PR #7088, opened 2020-05-19
Fix rsyslog in the dev env by using a local dir volume

wenottingham:
We broke this in the image refactor.


wenottingham:
(and by "we", I mean "me".)

squidboylan (approved):

rooftopcellist (approved):
I tested this locally and it works.

Issue #5571, opened 2019-12-24
Syncing all sources in parallel leads to missing hosts


binbashroot:
##### ISSUE TYPE
 - Bug Report

##### COMPONENT NAME
 - UI

##### SUMMARY

Using two completely separate satellite sources with the same "Organization" but different "Locations"  results in AWX only importing some, but not all,  of the hosts into a cumulative inventory.

##### ENVIRONMENT
* AWX version: 9.0.1.0
* AWX install method:  docker on linux
* Ansible version:  2.8.5
* Operating System: RHEL 7
* Web Browser: Chrome 79.0.3945.88 
* Red Hat Satellite 6.6

##### STEPS TO REPRODUCE

1. Inventories - > Add -> Enter Name ->Save -> Sources -> Add first Satellite server with creds using "Red Hat Satellite 6" for source-> Save
2. Inventories -> Select previously created inventory -> Sources -> Add -> Add 2nd Satellite server into source -> Save
3. Select Created Inventory -> Sources -> Click Sync All

"Overwrite, Overwrite Variables, or Update on Launch did not change results.

##### EXPECTED RESULTS
Expected a total of 350 hosts in Inventory.  

##### ACTUAL RESULTS
Total number of hosts ingested was 180.   Cumulative RH Satellite sources results in only a partial number of hosts in the inventory.   If two separate inventories are used each utilizing a separate Satellite server, the correct number of hosts are ingested in the inventory.

blomquisg:
@jladdjr check if this is fixed with your inventory plugin work

jladdjr:
Just ran through the steps to reproduce and I'm also finding that an inventory w/ two satellite inventory sources doesn't have the number of hosts I'd expect (a simple sum of the number of hosts returned by each inv. source). Digging to see what's going on. Wondering if this issue is specific to foreman or if it's a more general issue w/ awx inventories / inv. sources.

Note: In my testing, I used the new foreman plugin which awx is about to adopt.

ryanpetrello:
@elyezer @squidboylan @kdelee is this something any of you and yours have bandwidth to test now that we've merged the plugin?

If it doesn't "just work" in the new plugin, we're unlikely to fix it right now.

cc @blomquisg

jladdjr:
Got side-tracked, poking at this now.

jladdjr:
[...]

PR #7090, opened 2020-05-19
Updates to a given inventory should happen serially

jladdjr:
We need to ensure inventory updates for the same inventory happen serially.

Basic sanity test shows syncs happening serially. Need to follow-up with more tests to make sure there aren't any side-effects.

Note: Thought this might address https://github.com/ansible/awx/issues/5571, but I'm still seeing that issue w/ this fix.

### DIFF
--- a/awx/main/management/commands/inventory_import.py
+++ b/awx/main/management/commands/inventory_import.py
@@ -874,21 +874,20 @@ class Command(BaseCommand):
         Load inventory from in-memory groups to the database, overwriting or
         merging as appropriate.
         '''
-        with advisory_lock('inventory_{}_update'.format(self.inventory.id)):
-            # FIXME: Attribute changes to superuser?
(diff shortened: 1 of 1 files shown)


AlanCoding:
I think we should develop some means of testing, but I agree this should fix some issues.

AlanCoding (approved):
Confirmed that this fixes issue, not the issue you intended, but a bug nonetheless

jladdjr:
rebased, going to run this through some tests to make sure this still looks good

jladdjr:
The tests I ran for this are just about wrapped up. So far, I haven't seen anything come up that suggests this would introduce any regressions.

jladdjr:
Tests finished. Again, I'm not seeing any signs of a regression. 

Think this is good-to-go. cc @ryanpetrello @AlanCoding

Issue #526, opened 2017-10-30
Ability to store job templates in source control repo


jainnikhil30:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
There should be a way for job templates to be source controlled repo aka git.

##### STEPS TO REPRODUCE

This feature can used  if we have 2 ansible instances. Dev and prod. And then if we have a promotion model for all our playbooks from dev to prod. But then right now the job templates needs to be created manually.

perzizzle:
We have the environment you describe with a dev and production instance of tower for security reasons. One of the biggest problems we have is keeping job templates in sync. We have a series of scripts that use the tower api to validate the job templates are identical but they aren't perfect. We already store our inventory, and playbooks in source control.

shawnmason01:
Has anyone gotten this working successfully? I am needing to find a way to store my job templates in a git repo.

wenottingham:
https://github.com/ansible/awx/pull/7099 is a sample implementation of this.

This needs rethought/reworked in some ways:
- should use the new awx cli/modules import/export (as a format, and as an implementation)
- on import, it should probably be restricted to only creating templates & maybe inventories (even if more is in the export file)
- it needs a better authentication model (likely some sort of credential or user attached to the project)

There are more open questions that need thought through here.

gomker:
@wenottingham 
As a workaround I currenlty use existing Ansible Tower modules to export data to JSON and parse this data back into another tower to syncronize multiple Tower instances. In my use case I prefer to update existing template and create new when does not exist.
For credential managment I do this externally with other key management systems.

ryanpetrello:
@wenottingham it seems like there's a lot of overlap on this issue and https://github.com/ansible/awx/issues/578 - do we want to pick one and close the other?

PR #7099, opened 2020-05-20
[WIP] Project update sync assets

john-westcott-iv:
##### SUMMARY

This is an initial cut as having parts of Tower being controlled by code. A project update of an SCM based project will not attempt to create job templates and workflow job templates from files within the source. The current proposed files are:
    assets/job_templates.yml
    assets/workflow_job_template_nodes.yml
    assets/workflow_job_templates.yml
Which can be seen here: https://github.com/john-westcott-iv/test-playbooks/tree/master/assets

This PR adds a "Sync Assets" checkbox to the project UI screen. If enabled, during a manually invoked project sync the project_update.yml playbook will now include a role which will read these files and then leverage the awx.awx collection to build assets within tower.

Related #526
 

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API
 - UI

##### AWX VERSION

```
awx: 11.2.0
```

##### ADDITIONAL INFORMATION

TODO:
Sync with the awxkit import/export work to have the data files aligned. 

```

```


awx/api/serializers.py
```diff
@@ -1241,6 +1241,19 @@ def get_summary_fields(self, obj):
         return ret
 
 
+class DirectTokenSerializer():
```
ryanpetrello: I don't think this is something we'd ever expose to the API, so it's not really a "serializer" in that sense and probably doesn't belong in this file.  Given that you're just using `oauthlib.common.generate_token` and calling ORM code, why not just do _that_ instead (perhaps in a function) in `tasks.py` ?

awx/api/serializers.py
```diff
@@ -1241,6 +1241,19 @@ def get_summary_fields(self, obj):
...
+    def create_token(self, data):
+        if 'token' not in data:
+            data['token'] = generate_token()
+        if 'expires' not in data:
+            data['expires'] = now() + timedelta(
+                seconds=settings.OAUTH2_PROVIDER['ACCESS_TOKEN_EXPIRE_SECONDS']
```
ryanpetrello: Given that this is really intended to be an anonymous, single-use token, I think we should give it a *very* short expiration time, maybe something like 30 minutes (or however long we anticipate the actual sync process should take at maximum).

awx/playbooks/roles/sync_project/tasks/main.yml
```diff
@@ -0,0 +1,121 @@
...
+    ask_inventory_on_launch: "{{ item.ask_inventory_on_launch | default(omit) }}"
+    ask_job_type_on_launch: "{{ item.ask_job_type_on_launch | default(omit) }}"
+    ask_limit_on_launch: "{{ item.ask_limit_on_launch | default(omit) }}"
+    ask_scm_branch_on_launch: "{{ item.ask_scm_branch_on_launch | default(omit) }}"
+    ask_skip_tags_on_launch: "{{ item.ask_skip_tags_on_launch | default(omit) }}"
+    ask_tags_on_launch: "{{ item.ask_tags_on_launch | default(omit) }}"
+    ask_variables_on_launch: "{{ item.ask_variables_on_launch | default(omit) }}"
+    ask_verbosity_on_launch: "{{ item.ask_verbosity_on_launch | default(omit) }}"
+    become_enabled: "{{ item.become_enabled | default(omit) }}"
+    credential: "{{ item.credential | default(omit) }}"
```
ryanpetrello, line 32: If we're making something new, let's stop using this old singular `credential` key.

awx/playbooks/roles/sync_project/tasks/main.yml
```diff
@@ -0,0 +1,121 @@
...
+    project: "{{ awx_project }}"
+    scm_branch: "{{ item.scm_branch | default(omit) }}"
+    skip_tags: "{{ item.skip_tags | default(omit) }}"
+    start_at_task: "{{ item.start_at_task | default(omit) }}"
+    survey_enabled: "{{ item.survey_enabled | default(omit) }}"
+    survey_spec: "{{ item.survey_spec | default(omit) }}"
+    timeout: "{{ item.timeout | default(omit) }}"
+    tower_host: "{{ awx_host }}"
+    use_fact_cache: "{{ item.use_fact_cache | default(omit) }}"
+    vault_credential: "{{ item.vault_credential | default(omit) }}"
```
ryanpetrello, line 61: Same comment here.

awx/main/tasks.py
```diff
@@ -2393,6 +2410,14 @@ def post_run_hook(self, instance, status):
             if status == 'successful' and instance.launch_type != 'sync':
                 self._update_dependent_inventories(instance, dependent_inventory_sources)
 
+        # Clean up the oauth token if we are project syncing
+        if instance.project.sync_assets:
+            try:
+                OAuth2AccessToken.objects.filter(user=instance.created_by_id, description='project sync {}'.format(instance.id)).delete()
```
ryanpetrello: A pattern we usually use for this sort of thing is storing the reference to the token above:

https://github.com/ansible/awx/pull/7099/files#diff-9d4ea1dd908b35fb92eaede4bd10bb46R2080

...as `self.new_token = `

...and then down here, you can just `self.new_token.delete()`

awx/main/tasks.py
```diff
@@ -2393,6 +2410,14 @@ def post_run_hook(self, instance, status):
...
+            except Exception as e:
+                pass
```
ryanpetrello: If we fail to clean up the token for some reason, it's probably worth logging:

```python
logger.exception("Failed to cleanup OAuth2.0 token used for asset syncing")
```

awx/main/migrations/0116_v380_sync_assets.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.2.11 on 2020-05-04 02:26
+
+import awx.main.utils.polymorphic
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('main', '0115_v370_schedule_set_null'),
```
ryanpetrello, line 10: Just a heads up, @beeankha may beat you to claiming this `0116` number.

awx/api/serializers.py
john-westcott-iv: I thought about that initially but was thinking maybe this could be reused someday. How would you feel about having a get_token method in the User object? Then you could do something like: 
env['TOWER_OAUTH_TOKEN'] = User.objects.get(id=project_update.created_by_id).get_token('<description>')

awx/api/serializers.py
john-westcott-iv: This was something else I thought of but I wasn't sure how long. The default 1000 years is likely too much but I know there are some massive job syncs out there. Do we want to add another setting for this token? PROJECT_SYNC_ACCESS_TOKEN_EXPIRE_SECONDS and start with 24h as the default?

wenottingham: repeating from a different conversation - tying this to the launching user is tricky in the case of update-on-launch and scheduled project updates. We'd want to find a way to handle those

awx/playbooks/roles/sync_project/tasks/main.yml
john-westcott-iv: Good point, this was a bulk field copy from an `ansible-doc | grep ... | grep ...` command. For now I will leave this as is since this might be changing anyway to match the import/export code.

awx/main/tasks.py
john-westcott-iv: Fixed.

awx/main/tasks.py
john-westcott-iv: Added a message for both creation and deletion of tokens.

awx/api/serializers.py
john-westcott-iv: In the end I put this into the models/oauth as the User object is from django, not AWX.

john-westcott-iv:
This has been superseded by PR #7099

AndrewSav:
@john-westcott-iv what PR this has been superseded by? (It seems there is a mistake in the message above)

Issue #5996, opened 2020-02-19
Inventory should be removable on a JT


mabashian:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
We probably made it required initially because we didn't have the prompt on launch option in the form.  When this option becomes available we need to make the inventory removable.

The rule for inventory on JT is that either:

a) A default inventory is provided
or
b) Prompt on launch is checked

As a result, the user should be able to remove the selected inventory.

Marking as blocked until https://github.com/ansible/awx/issues/4967 lands

mabashian:
I went ahead and addressed this in #4967

dsesami:
conferred with @mabashian on this. There was a miscommunication, and this has not been implemented. bouncing back to needs_devel

dsesami:
This looks really good and handles all variations of selecting and unselecting prompt on launch for inventories.

PR #7101, opened 2020-05-20
make Inventory optional if prompt-on-launch is true

keithjgrant:
##### SUMMARY
JT Form: Makes Inventory field optional if "Prompt on Launch" is selected

Addresses #5996 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/screens/Template/shared/JobTemplateForm.jsx
+++ b/awx/ui_next/src/screens/Template/shared/JobTemplateForm.jsx
@@ -233,7 +233,7 @@ function JobTemplateForm({
         </FieldWithPrompt>
         <FieldWithPrompt
           fieldId="template-inventory"
-          isRequired
+          isRequired={!formikValues.ask_inventory_on_launch}
           label={i18n._(t`Inventory`)}
           promptId="template-ask-inventory-on-launch"
           promptName="ask_inventory_on_launch"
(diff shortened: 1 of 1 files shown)


awx/ui_next/src/screens/Template/shared/JobTemplateForm.jsx
```diff
@@ -244,10 +244,10 @@ function JobTemplateForm({
             value={inventory}
             onBlur={() => inventoryHelpers.setTouched()}
             onChange={value => {
-              inventoryHelpers.setValue(value.id);
+              inventoryHelpers.setValue(value ? value.id : null);
```
AlexSCorey, line 247: Not necessarily within the scope of this PR but I do think there is some value in setting the value of the Inventory (formik value) to the whole inventory object and then parsing through that object on submit.  It would would eliminate a 2nd state handler like we have on line 248.

AlexSCorey (approved):
Looks good to me.

awx/ui_next/src/screens/Template/shared/JobTemplateForm.jsx
keithjgrant: Yeah, I think I agree, though this fix would be bigger than the rest of this PR, so I'll defer it for later.

jakemcdermott (approved):

dsesami (approved):
works on add/edit. looks good to me.

Issue #7102, opened 2020-05-20
User/Team Access List Item 


AlexSCorey:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
 ~On the User/Team -> Access list if a user has multiple permission on a single resource, all of those permissions should be shown in a chip, on the list.~ It was decided not to do this 1 because of complications it would add to pagination, search etc.

- [x] There should be spacing between the label on the list item and the data 
![Screen Shot 2020-05-20 at 1 15 46 PM](https://user-images.githubusercontent.com/39280967/82476435-0e142b80-9a9c-11ea-991c-81502cab2c0e.png)

AlexSCorey:
completed by https://github.com/ansible/awx/pull/7183

Issue #6921, opened 2020-05-04
Users/Teams Access Add


AlexSCorey:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
See https://github.com/ansible/awx/issues/6597

PR #7103, opened 2020-05-20
Adds support to user and team access add

AlexSCorey:
##### SUMMARY
Adds support to user/team access add.  #6921.  This PR brought this issue to my attention. https://github.com/ansible/awx/issues/7102

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
add


nixocio:
I got this code locally, it seems to be sane the changes. I followed up with @AlexSCorey , and we will probably render the list of `access` a bit different  for `system administrators`.
image


 But it will not be addressed in this PR. Good job @AlexSCorey .

awx/ui_next/src/components/UserAccessAdd/UserAndTeamAccessAdd.jsx
```diff
@@ -0,0 +1,155 @@
...
+import getResourceTypes from './resources.data';
+
+const Grid = styled.div`
+  display: grid;
+  grid-gap: 20px;
+  grid-template-columns: 33% 33% 33%;
+  grid-template-columns: repeat(auto-fill, minmax(250px, 1fr));
+`;
+
+function UserAndTeamAccessAdd({
```
jakemcdermott: I think this is a fine and descriptive name for the component. Would you also rename the component directory to match? (`UserAccessAdd` --> `UserAndTeamAccessAdd`)

awx/ui_next/src/components/UserAccessAdd/resources.data.jsx
```diff
@@ -0,0 +1,208 @@
...
+import {
+  JobTemplatesAPI,
+  WorkflowJobTemplatesAPI,
+  CredentialsAPI,
+  InventoriesAPI,
+  ProjectsAPI,
+  OrganizationsAPI,
+} from '../../api';
+
+export default function getResourceTypes(i18n) {
```
jakemcdermott: I like what you're doing here, but I must nitpick the name of this function as well as the file you put it in. 

- This function returns configuration for the search component. What about `getSearchConfigs`, `getResourceSearchConfigs`, or something like that? 

- `resources.data.jsx` doesn't explain much. It also reuses a convention we've established for test data (`xyz.data..`) for an entirely different purpose, which I don't love. Since this module only exports a single function, why not name it after the function it exports? (`getResourceTypes.jsx` or whatever you rename the function to.)

jakemcdermott (changes requested):
From a code review perspective I think this looks good and I only have a few nitpicks for silly stuff like naming (see my review comments). Nice job here!

I _also_ pulled this down and tested it and noticed a few things that we'd probably want to address before merging:
- 'The Add User/Team Permissions' prompt lets you advance to the next step without selecting a resource. Instead of showing a blank page and making the user click `back`, can you make it so the `next` button is disabled on the first page until they select something?
- There is no loading indicator when loading second page of the prompt sequence. In general we _always_ need loading indicators when fetching content but it is particularly noticeable on this view when I tried it because the api requests involved are relatively heavy and take a moment to load.
- Applying search filters to the list of resources doesn't seem to work

awx/ui_next/src/components/UserAccessAdd/UserAndTeamAccessAdd.test.jsx
```diff
@@ -0,0 +1,225 @@
...
+
+    await act(async () =>
+      wrapper.find('Button[type="submit"]').prop('onClick')()
+    );
+
+    await expect(UsersAPI.associateRole).toHaveBeenCalled();
+    wrapper.update();
+    expect(wrapper.find('AlertModal').length).toBe(1);
+  });
+});
```
jakemcdermott: Solid tests :+1: nice work

awx/ui_next/src/components/UserAccessAdd/resources.data.jsx
AlexSCorey: > * There is no loading indicator when loading second page of the prompt sequence. In general we _always_ need loading indicators when fetching content but it is particularly noticeable on this view when I tried it because the api requests involved are relatively heavy and take a moment to load.

Yes I can add a loading indicator.  That component that creates that second view needs to be brought up to date and implement hooks.  I was going to do that, actually had the code written out for it, but decided that it didn't really belong in this PR.  Now I have an excused to bring it up to do and fix that loading issue!!

> * Applying search filters to the list of resources doesn't seem to work

That's weird. I wonder whats going on there.  Will dig in.

AlexSCorey:
> * There is no loading indicator when loading second page of the prompt sequence. In general we _always_ need loading indicators when fetching content but it is particularly noticeable on this view when I tried it because the api requests involved are relatively heavy and take a moment to load.

Yes I can add a loading indicator.  That component that creates that second view needs to be brought up to date and implement hooks.  I was going to do that, actually had the code written out for it, but decided that it didn't really belong in this PR.  Now I have an excused to bring it up to do and fix that loading issue!!

> * Applying search filters to the list of resources doesn't seem to work

That's weird. I wonder whats going on there.  Will dig in.

jakemcdermott (approved):
:+1:

dsesami:
https://github.com/ansible/awx/issues/7102 addresses most of the comments I was going to make :)

I'm making the following non-blocking, since I just need clarification:
- I don't know if this is patternfly-specific, but if you click a checkbox in the wizard (like when you're picking what roles to apply), only the text responds to clicks. If you click in the whitespace within the border, it can't be clicked. That feels non-intuitive, since the whole thing feels like a large button.

- Is there a way to remove roles? It feels like that should be included in this PR. If not, what's the link to it?

dsesami (approved):

Issue #7107, opened 2020-05-20
awxkit install fails on local clones


dsesami:
Note: i heard of another issue about this, but i couldn't find it. please link them together if you know the issue.

##### ISSUE TYPE
 - Bug Report

##### SUMMARY
`pip` does not honor symlinks in local clones. It got merged three weeks ago and they're going to have to revert it: https://github.com/pypa/pip/issues/7555

The issue has been somewhat existent since 2016: https://github.com/pypa/pip/issues/3500

Current workaround is rm awx/awxkit/VERSION and cp awx/VERSION awx/awxkit/VERSION.

This issue should be about either acknowledging the pip issue and monitoring it via this repo, or making the VERSION file not a symlink. 

##### STEPS TO REPRODUCE
run `pip install awx/awxkit`. 

##### EXPECTED RESULTS
Install succeeds.

##### ACTUAL RESULTS
VERSION file isn't found, failure occurs.

dsesami:
the commit that it's related to is https://github.com/ansible/awx/commit/296995e8bd62ff716223f352c2bbb96b55db7217

FileNotFoundError: Command "python setup.py egg_info" can't find (long temp path with VERSION at the end)

dsesami:
this got fixed. local clones work now.

PR #7114, opened 2020-05-21
don't use symlinks for awxkit VERSION

ryanpetrello:
as much as I'd like this to just work, certain versions of pip just
don't support it -_-

see: https://github.com/ansible/awx/issues/7107
see: https://github.com/pypa/pip/issues/7555

so unfortunately, we just have to track these as two distinct files and
just keep them in sync


shanemcd (approved):

kdelee (approved):
hallelujah

thought I was crazy but turns out it was a bug

dsesami (approved):

AlanCoding (approved):

Issue #7036, opened 2020-05-15
POC - Serve ui_next with same port as old_ui and different paths


nixocio:
##### ISSUE TYPE
 - Modify the current docker environment to serve both SPA in the same port and different paths.

##### SUMMARY
- Things to take in consideration:
Routing in both SPA and Django
Static Files Collection

This is the starting point: 
https://github.com/jakemcdermott/awx/tree/ui-next-integration

##### GOALS
Create a POC to show in the planning meeting with API team.

jakemcdermott:
:heavy_check_mark:

PR #7115, opened 2020-05-21
UI next integration

nixocio:
Add a new routing system to serve the new UI, without remove the old UI.

* Add  a new django app related to the new UI. Not affecting the usage of the old UI.
* Add a catch-all URL with the prefix `next`. 
* Add a `basename` to the routes of the the new UI URL, thus connecting them to the django new app.
* Modify the layout of images, to be built as they will be served later on.

See: https://github.com/ansible/awx/issues/7036


awx/ui_next/src/RootProvider.jsx
```diff
@@ -15,10 +15,10 @@ class RootProvider extends Component {
     const language = getLanguageWithoutRegionCode(navigator);
 
     return (
-      <HashRouter>
+      <HashRouter basename="/next">
```
keithjgrant: I noticed the URL ends up looking like `next/#next/users` with `next` both before and after the hash. Do we need it after the hash also?

tools/docker-compose/nginx.vh.default.conf
```diff
@@ -22,8 +22,6 @@ server {
 
     # HSTS (ngx_http_headers_module is required) (15768000 seconds = 6 months)
     add_header Strict-Transport-Security max-age=15768000;
-    add_header Content-Security-Policy "default-src 'self'; connect-src 'self' ws: wss:; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' *.pendo.io; img-src 'self' *.pendo.io data:; report-uri /csp-violation/";
```
matburt: Why remove the content security policy?

nixocio: The end goal is to keep it, it was just removed for demo purposes.

awx/ui_next/src/RootProvider.jsx
nixocio: That is how the front end routing system and back end got connected. Django is routing any URL that has `next` to the template `index.html`. The `index.html` has the link for the static files from the `npm run build`. Then frontend routing system now has this keyword `next` to be connected to an unique URL in backend always serving the same file. I think if we use `BrowserRouter` instead of the `HashRouter` we will not have this issue, probably.  But this was a temporary solution.

tools/docker-compose/nginx.vh.default.conf
jakemcdermott: Yep, these changes were necessary to demo but we don't intend to actually change this file.

[One of the patternfly components](https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/ChipGroup/ChipGroup.tsx#L120) is violating the CSP and preventing the UI from being served because it uses [a utility function that is unsafe](https://github.com/patternfly/patternfly-react/blob/2263de8361b196f4421a3adcbb01b4b2dd410d4a/packages/react-core/src/helpers/util.ts#L116).

Merging [awx/#7068](https://github.com/ansible/awx/pull/7068) _should_ clear this up because it will include [this commit](https://github.com/patternfly/patternfly-react/commit/1900655a54680a9cdd13a90db02b6005e1851e05#diff-d3283522647dde0024e18db90fe7b1ad) from patternfly-react, which removes the `new Function` call. I've verified this with some light manual testing locally but we'll also want qe to take a look.

awx/ui_next/src/RootProvider.jsx
jakemcdermott: @nixocio @keithjgrant I think we just need to change `basename` --> `baseName` and this should work with just the one leading `/next`:

ui-next-integration-01

jakemcdermott: @nixocio @keithjgrant On second look, basename is the correct prop name so perhaps we don't need it? :man_shrugging: works for me if that is the case. 

Another possibility is that we _may_ only need to set `basename` when running the development server (port 3000) and not for production builds. I suspect this is the case but haven't confirmed.

nixocio: You are right @jakemcdermott . I removed the `basename` and just worked fine. Thanks for verifying this.

jlmitch5 (approved):
this is awesome stuff!!

jlmitch5:
I just meant to leave a comment not an "approval", sorry abou that

awx/ui_next/src/RootProvider.jsx
jakemcdermott: Added `basenname` back in after switching to `BrowserRouter`

jakemcdermott:
regate

nixocio:
both_uis

PR #7116, opened 2020-05-21
Initial cut at tower_token module

john-westcott-iv:
##### SUMMARY

Initial creation of tower_token modules. 

I decided to mimic the ovirt module behavior of setting an ansible fact which can then be used directly in the tower_oatuhtoken parameter of other modules.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - Collection

##### AWX VERSION

```
awx: 11.2.0
```

##### ADDITIONAL INFORMATION

```
Blocked by https://github.com/ansible/awx/pull/7118. Once that is merged in we can fix the module linting problems with ease.
```

### DIFF
--- a/awx_collection/plugins/doc_fragments/auth.py
+++ b/awx_collection/plugins/doc_fragments/auth.py
@@ -31,6 +31,9 @@ options:
   tower_oauthtoken:
     description:
     - The Tower OAuth token to use.
+    - This value can be in one of two formats.
+    - A string which is the token itself. (i.e. bqV5txm97wqJqtkxlMkhQz0pKhRMMX)
+    - A dictionary structure as returned by the tower_token module.
     - If value not set, will try environment variable C(TOWER_OAUTH_TOKEN) and then config files
     type: raw
     version_added: "3.7"
--- a/awx_collection/plugins/doc_fragments/auth.py
+++ b/awx_collection/plugins/doc_fragments/auth.py
@@ -32,7 +32,8 @@ options:
     description:
     - The Tower OAuth token to use.
     - If value not set, will try environment variable C(TOWER_OAUTH_TOKEN) and then config files
-    type: str
+    type: raw
+    version_added: "3.7"
   validate_certs:
     description:
     - Whether to allow insecure connections to Tower or AWX.
--- a/awx_collection/plugins/module_utils/tower_api.py
+++ b/awx_collection/plugins/module_utils/tower_api.py
@@ -47,7 +47,7 @@ class TowerModule(AnsibleModule):
         tower_username=dict(required=False, fallback=(env_fallback, ['TOWER_USERNAME'])),
         tower_password=dict(no_log=True, required=False, fallback=(env_fallback, ['TOWER_PASSWORD'])),
         validate_certs=dict(type='bool', aliases=['tower_verify_ssl'], required=False, fallback=(env_fallback, ['TOWER_VERIFY_SSL'])),
-        tower_oauthtoken=dict(type='str', no_log=True, required=False, fallback=(env_fallback, ['TOWER_OAUTH_TOKEN'])),
+        tower_oauthtoken=dict(type='raw', no_log=True, required=False, fallback=(env_fallback, ['TOWER_OAUTH_TOKEN'])),
         tower_config_file=dict(type='path', required=False, default=None),
     )
     short_params = {
--- a/awx_collection/plugins/module_utils/tower_api.py
+++ b/awx_collection/plugins/module_utils/tower_api.py
@@ -3,7 +3,7 @@ __metaclass__ = type
 
 from ansible.module_utils.basic import AnsibleModule, env_fallback
 from ansible.module_utils.urls import Request, SSLValidationError, ConnectionError
-from ansible.module_utils.six import PY2
+from ansible.module_utils.six import PY2, string_types
 from ansible.module_utils.six.moves import StringIO
 from ansible.module_utils.six.moves.urllib.parse import urlparse, urlencode
 from ansible.module_utils.six.moves.urllib.error import HTTPError
@@ -98,10 +98,16 @@ class TowerModule(AnsibleModule):
 
         # Perform magic depending on whether tower_oauthtoken is a string or a dict
         if self.params.get('tower_oauthtoken'):
-            if type(self.params.get('tower_oauthtoken')) is dict:
-                setattr(self, 'oauth_token', self.params.get('tower_oauthtoken')['token'])
-            elif 'token' in self.params.get('tower_oauthtoken'):
-                setattr(self, 'oauth_token', self.params.get('tower_oauthtoken'))
+            token_param = self.params.get('tower_oauthtoken')
+            if type(token_param) is dict:
+                if 'token' in token_param:
+                    self.oauth_token = self.params.get('tower_oauthtoken')['token']
+                else:
+                    self.fail_json(msg="The provided dict in tower_oauthtoken did not properly contain the token entry")
+            elif isinstance(token_param, string_types):
+                self.oauth_token = self.params.get('tower_oauthtoken')
+            else:
+                self.fail_json(msg="The provided tower_oauthtoken type was not valid ({0}), please refer to ansible-doc for valid options".format(type(token_param).__name__))
 
         # Perform some basic validation
         if not re.match('^https{0,1}://', self.host):
--- a/awx_collection/plugins/module_utils/tower_api.py
+++ b/awx_collection/plugins/module_utils/tower_api.py
@@ -107,7 +107,8 @@ class TowerModule(AnsibleModule):
             elif isinstance(token_param, string_types):
                 self.oauth_token = self.params.get('tower_oauthtoken')
             else:
-                self.fail_json(msg="The provided tower_oauthtoken type was not valid ({0}), please refer to ansible-doc for valid options".format(type(token_param).__name__))
+                error_msg = "The provided tower_oauthtoken type was not valid ({0}). Valid options are str or dict.".format(type(token_param).__name__)
+                self.fail_json(msg=error_msg)
 
         # Perform some basic validation
         if not re.match('^https{0,1}://', self.host):
--- a/awx_collection/plugins/module_utils/tower_api.py
+++ b/awx_collection/plugins/module_utils/tower_api.py
@@ -96,6 +96,13 @@ class TowerModule(AnsibleModule):
             if direct_value is not None:
                 setattr(self, short_param, direct_value)
 
+        # Perform magic depending on whether tower_oauthtoken is a string or a dict
+        if self.params.get('tower_oauthtoken'):
+            if type(self.params.get('tower_oauthtoken')) is dict:
+                setattr(self, 'oauth_token', self.params.get('tower_oauthtoken')['token'])
+            elif 'token' in self.params.get('tower_oauthtoken'):
+                setattr(self, 'oauth_token', self.params.get('tower_oauthtoken'))
+
         # Perform some basic validation
         if not re.match('^https{0,1}://', self.host):
             self.host = "https://{0}".format(self.host)
@@ -504,6 +511,9 @@ class TowerModule(AnsibleModule):
                 item_name = existing_item['username']
             elif 'identifier' in existing_item:
                 item_name = existing_item['identifier']
+            elif item_type == 'o_auth2_access_token':
+                # An oauth2 token has no name, instead we will use its id for any of the messages
+                item_name = existing_item['id']
             else:
                 self.fail_json(msg="Unable to process delete of {0} due to missing name".format(item_type))
 
--- a/awx_collection/plugins/modules/tower_token.py
+++ b/awx_collection/plugins/modules/tower_token.py
@@ -20,8 +20,9 @@ author: "John Westcott IV (@john-westcott-iv)"
 version_added: "2.3"
 short_description: create, update, or destroy Ansible Tower tokens.
 description:
-    - Create, update, or destroy Ansible Tower tokens. See
+    - Create or destroy Ansible Tower tokens. See
       U(https://www.ansible.com/tower) for an overview.
+    - If you create a token it is your responsibility to delete the token.
 options:
     description:
       description:
@@ -60,22 +61,24 @@ extends_documentation_fragment: awx.awx.auth
 '''
 
 EXAMPLES = '''
-- name: Create a new token using an existing token
-  tower_token:
-    description: '{{ token_description }}'
-    scope: "write"
-    state: present
-    tower_oauthtoken: "{{ ny_existing_token }}"
-  register: new_token
+- block:
+    - name: Create a new token using an existing token
+      tower_token:
+        description: '{{ token_description }}'
+        scope: "write"
+        state: present
+        tower_oauthtoken: "{{ ny_existing_token }}"
 
-- name: Use our new token to make another call
-  tower_job_list:
-    tower_oauthtoken: "{{ tower_token }}"
+    - name: Use our new token to make another call
+      tower_job_list:
+        tower_oauthtoken: "{{ tower_token }}"
 
-- name: Delete our Token with the token we created
-  tower_token:
-    existing_token: "{{ tower_token }}"
-    state: absent
+  always:
+    - name: Delete our Token with the token we created
+      tower_token:
+        existing_token: "{{ tower_token }}"
+        state: absent
+      when: tower_token is defined
 '''
 
 RETURNS = '''
--- a/awx_collection/plugins/modules/tower_token.py
+++ b/awx_collection/plugins/modules/tower_token.py
@@ -98,7 +98,7 @@ EXAMPLES = '''
       when: tower_token is defined
 
 - name: Delete a token by its id
-  tower_toekn:
+  tower_token:
     existing_token_id: 4
     state: absent
 '''
--- a/awx_collection/plugins/modules/tower_token.py
+++ b/awx_collection/plugins/modules/tower_token.py
@@ -22,7 +22,10 @@ short_description: create, update, or destroy Ansible Tower tokens.
 description:
     - Create or destroy Ansible Tower tokens. See
       U(https://www.ansible.com/tower) for an overview.
-    - If you create a token it is your responsibility to delete the token.
+    - In addition, the module sets an Ansible fact which can be passed into other
+      tower_* modules as the parameter tower_oauthtoken. See examples for usage.
+    - Because of the sensitive nature of tokens, the created token value is only available once
+      through the Ansible fact. (See RETURN for details)
 options:
     description:
       description:
@@ -88,10 +91,14 @@ EXAMPLES = '''
       when: tower_token is defined
 '''
 
-RETURNS = '''
+RETURN = '''
 tower_token:
   type: dict
-  description: A Tower token object which can be used for auth or token deletion
+  description: An Ansible Fact variable representing a Tower token object which can be used for auth in subsequent modules. See examples for usage.
+  contains:
+    token:
+      description: The token that was generated. This token can never be accessed again, make sure this value is noted before it is lost.
+      type: str
   returned: on successful create
 '''
 
@@ -103,7 +110,6 @@ def return_token(module, last_response):
     # So the default module return would give you an ID but then the token would forever be masked on you.
     # This method will return the entire token object we got back so that a user has access to the token
 
-    module.json_output['token'] = last_response['token']
     module.json_output['ansible_facts'] = {
         'tower_token': last_response,
     }
--- a/awx_collection/plugins/modules/tower_token.py
+++ b/awx_collection/plugins/modules/tower_token.py
@@ -61,7 +61,7 @@ EXAMPLES = '''
         description: '{{ token_description }}'
         scope: "write"
         state: present
-        tower_oauthtoken: "{{ ny_existing_token }}"
+        tower_oauthtoken: "{{ my_existing_token }}"
 
     - name: Delete this token
       tower_token:
@@ -73,8 +73,8 @@ EXAMPLES = '''
         description: '{{ token_description }}'
         scope: "write"
         state: present
-        tower_username: "{{ ny_username }}"
-        tower_password: "{{ ny_password }}"
+        tower_username: "{{ my_username }}"
+        tower_password: "{{ my_password }}"
 
     - name: Use our new token to make another call
       tower_job_list:
--- a/awx_collection/plugins/modules/tower_token.py
+++ b/awx_collection/plugins/modules/tower_token.py
@@ -26,6 +26,10 @@ description:
       tower_* modules as the parameter tower_oauthtoken. See examples for usage.
     - Because of the sensitive nature of tokens, the created token value is only available once
       through the Ansible fact. (See RETURN for details)
+    - Due to the nature of tokens in Tower this module is not idempotent. A second will
+      with the same parameters will create a new token.
+    - If you are creating a temporary token for use with modules you should delete the token
+      when you are done with it. See the example for how to do it.
 options:
     description:
       description:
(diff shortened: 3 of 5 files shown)


awx_collection/plugins/modules/tower_token.py
```diff
@@ -0,0 +1,157 @@
...
+'''
+
+EXAMPLES = '''
+- name: Create a new token using an existing token
+  tower_token:
+    description: '{{ token_description }}'
+    scope: "write"
+    state: present
+    tower_oauthtoken: "{{ ny_existing_token }}"
+  register: new_token
```
kedark3: Is there a way to create token with username/password ? :) Absolutely Great work otherwise. Sincere thanks for Tower Collections team for taking customer feedback seriously :100:

john-westcott-iv: Yes, just use the tower_username/tower_password parameters instead of the tower_oauthtoken.

kedark3: Awesome. Do you mind adding and example,it will be tremendously useful.

awx_collection/plugins/modules/tower_token.py
```diff
@@ -0,0 +1,168 @@
...
+    tower_oauthtoken:
+      description:
+        - The Tower OAuth token to use.
+      required: False
+      type: str
+      version_added: "3.7"
+extends_documentation_fragment: awx.awx.auth
+'''
+
+EXAMPLES = '''
```
mshriver, line 67: Thanks for adding the Examples block, good timing.

awx_collection/plugins/modules/tower_token.py
john-westcott-iv: Done.

beeankha (approved):

awx_collection/plugins/doc_fragments/auth.py
```diff
@@ -32,7 +32,7 @@ class ModuleDocFragment(object):
     description:
     - The Tower OAuth token to use.
     - If value not set, will try environment variable C(TOWER_OAUTH_TOKEN) and then config files
-    type: str
+    type: raw
```
AlanCoding, line 38: if this type is raw, I want to see some information in the `description:` with direction to the user. When is it not a string, and what format should be expected in that case?

awx_collection/plugins/modules/tower_token.py
```diff
@@ -0,0 +1,162 @@
...
+- block:
+    - name: Create a new token using an existing token
+      tower_token:
+        description: '{{ token_description }}'
+        scope: "write"
+        state: present
+        tower_oauthtoken: "{{ ny_existing_token }}"
```
AlanCoding: `ny_existing_token` -> `my_existing_token`?

awx_collection/plugins/modules/tower_token.py
```diff
@@ -0,0 +1,162 @@
...
+    - name: Use our new token to make another call
+      tower_job_list:
+        tower_oauthtoken: "{{ tower_token }}"
+
+  always:
+    - name: Delete our Token with the token we created
+      tower_token:
+        existing_token: "{{ tower_token }}"
+        state: absent
+      when: tower_token is defined
```
AlanCoding, line 98: this looks like a duplicated task with "Delete this token"

awx_collection/plugins/modules/tower_token.py
```diff
@@ -0,0 +1,162 @@
...
+      tower_token:
+        description: '{{ token_description }}'
+        scope: "write"
+        state: present
+        tower_username: "{{ ny_username }}"
+        tower_password: "{{ ny_password }}"
+
+    - name: Use our new token to make another call
+      tower_job_list:
+        tower_oauthtoken: "{{ tower_token }}"
```
AlanCoding, line 91: I think the first task needs to add `register: result`, and then this task reference `"{{ result.tower_token }}"`, I don't think things can just get crammed into the globals. Not as a module plugin... maybe as an action plugin, but that would be hard to follow.

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -84,7 +84,10 @@ def __init__(self, argument_spec, **kwargs):
         if self.params.get('validate_certs') is not None:
             self.verify_ssl = self.params.get('validate_certs')
         if self.params.get('tower_oauthtoken'):
-            self.oauth_token = self.params.get('tower_oauthtoken')
+            if type(self.params.get('tower_oauthtoken')) is dict:
+                self.oauth_token = self.params.get('tower_oauthtoken')['token']
+            elif 'token' in self.params.get('tower_oauthtoken'):
+                self.oauth_token = self.params.get('tower_oauthtoken')
```
AlanCoding, line 108: After this line, the code should validate type of `self.oauth_token`, because it's no longer automatic. Here, only a string type should be acceptable.

awx_collection/tests/integration/targets/tower_token/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Create a Token
+      tower_token:
+        description: '{{ token_description }}'
+        scope: "write"
+        state: present
+      register: new_token
+
+    - name: Validate our token works by token
+      tower_job_list:
+        tower_oauthtoken: "{{ tower_token.token }}"
```
AlanCoding, line 66: I think this should be what was registered, and the documented return variable, `new_token.tower_token`

awx_collection/plugins/doc_fragments/auth.py
john-westcott-iv: Changed it to:
- tower_oauthtoken
        The Tower OAuth token to use.
        This value can be in one of two formats.
        A string which is the token itself. (i.e. bqV5txm97wqJqtkxlMkhQz0pKhRMMX)
        A dictionary structure as returned by the tower_token module.
        If value not set, will try environment variable `TOWER_OAUTH_TOKEN' and then config files

awx_collection/plugins/modules/tower_token.py
john-westcott-iv: Fixed, there was also a ny_password and ny_username.

awx_collection/plugins/modules/tower_token.py
john-westcott-iv: In the example we actually pull two tokens:
* block
  * pull a token
  * delete the token we pulled
  * pull a second token
  * use the second token to call tower_job_list
* always
  * delete the second token

awx_collection/plugins/modules/tower_token.py
john-westcott-iv: Up on line 107 we are doing the injection of the tower_token variable into the ansible_facts with this code snippet:
```
    module.json_output['ansible_facts'] = {
        'tower_token': last_response,
    }
```
This is similar the same as to what the ovirt_auth module does: https://github.com/ansible/ansible/blob/stable-2.9/lib/ansible/modules/cloud/ovirt/ovirt_auth.py#L279

awx_collection/tests/integration/targets/tower_token/tasks/main.yml
john-westcott-iv: We do document the return token as a dict (but not whats inside of the dict):
```
tower_token:
  type: dict
  description: A Tower token object which can be used for auth or token deletion
  returned: on successful create
```

awx_collection/plugins/module_utils/tower_api.py
john-westcott-iv: Not sure how you got the code like this, maybe an old commit? It looks like this:
```
        if self.params.get('tower_oauthtoken'):
            if type(self.params.get('tower_oauthtoken')) is dict:
                self.oauth_token = self.params.get('tower_oauthtoken')['token']
            elif 'token' in self.params.get('tower_oauthtoken'):
                self.oauth_token = self.params.get('tower_oauthtoken')
```

john-westcott-iv: We could I guess try and raise an exception if we fall down into an "else" condition here. LMK if you want me to do that.

awx_collection/plugins/doc_fragments/auth.py
AlanCoding: 👍  to that description

but for whatever reason, I'm not seeing that in the current diff

awx_collection/plugins/module_utils/tower_api.py
AlanCoding: yeah, I'm also not really quite sure what's going on here, and if you actually intended this diff.

Backing up, what you're doing is turning off the validation for the token parameter. That's fine is documented, and we give clean error messages when it doesn't fit the expected pattern.

In several ways, it seems like this still isn't finished, or at least the validation isn't.

What are our expectations?

 - If it is a dict, then it needs to have the key "token", it can have other keys, but this is a minimum condition. Ansible handles `KeyError`s poorly, so if it's a dict, but that dict doesn't have the token key, this should fail with a message to document the failure of expectations.
 - If it's a string, then it should be used as-is
 - If it's a list, or any other type, then it should fail, repeating or referring to the basic documentation of the token parameter

The current code has a few head-scratchers. where is it handed as a string? `elif 'token' in self.params.get('tower_oauthtoken'):` is assuming a dict, but the prior conditional is also assuming a dict.

awx_collection/plugins/doc_fragments/auth.py
john-westcott-iv: @beeankha asked me to hold off on the push until she finished up a yolo run. This will be up later today.

awx_collection/plugins/modules/tower_token.py
```diff
@@ -0,0 +1,162 @@
...
+      tower_token:
+        existing_token: "{{ tower_token }}"
+        state: absent
+      when: tower_token is defined
+'''
+
+RETURNS = '''
+tower_token:
+  type: dict
+  description: A Tower token object which can be used for auth or token deletion
```
AlanCoding: The ovirt example you link documents the type as `complex` and gives the sub-keys

https://github.com/ansible/ansible/blob/stable-2.9/lib/ansible/modules/cloud/ovirt/ovirt_auth.py#L158

I actually think that would be good. Even if it gave more keys than what are documented, this would be a good way to specify the 2 functional items - the token and the token_id. In the parameter documentation we discussed

> A dictionary structure as returned by the tower_token module.

It just defers here. We also don't have any real examples of the dictionary typed out. I think it needs to be somewhere more clearly stated that (1) the function of this data structure is to contain the token and the token id, and what the key names for those are and (2) also mention in the return docs that this is returned both in this key and in the ansible_facts.

awx_collection/plugins/modules/tower_token.py
```diff
@@ -0,0 +1,162 @@
...
+
+DOCUMENTATION = '''
+---
+module: tower_token
+author: "John Westcott IV (@john-westcott-iv)"
+version_added: "2.3"
+short_description: create, update, or destroy Ansible Tower tokens.
+description:
+    - Create or destroy Ansible Tower tokens. See
+      U(https://www.ansible.com/tower) for an overview.
```
AlanCoding, line 24: I feel like this is not enough detail, see your ovirt example

https://github.com/ansible/ansible/blob/stable-2.9/lib/ansible/modules/cloud/ovirt/ovirt_auth.py#L19-L20

> which should be later used in all other oVirt/RHV modules, so all modules don't need to perform login and logout.

It is implied in its wording that it's intended that you use the module, and then don't pass the token value manually, because it goes through ansible_facts. You can say any variation on this, but it needs to be said.

awx_collection/plugins/modules/tower_token.py
AlanCoding: doesn't look like that got pushed yet?

awx_collection/plugins/modules/tower_token.py
AlanCoding: but then if the task "Create a new token using username/password" failed, then it would try to delete the 1st token twice, failing somewhat erroneously.

Understanding better what this is trying to do, I think it would be best to split it into 2 separate examples, where they both delete the token in the end with the `always:` block, because that matches better with what we're trying to recommend.

Also, when you do the `tower_job_list`, it isn't necessary to pass the `tower_oauthtoken` parameter, right? Or am I misunderstanding?

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -487,6 +490,9 @@ def delete_if_needed(self, existing_item, on_delete=None):
                 item_name = existing_item['username']
             elif 'identifier' in existing_item:
                 item_name = existing_item['identifier']
+            elif item_type == 'o_auth2_access_token':
```
squidboylan, line 521: im not sure we'll hit this, testing locally when i output the token object i get:

```
TASK [tower_token : debug] ****************************************************************************************************************************************************************************************************************************************************************************
ok: [testhost] => {
    "msg": {                                                                                                                                                                                                                                                                                          
        "ansible_facts": {                                                                                                                                                                                                                                                                            
            "tower_token": {
                "application": null,                                                                                                                                                                                                                                                                  
                "created": "2020-06-01T18:48:22.790373Z",
                "description": "AWX-Collection-tests-tower_token-description-NtDlNNHIgxVWVYJp",                                                                                                                                                                                                       
                "expires": "3019-10-03T18:48:22.787511Z",
                "id": 461,                                                                                                                                                                                                                                                                            
                "modified": "2020-06-01T18:48:22.799678Z",                                                                                                                                                                                                                                            
                "refresh_token": null,
                "related": {                                                                                                                                                                                                                                                                          
                    "activity_stream": "/api/v2/tokens/461/activity_stream/",
                    "user": "/api/v2/users/1/"
                },
                "scope": "write",
                "summary_fields": {                                                                                                                                                                                                                                                                   
                    "user": {
                        "first_name": "",                                                                                                                                                                                                                                                             
                        "id": 1,
                        "last_name": "",                                                                                                                                                                                                                                                              
                        "username": "admin"
                    }
                },                                                                                                                                                                                                                                                                                    
                "token": "Zvk7PYb0eKusQn6fx0jTgGuvp8Yo9m",                                                                                                                                                                                                                                            
                "type": "o_auth2_access_token",                                                                                                                                                                                                                                                       
                "url": "/api/v2/tokens/461/",
                "user": 1
            }
        },
        "changed": true,
        "failed": false,
        "id": 461,
        "name": "unknown",                                                                                                                                                                                                                                                                            
        "token": "Zvk7PYb0eKusQn6fx0jTgGuvp8Yo9m",
        "warnings": [                                                                                                                                                                                                                                                                                 
            "You are running collection version devel but connecting to tower version 11.2.0"
        ]
    }                                                                                                                                                                                                                                                                                                 
}
```

and if it has a name we use the name, im having trouble locally deleting stuff, im not sure the delete actually works

awx_collection/tests/integration/targets/tower_token/tasks/main.yml
squidboylan: these tests still need to use `new_token` though right?

squidboylan: oh nvm, looks like `tower_token` works, i didnt realize modules could just register variables but thinking about it now it makes sense

awx_collection/plugins/module_utils/tower_api.py
squidboylan: i think that's suppsed to be `elif not 'token' in self.params.get('tower_oauthtoken'):` which gives us the old string behavior, my testing shows that the old string behavior is currently broken, im pretty sure that's why

john-westcott-iv: This has been updated, take a look and let me know what you think.

AlanCoding: 👍

awx_collection/plugins/modules/tower_token.py
john-westcott-iv: I'm a little torn on providing additional details about the return dict. The whole point of the module is to hide the login mechanics from someone. IMHO we should be trying to document "here is how you call this thing and pass it onto the next" vs. "here is how you call this thing, here are all of the internals of what you get back, now pass it onto the next". If there was other useful data within the dick maybe but I'm not sure the fields would provide any use. And people really should not have a need to "create" one of these dicts on their own to use for authentication (that would likely be more work than just cutting a token directly in Tower and using the string form). Do you see an advantage to documenting these files two fields?

As for the return 'token' value, I think that might just be an artifact from before getting the ansible_fact returned. Any objection to removing it from the response?

john-westcott-iv: Technically this module is the CRUD interface to the tokens endpoint and no other module includes a RETURN statement. But I think this one should because it’s setting the ansible_fact for token usage which deviates form the norm. In addition, the token string is only available though this fact once and can't be obtained through the API on subsequent calls.

Because of that, I pushed a change which modified the module description to add:
```
    - In addition, the module sets an Ansible fact which can be passed into other
      tower_* modules as the parameter tower_oauthtoken. See examples for usage.
    - Because of the sensitive nature of tokens, the created token value is only available once
      through the Ansible fact. (See RETURN for details)
```

Then In the return we say:
```
tower_token:
  type: dict
  description: An Ansible Fact variable representing a Tower token object which can be used for auth in subsequent modules. See examples for usage.
  contains:
    token:
      description: The token that was generated. This token can never be accessed again, make sure this value is noted before it is lost.
      type: str
  returned: on successful create
```

However, I’m not sure we want to dive into the data structure (i.e. the ID). Technically, like all other modules, the ID is in the return. But people really shouldn't care what else is in the dict and they shouldn't be trying to create one on their own. 

I also removed the token in the response (that was just an artifact from before the ansible_fact was returned). 

Let me know what people think about these changes.

awx_collection/plugins/module_utils/tower_api.py
john-westcott-iv: I pushed some updated code for this. Let me know if this resolves the issue.

squidboylan (approved):
LGTM

awx_collection/plugins/modules/tower_token.py
```diff
@@ -0,0 +1,168 @@
...
+'''
+
+RETURN = '''
+tower_token:
+  type: dict
+  description: An Ansible Fact variable representing a Tower token object which can be used for auth in subsequent modules. See examples for usage.
+  contains:
+    token:
+      description: The token that was generated. This token can never be accessed again, make sure this value is noted before it is lost.
+      type: str
```
AlanCoding, line 113: Again, I think documenting `id` (if that is what the key is) is a minimum requirement, because it is part of the function of the module `tower_token`, and a part of the interface between the module logic and the parameter structure passed into it.

awx_collection/plugins/modules/tower_token.py
```diff
@@ -0,0 +1,168 @@
...
+    scope:
+      description:
+        - Allowed scopes, further restricts user's permissions. Must be a simple space-separated string with allowed scopes ['read', 'write'].
+      required: False
+      type: str
+      default: 'write'
+      choices: ["read", "write"]
+    existing_token:
+      description: An existing token (for use with state absent)
```
AlanCoding: People will absolutely mis-construe this. People will read "token" as the token value, but that doesn't match the fact that it's a dict.

This single field is expected to be a mirror of all of the module params combined into a dict. It's fine to say that, and much better than not saying it.

What about idempotency? Is this module expected to be idempotent, and if not, what is expected of the user in order to write a playbook that doesn't leave around stale tokens?

awx_collection/plugins/modules/tower_token.py
```diff
@@ -0,0 +1,194 @@
...
+    description = module.params.get('description')
+    application = module.params.get('application')
+    scope = module.params.get('scope')
+    existing_token = module.params.get('existing_token')
+    existing_token_id = module.params.get('existing_token_id')
+    state = module.params.get('state')
+
+    if state == 'absent':
+        if not existing_token:
+            if not existing_token_id:
```
AlanCoding: it could make sense to mark these as mutually exclusive.

awx_collection/test/awx/test_token.py
```diff
@@ -0,0 +1,28 @@
...
+        'tower_host': None,
+        'tower_username': None,
+        'tower_password': None,
+        'validate_certs': None,
+        'tower_oauthtoken': None,
+        'tower_config_file': None,
+    }
+
+    result = run_module('tower_token', module_args, admin_user)
+    assert result, result.get('changed')
```
AlanCoding: I think is is backwards, `assert result` will always evaluate to true with a non-empty dict. You want to `assert result.get('changed'), result`, meaning that the result will be printed if it's not changed.

awx_collection/test/awx/test_token.py
```diff
@@ -0,0 +1,28 @@
...
+
+    tokens = OAuth2AccessToken.objects.filter(description='barfoo')
+    assert len(tokens) == 1, tokens[0].description == 'barfoo'
```
AlanCoding: maybe assert that the token's scope is read?

awx_collection/plugins/modules/tower_token.py
john-westcott-iv: Done, also added required_if so that one of them must be present if state == 'absent'

awx_collection/test/awx/test_token.py
john-westcott-iv: Thanks, fixed.

awx_collection/test/awx/test_token.py
john-westcott-iv: Thanks, resolved.

awx_collection/tests/integration/targets/tower_token/tasks/main.yml
```diff
@@ -0,0 +1,110 @@
...
+- assert:
+    that:
+      - results is failed
+      - '"The provided tower_oauthtoken type was not valid (list). Valid options are str or dict." == results.msg'
+
+- name: Try to delete a token with no existing_token or existing_token_id
+  tower_token:
+    state: absent
+  register: results
+  ignore_errors: True
```
AlanCoding: I think this is the last issue from CI

```
2020-06-08 10:53:56.769399 | static | linters run-test: commands[1] | yamllint -s .
2020-06-08 10:53:59.792596 | static | ./awx_collection/tests/integration/targets/tower_token/tasks/main.yml
2020-06-08 10:53:59.792784 | static |   34:18     warning  truthy value should be one of [false, true]  (truthy)
2020-06-08 10:53:59.792890 | static |   48:18     warning  truthy value should be one of [false, true]  (truthy)
```

john-westcott-iv: This one gets me all the time.

awx_collection/test/awx/test_token.py
```diff
@@ -0,0 +1,29 @@
...
+        'validate_certs': None,
+        'tower_oauthtoken': None,
+        'tower_config_file': None,
+    }
+
+    result = run_module('tower_token', module_args, admin_user)
+    assert result.get('changed'), result
+
+    tokens = OAuth2AccessToken.objects.filter(description='barfoo')
+    assert len(tokens) == 1, 'Rokens with description of barfoo != 0: {0}'.format(len(tokens))
```
AlanCoding: Rokens?

awx_collection/plugins/modules/tower_token.py
```diff
@@ -0,0 +1,201 @@
...
+
+  always:
+    - name: Delete our Token with the token we created
+      tower_token:
+        existing_token: "{{ tower_token }}"
+        state: absent
+      when: tower_token is defined
+
+- name: Delete a token by its id
+  tower_toekn:
```
AlanCoding: `tower_toekn` -> `tower_token`

awx_collection/plugins/modules/tower_token.py
AlanCoding: ok, resolved

this was the most important things out of the various discussions here

AlanCoding (approved):

PR #7118, opened 2020-05-21
Moving tower_oauthtoken into a doc fragement

john-westcott-iv:
##### SUMMARY

Move the tower_oauthtoken doc param out of the individual modules and into a doc fragment.

##### ISSUE TYPE

 - Docs Pull Request

##### COMPONENT NAME

 - Collection

##### AWX VERSION

```
awx: 11.2.0
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx_collection/plugins/modules/tower_send.py
+++ b/awx_collection/plugins/modules/tower_send.py
@@ -60,7 +60,7 @@ requirements:
   - six.moves.StringIO
   - sys
 
-extends_documentation_fragment: awx.awx.auth
+extends_documentation_fragment: awx.awx.auth_legacy
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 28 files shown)


beeankha (approved):

awx_collection/plugins/modules/tower_send.py
```diff
@@ -60,7 +60,7 @@
   - six.moves.StringIO
   - sys
 
-extends_documentation_fragment: awx.awx.auth
+extends_documentation_fragment: awx.awx.auth_legacy
```
AlanCoding, line 63: so the idea is that this won't match it?

https://github.com/ansible/awx/blob/c61e875bf83b60587b1ced02ca3229a36794bda3/awx_collection/tools/roles/template_galaxy/tasks/main.yml#L22

john-westcott-iv: No, we should change that too, I dropped the $ so that it will match.

AlanCoding (approved):

Issue #4680, opened 2019-09-07
Raise error if a group and host is created with the same name


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Create an inventory, inside that inventory create a host with name "foo". Then create a group with name "foo".

This breaks playbooks.

We should prevent the creation of the group with name "foo" because a host by that name exists, and return a 400 error message with details in this case.

Related https://github.com/ansible/awx/issues/3384

kdelee:
verified, tried it a few ways:

- made group first, tried to add host to inventory or in the group itself, both denied saying a group already has the name
- made the host first, tried to add the group and got an error saying a host already existed.

closing as verified, thanks!

PR #7119, opened 2020-05-21
added change so that groups and host names cannot share the same name

thedoubl3j:
##### SUMMARY

Change added so that a host name foo cannot be created in the same inventory that has a group named foo and vice versa. related #4680 

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API
##### AWX VERSION

```
awx: 11.2.0

```

##### ADDITIONAL INFORMATION

Before: you were allowed to create a host and group with the same name.
After: Will error and return a 400

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -1719,6 +1719,8 @@ class HostSerializer(BaseSerializerWithVariables):
         name = force_text(value or '')
         # Validate here only, update in main validate method.
         host, port = self._get_host_port_from_name(name)
+        if Group.objects.filter(name=value).exists():
+            raise serializers.ValidationError(_('Invalid group name. Name already exists as a Group.'))
         return value
 
     def validate_inventory(self, value):
@@ -1808,6 +1810,8 @@ class GroupSerializer(BaseSerializerWithVariables):
     def validate_name(self, value):
         if value in ('all', '_meta'):
             raise serializers.ValidationError(_('Invalid group name.'))
+        elif Host.objects.filter(name=value).exists():
+            raise serializers.ValidationError(_('Invalid group name. Name already exists as a Host.'))
         return value
 
     def validate_inventory(self, value):
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -1719,8 +1719,6 @@ class HostSerializer(BaseSerializerWithVariables):
         name = force_text(value or '')
         # Validate here only, update in main validate method.
         host, port = self._get_host_port_from_name(name)
-        if Group.objects.filter(name=value).exists():
-            raise serializers.ValidationError(_('Invalid group name. Name already exists as a Group.'))
         return value
 
     def validate_inventory(self, value):
@@ -1733,6 +1731,7 @@ class HostSerializer(BaseSerializerWithVariables):
 
     def validate(self, attrs):
         name = force_text(attrs.get('name', self.instance and self.instance.name or ''))
+        inventory = attrs.get('inventory', self.instance and self.instance.inventory or '')
         host, port = self._get_host_port_from_name(name)
 
         if port:
@@ -1741,7 +1740,9 @@ class HostSerializer(BaseSerializerWithVariables):
             vars_dict = parse_yaml_or_json(variables)
             vars_dict['ansible_ssh_port'] = port
             attrs['variables'] = json.dumps(vars_dict)
-
+        if Group.objects.filter(name=name, inventory=inventory).exists():
+            raise serializers.ValidationError(_('A Group with that name already exists.'))
+            
         return super(HostSerializer, self).validate(attrs)
 
     def to_representation(self, obj):
@@ -1807,11 +1808,16 @@ class GroupSerializer(BaseSerializerWithVariables):
             res['inventory'] = self.reverse('api:inventory_detail', kwargs={'pk': obj.inventory.pk})
         return res
 
+    def validate(self, attrs):
+        name = force_text(attrs.get('name', self.instance and self.instance.name or ''))
+        inventory = attrs.get('inventory', self.instance and self.instance.inventory or '')
+        if Host.objects.filter(name=name, inventory=inventory).exists():
+            raise serializers.ValidationError(_('A Host with that name already exists.'))
+        return super(GroupSerializer, self).validate(attrs)
+
     def validate_name(self, value):
         if value in ('all', '_meta'):
             raise serializers.ValidationError(_('Invalid group name.'))
-        elif Host.objects.filter(name=value).exists():
-            raise serializers.ValidationError(_('Invalid group name. Name already exists as a Host.'))
         return value
 
     def validate_inventory(self, value):
--- a/awx/main/tests/functional/api/test_inventory.py
+++ b/awx/main/tests/functional/api/test_inventory.py
@@ -59,6 +59,21 @@ def test_inventory_source_unique_together_with_inv(inventory_factory):
     is2 = InventorySource(name='foo', source='file', inventory=inv2)
     is2.validate_unique()
 
+@pytest.mark.django_db
+def test_inventory_host_name_unique(scm_inventory, post, admin_user):
+    inv_src = scm_inventory.inventory_sources.first()
+    group = inv_src.groups.create(name='barfoo', inventory=scm_inventory)
+    group.save()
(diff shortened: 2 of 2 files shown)


thedoubl3j:
big thanks to @rebeccahhh for all of the help

awx/api/serializers.py
```diff
@@ -1719,6 +1719,8 @@ def validate_name(self, value):
         name = force_text(value or '')
         # Validate here only, update in main validate method.
         host, port = self._get_host_port_from_name(name)
+        if Group.objects.filter(name=value).exists():
+            raise serializers.ValidationError(_('Invalid group name. Name already exists as a Group.'))
```
ryanpetrello: Maybe just "A Group with that name already exists." ?

You don't need to explicitly say that the *host* name is invalid, because the serializer will actually return the errant field like this in the API response (this is how the UI knows which field to draw the error on):

image

awx/api/serializers.py
```diff
@@ -1808,6 +1810,8 @@ def get_related(self, obj):
     def validate_name(self, value):
         if value in ('all', '_meta'):
             raise serializers.ValidationError(_('Invalid group name.'))
+        elif Host.objects.filter(name=value).exists():
+            raise serializers.ValidationError(_('Invalid group name. Name already exists as a Host.'))
```
ryanpetrello: Similarly, "A Host with that name already exists."

awx/main/tests/functional/api/test_inventory.py
```diff
@@ -60,6 +60,24 @@ def test_inventory_source_unique_together_with_inv(inventory_factory):
     is2.validate_unique()
 
 
+@pytest.mark.django_db
+def test_inventory_host_name_unique(scm_inventory, post, admin_user):
+    inv_src = scm_inventory.inventory_sources.first()
+    group = inv_src.groups.create(name='barfoo', inventory=scm_inventory)
+    group.save()
+    host1 = inv_src.hosts.create(name='barfoo', inventory=scm_inventory)
+    post(reverse('api:inventory_hosts_list', kwargs={'pk': host1.id}), admin_user, expect=400)
```
ryanpetrello: You're getting a 400 error here because this `POST` request isn't actually creating the host (because you already created it on the line right above, `host1 = ...`).  Instead, it's making a request like this (which is invalid):

```http
HTTP POST /api/v2/hosts/N/
{}
```

What you *want* is to make _this_ request and verify that the host *doesn't* get created.

```http
HTTP POST /api/v2/hosts/
{
    "name": "barfoo",
    "inventory_id": 1
}
```

You want to pass the parameters for creation as part of the POST request, and get make sure you get back a 400.  Maybe something like this:

```python
group.save()
resp = post(
    reverse('api:inventory_hosts_list'),
    {
        'name': 'barfoo',
        'inventory_id': scm_inventory.id,
    }
    admin_user,
    expect=400
)
```

Additionally, as you can see in your original test, "HTTP 400" can mean a lot of things - it might not mean you hit the code path you care about.  To make sure you're actually hitting the code path you expect, you might want to just add an additional assertion:

```python
assert "A Group with that name already exists." in resp.data.get('name', [])
```

awx/main/tests/functional/api/test_inventory.py
```diff
@@ -60,6 +60,24 @@ def test_inventory_source_unique_together_with_inv(inventory_factory):
...
+
+
+@pytest.mark.django_db   
+def test_inventory_group_name_unique(scm_inventory, post, admin_user):
+    inv_src = scm_inventory.inventory_sources.first()
+    host = inv_src.hosts.create(name='barfoo', inventory=scm_inventory)
+    host.save()
+    group = inv_src.groups.create(name='barfoo', inventory=scm_inventory)
+    post(reverse('api:inventory_groups_list', kwargs={'pk': group.id}), admin_user, expect=400)
```
ryanpetrello: Same comments as the above test.

awx/api/serializers.py
AlanCoding: Bigger issue - this need to be filtered to the inventory.

`Group.objects.filter(name=value)` -> `Group.objects.filter(name=value, inventory=self.instance.inventory)`

If there's not a `self.instance` that could be a problem, because then it changes from being a field simple field value validation to mutual field value validation. User gave an inventory value and a name value. Does the _combination_ of those two inputs create a situation which should be rejected?

To handle those cases, logic may have to be moved from `validate_name` to `validate`, where all attributes are simultaneously available.

ryanpetrello: Good point 👍

ryanpetrello: Also, what are the implications for smart inventory?

*Is* this something we can even enforce reliably at that level (it sounds like it would be difficult to do so)?

wenottingham: ... how would smart inventory get here? You can't post a host/group to it.

ryanpetrello: Okay, that's true - I was thinking if you changed the smart inventory filter such that it introduced a _new_ host with a collision, but smart inventories don't contain groups, so ignore me.

thedoubl3j: @ryanpetrello @AlanCoding pushed a change, let me know your thoughts when you have a moment.

awx/api/serializers.py
```diff
@@ -1739,7 +1739,9 @@ def validate(self, attrs):
             vars_dict = parse_yaml_or_json(variables)
             vars_dict['ansible_ssh_port'] = port
             attrs['variables'] = json.dumps(vars_dict)
-
+        if Group.objects.filter(name=name, inventory=attrs.get('inventory')).exists():
```
AlanCoding: I think you may need to check `self.instance.inventory` for the case that someone changes the name of an existing group. Similar to how below you did `attrs.get('name', self.instance and self.instance.name or '')`, a similar thing for "inventory" as opposed to "name"

awx/api/serializers.py
```diff
@@ -1731,15 +1731,18 @@ def validate_variables(self, value):
 
     def validate(self, attrs):
         name = force_text(attrs.get('name', self.instance and self.instance.name or ''))
+        inventory = attrs.get('inventory', self.instance and self.instance.inventory or '')
         host, port = self._get_host_port_from_name(name)
-
+
```
rebeccahhh: this is a tab from your editor, it's gonna need to be removed :)

rebeccahhh (commented):
The only other thing is we're gonna want to squash these commits a bit and make the messages more unified/useful for the commit history. Typically I say "no squashing" but in this case we should. 1 (maybe 2) commit should be expressive enough to communicate the change.

awx/main/tests/functional/api/test_inventory.py
```diff
@@ -60,6 +60,44 @@ def test_inventory_source_unique_together_with_inv(inventory_factory):
     is2.validate_unique()
 
 
+@pytest.mark.django_db
+def test_inventory_host_name_unique(scm_inventory, post, admin_user):
+    inv_src = scm_inventory.inventory_sources.first()
+    group = inv_src.groups.create(name='barfoo', inventory=scm_inventory)
+    group.save()
```
AlanCoding: there's no need to call the `.save()` after a `.create`. You should be able to delete this line

awx/main/tests/functional/api/test_inventory.py
```diff
@@ -60,6 +60,44 @@ def test_inventory_source_unique_together_with_inv(inventory_factory):
...
+
+    assert resp.status_code == 400
+    assert "A Group with that name already exists." in json.dumps(resp.data)
+
+
+@pytest.mark.django_db   
+def test_inventory_group_name_unique(scm_inventory, post, admin_user):
+    inv_src = scm_inventory.inventory_sources.first()
+    host = inv_src.hosts.create(name='barfoo', inventory=scm_inventory)
+    host.save()
```
AlanCoding: same for deleting this line

AlanCoding (approved):
I concur with what @rebeccahhh said about tabs & commits

The content here looks good to me. Once you're done with those administrative things I'm fine to see this merged.

thedoubl3j:
@AlanCoding removed @rebeccahhh  done. have 2 commits, left the first one hanging as it was the intial idea and then changed the second to document the moving of the logic and updates to subsequent tests.

rebeccahhh (approved):

ryanpetrello (approved):
👍

PR #7120, opened 2020-05-22
Add option to follow symlinks when scanning for playbooks

fherbert:
##### SUMMARY
Add a setting in settings -> jobs -> 'Follow Symlinks', which enabled or disables the option to find playbooks that are in a symlinked directory.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### AWX VERSION
```
awx: 11.2.0
```

##### ADDITIONAL INFORMATION
We have playbooks that are in symlinked folders which don't get discovered by awx.
This change allows those playbooks to be used.

By default, this feature is disabled so user experience will not change until setting is modified.

Once the setting is changed you won't see any difference in the list of playbooks until the project is synchronised.

### DIFF
--- a/awx/main/conf.py
+++ b/awx/main/conf.py
@@ -424,6 +424,16 @@ register(
 )
 
 register(
+    'AWX_SHOW_PLAYBOOK_LINKS',
+    field_class=fields.BooleanField,
+    default=False,
+    label=_('Follow symlinks'),
+    help_text=_('Follow symbolic links when scanning for playbooks'),
+    category=_('Jobs'),
+    category_slug='jobs',
+)
+
+register(
     'PRIMARY_GALAXY_URL',
     field_class=fields.URLField,
     required=False,
--- a/awx/main/conf.py
+++ b/awx/main/conf.py
@@ -428,8 +428,9 @@ register(
     field_class=fields.BooleanField,
     default=False,
     label=_('Follow symlinks'),
-    help_text=_('Follow symbolic links when scanning for playbooks. Be aware that setting this to True '
-        'can lead to infinite recursion if a link points to a parent directory of itself.'
+    help_text=_(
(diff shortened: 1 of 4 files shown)


awx/main/conf.py
```diff
@@ -423,6 +423,16 @@ def _load_default_license_from_file():
     category_slug='jobs',
 )
 
+register(
+    'AWX_SHOW_PLAYBOOK_LINKS',
+    field_class=fields.BooleanField,
+    default=False,
+    label=_('Follow symlinks'),
+    help_text=_('Follow symbolic links when scanning for playbooks'),
```
ryanpetrello: Can we add this explicit text to the end of the description?

> Be aware that setting this to True can lead to infinite recursion if a link points to a parent directory of itself.

ryanpetrello (approved):

AlanCoding:
This seems to cause the error:

```
awx_1        | 2020-05-27 12:25:29,890 ERROR    awx.main.tasks project_update 1937 (running) Post run hook errored.
awx_1        | Traceback (most recent call last):
awx_1        |   File "/awx_devel/awx/main/tasks.py", line 1501, in run
awx_1        |     self.post_run_hook(self.instance, status)
awx_1        |   File "/awx_devel/awx/main/tasks.py", line 2389, in post_run_hook
awx_1        |     p.playbook_files = p.playbooks
awx_1        |   File "/awx_devel/awx/main/models/projects.py", line 202, in playbooks
awx_1        |     for dirpath, dirnames, filenames in os.walk(smart_str(project_path, followlinks=settings.AWX_SHOW_PLAYBOOK_LINKS)):
awx_1        | TypeError: smart_text() got an unexpected keyword argument 'followlinks'
```

and I don't see any linked follow up, so I'll try to loop back around to that sometime.

ryanpetrello:
@AlanCoding https://github.com/ansible/awx/pull/7167

Issue #3461, opened 2019-03-21
Fallback instance_id values for inventory imports


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
 - API

##### SUMMARY
Right now, an inventory source type can only have one instance_id value.

This issue is to allow a list of values for instance_id, where it will find the identifier by looking at the first one, and the falling back to the 2nd, 3rd, etc.

##### ADDITIONAL INFORMATION

The reason is that the inventory plugins have changed the keys for hostvars, including the ones we use for instance_ids.

If we ever allow the YAML inventory config to be customizable (and we should), then we need to anticipate the scenario where we get the instance ID by the new name but not the old name.

Concretely...

| Source   | old key | new key     |
|----------|---------|-------------|
| ec2      | ec2_id  | instance_id |
| gce      | gce_id  | id          |
| azure_rm | id      | id          |

In https://github.com/ansible/awx/pull/3266, we have a solution to create the old key via the `compose` options from the old key. That will work for now, but again, we want to allow user customizations. We don't want to constrain them. Solving this issue will allows us to allow the users to remove that variable translation.

So, not urgent, but definitely what we want long-term. Ping @kdelee @wenottingham

AlanCoding:
Prototype branch: https://github.com/ansible/awx/compare/devel...AlanCoding:instance_id_fallbacks?expand=1

jladdjr:
Alas, I should not say that PR X fixes Issue Y, otherwise Zuul starts taking over.

kdelee:
@AlanCoding I'm reading this issue and trying to figure out if @jladdjr 's patch actually "closes" this issue or just implements it for vmware.

Seems like this issue is more general than that. Would you concur that Jim's patch covers this issue?

If so I would close as verified by regression testing w/ vmware as it is not breaking anything for us and provides flexiblity for either way the variable is specified, which AFAIK depends on version of inventory plugin you might be using.

AlanCoding:
This issue is a little too old.

We did use the `compose` dictionary to re-create the old values from the new values.

VMWare is only the exception because the value we're looking for is nested in a sub-dictionary, where we decided to axe `config.instanceuuid`. Also, since the old/new here was a matter of upper/lower case you shouldn't really have both co-existing in hostvars at the same time.

I would just close this issue, I think we had others which were more specifically tracking the error that the patch fixed. I don't expect we need anything else.

kdelee:
works for me, thanks

PR #7121, opened 2020-05-22
Fix vmware inventory sync to not delete/re-add hosts

jainnikhil30:
##### SUMMARY
This fixes the vmware inventory sync issue, where it deletes/re-adds the host even if nothing has changed for the host.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### ADDITIONAL INFORMATION

Checking the API from vmware the property was always instanceUuid

Vmware Esxi 6.0
https://code.vmware.com/apis/42/vsphere/doc/vim.vm.ConfigInfo.html

Vmware Esxi 5.5
https://code.vmware.com/apis/197/vsphere/doc/vim.vm.ConfigInfo.html

Earlier script used to lowercase every hostvar but the new inventory plugin doesn't do that and hence the instanceUuid doesnt match with instanceuuid.

Thanks to @Akasurde for lending his ESXI server for testing

### DIFF
--- a/awx/settings/defaults.py
+++ b/awx/settings/defaults.py
@@ -717,7 +717,7 @@ VMWARE_ENABLED_VAR = 'guest.gueststate'
 VMWARE_ENABLED_VALUE = 'running'
 
 # Inventory variable name containing the unique instance ID.
-VMWARE_INSTANCE_ID_VAR = 'config.instanceuuid'
+VMWARE_INSTANCE_ID_VAR = 'config.instanceUuid'
 
 # Filter for allowed group and host names when importing inventory
 # from VMware.


ryanpetrello (approved):

ryanpetrello:
cc @jladdjr @AlanCoding this looks pretty straightforward, and it sounds like @jainnikhil30 has spent some time testing it, so I'm gonna merge it.

jainnikhil30:
Just for information. Here is the inventory sync output without this change:

```
   21.903 INFO     Loaded 2 groups, 1 hosts
32
   21.917 WARNING  Host "vmware-vr-appliance-7.6.0.317-13027280_ovf10" has no "config.instanceuuid" variable
33
   21.959 DEBUG    Deleted host "vmware-vr-appliance-7.6.0.317-13027280_ovf10"
```

After the changes from PR:
```
      38.277 INFO     Loaded 2 groups, 1 hosts
32
   38.332 DEBUG    Inventory variables unmodified
33
   38.335 DEBUG    Group "guests" variables unmodified
34
   38.335 DEBUG    Group "sles11_64Guest" variables unmodified
35
   38.359 DEBUG    Host "VMware-vR-Appliance-7.6.0.317-13027280_OVF10_42290749-f59d-7541-3a6d-897a9721cfa7" variables updated
36
   38.378 DEBUG    Host "VMware-vR-Appliance-7.6.0.317-13027280_OVF10_42290749-f59d-7541-3a6d-897a9721cfa7" already in group "guests"
```

It doesn't deletes the host now.

ryanpetrello:
@jainnikhil30,

Does this work with Ansible 2.8?  If not, what's the fallout?

cc @AlanCoding @jladdjr

jainnikhil30:
@ryanpetrello 

Oh yeah, this will make things go bad with Ansible 2.8 because then Tower would go back to script, which is doing the lowercase on all host vars. This change will cause the same behavior on script, like every host will be deleted and re-added. I wonder if we can do a ansible version dependent change ?

AlanCoding:
Material from https://github.com/ansible/awx/issues/3461 might could possibly help resolve that?

jladdjr:
Chatted with @AlanCoding about this. It does sound like instead of changing the `VMWARE_INSTANCE_ID_VAR` to the new value, we would want to add the fallback code that alan mentioned so that we can support both old / new hostvars used to assign instance ids.

jainnikhil30:
@jladdjr @AlanCoding 

yeah makes sense. Thanks for taking a look into this.

jladdjr:
Fix that includes updates key for building vmware id as well as fallback mechanism to provide backwards compatibility:
https://github.com/ansible/awx/pull/7321/files

PR #7123, opened 2020-05-22
Adding templating of README.md file for Tower vs AWX collections

john-westcott-iv:
##### SUMMARY

Using the existing tools to dynamically generate a README.md from a template for the ansible.tower vs awx.awx collections.

##### ISSUE TYPE

 - Docs Pull Request

##### COMPONENT NAME

 - Collection

##### AWX VERSION

```
awx: 11.2.0
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx_collection/tools/roles/template_galaxy/tasks/main.yml
+++ b/awx_collection/tools/roles/template_galaxy/tasks/main.yml
@@ -38,3 +38,9 @@
     src: "{{ collection_path }}/tools/roles/template_galaxy/templates/galaxy.yml.j2"
     dest: "{{ collection_path }}/galaxy.yml"
     force: true
+
+- name: Tempalte the README.md file
+  template:
+    src: "{{ collection_path }}/tools/roles/template_galaxy/templates/README.md.j2"
+    dest: "{{ collection_path }}/README.md"
+    force: true
--- a/awx_collection/tools/roles/template_galaxy/templates/README.md.j2
+++ b/awx_collection/tools/roles/template_galaxy/templates/README.md.j2
@@ -1,4 +1,4 @@
-# AWX Ansible Collection
+# {% if collection_package | lower() == 'awx' %}AWX{% else %}Tower{% endif %} Ansible Collection
 
 [comment]: # (*******************************************************)
 [comment]: # (*                                                     *)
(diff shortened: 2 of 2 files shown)


awx_collection/README.md
```diff
@@ -1,7 +1,18 @@
...
+[comment]: # (*                                                     *)
+[comment]: # (*  This file is manageed by Ansible and not to be     *)
+[comment]: # (*  edited directly! Instead modify:                   *)
+[comment]: # (*  tools/roles/template_galaxy/templates/README.md.j2 *)
+[comment]: # (*                                                     *)
+[comment]: # (*  Changed to the base README.md file are lost upon   *)
+[comment]: # (*  the build of the collection                        *)
+[comment]: # (*******************************************************)
+
+This Ansible collection allows for easy interaction with an AWX server via Ansible playbooks.
```
AlanCoding, line 15: my only real comment here - could you instead just add this to the main `.gitignore`?

Otherwise, commits will be introducing the same diff to 2 files, and that's going to introduce synchronicity issues, which starts going wrong the second that another person contributes.

john-westcott-iv: We're kind of in a catch 22 because we want the README file in the repo for AWX but we want a different README to be generated dynamically for ansible.tower. I don't think .gitignore will work here because that only allows you to ignore files that aren't in the index. Since README.md is already there it will still show up in a git status on modification and will still be eligible for commits.

AlanCoding: if the problem is that the file is in the index, then can't you just delete it and then add it to the `.gitignore`?

> we want the README file in the repo for AWX

Why? Maybe it's nice to have, but I don't see that as necessary. If general direction is needed, it could be put in the [overall project README.md](https://github.com/ansible/awx/blob/devel/README.md)

The main service this README offers is the Galaxy page, and we already kind of preference that formatting over Github.

AlanCoding (approved):
I have my preference about .gitignore, but I don't want to make a blocker for merging this.

AlanCoding (changes requested):
I put up an alternative pull request which would make the management of this stuff more consistent

AlanCoding:
This has actually gone out-of-date with the current README.md file.

I would also really strongly push toward the replacement PR I put up, which I believe is mergeable now.

Issue #7052, opened 2020-05-17
awx 11.2 : unable to sync vmware vcenter inventory


lgrullon-gilead:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Upon upgrading from 11 to 11.2 I no longer able to sync my vcenter dynamic inventory objects within the GUI.

##### ENVIRONMENT
* AWX version: 11.2.0
* AWX install method: docker
* Ansible version:  ansible-2.9.7-1.el8.noarch
* Operating System: RHEL 8
* Web Browser: Chrome, Firefox, IE

##### STEPS TO REPRODUCE

When running the job for vmware inventory sync the job is failling with "ERROR! No inventory was parsed, please check your configuration and options."

##### EXPECTED RESULTS

The expected outcome should reflect successful and the inventory should get updated from vcenter objects.

##### ACTUAL RESULTS

 2.030 INFO     Updating inventory 4: vCenter-DI
    2.452 INFO     Reading Ansible inventory source: /tmp/awx_1072_i0i4h5_1/vmware_vm_inventory.yml
    2.454 INFO     Using VIRTUAL_ENV: /var/lib/awx/venv/ansible
    2.454 INFO     Using PATH: /var/lib/awx/venv/ansible/bin:/var/lib/awx/venv/awx/bin:/usr/pgsql-10/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
    2.454 INFO     Using PYTHONPATH: /var/lib/awx/venv/ansible/lib/python3.6/site-packages:
Traceback (most recent call last):
  File "/var/lib/awx/venv/awx/bin/awx-manage", line 8, in <module>
    sys.exit(manage())
  File "/var/lib/awx/venv/awx/lib/python3.6/site-packages/awx/__init__.py", line 152, in manage
    execute_from_command_line(sys.argv)
  File "/var/lib/awx/venv/awx/lib/python3.6/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/var/lib/awx/venv/awx/lib/python3.6/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/var/lib/awx/venv/awx/lib/python3.6/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
[...]

PR #7124, opened 2020-05-22
vmware_vm_inventory: Replace guest.guestId with config.guestId

Akasurde:
##### SUMMARY

guest.guestId is optional value and may not be populated everytime.
Use config.guestId instead which is mandatory.
Removed 'tag' property from properties since this property is never
populated.

Partially fixes https://github.com/ansible/awx/issues/7052

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
awx/main/models/inventory.py

### DIFF
--- a/awx/main/models/inventory.py
+++ b/awx/main/models/inventory.py
@@ -2275,7 +2275,7 @@ class vmware(PluginFileInjector):
             "customValue",  # optional
             "datastore",
             "effectiveRole",
-            "guestHeartbeatStatus",  # optonal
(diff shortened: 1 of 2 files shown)


Akasurde:
cc @AlanCoding

AlanCoding (approved):
thanks! All the checks I needed to run look alright, and I'm good to merge.

PR #7125, opened 2020-05-22
Adds basic credential plugin support to relevant fields in the static credential forms

mabashian:
##### SUMMARY
Background on credential plugins can be found here: https://docs.ansible.com/ansible-tower/3.5.0/html/administration/credential_plugins.html.

There are a handful of new components worth talking about:

1. CredentialPluginField.jsx: Since almost every field in the credential Type Details subform needed to have this button on the right hand side to initiate the cred plugin flow:

Screen Shot 2020-05-22 at 9 37 28 AM

I went ahead and created this component which handles everything except the input itself.  As far as I know there are three types of inputs that can be passed in as children: Text, Textarea and Password.

2. CredentialPluginSelected.jsx: This is the component that gets rendered (instead of the form field) when the user has configured a field to be retrieved from an external store.  This is what it looks like:

Screen Shot 2020-05-22 at 9 41 13 AM

Clicking (x) on the chip will remove the credential and bring the field back to an empty state.  Clicking on the key icon will let the user edit the credential/additional fields in the wizard.

3. CredentialPluginPrompt.jsx: This is the wizard that gets shown when the user clicks on the key icon on any available field.  There are always two steps.  First step is to select a credential.  The fields in the second step are determined by the type of credential selected in the first step.

4. PasswordInput.jsx: I decided to split PasswordField.jsx up into PasswordField and PasswordInput.  My reasoning for this was that I could reuse PasswordInput inside of CredentialPluginField.  The functionality should remain the same for existing password fields (User form).

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
This PR is currently without tests.  I will be adding _some_ test coverage shortly (especially for the plugin-related components).

The next step in this plan is to get rid of the static forms in favor of a form generator type solution so I'm not inclined to write extensive test coverage on components that I know will go away in the coming weeks.

There are a couple of things I left unimplemented:

1) The Test button - https://github.com/ansible/awx/issues/7126
2) Updating credential details to show fields that are configured to pull from an external source - https://github.com/ansible/awx/issues/7127

### DIFF
--- a/awx/ui_next/src/components/FormField/PasswordInput.test.jsx
+++ b/awx/ui_next/src/components/FormField/PasswordInput.test.jsx
@@ -1,0 +1,44 @@
+import React from 'react';
+import { Formik } from 'formik';
+import { mountWithContexts } from '../../../testUtils/enzymeHelpers';
+import { sleep } from '../../../testUtils/testUtils';
+import PasswordInput from './PasswordInput';
+
+describe('PasswordInput', () => {
+  test('renders the expected content', () => {
+    const wrapper = mountWithContexts(
+      <Formik
+        initialValues={{
+          password: '',
+        }}
+      >
+        {() => (
+          <PasswordInput id="test-password" name="password" label="Password" />
+        )}
+      </Formik>
+    );
+    expect(wrapper).toHaveLength(1);
+  });
+
+  test('properly responds to show/hide toggles', async () => {
+    const wrapper = mountWithContexts(
+      <Formik
+        initialValues={{
+          password: '',
+        }}
+      >
+        {() => (
+          <PasswordInput id="test-password" name="password" label="Password" />
+        )}
+      </Formik>
+    );
+    expect(wrapper.find('input').prop('type')).toBe('password');
+    expect(wrapper.find('EyeSlashIcon').length).toBe(1);
+    expect(wrapper.find('EyeIcon').length).toBe(0);
+    wrapper.find('button').simulate('click');
+    await sleep(1);
+    expect(wrapper.find('input').prop('type')).toBe('text');
+    expect(wrapper.find('EyeSlashIcon').length).toBe(0);
+    expect(wrapper.find('EyeIcon').length).toBe(1);
+  });
+});
--- a/awx/ui_next/src/components/FormField/PasswordInput.test.jsx
+++ b/awx/ui_next/src/components/FormField/PasswordInput.test.jsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import { Formik } from 'formik';
 import { mountWithContexts } from '../../../testUtils/enzymeHelpers';
-import { sleep } from '../../../testUtils/testUtils';
 import PasswordInput from './PasswordInput';
 
 describe('PasswordInput', () => {
@@ -36,7 +35,6 @@ describe('PasswordInput', () => {
     expect(wrapper.find('EyeSlashIcon').length).toBe(1);
     expect(wrapper.find('EyeIcon').length).toBe(0);
     wrapper.find('button').simulate('click');
-    await sleep(1);
     expect(wrapper.find('input').prop('type')).toBe('text');
     expect(wrapper.find('EyeSlashIcon').length).toBe(0);
     expect(wrapper.find('EyeIcon').length).toBe(1);
--- a/awx/ui_next/src/screens/Credential/CredentialEdit/CredentialEdit.jsx
+++ b/awx/ui_next/src/screens/Credential/CredentialEdit/CredentialEdit.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useCallback, useState, useEffect } from 'react';
 import { useHistory } from 'react-router-dom';
 import { object } from 'prop-types';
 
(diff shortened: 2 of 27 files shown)


mabashian:
Just pushed up a few tests for plugin-related components.  I'm planning on another commit to add test coverage for the plugin prompt component.

mabashian:
After talking with @trahman73 I've changed the UX of the selected plugin component just a bit:

Screen Shot 2020-05-26 at 3 11 44 PM

awx/ui_next/src/components/FormField/PasswordInput.test.jsx
```diff
@@ -0,0 +1,44 @@
...
+        {() => (
+          <PasswordInput id="test-password" name="password" label="Password" />
+        )}
+      </Formik>
+    );
+    expect(wrapper.find('input').prop('type')).toBe('password');
+    expect(wrapper.find('EyeSlashIcon').length).toBe(1);
+    expect(wrapper.find('EyeIcon').length).toBe(0);
+    wrapper.find('button').simulate('click');
+    await sleep(1);
```
jakemcdermott: what is this waiting for?

awx/ui_next/src/screens/Credential/shared/CredentialSubForms/GoogleComputeEngineSubForm.jsx
```diff
@@ -91,30 +96,38 @@ const GoogleComputeEngineSubForm = ({ i18n }) => {
...
           id="credential-sshKeyData"
           label={i18n._(t`RSA private key`)}
           name="inputs.ssh_key_data"
           type="textarea"
-          rows={6}
           validate={required(null, i18n)}
           isRequired
-        />
+        >
+          <TextArea
```
jakemcdermott, line 125: Should we use the file picker for ssh key and certificate fields? Given all the customization we'll end up doing for credential plugins I think they'll be simpler.

I'd also argue that they're safer by default: Unless we _also_ add drag-and-drop functionality, by using textareas we are inviting users to copy and paste secrets from their clipboard which is something they shouldn't do.

jakemcdermott (commented):
I think this looks great. I like the additional helper text and that we're using different icons for the secret lookup. Should we use file pickers for the ssh key and cert fields?

awx/ui_next/src/screens/Credential/CredentialEdit/CredentialEdit.jsx
```diff
@@ -12,45 +16,100 @@ function CredentialEdit({ credential, me }) {
...
+            data: { results: loadedInputSources },
+          },
+        ] = await Promise.all([
+          CredentialTypesAPI.read({
+            or__namespace: ['gce', 'scm', 'ssh'],
+          }),
+          CredentialsAPI.readInputSources(credential.id, { page_size: 200 }),
+        ]);
         setCredentialTypes(loadedCredentialTypes);
+        const inputSourcesMap = {};
```
jlmitch5: I think you could use an actual map here
const inputSourcesMap = loadedInputSources.map(inputSource => ({inputSource.input_field_name: inputSource}))

awx/ui_next/src/screens/Credential/CredentialEdit/CredentialEdit.jsx
```diff
@@ -12,45 +16,100 @@ function CredentialEdit({ credential, me }) {
...
+      await Promise.all([
+        CredentialsAPI.update(credential.id, {
+          user: (me && me.id) || null,
+          organization: (organization && organization.id) || null,
+          inputs: inputs || {},
+          ...remainingValues,
+        }),
+        ...destroyInputSources(pluginInputs),
+      ]);
+      await Promise.all(createAndUpdateInputSources(pluginInputs));
```
jlmitch5: i think this all looks good?  we don't have to put this in one of our useRequest-like hooks to fix the "calling stuff on unmounted component" issues do we?

awx/ui_next/src/screens/Credential/shared/CredentialSubForms/GoogleComputeEngineSubForm.jsx
jlmitch5: talked about this...we will

awx/ui_next/src/screens/Credential/shared/CredentialSubForms/GoogleComputeEngineSubForm.jsx
```diff
@@ -91,30 +96,38 @@ const GoogleComputeEngineSubForm = ({ i18n }) => {
           }}
         />
       </FormGroup>
-      <FormField
+      <CredentialPluginField
```
jlmitch5, line 99: I like how this basically just wraps the field...very easy to use!!

awx/ui_next/src/screens/Credential/shared/CredentialPlugins/CredentialPluginField.jsx
```diff
@@ -0,0 +1,103 @@
...
+  isDisabled: PropTypes.bool,
+};
+
+CredentialPluginField.defaultProps = {
+  validate: () => {},
+  isRequired: false,
+  isDisabled: false,
+};
+
+export default withI18n()(CredentialPluginField);
```
jlmitch5, line 103: this is a nice way to do this i think!

awx/ui_next/src/screens/Credential/shared/CredentialPlugins/CredentialPluginField.test.jsx
```diff
@@ -0,0 +1,85 @@
...
+      wrapper.unmount();
+    });
+    test('renders the expected content', () => {
+      expect(wrapper.find('CredentialPluginPrompt').length).toBe(0);
+      expect(wrapper.find('input').length).toBe(0);
+      expect(wrapper.find('KeyIcon').length).toBe(1);
+      expect(wrapper.find('CredentialPluginSelected').length).toBe(1);
+    });
+  });
+});
```
jlmitch5, line 85: nice tests

awx/ui_next/src/screens/Credential/shared/CredentialPlugins/CredentialPluginPrompt/CredentialsStep.jsx
```diff
@@ -0,0 +1,101 @@
...
+        {
+          name: i18n._(t`Name`),
+          key: 'name',
+        },
+      ]}
+    />
+  );
+}
+
+export default withI18n()(CredentialsStep);
```
jlmitch5, line 101: very simple

awx/ui_next/src/screens/Credential/shared/CredentialPlugins/CredentialPluginPrompt/MetadataStep.jsx
```diff
@@ -0,0 +1,157 @@
...
+          onClick={() => testMetadata()}
+        >
+          {i18n._(t`Test`)}
+        </TestButton>
+      </Tooltip>
+    </>
+  );
+}
+
+export default withI18n()(MetadataStep);
```
jlmitch5, line 157: i notice this is not using Formik...is that gonna cause any differences from our other forms (validation, styling, etc.)?

jlmitch5 (commented):
looking good!

awx/ui_next/src/components/FormField/PasswordInput.test.jsx
mabashian: Oh this came from https://github.com/ansible/awx/blob/devel/awx/ui_next/src/components/FormField/PasswordField.test.jsx#L39.  I just moved this test over when I split the field and the input.  I can look at refactoring that out of this test though @jakemcdermott?

awx/ui_next/src/screens/Credential/CredentialEdit/CredentialEdit.jsx
mabashian: ^^ @jlmitch5 and I talked about this one - going with a `.reduce` solution

awx/ui_next/src/screens/Credential/shared/CredentialPlugins/CredentialPluginPrompt/MetadataStep.jsx
mabashian: Ah yea - this component actually pulls the Formik state using `useFormikContext` here: https://github.com/ansible/awx/pull/7125/files#diff-98c8c894b8812d2914c1799d09dceb4cR26 so I think it should still have the same validation/styling afaik.

awx/ui_next/src/components/FormField/PasswordInput.test.jsx
mabashian: Hrm yea - it doesn't seem like that line is needed?  Tests pass without it.  I'll remove.

jlmitch5 (approved):
looks good to me!

PR #7135, opened 2020-05-23
Creation of tower_inventory_script module

sean-m-sullivan:
##### SUMMARY
Creation of tower_inventory_script module

I noticed that the collection lacked an inventory script module, and wanted to cut my teeth at creating one. 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - Collection

##### AWX VERSION
```
awx: 11.2.0
```

##### ADDITIONAL INFORMATION
new_name parameter does not work, tested it, compared code to other modules, may need to be removed.


wenottingham:
My concern is that this is a deprecated feature, so I'm a bit leery of having a module in the collection that may be out of date soon.

sean-m-sullivan:
I am looking at that now. Part of this was learning how to make a module. Do still see inventory scripts in 3.7. 

My gut opinion is that modules/updates should be tied to their respective tower/awx releases. What this means for the future of the collection as it applies to Tower/AWX and support of the collection, is something to be considered.

wenottingham:
Storing the custom scripts (or plugins) in a SCM project. It's a much better way to manage them regardless.

sean-m-sullivan:
True, I was just looking at that, Closing this. Thanks.

PR #7143, opened 2020-05-26
[WIP] Allow org admins to remove labels

rooftopcellist:
##### SUMMARY

Org Admin's couldn't unattach a label from a WFJT because the `can_unattach()` method in it's access class didn't cover that case.  

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
11.2.0
```

##### ADDITIONAL INFORMATION

I would like to add a test for this.  I am also still investigating why this is not needed for Job Templates.  There may be a better place for this fix.

### DIFF
--- a/awx/main/access.py
+++ b/awx/main/access.py
@@ -495,7 +495,7 @@ class NotificationAttachMixin(BaseAccess):
             # due to this special case, we use symmetrical logic with attach permission
             return self._can_attach(notification_template=sub_obj, resource_obj=obj)
         return super(NotificationAttachMixin, self).can_unattach(
-            obj, sub_obj, relationship, relationship, data=data
+            obj, sub_obj, relationship, data=data
         )
 
 


AlanCoding:
why is this not needed for attaching, in addition to JTs? also curious about that.

jbradberry:
:eyes:

jbradberry (changes requested):
Based on my investigation just now on the devel branch, the issue is that a `TypeError` is raised on line 498 due to a typo.  Fixing the typo appears to allow an org admin to remove the label without problem, unless there is more to the issue than I realize.

https://github.com/ansible/awx/blob/f05bd0568fb65a09fd2a83639578b70f799c131e/awx/main/access.py#L498

rooftopcellist:
@jbradberry is absolutely right.  I just updated the PR.  

@AlanCoding It is not needed for attaching, I just reverted my prior change to the unattach method.  

It looks like this was just a typo in this commit: https://github.com/ansible/awx/commit/4eab362318a706290ae005c2e5c2840605735962#diff-c14c9d695bfbeb70889f2b42b18ed42dR495

jbradberry (approved):
:moyai: :+1:

AlanCoding (approved):

Issue #7139, opened 2020-05-25
SumoLogic external logging in Ansible Tower 3.7 does not work


luckylittle:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
SumoLogic collector is typically URL with token in base64 format ending with two equal signs `==`. When this is saved in Tower, the incorrect URL will be stored in `/var/lib/awx/rsyslog/rsyslog.conf` where it automatically replaces those signs with `%3D%3D`. Underlying curl used by `rsyslogd` will further try to encode it and ends up with `%253D%253D`.

##### ENVIRONMENT
* AWX version: Red Hat Tower `3.7.0`
* AWX install method: https://releases.ansible.com/ansible-tower/setup-bundle/ansible-tower-setup-bundle-3.7.0-4.tar.gz
* Ansible version:  Red Hat Ansible `2.9.7`
* Operating System on the server: Red Hat Enterprise Linux release 8.2 (Ootpa)
* Web Browser: Chrome `80.0.3987.122` (Official Build) (64-bit) on Red Hat Enterprise Linux release 8.0 (Ootpa)

##### STEPS TO REPRODUCE
1. `https://<TOWER>/#/settings/system` - LOGGING
2. Logging aggregator type: `sumologic`
3. Logging aggregator: `https://<REDACTED>.sumologic.com/receiver/v1/http/<LONG TOKEN>==`

##### EXPECTED RESULTS

Integration with SumoLogic works. Logs will appear in the SumoLogic collector.

##### ACTUAL RESULTS

Integration with SumoLogic does not work due to incorrect parsing of the URL where external logs will be sent to. Logs do not appear in the SumoLogic collector.

##### ADDITIONAL INFORMATION

`cat /var/lib/awx/rsyslog/rsyslog.conf`

```
$DebugFile <REDACTED>
$DebugLevel 2
$WorkDirectory /var/lib/awx/rsyslog
$MaxMessageSize 700000
$IncludeConfig /var/lib/awx/rsyslog/conf.d/*.conf
main_queue(queue.spoolDirectory="/var/lib/awx" queue.maxdiskspace="1g" queue.type="Disk" queue.filename="awx-external-logger-backlog")
module(load="imuxsock" SysSock.Use="off")
input(type="imuxsock" Socket="/var/run/awx-rsyslog/rsyslog.sock" unlink="on")
template(name="awx" type="string" string="%rawmsg-after-pri%")
module(load="omhttp")
[...]

PR #7159, opened 2020-05-27
Correctly parse sumologic url paths

rooftopcellist:
##### SUMMARY
Related Issue: https://github.com/ansible/awx/issues/7139

 - Sumologic includes a token with a '==' at the end of it's host path.
   This gets encoded once by urlparse, and again by rsyslog when sending
   logs.  This change ensures it is only encoded.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
11.2.0
```

### DIFF
--- a/awx/main/tests/unit/api/test_logger.py
+++ b/awx/main/tests/unit/api/test_logger.py
@@ -35,7 +35,7 @@ data_loggly = {
 # Test reconfigure logging settings function
 # name this whatever you want
 @pytest.mark.parametrize(
-    'enabled, type, host, port, protocol, expected_config', [
+    'enabled, log_type, host, port, protocol, expected_config', [
         (
             True,
             'loggly',
@@ -135,9 +135,20 @@ data_loggly = {
                 'action(type="omhttp" server="yoursplunk.org" serverport="8088" usehttps="off" allowunsignedcerts="off" skipverifyhost="off" action.resumeRetryCount="-1" template="awx" errorfile="/var/log/tower/rsyslog.err" action.resumeInterval="5" restpath="services/collector/event")',  # noqa
             ])
         ),
+        (
+            True,  # valid sumologic config
+            'sumologic',
+            'https://endpoint5.collection.us2.sumologic.com/receiver/v1/http/ZaVnC4dhaV0qoiETY0MrM3wwLoDgO1jFgjOxE6-39qokkj3LGtOroZ8wNaN2M6DtgYrJZsmSi4-36_Up5TbbN_8hosYonLKHSSOSKY845LuLZBCBwStrHQ==', # noqa
+            None,
+            'https',
+            '\n'.join([
+                'template(name="awx" type="string" string="%rawmsg-after-pri%")\nmodule(load="omhttp")',
+                'action(type="omhttp" server="endpoint5.collection.us2.sumologic.com" serverport="443" usehttps="on" allowunsignedcerts="off" skipverifyhost="off" action.resumeRetryCount="-1" template="awx" errorfile="/var/log/tower/rsyslog.err" action.resumeInterval="5" restpath="receiver/v1/http/ZaVnC4dhaV0qoiETY0MrM3wwLoDgO1jFgjOxE6-39qokkj3LGtOroZ8wNaN2M6DtgYrJZsmSi4-36_Up5TbbN_8hosYonLKHSSOSKY845LuLZBCBwStrHQ==")',  # noqa
+            ])
+        ),
     ]
 )
-def test_rsyslog_conf_template(enabled, type, host, port, protocol, expected_config):
+def test_rsyslog_conf_template(enabled, log_type, host, port, protocol, expected_config):
     
     mock_settings, _ = _mock_logging_defaults()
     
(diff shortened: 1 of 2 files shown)


awx/main/utils/external_logging.py
```diff
@@ -15,6 +15,7 @@ def construct_rsyslog_conf_template(settings=settings):
     host = getattr(settings, 'LOG_AGGREGATOR_HOST', '')
     port = getattr(settings, 'LOG_AGGREGATOR_PORT', '')
     protocol = getattr(settings, 'LOG_AGGREGATOR_PROTOCOL', '')
+    type = getattr(settings, 'LOG_AGGREGATOR_TYPE', '')
```
ryanpetrello: Let's pick a different variable name that isn't a Python builtin.

rooftopcellist: Good call, that is fixed now (there and in the tests).

awx/main/utils/external_logging.py
```diff
@@ -34,7 +35,11 @@ def construct_rsyslog_conf_template(settings=settings):
         'input(type="imuxsock" Socket="' + settings.LOGGING['handlers']['external_logger']['address'] + '" unlink="on")',
         'template(name="awx" type="string" string="%rawmsg-after-pri%")',
     ])
-
+    # Sumologic has a '==' at the end of the host, which should not be parsed by urlparse 
+    # because it we then be parsed again by rsyslogd.  It will be added back below.
+    if log_type == 'sumologic':
```
ryanpetrello: Should we *only* restrict this behavior to sumologic?  What if there's some other HTTP-based endpoint that happens to have an `=` in the path?  It seems like we should circumvent the `=` escaping _regardless_ of the aggregator type (I can't imagine a scenario where we'd *ever* want `=` to turn into `%253D` on the other side of the HTTP request).

ryanpetrello:
@rooftopcellist have you actually reproduced the issue described at #7139?  I'm having trouble reproducing something like it with `devel` pointed at a local TCP server:

image

```python3
(foo) bash-4.4$ cat log.py
import json
import sys
from flask import Flask, escape, request

app = Flask(__name__)

@app.route('/', defaults={'path': ''}, methods=['POST'])
@app.route('/<path:path>', methods=['POST'])
def catch_all(path):
    print(request.environ['PATH_INFO'])
    return ''
```

image

If you're seeing this in certain environments and not others, this sounds like it could be a behavioral difference in `libcurl` between different environments (which we may need to accommodate?)

awx/main/utils/external_logging.py
rooftopcellist: I agree, we should do this for all logging aggregators, especially now that we are just telling urlparse to not escape `=`'s.

ryanpetrello (approved):
Let's squash these two commits, and then I say 🚢

awx/main/tests/unit/api/test_logger.py
```diff
@@ -135,9 +135,31 @@
                 'action(type="omhttp" server="yoursplunk.org" serverport="8088" usehttps="off" allowunsignedcerts="off" skipverifyhost="off" action.resumeRetryCount="-1" template="awx" errorfile="/var/log/tower/rsyslog.err" action.resumeInterval="5" restpath="services/collector/event")',  # noqa
             ])
         ),
+        (
+            True,  # valid sumologic config
+            'sumologic',
+            'https://endpoint5.collection.us2.sumologic.com/receiver/v1/http/ZaVnC4dhaV0qoiETY0MrM3wwLoDgO1jFgjOxE6-39qokkj3LGtOroZ8wNaN2M6DtgYrJZsmSi4-36_Up5TbbN_8hosYonLKHSSOSKY845LuLZBCBwStrHQ==', # noqa
```
ryanpetrello, line 141: @rooftopcellist you probably don't want to embed this raw URL in our source tree, and now that you've uploaded it in a commit, you might want to revoke the token if it's real.

awx/main/tests/unit/api/test_logger.py
```diff
@@ -135,9 +135,31 @@
...
+        ),
+        (
+            True,  # sumologic config with equals signs to check parsing
+            'sumologic',
+            'https://endpoint5.collection.us2.sumologic.com/receiver/v1/http/ZaVnC4dhaV0qoiETY0=MrM3==wwLoDgO1jFgjOxE6-39qokkj3LGtOroZ8wNaN2M6DtgYrJZsmSi4-36_Up5TbbN_8hosYonLKHSSOSKY845LuLZBCBwStrHQ==', # noqa
+            None,
+            'https',
+            '\n'.join([
+                'template(name="awx" type="string" string="%rawmsg-after-pri%")\nmodule(load="omhttp")',
+                'action(type="omhttp" server="endpoint5.collection.us2.sumologic.com" serverport="443" usehttps="on" allowunsignedcerts="off" skipverifyhost="off" action.resumeRetryCount="-1" template="awx" errorfile="/var/log/tower/rsyslog.err" action.resumeInterval="5" restpath="receiver/v1/http/ZaVnC4dhaV0qoiETY0=MrM3==wwLoDgO1jFgjOxE6-39qokkj3LGtOroZ8wNaN2M6DtgYrJZsmSi4-36_Up5TbbN_8hosYonLKHSSOSKY845LuLZBCBwStrHQ==")',  # noqa
```
ryanpetrello: Also, these two test cases look very similar to me - is there some difference I'm not spotting?

rooftopcellist: @ryanpetrello I've sprinkled a few other `=`'s throughout the url, for example `TY0=MrM3==ww`, just in case sumologic also uses `=`'s as characters in their tokens.

awx/main/tests/unit/api/test_logger.py
rooftopcellist: It's a fake one, I changed the token value and domain.  thanks though!

ryanpetrello (approved):
Thanks for digging into this @rooftopcellist!

PR #7171, opened 2020-05-27
Add "of" to "Run on" field in schedule form

### LINKED ISSUE #6718: Add "of" to "Run on" field in schedule form
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Screenshot_20200415_101505

This could be more readable by adding "of" to the "The" option so that it read like:

`The First ___ day of [month]`

nixocio:
Also, add a few strings to internationalization.

closes: https://github.com/ansible/awx/issues/6718

of_year


jlmitch5 (approved):
I _think_ this works.  I know some languages have different rules when it comes to language like "the" and "of" based on how they are structured.  Out of context, the translation may not be perfectly smooth.  I can't really think of a way to fix this though, so I think this is the best solution.

PR #7055, opened 2020-05-18
Mount bootstrapping scripts in dev

jakemcdermott:
##### SUMMARY
These scripts are [executed in the container](https://github.com/ansible/awx/blob/devel/tools/docker-compose.yml#L10) every time you run `make docker-compose` (or similar commands). 

Unexpectedly, editing these scripts will not change the behavior of the dev environment because docker-compose is actually running [copies of the scripts that are baked into the container image](https://github.com/ansible/awx/blob/devel/installer/roles/image_build/templates/Dockerfile.j2#L173) at build time.

This PR updates the compose file to ~map the bootstrapping scripts to their respective locations in the container (/usr/bin/)~ so that by default you can freely modify them and test the results without needing to rebuild the entire container image.

edit: new plan is to just remove them from the dockerfile

PR #7184, opened 2020-05-28
Remove dev env futzing of supervisord.conf permissions

wenottingham:
If we just link to where it will be in the dev env, we don't need to copy it at startup, and we don't need write permissions on it.

(variation on the theme of https://github.com/ansible/awx/pull/7055)


shanemcd (approved):

ryanpetrello:
@wenottingham @shanemcd we ready to merge this?

PR #7186, opened 2020-05-28
Relocate route config, remove route render props, refactor top-level routes

jakemcdermott:
##### SUMMARY
I would like our main `index.jsx` to be a standard one-line import of the app & render but doing so is a big refactor so I want to do it in two or three stages over the next few days. This PR is step one:

- It moves the router config to its own file. As the comments suggest, I'd prefer that file to just export the config but we'll need to wait for the next version of lingui. Until then, it exports a function that you pass `i18n`. Whenever lingui3 lands, this should be super easy to change to a simple config export.
- It removes unnecessary routes, route nesting, auth switching, etc. in the form of a much needed heavy-handed refactor
- Removes most of the route `render` functions, which are being phased out in later versions of react router. Only a few remain.

### DIFF
--- a/awx/ui_next/src/routeConfig.js
+++ b/awx/ui_next/src/routeConfig.js
@@ -1,0 +1,181 @@
+import { t } from '@lingui/macro';
+
+import Applications from './screens/Application';
+import Credentials from './screens/Credential';
+import CredentialTypes from './screens/CredentialType';
+import Dashboard from './screens/Dashboard';
+import Hosts from './screens/Host';
+import InstanceGroups from './screens/InstanceGroup';
+import Inventory from './screens/Inventory';
+import InventoryScripts from './screens/InventoryScript';
+import { Jobs } from './screens/Job';
+import ManagementJobs from './screens/ManagementJob';
+import NotificationTemplates from './screens/NotificationTemplate';
+import Organizations from './screens/Organization';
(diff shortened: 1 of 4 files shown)


awx/ui_next/src/routeConfig.js
```diff
@@ -0,0 +1,181 @@
...
+import AuthSettings from './screens/AuthSetting';
+import JobsSettings from './screens/JobsSetting';
+import SystemSettings from './screens/SystemSetting';
+import UISettings from './screens/UISetting';
+import License from './screens/License';
+import Teams from './screens/Team';
+import Templates from './screens/Template';
+import Users from './screens/User';
+
+// Ideally, this should just be a regular object that we export, but we
```
AlexSCorey, line 27: I think we need to start tracking these comments in the code with an issue.  I realize that when we adopt lingui3 this will be more of an enhancement, but it would be good to track.

AlexSCorey (approved):
The code looks good to me.  No blockers for approval, however I do think we need to have a way to track that comment. 

You do have a failing test

awx/ui_next/src/routeConfig.js
jakemcdermott: @AlexSCorey ok. done. https://github.com/ansible/awx/issues/7215

Issue #7060, opened 2020-05-18
Replace bubblewrap-based task execution in AWX with container-based execution


ryanpetrello:
This is a subtask of a larger enhancement/initiative surrounding execution environments: https://github.com/ansible/awx/issues/5157

AWX's current execution system is limited to executing jobs under [bubblewrap](https://github.com/containers/bubblewrap) for process isolation. For OpenShift and Kubernetes based deployments, this requires any container running jobs to run in privileged mode.

In a future release of AWX, we'd like to replace this execution mechanism with execution based strictly on containers, likely via Docker and/or podman.

This is a multi-stepped change that will require:

1.  Changes to the https://github.com/ansible/ansible-runner project itself to support container-based execution.
2. Modification to AWX's task execution to interface with runner to properly mount certain volume data (such as the appropriate project/SCM data that contains playbooks)
3.  Changes to the way that AWX executes and manages the JSON artifact data of inventory imports (today, inventory imports in AWX rely on a specific `awx-manage` process that is necessary to save inventory/group/host artifacts into the database).
4.  Certain types of "jobs" will continue to follow the model of local execution (meaning, they won't run in a container), in particular system jobs.

Other considerations include:

- How will this affect existing users that have grown accustomed to the "bwrap whitelist" mechanism, by which certain file paths can be "exposed" to the ansible-playbook runtime?  Should we continue to support this type of functionality?  How will that affect "Job Isolation"-specific settings that exist today.
- This new model will require AWX installations to either include to interact with a container registry.  What should be the default behavior for new installs (are users required to bring their own, or will all AWX installations provide a default container registry via the galaxy-ng project (https://github.com/ansible/galaxy_ng)?

Vacant0mens:
@ryanpetrello if container-based execution is what you're going for, within Kubernetes (if you're running un-privileged) the runner wouldn't be able to create containers on the Kubernetes cluster that AWX/ansible is running on. You would need an extra pod/container that runs Docker or some other container engine within the container that you would communicate to via an Api, or have the engine included in the runner's container or pod in order to call it locally.
If you're going to be calling a container runtime engine locally, it'd probably be easier to just run the tasks locally (within the container) as sub-processes or as a separate process. Isn't that what BubbleWrap is for though?

[...]

PR #3116, opened 2019-01-31
Removed unnecessary adding privileged service account

davidkirwan:
##### SUMMARY

AWX should not require cluster admin privileges to install on Openshift, nor should it add a privileged service account during install

Issue #3115 should be tackled to instead ensure the AWX container does not require root.

related #3114
related #3115

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - Installer

##### AWX VERSION

```
awx: 3.0.0
```

##### ADDITIONAL INFORMATION

```

```

PR #7188, opened 2020-05-29
Reshape security context for AWX containers

mo-saeed:
##### SUMMARY
We have Kubernetes cluster with [restricted pod security policy](https://raw.githubusercontent.com/kubernetes/website/master/content/en/examples/policy/restricted-psp.yaml) as the default policy.

When trying to install AWX with Kubernetes installer, AWX pod failed with the following errors:
```
unable to validate against any pod security policy: [spec.containers[1].securityContext.privileged: Invalid value: true: Privileged containers are not allowed]
```

```
 Error: the container has runAsNonRoot and the image has non-numeric user (memcache), cannot verify a user is non-root
```
So it's because awx-task is running as privileged container and unknow user for memcache and redis.

Change:
- Make enabling security context configurable 
- Add default privileged value to false for all containers 
- Add a user for both Memcache and Redis as official setup suggest https://github.com/helm/charts/blob/master/stable/redis/values.yaml#L186
https://github.com/helm/charts/blob/master/stable/memcached/values.yaml#L75

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - Installer (kubernetes 1.16)

##### AWX VERSION
11.2.0


ryanpetrello (changes requested):
The reason we do this is because of AWX's usage of https://github.com/containers/bubblewrap for process/playbook isolation inside our containers.

see: https://github.com/ansible/awx/pull/3116#issuecomment-459412897

Our long term plans to avoid this requirement are to move _away_ from bubblewrap, and instead support running playbooks via Docker/podman, or via container execution via the k8s API.  Details here:

https://github.com/ansible/awx/issues/7060

Until then, we're unlikely to change this default behavior.

mo-saeed:
Hi @ryanpetrello, Thanks for your review.

I have looked into the comment and the issue you mentioned and sounds to me it's different than this one here. I have tested this change and it worked fine with `"privileged": false`.

I am not sure I get this 
```
process/playbook isolation inside our containers
```
which containers do you mean? and what do you mean by process isolation? 

And in general, the value is still [configurable](https://github.com/mo-saeed/awx/blob/devel/installer/roles/kubernetes/defaults/main.yml#L20) so it can be adapted to be true or false based on each installation method and Kubernetes requirements.

Thanks again

ryanpetrello:
Hey @mo-saeed,

Here's the longer answer:

In a k8s installation today, the `awx_task` container is where `ansible-playbook` _actually runs_.  Inside of this container, multiple playbooks can and often _do_ run in parallel (alongside other AWX Python services).  Due to the nature of AWX's current architecture, this container **also** has direct access to the AWX database (because services running in the `awx_task` container - like the callback receiver which persists Ansible event data - _need_ direct access to the database):

image

This presents a security challenge - one of the guarantees that we *want* to provide is that playbooks executing in this container should *not* be able to interfere or interact with _other_ playbooks (because they could represent automation/content from some **other** tenant/Organization), and should _not_ have access to all of the aspects of the system available to the `awx` user (like, as described above, the database - because the users who write and run a playbook are not inherently trusted to have this level of system access on the underlying container - imagine a nefarious playbook that could affect the behavior of other playbooks, or worse, access restricted data from some _other_ tenant via the postgres database).

To solve this problem under the current architecture _today_, playbooks running in this environment are wrapped with https://github.com/containers/bubblewrap to provide a form of isolation apart from each other:

image

```shell
rpetrell@cloudshell:~ $  gcloud container clusters get-credentials awx-rpetrello --zone us-east1-b && kubectl exec awx-5fbd895548-w975l --namespace awx-rpetrello -c awx-task -- ps 
-ef | grep bwrap
Fetching cluster endpoint and auth data.
kubeconfig entry generated for awx-rpetrello.
root         141     131  8 16:36 pts/0    00:00:00 /usr/bin/bwrap --die-with-parent --unshare-pid --dev-bind / / --proc /proc --bind /tmp/ansible_runner_pi_t17kffo9/ansible_runner_pi_73z43a69/tmpzltwf6xa /etc/ssh --bind /tmp/ansible_runner_pi_
t17kffo9/ansible_runner_pi_73z43a69/tmpq4lfzcqv /etc/tower --bind /tmp/ansible_runner_pi_t17kffo9/ansible_runner_pi_73z43a69/tmpint4nhaj /tmp --bind /tmp/ansible_runner_pi_t17kffo9/ansible_runner_pi_73z43a69/tmpn4ozxgpa /var/lib/awx --bind /tmp
/ansible_runner_pi_t17kffo9/ansible_runner_pi_73z43a69/tmpv4xlf6u5 /var/log --ro-bind /var/lib/awx/venv/ansible /var/lib/awx/venv/ansible --ro-bind /var/lib/awx/venv/awx /var/lib/awx/venv/awx --bind /tmp/awx_1_3rno68df /tmp/awx_1_3rno68df --chd
ir /tmp/awx_1_3rno68df ansible -u admin localhost -i /tmp/awx_1_3rno68df/tmp7t_4me_u -e @/tmp/awx_1_3rno68df/env/extravars -m command -a sleep 30
```

One of the notable *downsides* of this approach is that bubblewrap _requires_ `privileged: true` to operate:

```shell
[root@a1cdb4c45f85 /]# bwrap --ro-bind /usr /usr --symlink usr/lib64 /lib64 --proc /proc --dev /dev --unshare-pid bash
bwrap: Creating new namespace failed: Operation not permitted
```

Most of this approach is from *before* AWX was open sourced, and it's a remnant of Red Hat Ansible Tower's architecture (all of this pre-dates even the existence of AWX and Tower as something that ran in containers vs VMs).  It's also an approach we intend to move away from (as described in https://github.com/ansible/awx/issues/7060).  Long term, we want to move to a model where _playbooks themselves_ are always run in distinct, isolated containers, entirely removing the need for something like https://github.com/containers/bubblewrap.

In the meantime, I disagree with:

> Add default privileged value to false for all containers

...because it means we're defaulting to something that's inherently insecure for multi-tenant use cases.

mo-saeed:
@ryanpetrello Thank you for the detailed explanation, now I get it :) 

So I have made some adaptions and hope it will address your concerns:
- Add default privileged value to false for all containers **except awx_task (default is true)**
- Add an option (default to false) to create a Pod security policy with allowed privileged containers to cover the case of having a cluster with default full restricted policy, then users can change the `awx_psp_create` to `true` to create awx's psp with allowed privileged containers.

Please let me know if you still have any concerns.

Thanks a lot for your time and support here.

kdelee (approved):
I tried this out deploying awx w/ containers built from this branch to a gke cluster and can confirm that if you don't do anything, the behavior is same as current installer -- e.g. task container is privileged and you can use bubblewrap.

Setting `task_security_context_privileged=false` deploys awx w/o the task container being privileged correctly makes this happen:
```
        name: awx-task
        resources:
          requests:
            cpu: 1500m
            memory: 2Gi
        securityContext:
          privileged: false
```

And if you try and launch a job w/ bubblewrap enabled, it fails when trying to call `bwrap`. Disabling process isolation in the settings fixes this and you are able to run jobs.

kdelee:
needs a rebase, had one conflict but it is not difficult to resolve:

```
diff --git a/installer/roles/kubernetes/templates/deployment.yml.j2 b/installer/roles/kubernetes/templates/deployment.yml.j2
index 243b235c91..d03f7ba1e6 100644
--- a/installer/roles/kubernetes/templates/deployment.yml.j2
+++ b/installer/roles/kubernetes/templates/deployment.yml.j2
@@ -159,7 +159,7 @@ spec:
             privileged: {{ web_security_context_privileged }}
 {% endif %}
 {% endif %}
-          image: "{{ kubernetes_web_image }}:{{ kubernetes_web_version }}"
+          image: "{{ kubernetes_awx_image }}:{{ kubernetes_web_version }}"

```

need to rename `kubernetes_web_image` there to `kubernetes_awx_image`

mo-saeed:
@kdelee Thank you for your approval. Can this be merged?

installer/roles/kubernetes/templates/deployment.yml.j2
```diff
@@ -270,6 +344,15 @@ spec:
               cpu: "{{ task_cpu_limit }}m"
 {% endif %}
         - name: {{ kubernetes_deployment_name }}-redis
+{% if redis_security_context_enabled is defined and redis_security_context_enabled | bool %}
+          securityContext:
+{% if redis_security_context_privileged is defined %}
+            privileged: {{ redis_security_context_privileged }}
+{% endif %}
+{% if redis_security_context_user is defined %}
+            runAsUser: {{ redis_security_context_user }}
```
shanemcd, line 347: I am not sure about this part. OpenShift (our primary downstream target) will automatically run under a random UID. Can you remove this separate conditional and include it in the `redis_security_context_enabled ` conditional above it?

installer/roles/kubernetes/templates/deployment.yml.j2
```diff
@@ -295,6 +378,15 @@ spec:
               cpu: "{{ redis_cpu_limit }}m"
 {% endif %}
         - name: {{ kubernetes_deployment_name }}-memcached
+{% if memcached_security_context_enabled is defined and memcached_security_context_enabled | bool %}
+          securityContext:
+{% if memcached_security_context_privileged is defined %}
+            privileged: {{ memcached_security_context_privileged }}
+{% endif %}
+{% if memcached_security_context_user is defined %}
+            runAsUser: {{ memcached_security_context_user }}
```
shanemcd: Same request as above.

installer/roles/kubernetes/templates/deployment.yml.j2
mo-saeed: Hi @shanemcd,

I am not sure if I get what do you want to change here, the `redis_security_context_enabled` is already in the above level so if it's false the user part will not be included anyway.

mo-saeed:
@shanemcd @kdelee It would be great if we can push this pull request somehow coz I need every time to resolve conflicts and that consumes time, so I appreciate your fast response here. 

Thanks a lot for your time and effort

benthomasson:
I'd like this change as well so that I can run AWX on openshift without needing a privileged container.

installer/roles/kubernetes/templates/deployment.yml.j2
shanemcd: Sorry, apparently I cant read jinja. It was hard to tell that this was all wrapped in the outer `if`.

shanemcd (approved):

PR #7190, opened 2020-05-29
Stop bouncing k8s deployment post-install

shanemcd:
We shouldnt need to do this now that RabbitMQ autoclustering is gone.


chrismeyersfsu (approved):

ryanpetrello (approved):
🎉

shanemcd:
Seems to work 🤷

Issue #7191, opened 2020-05-29
Invalid HTTP request when Conjur credential secret has a space in the name


AndrewCopeland:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Invalid HTTP request when Conjur credential secret has a space in the name

##### STEPS TO REPRODUCE

Create a conjur credential containing a space.
When retrieving the credential a 404 is returned from conjur.
The reason for this is because the conjur credential plugin url encodes the secretID with a `+` however conjur is expecting a `%20`.

##### EXPECTED RESULTS

To get back the credential content.

##### ACTUAL RESULTS

AWX fails to retrieve the credential and a 404 is returned.

ryanpetrello:
Hey @AndrewCopeland,

Mind giving this PR at try?

https://github.com/ansible/awx/pull/7200

Tovli:
Hi,
Can you say when will this fix be released? what version could we see it in?

Tovli:
This is related to:
https://github.com/ansible/awx/pull/7249

ryanpetrello:
@Tovli,

https://github.com/ansible/awx/pull/7214/

Tovli:
Thanks. So we should see it in version 12.
That's great.
Any date/month that you plan to release this version in?

ryanpetrello:
Hey @Tovli,

Probably somewhere in the coming weeks.  I'd say "likely by the end of June".

Tovli:
@ryanpetrello 
I actually see a release of 12.0.0 but this issue is open
Wasn't it part of 12.0.0 after all?

ryanpetrello:
Hey @Tovli,

The changes in question should be in the 12.0.0 release.  We keep issues open until we've done actual per-issue verification on our end (that's what the `state:needs_test` label on this issue represents).

one-t:
Tested - this is now working as expected

PR #7192, opened 2020-05-29
fix secret path url encoding. related #7191

AndrewCopeland:
##### SUMMARY
Decided to use `urllib.parse.quote` rather than `urllib.parse.quote_plus` on the `secret_path` because conjur expects spaces to be `%20` rather than `+`.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
And version

### DIFF
--- a/awx/main/credential_plugins/conjur.py
+++ b/awx/main/credential_plugins/conjur.py
@@ -1,7 +1,7 @@
 from .plugin import CredentialPlugin, CertFiles
 
 import base64
-from urllib.parse import urljoin, quote_plus
+from urllib.parse import urljoin, quote_plus, quote
 
 from django.utils.translation import ugettext_lazy as _
 import requests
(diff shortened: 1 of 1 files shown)


jakemcdermott:
Thanks for the PR @AndrewCopeland 

We ended up fixing `secret_path` and a few other fields in a different PR: https://github.com/ansible/awx/pull/7200 so I am going to close this.

orenbm:
@AndrewCopeland can you change the use of `quote_plus` to `quote` for all parameters? In case they use a Conjur user with a space (or any other parameter). We never use the plus in our decoding.

ryanpetrello:
Hey @orenbm,

That's precisely what we've done in this PR that ended up merging into AWX:

https://github.com/ansible/awx/pull/7200

(which will be in the next release of AWX)

orenbm:
hey @ryanpetrello :)

Thanks! I actually ran a test today and found that the fix you did (and i did) is incorrect. Please see [this PR](https://github.com/cyberark/ansible-conjur-collection/pull/11) for more info and the correct way to fix this.

ryanpetrello:
Thanks @orenbm, I think this PR that's about to go in covers what I missed: https://github.com/ansible/awx/pull/7249

Issue #289, opened 2017-09-25
Fan-out project updates across cluster


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

As an extension of #287, this issue is to implement the ability to easily fan-out a project update to all nodes in the cluster at once.

With 287 solved, this means that subsequent jobs will not need access to source control. Without the fan-out ability, the solution to low-availability source control will only work for a standalone deployment of AWX. The fan-out will allow a user to prepare their entire cluster for continuous operation with total loss of connectivity to source control.

##### STEPS TO REPRODUCE

 - Deploy AWX with multiple instances in the cluster
 - Update a project

##### EXPECTED RESULTS

The ability to have this update happen on all nodes on the cluster - bring them all up to sync with the current source control, such that subsequent jobs using that project do not need to reach out to source control.

##### ACTUAL RESULTS

Can not do this.

wenottingham:
Questions:
- what does this mean for offline cluster nodes that then rejoin the cluster?
- what does this mean if it fails on node X?
- what does this mean for `roles/requirements.yml` that refer to git HEAD of a role repo... that may change between syncs on different cluster nodes?

AlanCoding:
All good questions.

> what does this mean if it fails on node X?

Technically this is an optimization. If connectivity _does_ exist, then the project sync on a job ran on node X should sync to the correct revision before running a job.

That's just my 2 cents, of course. Same would apply to the 1st bullet, and I can't come up with an answer to the 3rd bullet off the top of my head.

AlanCoding:
After working on https://github.com/ansible/awx/pull/2287, I think I have a fairly good grasp on an implementation that would work.

 - on whatever node it lands on, perform a check type project update
 - in final_run_hook or post_run_hook
    - create "sync" type jobs for all nodes in system
    - submit a broadcast_all task with kwargs of node names & job ids
 - In that task, import `RunProjectUpdate`, with some slightly new re-entrant logic, pick up existing update id, and run the task
 - in the original node, poll the database for status of the child sync jobs, also watching for cancel_flag

AlanCoding:
I withdraw my last comment. Since the introduction of sliced jobs, it makes much more sense to continue the pattern of workflow-like jobs on an as-needed basic. My new preferred design:

 - as now, perform check type project update
 - Introduce new `changed` status to show when project_update.scm_revision != project.scm_revision, thereby signaling a change to canonical version
[...]

PR #7194, opened 2020-05-29
[WIP] cache downloaded roles & collections

wenottingham:
##### SUMMARY
Add some minimal caching of the current revision, populated the first time a job is run for that revision.

Clear it when the project revision changes, so we don't grow without bound.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -2333,6 +2333,20 @@ class RunProjectUpdate(BaseTask):
                 else:
                     self.original_branch = git_repo.active_branch
 
+    def clear_project_cache(self, instance, revision):
+        cache_dir = instance.get_cache_path()
+        if os.path.isdir(cache_dir):
+            for entry in os.listdir(cache_dir):
+                old_path = os.path.join(cache_dir, entry)
+                if entry != revision:
+                    # invalidate, then delete
+                    new_path = os.path.join(cache_dir,'.~~delete~~' + entry)
+                    try:
+                        os.rename(old_path, new_path)
+                        shutil.rmtree(new_path)
+                    except OSError:
+                        logger.warning(f"Could not remove cache directory {old_path}")
+
     @staticmethod
     def make_local_copy(project_path, destination_folder, scm_type, scm_revision):
         if scm_type == 'git':
@@ -2366,6 +2380,7 @@ class RunProjectUpdate(BaseTask):
         # To avoid hangs, very important to release lock even if errors happen here
         try:
             if self.playbook_new_revision:
+                self.clear_project_cache(instance, self.playbook_new_revision)
                 instance.scm_revision = self.playbook_new_revision
                 instance.save(update_fields=['scm_revision'])
             if self.job_private_data_dir:
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -2181,6 +2181,7 @@ class RunProjectUpdate(BaseTask):
             'scm_url': scm_url,
             'scm_branch': scm_branch,
             'scm_clean': project_update.scm_clean,
+            'project_cache': project_update.get_cache_path(),
             'roles_enabled': settings.AWX_ROLES_ENABLED,
             'collections_enabled': settings.AWX_COLLECTIONS_ENABLED,
         })
--- a/awx/main/models/projects.py
+++ b/awx/main/models/projects.py
@@ -604,9 +604,8 @@ class ProjectUpdate(UnifiedJob, ProjectOptions, JobNotificationMixin, TaskManage
         added_update_fields = []
         if not self.job_tags:
             job_tags = ['update_{}'.format(self.scm_type)]
-            if self.job_type == 'run':
-                job_tags.append('install_roles')
-                job_tags.append('install_collections')
+            job_tags.append('install_roles')
+            job_tags.append('install_collections')
             self.job_tags = ','.join(job_tags)
             added_update_fields.append('job_tags')
         if self.scm_delete_on_update and 'delete' not in self.job_tags and self.job_type == 'check':
--- a/awx/main/models/projects.py
+++ b/awx/main/models/projects.py
@@ -194,6 +194,11 @@ class ProjectOptions(models.Model):
             if not check_if_exists or os.path.exists(smart_str(proj_path)):
                 return proj_path
 
+    def get_cache_path(self):
+        local_path = os.path.basename(self.local_path)
+        if local_path:
+            return os.path.join(settings.PROJECTS_ROOT, '.__awx_cache', local_path)
+
     @property
     def playbooks(self):
         results = []
@@ -455,11 +460,12 @@ class Project(UnifiedJobTemplate, ProjectOptions, ResourceMixin, CustomVirtualEn
         )
 
     def delete(self, *args, **kwargs):
-        path_to_delete = self.get_project_path(check_if_exists=False)
+        paths_to_delete = (self.get_project_path(check_if_exists=False), self.get_cache_path())
         r = super(Project, self).delete(*args, **kwargs)
-        if self.scm_type and path_to_delete:  # non-manual, concrete path
-            from awx.main.tasks import delete_project_files
-            delete_project_files.delay(path_to_delete)
+        for path_to_delete in paths_to_delete:
+            if self.scm_type and path_to_delete:  # non-manual, concrete path
+                from awx.main.tasks import delete_project_files
+                delete_project_files.delay(path_to_delete)
         return r
 
 
--- a/awx/playbooks/project_update.yml
+++ b/awx/playbooks/project_update.yml
@@ -136,7 +136,7 @@
 
         - name: populate job directory with needed roles
           synchronize:
-            src: "{{ cache_dir }}/requirements_roles"
+            src: "{{ cache_dir }}/requirements_roles/"
             dest: "{{ roles_destination }}"
           when: roles_destination is defined and doesRequirementsExist.stat.exists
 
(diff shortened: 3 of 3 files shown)


awx/main/tasks.py
```diff
@@ -2327,6 +2328,19 @@ def pre_run_hook(self, instance, private_data_dir):
...
+    def clear_project_cache(self, instance, revision):
+        cache_dir = instance.get_cache_path()
+        for entry in os.listdir(cache_dir):
+            old_path = os.path.join(cache_dir, entry)
+            if entry != revision:
+                # invalidate, then delete
+                new_path = os.path.join(cache_dir,'.~~delete~~' + entry)
+                try:
+                    os.rename(old_path, new_path)
+                    shutil.rmtree(new_path)
```
wenottingham: this should be in a background thread, if we can stomach the complexity (we're inside the update lock here, and this may be a long deletion)

AlanCoding:
Remove the `if` from:

https://github.com/ansible/awx/blob/1a25d8072d7c9a3157f161e2bd57a6e0ea34988a/awx/main/models/projects.py#L607-L609

Explanation:

Caching implies a look & feel change. When someone clicks the update button (a manual update), the role & collection install does not happen. Instead, it is delayed until the last minute before the job run starts. The objective of the manual update changes with this PR. It's no longer only to update the revision in the database, but also allows updating the local cache on the instance where it runs. This allows earlier error reporting to the user, and also allows single-instance installs to run jobs with roles/collections offline, and would allow cluster installs to do the same if we did #289.

wenottingham:
> Remove the if from:

Done. This requires tweaking the project_update.yml to ensure we're using the new revision (for the update case) and the passed revision (for the run case).

AlanCoding:
Right, makes sense, both types of updates have a cache location, but the manual project update has no job location.

Going to have to start kicking the tires on testing for this.

wenottingham:
This will need checked that the cache isn't pulled into the backup in the backup role (although it's mostly harmless if it does.)

wenottingham:
Discussion: with this, we will not catch updates to the dependent roles/collections when you do a manual update, which is a use case we'd need to support. So further changes needed.

AlanCoding:
^ with SCM support for both roles / collections now, I want to enforce this in an integration test.

That's the big thing on my agenda, I think there's a lot of value we would get developing that.

AlanCoding:
I just started some basic testing of this. Putting aside the question of why using roles is not working, here are some other comments.

The return data structure of `synchronize` worries me, in the same way that the `copy` module did.

Screen Shot 2020-06-08 at 9 02 38 AM

Again, I _know_ that every time you are syncing to an empty directory. Like the copy module, the `synchronize` module is giving a line somewhere in its output corresponding to every file. This is a tiny test role, we need to test our solution for irresponsibly large roles.

With @ryanpetrello's recent related fixes, we're not sanitizing the return data for every module, but the large return data is still a liability, and I don't think we should have line-per-file scaling for the happy path (which we didn't for what he addressed). The project update event table would balloon for most users. I know how I would try to rework this (put it in `tasks.py`) but I don't want to touch code until I actually understand what the situation is.

----

Why are the roles not working in playbooks?

```json
"ANSIBLE_ROLES_PATH": "/tmp/awx_3659_nin6k6oo/requirements_roles:~/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles"
```

gets set in jobs. Since the `requirements_roles` path looks right, I had to set `AWX_CLEANUP_PATHS = False` to dig.

Problem is super obvious, the roles are doubly-nested under `requirements_roles`

```
[ec2-user@ip-10-0-9-74 ~]$ sudo ls /tmp/awx_3661_xnyqyov6/requirements_roles/requirements_roles
grog.debug-variable  yatesr.timezone
```

This content should be 1 level up.

The options passed to the `synchronize` module were:

```
            "src": "/var/lib/awx/projects/.__awx_cache/_3556__project_with_galaxy_requirements/31ddc6c1ce8532e947763c38b4751772bb85e207/requirements_roles",
            "dest": "/tmp/awx_3661_xnyqyov6/requirements_roles",
```

(among others)

The source structure:

```
ls /var/lib/awx/projects/.__awx_cache/_3556__project_with_galaxy_requirements/31ddc6c1ce8532e947763c38b4751772bb85e207/requirements_roles
grog.debug-variable  yatesr.timezone
```

So source is not double-nested. That implies a misunderstanding of the definition of the `src` or `dest` parameters to the synchronize module.

wenottingham:
Current behavior of this PR (note: all behavior is constrained to a single node where the task runs):

Task type | System State | Prior behavior (pinned content) | Prior behavior (unpinned content) | New behavior (pinned content) | New behavior (unpinned content)
---------| -------- | ------- | ------- | ------ | ---------
Project update | blank system | not synced | not synced | synced to cache | synced to cache 
Project update | cache present | not synced | not synced | synced to cache (forced refresh) | synced to cache (forced refresh)
Job project sync| blank system | synced to temp checkout| synced to temp checkout | synced to cache & copied to temp checkout | synced to cache & copied to temp checkout
Job project sync | cache present | synced to temp checkout | synced to temp checout | synced to cache only if missing, cache copied to temp checkout | synced to cache only if missing, cache copied to temp checkout

TL;DR of this table: if the user was relying on unversioned roles to be always updated when a job runs, this change will break that expectation (as it will use previously cached roles from a prior run of that project revision if present).

Cache is cleared:
- when delete_on_update is set (node where project update runs)
- on project revision change (node where project update runs)
- on project deletion (all nodes)

This means that the cache could grow unbounded on a (non-isolated) node that only ever runs the job and never runs the project update.

AlanCoding:
> if the user was relying on unversioned roles to be always updated when a _job runs_, this change will break that expectation

The caching behavior in this PR will not always update an unpinned role when _the project is updated_.

You have a 3 node cluster of A, B, and C. Last commit to source was 2 weeks ago. A project was created a week ago with an unpinned dependency on `geerlingguy.adminer`, and initial project update was on node B. A new version of the role (1.3.0) came out 2 days ago. Today, someone manually updates the project. This update runs on node A. It doesn't matter if the project update on node A uses `--force` or not, because when a job later runs on node B, the revision matches the revision of the project, so it will use that cache.

If the "job project sync" used the `--force` option, it would defeat the goal here. If it doesn't use that option, the command gives:

> [WARNING]: - geerlingguy.adminer (1.2.0) is already installed - use --force to change version to unspecified

The end result is that jobs will run with roles from before the last project update.

To illustrate a fix, you could change the cache directory so that it doesn't use the project revision, but uses the full timestamp of the project's last update. Some variation of this might actually work, but I'm not getting into that right now. All I'm trying to show here is that actually solving the problem isn't necessary hard, it just has this very meticulous hazard with it.

moonrail:
@AlanCoding
If I understand correctly, one can use "Update Revision on Launch" in combination with "Delete on Update" to always pull dependencies, even after this PR?

I'd like to note, that the ansible-galaxy option "--force" would not update dependencies of roles (meta/main.yml).
For this behaviour the option "--force-with-deps" is required.

In our company we are using Git Tags for versioning Roles, Playbooks &, more.
But instead of semver, we only use major.minor, to be able to develop patches for pinned versions, because the collection specific feature of ansible-galaxy to specify ranges of versions is not available for Git Tags.
This has the downside, that role dependencies have to be updated regularly, to get all changes, as the main project revision will not have changed.

But this is also the case on using ansible-galaxy dependencies for collections, as one can use version strings like:
version: '>=1.0.0,<2.0.0'

So users should really have a project option to disable cache, or maybe even better a Job Template option, that allows disabling cache just for this Template (or via Prompt on Launch, this Run).

AlanCoding:
My expectation would be that updating the project will clear the cache.

We cannot rely on `ansible-galaxy` to answer the question "is the current folder up to date?" That's not a feature it has. I don't see any valid use case for force or force-with-deps, because we would just start over with a new folder. There's no logic or performance benefit to running against an existing folder.

> To illustrate a fix, you could change the cache directory so that it doesn't use the project revision, but uses the full timestamp of the project's last update

That would entail additional schema changes, consider the project's `last_job_id` as a better alternative. This is easily accessible to both the project updates and job project syncs.

AlanCoding:
Proposed code change to address the scenario I described, based on the `last_job_id` solution discussed:

https://github.com/wenottingham/awx/compare/who-needs-cache-invalidation...AlanCoding:per_update_cache?expand=1

I'm running it through the tests now.

There is a balance to strike between blowing the cache away every run (always correct, very slow) and never blowing away the cache (rarely correctly, very fast). The code change I'm suggesting moves the needle slightly further in the direction of blowing away the cache more often, and being more correct.

There is another option to blow away the cache less often than either my change, or your current PR, but still maintaining correctness. However, this would involve more logic. I can return to that later, depending on interest.

There may be holes in my proposal.

In either case, I'm worried about holes due to the override-branch-on-launch feature. The clearing of the cache looks like it would clear the cache even if using a custom branch, which probably shouldn't happen.

wenottingham:
> Proposed code change to address the scenario I described, based on the last_job_id solution discussed:

This removes the bit where it clears the cache when delete_on_update is set - I think that's wrong, we want that to remove the cache overall. This may require passing the ID down into the project_update.yml and computing that value there.

AlanCoding:
The cache is removed by changing the cache folder name. In my proposed patch, the folder location changes with every project update (job project syncs excluded). This is not relevant for job project syncs because those will not run the "delete" tag to begin with.

In the current state of this PR, the cache folder comes from the scm revision, which may or may not change with a project update. So I get why you added it, but it makes no sense to keep if my other proposed changes are accepted.

AlanCoding:
a quick thought... my branch `per_update_cache` probably has an error where selecting a new branch on launch will run and use the cache from the base branch. Wasn't picked up by any of the tests, probably needs new tests developed for it.

AlanCoding:
Hope you don't mind that I close this, I'm trying to focus energy on the replacement PR.

Issue #7018, opened 2020-05-13
JT Launch prompts: credential validation improvements


keithjgrant:
Follow-up for #5909

##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
In the Credentials step of the JT launch prompt, the following behavior should be enforced:

Default credentials can be removed but they must be replaced with another credential of the same type. This is enforced by the api. For example, if there is a default machine credential on the JT then I can replace it with another machine credential but I cannot proceed to the next step without at least one machine credential.

unlikelyzero:
@keithjgrant added enhancement instead of feature as this seems like it may not be 100% critical for launch. Is that correct?

mabashian:
Gonna close this in favor of #8200

PR #7208, opened 2020-06-01
Add error feedback in Preview Step

### LINKED ISSUE #7016: JT Launch prompts: Error feedback in Preview Step
Follow-up for #5909

##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
After reaching the final step of JT Launch prompts (the preview step), the other steps validate their form data and show errors in the navigation menu on the left. HOWEVER, there is no explicit feedback to the user that there are errors.

At the top of the Preview Step, above the details, some sort of error message should appear indicating that errors need to be addressed before the job can launch.

Some of this is currently stubbed in, but it needs to be styled appropriately, and the errors need to be passed into the PreviewStep component (either all error messages or simply a boolean indicating that errors are present).

Additionally, the Launch button should probably be disabled until errors are addressed.

nixocio:
Add error feedback in Preview Step.

closes: https://github.com/ansible/awx/issues/7016

error_preview_step

### DIFF
--- a/awx/ui_next/src/components/LaunchPrompt/steps/useCredentialsStep.jsx
+++ b/awx/ui_next/src/components/LaunchPrompt/steps/useCredentialsStep.jsx
@@ -19,7 +19,8 @@ export default function useCredentialsStep(
     initialValues: getInitialValues(config, resource),
     validate,
     isReady: true,
-    error: null,
+    contentError: null,
+    formError: null,
     setTouched: setFieldsTouched => {
       setFieldsTouched({
         credentials: true,
--- a/awx/ui_next/src/components/LaunchPrompt/useSteps.js
+++ b/awx/ui_next/src/components/LaunchPrompt/useSteps.js
@@ -13,14 +13,13 @@ export default function useSteps(config, resource, i18n) {
     useOtherPromptsStep(config, resource, visited, i18n),
     useSurveyStep(config, resource, visited, i18n),
   ];
+
+  const formErrorsContent = steps
+    .filter(s => s?.formError && Object.keys(s.formError).length > 0)
+    .map(({ formError }) => formError);
+
   steps.push(
-    usePreviewStep(
-      config,
-      resource,
-      steps[3].survey,
-      {}, // TODO: formErrors ?
-      i18n
-    )
+    usePreviewStep(config, resource, steps[3].survey, formErrorsContent, i18n)
   );
 
   const pfSteps = steps.map(s => s.step).filter(s => s != null);
@@ -31,8 +30,9 @@ export default function useSteps(config, resource, i18n) {
     };
   }, {});
   const isReady = !steps.some(s => !s.isReady);
-  const stepWithError = steps.find(s => s.error);
-  const contentError = stepWithError ? stepWithError.error : null;
+
+  const stepWithError = steps.find(s => s.contentError);
+  const contentError = stepWithError ? stepWithError.contentError : null;
 
   const validate = values => {
     const errors = steps.reduce((acc, cur) => {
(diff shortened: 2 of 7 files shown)


awx/ui_next/src/components/LaunchPrompt/useSteps.js
```diff
@@ -13,14 +13,20 @@ export default function useSteps(config, resource, i18n) {
     useOtherPromptsStep(config, resource, visited, i18n),
     useSurveyStep(config, resource, visited, i18n),
   ];
+  const formErrors = steps
+    .filter(s => s.formError != null)
+    .filter(s => s !== undefined)
+    .filter(s => Object.keys(s.formError).length > 0);
```
jlmitch5: it's better to just chain these with &&, that way it doesn't have to iterate over the steps 3 times (not that it really matters in this case, because the number of steps is not large, but it's a better convention)

awx/ui_next/src/components/LaunchPrompt/useSteps.js
```diff
@@ -13,14 +13,20 @@ export default function useSteps(config, resource, i18n) {
...
+
+  const formErrorsContent = [];
```
jlmitch5: you don't need to separate these... just tack on a map to your filter statement.  I believe you could simplify to:

```
  const formErrorsContent = steps
    .filter(s => s?.formError && Object.keys(s.formError).length > 0) // the first part is equivalent to s !== undefined && s.formError != null
    .map(({formError}) => formError)
```

awx/ui_next/src/components/LaunchPrompt/useSteps.js
```diff
@@ -31,8 +37,10 @@ export default function useSteps(config, resource, i18n) {
     };
   }, {});
   const isReady = !steps.some(s => !s.isReady);
-  const stepWithError = steps.find(s => s.error);
-  const contentError = stepWithError ? stepWithError.error : null;
+
+  const stepWithError = steps.filter(s => s.contentError);
+
+  const contentError = stepWithError ? stepWithError.contentError : null;
```
jlmitch5, line 35: stepWithError is an array (or atleast looks to be one base on your filter statement above)...it's not an object.  I don't know exactly what you're trying to do, but this won't work.  the line 34 you deleted uses "find" which is like filter, but returns the first item of the array, rather than the array itself.  maybe you meant to use find instead of filter?

awx/ui_next/src/components/LaunchPrompt/steps/useCredentialsStep.jsx
```diff
@@ -19,7 +19,7 @@ export default function useCredentialsStep(
     initialValues: getInitialValues(config, resource),
     validate,
     isReady: true,
-    error: null,
+    contentError: null,
```
jlmitch5, line 22: why did this one not get the same pattern as the other steps (there's no formError key)

jlmitch5 (changes requested):
looks to be a few tweaks that need to be made, but awesome stuff!

awx/ui_next/src/components/LaunchPrompt/useSteps.js
nixocio: You are right. It was supposed to be a simple rename from 'error' to 'contentError'. Thanks. :eyes:

awx/ui_next/src/components/LaunchPrompt/steps/useCredentialsStep.jsx
nixocio: I will add the `formError` key returning `null`. There is another issue related to the validation of credentials https://github.com/ansible/awx/issues/7018

nixocio:
Updated per comments @jlmitch5

jlmitch5 (approved):
looks good

PR #7224, opened 2020-06-02
Now respecting the HAS_YAML variable

john-westcott-iv:
##### SUMMARY

Modify the tower_api library to respect the HAS_YAML variable. Right now it will attempt to load a configuration file as yaml even if HAS_YAML=False. This will cause a stack exception. In the new version yaml loading will not be attempted if the yaml library can not be loaded (it will fall back to ConfigParsing).

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - Collection
##### AWX VERSION

```
awx: 11.2.0
```

##### ADDITIONAL INFORMATION

```
To test the bug create a new virtualenv with no libraries installed. Create a simple playbook like:
---
- name: Quick test of loading tower config
  hosts: localhost
  connection: local
  gather_facts: False
  collections:
    - awx.awx
  tasks:
    - tower_organization:
        name: 'junk'

Run the playbook with the command like:
ansible-playbook -e ansible_python_interpreter=/<path to empty venv>/bin/python quick_test.yml

This will result in a stack trace because the yaml library is not present.

With this new version of the library the command will work.
```

### DIFF
--- a/awx_collection/plugins/module_utils/tower_api.py
+++ b/awx_collection/plugins/module_utils/tower_api.py
@@ -154,37 +154,44 @@ class TowerModule(AnsibleModule):
 
         # First try to yaml load the content (which will also load json)
         try:
-            config_data = yaml.load(config_string, Loader=yaml.SafeLoader)
-            # If this is an actual ini file, yaml will return the whole thing as a string instead of a dict
-            if type(config_data) is not dict:
-                raise AssertionError("The yaml config file is not properly formatted as a dict.")
+            try_config_parsing = True
+            if HAS_YAML:
+                try:
+                    config_data = yaml.load(config_string, Loader=yaml.SafeLoader)
+                    # If this is an actual ini file, yaml will return the whole thing as a string instead of a dict
+                    if type(config_data) is not dict:
+                        raise AssertionError("The yaml config file is not properly formatted as a dict.")
+                    try_config_parsing = False
 
-        except(AttributeError, yaml.YAMLError, AssertionError):
-            # TowerCLI used to support a config file with a missing [general] section by prepending it if missing
-            if '[general]' not in config_string:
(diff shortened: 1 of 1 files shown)


awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -154,37 +154,44 @@ def load_config(self, config_path):
...
-            if '[general]' not in config_string:
-                config_string = '[general]{0}'.format(config_string)
+                except(AttributeError, yaml.YAMLError, AssertionError):
+                    try_config_parsing = True
 
-            config = ConfigParser()
+            if try_config_parsing:
+                # TowerCLI used to support a config file with a missing [general] section by prepending it if missing
+                if '[general]' not in config_string:
+                    config_string = '[general]{0}'.format(config_string)
```
AlanCoding, line 172: this is going to be a conflict with the fix to add `\n`.

If you don't want to fix conflicts, you can add it here and I'll close the other one.

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -154,37 +154,44 @@ def load_config(self, config_path):
...
+                    if type(config_data) is not dict:
+                        raise AssertionError("The yaml config file is not properly formatted as a dict.")
+                    try_config_parsing = False
 
-        except(AttributeError, yaml.YAMLError, AssertionError):
-            # TowerCLI used to support a config file with a missing [general] section by prepending it if missing
-            if '[general]' not in config_string:
-                config_string = '[general]{0}'.format(config_string)
+                except(AttributeError, yaml.YAMLError, AssertionError):
+                    try_config_parsing = True
```
AlanCoding, line 167: non-blocker: I think a simple `pass` would be easier to read than redefinition

AlanCoding (approved):

PR #7228, opened 2020-06-02
Move to one container image rather than two mostly-identical ones

wenottingham:
##### SUMMARY
This creates one 'awx' container image used by both awx_web and awx_task.

This also tweaks the supervisor configuration to use /var/run/supervisor for
all sockets, and /etc/ for all config files.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - Installer


shanemcd:
Testing this out now.

Thought: maybe we should add a changelog entry as part of this PR.

shanemcd (approved):
This works for me on both k8s and local docker installs. Thanks Bill!

PR #7229, opened 2020-06-03
Comply with updated Ansible sanity test rules

AlanCoding:
##### SUMMARY
See:

https://github.com/ansible/proposals/issues/178#issuecomment-637965720

> In this case, version_added should not be there, because the option was there when the module was introduced to the collection.

That applies for the auto-migrated collections, and other maintainers still _can_ go their own way. But the tests don't allow for pinning versions from ansible-base. So we can't keep the old stuff. That decision has already been made.

There are some other wonky things in here, because we are balancing 2 different namespaces. For the 3 deprecated modules not shipped in the `ansible.tower` collection, we are free to reference `awx.awx` versions.

In the modules where we deprecate things which are in `ansible.tower`, we prefer Tower versions over AWX versions. Thus, we ignore those rules permanently for `awx.awx`. We might have to delete those rules for the `ansible.tower` collection (because it will become correct with the namespace change, and will trigger the unnecessary-ignore rule). Let me ping @felixfontein here, because I'm sure they weren't anticipating this situation of releasing the same content under 2 different namespaces.

##### ISSUE TYPE
 - Bugfix Pull Request

### DIFF
--- a/awx_collection/plugins/doc_fragments/auth.py
+++ b/awx_collection/plugins/doc_fragments/auth.py
@@ -33,7 +33,6 @@ options:
     - The Tower OAuth token to use.
     - If value not set, will try environment variable C(TOWER_OAUTH_TOKEN) and then config files
     type: str
-    version_added: "3.7"
   validate_certs:
     description:
     - Whether to allow insecure connections to Tower or AWX.
--- a/awx_collection/plugins/inventory/tower.py
+++ b/awx_collection/plugins/inventory/tower.py
@@ -12,7 +12,6 @@ DOCUMENTATION = '''
       - Matthew Jones (@matburt)
       - Yunfan Zhang (@YunfanZhang42)
     short_description: Ansible dynamic inventory plugin for Ansible Tower.
-    version_added: "2.7"
     description:
         - Reads inventories from Ansible Tower.
         - Supports reading configuration from both YAML config file and environment variables.
@@ -68,7 +67,6 @@ DOCUMENTATION = '''
             description: Make extra requests to provide all group vars with metadata about the source Ansible Tower host.
             type: bool
             default: False
-            version_added: "2.8"
 '''
 
 EXAMPLES = '''
--- a/awx_collection/plugins/lookup/tower_schedule_rrule.py
+++ b/awx_collection/plugins/lookup/tower_schedule_rrule.py
@@ -6,7 +6,6 @@ __metaclass__ = type
 DOCUMENTATION = """
     lookup: tower_schedule_rrule
     author: John Westcott IV (@john-westcott-iv)
-    version_added: "3.7"
     short_description: Generate an rrule string which can be used for Tower Schedules
     requirements:
       - pytz
--- a/awx_collection/plugins/modules/tower_credential.py
+++ b/awx_collection/plugins/modules/tower_credential.py
@@ -17,7 +17,6 @@ DOCUMENTATION = '''
 ---
 module: tower_credential
 author: "Wayne Witzel III (@wwitzel3)"
-version_added: "2.3"
 short_description: create, update, or destroy Ansible Tower credential.
 description:
     - Create, update, or destroy Ansible Tower credentials. See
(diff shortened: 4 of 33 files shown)


felixfontein:
@AlanCoding do I understand correctly that you have one set of sources (https://github.com/ansible/awx/tree/devel/awx_collection) for two different collections (awx.awx and ansible.tower)? Do both use the same version numbers, or do these also differ?

> [ansible/proposals#178 (comment)](https://github.com/ansible/proposals/issues/178#issuecomment-637965720)
> 
> > In this case, version_added should not be there, because the option was there when the module was introduced to the collection.
> 
> That applies for the auto-migrated collections, and other maintainers still _can_ go their own way. But the tests don't allow for pinning versions from ansible-base. So we can't keep the old stuff. That decision has already been made.

Well, the consensus seemed to be that `version_added` refers to the collection version the content is contained in. The typical user story is: I need collection x.y, and I need feature z in it. So which version of the collection do I need so I get this feature? If the user looks at the docs and finds a version number unrelated to the collection version number, this is not helpful.

I guess you can still mention this in `description`.

> There are some other wonky things in here, because we are balancing 2 different namespaces. For the 3 deprecated modules not shipped in the `ansible.tower` collection, we are free to reference `awx.awx` versions.

So both collections have **different** version schemes? I unfortunately only found one of them (https://galaxy.ansible.com/awx/awx), so I cannot compare.

> In the modules where we deprecate things which are in `ansible.tower`, we prefer Tower versions over AWX versions. Thus, we ignore those rules permanently for `awx.awx`. We might have to delete those rules for the `ansible.tower` collection (because it will become correct with the namespace change, and will trigger the unnecessary-ignore rule). Let me ping @felixfontein here, because I'm sure they weren't anticipating this situation of releasing the same content under 2 different namespaces.

If you release the same content under two different namespaces, you might want to template the content in a way that the correct version numbers are added in both namespaces.

AlanCoding:
> do I understand correctly that you have one set of sources (...) for two different collections

Yes, one is on Galaxy and the other is one cloud.redhat.com.

> Do both use the same version numbers, or do these also differ?

Versions are different, because versions are hard-and-fast bound to the server versions. It is exactly as confusing as what the original decisions for server versioning were. AWX leapfrogged Tower versions intentionally, so that they don't hit the 3.x.x series.

For an idealized case moving forward, I think we would have `version_added:` match the Ansible Tower (and `ansible.tower` collection) version, and the documentation could also helpfully state the AWX version. The Tower versions would always be x.x.0, because the dot-releases should never introduce new features.

Templating (replacing) the namespace is one thing, but replacing version numbers would require a mapping of versions of one server version to the other. Such a mapping never has existed. I don't want to start such a thing strictly for the purposes of the collection.

felixfontein:
Then why not just put the ansible.tower versions into version_added in the correct format (semver, i.e. x.y.0)? ansible-test only ensures that versions are of the correct format (i.e. semver for collection versions). Also, if you want to put other version numbers there, why not simply add ignore.txt entries?

AlanCoding:
Isn't that what I'm doing here?

I think I might see a point of confusion. There are 3 deprecated modules right now, and these were explicitly excluded from `ansible.tower`. So if `tower_send` shows an AWX version, that's because this is _only_ an AWX module.

felixfontein:
Ah, I eventually found that. I had to scroll down several pages before I found a version_added that wasn't removed. I was wondering why you're still removing all these numbers. Are these Tower versions, and is Tower using versions that are not of the form `x.y.z`?

AlanCoding:
The version numbers being removed are almost entirely Ansible versions from 2.9 or earlier. Even things that were added since then didn't use `awx.awx` versions, so it seems like there's no way to salvage them.

felixfontein:
I saw a lot of `3.7` removed, that's why I assumed that all `x.y` versions must be Tower.

AlanCoding:
oh right, 3.7.0 was the initial version of `ansible.tower`, but `awx.awx` existed for a few months before then. In terms of `ansible.tower` versioning, that would still make it the initial collection version.

squidboylan (approved):

Issue #6863, opened 2020-04-28
Add Copy Support for Projects and Inventories


AlexSCorey:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
See #4969

dsesami:
Some feedback:
- A copied project in the project list is missing Git revision info in the list (the sequence of characters and numbers that tells you the latest commit). However, copying the copy makes the Git revision appear, but then that second copy is missing the Git revision afterwards. Sorting also seems to fix it, so I guess the Git info is there, it's just not being rendered for some reason initially.

- An auditor is able to view the copy button for random projects with no apparent common aspects. I'm not really sure what makes those different in the copy list. However, clicking it causes an unauthorized error.

![Screenshot_20200515_132422](https://user-images.githubusercontent.com/19942822/82078782-b480bb00-96af-11ea-9a22-90c0c869caa4.png)


- The copying timestamp is off. While it should be `Name @ 1:25:39 PM`, it is `Name @ 1:25: 39 PM`. Note the space after the second colon.

AlexSCorey:
> Some feedback:
> 
> * A copied project in the project list is missing Git revision info in the list (the sequence of characters and numbers that tells you the latest commit). However, copying the copy makes the Git revision appear, but then that second copy is missing the Git revision afterwards. Sorting also seems to fix it, so I guess the Git info is there, it's just not being rendered for some reason initially.

A newly copied Project needs to be syned before the revision will apear

> * An auditor is able to view the copy button for random projects with no apparent common aspects. I'm not really sure what makes those different in the copy list. However, clicking it causes an unauthorized error.

Sys Auditors cannot see the copy button any more based on the ui in devel.

> ![Screenshot_20200515_132422](https://user-images.githubusercontent.com/19942822/82078782-b480bb00-96af-11ea-9a22-90c0c869caa4.png)
> 
> * The copying timestamp is off. While it should be `Name @ 1:25:39 PM`, it is `Name @ 1:25: 39 PM`. Note the space after the second colon.

The time stamp has also been fixed and is working properly in devel.
@dsesami

dsesami:
@AlexSCorey and I are discussing a placeholder for the git commit when a project hasn't been synced yet.

dsesami:
@AlexSCorey I've confirmed that the time stamp spacing issue still exists in devel. I'd like to keep this open until that last fix is put in.

AlexSCorey:
[...]

PR #7239, opened 2020-06-03
Adds ux for unsynced projects

AlexSCorey:
##### SUMMARY
After discussions with @dsesami and @trahman73 we decided that there needed to be a bit more help for users on newly copied/unsynced projects.  https://github.com/ansible/awx/issues/6863#issuecomment-636073900 

This PR disables the copy to clipboard button, and adds a message for the user to all unsynced projects.  
##### ISSUE TYPE
-enhancement

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
Screen Shot 2020-06-03 at 2 59 55 PM

### DIFF
--- a/awx/ui_next/src/components/ClipboardCopyButton/ClipboardCopyButton.jsx
+++ b/awx/ui_next/src/components/ClipboardCopyButton/ClipboardCopyButton.jsx
@@ -41,10 +41,10 @@ class ClipboardCopyButton extends React.Component {
 
   render() {
     const {
-      clickTip,
+      copyTip,
       entryDelay,
       exitDelay,
-      hoverTip,
+      copiedSuccessTip,
       isDisabled,
     } = this.props;
     const { copied } = this.state;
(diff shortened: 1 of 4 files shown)


dsesami (approved):
works great.

jakemcdermott (commented):
@AlexSCorey This looks good. As discussed offline, I think we'll want to adjust the wording slightly as "active" and "activate" don't fit with all scenarios. Once we update that I think this will be ready to go.

awx/ui_next/src/components/ClipboardCopyButton/ClipboardCopyButton.jsx
```diff
@@ -51,6 +57,7 @@ class ClipboardCopyButton extends React.Component {
         content={copied ? clickTip : hoverTip}
       >
         <Button
+          isDisabled={isDisabled}
           variant="plain"
           onClick={this.handleCopyClick}
           aria-label={hoverTip}
```
jlmitch5: the label doesn't change with the content, meaning the two could be different.

awx/ui_next/src/components/ClipboardCopyButton/ClipboardCopyButton.jsx
```diff
@@ -40,7 +40,13 @@ class ClipboardCopyButton extends React.Component {
   };
 
   render() {
-    const { clickTip, entryDelay, exitDelay, hoverTip } = this.props;
+    const {
+      clickTip,
+      entryDelay,
+      exitDelay,
+      hoverTip,
```
jlmitch5: can we change the name to copyTip copiedSuccessTip or something like that?  It's currently confusing, because it is the tip content that is changing...there is no difference in the way the tip is triggered and displayed as "click and hover" would have you believe.

jlmitch5 (changes requested):

jakemcdermott (approved):

jlmitch5 (approved):
looks good

Issue #6932, opened 2020-05-05
Replace local memcached usage with redis


ryanpetrello:
Now that every AWX node has a local independent redis server, we can probably just get rid of our memcached usage (and just use redis instead).

In modern versions of AWX, memcached is used specifically for:

- a local cache for AWX settings access (see: `awx.conf.settings`); any time `django.conf.settings` is accessed (e.g., `settings.TOWER_URL_BASE`), we *first* query memcached via Django's cache framework (https://docs.djangoproject.com/en/3.0/topics/cache/) to see if the value exists in memcached (if not, we pull the value from the database and _then_ place it in the cache)
- any time a setting value is *changed* via the API, we run a task on **every** AWX node, which removes the _old_ value from that nodes' local memcached
- there are various other places where AWX code interacts with memcached, but it generally does so using Django's cache interface i.e., `cache.get()`, `cache.set()`
- AWX *used to* use memcached as a holding space for in-flight playbook event data, and for fact gathering, but we've removed this type of usage

##### PLAN OF ACTION

I highly suspect that **all** of the above interactions are implemented *only* through Django's cache interface (i.e., I don't think we've got code that _directly_ accesses memcached using the `python-memcached` library), so replacing memcached with redis **should** be as simple as just swapping out the dependency and the Django caching configuration.

- investigate https://github.com/jazzband/django-redis as an alternative cache plugin, and add it to the list of dependencies in `./requirements`
- remove the `python-memcached` library from our list of dependencies in `./requirements`
- update our definition of `CACHES` under `awx.settings` to point at the redis Unix domain socket _instead of_ memcached
- remove any and all installer code and configuration that provisions memcached containers
   - in the local Docker installer
   - in the k8s installer
- remove any integration or functional tests that test/verify memcached connectivity at startup/install time

kdelee:
main thing I see as missing from our testing is this case:

 "I update settings on Tower A and look to see if updated on Tower B"

going to check that on a cluster

kdelee:
Allright, tried that out and adding a test to catch in future. Going to close as verified

PR #7240, opened 2020-06-03
remove memcache entirely, add in some django redis to cover it's absence

rebeccahhh:
##### SUMMARY

pertains to #6932 
removing the remainders of memcache and sub in django-redis to work with Redis

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 11.2.0
```


installer/roles/local_docker/templates/docker-compose.yml.j2
```diff
@@ -7,7 +7,7 @@ services:
     container_name: awx_web
     depends_on:
       - redis
-      - memcached
+      - django-redis
```
shanemcd:
```suggestion
```

I think we can just delete this line. This isnt a python dependency, but rather a docker-compose dependency. The `redis` above should be all we need.

awx/settings/defaults.py
```diff
@@ -441,8 +441,8 @@ def IS_TESTING(argv=None):
 # Django Caching Configuration
 CACHES = {
     'default': {
-        'BACKEND': 'django.core.cache.backends.memcached.MemcachedCache',
-        'LOCATION': 'unix:/var/run/memcached/memcached.sock'
+        'BACKEND': 'django_redis.cache.RedisCache',
+        'LOCATION': 'unix:/var/run/redis/redis.sock'
```
chrismeyersfsu: Can we specify an "isolated" database i.e. `'LOCATION': 'unix:/var/run/redis/redis.sock?db=1'`

rebeccahhh: I can put it in and give it a go

ryanpetrello (commented):
Let's add a note here:

https://github.com/ansible/awx/blob/devel/CHANGELOG.md#1200-tbd

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ This is a list of high-level changes for each release of AWX. A full list of com
 - Fixed a bug that prevented PagerDuty notifications from sending for workflow job template approvals (https://github.com/ansible/awx/issues/7094)
 - Fixed a bug that broke external log aggregation support for URL paths that include the = character (such as the tokens for SumoLogic) (https://github.com/ansible/awx/issues/7139)
 - Fixed a bug that prevented organization admins from removing labels from workflow job templates (https://github.com/ansible/awx/pull/7143)
+- Removed remaining uses of memcache and replaced it with django-redis (https://github.com/ansible/awx/pull/7240)
```
ryanpetrello: Since we didn't just remove the `python-memcached` library, but also an entire container/dependency of AWX, I think this should go near (or at) the very top, and should say something more like:

`- Removed memcached as a dependency of AWX (https://github.com/ansible/awx/issues/6932)`

People have come to be familiar w/ a `memcached` just existing in installs, and I think we should call it out prominently so as to avoid confusion.

ryanpetrello (approved):

installer/roles/kubernetes/tasks/main.yml
```diff
@@ -246,6 +246,8 @@
 
 - name: Migrate database
   shell: |
+    {{ kubectl_or_oc }} -n {{ kubernetes_namespace }} exec ansible-tower-management -- \
+      bash -c "echo 'DJANGO_REDIS_IGNORE_EXCEPTIONS = True' > /etc/tower/conf.d/redis.py"
```
ryanpetrello: @shanemcd @rebeccahhh What's the reasoning for not *always* doing this?  It seems like the old python-memcached library always operated with silent cache misses, and that much of the code is still structured to assume this type of behavior.

i.e., one of the current CI failures on this PR looks like it's failing because it's invoking some `awx-manage` command that tries (and fails) to talk to (a non-existent) redis because we found ourselves in a `settings.__getattr__` code path at import time.

CHANGELOG.md
rebeccahhh: :+1: makes sense

installer/roles/kubernetes/tasks/main.yml
shanemcd: I just worry that it will break sometime and we'll have no visibility into it until a user or customer finds it.

shanemcd (approved):

PR #7176, opened 2020-05-27
Add support Prompt on Launch for Workflow Job Template

nixocio:
Add support Prompt on Launch for Workflow Job Template

see: https://github.com/ansible/awx/issues/5819

PR #7242, opened 2020-06-03
Modify JobTemplateForm to use formik `setValue`

nixocio:
Modify JobTemplateForm to use formik `setValue`. Following the same
convention used in the `.WorkflowJobTemplateForm`. Also, replace the
usage of `useFormikContext` to `useField`.


jlmitch5 (approved):
awesome, thanks!

nixocio:
See: https://github.com/ansible/awx/pull/7176

PR #7244, opened 2020-06-03
Remove unneeded auth props

jakemcdermott:
##### SUMMARY
These props don't do anything anymore.


nixocio:
@jakemcdermott, the solution using `ProtectedRoute` is really cool.

nixocio:
I got this code locally. It seems to be working fine.

nixocio (approved):

jakemcdermott:
> @jakemcdermott the solution using ProtectedRoute is really cool.

Thanks - this is mostly because of some great feedback from @AlexSCorey. The way we had it originally was a bit more complicated, but after staring at it for a while we realized some of the routing logic was redundant and we found a way to simplify it.

PR #6430, opened 2020-03-26
Named URL contextual filters in lookups made by AWX collection

AlanCoding:
##### SUMMARY
Fixes https://github.com/ansible/awx/issues/5193

Apply a generalized solution for dependency processing, because I believe that addresses the reason it wound up removed in 9955ee6548d0335d44cf18d446d032031fc955e5

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - AWX collection

##### AWX VERSION
```
9.3.0
```

##### ADDITIONAL INFORMATION
We can absolutely break out the changes to `resolve_name_to_id` in another change set. It is requisite for this feature, in that it needs to turn on and off the parameters to it. But that is pretty minor.

We shouldn't stop there either, `get_one` needs to get gobbled up in that, as well as a few other methods that duplicate logic.

PR #7245, opened 2020-06-04
tower_job_template to use organizations

sean-m-sullivan:
##### SUMMARY
Update tower_job_template to use organization similarly to how tower_workflow_job_template does. 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - Collection

##### AWX VERSION
```
awx: 11.2.0
```

##### ADDITIONAL INFORMATION
 - Based on the organization use in tower_workflow_job_template
 - Used to help lookup the project and by extension the job template
 - The Organization is inferred from the associated project
 - If not provided, will lookup by name only, which does not work with duplicates.

 Tested against:
 - Multiple projects with the same name and different organizations
 - Multiple job templates with the same name with different organizations with identically named projects and different inventories (To help tell them apart)
 - Tower 3.6.4
 - Tower 3.7.0
 - AWX 11.2.0

### DIFF
--- a/awx_collection/plugins/modules/tower_job_template.py
+++ b/awx_collection/plugins/modules/tower_job_template.py
@@ -53,7 +53,6 @@ options:
         - If not provided, will lookup by name only, which does not work with duplicates.
         - Requires Tower Version 3.7.0 or AWX 10.0.0 IS NOT backwards compatible with earlier versions.
       type: str
-      version_added: 2.10
     project:
       description:
         - Name of the project to use for the job template.
--- a/awx_collection/plugins/modules/tower_job_template.py
+++ b/awx_collection/plugins/modules/tower_job_template.py
@@ -51,6 +51,7 @@ options:
         - Used to help lookup the object, cannot be modified using this module.
         - The Organization is inferred from the associated project
         - If not provided, will lookup by name only, which does not work with duplicates.
+        - Requires Tower Version 3.7.0 or AWX 10.0.0 IS NOT backwards compatible with earlier versions.
       type: str
       version_added: 2.10
     project:
@@ -338,10 +339,6 @@ from ..module_utils.tower_api import TowerModule
 import json
 
 
-def versiontuple(v):
-    return tuple(map(int, (v.split("."))))
-
-
 def update_survey(module, last_request):
     spec_endpoint = last_request.get('related', {}).get('survey_spec')
     if module.params.get('survey_spec') == {}:
@@ -437,9 +434,7 @@ def main():
     organization = module.params.get('organization')
     if organization:
         organization_id = module.resolve_name_to_id('organizations', organization)
-        tower_version = module.get_endpoint('ping')['json']['version']
-        if versiontuple(tower_version) >= versiontuple("3.7.0"):
-            search_fields['organization'] = new_fields['organization'] = organization_id
+        search_fields['organization'] = new_fields['organization'] = organization_id
 
     # Attempt to look up an existing item based on the provided data
     existing_item = module.get_one('job_templates', **{'data': search_fields})
--- a/awx_collection/plugins/modules/tower_job_template.py
+++ b/awx_collection/plugins/modules/tower_job_template.py
@@ -45,6 +45,14 @@ options:
       description:
         - Name of the inventory to use for the job template.
       type: str
+    organization:
+      description:
+        - Organization the job template exists in.
+        - Used to help lookup the object, cannot be modified using this module.
+        - The Organization is inferred from the associated project
+        - If not provided, will lookup by name only, which does not work with duplicates.
+      type: str
+      version_added: 2.10
     project:
       description:
         - Name of the project to use for the job template.
@@ -299,6 +307,7 @@ EXAMPLES = '''
   tower_job_template:
     name: "Ping"
     job_type: "run"
+    organization: "Default"
     inventory: "Local"
     project: "Demo"
     playbook: "ping.yml"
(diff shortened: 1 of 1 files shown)


awx_collection/plugins/modules/tower_job_template.py
```diff
@@ -415,19 +429,26 @@ def main():
...
 
+    new_fields = {}
+    search_fields = {'name': name}
+
+    # Attempt to look up the related items the user specified (these will fail the module if not found)
+    organization_id = None
+    organization = module.params.get('organization')
+    if organization:
+        organization_id = module.resolve_name_to_id('organizations', organization)
+        tower_version = module.get_endpoint('ping')['json']['version']
```
AlanCoding: I'm generally hesitant to issue an extra request for this purpose, although, we've recently made some changes so that we get it in the response header. Maybe we could develop a shared method which is a performant way to obtain/compare version. @john-westcott-iv I'm all ears if you have ideas for that.

My alternative would be to simply not make this request. Always include `organization` in the JT search payload if given, and document as a backward-incompatible change. This can still be avoided by users with older server versions by not specifying the parameter. Anyway, if they are facing that problem, they aren't matching server-collection versions, which we discourage.

sean-m-sullivan: I'm all for not getting tower version, or just dumping this out and specifying it in docs. 
We could just rely on the user not to supply the org if they are prior to 3.7 and add a note in documentation. 
At the time I was trying to just make sure it would work against 3.7, 3.6 and awx. If that's the consensus, I'll remove the parts that the check version.

john-westcott-iv: We do pull the version out of the headers in module_util/tower_api.py around like 224. It should be fairly trivial to make properties in the object to hold these. On the other side, if we are going to have something that compares version we probably would want to make a method in the tower_api object because you would need to make it work for both tower an AWX.

john-westcott-iv:
Unless I am missing the intention of this PR; as a general comment on this, the design pattern we chose for the collection was to go by name or ID only (you can pass the ID in as the name filed and it will find the data). Adding other criteria to lookup secondary objects will simply overburden the modules' parameters. For example, the name of a credential may not be unique and tower_job_template can take multiple credentials so how would you provide additional parameters needed to lookup all the credentials?

Instead, the plan is to provide a tower lookup plugin which will take "arbitrary" criteria and lookup the IDs of objects within Tower. I believe @AlanCoding has started that work. Then, if your installation does not use unique names you would be able to do something akin to:
```
- name: Build a job template
  tower_job_template:
    name: "{{ lookup('tower', {'name': 'My Job Template', 'organization': 'Default'}) }}"
    project: "{{ lookup('tower', {'name': 'My Project', 'organization': 'Default'}) }}"
```

AlanCoding:
> Adding other criteria to lookup secondary objects will simply overburden the modules' parameters. For example, the name of a credential may not be unique and tower_job_template can take multiple credentials so how would you provide additional parameters needed to lookup all the credentials?

Using related fields for lookups will never be airtight, but it's the status quo.

I don't like keeping unqualified name lookups, because it's a land mine for the user experience and unsupportable. Creating a "localhost" host will probably work once, and will certainly fail given time. If inventory is a hierarchal object, then why are organizations not?

We can add the lookup plugin and start pushing people towards it, but I don't see a way around using lookup fields like this. Eventually it can be generalized. My [WIP for that](https://github.com/ansible/awx/pull/6430) tried to generalize it _too much_, but we can converge on something, and in the meantime I'm in favor of adding some cases manually like this.

sean-m-sullivan:
My purpose for this was that the people I work with are planning on going to one tower, and the one thing holding them back was that job_template names could not be the same even if in different orgs, due to no related org. 

I believe their idea is they would have a dev org, which would hold dev playbooks and testing, and a prod ord which has the prod push branch templates and workflows. 

So the entire purpose of this was to search the org when looking up the name. I looked at how workflows did it, and did that solution. 

Do we want to remove the version lookup for now, and keep it as is similar to the workflow. and then once we land on the lookup, remove the code from both?

AlanCoding:
> Do we want to remove the version lookup for now, and keep it as is similar to the workflow. and then once we land on the lookup, remove the code from both?

Yes to removing the version lookup. Also, add to the README a note that this is backward-incompatible. Also, consider mentioning the backward-incompatibility in the documentation for the parameter. Correct docs never hurt.

Yes, I think it should work similar to workflow.

I don't see this functionality being removed. It could be refactored for all similar modules, that's all I was saying.

sean-m-sullivan:
So I've redone the code, and tested it vs 3.6.4, 3.7.0 and awx, It works in all 3, the 3.6 it works if you don't include the org, and don't have duplicate projects with the same name. I've included the following line in the documentation for the parameter
```
Requires Tower Version 3.7.0 or AWX 10.0.0 IS NOT backwards compatible with earlier versions.
```
I'm not sure where to put another part in the readme as you suggested.

awx_collection/plugins/modules/tower_job_template.py
```diff
@@ -45,6 +45,14 @@
...
         - Name of the inventory to use for the job template.
       type: str
+    organization:
+      description:
+        - Organization the job template exists in.
+        - Used to help lookup the object, cannot be modified using this module.
+        - The Organization is inferred from the associated project
+        - If not provided, will lookup by name only, which does not work with duplicates.
+      type: str
+      version_added: 2.10
```
AlanCoding: this version_added won't be accepted, it's complicated, we have to do an ignore rule for it. Might as well leave it off for now.

AlanCoding:
@sean-m-sullivan I think that text would be great to put in the documentation for the parameter. I think it's arguable that this _isn't_ backward incompatible since it doesn't cause problems if you don't use the parameter, so I'm fine to not put it in the README.md.

sean-m-sullivan:
Updated to make requested changes Removed the version from the comment, added a note about backward compatibility in the parameter. Removed the version check from code.

AlanCoding (approved):

PR #7247, opened 2020-06-04
Allow us to specify external redis/memcached and added a replicas var…

ilijamt:
##### SUMMARY
During the installation on k8s we can now specify.

- External redis
- External memcached
- Replicas for the **awx** deployment, which currently defaults to 1

Enabling external redis and/or memcached will not deploy their pods on k8s.

After this next step would be to split awx web and task container so they can be scaled as needed, or at least the web container for high availability.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - Installer

##### ADDITIONAL INFORMATION

```
$ kubectl version                                 
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.1", GitCommit:"d224476cd0730baca2b6e357d144171ed74192d6", GitTreeState:"clean", BuildDate:"2020-01-14T21:04:32Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.8", GitCommit:"ec6eb119b81be488b030e849b9e64fda4caaf33c", GitTreeState:"clean", BuildDate:"2020-03-12T20:52:22Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
```

I prepare the environment before running the installation of k8s.

```
kubectl delete namespace awx
kubectl create namespace awx
helm install memcached bitnami/memcached -n awx
helm install redis bitnami/redis --set usePassword=false -n awx
```

Inventory file is 
```
localhost ansible_connection=local ansible_python_interpreter="/usr/bin/env python3"
[all:vars]
dockerhub_base=11.2.0
kubernetes_context=minikube
kubernetes_namespace=awx
pg_image_registry=docker.io
pg_serviceaccount=awx
kubernetes_ingress_hostname=awx.local
awx_task_hostname=awx
awx_web_hostname=awxweb
postgres_data_dir="~/.awx/pgdocker"
host_port=80
host_port_ssl=443
docker_compose_dir="~/.awx/awxcompose"
pg_username=awx
pg_password=awxpass
pg_database=awx
pg_port=5432
admin_user=admin
admin_password=password
create_preload_data=True
secret_key=awxsecret
redis_hostname=redis-master
memcached_host=memcached
memcached_port=11211
```

Tested against:
- AWX 11.2.0 on k8s 
- AWX devel on k8s

### DIFF
--- a/installer/inventory
+++ b/installer/inventory
@@ -100,8 +100,8 @@ pg_port=5432
 #redis_port=6379
 #redis_db=0
 
-# How many replicas to spin off the awx pod, it defaults to 1 replica
-# replicas=1
+# How many replicas to spin off the awx pod, it defaults to empty which spins up 1 replica
+# replicas=
 
 # The following variable is only required when using the provided
 # containerized postgres deployment on OpenShift
--- a/installer/inventory
+++ b/installer/inventory
@@ -100,11 +100,6 @@ pg_port=5432
 #redis_port=6379
 #redis_db=0
 
-# If we want to use an external memcached both memcached_host and memcached_port need to be set, 
-# this is only applicable to k8s deployment and will skip the creation of the memcached container.
-#memcached_host=example.com
-#memcached_port=11211
-
 # How many replicas to spin off the awx pod, it defaults to 1 replica
 # replicas=1
 
--- a/installer/inventory
+++ b/installer/inventory
@@ -91,6 +91,23 @@ pg_database=awx
 pg_port=5432
 #pg_sslmode=require
 
+# Set redis_hostname if you have an external redis server, this is only applicable to k8s deployment and
+# will skip the creation of the redis container.
+#redis_hostname=
+#redis_username=
+#redis_ssl=false
+#redis_password=
+#redis_port=6379
+#redis_db=0
+
+# If we want to use an external memcached both memcached_host and memcached_port need to be set, 
+# this is only applicable to k8s deployment and will skip the creation of the memcached container.
+#memcached_host=example.com
+#memcached_port=11211
+
+# How many replicas to spin off the awx pod, it defaults to 1 replica
+# replicas=1
+
 # The following variable is only required when using the provided
 # containerized postgres deployment on OpenShift
 # pg_admin_password=postgrespass
--- a/installer/roles/kubernetes/defaults/main.yml
+++ b/installer/roles/kubernetes/defaults/main.yml
@@ -57,8 +57,6 @@ container_groups_image: "ansible/ansible-runner"
 
 uwsgi_bash: "bash -c"
 
-replicas: 1
-
 redis_ssl: false
 redis_port: 6379
 redis_db: 0
--- a/installer/roles/kubernetes/defaults/main.yml
+++ b/installer/roles/kubernetes/defaults/main.yml
@@ -56,3 +56,9 @@ ca_trust_bundle: "/etc/pki/tls/certs/ca-bundle.crt"
 container_groups_image: "ansible/ansible-runner"
 
 uwsgi_bash: "bash -c"
+
+replicas: 1
+
+redis_ssl: false
+redis_port: 6379
+redis_db: 0
--- a/installer/roles/kubernetes/templates/configmap.yml.j2
+++ b/installer/roles/kubernetes/templates/configmap.yml.j2
@@ -200,15 +200,6 @@ data:
     BROADCAST_WEBSOCKET_PORT = 8052
     BROADCAST_WEBSOCKET_PROTOCOL = 'http'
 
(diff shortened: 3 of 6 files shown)


wenottingham:
See #7240

ilijamt:
Should I remove the changes for memcached ? @wenottingham

ryanpetrello (changes requested):
(edited for clarification)

Our usage of Redis in awx is intended to run as a locally-bound service because there's currently no support written in awx for authentication or TLS encryption of the message bus.  Changing this would result in setups that are insecure by default.  Doing this _properly_ out of the box across awx's many deployment setups and topologies would be a pretty notable burden on us, and I'm not confident we have the bandwidth on our team to pull it off.

It's also critical for performance reasons that the service by as low latency as possible (as opposed to people pointing at some offsite redis with high network latency), and it doesn't get lower latency than a locally bound unix socket.

Also, as @wenottingham suggested, memcached is going away soon.

ilijamt:
> Redis is intended to run as a locally-bound service because it currently has no authentication or TLS encryption. Changing this would result in setups that are insecure by default.
> It's also critical for performance reasons that the service by as low latency as possible, and it doesn't get lower latency than a locally bound unix socket.

@ryanpetrello 
You can setup Redis to be secure with authentication and SSL, which is possible to be setup, this is for setups that will require to have multiple web containers for one, or even task container. As for latency, it should still be quite low latency compared to RabbitMQ before, but it will be a bit slower than running over a unix connection. In my setup latency is <1ms for req-response between the container and redis. which should be quite sufficient.

* https://redis.io/topics/encryption
* https://redis.io/commands/auth

Currently if you try to scale the deployment every pod has it's own redis instance. If it remains like it is, we will not be able to scale the web or the task container. Or is there another way to do this?

When running with an external Redis instance, all the registered task containers are picking up jobs. And removing the Redis containers will save resources when we use and external service. 

> Also, as @wenottingham suggested, memcached is going away soon.

I will remove Memcached, from the PR

wenottingham:
> Currently if you try to scale the deployment every pod has it's own redis instance. If it remains like it is, we will not be able to scale the web or the task container. 

Why not? You'd scale web and task the normal way, by scaling the pod.

ilijamt:
> Why not? You'd scale web and task the normal way, by scaling the pod.

Depending on which container we hit for running the application, that's the task container that will execute the application, having the coupled like that just creates a waste of resources, like running multiple web containers for every task container when they are not needed, this way we can decouple the task and the web container, and only run based on our needs. But to be able to do so we need a shared Redis. I've run a setup like this with RabbitMQ on AWX 9.2.0 (upgraded from older versions) for over a year, separately scale-able Task and Web containers with an external RabbitMQ and Memcached for fact caching which now can be replaced with Redis. This setup (was modified) run for 1000+ hosts without breaking a sweat, and we had 8 web containers, and 20 task containers, this way all the container were picking up tasks, and we could scale depending on the load, by spinning up or down both web and/or the task containers. I know that there are container groups and they are quite promising but still there is a lot of issues when running larger loads like this (from my own testing/experimentation)

With the removal of Memcached, we will soon need a way to share the fact cache between the containers, and if we have Redis attached to each pod, if the task runs in a different container it will not use fact caching it will have to gather and store it in it's own Redis.

I've done preliminary tests and scaling the pod with your suggestion does not scale for bigger workloads for smaller you will not notice anything but in my case it was quite noticeable, it will run the tasks predominantly on the task container paired with the web container that executed it. Another thing I noticed when the deployment is scaled up, is that the logs were not flowing easily you would have to refresh the page, but with an external Redis the tasks are picked up by random task containers and the whole experience and there are no hiccups that I could see.

Regarding security and low latency, we can use authentication and TLS coupled with Elasticache (Redis) or even a self-hosted Redis 6+. In my tests latency was <1ms which is more than fast enough for this, I think we can live even with a bit more.  

Do you guys see some flaw that I'm not aware in this setup, that would require some more insight?
Maybe I'm missing something.

shanemcd:
I think I'm going to vote against this. There are no plans to move away from the redis sidecar container. If you find that the default resource requests aren't working for you, feel free to play with the `*_mem_request` and `*_cpu_request` variables.

Hounddog:
Well i just stumbled upon another problem.
We use docker swarm and have constraints to run only the task on the manager node.
Even if i set the constraints to let Task/Web and Redis run on Manager Nodes i cannot guarantee that they will run on the same manager node. Having the redis host would enable us to keep running AWX in Docker Swarm. This is however now very limited

ilijamt:
I used to run AWX in Docker Swarm, you can make it run but will require some modifications to the install procedure and settings.py configuration that gets put into the container.

PR #7250, opened 2020-06-04
Remove style overrides from Lookup component

### LINKED ISSUE #6810: Remove style overrides from lookup inputs
##### ISSUE TYPE
 - Task

##### SUMMARY
We should remove the style overrides from the lookup component in favor of using default PatternFly styles.

https://www.patternfly.org/v4/documentation/react/components/inputgroup#basic

Screen Shot 2020-04-22 at 6 37 56 PM

It should look like this screenshot, except with the search icon to the left.

nixocio:
Remove style overrides from Lookup component

closes: https://github.com/ansible/awx/issues/6810

### DIFF
--- a/awx/ui_next/src/components/Lookup/Lookup.jsx
+++ b/awx/ui_next/src/components/Lookup/Lookup.jsx
@@ -14,7 +14,7 @@ import {
   Button,
   ButtonVariant,
   Chip,
-  InputGroup as PFInputGroup,
+  InputGroup,
   Modal,
 } from '@patternfly/react-core';
 import { withI18n } from '@lingui/react';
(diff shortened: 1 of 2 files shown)


awx/ui_next/src/components/Lookup/Lookup.jsx
```diff
@@ -33,15 +33,6 @@ const SearchButton = styled(Button)`
...
-const InputGroup = styled(PFInputGroup)`
-  ${props =>
-    props.multiple &&
-    `
-    --pf-c-form-control--Height: 90px;
-    overflow-y: auto;
-  `}
-`;
-
 const ChipHolder = styled.div`
```
jlmitch5, line 28: this needs to be removed as well

jlmitch5 (changes requested):
@nixocio this is more than just straight removing the overrides there...we need to make sure this still looks correct in the flow.  see the screenshot marly posted in the issue.

I know we'll need to fix the border for the custom ChipHolder.  (EDIT: I thought we were moving the icon from the left to the right, I read marly's issue to fast.  removed that part of this comment)

please post a screenshot in the description of the issue when tackling issues like this so we can inspect them when we review.

marshmalien:
https://github.com/ansible/awx/pull/7250/files#diff-7dd80f25de7c0280e8e3499a63d32242R118

We should update the search button variant to be `variant="control"`. That will get us the styles we are looking for. You can see how PatternFly uses the button variant in theses examples: https://www.patternfly.org/v4/documentation/react/components/inputgroup#with-multiple-group-siblings

I agree with @jlmitch5, we could remove the styles from search button, input group, and chip holder. We _might_ need to keep the chip holder height style to prevent the input group from squishing the chips.

nixocio:
@marshmalien, @jlmitch5 updated per comments. Thanks for your review.

lookup

marshmalien (approved):
🎉

awx/ui_next/src/components/Lookup/Lookup.jsx
nixocio: This variable ` --pf-c-form-control--Height: auto;` needs to be kept.

nixocio:
hey @jlmitch5, I removed the label `state:needs_review `, and later realized that I can not merge this one without your approval. Any more comments?

jlmitch5 (approved):
looks great!

PR #7257, opened 2020-06-04
Force worker processes to have a different signal handler from the parent

jbradberry:
##### SUMMARY

Situations have come up where the 5+ minute kill signal for
run_task_manager is emitted to the worker process running it, but
since the worker improperly inherited the AWXConsumerBase().stop()
handler a deadlock ultimately was triggered on the database
connection.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
awx: 11.2.0

### DIFF
--- a/awx/main/dispatch/worker/base.py
+++ b/awx/main/dispatch/worker/base.py
@@ -35,6 +35,7 @@ class WorkerSignalHandler:
 
     def __init__(self):
         self.kill_now = False
+        signal.signal(signal.SIGTERM, signal.SIG_DFL)
         signal.signal(signal.SIGINT, self.exit_gracefully)
 
     def exit_gracefully(self, *args, **kwargs):
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -581,3 +581,4 @@ class TaskManager():
                 logger.debug("Starting Scheduler")
                 with task_manager_bulk_reschedule():
                     self._schedule()
+                logger.debug("Finishing Scheduler")

### CODE REVIEW

File: awx/main/scheduler/task_manager.py [hunk 1]
```diff
@@ -576,8 +576,9 @@ def schedule(self):
         with advisory_lock('task_manager_lock', wait=False) as acquired:
             with transaction.atomic():
                 if acquired is False:
-                    logger.debug("Not running scheduler, another task holds lock")
+                    logger.warning("Not running scheduler, another task holds lock")
```
chrismeyersfsu: I imagine this happens a _lot_, and we expect that. I'd keep this lower than warning.

awx/main/scheduler/task_manager.py
```diff
@@ -576,8 +576,9 @@ def schedule(self):
...
                     return
                 logger.debug("Starting Scheduler")
                 with task_manager_bulk_reschedule():
                     self._schedule()
+                logger.debug("Finishing Scheduler")
```
chrismeyersfsu, line 584: Yes please. You can _probably_ use the log timestamp between the start and finish debug message, but would be nice to explicitly print the time taken.

awx/main/dispatch/worker/base.py
```diff
@@ -35,6 +35,7 @@ class WorkerSignalHandler:
 
     def __init__(self):
         self.kill_now = False
+        signal.signal(signal.SIGTERM, self.exit_gracefully)
```
chrismeyersfsu: :+1:

awx/main/scheduler/task_manager.py
ryanpetrello: Yep, it's probably not uncommon for multi-node setups that all run this on a periodic schedule to often race eachother 👍 It's _normal_ for only one to win.

jbradberry: sigh, ok, changed back.

ryanpetrello (approved):

awx/main/dispatch/worker/base.py
```diff
@@ -35,6 +35,7 @@ class WorkerSignalHandler:
 
     def __init__(self):
         self.kill_now = False
+        signal.signal(signal.SIGTERM, signal.SIG_DFL)
```
ryanpetrello, line 38: I like this much better than what I originally suggested about detecting the inherited signal handler from the parent process.

jbradberry:
regate

jbradberry:
mergeit

Issue #7256, opened 2020-06-04
Implement `Replace` functionality for credential passwords


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
<img width="517" alt="Screen Shot 2020-06-04 at 2 43 46 PM" src="https://user-images.githubusercontent.com/9889020/83798242-f9c25800-a671-11ea-8ab8-58b0583e82b7.png">

In the old UI, password fields in the credential subform are disabled on edit and we force the user to click this replace button in order to enable it and make a change.

@wenottingham @trahman73 is this a UX pattern that we want to carry over to the new app?  Right now the field looks like:

<img width="473" alt="Screen Shot 2020-06-04 at 2 48 10 PM" src="https://user-images.githubusercontent.com/9889020/83798579-75240980-a672-11ea-91fb-bb62d7743ead.png">

and the user can type in the input to overwrite:

<img width="482" alt="Screen Shot 2020-06-04 at 2 48 18 PM" src="https://user-images.githubusercontent.com/9889020/83798605-7ead7180-a672-11ea-8d79-b01312e2fe3c.png">

If we do want it - is it critical for an initial release?

mabashian:
Actually @jakemcdermott brought up a good point - maybe this button is used to remove a value from the field if necessary.  Not all the password fields are required so there is a use case for needing to remove a password but not replace it.

trahman73:
After our internal discussion, it seems we would want to carry over the same UX. Just want to make sure we have tooltips with explanation over the action icons. @wenottingham can let us know if this needs to be part of the initial release.

mabashian:
OK so after looking at the old UI, I think this is how the button works.

1) When the page is initially loaded and the user hasn't interacted with the field the tooltip on the button says `Replace`.
2) Clicking the button when it says `Replace` sets the value of the field to an empty string
3) The button now says `Revert`
4) Clicking the button when it says `Revert` sets the value of the field back to it's initial value (which I think is always `$encrypted$`

Note that this button should only appear when _editing_ the credential.  It should not appear when adding.

tiagodread:
We had a failure related to this change, which I'll be working on before we close the issue

tiagodread:
Fixed, closing

Issue #7127, opened 2020-05-22
Update credential details to display fields configured using cred plugins


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Credential details needs to be updated to display fields that are configured to be pulled from an external source (credential plugins).  I'm not sure that the UX has been worked out for this but we'd want to show:

1) Field name
2) Selected credential
3) Metadata (this could be multiple fields)

To set this scenario up you'd first need to create a credential and configure one of the fields to pull from an external secret store.

I'm marking this as blocked until https://github.com/ansible/awx/pull/7125 gets merged

unlikelyzero:
need to test

unlikelyzero:
Needs coverage

unlikelyzero:
@akus062381 there should be enough in our existing codebase to quickly add a test to verify the Credentials/Details page's use of External Credential references

Also, need to add a visual test or visual test stub

akus062381:
Manually tested.
Testability needed - see https://github.com/ansible/awx/issues/9687
Stubs added for coverage- see https://github.com/ansible/tower-qa/pull/6124

Issue #7126, opened 2020-05-22
Hook up credential plugin wizard Test button


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
<img width="1362" alt="Screen Shot 2020-05-22 at 9 50 07 AM" src="https://user-images.githubusercontent.com/9889020/82674720-bd264380-9c11-11ea-8310-4d238d5ddee4.png">

Clicking this button should POST to `/api/v2/credentials/:id/test/`.  We currently show the results in a toast message like so:

<img width="1239" alt="Screen Shot 2020-05-22 at 9 53 33 AM" src="https://user-images.githubusercontent.com/9889020/82675006-1ee6ad80-9c12-11ea-9d09-08ae2b2275de.png">

We could do that or we could show the results under the form.

Marking as blocked until https://github.com/ansible/awx/pull/7125 merges

unlikelyzero:
Still working on this. Blocked on akit commands.

unlikelyzero:
covered in credentials-mock.spec.js

Issue #6703, opened 2020-04-14
Add support for dropping a file on key fields so users don't copy/paste


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
This feature exists in the old app and we'll want to add support for it here as well.  Fields like `SSH Private Key` and `RSA Private Key` had logic built in to accept file drops in order to populate the field.

It looks like PF is using react-dropzone under the hood for it's file upload/drop functionality.  We could look in to pulling something like that in to achieve this.

Alternatively, maybe this is something that PF can help us out with.  Being able to import some kind of `Droppable` component that can wrap around an element would solve our problem here without having to pull in the dependency ourselves.

There may be other fields where this functionality is necessary - I just know it's prevalent on the credential form(s).

mabashian:
https://www.patternfly.org/v4/documentation/react/components/fileupload/#text-file-with-edits-allowed is already available via PF and _should_ be sufficient for our needs.  This UX would actually be better than the old app because users could drop or select their file from the file picker.

mabashian:
This can probably be combined with https://github.com/ansible/awx/issues/6113

PR #7258, opened 2020-06-04
Dynamically render credential subform fields

mabashian:
##### SUMMARY
Link #5879 

One of the first decisions I made was to actually not dynamically generate the _whole_ form but just the type details subform area.  All credentials have a Name/Description/Organization field so I felt like it was OK to leave that area of the form alone and just focus on the subform.  There are a number of new/updated components and I'm going to try to run through them and explain what's going on in each.

- `CredentialForm.jsx`: A couple of big changes in here.  Firstly, I found it easier to go ahead and initialize all the possible field values right off the bat rather than trying to do this on the fly as the credential type changed.  We'll see corresponding logic in the add/edit components that filters values down to only fields that belong to the selected credential type before sending the POST/PUT.  You'll also see an extension of the reset logic.  I tried to line this behavior up with the old application.  The basic gist is: if you change credential types it resets all the subform fields to their default value as provided by the API.  The one exception to this is when you switch back to the saved credential type (on edit).  In this case we'll revert the fields back to their _saved_ values.  The last change in here has to do with the subform - instead of having a big conditional where we include different subform components based on credential type we now render a `CredentialSubForm` component that handles the rendering of all of the relevant fields (when a cred type is selected).
- `CredentialSubForm.jsx`: This splits the fields into two buckets (bool & non-bool) then loops across each of those arrays.  For bool fields, they all get lumped together into a single OPTIONS `FormGroup`.  For non-bool fields, a `CredentialField` is rendered.
- `CredentialField.jsx`: This component introspects the field options and decides how it should be rendered.  For external cred types (like CyberArk and Hashi) we don't want to render the credential plugin fields to we render those as normal `FormGroup`'s.  If the field has `choices` then we know it should be a dropdown and render it appropriately.  By default, most fields are going to be rendering the `CredentialPluginField`.
- `CredentialPluginField.jsx`: This component handles all the plugin related logic and wraps the input in an InputGroup in order to add that `key` button.
- `BecomeMethodField.jsx`: This field needed special treatment because we maintain the array of default options on the client (not available via the API).
- `GceFileUploadField.jsx`: This is not actually a field but rather a shortcut users can use to populate the rest of the GCE subform.
- `CredentialAdd.jsx/CredentialEdit.jsx`: The big change here has to do with bundling up the payload that we send to `/api/v2/credentials`.  We now use the credential type OPTIONS response to filter out input fields that don't apply to the selected credential type.  There's also added logic for handling `Prompt on launch` inputs.  This differs from `Prompt on launch` fields in the JT form in that they don't have their own field on the model.  So, to enable prompting on a field we have to pass the `ASK` string in the payload.

**Outstanding issues (things that are not included in this PR):**
- `Replace` functionality on password fields: https://github.com/ansible/awx/issues/7256
- Update credential details to handle plugin input fields: https://github.com/ansible/awx/issues/7127
- Hook up credential plugin `Test` button: https://github.com/ansible/awx/issues/7126
- Drag and drop on key fields: https://github.com/ansible/awx/issues/6703
- Add file picker support for key fields: https://github.com/ansible/awx/issues/6113

Once we're settled on the approach I'll certainly add more test coverage to the form(s) but wanted to avoid unnecessary work if we decided to go in a different direction.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI


awx/ui_next/src/screens/Credential/shared/CredentialSubForm.jsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react';
```
jlmitch5: maybe we should name this file CredentialTypeSubForm?

awx/ui_next/src/screens/Credential/CredentialAdd/CredentialAdd.jsx
```diff
@@ -44,18 +59,17 @@ function CredentialAdd({ me }) {
           inputs: nonPluginInputs,
           ...remainingValues,
         });
-        const inputSourceRequests = [];
-        Object.entries(pluginInputs).forEach(([key, value]) => {
-          inputSourceRequests.push(
+
+        await Promise.all(
```
jlmitch5, line 64: at first i was thinking it'd be better to concat plugin/non-plugin together, but I think the way useCallback and await works...it'll still finish all this before returning that it's finished

awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx
```diff
@@ -3,54 +3,73 @@ import { Formik, useField } from 'formik';
...
+    .map(key => {
+      return {
+        value: credentialTypes[key].id,
+        key: credentialTypes[key].id,
+        label: credentialTypes[key].name,
+      };
+    })
+    .sort((a, b) => (a.label > b.label ? 1 : -1));
+
+  const resetSubFormFields = (newCredentialType, form) => {
```
jlmitch5, line 37: would this be better in "CredentialSubForm"?

awx/ui_next/src/screens/Credential/shared/CredentialFormFields/CredentialField.jsx
```diff
@@ -0,0 +1,167 @@
...
+};
+
+function CredentialField({ credentialType, fieldOptions, i18n }) {
+  const { values: formikValues } = useFormikContext();
+  const requiredFields = credentialType?.inputs?.required || [];
+  const isRequired = requiredFields.includes(fieldOptions.id);
+  const [subFormField, meta, helpers] = useField({
+    name: `inputs.${fieldOptions.id}`,
+    validate:
+      isRequired && !formikValues?.passwordPrompts[fieldOptions.id]
```
jlmitch5: can we unroll this from the ternaries and && into a function?

jlmitch5 (commented):
this is looking great!  just a couple things hear and there that might could be tweak, but i love the direction!

awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx
```diff
@@ -3,54 +3,73 @@ import { Formik, useField } from 'formik';
...
+    credentialTypes[newCredentialType].inputs.fields.forEach(
+      ({ ask_at_runtime, type, id, choices, default: defaultValue }) => {
```
AlexSCorey, line 40: I think formik's resetForm function can help with some of the complicated logic here. check this out https://github.com/ansible/awx/blob/devel/awx/ui_next/src/screens/Inventory/shared/InventorySourceForm.jsx#L28

If I had my way I would get the api to tell me whether the cred types are custom or not when we fetch them from the api.  Then, when in edit mode, I would only show those types in the drop down, and if there are no custom creds, the drop down is disabled.

awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx
```diff
@@ -135,80 +147,54 @@ function CredentialForm({
...
-      ssh_key_unlock: credential?.inputs?.ssh_key_unlock || '',
-      ssh_public_key_data: credential?.inputs?.ssh_public_key_data || '',
-      username: credential?.inputs?.username || '',
-    },
+    inputs: {},
+    passwordPrompts: {},
   };
 
+  Object.values(credentialTypes).forEach(credentialType => {
+    credentialType.inputs.fields.forEach(
```
AlexSCorey: I created a custom cred type in the old ui that did not have any input configuration or injector configuration and it broke this line of code.  I realize that a user creating such a Cred Type is unlikely, however, we should be prepared for it.

awx/ui_next/src/screens/Credential/shared/CredentialFormFields/CredentialField.jsx
AlexSCorey: +1

AlexSCorey (changes requested):
- [ ] We definitely need to figure out the bug mentioned below with a custom cred that has no input configuration.  

- [ ] I think we also need to re-think those nested ternaries as they can be hard to follow.

- [ ] As far as the reset function goes, I think I can be cleaned up a bit, especially the first if statement and I'd be happy to chat with you further about that if necessary.

awx/ui_next/src/screens/Credential/shared/CredentialSubForm.jsx
mabashian: After chatting with @jlmitch5 about this one I think we've settled on `TypeInputsSubForm`.  If anyone else has ideas for this component name then let us know.

awx/ui_next/src/screens/Credential/shared/CredentialFormFields/CredentialField.jsx
mabashian: Sure, np

awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx
jakemcdermott: Credential types have a `managed_by_tower` field. `false` means custom.

I think this drop-down would show _both_ custom and non-custom credentials, though. For example, I'd expect to be able to select the machine credential type or gce type even though both are managed (non-custom).

awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx
mabashian: 👍 will def fix this

awx/ui_next/src/screens/Credential/shared/CredentialFormFields/CredentialField.jsx
mabashian: Talked with @jlmitch5 about this one and I think we came up with more readable solution.

awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx
```diff
@@ -34,45 +29,6 @@ function CredentialFormFields({
     })
     .sort((a, b) => (a.label > b.label ? 1 : -1));
 
-  const resetSubFormFields = (newCredentialType, form) => {
```
AlexSCorey: https://media.giphy.com/media/Is1O1TWV0LEJi/giphy.gif

awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx
```diff
@@ -120,8 +76,16 @@ function CredentialFormFields({
             ...credentialTypeOptions,
           ]}
           onChange={(event, value) => {
-            credTypeHelpers.setValue(value);
-            resetSubFormFields(value, formik);
+            const { values, initialValues, resetForm } = formik;
```
AlexSCorey: wow, awesome!

awx/ui_next/src/screens/Credential/shared/CredentialFormFields/CredentialField.jsx
```diff
@@ -76,19 +76,20 @@ function CredentialField({ credentialType, fieldOptions, i18n }) {
   const { values: formikValues } = useFormikContext();
   const requiredFields = credentialType?.inputs?.required || [];
   const isRequired = requiredFields.includes(fieldOptions.id);
+  const validateField = () => {
```
AlexSCorey, line 79: 👍

AlexSCorey (approved):
I think this looks really good.

jlmitch5 (approved):
great stuff!

jakemcdermott (approved):

dsesami:
@mabashian what do you want the general rules to be for this for form persistence? it's super weird currently, so I'd like to get some defined behavior down.

1. Make some AWS creds, but don't save. Switch to something else and then switch back to AWS.
- The AWS creds are cleared out.

2. Make some cyberark creds and then save. Go back and edit.
- If you switch back to AWS, your previously existing fields are there! Well, not all of them. I had to go back and forth saving a cyberark and aws cred to do this, but things which i suspect use the same "key" are persisted between forms when editing, like "username".

dsesami:
followed up with @mabashian. Once we iron out the persistence in forms mentioned above + the weird issue of "Vault" appearing after the alphabetically-latter "VMware" this should be good for the purposes of this PR.

dsesami (approved):
barring the formik stuff I discussed with @mabashian which is in a different PR, this is good to go.

PR #7261, opened 2020-06-04
Add queue / instance group registration to heartbeat for k8s installs

shanemcd:
There is some history here.

https://github.com/ansible/awx/pull/7190 <- This PR was an attempt at fixing a bug notting ran into where some jobs on k8s installs would get stuck in Waiting forever.

The PR mentioned above introduced a bug where there are no instance groups on a fresh k8s-based install. This is because this process currently happens in the launch scripts, before the database is up.

With this patch, queue / instance group registration happens in the heartbeat, right after auto-registering the instance.

### DIFF
--- a/awx/main/management/commands/register_queue.py
+++ b/awx/main/management/commands/register_queue.py
@@ -16,31 +16,24 @@ class InstanceNotFound(Exception):
         super(InstanceNotFound, self).__init__(*args, **kwargs)
 
 
-class Command(BaseCommand):
+class RegisterQueue:
+    def __init__(self, queuename, controller, instance_percent, inst_min, hostname_list):
+        self.instance_not_found_err = None
+        self.queuename = queuename
+        self.controller = controller
+        self.instance_percent = instance_percent
+        self.instance_min = inst_min
+        self.hostname_list = hostname_list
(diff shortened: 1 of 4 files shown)


matburt (approved):

awx/main/management/commands/register_queue.py
```diff
@@ -16,44 +15,39 @@ def __init__(self, message, changed, *args, **kwargs):
...
-
-    def get_create_update_instance_group(self, queuename, instance_percent, instance_min):
+    def get_create_update_instance_group(self):
         created = False
         changed = False
-
-        (ig, created) = InstanceGroup.objects.get_or_create(name=queuename)
-        if ig.policy_instance_percentage != instance_percent:
-            ig.policy_instance_percentage = instance_percent
+        from awx.main.models import InstanceGroup
```
chrismeyersfsu: Don't love this. But what's the alternative. I think we can move `from awx.main.management.commands.register_queue import RegisterQueue` to the method it's called and return the import here to the top.

shanemcd: Sure. Will give that a shot.

PR #7262, opened 2020-06-04
remove LGPL source for dependencies we've since removed

ryanpetrello:


wenottingham (approved):
We should probably fix `awx/main/tests/functional/test_licenses.py` to catch this case, but yep.

ghjm (approved):

PR #7269, opened 2020-06-05
Use actual module parameters for inventory plugin

AlanCoding:
##### SUMMARY
As a followup to https://github.com/ansible/awx/pull/7216, this changes the _primary_ name for the parameters for the inventory plugin in the AWX collection to match what the modules will take.

This is a big change for the documented interface, so it may be best to delay until the next AWX release.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
11.2.0
```


AlanCoding:
@john-westcott-iv this could be absorbed into https://github.com/ansible/awx/pull/7286, or done afterwards. Since things are moving I'm closing this as something which could be revived later.

Issue #7015, opened 2020-05-13
JT Launch prompts: Credential Passwords step


keithjgrant:
Follow-up for #5909 

##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
This is the final step that needs to be coded up for the JT Launch prompt. Criteria:

- This should be the third step, immediately following the Credentials step
- This step can be shown in one of two situations:
    - Firstly, it should be shown if passwords_needed_to_start is _not_ an empty array _and_ ask_credential_on_launch is set to false.
    - Secondly, it should be shown if ask_credential_on_launch is set to true _and_ any of the selected credentials from that step prompt for passwords.
- This step is dynamic and should be able to support any number of password fields (the JT could have n number of different vault credentials all prompting for a password).
~~- Previously we showed these password fields on the same step as the credentials.  If it's easier to do that here as well then we can explore that.~~
~~- User cannot proceed to the next step until all required passwords are provided~~

mabashian:
Moving this to blocked while waiting for prompt on launch refactors

mabashian:
A good portion of this work is already completed here: https://github.com/mabashian/awx/tree/7015-prompt-cred-passwords but this branch includes passwords with the credential step.  If we want to split them out into 2 steps then this will need to be refactored (in addition to refactoring after POL changes).

unlikelyzero:
https://github.com/ansible/awx/pull/7281#issuecomment-691218496 note to cover

one-t:
Just want to check, are we expecting this PR to add prompting for schedules or will that be covered in another one? Currently that functionality does not appear to be implemented.

mabashian:
@one-t good question.  Schedule prompting will be handled right after this PR merges.  https://github.com/ansible/awx/issues/7692

unlikelyzero:
@one-t in this case, we will put a stubbed out test in the schedules folder with a //TBI link to where the test will be implemented

unlikelyzero:
https://github.com/ansible/awx/pull/9028#pullrequestreview-571571287

PR #7281, opened 2020-06-08
Updates credential details to support plugin and prompt fields

mabashian:
##### SUMMARY
Link #7127 

This PR accomplishes two things:

1) Support display of prompt on launch password fields in the Details view:
Screen Shot 2020-06-08 at 11 00 58 AM

- Previously we were showing the `ASK` text that the API returns.  I think this is a little bit more informative and it matches the pattern that we have on the JT details view.

2) Support display of plugin configured fields in the Details view:
Screen Shot 2020-06-08 at 11 01 29 AM

- Since we added support for credential plugins with https://github.com/ansible/awx/pull/7125 we needed to update the details view to display that plugin configuration.  Configuration consists of two things: a credential and some metadata.  After talking with @trahman73 we decided that the path of least resistance was to display the metadata as JSON in a codemirror underneath the field.  This way, a user can see both the credential and the metadata used when pulling this field from an external source.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/api/models/Credentials.js
+++ b/awx/ui_next/src/api/models/Credentials.js
@@ -35,7 +35,7 @@ class Credentials extends Base {
             },
           }
         );
-        if (data.next && requestCounter <= maxRequests) {
+        if (data?.next && requestCounter <= maxRequests) {
           return fetchInputSources(
             pageNo + 1,
             inputSources.concat(data.results)
--- a/awx/ui_next/src/api/models/Credentials.js
+++ b/awx/ui_next/src/api/models/Credentials.js
@@ -21,8 +21,11 @@ class Credentials extends Base {
   }
 
   readInputSources(id) {
+    const maxRequests = 5;
+    let requestCounter = 0;
     const fetchInputSources = async (pageNo = 1, inputSources = []) => {
       try {
+        requestCounter++;
         const { data } = await this.http.get(
           `${this.baseUrl}${id}/input_sources/`,
           {
@@ -32,7 +35,7 @@ class Credentials extends Base {
             },
           }
         );
-        if (data.next) {
+        if (data.next && requestCounter <= maxRequests) {
           return fetchInputSources(
             pageNo + 1,
             inputSources.concat(data.results)
--- a/awx/ui_next/src/api/models/Credentials.js
+++ b/awx/ui_next/src/api/models/Credentials.js
@@ -20,10 +20,31 @@ class Credentials extends Base {
     return this.http.options(`${this.baseUrl}${id}/access_list/`);
   }
 
-  readInputSources(id, params) {
-    return this.http.get(`${this.baseUrl}${id}/input_sources/`, {
-      params,
-    });
+  readInputSources(id) {
+    const fetchInputSources = async (pageNo = 1, inputSources = []) => {
+      try {
+        const { data } = await this.http.get(
+          `${this.baseUrl}${id}/input_sources/`,
+          {
+            params: {
+              page: pageNo,
+              page_size: 200,
+            },
+          }
+        );
+        if (data.next) {
+          return fetchInputSources(
+            pageNo + 1,
+            inputSources.concat(data.results)
+          );
+        }
+        return Promise.resolve(inputSources.concat(data.results));
+      } catch (error) {
+        return Promise.reject(error);
+      }
+    };
+
+    return fetchInputSources();
   }
 
   test(id, data) {
--- a/awx/ui_next/src/api/models/Credentials.js
+++ b/awx/ui_next/src/api/models/Credentials.js
@@ -41,7 +41,11 @@ class Credentials extends Base {
             inputSources.concat(data.results)
           );
         }
-        return Promise.resolve(inputSources.concat(data.results));
+        return Promise.resolve({
+          data: {
+            results: inputSources.concat(data.results),
+          },
+        });
       } catch (error) {
         return Promise.reject(error);
       }
--- a/awx/ui_next/src/screens/Credential/CredentialDetail/CredentialDetail.jsx
+++ b/awx/ui_next/src/screens/Credential/CredentialDetail/CredentialDetail.jsx
@@ -238,7 +238,9 @@ function CredentialDetail({ i18n, credential }) {
       </DetailList>
       {Object.keys(inputSources).length > 0 && (
         <PluginFieldText>
-          {i18n._(t`* This field will be retrieved from an external secret management system using the specified credential.`)}
+          {i18n._(
+            t`* This field will be retrieved from an external secret management system using the specified credential.`
+          )}
         </PluginFieldText>
       )}
       <CardActionsRow>
--- a/awx/ui_next/src/screens/Credential/CredentialDetail/CredentialDetail.jsx
+++ b/awx/ui_next/src/screens/Credential/CredentialDetail/CredentialDetail.jsx
@@ -60,7 +60,9 @@ function CredentialDetail({ i18n, credential }) {
         {
           data: { inputs: credentialTypeInputs, managed_by_tower },
         },
-        loadedInputSources,
+        {
+          data: { results: loadedInputSources },
+        },
       ] = await Promise.all([
         CredentialTypesAPI.readDetail(credential_type.id),
         CredentialsAPI.readInputSources(credentialId),
--- a/awx/ui_next/src/screens/Credential/CredentialDetail/CredentialDetail.jsx
+++ b/awx/ui_next/src/screens/Credential/CredentialDetail/CredentialDetail.jsx
@@ -64,12 +64,10 @@ function CredentialDetail({ i18n, credential }) {
           {
             data: { inputs: credentialTypeInputs, managed_by_tower },
           },
-          {
-            data: { results: loadedInputSources },
-          },
+          loadedInputSources,
         ] = await Promise.all([
           CredentialTypesAPI.readDetail(credential_type.id),
-          CredentialsAPI.readInputSources(credentialId, { page_size: 200 }),
+          CredentialsAPI.readInputSources(credentialId),
         ]);
 
         setFields(credentialTypeInputs.fields || []);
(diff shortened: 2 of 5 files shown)


awx/ui_next/src/screens/Credential/CredentialDetail/CredentialDetail.jsx
```diff
@@ -1,25 +1,36 @@
...
 } from '../../../components/DetailList';
+import ChipGroup from '../../../components/ChipGroup';
+import CodeMirrorInput from '../../../components/CodeMirrorInput';
+import CredentialChip from '../../../components/CredentialChip';
 import ErrorDetail from '../../../components/ErrorDetail';
 import { CredentialsAPI, CredentialTypesAPI } from '../../../api';
 import { Credential } from '../../../types';
 import useRequest, { useDismissableError } from '../../../util/useRequest';
 
+const PluginInputMetadata = styled(CodeMirrorInput)`
```
jlmitch5, line 26: aren't codemirrors already full-length?

awx/ui_next/src/screens/Credential/CredentialDetail/CredentialDetail.jsx
```diff
@@ -48,19 +60,33 @@ function CredentialDetail({ i18n, credential }) {
...
+        const [
+          {
+            data: { inputs: credentialTypeInputs, managed_by_tower },
+          },
+          {
+            data: { results: loadedInputSources },
+          },
+        ] = await Promise.all([
+          CredentialTypesAPI.readDetail(credential_type.id),
+          CredentialsAPI.readInputSources(credentialId, { page_size: 200 }),
```
jlmitch5: since we keep these functions in our api file now, we could do the recursive to last page thing there...would be a bit more elegant than keeping it in the views like we had to do in the old app.  but i'm also fine for keeping things simple like this.

jlmitch5 (commented):
looks good, just a couple questions

awx/ui_next/src/screens/Credential/CredentialDetail/CredentialDetail.jsx
mabashian: If we don't make it take up all the columns of the grid then it ends up looking like:

Screen Shot 2020-06-09 at 4 55 13 PM

awx/ui_next/src/screens/Credential/CredentialDetail/CredentialDetail.jsx
mabashian: @jlmitch5 see if https://github.com/ansible/awx/pull/7281/commits/834cfb82f8da49003633743599640d405c92cf47 is what you're looking for.  I like this pattern for requests that we know will be made recursively.

jakemcdermott: I'm ok with that approach if the recursion is capped to something reasonable and is guaranteed to terminate regardless of whether the server sends us correct data. A max request limit and counter will suffice.

awx/ui_next/src/screens/Credential/CredentialDetail/CredentialDetail.jsx
```diff
@@ -48,19 +60,31 @@ function CredentialDetail({ i18n, credential }) {
       setContentError(null);
       setHasContentLoading(true);
       try {
-        const {
```
mabashian, line 51: Reminder: useRequest

awx/ui_next/src/screens/Credential/CredentialDetail/CredentialDetail.jsx
mabashian: Sounds good @jakemcdermott - I added that logic and capped it at 5 requests.  The likelihood of more than 1000 input sources on a cred seems very low.

mabashian: We may decide that we want to build some kind of generic wrapper around this sort of logic so that we can recursively fetch from other endpoints without having to duplicate a lot of this.

awx/ui_next/src/screens/Credential/CredentialDetail/CredentialDetail.jsx
mabashian: @marshmalien - added this in.  Thanks for the snippet!

awx/ui_next/src/api/models/Credentials.js
```diff
@@ -16,8 +16,11 @@ class Credentials extends Base {
   }
 
   readInputSources(id) {
+    const maxRequests = 5;
```
AlexSCorey, line 24: At first this function was a bit confusing, but after looking at it again I like it.

awx/ui_next/src/screens/Credential/CredentialDetail/CredentialDetail.jsx
```diff
@@ -142,16 +224,26 @@ function CredentialDetail({ i18n, credential }) {
...
         <UserDateDetail
+          id="credential-last_modified-detail"
           label={i18n._(t`Last Modified`)}
           date={modified}
           user={modified_by}
         />
       </DetailList>
+      {Object.keys(inputSources).length > 0 && (
+        <PluginFieldText>
+          <Trans>
```
AlexSCorey: Why use `<Trans>` rather than doing 
```{i18n._(t * This field will be retrieved from an external secret management system using the specified credential.)```
? which has been our pattern to date

**Review by AlexSCorey — COMMENTED** [review]
I think this looks good.  Just 1 thing in question about `<Trans>`.

File: awx/ui_next/src/screens/Credential/CredentialDetail/CredentialDetail.jsx (hunk 5 above)
Comment (mabashian) [reply awx/ui_next/src/screens/Credential/CredentialDetail/CredentialDetail.jsx]: Oh I thought we had decided to use Trans when the string stands alone in the DOM (ie not in an attribute and doesn't live outside the render function).  I'll change this.  There's no real benefit to one over the other in this case I don't think.

File: awx/ui_next/src/api/models/Credentials.js [hunk 6]
```diff
@@ -15,10 +15,34 @@ class Credentials extends Base {
...
+            },
+          }
+        );
+        if (data.next && requestCounter <= maxRequests) {
+          return fetchInputSources(
+            pageNo + 1,
+            inputSources.concat(data.results)
+          );
+        }
+        return Promise.resolve(inputSources.concat(data.results));
```
jakemcdermott: The credential edit component expects this return structure to be `{ data: { results: [...] } }`.

jakemcdermott (changes requested):

awx/ui_next/src/api/models/Credentials.js
```diff
@@ -15,10 +15,34 @@ class Credentials extends Base {
...
+        const { data } = await this.http.get(
+          `${this.baseUrl}${id}/input_sources/`,
+          {
+            params: {
+              page: pageNo,
+              page_size: 200,
+            },
+          }
+        );
+        if (data.next && requestCounter <= maxRequests) {
```
jakemcdermott:
```suggestion
        if (data?.next && requestCounter <= maxRequests) {
```

awx/ui_next/src/api/models/Credentials.js
```diff
@@ -15,10 +15,34 @@ class Credentials extends Base {
...
+          return fetchInputSources(
+            pageNo + 1,
+            inputSources.concat(data.results)
+          );
+        }
+        return Promise.resolve(inputSources.concat(data.results));
+      } catch (error) {
+        return Promise.reject(error);
```
jakemcdermott, line 50: Need to surface this error in [CredentialEdit.jsx#L154](https://github.com/ansible/awx/blob/devel/awx/ui_next/src/screens/Credential/CredentialEdit/CredentialEdit.jsx#L154)

```jsx
return <ContentError error={error} />;
```

awx/ui_next/src/screens/Credential/CredentialDetail/CredentialDetail.jsx
```diff
@@ -75,34 +97,84 @@ function CredentialDetail({ i18n, credential }) {
 
   const { error, dismissError } = useDismissableError(deleteError);
 
-  const renderDetail = ({ id, label, type }) => {
-    let detail;
+  const renderDetail = ({ id, label, type, ask_at_runtime }) => {
```
jakemcdermott, line 102: No need to nest this function definition inside of the `CredentialDetail`, it can be moved up to module scope and out of the detail component.

awx/ui_next/src/api/models/Credentials.js
mabashian: https://github.com/ansible/awx/pull/7281/commits/bbafeadd0feeecf393b9ccc719485d01d242ef9d

awx/ui_next/src/api/models/Credentials.js
mabashian: https://github.com/ansible/awx/pull/7281/commits/b1a5b7a76750cce4538ca57f5c4d832264604b82

awx/ui_next/src/api/models/Credentials.js
mabashian: https://github.com/ansible/awx/pull/7281/commits/abb7486eb06399a7ce6facd516925118d6e02a30

jakemcdermott (approved):

unlikelyzero:
This PR is blocked by https://github.com/ansible/awx/issues/7814

mabashian:
OK @unlikelyzero I rebased after the PR for #7814 merged so I think this should be unblocked

unlikelyzero:
Awesome. I'm heads down on fixing the credential test so I won't be able to get to this today. I'll take it up first thing Monday

dsesami:
Having some trouble with prompt on launch.

1. Make a machine credential. Take all of the possible fields and set ALL of them to prompt on launch.
2. Edit a basic job template to use this machine credential.
3. Launch the job.

The prompt on launch modal does not prompt for the required fields -- it simply shows the final preview. Attempting to launch results in a 400.
Screenshot from 2020-09-11 13-22-39

dsesami:
This is going to be rebased when https://github.com/ansible/awx/issues/7015 is complete. After that, it can be put in needs_test for final review.

unlikelyzero:
Covered by 5627

PR #7286, opened 2020-06-08
Adding tower_api and tower_get_id lookup plugins

john-westcott-iv:
##### SUMMARY

Adding lookup plugins for AWX collections.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - Collection

##### AWX VERSION

```
awx: 11.2.0
```

##### ADDITIONAL INFORMATION

```

```


AlanCoding:
could this hypothetically replace `tower_job_list`?

john-westcott-iv:
> could this hypothetically replace `tower_job_list`?

List yes. But not cancel. In theory you could combine it with a loop for wait as well.

awx_collection/plugins/lookup/tower_get_id.py
```diff
@@ -0,0 +1,82 @@
...
+    description:
+      - The query parameters to search for in the form of key/value pairs.
+    type: dict
+    required: True
+"""
+
+EXAMPLES = """
+- name: Lookup a users ID
+  debug:
+    msg: "{{ query('awx.awx.tower_api', 'users', query_params={ 'username': 'admin' }) }}"
```
AlanCoding: What does `msg` wind up being in this case? A number like `"42"`?

awx_collection/plugins/lookup/tower_get_id.py
```diff
@@ -0,0 +1,82 @@
...
+"""
+
+RETURN = """
+_raw:
+  description:
+    - The ID found for the filter criteria
```
AlanCoding: might want to note - this can be cast to an integer, but it is not automatically.

awx_collection/tests/integration/targets/tower_lookup_get_id_plugin/tasks/main.yml
```diff
@@ -0,0 +1,78 @@
...
+      register: result
+
+    - assert:
+        that:
+          - result is failed
+          - "'You must pass exactly one endpoint to query' in result.msg"
+
+    - name: Try to load invalid endpoint
+      debug:
+        msg: "{{ query('awx.awx.tower_get_id', 'john', query_params={}, ) }}"
```
AlanCoding: this will be the same if you don't pass `query_params={}` right?

awx_collection/tests/integration/targets/tower_lookup_get_id_plugin/tasks/main.yml
```diff
@@ -0,0 +1,78 @@
...
+      ignore_errors: true
+      register: result
+
+    - assert:
+        that:
+          - result is failed
+          - "'The requested object could not be found at' in result.msg"
+
+    - name: Get the ID of the admin user
```
AlanCoding: Get the ID of the _first_ admin user

awx_collection/tests/integration/targets/tower_lookup_api_plugin/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+    - name: Test too many params (failure from validation of terms)
+      set_fact:
+        junk: "{{ query('awx.awx.tower_api', 'users', 'teams', query_params={}, ) }}"
+      ignore_errors: true
+      register: result
+
+    - assert:
+        that:
+          - result is failed
+          - "'You must pass exactly one endpoint to query' in result.msg"
```
AlanCoding, line 50: This is really overly repetitive with the other test. It's the same thing, just with a different lookup plugin. Could we just combine them in a loop?

awx_collection/tests/integration/targets/tower_lookup_api_plugin/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+          - "'The requested object could not be found at' in result.msg"
+
+    - name: Load user of a specific name
+      set_fact:
+        users: "{{ query('awx.awx.tower_api', 'users', query_params={ 'username' : user_creation_results['results'][0]['item'] }) }}"
+
+    - assert:
+        that:
+          - users['results'] | length() == 1
+          - users['count'] == 1
```
AlanCoding: you could do

```yaml
    - users['results'][0]['id'] == user_creation_results['results'][0]['id']
```

That would be a non-trivial assertion (the query was given the username, and here we expect the id).

awx_collection/tests/integration/targets/tower_lookup_api_plugin/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+
+    - name: Get the id of the admin users
```
AlanCoding: This is getting the id of the _first_ user - also the exact thing that `awx.awx.tower_get_id` does

awx_collection/tests/integration/targets/tower_lookup_api_plugin/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+      set_fact:
+        user_id: "{{ (query('awx.awx.tower_api', 'users', query_params=query_params) | json_query(jmes_query))[0] }}"
+      vars:
+        query_params:
+          username: "{{ user_creation_results['results'][0]['item'] }}"
+        jmes_query: 'results[*].id'
+
+    - assert:
+        that: "{{ user_id }} == {{ user_creation_results['results'][0]['id'] }}"
```
AlanCoding: It feels like this whole thing could be added to the last test as another assertion, and it would be better to delete this task.

awx_collection/tests/integration/targets/tower_lookup_api_plugin/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+
+    - assert:
+        that: users['results'] | length() == 2
+
+    - name: Get all users of a system through next attribute
+      set_fact:
+        users: "{{ query('awx.awx.tower_api', 'users', query_params={ 'page_size': 1, }, get_all=true ) }}"
+
+    - assert:
+        that: users['results'] | length() >= 3
```
AlanCoding: These last 2 tests are great tests.

However, it relies on assumptions if I read it correctly. You need to include the filter in query params - `query_params={ 'username__startswith': 'AWX-Collection-tests-tower_get_id-' }`, with the lookup plugin name replaced.

Even if it won't _fail_ without this, there could be 1,000 users, and the test will not finish in a reasonable time. That's very possible in the environments that @squidboylan and I run these in.

awx_collection/plugins/lookup/tower_api.py
```diff
@@ -0,0 +1,114 @@
...
+            argument_spec={}, direct_params=module_params,
+            error_callback=self.handle_error, warn_callback=self.warn_callback
+        )
+
+        self.set_options(direct=kwargs)
+
+        if self.get_option('get_all'):
+            return_data = module.get_all_endpoint(terms[0], data=self.get_option('query_params'))
+        else:
+            return_data = module.get_endpoint(terms[0], data=self.get_option('query_params'))
```
AlanCoding: what if someone wants to go rogue and GET an endpoint that's not formatted the usual way? Like, they want the config endpoint, or ping endpoint, or some obscure OAuth2 token endpoint. There's also data in sub-endpoints like `/api/v2/job_templates/N/launch/` and many others that have info that's important and not available elsewhere.

awx_collection/tests/integration/targets/tower_lookup_get_id_plugin/tasks/main.yml
john-westcott-iv: If you don't pass query_parameters this will just fail because query_parameters are required.

awx_collection/plugins/lookup/tower_get_id.py
john-westcott-iv: It should but you pointed out a bug to me. This has a call to tower_api instead of tower_get_api which would mean that msg would be an object instead of an int. I'll get that fixed.

awx_collection/plugins/lookup/tower_get_id.py
john-westcott-iv: Good call, done.

awx_collection/tests/integration/targets/tower_lookup_get_id_plugin/tasks/main.yml
john-westcott-iv: Fixed.

awx_collection/tests/integration/targets/tower_lookup_api_plugin/tasks/main.yml
john-westcott-iv: Are you thinking a separate target like `integration/targets/lookups_common/tasks/main.yml`?

awx_collection/tests/integration/targets/tower_lookup_api_plugin/tasks/main.yml
john-westcott-iv: My intention her was to validate that the query params were properly being sent back for the filtering (and that we got one user in the response). But I've added that assertion to make sure we also got the right one we expected.

awx_collection/tests/integration/targets/tower_lookup_api_plugin/tasks/main.yml
john-westcott-iv: Correct, this is doing it an "ansible" way instead of through the plugin. I did this initially because I started writing only this plugin. After trying some "ansible ways" to extract complex ids I decided to include the second plugin. The admin in the name was an artifact of initial writing tests using the "admin" id. But its been updated.

awx_collection/tests/integration/targets/tower_lookup_api_plugin/tasks/main.yml
john-westcott-iv: Agreeded, removed.

awx_collection/tests/integration/targets/tower_lookup_api_plugin/tasks/main.yml
john-westcott-iv: Good point on the 1,000 of users, that could throw off the get_all. I've add the username__startswith like you suggested so that at most we will only ever get back 3.

awx_collection/plugins/lookup/tower_api.py
john-westcott-iv: Good feedback. Maybe we add a parameter like "raw" which would just return the result['json'] file and not parse the objects out? If you like that idea should a raw return be the default or should we change it to you have to ask for raw? I think I am leaning towards raw being the default.

awx_collection/tests/integration/targets/tower_lookup_api_plugin/tasks/main.yml
john-westcott-iv: Actually, I changed this from startwith to endswith because the test ID should be unique incase you had leftover users from an old test.

awx_collection/plugins/lookup/tower_api.py
AlanCoding: > and not parse the objects out?

_does_ this parse anything? The `get_endpoint` method looks pretty bare bones. I didn't want to imply that it doesn't work right now, it's just not documented one way or the other. That's also a really high priority to write tests for. I think it would be worth some tests even for corner cases like `api/v2/job_templates/N/launch`, getting nice errors in those cases would be good

If it's given a "full" endpoint, then maybe `get_all` shouldn't be accepted. In that case, it's an endpoint of user's selection and we don't know if it'll be a list view.

awx_collection/tests/integration/targets/tower_lookup_api_plugin/tasks/main.yml
AlanCoding: I'm fine to add 3 new test files if it results in more targeted tests and fewer tasks. So 👍  from me to that idea. @squidboylan what do you think?

awx_collection/plugins/lookup/tower_api.py
john-westcott-iv: For parsing the objects I was really just thinking about the logic for "get_all" where it looks for the next entry link and go get more item and merges them into the results.

john-westcott-iv: I've modified the plugin to return generic GETs. I've also modified the tower_api.py method get_all to take an optional param to not have a fatal error if ['next'] isn't in the response. This should allow the plugin to do a get against anything.

AlanCoding:
Filed https://github.com/ansible/ansible/issues/70050 related to the problems of combining the test files into a common lookup plugin test.

AlanCoding:
I think I still need to establish some basic facts about how lookup plugins are expected to work.

consider:

```yaml
---
- hosts: localhost
  gather_facts: false
  connection: local
  collections:
    - awx.awx
  tasks:
    - name: debug all organizations
      debug:
        msg: "{{ item }}"
      with_awx.awx.tower_api: organizations
```

gives:

```
TASK [debug all organizations] ******************************************************************************************************************************************************
ok: [localhost] => (item=count) => {
    "msg": "count"
}
ok: [localhost] => (item=next) => {
    "msg": "next"
}
ok: [localhost] => (item=previous) => {
    "msg": "previous"
}
ok: [localhost] => (item=results) => {
    "msg": "results"
}
```

I want this to jive well with the basic documentation on lookup plugins, https://docs.ansible.com/ansible/latest/plugins/lookup.html, but even those docs are somewhat confusing to me.

There is a lot of looseness with types here too. If I give `with_awx.awx.tower_api: {}` then it gives a `KeyError` because it tries to grab `[0]` from a dict. I could actually make a `0` entry to the dict, which would be a dirty hack. I am still fumbling around with expectations.

To set some anchor points, I'm looking over the library of lookup plugins in:

https://github.com/ansible-collections/community.general/tree/master/plugins/lookup

Consider:

```yaml
  - debug:
      msg: 'key contains {{item}}'
    with_consul_kv:
      - 'key/to/retrieve'
  - name: Parameters can be provided after the key be more specific about what to retrieve
    debug:
      msg: 'key contains {{item}}'
    with_consul_kv:
      - 'key/to recurse=true token=E6C060A9-26FB-407A-B83E-12DDAFCB4D98'
  - name: retrieving a KV from a remote cluster on non default port
    debug:
      msg: "{{ lookup('consul_kv', 'my/key', host='10.10.10.10', port='2000') }}"
```

Is making use of `query` trying to fit a square peg into a round hole here? Should we, instead of returning a dict for the response, return a list with a single-element with that response? Should we correspondingly have some option to _expect_ a list from the API so that it more coherently loops over `results`? I'm leaning in this direction.

AlanCoding:
Right now this works:

```yaml
    - name: debug all organizations
      debug:
        msg: "{{ item['id'] }}"
      loop: "{{ query('awx.awx.tower_api', 'organizations', query_params={'description': 'foo'})[0].results }}"
      loop_control:
        label: "{{ item['name'] }}"
```

and it is relatively nice.

I would like that same thing to work without the `[0].results`. It just doesn't jive with what lookup plugins are.

I would also prefer, to be honest, to consolidate both of these into a single lookup plugin. These two things are related. Imagine adding new parameters:

 - flatten_lists: if provided, will promote results to the list which is returned, may default to True
 - expect_one: if provided, will fail if we get >1 result in a list view, or do not get an item or list view, default to False
 - only_ids: if provided, will not return dicts for each object, but just the id, applies for items and lists, default to False

I think these should be raised as a broader discussion.

AlanCoding:
Further reflecting on this, we want to _very carefully_ distinguish the options that affect return types, so this naming might be better:

 - `return_objects`: promote the `results` in list views to the top-level of the list returned. Default True
 - `return_ids`: promote the `id` key inside of an object to top-level of the list returned. Default False

john-westcott-iv:
So what do we want to do about non-object pages? i.e. api/v2/settings? Just return that response as is?

AlanCoding:
@john-westcott-iv `expect_one` would be a parameter that controls the error behavior. If it gets more than 1, it would throw an error. You could have `expect_object` which would strictly enforce that the page is either a list view or a detail view.

There's always a question of where do you land on strictness. Looking at other lookup plugins, I believe they tend to error on the side of being too loose. For instance, the `lookup` syntax itself will default to giving the _first_ item in a list 😱 . If you make the default behavior to error for >1 item, then I think you are conflicting with their philosophy. First, allow things to _work_, and then give users the ability to put in guard rails.

I just want to fit in well to the Ansible ecosystem.

awx_collection/plugins/lookup/tower_api.py
```diff
@@ -0,0 +1,102 @@
...
+options:
+  _terms:
+    description:
+      - The endpoint to query, i.e. teams, users, tokens, job_templates, etc.
+    required: True
+  query_params:
+    description:
+      - The query parameters to search for in the form of key/value pairs.
+    type: dict
+    required: False
```
AlanCoding, line 27: I find it tempting to give this an alias... maybe multiple aliases

```
   aliases: [query, data, filter, params, query_string, filter_string, url_dosey_doe]
```

AlanCoding:
Here, I put up my alternative in a branch:

https://github.com/ansible/awx/compare/devel...AlanCoding:combined_lookup?expand=1

I think I get the fact that the motivating use case is looking up related items, and that seems important enough that it warrants a separate plugin. But I would rather address that by having a conversation about defaults for the `return_*` parameters.

They are just too similar, I think that maintaining 2 of them is bound to cause problems. Keeping it the same will cause some more initial trouble for some users, but I think that's a worthwhile trade.

john-westcott-iv:
I started looking at this last night as well. For the section you have:
```
        if self.get_option('return_ids'):
            if 'results' in return_data:
                return_data['results'] = [item['id'] for item in return_data['results']]
            elif 'id' in return_data:
                return_data = return_data['id']
```
I'm wondering if we want to drop this or, if not, if we need a second parameter like object_id. This would cover objects that don't use `id` like user. Personally I don't mind having to pass it through json_query plugin in order to get IDs out of the objects.

AlanCoding:
> This would cover objects that don't use id like user

The user model does have ids. I agree we shouldn't do anything specific to `name` fields, but I don't think this does.

awx_collection/plugins/lookup/tower_api.py
```diff
@@ -0,0 +1,168 @@
...
+"""
+
+EXAMPLES = """
+- name: Load the UI settings
+  debug:
+    msg: "{{ query('awx.awx.tower_api', 'settings/ui') }}"
+
+- name: Lookup any users who are admins
+  debug:
+    msg: "{{ query('awx.awx.tower_api', 'users', query_params={ 'is_superuser': true }) }}"
```
AlanCoding: I ran through these examples with your branch, and I would like to add more elaboration.

```yaml
    - name: Load the UI settings  # plugin returns a list of 1 but lookup returns first element
      debug:
        msg: "{{ lookup('awx.awx.tower_api', 'settings/ui') }}"

    - name: Lookup any users who are admins  # use query to return the full list
      debug:
        msg: "{{ query('awx.awx.tower_api', 'users', query_params={ 'is_superuser': true }) | map(attribute='username') }}"

    # use lookup to return first item
    # the filter may return more than 1 item, it is users responsibility to assure this item is what they want
    # this can be used to pass related items to modules in the collection
    - name: Lookup a users ID
      debug:
        msg: "{{ lookup('awx.awx.tower_api', 'users', query_params={ 'username': 'admin' }, return_ids=True) }}"

    - name: debug all organizations in a loop  # use query to return a list
      debug:
        msg: "Organization description={{ item['description'] }} id={{ item['id'] }}"
      loop: "{{ query('awx.awx.tower_api', 'organizations') }}"
      loop_control:
        label: "{{ item['name'] }}"
```

This is probably going to be our primary blurb of documentation that users make use of, so we should make it count.

awx_collection/plugins/doc_fragments/auth_plugin.py
```diff
@@ -0,0 +1,48 @@
...
+            - Specify whether Ansible should verify the SSL certificate of Ansible Tower host.
+            - Defaults to True, but this is handled by the shared module_utils code
+        type: bool
+        env:
+            - name: TOWER_VERIFY_SSL
+        aliases: [ validate_certs ]
+
+notes:
+- If no I(config_file) is provided we will attempt to use the tower-cli library
+  defaults to find your Tower host information.
```
AlanCoding, line 43: > tower-cli library defaults

I'm trying to think if there's any reason for this bullet point to exist at all.... I'm not coming up with anything. I vote to delete it.

Should we also delete:

https://github.com/ansible/awx/blob/824d798d814cd63792c0a25a7ba4f55e9d16776e/awx_collection/plugins/doc_fragments/auth.py#L55-L56

?

Yes. Yes we should.

awx_collection/plugins/doc_fragments/auth_plugin.py
```diff
@@ -0,0 +1,48 @@
...
+- I(config_file) should contain Tower configuration in the following format
+    host=hostname
+    username=username
+    password=password
```
AlanCoding, line 47: I think it also accepts YAML? But it's not wrong, so this is a non-blocking comment...

awx_collection/plugins/lookup/tower_api.py
john-westcott-iv: Yes, I would like to expand this section to include examples of all of the parameters. In addition, I'd like to review the tests to make sure we have corresponding tests for every option.

AlanCoding: it might seem trite, but I also think it would be nice to have a line continuation example, for instance:

```yaml
  - name: Debug using the rrule plugin
    debug:
      msg: >
        {{ query(
          'awx.awx.tower_schedule_rrule',
          'minute',
          start_date='2020-4-16 03:45:07',
          end_on='2'
        ) }}
```

That's closer to how I would like to see a playbook actually written.

awx_collection/plugins/doc_fragments/auth_plugin.py
john-westcott-iv: Yea, this doc was almost a direct copy from the other one (as is also the case with the next comment). 

I feel like something should in here somewhere to indicate that we will, for backwards compatibility, search for this file like the old TowerCLI did. It would be nice to remove the TowerCLI reference, maybe we just have a small blurb with something like https://docs.ansible.com/ansible-tower/3.5.3/html/towerapi/tower_cli.html#file-locations?

Perhaps we let this go in as in as a clone and then cut a second PR to adjust this and the comment below?

awx_collection/tests/integration/targets/tower_lookup_api_plugin/tasks/main.yml
squidboylan: sorry didnt see this until today, I'm not too worried about more test files. It is kinda nice to keep all the tests needed to properly test one module in one target so running locally you dont have to think too hard about what to run, but if you think it's a good idea :+1: from me

AlanCoding:
Looks like this is about finished?

My only outstanding comment are the examples, you can copy and paste my suggestions exactly.

With that out of the way, I don't see any more work left.

awx_collection/plugins/lookup/tower_api.py
```diff
@@ -0,0 +1,202 @@
...
+  max_objects:
+    description:
+      - if C(return_all) is true, this is the maximum of number of objects to return from the list.
+      - If a list view returns more an max_objects an exception will be raised
+    type: integer
+    default: 1000
+
+notes:
+  - If the query is not filtered properly this can cause a performance impact.
+  - In addition, the built in threshold is 10,000 items; if the query returns more an exception will be thrown.
```
AlanCoding: this now conflicts with the `max_objects` help. I might have caused that, sorry. This needs to be rewritten to axed.

awx_collection/plugins/lookup/tower_api.py
```diff
@@ -0,0 +1,202 @@
...
+  tower_job_template:
+    name: "{{ job_template_name }}"
+    credentials: >-
+        {{ lookup(
+            'awx.awx.tower_api',
+            'credentials',
+            query_params={ 'name' : credential_name },
+            return_ids=True,
+            expect_one=True,
+            wantlist=True
```
AlanCoding: isn't lookup with `wantlist` the same as query? I would rather stick to query/lookup syntax

awx_collection/plugins/lookup/tower_api.py
```diff
@@ -0,0 +1,202 @@
...
+"""
+
+EXAMPLES = """
+- name: Load the UI settings
+  set_fact:
+    tower_settings: "{{ lookup('awx.awx.tower_api', 'settings/ui') }}"
+
+- name: Lookup any users who are admins
+  set_fact:
+    admin_user_list_view: "{{ lookup('awx.awx.tower_api', 'users', query_params={ 'is_superuser': true }, return_objects=False) }}"
```
AlanCoding: I really wanted to put together examples where the return type would be obvious from use. That's why I liked

```yaml
    - name: Lookup any users who are admins  # use query to return the full list
      debug:
        msg: "{{ query('awx.awx.tower_api', 'users', query_params={ 'is_superuser': true }) | map(attribute='username') }}"
```

But with what you have, the user has to trust that the naming `_list_view` is accurate (in addition to knowing what an AWX list view is to begin with).

And I also question if we need _an_ example of `return_objects=False` to begin with. It's just not useful in any way that I can image.

awx_collection/plugins/lookup/tower_api.py
```diff
@@ -0,0 +1,202 @@
...
+- name: Create a job template with a looked up credential from a folded lookup
+  tower_job_template:
+    name: "{{ job_template_name }}"
+    credentials: >-
+        {{ lookup(
+            'awx.awx.tower_api',
+            'credentials',
+            query_params={ 'name' : credential_name },
+            return_ids=True,
+            expect_one=True,
```
AlanCoding: this isn't a very natural example to have `expect_one=True` and passing a list. I would like to use something that's a natural fit. The example can be whatever you want it to be.

What if, instead, call `tower_group` to create a group of all hosts that start with `foo_`? Right now it's not using the ability to return a list in a way that's not very coherent to a human.

awx_collection/plugins/lookup/tower_api.py
```diff
@@ -0,0 +1,202 @@
...
+    actual_admin_user: "{{ lookup('awx.awx.tower_api', 'users', query_params={ 'username': 'admin' }, expect_one=True, wantlist=True) }}"
+
+- name: Get just the user ID of the admin user
+  set_fact:
+    admin_user_id: "{{ lookup('awx.awx.tower_api', 'users', query_params={ 'username': 'admin' }, return_ids=True, expect_one=True) }}"
+
+- name: Create a job template with a looked up credential from a folded lookup
+  tower_job_template:
+    name: "{{ job_template_name }}"
+    credentials: >-
```
AlanCoding: What about adding an example of a single Foreign Key, like the `admin_user_id`. For instance, create a credential and give ownership to an arbitrary admin with the `user:` option.

awx_collection/plugins/lookup/tower_api.py
```diff
@@ -0,0 +1,202 @@
...
+
+- name: Lookup any users who are admins
+  set_fact:
+    admin_user_list_view: "{{ lookup('awx.awx.tower_api', 'users', query_params={ 'is_superuser': true }, return_objects=False) }}"
+
+- name: Lookup any users who are admins and get their objects directly
+  set_fact:
+    admin_user_object: "{{ lookup('awx.awx.tower_api', 'users', query_params={ 'is_superuser': true } ) }}"
+
+- name: Lookup the admin user and fail if there are more than one, and make sure its a list (without this, the response would be an object)
```
AlanCoding: these read more like test cases than examples.

I would rather have examples that answer the question of how it would be useful in practice.

awx_collection/plugins/lookup/tower_api.py
```diff
@@ -0,0 +1,202 @@
...
+  set_fact:
+    actual_admin_user: "{{ lookup('awx.awx.tower_api', 'users', query_params={ 'username': 'admin' }, expect_one=True, wantlist=True) }}"
+
+- name: Get just the user ID of the admin user
+  set_fact:
+    admin_user_id: "{{ lookup('awx.awx.tower_api', 'users', query_params={ 'username': 'admin' }, return_ids=True, expect_one=True) }}"
+
+- name: Create a job template with a looked up credential from a folded lookup
+  tower_job_template:
+    name: "{{ job_template_name }}"
```
AlanCoding: these templates leave the examples as something which isn't exactly run-able either. It would be nice to be able to copy and paste it and run it

awx_collection/plugins/lookup/tower_api.py
john-westcott-iv: Removed the second bullet.

awx_collection/plugins/lookup/tower_api.py
john-westcott-iv: I tweaked it slightly:
```
    - name: Report the usernames of all users with admin privs
      debug:
        msg: "Admin users: {{ query('awx.awx.tower_api', 'users', query_params={ 'is_superuser': true }) | map(attribute='username') | join(', ') }}"
```

Let me know if you are good with that.

awx_collection/plugins/lookup/tower_api.py
john-westcott-iv: This is a fair comment, this was initially an attempt to lookup an assets ID which we talked about the other day. I like the group idea and will give that a try.

john-westcott-iv: So for the group example. The hosts list of tower_group is another place where it can't do ids in place of names. But I was able to come up with an example for it.

awx_collection/plugins/lookup/tower_api.py
john-westcott-iv: It is. Updated examples with wantlist=True to query.

awx_collection/plugins/lookup/tower_api.py
john-westcott-iv: I'm torn on this because for complex items (like a job template) we could take out the variables but if someone does not have an inventory we reference it will still fail. Unless we added "full examples" to make sure all other assets exist we run the risk of a copy/paste failing. Given that I'd almost rather see variables in there indicating that the example requires some data to function.

awx_collection/plugins/lookup/tower_api.py
john-westcott-iv: I added an example which uses a when clause to see if a user exists before applying a role. LMK what you think of that.

awx_collection/plugins/lookup/tower_api.py
AlanCoding: 🥇 💯 🥳  is how I feel about that

awx_collection/plugins/lookup/tower_api.py
```diff
@@ -0,0 +1,195 @@
...
+
+- name: Report the usernames of all users with admin privs
+  debug:
+    msg: "Admin users: {{ query('awx.awx.tower_api', 'users', query_params={ 'is_superuser': true }) | map(attribute='username') | join(', ') }}"
+
+- name: Lookup any users who are admins and get their objects directly
+  set_fact:
+    admin_user_object: "{{ lookup('awx.awx.tower_api', 'users', query_params={ 'is_superuser': true } ) }}"
+
+- name: Make sure user 'john' is an org admin of the default org if the user exists
```
AlanCoding, line 86: You mean make sure John is _not_ an admin of the default org, given that `state: absent`. Poor John.

(but speaking practically, it might be less awkward to delete the "state" and register" lines in the task)

awx_collection/plugins/lookup/tower_api.py
AlanCoding: I see what you're saying, yes, but only for the destructive cases.

Let's draw a line in the sand here, because the first half of these examples are beautiful and universal tasks which can be copied into any playbook and ran wherever auth is set up.

Printing the username of all admin users - always relevant. There is no case where this doesn't go right.

For the cases where you actually muck with stuff, either you put in tasks to setup prior data (org, inventory, etc.), or you don't try. If you're not trying, then sure, it doesn't matter whether the JT name is templated or not.

You could leave a comment saying that the tasks beyond that comment are not expected to be runnable because they need setup.

awx_collection/plugins/lookup/tower_api.py
```diff
@@ -0,0 +1,195 @@
...
+  set_fact:
+    tower_settings: "{{ lookup('awx.awx.tower_api', 'settings/ui') }}"
+
+- name: Report the usernames of all users with admin privs
+  debug:
+    msg: "Admin users: {{ query('awx.awx.tower_api', 'users', query_params={ 'is_superuser': true }) | map(attribute='username') | join(', ') }}"
+
+- name: Lookup any users who are admins and get their objects directly
+  set_fact:
+    admin_user_object: "{{ lookup('awx.awx.tower_api', 'users', query_params={ 'is_superuser': true } ) }}"
```
AlanCoding: I checked out your branch and ran it, and `admin_user_object` is a list of all admin users, with the full API information. You could say that in the `name:` or in a comment, or you could just change this to a `debug:` task. A similar statement goes for the "Load the UI settings" task. Unless you're going to show how to use the variable, I think there's negative value in doing a set_fact.

AlanCoding:
I think you could outright lose the `admin_user_object` example, and I would like to revive my suggestion:

```yaml
  - name: debug all organizations in a loop  # use query to return a list
    debug:
      msg: "Organization description={{ item['description'] }} id={{ item['id'] }}"
    loop: "{{ query('awx.awx.tower_api', 'organizations') }}"
    loop_control:
      label: "{{ item['name'] }}"
```

I just want people to write decent loops.

awx_collection/plugins/lookup/tower_api.py
john-westcott-iv: I've removed the admin_user_object and replaced it with your loop example.

awx_collection/plugins/doc_fragments/auth_plugin.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2017, Wayne Witzel III <wayne@riotousliving.com>
```
AlanCoding: Let's get rid of this one line.

Should be final comment here!

AlanCoding (approved):

awx_collection/tests/integration/targets/tower_lookup_api_plugin/tasks/main.yml
```diff
@@ -0,0 +1,243 @@
...
+
+    - assert:
+        that:
+          - results is failed
+          - "'Expected one object from endpoint users, but obtained 3' in results['msg']"
+
+    - name: Get the settings page
+      assert:
+        that:
+          - "'CUSTOM_LOGO' in {{ lookup(plugin_name, 'settings/ui' ) }}"
```
AlanCoding: It looks like this caused some new kind of test failure

```
  TASK [tower_lookup_api_plugin : Get the settings page] *************************
  fatal: [testhost]: FAILED! => {"msg": "The conditional check ''CUSTOM_LOGO' in {u'PENDO_TRACKING_STATE': u'off', u'MAX_UI_JOB_EVENTS': 4000, u'CUSTOM_LOGO': u'', u'UI_LIVE_UPDATES_ENABLED': True, u'CUSTOM_LOGIN_INFO': u''}' failed. The error was: template error while templating string: expected token ':', got 'string'. String: {% if 'CUSTOM_LOGO' in {u'PENDO_TRACKING_STATE': u'off', u'MAX_UI_JOB_EVENTS': 4000, u'CUSTOM_LOGO': u'', u'UI_LIVE_UPDATES_ENABLED': True, u'CUSTOM_LOGIN_INFO': u''} %} True {% else %} False {% endif %}"}
```

I don't quite know what that means, but this does seem duplicated with the other task that saves the result to `tower_settings`.

Also looks like this might be related to a bug in the assert module...

john-westcott-iv: It is a duplicate but it's odd that its failing in your tests (it passes in mine locally). I wonder if its a python/ansible version issue? Either way, we can remove this one.

squidboylan (approved):

PR #7300, opened 2020-06-09
Ignore the rsyslog.conf file

jbradberry:
##### SUMMARY
Ignore the rsyslog.conf file since it is a file generated during the build process.

Recently when running the dev environment, you would notice this when you run `git status`:

    Untracked files:
      (use "git add <file>..." to include in what will be committed)
	    tools/rsyslog/rsyslog.conf

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API
 - Installer

##### AWX VERSION
```
awx: 11.2.0
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ rsyslog.pid
 /tower-license/**
 tools/prometheus/data
 tools/docker-compose/Dockerfile
+tools/rsyslog/rsyslog.conf
 
 # Tower setup playbook testing
 setup/test/roles/postgresql


beeankha (approved):

rooftopcellist (approved):
I don't see us needing to change the default rsyslog.conf much, if at all.  I think we can safely put this in the `.gitignore`

ryanpetrello (changes requested):
I think this is actually just indicative of something in our dev env docker compose setup that we just don't need (and which @shanemcd very recently removed, because he noticed it, too):

https://github.com/ansible/awx/pull/7261/files#diff-0da361373f3517322c1b79b1e0288a12L37

jbradberry:
Per discussion, it appears that this problem had already been fixed in another way.  Closing.

PR #7305, opened 2020-06-09
patternfly v4 upgrade

jlmitch5:
This pulls in new patternfly dependencies and updates our codebase to fix any issues that arose.


awx/ui_next/src/components/Background/Background.jsx
```diff
@@ -1,18 +1,10 @@
 import React, { Fragment } from 'react';
 
-import { BackgroundImage, BackgroundImageSrc } from '@patternfly/react-core';
-
-const backgroundImageConfig = {
-  [BackgroundImageSrc.xs]: './images/pfbg_576.jpg',
-  [BackgroundImageSrc.xs2x]: './images/pfbg_576@2x.jpg',
-  [BackgroundImageSrc.sm]: './images/pfbg_768.jpg',
-  [BackgroundImageSrc.sm2x]: './images/pfbg_768@2x.jpg',
-  [BackgroundImageSrc.lg]: './images/pfbg_2000.jpg',
```
jakemcdermott, line 10: finally :+1:

unlikelyzero:
If you don't mind, I'd like to run this through our dependency pipeline as well as our e2e testsuite

unlikelyzero:
Adding needs_test as several of our tests are currently broken on this branch

awx/ui_next/src/components/About/About.test.jsx
```diff
@@ -14,6 +14,7 @@ describe('<About />', () => {
 
   test('close button calls onClose handler', () => {
     aboutWrapper = mountWithContexts(<About isOpen onClose={onClose} />);
+    console.log(aboutWrapper.debug())
```
AlexSCorey: Probably want to remove this

awx/ui_next/src/screens/Project/Project.jsx
```diff
@@ -121,6 +120,16 @@ class Project extends Component {
     const canToggleNotifications = isNotifAdmin;
 
     const tabsArray = [
+      {
```
AlexSCorey, line 123: Why are we adding this tab here?  No other resource has this tab unless it is a nested resource like Inventory Sources

awx/ui_next/src/screens/Team/TeamAccess/TeamAccessListItem.jsx
```diff
@@ -40,9 +40,6 @@ function TeamAccessListItem({ role, i18n, detailUrl, onSelect }) {
                     label={i18n._(t`Role`)}
                     value={
                       <Chip
-                        isReadOnly={
```
AlexSCorey, line 43: Was this removed intentionally?  We probably need to keep it so that only those that have permission to remove the item are given the button to do so.

AlexSCorey (changes requested):
I would like to see us begin to use the 'success' validate text too when the user has a validation error and then fixes the error. This is probably more of a comment for the team and for @trahman73 to think about and not really a blocker for this PR.
https://www.patternfly.org/v4/documentation/react/components/form

I'd also say that when QE gets their hands on this we should to all of there test suites, including the visual regression testing with users that have several different levels of permissions just to make sure we don't remove something that would give or take away a users rights.

awx/ui_next/src/screens/Project/Project.jsx
jlmitch5: we are removing the back button and moving to this pattern.  i need to add this to the rest of the views

awx/ui_next/src/screens/Team/TeamAccess/TeamAccessListItem.jsx
jlmitch5: it looks like the chips are failing tests too.  i'll take a look at those in just a second

awx/ui_next/src/components/About/About.test.jsx
jlmitch5: yep, thanks for the reminder

awx/ui_next/src/screens/Team/TeamAccess/TeamAccessListItem.jsx
jlmitch5: i had to add these back in.  their code mods deleted them erroneously

awx/ui_next/src/screens/Project/Project.jsx
AlexSCorey: this is good to know because I'm working on a details view right now

AlexSCorey (approved):
Those fixes look good to me.

unlikelyzero:
Removing needs_test as it has passed verification and @appuk has a branch for aligning testcode to application code.

jakemcdermott (approved):

PR #7308, opened 2020-06-10
Allow for SCM installs of the AWX collection

AlanCoding:
##### SUMMARY
This is a continuation of https://github.com/ansible/awx/pull/7123 (ping @john-westcott-iv ), to the logical conclusion of that train of thought.

This is also to obtain a form of compliance with https://github.com/ansible/ansible/pull/69154, such that you can install the collection with the requirements file:

```yaml
collections:
   - name: https://github.com/alancoding/awx.git#awx_collection,in_place_everything
     type: git
```

You can check `ansible-galaxy collection list` after this and see the placeholder version. ~We might want to get consistency with placeholder versions, and put it here too:~

https://github.com/ansible/awx/blob/442d539ff89ecfdd6cf7fb706d4e86f25aa96a8e/awx_collection/plugins/module_utils/tower_api.py#L36

~That just might take more work.~ (this has been done)

##### ISSUE TYPE
 - Feature Pull Request

##### WHAT THIS DOES

Instead of templating everything in-place (which has hazards for someone accidentally committing those changes), it templates it to a new "build" folder, and doesn't keep around old `*.tar.gz` files.

This separates the "source" version of the collection from the templated version of the collection.

It also templates 2 files a second time direction into the "source" version. This is to keep them synced with the template, and also allow for installing & using from source. If you see the version `0.0.1-devel` then you know you're running from source, that's the idea.

### DIFF
--- a/awx_collection/galaxy.yml
+++ b/awx_collection/galaxy.yml
@@ -1,0 +1,26 @@
+authors:
+- AWX Project Contributors <awx-project@googlegroups.com>
+dependencies: {}
+description: Ansible content that interacts with the AWX or Ansible Tower API.
+documentation: https://github.com/ansible/awx/blob/devel/awx_collection/README.md
+homepage: https://www.ansible.com/
+issues: https://github.com/ansible/awx/issues?q=is%3Aissue+label%3Acomponent%3Aawx_collection
+license:
+- GPL-3.0-only
+name: awx
+namespace: awx
+readme: README.md
+repository: https://github.com/ansible/awx
+tags:
+- cloud
+- infrastructure
+- awx
+- ansible
+- automation
+version: 11.2.0
+build_ignore:
+- tools
+- setup.cfg
+- galaxy.yml.j2
+- template_galaxy.yml
+- '*.tar.gz'
--- a/awx_collection/galaxy.yml
+++ b/awx_collection/galaxy.yml
@@ -1,3 +1,14 @@
+# (********************************************************)
+# (*                                                      *)
+# (*             WARNING                                  *)
+# (*                                                      *)
+# (*  This file is managed by Ansible and not to be       *)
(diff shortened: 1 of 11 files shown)


AlanCoding:
Pushed a commit which changes the module_utils version, so this will be good for consistency.

If there's anything else to do, it's to document how it is expected to operate.

awx_collection/tools/roles/template_galaxy/templates/galaxy.yml.j2
```diff
@@ -1,26 +1,38 @@
...
+# (*  edited directly! Instead modify:                    *)
+# (*  tools/roles/template_galaxy/templates/galaxy.yml.j2 *)
+# (*                                                      *)
+# (*  Changed to the base README.md file are lost upon    *)
+# (*  the build of the collection                         *)
+# (********************************************************)
+---
 authors:
-- AWX Project Contributors <awx-project@googlegroups.com>
+  - AWX Project Contributors <awx-project@googlegroups.com>
```
AlanCoding, line 14: this file was not previously subjected to yamllint... which may be an argument for making these changes.

awx_collection/galaxy.yml
```diff
@@ -0,0 +1,38 @@
+# (********************************************************)
+# (*                                                      *)
+# (*             WARNING                                  *)
+# (*                                                      *)
+# (*  This file is managed by Ansible and not to be       *)
+# (*  edited directly! Instead modify:                    *)
+# (*  tools/roles/template_galaxy/templates/galaxy.yml.j2 *)
+# (*                                                      *)
+# (*  Changed to the base README.md file are lost upon    *)
```
AlanCoding: need some wording changes

beeankha (approved):

john-westcott-iv:
This looks good to me.

PR #7311, opened 2020-06-10
set proper permissions for the redis socket

ryanpetrello:


elyezer (approved):

wenottingham:
this requires the emptyDir that is mounted here to inherit at least redis.redis 77x ownership. Is that the case?

ryanpetrello:
This isn't quite right yet for k8s - tinkering and will reopen another PR.

PR #7312, opened 2020-06-10
Add stub files for Credential Types

### LINKED ISSUE #7301: Custom Credential Type Routing Stub
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Stub out the routing and the needed components for Custom Credential Type

nixocio:
Add stub files for Credential Types. Routing system, and screens layout.

closes: https://github.com/ansible/awx/issues/7301

### DIFF
--- a/awx/ui_next/src/screens/CredentialType/CredentialTypeAdd/CredentialTypeAdd.jsx
+++ b/awx/ui_next/src/screens/CredentialType/CredentialTypeAdd/CredentialTypeAdd.jsx
@@ -1,0 +1,14 @@
+import React from 'react';
+import { Card, PageSection } from '@patternfly/react-core';
+
+function CredentialTypeAdd() {
+  return (
+    <PageSection>
+      <Card>
+        <div>Credentials Type Add</div>
+      </Card>
+    </PageSection>
+  );
+}
+
+export default CredentialTypeAdd;
--- a/awx/ui_next/src/screens/CredentialType/CredentialType.jsx
+++ b/awx/ui_next/src/screens/CredentialType/CredentialType.jsx
@@ -1,0 +1,25 @@
+import React from 'react';
+import { Route, Switch, Redirect } from 'react-router-dom';
+
+import CredentialTypeDetails from './CredentialTypeDetails';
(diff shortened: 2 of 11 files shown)


awx/ui_next/src/screens/CredentialType/CredentialType/CredentialType.jsx
```diff
@@ -0,0 +1,27 @@
...
+          from="/credential_types/:id"
+          to="/credential_types/:id/details"
+          exact
+        />
+        <Route path="/credential_types:/:id/edit">
+          <CredentialTypeDetails />
+        </Route>
+        <Route path="/credential_types:/:id/details">
+          <CredentialTypeEdit />
+        </Route>
```
marshmalien: The route paths and components should be swapped (the details component is currently under the edit route) and there is an extra colon after credential_types in the path `/credential_types:/:id/*`.

awx/ui_next/src/screens/CredentialType/CredentialTypeAdd/CredentialTypeAdd.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Card, PageSection } from '@patternfly/react-core';
+
+function CredentialTypeAdd() {
+  return (
+    <>
```
marshmalien: We can remove this React Fragment `<> </>` since we are returning a single element.

marshmalien (changes requested):
This looks great - I like how you decided to stub out the routes and components in a single PR. We could get rid of a few React Fragments since we only need it when we are returning more than one element. 

We should relocate CredentialType.jsx from `/CredentialType/CredentialType/CredentialType.jsx` to the top-level of the CredentialType directory. This component should be siblings with CredentialTypes.jsx.

awx/ui_next/src/screens/CredentialType/CredentialType/CredentialType.jsx
nixocio: Thanks for your review @marshmalien. Updated per comments.

marshmalien (approved):
🔥Nice!

PR #7313, opened 2020-06-10
remove some references to memcached

kdelee:
AFAIK we removed memcached depedency but I still see lots of references to it....this is a start but there are a good number of other references at least in comments + docs. Would be good to go ahead and clear up what that code is now doing if it is not using memcached


rebeccahhh (approved):
I think this is an excellent start

ryanpetrello (approved):

PR #7317, opened 2020-06-10
Force containers in k8s to run under root group

shanemcd:
Normally containers belong to the 'root' group, but for some reason the downstream red hat scl redis image only belongs to the 'redis' group by default. This fixes that.


ryanpetrello (approved):

ghjm:
Nicely done

PR #7319, opened 2020-06-10
Remove active_counts_by_org

wenottingham:
I was trying to parse the difference between this and the
(directly above) org_active_count from the comment, and then I
grepped and realized this function appears unused.


ryanpetrello (approved):
Yep, this does look unused.

PR #7321, opened 2020-06-10
Instance id fallback

### LINKED ISSUE #3461: Fallback instance_id values for inventory imports

jladdjr:
When creating instance id refer to _list_ of hostvars. If first hostvar is not available, walk down list of hostvars until one is defined. This allows awx to be backwards compatible when forming instance id.

Fixes https://github.com/ansible/awx/issues/3461

### DIFF
--- a/awx/main/management/commands/inventory_import.py
+++ b/awx/main/management/commands/inventory_import.py
@@ -271,7 +271,7 @@ class Command(BaseCommand):
                                      logging.DEBUG, 0]))
         logger.setLevel(log_levels.get(self.verbosity, 0))
 
-    def _get_instance_id(self, from_dict, default=''):
+    def _get_instance_id(self, variables, default=''):
         '''
         Retrieve the instance ID from the given dict of host variables.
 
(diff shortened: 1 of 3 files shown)


jladdjr:
@AlanCoding - base on https://github.com/ansible/awx/issues/3461, looks like we should also update the settings for gce, is that right? And are there any others that you're aware of?

AlanCoding:
no, I think we should only be doing this for vmware, because we solved the other cases by other means. VMware, due to the sub-properties, is the only one we made an exception for where the hostvars keys changed. Because of that, there's no workaround for it.

AlanCoding (changes requested):
passes testing 👍 

only remove the EC2 instance id change

jladdjr:
@AlanCoding reverted change to `EC2_INSTANCE_ID_VAR`

AlanCoding (approved):

jladdjr:
recheck

jladdjr:
~I think this is in Zuul limbo. Reaching out to some folks to try to get this unstuck.~ Nope, I'm just a Zuul noob still.

PR #7338, opened 2020-06-15
Hook up Test button on Metadata step in credential plugin wizard

mabashian:
##### SUMMARY
link #7126 

We're talking about this Test button:

Screen Shot 2020-06-15 at 11 20 44 AM

and not this Test button:

Screen Shot 2020-06-10 at 4 39 31 PM

The Test button on the Credential form will come in a later PR.  While they functionally do the same thing, the UX is a bit different.

Here's what it looks like in action:

cred_plugin_test

Note that I created `CredentialPluginTestAlert` in preparation for the other Test button.  I wanted to create a single button but the UX for the form level button is a bit different.  The form level button will pull up another form in a modal before firing off the `test` call.  Both buttons have this same `Alert` UX though so I went ahead and peeled that out into it's own component.

A note about the change to the `useRequest` hook: This may be the first time we're running in to this or maybe I solved it the wrong way but the use case I have is that the user can click this Test button as many times as they want.  If the first request fails but the second one succeeds then the error value from the first request is not wiped out and hangs around.  This was causing problems for conditional renders based on whether there was an error set.  It seemed to me the easiest way to solve this was to reset the error/result on each request.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI


mabashian:
recheck

awx/ui_next/src/screens/Credential/shared/CredentialPlugins/CredentialPluginPrompt/MetadataStep.jsx
```diff
@@ -27,6 +28,21 @@ function MetadataStep({ i18n }) {
...
+  const {
+    result: testPluginSuccess,
+    error: testPluginError,
+    request: testPluginMetadata,
+  } = useRequest(
+    useCallback(
+      async (credential, metadata) =>
+        CredentialsAPI.test(credential.id, {
+          metadata,
+        }),
```
keithjgrant: This function should access `selectedCredential.value` and `inputValues.value` directly here, rather than passing them in as parameters (this will prevent re-defening the function on every render down on line 161— use `onClick={testPluginMetadata}` instead)

keithjgrant: I think this might fix your problem with the error lingering from earlier requests, too, so you can clean up a lot of the state logic in CredentialPluginTestAlert, too.

…if that doesn't fix it, I suspect there may be a bug in useRequest that we'll have to address

awx/ui_next/src/screens/Credential/shared/CredentialPlugins/CredentialPluginTestAlert.jsx
```diff
@@ -0,0 +1,82 @@
...
+            t`Something went wrong with the request to test this credential and metadata.`
+          )
+        );
+      }
+      setTestVariant('danger');
+    } else if (successResponse) {
+      setTestMessage(i18n._(t`Test passed`));
+      setTestVariant('success');
+    }
+  }, [i18n, successResponse, errorResponse]);
```
keithjgrant, line 51: since these values are derived directly from props, they probably don't need to be stored in state/useEffect. Can you just extract them directly from `errorResponse` and `successResponse`?

awx/ui_next/src/screens/Credential/shared/CredentialPlugins/CredentialPluginPrompt/MetadataStep.jsx
mabashian: This should be addressed with https://github.com/ansible/awx/pull/7338/commits/0c9df15fa53ca1fb70ca09a34f5029c56d5fd344 iirc

mabashian:
After talking with @jakemcdermott and @trahman73 we decided to move the Test button down into the footer of the Wizard.  This was done here: https://github.com/ansible/awx/pull/7338/commits/bc05e9954977e098ad6f11a150bdcd8ac93dc1fa

jakemcdermott (approved):

unlikelyzero:
Ran the tests against this PR and two changes are necessary. Since i'm in all day training, I don't think I'm going to be able to close this by EOD

mabashian:
recheck

unlikelyzero (approved):

unlikelyzero:
Tested in pull/5517. Found and filed https://github.com/ansible/awx/issues/8241

Issue #6618, opened 2020-04-06
Dynamically update jobs lists based on websocket messages


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Socket behavior has changed in the old app - we should make it match up here.  Use https://github.com/ansible/awx/pull/5968#issuecomment-591565513 as a reference for expected behavior.

We'll need to subscribe to the jobs -> status_changed channel to get these updates.

When we get a message:

- If the job in question already exists in our list, we should use the contents of the websocket to update the row. When a job finishes and the sort order is -finished we should attempt to re-sort the list client side with the jobs without a finished timestamp at the top.
- If we get a new job and the sort order is -finished or -started, we should issue a request to see if the user has access to that job (/api/v2/unified_jobs?id__in=...). This request should be time-banded to every 5 seconds. Note that any search parameters applied are maintained in this request. If rows come back, we should add them to the top of the list and trim this list down to match the page size. If the sort order is something else we shouldn't try to add rows.

**Note that there are several other places where the jobs list component is used**

We have a number of completed jobs lists hanging around the app:

Job Template -> completed jobs
Host -> completed jobs
Inventory -> completed jobs
Smart Inventory -> completed jobs
Workflow Job Template -> completed jobs

We'll want to make sure that this logic works for all of these jobs lists.  The resource specific lists above **only show completed jobs** so we'll have to take that into account.

Blocked until #5957, #5958 land

jakemcdermott:
This issue will replace https://github.com/ansible/awx/issues/5885, which I'm closing.

> Create a nice abstraction, hook, utility, etc. for managing websocket subscriptions for job statuses 
> in the new ui. This will be used to drive status icons and other behavior that depends on the status 
> of running jobs.

dsesami:
This will be closed out after test stubs are created.

PR #5968, opened 2020-02-17
Changes ui-side behavior in response to websocket job status updates on several lists

mabashian:
##### SUMMARY
Part of #5883 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
It might be easiest for me to outline the expected behavior with these changes in order to illustrate what's changed and pose questions.

1. UI gets a socket message with a `status` _other than_ `pending`.
    - UI takes that message, loops across the available rows attempting to find a row that matches up with the `unified_job_id` provided in the socket message.  If it finds a matching row, we update the status of the row to match the status in the message.  (This same pattern will extend out to other fields (started/finished/etc) when provided by the api).

2. UI gets a socket message with at `status` of `pending`.
    - This is where things get interesting
    - This code treats this message like: This is a new job and you don't have it in your list.
        - Although now that I think about it, maybe this isn't quite true.  Could there be scenarios where we _do_ have this job in the list already but it's status is `new`?  If so, the code as it currently stands would potentially insert an extra job for this row (we don't check for this job in our existing list rows).
    - When we get a message like this, we have to go out and perform a GET in order to determine whether or not this particular user has sufficient privileges to see this job _and_ whether or not it fits in with the current search filters.
        - We only do this if the user is on page 1 of the jobs list
        - We only do this if the user has the default `-finished` or similar `-started` `order_by` filter applied to the list.  If they're sorting by something else we have no idea where this potential new row fits into the list.
        - We only do this if the user hasn't already filtered the list down by `id__in`.
            - There's a corner case here - if the user provided an `id__in` filter that matched a future job id.  This probably isn't a realistic scenario though.
        - These GET requests are time-banded on a 5 second interval.  For example, lets say a new job comes in over the websocket and we make a GET request for it: `/api/v2/unified_jobs?id__in=9000&count_disabled=1`.  Over the next 5 seconds, we hold on to all the id's of new jobs that come in and bundle them up into a single request to be executed 5 seconds after the initial request (id=9000).  That second request would look like `/api/v2/unified_jobs?id=9001,9002,9003`&count_disabled=1.  The cycle continues until there are no more new jobs in the past 5 seconds.
        - If we get rows back, this code adds them to the top of the list (this is why the sort order is important).  Page size is maintained by bumping the corresponding number of rows off the bottom.
        - Once the rows are added, future socket messages with updates to these rows fall into scenario 1 above.

[...]

PR #7342, opened 2020-06-15
Jobs List websockets

keithjgrant:
##### SUMMARY
Updated JobsList live based on websocket activity.

* status of already-loaded jobs is updated live
* if new messages come in regarding jobs not currently displayed, they are fetched via API
* these fetches are batched and throttled to no more than one call every five seconds

Addresses #6618

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
I spent a fair bit of time trying to figure out how to stub out Date.now() to test the useThrottle hook, but couldn't get it working properly.

This is a rough cut of websockets for jobs lists. All job lists should update based on WS activity now, though I some of the pages may need to be fine tuned further to get the exact behavior we want where it differs from the main Jobs list page.

Affected pages:

* Jobs List
* Job Template -> completed jobs
* Host -> completed jobs
* Inventory -> completed jobs
* Workflow Job Template -> completed jobs


mabashian:
Started scoping this out but am running in to:

Screen Shot 2020-06-17 at 5 55 08 AM

AlexSCorey:
> Started scoping this out but am running in to:
> 
> Screen Shot 2020-06-17 at 5 55 08 AM

@mabashian Doesn't websockets know that we aren't shaking hands anymore due to social distancing?

keithjgrant:
Oh, I totally forgot about that: websockets only work right now in Firefox after the conversion to react-scripts. @jakemcdermott is looking into that.

mabashian:
@keithjgrant ah thanks!  Will check it out in FF.

AlexSCorey (commented):
In pulling this down I see that if you launch a job, like a JT from the details view, it sends you to the job output.  However, if you immediately navigate back to the completed jobs list of the resource in question, you see the blinking status icon, but the status of the job never finalizes.  I think there are a couple of problems here and I'm not 100% whats going on but here is a guess on part of it: 
1)In `useWsJobs`, on mount, I think we need to check if we have any jobs that have `status: 'running'` and if so add those to the `JobsToFetch` array.
2) The job is now fetched properly, but it still is not updated in the list properly and I think that's because `PaginatedDataList` did not see a change and doesn't know to update

Just some thoughts.

awx/ui_next/src/components/JobList/useThrottle.test.jsx
```diff
@@ -0,0 +1,46 @@
...
+}
+function Test({ value }) {
+  const throttled = useThrottle(value, 500);
+  return <TestInner throttled={throttled} />;
+}
+
+jest.useFakeTimers();
+
+describe('useThrottle', () => {
+  test.skip('should throttle value', async () => {
```
AlexSCorey: did you intend to skip this test for some reason?

marshmalien:
list

When a job finishes and the sort order is -finished, the client needs to re-sort the jobs by their finished timestamp in descending order. In the recording above, the jobs updated through web sockets get added to the top of the list and stay in the same order regardless of finished time. There is more history about websocket list updates here: https://github.com/ansible/awx/pull/5968

awx/ui_next/src/components/JobList/useThrottle.test.jsx
keithjgrant: I wasn't able to get this test working, because it required stubbing out the Date object. I may have to delete it :/

keithjgrant:
> In pulling this down I see that if you launch a job, like a JT from the details view, it sends you to the job output. However, if you immediately navigate back to the completed jobs list of the resource in question, you see the blinking status icon, but the status of the job never finalizes. I think there are a couple of problems here and I'm not 100% whats going on but here is a guess on part of it:
> 1)In `useWsJobs`, on mount, I think we need to check if we have any jobs that have `status: 'running'` and if so add those to the `JobsToFetch` array.
> 2) The job is now fetched properly, but it still is not updated in the list properly and I think that's because `PaginatedDataList` did not see a change and doesn't know to update
> 
> Just some thoughts.

So the problem here is the useWsJobs hook assumed that the list was only completed jobs, so when a job completes it adds the job to the list — but this runs through a de-dupe function. Since the running job was already on the list from the initial load, a duplicate copy of the job was being added, then stripped off during de-duping. I updated this to run these completed jobs through the "add or update" logic instead of assuming it's not in the list yet.

The fact that the page initializes with still-running jobs in the "completed jobs" list is probably a bug (pre-existing), but I'd like to address that issue separate from this PR

awx/ui_next/src/components/JobList/useThrottle.test.jsx
AlexSCorey: Yeah, I'd rather have a deleted test, than a skipped test IMO

jakemcdermott:
I this one ready to merge? @marshmalien @AlexSCorey

jakemcdermott (approved):
lgtm, love the use of `useRef` for the ws object. @keithjgrant and I discussed a possible approach for testing that `useThrottle` hook but that's most likely a topic for a future pr.

AlexSCorey (approved):
You and I discussed the issues offline and you were able to fix all of my concerns. Good Work!

PR #7350, opened 2020-06-16
removing jinja2 warning in default venv

gamuniz:
##### SUMMARY

Removing the misleading jinja2 warning thrown by ansible-inventory command.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
make VERSION 
awx: 12.0.0

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/utils/common.py
+++ b/awx/main/utils/common.py
@@ -178,7 +178,7 @@ def _get_ansible_version(ansible_path):
     '''
     try:
         proc = subprocess.Popen([ansible_path, '--version'],
-                                stdout=subprocess.PIPE)
+                                stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         result = smart_str(proc.communicate()[0])
         return result.split('\n')[0].replace('ansible', '').strip()
     except Exception:


awx/main/utils/common.py
```diff
@@ -178,7 +178,7 @@ def _get_ansible_version(ansible_path):
     '''
     try:
         proc = subprocess.Popen([ansible_path, '--version'],
-                                stdout=subprocess.PIPE)
+                                stdout=subprocess.PIPE,sterr=subprocess.PIPE)
```
ryanpetrello: flake8 is probably gonna want a space after this `,`

ryanpetrello (approved):

ryanpetrello:
I want to test this a bit more - I'm a little concerned that by capturing stderr, we're going to fail parsing the ansible versions in some scenarios (i.e., is the version still on "line 0" of the combined stdout + stderr?)

ryanpetrello:
Yea, I don't know if this is the correct approach here - I'm worried about what other legitimate warnings something like this might cover up.

ryanpetrello (changes requested):
Talked some more to @gamuniz, I think I'm a 👎 on this approach.

PR #7351, opened 2020-06-16
Add section related to custom hooks

nixocio:
Add section related to custom hooks, and their usage.


awx/ui_next/CONTRIBUTING.md
```diff
@@ -243,6 +244,14 @@ About.defaultProps = {
 };
 ```
 
+### Custom Hooks
+
+There are currently a few custom hooks:
```
AlexSCorey, line 249: This is great! I do think we should provide the file path or file name to where the hooks are found.  Often there is more information about what the hooks do / what they return that would be helpful to the contributor.

AlexSCorey (changes requested):

AlexSCorey (approved):
LGTM

PR #6992, opened 2020-05-11
drastically optimize job host summary creation

ryanpetrello:
see: https://github.com/ansible/awx/issues/6991

PR #7352, opened 2020-06-16
fix a regression in how job host summaries are generated

ryanpetrello:
this change fixes a bug introduced in the optimization at https://github.com/ansible/awx/pull/6992

1. Create inventory with multiple hosts
2. Run a playbook with a limit to match only one host
3. Run job, verify that it only acts on the one host
4. Go to inventory host list and see that all the hosts have last_job updated to point to the job that only acted on one host.

### DIFF
--- a/awx/main/tests/functional/models/test_events.py
+++ b/awx/main/tests/functional/models/test_events.py
@@ -3,7 +3,7 @@ import pytest
 
 from django.utils.timezone import now
 
-from awx.main.models import Job, JobEvent, Inventory, Host
+from awx.main.models import Job, JobEvent, Inventory, Host, JobHostSummary
 
 
 @pytest.mark.django_db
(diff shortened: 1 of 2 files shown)


awx/main/tests/functional/models/test_events.py
```diff
@@ -153,3 +153,58 @@ def test_host_summary_generation_with_deleted_hosts():
...
+            'rescued': {},
+            'skipped': {},
+        },
+        host_map=host_map
+    ).save()
+
+    # since the playbook_on_stats only references one host,
+    # there should *only* be on JobHostSummary record (and it should
+    # be related to the appropriate Host)
+    assert JobHostSummary.objects.count() == 1
```
ryanpetrello, line 202: This assertion...

awx/main/tests/functional/models/test_events.py
```diff
@@ -153,3 +153,58 @@ def test_host_summary_generation_with_deleted_hosts():
...
+    for h in Host.objects.all():
+        if h.name == 'Host 1':
+            assert h.last_job_id == j.id
+            assert h.last_job_host_summary_id == JobHostSummary.objects.first().id
```
ryanpetrello, line 206: ...and this assertion are not _strictly_ addressed by the code changes in this patch (because the host summaries are generated properly before this patch, it's the "last job" reference that's behaving incorrectly), but I figured it was worth verifying while I was here.

awx/main/tests/functional/models/test_events.py
```diff
@@ -153,3 +153,58 @@ def test_host_summary_generation_with_deleted_hosts():
...
+        host_map=host_map
+    ).save()
+
+    # since the playbook_on_stats only references one host,
+    # there should *only* be on JobHostSummary record (and it should
+    # be related to the appropriate Host)
+    assert JobHostSummary.objects.count() == 1
+    for h in Host.objects.all():
+        if h.name == 'Host 1':
+            assert h.last_job_id == j.id
```
ryanpetrello, line 205: _This_ assertion is the point of the patch and reported issue.

AlanCoding (approved):
Oh that's a sneaky bug, once I read your comment about the host names in plabyook_on_stats events the picture became pretty clear to me.

Only thing that nags me for potential future bugs are cases where the hostname Ansible is given is transformed in the process of loading inventory. For example, automatically stripping whitespace, or converting special characters, or URL encoding stuff. If you went looking, you might be able to find something, but it would probably be addressed elsewhere.

kdelee (approved):
thanks for adding assertions for this in tests

ryanpetrello:
@AlanCoding @kdelee @elyezer 

Here's two playbook runs, second run has a limit pointed at one of the two hosts.

image

elyezer (approved):

ryanpetrello:
>  If you went looking, you might be able to find something, but it would probably be addressed elsewhere.

@AlanCoding I suspect that you're right, but I'd rather err on the side of technicalities causing us not to properly establish the Host --> Job link vs what we have now, which is "we're saving the wrong `.last_job_id` when you use `--limit`".

PR #7354, opened 2020-06-16
WIP: enforce bigint migrations on each task manager run (not process startup)

ryanpetrello:


ryanpetrello:
This addresses a race condition that's a little tricky to hit - I want to test this a bit to make sure it addresses the problem.

ryanpetrello:
I want to think more about this later.

PR #7359, opened 2020-06-16
Remove local port binding for postgres in dev env

shanemcd:
I dont think we need to bind this port at the host level. This will allow us to run the galaxy_ng dev tooling side-by-side with AWX.


ryanpetrello (approved):

chrismeyersfsu (approved):

chrismeyersfsu:
Update our `docker-compose-cluster.yml` to remove this port exposure also.

PR #7360, opened 2020-06-16
[foreman] add backwards support for ssl_verify in foreman plugin

jladdjr:
If you wanted to ensure certificates were validated when using the satellite / foreman inventory:
* the (old) script expected the `ssl_verify` source_var
* .. while the plugin expects `validate_certs`

This PR adds support for the old `ssl_verify` source_var in the plugin.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
awx 12.0.0
```

### DIFF
--- a/awx/main/models/inventory.py
+++ b/awx/main/models/inventory.py
@@ -2596,6 +2596,7 @@ class satellite6(PluginFileInjector):
 
     def inventory_as_dict(self, inventory_update, private_data_dir):
         ret = super(satellite6, self).inventory_as_dict(inventory_update, private_data_dir)
+        ret['validate_certs'] = False
 
         group_patterns = '[]'
         group_prefix = 'foreman_'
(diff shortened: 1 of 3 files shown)


awx/main/models/inventory.py
```diff
@@ -2596,6 +2596,7 @@ def get_plugin_env(self, inventory_update, private_data_dir, private_data_files)
 
     def inventory_as_dict(self, inventory_update, private_data_dir):
         ret = super(satellite6, self).inventory_as_dict(inventory_update, private_data_dir)
+        ret['validate_certs'] = False
```
AlanCoding, line 2599: I don't think we _need_ this line, and I'd rather leave it to the plugin's default in this case.

AlanCoding (approved):

awx/main/models/inventory.py
jladdjr: Okay - I thought that since we're advertising the settings in https://github.com/ansible-collections/community.general/blob/master/scripts/inventory/foreman.ini#L124, that it made sense to pin `validate_certs` to the static value in that file (in case things _do_ change in the plugin).

Then again, I can also see it making sense to follow the plug-in's behavior (and am happy to go either way).

Feel like we're stuck in this weird limbo right now where we want to support backwards compatiblity, but at some point it seems like we'll want to move on and adopt what's in the plugin without managing it / locking it down too much.

PR #7367, opened 2020-06-17
[WIP] Replace shallow clone with shutil copy with .git exclusion

AlanCoding:
##### SUMMARY
Experimental idea for some tech debt cleanup / simplification.

##### COMPONENT NAME
 - API

##### AWX VERSION
```
12.0.0
```

### DIFF
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -16,12 +16,10 @@ import stat
 import tempfile
 import time
 import traceback
-from distutils.dir_util import copy_tree
 from distutils.version import LooseVersion as Version
 import yaml
 import fcntl
 from pathlib import Path
-from uuid import uuid4
 import urllib.parse as urlparse
 
 # Django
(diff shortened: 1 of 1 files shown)


ryanpetrello:
@AlanCoding do we still need this?  I see it referenced in https://github.com/ansible/awx/pull/7643

AlanCoding:
no, it's not a part of that, but the latest idea I was throwing around was to expand this, using the new caching mechanisms, and thus remove the `GitPython` dependency entirely.

AlanCoding:
This has now collected 4 issues, which could probably all be consolidated into a single work item - to update and complete this.

There were significant conflicts introduced with the merge of https://github.com/ansible/awx/pull/7643, but those are generally moving in the same code design pattern. Nonetheless, the conflict is so large I think one would just have to start over and redo this.

AlanCoding:
Closing because this is tracked in an issue now.

Issue #7327, opened 2020-06-11
Delete Custom Credential Types


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Delete Custom Credential Types

nixocio:
Closing this one - Since related PR was already merged.

Issue #7324, opened 2020-06-11
List Custom Credential Type


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Add support for list Custom Credential Types

nixocio:
Closing this one - Since related PR was already merged.

PR #7369, opened 2020-06-17
Add Credential Type List and Delete

nixocio:
Add `Credential Type` List and Delete features.

See: #7324
Also:#7327

list_delete_credential_types

### DIFF
--- a/awx/ui_next/src/screens/Application/ApplicationsList/ApplicationList.test.jsx
+++ b/awx/ui_next/src/screens/Application/ApplicationsList/ApplicationList.test.jsx
@@ -21,7 +21,7 @@ const applications = {
           user_capabilities: { edit: true, delete: true },
         },
         url: '',
-        organiation: 10,
+        organization: 10,
       },
       {
         id: 2,
--- a/awx/ui_next/src/screens/CredentialType/CredentialTypeList/CredentialTypeListItem.jsx
+++ b/awx/ui_next/src/screens/CredentialType/CredentialTypeList/CredentialTypeListItem.jsx
@@ -1,0 +1,92 @@
+import React from 'react';
+import { string, bool, func } from 'prop-types';
+import { withI18n } from '@lingui/react';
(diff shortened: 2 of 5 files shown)


awx/ui_next/src/screens/CredentialType/CredentialTypeList/CredentialTypeListItem.jsx
```diff
@@ -0,0 +1,94 @@
...
+              </Link>
+            </DataListCell>,
+          ]}
+        />
+        <DataListAction
+          aria-label="actions"
+          aria-labelledby={labelId}
+          id={labelId}
+        >
+          {credentialType.summary_fields.user_capabilities.edit ? (
```
mabashian: This is a nit-pick (and I can tell this pattern came from somewhere else because I see it on other lists) but I feel like a conditional render `&&` would be better here than a ternary with an empty string.  So you could do:

```
          {credentialType.summary_fields.user_capabilities.edit && (
            <Tooltip content={i18n._(t`Edit credential type`)} position="top">
              <Button
                aria-label={i18n._(t`Edit credential type`)}
                variant="plain"
                component={Link}
                to={`/credential_types/${credentialType.id}/edit`}
              >
                <PencilAltIcon />
              </Button>
            </Tooltip>
          )}
```

I thought maybe the empty string was necessary for spacing of some kind but I tested it out locally and I don't think so.

awx/ui_next/src/screens/Application/ApplicationsList/ApplicationList.test.jsx
```diff
@@ -21,7 +21,7 @@ const applications = {
           user_capabilities: { edit: true, delete: true },
         },
         url: '',
-        organiation: 10,
```
mabashian, line 24: Nice

mabashian (approved):

dsesami (approved):

PR #7375, opened 2020-06-18
Use consistent layout for subforms in job/template forms

### LINKED ISSUE #5929: make checkbox fields consistent in ui_next
there are some inconsistencies between the checkbox groups on the project and template forms

project

Screen Shot 2020-02-13 at 10 58 17 AM

template

Screen Shot 2020-02-13 at 10 58 06 AM

- checkbox fields should put options into 3 columns (like templates)
- any dynamic fields based on checkbox options should be in gray/inset subform (like projects), and should have their fields in the standard 3 columns.
- seems like whatever fields are supposed to show up on proejct form when update revision on launch is checked are not.

vjanssens:
##### SUMMARY
This PR resolves #5929 by moving the related form fields in job template create/edit and workflow template create/edit forms inside a subform. This changes the layout to be consistent with the rest of the application.

I could not reproduce some of the issues mentioned in #5929, so I guess they are alread resolved.

- checkbox fields should put options into 3 columns (like templates) -> was already resolved
- any dynamic fields based on checkbox options should be in gray/inset subform (like projects), and should have their fields in the standard 3 columns. -> fixed by this PR
- seems like whatever fields are supposed to show up on project form when update revision on launch is checked are not. -> was already resolved

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
UI

##### AWX VERSION

```
12.0.0
```

##### ADDITIONAL INFORMATION

Job template form before:
5929-job-template-before

Job template form after:
5929-job-template-after

Workflow template form before
5929-workflow-template-before

Workflow template form after:
5929-workflow-template-after

### DIFF
--- a/awx/ui_next/src/screens/Template/shared/JobTemplateForm.jsx
+++ b/awx/ui_next/src/screens/Template/shared/JobTemplateForm.jsx
@@ -583,16 +583,15 @@ function JobTemplateForm({
                     </>
                   )}
 
-                  {allowCallbacks && enableWebhooks && (<br />)}
+                  {allowCallbacks && enableWebhooks && <br />}
 
                   {enableWebhooks && (
                     <>
-                      <Title size="md" headingLevel="h4">{i18n._(t`Webhook details`)}</Title>
+                      <Title size="md" headingLevel="h4">
+                        {i18n._(t`Webhook details`)}
+                      </Title>
                       <FormColumnLayout>
-                        <WebhookSubForm
-                          enableWebhooks={enableWebhooks}
-                          templateType={template.type}
-                        />
+                        <WebhookSubForm templateType={template.type} />
                       </FormColumnLayout>
                     </>
                   )}
--- a/awx/ui_next/src/screens/Template/shared/JobTemplateForm.jsx
+++ b/awx/ui_next/src/screens/Template/shared/JobTemplateForm.jsx
@@ -555,7 +555,7 @@ function JobTemplateForm({
                 <SubFormLayout>
                   {allowCallbacks && (
                     <>
-                      <Title size="md">
+                      <Title size="md" headingLevel="h4">
                         {i18n._(t`Provisioning Callback details`)}
                       </Title>
                       <FormColumnLayout>
@@ -580,13 +580,14 @@ function JobTemplateForm({
                           }
                         />
                       </FormColumnLayout>
-
-                      <br />
                     </>
                   )}
+
+                  {allowCallbacks && enableWebhooks && (<br />)}
+
                   {enableWebhooks && (
                     <>
-                      <Title size="md">{i18n._(t`Webhook details`)}</Title>
+                      <Title size="md" headingLevel="h4">{i18n._(t`Webhook details`)}</Title>
                       <FormColumnLayout>
                         <WebhookSubForm
                           enableWebhooks={enableWebhooks}
--- a/awx/ui_next/src/screens/Template/shared/JobTemplateForm.jsx
+++ b/awx/ui_next/src/screens/Template/shared/JobTemplateForm.jsx
@@ -9,6 +9,7 @@ import {
   Switch,
   Checkbox,
   TextInput,
+  Title,
 } from '@patternfly/react-core';
 import ContentError from '../../../components/ContentError';
 import ContentLoading from '../../../components/ContentLoading';
@@ -27,6 +28,7 @@ import {
   FormColumnLayout,
   FormFullWidthLayout,
   FormCheckboxLayout,
+  SubFormLayout,
 } from '../../../components/FormLayout';
 import { VariablesField } from '../../../components/CodeMirrorInput';
 import { required } from '../../../util/validators';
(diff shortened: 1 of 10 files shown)


awx/ui_next/src/screens/Template/shared/WorkflowJobTemplateForm.jsx
```diff
@@ -175,39 +182,50 @@ function WorkflowJobTemplateForm({
...
+            tooltip={i18n._(
+              t`If enabled, simultaneous runs of this workflow job template will be allowed.`
+            )}
+            label={i18n._(t`Enable Concurrent Jobs`)}
+          />
+        </FormCheckboxLayout>
+      </FormGroup>
+
+      {enableWebhooks && (
+        <>
```
jlmitch5: you can get rid of the <> fragment cotainer here, only needed for when there are > 1 child.

awx/ui_next/src/screens/Template/shared/WorkflowJobTemplateForm.jsx
```diff
@@ -175,39 +182,50 @@ function WorkflowJobTemplateForm({
...
+          <SubFormLayout>
+            <Title size="md">{i18n._(t`Webhook details`)}</Title>
```
jlmitch5: can you add headingLevel="4" prop to Title here?  It's required in the patternfly v4 upgrade that will be merging shortly.

awx/ui_next/src/screens/Template/shared/JobTemplateForm.jsx
```diff
@@ -543,30 +545,53 @@ function JobTemplateForm({
...
+                          id="template-host-config-key"
+                          name="host_config_key"
+                          label={i18n._(t`Host Config Key`)}
+                          validate={
+                            allowCallbacks ? required(null, i18n) : null
+                          }
+                        />
+                      </FormColumnLayout>
+
+                      <br />
```
jlmitch5: can you move the br out and put it between the two subforms?

```{allowCallbacks && enableWebhooks && (<br />)}```

**Review by jlmitch5 — CHANGES_REQUESTED** [review]
looking good, thanks for the contribution!  a few little things

**Comment by vjanssens** [comment]
Thanks @jlmitch5 for the review, i've created a commit to fix the issues. 

Unfortunately, there's a test failing which I'm unable to fix. 
```
Warning: An update to WebhookSubForm inside a test was not wrapped in act(...).
When testing, code that causes React state updates should be wrapped into act(...):

<WorkflowJobTemplateForm/> › webhooks and enable concurrent jobs functions properly

    Method “props” is meant to be run on 1 node. 0 found instead.

      207 |     ).toBe(true);
      208 |     expect(
    > 209 |       wrapper.find('input[aria-label="wfjt-webhook-key"]').prop('readOnly')
          |                                                            ^
      210 |     ).toBe(true);
      211 |     expect(
      212 |       wrapper.find('input[aria-label="wfjt-webhook-key"]').prop('value')
```

This is my first time using react, but this error seems a bit vague. I haven't touched code inside WebookSubForm, why would it be changed? Can you point me in the right direction?
When I check the checkbox manually, I can see the `wfjt-webhook-key` be visible.

mabashian:
@vjanssens I see a couple of failing tests.  One of them has to do with webhook_key and webhook_url not matching our expected values on form submission.  I took a look at this and those fields are not even valid when we POST/PUT on `/api/v2/job_templates` so I don't think they should be included in the payload.  I would fix that test with the following changes:

https://github.com/mabashian/awx/commit/56ea5aa9070962d77edbd8bdef9c0da1a3ccb042

The other failing test is a bit more complicated.  I stopped the test right before the failing line and took a look at the DOM.  The `ContentLoading` component is being shown inside the subform which means that isLoading is still `true` when we try to access the field in the test.  While `ContentLoading` is shown, those fields won't be available and the test will fail.  The way that webhook subform worked before was that it was mounted as soon as the form mounted.  Now that you've wrapped it in a conditional render I think things are getting a bit sideways.  We may have to rethink the approach of the subform component in order to get this working smoothly.  For example, if we're going to mount the component when `enableWebhooks` is true then we don't need that same check inside the webhook subform component.  It will always be true if the component is mounted.

vjanssens:
Thank you @mabashian. I've applied the fixes you proposed and removed the `enableWebhooks` conditional inside the WebhookSubForm, as I agree that double check was a bit strange.
I don't know the reasoning for why that state was passed to the component, but to me it is more straight forward to let the parent decide if it should be shown.

Fixed one of the tests, i'll need some more time to see if I can fix the other as well.

vjanssens:
recheck

vjanssens:
@mabashian I think I fixed the issue, I don't have any failing tests locally right now. However, the build seems to fail on something unrelated to my changes. Would you mind to take a look?

vjanssens:
Still not sure why the build fails. It does not seem to be related to my changes.

jlmitch5:
@vjanssens I just force pushed your branch rebased on the latest devel branch (no other changes)...I assume something was just messed up on the check.  I pulled down the code and all tests pass and the UI looks good.  I assume it was something about how we've just merged a PR related to the UI packaging.

I'm putting the mergeit label on there, so provided the checks pass, it will merge.

Thanks for the contribution!

jlmitch5 (approved):
looks good!

Issue #7376, opened 2020-06-18
API allows setting a credential owner to a user that is not part of the credential organization


elyezer:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

When creating a credential the following check is not done https://github.com/ansible/awx/blob/devel/awx/api/views/__init__.py#L1094-L1097. So one can go ahead and have a owner that is not part of the credential organization.

##### ENVIRONMENT
* AWX version: current devel
* AWX install method: docker on linux
* Ansible version:  X.Y.Z
* Operating System:
* Web Browser:

##### STEPS TO REPRODUCE

1) Create two organizations A and B (id 1 and 2 respectively)
2) Create a user and associate it with org B (id 2)
3) Perform a POST request to `/api/v2/credentials/` with the following content:

```json
{
    "name": "Credential on Org A",
    "description": "",
    "organization": 1,
    "credential_type": 1,
    "inputs": {},
    "user": 2,
    "team": null
}
```

Note: `credential_type` 1 translates to SSH/Machine credential.

##### EXPECTED RESULTS

The credential should not be created and as the error a similar message as the one provided here https://github.com/ansible/awx/blob/devel/awx/api/views/__init__.py#L1094-L1097 should be returned.

##### ACTUAL RESULTS

The credential is created and the owner is a user from an org different from the credential's org

##### ADDITIONAL INFORMATION

Performing a GET request at `/api/v2/organizations/2/credentials/` returns:

```
{"count":0,"next":null,"previous":null,"results":[]}
```

one-t:
This is now working as expected. PR for new test : https://github.com/ansible/tower-qa/pull/5295

PR #7381, opened 2020-06-18
Enforce single owner field when serializing creds

elyezer:
##### SUMMARY

The CredentialSerializerCreate expect a single owner field according to
its help text but was not validating that. This makes it validate for a
single owner field when creating a Credential.

Related #7376 

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 12.0.0
```

### DIFF
--- a/awx/main/tests/functional/api/test_credential.py
+++ b/awx/main/tests/functional/api/test_credential.py
@@ -61,6 +61,36 @@ def test_credential_validation_error_with_bad_user(post, admin, credentialtype_s
 
 
 @pytest.mark.django_db
+def test_credential_validation_error_with_no_owner_field(post, admin, credentialtype_ssh):
+    params = {
+        'credential_type': credentialtype_ssh.id,
+        'inputs': {'username': 'someusername'},
+        'name': 'Some name',
+    }
+    response = post(reverse('api:credential_list'), params, admin)
+    assert response.status_code == 400
+    assert response.data['detail'][0] == "Missing 'user', 'team', or 'organization'."
+
+
+@pytest.mark.django_db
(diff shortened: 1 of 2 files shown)


elyezer:
I was looking for adding tests for this, but I don't see any unit tests for the credential serializer. I will try to get some in place.

AlanCoding:
the validators are often indirectly tested via the unit tests that use mock requests. These tend to be around `awx/main/tests/functional/api/` but are scattered through more places.

elyezer:
I've updated with tests for some validation. I am not able to run locally so I will wait for PR verification and will update accordingly.

awx/main/tests/functional/api/test_credential.py
```diff
@@ -60,6 +60,36 @@ def test_credential_validation_error_with_bad_user(post, admin, credentialtype_s
...
+@pytest.mark.django_db
+def test_credential_validation_error_with_no_owner_field(post, admin):
+    params = {
+        'credential_type': 1,
+        'inputs': {'username': 'someusername'},
+        'name': 'Some name',
+    }
+    response = post(reverse('api:credential_list'), params, admin)
+    assert response.status_code == 400
+    assert response.data['detail'] == "Missing 'user', 'team', or 'organization'."
```
elyezer: Not sure how to access the detail information. Any help is appreciated.

I am trying to get my local dev env working so I can test locally.

AlanCoding: it might be in some other key like "msg" but it's hard to know without running and just printing out `response.data`

elyezer:
I finally got the tests passing, not being able to run them locally was really bad but I finally have everything in place.

AlanCoding (approved):

ryanpetrello (approved):

PR #7392, opened 2020-06-19
unpin collections used for inventory plugins

jladdjr:
Unpins collections in requirements/collections_requirements.yml

##### SUMMARY
Chatted with @kdelee about the idea of:

* leaving inventory plugin requirements unpinned in awx
* .. unless we are aware of an issue that needs to be resolved in a collection

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION

```
12.0.0
```

##### ADDITIONAL INFORMATION

If we continue to pin all inventory plugins, then we'll be constantly revving versions in https://github.com/ansible/awx/blob/devel/requirements/collections_requirements.yml. Seems like this churn may not be necessary.


jladdjr:
@kdelee @AlanCoding - have tested out these changes and not seeing any regressions. Can you all think of anything else we should do before merging?

kdelee:
:shipit:

kdelee (approved):

requirements/collections_requirements.yml
```diff
@@ -1,22 +1,12 @@
...
   - name: azure.azcollection
-    version: 0.1.2
   - name: amazon.aws
-    version: 0.1.1  # version 0.1.0 seems to have gone missing
   - name: theforeman.foreman
-    version: 0.8.1
   - name: google.cloud
-    version: 0.0.9  # contains PR 167, should be good to go
   - name: openstack.cloud
-    version: 0.0.1-dev85  # earlier had checksum mismatch
```
AlanCoding, line 14: question I would have asked is if this works with the `-dev` versions, but since this was created it looks like openstack.cloud has released a non-dev version number, so that should no longer be a stumbling block.

jladdjr: oo, good catch, and thanks for checking on the collection 👍

Issue #6842, opened 2020-04-27
"Playbook not found for project" error when the playbook exist but is invalid 


zaskan:
##### ISSUE TYPE
When a playbook is not valid but its filename is specified using a rest api enpoint, (e.g using job_templates endpoint) the system returns "Playbook not found for project".

##### SUMMARY
Error "Playbook not found for project", suggest users that yaml file doesn't exist at specified playbook path. In this particular situation error message like "Playbook specified doesn't contain any plays" or "Playbook specified is not a valid playbook" would clearly suggest user where problem lies.

cNeha:
Hello,  I would like to work on this request though I have noticed that PR #7401  is in closed state due to outdated code. Can I continue working on this request?

kjamaal:
Hi, wanted to work on this issue but there are dependency issues in devel that break tests. Posted about it in forum here https://forum.ansible.com/t/streamlining-awx-releases/6894/46?u=kjamaal. Should we leave this project alone until the re-architecture is done?

PR #7401, opened 2020-06-22
Created New Error Messages for Invalid and Non Yaml Playbook Files

humzam96:
##### SUMMARY
Upon Project Update a playbooks list was updated and only includes files that were valid. We added 2 new lists for invalid playbooks (playbooks that fail to match the regular expression in awx/main/utils/ansible.py) and non yaml files. This allows for more specific validation errors when validating in JobOptionsSerializer in awx/api/serializers.py. We tried looking for a solution to avoid creating a new migration, if there is a different suggestion we can try that too.

related #6842 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
awx: 12.0.0
```


awx/main/models/projects.py
```diff
@@ -431,13 +489,16 @@ def notification_templates(self):
         if self.organization is not None:
             error_notification_templates = set(error_notification_templates +
                                                list(base_notification_templates
-                                                    .filter(organization_notification_templates_for_errors=self.organization)))
+                                                    .filter(
```
wenottingham: There's a bunch of unrelated formatting changes in this PR. Please remove them.

humzam96: Thanks, for some reason those changes were recommended when tested on a local build, I just undid them and the Zuul tests passed.

humzam96:
Do you have any recommendations for updating this solution if it's not something that can be accepted?

awx/main/migrations/0118_v400_invalid_playbooks.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 2.2.11 on 2020-06-04 03:55
```
ryanpetrello, line 1: This migration is out of date and needs to be regenerated to point at the latest.

ryanpetrello:
Hey @humzam96,

This PR seems considerably out of date given code and database schema changes over the past 9 months.  If you're still interested in contributing this, please open another PR.

Thanks.

PR #7408, opened 2020-06-22
Add parameter update_password to tower_user module

koleo:
##### SUMMARY

Add a parameter **update_password** to **tower_user** module.

Available values are **always** (default) or **on_create**.

- When update_password=on_create, password is set only if a new user is created.
- When update_password=always, password is set for new and existing users (if password is provided). That's the way it works, currently.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - Collection

##### AWX VERSION

```
awx: 12.0.0
```

##### ADDITIONAL INFORMATION

This new parameter is particularly useful when you want to generate random passwords for new users but do not want to change passwords for any existing users.

The implementation is based on the System User module: https://docs.ansible.com/ansible/latest/modules/user_module.html#parameter-update_password

### DIFF
--- a/awx_collection/plugins/modules/tower_user.py
+++ b/awx_collection/plugins/modules/tower_user.py
@@ -55,6 +55,13 @@ options:
       description:
         - Write-only field used to change the password.
       type: str
+    update_password:
+      description:
+        - C(always) will update passwords if they differ.
+        - C(on_create) will only set the password for newly created users.
+      type: str
+      choices: [ always, on_create ]
+      default: always
     state:
       description:
         - Desired state of the resource.
@@ -115,6 +122,7 @@ def main():
         is_superuser=dict(type='bool', default=False, aliases=['superuser']),
         is_system_auditor=dict(type='bool', default=False, aliases=['auditor']),
         password=dict(no_log=True),
+        update_password=dict(choices=['always', 'on_create'], default='always', no_log=False),
         state=dict(choices=['present', 'absent'], default='present'),
     )
 
@@ -129,6 +137,7 @@ def main():
     is_superuser = module.params.get('is_superuser')
     is_system_auditor = module.params.get('is_system_auditor')
     password = module.params.get('password')
+    update_password = module.params.get('update_password')
     state = module.params.get('state')
 
     # Attempt to look up the related items the user specified (these will fail the module if not found)
@@ -158,7 +167,7 @@ def main():
         new_fields['is_superuser'] = is_superuser
     if is_system_auditor:
         new_fields['is_system_auditor'] = is_system_auditor
-    if password:
+    if password and (not existing_item or update_password == 'always'):
         new_fields['password'] = password
 
     # If the state was present and we can let the module build or update the existing item, this will return on its own
--- a/awx_collection/test/awx/test_user.py
+++ b/awx_collection/test/awx/test_user.py
@@ -44,3 +44,16 @@ def test_password_no_op_warning(run_module, admin_user, mock_auth_stuff, silence
     silence_warning.assert_called_once_with(
         "The field password of user {0} has encrypted data and "
         "may inaccurately report task is changed.".format(result['id']))
+
+
+@pytest.mark.django_db
+def test_update_password_on_create(run_module, admin_user, mock_auth_stuff):
+    for i in range(2):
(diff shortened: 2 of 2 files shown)


awx_collection/plugins/modules/tower_user.py
```diff
@@ -8,7 +8,7 @@
 __metaclass__ = type
 
 
-ANSIBLE_METADATA = {'metadata_version': '1.1',
+ANSIBLE_METADATA = {'metadata_version': '1.2',
```
AlanCoding: I don't think this is the right way to do this. The version_added should reference the collection version and not the metadata version too.

awx_collection/plugins/modules/tower_user.py
```diff
@@ -55,6 +55,14 @@
       description:
         - Write-only field used to change the password.
       type: str
+    update_password:
```
AlanCoding, line 58: I'm all in favor of getting this merged, but to be transparent I will state my only remaining hesitation.

We have other modules which take secrets, like `tower_credential` will accept secrets inside of the `inputs` field. It's possible that we could move this logic to the shared code in module_utils. For that, it would make more sense to have a general setting like `update_secrets` so there wouldn't be different parameters for each secret-ish field.

We can change the implementation later, but we can't rename the parameter as easily.

awx_collection/plugins/modules/tower_user.py
koleo: Okay, fixed, thanks.

awx_collection/plugins/modules/tower_user.py
koleo: IMHO an `update_secrets` parameter would only make sense if we rename `password` to `secret`. Once the `tower_user` module implements a `password` parameter, I think the best approach is to name this new one `update_password`.

As an example, [`user`](https://docs.ansible.com/ansible/latest/modules/user_module.html#parameter-password) module implements the parameters `password`/`update_password` and [`win_credential`](https://docs.ansible.com/ansible/latest/modules/win_credential_module.html#parameter-secret) implements the parameters `secret`/`update_secret`. But they all remain consistent in their names. 

As for whether to move it to module_utils, this can always be done later, if necessary.

awx_collection/plugins/modules/tower_user.py
```diff
@@ -158,7 +167,7 @@ def main():
         new_fields['is_superuser'] = is_superuser
     if is_system_auditor:
         new_fields['is_system_auditor'] = is_system_auditor
-    if password:
+    if password and (not existing_item or update_password == 'always'):
```
john-westcott-iv, line 170: Do we want to have an elif condition that calls module.warn() to have Ansible display a warning indicating that it is choosing to not set the password if this condition is set to ignore it and password was actually set?

beeankha (approved):
This looks OK to me!

awx_collection/plugins/modules/tower_user.py
koleo: For the use case I described in the PR comment, I would always set a (default or random) password in my playbook’s task as I can’t assume if user exists or not.

If we implement the log warn as you suggest, I will always get a warning in such a situation, which seems excessive...

AlanCoding:
Replying mostly to:

https://github.com/ansible/awx/pull/7408#discussion_r448494849

The main argument there is that the naming of the parameter here is desirable. I don't disagree.

I have put up a modified version of this at https://github.com/ansible/awx/pull/8087. This does the same thing via a parameter `update_secrets` as opposed to `update_password`. I completely agree that this is naming is _less good_. However, I would like to make a tradeoff to use the less obvious naming so that we have something more maintainable.

Taking logic added to a single module and migrating it to the shared module_utils is absolutely not trivial. If different modules wish to make use of a shared utility, but reference it by different names in different cases, then that would create additional overhead. We even have tests to enforce feature completeness, and these would require special exceptions for each case.

We have had a _strong_ push to keep these modules as generalized as possible, @john-westcott-iv has done wonders to this goal. Putting in non-API parameters to specific modules for specific purposes would be a step backwards for these goals, and people like John have invested a lot into development good designs.

AlanCoding:
I used your original commit, but put it in a new PR with `update_secrets`

PR #7413, opened 2020-06-22
Update awxkit credential page create_payload

elyezer:
##### SUMMARY
Only assign an organization if none of user, team or organization is
passed. When creating a credential just one of those fields are expected
and now that https://github.com/ansible/awx/pull/7381 is merged, that is
being enforced.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

awxkit

##### AWX VERSION

```
awx: 12.0.0
```

PS.: I am currently testing these changes

### DIFF
--- a/awxkit/awxkit/api/pages/credentials.py
+++ b/awxkit/awxkit/api/pages/credentials.py
@@ -287,7 +287,7 @@ class Credential(HasCopy, HasCreate, base.Base):
 
         credential_type, organization, user, team = filter_by_class(
             (credential_type, CredentialType), (organization, Organization), (user, User), (team, Team))
-        if not any((user, team, organization)):
+        if not all((user, team, organization)):
             organization = Organization
         self.create_and_update_dependencies(
             credential_type, organization, user, team)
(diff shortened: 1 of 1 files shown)


awxkit/awxkit/api/pages/credentials.py
```diff
@@ -323,7 +323,7 @@ def create(
         return self.update_identity(
             Credentials(
                 self.connection)).post(payload)
-
```
elyezer, line 326: I can drop these changes, my edit got the remove empty spaces at the end of the line when saving.

elyezer:
I've done some testing and this is ready for review

elyezer:
After more investigation I realized this is not needed and the changes will actually break the expected behavior. That said I am going to close this.

awxkit/awxkit/api/pages/credentials.py
```diff
@@ -287,7 +287,7 @@ def create_payload(
 
         credential_type, organization, user, team = filter_by_class(
             (credential_type, CredentialType), (organization, Organization), (user, User), (team, Team))
-        if not any((user, team, organization)):
+        if not all((user, team, organization)):
```
kdelee, line 290: I wonder if you want to check for each one individually.

If you have a user, get the org from it.

If you have a team, get the org from it,

If you have an org...use it.

PR #7420, opened 2020-06-23
Update awxkit credential creation

elyezer:
##### SUMMARY
Does not have an organization by default. Let `create_payload` decide if
it should be automatically created or not. This will avoid having more
than one owner when either user or team is passed to `create`.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - awxkit

##### AWX VERSION

```
devel
```

### DIFF
--- a/awxkit/awxkit/api/pages/credentials.py
+++ b/awxkit/awxkit/api/pages/credentials.py
@@ -310,7 +310,7 @@ class Credential(HasCopy, HasCreate, base.Base):
             credential_type=CredentialType,
             user=None,
             team=None,
-            organization=Organization,
+            organization=None,
             inputs=None,
             **kwargs):
         payload = self.create_payload(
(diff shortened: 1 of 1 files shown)


awxkit/awxkit/api/pages/credentials.py
```diff
@@ -310,7 +310,7 @@ def create(
             credential_type=CredentialType,
             user=None,
             team=None,
-            organization=Organization,
+            organization=None,
             inputs=None,
             **kwargs):
         payload = self.create_payload(
```
elyezer, line 316: `create_payload` will take care to provide the default Organization in case any of user, team or organization is provided. This is more to avoid having organization and user or team defined at the same time in case someone calls `create(user=user)` or `create(team=team)`

ryanpetrello (approved):

PR #7425, opened 2020-06-23
Don't traceback on KeyboardInterrupt

ghjm:
##### SUMMARY

Don't print a Python traceback when a user is running the awx CLI and presses Control-C.  Just exit.

##### ISSUE TYPE
 - Bughancement Pull Request

##### COMPONENT NAME
 - awxkit CLI

##### AWX VERSION
```
awx: 13.0.0
```


jbradberry (approved):
seems reasonable to me

Issue #7419, opened 2020-06-23
User Tokens List


AlexSCorey:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
We need to add the User Tokens List.  Based on old ui it seems that the user can only add tokens for themselves.  They can't add tokens for other users

https://tower-mockups.testing.ansible.com/patternfly/users/users-tokens/

PR #7436, opened 2020-06-24
Adds Tokens List

AlexSCorey:
##### SUMMARY
This addresses #7419.  This is the first PR to bring in Tokens to the new project.  The mock ups show a sync button, but there is no sync functionality, or sync button on this list on the old ui. Couple of things to consider about this list:
1)  Whether the tokens can be edited.  They can't in the old ui
2) How/whether to show a details view.  
3) This PR does not include add or edit functionality of a token.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI
##### ADDITIONAL INFORMATION
Screen Shot 2020-06-24 at 1 09 27 PM


awx/ui_next/src/screens/User/User.jsx
```diff
@@ -69,9 +69,16 @@ function User({ i18n, setBreadcrumb }) {
     },
     { name: i18n._(t`Teams`), link: `${match.url}/teams`, id: 2 },
     { name: i18n._(t`Access`), link: `${match.url}/access`, id: 3 },
-    { name: i18n._(t`Tokens`), link: `${match.url}/tokens`, id: 4 },
   ];
 
+  if (me?.id === Number(match.params.id)) {
```
jlmitch5, line 74: I would use Config here rather than pass down with props.

jlmitch5 (approved):
looks great!

dsesami (changes requested):
Currently, there's a bug if the token has an application associated with it.

1. In the old UI, make an application. Create a token, and then add the application to that token.
2. Visit the tokens list for the user in the new UI.

Screenshot from 2020-06-29 16-25-30

dsesami:
we've confirmed that the above issue is because applications are marked as mandatory in the new UI when they are actually optional. setting it to optional should fix this.

dsesami (approved):

PR #7437, opened 2020-06-24
Picky Openshift clusters need some more options

kdelee:
If a user does not have right privileges to the cluster, they cannot
install awx with these settings. Make sure we allow them to configure them
the way it works for their cluster.

These are blocking us installing awx on an internal openshift cluster where I have limited privileges.

Goal is no behavior change if don't change defaults. `awx_psp_privileged` and `task_security_context_privileged` are default `true` so `Add privileged SCC to service account` will continue to run by default and I set the defaults for the new vars to values that cause them to stll do the same thing if nothing modified:

```
awx_deployment_security_context_enabled: true
awx_deployment_fsgroup: 0
```

### DIFF
--- a/installer/roles/kubernetes/defaults/main.yml
+++ b/installer/roles/kubernetes/defaults/main.yml
@@ -14,6 +14,9 @@ kubernetes_awx_image: "{{ tower_package_name | default('ansible/awx') }}"
 kubernetes_web_image: "{{ kubernetes_awx_image }}"
 kubernetes_task_image: "{{ kubernetes_awx_image }}"
 
+awx_deployment_security_context_enabled: true
+awx_deployment_fsgroup: 0
+
 awx_psp_create: false
 awx_psp_name: 'awx'
 awx_psp_privileged: true
--- a/installer/roles/kubernetes/tasks/openshift.yml
+++ b/installer/roles/kubernetes/tasks/openshift.yml
@@ -32,6 +32,7 @@
 - name: Add privileged SCC to service account
   shell: |
     {{ openshift_oc_bin }} adm policy add-scc-to-user privileged system:serviceaccount:{{ openshift_project }}:awx
+  when: (awx_psp_privileged or web_security_context_privileged or task_security_context_privileged or redis_security_context_privileged) | bool
 
 # https://github.com/openshift/origin/issues/19182#issuecomment-378233606
 # If oc version ever grows a -o json option, remove the following tasks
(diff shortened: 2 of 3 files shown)


ryanpetrello:
@kdelee are you all intending to remove `fsGroup: 0` in your builds?

Doing so might break redis permissions in a way that its socket can't be accessed across container bounds (because the group won't match):

https://github.com/ansible/awx/commit/423df6618d1f5484ea428dcd32371e67bb4c49d4

cc @shanemcd

shanemcd:
Yeah... does Redis even work without this?

ryanpetrello:
What @shanemcd said - I'd verify that what you're doing here actually works - i.e., playbooks run, and you can see their streaming output across the UI of different pods.

kdelee:
ah ok well I've only gotten so far as "can I even install it this way", will test it more but at the moment I'm facing resource constraints on that cluster that are making tower really drag (even though I have a quota of 8Gi of memory they have placed a LimitRange that an individual pod can only have 2Gi of memory). So until I fix that can't really say how tower will actually perform after installed w/ the `fsGroup` attribute not set.

will investigate more.

essentially trying to work around an error where containers can never come up: get this message from openshift:

```
Error creating: pods "awx-7d47d79b4-" is forbidden: unable to validate against any security context constraint: [fsGroup: Invalid value: []int64{0}: 0 is not an allowed group]
```

kdelee:
Trying to read up on this `fsGroup` parameter.

Reading https://kubernetes.io/docs/tasks/configure-pod-container/security-context/ ...

> fsGroup: Volumes that support ownership management are modified to be owned and writable by the GID specified in fsGroup. See the Ownership Management design document for more details.

So if I understand this, it is making the volumes owned by user w/ GID=0, e.g. root.

I think this cluster prevents non-admin users from setting such permissions. Seems like this gets us back into the same pickle that prevents people from being able to use tower on openshift because of the privileged containers thing.

I wonder if there is a different GID I could use.

What all users need access to this volume? Can we make a group they belong to and use that GID?

chrismeyersfsu:
I _think_, if we took the time, we could make the redis container run as the same uid/gid as the AWX containers and then wouldn't need `fsGroup`.

shanemcd:
I think you would need to use `runAsUser`, which would also throw an error on a super locked down cluster.

kdelee:
closing this as not making any progress on it, thanks for insight about how it would not work!

Issue #7064, opened 2020-05-18
Implement AWX support for a data model and APIs to represent and manage Execution Environments


ryanpetrello:
This is a subtask of a larger enhancement/initiative surrounding execution environments: https://github.com/ansible/awx/issues/5157

Implement APIs for defining execution environment and their sources (registries).

- A model and API for defining, interacting with, and authenticating to one or more execution environments (e.g., name, URL, any necessary credential data).
  - A name
   - Credentials specific to registry authentication
   - A URL e.g., `quay.io/redhat/someimage:latest`
-  An interface for attaching execution environments to Organization, Inventories, and Job Templates.
-  RBAC for execution environments
- A documented process for registering new execution environments, and notifying AWX of their existance.

This work will be consumed by https://github.com/ansible/awx/issues/7060

ryanpetrello:
Additionally, we'll want to support this in our various SDK/CLI toolkit i.e., adding support creating and associating these in both the AWX collection and CLI.

jlmitch5:
UI work for execution environments:
- There will be a new credential type (container registry credential type).  Hopefully this will just work based on the credential type coming from the api once we build that (no UI config)
- New CRUD view for execution environments (i.e. images).  Detail view will have RBAC...would also need to add to the RBAC adder at the user level.  Fields are:  Fields are
  - name (text)
  - registry url (text)
  - credential selector (lookup, optional)
- Maybe: Execution environment detail view might have test button which shows success/failure notification when:
  - the image is accessible with the provided credentials
  - the test may also validate that containers built from the image provides access to runner and ansible
- We need fields to set execution environment where venv selectors are now (replaces the venv selectors)...in Job Template, Inventory sync, etc.

jbradberry:
Note that I've been convinced that having separate name and image url fields is redundant, so I've reduced it to a single field called `image`.

jbradberry:
Additional items and potential items that were left out of PR #7455 

- [x] Add a new `CredentialType`
- [ ] Add a configure-Tower-in-Tower setting for the global default EE
- [ ] Provide a way to install new bundled EEs
- [ ] Create a new Org-level EE admin RBAC role
  - [ ] Update the access class(es) to allow this to work
- [ ] Add a new `Project.default_environment` field, allowing a level of default EE between a full Org-level default and an explicit job template configured EE
  - [ ] Update the `Project` serializer to deal with this new field
- [ ] Add a new api sublist view for EE <- Unified Jobs
[...]

PR #7455, opened 2020-06-26
Add a new ExecutionEnvironment model

jbradberry:
##### SUMMARY
Modeling and additions to the API for the new execution environments.

- [x] a new `ExecutionEnvironment` model
- [ ] ~a new `CredentialType`~
- [x] a mixin to include in executable models to point to an `ExecutionEnvironment`
- [ ] ~a new configure-Tower-in-Tower setting to configure the global default EE~
- [x] update Organization to have an optional default EE
- [x] update the Org serializer to show and set the default EE
- [ ] ~update Project to have an optional default EE ????~
- [ ] ~update the Project serializer to show and set the default EE ????~
- [x] update the serializers for other objects to include the EE and `pull` fields
- [x] an RBAC access class
- [ ] ~add a new EE admin RBAC role ????~
- [x] an EE serializer
- [x] an API list view, and expose in the `/api/v2/` list
- [x] an API detail view
- [x] an API sublist view for Org <- EE relationships (_not_ the Org -> "Default EE" relationship)
- [x] an API sublist view for EE <- Unified Job Templates
- [ ] ~an API sublist view for EE <- Unified Jobs ????~
- [x] add a Swagger topic
- [x] make sure that ActivityStream captures changes and associations
- [x] an API sublist view for EE <- ActivityStreams
- [x] add new Page types to awxkit
- [x] make EEs exportable and hook up anything else necessary to allow the other CLI commands to work
- [ ] ~provide a way to create new EEs from scripts~

related #7064 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
awx: 13.0.0
```


awx/main/models/unified_jobs.py
```diff
@@ -59,7 +59,7 @@
 # NOTE: ACTIVE_STATES moved to constants because it is used by parent modules
 
 
-class UnifiedJobTemplate(PolymorphicModel, CommonModelNameNotUnique, NotificationFieldsModel):
+class UnifiedJobTemplate(PolymorphicModel, CommonModelNameNotUnique, ExecutionEnvironmentMixin, NotificationFieldsModel):
```
AlanCoding, line 62: > The execution environment used for this job

Not correct help_text for templates

AlanCoding:
Could we get some writeup in `docs/`?

This doesn't actually run jobs in the execution environment right now? Does that depend on https://github.com/ansible/ansible-runner/pull/470

I want to see what the contract is between the pieces of software. The best docs I've seen is this example:

https://github.com/jladdjr/ansible-runner/tree/04ea74015ece6bc08a6d734e9a4de6fe71c8eb06/test/integration/containerized/priv_data

```
---
idle_timeout: 60
job_timeout: 360
pexpect_timeout: 10
containerized: true
container_runtime: docker
```

Server code has a section for those settings in:

https://github.com/ansible/awx/blob/8ea31d8cdd2dff5ffa5aced9ce9d6bf4f3e944d5/awx/main/tasks.py#L1421-L1426

This remains very speculative, but I don't see why we can't add those settings there right now. We should also get the ability to run against ansible-runner branches. If we don't have that now, we should soon.

jbradberry:
@AlanCoding this is the very beginning of a draft.  It's a bit early for review.

awx/main/migrations/0118_v400_execution_environments.py
```diff
@@ -0,0 +1,68 @@
...
+        ),
+        migrations.CreateModel(
+            name='ExecutionEnvironment',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('created', models.DateTimeField(default=None, editable=False)),
+                ('modified', models.DateTimeField(default=None, editable=False)),
+                ('description', models.TextField(blank=True, default='')),
+                ('name', models.CharField(max_length=512)),
+                ('url', models.CharField(help_text='The registry location where the container is stored.', max_length=1024, verbose_name='URL')),
```
shanemcd: I'm wondering if it makes sense to have `name` and `url` be 2 separate things, or if there can just be one `name` field. What do you think?

jbradberry: You could be right, though I'm not sure about the 'name' field being the one that survives.  I'd think that if we kept only one that it should be the url field.

shanemcd: Mostly because they are technically referred to as "fully qualified image names"

jbradberry: hm, ok

awx/main/models/unified_jobs.py
jbradberry: Changed to: 'The default execution environment for jobs run by this organization.'

awx/main/models/execution_environments.py
```diff
@@ -0,0 +1,41 @@
...
+        null=True,
+        default=None,
+        blank=True,
+        on_delete=models.CASCADE,
+        related_name='%(class)ss',
+        help_text=_('The organization used to determine access to this execution environment.'),
+    )
+    image = models.CharField(
+        max_length=1024,
+        verbose_name=_('image location'),
```
kdelee, line 27: I would suggest, "Image name, which may include registry and tag"

awx/main/models/execution_environments.py
```diff
@@ -0,0 +1,41 @@
...
+    )
+    managed_by_tower = models.BooleanField(default=False, editable=False)
+    credential = models.ForeignKey(
+        'Credential',
+        related_name='%(class)ss',
+        blank=True,
+        null=True,
+        default=None,
+        on_delete=models.SET_NULL,
+    )
```
kdelee, line 38: would awx log in w/ the credential and will ansible-runner always just assume we are already authenticated w/ the registry?

What type of registry credentials are we going to support?

jbradberry: I don't know, I'm still pretty iffy on how this part will work.

awx/main/models/execution_environments.py
jbradberry: Generally the form field label gets set to this value, so do we really want to use something this verbose?

awx/main/models/execution_environments.py
AlanCoding: Do we have an idea of what _credential type_ this will need, and can we enforce that in the serializer or clean methods?

awx/main/models/execution_environments.py
AlanCoding: I would really like to see some content in the `docs/` folder with the exact syntax that is expected to work.

Take https://quay.io/repository/ansible-tower/ansible-runner as an example, we're sandboxing with this in other places too. Would there be no credential with image of `ansible-tower/ansible-runner`? Even if that's the case, it's not as obvious as it might seem, and it would be good to have an example.

shanemcd (approved):

jbradberry:
I'll capture the leftover bits (the ones we'll actually do, anyway) on another ticket.

Issue #7447, opened 2020-06-26
Can not Export inventory hosts


takezyou:
##### ISSUE TYPE
 - Bug or Fix

##### SUMMARY
Inventory does not include the host when doing an awx export

##### ENVIRONMENT
* AWX version: 13.0.0
* AWX install method: docker on linux
* Ansible version:  2.9.10
* Operating System: CentOS8
* AWX CLI version: 13.0.0

##### STEPS TO REPRODUCE
```
awx export --invetory 
```
- Inventory does not include the hosts

##### EXPECTED RESULTS
- Inventory contains the host

##### ACTUAL RESULTS
```
{
     "inventory": [
          {
               "name": "AnsibleServer",
               "description": "",
               "kind": "",
               "host_filter": null,
               "variables": "---",
               "insights_credential": null,
               "organization": {
                    "name": "Default",
                    "type": "organization"
               },
               "natural_key": {
                    "organization": {
                         "name": "Default",
                         "type": "organization"
                    },
                    "name": "AnsibleServer",
                    "type": "inventory"
               }
          },
          {
               "name": "Demo Inventory",
               "description": "",
               "kind": "",
               "host_filter": null,
               "variables": "",
               "insights_credential": null,
               "organization": {
                    "name": "Default",
                    "type": "organization"
               },
               "natural_key": {
                    "organization": {
                         "name": "Default",
                         "type": "organization"
                    },
                    "name": "Demo Inventory",
                    "type": "inventory"
               }
          },
```

takezyou:
@chrismeyersfsu 
I want to fix this myself.
Can I challenge you?

ryanpetrello:
Hey @takezyou,

We'd love a PR 😄 !

I'll leave @chrismeyersfsu' name on this since he's ultimately one of the people who would probably review such a PR.

takezyou:
Thank you!!😊
OK!! 
I'm creating a PR.
If I don't understand something, can I ask a question?

jbradberry:
@takezyou I'm also able to answer questions, as I did the original work on this feature.

takezyou:
@jbradberry Is there any way to disable page.PageCache()?

takezyou:
@jbradberry @chrismeyersfsu 
```
{
     "inventory": [
          {
               "name": "Demo Inventory",
               "description": "",
               "kind": "",
               "host_filter": null,
               "variables": "---",
               "insights_credential": null,
               "organization": {
                    "name": "Default",
[...]

PR #7460, opened 2020-06-28
[WIP]Fix cli export

takezyou:
##### SUMMARY

Bug fix export and import for AWX CLI.

related #7447 

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - CLI

##### AWX VERSION

```
13.1.0
```

### DIFF
--- a/awxkit/awxkit/api/pages/inventory.py
+++ b/awxkit/awxkit/api/pages/inventory.py
@@ -259,6 +259,7 @@ class Group(HasCreate, HasVariables, base.Base):
 
     dependencies = [Inventory]
     optional_dependencies = [Credential, InventoryScript]
+    NATURAL_KEY = ('name', 'inventory')
 
     @property
     def is_root_group(self):
@@ -367,6 +368,7 @@ page.register_page([resources.groups,
 class Host(HasCreate, HasVariables, base.Base):
 
     dependencies = [Inventory]
+    NATURAL_KEY = ('name', 'inventory')
 
     def payload(self, inventory, **kwargs):
         payload = PseudoNamespace(
--- a/awxkit/awxkit/api/pages/api.py
+++ b/awxkit/awxkit/api/pages/api.py
@@ -31,6 +31,8 @@ EXPORTABLE_RELATIONS = [
     'NotificationTemplates',
     'WorkflowJobTemplateNodes',
     'Credentials',
+    'Groups',
+    'Hosts'
 ]
 
 
@@ -38,6 +40,8 @@ EXPORTABLE_DEPENDENT_OBJECTS = [
     'Labels',
     'SurveySpec',
     'Schedules',
+    'Groups',
+    'Hosts',
     # WFJT Nodes are a special case, we want full data for the create
     # view and natural keys for the attach views.
     'WorkflowJobTemplateNodes',
@@ -89,6 +93,26 @@ class ApiV2(base.Base):
                 return None  # This foreign key has unresolvable dependencies
             fields[key] = rel_natural_key
 
+        natural_key = _page.get_natural_key(self._cache)
+
+        if natural_key['type'] == 'inventory':
+            related = self._export_inventories(_page, post_fields)
+            if related:
+                fields['related'] = related
+        else:
+            related = self._export_related(_page, post_fields)
+            if related:
+                fields['related'] = related
+
+        if natural_key is None:
+            log.error("Unable to construct a natural key for object %s.", _page.endpoint)
+            return None
+
+        fields['natural_key'] = natural_key
+
+        return utils.remove_encrypted(fields)
+
+    def _export_related(self, _page, post_fields):
         related = {}
         for key, rel_endpoint in _page.related.items():
             if key in post_fields or not rel_endpoint:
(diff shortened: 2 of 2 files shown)


awxkit/awxkit/api/pages/inventory.py
```diff
@@ -32,7 +32,7 @@
 class Inventory(HasCopy, HasCreate, HasInstanceGroups, HasVariables, base.Base):
 
     dependencies = [Organization]
-    NATURAL_KEY = ('organization', 'name')
+    NATURAL_KEY = ('organization', 'hosts', 'groups', 'name')
```
jbradberry: Hosts and groups should not be part of the natural key here.  The natural key is supposed to be a (mostly) fixed minimal subset of the information of the object needed to uniquely look it up.

awxkit/awxkit/api/pages/inventory.py
```diff
@@ -353,7 +354,7 @@ def remove_group(self, group):
 
 class Groups(page.PageList, Group):
 
-    pass
+    NATURAL_KEY = ('group',)
```
jbradberry: I'm pretty sure 'group' isn't a field or related on the Group model.  You should check the Django models under the awx/ subtree of this repo to get a sense of what fields are used for the uniqueness constraint for this model.

awxkit/awxkit/api/pages/inventory.py
```diff
@@ -434,7 +436,7 @@ def create(
 
 class Hosts(page.PageList, Host):
 
-    pass
+    NATURAL_KEY = ('host',)
```
jbradberry: As with Groups, I don't believe 'host' is a field on Host.

awxkit/awxkit/api/pages/inventory.py
jbradberry: Also, I just noticed that this is on Groups.  The plural page type shouldn't get the natural key, only the singular version.

jbradberry (changes requested):

jbradberry:
I'm afraid that this latest change has moved further away from where things probably should be.  Here's what I think:

- [ ] `Host` and `Group` should be in `EXPORTABLE_DEPENDENT_OBJECTS`
- [ ] they should _not_ be in `EXPORTABLE_RESOURCES`
- [ ] they should be in `EXPORTABLE_RELATIONS`
- [ ] looking at awx/main/models these are both `unique_together` on 'name' plus 'inventory', these should be the fields used for the `natural_key`

This would export the objects only as part of the related objects on exported inventories, and would create them in-place on that inventory during import, plus hooking up the relationships between them.

With all of that said, though, I'm not sure what would need to be done about inventory sources, inventory scripts, and special inventories.  Maybe @chrismeyersfsu can comment about this?

takezyou:
This ed6d68c commit can import but `maximum recursion depth exceeded while calling a Python object`  error.

ryanpetrello:
Hey @takezyou,

Thanks for taking a stab at this.  When you're done tinkering, let's please squash all these commits into a single one - it looks like there was a lot of activity here.

jladdjr:
hi @takezyou - I tried kicking the tires on these changes and found that while I can now see hosts in my export, the import isn't restoring the hosts. So, looks like we may be missing the second half.

For reference, here's the new export data that I'm seeing:
image


> With all of that said, though, I'm not sure what would need to be done about inventory sources, inventory scripts, and special inventories. 

^ as @jbradberry mentioned, inventories do touch on several other resources, though. Maybe, as a first step, we could have a PR that covers inventory basics (inventories, groups, hosts), and then grow that to include scripts, credentials, venvs, smart inventories, etc. There's a lot of ground to cover here.

jladdjr:
@takezyou - TLDR: looks like issue w/ exporting / importing hosts is tied to the underlying export code itself, and not the changes made in this PR.

---

@jbradberry and I took a look at this and found that the export logic doesn't seem to be handling the case of hosts correctly. Here, specifically: https://github.com/ansible/awx/blob/devel/awxkit/awxkit/api/pages/api.py#L109-L114

The issue is that exported hosts -- like this one:

image

.. are missing a `"natural_key"` section. That section is a cue to the export code that the JSON block represents _an object that should be created_, instead of just being a reference to the object.

I think @jbradberry can explain this much better than I can, but from what I gathered, things seem tricky because of variations on how endpoints work; some allow creating resources and also associating resources, some are restricted to just creating the resource. Apparently, the inventory / host case falls into one category, while WFJTs and WFJTNs fall into the other. Sounds like the code block I mentioned had one case in mind, but not the other.

takezyou:
@jladdjr Thank you comment!
I was able to notice the same problem when I was debugging it.

> I think @jbradberry can explain this much better than I can, but from what I gathered, things seem tricky because of variations on how endpoints work; some allow creating resources and also associating resources, some are restricted to just creating the resource. Apparently, the inventory / host case falls into one category, while WFJTs and WFJTNs fall into the other. Sounds like the code block I mentioned had one case in mind, but not the other.

@jbradberry What do you think about this?

jbradberry:
@takezyou I have nothing further to add than what @jladdjr said.

takezyou:
@jbradberry Please review.

jbradberry:
@takezyou I think that the underlying issue needs to be addressed, rather than bolting on support on a case-by-case basis.

jbradberry:
As noted, solving this is trickier than the approach outlined above.  Closing this in favor of #8368.

Issue #7454, opened 2020-06-26
Websockets aren't proxied correctly by ui development server (port 3001) causing them not to work in Chrome


jakemcdermott:
##### ISSUE TYPE
 - Bug Report

dsesami:
This was verified as part of the jobs list websockets verification, so it's in the clear.

PR #7469, opened 2020-06-29
Use a proxy config that works

### LINKED ISSUE #6520: Logout button on the API for ui-next does not work
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Click on the logout button on the API for ui-next makes a get request instead of a post request, due to which an error is thrown

##### STEPS TO REPRODUCE
1. https://localhost:3000/api
2. click on the logout button
Screen Shot 2020-04-01 at 10 28 17 AM

##### EXPECTED RESULTS
The logout should be successfull

##### ACTUAL RESULTS
An error is thrown:
Screen Shot 2020-04-01 at 10 28 34 AM

jakemcdermott:
#####  1. use a proxy config that works (addresses #7454)
The ui development server invoked by `react-scripts` is just the [webpack dev server](https://webpack.js.org/configuration/dev-server/) and the webpack dev server is just an [expressjs](https://expressjs.com/en/guide/using-middleware.html) server. When configuring proxy middleware for `react-scripts`, passing the path directly to the `use` method of the `expressjs` server, like [I did for our websockets proxy config](https://github.com/ansible/awx/blob/abcf8fea960b5c1c17d94f240a4f8aa546b0c81c/awx/ui_next/src/setupProxy.js#L32-L40) won't actually apply the proxy to that path. To correctly configure the proxy middleware, we must pass it the path directly:

```js
  // No!
  app.use(
    '/websocket',
    createProxyMiddleware({
      target: TARGET,
      secure: false,
      ws: true,
    })
  );

  // Yes!
  app.use(
    createProxyMiddleware('/websocket', {
      target: TARGET,
      secure: false,
      ws: true,
    })
  );
```

#####  2. Remove `proxy` field from package.json (resolves https://github.com/ansible/awx/issues/6520#issuecomment-641350592)
Workaround, no longer needed. 

#####  3. Replace dev server `TARGET_...` environment variables with a single environment variable.
Using separate variables for the host and port  is needlessly restrictive and can even cause problems with some deployments when trying to use the dev server. For example, to use the dev server wtih a default kubernetes install of awx you need to leave out the port and not use `https`, which can't be done with the current combination of vars. The simplest way to address this to just use one environment variable so you can pass in whatever you want.


awx/ui_next/src/setupProxy.js
```diff
@@ -1,35 +1,13 @@
...
-        Host: `localhost:${TARGET_PORT}`,
-        Origin: TARGET,
-        Referer: `${TARGET}/`,
-      },
-    })
-  );
-  app.use(
-    '/api',
-    createProxyMiddleware({
+    createProxyMiddleware(['/api', '/websocket'], {
```
AlexSCorey, line 7: Much cleaner, and it works!  Great job

AlexSCorey (approved):
Looks good to me.

PR #7485, opened 2020-06-30
replace certain terms with more inclusive language

ryanpetrello:
see: https://www.redhat.com/en/blog/making-open-source-more-inclusive-eradicating-problematic-language

### DIFF
--- a/awx/api/filters.py
+++ b/awx/api/filters.py
@@ -146,7 +146,7 @@ class FieldLookupBackend(BaseFilterBackend):
 
     # A list of fields that we know can be filtered on without the possiblity
     # of introducing duplicates
-    NO_DUPLICATES_WHITELIST = (CharField, IntegerField, BooleanField)
+    NO_DUPLICATES_ALLOW_LIST = (CharField, IntegerField, BooleanField)
 
     def get_fields_from_lookup(self, model, lookup):
 
@@ -205,7 +205,7 @@ class FieldLookupBackend(BaseFilterBackend):
         field_list, new_lookup = self.get_fields_from_lookup(model, lookup)
         field = field_list[-1]
 
-        needs_distinct = (not all(isinstance(f, self.NO_DUPLICATES_WHITELIST) for f in field_list))
+        needs_distinct = (not all(isinstance(f, self.NO_DUPLICATES_ALLOW_LIST) for f in field_list))
 
         # Type names are stored without underscores internally, but are presented and
         # and serialized over the API containing underscores so we remove `_`
--- a/awx/main/conf.py
+++ b/awx/main/conf.py
@@ -80,11 +80,11 @@ register(
 )
 
 register(
-    'PROXY_IP_WHITELIST',
+    'PROXY_IP_ALLOWED_LIST',
     field_class=fields.StringListField,
-    label=_('Proxy IP Whitelist'),
+    label=_('Proxy IP Allowed List'),
     help_text=_("If Tower is behind a reverse proxy/load balancer, use this setting "
-                "to whitelist the proxy IP addresses from which Tower should trust "
+                "to configure the proxy IP addresses from which Tower should trust "
                 "custom REMOTE_HOST_HEADERS header values. "
                 "If this setting is an empty list (the default), the headers specified by "
                 "REMOTE_HOST_HEADERS will be trusted unconditionally')"),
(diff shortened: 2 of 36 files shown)


matburt (approved):
👍

awx/settings/defaults.py
```diff
@@ -694,11 +694,6 @@ def IS_TESTING(argv=None):
     'cn-north-1': _('China (Beijing)'),
 }
 
-EC2_REGIONS_BLACKLIST = [
```
ryanpetrello, line 697: @AlanCoding I don't actually see where or how these settings are enforced?  Has this just been broken for a long time?

matburt: I thought these were converted to env variables and then passed to boto directly.

awx/settings/development.py
```diff
@@ -183,5 +183,4 @@
     except Exception:
         pass
 
-WEBSOCKET_ORIGIN_WHITELIST = ['https://localhost:8043', 'https://localhost:3000']
```
ryanpetrello, line 186: See: af84b25726326842d3e6a4b9f9de5e670945d070

awx/main/conf.py
```diff
@@ -80,11 +80,11 @@
...
 
 register(
-    'PROXY_IP_WHITELIST',
+    'PROXY_IP_ALLOWED_LIST',
     field_class=fields.StringListField,
-    label=_('Proxy IP Whitelist'),
+    label=_('Proxy IP Allowed List'),
     help_text=_("If Tower is behind a reverse proxy/load balancer, use this setting "
-                "to whitelist the proxy IP addresses from which Tower should trust "
+                "to allow the proxy IP addresses from which Tower should trust "
```
jbradberry: This construction, "to allow ... from which", seems particularly awkward to me.  I think maybe "to configure" works better here.

awx/main/models/notifications.py
```diff
@@ -262,25 +262,25 @@ class JobNotificationMixin(object):
...
-                                                               'total_inventory_sources', 'inventory_sources_with_failures',
-                                                               'organization_id', 'kind']},
-                                                {'project': ['id', 'name', 'description', 'status', 'scm_type']},
-                                                {'job_template': ['id', 'name', 'description']},
-                                                {'unified_job_template': ['id', 'name', 'description', 'unified_job_type']},
-                                                {'instance_group': ['name', 'id']},
-                                                {'created_by': ['id', 'username', 'first_name', 'last_name']},
-                                                {'labels': ['count', 'results']}]}]
+    JOB_FIELDS_ALLOWED_LIST = ['id', 'type', 'url', 'created', 'modified', 'name', 'description', 'job_type', 'playbook',
+                               'forks', 'limit', 'verbosity', 'job_tags', 'force_handlers', 'skip_tags', 'start_at_task',
```
rebeccahhh, line 266: what's up with these tabs?

rebeccahhh (approved):
:+1: I know we're going with "Allow" and "Block" but I do think "Block" and "Pass" would also be good, if we consider anything different at some point.

:rocket: it

awx/settings/defaults.py
AlanCoding: Yes, Matt is right, this might actually have been dropped on the floor in the transition to plugins

https://github.com/ansible/awx/blob/9.3.0/awx/main/models/inventory.py#L2095

We used this for the old setting `regions_exclude` for the script ini config.

The plugin has the option `regions`, but no include / exclude separate settings.

So you could check this setting in the modern loop for the plugin and exclude them... but we still use the keyword "all" if user doesn't provide regions. But if the user provided one of these regions, would we really want to remove them anyway? Outright removing this list is probably best.

awx/api/filters.py
```diff
@@ -146,7 +146,7 @@ class FieldLookupBackend(BaseFilterBackend):
 
     # A list of fields that we know can be filtered on without the possiblity
     # of introducing duplicates
-    NO_DUPLICATES_WHITELIST = (CharField, IntegerField, BooleanField)
+    NO_DUPLICATES_ALLOW_LIST = (CharField, IntegerField, BooleanField)
```
AlanCoding, line 149: The intent of the naming was "filtering these field types will not introduce duplicates"

you could also call it `DUPLICATE_SAFE` or `DISTINCT_NOT_NEEDED`

awx/main/models/notifications.py
AlanCoding: number of characters in variable changed. This is why many people dislike hanging indents.

AlanCoding (approved):

jladdjr (approved):

Issue #7326, opened 2020-06-11
Edit Custom Credential Type


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Edit Custom Credential Type

dsesami:
closing with the merge of #7489

unlikelyzero:
Reopening because no test coverage was referenced or added

PR #7489, opened 2020-06-30
Add edit credential types

nixocio:
Add feature to edit credential types.

See: https://github.com/ansible/awx/issues/7326

credential_type_edit

### DIFF
--- a/awx/ui_next/src/components/FormField/FormSubmitError.jsx
+++ b/awx/ui_next/src/components/FormField/FormSubmitError.jsx
@@ -22,6 +22,8 @@ function FormSubmitError({ error }) {
       Object.values(error.response.data).forEach(value => {
         if (Array.isArray(value)) {
           messages = messages.concat(value);
+        } else {
+          messages.push(value);
         }
       });
       setErrorMessage(messages.length > 0 ? messages : null);
--- a/awx/ui_next/src/screens/CredentialType/CredentialType.jsx
+++ b/awx/ui_next/src/screens/CredentialType/CredentialType.jsx
@@ -65,11 +65,6 @@ function CredentialType({ i18n, setBreadcrumb }) {
     },
   ];
 
-  let cardHeader = <RoutedTabs tabsArray={tabsArray} />;
-  if (pathname.endsWith('edit')) {
-    cardHeader = null;
-  }
-
   if (!isLoading && contentError) {
     return (
       <PageSection>
@@ -89,6 +84,11 @@ function CredentialType({ i18n, setBreadcrumb }) {
     );
   }
 
+  let cardHeader = <RoutedTabs tabsArray={tabsArray} />;
+  if (pathname.endsWith('edit')) {
+    cardHeader = null;
+  }
+
   return (
     <PageSection>
       <Card>
@@ -104,7 +104,7 @@ function CredentialType({ i18n, setBreadcrumb }) {
             {credentialType && (
               <>
                 <Route path="/credential_types/:id/edit">
-                  <CredentialTypeEdit />
+                  <CredentialTypeEdit credentialType={credentialType} />
                 </Route>
                 <Route path="/credential_types/:id/details">
                   <CredentialTypeDetails credentialType={credentialType} />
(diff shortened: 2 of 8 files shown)


nixocio:
~There is one extra case that I need to verify - changing this PR to WIP.~ Done. Ready to be reviewed.

awx/ui_next/src/screens/CredentialType/CredentialType.jsx
```diff
@@ -65,17 +65,12 @@ function CredentialType({ i18n, setBreadcrumb }) {
...
-    cardHeader = null;
-  }
-
   if (!isLoading && contentError) {
     return (
       <PageSection>
         <Card>
           <ContentError error={contentError}>
-            {contentError.response?.status === 404 && (
+            {contentError?.response?.status === 404 && (
```
AlexSCorey: I'm not sure that `contentError?.` is necessary.  It is checking if there is a `contentError`, which should be the case if it get past line 68

awx/ui_next/src/screens/CredentialType/CredentialTypeEdit/CredentialTypeEdit.jsx
```diff
@@ -1,10 +1,45 @@
...
+    } catch (error) {
+      setSubmitError(error);
+    }
+  };
+
+  const handleCancel = () => {
+    history.push(detailsUrl);
+  };
   return (
     <PageSection>
```
AlexSCorey: Screen Shot 2020-07-06 at 10 50 18 AM
You have an extra `<PageSection>`

awx/ui_next/src/components/FormField/FormSubmitError.jsx
```diff
@@ -22,6 +22,8 @@ function FormSubmitError({ error }) {
       Object.values(error.response.data).forEach(value => {
         if (Array.isArray(value)) {
           messages = messages.concat(value);
+        } else {
```
AlexSCorey, line 25: We have a util, which isn't in a util file, called getErrorMessage, which I think can handle all of these cases of error messages.  https://github.com/ansible/awx/blob/devel/awx/ui_next/src/components/ErrorDetail/getErrorMessage.js.  I thin we should, at some point, begin to migrate over to that file and let that handle our error messages.  If you want to investigate it, please feel free, but it isn't a blocker to merge this.

AlexSCorey (changes requested):
This looks really good so far.  Fix that 1 page section issue and if you want to touch base about the error handling ping me.

awx/ui_next/src/components/FormField/FormSubmitError.jsx
nixocio: Thanks for your suggestion. I looked in how we are handling this in other components. 

@keithjgrant filed https://github.com/ansible/awx/issues/7515, once this issue is sorted out, we will have a better pattern in how to deal those cases.

awx/ui_next/src/screens/CredentialType/CredentialTypeDetails/CredentialTypeDetails.jsx
```diff
@@ -70,7 +70,7 @@ function CredentialTypeDetails({ credentialType, i18n }) {
             <Button
               aria-label={i18n._(t`edit`)}
               component={Link}
-              to={`credential_types/${id}/edit`}
+              to={`/credential_types/${id}/edit`}
```
AlexSCorey, line 73: Good catch there

AlexSCorey (approved):
This looks good to me.  Well done!

dsesami:
Two notes:
1) this suffers from the same API char limit bug in https://github.com/ansible/awx/issues/7581. That does not block this PR, it is an API issue.
2) Can form validation on this block move the `!` bubble to the left side, like it is in most IDEs? it looks strange on the right side.
Screenshot from 2020-07-08 15-45-45

dsesami (approved):
approving as the above validation is related to the overall component, and not this PR.

PR #7491, opened 2020-07-01
Tower workflow schema

sean-m-sullivan:
##### SUMMARY
The schema option from the old tower workflow module is one that I know customers will still want. This takes the data model from awx export in development here: https://github.com/john-westcott-iv/awx/tree/import-export-collection and allows for that data model to be used as schema for a workflow. 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - Collection

##### AWX VERSION
```
awx: 11.2.0
```

##### ADDITIONAL INFORMATION
 - Based on the basics used in tower_workflow_job_template_node
 - Takes the schema from awx export  #7799
 - Does nodes and node associations in two steps, but one module.


awx_collection/plugins/modules/tower_workflow_job_template_schema.py
```diff
@@ -0,0 +1,378 @@
...
+__metaclass__ = type
+
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+---
+module: tower_workflow_job_template_node
```
AlanCoding: looks like this is copy+past from prior module

awx_collection/plugins/modules/tower_workflow_job_template_schema.py
```diff
@@ -0,0 +1,378 @@
...
+author: "John Westcott IV (@john-westcott-iv), Sean Sullivan (@Wilk42)"
+version_added: "2.9"
+short_description: create, update, or destroy Ansible Tower workflow job template graphs.
+description:
+    - Create, update, or destroy Ansible Tower workflow job template graphs.
+    - Use this to build a graph for a workflow, which dictates what the workflow runs.
+    - Replaces the deprecated tower_workflow_template module schema command.
+    - This module takes a josn list of inputs to first create nodes, then link them, and does not worry about ordering.
```
AlanCoding: josn -> json

awx_collection/plugins/modules/tower_workflow_job_template_schema.py
```diff
@@ -0,0 +1,378 @@
...
+        - workflow
+    organization:
+      description:
+        - The organization of the workflow job template the node exists in.
+        - Used for looking up the workflow, not a direct model field.
+      type: str
+    schema:
+      description:
+        - A json list of nodes and their coresponding options.
+      type: str
```
AlanCoding: Why would you take this as a JSON str? If it's ultimately a list, then use the "list" type, and since it's a list of nodes, I'm guessing that the elements (specified by the `elements:` key here) are "dict".

awx_collection/plugins/modules/tower_workflow_job_template_schema.py
```diff
@@ -0,0 +1,378 @@
...
+        - Job tags applied as a prompt, if job template prompts for job tags
+      type: str
+    skip_tags:
+      description:
+        - Tags to skip, applied as a prompt, if job tempalte prompts for job tags
+      type: str
+    limit:
+      description:
+        - Limit to act on, applied as a prompt, if job template prompts for limit
+      type: str
```
AlanCoding: I'm confused about these options, you said the `schema` is the important input and takes a list of nodes. These are node properties, so they shouldn't be inputs to the module, they should be nested properties in dicts inside the schema option, if I read it correctly.

awx_collection/plugins/modules/tower_workflow_job_template_schema.py
```diff
@@ -0,0 +1,378 @@
...
+      foo_key: bar_value
+
+- name: Create parent node for prior node
+  tower_workflow_job_template_node:
+    identifier: my-root-node
+    workflow: example-workflow
+    unified_job_template: jt-for-node-use
+    organization: Default
+    success_nodes:
+      - my-first-node
```
AlanCoding: It would be good to have examples with schema to start setting basic expectations

awx_collection/plugins/modules/tower_workflow_job_template_schema.py
sean-m-sullivan: Resolving and updating documentation

awx_collection/plugins/modules/tower_workflow_job_template_schema.py
sean-m-sullivan: Updated documentation for review. Matches output from Tower API, and the tower_export module.

awx_collection/plugins/modules/tower_workflow_job_template_schema.py
sean-m-sullivan: I've updated an example, and believe it should be a list item.

awx_collection/plugins/modules/tower_workflow_job_template_schema.py
sean-m-sullivan: @AlanCoding I'm planning in the future to do a simplified version along with the tower_export schema model, but it should be updated now.

wenottingham:
> Takes the schema from the upcoming awx export

Just to be clear - this format already exists - is this using the current awx cli export, or something else?

sean-m-sullivan:
> > Takes the schema from the upcoming awx export
> 
> Just to be clear - this format already exists - is this using the current awx cli export, or something else?


@wenottingham 
At the time I wrote that, It was not merged in yet. This is the output from the awxkit export.

Issue #7445, opened 2020-06-25
 Application Tokens


AlexSCorey:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Add the Tokens tab /list for Application.  Should hit api/v2/applications/1/tokens

PR #7497, opened 2020-07-01
Adds Application Token List with delete functionality

AlexSCorey:
##### SUMMARY
This address #7445.  It basically has the same functionality as the old UI, with 1 addition.  Now the user can delete Tokens from this view.  Only users that have admin or org admin privileges will be able to see this list, thus they also have delete privileges.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION
Screen Shot 2020-07-01 at 9 18 58 AM

### DIFF
--- a/awx/ui_next/src/api/index.js
+++ b/awx/ui_next/src/api/index.js
@@ -24,6 +24,7 @@ import Roles from './models/Roles';
 import Schedules from './models/Schedules';
 import SystemJobs from './models/SystemJobs';
 import Teams from './models/Teams';
+import Tokens from './models/Tokens';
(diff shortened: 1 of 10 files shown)


jlmitch5 (approved):
looks great!

dsesami (changes requested):
There isn't a way to differentiate between two items with Read/Write scope. We should add the scope to the list view so that the user can tell which token is which.

As you can see here, they look exactly the same, but one is read and one is write.

Screenshot from 2020-07-08 13-44-04

dsesami (approved):

Issue #7475, opened 2020-06-29
[awx_collection] Manage users in teams; manage teams/users in templates


iridian-ks:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Right now, I'm generating maybe ~100 job templatres with Ansible. Some of the jobs are duplicates and simply prefill different variables (environments, etc...). Different users have access to different environments. We use the automatic LDAP mapping to get users into Organizations, but it looks like a job/workflow template can only be a part of a single organization so creating these jobs and handing them off to different teams by leveraging the awx_collection seems impossible (at the moment) (I could also just be missing something).

My initial thought is to create Teams under my Organization and put users in those teams. Basically, a team per environment. Then map the job templates to those teams. When those users login they then see only the jobs they have access to. For example, devs can access dev environments and wouldn't be able to accidentally modify variables to target production, etc.

Either under `tower_team` or `tower_user` I feel like we should be able to add users to a team. It looks like that's not possible right now?

Also, under `tower_job_template` we should be able to manage the users/teams that have access to them. That also doesn't seem to be possible, right now?

I was very surprised to see the lack of functionality and my first thought is that I'm simply missing something, which might be the case and I'm just not smart enough to find the right module to manage the mappings.

wenottingham:
This is what `tower_role` is for.

iridian-ks:
I knew I was missing something... sometimes in Ansible module documentation you can find related modules. I'm wondering if I could take a stab at providing links in the docs so that `tower_team` and `tower_job_template` can link out to `tower_role`. Maybe that could help others?

wenottingham:
Closing this specific issue, as the functionality is there.

PR #7503, opened 2020-07-01
Added documentation notes linking to tower_role for tower_team and to…

iridian-ks:
…wer_job_template

##### SUMMARY

Related to issue #7475 

Just adding some links to the docs.

##### ISSUE TYPE

 - Docs Pull Request

##### COMPONENT NAME

 - awx_collection

##### AWX VERSION

N/A

### DIFF
--- a/awx_collection/plugins/modules/tower_job_template.py
+++ b/awx_collection/plugins/modules/tower_job_template.py
@@ -282,6 +282,7 @@ notes:
   - JSON for survey_spec can be found in Tower API Documentation. See
     U(https://docs.ansible.com/ansible-tower/latest/html/towerapi/api_ref.html#/Job_Templates/Job_Templates_job_templates_survey_spec_create)
     for POST operation payload example.
+  - To manage permissions for job templates. See U(https://docs.ansible.com/ansible/latest/modules/tower_role_module.html)
 '''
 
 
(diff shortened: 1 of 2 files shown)


awx_collection/plugins/modules/tower_job_template.py
```diff
@@ -282,6 +282,7 @@
   - JSON for survey_spec can be found in Tower API Documentation. See
     U(https://docs.ansible.com/ansible-tower/latest/html/towerapi/api_ref.html#/Job_Templates/Job_Templates_job_templates_survey_spec_create)
     for POST operation payload example.
+  - To manage permissions for job templates. See U(https://docs.ansible.com/ansible/latest/modules/tower_role_module.html)
```
wenottingham, line 285: This (and the other change) should either link to the collection docs on galaxy/Automation Hub (which makes it mor complicated), or just not have a link and just refer to `tower_role` by name.

AlanCoding:
I don't like the idea of linking these here, but that's not a strong opposition.

The main problem here is that it links to the wrong thing, very recently, the Ansible docs team developed a solution for publishing collection docs. This has been done for the `awx.awx` collection, see:

https://docs.ansible.com/ansible/2.10/collections/awx/awx/

Although it would be coherent to link here, it would result in that page linking to itself... and, we don't have a way to reference the latest, as opposed to 2.10, and the organization structure for this could still change with later Ansible versions. I just want to document that as I close this, @ryanpetrello has been fielding some similar questions on this, and kudos to @acozine for getting the docs pages together

PR #7525, opened 2020-07-01
don't allow users to specify $encrypted$ for encrypted credential fields

ryanpetrello:
this keyword only has value when you _update_ an existing credential

### DIFF
--- a/awx/main/fields.py
+++ b/awx/main/fields.py
@@ -637,6 +637,14 @@ class CredentialInputField(JSONSchemaField):
             else:
                 decrypted_values[k] = v
 
+        # don't allow secrets with $encrypted$ on new object creation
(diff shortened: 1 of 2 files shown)


ryanpetrello:
cc @john-westcott-iv

ryanpetrello:
image

AlanCoding (approved):

jakemcdermott:
Creating a new credential with input sources via the UI might depend on this behavior.

I tried to test this but ran into: https://github.com/ansible/awx/issues/7530

elyezer (approved):
LGTM considering API side. I also ran into https://github.com/ansible/awx/issues/7530 when trying thru the UI.

Issue #7483, opened 2020-06-30
Custom credential types with boolean fields cannot be created or edited using the AWX console


alanbchristie:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
I am creating custom credential types (and custom credentials) using the Ansible awx.awx collection's `awx.awx.tower_credential_type` and `awx.awx.tower_credential` modules. I can create credentials using the `tower-cli` but not using the AWX console.

You cannot edit existing or create new custom credentials in the console because the **SAVE** button is never enabled. When I click the checkbox for the boolean field that's presented for a boolean property the text **Please enter a value** is printed.

Custom credential types that do not contain booleans can be edited and created.

##### ENVIRONMENT
* AWX version: 11.2.0
* AWX install method: kubernetes
* Ansible version:  2.9.7
* Operating System: kubernetes
* Web Browser: Safari or Chrome

##### STEPS TO REPRODUCE

Create a custom credential type (using the Ansible `awx.awx.tower_credential` module). Here's a snippet from my playbook that creates my own `kubernetes` type. The field that causes the problem is `verify_ssl`: -

```
- name: Add kubernetes credential type
  awx.awx.tower_credential_type:
    name: kubernetes
    description: Credentials for a Kubernetes/OpenShift cluster
    kind: cloud
    inputs:
      fields:
      - id: name
        type: string
        label: A symbolic name for the cluster
      - id: host
        type: string
        label: API host
      - id: api_key
        type: string
        label: API Key (Token)
        secret: true
      - id: verify_ssl
        type: boolean
        label: Verify SSL
      required:
      - name
      - host
      - api_key
      - verify_ssl
    injectors:
      env:
        K8S_AUTH_NAME: "{{ '{{' }} name {{ '}}' }}"
        K8S_AUTH_HOST: "{{ '{{' }} host {{ '}}' }}"
        K8S_AUTH_API_KEY: "{{ '{{' }} api_key {{ '}}' }}"
        K8S_AUTH_VERIFY_SSL: "{{ '{{' }} verify_ssl {{ '}}' }}"
```

You can create credentials of this type using the `awx.awx.tower_credential` module. Here's a snippet that successfully creates instances of the credential that can be used in templates: -

```
- name: Add kubernetes credential
  awx.awx.tower_credential:
    name: "{{ item.name }}"
    credential_type: kubernetes
    description: "{{ item.description|default('', true) }}"
    organization: "{{ tower.organisation }}"
    inputs:
[...]

PR #7537, opened 2020-07-02
Fix custom credential field validation

jakemcdermott:
##### SUMMARY
for #7483 

User has a custom credential with `required` booleans and at least one custom field id that is identical to the name of a field on the actual credential model (`name`, `description`, etc).

- Ignore required field validations for booleans. You can't _not_ provide a value for these. Not checking the box is implicitly providing a value of `false`. 
- Underscore the property names of the common credential model fields on the ui state to avoid collisions w/ custom cred field ids.

### DIFF
--- a/awx/ui/client/features/credentials/add-edit-credentials.controller.js
+++ b/awx/ui/client/features/credentials/add-edit-credentials.controller.js
@@ -53,16 +53,19 @@ function AddEditCredentialsController (
             vm.form.disabled = !isEditable;
         }
 
-        vm.form.organization._disabled = !isOrgEditableByUser;
+        vm.form._organization._disabled = !isOrgEditableByUser;
         // Only exists for permissions compatibility
         $scope.credential_obj = credential.get();
 
-        vm.form.organization._resource = 'organization';
-        vm.form.organization._model = organization;
-        vm.form.organization._route = 'credentials.edit.organization';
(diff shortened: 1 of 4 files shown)


awx/ui/client/features/credentials/add-edit-credentials.controller.js
```diff
@@ -53,16 +53,19 @@ function AddEditCredentialsController (
...
-        vm.form.organization._resource = 'organization';
-        vm.form.organization._model = organization;
-        vm.form.organization._route = 'credentials.edit.organization';
-        vm.form.organization._value = credential.get('summary_fields.organization.id');
-        vm.form.organization._displayValue = credential.get('summary_fields.organization.name');
-        vm.form.organization._placeholder = strings.get('inputs.ORGANIZATION_PLACEHOLDER');
+        // Custom credentials can have input fields named 'name', 'organization',
+        // 'description', etc. Underscore these variables to make collisions
+        // less likely to occur.
+        vm.form._organization._resource = 'organization';
```
marshmalien, line 63: 👍 Nice

marshmalien (approved):
Tested this locally. Looks great!

AlexSCorey (approved):
No concerns on my end. Looks good to me.

dsesami:
Booleans in custom credential fields appear to work normally on both save and edit. :+1:

dsesami (approved):

PR #7540, opened 2020-07-02
Add Websocket support to Projects List

keithjgrant:
##### SUMMARY
Adds websocket support to project list, showing live data for most recent sync status.

The general approach is modeled after websockets on the JobsList: list of loaded projects is passed into a hook (`useWsProjects`), which then returns a copy of the list. The hook subscribes to websocket events and updates the returned list with live sync data. If navigation happens in the main component (ProjectList), such as pagination/sorting/searching, new data is fetch via the API as normal and the hook begins working with this new list.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI


awx/ui_next/src/screens/Project/ProjectList/useWsProjects.js
```diff
@@ -0,0 +1,85 @@
...
+    };
+    setProjects([
+      ...projects.slice(0, index),
+      updatedProject,
+      ...projects.slice(index + 1),
+    ]);
+  }, [lastMessage]); // eslint-disable-line react-hooks/exhaustive-deps
+
+  useEffect(() => {
+    ws.current = new WebSocket(`wss://${window.location.host}/websocket/`);
```
jakemcdermott, line 41: It is starting to look like we should move the ws creation boilerplate to a utility function of some sort. We're doing this in a few places now and, as far as I can tell, the only difference between the different invocations is the subscribed `groups` and the action to perform `onmessage`:

Something like:

```jsx
import createWebsocket from 'utils/createWebsocket';

...

export default function useWsProjects(initialProjects) {
  ...
 
  useEffect(() => {
    ws.current = createWebsocket({
      groups: {
        jobs: ['status_changed'],
        control: ['limit_reached_1'],
      },
      onMessage: setLastMessage,
    });

    return () => {
      ws.current.close();
    };
  }, []);

  ...
```

Thoughts? I could also see it being too early to tell at this point. Up to you.

awx/ui_next/src/screens/Project/ProjectList/useWsProjects.js
```diff
@@ -0,0 +1,85 @@
...
+          finished: lastMessage.finished,
+        },
+      },
+    };
+    setProjects([
+      ...projects.slice(0, index),
+      updatedProject,
+      ...projects.slice(index + 1),
+    ]);
+  }, [lastMessage]); // eslint-disable-line react-hooks/exhaustive-deps
```
jakemcdermott, line 38: Do we need to monitor a `lastMessage` state variable with `useEffect` ? Adding `useEffect` _seems_ like it might be adding an unnecessary second step to event processing but it may just be that I'm missing something about a problem it solves.

What about putting the project updating logic in a function and triggering it directly from the events emitted by the websocket? `setState` would potentially get called for each ws event but we could always debounce or throttle the call with a utility.

jakemcdermott (commented):
@keithjgrant I tested this and it seems to work well :+1: 

wsprojects

I have a few questions regarding the hook implementation (see review comments). Please let me know what you think.

awx/ui_next/src/screens/Project/ProjectList/useWsProjects.js
keithjgrant: Yeah I agree. I was definitely starting to think about this. I'm starting on Inventory lists for my next PR, so I think I'll tackle it along with that change to help make sure it covers a broad range of uses.

awx/ui_next/src/screens/Project/ProjectList/useWsProjects.js
keithjgrant: This approach is based on one I'd seen in some examples online with WS and React hooks. I want to say there was a reason it didn't work the other way, but I'll verify that...

awx/ui_next/src/screens/Project/ProjectList/useWsProjects.js
keithjgrant: actually, your suggestion is a little different than my initial thought. I had considered making a `useWebsocket` hook that handled this as well as onMessage handling, then returning `lastMessage` for the component to consume

jakemcdermott (approved):

awx/ui_next/src/screens/Project/ProjectList/useWsProjects.js
jakemcdermott: sounds good

keithjgrant:
Cypress tests pass.

Spoke with @dsesami — since my next PR will involve some re-factoring of this, QE will take a closer look at this after those changes are in.

PR #7552, opened 2020-07-06
Allow anyone to start awx and awx_task [#7545]

### LINKED ISSUE #7545: Permission denied for launch_awx.sh and launch_awx_task.sh
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

I am currently working on updating @ikke-t [awx_pod playbook](https://github.com/ikke-t/awx_pod) for using awx 13.

My deployment is a bit different from the installation instructions here, but I would expect that the error below appears when using e.g. docker-compose as well.

I am using the following image: ``docker.io/ansible/awx:13.0.0``

The following files have permissions 0700:

* /usr/bin/launch_awx.sh
* /usr/bin/launch_awx_task.sh

which leads to issues when running ``docker.io/ansible/awx:13.0.0`` with non-root user ``1000``

imho this files should have 0755.

##### ENVIRONMENT
* AWX version: 13.0.0
* AWX install method: podman / awx_pod playbook
* Ansible version:  2.9.10
* Operating System: EL8 (CentOS 8)

##### STEPS TO REPRODUCE

```
$ ansible-playbook -K -i centos-8, awx-install.yml
$ podman logs -f awx_awxweb
$ podman logs -f awx_awxtask
```

##### EXPECTED RESULTS

```
RESULT 2
OK
```

##### ACTUAL RESULTS

Permission denied for launch_awx.sh and launch_awx_task.sh

```
$ podman logs awx_awxweb
/bin/sh: /usr/bin/launch_awx.sh: Permission denied

$ podman logs awx_awxtask
/bin/sh: /usr/bin/launch_awx_task.sh: Permission denied
```

##### ADDITIONAL INFORMATION

Creating a derived container using

``
RUN chmod 755 /usr/bin/launch_awx*.sh
``

resolves this issue.

ikke-t:
##### SUMMARY
Fixes #7545 by adding exec bits to start scripts for awx and awx task processes.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - Installer

##### AWX VERSION
13, git devel head.

##### ADDITIONAL INFORMATION
This change is needed, as at current AWX start fails if run with podman, as non root. Podman uses non root user, so I add here read and execution bit to group and others as well, so the processes get started. Without this, start fails with "no permissions".

### DIFF
--- a/installer/roles/image_build/tasks/main.yml
+++ b/installer/roles/image_build/tasks/main.yml
@@ -110,14 +110,14 @@
   copy:
     src: launch_awx.sh
     dest: "{{ docker_base_path }}/launch_awx.sh"
-    mode: '0700'
+    mode: '0755'
   delegate_to: localhost
 
 - name: Stage launch_awx_task
   template:
     src: launch_awx_task.sh.j2
     dest: "{{ docker_base_path }}/launch_awx_task.sh"
-    mode: '0700'
+    mode: '0755'
   delegate_to: localhost
 
 - name: Stage rsyslog.conf


installer/roles/image_build/templates/Dockerfile.j2
```diff
@@ -209,7 +209,8 @@ RUN for dir in \
 # Adjust any remaining permissions
 RUN chmod u+s /usr/bin/bwrap ; \
     chgrp -R root ${COLLECTION_BASE} ; \
-    chmod -R g+rw ${COLLECTION_BASE}
+    chmod -R g+rw ${COLLECTION_BASE} ; \
+    chmod ugo+rx /usr/bin/launch_awx.sh /usr/bin/launch_awx_task.sh
```
wenottingham: Why do you need both? The permissions should be respected if the files are already that permission.

ikke-t: OK, if so, I'll force push dropping that part. I was thinking about the same, but as I was unsure about the build process, I put it to both. Thanks for simplifying this!

ikke-t: I reverted this dockerfile change now.

ryanpetrello (approved):

Issue #7546, opened 2020-07-03
HashiCorp Vault Secret Lookup: Test button don't work


stasonspb:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Can't get test form for Lookup type credential

##### ENVIRONMENT
* AWX version: 13.0.0
* AWX install method:  docker on linux
* Ansible version:  2.9.10
* Operating System: linux
* Web Browser: Firefox, Chrome

##### STEPS TO REPRODUCE

Add a new credential
type: “HashiCorp Vault Secret Lookup”
set: Name, Server URL, Token
Click the "test" button

##### EXPECTED RESULTS

get the form "Test External Credential"
![image](https://user-images.githubusercontent.com/16939849/86481748-45dbe780-bd59-11ea-9d39-bba164335a21.png)

##### ACTUAL RESULTS

get nothing

##### ADDITIONAL INFORMATION

Same problem for another Lookup credential type(CyberArk Conjur Secret Lookup and CyberArk AIM Central Credential Provider Lookup)
Button works on version 12.0.0

ryanpetrello:
Hey @stasonspb,

Thanks for reporting this - this looks related to a recent dependency upgrade we made (to jQuery cc @jakemcdermott).

This should be resolved in the next release of AWX: https://github.com/ansible/awx/pull/7553

venu-devannagari:
Hi did it resolved..

stasonspb:
For me issue is resolved in the latest version(14.0.0)

wenottingham:
Closing as fixed.

Issue #7534, opened 2020-07-02
Edit Schedule view is broken in AWX 13.0.0


kenchrcum:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Schedules cannot be edited in AWX 13.0.0. It seems to be some kind of templating error.

##### ENVIRONMENT
* AWX version: 13.0.0
* AWX install method: docker
* Ansible version:  2.9.10
* Operating System: centos
* Web Browser: chrome

##### STEPS TO REPRODUCE

Install fresh AWX deployment and click on Schedules -> Cleanup Activity Schedule

##### EXPECTED RESULTS

Possibility to edit Scheduled Job

##### ACTUAL RESULTS

![awx-schedule-bug](https://user-images.githubusercontent.com/21356628/86352136-9d992680-bc65-11ea-8795-071af4b314ff.png)

Luke239847:
Got the same errors.
Seems to occur only when the _Repeat Frequency_ is set to "Week"

jakemcdermott:
Hey @kenchrcum  - we've just merged a patch for this. It'll be fixed in the next release of awx.

Aech1977:
is there a workaround until this pull request is approved and merged (or did it happen already and I'm just dense)?

ryanpetrello:
@Aech1977 the fix has merge into AWX, but we haven't made an official release of AWX since then (we're likely going to release a new version in the coming weeks).

Aech1977:
I'll sit tight then.  Thanks.  Another dumb question I know I came across it before but where can I see release versions?
Thanks for the speedy response.

ryanpetrello:
Hey @Aech1977,

New releases show up here: https://github.com/ansible/awx/releases
We publish official images to Dockerhub: https://hub.docker.com/r/ansible/awx

Additionally, we have an official changelog at: https://github.com/ansible/awx/blob/devel/CHANGELOG.md

...and we announce releases on our mailing list: https://groups.google.com/forum/#!forum/awx-project

Aech1977:
thanks

kenchrcum:
Fix in version 14.0.0 verified. Thank you everybody :)

PR #7553, opened 2020-07-06
Fix self-closing tags for jquery 3.5 compat

### LINKED ISSUE #7530: 'Next' button is missing from credential plugin prompt
##### ISSUE TYPE
Seems to be devel only. Did not reproduce on recent release. I do not know the cause nor am I closer to it after an hour digging into this. I think I may need to bisect the devel commits to find it.

Screenshot from 2020-07-01 19-30-41

jakemcdermott:
Fixes https://github.com/ansible/awx/issues/7530, https://github.com/ansible/awx/issues/7546 and part of https://github.com/ansible/awx/issues/7534


jlmitch5 (approved):
looks good!

jakemcdermott:
also addresses https://github.com/ansible/awx/issues/7606

Issue #7433, opened 2020-06-24
User Tokens Add


AlexSCorey:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Add functionality to add a token in Users https://tower-mockups.testing.ansible.com/patternfly/users/users-tokens-add/

PR #7555, opened 2020-07-06
Adds User TokenAdd Functionality

AlexSCorey:
##### SUMMARY
This addresses #7433 and adds support to add a token.  It also adds an Application Look up.  

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
Screen Shot 2020-07-06 at 12 07 27 PM

### DIFF
--- a/awx/ui_next/src/screens/User/shared/UserTokenForm.jsx
+++ b/awx/ui_next/src/screens/User/shared/UserTokenForm.jsx
@@ -1,0 +1,127 @@
+import React from 'react';
+import { withI18n } from '@lingui/react';
+import { t } from '@lingui/macro';
+import { Formik, useField } from 'formik';
+import { Form, FormGroup } from '@patternfly/react-core';
+import AnsibleSelect from '../../../components/AnsibleSelect';
+import FormActionGroup from '../../../components/FormActionGroup/FormActionGroup';
+import FormField, {
+  FormSubmitError,
+  FieldTooltip,
+} from '../../../components/FormField';
+import ApplicationLookup from '../../../components/Lookup/ApplicationLookup';
+import { required } from '../../../util/validators';
+
+import { FormColumnLayout } from '../../../components/FormLayout';
+
+function UserTokenFormFields({ i18n }) {
+  const [applicationField, applicationMeta, applicationHelpers] = useField(
+    'application'
+  );
+
+  const [scopeField, scopeMeta, scopeHelpers] = useField({
+    name: 'scope',
+    validate: required(i18n._(t`Please enter a value.`), i18n),
+  });
+
+  return (
+    <>
+      <FormGroup
+        fieldId="application-lookup"
+        name="application"
+        validated={
+          !applicationMeta.touched || !applicationMeta.error
+            ? 'default'
+            : 'error'
+        }
+        helperTextInvalid={applicationMeta.error}
+      >
+        <ApplicationLookup
+          value={applicationField.value}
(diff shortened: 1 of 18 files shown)


awx/ui_next/src/screens/User/shared/UserTokenForm.jsx
```diff
@@ -0,0 +1,131 @@
...
+            {i18n._(t`Scope`)}
+            <span className="pf-c-form__label-required" aria-hidden="true">
+              *
+            </span>
+            <FieldTooltip
+              content={i18n._(t`Specify a scope for the token's access`)}
+            />
+          </span>
+        }
+      >
```
keithjgrant: FormGroup has an `isRequired` prop you can set to add the red asterisk, so you don't have to do it by hand here

awx/ui_next/src/screens/User/shared/UserTokenForm.test.jsx
```diff
@@ -0,0 +1,143 @@
...
+      wrapper = mountWithContexts(
+        <UserTokenForm handleSubmit={jest.fn()} handleCancel={handleCancel} />
+      );
+    });
+    await waitForElement(wrapper, 'ContentLoading', el => el.length === 0);
+    expect(handleCancel).not.toHaveBeenCalled();
+    wrapper.find('button[aria-label="Cancel"]').invoke('onClick')();
+    expect(handleCancel).toBeCalled();
+  });
+  // test('should throw error on submit without scope value', async() => {
```
keithjgrant: Is this commented test needed?

awx/ui_next/src/screens/User/shared/UserTokenForm.jsx
AlexSCorey: @keithjgrant I know about the `isRequired` prop. However, since the PF4 upgrade the alignment of the labels has been off.  See https://github.com/ansible/awx/issues/7556.  This is an attempt to solve that problem, but I've found a better solution.

awx/ui_next/src/screens/User/shared/UserTokenForm.test.jsx
AlexSCorey: I meant to update that test.  I'll do it.

keithjgrant (approved):
👍🏻

dsesami (changes requested):
There's a bug on this one and I don't know if it's in the other form, so I'd take a peek.

1. Add a token.
2. Save.
3. Try to click the add button again.

It starts to append to the URL like ```/add/add/add/add/add``` and doesn't work.

dsesami (approved):
This PR brought up https://github.com/ansible/awx/issues/7581. Once that's fixed, we'll need to make sure the UI bubbles up validation.

What's left: this brought up the question of whether or not the description should be added to the main list. since that's technically outside of the scope of this PR, i'm approving it for now, but if you'd like one more pass while we're in this block of code, I'm not against it.

AlexSCorey:
> This PR brought up #7581. Once that's fixed, we'll need to make sure the UI bubbles up validation.
> 
> What's left: this brought up the question of whether or not the description should be added to the main list. since that's technically outside of the scope of this PR, i'm approving it for now, but if you'd like one more pass while we're in this block of code, I'm not against it.

after convo with @trahman73 we are not going to add the description to this list item.  We will, of course, add it to the details page.

PR #7583, opened 2020-07-08
added OIDC tab to auth settings

emargolin:
##### SUMMARY
Added an OpenID Connect button in the Settings > Authorization section.

Screen Shot 2020-07-08 at 12 45 00 PM

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - UI

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/ui/client/src/configuration/forms/auth-form/configuration-auth.controller.js
+++ b/awx/ui/client/src/configuration/forms/auth-form/configuration-auth.controller.js
@@ -45,7 +45,7 @@ export default [
             if(!_.get($scope.$parent, [formTracker.currentFormName(), '$dirty'])) {
                 authVm.activeTab = tab;
                 authVm.activeAuthForm = getActiveAuthForm(tab);
-                formTracker.setCurrentAuth(authVm.activeAuthForm);
+                formTracker.setCurrentAuth(authVm.activeAuthForm); 
                 startCodeMirrors();
             } else {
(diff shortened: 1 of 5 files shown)


ryanpetrello:
Hey @emargolin,

Did you mean to open this?

It looks like it has some private hostnames/usernames in it.

emargolin:
Hi Ryan,

Yes! I meant to make a pull request with some changes onto my branch, but
didn't mean to keep the private hostnames. Should I make any changes?

Thanks,
Elena

emargolin (commented):
Took out private hostnames.

wenottingham:
Hi -

Outside of any accidental changes to the inventory file, this isn't something we'd ever accept like this. We're not going to add UI code for features that aren't actually in AWX; this would need to be pared with a backend implementation.

ryanpetrello:
Hey @emargolin,

What @wenottingham said.  This JavaScript code looks like a step in the right direction _if there were_ an official backend generic OIDC implementation to match - but that's the part that's missing for us to consider merging something like this.  @wenottingham, @matburt and I spoke with Suresh Kumar earlier today - before anything like this could land in AWX in an _official_ supported capacity, it would require support in the underlying Python libraries (namely `python-social-auth`) that AWX relies on.

This pull request doesn't look wrong as a necessary step in the direction of official support, but it's also probably premature.

PR #7584, opened 2020-07-08
use jinja2.sandbox for credential type injectors

ryanpetrello:


ghjm (approved):
LGTM.  (Did not actually test it though.)

jladdjr (approved):

PR #7585, opened 2020-07-08
Safely enable HTML in Custom Login Info

philipsd6:
###### Rationale:
We are required to include a link to our privacy policy on the login page. In plain text this looks like:

    Please review our GDPR privacy statement before proceeding:
    https://global.intranet.company.com/homesites/GDPRHOME/GDPR%20documents/Company%20Internal%20Technologies%20Privacy%20Notice_version%2019062019.pdf

This is not very user friendly, so we would like to embed this as an HTML link.

###### Changes:
* Use ng-bind-html to safely include HTML content in the Custom Login Info.
  > "Evaluates the expression and inserts the resulting HTML into the element in a secure way. By default, the resulting HTML content will be sanitized using the $sanitize service."
  > [https://docs.angularjs.org/api/ng/directive/ngBindHtml](https://docs.angularjs.org/api/ng/directive/ngBindHtml)

* In order to maintain backwards compatibility with the `white-space: pre-wrap` CSS, a simple check for HTML-like contents is added to only switch to `white-space: normal` if the content looks like HTML.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - UI

##### AWX VERSION

```
awx: 13.0.0
```

##### ADDITIONAL INFORMATION

A picture is worth a thousand words:
###### Before, in plain text:
Before
###### After, in basic HTML
jjPsUwlqqH

### DIFF
--- a/awx/ui/client/src/login/loginModal/loginModal.partial.html
+++ b/awx/ui/client/src/login/loginModal/loginModal.partial.html
@@ -98,7 +98,12 @@
                         </div>
                     </div>
                 </form>
-                <div id="login_modal_notice" class="LoginModalNotice" ng-if="customLoginInfoPresent"><div class="LoginModalNotice-title" translate>NOTICE</div>{{ customLoginInfo | sanitize }}</div>
+                <div id="login_modal_notice" class="LoginModalNotice" ng-if="customLoginInfoPresent">
+                    <div class="LoginModalNotice-title" translate>NOTICE</div>
+                    <ng-bind-html ng-bind-html="customLoginInfo"
+                        ng-style="{'white-space' : customLoginInfoIsHTML ? 'initial' : 'pre-wrap'}">
+                    </ng-bind-html>
+                </div>
             </div>
             <div class="LoginModal-footer">
                 <div class="LoginModal-footerBlock">
--- a/awx/ui/client/src/login/loginModal/loginModal.controller.js
+++ b/awx/ui/client/src/login/loginModal/loginModal.controller.js
@@ -88,6 +88,7 @@ export default ['$log', '$cookies', '$rootScope', 'ProcessErrors',
         }
         scope.customLoginInfo = $AnsibleConfig.custom_login_info;
         scope.customLoginInfoPresent = (scope.customLoginInfo) ? true : false;
+        scope.customLoginInfoIsHTML = /<\/?[a-z][\s\S]*>/i.test(scope.customLoginInfo);
     });
 
     if (scope.removeAuthorizationGetLicense) {
--- a/awx/ui/client/src/login/loginModal/loginModalNotice.block.less
+++ b/awx/ui/client/src/login/loginModal/loginModalNotice.block.less
@@ -12,7 +12,6 @@
     color: @login-notice-text;
     overflow-y: scroll;
     overflow-x: visible;
-    white-space: pre-wrap;
 }
 
 .LoginModalNotice-title {


awx/ui/client/src/login/loginModal/loginModal.partial.html
```diff
@@ -98,7 +98,11 @@
                         </div>
                     </div>
                 </form>
-                <div id="login_modal_notice" class="LoginModalNotice" ng-if="customLoginInfoPresent"><div class="LoginModalNotice-title" translate>NOTICE</div>{{ customLoginInfo | sanitize }}</div>
+                <div id="login_modal_notice" class="LoginModalNotice" ng-if="customLoginInfoPresent"
+                    ng-style="{'white-space' : (customLoginInfoIsHTML) ? 'normal': 'pre-wrap'}">
```
jakemcdermott: This changes the appearance of seemingly-not-html login info for me:

before:
Screenshot from 2020-07-09 09-45-10

after:
Screenshot from 2020-07-09 09-46-51

Interestingly, I noticed the text is rendered correctly if I just hard-code the `customLoginInfoIsHTML` to `true`, forcing it to ignore the `pre-wrap`.

jakemcdermott (changes requested):
Hey @philipsd6 

Thanks for the contribution. I think I'm on board with the idea of safely enabling html for custom login info. 

That said, will you please explain a little more about why we'd need to _add_ `pre-wrap` css for seemingly-non-html login info? You mentioned this was for backward compatibility purposes but it isn't clear to me what this would make us backward compatible _with_. Is this for prior awx/tower versions? browsers? etc.

awx/ui/client/src/login/loginModal/loginModal.partial.html
philipsd6: The reason I added the `customLoginInfoIsHTML` check to switch between `normal` and `pre-wrap` is because the original state of the element before this PR was `pre-wrap`, as you can see in [/awx/ui/client/src/login/loginModal/loginModalNotice.block.less#L15](https://github.com/ansible/awx/blob/d0ac02826537343880b09282666639cf5135779c/awx/ui/client/src/login/loginModal/loginModalNotice.block.less#L15).

(Technically, now that this PR sets the value, that line could be eliminated!)

I think I forgot to force-reload after my change to verify the plain text -- I'm seeing that it behaves oddly now as well. In your case, since you only had a single line of plain text, it didn't matter that you used `normal` white-space, but if you had multi-line plain text, it would wrap it into one line.

Trying to switch to `normal` was *my* design decision -- there's an argument to be made for leaving the CSS alone so that people could use simple `<em>` and `<b>` markup, and still have lines and wrapping done properly without having to use `<br />` or `<p>` tags. The only downside is that those of us who want to dump larger HTML fragments like this:
```html
<p>
Please review our <a href="https://global.intranet.company.com/homesites/GDPRHOME/GDPR%20documents/Company%20Internal%20Technologies%20Privacy%20Notice_version%2019062019.pdf">GDPR Privacy Statement</a>.
</p>
<p>
For your reference:
<ul>
<li><a href="https://global.intranet.company.com/confluence/x/oRQQw">User Guide</a</li>
<li><a href="https://global.intranet.company.com/confluence/display/AWX/FAQ">FAQ</a</li></ul>
</p>
```
would have to reformat them to be line-oriented like this:
```html
Please review our <a href="https://global.intranet.company.com/homesites/GDPRHOME/GDPR%20documents/Company%20Internal%20Technologies%20Privacy%20Notice_version%2019062019.pdf">GDPR Privacy Statement</a>.

For your reference:<ul><li><a href="https://global.intranet.company.com/confluence/x/oRQQw">User Guide</a</li><li><a href="https://global.intranet.company.com/confluence/display/AWX/FAQ">FAQ</a</li></ul>

philipsd6: let me know what you think @jakemcdermott, and I will look deeper into how to properly manage the css for this element.

philipsd6: Ok, I figured it out -- originally the custom info was templated directly into the `div#login_modal_notice` container, so the CSS needed to be applied to that element. Since I'm using ng-bind-html, a new element is created in that div, so that's where the CSS needs to apply. So I removed the style from `loginModalNotice.block.less` and set it on the `<ng-bind-html>` element dynamically instead.

There's now three commits, but I can squash them as you wish.

I've also rebased on latest devel.

jakemcdermott (approved):
@philipsd6 Looks good, thanks!

philipsd6:
whoops... I forgot to update the tooltip help for this setting! It still says custom HTML isn't supported, but now it is!

philipsd6:
I will open up a new PR to fix the help_text.

PR #7590, opened 2020-07-09
Adding hosts from smart inventories to dashboard count

john-westcott-iv:
##### SUMMARY

On the dashboard (and likely other places) hosts which are included in smart inventories do not appear in counts. This change corrects the dashboard.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API
 
##### AWX VERSION

```
awx: 13.0.0
```

##### ADDITIONAL INFORMATION

Create a smart inventory.
Create a new user in a different org.
Give that user read permissions to the smart inventory.
Load the users dashboard. The hosts = 0 inventories = 1.

After this change hosts = <number of hosts in inventory>

```

```

### DIFF
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -212,10 +212,20 @@ class DashboardView(APIView):
                           'inventory_failed': groups_inventory_failed}
 
         user_hosts = get_user_queryset(request.user, models.Host)
+        # Also include hosts from smart inventories that may be created from inventories the user does not have access too
+        unique_smart_hostnames = []
+        for inventory in user_inventory:
+            if inventory.kind == 'smart':
+                for host in inventory.hosts.filter(**{}).only(*['name']):
+                    if host.name not in unique_smart_hostnames:
+                        unique_smart_hostnames.append(host.name)
+        for host in user_hosts:
+            if host.name not in unique_smart_hostnames:
+                unique_smart_hostnames.append(host.name)
         user_hosts_failed = user_hosts.filter(last_job_host_summary__failed=True)
         data['hosts'] = {'url': reverse('api:host_list', request=request),
                          'failures_url': reverse('api:host_list', request=request) + "?last_job_host_summary__failed=True",
-                         'total': user_hosts.count(),
+                         'total': len(unique_smart_hostnames),
                          'failed': user_hosts_failed.count()}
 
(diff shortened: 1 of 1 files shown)


awx/api/views/__init__.py
```diff
@@ -212,10 +212,20 @@ def get(self, request, format=None):
                           'inventory_failed': groups_inventory_failed}
 
         user_hosts = get_user_queryset(request.user, models.Host)
+        # Also include hosts from smart inventories that may be created from inventories the user does not have access too
+        unique_smart_hostnames = []
+        for inventory in user_inventory:
+            if inventory.kind == 'smart':
+                for host in inventory.hosts.filter(**{}).only(*['name']):
```
wenottingham, line 219: I defer to eng, but I'd be worried about performance here.

ryanpetrello (changes requested):
In practice, smart inventory host calculation can be **very** slow, because users can define whatever (inperformant) query they like.  We're fairly aggressive about avoiding slow queries on the dashboard (it's gotten us into the trouble in the past performance-wise, because it's the first page users land on when they log in, and so a slow dashboard is both compounding, and _very_ noticeable), so I'm probably a 👎 on this idea.

ryanpetrello:
@john-westcott-iv 

We *do* have a feature that enables an automatic background build/calculation of smart inventory membership:

https://docs.ansible.com/ansible-tower/latest/html/userguide/inventories.html#smart-inventories

image

If you enable this, we keep around distinct `SmartInventoryMembership` records that are queryable in a performant way.  But it's off by default, so it's probably not something we can rely on for accuracy; I doubt most users are aware of it cc @AlanCoding @chrismeyersfsu

AlanCoding:
IIRC, we changed things so that we update that on job runs, so it might not be horrendously inaccurate.

I kind of see how that approach might work, but I don't have much appetite for fixing that, and it would obfuscate the meaning of the count as well.

```
Create a smart inventory.
Create a new user in a different org.
Give that user read permissions to the smart inventory.
Load the users dashboard. The hosts = 0 inventories = 1.
```

But does that mean that the superuser will see 2 hosts in the count? That could be problematic for the user to interpret it. I would want to rewind this to the original issue that people report, and very likely solve it with new fields, instead of changing what existing fields represent.

PR #7601, opened 2020-07-10
Update .gitignore to ignore translation files

nixocio:
See: https://lingui.js.org/guides/excluding-build-files.html


jakemcdermott:
talked about this offline w/ @nixocio 

He explained that there's actually a bit more to this than just modifying the `.gitignore`. We're going to track the work in a GH issue & explore the solution at a future date: https://github.com/ansible/awx/issues/7602

Issue #5236, opened 2019-11-04
Add sparkline of recent jobs to expanded host row


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
The awx-pf interface for hosts exposes recent jobs in a sparkline (like templates) but the one piece of information that we're missing is the job `type`.  The `type` is important because it allows us to navigate directly to the correct output page for the job in question without having to make an extra request.

I tried this chunk:

```
        d.setdefault('recent_jobs', [{
            'id': j.job.id,
            'name': j.job.job_template.name if j.job.job_template is not None else "",
            'status': j.job.status,
            'finished': j.job.finished,
            'type': j.job.get_real_instance_class()._meta.verbose_name.replace(' ', '_')
        } for j in obj.job_host_summaries.select_related('job__job_template').order_by('-created')[:5]])
```

in `def get_summary_fields(self, obj):` in the `HostSerializer` but wasn't sure if that was performant or correct.

mabashian:
If this isn't possible - there are some ways we can get around it.  We can set up a generic job endpoint that acts as a router and redirects to the correct output page after a GET request to the job (to get the type).

mabashian:
Note that these links:

<img width="1485" alt="Screen Shot 2019-11-04 at 4 37 22 PM" src="https://user-images.githubusercontent.com/9889020/68160292-7432ca00-ff21-11e9-9c09-ba9a89d6be16.png">

will be broken until we come up with a solution

unlikelyzero:
Extra network call
<img width="1911" alt="Screen Shot 2020-07-16 at 12 02 32 PM" src="https://user-images.githubusercontent.com/12446869/87694435-364a9e80-c75c-11ea-9108-116ae4b5c0db.png">

jakemcdermott:
https://github.com/ansible/awx/pull/7631#discussion_r455982679 :eye: :eye:

kdelee:
Unit tests added for this attribute in the api are sufficient for future regression testing, going to call this verified from the API perspective and remove the `component:api` label
tests: https://github.com/ansible/awx/pull/7611/files#diff-c967e9f6369e046cb79c1676024a5187a3f25f37f6d154320594b6ba3479b0cdR38

unlikelyzero:
Note: We can cover this feature by simply adding a feature test here https://github.com/ansible/awx/issues/5236#issuecomment-549558786

unlikelyzero:
[...]

PR #7611, opened 2020-07-13
add job type to Host summary_fields.recent_jobs

ryanpetrello:
see: https://github.com/ansible/awx/issues/5236

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -1697,6 +1697,7 @@ class HostSerializer(BaseSerializerWithVariables):
         d.setdefault('recent_jobs', [{
             'id': j.job.id,
             'name': j.job.job_template.name if j.job.job_template is not None else "",
+            'type': j.job.get_real_instance_class()._meta.verbose_name.replace(' ', '_'),
             'status': j.job.status,
             'finished': j.job.finished,
         } for j in obj.job_host_summaries.select_related('job__job_template').order_by('-created')[:5]])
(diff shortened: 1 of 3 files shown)


awx/api/serializers.py
```diff
@@ -1697,6 +1697,7 @@ def get_summary_fields(self, obj):
         d.setdefault('recent_jobs', [{
             'id': j.job.id,
             'name': j.job.job_template.name if j.job.job_template is not None else "",
+            'type': j.job.get_real_instance_class()._meta.verbose_name.replace(' ', '_'),
```
jakemcdermott: We do this here as well: https://github.com/ansible/awx/blob/devel/awx/main/models/unified_jobs.py#L1224

I'd imagine that these should _always_ emit the same keys so it might be worth it to just make this a property on the UJ model & add a simple test to check an instance of each job type against a list of expected values.

I could also see it being too early to tell and wanting to hold off. Up to you.

jakemcdermott (approved):

awx/api/serializers.py
ryanpetrello: Done!

unlikelyzero:
Running yolo

AlanCoding (approved):

chrismeyersfsu (approved):

unlikelyzero:
Tested:
- Verify event data refreshes the data displayed on the dashboard
- Verify type 'job' data is received by the browser websocket client
- Verify type 'inventory_update' data is received by the browser websocket client
- Verify type 'workflow_approval' data is received by the browser websocket client
- Verify type 'workflow_job' data is received by the browser websocket client
- Verify type 'project_update' data is received by the browser websocket client
- **Verify Slice Jobs appear as type 'job' and data is received by the browser websocket client**
- System Jobs do not have a type field

@jakemcdermott @ryanpetrello while testing this, I got myself into a scenario where a SJT was stuck in pending and I stopped receiving any websocket events. I can't repro but keep that in mind

PR #7614, opened 2020-07-13
Add pagination/item count to DataListToolbar

### LINKED ISSUE #5626: Add result count to top bar
We will be adding the result count to the top bar.

image

image (1)

When the screen size causes search/sort to be hidden inside the toggle-able search icon, the count should still be to the right of this (i.e. always displayed)

vjanssens:
### SUMMARY
This PR resolves #5626 by adding the result count to the DataListToolbar. It auto-updates the count when filters are applied, it is automatically shown on pages using `<DataList>` and remains visible on smaller devices.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

image

image

### DIFF
--- a/awx/ui_next/src/components/DataListToolbar/DataListToolbar.jsx
+++ b/awx/ui_next/src/components/DataListToolbar/DataListToolbar.jsx
@@ -37,7 +37,7 @@ class DataListToolbar extends React.Component {
       additionalControls,
       i18n,
       qsConfig,
-      pagination
+      pagination,
     } = this.props;
 
     const showExpandCollapse = onCompact && onExpand;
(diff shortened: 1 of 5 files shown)


jlmitch5:
thanks for the contribution!  I double-checked and there's been some slight shifting around to the requirement here, apologies for not updating the card, I didn't realize there was this change until just now.  It sounds like to be consistent with other apps using the PF design system, we want to utilize the compact pagination option, right aligned to the top of the toolbar.  would you be able to update your PR to use that instead?

https://www.patternfly.org/v4/documentation/react/components/pagination#compact

vjanssens:
No problem! It took me a bit, as I'm not very familiar with React but I was able to pass a Pagination component from the PaginatedDataList component down to ListHeader and down to DataListToolbar. Result is:

afbeelding

afbeelding

I do need to add/fix tests for this change. Can you let me know if I'm heading in the right direction?

dsesami (approved):
Looks good to me. I've gone around and asked to see if anyone else wants to approve this PR, if it works out then I'll just apply the `mergeit` tag, otherwise they can do so.

mabashian (approved):

PR #7621, opened 2020-07-15
Remove case sensitive hostcount

gamuniz:
##### SUMMARY

remove case sensitivity from active_hosts to give a more accurate host count

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
AWX 13.0
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/managers.py
+++ b/awx/main/managers.py
@@ -7,7 +7,7 @@ import os
 
 from django.db import models
 from django.conf import settings
-
+from django.db.models.functions import Lower
 from awx.main.utils.filters import SmartFilter
 from awx.main.utils.pglock import advisory_lock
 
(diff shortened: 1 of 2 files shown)


wenottingham:
At this point, we're unlikely to change this; this number doesn't drive behavior in any case.

tchellomello:
@wenottingham  I think we should revisit this since I many times had to use this query with customers as they want a more accurate statement of their hosts. We in support often stumble on phone calls explaining this to customers and having this merged would help a lot!

gforster:
Request to reopen - this does drive behavior in Tower for license count

PR #7624, opened 2020-07-15
Initial Execution Environment integration

shanemcd:


awx/main/tasks.py
```diff
@@ -2439,6 +2471,19 @@ def should_use_proot(self, project_update):
         '''
         return getattr(settings, 'AWX_PROOT_ENABLED', False)
 
+    def build_execution_environment_params(self, instance, cwd):
+        project_path = os.path.dirname(instance.get_project_path(check_if_exists=False))
+        execution_environment_params = {
+            "process_isolation": True,
+            "container_image": settings.AWX_EXECUTION_ENVIRONMENT_DEFAULT_IMAGE,
+            "container_volume_mounts": [
+                f"{project_path}:{project_path}",
```
AlanCoding: As per https://github.com/ansible/awx/issues/7625, after you rebase, you may be able to get things working by adding a mount to `instance. get_cache_path()`, mounting to same path like how you do with project path, like `f"{cache_path}:{cache_path}"`

I also don't really know what `cwd` is for project updates, and can't think of a reason why it would be needed.

awx/main/tasks.py
```diff
@@ -2120,7 +2134,25 @@ def _build_scm_url_extra_vars(self, project_update):
...
+        runner_inventory_folder = os.path.join(private_data_dir, 'inventory')
+        fn = os.path.join(runner_inventory_folder, 'hosts')
+        if not os.path.exists(runner_inventory_folder):
+            # Ansible Runner requires that this directory exists.
+            # Specifically, when using process isolation
+            os.mkdir(runner_inventory_folder)
+        with open(fn, 'w') as f:
+            os.chmod(fn, stat.S_IRUSR | stat.S_IXUSR | stat.S_IWUSR)
+            f.write('#! /usr/bin/env python3\n# -*- coding: utf-8 -*-\nprint(%r)\n' % json_data)
+        return fn
```
AlanCoding: I'd like to figure out how to do this more cleanly. In my own testing I've had pretty good luck just not including any hosts, because `ansible-playbook` doesn't need `-i` to work, and it automatically adds localhost.

jbradberry (approved):

awx/main/tasks.py
shanemcd: This has been addressed.

requirements/requirements_git.txt
```diff
@@ -0,0 +1 @@
+git+git://github.com/shanemcd/ansible-runner@ee-fact-caching#egg=ansible-runner
```
AlanCoding: I don't see any blockers in this branch https://github.com/ansible/ansible-runner/compare/devel...shanemcd:ee-fact-caching

Should we go ahead and see about getting this merged into the `devel` branch for runner 2.0?

awx/settings/defaults.py
```diff
@@ -66,6 +66,8 @@ def IS_TESTING(argv=None):
 AWX_CONTAINER_GROUP_DEFAULT_NAMESPACE = 'default'
 AWX_CONTAINER_GROUP_DEFAULT_IMAGE = 'ansible/ansible-runner'
 
+AWX_EXECUTION_ENVIRONMENT_DEFAULT_IMAGE = 'docker.io/ansible/ansible-runner:devel'
```
AlanCoding: I'm going to be submitting a patch to update this to the location of the new "AWX default execution environment". It may be that some tests are failing b/c this image lacks some needed stuff, but that's probably not more than a handful of them.

@kdelee just wanted to FYI ping you to point out this critical integration point.

awx/main/tasks.py
AlanCoding: I'm testing a patch that does `f"{project_path}:{project_path}:Z"` here to address a permissions error... we will see how that goes.

Issue #7511, opened 2020-07-01
Mobile friendly nav wacky when entries have the same name under different categories


chrismeyersfsu:
##### SUMMARY
Rows per page is not respected when copying resources.

##### STEPS TO REPRODUCE

* Shrink browser until mobile view
* On the nav, click Settings --> Jobs

##### EXPECTED RESULTS

* Only the `Settings` category to be expanded

##### ACTUAL RESULTS

* `Views` category and `Settings` expand

![image](https://user-images.githubusercontent.com/722880/86275735-b809cc00-bba1-11ea-9e5f-c9bea474df30.png)
![image](https://user-images.githubusercontent.com/722880/86275755-c526bb00-bba1-11ea-834b-8729ffcda438.png)

dsesami:
It looks like this works without issue now.

PR #7628, opened 2020-07-15
Improve nav behavior

jakemcdermott:
nav

**Use more selective route matching when determining if a nav item is active**
views/jobs and settings/jobs were colliding, addresses #7511 

**Don't automatically collapse nav groups when user navigates to a different group**
It shouldn't do this so I fixed it. Navbar sections should stay expanded until a user explicitly closes them. It works like the pf examples now https://www.patternfly.org/v4/documentation/react/components/nav#default-link3


keithjgrant (approved):
lgtm

PR #7631, opened 2020-07-16
Use summary job_type on host list

jakemcdermott:
##### SUMMARY
https://github.com/ansible/awx/issues/5236

### DIFF
--- a/awx/ui_next/src/screens/Host/HostList/HostListItem.jsx
+++ b/awx/ui_next/src/screens/Host/HostList/HostListItem.jsx
@@ -30,11 +30,6 @@ const DataListAction = styled(_DataListAction)`
 
 function HostListItem({ i18n, host, isSelected, onSelect, detailUrl }) {
   const labelId = `check-action-${host.id}`;
-  const recentPlaybookJobs = host.summary_fields.recent_jobs.map(job => ({
-    ...job,
-    type: 'job',
-  }));
-
   return (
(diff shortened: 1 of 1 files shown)


ryanpetrello (approved):

awx/ui_next/src/screens/Host/HostList/HostListItem.jsx
```diff
@@ -30,11 +30,6 @@ const DataListAction = styled(_DataListAction)`
 
 function HostListItem({ i18n, host, isSelected, onSelect, detailUrl }) {
   const labelId = `check-action-${host.id}`;
-  const recentPlaybookJobs = host.summary_fields.recent_jobs.map(job => ({
-    ...job,
-    type: 'job',
-  }));
-
```
jakemcdermott, line 37: @unlikelyzero This doesn't seem to change the two network calls we were seeing. That must be due to something else and it is not fixed in this PR. 

That said, you can still verify that this PR addresses the issue by hovering your mouse over the link and looking at the href that shows up:

host_link

(Look at bottom left)

With an old API + this change, you'll see "undefined".

With a new api + this change you'll see the correct job link.

jakemcdermott: The reason you can't check with an old api + old ui_next is because we were hard-coding the job_type, which I remove i this PR.

Issue #7557, opened 2020-07-06
List Instance Groups


nixocio:
ISSUE TYPE

    Feature Idea

SUMMARY

Add support for list Instance Groups

unlikelyzero:
@nixocio please reach out to @dsesami or myself when you're nearing completion

mabashian:
@nixocio I see that #7633 has merged, does that mean that this issue should be closed?

unlikelyzero:
@mabashian we're leaving this open to link back to the test stubs or net new functional tests

nixocio:
@unlikelyzero, I just closed 2 issues related to instance groups yesterday. I will re-open them.

dsesami:
This is looking really good, stubs are open and can track this, so this can be closed.

PR #7633, opened 2020-07-16
Add feature list instance groups

nixocio:
Add feature list instance groups.

Link: https://github.com/ansible/awx/issues/7557

image

### DIFF
--- a/awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupList.jsx
+++ b/awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupList.jsx
@@ -1,14 +1,213 @@
-import React from 'react';
+import React, { useEffect, useCallback } from 'react';
+import { useLocation, useRouteMatch } from 'react-router-dom';
+import { withI18n } from '@lingui/react';
+import { t } from '@lingui/macro';
 import { Card, PageSection } from '@patternfly/react-core';
 
-function InstanceGroupList() {
+import { InstanceGroupsAPI } from '../../../api';
+import { getQSConfig, parseQueryString } from '../../../util/qs';
+import useRequest, { useDeleteItems } from '../../../util/useRequest';
+import useSelected from '../../../util/useSelected';
+import PaginatedDataList, {
+  ToolbarDeleteButton,
+  ToolbarAddButton,
+} from '../../../components/PaginatedDataList';
+import ErrorDetail from '../../../components/ErrorDetail';
+import AlertModal from '../../../components/AlertModal';
+import DatalistToolbar from '../../../components/DataListToolbar';
+import InstanceGroupListItem from './InstanceGroupListItem';
+
+const QS_CONFIG = getQSConfig('instance_group', {
+  page: 1,
+  page_size: 20,
+});
+
+function modifyInstanceGroups(items = []) {
+  return items.map(item => {
+    const clonedItem = {
+      ...item,
+      summary_fields: {
+        ...item.summary_fields,
+        user_capabilities: {
+          ...item.summary_fields.user_capabilities,
+        },
+      },
+    };
+    if (clonedItem.name === 'tower') {
+      clonedItem.summary_fields.user_capabilities.delete = false;
+    }
+    return clonedItem;
+  });
+}
+
+function InstanceGroupList({ i18n }) {
+  const location = useLocation();
+  const match = useRouteMatch();
+
+  const {
+    error: contentError,
+    isLoading,
+    request: fetchInstanceGroups,
+    result: { instanceGroups, instanceGroupsCount, actions },
+  } = useRequest(
+    useCallback(async () => {
+      const params = parseQueryString(QS_CONFIG, location.search);
+
+      const [response, responseActions] = await Promise.all([
+        InstanceGroupsAPI.read(params),
+        InstanceGroupsAPI.readOptions(),
+      ]);
+
+      return {
+        instanceGroups: response.data.results,
+        instanceGroupsCount: response.data.count,
+        actions: responseActions.data.actions,
+      };
+    }, [location]),
+    {
+      instanceGroups: [],
+      instanceGroupsCount: 0,
+      actions: {},
+    }
+  );
+
+  useEffect(() => {
+    fetchInstanceGroups();
+  }, [fetchInstanceGroups]);
+
+  const { selected, isAllSelected, handleSelect, setSelected } = useSelected(
+    instanceGroups
+  );
+
+  const modifiedSelected = modifyInstanceGroups(selected);
+
+  const {
+    isLoading: deleteLoading,
+    deletionError,
+    deleteItems: deleteInstanceGroups,
+    clearDeletionError,
+  } = useDeleteItems(
+    useCallback(async () => {
+      await Promise.all(
+        selected.map(({ id }) => InstanceGroupsAPI.destroy(id))
+      );
+    }, [selected]),
+    {
+      qsConfig: QS_CONFIG,
+      allItemsSelected: isAllSelected,
+      fetchItems: fetchInstanceGroups,
+    }
+  );
+
+  const handleDelete = async () => {
+    await deleteInstanceGroups();
+    setSelected([]);
+  };
+
+  const canAdd = actions && actions.POST;
+
+  function cannotDelete(item) {
+    return !item.summary_fields.user_capabilities.delete;
+  }
+
+  const pluralizedItemName = i18n._(t`Instance Groups`);
+
+  let errorMessageDelete = '';
+
+  if (modifiedSelected.some(item => item.name === 'tower')) {
+    const itemsUnableToDelete = modifiedSelected
+      .filter(cannotDelete)
+      .filter(item => item.name !== 'tower')
+      .map(item => item.name)
+      .join(', ');
+
+    if (itemsUnableToDelete) {
+      if (modifiedSelected.some(cannotDelete)) {
+        errorMessageDelete = i18n._(
+          t`You do not have permission to delete ${pluralizedItemName}: ${itemsUnableToDelete}. `
+        );
+      }
+    }
+
+    if (errorMessageDelete.length > 0) {
+      errorMessageDelete = errorMessageDelete.concat('\n');
+    }
+    errorMessageDelete = errorMessageDelete.concat(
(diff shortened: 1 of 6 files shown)


awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupList.jsx
```diff
@@ -1,14 +1,209 @@
...
+  if (itemsUnableToDelete) {
+    if (modifiedSelected.some(cannotDelete)) {
+      errorMessageDelete = i18n._(
+        t`You do not have permission to delete ${pluralizedItemName}: ${itemsUnableToDelete}. `
+      );
+    }
+  }
+
+  if (modifiedSelected.some(item => item.name === 'tower')) {
+    if (errorMessageDelete.length > 0) {
```
AlexSCorey, line 132: Just need to check my thinking here.  This `if` block adds errorMessages that are not the error message that says you can't delete the Tower Instance group.  Is that right?

awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupList.jsx
```diff
@@ -1,14 +1,209 @@
...
+import AlertModal from '../../../components/AlertModal';
+import DatalistToolbar from '../../../components/DataListToolbar';
+import InstanceGroupListItem from './InstanceGroupListItem';
+
+const QS_CONFIG = getQSConfig('instance_group', {
+  page: 1,
+  page_size: 20,
+});
+
+const modifyInstanceGroups = (items = []) => {
```
AlexSCorey: You _could_ do this work inside the fetch request which would allow you to remove this block of code and that code on line 73.  I think doing that would make it a bit easier to maintain this file in the future as you won't have to move up and down the file to remember whats happening with the instance groups.

awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupList.jsx
```diff
@@ -1,14 +1,209 @@
...
+    setSelected([]);
+  };
+
+  const canAdd = actions && actions.POST;
+
+  function cannotDelete(item) {
+    return !item.summary_fields.user_capabilities.delete;
+  }
+
+  let itemsUnableToDelete = false;
```
AlexSCorey: I don't see where this is being used in this component

awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupListItem.jsx
```diff
@@ -0,0 +1,178 @@
...
+                  <Badge isRead>{instanceGroup.instances}</Badge>
+                </ListGroup>
+              ) : null}
+            </DataListCell>,
+
+            <DataListCell
+              key="capacity"
+              aria-label={i18n._(t`instance group used capacity`)}
+            >
+              {!isContainerGroup(instanceGroup) ? (
```
AlexSCorey: I think we should avoid nested ternary operators if we can. You can call a function here and return the JSX

AlexSCorey (changes requested):
This is a great start. Let me know if you have questions or want to discuss anything. I will probably be delayed as I'm on PTO Mon-Fri

awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupList.jsx
```diff
@@ -1,14 +1,209 @@
...
+                  <ToolbarDeleteButton
+                    key="delete"
+                    onDelete={handleDelete}
+                    itemsToDelete={modifiedSelected}
+                    pluralizedItemName={i18n._(t`Instance Groups`)}
+                    errorMessage={
+                      modifiedSelected.some(item => item.name === 'tower')
+                        ? errorMessageDelete
+                        : ''
+                    }
```
marshmalien: It looks like the prop `errorMessage` only returns `errorMessageDelete` when an instance group name is "tower". On lines ~115-130, the `if` statements perform many checks to build up the `errorMessageDelete` string, but since `errorMessage` only cares about one condition (name = "tower"), could we simplify how we build the error message?

awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupList.jsx
```diff
@@ -1,14 +1,209 @@
...
+
+const QS_CONFIG = getQSConfig('instance_group', {
+  page: 1,
+  page_size: 20,
+});
+
+const modifyInstanceGroups = (items = []) => {
+  return items.map(function updateDelete(item) {
+    if (item.name === 'tower') {
+      item.summary_fields.user_capabilities.delete = false;
```
marshmalien: Even though `.map()` returns a new Array, this line mutates the original instanceGroups object. We could avoid this by making a copy of the `items` array before updating it.

awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupList.jsx
nixocio: @AlexSCorey here. Are you mentioning `itemsUnableToDelete`? See line 115.

awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupList.jsx
nixocio: You are right.

awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupList.jsx
nixocio: @marshmalien, we need to cover 3 different use cases. 

- Only `tower` instance selected
- `tower` instance and another `foo` instance, 
-  just `foo` instance.

I treated the last case as a regular case that we already have in our code basis. 

I will try to refactor and simplify as per your suggestion.

awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupList.jsx
```diff
@@ -1,14 +1,209 @@
...
+    }, [selected]),
+    {
+      qsConfig: QS_CONFIG,
+      allItemsSelected: isAllSelected,
+      fetchItems: fetchInstanceGroups,
+    }
+  );
+
+  const handleDelete = async () => {
+    await handleDeleteInstanceGroups();
```
marshmalien: This is a little nitpicky -  since `handleDeleteInstanceGroups` isn't tied directly to a prop/event handler, I would drop "handle" from the beginning of the name.

awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupList.jsx
nixocio: I tried that one before it was convoluted.

awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupList.jsx
nixocio: Updated.

awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupList.jsx
nixocio: @marshmalien, I refactored this logic. If the `tower` instance group is not selected the `errorMessageDelete` will be empty, and `ToolbarDeleteButton` will handle the error message as we currently do for all other cases. If the `tower` instance is selected, then logic to create the error message will evaluate inside `InstanceGroupList`.

awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupList.jsx
nixocio: Updated.

awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupList.jsx
nixocio: Updated.

nixocio:
Thanks @AlexSCorey and @marshmalien  for your review. I updated this PR.  Ready to be reviewed again.

awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupListItem.jsx
nixocio: I will update this one as well.

unlikelyzero:
@nixocio for the utilization bar, is there clear way to tell what % the bar is rendering from a css, svg, or element id?

mabashian (approved):

nixocio:
> @nixocio for the utilization bar, is there clear way to tell what % the bar is rendering from a css, svg, or element id?

image

marshmalien (approved):
Awesome work here. This looks good to go after you update each item in the mock `instanceGroups` data in `InstanceGroupList.test.jsx` with a url key and value 👍

awx/ui_next/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupList.test.jsx
```diff
@@ -0,0 +1,190 @@
...
+
+const instanceGroups = {
+  data: {
+    results: [
+      {
+        id: 1,
+        name: 'Foo',
+        type: 'instance_group',
+        consumed_capacity: 10,
+        summary_fields: { user_capabilities: { edit: true, delete: true } },
```
marshmalien, line 23: To prevent a prop type error from showing up in the unit tests, we need to add a `url` key and value to each of these result items.

nixocio: Updated.

dsesami (approved):

AlexSCorey:
regate

AlexSCorey (approved):
Good work

Issue #5518, opened 2019-12-16
Cache the roles directory based on requirements.yml


mrmibrown:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Cache the roles directory based on requirements.yml to expedite jobs with larger repositories.

wenottingham:
@AlanCoding do you have idea on how this would be done?

AlanCoding:
Yes, I really wanted to do this in spare time, but I'm too far behind at the moment.

To correctly handle different branches, it makes sense to organize by the commit hash. Either project updates or project syncs could run the command, and then save the role contents in a folder specific to the project and specific to the commit (not in the project source tree).

This wouldn't change the job folder structure that we have now, but it would change user experience with updating projects and debugging Galaxy-related errors (hopefully for the better).

There are certainly more details to nit pick over, but in general terms I don't think there should be any problem with this.

wenottingham:
Well, I'd point immediately at the cache invalidation/cleanup issue, especially when it comes to promptable branches.

AlanCoding:
that's a fair point, and because of that it might not be realistic to cache anything other than default branches.

n2aws:
I see this is labelled as an enhancement, but I think we should consider the current behavior in 9.1.0, 9.1.1, and 9.2.0 to be a "bug" as the change broke how roles/requirements.yml works, and several people have entire jobs that broke with this.

AlanCoding:
support for collections from SCM just merged

https://github.com/ansible/ansible/pull/69154

trying to think of how that could impact any solution here.

AlanCoding:
This is being merged, I'm leaving the issue open for @elyezer to close when ready

elyezer:
This has been verified and works as expected.

PR #5429, opened 2019-12-03
Allow for skipping project sync for SCM inventory updates

AlanCoding:
##### SUMMARY
If you have a project that is already fully-synced with the remote locally, it won't do anything when you launch a job using it. The same pre-check was not implemented for inventory updates just due to sheer time constraints and aversion to more changes.

This makes it so that inventory updates can skip their project sync if all the boxes are checked.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
9.0.1
```

##### ADDITIONAL INFORMATION
Let me get some screenshots

Screen Shot 2019-12-03 at 10 01 33 AM

This still has a bit of an issue in that it doesn't link to the `source_project`, that may require either API or UI changes.

There is also more consolidation of logic that should be done here.

Issue #7625, opened 2020-07-15
Handle project and galaxy caching in Execution Environments


shanemcd:
##### ISSUE TYPE
 - Feature Idea

We need to ensure the work done in https://github.com/ansible/awx/issues/5518 does not regress once we move to Execution Environments.

AlanCoding:
I hope that I now have https://github.com/ansible/awx/pull/7643 stabilized. Let me give a synopsis of the interfaces here.

The playbook variables `roles_destination` and `collections_destination` are being removed in that pull request. Those were destinations in a job folder, which we had to give access to via bubblewrap. I want to be very specific about this - project updates should only need special access to the projects folder.

The way that's done is exposing `settings.PROJECTS_ROOT` as a bubblewrap show path. This gives access to all project folders, which might not be ideal. If you wanted to be more explicit, you could give it access to _only_ the folders:

 - `/var/lib/awx/projects/_42__project_name`
 - `/var/lib/awx/projects/.__awx_cache/_42__project_name`

The first one is the project folder, which has been a thing for as long as Ansible Tower has. The latter has everything we need for role & collection caching. Obviously, you should expect the project update to be writing to and making changes to these folders.

In my initial implementation (now abandoned), I hit one surprising issue which I want to disclose:

```
awx_1       | 2020-07-17 20:42:54,869 ERROR    awx.main.tasks project_update 93 (running) Post run hook errored.
awx_1       | Traceback (most recent call last):
awx_1       |   File "/awx_devel/awx/main/tasks.py", line 1499, in run
awx_1       |     self.post_run_hook(self.instance, status)
awx_1       |   File "/awx_devel/awx/main/tasks.py", line 2410, in post_run_hook
awx_1       |     os.rename(stage_subpath, dest_subpath)
awx_1       | OSError: [Errno 18] Invalid cross-device link: '/var/lib/awx/projects/.__awx_cache/_67__branch_override_with_different_role_requirements_/stage/requirements_roles' -> '/tmp/awx_92_7a3rost5/requirements_roles'
```

This came from the fact that our _development environment maps the project folder_, and I tried to use `os.rename` to move some cache content from inside the project folder to the job private data directory under `/tmp`. That's not allowed, you can copy across volumes, but you can't _move_ it.

Obvious solution to that - don't design anything to move stuff between folders which might be in different volumes. We shouldn't be doing any of that with the updated PR. Even if you mounted separate volumes for the project folder and the project's cache folder, I don't anticipate problems.

[...]

PR #7643, opened 2020-07-17
Implement per-update cache for roles and collections

AlanCoding:
##### SUMMARY
This is a direct continuation of https://github.com/ansible/awx/pull/7194

Connect https://github.com/ansible/awx/issues/5518

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
13.0.0
```

##### ADDITIONAL INFORMATION
I have to be really picky about what to refactor here.

See #5429, that has a lot of background for some of the patterns here. I think `make_local_copy` should be a method on the project model. I'm trying to allow for that in the future, but keep changes here minimal.

That PR also may reflect some shortcomings of this which still remain. For SCM inventory, roles might not be getting installed in some cases, and that still needs some investigation.

### DIFF
--- a/awx/main/models/projects.py
+++ b/awx/main/models/projects.py
@@ -604,9 +604,8 @@ class ProjectUpdate(UnifiedJob, ProjectOptions, JobNotificationMixin, TaskManage
         added_update_fields = []
         if not self.job_tags:
             job_tags = ['update_{}'.format(self.scm_type)]
-            if self.job_type == 'run':
-                job_tags.append('install_roles')
-                job_tags.append('install_collections')
+            job_tags.append('install_roles')
+            job_tags.append('install_collections')
             self.job_tags = ','.join(job_tags)
             added_update_fields.append('job_tags')
         if self.scm_delete_on_update and 'delete' not in self.job_tags and self.job_type == 'check':
--- a/awx/main/models/projects.py
+++ b/awx/main/models/projects.py
@@ -424,6 +424,10 @@ class Project(UnifiedJobTemplate, ProjectOptions, ResourceMixin, CustomVirtualEn
         return False
 
     @property
+    def cache_id(self):
+        return str(self.last_job_id)
+
+    @property
     def notification_templates(self):
         base_notification_templates = NotificationTemplate.objects
         error_notification_templates = list(base_notification_templates
@@ -560,6 +564,19 @@ class ProjectUpdate(UnifiedJob, ProjectOptions, JobNotificationMixin, TaskManage
     def result_stdout_raw(self):
         return self._result_stdout_raw(redact_sensitive=True)
 
+    @property
+    def branch_override(self):
+        """Whether a branch other than the project default is used."""
+        if not self.project:
+            return True
+        return bool(self.scm_branch and self.scm_branch != self.project.scm_branch)
+
+    @property
+    def cache_id(self):
+        if self.branch_override or self.job_type == 'check' or (not self.project):
+            return str(self.id)
+        return self.project.cache_id
+
     def result_stdout_raw_limited(self, start_line=0, end_line=None, redact_sensitive=True):
         return self._result_stdout_raw_limited(start_line, end_line, redact_sensitive=redact_sensitive)
 
@@ -603,9 +620,7 @@ class ProjectUpdate(UnifiedJob, ProjectOptions, JobNotificationMixin, TaskManage
     def save(self, *args, **kwargs):
         added_update_fields = []
         if not self.job_tags:
-            job_tags = ['update_{}'.format(self.scm_type)]
-            job_tags.append('install_roles')
-            job_tags.append('install_collections')
+            job_tags = ['update_{}'.format(self.scm_type), 'install_roles', 'install_collections']
             self.job_tags = ','.join(job_tags)
             added_update_fields.append('job_tags')
         if self.scm_delete_on_update and 'delete' not in self.job_tags and self.job_type == 'check':
--- a/awx/main/models/projects.py
+++ b/awx/main/models/projects.py
@@ -194,6 +194,11 @@ class ProjectOptions(models.Model):
             if not check_if_exists or os.path.exists(smart_str(proj_path)):
                 return proj_path
 
+    def get_cache_path(self):
+        local_path = os.path.basename(self.local_path)
+        if local_path:
+            return os.path.join(settings.PROJECTS_ROOT, '.__awx_cache', local_path)
+
     @property
     def playbooks(self):
         results = []
@@ -455,11 +460,12 @@ class Project(UnifiedJobTemplate, ProjectOptions, ResourceMixin, CustomVirtualEn
         )
 
     def delete(self, *args, **kwargs):
-        path_to_delete = self.get_project_path(check_if_exists=False)
+        paths_to_delete = (self.get_project_path(check_if_exists=False), self.get_cache_path())
         r = super(Project, self).delete(*args, **kwargs)
-        if self.scm_type and path_to_delete:  # non-manual, concrete path
-            from awx.main.tasks import delete_project_files
-            delete_project_files.delay(path_to_delete)
+        for path_to_delete in paths_to_delete:
+            if self.scm_type and path_to_delete:  # non-manual, concrete path
+                from awx.main.tasks import delete_project_files
+                delete_project_files.delay(path_to_delete)
         return r
 
 
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -1865,44 +1865,32 @@ class RunJob(BaseTask):
         project_path = job.project.get_project_path(check_if_exists=False)
         job_revision = job.project.scm_revision
         sync_needs = []
-        all_sync_needs = ['update_{}'.format(job.project.scm_type), 'install_roles', 'install_collections']
+        source_update_tag = 'update_{}'.format(job.project.scm_type)
+        branch_override = bool(job.scm_branch and job.scm_branch != job.project.scm_branch)
         if not job.project.scm_type:
             pass # manual projects are not synced, user has responsibility for that
         elif not os.path.exists(project_path):
             logger.debug('Performing fresh clone of {} on this instance.'.format(job.project))
-            sync_needs = all_sync_needs
-        elif not job.project.scm_revision:
-            logger.debug('Revision not known for {}, will sync with remote'.format(job.project))
-            sync_needs = all_sync_needs
-        elif job.project.scm_type == 'git':
+            sync_needs.append(source_update_tag)
+        elif job.project.scm_type == 'git' and job.project.scm_revision and (not branch_override):
             git_repo = git.Repo(project_path)
             try:
-                desired_revision = job.project.scm_revision
-                if job.scm_branch and job.scm_branch != job.project.scm_branch:
-                    desired_revision = job.scm_branch  # could be commit or not, but will try as commit
-                current_revision = git_repo.head.commit.hexsha
-                if desired_revision == current_revision:
-                    job_revision = desired_revision
+                if job_revision == git_repo.head.commit.hexsha:
                     logger.debug('Skipping project sync for {} because commit is locally available'.format(job.log_format))
                 else:
-                    sync_needs = all_sync_needs
+                    sync_needs.append(source_update_tag)
             except (ValueError, BadGitName):
                 logger.debug('Needed commit for {} not in local source tree, will sync with remote'.format(job.log_format))
-                sync_needs = all_sync_needs
+                sync_needs.append(source_update_tag)
         else:
-            sync_needs = all_sync_needs
-        # Galaxy requirements are not supported for manual projects
-        if not sync_needs and job.project.scm_type:
-            # see if we need a sync because of presence of roles
-            galaxy_req_path = os.path.join(project_path, 'roles', 'requirements.yml')
-            if os.path.exists(galaxy_req_path):
-                logger.debug('Running project sync for {} because of galaxy role requirements.'.format(job.log_format))
-                sync_needs.append('install_roles')
+            logger.debug('Project not available locally {}, will sync with remote'.format(job.project))
+            sync_needs.append(source_update_tag)
 
-            galaxy_collections_req_path = os.path.join(project_path, 'collections', 'requirements.yml')
-            if os.path.exists(galaxy_collections_req_path):
-                logger.debug('Running project sync for {} because of galaxy collections requirements.'.format(job.log_format))
-                sync_needs.append('install_collections')
+        cache_id = str(job.project.last_job_id)  # content cache - for roles and collections
+        has_cache = os.path.exists(os.path.join(job.project.get_cache_path(), cache_id))
+        # Galaxy requirements are not supported for manual projects
+        if job.project.scm_type and ((not has_cache) or branch_override):
+            sync_needs.extend(['install_roles', 'install_collections'])
 
         if sync_needs:
             pu_ig = job.instance_group
@@ -1920,7 +1908,7 @@ class RunJob(BaseTask):
                 execution_node=pu_en,
                 celery_task_id=job.celery_task_id
             )
-            if job.scm_branch and job.scm_branch != job.project.scm_branch:
+            if branch_override:
                 sync_metafields['scm_branch'] = job.scm_branch
             if 'update_' not in sync_metafields['job_tags']:
                 sync_metafields['scm_revision'] = job_revision
@@ -1952,10 +1940,7 @@ class RunJob(BaseTask):
             if job_revision:
                 job = self.update_model(job.pk, scm_revision=job_revision)
             # Project update does not copy the folder, so copy here
-            RunProjectUpdate.make_local_copy(
-                project_path, os.path.join(private_data_dir, 'project'),
-                job.project.scm_type, job_revision
-            )
+            RunProjectUpdate.make_local_copy(job.project, private_data_dir, scm_revision=job_revision)
 
         if job.inventory.kind == 'smart':
             # cache smart inventory memberships so that the host_filter query is not
@@ -1995,10 +1980,7 @@ class RunProjectUpdate(BaseTask):
 
     @property
     def proot_show_paths(self):
-        show_paths = [settings.PROJECTS_ROOT]
-        if self.job_private_data_dir:
-            show_paths.append(self.job_private_data_dir)
-        return show_paths
+        return [settings.PROJECTS_ROOT]
 
     def __init__(self, *args, job_private_data_dir=None, **kwargs):
         super(RunProjectUpdate, self).__init__(*args, **kwargs)
@@ -2165,8 +2147,7 @@ class RunProjectUpdate(BaseTask):
         extra_vars.update(extra_vars_new)
 
         scm_branch = project_update.scm_branch
-        branch_override = bool(scm_branch and project_update.scm_branch != project_update.project.scm_branch)
-        if project_update.job_type == 'run' and (not branch_override):
+        if project_update.job_type == 'run' and (not project_update.branch_override):
             if project_update.project.scm_revision:
                 scm_branch = project_update.project.scm_revision
             elif not scm_branch:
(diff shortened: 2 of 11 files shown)


awx/main/tasks.py
```diff
@@ -1997,10 +2001,7 @@ class RunProjectUpdate(BaseTask):
 
     @property
     def proot_show_paths(self):
-        show_paths = [settings.PROJECTS_ROOT]
-        if self.job_private_data_dir:
-            show_paths.append(self.job_private_data_dir)
-        return show_paths
+        return [settings.PROJECTS_ROOT]
```
wenottingham, line 1982: ?

AlanCoding: ? as in, "why did remove the job dir here" or ? as in, "why did use `@property` when I could have make it an actual property?"

For the latter question, the answer is CTiT, I don't want it to be looking up the setting on load of the module.

The former question is much bigger. I removed the `synchronize` module from the playbook. By doing this, the duties of the project update playbook become much more static, because given a certain project, it always downloads to the same "stage" folder for the roles/collections. Thinking there relates to https://github.com/ansible/awx/issues/7625, I want to have a project update interface that works for that, while also working for caching roles / collections. The project cache folder is already a subfolder of the PROJECTS_ROOT folder, so it no longer needs to be added here.

I'm going to be writing docs on this. The overall function will be much easier to explain with these changes.

awx/playbooks/project_update.yml
```diff
@@ -1,6 +1,9 @@
 ---
 # The following variables will be set by the runner of this playbook:
-# project_path: PROJECTS_DIR/_local_path_
+# projects_root: Global location for caching project checkouts and roles and collections
+#     should not have trailing slash on end
+# local_path: Path within projects_root to use for this project
+# project_path: projects_root/local_path
```
AlanCoding: My intent here is that `project_path` will be removed, replaced with `{{projects_root}}/{{local_path}}` in the playbook. After that, the playbook will no longer document this variable, or have it passed in extra_vars.

I just didn't want to do that in this diff, but I could be convinced otherwise.

AlanCoding:
I do have a few other corner case concerns:

 - [x] the project update `pre_run_hook` (with lock held) should assure that the "stage" directory doesn't exist. It never should, so if it does, it should log warnings
 - [x] we _might_ need to clear the cache more aggressively for the branch override cases, because it could leave an unlimited number of caches around. This might be workable by clearing all but `last_job_id` in both cases...

I'm still trying to challenge myself to come up with more potential issues. I feel much better about SCM inventory syncs after the last few commits, I expect it will probably work right, but might give erroneous warnings, not fully utilize the cache, and still be rough around the edges. I would hate to wind up effectively re-doing #5429, but I think it may come to that.

AlanCoding:
My only remaining issue with the code as it stands is that the logic for needs roles/collections is much more complicated than it needs to be to cover cases we would actually hit. I may or may not make another change for that.

AlanCoding:
I think that this is cleaner, but it should make any functional difference, and I won't put it in this branch until I'm sure it will not break things.

```diff
diff --git a/awx/main/tasks.py b/awx/main/tasks.py
index 6bad842bce..f9ed09ab28 100644
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -1872,48 +1872,34 @@ class RunJob(BaseTask):
         elif not os.path.exists(project_path):
             logger.debug('Performing fresh clone of {} on this instance.'.format(job.project))
             sync_needs = all_sync_needs
-        elif not job.project.scm_revision:
-            logger.debug('Revision not known for {}, will sync with remote'.format(job.project))
-            sync_needs = all_sync_needs
-        elif job.project.scm_type == 'git':
+        elif job.project.scm_type == 'git' and job.project.scm_revision and (not branch_override):
             git_repo = git.Repo(project_path)
             try:
-                desired_revision = job.project.scm_revision
-                if branch_override:
-                    desired_revision = job.scm_branch  # could be commit or not, but will try as commit
-                current_revision = git_repo.head.commit.hexsha
-                if desired_revision == current_revision:
-                    job_revision = desired_revision
+                if job_revision == git_repo.head.commit.hexsha:
                     logger.debug('Skipping project sync for {} because commit is locally available'.format(job.log_format))
+                    cache_id = str(job.project.last_job_id)
+                    has_cache = os.path.exists(os.path.join(job.project.get_cache_path(), cache_id))
+                    if has_cache:
+                        logger.debug('Using cached roles or collections for {}'.format(job.log_format))
+                    else:
+                        # see if we need a sync because of presence of roles
+                        roles_exist = os.path.exists(os.path.join(project_path, 'roles', 'requirements.yml'))
+                        if roles_exist:
+                            logger.debug('Running project sync for {} for role requirements.'.format(job.log_format))
+                            sync_needs.append('install_roles')
+
+                        collections_exist = os.path.exists(os.path.join(project_path, 'collections', 'requirements.yml'))
+                        if collections_exist:
+                            logger.debug('Running project sync for {} for collection requirements.'.format(job.log_format))
+                            sync_needs.append('install_collections')
                 else:
                     sync_needs = all_sync_needs
             except (ValueError, BadGitName):
                 logger.debug('Needed commit for {} not in local source tree, will sync with remote'.format(job.log_format))
                 sync_needs = all_sync_needs
         else:
+            logger.debug('Project not available locally {}, will sync with remote'.format(job.project))
             sync_needs = all_sync_needs
-        # Galaxy requirements are not supported for manual projects
-        if not sync_needs and job.project.scm_type:
-            has_cache = False
-            if not branch_override:
-                cache_id = str(job.project.last_job_id)
-                has_cache = os.path.exists(os.path.join(job.project.get_cache_path(), cache_id))
-            # see if we need a sync because of presence of roles
-            galaxy_req_path = os.path.join(project_path, 'roles', 'requirements.yml')
-            roles_exist = os.path.exists(galaxy_req_path)
-            if roles_exist and not has_cache:
-                logger.debug('Running project sync for {} because of galaxy role requirements.'.format(job.log_format))
-                sync_needs.append('install_roles')
-            elif roles_exist:
-                logger.debug('Using cached roles for {}'.format(job.log_format))
-
-            galaxy_collections_req_path = os.path.join(project_path, 'collections', 'requirements.yml')
-            collections_exist = os.path.exists(galaxy_collections_req_path)
-            if collections_exist and not has_cache:
-                logger.debug('Running project sync for {} because of galaxy collections requirements.'.format(job.log_format))
-                sync_needs.append('install_collections')
-            elif collections_exist:
-                logger.debug('Using cached collections for {}'.format(job.log_format))
 
         if sync_needs:
             pu_ig = job.instance_group
@@ -2423,7 +2409,7 @@ class RunProjectUpdate(BaseTask):
                 instance.save(update_fields=['scm_revision'])
             # Roles and collection folders copy - both to durable cache and job dir
             stage_path = os.path.join(instance.get_cache_path(), 'stage')
-            if instance.branch_override:
+            if instance.branch_override or instance.job_type == 'check':
                 # We actually do not want to cache these cases, so we just set an id
                 # which we are sure that nothing will reference for its cache
                 cache_id = str(instance.id)
```

AlanCoding:
Did another bit of a pivot, some major items changed:

 - [x] always create the stage dir in pre_run_hook
 - [x] always sync both roles and collections if either
        branch_override, or cache dir does not exist
 - [x] (need either way) only move stage to cache if successful
 - [x] change creation of stage to mkdirs
 - [x] do not populate cache if did not run galaxy tags

awx/main/tasks.py
wenottingham: Thanks, it was the former question, it wasn't obvious to me from looking at it.

awx/main/tasks.py
```diff
@@ -2322,27 +2305,58 @@ def pre_run_hook(self, instance, private_data_dir):
...
+            shutil.rmtree(stage_path)
+        # Important - the presence of an empty cache will indicate that a given
+        # project revision did not have any roles or collections
+        os.makedirs(stage_path)
 
     @staticmethod
-    def make_local_copy(project_path, destination_folder, scm_type, scm_revision):
-        if scm_type == 'git':
+    def clear_project_cache(cache_dir, keep_value):
+        if os.path.isdir(cache_dir):
```
AlanCoding, line 2315: You could also test `os.path.exists`, but I don't think it should matter at all. I really don't expect this check is needed given the other changes. However, I want to keep the check as a defensive measure. The contract may change in the future such that the stage dir is not always populated. Or error conditions could cause that...

awx/main/models/projects.py
```diff
@@ -597,10 +620,7 @@ def preferred_instance_groups(self):
     def save(self, *args, **kwargs):
         added_update_fields = []
         if not self.job_tags:
-            job_tags = ['update_{}'.format(self.scm_type)]
-            if self.job_type == 'run':
-                job_tags.append('install_roles')
-                job_tags.append('install_collections')
+            job_tags = ['update_{}'.format(self.scm_type), 'install_roles', 'install_collections']
```
AlanCoding, line 623: Let me get this out of the way:

With role/collection caching in place, it no longer makes _any_ sense to have 2 separate tags for these things. It should just be 1 tag, say "install_content". If you follow all the code here, you will see that one of them is never used without the other.

I am not consolidating those tags right now, because `job_tags` are a primary part of the model and something that the user sees. In the effort to maintain behavior consistency I'm not changing it now, but after this PR, we could/should make that change for `devel` branch.

awx/main/tasks.py
```diff
@@ -2632,7 +2675,7 @@ def pre_run_hook(self, inventory_update, private_data_dir):
                 _eager_fields=dict(
                     launch_type="sync",
                     job_type='run',
-                    job_tags='update_{},install_collections'.format(source_project.scm_type),  # roles are never valid for inventory
+                    job_tags='update_{},install_roles,install_collections'.format(source_project.scm_type),  # roles are never valid for inventory
```
AlanCoding: I'm still on the fence about the choices made here.

Background on change - here we are caching _content_, roles & collections, and marking that cache with an identifier. SCM inventory cannot use roles, only collections. But... an SCM inventory could trigger a local cache update, which later gets used by a job. So yes, we do need to update roles here, just to keep the cache mechanics simple.

However, I'm still not detecting whether the cache exists or not. That would be getting into what #5429 was trying to do. This means that the cache would be blown away and both roles & collections re-downloaded for every SCM inventory update. This is gory because right now it would leave a warning log:

https://github.com/ansible/awx/pull/7643/files#diff-9d4ea1dd908b35fb92eaede4bd10bb46R2409

So options are:
1. downgrade the `warn` log to `debug` or `info`, accept that we're not solving the caching problem for SCM inv
2. check if cache exists here, like in the `RunJob` pre_run_hook

This is a pained decision for me right now. I don't want to creep scope further, but I want this to make sense.


Since the SCM inventory obliged to download _roles_, I'm leaning toward option (2), because it would be seen as a regression for it to start doing 20 GB downloads in all these cases where the project does not even have collection requirements.

awx/main/tests/functional/test_tasks.py
```diff
@@ -30,7 +30,7 @@ class TestDependentInventoryUpdate:
     def test_dependent_inventory_updates_is_called(self, scm_inventory_source, scm_revision_file):
         task = RunProjectUpdate()
         task.revision_path = scm_revision_file
-        proj_update = ProjectUpdate.objects.create(project=scm_inventory_source.source_project)
+        proj_update = scm_inventory_source.source_project.create_project_update()
```
AlanCoding, line 33: We can't do `objects.create()` because that won't set `local_path`. It has to be created via the conical means of creating it _from_ the project.

awx/main/tests/unit/test_tasks.py
```diff
@@ -61,7 +61,10 @@ def patch_Job():
 
 @pytest.fixture
 def job():
-    return Job(pk=1, id=1, project=Project(), inventory=Inventory(), job_template=JobTemplate(id=1, name='foo'))
+    return Job(
+        pk=1, id=1,
+        project=Project(local_path='/projects/_23_foo'),
```
AlanCoding, line 66: Again, `local_path` _has_ to be set for all non-manual projects. This is an oddball case because it's an object not even saved to the database, so our normal means of setting it won't work.

AlanCoding:
I filed a related issue at https://github.com/ansible/awx/issues/7665

I've been diving into the `local_cache` problem because `local_path = os.path.basename(self.local_path)` causes me great anxiety. In two seconds flat, I'm sure the user would input `foo/`, which will give empty string, and potentially damage the integrity of the projects _root_ folder.

For the purposes of this PR, I have satisfied myself that it is automatically set by the API. I don't think there is any legacy data that might cause an issue, since most of this logic has been around for a very long time.

AlanCoding:
@tvo318 I added some docs here, but are very technical.

For the user's experience, @gamuniz noted that this detail is something we should call out - my attempted wording:

> if the user hits the "update" button for a project, then the update will re-download roles even if the revision of the project did not change.

You might also add "otherwise, jobs will use roles cached from the last update", or something along those lines.

*everything also applies to collections. We might start saying content=(roles and collections), but I defer to your judgement there.

AlanCoding:
@chrismeyersfsu is there any reason to keep these lines around?

https://github.com/ansible/awx/blob/b36c5d920993a8b06ea2e3f38fcccac87a13bc15/awx/main/tasks.py#L2036-L2040

I don't think they do what they claim to do, but I don't really know what's up with them. My preference would be to delete them.

AlanCoding:
Things that we could do later, targeting `devel`:

 - ~do not have project syncs call `make_local_copy`, do in job/inventory task~
   ~do not pass job_private_data_dir to project update at all, it is none of its business~
   (no solution for non-git, I take it back)
 - consolidate logic to check for sync needs in job/inventory tasks
 - combine install_roles and install_collections tags into single tag
 - remove `project_path` from project_update.yml playbook variables, just use its components
 - copy instead of doing a shallow git clone #7367
 - simplify logic for project `local_path` setting

After this merges, I should probably refresh these and file an issue.

AlanCoding:
I've thought about https://github.com/ansible/awx/pull/7643#issuecomment-661318048 more... no good can come from those lines of code. It was put there back when we assumed we were passing that directory to Galaxy CLI... we're not. Now we copy the folders from the cache. It gets worse. Sometimes we do hit this code path before a job, but other times we don't, so sometimes we're merging directories with different filesystem permissions. It's nothing but a liability. It must be destroyed.

AlanCoding:
Tinkering with doing the things I mentioned in https://github.com/ansible/awx/pull/7643#issuecomment-661318780, it would look like this (have not ran any tests):

https://github.com/AlanCoding/awx/commit/02cda85d4015d3c49a25b75f0fb6572e5f7fab92

This convinces me that _now_ is not the right time to do these things. The most major expected gains are still not really possible because we support branch override for non-git types. That is something we could consider. Consolidating the sync needs code just makes it more obvious that we need to consolidate the code for _creating and running the sync_.

awx/main/tasks.py
```diff
@@ -2322,27 +2295,58 @@ def pre_run_hook(self, instance, private_data_dir):
...
             project_path = instance.project.get_project_path(check_if_exists=False)
             if os.path.exists(project_path):
                 git_repo = git.Repo(project_path)
                 if git_repo.head.is_detached:
                     self.original_branch = git_repo.head.commit
                 else:
                     self.original_branch = git_repo.active_branch
+        stage_path = os.path.join(instance.get_cache_path(), 'stage')
+        if os.path.exists(stage_path):
+            logger.warning('{0} cache staging area unexpectedly existed before update.')
```
AlanCoding: whoops! missing the `.format`

awx/main/tasks.py
```diff
@@ -2322,8 +2295,7 @@ def pre_run_hook(self, instance, private_data_dir):
             os.mkdir(settings.PROJECTS_ROOT)
         self.acquire_lock(instance)
         self.original_branch = None
-        if (instance.scm_type == 'git' and instance.job_type == 'run' and instance.project and
-                instance.scm_branch != instance.project.scm_branch):
+        if instance.scm_type == 'git' and instance.branch_override:
```
AlanCoding, line 2298: Explanation: this is changed due to introduction of new property `branch_override`, and it's also true that `branch_override=True` implies `instance.job_type == 'run'` so I removed that from the condition.

awx/main/models/projects.py
```diff
@@ -194,6 +194,11 @@ def get_project_path(self, check_if_exists=True):
             if not check_if_exists or os.path.exists(smart_str(proj_path)):
                 return proj_path
 
+    def get_cache_path(self):
+        local_path = os.path.basename(self.local_path)
+        if local_path:
```
shanemcd, line 199: Should we put a `logger.warn` in an `else` here?

ryanpetrello: Or perhaps a `.debug()`, `warn` might be too noisy.

awx/main/models/projects.py
```diff
@@ -418,6 +423,10 @@ def needs_update_on_launch(self):
                 return True
         return False
 
+    @property
+    def cache_id(self):
+        return str(self.last_job_id)
```
chrismeyersfsu, line 428: This will never be a project `sync` job right?

awx/main/tasks.py
```diff
@@ -2332,17 +2304,48 @@ def pre_run_hook(self, instance, private_data_dir):
...
     @staticmethod
-    def make_local_copy(project_path, destination_folder, scm_type, scm_revision):
-        if scm_type == 'git':
+    def clear_project_cache(cache_dir, keep_value):
+        if os.path.isdir(cache_dir):
+            for entry in os.listdir(cache_dir):
+                old_path = os.path.join(cache_dir, entry)
+                if entry not in (keep_value, 'stage'):
+                    # invalidate, then delete
+                    new_path = os.path.join(cache_dir,'.~~delete~~' + entry)
```
wenottingham, line 2320: I initially added this out of an idea to do these deletes in the background outside of the lock.

I could never get that to work, though. (Just commenting for posterity)

awx/main/tasks.py
```diff
@@ -1865,44 +1865,31 @@ def pre_run_hook(self, job, private_data_dir):
         project_path = job.project.get_project_path(check_if_exists=False)
         job_revision = job.project.scm_revision
         sync_needs = []
-        all_sync_needs = ['update_{}'.format(job.project.scm_type), 'install_roles', 'install_collections']
+        source_update_tag = 'update_{}'.format(job.project.scm_type)
+        branch_override = bool(job.scm_branch and job.scm_branch != job.project.scm_branch)
```
ryanpetrello, line 1869: Why not just use `job.branch_override` here?

ryanpetrello: Oh, ignore me, this is a `Job`, not a `ProjectUpdate`.

awx/main/models/projects.py
AlanCoding: This is a mess. The only reason it's used is because this is the status-quo:

https://github.com/ansible/awx/blob/8b97e3f8bb168f77045ffb2f15cfffc8e77fa005/awx/main/models/projects.py#L191

If `local_path` is ever `None`, it will error in almost any case. We don't handle the null case. What would it even mean if you did get that? That would happen if `local_path` is `foo/`, but we auto-generate this field, and I can't figure out anyway the auto-generation would produce that. _Maybe_ manual projects could accept this kind of value. If it did, this wouldn't be right. You would prefer to `rstrip` the slash instead.

It looks like the code is, for some reason, expecting to be given an absolute path. I don't know why.

awx/main/models/projects.py
AlanCoding: You're looking at the `Project` model definition. It is also defined for `ProjectUpdate` down below. If it's a normal sync, then it return's the project's last job (for core functionality). If it's a branch override case, we agreed to give up on cache invalidation, so it uses its own id, simply to make sure that it is never mistaken for the default branch's cache by other syncs.

matburt (approved):

awx/main/models/projects.py
```diff
@@ -455,11 +464,12 @@ def _get_related_jobs(self):
...
-        path_to_delete = self.get_project_path(check_if_exists=False)
+        paths_to_delete = (self.get_project_path(check_if_exists=False), self.get_cache_path())
         r = super(Project, self).delete(*args, **kwargs)
-        if self.scm_type and path_to_delete:  # non-manual, concrete path
-            from awx.main.tasks import delete_project_files
-            delete_project_files.delay(path_to_delete)
+        for path_to_delete in paths_to_delete:
+            if self.scm_type and path_to_delete:  # non-manual, concrete path
+                from awx.main.tasks import delete_project_files
+                delete_project_files.delay(path_to_delete)
```
chrismeyersfsu, line 472: This isn't something you need to address, just FYI. Running a task now goes over NOTIFY/LISTEN interface. `NOTIFY` has a maximum payload size of 8,000. Apply that to this case and if the path string is longer than 8,000 characters, the cleanup will not run.

Here is a pattern when the payload is large.

https://github.com/ansible/awx/blob/d8f6ea3fe89eaa6aea375dfa0280b43c66899ae0/awx/api/generics.py#L987
https://github.com/ansible/awx/blob/d8f6ea3fe89eaa6aea375dfa0280b43c66899ae0/awx/main/tasks.py#L2938

chrismeyersfsu (approved):

awx/main/models/projects.py
AlanCoding: huh, I would think that could be a problem for the `handle_setting_changes` task, but it looks like the 8,000 limit may be specifically selected to be more than the character count for a listing of all setting keys, which comes to 6,845.

The project folder _could_ be anywhere if the user configured it in CTiT, but that's not a use case that concerns me a lot. The project name is in both of these paths, but that's capped at 512. It actually helps that we issue the task twice, instead of putting more data in the args.

Issue #7473, opened 2020-06-29
Smart Inventory Edit Form


marshmalien:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
More details in smart inventory add form issue: #7472 

Fields:
1. Name 
    * Required
2. Description 
3. Organization 
    * Required
    * Organizations populate from `/api/v2/organizations/?page_size=5&order_by=name&role_level=inventory_admin_role`
4. Smart host filter
    * Required
    * In the old UI we disable this field until Organization has a value 
    * Hosts populate from ` /api/v2/hosts/?order_by=name&page_size=5&inventory__organization=:org-id`
    * Mockup:  https://tower-mockups.testing.ansible.com/patternfly/inventories/inventories-smart-edit-filter/
5. Instance groups
    * Instance groups populate from `/api/v2/instance_groups/?order_by=name&page_size=5`
6. Variables

Action Buttons:
1. Cancel
    * Navigate to inventories list
2. Save 
    * Navigate to smart inventory detail

Link to mockup: https://tower-mockups.testing.ansible.com/patternfly/inventories/inventories-smart-edit/

dsesami:
closed with the merge of #7644

Issue #7472, opened 2020-06-29
Smart Inventory Add Form


marshmalien:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Fields:
1. Name 
    * Required
2. Description 
3. Organization 
    * Required
    * Organizations populate from `/api/v2/organizations/?page_size=5&order_by=name&role_level=inventory_admin_role`
4. Smart host filter
    * Required
    * In the old UI we disable this field until Organization has a value 
    * Hosts populate from ` /api/v2/hosts/?order_by=name&page_size=5&inventory__organization=:org-id`
    * Lookup UX will need to consider how to populate filter keys: https://github.com/ansible/awx/issues/5513
    * The host filter supports relational queries, `and` `or` boolean logic, and expression grouping via `()`. I'm not sure what the UX will look like, but this advanced filter work will probably follow the same pattern as advanced search. 
        * Example: `(name=my_host and groups__name="my group") or (name=my_host2 and groups__name=my_group2)`
5. Instance groups
    * Instance groups populate from `/api/v2/instance_groups/?order_by=name&page_size=5`
6. Variables

Action Buttons:
1. Cancel
    * Navigate to Inventories list
2. Save 
    * Navigate to Smart inventory detail on successful save

Link to mockup: https://tower-mockups.testing.ansible.com/patternfly/inventories/inventories-smart-add/

dsesami:
closed with the merge of #7644

Issue #5513, opened 2019-12-16
Determine the keys to use for filtering in `host_filter`


wenottingham:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

As we move to the new search flow in ui_next where there is a starting set of keys for easier filtering, we should determine how we would want to do this in the host_filter for smart inventory, and if we would want to populate keys based on specific ansible fact names (e.g., `ansible_system`, `ansible_distribution_version`, `ansible_os_family`, etc.)

unlikelyzero:
covered in https://github.com/ansible/awx/pull/7644

unlikelyzero:
Re-opening to re-evalutate what was tested in  #7644

unlikelyzero:
@tiagodread to test this area and update the associated tests

tiagodread:
Trying to select the key `Ansib`:
 
![chrome-capture (17)](https://user-images.githubusercontent.com/12464858/119730815-8d290600-be4c-11eb-97be-9559f5c7f358.gif)

nixocio:
May be related: https://github.com/ansible/awx/issues/11985

Issue #3427, opened 2019-03-14
Include Host_filter OR Logic in Smart Inventory UI


jamesmarshall24:
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME

 - UI

##### SUMMARY

[Smart inventory](https://docs.ansible.com/ansible-tower/latest/html/userguide/inventories.html#smart-inventories) host_filter is currently set AND. Add ability to allow OR set for host_filter.

mrmibrown:
Here's a work-around for this issue:
https://access.redhat.com/solutions/4268032

shanemcd:
We are moving away from smart inventories and recently introduced a new feature called Constructed Inventory. That merged in https://github.com/ansible/awx/pull/13448 and got released with AWX 22. Constructed Inventories allow for querying host vars. I'm going to go ahead and close this and can also follow up with a link to a blog post that @AlanCoding is currently working on.

Docs to new constructed inventory source:

https://github.com/ansible/awx/blob/devel/docs/inventory/constructed_inventory.md

PR #7644, opened 2020-07-17
Add smart inventory add/edit forms and host filter lookup

marshmalien:
##### SUMMARY
Issues: #7473 #7472

This PR adds the smart inventory add and edit forms, as well as the host filter lookup. A Smart Inventory is a collection of hosts defined by the host filter search. The host filter only applies to hosts of inventories that belong to the Smart Inventory's organization, therefore, the host filter lookup will remain disabled until the organization field is given a value.  If a user does not have permission to POST a smart inventory, the form save button will remain disabled. 

The smart host filter lookup works a little differently from our other lookups. Instead of selecting hosts from a list, you create a search tag that contains the hosts you want. I created a separate HostFilterUtils file with helper methods to handle converting the query string and host filter string to a searchParams object and back. The host filter lookup doesn't handle advanced search features such as conditional `or`, compound expressions `()`, or relational queries using `host_filter`. The lookup exposes the following filter keys: name, id, groups__name, inventory, instance_id, enabled, last_job, and insights_system_id. 

This issue #5513 mentions populating keys based on specific ansible fact names, but I'm not sure it is possible to search on `host_filter` with our current basic search toolbar. I may be wrong about that assumption, but I wasn't successful at finding a way to prepend `host_filter` to a namespaced query string
`?smart_hosts.host_filter=ansible_facts__ansible_processor_vcpus=8` without breaking the search chip groups. My goal was to get as far as possible without editing the Search component. 

**_Smart inventory form_**
Screen Shot 2020-07-17 at 3 21 48 PM

**_Host filter modal_**
Screen Shot 2020-07-17 at 4 10 05 PM

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - UI

### DIFF
--- a/awx/ui_next/src/components/Lookup/HostFilterLookup.jsx
+++ b/awx/ui_next/src/components/Lookup/HostFilterLookup.jsx
@@ -1,0 +1,345 @@
+import React, { useState, useEffect, useCallback } from 'react';
+import { withRouter, useHistory, useLocation } from 'react-router-dom';
+import { number, func, bool, string } from 'prop-types';
+import { withI18n } from '@lingui/react';
+import styled from 'styled-components';
+import { t } from '@lingui/macro';
+import { SearchIcon } from '@patternfly/react-icons';
+import {
+  Button,
+  ButtonVariant,
+  Chip,
+  FormGroup,
+  InputGroup,
+  Modal,
+} from '@patternfly/react-core';
+import ChipGroup from '../ChipGroup';
+import DataListToolbar from '../DataListToolbar';
+import LookupErrorMessage from './shared/LookupErrorMessage';
+import PaginatedDataList, { PaginatedDataListItem } from '../PaginatedDataList';
+import { HostsAPI } from '../../api';
+import { getQSConfig, mergeParams, parseQueryString } from '../../util/qs';
+import useRequest, { useDismissableError } from '../../util/useRequest';
+import {
+  removeDefaultParams,
+  removeNamespacedKeys,
+  toHostFilter,
+  toQueryString,
+  toSearchParams,
+} from './shared/HostFilterUtils';
+
+const ChipHolder = styled.div`
+  --pf-c-form-control--Height: auto;
+  .pf-c-chip-group {
+    margin-right: 8px;
+  }
+`;
+
+const ModalList = styled.div`
+  .pf-c-toolbar__content {
+    padding: 0 !important;
+  }
+`;
+
+const useModal = () => {
+  const [isModalOpen, setIsModalOpen] = useState(false);
+
+  function toggleModal() {
+    setIsModalOpen(!isModalOpen);
+  }
+
+  function closeModal() {
+    setIsModalOpen(false);
+  }
+
+  return {
+    isModalOpen,
+    toggleModal,
+    closeModal,
+  };
+};
+
+const QS_CONFIG = getQSConfig(
+  'smart_hosts',
+  {
+    page: 1,
+    page_size: 5,
+    order_by: 'name',
+  },
+  ['id', 'page', 'page_size', 'inventory']
+);
+
+const buildSearchColumns = i18n => [
+  {
+    name: i18n._(t`Name`),
+    key: 'name',
+    isDefault: true,
+  },
+  {
+    name: i18n._(t`ID`),
+    key: 'id',
+  },
+  {
+    name: i18n._(t`Group`),
+    key: 'groups__name',
+  },
+  {
+    name: i18n._(t`Inventory`),
+    key: 'inventory',
+  },
+  {
+    name: i18n._(t`Enabled`),
+    key: 'enabled',
+  },
+  {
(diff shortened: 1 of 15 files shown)


mabashian:
It looks like something's getting a little sideways when searching for a string with a space in it.  It looks like the space is getting encoded and then displayed that way.  When I open up the host filter modal my matching hosts disappear:

test_smart_inv

mabashian:
You may have already talked with Bill/Taufique about this but is the plan for this field to just show the raw string:

Screen Shot 2020-07-22 at 9 15 17 AM

or do we need to somehow convert that in to "search tags" to match:

Screen Shot 2020-07-22 at 9 16 42 AM

mabashian:
I don't think we need this prompt on launch - maybe just some copy/paste?

Screen Shot 2020-07-22 at 9 26 18 AM

mabashian:
Screen Shot 2020-07-22 at 9 29 39 AM

👍 on making this disabled until the user selects an Org.  You may have already discussed this but should we add a tooltip to this field indicating that an Org needs to be selected?  We could also hide the field while Org is empty but meh, idk if I like that better or not.

I'm fine leaving it the way it is, just figured I'd ask.

mabashian:
I'm a little bit concerned about including this field in the default search values:

Screen Shot 2020-07-22 at 9 37 10 AM

at first I thought that it was the key: `ansible_facts`.  Then I went and looked at the OPTIONS and realized it's `The date and time ansible_facts was last modified.`.  I'm not sure that that's super useful when creating a host filter and may just be confusing to folks that are trying to build a query using `ansible_facts`.

mabashian:
It looks like these come straight from the OPTIONS response:

Screen Shot 2020-07-22 at 9 44 39 AM

but I'm wondering if we should update Inventory -> Inventory ID since it seems like that field requires a number.

Also wondering if we have a pattern for boolean fields like `Enabled`.  This field only really has two options `true` and `false`.  Rather than forcing them to type it in could we just give them a dropdown with those two options?

This may be bleeding into a larger discussion about what fields we want to make available on the "basic" search level and what we want to make available in "advanced".

mabashian:
Outside the scope of this particular PR (though related) is https://github.com/ansible/awx/issues/3427 i.e. UI support for `OR` logic in the host_filter.  This is already supported by the API.  The most basic solution to this problem would be to expose a simple text input that a user could opt in to when creating a smart inventory host filter where a complete `host_filter` string could be supplied.  This would serve 2 purposes:

1) Users wanting to specify complex/non-ui supported host filter strings could do so in the UI and wouldn't have to go through the API browser
2) If the UI detects a string that we can't parse into consumable search tags then we just show the string in the input.

I wouldn't expect any of this work to be included in this particular PR but it may impact UX decisions so I felt like it was worth bringing up.

@wenottingham thoughts/opinions?

mabashian:
This may be related to https://github.com/ansible/awx/pull/7644#issuecomment-662440983 but a filter with multiple `name` tags seems to be getting wonky as well:

test_smart_inv_2

If I go in and check for matching hosts after saving there are 0 instead of 1 :(

marshmalien:
> You may have already talked with Bill/Taufique about this but is the plan for this field to just show the raw string:
> 
> Screen Shot 2020-07-22 at 9 15 17 AM
> 
> or do we need to somehow convert that in to "search tags" to match:
> 
> Screen Shot 2020-07-22 at 9 16 42 AM

I think the plan is to display this using PatternFly LabelGroups and [Labels](https://www.patternfly.org/v4/documentation/react/components/label) similar to how we display it using ChipGroups and Chips. The PatternFly LabelGroup component is still in the design phase.

wenottingham:
@mabashian having a text input could work. How would we expose it? (I hate an 'advanced' checkbox, but not sure what else it would be.)

mabashian:
@wenottingham yea I think this would be exposed in the advanced section.  @jlmitch5 is working on that portion separately as far as I understand so we can talk more about it on his side of things.  I do think there's a place for something like that in the UI but we'll have to make sure it's a last resort and not something we expect users to use frequently.

mabashian (approved):

dsesami:
spoke with @marshmalien about this. some notes:

1) It might be good to have X's to delete chips with the search filter. Marly has ideas for later improvements to this area, but those don't apply quite yet. 
2) the search filter modal on the add/edit form does not show any info besides names. This might trip up users who want to filter by (for example) group, but can't immediately tell what group the current list of hosts are in.

marshmalien:
Will address adding a host filter string text input in this issue: https://github.com/ansible/awx/issues/7850
and expanding host list content in this issue: https://github.com/ansible/awx/issues/7853

dsesami:
spoke with @marshmalien. We'll probably get into the ideas for expanded views in a future issue or as possible parts of #7850

dsesami (approved):

nixocio:
Related: https://github.com/ansible/awx/issues/11017

Issue #4736, opened 2019-09-13
docker-compose install gets "no pg_hba.conf entry"


javexed:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Standard docker install fails to work.

##### ENVIRONMENT
* AWX version: latest (60fc952)
* AWX install method: docker on linux
* Ansible version:  2.8.4
* Operating System: CentOS 3.10.0-957.el7.x86_64
* Web Browser: n/a

##### STEPS TO REPRODUCE

Standard git pull, from awx/installer, run sudo ansible-playbook -i inventory install.yml

##### EXPECTED RESULTS

A working AWX install

##### ACTUAL RESULTS

From the docker logs of the postgres image, I see lots of these:
no pg_hba.conf entry for host

The location as per the docker-compose is:
      PGDATA: /var/lib/postgresql/data/pgdata
with the volume mapped as:
      - /tmp/pgdocker/10/data/:/var/lib/postgresql/data/pgdata:Z
Running inside the container:
cat $PGDATA/pg_hba.conf works

So postgres is looking for another location? Will keep running tests to see if I can figure out where it is trying to pull from.

ryanpetrello:
@shanemcd @rooftopcellist is this related to the recent postgres10 update?

javexed:
Ok, got the fix, I think. 
Based the solution from here: https://github.com/docker-library/postgres/issues/482
Added the following line to pg_hba.conf
```bash
host all all 0.0.0.0/0 trust
```
NOTE: no idea if this is kosher.
Then did a 
```bash
docker-compose restart
```

javexed:
Ok, reading from here: https://github.com/kartoza/docker-postgis/issues/83

It looks like we should be able to fix the issue by adding the following line to 
https://github.com/ansible/awx/blob/devel/installer/roles/local_docker/templates/docker-compose.yml.j2
in the postgres section
```bin
ALLOW_IP_RANGE: 0.0.0.0/0
```

However, doing that doesn't seem to fix the issue. Wondering if it is that issue mentioned in that ticket.
Thus, the band-aid solution is to do the above, namely manually edit the pg_hba.conf file and then restart.

treyd:
Has anyone successfully worked around the `pg_hba.conf` issue and gotten AWX to deploy on a local docker?  I've tried but now I'm getting messages about the `awx` and `postgres` databases not existing.

rooftopcellist:
[...]

PR #7698, opened 2020-07-21
Use default path for pgdata volume in docker-compose template

geerlingguy:
##### SUMMARY

Use the default pgdata volume path for the `postgres:10` docker container in the templated `docker-compose.yml.j2` file.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - Installer

##### AWX VERSION

```
make: Nothing to be done for `version'. (devel)
```

##### ADDITIONAL INFORMATION

See https://github.com/ansible/awx/issues/4736#issuecomment-662067865 — the Docker library `postgres:10` image that is used in the docker-compose configuration recommends using the default volume path of `/var/lib/postgresql/data` for a mounted data directory.

Currently AWX sets this path to `/var/lib/postgresql/data/pgdata` and overrides the `PGDATA` environment variable, but that doesn't seem like it's necessary (unless there's something I'm missing).

### DIFF
--- a/installer/roles/local_docker/templates/docker-compose.yml.j2
+++ b/installer/roles/local_docker/templates/docker-compose.yml.j2
@@ -167,12 +167,11 @@ services:
     container_name: awx_postgres
     restart: unless-stopped
     volumes:
-      - {{ postgres_data_dir }}/10/data/:/var/lib/postgresql/data/pgdata:Z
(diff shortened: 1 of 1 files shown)


rooftopcellist (approved):
I agree, using the default is better `/var/lib/postgresql/data`.  I don't remember there being any specific reason why we have it the way it is.  I checked out these changes and was able to run awx without issue.

ryanpetrello:
regate

Issue #7699, opened 2020-07-21
Can't install project using docker-compose on Mac from devel or 13.0.0


geerlingguy:
##### ISSUE TYPE

 - Bug Report

##### SUMMARY

I've tried cloning the project and following the [Docker Compose installation instructions](https://github.com/ansible/awx/blob/devel/INSTALL.md#docker-compose) for both `devel` and `13.0.0`, and neither version results in a working installation on my Mac.

##### ENVIRONMENT
* AWX version: `13.0.0` or `devel`
* AWX install method: docker for mac
* Ansible version: 2.9.10
* Operating System: macOS (10.15.5)
* Web Browser: Safari

##### STEPS TO REPRODUCE

  1. `git clone https://github.com/ansible/awx.git`
  2. `cd awx/installer`
  3. `ansible-playbook -i inventory install.yml`

##### EXPECTED RESULTS

After some time for database migration to complete, the AWX installation should be accessible at http://localhost/

##### ACTUAL RESULTS

  1. The page http://localhost/ remains stuck at 'Server Error' indefinitely (waited 20 minutes).
  1. The awx_task and awx_web container logs show:

     ```
     django.db.utils.OperationalError: FATAL:  no pg_hba.conf entry for host "172.24.0.5", user "awx", database "awx", SSL off
     ```

  1. The awx_posgres container logs show:

     ```
     2020-07-21 20:43:10.990 UTC [1466] FATAL:  no pg_hba.conf entry for host "172.24.0.5", user "awx", database "awx", SSL off
     ```

##### ADDITIONAL INFORMATION

Related issue: https://github.com/ansible/awx/issues/4736

Full initial startup log of the postgres container (using `docker logs awx_postgres`):

```
The files belonging to this database system will be owned by user "postgres".
This user must also own the server process.

The database cluster will be initialized with locale "en_US.utf8".
The default database encoding has accordingly been set to "UTF8".
The default text search configuration will be set to "english".

Data page checksums are disabled.

fixing permissions on existing directory /var/lib/postgresql/data/pgdata ... ok
creating subdirectories ... ok
selecting default max_connections ... 100
selecting default shared_buffers ... 128MB
selecting default timezone ... Etc/UTC
selecting dynamic shared memory implementation ... posix
creating configuration files ... ok
running bootstrap script ... ok
performing post-bootstrap initialization ... ok
syncing data to disk ... 
PostgreSQL Database directory appears to contain a database; Skipping initialization

[...]

PR #7701, opened 2020-07-21
Use a docker volume instead of bind mount for postgres data in docker-compose

geerlingguy:
##### SUMMARY

For background, see: https://github.com/ansible/awx/issues/7699

Currently, on Docker for Mac, if you try installing with the Docker Compose configuration, the first time the postgres container starts, it doesn't initialize a database correctly, due to some issue with the default bind mount.

It would make for a much faster experience, and also solve this bug, to use a Docker volume instead of a bind mount by default for the postgres container. That's what this PR does.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - Installer

##### AWX VERSION

```
make: Nothing to be done for `version'. (devel)
```

##### ADDITIONAL INFORMATION

Alternatively, the configuration for the mount could be templated, so you could choose a bind mount or a docker volume.

Also, I'm not sure what the rationale is for a bind mount (vs a Docker volume), but if the main concern is backup, it's easy enough to [back up the contents of a Docker volume](https://docs.docker.com/storage/volumes/#backup-restore-or-migrate-data-volumes).

### DIFF
--- a/installer/roles/local_docker/templates/docker-compose.yml.j2
+++ b/installer/roles/local_docker/templates/docker-compose.yml.j2
@@ -167,7 +167,7 @@ services:
     container_name: awx_postgres
     restart: unless-stopped
     volumes:
-      - {{ postgres_data_dir }}/10/data/:/var/lib/postgresql/data:Z
+      - postgres-data:/var/lib/postgresql/data
     environment:
       POSTGRES_USER: {{ pg_username }}
       POSTGRES_PASSWORD: {{ pg_password }}
@@ -195,3 +195,4 @@ volumes:
   supervisor-socket:
   rsyslog-socket:
   rsyslog-config:
+  postgres-data:


shanemcd:
At the very least we'll need a changelog entry to go along with this. It'd also be nice to describe or link to a migration path. We'll also need to bump the major version before putting this out in a release.

geerlingguy:
Here are the commands to migrate from a bind mount to a docker volume:

1. Tear down the current running instance (e.g. `docker-compose down -v`).
2. _Don't_ delete the current data directory (in `{{ postgres_data_dir }}/10/data`), because that's what we're going to copy to the new instance.
3. Update to the newer instance using the installer after this PR is merged and there's a new major version.
4. Once the environment is built, turn it off (even if it's still initializing a new database):

   ```
   cd ~/.awx/awxcompose
   docker-compose stop
   ```

5. Create a container with the postgres volume mounted to copy the files from your local path into the postgres-data Docker volume:

   ```
   docker container create --name temp -v awxcompose_postgres-data:/data busybox
   cd ~/.awx/pgdocker/10/data
   docker cp . temp:/data
   docker rm temp
   ```

6. Re-start the new AWX environment:

   ```
   cd ~/.awx/awxcompose
   docker-compose start
   ```

7. Once you can confirm everything's working correctly, delete the old local data directory:

   ```
   rm -rf ~/.awx/pgdocker
   ```

I just tested this process and at the end, the installation seems to be working normally with the new volume layout: I installed using the original bind mount, logged in and ran one job, then did everything listed above, and now with the new volume layout, it's all working correctly, and I could run another job:

Screen Shot 2020-07-21 at 8 48 11 PM

geerlingguy:
@shanemcd - Regarding the changelog entry... is this something that you'd want me to add in a new section in the existing changelog, or is there some other process for adding changelog fragments? I didn't see anything other than the main 'CHANGELOG.md' file in this repo.

ryanpetrello:
I'm also fine with this, but I'd want to put together some instructions similar to @geerlingguy's comment, and include them in the release announcement so that people using local Docker know how to migrate their data.

shanemcd:
Looks like there are now some conflicts.

Regarding the upgrade instructions, maybe it's ok to just link to the comment above.

ryanpetrello:
@geerlingguy I'm actually a little concerned that this change is going to result in a lot of flak from the community on upgrade (because people won't read, and they'll be confused and surprised when suddenly all their data is "gone").

I'll admit that "runs on Docker for Mac" isn't a path we focus much of our effort on - do you know if there's any other alternative here to resolve the referenced bug that doesn't involve a breaking change for every AWX user on upgrade?  If not, I think the cost might be too high to make an argument for this change.

geerlingguy:
@ryanpetrello - The main motivation is to make it so AWX can be installed on Macs using Docker for Mac without having to dig through the issue queues and figure out that you have to start the environment, then delete the postgres data folder on the Mac, then restart the `postgres` and `task` containers, otherwise you just get 'Server Error' and your CPU goes crazy as Docker logs fill up with `pg_hba.conf` errors.

Considering it took me about 3 hours to figure out how to get AWX started on my Mac yesterday without ever having installed it on this particular Mac, I'd argue it might be worth it for the first-time user experience.

The problem of users not having a proper upgrade path / release notes is one that's been kind of par for the course running AWX, so any notes (like instructions for converting the mount) would be an improvement over 'here's a link to the changelog between 12 and 13, good luck with the upgrade' (IMHO).

ryanpetrello:
> The problem of users not having a proper upgrade path / release notes is one that’s been kind of par for the course running AWX

...and it's been something we've gotten roasted alive for, and that we're trying to better about 😄.  Recent releases over the past year of AWX have been much better about this, and we *do* publish release notes/a changelog now.

I don't have metrics to confirm it, but I suspect that the number of people running AWX on Docker for Mac and encountering this issue is _far lower_ than the number of setups we'll unexpectedly break on upgrade if we make a change like this.  Instructions or not, it _will_ be perceived as a breaking change, and it'll be a surprising nuisance to many.

shanemcd:
> The problem of users not having a proper upgrade path / release notes is one that's been kind of par for the course running AWX

This hasnt been true for a long time.

Are we sure switching to a volume is the only solution here?

geerlingguy:
@shanemcd - I know that things have gotten better, I'm just saying historically it's been a problem and having good enough release notes that people are more confident in the upgrade process would help people see it's gotten better. I realize my earlier comment reads a bit harsh, that was not intended :)

> Are we sure switching to a volume is the only solution here?

I tried setting up the mount with no `:Z`, with `:cached`, and no other way of mounting it seemed to make it work. Besides, using a docker volume makes performance a _lot_ better on Mac since the volume is inside the virtualized Docker environment and not shared across the macOS filesystem (which is 20-100x slower)—plus it's generally preferred to use a Docker volume unless you need frequent access to the files on the volume from your host OS (e.g. if you're developing a web app and need to modify files in the mount via an IDE).

I can't think of a use case where it's important to have the postgres data files conveniently accessible on the host machine where running AWX, except for easier backup (since you could just drag and drop the folder somewhere to back it up, instead of having to run a script with a few `docker` commands).

shanemcd:
I think the only way we are going to merge this is if the installer itself attempted to do the right thing. You can see one example of another time we did the heavy lifting for users here: https://github.com/ansible/awx/blob/devel/installer/roles/local_docker/tasks/upgrade_postgres.yml

We'd want to call the steps described in your comment above sometime after ^ runs. Then in the changelog we can just say "hey, your data now lives somewhere else".

ryanpetrello:
@geerlinguy sorry for the delay on reviewing/merging this PR - we've since replaced the local Docker installer in `devel` (https://groups.google.com/g/awx-project/c/47MjWSUQaOc/m/bCjSDn0eBQAJ https://github.com/ansible/awx/tree/devel/tools/docker-compose) and thus the next major release of AWX, so I suspect this PR is obsolete now.

PR #7705, opened 2020-07-22
Change Dockerfile to copy custom venv

anxstj:
##### SUMMARY

The container build process has changed with AWX 12.0.0. The new process copies the default virtual environment from a build container to the final awx container. Unfortunately, the custom virtual environments, which are located in /opt, get lost with this.

This PR updates the documentation to highlight that the custom virtual environment path must be copied to the final awx container, too.

Another option would be to change the Dockerfile to always copy the /opt directory, regardless of custom virtual environments or not. But I'm not sure if /opt is really the "supported" place for custom virtual environments.

##### ISSUE TYPE

 - Docs Pull Request

##### COMPONENT NAME

 - Installer

##### AWX VERSION

```
awx: 13.0.0

```


ryanpetrello (approved):
Thanks!

Issue #7658, opened 2020-07-20
Analytics fail to upload if the size is greater than 100MB


benthomasson:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
A customer's analytics uploads were failing since the bundle size was greater than 100MB.  This is due to an upload size limit in the upload service.    Tower should upload bundles smaller than 100MB by building a bundle and if it is too big then divide the time period in half and send two bundles instead. Recurse until small enough bundles are reached.    

##### ENVIRONMENT
* AWX version: X.Y.Z
* AWX install method: openshift, minishift, docker on linux, docker for mac, boot2docker
* Ansible version:  X.Y.Z
* Operating System:
* Web Browser:

wenottingham:
The simplest way to do this is probably split how we gather the data and do uploads.
- generate full upload minus job events
- check the last submission date
- for every day since prior submission, create a daily dump of job events, and send each separately

benthomasson:
That will probably be less CPU and memory intensive on the query side.   Paginating the events and sending a constant number in a larger number of bundles would be better for the sender and the receiver.

jctanner:
functional example of how to split at ship time https://gist.github.com/jctanner/db5a4d5e6a054ac0468ef09ed42ea276

Tested with local api and did not see any errors, but also didn't validate the data was processed correctly.

```
fastapi_1           | {"@timestamp": "2020-07-22T19:34:22.438Z", "@version": 1, "source_host": "4c2715477e42", "name": "tower_analytics_report.processor.tower_analytics_processor", "args": [], "levelname": "INFO", "levelno": 20, "pathname": "./tower_analytics_report/processor/tower_analytics_processor.py", "filename": "tower_analytics_processor.py", "module": "tower_analytics_processor", "stack_info": null, "lineno": 840, "funcName": "handle_events_table", "created": 1595446462.438589, "msecs": 438.58909606933594, "relativeCreated": 8097716.229200363, "thread": 140240822705480, "threadName": "MainThread", "processName": "SpawnProcess-1", "process": 32, "message": "handle_events_table.merge_encrypted"}
[...]

PR #5721, opened 2020-01-21
Change how analytics is gathered to only gather once per interval

wenottingham:
##### SUMMARY

Move to just a periodic beat task, but store last collection time so we only gather on the specified (configurable) interval.

Use an advisory lock so multiple nodes don't try and send at once.

##### ISSUE TYPE
 - Feature Pull Request

##### ADDITIONAL INFORMATION

This is probably abuse of the settings infrastructure to store the last collection time there.

PR #7709, opened 2020-07-22
Adjust analytics gathering

wenottingham:
##### SUMMARY

Two major changes:

1. When collecting, don't collect everything in one big bundle. Split out jobs/workflow jobs/job events into their own collection, and collect a series of 20 minute bundles for those, sending each separately. 
2. ~Allow configurable subsets of data collection, via AUTOMATION_ANALYTICS_COLLECTOR_SUBSET.  Collector names to put in this list can be pulled from /api/v2/config. (No UI support at this time.)~ Removed the config setting until we're sure that we won't delete this mechanism as we fix it more.
3. Spit table dumps at ~200MB to ensure we don't hit a size limit

##### ISSUE TYPE
 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### POTENTIAL PITFALLS

This does not check that it does not generate a too large bundle, but instead relies on "never ship a CSV over 200MB", which should accomplish the same task.

When doing a 4 week collection, this can produce ~150 bundles over a period of minutes.


wenottingham:
cc @jctanner as well

For https://github.com/ansible/awx/issues/7658

wenottingham:
~This behavior change is *only* for analytics gathered from Tower tasks. Running `awx-manage gather_analytics` still gets you one big firehosed bundle.~ This has changed, you get a number of bundles regardless.

awx/main/tasks.py
```diff
@@ -354,6 +355,21 @@ def send_notifications(notification_list, job_id=None):
...
+    def _gather_and_ship(subset, since=None, until=None):
+        error = 0
+        try:
+            tgz = analytics.gather(subset=subset, since=since, until=until)
+            if not tgz:
+                return error
+            logger.info('gathered analytics: {}'.format(tgz))
+            analytics.ship(tgz)
+        except Exception:
+            error += 1
```
Ladas: let log the exception, so it doesn't get lost:

```
except Exception as e:
    logging.error(e, exc_info=True)
```

awx/main/analytics/collectors.py
```diff
@@ -290,12 +298,16 @@ def _copy_table(table, query, path):
                                  FROM main_unifiedjob
                                  JOIN django_content_type ON main_unifiedjob.polymorphic_ctype_id = django_content_type.id
                                  LEFT JOIN main_organization ON main_organization.id = main_unifiedjob.organization_id
-                                 WHERE (main_unifiedjob.created > {0} OR main_unifiedjob.finished > {0})
+                                 WHERE ((main_unifiedjob.created > {0} AND main_unifiedjob.created < {1})
```
Ladas: we should probably do `(main_unifiedjob.created > {0} AND main_unifiedjob.created <= {1})` the the `<=` less or equal

Although the probability we'll hit this in prod is near 0.

We might hit this in a perf test env, where we'll probably have mock up data aligned to full seconds, not microseconds

awx/main/tasks.py
wenottingham: Updated with logger.exception

awx/main/analytics/collectors.py
wenottingham: Done

awx/main/analytics/collectors.py
```diff
@@ -217,30 +218,34 @@ def job_instance_counts(since):
...
-            cursor.copy_expert(query, file)
-            file.close()
-        return file_path
+def _copy_table(table, query, path):
+    file_path = os.path.join(path, table + '_table.csv')
+    file = open(file_path, 'w', encoding='utf-8')
+    with connection.cursor() as cursor:
+        cursor.copy_expert(query, file)
+        file.close()
+    # Ensure we actually dumped data, and not just headers
```
wenottingham: Spent 20 minutes staring at psycopg to see if there's a better way to do this. I'd really like to find one.

awx/main/tasks.py
```diff
@@ -372,16 +389,30 @@ def gather_analytics():
...
+            if subset:
+                errors += _gather_and_ship(subset, since=since, until=gather_time)
+            for collector in incremental_collectors:
+                start = since
+                until = None
+                while start < gather_time:
+                    until = start + timedelta(minutes=20)
+                    if (until > gather_time):
+                        until = gather_time
+                    errors += _gather_and_ship(collector, since=start, until=until)
```
Ladas: @wenottingham just to make sure, do we always send `config.json` with each payload? I got lost in debugging of the path :-)

config.json is required, since all data saved is tied to cluster_id we get from this

wenottingham: No, it doesn't guarantee that at the moment, will need fixed.

Ladas: Maybe `manifest.json` too, we don't use the versions anywhere now, but each payload should probably have it just for sure, what do you think @benthomasson ? (processor treats it as optional though)

wenottingham: manifest is always added, config.json would go in the same block

awx/main/tasks.py
```diff
@@ -354,6 +355,22 @@ def send_notifications(notification_list, job_id=None):
...
 @task(queue=get_local_queuename)
 def gather_analytics():
+    def _gather_and_ship(subset, since=None, until=None):
+        try:
+            tgz = analytics.gather(subset=subset, since=since, until=until)
+            # empty analytics without raising an exception is not an error
+            if not tgz:
+                return True
+            logger.info('gathered analytics: {}'.format(tgz))
+            analytics.ship(tgz)
```
Ladas: hmm, so one new issue and one old issue

new issue
-------------

https://github.com/ansible/awx/blob/devel/awx/main/analytics/core.py#L186-L189 will save the last updated timestamp for each collection, so if this fails in the middle, we'll miss a lot of data

So we should save the state.last_run after we've sent all the bundles, right?

And old issue
-------------------

It gets fresh `now()` before shipping, so this creates a gap now? https://github.com/ansible/awx/blob/devel/awx/main/analytics/core.py#L186-L189

E.g. I'll be collecting events where the `now()` is `2020-07-01T12:00:00`, but this will take a while and when I want to ship this I take another `now()` at. e.g. `2020-07-01T13:00:00` and say this is the last run

That will create 1h of gap. A data we'll never send to analytics. Should I create separate issue for this?

-----

Basically, we should take `now()` snapshot at the start and work with it for all collections and set it as `state.last_run` after all payloads were sent. That should fix both these bugs. (so setting up ahead what time window we're processing and mark the whole window as processed)

wenottingham: The solo model there shouldn't actually used now that this PR adds the `since` and `until` parameters to gather() and always passes them.
 
It probably should be removed for clarity.

awx/main/analytics/collectors.py
jameswnl: `select count(*)` should be a cheap call.  Looks like `created` field is indexed
(but `modified` looks like not indexed)

awx/main/tasks.py
```diff
@@ -372,16 +389,29 @@ def gather_analytics():
...
+                start = since
+                until = None
+                while start < gather_time:
+                    until = start + timedelta(minutes=20)
+                    if (until > gather_time):
+                        until = gather_time
+                    if not _gather_and_ship(incremental_collectors, since=start, until=until):
+                        break
+                    start = until
+                    settings.AUTOMATION_ANALYTICS_LAST_GATHER = until
```
Ladas, line 417: @wenottingham hm, I think we were using DB `TowerAnalyticsState` table, because the collection can be made by a different host each time?

settings.AUTOMATION_ANALYTICS_LAST_GATHER will store it only on 1 host? So I think we need to use the DB here.

awx/main/analytics/core.py
```diff
@@ -51,40 +74,26 @@ def projects_by_scm_type():
...
-    :pararm module: the module to search for registered analytic collector
+    :param module: the module to search for registered analytic collector
                     functions; defaults to awx.main.analytics.collectors
     """
 
-    run_now = now()
-    state = TowerAnalyticsState.get_solo()
-    last_run = state.last_run
-    logger.debug("Last analytics run was: {}".format(last_run))
+    last_run = since or settings.AUTOMATION_ANALYTICS_LAST_GATHER or (now() - timedelta(weeks=4))
```
Ladas, line 103: the `since` and `until` will be always set now correct? So we should just do

```
last_run = since
```

So we make sure the time window is set only on 1 place.

awx/main/analytics/core.py
```diff
@@ -51,40 +74,26 @@ def projects_by_scm_type():
...
-    manifest[file_name] = version
-
-    def decorate(f):
-        return f
-
-    return decorate
-
-
-def gather(dest=None, module=None, collection_type='scheduled'):
+def gather(dest=None, module=None, subset = None, since = None, until = now(), collection_type='scheduled'):
```
Ladas, line 85: maybe `until=None` ? We should always set it outside of this function, so all collections have the same until.

awx/main/tasks.py
```diff
@@ -372,16 +389,29 @@ def gather_analytics():
...
+            if subset:
+                _gather_and_ship(subset, since=since, until=gather_time)
```
Ladas, line 419: We should probably do `...last_gather = gather_time` here, so after everything is sent, to make sure we get all data in a case of failure somewhere in the middle.

(We'd need to store another timestamp, if we'd want to record incremental collectors and the rest of `subset`. So then in a case of failure, incremental collectors would continue from last batch, but the rest of the subset would be sent for the whole missing time.)

Also setting it only if there were no errors

awx/main/tasks.py
Ladas: hm I see, this is a specific settings implementation, not django settings.

So it seems like this will get saved to DB if the setting is part of `all_supported_settings` https://github.com/ansible/awx/blob/68c62d74f6dcbc1bfe49db5c209299f9f9905845/awx/conf/settings.py#L443 ?

Otherwise it gets stored only in memory.

Given we were using another DB table `TowerAnalyticsState`, I'd guess the AUTOMATION_ANALYTICS_LAST_GATHER change is stored only in memory :-)

Ladas: Hm, it looks like it is registered, so the change should be persisted to the DB https://github.com/ansible/awx/blob/68c62d74f6dcbc1bfe49db5c209299f9f9905845/awx/main/conf.py#L817

So the question is, why were we using another table to store this information? (it's either that it doesn't save properly to the DB or that the author didn't know about this :-) )

Ladas: I see, there was always
```
last_gather = Setting.objects.filter(key='AUTOMATION_ANALYTICS_LAST_GATHER').first()
```

But it was overridden later by 
```
state = TowerAnalyticsState.get_solo()	   
last_run = state.last_run
```

So that is quire strange :-o

Ladas: ok now I get it, so the original last run was the 
```
state = TowerAnalyticsState.get_solo()	   
last_run = state.last_run
```

Then we added the setting as part of fixing the bug where all systems were sending the duplicate payload 
https://github.com/ansible/awx/pull/5721/files

Ok, so this should work fine (we can mark this as resolved)

Funny comment btw. :-D 

```
ADDITIONAL INFORMATION
This is probably abuse of the settings infrastructure to store the last collection time there.
```

awx/main/analytics/core.py
Ladas: ideally we'd change the method to:
```
def gather(since, until, dest=None, module=None, subset = None, collection_type='scheduled')
```

Since the `since` and `until` must be always sent, as the fixed time window

awx/main/tasks.py
```diff
@@ -354,6 +355,22 @@ def send_notifications(notification_list, job_id=None):
 
 @task(queue=get_local_queuename)
 def gather_analytics():
+    def _gather_and_ship(subset, since=None, until=None):
```
Ladas: again, can be:

```
_gather_and_ship(subset, since, until):
```

since the time window must be always defined

awx/main/analytics/collectors.py
```diff
@@ -217,30 +218,34 @@ def job_instance_counts(since):
...
+    data = file.readline()
+    file.close()
+    if not data:
+        os.remove(file_path)
+        return None
+    return file_path
 
+
+@register('events_table', '1.1', format='csv', description=_('Automation task records'), expensive=True)
+def events_table(since, full_path, until = now(), **kwargs):
```
Ladas: also 
```
events_table(since, full_path, until, **kwargs)
```
and the same for below

awx/main/analytics/collectors.py
Ladas: well, reading 2 lines out of csv is probably much quicker than extra query

jameswnl: Reading csv is not necessarily quicker. We are comparing the cost of `[select * query + write-csv + read-csv]` vs `select count(*)` for an empty bundle.  The cost also depends on the probability of an empty 20-minutes window.  I guess the difference won't have much impact to the big picture.

However a `select count(*)` check is much cleaner/clearer in code readability IMHO.

Ladas: In current state maybe, but we want to slowly get to place where we load a bigger chunk from DB and divide the csv (if it goes over 100MB) rather than doing many small queries. That will be more effective on tower side and analytics processor. 

Then this workflow is fine :-)

awx/main/tasks.py
wenottingham: Yea, TowerAnalyticsState was the old thing that should have been removed when we first implemented the settings-based implementation of tracking the last run.

awx/main/tasks.py
wenottingham: So the way I intended to code this (I may have gotten it wrong) was to shift `settings.AUTOMATION_ANALYTICS_LAST_GATHER` to "the latest date we have job information for".

Hence we update it each time sending a time slice of the data succeeds, stop collecting that time sliced data entirely if one slice fails, and then send the non-automation-record data (counts, config,json, instance_info, etc) *without* updating the last gather date. 

This is a semantic change of what it represents, but as long as the data processor does the right thing, I think it's more accurate, and will ensure we (in theory) won't miss any job records.

awx/main/analytics/core.py
wenottingham: awx/main/mangement/commands/gather_analytics.py (manual gathering) doesn't set both timestamps when calling `gather()`. It could, and then we could change this.

awx/main/tasks.py
Ladas: Ah ok, only the `incremental_collectors` are time constrained, so for the rest of the subset, the `AUTOMATION_ANALYTICS_LAST_GATHER` doesn't matter. So then this is correct. You can mark this as resolved. :-)

awx/main/analytics/collectors.py
```diff
@@ -217,30 +218,34 @@ def job_instance_counts(since):
     return counts
 
 
-@register('query_info', '1.0')
-def query_info(since, collection_type):
+@register('query_info', '1.0', description=_('Metadata about the analytics collected'))
+def query_info(since, collection_type, **kwargs):
     query_info = {}
     query_info['last_run'] = str(since)
     query_info['current_time'] = str(now())
```
Ladas: this should probably have the gather time? Although we don't use it anywhere

benthomasson:
Can we incorporate https://github.com/ansible/awx/issues/7658#issuecomment-662655556 to guarantee known file sizes?
This splits the csv files after the queries are done.

wenottingham:
~If we do that, we don't need to necessarily gather quite as often. It's a large chunk of code I'd rather not have to include, though~

ETA: added something similar, but done before we create the tarball. Now to test.....

awx/main/analytics/core.py
wenottingham: I've updated the management command to *allow* for setting a since and until, but don't want to require it. So I'll leave this like this.

awx/main/analytics/core.py
wenottingham: see above

awx/main/tasks.py
wenottingham: done

awx/main/analytics/collectors.py
wenottingham: Yep, fixed.

awx/main/analytics/core.py
Ladas: so we load this in tasks already as
```
last_gather = Setting.objects.filter(key='AUTOMATION_ANALYTICS_LAST_GATHER').first()
```

So Ideally we don't do this on multiple places.

To avoid bugs, there should be 1 place setting the timeframe, then all collections would use the same timeframe.

awx/main/tasks.py
```diff
@@ -372,16 +391,29 @@ def gather_analytics():
...
+                    incremental_collectors.append(collector)
+
+            # Cap gathering at 4 weeks of data if there has been no data gathering
+            since = last_time or (gather_time - timedelta(weeks=4))
+
+            if incremental_collectors:
+                start = since
+                until = None
+                while start < gather_time:
+                    until = start + timedelta(hours = 4)
```
Ladas, line 411: @benthomasson @jctanner @wenottingham just to make sure, the probability we'd blow a /tmp disk size with 4h chunk is close to 0, right?

awx/main/tasks.py
```diff
@@ -372,16 +391,29 @@ def gather_analytics():
...
-                    os.remove(tgz)
+            subset = list(all_collectors().keys())
+            incremental_collectors = []
+            for collector in expensive_collectors():
+                if collector in subset:
+                    subset.remove(collector)
+                    incremental_collectors.append(collector)
+
+            # Cap gathering at 4 weeks of data if there has been no data gathering
+            since = last_time or (gather_time - timedelta(weeks=4))
```
Ladas, line 405: @wenottingham we wanted to go over 4 weeks in https://github.com/RedHatInsights/tower-analytics-backend/issues/97 ?

Is this the right time? :-)

Then we should probably send the latest chunk first (so customers can see their latest data quickly)

wenottingham: We'd need to use different timestamping/tracking then.

awx/main/tasks.py
wenottingham: Not zero, but less than we're doing now with a potential 4 week query.

awx/main/analytics/core.py
```diff
@@ -51,40 +79,35 @@ def projects_by_scm_type():
...
+        path = os.path.join(destdir, 'manifest.json')
+        with open(path, 'w', encoding='utf-8') as f:
+            try:
+                json.dump(manifest, f)
+            except Exception:
+                f.close()
+                os.remove(f.name)
+                logger.exception("Could not generate manifest.json")
+
+    last_run = since or settings.AUTOMATION_ANALYTICS_LAST_GATHER or (now() - timedelta(weeks=4))
```
Ladas, line 103: @wenottingham my old comment meant that we already load this config once in tasks
```
last_gather = Setting.objects.filter(key='AUTOMATION_ANALYTICS_LAST_GATHER').first()
```

So ideally we'd load this somewhere above, once. So we avoid any bugs with different collections getting different timeframes. (so making sure there is 1 time window we collect and ack for all collections)

awx/main/analytics/collectors.py
wenottingham: Given how we pass in the query (and mangle it in the unit & integration tests), I'd rather not redo this for this PR. But good to keep in mind if we change this later.

awx/main/analytics/collectors.py
wenottingham: Updated.

awx/main/tasks.py
wenottingham: We could move this to a temporary directory under /var/lib/awx, but at a certain point if you're running out of space there's not much we can do.

awx/main/tasks.py
wenottingham: Resolving this, we'd need a different way to track what we've gathered and sent to do most-recent-first. Not saying we can't, but don't want to do that now.

awx/main/analytics/core.py
wenottingham: This is this way to handle being acalled from the management command (awx/main/management/commands/gather_analytics.py)

wenottingham:
OK, seems to be doing the right thing, splitting as needed, and sends data that is appearing on cloud.redhat.com

awx/main/tasks.py
Ladas: so let's try to see how this performs in the perf tests, it would have to be quite huge env so the 4h goes over /tmp (cc @kdelee )

And as another PR, we could try what @benthomasson mentioned I think, which is pagination. Using marker based pagination always getting e.g. max 100k events/jobs/etc. (the marker should be something unique like `created, id` combination, that we can use as a cursor moving through data , but maybe just `created` will work, it would get stuck if we'd have 100k events with the same created :-) ) Or we could try actual PG cursor.

In any way, the pagination will help us control max records we put under /tmp . (and the loading of the data will be probably more effective)

awx/main/analytics/core.py
Ladas: I see, then lets keep it this way (we can tweak it later if needed)

awx/main/analytics/collectors.py
```diff
@@ -31,8 +32,8 @@ def something(since):
 '''
 
 
-@register('config', '1.1')
-def config(since):
+@register('config', '1.1', description=_('General platform configuration. This collector cannot be disabled.'))
```
ryanpetrello: "This collector cannot be disabled" (I think we removed that functionality from this PR)

awx/main/analytics/collectors.py
ryanpetrello: `SELECT COUNT(*)` is going to be _incredibly_ slow for customers with very large `main_jobevent` tables (billions of rows) https://wiki.postgresql.org/wiki/Slow_Counting.  We go to great lengths to avoid ever getting an accurate `COUNT(*)` on this table - we should avoid doing it here at all costs.

ryanpetrello: If you wanted to count lines/rows, you could pass in a file-like (maybe a subclass of `io. io.TextIOBase` or something) object that counted lines as they were written by `copy_expert`.  Or you could just do what you're doing here, which is probably clear enough.

awx/main/analytics/collectors.py
```diff
@@ -262,11 +267,14 @@ def _copy_table(table, query, path):
                               main_jobevent.event_data::json->'res'->'warnings' AS warnings,
                               main_jobevent.event_data::json->'res'->'deprecations' AS deprecations
                               FROM main_jobevent 
-                              WHERE main_jobevent.created > {}
-                              ORDER BY main_jobevent.id ASC) TO STDOUT WITH CSV HEADER'''.format(since.strftime("'%Y-%m-%d %H:%M:%S'"))
-    if not subset or 'events' in subset:
-        _copy_table(table='events', query=events_query, path=full_path)
+                              WHERE (main_jobevent.created > {} AND main_jobevent.created <= {})
+                              ORDER BY main_jobevent.id ASC) TO STDOUT WITH CSV HEADER
+                   '''.format(since.strftime("'%Y-%m-%d %H:%M:%S'"),until.strftime("'%Y-%m-%d %H:%M:%S'"))
```
ryanpetrello: Is there some reason `.isoformat()` doesn't work here instead of spelling it out something similar each time with `strftime`?

awx/main/analytics/collectors.py
ryanpetrello: Technically speaking @wenottingham, it might actually be better to implement the splitting here _at this layer_ instead of writing one file and then chunking it up into multiple later.  You've already got a file-like interface here; you _could_ count bytes as you go, and swap file references at certain thresholds.

awx/main/analytics/collectors.py
wenottingham: Thanks, fixed.

awx/main/analytics/collectors.py
Ladas: @ryanpetrello I think @jameswnl referred to `quick count` if PG would do index scan. So e.g. `select count(*) from events where created > ... and created < ...` should do index scan, if it's small enough portion of the table. (it should be quicker than fetching that timeframe, since it just does scan+count)

But yeah, if PG will choose sequential scan for the the count or the copy on large table, the perf will be horrible. :-)

jameswnl: Yeah, if we are uncertain on the cost of `select count(*)`, we should stay with the conservative one.

awx/main/analytics/collectors.py
wenottingham: did this

wenottingham:
> Technically speaking @wenottingham, it might actually be better to implement the splitting here at this layer instead of writing one file and then chunking it up into multiple later. You've already got a file-like interface here; you could count bytes as you go, and swap file references at certain thresholds.

I've done this now, works in testing. Although if psycopg2:copy_expert ever stops line-buffering its writes this will all fail horribly.

wenottingham:
recheck

awx/main/analytics/collectors.py
```diff
@@ -217,30 +219,71 @@ def job_instance_counts(since):
...
-def copy_tables(since, full_path, subset=None):
-    def _copy_table(table, query, path):
-        file_path = os.path.join(path, table + '_table.csv')
-        file = open(file_path, 'w', encoding='utf-8')
-        with connection.cursor() as cursor:
-            cursor.copy_expert(query, file)
-            file.close()
-        return file_path
+'''
+The event table can be *very* large, and we have a 100MB upload limit.
```
rooftopcellist, line 232: This note should probably be updated to 200MB

awx/main/analytics/core.py
```diff
@@ -154,6 +230,9 @@ def ship(path):
     if not path:
         logger.error('Automation Analytics TAR not found')
         return
+    if not os.path.exists(path):
+        logger.error('Automation Analytics TAR {} not found'.format(path))
```
rooftopcellist, line 234: This is kicking back an error for every letter of [this](https://github.com/ansible/awx/blob/devel/awx/main/analytics/core.py#L35) error message when the license is not enabled.

awx/main/analytics/collectors.py
wenottingham: From chat - the upload limit on the final compressed tarball is 100MB. We just cap any single file at 200MB uncompressed

rooftopcellist (changes requested):

awx/main/management/commands/gather_analytics.py
```diff
@@ -15,6 +18,10 @@ def add_arguments(self, parser):
                             help='Gather analytics without shipping. Works even if analytics are disabled in settings.')
         parser.add_argument('--ship', dest='ship', action='store_true',
                             help='Enable to ship metrics to the Red Hat Cloud')
+        parser.add_argument('--since', dest='since', action='store',
+                            help='Start date for collection')
+        parser.add_argument('--until', dest='until', action='store',
+                            help='End date for collection')
```
rebeccahhh, line 24: Big fan of this addition

awx/main/analytics/core.py
wenottingham: https://github.com/ansible/awx/pull/7709/commits/13802fcf2bfa137aa642ee9b5eb2d187b420b269 should cover this, I think.

rooftopcellist:
recheck

rooftopcellist (approved):

awx/settings/defaults.py
```diff
@@ -1001,7 +1001,7 @@ def IS_TESTING(argv=None):
             'level': 'INFO',  # very verbose debug-level logs
         },
         'awx.analytics': {
-            'handlers': ['external_logger'],
+            'handlers': ['task_system', 'external_logger'],
```
AlanCoding, line 1004: This absolutely hammers the log file `/var/log/tower/task_system.log` with messages like:

```
2020-09-17 01:20:52,346 INFO     awx.analytics.performance api response times
2020-09-17 01:20:52,683 INFO     awx.analytics.job_events Event data saved.
2020-09-17 01:20:59,555 INFO     awx.analytics.activity_stream Activity Stream update entry for job_template
```

These have no more information than this, and it's not usable. This was never intended to be sent to file based logs, only to the external logger. What motivated this specific change?

A simple test run now generates 1.5 MB for this whereas tower.log is 130 KB, and I would expect them to be pretty similar.

wenottingham: The process of sending analytics is split between awx.main.analytics and awx.main.tasks.

Before this change, both of these would log, and they'd end up in entirely separate places. To find errors, you'd have to look in a different place based on which particular line of code threw the error.

wenottingham: Whoops, this adjusted the wrong logger.  PR incoming.

PR #7719, opened 2020-07-24
work around a memory leak in channels_redis

ryanpetrello:
cc @gamuniz

### DIFF
--- a/awx/main/consumers.py
+++ b/awx/main/consumers.py
@@ -1,3 +1,4 @@
+import collections
 import json
 import logging
 import time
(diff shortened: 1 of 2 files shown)


ryanpetrello:
I filed an upstream issue to track the observed memory leak in `channels-redis`:

https://github.com/django/channels_redis/issues/212

ryanpetrello:
@chrismeyersfsu I may tinker w/ this a bit more - I'm not totally convinced that it doesn't also leak a little.

ryanpetrello:
Closing this, because it also leaks.

I think we need to fix this in `channels_redis`.

Issue #7339, opened 2020-06-15
Add Test button to credential form for external credential types


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
This is sort of related to #7338 in that they both interact with the same endpoint but the UX is a bit different on this button.  To be clear, this is the button we're talking about in this issue:

<img width="1513" alt="Screen Shot 2020-06-15 at 11 32 07 AM" src="https://user-images.githubusercontent.com/9889020/84676834-f180dd00-aefb-11ea-8ac1-0362ca54a84a.png">

Clicking on that button should open up a modal that looks like:

<img width="818" alt="Screen Shot 2020-06-15 at 11 33 09 AM" src="https://user-images.githubusercontent.com/9889020/84676929-0493ad00-aefc-11ea-807d-26d3968038d4.png">

These are the same form fields present in the Metadata step of the credential plugin wizard (based on credential type).

Once a user fills in the required fields and submits, a request to `/api/v2/credentials/:id/test` should be issued and an alert shown in the upper right corner indicating the result of the test:

<img width="1274" alt="Screen Shot 2020-06-15 at 11 34 41 AM" src="https://user-images.githubusercontent.com/9889020/84677091-399fff80-aefc-11ea-9d39-6679847d2a35.png">

mabashian:
@unlikelyzero I think we can close this now?

unlikelyzero:
covered by credential-mock

PR #7730, opened 2020-07-27
Adds support for a Test button on the credential form when the credential type is 'external'

mabashian:
##### SUMMARY
link #7339 

Here's the button:

Screen Shot 2020-07-27 at 4 18 03 PM

This button only appears on the add/edit form when the user has selected a credential type with a `kind` of `external`.  The button is _disabled_ until the form valid and all required fields are provided.

Clicking the button will open up a modal which prompts users to provide values for the `metadata` fields.  Some of these fields may be required.

We use a combination of the `input` values from the underlying Credential form along with the `metadata` field values to POST to either the credential's `test` endpoint or the credential type's `test` endpoint.  We use the credential type endpoint _unless_ the user is editing a credential and hasn't changed the credential type from whatever was previously saved.

Executing the test will show an Alert in the upper right corner akin to https://github.com/ansible/awx/pull/7338.

There's actually some overlapping components between this PR and that other one.  That'll get cleaned up when one of these merges and the other is rebased.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/screens/Credential/shared/CredentialFormFields/CredentialPlugins/CredentialPluginTestAlert.jsx
+++ b/awx/ui_next/src/screens/Credential/shared/CredentialFormFields/CredentialPlugins/CredentialPluginTestAlert.jsx
@@ -1,0 +1,91 @@
+import React, { useEffect, useState } from 'react';
+import { withI18n } from '@lingui/react';
+import { t } from '@lingui/macro';
+import { string, shape } from 'prop-types';
+import {
+  Alert,
+  AlertActionCloseButton,
+  AlertGroup,
+} from '@patternfly/react-core';
+
+function CredentialPluginTestAlert({
+  i18n,
+  credentialName,
+  successResponse,
+  errorResponse,
+}) {
+  const [testMessage, setTestMessage] = useState('');
+  const [testVariant, setTestVariant] = useState(false);
+
+  useEffect(() => {
+    if (errorResponse) {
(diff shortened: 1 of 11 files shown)


jakemcdermott (approved):

awx/ui_next/src/screens/Credential/shared/CredentialFormFields/CredentialPlugins/CredentialPluginTestAlert.jsx
```diff
@@ -0,0 +1,91 @@
...
+  i18n,
+  credentialName,
+  successResponse,
+  errorResponse,
+}) {
+  const [testMessage, setTestMessage] = useState('');
+  const [testVariant, setTestVariant] = useState(false);
+
+  useEffect(() => {
+    if (errorResponse) {
```
jakemcdermott, line 21: This conditional might not be needed. I think the conditional below it should be enough:

`if (errorResponse?.response?.data?.inputs) {`

jakemcdermott:
LGTM so far. I would like to take another look once we merge https://github.com/ansible/awx/pull/7338 and integrate the changes into this.

mabashian:
Sounds good @jakemcdermott

unlikelyzero:
@dsesami hey, this is one i'm working on atm

unlikelyzero (approved):

unlikelyzero:
Tested via https://github.com/ansible/tower-qa/pull/5449

PR #7738, opened 2020-07-28
Test console cleanup

mabashian:
##### SUMMARY
All of the noise fell into one of 3 categories:

1) Console errors regarding isValid - I think that this prop changed to `validated` with the most recent PF major version bump but there were a few places where we were still using it and it needed to be cleaned up.
2) Console errors regarding certain code snippets in tests not being wrapped in act().
3) Socket debug messages - connect/disconnect, that type of thing.

I addressed the first two with code changes.  To address the third, I decided to globally mock `console.debug` to prevent those messages from being logged to the console.  At this point I don't know if I see a lot of value in having debug statements show up in the output of tests but if someone has a good reason why they're needed then I'm happy to come up with an alternative solution.

The only thing that's still being logged is a warning in the App.jsx test:

```
console.warn node_modules/tiny-warning/dist/tiny-warning.cjs.js:13
Warning: You are attempting to use a basename on a page whose URL path does not begin with the basename. Expected path "/" to begin with "/next".
```

which I think is fine.  The `/next` solution will be around for a while but eventually we'll get rid of it when the app can fully replace the existing UI.  At that time this would go away.

### DIFF
--- a/awx/ui_next/src/setupTests.js
+++ b/awx/ui_next/src/setupTests.js
@@ -12,3 +12,10 @@ require('@nteract/mockument');
 
 // eslint-disable-next-line import/prefer-default-export
 export const asyncFlush = () => new Promise(resolve => setImmediate(resolve));
+
+// this ensures that debug messages don't get logged out to the console
+// while tests are running i.e. websocket connect/disconnect
+global.console = {
+  ...console,
+  debug: jest.fn(),
+};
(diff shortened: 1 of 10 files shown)


awx/ui_next/src/setupTests.js
```diff
@@ -12,3 +12,10 @@ require('@nteract/mockument');
 
 // eslint-disable-next-line import/prefer-default-export
 export const asyncFlush = () => new Promise(resolve => setImmediate(resolve));
+
+// this ensures that debug messages don't get logged out to the console
+// while tests are running i.e. websocket connect/disconnect
+global.console = {
+  ...console,
+  debug: jest.fn(),
```
jakemcdermott, line 20: :+1: works for me. nice

jakemcdermott (approved):

Issue #7740, opened 2020-07-28
Pass 2 of making options requests to get possible keys for advanced search


jlmitch5:
There are a handful of screens that don't have anything populated in the options for the key typeahead, as they don’t use the useRequest hook yet. In terms of work, I think it makes sense that they should be converted first so that they can grab the options request to get possible advanced keys.

- UserList.jsx - https://github.com/ansible/awx/pull/7743
- WFJTList - https://github.com/ansible/awx/pull/7745
- (Templates) ProjectsList.jsx - https://github.com/ansible/awx/pull/7752
- JTList - https://github.com/ansible/awx/pull/7754
- InventorySourcesList.jsx - https://github.com/ansible/awx/pull/7755
- ProjectJobTemplatesList.jsx - https://github.com/ansible/awx/pull/7759
- OrganizationTeamList.jsx - https://github.com/ansible/awx/pull/7761
- InventoryGroupsList.jsx - https://github.com/ansible/awx/pull/7765
- ResourceAccessList.jsx - https://github.com/ansible/awx/pull/7772
- AddResourceRole lists
- UserAndTeamAccessAdd /SelectResourceStep lists*
- AssociateModal lists*
- NotificationList.jsx - https://github.com/ansible/awx/pull/7785
- OrganizationLookup - https://github.com/ansible/awx/pull/7786

*these actually need to be passed an additional options request from the places they are called, they have useRequest in use right now, so bit of a different conversion

mabashian:
I'm gonna slowly chip away at this list - if anyone else is interested in grabbing some of these just let me know so that we don't duplicate efforts.

I'll probably do each component in a separate PR

mabashian:
I think these:

```
AddResourceRole lists
UserAndTeamAccessAdd /SelectResourceStep lists*
AssociateModal lists*
```

will be a bit easier to address once #7739 merges since we'll be able to add the options request and hook up advanced search at the same time.

So I'm gonna throw this back into needs_devel for that work

jlmitch5:
DONE: As part of the follow up to this, we want to slightly tweak the logic of the options calls to get search keys in all the existing places

![Screen Shot 2020-08-05 at 1 49 15 PM](https://user-images.githubusercontent.com/1342624/89447125-7aa9d880-d723-11ea-91d2-f4be8dee200d.png)

jlmitch5:
[...]

PR #7743, opened 2020-07-28
Converts UserList to functional component

mabashian:
##### SUMMARY
Link https://github.com/ansible/awx/issues/7740

This addresses UserList.jsx


mabashian:
@nixocio pointed out that we not have a `useSelected` hook which can probably be used here.  I'll take a look at that before we merge this in.

nixocio (approved):

jlmitch5 (approved):
Wow such a big improvement!

PR #7745, opened 2020-07-28
Converts WorkflowJobTemplatesList to use useRequest in preparation for advanced search

mabashian:
##### SUMMARY
link https://github.com/ansible/awx/issues/7740
See title.  I didn't include an OPTIONS request because it would generate an unused variable but that can easily be changed when advanced search is in play.


jakemcdermott:
@jlmitch5 Should we fast-track this PR and merge it now so you can rebase it into adv search?

jakemcdermott (approved):

PR #7752, opened 2020-07-29
Converts ProjectsList to use useRequest in preparation for advanced search

mabashian:
##### SUMMARY
link #7740
See title. I didn't include an OPTIONS request because it would generate an unused variable but that can easily be changed when advanced search is in play.


jlmitch5 (approved):
Nice!

PR #7757, opened 2020-07-29
WIP: cache main_jobevent Insights gathering to cut down on expensive queries

ryanpetrello:
cc @Ladas @benthomasson


ryanpetrello:
Downside to this approach is you have to collect on _every_ node.

Issue #5150, opened 2019-10-29
Support arbitrary custom inventory plugin configuration


wenottingham:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Followon of https://github.com/ansible/awx/issues/5149 (and its predecessors).

Given the ability to expose a custom inventory plugin via a collection for use as an inventory source, we should allow configuring a generic custom inventory source where the configuration data is merely "a plugin configuration file".

From a related (duplicate) issue filed by @AlanCoding:

Right now, there is no way to use an inventory plugin-style YAML configuration aside from checking it into source control. We have had several inquiries related to this, so this is proposed as a solution to resolve all of those use cases.

Things we can say this feature would necessary need to have:

 - form elements
   - Place to select a credential or multiple credentials
   - Place to enter YAML content of the inventory plugin (no JSON)
   - (unclear) filename may or may not need to be independently specified, the only known stubborn exception that could require this is [VMWare](https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/inventory/vmware_vm_inventory.py#L278)
   - any other shared elements with whatever type of resource it is (like NAME)
 - The YAML content will be required to have `plugin: foo` as a top-level entry
 - The form will enforce that supported types use a credential of its corresponding type
   - possible additional feature - allow custom types to claim an inventory plugin namespace
 - Inventory updates already use plugins, this will do very little different, just using the config that the user gives
 - A "convert to plugin" button on existing inventory sources that work with plugins
   - this will start the user out with a configuration that gives compatibility with existing behavior, which they can then modify at will

Questions still remain. Will this be another type, similar to inventory scripts? Or will it be another option for `source` of an inventory source? Will there be any way for the user to alter ansible.cfg options, like the group name conversion?

AlanCoding:
Same as https://github.com/ansible/awx/issues/4306?

AlanCoding:
the main important discussion point I want to copy over from the other issue:

> A "convert to plugin" button on existing inventory sources that work with plugins
this will start the user out with a configuration that gives compatibility with existing behavior, which they can then modify at will

And this is all non-specific to collections. I don't see any tie-ins or conflicts.

ryanpetrello:
@AlanCoding is this issue description:

https://github.com/ansible/awx/issues/4306

...still indicative of the work we want to do for this enhancement?

cc @chrismeyersfsu

ryanpetrello:
[...]

PR #7763, opened 2020-07-29
Feature inv plugin conf

chrismeyersfsu:
related to https://github.com/ansible/awx/issues/5150

### TODO:
- [x] After talking with @AlanCoding . I have `source="scm"` all wrong. `source_vars` are irrelevant  for SCM source & we should be pointing to the inventory file in the SCM, which will point to the inventory plugin via `plugin: `
- [x] Remove the migration rewind before merge.


awx/main/models/inventory.py
```diff
@@ -2028,76 +1467,10 @@ def get_plugin_env(self, *args, **kwargs):
...
-            'gce_image': 'image',
-            # We need this as long as hostnames is non-default, otherwise hosts
-            # will not be addressed correctly, was returned in script
-            'ansible_ssh_host': 'networkInterfaces[0].accessConfigs[0].natIP | default(networkInterfaces[0].networkIP)'
-        }
-
     def inventory_as_dict(self, inventory_update, private_data_dir):
-        ret = super(gce, self).inventory_as_dict(inventory_update, private_data_dir)
+        ret = super().inventory_as_dict(inventory_update, private_data_dir)
         credential = inventory_update.get_cloud_credential()
```
chrismeyersfsu, line 1494: This is the 1 inventory plugin type where the config lives in a credential and has to be transferred to the inventory `gce.yml`. We can solve this in the inventory plugin by accepting the project list as an env var & update our credential type accordingly.

awx/main/migrations/0118_v380_inventory_plugins.py
```diff
@@ -0,0 +1,83 @@
...
+            with open(BACKUP_FILENAME, 'w') as fh:
+                json.dump(source_vars_backup, fh)
+
+            injector = FrozenInjectors[inv_source_obj.source]()
+            new_inv_source_vars = injector.inventory_as_dict(inv_source_obj, None)
+            inv_source_obj.source_vars = new_inv_source_vars
+            inv_source_obj.save()
+
+
+def inventory_source_vars_backward(apps, schema_editor):
```
chrismeyersfsu: I plan on removing this before merge. It's nice to be able to backward when developing but alas customers almost certainly cram sensitive info into their source_vars

awx/main/models/inventory.py
```diff
@@ -1419,6 +1158,20 @@ def clean_source_path(self):
             raise ValidationError(_("Cannot set source_path if not SCM type."))
         return self.source_path
 
+    def clean_source_vars(self):
+        injector = self.injectors.get(self.source)
+        source_vars = dict(self.source_vars_dict) # make a copy
+        if injector and self.source_vars_dict.get('plugin', '') != injector.get_proper_name():
+            source_vars['plugin'] = injector.get_proper_name()
```
chrismeyersfsu: Add or correct `plugin: a.b.c` for user based on `source` property.

awx/main/models/inventory.py
```diff
@@ -1419,6 +1158,20 @@ def clean_source_path(self):
...
+        elif not injector:
```
chrismeyersfsu: `scm` source type requires the user to specify `plugin: a.b.c`, we can't do it for them. Note that this _could_ break some flows because we don't allow the model to be created/modified if the structure of their `source_vars` isn't correct. We _could_ instead allow `plugin: ` to not be present and fail at runtime.

AlanCoding:
What UI work would be needed to wrap this up?

awx/main/migrations/0118_v380_inventory_plugins.py
```diff
@@ -0,0 +1,83 @@
...
+
+from ._inventory_source_vars import FrozenInjectors
+
+
+logger = logging.getLogger('awx.main.migrations')
+BACKUP_FILENAME = '/tmp/tower_migration_inventory_source_vars.json'
+
+
+def _get_inventory_sources(InventorySource):
+    # TODO: Maybe pull the list of cloud sources from code
```
AlanCoding: I don't know, it's probably better to use a frozen list for the migration

chrismeyersfsu:
Re: UI work needed. Nothing new.
image
Remove instance filters, regions, only group by.

wenottingham:
Thinking of this from a user perspective, this trades in reduced complexity, and increased flexibility, at the expense of usability niceness.

We probably need an option to generate what the frozen migration would generate from a blank config so that users can make new sources that match what a migrated source would be.

Why do <source>_ENABLED_VAR/_ENABLED_VALUE/_GROUP_FILTER still exist in *settings*? If we're doing that much change, we should fix that so it's per-source as well.

awx/ui/client/src/inventories-hosts/inventories/related/sources/sources.form.js
```diff
@@ -126,46 +126,6 @@ export default ['NotificationsList', 'i18n', function(NotificationsList, i18n){
                 includeInventoryFileNotFoundError: true,
                 subForm: 'sourceSubForm'
             },
```
jakemcdermott, line 128: Screenshot from 2020-08-04 17-51-47

awx/ui_next/src/screens/Inventory/shared/InventorySourceForm.jsx
```diff
@@ -200,21 +200,21 @@ const InventorySourceForm = ({
...
-    source_regions: source?.source_regions || '',
     source_script: source?.summary_fields?.source_script || null,
     source_vars: source?.source_vars || '---\n',
     update_cache_timeout: source?.update_cache_timeout || 0,
     update_on_launch: source?.update_on_launch || false,
     update_on_project_update: source?.update_on_project_update || false,
     verbosity: source?.verbosity || 1,
+    enabled_var: source?.enabled_var || '',
+    enabled_value: source?.enabled_value || '',
+    host_filter: source?.host_filter || '',
```
jakemcdermott, line 217: Screenshot from 2020-08-04 17-49-01

awx/main/migrations/0118_inventory_plugins.py
```diff
@@ -0,0 +1,93 @@
...
+            with open(BACKUP_FILENAME, 'w') as fh:
+                json.dump(source_vars_backup, fh)
+
+            injector = FrozenInjectors[inv_source_obj.source]()
+            new_inv_source_vars = injector.inventory_as_dict(inv_source_obj, None)
+            inv_source_obj.source_vars = yaml.dump(new_inv_source_vars)
+            inv_source_obj.save()
+
+
+def inventory_source_vars_backward(apps, schema_editor):
```
chrismeyersfsu: TODO: Remove the rewind. Users could have sensitive data.

awx/ui/client/src/inventories-hosts/inventories/related/sources/sources.form.js
```diff
@@ -340,6 +300,36 @@ export default ['NotificationsList', 'i18n', function(NotificationsList, i18n){
                 ngDisabled: '!(inventory_source_obj.summary_fields.user_capabilities.edit || canAdd)',
                 subForm: 'sourceSubForm'
             },
+            host_filter: {
+                label: i18n._("Host Filter"),
+                type: 'text',
+                dataTitle: i18n._('Host Filter'),
+                dataPlacement: 'right',
+                awPopOver: "<p>" + i18n._("Regular expression where only matching hosts will be imported.") + "</p>",
```
AlanCoding: I think it would be good to mention that this is client-side filtering.

I'm still very worried about confusion with these fields, because there is duplication between them and the inventory plugin options themselves. I wouldn't even hesitate saying that "this is not an inventory plugin parameter"

AlanCoding:
I'm pulling this down and trying to test some things

Screen Shot 2020-08-21 at 10 44 00 AM

I think this should be a little more explicit in order to avoid the inevitable confusion. The Source Variables (can we please call this something else?) may still have filtering options. But this performs an additional filter after the host list is obtained with regex filtering. Also, does anyone know how to write these? Does it filter out hosts if it matches a given expression, or does it only _include_ hosts that match the expression? If I give multiple terms... are they an OR condition or AND?

elyezer:
I am exploring this and found that the value of `source_vars` returned by the API is not consistent. For example, custom source returns the `source_vars` as being a JSON string but others are returning a YAML string. That makes me thing if the source vars should be normalized to JSON (what we had previously) for everything.

chrismeyersfsu:
`source_vars` is a TextField and is not coerced into JSON or YAML when stored because there was a want to store comments. So the behavior you are seeing is expected.

The migration in this PR is the exception. We coerce the `source_vars` to yaml.

elyezer:
@chrismeyersfsu thanks for the clarification, wanted to make sure we were not missing anything.

dsesami (changes requested):
This looks great from a UI standpoint, but I'd like to request a small adjustment:
`1.` Require both enabled var and value and throw an error if only one is set. This is not what the API does, and I assume (perhaps incorrectly) that there is a reason for that.
`2.` If we want to keep validation as-is, can we update the tooltip on enabled_value to mention that it is only used if enabled_var is set, and vice-versa?

@jakemcdermott have agreed to do `#1` and `#2` for ui_next and `#2` for the classic ui.

bonus points: can we have the client double-check that the regex syntax for the host filter is valid? small quality of life change

awx/ui/client/src/inventories-hosts/inventories/related/sources/sources.form.js
jakemcdermott: updated:
Screenshot from 2020-08-27 12-18-48
Screenshot from 2020-08-27 12-19-01
Screenshot from 2020-08-27 12-19-13

cc @dsesami

jakemcdermott:
Alright, this is ready for more review and testing. Thanks for taking a look

awx/ui_next/src/screens/Inventory/shared/InventorySourceSubForms/SharedFields.jsx
```diff
@@ -192,6 +192,7 @@ export const HostFilterField = withI18n()(({ i18n }) => {
       )}
       name="host_filter"
       type="text"
+      validate={regExp(i18n)}
     />
```
jakemcdermott, line 196: Screenshot from 2020-08-27 11-59-49
cc @dsesami

awx/ui/client/src/inventories-hosts/inventories/related/sources/sources.form.js
```diff
@@ -340,6 +300,36 @@ export default ['NotificationsList', 'i18n', function(NotificationsList, i18n){
...
+            enabled_value: {
+                label: i18n._("Enabled Value"),
+                type: 'text',
+                dataTitle: i18n._('Enabled Value'),
+                dataPlacement: 'right',
+                awPopOver: "<p>" + i18n._("This field is ignored unless an Enabled Variable is set. If the enabled variable matches this value, the host will be enabled on import.") + "</p>",
+                dataContainer: 'body',
+                ngDisabled: '!(inventory_source_obj.summary_fields.user_capabilities.edit || canAdd)',
+                subForm: 'sourceSubForm'
+            },
```
jakemcdermott, line 332: Screenshot from 2020-08-27 12-18-48
Screenshot from 2020-08-27 12-19-01
Screenshot from 2020-08-27 12-19-13

cc @dsesami @AlanCoding @chrismeyersfsu

jlmitch5 (approved):
UI code looks good!

AlanCoding:
put up https://github.com/chrismeyersfsu/awx/pull/6

unlikelyzero:
Please merge this at the same time:
https://github.com/ansible/tower-qa/pull/5464

Unrelated bugs discovered during testing:
https://github.com/ansible/awx/issues/8044
https://github.com/ansible/awx/issues/8045

awx/main/tasks.py
```diff
@@ -2567,23 +2562,18 @@ def build_args(self, inventory_update, private_data_dir, passwords):
...
-        if getattr(settings, '%s_HOST_FILTER' % src.upper(), False):
-            args.extend(['--host-filter',
-                         getattr(settings, '%s_HOST_FILTER' % src.upper())])
+        if inventory_update.enabled_var:
+            args.extend(['--enabled-var', shlex.quote(inventory_update.enabled_var)])
+            args.extend(['--enabled-value', shlex.quote(inventory_update.enabled_value)])
+        else:
+            if getattr(settings, '%s_ENABLED_VAR' % src.upper(), False):
+                args.extend(['--enabled-var',
+                            getattr(settings, '%s_ENABLED_VAR' % src.upper())])
```
AlanCoding, line 2563: so now these become the defaults?

AlanCoding (approved):
From a technical perspective, I think got all the questions answered about expectations for how problems are solved. Really happy to see this happen 👍  it'll help a lot moving to jobs in containers

awx/main/tasks.py
AlanCoding: answer: yes

dsesami (approved):
approved from a UI perspective.

ryanpetrello:
recheck

PR #7772, opened 2020-07-30
Converts ResourceAccessList to functional component

mabashian:
##### SUMMARY
link https://github.com/ansible/awx/issues/7740

This component now uses useRequest as well as useDeleteItems.

Since this component has custom "add" and "delete" modals it's a little messier than I would like but it should do for now.

I also had to clean up the tests a bit since they were doing some unsavory checks which of course broke when the underlying component internals changed.


AlexSCorey (approved):
This looks good to me.

Issue #7769, opened 2020-07-30
Update conditional in project update playbook - require min Ansible 2.9 to install collections


AlanCoding:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
This needs to be backed off from 2.8 to 2.9:

https://github.com/ansible/awx/blob/708244881cc6cd04722d3bc36d7d6036bffc329a/awx/playbooks/project_update.yml#L165

Because if you run with Ansible 2.8, it gives blank `stdout` and that's confusing.

If my research is correct, @jborean93 added the `ansible-galaxy collection` CLI in https://github.com/ansible/ansible/pull/57106, which was 2.9. Version 2.8 seems to mistake these invocations for a role command, and it doesn't work right.

kdelee:
This is done, now collections will only be installed if ansible running the project update is >= 2.9

AlanCoding:
We might also want to change some conditional skips in tower-qa, because the reason we saw this was because some tests were failing unexpectedly in 2.8. Best approach may be just to run tests with 2.8 and see what happens.

PR #7782, opened 2020-07-31
only support ansible-galaxy collection in 2.9+

ryanpetrello:
see: https://github.com/ansible/awx/issues/7769


AlanCoding (approved):

wenottingham:
We're going to be tied to 2.9+ soon enough, but it's fine to have this still be a check.

ryanpetrello:
regate

PR #7787, opened 2020-07-31
Fix collection test inventory

squidboylan:


kdelee (approved):
pesky ole "no host in inventory"

AlanCoding (approved):

PR #7790, opened 2020-07-31
work around a memory leak in channels_redis

ryanpetrello:
see: https://github.com/django/channels_redis/issues/212
see: https://github.com/django/channels_redis/pull/213/

### DIFF
--- a/awx/main/consumers.py
+++ b/awx/main/consumers.py
@@ -1,3 +1,4 @@
+import collections
 import json
 import logging
 import time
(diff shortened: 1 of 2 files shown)


awx/main/consumers.py
```diff
@@ -12,12 +13,53 @@
...
+            if self._expires[k] < time.time():
+                expired.append(k)
+            else:
+                # as this is an OrderedDict, every key after this
+                # was inserted *later*, so if _this_ key is *not* expired,
+                # the ones after it aren't either (so we can stop iterating)
+                break
+        for k in expired:
+            del self._expires[k]
+            del self[k]
```
jakemcdermott, line 54: I think `self.pop(k, None)` would work the same w/out needing the `del` + `__delitem__` override

ryanpetrello: Yes, **but** the reason I defined a `__delitem__` is because the underlying redis cache layer *itself* is asyncio code that sometimes *also* deletes these keys as part of callbacks.  So there are situations where there's a race to see which coroutine will clean up first.  In other words, code in `channels_redis` tries to  `del self.receive_buffer[channel]`, but my code here has already beat it to the punch (and it doesn't like that). So I decided to just make KeyErrors on `del` just be an "okay fine 🤷" to simplify my life.

jakemcdermott: Ah. Yep. I see now: https://github.com/django/channels_redis/blob/master/channels_redis/core.py#L462

Unless you override the `del`, `channels_redis` will hit the error if it loses the race.

AlanCoding:
Is the state of this that it's still unclear if fix will land here vs. the upstream library?

ryanpetrello:
@AlanCoding I'd like this to land upstream in `channels_redis`, but if it doesn't, we may carry this on our side for awhile.

ryanpetrello:
Closing this for now because it doesn't appear to work.

PR #7794, opened 2020-08-02
Enable custom settings during installation

mamercad:
##### SUMMARY

Adds the ability to deploy custom settings during installation.
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - Installer

##### AWX VERSION

```
awx: 13.0.0
```

##### ADDITIONAL INFORMATION

The input required is a JSON input file with AWX/Tower settings that the user would like to set. This file is turned into a Kubernetes/OpenShift ConfigMap, which is subsequently mounted into a new Kubernetes/OpenShift Job (which is responsible for applying the settings). The settings job is based on `alpine:latest` and does two things: installs `curl` and hits the AWX/Tower API in order to apply the settings.

Example from my home lab:

```
❯ kubectl -n awx get configmap awx-settings -o jsonpath='{.data}'
map[settings.jonagold.json:{
     "ACTIVITY_STREAM_ENABLED": true,
     "ACTIVITY_STREAM_ENABLED_FOR_INVENTORY_SYNC": false,
     "ORG_ADMINS_CAN_SEE_ALL_USERS": true,
     "MANAGE_ORGANIZATION_AUTH": true,
     "TOWER_URL_BASE": "http://192.168.1.201",
     "REMOTE_HOST_HEADERS": [
          "HTTP_X_FORWARDED_FOR"
     ]
}
]
```

```
❯ kubectl -n awx get job awx-settings-job
NAME               COMPLETIONS   DURATION   AGE
awx-settings-job   1/1           4s         30m
```

```
❯ awx settings list | grep TOWER_URL_BASE
     "TOWER_URL_BASE": "http://192.168.1.201",
```

### DIFF
--- a/installer/inventory
+++ b/installer/inventory
@@ -139,6 +139,12 @@ secret_key=awxsecret
 # it manages from the docker host, you can set this to turn it into a volume for the container.
 #project_data_dir=/var/lib/awx/projects
 
+# Controls if custom settings (overrides) are deployed during installation.
+# For a complete list of settings: `awx settings list -f json`.
+# See inventory/settings.example.json for an example; the settings defined
+# in this file will be `PATCH`ed in.
+#custom_settings_file=/path/to/your/custom_settings.json
+
 # AWX custom virtual environment folder. Only usable for local install.
 #custom_venv_dir=/opt/my-envs/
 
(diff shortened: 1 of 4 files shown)


wenottingham:
I feel like doing this via a separate pod isn't right... especially when you could just configmap in a settings file?

mamercad:
> I feel like doing this via a separate pod isn't right... especially when you could just configmap in a settings file?

The thought behind using a Kubernetes Job was that deploying settings was a one-time operation (not that it *has* to be that way).

ryanpetrello:
@mamercad sorry for the delay on reviewing this PR - we've in the process of replacing the k8s installer in `devel` (https://groups.google.com/g/awx-project/c/47MjWSUQaOc/m/bCjSDn0eBQAJ https://github.com/ansible/awx/tree/devel/tools/docker-compose) and thus the next major release of AWX with an operator (https://github.com/ansible/awx-operator), so I suspect this PR is obsolete now.

Issue #7795, opened 2020-08-03
Using "yaml" as a format when exporting a resource on Python 3 generates garbage


zedr:
##### ISSUE TYPE
 - If one runs the command `awx export -f yaml` using a Python 3 interpreter, the literal representation of the generated bytestring will be printed, instead of the proper, parseable YAML representation.

##### SUMMARY
The command `awx export -f yaml` will print the literal representation of the generated bytestring, e.g. `b"workflow_job_templates:\n- allow_simultaneous: true\n ...` if the interpreter is Python 3. This happens because the behaviour of the `print()` function, in respect to how it handles bytestrings (`bytes`), has changed compared to Python 2.

For example:
```
Python 2.7.5 (default, Mar 20 2020, 17:08:22) 
>>> print(b'hello')
hello
```

```
Python 3.6.8 (default, Sep 26 2019, 11:57:09)
>>> print(b'hello')
b'hello'
```

In `awxkit.awxkit.cli.resource.py:231`:
```
            formatted = format_response(
                Page.from_json(response, connection=connection),
                fmt=client.get_config('format'),
                filter=_filter
            )
           print(formatted)
```
`formatted` is always of type `str` in Python 2. In Python 3 it is of type `str` if the JSON formatter is chosen, and of type `bytes` if the YAML formatter is chosen.

##### ENVIRONMENT
* AWX version: 13.0.0
* AWX install method: source
* Ansible version:  2.9.10
* Operating System: Linux (RHEL 7)
* Web Browser: Google Chrome

##### STEPS TO REPRODUCE

1. Install awx in a Python 3 virtual environment
2. `awx -f yaml export`

##### EXPECTED RESULTS

The command will should print a parseable representation of the YAML payload, e.g. (ellipsed for brevity)

```yaml
workflow_job_templates:
- allow_simultaneous: true
...
```

##### ACTUAL RESULTS

The command will print the following garbage on stdout (ellipsed for brevity):
```
b"workflow_job_templates:\n- allow_simultaneous: true\n ...`
```

kdelee:
Tests passing with patch merged, also verified manually locally using `awx me -f yaml` as well as ` awx workflow_job_templates list -f yaml` going to call this verified, thanks for the patch!

PR #6327, opened 2020-03-18
remove python2 support from awxkit

ryanpetrello:

PR #7796, opened 2020-08-03
Fix garbage being printed when exporting as YAML - related #7795

zedr:
##### SUMMARY
This resolves issue #7795, by passing the `encoding` keyword argument
only when the code is run on a Python 2 interpreter.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
CLI

##### AWX VERSION
```
awx: 13.0.0
```

### DIFF
--- a/awxkit/awxkit/cli/format.py
+++ b/awxkit/awxkit/cli/format.py
@@ -130,7 +130,6 @@ def format_yaml(output, fmt):
     return yaml.safe_dump(
         output,
         default_flow_style=False,
-        encoding='utf-8',
         allow_unicode=True
     )
 


ryanpetrello (approved):
Thanks for contributing this @zedr - good find!

ryanpetrello:
@zedr, could you actually just update this to remove the argument entirely?  The awx CLI actually no longer supports python2 at all:

>10.0.0 (Mar 30, 2020)
> As of AWX 10.0.0, the official AWX CLI no longer supports Python 2 (it requires at least Python 3.6) (https://github.com/ansible/awx/pull/6327)

https://github.com/ansible/awx/blob/devel/CHANGELOG.md

ryanpetrello (approved):

ryanpetrello:
@zedr I'm gonna run through a few tests on my end, and then I'll merge this.  Thanks for reporting it!

PR #7799, opened 2020-08-03
Adding import/export modules around AWX Kit features

john-westcott-iv:
##### SUMMARY

This PR adds a tower_import and tower_export modules to wrap the awxkit features.
It includes changes to the module_utils library structure:
   * Original TowerModule becomes TowerLegacyModule
   * TowerModule from tower_api becomes TowerAPIModule (and has some base stuff split out)
   * A new TowerAWXKitModule is created in tower_awxkit.py
   * A real base TowerModule is created in tower_module.py
       * TowerModule is responsible for loading config files and having the base connection argument spec, dealing with exit/warning, etc
       * TowerAWXKit and TowerAPI are responsible for interfacing with Tower
       * TowerAWXKitModule and TowerAPIModule are child classes of TowerModule

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - Collection

##### AWX VERSION

```
awx: 13.0.0
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx_collection/plugins/module_utils/tower_awxkit.py
+++ b/awx_collection/plugins/module_utils/tower_awxkit.py
@@ -1,0 +1,50 @@
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+from . tower_module import TowerModule
+from ansible.module_utils.basic import missing_required_lib
+
+try:
+    from awxkit.api.client import Connection
+    from awxkit.api.pages.api import ApiV2
+    from awxkit.api import get_registered_page
+    HAS_AWX_KIT = True
+except ImportError:
+    HAS_AWX_KIT = False
+
+
+class TowerAWXKitModule(TowerModule):
+    connection = None
+    apiV2Ref = None
+
+    def __init__(self, argument_spec, **kwargs):
+        kwargs['supports_check_mode'] = False
+
+        super(TowerAWXKitModule, self).__init__(argument_spec=argument_spec, **kwargs)
+
+        # Die if we don't have AWX_KIT installed
+        if not HAS_AWX_KIT:
+            self.exit_module(msg=missing_required_lib('awxkit'))
+
+        # Establish our conneciton object
+        self.connection = Connection(self.host, verify=self.verify_ssl)
+
+    def authenticate(self):
+        try:
+            self.connection.login(username=self.username, password=self.password, token=self.oauth_token)
+            # If we have neither of these, then we can try un-authenticated access
+            self.authenticated = True
+        except Exception:
+            self.exit_module("Failed to authenticate")
+
+    def get_api_v2_object(self):
+        if not self.apiV2Ref:
+            if not self.authenticated:
+                self.authenticate()
+            v2_index = get_registered_page('/api/v2/')(self.connection).get()
+            self.api_ref = ApiV2(connection=self.connection, **{'json': v2_index})
+        return self.api_ref
+
+    def logout(self):
+        if self.authenticated:
+            self.connection.logout()
--- a/awx_collection/plugins/module_utils/tower_awxkit.py
+++ b/awx_collection/plugins/module_utils/tower_awxkit.py
@@ -31,9 +31,12 @@ class TowerAWXKitModule(TowerModule):
 
     def authenticate(self):
         try:
-            self.connection.login(username=self.username, password=self.password, token=self.oauth_token)
-            # If we have neither of these, then we can try un-authenticated access
-            self.authenticated = True
+            if self.oauth_token:
+                self.connection.login(None, None, token=self.oauth_token, auth_type='Bearer')
+                self.authenticated = True
+            elif self.username:
+                self.connection.login(username=self.username, password=self.password)
+                self.authenticated = True
         except Exception:
             self.exit_json("Failed to authenticate")
 
--- a/awx_collection/plugins/module_utils/tower_awxkit.py
+++ b/awx_collection/plugins/module_utils/tower_awxkit.py
@@ -24,7 +24,7 @@ class TowerAWXKitModule(TowerModule):
 
         # Die if we don't have AWX_KIT installed
         if not HAS_AWX_KIT:
-            self.exit_module(msg=missing_required_lib('awxkit'))
+            self.exit_json(msg=missing_required_lib('awxkit'))
 
         # Establish our conneciton object
         self.connection = Connection(self.host, verify=self.verify_ssl)
@@ -35,7 +35,7 @@ class TowerAWXKitModule(TowerModule):
             # If we have neither of these, then we can try un-authenticated access
             self.authenticated = True
         except Exception:
-            self.exit_module("Failed to authenticate")
+            self.exit_json("Failed to authenticate")
 
     def get_api_v2_object(self):
         if not self.apiV2Ref:
@@ -48,3 +48,5 @@ class TowerAWXKitModule(TowerModule):
     def logout(self):
         if self.authenticated:
             self.connection.logout()
+
+    
--- a/awx_collection/plugins/module_utils/tower_awxkit.py
+++ b/awx_collection/plugins/module_utils/tower_awxkit.py
@@ -32,7 +32,7 @@ class TowerAWXKitModule(TowerModule):
     def authenticate(self):
         try:
             if self.oauth_token:
-                self.connection.login(None, None, token=self.oauth_token, auth_type='Bearer')
+                self.connection.login(None, None, token=self.oauth_token)
                 self.authenticated = True
             elif self.username:
                 self.connection.login(username=self.username, password=self.password)
--- a/awx_collection/plugins/module_utils/tower_awxkit.py
+++ b/awx_collection/plugins/module_utils/tower_awxkit.py
@@ -48,5 +48,3 @@ class TowerAWXKitModule(TowerModule):
     def logout(self):
         if self.authenticated:
             self.connection.logout()
-
-    
--- a/awx_collection/plugins/module_utils/tower_module.py
+++ b/awx_collection/plugins/module_utils/tower_module.py
@@ -215,7 +215,6 @@ class TowerModule(AnsibleModule):
                 else:
                     setattr(self, honorred_setting, config_data[honorred_setting])
 
-
     def logout(self):
         # This method is intended to be overridden
         pass
--- a/awx_collection/plugins/module_utils/tower_module.py
+++ b/awx_collection/plugins/module_utils/tower_module.py
@@ -1,0 +1,240 @@
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+from ansible.module_utils.basic import AnsibleModule, env_fallback
+from ansible.module_utils.six import PY2, string_types
+from ansible.module_utils.six.moves import StringIO
+from ansible.module_utils.six.moves.urllib.parse import urlparse
+from ansible.module_utils.six.moves.configparser import ConfigParser, NoOptionError
+from socket import gethostbyname
+import re
+from os.path import isfile, expanduser, split, join, exists, isdir
+from os import access, R_OK, getcwd
+from distutils.util import strtobool
+
+try:
+    import yaml
+    HAS_YAML = True
+except ImportError:
+    HAS_YAML = False
+
+
+class ConfigFileException(Exception):
+    pass
+
+
+class ItemNotDefined(Exception):
+    pass
+
+
+class TowerModule(AnsibleModule):
+    url = None
+    AUTH_ARGSPEC = dict(
+        tower_host=dict(required=False, fallback=(env_fallback, ['TOWER_HOST'])),
+        tower_username=dict(required=False, fallback=(env_fallback, ['TOWER_USERNAME'])),
+        tower_password=dict(no_log=True, required=False, fallback=(env_fallback, ['TOWER_PASSWORD'])),
+        validate_certs=dict(type='bool', aliases=['tower_verify_ssl'], required=False, fallback=(env_fallback, ['TOWER_VERIFY_SSL'])),
+        tower_oauthtoken=dict(type='raw', no_log=True, required=False, fallback=(env_fallback, ['TOWER_OAUTH_TOKEN'])),
+        tower_config_file=dict(type='path', required=False, default=None),
+    )
+    short_params = {
+        'host': 'tower_host',
+        'username': 'tower_username',
+        'password': 'tower_password',
+        'verify_ssl': 'validate_certs',
+        'oauth_token': 'tower_oauthtoken',
+    }
+    host = '127.0.0.1'
+    username = None
+    password = None
+    verify_ssl = True
+    oauth_token = None
+    oauth_token_id = None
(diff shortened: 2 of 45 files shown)


awx_collection/plugins/module_utils/tower_awxkit.py
```diff
@@ -0,0 +1,50 @@
...
+            self.exit_module(msg=missing_required_lib('awxkit'))
+
+        # Establish our conneciton object
+        self.connection = Connection(self.host, verify=self.verify_ssl)
+
+    def authenticate(self):
+        try:
+            self.connection.login(username=self.username, password=self.password, token=self.oauth_token)
+            # If we have neither of these, then we can try un-authenticated access
+            self.authenticated = True
```
AlanCoding: If you don't have username/pass or token, then wouldn't that error and put you in the `except` block?

john-westcott-iv: Those should all have a default of None from the TowerModule code:
```
    username = None
    password = None
    verify_ssl = True
    oauth_token = None
```

So they should never be "not defined" and cause an error.

awx_collection/plugins/module_utils/tower_module.py
```diff
@@ -0,0 +1,240 @@
...
+
+class ConfigFileException(Exception):
+    pass
+
+
+class ItemNotDefined(Exception):
+    pass
+
+
+class TowerModule(AnsibleModule):
```
AlanCoding, line 30: It doesn't make a whole lot of sense to me that the common `TowerModule` would subclass `AnsibleModule`. Because of this, we still maintain the hack of the `direct_params` in `__init__`. I thought getting rid of that would kind of be the whole point of separating the classes.

The way it is now, it looks like you could have avoided moving all this code and just had the towerkit class subclass the old class.

But maybe this is your objective and you just haven't gotten around to everything yet.

awx_collection/plugins/module_utils/tower_module.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def logout(self):
+        # This method is intended to be overridden
+        pass
+
+    def fail_json(self, **kwargs):
+        # Try to log out if we are authenticated
+        self.logout()
+        if self.error_callback:
+            self.error_callback(**kwargs)
```
AlanCoding, line 226: This is another hack which I thought separating the classes would help solve. The `TowerAPIModule` class could just pass `self.fail_json` as the error callback, instead of having `self.fail_json` be the default behavior, which isn't valid for non-module plugins.

awx_collection/plugins/modules/tower_export.py
```diff
@@ -0,0 +1,155 @@
...
+
+def main():
+    argument_spec = dict(
+        all=dict(type='bool', default=False),
+    )
+
+    # We are not going to raise an error here because the __init__ method of TowerAWXKitModule will do that for us
+    if HAS_EXPORTABLE_RESOURCES:
+        for resource in EXPORTABLE_RESOURCES:
+            argument_spec[resource] = dict()
```
AlanCoding: 👍  clever way to avoid writing more code for the other parameters.

awx_collection/tests/integration/targets/tower_export/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
...
+    - name: Export all tower assets
+      tower_export:
+        all: True
+      register: all_assets
+
+    - assert:
+        that:
+          - all_assets is not changed
+          - all_assets is successful
+          - all_assets['assets']['organizations'] | length() >= 2
```
AlanCoding, line 35: 👍  I'm fine with this structure for tests. We need to confirm that they work, but this is what I expected.

awx_collection/plugins/modules/tower_export.py
```diff
@@ -0,0 +1,155 @@
...
+    module.json_output['changed'] = False
+
+    # The exporter code currently works like the following:
+    #   Empty list == all assets of that type
+    #   string = just one asset of that type (by name)
+    #   None = skip asset type
+    # Here we are going to setup a dict of values to export
+    export_args = {}
+    for resource in EXPORTABLE_RESOURCES:
+        if module.params.get('all') or module.params.get(resource) == 'all':
```
AlanCoding, line 131: Seems like we should add some really simple validation for the case that 

```python
module.params.get('all') and module.params.get(resource) not in (None, 'all')
```

In that case, you're getting conflicting instructions from the user. The "all" parameter dictates that you should export all credentials, but the credential param says to export just "my-cred", for example.

awx_collection/tests/integration/targets/tower_export/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
...
+        inventory: 'all'
+        organizations: "{{ org_name1 }}"
+      register: mixed_export
+
+    - assert:
+        that:
+          - mixed_export is successful
+          - mixed_export is not changed
+          - mixed_export['assets']['inventory'] | length() >= 1
+          - mixed_export['assets']['organizations'] | length() == 1
```
AlanCoding, line 61: 👍  good to have coverage of this.

awx_collection/plugins/modules/tower_host.py
```diff
@@ -89,7 +89,7 @@ def main():
     )
 
     # Create a module for ourselves
-    module = TowerModule(argument_spec=argument_spec)
+    module = TowerAPIModule(argument_spec=argument_spec)
```
jbradberry, line 92: The new import/export code does not support hosts or groups, though there is an open issue to add them in.

awx_collection/tests/integration/targets/tower_import/tasks/main.yml
```diff
@@ -0,0 +1,108 @@
...
+              natural_key:
+                name: "Default"
+                type: "organization"
+      register: import_output
+
+    - assert:
+        that:
+          - import_output is changed
+
+    - name: "Import something again (awxkit is not idempotent, this tests a filure)"
```
jbradberry: filure -> failure

awx_collection/plugins/modules/tower_export.py
```diff
@@ -0,0 +1,155 @@
...
+    if not HAS_EXPORTABLE_RESOURCES:
+        module.fail_json(msg="Your version of awxkit does not have import/export")
+
+    # The export process will never change a Tower system
+    module.json_output['changed'] = False
+
+    # The exporter code currently works like the following:
+    #   Empty list == all assets of that type
+    #   string = just one asset of that type (by name)
+    #   None = skip asset type
```
jbradberry: This is not correct.  The actual behavior is:

empty string: all assets of this type
non-empty string: one specific asset by name or id
asset type not present in dict or is `None`: skip unless the export kwargs was missing all asset types, then export all instances of all types

awx_collection/plugins/modules/tower_export.py
jbradberry: Should we consider moving away from the use of `'all'` flags?  The actual semantics of the import/export code that I wrote is unambiguous, the case that @AlanCoding is talking about here cannot arise.

awx_collection/plugins/module_utils/tower_module.py
john-westcott-iv: Correct, there is a plan in place but it will not be part of this PR.

awx_collection/plugins/modules/tower_export.py
john-westcott-iv: I'm open to input but I think I am partial to the explicit `all` parameter and keywords. We could make it do something like:
```
- name: Export all resources
   tower_export:
   register: all_exported_data
```

And then have something like your code:
```
        # If no resource parameters are explicitly used, export everything.
        all_resources = all(module.params.get(resource) is None for resource in EXPORTABLE_RESOURCES)
```

But then we are back to our existing loop, just changing `module.params.get('all')` to `all_resources`.

I think I am also partial to the explicit:
```
- name: Export all credentials
  tower_export:
    credentials: 'all'
```

as opposed to :
```
- name: Export all credentials
  tower_export:
    credentials: ''
```

The second syntax looks like it could be confused with "nothing" vs an all. Anyone else have thoughts on this?

awx_collection/plugins/modules/tower_export.py
john-westcott-iv: Comment updated and [] changed to ''.

awx_collection/plugins/modules/tower_export.py
AlanCoding: Hmm, as a gut reaction I personally like the aesthetics of this

```yaml
- name: Export all resources
   tower_export:
   register: all_exported_data
```

But the behavior is "jumpy". If you have `credentials: "foo"` then it exports 1 thing, but take it away and it exports 1,000 things. If this syntax returned _nothing_ that would give a better impression of behavioral stability. I'm just saying it's arguable so I don't want to take a position.

I don't have a problem with the current "all" param. The validation statement I'm talking about would just be 1 conditional.

I also don't want to care about giving blank values. Ansible has an `omit` syntax. You could throw an error for blank values for clarity, but it doesn't bother me either way, it shouldn't be your problem to deal with.

AlanCoding:
👍  looking so far. I'm kicking off some tests, once those are passing, then I'm going to be pushing to get this ready for merge.

awx_collection/plugins/modules/tower_import.py
```diff
@@ -0,0 +1,99 @@
...
+      type: dict
+requirements:
+  - "awxkit >= 9.3.0"
+extends_documentation_fragment: awx.awx.auth
+'''
+
+EXAMPLES = '''
+- name: Import all tower assets
+  tower_import:
+    assets: "{{ export_output.assets }}"
```
AlanCoding, line 45: you should spell it out for people that `export_output` is from a `register:` from `tower_export`.

awx_collection/plugins/modules/tower_import.py
```diff
@@ -0,0 +1,99 @@
...
+
+- name: Import orgs from a json file
+  tower_import:
+    assets: "{{ lookup('file', 'org.json') | from_json() }}"
```
AlanCoding, line 49: Again, what is the origin of the file? Is it from the `awx export` command?

awx_collection/plugins/modules/tower_import.py
```diff
@@ -0,0 +1,99 @@
...
+    try:
+        module.json_output['changed'] = module.get_api_v2_object().import_assets(assets)
+    except Exception as e:
+        module.fail_json(msg="Failed to import assets {0}".format(e))
+    finally:
+        # Finally consume the logs incase there were any errors and die if there were
+        log_contents = log_capture_string.getvalue()
+        log_capture_string.close()
+        if log_contents != '':
+            module.fail_json(msg=log_contents)
```
AlanCoding, line 99: this seems a little extreme... wouldn't you just want to warn with the log contents? Wouldn't the CLI code raise an exception if something was actually a problem? Asking @jbradberry here too

jbradberry: I think I agree, this seems a bit much.

awx_collection/tests/integration/targets/tower_export/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
...
+
+    - name: Create an inventory
+      tower_inventory:
+        name: "{{ inventory_name1 }}"
+        organization: "{{ org_name1 }}"
+
+    - name: Export all tower assets
+      tower_export:
+        all: true
+      register: all_assets
```
AlanCoding, line 29: The kind of failure I'm seeing for this is:

```
  TASK [tower_export : Export all tower assets] **********************************
  ok: [testhost]
  
  TASK [tower_export : assert] ***************************************************
  fatal: [testhost]: FAILED! => {"msg": "The conditional check 'all_assets['assets']['organizations'] | length() >= 2' failed. The error was: error while evaluating conditional (all_assets['assets']['organizations'] | length() >= 2): 'dict object' has no attribute 'assets'"}
```

It is possible that this failure is due to missing the `awxkit` library.

Is there any logic for detection of importability of `awxkit`? I would like to have a minimal check for that and give an error message that makes sense.

squidboylan: unrelated but im a bit worried about exporting everything in our tests, this could add a significant amount of time to our testing, we might want to have a discussion about how we can export less while feeling we have enough coverage

john-westcott-iv: @AlanCoding There is logic in both tower_awxkit.py:
```
        if not HAS_AWX_KIT:
            self.exit_json(msg=missing_required_lib('awxkit'))
```

As well as more specific logic in tower_import/export:
```
    if not HAS_EXPORTABLE_RESOURCES:
        module.fail_json(msg="Your version of awxkit does not have import/export")
```

I'm not sure what would cause an `ok` with no `assets` entry, can you have it dump out the all_assets variable?

awx_collection/tests/integration/targets/tower_export/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
...
+
+    - assert:
+        that:
+          - all_assets is not changed
+          - all_assets is successful
+          - all_assets['assets']['organizations'] | length() >= 2
```
beeankha, line 35: When I debug `all_assets`, I get:
```
TASK [debug] ******************************************************************************************************************************************
ok: [localhost] => {
    "all_assets": {
        "assets": {
            "credential_types": null,
            "credentials": null,
            "inventory": null,
            "inventory_sources": null,
            "job_templates": null,
            "notification_templates": null,
            "organizations": null,
            "projects": null,
            "teams": null,
            "users": null,
            "workflow_job_templates": null
        },
        "changed": false,
        "failed": false
    }
}
```

...and the test result is:
```
TASK [assert] **************************************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "The conditional check 'all_assets['assets']['organizations'] | length() >= 2' failed. The error was: Unexpected templating type error occurred on ({% if all_assets['assets']['organizations'] | length() >= 2 %} True {% else %} False {% endif %}): object of type 'NoneType' has no len()"}
```

awx_collection/tests/integration/targets/tower_import/tasks/main.yml
```diff
@@ -0,0 +1,108 @@
...
+                notification_templates_error: []
+                notification_templates_approvals: []
+              natural_key:
+                name: "Default"
+                type: "organization"
+      register: import_output
+
+    - assert:
+        that:
+          - import_output is changed
```
beeankha, line 34: Here I'm seeing:
```
TASK [Import something] ****************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to import assets argument of type 'NoneType' is not iterable"}
```

beeankha (approved):

AlanCoding:
recheck

AlanCoding (approved):

beeankha:
recheck

Issue #7798, opened 2020-08-03
Exported Workflow Job Templates refer to inventories by id instead of natural key 


zedr:
##### ISSUE TYPE
 - The API endpoint for a workflow_job_template  does not provide a representation that includes the natural key of the related Inventory, which is problematic when migrating it between different instances of AWX/Tower.

##### SUMMARY
The API endpoint for a workflow_job_template  does not provide a representation that includes the natural key of the related Inventory, if set. Instead, it just uses the numerical `id`, which is problematic when attempting to export and import the workflow between different instances of AWX/Tower, since the API will attempt to use the given numerical `id` to associate an inventory to the import workflow, which will most likely be different on the target instance of AWX/Tower. This behaviour is dangerous because if the target instance of AWX/Tower has a different inventory with the same numerical `id`, no warnings will be raised, causing the workflow to be associated with the wrong template.

##### ENVIRONMENT
* AWX version: 3.7.1 (Tower)
* AWX install method: openshift
* Ansible version:  2.9.9
* Operating System: OpenShift
* Web Browser: Google Chrome

##### STEPS TO REPRODUCE

1. Export a `workflow_job_template` that is associated with an inventory
2. Observe the value of the `workflow_job_templates.[].inventory` attribut in the exported YAML.

##### EXPECTED RESULTS

It is reasonable to expect that the following YAML document would be generated for an example workflow_job_template object.

```yaml
workflow_job_templates:
- allow_simultaneous: true
  ask_inventory_on_launch: false
  ask_limit_on_launch: false
  ask_scm_branch_on_launch: true
  ask_variables_on_launch: true
  description: My VM provisioning
  extra_vars: ''
  inventory:
    name: my-vms
    type: inventory
    organization:
       name: ACME
       type: organization
  limit: null
  name: my-workflow
  natural_key:
    name: my-workflow
    organization:
      name: ACME
      type: organization
    type: workflow_job_template
  organization:
    name: ACME
    type: organization
  related:
    notification_templates_approvals: []
    notification_templates_error: []
    notification_templates_started: []
    notification_templates_success: []
    schedules: []
    survey_spec: {}
    workflow_nodes: []
  scm_branch: null
  survey_enabled: false
  webhook_credential: null
```
The document above, when imported, will generate the association with the inventory correctly on Tower 3.7.1. This is the expected behaviour, relying on the natural key of the object instead of the `id`.

##### ACTUAL RESULTS

YAML such as the following is generated:
```yaml
[...]

PR #7805, opened 2020-08-03
Serialize Workflow Job Template inventories by natural key - related #7798

zedr:
##### SUMMARY
Related #7798

This changeset introduces two changes:

 1. Update the API representation of Workflow Job Templates to use the
 natural key of the Inventory type instead of its id;
 2. Override the related property of the CLI's WorkflowJobTemplate page
 type to patch the related references during the export process,
 allowing the resource to be serialised using the natural key of the
 Inventory type instead of the id.

Change n.2 is a workaround that is used when exporting resources from
AWX/Tower instances that don't have change n.1. It can be removed in the
future.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API
 - CLI

##### AWX VERSION
```
awx: 13.0.0
```

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -3404,6 +3404,12 @@ class WorkflowJobTemplateSerializer(JobTemplateMixin, LabelsListMixin, UnifiedJo
             res['organization'] = self.reverse('api:organization_detail',   kwargs={'pk': obj.organization.pk})
         if obj.webhook_credential_id:
             res['webhook_credential'] = self.reverse('api:credential_detail', kwargs={'pk': obj.webhook_credential_id})
+        if obj.inventory:
+            res['inventory'] = self.reverse(
+                'api:inventory_detail', kwargs={
+                    'pk': obj.inventory.pk
+                }
+            )
         return res
 
     def validate_extra_vars(self, value):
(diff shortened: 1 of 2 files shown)


awx/api/serializers.py
```diff
@@ -3404,6 +3404,12 @@ def get_related(self, obj):
             res['organization'] = self.reverse('api:organization_detail',   kwargs={'pk': obj.organization.pk})
         if obj.webhook_credential_id:
             res['webhook_credential'] = self.reverse('api:credential_detail', kwargs={'pk': obj.webhook_credential_id})
+        if obj.inventory:
```
jbradberry, line 3407: :+1:

awxkit/awxkit/api/pages/workflow_job_templates.py
```diff
@@ -15,6 +15,27 @@ class WorkflowJobTemplate(HasCopy, HasCreate, HasNotifications, HasSurvey, Unifi
     optional_dependencies = [Organization]
     NATURAL_KEY = ('organization', 'name')
 
+    @property
+    def related(self):
```
jbradberry, line 19: I'm pretty uncomfortable with doing it this way.  However, fortunately, I do note that the `inventory` field shows up as of type `'id'` in the `OPTIONS` call on the api endpoint even under current devel awx.  I think it would be easy enough to check for the case of an id field that fails to show up in related in the `_export()` method, and attempt to construct the natural key by other means (probably by doing a `_filtered_list()` call).

zedr: OK. Good suggestion. I'll look into it and push an update if I get something working.

jbradberry:
I've gone ahead and rebased this branch against the latest devel, and made the change I recommended.  The new PR is #8353.

PR #7808, opened 2020-08-04
Allow YAML as a CLI import format

zedr:
##### SUMMARY
This changset allows the import of YAML formatted resources. The CLI
user can indicate which format to use with the `-f, --format` option.
The CLI help text has been amended to reflect the new feature.
    
The AWX CLI `export` subcommand offers the option of formatting the output
as YAML or JSON, so it makes sense that the `import` subcommand reflects
this.

A simple test is also provided. In order to ease the task of testing
commands that import resources by reading the stdin, the CLI has been
extended to allow specifying an alternative file descriptor for stdin,
similarly to stdout and stderr.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
- CLI

##### AWX VERSION
```
awx: 13.0.0
```

### DIFF
--- a/awxkit/awxkit/cli/resource.py
+++ b/awxkit/awxkit/cli/resource.py
@@ -1,8 +1,9 @@
+import yaml
 import json
 import os
-import sys
 
 from awxkit import api, config
+from awxkit.exceptions import ImportExportError
 from awxkit.utils import to_str
 from awxkit.api.pages import Page
 from awxkit.api.pages.api import EXPORTABLE_RESOURCES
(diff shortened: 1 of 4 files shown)


ryanpetrello:
@jbradberry @elyezer this seems like a pretty reasonable ask - any thoughts on this?

jbradberry:
This seems like a reasonable idea to me, and I don't see any problems offhand with the code here.

jbradberry (approved):

awxkit/awxkit/cli/resource.py
```diff
@@ -135,7 +136,13 @@ def handle(self, client, parser):
             parser.print_help()
             raise SystemExit()
 
-        data = json.load(sys.stdin)
+        format = getattr(client.args, 'conf.format')
```
ryanpetrello: Is there a reason you didn't just do `client.get_config('format')` ?

zedr: I simply wasn't aware of it :smile:

ryanpetrello (approved):

PR #7811, opened 2020-08-04
mark PRIMARY_GALAXY_USERNAME and PRIMARY_GALAXY_PASSWORD as deprecated

ryanpetrello:


awx/main/conf.py
```diff
@@ -458,7 +458,8 @@ def _load_default_license_from_file():
     required=False,
     allow_blank=True,
     label=_('Primary Galaxy Server Username'),
-    help_text=_('For using a galaxy server at higher precedence than the public Ansible Galaxy. '
+    help_text=_('This setting is deprecated and will be removed in a future release. '
```
beeankha: I know this is nitpicky, but I feel like having `[]` or `()` around the deprecation notice makes the docstring "flow" a bit better... Otherwise it kinda blends in with the rest.

beeankha (approved):
Non-blocking comment added!

ryanpetrello:
@beeankha 
image

ryanpetrello:
cc @tvo318

AlanCoding (approved):

Issue #7813, opened 2020-08-04
Improve support for fetching content from multiple Galaxy-like endpoints


ryanpetrello:
Today, AWX users can specify Ansible role/collection dependencies in their SCM repositories in a `requirements.yml` file, and AWX will install these as part of the Project Update.  Currently, AWX has support for pulling this content from https://galaxy.ansible.org or *one* Galaxy-like API such as Red Hat's cloud-based Ansible Automation Hub.

The model for configuring Galaxy credentials today is too simple (you can only specify *one* "private" Galaxy, ordering is complicated, and it's a set of *global* settings that can't be differentiated per Organization).

This issue is meant to track a change to this interface whereby Galaxy credentials will be assigned to Organizations, and can be unlimited and ordered as the user sees fit.

### Scope of API Change

- [x]  Introduce a new CredentialType of e.g., `kind="galaxy"` which represents a content source for fetching collections defined in `requirements.yml` when project updates are run (e.g., galaxy.ansible.com, cloud.redhat.com, on-premise Automation Hub).  This new type will represent a URL and (optional) authentication details necessary to construct the environment variables _when a project update runs_ `ansible-galaxy collection install` as described in https://docs.ansible.com/ansible/latest/user_guide/collections_using.html#configuring-the-ansible-galaxy-client. It will have fields which map directly to the configuration options exposed to the Ansible Galaxy CLI, e.g., per-server:
    - `url` (required str)
    - `auth_url` (optional str)
    - `token` (optional str, encrypted)

Note that a URL will *always* be required, but authentication (as in the case of galaxy.ansible.com) is optional. 

**This new solution _will not_ support basic authentication, because several of these APIs will be removing support for basic auth in the next year** (see: https://github.com/ansible/awx/pull/7811/)

- [x]  Add an API endpoint for associating an _ordered list_ of these credentials _at the Organization level_.  For example, an Organization may have an empty list (the default), or one or more items in the list (i.e., Public Galaxy, _then_ On-Prem AH).  This might be something like:

`/api/v2/organizations/N/galaxy_credentials/`

- [x]  Write code to detect and migrate the *existing* Galaxy-oriented setting values in such a way that post-upgrade, proper credentials are created and attached to _every_ Organization in the AWX install:
    - `PRIMARY_GALAXY_URL`
    - `PRIMARY_GALAXY_USERNAME`
    - `PRIMARY_GALAXY_PASSWORD`
    - `PRIMARY_GALAXY_TOKEN`
[...]

PR #7817, opened 2020-08-04
Support Organization-scoped Galaxy installs

ryanpetrello:
see: https://github.com/ansible/awx/issues/7813

cc @cutwater @bmbouter

```shell
curl -sk "https://awx.example.org/api/v2/organizations/1/galaxy_credentials/" | jq '.results[] | .inputs'
{
  "url": "https://cloud.redhat.com/api/automation-hub/",
  "token": "$encrypted$",
  "auth_url": "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token"
}
{
  "url": "https://galaxy.ansible.com/"
}
```

image

image


awx/main/redact.py
```diff
@@ -13,11 +11,13 @@ class UriCleaner(object):
     @staticmethod
     def remove_sensitive(cleartext):
         # exclude_list contains the items that will _not_ be redacted
-        exclude_list = [settings.PUBLIC_GALAXY_SERVER['url']]
-        if settings.PRIMARY_GALAXY_URL:
-            exclude_list += [settings.PRIMARY_GALAXY_URL]
-        if settings.FALLBACK_GALAXY_SERVERS:
-            exclude_list += [server['url'] for server in settings.FALLBACK_GALAXY_SERVERS]
+        # TODO: replace this with the server URLs from proj.org.credentials
```
ryanpetrello: I still need to refactor this.

awx/settings/defaults.py
```diff
@@ -569,28 +569,9 @@ def IS_TESTING(argv=None):
...
-    'url': 'https://galaxy.ansible.com'
-}
-
 # Applies to any galaxy server
 GALAXY_IGNORE_CERTS = False
 
-# List of dicts of fallback (additional) Galaxy servers.  If configured, these
-# will be higher precedence than public Galaxy, but lower than primary Galaxy.
-# Available options: 'id', 'url', 'username', 'password', 'token', 'auth_url'
-FALLBACK_GALAXY_SERVERS = []
```
ryanpetrello, line 590: @wenottingham @AlanCoding I decided not to put this into the migration; I can't see where we ever actually documented this as a feature (it's not exposed in the API, either), and given the landscape of "Galaxy-like services" _before now_, I'd argue we can probably assume there's not (m)any people who've done this (unless you two know of any).

awx/ui/client/src/configuration/forms/jobs-form/configuration-jobs.form.js
```diff
@@ -74,31 +74,6 @@ export default ['i18n', function(i18n) {
             AWX_SHOW_PLAYBOOK_LINKS: {
                 type: 'toggleSwitch',
             },
-            PRIMARY_GALAXY_URL: {
```
ryanpetrello, line 77: Look, @mabashian, I know JavaScript!

awx/main/tasks.py
```diff
@@ -2046,35 +2046,24 @@ def build_env(self, project_update, private_data_dir, isolated=False, private_da
...
-                    if not server.get(key):
-                        continue
-                    env_key = ('ANSIBLE_GALAXY_SERVER_{}_{}'.format(server.get('id', 'unnamed'), key)).upper()
-                    env[env_key] = server[key]
-            if galaxy_servers:
-                # now set the precedence of galaxy servers
-                env['ANSIBLE_GALAXY_SERVER_LIST'] = ','.join([server.get('id', 'unnamed') for server in galaxy_servers])
+
+        # build out env vars for Galaxy credentials (in order)
+        galaxy_server_list = []
```
ryanpetrello, line 2032: @wenottingham @rebeccahhh @chrismeyersfsu @AlanCoding one behavioral change worth pointing out here is that *for new installs* and *new Organizations* (ignore old settings that get migrated on upgrade), the default behavior is "don't pull from galaxy.ansible.com by default".  I _think_ this is what I heard @wenottingham say he wanted - is that right?

awx/main/redact.py
ryanpetrello: Actually, @AlanCoding I don't think we actually need this anymore at all - the point of this code is to filter HTTP basic auth creds out of URLs - starting w/ this PR we're *only* supporting tokens, so I don't actually think we need this anymore.

awx/main/redact.py
```diff
@@ -12,23 +10,13 @@ class UriCleaner(object):
 
     @staticmethod
     def remove_sensitive(cleartext):
-        # exclude_list contains the items that will _not_ be redacted
```
ryanpetrello, line 15: There's not really anything to redact here (we don't support basic auth anymore).

Also, all our credentialing data is passed in via environment variables now (a token).

Here's some stdout that illustrates that these URLs render just fine without this code:

image

ryanpetrello: I'm a little unclear why were doing this in the first place, @AlanCoding:

https://github.com/ansible/awx/commit/d59d8562db35a8b1d7769cc8081c84028779c931

The point of this function is to take a giant blob of stdout and discover embedded URLs (printed by e.g., `git`) like:

`https://USER:PASS@foo`

...and turn them into `https://*********@foo`

Why would we ever want to _whitelist_ Galaxy/Hub URLs to _prevent_ that behavior?  I don't think we need this code at all anymore since we're losing basic auth support, but I think it was flawed to start with.

AlanCoding: If it was important we would have test coverage for it. Maybe I was going crazy, I'm not sure.

awx/main/migrations/_galaxy.py
```diff
@@ -0,0 +1,85 @@
...
+    Credential = apps.get_model('main', 'Credential')
+    Setting = apps.get_model('conf', 'Setting')
+
+    galaxy_type = CredentialType.objects.get(kind='galaxy')
+    private_galaxy_url = Setting.objects.filter(key='PRIMARY_GALAXY_URL').first()
+
+    # by default, prior versions of AWX/Tower automatically pulled content
+    # from galaxy.ansible.com
+    public_galaxy_enabled = True
+    public_galaxy_setting = Setting.objects.filter(key='PUBLIC_GALAXY_ENABLED').first()
```
AlanCoding, line 33: 👍  nice, that's how I would have done this inside of migrations.

awx/main/migrations/_galaxy.py
```diff
@@ -0,0 +1,85 @@
...
+        if public_galaxy_enabled:
+            # If public Galaxy was enabled, make a credential for it
+            cred = Credential(
+                created=now(),
+                modified=now(),
+                name='Ansible Galaxy',
+                organization=org,
+                credential_type=galaxy_type,
+                inputs = {
+                    'url': 'https://galaxy.ansible.com/'
```
AlanCoding: I'm a little confused by this in particular. What happens if you create a new organization after this migration has finished? What credentials does it have associated, and how does it go about doing Galaxy commands before the user has associated any credentials?

Also, why would each organization get a _new_ public galaxy "credential"? Wouldn't all orgs just share the same credential? It's 1 URL.

awx/main/migrations/_galaxy.py
```diff
@@ -0,0 +1,85 @@
...
+                name = f'Ansible Automation Hub ({private_galaxy_url.value})'
+            cred = Credential(
+                created=now(),
+                modified=now(),
+                name=name,
+                organization=org,
+                credential_type=galaxy_type,
+                inputs=inputs
+            )
+            cred.save()
```
AlanCoding, line 82: I am going to have a lot of RBAC questions that stem from this. Before we even get into any feature-specific discussions, does this correctly populate the credential's role fields and their parents/ancestors? I don't trust that it does in migrations.

You made the organization of the credential the same as the organization it gets associated. Do you intend to enforce this for new credentials?

awx/main/migrations/_galaxy.py
AlanCoding: Also, shouldn't the public Galaxy credential be visible to all users? In this implementation, it seems like it would only be visible to superusers.

ryanpetrello: > I'm a little confused by this in particular. What happens if you create a new organization after this migration has finished? What credentials does it have associated...

None.  We're changing the default behavior when you create new Organizations to *not* pull unofficial community content (cc @wenottingham in case I've misunderstood this).

ryanpetrello: > Also, why would each organization get a new public galaxy "credential"? Wouldn't all orgs just share the same credential? It's 1 URL.

That _might_ be a better way to do this.  Who's allowed to edit that one global credential, though?  How would we even represent that with our current RBAC model?

awx/main/migrations/_galaxy.py
ryanpetrello: Good questions - this is an area where our usage of a credential to represent "pull from galaxy.ansible.com" here is a _little_ weird.  It doesn't really make sense to create one of these credentials outside of the context of its association with at least one Organization (i.e.., there's no reason for somebody lower than Org Admin to make one of these, because they can't actually assign it to an Org).

awx/main/migrations/_galaxy.py
wenottingham: How would each org be able to use this credential without weird backend logic on org creation?

awx/main/tasks.py
wenottingham: We'd need to document this, but I'm OK with it

awx/settings/defaults.py
wenottingham: Depends on what folks like GTS are doing. But I think the new way of setting this up is more clear.

docs/collections.md
```diff
@@ -85,28 +85,22 @@ job runs. It will populate the cache id set by the last "check" type update.
...
-primary galaxy server if they are available there. If a requirement is
-not available from the primary galaxy server, then it will fallback to
-downloading it from the public Galaxy server.
+You can specify a list of zero or more servers to download roles and
+collections from for AWX Project Updates.  This is done by associating Galaxy
+credentials (in sequential order) via the API at
+`/api/v2/organizations/N/galaxy_credentials/`.  Authentication
+via an API token is optional (i.e., https://galaxy.ansible.com/), but other
+content sources (such as Red Hat Ansible Automation Hub) require proper
+configuration of the Auth URL and Token.
```
wenottingham, line 99: I think onprem hub can do token only w/o auth URL.

awx/main/migrations/_galaxy.py
ryanpetrello: > How would each org be able to use this credential without weird backend logic on org creation?

Yea.  I'll admit that in the case of "Public Galaxy", creating a cred per org is a little bit of a square peg in a round hole here, but I don't think I want to do something super crazy to accommodate sharing *one* Credential object globally.

docs/collections.md
ryanpetrello: My understanding is cloud.redhat.com is token + Auth URL (basically, a keycloak server), and that "on premise Automation Hub" is a URL and a token.  In any event, the permutation of how you specify all of this is different for each implementation, which is why everything but the "URL" is optional.

cc @chrismeyersfsu @rebeccahhh do I have this right?

awx/main/migrations/_galaxy.py
ryanpetrello: >  does this correctly populate the credential's role fields and their parents/ancestors?

It _seems_ like it does?

```In [8]: Credential.objects.first().name
Out[8]: 'Ansible Galaxy'

In [9]: Credential.objects.first().inputs
Out[9]: {'url': 'https://galaxy.ansible.com/'}

In [10]: Credential.objects.first().admin_role.ancestors.all()
Out[10]: <QuerySet [<Role: Admin-36>, <Role: Admin-2>, <Role: Credential Admin-6>, <Role: System Administrator-1>]>

In [12]: Credential.objects.first().admin_role.parents.all()
Out[12]: <QuerySet [<Role: Credential Admin-6>, <Role: System Administrator-1>]>
```

awx/main/migrations/_galaxy.py
ryanpetrello: One potential solution - Galaxy creds must belong to an Org (not a team or user).  I think that's pretty reasonable, because there's no context you'd use them other than via their ordered assignment to an Organization (which is something an Org Admin and higher would have exclusively have access to).

image

awx/settings/defaults.py
ryanpetrello: Nevermind, I decided to just support it and write some tests.

awx/main/tests/functional/test_galaxy_credential_migration.py
```diff
@@ -0,0 +1,111 @@
...
+            'username': 'user',
+            'password': 'pass',
+            'token': 'fallback123',
+        }]
+        galaxy.migrate_galaxy_settings(apps, None)
+    finally:
+        settings.FALLBACK_GALAXY_SERVERS = []
+    assert org.galaxy_credentials.count() == 3
+    creds = org.galaxy_credentials.all()
+    assert creds[0].name == 'Private Galaxy (https://example.org/)'
```
ryanpetrello, line 107: cc @AlanCoding @wenottingham @rebeccahhh @chrismeyersfsu 

Highest precedence is "private Hub" aka `PRIMARY_GALAXY_XXXXXX`
...followed by anything in `FALLBACK_GALAXY_SERVERS`
...followed by galaxy.ansible.com **iff** `PUBLIC_GALAXY_ENABLED` is not explicitly set to False.

Yes?

awx/main/migrations/_galaxy.py
ryanpetrello: Another approach I like (which I've pushed to this PR) is to just introduce the concept of `Credential.managed_by_tower`, just like we did with `CredentialType.managed_by_tower`, and to have one global, non-editable Credential that represents galaxy.ansible.com

jakemcdermott: > What happens if you create a new organization after this migration has finished? What credentials does it have associated, and how does it go about doing Galaxy commands before the user has associated any credentials?

Could all new organizations start with a default credential in their Galaxy credential list? The default credential could point to `galaxy.ansible.com`.

jakemcdermott: Alternatively (& I think this might be preferable to what I mentioned above):

We could address this by showing a clear error / warning message if the user tries to run a JT that needs to fetch content without having configured a usable content credential that the job can use. There's a variety of places we could surface such a message (e.g job details page).

awx/ui/client/src/organizations/galaxy-credentials-multiselect/galaxy-credentials-modal/galaxy-credentials-modal.directive.js
```diff
@@ -0,0 +1,123 @@
...
+                        });
+                    });
+            }
+
+            init();
+
+            $scope.$on("selectedOrDeselected", function(e, value) {
+                let item = value.value;
+                if (value.isSelected) {
+                    if(!$scope.igTags) {
```
jakemcdermott: nitpick, but naming these `igTags` seems instance-group specific.

awx/ui/client/src/organizations/galaxy-credentials-multiselect/galaxy-credentials-modal/galaxy-credentials-modal.directive.js
```diff
@@ -0,0 +1,123 @@
...
+        },
+
+        controller: ['$scope', '$compile', 'QuerySet', 'GetBasePath','generateList', 'CredentialList', function($scope, $compile, qs, GetBasePath, GenerateList, CredentialList) {
+
+            function init() {
+
+                    $scope.credential_queryset = {
+                        order_by: 'name',
+                        page_size: 5,
+                        credential_type__kind: 'galaxy'
```
jakemcdermott, line 32: :+1:

jakemcdermott (commented):
@mabashian 3.8 ui lgtm. If we can get to it, I think it would be good to do the ui_next additions as well while the requirements and expected behaviors are fresh in everyone's mind.

@ryanpetrello The organization lists are always showing 0 results for me:

```shell
docker exec -i tools_awx_1 sh <<-EOSH
  awx-manage createsuperuser --noinput --username=admin --email=null@ansible.com
  awx-manage update_password --username=admin --password=password
EOSH
curl -f -k -X GET --user admin:password https://localhost:8043/api/v2/organizations/

{"count":0,"next":null,"previous":null,"results":[]}
```

awx/ui/client/src/organizations/galaxy-credentials-multiselect/galaxy-credentials-modal/galaxy-credentials-modal.directive.js
mabashian: Ha yea I copied this code over - will update

awx/main/migrations/_galaxy.py
ryanpetrello: > Could all new organizations start with a default credential in their Galaxy credential list? The default credential could point to galaxy.ansible.com.

What I've discussed with @wenottingham @AlanCoding and @mabashian is just this - we're going to make it so that the Org Creation form pre-fills with "Ansible Galaxy".

awx/main/tasks.py
rebeccahhh: duly noted :+1:

ryanpetrello: So this is the API behavior we landed on, but I think as a nicety in the UI, when you create a new org, we're talking about pre-filling the field to include Ansible Galaxy.  So at the time users create an org, they are immediately given the choice to opt in/out of "community Galaxy" at org creation time.  But if you're automating org creation (via e.g., the CLI or modules), you've got to explicit set these credentials up yourself to represent where you want to pull content from by default.

ryanpetrello: see: https://github.com/ansible/awx/pull/7817#discussion_r465950688

docs/collections.md
rebeccahhh: sounds right to me, on prem Automation Hub should be an API token and URL.
Also yes, it is true that each implementation is different and we do need those other fields to be optional.

awx/main/tests/functional/test_galaxy_credential_migration.py
rebeccahhh: not sure if we're using the terminology of `FALLBACK` but other than that yes, that all looks correct to me.
(and by not sure, I literally just don't know, it probably is :woman_shrugging:)

ryanpetrello: I debated not actually migrating these `FALLBACK_GALAXY_SERVERS` at all, because I'm not confident anybody actually uses them (they're not exposed in the API or documented anywhere that I can tell).

ryanpetrello:
recheck

elyezer (commented):
We will need to also update awxkit with the following, do you want handle it on this PR or want me to send a separate one?

```diff
diff --git a/awxkit/awxkit/api/pages/credentials.py b/awxkit/awxkit/api/pages/credentials.py
index d5ba0c4d05..f964ae38e4 100644
--- a/awxkit/awxkit/api/pages/credentials.py
+++ b/awxkit/awxkit/api/pages/credentials.py
@@ -82,6 +82,7 @@ def config_cred_from_kind(kind):
 
 credential_type_name_to_config_kind_map = {
     'amazon web services': 'aws',
+    'ansible galaxy/automation hub api token': 'galaxy',
     'ansible tower': 'tower',
     'google compute engine': 'gce',
     'insights': 'insights',
```

elyezer (commented):
We will need to also update awxkit with the following, do you want handle it on this PR or want me to send a separate one?

```diff
diff --git a/awxkit/awxkit/api/pages/credentials.py b/awxkit/awxkit/api/pages/credentials.py
index d5ba0c4d05..f964ae38e4 100644
--- a/awxkit/awxkit/api/pages/credentials.py
+++ b/awxkit/awxkit/api/pages/credentials.py
@@ -82,6 +82,7 @@ def config_cred_from_kind(kind):
 
 credential_type_name_to_config_kind_map = {
     'amazon web services': 'aws',
+    'ansible galaxy/automation hub api token': 'galaxy',
     'ansible tower': 'tower',
     'google compute engine': 'gce',
     'insights': 'insights',
diff --git a/awxkit/awxkit/api/pages/organizations.py b/awxkit/awxkit/api/pages/organizations.py
index 413ecf4961..e1d13a0013 100644
--- a/awxkit/awxkit/api/pages/organizations.py
+++ b/awxkit/awxkit/api/pages/organizations.py
@@ -22,6 +22,24 @@ class Organization(HasCreate, HasInstanceGroups, HasNotifications, base.Base):
         with suppress(exc.NoContent):
             self.related.users.post(user)
 
+    def add_galaxy_credential(self, credential):
+        if isinstance(credential, page.Page):
+            credential = credential.json
+        with suppress(exc.NoContent):
+            self.related.galaxy_credentials.post({
+                "id": credential.id,
+            })
+
+    def remove_galaxy_credential(self, credential):
+        if isinstance(credential, page.Page):
+            credential = credential.json
+        with suppress(exc.NoContent):
+            self.related.galaxy_credentials.post({
+                "id": credential.id,
+                "disassociate": True,
+            })
+
+
     def payload(self, **kwargs):
         payload = PseudoNamespace(name=kwargs.get('name') or 'Organization - {}'.format(random_title()),
                                   description=kwargs.get('description') or random_title(10))
```

ryanpetrello:
@elyezer feel free to add commits to this PR 😄

elyezer:
It seems that with these changes AWX is not keeping the behavior it had before when a galaxy server config was present on the `ansible.cfg` config file on the project's repository. Steps to reproduce:

1) Create a project that has a `ansible.cfg` with a invalid Galaxy url, for example https://github.com/ansible/test-playbooks/pull/129/files
2) Sync the project and the expected behavior would be to fail or error but it is succeeding

Is this expected or should it keep the old behavior?

awx/api/serializers.py
```diff
@@ -2603,13 +2605,31 @@ def get_summary_fields(self, obj):
...
         return ret
 
+    def validate_organization(self, org):
+        if (
+            self.instance and
+            self.instance.credential_type.kind == 'galaxy' and
+            org is None
+        ):
+            raise serializers.ValidationError(_(
+                "Galaxy credentials must be owned by an Organization."
```
chrismeyersfsu, line 2629: What if the org is deleted?

awx/api/serializers.py
```diff
@@ -2674,6 +2694,14 @@ def validate(self, attrs):
         if attrs.get('team'):
             attrs['organization'] = attrs['team'].organization
 
+        if (
+            attrs['credential_type'].kind == 'galaxy' and
+            list(owner_fields) != ['organization']
+        ):
+            raise serializers.ValidationError({"organization": _(
+                "Galaxy credentials must be owned by an Organization."
```
chrismeyersfsu, line 2702: Yes. @alancoding wants to do away with team ownership and this aligns.

awx/main/management/commands/create_preload_data.py
```diff
@@ -42,6 +42,16 @@ def handle(self, *args, **kwargs):
                                               },
                                               created_by=superuser)
                 c.admin_role.members.add(superuser)
+                public_galaxy_credential = Credential(
```
chrismeyersfsu, line 45: I'm on my cell phone. I assume this only runs on fresh installs and the migration will create this otherwise.

ryanpetrello: Yep, here:

https://github.com/ansible/awx/pull/7817/files#diff-990db3abe66cd302b90fe644303d7879R50
https://github.com/ansible/awx/pull/7817/files#diff-137317559e46e7be3f56a14d5b5fe1e7R18-R20

unlikelyzero:
Issues:

Add/ Edit Credential form does not list Organization as a required field when creating or editing an 'Ansible Galaxy/Automation Hub API Token' credential
In-line helper text for 'Ansible Galaxy/Automation Hub API Token' could be more helpful

Tests:
Can create an Ansible Galaxy/Automation Hub API Token credential type
Can create an Ansible Galaxy/Automation Hub API Token credential type configured for cloud.redhat
Can create an Ansible Galaxy/Automation Hub API Token credential type configured for an on-prem AH server
Can assign a Ansible Galaxy/Automation Hub API Token credentials to a new Organization
Can assign multiple Ansible Galaxy/Automation Hub API Token credentials to a new Organization
Can assign a Ansible Galaxy/Automation Hub API Token credentials to an existing Organization
Can assign multiple Ansible Galaxy/Automation Hub API Token credentials to an existing Organization
Can remove Ansible Galaxy/Automation Hub API Token credentials from an existing Organization
If a project cannot reach the first Ansible Galaxy/Automation Hub API Token credential server configured, it will fallback to the second Ansible Galaxy/Automation Hub API Token credential

awx/main/models/credential/__init__.py
```diff
@@ -1173,6 +1178,38 @@ def create(self):
 )
 
 
+ManagedCredentialType(
+    namespace='galaxy_api_token',
+    kind='galaxy',
+    name=ugettext_noop('Ansible Galaxy/Automation Hub API Token'),
```
AlanCoding, line 1184: These highly descriptive names are kind of messy from the user's perspective using the name to reference the object, from the CLI or the collection like

```yaml
tower_credential:
  credential_type: "Ansible Galaxy/Automation Hub API Token"
```

rooftopcellist (approved):

awx/api/serializers.py
AlanCoding: I don't want to muck with the credential organization set null behavior

https://github.com/ansible/awx/blob/f4af74dabebc4d716ffe470b834c23e293f5c94f/awx/main/models/credential/__init__.py#L103-L109

So it would leave it null. I'm not too terribly bothered by that. As long as an org admin can still _remove_ it, I'm okay with it.

(note there are 2 orgs in play here, the one who owns the credential and the one using the credential, I am talking about the ownership org being deleted)

awx/api/serializers.py
```diff
@@ -2674,6 +2694,14 @@ def validate(self, attrs):
         if attrs.get('team'):
             attrs['organization'] = attrs['team'].organization
 
+        if (
+            attrs['credential_type'].kind == 'galaxy' and
```
AlanCoding, line 2698: what if `attrs['credential_type']` is null or the key is not present?

awx/main/models/credential/__init__.py
```diff
@@ -1173,6 +1178,38 @@ def create(self):
...
+                'The URL of a Keycloak server token_endpoint, if using '
+                'SSO auth.'
+            )
+        },{
+            'id': 'token',
+            'label': ugettext_noop('API Token'),
+            'type': 'string',
+            'secret': True,
+            'help_text': ugettext_noop(
+                'A token to use for authentication against the Galaxy instance.'
```
AlanCoding, line 1205: To get legalistic with this - the token could be used for either Galaxy or Automation Hub, but the auth_url would only be used with Automation Hub.

awx/main/tasks.py
```diff
@@ -2138,8 +2141,8 @@ def build_extra_vars_file(self, project_update, private_data_dir):
             'scm_url': scm_url,
             'scm_branch': scm_branch,
             'scm_clean': project_update.scm_clean,
-            'roles_enabled': settings.AWX_ROLES_ENABLED,
-            'collections_enabled': settings.AWX_COLLECTIONS_ENABLED,
+            'roles_enabled': galaxy_creds_are_defined and settings.AWX_ROLES_ENABLED,
+            'collections_enabled': galaxy_creds_are_defined and settings.AWX_COLLECTIONS_ENABLED,
```
AlanCoding, line 2145: I would predict users having trouble debugging this, and I would like to put a docs note for it.

@tvo318 just to throw out some initial statement and wording:

If an organization does not have any Galaxy credentials associated with it, it will not download roles or collections from requirements files. On creation of a new organization:

 - the UI will automatically associate the public Galaxy credential (for now, may change in later releases)
 - organizations created via the API need to have the public Galaxy credential added in order to make use of roles/collection requirements files.

AlanCoding:
@john-westcott-iv recently added some tests to help keep the AWX collection up-to-date with the API, but for now related lists are out of scope of those checks (it is hard to determine what the editable relationships are)

Given the default API behavior, I do see a lot of value in making the corresponding collection update.

https://github.com/ansible/awx/blob/devel/awx_collection/plugins/modules/tower_organization.py

We would be adding the galaxy_credentials relationship as a list field. This can be modeled off a field like `notification_templates_started`.

Additionally, since the public Galaxy credential is statically named, I think we should add that to the examples for `tower_organization`, because the named reference will be universally correct.

Spitball:

```yaml
- name: Create tower organization
  tower_organization:
    name: "My New Organization"
    description: "This has public galaxy enabled and will install roles and collections"
    galaxy_credentials:
      - Ansible Galaxy
```

awx/main/models/credential/__init__.py
AlanCoding: but no change here makes a lot of sense. We might want to add an issue to reference credential types by `namespace` in the collection modules.

jakemcdermott (approved):

Issue #7814, opened 2020-08-04
API error on a credential input field causes page to crash


mabashian:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
This would apply to any form where some of the values are nested under a single top-level field.  It's easiest to see on credentials where the model looks like:

```
name
description
....
inputs
    url
    token
    api_version
```

If there's an error with one of the `inputs` fields when POSTing or PUTing on this endpoint the response will look something like:

```
{
    inputs: {
        url: ["Invalid URL: Missing url scheme (http, https, etc.)"]
    }
}
```

So the actual error is not at the top level but rather nested down one level.  This means we're falling in to this bit of logic:

https://github.com/ansible/awx/blob/devel/awx/ui_next/src/components/FormField/FormSubmitError.jsx#L26

and we're pushing an _object_ onto the `messages` array.  Then we're trying to display that value here:

https://github.com/ansible/awx/blob/devel/awx/ui_next/src/components/FormField/FormSubmitError.jsx#L47

and React doesn't like that:

<img width="1678" alt="Screen Shot 2020-08-04 at 12 23 02 PM" src="https://user-images.githubusercontent.com/9889020/89318789-45cb5200-d64d-11ea-88c1-985e5c0eb7d0.png">

##### STEPS TO REPRODUCE
1. Attempt to create a Credential of type CyberArk AIM Central Credential...
2. For the CyberArk AIM URL field type in an arbitrary string like `foo`
3. Hit save

##### EXPECTED RESULTS
Error shown under field

##### ACTUAL RESULTS
<img width="1678" alt="Screen Shot 2020-08-04 at 12 23 02 PM" src="https://user-images.githubusercontent.com/9889020/89318955-81661c00-d64d-11ea-8c9c-8bb52d4f7893.png">

PR #7821, opened 2020-08-04
Handle form submission errors that may be deeply nested in the return object

mabashian:
##### SUMMARY
link #7814 

The issue here was that the error message for an error on the `url` field in the credential form was nested down in the second level of the return object.  The code that we previously had in FormSubmitError didn't handle that very well so I built a recursive solution that will traverse the entire object and gather up all of the strings in an array to be displayed in the Alert.  This solution also handles arrays of strings which can happen when a field has multiple errors.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/components/FormField/FormSubmitError.jsx
+++ b/awx/ui_next/src/components/FormField/FormSubmitError.jsx
@@ -2,6 +2,26 @@ import React, { useState, useEffect } from 'react';
 import { useFormikContext } from 'formik';
 import { Alert } from '@patternfly/react-core';
 
+const findErrorStrings = (obj, messages = []) => {
+  if (typeof obj === 'string') {
+    messages.push(obj);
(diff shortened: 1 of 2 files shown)


jakemcdermott (approved):

awx/ui_next/src/components/FormField/FormSubmitError.jsx
```diff
@@ -2,6 +2,26 @@ import React, { useState, useEffect } from 'react';
 import { useFormikContext } from 'formik';
 import { Alert } from '@patternfly/react-core';
 
+const findErrorStrings = (obj, messages = []) => {
+  if (typeof obj === 'string') {
+    messages.push(obj);
```
jakemcdermott, line 7: lgtm

jakemcdermott (approved):

PR #7823, opened 2020-08-04
Catching runtime error

john-westcott-iv:
##### SUMMARY

It was found that a job which was running during an AWX backup and restored into a new cluster was throwing a 500 error when attempting to be canceled. See https://access.redhat.com/support/cases/#/case/02715439 for full issue and stack. In 
awx/main/dispatch/control.py if we can't reach the dispatcher we:
```
                if reply is None:
                    logger.error(f'{self.service} did not reply within {timeout}s')
                    raise RuntimeError(f"{self.service} did not reply within {timeout}s")
                break
```

But the cancel code was only excepting socket.timeout which was causing the raise to bubble up into a 500. This fix catches both the socket.timeout as well as the returned RuntimeError and then properly allows the job to be canceled when the dispatcher is unreachable.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 13.0.0
```

##### ADDITIONAL INFORMATION

```
See access ticket for details on problem replication.
```

### DIFF
--- a/awx/main/models/unified_jobs.py
+++ b/awx/main/models/unified_jobs.py
@@ -1366,7 +1366,7 @@ class UnifiedJob(PolymorphicModel, PasswordFieldsModel, CommonModelNameNotUnique
                 running = self.celery_task_id in ControlDispatcher(
                     'dispatcher', self.controller_node or self.execution_node
                 ).running(timeout=timeout)
-            except socket.timeout:
+            except (socket.timeout, RuntimeError):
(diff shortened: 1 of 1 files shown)


ryanpetrello (approved):
Yep, this makes perfect sense to me; this is something that just got lost in translation when we moved from RabbitMQ to Redis.  Thanks for digging into this!

Issue #7619, opened 2020-07-14
tower_credentials.ply


jpell958:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
When the **input** parameter for tower_credentials.ply is not supplied, our expected behavior of tower_credentials would be to not change any of the **input** values like usernames and  passwords .  However, we find that any username or password previously set is wiped out and set to blank.

##### ENVIRONMENT
* AWX version: 12.0.0
* AWX install method: we developed an Ansible script to install 
* Ansible version:  2.9.9
* Operating System: RHEL Linux 7
* Web Browser: Chromium Edge 

##### STEPS TO REPRODUCE

1. Create a Machine credential on AWX with a username and password set.
2. Using tower_credential update the same credential but do not use the inputs parameter

##### EXPECTED RESULTS

tower_credential should not change any existing input fields if input parameter is not supplied; i.e.: usernames and passwords previously set should not change

##### ACTUAL RESULTS

tower_credential is changing previously set input fields; i.e.: usernames and passwords 

##### ADDITIONAL INFORMATION

none

jakemcdermott:
> When the input parameter for tower_credentials.ply is not supplied, our expected behavior of tower_credentials would be to not change any of the input values like usernames and passwords

I'm not sure I agree. Correctly designed ansible modules are idempotent and so not providing input values to the module _could_ be construed as "put the credential into a state with no inputs (clear them)"

More importantly this probably depends on how other awx modules treat this sort of thing. @AlanCoding thoughts?

jpell958:
I wish to be able to change any of the non-input credential's attributes without knowing or modifying the password or other inputs.  There should be a way to to keep current input values in place while updating other attributes.  If you don't specify the input field leave those items alone.  If you want to wipe certain one's then post a null or "" - thoughts?  

I would like to code this and do a pull request

AlanCoding:
I agree with @jpell958's expectation. This is the whole not-provided vs. blank issue.

It is _extremely_ uncommon that an API parameter should have a default. Some parameters related to the _module action_ should validly have a default. These are parameters that tell the module _what to do_, such as `state='present'` this is a sensible default. Some API fields may be _required_ because it's hard to make sense of intent without it. However, the vast majority of API fields should be _optional_ and have _no default_.

[...]

PR #7840, opened 2020-08-05
Make 'inputs' Unchanged When Blank in Credentials Module

beeankha:
Addressing Issue https://github.com/ansible/awx/issues/7619

##### SUMMARY

Previously-set `inputs` parameters will no longer be wiped out when no inputs are provided in a separate playbook task for the same credential.  When inputs ARE present, passwords still come back as `$encrypted$` so those always return as "changed" if the input specifies it. 

(this solution has been discussed/developed with @john-westcott-iv )

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - Collections

### DIFF
--- a/awx_collection/plugins/modules/tower_credential.py
+++ b/awx_collection/plugins/modules/tower_credential.py
@@ -384,19 +384,25 @@ def main():
         team_id = module.resolve_name_to_id('teams', team)
 
     # Create credential input from legacy inputs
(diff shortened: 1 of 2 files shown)


AlanCoding (approved):

TristanCacqueray:
recheck

beeankha:
regate

PR #7842, opened 2020-08-05
Remove showExpandCollapse prop from the DataListToolbar calls

nixocio:
Remove showExpandCollapse prop from the DataListToolbar calls. This is
not an expected prop to be passed to this component.

Inside DataListToolbar.

```
  const showExpandCollapse = onCompact && onExpand;
```

In order to use this feature, `onCompact` and `onExpand` props should
be passed.

### DIFF
--- a/awx/ui_next/src/components/DataListToolbar/DataListToolbar.test.jsx
+++ b/awx/ui_next/src/components/DataListToolbar/DataListToolbar.test.jsx
@@ -36,7 +36,6 @@ describe('<DataListToolbar />', () => {
       <DataListToolbar
         qsConfig={QS_CONFIG}
         isAllSelected={false}
-        showExpandCollapse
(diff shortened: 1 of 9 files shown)


nixocio:
See: https://github.com/ansible/awx/blob/1c0dc1cccb19feb41d2d5c5fdfdb81647515efb8/awx/ui_next/src/components/DataListToolbar/DataListToolbar.jsx#L20

jakemcdermott (changes requested):
:+1: Nice find

@nixocio Let's remove it from the associated unit test as well: https://github.com/ansible/awx/blob/devel/awx/ui_next/src/components/DataListToolbar/DataListToolbar.test.jsx#L39

nixocio:
done @jakemcdermott.

jakemcdermott (approved):

Issue #5910, opened 2020-02-11
Support prompting on WFJT Launch


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
See #5909

Launching a WFJT is a bit different from launching a JT.  It has the possibility to display a subset of the steps to the launching user:

Note that the api provides us with a convenience endpoint: `/api/v2/workflow_job_templates/:id/launch/`.  This is where we should pull all of these flags from.

Potential Steps:

1) Inventory
    - This step is only shown if ask_inventory_on_launch is set to true
    - If there's a default inventory it should be selected
    - If there's no default inventory, the user can proceed to the next step without selecting an inventory.  This differs from JT launch.
    - If there's a default inventory, the user can only proceed if an inventory is selected (either the default or a replacement).
2) Other Prompts
    - This step is shown when one or more of the follow flags is set to true:
        - ask_scm_branch_on_launch
        - ask_limit_on_launch
        - ask_variables_on_launch
    - None of these fields are required and the user can proceed to the next step without specifying any of them
3) Survey
    - This step is shown when survey_enabled is set to true
    - An extra request to get the survey spec (questions) is required in order to render this step
    - Survey questions will be dynamically rendered based on the survey spec and should be presented in the same order as they appear in the spec.
    - Any required fields must be filled out before the user can proceed to the Preview step.
4) Preview
    - This step will _always_ be shown as long as some prompting has occurred (i.e. this wizard is actually shown).
    - This is a read-only step where all the prompted values are combined with WFJT defaults to give the launching user a snapshot of the configuration they're launching with.
    - Clicking the Launch button should hit the launch endpoint with all the specified overrides.

mabashian:
I'm gonna mark this as blocked until #5909 is done.  It possible that the implementation there will be 100% of what we need here and I don't want to duplicate effort.

nixocio:
Since #5909 is completed removing the `blocked` label.

dsesami:
Update: I suspect this has something to do with the current containers and not the WFJT form.

~I try to enable all of the workflow job template fields within the UI on devel, and a 500 error is generated when I try to save (usually on edit, not initial creation).~

Confirmed the above was due to a messed up DB.

unlikelyzero:
covered by 5465

PR #7851, opened 2020-08-06
Support workflow prompting on launch

mabashian:
##### SUMMARY
Link #5910 

The only real difference between workflow job launch prompting and job template job launch prompting is that with workflows the inventory is an optional field _if_ a default inventory is **not** present.  So I had to add a bit of conditional logic to the validator function for the inventory step to take that into account.  Also I _think_ that the validate passed into useField is not necessary in either case since we're handling validation up at a higher level.  If that is necessary for job template launching then we'll need to conditionally apply it since the field is only sometimes required when launching a workflow.  Otherwise the current solution was already sufficient for covering all of the possible workflow prompts.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI


marliana (approved):

marshmalien (approved):
LGTM 👍

Issue #7793, opened 2020-07-31
Importing workflow_job_templates does not set node job_templates correctly


zedr:
##### ISSUE TYPE
When importing workflow_job_templates, the `unified_job_template` field in each associated workflow_job_template_node is incorrectly set to `null`.

##### SUMMARY
AWX CLI has recently introduced the ability to export and import many types of AWX objects, including Workflow Job Templates. These objects have related objects called Workflow Job Template Nodes, that can point to an existing Job Template.

Unfortunately, when importing a previously exported resource, any Workflow Job Template Node that points to a Job Template, using its natural key, will not be able to find that object, even it it is present. The result a mangled workflow.

I've run into this problem when exporting a Workflow Job Template from an AWX 9.2.0 to Tower 3.7.1

##### ENVIRONMENT
* AWX version: Tower 3.7.1
* AWX install method: openshift
* Ansible version:  2.9.9
* Operating System: OpenShift 3.11, RHEL
* Web Browser: Google Chrome

##### STEPS TO REPRODUCE

Export a Workflow Job Tempates with nodes that point to existing Job Templates, e.g.

```
workflow_job_templates:
-   allow_simultaneous: true
    ask_inventory_on_launch: false
    ask_limit_on_launch: false
    ask_scm_branch_on_launch: true
    ask_variables_on_launch: true
    description: WF MyTest provisioning
    extra_vars: ''
    inventory: 25
    limit: null
    name: WF-infra_provisioning
    natural_key:
        name: WF-infra_provisioning
        organization:
            name: ITDP
            type: organization
        type: workflow_job_template
    organization:
        name: ITDP
        type: organization
    related:
        notification_templates_approvals: []
        notification_templates_error: []
        notification_templates_started: []
        notification_templates_success: []
        schedules: []
        survey_spec: {}
        workflow_nodes:
        -   all_parents_must_converge: false
            diff_mode: null
            extra_data: {}
            inventory: null
            job_tags: null
            limit: null
            natural_key:
                type: workflow_job_template_node
                workflow_job_template:
                    name: WF-infra_provisioning
                    organization:
                        name: ITDP
                        type: organization
                    type: workflow_job_template
            related:
                always_nodes: []
                credentials: []
                failure_nodes: []
                success_nodes:
                -   type: workflow_job_template_node
                    workflow_job_template:
                        name: WF-infra_provisioning
                        organization:
                            name: ITDP
                            type: organization
                        type: workflow_job_template
            scm_branch: null
[...]

PR #7856, opened 2020-08-07
Fix early return in assign related method

zedr:
##### SUMMARY
This change fixes an erroneus early return in a private method that was
preventing more than one type of related object from being correctly
assigned to the parent object, and therefore imported.

Also, a minor spelling mistake was corrected.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
- CLI

##### AWX VERSION
```
awx: 14.0.0
```

##### ADDITIONAL INFORMATION
See #7793.

### DIFF
--- a/awxkit/awxkit/api/pages/api.py
+++ b/awxkit/awxkit/api/pages/api.py
@@ -268,9 +268,9 @@ class ApiV2(base.Base):
     def _assign_related(self):
         for _page, name, related_set in self._related:
             endpoint = _page.related[name]
-            if isinstance(related_set, dict):  # Relateds that are just json blobs, e.g. survey_spec
+            if isinstance(related_set, dict):  # Related that are just json blobs, e.g. survey_spec
                 endpoint.post(related_set)
-                return
+                continue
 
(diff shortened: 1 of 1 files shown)


jbradberry (approved):
Oh.  Oops.

awxkit/awxkit/api/pages/api.py
```diff
@@ -268,9 +268,9 @@ def _assign_roles(self):
     def _assign_related(self):
         for _page, name, related_set in self._related:
             endpoint = _page.related[name]
-            if isinstance(related_set, dict):  # Relateds that are just json blobs, e.g. survey_spec
+            if isinstance(related_set, dict):  # Related that are just json blobs, e.g. survey_spec
```
jbradberry, line 271: The original here wasn't a spelling mistake.  Maybe "Related objects that are..." would be clearer, though?

ryanpetrello (approved):
Great find, @zedr - thanks for the patch!

awxkit/awxkit/api/pages/api.py
zedr: Yes, possibly.

Issue #7767, opened 2020-07-29
Edit instance groups


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Add feature to edit instance groups.

dsesami:
This feature was merged, stubs are written. :100:

PR #7864, opened 2020-08-10
Add feature to edit instance group

nixocio:
Add feature to edit instance group.

See: https://github.com/ansible/awx/issues/7767

image

### DIFF
--- a/awx/ui_next/src/screens/InstanceGroup/shared/InstanceGroupForm.jsx
+++ b/awx/ui_next/src/screens/InstanceGroup/shared/InstanceGroupForm.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { func, shape } from 'prop-types';
-import { Formik } from 'formik';
+import { Formik, useField } from 'formik';
 import { withI18n } from '@lingui/react';
 import { t } from '@lingui/macro';
 import { Form } from '@patternfly/react-core';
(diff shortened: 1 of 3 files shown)


awx/ui_next/src/screens/InstanceGroup/shared/InstanceGroupForm.jsx
```diff
@@ -11,15 +11,17 @@ import { required, minMaxValue } from '../../../util/validators';
...
     <>
       <FormField
+        name="name"
         id="instance-group-name"
         label={i18n._(t`Name`)}
-        name="name"
         type="text"
         validate={required(null, i18n)}
         isRequired
+        isDisabled={instanceGroupNameField.value === 'tower'}
```
jakemcdermott, line 24: In addition to the validation ranges you've provided, please add a min & max value for the inputs - They're supported and improve the UX when setting these values with a mouse wheel or keyboard arrows:

Screenshot from 2020-08-11 09-33-22

awx/ui_next/src/screens/InstanceGroup/shared/InstanceGroupForm.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
```
jakemcdermott, line 1: @nixocio Since IGs don't have an add subscreen, this component isn't actually shared. I think we should move this file to `InstanceGroup/InstanceGroupEdit` and delete the `shared` folder (for now, that is. Container Groups may actually have a shared form but I presume you'll do that in a different PR)

jakemcdermott (changes requested):

awx/ui_next/src/screens/InstanceGroup/shared/InstanceGroupForm.jsx
nixocio: Thanks for your review @jakemcdermott. Updating now.

awx/ui_next/src/screens/InstanceGroup/shared/InstanceGroupForm.jsx
nixocio: `InstanceGroupForm` is shared between the `InstanceGroupAdd` and `InstanceGroupEdit` even in this case, is your suggestion still valid? I was trying to follow the same pattern that already in code basis. But it seems to be valid just in cases we have a shared `subscreen`. I will evaluate this change.

jakemcdermott: @nixocio Whoops, you're right. IGs _do_ have an add form. Never mind! It is fine as is.

jakemcdermott (approved):

PR #7873, opened 2020-08-11
Project update

sean-m-sullivan:
##### SUMMARY
Adding a module to the awx collection that updates project similar to tower_job_launch

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
Collection

##### AWX VERSION
awx: 14.0.0


awx_collection/tests/integration/targets/tower_project_update/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
+---
+- name: Update a project without waiting
+  tower_project_update:
+    name: "AWX-Collection-tests-tower_job_template-proj-{{ test_id }}"
```
AlanCoding: maybe I missed something, but doesn't a project with this name need to be created first?

awx_collection/plugins/modules/tower_project_update.py
```diff
@@ -0,0 +1,146 @@
...
+
+    # Grab our start time to compare against for the timeout
+    start = time.time()
+
+    project_url = result['json']['url']
+    while not result['json']['finished']:
+        # If we are past our time out fail with a message
+        if timeout and timeout < time.time() - start:
+            module.json_output['msg'] = "Monitoring aborted due to timeout"
+            module.fail_json(**module.json_output)
```
AlanCoding: @john-westcott-iv I think it would be interesting to look at shared utilities between this as tower_job_launch, these patterns feel very familiar

awx_collection/plugins/modules/tower_project_update.py
```diff
@@ -0,0 +1,146 @@
...
+          amount of seconds
+      type: int
+extends_documentation_fragment: awx.awx.auth
+'''
+
+RETURN = '''
+project_info:
+    description: dictionary containing information about the project updated
+    returned: If project synced
+    type: dict
```
john-westcott-iv: I'm torn on this return. We do a similar dict return in tower_workflow_launch. However, in tower_job_cancel, tower_job_launch, and tower_job_wait we are returning specific items (i.e. id, time elapsed, status, etc). I think I like returning specific documented things better than a generic hash with no definition of what is in it. Thoughts?

awx_collection/plugins/modules/tower_project_update.py
```diff
@@ -0,0 +1,146 @@
...
+      description:
+        - Organization the project exists in.
+        - Used to help lookup the object, cannot be modified using this module.
+        - If not provided, will lookup by name only, which does not work with duplicates.
+      type: str
+    wait:
+      description:
+        - Wait for the project to update.
+      default: True
+      type: bool
```
john-westcott-iv, line 39: If we are going to make wait optional do we need a second tower_project_wait (like tower_job_launch and tower_job_wait) so that you can launch a project update, do some stuff in the background and then bring the wait into the foreground to get the results? This is really the async function in Ansible but I think our modules and pattern was developer before async was an actual thing.

awx_collection/plugins/modules/tower_project_update.py
john-westcott-iv: Yes, I agree, we could probably make a wait_until_finished method which took a URL, interval, error msg, etc and performed this loop.

awx_collection/plugins/modules/tower_project_update.py
john-westcott-iv: It looks like below we are returning specific pieces of information, not a project_info dict.

awx_collection/plugins/modules/tower_project_update.py
```diff
@@ -0,0 +1,146 @@
...
+
+ANSIBLE_METADATA = {'metadata_version': '1.0',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+---
+module: tower_project_update
+author: "Sean Sullivan (@sean-m-sullivan)"
+short_description: Sync a Project in Ansible Tower
```
john-westcott-iv: We use both the Sync and Update terms in the module and its name. We should pick one.

awx_collection/plugins/modules/tower_project_update.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def main():
+    # Any additional arguments that are not fields of the item can be added here
+    argument_spec = dict(
+        name=dict(required=True, aliases=['project']),
+        organization=dict(),
+        wait=dict(required=False, default=True, type='bool'),
+        interval=dict(required=False, default=1.0, type='float'),
+        timeout=dict(required=False, default=None, type='int'),
```
john-westcott-iv: You can remove required=False here.

awx_collection/tests/integration/targets/tower_project_update/tasks/main.yml
john-westcott-iv: The object name structure is also: `AWX-Collection-tests-<module>-<asset type>-<id>`. So tower_job_template should be later to tower_project_update. And, since you use it down below it should probably be in a variable. Check out tower_export for an example of creating a test_id, creating multiple assets and then running things in a block to try and ensure cleanup on a failed test.

john-westcott-iv (changes requested):

awx_collection/tests/integration/targets/tower_project_update/tasks/main.yml
sean-m-sullivan: It does need to be created, which is why I took the info from the tower project job thinking it would be run after.

awx_collection/plugins/modules/tower_project_update.py
sean-m-sullivan: I took the tower_job_launch as a template to use to make this, and changed it to update project. So it is definitely the same code.

awx_collection/plugins/modules/tower_project_update.py
sean-m-sullivan: Will update that.

awx_collection/plugins/modules/tower_project_update.py
sean-m-sullivan: Good question, I had more thought of it, that You can wait for the results, but a given Job template is blocked until the corresponding project is updated, so if you launch a template afterwards it is still effectively doing the wait. Leaving the option up to what the user decides.

awx_collection/tests/integration/targets/tower_project_update/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+    organization: Default
+    wait: False
```
beeankha: Any `False` or `True` values in the test playbooks need to be lower case to pass lint check

awx_collection/tests/integration/targets/tower_project_update/tasks/main.yml
sean-m-sullivan: Updating this to include its own project.

sean-m-sullivan: Thanks for the tip using that as a reference

awx_collection/plugins/modules/tower_project_update.py
sean-m-sullivan: Updated to include better information about what is returned

awx_collection/plugins/modules/tower_project_update.py
sean-m-sullivan: Do we want to do this now, or do this as a separate update.

awx_collection/plugins/modules/tower_project_update.py
```diff
@@ -0,0 +1,152 @@
...
+    - Update a Ansible Tower Project. See
+      U(https://www.ansible.com/tower) for an overview.
+options:
+    name:
+      description:
+        - The name of the project to update.
+      required: True
+      type: str
+      aliases:
+        - project
```
AlanCoding, line 28: @john-westcott-iv this wouldn't allow specifying the project by id. Do we have a pattern for that?

awx_collection/plugins/modules/tower_project_update.py
AlanCoding: I'm fine to punt, maybe leave a comment about it

awx_collection/plugins/modules/tower_project_update.py
AlanCoding: I'm always in favor of fewer modules. Maybe we could add the wait option there, and draft an example with `awx.awx.tower_api` lookup plugin to deprecate the tower_job_wait module?

awx_collection/plugins/modules/tower_project_update.py
sean-m-sullivan: I did not think we have a pattern for that in other modules, but I'll look

awx_collection/plugins/modules/tower_project_update.py
sean-m-sullivan: If we were to put it in tower_api, we could build it into the "Create" modules like tower_project, tower_job_template, with less required options, that would fail on create, but would allow for an "update", First thing I tried with tower_project was doing this, and it overwrote the things to blank that I didn't provide.

awx_collection/plugins/modules/tower_project_update.py
sean-m-sullivan: I'm inclined to punt it atm, but put it on my docket of things to do. Been putting in the work on the schema PR, got a good set of test cases for it, updated Documentation, but ran into snag where its throwing an error, figuring that out.

awx_collection/plugins/modules/tower_project_update.py
john-westcott-iv: In this case it would not because get_one does not attempt to do an ID if it can't find something by name. resolve_name_to_id does allow for a ID and will try an ID after failing to find something by name. But resolve_name_to_id does not allow for additional search criteria. We should think about combining these functionalities.

john-westcott-iv (commented):
One kind of high level thing here. For tower_job we have: launch, wait, cancel and list. List may not be applicable to this but do we want to follow the same patter with launch, cancel and wait for tower_project_update? (i.e. make this tower_project_update_launch and add a wait and cancel modules).

awx_collection/tests/integration/targets/tower_project_update/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+- name: Update a project and wait
+  tower_project_update:
+    name: "{{ project_name1 }}"
+    organization: Default
+    wait: true
+  register: result
+
+- assert:
+    that:
+      - result is changed
```
squidboylan: If this returns the job data we should assert that it's done (to verify that `wait` actually works), im not exactly sure what data we get back here

sean-m-sullivan: I've updated this to successful here was the result info
```
{
    "result": {
        "changed": true,
        "failed": false,
        "id": 925,
        "status": "successful",
    }
}
```

sean-m-sullivan:
> One kind of high level thing here. For tower_job we have: launch, wait, cancel and list. List may not be applicable to this but do we want to follow the same patter with launch, cancel and wait for tower_project_update? (i.e. make this tower_project_update_launch and add a wait and cancel modules).
@john-westcott-iv 
So I've looked into this, and at least I thought I'd break down thoughts on these. 

wait: 
because projects are short and don't make changes, I think its applicable to put not have a separate module, especially as tower_project and tower_project update have wait built in. That being said, tower_workflow_launch has wait built in, while tower_job_launch does not, thus requiring job_launch + job wait to achieve the same results. 

cancel:
I believe this is applicable for jobs, but general project updates will either quickly fail, timeout, or succeed and don't make changes versus workflows and jobs where you might want to cancel them via module. 

Launch:
I think this is just down to nomenclature, projects update, jobs launch, as per the api, 
/api/v2/projects/{id}/update/
/api/v2/job_templates/{id}/launch/
/api/v2/workflow_job_templates/{id}/launch/

sean-m-sullivan:
@AlanCoding @john-westcott-iv 

I'm working on the name_to_id, but think it may be beyond the scope of this merge, same with the wait as a function, Do we want to merge this without those items? or should we wait on those changes.

awx_collection/plugins/modules/tower_project_update.py
sean-m-sullivan: This module now resolves by id or name. I tried to do it as a global change but when coupled with project_update, it would change the name to the ID#, as that project hardcodes where it gets the name from the module args. If we want to implement this collection wide, we would need to rewrite all the modules where this is the case.

awx_collection/plugins/modules/tower_project_update.py
sean-m-sullivan: Added #8022 to address this, once we have some feedback on this, or even after this is merged, will update this module to use the new method.

AlanCoding:
This is looking good to me, if integration tests pass then I agree it makes sense and then later return to the consolidation of code you've already started on.

AlanCoding (approved):
I think all the feedback has been addressed. After I get integration results I'll check back in.

beeankha (approved):

sean-m-sullivan:
recheck

sean-m-sullivan:
@john-westcott-iv wait changed merged, should satisfy the change request.

awx_collection/plugins/modules/tower_project_update.py
```diff
@@ -0,0 +1,150 @@
...
+
+    # Extract our parameters
+    name = module.params.get('name')
+    organization = module.params.get('organization')
+    wait = module.params.get('wait')
+    interval = module.params.get('interval')
+    timeout = module.params.get('timeout')
+
+    # Attempt to look up project based on the provided name or id
+    if name.isnumeric():
```
AlanCoding: this is breaking tests on python2 because this syntax only works for python3

AlanCoding:
recheck

awx_collection/plugins/modules/tower_project_update.py
sean-m-sullivan: Updated to use isdigit instead, added more tests and better fail messages.

awx_collection/plugins/modules/tower_project_update.py
```diff
@@ -0,0 +1,150 @@
...
+    organization = module.params.get('organization')
+    wait = module.params.get('wait')
+    interval = module.params.get('interval')
+    timeout = module.params.get('timeout')
+
+    # Attempt to look up project based on the provided name or id
+    if name.isdigit():
+        results = module.get_endpoint('projects', **{'data': {'id': name}})
+        if results['json']['count'] == 0:
+            module.fail_json(msg='Could not find Project with ID: {0}'.format(name))
```
AlanCoding: this hit

```
plugins/modules/tower_project_update.py:109:85: W291: trailing whitespace
```

sean-m-sullivan:
recheck

AlanCoding (approved):

sean-m-sullivan:
regate

sean-m-sullivan:
recheck

john-westcott-iv (approved):

PR #7895, opened 2020-08-12
update existing relatedSearchKey requests to new convention and fix U…

jlmitch5:
part of #7740

this is a quick refactoring of all the existing search key login in the app.  i still need to add the login to routes that were recently converted to useRequest


marshmalien (approved):
🔥 LGTM!

PR #7904, opened 2020-08-13
Use 'secure' cookies in containers/OCP as well.

wenottingham:
##### SUMMARY

This flips the default settings for the session & CSRF cookies for containerized deployments.
These settings were always configurable. I think they were unset before due to some load balancer
weirdness(?)

Tested:
- [:heavy_check_mark:] dev env
- [:heavy_check_mark:] OCP
- [:heavy_check_mark:] local docker

Untested:
- [ ] k8s

##### ISSUE TYPE
 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME
 - API
 - Installer

### DIFF
--- a/awx/settings/defaults.py
+++ b/awx/settings/defaults.py
@@ -232,6 +232,7 @@ START_TASK_LIMIT = 100
 
 # Disallow sending session cookies over insecure connections
 SESSION_COOKIE_SECURE = True
+SESSION_COOKIE_HTTPONLY = True
 
 # Seconds before sessions expire.
 # Note: This setting may be overridden by database settings.
--- a/awx/settings/development.py
+++ b/awx/settings/development.py
@@ -57,10 +57,10 @@ mimetypes.add_type("image/svg+xml", ".svg", True)
 mimetypes.add_type("image/svg+xml", ".svgz", True)
 
 # Disallow sending session cookies over insecure connections
-SESSION_COOKIE_SECURE = False
+SESSION_COOKIE_SECURE = True
 
 # Disallow sending csrf cookies over insecure connections
-CSRF_COOKIE_SECURE = False
+CSRF_COOKIE_SECURE = True
 
 # Override django.template.loaders.cached.Loader in defaults.py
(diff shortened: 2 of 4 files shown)


ryanpetrello:
@matburt @shanemcd either of you remember why did this?  Something something k8s load balancer?

matburt:
I'm pretty sure this has to do with us not using ssl configured on our internal webserver so that the Router or Ingress Controller could manage that. We'll want to verify everything is kosher there

shanemcd:
Definitely want to test this on GKE.

wenottingham:
@shanemcd (gentle prod)

shanemcd:
Testing this on GKE now.

shanemcd:
I dont seem to be able to log in via the UI or the API browser with this patch. UI just throws a generic "Invalid username and/or password." message. The API shows:

image

ryanpetrello:
@shanemcd

https://github.com/ansible/awx/pull/7904/files#diff-0f169a0b2ffca626e39297fce9f5dbb5R66

Are you actually running your GKE test behind load balancing w/ TLS and not plain HTTP?  This flag ^ makes it so that the `Set-Cookie` for the CSRF cookie will specify `secure`, which means the browser is only to use the cookie *iff* the request in the web browser is actually HTTPS.  I'm guessing that's not the case in your testing, hence the CSRF failure (because the CSRF cookie isn't working).

shanemcd:
Nope, the default Ingress created by GKE is just HTTP 😞

ryanpetrello:
Yea, I think to effectively test this on GKE, we'd have to set up HTTPS.  Also, if we merged something like this, we'd probably have to make it pretty clear to the AWX community what changed and why we changed it (and how to disable it if they're not inclined to set up load balancing w/ SSL termination).  Otherwise, people are gonna upgrade and not be able to log in.

ryanpetrello:
@wenottingham if you want to rebase this and remove the `./installer` bits, it's technically likely mergeable in that state.

But we probably also want to track the work on the ansible/awx-operator side too to make _that_ actually happen cc @shanemcd @Spredzy

wenottingham:
Rebased, needs retesting.

Issue #6603, opened 2020-04-06
Add adhoc command form to inventory


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
An adhoc command can be executed against hosts within an inventory.  This form should be accessible after selecting hosts or groups on the following pages:

- [x] Inventory -> Hosts

- [x]  Inventory -> Host -> Groups
- [x]  Inventory -> Groups
- [x] Inventory -> Group -> Hosts
- [x] Smart inventory -> Hosts

There should be a button on each of these lists that is disabled when no rows are selected and enabled when at least one row is selected.

Clicking the button should take you to a standalone form:

<img width="1526" alt="Screen Shot 2020-04-06 at 1 29 59 PM" src="https://user-images.githubusercontent.com/9889020/78587297-bc288680-780a-11ea-89cf-ed8752440f53.png">

Here's a link to the mockup: https://tower-mockups.testing.ansible.com/patternfly/inventories/inventories-groups-run/

The `limit` field should be _pre-populated_ with a comma separated string of the names of all the selected resources from list.  For example, if I selected these 3 hosts from the list [localhost1, localhost2, localhost3], the limit field should look like:

<img width="724" alt="Screen Shot 2020-04-06 at 1 32 41 PM" src="https://user-images.githubusercontent.com/9889020/78587554-1b869680-780b-11ea-932f-fff8e0e6ca49.png">

when the form is initially loaded.

Clicking the launch button should take you to the output page for the adhoc command.

mabashian:
Note that running adhoc commands on the Hosts (/#/hosts) list is not supported in the old application.  It does _seem_ like that's possible based on the OPTIONS response on `/api/v2/ad_hod_commands` - the inventory field is not required.  @wenottingham should we look at supporting running adhoc commands from the `ALL HOSTS` list?  If that's actually possible/something we want to do it'd probably be easy to roll into this work.

dsesami:
This can be closed with the merge of #7905 and also the opening of a stubs PR. I will close it when that is ready.

dsesami:
This was prematurely closed as only Inventory -> Groups was verified. Work needs to be done to hook up the rest of them, as #7905 was mostly a PoC to ensure that it worked on one field first.

tiagodread:
Verified all remaining lists, tests covered by 5602

nixocio:
This issue seems to be related to https://github.com/ansible/awx/issues/612

PR #7905, opened 2020-08-13
Add Ad Hoc Commands

AlexSCorey:
##### SUMMARY
This addresses #6603 for just Inventory Groups List.  It proposes a new workflow for the ad hoc commands so that we treat it more like prompt on launch. Talking with @trahman73 about the error handling of this feature we also introduce a new work flow for errors. We decided that because this feature is now in a wizard on top of a list that all of the content errors(ususally rendered inside a card on a page and not in a modal) would be rendered inside a modal.  The tooltips, while complete, could probably use an extra set of eyes. I modified them a bit, from the classic ui tooltips, so that they could be translated a bit easier and with more context.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION

ad hoc

### DIFF
--- a/awx/ui_next/src/screens/Inventory/InventoryGroups/InventoryGroupsList.jsx
+++ b/awx/ui_next/src/screens/Inventory/InventoryGroups/InventoryGroupsList.jsx
@@ -160,7 +160,7 @@ function InventoryGroupsList({ i18n }) {
         >
           {({ openAdHocCommands }) => (
             <DropdownItem
-              key="add"
+              key="run command"
               onClick={openAdHocCommands}
               isDisabled={groupCount === 0}
             >
@@ -171,6 +171,7 @@ function InventoryGroupsList({ i18n }) {
         <DropdownItem
           variant="danger"
           aria-label={i18n._(t`Delete`)}
+          key="delete"
           onClick={toggleModal}
           isDisabled={selected.length === 0 || selected.some(cannotDelete)}
         >
--- a/awx/ui_next/src/screens/Inventory/InventoryGroups/InventoryGroupsList.jsx
+++ b/awx/ui_next/src/screens/Inventory/InventoryGroups/InventoryGroupsList.jsx
@@ -2,7 +2,13 @@ import React, { useCallback, useState, useEffect } from 'react';
 import { useParams, useLocation } from 'react-router-dom';
 import { withI18n } from '@lingui/react';
 import { t } from '@lingui/macro';
-import { Button, Tooltip } from '@patternfly/react-core';
+import {
+  Button,
+  Tooltip,
+  DropdownItem,
+  ToolbarGroup,
+  ToolbarItem,
+} from '@patternfly/react-core';
 import { getQSConfig, parseQueryString } from '../../../util/qs';
 import useSelected from '../../../util/useSelected';
 import useRequest from '../../../util/useRequest';
@@ -17,6 +23,7 @@ import PaginatedDataList, {
 import InventoryGroupItem from './InventoryGroupItem';
 import InventoryGroupsDeleteModal from '../shared/InventoryGroupsDeleteModal';
 import AdHocCommandsButton from '../../../components/AdHocCommands/AdHocCommands';
+import { Kebabified } from '../../../contexts/Kebabified';
 
 const QS_CONFIG = getQSConfig('group', {
   page: 1,
@@ -141,9 +148,37 @@ function InventoryGroupsList({ i18n }) {
     setSelected([]);
     setIsDeleteLoading(false);
   };
-
   const canAdd =
     actions && Object.prototype.hasOwnProperty.call(actions, 'POST');
+  const kebabedAdditionalControls = () => {
+    return (
+      <>
+        <AdHocCommandsButton
+          adHocItems={selected}
+          apiModule={InventoriesAPI}
+          itemId={parseInt(inventoryId, 10)}
+        >
+          {({ openAdHocCommands }) => (
+            <DropdownItem
+              key="add"
+              onClick={openAdHocCommands}
+              isDisabled={groupCount === 0}
+            >
+              {i18n._(t`Run command`)}
+            </DropdownItem>
+          )}
+        </AdHocCommandsButton>
+        <DropdownItem
+          variant="danger"
+          aria-label={i18n._(t`Delete`)}
+          onClick={toggleModal}
+          isDisabled={selected.length === 0 || selected.some(cannotDelete)}
+        >
+          {i18n._(t`Delete`)}
+        </DropdownItem>
+      </>
+    );
+  };
 
   return (
     <>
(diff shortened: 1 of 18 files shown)


trahman73:
Nice job @AlexSCorey I'll leave my UX review below.

All examples are on the mockups here:
https://tower-mockups.testing.ansible.com/patternfly/inventories/inventories-groups/
https://tower-mockups.testing.ansible.com/patternfly/inventories/inventories-groups-run/
https://tower-mockups.testing.ansible.com/patternfly/inventories/inventories-groups-run-creds/

- Move Run command button after Add button (Delete will be last)
- Sentence casing on the main button to read "Run command" 
- Change the wizard title from "Ad Hoc Commands" to "Run command" to match the action button
- Sentence case the left nav of the wizard to read "Machine credential"
- Remove the Machine credential heading with (?) from the second step of the wizard (the left nav is good enough here)

Thanks and please let me know if you have any questions on the above.

AlexSCorey:
> * Sentence casing on the main button to read "Run command"
> * Remove the Machine credential heading with (?) from the second step of the wizard (the left nav is good enough here)
> 
@trahman73 The main button is currently sentence cased.  As you saying that it should not be sentence cased?  Also, the heading on step 2 is just a standard label heading.  I can remove that, and, they can't launch the job until they select a credential so that asterisk isn't really necessary, but what about the tooltip?

trahman73:
Disregard the main button comment as it is correct. We can remove the heading on the Machine credential step (including the tooltip) and disable launch until a credential is selected.

mabashian:
@AlexSCorey @trahman73 @wenottingham 

Was thinking a bit about the UX of the run command button and wanted to run an idea by y'all.  The current implementation matches up with the old UI in that the button is disabled until something is selected (group, host, etc).  This action pre-fills the limit with whatever has been selected.  I think that's great and we should keep that but I'm now wondering why that button is disabled until you select something.  The `limit` is not a required field so you could run an adhoc command against all the hosts/groups in the list without providing a limit.  Should we make this button enabled by default (maybe as long as there's at least one host/group in the list)?  Clicking the button without first selecting a host/group would just result in an empty limit, which is valid.

mabashian:
@AlexSCorey I think we should preselect a verbosity for the user on initial load.  Looks like this is the one that's selected in the old UI:

Screen Shot 2020-08-17 at 10 11 32 AM

Additionally, I think we should display the verbosity options like we do on the JT form:

Screen Shot 2020-08-17 at 10 14 34 AM

mabashian:
@AlexSCorey this Arguments field actually isn't required:

Screen Shot 2020-08-17 at 10 12 57 AM

mabashian:
@trahman73 @AlexSCorey I'm a little worried about using a tag-based approach for the limit pre-fill.  It feels like a mixture of two different UX flows (tags & an input).  Patterns within the limit are flexible and there's some syntactical sugar that's supported: https://docs.ansible.com/ansible/latest/user_guide/intro_patterns.html#common-patterns.  It looks like a user could still do something like that on the input side of the field but they wouldn't be able to do any modifications to the tagged hosts/groups.  I'm just wondering if it would be easier to understand as a single input.  Pre-filled hosts/groups would be separated by a colon or comma in the input string.

trahman73:
@AlexSCorey I agree with @mabashian about keeping the Run command button active.
We can change the Limit field to a regular text input as per the additional feedback. I will update the mocks as well from the feedback so far.

awx/ui_next/src/screens/Inventory/InventoryGroups/InventoryGroupsList.jsx
```diff
@@ -125,9 +134,37 @@ function InventoryGroupsList({ i18n }) {
...
+    return (
+      <>
+        <AdHocCommandsButton
+          adHocItems={selected}
+          apiModule={InventoriesAPI}
+          itemId={parseInt(inventoryId, 10)}
+        >
+          {({ openAdHocCommands }) => (
+            <DropdownItem
+              key="add"
```
jlmitch5: probably want a different key name

awx/ui_next/src/screens/Inventory/InventoryGroups/InventoryGroupsList.jsx
```diff
@@ -125,9 +134,37 @@ function InventoryGroupsList({ i18n }) {
...
+              onClick={openAdHocCommands}
+              isDisabled={groupCount === 0}
+            >
+              {i18n._(t`Run command`)}
+            </DropdownItem>
+          )}
+        </AdHocCommandsButton>
+        <DropdownItem
+          variant="danger"
```
jlmitch5, line 172: probably want key prop here

jlmitch5 (commented):
the kebabification of the action buttons looks good, just some key prop changes!

jlmitch5 (approved):
looks good

unlikelyzero:
@one-t can you help us identify some test cases that would flesh out the UI.

dsesami (changes requested):
Two issues:

1. In the legacy UI, the tooltip for Parameters changes dynamically based on what module is selected. It provides a link to the ansible documentation. The hyperlink is correct, but the ansible docs currently improperly redirect -- I'm going to file a separate issue for that. We should provide a link to the docs depending on which module is selected, though.
2. Follow these steps:
- Create an inventory with just one host, `localhost`. This is also the default Demo Inventory if you have that.
- Assign that host to Group1
- Launch any command against Group1. The simplest example is `ping`. In the old UI, this succeeds. In the new UI, I receive a FAILED result, giving me a usage message for `ping`, which implies that the command has syntax errors.

Screenshot from 2020-09-01 11-24-28

The same `ping` against the same group works in the old UI:
Screenshot from 2020-09-01 11-24-52

dsesami:
Also, when linking to modules, I've been informed that the old UI uses an outdated link mechanism. Modules links should be formed like this: https://docs.ansible.com/ansible/latest/modules/ping_module.html

AlexSCorey:
> Two issues:
> 
> 1. In the legacy UI, the tooltip for Parameters changes dynamically based on what module is selected. It provides a link to the ansible documentation. The hyperlink is correct, but the ansible docs currently improperly redirect -- I'm going to file a separate issue for that. We should provide a link to the docs depending on which module is selected, though.
> 2. Follow these steps:
> 
> * Create an inventory with just one host, `localhost`. This is also the default Demo Inventory if you have that.
> * Assign that host to Group1
> * Launch any command against Group1. The simplest example is `ping`. In the old UI, this succeeds. In the new UI, I receive a FAILED result, giving me a usage message for `ping`, which implies that the command has syntax errors.
> 
> Screenshot from 2020-09-01 11-24-28
> 
> The same `ping` against the same group works in the old UI:
> Screenshot from 2020-09-01 11-24-52

AlexSCorey:
recheck

dsesami:
this is almost ready -- all that's left is the hyperlinking to work properly in the tooltip for arguments.

dsesami (approved):

dsesami:
regate

dsesami:
recheck

Issue #7885, opened 2020-08-11
Update navigation bar and routing system to add execution environments


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Update navigation bar and routing system to add execution environments.

unlikelyzero:
moving to https://github.com/ansible/awx/issues/7884

Issue #7884, opened 2020-08-11
[UI] Execution environments


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Issue to track the progress of UI work related to execution environments.

~1 - Update navigation bar and routing system to add execution environments. https://github.com/ansible/awx/issues/7885~
~2 - List execution environments - https://github.com/ansible/awx/issues/7886~
~3 - Add/Edit execution environments - https://github.com/ansible/awx/issues/7887~
~4 - Add details execution environments - https://github.com/ansible/awx/issues/8171~
~5 - Add access view for execution environments - https://github.com/ansible/awx/issues/8190~
~6 - Add execution environments view for organizations -  https://github.com/ansible/awx/issues/8210 -> #8337-> #9384~
~7 - Update usage of user_capabilities for execution environments - https://github.com/ansible/awx/issues/8216~
~8 - Add option to select a default global EE on settings - https://github.com/ansible/awx/issues/9088~
~9 - Update EE Add/List/Edit/Details with latest API changes - https://github.com/ansible/awx/issues/9087 ->  #9179~
~10 - Add execution environment field to a few screens - https://github.com/ansible/awx/issues/9189~
~11 -  Remove the custom venv fields from the UI - https://github.com/ansible/awx/issues/9190~
~12 -  Surface missing-resource warnings on list items for UJTs that were using custom virtualenvs - https://github.com/ansible/awx/issues/9207~
~13 - A pull option to EE - https://github.com/ansible/awx/issues/9109~
~14 - User should not modify name and image managed by awx. https://github.com/ansible/awx/issues/9250~
~15 - Add EE to the Activity Stream - https://github.com/ansible/awx/issues/9308~
~16 - Add copy functionality to EE - https://github.com/ansible/awx/issues/9307~
~17 - Re-evaluate RBAC for EE as per latest API changes - https://github.com/ansible/awx/issues/9416~
18 - Filter by container registry credential to be associated with EE - https://github.com/ansible/awx/issues/9628
~19 - Remove EE field from Project screen - https://github.com/ansible/awx/issues/9719~
~20 - Add screen to represent EE unified job templates endpoint - https://github.com/ansible/awx/issues/9723~
21 - Add link to the docs in how to replace a custom virtual env with an EE - https://github.com/ansible/awx/issues/9846
~22 - EE RBAC violation when EE org changes - https://github.com/ansible/awx/issues/9769~
[...]

PR #7909, opened 2020-08-15
Add execution environments files

nixocio:
Update navigation bar and routing system to add execution environments.
Also, add stub files for the remaining related work.

See: https://github.com/ansible/awx/issues/7885
Also: https://github.com/ansible/awx/issues/7884


AlexSCorey (approved):
Looks good to me.

jakemcdermott (approved):

Issue #5684, opened 2020-01-16
Users List - Add Type column


marshmalien:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Mockup: https://tower-mockups.testing.ansible.com/patternfly/users/users/

Add column for user "type" Normal User, System Administrator, System Auditor to the user list.

<img width="1197" alt="Screen Shot 2020-01-16 at 4 45 48 PM" src="https://user-images.githubusercontent.com/15881645/72565384-b9f40300-387f-11ea-9ab8-d0aeb1c198e2.png">

dsesami:
This column appears normally now. :+1: and is already stubbed in users-crud-spec

PR #7910, opened 2020-08-16
Add type column to users list

nixocio:
Add type column to users list. Also, update `UserListItem` to be a
functional component.

See: https://github.com/ansible/awx/issues/5684

image


jakemcdermott:
recheck

jakemcdermott (approved):

nixocio:
recheck

PR #7911, opened 2020-08-16
Add Remote Archive SCM Type to support using artifacts and releases as projects

philipsd6:
##### SUMMARY

Some users need to provide projects based on a build process that produces a versioned artifact, or release, containing all the requirements for that project in a single archive.

This PR provides a new SCM type "Remote Archive" that accepts an URL to a remote archive, such as a [GitHub Release](https://github.com/philipsd6/test-playbook/archive/v0.0.1.zip) or a build artifact stored in [Artifactory](https://jfrog.com/artifactory/) and unpacks it into the project path for use. 

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - UI

##### AWX VERSION

```
awx: 14.0.0
```

##### ADDITIONAL INFORMATION

- An important design choice was to ensure that an archive that contains a single root directory (most well-formed archives unpack into a versioned directory) is skipped in the unpacking, as otherwise changing the SCM URL to point to a new archive would cause Job Template references to playbooks to be broken. 
- I chose to implement the core as an action plugin, as implementing the feature in pure Ansible was proving impossible due to the need to support `scm_clean` and stripping leading components.
- The checksum of the downloaded archive is used as the repository version
- Note: Since this feature is intended to support the concept of unchanging artifacts, it might be advisable to disable galaxy integration (for roles, at least) though I did not implement that.


philipsd6:
recheck

philipsd6:
All tests are successful now; The non-voting awx-detect-schema-change test is correctly noting the schema change due to the addition of `archive` as an scm type.

awx/main/migrations/0118_v400_add_remote_archive_scm_type.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.2.11 on 2020-08-18 22:35
```
ryanpetrello: We recently stopped embedding downstream versions in these filenames (because it's a remnant of how we managed migrations in very old version of Django) - can you remove the `v400` ?

ryanpetrello:
@philipsd6 wow, this is quite a pull request 😄 - you seem to have covered _everything_ from the backend to frontend!

Thanks for putting this together.  Would you mind filing an issue as a feature request that outlines and describes this use case/feature?  That way we can discuss it and track progress there (instead of strictly on this PR).

awx/main/migrations/0118_v400_add_remote_archive_scm_type.py
philipsd6: I wasn't sure how to create the migration, since there doesn't seem to be any documentation about this in the contribution guide, so I was trying to match existing, but yes, I'll change it!

awx/ui/client/src/projects/add/projects-add.controller.js
```diff
@@ -170,6 +170,14 @@ export default ['$scope', '$location', '$stateParams', 'GenerateForm',
                             $scope.lookupType = 'scm_credential';
                             $scope.scmBranchLabel = i18n._('SCM Branch/Tag/Revision');
                             break;
+                        case 'archive':
+                            $scope.credentialLabel = "SCM " + i18n._("Credential");
+                            $scope.urlPopover = '<p>' + i18n._('Example URLs for Remote Archive SCM include:') + '</p>' +
+                                '<ul class=\"no-bullets\"><li>https://github.com/philipsd6/test-playbook/archive/v0.0.1.tar.gz</li>' +
+                                '<li>http://github.com/philipsd6/test-playbook/archive/v0.0.2.zip</li></ul>';
```
wenottingham: I get the idea here, but we'd need to replace this with either a generic example URL that doesn't actually exist, or an upstream URL/reference.

awx/playbooks/project_update.yml
```diff
@@ -101,6 +101,50 @@
...
+        - name: Get archive from url
+          get_url:
+            url: "{{ scm_url|quote }}"
+            dest: "{{ project_path|quote }}/.archive/"
+            url_username: "{{ scm_username|default(omit) }}"
+            url_password: "{{ scm_password|default(omit) }}"
+            force_basic_auth: true
+          register: get_archive
+
+        - name: Unpack archive
```
matburt, line 119: Given the frequency by which updates can run and the impact from unarchiving we could check the checksum at this step... compare it to the current scm version and then NOT do this or run the archiving if they are unchanged.

matburt (changes requested):
I have one small request on using the checksum to not repeat the unarchive operation. This looks great and is a great contribution. Needs docs and QE signoff.

wenottingham:
> Note: Since this feature is intended to support the concept of unchanging artifacts, it might be advisable to disable galaxy integration (for roles, at least) though I did not implement that.

I think it's fine to keep galaxy role/collection downloads with this idea - that allows for separation of artifacts for reuse and not having to bundle everything into each project artifact. (Those content downloads can still be disabled globally).

One comment on this is that collections are growing more standardized support for including playbooks. When that time comes, we'd expect to have a collection as a project source, which will likely require some changes/adjustments here - if collections-as-projects are implemented separately, you'll have two *very* similar infrastructures side by side.

awx/playbooks/project_update.yml
philipsd6: The unarchive step only runs if the get_archive has changed, which should cover this requirement.

awx/ui/client/src/projects/add/projects-add.controller.js
philipsd6: OK, how about I just update these to `s/philipsd6/username/` to match the style in the Mercurial sample.

philipsd6:
@wenottingham keeping galaxy role / collection downloads enabled is fine -- if necessary gatekeeping can be managed on the SDL side by disallowing any `requirements.yml` files to be present. 

Regarding the collection as project source idea -- I was thinking along the same lines, although this concept of mine has been on my back burner since before that became a thing. Right now collections are still tied to Galaxy, and if we tied this feature to collections only, we would still have to ensure that we could pull an internally build collection from alternative stores, such as our Artifactory. 

To be more compatible with this potential future though, should we switch to calling this "collection" instead of "archive"?

awx/ui/client/src/projects/add/projects-add.controller.js
philipsd6: I updated these URLS to be:
* https://github.com/username/project/archive/v0.0.1.tar.gz
* http://github.com/username/project/archive/v0.0.2.zip

awx/playbooks/project_update.yml
philipsd6: I think the main issue with the project update frequency, is that `get_url` will "download the file every time and replace the file if the contents change."

So the archive will always be downloaded, but only unpacked if it changes. I expect that downloading is the heavier operation, but there's no avoiding that unless we increase the complexity by trying to handle etag or last-modified headers from the server.

@matburt do you think I should try to implement etag and last-modified support, or shall we just mark this conversation as resolved?

wenottingham:
As long as it's downloading an arbitrary archive, and not limited to things specifically in collection format, then we shouldn't call them collections.

philipsd6:
> As long as it's downloading an arbitrary archive, and not limited to things specifically in collection format, then we shouldn't call them collections.

Collections should eventually encompass all our requirements for immutable artifacts. The main issue comes down to whether we are forced to use a Galaxy (or Redhat Automation Hub) server or if we can use arbitrary sources for collection archives. So for now, I would agree that we keep this named "archive" and then explore the idea of implementing this concept using Collections separately. 

But if this PR is accepted, I think the documentation for it will show creating an archive in a collection-style format as the source for the project.

awx/ui/client/src/projects/add/projects-add.controller.js
wenottingham: That works for me.

awx/playbooks/project_update.yml
matburt: I think it's fine to resolve this, there's no guarantee that those properties will exist on the remote content server.

matburt (approved):

ryanpetrello:
First pass, I gave this a try and it looks like it works pretty well:

image
image
image

```shell
~ curl -sLO "https://github.com/ansible/awx/archive/14.0.0.tar.gz"
~ sha1sum 14.0.0.tar.gz
927ca47df097b6761ce6d95de355c77fa22bac37  14.0.0.tar.gz
```

Playbooks are properly discovered in tarballs:

image

I'm gonna run some additional tests (and maybe with some new ones) before putting a 👍  on this.

awx/playbooks/action_plugins/project_archive.py
```diff
@@ -0,0 +1,82 @@
...
+
+            if is_dir:
+                os.makedirs(dest, exist_ok=True)
+            else:
+                try:
+                    member_f = archive.open(member)
+                except TypeError:
+                    member_f = tarfile.ExFileObject(archive, member)
+
+                with open(dest, "wb") as f:
```
chrismeyersfsu, line 75: This _seems_ like it's not vulnerable to the known `tarfile` attacks .. but I just can't say for sure https://bugs.python.org/issue21109

ryanpetrello:
@philipsd6 would you also add a line to the changelog for the next release (14.1.0)?

see: https://github.com/ansible/awx/blob/devel/CHANGELOG.md

Thanks!

awx/playbooks/action_plugins/project_archive.py
philipsd6: I am 99% sure it's not vulnerable, as the file in the archive is opened for reading only, and the destination filename is always prefixed with the project_path on #L57, and currently, the plugin only handles directories and files, *not* symlinks.

jakemcdermott (approved):
Frontend lgtm @philipsd6 - thanks!

awx/playbooks/action_plugins/project_archive.py
philipsd6: ...though not handling symlinks might be a problem if an archive contains a project with symlinks in var files, like:
```
rhel7.yml
rhel8.yml -> rhel7.yml
```
Do you think it would make sense to implement support for this? If so, I'll ensure the resulting code isn't vulnerable to a tarfile attack.

But maybe it's better to enhance this feature in a second pass later.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ This is a list of high-level changes for each release of AWX. A full list of com
 - Fixed a bug in the AWX CLI that prevented Workflow nodes from importing properly - https://github.com/ansible/awx/issues/7793
 - Fixed a bug in the awx.awx collection release process that templated the wrong version - https://github.com/ansible/awx/issues/7870
 - Fixed a bug that caused errors rendering stdout that contained UTF-16 surrogate pairs - https://github.com/ansible/awx/pull/7918
+-  Added the Remote Archive SCM Type to support using immutable artifacts and releases as projects - https://github.com/ansible/awx/pull/7911
```
ryanpetrello: Minor nitpick 😄 

Let's put this at the top, because it's a new feature:

```diff
diff --git a/CHANGELOG.md b/CHANGELOG.md
index 648e4adbf3..3d6a657365 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This is a list of high-level changes for each release of AWX. A full list of com

 ## 14.1.0 (TBD)
 - AWX images can now be built on ARM64 - https://github.com/ansible/awx/pull/7607
+- Added the Remote Archive SCM Type to support using immutable artifacts and releases (such as tarballs and zip files) as projects - https://github.com/ansible/awx/issues/7954
 - Deprecated official support for Mercurial-based project updates - https://github.com/ansible/awx/issues/7932
 - Added resource import/export support to the official AWX collection - https://github.com/ansible/awx/issues/7329
 - Added the ability to import YAML-based resources (instead of just JSON) when using the AWX CLI - https://github.com/ansible/awx/pull/7808
@@ -13,7 +14,6 @@ This is a list of high-level changes for each release of AWX. A full list of com
 - Fixed a bug in the AWX CLI that prevented Workflow nodes from importing properly - https://github.com/ansible/awx/issues/7793
 - Fixed a bug in the awx.awx collection release process that templated the wrong version - https://github.com/ansible/awx/issues/7870
 - Fixed a bug that caused errors rendering stdout that contained UTF-16 surrogate pairs - https://github.com/ansible/awx/pull/7918
--  Added the Remote Archive SCM Type to support using immutable artifacts and releases as projects - https://github.com/ansible/awx/pull/7911
```

philipsd6: Awww... makes me feel extra special ;)

philipsd6: Done!

awx/ui/client/src/projects/add/projects-add.controller.js
```diff
@@ -170,6 +170,14 @@ export default ['$scope', '$location', '$stateParams', 'GenerateForm',
                             $scope.lookupType = 'scm_credential';
                             $scope.scmBranchLabel = i18n._('SCM Branch/Tag/Revision');
                             break;
+                        case 'archive':
+                            $scope.credentialLabel = "SCM " + i18n._("Credential");
+                            $scope.urlPopover = '<p>' + i18n._('Example URLs for Remote Archive SCM include:') + '</p>' +
```
ryanpetrello, line 175: cc @jakemcdermott 

Do we want a credential at all for these?  What would that credential even be?  I could *maybe* see some argument for HTTP-like (`Authorization` headers?) credentials, but I don't think it's a requirement on the first pass here, and it's kind of confusing that we ask for this in the UI (because I don't think setting a credential here actually *does* anything).

ryanpetrello:
@philipsd6 Could you apply a diff like this as well?  The UI prevents this, but we also don't want people setting this for archive type projects.

```diff
diff --git a/awx/api/serializers.py b/awx/api/serializers.py
index be6a9d640b..429bf512be 100644
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -1336,6 +1336,8 @@ class ProjectOptionsSerializer(BaseSerializer):
             attrs.pop('local_path', None)
         if 'local_path' in attrs and attrs['local_path'] not in valid_local_paths:
             errors['local_path'] = _('This path is already being used by another manual project.')
+        if attrs.get('scm_branch') and scm_type == 'archive':
+            errors['scm_branch'] = _('SCM branch cannot be used with archive projects.')
         if attrs.get('scm_refspec') and scm_type != 'git':
             errors['scm_refspec'] = _('SCM refspec can only be used with git projects.')
```

image

awx/ui/client/src/projects/add/projects-add.controller.js
philipsd6: Actually, I pass the `scm_username` and `scm_password` on to the `get_url` module, in case authentication is required for downloading the archive, which could be the case in an Artifactory use case. Don't those vars come from the credential?

wenottingham: Yes, we'd likely need the username/password for basic auth scenarios.

ryanpetrello: I'm not sure this is *necessary* on the first pass of this PR, personally - I think it's reasonable to just say on the first pass this is a publicly reachable URL.

(or at least, if we implemented credential support, I'd say stick it in another PR)

wenottingham: This PR already includes credential support, see the get_url task ? It's just piggybacking on the existing SCM credentials

ryanpetrello: ...

You are totally right.  I blame Friday brain.

Okay, I think this is fine in its current state on a first pass (if somebody wants e.g., oauth2 tokens, we can add it later).  I'll test user+pass as well.

ryanpetrello: @philipsd6,

I just tested this, and I don't think these extra vars are being passed through to the underlying playbook, so it doesn't work.  I had to make this change to make it work for me:

```diff
diff --git a/awx/main/tasks.py b/awx/main/tasks.py
index 06c740c129..3e70f93c19 100644
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -2105,7 +2105,7 @@ class RunProjectUpdate(BaseTask):
                     scm_username = False
             elif scm_url_parts.scheme.endswith('ssh'):
                 scm_password = False
-            elif scm_type == 'insights':
+            elif scm_type in ('insights', 'archive'):
                 extra_vars['scm_username'] = scm_username
                 extra_vars['scm_password'] = scm_password
             scm_url = update_scm_url(scm_type, scm_url, scm_username,
```

Could you add a commit like this one?

philipsd6: How about we leave the support for Basic Authentication here as is, as it's documented in the playbook for supporting SVN and Insights?

philipsd6: Friday brain indeed! Replied before I reloaded... let me look into your suggestion and update the PR over the weekend.

philipsd6: @ryanpetrello thanks for catching this, I've updated the PR to ensure the credentials are passed for the archive scm type.

ryanpetrello (approved):
This looks great - thanks for addressing all the feedback!

I'm going to run this through a bit more automated testing I have, then merge it 👍

ryanpetrello:
regate

PR #7918, opened 2020-08-17
make event stdout encoding more resilient to UTF-16 surrogate pairs

ryanpetrello:
Given that we're returning literally any bytestream that Ansible can muster, it probably makes sense to relax `errors='strict'` a bit to avoid 500 errors.

see: https://unicodebook.readthedocs.io/unicode_encodings.html#utf-16-surrogate-pairs

Run this adhoc command w/ `-m shell`:

`printf '\uD83D\uDE00'`

image

after (better than a 500 error):

image

### DIFF
--- a/awx/api/renderers.py
+++ b/awx/api/renderers.py
@@ -7,6 +7,24 @@ from prometheus_client.parser import text_string_to_metric_families
 # Django REST Framework
 from rest_framework import renderers
 from rest_framework.request import override_method
+from rest_framework.utils import encoders
+
+
+class SurrogateEncoder(encoders.JSONEncoder):
+
+    def encode(self, obj):
+        ret = super(SurrogateEncoder, self).encode(obj)
+        try:
+            ret.encode()
+        except UnicodeEncodeError as e:
+            if 'surrogates not allowed' in e.reason:
+                ret = ret.encode('utf-8', 'replace').decode()
+        return ret
+
+
+class DefaultJSONRenderer(renderers.JSONRenderer):
+
+    encoder_class = SurrogateEncoder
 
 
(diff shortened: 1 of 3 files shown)


ryanpetrello:
recheck

jakemcdermott (approved):

ryanpetrello:
There's also apparently libraries for actually handling these more gracefully: https://pypi.org/project/surrogates/

...but I also don't know that I want to add a dependency to deal with something so esoteric.

awx/api/renderers.py
```diff
@@ -7,6 +7,24 @@
 # Django REST Framework
 from rest_framework import renderers
 from rest_framework.request import override_method
+from rest_framework.utils import encoders
+
+
+class SurrogateEncoder(encoders.JSONEncoder):
+
+    def encode(self, obj):
+        ret = super(SurrogateEncoder, self).encode(obj)
```
ryanpetrello, line 16: This is a little gross, but I couldn't find a less gross way to do this.

tl;dr there’s code further down the stack in DRF that ends up calling `.encode()` (and it blows up), so I’m calling it, seeing if it blows up, and then fixing the problem (so it won’t blow up later)

awx/api/renderers.py
```diff
@@ -7,6 +7,24 @@
...
+        try:
+            ret.encode()
+        except UnicodeEncodeError as e:
+            if 'surrogates not allowed' in e.reason:
```
jakemcdermott, line 20: since `ret` is just a python str this assumption lgtm :+1:

ryanpetrello:
recheck

ryanpetrello:
recheck

Issue #6172, opened 2020-03-04
Add Variables Field to Schedule Details


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Link to mockups: https://tower-mockups.testing.ansible.com/patternfly/templates/templates-schedules-detail/

I peeled this out of https://github.com/ansible/awx/pull/6170 because I think it's going to require us to extend the codemirror components to support an initial value of `{}`.  I believe those components are currently expecting a string.  I didn't want to muddy that initial PR with these changes.

dsesami:
Looks great!

unlikelyzero:
Covered by 4560

PR #7921, opened 2020-08-17
Adds extra variables to schedule details

mabashian:
##### SUMMARY
link #6172 

The `extra_data` field (extra vars) on a schedule is a JSON blob rather than a string like some of the other models.  I needed to make some modifications to `VariablesDetail` in order to account for this.  If we detect that the value passed in is in fact JSON (not stringified JSON) then it should now handle converting that JSON to it's stringified version (which is what codemirror expects).

I also made some modifications to the conditional render params on the prompt fields.  We're now checking whether or not the field is promptable (from the launch endpoint) and using that to determine whether or not a field should be shown in the details.  The scenario that we're addressing by doing this is if a field _was_ promptable when the schedule was created but no longer is.  We don't want to show the prompted value since it won't be used when the schedule runs.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/components/CodeMirrorInput/VariablesDetail.jsx
+++ b/awx/ui_next/src/components/CodeMirrorInput/VariablesDetail.jsx
@@ -1,10 +1,15 @@
 import 'styled-components/macro';
 import React, { useState, useEffect } from 'react';
-import { string, node, number } from 'prop-types';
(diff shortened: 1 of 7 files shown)


jlmitch5 (approved):
looks good!

jakemcdermott (approved):

PR #7925, opened 2020-08-18
Fix bug tower_project checks creds on removal

### LINKED ISSUE #7924: tower_project should not check for credentials on state=absent
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
awx.awx.tower_project checks for credential on state == absent, and it shouldn't care if we're removing it (eg credential gets removed before this is called)

##### ENVIRONMENT
* AWX version: 3.7.1
* AWX install method: Ansible Tower 3.7.1 - setup.sh
* Ansible version:  2.9.1
* Operating System: Red Hat 7.9
* Web Browser: Chrome

##### STEPS TO REPRODUCE

Create a project with a credential
Remove the credential before removing the project

##### EXPECTED RESULTS

Project should be removed

##### ACTUAL RESULTS

Error is received when trying to remove it.
fatal: [awx-node01]: FAILED! => {"changed": false, "msg": "The credentials git_creds was not found on the Tower server"}

DLV111:
##### SUMMARY
This fixes #7924 
On removal of a project checks are done to see if the credentials exists or not - we shouldn't care on removal, so only perform these checks when the state is not absent.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION

Unknown - this is using Ansible Tower 3.7.1
```

```

##### ADDITIONAL INFORMATION
See linked bug/issue #7924 

Before
```
TASK [IS.compute.ansible_tower_administration : tower_project | Add tower project] ***************************************************************************************************************
fatal: [awx-node01]: FAILED! => {"changed": false, "msg": "The credentials git_creds was not found on the Tower server"}

```
After

```
TASK [IS.compute.ansible_tower_administration : tower_project | Add tower project] ***************************************************************************************************************
ok: [awx-node01]
```


DLV111:
I believe the API failures are known based on looking at other PR's in flight. Not sure if this is the right solution, but it works for me so figured i'd send it through and see what happens!

AlanCoding (approved):

john-westcott-iv (changes requested):
Thanks for the contribution! The pattern that we established for doing non-essential lookups is to move them under the delete if needed call. If you look at tower_job_template you will see that we lookup the org first (because its required to lookup the template), then lookup the template and then call delete if state is absent. Later on we do the lookups on inventory/project/etc. Can you modify this PR to match that pattern?

AlanCoding:
^ that should be fairly simple to act on

DLV111:
Hey - i had a look at this, and i'm not 100% sure on how to implement - the scm_credential is needed for the creation of the project - but isn't required for a delete. The current issue is that it's being looked up on a delete. I can't really move it to delete_if_needed as we don't want to remove the credential, we only want to ensure we don't check its there on removal - as it may have been deleted, or used by other areas of tower.

I hit this as we create all these items via the API in Tower 3.7.1, and the removal happen in the same order, so the credentials are removed before the project is deleted.

john-westcott-iv:
To follow the existing pattern, instead of doing this:
`if credential is not None and state != 'absent':`

Move the original `if credential is not None` below the `delete_if_needed` call like:
```
    # Attempt to look up the related items the user specified (these will fail the module if not found)
    org_id = module.resolve_name_to_id('organizations', organization)

    # Attempt to look up project based on the provided name and org ID
    project = module.get_one('projects', **{
        'data': {
            'name': name,
            'organization': org_id
        }
    })

    if state == 'absent':
        module.delete_if_needed(project)

    # Attempt to look up the related items the user specified (these will fail the module if not found)
    if credential is not None and state != 'absent':
        credential = module.resolve_name_to_id('credentials', credential)
```

This makes it so the credential is not looked up until after we decide to delete it.

We are attempting to follow a pattern so that (eventually) we may be able to use Ansible itself to generate these modules automatically (see awx_collections/tools/generate.yml).

ryanpetrello:
@dlevene1 do you mind rebasing and updating this PR?

Thanks.

ryanpetrello:
Actually @AlanCoding do we need this PR anymore given https://github.com/ansible/awx/pull/8109/files#r485843956?

Are we ready to verify https://github.com/ansible/awx/issues/7924 ?

AlanCoding:
Yes, we applied the standard practice to this case. Thanks.

Issue #7801, opened 2020-08-03
Associate instances to an instance group


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Associate instances to an instance group.
See: https://tower-mockups.testing.ansible.com/patternfly/instance-groups/instance-groups-instances/

PR #7929, opened 2020-08-18
Associate instances to instance groups

nixocio:
Associate instances to instance groups.

See: https://github.com/ansible/awx/issues/7801
associate_updated


nixocio:
I have been working on this feature in the last couple of days. I would like get feedback.

I spent sometime adding a work around for the slider for the forks. Then I decided to wait for the slider to be in PF in order to implement this part of this feature.

https://tower-mockups.testing.ansible.com/patternfly/instance-groups/instance-groups-instances/

See: https://github.com/patternfly/patternfly/issues/296

awx/ui_next/src/screens/InstanceGroup/Instances/InstanceListItem.jsx
```diff
@@ -0,0 +1,139 @@
...
+      return (
+        <Progress
+          value={100 - item.percent_capacity_remaining}
+          measureLocation={ProgressMeasureLocation.top}
+          size={ProgressSize.sm}
+          title={i18n._(t`Used capacity`)}
+        />
+      );
+    }
+    return <span css="color: red">{i18n._(t`Unavailable`)}</span>;
```
keithjgrant: this should use a patternfly CSS variable instead of `red` (`color: var(--pf-global--danger-color--200);`) https://pf4.patternfly.org/documentation/overview/global-css-variables

awx/ui_next/src/components/PaginatedDataList/ToolbarAddButton.jsx
```diff
@@ -6,14 +6,23 @@ import { withI18n } from '@lingui/react';
 import { t } from '@lingui/macro';
 import { useKebabifiedMenu } from '../../contexts/Kebabified';
 
-function ToolbarAddButton({ linkTo, onClick, i18n, isDisabled }) {
+function ToolbarAddButton({
+  linkTo,
+  onClick,
+  i18n,
+  isDisabled,
+  defaultLabel = '',
```
AlexSCorey: rather than setting the default to an empty string here, set it to `i18n.(t`Add`)` and you can remove the code on line 24

awx/ui_next/src/components/PaginatedDataList/ToolbarAddButton.jsx
```diff
@@ -23,28 +32,28 @@ function ToolbarAddButton({ linkTo, onClick, i18n, isDisabled }) {
         to={linkTo}
         onClick={!onClick ? undefined : onClick}
       >
-        {i18n._(t`Add`)}
+        {label}
```
AlexSCorey: this becomes `{defaultLabel}` as well as the other labels for the rest of the file

AlexSCorey (changes requested):

awx/ui_next/src/components/PaginatedDataList/ToolbarAddButton.jsx
nixocio: Thanks for your review @AlexSCorey, updated per your suggestions.

AlexSCorey (approved):
Looks good.  I had an idea to change where `canAdd` is defined a bit, but I'll bring that up offline to see what the group says.

dsesami:
I suspect that the AWX instance group should not allow disassociation. can you verify this? 
Screenshot from 2020-08-26 12-04-11

nixocio:
@dsesami, that is the current implementation. The only instance that can not be associated/disassociated currently is an `isolated` instance.

dsesami:
we've identified that instance group disassociation does not behave properly with isolated nodes, so this is going back into review.

nixocio:
Thanks for looking into that @dsesami.

nixocio:
After re-testing this isolated instances can not be associated/disassociated with the changes introduced by this PR, this is the same behavior in the current ui.

image

image

dsesami (approved):
Looks good. I've identified an issue with resource utilization being marked as `Unavailable`, but that is unrelated to this PR and was discussed with @nixocio already.

Issue #7756, opened 2020-07-29
Creating a new user in ui-next does not associate it with an organization


dsesami:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Creating a user and associating it with an org in the add form does not actually attach the user to the org. It is also probably important to check the edit form when fixing this.

##### STEPS TO REPRODUCE
1. Create an organization.
2. Go to the user creation form.
3. Attempt to associate the user with the org when creating.

##### EXPECTED RESULTS
1. The user appears in the org access list as a member
2. The org appears in the org list for the user

##### ACTUAL RESULTS
No association occurs.

dsesami:
fixed with the merge of #7933

PR #7933, opened 2020-08-19
Use organization api to create users

jakemcdermott:
##### SUMMARY
for #7756 

* Use organization api to create users. This ensures that the user will be related to the chosen organization when it is created.
* Also makes user organization names bold to match other similar lists


dsesami (approved):

dsesami:
tests passed. thanks!

PR #7937, opened 2020-08-19
clean up old authtoken support

ryanpetrello:
just use Bearer tokens - those are the only type of tokens we support

### DIFF
--- a/awxkit/awxkit/api/client.py
+++ b/awxkit/awxkit/api/client.py
@@ -15,12 +15,11 @@ class ConnectionException(exc.Common):
 
 
 class Token_Auth(requests.auth.AuthBase):
-    def __init__(self, token, auth_type='Token'):
+    def __init__(self, token):
         self.token = token
-        self.auth_type = auth_type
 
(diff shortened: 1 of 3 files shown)


beeankha (approved):

awxkit/awxkit/api/client.py
```diff
@@ -57,7 +56,7 @@ def login(self, username=None, password=None, token=None, **kwargs):
             else:
                 self.session.auth = (username, password)
         elif token:
-            self.session.auth = Token_Auth(token, auth_type=kwargs.get('auth_type', 'Token'))
```
ryanpetrello, line 60: This default is wrong, and doesn't even work anymore.

See: https://github.com/ansible/awx/commit/28512e042b4dfe16ef75d39aea79f2d82459b3a9#diff-763ee29332475dd80b6d51f1eeb8a2aeL260-L262

kdelee (approved):
:1st_place_medal:

john-westcott-iv (approved):

Issue #7157, opened 2020-05-26
Add status from the websockets for JobDetail


nixocio:
This issue is a place-holder to be explored,  related to the comment below:

https://github.com/ansible/awx/blob/d42f57d726c74d1f74c7114f522e15fc72d706a4/awx/ui_next/src/screens/Job/JobDetail/JobDetail.jsx#L126

mabashian:
Yea this is still an issue - the status icon doesn't update dynamically on the job output page

unlikelyzero:
Can close alongside the merge of 5501

nixocio:
This issue was addressed by https://github.com/ansible/awx/pull/8726

unlikelyzero:
https://github.com/ansible/awx/pull/8726#issuecomment-772746083

Issue #6616, opened 2020-04-06
Dynamically update workflow results interface based on websocket messages


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
When a user is on a workflow results page:

<img width="1638" alt="Screen Shot 2020-04-06 at 4 04 50 PM" src="https://user-images.githubusercontent.com/9889020/78600131-5d6e0780-7820-11ea-9120-29a047677924.png">

We want to update the status of jobs on nodes in real time by listening to messages that come across the websocket.

Hopefully by the time someone picks this up, we'll have a websocket implementation in place so the actually subscription process will be trivial.

The code that handles this in the old app is found here: https://github.com/ansible/awx/blob/devel/awx/ui/client/src/workflow-results/workflow-results.controller.js#L264

The gist of it is that we listen for job update messages.  When we get one, we check to see whether or not it belongs to the workflow we're viewing.  If it does, we go out and update the job reference on the node to reflect the new status and redraw the graph.

A user should be able to watch the entire workflow run in real time without any refreshes of the page.

mabashian:
Note that as part of this work we'll want to update the status of the outer workflow job as well:

<img width="1638" alt="Screen Shot 2020-04-06 at 4 04 50 PM" src="https://user-images.githubusercontent.com/9889020/78600554-17657380-7821-11ea-9bc7-6d11c8977cca.png">

When the job actually finishes, we may be issuing a refresh of the job from the api to get all the updated fields that become available after the job finishes.

jakemcdermott:
related, because this is job status subscription https://github.com/ansible/awx/issues/5885

Issue #5501, opened 2019-12-13
Kubernetes/Openshift Ansible installer is not working on RHEL 7


megabreit:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
The Kubernetes/Openshift Ansible installer does not work on RHEL 7

##### ENVIRONMENT
* AWX version: 9.0.1
* AWX install method: openshift
* Ansible version:  2.8.7 (from Redhat repository rhel-7-server-ansible-2.8-rpms)
* Operating System: RHEL 7.7
* Web Browser: none
* Jinja2 version: python-jinja2-2.7.2-4.el7.noarch

##### STEPS TO REPRODUCE

oc login
ansible-playbook -i inventory install.yml -e openshift_token=$(oc whoami -t) -e openshift_user=clusteradminuser -vvv
Taget cluster is Openshift 3.11, oc binary is 3.11 too.

##### EXPECTED RESULTS

successfull installation

##### ACTUAL RESULTS

Playbook terminates:
```
TASK [kubernetes : Find cluster for current context] ********************************************************************************************************************
task path: /home/XXXX/src/awx/installer/roles/kubernetes/tasks/main.yml:203
The full traceback is:
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/ansible/executor/task_executor.py", line 145, in run
    res = self._execute()
  File "/usr/lib/python2.7/site-packages/ansible/executor/task_executor.py", line 592, in _execute
    self._task.post_validate(templar=templar)
  File "/usr/lib/python2.7/site-packages/ansible/playbook/task.py", line 268, in post_validate
    super(Task, self).post_validate(templar)
  File "/usr/lib/python2.7/site-packages/ansible/playbook/base.py", line 435, in post_validate
    value = templar.template(getattr(self, name))
  File "/usr/lib/python2.7/site-packages/ansible/template/__init__.py", line 584, in template
    disable_lookups=disable_lookups,
  File "/usr/lib/python2.7/site-packages/ansible/template/__init__.py", line 539, in template
    disable_lookups=disable_lookups,
  File "/usr/lib/python2.7/site-packages/ansible/template/__init__.py", line 804, in do_template
    res = j2_concat(rf)
  File "<template>", line 11, in root
  File "/usr/lib/python2.7/site-packages/jinja2/filters.py", line 740, in do_list
    return list(value)
  File "/usr/lib/python2.7/site-packages/jinja2/filters.py", line 931, in _select_or_reject
[...]

PR #7940, opened 2020-08-19
Update job status and workflow node job status based on websocket events

mabashian:
##### SUMMARY
link #6616
link https://github.com/ansible/awx/issues/7157

This PR does a couple of things:

1) Workflow results will now update (job status) in real time based on consuming websocket messages.
workflow_sockets
2) When a workflow _finishes_ running, the nodes are refreshed via the API so that we can display all data that comes available after a job finishes.
3) The status of any job (playbook, project sync, etc) will now update in real time based on consuming websocket messages.
playbook_sockets
4) Once a job _finishes_ running, the job is refreshed via the API so that we can display all data that comes available after a job finishes.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/components/Workflow/WorkflowNodeHelp.test.jsx
+++ b/awx/ui_next/src/components/Workflow/WorkflowNodeHelp.test.jsx
@@ -9,11 +9,15 @@ describe('WorkflowNodeHelp', () => {
   });
   test('renders the expected content for a completed job template job', () => {
     const node = {
-      job: {
-        name: 'Foo Job Template',
-        elapsed: 9000,
-        status: 'successful',
-        type: 'job',
+      originalNodeObject: {
+        summary_fields: {
+          job: {
+            name: 'Foo Job Template',
+            elapsed: 9000,
+            status: 'successful',
+            type: 'job',
+          },
+        },
       },
       unifiedJobTemplate: {
         name: 'Foo Job Template',
--- a/awx/ui_next/src/screens/Job/useWsJob.js
+++ b/awx/ui_next/src/screens/Job/useWsJob.js
@@ -1,0 +1,53 @@
+import { useState, useEffect } from 'react';
+import { useParams } from 'react-router-dom';
+import useWebsocket from '../../util/useWebsocket';
+import { JobsAPI } from '../../api';
+
+export default function useWsJob(initialJob) {
+  const { type } = useParams();
+  const [job, setJob] = useState(initialJob);
+  const lastMessage = useWebsocket({
+    jobs: ['status_changed'],
+    control: ['limit_reached_1'],
+  });
+
+  useEffect(() => {
+    setJob(initialJob);
+  }, [initialJob]);
+
+  useEffect(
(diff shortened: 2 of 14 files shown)


mabashian:
Note to self - remove this line: https://github.com/ansible/awx/blob/d42f57d726c74d1f74c7114f522e15fc72d706a4/awx/ui_next/src/screens/Job/JobDetail/JobDetail.jsx#L126

awx/ui_next/src/components/Workflow/WorkflowNodeHelp.test.jsx
```diff
@@ -9,11 +9,15 @@ describe('WorkflowNodeHelp', () => {
...
-      job: {
-        name: 'Foo Job Template',
-        elapsed: 9000,
-        status: 'successful',
-        type: 'job',
+      originalNodeObject: {
+        summary_fields: {
+          job: {
+            name: 'Foo Job Template',
+            elapsed: 9000,
```
jakemcdermott, line 16:
```suggestion
            elapsed: 9001,
```

awx/ui_next/src/screens/Job/useWsJob.js
```diff
@@ -0,0 +1,51 @@
...
+        setJob(data);
+      }
+
+      if (!job || lastMessage?.unified_job_id !== job.id) {
+        return;
+      }
+
+      if (
+        ['successful', 'failed', 'error', 'cancelled'].includes(
+          lastMessage.status
```
jakemcdermott, line 31: > Once a job finishes running, the job is refreshed via the API so that we can display all data that comes available after a job finishes.

I like the approach of adding the auto-reload to the hook in order to fetch the summary data when a final status event is received. :+1: 

The summary statistics may not be correct until the job details endpoint returns `"event_processing_finished": true,`.  Depending on the scale of the job, this can happen _long_ after the final job status message is received. 

To detect when event processing is actually finished, we might need to enter a retry loop and poll the api until the `event_processing_finished` flag is `true` or some reasonable retry limit is reached.

jakemcdermott (approved):
LGTM. I would like to see a retry loop added for `event_processing_finished` (see: https://github.com/ansible/awx/pull/7940/files#r474040690) but doesn't necessarily need to be in this PR.

unlikelyzero:
@mabashian this bug is live on devel and this branch: https://github.com/ansible/awx/issues/8076
 Will finish testing by EOD today

dsesami:
@unlikelyzero verify that the Jobs List websocket events work alongside the workflow in two different windows. I'd like to make sure it all lines up.

unlikelyzero:
@mabashian we have a working test for this but need some sort of testability added to the sparkline elements. At the moment, there is no way to determine their state.

mabashian:
recheck

mabashian:
recheck

mabashian:
recheck

unlikelyzero:
tested via #5501

unlikelyzero (approved):

Issue #7498, opened 2020-07-01
the tower_role module should accept ids for related objects


AlanCoding:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Parameters to the `tower_role` module will not resolve integer-like inputs as ids, only as names.

They should.

##### STEPS TO REPRODUCE

See details from

https://groups.google.com/g/awx-project/c/mRuVUCtgMhQ

something like

```yaml
  - name: Give team access to credential
    tower_role:
      credential: 5
      role: use
      team: testteam
```

##### EXPECTED RESULTS

looks up credential with id=5

##### ACTUAL RESULTS

problems finding credential

jmartensson:
It not only tower_role, I also have issues when using tower_credential
```
   - name: Change owner of credential
     tower_credential:
       name: 62
       description: mycredential
       organization: myneworg
       credential_type: Machine
       state: present
```
This ends up with a new credential called '62' instead of making changes to the credential with id '62'

AlanCoding:
that's a separate issue. I can't accept the double-use of a parameter to take `credential: foo` as well as `credential: 42`, but IMO `name: 42` is a bridge too far. That's not even idempotent.

This issue is because a _specific module_ didn't apply the same pattern as other modules. Only `tower_role` needs to be fixed.

To take an id for modification requires changing _all_ modules. If this was implemented, I'd want it implemented as `id: 42`, because that's less confusing. From that, a lot of conditional handling would have to fall out, because the means of lookup would change depending on what parameters the user gave.

AlanCoding:
Also see related issue https://github.com/ansible/awx/issues/5193. This kind of addressability problem is a pain point, and we have some competing proposals.

beeankha:
The current `devel` module_utils favors name over ID when multiple objects are found. While this particular issue was being worked on, it became clear that we actually want to do the _opposite_ and favor the ID over name. This new default behavior would result in a much more precise lookup Collections-wide.

PR #7539, opened 2020-07-02
tower_role.py 204 error field detail 

chris93111:
##### SUMMARY

ansible crash because the api don't return detail field in json

fix error 204 detail > msg return api

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

awx_collection

##### AWX VERSION
13.0.0

##### ADDITIONAL INFORMATION

The full traceback is:
Traceback (most recent call last):
  File "<stdin>", line 113, in <module>
  File "<stdin>", line 105, in _ansiballz_main
  File "<stdin>", line 48, in invoke_module
  File "/tmp/ansible_tower_role_payload_VOI9Tk/__main__.py", line 186, in <module>
  File "/tmp/ansible_tower_role_payload_VOI9Tk/__main__.py", line 173, in main
KeyError: 'detail'

failed: [localhost] (item={u'name': u'DQT_GDE_exploit', u'desc': u'equipe GDE exploit'}) => {
    "changed": false, 
    "item": {
        "desc": "equipe GDE exploit", 
        "name": "DQT_GDE_exploit"
    }, 
    "module_stderr": "Traceback (most recent call last):\n  File \"<stdin>\", line 113, in <module>\n  File \"<stdin>\", line 105, in _ansiballz_main\n  File \"<stdin>\", line 48, in invoke_module\n  File \"/tmp/ansible_tower_role_payload_VOI9Tk/__main__.py\", line 186, in <module>\n  File \"/tmp/ansible_tower_role_payload_VOI9Tk/__main__.py\", line 173, in main\nKeyError: 'detail'\n", 
    "module_stdout": "", 
    "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error", 
    "rc": 1
}

with msg field

failed: [localhost] (item={u'name': u'DQT_GDE_exploit', u'desc': u'equipe GDE exploit'}) => {
    "changed": false, 
    "invocation": {
        "module_args": {
            "credential": null, 
            "inventory": null, 
            "job_template": null, 
            "organization": "CNAF", 
            "project": null, 
            "target_team": null, 
            "team": "DQT_GDE_exploit", 
            "tower_config_file": null, 
            "tower_host": "https://xxxxxxxxxxxxxx", 
            "tower_oauthtoken": null, 
            "tower_password": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER", 
            "tower_username": "admin", 
            "user": null, 
            "validate_certs": false, 
            "workflow": null
        }
[...]

PR #7945, opened 2020-08-19
Get tower_role Module to Accept IDs for Related Objects

beeankha:
##### SUMMARY

Addressing Issue https://github.com/ansible/awx/issues/7498

##### ISSUE TYPE

 - Bugfix Pull Request

### DIFF
--- a/awx_collection/plugins/module_utils/tower_api.py
+++ b/awx_collection/plugins/module_utils/tower_api.py
@@ -105,32 +105,32 @@ class TowerAPIModule(TowerModule):
 
         return response['json']['results'][0]
 
-    def resolve_name_to_id(self, endpoint, name_or_id):
-        # Try to resolve the object by name
+    def get_one_by_name_or_id(self, endpoint, name_or_id):
         name_field = 'name'
-        if endpoint == 'users':
+        if endpoint is 'users':
             name_field = 'username'
 
-        response = self.get_endpoint(endpoint, **{'data': {name_field: name_or_id}})
-        if response['status_code'] == 400:
-            self.fail_json(msg="Unable to try and resolve {0} for {1} : {2}".format(endpoint, name_or_id, response['json']['detail']))
+        query_params = {'or__{0}'.format(name_field): name_or_id}
+        try:
+            query_params['or__id'] = int(name_or_id)
+        except ValueError:
+            # If we got a value error than we didn't have an integer so we can just pass and fall down to the fail
+            pass
 
+        response = self.get_endpoint(endpoint, **{'data': query_params})
         if response['json']['count'] == 1:
-            return response['json']['results'][0]['id']
+            return response['json']['results'][0]
+        elif response['json']['count'] == 2:
+            for tower_object in response['json']['results']:
+                return tower_object
+        # We shouldn't get here because we found 2 objects and ID has to be unique, so one of the objects must have a matching name
         elif response['json']['count'] == 0:
-            try:
-                int(name_or_id)
-                # If we got 0 items by name, maybe they gave us an ID, let's try looking it up by ID
-                response = self.head_endpoint("{0}/{1}".format(endpoint, name_or_id), **{'return_none_on_404': True})
-                if response is not None:
-                    return name_or_id
-            except ValueError:
-                # If we got a value error than we didn't have an integer so we can just pass and fall down to the fail
-                pass
-
             self.fail_json(msg="The {0} {1} was not found on the Tower server".format(endpoint, name_or_id))
         else:
-            self.fail_json(msg="Found too many names {0} at endpoint {1} try using an ID instead of a name".format(name_or_id, endpoint))
+            self.fail_json(msg="Found too many names {0} at endpoint {1}, try using an ID instead of a name".format(name_or_id, endpoint))
+
+    def resolve_name_to_id(self, endpoint, name_or_id):
+        return self.get_one_by_name_or_id(endpoint, name_or_id)['id']
 
     def make_request(self, method, endpoint, *args, **kwargs):
         # In case someone is calling us directly; make sure we were given a method, let's not just assume a GET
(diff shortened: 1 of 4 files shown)


awx_collection/plugins/modules/tower_role.py
```diff
@@ -170,14 +170,14 @@ def main():
                 if response['status_code'] == 204:
                     module.json_output['changed'] = True
                 else:
-                    module.fail_json(msg="Failed to grant role {0}".format(response['json']['detail']))
+                    module.fail_json(msg="Failed to grant role. {0}".format(response['json'].get('detail', response['json'].get('msg', 'unknown'))))
```
AlanCoding, line 172: Link #7539

AlanCoding: This is good, looking for detail and then msg is ideal.

awx_collection/plugins/modules/tower_role.py
```diff
@@ -130,7 +130,7 @@ def main():
 
         resource_name = params.get(param)
         if resource_name:
-            resource = module.get_one(endpoint, **{'data': {name_field: resource_name}})
+            resource = module.get_one_by_name_or_id(module.param_to_endpoint(param), resource_name)
```
AlanCoding, line 132: After you make this change `name_field` is unused, so you should delete its assignment on line 129

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -105,32 +105,36 @@ def get_one(self, endpoint, *args, **kwargs):
...
-                # If we got a value error than we didn't have an integer so we can just pass and fall down to the fail
-                pass
-
             self.fail_json(msg="The {0} {1} was not found on the Tower server".format(endpoint, name_or_id))
         else:
-            self.fail_json(msg="Found too many names {0} at endpoint {1} try using an ID instead of a name".format(name_or_id, endpoint))
+            self.fail_json(msg="Found too many names {0} at endpoint {1}, try using an ID instead of a name".format(name_or_id, endpoint))
+
+    def resolve_name_to_id(self, endpoint, name_or_id):
+        return self.get_one_by_name_or_id(endpoint, name_or_id)['id']
```
AlanCoding, line 140: 👍  good change to the method call pattern here

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -105,32 +105,36 @@ def get_one(self, endpoint, *args, **kwargs):
...
             name_field = 'username'
 
-        response = self.get_endpoint(endpoint, **{'data': {name_field: name_or_id}})
-        if response['status_code'] == 400:
-            self.fail_json(msg="Unable to try and resolve {0} for {1} : {2}".format(endpoint, name_or_id, response['json']['detail']))
+        query_params = {'or__{0}'.format(name_field): name_or_id}
+        try:
+            query_params['or__id'] = int(name_or_id)
+        except ValueError:
+            # If we got a value error than we didn't have an integer so we can just pass and fall down to the fail
```
AlanCoding: this comment should be re-worded or deleted

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -105,32 +105,36 @@ def get_one(self, endpoint, *args, **kwargs):
...
+        if response['status_code'] != 200:
+            self.fail_json(msg="Failed to query endpoint {0} for {1} {2} ({3}), see results".format(endpoint, name_field, name_or_id, response['status_code']), resuls=response)
 
         if response['json']['count'] == 1:
-            return response['json']['results'][0]['id']
+            return response['json']['results'][0]
+        elif response['json']['count'] == 2:
+            for tower_object in response['json']['results']:
+                if tower_object['name'] == name_or_id:
+                    return tower_object
```
AlanCoding, line 133: What scenario is this `elif` trying to address?

If it's addressing the case of >1 object with the same name, then that stands in disagreement to the error case below:

> Found too many names {0} at endpoint

If there are >1 objects with the same name, it should hit that error, and "guessing", and returning the first object here is contradictory.

----

If, on the other hand, this block is designed to fish out an object with "id" equal to `name_or_id` when there are _other_ objects with the integer as a string "name"... then other details are not right. In that scenario, this should not be only for `== 2` objects, you should run this when the count is anything `>1`. You should also only exit when `tower_object['id'] == name_or_id`, as opposed to 'name'.

AlanCoding (changes requested):
My reading is that you are improving on @john-westcott-iv's implementation to doing the same thing in 1 API request which was previously done in 2 requests. This is done using the `or__` query string syntax.

I don't think this is necessary, you could keep the old logic and I think that would be fine, it would still fix the bug. But if you do keep the `or__` stuff there are a few things to patch up.

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -105,32 +105,49 @@ def get_one(self, endpoint, *args, **kwargs):
...
+            object_to_return = None
+            for tower_object in response['json']['results']:
+                # Name takes priority, so we match on name first
+                # If we already have an object to return and it has the right name, we need to fail
+                if object_to_return and object_to_return['name'] == name_or_id:
+                    self.fail_json(msg="The requested name or id was ambiguous and resulted in too many items")
+                # If we found an object ID with this name (which has to be unique) and we don't already have
+                # something to return, select this object for now
+                elif tower_object['id'] == name_or_id and not object_to_return:
+                    object_to_return = tower_object
```
AlanCoding: I still think there's something problematic here. I have drafted a test case where I'm creating a new team, and I give its related Organization as "1". Now there are 2 organizations present in the database:

 - Organization with name "foo" and id of 1
 - Organization with name "1" and id of 2

I give the related organization as `organization: 1`. This integer gets converted to a string by Ansible and comes in here as `name_or_id`. If it hits the org id=1 first, then that gets assigned to `object_to_return`. Thus, it fails on the 2nd item. However, if it hits the org id=2 first, then it does not fail. That's not good.

AlanCoding (approved):
awesome, good work!

Issue #7930, opened 2020-08-18
List jobs related to a certain instance group


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
List jobs related to a certain instance group.

See: https://tower-mockups.testing.ansible.com/patternfly/instance-groups/instance-groups-jobs/

nixocio:
As per feedback from @trahman73, the user should be able to delete jobs in this screen as well. Mock ups will be updated to add this change. 

~I will update this screen also.~ Already done.

dsesami:
bit of feedback here -- there is no way to identify if the timestamp in the list is the start time or the finish time. This could be confusing to users, so it should either be labeled, or both values should be visible in an obvious way. Everything else seems solid.

edit: this will be handled in a future issue.

PR #7951, opened 2020-08-20
Add list of jobs for instance groups

nixocio:
Add list of jobs for instance groups.

See: https://github.com/ansible/awx/issues/7930

image


jakemcdermott (approved):
code  lgtm &  tested it lightly - nice work!

PR #7958, opened 2020-08-21
Fix container groups

shanemcd:
There's a bug in the upstream community.kubernetes plugin. We can open up a
follow-up PR once that has been patched.


shanemcd:
Upstream issue: https://github.com/ansible-collections/community.kubernetes/issues/200

ryanpetrello (approved):

Issue #6113, opened 2020-02-27
Update relevant credential form fields from text area to file picker


jlmitch5:
the File Picker component is slated for the early March release of PF-react.  Once it is released, we should update the text area based fields (private and public SSH key) to this component.

marshmalien:
PatternFly File Upload: https://www.patternfly.org/v4/documentation/react/components/fileupload

mabashian:
Marking this as blocked until we upgrade PF - unless I'm mistaken and we already have access to this.  I'm thinking about bringing us up to the latest PF early next week.

jlmitch5:
I believe this is now merged in PF-react so can be done, though I'm not sure if we want to tackle this after we move to the more generator based cred form creation or not.

mabashian:
This can probably be combined with https://github.com/ansible/awx/issues/6703

dsesami:
@unlikelyzero is this good to close as we tested file upload?

unlikelyzero:
🎉

PR #7960, opened 2020-08-21
Support file upload/drag and drop on credential mutliline fields

mabashian:
##### SUMMARY
link #6113 
link #6703 

Here's upload in action:

upload

And here's the drag and drop:

dragndrop

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/index.jsx
+++ b/awx/ui_next/src/index.jsx
@@ -7,8 +7,8 @@ import { BrandName } from './variables';
 document.title = `Ansible ${BrandName}`;
 
 ReactDOM.render(
-  <React.StrictMode>
-    <App />
-  </React.StrictMode>,
+  // <React.StrictMode>
+  <App />,
+  // </React.StrictMode>,
   document.getElementById('app') || document.createElement('div')
 );
--- a/awx/ui_next/src/index.jsx
+++ b/awx/ui_next/src/index.jsx
@@ -7,8 +7,8 @@ import { BrandName } from './variables';
 document.title = `Ansible ${BrandName}`;
 
 ReactDOM.render(
-  // <React.StrictMode>
-  <App />,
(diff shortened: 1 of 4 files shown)


jakemcdermott (approved):

jlmitch5 (approved):
very nice!

nixocio:
That is neat feature.

unlikelyzero:
@mabashian @jlmitch5 this is our first case of os native uploading in cypress. We expect this to take a full week in terms of turnaround but it will help establish some best practices for os interaction.

dsesami:
Is it possible to create a preventative hard limit on filesize upload? For example, if I paste the wrong file into the RSA private key field, I can successfully freeze up the UI (something like 50MB isn't recoverable and you have to kill the page and come back). It would be nice if it simply shut down the upload if it was too large before the upload itself finished.

unlikelyzero:
@mabashian we're able to use the drag and drop field but cannot use this button upload element. Can you sync with us to go over our options?

mabashian:
@dsesami yes we can implement a max file size.  Here's the library that the PF component is using under the hood: https://react-dropzone.js.org/#!/Dropzone

unlikelyzero:
@mabashian Filed as https://github.com/ansible/awx/issues/8082 . As soon as we sort out the file upload problems with our test framework, this can be merged

mabashian:
Decided to file an enhancement for the file upload size limit: https://github.com/ansible/awx/issues/8083

unlikelyzero:
@mabashian can we get IDs added to the Save and Cancel buttons?
Screen Shot 2020-09-04 at 10 22 16 AM

unlikelyzero:
verified via 5484

Issue #7353, opened 2020-06-16
Enforce feature-completeness of the AWX collection


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
The AWX collection has a [module generator](https://github.com/ansible/awx/blob/devel/awx_collection/tools/generate.yml) that will use the API OPTIONS to produce a module with all of the same fields that the API has.

However, that has only be employed for adding new modules. We also need to devise a way to keep the collection up-to-date with API changes through some form of automation.

For a first pass, we would like to enforce that:

 - all top-level endpoints from the API (or other metric) have a module corresponding to them
 - every module has all of the fields that the API has

We can still keep exclusion lists and other exceptions, as necessary.

If we create this enforcement mechanism, we could run it in the Zuul CI so that we know that it's being kept up with every merge.

beeankha:
PR https://github.com/ansible/awx/pull/7963 has been merged, closing this issue!

PR #7963, opened 2020-08-21
Collections Completeness Tests

john-westcott-iv:
##### SUMMARY
Adds a test to test_collection which validates for feature completness.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - Collection

##### AWX VERSION

```
awx: 14.0.0
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx_collection/test/awx/test_completeness.py
+++ b/awx_collection/test/awx/test_completeness.py
@@ -4,6 +4,7 @@ __metaclass__ = type
 import pytest
 
 from awx.main.tests.functional.conftest import _request
+from ansible.module_utils.six import PY2, string_types
 import yaml
 import os
 import re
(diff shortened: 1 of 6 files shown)


AlanCoding:
This looks really awesome.

Getting into the technical nitty gritty, I'm interested if we could "map" `tower_workflow_launch` to `workflow_jobs`, because the field comparisons should still be meaningful there. The editable / read-only distinction won't be accurate, but I objectively we would still like to look for field parity there.

john-westcott-iv:
Adding reference to #7353

awx_collection/test/awx/test_completeness.py
```diff
@@ -0,0 +1,251 @@
...
+import re
+
+# Analysis variables
+# -----------------------------------------------------------------------------------------------------------
+
+# Read-only endpoints are dynamically created by an options page with no POST section.
+# Noramlly a read-only endpoint should not have a module (i.e. /api/v2/me) but sometimes we reuse a name
+# For example, we have a tower_role module but /api/v2/roles is a read only endpoint.
+# Rhis list indicates which read-only endpoints have associated modules with them.
+read_only_endpoints_with_modules = ['tower_settings', 'tower_role']
```
AlanCoding, line 18: But both of these wind up being exceptions anyway don't they?

awx_collection/test/awx/test_completeness.py
```diff
@@ -0,0 +1,251 @@
...
+]
+
+# Some modules take additional parameters that do not appear in the API
+# Add the module name as the key with the value being the list of params to ignore
+no_api_parameter_ok = {
+    # The credential module used to take a lot of parameters which are now combined into "inputs"
+    'tower_credential': [
+        'authorize', 'authorize_password', 'become_method', 'become_password', 'become_username', 'client',
+        'domain', 'host', 'kind', 'password', 'project', 'secret', 'security_token', 'ssh_key_data',
+        'ssh_key_unlock', 'subscription', 'tenant', 'username', 'vault_id', 'vault_password',
```
AlanCoding: instead of enumerating these, could you instead search for "deprecated" in the help text and skip in those cases instead?

awx_collection/test/awx/test_completeness.py
```diff
@@ -0,0 +1,251 @@
...
+    # Always_nodes, failure_nodes, success_nodes, credentials is how we handle associations
+    # Organization is how we looking job templates
+    'tower_workflow_job_template_node': ['always_nodes', 'failure_nodes', 'success_nodes', 'credentials', 'organization'],
+    # Survey is how we handle associations
+    'tower_workflow_job_template': ['survey'],
+    # These fields get wrapped into notification_configuration
+    'tower_notification_template': [
+        'account_sid', 'account_token', 'channels', 'client_name', 'color', 'from_number', 'headers', 'host',
+        'message_from', 'nickname', 'notify', 'password', 'port', 'recipients', 'sender', 'server',
+        'service_key', 'subdomain', 'targets', 'to_numbers', 'token', 'url', 'use_ssl', 'use_tls', 'username',
```
AlanCoding: Same here, I would prefer some more general solution for deprecated params

awx_collection/test/awx/test_completeness.py
```diff
@@ -0,0 +1,251 @@
...
+    'tower_token': ['existing_token', 'existing_token_id'],
+    # Children and hosts are how tower_group deals with associations
+    'tower_group': ['children', 'hosts'],
+    # Credential and Vault Credential are a legacy fields for old Towers
+    # Credentials/labels/survey spec is now how we handle associations
+    # We take an organization here to help with the lookups only
+    'tower_job_template': ['credential', 'vault_credential', 'credentials', 'labels', 'survey_spec', 'organization'],
+    # Always_nodes, failure_nodes, success_nodes, credentials is how we handle associations
+    # Organization is how we looking job templates
+    'tower_workflow_job_template_node': ['always_nodes', 'failure_nodes', 'success_nodes', 'credentials', 'organization'],
```
AlanCoding: Could we also do some general solution (via parameter help text) for related fields?

People are general averse to hard lists of fields, and in cases like these, we can get real value added for the user by putting this signals in a standard form in the help text in DOCUMENTATION.

awx_collection/test/awx/test_completeness.py
```diff
@@ -0,0 +1,251 @@
...
+        'message_from', 'nickname', 'notify', 'password', 'port', 'recipients', 'sender', 'server',
+        'service_key', 'subdomain', 'targets', 'to_numbers', 'token', 'url', 'use_ssl', 'use_tls', 'username',
+    ]
+}
+
+# When this tool was created we were not feature complete. Adding something in here indicates a module
+# that needs to be developed. If the module is found on the file system it will auto-detect that the
+# work is being done and will bypass this check. At some point this module should be removed from this list.
+needs_development = [
+    'tower_ad_hoc_command', 'tower_application', 'tower_instance_group', 'tower_inventory_script',
```
AlanCoding, line 56: `tower_inventory_script` is for a deprecated endpoint, so you could move this over to the ignore list instead.

awx_collection/test/awx/test_completeness.py
john-westcott-iv: Deprecated might catch some of them but not sure I want to rely specifically on that.

AlanCoding (approved):
Thanks for implementing the changes from our discussion, the exceptions list seems very reasonable to me here. I'm good to merge this as something that closes #7353

beeankha (approved):

PR #7965, opened 2020-08-21
adding in inventory to unified job query for analytics

rebeccahhh:
##### SUMMARY

adds inventory name and ID to collector

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 14.0.0
```

### DIFF
--- a/awx/main/analytics/collectors.py
+++ b/awx/main/analytics/collectors.py
@@ -228,7 +228,7 @@ def query_info(since, collection_type):
 
 # Copies Job Events from db to a .csv to be shipped
 @table_version('events_table.csv', '1.1')
-@table_version('unified_jobs_table.csv', '1.0')
+@table_version('unified_jobs_table.csv', '1.1')
 @table_version('unified_job_template_table.csv', '1.0')
 @table_version('workflow_job_node_table.csv', '1.0')
 @table_version('workflow_job_template_node_table.csv', '1.0')
@@ -272,6 +272,8 @@ def copy_tables(since, full_path, subset=None):
                                  django_content_type.model,
                                  main_unifiedjob.organization_id,
                                  main_organization.name as organization_name,
+                                 main_job.inventory_id, 
+                                 main_inventory.name, 
                                  main_unifiedjob.created,  
                                  main_unifiedjob.name,  
                                  main_unifiedjob.unified_job_template_id, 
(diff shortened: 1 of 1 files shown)


wenottingham:
cc @Ladas - this will need processor support once it lands

rooftopcellist:
@rebeccahhh it looks like this test might need to be updated based on the test failure - https://github.com/ansible/awx/blob/b69b53f52774784cf88816af507c65b96c79c6bd/awx/main/tests/functional/analytics/test_collectors.py#L70

Ladas:
it will need to join to `main_inventory` table most likely

rebeccahhh:
@rooftopcellist actually @Ladas is right, we're going to have to do a table update. I opened this PR quite prematurely :sweat_smile: 
Working on that now.

rebeccahhh:
Alright updated, should be passing that test now @rooftopcellist

awx/main/analytics/collectors.py
```diff
@@ -262,6 +264,7 @@ def _copy_table(table, query, path):
                               main_jobevent.event_data::json->'res'->'warnings' AS warnings,
                               main_jobevent.event_data::json->'res'->'deprecations' AS deprecations
                               FROM main_jobevent 
+                              LEFT JOIN main_inventory ON main_inventory.id = main_job.inventory_id
```
jbradberry: This seems to be a typo, it looks like we're dealing with `main_jobevent` in this query

awx/main/analytics/collectors.py
```diff
@@ -287,8 +292,9 @@ def _copy_table(table, query, path):
                                  main_unifiedjob.elapsed, 
                                  main_unifiedjob.job_explanation, 
                                  main_unifiedjob.instance_group_id
-                                 FROM main_unifiedjob
+                                 FROM main_unifiedjob, main_job
```
jbradberry: I think this isn't the type of join we want, this is a type of implicit inner join called an equi join.

rebeccahhh: wouldn't that just require changing the left join below, not this line?

awx/main/analytics/collectors.py
rebeccahhh: ope, that's what I get for typing too fast, :+1:

awx/main/analytics/collectors.py
rebeccahhh: :facepalm: nvm

jbradberry (approved):
This looks ok to me now

awx/main/analytics/collectors.py
```diff
@@ -254,6 +254,8 @@ def _copy_table(table, query, path):
                               main_jobevent.task,
                               main_jobevent.role, 
                               main_jobevent.job_id, 
+                              main_jobevent.inventory_id,
```
rooftopcellist: I think you need `main_inventory.id,` here instead.

rebeccahhh: I actually need main_job.inventory_id so that I can match it to the inventory table, should be fixed

rebeccahhh:
@Ladas @notting merging this today is likely

beeankha (approved):

wenottingham (approved):

kdelee:
@rebeccahhh in the future you can run YOLO with TESTEXPR=analytics

need to update a test because of the version change for the unified jobs csv

rebeccahhh:
@kdelee good to know, I was on PTO when this merged, I'll work with @rooftopcellist to fix this. :smile_cat: 
update--should be taken care of

Issue #7848, opened 2020-08-06
Dispatcher unable to handle absent nodes


john-westcott-iv:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

To start, we are backing up a cluster and restoring to servers in a different data center which may have different names. In the database the main_instance table has entries with the same UUIDs but different hostnames. When this happens the dispatcher for running jobs becomes unreachable and you can no longer cancel jobs. 

PR https://github.com/ansible/awx/pull/7823 fixes a 500 error from happening within the UI when trying to cancel a job running on a non-reachable dispatcher. While it works for a job template, workflow job templates appear to "resurrect" themselves in the UI. https://drive.google.com/file/d/1KBdTjOeY1Jag1eISlKbR9WJhiwwv9x_k/view?usp=sharing This appears to be happening because on a workflow only the cancel_flag is set too True and then the dispatcher picks up this flag and actually performs the job cancel. So the UI makes it looked canceled but then (after the status is reloaded) the UI puts it back into its running state.

Should the reaper logic should be updated to handle dispatcher processes which have been lost and will not return?

##### ENVIRONMENT
* AWX version: 13.0.0 (Tower 3.7.1)
* AWX install method: Tower Cluster
* Ansible version:  X.Y.Z
* Operating System: RHEL
* Web Browser: Any

##### STEPS TO REPRODUCE
Build a 3 node cluster with external database.
Start a workflow running on the cluster.
Backup the cluster.

Restore the backup into a new 3 node cluster.
The workflow that was running in the old cluster during the backup will not be cancelable. 
NOTE: there is a PR to the backup code #841 which will auto-cancel running jobs during a restore. If the above is tested with that PR this will not be reproducible.

##### EXPECTED RESULTS
Canceling a job in the UI will result in the job being canceled.

##### ACTUAL RESULTS
Jobs are stuck running in queue.

wenottingham:
`with the same UUIDs but different hostnames`

How???

ryanpetrello:
To reproduce something like this in a fairly contrived way:

1.  Launch a job and let it finish.
2.  `Job.objects.filter(pk=N).update(status='running', execution_node='some-nonsense.example.org')`
3. Even if you restart services, this just stays in `running` forever.

kdelee:
Tried this out and worked like charm! Thanks

```
[...]

PR #7966, opened 2020-08-21
improve job reaping for jobs that were started on a missing Instance

ryanpetrello:
see: https://github.com/ansible/awx/issues/7848

### DIFF
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -14,8 +14,10 @@ from django.utils.translation import ugettext_lazy as _, gettext_noop
 from django.utils.timezone import now as tz_now
 
 # AWX
+from awx.main.dispatch.reaper import reap_job
 from awx.main.models import (
     AdHocCommand,
+    Instance,
     InstanceGroup,
     InventorySource,
     InventoryUpdate,
@@ -515,6 +517,20 @@ class TaskManager():
                 task.job_explanation = timeout_message
                 task.save(update_fields=['status', 'job_explanation', 'timed_out'])
 
+    def reap_jobs_from_orphaned_instances(self):
+        # discover jobs that are in running state but aren't on an execution node
+        # that we know about; this is a fairly rare event, but it can occur if you,
+        # for example, SQL backup an awx install with running jobs and restore it
+        # elsewhere
+        for j in UnifiedJob.objects.filter(
+            status__in=['pending', 'waiting', 'running'],
+        ).exclude(
+            execution_node__in=Instance.objects.values_list('hostname', flat=True)
+        ):
+            if j.execution_node and not j.is_containerized:
+                logger.error(f'{j.execution_node} is not a registered instance; reaping {j.log_format}')
+                reap_job(j, 'failed')
+
     def calculate_capacity_consumed(self, tasks):
         self.graph = InstanceGroup.objects.capacity_values(tasks=tasks, graph=self.graph)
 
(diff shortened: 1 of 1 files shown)


awx/main/scheduler/task_manager.py
```diff
@@ -567,6 +583,7 @@ def _schedule(self):
             self.spawn_workflow_graph_jobs(running_workflow_tasks)
 
             self.timeout_approval_node()
+            self.reap_jobs_from_orphaned_instances()
```
ryanpetrello, line 586: I put this in the task manager's lock because I don't want multiple jobs in the cluster to all run this same code at once (it only needs to happen on *one* node).

awx/main/scheduler/task_manager.py
```diff
@@ -515,6 +517,20 @@ def timeout_approval_node(self):
...
 
+    def reap_jobs_from_orphaned_instances(self):
+        # discover jobs that are in running state but aren't on an execution node
+        # that we know about; this is a fairly rare event, but it can occur if you,
+        # for example, SQL backup an awx install with running jobs and restore it
+        # elsewhere
+        for j in UnifiedJob.objects.filter(
+            status__in=['pending', 'waiting', 'running'],
+        ).exclude(
+            execution_node__in=Instance.objects.values_list('hostname', flat=True)
```
ryanpetrello, line 528: `execution_node` is *not* indexed, but `status` is, and there *should* be relatively few jobs in this status at a given time.

awx/main/scheduler/task_manager.py
```diff
@@ -515,6 +517,20 @@ def timeout_approval_node(self):
...
+        ):
+            if j.execution_node and not j.is_containerized:
```
ryanpetrello, line 530: re: `is_containerized`, I think w/ container groups we set this execution_node to be the hostname of the pod or something, but I can't remember

cc @shanemcd

matburt (approved):

awx/main/scheduler/task_manager.py
chrismeyersfsu: Back of the envelope math:
Assume 10,000 jobs. How big is a job? https://dba.stackexchange.com/questions/23879/measure-the-size-of-a-postgresql-table-row Average job size + averaged unifiedjob size in an AWX DB came out to `3kb` for me. `10,000 * 3kb = 30MB` Let say a modest Postgres has 16GB and 3% is allocated to `work_mem` = `480MB`. Plenty of working memory to do the operation in question.

chrismeyersfsu (approved):

chrismeyersfsu:
Thought. Don't we have other reaper code that performs queries similar to this? I'd rather reuse that gathered query and do the reaping there.

ryanpetrello:
> Don't we have other reaper code that performs queries similar to this? I'd rather reuse that gathered query and do the reaping there.

Peers in a cluster will discover nodes that are *down* and reap _their_ jobs.  But in this circumstance, the `execution_node` doesn't match any known `Instance` records.

chrismeyersfsu:
True, the other reaper code would only select jobs that are in status `waiting, running` while this reaper includes `pending` also. So not a great unified query. Alright, this works for me.

awx/main/scheduler/task_manager.py
shanemcd: Yeah here's the codepath for that: https://github.com/ansible/awx/blob/feb9c5700cd700c6e8d972854e454fbd2871d053/awx/main/scheduler/task_manager.py#L237-L261

Issue #7876, opened 2020-08-11
Notification Templates add/edit form


keithjgrant:
Follow-up on #7867

##### ISSUE TYPE
 - Feature Idea

tiagodread:
Test and stubs are written and the PR was merged

Issue #7773, opened 2020-07-30
Evaluate disable form submit buttons per PF guidance


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
PF doc: https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/forms#enabling-and-disabling-buttons

> "Form submit buttons should be disabled until a user starts filling out the form or edits it in any way."

jlmitch5:
We should disable the save button until a user starts filling out the form or edits it in any way, or if the form is invalid with the client-side validators return false, such as when they do when you have not entered a value for a required field.

we should pass a new prop to `FormActionGroup` so that we can set this information on the isDisabled prop of the save button.  It should be something like:

```
saveIsDisabled={!(formik.dirty || formik.isValid)}
```

nixocio:
As per @trahman73 feedback.

>we need to address this in wizards as well..... we currently allow you to go to the next step even if you haven't completed any required fields in the current step

AlexSCorey:
removing myself to reflect accurate priorities

Issue #7515, opened 2020-07-01
Refactor FormSubmitError to use fully unit tested util


keithjgrant:
##### ISSUE TYPE
 - Improvement

##### SUMMARY
The FormSubmitError component does some complicated parsing of API error responses, which it then passes to Formik and/or displays as a general error to the user. This logic should be moved into a util function with a full array of unit tests to account for all the various structures that the API might return.

keithjgrant:
@mabashian — I'd like to pick your brain a bit on some of the changes you made here to make sure I understand all the permutations of these error messages before proceeding.

AlexSCorey:
We have a util, in there also `getErrorMessage` used by `<ErrorDetail>` which could be use to generate the error messages for `<FormSubmitError>`

PR #7967, opened 2020-08-21
Notifications form

keithjgrant:
##### SUMMARY
Adds notification template form and add/edit pages

addresses #7876 

Note: this probably needs a touch more polish but I wanted to get some eyes on it while I'm out on PTO next week

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/screens/NotificationTemplate/shared/CustomMessagesSubForm.jsx
+++ b/awx/ui_next/src/screens/NotificationTemplate/shared/CustomMessagesSubForm.jsx
@@ -1,0 +1,213 @@
+import 'styled-components/macro';
+import React from 'react';
+import { withI18n } from '@lingui/react';
+import { t } from '@lingui/macro';
+import { useField } from 'formik';
+import {
+  FormGroup,
+  Title,
+  Switch,
+  Text,
+  TextVariants,
+} from '@patternfly/react-core';
+import {
+  FormColumnLayout,
+  FormFullWidthLayout,
+  SubFormLayout,
+} from '../../../components/FormLayout';
+import FormField, {
+  PasswordField,
+  CheckboxField,
+  FieldTooltip,
+} from '../../../components/FormField';
+import AnsibleSelect from '../../../components/AnsibleSelect';
+import { CodeMirrorField } from '../../../components/CodeMirrorInput';
+import {
+  combine,
+  required,
+  requiredEmail,
+  url,
+} from '../../../util/validators';
+import { NotificationType } from '../../../types';
+
+function CustomMessagesSubForm({ defaultMessages, type, i18n }) {
+  const [useCustomField, , useCustomHelpers] = useField('useCustomMessages');
+  const showMessages = type !== 'webhook';
+  const showBodies = ['email', 'pagerduty', 'webhook'].includes(type);
+
+  return (
+    <>
+      <Switch
+        id="toggle-custom-messages"
+        label={i18n._(t`Customize messages…`)}
+        isChecked={!!useCustomField.value}
+        onChange={() => useCustomHelpers.setValue(!useCustomField.value)}
+      />
+      {useCustomField.value && (
+        <SubFormLayout>
+          <Text
+            className="pf-c-content"
+            css="margin-bottom: var(--pf-c-content--MarginBottom)"
+          >
+            <small>
+              Use custom messages to change the content of notifications sent
+              when a job starts, succeeds, or fails. Use curly braces to access
+              information about the job:{' '}
+              <code>
+                {'{{'} job_friendly_name {'}}'}
+              </code>
+              ,{' '}
+              <code>
+                {'{{'} url {'}}'}
+              </code>
+              , or attributes of the job such as{' '}
+              <code>
+                {'{{'} job.status {'}}'}
+              </code>
+              . You may apply a number of possible variables in the message.
+              Refer to the{' '}
+              <a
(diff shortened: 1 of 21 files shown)


AlexSCorey:
@keithjgrant 
Just poking at this a bit:
1) I was able to create an email notification, but then I edited it, and removed the recipients list (which says it's required) and I was able to still save.  However, it did not save the removed, recipients list
2) On the list, it seems like in most places where we have an inline label we are not using a colon to separate the label from the value.  We do use a colon in Survey List, but that's the only place that I found it in a brief search.  @trahman73 can you comment on whether to use the colon?
3) Disable the save button.  This is a new pattern that we need the implement here and in all other forms. (https://github.com/ansible/awx/issues/7773)

trahman73:
@AlexSCorey That's correct, we do not use a colon to separate a label from it's value.

awx/ui_next/src/screens/NotificationTemplate/shared/CustomMessagesSubForm.jsx
```diff
@@ -0,0 +1,232 @@
...
+        onChange={() => useCustomHelpers.setValue(!useCustomField.value)}
+      />
+      {useCustomField.value && (
+        <SubFormLayout>
+          <Text
+            className="pf-c-content"
+            css="margin-bottom: var(--pf-c-content--MarginBottom)"
+          >
+            <small>
+              Use custom messages to change the content of notifications sent
```
mabashian: Hrmmm we'll have to figure out how to translate all of this

mabashian:
@keithjgrant I went ahead and rebased this to make it a bit easier to review.  I noticed that there were some failing tests in the cred form test suite so I pushed a fix for that.  It looks like the only thing left that's failing in the awx-ui-next is a lint failure:

```
2020-08-26 14:15:03.399065 | static | /awx_devel/awx/ui_next/src/components/FormField/ArrayTextField.jsx
2020-08-26 14:15:03.399143 | static |   43:19  error  'value' is already declared in the upper scope  no-shadow
```

mabashian:
This may be unrelated to your PR but after I save a change to a notification and get redirected to the details view, I don't see that change immediately reflected.  I have to refresh the page.  Perhaps there's some kind of race going on?

notif_details

keithjgrant:
> This may be unrelated to your PR but after I save a change to a notification and get redirected to the details view, I don't see that change immediately reflected. I have to refresh the page. Perhaps there's some kind of race going on?

I remember this being an issue with other details pages a while back. I'll track down what we did as a fix then update this one to the same behavior

keithjgrant:
recheck

tiagodread:
- Trying to create  a notification without type organization and notification type (only name) The UI doesn’t show any error message

- Trying to create a notification without organization the message "This field is required." is showing twice:

> 1. Fill name
> 2. Choose notification type `Grafana`
> 3. Fill Grafana URL and Grafana API Key
> 4. Click on Save button

image

- After saving a notification on the details tab the metadata `Destination Channels` shows all destinations without any space and this happens on all the places where you can specify multiple channels or emails to receive a notification:

image

- Trying to create a notification I have to check and uncheck the `Disable SSL verification` if i want to save the form without checking this option (this also happens in all forms that use this select field):

notfcation_0

tiagodread:
- Trying to create a notification type mattermost the field `Icon URL`must is required even the label showing not required(this also happens for rocketchat):

notfcation_1

- I don’t know why, but when you click on any label on `pagerduty` form 2x the page crashes totally and this just happens on this form (I verified all of them):

notfcation_2

Console:

`react-dom.development.js:13413 Uncaught Error: Objects are not valid as a React child (found: object with keys {_activeMessages, _catalogs, t, select, plural, selectOrdinal, _dev, _activeLanguageData, _language, _locales, _missing}). If you meant to render a collection of children, use an array instead.`

- Trying to create a notification type twilio the fields `Source phone number` and `Destination SMS number(s)` are accepting letters, is this correct ?

image

- Trying to create a email notification type, the fields `Host, Recipient list, Port and Time out` doesn’t have validation, so I can save the form like this:

image

tiagodread (changes requested):
There’s some issues that we have to fix first before merging this PR

keithjgrant:
@tiagodread I've addressed all the issues listed above, except the following items:

> Trying to create a notification without organization the message "This field is required." is showing twice

This is a known bug I plan to address in https://github.com/ansible/awx/issues/7515
I did tweak the layout a little so the first error shows in red as it is meant to and the second error sits beneath the last form field.

> After saving a notification on the details tab the metadata `Destination Channels` shows all destinations without any space and this happens on all the places where you can specify multiple channels or emails to receive a notification

I've fixed this bug in a separate forthcoming PR

> Trying to create a notification type twilio the fields Source phone number and Destination SMS number(s) are accepting letters, is this correct ?

This matches behavior of the old UI, though it certainly could be improved in the future (maybe file a low priority issue for follow-up? phone number validation is notoriously difficult to do well, made even more so by the multi-input field here)

> Trying to create a email notification type, the fields Host, Recipient list, Port and Time out doesn’t have validation

I fixed `port` and `timeout` validation, but let the other fields un-validated to match behavior of the old UI.

mabashian (approved):

tiagodread:
@keithjgrant is it possible to add unique ids for these custom message inputs elements?

tiagodread (approved):
Retested, tests and stubs are written!
👍🏽

PR #7969, opened 2020-08-23
Add label to identify isolated instance group JobDetail screen

nixocio:
Add label to identify isolated instance group JobDetail screen as per
mock up.

See: https://tower-mockups.testing.ansible.com/patternfly/jobs/jobs-detail/


jlmitch5 (approved):
looks good

PR #7974, opened 2020-08-24
Adds fix to load template details 

### LINKED ISSUE #6561: WFJT Details Won't Load for users that do not have Add/Edit Permissions
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
If a `template.summary_fields.user_capabilities.edit: false` or  `template.summary_fields.user_capabilities.delete: false`
the WFJT details page won't load.

##### ENVIRONMENT
* AWX version: 2.9
* AWX install method:  docker for mac
* Ansible version:  9.3
* Operating System: Mojave
* Web Browser: Chrome

##### STEPS TO REPRODUCE

Create a wfjt with webhooks enabled and a webhook key.  Allow another use to have execute permissions on the wfjt.  Sign in as that other user and click on the wfjt so that it should navigate you to the details page.
##### EXPECTED RESULTS

The details page loads
##### ACTUAL RESULTS

The details page fails to load because the user does not have the proper permissions to hit the api to get the webhook_key.  

##### ADDITIONAL INFORMATION

We need to set up some sort of conditional that if the user does not have proper permissions to get the webhook_key then that api call should not be made. Instead we should probably show `ENCRYPTED` in the field where the webhook_key would be.

@trahman73 can you comment on this?

AlexSCorey:
##### SUMMARY
this resolves #6561.

It adds an options request to determine if the user has to ability to read the webhook key and if they do, it will fetch the key, otherwise it won't.

##### ISSUE TYPE
 - Docs Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/screens/Template/Template.test.jsx
+++ b/awx/ui_next/src/screens/Template/Template.test.jsx
@@ -18,11 +18,16 @@ const mockMe = {
   is_system_auditor: false,
 };
 describe('<Template />', () => {
+  let wrapper;
   beforeEach(() => {
     JobTemplatesAPI.readDetail.mockResolvedValue({
       data: mockJobTemplateData,
     });
-
+    JobTemplatesAPI.readTemplateOptions.mockResolvedValue({
+      data: {
+        actions: { PUT: true },
+      },
+    });
     OrganizationsAPI.read.mockResolvedValue({
       data: {
         count: 1,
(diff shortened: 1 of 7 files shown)


awx/ui_next/src/screens/Template/Template.test.jsx
```diff
@@ -117,4 +133,62 @@ describe('<Template />', () => {
...
+                  path: '/templates/job_template/1/foobar',
+                },
+              },
+            },
+          },
+        }
+      );
+    });
+    expect(JobTemplatesAPI.readWebhookKey).not.toHaveBeenCalled();
+  });
```
jlmitch5, line 193: nice tests

jlmitch5 (approved):
very nice

unlikelyzero (approved):

unlikelyzero:
Auditor is able to view WFJT Details view for a WFJT with and without Webhooks enabled

Issue #7977, opened 2020-08-24
Inventory Host Facts display strange loading behavior


jakemcdermott:
**expected behavior**
The details should appear loading until the data request resolves.

**actual behavior**
The fact component displays momentarily, disappears, then shows again when the request resolves.

**additional information**
![host_facts_load1](https://user-images.githubusercontent.com/9753817/91073160-a0cae600-e608-11ea-8936-a4d7d110413f.gif)

dsesami:
This appears to simply no longer flicker. :100:

PR #7978, opened 2020-08-24
Don't display facts until facts are loaded

jakemcdermott:
##### SUMMARY
for #7977 

**before:**
host_facts_load1

**after:**
host_facts_load2

### DIFF
--- a/awx/ui_next/src/screens/Inventory/InventoryHostFacts/InventoryHostFacts.jsx
+++ b/awx/ui_next/src/screens/Inventory/InventoryHostFacts/InventoryHostFacts.jsx
@@ -11,32 +11,31 @@ import useRequest from '../../../util/useRequest';
 import { HostsAPI } from '../../../api';
 
 function InventoryHostFacts({ i18n, host }) {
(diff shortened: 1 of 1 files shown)


awx/ui_next/src/screens/Inventory/InventoryHostFacts/InventoryHostFacts.jsx
```diff
@@ -11,32 +11,31 @@ import useRequest from '../../../util/useRequest';
...
-    return <ContentLoading />;
-  }
+    request();
+  }, [request]);
 
   if (error) {
     return <ContentError error={error} />;
   }
 
+  if (isLoading || result === null) {
```
jakemcdermott, line 31: In my opinion, there isn't a single default initial value for `isLoading` that's appropriate for _all_ use cases of `useRequest`.

Eventually, I'd like to look into updating `useRequest` to allow setting initial values other than `result` but that's a much bigger change. In the meantime, this works.

mabashian (approved):

jakemcdermott:
regate

PR #7983, opened 2020-08-24
Fix pylint Errors in Collections Modules

beeankha:
##### SUMMARY

Fixing miscellaneous `pylint` failures in Collections modules.

##### ISSUE TYPE

 - Collections

### DIFF
--- a/awx_collection/test/awx/conftest.py
+++ b/awx_collection/test/awx/conftest.py
@@ -187,7 +187,7 @@ def run_module(request, collection_import):
         try:
             result = json.loads(module_stdout)
         except Exception as e:
-            raise Exception('Module did not write valid JSON, error: {0}, stdout:\n{1}'.format(str(e), module_stdout))
+            module.fail_json(msg='Module did not write valid JSON, error: {0}, stdout:\n{1}'.format(str(e), module_stdout))
         # A module exception should never be a test expectation
         if 'exception' in result:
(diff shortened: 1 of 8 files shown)


beeankha:
`PEP 3135 New Super` information: https://www.python.org/dev/peps/pep-3135/

AlanCoding (approved):
If test results look good, then merge away!

awx_collection/test/awx/conftest.py
```diff
@@ -187,7 +187,7 @@ def mock_load_params(self):
         try:
             result = json.loads(module_stdout)
         except Exception as e:
-            raise Exception('Module did not write valid JSON, error: {0}, stdout:\n{1}'.format(str(e), module_stdout))
+            module.fail_json(msg='Module did not write valid JSON, error: {0}, stdout:\n{1}'.format(str(e), module_stdout))
```
AlanCoding: We never initialized something like the `module` variable here, if you did, I think it would look like `module = resource_class()`, but I'm not a fan of that pattern for this context

AlanCoding (changes requested):
sorry I was going to fast for myself, I think we need a different kind of solution for the conftest.py file here for the "raise from" issue

AlanCoding (approved):

kdelee (approved):
:1st_place_medal:

PR #7987, opened 2020-08-25
Add advanced search keys for InstanceGroup and CredentialType Lists

nixocio:
Add advanced search keys for `InstanceGroup` and `CredentialType` Lists.

See: https://github.com/ansible/awx/pull/7895/files


AlexSCorey (approved):
This looks good to me.

PR #7991, opened 2020-08-25
hashivault_kv auth_path moved from metadata to inputs

bbayszczak:
##### SUMMARY
The auth_path is used with the approle auth method

As specified here: https://github.com/ansible/awx-custom-credential-plugin-example/blob/master/awx_custom_credential_plugin_example/__init__.py
`inputs.fields represents fields the user will specify *when they create* a credential of this type`
`inputs.metadata represents values the user will specify *every time* they link two credentials together*`

It's not linked to the secret we are reading but to the auth method so this parameter has to be moved to inputs.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - credential_plugins/hashivault

##### AWX VERSION
awx: 14.0.0

### DIFF
--- a/awx/main/credential_plugins/hashivault.py
+++ b/awx/main/credential_plugins/hashivault.py
@@ -41,11 +41,11 @@ base_inputs = {
         'secret': True,
         'help_text': _('The Secret ID for AppRole Authentication')
     }, {
-        'id': 'auth_path',
-        'label': _('Path to Auth'),
+        'id': 'approle_auth_path',
+        'label': _('Path to Approle Auth'),
         'type': 'string',
         'multiline': False,
-        'help_text': _('The path where the Authentication method is mounted e.g, approle')
+        'help_text': _('The path where the AppRole Authentication method is mounted e.g, approle')
     }
     ],
     'metadata': [{
@@ -53,6 +53,12 @@ base_inputs = {
         'label': _('Path to Secret'),
         'type': 'string',
         'help_text': _('The path to the secret stored in the secret backend e.g, /some/secret/')
+    }, {
+        'id': 'auth_path',
+        'label': _('Path to Auth'),
+        'type': 'string',
+        'multiline': False,
+        'help_text': _('The path where the Authentication method is mounted e.g, approle')
     }],
     'required': ['url', 'secret_path'],
 }
@@ -119,7 +125,11 @@ def handle_auth(**kwargs):
 def approle_auth(**kwargs):
     role_id = kwargs['role_id']
     secret_id = kwargs['secret_id']
-    auth_path = kwargs.get('auth_path') or 'approle'
+    # we first try to use the 'auth_path' from the metadata
+    # if not found we try to fetch the 'approle_auth_path' from inputs
+    # if not found we use the default value 'approle'
+    auth_path = \
+        kwargs.get('auth_path', kwargs.get('approle_auth_path', "approle"))
 
     url = urljoin(kwargs['url'], 'v1')
     cacert = kwargs.get('cacert', None)
--- a/awx/main/credential_plugins/hashivault.py
+++ b/awx/main/credential_plugins/hashivault.py
@@ -40,6 +40,12 @@ base_inputs = {
         'multiline': False,
         'secret': True,
         'help_text': _('The Secret ID for AppRole Authentication')
+    }, {
+        'id': 'auth_path',
+        'label': _('Path to Auth'),
+        'type': 'string',
+        'multiline': False,
+        'help_text': _('The path where the Authentication method is mounted e.g, approle')
     }
     ],
     'metadata': [{
@@ -47,11 +53,6 @@ base_inputs = {
         'label': _('Path to Secret'),
         'type': 'string',
         'help_text': _('The path to the secret stored in the secret backend e.g, /some/secret/')
-    },{
-        'id': 'auth_path',
-        'label': _('Path to Auth'),
-        'type': 'string',
-        'help_text': _('The path where the Authentication method is mounted e.g, approle')
     }],
     'required': ['url', 'secret_path'],
 }
--- a/awx/main/credential_plugins/hashivault.py
+++ b/awx/main/credential_plugins/hashivault.py
@@ -41,11 +41,12 @@ base_inputs = {
         'secret': True,
         'help_text': _('The Secret ID for AppRole Authentication')
     }, {
-        'id': 'approle_auth_path',
+        'id': 'default_auth_path',
         'label': _('Path to Approle Auth'),
         'type': 'string',
         'multiline': False,
-        'help_text': _('The path where the AppRole Authentication method is mounted e.g, approle')
+        'default': 'approle',
+        'help_text': _('The AppRole Authentication path to use if one isn\'t provided in the metadata when linking to an input field. Defaults to \'approle\'')
     }
     ],
     'metadata': [{
@@ -126,10 +127,8 @@ def approle_auth(**kwargs):
     role_id = kwargs['role_id']
     secret_id = kwargs['secret_id']
     # we first try to use the 'auth_path' from the metadata
-    # if not found we try to fetch the 'approle_auth_path' from inputs
-    # if not found we use the default value 'approle'
-    auth_path = \
-        kwargs.get('auth_path', kwargs.get('approle_auth_path', "approle"))
+    # if not found we try to fetch the 'default_auth_path' from inputs
+    auth_path = kwargs.get('auth_path') or kwargs['default_auth_path']
 
     url = urljoin(kwargs['url'], 'v1')
     cacert = kwargs.get('cacert', None)
--- a/awx_collection/test/awx/test_credential_input_source.py
+++ b/awx_collection/test/awx/test_credential_input_source.py
@@ -122,7 +122,8 @@ def source_cred_hashi_secret(organization):
             "url": "https://secret.hash.example.com",
             "token": "myApiKey",
             "role_id": "role",
-            "secret_id": "secret"
+            "secret_id": "secret",
+            "approle_auth_path": "path-to-approle"
         }
     )
 
@@ -157,6 +158,8 @@ def test_hashi_secret_credential_source(run_module, admin_user, organization, so
     assert cis.metadata['secret_backend'] == "backend"
     assert cis.metadata['secret_key'] == "a_key"
     assert cis.source_credential.name == source_cred_hashi_secret.name
+    assert cis.source_credential.approle_auth_path == \
+           source_cred_hashi_secret.inputs["approle_auth_path"]
     assert cis.target_credential.name == tgt_cred.name
     assert cis.input_field_name == 'password'
     assert result['id'] == cis.pk
--- a/awx_collection/test/awx/test_credential_input_source.py
+++ b/awx_collection/test/awx/test_credential_input_source.py
@@ -123,7 +123,7 @@ def source_cred_hashi_secret(organization):
             "token": "myApiKey",
             "role_id": "role",
             "secret_id": "secret",
-            "approle_auth_path": "path-to-approle"
+            "default_auth_path": "path-to-approle"
         }
     )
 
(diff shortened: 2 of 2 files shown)


awx/main/credential_plugins/hashivault.py
```diff
@@ -40,18 +40,19 @@
         'multiline': False,
         'secret': True,
         'help_text': _('The Secret ID for AppRole Authentication')
+    }, {
+        'id': 'auth_path',
+        'label': _('Path to Auth'),
+        'type': 'string',
+        'multiline': False,
+        'help_text': _('The path where the Authentication method is mounted e.g, approle')
```
jakemcdermott: I _think_ more than one auth path can be available to a given AppRole role_id and AppRole secret_id, which _may_ be the reason why the auth path was originally set as a metadata variable instead of an input.

However, I do see the rationale for grouping the auth path with the inputs instead because it is probably _more_ convenient to just make a separate hashi credential for each auth_login path than it would be to provide the same login paths every time you link the hashi credential to an input field.

Based on my understanding of https://learn.hashicorp.com/tutorials/vault/approle#policy-requirements, I think this pr is probably the way to go. An app may have multiple approle auth paths it can use, but it doesn't _seem_ like you'd typically use more than one at a time for most situations.

That said, I may be overlooking a use case that depends on the current configuration. 

cc @kawsark  @ryanpetrello Thoughts?

ryanpetrello: Also worth mentioning that changing this is likely going to be a breaking change and potentially require a messy data migration for people who've already set these up.

kawsark: Agreed with Ryan. Also often larger customers have 100s of AppRole Auth methods setup (typically one per application) and it might be cumbersome to create a new hashi credential for each auth_login path. Usually the auth login path can be determined by the CI/CD Pipeline based on variables.

jakemcdermott: > Also often larger customers have 100s of AppRole Auth methods setup (typically one per application) and it might be cumbersome to create a new hashi credential for each auth_login path

This is great info, thank you @kawsark. This is a compelling reason to leave the current variable as-is. 

> Usually the auth login path can be determined by the CI/CD Pipeline based on variables.

I think this implies linking input fields to external credentials dynamically in a ci job run? (cool!)

Do you feel that there is also a (separate) use case for keeping an auth path scoped to the whole hashi credential? Perhaps we can accommodate both use cases with a _default_ approle auth path that lives in the inputs. The default will be overridden by any metadata auth paths that are provided. 

Pros:
- additive, so non-breaking & no messy data migrations
- leaves currently addressed use cases intact
- solves this problem:
  > It's not linked to the secret we are reading [#7991](https://github.com/ansible/awx/pull/7991#issue-473306135)

Cons:
- Yet another field, complexity, etc.

cc @ryanpetrello @wenottingham

bbayszczak: Thanks for your reviews

The way approle works is:
  - you create an `approle` auth method at a specific path in Vault. Using this path you will be able to interact with this approle 

```
# `zuul` is the path at wchich the approle auth method will be created 
$> vault auth enable -path=zuul approle
Success! Enabled approle auth method at: zuul/
```

  - for each approle auth method, you can create several roles, each role have specific rights/associated policies
```
# here we create a role called `repo1` with the policy `default` associated
$> vault write auth/zuul/role/repo1 token_policies=default
Success! Data written to: auth/zuul/role/repo1
```

  - for each role you have one `role_id` and you can create several `secret_id`
```
# we read the role_id of this role
$> vault read auth/zuul/role/repo1/role-id
Key        Value
---        -----
role_id    620272b1-00db-b06d-8c25-7b41c6d77da7

# we generate a secret_id for this role
$> vault write -f auth/zuul/role/repo1/secret-id
Key                   Value
---                   -----
secret_id             650c5fbb-8f77-e556-17bf-d37f3001a5f4
secret_id_accessor    75aa1796-c08a-d5bf-824a-4a78d887de0e
```
  - you use the previously generated `role_id` and `secret_id` to login using the path of the approle auth method
The tuple `role_id` and `secret_id` depend on only one approle (you don't need to specify the role name during login as a `role_id` match only one role)
```
$> vault write auth/zuul/login role_id=620272b1-00db-b06d-8c25-7b41c6d77da7 secret_id=650c5fbb-8f77-e556-17bf-d37f3001a5f4
Key                     Value
---                     -----
token                   s.hTMy2GLh9WUyTnoZLZxwhrqe
token_accessor          pA0pC3o3Qurs8atlH8PMGl8w
token_duration          768h
token_renewable         true
token_policies          ["default"]
identity_policies       []
policies                ["default"]
token_meta_role_name    repo1
```

As you can see a tuple `role_id` and `secret_id` match only one approle auth method, only one path

I don't think it's possible to use an approle to auth against several auth methods as the role_id is generated by Vault for each role, that's why I made this change

@jakemcdermott 
> An app may have multiple approle auth paths it can use

That's right but the app will need a `role_id` and a `secret_id` for each approle auth path (and also for different roles at the same approle auth path)

@ryanpetrello 
> Also worth mentioning that changing this is likely going to be a breaking change and potentially require a messy data migration for people who've already set these up.

That's completely right but having the approle auth path elsewhere than in the credential is only data duplication

@kawsark 
I can't see the use case you're talking about as, from my point of view, there is already different credentials for each auth_login path

To add more context, we are using Vault since several years now and we're using several approles/roles
But we started the AWX setup recently so maybe I'm missing something on a pure AWX point of view ? (on the way credentials are stored/used ?)

jakemcdermott: This is a great conversation and there is some excellent information here. Thanks.

**problem: 1**: 
> 100s of AppRole Auth methods setup (typically one per application) and it might be cumbersome to create a new hashi credential for each auth_login path. https://github.com/ansible/awx/pull/7991#discussion_r477383273

**problem: 2**: 
> from my point of view, there is already different credentials for each auth_login path https://github.com/ansible/awx/pull/7991#discussion_r477437118

**problem 3**: 
> Also worth mentioning that changing this is likely going to be a breaking change and potentially require a messy data migration for people who've already set these up. https://github.com/ansible/awx/pull/7991#discussion_r477350244


_Adding_ a default auth_path to the input fields and overriding it w/ the metadata path if both are provided will address all of these problems and concerns. We can explain the override behavior with appropriate variable naming and informative help text on the fields. Thoughts? Other ideas?

bbayszczak: > _Adding_ a default auth_path to the input fields and overriding it w/ the metadata path if both are provided will address all of these problems and concerns.

This looks like the right way to do it 👍 

I'll work on it then submit a new commit for review

Thanks

jakemcdermott: @bbayszczak

Sounds good - thank you.

jakemcdermott (changes requested):
(See discussion and next steps in the review comments: https://github.com/ansible/awx/pull/7991/files#r478259375)

Current plan is to add a default auth path to the input fields. I'll take another look when that is ready, thank you!

awx/main/credential_plugins/hashivault.py
```diff
@@ -40,17 +40,24 @@
         'multiline': False,
         'secret': True,
         'help_text': _('The Secret ID for AppRole Authentication')
+    }, {
+        'id': 'approle_auth_path',
```
jakemcdermott:
```suggestion
        'id': 'default_auth_path',
```

awx/main/credential_plugins/hashivault.py
```diff
@@ -40,17 +40,24 @@
...
+        'label': _('Path to Approle Auth'),
+        'type': 'string',
+        'multiline': False,
+        'help_text': _('The path where the AppRole Authentication method is mounted e.g, approle')
```
jakemcdermott:
```suggestion
        'help_text': _('The AppRole Authentication path to use if one isn\'t provided in the metadata when linking to an input field. Defaults to \'approle\'.)
```

awx/main/credential_plugins/hashivault.py
```diff
@@ -40,17 +40,24 @@
         'multiline': False,
         'secret': True,
         'help_text': _('The Secret ID for AppRole Authentication')
+    }, {
+        'id': 'approle_auth_path',
+        'label': _('Path to Approle Auth'),
+        'type': 'string',
+        'multiline': False,
```
jakemcdermott, line 47:
```suggestion
        'multiline': False,
        'default': 'approle',
```

awx/main/credential_plugins/hashivault.py
```diff
@@ -118,7 +125,11 @@ def handle_auth(**kwargs):
 def approle_auth(**kwargs):
     role_id = kwargs['role_id']
     secret_id = kwargs['secret_id']
-    auth_path = kwargs.get('auth_path') or 'approle'
+    # we first try to use the 'auth_path' from the metadata
+    # if not found we try to fetch the 'approle_auth_path' from inputs
+    # if not found we use the default value 'approle'
+    auth_path = \
+        kwargs.get('auth_path', kwargs.get('approle_auth_path', "approle"))
```
jakemcdermott:
```suggestion
        kwargs.get('auth_path') or kwargs['default_auth_path']
```

Note: the `or` is just to keep this implementation robust and behaving correctly in case the value of `auth_path` ever gets saved to `''` (empty string).

Screenshot from 2020-08-31 14-48-53

awx_collection/test/awx/test_credential_input_source.py
```diff
@@ -122,7 +122,8 @@ def source_cred_hashi_secret(organization):
             "url": "https://secret.hash.example.com",
             "token": "myApiKey",
             "role_id": "role",
-            "secret_id": "secret"
+            "secret_id": "secret",
+            "approle_auth_path": "path-to-approle"
```
jakemcdermott:
```suggestion
            "default_auth_path": "path-to-approle"
```

awx_collection/test/awx/test_credential_input_source.py
```diff
@@ -157,6 +158,8 @@ def test_hashi_secret_credential_source(run_module, admin_user, organization, so
     assert cis.metadata['secret_backend'] == "backend"
     assert cis.metadata['secret_key'] == "a_key"
     assert cis.source_credential.name == source_cred_hashi_secret.name
+    assert cis.source_credential.approle_auth_path == \
+           source_cred_hashi_secret.inputs["approle_auth_path"]
```
jakemcdermott:
```suggestion
           source_cred_hashi_secret.inputs["default_auth_path"]
```

jakemcdermott (changes requested):
This is looking good - I have a few suggested changes for naming and configuration (see review comments). Once those are in I'll pull this pr down and test it out. Thanks!

awx/main/credential_plugins/hashivault.py
```diff
@@ -118,7 +125,11 @@ def handle_auth(**kwargs):
 def approle_auth(**kwargs):
     role_id = kwargs['role_id']
     secret_id = kwargs['secret_id']
-    auth_path = kwargs.get('auth_path') or 'approle'
+    # we first try to use the 'auth_path' from the metadata
+    # if not found we try to fetch the 'approle_auth_path' from inputs
+    # if not found we use the default value 'approle'
```
jakemcdermott:
```suggestion
```

jakemcdermott: This comment can be removed if we configure a `default` for the field: https://github.com/ansible/awx/pull/7991/files#r480304473)

awx/main/credential_plugins/hashivault.py
bbayszczak: Right, I did not think about the empty string use case, thanks

bbayszczak:
recheck

bbayszczak:
recheck

jakemcdermott (approved):
Code lgtm. I debugged and tested this to verify that pre-existing hashi creds with and without metadata approle paths will construct the same auth url before and after applying the change.

jakemcdermott (changes requested):
@bbayszczak The changes LGTM. 

~Please rebase this PR branch with the latest changes from devel. Once that is done we can merge this.~ Thank you!

Edit: Nevermind the rebase - this can be merged as-is. :+1:

jakemcdermott (approved):

Issue #7364, opened 2020-06-17
Add ability to cancel jobs from jobs list


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
We need to provide a way for users to cancel a job from the jobs list.  Currently only relaunch and delete are possible:

<img width="1420" alt="Screen Shot 2020-06-17 at 5 34 14 AM" src="https://user-images.githubusercontent.com/9889020/84881684-36238a00-b05c-11ea-9b7d-6fe40f9e57b2.png">

There are two options:

1) We could add a cancel icon to each row (like the old app)
2) We could add another list level action button next to `Delete` and allow users to select multiple rows before clicking `Cancel`.  This would behave like batch delete where we still make individual requests for each job.

@wenottingham @trahman73 any preference on one or the other?

trahman73:
@mabashian I'm fine with either option, but after our discussion it sounds like option 1 would be easier to implement.

wenottingham:
I think we're going to get asked for the "batch" variant, so if it's doable, that would be nice.

tiagodread:
Test and stubs in place PR already merged

PR #7993, opened 2020-08-25
Adds cancel button to jobs list

mabashian:
##### SUMMARY
link #7364 

Screen Shot 2020-08-25 at 1 54 32 PM

canceljob

This feature adds a cancel button next to the delete button on the jobs list.  This cancel button behaves like the delete button in that a user can select multiple jobs to cancel at one time.  Some general rules surrounding the button:

1) Button should be disabled until a user selects at least one job
2) Button should be disabled if user selects a job that they don't have the ability to cancel (summary_fields.user_capabilities.start is false)
3) API requests should not be made for jobs that are in a finished state.  A user can still select these rows but confirming a cancel will not actually send the request(s).  I confirmed this behavior with Taufique and Bill.

I also wanted to use the `copy` related endpoint provided on all of the job objects rather than having a big `switch` statement covering all of the different job types.  Since we didn't have anything that would support taking in an arbitrary endpoint I created `Related.js` in our models directory.  This model has generic methods covering GET/POST/DELETE/PUT/PATCH.  It doesn't inherit from the base model so I needed to pull the axios http instantiation out into something that could be consumed by more than one model.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI


awx/ui_next/src/components/JobList/JobListCancelButton.jsx
```diff
@@ -0,0 +1,141 @@
...
+
+  const renderTooltip = () => {
+    const jobsUnableToCancel = jobsToCancel
+      .filter(cannotCancel)
+      .map(job => job.name)
+      .join(', ');
+    if (jobsToCancel.some(cannotCancel)) {
+      return (
+        <div>
+          {i18n.plural({
```
mabashian: i18n.plural is no good -> see #8002 

Use low level i18n._ instead

jakemcdermott (changes requested):
Looking good & works well. We dove into job-switching problem offline. Ultimately @mabashian and I decided on adding a util to select the job model given a `job` object as a way to avoid having to add a switch everywhere a job api is used.

jakemcdermott (approved):

unlikelyzero:
Tiago will be working on this one today

tiagodread:
@mabashian when I tried to cancel a running job using the Cancel button inside kebab (when the advanced search is enabled) nothing happens (the same occurred for delete button)

cancel_button

tiagodread:
recheck

tiagodread (approved):
Tests and stubs in place 👍🏽

tiagodread:
regate

PR #8003, opened 2020-08-26
add certificates to the list of CAs trusted (in case anchors is mounted)

velzend:
##### SUMMARY
The local docker example also includes volume mount for `/etc/pki/ca-trust/source/anchors`. 
Certificates are not automatically added to the trusted list of CAs without running `update-ca-trust`.

Run this command later via `docker exec`.

Another alternative is to change the container mount point to the compiled bundle, located on the OS and in the container at: `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem`.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API
 - Installer

##### AWX VERSION

```
awx: 14.1.0
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/installer/roles/image_build/templates/launch_awx_task.sh.j2
+++ b/installer/roles/image_build/templates/launch_awx_task.sh.j2
@@ -5,6 +5,8 @@ if [ `id -u` -ge 500 ]; then
     rm /tmp/passwd
 fi
 
+update-ca-trust
+
 source /etc/tower/conf.d/environment.sh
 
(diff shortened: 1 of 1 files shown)


ryanpetrello:
@velzend sorry for the delay on reviewing/merging this PR - we've since replaced the local Docker installer in `devel` (https://groups.google.com/g/awx-project/c/47MjWSUQaOc/m/bCjSDn0eBQAJ https://github.com/ansible/awx/tree/devel/tools/docker-compose) and thus the next major release of AWX, so I suspect this PR is obsolete now.

Issue #8006, opened 2020-08-26
using set_stats with large amounts of data slows down UI and API endpoints


ryanpetrello:
1.  Create an inventory w/ 25+ (local connection) hosts.
2.  Run this playbook w/ a 100MB artifact file for `set_stats`:

```yaml
---
- hosts: all
  connection: local
  tasks:
  - set_stats:
      data:
        random_data: "{{ lookup('file', 'random_data.txt') }}"
    delegate_to: localhost
```

```shell
bash-4.4$ base64 /dev/urandom | head -c 100000000 > random_data.txt
bash-4.4$ du -h random_data.txt
96M	random_data.txt
```

3.  Go to `/api/v2/inventories/N/hosts/`
4.  Become very, very sad:

<img width="372" alt="image" src="https://user-images.githubusercontent.com/214912/91335777-e15c6800-e79e-11ea-8d52-f29c2efdbcf4.png">

one-t:
Tested using a slightly modified example playbook on a local k8s install.

```
---
- hosts: all
  connection: local
  tasks:
  - shell: 'base64 /dev/urandom | head -c 100000000 > random_data.txt'
    run_once: true
  - set_stats:
      data:
        random_data: "{{ lookup('file', 'random_data.txt') }}"
    delegate_to: localhost
```
This is now working as expected, response times were around 0.05 seconds, which is normal for this installation.

PR #8007, opened 2020-08-26
defer loading Job.artifacts and .extra_vars on host views to improve performance

ryanpetrello:
see: https://github.com/ansible/awx/issues/8006

this data can be *really* large, and we don't actually need it for the summary fields on this API endpoint


ryanpetrello:
Before:
image

After:
image

jbradberry (approved):

AlanCoding (approved):
I wish that we had a better solution with values_list or something

matburt (approved):

PR #8022, opened 2020-08-27
Add wait function and update collection modules

sean-m-sullivan:
##### SUMMARY
Adding a wait function to be used by multiple modules when waiting on an endpoint to update. 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
Collection

##### AWX VERSION
```
awx: 14.0.0
```

##### ADDITIONAL INFORMATION
As discussed in PR #7873 , adding a wait function for use in modules.


awx_collection/plugins/modules/tower_job_wait.py
```diff
@@ -93,20 +93,6 @@
...
-
-
-def check_job(module, job_url):
-    response = module.get_endpoint(job_url)
-    if response['status_code'] != 200:
-        module.fail_json(msg="Unable to read job from Tower {0}: {1}".format(response['status_code'], module.extract_errors_from_response(response)))
-
-    # Since we were successful, extract the fields we want to return
-    for k in ('id', 'status', 'elapsed', 'started', 'finished'):
-        module.json_output[k] = response['json'].get(k)
```
john-westcott-iv, line 106: For tower_job_wait this was a legacy module which we attempted to keep backwards compatibility. In most of the new returns we try to not return specific pieces of data but this module does to achieve the backwards compatibly. Without this logic being moved between your result=module.wait_on_url and the modue.exit_json the unit tests will fail.

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -588,3 +589,26 @@ def is_job_done(self, job_status):
             return False
         else:
             return True
+
+    def wait_on_url(self, object_name=None, object_type=None, url=None, timeout=None, interval=None):
```
john-westcott-iv: The modules have defaults for their wait parameters but here wait is set to None. if someone used this function and forgot to set wait to something you will likely get exceptions in the code. I'd rather see these wait have a default like 30 to prevent that from ever happening.

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -588,3 +589,26 @@ def is_job_done(self, job_status):
...
+        # Grab our start time to compare against for the timeout
+        start = time.time()
+        result = self.get_endpoint(url)
+        while not result['json']['finished']:
```
john-westcott-iv, line 597: Since we are making this generic, do we want to test to see if 'json' and 'json'['finished'] actually exist in the result? Imagine if someone gave you a URL that did not confirm to the standard one we have been checking so finished is not in there. If that were the case they would get a stack trace in their ansible run instead of a "pretty" error message like "Finished was not returned in this request".

awx_collection/plugins/module_utils/tower_api.py
john-westcott-iv: Also, consider if you want object_name, object_type and URL to be defaulted to None. If object_name and type are None that will make the fail message on line 600 to say something like 'Monitoring of None "None" aborted due to timeout' which is not super helpful. If URL is none you are going to get an exception when you try to call self.get_endpoint(url). I'd rather see these not have a default of None and be non-optional parameters.

john-westcott-iv (changes requested):

awx_collection/plugins/modules/tower_job_wait.py
sean-m-sullivan: I've updated this, and ran the tests to make sure it passed the assertions, Also added a comment to note this is code for legacy compatibility so it is not removed in the future.

awx_collection/plugins/module_utils/tower_api.py
sean-m-sullivan: Updated to be non default parameters, also saw how the error is clearer, set default times for timeout and interval.

awx_collection/plugins/module_utils/tower_api.py
sean-m-sullivan: All the endpoints I checked had finished because they are all job based, but I added in a check that will fail with a pretty message if this is not the case.

awx_collection/plugins/modules/tower_job_wait.py
sean-m-sullivan: Actually having a second look at this, after it failed in the python integration test, I took a look at the output, 
```
{
    "wait_results": {
        "changed": false,
        "failed": false,
        "status": "successful",
    }
}
```
Was to generic, so I took that for loop and made it default across the board. 
```
{
    "wait_results": {
        "changed": false,
        "elapsed": 0.0,
        "failed": true,
        "finished": "2020-08-28T13:21:52.001950Z",
        "id": 1248,
        "msg": "The job \"1248\" failed",
        "started": null,
        "status": "canceled",
    }
}
```

beeankha:
recheck

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -588,3 +589,34 @@ def is_job_done(self, job_status):
...
+
+            # Put the process to sleep for our interval
+            time.sleep(interval)
+
+            result = self.get_endpoint(url)
+            self.json_output['status'] = result['json']['status']
+
+        # If the job has failed, we want to raise a task failure for that so we get a non-zero response.
+        if result['json']['failed']:
+            self.json_output['msg'] = 'The {0} "{1}" failed'.format(object_type, object_name)
```
beeankha: The one `awx-ansible-module` CI test that is erroring out I _think_ directly comes back to this part of the code.  

The error message is: `AssertionError: assert {'msg': 'The job "1" failed'} != {'msg': 'Job with id 1 failed'}`, which makes me think that we either need to change the expected output of the `test_job_wait_failed ` unit test in `/awx_devel/awx_collection/test/awx/test_job.py`, or the format of the error message in this part of the module utils.

AlanCoding:
@sean-m-sullivan we merged a few test fixes, if you could rebase this I'll be happy to kick things off again.

sean-m-sullivan:
@AlanCoding Done, I think its kicking off on its own.

awx_collection/plugins/module_utils/tower_api.py
sean-m-sullivan: I could change it to the old format, or update the Test. I like the idea of the new format, and either change is easy, I think its down to, what do we prefer? Keeping with legacy format, the new one, or something else?

sean-m-sullivan: @beeankha updated to pass tests and look pretty for everything but tower_job_wait.

sean-m-sullivan: Updated to work for legacy and have nice output for other.

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -588,3 +589,42 @@ def is_job_done(self, job_status):
...
+    def wait_on_url(self, url, object_name, object_type, timeout=30, interval=10):
+        # Grab our start time to compare against for the timeout
+        start = time.time()
+        result = self.get_endpoint(url)
+        while not result['json']['finished']:
+            # If we are past our time out fail with a message
+            if timeout and timeout < time.time() - start:
+                # Account for Legacy messages
+                if object_type is 'legacy_job_wait':
+                    self.json_output['msg'] = 'Monitoring of Job - {1} aborted due to timeout'.format(object_name)
```
AlanCoding: This format statement uses `{1}` but it looks like it should use the 0 index.

sean-m-sullivan: Updated to 0

AlanCoding (approved):

beeankha (approved):

john-westcott-iv (approved):

beeankha:
regate

PR #8029, opened 2020-08-27
Run project updates from a copy of the playbook vendoring directory

AlanCoding:
##### SUMMARY
This is for consistency and compliance with how `ansible-runner` works.

Previously, project updates did have a private data directory, but they ran the playbook inside of the `awx/playbooks` directory.

This makes the change that we _copy_ the playbook and other associated Ansible content, and then run with the _working directory_ inside of the temporary folder.

The idea here is that this will be easier to grok and manage as containerized job runs & the receptor mesh is implemented. We ship private data dirs, not some crazy list of files elsewhere on the computer.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
14.1.0
```

##### ADDITIONAL INFORMATION
This has already been merged into the `shanemcd:execution-environments` branch. But I want it here too.

### DIFF
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -2169,7 +2169,7 @@ class RunProjectUpdate(BaseTask):
         self._write_extra_vars_file(private_data_dir, extra_vars)
 
     def build_cwd(self, project_update, private_data_dir):
-        return self.get_path_to('..', 'playbooks')
+        return os.path.join(private_data_dir, 'project')
 
     def build_playbook_path_relative_to_cwd(self, project_update, private_data_dir):
         return os.path.join('project_update.yml')
@@ -2309,6 +2309,12 @@ class RunProjectUpdate(BaseTask):
             logger.warning('{0} unexpectedly existed before update'.format(stage_path))
             shutil.rmtree(stage_path)
         os.makedirs(stage_path)  # presence of empty cache indicates lack of roles or collections
+
+        # the project update playbook is not in a git repo, but uses a vendoring directory
+        # to be consistent with the ansible-runner model,
+        # that is moved into the runner projecct folder here
+        awx_playbooks = self.get_path_to('..', 'playbooks')
+        copy_tree(awx_playbooks, os.path.join(private_data_dir, 'project'))
 
     @staticmethod
     def clear_project_cache(cache_dir, keep_value):


matburt (approved):

ghjm:
Makes sense to me.  Should we worry at all about the time and space requirements of making a second copy?  How big is the biggest project someone might have?

ghjm (approved):

AlanCoding:
> How big is the biggest project someone might have?

No, that's not what this is. The `awx_playbooks` path points to the folder https://github.com/ansible/awx/tree/devel/awx/playbooks. It's not a git repo, it's just a folder with playbooks and plugins in it, existing inside of the AWX install directory. That's a static size (discounting code changes we, ourselves, make). Including subdirectories, it comes to 12 files.

The project update needs `project_update.yml` to run. This copies that playbook to the _private data dir_ for each _project update_ (because all job types are ran via `ansible-runner`). Then the project update will run, and make changes to the specific project folder like `/var/lib/awx/projects/_4_foo`.

We use a lock so that only 1 update is using (reading & writing) the `_4_foo` folder at the same time. Before a job runs, we will copy contents from that folder into the _job private data dir_. This is a different folder from the project update's private data dir. This copy has serious concerns about time/space, but nothing about that is changing here.

While we have a lock for multiple local updates of the same project, multiple updates of _different_ projects can happen simultaneously, and they all read `project_update.yml`, the benefit gained is that we will not volume-mount the same place for multiple concurrent jobs.

Philosophically, it's nicer organizationally that we don't stuff outside of a unified job's private data dir. Removing inventory scripts helped with that, and adding this copy leaves things more consistent so we can better refactor task logic as needed.

AlanCoding:
d'oh, I should remove this too:

https://github.com/ansible/awx/blob/45f7e4a66306ffe097d0e293119dd3bc54cc9a15/awx/main/tasks.py#L1980-L1982

The whole point is that we're cutting down on special access of this type.

ghjm:
Ah ok, you're right, I misunderstood what this is entirely.

AlanCoding:
Oh boy, now that I say that about the project root being unnecessarily whitelisted, I am finding all kinds of old stuff that shouldn't actually be there in `tasks.py`. There's good agreement and confidence about this, so I think I will punt that cleanup to a secondary PR.

AlanCoding:
regate

PR #8030, opened 2020-08-27
Execution Environments

shanemcd:
This is a long-running feature branch where Execution Environments is being implemented. A proper upgrade plan for the community will need to be in place before this can be merged and included in a release of AWX.

## TODOs
- [x] Documentation
- [x] EE Module in AWX collection
- [x] Integrate Receptor


shanemcd:
recheck

jakemcdermott:
recheck

jakemcdermott:
regate

jakemcdermott:
recheck

shanemcd:
@AlanCoding looks like you may have broken the build with that last commit. Can you take a look? Maybe we should go back to only doing PRs to this branch.

shanemcd:
recheck

shanemcd:
recheck

shanemcd:
recheck

awx/api/metadata.py
```diff
@@ -209,7 +209,7 @@ def determine_actions(self, request, view):
                     continue
 
                 if field == "pod_spec_override":
-                    meta['default'] = PodManager().pod_definition
+                    meta['default'] = AWXReceptorJob().pod_definition
```
AlanCoding, line 212: It would be nice if this pod definition was not defined in 2 places to be DRY.

awx/main/access.py
```diff
@@ -1308,6 +1308,53 @@ def can_unattach(self, obj, sub_obj, relationship, *args, **kwargs):
...
+        return ExecutionEnvironment.objects.filter(
+            Q(organization__in=Organization.accessible_pk_qs(self.user, 'execution_environment_admin_role')) |
+            Q(organization__isnull=True)
+        ).distinct()
+
+    @check_superuser
+    def can_add(self, data):
+        if not data:  # So the browseable API will work
+            return Organization.accessible_objects(self.user, 'execution_environment_admin_role').exists()
+        return self.check_related('organization', Organization, data)
```
AlanCoding: We should test that if the organization is null, a non-superuser can not add the execution environment. It might be, I just wanted to make the obvious note. I don't think `check_related` does this, that's what the `mandatory` kwarg was for.

awx/main/access.py
```diff
@@ -1308,6 +1308,53 @@ def can_unattach(self, obj, sub_obj, relationship, *args, **kwargs):
...
+
+    @check_superuser
+    def can_change(self, obj, data):
+        if obj and obj.organization_id is None:
+            raise PermissionDenied
+        if self.user not in obj.organization.execution_environment_admin_role:
+            raise PermissionDenied
+        org_pk = get_pk_from_dict(data, 'organization')
+        if obj and obj.organization_id != org_pk:
```
AlanCoding: covering these kinds of cases was why we have the `check_related` method to begin with

awx/main/access.py
```diff
@@ -1308,6 +1308,53 @@ def can_unattach(self, obj, sub_obj, relationship, *args, **kwargs):
...
+        if not data:  # So the browseable API will work
+            return Organization.accessible_objects(self.user, 'execution_environment_admin_role').exists()
+        return self.check_related('organization', Organization, data)
+
+    @check_superuser
+    def can_change(self, obj, data):
+        if obj and obj.organization_id is None:
+            raise PermissionDenied
+        if self.user not in obj.organization.execution_environment_admin_role:
+            raise PermissionDenied
```
AlanCoding, line 1347: Other methods just return `True` in these cases. I'm not against changing them all to raising instead, but it's inconsistency for no reason.

awx/main/migrations/0125_more_ee_modeling_changes.py
```diff
@@ -0,0 +1,46 @@
+# Generated by Django 2.2.16 on 2020-11-19 16:20
```
AlanCoding, line 1: are we going to combine these migrations into 1 file before merge?

AlanCoding (approved):

unlikelyzero:
@nixocio based on some test PR feedback, would it be possible to replace the data-cy attribute with a data-ouia attribute?

shanemcd:
recheck

Issue #6367, opened 2020-03-20
Support all unified job template actions (update, approve, etc.)


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Right now with the AWX collection, you can:

 - launch jobs
 - launch workflows

It should also be possible to take the corresponding action for all the other types:

 - update inventory sources
 - update projects
 - approve approval templates
 - launch system job templates (module does not exist)

Prior attempted implementation was https://github.com/ansible/awx/pull/5856 for inventory sources.

john-westcott-iv:
We already have some pattern established for this with job templates. We have tower_job_template to manage a JT and then tower_job_[launch|wait|cancel]. If we need to maintain that pattern for backwards compatibility should that just be our pattern?

So for projects it would be something like: tower_project (to manage a project) and then tower_project_update_[launch|wait]. 

I like the ideas of separating the "CRUD" from "the actions". It feels like that would better mimic how the API works (i.e. a job template is a different thing that a job) and I think having more modules might be clearer to users rather than having only a couple with a ton of different states. (like: 'updated', 'launched', 'canceled', 'present', 'absent', etc). From an auto-generation perspective having a CRUD and "action" templates may also be easier than more complex modules with multiple states. 

If we truly do want to go for less modules (and are ok with breaking the existing pattern) I think I would still prefer tower_job_template and tower_job and let _job_template take the standard absent, present and let job take states: launched, canceled, waited (not sure I like 'waited' but stated are past test).

AlanCoding:
There are a _lot_ of "action" endpoints, although it depends on how you count association.

CRUD modules work with the language of Ansible, and I don't see that for "action" modules, which sound like an anti-pattern. Ideally, if something is truly an action, then it shouldn't be supported in the collection. Association is an action, but can be put into the language of desired state by the client.

I want to ask these questions, debate proposed patterns, and hopefully get it right. If we deprecate something, then fine, we should.

> I think I would still prefer tower_job_template and tower_job and let _job_template take the standard absent, present and let job take states: launched, canceled, waited 

I would like to know more about how this would look. I don't quite get it.

AlanCoding:
Link https://github.com/ansible/awx/pull/7873

AlanCoding:
out of date after feature completeness testing, and the sub-issues have new issues opened for them

AlanCoding:
Link https://github.com/ansible/awx/issues/8009

PR #7616, opened 2020-07-13
Create tower_inventory_refresh.py awx_collection

chris93111:
##### SUMMARY
This module update all sources in inventory

-->

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
awx_collection

##### AWX VERSION
```
13.0.0
```

Issue #5193, opened 2019-10-31
In tower_inventory_source, take organization parameter for inventory org


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
If you create an inventory source like this:

```yaml
  tower_inventory_source:
    name: source-test-inventory
    inventory: openstack-test-inventory
    source: ec2
```

It will fail if there are 2 versions of "openstack-test-inventory" in different organizations.

This module should have the organization parameter added, and filter the inventory according to that.

Basically this is to track cherry-picking https://github.com/ansible/ansible/pull/63100

AlanCoding:
regressed

beeankha:
https://github.com/ansible/awx/pull/8033 is now merged, closing this issue!

PR #8033, opened 2020-08-27
Add New tower_inventory_source_update Module

beeankha:
##### SUMMARY

Addressing Issue https://github.com/ansible/awx/issues/6367

Adding a new `tower_inventory_source_update` module, which will enable Collections users to update their inventory sources via playbook.  Also adding corresponding integration test and updating `awx_collection/test/awx/test_completeness.py` to reflect changes.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API
- Collections

### DIFF
--- a/awx_collection/plugins/modules/tower_inventory_source.py
+++ b/awx_collection/plugins/modules/tower_inventory_source.py
@@ -131,7 +131,6 @@ options:
     organization:
       description:
         - Name of the inventory source's inventory's organization.
-      required: True
       type: str
 extends_documentation_fragment: awx.awx.auth
 '''
@@ -174,7 +173,7 @@ def main():
         enabled_value=dict(),
         host_filter=dict(),
         credential=dict(),
-        organization=dict(required=True),
+        organization=dict(),
         overwrite=dict(type='bool'),
         overwrite_vars=dict(type='bool'),
         custom_virtualenv=dict(),
@@ -203,14 +202,10 @@ def main():
     source_project = module.params.get('source_project')
     state = module.params.get('state')
 
-    # Attempt to look up inventory source based on the provided name and inventory ID
-    org_id = module.resolve_name_to_id('organizations', organization)
-    inventory_object = module.get_one('inventories', **{
-        'data': {
-            'name': inventory,
-            'organization': org_id,
-        }
-    })
+    lookup_data = {'name': inventory}
+    if organization:
+        lookup_data['organization'] = module.resolve_name_to_id('organizations', organization)
+    inventory_object = module.get_one('inventories', data=lookup_data)
 
     if not inventory_object:
         module.fail_json(msg='The specified inventory was not found.')
--- a/awx_collection/plugins/modules/tower_inventory_source.py
+++ b/awx_collection/plugins/modules/tower_inventory_source.py
@@ -131,6 +131,7 @@ options:
     organization:
       description:
         - Name of the inventory source's inventory's organization.
+      required: True
       type: str
 extends_documentation_fragment: awx.awx.auth
 '''
@@ -173,7 +174,7 @@ def main():
         enabled_value=dict(),
         host_filter=dict(),
         credential=dict(),
-        organization=dict(),
+        organization=dict(required=True),
         overwrite=dict(type='bool'),
         overwrite_vars=dict(type='bool'),
         custom_virtualenv=dict(),
@@ -196,29 +197,34 @@ def main():
     name = module.params.get('name')
     new_name = module.params.get('new_name')
     inventory = module.params.get('inventory')
+    organization = module.params.get('organization')
     source_script = module.params.get('source_script')
     credential = module.params.get('credential')
     source_project = module.params.get('source_project')
     state = module.params.get('state')
 
-    new_fields = {}
-    organization_id = None
-    organization = module.params.get('organization')
-    if organization:
-        organization_id = module.get_one_by_name_or_id('organizations', organization)
-
     # Attempt to look up inventory source based on the provided name and inventory ID
-    inventory_id = module.resolve_name_to_id('inventories', inventory)
-    inventory_source = module.get_one('inventory_sources', **{
+    org_id = module.resolve_name_to_id('organizations', organization)
+    inventory_object = module.get_one('inventories', **{
+        'data': {
+            'name': inventory,
+            'organization': org_id,
+        }
+    })
+
+    if not inventory_object:
+        module.fail_json(msg='The specified inventory was not found.')
+
+    inventory_source_object = module.get_one('inventory_sources', **{
         'data': {
             'name': name,
-            'inventory': inventory_id,
+            'inventory': inventory_object['id'],
         }
     })
 
     if state == 'absent':
         # If the state was absent we can let the module delete it if needed, the module will handle exiting from this
-        module.delete_if_needed(inventory_source)
+        module.delete_if_needed(inventory_source_object)
 
     # Attempt to look up associated field items the user specified.
     association_fields = {}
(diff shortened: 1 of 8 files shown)


beeankha:
This module is in need of a `wait` parameter, and is blocked until merge of PR https://github.com/ansible/awx/pull/8022

sean-m-sullivan:
Looks good!, I like the use of the lookup plugin example/tests, it helps to get examples like these out there, especially for lesser known lookup functions. Waiting to review till the wait function is added.

AlanCoding:
This will do the same in effect as #7616, but that's not super obvious to figure out by the names. Would it be better to make the 'update all' functionality some sort of parameter on this module?

awx_collection/plugins/modules/tower_inventory_source_update.py
```diff
@@ -0,0 +1,100 @@
...
+short_description: Update inventory source(s).
+description:
+    - Update Ansible Tower inventory source(s). See
+      U(https://www.ansible.com/tower) for an overview.
+options:
+    inventory:
+      description:
+        - Name of the inventory that contains the inventory source(s) to update.
+      required: True
+      type: str
```
AlanCoding, line 29: Here, again, I think it would make the most sense to add an `organization` field to help with the lookup of the inventory so that you don't get into an unresolvable "duplicates returned" situation.

awx_collection/plugins/modules/tower_inventory_source_update.py
```diff
@@ -0,0 +1,100 @@
...
+extends_documentation_fragment: awx.awx.auth
+'''
+
+EXAMPLES = '''
+- name: Update a single inventory source
+  tower_inventory_source_update:
+    inventory: "My Inventory"
+    inventory_source: "Example Inventory Source"
```
AlanCoding, line 62: Will this wait until the update has finished? Either way, could we get an example of showing how to do that?

awx_collection/tests/integration/targets/tower_inventory_source_update/tasks/main.yml
```diff
@@ -0,0 +1,80 @@
...
+    - assert:
+        that:
+          - "result is changed"
+
+  always:
+    - name: Delete Inventory
+      tower_inventory:
+        name: "{{ inv_name }}"
+        organization: Default
+        state: absent
```
AlanCoding, line 105: If an update is running, then deleting an inventory will return an error code IIRC, so I don't think this would work unless the updates finish or are fully canceled.

awx_collection/plugins/modules/tower_inventory_source_update.py
sean-m-sullivan: I believe she is waiting for the wait from tower_project_update to be merged to add that code.

beeankha: Yep, @sean-m-sullivan is correct!

awx_collection/plugins/modules/tower_inventory_source.py
```diff
@@ -128,6 +128,10 @@
         - list of notifications to send on error
       type: list
       elements: str
+    organization:
+      description:
+        - Name of organization for project.
```
AlanCoding: name of organization for the inventory source's inventory's organization

Also, connect #5193 as that will be resolved by this

awx_collection/plugins/modules/tower_inventory_source.py
```diff
@@ -195,6 +201,12 @@ def main():
...
     state = module.params.get('state')
 
+    new_fields = {}
+    organization_id = None
+    organization = module.params.get('organization')
+    if organization:
+        organization_id = module.get_one_by_name_or_id('organizations', organization)
+
     # Attempt to look up inventory source based on the provided name and inventory ID
     inventory_id = module.resolve_name_to_id('inventories', inventory)
```
AlanCoding: Need to actually make use of `organization_id` in the query for this to make sense.

awx_collection/plugins/modules/tower_inventory_source.py
beeankha: Oops, bad copypasta

awx_collection/plugins/modules/tower_inventory_source.py
AlanCoding: In other cases this was done by using `get_one`, but in this case it seems appropriate for a new helper method that will fail if multiples are returned like `get_exactly_one`

awx_collection/test/awx/test_inventory_source.py
```diff
@@ -45,30 +46,6 @@ def test_inventory_source_create(run_module, admin_user, base_inventory, project
     }
 
 
-@pytest.mark.django_db
-def test_create_inventory_source_implied_org(run_module, admin_user):
```
beeankha: I'm removing this test because there will no longer be implied orgs; the `organization` parameter is now explicitly required by the `tower_inventory_source` module (README has been updated with this info).

sean-m-sullivan (approved):
Looks good and I believe complete.

awx_collection/plugins/modules/tower_inventory_source.py
```diff
@@ -190,23 +197,34 @@ def main():
...
     state = module.params.get('state')
 
     # Attempt to look up inventory source based on the provided name and inventory ID
-    inventory_id = module.resolve_name_to_id('inventories', inventory)
-    inventory_source = module.get_one('inventory_sources', **{
+    org_id = module.resolve_name_to_id('organizations', organization)
+    inventory_object = module.get_one('inventories', **{
+        'data': {
+            'name': inventory,
+            'organization': org_id,
```
AlanCoding: wouldn't this be a problem in the event that the user didn't provide `organization`? In that case you get `None` for this, but in the query params I think that would cause it to filter for inventories with a null organization, which is probably 0.

awx_collection/tests/integration/targets/tower_inventory_source_update/tasks/main.yml
```diff
@@ -0,0 +1,116 @@
...
+      register: created_inventory
+
+    - name: Create an Inventory Source
+      tower_inventory_source:
+        name: "{{ inv_source1 }}"
+        source: scm
+        source_project: "{{ project_name }}"
+        source_path: inventories/inventory.ini
+        description: Source for Test inventory
+        organization: "{{ created_org.id }}"
```
AlanCoding, line 51: If the next task is for testing org-based lookup, I think it makes sense to not provide `organization:` here. The inventory name should be sufficiently unique.

beeankha: `organization` is now a required parameter for the inventory source module so I can't take this out.  I forgot to update the name for the next task in this playbook.

beeankha: I’ve updated the module so that the org parameter isn’t required anymore, but I _do_ have this integration test create two same-named inventories associated with two separate orgs.  Thus, if I don’t specify the organization here, too many results get returned from the API.

awx_collection/plugins/modules/tower_inventory_source.py
```diff
@@ -190,23 +196,30 @@ def main():
...
-    # Attempt to look up inventory source based on the provided name and inventory ID
-    inventory_id = module.resolve_name_to_id('inventories', inventory)
-    inventory_source = module.get_one('inventory_sources', **{
+    lookup_data = {'name': inventory}
+    if organization:
+        lookup_data['organization'] = module.resolve_name_to_id('organizations', organization)
+    inventory_object = module.get_one('inventories', data=lookup_data)
+
+    if not inventory_object:
+        module.fail_json(msg='The specified inventory was not found.')
```
AlanCoding: Non-blocking comment, but I think it would be helpful to put `lookup_data` in this message with a `.format`.

AlanCoding (approved):
awesome work!

AlanCoding:
recheck

Issue #7467, opened 2020-06-29
Instance Groups


nixocio:
Issue to track work related to Instance Groups/Container groups.

 1. Modify routing system to add those new routes. https://github.com/ansible/awx/issues/7471
 2. List Instance Groups https://github.com/ansible/awx/issues/7557

The next issues need to differentiate `instance groups` and `container groups`.

Instance groups:

 3. Add Instance Groups https://github.com/ansible/awx/issues/7744
 4. Edit Instance Groups https://github.com/ansible/awx/issues/7767
 5. Delete Instance Groups https://github.com/ansible/awx/issues/7642
 6. Details page for Instance Groups - https://github.com/ansible/awx/issues/7640
 7. Associate instance with instance groups - https://github.com/ansible/awx/issues/7801
 8. List jobs related to instances - https://github.com/ansible/awx/issues/7930
 9. Add controller label - https://github.com/ansible/awx/issues/8244
10. Add slider to instance groups - https://github.com/ansible/awx/issues/7777

Container groups:

 1. Add/Edit Container Groups - https://github.com/ansible/awx/issues/7955
 2. Details page for Container Groups - https://github.com/ansible/awx/issues/8073
 3. Delete Container Groups - It was created as a side effect of the other items.
 4. List jobs related to Container Groups - https://github.com/ansible/awx/issues/8091

See : https://tower-mockups.testing.ansible.com/patternfly/instance-groups/instance-groups/
To be expanded.

unlikelyzero:
@nixocio do you want to close this to track the individual work in the remaining tickets?

unlikelyzero:
note: this was discovered https://github.com/ansible/awx/issues/9191

unlikelyzero:
Closing as everything is handle as an individual ticket

PR #8035, opened 2020-08-28
Update instance groups

nixocio:
* Simplify criteria to instance group to be considered unavailable
* Round values for used capacity

See: https://github.com/ansible/awx/issues/7467


nixocio:
Now that the PR to associate/disassociate instances was merged I did a full comparison with the current UI implementation and I added those updates.

dsesami (approved):

jakemcdermott (approved):

PR #8041, opened 2020-08-28
Adds support for pending deletion on inventory list

### LINKED ISSUE #7680: Support Pending Delete on inventory list rows
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Since it can take a really long time to delete inventories, the API will actually mark inventories using the `pending_deletion` field to let us know that they are in the process of deleting an inventory.  When we detect a row with this field set to true we should remove all the action icons in favor of some text that says `Pending delete` or `Pending deletion`.  We probably also want to generally disable the row as users won't be able to take any actions on it while it's in this state.

mabashian:
##### SUMMARY
link #7680 

This issue is a bit nuanced because it breaks from our normal delete pattern.  A couple of notes:

1) We're no longer refreshing the list after issuing successful delete requests.  This is because inventories are placed in a `pending_deletion` state after the delete request is processed.  Instead, we'll wait for a socket message that tells us that an inventory has been deleted.
2) We're now updating the `pending_deletion` field on inventory rows based on socket messages.
3) As stated in 1, we're now performing a refresh of the list only when we get a socket message saying that an inventory has been deleted.  This request should take pagination into account.
4) I added some warning text on the delete modal indicating that the inventory(ies) would be placed in a pending delete state.

List:

Screen Shot 2020-08-28 at 2 24 51 PM

Modal:

Screen Shot 2020-08-28 at 2 24 10 PM

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/screens/Inventory/InventoryList/InventoryListItem.jsx
+++ b/awx/ui_next/src/screens/Inventory/InventoryList/InventoryListItem.jsx
@@ -8,6 +8,7 @@ import {
   DataListItem,
   DataListItemCells,
   DataListItemRow,
+  Label,
   Tooltip,
 } from '@patternfly/react-core';
 import { PencilAltIcon } from '@patternfly/react-icons';
@@ -69,6 +70,7 @@ function InventoryListItem({
       <DataListItemRow>
         <DataListCheck
           id={`select-inventory-${inventory.id}`}
+          isDisabled={inventory.pending_deletion}
           checked={isSelected}
           onChange={onSelect}
           aria-labelledby={labelId}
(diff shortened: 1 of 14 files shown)


velzend:
Nice feature...

awx/ui_next/src/screens/Inventory/InventoryList/InventoryListItem.jsx
```diff
@@ -79,52 +81,63 @@ function InventoryListItem({
...
-              onDoneLoading={() => setIsDisabled(false)}
-              helperText={{
-                tooltip: i18n._(t`Copy Inventory`),
-                errorMessage: i18n._(t`Failed to copy inventory.`),
-              }}
-            />
-          )}
-        </DataListAction>
+                onLoading={() => setIsDisabled(true)}
+                onDoneLoading={() => setIsDisabled(false)}
```
keithjgrant: These should be wrapped in useCallback to keep them hooks-safe (CopyButton references them in a useEffect)

awx/ui_next/src/screens/Inventory/InventoryList/useWsInventories.js
```diff
@@ -64,16 +75,41 @@ export default function useWsProjects(
...
+          return;
+        }
+      } else {
+        if (!['pending', 'waiting', 'running'].includes(lastMessage.status)) {
+          enqueueId(lastMessage.inventory_id);
+          return;
+        }
+
+        updatedInventory.isSourceSyncRunning = true;
+      }
```
keithjgrant: A lot of nested if/else can be tricky to to follow. Maybe move part of this into a helper function(s), or maybe handle cases one at a time, returning out of this function when possible (so you don't need an `else`)?

keithjgrant (changes requested):
Looks good, just a couple fixes

mabashian:
@keithjgrant OK I flattened out those nested if/else's into individual cases.  If that's not what you were expecting then let me know and I'll try again.

I also came across a bug where new rows weren't fetched after copying an inventory _if_ an inventory source sync had finished before attempting to copy.  This was due to `id__in` being added to the query params.  I went ahead and fixed that in a separate commit to this PR.

unlikelyzero:
@dsesami this is ready for test

keithjgrant (approved):
looks great!

dsesami:
@mabashian this needs a rebase, then I can knock it out pretty quick.

dsesami:
I managed to crash the page with this console error: 
```Failed to load resource: the server responded with a status of 409 (Conflict)```

Recreate steps:
1. Fire up a JT that uses ansible/test-playbooks as its project. I used `ping-20.yml`, but any task that takes 5 seconds or more will do fine. 
2. Open two tabs. One tab will launch the JT, the other is the inventory view.
3. Launch the JT, then immediately try to delete the inventory while the Job is running. The page will crash.

dsesami (changes requested):
see above comment

dsesami:
spoke with mike -- same bug as above, but with the details page's delete button instead of the list delete button.

unlikelyzero:
@dsesami @mabashian in that case, do you think we should move it out of this ticket into its own bug or address it right now in this PR?

mabashian:
@unlikelyzero @dsesami nah I can hopefully fix it pretty quickly.  If it gets messy then yea lets peel it out.

dsesami (approved):
Works great.

Issue #8011, opened 2020-08-26
Create a tower_instance_group module


john-westcott-iv:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Per PR #7963 we are missing a module to interact with the instance_groups endpoint

appuk:
@beeankha I was able to create and edit an instance group using this module, but when I tried to delete the object using `state: absent`, it did not delete the instance group, and the task passed as ok

ryanpetrello:
@beeankha or @john-westcott-iv can one of you take a peek at this as you have available time?

jbradberry:
After taking a quick look at the merged PR, it looks to me as if the problem is a minor typo, `if state is 'absent'` instead of `if state == 'absent'`.

ryanpetrello:
@appuk can you give this another try?

appuk:
The instance group is deleted on stating `state: absent`, Hence closing!

PR #8043, opened 2020-08-28
Adding tower_instance_group module

john-westcott-iv:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - Collection

##### AWX VERSION

```
awx: 14.1.0
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx_collection/plugins/modules/tower_instance_group.py
+++ b/awx_collection/plugins/modules/tower_instance_group.py
@@ -1,0 +1,157 @@
+#!/usr/bin/env python
+# coding: utf-8 -*-
+
+
+# (c) 2020, John Westcott IV <john.westcott.iv@redhat.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+---
+module: tower_instance_group
+author: "John Westcott IV (@john-westcott-iv)"
+version_added: "4.0"
+short_description: create, update, or destroy Ansible Tower instance groups.
+description:
+    - Create, update, or destroy Ansible Tower instance groups. See
+      U(https://www.ansible.com/tower) for an overview.
+options:
+    name:
+      description:
+        - Name of this instance group.
+      required: True
+      type: str
+    new_name:
+      description:
+        - Setting this option will change the existing name (looked up via the name field.
+      required: True
+      type: str
+    credential:
+      description:
+        - Credential to authenticate with Kubernetes or OpenShift.  Must be of type "Kubernetes/OpenShift API Bearer Token”.
+      required: False
+      type: str
+    policy_instance_percentage:
+      description:
+        - Minimum percentage of all instances that will be automatically assigned to this group when new instances come online.
+      required: False
+      type: int
+      default: '0'
+    policy_instance_minimum:
+      description:
+        - Static minimum number of Instances that will be automatically assign to this group when new instances come online.
+      required: False
+      type: int
+      default: '0'
+    policy_instance_list:
+      description:
+        - List of exact-match Instances that will be assigned to this group
+      required: False
+      type: dict
+    pod_spec_override:
+      description:
+        - A custom Kubernetes or OpenShift Pod specification.
+      required: False
+      type: str
+    instances:
+      description:
+        - The instances associated with this instance_group
+      required: False
+      type: list
+    state:
+      description:
+        - Desired state of the resource.
+      choices: ["present", "absent"]
+      default: "present"
+      type: str
+extends_documentation_fragment: awx.awx.auth
+'''
+
+EXAMPLES = '''
+'''
+
+from ..module_utils.tower_api import TowerAPIModule
+
+
+def main():
+    # Any additional arguments that are not fields of the item can be added here
+    argument_spec = dict(
+        name=dict(required=True),
+        new_name=dict(),
+        credential=dict(),
+        policy_instance_percentage=dict(type='int', default='0'),
+        policy_instance_minimum=dict(type='int', default='0'),
+        policy_instance_list=dict(type='dict'),
+        pod_spec_override=dict(),
+        instances=dict(required=False, type="list", default=None),
+        state=dict(choices=['present', 'absent'], default='present'),
+    )
+
+    # Create a module for ourselves
+    module = TowerAPIModule(argument_spec=argument_spec)
+
+    # Extract our parameters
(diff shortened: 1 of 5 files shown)


awx_collection/tests/integration/targets/tower_instance_group/tasks/main.yml
```diff
@@ -0,0 +1,63 @@
...
+      register: result
+
+    - assert:
+        that:
+          - "result is changed"
+
+    - name: Create a container group
+      tower_instance_group:
+        name: "{{ group_name2 }}"
+        credential: "{{ cred_name1 }}"
```
AlanCoding, line 43: @shanemcd does that look good to you? (I'm not as familiar with the API for this)

awx_collection/test/awx/test_instance_group.py
```diff
@@ -0,0 +1,69 @@
...
+    assert ig.policy_instance_percentage == 34
+    assert ig.policy_instance_minimum == 12
+
+    # Create a new instance in the DB
+    new_instance = Instance.objects.create(hostname='foo.example.com')
+
+    # Set the new instance group only to the one instnace
+    result = run_module('tower_instance_group', {
+        'name': 'foo-group',
+        'instances': new_instance.hostname,
```
AlanCoding: This is weird, instances plural, but you give it a single string, not a list?

awx_collection/plugins/modules/tower_instance_group.py
```diff
@@ -0,0 +1,157 @@
...
+      description:
+        - Static minimum number of Instances that will be automatically assign to this group when new instances come online.
+      required: False
+      type: int
+      default: '0'
+    policy_instance_list:
+      description:
+        - List of exact-match Instances that will be assigned to this group
+      required: False
+      type: dict
```
AlanCoding: why is this a dict, as opposed to a list?

awx_collection/plugins/modules/tower_instance_group.py
```diff
@@ -0,0 +1,157 @@
...
+    pod_spec_override:
+      description:
+        - A custom Kubernetes or OpenShift Pod specification.
+      required: False
+      type: str
+    instances:
+      description:
+        - The instances associated with this instance_group
```
AlanCoding, line 65: We should get some info about the durability of this change. Does it remove instances, or just add them?

AlanCoding: Also, when will the policy rules undo this...

beeankha (approved):
Some `pep8`/`pylint`/sanity stuff needs to get fixed, otherwise I think this PR is good to go

AlanCoding:
This looks good, we just need to get `tower_instance_group` hooked up to our integration testing and verified before merge.

awx_collection/plugins/modules/tower_instance_group.py
john-westcott-iv: It was kid of unclear what this does since there is also the "normal" association pattern as well.

awx_collection/plugins/modules/tower_instance_group.py
john-westcott-iv: That came from the API:
```
            "policy_instance_list": {
                "type": "json",
                "required": false,
                "label": "Policy Instance List",
                "help_text": "List of exact-match Instances that will be assigned to this group",
                "filterable": true,
                "child": {
                    "type": "string",
                    "required": true,
                    "read_only": false
                }
            },
```

awx_collection/plugins/modules/tower_instance_group.py
john-westcott-iv: This just flows into our standard association pattern.

awx_collection/test/awx/test_instance_group.py
john-westcott-iv: I'm thinking Ansible might be converting the string into an array on the fly? I'll update this so that it comes in as an Array to better match the type.

awx_collection/plugins/modules/tower_instance_group.py
AlanCoding: ah, so yeah, I think that's because the API doesn't actually have a list type. My concern here is whether this is actually a usable field. The API really needs a list, but I don't know for sure that Ansible itself will allow a list in a dict field. I might see if I can suggest an example for this real quick and also determine if it does or doesn't work.

AlanCoding: Yeah, this is what I hit

```
TASK [Create an instance group with list] *******************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "argument policy_instance_list is of type <class 'list'> and we were unable to convert to dict: <class 'list'> cannot be converted to a dict"}
```

from

```yaml
    - name: Create an instance group with list
      tower_instance_group:
        name: "alan instance group"
        policy_instance_list:
          - awx
```

I think that the easy solution here is to make it the a list type. Tested with

```diff
diff --git a/awx_collection/plugins/modules/tower_instance_group.py b/awx_collection/plugins/modules/tower_instance_group.py
index 6424be4b3b..9a7db2f814 100644
--- a/awx_collection/plugins/modules/tower_instance_group.py
+++ b/awx_collection/plugins/modules/tower_instance_group.py
@@ -54,7 +54,7 @@ options:
       description:
         - List of exact-match Instances that will be assigned to this group
       required: False
-      type: dict
+      type: list
     pod_spec_override:
       description:
         - A custom Kubernetes or OpenShift Pod specification.
@@ -88,7 +88,7 @@ def main():
         credential=dict(),
         policy_instance_percentage=dict(type='int', default='0'),
         policy_instance_minimum=dict(type='int', default='0'),
-        policy_instance_list=dict(type='dict'),
+        policy_instance_list=dict(type='list'),
         pod_spec_override=dict(),
         instances=dict(required=False, type="list", default=None),
         state=dict(choices=['present', 'absent'], default='present'),
```

and that seems to work fine.

AlanCoding (changes requested):
I gave a diff for policy_instance_list, other than that, all tests are passing. If we can get that change I would want to merge.

beeankha:
Officially connecting the associated AWX issue https://github.com/ansible/awx/issues/8011

awx_collection/test/awx/test_instance_group.py
```diff
@@ -0,0 +1,71 @@
...
+    all_instance_names = []
+    for instance in ig.instances.all():
+        all_instance_names.append(instance.hostname)
+
+    assert new_instance.hostname in all_instance_names, 'Failed to add instance to group'
+    assert len(all_instance_names) == 1, 'Too many instances in group {0}'.format(','.join(all_instance_names))
+
+
+@pytest.mark.django_db
+def test_continer_group_create(run_module, admin_user, kube_credential):
```
piroux: typo ;)

AlanCoding (approved):

PR #8047, opened 2020-08-31
Remove more special access to folders outside job private_data_dir

AlanCoding:
##### SUMMARY
This removes 2 references to locations that no longer exist in source:

 - awx/plugins/callback
 - awx/plugins/inventory
 - (later addition) awx/lib

The first one was deleted when the integration with `ansible-runner` was done

The second folder was deleted when the transition to inventory plugins, from inventory scripts, was completed.

This does 1 more thing, it copies the Ansible library directory for use in fact caching to the job private data directory. This avoids one more need for the job to access things outside of its private data directory at a relatively trivial cost. It is really needed for running jobs in containers, but like #8029, I thought it was cheap enough that we could make it a standard pattern now so that we further narrow down what we need to support.

##### ISSUE TYPE
 - Tech debt

##### COMPONENT NAME
 - API

##### AWX VERSION
```
14.1.0
```

##### ADDITIONAL INFORMATION
As of right now, still getting test data, but turnaround should be fast, will take it out of draft mode then.

### DIFF
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -1630,21 +1630,10 @@ class RunJob(BaseTask):
 
         return passwords
 
-    def add_ansible_venv(self, venv_path, env, isolated=False):
-        super(RunJob, self).add_ansible_venv(venv_path, env, isolated=isolated)
-        # Add awx/lib to PYTHONPATH.
-        env['PYTHONPATH'] = env.get('PYTHONPATH', '') + self.get_path_to('..', 'lib') + ':'
-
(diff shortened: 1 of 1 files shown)


ghjm (approved):
To the extent I understand this, LGTM

jbradberry (changes requested):
Capturing it here for posterity, but I found and we discussed in Slack a couple more uses of `self.get_path_to('..', 'plugins', 'callback')` that should get axed.

jbradberry (approved):
Seems good to me.

PR #8057, opened 2020-09-01
update the project archive sync support for py2 compatability

ryanpetrello:
playbook runs in AWX are still supported in python2, so this adds
support for archive-based project syncs in py2

cc @philipsd6

### DIFF
--- a/awx/playbooks/action_plugins/project_archive.py
+++ b/awx/playbooks/action_plugins/project_archive.py
@@ -2,14 +2,20 @@ from __future__ import absolute_import, division, print_function
 
 __metaclass__ = type
 
-import zipfile
-import tarfile
+import errno
 import os
+import tarfile
+import zipfile
 
 from ansible.plugins.action import ActionBase
 from ansible.utils.display import Display
 
 display = Display()
+
+try:
+    from zipfile import BadZipFile
+except ImportError:
+    from zipfile import BadZipfile as BadZipFile  # py2 compat
 
 
 class ActionModule(ActionBase):
@@ -26,14 +32,15 @@ class ActionModule(ActionBase):
             archive = zipfile.ZipFile(src)
             get_filenames = archive.namelist
             get_members = archive.infolist
-        except zipfile.BadZipFile:
-            archive = tarfile.open(src)
+        except BadZipFile:
+            try:
+                archive = tarfile.open(src)
+            except tarfile.ReadError:
+                result["failed"] = True
+                result["msg"] = "{0} is not a valid archive".format(src)
+                return result
             get_filenames = archive.getnames
             get_members = archive.getmembers
-        except tarfile.ReadError:
-            result["failed"] = True
-            result["msg"] = "{0} is not a valid archive".format(src)
-            return result
 
         # Most well formed archives contain a single root directory, typically named
         # project-name-1.0.0. The project contents should be inside that directory.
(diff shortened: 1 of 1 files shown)


awx/playbooks/action_plugins/project_archive.py
```diff
@@ -26,14 +32,15 @@ def run(self, tmp=None, task_vars=None):
             archive = zipfile.ZipFile(src)
             get_filenames = archive.namelist
             get_members = archive.infolist
-        except zipfile.BadZipFile:
-            archive = tarfile.open(src)
+        except BadZipFile:
+            try:
+                archive = tarfile.open(src)
+            except tarfile.ReadError:
```
ryanpetrello, line 38: This lack of indentation was actually a legitimate bug we just missed in code review.

awx/playbooks/action_plugins/project_archive.py
```diff
@@ -2,15 +2,21 @@
...
 
 from ansible.plugins.action import ActionBase
 from ansible.utils.display import Display
 
 display = Display()
 
+try:
+    from zipfile import BadZipFile
+except ImportError:
+    from zipfile import BadZipfile as BadZipFile  # py2 compat
```
ryanpetrello, line 18: Booooooooooooo python2

awx/playbooks/action_plugins/project_archive.py
```diff
@@ -62,10 +69,19 @@ def run(self, tmp=None, task_vars=None):
             try:
                 is_dir = member.is_dir()
             except AttributeError:
-                is_dir = member.isdir()
+                try:
+                    is_dir = member.isdir()
+                except AttributeError:
+                    is_dir = member.filename[-1] == '/'  # py2 compat for ZipInfo
```
ryanpetrello, line 75: python2's `zipfile` doesn't have this, but this is what py3 does, so I just copied it here <shrug>

awx/playbooks/action_plugins/project_archive.py
```diff
@@ -62,10 +69,19 @@ def run(self, tmp=None, task_vars=None):
...
 
             if is_dir:
-                os.makedirs(dest, exist_ok=True)
+                try:
+                    os.makedirs(dest)
```
ryanpetrello, line 79: poor man's `exist_ok=True` (which also doesn't exist in py2)

kdelee (approved):
:+1: fixes the tests that were failing, works for me!

philipsd6:
@ryanpetrello I had the understanding that AWX was solidly Python3 at this point, and even if a project was associated with a Py2 venv, the project sync would be done with the default Py3, no? Seems unreasonable for the built ins to not be able to depend on the main Python version! In any case, as my understanding was apparently wrong, thanks for catching this and fixing it!

P.S. I'm **so** tired of writing backwards compatible Python... Sometimes I even get bit when using `{k: v for k, v in ...}` instead of `dict((k, v) for k, v in ...)`

ryanpetrello:
Hey @philipsd6,

Hey no big deal.  I'm a maintainer and I forgot too 😂.

When a project update runs, it uses the default "Ansible" virtualenv, and we still technically support systems (like RHEL7) that have python2 as their default system Python for _playbook execution_.

The Python code that AWX itself runs in _its_ various processes is all python3 only, though.

PR #8061, opened 2020-09-02
Purge environment variables to work with ansible-runner changes

AlanCoding:
These changes correspond to https://github.com/ansible/ansible-runner/pull/507, but they shouldn't break anything if that is not merged first either.


jbradberry (approved):
This seems to make sense to me.

AlanCoding:
recheck

PR #8064, opened 2020-09-02
Respect user proot show paths when using containers

AlanCoding:
##### SUMMARY
I'm actually using this to fix some tests, but it's also necessary to support in general.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
20.0.0
```

### DIFF
--- a/awx/main/models/credential/__init__.py
+++ b/awx/main/models/credential/__init__.py
@@ -559,7 +559,12 @@ class CredentialType(CommonModelNameNotUnique):
 
             if extra_vars:
                 path = build_extra_vars_file(extra_vars, private_data_dir)
-                args.extend(['-e', '@%s' % path])
+                # FIXME: develop some better means of referencing paths inside containers
+                container_path = os.path.join(
+                    '/runner',
+                    os.path.basename(path)
+                )
+                args.extend(['-e', '@%s' % container_path])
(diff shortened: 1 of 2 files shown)


awx/main/models/credential/__init__.py
```diff
@@ -559,7 +559,12 @@ def build_extra_vars_file(vars, private_dir):
 
             if extra_vars:
                 path = build_extra_vars_file(extra_vars, private_data_dir)
-                args.extend(['-e', '@%s' % path])
+                # FIXME: develop some better means of referencing paths inside containers
+                container_path = os.path.join(
+                    '/runner',
+                    os.path.basename(path)
+                )
+                args.extend(['-e', '@%s' % container_path])
```
AlanCoding, line 567: @ryanpetrello just to let you know what's going on for running jobs in containers. We can't ever reference any absolute paths 😝 

This won't be the last of this kind of finding... but clearly since this broke all credentials that used extra vars its impact was relatively high. Just off the top of my head, I would guess that credentials that use an env var to reference a file would also not work.

AlanCoding:
Closing and pushing to main at https://github.com/ansible/awx/pull/8030

PR #8069, opened 2020-09-03
add the ability to prevent SAML auto-population behavior

ryanpetrello:


awx/sso/conf.py
```diff
@@ -919,6 +919,17 @@ def get_saml_entity_id():
     return settings.TOWER_URL_BASE
 
 
+register(
+    'SAML_AUTO_CREATE_OBJECTS',
+    field_class=fields.BooleanField,
+    default=True,
+    label=_('SAML Assertion Consumer Service (ACS) URL'),
```
wenottingham: c&p leftover

one-t (approved):
LGTM, I suppose the use case here is that they want to manually create the groups and teams in tower and then manage the membership using SAML attributes. Presumably because they have a bunch of non-tower related groups that are getting auto-created. This could also be accomplished by selectively releasing them from the IDP side (since you can dictate what gets released to each application).

Issue #7955, opened 2020-08-20
Add/Edit container group


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Add/Edit container group.

See: https://tower-mockups.testing.ansible.com/patternfly/instance-groups/instance-groups-container-add/

nixocio:
I ended up working in adding/editing a container group at the same time, since they are highly coupled - sharing the same form.

mabashian:
@nixocio is this issue still valid?  I see that #8070 has merged.  Should this be closed?

nixocio:
@mabashian, @dsesami verified this one. Can we close this one @dsesami?

dsesami:
This was captured within the scope of other verification and was stubbed out etc., this is good to close.

PR #8070, opened 2020-09-03
Add/Edit Container Groups

nixocio:
Add/Edit container groups.

See: https://github.com/ansible/awx/issues/7955

image

### DIFF
--- a/awx/ui_next/src/screens/InstanceGroup/shared/ContainerGroupForm.jsx
+++ b/awx/ui_next/src/screens/InstanceGroup/shared/ContainerGroupForm.jsx
@@ -1,0 +1,144 @@
+import React from 'react';
+import { func, shape } from 'prop-types';
+import { Formik, useField } from 'formik';
+import { withI18n } from '@lingui/react';
+import { t } from '@lingui/macro';
+import { Form, FormGroup } from '@patternfly/react-core';
+import { jsonToYaml } from '../../../util/yaml';
+
+import FormField, {
+  FormSubmitError,
+  CheckboxField,
+} from '../../../components/FormField';
+import FormActionGroup from '../../../components/FormActionGroup';
+import { required } from '../../../util/validators';
+import {
+  FormColumnLayout,
+  FormFullWidthLayout,
+  FormCheckboxLayout,
+  SubFormLayout,
+} from '../../../components/FormLayout';
+import CredentialLookup from '../../../components/Lookup/CredentialLookup';
+import { VariablesField } from '../../../components/CodeMirrorInput';
+
+function ContainerGroupFormFields({ i18n }) {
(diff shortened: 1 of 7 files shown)


awx/ui_next/src/screens/InstanceGroup/shared/ContainerGroupForm.jsx
```diff
@@ -0,0 +1,144 @@
...
+        tooltip={i18n._(
+          t`Credential to authenticate with Kubernetes or OpenShift.  Must be of type "Kubernetes/OpenShift API Bearer Token”.`
+        )}
+      />
+
+      <FormGroup
+        fieldId="container-groups-option-checkbox"
+        label={i18n._(t`Options`)}
+      >
+        <FormCheckboxLayout>
```
jakemcdermott, line 60: @nixocio 

In terms of functionality, this is very similar to the notification config overrides we do on the notification form and the UX should be consistent for both features.

- In the 3.8 ui we use a toggle-switch for container groups and notifications
- The mockups [show a checkbox](https://tower-mockups.testing.ansible.com/patternfly/instance-groups/instance-groups-container-edit/) for container groups but don't specify any UX for notifications
- [We intend to use a switch](https://github.com/ansible/awx/pull/7967/files#) for notifications in ui_next


Personally, I prefer the toggle switch but I think the important thing is that we're consistent with the type of input field we use for configuration overrides.

cc @trahman73 @mabashian @keithjgrant  thoughts?

jakemcdermott (changes requested):
The code, functionality, and tests look good to me. Nice work!

I would like for https://github.com/ansible/awx/pull/8070#discussion_r483779879 to be resolved (in some fashion) prior to merging. Depending on what is decided, this may mean changing the override field to a toggle.

awx/ui_next/src/screens/InstanceGroup/shared/ContainerGroupForm.jsx
nixocio: There was a slack discussion about the need to use `checkbox`. I think the mock up was updated upon the outcome of that discussion.

jakemcdermott: :+1: Nice. Sounds good to me.

jakemcdermott (approved):

jakemcdermott (approved):
https://github.com/ansible/awx/pull/8070#discussion_r483803623

dsesami:
I'm sitting with @nixocio on Monday 9/14 to discuss this PR in detail.

dsesami (approved):
Covered this at length with @nixocio . Looks great. :+1: Will follow up on some improvement ideas at dev sync.

PR #8072, opened 2020-09-03
Modify get_one method to allow IDs in addition to names

john-westcott-iv:
##### SUMMARY

The get_one method took a data parameter with search criteria to find an item. It is used in many of the modules. Based on its current implementation one would do something like:
```
data: {
    'name': item_name,
    'org': org_id,
}
```
This does not lend well with being able to lookup by name or ID which is something we want the collection to be able to do.

This change introduces a parameter called name_or_id which will modify the behavior of get_one to properly search on either a name or an ID.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - Collection

##### AWX VERSION

```
awx: 14.1.0
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx_collection/plugins/module_utils/tower_api.py
+++ b/awx_collection/plugins/module_utils/tower_api.py
@@ -52,20 +52,24 @@ class TowerAPIModule(TowerModule):
         return TowerAPIModule.IDENTITY_FIELDS.get(endpoint, 'name')
 
     def get_item_name(self, item, allow_unknown=False):
-        if 'name' in item:
-            return item['name']
+        if item:
+            if 'name' in item:
+                return item['name']
 
-        for field_name in TowerAPIModule.IDENTITY_FIELDS.values():
-            if field_name in item:
-                return item[field_name]
+            for field_name in TowerAPIModule.IDENTITY_FIELDS.values():
+                if field_name in item:
+                    return item[field_name]
 
-        if item.get('type', None) in ('o_auth2_access_token', 'credential_input_source'):
-            return item['id']
+            if item.get('type', None) in ('o_auth2_access_token', 'credential_input_source'):
+                return item['id']
 
         if allow_unknown:
             return 'unknown'
 
-        self.exit_json(msg='Cannot determine identity field for {0} object.'.format(item.get('type', 'unknown')))
+        if item:
+            self.exit_json(msg='Cannot determine identity field for {0} object.'.format(item.get('type', 'unknown')))
+        else:
+            self.exit_json(msg='Cannot determine identity field for Undefined object.')
 
     def head_endpoint(self, endpoint, *args, **kwargs):
         return self.make_request('HEAD', endpoint, **kwargs)
--- a/awx_collection/plugins/module_utils/tower_api.py
+++ b/awx_collection/plugins/module_utils/tower_api.py
@@ -25,6 +25,11 @@ class TowerAPIModule(TowerModule):
     }
     session = None
     cookie_jar = CookieJar()
+    IDENTITY_FIELDS = {
+        'users': 'username',
+        'workflow_job_template_nodes': 'identifier',
+        'instances': 'hostname'
+    }
 
     def __init__(self, argument_spec, direct_params=None, error_callback=None, warn_callback=None, **kwargs):
         kwargs['supports_check_mode'] = True
@@ -44,30 +49,20 @@ class TowerAPIModule(TowerModule):
 
     @staticmethod
     def get_name_field_from_endpoint(endpoint):
-        if endpoint == 'users':
-            return 'username'
-        elif endpoint == 'instances':
-            return 'hostname'
-        return 'name'
+        return TowerAPIModule.IDENTITY_FIELDS.get(endpoint, 'name')
 
-    @staticmethod
-    def get_item_name(item):
+    def get_item_name(self, item):
         if 'name' in item:
             return item['name']
-        elif 'username' in item:
-            return item['username']
-        elif 'identifier' in item:
-            return item['identifier']
-        elif 'hostname' in item:
-            return item['hostname']
-        elif item.get('type', None) == 'o_auth2_access_token':
-            # An oauth2 token has no name, instead we will use its id for any of the messages
-            rerurn item['id']
-        elif item.get('type', None) == 'credential_input_source':
-            # An credential_input_source has no name, instead we will use its id for any of the messages
-            return existing_item['id']
-        return 'Unknown'
 
+        for field_name in TowerAPIModule.IDENTITY_FIELDS.values():
+            if field_name in item:
+                return item[field_name]
+
+        if item.get('type', None) in ('o_auth2_access_token', 'credential_input_source'):
+            return item['id']
+
+        self.exit_json(msg='Cannot determine identity field for {0} object.'.format(item.get('type', 'unknown')))
 
     def head_endpoint(self, endpoint, *args, **kwargs):
         return self.make_request('HEAD', endpoint, **kwargs)
@@ -141,18 +136,18 @@ class TowerAPIModule(TowerModule):
             self.fail_json(msg="The endpoint did not provide count and results")
 
         if response['json']['count'] == 0:
-            return None, name_or_id
+            return None
         elif response['json']['count'] > 1:
             if name_or_id:
                 # Since we did a name or ID search and got > 1 return something if the id matches
                 for asset in response['json']['results']:
                     if asset['id'] == name_or_id:
-                        return asset, asset['id'][name_field]
+                        return asset
             # We got > 1 and either didn't find something by ID (which means multiple names)
             # Or we weren't running with a or search and just got back too many to begin with.
             self.fail_json(msg="An unexpected number of items was returned from the API ({0})".format(response['json']['count']))
 
-        return response['json']['results'][0], response['json']['results'][0][name_field]
+        return response['json']['results'][0]
 
     def get_one_by_name_or_id(self, endpoint, name_or_id):
         name_field = self.get_name_field_from_endpoint(endpoint)
--- a/awx_collection/plugins/module_utils/tower_api.py
+++ b/awx_collection/plugins/module_utils/tower_api.py
@@ -51,7 +51,7 @@ class TowerAPIModule(TowerModule):
     def get_name_field_from_endpoint(endpoint):
         return TowerAPIModule.IDENTITY_FIELDS.get(endpoint, 'name')
 
-    def get_item_name(self, item):
+    def get_item_name(self, item, allow_unknown=False):
         if 'name' in item:
             return item['name']
 
@@ -61,6 +61,9 @@ class TowerAPIModule(TowerModule):
 
         if item.get('type', None) in ('o_auth2_access_token', 'credential_input_source'):
             return item['id']
+
+        if allow_unknown:
+            return 'unknown'
 
         self.exit_json(msg='Cannot determine identity field for {0} object.'.format(item.get('type', 'unknown')))
 
@@ -358,7 +361,7 @@ class TowerAPIModule(TowerModule):
                 item_url = existing_item['url']
                 item_type = existing_item['type']
                 item_id = existing_item['id']
-                item_name = self.get_item_name(existing_item)
+                item_name = self.get_item_name(existing_item, allow_unknown=True)
             except KeyError as ke:
                 self.fail_json(msg="Unable to process delete of item due to missing data {0}".format(ke))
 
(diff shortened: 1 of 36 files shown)


awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -88,7 +88,28 @@ def get_all_endpoint(self, endpoint, *args, **kwargs):
             response['json']['next'] = next_page
         return response
 
-    def get_one(self, endpoint, *args, **kwargs):
+    def get_name_field_from_endpoint(self, endpoint):
+        if endpoint == 'users':
+            return 'username'
+        elif endpoint == 'instances':
+            return 'hostname'
+        return 'name'
```
AlanCoding: You're cycling over potential identity fields, but that's the same thing done in other cases like this:

https://github.com/ansible/awx/blob/c4393c4827691139c473b3eaa0a05c2a2b4e6aac/awx_collection/plugins/module_utils/tower_api.py#L349-L360

Your implementation manages to reuse `get_name_field_from_endpoint` in 2 places, but this logic is still very fractured and duplicated.

My thoughts on implementation here is that you probably want a global OrderedDict or simple list of tuples so that you can cycle over them in both cases. That way, you can map the field name exceptions to their endpoints, users-username, instances-hostname, workflow_job_template_nodes-identifier.

I really want this helper method to be totally comprehensive, in a way that will cause it to get used over and over again. Probably a lot. You could have a base method that returns the field name (like username) and wrap that with another method that returns the value (like joe).

beeankha:
recheck

AlanCoding:
This is the direction I would try to go in, I can polish it up to be a usable branch if you want. For now, I've just verified that the credential unit tests pass:

```diff
diff --git a/awx_collection/plugins/module_utils/tower_api.py b/awx_collection/plugins/module_utils/tower_api.py
index cf2860a87b..fdfede2553 100644
--- a/awx_collection/plugins/module_utils/tower_api.py
+++ b/awx_collection/plugins/module_utils/tower_api.py
@@ -25,6 +25,11 @@ class TowerAPIModule(TowerModule):
     }
     session = None
     cookie_jar = CookieJar()
+    IDENTITY_FIELDS = {
+        'users': 'username',
+        'workflow_job_template_nodes': 'identifier',
+        'instances': 'hostname'
+    }
 
     def __init__(self, argument_spec, direct_params=None, error_callback=None, warn_callback=None, **kwargs):
         kwargs['supports_check_mode'] = True
@@ -44,30 +49,28 @@ class TowerAPIModule(TowerModule):
 
     @staticmethod
     def get_name_field_from_endpoint(endpoint):
-        if endpoint == 'users':
-            return 'username'
-        elif endpoint == 'instances':
-            return 'hostname'
+        for candidate_endpoint, field_name in TowerAPIModule.IDENTITY_FIELDS.items():
+            if endpoint == candidate_endpoint:
+                return field_name
         return 'name'
 
     @staticmethod
     def get_item_name(item):
         if 'name' in item:
             return item['name']
-        elif 'username' in item:
-            return item['username']
-        elif 'identifier' in item:
-            return item['identifier']
-        elif 'hostname' in item:
-            return item['hostname']
-        elif item.get('type', None) == 'o_auth2_access_token':
+
+        for field_name in TowerAPIModule.IDENTITY_FIELDS.values():
+            if field_name in item:
+                return item[field_name]
+
+        if item.get('type', None) in ('o_auth2_access_token', 'credential_input_source'):
             # An oauth2 token has no name, instead we will use its id for any of the messages
-            rerurn item['id']
-        elif item.get('type', None) == 'credential_input_source':
             # An credential_input_source has no name, instead we will use its id for any of the messages
-            return existing_item['id']
-        return 'Unknown'
+            return item['id']
 
+        raise Exception(
+            'Cannot determine identity field for {} object.'.format(item.get('type', 'unknown'))
+        )
 
     def head_endpoint(self, endpoint, *args, **kwargs):
         return self.make_request('HEAD', endpoint, **kwargs)
@@ -141,18 +144,18 @@ class TowerAPIModule(TowerModule):
             self.fail_json(msg="The endpoint did not provide count and results")
 
         if response['json']['count'] == 0:
-            return None, name_or_id
+            return None
         elif response['json']['count'] > 1:
             if name_or_id:
                 # Since we did a name or ID search and got > 1 return something if the id matches
                 for asset in response['json']['results']:
                     if asset['id'] == name_or_id:
-                        return asset, asset['id'][name_field]
+                        return asset
             # We got > 1 and either didn't find something by ID (which means multiple names)
             # Or we weren't running with a or search and just got back too many to begin with.
             self.fail_json(msg="An unexpected number of items was returned from the API ({0})".format(response['json']['count']))
 
-        return response['json']['results'][0], response['json']['results'][0][name_field]
+        return response['json']['results'][0]
 
     def get_one_by_name_or_id(self, endpoint, name_or_id):
         name_field = self.get_name_field_from_endpoint(endpoint)
diff --git a/awx_collection/plugins/modules/tower_credential.py b/awx_collection/plugins/modules/tower_credential.py
index 146484ca66..f244ae812e 100644
--- a/awx_collection/plugins/modules/tower_credential.py
+++ b/awx_collection/plugins/modules/tower_credential.py
@@ -370,7 +370,7 @@ def main():
     if organization:
         lookup_data['organization'] = org_id
 
-    credential, name = module.get_one('credentials', name_or_id=name, **{'data': lookup_data})
+    credential = module.get_one('credentials', name_or_id=name, **{'data': lookup_data})
 
     if state == 'absent':
         # If the state was absent we can let the module delete it if needed, the module will handle exiting from this
@@ -396,7 +396,7 @@ def main():
 
     # Create the data that gets sent for create and update
     credential_fields = {
-        'name': new_name if new_name else name,
+        'name': new_name if new_name else (module.get_item_name(credential) if credential else name),
         'credential_type': cred_type_id,
     }
     if has_inputs:
```

I get that the determination of name is a little complicated. If you didn't want to do the in-line conditional, it actually makes more sense to me to omit the "name" field in `credential_fields`, unless either `new_name` is not `None` or `credential` is `None`. In other words, the name is only functional in the event of changing the name, or creating a new object.

john-westcott-iv:
@AlanCoding suggestion implemented. However, in get_item_name instead of raising an exception I made it non-static with a module.exit_json if we can't get an identifier. That way we don't need try/catch blocks in any module that uses the `'name': new_name if new_name else (module.get_item_name(credential) if credential else name)` pattern.

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -88,7 +113,21 @@ def get_all_endpoint(self, endpoint, *args, **kwargs):
             response['json']['next'] = next_page
         return response
 
-    def get_one(self, endpoint, *args, **kwargs):
+    def get_one(self, endpoint, name_or_id=None, *args, **kwargs):
+        if name_or_id:
+            name_field = self.get_name_field_from_endpoint(endpoint)
+            new_args = kwargs.get('data', {}).copy()
```
AlanCoding, line 123: This will result in queries like `/api/v2/inventories/?or__id=42&or__name=42&organization=1`

I'm still reading through the rest of this, but I want to be extra sure that this gets tested, because there is some fundamental ambiguity to this. Is this expressing `(name=42 OR id=42) AND organization=1`, or is it expressing `id=42 OR (name=42 AND organization=1)`? These are very much not the same thing, and failing test cases can be produced depending on your assumed expectations. I feel like it would have been easier to juggle the corner cases that arise from this if we kept the id check as a separate request.

AlanCoding: just to follow through what this does:

https://github.com/ansible/awx/blob/90c3bfc6ae92aa087f317e7515aeadef5cbfe8f5/awx/api/filters.py#L303-L305

(this code is not easy to read)

https://github.com/ansible/awx/blob/90c3bfc6ae92aa087f317e7515aeadef5cbfe8f5/awx/api/filters.py#L322-L323

https://github.com/ansible/awx/blob/90c3bfc6ae92aa087f317e7515aeadef5cbfe8f5/awx/api/filters.py#L344-L350

If that wasn't hard enough to read, let's look at where the real action happens:

https://github.com/ansible/awx/blob/90c3bfc6ae92aa087f317e7515aeadef5cbfe8f5/awx/api/filters.py#L369

This constructs a crazy complicated queryset in our cases that would (psuedo-code) look like:

```python
Inventory.objects.filter(
  organization=1,
  (Q(name=42) | Q(id=42))
)
```

awx_collection/plugins/modules/tower_job_template.py
```diff
@@ -453,9 +453,8 @@ def main():
         new_fields['inventory'] = module.resolve_name_to_id('inventories', inventory)
     if project is not None:
         if organization_id is not None:
-            project_data = module.get_one('projects', **{
+            project_data, project = module.get_one('projects', name_or_id=project, **{
```
AlanCoding: this doesn't look right here. It should return just one value now.

awx_collection/tools/roles/generate/templates/tower_module.j2
```diff
@@ -175,9 +175,8 @@ def main():
 {% endfor %}
 
     # Attempt to look up an existing item based on the provided data
-    existing_item = module.get_one('{{ item_type }}', **{
+    existing_item, name = module.get_one('{{ item_type }}', name_or_id={{ name_option }}, **{
```
AlanCoding: Again, I think you dropped the 2 returned items so this needs to be updated

AlanCoding (changes requested):
I've pointed out what I think are 2 bugs where it expects 2 values returned but only returns 1. Need a patch to those, and otherwise this looks ready to go.

awx_collection/plugins/modules/tower_job_template.py
john-westcott-iv: Thanks for catching this one.

awx_collection/tools/roles/generate/templates/tower_module.j2
john-westcott-iv: Updated and also fixed the call for name in the data structure down below for the new `new_name if new_name else (module... if existing_object else {{ name_field }}`

AlanCoding (approved):

awx_collection/tests/integration/targets/tower_inventory/tasks/main.yml
```diff
@@ -29,7 +29,7 @@
       tower_inventory:
         name: "{{ inv_name1 }}"
         organization: Default
-        insights_credential: "{{ cred_name1 }}"
+        insights_credential: "{{ results.id }}"
```
AlanCoding: > The task includes an option with an undefined variable. The error was: 'results' is undefined

results -> result?

awx_collection/tests/integration/targets/tower_credential_input_source/tasks/main.yml
```diff
@@ -27,17 +27,17 @@
...
-      - "result is changed"
+      - "target_cred_result is changed"
 
 - name: Add credential Input Source
   tower_credential_input_source:
     input_field_name: password
-    target_credential: "{{ target_cred_name }}"
-    source_credential: "{{ src_cred_name }}"
+    target_credential: "{{ target_cred_result.id }}"
+    source_credential: "{{ src_cred_result.id }}"
```
AlanCoding: right below this we seem to be getting and error:

```
  PLAY [testhost] ****************************************************************
  
  TASK [Gathering Facts] *********************************************************
  ok: [testhost]
  
  TASK [tower_credential_input_source : Generate names] **************************
  ok: [testhost]
  
  TASK [tower_credential_input_source : Add Tower credential Lookup] *************
  changed: [testhost]
  
  TASK [tower_credential_input_source : assert] **********************************
  ok: [testhost] => {
      "changed": false, 
      "msg": "All assertions passed"
  }
  
  TASK [tower_credential_input_source : Add Tower credential Target] *************
  changed: [testhost]
  
  TASK [tower_credential_input_source : assert] **********************************
  ok: [testhost] => {
      "changed": false, 
      "msg": "All assertions passed"
  }
  
  TASK [tower_credential_input_source : Add credential Input Source] *************
  changed: [testhost]
  
  TASK [tower_credential_input_source : assert] **********************************
  fatal: [testhost]: FAILED! => {"msg": "The conditional check 'result is changed' failed. The error was: The 'changed' test expects a dictionary"}
  
  PLAY RECAP *********************************************************************
  testhost                   : ok=7    changed=3    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0   
```

Maybe this task is missing a `register:`?

Maybe you missed a rename `result` -> `target_cred_result`?

awx_collection/tests/integration/targets/tower_inventory/tasks/main.yml
john-westcott-iv: Correct!

awx_collection/tests/integration/targets/tower_credential_input_source/tasks/main.yml
john-westcott-iv: It was missing a register. I also updated this test to be more friendly while I was in there.

beeankha (approved):

Issue #7655, opened 2020-07-20
Scheduler/Dispatcher does not start Jobs with >600 pending Jobs


moonrail:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
With a lot of pending Jobs (e.g. if Instances were disabled/down for maintenance, Cluster was restarted, Failover, ...) the Job-Scheduler/Dispatcher "run_task_manager" will not finish in under 5 Minutes and be killed by: https://github.com/ansible/awx/blob/devel/awx/main/dispatch/pool.py#L385
From our observation Jobs will only be executed, if Scheduler/Dispatcher ran through all pending/new Jobs and dispatched/skipped them.
Therefore no Jobs will be executed, and "run_callback_receiver" will never be run.

We've tried it with following setups:
- kubernetes with 15 instances, dedicated db
- just for testing: docker on linux (5 instances, with modified settings.py for unique UUIDs & Hostnames, dedicated db)

##### ENVIRONMENT
* AWX version: 13.0.0
* AWX install method: minishift, docker on linux
* Ansible version:  2.9.10
* Operating System: CentOS8
* Web Browser: Firefox, Chromium

##### STEPS TO REPRODUCE

First method:
- have a cluster with e.g. 10 instances
- spawn >600 Jobs (depends on latency to DB & instance resources)
- shut down DB or all instances
- start DB or all instances again

Second method:
- have a cluster with e.g. 10 instances
- set instances to disabled
- spawn >600 Jobs (depends on latency to DB & instance resources)
- enable instances again

##### EXPECTED RESULTS

Either:
- Jobs are being run
Or:
- Job is being run right after Scheduler/Dispatcher dispatched it, not after Scheduler/Dispatcher is done
  - this way at least some Jobs will be executed

##### ACTUAL RESULTS

- no jobs are executed, as Scheduler/Dispatcher never finishes processing queue before 5 minutes

##### ADDITIONAL INFORMATION

Example log output of the Scheduler/Dispatcher being killed:
```
awx_task | 2020-07-16T12:05:24.565256659Z 2020-07-16 12:05:24,564 DEBUG    awx.main.scheduler job 657142 (pending) couldn't be scheduled on graph, waiting for next cycle
awx_task | 2020-07-16T12:05:25.381709968Z 2020-07-16 12:05:25,381 DEBUG    awx.main.scheduler No instance available in group tower to run job job 657143 (pending) w/ capacity requirement 2
[...]

PR #8074, opened 2020-09-03
Prevent task manager timeout by limiting number of jobs to start

fosterseth:
##### SUMMARY

related #7655 

There are two expensive code loops in the task manager:
- Finding the preferred instance to run a job on (in `process_pending_tasks()`)
- Starting a task (db saves take a while)

If you queue up 500+ jobs (concurrent enabled), have a lot of instances and organizations, it's very possible that the task manager will time out (it reaps after running for 5 minutes). After a time out, it loses progress and the next cycle will start again from scratch. This creates a scenario where no jobs can be started.

The change here limits the number of jobs that the task manager can start to 150. Once this limit is reached, it can no longer turn pending jobs into running jobs. If the limit is reached, it will immediately schedule another task manager to run once the current one finishes, that way another set of 100 jobs can be started.

There is still a potential scenario where the task manager will time out. If we are under the limit (150), but we are at max capacity across available instances, we can still hit that heavy `find preferred instance` loop and might hit a timeout. Maybe this isn't so bad given we won't be able to start jobs anyways. Even if this happens we still will get jobs starting eventually.

To explain the above, say you have jobA and jobB. jobA can only run on instanceA, and jobB can only run on instanceB.

- Queue up 5k of jobA, followed by 5k of jobB. 

- Say instanceA is at max capacity, while instanceB has plenty of capacity left.

- The next task manager run will loop through jobA first, then jobB (order of job creation).

- The task manager spends all of its time in `find preferred instance` for jobA and eventually times out. It didn't even get to jobB, which technically could be started because instanceB has capacity left and we are not at the `start_task_limit`.

I think the above scenario is unlikely for most situations, and even in that case we don't hit a deadlock and jobs will eventually complete, albeit much slower than it should.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 14.1.0
```

### DIFF
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -12,6 +12,7 @@ import random
 from django.db import transaction, connection
 from django.utils.translation import ugettext_lazy as _, gettext_noop
 from django.utils.timezone import now as tz_now
+from django.conf import settings
 
 # AWX
 from awx.main.dispatch.reaper import reap_job
@@ -45,6 +46,12 @@ class TaskManager():
 
     def __init__(self):
         self.graph = dict()
+        # start task limit indicates how many pending jobs can be started on this
+        # .schedule() run. Starting jobs is expensive, and there is code in place to reap
+        # the task manager after 5 minutes. At scale, the task manager can easily take more than
+        # 5 minutes to start pending jobs. If this limit is reached, pending jobs
+        # will no longer be started and will be started on the next task manager cycle.
+        self.start_task_limit = settings.START_TASK_LIMIT
         for rampart_group in InstanceGroup.objects.prefetch_related('instances'):
             self.graph[rampart_group.name] = dict(graph=DependencyGraph(rampart_group.name),
                                                   capacity_total=rampart_group.capacity,
@@ -189,6 +196,10 @@ class TaskManager():
         return result
 
     def start_task(self, task, rampart_group, dependent_tasks=None, instance=None):
+        self.start_task_limit -= 1
+        if self.start_task_limit == 0:
+            # schedule another run immediately after this task manager
+            schedule_task_manager()
         from awx.main.tasks import handle_work_error, handle_work_success
 
         dependent_tasks = dependent_tasks or []
@@ -448,6 +459,8 @@ class TaskManager():
     def process_pending_tasks(self, pending_tasks):
         running_workflow_templates = set([wf.unified_job_template_id for wf in self.get_running_workflow_jobs()])
         for task in pending_tasks:
+            if self.start_task_limit <= 0:
+                break
             if self.is_job_blocked(task):
                 logger.debug("{} is blocked from running".format(task.log_format))
                 continue
--- a/awx/settings/defaults.py
+++ b/awx/settings/defaults.py
@@ -203,6 +203,9 @@ JOB_EVENT_MAX_QUEUE_SIZE = 10000
 # The number of job events to migrate per-transaction when moving from int -> bigint
 JOB_EVENT_MIGRATION_CHUNK_SIZE = 1000000
 
+# The maximum allowed jobs to start on a given task manager cycle
+START_TASK_LIMIT = 100
+
 # Disallow sending session cookies over insecure connections
 SESSION_COOKIE_SECURE = True
 


ryanpetrello (approved):
Thanks for the detailed explanation @fosterseth

This seems pretty reasonable to me.  I'd love for @rebeccahhh and @chrismeyersfsu - who have more task manager experience than I do - to weigh in as well.

ryanpetrello:
@moonrail @psuriset either of you have any interest in giving this patch a whirl (see: https://github.com/ansible/awx/issues/7655)

awx/main/scheduler/task_manager.py
```diff
@@ -45,6 +45,12 @@ class TaskManager():
 
     def __init__(self):
         self.graph = dict()
+        # start task limit indicates how many pending jobs can be started on this
+        # .schedule() run. Starting jobs is expensive, and there is code in place to reap
+        # the task manager after 5 minutes. At scale, the task manager can easily take more than
+        # 5 minutes to start pending jobs. If this limit is reached, pending jobs
+        # will no longer be started and will be started on the next task manager cycle.
+        self.start_task_limit = 150
```
ryanpetrello: Curious, why 150, and not - say - 100?

psuriset:
@ryanpetrello We will trigger few perf tests

awx/main/scheduler/task_manager.py
fosterseth: No particular reason -- the original issue involved around 300 jobs locking up the task manager, so I figured half of that would be a safe number.

AlanCoding: I could see a benefit to keeping the definition of this constant in settings.

awx/main/scheduler/task_manager.py
```diff
@@ -451,6 +461,8 @@ def process_pending_tasks(self, pending_tasks):
             if self.is_job_blocked(task):
                 logger.debug("{} is blocked from running".format(task.log_format))
                 continue
+            if self.start_task_limit <= 0:
+                continue
```
AlanCoding: wouldn't it be better to put this before `self.is_job_blocked`?

Or even, if you hit the condition, `break` instead of `continue`

awx/main/scheduler/task_manager.py
fosterseth: i'll look into adding this as a setting

awx/main/scheduler/task_manager.py
fosterseth: yes you're right, can `break` and be before `is_job_blocked`

AlanCoding (approved):
I only left minor non-blocking comments, after your testing is complete, I'd say go ahead with merge

chrismeyersfsu:
This is a great small patch!

It's worth mentioning the root cause here as well as possible solutions.

### Root Cause
`preferred_instance_groups` gets called for each job that is considered via `process_pending_tasks()` that @fosterseth mentioned. This flow results in MANY individual queries.

### Possible Solutions
1. leverage `prefetch_related`
2. `instance_groups` is a field on `UnifiedJobTemplate`, this is good. We could query all `instance_groups` for all unique job ids that we care about and use the results as a lookup table for the task manager cycle.

Below is a summary of the logic implemented in each class definition of `preferred_instance_groups`. After looking at the logic in each implementation we know that solution `2.` is possible.

<p align="center"><b>preferred_instance_groups</b></p>

| Model                                                        | Description                                                  |
| ------------------------------------------------------------ | ------------------------------------------------------------ |
| <a href="https://github.com/ansible/awx/blob/devel/awx/main/models/jobs.py#L778">Job</a> | job_template.instance_groups + inventory.instance_groups + organization.instance_groups |
| <a href="https://github.com/ansible/awx/blob/devel/awx/main/models/workflow.py#L667">WorkflowJob</a> | []                                                           |
| <a href="https://github.com/ansible/awx/blob/devel/awx/main/models/ad_hoc_commands.py#L215">Ad Hoc</a> | inventory.instance_groups + inventory.organization.instance_groups |
| <a href="https://github.com/ansible/awx/blob/devel/awx/main/models/projects.py#L610">ProjectUpdate</a> | super.preferred_instance_groups + organization.instance_groups     |
| <a href="https://github.com/ansible/awx/blob/devel/awx/main/models/unified_jobs.py#L1410">UnifiedJob</a> | unified_job_template.instance_groups |
| <a href="https://github.com/ansible/awx/blob/devel/awx/main/models/inventory.py#L1302">InventoryUpdate</a> | inventory_source.inventory.instance_groups + inventory_source.inventory.organization.instance_groups |

fosterseth:
@chrismeyersfsu your second solution sounds like a straightforward and really nice improvement. I think I'll implement that in a separate PR. Thanks for laying out the details.

PR #8075, opened 2020-09-03
if redis is unreachable, set instance capacity to zero

ryanpetrello:


AlanCoding:
So this is called once per heartbeat?

https://github.com/ansible/awx/blob/967e35fec911f5d707d923a3be3b38dcbdac7e35/awx/main/tasks.py#L418

Can we get into a situation where it takes 60 seconds to timeout or something?

AlanCoding (approved):
it's probably safe

ryanpetrello:
> Can we get into a situation where it takes 60 seconds to timeout or something?

Given that it's a local domain socket, I doubt it.  If so, you've got bigger problems.  I think this is relatively safe.

Issue #8073, opened 2020-09-03
Add details page for container groups


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Add details page for container groups

nixocio:
@unlikelyzero, can we close this one?

unlikelyzero:
@nixocio we will not be in the critical path for EE but will still be in the critical path for Container Groups. In this case, Danny didn't link to the tests which covered this functionality nor the PR which was in progress. We will leave this post-merge needs_test issue up to ensure that we had and have adequate coverage.

For now, our priority will remain the open awx pulls and get back to these post-merge needs_test issues

nixocio:
Thanks @unlikelyzero.

tiagodread:
Covered by:
- container-group-crud.spec.js
- container-group-operations.spec.js
- container-group-tabs.spec.js

Closing this issue.

PR #8078, opened 2020-09-03
Add Container Group details

nixocio:
Add Container Group details.

See: https://github.com/ansible/awx/issues/8073

* Details Page with `pod_spec_override`  modified.

image

* Details Page with `pod_spec_override`  not modified.

image

### DIFF
--- a/awx/ui_next/src/screens/InstanceGroup/ContainerGroupDetails/ContainerGroupDetails.jsx
+++ b/awx/ui_next/src/screens/InstanceGroup/ContainerGroupDetails/ContainerGroupDetails.jsx
@@ -1,14 +1,118 @@
-import React from 'react';
-import { Card, PageSection } from '@patternfly/react-core';
+import React, { useCallback } from 'react';
+import { withI18n } from '@lingui/react';
+import { t } from '@lingui/macro';
+import { Link, useHistory } from 'react-router-dom';
+import { Button, Label } from '@patternfly/react-core';
 
-function ContainerGroupDetails() {
+import { VariablesDetail } from '../../../components/CodeMirrorInput';
+import AlertModal from '../../../components/AlertModal';
+import { CardBody, CardActionsRow } from '../../../components/Card';
+import DeleteButton from '../../../components/DeleteButton';
+import {
+  Detail,
+  DetailList,
+  UserDateDetail,
+} from '../../../components/DetailList';
+import useRequest, { useDismissableError } from '../../../util/useRequest';
+import { jsonToYaml, isJsonString } from '../../../util/yaml';
+import { InstanceGroupsAPI } from '../../../api';
+
+function ContainerGroupDetails({ instanceGroup, i18n }) {
+  const { id, name } = instanceGroup;
+
+  const history = useHistory();
+
+  const {
+    request: deleteInstanceGroup,
+    isLoading,
+    error: deleteError,
+  } = useRequest(
+    useCallback(async () => {
+      await InstanceGroupsAPI.destroy(id);
+      history.push(`/instance_groups`);
+    }, [id, history])
+  );
+
+  const { error, dismissError } = useDismissableError(deleteError);
+
   return (
-    <PageSection>
-      <Card>
-        <div>Container group details</div>
-      </Card>
-    </PageSection>
+    <CardBody>
+      <DetailList>
+        <Detail
+          label={i18n._(t`Name`)}
+          value={instanceGroup.name}
+          dataCy="container-group-detail-name"
+        />
+        <Detail
+          label={i18n._(t`Type`)}
+          value={i18n._(t`Container group`)}
+          dataCy="container-group-type"
+        />
+        <Detail
+          label={i18n._(t`Credential`)}
+          value={
+            <Label variant="outline" color="blue">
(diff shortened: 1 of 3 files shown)


awx/ui_next/src/screens/InstanceGroup/ContainerGroupDetails/ContainerGroupDetails.jsx
```diff
@@ -1,14 +1,137 @@
...
+        />
+        <UserDateDetail
+          label={i18n._(t`Last Modified`)}
+          date={instanceGroup.modified}
+          user={instanceGroup.summary_fields.modified_by}
+        />
+        <VariablesDetail
+          label={i18n._(t`Pod spec override`)}
+          value={
+            isJsonString(instanceGroup.pod_spec_override)
```
jakemcdermott: @nixocio To avoid showing a blank configuration, we should use the default value when a `pod_spec_override` hasn't been provided.

- First we make an options request to `api/v2/instance_groups`. 
  - You can include that with https://github.com/ansible/awx/blob/devel/awx/ui_next/src/screens/InstanceGroup/InstanceGroup.jsx#L36 using `Promise.all` if you'd like.
- when `instanceGroup.pod_spec_override` isn't defined, use the value of `actions.PUT.pod_spec_override.default` from the options data instead

jakemcdermott (changes requested):

awx/ui_next/src/screens/InstanceGroup/ContainerGroupDetails/ContainerGroupDetails.jsx
```diff
@@ -1,14 +1,137 @@
...
+            label={i18n._(t`Used capacity`)}
+            content={`${Math.round(
+              100 - instanceGroup.percent_capacity_remaining
+            )} %`}
+            dataCy="container-group-used-capacity"
+          />
+        ) : (
+          <Detail
+            label={i18n._(t`Used capacity`)}
+            value={<Unavailable>{i18n._(t`Unavailable`)}</Unavailable>}
```
jakemcdermott: My understanding is that container groups don't have capacity values like instance groups do. Container group instances are ephemeral and so container groups don't really have dedicated instances in the traditional sense.

[We don't show any capacity values on the list item](https://user-images.githubusercontent.com/9753817/92267485-bb297d00-eeae-11ea-840f-f3806e134621.png) and [it doesn't it seem like you intend to make these fields editable](https://github.com/ansible/awx/pull/8070/files#diff-b8b640dae39bb9d0c296b191b0378519R30) on the form, either, so I wanted to confirm that these fields should be included on this component.

awx/ui_next/src/screens/InstanceGroup/ContainerGroupDetails/ContainerGroupDetails.jsx
nixocio: Thanks for your review @jakemcdermott, I did think about that...but we currently do not do that in the `classic` ui. I am +1 to add that, but I verified the api itself, and it is an empty string when no `pod_spec_override` was provided.  I can update that for sure.

awx/ui_next/src/screens/InstanceGroup/ContainerGroupDetails/ContainerGroupDetails.jsx
nixocio: I will verify this one. Thanks.

awx/ui_next/src/screens/InstanceGroup/ContainerGroupDetails/ContainerGroupDetails.jsx
jakemcdermott: In the old ui I think we hide the field when it isn't overridden.

We can hide the field, too, if you'd like. The important thing is that we don't present the configuration as blank when an override isn't being used because that is potentially misleading.

nixocio: Updated. Please, see images above.

awx/ui_next/src/screens/InstanceGroup/ContainerGroupDetails/ContainerGroupDetails.jsx
nixocio: Updated. Please, see images above.

jakemcdermott (approved):

dsesami:
see https://github.com/ansible/awx/pull/8070#issuecomment-690260512 -- this will be followed up on soon

dsesami:
Found one bug. 

Pod spec override does not do JSON/YAML validation when saving.

1. Take a container group
2. do something to the JSON or the YAML, does not matter. remove a quote, break the syntax somehow
3. save (this should fail but it succeeds)

dsesami (changes requested):
once the validation goes in on save as described above, then this is good to go :+1:

nixocio:
@dsesami, this bug that you reported is slightly more involving. Not being introduced by the code in this PR. 

I filed 2 issues related to this.  
See: https://github.com/ansible/awx/issues/8181 
Also: https://github.com/ansible/awx/issues/8182

dsesami (approved):

PR #8084, opened 2020-09-04
remove old fact scanning plugins

ryanpetrello:
@AlanCoding @matburt @wenottingham convince me that we still rely on this/that it's still in use


AlanCoding (approved):
It was actually kind of hazardous that we were setting ANSIBLE_LIBRARY in the first place, because that can blow away the user's settings

ryanpetrello:
Okay, I ran through some test and feel confident that we don't need this anymore.

PR #8087, opened 2020-09-04
Add new option update_secrets to allow lazy or strict updating

AlanCoding:
This is my attempt to modify the solution for `tower_user` updating passwords, so that it can be rolled out generally to other modules, and so that it follows a universal API principal.

### DIFF
--- a/awx_collection/plugins/module_utils/tower_api.py
+++ b/awx_collection/plugins/module_utils/tower_api.py
@@ -28,6 +28,7 @@ class TowerAPIModule(TowerModule):
         'workflow_job_template_nodes': 'identifier',
         'instances': 'hostname'
     }
+    ENCRYPTED_STRING = "$encrypted$"
 
     def __init__(self, argument_spec, direct_params=None, error_callback=None, warn_callback=None, **kwargs):
         kwargs['supports_check_mode'] = True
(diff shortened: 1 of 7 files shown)


awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -24,7 +24,7 @@ class TowerAPIModule(TowerModule):
         'tower': 'Red Hat Ansible Tower',
     }
     session = None
-    cookie_jar = CookieJar()
```
AlanCoding, line 25: @john-westcott-iv this particular thing wasn't getting used.

awx_collection/plugins/modules/tower_user.py
```diff
@@ -55,6 +55,12 @@
       description:
         - Write-only field used to change the password.
       type: str
+    update_secrets:
+      description:
+        - C(true) will always password if user specifies password and API gives $encrypted$ for password.
```
beeankha: Sentence here is incomplete, looks like it should be "... will always **change** password if user specifies..."

beeankha (approved):
Other than minor comment re: parameter description text, this looks good to me!

AlanCoding:
regate

AlanCoding:
regate

AlanCoding:
regate

AlanCoding:
regate

AlanCoding:
regate

PR #8094, opened 2020-09-08
Update Django and channels_redis

ryanpetrello:
see: https://www.djangoproject.com/weblog/2020/sep/01/security-releases/

This isn't _that_ critical given that we run Python 3.6 (not 3.7+).

but also related: https://github.com/django/channels_redis/issues/212

### DIFF
--- a/awx/main/consumers.py
+++ b/awx/main/consumers.py
@@ -1,5 +1,3 @@
-import collections
-import functools
 import json
 import logging
 import time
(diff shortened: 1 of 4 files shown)


awx/main/consumers.py
```diff
@@ -14,40 +12,12 @@
...
 from channels.layers import get_channel_layer
 from channels.db import database_sync_to_async
-from channels_redis.core import RedisChannelLayer
 
 
 logger = logging.getLogger('awx.main.consumers')
 XRF_KEY = '_auth_user_xrf'
 
 
-class BoundedQueue(asyncio.Queue):
```
ryanpetrello, line 24: replaced by: https://github.com/django/channels_redis/pull/219

beeankha (approved):

jbradberry (approved):

Issue #4254, opened 2019-07-02
Auto populate lookup fields in add forms when user only has access to one resource


mabashian:
##### SUMMARY
On certain add forms in the UI lookup fields like Inventory/Project/Organization (that are required) are automatically populated when the user only has access to one of these resources.  Since the field is required and the user only has access to one we can safely assume that that's the resource they'll select.

In order to achieve this, we'll need to identify all of the fields that should behave this way/do behave this way in `awx/ui`.  This will only impact add forms.  Edit forms should not behave this way.  If we detect that the user has access to more than one of these resources the field should be blank.

mabashian:
Link https://github.com/ansible/awx/issues/4206 for original issue.  This is a subset of those proposed changes.

mabashian:
Adding machine credential in the adhoc command wizard to this issue

mabashian:
OK here's my best effort on all of the places where this should happen:

1) User Form
   - Organization Lookup
2) Application Form
   -  Organization Lookup
3) Inventory Form
   - Organization Lookup
4) Smart Inventory Form
   - Organization Lookup
5) Project Form 
   - Organization Lookup
   - Red Hat Insights -> Credential Lookup
6) Team Form 
   - Organization Lookup
7) Job Template Form 
   - Project Lookup (this was already in place but I normalized the implementation)
8) Inventory Source Form 
   - Sourced from project -> Project Lookup (this was already in place but I normalized the implementation)
   - Google Compute Engine -> Credential Lookup
   - Microsoft Azure Resource Manager -> Credential Lookup
   - VMware vCenter -> Credential Lookup
   - Red Hat Satellite 6 -> Credential Lookup
   - OpenStack -> Credential Lookup
   - Red Hat Virtualization-> Credential Lookup
   - Ansible Tower -> Credential Lookup

------ These are not included in #8104 and will have to be added in other PR(s) ------

9) Notification Template Form 
   - Organization Lookup
10) Container Group Form 
    - Credential Lookup
11) Adhoc Command Form 
    - Machine Credential Lookup

tiagodread:
Tests and PR already merged 👍🏽

PR #8104, opened 2020-09-08
Auto populate various required lookups on various forms

mabashian:
##### SUMMARY
link #4254 

Auto populates the following fields on the following forms if the user only has access to one resource for the field in question:

1) User Form
   - Organization Lookup
2) Application Form
   -  Organization Lookup
3) Inventory Form
   - Organization Lookup
4) Smart Inventory Form
   - Organization Lookup
5) Project Form 
   - Organization Lookup
   - Red Hat Insights -> Credential Lookup
6) Team Form 
   - Organization Lookup
7) Job Template Form 
   - Project Lookup (this was already in place but I normalized the implementation)
8) Inventory Source Form 
   - Sourced from project -> Project Lookup (this was already in place but I normalized the implementation)
   - Google Compute Engine -> Credential Lookup
   - Microsoft Azure Resource Manager -> Credential Lookup
   - VMware vCenter -> Credential Lookup
   - Red Hat Satellite 6 -> Credential Lookup
   - OpenStack -> Credential Lookup
   - Red Hat Virtualization-> Credential Lookup
   - Ansible Tower -> Credential Lookup

In order to achieve this and make it as uniform as possible I added the useAutoPopulateLookup hook.  This simple hook will trigger the callback to update the field value the first time through if enabled.  This hook was added to the CredentialLookup, OrganizationLookup, and ProjectLookup components.

As a result of these changes, a lot of `onChange` functions in various forms needed to be wrapped in `useCallback` to prevent unnecessary extra requests being triggered in the lookup.  The `onChange` function is now referenced inside the useRequest hook and as a result is in it's dependency array and shouldn't be changing with each render (hence useCallback).  There was one snag with this pattern: https://github.com/formium/formik/issues/2268.  The `useField` helpers in Formik were causing the callback function to be re-generated with each render (and as result were causing numerous extra GET requests by the lookup components).  Fortunately, `setFieldValue` does not suffer the same fate: https://github.com/formium/formik/issues/2268#issuecomment-682685788.  So I refactored a lot of our functions to use that inside the `useCallback` hooks.

Here's a quick example of what it looks like:

autopoporg

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/util/useAutoPopulateLookup.jsx
+++ b/awx/ui_next/src/util/useAutoPopulateLookup.jsx
@@ -1,0 +1,24 @@
+import { useCallback, useRef } from 'react';
+
+/**
+ * useAutoPopulateLookup hook [... insert description]
+ * Param: [... insert params]
+ * Returns: {
+ *  [... insert returns]
+ * }
+ */
+
+export default function useAutoPopulateLookup(populateLookupField) {
+  const isFirst = useRef(true);
+
+  return useCallback(
+    results => {
+      if (isFirst.current && results.length === 1) {
+        populateLookupField(results[0]);
+      }
+
+      isFirst.current = false;
+    },
+    [populateLookupField]
+  );
+}
--- a/awx/ui_next/src/components/Lookup/CredentialLookup.jsx
+++ b/awx/ui_next/src/components/Lookup/CredentialLookup.jsx
@@ -10,6 +10,7 @@ import { getQSConfig, parseQueryString, mergeParams } from '../../util/qs';
 import { FieldTooltip } from '../FormField';
 import Lookup from './Lookup';
 import OptionsList from '../OptionsList';
+import useAutoPopulateLookup from '../../util/useAutoPopulateLookup';
 import useRequest from '../../util/useRequest';
 import LookupErrorMessage from './shared/LookupErrorMessage';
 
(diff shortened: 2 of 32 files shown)


unlikelyzero:
@dsesami @tiagodread , @mabashian is going on PTO this week. I think we should give this some early feedback today.

awx/ui_next/src/util/useAutoPopulateLookup.jsx
```diff
@@ -0,0 +1,24 @@
+import { useCallback, useRef } from 'react';
+
+/**
```
jlmitch5, line 3: I think we should either fill this in or get rid of the comment

awx/ui_next/src/util/useAutoPopulateLookup.jsx
```diff
@@ -0,0 +1,24 @@
...
+    results => {
+      if (isFirst.current && results.length === 1) {
+        populateLookupField(results[0]);
+      }
+
+      isFirst.current = false;
+    },
+    [populateLookupField]
+  );
+}
```
jlmitch5, line 24: this makes sense to me.

jlmitch5 (approved):
looks good, aside from unused doc comment

tiagodread:
@mabashian As a normal user who has admin access to an Organization and Projects I can’t choose a project on Job template form because the lookup field is disabled.

Steps to reproduce:
1. Create a organization,  2 projects
2. Create a normal user and give him access as admin to this organization and projects
3. Login as this user
4. Navigate to Job template then click on Add button

**Result:**
This user can’t choose a project because de field is disabled

image

**Expected Result:**
in the old UI this user can choose a project and create a Job template

image


Extra information: Instead of creating 2 projects, create only one, the project field is auto populated but still disabled

tiagodread (approved):
Tests and stubs covered on 5530 👍🏽

Issue #7877, opened 2020-08-11
Notification Details: Display custom notification messages


keithjgrant:
Follow-up on #7867

Display the custom notification text, if different than the default messages.

##### ISSUE TYPE
 - Feature Idea

unlikelyzero:
@keithjgrant , @tiagodread and @unlikelyzero will be testing this when it's first available for `state:needs_review`. We've got most of the tests already written

tiagodread:
Everything OK, tests are written and the PR already merged

PR #8105, opened 2020-09-08
Notification Detail: show custom messages

keithjgrant:
##### SUMMARY
Adds custom messages to the Notification Detail, if they have been customized. I also created an ArrayDetail component to clean up the display of multi-value details (such as Slack destination channels, Twilio destination numbers, etc)

Addresses #7877 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/screens/NotificationTemplate/NotificationTemplateDetail/NotificationTemplateDetail.jsx
+++ b/awx/ui_next/src/screens/NotificationTemplate/NotificationTemplateDetail/NotificationTemplateDetail.jsx
@@ -435,7 +435,7 @@ function CustomMessageDetails({ messages, defaults, type, i18n }) {
         <CodeDetail
           label={i18n._(t`Workflow approved message`)}
           value={
-            messages.workflow_approval.approved.message ||
+            messages.workflow_approval?.approved?.message ||
             defaults.workflow_approval.approved.message
           }
           mode="jinja2"
@@ -447,7 +447,7 @@ function CustomMessageDetails({ messages, defaults, type, i18n }) {
         <CodeDetail
           label={i18n._(t`Workflow approved message body`)}
           value={
-            messages.workflow_approval.approved.body ||
+            messages.workflow_approval?.approved?.body ||
             defaults.workflow_approval.approved.body
           }
           mode="jinja2"
@@ -459,7 +459,7 @@ function CustomMessageDetails({ messages, defaults, type, i18n }) {
         <CodeDetail
           label={i18n._(t`Workflow denied message`)}
           value={
-            messages.workflow_approval.denied.message ||
+            messages.workflow_approval?.denied?.message ||
             defaults.workflow_approval.denied.message
           }
           mode="jinja2"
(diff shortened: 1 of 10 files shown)


unlikelyzero:
@tiagodread I think you can start adding tests for this one even in it's current state

tiagodread:
@keithjgrant tests are written and once the PR receives dev review, I'll run the tests again and it can be merged.

awx/ui_next/src/screens/NotificationTemplate/NotificationTemplateDetail/NotificationTemplateDetail.jsx
```diff
@@ -358,4 +371,164 @@ function NotificationTemplateDetail({ i18n, template }) {
...
+          mode="jinja2"
+          rows="6"
+          fullWidth
+        />
+      )}
+      {showMessages && (
+        <CodeDetail
+          label={i18n._(t`Workflow approved message`)}
+          value={
+            messages.workflow_approval.approved.message ||
```
jakemcdermott:
```suggestion
            messages?.workflow_approval?.approved?.message ||
```

@keithjgrant 
`messages.workflow_approval` can be `null` so I think we need to check before referencing its properties or use optional chaining in the various places it is referenced:

Screenshot from 2020-09-14 13-17-15

Screenshot from 2020-09-14 13-18-06

awx/ui_next/src/screens/NotificationTemplate/shared/hasCustomMessages.js
```diff
@@ -0,0 +1,40 @@
...
+  if (!messages) {
+    return false;
+  }
+
+  return (
+    isCustomized(messages.started, defaults.started) ||
+    isCustomized(messages.success, defaults.success) ||
+    isCustomized(messages.error, defaults.error) ||
+    isCustomized(
+      messages.workflow_approval.approved,
```
jakemcdermott:
```suggestion
      messages?.workflow_approval?.approved,
```

jakemcdermott (changes requested):

awx/ui_next/src/screens/NotificationTemplate/NotificationTemplateDetail/NotificationTemplateDetail.jsx
jakemcdermott: probably needed here as well: https://github.com/ansible/awx/blob/devel/awx/ui_next/src/screens/NotificationTemplate/shared/NotificationTemplateForm.jsx#L144-L170

awx/ui_next/src/screens/NotificationTemplate/shared/hasCustomMessages.js
jakemcdermott: probably needed here as well: https://github.com/ansible/awx/blob/devel/awx/ui_next/src/screens/NotificationTemplate/shared/NotificationTemplateForm.jsx#L144-L170

jakemcdermott (approved):

jakemcdermott:
After looking a bit further into the `null` `workflow_approval` field w/ @keithjgrant I'm starting to think this might be an unaddressed data migration issue with pre-existing notification templates and the (more recently added) workflow approval feature. It's probably simpler to just handle the `null` in the ui at this point so that's what we're doing.

tiagodread:
@keithjgrant After clicking on the select all notifications on notification templates list isn’t possible to unselect items:

chrome-capture (5)

tiagodread (approved):
Tests are written 👍🏽

PR #8109, opened 2020-09-09
Hack to delete orphaned organizations, consolidate get_one methods

AlanCoding:
##### SUMMARY
This addresses some test fails we get when running the `demo_data` target.

This happens because of the series of events:

 - the `awx-manage create_preload_data` command was ran
 - the "Default" organization gets deleted, for some reason or another
 - the collection integration tests are ran, running demo_data first

When trying to create the job template, it fails because the "Demo Project" has 2 items that match the query.

This is because before getting to the JT creation, it creates the "Demo Project" in the "Default" organization. But 1 "Demo Project" already exists with a null organization.

This is pretty ugly, I don't like it, I don't think anyone should like it. But right now, for stability, we should have at least _some_ means of cleaning up orphaned objects so that playbooks written using the `awx.awx` collection can self-correct, and if this becomes a formula for that, it's not the worst thing that could happen.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
14.1.0
```

##### ADDITIONAL INFORMATION
I kind of have a draft for an improved solution, I started down this path but haven't been spending as much time directly writing stuff in the collection lately:

 - Instead of combining `name_or_id` with the query parameters, _first_ search exclusively with `{"name": "<name>"}`
 - _Then_, if that comes back with 0 results, try request with `{"id": <id>}` if the id is integer like
 - If the id is not integer-like, or both prior requests came back with >1, then 0 results, then combine name with lookup data

This would make 1 request into potentially 3 requests, but I don't think it would do this _very often_ because in 99% of cases the first request will find the object. Even if that's 90% instead... it's not a performance hit big enough to care about.

Anyway, I do not do that here, because I expect it to be more controversial.

### DIFF
--- a/awx_collection/plugins/module_utils/tower_api.py
+++ b/awx_collection/plugins/module_utils/tower_api.py
@@ -123,12 +123,12 @@ class TowerAPIModule(TowerModule):
             if name_field in new_data:
                 self.fail_json(msg="You can't specify the field {0} in your search data if using the name_or_id field".format(name_field))
 
-            new_data['or__{0}'.format(name_field)] = name_or_id
             try:
                 new_data['or__id'] = int(name_or_id)
+                new_data['or__{0}'.format(name_field)] = name_or_id
             except ValueError:
                 # If we get a value error, then we didn't have an integer so we can just pass and fall down to the fail
-                pass
+                new_data[name_field] = name_or_id
             new_kwargs['data'] = new_data
 
         response = self.get_endpoint(endpoint, **new_kwargs)
@@ -160,11 +160,13 @@ class TowerAPIModule(TowerModule):
 
     def fail_wanted_one(self, response, endpoint, query_params):
         sample = response.copy()
-        if len(sample['json']['results']) < 1:
+        if len(sample['json']['results']) > 1:
             sample['json']['results'] = sample['json']['results'][:2] + ['...more results snipped...']
+        url = self.build_url(endpoint, query_params)
+        display_endpoint = url.geturl()[len(self.host):]  # truncate to not include the base URL
         self.fail_json(
             msg="Request to {0} returned {1} items, expected 1".format(
-                self.build_url(endpoint, query_params).geturl(), response['json']['count']
+                display_endpoint, response['json']['count']
             ),
             query=query_params,
             response=sample,
--- a/awx_collection/plugins/module_utils/tower_api.py
+++ b/awx_collection/plugins/module_utils/tower_api.py
@@ -4,11 +4,9 @@ __metaclass__ = type
 from . tower_module import TowerModule
 from ansible.module_utils.urls import Request, SSLValidationError, ConnectionError
 from ansible.module_utils.six import PY2
-from ansible.module_utils.six.moves.urllib.parse import urlencode
 from ansible.module_utils.six.moves.urllib.error import HTTPError
 from ansible.module_utils.six.moves.http_cookiejar import CookieJar
 import time
-import re
 from json import loads, dumps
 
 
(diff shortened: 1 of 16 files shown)


awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -4,11 +4,9 @@
 from . tower_module import TowerModule
 from ansible.module_utils.urls import Request, SSLValidationError, ConnectionError
 from ansible.module_utils.six import PY2
-from ansible.module_utils.six.moves.urllib.parse import urlencode
 from ansible.module_utils.six.moves.urllib.error import HTTPError
 from ansible.module_utils.six.moves.http_cookiejar import CookieJar
 import time
-import re
```
AlanCoding, line 11: @john-westcott-iv it looks like these are `flake8` failures that, for whatever reason, were not getting picked up by the test targets we're running. The unused imports are largely leftover from splitting into 2 different classes.

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -143,63 +142,42 @@ def get_one(self, endpoint, name_or_id=None, *args, **kwargs):
...
+                    if str(asset['id']) == name_or_id:
                         return asset
             # We got > 1 and either didn't find something by ID (which means multiple names)
             # Or we weren't running with a or search and just got back too many to begin with.
-            self.fail_json(msg="An unexpected number of items was returned from the API ({0})".format(response['json']['count']))
+            self.fail_wanted_one(response, endpoint, new_kwargs.get('data'))
 
         return response['json']['results'][0]
 
-    def get_one_by_name_or_id(self, endpoint, name_or_id):
```
AlanCoding, line 159: This does almost the exact same thing as `get_one`, and it was relatively trivial to consolidate them.

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -212,46 +190,41 @@ def make_request(self, method, endpoint, *args, **kwargs):
             # If we have a oauth token, we just use a bearer header
             headers['Authorization'] = 'Bearer {0}'.format(self.oauth_token)
 
-        # Update the URL path with the endpoint
-        self.url = self.url._replace(path=endpoint)
```
AlanCoding, line 216: @john-westcott-iv the variable `self.url` is defined in the parent class `TowerModule`. I'm fine to have the subclasses reference it (like other config params), but I don't want them to _modify_ it. That makes state hard to track. By only working with a local `url` instead of `self.url` we get rid of the `finally:` block to reset it to the prior value. It's easier to just never modify instance data so we don't have to worry about state.

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -627,7 +598,7 @@ def wait_on_url(self, url, object_name, object_type, timeout=30, interval=10):
             # If we are past our time out fail with a message
             if timeout and timeout < time.time() - start:
                 # Account for Legacy messages
-                if object_type is 'legacy_job_wait':
+                if object_type == 'legacy_job_wait':
```
AlanCoding, line 611: again, flake8 failures that were not discovered before

awx_collection/plugins/module_utils/tower_module.py
```diff
@@ -112,6 +112,23 @@ def __init__(self, argument_spec=None, direct_params=None, error_callback=None,
...
+        if not endpoint.endswith('/') and '?' not in endpoint:
+            endpoint = "{0}/".format(endpoint)
+
+        # Update the URL path with the endpoint
+        url = self.url._replace(path=endpoint)
+
+        if query_params:
+            url = url._replace(query=urlencode(query_params))
+
+        return url
```
AlanCoding, line 130: Here again, trying to make `TowerModule` class the owner of any gymnastics that has to go on for `self.url`

awx_collection/plugins/modules/tower_project.py
```diff
@@ -234,21 +233,22 @@ def main():
...
-        }
-    })
+    project = module.get_one('projects', name_or_id=name, data=lookup_data)
 
     if state == 'absent':
         # If the state was absent we can let the module delete it if needed, the module will handle exiting from this
         module.delete_if_needed(project)
 
+    if credential is not None:
+        credential = module.resolve_name_to_id('credentials', credential)
```
AlanCoding, line 250: I kind of just made the change suggested in https://github.com/ansible/awx/pull/7925

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -117,22 +115,23 @@ def get_all_endpoint(self, endpoint, *args, **kwargs):
...
-            if name_field in new_args:
+            new_data = kwargs.get('data', {}).copy()
+            if name_field in new_data:
                 self.fail_json(msg="You can't specify the field {0} in your search data if using the name_or_id field".format(name_field))
 
-            new_args['or__{0}'.format(name_field)] = name_or_id
             try:
-                new_args['or__id'] = int(name_or_id)
+                new_data['or__id'] = int(name_or_id)
+                new_data['or__{0}'.format(name_field)] = name_or_id
```
AlanCoding, line 128: @beeankha I made this change in part because it made the error message better.

But I also don't know how correct it is to do `/api/v2/inventories/?or__name=foo&organization=2`. If that's the query you're doing, it's better to just get rid of the `or__`, because that only matters if the id-based lookup is in play, and it's not here.

kdelee (approved):
:bow: thanks for digging into this Alan!

squidboylan (approved):

beeankha (approved):

AlanCoding:
regate

AlanCoding:
regate

PR #8113, opened 2020-09-09
Start hooking in execution environments

jbradberry:
##### SUMMARY

- add a user-configurable setting for the global default EE
- add the ability to set Project-level default execution environments
- add an execution environment admin role that Org admins can delegate
- add a new Container Registry managed credential type

related #7064 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
awx: 14.1.0
```

This PR has been closed, and the changes ported over to https://github.com/shanemcd/awx/pull/19


AlanCoding:
whoops, looks like there was a rebasing issue

awx/main/conf.py
```diff
@@ -164,6 +165,18 @@ def _load_default_license_from_file():
     read_only=True,
 )
 
+register(
+    'DEFAULT_EXECUTION_ENVIRONMENT',
+    field_class=fields.PrimaryKeyRelatedField,
+    allow_null=True,
```
AlanCoding, line 170: What would be the behavior when this is null?

jbradberry: The most recent `managed_by_tower = True` version would be used.

AlanCoding: So then we still need some code to pre-create a default EE? Or is that already in the branch somewhere?

jbradberry: Yes, we still need it and it isn't in-place yet.

AlanCoding: What if, instead of creating it in a migration of create_preload_data (I do not believe the latter is reliable), we create the default execution environment ad-hoc if a job tries to launch while no `managed_by_tower` execution environments exist? Just thinking that might be easier and more reliable.

awx/main/models/unified_jobs.py
```diff
@@ -376,6 +394,8 @@ def create_unified_job(self, **kwargs):
             for fd, val in eager_fields.items():
                 setattr(unified_job, fd, val)
 
+        unified_job.execution_environment = self.resolve_execution_environment()
```
AlanCoding, line 397: this seems fine to me, but with this addition, "execution_environment" should be removed from the list here.

https://github.com/ansible/awx/blob/4f1eafec26487892c58bbe49856c3e58755613d4/awx/main/models/jobs.py#L287

AlanCoding: No, maybe I'm wrong, because you check

```python
        if self.execution_environment is not None:
            return self.execution_environment
```

so I guess the pattern is that `execution_environment` on the job can be set by the unified job template, and if it is, that comes highest precedence, so it is respected.

If going _this_ route, then it seems `execution_environment` should be added to the unified job fields for inventory sources, since those have historically allowed overriding at the UJT level too.

awx/main/models/unified_jobs.py
```diff
@@ -338,6 +339,23 @@ def notification_templates(self):
...
+        """
+        if self.execution_environment is not None:
+            return self.execution_environment
+        if getattr(self, 'project_id', None) and self.project.default_environment is not None:
+            return self.project.default_environment
+        if getattr(self, 'organization', None) and self.organization.default_environment is not None:
+            return self.organization.default_environment
+        if getattr(self, 'inventory', None) and self.inventory.organization is not None:
+            if self.inventory.organization.default_environment is not None:
+                return self.inventory.organization.default_environment
```
AlanCoding, line 354: It seems strange that inventory would come after organization in the selection logic.

Also, that creates a lot of confusion with inventory updates, since there would be an EE on inventory and an EE on the inventory source. IIRC, with custom virtual environments, inventory sources could have a custom venv specified, but inventory could not.

awx/main/models/credential/__init__.py
```diff
@@ -1187,6 +1186,37 @@ def create(self):
...
+            'id': 'username',
+            'label': ugettext_noop('Username'),
+            'type': 'string',
+        }, {
+            'id': 'password',
+            'label': ugettext_noop('Password'),
+            'type': 'string',
+            'secret': True,
+        }, {
+            'id': 'token',
```
AlanCoding, line 1209: So this isn't hooked up yet, I'd like to get some definition around that. What I've seen before is

```
podman login -u='user' -p='token'
```

That's used for pushing. So obviously you need your token to push to quay.io. However, I have not experimented with pulling from quay private repos or anything like that. I'm also unclear how this will work with Automation Hub authentication, to use their image registry, and the details of that authentication are certain to be different. But this is just 1 credential type, so how would we deal with that? And does it even work?

Also, does this need some form of support on the ansible-runner side? I'm tempted to say that we're getting too far ahead of ourselves, and maybe we can just get through the upstream release without any support for pulling images that require authentication.

awx/main/models/unified_jobs.py
jbradberry: `Inventory` does not have an EE.  It is used here solely for the purpose of inferring an `Organization`, for jobs that fail to have one more directly defined.

`InventorySource` _does_ have an EE, however, I have been very clear about maintaining a strict separation between "EE as explicitly set for use when executing" (always `self.execution_environment`) and "EE as a fallback option" (always `self.default_environment`).

jbradberry: To be more clear, `InventorySource` has `self.execution_environment`.  The only models that have `self.default_environment` are / will be `Organization` and `Project`.

awx/main/models/unified_jobs.py
jbradberry: > this seems fine to me, but with this addition, "execution_environment" should be removed from the list here.
> 
> https://github.com/ansible/awx/blob/4f1eafec26487892c58bbe49856c3e58755613d4/awx/main/models/jobs.py#L287

What's the reasoning behind that?

jbradberry: > No, maybe I'm wrong, because you check
> 
> ```python
>         if self.execution_environment is not None:
>             return self.execution_environment
> ```
> 
> so I guess the pattern is that `execution_environment` on the job can be set by the unified job template, and if it is, that comes highest precedence, so it is respected.
> 
> If going _this_ route, then it seems `execution_environment` should be added to the unified job fields for inventory sources, since those have historically allowed overriding at the UJT level too.

The `execution_environment` field on a job (as opposed to a job template) is supposed to be the already precedence-resolved EE to be used while executing.

As I noted in another comment, `InventorySource` does have `execution_environment` defined on itself, by virtue of inheriting from `UnifiedJobTemplate`.

awx/main/models/unified_jobs.py
AlanCoding: Thanks, we're on the same page about inventory source.

But in what case is it desirable to use `inventory.organization.default_environment`? I don't think that's desirable to JTs, it would confuse more than it would help. I could see that for the case of inventory updates where the inventory source does not set an EE.

So maybe make this conditional specific to the inventory update job type?

awx/main/models/credential/__init__.py
jbradberry: The three registries I looked into all take either username+password or a token.  If someone has specific knowledge of what changes are needed here to do the right thing, I would love to hear it.

awx/main/models/unified_jobs.py
AlanCoding: This is the list of fields for InventorySource to copy over when `create_unified_jobs` is called

https://github.com/ansible/awx/blob/4f1eafec26487892c58bbe49856c3e58755613d4/awx/main/models/inventory.py#L1030-L1034

This should contain `execution_environment` by the same measure that JT does, if that's the design.

awx/main/models/unified_jobs.py
jbradberry: Is there a non-contrived case where a JT would not have an Org but would wind up getting its EE from an inventory Org instead?  And would using that EE actually be harmful instead of falling back all of the way to the global default EE?

awx/main/models/unified_jobs.py
jbradberry: It made sense when I did it in the first PR, but I suspect that you are right that it is no longer needed then.

PR #8121, opened 2020-09-10
address a few follow-up issues for Org -> Galaxy Credentials support

ryanpetrello:
- add support for managing galaxy creds in the tower organization module
- fix a minor serializer bug

### DIFF
--- a/awx_collection/tests/integration/targets/tower_organization/tasks/main.yml
+++ b/awx_collection/tests/integration/targets/tower_organization/tasks/main.yml
@@ -12,6 +12,8 @@
 - name: "Create a new organization"
   tower_organization:
     name: "{{ org_name }}"
+    galaxy_credentials:
+      - Ansible Galaxy
   register: result
 
 - assert:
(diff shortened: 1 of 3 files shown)


awx_collection/tests/integration/targets/tower_organization/tasks/main.yml
```diff
@@ -12,6 +12,8 @@
 - name: "Create a new organization"
   tower_organization:
     name: "{{ org_name }}"
+    galaxy_credentials:
+      - Ansible Galaxy
   register: result
```
AlanCoding, line 17: I'd also like to put this in `EXAMPLES` in `awx_collection/plugins/modules/tower_organization.py`

because that is what will go live at https://docs.ansible.com/ansible/2.10/collections/awx/awx/tower_organization_module.html

AlanCoding (approved):

awx_collection/tests/integration/targets/tower_organization/tasks/main.yml
beeankha: Agreed.

beeankha (approved):
LGTM!

PR #8125, opened 2020-09-10
Fix malformed urls

jakemcdermott:
##### SUMMARY
The `baseUrl` in our models contains a trailing slash:

```javascript
this.baseUrl = '/api/v2/workflow_job_templates/';`
```

and so an additional slash before joining it to an id can cause problems in the build:

```javascript
return this.http.options(`${this.baseUrl}/${id}/`); // '/api/v2/workflow_job_templates//:id'
```

### DIFF
--- a/awx/ui_next/src/api/models/JobTemplates.js
+++ b/awx/ui_next/src/api/models/JobTemplates.js
@@ -25,7 +25,7 @@ class JobTemplates extends SchedulesMixin(
   }
 
   readTemplateOptions(id) {
-    return this.http.options(`${this.baseUrl}/${id}/`);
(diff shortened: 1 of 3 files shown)


dsesami:
~role disassociation from a WFJT needs to be added~

this is not true; rogue caching in action :)

dsesami (approved):

dsesami:
recheck

jakemcdermott:
recheck

dsesami:
recheck

dsesami:
regate (we'll keep trying this until it goes)

dsesami:
regate

dsesami:
recheck

shanemcd:
recheck

PR #8126, opened 2020-09-10
Changes to get execution environments factories working

AlanCoding:


awxkit/awxkit/api/pages/execution_environments.py
```diff
@@ -19,6 +20,29 @@ class ExecutionEnvironment(HasCreate, base.Base):
     dependencies = [Organization, Credential]
     NATURAL_KEY = ('organization', 'image')
 
+    # fields are image, organization, managed_by_tower, credential
+    def create(self, image='quay.io/ansible/ansible-runner:devel', credential=None, **kwargs):
```
jbradberry, line 24: Is this truly what we want for the default image?

AlanCoding: This isn't the AWX default image. _That_ will include some [content](https://github.com/ansible/awx/blob/devel/requirements/collections_requirements.yml) that we already provide in the Ansible venv. For that, I've built https://quay.io/repository/alancoding/awx-ee?tab=info, which needs to be rebuilt, and an official image is still TBD.

No matter what kind of thing you're making, there is no viable pathway to making a working image unless you use the ansible-runner [Dockerfile](https://github.com/ansible/ansible-runner/blob/devel/Dockerfile). Because if you don't use that, it won't work with ansible-runner, and thus not work with AWX in any way, shape, or form.

Out of the images that are built from this Dockerfile, you may consider https://hub.docker.com/r/ansible/ansible-runner or https://quay.io/repository/ansible/ansible-runner. Out of these two, only the quay.io one is built continuously on an automated trigger. The Dockerhub one is effectively already out-of-date and non-functional.

But for making a custom execution environment, you don't _want_ the AWX default execution environment that has `azure.azcollection` in it, because that's probably not necessary for whatever minor thing it is you're trying to verify in a test. If you want the minimum thing that basically works, that's the quay.io ansible-runner image... for now.

jbradberry (approved):

PR #8131, opened 2020-09-11
Remove psutil from ansible venv

wenottingham:
Hypothesis: it is not needed by anything, this is all encapsulated in
runner (which doesn't run out of this venv).

### DIFF
--- a/requirements/requirements_ansible.in
+++ b/requirements/requirements_ansible.in
@@ -51,8 +51,6 @@ netaddr
 # oVirt/RHV
 ovirt-engine-sdk-python==4.3.0    # minimum set inside Ansible facts module requirements
 pycurl==7.43.0.1    # higher versions will not install without SSL backend specified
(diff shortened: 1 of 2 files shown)


rooftopcellist (approved):
LGTM.  Hurray for removing deps!

beeankha (approved):

rebeccahhh:
~what about [here](https://github.com/ansible/awx/blob/defa334bd924954058e0e0494ee763b699e6dbd0/awx/main/dispatch/pool.py#L105) and [here?](https://github.com/ansible/awx/blob/8996d0a464ca33e862c43d1ac284f0cd712a4312/awx/main/utils/common.py#L714)~
retracted, I was multitasking poorly : )

wenottingham:
@AlanCoding pointed me to the runner callback plugin, which runs in this context. Oops.

PR #8132, opened 2020-09-11
Make ansible venv psutil match awx venv version

wenottingham:
##### SUMMARY

The other side of https://github.com/ansible/awx/pull/8131 - if we do need it, it should match everywhere.


ryanpetrello (approved):
WFM

PR #8136, opened 2020-09-13
Support `!import` and `!include` in `awx import -f yaml` command

### LINKED ISSUE #8135: Support `!include` and `!import` constructors in `awx import -f yaml` command
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Support `!import` and `!include` with `awx import -f yaml` command. This is useful when working with a large number of AWX resources. This allows breaking config into smaller files, and having a top-level file like this:

```yaml
# resources.yaml
projects:
- !include projects/project1.yaml
- !include projects/project2.yaml
job_templates:
- !include jobs/job1.yaml

# ....
```

Which allows managing resources in individual files (e.g. using a friendly directory structure), and still being able to import with `awx import -f yaml < resources.yaml`.

#### SUGGESTED IMPLEMENTATION

Use `awxkit.yaml_file.Loader` instead of `yaml.safe_load` in https://github.com/ansible/awx/blob/devel/awxkit/awxkit/cli/resource.py#L143

neoaggelos:
##### SUMMARY

Closes #8135

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 14.1.0
```

##### ADDITIONAL INFORMATION

The description in #8135 should be enough, let me know if something is not clear.

```

```

### DIFF
--- a/awxkit/awxkit/cli/resource.py
+++ b/awxkit/awxkit/cli/resource.py
@@ -2,7 +2,7 @@ import yaml
 import json
 import os
 
-from awxkit import api, config
+from awxkit import api, config, yaml_file
 from awxkit.exceptions import ImportExportError
 from awxkit.utils import to_str
 from awxkit.api.pages import Page
@@ -140,7 +140,7 @@ class Import(CustomCommand):
         if fmt == 'json':
             data = json.load(client.stdin)
         elif fmt == 'yaml':
-            data = yaml.safe_load(client.stdin)
+            data = yaml.load(client.stdin, Loader=yaml_file.Loader)
         else:
             raise ImportExportError("Unsupported format for Import: " + fmt)
 
--- a/awxkit/test/cli/test_format.py
+++ b/awxkit/test/cli/test_format.py
@@ -63,6 +63,7 @@ def test_yaml_import():
           - name: Workflow1
         """
     )
+    yaml_fd.name = 'file.yaml'
     cli = CLI(stdin=yaml_fd)
     cli.parse_args(['--conf.format', 'yaml'])
     cli.v2 = MockedV2()


neoaggelos:
This is a proposal solution, keeping the diff as small as possible. I can update/change as required if something more complex may be needed.

awxkit/awxkit/cli/resource.py
```diff
@@ -140,7 +140,7 @@ def handle(self, client, parser):
         if fmt == 'json':
             data = json.load(client.stdin)
         elif fmt == 'yaml':
-            data = yaml.safe_load(client.stdin)
```
ryanpetrello, line 143: There's some alarm bells going off for me for the security implications here:

https://pyyaml.org/wiki/PyYAMLDocumentation

> Warning: It is not safe to call yaml.load with any data received from an untrusted source! yaml.load is as powerful as pickle.load and so may call any Python function. Check the yaml.safe_load function though.

neoaggelos: Yes, I am aware of `load` vs `safe_load`, but this is not an issue here, since we do specify a "safe" loader.

The safety implications of `yaml.load()` stem from the fact that `yaml.Loader` is unsafe. Here, we use `awxkit.yaml_file.Loader` which inherits from `yaml.SafeLoader`, so to me it seems equivalent from a safety standpoint.

Also, `yaml.safe_load()` is simply an alias for `yaml.load(..., loader=SafeLoader)`.

ryanpetrello: Ah, yep!

Sorry, I wasn't looking closely enough at this PR.  Thanks, @neoaggelos.

ryanpetrello (approved):
This looks pretty reasonable to me.

@jbradberry or @elyezer either of you want to give it awhirl and sign off?

jbradberry (approved):
Works for me, barring some problems with a dependency that wasn't installed.

PR #8138, opened 2020-09-14
K8s ServiceAccount variabilization

moreiramarti:
##### SUMMARY
K8s service account variabilization

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - Installer

##### AWX VERSION
awx: 14.1.0

##### ADDITIONAL INFORMATION
Keep same names between k8s resources

### DIFF
--- a/installer/roles/kubernetes/defaults/main.yml
+++ b/installer/roles/kubernetes/defaults/main.yml
@@ -40,6 +40,7 @@ openshift_pg_emptydir: false
 openshift_pg_pvc_name: postgresql
 
 kubernetes_deployment_name: awx
+kubernetes_serviceaccount_name: awx
 
 kubernetes_deployment_replica_size: 1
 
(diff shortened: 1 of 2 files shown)


bbayszczak (approved):

installer/roles/kubernetes/defaults/main.yml
```diff
@@ -40,6 +40,7 @@ openshift_pg_emptydir: false
 openshift_pg_pvc_name: postgresql
 
 kubernetes_deployment_name: awx
+kubernetes_serviceaccount_name: "{{ kubernetes_deployment_name }}"
```
jakemcdermott:
```suggestion
kubernetes_serviceaccount_name: awx
```

This could be a breaking change for those already using a non-default `kubernetes_deployment_name`. Can you set the default value to 'awx` instead?

jakemcdermott (changes requested):

installer/roles/kubernetes/defaults/main.yml
moreiramarti: Sure, I hadn't thought about that
It's pushed

jakemcdermott (approved):
lgtm. thanks @moreiramarti

PR #8141, opened 2020-09-14
Print one targz per line (analytics gather)

kdelee:
Printing out a python like list is hard to process for tests
Better to print out one tarball per line

still testing this

follow up to https://github.com/ansible/awx/pull/7709


hendersonreed (commented):
Looks correct to me!

hendersonreed (approved):

beeankha (approved):

rooftopcellist (approved):
This looks good to me.  I don't see any way that this could cause any issues as it is only used when run manually and only changes the format of the log messages.  If it makes it easier to parse, then I'm OK with this.

PR #8146, opened 2020-09-15
Job nodes sometimes take time to appear

kdelee:
poll until it shows up, but give up eventually

This helps tower-qa use of awxkit but should be transparent to anyone else


chrismeyersfsu:
I wouldn't needing to poll. We create a workflow job and copy the workflow job template nodes over to new workflow job nodes. All in the request/response cycle as a transaction. https://github.com/ansible/awx/blob/devel/awx/api/views/__init__.py#L3214

kdelee:
:sob:

PR #8147, opened 2020-09-15
move an optional import for awxkit

ryanpetrello:
I'm not sure that this function is actually in use anywhere anymore, but
it shouldn't be a top-level import because it represents an optional
dependency.


jbradberry (approved):
Seems reasonable for now.

Issue #7878, opened 2020-08-11
Notifications List: websocket support


keithjgrant:
Follow-up on #7867

Update status labels beside notifications as they change via websocket messages.

##### ISSUE TYPE
 - Feature Idea

tiagodread:
Tests covered by 5534 PR already merged

PR #8150, opened 2020-09-15
Update status after sending test notification

keithjgrant:
##### SUMMARY
When sending a test notification, the status of the notification template is set to "Pending". This adds a poll request on a 5 second interval to detect when the status changes to Successful, Error, or any other state.

Addresses #7878 — after digging into the old UI, I discovered this is not accomplished with websockets, since notification status information isn't available via WS. This mimics the polling behavior of the old UI.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI


jakemcdermott (approved):
lgtm

tiagodread (approved):
Tests are Covered by 5534, after fixing the broken test we can merge.

PR #8152, opened 2020-09-16
Fix AWX collection import test interference with Default organization

AlanCoding:
Other tests rely on the Default organization existing, and this test was renaming it.

### DIFF
--- a/awx_collection/tests/integration/targets/tower_import/tasks/main.yml
+++ b/awx_collection/tests/integration/targets/tower_import/tasks/main.yml
@@ -25,7 +25,7 @@
                 notification_templates_error: []
                 notification_templates_approvals: []
               natural_key:
-                name: "Default"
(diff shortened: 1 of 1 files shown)


squidboylan (approved):

beeankha (approved):

awx_collection/tests/integration/targets/tower_import/tasks/main.yml
```diff
@@ -48,15 +48,15 @@
...
+                name: "{{ org_name1 }}"
                 type: "organization"
       register: import_output
       ignore_errors: true
 
     - assert:
         that:
-          - import_output is failed
-          - "'Organization with this Name already exists' in import_output.msg"
+          - import_output is not failed
```
AlanCoding, line 58: Change of expectation here, because it is believed that the failure was related to renaming, which we're getting rid of here.

PR #8166, opened 2020-09-17
Fix analytics logging

wenottingham:
The analytics change PR adjusted the logging for awx.analytics,
which solved the issue, but should have used the targeted awx.main.analytics.

Also flip a couple of loggers to use the regular awx.analytics (awx analytics)
logger instead of awx.main.analytics (the automation anayltics task system).

### DIFF
--- a/awx/settings/defaults.py
+++ b/awx/settings/defaults.py
@@ -987,6 +987,11 @@ LOGGING = {
             'handlers': ['task_system', 'external_logger'],
             'propagate': False
         },
+        'awx.main.analytics': {
+            'handlers': ['task_system', 'external_logger'],
+            'level': 'INFO',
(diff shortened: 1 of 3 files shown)


awx/settings/defaults.py
```diff
@@ -1001,7 +1006,7 @@ def IS_TESTING(argv=None):
             'level': 'INFO',  # very verbose debug-level logs
         },
         'awx.analytics': {
-            'handlers': ['task_system', 'external_logger'],
```
wenottingham, line 1004: Specific revert of a commit in #7709, see https://github.com/ansible/awx/pull/7709#discussion_r489877104

AlanCoding (approved):
looks good, thanks

PR #8167, opened 2020-09-17
Add support for a `--status` to the callback receiver (and improve our approach to stats collection in general)

ryanpetrello:
image

Worth mentioning that I removed the `finished` and `current_task` state tracking from the callback receiver because all of that IPC at high volume workloads is just too expensive and not worth it.

### DIFF
--- a/awx/main/dispatch/control.py
+++ b/awx/main/dispatch/control.py
@@ -2,6 +2,9 @@ import logging
 import uuid
 import json
 
+from django.conf import settings
+import redis
+
 from awx.main.dispatch import get_local_queuename
 
 from . import pg_bus_conn
@@ -21,7 +24,9 @@ class Control(object):
         self.queuename = host or get_local_queuename()
 
     def status(self, *args, **kwargs):
-        return self.control_with_reply('status', *args, **kwargs)
+        r = redis.Redis.from_url(settings.BROKER_URL)
+        stats = r.get(f'awx_{self.service}_statistics') or b''
+        return stats.decode('utf-8')
 
     def running(self, *args, **kwargs):
         return self.control_with_reply('running', *args, **kwargs)
--- a/awx/main/dispatch/pool.py
+++ b/awx/main/dispatch/pool.py
@@ -5,6 +5,7 @@ import signal
 import sys
 import time
 import traceback
+from datetime import datetime
 from uuid import uuid4
 
 import collections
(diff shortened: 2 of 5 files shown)


ryanpetrello:
cc @gamuniz please feel free to give this a whirl

awx/main/dispatch/pool.py
```diff
@@ -309,6 +337,10 @@ def __init__(self, *args, **kwargs):
         # max workers can't be less than min_workers
         self.max_workers = max(self.min_workers, self.max_workers)
 
+    def debug(self, *args, **kwargs):
+        self.cleanup()
+        return super(AutoscalePool, self).debug(*args, **kwargs)
```
ryanpetrello, line 342: The autoscale pool used by the dispatcher cleans up its state to detect orphaned processes before stats are reported.

awx/main/dispatch/pool.py
```diff
@@ -236,13 +258,13 @@ def up(self):
...
     def debug(self, *args, **kwargs):
-        self.cleanup()
         tmpl = Template(
+            'Recorded at: {{ dt }} \n'
             '{{ pool.name }}[pid:{{ pool.pid }}] workers total={{ workers|length }} {{ meta }} \n'
             '{% for w in workers %}'
             '.  worker[pid:{{ w.pid }}]{% if not w.alive %} GONE exit={{ w.exitcode }}{% endif %}'
             ' sent={{ w.messages_sent }}'
-            ' finished={{ w.messages_finished }}'
+            '{% if w.messages_finished %} finished={{ w.messages_finished }}{% endif %}'
```
ryanpetrello, line 267: For the callback receiver, this number will always be zero (because we don't track it).

awx/main/dispatch/pool.py
```diff
@@ -74,7 +83,8 @@ def put(self, body):
             if not body.get('uuid'):
                 body['uuid'] = str(uuid4())
             uuid = body['uuid']
-        self.managed_tasks[uuid] = body
+        if self.track_managed_tasks:
+            self.managed_tasks[uuid] = body
```
ryanpetrello, line 87: This should also cut down notably on memory consumption for high workloads that have a backlog cc @AlanCoding

awx/main/dispatch/control.py
```diff
@@ -21,7 +24,9 @@ def __init__(self, service, host=None):
         self.queuename = host or get_local_queuename()
 
     def status(self, *args, **kwargs):
-        return self.control_with_reply('status', *args, **kwargs)
+        r = redis.Redis.from_url(settings.BROKER_URL)
+        stats = r.get(f'awx_{self.service}_statistics') or b''
```
ryanpetrello, line 28: Each service (callback, dispatcher) has its own specific key in redis where the service periodically does a ` .put()` containing a view of internal state.

ryanpetrello: This model more closely matches the behavior in the new broadcast stats support @chrismeyersfsu added (and I like it better, because it means that as long as redis is running, we can still see the "last stats" even if one of these processes is having trouble).

chrismeyersfsu (approved):

ryanpetrello:
regate

PR #8174, opened 2020-09-18
`local_path` can now contain several folder levels

bbayszczak:
##### SUMMARY
`local_path` was previously defined using `os.path.basename` which means means specifying a `local_path` with several folder levels ended up in a wrong `project_path`

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - Models

##### AWX VERSION
awx: 14.1.0

##### ADDITIONAL INFORMATION

Lets take the following folder structure

```
/tmp/project
    ├── subproject1
    │   └── ansible
    ├── subproject2
    │   └── ansible
    └── subproject3
        └── ansible
```

If we configure a project for `subproject1`:
 - The `project_root` is `/tmp/project`
 - We specify the `local_path`: `subproject1/ansible`
Without this commit we ended the `local_path` `/tmp/project/ansible` which doesn't exists

### DIFF
--- a/awx/main/models/projects.py
+++ b/awx/main/models/projects.py
@@ -189,9 +189,8 @@ class ProjectOptions(models.Model):
         return cred
 
     def get_project_path(self, check_if_exists=True):
-        local_path = os.path.basename(self.local_path)
-        if local_path and not local_path.startswith('.'):
-            proj_path = os.path.join(settings.PROJECTS_ROOT, local_path)
+        if self.local_path and not self.local_path.startswith('.'):
+            proj_path = os.path.join(settings.PROJECTS_ROOT, self.local_path)
             if not check_if_exists or os.path.exists(smart_str(proj_path)):
                 return proj_path
 


bbayszczak:
Looks too big, maybe I'm missing a point ?

bbayszczak:
recheck

AlanCoding:
In what case are you specifying `local_path` for a project? Manual projects?

bbayszczak:
We were trying  to use the `local_path` to specify the path of our ansible folder in our repository for job templates
All our infra repository are structured like this:

```
/repo
    ├── terraform
    │   └── ...
    ├── ansible
        └── inventory
        └── roles
        └── play.yaml
        └── ...
```

We wanted to avoid having to create `collections` and `roles` folder at the root of our repository

AlanCoding:
That is not what it was intended to be used for. When you run a project update, it updates that particular path, and I don't know what the Ansible `git` module is going to do when given a path, not to a repo, but a subdirectory inside of it (`git` itself is one thing, but the module adds logic on top of it).

This `local_path` field should be for internal use only, aside from manual projects, which is why I ask. There are all kinds of implications of allowing this kind of use, and those need to be vetted from the original use case.

PR #8177, opened 2020-09-18
fix an rrule bug that causes improper HOURLY/MINUTELY calculation

ryanpetrello:
see: https://github.com/ansible/awx/issues/8071

### DIFF
--- a/awx/main/models/schedules.py
+++ b/awx/main/models/schedules.py
@@ -205,10 +205,15 @@ class Schedule(PrimordialModel, LaunchTimeConfig):
                     'A valid TZID must be provided (e.g., America/New_York)'
                 )
 
-        if fast_forward and ('MINUTELY' in rrule or 'HOURLY' in rrule):
+        if (
+            fast_forward and
+            ('MINUTELY' in rrule or 'HOURLY' in rrule) and
+            'COUNT=' not in rrule
+        ):
             try:
                 first_event = x[0]
-                if first_event < now():
+                # If the first event was over a week ago...
+                if (now() - first_event).days > 7:
                     # hourly/minutely rrules with far-past DTSTART values
                     # are *really* slow to precompute
                     # start *from* one week ago to speed things up drastically
--- a/awx/main/tests/functional/models/test_schedule.py
+++ b/awx/main/tests/functional/models/test_schedule.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timedelta
 from contextlib import contextmanager
 
 from django.utils.timezone import now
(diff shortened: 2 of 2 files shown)


awx/main/models/schedules.py
```diff
@@ -205,10 +205,15 @@ def rrulestr(cls, rrule, fast_forward=True, **kwargs):
                     'A valid TZID must be provided (e.g., America/New_York)'
                 )
 
-        if fast_forward and ('MINUTELY' in rrule or 'HOURLY' in rrule):
+        if (
+            fast_forward and
+            ('MINUTELY' in rrule or 'HOURLY' in rrule) and
+            'COUNT=' not in rrule
```
ryanpetrello, line 211: We really only need this optimization if people set up *unbounded* per minute/hour rules in the past.

awx/main/models/schedules.py
```diff
@@ -205,10 +205,15 @@ def rrulestr(cls, rrule, fast_forward=True, **kwargs):
...
+        ):
             try:
                 first_event = x[0]
-                if first_event < now():
+                # If the first event was over a week ago...
+                if (first_event - now()).days > 7:
```
ryanpetrello, line 216: Similarly, if the first per-minute/per-hour event *is actually in the past week*, it's not going to be expensive, so we can skip this optimization.

jakemcdermott:
arrrr

AlanCoding (approved):

jakemcdermott (approved):

awx/main/tests/functional/models/test_schedule.py
```diff
@@ -161,6 +161,57 @@ def test_computed_fields_turning_off_by_deleting(self, job_template):
...
+    first_event = sched.rrulestr(sched.rrule)[0]
+    assert first_event.replace(tzinfo=None) == recent
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('freq, delta', (
+    ('MINUTELY', 1),
+    ('HOURLY', 1)
+))
+def test_really_old_dtstart(job_template, freq, delta):
```
jakemcdermott, line 189: lgtm

awx/main/models/schedules.py
```diff
@@ -205,10 +205,15 @@ def rrulestr(cls, rrule, fast_forward=True, **kwargs):
...
+        if (
+            fast_forward and
+            ('MINUTELY' in rrule or 'HOURLY' in rrule) and
+            'COUNT=' not in rrule
+        ):
             try:
                 first_event = x[0]
-                if first_event < now():
+                # If the first event was over a week ago...
+                if (now() - first_event).days > 7:
```
ryanpetrello, line 216: Similarly, if the first per-minute/per-hour event is actually _in the past week_, it's not going to be expensive to burst out the rrule, so we can skip this optimization.

PR #8185, opened 2020-09-18
Use inventory and env private_data_dir subfolders

AlanCoding:
##### SUMMARY
This is another part of trying to get AWX to follow `ansible-runner` best practices.

See docs:

https://ansible-runner.readthedocs.io/en/stable/intro.html#runner-input-directory-hierarchy

So what runner & people expect is:

```
.
├── env
│   ├── envvars
│   ├── extravars
│   ├── passwords
│   ├── cmdline
│   ├── settings
│   └── ssh_key
├── inventory
│   └── hosts
└── project
    ├── test.yml
    └── roles
        └── testrole
            ├── defaults
            ├── handlers
            ├── meta
            ├── README.md
            ├── tasks
            ├── tests
            └── vars
```

What we were producing was something more like

```
├── cp
├── env
│   ├── cmdline
│   ├── envvars
│   ├── extravars
│   ├── passwords
│   ├── settings
│   └── ssh_key
├── project
│   ├── ansible.cfg
│   ├── ansible_env.yml
│   ├── async_tasks.yml
<snip>
│   └── vault.yml
└── tmp3evnorsu
```

There was no inventory folder, and inventory was stored in `tmp3evnorsu` as a top-level file. Any credential files were also saved in that same structure.

With this change, it's more like:

```
├── cp
├── env
│   ├── cmdline
│   ├── envvars
│   ├── extravars
│   ├── passwords
│   └── settings
├── inventory
│   └── tmp94xjm0d7
└── project
    ├── ansible.cfg
    ├── ansible_env.yml
    ├── async_tasks.yml
<snip>
    └── vault.yml
```

and any credential files (starting with `tmp`) are dumped into the `env` folder.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
14.1.0
```

##### ADDITIONAL INFORMATION
The motivation for this is that `ansible-runner` may require specific mounts for every file of directory in the private_data_dir. As such, we really don't want randomized filenames in the top level directory. I don't know for sure this is the direction we are going to go, but this seems like a defensible change on its own, to better adhere to the practices someone familiar with `ansible-runner` would expect.

### DIFF
--- a/awx/main/models/credential/injectors.py
+++ b/awx/main/models/credential/injectors.py
@@ -6,6 +6,8 @@ import tempfile
 
 from django.conf import settings
 
+from awx.main.utils.execution_environments import to_container_path
+
 
 def aws(cred, env, private_data_dir):
     env['AWS_ACCESS_KEY_ID'] = cred.get_input('username', default='')
@@ -30,9 +32,9 @@ def gce(cred, env, private_data_dir):
     json.dump(json_cred, f, indent=2)
     f.close()
     os.chmod(path, stat.S_IRUSR | stat.S_IWUSR)
-    cred_path = os.path.join('/runner', 'env', os.path.basename(path))
-    env['GCE_CREDENTIALS_FILE_PATH'] = cred_path
-    env['GCP_SERVICE_ACCOUNT_FILE'] = cred_path
+    container_path = to_container_path(path, private_data_dir)
+    env['GCE_CREDENTIALS_FILE_PATH'] = container_path
+    env['GCP_SERVICE_ACCOUNT_FILE'] = container_path
 
     # Handle env variables for new module types.
     # This includes gcp_compute inventory plugin and
@@ -103,8 +105,7 @@ def openstack(cred, env, private_data_dir):
     yaml.safe_dump(openstack_data, f, default_flow_style=False, allow_unicode=True)
     f.close()
     os.chmod(path, stat.S_IRUSR | stat.S_IWUSR)
-    # TODO: constant for container base path
-    env['OS_CLIENT_CONFIG_FILE'] = os.path.join('/runner', 'env', os.path.basename(path))
+    env['OS_CLIENT_CONFIG_FILE'] = to_container_path(path, private_data_dir)
 
 
 def kubernetes_bearer_token(cred, env, private_data_dir):
@@ -116,6 +117,6 @@ def kubernetes_bearer_token(cred, env, private_data_dir):
         with os.fdopen(handle, 'w') as f:
             os.chmod(path, stat.S_IRUSR | stat.S_IWUSR)
             f.write(cred.get_input('ssl_ca_cert'))
-        env['K8S_AUTH_SSL_CA_CERT'] = os.path.join('/runner', 'env', os.path.basename(path))
+        env['K8S_AUTH_SSL_CA_CERT'] = to_container_path(path, private_data_dir)
     else:
         env['K8S_AUTH_VERIFY_SSL'] = 'False'
--- a/awx/main/models/credential/injectors.py
+++ b/awx/main/models/credential/injectors.py
@@ -25,7 +25,7 @@ def gce(cred, env, private_data_dir):
         env['GCE_PROJECT'] = project
     json_cred['token_uri'] = 'https://oauth2.googleapis.com/token'
 
-    handle, path = tempfile.mkstemp(dir=private_data_dir)
+    handle, path = tempfile.mkstemp(dir=os.path.join(private_data_dir, 'env'))
     f = os.fdopen(handle, 'w')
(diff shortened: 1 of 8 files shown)


AlanCoding:
recheck

awx/main/tasks.py
```diff
@@ -1478,10 +1479,6 @@ def run(self, pk, **kwargs):
                     module_args = ansible_runner.utils.args2cmdline(
                         params.get('module_args'),
                     )
-                shutil.move(
-                    params.pop('inventory'),
-                    os.path.join(private_data_dir, 'inventory')
-                )
```
AlanCoding: I want this PR, if nothing else, for this. I hit a lot of errors related to container groups with execution environments that came back to this inventory move... which is something we shouldn't need and shouldn't exist.

chrismeyersfsu (approved):

ryanpetrello:
`<low whistle>` this PR is pretty old.

@AlanCoding @shanemcd does this make sense as something to throw into the EE branch while we're here, or do you think it would require an excessive amount of rebasing at this point?

AlanCoding:
I'm leaning towards letting `execution-environments` merge and then rebasing this. _This_ isn't that much content to rebase.

ryanpetrello:
@AlanCoding (when you're back) this might be good to rebase and work on getting in.

AlanCoding:
Hit friction in the rebase

```
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/tasks.py", line 1377, in run
    'inventory': self.build_inventory(self.instance, private_data_dir),
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/tasks.py", line 1072, in build_inventory
    os.makedirs(path, mode=0o700)
  File "/usr/lib64/python3.8/os.py", line 223, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/tmp/pdd_wrapper_19_w9mwdkjf/awx_19_0ja9wg3g/inventory'
```

That's due to deeper assumptions about directory creation order. I believe the `os.makedirs(path, mode=0o700)` line just gets deleted for this reason.

awx/main/models/credential/injectors.py
```diff
@@ -96,25 +97,25 @@ def _openstack_data(cred):
...
     env['K8S_AUTH_API_KEY'] = cred.get_input('bearer_token', default='')
     if cred.get_input('verify_ssl') and 'ssl_ca_cert' in cred.inputs:
         env['K8S_AUTH_VERIFY_SSL'] = 'True'
-        handle, path = tempfile.mkstemp(dir=private_data_dir)
+        handle, path = tempfile.mkstemp(dir=os.path.join(private_data_dir, 'env'))
         with os.fdopen(handle, 'w') as f:
             os.chmod(path, stat.S_IRUSR | stat.S_IWUSR)
             f.write(cred.get_input('ssl_ca_cert'))
-        env['K8S_AUTH_SSL_CA_CERT'] = os.path.join('/runner', os.path.basename(path))
+        env['K8S_AUTH_SSL_CA_CERT'] = os.path.join('/runner', 'env', os.path.basename(path))
```
AlanCoding: I'm much less happy with these than I thought I would be.

There are tons of places in code where we put something in a directory, and other part of code relies on knowledge of how the directory & naming of that file was done. I would like something more robust to changes, and harder to get wrong for the host / container paths.

awx/main/tests/unit/test_tasks.py
```diff
@@ -1053,7 +1068,7 @@ def test_gce_credentials(self, private_data_dir, job):
         safe_env = {}
         credential.credential_type.inject_credential(credential, env, safe_env, [], private_data_dir)
         runner_path = env['GCE_CREDENTIALS_FILE_PATH']
-        local_path = os.path.join(private_data_dir, os.path.basename(runner_path))
+        local_path = runner_path.replace('/runner', private_data_dir)  # container path to host path
```
AlanCoding: I know these are only the tests, but this looks like the direction I would like to start taking things in.

There _might_ be corner cases that this will have problems with... but in the big picture, this is a durable way to switch from the container to host path.

awx/main/models/credential/injectors.py
AlanCoding: So, to illustrate this point, let me write logic similar to what I've done in tests here

```python
env['K8S_AUTH_SSL_CA_CERT'] = path.replace(private_data_dir, '/runner')
```

Now, maybe this would be best implemented in a utility method (for better future control and error handling).

In general terms, this is where I'm interested in going, and I believe something like this would be hugely valuable for maintainability.

AlanCoding:
I'm happy with how the code here looks now.

Issue #8183, opened 2020-09-18
Add Setting detail views 


marshmalien:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Mockup: https://tower-mockups.testing.ansible.com/patternfly/settings/settings/

This issue tracks adding the Detail view to all setting views. 

* Setting content should be derived from the setting OPTIONS request
* Setting fields without a value should have a Detail value of "Unconfigured"
* Setting fields with sensitive information should have a Detail value of "Encrypted"

**RBAC**

* RBAC side-nav rules will be handled by this issue: https://github.com/ansible/awx/issues/4426 
* Only superusers and system auditors have permission to view settings
  * Redirect users without setting permissisons to home "/"
* Only superusers have permission to edit settings
  * If user is not a superuser, hide the edit button

**VIEWS**

- <a href="#user-content-activity-stream">Activity Stream</a>
- <a href="#user-content-azure">Azure AD</a>
- Github
    - <a href="#user-content-github-default">GitHub Default</a>
    - <a href="#user-content-github-organization">GitHub Organization</a>
    - <a href="#user-content-github-team">GitHub Team</a>
- <a href="#user-content-google">Google OAuth 2</a>
- <a href="#user-content-jobs">Jobs</a>
- LDAP
    - <a href="#user-content-ldap">Default</a>
    - LDAP 1
    - LDAP 2
    - LDAP 3
    - LDAP 4
    - LDAP 5
- <a href="#user-content-logging">Logging </a>
- <a href="#user-content-system">Miscellaneous </a>
- <a href="#user-content-radius">RADIUS </a>
- <a href="#user-content-saml">SAML </a>
- <a href="#user-content-tacacs">TACACS </a>
- <a href="#user-content-ui">User Interface</a>

<h2 id="activity-stream">Activity Stream</h2>

  * Enable Activity Stream
  * Enable Activity Stream for Inventory Sync

  
<h2 id="azure">Azure AD</h2>

  * Azure AD OAuth2 Callback URL
  * Azure AD OAuth2 Key
  * Azure AD OAuth2 Secret
  * Azure AD OAuth2 Organization Map
  * Azure AD OAuth2 Team Map

<h2 id="github-default">Github Default</h2>

  * GitHub OAuth2 Callback URL
  * GitHub OAuth2 Key
  * GitHub OAuth2 Secret
  * GitHub OAuth2 Organization Map
  * GitHub OAuth2 Team Map

[...]

PR #8186, opened 2020-09-18
Add setting details and unit tests

marshmalien:
##### SUMMARY
Issue: #8183

This PR adds details views to all setting/* routes. 

* Add complete configuration of setting breadcrumbs
* Add content to all setting detail views
* Setting fields with no value are displayed as "Unconfigured"
* Setting fields with a sensitive value are displayed as "Encrypted"
* Redirect any user who is not a superuser or system auditor to home
* Hide the ability to edit from non-superusers

The setting detail content is created by keying off of the setting's GET and OPTIONS results. At the top-level in Settings.jsx, the component fetches all of the OPTIONS from `/settings/all/` and stores it in the setting Context Provider. This provides a way for the children components (details/edit) to access the OPTIONS data without making extra API requests and it removes the need to drill down an additional prop. 

Each individual detail is created by passing data through a `<SettingDetail />` component. This component will render the appropriate Detail based on the detail's data "type". 

Design-wise, any detail rendered in a CodeMirror component should be shown at the bottom of the detail card. This required extra data manipulation in a few views (Jobs, LDAP, Logging, Misc. System), where I had to explicitly move the insertion order of "nested" data types to the bottom of the list. This is done via a util called `sortNestedDetails` (Thanks Kersom for helping me refactor this). A few views also use an additional util called `pluck` that helps by creating a new object based on select object properties. 

Since all of the setting detail unit tests asserted whether a detail or detail variable existed, I created also created a `settingTestUtils` file with a couple assertion methods. This could be moved to the top-level utils, but I wanted to hear what others thought. 

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - UI


unlikelyzero:
We noticed these differences between the old ui and the new ui:

- Tooltips are missing
- Revert to default value on fields is missing
- Revert All values to default is missing
- It is not immediately apparent that GitHub Categories are listed by Tab instead of Dropdown
- Mandatory/Required settings are not apparent on Details

Other notes:
- If a sub-setting has many tabs then clicking on a breadcrumb will not redirect the user but result in 404. i.e. Click on the breadcrumb for GitHub Auth from one of the Github Auth tabs
- We should have aria labels on these fields. We should have ids
- Responsivity issues at 1260 make the triple columns hard to read, maybe consider changing the responsivity from 2 to 3 columns at a higher width. Best example 1260 x 768 for Miscellaneous System / Details. General issues with dd/dt tables
- Custom Logo has broken image tag
- Is activity stream going to be a part of the ui_next project? Would we want to remove these settings?

Our plan is as follows:
- We will add some happy path tests for general navigation for all settings
- We will add happy path tests for modifying Job settings and Systems Settings
- We will add a visual test for each Settings Page

marshmalien:
https://github.com/ansible/awx/pull/8186#issuecomment-696928780

* Across the new UI we don’t show tooltips in the Detail views, but since it has been a popular request for Settings, I plan on adding them in this PR. 
* Revert is in the Form view, not Details.
* Revert All is in the Form view, not Details.
* I developed different navigation POCs and demoed them for @dsesami and @trahman73 . We came to the conclusion that Tab navigation was preferred over dropdown navigation. Do you have a suggestion for making the tab categories more apparent?
* Required fields will be identified in the Form view, not Details.
* General dd/dt table issues should be handled in a separate issue. 
* We will have activity stream in the new UI.
* A follow-up commit can handle: 
    * Broken redirects
    * Aria-labels and Ids
    * Not configured custom logo value
    * Detail popovers

unlikelyzero:
@marshmalien new field in devel: https://github.com/ansible/awx/pull/8213
cc @ryanpetrello

unlikelyzero:
@marshmalien 
We don't necessarily need the tooltips in the first PR (from QE), however, we do need some "testability" for our test automation. 

As I understand it, aria labels are requirement for a11y. Idk when we need them by, but I would imagine before we GA. Sometimes we use aria labels instead of unique IDs in test automation to satisfy our "testability" requirements.

jakemcdermott (approved):
There's a few more updates on the way which will address some of the qe concerns mentioned above & a few nitpicks I brought up w/ @marshmalien offline but otherwise this lgtm :+1:

jlmitch5 (approved):
similar to @jakemcdermott, talked with @marshmalien a bit about a couple things, but this pr looks good!  lots of tests is great, and should be easy to follow the pattern as we add new settings.

tiagodread:
@marshmalien I saw a curious thing, when the settings page load all the data is requested and if I click on `Azure` the only request is a GET for `/azuread-oauth2/` this happens for almost everything, except for Github and MIsc System.

notif

marshmalien:
@tiagodread I leveraged the setting categories to populate most of the detail fields since they (for the most part) mapped 1:1 to the UI views. For example, the GitHub Organization view is populated by the API response from `/api/v2/settings/github-org/` and the GitHub Team view is populated from the response from `/api/v2/settings/github-team/`. There were a couple setting views where the setting category api response did not map 1:1 to the UI. In Miscellaneous System settings, the “system” setting category `/api/v2/settings/system/` endpoint did not contain fields SESSION_COOKIE_AGE, SESSIONS_PER_USER, REFRESH_TOKEN_EXPIRE_SECONDS, etc., so instead I decided to fetch data from `/settings/all` and pluck only the fields that we wanted to show.

marshmalien:
@unlikelyzero Wouldn’t adding ids or an attribute other than aria-label be a better approach to satisfy testability requirements? We should be careful when adding aria-labels because sometimes they make accessibility worse by overriding or misrepresenting information. [No ARIA is better than Bad ARIA.]( https://www.w3.org/TR/wai-aria-practices-1.1/#no_aria_better_bad_aria)

unlikelyzero:
IDs would be preferred. Thank you for that ARIA link

marshmalien:
I added a few follow-up commits that should fix broken redirects, add detail popovers, handle not configured image details, and add unit values to detail fields that contain integers. The last item required a few additions to the api. I added an additional property “unit” to the settings OPTIONS response to help the UI render the correct unit behind Integer fields, ex. TCP Connection Timeout: 5 seconds. 

@tiagodread  found a bug where we were passing the value “[object Object]”  to all "Back to Location" tab aria-labels. The screen reader would say something like “Tab object object button” instead of “Back to Settings”. I pushed up a fix to RoutedTabs to catch when we pass an “Object” to the aria-label prop and render an empty string instead. Now whenever a screen read encounters a tab with an icon, it will read the visible text instead of relying on the aria-label.

_Popover_
Screen Shot 2020-09-24 at 2 19 34 PM

ryanpetrello (approved):

tiagodread (approved):
Visual tests 5559 and functional tests 5569 in place 👍🏽 
Great job!

PR #8191, opened 2020-09-21
remove multiprocessing.Queue usage from the callback receiver

ryanpetrello:
instead, just have each worker connect directly to redis
this has a few benefits:

- it's simpler to explain and debug
- back pressure on the queue keeps messages in redis (which is
  observable, and survives the restart of Python processes)
- it's likely more performant at high loads than a single consumer reading from redis and distributing to workers via per-worker IPC pipes

### DIFF
--- a/awx/main/dispatch/worker/base.py
+++ b/awx/main/dispatch/worker/base.py
@@ -132,22 +132,9 @@ class AWXConsumerRedis(AWXConsumerBase):
         super(AWXConsumerRedis, self).run(*args, **kwargs)
         self.worker.on_start()
 
-        time_to_sleep = 1
         while True:
-            while True:
-                try:
-                    res = self.redis.blpop(self.queues)
-                    time_to_sleep = 1
-                    res = json.loads(res[1])
-                    self.process_task(res)
-                except redis.exceptions.RedisError:
-                    time_to_sleep = min(time_to_sleep * 2, 30)
-                    logger.exception(f"encountered an error communicating with redis. Reconnect attempt in {time_to_sleep} seconds")
-                    time.sleep(time_to_sleep)
-                except (json.JSONDecodeError, KeyError):
-                    logger.exception("failed to decode JSON message from redis")
-                if self.should_stop:
-                    return
+            logger.debug(f'{os.getpid()} is alive')
+            time.sleep(60)
 
 
 class AWXConsumerPG(AWXConsumerBase):
--- a/awx/main/dispatch/worker/callback.py
+++ b/awx/main/dispatch/worker/callback.py
@@ -1,4 +1,5 @@
 import cProfile
+import json
 import logging
 import os
 import pstats
@@ -6,12 +7,15 @@ import signal
 import tempfile
 import time
 import traceback
-from queue import Empty as QueueEmpty
 
 from django.conf import settings
 from django.utils.timezone import now as tz_now
 from django.db import DatabaseError, OperationalError, connection as django_connection
 from django.db.utils import InterfaceError, InternalError, IntegrityError
+
+import psutil
+
+import redis
 
 from awx.main.consumers import emit_channel_notification
 from awx.main.models import (JobEvent, AdHocCommandEvent, ProjectUpdateEvent,
@@ -24,10 +28,6 @@ from .base import BaseWorker
 
 logger = logging.getLogger('awx.main.commands.run_callback_receiver')
 
-# the number of seconds to buffer events in memory before flushing
-# using JobEvent.objects.bulk_create()
-BUFFER_SECONDS = .1
-
 
 class CallbackBrokerWorker(BaseWorker):
     '''
@@ -39,21 +39,57 @@ class CallbackBrokerWorker(BaseWorker):
     '''
 
     MAX_RETRIES = 2
+    last_stats = time.time()
+    total = 0
+    last_event = ''
     prof = None
 
     def __init__(self):
         self.buff = {}
+        self.pid = os.getpid()
+        self.redis = redis.Redis.from_url(settings.BROKER_URL)
+        for key in self.redis.keys('awx_callback_receiver_statistics_*'):
+            self.redis.delete(key)
 
     def read(self, queue):
         try:
-            return queue.get(block=True, timeout=BUFFER_SECONDS)
-        except QueueEmpty:
-            return {'event': 'FLUSH'}
+            res = self.redis.blpop(settings.CALLBACK_QUEUE, timeout=settings.JOB_EVENT_BUFFER_SECONDS)
+            if res is None:
+                return {'event': 'FLUSH'}
+            self.total += 1
+            return json.loads(res[1])
+        except redis.exceptions.RedisError:
+            logger.exception("encountered an error communicating with redis")
+            time.sleep(1)
+        except (json.JSONDecodeError, KeyError):
+            logger.exception("failed to decode JSON message from redis")
+        finally:
+            self.record_statistics()
+        return {'event': 'FLUSH'}
+
+    def record_statistics(self):
+        # buffer stat recording to once per (by default) 5s
+        if time.time() - self.last_stats > settings.JOB_EVENT_STATISTICS_INTERVAL:
+            try:
+                self.redis.set(f'awx_callback_receiver_statistics_{self.pid}', self.debug())
+                self.last_stats = time.time()
+            except Exception:
+                logger.exception("encountered an error communicating with redis")
+                self.last_stats = time.time()
+
+    def debug(self):
+        return f'.  worker[pid:{self.pid}] sent={self.total} rss={self.mb}MB {self.last_event}'
+
+    @property
+    def mb(self):
+        return '{:0.3f}'.format(
+            psutil.Process(self.pid).memory_info().rss / 1024.0 / 1024.0
+        )
 
     def toggle_profiling(self, *args):
         if self.prof:
             self.prof.disable()
-            filename = f'callback-{os.getpid()}.pstats'
+            filename = f'callback-{self.pid}.pstats'
             filepath = os.path.join(tempfile.gettempdir(), filename)
             with open(filepath, 'w') as f:
                 pstats.Stats(self.prof, stream=f).sort_stats('cumulative').print_stats()
@@ -108,6 +144,8 @@ class CallbackBrokerWorker(BaseWorker):
     def perform_work(self, body):
         try:
             flush = body.get('event') == 'FLUSH'
+            if flush:
+                self.last_event = ''
             if not flush:
                 event_map = {
                     'job_id': JobEvent,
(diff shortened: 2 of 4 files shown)


awx/main/dispatch/worker/base.py
```diff
@@ -132,22 +132,9 @@ def run(self, *args, **kwargs):
...
-                except redis.exceptions.RedisError:
-                    time_to_sleep = min(time_to_sleep * 2, 30)
-                    logger.exception(f"encountered an error communicating with redis. Reconnect attempt in {time_to_sleep} seconds")
-                    time.sleep(time_to_sleep)
-                except (json.JSONDecodeError, KeyError):
-                    logger.exception("failed to decode JSON message from redis")
-                if self.should_stop:
-                    return
+            logger.debug(f'{os.getpid()} is alive')
+            time.sleep(60)
```
ryanpetrello, line 137: Might be able to just replace this w/ a loop that does a `.join()` on the children, will experiment.

ryanpetrello:
This is WIP because I can't merge this until I fix `run_callback_receiver --status`

awx/main/dispatch/worker/callback.py
```diff
@@ -123,6 +163,8 @@ def perform_work(self, body):
                         job_identifier = body[key]
                         break
 
+                self.last_event = f'\n\t- {cls.__name__} for #{job_identifier} ({body.get("event", "")} {body.get("uuid", "")})'  # noqa
```
ryanpetrello, line 164: Kapture 2020-09-22 at 18 07 21

👀 @gamuniz @chrismeyersfsu @AlanCoding @ghjm @jakemcdermott

awx/main/dispatch/worker/base.py
```diff
@@ -132,22 +132,9 @@ def run(self, *args, **kwargs):
...
         self.worker.on_start()
 
-        time_to_sleep = 1
         while True:
-            while True:
-                try:
-                    res = self.redis.blpop(self.queues)
-                    time_to_sleep = 1
-                    res = json.loads(res[1])
-                    self.process_task(res)
```
AlanCoding, line 142: After this diff, `process_task` still exists and is being used, apparently by the postgres consumer. This method writes to the multiprocessing queue.

So let me echo my understanding - you are removing use of the multiprocessing queue _only for_ the redis-connected callback receiver. However, other messages, such as "delete this inventory" or "start running this job" are sent through the postgres messaging system, and still go through the dispatcher's multiprocessing queue. So one node in the cluster may have 4 or 8 redis connections, but 1 connection to the postgres message bus.

I'm just trying to state facts, and hopefully get the right. I don't recall any evidence that the main dispatcher ever had significant buildup of messages in the queue. Perhaps this is because autoscaling is enabled for the ordinary dispatcher but not for the callback receiver (going by my memory, it may be the other way around).

ryanpetrello: Yep, you've got it right.

So under this new model, `process_task`, and the model where we utilize a `multiprocessing.Queue` to dispatch messages to the worker processes is really only used but the dispatcher.  The callback receiver still has a "main process", but it only exists to fork its children and hang out until they exit.

ryanpetrello: > I don't recall any evidence that the main dispatcher ever had significant buildup of messages in the queue. Perhaps this is because autoscaling is enabled for the ordinary dispatcher but not for the callback receiver.

This is mostly because the volume is way lower - you might be running hundreds (or _maybe_ thousands) of parallel jobs/tasks, but it's not uncommon for the callback receiver to deal in far more events in parallel.

AlanCoding: The greatest volume we had there was the computed fields task. That did fail things in fun ways. I think those tended to be related to the autoscaling, memory, database connection limit, and so on. So filling up _that_ IPC queue didn't tend to come up much, and computed fields should be muzzled now anyway. So I agree with your assessment about volume.

awx/main/dispatch/worker/callback.py
```diff
@@ -39,16 +43,50 @@ class CallbackBrokerWorker(BaseWorker):
...
+        except (json.JSONDecodeError, KeyError):
+            logger.exception("failed to decode JSON message from redis")
+        finally:
+            self.record_statistics()
+        return {'event': 'FLUSH'}
+
+    def record_statistics(self):
+        if time.time() - self.last_stats > 1:  # buffer stat recording to once per second
+            try:
+                self.redis.set(f'awx_callback_receiver_statistics_{os.getpid()}', self.debug())
```
AlanCoding: This seems a _little_ non-ideal. The fraction of the time that a human is actively watching the `--status` command is minuscule. I don't think it's particularly expensive, maybe like 2e-4 seconds (per every second), but it's adding to the overall background noise for messages that will never be read. I don't have an alternative idea that's not ugly, so it's just a passing thought.

AlanCoding (approved):

awx/main/dispatch/worker/callback.py
ryanpetrello: Yep, it _does_ have a cost.  That said, I think it's a drop in the bucket compared to other bottlenecks in event processing like (for example), the overhead of the Django ORM.

AlanCoding: How often are workers recycled, and do the keys get cleared? Should this line pass kwarg `ex=~1`?

https://redis-py.readthedocs.io/en/stable/_modules/redis/client.html

awx/main/dispatch/worker/base.py
ryanpetrello: Yea, the dispatcher, unlike the callback receiver, can detect when all the workers are busy, and can autoscale up *new* workers to _avoid_ a backlog in the IPC queues.  There *is* a limit to this, of course, but on well-provisioned hardware and clusters, it's difficult to hit (we're talking hundreds to thousands of tasks).

awx/main/dispatch/worker/callback.py
ryanpetrello: Callback workers don't get recycled; they exist until the process exits.

At process startup time, I unset all of these keys to wipe out the stats from the previous workers:

https://github.com/ansible/awx/pull/8191/files#diff-f37b92a11438678a6a32ac23a7790f05R54

ryanpetrello:
Some more metrics on the current version of this PR, single node install, `m5.16xlarge`, processing a queue backlog of 5M+ small events, external logging disabled:

```shell
[root@ip-10-0-15-214 ec2-user]# awx-manage print_settings JOB_EVENT_WORKERS
JOB_EVENT_WORKERS                        = 16
[root@ip-10-0-15-214 ec2-user]# awx-manage print_settings UI_LIVE_UPDATES_ENABLED
UI_LIVE_UPDATES_ENABLED                  = False
```

```shell
[root@ip-10-0-15-214 ec2-user]# awx-manage callback_stats
main_jobevent
↳  last minute 1835721
main_inventoryupdateevent
↳  last minute 0
main_projectupdateevent
↳  last minute 0
main_adhoccommandevent
↳  last minute 0
```

Obviously your mileage will vary depending on the actual size of your stdout and other parallel work, and things like database IOPS; these numbers represent "best case" performance in a single-node system that otherwise has no other load or CPU contention.

Turning on external logging generally adds a 20-30% cost due to the all of the string munging/formatting we do in Python, and the overhead of cpython's `logging` module for record generation, including the overhead our various log handlers and filters.  There might be another 5-10% optimization we could squeeze out of this if we really wanted to.

This could *probably* be made slightly faster by throwing more IOPs at a separate distinct database VM, but at this point we're coming up against the overhead of the Django ORM and its `bulk_create` implementation.  I suspect that re-writing the insertion in vanilla `psycopg2` and `copy_from` is one of the last remaining drastic improvements we could make to this model in Python.

awx/main/dispatch/worker/callback.py
```diff
@@ -26,7 +30,7 @@
 
 # the number of seconds to buffer events in memory before flushing
 # using JobEvent.objects.bulk_create()
-BUFFER_SECONDS = .1
+BUFFER_SECONDS = 1
```
chrismeyersfsu: put in `settings/defaults.py`

awx/main/dispatch/worker/callback.py
```diff
@@ -39,16 +43,51 @@ class CallbackBrokerWorker(BaseWorker):
...
+    total = 0
+    last_event = ''
     prof = None
 
     def __init__(self):
         self.buff = {}
+        self.pid = os.getpid()
+        self.redis = redis.Redis.from_url(settings.BROKER_URL)
+        for key in self.redis.keys('awx_callback_receiver_statistics_*'):
+            self.redis.delete(key)
```
chrismeyersfsu, line 52: could race other callback workers on init and cause a traceback and would then get re-spawned? Should at least try except this.

awx/main/dispatch/worker/callback.py
```diff
@@ -39,16 +43,51 @@ class CallbackBrokerWorker(BaseWorker):
...
+            logger.exception("encountered an error communicating with redis")
+            time.sleep(1)
+        except (json.JSONDecodeError, KeyError):
+            logger.exception("failed to decode JSON message from redis")
+        finally:
+            self.record_statistics()
+        return {'event': 'FLUSH'}
+
+    def record_statistics(self):
+        if time.time() - self.last_stats > 5:  # buffer stat recording to once per 5s
```
chrismeyersfsu: make the `5` a setting in `settings/defaults.py`

awx/main/dispatch/worker/callback.py
```diff
@@ -39,16 +43,51 @@ class CallbackBrokerWorker(BaseWorker):
...
+                logger.exception("encountered an error communicating with redis")
+                self.last_stats = time.time()
+
+    def debug(self):
+        return f'.  worker[pid:{self.pid}] sent={self.total} rss={self.mb}MB {self.last_event}'
+
+    @property
+    def mb(self):
+        return '{:0.3f}'.format(
+            psutil.Process(os.getpid()).memory_info().rss / 1024.0 / 1024.0
```
chrismeyersfsu: replace os.getpid() with self.pid.

awx/main/dispatch/worker/callback.py
chrismeyersfsu: This leaves me wanting more. Having a snapshot of the last job event processed is interesting but I suppose I want a log of all events processed. Not saying this is the place for that. But maybe this could be _more_ of a summary. Last 10-20 processed?

awx/main/dispatch/worker/callback.py
ryanpetrello: Key miss on delete don't raise a traceback:

```
In [4]: r.delete('foo')
Out[4]: 0
```

awx/main/dispatch/worker/callback.py
ryanpetrello: 👍

awx/main/dispatch/worker/callback.py
ryanpetrello: 👍

awx/main/dispatch/worker/callback.py
ryanpetrello: 👍

chrismeyersfsu (approved):

Issue #416, opened 2017-10-12
Audit AWX services that can control CA cert verification


jangsutsr:
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
 - API

##### SUMMARY
AWX backend is not only a REST service provider, but also a client to multiple third-party services, like log aggregators, notification backends, inventory sources and SSO providers. Ideally when AWX communicates with these services (via HTTPS in most of the cases), user should have the freedom to use custom CA certificates for verification, and to toggle off CA cert verification in general.

This issue is meant to be a tracker of all issues and pull requests that's related to providing CA cert verification toggle and customization for one or more third-party services, so that we can implement global verification toggle and customization Tower configurations when we have enough individual solutions.

This won't be an easy one-off work, as AWX rely on a non-trivial number of services, and individual implementation could be quite scattered around the code base. So any issues or pull requests that could enrich the list below is more than welcomed.

- [x] Logging handlers.
- [ ] Openstack inventory source: #404 #405.
- [ ] LDAP auth provider: #411.
- [ ] source control: #490

ghjm:
If we do this audit, what we should do is make sure that all Tower services use the system trust store and do not install private trust stores of their own.  The main difficulty with this is that if you use the Python Requests module from upstream, by default it uses its own trust store from certifi.  Requests as packaged on RHEL uses the RHEL trust store.  I _think_ recent Tower is already fully compliant with this approach, but auditing it couldn't hurt.

shanemcd:
All PR/Issues mentioned here are either closed or merged. Closing.

PR #8192, opened 2020-09-21
Override certifi in requests with the system cert store.

wenottingham:
There's not a good way to do this with pip, so just overwrite
the requests file after installing it.

Add a stub certifi library for things that call `certifi.where()` for their own use.


wenottingham:
For at least a part of https://github.com/ansible/awx/issues/416

ghjm:
I don't understand what's going on here with the virtualenvs.  Have we stopped using --system-site-packages?  Are we taking care of platform-only Python packages that aren't available via pip, like SELinux?  Or the plethora of platform-patched Python packages? Maybe this has all already been discussed and I just haven't been involved.

In any case, yes, if we're not using --system-site-packages, copying certs.py seems like a reasonable thing to do.

wenottingham:
We're not using system-site-packages, no.

That being said, if we really want this to affect everything, we probably need a fake certifi.where(), as other libs call that directly. Sigh.

PR #8193, opened 2020-09-21
add a few additional optimizations to the callback receiver

ryanpetrello:
cc: @gamuniz 

Many events, few tasks:
Kapture 2020-09-21 at 19 10 48

Many events, many tasks:
test

### DIFF
--- a/awx/conf/settings.py
+++ b/awx/conf/settings.py
@@ -17,6 +17,8 @@ from django.utils.functional import cached_property
 # Django REST Framework
 from rest_framework.fields import empty, SkipField
 
+import cachetools
+
 # Tower
 from awx.main.utils import encrypt_field, decrypt_field
 from awx.conf import settings_registry
@@ -27,6 +29,8 @@ from awx.conf.migrations._reencrypt import decrypt_field as old_decrypt_field
 # ready (or during migrations).
 
 logger = logging.getLogger('awx.conf.settings')
+
+SETTING_MEMORY_TTL = 5 if 'callback_receiver' in ' '.join(sys.argv) else 0
 
 # Store a special value to indicate when a setting is not set in the database.
 SETTING_CACHE_NOTSET = '___notset___'
@@ -406,6 +410,7 @@ class SettingsWrapper(UserSettingsHolder):
     def SETTINGS_MODULE(self):
         return self._get_default('SETTINGS_MODULE')
 
+    @cachetools.cached(cache=cachetools.TTLCache(maxsize=2048, ttl=SETTING_MEMORY_TTL))
     def __getattr__(self, name):
         value = empty
         if name in self.all_supported_settings:
--- a/awx/main/models/events.py
+++ b/awx/main/models/events.py
@@ -4,6 +4,7 @@ import datetime
 import logging
 from collections import defaultdict
 
+from django.conf import settings
 from django.db import models, DatabaseError, connection
 from django.utils.dateparse import parse_datetime
 from django.utils.text import Truncator
@@ -57,7 +58,18 @@ def create_host_status_counts(event_data):
     return dict(host_status_counts)
 
 
+MINIMAL_EVENTS = set([
+    'playbook_on_play_start', 'playbook_on_task_start',
+    'playbook_on_stats', 'EOF'
+])
+
+
 def emit_event_detail(event):
+    if (
+        settings.UI_LIVE_UPDATES_ENABLED is False and
+        event.event not in MINIMAL_EVENTS
+    ):
+        return
     cls = event.__class__
     relation = {
         JobEvent: 'job_id',
--- a/awx/main/models/events.py
+++ b/awx/main/models/events.py
@@ -380,10 +380,11 @@ class BasePlaybookEvent(CreatedModifiedModel):
             value = force_text(event_data.get(field, '')).strip()
             if value != getattr(self, field):
                 setattr(self, field, value)
-        analytics_logger.info(
-            'Event data saved.',
-            extra=dict(python_objects=dict(job_event=self))
-        )
+        if settings.LOG_AGGREGATOR_ENABLED:
+            analytics_logger.info(
+                'Event data saved.',
+                extra=dict(python_objects=dict(job_event=self))
+            )
 
     @classmethod
     def create_from_data(cls, **kwargs):
--- a/awx/ui/client/features/output/stream.service.js
+++ b/awx/ui/client/features/output/stream.service.js
@@ -119,6 +119,10 @@ function OutputStream ($q) {
         this.counters.ready = ready;
         this.counters.used = used;
         this.counters.missing = missing;
+
+        if (!window.liveUpdates) {
+            this.counters.ready = event.counter;
+        }
     };
 
     this.bufferEmpty = threshold => {
@@ -140,6 +144,10 @@ function OutputStream ($q) {
     this.isReadyToRender = () => {
         const { total } = this.counters;
         const readyCount = this.getReadyCount();
+
+        if (!window.liveUpdates) {
+            return true;
+        }
 
         if (readyCount <= 0) {
             return false;
--- a/awx/ui/test/unit/components/stream.unit.js
+++ b/awx/ui/test/unit/components/stream.unit.js
@@ -51,7 +51,13 @@ describe('Output | StreamService', () => {
     });
 
     describe('isReadyToRender', () => {
-        it("it's never ready to render unless the result of getReadyCount is greater than 0", () => {
+        it("it's never ready to render when live updates are enabled unless the result of getReadyCount is greater than 0", () => {
+            delete window.liveUpdates;
+            Object.defineProperty(window, 'liveUpdates', {
+                value: true,
+                writable: false
+            });
+
             const params = [
                 [-1, false],
                 [0, false],


awx/main/models/events.py
```diff
@@ -58,6 +59,11 @@ def create_host_status_counts(event_data):
 
 
 def emit_event_detail(event):
+    if (
+        settings.UI_LIVE_UPDATES_ENABLED is False and
+        event.event not in ('playbook_on_stats', 'EOF')
```
ryanpetrello: I'd like to maybe make this be `not startswith("runner_")`, but I'm having some trouble getting the UI to behave in a reasonable way cc @jakemcdermott is taking a look, I think.

wenottingham:
How hard is a placeholder in the text window in this case?

ryanpetrello:
@wenottingham see updated behavior in GIFs above.

ryanpetrello:
Here are some before/after numbers of this PR processing a large backlog of events on a single node `m5.4xlarge` install.  Both are using the default `settings.JOB_EVENT_WORKERS = 4`, and the external logger is disabled.

# before (devel)

```
[root@ip-10-0-4-131 ec2-user]# awx-manage callback_stats
main_jobevent
↳  last minute 123999

main_inventoryupdateevent
↳  last minute 0

main_projectupdateevent
↳  last minute 0

main_adhoccommandevent
↳  last minute 0
```

# after (`UI_LIVE_UPDATES_ENABLED` = `False`)

```
[root@ip-10-0-5-5 ec2-user]# awx-manage callback_stats
main_jobevent
↳  last minute 532999

main_inventoryupdateevent
↳  last minute 0

main_projectupdateevent
↳  last minute 0

main_adhoccommandevent
↳  last minute 0
```

ryanpetrello:
Here is the same comparison again, but with logging configured and pointed at a local UDP port via rsyslogd:

image

# before (devel)

```
[root@ip-10-0-4-131 ec2-user]# awx-manage callback_stats
main_jobevent
↳  last minute 91999

main_inventoryupdateevent
↳  last minute 0

main_projectupdateevent
↳  last minute 0

main_adhoccommandevent
↳  last minute 0
```

# after (UI_LIVE_UPDATES_ENABLED = False)

```
[root@ip-10-0-5-5 ec2-user]# awx-manage callback_stats
main_jobevent
↳  last minute 334999

main_inventoryupdateevent
↳  last minute 0

main_projectupdateevent
↳  last minute 0

main_adhoccommandevent
↳  last minute 0
```

ryanpetrello:
With `UI_LIVE_UPDATES_ENABLED = False` and `JOB_EVENT_WORKERS=16` on an `m5.4xlarge` (logging disabled), I'm able to break 1M inserts per minute:

```
[root@ip-10-0-5-5 ec2-user]# awx-manage callback_stats
main_jobevent
↳  last minute 1096363

main_inventoryupdateevent
↳  last minute 0

main_projectupdateevent
↳  last minute 0

main_adhoccommandevent
↳  last minute 0
```

awx/main/models/events.py
```diff
@@ -57,7 +58,18 @@ def create_host_status_counts(event_data):
...
 
+MINIMAL_EVENTS = set([
+    'playbook_on_play_start', 'playbook_on_task_start',
+    'playbook_on_stats', 'EOF'
+])
+
+
 def emit_event_detail(event):
+    if (
+        settings.UI_LIVE_UPDATES_ENABLED is False and
```
jakemcdermott, line 69: I'm wondering if we should add a new setting instead of reusing `UI_LIVE_UPDATES_ENABLED` so that the ui can handle them independently. 

cc @wenottingham  @ryanpetrello @mabashian  - thoughts? Are there situations where we might want to turn off job list updating without reducing job_event broadcast (or vice versa?)

ryanpetrello: Maybe so?  That said, the notion of asking a user to properly understand and toggle two things independently concerns me a little 😄.  One of my goals with this PR is for `UI_LIVE_UPDATES_ENABLED` to effectively serve as a one-click "low bandwidth mode" for users who want higher event throughput.

jakemcdermott: :thumbsup: sgtm

wenottingham: UI_LIVE_UPDATES_ENABLED was originally designed for the old list update scheme (where websocket triggered an API refresh) - correct?

I'd be more tempted to tweak it so it means what this PR does instead and still allow list refreshes with it.

ryanpetrello:
@moonrail @psuriset either of you have any time to give this patch a whirl in comparison with stable builds?  Specifically, I'm curious to see how event processing time improves.

awx/main/models/events.py
jakemcdermott: Yep - I think you both have a point. Keeping it as a single setting makes sense.

awx/main/models/events.py
```diff
@@ -368,10 +380,11 @@ def _send_notifications():
             value = force_text(event_data.get(field, '')).strip()
             if value != getattr(self, field):
                 setattr(self, field, value)
-        analytics_logger.info(
-            'Event data saved.',
-            extra=dict(python_objects=dict(job_event=self))
-        )
+        if settings.LOG_AGGREGATOR_ENABLED:
```
ryanpetrello, line 383: Our custom filter technically checks this flag, too, but it does so *after* making the logging record and looping through all the handlers (which is actually pretty expensive).  If logging is disabled, it's cheaper to just not call this `.info()` function at all.

awx/main/models/events.py
wenottingham: Note: that means we may want to *remove* the bit where UI_LIVE_UPDATES_ENABLED stops list updates. But that can be a second PR.

awx/conf/settings.py
```diff
@@ -406,6 +410,7 @@ def _get_default(self, name):
     def SETTINGS_MODULE(self):
         return self._get_default('SETTINGS_MODULE')
 
+    @cachetools.cached(cache=cachetools.TTLCache(maxsize=2048, ttl=SETTING_MEMORY_TTL))
```
ryanpetrello, line 413: In my profiling for high workloads, the cache hits for redis actually add up and end up being quite expensive in bulk;  even though it's just a round trip to a domain socket, it's still 10+ round trips per emitted event (because we access a handful of distinct settings to determine how to log) and that has a real cost when you multiply it by millions of events

For the purposes of the callback receiver, it's probably not imperative that we have real-time correctness for these setting values (caching for a few seconds locally in memory makes a notable performance improvement).

awx/main/models/events.py
ryanpetrello: Maybe so.  I suppose you could argue that at this point, we've probably mostly resolved the original problem that `UI_LIVE_UPDATES_ENABLED` was originally meant to paper over.

mabashian: > Maybe so. I suppose you could argue that at this point, we've probably mostly resolved the original problem that UI_LIVE_UPDATES_ENABLED was originally meant to paper over.

Yea that _should_ be the case at this point

mabashian (approved):
UI changes lgtm

awx/main/models/events.py
jakemcdermott: Ok - just chatted w/ @mabashian about this. I'm not entirely sure what the scope of work would be off the top of my head but from what we can see it may not be a risky change & something we can fit in. I'll take a closer look tomorrow.

ryanpetrello: Just to be clear, let's aim for this as a separate `devel` PR, though.

jakemcdermott: k, sounds good. Topic for a future PR and we'll track that work in this issue: https://github.com/ansible/awx/issues/8211

rebeccahhh (approved):

awx/main/models/events.py
```diff
@@ -57,7 +58,18 @@ def create_host_status_counts(event_data):
     return dict(host_status_counts)
 
 
+MINIMAL_EVENTS = set([
```
jbradberry, line 61: I would think that with a set this small that the hash calculation would overwhelm the big-O lookup advantage you gain by using a set instead of a tuple.

shanemcd (approved):

mcharanrm:
Hello @ryanpetrello, @kdelee CC: @psuriset, We deployed patch on 3 node cluster and all the 3 nodes are virtual machines (created using libvirt) with RHEL 7.8 installed. Each vm's configured with 4vcpus, 23G RAM, and 75G of shared disk space. We can see 33.8% increase in events processing rate after changing `UI_LIVE_UPDATES_ENABLED: false` and having higher event processing rate can explains less event lag between job run time and jobs events processing time. 
Please take a look at our [CPT graphs](http://kibana.perf.lab.eng.bos.redhat.com:5706/s/tower-cpt/app/kibana#/dashboard/eae2aab0-bd78-11ea-83da-d76f25b0f35c?_g=(refreshInterval:(pause:!t,value:0),time:(from:'2020-09-23T14:30:09.576Z',mode:absolute,to:'2020-09-24T09:19:09.576Z'))&_a=(description:'',filters:!(('$state':(store:appState),exists:(field:name.keyword),meta:(alias:!n,disabled:!f,index:'6488a250-a908-11ea-83da-d76f25b0f35c',key:name.keyword,negate:!f,type:exists,value:exists))),fullScreenMode:!f,options:(darkTheme:!f,hidePanelTitles:!f,useMargins:!t),panels:!((embeddableConfig:(),gridData:(h:15,i:'6',w:24,x:0,y:59),id:'4842b420-d7cd-11ea-83da-d76f25b0f35c',panelIndex:'6',type:visualization,version:'6.8.4'),(embeddableConfig:(),gridData:(h:15,i:'7',w:24,x:24,y:59),id:'1d93d7f0-d7cc-11ea-83da-d76f25b0f35c',panelIndex:'7',type:visualization,version:'6.8.4'),(embeddableConfig:(),gridData:(h:15,i:'8',w:24,x:0,y:74),id:f49e80a0-db01-11ea-83da-d76f25b0f35c,panelIndex:'8',type:visualization,version:'6.8.4'),(embeddableConfig:(),gridData:(h:15,i:'9',w:24,x:24,y:74),id:a21fba60-db01-11ea-83da-d76f25b0f35c,panelIndex:'9',type:visualization,version:'6.8.4'),(embeddableConfig:(),gridData:(h:15,i:'10',w:48,x:0,y:14),id:fad4a9c0-a98f-11ea-83da-d76f25b0f35c,panelIndex:'10',type:visualization,version:'6.8.4'),(embeddableConfig:(),gridData:(h:15,i:'11',w:24,x:24,y:29),id:bea6feb0-bfc1-11ea-83da-d76f25b0f35c,panelIndex:'11',type:visualization,version:'6.8.4'),(embeddableConfig:(),gridData:(h:15,i:'12',w:24,x:0,y:29),id:a0f781f0-bfc1-11ea-83da-d76f25b0f35c,panelIndex:'12',type:visualization,version:'6.8.4'),(embeddableConfig:(),gridData:(h:14,i:'13',w:48,x:0,y:0),id:'7231c7b0-dd3d-11ea-83da-d76f25b0f35c',panelIndex:'13',type:visualization,version:'6.8.4'),(embeddableConfig:(),gridData:(h:15,i:'14',w:48,x:0,y:44),id:c3635a00-f803-11ea-83da-d76f25b0f35c,panelIndex:'14',type:visualization,version:'6.8.4')),query:(language:lucene,query:''),timeRestore:!t,title:tower_cpt_dashboard,viewMode:edit)). 
Screenshot from 2020-09-24 16-36-44


Please let me know if there is anything else that i need to test it.

ryanpetrello:
This looks great - thank you @mcharanrm!

Issue #8190, opened 2020-09-21
Add access view for Execution Environments


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Add access view for Execution Environments.

See: https://tower-mockups.testing.ansible.com/patternfly/exec-env/exec-env-detail/

nixocio:
See: https://github.com/ansible/awx/pull/8198

PR #8197, opened 2020-09-22
Add access view for Execution Environments

nixocio:
Add access view for Execution Environments

See: https://github.com/ansible/awx/issues/8190

image


nixocio:
This data is not yet implement on the API side, but for the U addition it is the same boilerplate used in different resources.

nixocio:
I will rename this branch.

PR #8198, opened 2020-09-22
Add access view for Execution Environments

nixocio:
Add access view for Execution Environments

See: https://github.com/ansible/awx/issues/8190

image


nixocio:
@jakemcdermott, This data is not yet implement on the API side, but for the UI addition it is the same boilerplate used in different resources.

jlmitch5 (approved):

nixocio:
This PR will be probably closed. Based on the changes for RBAC for EE.

jakemcdermott:
@nixocio, @jbradberry, and I discussed this offline. It's possible we'll revisit this based on feedback but as of right now there doesn't seem to be much of a need for a dedicated access tab.

There will be an org-level admin role for EE's but EE's themselves won't have roles, which makes the permission scenarios relatively very simple.

A similar pattern exists for Notification Templates, presumably for similar reasons. There's a "Notification Admin" role assignable to users at the org level but Notification Templates themselves don't have roles and we do not show an access ui for Notification Templates.

Issue #8145, opened 2020-09-15
Database deadlock when running Facts job


jpell958:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
An AWX database deadlock occurs when facts jobs (package facts and gather facts) are running in slices with 4,000 targets.

##### ENVIRONMENT
* AWX version: 13.0
* AWX install method: developed Ansible playbook
* Ansible version:  2.9.11
* Operating System: RHEL 7
* Web Browser: Chromium Edge

##### STEPS TO REPRODUCE

The issue happens every time we run a playbook with this specific set of tasks:
```
---
- name: Collect Package Facts
 package_facts:
    manager: "auto"
  tags:
    - gather-facts

- name: Collect All Facts
  setup:
    gather_subset: all
  tags:
    - gather-facts
```
More details will be added to this case as they are gathered

##### EXPECTED RESULTS

- No deadlock when running facts job
- Successfully receive EOF event.

##### ACTUAL RESULTS

- Deadlock occurs when running facts job
- No EOF event received.

Example query used:
```
UPDATE "main_host" SET "last_job_id" = (CASE WHEN
("main_host"."id" = 8750) THEN 252213 WHEN
("main_host"."id" = 8543) THEN 252212 WHEN
("main_host"."id" = 2245) THEN 252211 WHEN
("main_host"."id" = 2246) THEN 252210 WHEN
("main_host"."id" = 3554) THEN 252268 WHEN
("main_host"."id" = 3553) THEN 252208 WHEN
("main_host"."id" = 3541) THEN 252207 WHEN
("main_host"."id" = 3542)
THEN 252269 WHEN ("main_host"."id" = 428) THEN 252205
WHEN ("main_host"."id" = 651) THEN 252204 WHEN
("main_host"."id" = 10030) THEN 252203 WHEN
("main_host"."id" = 511) THEN 252202 WHEN ("main_host
"."id" = 429) THEN 252213 WHEN ("main_host"."id" = 512)
THEN 252212 WHEN ("main_host"."id" = 294) THEN 252211
WHEN ("main_host"."id" = 295) THEN 252210 WHEN
("main_host"."id" = 296) THEN 252268 WHEN
("main_host"."id" = 335) THEN 252208 WHEN
("main_host"."id" = 430) THEN 252207 WHEN
("main_host"."id" = 539) THEN 252269 WHEN
("main_host"."id" = 297) THEN 252205 WHEN
("main_host"."id" = 720) THEN
[...]

PR #8206, opened 2020-09-22
change host -> last_job_id bulk update query to avoid locking issues

ryanpetrello:
see: https://github.com/ansible/awx/issues/8145

cc @jpell958

### DIFF
--- a/awx/main/models/events.py
+++ b/awx/main/models/events.py
@@ -520,13 +520,21 @@ class JobEvent(BasePlaybookEvent):
                 (summary['host_id'], summary['id'])
                 for summary in JobHostSummary.objects.filter(job_id=job.id).values('id', 'host_id')
             )
+            updated_hosts = set()
             for h in all_hosts:
                 # if the hostname *shows up* in the playbook_on_stats event
                 if h.name in hostnames:
                     h.last_job_id = job.id
+                    updated_hosts.add(h)
                 if h.id in host_mapping:
                     h.last_job_host_summary_id = host_mapping[h.id]
-            Host.objects.bulk_update(all_hosts, ['last_job_id', 'last_job_host_summary_id'])
+                    updated_hosts.add(h)
+
+            Host.objects.bulk_update(
+                list(updated_hosts),
+                ['last_job_id', 'last_job_host_summary_id'],
+                batch_size=100
+            )
 
 
     @property


ryanpetrello:
@AlanCoding I've made some more refinements to this query (see related issue) - can I get your thoughts?

awx/main/models/events.py
```diff
@@ -520,13 +520,22 @@ def _update_host_summary_from_stats(self, hostnames):
...
                     h.last_job_host_summary_id = host_mapping[h.id]
-            Host.objects.bulk_update(all_hosts, ['last_job_id', 'last_job_host_summary_id'])
+                    updated_hosts.add(h)
+
+            def _chunks(x, n):
+                for i in range(0, len(x), n):
+                    yield x[i:i + n]
+
+            for chunk in _chunks(list(updated_hosts), 100):
+                Host.objects.bulk_update(chunk, ['last_job_id', 'last_job_host_summary_id'])
```
AlanCoding: bulk_update takes in a `batch_size` kwarg

https://docs.djangoproject.com/en/2.2/ref/models/querysets/

shouldn't we prefer to use that instead of chunking it ourselves?

awx/main/models/events.py
```diff
@@ -520,13 +520,22 @@ def _update_host_summary_from_stats(self, hostnames):
...
             )
+            updated_hosts = set()
             for h in all_hosts:
                 # if the hostname *shows up* in the playbook_on_stats event
                 if h.name in hostnames:
                     h.last_job_id = job.id
+                    updated_hosts.add(h)
                 if h.id in host_mapping:
                     h.last_job_host_summary_id = host_mapping[h.id]
-            Host.objects.bulk_update(all_hosts, ['last_job_id', 'last_job_host_summary_id'])
```
ryanpetrello, line 529: One notable change is that the prior version (here) was updating `all_hosts`, even though many of those didn't actually _change_.  The new version tracks which hosts are actually *part* of this run, and only updates those.  I expect that this explains the behavior described in https://github.com/ansible/awx/issues/8145; in a scenario where you've got many parallel job slices all finishing around the same time, they're all competing with each other to update all of the same rows (even though each slice only represents some percentage of the total inventory).

awx/main/models/events.py
```diff
@@ -520,13 +520,22 @@ def _update_host_summary_from_stats(self, hostnames):
...
+                    updated_hosts.add(h)
+
+            def _chunks(x, n):
+                for i in range(0, len(x), n):
+                    yield x[i:i + n]
+
+            for chunk in _chunks(list(updated_hosts), 100):
```
ryanpetrello: The other change is that I decided to split these updates into more reasonable chunks; I don't want to give up the performance benefits by calling `host.save()` per-Host, but it's also true that the list of hosts could be really large.

awx/main/models/events.py
ryanpetrello: Oh!  I didn't know about this.  Thanks 😄

awx/main/models/events.py
```diff
@@ -520,13 +520,21 @@ def _update_host_summary_from_stats(self, hostnames):
...
+                    updated_hosts.add(h)
                 if h.id in host_mapping:
                     h.last_job_host_summary_id = host_mapping[h.id]
-            Host.objects.bulk_update(all_hosts, ['last_job_id', 'last_job_host_summary_id'])
+                    updated_hosts.add(h)
+
+            Host.objects.bulk_update(
+                list(updated_hosts),
+                ['last_job_id', 'last_job_host_summary_id'],
+                batch_size=100
```
ryanpetrello, line 536: The other change is that I decided to split these updates into more reasonable chunks; I don't want to give up the performance benefits by calling host.save() per-Host, but it's also true that the list of hosts could be really large.

awx/main/models/events.py
```diff
@@ -520,13 +520,21 @@ def _update_host_summary_from_stats(self, hostnames):
                 (summary['host_id'], summary['id'])
                 for summary in JobHostSummary.objects.filter(job_id=job.id).values('id', 'host_id')
             )
+            updated_hosts = set()
             for h in all_hosts:
```
AlanCoding, line 524: okay, so `host_map` is _not_ down-selected because it's sent over the wire as an artifact from the inventory generation. Is that right? Then `all_hosts` follows from that, and the only time we get what the limited hosts were are from the stats event, which comes through in `host_mapping`.

`hostnames` comes from...

https://github.com/ansible/awx/blob/de5f996358563eb900645d3620ace82e77451ad7/awx/main/models/events.py#L476-L483

Note that, as I write that, I see clear as day that its criteria is different from L504 above

```python
                for stat in ('changed', 'dark', 'failures', 'ignored', 'ok', 'processed', 'rescued', 'skipped'):
```

which is not what you came here to fix... but...

AlanCoding (approved):

awx/main/models/events.py
AlanCoding: Oh you know what, that might actually be https://github.com/ansible/ansible-runner/issues/196

awx/main/models/events.py
AlanCoding: I get the point of experimenting with `host.save()`, but for the purpose of finalizing this for merge, it would be a hard sell to use that as opposed to `bulk_update` with a batch size of, say, 10.

You're not in a transaction here, are you? If you are, I can see that undermining the goal of smaller batch sizes... but being in the callback receiver, I didn't think you were.

ryanpetrello: Yea, in my testing, distinct `.save()` calls just aren't gonna work (too slow for large inventories).

Issue #7402, opened 2020-06-22
Scheduled jobs that end with timeouts do not trigger a failed notification


hluaces:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

Scheduled jobs that end up with error because their running time is over the specified timeout threshold for said job are not triggering a "failed notification". This used to work in AWX 8.0.0.

##### ENVIRONMENT
* AWX version: 12.0.0
* AWX install method: docker on Linux
* Ansible version:  2.9.9
* Operating System: CentOS7
* Web Browser: N/A

##### STEPS TO REPRODUCE

1. Create some notifications in an AWX instance. I'm my case I'm trying with "email" and "Slack" notification types.

2. Create or modify an AWX job and set the following:

   * An "On error" notification.
   * A timeout. In order to reproduce the bug I recommend setting up an abnormally short timeout. In my case I'm trying with 5s.

3. Run the job and be sure to make it fail due to timeout.

4. Verify that no "failure notification" is triggered.

##### EXPECTED RESULTS

A "failure" notification should be triggered. This is the behavior that AWX 8.0.0 used to have.

##### ACTUAL RESULTS

No notification is triggered and the job fails silently.

##### ADDITIONAL INFORMATION

Screenshot of a failed job reporting a timeout error:

![image](https://user-images.githubusercontent.com/3204332/85262998-d4ee2300-b46e-11ea-9e9b-a9a8be067c50.png)

The notification configuration for the failed job:

![image](https://user-images.githubusercontent.com/3204332/85263134-0535c180-b46f-11ea-9178-b3fd7a4679e2.png)

The failing job configuration used to reproduce the issue. Please note that the low timeout is there just for demonstration purposes. The original failing job has a timeout of 3600s.

![image](https://user-images.githubusercontent.com/3204332/85263313-462dd600-b46f-11ea-84de-abb3ceac1de8.png)

masbahnana:
1) Created slack notification
2) Created Job template with a playbook that takes 30 seconds and a time out of 5 seconds (so it would fail with timeout)
3) Created schedule to launch job every minute
4) Associated notification w/ job and enabled it for failures (to get notification when jobs failed)
5) Waited and observed that job did lauch, failed, and I got notification

PR #8221, opened 2020-09-23
Enable "On Fail" Notifications to Send Upon Timeout of Job Templates

beeankha:
##### SUMMARY

Fixing the bug described in Issue https://github.com/ansible/awx/issues/7402

Job Templates with a timeout and "send on fail" notifications attached were not correctly sending a notification upon timeout.  This PR should fix that faulty behavior.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -1506,6 +1506,8 @@ class BaseTask(object):
                 self.instance.job_explanation = "Job terminated due to timeout"
                 status = 'failed'
                 extra_update_fields['job_explanation'] = self.instance.job_explanation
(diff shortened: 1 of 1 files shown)


beeankha:
I tested this locally with a JT that had 200 forks (just like in the example of the filed issue), 0 forks, `enable fact cache` toggled on and then off, scheduled, not scheduled, etc.

rebeccahhh (approved):
+1 (especially for the comment, I dig comments)

ryanpetrello (approved):

rh-dluong:
Hello, I'd like to request that the fix for this also be backported to Tower 3.8.

PR #8229, opened 2020-09-24
Make filter a bit more consistent accross UI

### LINKED ISSUE #7410: Add `description` to the default search options on relevant lists
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
~~Seems reasonable that a user may want to search by description on this list.  That would make the defaults:~~

~~Name~~
~~Description~~
~~Created By~~
~~Modified By~~

Repurposing this as I think there are a number of lists that could benefit from adding this option to the default search.

- Organizations list
- Teams list
- Notifications list

nixocio:
Add `description`, `created_by` and `modified_by` when those fields are
available.

See: https://github.com/ansible/awx/issues/7410

### DIFF
--- a/awx/ui_next/src/components/Schedule/ScheduleList/ScheduleList.jsx
+++ b/awx/ui_next/src/components/Schedule/ScheduleList/ScheduleList.jsx
@@ -140,6 +140,18 @@ function ScheduleList({
             key: 'name__icontains',
             isDefault: true,
           },
+          {
+            name: i18n._(t`Description`),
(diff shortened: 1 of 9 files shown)


jakemcdermott (approved):

dsesami (changes requested):
Notification templates have a description field and should be added as well.

Schedules (from both the main schedule view and the job template tab view) have data containing "created by" and "modified by" fields. this should be included.

dsesami (approved):

Issue #7751, opened 2020-07-29
Delete button tooltip missing username when user cannot be deleted


mabashian:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
<img width="1158" alt="Screen Shot 2020-07-29 at 8 35 41 AM" src="https://user-images.githubusercontent.com/9889020/88800722-9850ce00-d176-11ea-8650-eec148f2351e.png">

If you're logged in as a superuser and you select your own superuser account to be deleted the tooltip is missing the username.  IIRC we should be listing out all the resources that you don't have the ability to delete in that tooltip.

tiagodread:
PR already merged

unlikelyzero:
Re-opening to add coverage for https://github.com/ansible/awx/pull/8230

mabashian:
@unlikelyzero looks like this could be closed?

unlikelyzero:
needs additional tests added

unlikelyzero:
https://github.com/ansible/tower-qa/pull/5626

PR #8230, opened 2020-09-24
Add username to tooltip

nixocio:
Add username to tooltip when user cannot be deleted.

See: https://github.com/ansible/awx/issues/7751

Users:

image

Others:

image

### DIFF
--- a/awx/ui_next/src/components/PaginatedDataList/ToolbarDeleteButton.jsx
+++ b/awx/ui_next/src/components/PaginatedDataList/ToolbarDeleteButton.jsx
@@ -87,7 +87,7 @@ function ToolbarDeleteButton({
   const renderTooltip = () => {
     const itemsUnableToDelete = itemsToDelete
       .filter(cannotDelete)
-      .map(item => item.name)
+      .map(item => item.name || item.username)
       .join(', ');
     if (itemsToDelete.some(cannotDelete)) {
       return (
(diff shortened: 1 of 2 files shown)


unlikelyzero:
@nixocio what does this look like with 25 usernames with character length of 13?

nixocio:
@unlikelyzero, like this. 

image

I did not introduce any changes in how the tooltip is being displayed. Just added a missing field. The same behavior will be present in all lists.

awx/ui_next/src/components/PaginatedDataList/ToolbarDeleteButton.jsx
```diff
@@ -84,10 +84,15 @@ function ToolbarDeleteButton({
...
+  function verifyUsername(item) {
+    return item.username;
+  }
+
   const renderTooltip = () => {
+    const criteria = itemsToDelete.every(verifyUsername) ? 'username' : 'name';
     const itemsUnableToDelete = itemsToDelete
       .filter(cannotDelete)
-      .map(item => item.name)
+      .map(item => item[criteria])
```
jlmitch5: why not just:

.map(item => item.username || item.name) ?

jlmitch5 (changes requested):

awx/ui_next/src/components/PaginatedDataList/ToolbarDeleteButton.jsx
nixocio: @jlmitch5, updated per your suggestion. Thanks.

jlmitch5 (approved):
looks good!

tiagodread (approved):
👍🏽

Issue #8219, opened 2020-09-23
[Notification template] "name" field triggers unexpected GET requests


tiagodread:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

Trying to create a new notification template every char filled on the name field trigger a GET request for `/api/v2/organizations/`

note: this causes significant performance slowdown in the UI.

##### ENVIRONMENT
* AWX version: 2.9.11
* AWX install method:docker for mac
* Ansible version:  14.1.0
* Operating System: Catalina
* Web Browser: Chrome

##### STEPS TO REPRODUCE

1. Navigate to Notification template
2. Open the network tab on browser
3. Type something on the name field

##### EXPECTED RESULTS

No API request should be made

##### ACTUAL RESULTS

Get request for `/api/v2/organizations/?order_by=name&page=1&page_size=5` on every character filled 

##### ADDITIONAL INFORMATION

![notification](https://user-images.githubusercontent.com/12464858/94059348-a870dc80-fdb8-11ea-86fd-87f58742ef1b.gif)

tiagodread:
This problem is happening for each field for each notification type.

dsesami:
It appears that this also occurs when doing Container Groups > New > selecting a credential.

tiagodread:
Tests are OK, PR already merged.

PR #8232, opened 2020-09-24
Fix extra GET requests on Notif Template/Container Groups forms

mabashian:
##### SUMMARY
link #8219 

I also hooked up autopopulate on those two forms because that was what was causing the issue.  `onChange` functions that get passed to lookups with autopopulate functionality now have to be wrapped in `useCallback` to stop them from being redefined on every render.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI


jakemcdermott (approved):
The code changes look reasonable to me. If QE can confirm that we've resolved the issue in all of our forms then let's go ahead and merge this.

tiagodread (approved):
I ran notification templates tests and now everything is ok!

PR #6032, opened 2020-02-21
migrate event table primary keys from integer to bigint

ryanpetrello:
see: https://github.com/ansible/awx/issues/6010

tl;dr 
1.  Rename old event table (`_old_main_jobevent`), copy its schema to new table (`main_jobevent`) but `id` --> `bigint`, Django migration is **done**
2.  AWX starts, you can’t see any events/stdout because `main_jobevent` is empty
3.  Dispatcher startup code notices the existence of the old table (that has all the data), shovels it into the new table in chunks until the old table is empty
4.  Old (empty) table is deleted, so the dispatcher startup code no longer attempts data shoveling

PR #8236, opened 2020-09-25
refactor some callback receiver code

ryanpetrello:
the bigint migration (https://github.com/ansible/awx/pull/6032) removed the foreign key constraints for:

- host_id
- job_id (and projectupdate_id, etc...)

I'd like to say this was intentional, but I actually think is was an unexpected
side effect of my work that happened to work out in our favor.

because of this, we don't really need to check explicitly for a host_id
IntegrityError anymore (because it won't occur)

additionally, while it's possible to insert an event with a mismatched
job_id now (for example, you can totally start a long-running job, and
delete the job record in the background using the ORM or psql), doing
so results in DoesNotExist errors in the code that handles the
playbook_on_stats events

cc @gamuniz

### DIFF
--- a/awx/main/dispatch/worker/callback.py
+++ b/awx/main/dispatch/worker/callback.py
@@ -11,7 +11,7 @@ import traceback
 from django.conf import settings
 from django.utils.timezone import now as tz_now
 from django.db import DatabaseError, OperationalError, connection as django_connection
-from django.db.utils import InterfaceError, InternalError, IntegrityError
+from django.db.utils import InterfaceError, InternalError
 
 import psutil
 
@@ -120,20 +120,12 @@ class CallbackBrokerWorker(BaseWorker):
                     e.modified = now
                 try:
                     cls.objects.bulk_create(events)
-                except Exception as exc:
+                except Exception:
                     # if an exception occurs, we should re-attempt to save the
                     # events one-by-one, because something in the list is
-                    # broken/stale (e.g., an IntegrityError on a specific event)
+                    # broken/stale
                     for e in events:
                         try:
-                            if (
-                                isinstance(exc, IntegrityError) and
-                                getattr(e, 'host_id', '')
-                            ):
-                                # this is one potential IntegrityError we can
-                                # work around - if the host disappears before
-                                # the event can be processed
-                                e.host_id = None
                             e.save()
                         except Exception:
                             logger.exception('Database Error Saving Job Event')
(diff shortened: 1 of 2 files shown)


awx/main/dispatch/worker/callback.py
```diff
@@ -123,17 +123,9 @@ def flush(self, force=False):
                 except Exception as exc:
                     # if an exception occurs, we should re-attempt to save the
                     # events one-by-one, because something in the list is
-                    # broken/stale (e.g., an IntegrityError on a specific event)
+                    # broken/stale
                     for e in events:
                         try:
-                            if (
-                                isinstance(exc, IntegrityError) and
-                                getattr(e, 'host_id', '')
```
ryanpetrello, line 131: This foreign key constraint just doesn't exist anymore at all (go ahead and go check 😄 `\d main_jobevent`).

awx/main/models/events.py
```diff
@@ -349,31 +350,36 @@ def _update_from_event_data(self):
...
-                    job_id=self.job_id, uuid__in=failed
-                ).update(failed=True)
-
-                # send success/failure notifications when we've finished handling the playbook_on_stats event
-                from awx.main.tasks import handle_success_and_failure_notifications  # circular import
-
-                def _send_notifications():
-                    handle_success_and_failure_notifications.apply_async([self.job.id])
-                connection.on_commit(_send_notifications)
+                try:
```
ryanpetrello, line 353: The `job_id` constraint doesn't exist either, so it's totally possible to insert a `playbook_on_stats` for a job that no longer exists. When that's the case, this code will blow up.  This is probably a _little_ difficult to hit in practice, because the API actually prevents you from deleting a job that's still processing events.

awx/main/models/events.py
```diff
@@ -497,7 +503,11 @@ def _hostnames(self):
 
     def _update_host_summary_from_stats(self, hostnames):
         with ignore_inventory_computed_fields():
-            if not self.job or not self.job.inventory:
+            try:
```
ryanpetrello, line 506: This code will also blow up in the scenario described above.

ryanpetrello:
recheck

awx/main/models/events.py
chrismeyersfsu: "the API actually prevents you from deleting a job that's still processing events, until 2 minutes after the job is finished" But yeah, it's still hard to hit.

chrismeyersfsu (approved):

awx/main/models/events.py
ryanpetrello: Yea, it *can* happen.  Kind of hard to get here, but better to have this work properly 👍

Issue #7417, opened 2020-06-23
auto-logout warning is behind survey editor


tp199314:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
I was editing a survey of a job template for a long time with many modifications and additions. Sadly, AWX ran into some kind of timeout and when hitting "save" i got the login screen. 

##### ENVIRONMENT
* AWX version: 12.0.0
* AWX install method: docker on linux
* Ansible version:  2.9.9
* Operating System: Ubuntu (AWX Server) Windows 10 (Client)
* Web Browser: Chrome

##### STEPS TO REPRODUCE

1. Open job template
2. Hit "edit survey"
3. Edit something and wait a huge amount of time (or so)
4. Hit "save"

##### EXPECTED RESULTS

Would be nice to deactivate the timeout or to prevent dataloss in another way.

##### ACTUAL RESULTS

Got redirected to login screen - all added/edited data was gone.

##### ADDITIONAL INFORMATION

I noticed, that there were a auto-logout-warning the other day, maybe it is possible to show this in front of the survey-edit-screen. :)

wenottingham:
> I noticed, that there were a auto-logout-warning the other day, maybe it is possible to show this in front of the survey-edit-screen. :)

That's the bug. We'll look at this in the ongoing UI rewrite.

Issue #4745, opened 2019-09-13
invalid credentials warning appearing below session time out - session timeout warning persisting


rebeccahhh:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

After a session has timed out and when attempting to re-login if incorrect credentials are entered the warning regarding invalid username and password is appearing below the session timeout warning. Session timeout warning is persisting instead of being replaces, leaving the sign in with 2 warnings. 

##### ENVIRONMENT
* AWX version: 7.0.0
* AWX install method: docker on linux
* Ansible version:  2.10.0.dev0
* Operating System: Fedora 30
* Web Browser: Firefox

##### STEPS TO REPRODUCE

Log into AWX, ignore it until it times out, go to log back in but use bad credentials.

##### EXPECTED RESULTS

Output should replace the timeout warning with the invalid credentials warning as it does if you are signed out and when signing in use bad credentials:
![image](https://user-images.githubusercontent.com/11764497/64882277-c98de180-d62a-11e9-9e02-7d4e8671c89d.png)

##### ACTUAL RESULTS

Instead both warnings appear:
![image](https://user-images.githubusercontent.com/11764497/64881953-102f0c00-d62a-11e9-889f-c8ecd209857d.png)

wenottingham:
punting low-priority UI nits to ui-next

AlexSCorey:
@rebeccahhh is this an issue in the current ui? or in the new ui?

rebeccahhh:
@AlexSCorey current, I haven't validated if it is present or not in the new UI (I would assume it is not).

unlikelyzero:
will retest in https://github.com/ansible/awx/pull/8250

akus062381:
Manually verified this has been fixed

Issue #4205, opened 2018-12-03
Track session timeout and prompt user to continue session when it's about to expire


mabashian:
I'd be happy to discuss other UX designs to solve this problem but I do think we want to give the user the ability to extend their session when it's about to expire.  This would require the UI to do a couple of things:

1) We'll need to probably keep a timer running that gets updated with each api response to track how much time is left on a particular session.
2) When that timer gets down to some threshold (was 60 seconds in old awx) we prompt the user to extend their session.  
3) If the user does click to extend their session... ??? I'm actually not too sure how this currently works but I would assume a request needs to be made to the api which would inherently extend the session.
4) If the user does not extend their session, the UI should hit the logout endpoint(? maybe) and redirect the user to the login screen
5) In the past, we've shown a message to the user saying that their session expired.  If/when we have a generic message field on the pf login component we could put it there or we could just go to the login screen with no message

jlmitch5:
As part of https://github.com/ansible/awx-pf/pull/156 401 authentication is checked based on an api response returning a 401 (instead of checking the cookie).  This works, but it does cause the UI to show the route before loading the login page for a split second.  There are a couple ways we can handle this:

- moving back to checking the cookie
- block rendering of the UI until the first api response is returned

The thing we need to be careful of is we do want the message that the user has been logged out to be population by the root dialog provider (which is why I had to move away from the check of api.isAuthenticated called from index.jsx)

We will also need to make sure other routes are not able to be accessed when the user is logged out.  There is currently an issue where trying to access `/#/foo` redirects the user to dashboard instead of login.

mabashian:
Adding onto this issue something currently in awx-pf.  The `logged out` text on the login screen seems to be shown inconsistently.  I believe the expected behavior is that it should be shown after logout.  Currently it's being shown on load/refresh but not after logging out:

![logout](https://user-images.githubusercontent.com/9889020/56153186-8f3ced00-5f83-11e9-8b28-f2f0922753e5.gif)

@jlmitch5 is planning on addressing this as part of his work on this issue.

keithjgrant:
Some miscellaneous thoughts while discussing #7210 (copied from slack):

>  I think we need the app to go through the login/auth validation logic before <Route>s  get mounted at or near the top level of the component hierarchy. That would only execute when the page is first opened/refreshed, since all routing afterwards happens through react-router
> Somewhere in that same logic, we could add a setTimeout or whatever to periodically confirm with the API that the session is still active
[...]

PR #8250, opened 2020-09-28
Add session timeout support

jakemcdermott:
#### SUMMARY
Intercept HTTP(S) responses and store expiration time from headers in web storage. Drive expiration timers in app container across all tabs with browser storage events and accompanying react hooks integration. Show a warning with logout countdown and continue button when session is nearly expired.

- not addressed by this pr: Login page behavior (idle logout messages, etc.). That will be handled in a separate issue: https://github.com/ansible/awx/issues/8272

##### Local Storage Integration
The `useStorage` hook integrates with the browser's localStorage api. It accepts a localStorage key as its only argument and returns a state variable and setter function for that state variable. When local storage is shared, the hook enables bidirectional data transfer between tabs via an event listener that is registered with the Web Storage api. This means that updates to the state variable using the setter function on one tab will also update the state variable on any other tab using this hook with the same key and vice-versa.

```jsx
function useStorage(key) {
  const [storageVal, setStorageVal] = useState(
    window.localStorage.getItem(key)
  );
  window.addEventListener('storage', () => {
    const newVal = window.localStorage.getItem(key);
    if (newVal !== storageVal) {
      setStorageVal(newVal);
    }
  });
  const setValue = val => {
    window.localStorage.setItem(key, val);
    setStorageVal(val);
  };
  return [storageVal, setValue];
}
```

Sequence Diagram for useStorage

The `useStorage` hook currently lives in the `AppContainer` component. It can be relocated to a more general location should and if the need ever arise

##### Session Expiration
Session timeout state is communicated to the client in the HTTP(S) response headers. Every HTTP(S) response is intercepted to read the session expiration time before being passed into the rest of the application. A timeout date is computed from the intercepted HTTP(S) headers and pushed into local storage, where it can be read using standard Web Storage apis or other utilities, such as `useStorage`.

#### ADDITIONAL INFORMATION
session


keithjgrant:
addresses #4205

awx/ui_next/src/api/Base.js
```diff
@@ -10,6 +10,16 @@ const defaultHttp = axios.create({
   },
 });
 
+defaultHttp.interceptors.response.use(response => {
+  const timeout = response?.headers['session-timeout'];
+  if (timeout) {
+    const timeoutDate = new Date().getTime() + timeout * 1000;
+    window.localStorage.setItem('session-timeout', String(timeoutDate));
```
nixocio: Perhaps we could write like
```suggestion
    window.localStorage.setItem('awx-session-timeout', String(timeoutDate));
   
```
To avoid conflict with other keys in `localStorage`.

jakemcdermott: HTTP(s) activity from GUI interaction tends to be bursty. Someday, we _may_ want to debounce this `setItem` call so that it only happens on the last response.

We can easily optimize this later by wrapping the function call in a debounce utility. That said, I'd prefer to hold off on that until we actually know it's needed.

nixocio:
Neat Diagram.

Edit: Helpful as well!

awx/ui_next/src/components/AppContainer/AppContainer.jsx
```diff
@@ -34,6 +38,45 @@ const PageHeader = styled(PFPageHeader)`
...
+ *      storage value and vice-versa.
+ *   3. When local storage is shared across browser tabs, the data
+ *      binding is also shared across browser tabs. This means that
+ *      updates to the state variable using the setter function on
+ *      one tab will also update the state variable on any other tab
+ *      using this hook with the same key and vice-versa.
+ */
+function useStorage(key) {
+  const [storageVal, setStorageVal] = useState(
+    window.localStorage.getItem(key)
```
nixocio, line 71: If later on we need to improve performance. We can evaluate the usage of `useState Lazy initialization` here. See: https://reactjs.org/docs/hooks-reference.html#lazy-initial-state

jakemcdermott: :man_shrugging: Perhaps. Although I wouldn't expect a bottleneck to happen here as this is an initialization step and doesn't run _too_ frequently. 

To be clear, I don't expect performance issues to be of high concern w/ this PR. The reason I brought it in my last comment was to point out where in the code we'd probably want to look first if it starts to be a concern in the future.

awx/ui_next/src/api/Base.js
jakemcdermott: Done

jakemcdermott: I decided to debounce this function call. While I don't think it's a situation that users will encounter, I noticed some slowdown with our auto-refreshing dev server on firefox when many tabs were opened and this helped.

awx/ui_next/src/components/AppContainer/AppContainer.jsx
```diff
@@ -132,6 +222,28 @@ function AppContainer({ i18n, navRouteConfig = [], children }) {
...
+            {i18n._(t`Continue`)}
+          </Button>,
+          <Button key="logout" variant="secondary" onClick={handleLogout}>
+            {i18n._(t`Logout`)}
+          </Button>,
+        ]}
+      >
+        {i18n._(t`You will be logged out in`)}{' '}
+        {Number(Math.floor(timeRemaining / 1000))}{' '}
+        {i18n._(t`seconds due to inactivity.`)}
```
marshmalien: Could we wrap this entire message for translation instead of splitting it up into chunks?
```suggestion
        {i18n._(t`You will be logged out in ${Number(Math.floor(timeRemaining / 1000))} seconds due to inactivity.`)}

```

marshmalien (approved):
Thank you for writing detailed and thorough documentation! Found a couple issues and discussed them with @jakemcdermott. One issue was due to the API allowing a delay of ~950 years, but JavaScript timers (setInterval and setTimeout) allow a max delay of `2147483647` seconds. 

The fixes look good to me! I only have one nitpick but otherwise 👍

awx/ui_next/src/components/AppContainer/AppContainer.jsx
jakemcdermott: Good idea. Fixed!

unlikelyzero:
@jakemcdermott we won't be adding automated coverage for session timeouts. I would like to put this into needs_test to 🔨  on it for an hour. Let me know when it's in a state that's appropriate -- even it if is just a needs_test ticket post merge

jakemcdermott:
@unlikelyzero Sounds good. I intend to move this over to needs test in the next day or so. I'll add the tag and ping.

awx/ui_next/src/components/AppContainer/AppContainer.jsx
```diff
@@ -13,13 +14,26 @@ import styled from 'styled-components';
...
+// The maximum supported timeout for setTimeout(), in milliseconds,
+// is the highest number you can represent as a signed 32bit
+// integer (approximately 25 days)
+const MAX_TIMEOUT = 2 ** (32 - 1) - 1;
+
+// The number of seconds the session timeout warning is displayed
+// before the user is logged out. Increasing this number (up to
+// the total session time, which is 1800s by default) will cause
+// the session timeout warning to display sooner.
+const SESSION_WARNING_DURATION = 10;
```
jakemcdermott, line 35:
```suggestion
const SESSION_WARNING_DURATION = 60;
```

unlikelyzero:
Found some issues with the e2e tests. Going to iron those out before merge

unlikelyzero:
@jakemcdermott hey, did you want to rerun yolok?

unlikelyzero:
May need to add `cy.reload` and network captures to `AfterEach`

tiagodread:
Added [stubs](https://github.com/ansible/tower-qa/pull/5743/commits/c2c39dd2354c5ef694abc908734f98196df531fd) for session time out in 5743

unlikelyzero:
note: retest this https://github.com/ansible/awx/issues/4745 and this https://github.com/ansible/awx/issues/7417

Issue #8202, opened 2020-09-22
Launching job with vault password 400s when default machine cred is replaced


mabashian:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Really odd behavior that I came across while building out credential prompting in the new UI.  This can be reproduced via the API browser though.

I created a JT with a default machine credential (that prompted for some passwords) and a default vault credential (that prompted for password).

If I swap the machine credential for one that does not prompt for a password then the API responds with a 400 on launch complaining about the vault password: `{"passwords_needed_to_start":["vault_password.A"]}`

In the example below credential with ID = 1 is the machine credential that prompts for ssh and become passwords.  Credential ID = 2 is the machine credential that does not prompt for passwords.  Credential ID = 4 is the vault credential that prompts for password.

A normal launch without changing credentials:

![good_launch](https://user-images.githubusercontent.com/9889020/93922608-26ad8000-fce0-11ea-8b61-8604ec329d83.gif)

A launch where I change the machine credential:

![bad_launch](https://user-images.githubusercontent.com/9889020/93922684-493f9900-fce0-11ea-93a5-891b46041a13.gif)

one-t:
Confirmed that this is now working as expected in devel.

PR #8251, opened 2020-09-28
fix a bug that can break password prompting in certain scenarios

ryanpetrello:
see: https://github.com/ansible/awx/issues/8202

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -4157,7 +4157,10 @@ class JobLaunchSerializer(BaseSerializer):
         # verify that credentials (either provided or existing) don't
         # require launch-time passwords that have not been provided
         if 'credentials' in accepted:
-            launch_credentials = accepted['credentials']
+            launch_credentials = Credential.unique_dict(
+                list(template_credentials.all()) +
+                list(accepted['credentials'])
+            ).values()
         else:
             launch_credentials = template_credentials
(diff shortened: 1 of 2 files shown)


awx/api/serializers.py
```diff
@@ -4157,7 +4157,10 @@ def validate(self, attrs):
         # verify that credentials (either provided or existing) don't
         # require launch-time passwords that have not been provided
         if 'credentials' in accepted:
-            launch_credentials = accepted['credentials']
+            launch_credentials = (
+                list(accepted['credentials']) +
```
ryanpetrello: @AlanCoding this `accepted` list that comes back from `template._accept_or_ignore_job_kwargs` *only* includes _net new_ credentials - not ones that **already exist** on the JT prior to launch, so we end up looking at the wrong list of materialized credentials to determine _which credentials need passwords_.  See tests below.

ryanpetrello:
@mabashian you mind giving this a whirl?

awx/api/serializers.py
AlanCoding: This might give duplicates inside of `launch_credentials`, but from reading the code after this carefully, I can't see any harm which would come from that.

AlanCoding (approved):

AlanCoding (approved):

awx/api/serializers.py
ryanpetrello: I changed this slightly to address this.

PR #8253, opened 2020-09-28
Fix Approval Node Edit Permissions

beeankha:
##### SUMMARY

Currently, users with the proper org/wfjt admin permissions are unable to edit approval nodes in workflow job templates that they should have edit permissions for.  This PR should fix that bug.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/access.py
+++ b/awx/main/access.py
@@ -2750,6 +2750,9 @@ class WorkflowApprovalTemplateAccess(BaseAccess):
         else:
             return (self.check_related('workflow_approval_template', UnifiedJobTemplate, role_field='admin_role'))
 
+    def can_change(self, obj, data):
+        return self.user.can_access(WorkflowJobTemplate, 'change', obj.workflow_job_template, data={})
+
     def can_start(self, obj, validate_license=False):
         # for copying WFJTs that contain approval nodes
         if self.user.is_superuser:


ryanpetrello (approved):
Looks reasonable to me, but I helped write it, so I'd say let's see what @AlanCoding thinks.

awx/main/access.py
```diff
@@ -2750,6 +2750,9 @@ def can_add(self, data):
         else:
             return (self.check_related('workflow_approval_template', UnifiedJobTemplate, role_field='admin_role'))
 
+    def can_change(self, obj, data):
+        return self.user.can_access(WorkflowJobTemplate, 'change', obj.workflow_job_template, data=data)
```
AlanCoding: Instead of passing `data=data`, I would much prefer `data={}`. The data you get contain fields for modification of the workflow approval template, not fields for the WFJT, which is how it would be interpreted in this call.

I'm also happy to take your word that the approval node's WFJT cannot be modified and cannot be null, so that this is not complicated further.

ryanpetrello: Agreed, `data={}` *is* probably what we want here.

ryanpetrello (approved):

Issue #8262, opened 2020-09-29
Add Support for Inventory Related Group Disassociate functionality.


AlexSCorey:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
See Inventory -> Group -> Host -> Disassociate for an example
Blocked by #8238

tiagodread:
While verifying this https://github.com/ansible/awx/pull/8263 I found this:

The select Groups modal has an advance search that enables the kebab button but there’s nothing inside it:

![modal](https://user-images.githubusercontent.com/12464858/96641682-755c3300-12fb-11eb-9078-50e24c6928f6.gif)

unlikelyzero:
We need to ensure we have coverage for this one

unlikelyzero:
Quite a few things were discovered in testing (5 bugs?). Whoever picks up this ticket, please ensure that we have stubs for the bugs and at least one test here.

akus062381:
@unlikelyzero we are covered here. See inventory-group-tabs.spec.js.
I also verified good locators. Closing this issue

Issue #8261, opened 2020-09-29
Support Add Functionality For Inventory Related Group


AlexSCorey:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
See Inventory -> Group -> Host -> Add
blocked by #8238

tiagodread:
PR already merged tests covered by 5683

Issue #8238, opened 2020-09-25
Add Inventory Related Groups List


AlexSCorey:
##### ISSUE TYPE
 - Feature

##### SUMMARY
Add Inventory -> Groups -> Related Groups

Add the list and tool bar

tiagodread:
Tests covered by:
```
/inventoriesinventory-adhoc-commands.spec.js
/inventories/inventory-group-tabs.spec.js
/user-journeys/auditor-journey.spec.js
```

PR #8263, opened 2020-09-29
8238 add related groups

AlexSCorey:
##### SUMMARY
This addresses https://github.com/ansible/awx/issues/8238.  It does not add functionality for Adding new or existing inventory or Disassociating an inventory.  

Part of the work in this PR includes refactoring the Add Drop Down button that is used by this list and, by the Inventory Group Host List so that both lists can use that component.  It also includes refactoring so that those tool bar action items appear properly in the drop down during Advanced Search.  

It does render the modals necessary to add and disassociate list items, but there is no persistence.  See https://github.com/ansible/awx/issues/8262 and https://github.com/ansible/awx/issues/8261

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
Screen Shot 2020-09-29 at 11 53 34 AM
Screen Shot 2020-09-29 at 11 54 17 AM

### DIFF
--- a/awx/ui_next/src/screens/Inventory/InventoryRelatedGroups/InventoryRelatedGroupList.jsx
+++ b/awx/ui_next/src/screens/Inventory/InventoryRelatedGroups/InventoryRelatedGroupList.jsx
@@ -1,12 +1,6 @@
 import React, { useCallback, useEffect, useState } from 'react';
 import { withI18n } from '@lingui/react';
 import { t } from '@lingui/macro';
-import {
-  Button,
-  Tooltip,
-  DropdownItem,
-  ToolbarItem,
-} from '@patternfly/react-core';
 import { useParams, useLocation, useHistory } from 'react-router-dom';
 
 import { GroupsAPI, InventoriesAPI } from '../../../api';
@@ -18,7 +12,6 @@ import DataListToolbar from '../../../components/DataListToolbar';
 import PaginatedDataList from '../../../components/PaginatedDataList';
 import InventoryGroupRelatedGroupListItem from './InventoryRelatedGroupListItem';
 import AddDropdown from '../shared/AddDropdown';
-import { Kebabified } from '../../../contexts/Kebabified';
 import AdHocCommands from '../../../components/AdHocCommands/AdHocCommands';
 import AssociateModal from '../../../components/AssociateModal';
 import DisassociateButton from '../../../components/DisassociateButton';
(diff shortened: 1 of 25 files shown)


awx/ui_next/src/screens/Inventory/InventoryRelatedGroups/InventoryRelatedGroupList.jsx
```diff
@@ -0,0 +1,255 @@
...
+                        >
+                          {i18n._(t`Run command`)}
+                        </DropdownItem>
+                      )}
+                    </AdHocCommandsButton>
+                  ) : (
+                    <ToolbarItem>
+                      <Tooltip
+                        content={i18n._(
+                          t`Select an inventory source by clicking the check box beside it. The inventory source can be a single host or a selection of multiple hosts.`
```
jakemcdermott: I wish our code formatter would automatically handle these strings for us, but it doesn't : \  Would you please format this for (source code) readability?

I'm not picky when it comes to how you want to do it. Just make sure the entire string can be read in a code review window / reasonably sized monitor without wrapping or side scrolling. < 100 or < 120 chars is a reasonable number to aim for.

jakemcdermott: For example, 

I reviewed this community pr the other day (proof-reading for typos and ..string escape errors ;) )


lines

awx/ui_next/src/screens/Inventory/InventoryRelatedGroups/InventoryRelatedGroupList.jsx
```diff
@@ -0,0 +1,255 @@
...
+            detailUrl={`/inventories/inventory/${inventoryId}/groups/${o.id}/details`}
+            editUrl={`/inventories/inventory/${inventoryId}/groups/${o.id}/edit`}
+            isSelected={selected.some(row => row.id === o.id)}
+            onSelect={() => handleSelect(o)}
+          />
+        )}
+        emptyStateControls={
+          canAdd && (
+            <AddDropdown
+              key="associate"
```
jakemcdermott: Does this need a key? Generally I see those used for lists of elements, where each one needs to be unique.

ditto https://github.com/ansible/awx/pull/8263/files#diff-876539ff73accea8ed0180b306a637c3R210 and anywhere else not mentioned

awx/ui_next/src/screens/Inventory/shared/AddDropdown.jsx
```diff
@@ -0,0 +1,88 @@
...
+  const element = useRef(null);
+
+  useEffect(() => {
+    const toggle = e => {
+      if (!isKebabified && (!element || !element.current.contains(e.target))) {
+        setIsOpen(false);
+      }
+    };
+
+    document.addEventListener('click', toggle, false);
```
jakemcdermott, line 25: Is there some sort of `onClick` prop or handler that we can use instead? I _usually_ don't expect to see click listeners in react components.

jakemcdermott: update: It's for making sure the dropdown closes when you click out of it. :+1: wfm

unlikelyzero:
@dsesami to add happy path, regular stubs, and a net new stub for RBAC
@tiagodread has an example of the RBAC tests with kebab

jakemcdermott (approved):

dsesami (dismissed):
normal half-screen-size widths seem to make this list go a bit wonky

Screenshot from 2020-10-06 15-51-58

dsesami:
adding/disassociating existing groups doesn't seem to work. is this implemented?

also -- would we want a way to show the parent group/root group on this page?

AlexSCorey:
> adding/disassociating existing groups doesn't seem to work. is this implemented?
> 
> also -- would we want a way to show the parent group/root group on this page?

Add/disassociate do not work here.  @trahman73 can you comment on parent group/root group on this page?

> normal half-screen-size widths seem to make this list go a bit wonky
> 
> Screenshot from 2020-10-06 15-51-58

The weird wrapping is a known PF issue.  There have been discussions about ways to fix it, but we have opted to allow PF to fix it rather than fixing it ourselves and then bring in their change.

jakemcdermott:
* The latest changes are to bring in some of the pre-existing patterns for handling outside-modal clicks from within the kebab menu and those lgtm :+1: 

* Regarding the [wrapping issue](https://github.com/ansible/awx/pull/8263#pullrequestreview-503308017) @dsesami mentioned, @AlexSCorey and I looked into this more and found we were able to resolve it by changing the directional params of the dropdown component

* We want to look into combining `screens/Inventory/shared/AddDropdown` and `components/AddDropdownButton` but will do that with a separate issue and pr after this one is merged. We'll track that work in this issue: https://github.com/ansible/awx/issues/8361

* Re: root groups, etc. (cc @dsesami) We may want to add that in as a later enhancement but it's not within the scope of the original issue. This isn't something we show on the group->related groups tab in the old ui, either, so I'd prefer it if we'd card that up as a possible future enhancement for a later time. It needs some additional discussion & ux considerations for sure. 

Screenshot from 2020-10-09 11-49-06

jakemcdermott (approved):

jakemcdermott:
regate

tiagodread:
Hey @AlexSCorey found some items:

- The advanced search results don’t match 

kebab

- The order feature on toolbar looks like doesn’t work 

order

- Trying to run a command using the variables `Foo:Bar` (without spacing) the modal doesn’t show that there’s an error on that field and allow the user execute but it returns an error:

command

- We have some a11y items found using axe [chrome extension](https://chrome.google.com/webstore/detail/axe-web-accessibility-tes/lhdoppojpmngadmnindnejefpokejbdd)

Steps to reproduce:
1. Install this chrome extension
2. Inspect the page 
3. Go to the axe tab
4. Click on analyse button

tiagodread (approved):
👍🏽

AlexSCorey:
regate

Issue #8252, opened 2020-09-28
JT tabs changes the items every time that you click on a tab


tiagodread:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

JT tabs items changes the position every time that you click on a tab

##### ENVIRONMENT
* AWX version: 14.1.0
* AWX install method: docker for mac
* Ansible version:  2.9.11
* Operating System: Catalina
* Web Browser: Chrome

##### STEPS TO REPRODUCE

1. Navigate to Job templates
2. Navigate through the available tabs

##### EXPECTED RESULTS

The tab items should be static and don’t change the position every time

##### ACTUAL RESULTS

The items changed the position 

##### ADDITIONAL INFORMATION

This behaviour broken a test case `onjob-template-tabs.spec.js`

![flick](https://user-images.githubusercontent.com/12464858/94482787-36c9d180-01b0-11eb-8bf8-bdd68a0e0870.gif)

tiagodread:
PR already merged 👍🏽

PR #8267, opened 2020-09-29
Reset error/result only after the next request has resolved to prevent render flickering

mabashian:
##### SUMMARY
link #https://github.com/ansible/awx/issues/8252

The reason why this hook was modified in the first place was because a user can press the TEST button over and over again on several credential plugin views without re-initializing the hook.  Once an error was set, it was never cleared and would always be displayed.  I've modified the changes to be a bit safer now and should mitigate the mentioned issue.  I've tested the credential plugin TEST buttons and they still function as expected.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/util/useRequest.js
+++ b/awx/ui_next/src/util/useRequest.js
@@ -38,18 +38,16 @@ export default function useRequest(makeRequest, initialValue) {
     request: useCallback(
       async (...args) => {
         setIsLoading(true);
-        if (isMounted.current) {
-          setResult(initialValue);
-          setError(null);
-        }
         try {
           const response = await makeRequest(...args);
           if (isMounted.current) {
             setResult(response);
+            setError(null);
           }
         } catch (err) {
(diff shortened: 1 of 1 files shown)


awx/ui_next/src/util/useRequest.js
```diff
@@ -38,18 +38,16 @@ export default function useRequest(makeRequest, initialValue) {
...
         try {
           const response = await makeRequest(...args);
           if (isMounted.current) {
             setResult(response);
+            setError(null);
           }
         } catch (err) {
           if (isMounted.current) {
             setError(err);
+            setResult(initialValue);
```
marshmalien, line 50: Nice 👍

marshmalien (approved):

tiagodread (approved):
Tested using UI broken test and now is everything ok 👍🏽

PR #8302, opened 2020-10-01
Fix Show/Hide i18n buttons

koleo:
##### SUMMARY

When i18n is enabled, dynamic Show/Hide buttons in INPUTs are translated. For example in French language SHOW/HIDE action is displayed as AFFICHER/MASQUER.

When the user click on one of these buttons, the button label changes dynamically (SHOW becomes HIDE, and vice-versa).

The button label was properly translated the first time the page was displayed, but when the label changed after clicking the button, it went back to the English label.

- Fixing translation for the dynamic label change
- Using an alternate method to detect the state: it's better to check the input type (text or password) rather than relying on the label which can vary from one language to another.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

##### AWX VERSION

```
awx: 14.1.0
```

##### ADDITIONAL INFORMATION

You can reproduce the bug or test the fix by:

- set up your browser to "French language" or any other language available in AWX
- Go to the "Add user" form in the UI (https://.../#/users/add) and click on "SHOW" or equivalent (this should appear as "AFFICHER" in French)

### DIFF
--- a/awx/ui/client/src/shared/directives.js
+++ b/awx/ui/client/src/shared/directives.js
@@ -1207,18 +1207,19 @@ function(SettingsUtils, i18n, $rootScope) {
     };
 }])
 
-.directive('awPasswordToggle', [function() {
+.directive('awPasswordToggle', ['i18n',
+function(i18n) {
     return {
         restrict: 'A',
         link: function(scope, element) {
             $(element).click(function() {
-                var buttonInnerHTML = $(element).html();
-                if (buttonInnerHTML.indexOf("Show") > -1) {
-                    $(element).html("Hide");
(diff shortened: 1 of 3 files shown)


unlikelyzero:
@koleo we've moved to a new UI and this implementation will need to be adjusted. Do you mind re-applying this fix to the new ui_next src?

koleo:
Hi, sure, I'll check that soon. Thanks.

ryanpetrello:
Hey @koleo,

What @unlikelyzero said - the UI has been entirely replaced in newer versions of awx, so this code is quite out of date.  If you find that this is still a necessary improvement in the new UI, please feel free to open a new PR.

Thanks!

Issue #8127, opened 2020-09-10
awx.awx collection modules should check for compatible version of awxkit


snooyen:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Caused by #8108 
@john-westcott-iv 

awx.awx.tower_* modules should:

1. Fail with error if `awxkit` is not installed
2. Fail with error if awxkit does not include export/import functionality

##### ENVIRONMENT
* AWX version: 14.1.0
* AWX install method: K8s
* Ansible version:  2.9.11
* Operating System: Mac OSX
* Web Browser: Chrome

##### STEPS TO REPRODUCE

1. Run `awx.awx.tower_export` in a Python environment that does not have `awxkit` installed.

##### EXPECTED RESULTS
Task should fail due to missing `awxkit` dependency

##### ACTUAL RESULTS
Task fails with incorrect error message.  See #8108 for details.

squidboylan:
This has been fixed by https://github.com/ansible/awx/pull/8308 which I have confirmed with manual testing

PR #8308, opened 2020-10-01
correct error message when calling tower_AWXKit collections module and venv doesn't have it

rebeccahhh:
#### SUMMARY

relates to #8127 -  The reason this issue arose in the first place is due to attempting to use an Ansible module (exit_json) when the correct one was instead fail_json. This triggered a different error message entirely, obfuscating from the venv's lack of correct set up. The logic of alerting if and when AWXKit was not present in a venv calling for it was already present within the module and now can be properly utilized.
Specifics:
- code was originally supposed to be hitting [this](https://github.com/ansible/ansible/blob/5cd489af061454565e036dd580d25e41487fbfcd/lib/ansible/module_utils/basic.py#L646) module, but instead was hitting [this](https://github.com/ansible/ansible/blob/5cd489af061454565e036dd580d25e41487fbfcd/lib/ansible/module_utils/basic.py#L1563) (and correctly so) due to it attempting to use unsupported parameters within the exit_json ansible module.
- pertinent ansible documentation regarding exit_json vs. fail_json [here](https://docs.ansible.com/ansible/latest/reference_appendices/module_utils.html)

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API
- Collections

##### AWX VERSION

```
awx: 14.1.0
```

##### ADDITIONAL INFORMATION

prior to change the error spit out would have been:
```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Unsupported parameters for (tower_export) module: inventory Supported parameters include: all, tower_config_file, tower_host, tower_oauthtoken, tower_password, tower_username, validate_certs"}
```
post change it should be: 
```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to import the required Python library (awxkit) on localhost.localdomain's Python /<path to venv>/bin/python3. Please read the module documentation and install it in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter"}
```

### DIFF
--- a/awx_collection/plugins/module_utils/tower_awxkit.py
+++ b/awx_collection/plugins/module_utils/tower_awxkit.py
@@ -24,7 +24,7 @@ class TowerAWXKitModule(TowerModule):
 
         # Die if we don't have AWX_KIT installed
         if not HAS_AWX_KIT:
-            self.exit_json(msg=missing_required_lib('awxkit'))
+            self.fail_json(msg=missing_required_lib('awxkit'))
 
         # Establish our conneciton object
         self.connection = Connection(self.host, verify=self.verify_ssl)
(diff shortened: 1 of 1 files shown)


beeankha (approved):
Nice catch! 👍

AlanCoding (approved):

squidboylan (approved):
WFM!

Issue #4540, opened 2019-08-21
Fails to detect roles/requirements.yaml


Aethylred:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
When a project includes a `roles/requirements.yaml` file it is not detected or processed:
```
TASK [detect requirements.yml] *************************************************
ok: [localhost, -> localhost] => {"changed": false, "stat": {"exists": false}}
```
It looks like the task to detect the file looks specifically for `requirements.yml` and not also `requirements.yaml`.

This is contrary to the YAML offical extension: https://yaml.org/faq.html

(Generally because both `.yaml` and `.yml` are in common usage, they should be considered interchangeable as different people like different things for their own reasons. I'd expect AWX/Tower to check and use a file with either extension, and throwing an error if it finds both)

##### ENVIRONMENT
* AWX version: 3.5.2
* AWX install method: Tower
* Ansible version:  2.8
* Operating System: CentOS 7
* Web Browser: Chrome

##### STEPS TO REPRODUCE

Create an AWX/Tower role and specify requirements in `roles/requirements.yaml` and clone into a project

rename to requirements.yml and it works.

##### EXPECTED RESULTS

I expected that the roles specified in `requirements.yaml` would be installed

##### ACTUAL RESULTS

The file `requirements.yaml` was ignored and roles were not installed.

##### ADDITIONAL INFORMATION

We use Ansible Tower, but I think AWX has this issue too.

ryanpetrello:
Agreed, it sounds like Ansible itself supports both:

https://docs.ansible.com/ansible/latest/reference_appendices/galaxy.html

> Beginning with Ansible 1.8 it is possible to install multiple roles by including the roles in a requirements.yml file. The format of the file is YAML, and the file extension must be either .yml or .yaml.

ryanpetrello:
@AlanCoding @chrismeyersfsu we should probably verify that this also works properly for new collections support in https://github.com/ansible/awx/pull/4265

https://github.com/ansible/awx/pull/4265/files#diff-f069d8ac794ad13606442d2194490282R141

AlanCoding:
Yes, it needs to be addressed in a number of places.

```
(awx) bash-4.2$ grep -r "requirements\.yml" awx/main/
awx/main/tasks.py:            galaxy_req_path = os.path.join(project_path, 'roles', 'requirements.yml')
[...]

Issue #6270, opened 2020-03-12
Support unified requirements.yml file roles/collections


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
See: https://github.com/ansible/ansible/pull/67843

Right now, we install files

```
roles/requirements.yml
collections/requirements.yml
```

The first will install with `ansible-galaxy install`, the second will install with `ansible-galaxy collection install`.

That PR will allow installing _both types_ with `ansible-galaxy install`.

Assuming this is the final form that merges (there seems little doubt), none of this will make a lot of sense. It doesn't make sense to allow putting collection requirements in the roles folder.

Tentative idea of how to react to this:

 - Depending on Ansible version, switch to `ansible-galaxy role install roles/requirements.yml`
 - Look for a unified requirements.yml file in a 3rd, new, yet-to-be-decided location

geerlingguy:
For a unified file, why not `requirements.yml` in the root of the project folder?

wenottingham:
Because that's not where anyone has put it before. I would not want to invent a new thing, especially since download-at-execution is a pardigm we'd like to move away from.

geerlingguy:
> Because that's not where anyone has put it before.

That's where I've always put it—for any kind of project where you have to manage dependencies ([node.js `package.json`](https://nodejs.org/en/knowledge/getting-started/npm/what-is-the-file-package-json/), [php `composer.json`](https://getcomposer.org/doc/01-basic-usage.md#composer-json-project-setup), [ruby `Gemfile`](https://bundler.io/gemfile.html), [python pip `requirements.txt`](https://pip.readthedocs.io/en/1.1/requirements.html), [rust `cargo.toml`](https://doc.rust-lang.org/cargo/guide/dependencies.html), etc.), the standard seems to be "put the list of dependencies/requirements in a file in the project root, and when you are going to build or run the project, run the command using that root file.

I don't understand the reasoning behind splitting the requirements in two files (current behavior) or stashing an Ansible requirements file in some subdirectory... I think if anything, following the precedent of other software ecosystems makes for the least end-user frustration, when they come in and realize they have to do special (and different) things to structure their projects for Ansible Tower/AWX.

geerlingguy:
> download-at-execution is a pardigm we'd like to move away from

I dislike that too; for most of my Ansible projects I just commit all the dependencies to the codebase so I don't have to rely on galaxy (or AH) to be up and reachable for all the role/collection installs.

[...]

PR #8312, opened 2020-10-02
related #4540 updating project_update.yml to accept yaml extension 

thedoubl3j:
##### SUMMARY

Related #4540 changes allows for the detection of roles/requirements.yaml. This is mainly a shepherd PR from @rebeccahhh and a community member from earlier in the year.   

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
$ make VERSION
$ awx: 15.0.0

```

##### ADDITIONAL INFORMATION

Tower-qa PR [here](https://github.com/ansible/tower-qa/pull/5585)

### DIFF
--- a/awx/playbooks/project_update.yml
+++ b/awx/playbooks/project_update.yml
@@ -159,23 +159,29 @@
   gather_facts: false
   connection: local
   name: Install content with ansible-galaxy command if necessary
+  vars:
+    yaml_exts:
+      - {ext: .yml}
+      - {ext: .yaml}
   tasks:
 
     - block:
-        - name: detect requirements.yml
+        - name: detect roles/requirements.(yml/yaml)
           stat:
-            path: '{{project_path|quote}}/roles/requirements.yml'
+            path: "{{project_path|quote}}/roles/requirements{{ item.ext }}"
+          with_items: "{{ yaml_exts }}"
           register: doesRequirementsExist
 
-        - name: fetch galaxy roles from requirements.yml
+        - name: fetch galaxy roles from requirements.(yml/yaml)
           command: >
-            ansible-galaxy role install -r roles/requirements.yml
+            ansible-galaxy role install -r {{ item.stat.path }}
             --roles-path {{projects_root}}/.__awx_cache/{{local_path}}/stage/requirements_roles
             {{ ' -' + 'v' * ansible_verbosity if ansible_verbosity else '' }}
           args:
             chdir: "{{project_path|quote}}"
           register: galaxy_result
-          when: doesRequirementsExist.stat.exists
+          with_items: "{{ doesRequirementsExist.results }}"
+          when: item.stat.exists
           changed_when: "'was installed successfully' in galaxy_result.stdout"
           environment:
             ANSIBLE_FORCE_COLOR: false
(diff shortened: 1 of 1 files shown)


awx/playbooks/project_update.yml
```diff
@@ -186,20 +192,22 @@
...
-            path: '{{project_path|quote}}/collections/requirements.yml'
+            path: "{{project_path|quote}}/collections/requirements{{ item.ext }}"
+          with_items: "{{ yaml_exts }}"
           register: doesCollectionRequirementsExist
 
-        - name: fetch galaxy collections from collections/requirements.yml
+        - name: fetch galaxy collections from collections/requirements.(yml/yaml)
           command: >
-            ansible-galaxy collection install -r collections/requirements.yml
+            ansible-galaxy collection install -r collections/requirements{{ item.item.ext }}
```
AlanCoding: I feel like it would be nicer to use `{{ item.path }}` or something like that, because you're kind of re-constructing the path when it was already given in the prior task and saved in the result.

AlanCoding:
My concerns before were that:

 - this required messy changes in `tasks.py`
 - the output of the project playbook was _really_ noisy, and this was making it worse

The first point is fixed, because the caching mechanism got rid of a lot of that complexity.

The second point has been mitigated by using tags so that we don't show as much in completely irrelevant cases. I'm fine with the value/noise ratio for the added skipped tasks I see will come from this.

I'm fine to give my approval if someone says it's been tested properly.

awx/playbooks/project_update.yml
rebeccahhh: I agree with this idea.

thedoubl3j: change made

wenottingham:
If this works, it seems reasonable. It looks like it would download both if someone had both of requirements.yml and requirements.yaml... which is weird, but if they do that in their repo, I guess they meant it.

awx/playbooks/project_update.yml
```diff
@@ -159,23 +159,29 @@
...
   name: Install content with ansible-galaxy command if necessary
+  vars:
+    yaml_exts:
+      - {ext: .yml}
+      - {ext: .yaml}
   tasks:
 
     - block:
-        - name: detect requirements.yml
+        - name: detect requirements.(yml/yaml)
```
AlanCoding: minor naming comment - this should say "detect _roles/_ requirements.(yml/yaml)", like the other one for collections

AlanCoding:
This still feels a little messy to me. The alternative that begs attention is:

```yaml
        - name: fetch galaxy collections from collections/requirements.(yml/yaml)
          command: >
            ansible-galaxy collection install -r {{ item}}
            --collections-path {{projects_root}}/.__awx_cache/{{local_path}}/stage/requirements_collections
            {{ ' -' + 'v' * ansible_verbosity if ansible_verbosity else '' }}
          args:
            chdir: "{{project_path|quote}}"
          register: galaxy_collection_result
          with_fileglob: "{{project_path|quote}}/collections/requirements.ya?ml"
          changed_when: "'Installing ' in galaxy_collection_result.stdout"
          environment:
            ANSIBLE_FORCE_COLOR: false
            ANSIBLE_COLLECTIONS_PATHS: "{{projects_root}}/.__awx_cache/{{local_path}}/stage/requirements_collections"
            GIT_SSH_COMMAND: "ssh -o StrictHostKeyChecking=no"
```

This would still need some testing.

This eliminates a lot of the syntax legwork going on in the playbook. There's no conditional for the file existing, there's no separate task for discovering the file. It loops over the any files in the (strictly 2) allowable places.

The messiness isn't bad right now, but adding `requirements.yml` to the mix would mean that there are 3 different types of files to get things from. That means that there are 6 different files we could theoretically install. Add the "detection" tasks, and you have 12 tasks to do what you're trying to do.

The `with_fileglob` method here would actually eliminate need for the `block:` syntax, since the tasks themselves would absorb those `when:` conditionals.

In fact, if we're actually planning on doing #6270, I would go even further to a crazy extreme.

There should only be 1 task for what we're doing here. Since we're supporting `requirements.yml/yaml`, `roles/requirements.yml/yaml` and `collections/requirements.yml/yaml`, then _those_ should be the loop. The enablement playbook variable and the command suffix (" role", " collection", "") could be parameterized in the dict in playbook `vars:`. That would justify its existence, whereas the current yaml_ext dict is just kind of awkward.

thedoubl3j:
worked with @rebeccahhh on getting an idea of what that would look like @AlanCoding currently testing locally before pushing

thedoubl3j:
I know @AlanCoding is out now but wanted to drop the reasoning of why we didn't adopt the all in one task or address the example he listed. 

After testing, we knew that what was there worked and it had test coverage. We also had another contributor (@sbaerlocher)  follow the same pattern to add functionality for #6270 so we can get 2 birds here and not need a lot of rewrite. 

The 3 in 1 would also introduce a lot of complexity that would take more time to get working so a simpler approach (all be it more lines) was taken. 

We also now have a base to go from for the refactor that is in a slight working condition that can get worked on after.

beeankha (approved):
LGTM! 👍

PR #3138, opened 2019-02-01
[WIP] Remove job isolation from management command logic

AlanCoding:
##### SUMMARY
Fixes bug with inventory update isolation

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
3.3.0
```

##### ADDITIONAL INFORMATION
Job isolation for inventory updates is already performed by the dispatcher task code. There is no need to do it here, and it breaks everything whenever something changes.

I thought I saw this done in the PR https://github.com/ansible/awx/pull/3041, but I think it was removed from that diff.

PR #451, opened 2017-10-17
[WIP] Proof of principle of directly calling task management commands

AlanCoding:
Related to #200, as a pre-req item that I'm just throwing out there as a possibility.

Immediate benefit: This eliminates the time taken for initialization of the inventory import `awx-manage` command inside of the task code. Instead of creating a new subprocess to do this, it just dives straight into the command, reducing the time for a simple inventory import from about 12.5 to 8.5 seconds. Not monumental, sure, but a large fraction of the remaining time is `ansible-inventory` after all.

The reason I find this cleaner is because we are only creating a new runtime context for running the actual user code, as opposed to our own code. The inventory import command, in particular, can not pass off the responsibility for management of the bwrap isolation to that task code, because we still try to enforce that isolated if it's ran from the CLI.

We can also (somewhat more easily) extend this to system jobs. No isolation is needed there, because it's 100% AWX code.

This still needs some place to hook into in order to implement the output event callback (related to the new feature event-izing all stdout), plus the cancel callback.

PR #8323, opened 2020-10-05
Refactor to hook in programmatic use of inventory import saving-to-DB code

AlanCoding:
##### SUMMARY
~Sharing as a WIP because I may be pulled away for an extended leave soon.~

Refactor of inventory updates so that we can easily containerize the `ansible-inventory` subprocess.

This changes the `tasks.py` code to call `ansible-inventory` as opposed to `awx-manage`. We want inventory updates to run in the control plane, where the Ansible install will be available, but AWX utilities will not. Correspondingly, the processing of saving the data that is returned is shifted to the main code - instead of saving in the management command, we save in `pre_run_hook`, and some clever tricks are employed to emit logs in the same style they always have been.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
15.0.0
```

##### ADDITIONAL INFORMATION
This is picking up some really old work, but only taking the _absolutely_ necessary parts to get inventory updates into containers.

~As of opening this PR, this is only progressing to the point of making the `awx-manage inventory_import` interface work.~

~With this established, I'm diving into `awx/main/tasks.py` in order to change the `args` file from `awx-manage` to `ansible-inventory`, then the process of saving the inventory will get pushed to `post_run_hook`.~

Old related work includes https://github.com/ansible/awx/pull/451 and #3138

### DIFF
--- a/awx/main/management/commands/inventory_import.py
+++ b/awx/main/management/commands/inventory_import.py
@@ -863,68 +863,69 @@ class Command(BaseCommand):
         self.inventory_update.save(update_fields=['org_host_limit_error'])
 
     def handle(self, *args, **options):
-        # Load inventory and related objects from database.
-        inventory_name = options.get('inventory_name', None)
-        inventory_id = options.get('inventory_id', None)
-        if inventory_name and inventory_id:
-            raise CommandError('--inventory-name and --inventory-id are mutually exclusive')
-        elif not inventory_name and not inventory_id:
-            raise CommandError('--inventory-name or --inventory-id is required')
+        with advisory_lock('inventory_{}_import'.format(self.inventory.id)):
+            # Load inventory and related objects from database.
+            inventory_name = options.get('inventory_name', None)
+            inventory_id = options.get('inventory_id', None)
+            if inventory_name and inventory_id:
+                raise CommandError('--inventory-name and --inventory-id are mutually exclusive')
+            elif not inventory_name and not inventory_id:
+                raise CommandError('--inventory-name or --inventory-id is required')
 
-        # Obtain rest of the options needed to run update
-        raw_source = options.get('source', None)
-        if not raw_source:
-            raise CommandError('--source is required')
-        verbosity = int(options.get('verbosity', 1))
-        self.set_logging_level(verbosity)
-        venv_path = options.get('venv', None)
+            # Obtain rest of the options needed to run update
+            raw_source = options.get('source', None)
+            if not raw_source:
+                raise CommandError('--source is required')
+            verbosity = int(options.get('verbosity', 1))
+            self.set_logging_level(verbosity)
+            venv_path = options.get('venv', None)
 
-        # Load inventory object based on name or ID.
-        if inventory_id:
-            q = dict(id=inventory_id)
-        else:
-            q = dict(name=inventory_name)
-        try:
-            inventory = Inventory.objects.get(**q)
-        except Inventory.DoesNotExist:
-            raise CommandError('Inventory with %s = %s cannot be found' % list(q.items())[0])
-        except Inventory.MultipleObjectsReturned:
-            raise CommandError('Inventory with %s = %s returned multiple results' % list(q.items())[0])
-        logger.info('Updating inventory %d: %s' % (inventory.pk, inventory.name))
+            # Load inventory object based on name or ID.
+            if inventory_id:
+                q = dict(id=inventory_id)
+            else:
+                q = dict(name=inventory_name)
+            try:
+                inventory = Inventory.objects.get(**q)
+            except Inventory.DoesNotExist:
+                raise CommandError('Inventory with %s = %s cannot be found' % list(q.items())[0])
+            except Inventory.MultipleObjectsReturned:
+                raise CommandError('Inventory with %s = %s returned multiple results' % list(q.items())[0])
+            logger.info('Updating inventory %d: %s' % (inventory.pk, inventory.name))
 
 
-        # Create ad-hoc inventory source and inventory update objects
-        with ignore_inventory_computed_fields():
-            source = Command.get_source_absolute_path(raw_source)
+            # Create ad-hoc inventory source and inventory update objects
+            with ignore_inventory_computed_fields():
+                source = Command.get_source_absolute_path(raw_source)
 
-            inventory_source, created = InventorySource.objects.get_or_create(
-                inventory=inventory,
-                source='file',
-                source_path=os.path.abspath(source),
-                overwrite=bool(options.get('overwrite', False)),
-                overwrite_vars=bool(options.get('overwrite_vars', False)),
-            )
-            inventory_update = inventory_source.create_inventory_update(
-                _eager_fields=dict(
-                    job_args=json.dumps(sys.argv),
-                    job_env=dict(os.environ.items()),
-                    job_cwd=os.getcwd())
-            )
+                inventory_source, created = InventorySource.objects.get_or_create(
+                    inventory=inventory,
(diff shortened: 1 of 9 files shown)


awx/main/tasks.py
```diff
@@ -2707,6 +2680,121 @@ def pre_run_hook(self, inventory_update, private_data_dir):
...
+            options['exclude_empty_groups'] = True
+        if getattr(settings, '%s_INSTANCE_ID_VAR' % src.upper(), False):
+            options['instance_id_var'] = getattr(settings, '%s_INSTANCE_ID_VAR' % src.upper())
+
+        # Verbosity is applied to saving process, as well as ansible-inventory CLI option
+        if inventory_update.verbosity:
+            options['verbosity'] = inventory_update.verbosity
+
+        # Mock ansible-runner events
+        class CallbackHandler(logging.Handler):
```
AlanCoding: This class could be moved to our `awx.main.utils.handlers`, but a lot of it is quirky stuff specific to inventory updates.

AlanCoding:
This work is starting to stabilize nicely, the largest single shortcoming right now is that the line numbers are not consistent. We need some new testing for that, and _probably_ have to record the last-seen `end_line` from `event_handler`, and this will come with some vague plans about how to handle that with the move to receptor.

AlanCoding:
Remaining cases that are getting picked up by integration tests:

 - Certain log lines were removed from output, like "Updating inventory" - these were removed because they come _before_ the `ansible-inventory` call, and now I'm only doing the faux log output for the _save_ step
 - Some error reporting was removed like "Source does not exist", which now surfaces with the `ansible-inventory` error... this could be made a pre-check in `tasks.py`, but I don't feel strongly that it should be
 - We lost redirection of some warnings like "[WARNING]: Invalid characters were found in group names but not replaced, use", but I also don't want to get rid of those

Basically, these are look & feel changes. Some will require test changes, but we need to get into the mindset that many fundamental changes are expected, and we need some good messaging around those.

AlanCoding:
Integration test agenda:

 - Write new test that assures correct ordering of inventory update events, meaning that the range of line numbers all add up
   this will correspond to a change which still needs to be made for `event_handler` to cache `end_line` locally
 - Split canceling test into 2, right now it runs a sleep script, and has a wait step in the wrong place
   Need 1 test that waits like ~5 seconds after status="running" (as act of faith) and cancels `ansible-inventory` process
   Need 1 test that imports a simple ini inventory with... say... 30 thousand hosts, and assures speedy cancellation after `ansible-inventory` sub-process is completed.

unlikelyzero:
@jladdjr please ensure that the patternfly_e2e component is passing in yolo before merge. Please reach out to me or @tiagodread if it's not

jladdjr:
- [ ] @unlikelyzero you got it 👍

unlikelyzero:
Note: this affects the default JT details page and so we'll need to update the UI Tests to match this new functionality

jladdjr:
Wrote up issue for the runner line count being off-by-one because of extra newlines showing up in `ansible-inventory` output:
https://github.com/ansible/ansible-runner/issues/566

Have a fix for that issue up here:
https://github.com/ansible/ansible-runner/pull/567

awx/main/tasks.py
```diff
@@ -2805,6 +2806,8 @@ def emit(self, record):
             # for us.
             save_status, tb, exc = cmd.perform_update(options, data, inventory_update)
         except Exception as raw_exc:
+            if exc is None:
```
jladdjr: is it even possible for `perform_update` to return any values if it threw an exception? wondering if `exc` should be dropped entirely in the except clause and replaced with `raw_exc`. I'm fuzzy on the context here, though, so not sure that I'm reading this correctly.

In any case, sounds like maybe you caught the root cause of some host limit failures, which is awesome.

AlanCoding: Oh yeah, you might have a point, I knew this was ugly, but maybe it's just not even correct given that.

If you read the code of the method, it's a mess. A license error can result in multiple different scenarios. I could throw and uncaught `CommandError`, or it could use the internal license error exception and be caught within the method. If the method catches it, then it won't throw an exception here, but `exc` will be non-null.

We should refactor that if possible.

One thing to be mindful of - our behavior should be the same for all license error scenarios. That's going to be kind of hard, because if it's a `CommandError`, then how do you know if it was a license failure? If refactoring, you might start with that, and maybe enforce that `perform_update` _never_ throws a `CommandError`, and replace those cases with license error? I end with a question mark, because this is speculative, and it needs more thought and consideration.

awx/main/management/commands/inventory_import.py
```diff
@@ -959,34 +904,100 @@ def mark_org_limits_failure(self, save=True):
...
-        venv_path = options.get('venv', None)
+        # Load inventory and related objects from database.
+        inventory_name = options.get('inventory_name', None)
+        inventory_id = options.get('inventory_id', None)
+        if inventory_name and inventory_id:
+            raise CommandError('--inventory-name and --inventory-id are mutually exclusive')
+        elif not inventory_name and not inventory_id:
+            raise CommandError('--inventory-name or --inventory-id is required')
+
+        with advisory_lock('inventory_{}_import'.format(inventory_id)):
```
AlanCoding, line 918: I think that we shouldn't use 2 locks for this, because I don't think it's necessary and it's complicating other considerations.

The bug you fixed came down to the fact that we prematurely initialized the `inventory` object. In this PR, it is passed into the `perform_update` method via `inventory_update.inventory`. We need load inventory after we obtain the lock (so that we get updated `inventory.variables`, in particular), but we can re-load it.

Where we obtain the lock inside of `perform_update`, where we used to do `self.load_inventory_from_database()`, we just instead just do `inventory_update.inventory.refresh_from_db()`, and then that should be plenty to get rid of the outer lock inside of `handle` IMO

jladdjr:
To give a general update on the state of things here, @AlanCoding and I have continued seeing some unexpected fallout because of this refactor. We've pursued a number of theories here (related to subscriptions, license caching, locks, and even taken time to sanity check how we're exercising our tests, too), but still haven't gotten to the bottom of things here. If things are a bit quiet here, it's because we're still off looking into different theories.

This morning we reached out to @rooftopcellist to get some more background on the recent subscription changes in case our changes here are some how causing some unexpected fallout related to subscriptions. We've seen some odd behavior with notifications that seems to point in that direction.

AlanCoding:
Before anyone else gets too far into this, I'd like to get the patches from https://github.com/AlanCoding/awx/pull/8/ incorporated, then it will be closer to "trying to wrap up" as opposed to "just got started"

awx/main/tasks.py
```diff
@@ -2707,6 +2703,130 @@ def pre_run_hook(self, inventory_update, private_data_dir):
...
+
+        handler = CallbackHandler(
+            self.event_handler, self.cancel_callback,
+            verbosity=inventory_update.verbosity,
+            job_timeout=self.get_instance_timeout(self.instance),
+            counter=self.event_ct
+        )
+        inv_logger = logging.getLogger('awx.main.commands.inventory_import')
+        handler.formatter = inv_logger.handlers[0].formatter
+        inv_logger.handlers[0] = handler
```
AlanCoding, line 2755: After this is finished, we should probably reset this logger back to how it was before.

unlikelyzero:
@AlanCoding @jladdjr please let us know as you get closer to merge and we'll make sure the e2e tests are 🟢

shanemcd (approved):

jladdjr:
🎉

Issue #5209, opened 2019-11-01
Add Support for BitBucket webhooks


bkmeneguello:
##### ISSUE TYPE
- Feature Request

##### SUMMARY
Restore commit 885841caeaea0f7c86d39d6e7a66b960124289c2 to add support for BitBucket webhooks, since Server edition supports signed payloads and Cloud edition suggests IP Whitelists ([manage webhooks](https://confluence.atlassian.com/bitbucket/manage-webhooks-735643732.html) and [ip whitelist](https://confluence.atlassian.com/bitbucket/what-are-the-bitbucket-cloud-ip-addresses-i-should-use-to-configure-my-corporate-firewall-343343385.html))

@jbradberry

rmauc:
+1 for Bitbucket Server

nitinkansal1984:
+1 for bitbucket

Forest-Troll:
I try restore this commit with some changes. It fine work with bitbucket 5.12 and AWX 9.1.0 for me. Not tested this with last environment yet.

https://github.com/Forest-Troll/awx/tree/bitbucket-return

ryanpetrello:
Support for this in AWX is blocked, because Bitbucket does not support signed webhooks in its free product.  Details at:

https://github.com/ansible/awx/pull/8328#issuecomment-707171773

christianmarangoni:
+1

rmauc:
Is there any update on this RFE?
We are looking for the webhook to Bitbucket Server (on-prem) V7.6.0
Thanks

jvandermeulen:
+1 for Bitbucket Server

jbradberry:
@rmauc @jvandermeulen https://github.com/ansible/awx/pull/8328#issuecomment-707171773 as noted above.

rmauc:
Is waiting for the free product really justification to restrict enterprise customers from fully benefiting from the Tower feature set?
Afterall, you can get 10 x Bitbucket Server licenses for $10USD per year. https://www.atlassian.com/licensing/bitbucket#server-licensing

opfpqgoon:
Why could we not enable bitbucket on-prem (server/datacenter) Webhooks with clear documentation & UI tooltips explaining that this does not work with bitbucket cloud?

Having bitbucket Webhooks integration with AWX would massively simplify CI/CD workflows as there's no longer a need for an intermediary service to receive Webhooks and trigger tower-cli commands.

[...]

PR #8328, opened 2020-10-06
Add support for BitBucket Webhooks

Forest-Troll:
##### SUMMARY
Restore commit https://github.com/ansible/awx/commit/885841caeaea0f7c86d39d6e7a66b960124289c2  with some additional fix for integrate with Bitbucket Server.
related #5209

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION

```
awx: 15.0.0
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/migrations/0092_v360_webhook_mixin.py
+++ b/awx/main/migrations/0092_v360_webhook_mixin.py
@@ -24,7 +24,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='jobtemplate',
             name='webhook_service',
-            field=models.CharField(blank=True, choices=[('github', 'GitHub'), ('gitlab', 'GitLab'), ('bitbucket', 'Bitbucket')]], help_text='Service that webhook requests will be accepted from', max_length=16),
+            field=models.CharField(blank=True, choices=[('github', 'GitHub'), ('gitlab', 'GitLab'), ('bitbucket', 'Bitbucket')], help_text='Service that webhook requests will be accepted from', max_length=16),
         ),
         migrations.AddField(
             model_name='workflowjobtemplate',
--- a/awx/main/migrations/0092_v360_webhook_mixin.py
+++ b/awx/main/migrations/0092_v360_webhook_mixin.py
@@ -24,7 +24,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='jobtemplate',
             name='webhook_service',
-            field=models.CharField(blank=True, choices=[('github', 'GitHub'), ('gitlab', 'GitLab')], help_text='Service that webhook requests will be accepted from', max_length=16),
+            field=models.CharField(blank=True, choices=[('github', 'GitHub'), ('gitlab', 'GitLab'), ('bitbucket', 'Bitbucket')]], help_text='Service that webhook requests will be accepted from', max_length=16),
         ),
         migrations.AddField(
             model_name='workflowjobtemplate',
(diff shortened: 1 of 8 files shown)


awx/main/migrations/0092_v360_webhook_mixin.py
```diff
@@ -39,7 +39,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='workflowjobtemplate',
             name='webhook_service',
-            field=models.CharField(blank=True, choices=[('github', 'GitHub'), ('gitlab', 'GitLab')], help_text='Service that webhook requests will be accepted from', max_length=16),
+            field=models.CharField(blank=True, choices=[('github', 'GitHub'), ('gitlab', 'GitLab'), ('bitbucket', 'Bitbucket')], help_text='Service that webhook requests will be accepted from', max_length=16),
```
wenottingham, line 42: I do not believe you can rewrite past migrations in this way.

Forest-Troll: But hooks with BitbucketServers was worked.
But callback status to BitbucketServer not working yet...

I changed cipher for has validate to sha256, and added empty method for get_event_status_api, nothing more.

Oh, I think i know what you mean :-)

ryanpetrello: No, this needs to be a new migration.  You should be able to generate one using `awx-manage makemigrations`.

ryanpetrello:
@jbradberry I remember we looked at this initially, but eventually yanked support for some reason.  Do either of you remember why, and does this PR address it (or has some needed support changed on the Bitbucket end, somehow?)

jbradberry:
> @jbradberry I remember we looked at this initially, but eventually yanked support for some reason. Do either of you remember why, and does this PR address it (or has some needed support changed on the Bitbucket end, somehow?)

Yes, we needed to obtain a license for Bitbucket Server and it didn't look like we were going to get it in time.  The free website version of Bitbucket does not support signed webhooks, so support for that should be excluded, if that can be done easily.  It also wasn't clear without trying it out whether the machinery that I built that worked for Github and Gitlab would work for Bitbucket without serious elbow grease.

I'm pretty sure that this PR is just doing a `git revert ...` on the commit where I yanked out the Bitbucket code and doesn't solve the problems of 1) knowing whether the code I wrote blindly would even work correctly, and 2) migrating data that has already landed.

ryanpetrello:
> The free website version of Bitbucket does not support signed webhooks

Yes, now I remember.  For me, if there's not some _obvious_ way to detect free vs paid and restrict access on the AWX side, then I don't think we should merge this, because I expect the vast majority of users will end up trying to use it with the free version.

*Maybe* we could interpret free vs paid as "if it doesn't _have_ a signature, it is forbidden".  That's a pretty lousy user experience, though, and I'm guessing it'll confuse users (and likely prompt users to ask *us* for an "I don't care about security" button).

Honestly, I'm a little surprised and disappointed that Bitbucket is making people pay for secure defaults (clearly it works if it's in the enterprise product, and they've chosen not to provide it in the free web service), but c’est la vie, I guess.

So to summarize, I'd say we're back at the same place we were a year ago when we decided not to support Bitbucket.

ryanpetrello (changes requested):
Given my comments on this PR, I vote 👎 until Bitbucket can properly support signed web hooks in its free product.

opfpqgoon:
Given that trial licenses of Atlassian products can be requested pretty easily (https://confluence.atlassian.com/bitbucketserver/install-a-bitbucket-server-trial-867192384.html), I don't fully understand why we need to wait on bitbucket cloud to support signed Webhooks.  Given that Atlassian is completely revamping their license model in February, there are quite a few enterprise teams who will be moving from Server licensing to Datacenter licensing.

Why could we not enable bitbucket on-prem (server/datacenter) Webhooks with clear documentation & UI tooltips explaining that this does *not* work with bitbucket cloud?

Having bitbucket Webhooks integration with AWX would massively simplify CI/CD workflows as there's no longer a need for an intermediary service to receive Webhooks and trigger tower-cli commands.

I'd be happy to help test anything related to bitbucket on-prem webhook integration.

tech2734:
I am running into more and more clients that are using Bitbucket and have a pretty clear need for this functionality.   Please reconsider supporting it.  I would help out with testing as well.  thanks.

PR #8333, opened 2020-10-06
reduce per-job database query count

chrismeyersfsu:
* Do not query the database for the set of Instance that belong to the
group for which we are trying to fit a job on, for each job.
* Instead, cache the set of instances per-instance group.

Before these changes, the task manager would take 1-2 seconds to decide and begin running a job. With these changes, it takes about 0.05 seconds per job.

| Name                                                | 200 jobs           | 1,000 jobs         |
| --------------------------------------------------- | ------------------ | ------------------ |
| Fit optimization                                    | 11s \| 0.055 s/job | 58s \| 0.058 s/job |

### DIFF
--- a/awx/main/models/ha.py
+++ b/awx/main/models/ha.py
@@ -261,18 +261,20 @@ class InstanceGroup(HasPolicyEditsMixin, BaseModel, RelatedJobsMixin):
         app_label = 'main'
 
 
-    def fit_task_to_most_remaining_capacity_instance(self, task):
+    @staticmethod
+    def fit_task_to_most_remaining_capacity_instance(task, instances):
         instance_most_capacity = None
-        for i in self.instances.filter(capacity__gt=0, enabled=True).order_by('hostname'):
+        for i in instances:
             if i.remaining_capacity >= task.task_impact and \
                     (instance_most_capacity is None or
                      i.remaining_capacity > instance_most_capacity.remaining_capacity):
                 instance_most_capacity = i
         return instance_most_capacity
 
-    def find_largest_idle_instance(self):
+    @staticmethod
+    def find_largest_idle_instance(instances):
         largest_instance = None
-        for i in self.instances.filter(capacity__gt=0, enabled=True).order_by('hostname'):
+        for i in instances:
             if i.jobs_running == 0:
                 if largest_instance is None:
                     largest_instance = i
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -7,6 +7,7 @@ import logging
 import uuid
 import json
 import random
+from types import SimpleNamespace
 
 # Django
 from django.db import transaction, connection
@@ -45,6 +46,15 @@ logger = logging.getLogger('awx.main.scheduler')
 class TaskManager():
 
     def __init__(self):
+        '''
+        Do NOT put database queries or other potentially expensive operations
+        in the task manager init. The task manager object is created every time a
+        job is created, transitions state, and every 30 seconds on each tower node.
+        More often then not, the object is destroyed quickly because the NOOP case is hit.
+
+        The NOOP case is short-circuit logic. If the task manager realizes that another instance
+        of the task manager is already running, then it short-circuits and decides not to run.
+        '''
         self.graph = dict()
         # start task limit indicates how many pending jobs can be started on this
         # .schedule() run. Starting jobs is expensive, and there is code in place to reap
@@ -52,10 +62,30 @@ class TaskManager():
         # 5 minutes to start pending jobs. If this limit is reached, pending jobs
         # will no longer be started and will be started on the next task manager cycle.
         self.start_task_limit = settings.START_TASK_LIMIT
+
+    def after_lock_init(self):
+        '''
+        Init AFTER we know this instance of the task manager will run because the lock is acquired.
+        '''
+        instances = Instance.objects.filter(capacity__gt=0, enabled=True)
+        self.real_instances = {i.hostname: i for i in instances}
+
+        instances_partial = [SimpleNamespace(obj=instance,
+                                             remaining_capacity=instance.remaining_capacity,
+                                             capacity=instance.capacity,
+                                             jobs_running=instance.jobs_running,
+                                             hostname=instance.hostname) for instance in instances]
+
+        instances_by_hostname = {i.hostname: i for i in instances_partial}
+
         for rampart_group in InstanceGroup.objects.prefetch_related('instances'):
             self.graph[rampart_group.name] = dict(graph=DependencyGraph(rampart_group.name),
                                                   capacity_total=rampart_group.capacity,
-                                                  consumed_capacity=0)
+                                                  consumed_capacity=0,
+                                                  instances=[])
+            for instance in rampart_group.instances.filter(capacity__gt=0, enabled=True).order_by('hostname'):
+                if instance.hostname in instances_by_hostname:
+                    self.graph[rampart_group.name]['instances'].append(instances_by_hostname[instance.hostname])
 
     def is_job_blocked(self, task):
         # TODO: I'm not happy with this, I think blocking behavior should be decided outside of the dependency graph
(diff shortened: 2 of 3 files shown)


fosterseth (approved):

fosterseth:
looks good, tested the patch and can confirm the speed boost

awx/main/scheduler/task_manager.py
```diff
@@ -484,15 +513,22 @@ def process_pending_tasks(self, pending_tasks):
...
                 remaining_capacity = self.get_remaining_capacity(rampart_group.name)
                 if not rampart_group.is_containerized and self.get_remaining_capacity(rampart_group.name) <= 0:
                     logger.debug("Skipping group {}, remaining_capacity {} <= 0".format(
                                  rampart_group.name, remaining_capacity))
                     continue
 
-                execution_instance = rampart_group.fit_task_to_most_remaining_capacity_instance(task)
+                execution_instance = rampart_group.fit_task_to_most_remaining_capacity_instance(task, self.graph[rampart_group.name]['instances'])
                 if execution_instance:
+                    execution_instance.capacity -= task.task_impact
```
fosterseth: could we be double subtracting here if `idle_instance_that_fits == execution_instance`?

fosterseth:
we should also make sure the changes here are compatible with this code that we call on running tasks, as it looks to modify your `self.graph` before we process any pending tasks

https://github.com/ansible/awx/blob/37b3cc72b27fc06e3c6b1ca5e5bf5f7574dc366d/awx/main/scheduler/task_manager.py#L548

https://github.com/ansible/awx/blob/37b3cc72b27fc06e3c6b1ca5e5bf5f7574dc366d/awx/main/managers.py#L203

moonrail:
@chrismeyersfsu 
Thank you for mentioning it in #7655

This is game changing - I can confirm a ton faster processing time for starting jobs - your stated 20 times are easily valid.

With START_TASK_LIMIT 10 one run with starting 10 jobs takes < 2 seconds, so > 6 seconds improvement.

With START_TASK_LIMIT 100 one run with starting 100 jobs takes < 7 seconds, so ~20-60s improvement. I've also not experienced the task manager now getting killed, as it did not run longer than 5 minutes.

Still - I have found two potential issues:
1. The task manager is run continiously (which is good), but at this new lightning fast pace it also leads to a lot of log-output - maybe a problem with external logging arises when DEBUG-Level is required because of this? I have currently no environment to test this.
2. This is probably not directly an issue with this PR, but since you've added some sort of instance-cache it could be:
From my obvservations the task manager is started on a node and runs there for X iterations/time, until it gives up lock and another node sets lock and it runs there.
I've noticed, that the Task Manager does not use "newly" enabled instances since its start on a specific node.
<br>For this I've set all instances to disabled, started 1k jobs and watched the Task Manager logs. It ran in a loop.
I've then enabled all 15 instanced, but no jobs were started in further Task Manager iterations as there was "0 capacity remaining".
As soon as the Task Manager switched nodes, it detected available instances.

awx/main/scheduler/task_manager.py
```diff
@@ -45,17 +46,46 @@
 class TaskManager():
 
     def __init__(self):
+        '''
+        Do NOT put database queries or other potentially expensive operations
```
ryanpetrello, line 50: It's times like these that I employ one of my favorite comments:

https://github.com/ansible/awx/blob/devel/awx/main/tasks.py#L1213

This may just be my excuse to use emojis in source code, though.

ryanpetrello (approved):
I like this PR a lot 👍

awx/main/scheduler/task_manager.py
```diff
@@ -483,24 +512,23 @@ def process_pending_tasks(self, pending_tasks):
...
                     logger.debug("Skipping group {}, remaining_capacity {} <= 0".format(
                                  rampart_group.name, remaining_capacity))
                     continue
 
-                execution_instance = rampart_group.fit_task_to_most_remaining_capacity_instance(task)
-                if execution_instance:
-                    logger.debug("Starting {} in group {} instance {} (remaining_capacity={})".format(
-                                 task.log_format, rampart_group.name, execution_instance.hostname, remaining_capacity))
-                elif not execution_instance and idle_instance_that_fits:
+                execution_instance = InstanceGroup.fit_task_to_most_remaining_capacity_instance(task, self.graph[rampart_group.name]['instances']) or \
```
ryanpetrello, line 521: Conspiracy theory time, come along and join me @chrismeyersfsu @fosterseth 

1.  The task manager starts, obtains the lock, calls `after_lock_init`, and has a set of known instances.  We arrive at a moment in time _right before_ this code.
2. The user removes an instance via the API, or via `awx-manage deprovision_instance`, and it's the one we happen to pick here via `execution_instance`.
3.  What **terrible, terrible** thing happens next?

awx/main/models/ha.py
```diff
@@ -261,18 +261,20 @@ class Meta:
         app_label = 'main'
 
 
-    def fit_task_to_most_remaining_capacity_instance(self, task):
+    @staticmethod
+    def fit_task_to_most_remaining_capacity_instance(task, instances):
         instance_most_capacity = None
-        for i in self.instances.filter(capacity__gt=0, enabled=True).order_by('hostname'):
+        for i in instances:
```
ryanpetrello, line 267: @chrismeyersfsu @fosterseth to address my race condition comment below, perhaps we should sprinkle in some sort of `refresh_from_db` call with an `try...catch` to make _absolutely sure_ the selected instance *actually* still exists (and wasn't deleted in the time it took the task manager to start and enter this code block?)

awx/main/scheduler/task_manager.py
fosterseth: hmm good point. Not sure of the behavior, but this method seems to go along with this scenario https://github.com/ansible/awx/blob/e6c124962bca6215b8b6d4673ec252195df234be/awx/main/scheduler/task_manager.py#L533

but that might be more for jobs after `.save()`. I think in the case you describe, `task.save()` should fail

ryanpetrello: I think this is *very unlikely* to happen, but we should probably test to make sure that it doesn't result in something bad, like a task never running, and getting stuck in `waiting` forever (or worse, a task-manager breaking exception).

fosterseth: okay I tried this on a local 3 node cluster install. If I delete the execution instance the job is assigned to before the job is saved in `start_task` method, it still runs successfully on that instance.

image

Says it fails in the explanation but actually succeeds (I see hello world playbook run output)

I get the following errors in the log, on loop

```
tools_awx_1_1 | 2020-10-12 18:47:16,633 ERROR    awx.main.dispatch encountered an error communicating with redis to store dispatcher statistics
tools_awx_1_1 | Traceback (most recent call last):
tools_awx_1_1 |   File "/awx_devel/awx/main/dispatch/worker/base.py", line 111, in record_statistics
tools_awx_1_1 |     self.redis.set(f'awx_{self.name}_statistics', self.pool.debug())
tools_awx_1_1 |   File "/awx_devel/awx/main/dispatch/pool.py", line 341, in debug
tools_awx_1_1 |     self.cleanup()
tools_awx_1_1 |   File "/awx_devel/awx/main/dispatch/pool.py", line 436, in cleanup
tools_awx_1_1 |     reaper.reap(excluded_uuids=running_uuids)
tools_awx_1_1 |   File "/awx_devel/awx/main/dispatch/reaper.py", line 38, in reap
tools_awx_1_1 |     (changed, me) = Instance.objects.get_or_register()
tools_awx_1_1 |   File "/awx_devel/awx/main/managers.py", line 150, in get_or_register
tools_awx_1_1 |     return (False, self.me())
tools_awx_1_1 |   File "/awx_devel/awx/main/managers.py", line 108, in me
tools_awx_1_1 |     raise RuntimeError("No instance found with the current cluster host id")
tools_awx_1_1 | RuntimeError: No instance found with the current cluster host id
.
.
tools_awx_1_1 | 2020-10-12 18:47:13,473 ERROR    awx.main.wsbroadcast AWX is currently installing/upgrading.  Trying again in 5s...
```

ryanpetrello: 💥

ryanpetrello:
replaced by https://github.com/ansible/awx/pull/8403

Issue #8344, opened 2020-10-07
Enhance tower_job_wait to support project_update/inventory_update/workflow_job in addtion to job


donald-picard-kr:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
tower_job_wait only supports waiting on tasks of the type 'job'

I would like to see this module (or potenitally a seperate one, the code I wrote is to suppor this in one module) support waiting for other similar jobs
Addtional "job" types allowed
- 'project_update'
- 'job'   # default
- 'inventory_update'
- 'workflow_job'

I have started development work on this feature. A pull will be incoming.

donald-picard-kr:
This has been released in v16.0.0 
it is working and has been tested with
project_updates, inventory_updates, workflow_jobs against our production Ansible Tower 3.7.2 without any issues. I do not have an awx server to test against but the below integration tests are implemented. 

This is in addition to the integration testing completed in the referenced pull request. 
this included 'workflow_jobs' and the default (var 'job_type' is not defined) which is 'jobs'

jobs is the original implementation that of the module that was left as the default wait type.

PR #8345, opened 2020-10-07
add support for wait of project_updates jobs, inventory_update, and w…

donald-picard-kr:
…orkflow_job.

##### SUMMARY

This is an enhancement to allow the tower job wait module to support waiting for other similar jobs
Addtional "job" types allowed
- 'project_update'
- 'job'   # default
- 'inventory_update'
- 'workflow_job'

I wrote this some time ago and will need to rebase this to get it into this branch
My code is based off off commit 2e4e687d694c9b559189224e785ae7c104145bb4 - Commits on Feb 26, 2020

related #8344

https://github.com/ansible/awx/issues/8344

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME

 - tower_job_wait

##### AWX VERSION
```
awx: 9.2.0
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx_collection/plugins/modules/tower_job_wait.py
+++ b/awx_collection/plugins/modules/tower_job_wait.py
@@ -30,7 +30,7 @@ options:
     interval:
       description:
         - The interval in sections, to request an update from Tower.
-        - For backwards compatability if unset this will be set to the average of min and max intervals
+        - For backwards compatibility if unset this will be set to the average of min and max intervals
       required: False
       default: 1
       type: float
@@ -48,6 +48,12 @@ options:
       description:
         - Maximum time in seconds to wait for a job to finish.
       type: int
+    job_type:
+      description:
+        - Job type to wait for
+      choices: ['project_updates', 'jobs', 'inventory_updates', 'workflow_jobs']
+      default: 'jobs'
+      type: str
 extends_documentation_fragment: awx.awx.auth
 '''
 
@@ -99,6 +105,7 @@ def main():
     # Any additional arguments that are not fields of the item can be added here
     argument_spec = dict(
         job_id=dict(type='int', required=True),
+        job_type=dict(choices=['project_updates', 'jobs', 'inventory_updates', 'workflow_jobs'], default='jobs'),
         timeout=dict(type='int'),
         min_interval=dict(type='float'),
         max_interval=dict(type='float'),
@@ -110,6 +117,7 @@ def main():
 
     # Extract our parameters
     job_id = module.params.get('job_id')
+    job_type = module.params.get('job_type')
     timeout = module.params.get('timeout')
     min_interval = module.params.get('min_interval')
     max_interval = module.params.get('max_interval')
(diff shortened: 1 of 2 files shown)


awx_collection/plugins/modules/tower_job_wait.py
```diff
@@ -92,6 +104,7 @@
 
 from ..module_utils.ansible_tower import TowerModule, tower_auth_config, tower_check_mode
 from ansible.module_utils.six import PY2
+from ansible.module_utils.basic import AnsibleModule
```
AlanCoding: Looks like an unused import, the needed logic is obtained from `TowerModule` which has this as a parent class.

donald-picard-kr: Ok I will remove that line. I built this feature a few years ago and updated it in January or so. I see that a lot changed since January. I am working on trying to make this work with the new code. 
I see that `awx_collection/plugins/modules/tower_job_launch.py` has some of what I need it it so I won't have to recode too much.
Thank you for this help. I appreciate it.

I am looking at the `module.get_one()` logic right now.

donald-picard-kr:
The new code logic is very different than what used to be here. I ended up rewritting the enhancement after looking what was current. It looks like this should work.

awx_collection/plugins/modules/tower_job_wait.py
donald-picard-kr: @AlanCoding 
I rewrote this to use the new code base. It now builds sucessfully, passing the build checks.

awx_collection/plugins/modules/tower_job_wait.py
```diff
@@ -48,6 +48,11 @@
       description:
         - Maximum time in seconds to wait for a job to finish.
       type: int
+    job_type:
+      description:
+        - Job type to wait for
+      choices: ['project_updates', 'jobs', 'inventory_updates', 'workflow_jobs']
+      default: 'jobs'
```
beeankha, line 55: `type: str` needs to be specified here.

beeankha:
@donald-picard-kr thank you for this PR!  Could you please rebase against the `devel` branch and also add a task or two into the [integration test playbook](https://github.com/ansible/awx/blob/devel/awx_collection/tests/integration/targets/tower_job_wait/tasks/main.yml) to check that the new `job_type` parameter is valid? I also added a comment above about `type: str` needing to be specified in the module options.

awx_collection/plugins/modules/tower_job_wait.py
```diff
@@ -130,14 +137,14 @@ def main():
...
-    job = module.get_one('jobs', **{
+    job = module.get_one(job_type, **{
         'data': {
             'id': job_id,
         }
     })
 
     if job is None:
-        module.fail_json(msg='Unable to wait on job {0}; that ID does not exist in Tower.'.format(job_id))
+        module.fail_json(msg="Unable to wait on " + job_type.rstrip("s") + ' {0}; that ID does not exist in Tower.'.format(job_id))
```
beeankha: This is super nitpicky, but can we use either `'` or `"` vs both here?

donald-picard-kr: I addressed this issue. I will address the others one at a time.

awx_collection/plugins/modules/tower_job_wait.py
donald-picard-kr: I addressed this one as well. I will pull in devel and add testing next.

beeankha: Awesome, thank you @donald-picard-kr !

donald-picard-kr: > @donald-picard-kr thank you for this PR! Could you please rebase against the `devel` branch and also add a task or two into the [integration test playbook](https://github.com/ansible/awx/blob/devel/awx_collection/tests/integration/targets/tower_job_wait/tasks/main.yml) to check that the new `job_type` parameter is valid? I also added a comment above about `type: str` needing to be specified in the module options.

@beeankha I think I have addressed all of your comments.

beeankha (approved):

PR #8348, opened 2020-10-07
Rename inventory_source param to name in tower_inventory_source_update

### LINKED ISSUE #8347: collection module tower_inventory_source_update should have "name" instead of "inventory_source"
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Module that works on an entity should reference that entity by `name`. It feels more ansibly anyway.

nesanton:
##### SUMMARY

* fixes #8347
* Rename inventory_source to name in the tower_inventory_source_update
* Allow to specify both name or id for `name` and `inventory` params

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

 - tower_inventory_source_update

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx_collection/plugins/modules/tower_inventory_source_update.py
+++ b/awx_collection/plugins/modules/tower_inventory_source_update.py
@@ -22,14 +22,14 @@ description:
     - Update Ansible Tower inventory source(s). See
       U(https://www.ansible.com/tower) for an overview.
 options:
-    inventory:
+    name:
       description:
-        - Name of the inventory that contains the inventory source(s) to update.
+        - The name or id of the inventory source to update.
       required: True
       type: str
-    inventory_source:
+    inventory:
       description:
-        - The name of the inventory source to update.
+        - Name or id of the inventory that contains the inventory source(s) to update.
       required: True
       type: str
     organization:
@@ -58,14 +58,14 @@ extends_documentation_fragment: awx.awx.auth
 EXAMPLES = '''
 - name: Update a single inventory source
   tower_inventory_source_update:
+    name: "Example Inventory Source"
     inventory: "My Inventory"
-    inventory_source: "Example Inventory Source"
     organization: Default
 
 - name: Update all inventory sources
   tower_inventory_source_update:
+    name: "{{ item }}"
     inventory: "My Other Inventory"
-    inventory_source: "{{ item }}"
   loop: "{{ query('awx.awx.tower_api', 'inventory_sources', query_params={ 'inventory': 30 }, return_ids=True ) }}"
 '''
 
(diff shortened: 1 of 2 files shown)


john-westcott-iv:
We should add aliases to these parameters so any existing playbooks continue to work.

nesanton:
Absolutely!
There you go. I only changed the `name` param, so bringing back `inventory_source` as an alias to `name`

nesanton:
@john-westcott-iv @AlanCoding would this work?

nesanton:
@beeankha

awx_collection/plugins/modules/tower_inventory_source_update.py
```diff
@@ -88,8 +90,8 @@
 def main():
     # Any additional arguments that are not fields of the item can be added here
     argument_spec = dict(
+        name=dict(required=True),
```
beeankha: This needs to have the alias info in it, like so:
`name=dict(required=True, aliases=['inventory_source']),`

awx_collection/tests/integration/targets/tower_inventory_source_update/tasks/main.yml
```diff
@@ -84,8 +84,8 @@
 
     - name: Test Inventory Source Update for All Sources
       tower_inventory_source_update:
+        name: "{{ item.name }}"
         inventory: "{{ inv_name }}"
-        inventory_source: "{{ item.name }}"
```
beeankha, line 88: For this integration test playbook task, it would be nice to have a check to make sure that the previously-implemented param of `inventory_source` still works, and state what we're doing in the task name.  Something like:

```
    - name: Test Inventory Source Update for All Sources (using inventory_source param vs name)
      tower_inventory_source_update:
        inventory_source: "{{ item.name }}"
```

awx_collection/plugins/modules/tower_inventory_source_update.py
nesanton: oh, right, added it to docstring and forgot the arg spec

awx_collection/tests/integration/targets/tower_inventory_source_update/tasks/main.yml
nesanton: Added!

rebeccahhh:
@nesanton what version of AWX is this?

nesanton:
> @nesanton what version of AWX is this?

collection v15.0.1 against Tower 3.7.3

beeankha (approved):
Thanks for doing this work, @nesanton !

Issue #7721, opened 2020-07-24
Update AddDropDownButton to show down caret


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
From feedback from @trahman73.

> Add a down caret to the Add button to show users that there will be additional choices.

Example:

<img width="935" alt="Screen Shot 2020-07-24 at 11 12 56 AM" src="https://user-images.githubusercontent.com/9053044/88408490-a03cf680-cda1-11ea-94bf-a0abad01445e.png">

tiagodread:
Tests in place PR already merged

PR #8351, opened 2020-10-08
Update AddDropDownButton to show down caret

nixocio:
Update AddDropDownButton to show down caret

See: https://github.com/ansible/awx/issues/7721

Before:

image

With this change:
image


jakemcdermott (approved):
:+1:

nixocio:
regate

jakemcdermott:
recheck

unlikelyzero:
@tiagodread 
1. let's add a visual check for the Instance Page
2. We need to determine if there are scenarios where a user can only ADD based on a permission. @mabashian @jakemcdermott do you know of one? Maybe WFJT permissions?
3. All cypress tests need to pass

tiagodread (approved):
👍🏽

Issue #6857, opened 2020-04-28
[ui_next] WF node preview, values for 'Detail__DetailName: Labels, Job Tags, Show Changes', needs to be marked for translations


sshedmake:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
With Japanese (dummy translation) google-chrome, values for 'Detail__DetailName: Labels, Job Tags, Show Changes' at WorkFlow Node preview screen, appears in English (instead of Japanese)

##### ENVIRONMENT
* AWX version: 10.0.0
* AWX install method: docker on linux
* Ansible version:  2.10.0.dev0
* Operating System: Fedora 31
* Web Browser: google-chrome

##### STEPS TO REPRODUCE
1. Do a successful login to the ui_next interface in the preferred language (say Japanese)
2. Navigate to any preconfigured WorkFlow Node preview screen via ~/templates/workflow_job_template/*/visualizer url path
3. Hover for 'WorkflowChart Tooltip Actions - node details'
4. To view node details, click on node preview icon with letter 'i'
5. Observe for respective values of 'Detail__DetailName: Labels, Job Tags, Show Changes'

##### EXPECTED RESULTS
Content should be in Japanese

##### ACTUAL RESULTS
2 more, Show Less, On|Off

##### ADDITIONAL INFORMATION
1. Screenshot higlighting values for 'Detail__DetailName: Labels, Job Tags, Show Changes', appearing in English, with Japanese google-chrome:

![output_wf_node-preview](https://user-images.githubusercontent.com/12151785/80512506-362bd580-899b-11ea-9f62-05b2e8fce0e1.gif)

unlikelyzero:
Putting into needs test to verify once pseudolocalization is available or real translations have arrived

mabashian:
PF does give us the ability to overwrite those chip strings with our own translated versions: http://patternfly-react.surge.sh/components/chip-group#chipgroup

The on/off looks like it's been marked: https://github.com/ansible/awx/blob/devel/awx/ui_next/src/components/PromptDetail/PromptDetail.jsx#L253

tiagodread:
These fields are now marked to be translated cc @mabashian 

![image](https://user-images.githubusercontent.com/12464858/118718498-b0750500-b7fd-11eb-87ea-6543b0aef2d0.png)

PR #8352, opened 2020-10-08
Mark missing words for translation

nixocio:
Add missing words for translation.

`...more`, and `Show Less` were already marked for translation in a
previous PR, since this code is shared as part of the `ChipGroup` code.

See: https://github.com/ansible/awx/issues/6857


marshmalien (approved):
💯

tiagodread (approved):

PR #8353, opened 2020-10-08
Zedr wfjt inventories

jbradberry:
##### SUMMARY
The current version of AWX fails to include the url to a non-null Inventory foreign key in the API results for a Workflow Job Template.  This PR finishes off work from @zedr to both add that link the the API output, but also attempt a reasonable fallback when using the awx-cli export function against a version of AWX that does not have this fix.

related #7798 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
awx: 15.0.0
```

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -3438,6 +3438,12 @@ class WorkflowJobTemplateSerializer(JobTemplateMixin, LabelsListMixin, UnifiedJo
             res['organization'] = self.reverse('api:organization_detail',   kwargs={'pk': obj.organization.pk})
         if obj.webhook_credential_id:
             res['webhook_credential'] = self.reverse('api:credential_detail', kwargs={'pk': obj.webhook_credential_id})
+        if obj.inventory:
+            res['inventory'] = self.reverse(
+                'api:inventory_detail', kwargs={
+                    'pk': obj.inventory.pk
+                }
+            )
         return res
 
     def validate_extra_vars(self, value):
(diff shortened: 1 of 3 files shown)


awx/api/serializers.py
```diff
@@ -3438,6 +3438,12 @@ def get_related(self, obj):
             res['organization'] = self.reverse('api:organization_detail',   kwargs={'pk': obj.organization.pk})
         if obj.webhook_credential_id:
             res['webhook_credential'] = self.reverse('api:credential_detail', kwargs={'pk': obj.webhook_credential_id})
+        if obj.inventory:
+            res['inventory'] = self.reverse(
+                'api:inventory_detail', kwargs={
+                    'pk': obj.inventory.pk
```
AlanCoding: could be more efficient to consistently use `obj.inventory_id` kind of syntax. It's the pattern I would generally prefer, not a blocker for this specifically, because current code is highly inconsistent.

jbradberry: Fair enough, I had that thought as well.

rebeccahhh (approved):

ryanpetrello:
@zedr you interested in giving this PR a whirl?

zedr:
Good for me!

squidboylan:
works for me, confirmed that the `related` field in the api view of a wfjt contains an `inventory` field

Issue #8013, opened 2020-08-26
Create a tower_workflow_approval module


john-westcott-iv:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Per PR #7963 we are missing a module to interact with the workflow_approvals endpoint

sean-m-sullivan:
@john-westcott-iv were you meaning just adding them like the above PR, or a module to actually approve a workflow via module?
If the former we can close this, if the later it still needs work.

yagomarques:
it's working, we use those playbooks point at an AWX machine.

deny scenario
```
- name: Wait for approval node to activate and approve
  hosts: localhost
  tasks:
    - awx.awx.tower_workflow_approval:
        workflow_job_id: "9"
        name: test1
        action: deny
```
![image](https://user-images.githubusercontent.com/18562328/120387177-f18e0e80-c2ff-11eb-8ee7-b99180c9df78.png)

after we launched another one passing an approved
 ```
name: Wait for approval node to activate and approve
  hosts: localhost
  tasks:
    - awx.awx.tower_workflow_approval:
        workflow_job_id: "11"
        name: test1
        action: approve
```

![image](https://user-images.githubusercontent.com/18562328/120387233-01a5ee00-c300-11eb-9d3a-9d292d5ab77b.png)
to finish we exec a job that take a long time and we put on our playbook a small timeout.
```
 name: Wait for approval node to activate and approve
  hosts: localhost
  tasks:
    - awx.awx.tower_workflow_approval:
        workflow_job_id: "14"
        name: test1
        timeout: 1
        action: approve
```
it's the result
```
TASK [awx.awx.tower_workflow_approval] **********************************************************************************************************************************
[WARNING]: You are using the awx version of this collection but connecting to Red Hat Ansible Tower
fatal: [localhost]: FAILED! => {"changed": false, "elapsed": null, "finished": null, "id": null, "msg": "Monitoring of Workflow Approval - test1 aborted due to timeout, workflow_jobs/19/workflow_nodes/", "started": null, "status": null}

PLAY RECAP **************************************************************************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0 
```
```

PR #8364, opened 2020-10-10
Workflow approval

sean-m-sullivan:
##### SUMMARY
Add Ability to add Approval nodes for workflow job template node module to resolve Issue #8013 
Added test for linking of workflow nodes. 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - Collection

##### AWX VERSION
```
awx: 15.0.0
```
Resolves Issue
 #8013


awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -337,7 +337,7 @@ def authenticate(self, **kwargs):
         # If we have neither of these, then we can try un-authenticated access
         self.authenticated = True
 
-    def delete_if_needed(self, existing_item, on_delete=None):
+    def delete_if_needed(self, existing_item, on_delete=None, on_continue=None):
```
john-westcott-iv: I'm not sure we want to call these `on_continue`. If I am reading this right, on_continue is really telling the function to just not auto-exit for you, is that correct? If so maybe you do something like auto_exit=True and then do something like:
```
if auto_exit:
    self.exit_json(**self.json_output)
```
You don't need to return self.json_output because the calling code can just do `module.json_output` to get at it.

The on_delete if accepting and calling a function if a delete happened, thus its name. On a delete of an object, call this additional function.

awx_collection/tests/integration/targets/tower_workflow_job_template/tasks/main.yml
```diff
@@ -145,6 +146,36 @@
...
+    that:
+      - "no_name_results.msg == 'Approval node name is required to create approval node.'"
+
+- name: Create approval node
+  tower_workflow_job_template_node:
+    identifier: approval_test
+    approval_node:
+      name: "{{ approval_node_name }}"
+      timeout: 900
+    workflow: "{{ wfjt_name }}"
```
john-westcott-iv, line 168: Can we also test a state == absent in here to make sure that is working?

awx_collection/plugins/modules/tower_workflow_job_template_node.py
```diff
@@ -217,7 +236,13 @@ def main():
     existing_item = module.get_one('workflow_job_template_nodes', **{'data': search_fields})
 
     if state == 'absent':
+        # Look up existing approval node for deletion
+        if existing_item['related'].get('unified_job_template') is not None:
+            existing_approval_node = module.get_endpoint(existing_item['related']['unified_job_template'])['json']
         # If the state was absent we can let the module delete it if needed, the module will handle exiting from this
+        # Delete the Approval Node
+        module.delete_if_needed(existing_approval_node, on_continue=True,)
```
john-westcott-iv: I haven't use approval nodes enough to know the mechanics... but if you delete the node the corresponding approval_node won't be auto-deleted?

awx_collection/plugins/modules/tower_workflow_job_template_node.py
```diff
@@ -93,6 +93,24 @@
         - Can be a job template, project, inventory source, etc.
         - Omit if creating an approval node (not yet implemented).
       type: str
+    approval_node:
```
john-westcott-iv, line 97: Should approval node and unifited_job_template be mutually exclusive? Or can you do both at the same time?

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -438,6 +446,15 @@ def create_if_needed(self, existing_item, new_item, endpoint, on_create=None, it
...
+            # 200 is response from approval node creation
+            elif response['status_code'] == 200 and item_type == 'workflow_job_template_approval_node':
+                self.json_output['name'] = 'unknown'
+                for key in ('name', 'username', 'identifier', 'hostname'):
+                    if key in response['json']:
+                        self.json_output['name'] = response['json'][key]
+                self.json_output['id'] = response['json']['id']
+                self.json_output['changed'] = True
+                item_url = response['json']['url']
             else:
```
john-westcott-iv: Can we handle this in the code after the on_continue? This seems very specific to workflow_job_template_approval_nodes and I don't think we want to fill the common library with specific one off behaviors.

awx_collection/plugins/module_utils/tower_api.py
sean-m-sullivan: Agree on the name chnage, and will look at return. Will do the module.json_output as well

awx_collection/plugins/modules/tower_workflow_job_template_node.py
sean-m-sullivan: It's hard to tell as you can't get a list of them, will do more research, this was abondence of caution like if you delete a project it doesn't delete a template.

awx_collection/plugins/modules/tower_workflow_job_template_node.py
sean-m-sullivan: Mutually exclusive, I did not test invoking both, but in theory the node would be created then converted to approval.

awx_collection/plugins/module_utils/tower_api.py
sean-m-sullivan: We could definitly put the logic after on continue, but the item_url still needs population and it will be different for the workflow approval, and its a check if it came back ok before attempting associations. This I'm not so sure on moving, as it logically makes sense to keep the block of 441-464 together.

awx_collection/plugins/modules/tower_workflow_job_template_node.py
sean-m-sullivan: Did the testing, this is not needed at all, reverting to original

awx_collection/tests/integration/targets/tower_workflow_job_template/tasks/main.yml
sean-m-sullivan: Added absent the node.

awx_collection/plugins/module_utils/tower_api.py
sean-m-sullivan: in terms of self.exit_json, that looks to be the norm for tower_api and is in 10 other places in the plugin. Right now just replicating the norm.

sean-m-sullivan:
regate

sean-m-sullivan:
Did some poking around and on a whim added the UnififedJobTemplate import and attempted to get the info, changed approval node name to unique and was able to get the tests to be successful and logical

beeankha (approved):
Thank you @sean-m-sullivan !

awx_collection/plugins/modules/tower_workflow_job_template_node.py
```diff
@@ -264,10 +291,42 @@ def main():
...
+        # Find created workflow node ID
+        search_fields = {'identifier': identifier}
+        search_fields['workflow_job_template'] = workflow_job_template_id
+        workflow_job_template_node = module.get_one('workflow_job_template_nodes', **{'data': search_fields})
+        workflow_job_template_node_id = workflow_job_template_node['id']
+        module.json_output['workflow_node_id'] = workflow_job_template_node_id
+        # Due to not able to lookup workflow_approval_templates, find the existing item in another place
+        if workflow_job_template_node['related'].get('unified_job_template') is not None:
+            existing_item = module.get_endpoint(workflow_job_template_node['related']['unified_job_template'])['json']
+        approval_endpoint = 'workflow_job_template_nodes/' + str(workflow_job_template_node_id) + '/create_approval_template/'
```
john-westcott-iv: I don't think this is wrong but stylistically we have been using the format function like:
`approval_endpoint = 'workflow_job_template_nodes/{0}/create_approval_template/'.format(workflow_job_template_node_id)`

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -375,7 +379,10 @@ def delete_if_needed(self, existing_item, on_delete=None):
                 else:
                     self.fail_json(msg="Unable to delete {0} {1}: {2}".format(item_type, item_name, response['status_code']))
         else:
-            self.exit_json(**self.json_output)
+            if auto_exit:
+                self.exit_json(**self.json_output)
+            else:
+                return None
```
john-westcott-iv: Is there anyway we can standardize on what gets returned if you are doing `auto_exit=False`? We have return self.json_output, None and I think another two options below. If we can't standardize on a return can we just return nothing and then have methods to get at the required data from the TowerAPIModule directly? i.e. for json_output you can reference `module.json_output` from the calling module.

awx_collection/plugins/module_utils/tower_api.py
john-westcott-iv: I guess I am also curious why this endpoint is returning a 200 instead of a 201. That is probably more of a Tower question than a collection question but it's breaking our "pattern" of expecting a 201 to be returned for something new. @ryanpetrello any idea why this endpoint would return a 200 instead of a 201 for an on create? Any idea if there are other endpoint that return a 200 instead of a 201?

We have, on other functions, done things like `if response['status_code'] in [200, 201]:` if the return code varies. Or, if we really need to be more specific for this use case could we do something like:
```
if response['status_code'] == 201  or (response['status_code'] == 200 and item_type == 'workflow_job_template_approval_node'):
```

That would eliminate redundant code (let me know if I am misreading and if and elif are not identical code).

ryanpetrello: Generally speaking, our endpoints that *create* things should be returning HTTP 200.  HTTP 201 also isn't uncommon, but it's generally reserved for situations where we've created something that's still pending, like job launches:

```
awx/api/generics.py
626:        return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
664:            if response.status_code != status.HTTP_201_CREATED:
698:            return Response(data, status=status.HTTP_201_CREATED, headers=headers)
1024:        return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)

awx/api/views/webhooks.py
47:        return Response({'webhook_key': obj.webhook_key}, status=status.HTTP_201_CREATED)

awx/api/views/__init__.py
2431:            return Response(data, status=status.HTTP_201_CREATED, headers=headers)
2887:        return Response(status=status.HTTP_201_CREATED, headers=headers)
3222:        return Response(data, status=status.HTTP_201_CREATED, headers=headers)
3254:        return Response(data, status=status.HTTP_201_CREATED, headers=headers)
3452:        return Response(data, status=status.HTTP_201_CREATED, headers=headers)
3640:            return Response(data, status=status.HTTP_201_CREATED, headers=headers)
3698:        return Response(data, status=status.HTTP_201_CREATED, headers=headers)
3878:        if response.status_code != status.HTTP_201_CREATED:
3985:            return Response(data, status=status.HTTP_201_CREATED, headers=headers)
```

ryanpetrello: Hrm...

ryanpetrello: This is probably why:

image

Should we just change this API code to be a 201?

ryanpetrello: Here ya go, let's do this instead: https://github.com/ansible/awx/pull/8416

sean-m-sullivan: I think that would work, I was wondering it myself @john-westcott-iv the 200+item was more being overly cautious, @ryanpetrello working on new build with John's suggestions where I change this to 201

awx_collection/plugins/module_utils/tower_api.py
sean-m-sullivan: changed to return self.json_output , since we are adding things to it as we go, and then can be consistent, unless you were thinking of something different?

kedark3:
@sean-m-sullivan this will be a great addition in our tower_configuration repo as well later. Thanks for this work.

john-westcott-iv (approved):

Issue #5685, opened 2020-01-16
Add "external source" enhancements to Users list, detail, and form


marshmalien:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Users List
- Denote that a user comes from an external source (ie LDAP) as a tag next to the username

<img width="401" alt="Screen Shot 2020-01-16 at 4 47 33 PM" src="https://user-images.githubusercontent.com/15881645/72565484-f6bffa00-387f-11ea-99c7-597eebe54354.png">

User Detail
- Denote that a user comes from an external source (ie LDAP) as a tag under Type detail label
<img width="1198" alt="Screen Shot 2020-01-16 at 4 48 56 PM" src="https://user-images.githubusercontent.com/15881645/72565639-4bfc0b80-3880-11ea-95ff-0c67080537bc.png">

User Form
- Username field is required if user is not from an external source
- Password field should be hidden if user is from external source (ldap/radius/social auth users)
For reference, see [angular tower users form](https://github.com/ansible/awx/blob/devel/awx/ui/client/src/users/users.form.js#L62)

tiagodread:
PR already merged, closing this issue

PR #8369, opened 2020-10-12
Add general updates for User screen

nixocio:
* Add type of login used as part of UserListItem.
* Add type of login used as part of UserDetail.
* Hide password field, UserForm, in case login method is LDAP or Social.
* Make username field, UserForm, not required in case login is LDAP or
Social.

See: https://github.com/ansible/awx/issues/5685

### DIFF
--- a/awx/ui_next/src/screens/User/UserDetail/UserDetail.jsx
+++ b/awx/ui_next/src/screens/User/UserDetail/UserDetail.jsx
@@ -3,7 +3,7 @@ import { Link, useHistory } from 'react-router-dom';
 import { withI18n } from '@lingui/react';
 import { t } from '@lingui/macro';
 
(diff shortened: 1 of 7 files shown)


mabashian (approved):
Checked this out and compared it to the old app.  The feature parity looks good to me.

The only thought I had was that maybe we could expose this attribute:

```
            "auth": [
                {
                    "provider": "github",
                    "uid": "9999999"
                }
            ]
```

in the details view of the user.

Because this is an array, I assume multiple values can exist in here but I don't actually know that for a fact.

We could include that work in this PR _or_ create a new enhancement issue for it.  I'll leave that up to you @nixocio

nixocio:
hey @mabashian, I like that idea, I will file an enhancement issue for it.

tiagodread (approved):
👍🏽

Issue #8359, opened 2020-10-09
awx cli extra_vars not working


Besstheone:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
awxkit (version 15.0.0) not accepting extra_vars on prompt but working with files

##### ENVIRONMENT
* AWX version: 15.0.0
* AWX install method: docker on linux
* Ansible version:  2.9.13
* Operating System:
* Web Browser:

##### STEPS TO REPRODUCE

call awx cli with extravars
![image](https://user-images.githubusercontent.com/39663221/95584991-773e1080-0a36-11eb-8628-a9b707ede94b.png)

##### EXPECTED RESULTS
job launched
##### ACTUAL RESULTS

![image](https://user-images.githubusercontent.com/39663221/95585011-81f8a580-0a36-11eb-96f1-0940460c54d4.png)

ryanpetrello:
Hey @Besstheone,

The `--extra-vars` argument is supposed to be valid JSON/YAML, not Ansible-style extra vars.  I'll open a PR that clarifies this a bit (instead of this confusing error message).

kdelee:
@elyezer happened to be in the neighborhood and tried this out, I think we can close it

This command:
```
awx job_templates launch --extra_vars vm_name=foo
```

makes this error come out:
```
awx job_templates launch: argument --extra_vars: vm_name=foo is not valid JSON or YAML
```

PR #8370, opened 2020-10-12
fix a confusing error message when extra vars aren't valid JSON/YAML

ryanpetrello:
see: https://github.com/ansible/awx/issues/8359

### DIFF
--- a/awxkit/awxkit/cli/options.py
+++ b/awxkit/awxkit/cli/options.py
@@ -178,6 +178,9 @@ class ResourceOptionsParser(object):
                     except Exception:
                         raise argparse.ArgumentTypeError("{} is not valid JSON or YAML".format(v))
 
+                if not isinstance(parsed, dict):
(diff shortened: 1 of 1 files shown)


awxkit/awxkit/cli/options.py
```diff
@@ -178,6 +178,9 @@ def json_or_yaml(v):
                     except Exception:
                         raise argparse.ArgumentTypeError("{} is not valid JSON or YAML".format(v))
 
+                if not isinstance(parsed, dict):
+                    raise argparse.ArgumentTypeError("{} is not valid JSON or YAML".format(v))
```
AlanCoding, line 182: It might be nice to add some verbiage saying that it parsed, but isn't a dictionary. It's good to raise this error, but the next steps for the user are very unclear. It's hard enough for ourselves when we're already used to these quirks.

non-blocking comment

AlanCoding (approved):
The old "YAML document is a string" strikes again

Issue #8374, opened 2020-10-12
Link workflow approval node to approval details in output


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
In workflows with approval steps we should navigate the user to the details view of the approval when they click on the node in the output:

<img width="1622" alt="Screen Shot 2020-10-12 at 2 22 03 PM" src="https://user-images.githubusercontent.com/9889020/95778602-786d7880-0c96-11eb-9f50-f30e74c117db.png">

should take the user to:

<img width="1680" alt="Screen Shot 2020-10-12 at 2 23 02 PM" src="https://user-images.githubusercontent.com/9889020/95778619-7e635980-0c96-11eb-802a-e28b49eff0dd.png">

This would be consistent with our behavior for other node types.

mabashian:
Blocked until #8375 lands

akus062381:
Resolved

Issue #8373, opened 2020-10-12
Add a workflow approvals list


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
API gives us `/api/v2/workflow_approvals` just like many other lists.  We can use this to build our familiar list + details pattern alongside https://github.com/ansible/awx/issues/7112.

Additionally, we can use this as the interface for approving/denying workflow approvals rather than having a one-off drawer.

This list should feature the following things:

1) Show the workflow approval name along with a link to the approval details view
2) Show the job name along with a link to the job that contains this workflow approval
3) The status of the approval if !pending
4) If/When the approval expires
5) Buttons to approve/deny the approval when rbac permits
6) Button to delete the approval when rbac permits

tiagodread:
Closing this after merging #8375 tests already merged.

Issue #8372, opened 2020-10-12
Add pending workflow approval count/link to page header


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
When we added this feature to the old UI, the entry point for approving/denying workflow approvals was here in the header:

<img width="1680" alt="Screen Shot 2020-10-12 at 2 02 00 PM" src="https://user-images.githubusercontent.com/9889020/95777155-a69d8900-0c93-11eb-8515-0e7a06182589.png">

We would listen for workflow approval job websocket messages and update the count by [GETting `/api/v2/workflow_approvals?status=pending&page_size=1`](https://github.com/ansible/awx/blob/devel/awx/ui/client/src/app.js#L220-L232)

Clicking on this link _used_ to open up a drawer from the right hand side where we displayed the pending approvals.

I think that this button should now link the user to `/workflow_approvals?workflow_approvals.status=pending` so that they see:

<img width="1432" alt="Screen Shot 2020-10-12 at 2 09 03 PM" src="https://user-images.githubusercontent.com/9889020/95777603-83bfa480-0c94-11eb-93a0-c7a83906d534.png">

mabashian:
Blocked until #8375 lands

mabashian:
Alternatively this could go in the side nav

<img width="284" alt="Screen Shot 2021-02-03 at 4 26 01 PM" src="https://user-images.githubusercontent.com/9889020/106811609-95aac600-663c-11eb-859d-a6c3714f7e78.png">

and we wouldn't have to add anything to the top nav

Issue #8273, opened 2020-09-30
Add Workflow Approvals to ui_next


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
_Intentionally leaving this blank as I explore options for this_

mabashian:
OK, I've decided to split this work up into a few different issues:

1) [Add pending workflow approval count/link to page header](https://github.com/ansible/awx/issues/8372)
2) [Add a 'details' view for a completed approval node](https://github.com/ansible/awx/issues/7112).  I'd actually expand this out to just a generic 'details' view for all workflow approvals.
3) [Add a workflow approvals list](https://github.com/ansible/awx/issues/8373)
4) [Link workflow approval node to approval details in output](https://github.com/ansible/awx/issues/8374)

sugitk:
I saw a new Web UI for Workflow Approvals, I would like to know how many approvals are pending in the queue.  If possible, can you append a number of them in this menu?

<img width="439" alt="スクリーンショット 2020-12-13 20 31 35" src="https://user-images.githubusercontent.com/2234991/102010735-61750080-3d83-11eb-9034-0e4308e9037f.png">

unlikelyzero:
@sugitk please follow https://github.com/ansible/awx/issues/8372 for updates

mabashian:
Closing this issue since the work has already been split into smaller issues.

PR #8375, opened 2020-10-12
Adds workflow approval list and details

mabashian:
##### SUMMARY
Meta issue https://github.com/ansible/awx/issues/8273
Addresses #8373 
Sort of addresses https://github.com/ansible/awx/issues/7112 (the rest of it captured in https://github.com/ansible/awx/issues/8374)
Does not include https://github.com/ansible/awx/issues/8372

This PR adds a new list `/workflow_approvals` along with a details view for items in that list.

The entry point for this list is here:

Screen Shot 2020-10-12 at 2 36 43 PM

Since this list does not exist in the old UI I figured this was as good a spot as any.  We can move it if we need to.

The list itself is pretty straightforward but here's an explanation of the columns:

Screen Shot 2020-10-12 at 2 45 12 PM

Screen Shot 2020-10-12 at 2 47 18 PM

Screen Shot 2020-10-12 at 2 50 26 PM

Some examples of tooltip on status:

Screen Shot 2020-10-12 at 2 51 00 PM
Screen Shot 2020-10-12 at 2 51 06 PM

Screen Shot 2020-10-12 at 2 54 38 PM

Here's a look at the details view for a pending approval:

Screen Shot 2020-10-12 at 2 55 15 PM

a lot of the same information from the list row is shown here.

Here's a look at the details view for an approved approval:

Screen Shot 2020-10-12 at 2 56 11 PM

The action buttons/rbac should be the same on the details view as they are on the list.

**Websockets**

OK so here's how websockets work on this list.  This list is subscribed to the jobs -> status_changed channel just like the jobs list.  We're only interested in jobs with the type: `workflow_approval` though.  The response to events that match our criteria is pretty straightforward: we refresh the list contents.  This is throttled but that throttle is currently set to just 1 second.  Unlike the jobs list, I'm not anticipating a tremendous amount of activity on this list in a short amount of time but if we need to increase that throttle threshold then we can.  Here's when we refresh the list:

1) When the job in question _is not_ in our list _and_ it's status is `pending`.
2) When the job in question _is_ in our list _and_ it's status is not in `['new', 'pending', 'waiting', 'running']` (i.e. the job finished).

^^ Scenario 2 creates an interesting challenge when the user that's viewing the list is also the user that is acting on the row.  When the user clicks approve/deny and the API responds with a 204 there's a bit of a delay before the socket message comes through and tells us that the job has moved to a completed status.  During that time, I hide the action buttons via a boolean on the list row state.

**Websockets are where I spent a majority of my time on this feature and I'm still not sure that it's quite right.  I'm definitely open to alternative approaches to keeping this list up to date using the data that we get from the job status channel.**

Unfortunately, we have to hit the API to retrieve the row data as the socket message doesn't have much more than the status and the job id.  There could be a solution where we fetch specific rows instead of refetching all of them.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI


wenottingham:
Looks generally good. I do think we need to move the list - workflow approvals are the same conceptual things as job runs, and so should be in that area.

unlikelyzero:
@tiagodread note: This is net new and not absolutely critical for this week.

mabashian:
OK, made the following updates based on feedback from Taufique and Bill.

1) Moved Workflow Approvals out of Administration and in to Views
2) Moved approve/deny buttons from the individual list row items up to the list action level (alongside delete).  This means multiple rows can be approved/denied at the same time.
3) Updated the UX on the status labels to match the mockups: https://tower-mockups.testing.ansible.com/patternfly/workflow-approvals/wf-approvals/

jakemcdermott (commented):
- List item, approval button, new view, and tests, etc. LGTM
- let's double-check w/ @keithjgrant on the websocket hook usage

awx/ui_next/src/screens/WorkflowApproval/shared/WorkflowApprovalStatus.jsx
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import { withI18n } from '@lingui/react';
+import { t } from '@lingui/macro';
+import { Label, Tooltip } from '@patternfly/react-core';
+import { CheckIcon, InfoCircleIcon } from '@patternfly/react-icons';
+import { WorkflowApproval } from '../../../types';
+import { formatDateString } from '../../../util/dates';
+
+function WorkflowApprovalStatus({ workflowApproval, i18n }) {
```
keithjgrant, line 9: We have a generic status label component at `components/StatusLabel`. Do you think it could meet your needs here instead of a new component?

awx/ui_next/src/screens/WorkflowApproval/WorkflowApprovalDetail/WorkflowApprovalDetail.jsx
```diff
@@ -0,0 +1,246 @@
...
+          isOpen={deleteError}
+          variant="error"
+          title={i18n._(t`Error!`)}
+          onClose={dismissDeleteError}
+        >
+          {i18n._(t`Failed to delete workflow approval.`)}
+          <ErrorDetail error={deleteError} />
+        </AlertModal>
+      )}
+      {approveError && (
```
AlexSCorey, line 231: You could combine these 2 AlertModals, and would you also only need to have 1 useDismissableError above, which would be something like
```const {error, dismissError} = useDismissableError(approveError || denyError)```
and then the alertModals would be
```<AlertModal
          isOpen={error}
          variant="error"
          title={i18n._(t`Error!`)}
          onClose={dismissError}
        >
         approveError ?  {i18n._(t`Failed to approve workflow approval.`)} 
            : {i18n._(t`Failed to deny workflow approval.`)}
          <ErrorDetail error={approveError} />
        </AlertModal>
```

awx/ui_next/src/screens/WorkflowApproval/WorkflowApprovalList/WorkflowApprovalList.jsx
```diff
@@ -0,0 +1,277 @@
...
+      relatedSearchableKeys: [],
+      searchableKeys: [],
+    }
+  );
+
+  useEffect(() => {
+    fetchWorkflowApprovals();
+  }, [fetchWorkflowApprovals]);
+
+  // TODO: update QS_CONFIG to be safe for deps array
```
AlexSCorey: Is this a note for your current work that you forgot to remove, or is this future work?

awx/ui_next/src/screens/WorkflowApproval/WorkflowApprovalList/WorkflowApprovalList.test.jsx
```diff
@@ -0,0 +1,317 @@
+import React from 'react';
+import { act } from 'react-dom/test-utils';
+import { WorkflowApprovalsAPI } from '../../../api';
+import { mountWithContexts } from '../../../../testUtils/enzymeHelpers';
+import WorkflowApprovalList from './WorkflowApprovalList';
+
+jest.mock('../../../api');
+
+const mockWorkflowApprovals = [
```
AlexSCorey: Might consider putting this inside of a data.json file to reduce the # of lines of code in this file

awx/ui_next/src/screens/WorkflowApproval/WorkflowApprovalList/useWsWorkflowApprovals.js
```diff
@@ -0,0 +1,61 @@
+import { useState, useEffect } from 'react';
```
AlexSCorey, line 1: was does the Ws in `useWsWorkflowApprovals` stand for?

awx/ui_next/src/screens/WorkflowApproval/WorkflowApprovalList/WorkflowApprovalList.jsx
keithjgrant: that's an old TODO of mine, copied over from somewhere else I think

tiagodread:
Hey @mabashian !
- I tried to delete a workflow approval before approving or deny and I got this result (on the list or details):

approval

- The simple search using name doesn’t work:
search

- If i delete the job the workflow approval associated with that job show `undefined` for the job name and if I click on it I got 404:

deleted

awx/ui_next/src/screens/WorkflowApproval/WorkflowApprovalList/WorkflowApprovalListApproveButton.jsx
```diff
@@ -0,0 +1,76 @@
...
+          onClick={onApprove}
+        >
+          {i18n._(t`Approve`)}
+        </DropdownItem>
+      ) : (
+        <Tooltip content={renderTooltip()} position="top">
+          <div>
+            <Button
+              isDisabled={isDisabled}
+              aria-label={i18n._(t`Approve`)}
```
unlikelyzero, line 54: do aria-labels need to be translated?

unlikelyzero:
We have tests pending the merge of this PR. Please let us know when it's ready for final merge

tiagodread:
@mabashian can you add some uniques ids for the data on datails tab specially for the status

awx/ui_next/src/screens/WorkflowApproval/shared/WorkflowApprovalStatus.jsx
mabashian: 👍 will check that out

awx/ui_next/src/screens/WorkflowApproval/WorkflowApprovalList/WorkflowApprovalList.test.jsx
mabashian: 👍

awx/ui_next/src/screens/WorkflowApproval/WorkflowApprovalList/useWsWorkflowApprovals.js
mabashian: That's the naming pattern we have for all of our websocket hooks.  They all start with `useWs` and then we concat the resource on the end.

awx/ui_next/src/screens/WorkflowApproval/shared/WorkflowApprovalStatus.jsx
mabashian: Yea there's definitely some overlap here.  If we want to keep the same UX then I'd have to refactor this component quite a bit with several conditionals.  I'm a little reluctant to do that given how clean this component (StatusLabel) is.

mabashian:
OK @tiagodread @unlikelyzero I think this is ready for your eyes again.  I've added data-cy attributes to all the details: https://github.com/ansible/awx/pull/8375/files#diff-eeea31a09935ea3101ae734da15ca14d933119b4bb0819194fb9ff24a074268cR85-R183

unlikelyzero:
@mabashian hey, we're not longer using data-cy attributes but trying to use the dataouia attributes instead. Is there enough to glean from this PR?

https://github.com/ansible/awx/pull/8472#discussion_r517504241

tiagodread:
Hey @mabashian looks like the error `- I tried to delete a workflow approval before approving or deny and I got this result (on the list or details)` was fixed for the list, but, i can try to delete a pending workflow approval in the details page:

wfaproval

tiagodread:
@mabashian awesome job!!!

Tests covered by a mix of manual and automated test cases in here https://github.com/ansible/tower-qa/pull/5656

tiagodread (approved):

mabashian:
@keithjgrant @jakemcdermott @AlexSCorey any other thoughts on this one before merge?

jakemcdermott (approved):

PR #8386, opened 2020-10-13
Fixing resetting of variables and adding unit test for condition

### LINKED ISSUE #8338: tower_job_template.py
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Similar to our issue #7766, we found that this module ignores setting any job template Boolean or Integer parameter values, to its default value.  For example the following parameters can no longer be set as Zero or False after it has been modified via GUI or Tower_Module

        timeout: 3600
        allow_simultaneous: true
        ask_limit_on_launch: true

We only tested a handful of parameters but expect that they all behave in the same way.  We request that all parameters for this module be fully tested

##### ENVIRONMENT

- AWX version: 13.0.0
- AWX install method: we developed an Ansible script to install
- Ansible version: 2.9.11
- Operating System: RHEL Linux 7
- Web Browser: Chromium Edge

##### STEPS TO REPRODUCE

Set any parameter value to something other than its default via GUI or tower_job_template module and then try to set the value back to its default value using tower_job_template module.

##### EXPECTED RESULTS

Module should follow the behavior of AWX GUI and allow parameters to be changed back to their default values

##### ACTUAL RESULTS

Parameter value remains as previously set and will not change back to its default value as directed by tower_job_template payload parameters

##### ADDITIONAL INFORMATION

none

john-westcott-iv:
##### SUMMARY

Fixes #8338 

##### ISSUE TYPE

  - Bugfix Pull Request
 
##### COMPONENT NAME

 - Collection

##### AWX VERSION

```
awx: 15.0.0
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx_collection/test/awx/test_job_template.py
+++ b/awx_collection/test/awx/test_job_template.py
@@ -36,6 +36,53 @@ def test_create_job_template(run_module, admin_user, project, inventory):
 
 
 @pytest.mark.django_db
+def test_resete_job_template_values(run_module, admin_user, project, inventory):
+
+    module_args = {
+        'name': 'foo', 'playbook': 'helloworld.yml',
+        'project': project.name, 'inventory': inventory.name,
+        'extra_vars': {'foo': 'bar'},
+        'job_type': 'run',
+        'state': 'present',
+        'forks': 20,
+        'timeout': 50,
+        'allow_simultaneous': True,
+        'ask_limit_on_launch': True,
+    }
+
+    result = run_module('tower_job_template', module_args, admin_user)
+
+    jt = JobTemplate.objects.get(name='foo')
+    assert jt.forks == 20
+    assert jt.timeout == 50
+    assert jt.allow_simultaneous
+    assert jt.ask_limit_on_launch
+
+    module_args = {
+        'name': 'foo', 'playbook': 'helloworld.yml',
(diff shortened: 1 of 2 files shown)


awx_collection/test/awx/test_job_template.py
```diff
@@ -35,6 +35,53 @@ def test_create_job_template(run_module, admin_user, project, inventory):
     assert jt.inventory_id == inventory.id
 
 
+@pytest.mark.django_db
+def test_resete_job_template_values(run_module, admin_user, project, inventory):
```
beeankha: Minor typo: should be `reset` vs `resete`

awx_collection/test/awx/test_job_template.py
```diff
@@ -35,6 +35,53 @@ def test_create_job_template(run_module, admin_user, project, inventory):
...
+
+    jt = JobTemplate.objects.get(name='foo')
+    assert jt.forks == 0
+    assert jt.timeout == 0
+    assert not jt.allow_simultaneous
+    assert not jt.ask_limit_on_launch
+
+
+
+
```
beeankha: Another minor edit: there should be only 2 blank lines here vs 4

beeankha (approved):
Made some minor edit suggestions, otherwise looks great!  👍

PR #8394, opened 2020-10-15
centralize reusable profiling code

chrismeyersfsu:
see doc updates for usage

Verified that `TimingMiddleware` continues to work.
image

### DIFF
--- a/awx/main/middleware.py
+++ b/awx/main/middleware.py
@@ -1,13 +1,9 @@
 # Copyright (c) 2015 Ansible, Inc.
 # All Rights Reserved.
 
-import uuid
 import logging
 import threading
 import time
-import cProfile
-import pstats
-import os
 import urllib.parse
 
 from django.conf import settings
@@ -22,6 +18,7 @@ from django.urls import reverse, resolve
 
 from awx.main.utils.named_url_graph import generate_graph, GraphNode
 from awx.conf import fields, register
+from awx.main.utils.profiling import AWXProfiler
 
 
 logger = logging.getLogger('awx.main.middleware')
@@ -32,11 +29,14 @@ class TimingMiddleware(threading.local, MiddlewareMixin):
 
     dest = '/var/log/tower/profile'
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.prof = AWXProfiler("TimingMiddleware")
+
     def process_request(self, request):
         self.start_time = time.time()
         if settings.AWX_REQUEST_PROFILE:
-            self.prof = cProfile.Profile()
-            self.prof.enable()
+            self.prof.start()
 
     def process_response(self, request, response):
         if not hasattr(self, 'start_time'):  # some tools may not invoke process_request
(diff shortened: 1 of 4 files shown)


shanemcd:
recheck

awx/main/middleware.py
```diff
@@ -32,45 +29,28 @@ class TimingMiddleware(threading.local, MiddlewareMixin):
 
     dest = '/var/log/tower/profile'
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.prof = AWXProfiler("TimingMiddleware")
+
     def process_request(self, request):
         self.start_time = time.time()
+        print(f"Setting is {settings.AWX_REQUEST_PROFILE}")
```
ryanpetrello: 👃

awx/main/utils/profiling.py
```diff
@@ -0,0 +1,148 @@
...
+            self.generate_results()
+            res = self.output_results()
+            self.started = False
+            return res
+        else:
+            print("AWXProfiler::stop() called without calling start() first", file=sys.stderr)
+            return None
+
+
+def profile(name, *init_args, **init_kwargs):
```
ryanpetrello, line 140: ❤️

awx/main/middleware.py
```diff
@@ -32,45 +29,28 @@ class TimingMiddleware(threading.local, MiddlewareMixin):
...
         if settings.AWX_REQUEST_PROFILE:
-            self.prof = cProfile.Profile()
-            self.prof.enable()
+            print("Starting it")
```
ryanpetrello: Do we need these print statements?

awx/main/middleware.py
```diff
@@ -32,45 +29,28 @@ class TimingMiddleware(threading.local, MiddlewareMixin):
...
         if settings.AWX_REQUEST_PROFILE:
-            self.prof.disable()
-            cprofile_file = self.save_profile_file(request)
-            response['cprofile_file'] = cprofile_file
+            response['cprofile_file'] = self.prof.stop()
         perf_logger.info('api response times', extra=dict(python_objects=dict(request=request, response=response)))
         return response
 
-    def save_profile_file(self, request):
-        if not os.path.isdir(self.dest):
```
ryanpetrello, line 54: Nice to lose all this.

awx/main/utils/profiling.py
```diff
@@ -0,0 +1,148 @@
...
+import functools
+import pstats
+import os
+import uuid
+import datetime
+import json
+import sys
+
+
+class AWXBase:
```
ryanpetrello: Better name, maybe `AWXProfileBase` ?

docs/debugging.md
```diff
@@ -98,3 +98,81 @@ sdb-listen
...
+
+Now, invoke the function being timed. Each run of the timed function will result
+in a file output to `dest`. The timing data will be in the file name.
+
+```
+bash-4.4$ ls -al /var/log/tower/timing/
+total 16
+drwxr-xr-x 2 awx  root 4096 Oct 15 13:25  .
+drwxrwxr-x 1 root root 4096 Oct 15 13:25  ..
+-rw-r--r-- 1 awx  root   74 Oct 15 13:25 '0:00:02.002159-seconds-my_task_manager_timing-e8c03054-7ea3-43ea-902e-b5f3561cbae7.time'
```
ryanpetrello: Why are there single quotes around this filename?

docs/debugging.md
```diff
@@ -98,3 +98,81 @@ sdb-listen
...
+will result in a file output to `dest` containing the profile data summary as
+well as a dot graph if enabled. The profile data summary can be viewed in a
+text editor. The dot graph can be viewed using `xdot`.
+
+```
+bash-4.4$ ls -al /var/log/tower/profile/
+total 24
+drwxr-xr-x 2 awx  root 4096 Oct 15 13:23 .
+drwxrwxr-x 1 root root 4096 Oct 15 13:23 ..
+-rw-r--r-- 1 awx  root  635 Oct 15 13:23 task_manager_profile-2.001s-2303-272858af-3bda-45ec-af9e-7067aa86e4f3.dot
```
ryanpetrello: One nice thing about the old format is that it _started_ with the timing, so you could sort the filenames with `ls -n` to see the slowest/fastest requests in order.  Might want to change this back to e.g.,

`2.00s-task_manager_profile-<UUID>.dot`

ryanpetrello (commented):
I like the way this is going.  Are you going to update the callback receiver to make use of this refactored code (instead of its own cProfile usage)?

docs/debugging.md
chrismeyersfsu: Will double check. I think it's just my zsh terminal

chrismeyersfsu:
image
* confirmed, `'filename'` is because of my shell
* Moved duration to front of filename
* fixed Timing to output `total_seconds()` rather than `hh:mm:ss`
* removed debug print statements
* renamed base class
* replaced callback receiver profiler with this one

awx/main/middleware.py
```diff
@@ -32,45 +29,25 @@ class TimingMiddleware(threading.local, MiddlewareMixin):
...
     def process_response(self, request, response):
         if not hasattr(self, 'start_time'):  # some tools may not invoke process_request
             return response
         total_time = time.time() - self.start_time
         response['X-API-Total-Time'] = '%0.3fs' % total_time
         if settings.AWX_REQUEST_PROFILE:
-            self.prof.disable()
-            cprofile_file = self.save_profile_file(request)
-            response['cprofile_file'] = cprofile_file
+            response['cprofile_file'] = self.prof.stop()
```
ryanpetrello: If we want to return this in the HTTP response, we should make this follow the convention of other custom headers, `X-API-Profile-File`

ryanpetrello (approved):
Thanks for cleaning up my mess.

PR #8398, opened 2020-10-15
Fixing stack trace from an error not having a detail

john-westcott-iv:
##### SUMMARY

Was attempting to add an SCM credential to a job template and got a stack trace. The API was returning:
```
{'error': 'Cannot assign a Credential of kind `scm`.'}
```

Which did not contain a "detail" field. This change attempts to return 'detail' but if that DNE than just send back the entire json response instead of generating a stack. In the future I want to create a `get_error` method which would look for all sorts of things.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - Collections

##### AWX VERSION

```
awx: 15.0.0
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx_collection/plugins/module_utils/tower_api.py
+++ b/awx_collection/plugins/module_utils/tower_api.py
@@ -393,7 +393,7 @@ class TowerAPIModule(TowerModule):
             if response['status_code'] == 204:
                 self.json_output['changed'] = True
             else:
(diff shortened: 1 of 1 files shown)


beeankha (approved):
Thank you for fixing these up, @john-westcott-iv  !

Issue #8200, opened 2020-09-22
Enforce credential replacement rules on job launch prompt


mabashian:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
When launching a job template/configuring a schedule/configuring a workflow node where the job template prompts for credentials, a user should not be able to remove a credential without replacing it with one of the same type.  For example, if a job template has a default machine credential, the user must use that default or replace it with another machine credential.  This applies to all credential types.  Vault credentials have slightly different rules since a user can specify multiple default vault credentials.  If a job template has a default vault credential with ID: `foo` then the user must use that default or replace it with another vault credential with ID: `foo`.

In the old UI we showed this message on the prompt step if a user removed one of the default credentials:

<img width="780" alt="Screen Shot 2020-09-22 at 12 39 26 PM" src="https://user-images.githubusercontent.com/9889020/93911688-b64b3280-fcd0-11ea-8782-997a0c656842.png">

<img width="822" alt="Screen Shot 2020-09-22 at 12 40 11 PM" src="https://user-images.githubusercontent.com/9889020/93911755-cd8a2000-fcd0-11ea-9bb2-0e89ae33f361.png">

We should probably consider showing something similar to let the user know what the expectation is.

unlikelyzero:
Note: we will need to add this to our coverage

mabashian:
Same thing applies to workflow nodes and schedules.

mabashian:
Wait until #8399 merges

akus062381:
coverage added in 
https://github.com/ansible/tower-qa/pull/5871

Issue #5913, opened 2020-02-11
Support prompting on workflow nodes


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
When a user selects a JT or WFJT with promptable fields while creating or editing a workflow node we want to give that user the ability to fill in the prompts right there in the wizard.  These fields have to be provided with the node and will not be prompted for at launch time.

There's a significant amount of overlap with #5909 and #5910 so hopefully we can re-use some of those same steps here.

Like JT schedules, credentials that require passwords cannot be used in a WF node job.  We don't store those passwords and they can't be provided at launch time so the user will need to _replace_ those credentials if they are present on the JT that's been selected as a node.

Some extra text needs to be added to the inventory step when it's promptable.  The global WFJT Inventory will trump and node level inventory that's provided.  We want to communicate that to the user.  This is probably only necessary if ask_inventory_on_launch is set to true on the outer workflow.

The preview step will differ a little bit from the launch preview step because we'll want to show the JT/WFJT as well as the edge type (success/failure/always).  Probably also need to show something about it's convergence state in the preview as well.

Going to set this as blocked until #5909 and #5910 land.

Link to mockups: https://tower-mockups.testing.ansible.com/patternfly/workflows/workflows-add-step2a/

mabashian:
WFJT prompting hasn't been merged yet but the PR is quite small and I don't think it will impact this work at all so I removed the `blocked` label.

mabashian:
@AlexSCorey moving this into in_progress since I know you've done a lot of work on it already

AlexSCorey:
@mabashian good call, thanks

unlikelyzero:
@AlexSCorey i'll be testing writing up the testcases for this next week

AlexSCorey:
POL bugs reported by QE that should make it into this work if possible.

- [x] Unique IDs for Next, Back and Cancel Buttons

- [ ] Variable field with wrong format like `Foo:Bar` without spacing doesn't display the validation error

- [ ] Trying to launch a JT without enabling the variables POL with survey the variable field doesn't appear on the preview step

PR #8201, opened 2020-09-22
[WIP]5913 Adds WFNode POL

AlexSCorey:
##### SUMMARY
This addresses #5913 but includes a fairly significant re-architect of the JT POL work as well.  The effort here was to allow formik to handle as much of the errors as possible.  

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
WFNode

PR #8399, opened 2020-10-15
Add WF Node Promptability

AlexSCorey:
##### SUMMARY
This addresses https://github.com/ansible/awx/issues/5913

There are known issues with credentials https://github.com/ansible/awx/issues/8200
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI


awx/ui_next/src/components/LaunchPrompt/LaunchPrompt.jsx
```diff
@@ -107,16 +107,22 @@ function LaunchPrompt({ config, resource = {}, onLaunch, onCancel, i18n }) {
...
-        inventory: resource.summary_fields?.inventory || null,
-        credentials: resource.summary_fields?.credentials || null,
-        diff_mode: resource.diff_mode || false,
-        extra_vars: resource.extra_vars || '---',
-        job_type: resource.job_type || '',
-        job_tags: resource.job_tags || '',
-        skip_tags: resource.skip_tags || '',
-        scm_branch: resource.scm_branch || '',
-        limit: resource.limit || '',
+        verbosity: config.ask_verbosity_on_launch && (resource.verbosity || 0),
```
mabashian: I'm curious what you were running in to that made you make this change?  Was it something with the preview step or?

awx/ui_next/src/components/LaunchPrompt/LaunchPrompt.jsx
```diff
@@ -107,16 +107,22 @@ function LaunchPrompt({ config, resource = {}, onLaunch, onCancel, i18n }) {
...
+        inventory:
+          config.ask_inventoryon_launch && resource.summary_fields?.inventory,
```
mabashian: This should be `ask_inventory_on_launch`

awx/ui_next/src/components/LaunchPrompt/LaunchPrompt.jsx
AlexSCorey: Precisely. The Preview Step will show the default resource values in  Prompted Values, even if they are not Prompt on Launch values. So this addition prevents those values from being added on load so that they don't show up in the preview step.

mabashian:
@AlexSCorey running into a console error when trying to launch a job that prompts for only inventory (without a default inventory)

prompt_inv_no_default

Note that the only thing this JT prompts for is Inventory and there is no default.

Screen Shot 2020-10-19 at 4 17 36 PM

awx/ui_next/src/components/LaunchPrompt/LaunchPrompt.jsx
mabashian: @AlexSCorey yea this will cause the default inventory to not be selected on launch.  Was about to create a new comment for it when I ran across it in the UI but fixing this variable should fix it.

Before:
before

After:
after

mabashian:
@AlexSCorey here's a different console error that's thrown when I open up the add node modal:

Screen Shot 2020-10-19 at 4 52 54 PM

I think this is getting thrown because the only thing in our initialValues is nodeResource: https://github.com/ansible/awx/pull/8399/files#diff-7e38ebc07d909e7c52a22d1502ac14eaebbd23fbcb9427224255ee15892e67d6R247-R251

You're generating intialValues in your hooks that get aggregated here: https://github.com/ansible/awx/pull/8399/files#diff-a55638c82435d421cc96f015b4d4d4b33fc2cabffaa13e3b3008209ccb1c8d37R83

but it's never actually used: https://github.com/ansible/awx/pull/8399/files#diff-7e38ebc07d909e7c52a22d1502ac14eaebbd23fbcb9427224255ee15892e67d6R110-R114 when you reference the hook.

My memory is fuzzy now but this sounds familiar.  Did we talk about a need for splitting the logic for generating initialValues from the rest of the logic? cc @keithjgrant

mabashian:
If I create a node with a prompted inventory and then immediately try to edit that same node I'm getting this error:

create_then_edit

Screen Shot 2020-10-19 at 5 02 49 PM

mabashian:
I haven't looked closely at how this particular list is working (I know that a request to the `launch` endpoint is necessary for determining what needs to get prompted) but it looks like the templates list is reloading after every selection and it's a bit jarring:

list_reload

I would expect a single GET request to the `launch` endpoint to be made when the user selects a row.  We could even store the response with the row to prevent the need for future requests to the `launch` endpoint for that particular row if we want to.

mabashian:
After I create a node with a prompted inventory (workflow is saved) and I go back in and edit that node, the preview step seems mostly empty (it's full of info when adding the node):

edit_node_preview

unlikelyzero:
Writing up some tests for this now
Need to retest: https://github.com/ansible/awx/pull/8201#issuecomment-697980349

unlikelyzero:
Found so far:

- Cannot Save a WFJT when the inventory for a JT has been deleted.

unlikelyzero:
@AlexSCorey @mabashian from the bug bash:
https://github.com/ansible/awx/issues/8621

- Wf node modal makes API calls every time I select a different JT and this disables the save button for a while. This is annoying. Seems like its trying to generate whether I get next steps. No caching is happening because when I switch between JTs it lags a ton and does not seem to reuse info

mabashian:
> Wf node modal makes API calls every time I select a different JT and this disables the save button for a while. This is annoying

I mean.... if we don't disable the button you can click it before we've figured out whether or not we need to show next steps and you've created a node with a busted state.  That seems significantly worse.

> No caching is happening because when I switch between JTs it lags a ton and does not seem to reuse info

This is something we could look in to but my response would be... do we expect a user to actually do this?  Click back and forth between two resources?  I think it's much more likely that a user would come in, search the list for what they're looking for and make one selection.  Adding a cache introduces more complex code to an already complex component.

@unlikelyzero ^^

unlikelyzero:
Potentially, Tiago will perform the testing on thursday/friday

tiagodread:
Hey @AlexSCorey I found this so far:

- [ ] Trying to launch a WFJT with a JT with or without prompting fields:

image

- [ ] Page crash if Variables field contain problems related any validation error:

prompt1

- [ ]  If a JT is configured with 'Job Type' POL and a user adds that node to WFViz, the WFViz Add/Edit Node Wizard choses Run as the default option but appear to select a choice.

image
image

- [ ] None additional text is displayed to the inventory step when it's promptable when the outer WFJT inventory is also marked as POL.

mabashian:
@tiagodread

> If a JT is configured with 'Job Type' POL and a user adds that node to WFViz, the WFViz Add/Edit Node Wizard choses Run as the default option but appear to select a choice.

The job type field is a required field with or without prompting so there's always a default value.

> None additional text is displayed to the inventory step when it's promptable when the outer WFJT inventory is also marked as POL.

Good catch.  In the interest of moving this PR forward would it be OK if we split this out into a separate, followup issue?

unlikelyzero:
@mabashian we're going to do a final test tomorrow with the latest fixes. It's likely we'll have to merge it with issues filed. I think we should block the merge on anything `priority:high`

tiagodread:
Hey @AlexSCorey @mabashian we went through all items in [here](https://github.com/ansible/awx/pull/8201#issuecomment-697980349) most of them was fixed but we are still facing some

- [ ] JT with survey multi choice (single select) required but none default option appear like an empty choice:

image
image

- [ ] If the survey step contains any error, is not clear which field this problem is related to

- [ ] If a WFJT or JT is configured for POL with Branch Override and a user adds that WFJT or JT to a node in the WFViz, the WFViz Add/Edit Node Wizard does not respect their prompted input after save. (note: this requires additional Project configuration)

- [ ] Trying to edit the WF node from the "View node details" when you try to save you see:

image

- [ ] Activity and organization information in the Preview step overpoem each other

image

tiagodread:
> @tiagodread
> 
> > If a JT is configured with 'Job Type' POL and a user adds that node to WFViz, the WFViz Add/Edit Node Wizard choses Run as the default option but appear to select a choice.
> 
> The job type field is a required field with or without prompting so there's always a default value.
> 
> > None additional text is displayed to the inventory step when it's promptable when the outer WFJT inventory is also marked as POL.
> 
> Good catch. In the interest of moving this PR forward would it be OK if we split this out into a separate, followup issue?

Opened https://github.com/ansible/awx/issues/8789 as a follow-up this

mabashian:
@tiagodread 

 > JT with survey multi choice (single select) required but none default option appear like an empty choice:

Should be fixed with https://github.com/ansible/awx/pull/8399/commits/dbebde735b527e85c48d832da943ea9885580625

The display was also incorrect on the survey preview modal so I fixed that too.

mabashian:
@tiagodread https://github.com/ansible/awx/pull/8399/commits/ee88b46fbd6ef4abcd8b38117ec5c50adef0958f adds ID's to the wizard steps.

Here are all of the ID's:

`run-type-step`
`node-type-step`
`inventory-step`
`credentials-step`
`other-prompts-step`
`survey-step`
`preview-step`

tiagodread:
Added a [commit](https://github.com/ansible/tower-qa/pull/5743/commits/7d65f2289f6edd8ed23fc82c413a52093d880230) to cover the main flow in 8399

mabashian:
@tiagodread 

> Page crash if Variables field contain problems related any validation error

OK, I think I've got this working reasonably well.  It's not crashing anymore and I think the behavior is actually expected.

bad_yaml

`foo:bar` gets translated to:

```
'0': f
'1': o
'2': o
'3': ':'
'4': b
'5': a
'6': r
```

I believe this is the same behavior exhibited in the old UI and the page no longer crashes.

tiagodread:
> @tiagodread
> 
> > Page crash if Variables field contain problems related any validation error
> 
> OK, I think I've got this working reasonably well. It's not crashing anymore and I think the behavior is actually expected.
> 
> bad_yaml
> 
> `foo:bar` gets translated to:
> 
> ```
> '0': f
> '1': o
> '2': o
> '3': ':'
> '4': b
> '5': a
> '6': r
> ```
> 
> I believe this is the same behavior exhibited in the old UI and the page no longer crashes.

Hey @mabashian while verifying, I realised that this also happen if the JT node without survery, if you try:

1. create a JT
2. enable prompt for variables
3. add this JT to a Workflow
4. fill in `foo:bar` hit save

prompt

unlikelyzero:
@mabashian @tiagodread if we can determine that the bug here https://github.com/ansible/awx/pull/8399#issuecomment-744481345 is not specific to this PR, then i think we should file it as a high priority bug to address outside the pr

mabashian:
@unlikelyzero upon closer inspection this is something that we should fix in this PR and isn't widespread.  I'm working on it right now.

mabashian:
@tiagodread here are the followup issues: https://github.com/ansible/awx/issues/8907 and https://github.com/ansible/awx/issues/8906.

https://github.com/ansible/awx/pull/8399#issuecomment-744481345 should be addressed by my latest commit though.

mabashian (approved):

tiagodread (approved):
Great job!

Verified that workflow node prompt works properly, added a test to cover it on 
`workflow-viz-jt-prompt-and-survey.spec.js`and `workflow-viz-wfjt-prompt-and-survey.spec.js`and

Issue #8401, opened 2020-10-15
Add additional filters to dashboard graph


jlmitch5:
In the transition from nvd3 to using the graphs based on automation analytics line chart, we've lost the status-based and hour-based (last 24 hours) filters.

This tracks adding those back (which will be done after they are added to automation analytics in order to stay in sync.

unlikelyzero:
Marking as enhancement instead of feature in case we want to push past GA

tiagodread:
Missing filter by view successful and failed?
![image](https://user-images.githubusercontent.com/12464858/100460523-1631d100-30a6-11eb-8b3f-18bb159643b7.png)

mabashian:
I'm marking this as `high` but also `blocked`.  @jlmitch5 when you get back lets touch base on this one.

unlikelyzero:
@tiagodread , this should be relatively straightforward to implement

mabashian:
I think we're going to have to roll this ourselves.  Removing the blocked label and putting this back in needs_devel.

Issue #8400, opened 2020-10-15
Add dashboard to ui_next


jlmitch5:
This card tracks adding the dashboard to ui_next.  The big change from ui -> ui_next will be the recently ran jobs and used templates need to be more full featured as they will be replacing portal mode.

unlikelyzero:
Closing. Found one bug post-merge https://github.com/ansible/awx/issues/8439

Issue #6756, opened 2020-04-20
Dashboard revision


wbieniek:
##### ISSUE TYPE
Dashboard page enhancements 

##### SUMMARY
It would be nice to see a button on the Dashboard with currently running and pending  job numbers. How about alerting for longer the expecting running jobs?
Also last few jobs statistics  would be  nice as well.
Even better if there were list of items we could have we could customize our Dashboard to our needs.
I would reduce size of the graph as well.

jlmitch5:
While it does not feature your requests, we have revamped the AWX dashboard in the new UI.

> It would be nice to see a button on the Dashboard with currently running and pending job numbers. How about alerting for longer the expecting running jobs?  Also last few jobs statistics would be nice as well.

While we aren't currently planning on including these features, job and template running/viewing/filtering functionality should be massively improved over the legacy dashboard, due to the fact the lists are fully-featured.

> Even better if there were list of items we could have we could customize our Dashboard to our needs.  I would reduce size of the graph as well.

We have discussed this a lot over the years and would love to have a configurable widget library for users to be able to customize their dashboard view, but the priority at the current moment with the new UI was getting as close to parity with the existing UI as possible.

jlmitch5:
We've now implemented the new dashboard.  Closing this.

PR #8402, opened 2020-10-15
add ui_next dashboard

jlmitch5:
Link https://github.com/ansible/awx/issues/8400 https://github.com/ansible/awx/issues/6756

Screen Shot 2020-10-15 at 3 33 45 PM

Screen Shot 2020-10-15 at 3 46 47 PM

This pr adds the dashboard to ui_next.  The big changes are:
- Dashboard counts.  These are identical (in terms of values and where they link to) compared to the legacy dashboard.
- Dashboard graph.  This is based off the line chart in Automation Analytics.  The UX should be near-100% consistent with AA's chart.  In the transition, a few of the filters were lost (namely status-based and hour based time granularity).  These will be added once they are supported in AA.  Tracking here: https://github.com/ansible/awx/issues/8401
- Dashboard lists.  These are identical to the full-featured lists on the Jobs and Templates sections of the UI, respectively.  The new graphs being full-featured makes portal mode redundant.

### DIFF
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -316,6 +316,9 @@ class DashboardJobsGraphView(APIView):
         if period == 'month':
             end_date = start_date - dateutil.relativedelta.relativedelta(months=1)
             interval = 'days'
+        elif period == 'two_weeks':
+            end_date = start_date - dateutil.relativedelta.relativedelta(weeks=2)
+            interval = 'days'
         elif period == 'week':
             end_date = start_date - dateutil.relativedelta.relativedelta(weeks=1)
             interval = 'days'
--- a/awx/ui_next/src/screens/Dashboard/shared/DashboardTemplateList.jsx
+++ b/awx/ui_next/src/screens/Dashboard/shared/DashboardTemplateList.jsx
@@ -140,7 +140,6 @@ function DashboardTemplateList({ i18n }) {
     jtActions && Object.prototype.hasOwnProperty.call(jtActions, 'POST');
   const canAddWFJT =
     wfjtActions && Object.prototype.hasOwnProperty.call(wfjtActions, 'POST');
-  // spreading Set() returns only unique keys
   const addButtonOptions = [];
 
   if (canAddJT) {
--- a/awx/ui_next/src/screens/Dashboard/shared/DashboardTemplateList.jsx
+++ b/awx/ui_next/src/screens/Dashboard/shared/DashboardTemplateList.jsx
@@ -19,18 +19,24 @@ import useRequest, { useDeleteItems } from '../../../util/useRequest';
 import { getQSConfig, parseQueryString } from '../../../util/qs';
 import useWsTemplates from '../../../util/useWsTemplates';
 import AddDropDownButton from '../../../components/AddDropDownButton';
-import TemplateListItem from './TemplateListItem';
 
-// The type value in const QS_CONFIG below does not have a space between job_template and
-// workflow_job_template so the params sent to the API match what the api expects.
-const QS_CONFIG = getQSConfig('template', {
-  page: 1,
-  page_size: 20,
-  order_by: 'name',
-  type: 'job_template,workflow_job_template',
-});
+import DashboardTemplateListItem from './DashboardTemplateListItem';
 
-function TemplateList({ i18n }) {
+const QS_CONFIG = getQSConfig(
+  'template',
+  {
+    page: 1,
+    page_size: 5,
+    order_by: 'name',
+    type: 'job_template,workflow_job_template',
+  },
+  ['id', 'page', 'page_size']
+);
+
+function DashboardTemplateList({ i18n }) {
+  // The type value in const QS_CONFIG below does not have a space between job_template and
+  // workflow_job_template so the params sent to the API match what the api expects.
+
   const location = useLocation();
 
   const [selected, setSelected] = useState([]);
@@ -244,7 +250,7 @@ function TemplateList({ i18n }) {
             />
           )}
           renderItem={template => (
-            <TemplateListItem
+            <DashboardTemplateListItem
               key={template.id}
               value={template.name}
               template={template}
@@ -271,4 +277,4 @@ function TemplateList({ i18n }) {
   );
 }
 
-export default withI18n()(TemplateList);
+export default withI18n()(DashboardTemplateList);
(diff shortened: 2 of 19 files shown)


awx/api/views/__init__.py
```diff
@@ -316,6 +316,9 @@ def get(self, request, format=None):
         if period == 'month':
             end_date = start_date - dateutil.relativedelta.relativedelta(months=1)
             interval = 'days'
+        elif period == 'two_weeks':
```
mabashian, line 319: @ryanpetrello this look good to you?  Just want to make sure we're not sneaking something by y'all without a review.

mabashian:
Now that we've got recent jobs/templates tabbed out at the bottom my first instinct was... why not have all 3 things in tabs like:

Screen Shot 2020-10-19 at 10 16 56 AM

then there would be little need for scrolling.

Of course, this would mean that recent jobs is now an extra click away.  I don't feel strongly about this, it just came to mind as an option.  It really depends on how this page is actually used/what users care about seeing.

@trahman73 @wenottingham 👍 / 👎 ? (This could easily be changed in the future as well if we'd rather just stick to the current UX initially.)

awx/ui_next/src/screens/Dashboard/shared/DashboardTemplateList.jsx
```diff
@@ -0,0 +1,280 @@
...
+    } else {
+      setSelected(selected.concat(template));
+    }
+  };
+
+  const canAddJT =
+    jtActions && Object.prototype.hasOwnProperty.call(jtActions, 'POST');
+  const canAddWFJT =
+    wfjtActions && Object.prototype.hasOwnProperty.call(wfjtActions, 'POST');
+  // spreading Set() returns only unique keys
```
mabashian: Do we still need this?  Not sure which line it's referencing.

mabashian:
I think this is looking fantastic @jlmitch5!  Curious about `ChartTooltip`/`LineChart` in `Dashboard/shared`.  My assumption is that those files are copied over from the other repo.  Is that right?  Just wondering if we should have a discussion about how to go about sharing components that are not in PF.  Keeping distinct copies in each repo means that we have control over it and won't get blindsided by changes needed for AA that may break something we do in AWX.  At the same time, we'll have to make sure our copy of these components stays up to date if they change in AA.  You may have already talked about this and I'm late to the party but I figure it's worth a conversation.  cc @jakemcdermott

mabashian:
@unlikelyzero @tiagodread I don't anticipate a tremendous amount of change/refactor on this PR so it's probably safe for you all to start getting familiar with it.

unlikelyzero:
@mabashian we are going to do a bug bash on it on wednesday and take a first look at this tomorrow

unlikelyzero:
Note: we will need to put in //to be implemented for https://github.com/ansible/awx/issues/8401

trahman73:
@mabashian That's not a bad idea to combine all 3 views into a tabbed card as shown. I don't think it's a big deal to have to click into the Recent Jobs and Recent Templates tab to get to that content. I'm curious about more opinions as I don't really know how this page is currently used.

jakemcdermott:
@mabashian 
- I do like the idea of adding the third tab. seems like a nicer ux overall. FWIW I'm for it but curious what @jlmitch5 thinks.
- The plan was to just duplicate the chart code over to awx and not worry about making a shared repo or lib at this time. I think we should stick to this plan. IMO there's no need to dive into the complexity of cross-project code sharing until after 4.0
- I heard something about having to move to victory charts at some point. That seems... non-essential as a first step and again would prefer post 4.0 unless it's a blocker for some reason.

unlikelyzero:
@trahman73 @jlmitch5 @jakemcdermott the 3 tab design also improves client-side performance. Our weighted lighthouse score improved by 3/100 points between revisions

unlikelyzero:
@jlmitch5 A bug-free PR!
- will the Chart Hoverover text be translatable?

The only point of feedback is that we'd like to have this a bit more testable. 
- Can you work with us to come up with a deterministic way of hovering over a datapoint? There are no tests for the AA version of this chart so this might involve some minor changes
- Could you add unique IDs to the Cards and chart filter dropdowns

We have an associated PR that we'd like to have merged alongside this PR. It should contain:

- Any associated testability improvements.
- All stubs which were executed manually
- All stubs for future work with //to be implemented
- A visual test for the Dashboard and Graph
- A single test to validate the correctness of the cards.
- A single test to validate the correctness of the chart hoverover
- Re-used tests for CRUD and Search on Job List
- Re-used tests for CRUD and Search on Recent Templates

jakemcdermott:
recheck

mabashian (approved):

mabashian:
^^ I think this is a legitimate failure

```
2020-10-21 12:49:09.462583 | static | FAIL src/screens/Dashboard/Dashboard.test.jsx
2020-10-21 12:49:09.462961 | static |   ● <Dashboard /> › renders jobs list by default
2020-10-21 12:49:09.462997 | static |
2020-10-21 12:49:09.463381 | static |     expect(received).toBe(expected) // Object.is equality
2020-10-21 12:49:09.463416 | static |
2020-10-21 12:49:09.463505 | static |     Expected: 1
2020-10-21 12:49:09.463677 | static |     Received: 0
2020-10-21 12:49:09.463713 | static |
2020-10-21 12:49:09.463762 | static |       31 |
2020-10-21 12:49:09.464156 | static |       32 |   test('renders jobs list by default', () => {
2020-10-21 12:49:09.464786 | static |     > 33 |     expect(pageWrapper.find('JobList').length).toBe(1);
2020-10-21 12:49:09.465935 | static |          |                                                ^
2020-10-21 12:49:09.466021 | static |       34 |   });
2020-10-21 12:49:09.466059 | static |       35 |
2020-10-21 12:49:09.466184 | static |       36 |   test('renders template list when the active tab is changed', async () => {
2020-10-21 12:49:09.466211 | static |
2020-10-21 12:49:09.466396 | static |       at Object.test (src/screens/Dashboard/Dashboard.test.jsx:33:48)
```

Test probably need to be updated after changing to the 3 tabs view

unlikelyzero:
@jlmitch5 we're almost done with the tests. We just need unique IDs for the tabs and the cards. Is that something you can get in tomorrow?

unlikelyzero:
Associated Tests in 5622

PR #8403, opened 2020-10-15
Improve general performance for a variety of high-load job launch use cases

chrismeyersfsu:
reduce per-job database query count

Do not query the database for the set of Instance that belong to the
group for which we are trying to fit a job on, for each job.
Instead, cache the set of instances per-instance group.
reduce parent->child lock contention

We update the parent unified job template to point at new jobs
created. We also update a similar foreign key when the job finishes
running. This causes lock contention when the job template is
allow_simultaneous and there are a lot of jobs from that job template
running in parallel. I've seen as bad as 5 minutes waiting for the lock
when a job finishes.
This change moves the parent->child update to OUTSIDE of the
transaction if the job is allow_simultaneous (inherited from the parent
unified job). We sacrafice a bit of correctness for performance. The
logic is, if you are launching 1,000 parallel jobs do you really care
that the job template contains a pointer to the last one you launched?
Probably not. If you do, you can always query jobs related to the job
template sorted by created time.

### DIFF
--- a/awx/main/models/unified_jobs.py
+++ b/awx/main/models/unified_jobs.py
@@ -873,7 +873,13 @@ class UnifiedJob(PolymorphicModel, PasswordFieldsModel, CommonModelNameNotUnique
 
         # If status changed, update the parent instance.
         if self.status != status_before:
-            self._update_parent_instance()
+            # Update parent outside of the transaction for Job w/ allow_simultaneous=True
+            # This dodges lock contention at the expense of the foreign key not being
+            # completely correct.
+            if getattr(self, 'allow_simultaneous', False):
+                connection.on_commit(self._update_parent_instance)
+            else:
+                self._update_parent_instance()
 
         # Done.
         return result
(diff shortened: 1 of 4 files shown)


awx/main/models/unified_jobs.py
```diff
@@ -871,9 +871,15 @@ def save(self, *args, **kwargs):
...
         result = super(UnifiedJob, self).save(*args, **kwargs)
 
-        # If status changed, update the parent instance.
-        if self.status != status_before:
-            self._update_parent_instance()
+        # Update parent outside of the transaction for Job w/ allow_simultaneous=True
+        # This dodges lock contention at the expense of the foreign key not being
+        # completely correct.
+        if getattr(self, 'allow_simultaneous', False):
+            connection.on_commit(self._update_parent_instance)
```
ryanpetrello: Is this necessary if `self.status == status_before:` ?

Would this be better?

```python
if self.status != status_before:
    if getattr(self, 'allow_simultaneous', False):
        connection.on_commit(self._update_parent_instance)
    else:
        self._update_parent_instance()
```

ryanpetrello (approved):

fosterseth (approved):

ryanpetrello:
regate

Issue #8407, opened 2020-10-16
Installer missing docker_image command in Ansible 2.10.2


marcolussetti:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

Installer does not run in my environment complaining that docker_image cannot be found.

##### ENVIRONMENT
* AWX version: 15.0
* AWX install method: docker on linux
* Ansible version:  2.10.2
* Operating System: Centos 8.2
* Web Browser: N/A

##### STEPS TO REPRODUCE
1. Blank CentOS 8.2 install
2. `pip install ansible`
3. Edit inventory as appropriate
4. Run installer

##### EXPECTED RESULTS
Playbooks run to completion, and AWX is installed.

##### ACTUAL RESULTS
Playbook errors out noting that docker_image is not recognized.

marcolussetti:
Installing the appropriate collection with `ansible-galaxy collection install community.general` fixes the issue.

ryanpetrello:
Install instructions added at: https://github.com/ansible/awx/pull/8408 (thanks @marcolussetti)

PR #8408, opened 2020-10-16
Point installer to community.general.docker_image

marcolussetti:
Fixes issue in Ansible 2.10.2 where docker_image command is not found.

##### SUMMARY

Fixes issue #8407 in my environment. The change to the `main.yml` tasks file is not strictly necessary, but I figured it might make sense to do as well?

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - Installer

##### AWX VERSION

```
awx: 15.0.0
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/installer/roles/image_build/tasks/main.yml
+++ b/installer/roles/image_build/tasks/main.yml
@@ -42,7 +42,7 @@
   delegate_to: localhost
 
 - name: Build sdist builder image
-  community.general.docker_image:
+  docker_image:
     build:
       path: "{{ role_path }}/files"
       dockerfile: Dockerfile.sdist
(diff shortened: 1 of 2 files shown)


installer/roles/image_build/tasks/main.yml
```diff
@@ -42,7 +42,7 @@
   delegate_to: localhost
 
 - name: Build sdist builder image
-  docker_image:
+  community.general.docker_image:
```
wenottingham: This wouldn't be backwards compatible with running on a bare 2.9.

marcolussetti: Understood. Will switch back & add note to the INSTALL.md that the requirement is only for ansible >= 2.10

marcolussetti: Switched back & marked conversation as resolved.

ryanpetrello (approved):

Issue #8331, opened 2020-10-06
Integer and float surveys are broken


dsesami:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
You can't create a survey in a job template of `Integer` type or `Float` type.
Replicating the survey question like in the screenshot + saving results in the console errors shown.

Also, you can't set textarea/text default answers to just a number after you fail the initial Save -- it is interpreted as a number and returns a 400. This will probably go away after the first issue is fixed.

![Screenshot from 2020-10-06 14-05-50](https://user-images.githubusercontent.com/19942822/95242465-339c9a00-07dd-11eb-90cb-8b8d98553b90.png)

unlikelyzero:
May be a dupe of https://github.com/ansible/awx/issues/8222

tiagodread:
PR already merged and tests in place

Issue #8222, opened 2020-09-23
The JT Survey Edit Question page does not allow a default answer if Answer Type is Integer or Float


unlikelyzero:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
The JT Survey Edit Page does not allow a default answer if Answer Type is Integer or Float

##### ENVIRONMENT
* AWX version: aac17b9d2c7ae17c2dc204de11490401bc75f7a2

##### STEPS TO REPRODUCE
1. Navigate to the Survey Edit Question page.
2. Attempt to create or edit a question with Integer or Float type
3. Enter an integer value into Default answer field.
4. Click Save
Note: Inert Action.

##### EXPECTED RESULTS
1. Save Survey Question

##### ACTUAL RESULTS
1. Inert action

##### ADDITIONAL INFORMATION
```
Warning: An unhandled error was caught during low priority validation in <Formik validate /> TypeError: value.trim is not a function
    at Object.validate (validators.jsx:21)
    at Formik.tsx:260
    at new Promise (<anonymous>)
    at Formik.tsx:260
    at Formik.tsx:276
    at Array.map (<anonymous>)
    at Formik.tsx:274
    at Formik.tsx:299
    at Formik.tsx:330
    at unstable_runWithPriority (scheduler.development.js:697)
    at Formik.tsx:329
    at Formik.tsx:1214
    at Formik.tsx:601
    at Formik.tsx:1214
    at Formik.tsx:657
    at Formik.tsx:670
    at Object.onChange (Formik.tsx:1214)
    at Object.onChange (FormField.jsx:70)
    at TextInputBase.handleChange (TextInput.tsx:80)
    at HTMLUnknownElement.callCallback (react-dom.development.js:188)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:237)
    at invokeGuardedCallback (react-dom.development.js:292)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:306)
    at executeDispatch (react-dom.development.js:389)
    at executeDispatchesInOrder (react-dom.development.js:414)
    at executeDispatchesAndRelease (react-dom.development.js:3278)
    at executeDispatchesAndReleaseTopLevel (react-dom.development.js:3287)
    at Array.forEach (<anonymous>)
    at forEachAccumulated (react-dom.development.js:3257)
    at runEventsInBatch (react-dom.development.js:3304)
    at runExtractedPluginEventsInBatch (react-dom.development.js:3514)
[...]

PR #8419, opened 2020-10-20
Fix integer/float errors in survey

keithjgrant:
##### SUMMARY
Fix console errors for integer & float default values when setting up a template survey

addresses #8331 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/util/validators.jsx
+++ b/awx/ui_next/src/util/validators.jsx
@@ -68,7 +68,6 @@ export function noWhiteSpace(i18n) {
 
 export function integer(i18n) {
   return value => {
-    console.log(value);
     const str = String(value);
     if (/[^0-9]/.test(str)) {
       return i18n._(t`This field must be an integer`);
--- a/awx/ui_next/src/util/validators.jsx
+++ b/awx/ui_next/src/util/validators.jsx
@@ -68,11 +68,22 @@ export function noWhiteSpace(i18n) {
 
 export function integer(i18n) {
   return value => {
+    console.log(value);
     const str = String(value);
     if (/[^0-9]/.test(str)) {
       return i18n._(t`This field must be an integer`);
     }
     return undefined;
+  };
+}
+
+export function number(i18n) {
+  return value => {
+    const str = String(value);
+    if (/^[0-9]*(\.[0-9]*)?$/.test(str)) {
+      return undefined;
+    }
+    return i18n._(t`This field must be a number`);
   };
 }
 
--- a/awx/ui_next/src/util/validators.jsx
+++ b/awx/ui_next/src/util/validators.jsx
@@ -79,7 +79,11 @@ export function integer(i18n) {
 export function number(i18n) {
   return value => {
     const str = String(value);
-    if (/^[0-9]*(\.[0-9]*)?$/.test(str)) {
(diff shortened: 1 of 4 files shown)


awx/ui_next/src/util/validators.jsx
```diff
@@ -76,6 +77,16 @@ export function integer(i18n) {
   };
 }
 
+export function number(i18n) {
+  return value => {
+    const str = String(value);
+    if (/^[0-9]*(\.[0-9]*)?$/.test(str)) {
```
jakemcdermott: Probably should also handle:
  - negative numbers `/^[0-9]*(\.[0-9]*)?$/.test(String(-2.2))`
  - big numbers `/^[0-9]*(\.[0-9]*)?$/.test(String(999999999999999999999.9))`

awx/ui_next/src/util/validators.jsx
```diff
@@ -68,6 +68,7 @@ export function noWhiteSpace(i18n) {
 
 export function integer(i18n) {
   return value => {
+    console.log(value);
```
jakemcdermott:
```suggestion
```

jakemcdermott (commented):
lgtm, just a few nits & I think we should merge this.

jakemcdermott (changes requested):
lgtm, just a few nits & I think we should test/merge this.

jakemcdermott (approved):

tiagodread:
@keithjgrant I tried to create a float survey and got this result:

survey

Also, try add one each survey type into JT, when you add Multiple choice option and try launch the JT the page crashes totally:

survey1

keithjgrant:
> @keithjgrant I tried to create a float survey and got this result:

I get this error in the old UI as well... I think this should be filed as an API bug. It's not accepting `2` as a value

> Also, try add one each survey type into JT, when you add Multiple choice option and try launch the JT the page crashes totally:

Should be fixed now

unlikelyzero:
please check this issue at the same time https://github.com/ansible/awx/issues/8222

unlikelyzero:
@keithjgrant @tiagodread please check #8222 and also file the float bug against the API

tiagodread:
@keithjgrant I saw that the ui_next and old UI is passing the `default` as integer:

```
{
    "description": "",
    "name": "",
    "spec": [
        {
            "choices": "",
            "default": 2,
            "max": 1024,
            "min": 0,
            "new_question": true,
            "question_description": "",
            "question_name": "1.0+1.0",
            "required": true,
            "type": "float",
            "variable": "float"
        }
    ]
}
```

changing 2 to 2.0 the API can save successfully

keithjgrant:
JavaScript doesn't distinguish between 2 and 2.0. I might be able to send the string "2.0", but a string is definitely not a float

keithjgrant:
@tiagodread JavaScript doesn't distinguish between 2 and 2.0. The only other option I have is to send the string "2.0", but a string is definitely not a float, and the API returns the same error.

tiagodread:
Open a new issue to fix only Float default value here https://github.com/ansible/awx/issues/8460

tiagodread (approved):
Good Job!
Tests covered by 5635

Issue #8429, opened 2020-10-21
Expose the SCM Branch name in the playbook environment


Malizor:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Today if I launch a job and put `foobar` in the "SCM Branch" field, the corresponding commit hash is exposed as `awx_project_revision` but the proper human readable name of my branch/tag `foobar` remains inaccessible.

A use case to access this information is to put it in configuration files (eg. in a UI to document which version of the platform was deployed). Many more can be envisioned too.

I see two paths for implementation:
* Change the `awx_project_revision` content to keep the SCM Branch as it was in the input (stop turning it into a hash)
* Expose this into a variable of its own, like `awx_project_scm_branch`

jakemcdermott:
Thought this was ui related but on second look it isn't.

kdelee:
The variable is now available. Put up a PR in the tower docs repo, so not going to close yet, but once the docs are in we can close.

Malizor:
[This page](https://docs.ansible.com/ansible-tower/latest/html/userguide/job_templates.html) seems to be the best candidate for adding the new variables documentation, but I can't find the related repository I should send a PR to.
Could someone point it to me?

pmanno:
Hi,

Any update on this? I can't seem to find any info in the docs on it.

Also, I want to parameterize the requirements.yml file of the project, so I can update the project branch and all referenced roles. Should be able to just set

version: {{ scm_revision} }  in requirements.yml for that, right? 

Thanks,
Paul

ckabalan:
+1 for this feature

Our use case is that we have playbooks that call other playbooks by name and many times we need to update multiple playbooks to create a new release. For example, we have a playbook that does the infrastructure provisioning phase of a VM deployment, then another that runs to configure the OS, then another that runs to test the resulting VM. Being able to call these by name using `tower_job_launch` with the SCM branch variable would allow us to say "VM Deploy feature-name" would also launch the templates "OS Configure feature-name" and "VM Test feature-name".

joshebosh:
+1 for this feature... I have roles that do other git repos and would be nice to checkout branches based on the value in Project > SCM BRANCH box....

wenottingham:
This has been fixed for a while. Closing.

ryedin:
Not doubting that it's fixed, but it may not be documented well still. Wanted this feature, looked through the docs, googled, found this issue, back to looking through the docs.

[...]

PR #8433, opened 2020-10-21
Expose the SCM Branch name in the playbook environment

Malizor:
##### SUMMARY
Expose more job setting variables in the playbook environment.

related #8429

A playbook wanting to know the branch or tag name it is deployed from has currently no way to do so.

With this change, any configured default project SCM branch is exposed as `{awx,tower}_project_scm_branch`.
If the SCM Branch is overwritten in the job, the related value is exposed as `{awx,tower}_job_scm_branch`.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
awx: 15.0.1
```


ryanpetrello (approved):

DE110283:
In which Tower version this will be available? Just tested in version 3.8.3 and it is still missing... (tower_job_scm_branch and tower_project_scm_branch not available in job run).

jasonk876:
Is this available? Believe this would fix an issue that we are having.

Issue #7595, opened 2020-07-09
Rename users and access tabs


marshmalien:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Task: Update subtab names

<img width="366" alt="Screen Shot 2020-07-09 at 3 22 26 PM" src="https://user-images.githubusercontent.com/15881645/87082005-4d3b3f00-c1f8-11ea-856a-a138f5da3081.png">

<img width="506" alt="Screen Shot 2020-07-09 at 3 25 40 PM" src="https://user-images.githubusercontent.com/15881645/87082160-8ecbea00-c1f8-11ea-9ab9-80d14a9c4065.png">

[These tables](https://docs.google.com/spreadsheets/d/1DSgZC1JJYI6oCp47QRjWhDAc3CznB6tDZI3GuhxLdqc/edit?usp=sharing) show the sub-tabs titles and content for each top-level view (User, Team, Organization) in the old and new UI. Under **User** and **Team**, I’ve crossed out the current list tab title and replaced it with the new proposed name. These new names should make the tab titles and their list content more consistent across these objects.

Currently, the “Access” tab in Users and Teams shows assigned resource roles, but in Organization, the same tab shows a list of users that have access to the organization via some role. Basically, the idea is to rename the tabs to match the API endpoint we are receiving information from.

##### ADDITIONAL INFORMATION
https://docs.google.com/spreadsheets/d/1DSgZC1JJYI6oCp47QRjWhDAc3CznB6tDZI3GuhxLdqc/edit?usp=sharing

cc @tvo318 @trahman73

unlikelyzero:
Please run through the e2e suite. This will likely break some tests that depend on the tabs

mabashian:
@unlikelyzero looks like this got merged without test.  Applying needs_test now.

unlikelyzero:
Adding @akus062381 to this to cover as a part of the access list testing on JT forms

akus062381:
Read what's listed and figure out if the current implementation is correct.

akus062381:
There is a Google doc where I have compiled everything, and specifies some issues discovered here. 
This is in progress- will meet with Dev to determine which are bugs and which aren't, and file bug reports accordingly.

akus062381:
this has been handled by breaking smaller issues out into their own issue files.

PR #8441, opened 2020-10-22
Rename users and access tabs

nixocio:
Rename users and access tabs

See: https://github.com/ansible/awx/issues/7595


nixocio:
@unlikelyzero, this change will break a few tests, I already ran the test suite, and I will fix the tests myself.

marshmalien (approved):
Perfect ✨

Issue #7665, opened 2020-07-20
API should not accept modification of local_path for nom-manual projects


AlanCoding:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
The API will accept a PATCH to edit `local_path` for a git project, but the change will not take effect.

Some version of a 400 error would be preferable to this. This is a valid field to edit in the case of manual projects, so there's no valid justification that it's "not a field".

##### ENVIRONMENT
* AWX version: 13.0.0
* AWX install method: N/A
* Ansible version:  N/A

##### STEPS TO REPRODUCE

PATCH `/api/v2/projects/6/`

```json
{
  "description": "asdf",
  "local_path": "_6__demo_project_edited"
}
```

##### EXPECTED RESULTS

One of:

 - the local_path value changes to "_6__demo_project_edited"
 - get error

##### ACTUAL RESULTS

No error

description changes to "asdf"

the `local_path` remains at prior value of "_6__demo_project"

kdelee:
Tested this with doing PATCH to a project with:

```
{
    "local_path": "_188__project_FOOEDIT"
}
```

got expected:
```
{
    "local_path": [
        "Cannot change local_path for git-based projects"
    ]
}
```

works for me! thanks

PR #8447, opened 2020-10-23
related #7665 added 400 if user attempts an edit to local_path in a SCM project

thedoubl3j:
##### SUMMARY

fix for #7665 to not accept changes to the local_path if its a non-manual project. Change is adding a logical comparison of the patch value of `local_path` to the instance (stored/current) value of `local_path` and if they differ, fail with a simple error stating that local_path cannot be edited for an SCM project. 

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
$ make VERSION
awx: 15.0.1

```

##### ADDITIONAL INFORMATION

Big thanks to @rebeccahhh for the "print testing" tips that can be done with the serializer

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -1333,6 +1333,8 @@ class ProjectOptionsSerializer(BaseSerializer):
             scm_type = attrs.get('scm_type', u'') or u''
         if self.instance and not scm_type:
             valid_local_paths.append(self.instance.local_path)
+        if scm_type and self.instance.local_path != attrs.get('local_path'):
+            errors['local_path'] = _(f'Cannot change local_path for {scm_type}-based projects')
         if scm_type:
             attrs.pop('local_path', None)
         if 'local_path' in attrs and attrs['local_path'] not in valid_local_paths:
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -1333,7 +1333,7 @@ class ProjectOptionsSerializer(BaseSerializer):
             scm_type = attrs.get('scm_type', u'') or u''
         if self.instance and not scm_type:
             valid_local_paths.append(self.instance.local_path)
-        if scm_type and self.instance.local_path != attrs.get('local_path'):
+        if scm_type and "local_path" in attrs and self.instance.local_path != attrs['local_path']:
             errors['local_path'] = _(f'Cannot change local_path for {scm_type}-based projects')
         if scm_type:
             attrs.pop('local_path', None)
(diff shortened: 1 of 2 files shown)


awx/api/serializers.py
```diff
@@ -1333,6 +1333,8 @@ def validate(self, attrs):
             scm_type = attrs.get('scm_type', u'') or u''
         if self.instance and not scm_type:
             valid_local_paths.append(self.instance.local_path)
+        if scm_type and self.instance.local_path != attrs['local_path']:
+            errors['local_path'] = _('Cannot change local_path for projects with SCM')
```
ryanpetrello: Could we make this error message a bit more explicit, i.e..,

`_(f'Cannot change local_path for {scm_type}-based projects')`

ryanpetrello (commented):
Thanks for digging into this @thedoubl3j 😄 

Would you mind writing a test that verified the behavior here?  This might be a good spot for it:

https://github.com/ansible/awx/blob/devel/awx/main/tests/functional/api/test_project.py

Mostly something that:

1.  Creates a git-based project.
2.  Tries to submit an `HTTP PATCH` like the one described in https://github.com/ansible/awx/issues/7665
3.  Makes an assertion about the response status code (400), and that the error message in the HTTP response matches the error message in this PR.

ryanpetrello:
Specifically, here's a test you could probably start with:

https://github.com/ansible/awx/blob/devel/awx/main/tests/functional/api/test_project.py#L44

It uses the `project` fixture to make a project, then does an HTTP PATCH against it.  You could modify this slightly to send the `local_path` argument, and make assertions with `expect=400`, and `resp.data`.

thedoubl3j:
for sure @ryanpetrello working on some other stuff but might be able to get something up by EoD

ryanpetrello (approved):
This looks great - thanks!

awx/api/serializers.py
```diff
@@ -1333,6 +1333,8 @@ def validate(self, attrs):
             scm_type = attrs.get('scm_type', u'') or u''
         if self.instance and not scm_type:
             valid_local_paths.append(self.instance.local_path)
+        if scm_type and self.instance.local_path != attrs.get('local_path'):
```
ryanpetrello: Regarding failing tests, you might want something like this:

```python
if scm_type and "local_path" in attrs and and self.instance.local_path != attrs['local_path']:
```

thedoubl3j: :+1: tested locally.

ryanpetrello (approved):
🎉

PR #8462, opened 2020-10-26
cast create_preload_data to boolean with `create_preload_data | bool` in launch_awx_task.sh.j2

jakegatsby:
##### SUMMARY

The `create_preload_data` var from the installer inventory file is a string, and currently always evaluates to `True` in the `launch_awx_task.sh.j2` Jinja template.  This results in the preload data always being populated, regardless of setting in the inventory file.  This was tested using the docker-compose deployment method and discussed in #ansible-awx.  Also, piping to the `bool` filter is used for all the other boolean type vars in the inventory when used in jinja templates as part of the installation process.

##### ISSUE TYPE

Bugfix Pull Request

##### COMPONENT NAME

Installer

##### AWX VERSION

```
awx: 15.0.1
```

##### ADDITIONAL INFORMATION

```

```


ryanpetrello (approved):

ryanpetrello:
Thanks @jakegatsby!

Issue #7130, opened 2020-05-22
Users Team view: add to team


keithjgrant:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
On the User Teams page, implement an "Add" button to add the user to teams. Mockup: https://tower-mockups.testing.ansible.com/patternfly/users/users-teams-add/

This is broken out from #6594

nixocio:
Related - https://github.com/ansible/awx/issues/6593

tiagodread:
PR already merged and tests in place

unlikelyzero:
Hey, I don't see coverage for the dissociate case. 
All I see is 
user-tabs.spec.js
context.skip('Teams tab', function() {}) //TBI```

tiagodread:
@unlikelyzero this PR covers association and disassociation https://github.com/ansible/tower-qa/pull/5675

tiagodread:
PR already merged, tests in place, closing this issue.

Issue #8237, opened 2020-09-25
Fix Kebabified Toolbar Buttons


AlexSCorey:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
On some lists, where we have one off buttons (add, edit) the kebabified buttons do not render properly when advanced search is activated.  

Those lists are:

- [x] Inventory Host Groups List

- [x] Inventory Groups Host List

- [x] Inventory Groups Nested Groups List

##### ENVIRONMENT
* AWX version: 14.1
* AWX install method: docker for mac
* Ansible version:  2.9.1
* Operating System: Mojave
* Web Browser:Chrome

##### STEPS TO REPRODUCE
Go to any of the lists above.  In the search drop down select Advanced then click on the kebabed icon on the right hand side of the toolbar to see the buttons appear.  

##### EXPECTED RESULTS

![Screen Shot 2020-09-25 at 11 29 31 AM](https://user-images.githubusercontent.com/39280967/94285965-6a8fc780-ff22-11ea-9ef9-f71bf31fd171.png)

##### ACTUAL RESULTS

![Screen Shot 2020-09-25 at 11 30 23 AM](https://user-images.githubusercontent.com/39280967/94286092-898e5980-ff22-11ea-9e25-81016787917d.png)

tiagodread:
PR already merged, worked as expected

tiagodread:
More places with button inside kebab:

**Job Templates:**

- [x]  Access tab -> advanced search -> kebab -> Add

![image](https://user-images.githubusercontent.com/12464858/97215814-b6dd5a00-17a3-11eb-811f-cab797505ce6.png)

- [x] Access tab -> Add -> Users or team -> advanced search -> kebab -> empty (same happen in WFJT Node) 

![image](https://user-images.githubusercontent.com/12464858/97216158-2e12ee00-17a4-11eb-8ac0-0eaf7aeaaaa5.png)

**Credentials:**

- [x] Access tab -> advanced search -> kebab -> Add

![image](https://user-images.githubusercontent.com/12464858/97216432-93ff7580-17a4-11eb-9284-078691212ce5.png)

**Projects:**

- [x] Access tab -> advanced search -> kebab -> Add

![image](https://user-images.githubusercontent.com/12464858/97216573-d1640300-17a4-11eb-9fbf-2cbd4381e515.png)
- [x] Notifications -> advanced search -> kebab -> empty

[...]

Issue #4212, opened 2019-01-23
Establish and implement loading state


mabashian:
In yolotower we use the `Loading...` spinner down in the bottom right corner to indicate that we're interacting with the api.  We also slap a transparent div over the entire application to prevent the user from clicking on anything while we're making our requests.

We'll want to establish a ux flow around "loading" when: 

a) Initially loading a list
b) Navigating from one page to another
c) Sorting the list
d) (In the future) performing a search

We have the current loading screen mockup:

<img width="1427" alt="screen shot 2019-01-23 at 10 03 17 am" src="https://user-images.githubusercontent.com/9889020/51615558-269b2280-1ef6-11e9-9cf6-8db69096044e.png">

My initial impression was that this would be shown _instead_ of the list during the aforementioned actions but as @marshmalien pointed out to me, this would cause a flash when going from page to page which we currently don't have.

I've opened up this issue to track discussion and eventually outline exactly what needs to be implemented.

marshmalien:
After discussing this issue with @mabashian and @trahman73, we've proposed the following:

#### On initial list load:
  * List is in a loading state (fetching data), resource count is unknown:
    * Show empty card panel with spinner icon centered within the list panel
    * Hide toolbar, delete, checkbox
    * Show Add button
    * Mockup:
![pf-orgs-list-empty-loading](https://user-images.githubusercontent.com/15881645/51635846-c4f2ac80-1f25-11e9-8b13-2fd1ef56fc1a.png)

  * List is not in a loading state. There are no resources:
    * Show PF-4 EmptyState component
    * Hide toolbar, delete, checkbox
    * Show Add button
    * Mockup:
![pf-orgs-list-empty-list](https://user-images.githubusercontent.com/15881645/51635801-a55b8400-1f25-11e9-836c-d12da3a594f8.png)

  * List is not in a loading state. There are resources:
    * Show resource list

#### Paginating through list
  * List is in a loading state (fetching data):
    * Keep the current list
    * Overlay a spinner centered within the list panel (Sync-up with @trahman73 to determine the height)
    * Render new list once api request completes
    * Add Spinner icon: fa-spinner
    * Add overlay with opacity. See mockup. 
    * Mockup: 
![image](https://user-images.githubusercontent.com/15881645/51637342-86f78780-1f29-11e9-89fb-d47da53c8b2e.png)

#### Sorting the list
  * List is in a loading state:
    * Keep the current list
    * Overlay a spinner centered within the list panel
[...]

PR #8463, opened 2020-10-26
Add feature to associate teams to users

nixocio:
Add feature to associate teams to users. For the time being when
associating Users to a team, the User will be associated with `member_role` only. And when `disassociating` the User from a team all related roles - member, read, and admin will be removed.

Also, fix a bug related to search not being cleared after closing/cancel
the `AssociateModal`.

See: https://github.com/ansible/awx/issues/7130

image

image

image

### DIFF
--- a/awx/ui_next/src/screens/User/UserTeams/UserTeamList.test.jsx
+++ b/awx/ui_next/src/screens/User/UserTeams/UserTeamList.test.jsx
@@ -1,83 +1,244 @@
 import React from 'react';
 import { act } from 'react-dom/test-utils';
-import { UsersAPI } from '../../../api';
-import { mountWithContexts } from '../../../../testUtils/enzymeHelpers';
+import { createMemoryHistory } from 'history';
+
+import { UsersAPI, TeamsAPI } from '../../../api';
+import {
+  mountWithContexts,
+  waitForElement,
+} from '../../../../testUtils/enzymeHelpers';
 
 import UserTeamList from './UserTeamList';
 
 jest.mock('../../../api');
+jest.mock('react-router-dom', () => ({
+  ...jest.requireActual('react-router-dom'),
+  useParams: () => ({
+    id: 1,
+    userId: 2,
+  }),
+}));
 
-const mockAPIUserTeamList = {
-  data: {
-    count: 3,
-    results: [
-      {
-        name: 'Team 0',
-        id: 1,
-        url: '/teams/1',
-        summary_fields: {
-          user_capabilities: {
-            delete: true,
-            edit: true,
-          },
+const mockAPIUserTeamList = [
+  {
+    name: 'Team 0',
+    id: 1,
+    url: '/teams/1',
+    summary_fields: {
+      user_capabilities: {
+        delete: true,
+        edit: true,
+      },
+      object_roles: {
+        member_role: {
+          id: 42,
+        },
+        admin_role: {
+          id: 43,
+        },
+        read_role: {
+          id: 44,
         },
       },
-      {
-        name: 'Team 1',
-        id: 2,
-        url: '/teams/2',
-        summary_fields: {
-          user_capabilities: {
-            delete: true,
-            edit: true,
-          },
-        },
-      },
-      {
-        name: 'Team 2',
-        id: 3,
-        url: '/teams/3',
-        summary_fields: {
-          user_capabilities: {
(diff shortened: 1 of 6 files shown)


AlexSCorey:
During advanced search the toolbar buttons should be dropdown items, not buttons 
Screen Shot 2020-10-28 at 10 08 49 AM

AlexSCorey:
~~I don't think we need the row item label for `Organization` here~~ Realizing that most people won't name their Organizations "Organization 0" so we probably do need the label until we the the Table in.
Screen Shot 2020-10-28 at 10 10 01 AM

AlexSCorey:
this is weird. Looks like it is also on Organization so probably an issue with search.  (https://github.com/ansible/awx/issues/8474).  Just FYI
weird filter

AlexSCorey:
Signed in as a Normal User the Add button is activated when it shouldn't be.
normal user

awx/ui_next/src/screens/User/UserTeams/UserTeamList.test.jsx
```diff
@@ -1,83 +1,127 @@
...
+      },
+    });
+
+    UsersAPI.readRoleOptions.mockResolvedValue(options);
+    const history = createMemoryHistory({
+      initialEntries: ['/users/1/teams'],
+    });
     await act(async () => {
-      wrapper = mountWithContexts(<UserTeamList />);
+      wrapper = mountWithContexts(
```
AlexSCorey: Why are you using `<Route>` here?

AlexSCorey (changes requested):
in UserTeamList you will want to add a test to be sure that you are properly associating a team, and then updating the list. Same thing for disassociate.

awx/ui_next/src/screens/User/UserTeams/UserTeamList.test.jsx
nixocio: Removed.

nixocio:
Thanks for your review @AlexSCorey. Updated per comments.

> During advanced search the toolbar buttons should be dropdown items, not buttons

https://github.com/ansible/awx/issues/8237

There is an issue tracking this problem - it was not introduced by this PR.

tiagodread:
@nixocio I Tried to click on Add button but it doesn’t appear

nixocio:
Thanks @tiagodread, ~updating~ updated.

nixocio:
hey @AlexSCorey, do you have any more comments?

tiagodread:
Hey @nixocio after making a search for teams -> closing the modal -> Clicking on Associate the modal opens with the old search state, we are not resetting the state of the modal

modal

nixocio:
> Hey @nixocio after making a search for teams -> closing the modal -> Clicking on Associate the modal opens with the old search state, we are not resetting the state of the modal
> 
> modal

@tiagodread, this issue was not introduced by this PR, but I added a fix here as well.

The same behavior was present in other screens using the `AssociateModal`. Thanks!

AlexSCorey (approved):
This looks good to me.  You've addressed all of my concerns. Good work!

tiagodread:
When a user is associated to a team and then navigate to this user -> teams tab and try to disassociate that team nothing happens

teams

nixocio:
@tiagodread, the last reported issue was addressed. Please, take another look.

tiagodread:
@nixocio now we can successfully associate and disassociate teams to users, great job!!!

My only concern about it is that if you try to associate/disassociate a ton of teams it'll take long time to complete all xhr requests specially for slow connections. 
teams

A suggestion could be disable the buttons Associate/Disassociate until these xhr requests finish or something different to indicates that something is happening in background, what do you think about it ?

nixocio:
> @nixocio now we can successfully associate and disassociate teams to users, great job!!!
> 
> My only concern about it is that if you try to associate/disassociate a ton of teams it'll take long time to complete all xhr requests specially for slow connections.
> teams
> 
> A suggestion could be disable the buttons Associate/Disassociate until these xhr requests finish or something different to indicates that something is happening in background, what do you think about it ?

Hey @tiagodread, that is a valid point. There is an issue tracking related work here https://github.com/ansible/awx/issues/4212, I added a comment for delete/disassociating actions as well.

tiagodread (approved):
Great Job Kersom 🎉

Issue #8453, opened 2020-10-25
Username is a required field


nixocio:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Username is a required field by the API.

![Screenshot_2020-10-25 Ansible AWX(1)](https://user-images.githubusercontent.com/9053044/97114170-47913880-16c5-11eb-8e10-621b6b72e1b2.png)

tiagodread:
Tests covered by 5658 and PR already merged

PR #8468, opened 2020-10-27
Fix username as a required field

nixocio:
Fix username as a required field. `UserForm` is used for adding and
editing an user. When adding an user, the initial user value is a `{}`
update logic to cover this case.

Also, add unit-tests to cover this particular case.

See: https://github.com/ansible/awx/issues/8453

Add

image

Edit

image

### DIFF
--- a/awx/ui_next/src/screens/User/shared/UserForm.jsx
+++ b/awx/ui_next/src/screens/User/shared/UserForm.jsx
@@ -68,11 +68,9 @@ function UserFormFields({ user, i18n }) {
         name="username"
         type="text"
         validate={
-          !ldapUser && externalAccount === null
-            ? required(null, i18n)
-            : () => undefined
+          !ldapUser && !externalAccount ? required(null, i18n) : () => undefined
         }
-        isRequired={!ldapUser && externalAccount === null}
+        isRequired={!ldapUser && !externalAccount}
       />
       <FormField
         id="user-email"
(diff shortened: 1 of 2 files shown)


awx/ui_next/src/screens/User/shared/UserForm.jsx
```diff
@@ -68,11 +68,9 @@ function UserFormFields({ user, i18n }) {
         name="username"
         type="text"
         validate={
-          !ldapUser && externalAccount === null
-            ? required(null, i18n)
-            : () => undefined
+          !ldapUser && !externalAccount ? required(null, i18n) : () => undefined
         }
-        isRequired={!ldapUser && externalAccount === null}
+        isRequired={!ldapUser && !externalAccount}
```
jakemcdermott, line 73: JavaScript surprise: If `externalAccount` is `{}`, then `!externalAccount` will be _false_. 

Screenshot from 2020-10-27 17-01-41

Would that be the expected behavior here? The reason I ask is because in the PR summary there is a mention of `{}` being used as the default user value.

nixocio: It was the initial value of user `{}`, then `externalAccount` was `undefined`. I will add more unit-tests around this. Thanks for your review.

nixocio: @jakemcdermott, I added more unit-tests.

jakemcdermott (approved):

tiagodread (approved):
Tests covered by 5658

PR #8471, opened 2020-10-27
fix a minor bug introduced in Project save validation

ryanpetrello:
This is what happens when Ryan writes code examples without testing them.


chrismeyersfsu:
so ... did you test this one?

chrismeyersfsu (approved):

Issue #8293, opened 2020-09-30
Misc. system settings form 


marshmalien:
##### ISSUE TYPE
 - Feature Idea

URL: `/next/settings/miscellaneous_system/edit`

* Help text should be displayed in the form field popover
* Handle errors
* Internalization

##### ISSUE TYPE
 - Feature Idea

#### Fields
  * Base URL of the Tower host
	  * Required
	  * URL Text field
  * All Users Visible to Organization Admins
	  * Required
	  * Toggle
  * Organization Admins Can Manage Users and Teams
	  * Required
	  * Toggle
  * Idle Time Force Log Out
	  * Required
	  * Integer
	  * Default: 1800
  * Maximum number of simultaneous logged in sessions
	  * Required
	  * Integer
	  * Default: -1
  * Enable HTTP Basic Auth
	  * Required
	  * Toggle
  * Allow External Users to Create OAuth2 Tokens
	  * Toggle
  * Login redirect override URL
	  * Text
  * Access Token Expiration (OAUTH2_PROVIDER)
	  * Required
	  * Integer
	  * Default: 600
  * Refresh Token Expiration (OAUTH2_PROVIDER)
	  * Required
	  * Integer
	  * Default: 2628000
  * Authorization Code Expiration (OAUTH2_PROVIDER)
	  * Required
	  * Integer
	  * Default: 31536000000
  * Remote Host Headers
	  * Required
	  * CodeMirror
  * Custom virtual environment paths
	  * CodeMirror
  * Gather data for Automation Analytics
	  * Toggle
  * Red Hat customer username
	  * Text
  * Red Hat customer password
	  * Encrypted
  * Automation Analytics upload URL
	  * URL text field
  * Automation Analytics Gather Interval
	  * Integer
	  * Default: 14400
  * Action Buttons
	  * Save
	  * Cancel
	  * Revert all

<img width="927" alt="Screen Shot 2020-09-30 at 5 12 37 PM" src="https://user-images.githubusercontent.com/15881645/94740166-29435180-0340-11eb-9737-e5a471db8625.png">

unlikelyzero:
covered by https://github.com/ansible/tower-qa/pull/5734

Issue #8292, opened 2020-09-30
Logging system form 


marshmalien:
##### ISSUE TYPE
 - Feature Idea

URL: `/next/settings/logging/edit`

* Help text should be displayed in the form field popover
* Handle errors
* Internalization

##### ISSUE TYPE
 - Feature Idea

#### Fields
  * Logging aggregator
	  * Text
  * Logging aggregator port
	  * Text
  * Logging aggregator type
	  * Select
  * Logging aggregator username
	  * Text
  * Logging aggregator password/token
	  * Encrypted
  * Loggers sending data to log aggregator form
	  * CodeMirror
  * Log system tracking facts individually
	  * Toggle
  * Logging aggregator protocol
	  * Select
  * TCP connection timeout
	  * Required
	  * Hide if logging aggregator protocol === "UDP"
  * Logging aggregator level threshold
	  * Select
  * Enable/Disable HTTPS certificate verification
	  * Switch
  * Action Buttons
	  * Save
	  * Cancel
	  * Revert all
	  * Test Logging
* Enable External Logging
    * In the old ui, this toggle is disabled when: 
```
disabled="(!LOG_AGGREGATOR_ENABLED && 
(!requiredLogValues.LOG_AGGREGATOR_HOST || 
requiredLogValues.LOG_AGGREGATOR_HOST === '' || 
!requiredLogValues.LOG_AGGREGATOR_TYPE || 
requiredLogValues.LOG_AGGREGATOR_TYPE === '')) || 
$rootScope.user_is_system_auditor"
```

unlikelyzero:
covered by https://github.com/ansible/tower-qa/pull/5734

Issue #8287, opened 2020-09-30
Activity stream settings form 


marshmalien:
##### ISSUE TYPE
 - Feature Idea

URL: `/next/settings/activity_stream/edit`

* Help text should be displayed in the form field popover
* Handle errors
* Internalization

##### ISSUE TYPE
 - Feature Idea

#### Fields
  * Enable Activity Stream
	  * Toggle
	  * Required
  * Enable Activity Stream for Inventory Sync
	  * Toggle
	  * Required
  * Action Buttons
	  * Save
	  * Cancel
	  * Revert all

unlikelyzero:
covered by https://github.com/ansible/tower-qa/pull/5734

Issue #8286, opened 2020-09-30
AzureAD settings form 


marshmalien:
##### ISSUE TYPE
 - Feature Idea

URL: `/next/settings/azure/edit`

* Help text should be displayed in the form field popover
* Handle errors
* Internalization

#### Fields
  * Azure AD OAuth2 Callback URL
	  * Read-only
	  * Depends on `TOWER_URL_BASE`
	  * No revert 
  * Azure AD OAuth2 Key
	  * Text field
  * Azure AD OAuth2 Secret
	  * Encrypted
  * Azure AD OAuth2 Organization Map
	  * CodeMirror field
  * Azure AD OAuth2 Team Map
	  * CodeMirror field
  * Action Buttons
	  * Save
	  * Cancel
	  * Revert all

<img width="804" alt="Screen Shot 2020-09-30 at 4 14 47 PM" src="https://user-images.githubusercontent.com/15881645/94735013-13ca2980-0338-11eb-9d47-7c40d46cced3.png">

unlikelyzero:
covered by https://github.com/ansible/tower-qa/pull/5734

PR #8472, opened 2020-10-27
Add setting "system" category forms and tests

marshmalien:
### SUMMARY
Activity Stream: [https://github.com/ansible/awx/issues/8287](https://github.com/ansible/awx/issues/8287)
Logging: [https://github.com/ansible/awx/issues/8292](https://github.com/ansible/awx/issues/8292)
Misc System: [https://github.com/ansible/awx/issues/8293](https://github.com/ansible/awx/issues/8293)
AzureAD: https://github.com/ansible/awx/issues/8286

Adds "System" setting category forms: Activity Stream, Logging, and Miscellaneous System. The patterns used in this area are repeated in all setting forms, so I would like feedback on the technical approach and UX. 

The new field revert button will behave differently from the old ui revert button. In the old UI, clicking the field-level revert button executes the request immediately; it sends a PATCH request to the API without requiring the user to confirm the action in an alert modal. In the new UI, the revert button does not execute the request immediately; instead, the button will update the form with the system default value and switch the button text from "revert" to "undo". If the user is happy with their selection, they can choose to save the form. If the user selected revert by accident, the undo button will allow the user to recover from their action by reverting the field value back to the initial form value. The revert button will appear disabled if the current, initial, and default field values are all equal.

setting-undo

### RBAC
To guarantee that only superusers have permission to view setting forms, the edit route checks that the user has system administration privileges before rendering the edit component. If the user does not have system administration privileges, a Redirect will take the user back to the setting details view.

```jsx
{me?.is_superuser ? (
  <ActivityStreamEdit />
) : (
  <Redirect to={`${baseURL}/details`} />
)}
```

### Additional components and hook
- useModal - handles opening and closing a modal
- RevertAllAlert modal - Confirmation alert modal shown when the user chooses to "Revert all".
- RevertButton - Button that handles reverting the form field value to either the default system setting value or initial form value.
- RevertFormActionGroup - Form action buttons that include Save, Revert all to default, and Cancel by default.
- LoggingTestAlert toast - Toast alert with the status of the test log message.

test-log
test-log-err


awx/ui_next/src/components/FormActionGroup/FormActionGroup.jsx
```diff
@@ -24,16 +18,6 @@ const FormActionGroup = ({
...
-        {onRevert && (
-          <Button
-            aria-label={i18n._(t`Revert`)}
-            variant="secondary"
-            type="button"
-            onClick={onRevert}
-          >
-            {i18n._(t`Revert`)}
-          </Button>
-        )}
```
marshmalien, line 36: Since the ability to revert is a settings-only feature, I removed the onRevert button from the top-level FormActionGroup component and created a custom  action button group for Settings called RevertFormActionGroup. The RevertFormActionGroup component allows the form to pass in additional action buttons as `{ children }` and keeps the top-level FormActionGroup generic.

awx/ui_next/src/api/models/Settings.js
```diff
@@ -21,6 +21,10 @@ class Settings extends Base {
   readCategoryOptions(category) {
     return this.http.options(`${this.baseUrl}${category}/`);
   }
+
+  test(category, data) {
```
jlmitch5: I wonder if we should attempt to rename this with a verb so the REST action is clear...createTest maybe?

awx/ui_next/src/screens/Setting/shared/RevertAllAlert.jsx
```diff
@@ -0,0 +1,39 @@
...
+          key="cancel"
+          variant="secondary"
+          aria-label={i18n._(t`Cancel revert`)}
+          onClick={onClose}
+        >
+          {i18n._(t`Cancel`)}
+        </Button>,
+      ]}
+    >
+      {i18n._(t`This will revert all configuration values to their
```
jlmitch5: Should we indicate in this message that revert all is for the particular form you are on, not every setting configuration value across all forms?

awx/ui_next/src/screens/Setting/shared/RevertButton.jsx
```diff
@@ -0,0 +1,66 @@
...
+import PropTypes from 'prop-types';
+import { withI18n } from '@lingui/react';
+import { t } from '@lingui/macro';
+import { useField } from 'formik';
+import { Button, Tooltip } from '@patternfly/react-core';
+import styled from 'styled-components';
+
+const ButtonWrapper = styled.div`
+  margin-left: auto;
+  &&& {
```
jlmitch5, line 11: TIL about this trick for bumping the specificity, cool stuff

awx/ui_next/src/screens/Setting/shared/SharedFields.jsx
```diff
@@ -0,0 +1,258 @@
...
+    </FormFullWidthLayout>
+  );
+});
+ObjectField.propTypes = {
+  name: string.isRequired,
+  config: shape({}).isRequired,
+  isRequired: bool,
+};
+
+export { BooleanField, ChoiceField, EncryptedField, InputField, ObjectField };
```
jlmitch5, line 260: very cool we've been able to keep this pseudo-form generator very tight!

awx/ui_next/src/screens/Setting/shared/ShareFields.test.jsx
```diff
@@ -0,0 +1,152 @@
+import React from 'react';
```
jlmitch5: This file is supposed to be named "SharedFields" I think

awx/ui_next/src/screens/Setting/ActivityStream/ActivityStreamEdit/ActivityStreamEdit.jsx
```diff
@@ -1,25 +1,137 @@
...
+  RevertAllAlert,
+  RevertFormActionGroup,
+} from '../../shared';
+import useModal from '../../../../util/useModal';
+import useRequest from '../../../../util/useRequest';
+import { SettingsAPI } from '../../../../api';
+
+function ActivityStreamEdit() {
+  const history = useHistory();
+  const { isModalOpen, toggleModal, closeModal } = useModal();
```
jlmitch5, line 22: Slick way to handle this kind of thing!

awx/ui_next/src/screens/Setting/Logging/LoggingEdit/LoggingEdit.test.jsx
```diff
@@ -1,16 +1,318 @@
...
+      wrapper = mountWithContexts(
+        <SettingsProvider value={mockAllOptions.actions}>
+          <LoggingEdit />
+        </SettingsProvider>
+      );
+    });
+    await waitForElement(wrapper, 'ContentLoading', el => el.length === 0);
+    expect(wrapper.find('ContentError').length).toBe(1);
+  });
 });
```
jlmitch5, line 318: extremely thorough tests here!

jlmitch5 (commented):
Looks great!  I think the settings code has turned out incredibly simple for how much is going on, great job.  Just a few small things, mostly UX/wording related and not really code related.

awx/ui_next/src/screens/Setting/shared/ShareFields.test.jsx
marshmalien: 😄 I think you're right.

awx/ui_next/src/screens/Setting/shared/RevertAllAlert.jsx
marshmalien: I consulted with @tvo318 on this one. What do you think about:
```
This will revert all configuration values on this page to their factory defaults. Are you sure you want to proceed?
```

awx/ui_next/src/screens/Setting/AzureAD/AzureADEdit/AzureADEdit.jsx
```diff
@@ -1,25 +1,146 @@
...
+    useCallback(async values => {
+      const result = await SettingsAPI.updateAll(values);
+      return result;
+    }, []),
+    null
+  );
+
+  useEffect(() => {
+    if (submitResult) {
+      history.push('/settings/azure/details');
```
AlexSCorey: Curious why you are navigating this way instead of doing the navigation inside of the useRequest [here](https://github.com/ansible/awx/pull/8472/files#diff-eb84bb28b0549e1934e10a2c5de60ac40f5224fccd28a4a414fa0813c65b0c8bR50).  Is it possible that a submitResult not be returned?

awx/ui_next/src/api/models/Settings.js
AlexSCorey: I agree, I think this should be createTest as well.

awx/ui_next/src/screens/Setting/ActivityStream/ActivityStreamEdit/ActivityStreamEdit.jsx
```diff
@@ -1,25 +1,137 @@
...
+  const { PUT: options } = useSettings();
+
+  const {
+    isLoading,
+    error,
+    request,
+    result: {
+      ACTIVITY_STREAM_ENABLED,
```
AlexSCorey, line 30: I'm seeing this all caps, constant syntax more and more.  Is this the direction the team is moving in?  I like it and think it makes sense.

awx/ui_next/src/screens/Setting/Logging/LoggingEdit/LoggingEdit.jsx
```diff
@@ -1,23 +1,272 @@
...
+                    config={logging.LOG_AGGREGATOR_LEVEL}
+                  />
+                  {['tcp', 'https'].includes(
+                    formik.values.LOG_AGGREGATOR_PROTOCOL
+                  ) && (
+                    <InputField
+                      name="LOG_AGGREGATOR_TCP_TIMEOUT"
+                      config={logging.LOG_AGGREGATOR_TCP_TIMEOUT}
+                      type="number"
+                      isRequired
```
nixocio, line 213: What about we add more specific validator for `type="number"` ? We do have `number` or `integer` validators.

image

AlexSCorey (commented):
I think this looks really good. I can the tests locally and they all pass. You were able to address all of my concerns offline.  I did find some thing about the tooltips, which you said you'd look into.  Overall, awesome work!

awx/ui_next/src/screens/Setting/AzureAD/AzureADEdit/AzureADEdit.jsx
marshmalien: Both patterns achieve the same goal. It is possible that a submitResult not be returned if there is an error or if the request function does not return a result. I've seen both patterns used in our codebase, but I'm open to moving navigation into the request function, for brevity.

nixocio (commented):
Great job @marshmalien. I got this code locally, and it seems to be working as expected. I left a minor suggestion.

awx/ui_next/src/screens/Setting/ActivityStream/ActivityStreamEdit/ActivityStreamEdit.jsx
marshmalien: This is all caps because the fields are all caps in the API response. I wasn't trying to change the syntax, more so, I was trying to avoid doing a destructuring reassignment for the sake of casing.

awx/ui_next/src/screens/Setting/shared/RevertAllAlert.jsx
jlmitch5: Yes that's good!

awx/ui_next/src/screens/Setting/Logging/LoggingEdit/LoggingEdit.jsx
marshmalien: Of course Firefox will let us type non-numeric characters in a `type="number"` input field 🙃See:https://bugzilla.mozilla.org/show_bug.cgi?id=1398528
If I remove the type attribute from TextInput, the validators will work as expected.
input

unlikelyzero:
Feedback:
- If entering an incorrect format into a CodeMirror component (i.e. Azure AD Map), the except is thrown to console without feedback to the user
- There doesn't seem to be a unique attribute associated with the individual 'Reverts' or 'Undo'. Is that something you could add via ID or aria?
- Error on when clicking into `/next/settings/miscellaneous_system/edit`page. 
```
TypeError: Cannot set property 'value' of undefined at https://localhost:3001/static/js/main.chunk.js:87121:29 at Array.forEach (<anonymous>) at https://localhost:3001/static/js/main.chunk.js:87119:29 at async https://localhost:3001/static/js/main.chunk.js:113038:26
```

jlmitch5 (approved):
looks good!

marshmalien:
@unlikelyzero 
> If entering an incorrect format into a CodeMirror component (i.e. Azure AD Map), the except is thrown to console without feedback to the user.

Updated PR to correctly handle code field error messages.
Screen Shot 2020-11-03 at 2 36 17 PM

>There doesn't seem to be a unique attribute associated with the individual 'Reverts' or 'Undo'. Is that something you could add via ID or aria?

All revert buttons should have an aria-label="Revert". I added a data-cy prop that concats the field id and "-revert" together. Example: `data-cy=AUTOMATION_ANALYTICS_URL-revert`. 

>Error on when clicking into /next/settings/miscellaneous_system/edit page.

Updated PR to handle rendering empty fields.

awx/ui_next/src/screens/Setting/shared/RevertButton.jsx
```diff
@@ -45,7 +45,7 @@ function RevertButton({ i18n, id, defaultValue, isDisabled = false }) {
       <ButtonWrapper>
         <Button
           aria-label={isRevertable ? i18n._(t`Revert`) : i18n._(t`Undo`)}
-          data-cy={`${id}-revert`}
+          data-ouia-id={`${id}-revert`}
```
quarckster: it should be `ouiaId`:
```js
ouiaId={`${id}-revert`}
```
during the app building it will be transformed into `data-ouia-component-id`

quarckster (changes requested):

awx/ui_next/src/screens/Setting/shared/RevertButton.jsx
marshmalien: Thanks! I can confirm that adding `ouiaId` to the Revert Button transforms into the following additional attributes:
* data-ouia-component-type="PF4/Button" 
* data-ouia-safe="true" 
* data-ouia-component-id="TOWER_URL_BASE-revert"

jakemcdermott (approved):

unlikelyzero:
Need to update this line to provide better instruction to user on how the main logging toggle can be enabled. https://github.com/ansible/awx/blob/10664d1931d789c281e2c9d3d2f14c0a69cb9e13/awx/main/conf.py#L659

unlikelyzero:
recheck

unlikelyzero:
@marshmalien just a single defect was found on this last pass:
'Reverting' the 'Base URL of the Tower host' under /Miscellaneous_system/edit changes the URL to localhost

marshmalien:
@unlikelyzero Is that different from the old UI? The field should revert to the same value that is returned from `/api/v2/settings/all/` OPTIONS response. 
```
"TOWER_URL_BASE": {
  "type": "string",
  "required": true,
  "label": "Base URL of the Tower host",
  "help_text": "This setting is used by services like notifications to render a valid url to the Tower host.",
  "category": "System",
  "category_slug": "system",
  "default": "https://localhost:3001"
},
```

unlikelyzero:
covered by https://github.com/ansible/tower-qa/pull/5734

PR #8483, opened 2020-10-28
Supports Associating Related Inventory Group

AlexSCorey:
#8261 

This also adds support for creating a new inventory group and
associating it at the same time.

##### SUMMARY
This supports associate/add functionality for Inventory Related Groups List.  It also introduces a slightly new pattern for navigating to the list via `Back to Groups` tab.  The use case there was we needed to be have access to both the Related Groups list we navigated from, to create a new Inventory Group.  On Save we are navigated to the details page. On the details page (the same component we use to create a new inventory group from the Inventory Group List)we have tabs labeled `Back to Groups` (which goes back to the Inventory Related Group List we were on when we clicked to create a new Inventory Group, `prevGroupId`) `Details`, `Related Groups` (which navigates to the related groups list for this group we just created.  This list will initially be empty as this group was just created, `groupId`), `Hosts`.  

https://github.com/ansible/awx/compare/devel...AlexSCorey:8261-AddRelatedGroups?expand=1#diff-49a8ef58c64d4f46d7d874a522bb1705779573919b4793d1d2adbf6b8abfca24R57

I deliberately haven't written any tests for this bit of code because I wanted some feedback on it before moving forward.  Everything else has tests.
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
Screen Shot 2020-10-28 at 3 59 35 PM
Screen Shot 2020-10-28 at 3 59 15 PM
Screen Shot 2020-10-28 at 3 59 06 PM

### DIFF
--- a/awx/ui_next/src/api/models/Groups.js
+++ b/awx/ui_next/src/api/models/Groups.js
@@ -37,6 +37,17 @@ class Groups extends Base {
   readChildren(id, params) {
     return this.http.get(`${this.baseUrl}${id}/children/`, { params });
   }
+
+  associateChildGroup(id, childId) {
+    return this.http.post(`${this.baseUrl}${id}/children/`, { id: childId });
+  }
+
+  disassociateChildGroup(id, childId) {
+    return this.http.post(`${this.baseUrl}${id}/children/`, {
+      disassociate: id,
+      id: childId,
+    });
+  }
 }
 
 export default Groups;
--- a/awx/ui_next/src/api/models/Groups.js
+++ b/awx/ui_next/src/api/models/Groups.js
@@ -48,6 +48,12 @@ class Groups extends Base {
       id: childId,
     });
   }
+
+  readPotentialGroups(id, params) {
+    return this.http.get(`${this.baseUrl}${id}/potential_children/`, {
+      params,
+    });
+  }
 }
 
 export default Groups;
--- a/awx/ui_next/src/screens/Inventory/Inventories.jsx
+++ b/awx/ui_next/src/screens/Inventory/Inventories.jsx
@@ -69,6 +69,12 @@ function Inventories({ i18n }) {
         [`${inventoryGroupsPath}/${nested?.id}/nested_hosts/add`]: i18n._(
           t`Create new host`
         ),
+        [`${inventoryGroupsPath}/${nested?.id}/nested_groups`]: i18n._(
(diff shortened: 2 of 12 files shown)


mabashian:
It looks like we're missing the breadcrumb on the related groups list:

Screen Shot 2020-10-29 at 10 17 55 AM

Here's what related hosts looks like:

Screen Shot 2020-10-29 at 10 18 01 AM

mabashian:
So I do have some thoughts on:

`Save we are navigated to the details page`

We kind of have two different flows for the back button on a group's details view now and I'm wondering if we should just pick one behavior.

As you pointed out, in this new flow where a user creates a new related group, the back button on the details takes you back to the _related groups_ list from whence you came.

new_related_group

If you then navigate to the details view for this group (via related groups) and hit the back tab it takes you all the way back out to the root level groups.

existing_related_group

If possible, I feel like we should pick one of these patterns for both scenarios.

marshmalien:
Alex and I chatted more about this flow and came up with a simplified solution. The "Back to ***List***" tab and its corresponding ***List*** breadcrumb should navigate to the same location. This matches the pattern we have in the rest of the app and removes the need to track the location you navigated from. Thoughts? 👍 👎 

Screen Shot 2020-10-30 at 12 07 34 PM
Screen Shot 2020-10-30 at 12 08 53 PM
Screen Shot 2020-10-30 at 12 20 38 PM

unlikelyzero:
@tiagodread 
To add coverage for the navigation between hosts / inventory contexts https://github.com/ansible/awx/pull/8483#issuecomment-719652094

tiagodread:
Hey @AlexSCorey I found some issues:

- [x] The select group modal for existing groups shows groups that actually are already related, in the old UI this groups already related is not displayed:

related1

- [x] Trying to add the related group `group 1` to group `root group` and then going to group `group 1` and add the related group `root group` the user see the error: `Cyclical Group association.` Note: The same problem occurs on the old UI and this is a good opportunity to fix it for the UI next. 

related2

- [x] Missing aria-labels on `Back to Groups`

tiagodread:
@AlexSCorey the basic and advanced search also sort feature in the `Add existing group` modal doesn’t work after the last commit.

tiagodread (approved):
Tests covered by 5683

PR #8518, opened 2020-11-02
Change import/export Collections Module to Detect Errors vs Warnings

beeankha:
##### SUMMARY

The `log_contents` error detection was picking up on `WARNING` level logs vs `ERROR` and above.   If an unprivileged user tries to import/export assets that they don't have permissions to, they get a warning; it is not necessary in this case to make the module fail out completely, since other assets will still get imported/exported.

##### ISSUE TYPE

 - Collections Pull Request

##### AWX VERSION

```
awx: 15.0.1
```


beeankha:
Closing in order to make a new PR in the correct place.

Issue #8098, opened 2020-09-08
Inventory does not update host list if sources are deleted


dsesami:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Essentially, if you launch a template with an inventory that was synced with a source, and then delete that source, the inventory then cannot be synced and becomes "stale".

It should be noted that deleting an inventory source deletes its respective hosts automatically in the old UI. 

##### STEPS TO REPRODUCE
1. Create a job template, WFJT, whatever can utilize a specific inventory.
2. In the inventory, add a source from the Sources tab > using ansible/test-playbooks > from Project root > something like "Create 10 hosts" as your initialization file.
3. Sync the source so 10 hosts appear. these are blank hosts, so they won't work.
4. Launch a job. As expected, it fails as it cannot connect to blank hosts.
5. Go back to the inventory and delete the bad inventory source.
6. Go back to the job and try to launch it again.

##### EXPECTED RESULTS
There is some way to update/re-sync the inventory automatically so it only uses the default localhost host and passes, after the inventory source with the empty hosts is deleted.

##### ACTUAL RESULTS
The host list cannot be re-synced as a whole, and the empty hosts are "stale" and stay. They can be manually deleted from the Inventory > Hosts tab, but this seems non-ideal. The old UI deletes this automatically.

nixocio:
In order to recreate this issue.

Go to`Inventories>{anyInventory}Sources>Create new source`.

Create a source inventory which related project contains a file like, `foo.ini`:

```
host1
host2
host3
host4
host5
host6
host7
host8
host9
host10

```

mabashian:
@unlikelyzero moving this to needs_test since #8519 has merged.  Feel free to close if you already have coverage.

unlikelyzero:
closing with https://github.com/ansible/awx/pull/8519#pullrequestreview-530295121

PR #8519, opened 2020-11-02
Remove groups/hosts when deleting inventory sources

nixocio:
Remove related resources groups/hosts when deleting inventory sources.

The current UI deletes `groups` and `hosts` once the inventory source is
deleted. Add this behavior to the new UI.

See: https://github.com/ansible/awx/issues/8098

### DIFF
--- a/awx/ui_next/src/screens/Inventory/InventorySources/InventorySourceList.jsx
+++ b/awx/ui_next/src/screens/Inventory/InventorySources/InventorySourceList.jsx
@@ -105,8 +105,37 @@ function InventorySourceList({ i18n }) {
   );
   const { error: syncError, dismissError } = useDismissableError(syncAllError);
 
+  const deleteRelatedInventoryResources = resourceId => {
+    return [
+      InventorySourcesAPI.destroyHosts(resourceId),
+      InventorySourcesAPI.destroyGroups(resourceId),
+    ];
+  };
+
+  const {
+    isLoading: deleteRelatedResourcesLoading,
+    deletionError: deleteRelatedResourcesError,
+    deleteItems: handleDeleteRelatedResources,
+  } = useDeleteItems(
+    useCallback(async () => {
+      return (
+        Promise.all(
+          selected
+            .map(({ id: resourceId }) =>
+              deleteRelatedInventoryResources(resourceId)
+            )
+            .flat()
+        ),
+        []
+      );
+    }, [selected])
+  );
+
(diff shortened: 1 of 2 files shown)


awx/ui_next/src/screens/Inventory/InventorySources/InventorySourceList.jsx
```diff
@@ -105,8 +105,33 @@ function InventorySourceList({ i18n }) {
   );
   const { error: syncError, dismissError } = useDismissableError(syncAllError);
 
+  const deleteRelatedInventoryResources = resourceId => {
```
jlmitch5, line 108: the use of async is confusing here.  it doesn't look like you're returning anything from this function.  so selected.map would return an array of undefineds...so the Promise.all below will never resolve  (or will instantly resolve regardless of if the functions are  called.  I believe this will work correctly:

```
const deleteRelatedInventoryResources = resourceId => {
    return [InventorySourcesAPI.destroyHosts(resourceId), 
    InventorySourcesAPI.destroyGroups(resourceId)];
  };

  const {
    isLoading: deleteRelatedResourcesLoading,
    deletionError: deleteRelatedResourcesError,
    deleteItems: handleDeleteRelatedResources,
  } = useDeleteItems(
    useCallback(async () => {
      return (
        Promise.all(
          selected.map(({ id: resourceId }) =>
            deleteRelatedInventoryResources(resourceId)
          ).flat();
          console.log("you'll get here after everything is resolved");
        ),
        []
      );
    }, [selected])
  );
```

jlmitch5 (changes requested):
Looks good, one small thing.

awx/ui_next/src/screens/Inventory/InventorySources/InventorySourceList.jsx
nixocio: @jlmitch5, thanks for your review. Updating now.

AlexSCorey (commented):
I reached out to Kerom about the same issue that John found re: async stuff.  Once that's addressed I'm ok with this.

jlmitch5 (approved):
looks good!

tiagodread (approved):
Verified that:

1. After deleting a inventory source all hosts and groups imported are successfully deleted
2. Groups created manually after import hosts/groups from source are not deleted when the inventory source is deleted (working as the old ui)

I'm still working on the UI tests to cover this

Issue #7636, opened 2020-07-16
[ui_next] Surveys cannot be enabled via Survey Toggle for WFJT


unlikelyzero:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Surveys cannot be enabled via Survey Toggle for WFJT

##### ENVIRONMENT
* AWX version: 13

##### STEPS TO REPRODUCE
1. Create a WFJT
2. Navigate to Survey tab and create a survey
3. Click 'Enable Surveys toggle'
Note: Error message

##### EXPECTED RESULTS
Survey toggled on

##### ACTUAL RESULTS
Survey not toggled on
Error message

unlikelyzero:
Covered in workflow-prompts-and-surveys.spec

PR #8520, opened 2020-11-02
Toggles WFJT survey on and off

AlexSCorey:
##### SUMMARY
This addresses #7636 to toggle a wfjt survey and adds unit tests

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
toggle survey

### DIFF
--- a/awx/ui_next/src/screens/Template/TemplateSurvey.jsx
+++ b/awx/ui_next/src/screens/Template/TemplateSurvey.jsx
@@ -68,11 +68,17 @@ function TemplateSurvey({ template, canEdit, i18n }) {
 
   const { request: toggleSurvey, error: toggleError } = useRequest(
     useCallback(async () => {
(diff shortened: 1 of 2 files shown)


jakemcdermott (approved):

unlikelyzero:
@AlexSCorey @marshmalien Is this toggle different than the implementation within https://github.com/ansible/awx/pull/8472 ?
I'm not seeing `checked` as a property to test against

marshmalien:
@unlikelyzero  Both toggles use the same PF React Switch component. I don't think we should depend on the value of the `checked` attribute in the browser's HTML inspector. However, the DOM property does show the current `checked` value. In Chrome you can find this information under the Properties tab.

Screen Shot 2020-11-03 at 3 53 47 PM

PR #8526, opened 2020-11-03
Convert WorkflowJobTemplate.jsx to functional component

mabashian:
##### SUMMARY
This alleviates an issue where several extraneous network requests are made each time the workflow node wizard is closed.  Rather than rolling this in to workflow node POL I figured it would be better to have it's own PR.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/screens/Template/WorkflowJobTemplate.jsx
+++ b/awx/ui_next/src/screens/Template/WorkflowJobTemplate.jsx
@@ -1,317 +1,265 @@
-import React, { Component } from 'react';
+import React, { useEffect, useCallback } from 'react';
 import { t } from '@lingui/macro';
 import { withI18n } from '@lingui/react';
 import { CaretLeftIcon } from '@patternfly/react-icons';
 import { Card, PageSection } from '@patternfly/react-core';
-import { Switch, Route, Redirect, withRouter, Link } from 'react-router-dom';
+import {
+  Switch,
+  Route,
+  Redirect,
+  Link,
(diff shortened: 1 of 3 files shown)


jlmitch5 (approved):
i love a good functional component conversion, nice stuff

awx/ui_next/src/screens/Template/WorkflowJobTemplate.jsx
```diff
@@ -1,317 +1,265 @@
...
+              />
+            </Route>
+          )}
+          {!hasRolesandTemplateLoading && (
             <Route key="not-found" path="*">
               <ContentError isNotFound>
                 {match.params.id && (
                   <Link
-                    to={`/templates/workflow_job_template/${match.params.id}/details`}
+                    to={`/templates/${match.params.templateType}/${match.params.id}/details`}
```
AlexSCorey, line 250: since you already are using `useParams()` above, might as well use the `templateType` and `id` value from it here too. It would shorten this up a bit.

AlexSCorey (approved):
This looks good to me.  I don't have any blockers so I'm going to approve, but if feel inclined to update the useParams piece please do so. There were other issues that I mentioned offline, which have already been addressed.

Good work!

unlikelyzero:
@mabashian please run the existing yolok tests. There should be coverage and we're explicitly waiting for certain requests to complete after WFViz close

Issue #4968, opened 2019-10-09
Host Config Key field should be marked as required when provisioning callback checkbox checked


mabashian:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
A few differences I noticed between the existing implementation of provisioning callbacks and the ui_next version.

1. Host Config Key should have a little * next to it to denote that it's required when the checkbox is checked.
~~2. With provisioning callbacks enabled, a successful save of the form doesn't show the old modal:~~

trahman73:
@mabashian In current Tower, I believe the modal was necessary to point out the Host Config Key. Now that we have a Details view for ui-next, my opinion is that we no longer need the modal.

mabashian:
Refactoring this issue - we won't worry about the modal but we should update the field to display as required:

<img width="1371" alt="Screen Shot 2020-09-25 at 11 12 57 AM" src="https://user-images.githubusercontent.com/9889020/94284394-492ddc00-ff20-11ea-8976-28bfdb2b2d0d.png">
<img width="612" alt="Screen Shot 2020-09-25 at 11 13 18 AM" src="https://user-images.githubusercontent.com/9889020/94284395-49c67280-ff20-11ea-8b09-9d9d6058fa38.png">

tiagodread:
PR already merged

PR #8545, opened 2020-11-05
Mark `Host Config Key` as a required field

nixocio:
Mark `Host Config Key` as a required field

See:https://github.com/ansible/awx/issues/4968

image


jakemcdermott (approved):

tiagodread (approved):
👍🏽  Worked as expected

PR #8580, opened 2020-11-12
forcibly close DB and cache socket connection post-fork

ryanpetrello:
we've seen evidence of a race condition on fork for `awx.conf.model.Setting`
access; in the past, we've attempted to solve this by explicitly closing
connections pre-fork, but we've seen evidence that this isn't always
good enough

this patch is an attempt to close connections post-fork so that sockets
aren't inherited post fork, leading to bizarre race conditions in
setting access

### DIFF
--- a/awx/conf/settings.py
+++ b/awx/conf/settings.py
@@ -4,6 +4,7 @@ import logging
 import sys
 import threading
 import time
+import os
 
 # Django
 from django.conf import LazySettings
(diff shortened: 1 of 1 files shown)


awx/conf/settings.py
```diff
@@ -255,6 +257,16 @@ def __init__(self, default_settings, cache, registry):
         self.__dict__['cache'] = EncryptedCacheProxy(cache, registry)
         self.__dict__['registry'] = registry
 
+    def __clean_on_fork__(self):
+        pid = os.getpid()
+        if pid not in self.__dict__['__forks__']:
+            self.__dict__['__forks__'][pid] = True
```
ryanpetrello, line 273: We have to access this via `__dict__` instead of `getattr` to avoid a recursive loop (because `__getattr__` calls *this* function via `_get_local`).

ryanpetrello: I also tested this out by printing this dict in a tight loop of `awx-python` forks, and the data structure for each process never grows beyond two keys (the parent pid, and the current process pid).

So in effect, this branch is entered *one time* when a child process does `getattr(settings, ...`), and then after that, it's a no-op.

matburt (approved):
This harkens back to when Ansible would change pids occasionally on us during event processing years ago.

PR #8582, opened 2020-11-12
Revert portions of 40a10dcc5f00d60284aa507b418b5e758094d437

wenottingham:
Keep some of the collectors, while removing ones we calculate
in other ways.

### DIFF
--- a/awx/main/analytics/collectors.py
+++ b/awx/main/analytics/collectors.py
@@ -65,6 +65,7 @@ def config(since, **kwargs):
     }
 
 
+@register('counts', '1.0', description=_('Counts of objects such as organizations, inventories, and projects'))
 def counts(since, **kwargs):
     counts = {}
     for cls in (models.Organization, models.Team, models.User,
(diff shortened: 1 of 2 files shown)


awx/main/analytics/collectors.py
```diff
@@ -86,7 +86,6 @@ def counts(since, **kwargs):
     inv_counts['smart'] = inv_counts.get('smart', 0)
     counts['inventories'] = inv_counts
     
-    counts['unified_job'] = models.UnifiedJob.objects.exclude(launch_type='sync').count() # excludes implicit project_updates
     counts['active_host_count'] = models.Host.objects.active_count()
```
Ladas: @wenottingham hm, how would we use the the active host count? From code it does:

```
self.order_by().exclude(inventory_sources__source='tower').values('name').distinct().count()
```

So unique host count across all inventories? Not sure why we're excluding tower source. We can get this, only if you're running some jobs on those inventories.

Ladas: And how this relates to `inventory_counts` below

wenottingham: It's a de-duped count of hosts across inventories. The reason that tower source is included is because hosts sourced from another Tower instance don't "count" for licensing reasons (as they are already counted under the master.)

wenottingham: Yea, I just wont' drop this for now

ryanpetrello (approved):
Yep, this looks good.  I'm fine w/ this if @Ladas and @benthomasson are 👍

PR #8588, opened 2020-11-13
Update usage of useDeleteItems

nixocio:
* Update usage of useDeleteItems.
* Remove async when not necessary
* Update return of promises

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function

### DIFF
--- a/awx/ui_next/src/screens/Application/ApplicationDetails/ApplicationDetails.jsx
+++ b/awx/ui_next/src/screens/Application/ApplicationDetails/ApplicationDetails.jsx
@@ -4,7 +4,7 @@ import { t } from '@lingui/macro';
 import { Link, useHistory } from 'react-router-dom';
 import { Button } from '@patternfly/react-core';
 
-import { useDeleteItems } from '../../../util/useRequest';
+import useRequest, { useDismissableError } from '../../../util/useRequest';
 import AlertModal from '../../../components/AlertModal';
 import { CardBody, CardActionsRow } from '../../../components/Card';
 import { Detail, DetailList } from '../../../components/DetailList';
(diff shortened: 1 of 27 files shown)


nixocio:
Currently running e2e tests.

unlikelyzero:
@nixocio if the tests pass, please merge! Right now we have a bug with the yolok pipeline: if a visual test fails, it will impact the results on the functional tests. We have a ticket to fix that https://github.com/ansible/tower-qa/issues/5711

For now, just run this PR without the visual test option

awx/ui_next/src/screens/Application/ApplicationDetails/ApplicationDetails.jsx
```diff
@@ -21,11 +21,11 @@ function ApplicationDetails({
...
     isLoading: deleteLoading,
-    deletionError,
-    deleteItems: handleDeleteApplications,
+    error: deletionError,
+    request: deleteApplications,
     clearDeletionError,
-  } = useDeleteItems(
-    useCallback(async () => {
+  } = useCallback(
+    useRequest(async () => {
```
keithjgrant: This should be reversed to `useRequest(useCallback(...`

awx/ui_next/src/screens/Application/ApplicationDetails/ApplicationDetails.jsx
```diff
@@ -21,11 +21,11 @@ function ApplicationDetails({
   const history = useHistory();
   const {
     isLoading: deleteLoading,
-    deletionError,
-    deleteItems: handleDeleteApplications,
+    error: deletionError,
+    request: deleteApplications,
     clearDeletionError,
```
keithjgrant: useRequest doesn't return a `clearDeletionError`. You'll have to pass `deletionError` into a `useDismissableError` below this to get the clearDeletionError function

keithjgrant (changes requested):
Looks good for the most part. Just a couple issues in one file

awx/ui_next/src/screens/Application/ApplicationDetails/ApplicationDetails.jsx
nixocio: @keithjgrant, updates per comments.

keithjgrant (approved):
lgtm!

nixocio:
I ran the visual tests and e2e tests, no regression was added by those changes.

Issue #8276, opened 2020-09-30
Make org a required field for galaxy credentials


mabashian:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
<img width="1429" alt="Screen Shot 2020-09-30 at 11 29 07 AM" src="https://user-images.githubusercontent.com/9889020/94706492-4102e180-0310-11eb-904f-68234cee0489.png">

Ideally we don't want to see this error in the UI

tiagodread:
Verified that Org is required only for Galaxy type, also added a stub for it in 5744

PR #8594, opened 2020-11-16
Mark organization as a required field for Galaxy crendentials

nixocio:
Mark organization as a required field for Galaxy credential type. 

See: https://github.com/ansible/awx/issues/8276

image

Non required field.

image

### DIFF
--- a/awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx
+++ b/awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx
@@ -22,10 +22,24 @@ function CredentialFormFields({
   initialValues,
 }) {
   const { setFieldValue } = useFormikContext();
-  const [orgField, orgMeta, orgHelpers] = useField('organization');
+
   const [credTypeField, credTypeMeta, credTypeHelpers] = useField({
     name: 'credential_type',
     validate: required(i18n._(t`Select a value for this field`), i18n),
+  });
+
+  const isGalaxyCredential =
+    !!credTypeField.value &&
+    credentialTypes[credTypeField.value].kind === 'galaxy';
+
+  const [orgField, orgMeta, orgHelpers] = useField({
+    name: 'organization',
+    validate:
+      isGalaxyCredential &&
+      required(
+        i18n._(t`Galaxy credentials must be owned by an Organization.`),
+        i18n
+      ),
   });
 
   const credentialTypeOptions = Object.keys(credentialTypes)
@@ -108,6 +122,7 @@ function CredentialFormFields({
         value={orgField.value}
         touched={orgMeta.touched}
         error={orgMeta.error}
+        required={isGalaxyCredential}
       />
       <FormGroup
         fieldId="credential-Type"
--- a/awx/ui_next/src/screens/Credential/shared/CredentialForm.test.jsx
+++ b/awx/ui_next/src/screens/Credential/shared/CredentialForm.test.jsx
@@ -4,6 +4,7 @@ import { mountWithContexts } from '../../../../testUtils/enzymeHelpers';
 import machineCredential from './data.machineCredential.json';
 import gceCredential from './data.gceCredential.json';
 import scmCredential from './data.scmCredential.json';
+import galaxyCredential from './data.galaxyCredential.json';
 import credentialTypesArr from './data.credentialTypes.json';
 import CredentialForm from './CredentialForm';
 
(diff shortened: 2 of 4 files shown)


awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx
```diff
@@ -84,6 +84,16 @@ function CredentialFormFields({
     [setFieldValue]
   );
 
+  function verifyGalaxyCredential(credentialId) {
```
mabashian: I'd suggest changing this to `isGalaxyCredential` or `isGalaxyCredentialType` for readability.

awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx
```diff
@@ -84,6 +84,16 @@ function CredentialFormFields({
...
+    if (credentialId) {
+      return (
+        credentialTypes[credentialId].name ===
```
mabashian: Rather than doing a string match on the name, I think it'd be better to match on the `kind` field.

```
credentialTypes[credentialId].kind === 'galaxy`
```

This is much less likely to change in the future and while I don't think `Ansible Galaxy/Automation Hub API Token` would ever get translated, if it did this logic would break.

awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx
```diff
@@ -108,6 +118,8 @@ function CredentialFormFields({
         value={orgField.value}
         touched={orgMeta.touched}
         error={orgMeta.error}
+        required={verifyGalaxyCredential(credTypeField.value)}
```
mabashian: I suppose you could just store this value in a variable instead of calling a function (since you have to do it twice).

```
  const isGalaxyCredential =
    !!credTypeField.value &&
    credentialTypes[credTypeField.value].kind === 'galaxy'
      ? true
      : false;

...

      <OrganizationLookup
        helperTextInvalid={orgMeta.error}
        isValid={!orgMeta.touched || !orgMeta.error}
        onBlur={() => orgHelpers.setTouched()}
        onChange={onOrganizationChange}
        value={orgField.value}
        touched={orgMeta.touched}
        error={orgMeta.error}
        required={isGalaxyCredential}
        autoPopulate={isGalaxyCredential}
      />
```

mabashian:
Thinking about auto-populating the organization here.  I know that we have this general rule where we want to auto-populate required lookup fields but this case is a little different.  This is the only credential type that requires an org.  If this field gets auto-populated and you change to a different credential type, the selected org is still there which may be undesirable.  We could fix that by wiping the org whenever you change credential types but that could also be undesirable so I'm kind of inclined to say we shouldn't auto-populate the org in this particular scenario.  That matches up with the old UI and I wonder if we did that for a reason.  CC @trahman73  @wenottingham in case y'all have other thoughts.

trahman73:
@mabashian I agree that we shouldn't auto populate this one particular Org field. The backup can be that we auto populate, but remove when cred type is changed as you suggested.

nixocio:
Even though we auto-populate the Organization once `galaxy` kind is selected. Once the credential type is modified we give give the user the option to remove the auto-populated Organization.

image

~It seems to be a sane choice to remove the Organization when the credential kind is different from `galaxy`.~ 

After following up with @mabashian, I better understood the issue here. For the time being it seems to be a better fit to remove the auto-populate for this field. We can always revisit this later on. Thanks @mabashian.

nixocio:
@mabashian, updated per comments.

mabashian (approved):

tiagodread:
Hey @nixocio Now the organization is marked as required for Galaxy and non required for others, but I found a problem trying to edit, steps to reproduce:

1. Edit any credential existed and set the credential type to Ansible Galaxy/Automation Hub API Token

cred

nixocio:
> Hey @nixocio Now the organization is marked as required for Galaxy and non required for others, but I found a problem trying to edit, steps to reproduce:
> 
>     1. Edit any credential existed and set the credential type to Ansible Galaxy/Automation Hub API Token
> 
> 
> cred

~I will take a look.~ Thanks. Done. 

The issue was that we were initially relying on the API validation for this field. Once the credential was created as a different type of `Ansible Galaxy/Automation Hub API Token`, when modifying the API is not raising any error.I added UI validation for this scenario now.

tiagodread (approved):
Verified that Org is required only for Galaxy type, also added a stub for it in 5744

PR #8598, opened 2020-11-17
Initial integration of Receptor into AWX

shanemcd:


awx/main/tasks.py
```diff
@@ -1482,38 +1479,85 @@ def run(self, pk, **kwargs):
...
-                    pod_manager=pod_manager
-                )
-                status, rc = isolated_manager_instance.run(self.instance,
-                                                           private_data_dir,
-                                                           params.get('playbook'),
-                                                           params.get('module'),
-                                                           module_args,
-                                                           ident=str(self.instance.pk))
-                self.finished_callback(None)
+            if not isinstance(self.instance, ProjectUpdate):
```
ryanpetrello: This block is a little weird since they currently both appear to do the same thing.  Maybe it would be better to just leave an `if not isinstance(self.instance, ProjectUpdate):` statement w/ a code comment inside of it for now?

shanemcd: I'm not quite ready for this level of feedback yet, hence the Draft status. This conditional is just a hack that I plan on removing.

awx/main/tasks.py
```diff
@@ -1482,38 +1479,85 @@ def run(self, pk, **kwargs):
...
+                        return result('canceled', 1)
+                    time.sleep(1)
+
+            # Both "processor" and "cancel_watcher" are spawned in separate threads.
+            # We wait for the first one to return. If cancel_watcher returns first,
+            # we yank the socket out from underneath the processor, which will cause it
+            # to exit. A reference to the processor_future is passed into the cancel_watcher_future,
+            # Which exits if the job has finished normally. The context manager ensures we do not
+            # leave any threads laying around.
+            with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
```
ryanpetrello: This is some fancy new python3 niceness, and I think it's a reasonable way to accomplish this, but we'll want to pay careful attention to integration results to make sure we don't notice any sort of weird race conditions/hangs related to our logging code and this threadpool.  This sort of thing has bitten us in the past due to cpython bugs e.g.,

https://bugs.python.org/issue37429
https://github.com/ansible/awx/commit/5364e7839724026
https://bugs.python.org/issue38884

execution-environment.yml
```diff
@@ -2,9 +2,14 @@
...
 dependencies:
   galaxy: requirements/collections_requirements.yml
+  system: requirements/bindep.txt
 additional_build_steps:
   prepend:
     - RUN pip3 install --upgrade pip setuptools
   append:
     - COPY --from=quay.io/project-receptor/receptor /usr/bin/receptor /usr/bin/receptor
     - RUN mkdir -m 0770 /var/run/receptor
+    - RUN pip3 install remote-pdb
```
ryanpetrello, line 12: Did you intend to leave these in?

awx/main/tasks.py
ryanpetrello: Okey-dokey.

awx/main/tasks.py
```diff
@@ -1482,38 +1479,85 @@ def run(self, pk, **kwargs):
...
+
+            # We establish a connection to the Receptor socket and submit our work, passing
+            # in the right side of our socketpair for reading.
+            receptor_ctl = ReceptorControl('/var/run/receptor/receptor.sock')
+            result = receptor_ctl.submit_work(worktype=worktype,
+                                              payload=sockout.makefile('rb'))
+            sockin.close()
+            sockout.close()
+
+            self.update_model(self.instance.pk, receptor_work_id=result['unitid'])
```
AlanCoding: I was really hoping that receptor would let us specify the work ID, because it should be fairly trivial to give a unique ID with the fields already on the unified job model. Then we wouldn't have to muck with tracking a new field. However...

https://github.com/project-receptor/receptor/blob/f7c245c18911a7ea5edb808bf568c39be21b17a6/receptorctl/receptorctl/socket_interface.py#L148-L151

I realize that the receptorctl code shows no sign of supporting this. Need to cross-reference against the `work submit` CLI itself, and I'm a little rusty to get to its help text right now.

jbradberry: cc @ghjm

tools/docker-compose/receptor.conf
```diff
@@ -0,0 +1,27 @@
...
+#     insecureskipverify: true
+
+# - ws-peer:
+#     tls: insecure
+#     address: wss://receptor-receptor.apps.tower.ocp4.testing.ansible.com
+
+- work-kubernetes:
+    worktype: ocp
+    namespace: receptor
+    image: quay.io/shanemcd/ee
```
AlanCoding: Ultimately this will need to come from the execution environment model?

That question might be getting too far ahead of things, and if it is, that's fine.

shanemcd:
Folks, please wait until this PR is marked as ready for review before piling on.

awx/main/tasks.py
shanemcd: I dont think we want a receptor network being tied to a single Tower installation. If we made this value derived from something like the job id, we're going to run into conflicts.

awx/main/tasks.py
shanemcd: Would you feel better with using `ProcessPoolExecutor`?

ryanpetrello: Nope (for reasons we privately discussed) 😄

execution-environment.yml
ryanpetrello: @shanemcd do we still need this (and elsewhere in this PR)

AlanCoding:
recheck

shanemcd:
recheck

awx/main/tasks.py
```diff
@@ -3137,3 +3009,212 @@ def deep_copy_model_obj(
...
+            execution_environment_params = self.task.build_execution_environment_params(self.task.instance)
+            self.runner_params['settings'].update(execution_environment_params)
+
+    def run(self):
+        # Create a socketpair. Where the left side will be used for writing our payload
+        # (private data dir, kwargs). The right side will be passed to Receptor for
+        # reading.
+        sockin, sockout = socket.socketpair()
+
+        threading.Thread(target=self.transmit, args=[sockin]).start()
```
AlanCoding, line 3040: You never do a `.join()` on this do you?

Maybe I'm going to far with this, but some thoughts:

 - it doesn't seem to make sense to start `self.processor` if the transmit thread is still running, so maybe a `.join()` on the transmit thread would be good to have assured before processing starts. Otherwise there's nothing to process.
 - We do need to run the cancel watcher while this is running, because it could get canceled while a large amount of data is being transmitted. Maybe we could poll the cancel callback until that transmit thread finishes?

awx/main/tasks.py
```diff
@@ -3137,3 +3009,212 @@ def deep_copy_model_obj(
...
+        receptor_ctl = ReceptorControl('/var/run/receptor/receptor.sock')
+        result = receptor_ctl.submit_work(worktype=self.work_type,
+                                          payload=sockout.makefile('rb'),
+                                          params=self.receptor_params)
+        unit_id = result['unitid']
+
+        sockin.close()
+        sockout.close()
+
+        resultsock, resultfile = receptor_ctl.get_work_results(unit_id,
```
AlanCoding: this is expecting the return from here right?

https://github.com/project-receptor/receptor/blob/e293394235082df0b5f2cf037a43d13c6a1f3787/receptorctl/receptorctl/socket_interface.py#L233

I'd like to make it a little more clear what `resultfile` is, which comes from `self._sockfile = self._socket.makefile('rwb')`

There's 2 types of things being passed around, sockets and sockfiles. Some are made here, and some come from receptor. I'd like to try to get the naming to be as indicative of this as possible, like `result_sockfile`

awx/main/tasks.py
```diff
@@ -3137,3 +3009,212 @@ def deep_copy_model_obj(
...
+        return res
+
+    # Spawned in a thread so Receptor can start reading before we finish writing, we
+    # write our payload to the left side of our socketpair.
+    def transmit(self, _socket):
+        if self.work_type == 'local':
+            self.runner_params['only_transmit_kwargs'] = True
+
+        ansible_runner.interface.run(streamer='transmit',
+                                     _output=_socket.makefile('wb'),
```
AlanCoding, line 3087: writable sockfile, in the same way that `resultfile` is. Am I getting this correct?

awx/main/tasks.py
```diff
@@ -3137,3 +3009,212 @@ def deep_copy_model_obj(
...
+            cancel_watcher_future = executor.submit(self.cancel_watcher, processor_future)
+            futures = [processor_future, cancel_watcher_future]
+            first_future = concurrent.futures.wait(futures,
+                                                   return_when=concurrent.futures.FIRST_COMPLETED)
+
+            res = list(first_future.done)[0].result()
+            if res.status == 'canceled':
+                receptor_ctl.simple_command(f"work cancel {unit_id}")
+                resultsock.shutdown(socket.SHUT_RDWR)
+                resultfile.close()
```
AlanCoding, line 3072: I don't understand why the shutdown/close of the socket here would be specific to canceling.

awx/main/tasks.py
```diff
@@ -3137,3 +3009,212 @@ def deep_copy_model_obj(
...
+        if self.task.instance.is_container_group_task:
+            work_type = 'ocp'
+        else:
+            work_type = 'local'
+
+        return work_type
+
+    def cancel_watcher(self, processor_future):
+        while True:
+            if processor_future.done():
```
AlanCoding, line 3133: I'm getting more and more worried about the _transmit_ thread hitting an error while this cancel_watcher is serving as the main loop, and is being counted on for the exit condition.

awx/main/tasks.py
```diff
@@ -3137,3 +3009,212 @@ def deep_copy_model_obj(
...
+        # Which exits if the job has finished normally. The context manager ensures we do not
+        # leave any threads laying around.
+        with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
+            processor_future = executor.submit(self.processor, resultfile)
+            cancel_watcher_future = executor.submit(self.cancel_watcher, processor_future)
+            futures = [processor_future, cancel_watcher_future]
+            first_future = concurrent.futures.wait(futures,
+                                                   return_when=concurrent.futures.FIRST_COMPLETED)
+
+            res = list(first_future.done)[0].result()
```
AlanCoding, line 3068: Just trying to follow the code and jot down notes - `[0]` here could give the result from either the processing thread or the cancel watcher thread. It just doesn't matter because they give the same format of output, and the cancel watcher may outright give the result from the processing thread.

AlanCoding:
I want to test the scenario where the remote job is killed half-way through. I suspect that `ansible-runner process` may not be designed in a way that can realize that an exit. The only way I can see us dealing with that is adding a check for `receptorctl status` inside of the cancel callback loop - basically reaper logic inside the controller process.

awx/main/tasks.py
```diff
@@ -3137,3 +3009,212 @@ def deep_copy_model_obj(
...
+
+    @property
+    def receptor_params(self):
+        if self.task.instance.is_container_group_task:
+            spec_yaml = yaml.dump(self.pod_definition, explicit_start=True)
+            kubeconfig_yaml = yaml.dump(self.kube_config, explicit_start=True)
+
+            receptor_params = {
+                "secret_kube_pod": spec_yaml,
+                "secret_kube_config": kubeconfig_yaml
```
AlanCoding, line 3110: there are some cases we are seeing:

```
  File "/home/ec2-user/tower-qa/tests/api/external_jobs/test_container_groups.py", line 313, in test_container_group_sliced_jobs_exceeding_resource_quota
    job.assert_status(['running', 'pending', 'waiting'])
  File "/home/ec2-user/venvs/venv/lib64/python3.6/site-packages/awxkit/api/mixins/has_status.py", line 93, in assert_status
    raise AssertionError(msg)
AssertionError: Job-3243 has status of error, which is not in ['running', 'pending', 'waiting'].
result_traceback:
  File "/var/lib/awx/venv/awx/lib64/python3.6/site-packages/awx/main/tasks.py", line 1425, in run
    res = receptor_job.run()
  File "/var/lib/awx/venv/awx/lib64/python3.6/site-packages/awx/main/tasks.py", line 3036, in run
    params=self.receptor_params)
  File "/var/lib/awx/venv/awx/lib64/python3.6/site-packages/receptorctl/socket_interface.py", line 187, in submit_work
    raise RuntimeError(errmsg)
RuntimeError: Failed to start work unit: secret_kube_config provided but not allowed
```

Looks like 2 places? So not a lot, don't know what's up with that.

AlanCoding: raised from

https://github.com/project-receptor/receptor/blob/e293394235082df0b5f2cf037a43d13c6a1f3787/receptorctl/receptorctl/socket_interface.py#L187

awx/main/isolated/manager.py
```diff
@@ -45,28 +44,12 @@ def __init__(self, event_handler, canceled_callback=None, check_callback=None, p
...
-                inventory['all']['hosts'][host] = {
-                    "ansible_connection": "kubectl",
-                    "ansible_kubectl_config": path,
-                }
-        else:
-            inventory = '\n'.join([
-                '{} ansible_ssh_user={}'.format(host, settings.AWX_ISOLATED_USERNAME)
-                for host in hosts
-            ])
+        inventory = '\n'.join([
```
jbradberry, line 49: a minor complaint, but `.join(...)` doesn't need a list comprehension

AlanCoding (approved):
There are still 2 PRs open against this https://github.com/shanemcd/awx/pulls, but it's relatively easy to change the base.

There's a lot of work still going on, but with the latest round of issues and patches, I feel like the form of work is taking shape. This has been the primary branch we've worked off for a while, so we should go ahead and merge it into `execution-environments` whenever.

shanemcd:
recheck

jakemcdermott:
doit

jakemcdermott (approved):

jakemcdermott (approved):

jakemcdermott (approved):

jakemcdermott (approved):

Issue #8501, opened 2020-10-30
Attempting to delete a workflow approval template results in 500 error


mabashian:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
<img width="1148" alt="Screen Shot 2020-10-30 at 11 16 35 AM" src="https://user-images.githubusercontent.com/9889020/97722929-94985480-1aa1-11eb-9f42-21eac7c2003f.png">

See the bottom of the issue for full traceback.

##### STEPS TO REPRODUCE
1. Create a workflow
2. Create a workflow approval node
3. Save workflow
4. Attempt to delete the created workflow approval template via the API browser

##### EXPECTED RESULTS
Successful delete (?)

##### ACTUAL RESULTS
Error

##### ADDITIONAL INFORMATION
Full error:

```
awx_1       | 2020-10-30 15:16:33,114 ERROR    django.request Internal Server Error: /api/v2/workflow_approval_templates/16/
awx_1       | Traceback (most recent call last):
awx_1       |   File "/venv/awx/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
awx_1       |     response = get_response(request)
awx_1       |   File "/venv/awx/lib/python3.6/site-packages/django/core/handlers/base.py", line 115, in _get_response
awx_1       |     response = self.process_exception_by_middleware(e, request)
awx_1       |   File "/venv/awx/lib/python3.6/site-packages/django/core/handlers/base.py", line 113, in _get_response
awx_1       |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
awx_1       |   File "/usr/lib64/python3.6/contextlib.py", line 52, in inner
awx_1       |     return func(*args, **kwds)
awx_1       |   File "/venv/awx/lib/python3.6/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
awx_1       |     return view_func(*args, **kwargs)
awx_1       |   File "/venv/awx/lib/python3.6/site-packages/django/views/generic/base.py", line 71, in view
awx_1       |     return self.dispatch(request, *args, **kwargs)
awx_1       |   File "./awx/api/generics.py", line 328, in dispatch
awx_1       |     return super(APIView, self).dispatch(request, *args, **kwargs)
awx_1       |   File "/venv/awx/lib/python3.6/site-packages/rest_framework/views.py", line 505, in dispatch
awx_1       |     response = self.handle_exception(exc)
[...]

PR #8612, opened 2020-11-17
Allow deletion of workflow approval template

fosterseth:
##### SUMMARY

related #8501 

Create a workflow with approval node. Attempting to delete the workflow approval template results in a 500 server error.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 15.0.1
```

### DIFF
--- a/awx/main/models/workflow.py
+++ b/awx/main/models/workflow.py
@@ -674,7 +674,7 @@ class WorkflowJob(UnifiedJob, WorkflowJobOptions, SurveyJobMixin, JobNotificatio
         return self.status == 'running'
 
 
-class WorkflowApprovalTemplate(UnifiedJobTemplate):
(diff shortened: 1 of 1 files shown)


rooftopcellist:
I just tested this out, this indeed fixes the 500 and allows for deletion of the WF approval template.  

I created a WF approval template:
image


Then successfully deleted it, but instead of recieving a 200, the response has a 204 in it.  

image

AlanCoding (approved):

rooftopcellist:
I tested this, and am now able to delete a WF approval template without getting a 500 using the `/api/v2/workflow_approval_templates` endpoint.

rooftopcellist (approved):

PR #8614, opened 2020-11-17
analytics: include modified column and full event_data for playbook_on_stats

jctanner:
##### SUMMARY
Fixes https://github.com/RedHatInsights/tower-analytics-backend/issues/514

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME
analytics

##### AWX VERSION

```
awx: 15.0.1
```

### DIFF
--- a/awx/main/analytics/collectors.py
+++ b/awx/main/analytics/collectors.py
@@ -280,7 +280,7 @@ def _copy_table(table, query, path):
     return file.file_list()
 
 
-@register('events_table', '1.1', format='csv', description=_('Automation task records'), expensive=True)
+@register('events_table', '1.2', format='csv', description=_('Automation task records'), expensive=True)
 def events_table(since, full_path, until, **kwargs):
     events_query = '''COPY (SELECT main_jobevent.id, 
                               main_jobevent.created,
--- a/awx/main/analytics/collectors.py
+++ b/awx/main/analytics/collectors.py
@@ -288,6 +288,7 @@ def events_table(since, full_path, until, **kwargs):
                               main_jobevent.parent_uuid,
                               main_jobevent.event, 
                               main_jobevent.event_data::json->'task_action' AS task_action,
+                              (CASE WHEN event = 'playbook_on_stats' THEN event_data END),
                               main_jobevent.failed, 
                               main_jobevent.changed, 
                               main_jobevent.playbook, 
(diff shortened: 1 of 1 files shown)


awx/main/analytics/collectors.py
```diff
@@ -309,6 +309,25 @@ def events_table(since, full_path, until, **kwargs):
     return _copy_table(table='events', query=events_query, path=full_path)
 
 
+@register('events_table_playbook_stats', '1.1', format='csv', description=_('playbook on stats records'), expensive=True)
+def events_table_playbook_stats(since, full_path, until, **kwargs):
+    events_query = '''COPY (SELECT main_jobevent.id,
```
ryanpetrello: @jctanner isn't this data already captured in https://github.com/ansible/awx/pull/8614/files#diff-5e7230fef85166a18967e2216535cd734d66de8b26983c303981e3c82b6fa1afR284 ?

jctanner: @ryanpetrello the difference is that this gets the full event payload for playbook_on_stats events.

ryanpetrello: I see.  Is there anything sensitive in this payload?  Is there some way we could combine both queries into one, but only conditionally return event data if the event actually *is* `playbook_on_stats`.  I ask because at large scale, this is a *very* expensive/slow query (to effectively run twice).

jctanner: It will contain all inventory hostnames. If there's some sort of conditional magic in sql to add a field when the value of another field is a certain string, the the other collector could be updated. However, that is beyond my level of ability.

ryanpetrello: Something like this might work?

```sql
SELECT event, (CASE WHEN event = 'playbook_on_stats' THEN event_data END) FROM main_jobevent;
```

Ladas: @jctanner @ryanpetrello yeah, I think getting it in 1 query like this will be more effective on tower side (we'll do only 1 index scan of events)

jctanner: I will try to modify the other collector with that SQL example and revise the PR.

jctanner: @ryanpetrello it seems to work, ty!

wenottingham:
@Ladas does this need a version update on events_table with this change?

jctanner:
@wenottingham are you referring to the second arg in the `@register` decorator?

wenottingham:
Yes, the idea is we bump that version whenever we change the format/schema of what we're sending (for the processor to act on). This is a corner case as to whether we're really changing the schema or not - do we need it here?

Ladas:
@wenottingham @jctanner yeah, it adds a new column to the csv, so we should bump the version

jctanner:
With 10K hosts and a single task, we exceed the default csv field size limits ...

```
[root@localhost extract]# python -c "import csv; from pprint import pprint; reader = csv.reader(open('events_table.csv
')); data = list(reader); pprint([[x[0],x[4],x[5],x[6]] for x in data][-10:])"
  File "<string>", line 1, in <module>
_csv.Error: field larger than field limit (131072)
```

If we use csv reader, we'll need to set `csv.field_size_limt` per https://stackoverflow.com/a/15063941

jctanner:
The AA db schema will need an update ...

```
postgres                    | 2020-11-18 23:13:31.564 UTC [1361] ERROR:  column "playbook_on_stats" of relation "job_events_tmp" does not exist
```

jctanner:
The AA backend PR has been merged, so this is ready for review/revision/merge/etc.

ryanpetrello (approved):

Ladas (approved):
Looks good :+1:

Ladas:
@jctanner @ryanpetrello can we backport up to 3.7? (cc @wenottingham )

ryanpetrello:
This patch is simple, I'm fine w/ this in 3.7.5 and 3.8.1.

PR #8615, opened 2020-11-17
Fix the navbar

jakemcdermott:
##### SUMMARY
The expected behavior for the navbar is to start expanded.


jakemcdermott:
@unlikelyzero @AlexSCorey A test should have failed somewhere when we broke this. Can you make sure a test gets added somewhere for this behavior?

unlikelyzero (approved):

PR #8616, opened 2020-11-18
add postgresql volume capacity for openshift persistent volume claim …

ljuaneda:
##### SUMMARY
Optionnal PersistentVolumeClaim creation for the postgreSQL database
Optionnal Docker Image name spec for PostgreSQL to handle properly disconnected installations

### DIFF
--- a/installer/roles/kubernetes/defaults/main.yml
+++ b/installer/roles/kubernetes/defaults/main.yml
@@ -38,6 +38,7 @@ kubernetes_redis_config_mount_path: "/usr/local/etc/redis/redis.conf"
 
 openshift_pg_emptydir: false
 openshift_pg_pvc_name: postgresql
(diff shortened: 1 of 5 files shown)


ljuaneda:
I've also added the ability to specify your own route host name in Openshift.
It's useful if you're using openshift router sharding

ryanpetrello:
@ljuaneda sorry for the delay on reviewing this PR - we've in the process of replacing the k8s installer in `devel` (https://groups.google.com/g/awx-project/c/47MjWSUQaOc/m/bCjSDn0eBQAJ https://github.com/ansible/awx/tree/devel/tools/docker-compose) and thus the next major release of AWX with an operator (https://github.com/ansible/awx-operator), so I suspect this PR is obsolete now.

PR #8617, opened 2020-11-18
Avoid invalid group characters in default inventory script factory

AlanCoding:
##### SUMMARY
This avoids the warning:

> [WARNING]: Invalid characters were found in group names but not replaced, use

When this script is used.

Pretty soon, we need to kill off this script, but we also need to silence this warning to clean up some tests right now.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API


jladdjr (approved):

jladdjr:
@AlanCoding - I think we can merge this one if you're ready

AlanCoding:
We only have 1 test change needed where it hard-coded "group-" and we just change that to "group_"

I'll see about getting that change in and merging this.

Issue #4204, opened 2018-12-03
Add support for social auth on login


mabashian:
AWX currently supports social auth options such as Google and Github (there may be others).  You can see the documentation for that [here](https://docs.ansible.com/ansible-tower/latest/html/administration/social_auth.html).  Note that when these social auth options are enabled/configured a logo appears on the login modal.  Clicking that logo should redirect the user to the platform in question where the user will authenticate.  Upon authentication, a callback url will be hit and the user will be logged in.

@trahman73 afaik the patternfly react login component doesn't have support for these social auth buttons.  If you have ideas on where they should go/how they should function i'd be interested in hearing about that.

trahman73:
@mabashian The React login component doesn't currently support social auth logins, but the PF4 core module does.

https://pf-next.com/components/login/examples-full/?component=Login

We can request the PF4 React Login dev to include support for social auths or we can build our own following the design from the above link. The latter might be more difficult as the login is a single page component.

nixocio:
Social auth is listed here now - https://www.patternfly.org/v4/components/login-page/react/#props

unlikelyzero:
covered in 5783

PR #8619, opened 2020-11-18
Convert Login.jsx to functional component

mabashian:
##### SUMMARY
Prep work for #4204 

This should behave the same way that the login used to behave with a couple of exceptions:

1) The Log In button will now be disabled while the authentication request is pending
2) An error modal is shown when we fail to fetch the custom login info

I'm open to removing that modal if folks feel strongly about it.  I just thought it might be nice to show the error in case it could be fixed by the user.  On the other hand, if there is an error on that endpoint the modal would get shown each time they hit the login page which could be quite annoying.

### DIFF
--- a/awx/ui_next/src/App.test.jsx
+++ b/awx/ui_next/src/App.test.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-
+import { act } from 'react-dom/test-utils';
 import { mountWithContexts } from '../testUtils/enzymeHelpers';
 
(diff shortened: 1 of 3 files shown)


jakemcdermott (approved):

unlikelyzero:
@mabashian if the visual tests for login pass, this is good to merge

unlikelyzero:
Green and covered in visualintegration/sanity login

Issue #8629, opened 2020-11-18
Observability and metrics for AWX Job execution pipeline


chrismeyersfsu:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
* How many jobs per minute<or some time unit> is my AWX capable of running?
To answer this question you must consider the entire job running pipeline from the API, where jobs are launched or created, to the Task Manager, where jobs are put into waiting, through the Dispatcher, where jobs are put into running & finished, next is the callback receiver where job events from running jobs are saved to the database, websockets where job events are sent to clients in the browser, and even notifications, which run after everything is complete. The runtime of each of these subsystems is needed to answer the seemingly simple question of "How many jobs per second is my AWX capable of running?"

### Define Metrics
Concretely, the first step to answering the above question is to first understand the performance of each subsystem. To do so, we must be able to observe those subsystem. We do so through metrics. 

To start with we should define the sub-systems in the pipeline. Then, discover what metrics tell the story. Defining the metrics that are valuable is the hard part. My thought is that each subsystem is like a queue so information that you might want to know about a queue would be useful i.e. depth, push vs. pop rate, average per-item duration in the queue, max latency of item in the queue. We have successfully defined the metrics that provide value for the callback receiver.
 
| Sub-System        | Metrics                                                      |
| ----------------- | ------------------------------------------------------------ |
| API               | job launches per time unit                                   |
| Task Manager      | average time to first process a pending job, number of jobs transitioned from pending -> waiting per time unit (not sure about these) |
| Dispatcher        | latency from waiting -> running start, average time spent pre-start work, average time spent post-run work |
| Callback Receiver | Events per time processed (existing stats)                   |
| Websockets        | Similar to the callback receiver + number of messages dropped when overloaded                 |
| Notifications     | Maybe this rolls into the dispatcher as post-run work        |

To be clear, the above metrics are mostly guesses. We do know what metrics are useful for the callback receiver  & some of the websockets. I've explored tracing individual job timings as they flow through the task manager and below here https://github.com/ansible/awx/pull/8631

### Work Items
* Save metrics in Redis https://github.com/ansible/awx/issues/9012
* Expose metrics in the API as Prometheus data https://github.com/ansible/awx/issues/9019
* Broadcast metrics to every AWX node https://github.com/ansible/awx/issues/9020
* metrics dashboard https://github.com/ansible/awx/issues/9056
* periodically output metrics to log file https://github.com/ansible/awx/issues/9021
[...]

PR #8630, opened 2020-11-18
WIP cbreceiver -> statsd -> collectd

chrismeyersfsu:
* /var/log/tower/collectd.log can be consumed by the sosreport and
pushed into an elk stack

related to https://github.com/ansible/awx/issues/8629


chrismeyersfsu:
We are going to use Redis instead of a new service to store/share stats data.

PR #8631, opened 2020-11-18
WIP per-job time spent in pipeline step

chrismeyersfsu:
Output timing data to rsyslog w/ an elk stack on the other side. Looking at tracing job execution time through our pipeline.


ryanpetrello:
@fosterseth does your work replace this cc @chrismeyersfsu or is this another distinct thing?

PR #8640, opened 2020-11-19
Adds filter by role on Org access lists

### LINKED ISSUE #8130: Add an interface for displaying org administrators
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
In the old UI we have this page:

Screen Shot 2020-09-11 at 11 20 25 AM

that shows all the users who have an admin role on the Org in question.

I attempted to gather this same information via the advanced search on the access list but was unable to do so.  That might be because I don't understand how `roles__search` works 🤷‍♂️ .  Maybe someone on the API team can shed some light on this.

We either need a definitive way to filter the access list down to just users who have the Admin role _or_ we need to add this tab to the Org details and have it pull from `/api/v2/organizations/:id/admins`.  

Ideally, I think we'd rather have the filtering working if we had to choose between one or the other.

AlexSCorey:
##### SUMMARY
This will solve #8130 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
roles search

### DIFF
--- a/awx/ui_next/src/components/ResourceAccessList/ResourceAccessList.jsx
+++ b/awx/ui_next/src/components/ResourceAccessList/ResourceAccessList.jsx
@@ -2,7 +2,7 @@ import React, { useCallback, useEffect, useState } from 'react';
 import { useLocation } from 'react-router-dom';
 import { withI18n } from '@lingui/react';
 import { t } from '@lingui/macro';
-import { TeamsAPI, UsersAPI } from '../../api';
+import { RolesAPI, TeamsAPI, UsersAPI } from '../../api';
 import AddResourceRole from '../AddRole/AddResourceRole';
 import AlertModal from '../AlertModal';
 import DataListToolbar from '../DataListToolbar';
@@ -26,7 +26,13 @@ function ResourceAccessList({ i18n, apiModel, resource }) {
   const location = useLocation();
 
   const {
-    result: { accessRecords, itemCount, relatedSearchableKeys, searchableKeys },
+    result: {
+      accessRecords,
+      itemCount,
+      relatedSearchableKeys,
+      searchableKeys,
+      organizationRoles,
+    },
     error: contentError,
     isLoading,
     request: fetchAccessRecords,
@@ -37,6 +43,41 @@ function ResourceAccessList({ i18n, apiModel, resource }) {
         apiModel.readAccessList(resource.id, params),
         apiModel.readAccessOptions(resource.id),
       ]);
+
+      // Eventually this could be expanded to other access lists.
+      // We will need to combine the role ids of all the different level
+      // of resource level roles.
+
+      let orgRoles;
+      if (location.pathname.includes('/organizations')) {
+        const {
+          data: { results: roles },
+        } = await RolesAPI.read({ content_type__isnull: true });
+        const sysAdmin = roles.filter(
+          role => role.name === 'System Administrator'
+        );
+        const sysAud = roles.filter(role => {
+          let auditor;
+          if (role.name === 'System Auditor') {
+            auditor = role.id;
+          }
+          return auditor;
+        });
+
+        orgRoles = Object.values(resource.summary_fields.object_roles).map(
+          opt => {
+            let item;
+            if (opt.name === 'Admin') {
+              item = [`${opt.id}, ${sysAdmin[0].id}`, opt.name];
+            } else if (sysAud[0].id && opt.name === 'Auditor') {
+              item = [`${sysAud[0].id}, ${opt.id}`, opt.name];
+            } else {
+              item = [`${opt.id}`, opt.name];
+            }
+            return item;
+          }
+        );
+      }
       return {
         accessRecords: response.data.results,
         itemCount: response.data.count,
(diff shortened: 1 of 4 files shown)


awx/ui_next/src/components/ResourceAccessList/ResourceAccessList.jsx
```diff
@@ -25,8 +25,51 @@ function ResourceAccessList({ i18n, apiModel, resource }) {
...
+      if (id.indexOf(details.type) > -1) {
+        OrgResourceAdminId = value.id;
+      }
+    });
+
+    const options = Object.entries(details.summary_fields.object_roles).map(
+      opt => {
+        const orgAdmin = org?.summary_fields?.object_roles?.admin_role.id;
+        const orgExecute = org?.summary_fields?.object_roles?.execute_role.id;
+        if (opt[1].name.includes('Admin')) {
```
mabashian: Matching on strings that come back from the API always makes me a little nervous.  This mostly stems from the fact that strings _could_ be translated.  I'm not sure if these particular strings get translated or not (or whether they could be in the future).  I'm wondering if we could do something like this instead:

```
        if (opt[0] === 'admin_role') {
          return [
            [opt[1].id, orgAdmin, OrgResourceAdminId].join(','),
            opt[1].name,
          ];
        }
        if (opt[0] === 'execute_role') {
          return [
            [opt[1].id, orgAdmin, orgExecute, OrgResourceAdminId].join(','),
            opt[1].name,
          ];
        }
```

I'm having some trouble fully wrapping my head on what's going on here.  I may need you to give me the rundown so that I understand how we're generating these options for the dropdown.

mabashian: OK, I see the data structure now.  We want to build a concatenated list of role id's to associate with the "Admin" or "Execute" options like this:

```
["108,2,9", "Admin"]
```

mabashian:
One thing I did notice was I'm not sure we're capturing team level roles in our filters:

Screen Shot 2020-11-20 at 10 08 21 AM

Screen Shot 2020-11-20 at 10 08 32 AM

This also got me thinking about higher level roles like System Administrator and System Auditor.  Should they be included in each of these buckets?  Right now it looks like they get filtered out.

I really like this feature and I think it will be a huge improvement over the old UI - I'm just wondering if we need to take it a bit further.

unlikelyzero:
@AlexSCorey realistically, we may not be able to get this merged in this week

AlexSCorey:
@mabashian I had to pare this PR down as the api does not have functionality to filter the users down if they have been granted access via a team.

unlikelyzero:
@AlexSCorey do you have a ticket field for the functionality that's missing and required for the API?

AlexSCorey:
@unlikelyzero https://github.com/ansible/awx/issues/8708

unlikelyzero:
@mabashian ```This also got me thinking about higher level roles like System Administrator and System Auditor. Should they be included in each of these buckets? Right now it looks like they get filtered out.```

I think we should consider leaving them out. There was some confusion in the bug bash once there were more than 2 system admins. https://github.com/ansible/awx/issues/8818

unlikelyzero:
@AlexSCorey can you rebase now that tables have landed?

AlexSCorey:
@unlikelyzero updated

unlikelyzero:
@AlexSCorey system administrators and system auditors do not appear when I select all roles under 'Filter by Roles'. I think we should add those two role types

jakemcdermott:
OK this is shaping up nicely. Nice work, @AlexSCorey !

My understanding is that there's a few more things we want to add to this PR. We're going to connect w/ @jbradberry later in the week just to see about what the best way to get the sysadmin and auditor role information

jakemcdermott (approved):
:+1:

akus062381:
@AlexSCorey found this bug while trying to test. Putting it back into needs_devel status.

Screen+Recording+2021-02-10+at+04 50 32+PM

unlikelyzero:
@AlexSCorey almost done. I've got a PR up 5978

There's a bug with Team's Roles not appearing that needs to be fixed
Screen Recording 2021-02-13 at 10 39 52 AM

I'm attempting to write a test to cover the removal of User and Team roles from the access list by I do not have a clear way to locate the X "close" component. Would it be possible to make those unique with a deterministic ID?

AlexSCorey:
recheck

awx/ui_next/src/components/ResourceAccessList/ResourceAccessListItem.jsx
```diff
@@ -56,6 +56,8 @@ function ResourceAccessListItem({ accessRecord, onRoleDelete, i18n }) {
           onRoleDelete(role, accessRecord);
         }}
         isReadOnly={!role.user_capabilities.unattach}
+        ouiaId={`${role.name}-${role.id}`}
```
unlikelyzero, line 59: here

unlikelyzero:
Covered in 5978
Adding https://github.com/ansible/awx/issues/8708 to the next release milestone

unlikelyzero (approved):

PR #8650, opened 2020-11-20
Add support for Insights as an inventory source

wenottingham:
##### SUMMARY

Use the insights inventory plugin as an inventory source.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION

Requires at minimum https://github.com/RedHatInsights/ansible-collections-insights/pull/14


requirements/collections_requirements.yml
```diff
@@ -8,5 +8,6 @@ collections:
   - name: openstack.cloud
   - name: community.vmware
   - name: ovirt.ovirt
+  - name: redhat.insights
```
wenottingham: Herein lies the issue - the build process cannot handle this automatically right now.

AlanCoding: Could we publish a _different_ execution environment for this source?

In the bigger picture, I'd like to move away from this single default EE which is not a combination of very unrelated content.

Ideally, we should discover what collections are in an execution environment, and when creating an inventory source, give a listing of execution environments which could plausibly run it. The FQCN `redhat.insights.insights`, which we put into the code itself, gives the collection right there.

wenottingham: Eh, I think if we start down the road of one-EE-per-type, and inspecting them, it's a much larger and fundamental change.

AlanCoding: Setting aside how distant-future it is, or even how much effort it takes, I think one-EE-per-type would be the wrong answer. Some collections, after all, are dependencies of other collections and it wouldn't make sense to cut an image for both. So my idea is 1 EE to several inventory sources (starting whenever a concrete need arises). How many, and how to organize them, I don't know.

wenottingham: Yes, mostly just thinking the organizational/model/etc work is all in going from 1 to >1, and the specific number >1 doesn't really matter.

awx/main/models/inventory.py
```diff
@@ -1582,5 +1583,19 @@ class tower(PluginFileInjector):
...
+    plugin_name = 'insights'
+    namespace = 'redhat'
+    collection = 'insights'
+
+    def get_plugin_env(self, inventory_update, private_data_dir, private_data_files):
+        credential = inventory_update.get_cloud_credential()
+        ret = super(insights, self).get_plugin_env(inventory_update, private_data_dir, private_data_files)
+        if credential:
+            ret['INSIGHTS_USER'] = credential.get_input('username', default='')
+            ret['INSIGHTS_PASSWORD'] = credential.get_input('password', default='')
```
AlanCoding: We don't need this and it not DRY. See other examples like rhv:

https://github.com/ansible/awx/blob/698844db4736421ef5dff7b684a4d0870cecd9b3/awx/main/models/inventory.py#L1556

That will tell it to fill in the env vars from the credential type definition at:

https://github.com/ansible/awx/blob/698844db4736421ef5dff7b684a4d0870cecd9b3/awx/main/models/credential/__init__.py#L1070

For your insights credential type, you've already put `INSIGHTS_USER` in the `injector`. So add `base_injector = 'template' ` and then remove the entire `get_plugin_env` method here.

If people dislike this strongly, then we should get rid of `base_injector` and _always_ use the managed/template injector logic if it's present for the credential type. Otherwise we repeat ourselves.

AlanCoding:
From what I can tell, the build process for the collection templates the namespace:

https://github.com/RedHatInsights/ansible-collections-insights/blob/fa3e8a37ecc52c29fc4cec8c0f3a7727962aca6d/.github/workflows/galaxy-release.yml#L27

This means that there's an upstream version at https://galaxy.ansible.com/redhatinsights/insights

We should resolve this current item using the upstream collection. We have no construct of supporting downstream collections. Make it work with the upstream and worry about that someday in the future.

wenottingham:
Yea, but the templated build process for the upstream collection is broken. (The plugin doesn't work)

awx/main/models/inventory.py
wenottingham: Updated.

AlanCoding:
> Yea, but the templated build process for the upstream collection is broken. (The plugin doesn't work)

I want to narrow down what this is saying. You mean it doesn't work because:

https://github.com/RedHatInsights/ansible-collections-insights/blob/fa3e8a37ecc52c29fc4cec8c0f3a7727962aca6d/plugins/inventory/insights.py#L91

that hard-codes `NAME = 'redhat.insights.insights'`.

That should be fixed on the side of the collection. It was fixed for the `awx.awx`/`ansible.tower` collection. You can go suggest this to the collection maintainers.

https://github.com/ansible/awx/blob/b5c2a6ad6568cfeb55fae8bf3058c7eb53088b3d/awx_collection/tools/roles/template_galaxy/tasks/main.yml#L31-L35

They are already template things in a playbook, so nothing is fundamentally different.

----

So then assuming that gets fixed, inventory updates still won't work, because in this PR `redhat.insights.insights` is hard-coded in the AWX source. My suggestion is that it should be changed to `redhatinsights.insights.insights` so that it's the upstream collection.

After that, the remaining problem is making it work with downstream. Working with the downstream collection is already a problem with many other collections in our manifest, and we don't yet have a specific outline of how that will work.

wenottingham:
> that hard-codes NAME = 'redhat.insights.insights'.

Yes, that. I'm not clear why they did their upstream/downstream split this way.

AlanCoding:
Scratch that, they did correctly do the replacement of that line in code for the officially released version. The inventory plugin should work fine as `redhatinsights.insights.insights`.

Circling back, my suggestion is to add the upstream collection and get it working, then plan the split for downstream on its own.

ryanpetrello:
@wenottingham what's the plan for following up on this post-EE merge.  Is this work tracked in a ticket/issue somewhere?

wenottingham:
Rebased.

It will be in certain execution environments out of the box, so it could just work.

shanemcd:
@wenottingham Is this ready for review?

wenottingham:
Waiting on upstream plugin fix to enable using insights remediations as-is, but this code should be reviewable in its current state.

wenottingham:
recheck

AlanCoding:
Trying to take stock again, I believe this requires https://github.com/ansible/awx-ee/pull/23 to work, is that right?

wenottingham:
Correct. I'm mystified by the remaining unit test failure - it reads as if all sources should fail that way, not this one. (The key for the file is `inventory/insights.yml`, due to some pathing changes).

AlanCoding:
@wenottingham delete `awx/main/tests/data/inventory/plugins/insights/files/insights.yml` and it passes.

The intent of the "file_reference" is when the inventory update requires a _separate_ credential file (not for the inventory file itself). If I view the `env.json` I can clearly see that no such file is needed to setup insights credentials here.

wenottingham:
... that then leaves the question of why data/inventory/plugins/tower/files/tower.yml is there, as it is equivalent.

AlanCoding:
@wenottingham Update on https://github.com/ansible/awx-ee/pull/23 - we merged it and here I am confirming that the image built and pushed.

```
$ podman run --rm --tty --interactive quay.io/ansible/awx-ee:devel ansible-galaxy collection list
[WARNING]: You are running the development version of Ansible. You should only run Ansible from
"devel" if you are modifying the Ansible engine, or trying out features under development. This is a
rapidly changing source of code and can become unstable at any point.

# /usr/share/ansible/collections/ansible_collections
Collection              Version
----------------------- -------
amazon.aws              1.5.0  
ansible.posix           1.2.0  
ansible.windows         1.6.0  
awx.awx                 19.2.0 
azure.azcollection      1.6.0  
community.vmware        1.10.0 
google.cloud            1.0.2  
kubernetes.core         1.2.1  
openstack.cloud         1.4.0  
ovirt.ovirt             1.5.0  
redhatinsights.insights 1.0.4  
theforeman.foreman      2.1.0
```

awx/main/models/inventory.py
```diff
@@ -1558,5 +1559,21 @@ class tower(PluginFileInjector):
     collection = 'awx'
 
 
+class insights(PluginFileInjector):
+    plugin_name = 'insights'
+    base_injector = 'template'
+    namespace = 'redhatinsights'
+    collection = 'insights'
+    downstream_namespace = 'redhat'
+    downstream_collection = 'insights'
```
AlanCoding, line 1551: :open_mouth: 

:hugs: 

:+1: 

link https://github.com/ansible/awx/pull/10382

AlanCoding (approved):
The inventory injector stuff w/ insights collection should work and follows all the standard patterns. That part has my blessing.

I've asked some help for reviewing host insights endpoint removal, as I may not have the background for it, and I'll be out tomorrow

chrismeyersfsu:
Looks like we probably don't need to figure out if hosts are insights hosts anymore. I'm going to submit a PR to this PR to remove tower code that consumes https://github.com/ansible/awx-facts-playbooks/blob/master/library/scan_insights.py

chrismeyersfsu (approved):

wenottingham:
> Looks like we probably don't need to figure out if hosts are insights hosts anymore. I'm going to submit a PR to this PR to remove tower code that consumes https://github.com/ansible/awx-facts-playbooks/blob/master/library/scan_insights.py

That is most of the goal of this patchset, yes - want to greatly simplify the process of consuming insights remediations.

wenottingham:
regate

PR #8651, opened 2020-11-20
Clean up some old sourve var settings in unit tests

wenottingham:
Also if you want to use assert, use assert.

Noticed in passing while adding another inventory source.


AlanCoding (approved):
looks great, thanks!

Issue #8530, opened 2020-11-03
Incorrect element ID for instance groups layout


unlikelyzero:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
The Instance Group Page has an underscore in the element ID. `#instance_group-list-toolbar` should be `#instance-group-list-toolbar`
##### ENVIRONMENT
* AWX version: 114bcd0349d571a082dc7ca92cb68980cdcec214

mabashian:
I'm assuming the id should be `instance-group-list-toolbar`.  The two strings listed initially look exactly the same to me.

unlikelyzero:
@mabashian yes exactly

one-t:
Hey, @mabashian - can you rebase this to pick up the node version update? Thanks!

PR #8654, opened 2020-11-23
Swap underscores for hyphens in qs namespace strings

mabashian:
##### SUMMARY
link #8530 

We currently have both patterns (underscores and hyphens) being used in these strings.  While I feel like the underscores make sense for display in the url, they don't look great in our element id's.

Since this string is used in both of those places it's going to look funny somewhere.  At least this way they're all consistent.

Example of hyphen already being used: https://github.com/ansible/awx/blob/devel/awx/ui_next/src/screens/NotificationTemplate/NotificationTemplateList/NotificationTemplateList.jsx#L19

### DIFF
--- a/awx/ui_next/src/components/Lookup/InstanceGroupsLookup.jsx
+++ b/awx/ui_next/src/components/Lookup/InstanceGroupsLookup.jsx
@@ -13,7 +13,7 @@ import useRequest from '../../util/useRequest';
 import Lookup from './Lookup';
 import LookupErrorMessage from './shared/LookupErrorMessage';
 
-const QS_CONFIG = getQSConfig('instance_groups', {
+const QS_CONFIG = getQSConfig('instance-groups', {
   page: 1,
   page_size: 5,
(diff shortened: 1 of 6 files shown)


jakemcdermott (approved):

unlikelyzero:
This will need some slight tweaks to the e2e tests to match the IDs

Also will need to uncomment this test https://github.com/ansible/tower-qa/blob/49b467f00fc220d6e84c6b52a48d7062cb151bce/ui-tests/awx-pf-tests/cypress/integration/sanity.spec.js#L159

unlikelyzero:
@one-t to re-run yolok to green

one-t:
Hey, @mabashian - can we get another rebase on this one? just need to sync things up again. Thanks!

mabashian:
@one-t just rebased - sorry for the delay!

one-t (approved):

one-t:
Tests: https://github.com/ansible/tower-qa/pull/5796

Issue #8606, opened 2020-11-17
Credential field missing from adhoc command output details


mabashian:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Here's the output details:

<img width="1680" alt="Screen Shot 2020-11-17 at 12 46 34 PM" src="https://user-images.githubusercontent.com/9889020/99427155-0d394680-28d3-11eb-83e7-797b0dfb26a4.png">

Here's the object:

<img width="674" alt="Screen Shot 2020-11-17 at 12 46 48 PM" src="https://user-images.githubusercontent.com/9889020/99427179-14605480-28d3-11eb-839d-312cf29f8134.png">

##### STEPS TO REPRODUCE
1. Launch an adhoc command and select a machine credential
2. Observe output results

##### EXPECTED RESULTS
Selected credential to be listed

##### ACTUAL RESULTS
Selected credential is not listed

tiagodread:
Closing this issue, the PR was merged and test are covered in 5745

PR #8655, opened 2020-11-23
Display machine credential in job details when present

mabashian:
##### SUMMARY
link https://github.com/ansible/awx/issues/8606

A machine credential is required in order to run an adhoc command so we should make sure that credential is displayed on the output details page.

This field can/should only be present in the summary_fields for an adhoc command (based on my local testing) so I don't think there's any risk that we'd display double credentials.

Screen Shot 2020-11-23 at 11 45 19 AM

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI


jakemcdermott (approved):

tiagodread (approved):
Worked as expected, I cover this in 5745

Issue #5683, opened 2020-01-16
Move Created By and Modified By fields


marshmalien:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Move **Created By** and **Last Modified** to the very bottom of the Detail List, but before the "full card width" detail items like labels, instance groups, and variables. 

_Actual Result_
<img width="1272" alt="Screen Shot 2020-07-27 at 10 51 00 AM" src="https://user-images.githubusercontent.com/15881645/88558155-2e5eea00-cff9-11ea-962a-5eaa7b27c0a0.png">

_Expected Result_
<img width="1345" alt="Screen Shot 2020-07-27 at 11 02 24 AM" src="https://user-images.githubusercontent.com/15881645/88557952-e8098b00-cff8-11ea-96ba-1c2f20762dba.png">

##### ADDITIONAL INFORMATION
This applies to all Detail views, so whoever picks this up will have to audit ui_next for all occurrences.

unlikelyzero:
@marshmalien still an issue?

marshmalien:
@unlikelyzero Yes, just verified with @trahman73 and it is still an issue.

unlikelyzero:
covered in https://github.com/ansible/tower-qa/pull/5829

PR #8658, opened 2020-11-23
Move jt/wfjt created/modified details to the end right before the full width details

mabashian:
##### SUMMARY
link #5683 

Screen Shot 2020-11-23 at 2 37 42 PM
Screen Shot 2020-11-23 at 2 20 57 PM

I also moved the Options detail in WorkflowJobTemplateDetail.jsx to come after the webhook details.  This makes it consistent with JobTemplateDetail.jsx.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

@unlikelyzero @tiagodread I assume this will cause a break in the visual tests

### DIFF
--- a/awx/ui_next/src/screens/Application/ApplicationDetails/ApplicationDetails.jsx
+++ b/awx/ui_next/src/screens/Application/ApplicationDetails/ApplicationDetails.jsx
@@ -7,7 +7,11 @@ import { Button } from '@patternfly/react-core';
 import useRequest, { useDismissableError } from '../../../util/useRequest';
 import AlertModal from '../../../components/AlertModal';
 import { CardBody, CardActionsRow } from '../../../components/Card';
-import { Detail, DetailList } from '../../../components/DetailList';
+import {
+  Detail,
+  DetailList,
+  UserDateDetail,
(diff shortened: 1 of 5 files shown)


jakemcdermott (approved):

unlikelyzero:
Since this impacts all details views, we'll need to ensure that the e2e tests are green and the visual tests pass.

unlikelyzero:
This will just need some visual test changes. If we don't have visual tests for this view, let's add them now

tiagodread:
@mabashian we are not displaying `Created` and `Last Modified` data for some details page, I went through all details page checking the API response, and here is the list:

- Jobs: has data but not displayed
- notification templates: has data but not displayed
- applications: has data but not displayed

mabashian:
@tiagodread 

Screen Shot 2021-01-04 at 3 20 25 PM
Screen Shot 2021-01-04 at 3 16 19 PM
Screen Shot 2021-01-04 at 3 13 17 PM

Including the user in the detail depends on whether its available via the api.  In some cases it is not.

tiagodread (approved):
Added visual tests in 5829

Issue #7522, opened 2020-07-01
JT survey edit breadcrumb clicking goes to blank view


chrismeyersfsu:
##### STEPS TO REPRODUCE

* Create a JT with a survey
* Add a survey item
* Click to edit that survey item
* Note the breadcrumbs at the top, click "Edit Question"

##### EXPECTED RESULTS

Probably for it to not be clickable

##### ACTUAL RESULTS

Sent to a blank view. Also note the "Back to Templates" is blue and underlined when I am editing a survey question. The survey question should probably be the thing blue and underlined.

![image](https://user-images.githubusercontent.com/722880/86284135-45541d00-bbb0-11ea-99c5-175633ea8273.png)

![image](https://user-images.githubusercontent.com/722880/86284123-4127ff80-bbb0-11ea-8eda-88d8c6847422.png)

unlikelyzero:
During `needs_test` verification, please add testcase (or stub)

mabashian:
`Edit Question` should not be a link in this case

tiagodread:
Verified that:

The user can create and edit a JT survey
Can navigate to survey list using breadcumb links
Can enable survey
Can delete survey
Can launch a JT with survey enabled

Added coverage in 5819

PR #8667, opened 2020-11-24
Fix issue with broken survey question edit breadcrumb by altering the url scheme

mabashian:
##### SUMMARY
link #7522 

I couldn't come up with a different way to solve this problem.  We define a static breadcrumb object with exact url's whenever the survey component is mounted.  At that point in time, we have no idea what all of the survey question variables are and therefore cannot generate static breadcrumb strings for all the possibilities.

To get around this, I removed the question variable from the route and instead put it in query params.

I also added some logic to redirect the user back out to the survey list when the query param was missing or the variable was not present in the survey spec.

Screen Shot 2020-11-24 at 12 44 31 PM

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/screens/Template/Survey/SurveyQuestionEdit.jsx
+++ b/awx/ui_next/src/screens/Template/Survey/SurveyQuestionEdit.jsx
@@ -1,5 +1,10 @@
 import React, { useState } from 'react';
-import { useHistory, useRouteMatch } from 'react-router-dom';
+import {
+  Redirect,
+  useHistory,
+  useLocation,
+  useRouteMatch,
+} from 'react-router-dom';
 import ContentLoading from '../../../components/ContentLoading';
 import { CardBody } from '../../../components/Card';
 import SurveyQuestionForm from './SurveyQuestionForm';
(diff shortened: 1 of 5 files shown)


unlikelyzero:
Needs a test

awx/ui_next/src/screens/Template/Survey/SurveyQuestionEdit.jsx
```diff
@@ -8,12 +13,23 @@ export default function SurveyQuestionEdit({ survey, updateSurvey }) {
   const [formError, setFormError] = useState(null);
   const history = useHistory();
   const match = useRouteMatch();
+  const { search } = useLocation();
+  const queryParams = new URLSearchParams(search);
+  const questionVariable = queryParams.get('question_variable');
 
   if (!survey) {
```
AlexSCorey, line 20: I know this is not part of this PR, but to me, this should be handled in TemplateSurvey.jsx where we fetch the survey.

marshmalien (approved):
I think this is a fine solution and it solves the issue with minimal code changes. I think there is a way to keep the variable name in the URL `/templates/${template.type}/${template.id}/survey/${variable}/edit`, but it would require drilling down `template` and `setBreadcrumb` from Template.jsx => TemplateSurvey.jsx => SurveyQuestionEdit.jsx. 

Within SurveyQuestionEdit, we could call `setBreadcrumb(template, null, variable)` and update Templates.jsx `breadcrumbConfig` to accept three parameters: template, schedule, and variable. Having said that, implementing the above versus what you have in this PR is probably not worth the trouble.

My only suggestion is that we should trigger the survey form to re-render whenever the  variable param changes. This could be done by adding a `key` prop or `enableReinitialize` to Formik in SurveyQuestionForm.jsx.

unlikelyzero:
We need to add a real test here. I am certain that this is going to be impacted by later changes

tiagodread (approved):
Verified that:
- The user can create and edit a JT survey
- Can navigate to survey list using breadcumb links
- Can enable survey
- Can delete survey
- Can launch a JT with survey enabled
Added coverage in 5819

PR #8672, opened 2020-11-24
Allow combining multiple subscriptions if they are in a manifest.

wenottingham:
The rules:
- subs must be of same SKU type (all MCTxxxx, all SERxxxx, etc)
~- subs must be of the same product name (aside from the managed node count)~
- when combining, we take the earliest end date

### DIFF
--- a/awx/main/utils/licensing.py
+++ b/awx/main/utils/licensing.py
@@ -11,7 +11,7 @@ The Licenser class can do the following:
 
 import base64
 import configparser
-from datetime import datetime
+from datetime import datetime, timezone
 import collections
 import copy
 import io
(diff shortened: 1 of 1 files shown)


rooftopcellist (approved):

wenottingham:
recheck

awx/main/utils/licensing.py
```diff
@@ -131,21 +134,69 @@ def _unset_attrs(self):
...
+        def is_appropriate_manifest_sub(sub):
+            if sub['pool']['activeSubscription'] is False:
+                return False
+            now = datetime.now(timezone.utc)
+            if parse_date(sub['startDate']) > now:
+                return False
+            if parse_date(sub['endDate']) < now:
+                return False
+            products = sub['pool']['providedProducts']
+            if any(map(lambda product: product.get('productId', None) == "480", products)):
```
jakemcdermott:
```suggestion
            if any(product.get('productId') == '480' for product in products):
```

Issue #8605, opened 2020-11-17
Attempting to create custom credential throws JS error


mabashian:
##### ISSUE TYPE
 - Bug Report

##### ENVIRONMENT
* AWX version: X.Y.Z
* AWX install method: openshift, minishift, docker on linux, docker for mac, boot2docker
* Ansible version:  X.Y.Z
* Operating System:
* Web Browser:

##### STEPS TO REPRODUCE
1. Create a custom credential type with the following Input Configuration:

```
fields:
  - id: username
    type: string
    label: Username
  - id: password
    type: string
    label: Password
    secret: true
  - id: multi_choice
    type: string
    label: Multiple Choice
    choices:
      - a
      - b
      - c
  - id: bool
    type: boolean
    label: Bool
  - id: ssh_key
    type: string
    label: SSH Key
    format: ssh_private_key
  - id: ssh_key_test
    type: string
    label: SSH Key
    multiline: true
required:
  - username
  - password
```
2. Attempt to create a new credential by only filling in the username/password

##### EXPECTED RESULTS
New credential created

##### ACTUAL RESULTS
Error in the console:

```
TypeError: Cannot read property 'credential' of undefined
    at CredentialAdd.jsx:46
    at Array.forEach (<anonymous>)
    at CredentialAdd.jsx:44
    at useRequest.js:42
    at handleSubmit (CredentialAdd.jsx:115)
    at onSubmit (CredentialForm.jsx:215)
    at Formik.tsx:866
    at Formik.tsx:1214
    at Formik.tsx:773 "
    at FormSubmitError (https://localhost:3001/static/js/main.chunk.js:11768:3)
...
```

tiagodread:
PR merged, tests in place

PR #8683, opened 2020-11-25
Avoid prop reference error when saving cred

jakemcdermott:
##### SUMMARY
for #8605


unlikelyzero:
Will need a real test here

nixocio (approved):

Issue #8666, opened 2020-11-24
[ui_next] Token modal does not display when creating token/application


unlikelyzero:
### ISSUE TYPE
 - Bug Report

##### SUMMARY
Token modal does not display when creating token/application

##### ENVIRONMENT
* AWX version: a50034be3c9d8257e02a430ff5fc8bd6901b2b58

tiagodread:
PR already merged, tests covered in 5776

PR #8684, opened 2020-11-25
Show modals with app/token info after creation

mabashian:
##### SUMMARY
link #8666 

Certain fields are only exposed in the response when an Application or Token are created.  We display this data via a modal that gets shown right after the resource is created.

Application:

new_app

Token:

new_token

I decided to use http://patternfly-react.surge.sh/components/clipboard-copy for some of these fields as it seems likely that they would be copied out of this modal.  I went with the expand/collapse mode in case the screen is small and the strings are cut off.

There were also a couple of fields missing from the Application and Token details views so I added those.

I also found the first column in the token list to be confusing.  For personal access tokens (tokens without an app) we displayed a string with a link to the token details.  For application tokens we showed a label (Application) and then the Application name with a link to the token details.  My initial thought was that link would take me to the Application.

I prefer something like this:

Screen Shot 2020-11-25 at 12 40 19 PM

Where the first column always links the user to the token details and the second column links the user to the Application (where present).

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/screens/Application/ApplicationAdd/ApplicationAdd.jsx
+++ b/awx/ui_next/src/screens/Application/ApplicationAdd/ApplicationAdd.jsx
@@ -8,7 +8,7 @@ import ApplicationForm from '../shared/ApplicationForm';
 import { ApplicationsAPI } from '../../../api';
 import { CardBody } from '../../../components/Card';
 
-function ApplicationAdd() {
+function ApplicationAdd({ onSuccessfulAdd }) {
   const history = useHistory();
   const [submitError, setSubmitError] = useState(null);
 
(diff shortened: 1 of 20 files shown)


tiagodread:
recheck

tiagodread:
@mabashian should the user be able to edit the token? In the old UI you can only delete the token.
If the answer is **Yes**, then, the Edit button doesn’t work in the token details tab.

unlikelyzero:
This will need a real test (not a stub)

mabashian:
Note to self: we should add some text to the modals letting the user know that they won't be able to see the token again after they close the modal and that they should copy it now.

jakemcdermott (approved):

unlikelyzero:
@tiagodread @mabashian do we have adequate unit test coverage here?

Also, can we ensure that the extra text for copying before close is i18n'd in the unit tests?

tiagodread:
@mabashian can you add some uniqueness id for each  `<dd>` and `<dt>` element in the details view?

image

mabashian:
> Note to self: we should add some text to the modals letting the user know that they won't be able to see the token again after they close the modal and that they should copy it now.

Screen Shot 2020-12-08 at 1 52 14 PM
Screen Shot 2020-12-08 at 1 48 37 PM

mabashian:
@tiagodread @unlikelyzero I think i've addressed everything listed here

tiagodread (approved):
Awesome!
Worked as expected, added coverage in 5776 👍🏽

PR #8706, opened 2020-11-30
Pre-lingui upgrade 

AlexSCorey:
##### SUMMARY
In preparation for a Ling ui upgrade I am auditing our files that use withI18n looking for those files that still use React classes.  Where it makes sense to move those over to hook I am doing so. This is the first PR doing that.  This will be followed by at least 1 or 2 more PRs doing this.  No functionality has changed.

12/15/2020 ran e2e tests on this PR and everything passed
##### ISSUE TYPE
- Preparation PR

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/components/About/About.jsx
+++ b/awx/ui_next/src/components/About/About.jsx
@@ -12,8 +12,8 @@ import {
 import { BrandName } from '../../variables';
 import brandLogoImg from './brand-logo.svg';
 
-class About extends React.Component {
-  static createSpeechBubble(version) {
+function About({ ansible_version, version, isOpen, onClose, i18n }) {
+  const createSpeechBubble = () => {
     let text = `${BrandName} ${version}`;
     let top = '';
     let bottom = '';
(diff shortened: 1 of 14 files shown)


nixocio:
I would suggest to run e2e tests to verify that no regression was added.

AlexSCorey:
> I would suggest to run e2e tests to verify that no regression was added.

yep that's the plan.  Pipeline are down right now but I'm trying to run them locally.  When I get 👍  from that Ill mark this PR for needs review

unlikelyzero:
@AlexSCorey we didn't find any other teams using a i18n linter or linting strategy. The closest we got was https://www.npmjs.com/package/i18next-pseudo which is being run by the openshift/console project to check i18n of strings pre-merge

jakemcdermott (approved):
LGTM - This is basically just a refactor for some of our components. All unit, functional, and e2e test results look good

jlmitch5 (approved):
looking good!

unlikelyzero:
@tiagodread and I met with @AlexSCorey and noticed that the inventory lookup field did not appear as 'enabled' on Edit forms. Will need to investigate further and apply mergeit if not related to the changes in this PR

unlikelyzero:
@AlexSCorey i'm seeing hundreds of API requests on edit pages. I think that's why the visual tests were failing: the inventory lookups never stop firing
Screen Shot 2020-12-15 at 7 50 30 PM

unlikelyzero (dismissed):
something is blowing up the edit forms. We need to take a hard look at this one when you get back from break. 
https://share.getcloudapp.com/kpuN6LlN

jakemcdermott:
Very nice find @unlikelyzero! - looks like we have a misconfigured hook somewhere in the form. We'll just have to address this once folks are back before merging.

jakemcdermott:
Note: I decided to just fix this: https://github.com/ansible/awx/pull/8706/commits/5cb580be7a5cc72c43115126c10ec269e98802ae

jakemcdermott:
regate

PR #8710, opened 2020-11-30
Remove old radius library from requirements

wenottingham:
If it's not required by django-radius and not imported anywhere in AWX, we probably don't need it.

### DIFF
--- a/docs/licenses/python-radius.txt
+++ b/docs/licenses/python-radius.txt
@@ -1,9 +1,0 @@
-MIT License
-
-Copyright (c) <year> <copyright holders>
-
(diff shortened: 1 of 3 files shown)


shanemcd:
@wenottingham What about: 

https://github.com/ansible/awx/blob/12edbdab110b8c1f6dd95d0f8cad643bff8276b2/requirements/requirements.in#L19

https://github.com/ansible/awx/blob/12edbdab110b8c1f6dd95d0f8cad643bff8276b2/requirements/requirements.txt#L35

https://github.com/ansible/awx/blob/12edbdab110b8c1f6dd95d0f8cad643bff8276b2/requirements/requirements.txt#L44

shanemcd (approved):

ryanpetrello:
Good find, it uses `pyrad`.

```shell
~/venvs/tmp-6e7bd8b5f286fae pip install django-radius && pip freeze
-f file:///Users/ryan/.pip/wheelhouse
django-radius==1.5.0
future==0.16.0
netaddr==0.8.0
pyrad==2.4
six==1.15.0
```

ryanpetrello:
regate

PR #8718, opened 2020-12-01
add missing imports

chrismeyersfsu:
https://github.com/ansible/awx/blob/devel/setup.cfg#L16 linting the migrations is explicitly excluded because the django generated migrations don't conform to our standards. This is why the missing imports were "allowed" to merge :+1: 

I manually ran flake8 against the migrations utility file and found these missing imports.

related to https://github.com/ansible/tower/issues/4722

### DIFF
--- a/awx/main/migrations/_inventory_source_vars.py
+++ b/awx/main/migrations/_inventory_source_vars.py
@@ -1,9 +1,13 @@
 import json
+import re
+import logging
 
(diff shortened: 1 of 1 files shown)


AlanCoding (approved):

ryanpetrello (approved):

chrismeyersfsu:
recheck

PR #8719, opened 2020-12-01
correct a note about TZID support

ryanpetrello:


beeankha (approved):

jbradberry (approved):
I see that our tests do make use of it, so :rocket:

Issue #8709, opened 2020-11-30
job_explanation and result_traceback are not shown for failed jobs


AlanCoding:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Inventory updates (and maybe other job types, have not checked) with "error" status are not displaying information in the UI that is expected to be shown.

##### ENVIRONMENT
* AWX version: 16.0.0
* AWX install method: N/A
* Ansible version:  N/A
* Operating System:
* Web Browser: Chrome

##### STEPS TO REPRODUCE

Run an inventory update that produces a traceback in the task system... you could modify the code to raise an exception manually, which might be the easiest way to do this.

##### EXPECTED RESULTS

The old UI would show information from 2 important fields:

 - `job_explanation`
 - `result_traceback`

It selectively shows these when:

```json
    "status": "error",
```

##### ACTUAL RESULTS

Not enough detail:

![Screenshot from 2020-11-30 14-07-11](https://user-images.githubusercontent.com/1385596/100653168-fd2f5700-3315-11eb-82b9-c04506730f33.png)

##### ADDITIONAL INFORMATION

This is important for cases like license errors or when allocated host limit is exceeded (either system-wide or org-specific)

unlikelyzero:
https://github.com/ansible/awx/pull/8726#issuecomment-772746083

PR #8726, opened 2020-12-01
Show job traceback stdout and error details

jakemcdermott:
##### SUMMARY
For #8709 

* Render result tracebacks as stdout
* Display status explanation message when available

##### ADDITIONAL INFORMATION
job_traceback

### DIFF
--- a/awx/ui_next/src/screens/Job/JobOutput/JobOutput.jsx
+++ b/awx/ui_next/src/screens/Job/JobOutput/JobOutput.jsx
@@ -225,7 +225,7 @@ class JobOutput extends Component {
     this.state = {
       contentError: null,
       deletionError: null,
-      hasContentLoading: !props?.job?.result_traceback,
+      hasContentLoading: true,
       results: {},
       currentlyLoading: [],
       remoteRowCount: 0,
@@ -241,7 +241,6 @@ class JobOutput extends Component {
 
     this._isMounted = false;
     this.loadJobEvents = this.loadJobEvents.bind(this);
-    this.loadTracebackEvents = this.loadTracebackEvents.bind(this);
     this.handleDeleteJob = this.handleDeleteJob.bind(this);
     this.rowRenderer = this.rowRenderer.bind(this);
     this.handleHostEventClick = this.handleHostEventClick.bind(this);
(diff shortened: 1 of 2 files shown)


AlanCoding:
Looks great! thanks

awx/ui_next/src/screens/Job/JobOutput/JobOutput.jsx
```diff
@@ -195,8 +198,32 @@ class JobOutput extends Component {
...
+    // with the traceback as the stdout text
+    const { job } = this.props;
+    const remoteRowCount = 1;
+    const results = [
+      {
+        counter: -1,
+        created: null,
+        event: null,
+        type: null,
+        stdout: job?.result_traceback || '',
```
AlexSCorey: Is there a case where there would be a traceback that does not have a value in `job.result_traceback`?  If so, rather than show an empty string should we show some default string value instead?

AlanCoding: In most cases, if status is "error" then the result_traceback should be non-blank. However, I would prefer not to rely on that assumption. It's also possible (although maybe not important) that the normal standard out can be non-blank and the result_traceback can be non-blank due to inventory updates hitting a license error after attempting to save hosts (DB transaction gets rolled back).

I don't want to be opinionated, just trying to help define the envelope of possible API behaviors.

jakemcdermott: @AlexSCorey 

That's a good question. 

No - At this point, this code path isn't reached _unless_ `result_traceback` is defined: https://github.com/jakemcdermott/awx/blob/fix-8709/awx/ui_next/src/screens/Job/JobOutput/JobOutput.jsx#L221

Since any code that _does_ use this object assumes it's a string, the `''` default I'm using for this field is just some defensive coding on my part in case things change in the future.

jakemcdermott: > In most cases, if status is "error" then the result_traceback should be non-blank. However, I would prefer not to rely on that assumption. 

That's some good info, thanks @AlanCoding. The way I've currently implemented this doesn't rely on that assumption, so I think we are good here. The actual value of `status` isn't considered at all. We only check `results_traceback` when deciding to render the traceback instead of stdout.

> It's also possible (although maybe not important) that the normal standard out can be non-blank and the result_traceback can be non-blank due to inventory updates hitting a license error after attempting to save hosts (DB transaction gets rolled back).

@AlanCoding 

In this scenario, the normal standard out would be ignored and we'd only show the traceback. 

It sounds like showing _both_ might be useful, though. Do you have a test case I can use to reproduce the scenario locally? I'll see about squeezing some changes into this PR to handle both outputs.

unlikelyzero:
@jakemcdermott we're going to take our first look at this on wednesday

unlikelyzero:
@jakemcdermott @AlanCoding is ERROR the only state which includes a status explanation? Should we test FAILED state as well?

unlikelyzero:
covered in job-template-tabs and job-result-output

awx/ui_next/src/screens/Job/JobOutput/JobOutput.jsx
AlanCoding: I think violating the organization `max_hosts` is the best way to hit this.

jakemcdermott:
I'm adding some things to this PR: https://github.com/ansible/awx/pull/8726#discussion_r546026336

I'll ping for another round of reviews when done.

unlikelyzero:
https://github.com/ansible/awx/issues/8995 has some additional details on reproduction steps

one-t:
Stubs added here: https://github.com/ansible/tower-qa/pull/5938

I've tested this with Project syncs, Inventory Syncs, Job Template runs and it is working as expected now.

one-t (approved):

Issue #6189, opened 2020-03-05
Convert item lists to Tables


keithjgrant:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

We've had to do some careful custom CSS to keep the columns aligned in the data lists (e.g. JobsList, etc). In reality, this is tabular data and should probably be converted to use tables.

##### ADDITIONAL INFORMATION

~~PF Tables are not ready to do this yet. The current API is too cumbersome and not worth the work. Blocked by https://github.com/patternfly/patternfly-react/issues/2358 unless we want to roll our own components and copy in PatternFly CSS by hand.~~

All primary lists are converted in devel and pending PRs. Here are secondary lists (e.g. hosts under an inventory) that need to be converted and/or merged with the primary list component:

- [x] Host list under an Inventory
- [x] Team list under a User
- [x] Organization list under User
- [x] Notification list under a Template
- [x] Notification list under a Project
- [x] Notification list under an Organization
- [x] Any component that uses `<OptionsList>`
- [x] Access & Roles under a Team
 - [x] Instance Groups and Related tabs (in progress see https://github.com/ansible/awx/issues/10160)
 - [x] Inventories -> Inventory -> Groups tab -> Group -> Related group tab list and Hosts tab list (in progress https://github.com/ansible/awx/issues/10132)
 - [x] Teams and tabs under teams (in progress https://github.com/ansible/awx/issues/10088)
 - [x] Final search for anywhere remaining using PaginatedDataList that should be converted

keithjgrant:
Patternfly's draft PR on their portion of this work: https://github.com/patternfly/patternfly-react/pull/4950

unlikelyzero:
Please test 
https://github.com/ansible/awx/issues/6300
https://github.com/ansible/awx/issues/7626
https://github.com/ansible/awx/issues/6189
https://github.com/ansible/awx/issues/8086
https://github.com/ansible/awx/issues/7563

tiagodread:
Jobs list covered here ansible/tower-qa#5909

unlikelyzero:
@tiagodread did you test these https://github.com/ansible/awx/issues/6189#issuecomment-744424028

akus062381:
resolved.

keithjgrant:
Reopening. I've got a few more to finish converting before this is all the way completed

tiagodread:
all merged, we can safely close this.

PR #8729, opened 2020-12-01
Create PaginatedTable component

keithjgrant:
##### SUMMARY
* Adds PaginatedTable, HeaderRow, HeaderCell components to eventually replace PaginatedDataTables for item list layouts
* Converts Org list and Inventories list to use PaginatedTable

Addresses #6189

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI


awx/ui_next/src/components/PaginatedTable/PaginatedTable.jsx
```diff
@@ -0,0 +1,194 @@
...
+  } else {
+    Content = (
+      <TableComposable aria-label={dataListLabel}>
+        {headerRow}
+        <Tbody>{items.map(renderRow)}</Tbody>
+      </TableComposable>
+    );
+  }
+
+  const ToolbarPagination = (
```
AlexSCorey, line 99: This seems to be a duplicate of the same component that starts[here](https://github.com/ansible/awx/pull/8729/files#diff-b942f30c50317e77ddaf815b6d7e3bbe508ea5fa74e43318eee5a31050af5f57R135) with a few exceptions.  Could they be combined?

awx/ui_next/src/screens/Inventory/InventoryList/InventoryList.jsx
```diff
@@ -188,6 +191,17 @@ function InventoryList({ i18n }) {
           ]}
           toolbarSearchableKeys={searchableKeys}
           toolbarRelatedSearchableKeys={relatedSearchableKeys}
+          headerRow={
+            <HeaderRow defaultSortKey="name" qsConfig={QS_CONFIG}>
```
AlexSCorey: Since we will be doing this on each list I wonder if there is some sort of hook or a component we can use to generate the header rows for us.  I'm imagining something where we send it an array or column header names and it returns the correct header cells.

awx/ui_next/src/components/PaginatedTable/ActionsTd.jsx
```diff
@@ -0,0 +1,33 @@
...
+  ${props => {
+    const columns = '40px '.repeat(props.numActions || 1);
+    return css`
+      grid-template-columns: ${columns};
+    `;
+  }}
+`;
+
+export default function ActionsTd({ numActions = 1, children }) {
+  const width = numActions * 40;
```
AlexSCorey, line 22: Could you automatically get the number of actions from the number of children?  If so, doing it that way would reduce the number of props being passed in.  Additionally, I have a preference for not using styled components and css html attributes in the same file where possible. I think doing all of our styling in same place/same way is slightly easier to maintain for the future.

awx/ui_next/src/screens/Inventory/InventoryList/InventoryList.jsx
keithjgrant: HeaderRow is a custom component to take care of a bunch of the sorting logic for us already. I could toy with passing the column data as a prop or something.

I also considered exporting the HeaderRow from the item file (e.g. export InventoryHeaderRow from InventoryListItem) so that the column data all lives in the same place instead of keeping it in sync across two files

awx/ui_next/src/components/PaginatedTable/ActionsTd.jsx
keithjgrant: My initial thought was that doesn't work because some rows are missing certain actions... but now that I think about it again, the way it is now they still need to pass an empty div in the missing slots which isn't ideal...

I may need to put an index prop on the children so they can position themselves correctly in the grid

awx/ui_next/src/components/DataListToolbar/DataListToolbar.jsx
```diff
@@ -93,9 +93,11 @@ function DataListToolbar({
               onRemove={onRemove}
             />
           </ToolbarItem>
-          <ToolbarItem>
-            <Sort qsConfig={qsConfig} columns={sortColumns} onSort={onSort} />
-          </ToolbarItem>
+          {sortColumns && (
```
jlmitch5, line 96: it's nice that we've included this as a fallback until we convert everything and remove sortColumns across the app.

awx/ui_next/src/screens/Inventory/InventoryList/InventoryList.jsx
```diff
@@ -188,6 +191,17 @@ function InventoryList({ i18n }) {
...
+              <HeaderCell sortKey="name">{i18n._(t`Name`)}</HeaderCell>
```
jlmitch5, line 194: This is a very nice, declarative interface!

jakemcdermott:
This is great

* :+1: from me. I don't have any major suggestions  as far as implementation goes. I **really** like how the bulk of the conversion changes are concentrated in the individual `ListItem`.  I think this will be extremely helpful for converting the lists throughout the app. 

* I pulled this down and tested it out a bit (inventories page) and it works well. It's an overall great improvement to the UX

* ~I _did_ see some errors when running the `make ui-devel` build (npm run build). I can't tell for sure but they seem related to module importing (?). I'm wondering if this is the same issue @ jlmitch5 ran into a while back when previewing some unreleased toolbar components?~  cc @keithjgrant @jlmitch5

 **edit:** nevermind on that last part, see comment below:

jakemcdermott:
Nevermind - The error I saw was from building on node 10, rebasing, and then trying a rebuild on node 14 without reinstalling deps (the node version bump landed recently)

A full clean & build after rebase fixed things.

awx/ui_next/src/components/PaginatedTable/ActionsTd.jsx
keithjgrant: OK, I reworked this to determine based on the number of children, and use ActionItems as children to render those children to null for actions/buttons where there user doesn't have RBAC permissions

jakemcdermott:
:wave: @AlexSCorey @jlmitch5 @keithjgrant 

Is this one ready for test?

AlexSCorey (approved):
All of my concerns have been addressed.

awx/ui_next/src/screens/Inventory/InventoryList/InventoryList.jsx
jakemcdermott: > This is a very nice, declarative interface!

@jlmitch5 I agree!

awx/ui_next/src/screens/Inventory/InventoryList/InventoryListItem.jsx
```diff
@@ -85,108 +56,84 @@ function InventoryListItem({
...
+        {inventory.kind === 'smart'
+          ? i18n._(t`Smart Inventory`)
+          : i18n._(t`Inventory`)}
+      </Td>
+      <Td key="organization" dataLabel={i18n._(t`Organization`)}>
+        <Link
+          to={`/organizations/${inventory.summary_fields.organization.id}/details`}
+        >
+          {inventory.summary_fields.organization.name}
+        </Link>
```
jakemcdermott:
```suggestion
        <Link
          to={`/organizations/${inventory?.summary_fields?.organization?.id}/details`}
        >
          {inventory?.summary_fields?.organization?.name}
        </Link>
```

@keithjgrant We'll need to handle the intermediate state for list items when they're being deleted. It appears that inventories can lose their reference to the related organization during that time and so we need to avoid referencing the field to avoid an error:

Screenshot from 2020-12-17 13-04-42

(Found w/ the latest e2e/visual run :tada: ! :tada: !)

jakemcdermott (changes requested):

jakemcdermott (approved):

PR #8751, opened 2020-12-04
Separate Kubernetes Ingress TLS spec from backend spec

### LINKED ISSUE #8750: Kubernetes Ingress TLS spec is coupled to service routing spec
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

The Kubernetes Ingress templates couple the `.spec.tls` section with the `.spec.rules` vs. `.spec.backend` section.

TLS Cert serving from Ingress controllers is not tied to the `rules` nor `backend` but actually to the common name "CN" in the cert.

See:
* Kubernetes: https://v1-17.docs.kubernetes.io/docs/concepts/services-networking/ingress/#tls
    > Referencing this secret in an Ingress tells the Ingress controller to secure the channel from the client to the load balancer using TLS. You need to make sure the TLS secret you created came from a certificate that contains a Common Name (CN), also known as a Fully Qualified Domain Name (FQDN) for `sslexample.foo.com`.
* GKE: https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-multi-ssl#the_hosts_field_of_an_ingress_object
    > GKE reads the Common Name (CN) from the certificate in the Secret. If the Common Name matches the domain name in a client request, then the load balancer presents the matching certificate to the client.

This coupling can be seen here: https://github.com/ansible/awx/blob/devel/installer/roles/kubernetes/templates/deployment.yml.j2#L549-L554

The bug is that you cannot properly template out the following case
```yaml
kubernetes_ingress_hostname: "" # unset, because we want a single Service Ingress
kubernetes_ingress_tls_secret: "my-tls-secret" # defined, but will be unused due to the templating
```

hjkatz:
##### SUMMARY

Fixes #8750 

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - Installer

##### AWX VERSION

```
15.0.1
```

### DIFF
--- a/installer/roles/kubernetes/templates/deployment.yml.j2
+++ b/installer/roles/kubernetes/templates/deployment.yml.j2
@@ -546,12 +546,6 @@ metadata:
 
 spec:
 {% if kubernetes_ingress_hostname is defined %}
-{% if kubernetes_ingress_tls_secret is defined %}
-  tls:
-  - hosts:
-    - {{ kubernetes_ingress_hostname }}
-    secretName: {{ kubernetes_ingress_tls_secret }}
-{% endif %}
   rules:
   - host: {{ kubernetes_ingress_hostname }}
     http:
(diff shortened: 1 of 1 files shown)


hjkatz:
@shanemcd Thanks for the quick responses and the tips on the other PRs. Here's one more then I think we'll be able to install this into our GKE cluster successfully (Ingress and all)!

One more question, what's this repo's release schedule/process look like? We would love to have a version cut with these changes so we can move away from manually changing the templates in our install process.

shanemcd:
@hjkatz We're going to be putting out AWX 16.0 sometime early next week. Probably Monday or Tuesday.

shanemcd (approved):

Issue #2056, opened 2018-07-09
Introduce a restrictive Content Security Policy


ryanpetrello:
Add a restrictive Content-Security-Policy (https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) to awx to drastically reduce the surface area for a wide variety of in-browser security issues.

This is **really easy** to do on the backend:

```python
response['Content-Security-Policy'] = "default-src 'self'; report-uri /csp-violation/"
```

It's _hard_ on the front-end, because we'd need to identify any place we (or third-party libraries) inject inline JS and CSS.

Current roadblocks **may** include:

* Our version of Angular _itself_ (need to read up on https://code.angularjs.org/1.6.10/docs/api/ng/directive/ngCsp) 
* https://github.com/josdejong/mathjs/issues/401 (used by the network UI - it sounds like we're only using one or two small functions from this dependency, so maybe we can just lose it entirely?)
* Anything related to the pendo service?

benthomasson:
mathjs v4.0.0 removes eval: https://github.com/josdejong/mathjs/issues/682#issuecomment-367286156

mabashian:
OK so I took a look at `ng-csp` and I think I understand what it does.  By adding that directive to the dom (body) it lets angular know that it should not make use of `eval()` or inline styles/resources _internally_.  The UI will need to import `angular-csp.css` to get things like ng-cloak/ng-show/ng-hide to work since they'll no longer be able to depend on inline styles.  

UI will need to make sure it uses directives like `ng-click` rather than `onclick` as these won't be allowed by the CSP.

[Here's a short presentation on csp + angular 1.x](https://www.youtube.com/watch?time_continue=703&v=FIt9CITQFtY)

ryanpetrello:
If we addressed this (and I think we should), we're unlikely to do so in the "old" AWX UI, but instead would start enforcing it in the replacement UI (ui-next).

kdelee:
@ryanpetrello @jakemcdermott  do you have any particular suggestions for testing this, either from an API or UI perspective? This has been in a while and has not caused any problems for our API regression testing, so I'm happy with the way it is.

wenottingham:
I would think if both API regression testing and UI functional testing works... it works.

ryanpetrello:
We could *try* to come up with some testing that did something you weren't allowed to do (like ran JavaScript on the fly).

ryanpetrello:
[...]

PR #8754, opened 2020-12-04
Introduce a strict Content-Security-Policy

ryanpetrello:
see: https://github.com/ansible/awx/issues/2056

### DIFF
--- a/awx/ui_next/public/index.html
+++ b/awx/ui_next/public/index.html
@@ -7,7 +7,7 @@
       </script>
       <meta
         http-equiv="Content-Security-Policy"
-        content="default-src 'self'; connect-src 'self' ws: wss:; style-src 'self' 'nonce-{{ csp_nonce }}' *.pendo.io; script-src 'self' 'nonce-{{ csp_nonce }}' *.pendo.io; img-src 'self' *.pendo.io data:; report-uri /csp-violation/"
+        content="default-src 'self'; connect-src 'self' ws: wss:; style-src 'self' 'nonce-{{ csp_nonce }}' *.pendo.io; script-src 'self' 'nonce-{{ csp_nonce }}' *.pendo.io; img-src 'self' *.pendo.io data:;"
       />
     <% } %>
     <meta charset="utf-8" />
--- a/awx/ui_next/public/index.html
+++ b/awx/ui_next/public/index.html
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    <script nonce="{{csp_nonce}}" type="text/javascript">window.NONCE_ID = '{{csp_nonce}}';</script>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta name="theme-color" content="#000000" />
@@ -8,6 +9,7 @@
       name="description"
       content="AWX"
     />
+    <meta http-equiv="Content-Security-Policy" content="default-src 'self'; connect-src 'self' ws: wss:; style-src 'self' 'nonce-{{csp_nonce}}' *.pendo.io; script-src 'self' 'nonce-{{csp_nonce}}' *.pendo.io; img-src 'self' *.pendo.io data:; report-uri /csp-violation/">
     <title>AWX</title>
   </head>
   <body>
--- a/awx/ui_next/public/index.html
+++ b/awx/ui_next/public/index.html
@@ -1,7 +1,15 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    <script nonce="{{csp_nonce}}" type="text/javascript">window.NONCE_ID = '{{csp_nonce}}';</script>
+    <% if (process.env.NODE_ENV === 'production') { %>
+      <script nonce="{{ csp_nonce }}" type="text/javascript">
+        window.NONCE_ID = '{{ csp_nonce }}';
+      </script>
+      <meta
+        http-equiv="Content-Security-Policy"
+        content="default-src 'self'; connect-src 'self' ws: wss:; style-src 'self' 'nonce-{{ csp_nonce }}' *.pendo.io; script-src 'self' 'nonce-{{ csp_nonce }}' *.pendo.io; img-src 'self' *.pendo.io data:; report-uri /csp-violation/"
+      />
+    <% } %>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta name="theme-color" content="#000000" />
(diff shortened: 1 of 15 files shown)


unlikelyzero:
Given that the pipes are borked. I think we'll have to run against a yolo'd instance locally. The key areas to test manually would be file upload and download on windows.

jakemcdermott:
recheck

ryanpetrello (commented):
I poked at this for awhile, and it seems to work really well for me 🎉!

ryanpetrello (commented):
@jakemcdermott after this merges, gimme a heads up, because we'll need to remove the CSP header from some downstream nginx configs, too.

ryanpetrello:
This PR is *just about* wrapped up, but I think @nixocio is going to add a server-side endpoint for reporting CSP violations (at `/csp-violation/`) so that we can spot violations in our server logs.

Probably the simplest thing to do here is to report this via the general AWX logger e.g.,

`logging.getLogger('awx').error(...)`

awx/ui_next/public/index.html
```diff
@@ -1,6 +1,15 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    <% if (process.env.NODE_ENV === 'production') { %>
```
ryanpetrello, line 4: @jakemcdermott how does this get set?  Will this take effect for official AWX builds?

awx/ui_next/src/setupCSP.js
```diff
@@ -0,0 +1,30 @@
...
+      'violated-directive': e.violatedDirective,
+    },
+  };
+  if (e.sourceFile) violation['csp-report']['source-file'] = e.sourceFile;
+  if (e.lineNumber) violation['csp-report']['line-number'] = e.lineNumber;
+  if (e.columnNumber) violation['csp-report']['column-number'] = e.columnNumber;
+
+  const xhr = new XMLHttpRequest();
+  xhr.open('POST', '/csp-violation/', true);
+  xhr.setRequestHeader('content-type', 'application/csp-report');
```
ryanpetrello, line 28: This may need to be `application/json` for DRF to be happy (and not require a bunch of weird gymnastics w/ DRF POST body parsers).

awx/ui_next/public/index.html
jakemcdermott: Yes

awx/ui_next/src/setupCSP.js
jakemcdermott: The specification is to use a `Content-Type` header field of `application/csp-report`: https://w3c.github.io/webappsec-csp/2/#violation-reports

awx/ui_next/public/index.html
jakemcdermott: > how does this get set?

This isn't the final `index.html` that gets packaged in the build or collected by django. It's actually an `.ejs` template. 

Were it configurable by `react-scripts` (it isn't), I would name this file `index.html.j2.ejs`. `react-scripts` processes it as an `.ejs` template to produce the `jinja` template that django uses to serve the actual html file.

jakemcdermott: > Will this take effect for official AWX builds?

Yes

> how does this get set?

This isn't the final index.html that gets packaged in the build and collected by django. It's actually (and always has been since the angular days) an `.ejs` template.

When you run `make ui-devel`, react-scripts processes it as an `.ejs` template and outputs the jinja/djt template that django uses to render the actual index.html in production. Were it configurable by react-scripts (it isn't), I would rename this source file to `index.html.j2.ejs`. and the one django consumes to `index.html.j2` to make this pipeline more explicit.

awx/ui_next/src/setupCSP.js
jakemcdermott: @nixocio Tells me this is already wired up: https://github.com/ansible/awx/blob/devel/awx/urls.py#L25

Screenshot from 2020-12-07 20-52-47

:man_shrugging:

ryanpetrello: https://github.com/ansible/awx/commit/eacf819caf4

ryanpetrello:
recheck

ryanpetrello:
@nixocio @jakemcdermott I'm still seeing some UI test failures on this PR:

https://ansible.softwarefactory-project.io/logs/54/8754/75db4c9437bf5f4dbf32a5b69d6f15d772d9068a/check/awx-ui/8a02ba8/log-classify.html

Otherwise, I'm ready to merge this whenever this is addressed and when @unlikelyzero gives a 👍

awx/ui_next/public/index.html
nixocio: Good explanation @jakemcdermott

awx/ui_next/public/index.html
```diff
@@ -1,6 +1,15 @@
...
+      <script nonce="{{ csp_nonce }}" type="text/javascript">
+        window.NONCE_ID = '{{ csp_nonce }}';
+      </script>
+      <meta
+        http-equiv="Content-Security-Policy"
+        content="default-src 'self'; connect-src 'self' ws: wss:; style-src 'self' 'nonce-{{ csp_nonce }}' *.pendo.io; script-src 'self' 'nonce-{{ csp_nonce }}' *.pendo.io; img-src 'self' *.pendo.io data:;"
+      />
+    <% } %>
```
keithjgrant, line 12: Is it okay to use the same nonce for styles and scripts? Since a nonce can "only be used once," I'm vaguely worried some browser may reject the second one

jakemcdermott: That's a good question. I _think_ we're ok there. 

My understanding is that the nonce should be used only once _per page load_. The way `__webpack_nonce__` works seems to support this, in that it takes one `nonce` value and applies it globally to both `style-src` and `inline-src`.

ryanpetrello: Yep, the nonce is intended as a per-page load nonce, I think (what @jakemcdermott said).  I don't think two separate nonces buys us anything - it's just two server-side secrets we'd end up generating per request instead.

Maybe there's some browser out there that treats this differently, but my Chrome and Safari testing seems to work fine.  If anybody's breaking the rules or has weird support, you know which vendor it's probably gonna be 😉.

nixocio: ~I used firefox with those changes and no errors reported~.  Well, errors were reported just down below.

jakemcdermott (approved):

nixocio (approved):

unlikelyzero:
Tested it based on a build from 9am. Looks like there's an issue with websockets. I need to drop for the next couple hours but I'll retest with a latest build and compare it to devel (just to make sure it's not an existing problem)

unlikelyzero:
Does not work with Firefox ESR. Will need to rethink implementation

unlikelyzero (changes requested):
Seeing these errors in Firefox ESR in Windows 10

Content Security Policy: The page's settings blocked the loading of a resource at inline ("style-src").

jakemcdermott:
Confirming what @unlikelyzero is seeing - On Firefox ESR 78:

Screenshot from 2020-12-09 12-33-40

ryanpetrello:
Good find @unlikelyzero @jakemcdermott 😄

jakemcdermott:
@unlikelyzero found it! Thank you @unlikelyzero

unlikelyzero (approved):
Follow up ticket for Firefox ESR support filed

swathit0717:
I am running AWX - 16.0.0. Looks like "Introduce a strict Content-Security-Policy #8754" has already been merged to 16.0.0. When I tried to check the CSP header using "CSP Evaluator" Chrome extension, I observe that "No Content Security Policy has been set".
Later, I tried with AWX -15.0.0, there is CSP header already set. Is there a way to modify CSP header to remove the 'unsafe-inline' directive. Please suggest. 
imagex

Issue #7879, opened 2020-08-11
Notifications: Copy & advanced search


keithjgrant:
Follow-up on #7867

Add ability to copy notifications and flush out advanced search

##### ISSUE TYPE
 - Feature Idea

mabashian:
Started this work here: https://github.com/mabashian/awx/tree/7879-notif-copy-search.  Going to have to pivot to another bug for a bit but wanted to document the progress so I know where to look when I get back to this.

PR #8780, opened 2020-12-07
Add support for notification template copy and advanced search

mabashian:
##### SUMMARY
link #7879 

Copy:

copy

Search:

advanced_search

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/components/CopyButton/CopyButton.jsx
+++ b/awx/ui_next/src/components/CopyButton/CopyButton.jsx
@@ -34,7 +34,7 @@ function CopyButton({
     <>
       <Tooltip content={helperText.tooltip} position="top">
         <Button
-          isDisabled={isDisabled}
+          isDisabled={isLoading || isDisabled}
           aria-label={i18n._(t`Copy`)}
           variant="plain"
           onClick={copyItemToAPI}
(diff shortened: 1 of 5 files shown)


nixocio (approved):
I got this code locally, changes seem to be working as expected.

tiagodread:
@mabashian found this items:

- [ ] If the copy button was clicked twice before the first request finish the user see:

image

- [ ] Advanced search for type doesn’t work:

image

- [ ] (Not related to this PR) The user can hit the test button many times while the notification test is still running

mabashian:
@tiagodread 

> If the copy button was clicked twice before the first request finish the user see:

Good catch, this actually applies to all the copy buttons in the app but they all share the same component so the fix here should apply to all.

> Advanced search for type doesn’t work

This looks like an API bug to me.  I filed this issue: https://github.com/ansible/awx/issues/8919.  The field that you're searching on isn't actually the field you wan though.  You want to be searching on `notification_type`.  `type` is a field with only one possible value right now: `notification_template`.  Im not quite sure what the function of this field is.

notif_search

> (Not related to this PR) The user can hit the test button many times while the notification test is still running

I fixed this

mabashian:
OK @tiagodread i think this is ready for another look

tiagodread (approved):
Verified that the user can now copy notifications and it keep working on different places like 

```
templates
credentials
projects
inventories
notifications

```

Tests covered by 5804

Issue #8291, opened 2020-09-30
LDAP [Default, 1-5] settings form


marshmalien:
##### ISSUE TYPE
 - Feature Idea

URL: `/next/settings/ldap/:category/edit`

* Help text should be displayed in the form field popover
* Handle errors
* Internalization

##### ISSUE TYPE
 - Feature Idea

#### LDAP Default
  * LDAP Server URI
	  * Text
  * LDAP Bind DN
	  * Text
  * LDAP Bind Password
	  * Encrypted
  * LDAP User DN Template
	  * Text
  * LDAP Group Type
	  * Group
  * LDAP Require Group 
	  * Text
  * LDAP Deny Group
	  * Text
  * LDAP Start TLS
	  * Toggle
  * LDAP User Search
	  * CodeMirror
  * LDAP Group Search
	  * CodeMirror
  * LDAP User Attribute Map
	  * CodeMirror
  * LDAP Group Type Parameters
	  * CodeMirror
  * LDAP User Flags by Group
	  * CodeMirror
  * LDAP Organization Map
	  * CodeMirror
  * LDAP Team Map
	  * CodeMirror
  * Action Buttons
	  * Save
	  * Cancel
	  * Revert all

<img width="894" alt="Screen Shot 2020-09-30 at 4 47 52 PM" src="https://user-images.githubusercontent.com/15881645/94737957-b1275c80-033c-11eb-975e-9f6667477b43.png">

tiagodread:
Tests covered by https://github.com/ansible/tower-qa/pull/5846

PR #8761, opened 2020-12-04
Add all github category setting edit forms

marshmalien:
##### SUMMARY
Issue: #8288
This PR adds the following github category forms:
* Default `/settings/github/default/edit`
* Organization `/settings/github/organization/edit`
* Team `/settings/github/team/edit`
Screen Shot 2020-12-04 at 3 55 47 PM

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - UI

PR #8781, opened 2020-12-07
Add all LDAP (Default, 1-5) setting forms

marshmalien:
##### SUMMARY
Issue: https://github.com/ansible/awx/issues/8291
This PR adds LDAP setting forms Default, 1-5. 

Screen Shot 2020-12-04 at 4 20 40 PM

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - UI


awx/ui_next/src/screens/Setting/LDAP/LDAPEdit/LDAPEdit.jsx
```diff
@@ -1,25 +1,218 @@
...
+
+function LDAPEdit() {
+  const history = useHistory();
+  const { isModalOpen, toggleModal, closeModal } = useModal();
+  const { PUT: options } = useSettings();
+  const {
+    params: { category },
+  } = useRouteMatch('/settings/ldap/:category/edit');
+  const ldapCategory =
+    category === 'default' ? 'AUTH_LDAP_' : `AUTH_LDAP_${category}_`;
```
jakemcdermott, line 44: This is cool. The mutli-ldap subcategories were tough to deal with in the classic ui. Generating some route keys for the different ldap configs and using `ldapCategory` to reference the fields solves that problem. :+1:

awx/ui_next/src/screens/Setting/LDAP/LDAPEdit/LDAPEdit.jsx
```diff
@@ -1,25 +1,218 @@
...
+      [`${ldapCategory}USER_FLAGS_BY_GROUP`]: formatJson(
+        form[`${ldapCategory}USER_FLAGS_BY_GROUP`]
+      ),
+      [`${ldapCategory}USER_SEARCH`]: formatJson(
+        form[`${ldapCategory}USER_SEARCH`]
+      ),
+    });
+  };
+
+  const handleRevertAll = async () => {
```
jakemcdermott, line 131: @marshmalien Is there a way to use `ldapCategory` in the revert handler? The current implementation will revert the settings across the different ldap configs (default->5). I think that might be unexpected for some. It's also different from the other auth settings screens . The [github auth settings](https://github.com/ansible/awx/pull/8761/files), for example, reverts only what's on the tab.

marshmalien: Yes, good catch and an oversight on my part. I agree, this would be unexpected and we should definitely revert them by category.

marshmalien: Fixed in 79c0529

jakemcdermott (approved):

tiagodread:
This looks great!

The screenshot below is a real config that really works and might be better if the fields width gets bigger (entire line) just because the configs for LDAP can be large. cc @one-t 

image


The only ask is that we change the data-cy attributes to data-ouia-id's if possible

marshmalien:
Latest UX changes include: 
* Dynamically resize codemirror height 
* Show placeholder text in codemirror inputs
* Expand select form fields to full width per @one-t's suggestions
* Add ouiaId props 
Screen Shot 2021-01-13 at 2 46 22 PM

tiagodread:
recheck

tiagodread (approved):
This is awesome 👏🏽 
Tests covered by https://github.com/ansible/tower-qa/pull/5846

tiagodread:
regate

Issue #7004, opened 2020-05-12
Empty state appears briefly after adding a survey


appuk:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
After adding a Survey, the Empty state page appears briefly before the survey list actually populates with the newly added survey

##### STEPS TO REPRODUCE
1. Create a JT
2. Go to the surveys tab
3. Add a new survey
4. click save

##### EXPECTED RESULTS
The page should be directed to the Survey list page and the newly added survey should be present

##### ACTUAL RESULTS
The page redirects to the Survey list page, but it does not populate the survey list immediately causing it to show the Empty state for like a second
This also causes Flake in the Cypress tests, but overall the behavior is a bit annoying

![survey](https://user-images.githubusercontent.com/11617351/81752280-3047f100-947f-11ea-9784-8f64b9447b1d.gif)

unlikelyzero:
still an issue. Should be tracking in https://github.com/ansible/awx/issues/4212

tiagodread:
This was fixed by https://github.com/ansible/awx/pull/8802, once this PR is merged we can safely close this issue

tiagodread:
Fixed

PR #8802, opened 2020-12-09
Adds Spinner to Loading state

AlexSCorey:
##### SUMMARY
This addresses #4212 and offers a POC for how some changes that would need to be made throughout the app with Hosts as the example.  One other thing that I introduce here is default `isLoading` state for useRequest. To do this I pass a boolean value to useRequest.  Looking for feedback on implementation here.  Once we are good with that I can fine tune and make adjustments throughout.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
loading

### DIFF
--- a/awx/ui_next/src/components/LoadingSpinner/LoadingSpinner.jsx
+++ b/awx/ui_next/src/components/LoadingSpinner/LoadingSpinner.jsx
@@ -1,0 +1,23 @@
+import React from 'react';
+
+import { Spinner } from '@patternfly/react-core';
+import styled from 'styled-components';
+
+const UpdatingContent = styled.div`
+  position: fixed;
+  top: 50%;
+  left: 50%;
+  z-index: 300;
+  width: 100%;
+  height: 100%;
+  & + * {
+    opacity: 0.5;
+  }
+`;
+
+const LoadingSpinner = () => (
+  <UpdatingContent>
+    <Spinner />
+  </UpdatingContent>
+);
+export default LoadingSpinner;
--- a/awx/ui_next/src/screens/Host/Host.jsx
+++ b/awx/ui_next/src/screens/Host/Host.jsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useCallback, useEffect } from 'react';
 import { withI18n } from '@lingui/react';
 import { t } from '@lingui/macro';
 import {
@@ -20,29 +20,22 @@ import HostDetail from './HostDetail';
 import HostEdit from './HostEdit';
 import HostGroups from './HostGroups';
 import { HostsAPI } from '../../api';
+import useRequest from '../../util/useRequest';
 
 function Host({ i18n, setBreadcrumb }) {
-  const [host, setHost] = useState(null);
-  const [contentError, setContentError] = useState(null);
-  const [hasContentLoading, setHasContentLoading] = useState(true);
-
   const location = useLocation();
   const match = useRouteMatch('/hosts/:id');
+  const { error, isLoading, result: host, request: fetchHost } = useRequest(
+    useCallback(async () => {
+      const { data } = await HostsAPI.readDetail(match.params.id);
+      setBreadcrumb(data);
+      return data;
+    }, [match.params.id, setBreadcrumb])
+  );
 
   useEffect(() => {
-    (async () => {
-      setContentError(null);
-      try {
-        const { data } = await HostsAPI.readDetail(match.params.id);
-        setHost(data);
-        setBreadcrumb(data);
-      } catch (error) {
-        setContentError(error);
-      } finally {
-        setHasContentLoading(false);
-      }
-    })();
-  }, [match.params.id, location, setBreadcrumb]);
+    fetchHost();
+  }, [fetchHost, location]);
 
   const tabsArray = [
     {
@@ -77,7 +70,7 @@ function Host({ i18n, setBreadcrumb }) {
     },
   ];
 
-  if (hasContentLoading) {
+  if (isLoading) {
     return (
       <PageSection>
         <Card>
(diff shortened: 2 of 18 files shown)


awx/ui_next/src/screens/Host/Host.jsx
```diff
@@ -116,23 +101,25 @@ function Host({ i18n, setBreadcrumb }) {
...
-            <Route path="/hosts/:id/groups" key="groups">
-              <HostGroups host={host} />
-            </Route>,
-            <Route path="/hosts/:id/completed_jobs" key="completed-jobs">
-              <JobList defaultParams={{ job__hosts: host.id }} />
-            </Route>,
-          ]}
+
+          <Route path="/hosts/:id/details" key="details">
+            {isLoading ? <ContentLoading /> : <HostDetail host={host} />}
```
jlmitch5: I understand we are going to need to do some across-the-app kind of changes.

Is there a way we can use this `isLoading ? <ContentLoading /> : <Screen />` pattern always?  if there are some that we have to do this, and then sometimes you have to pass as prop, it's going to be make it more difficult to implement as well as when we add new views.

awx/ui_next/src/screens/Template/TemplateList/TemplateList.jsx
```diff
@@ -163,7 +163,7 @@ function TemplateList({ i18n }) {
   const addButton = (
     <AddDropDownButton key="add" dropdownItems={addDropDownButton} />
   );
-
+  console.log(isLoading, 'isLoading');
```
jlmitch5: remove this log

jlmitch5 (changes requested):
Love the feature, would be great if there was a consistent way of doing it across all views in the app.  I'd definitely prefer not passing as props if we can just render it at the route level.

For screens that make requests, they can set isLoading internally and render the ContentLoading component, I think that's fine.

awx/ui_next/src/screens/Host/Host.jsx
AlexSCorey: Sometimes, if the child component does not make any api calls on its own I can just put the content loading like isLoading ? <contentLoading> : <screen> other times, if the child or grand child makes api calls I need to pass the loading state down so that it doesn’t flash the loading screen

awx/ui_next/src/components/LoadingSpinner/LoadingSpinner.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+
+import { Spinner as PFSpinner } from '@patternfly/react-core';
+import styled from 'styled-components';
+
+const Spinner = styled(PFSpinner)`
+  position: fixed;
+  top: 50%;
+  left: 50%;
```
keithjgrant, line 9: This doesn't exactly center the element; it puts the top left corner in the center. You probably want something more like

```
top: calc(50% - 28px);
left: calc(50% - 28px);
```
(28px is about half the spinner's size)

awx/ui_next/src/components/LoadingSpinner/LoadingSpinner.jsx
```diff
@@ -0,0 +1,26 @@
...
+`;
+const UpdatingContent = styled.div`
+  position: absolute;
+  width: 100%;
+  height: 100%;
+  z-index: 300;
+  & + * {
+    opacity: 0.5;
+  }
+`;
```
keithjgrant: the `z-index` is the only style doing anything here. delete the rest

awx/ui_next/src/components/LoadingSpinner/LoadingSpinner.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+
+import { Spinner as PFSpinner } from '@patternfly/react-core';
+import styled from 'styled-components';
+
+const Spinner = styled(PFSpinner)`
```
keithjgrant: nit-picky, but it makes more sense to me to list the container (`UpdatingContent`) above the spinner

AlexSCorey:
Loading state meetings decisions reached:

- For Loading state: As a first step simply swap out the add the Spinner to the Content loading component without adjusting which components are rendering a loading state. From there see if we can eliminate as much flashing ass possible. Also initiate useRequest with isLoading=true state to eliminate the rendering of  ContentError before the request is fired off/`isLoading` is true.
For spinner vs skeleton: Spinner first and then consider and implement skeleton where appropriate.

jlmitch5 (approved):
Based on our meeting, I'm happy with the path we are moving forward with the loading state, great job on this Alex!

unlikelyzero:
@tiagodread this will be a great use of the cy.route2 network delay intercept

tiagodread:
Hey @AlexSCorey found so far:

- JT Schedules tab showing the spinner twice:

schedules

tiagodread:
@AlexSCorey verified that the loading spinner is displayed in several places like, lists, details, forms, etc. Here are places where doesn't contain loading spinner while the data is been fetched

- Dashboard 
- Workflow Approvals -> Item details 
- Job Templates -> Survey tab
- Inventories list (tables) sort, search, pagination
- Organization list (tables) sort, search, pagination

tiagodread:
This PR isn't fixing the problem https://github.com/ansible/awx/issues/7004, still an issue @AlexSCorey

tiagodread:
Trying to edit a host, after hit save button the data isn't updated on the details view, we have 2 faling tests because of it.

edit-host

unlikelyzero:
@AlexSCorey @tiagodread @one-t let's sync tomorrow on how this transition state impacts our lighthouse performance score. Shouldn't take more than 15 minutes

tiagodread (approved):
All items were addressed, tests covered manually and https://github.com/ansible/tower-qa/pull/5861

PR #8833, opened 2020-12-10
Sync the is_superuser flag when the members relationship changes

AlanCoding:
This was giving the wrong behavior before.

work started by @rebeccahhh at

https://github.com/ansible/awx/compare/devel...rebeccahhh:devel

### DIFF
--- a/awx/main/signals.py
+++ b/awx/main/signals.py
@@ -123,10 +123,23 @@ def sync_superuser_status_to_rbac(instance, **kwargs):
 
 def sync_rbac_to_superuser_status(instance, sender, **kwargs):
     'When the is_superuser flag is false but a user has the System Admin role, update the database to reflect that'
-    if kwargs['action'] in ['post_add', 'post_remove']:
-        if instance.singleton_name == ROLE_SINGLETON_SYSTEM_ADMINISTRATOR:
-            new_status_value = bool(kwargs['action'] == 'post_add')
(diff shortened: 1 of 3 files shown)


awx/main/signals.py
```diff
@@ -121,6 +121,27 @@ def sync_superuser_status_to_rbac(instance, **kwargs):
         Role.singleton(ROLE_SINGLETON_SYSTEM_ADMINISTRATOR).members.remove(instance)
 
 
+def sync_rbac_to_superuser_status(instance, sender, **kwargs):
+    'When the is_superuser flag is false but a user has the System Admin role, update the database to reflect that'
+    if kwargs['action'] in ['post_add', 'post_remove', 'post_clear']:
+        new_status_value = bool(kwargs['action'] == 'post_add')
+        if hasattr(instance, 'singleton_name'):  # duck typing, role.members.add() vs user.roles.add()
```
AlanCoding, line 128: If someone hates me too bad for this, you could use the `reverse` kwarg, but how do you know at a glance which is the reverse and which is the non-reverse?

rebeccahhh (approved):
we have passing testing

awx/main/signals.py
jbradberry: I was going to say something about the use of `singleton_name` here, but apparently that's an actual field on the `Role` model.

jbradberry (approved):
This looks reasonable to me.

PR #8865, opened 2020-12-11
Improve the behavior of EE resolution for ad hoc commands

jbradberry:
##### SUMMARY
- call resolve_execution_environment during AdHocCommand.save()
- wrap the fallback call of the resolver in tasks.py in disable_activity_stream()

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
awx: 16.0.0
```

### DIFF
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -882,8 +882,11 @@ class BaseTask(object):
 
     def build_execution_environment_params(self, instance):
         if instance.execution_environment_id is None:
-            self.instance = instance = self.update_model(
-                instance.pk, execution_environment=instance.resolve_execution_environment())
+            from awx.main.signals import disable_activity_stream
+
+            with disable_activity_stream():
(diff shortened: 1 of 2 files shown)


awx/main/tasks.py
```diff
@@ -882,8 +882,11 @@ def get_path_to(self, *args):
 
     def build_execution_environment_params(self, instance):
         if instance.execution_environment_id is None:
-            self.instance = instance = self.update_model(
-                instance.pk, execution_environment=instance.resolve_execution_environment())
+            from awx.main.signals import disable_activity_stream
+
+            with disable_activity_stream():
+                self.instance = instance = self.update_model(
+                    instance.pk, execution_environment=instance.resolve_execution_environment())
```
AlanCoding, line 889: I don't think this should ever be needed in this new system, but it shouldn't do harm

AlanCoding (approved):

PR #8867, opened 2020-12-11
Wrap functions in useCallback to avoid extra requests

nixocio:
Wrap functions in useCallback to avoid extra requests.

`make ui-devel` I noticed a few extra requests with the build.

with useCallback.

image

without useCallback.

image


jakemcdermott (approved):

jakemcdermott:
e2e is :green_circle: : except for a known flakey test that's on devel: https://dashboard.cypress.io/projects/ui2mq7/runs/3616/overview

This is a refactor and tests lgtm - merging.

Issue #7692, opened 2020-07-21
Support promptable fields on schedules


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
When creating/editing a schedule for a JT or a WFJT we need to expose fields that are promptable on launch in the form so that the user can specify values for those fields before saving the schedule.  In the old UI we exposed all the promptable fields (except extra variables) through a button beneath the form like so:

![schedule_prompts](https://user-images.githubusercontent.com/9889020/88088499-80ff5880-cb58-11ea-9fc9-6930f27988e8.gif)

~~We could follow the same pattern in ui_next but I've never really loved that pattern and it's a bit of a one-off solution.  I can't think of any other forms where some of the fields are in the form and the rest are in a modal that gets opened after clicking a button.  We did that originally because we could reuse the launch prompt code and get it done pretty quickly.  I feel like these prompts are _part_ of the form and while it _could_ make the form quite long (particularly the survey portion) I'm wondering if these promptable fields should just be included in the form.  This would reduce clicks and potentially make the schedule form much easier to read/understand if all the fields were visible in once place.~~

There's also odd inconsistencies with extra variables where that gets shown in the form and not in the modal.  I just feel like it'd be better if it was all in one form or if we rethought the UX of creating a schedule in general.

@trahman73 @wenottingham I'm open to discussion on this one if you all have strong feelings and/or other ideas.

UI team decided to duplicate the workflow from the old ui.  1 thing to consider, and an areas of improvement, is to figure out a way to let the user know that there is some missing, required, prompt data (ie. inventory). @trahman73

mabashian:
I'm going to mark this as blocked until we nail down the UX

unlikelyzero:
To the future tester: we should audit cypress/integration/job-templates/job-template-prompt-on-launch.spec.js for applicable testcases

AlexSCorey:
after conversation with @trahman73 @mabashian @wenottingham we decided that this issue should utilize a single form that has headers for the Schedule form, and then a sub section for the Promptable fields. Will probably need another sub section to differentiate Surveys from the rest of the Promptable fields.  Mock up to come....

AlexSCorey:
Met with the UI team today to discuss this issue. We decided that for several reasons we will duplicate the workflow on from the old ui.

unlikelyzero:
These may or may not need to split out into this ticket https://github.com/ansible/awx/issues/8692

tiagodread:
[...]

Issue #6510, opened 2020-03-31
Prompted fields on schedules are broken


dsesami:
EDIT: see https://github.com/ansible/awx/issues/6510#issuecomment-692930141

##### ISSUE TYPE
- Bug/Feature-ish

##### SUMMARY
Follow-up to work in https://github.com/ansible/awx/issues/5859

See this screenshot:
![Screenshot from 2020-03-31 15-09-59](https://user-images.githubusercontent.com/19942822/78066651-5a2bd500-7363-11ea-909d-cacfb6ab0fc2.png)

"Prompted Fields" doesn't look like a header for a group of items below it. It just looks like a regular item with no value. The work in this involves creating a proper subsection for prompted fields (if any exist) for the schedule.

dsesami:
Prompted fields on a JT schedule are actually entirely broken now and do not appear at all.

![Screenshot from 2020-09-15 15-29-19](https://user-images.githubusercontent.com/19942822/93255694-42f17000-f768-11ea-89b3-0b7037105de6.png)

unlikelyzero:
Note: we should convert the stubs to tests on at least prompts on schedule details

mabashian:
Linking this with #7692

unlikelyzero:
coverage added

PR #8869, opened 2020-12-11
Fix prompted values section of schedule details

mabashian:
##### SUMMARY
link #6510 

Talked with @marshmalien and we came up with this UX pattern:

Screen Shot 2020-12-11 at 11 11 48 AM

We think that this most closely matches the direction that PF is going with subforms.

I made the same UX changes to the PromptDetail component since we also show prompted values in a similar way in that component.

@one-t @unlikelyzero @tiagodread - to whoever tests this.  Note that prompting on the schedules form is not currently supported.  See #7692.  So you'll have to interact with the api to set this up.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/components/PromptDetail/PromptDetail.jsx
+++ b/awx/ui_next/src/components/PromptDetail/PromptDetail.jsx
@@ -5,7 +5,7 @@ import { withI18n } from '@lingui/react';
 import { t, Trans } from '@lingui/macro';
 import { Link } from 'react-router-dom';
 import styled from 'styled-components';
-import { Chip, Divider } from '@patternfly/react-core';
+import { Chip, Divider, Title } from '@patternfly/react-core';
 import { toTitleCase } from '../../util/strings';
 
 import CredentialChip from '../CredentialChip';
@@ -18,9 +18,19 @@ import PromptInventorySourceDetail from './PromptInventorySourceDetail';
 import PromptJobTemplateDetail from './PromptJobTemplateDetail';
 import PromptWFJobTemplateDetail from './PromptWFJobTemplateDetail';
 
-const PromptHeader = styled.h2`
-  font-weight: bold;
-  margin: var(--pf-global--spacer--lg) 0;
+const PromptTitle = styled(Title)`
+  margin-top: 40px;
+  --pf-c-title--m-md--FontWeight: 700;
+  grid-column: 1 / -1;
+`;
+
+const PromptDivider = styled(Divider)`
+  margin-top: var(--pf-global--spacer--lg);
+  margin-bottom: var(--pf-global--spacer--lg);
+`;
+
+const PromptDetailList = styled(DetailList)`
+  padding: 0px 20px;
 `;
 
 function formatTimeout(timeout) {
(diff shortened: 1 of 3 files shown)


awx/ui_next/src/components/Schedule/ScheduleDetail/ScheduleDetail.jsx
```diff
@@ -140,18 +151,25 @@ function ScheduleDetail({ schedule, i18n }) {
     survey_enabled,
   } = launchData || {};
 
+  const showCredentialsDetail =
+    ask_credential_on_launch && credentials.length > 0;
+  const showInventoryDetail = ask_inventory_on_launch && inventory;
```
marshmalien, line 165: I like these additions. Do you think we should do this for other fields like `ask_skip_tags_on_launch` and `ask_tags_on_launch`?

I created a job template with zero skip tags and selected skip tags prompt on launch. Below we show a Prompted Values section, but no skip tags detail field. Do we want to show an empty Prompted Values section?
Screen Shot 2020-12-11 at 4 17 54 PM

awx/ui_next/src/components/PromptDetail/PromptDetail.jsx
```diff
@@ -18,9 +18,19 @@ import PromptInventorySourceDetail from './PromptInventorySourceDetail';
 import PromptJobTemplateDetail from './PromptJobTemplateDetail';
 import PromptWFJobTemplateDetail from './PromptWFJobTemplateDetail';
 
-const PromptHeader = styled.h2`
-  font-weight: bold;
-  margin: var(--pf-global--spacer--lg) 0;
+const PromptTitle = styled(Title)`
+  margin-top: 40px;
```
marshmalien:
```suggestion
  margin-top: var(--pf-global--spacer--xl);
```
We could make them all PatternFly utility variables. Either way works for me.

awx/ui_next/src/components/PromptDetail/PromptDetail.jsx
```diff
@@ -18,9 +18,19 @@ import PromptInventorySourceDetail from './PromptInventorySourceDetail';
...
+  grid-column: 1 / -1;
+`;
+
+const PromptDivider = styled(Divider)`
+  margin-top: var(--pf-global--spacer--lg);
+  margin-bottom: var(--pf-global--spacer--lg);
+`;
+
+const PromptDetailList = styled(DetailList)`
+  padding: 0px 20px;
```
marshmalien:
```suggestion
  padding: 0px var(--pf-global--spacer--lg);
```

awx/ui_next/src/components/Schedule/ScheduleDetail/ScheduleDetail.jsx
mabashian: Oof, yes I'll make those same changes for those fields.  Good catch.

mabashian:
OK @marshmalien I think this is ready for another look

mabashian:
Before we merge this I need to write better tests around checking for when the `Prompted Values` section is hidden/shown.

marshmalien (approved):
🧨 Tested out the latest changes locally and this look good to me. Thanks for adding test coverage for those two additional cases!

one-t:
This looks good to me- we have a stub for this already in tower-qe.
image

Issue #8856, opened 2020-12-10
[ui_next] No identifying information that an ad-hoc command was run as an ad-hoc command in job details view


unlikelyzero:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
No identifying information that an ad-hoc command was run as an ad-hoc command in job details view

##### ENVIRONMENT
* AWX version: 12/10

##### STEPS TO REPRODUCE
1. Run ad-hoc command job
2. Navigate to details view of that job

##### EXPECTED RESULTS
Something to describe job type as ad-hoc command vs playbook run

##### ACTUAL RESULTS
No identifying information

##### ADDITIONAL INFORMATION
<img width="622" alt="Screen Shot 2020-12-10 at 6 33 01 PM" src="https://user-images.githubusercontent.com/12446869/101842496-2752f700-3b16-11eb-8307-7f313aaf5554.png">

unlikelyzero:
Covered by unit test

PR #8888, opened 2020-12-14
 Make display type unique for different job types and consistent between job list and details

jakemcdermott:
##### SUMMARY
for #8856 

Make display type unique for different job types and consistent between job list and details:

- Fix key for adhoc command display type in job list item
- Use `type` field as key for job display type on job details
- Match displayed type text for job list and details


keithjgrant (approved):
lgtm

Issue #8890, opened 2020-12-14
Advanced search host filter chips not rendered in smart inventory form


mabashian:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
This logic here: https://github.com/ansible/awx/blob/devel/awx/ui_next/src/components/Lookup/HostFilterLookup.jsx#L275 means that we will only ever render chips that fit our default search options.  If you create a filter using the advanced search it _appears_ to not work (even though it is working, just not rendering).

![host_filter](https://user-images.githubusercontent.com/9889020/102125727-1d800b00-3e18-11eb-8ef8-1643f0336d98.gif)

nixocio:
I was not able to re-create this on devel. It seems like it was already fixed.

akus062381:
this is all set. closing

akus062381:
@mabashian - Tiago brought something to my attention that I had missed before, that when you perform an advanced search here, the chip showing the filter parameter is not visible. It is only visible when performing a basic search. I am reopening this and putting it back into needs_devel

tiagodread:
The modals in general is not displaying all items in the toolbar when there is only on item in the list because looks like the modal adjust the high according to the number of items in the list. I Wonder if the modal should have a defined high even if there is only one item, to avoid this behaviour @mabashian 

![image](https://user-images.githubusercontent.com/12464858/118705936-32116680-b7ef-11eb-88e6-33bdaad614ce.png)

mabashian:
OK, it seems like there are some fields where the chip is displayed correctly but there are some advanced fields where it is not.  If you search for `id` or `inventory` for example (both id's) the chip renders correctly.  If you search for `name`, `description`, or `type` the chip does not render correctly.

Example using name:

![host_filter](https://user-images.githubusercontent.com/9889020/118708343-9d583a00-b7e9-11eb-9322-de0ea0812895.gif)

akus062381:
this is fixed and verified! stubs already exist.

Issue #7700, opened 2020-07-21
Add shortcut to create smart inventory from /#/hosts list


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
In the old UI we had this button on the Hosts list that would let users create a smart inventory using the current search filter items:

<img width="1535" alt="Screen Shot 2020-07-21 at 5 21 39 PM" src="https://user-images.githubusercontent.com/9889020/88108396-c0d53880-cb76-11ea-8579-bdf0e2b97b1c.png">
<img width="1532" alt="Screen Shot 2020-07-21 at 5 22 05 PM" src="https://user-images.githubusercontent.com/9889020/88108397-c16dcf00-cb76-11ea-8060-b69a5f36edd3.png">

All it would do is pass the search params though to the form and then pre-populate that field.  I think we should do something similar in ui_next.  The button is only clickable when the user has specified at least one search term so the button could be shown only after a search is executed _or_ we could always show it and just disable it when there are no searches.

tiagodread:
Tests covered by ansible/tower-qa#5867

PR #8896, opened 2020-12-14
Adds smart inventory button on host list

mabashian:
##### SUMMARY
link #7700 

smart_inv

Note that I filed https://github.com/ansible/awx/issues/8890 while working on this PR.  If you attempt to create a filter that does not match one of these: https://github.com/ansible/awx/blob/devel/awx/ui_next/src/components/Lookup/HostFilterLookup.jsx#L74-L109 then the tag _will not show up_ in the host filter field.  The value is still there and you can save.  You can also see the value inside the modal.

The Smart Inventory button is disabled until a search is performed.  The button is hidden from users that don't have the ability to create a host as was the case in the old UI.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/components/Lookup/HostFilterLookup.jsx
+++ b/awx/ui_next/src/components/Lookup/HostFilterLookup.jsx
@@ -12,6 +12,7 @@ import {
   FormGroup,
   InputGroup,
   Modal,
+  Tooltip,
 } from '@patternfly/react-core';
 import ChipGroup from '../ChipGroup';
 import Popover from '../Popover';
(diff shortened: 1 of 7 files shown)


jakemcdermott (approved):

tiagodread:
Hey @mabashian do you mind rebasing this one?

tiagodread:
@mabashian some feedback:

- ~Smart host filter fields empty and disabled after searching hosts by description:~

smart1

- ~I cant open the Smart host filter modal when the search term doesn't match because the button is disabled:~

smart

I found these two items above because i didn't know that I need fill an Organization first, to be able to click on **Smart host filter lookup button** This is not intuitive, in the old UI the users were able to see this help text:

image

We are missing this helper text in the ui_next

- Smart inventory button on kebab after advanced search:

image

tiagodread:
@mabashian Smart inventory button on kebab after advanced search:

image

mabashian:
@tiagodread ok I fixed the styling of that button.  The options are getting chopped off at certain resolutions.  I think this is an issue for PF though since we use their Toolbar component under the hood:

Screen Shot 2021-01-15 at 2 54 43 PM

tiagodread (approved):
Tests covered by https://github.com/ansible/tower-qa/pull/5867

Issue #8741, opened 2020-12-03
Dynamic AWS inventory source overwrite existing variables when host already exist


anthoc:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

We are using multiple inventory source for the same inventory and for same hosts.
Some variable are from a dynamic AWS source (like infrastructure information)
Some are from a yml file pulled on a gitlab server (custom variables)
Some are declared manually inside the variables field of the host in the AWX dashboard.

We configured each source with "overwrite" checked, and "overwrite variables" unchecked

When we are running an inventory sync from the gitlab source, the variables added from aws sync and manually added are still there when the sync is done.

When we are running an inventory sync from the AWS source, only the AWS variables are there after the sync, all the variables obtained from gitlab sync and manually added are removed. 

And the group issued from gitlab source are gone too.

##### ENVIRONMENT
* AWX version: 15.0.1
* AWX install method: docker
* Ansible version:  2.9.14
* Operating System: Debian 10
* Web Browser: Google Chrome 87.0.4280.88

##### STEPS TO REPRODUCE

- Add an AWS source in an inventory with "overwrite checked" and "overwrite variables" unchecked
- Add a gitlab source in the same inventory with "overwrite checked" and "overwrite variables" unchecked
- Sync the gitlab source => the host is created and gitlab variables and groups are added
- Sync the AWS source => gitlab variables and groups are removed, AWS variables and groups are added
- Sync the gitlab source => gitlab variables and groups are added and AWS variables and groups are still there

##### EXPECTED RESULTS

I am not sure, but I expected the same behaviour from both sources (aws and gitlab).
From what I understood, with this configuration variables issued from different sources are supposed to coexist when the host have the same name. 

##### ACTUAL RESULTS

Existing variables from other sources are always removed by AWS source (maybe the host is completely recreated)

anthoc:
Actually I checked, and even if the host have the same name, AWS inventory source always completely recreate the host (the ID is different) when there are variables manually added or from a different source, whereas gitlab source just append the variables to the existing host, when we trigger the sync from AWS.

anthoc:
I enabled the DEBUG mode in the option of the AWS source and here is what I saw : 

```
14.483 DEBUG Finished loading from source: /tmp/awx_730_09rh8iqc/aws_ec2.yml
14.483 INFO  Processing JSON output...
14.483 DEBUG Loaded group: all
14.483 DEBUG Loaded group: aws_ec2
[...]

PR #8908, opened 2020-12-15
Flip the instance ID var for AWS to `instance_id`

wenottingham:
This is always returned by the plugin, ec2_id was what the old script returned.

##### SUMMARY
See https://github.com/ansible/awx/issues/8741

Easiest way to check: without this, every time you sync it will recreate all the hosts (look at the "id" of the hosts to confirm)

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API


AlanCoding (approved):
I'm wondering if this is a problem for other sources, I thought that with the config changes we were going to put in fallbacks for everything like https://github.com/ansible/awx/pull/7321, but it's true that the old values are no longer necessary

PR #8912, opened 2020-12-16
Update consumers.py

Saurabh-Thakre:
Caught a syntax issue at line number 78, modified the syntax to return correct value of nonce_diff.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

  - Logging Error

##### AWX VERSION

```
AWX version: 16.0.0
```

##### ADDITIONAL INFORMATION

```
--- Logging error ---
    raise ValueError("Potential replay attack or machine(s) time out of sync by {nonce_diff} seconds.")
ValueError: Potential replay attack or machine(s) time out of sync by {nonce_diff} seconds.
BlockingIOError: [Errno 11] Resource temporarily unavailable
```
Above is what the code is returning as of now. We could see ValueError is not returning value of **nonce_diff** and instead its treating nonce_diff as a plain string in quotes.

So just modified line number 78 with correct syntax, which will return the value of nonce_diff

### DIFF
--- a/awx/main/consumers.py
+++ b/awx/main/consumers.py
@@ -75,7 +75,7 @@ class WebsocketSecretAuthHelper:
         nonce_diff = now - nonce_parsed
         if abs(nonce_diff) > nonce_tolerance:
             logger.warn(f"Potential replay attack or machine(s) time out of sync by {nonce_diff} seconds.")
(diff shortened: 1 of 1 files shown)


Spredzy (approved):
LGTM

Issue #8207, opened 2020-09-22
Disable cancel button when job is complete


dsesami:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
This code is reused in a few places. One example is Job Template -> Completed Jobs
![Screenshot from 2020-09-22 16-25-12](https://user-images.githubusercontent.com/19942822/93933804-33d26b00-fcf0-11ea-9e70-1217d192be1d.png)

The `Cancel Job` button is enabled even when jobs are complete. Also, this flow seems a bit unintuitive. Should this be a one-button click cancel button that switches out with the launch job button, like the other UI?

mabashian:
> The Cancel Job button is enabled even when jobs are complete

We can fix this.

It was a conscious decision to make this a multi-select thing.  Of course we could go back to individual buttons if needed.

unlikelyzero:
this is a dupe of https://github.com/ansible/awx/issues/8207

PR #8928, opened 2020-12-16
Disable cancel button on jobs list when one or more selected jobs is not running

mabashian:
##### SUMMARY
link #8207 

I also added separate messaging for jobs that you don't have permission to cancel v. jobs that cannot be canceled because they aren't running.

Screen Shot 2020-12-16 at 12 57 40 PM

Screen Shot 2020-12-16 at 12 53 35 PM

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/components/JobList/JobListCancelButton.jsx
+++ b/awx/ui_next/src/components/JobList/JobListCancelButton.jsx
@@ -53,7 +53,7 @@ function JobListCancelButton({ i18n, jobsToCancel, onCancel }) {
       return (
         <div>
           {cannotCancelPermissions.length > 0 && (
-            <>
+            <div>
               {i18n._(
                 '{numJobsUnableToCancel, plural, one {You do not have permission to cancel the following job:} other {You do not have permission to cancel the following jobs:}}',
                 {
@@ -61,10 +61,10 @@ function JobListCancelButton({ i18n, jobsToCancel, onCancel }) {
                 }
               )}
               {' '.concat(cannotCancelPermissions.join(', '))}
-            </>
+            </div>
           )}
           {cannotCancelNotRunning.length > 0 && (
-            <>
+            <div>
               {i18n._(
                 '{numJobsUnableToCancel, plural, one {You cannot cancel the following job because it is not running:} other {You cannot cancel the following jobs because they are not running:}}',
                 {
@@ -72,7 +72,7 @@ function JobListCancelButton({ i18n, jobsToCancel, onCancel }) {
                 }
               )}
               {' '.concat(cannotCancelNotRunning.join(', '))}
-            </>
+            </div>
           )}
         </div>
       );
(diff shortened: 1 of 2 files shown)


awx/ui_next/src/components/JobList/JobListCancelButton.jsx
```diff
@@ -33,20 +40,40 @@ function JobListCancelButton({ i18n, jobsToCancel, onCancel }) {
...
+              )}
+              {' '.concat(cannotCancelPermissions.join(', '))}
+            </div>
+          )}
+          {cannotCancelNotRunning.length > 0 && (
+            <div>
+              {i18n._(
+                '{numJobsUnableToCancel, plural, one {You cannot cancel the following job because it is not running:} other {You cannot cancel the following jobs because they are not running:}}',
+                {
+                  numJobsUnableToCancel: cannotCancelNotRunning.length,
```
jakemcdermott, line 71: @mabashian 
lgtm

Since we're just listing out the names, it seems the tooltip behavior would still be reasonable even when there's a _lot_ of jobs to list.

jakemcdermott (approved):

awx/ui_next/src/components/JobList/JobListCancelButton.jsx
```diff
@@ -62,7 +89,9 @@ function JobListCancelButton({ i18n, jobsToCancel, onCancel }) {
   };
 
   const isDisabled =
-    jobsToCancel.length === 0 || jobsToCancel.some(cannotCancel);
+    jobsToCancel.length === 0 ||
+    jobsToCancel.some(cannotCancelBecausePermissions) ||
+    jobsToCancel.some(cannotCancelBecauseNotRunning);
 
   const cancelJobText = i18n._(
     '{zeroOrOneJobSelected, plural, one {Cancel job} other {Cancel jobs}}',
```
jakemcdermott, line 97: (This isn't directly related to the changes to this PR, but something I noticed).

We show `Cancel job(s)` for one action button and just `Delete` for the other. In my opinion, leaving it as `Cancel` would be more cohesive.

edit: Then again, maybe that could be ambiguous? Since sometimes a `Cancel` button means to close out of something e.g a form or modal.

mabashian: > edit: Then again, maybe that could be ambiguous? Since sometimes a Cancel button means to close out of something e.g a form or modal.

Yea I did the exact same thing.  At first I made it `Cancel` to be consistent but thin I was like wait.... that's kind of weird.  Ended up going with Cancel job/jobs.  It's a weird case and I'm happy to change it if we can come up with something better.  cc @trahman73 if you're around and have any ideas.

trahman73: In the old UI, cancel used to be an icon that had a tooltip that said "Cancel job" I believe? Now that we are using buttons, keeping the "Cancel job" text feels right to me as just having "Cancel" follows our pattern of leaving an editable area.

Issue #8936, opened 2020-12-17
Page crash trying to launch a job template


tiagodread:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Page crash trying to launch a job template from different places (list, details, etc)

##### STEPS TO REPRODUCE

1. Create a JT
2. Try to launch

##### ACTUAL RESULTS

![image](https://user-images.githubusercontent.com/12464858/102494258-2577bf00-4053-11eb-98dc-cd7f89b5e257.png)

PR #8938, opened 2020-12-17
Don't build job redirect until data available

jakemcdermott:
##### SUMMARY
Job and template route cleanup, addresses #8936 

- Add missing slash to templates routing key
- In job redirect, don't build redirect until data available


mabashian (approved):

tiagodread (approved):
Broken tests are green now

Issue #8884, opened 2020-12-14
If browser not set to a locale with strings, templated text appears


adpavlov:
##### ISSUE TYPE
 - Bug Report

##### ENVIRONMENT
* AWX version: 16.0.0
* AWX install method: docker on linux
* Ansible version:  2.9.15
* Operating System: macOS 11.0.1
* Web Browser: Firefox, Safari, Edge, Opera

##### STEPS TO REPRODUCE
Upgrade AWX to 16.0.0

##### EXPECTED RESULTS

Normal UI
##### ACTUAL RESULTS

Some elements are not decoded properly

![image](https://user-images.githubusercontent.com/51116816/102063078-9fffcf00-3e06-11eb-890b-a89694c31a15.jpeg)

![image](https://user-images.githubusercontent.com/51116816/102063160-b86fe980-3e06-11eb-900d-a9f9d7926826.jpeg)

![image](https://user-images.githubusercontent.com/51116816/102063228-ca518c80-3e06-11eb-986b-1c088dbcc894.jpeg)

unlikelyzero:
@adpavlov try:
```
make clean-ui
make ui-devel
```

adpavlov:
> @adpavlov try:
> 
> ```
> make clean-ui
> make ui-devel
> ```

Sorry, not sure I know where to issue those commands. I'm using docker image from dockerhub, not building it myself.
Same installation works fine from windows based browsers.

jakemcdermott:
@adpavlov 

Do you see any errors in the browser console on page load? (Cmd + Opt + J, I believe)

adpavlov:
> @adpavlov
> 
> Do you see any errors in the browser console on page load? (Cmd + Opt + J, I believe)

nothing except insecure protocol warning.
And this is what I see in web inspector

`<h2 class="pf-c-title pf-m-3xl">Welcome to Ansible {brandName}! Please Sign In.</h2>`
vs windows browsing
`<h2 class="pf-c-title pf-m-3xl">Welcome to Ansible AWX! Please Sign In.</h2>`

jakemcdermott:
@adpavlov  

We weren't able to reproduce this using the default install settings and mac os 10.15.7:

![image (12)](https://user-images.githubusercontent.com/9753817/102120599-c7f43000-3e10-11eb-8dcd-268e6aa0ce36.png)

[...]

PR #8959, opened 2020-12-21
Default to lang in catalog when browser preference unavailable

jakemcdermott:
for https://github.com/ansible/awx/issues/8884
##### SUMMARY
When strings aren't available for the browser's preferred locale, default to one with strings to avoid displaying raw javascript
variables.


marshmalien (approved):

jakemcdermott:
e2e :green_circle:

Issue #8962, opened 2020-12-21
Job output page broken when navigating directly by URL


marshmalien:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Job output page displays Not Found error when navigating via URL from one job output page to another job output page with a different job type.

![loading broken](https://user-images.githubusercontent.com/15881645/102817273-7bbf6780-439d-11eb-9b66-d5e2e5f3975c.gif)

##### STEPS TO REPRODUCE
1. Navigate to a playbook run job output page `/jobs/playbook/:jobId/output`
2. Update URL to a project update job output page `/jobs/project/:projectId` (do not include a trailing `/output`)
    * Refreshing the browser removes the error and successfully renders page

##### EXPECTED RESULTS
Job output page successfully renders with no error

##### ACTUAL RESULTS
Error 404 "Not Found" after network request to `/api/v2/project_updates/:jobId/events/?order_by=start_line&page_size=5`

unlikelyzero:
No Unit Tests were added in the PR.

QE Coverage will need to be added for this route and the WFJT routes for completed jobs and jobs which are currently running

one-t:
These items are already stubbed out and present in the backlog.

PR #8963, opened 2020-12-21
Fix navigation bug in job output 

marshmalien:
##### SUMMARY
#8962 

When navigating through the URL, the `type` param and  `job` object would get out of sync and the JobOutput component would reference the "previous" `job`. The error only appears when `componentDidMount` is invoked because it is in that lifecycle method where we attempt to fetch job events with the stale job object. 

loading

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - UI

### DIFF
--- a/awx/ui_next/src/screens/Job/Job.jsx
+++ b/awx/ui_next/src/screens/Job/Job.jsx
@@ -14,15 +14,15 @@ import { CaretLeftIcon } from '@patternfly/react-icons';
 import { Card, PageSection } from '@patternfly/react-core';
 import { JobsAPI } from '../../api';
 import ContentError from '../../components/ContentError';
+import ContentLoading from '../../components/ContentLoading';
 import RoutedTabs from '../../components/RoutedTabs';
 import useRequest from '../../util/useRequest';
 import JobDetail from './JobDetail';
 import JobOutput from './JobOutput';
 import { WorkflowOutput } from './WorkflowOutput';
 import useWsJob from './useWsJob';
-import { JOB_TYPE_URL_SEGMENTS } from '../../constants';
 
-function Job({ i18n, lookup, setBreadcrumb }) {
+function Job({ i18n, setBreadcrumb }) {
   const { id, type } = useParams();
   const match = useRouteMatch();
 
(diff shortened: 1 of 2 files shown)


awx/ui_next/src/screens/Job/Job.jsx
```diff
@@ -14,15 +14,15 @@ import { CaretLeftIcon } from '@patternfly/react-icons';
...
+import ContentLoading from '../../components/ContentLoading';
 import RoutedTabs from '../../components/RoutedTabs';
 import useRequest from '../../util/useRequest';
 import JobDetail from './JobDetail';
 import JobOutput from './JobOutput';
 import { WorkflowOutput } from './WorkflowOutput';
 import useWsJob from './useWsJob';
-import { JOB_TYPE_URL_SEGMENTS } from '../../constants';
 
-function Job({ i18n, lookup, setBreadcrumb }) {
```
marshmalien, line 25: I couldn't find where we were passing the `lookup` prop and it was always `undefined` so I removed it.

awx/ui_next/src/screens/Job/Job.jsx
```diff
@@ -78,20 +83,10 @@ function Job({ i18n, lookup, setBreadcrumb }) {
     );
   }
 
-  if (lookup && job) {
```
marshmalien, line 81: Since `lookup` was always `undefined`, our code never entered this block so I removed this as well. `JobTypeRedirect` component located in the parent seemed to handle this exact redirect logic.

awx/ui_next/src/screens/Job/Jobs.jsx
```diff
@@ -63,11 +54,7 @@ function Jobs({ i18n }) {
           <TypeRedirect view="output" />
         </Route>
         <Route path={`${match.path}/:type/:id`}>
-          <Job
-            history={history}
-            location={location}
```
marshmalien, line 68: We aren't using `history` or `location` in < Job /> so ✂️

jakemcdermott (approved):

TownGeekAus (approved):

jakemcdermott:
e2e :green_circle:

PR #8970, opened 2020-12-22
Convert JobList to PaginatedTable

keithjgrant:
##### SUMMARY
Converts JobList component to use PaginatedTable. I also added the expandable content rows while I was in there

addresses #6189

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

Screen Shot 2020-12-23 at 2 34 21 PM

### DIFF
--- a/awx/ui_next/src/components/DetailList/LaunchedByDetail.jsx
+++ b/awx/ui_next/src/components/DetailList/LaunchedByDetail.jsx
@@ -1,0 +1,51 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import { t } from '@lingui/macro';
+import Detail from './Detail';
+
+const getLaunchedByDetails = ({ summary_fields = {}, related = {} }) => {
+  const {
+    created_by: createdBy,
+    job_template: jobTemplate,
+    schedule,
+  } = summary_fields;
+  const { schedule: relatedSchedule } = related;
+
+  if (!createdBy && !schedule) {
+    return {};
+  }
+
+  let link;
+  let value;
+
+  if (createdBy) {
+    link = `/users/${createdBy.id}`;
+    value = createdBy.username;
+  } else if (relatedSchedule && jobTemplate) {
+    link = `/templates/job_template/${jobTemplate.id}/schedules/${schedule.id}`;
+    value = schedule.name;
+  } else {
+    link = null;
+    value = schedule.name;
+  }
+
+  return { link, value };
+};
+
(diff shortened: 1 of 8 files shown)


nixocio (approved):

jakemcdermott (approved):

tiagodread:
Hey @keithjgrant I've verified that the user can:

- navigate to job details
- relaunch a job
- use the basic and advanced search
- delete a single or some finished jobs
- cancel a single or some running jobs
- sort jobs by name, status, type, start time and finish time 
- change the number of job items on the list
- paginate through job items 
- expand/collapse the row
- The loading state component is displayed

Responsive tests against:
1024x768 and 1920x1080 

**Considerations:**
Are we going to cover these `launched` by and `Project` sorting options?
image

If the user selects a running/pending and a successful job the delete button is enabled and trying to delete both:

image

Are these data `Started Time` and `Finish Time` on the <thead> and `expandable-row`  the same data or they are duplicate?

image

Could you add a data-ouia-component-id (when available) or IDs on:
```
<table> 
<thead> <button> for each sorting option (Name, Status, Start time...)
<tr> expanded with something similar to the original row, like `expanded-job-row-8` 
```

keithjgrant:
> Are we going to cover these `launched` by and `Project` sorting options?

At this point we don't have plans for additional sorting options beyond the column headers, but it's worth revisiting at some point to bring them back in an enhancement (it would take a bit of UI we haven't mocked up or worked through yet)

> If the user selects a running/pending and a successful job the delete button is enabled and trying to delete both:
> 
> image

It looks like this is the current behavior in devel as well. Can you file an issue and we can deal with it separately from this PR?

> Are these data `Started Time` and `Finish Time` on the and `expandable-row` the same data or they are duplicate?
> 
> image

Yeah, these are duplicated from the non-expanded row. Though that's probably not necessary... Thoughts on this @trahman73 ?

> Could you add a data-ouia-component-id (when available) or IDs on:
> 
> ```
> <table> 
> <thead> <button> for each sorting option (Name, Status, Start time...)
> <tr> expanded with something similar to the original row, like `expanded-job-row-8` 
> ```

👍🏻

tiagodread:
Hey @keithjgrant look

Steps to reproduce:
1 click on the checkbox to select all rows

select all

On the checkbox row the aria-label is incorrect, "Select all rows":

image

tiagodread (approved):
Great job, added a PR to fix some broken tests and add more coverage about the list here https://github.com/ansible/tower-qa/pull/5909

PR #8980, opened 2020-12-26
Simplify image build

shanemcd:
Depends-On: https://github.com/ansible/zuul-jobs/pull/46

This does a few things:

- Removes need for awx_sdist_builder image
- Reorders Dockerfile steps to optimize image cache between prod and dev builds
- Unifies VENV_BASE and COLLECTION_BASE in prod and dev builds


.dockerignore
```diff
@@ -1,2 +1 @@
-.git
```
AlanCoding, line 1: Why are you removing this? Using `make docker-compose-build` runs

https://github.com/ansible/awx/blob/642e6f792cad6365506d2ae1f48f2a828b290fb8/Makefile#L583-L584

And this passes `.` for the build context. I agree it would be better to use another folder for the build context, but I don't see anything in this PR that makes that change. Without that, we shouldn't be including folders like this.

Makefile
```diff
@@ -19,7 +19,8 @@ PYCURL_SSL_LIBRARY ?= openssl
 COMPOSE_TAG ?= $(GIT_BRANCH)
 COMPOSE_HOST ?= $(shell hostname)
 
-VENV_BASE ?= /venv
+VENV_BASE ?= /var/lib/awx/venv/
```
AlanCoding, line 22: :+1:  I support this

installer/roles/image_build/templates/Dockerfile.j2
```diff
@@ -1,21 +1,19 @@
...
 
 # Locations - set globally to be used across stages
-ARG VENV_BASE="{% if not build_dev|bool %}/var/lib/awx{% endif %}/venv"
-ARG COLLECTION_BASE="{% if not build_dev|bool %}/var/lib/awx{% endif %}/vendor/awx_ansible_collections"
+ARG COLLECTION_BASE="/var/lib/awx/vendor/awx_ansible_collections"
 
 # Build container
 FROM centos:8 as builder
 
-ARG VENV_BASE
```
AlanCoding, line 16: So `VENV_BASE` becomes _totally_ not customizable.

.dockerignore
shanemcd: One of the things I've done in this PR is "Removes need for awx_sdist_builder image". To do that, We are now building the sdist in the first stage of the image build. A UI target currently calls git. /cc @jakemcdermott is this something we can fix / change?

https://github.com/ansible/awx/blob/642e6f792cad6365506d2ae1f48f2a828b290fb8/installer/roles/image_build/templates/Dockerfile.j2#L79-L82

https://github.com/ansible/awx/blob/642e6f792cad6365506d2ae1f48f2a828b290fb8/Makefile#L484

installer/roles/image_build/templates/Dockerfile.j2
```diff
@@ -72,16 +70,21 @@ RUN cd /tmp && make requirements_collections
...
-RUN mkdir -p -m 755 /var/lib/awx && \
-    OFFICIAL=yes /var/lib/awx/venv/awx/bin/pip install /tmp/{{ awx_sdist_file }}
+# Use the distro provided npm to bootstrap our required version of node
+RUN npm install -g n && n 14.15.1 && dnf remove -y nodejs
+
+# Copy source into builder, build sdist, install it into awx venv
+COPY . /tmp/src/
+WORKDIR /tmp/src/
+RUN make sdist && \
+    /var/lib/awx/venv/awx/bin/pip install dist/awx-$(cat VERSION).tar.gz
```
AlanCoding, line 82: trying to grok the removal of the sdist image.

This line does what that used to do, `make sdist`. This seems fine.

.dockerignore
AlanCoding: that was such an expected answer! Yeah, we could refactor that but I'm also fine leaving it temporarily if it still builds fast enough.

AlanCoding (approved):

shanemcd:
regate

shanemcd:
recheck

shanemcd:
regate

Issue #8847, opened 2020-12-10
[ui_next] Project details view needs sync button


unlikelyzero:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Can only update projects from the list view, but can update inventory sources from the detail view and list view.
https://tower-mockups.testing.ansible.com/patternfly/projects/projects-detail/

##### ENVIRONMENT
* AWX version: X.Y.Z
* AWX install method: openshift, minishift, docker on linux, docker for mac, boot2docker
* Ansible version:  X.Y.Z
* Operating System:
* Web Browser:

kannurkars:
looks like this fix broke job output of the failed jobs, 
Ref: #9117
even tough api/v2/project_updates/110/stdout has the output of the job it says internel server error in job out page in UI.

```
HTTP 500 Internal Server Error
Content-Type: application/json
Vary: Accept

{
    "detail": "A server error has occurred."
}
```

unlikelyzero:
https://github.com/ansible/awx/pull/9003#issuecomment-755752991

PR #9003, opened 2021-01-05
Adds sync button to project details page

AlexSCorey:
##### SUMMARY
Addresses #8847

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
Screen Shot 2021-01-05 at 9 51 30 AM

### DIFF
--- a/awx/ui_next/src/screens/Project/ProjectDetail/ProjectDetail.jsx
+++ b/awx/ui_next/src/screens/Project/ProjectDetail/ProjectDetail.jsx
@@ -19,6 +19,7 @@ import CredentialChip from '../../../components/CredentialChip';
 import { ProjectsAPI } from '../../../api';
 import { toTitleCase } from '../../../util/strings';
 import useRequest, { useDismissableError } from '../../../util/useRequest';
+import ProjectSyncButton from '../shared/ProjectSyncButton';
 
 function ProjectDetail({ project, i18n }) {
   const {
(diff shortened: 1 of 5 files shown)


nixocio (approved):
Those changes address the mentioned issue.

nixocio:
One thing that I noticed reviewing this PR is that the user triggers an action, but it does not get any feedback about that action. Perhaps, we can file another issue to either redirect the user to the main list where we show the progress of the sync or another solution to display to the user progress of the sync.

unlikelyzero:
@AlexSCorey I think we should add something like @nixocio is suggesting. My first thought while using this was to check the console log because I thought that it hadn't completed the operation. If that's necessary in a follow up ticket (subscribing to the websocket event), then we can just file the RFE, too.

We could also show the sync history in some way, to make it clearer that the action was performed

Would it be possible to get a unique id on this element as well?

trahman73:
I think a toast message would fit nicely here to show the user the sync action is completed.
image
image

unlikelyzero:
@AlexSCorey hey, I'm seeing two api requests being sent on the click event. Is it possible to reduce it to a single POST instead of the GET can_update: true -> POST ?

unlikelyzero:
covered by https://github.com/ansible/tower-qa/pull/5832

PR #9009, opened 2021-01-05
consolidate settings and delete dead settings

chrismeyersfsu:
Less duplicate settings. Dev closer to production.


awx/settings/local_settings.py.docker_compose
```diff
@@ -48,56 +48,12 @@ if "pytest" in sys.modules:
         }
     }
 
-# Absolute filesystem path to the directory to host projects (with playbooks).
-# This directory should NOT be web-accessible.
-PROJECTS_ROOT = '/var/lib/awx/projects/'
```
chrismeyersfsu, line 53: Dev and prod store projects here so unify in `defaults.py`

awx/settings/local_settings.py.docker_compose
```diff
@@ -48,56 +48,12 @@ if "pytest" in sys.modules:
...
-
 # Location for cross-development of inventory plugins
 AWX_ANSIBLE_COLLECTIONS_PATHS = '/var/lib/awx/vendor/awx_ansible_collections'
 
-# Absolute filesystem path to the directory for job status stdout
-# This directory should not be web-accessible
-JOBOUTPUT_ROOT = os.path.join(BASE_DIR, 'job_status')
```
chrismeyersfsu, line 60: Dev container didn't use to have `/var/lib/awx` but it does now. So set both prod and dev job otuput to be a subdir in there. Note that the code creates the `job_status` dir if if doesn't exist at job stdout download/generation time.

awx/settings/local_settings.py.docker_compose
```diff
@@ -48,56 +48,12 @@ if "pytest" in sys.modules:
...
-
 # The UUID of the system, for HA.
 SYSTEM_UUID = '00000000-0000-0000-0000-000000000000'
 
-# Local time zone for this installation. Choices can be found here:
```
chrismeyersfsu, line 65: These values are the same as what is in defaults.py

awx/settings/local_settings.py.docker_compose
```diff
@@ -108,40 +64,6 @@ PROXY_IP_WHITELIST = []
 # Enable logging to syslog. Setting level to ERROR captures 500 errors,
 # WARNING also logs 4xx responses.
 
-LOGGING['handlers']['syslog'] = {
```
chrismeyersfsu, line 111: redundant. Same value as `defaults.py`

awx/settings/local_settings.py.docker_compose
```diff
@@ -108,40 +64,6 @@ PROXY_IP_WHITELIST = []
...
-    'level': 'WARNING',
-    'filters': ['require_debug_false'],
-    'class': 'logging.NullHandler',
-    'formatter': 'simple',
-}
-
-LOGGING['loggers']['django.request']['handlers'] = ['console']
```
chrismeyersfsu, line 118: This part _may_ be controversial. From what I could tell, this doesn't change what goes to the console in the dev env. It just stops logs from going to files.

Way back in the day, church used to do his devel outside of a container or even maybe a VM. So it was important to not litter your host machine with log files.

Today, all dev takes place in containers. The log files are rotated so no risk of eating up all the disk space either.

awx/settings/local_settings.py.docker_compose
```diff
@@ -154,74 +76,6 @@ LOGGING['handlers']['management_playbooks'] = {'class': 'logging.NullHandler'}
 #LOGGING['loggers']['django_auth_ldap']['handlers'] = ['console']
 #LOGGING['loggers']['django_auth_ldap']['level'] = 'DEBUG'
 
-###############################################################################
```
chrismeyersfsu, line 157: Dead

awx/settings/defaults.py
```diff
@@ -125,10 +125,10 @@ def IS_TESTING(argv=None):
 # Absolute filesystem path to the directory for job status stdout (default for
 # development and tests, default for production defined in production.py). This
 # directory should not be web-accessible
-JOBOUTPUT_ROOT = os.path.join(BASE_DIR, 'job_output')
+JOBOUTPUT_ROOT = '/var/lib/awx/job_status/'
 
 # Absolute filesystem path to the directory to store logs
-LOG_ROOT = os.path.join(BASE_DIR)
+LOG_ROOT = '/var/log/tower/'
```
chrismeyersfsu, line 131: Absolute path to where logs will live

awx/settings/production.py
```diff
@@ -46,16 +42,6 @@
 
 AWX_ISOLATED_USERNAME = 'awx'
 
-LOGGING['handlers']['tower_warnings']['filename'] = '/var/log/tower/tower.log'  # noqa
```
chrismeyersfsu, line 49: `LOG_ROOT` is now set to `/var/log/tower` instead of dynamically being set to the path where the awx code lives. Therefore, these overrides are not needed. Instead, the `LOGGING['handlers']` settings already defined in defaults.py will be the authority of record. A single place.

awx/settings/local_settings.py.docker_compose
AlanCoding: This probably needs manual verification that the dev container can still produce stdout downloads

awx/settings/local_settings.py.docker_compose
rebeccahhh: given your last point I'm not sure this is actually controversial.

AlanCoding: but then are unit tests and stuff like that going to dump a whole bunch of garbage to these log files?

AlanCoding: This is also done here:

https://github.com/ansible/awx/blob/abc6a842105999bfaadf312ed1fb9596574355bf/installer/roles/image_build/files/settings.py#L55

Sharing code between development and this seems like a good idea, but I don't like the idea of removing this in the name of simplicity. Logging to the console is the doctrine of Kubernetes

rebeccahhh (approved):
this all looks good, I agree with @AlanCoding 's manual test suggestion.

awx/settings/defaults.py
```diff
@@ -125,10 +125,10 @@ def IS_TESTING(argv=None):
...
 
 # The heartbeat file for the tower scheduler
 SCHEDULE_METADATA_LOCATION = os.path.join(BASE_DIR, '.tower_cycle')
```
wenottingham, line 134: Can we change this (and others) and effectively remove BASE_DIR entirely?

AlanCoding: `BASE_DIR` is where the actual AWX code lives. In the development environment, this location (`/awx_devel`) is different from the `/var/lib` locations. I think its naming is poor, but I don't think we can get rid of it easily.

awx/settings/local_settings.py.docker_compose
```diff
@@ -154,74 +76,6 @@ LOGGING['handlers']['management_playbooks'] = {'class': 'logging.NullHandler'}
...
-# SCM TEST SETTINGS
-###############################################################################
-
-# Define these variables to enable more complete testing of project support for
```
ryanpetrello, line 161: 🔪🔪🔪

ryanpetrello (approved):
Man, doing some housecleaning in these files was a _long_ time coming - thank you 😄.

Issue #8921, opened 2020-12-16
Prevent users from selecting credentials that prompt for passwords on workflow nodes


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
This is residual work from #8399.  If a Job Template references credentials that prompt for passwords then that template cannot be used in a workflow.  Additionally, if the JT prompts for credentials a user should not be able to select a credential that prompts for password(s).  Attempting to save this node will result in an error from the api.

![image (1)](https://user-images.githubusercontent.com/9889020/102362689-b16de600-3f82-11eb-9a23-2b06706a787f.png)

AlexSCorey:
We need to do something similar for schedules

tiagodread:
scenario from https://github.com/ansible/awx/pull/9210

- The credential password prompt is not being required on the schedule modal wizard:
1. Create a machine credential OR a vault credential with prompt enabled for password
2. Create a JT and mark Credentials as prompt
2. Create a schedule for this JT
3. On the Credentials step choose the credential created before (with prompt)

The step `Credential passwords` step to fill the credential password isn't displayed: 

![schedules7](https://user-images.githubusercontent.com/12464858/107773720-d54c6e80-6d1c-11eb-90bf-03099846dfdf.gif)

mabashian:
Yea @tiagodread the credential password step shouldn't be shown in that case because the user shouldn't be able to select a credential that prompts for passwords on a schedule/workflow node.  We'll make sure to address that when working this issue.

mabashian:
The UX in the old UI is pretty poor here:

<img width="1664" alt="Screen Shot 2021-02-24 at 1 13 34 PM" src="https://user-images.githubusercontent.com/9889020/109045918-29067280-76a2-11eb-962c-21fdf30d9ed0.png">

The row is selectable but the button stays disabled giving no indication as to why.

I think we should disable the row and add a tooltip explaining that JT's that have credentials that prompt for passwords are not permitted in workflow nodes.

mabashian:
Note that the described case above is slightly different but related to credential password prompting.  We should fix both.

mabashian:
Here's what we need to do here:

1) Prevent users from _selecting_ credentials that prompt for passwords on workflow nodes/schedules where credential is promptable
2) If a JT prompts for credentials and a default credential prompts for passwords then the credential must be replaced with one that does not prompt for password(s) in the schedule/wf node before it can be saved.
[...]

PR #8738, opened 2020-12-03
Adds vault IDs to Vault credential chips

AlexSCorey:
##### SUMMARY
This addresses #8532 This adds the vault ID to the vault credential name.  I show it the same way that it's shown in the old ui.  However, since we display the id each time we display the credential name I wonder if we should just build the name with the ID when the credential is created and save it that way. That would potentially save several api calls we need to make the get the vault id and it would eliminate places where we build the name in the ui....on `MultiCredentialsLookup.jsx` and `CredentialChip.jsx`.

There are a few places where this hasn't been implemented because those parts are actively in flight right now. https://github.com/ansible/awx/issues/8735

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
Screen Shot 2020-12-03 at 10 29 11 AM

PR #9028, opened 2021-01-06
Add support for credential password prompting on job launch

mabashian:
##### SUMMARY
link #7015 

The main problem that I ran in to while adding support for this feature is the fact that each time formik's validation runs, errors for fields that are not shown get wiped.  To get around this, I added a method to each step that will re-apply errors whenever a user navigates to a particular step.  Each step is also responsible for determining whether or not that step has an error.

The scope of these changes will require us to re-test all of the launch prompting scenarios.  As far as credentials go, there are 3 scenarios where credential passwords.

1) A JT can have default credentials that require passwords on launch but the JT itself doesn't prompt for credentials.
2) A JT can have default credentials that require passwords on launch and the JT can also prompt for credentials.  This allows users to select different credentials.
3) A JT can not have default credentials but prompt for credentials.  This would allow a user to select credentials that prompt for passwords.

Note that this issue: https://github.com/ansible/awx/issues/8200 is still not implemented.

https://github.com/ansible/awx/issues/8921 is also not implemented.

Both of these issues may be encountered during testing.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI


AlexSCorey:
@mabashian Screen Shot 2021-01-07 at 1 52 36 PM
Should we somehow indicate which credential needs the password? ie. should we show the credential name?

awx/ui_next/src/components/LaunchPrompt/steps/CredentialPasswordsStep.jsx
```diff
@@ -0,0 +1,129 @@
...
+        />
+      )}
+      {vaultsThatPrompt.map(credId => (
+        <PasswordField
+          id={`launch-vault-password-${credId}`}
+          key={credId}
+          label={
+            credId === ''
+              ? i18n._(t`Vault password`)
+              : i18n._(t`Vault password - ${credId}`)
```
AlexSCorey: In this PR https://github.com/ansible/awx/pull/8738 we are using `|` to separate the credential name from the vault ID.  I realize this is not exactly the same use case/situation, but thought I'd share.  We may want to have some consistency with the naming pattern with these credentials.

awx/ui_next/src/components/LaunchPrompt/steps/useCredentialPasswordsStep.jsx
```diff
@@ -0,0 +1,342 @@
...
+    return initialValues;
+  }
+
+  if (
+    !launchConfig.ask_credential_on_launch &&
+    launchConfig.passwords_needed_to_start
+  ) {
+    launchConfig.passwords_needed_to_start.forEach(password => {
+      if (password === 'ssh_password') {
+        initialValues.credentialPasswordSsh = '';
```
AlexSCorey: Why are we added they keys here, only to switch them back to keys that the api recognizes in https://github.com/ansible/awx/pull/9028/files#diff-b4b7e4b3882a0650f7db64d9166a39cc1c7ecae225a73f41efaaa9bc0334286fR1?  It seems that if we didn't switch keys a lot of the validation could be consolidated into a forEach loop making this code more concise and easier to read.

awx/ui_next/src/util/prompt/getCredentialPasswords.js
```diff
@@ -0,0 +1,29 @@
+export default function getCredentialPasswords(values) {
```
AlexSCorey: This could possibly be shortened or maybe even eliminated if we can avoid created keys in the formik values object that the api doesn't recognize.

AlexSCorey (commented):
My gut tells me that the validation on `useCredentialPasswordStep` can be improved.  Let me know if you want to discuss.

tiagodread:
@mabashian @one-t 

- Trying to relaunch a JT with credentials prompts enabled and JT credentials prompt enabled:
image

Full Flow:
flowprompt

Note that, when I launch the JT for the first time, the `Private key passphrase *` is required on wizard but on relaunch this field isn't displayed.


- Should this icon on step 2 (with error)  be at the same line with the text?

image

tiagodread (approved):
Verified that:
- POL regression
- The user can prompt credentials without POL on credential
- The user can prompt credentials with POL on credential
- The user can prompt using different credentials like machine and vault
- The user can relaunch a job and prompt credentials fields

PR with stubs and a test https://github.com/ansible/tower-qa/pull/5871

AlexSCorey (approved):
I think these changes are an improvement to reduce some of the complexity in the validation function.  I think this is a big step to get to feature parity for all of the prompt on launch features.  Well done!

PR #9046, opened 2021-01-08
Do not drop inventories marked as has_inventory_sources

jbradberry:
##### SUMMARY

The old tool did not do this aggressive filtering.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - CLI

##### AWX VERSION
```
awx: 16.0.0
```


natman (commented):
It does the job, but it should be good to have an option to export only sources without a long list of hosts that can be synced in a second time on the host destination.

ryanpetrello (approved):

PR #9054, opened 2021-01-11
Pin cryptography to pre 3 release

Spredzy:
Pin cryptography to pre `2.9.2` release. The recent autobahn version allows it.

### DIFF
--- a/requirements/README.md
+++ b/requirements/README.md
@@ -140,6 +140,10 @@ The offline installer needs to have functionality confirmed before upgrading the
 Versions need to match the versions used in the pip bootstrapping step
 in the top-level Makefile.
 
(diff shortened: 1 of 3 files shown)


AlanCoding (approved):

AlanCoding (approved):

chrismeyersfsu:
We need to put an entry in requirements.in This way, if a y or z stream is released (security updates) we pick it up. Also, by just living in requirements.txt it has a chance to be overridden by the 3.x version

Edit: You beat me to it :)

chrismeyersfsu (approved):

rooftopcellist (approved):

Issue #9049, opened 2021-01-08
Table column rows break from toolbar when in responsive resolutions between 600 < x < 1024


unlikelyzero:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Table column rows break from toolbar when in responsive resolutions between 600 < x < 1024

##### ENVIRONMENT
* AWX version: aa87a99441b595c3760a0950968b64a664382bfd

##### STEPS TO REPRODUCE
1. Open up the inventories list view
2. Narrow the width of the browser window between 600 and 1024 pixels

##### EXPECTED RESULTS

Table to match toolbar width

##### ACTUAL RESULTS

Table width recalculated and out of alignment with toolbar

##### ADDITIONAL INFORMATION

<img width="1327" alt="Screen Shot 2021-01-08 at 3 37 35 PM" src="https://user-images.githubusercontent.com/12446869/104062321-3f2aa300-51c8-11eb-9c5c-e876c72719dd.png">

keithjgrant:
after discussion with @trahman73, we decided to hide the Groups, Hosts, and Sources columns for viewports that are too narrow to accommodate them. This is an approach that some other teams are taking in similar circumstances, and since we have the details view to see this data, it's simply a nice-to-have in the table view.

tiagodread:
worked as expected, we have already a visual test to cover this list in /visualintegration/inventory.spec.js

PR #9057, opened 2021-01-11
Fix inventory list responsive behavior

keithjgrant:
##### SUMMARY

Hides Groups, Hosts, and Sources columns in Inventory List table when screen width is too narrow to show them.

Addresses #9049 

Screen Shot 2021-01-11 at 9 49 04 AM

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/components/PaginatedTable/HeaderRow.jsx
+++ b/awx/ui_next/src/components/PaginatedTable/HeaderRow.jsx
@@ -71,6 +71,7 @@ export function HeaderCell({
   sortBy,
   columnIndex,
   idPrefix,
+  className,
   children,
 }) {
   const sort = sortKey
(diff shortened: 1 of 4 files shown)


nixocio (approved):

jlmitch5 (approved):
nice

tiagodread:
Add a visual test to cover this one

tiagodread:
Hey @keithjgrant from 993px to 1081 we are not hiding these columns and the problem was not fixed, look:

resize

keithjgrant:
> Hey @keithjgrant from 993px to 1081 we are not hiding these columns and the problem was not fixed, look:

Yeah, it looks like this can vary a lot depending on the length of words in the Inventory and Organization names. I was able to make on that broke even at 1300px wide. I'm beginning to think the best answer might be to remove Groups/Hosts/Sources columns altogether. (allowing word-breaks doesn't really help because the browser seems to layout the table before making those decisions).

tiagodread:
@keithjgrant We are hiding Groups, Hosts and Sources with < 993px from 993px to 1057px the user can see this problem, in addition, the sidebar menu also is pushing the page to the right:

resize

tiagodread:
@keithjgrant from 1200px (when we hide the sidebar menu) to 1220px the user still can see this problem:

resize

tiagodread (approved):
worked as expected, we have already a visual test to cover this list in `/visualintegration/inventory.spec.js`

PR #9059, opened 2021-01-12
Update inventory

Gr4cchus:
##### SUMMARY

In the docs it says `If these variables are present then all deployments will use these hosted images. If the variables are not present then the images will be built during the install.` and lists these as the variables:
```
dockerhub_base=ansible
dockerhub_version=latest
```
shouldnt that second line also be included here? Or if its implicit that it uses `latest` comment it somewhere in the config so the user knows this?

##### ISSUE TYPE

 - Docs Pull Request

##### COMPONENT NAME

 - API
 - UI
 - Installer

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/installer/inventory
+++ b/installer/inventory
@@ -7,6 +7,7 @@ localhost ansible_connection=local ansible_python_interpreter="/usr/bin/env pyth
 # be selected based on: latest, 1, 1.0, 1.0.0, 1.0.0.123
 # by default the base will be used to search for ansible/awx
 dockerhub_base=ansible
+dockerhub_version=latest
 
 # Openshift Install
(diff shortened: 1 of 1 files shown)


ryanpetrello:
@Gr4cchus sorry for the delay on reviewing/merging this PR - we've since replaced the local Docker installer in `devel` (https://groups.google.com/g/awx-project/c/47MjWSUQaOc/m/bCjSDn0eBQAJ https://github.com/ansible/awx/tree/devel/tools/docker-compose) and thus the next major release of AWX, so I suspect this PR is obsolete now.

PR #9069, opened 2021-01-12
remove python log rotation in favor of system

chrismeyersfsu:
* The cron ran logrotate will now rotate our log files instead of python


ryanpetrello (approved):
caveats about verification, testing, etc... but yes this makes general sense

We should make sure we actually test this in a live system that passes over multiple log rotation boundaries, and make sure all underlying systems are okay with it

It might even be worth doing some general chaos engineering and setting up a build where we periodically just fully wipe out all awx-specific logs under `/var/log/tower/` and make sure things recover in a reasonable way

chrismeyersfsu:
recheck

Issue #9077, opened 2021-01-13
Partial Migrations on Local Docker Install


rooftopcellist:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

I seeing migrations only partially run on Local Docker installations.  It seems as those sometimes, the migrations will run in the background and finish, but other times, only 5 migrations are run.  

##### ENVIRONMENT
* AWX version: X.Y.Z
* AWX install method: openshift, minishift, docker on linux, docker for mac, boot2docker
* Ansible version:  X.Y.Z
* Operating System:
* Web Browser:

##### STEPS TO REPRODUCE

```
ansible-playbook -i inventory install.yml -vv
```

##### EXPECTED RESULTS

Successful install, no errors.  

##### ACTUAL RESULTS

Only 5 migrations run (none from the `main` app), installer fails early on "Create Preload data" task.  

##### ADDITIONAL INFORMATION

Traceback from installer:

```
TASK [local_docker : Create Preload data] **************************************************************************************
task path: /home/chadams/awx/installer/roles/local_docker/tasks/compose.yml:53
[...]

PR #9078, opened 2021-01-13
Fix Local Docker partial migration and unavailable superuser issues

rooftopcellist:
##### SUMMARY
Fixes migration issues mentioned in https://github.com/ansible/awx/issues/9077

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
devel
```

### DIFF
--- a/installer/roles/local_docker/tasks/compose.yml
+++ b/installer/roles/local_docker/tasks/compose.yml
@@ -36,6 +36,11 @@
   register: awx_secret_key
 
 - block:
+    - name: Run migrations in task container
+      shell: docker-compose run --rm --service-ports task awx-manage migrate --no-input
+      args:
+        chdir: "{{ docker_compose_dir }}"
+
     - name: Start the containers
       docker_compose:
         project_src: "{{ docker_compose_dir }}"
@@ -49,6 +54,11 @@
     - name: Update CA trust in awx_task container
       command: docker exec awx_task '/usr/bin/update-ca-trust'
       when: awx_compose_config.changed or awx_compose_start.changed
+
+    - name: Wait for launch script to create user
+      wait_for:
+        timeout: 10
+      delegate_to: localhost
 
     - name: Create Preload data
       command: docker exec awx_task bash -c "/usr/bin/awx-manage create_preload_data"
--- a/installer/roles/local_docker/tasks/upgrade_postgres.yml
+++ b/installer/roles/local_docker/tasks/upgrade_postgres.yml
@@ -1,26 +1,24 @@
 ---
-- name: Check for existing Postgres data
-  stat:
-    path: "{{ postgres_data_dir }}/pgdata/PG_VERSION"
+
+- name: Register temporary docker container
+  set_fact:
+    container_command: "docker run -v '{{ postgres_data_dir | realpath }}:/var/lib/postgresql' centos:8 bash "
+
+- name: Check for existing Postgres data (run from inside the container for access to file)
+  shell:
+    cmd: "{{ container_command }} [[ -f /var/lib/postgresql/10/data/PG_VERSION ]] && echo 'exists'"
   register: pg_version_file
+  ignore_errors: true
 
 - name: Record Postgres version
-  set_fact:
-    old_pg_version: "{{ lookup('file', postgres_data_dir + '/pgdata/PG_VERSION') }}"
-  when: pg_version_file.stat.exists
+  shell: "{{ container_command }} cat var/lib/postgresql/10/data/PG_VERSION"
+  register: old_pg_version
+  when: pg_version_file.stdout == 'exists'
 
 - name: Determine whether to upgrade postgres
   set_fact:
-    upgrade_postgres: "{{ old_pg_version is defined and old_pg_version == '9.6' }}"
-
-- name: Set up new postgres paths pre-upgrade
-  file:
-    state: directory
-    path: "{{ item }}"
-    recurse: true
-  when: upgrade_postgres | bool
-  with_items:
-    - "{{ postgres_data_dir }}/10/data"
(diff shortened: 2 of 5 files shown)


installer/roles/local_docker/tasks/compose.yml
```diff
@@ -50,6 +56,11 @@
       command: docker exec awx_task '/usr/bin/update-ca-trust'
       when: awx_compose_config.changed or awx_compose_start.changed
 
+    - name: Wait for launch script to create user
+      wait_for:
+        timeout: 5
```
rooftopcellist: Fixes Nonetype superuser [issue](https://github.com/ansible/awx/issues/9077#issuecomment-759749936)

installer/roles/local_docker/tasks/upgrade_postgres.yml
```diff
@@ -1,26 +1,31 @@
...
-  stat:
-    path: "{{ postgres_data_dir }}/pgdata/PG_VERSION"
+
+- name: Register full path to postgres data dir
+  shell: "echo {{ postgres_data_dir }}"
+  register: fq_postgres_data_dir
+
+- name: Register temporary docker container
+  set_fact:
+    container_command: "docker run -v '{{ fq_postgres_data_dir.stdout }}:/var/lib/postgresql' centos:8 bash "
```
jbradberry: See also: https://github.com/ansible/awx/pull/9065#discussion_r556864979

installer/roles/local_docker/tasks/upgrade_postgres.yml
```diff
@@ -1,26 +1,31 @@
...
+
+- name: Check for existing Postgres data (run from inside the container for access to file)
+  shell:
+    cmd: "{{ container_command }} [[ -f /var/lib/postgresql/10/data/PG_VERSION ]] && echo 'exists'"
```
jbradberry, line 9: I think that using https://docs.ansible.com/ansible/devel/collections/ansible/builtin/stat_module.html would be a more idiomatic Ansible thing to do here.

awx/main/management/commands/create_preload_data.py
```diff
@@ -10,7 +10,7 @@
 class Command(BaseCommand):
     """Create preloaded data, intended for new installs
     """
-    help = 'Creates a preload tower data iff there is none.'
+    help = 'Creates a preload tower data if there is none.'
```
ryanpetrello: I dunno who wrote this originally, but I'll bet this wasn't a typo 😉

https://en.wikipedia.org/wiki/If_and_only_if#Origin_of_iff_and_pronunciation

ryanpetrello: My suspicions are confirmed, @matburt!

https://github.com/ansible/awx/commit/6709654e0471c49c13f8d448ee9162087551984a

rooftopcellist: aha! good point.  thanks for pointing that out.

installer/roles/local_docker/tasks/upgrade_postgres.yml
rooftopcellist: @jbradberry is there a good way to stat this file from inside the container with ansible?  The issue is that we can't run this from the host without sudo otherwise.

chrismeyersfsu: ansible has a docker connection module https://docs.ansible.com/ansible/latest/scenario_guides/guide_docker.html#connecting-to-the-docker-api

rooftopcellist: I attempted to do this here: https://github.com/rooftopcellist/awx/compare/pg12-upgrade-migrations...rooftopcellist:pg12-upgrade-migrations-docker?expand=1

But ran into 2 issues:
1. If one of these tasks fail, the installer will leave behind a tmp container which will cause errors in later runs.  (Not that we couldn't solve this one, it would just make the playbook messy.  
2. I am running in to permission (?) issues when using this method.  
```
TASK [local_docker : Check for existing Postgres data (run from inside the container for access to file)] ************************************************************************************************************************************
task path: /home/chadams/awx/installer/roles/local_docker/tasks/upgrade_postgres.yml:13
fatal: [localhost]: UNREACHABLE! => {"changed": false, "msg": "Failed to create temporary directory.In some cases, you may have been able to authenticate and did not have permissions on the target directory. Consider changing the remote tmp path in ansible.cfg to a path rooted in \"/tmp\", for more error information use -vvv. Failed command was: ( umask 77 && mkdir -p \"` echo ~/.ansible/tmp `\"&& mkdir \"` echo ~/.ansible/tmp/ansible-tmp-1610982431.4560068-986689-254466671929443 `\" && echo ansible-tmp-1610982431.4560068-986689-254466671929443=\"` echo ~/.ansible/tmp/ansible-tmp-1610982431.4560068-986689-254466671929443 `\" ), exited with result 1", "unreachable": true}
```

I think we should merge what is currently in the PR.

shanemcd (approved):

installer/roles/local_docker/tasks/upgrade_postgres.yml
```diff
@@ -1,26 +1,24 @@
 ---
-- name: Check for existing Postgres data
-  stat:
-    path: "{{ postgres_data_dir }}/pgdata/PG_VERSION"
+
+- name: Register temporary docker container
+  set_fact:
+    container_command: "docker run -v '{{ postgres_data_dir | realpath }}:/var/lib/postgresql' centos:8 bash "
```
rooftopcellist, line 5: Using `realpath` here is not what we need.  

```
ok: [localhost] => {"ansible_facts": {"container_command": "docker run -v '/home/chadams/awx/installer/~/.awx/pgdocker:/var/lib/postgresql' centos:8 bash "}, "changed": false}
```

Issue #8465, opened 2020-10-27
Page title with different text size across ui_next


tiagodread:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

The dashboard and Management Jobs has a different text size <h2> while in other place usingn <h1>

##### ENVIRONMENT
* AWX version: 3.8.0
* AWX install method: docker for mac
* Ansible version:  2.9.13
* Operating System: Catalina
* Web Browser: Chrome

##### STEPS TO REPRODUCE

1. Navigate to dashboard or Management jobs

##### EXPECTED RESULTS

Page title should display the title using the same pattern for all places

##### ACTUAL RESULTS

Different tags in some pages

##### ADDITIONAL INFORMATION

![image](https://user-images.githubusercontent.com/12464858/97308146-4be16100-183f-11eb-9785-dc2938026d97.png)

![image](https://user-images.githubusercontent.com/12464858/97308170-53086f00-183f-11eb-97f5-d6d2b9ee1096.png)

unlikelyzero:
@tiagodread does this violate an a11y standard?

tiagodread:
> @tiagodread does this violate an a11y standard?

Not necessarily but we need keep the consistency between the importance of that information for the user (reading or listening)

tiagodread:
Closing this after merge #8482

jakemcdermott:
reopening - in the above screenshots, we actually want the font size shown in the dashboard image for all of the title bars.

jlmitch5:
This is what we want it to look like:

https://www.patternfly.org/v4/demos/primary-detail/react-demos/primary-detail-content-padding/

So for example

```
Organizations > Ansible (two breadcrum links)
Details (on no line, title component)
```

mabashian:
I believe that @jlmitch5 addressed this with #9083

tiagodread:
Yeah, we can close this issue.

Issue #7727, opened 2020-07-27
Add title under breadcrumb


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
<img width="1460" alt="Screen Shot 2020-07-27 at 1 11 29 PM" src="https://user-images.githubusercontent.com/9889020/88570850-c5806d80-d00a-11ea-8960-15e0861248f8.png">

We used to have something like this supported in the breadcrumb component but I think that support was dropped.  If we want to do something like this now we'll have to add it ourselves.

Will be important to work with @trahman73 to determine when/where to add this.

mabashian:
We're now simply following the PF pattern for breadcrumbs.  Closing

unlikelyzero:
https://github.com/ansible/awx/pull/9083#issuecomment-765005073

Issue #7230, opened 2020-06-03
Allow search for actions performed by user in activity stream


davidtluong:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Right now there's no way to search for actions performed by a certain user in activity stream.  The request is for there to be a field that you can search for by username and return the actions performed by the user.

wenottingham:
You can tack `actor__username:<username>` into the URL bar and this will work.

This should be fixed in the UI rewrite to be supported in the search widget.

mabashian:
This was addressed in #9083 

<img width="1503" alt="Screen Shot 2021-02-04 at 11 27 04 AM" src="https://user-images.githubusercontent.com/9889020/106923341-f50be300-66db-11eb-8b1b-077d33fa2a7f.png">

Issue #6866, opened 2020-04-28
Activity Stream


unlikelyzero:
Stub for development

Note: We should rethink default search filtering. 'admin' doesn't return what a user would think to see, for example.

jlmitch5:
Things that need new activity stream entries (API and UI): - 
-  (no issue yet) - possibly need to add as entries for execution environments api model(s) (and display in UI)?
- https://github.com/ansible/awx/issues/5011 Ad Hoc Commands as entries links (need to be added to api and UI)

Outstanding UI-only issues:
- https://github.com/ansible/awx/issues/3521 - buggy workflow node deletion as entry link
- https://github.com/ansible/awx/issues/5030 - buggy workflow schedule as entry link
- https://github.com/ansible/awx/issues/5299 - line wrap broken in as detail modal “action” detail item
- https://github.com/ansible/awx/issues/3786 - activity stream column spacing is poor  

Translation (both UI and API):
- https://github.com/ansible/awx/issues/3828
  -  On the UI side: a few interstitial connecting words and the pastTense helper needs to be translated (in old code base this is the build-description.factory.js file).
  -  On the api side: the “object1” and “object2” would need to be translated, or the UI would not to get all the possible object1/object2 values so we can  hold the translations

  Outstanding API-only Issues:
-  https://github.com/ansible/awx/issues/1332 - extraneous activity stream entries on initial startup
- https://github.com/ansible/awx/issues/1513 - related job as entries don’t have as much info in them as regular jobs
- https://github.com/ansible/awx/issues/2535 - extraneous notification template as entries
- https://github.com/ansible/awx/issues/2828 - start/end time of job’s activity stream entries not included
- https://github.com/ansible/awx/issues/3693 - canceled_by not included in as entries

rooftopcellist:
@jlmitch5 for https://github.com/ansible/awx/issues/3828, that requires marking those strings for translation before the go in the db (saved on the activity stream object) as well as when they are pulled from the db.  

Here is another example of where we had to do this: https://github.com/ansible/awx/pull/7045/files

Though performance may be a consideration here because there are so many activity stream objects.  

--------------------------------------------------------

Here is a quick POC for the API side of that issue - https://github.com/ansible/awx/compare/devel...rooftopcellist:activity_stream_translations?expand=1

mabashian:
Some notes: 

[...]

PR #9083, opened 2021-01-14
Update Breadcrumbs/Add Activity Stream UI

jlmitch5:
This PR is broken up into 3 main commits. Some detail on each:

TODO:
- Adding some more detailed tests around the description/link building

## "update Breadcrumb component to ScreenHeader"

Relevant issues: https://github.com/ansible/awx/issues/7727, https://github.com/ansible/awx/issues/8465

Screen Shot 2021-01-14 at 10 36 38 AM

Part of the requirements for the Activity Stream UI (that is pulled over from the legacy UI), is that each view (aside from Managment Jobs), has an activity stream link which pre-filters the view to the specific type of resource(s) that page shows.  So for example, when you are on the Organizations page and you click the Activity Stream link in the top-right header, it should take you to the Activity Stream route and only show you Organization-based activity (creating, editing, deleting, etc.)

Because we already had a component that rendered the header across the app in a consistent way (`<Breadcrumbs />`, I decided to extend that component (with an additional prop for the stream type to filter), and rename it to `<ScreenHeader />` in order for the name to better encapsulate its new responsibilities.

While making these changes, I decided to tackle an issue that was raised a few weeks ago...Patternfly has stopped rendering the `BreadcrumbHeading` component (always our last breadcrumb) in a similar way to the `Title` component.  I talked with taufique, and he said we still want this functionality--projects are moving to a system where this functionality is implemented by rolling your own solution with `Title` for the last breadcrumb.  Basically, I just duplicated the recursive Crumb component, and only render the final step (as `Title`).  I also get rid of the `Breadcrumbs` bar (and the unnecessary top padding it creates) when there is only one crumb.

For testing, due to the way this utilizes the `useRouteMatch` hook, the solution I settled on was to utilize the actual React Router lib in the relevant files as they would fail if I didn't (open to other approaches on this point if anyone has any better ideas).

## "don't strip out non-namespaced params when encoding url search params"

When designing the qs libs, we made sure all lists had a namespace with which we could add to the params, so that when interacting with various UI elements which update the list (such as search, sort and pagination/page size setting), any other search params that were hanging around would not be stripped out.  while adding the params was done correctly, it seemed like the encoding side of things was not.  I updated the qs lib's function for encoding the params into the url param to accept an addtional "other search params" argument, and in all the places where it is called, I get those other params so they can be passed.

This was necessary because the activity stream type needed to be a search param, but it kept getting stripped out and set back to "all activity".

## "add activity stream ui"

Relevant Issues:
- Root activity stream UI issue https://github.com/ansible/awx/issues/6866
- Add `?actor__username` search column: https://github.com/ansible/awx/issues/7230
- Add workflow template schedule link: https://github.com/ansible/awx/issues/5030
- Table column responsiveness: https://github.com/ansible/awx/issues/3786

Screen Shot 2021-01-14 at 10 37 11 AM
Screen Shot 2021-01-14 at 10 37 25 AM

This commit contains all the new activity stream UI code (api model, route, list, etc.).  It uses the new react-table component. 
 It is virtually identical to the old UI, with a few small differences which are:
- there are a few additional/updated activity stream types in the dropdown to cover all screens in ui_next.  oauth2 applications and tokens have been consolidated to a single type so that they can be linked to together in the applications routes.
- anchor links have been updated in a few places to match the equivalent routes in the new UI.  The most notable change in this respect is that settings-related events used to link to "sub-sections", but because these sub-sections are defined in the UI (as opposed to the small number of categories the api defines them under), I just updated these links to the root settings page.
- the description/link string builder is virtually identical to that in the old UI (which I did so that functionality would be identical), but has been updated to conform to our coding standards (no more `string +=`).  It also does not render html strings (which is against our new CSP), but rather JSX components.
- sorting can be done by timestamp and initiated by (by userrname), but not the description column.  The description column's params that were used in the old UI were actually not the same ones that are displayed, so they didn't actually work correctly.  There's no real way to sort this way because the descrriptions are generated from a bunch of different pieces of data from the api.  Note for initiated by "system" (even usernames that come alphabetically after it) will always be at the end, because it is written when no actor__username is given by the api.
- Note that "keyword search" maps to ?search= on the backend.  It can return puzzling results, but after talking with Taufique it made more sense to leave it (to maintain consistent functionality across old and new UI).  It would be nice in the future to see if there are ways to retool the activity stream's API view to return the description and initiated by data in a way that is more accurately searchable.


unlikelyzero:
@qe we should add a visual test here https://github.com/ansible/awx/issues/3786#issuecomment-760354886

tiagodread:
Hey @jlmitch5 this looks awesome, I've found so far:

- Performing a search and then changing the activity stream type, the search term still set:

as1

- Missing space between text and link:

image


- Paginate through the list and then select an Activity Stream type that doesn't contain events, the `Not Found` view is displayed instead of `No Events Found` view.

image

- Trying to perform an advanced search with `Key` set to Type:

image

- Team -> Roles Tab missing title:

image

- JT -> Schedules tab -> Add schedule -> Schedule details missing title:

image

- Inventories -> groups tab -> Add group -> group details -> Related Groups tab only appear "Groups" as title AND navigating between the tabs Related Groups and Hosts you can see that the UI has a resizing (updown) weird problem 

as2

- Inventories -> inventory -> Sources Tab -> Add Source -> Source notification tab missing title:

image

jakemcdermott (approved):

jlmitch5:
Thanks for the review @tiagodread !

> Performing a search and then changing the activity stream type, the search term still set:

Fixed

> Missing space between text and link:

Fixed

> Paginate through the list and then select an Activity Stream type that doesn't contain events, the `Not Found` view is displayed instead of `No Events Found` view.

I've reset the page to 1 when the activity stream type changes

> Trying to perform an advanced search with `Key` set to Type:

We talked about this offline.  It's due to the fact type doesn't accept the __icontains suffix operator.

jlmitch5:
@tiagodread fixed all the breadcrumb issues you mentioned.  Thanks again for such a thorough review!

- Team -> Roles Tab missing title:

Fixed

- JT -> Schedules tab -> Add schedule -> Schedule details missing title:

Fixed

- Inventories -> groups tab -> Add group -> group details -> Related Groups tab only appear "Groups" as title

Fixed

- navigating between the tabs Related Groups and Hosts you can see that the UI has a resizing (updown) weird problem 

Fixed (and I added a min-height to the title section, so even if it's not rendered yet, it's render want cause updown flashing in any case)

- Inventories -> inventory -> Sources Tab -> Add Source -> Source notification tab missing title:

Fixed

unlikelyzero:
@jlmitch5 @marshmalien  added https://github.com/ansible/awx/issues/9122 to track the settings work

unlikelyzero:
covered by https://github.com/ansible/tower-qa/pull/5873

Issue #9085, opened 2021-01-14
Certain playbook output has unexpected delays


ryanpetrello:
Run this playbook on an awx install that is doing little or nothing else:

```yaml
---

- hosts: all
  gather_facts: false
  tasks:
    - shell: 'echo "Hello, World" && sleep .1'
      with_sequence: start=0 end=100
```

...observe that none of the events are recorded, nor does any of the stdout show up in the UI *until* the playbook command finishes executing (then everything shows up at the end in a giant burst).

kdelee:
fwiw I started to look at this in my local dev environment

I noticed that running something like https://github.com/ansible/test-playbooks/blob/master/chatty_tasks.yml with 1 host and 5000 messages, I did not get any output until the end of the playbook run.

Tried blowing away all my containers and starting fresh, but now I'm having trouble getting ti to run at all (stuck at `Attatching to tools_awx_1`).

will continue tomorrow

kdelee:
@ryanpetrello I'm still seeing some weird behavior when I try also on an ec2 instance

here's a gif trying to capture what I'm seeing.

Essentially a little bit of stdout will show when launching the job, but then it just does not update until I do a hard refresh

![events_not_streaming](https://user-images.githubusercontent.com/29239279/105882457-a7a0bf00-5fd3-11eb-80f6-316e40caa54b.gif)

not saying your change was not good -- but does not seem sufficient to say that there is not any other "unexpected delays"

ryanpetrello:
@kdelee for what it's worth, the new UI may make this hard to troubleshoot/verify - @jakemcdermott suggested to me that there may still be some streaming stdout warts?

@kdelee if you open the websocket tab in a browser console, are you able to see messages come in?

kdelee:
not that I can tell, but maybe I'm not reading this right.

Also worth mentioning that the "elapsed" timer does not update until the end of the job either

Here is a video. Wait for it -- it looks static but thats because nothing updates until after the job is done (is about 1 min and change)
![events_delayed](https://user-images.githubusercontent.com/29239279/106165489-2a9f5200-6159-11eb-81f3-1d73d3cc6f5b.gif)

kdelee:
We're still seeing this

ryanpetrello:
@kdelee I wonder if this is similar to https://github.com/ansible/awx/issues/9747

kdelee:
It is hard to fully determine if this is fixed until the websocket problems are fixed.

PR #9086, opened 2021-01-14
further optimize callback receiver buffering for certain situations

ryanpetrello:
see: https://github.com/ansible/awx/issues/9085

I'm still tinkering with this a bit (I think I'm responsible for the change that busted it)

### DIFF
--- a/awx/main/dispatch/worker/callback.py
+++ b/awx/main/dispatch/worker/callback.py
@@ -38,6 +38,7 @@ class CallbackBrokerWorker(BaseWorker):
 
     MAX_RETRIES = 2
     last_stats = time.time()
+    last_flush = time.time()
     total = 0
     last_event = ''
     prof = None
@@ -52,7 +53,7 @@ class CallbackBrokerWorker(BaseWorker):
 
     def read(self, queue):
         try:
-            res = self.redis.blpop(settings.CALLBACK_QUEUE, timeout=settings.JOB_EVENT_BUFFER_SECONDS)
+            res = self.redis.blpop(settings.CALLBACK_QUEUE, timeout=1)
             if res is None:
                 return {'event': 'FLUSH'}
             self.total += 1
@@ -102,6 +103,7 @@ class CallbackBrokerWorker(BaseWorker):
         now = tz_now()
         if (
             force or
+            (time.time() - self.last_flush) > settings.JOB_EVENT_BUFFER_SECONDS or
             any([len(events) >= 1000 for events in self.buff.values()])
         ):
             for cls, events in self.buff.items():
@@ -124,6 +126,7 @@ class CallbackBrokerWorker(BaseWorker):
                 for e in events:
                     emit_event_detail(e)
             self.buff = {}
+            self.last_flush = time.time()
 
     def perform_work(self, body):
         try:
--- a/awx/settings/defaults.py
+++ b/awx/settings/defaults.py
@@ -196,9 +196,9 @@ LOCAL_STDOUT_EXPIRE_TIME = 2592000
 # events into the database
 JOB_EVENT_WORKERS = 4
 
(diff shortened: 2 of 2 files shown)


awx/settings/defaults.py
```diff
@@ -198,7 +198,7 @@ def IS_TESTING(argv=None):
 
 # The number of seconds (must be an integer) to buffer callback receiver bulk
```
wenottingham: Please update this if you're changing it as below

ryanpetrello: > must be an integer

...and now I remember why I changed this.  I think the `timeout=` argument to the redis `blpop` must be an integer.

https://github.com/ansible/awx/pull/9086/files#diff-915de8c0ee78894addb6ce93fd72bda6adc7b9a0bf2ba54ae9e9f1a6c0d8f011R56

This PR probably still needs some adjusting and testing on my end, but I've got an idea on what to adjust now.

awx/main/dispatch/worker/callback.py
```diff
@@ -52,7 +53,7 @@ def __init__(self):
 
     def read(self, queue):
         try:
-            res = self.redis.blpop(settings.CALLBACK_QUEUE, timeout=settings.JOB_EVENT_BUFFER_SECONDS)
+            res = self.redis.blpop(settings.CALLBACK_QUEUE, timeout=1)
```
ryanpetrello, line 56: @chrismeyersfsu I think we should just set this to `1`, as `blpop` doesn't let you set this lower.  This timeout is mostly to avoid a scenario where we have a partially filled buffer in memory, and the read from redis hangs for a really long time.  The buffering/flushing code below is basically meant to flush "at least every .1s, or whenever the buffer is full, whichever comes first", and I goofed this up when I refactored it last year.

awx/main/dispatch/worker/callback.py
```diff
@@ -102,6 +104,7 @@ def flush(self, force=False):
         now = tz_now()
         if (
             force or
+            (time.time() - self.last_flush) > settings.JOB_EVENT_BUFFER_SECONDS or
```
ryanpetrello, line 106: @chrismeyersfsu *this* code, however, is called after every redis read, and should enforce an *actual* "time that has passed since the last *flush*" check per-read

So to summarize:

1.  Do a redis blpop.  If it blocks for >1s, that means `callback_tasks` is empty and there's nothing _new_, but the buffer may have _some_ data in it.  Flush.
2. If the buffer is full of (1000) events, flush.
3. If `JOB_EVENT_BUFFER_SECONDS` _actual seconds_ have passed since the last time we flushed, flush.  This is the situation described in https://github.com/ansible/awx/issues/9085 - if you emit events in a loop with just the right timing, you never arrive at case number 1 (`blpop` never times out), and it takes a *long* time to reach case 2 (you've buffered 1000 events).  In _real world deployments_, it's probably somewhat difficult to encounter this unless you have a very slow trickle of event data (like the example in my issue).

ryanpetrello: @chrismeyersfsu I'm fairly confident that this fixes the use case outlined in the associated issue, but I want to try all of this together in a large-volume deployment on a large VM (to make sure we're not losing some of our performance gains in prior releases if there's backpressure in redis)

If things look good, we can let this bake in `devel` awhile.

ryanpetrello:
This patch on a fairly beefy VM (I'm happy with these numbers):

image

chrismeyersfsu:
What is the average size of our batch when events are getting created at full speed? I want to make sure that 0.25 seconds i enough to batch 1,000 requests.

ryanpetrello:
@chrismeyersfsu I don't have my test environment spun up anymore, but when I was testing this, I was seeing inserts in the high 900's range.

chrismeyersfsu (approved):

PR #9089, opened 2021-01-14
Survey update to workflows

sean-m-sullivan:
##### SUMMARY
It was pointed out that the awx endpoints refer to survey_spec. 
Job templates refer to survey_spec
However workflows were just survey.

This update changed the module option to survey_spec, but keeps the alias of survey to be backwards compatible. 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - Collection
 - tower_role

##### AWX VERSION

```
awx: 16.0.0
```


AlanCoding (approved):

AlanCoding:
Trying to run tests:

```
Error Message
The test `ansible-test sanity --test validate-modules` failed with 1 error:
Stacktrace
plugins/modules/tower_workflow_job_template.py:0:0: nonexistent-parameter-documented: Argument 'survey' is listed in DOCUMENTATION.options, but not accepted by the module argument_spec
```

So it looks like that still needs fixing before we can merge.

Granted, I don't know exactly why this is failing, given your code.

sean-m-sullivan:
@AlanCoding I found that the alias had not actually been set on line 166 of the workflow module, so this is why it most likely failed.

beeankha (approved):

PR #9093, opened 2021-01-15
Fix setting toggle on click behavior

marshmalien:
##### SUMMARY
This PR fixes a bug where the setting toggle value did not change on click. The form group used the same `id` attr value as the form label and input which conflicted with the label control. To fix this, I updated the form group `id` to be different from the form label and input. 

toggle bug

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - UI


jlmitch5 (approved):
nice

unlikelyzero:
Added coverage here https://github.com/ansible/tower-qa/pull/5901

PR #9094, opened 2021-01-15
Adds translation linting and addresses issues the linter found

AlexSCorey:
##### SUMMARY
This PR is a POC and my recommendation for moving forward with our existing translation library based on what I've found. 

The reason for the investigation
Ling UI, our current translation library, is deprecating the HOC(`withI18n`) which we use extensively throughout the project.  Thus, we need to upgrade this library, which involves touching a ton of files to convert them over to React hooks (more within our existing pattern), or use the `<I18n>` component (not really within our existing pattern overall, though we do use it in some places currently).  We also wanted to explore the ways to shorten the translation workflow loop by adding functionality for linting for translated strings.  We thought this was a great time to explore other translation libraries to see if we wanted to migrate to another, or stay with what we have. 

My recommendation
I explored I18next, i18n-js, and LingUI.  I determined, fairly quickly that I18next was a good candidate for us as well as keeping LingUI, and decided that I18n-js wasn't a good candidate because it didn't give us any additional functionality.  I enventually discovered that we can keep LingUI and add functionality for linting for translations via https://github.com/edvardchen/eslint-plugin-i18next.  Going with this choice means we don't have to figure out new tooling, and we get to keep a library that we are already comfortable with that seems to be well maintained.  

About this PR
This PR brings in eslint-plugin-i18next, to be used along side LingUI, and fixes all the files that need too be fixed.  This PR adds linting for only those strings inside the JSX and does not lint strings outside of it. Therefore, if a string is defined in a variable, an object, or an array outside of JSX and that string needs to be translated but isn't the linter will not catch it.  This linter does have quite a bit of flexibility so we can probably figure out a way to check for those situations also but that wasn't a deal breaker for me.  

To lint
`npm run lint`

##### ISSUE TYPE
-tooling
-enhancement
##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/components/CodeMirrorInput/VariablesDetail.jsx
+++ b/awx/ui_next/src/components/CodeMirrorInput/VariablesDetail.jsx
@@ -1,6 +1,7 @@
 import 'styled-components/macro';
 import React, { useState, useEffect } from 'react';
 import { node, number, oneOfType, shape, string, arrayOf } from 'prop-types';
+import { Trans, withI18n } from '@lingui/react';
 import { Split, SplitItem, TextListItemVariants } from '@patternfly/react-core';
 import { DetailName, DetailValue } from '../DetailList';
 import MultiButtonToggle from '../MultiButtonToggle';
@@ -111,7 +112,7 @@ function VariablesDetail({ dataCy, helpText, value, label, rows, fullHeight }) {
             css="color: var(--pf-global--danger-color--100);
             font-size: var(--pf-global--FontSize--sm"
           >
-            Error: {error.message}
+            <Trans>Error:</Trans> {error.message}
           </div>
         )}
       </DetailValue>
@@ -131,4 +132,4 @@ VariablesDetail.defaultProps = {
   helpText: '',
 };
 
-export default VariablesDetail;
+export default withI18n()(VariablesDetail);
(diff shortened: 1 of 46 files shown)


awx/ui_next/src/components/CodeMirrorInput/VariablesDetail.jsx
```diff
@@ -111,7 +112,7 @@ function VariablesDetail({ dataCy, helpText, value, label, rows, fullHeight }) {
             css="color: var(--pf-global--danger-color--100);
             font-size: var(--pf-global--FontSize--sm"
           >
-            Error: {error.message}
+            <Trans>Error:</Trans> {error.message}
```
keithjgrant, line 115: Did you confirm this text renders ok? We've had some issues with <Trans> contents not displaying in the past

AlexSCorey: It does, yes

jakemcdermott (approved):

unlikelyzero:
@AlexSCorey I don't think that this will break any user facing problems which would cause a failed test, but please run this through yolok to make sure that it's green. If anything is broken, please sync with ui-qe

Also, it might be worth creating a unit test to verify that this fails in a meaningful way.

AlexSCorey:
https://github.com/ansible/tower-qa/actions/runs/528335582 workflow tests passed

unlikelyzero (approved):
This is a gamechanger!

Issue #9010, opened 2021-01-05
Elements in a row shift on table sort action


unlikelyzero:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Elements in a row shift on table sort action

##### ENVIRONMENT
* AWX version: 96a7fe0035427774f4eac93983cbb887ab275a42

##### STEPS TO REPRODUCE
1. Navigate to a page with a table view
2. Click on sort
Note: the tables shifts

##### EXPECTED RESULTS
Table not to shift
##### ACTUAL RESULTS
Elements shift

##### ADDITIONAL INFORMATION
![Screen Recording 2021-01-05 at 01 57 35 PM](https://user-images.githubusercontent.com/12446869/103687586-8281d900-4f5e-11eb-820f-1b935e446397.gif)

keithjgrant:
We decided to punt this and address it alongside #7548 — it might make more sense to hide the column headers while the skeleton loaders display, thus making the shifting columns widths hidden during load and acceptable is is

Issue #7548, opened 2020-07-04
Sort on lists does not have a consistent, deterministic 'finished' or 'ready' state


unlikelyzero:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Sort on lists does not have a consistent, deterministic 'finished' or 'ready' state. This makes test automation non-deterministic. The reason I bring this up now, is that the behavior is being seen on a small dataset with few resources. Whatever solution we come up with should be to handle large datasets.

Our sort list test performs the following:
1. Create Resource 'A' and 'B'
2. Navigate to Resource List by url
3. Verify that Resource 'A' and 'B' are listed
4. Click 'Sort' by 'Name' 'A->Z' button.
5. Verify that the url is `schedules?schedule.order_by=name`
    Verify the resource order by the aria-label attribute.
6. Click 'Sort' by 'Name' 'Z->A' button.
7. Verify that the url is `schedules?schedule.order_by=-name`
    Verify the resource order by the aria-label attribute.

Problems:
The time between step 6 and 7 can be greater than 5seconds. We know this because we cheated and inserted a 5s wait.
The previous order of elements on screen is incorrect until it is refreshed.
The time between the XHR request being received that the resource order changing to match the sort order varies.
The time between the sort order changing and the url updating varies.
The time between 'Sort' button click and the 'Sort' button element updating to 'Z->A' does not reflect the state of the listed resources.

Solutions:
1. Do not show list of resources until order is resource order is correct. i.e. Click on the sort button should hide resources until they're refreshed and displayed in the specified order.
2. Provide some indication that the sort order is correct.
3. Potentially implement OUIA pagesafe attribute

##### ADDITIONAL INFORMATION

https://dashboard.cypress.io/projects/ui2mq7/runs/1456/failures/bfd7df5e-48b7-457d-acd2-1f919c22ffb4/video
https://dashboard.cypress.io/projects/ui2mq7/runs/1498/failures/3895008f-bb6e-4022-bfbf-60d3d8d1faee/video

mabashian:
Skeleton loading

keithjgrant:
Ideally, I'd like to find a solution that doesn't hide the list while loading a new list, because this can cause the page to jump around and/or generally draw more of the user's attention to the slow load time. I'm inclined to look for a way we could show a loading indicator somewhere while an updated list is loading.

keithjgrant:
[...]

PR #9095, opened 2021-01-15
Explicitly set table column widths

keithjgrant:
##### SUMMARY
Adds `width` prop to HeaderCell component for list tables, to minimize column shifting when changing pagination. Updates Org and Inventories lists with some specified widths.

Note this is not 100% foolproof, as columns will still shrink to fit when the browser width is constrained — but IMO this is the preferred behavior for that case, since fitting the data on screen is more important than preventing shift.

Adresses #9010 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/components/PaginatedTable/HeaderRow.jsx
+++ b/awx/ui_next/src/components/PaginatedTable/HeaderRow.jsx
@@ -1,3 +1,4 @@
+import 'styled-components/macro';
 import React from 'react';
 import { useLocation, useHistory } from 'react-router-dom';
 import { Thead, Tr, Th as PFTh } from '@patternfly/react-table';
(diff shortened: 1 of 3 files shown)


jakemcdermott (approved):

keithjgrant:
I discussed this with @jlmitch5 yesterday and he had some concerns about simply using percentages to control columns widths, as this may not meet all our needs and in some cases can be confusing because we I didn't allocate the full 100% of width among all the columns (leaving some to the browser's table layout algorithm).

I'm going to experiment a bit with other possible approaches/tweaks to the way I have it currently.

keithjgrant:
We've decided to shelf this for a while, and circle back when addressing #7548 so we can take a more holistic approach to both problems at once

PR #9098, opened 2021-01-15
Update project_update.yml

dlemons-redhat:
##### SUMMARY

In project_update.yml, Set the depth of the pull to 1 in order to save space and bandwidth in the SCM transactions.  Tower will not be taking advantage of all of the prior revisions.  

This issue impacts a customer whose repositories have outgrown the disks of their Tower VMs.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - Project update functionality of UI
 

##### AWX VERSION

```
any I guess
```

##### ADDITIONAL INFORMATION

```
vim /var/lib/awx/venv/awx/lib/python3.6/site-packages/awx/playbooks/project_update.yml

[1]+  Stopped                 vim /var/lib/awx/venv/awx/lib/python3.6/site-packages/awx/playbooks/project_update.yml
[root@localhost projects]# du -sh *
148K	_6__demo_project
0	_6__demo_project.lock
508K	_8__splunk
0	_8__splunk.lock

## run update in the Tower UI

[root@localhost projects]# du -sh *
148K	_6__demo_project
0	_6__demo_project.lock
184K	_8__splunk
0	_8__splunk.lock
[root@localhost projects]# pwd
/var/lib/awx/projects

```

### DIFF
--- a/awx/playbooks/project_update.yml
+++ b/awx/playbooks/project_update.yml
@@ -39,6 +39,7 @@
             refspec: "{{scm_refspec|default(omit)}}"
             force: "{{scm_clean}}"
             accept_hostkey: "{{scm_accept_hostkey|default(omit)}}"
+            depth: 1
           register: git_result
 
         - name: Set the git repository version


AlanCoding (commented):
This might be faster on the first pull, but that doesn't mean it's faster or better in the long term. We omit the git history when we copy the project tree into the folder where the job runs. There are a lot of things that can be happening in the project folder, like switching between branches and this is unlikely to work well with those features we need to support.

dlemons-redhat:
Looking at the playbook, I don't see where changing branches will have a significant impact.  They will just be pulling the latest version, which is the only revision tower operates on.  To clarify, I am referring to the behavior described in the docs:

Typically, during a project update, the revision of the default branch (specified in the SCM Branch field of the project) is stored when updated, and jobs using that project will employ this revision. Providing a non-default SCM Branch (not a commit hash or tag) in a job, the newest revision is pulled from the source control remote immediately before the job starts. This revision is shown in the Revision field of the job and its respective project update.

source: https://docs.ansible.com/ansible-tower/latest/html/userguide/job_branching.html


Updated initial PR with customer reference.

AlanCoding:
I ran a subset of our full integration tests to get a specific thing this breaks, and projects using parameters like this error:

```
            scm_refspec=f"refs/pull/{pr_number}/head:refs/remotes/origin/pull/{pr_number}/head",
            scm_branch=f"pull/{pr_number}/head"
```

This is a documented thing that we support. Running it combined with this diff gives output like this:

```
PLAY [Update source tree if necessary] *****************************************

TASK [update project using git] ************************************************
fatal: [localhost]: FAILED! => {"changed": false, "cmd": "/usr/bin/git checkout --force pull/145/head", "msg": "Failed to checkout pull/145/head", "rc": 1, "stderr": "error: pathspec 'pull/145/head' did not match any file(s) known to git\\n", "stderr_lines": ["error: pathspec 'pull/145/head' did not match any file(s) known to git"], "stdout": "", "stdout_lines": []}

PLAY RECAP *********************************************************************
localhost                  : ok=0    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0 
```

There's probably an issue with the git module, these may be mutually exclusive parameters but not documented. It's very possible there's something you could do in ansible-core to make progress on this, but we are a long way away from AWX being able to support this sanely.

Issue #8544, opened 2020-11-05
awx.awx launch modules do not allow survey only variables when ask_variables_on_launch is false


sean-m-sullivan:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
The following modules:
awx.awx.tower_workflow_launch
awx.awx.tower_job_launch

Do not allow a user to pass survey only vars through the api, as when using the api interface through the uri module or other methods. Believe this is a logic problem in the above modules but not sure on how best to address it. 

##### ENVIRONMENT
* AWX version: 15.0.1

##### STEPS TO REPRODUCE
Attempt to launch a job with extra vars and survey enabled, but ask_variables_on_launch is false
Attempt to launch a workflow with extra vars and survey enabled, but ask_variables_on_launch is false

##### EXPECTED RESULTS
Job is launched without issue as it does when using the URI module. 

##### ACTUAL RESULTS
Error occurs because ask_variables_on_launch is not enabled.

##### PROPOSED ACTIONS
Either remove the check for extra vars, or add logic to allow extra vars when survey is enabled.

sean-m-sullivan:
adding more info. 
When Only a survey is enabled on a template/workflow, not ask extra vars, this should work. When those 3 variables are the survey variables.
```
    - name: Start Job
      uri:
        url: "https://{{tower_server}}/api/v2/job_templates/21/launch/"
        method: POST
        user: admin
        password: "password"
        body: 
          extra_vars:
              username: "admin"
              password: "password"
              user_env_type: "lab"
        force_basic_auth: yes
        status_code: 201
        body_format: json
        validate_certs: no
```

if you add a 4th variable that isn't a survey variable, it will fail.

because ask_extra_vars on launch isn't checked.

john-westcott-iv:
I think this is a limitation in the API itself, not in something the modules are doing. However, you said that the URI module works? Do you have an example of that?

sean-m-sullivan:
Using the above URI with Body/Body commented out when Only Survey Vars is No, I came up with the following
| Survey Enabled | Extra vars at launch | only Survey Vars | Result   |
|--------------- | -------------------- |----------------- |--------- |
|       Yes      |          Yes         |       Yes        | Launches |
|       Yes      |          Yes         |       No         | Launches |
|       Yes      |          No          |       No         | Fails Says missing survey Variables |
|       No       |          No          |       No         | Launches |
|       No       |          Yes         |       No         | Launches |
[...]

PR #9104, opened 2021-01-18
Survey logic for awx modules

sean-m-sullivan:
##### SUMMARY
Update survey logic to pass survey only vars through the api, and add tests to reflect this. This will fail if the user tries to pass to many variables if only survey is enabled. However this mirrors the action of the API. 

As referenced in Issue #8544

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
 - Collection
 - tower_role

##### AWX VERSION
```
awx: 16.0.0
```


beeankha (approved):

awx_collection/plugins/modules/tower_job_launch.py
```diff
@@ -225,6 +224,9 @@ def main():
     for variable_name in check_vars_to_prompts:
         if module.params.get(variable_name) and not job_template[check_vars_to_prompts[variable_name]]:
             param_errors.append("The field {0} was specified but the job template does not allow for it to be overridden".format(variable_name))
+    # Check if Either ask_variables_on_launch, or survey_enabled is enabled for use of extra vars.
+    if module.params.get('extra_vars') and not (job_template['ask_variables_on_launch'] or job_template['survey_enabled']):
+        param_errors.append("The field extra_vars was specified but the job template does not allow for it to be overridden")
```
AlanCoding, line 229: My guidance on these cases is that we should rely on The API validation if possible. This doesn't do any harm, but it's obviously not comprehensive, and it would be burdensome to make it so. In other words, just don't validate inputs if the API is likely to give a descriptive message, and let the user see that message.

That's my general guiding principle for these cases, and mean it only as a suggestion. Again, this type of thing shouldn't do _harm_. I would just be interested in deleting blocks like these and see if we can get the tests adjusted to pass with adjusted verbiage.

AlanCoding (approved):

awx_collection/plugins/modules/tower_job_launch.py
sean-m-sullivan: Right now we could do that and I believe we would just need to change the assertions in our tests, however I don't believe it gives a good error message, and thought this was the best way to go around it our earlier restriction, while allowing everything the tower api allows through.

PR #9113, opened 2021-01-19
Fix bind-mount problem with pg12 upgrades

rooftopcellist:
##### SUMMARY
Issue: https://github.com/ansible/awx/issues/9077

There was an issue in https://github.com/ansible/awx/pull/9078 that I found out after merging where the old postgres:10 container was mistakenly being used during the task that runs the migrations.   This PR fixes that and a couple other things.  

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - Installer

##### AWX VERSION

```
devel
```

### DIFF
--- a/installer/roles/local_docker/tasks/compose.yml
+++ b/installer/roles/local_docker/tasks/compose.yml
@@ -36,6 +36,12 @@
   register: awx_secret_key
 
 - block:
+    - name: Remove AWX containers before migrating postgres so that the old postgres container does not get used
+      docker_compose:
+        project_src: "{{ docker_compose_dir }}"
+        state: absent
+      ignore_errors: true
+
     - name: Run migrations in task container
       shell: docker-compose run --rm --service-ports task awx-manage migrate --no-input
       args:
--- a/installer/roles/local_docker/tasks/upgrade_postgres.yml
+++ b/installer/roles/local_docker/tasks/upgrade_postgres.yml
@@ -1,27 +1,38 @@
 ---
 
+- name: Get full path of postgres data dir
+  shell: "echo {{ postgres_data_dir }}"
+  register: fq_postgres_data_dir
+
 - name: Register temporary docker container
   set_fact:
-    container_command: "docker run -v '{{ postgres_data_dir | realpath }}:/var/lib/postgresql' centos:8 bash "
+    container_command: "docker run --rm -v '{{ fq_postgres_data_dir.stdout }}:/var/lib/postgresql' centos:8 bash -c "
 
 - name: Check for existing Postgres data (run from inside the container for access to file)
   shell:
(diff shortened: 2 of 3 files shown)


installer/roles/local_docker/tasks/compose.yml
```diff
@@ -36,6 +36,12 @@
   register: awx_secret_key
 
 - block:
+    - name: Remove AWX containers before migrating postgres so that the old postgres container does not get used
```
rooftopcellist, line 39: Removing the old postgres container here prevents docker-compose from using it, as it satisfies the `depends_on` for the task container.  The result is that postgres:12 is used every time, regardless of what other postgres containers are available.

installer/roles/local_docker/tasks/upgrade_postgres.yml
```diff
@@ -1,43 +1,59 @@
 ---
 
+- name: Get full path of postgres data dir
+  shell: "echo {{ postgres_data_dir }}"
+  register: fq_postgres_data_dir
```
rooftopcellist, line 5: `real_path` does not work here.  More info here: https://github.com/ansible/awx/pull/9078/files#r559794551

installer/roles/local_docker/tasks/upgrade_postgres.yml
```diff
@@ -1,43 +1,59 @@
...
+  when: old_pg_version.changed
+
+- name: Set up new postgres paths pre-upgrade
+  shell: |
+    {{ container_command }} "mkdir -p /var/lib/postgresql/12/data/"
+  when: upgrade_postgres | bool
 
 - name: Stop AWX before upgrading postgres
-  docker_service:
+  docker_compose:
```
rooftopcellist, line 35: `docker_service` is deprecated in favor of `docker_compose`.

installer/roles/local_docker/templates/docker-compose.yml.j2
```diff
@@ -178,7 +178,7 @@ services:
     container_name: awx_postgres
     restart: unless-stopped
     volumes:
-      - "{{ postgres_data_dir }}:/var/lib/postgresql:Z"
```
rooftopcellist, line 181: The `12/data` directory must explicitly be bind-mounted to `/var/lib/postgresql/data` where postgresql expects it, else it will init it's own data directory and none of the old data will be in the new AWX install.

ryanpetrello (approved):

rooftopcellist (commented):
I just ran through a fresh install and an upgrade from 16.0.0, both were successful.  

Data that was created on 16.0.0 was successfully persisted after the upgrade.  
image

beeankha (approved):

PR #9114, opened 2021-01-19
Convert TemplateList to PaginatedTable

keithjgrant:
##### SUMMARY
Converts TemplateList component to use PaginatedTable. I also added the expandable content rows

addresses #6189

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI
 - 
Screen Shot 2021-01-19 at 2 59 47 PM


jakemcdermott (approved):

tiagodread:
Awesome Job @keithjgrant, so far, I've found these items:

- Expanded view missing `Last modified` and `Last Ran`:

image

- Workflow Job Template doesn’t contain project, we are displaying on the Expanded view:

image
Old:
image

- Inventory list contains a column header for `Actions`:

image


But this was not applied for `Jobs list` AND `Job Template list`:

image
 
- The column Last Run doesn’t make sense when you try to sort IMO, because this is showing the playbook activity instead of the Last Run timestamp value, maybe if we rename this column to `Activity` (not sortable) and add a new column for `Last Run` with timestamp (sortable) would be better. cc @keithjgrant @trahman73 

image
 
- Add data-ouia-id (when available) or IDs on:

 - each sort option (name, type, Last run...)
 - each action button
 - each <dd> element on expanded view

trahman73:
@tiagodread If it is not sortable, we can perhaps show Activity in the expanded area?

tiagodread:
> @tiagodread If it is not sortable, we can perhaps show Activity in the expanded area?

Yeah @trahman73 we can do that, actually looks better than my suggestion 😄

keithjgrant:
@tiagodread Just pushed fixes for the items you listed.

I also changed the Last Run column to display a timestamp of the last run, leaving the detail sparkview in the expanded column

tiagodread:
recheck

tiagodread:
@keithjgrant the Dashboard -> Recent Templates List don't inherit the `Templates` the same PaginatedTable list?

jt

jakemcdermott (approved):

awx/ui_next/src/components/DetailList/Detail.jsx
```diff
@@ -34,6 +34,7 @@ const DetailValue = styled(
 `;
 
 const Detail = ({
+  id,
```
tiagodread, line 37: @keithjgrant defined but never used

tiagodread: @keithjgrant defined but never used

tiagodread (commented):
only a small lint error, after fixing this, this is ready to go.

tiagodread (approved):
Nice! fixed broken tests here https://github.com/ansible/tower-qa/pull/5917

PR #9118, opened 2021-01-20
[WIP] Replace license refs with subscription refs where possible without ne…

rooftopcellist:
…cessitating a migration

##### SUMMARY

We still have a number of "license" references that should be changed to "subscriptions".  This is some follow-up cleanup from when we moved to a subscription-based model.  

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API
 - UI

### DIFF
--- a/awx/api/generics.py
+++ b/awx/api/generics.py
@@ -188,7 +188,7 @@ class APIView(views.APIView):
         Log warning for 400 requests.  Add header with elapsed time.
         '''
         from awx.main.utils import get_licenser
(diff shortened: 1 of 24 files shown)


ryanpetrello:
If there's not user-facing strings or messaging here, I'd consider this `priority:very_low`; certainly not worth spending cycles on the necessary code changes in QE _right now_ 😄.

rooftopcellist:
Agreed, it looks like there are the tests are not happy with it either, so I'm just going to close this.  

image

PR #9119, opened 2021-01-20
Adding option to use custom ansible-runner

john-westcott-iv:
##### SUMMARY

In developing an AWX feature I needed to make modifications to ansible-runner. This PR allows me to place ansible-runner code in the tools directory which will be picked up and installed as part of the `make docker-compose-build` phase. If no source code is provided, the default version of ansible-runner will be installed from the existing requirements/requirements.txt.

##### ISSUE TYPE

 - Feature Pull Request
 
##### COMPONENT NAME

 - Installer

##### AWX VERSION

```
awx: 16.0.0
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,7 @@ awx/ui_next/instrumented
 rsyslog.pid
 tools/prometheus/data
 tools/docker-compose/Dockerfile
+tools/ansible-runner
 
 # Tower setup playbook testing
 setup/test/roles/postgresql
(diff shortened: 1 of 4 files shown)


wenottingham:
Could this be done with a git+... reference in the requirements/requirements.txt?

shanemcd:
@wenottingham yes. That would be the preferred way of doing this. I'm going to close since we already have this ability.

shanemcd:
See here for an example of how to do this: https://github.com/ansible/awx/blob/execution-environments/requirements/requirements_git.txt

Issue #5157, opened 2019-10-29
Execution environment definition, creation, and management


wenottingham:
##### ISSUE TYPE
 - Feature Idea

(I expect subtasks will live off of here.)

##### SUMMARY

Ansible is moving to collections. Collections have dependencies. Dependencies complicate providing coherent execution of playbooks. We have historically hidden this complexity for cloud modules, but it seeps in occasionally (especially with Azure), and we leave custom venv creation entirely to our users. This is complicated by the fact that the AWX operator is often not the playbook author. Our users aren’t venv and pip experts; we need to make things easier for them where we can.

Hence, AWX needs to expand to provision and manage execution environments, including how they get to execution nodes.

**What is an execution environment?**

An execution environment definition consists of:
- an ansible version
- a python version (defaults to system python)
- a tiny set of required python libraries
- zero or more collections (optional, see below). This could be sourced from the requirements present in a playbook repo
- python dependencies of those collections

The concept of specifying a set of collections for an environment is to resolve and install their dependencies. The collections themselves are not required to be in the environment (and arguably should not be.)

An execution environment is built from this definition. This will likely be a container image.

**What should we provide to users?**

1. Create a standardized definition of an execution environment. Given our ecosystem and the cloud world at large, I sense… YAML.
2. Create a tool, that given this definition, creates an execution environment as a container image. It would do so by reading any provided collections, extracting the dependency information, and then creating the environment in a container image.
- it should warn/error on dependency issues
- it should warn/error if the collections do not claim compatibility with the provided Ansible version
3. Add execution environments as first-order objects in AWX
4. Read environment definitions from SCM as needed
5. Manage how execution environments get on execution nodes. This is likely via container pull.
6. Replace the existing ‘virtual environment’ construct with this

AlanCoding:
I was going to try to post a bunch of links here from Ansible core. The [roadmap](https://docs.ansible.com/ansible/devel/roadmap/ROADMAP_2_10.html) for 2.10 isn't enlightening for me. Maybe there's a longer-term roadmap somewhere? Some requests, like https://github.com/ansible/ansible/issues/58266, are counter-productive.

There are some surprising formats used for Ansible module python requirements

 - https://github.com/ansible/ansible/blob/5349b3ae4c056ca5f5a32bc41a5e7eb9d60a7f46/lib/ansible/modules/cloud/docker/docker_volume_info.py#L39
[...]

PR #9124, opened 2021-01-21
openstacksdk up to update to latest

VitalyKapustin-git:
##### SUMMARY
Openstack.cloud collection modules require openstacksdk>0.37.
Example:
TASK [openstack.cloud.project] *****************************************
fatal: [192.168.89.5]: FAILED! => {"changed": false, "msg": "To utilize this module, the installed version of the openstacksdk library MUST be >=0.45.1."}

So just removed concrete version and let pip to instal latest from repo.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - Installer

### DIFF
--- a/requirements/requirements_ansible.in
+++ b/requirements/requirements_ansible.in
@@ -61,7 +61,7 @@ pywinrm[kerberos]==0.3.0
 requests
 requests-credssp==1.0.2   # For windows authentication awx/issues/1144
 # OpenStack
-openstacksdk==0.37.0
+openstacksdk
 # Openshift/k8s
 openshift>=0.11.0  # minimum version to pull in new pyyaml for CVE-2017-18342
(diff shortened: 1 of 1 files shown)


AlanCoding:
This still leaves this with older version

https://github.com/ansible/awx/blob/accb3e0a9acc84a504d3e9ec02ca2124a58c57b1/requirements/requirements_ansible.txt#L80

you can run the script in that folder, and hopefully it should resolve dependencies automatically to just do this targeted upgrade

ryanpetrello:
Hey @VitalyKapustin-git,

This PR is unfortunately out of date, because the next major version of AWX will specify this (and other) dependency via a default execution environments (we're removing virtual environments).  Details here:

https://github.com/ansible/awx/issues/5157

Issue #578, opened 2017-11-03
Feature idea: programmatic job/workflow templates & surveys in SCM


wenottingham:
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
 - API
 - UI

##### SUMMARY

We already pull playbooks from source control - it would be neat if the job template & workflow definitions (including surveys, potentially even roles) could be sourced from there as well.

(Probably a very large discussion of *how* this would work, and what to support, is required.)

##### ENVIRONMENT
* AWX version: any

lhoss:
Very good point!
In comparing AWX with Jenkins+Ansible+Pipeline(plugins), it is IMHO a very strong point for the Jenkins solution to be able to source-control the pipeline code (including any ansible steps).
And us having >50 playbooks (most based on a role), having a way to manage+auto-generate Job/Templates for >80% of those playbooks, could be highly useful

hemskgren:
Very good point!

perzizzle:
I would be very interested in this feature.

ghost:
Hi Guys, any timelines when this feature is coming out ?

powertim:
I'm also interested to test this new feature !

ghost:
+1

abedwardsw:
+1 I think this is a really good idea, I was searching for something similar.  In general I find we end up with too much manual configuration of jobs in AWX, having it sourced from Git would provide much needed version control (rollback capabilities) and simplify the process of moving between awx instances.

@wenottingham have you given much more thought to this feature?  Is the implementation significant?

If implementation is significant, wonder if a easier starting point might be the ability to have the tower-cli synchronize a git repo to a org for example.  Kinda infrastructure as code approach.

wenottingham:
What do you mean by significant? I suspect there's a decent amount of design that needs to go into it to make sure it's sufficiently future proof, and doesn't clash horribly with tower-cli and other things.

abedwardsw:
by significant I was wondering if this is a large change requiring weeks of effort.  it sounds like it is. ;-) .  Would be a great feature if/when someone has time to implement.

MarkusTeufelberger:
Having a role/playbook to interact with AWX using the https://docs.ansible.com/ansible/latest/modules/tower_job_template_module.html is a workaround, but then I'd probably need to lock down user interaction on the server (and/or an easier way to export the current templates into such tasks).

[...]

PR #9129, opened 2021-01-21
[WIP] Project update sync assets (round 2)

john-westcott-iv:
##### SUMMARY

This PR replaces PR #7099

This is an initial cut as having parts of Tower being controlled by code. A project update of an SCM based project will not attempt to create job templates and workflow job templates from files within the source. The current proposed files are:
assets/job_templates.yml
assets/workflow_job_template_nodes.yml
assets/workflow_job_templates.yml

This PR (currently):
* Adds a "Sync Assets" checkbox to the project UI screen. If enabled, during a manually invoked project sync the project_update.yml playbook will now include a role which will read these files and then leverage the awx.awx collection to build assets within tower.
* Adds an export endpoint to project API (still needs work).
* Adds a PROJECT_SYNC_ACCESS_TOKEN_EXPIRE_SECONDS setting
* Modifies UI for project export job types

Related
#526
#578

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API
 - UI

##### AWX VERSION

```
16.0.0
```

##### ADDITIONAL INFORMATION

```

```


shanemcd:
Like we talked about earlier, I would be interested in seeing if we can associate a Tower credential to the project and use that instead of obtaining an oauth token behind the scenes.

john-westcott-iv:
@ryanpetrello @wenottingham Should we allow for relaunch on a project export. Keeping in mind that an export takes a commit message and a list of job template IDs. On a failure I suppose it would make sense. However, on a successful job export would we want people to be able to relaunch with the same commit message?

john-westcott-iv:
> Like we talked about earlier, I would be interested in seeing if we can associate a Tower credential to the project and use that instead of obtaining an oauth token behind the scenes.

@shanemcd @wenottingham If we use a Tower credential type for the project export it has a field called Ansible Tower Hostname (thats required). I feel like I should respect this when making the connection to Tower (right now I use TOWER_URL_BASE setting). This could lead to awkwardness on an export if someone does not populate that correctly. Any thoughts on this?

john-westcott-iv:
recheck

awx/ui_next/src/screens/Project/shared/ProjectSubForms/SharedFields.jsx
```diff
@@ -93,6 +93,15 @@ export const ScmTypeOptions = withI18n()(
                   revision of the project prior to starting the job.`
             )}
           />
+          <CheckboxField
+            id="option-sync_assets"
```
unlikelyzero, line 97: do we want underscores here? @mabashian

unlikelyzero:
@john-westcott-iv , i'll take a stab at testing this from the ui-side when you're ready for review

shanemcd:
Holy conflicts. Please let me know if you still want this.

john-westcott-iv:
I def sill want it but didn't have the time to put it in before the 4.0 release so this will have to be completely refactored later. Closing for now.

Issue #8749, opened 2020-12-04
Log Job Lifecycle


chrismeyersfsu:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Users want to know why a job isn't running.

More generally, we want to tell user what and when things are happening for a Job in AWX. We can accomplish this if we know the log the lifecycle of a job.

### Job Life-cycle

Many sub-systems in AWX perform operations, both serially and in parallel, on a single Job. Each operation takes time and contributes to the total runtime of job outside of the actual playbook run. 

| State                           | Description                                                  | Reason(s)                                                    |
| ------------------------------- | ------------------------------------------------------------ | ------------------------------------------------------------ |
| pending_epoch                   | {job.type}-{job.id} created                                  | Job created in the database and has yet to be seen by the task manager |
| pending_ack                     | {job.type}-{job.id} acknowledged                             | Job seen for the first time by the task manager              |
| pending_blocked_business        | {job.type2}-{job.id2} blocked by {job.type2}-{job.id2}       | Job blocked on Project Update<br />Job blocked on Inventory Update<br />Project Update blocked on <br />Job blocked on same-job-template Job (allow_simultaneous=False)<br /><br />Project Update blocked on previous Project Update<br /><br />Inventory Update blocked on previous Inventory Update |
| pending_blocked_capacity        | {job.type}-{job.id} needs {capacity_needed}. Considered groups [({group.name[i]}, group.capacity_remaining[i]), ] | Job blocked because the amount of capacity required by the task exceeds any single group remaining capacity |
| waiting                         | {job.type}-{job.id} transitioned to waiting, will run on node {job.execution_node} and consumes capacity {job.capacity_consumed} from group {group.name} leaving the group with {group.remaining_capacity} capacity | Job is not blocked by business logic and enough capacity is available to run. |
| running_pre                     | {job.type}-{job.id} starting                                 | Job has been picked up by the dispatcher and is set from waiting to running |
| running_started                 | {job.type}-{job.id} starting playbook run                    | Job has finished pre-run checks and parameter gathering and is going to start running the playbook |
| running_finished                | {job.type}-{job.id} finished playbook run                    | Job has finished running the playbook                        |
| running_post                    | {job.type}-{job.id}                                          |                                                              |
| running_finalize                |                                                              |                                                              |
| running_finished                | {job.type}-{job.id} finish                                   | Job status set to final state. At this point, the job status is set and will not change (i.e. successful, failed, error, canceled) |
[...]

PR #9130, opened 2021-01-21
log job lifecycle

fosterseth:
feature issue https://github.com/ansible/awx/issues/8749

##### SUMMARY

example output in `/var/log/tower/job_lifecycle.log`
```json
{"type": "job", "task_id": 353, "state": "created", "template_name": "sleep", "time": "2021-01-28T04:25:46.661029+00:00"}
{"type": "job", "task_id": 353, "state": "acknowledged", "template_name": "sleep", "time": "2021-01-28T04:25:47.125391+00:00"}
{"type": "job", "task_id": 350, "state": "blocked", "template_name": "sleep", "blocked_by": "projectupdate-351", "time": "2021-01-28T04:25:47.166337+00:00"}
{"type": "job", "task_id": 352, "state": "blocked", "template_name": "sleep", "blocked_by": "projectupdate-351", "time": "2021-01-28T04:25:47.176640+00:00"}
{"type": "job", "task_id": 353, "state": "blocked", "template_name": "sleep", "blocked_by": "projectupdate-351", "time": "2021-01-28T04:25:47.177712+00:00"}
```

these events also show up in the stdout log
`awx_1       | 2021-02-03 18:41:05,656 DEBUG    awx.analytics.job_lifecycle job-740 created`

##### Dependency Graph Changes

The `DependencyGraph` tells us _what_ is blocked, but doesn't tell us what is doing the blocking. I rewrote it to expose that information.

e.g. here is the new representation in `dependency_graph.data`
```python
{'inventory_source_ids': {},
 'inventory_source_updates': {},
 'inventory_updates': {},
 'job_template_jobs': {},
 'project_updates': {12: <ProjectUpdate: 2021-01-14 05:48:48.599418+00:00-2-pending>},
 'system_job': None,
 'workflow_job_template_jobs': {}}
```
now instead of `is_job_blocked` in the task manager, we have `job_blocked_by` which returns a `UnifiedJob` object, or `None` if it is not blocked.

##### Job Explanation Changes
We can also update `job_explanation` with some information about why a job is blocked.
e.g.
`waiting until the project update finishes`

To reduce performance loss with `task.save()`, we only update this `job_explanation` if it is different from the current job_explanation, and if more than 30 seconds have passed since `job.created`.

There is a PR open that will display job explanation in the UI, so this might be our mechanism of displaying information to the user.

https://github.com/ansible/awx/pull/8726/

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 16.0.0
```

### DIFF
--- a/awx/main/models/unified_jobs.py
+++ b/awx/main/models/unified_jobs.py
@@ -55,7 +55,7 @@ from awx.main.fields import JSONField, AskForField, OrderedManyToManyField
 __all__ = ['UnifiedJobTemplate', 'UnifiedJob', 'StdoutMaxBytesExceeded']
 
 logger = logging.getLogger('awx.main.models.unified_jobs')
-
+logger_job_lifecycle = logging.getLogger('awx.analytics.job_lifecycle')
 # NOTE: ACTIVE_STATES moved to constants because it is used by parent modules
 
 
@@ -420,7 +420,7 @@ class UnifiedJobTemplate(PolymorphicModel, CommonModelNameNotUnique, Notificatio
         # have been associated to the UJ
         if unified_job.__class__ in activity_stream_registrar.models:
             activity_stream_create(None, unified_job, True)
-
+        unified_job.log_lifecycle("created")
         return unified_job
 
     @classmethod
@@ -862,7 +862,7 @@ class UnifiedJob(PolymorphicModel, PasswordFieldsModel, CommonModelNameNotUnique
             self.unified_job_template = self._get_parent_instance()
             if 'unified_job_template' not in update_fields:
                 update_fields.append('unified_job_template')
-        
+
         if self.cancel_flag and not self.canceled_on:
             # Record the 'canceled' time.
             self.canceled_on = now()
@@ -1010,6 +1010,7 @@ class UnifiedJob(PolymorphicModel, PasswordFieldsModel, CommonModelNameNotUnique
             event_qs = self.get_event_queryset()
         except NotImplementedError:
             return True  # Model without events, such as WFJT
+        self.log_lifecycle("event_processing_finished")
         return self.emitted_events == event_qs.count()
 
     def result_stdout_raw_handle(self, enforce_max_bytes=True):
@@ -1318,6 +1319,10 @@ class UnifiedJob(PolymorphicModel, PasswordFieldsModel, CommonModelNameNotUnique
         if 'extra_vars' in kwargs:
             self.handle_extra_data(kwargs['extra_vars'])
 
+        # remove any job_explanations that may have been set while job was in pending
+        if self.job_explanation != "":
+            self.job_explanation = ""
+
         return (True, opts)
 
     def signal_start(self, **kwargs):
@@ -1484,3 +1489,17 @@ class UnifiedJob(PolymorphicModel, PasswordFieldsModel, CommonModelNameNotUnique
     @property
     def is_containerized(self):
         return False
+
+    def log_lifecycle(self, state, blocked_by=None):
+        extra={'type': self._meta.model_name,
+               'task_id': self.id,
+               'state': state}
+        if self.unified_job_template:
+            extra["template_name"] = self.unified_job_template.name
+        if state == "blocked" and blocked_by:
+            blocked_by_msg = f"{blocked_by._meta.model_name}-{blocked_by.id}"
+            msg = f"{self._meta.model_name}-{self.id} blocked by {blocked_by_msg}"
+            extra["blocked_by"] = blocked_by_msg
+        else:
+            msg = f"{self._meta.model_name}-{self.id} {state.replace('_', ' ')}"
+        logger_job_lifecycle.debug(msg, extra=extra)
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -64,6 +64,8 @@ class TaskManager():
         # will no longer be started and will be started on the next task manager cycle.
         self.start_task_limit = settings.START_TASK_LIMIT
 
+        self.time_delta_job_explanation = timedelta(seconds=30)
+
     def after_lock_init(self):
         '''
         Init AFTER we know this instance of the task manager will run because the lock is acquired.
@@ -80,7 +82,7 @@ class TaskManager():
         instances_by_hostname = {i.hostname: i for i in instances_partial}
 
         for rampart_group in InstanceGroup.objects.prefetch_related('instances'):
-            self.graph[rampart_group.name] = dict(graph=DependencyGraph(rampart_group.name),
+            self.graph[rampart_group.name] = dict(graph=DependencyGraph(),
                                                   capacity_total=rampart_group.capacity,
                                                   consumed_capacity=0,
                                                   instances=[])
@@ -88,18 +90,21 @@ class TaskManager():
                 if instance.hostname in instances_by_hostname:
                     self.graph[rampart_group.name]['instances'].append(instances_by_hostname[instance.hostname])
 
-    def is_job_blocked(self, task):
+    def job_blocked_by(self, task):
         # TODO: I'm not happy with this, I think blocking behavior should be decided outside of the dependency graph
         #       in the old task manager this was handled as a method on each task object outside of the graph and
         #       probably has the side effect of cutting down *a lot* of the logic from this task manager class
         for g in self.graph:
-            if self.graph[g]['graph'].is_job_blocked(task):
-                return True
+            blocked_by = self.graph[g]['graph'].task_blocked_by(task)
+            if blocked_by:
+                return blocked_by
 
         if not task.dependent_jobs_finished():
-            return True
+            blocked_by = task.dependent_jobs.first()
+            if blocked_by:
+                return blocked_by
 
-        return False
+        return None
 
     def get_tasks(self, status_list=('pending', 'waiting', 'running')):
         jobs = [j for j in Job.objects.filter(status__in=status_list).prefetch_related('instance_group')]
@@ -312,6 +317,7 @@ class TaskManager():
             with disable_activity_stream():
                 task.celery_task_id = str(uuid.uuid4())
                 task.save()
+                task.log_lifecycle("waiting")
 
             if rampart_group is not None:
                 self.consume_capacity(task, rampart_group.name)
@@ -450,6 +456,7 @@ class TaskManager():
     def generate_dependencies(self, undeped_tasks):
         created_dependencies = []
         for task in undeped_tasks:
+            task.log_lifecycle("acknowledged")
             dependencies = []
             if not type(task) is Job:
                 continue
@@ -489,11 +496,18 @@ class TaskManager():
 
     def process_pending_tasks(self, pending_tasks):
         running_workflow_templates = set([wf.unified_job_template_id for wf in self.get_running_workflow_jobs()])
+        tasks_to_update_job_explanation = []
         for task in pending_tasks:
             if self.start_task_limit <= 0:
                 break
-            if self.is_job_blocked(task):
-                logger.debug("{} is blocked from running".format(task.log_format))
+            blocked_by = self.job_blocked_by(task)
+            if blocked_by:
+                task.log_lifecycle("blocked", blocked_by=blocked_by)
+                job_explanation = gettext_noop(f"waiting for {blocked_by._meta.model_name}-{blocked_by.id} to finish")
+                if task.job_explanation != job_explanation:
+                    if task.created < (tz_now() - self.time_delta_job_explanation):
+                        task.job_explanation  = job_explanation
+                        tasks_to_update_job_explanation.append(task)
                 continue
             preferred_instance_groups = task.preferred_instance_groups
             found_acceptable_queue = False
@@ -539,7 +553,17 @@ class TaskManager():
                     logger.debug("No instance available in group {} to run job {} w/ capacity requirement {}".format(
                                  rampart_group.name, task.log_format, task.task_impact))
             if not found_acceptable_queue:
+                task.log_lifecycle("needs_capacity")
+                job_explanation = gettext_noop("This job is not ready to start because there is not enough available capacity.")
+                if task.job_explanation != job_explanation:
+                    if task.created < (tz_now() - self.time_delta_job_explanation):
+                        # Many launched jobs are immediately blocked, but most blocks will resolve in a few seconds.
+                        # Therefore we should only update the job_explanation after some time has elapsed to
+                        # prevent excessive task saves.
+                        task.job_explanation = job_explanation
+                        tasks_to_update_job_explanation.append(task)
                 logger.debug("{} couldn't be scheduled on graph, waiting for next cycle".format(task.log_format))
+        UnifiedJob.objects.bulk_update(tasks_to_update_job_explanation, ['job_explanation'])
 
     def timeout_approval_node(self):
         workflow_approvals = WorkflowApproval.objects.filter(status='pending')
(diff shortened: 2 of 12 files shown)


fosterseth:
@gamuniz 

What are you thoughts on the information output to `/var/log/tower/job_lifecycle.log` (see above)

if there is additional information helpful for debugging purposes, we could include them.

gamuniz:
@fosterseth that looks really good and should be super helpful, this should be exposed to awx logger by default right?

awx/main/models/unified_jobs.py
```diff
@@ -397,6 +397,7 @@ def create_unified_job(self, **kwargs):
             # credentials and labels
             unified_job.save()
 
+        logger_job_lifecycle.info(f"{unified_job._meta.model_name}-{unified_job.id} created", extra={'type': unified_job._meta.model_name, 'template_name': unified_job.unified_job_template.name, 'job_id': unified_job.id, 'state': 'created'})
```
ryanpetrello: Could we make this a function on `model.UnifiedJob` and use it everywhere below e.g.,

```python
def log_lifecycle(self, status, **extra):
    ....

uj.log_lifecycle('created')
uj.log_lifecycle('blocked', blocked_by='blah blah blah')
```

awx/main/scheduler/task_manager.py
```diff
@@ -492,8 +498,14 @@ def process_pending_tasks(self, pending_tasks):
         for task in pending_tasks:
             if self.start_task_limit <= 0:
                 break
-            if self.is_job_blocked(task):
+            blocked_by = self.job_blocked_by(task)
+            if blocked_by:
                 logger.debug("{} is blocked from running".format(task.log_format))
+                logger_job_lifecycle.info(f"{task._meta.model_name}-{task.id} blocked by {blocked_by._meta.model_name}-{blocked_by.id}", extra={'type': task._meta.model_name, 'template_name': task.unified_job_template.name, 'job_id': task.id, 'state': 'blocked', 'blocked_by': f"{blocked_by._meta.model_name}-{blocked_by.id}"})
+                job_explanation = str.format(f"waiting for {blocked_by._meta.model_name}-{blocked_by.id} to finish")
+                if task.job_explanation != job_explanation and task.created < (tz_now() - timedelta(seconds=30)):
```
ryanpetrello: Yea, I think this is a reasonable approach to avoid a stampede of updates.

awx/main/scheduler/task_manager.py
```diff
@@ -539,6 +552,10 @@ def process_pending_tasks(self, pending_tasks):
                     logger.debug("No instance available in group {} to run job {} w/ capacity requirement {}".format(
                                  rampart_group.name, task.log_format, task.task_impact))
             if not found_acceptable_queue:
+                job_explanation = "This job is not ready to start because there is not enough available capacity."
+                if task.job_explanation != job_explanation and task.created < (tz_now() - timedelta(seconds=30)):
+                    task.job_explanation = "This job is not ready to start because there is not enough available capacity."
```
ryanpetrello: nitpick, but `task.job_explanation = job_explanation` no need to duplicate this string.

awx/main/scheduler/task_manager.py
```diff
@@ -539,6 +552,10 @@ def process_pending_tasks(self, pending_tasks):
                     logger.debug("No instance available in group {} to run job {} w/ capacity requirement {}".format(
                                  rampart_group.name, task.log_format, task.task_impact))
             if not found_acceptable_queue:
+                job_explanation = "This job is not ready to start because there is not enough available capacity."
```
ryanpetrello: This string should be wrapped in i18n `_(...)`

awx/main/tasks.py
```diff
@@ -1186,16 +1186,19 @@ def pre_run_hook(self, instance, private_data_dir):
         '''
         Hook for any steps to run before the job/task starts
         '''
+        logger_job_lifecycle.info(f"{instance._meta.model_name}-{instance.id} pre run", extra={'type': instance._meta.model_name, 'template_name': instance.unified_job_template.name, 'job_id': instance.id, 'state': 'pre_run'})
```
ryanpetrello: I like the use of the hooks for this 👍

ryanpetrello: Instead of `'job_id'` could we use `self.event_data_key` for accuracy for all types of "jobs"?

awx/main/tasks.py
```diff
@@ -1186,16 +1186,19 @@ def pre_run_hook(self, instance, private_data_dir):
...
 
     def post_run_hook(self, instance, status):
         '''
         Hook for any steps to run before job/task is marked as complete.
         '''
+        logger_job_lifecycle.info(f"{instance._meta.model_name}-{instance.id} post run", extra={'type': instance._meta.model_name, 'template_name': instance.unified_job_template.name, 'job_id': instance.id, 'state': 'post_run'})
```
ryanpetrello: Same comment as above here.

awx/main/tasks.py
```diff
@@ -1186,16 +1186,19 @@ def pre_run_hook(self, instance, private_data_dir):
...
 
     def final_run_hook(self, instance, status, private_data_dir, fact_modification_times, isolated_manager_instance=None):
         '''
         Hook for any steps to run after job/task is marked as complete.
         '''
+        logger_job_lifecycle.info(f"{instance._meta.model_name}-{instance.id} finalize run", extra={'type': instance._meta.model_name, 'template_name': instance.unified_job_template.name, 'job_id': instance.id, 'state': 'finalize_run'})
```
ryanpetrello: (and here)

awx/main/scheduler/task_manager.py
```diff
@@ -492,8 +498,14 @@ def process_pending_tasks(self, pending_tasks):
         for task in pending_tasks:
             if self.start_task_limit <= 0:
                 break
-            if self.is_job_blocked(task):
+            blocked_by = self.job_blocked_by(task)
+            if blocked_by:
                 logger.debug("{} is blocked from running".format(task.log_format))
+                logger_job_lifecycle.info(f"{task._meta.model_name}-{task.id} blocked by {blocked_by._meta.model_name}-{blocked_by.id}", extra={'type': task._meta.model_name, 'template_name': task.unified_job_template.name, 'job_id': task.id, 'state': 'blocked', 'blocked_by': f"{blocked_by._meta.model_name}-{blocked_by.id}"})
+                job_explanation = str.format(f"waiting for {blocked_by._meta.model_name}-{blocked_by.id} to finish")
```
ryanpetrello: Do we need the `str.format` since we're using an f-string?  Let's also wrap this string in i18n `_()` function call so it gets marked for translation.

awx/settings/defaults.py
```diff
@@ -940,6 +943,15 @@ def IS_TESTING(argv=None):
             'backupCount': 5,
             'formatter':'simple',
         },
+        'job_lifecycle': {
+            'level': 'INFO',
+            'class':'logging.handlers.RotatingFileHandler',
```
ryanpetrello: You might want to sync up w/ @chrismeyersfsu on his impending logrotate work, as he's been planning on removing these rotating file handlers.

ryanpetrello (changes requested):
This is coming along nicely - I've got some minor feedback.

awx/main/tasks.py
fosterseth: thinking about parsing this json -- would we want a fixed key here across all entries? maybe it could be named `task_id` to be more generic

awx/main/scheduler/task_manager.py
fosterseth: good catch thanks. I also realize I need to unset/change this job_explanation when we start the task

awx/main/scheduler/task_manager.py
fosterseth: I am not sure if this is still relevant https://github.com/ansible/awx/blob/dfa65225d90af717b5011580a4ed35d376dd72e0/awx/api/serializers.py#L806

does this mean we shouldn't mark job_explanation for translation?

ryanpetrello: Hey, good catch!  Yep, I think you're right about this.

ryanpetrello: Well.

ryanpetrello: I think in this case we might *actually* want to wrap this in `ugettext_noop` (not `ugettext_lazy`).  This will *mark* the string for translation (so it shows up in our generated i18n files), but it won't *actually* translate it here.

https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.translation.ugettext_noop

awx/main/tasks.py
ryanpetrello: Yea, that's a good point.  If we want to stick with `job_id`, that probably makes sense considering I think there's additional context in here that tells you what *type* of job it is.

awx/main/scheduler/task_manager.py
ryanpetrello: Yes, good call.  We may want to make sure that when a job goes into `waiting`, we unset this explicitly.

fosterseth:
image

in development environment, job_lifecycle logs can be blue, if we want that

ryanpetrello:
> in development environment, job_lifecycle logs can be blue, if we want that

I like it 😄

awx/main/models/unified_jobs.py
```diff
@@ -1484,3 +1489,17 @@ def is_isolated(self):
...
         return False
+
+    def log_lifecycle(self, state, blocked_by=None):
+        extra={'type': self._meta.model_name,
+               'task_id': self.id,
+               'state': state}
+        if self.unified_job_template:
+            extra["template_name"] = self.unified_job_template.name
+        if state == "blocked" and blocked_by:
+            blocked_by_msg = f"{blocked_by._meta.model_name}-{blocked_by.id}"
```
ryanpetrello, line 1500: We've got a property that's pretty similar to this that we use in other logging contexts:

`UnifiedJob.log_format`

It also includes the current status of the job.  If that's too verbose, that's fine, but it might be an opportunity for reuse here.

ryanpetrello:
```python
In [1]: Job.objects.first().log_format
Out[1]: 'job 1 (successful)'```

Comment (ryanpetrello) [reply awx/main/models/unified_jobs.py:1500]:
```python
In [4]: [j.log_format for j in UnifiedJob.objects.all()]
Out[4]:
['job 1 (successful)',
 'project_update 2 (successful)',
 'ad_hoc_command 3 (successful)',
 'ad_hoc_command 4 (successful)',
 'ad_hoc_command 5 (successful)',
 'job 6 (successful)',
 'project_update 7 (successful)']```

Comment (ryanpetrello) [reply awx/main/models/unified_jobs.py:1500]: You know @fosterseth ignore me, I like the way yours looks better.

File: awx/settings/defaults.py (hunk 9 above)
Comment (ryanpetrello) [reply awx/settings/defaults.py]: This definitely needs an update now that @chrismeyersfsu has merged his log rotation PR, but I don't expect anything super complicated.

**Review by ryanpetrello — APPROVED** [review]
Really great work on this - I like how it turned out, and this is going to be really useful to figure out why jobs aren't running yet.

I'd say once we've tossed this through an integration run, I'm a 👍 on merging this.

**Comment by ryanpetrello** [comment]
@fosterseth,

@jakemcdermott had mentioned to me the possibility of folding his work on https://github.com/ansible/awx/pull/8726 *into* this PR and verifying and landing them around the same time.  Is this something you two have discussed as an option?

**Comment by fosterseth** [comment]
I've gone back and forth on whether we should provide a `waiting_on` field that would identify the job that is causing the block. We could then provide that job as a link in the ui, which I think would be really nice
```python
    waiting_on = models.ForeignKey(
        'self',
        null=True,
        editable=False,
        related_name='+',
        on_delete=polymorphic.SET_NULL,
    )
```
e.g.
image

Currently, we can stuff that info into the `job_explanation`, but it won't be a link in the UI or anything.

Because waiting_on could be a projectupdate/job/inventorysource, we would also need another field in the json that identifies the type of resource, so that we can get the appropriate link.

@jakemcdermott  and I brainstormed some ideas how to display this link. One thought was, if waiting_on is null, then just display job_explanation. If not null, then have a predefined string with the link to the waiting_on resource (e.g. "waiting on <project update 123>"). The downside of this is that there might be case we want job_explanation information to take a priority over waiting_on information, so having these two fields compete for the same space in the UI is not ideal. Having a separate "Waiting On" field in the UI solves this, but then again it's yet another field on the detail page, introducing more clutter.

Maybe we should punt the UI link idea until later, and settle with the generic text in the job_explanation for now?

ryanpetrello:
@fosterseth I think the current implementation based on just the `job_explanation` changes in @jakemcdermott's PR is probably fine for a first mergeable pass on this work - we can always scope out improvements down the line 😄

jakemcdermott (approved):

ryanpetrello:
@fosterseth before we merge this, can we squash a few commits (or at least, improve a few of the latter commit messages)?

fosterseth:
@ryanpetrello Yep I plan to squash them all into one

fosterseth:
required tower packaging changes https://github.com/fosterseth/tower-packaging/tree/add_log_rotate_job_lifecycle

fosterseth:
states for all jobs

- created
- acknowledged
- waiting
- pre run
- preparing playbook
- running playbook
- post run
- finalize run

special processing for some jobs

- start job fact cache
- finish job fact cache
- event processing cache
- notifications sent

stuck in pending
- blocked by
- needs capacity

awx/main/scheduler/task_manager.py
```diff
@@ -489,11 +494,18 @@ def generate_dependencies(self, undeped_tasks):
...
+        tasks_to_update_job_explanation = []
         for task in pending_tasks:
             if self.start_task_limit <= 0:
                 break
-            if self.is_job_blocked(task):
-                logger.debug("{} is blocked from running".format(task.log_format))
+            blocked_by = self.job_blocked_by(task)
+            if blocked_by:
+                task.log_lifecycle("blocked", blocked_by=blocked_by)
+                job_explanation = gettext_noop(f"waiting for {blocked_by._meta.model_name}-{blocked_by.id} to finish")
```
elyezer, line 506: Is it possible to make the blocked by job name to become a link in the UI so we can easily navigate to it and follow if we want to?

fosterseth: exactly, we plan to make this a link in the UI, which would be helpful to users

fosterseth:
was curious how big this file gets -- here it is after running the yolo tests

```
[root@ip-10-0-4-116 tower]# ls -lh
total 1.7M
-rw-r-----. 1 awx  awx   52K Feb  3 02:58 callback_receiver.log
-rw-r-----. 1 awx  awx   99K Feb  3 02:58 dispatcher.log
-rw-r-----. 1 awx  awx     0 Feb  3 01:55 isolated_manager.log
-rw-r-----. 1 awx  awx  667K Feb  3 02:58 job_lifecycle.log
-rw-r-----. 1 awx  awx     0 Feb  3 01:55 management_playbooks.log
drwxr-xr-x. 2 awx  awx     6 Feb  3 02:14 playbook_profiling
-rw-------. 1 root root 658K Feb  3 02:11 setup-2021-02-03-01:50:34.log
-rw-r-----. 1 awx  awx   32K Feb  3 02:58 task_system.log
-rw-r-----. 1 awx  awx  212K Feb  3 03:05 tower.log
-rw-r-----. 1 awx  awx     0 Feb  3 01:55 tower_rbac_migrations.log
-rw-r-----. 1 awx  awx     0 Feb  3 01:55 tower_system_tracking_migrations.log
-rw-r-----. 1 awx  awx     0 Feb  3 01:55 wsbroadcast.log
```

ryanpetrello (approved):

PR #9143, opened 2021-01-22
expose active host count on the Organization detail view

ryanpetrello:


jbradberry (approved):
I think this will probably be ok.

wenottingham (approved):

termlen0:
Pasting below, feeback for this issue:
cc @wenottingham 
```
We’d like Ansible Tower to have a view that shows the hosts that a specific organization or organizations in Tower with no activity (playbooks, ad hoc commands, etc) for X period of time (where X can be defined). 

 

As a concrete example to help understand that request:  We need to understand if an Org has, say, 100 host licenses, and they only have used 10 of them regularly in the last 6 months.  This will help us with the follow-up/pro-active conversation with those teams to:

 

    determine if the licenses are simply not needed anymore (if they’re not, they’re burning money and we can repurpose them to another team who could use them).
    determine if they’re going unused because they are having challenges creating building automation.  In that case, our team can provide assistance to them so they can benefit```

PR #9145, opened 2021-01-22
WIP Inclusion of the EE option in the payloads within the Organization and Projects.

yagomarques:
##### SUMMARY
We are implementing new tests for Execution Environment. For do that we need to use the factories for create organizations and projects with Execution Environment

##### ISSUE TYPE
 - https://github.com/ansible/awx/issues/7064


awxkit/awxkit/api/pages/organizations.py
```diff
@@ -67,3 +79,4 @@ class Organizations(page.PageList, Organization):
 page.register_page([resources.organizations,
                     resources.user_organizations,
                     resources.project_organizations], Organizations)
+
```
kdelee: please remove this extra new line

AlanCoding:
this does look necessary, but maybe the base branch was rebased.

Also, I would like to keep work on the `shanemcd:execution-environments` branch, because that's where most of the focus has been lately, and where we're likely to run tests from.

kdelee:
@AlanCoding you don't say so! just got it fixed as far as when pointed to the ansible fork execution environments branch

can we merge it here or does it need to go into shane's first?

awxkit/awxkit/api/pages/organizations.py
```diff
@@ -43,6 +43,18 @@ def remove_galaxy_credential(self, credential):
     def payload(self, **kwargs):
         payload = PseudoNamespace(name=kwargs.get('name') or 'Organization - {}'.format(random_title()),
                                   description=kwargs.get('description') or random_title(10))
+        
+        for fk_field in ('inventory', 'default_environment'):
```
kdelee: why do you have this `inventory` item in this list? I think you need to remove that

awxkit/awxkit/api/pages/projects.py
```diff
@@ -217,3 +227,4 @@ class Playbooks(base.Base):
 
 
 page.register_page(resources.project_playbooks, Playbooks)
+
```
kdelee: please remove extra new line

AlanCoding:
It's possible that we can merge this here and rebase that other branch, I don't want to be particular about this. @shanemcd

awxkit/awxkit/api/pages/projects.py
```diff
@@ -43,6 +43,16 @@ def payload(self, organization, scm_type='git', **kwargs):
             'allow_override')
         update_payload(payload, fields, kwargs)
 
+        for fk_field in ('execution_environment', 'default_environment'):
```
AlanCoding, line 46: do projects take _both_ of these fields??

yagomarques: the projects api has this field, and thinking now we also need to modify the [job_template](https://github.com/ansible/awx/blob/execution-environments/awxkit/awxkit/api/pages/job_templates.py#L105), as it's the same case. What do you think @AlanCoding @kdelee?

rebeccahhh: I believe it's on my to do list to modify unified_jobs

awxkit/awxkit/api/pages/organizations.py
```diff
@@ -39,10 +39,20 @@ def remove_galaxy_credential(self, credential):
                 "disassociate": True,
             })
 
-
     def payload(self, **kwargs):
         payload = PseudoNamespace(name=kwargs.get('name') or 'Organization - {}'.format(random_title()),
                                   description=kwargs.get('description') or random_title(10))
+        
+        for fk_field in ('default_environment'):
```
yagomarques: I found a bug, I need to fix it before Merge. Basically, when I have only one parameter, it doesn't execute the action successfully. @kdelee

shanemcd:
I'm fine with merging here and rebasing my branch, after we get tests passing of course.

awxkit/awxkit/api/pages/projects.py
AlanCoding: My question is more basic here. Look at the help text for the two fields

`execution_environment`

> The container image to be used for execution.

`default_environment`

> The default execution environment for jobs run using this project.

What is the intent here? Is it that project updates will run in the `execution_environment` container? Because if it is, that's going to be a problem.

We don't allow customizing the custom virtual environment of project updates now (only that of jobs that use the project), and we won't allow customizing the execution environment. Our discussions of the OpenShift solution for project updates are completely incompatible with letting users select the EE to run a _project update_ in. This question is mostly directed at @jbradberry, but I also don't want to add defunct fields here because it indicates an intent to test them.

AlanCoding: I filed https://github.com/shanemcd/awx/pull/27

So this awxkit code should only need the `default_environment` field. However, the other field should not do any harm here.

jbradberry: @yagomarques job templates do not have `default_environment`.

jbradberry: @AlanCoding the `execution_environment` field is brought in by `Project` being a subclass of `UnifiedJobTemplate`.  If we want to prevent any kind of customization for project updates, we are going to have to override the EE resolver for projects.  Also, we're going to want to either prevent this field from showing up in the UI and API serializer, or we'll have to break up the usage of `ExecutableEnvironmentMixin` to the explicit job template and job types.  This might also have implications due to the field not being on the common parent table, I haven't thought that part through yet.

awxkit/awxkit/api/pages/organizations.py
jbradberry: You need a comma for it to be a tuple.  `for fk_field in ('default_environment',):`

awxkit/awxkit/api/pages/projects.py
AlanCoding: > the execution_environment field is brought in by Project being a subclass of UnifiedJobTemplate. If we want to prevent any kind of customization for project updates, we are going to have to override the EE resolver for projects. Also, we're going to want to either prevent this field from showing up in the UI and API serializer

Agreed, this is what my linked PR does and I'm satisfied with the implications of it.

> or we'll have to break up the usage of ExecutableEnvironmentMixin to the explicit job template and job types

Or leave it where it is, and let it be `None` on the `Project` model and don't touch it. It could still be edited in the ORM, but you could lock down editing it if you were really worried.

There are other cases where a field on the UJT model does not apply to all job types.

jbradberry: @AlanCoding What linked PR?  I don't see that you have a PR linked anywhere in here.

AlanCoding: https://github.com/shanemcd/awx/pull/27

jbradberry: ... why the heck did I think that was an issue instead of a PR?  :confused:

AlanCoding (approved):

jbradberry:
@yagomarques I almost never say this, but I think that you need to squash or otherwise clean up your commit history here.

AlanCoding:
got some lint errors

```
2021-01-26 18:34:50.140766 | static | ./awxkit/awxkit/api/pages/organizations.py:67:1: E305 expected 2 blank lines after class or function definition, found 1
2021-01-26 18:34:50.141072 | static | ./awxkit/awxkit/api/pages/projects.py:229:59: W292 no newline at end of file
2021-01-26 18:34:50.185792 | static | make: *** [Makefile:319: flake8] Error 1
```

yagomarques:
> @yagomarques I almost never say this, but I think that you need to squash or otherwise clean up your commit history here.

Issue #4426, opened 2019-08-05
Add RBAC rules to side-nav


mabashian:
##### SUMMARY
The side-nav in `ui_next` does not have any RBAC rules baked in yet.  This issue will be used to track that work.  https://github.com/ansible/awx/blob/devel/awx/ui/client/lib/components/layout/layout.partial.html can be used as a reference.  In general, we want to do the following:

1. Hide Settings, Management Jobs, and Credential Types from _all users except system administrators and system auditors_.  Note that these two flags come from `/api/v2/me` (is_superuser, is_system_auditor).
2. Hide Instance Groups and Integrations (Applications) from _all users except system administrators, system auditors, and org administrators_.  Determining whether a user is an org admin is done with a request to `/api/v2/users/:user_id/admin_of_organizations/` and checking that the count is greater than one.
3. Hide Notifications from _all users except system administrators, system auditors, org administrators, and notification administrators_.    Determining whether a user is an org admin is done with a request to `api/v2/organizations/?role_level=notification_admin_role` and checking that the count is greater than one.

If none of the `Administration` or `Settings` options are shown to a user then those headers should be hidden as well.

AlexSCorey:
backing out for now

jakemcdermott:
I would like to discuss with the api team the possibility of adding an endpoint (or more information to an existing API) to help solve this, as I believe it would reduce a lot of complexity and http calls.

unlikelyzero:
@nixocio as I understand it, we will need to add some api changes to the /me and /config API endpoints. Is that correct @jakemcdermott ?

nixocio:
It would be a nice to have, but not a blocker. I was waiting for the wizard PR work to be completed to move forward with this one.

nixocio:
There is now also the execution environment admin that needs to be taken in consideration for the side-nav.

tiagodread:
This is now fixed

PR #9159, opened 2021-01-25
Add RBAC rules to side-nav

nixocio:
Add RBAC rules to side-nav.

See: https://github.com/ansible/awx/issues/4426

https://user-images.githubusercontent.com/9053044/105753752-2c79d300-5f17-11eb-858c-c28ef5bc97f3.mov

### DIFF
--- a/awx/ui_next/src/App.jsx
+++ b/awx/ui_next/src/App.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import {
   useRouteMatch,
   useLocation,
@@ -13,6 +13,7 @@ import AppContainer from './components/AppContainer';
 import Background from './components/Background';
 import NotFound from './screens/NotFound';
 import Login from './screens/Login';
+import { MeAPI, UsersAPI, OrganizationsAPI } from './api';
 
 import ja from './locales/ja/messages';
 import en from './locales/en/messages';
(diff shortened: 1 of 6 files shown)


unlikelyzero:
@jakemcdermott do you have a link to the api change to reduce the number of requests based on the changes to the /me/config endpoint?

awx/ui_next/src/App.jsx
```diff
@@ -39,6 +46,48 @@ function App() {
   const match = useRouteMatch();
   const { hash, search, pathname } = useLocation();
 
+  const fetchUserData = async () => {
```
keithjgrant, line 49: This adds a lot of logic to the App component... I'm wondering if it can be moved down into a new component instead (maybe something wrapping `<Login>`? We did a lot of work last year to keep these top level components nice and streamlined, and I'd hate to lose some of that

nixocio: hey @keithjgrant, thanks for your comment. I did think about a context that could be updated by the login component, and store information about the logged user. This could avoid de-duplicate a few requests inside the AppContainer that check the same data. The drawback is that we will have one of more context. In this case we will probably need to wrap the whole application on this new context.  I will explore the idea to move logic to the login component. If you have a specific suggestion, just let me know.

keithjgrant: Yeah, I'd avoid a new context I think. I'd see what would happen you either move it into tho Login component, or create a new component that wraps the Login component to encapsulate that logic and expose the callbacks props needed to interact with it

tiagodread:
Waiting on API change cc @jakemcdermott

mabashian:
At the moment I believe this is blocked by #9908 @nixocio is that correct?  If not feel free to remove the blocked label.

nixocio:
Closing this one in favor of https://github.com/ansible/awx/pull/10249

Issue #7676, opened 2020-07-20
Add tooltips to inventory sync status icons on Inventory List


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
<img width="1435" alt="Screen Shot 2020-07-20 at 2 32 14 PM" src="https://user-images.githubusercontent.com/9889020/87973093-0241e700-ca96-11ea-9643-da37af22e04b.png">

Currently hovering over these icons does nothing.  In the old app we showed these tooltips:

<img width="366" alt="Screen Shot 2020-07-20 at 2 35 09 PM" src="https://user-images.githubusercontent.com/9889020/87973358-62d12400-ca96-11ea-9896-8e606328d14e.png">
<img width="433" alt="Screen Shot 2020-07-20 at 2 35 14 PM" src="https://user-images.githubusercontent.com/9889020/87973361-6369ba80-ca96-11ea-90ae-08ccf7d86cf9.png">
<img width="413" alt="Screen Shot 2020-07-20 at 2 35 23 PM" src="https://user-images.githubusercontent.com/9889020/87973363-6369ba80-ca96-11ea-8d01-f8ea4ccb7b26.png">

We probably want to do something similar

nixocio:
In the old ui, we do have tooltips and popovers. Tooltips being displayed when hovering over the cloud icon, and a popover being displayed once the cloud icon is clicked.

<img width="1565" alt="image" src="https://user-images.githubusercontent.com/9053044/100005127-78cd5900-2d96-11eb-926d-4c2fb8e24bfe.png">

tiagodread:
PR already merged, tests in place,  we can safely close this issue after merging https://github.com/ansible/tower-qa/pull/5967

PR #9170, opened 2021-01-26
Add tooltips to inventory sync status

nixocio:
Add tooltips to inventory sync status on Inventory List.

See: https://github.com/ansible/awx/issues/7676

image

image


jakemcdermott (approved):

tiagodread (approved):
Ok, verified that the tooltip renders for the  different status:
- not configured inventory for sync
- no inventory sync failures
- one or more sourced sync with failures

PR #9172, opened 2021-01-26
Make sure project updates run in default EE

AlanCoding:
replaces https://github.com/shanemcd/awx/pull/27 because base branch moved

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -1394,7 +1394,7 @@ class ProjectSerializer(UnifiedJobTemplateSerializer, ProjectOptionsSerializer):
 
     class Meta:
         model = Project
-        fields = ('*', 'organization', 'scm_update_on_launch',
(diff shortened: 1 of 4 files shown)


AlanCoding:
I rebuilt the UI, and navigated around. I'm not seeing any console errors.

It looks like we have not yet hooked up the fields for JT / project

@nixocio, does that sound right to you? Can you anticipate any fallout from removing this field from the API?

jbradberry (approved):
Still looks good to me, but per my comment elsewhere, we need to make sure that the UI also loses this field for Projects.

AlanCoding:
I can't find it anywhere in the UI, navigating to the project form, neither field is present. It looks like the UI has only added the execution environments list so-far.

nixocio:
hey @AlanCoding, that seems right.

nixocio:
Also we are tracking the work related to EE and UI on this issue, https://github.com/ansible/awx/issues/7884, if we are missing something, please file an issue.

Issue #9109, opened 2021-01-19
A pull option for execution environments


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Scenario:

I have an execution environment that targets some `:latest` image on quay.io. The image has a bug, I run a job, and it fails. The bug is fixed, and someone uploads a new version of that imate.

AWX needs a way for the user to force the worker node to refresh the image from the registry.

Earlier we discussed a `pull` field for the `ExecutionEnvironment` model (@jbradberry started on the model change), but right now I don't see mention of this in any other existing issues.

This would be an important feature gap, but it's not necessarily planned for the initial roll-out of execution environments.

shanemcd:
I thought we had this on the EE branch? Did it get removed at some point? I would love to see this in our initial effort.

jbradberry:
> I thought we had this on the EE branch? Did it get removed at some point? I would love to see this in our initial effort.

Uh, you removed it.

AlanCoding:
Let's talk about the full implementation.

https://github.com/ansible/ansible-runner/blob/3d4a6c2ac42bdf0669feed19cdd6f9800ec9f57d/ansible_runner/runner_config.py#L846-L847

Maaaybe this would be achieved by the ansible-runner setting `container_options=['--pull']`?

EDIT: this was wrong, keep reading...

shanemcd:
@jbradberry Wasn't it in the wrong place?

Edit: I remember now: your original implementation put it on the mix-in, which we decided against. We still need to add this option at the EE level.

shanemcd:
I remember now: your original implementation put it on the mix-in, which we decided against. We still need to add this option at the EE level.

shanemcd:
@rebeccahhh This should be pretty easy to add. Can you look into this if you have time this week?

shanemcd:
@AlanCoding According to the man pages for podman/docker:

```
--pull string                    Pull image before running ("always"|"missing"|"never") (default "missing")
```

So I think we would want `--pull=always`.

AlanCoding:
wouldn't you want to give all the options? What if someone wanted to populate the cache manually and use "never" for security reasons?

To set expectations of how this operates, I experimented with this locally:

```
time podman run --rm --tty --interactive quay.io/ansible/ansible-runner:devel whoami
time podman run --rm --tty --interactive quay.io/ansible/ansible-runner:devel whoami --pull=always
[...]

PR #9179, opened 2021-01-27
add container options to EE model

rebeccahhh:
##### SUMMARY

Add pull options to the EE model, there will be more that needs to be done I would imagine to make this useful. Should address #9109 

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 17.0.1
```


awx/ui_next/src/screens/ExecutionEnvironment/ExecutionEnvironmentList/ExecutionEnvironmentList.jsx
```diff
@@ -147,7 +147,7 @@ function ExecutionEnvironmentList({ i18n }) {
                 showSelectAll
                 isAllSelected={isAllSelected}
                 onSelectAll={isSelected =>
-                  setSelected(isSelected ? [...executionEnvironments] : [])
+                  setSelected(isSelected ? [...s] : [])
```
shanemcd: Were the changes to these js files intentional?

awx/api/metadata.py
```diff
@@ -141,7 +141,7 @@ def get_field_info(self, field):
                 pass
         if field.field_name == 'type':
             field_info['type'] = 'choice'
-        elif field.field_name in ('url', 'custom_virtualenv', 'token'):
+        elif field.field_name in ('url', 'token'): #'custom_virtualenv',
```
shanemcd: I'd lean towards just deleting code rather than leaving behind stuff commented out. The diff will be more clear that way too.

ryanpetrello: Not to mention, deleting code is _extremely_ cathartic **and** it aligns with an admirable personal goal, which is to leave this world having deleted more code than you added.

awx/ui_next/src/screens/ExecutionEnvironment/ExecutionEnvironmentList/ExecutionEnvironmentList.jsx
rebeccahhh: oh wow, it was definitely not, sorry about that

awx/api/metadata.py
rebeccahhh: do we want to delete it permanently though? or keep it while people are migrating?

shanemcd: Followed up in DM but also posting here for posterity:

> We need to be able to tell on the backend if something had a custom venv so that we know if they need to specify a custom EE.
> We dont need to expose it in the API anymore, so it’s ok to go ahead and remove it from any serializers.

shanemcd: Reading through this diff a bit more... We might end up needing _most_ of the code commented out here. We can definitely get rid of any references to custom venvs in the serializers / API, though. We just want to keep any resolving logic, so that we can prompt users in the appropriate situations.

rebeccahhh: I pulled all changes out _except_ for the relevant model change. Since I'm not fully caught up on everything here I'm quite sure this change alone will not be enough and I'll need to hook in some more changes but this is a better, simpler place to start.

awx/main/models/execution_environments.py
```diff
@@ -35,6 +41,11 @@ class Meta:
         default=None,
         on_delete=models.SET_NULL,
     )
+    container_options = models.CharField(
```
ryanpetrello: This change probably also needs a migration file so the database schema gets updated on upgrade - you should be able to generate one via `awx-manage makemigrations`.

awx/main/migrations/0126_executionenvironment_container_options.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('main', '0125_more_ee_modeling_changes'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='executionenvironment',
+            name='container_options',
+            field=models.CharField(choices=[('always', 'Always pull container before running.'), ('missing', 'No pull option has been selected'), ('never', 'Never cull container before running')], default='missing', help_text='Pull image before running?', max_length=1024),
```
shanemcd:
```suggestion
            field=models.CharField(choices=[('always', 'Always pull container before running.'), ('missing', 'No pull option has been selected'), ('never', 'Never pull container before running')], default='missing', help_text='Pull image before running?', max_length=1024),
```

rebeccahhh: that would have meant something very different 😅

rebeccahhh:
recheck

AlanCoding:
@rebeccahhh can you rebase? Sorry I got to this late.

jakemcdermott:
sup

ui here: https://github.com/ansible/awx/pull/9179/commits/b74bc3054fd1c63f0fea5bdc1e82dd3126521b5a

jakemcdermott:
eefields

jakemcdermott:
recheck

jakemcdermott:
recheck

awx/ui_next/src/screens/ExecutionEnvironment/shared/ExecutionEnvironmentForm.jsx
```diff
@@ -37,8 +47,28 @@ function ExecutionEnvironmentFormFields({ i18n, me, executionEnvironment }) {
...
+
   return (
     <>
+      <FormField
+        id="execution-environment-name"
+        label={i18n._(t`Name`)}
+        name="name"
+        type="text"
+        validate={required(null, i18n)}
+        isRequired
```
nixocio, line 70: Not sure if name will be a required field on the API side.

awx/main/models/execution_environments.py
```diff
@@ -12,6 +12,12 @@ class ExecutionEnvironment(CommonModel):
     class Meta:
         ordering = ('-created',)
 
+    PULL_CHOICES = [
+        ('always', _("Always pull container before running.")),
+        ('missing', _("No pull option has been selected")),
```
nixocio:
```suggestion
        ('missing', _("No pull option has been selected.")),
```

awx/main/models/execution_environments.py
```diff
@@ -12,6 +12,12 @@ class ExecutionEnvironment(CommonModel):
...
+        ('never', _("Never pull container before running"))
```
nixocio:
```suggestion
        ('never', _("Never pull container before running."))
```

awx/ui_next/src/screens/ExecutionEnvironment/shared/ExecutionEnvironmentForm.jsx
rebeccahhh: I can make it one, and in fact think it's prudent to have it be

rebeccahhh:
@jakemcdermott we elected to change "container options" to instead simply be "pull"
I believe I got everything in the UI that is affected by the change but if I was a bull in a china shop please forgive me ^_^

shanemcd (approved):

jakemcdermott (approved):

awx/ui_next/src/screens/ExecutionEnvironment/shared/ExecutionEnvironmentForm.jsx
```diff
@@ -37,19 +47,58 @@ function ExecutionEnvironmentFormFields({ i18n, me, executionEnvironment }) {
...
       />
+      <FormGroup
+        fieldId="execution-environment-container-options"
+        helperTextInvalid={containerOptionsMeta.error}
+        validated={
+          !containerOptionsMeta.touched || !containerOptionsMeta.error
+            ? 'default'
+            : 'error'
+        }
+        label={i18n._(t`Container Pull Option`)}
```
jakemcdermott:
```suggestion
        label={i18n._(t`Pull`)}
```

(To make it consistent w/ the Details panel)

PR #9185, opened 2021-01-28
Add orgs job launch

sean-m-sullivan:
##### SUMMARY
add orgainzation lookup to job launch module to be in line with other modules

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - awx-collection

##### AWX VERSION
17.0.1


john-westcott-iv:
This field used to be useful before we had the lookup plugin. IMHO now that we have the lookup plugin I'd rather have people use that to filter out correct templates instead of having an organization option that is only used to assist with lookups. How do other people feel about this?

sean-m-sullivan:
I can see the argument for the lookup plugin, was more going through the modules to use them in roles, and noticed that this was the one without the lookup in it. I'd vote for keeping the status quo with how we've been doing it, but do see the merits of forcing users to use the lookup plugin.

awx_collection/test/awx/test_job_template.py
```diff
@@ -85,6 +85,7 @@ def test_job_launch_with_prompting(run_module, admin_user, project, inventory, m
     JobTemplate.objects.create(
         name='foo',
         project=project,
+        organization='Default',
```
beeankha: If you add `organization` as an argument for this test and then `organization=organization,` on this line, this test (and thus CI) should pass.

beeankha (approved):

AlanCoding (approved):

beeankha:
regate

PR #9186, opened 2021-01-28
Update tower_instance_group.py

axil76:
##### SUMMARY
Add the is_containerized and summary_fields parameters for the configation import containing group containers

### DIFF
--- a/awx_collection/plugins/modules/tower_instance_group.py
+++ b/awx_collection/plugins/modules/tower_instance_group.py
@@ -92,6 +92,8 @@ def main():
         pod_spec_override=dict(),
         instances=dict(required=False, type="list", default=None),
         state=dict(choices=['present', 'absent'], default='present'),
+        is_containerized=dict(type='bool'),
+        summary_fields=dict(),
     )
 
     # Create a module for ourselves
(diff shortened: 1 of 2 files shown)


awx_collection/plugins/modules/tower_instance_group.py
```diff
@@ -92,6 +92,8 @@ def main():
         pod_spec_override=dict(),
         instances=dict(required=False, type="list", default=None),
         state=dict(choices=['present', 'absent'], default='present'),
+        is_containerized=dict(type='bool'),
```
ryanpetrello, line 95: cc @AlanCoding @shanemcd (is this field name changing soon?)

AlanCoding:
These are read-only fields. Because of that, I think the argument_spec is the wrong place for them.

Can you give more details about your use case, preferably by sharing a playbook?

I appreciate that the returned result from a module should contain read-only fields like summary_fields. This may be fine in the top-level keys, or they may need to be in some other place, or we may have to direct people to use the lookup plugin. It just depends and we need more info about the use case.

awx_collection/plugins/modules/tower_instance_group.py
AlanCoding: Checking the current API in `devel` branch

```json
            "capacity": 298,
            "committed_capacity": 0,
            "consumed_capacity": 0,
            "percent_capacity_remaining": 100.0,
            "jobs_running": 0,
            "jobs_total": 0,
            "instances": 1,
            "controller": null,
            "is_controller": false,
            "is_isolated": false,
            "is_container_group": false,
            "credential": null,
            "policy_instance_percentage": 100,
            "policy_instance_minimum": 0,
            "policy_instance_list": [],
            "pod_spec_override": "",
```

(for instance group)

So you are right @ryanpetrello, this field no longer exists

AlanCoding:
I can discuss further with you how these modules expect read-only fields should be referenced in playbooks, but I don't see a way forward for this patch.

PR #9084, opened 2021-01-14
Extend launched by details to handle scheduled JTs and WFJTs

marshmalien:
##### SUMMARY
Issue: https://github.com/ansible/awx/issues/5057

Launched By detail should link to the schedule that launched the job. 

Screen Shot 2021-01-14 at 11 41 49 AM

##### ISSUE TYPE

 - Enhancement Pull Request

##### COMPONENT NAME

 - UI

PR #9201, opened 2021-02-01
Properly set launched_by for jobs launched by scheduled workflows : Issue 5057

### LINKED ISSUE #5057: Fix "Launched By" for JTs in scheduled workflows
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Background issue for https://github.com/ansible/awx/issues/5048.

The UI includes a "Launched By" tag to denote how something was launched.
It is propagated by some combination of:
- created_by (for the launching user)
- related schedule (for scheduled jobs)

This means you get the following:
- JT launched manually: user who launched job
- JT launched from schedule: schedule that launched job
- JT launched from a workflow manually: user who launched workflow
- Workflow launched manually: user who launched workflow
- Workflow launched from schedule: schedule that launched workflow
- **JT launched from a workflow launched by a schedule: nothing!**

For now, we're just not going to show the field in that final case, but we should propagate something via the API to give useful data here.

amolgautam25:
##### SUMMARY

with respect to the issue 5057, there should be a way by which the UI should know , if a job was launched from a workflow , and that workflow was launched from a schedule.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 17.0.1
```

##### ADDITIONAL INFORMATION

```
This PR tries to solve the issue 5057. 
There is an edge case , where the API response is not very clear. When a JT is invoked from a workflow , and that workflow is invoked from a schedule, the API response does not convey how the job was launched. 

So, I have added the schedule id of the schedule that invoked the workflow and which in turn invokes the JT. The new key in API response is 'launched_by', and it has the schedule id. In all  the 5 previous cases mentioned in the initial issue , the 'launched_by' field is blank ( that is empty string ( "" )).

 
```


jbradberry (commented):
I'm afraid that this breaks cases that already work.  This code needs to be able to return a user when the launch type is manual, relaunch(?), or a workflow that was launched by a user.  Also, we need to return the schedule in the case where this object is a non-workflow job launched directly from a schedule.

I believe that we probably also want to return information from this method as a dict, so that we can include what type of object we are returning and also the launch type string, but we should probably have a discussion with the UI team to find out what they want exactly.

fosterseth:
here is the logic for the UI to determine what the link points to, `LaunchByDetail.jsx`

```javascript
  if (createdBy) {
    link = `/users/${createdBy.id}`;
    value = createdBy.username;
  } else if (relatedSchedule && jobTemplate) {
    link = `/templates/job_template/${jobTemplate.id}/schedules/${schedule.id}`;
    value = schedule.name;
  } else {
    link = null;
    value = schedule.name;
  }
```

My understanding was that we could add a 3rd `else` clause to look for the `launched_by` field for any job where `created_by` and `related_schedule` isn't defined.

I'm not seeing how this would break other cases, given `launched_by` isn't a field anywhere else

**edit: scratch that, there is a PR that reworks the code I posted above**

jbradberry:
> I'm not seeing how this would break other cases, given `launched_by` isn't a field anywhere else

Well, ok, that's technically true.  However, adding a summary field that only returns info for one specific case doesn't really relieve the underlying problem, which is that the UI resolving code needs to exist and be kept up to date as new cases are added.  My understanding of the original issue is that we want to return a new piece of summary info via the API that removes the need to do this.

awx/main/models/unified_jobs.py
```diff
@@ -1503,3 +1503,18 @@ def log_lifecycle(self, state, blocked_by=None):
...
+        workflow=self.get_workflow_job()
+        if workflow is None:
+            return ""
+
+        #if that workflow was launched by a schedule, get that schedule
+        if(workflow.launch_type=="scheduled"):
+            return workflow.schedule.id
+        else:
+            return ""
+        #here , the return value is the schedule id , but we can also return the workflow id . Need to discuss.
```
ryanpetrello: Given the description at https://github.com/ansible/awx/issues/5057 it seems like what we want here is probably some level of specificity about *what* launched the workflow, not just an integer representing the ID of some object?

Was it a user?  Which user?

* JT launched manually: user who launched job
* JT launched from a workflow manually: user who launched workflow
* Workflow launched manually: user who launched workflow

No?  Was it a schedule?  Which schedule?

* JT launched from schedule: schedule that launched job
* Workflow launched from schedule: schedule that launched workflow

We probably need to think about and coordinate with somebody in the UI (@marshmalien?) about how they want to consume this information.

It looks like there's already some front-end logic being added for this over at https://github.com/ansible/awx/pull/9084.  That might be a good starting point for figuring out how we can move that logic and expose it in the API (so that the UI and other clients don't have to generate it themselves).

Instead of returning an arbitrary integer here, it might make sense to represent this as some type of data structure that specifies the *type* of resource, it's identifier, and maybe even a related link to it so it can be referenced by the UI or other API clients, e.g.,

`{"user": 100, "url": "..."}`
`{"schedule": 200, "url": "..."}`

ryanpetrello: ...and now I realized this is pretty similar to feedback that @jbradberry already left at https://github.com/ansible/awx/pull/9201#pullrequestreview-580726528

Sorry about that, @jbradberry, I should've read more closely 😄

awx/main/models/unified_jobs.py
```diff
@@ -1512,3 +1514,52 @@ def log_lifecycle(self, state, blocked_by=None):
...
         logger_job_lifecycle.debug(msg, extra=extra)
+
+#not doing the recursive logic right now , but just trying to make sure it returns something , the logic is basically what is used in the UI.
+    @property
+    def launched_by(self):
+        
+        # if the job is some maintainance task
+        #sdb.set_trace()
+        if hasattr(self,'job_type'):
+            if self.job_type=="cleanup_activity_stream":
```
jbradberry: What was the issue with this, exactly?

awx/main/models/unified_jobs.py
```diff
@@ -1512,3 +1514,52 @@ def log_lifecycle(self, state, blocked_by=None):
...
+            else:
+                return None
+        else:
+            return None
+
+    @property
+    def ancestor_job(self):
+        
+
+        if hasattr(self,'get_workflow_job()'):
```
jbradberry: Two things,
- `hasattr` is not going to work properly with the parentheses.  `hasattr(self, 'get_workflow_job')`.  Similarly on line 1557.
- guarding against the method not being there isn't necessary.  `get_workflow_job` is defined on `UnifiedJob`, so will be available for all job types.

awx/api/serializers.py
```diff
@@ -140,6 +142,8 @@
     'credential_type': DEFAULT_SUMMARY_FIELDS,
 }
 
+import sdb
```
kdelee: you need to remove this or comment this out in order to install a prod instance because this is only available in the development environment. Just have it enabled when developing.

One thing that can help when you only want to add some changes to a file to git is `git add -i`

https://git-scm.com/book/en/v2/Git-Tools-Interactive-Staging

awx/main/models/unified_jobs.py
```diff
@@ -52,6 +52,8 @@
 from awx.main.consumers import emit_channel_notification
 from awx.main.fields import JSONField, JSONBField, AskForField, OrderedManyToManyField
 
+import sdb
```
kdelee: same thing here

awx/main/models/unified_jobs.py
```diff
@@ -1512,3 +1514,47 @@ def log_lifecycle(self, state, blocked_by=None):
...
+#not doing the recursive logic right now , but just trying to make sure it returns something , the logic is basically what is used in the UI.
+    @property
+    def launched_by(self):
+        
+        #sdb.set_trace()
+        #checking if the current job was started by a workflow job.
+       
+        ancestor_job=self.ancestor_job
+
+        if hasattr(ancestor_job,'launch_type'):
```
jbradberry: All `UnifiedJob`s are guaranteed to have a `.launch_type`.  This condition isn't necessary.

    >>> from collections import Counter
    >>> Counter(hasattr(j, 'launch_type') for j in UnifiedJob.objects.all())
    Counter({True: 7923})

awx/main/models/unified_jobs.py
```diff
@@ -1512,3 +1514,47 @@ def log_lifecycle(self, state, blocked_by=None):
...
+            else:
+                return None
+        else:
+            return None
+
+    @property
+    def ancestor_job(self):
+        
+
+        if hasattr(self,'get_workflow_job'):
```
jbradberry: Similarly for this check:

    >>> Counter(hasattr(j, 'get_workflow_job') for j in UnifiedJob.objects.all())
    Counter({True: 7923})

awx/api/serializers.py
```diff
@@ -817,6 +824,25 @@ def to_representation(self, obj):
...
+        dictionary ={}
+        if obj.launched_by is not None:
+            dictionary = model_to_dict(obj.launched_by)
+            return dictionary
+        else:
+            return {}
+    
+    def get_ancestor_job(self,obj):
+
+        if obj.ancestor_job is not None:
```
jbradberry: Your `ancestor_job` property is written to always return something.  It should be reasonable for this method to just `return model_to_dict(obj.ancestor_job)`

awx/main/models/unified_jobs.py
```diff
@@ -1512,3 +1514,47 @@ def log_lifecycle(self, state, blocked_by=None):
...
+            if ancestor_job.launch_type=="manual":
+                return ancestor_job.created_by
+            elif ancestor_job.launch_type=="scheduled":
+                return ancestor_job.schedule
+            elif ancestor_job.launch_type=="dependency":
+                if ancestor_job.job_type_name=="project_update":
+                    return ancestor_job.project
+                elif ancestor_job.job_type_name=="inventory_update":
+                    return ancestor_job.inventory
+            else:
```
jbradberry: For the webhook `launch_type`, we probably want to return `ancestor_job.unified_job_template`.

jbradberry:
For cleanup purposes, please

- remove all uses and imports of sdb, even if commented out
- put spaces around assignment `=`, operators such as `==`, and spaces after your commas
- reduce the number of blank lines added, such as lines 1544 and 1545.  Some is ok, if it improves readability.

awx/main/models/unified_jobs.py
jbradberry: I suspect,

- `'relaunch'` can only happen manually, so `.created_by`?
- `'sync'` is a project sync, so `.project`?
- `'scm'` is an inventory sync, so `.inventory`?
- `'callback'` says that it is started via a host callback, so `.limit`?  Except that that is a string so we're sad?

awx/api/serializers.py
amolgautam25: The `model_to_dict` does not handle `None` . It throws an error. `NoneType` returns `None` here and it breaks . That is why i did this :
```
if obj.launched_by is not None:
            dictionary = model_to_dict(obj.launched_by)
            return dictionary
        else:
            return {}

```

If `None` goes into model_to_dict , it breaks.

awx/main/models/unified_jobs.py
amolgautam25: Added `relaunch` condition.
Added `webhook` condition. 
Added `sync` condition. ( need to test it )
Added `scm` condition . ( need to test it )

awx/main/models/unified_jobs.py
amolgautam25: Removed the `hasattr()` condition

awx/main/models/unified_jobs.py
amolgautam25: Removed the hasattr() condition

awx/main/models/unified_jobs.py
```diff
@@ -1512,3 +1512,35 @@ def log_lifecycle(self, state, blocked_by=None):
...
+            return ancestor_job.project
+        elif ancestor_job.launch_type == "scm":
+            return ancestor_job.inventory
+
+    @property
+    def ancestor_job(self):
+        workflow_job = self.get_workflow_job()    
+        if workflow_job is not None:
+            ancestor_list=workflow_job.get_ancestor_workflows()
+            if not ancestor_list
```
jbradberry: forgot the `:`

kdelee:
Schema check is failing which is expected because you are changing the schema.

I would like to point out that in the OPTIONS request we are not getting a great description of what this field is

A couple other nearby fields:
```
 `execution_node`: The node the job executed on. (string)\n*
 `controller_node`: The instance that managed the isolated execution environment. (string)\n* `launched_by`:  (field)\n*
```
this one:
```
`ancestor_job`:  (field)\
```
Not sure where we put that in. Maybe it gets pulled from doc strings?

awx/api/serializers.py
```diff
@@ -713,14 +714,17 @@ class UnifiedJobSerializer(BaseSerializer):
         read_only=True
     )
 
+    launched_by=serializers.SerializerMethodField()
+    ancestor_job=serializers.SerializerMethodField()
```
jbradberry: These need to have extra spaces around the `=`.

awx/api/serializers.py
```diff
@@ -713,14 +714,17 @@ class UnifiedJobSerializer(BaseSerializer):
...
+
     class Meta:
         model = UnifiedJob
         fields = ('*', 'unified_job_template', 'launch_type', 'status',
                   'execution_environment',
                   'failed', 'started', 'finished', 'canceled_on', 'elapsed', 'job_args',
                   'job_cwd', 'job_env', 'job_explanation',
                   'execution_node', 'controller_node',
-                  'result_traceback', 'event_processing_finished')
+                  'result_traceback', 'event_processing_finished', 'launched_by','ancestor_job')
```
jbradberry: This needs a space after the `,`

awx/api/serializers.py
```diff
@@ -817,6 +821,18 @@ def to_representation(self, obj):
...
         return ret
 
+    def get_launched_by(self,obj):
+        launched_by = obj.launched_by
+        if launched_by is not None:
+            return model_to_dict(launched_by)
+        return {}
+    
+    def get_ancestor_job(self,obj):
+        ancestor_job=obj.ancestor_job
```
jbradberry: Spaces around the `=`.

awx/main/models/unified_jobs.py
```diff
@@ -1512,3 +1512,35 @@ def log_lifecycle(self, state, blocked_by=None):
         else:
             msg = f"{self._meta.model_name}-{self.id} {state.replace('_', ' ')}"
         logger_job_lifecycle.debug(msg, extra=extra)
+
+    @property
+    def launched_by(self):
+        ancestor_job=self.ancestor_job
```
jbradberry: and here.

awx/main/models/unified_jobs.py
```diff
@@ -1512,3 +1512,35 @@ def log_lifecycle(self, state, blocked_by=None):
...
+        elif ancestor_job.launch_type == "sync":
+            return ancestor_job.project
+        elif ancestor_job.launch_type == "scm":
+            return ancestor_job.inventory
+
+    @property
+    def ancestor_job(self):
+        workflow_job = self.get_workflow_job()    
+        if workflow_job is not None:
+            ancestor_list=workflow_job.get_ancestor_workflows()
```
jbradberry: and here.

ryanpetrello (approved):

awx/main/models/unified_jobs.py
```diff
@@ -1512,3 +1512,35 @@ def log_lifecycle(self, state, blocked_by=None):
...
+
+    @property
+    def launched_by(self):
+        ancestor_job = self.ancestor_job
+        if ancestor_job.launch_type == "manual" or ancestor_job.launch_type == "relaunch":
+            return ancestor_job.created_by
+        elif ancestor_job.launch_type == "scheduled":
+            return ancestor_job.schedule
+        elif ancestor_job.launch_type == "workflow":
+            return ancestor_job.source_workflow_job
```
jbradberry: `source_workflow_job` apparently isn't something provided on the model, it's constructed in the serializer from the `.workflow_job_id`.

jbradberry: I don't think this case will ever actually be hit, since by using `.ancestor_job` we are climbing the ladder of workflows until we hit the one that wasn't launched by yet another workflow.  However, in the interest of making sure something weird hasn't happened, this should probably return `ancestor_job.workflow_job`

jbradberry: ... and it turns out _that_ isn't a field either, it's just calculating from `.get_workflow_job()`.

jbradberry: I guess we should just return `ancestor_job.get_workflow_job()` here.

jbradberry:
> Schema check is failing which is expected because you are changing the schema.
> 
> I would like to point out that in the OPTIONS request we are not getting a great description of what this field is
> 
> A couple other nearby fields:
> 
> ```
>  `execution_node`: The node the job executed on. (string)\n*
>  `controller_node`: The instance that managed the isolated execution environment. (string)\n* `launched_by`:  (field)\n*
> ```
> 
> this one:
> 
> ```
> `ancestor_job`:  (field)\
> ```
> 
> Not sure where we put that in. Maybe it gets pulled from doc strings?

@kdelee Some of our `SerializerMethodField`s have `help_text` defined on them, but I don't see anything more interesting than that going on for existing ones.

@amolgautam25 you should probably come up with some reasonable `help_text` for these serializer fields.

amolgautam25:
regate

ryanpetrello (approved):
This PR looks good to me - @yagomarques or @shebangbash do either of you have some cycles to give this a spin through an integration run?

awx/main/models/unified_jobs.py
```diff
@@ -1512,3 +1512,35 @@ def log_lifecycle(self, state, blocked_by=None):
...
+            return ancestor_job.created_by
+        elif ancestor_job.launch_type == "scheduled":
+            return ancestor_job.schedule
+        elif ancestor_job.launch_type == "workflow":
+            return ancestor_job.get_workflow_job()
+        elif ancestor_job.launch_type == "dependency":
+            if ancestor_job.job_type_name == "project_update":
+                return ancestor_job.project
+            elif ancestor_job.job_type_name == "inventory_update":
+                return ancestor_job.inventory
```
kdelee: @jbradberry not sure I fully understand why its getting set to the inventory, when really it is being launched as a dependency of the job template launch.

ryanpetrello: I guess these objects are the *closest* approximation to an object we have?

But I'll agree, it's not really "the inventory" that launched it, it's Tower's task system.

awx/main/models/unified_jobs.py
```diff
@@ -1512,3 +1512,35 @@ def log_lifecycle(self, state, blocked_by=None):
...
+        ancestor_job = self.ancestor_job
+        if ancestor_job.launch_type == "manual" or ancestor_job.launch_type == "relaunch":
+            return ancestor_job.created_by
+        elif ancestor_job.launch_type == "scheduled":
+            return ancestor_job.schedule
+        elif ancestor_job.launch_type == "workflow":
+            return ancestor_job.get_workflow_job()
+        elif ancestor_job.launch_type == "dependency":
+            if ancestor_job.job_type_name == "project_update":
+                return ancestor_job.project
```
kdelee: @jbradberry not sure I fully understand why its getting set to the project, when really it is being launched as a dependency of the job template launch.

awx/main/models/unified_jobs.py
amolgautam25: "inventory" is not really the object that launched it, but i took a close approximation. From what i understand , the task manager was launching the job. But i was not able to figure out what should be returned in this case. ( I might be wrong here )

awx/main/models/unified_jobs.py
jbradberry: Not sure why I'm getting @ mentioned, this isn't my PR.  I offered some advice, but if other people know of a better model relationship to be used, by all means talk it out.

awx/api/serializers.py
```diff
@@ -817,6 +821,18 @@ def to_representation(self, obj):
         ret['job_explanation'] = _(obj.job_explanation)
         return ret
 
+    def get_launched_by(self,obj):
+        launched_by = obj.launched_by
+        if launched_by is not None:
+            return model_to_dict(launched_by)
```
ryanpetrello: @amolgautam25 this is pretty tall:

```
> model_to_dict(UnifiedJobTemplate.objects.first())
{'created': '2021-03-15 16:35:29.804657+00:00',
 'created_by': None,
 'current_job': None,
 'description': 'Remove activity stream history',
 'execution_environment': None,
 'id': 2,
 'job_type': 'cleanup_activitystream',
 'last_job': None,
 'last_job_failed': False,
 'last_job_run': None,
 'modified': '2021-03-15 16:35:29.804657+00:00',
 'modified_by': None,
 'name': 'Cleanup Activity Stream',
 'next_job_run': '2021-03-16 16:35:29+00:00',
 'next_schedule': 'Cleanup Activity Schedule_t2_2_2021-03-16 16:35:29+00:00',
 'old_pk': None,
 'organization': None,
 'polymorphic_ctype': 'system job template',
 'status': 'ok',
 'unifiedjobtemplate_ptr': 'Cleanup Activity Stream-2'}
```

Is this the data we want to surface here?

cc @jbradberry @kdelee

This is a bunch of metadata *about* the job, but it doesn't seem like it tells you what it *is* in a deterministic way (which is what I'd expect an API consumer like the UI to _want_: https://github.com/ansible/awx/pull/9201#discussion_r573000566)

In line w/ my comment above, things I'd expect some consumer of this API to want to know about an arbitrary object:

1.  What "type" of object it is.
2. What it's unique identifier (primary key?) is, and probably its "name"
3. A URL reference to the object

e.g.,

```json
{
    "type": "user",
    "id": 123,
    "name": "Ryan",
    "url": "/api/v2/users/123/"
}
```

https://github.com/ansible/awx/pull/9084/files#diff-89bcbc225ce4bf098ef34185ff1312dcb1d15aef32de9b1e40949014e143bd8eR9-R28

jbradberry: Probably not.  We need the type, the name, and the id, IIRC.

jbradberry: @ryanpetrello yeah, the api url sounds like a reasonable thing to want, for consistency.  The immediate use of these additions, though, is to construct a UI route.

ryanpetrello (changes requested):
Left some comments requesting some changes.

awx/api/serializers.py
ryanpetrello: If we did something _like this_, though, we could represent things like system dependencies like:

```
{
    "type": "dependency",
    "id": null,
    "name": "System Dependency",
    "url": null
}
```

awx/main/models/unified_jobs.py
jbradberry: https://github.com/ansible/awx/blob/f850f8d/awx/main/scheduler/task_manager.py#L352-L381

By my reading of this code, we don't store any information about what job triggered the update for either case.

kdelee: k, we will stick with this then

awx/api/serializers.py
kdelee: :eyes: Yeah if we could trim this down and have consistent contents that would be ideal

so instead of `model_to_dict` we would have

dependency
```
    "type": "dependency",
    "id": null,
    "name": "System Dependency",
    "url": null
```
workflow triggered job
```
    "type": "workflow",
    "id": 123,
    "name": "Workflow that triggered this job Name",
    "url": /api/v2/workflows/123
```

webhook
```
    "type": "job_template",
    "id": 345,
    "name": "Job Template That Has the Webook Name",
    "url": /api/v2/job_template/345
```

... etc

amolgautam25: I have made these changes. 
`launched_by` returns 4 attributes : 
`id , type , name , url`

Right now the url point towards the API endpoints as mentioned above.

amolgautam25: cc @ryanpetrello @kdelee @jbradberry

awx/main/models/unified_jobs.py
```diff
@@ -1512,3 +1512,77 @@ def log_lifecycle(self, state, blocked_by=None):
...
+    @property
+    def launched_by(self):
+        ancestor_job = self.ancestor_job
+        launched_by = {}
+        if ancestor_job.launch_type == "manual" or ancestor_job.launch_type == "relaunch":
+            user_obj=ancestor_job.created_by
+            launched_by['id'] = user_obj.id
+            launched_by['name'] = user_obj.username
+            launched_by['type'] = "User"
+            launched_by['url'] = "/api/v2/users/" + str(user_obj.id)
```
ryanpetrello: Most objects in the system have a method you can call to generate this: `get_absolute_url`

```python
In [1]: User.objects.first().get_absolute_url()
Out[1]: '/api/v2/users/1/'
```

awx/main/models/unified_jobs.py
```diff
@@ -1512,3 +1512,77 @@ def log_lifecycle(self, state, blocked_by=None):
         else:
             msg = f"{self._meta.model_name}-{self.id} {state.replace('_', ' ')}"
         logger_job_lifecycle.debug(msg, extra=extra)
+
+    @property
+    def launched_by(self):
+        ancestor_job = self.ancestor_job
+        launched_by = {}
+        if ancestor_job.launch_type == "manual" or ancestor_job.launch_type == "relaunch":
+            user_obj=ancestor_job.created_by
```
ryanpetrello: spaces between the `=`

awx/api/serializers.py
```diff
@@ -713,14 +714,17 @@ class UnifiedJobSerializer(BaseSerializer):
         read_only=True
     )
 
+    launched_by = serializers.SerializerMethodField(help_text=_('Object that launched the current job'))
+    ancestor_job = serializers.SerializerMethodField(help_text=_('The parent workflow that started the job, if not then ancestor_job returns self '))
```
ryanpetrello: Does this field _need_ to be exposed to the API, or is it just a function that's used internally by `launched_by`?

cc @marshmalien @wenottingham

awx/main/models/unified_jobs.py
```diff
@@ -1512,3 +1512,77 @@ def log_lifecycle(self, state, blocked_by=None):
         else:
             msg = f"{self._meta.model_name}-{self.id} {state.replace('_', ' ')}"
         logger_job_lifecycle.debug(msg, extra=extra)
+
+    @property
+    def launched_by(self):
+        ancestor_job = self.ancestor_job
+        launched_by = {}
+        if ancestor_job.launch_type == "manual" or ancestor_job.launch_type == "relaunch":
```
ryanpetrello: There's a lot of duplication in this `if` statement.  You could probably refactor most of this to do e.g.,

```python
if ancestor_job.launch_type in ('manual', 'relaunch'):
    obj = ancestor_job.created_by
elif ancestor_job.launch_type == "scheduled":
    obj = ancestor_job.schedule
elif:
    ...
elif:
    ...
return {
    'id': obj.id,
    'name': getattr(obj, 'name', None) or obj.username,
    'type': obj.__class__.__name__,
    'url': obj.get_absolute_url()
}

Comment (ryanpetrello) [reply awx/main/models/unified_jobs.py]: This could probably be refined even further by removing the big `if/elif/elif/elif` statement and using a dict to map launch types to related objects e.g.,

```python
attr = {
    'manual': 'created_by',
    'relaunch': 'created_by',
    'scheduled': 'schedule',
    ...
}
obj = getattr(ancestor_job, attr, None)
if obj is not None:
    return {
        ...
    }
return {}
```

awx/api/serializers.py
jbradberry: @ryanpetrello it does.  Per our conversation with @marshmalien a while back, the UI route built up to the `launched_by` object is relative to the job that it is referenced by, the `ancestor_job`.

awx/main/models/unified_jobs.py
jbradberry: 100% this.

awx/main/models/unified_jobs.py
jbradberry: I concur.

awx/api/serializers.py
ryanpetrello: Gotcha 👍

awx/main/models/unified_jobs.py
jbradberry: More like

```python
attr = { ... }
obj = getattr(ancestor_job, attr.get(ancestor_job.launch_type, ''), None)
if obj is not None:
    return { ... }
return {}
```

awx/main/models/unified_jobs.py
```diff
@@ -1512,3 +1512,36 @@ def log_lifecycle(self, state, blocked_by=None):
...
+        "workflow" : "workflow", 
+        "webhook" : "job_template" , 
+        "sync" : "project" ,
+        "scm" : "inventory",
+        }
+
+        obj = getattr(ancestor_job, attr.get(ancestor_job.launch_type, ''), None)
+        if obj is not None:
+            return {
+            'id': obj.id,
```
ryanpetrello: This likely needs indentation.

awx/main/models/unified_jobs.py
```diff
@@ -1512,3 +1512,36 @@ def log_lifecycle(self, state, blocked_by=None):
         else:
             msg = f"{self._meta.model_name}-{self.id} {state.replace('_', ' ')}"
         logger_job_lifecycle.debug(msg, extra=extra)
+
+    @property
+    def launched_by(self):
+        ancestor_job = self.ancestor_job
+        attr = {"manual" : "created_by",
+        "relaunch" : "created_by",
```
ryanpetrello: This likely needs indentation.

awx/api/serializers.py
```diff
@@ -64,6 +64,7 @@
 from awx.main.models.rbac import (
     get_roles_on_resource, role_summary_fields_generator
 )
+from awx.main.utils.common import model_to_dict
```
jbradberry: I think this is now no longer used.

awx/main/models/unified_jobs.py
```diff
@@ -1512,3 +1512,37 @@ def log_lifecycle(self, state, blocked_by=None):
...
+            "sync" : "project" ,
+            "scm" : "inventory",
+            }
+
+        obj = getattr(ancestor_job, attr.get(ancestor_job.launch_type, ''), None)
+        if obj is not None:
+            return {
+                'id': obj.id,
+                'name': getattr(obj, 'name', None) or obj.username,
+                'type': obj.__class__.__name__,
```
jbradberry: I believe that there is some standardized way we expose the type of objects in the API.  Let me see if I can't find it...

awx/api/serializers.py
amolgautam25: It is being used in line 831. 
```
def get_ancestor_job(self,obj):
        ancestor_job = obj.ancestor_job
        if ancestor_job is not None:
            return model_to_dict(ancestor_job)
        return {}```

Comment (jbradberry) [reply awx/api/serializers.py]: That is likely going to have the same objection then as the other usage of it -- it is returning too much info.

File: awx/main/models/unified_jobs.py (hunk 27 above)
Comment (jbradberry) [reply awx/main/models/unified_jobs.py]: The function `get_type_for_model` from `awx.main.utils`.  This is what we use for the `BaseSerializer.type` serializer method field.

File: awx/main/models/unified_jobs.py [hunk 28]
```diff
@@ -1512,3 +1512,37 @@ def log_lifecycle(self, state, blocked_by=None):
...
+        ancestor_job = self.ancestor_job
+        attr = {
+            "manual" : "created_by",
+            "relaunch" : "created_by",
+            "scheduled" : "schedule" , 
+            "workflow" : "workflow", 
+            "webhook" : "job_template" , 
+            "sync" : "project" ,
+            "scm" : "inventory",
+            }
```
amolgautam25: There is one special case here which confuses me. 
```launch_type == dependency```

I used a nested loop logic for this ```launch_type```

```
elif ancestor_job.launch_type == "dependency":
      if ancestor_job.job_type_name == "project_update":
      ......
      elif ancestor_job.job_type_name == "inventory_update":
      .......
```
Right now , i am not sure how to achieve the same result in this line : 
 
```
obj = getattr(ancestor_job, attr.get(ancestor_job.launch_type, ''), None)

awx/api/serializers.py
ryanpetrello: I wonder if this would make more sense to expose as a related URL (or maybe a summary field) instead of as an actual field on the job record.  @jbradberry what do you think?

jbradberry: The UI needs to construct a route around the info from it.

awx/main/models/unified_jobs.py
ryanpetrello: cc @kdelee given that the link to inventory/project is _a little weird/confusing_ (it's not *that* object that "launched it"), should we change this special case up as previously discussed to be something like e.g.,

```python
if ancestor_job.launch_type == "dependency":
    return {
        'id': None,
        'name': 'Generated by AWX',
        'type': 'Dependency',
        'url': None
    }
attr = {
    ...

File: awx/main/models/unified_jobs.py (hunk 27 above)
Comment (ryanpetrello) [reply awx/main/models/unified_jobs.py]: Agreed, this is what we should use ☝️

Comment (amolgautam25) [reply awx/main/models/unified_jobs.py]: Updated the PR with ```get_type_for_model```

File: awx/main/models/unified_jobs.py (hunk 28 above)
Comment (ryanpetrello) [reply awx/main/models/unified_jobs.py]: cc @mabashian

**Comment by jbradberry** [comment]
The linter is complaining.

```
2021-03-16 16:19:08.326276 | static | mkdir -p reports
2021-03-16 16:19:26.052975 | static | ./awx/main/models/unified_jobs.py:1527:13: E123 closing bracket does not match indentation of opening bracket's line
2021-03-16 16:19:26.053243 | static | ./awx/main/models/unified_jobs.py:1536:17: E123 closing bracket does not match indentation of opening bracket's line
2021-03-16 16:19:26.082734 | static | make: *** [Makefile:278: flake8] Error 1
2021-03-16 16:19:26.083211 | static | ERROR: InvocationError for command /usr/bin/make flake8 (exited with code 2)
```

File: awx/api/serializers.py (hunk 26 above)
Comment (jbradberry) [reply awx/api/serializers.py]: Well, ok, maybe a summary field _would_ be the thing to do.

File: awx/main/models/unified_jobs.py (hunk 28 above)
Comment (amolgautam25) [reply awx/main/models/unified_jobs.py]: I have added a seperate check for ```launch_type == "dependency":```

File: awx/api/serializers.py [hunk 29]
```diff
@@ -817,6 +821,16 @@ def to_representation(self, obj):
...
         return ret
 
+    def get_launched_by(self,obj):
+        if obj is not None:
+            return obj.launched_by
+    
+    def get_ancestor_job(self,obj):
+        ancestor_job = obj.ancestor_job
+        if ancestor_job is not None:
+            return model_to_dict(ancestor_job)
```
ryanpetrello: @marshmalien can you provide some input on the format of data you're looking for here?

Do you *just* need a URL?

awx/api/serializers.py
amolgautam25: Working on it.

awx/api/serializers.py
ryanpetrello: @amolgautam25 @jbradberry we discussed this privately, but I'll repeat here

I think this makes _more_ sense as a summary field, **not** an object-level field.  We should edit `UnifiedJobSerializer.get_summary_fields()` to do something like this, I think:

```python
if self.is_detail_view:
    ancestor = obj.ancestor_job
    if ancestor != obj:
        summary_fields['ancestor_job'] = {
            'id' : ancestor.id,
            'name': ...
            'type': ...
            'url': ...
        }
```

awx/api/serializers.py
ryanpetrello: relevant comment: https://github.com/ansible/awx/pull/9201/files#r595450489

ryanpetrello: With the summary field added, we shouldn't need this anymore.

awx/api/serializers.py
ryanpetrello: With the summary field added, we shouldn't need this `SerializerMethodField` anymore.

awx/api/serializers.py
```diff
@@ -713,14 +714,17 @@ class UnifiedJobSerializer(BaseSerializer):
...
+
     class Meta:
         model = UnifiedJob
         fields = ('*', 'unified_job_template', 'launch_type', 'status',
                   'execution_environment',
                   'failed', 'started', 'finished', 'canceled_on', 'elapsed', 'job_args',
                   'job_cwd', 'job_env', 'job_explanation',
                   'execution_node', 'controller_node',
-                  'result_traceback', 'event_processing_finished')
+                  'result_traceback', 'event_processing_finished', 'launched_by', 'ancestor_job')
```
ryanpetrello: With the summary field added, we shouldn't need this anymore.

awx/api/serializers.py
ryanpetrello: With the summary field added, we shouldn't need this method anymore.

awx/api/serializers.py
amolgautam25: I have removed it.

awx/api/serializers.py
amolgautam25: I have removed it.

awx/api/serializers.py
amolgautam25: I have removed it.

awx/main/models/unified_jobs.py
amolgautam25: Indented.

awx/main/models/unified_jobs.py
amolgautam25: Indented.

awx/api/serializers.py
amolgautam25: I have removed it.

rebeccahhh:
someone can feel free to disagree with me, and I don't advocate for this _too_ often, but you have 17 commits in here. I would recommend an interactive rebase so you can clean them up into more organized logical blocks and the ones that are small fixes like indentation and such can just be squashed into those larger logical blocks for clarity.

ryanpetrello (approved):
This is looking good to me, and is probably ready for @kdelee to take a look.

ryanpetrello:
@amolgautam25 I agree with @rebeccahhh, I think it's worth rebasing this PR down to one descriptive commit prior to merging.

awx/api/serializers.py
```diff
@@ -773,6 +774,16 @@ def get_summary_fields(self, obj):
                 if val is not None:
                     summary_fields['source_workflow_job'][field] = val
 
+        if self.is_detail_view:
+            ancestor = obj.ancestor_job
+            if ancestor != obj:
+                summary_fields['ancestor_job']= {
```
jbradberry: spaces around the `=`, please.

awx/api/serializers.py
```diff
@@ -817,6 +828,10 @@ def to_representation(self, obj):
         ret['job_explanation'] = _(obj.job_explanation)
         return ret
 
+    def get_launched_by(self,obj):
```
jbradberry: space after the `,`, please

awx/main/models/unified_jobs.py
```diff
@@ -1512,3 +1512,45 @@ def log_lifecycle(self, state, blocked_by=None):
...
+        if ancestor_job.launch_type == "dependency":
+            return {
+                'id': None,
+                'name': 'Generated by AWX',
+                'type': 'Dependency',
+                'url': None
+            }
+
+        attr = {
+            "manual" : "created_by",
```
jbradberry: These `:` characters here and below shouldn't get a space in front of them.

awx/api/serializers.py
```diff
@@ -773,6 +774,16 @@ def get_summary_fields(self, obj):
                 if val is not None:
                     summary_fields['source_workflow_job'][field] = val
 
+        if self.is_detail_view:
+            ancestor = obj.ancestor_job
+            if ancestor != obj:
```
jbradberry, line 807: Do we actually want to make this conditional?  I would think it would be nice for the UI code if they didn't have to defensively check if this was here and fall back to the full data of the current job if not.

awx/api/serializers.py
```diff
@@ -773,6 +774,16 @@ def get_summary_fields(self, obj):
...
+                summary_fields['ancestor_job']= {
+                    'id' : ancestor.id,
+                    'name': getattr(ancestor, 'name', None),
```
jbradberry: I think `.name` is guaranteed to be on a `UnifiedJob`.

ryanpetrello: Yep.

awx/api/serializers.py
ryanpetrello: I think so (reasons privately discussed).

yagomarques:
Hey @amolgautam25, on this scenario - `normal job template launched by a workflow` 

Why we had on the `job.ancestor_job.unified_job_template` the `workflow.name` concatenated with the `workflow.id`? Would it be better if has the name or the id only? 

example:
- `wjft` is a workflow representation that i will use to run a job. 
```
(Pdb) wfjt.name
'WorkflowJobTemplate - TunePermission♪'
(Pdb) wfjt.id
23
```
- the `job` is a variable that contains the result of a job execution.
 ```
(Pdb) job.ancestor_job
{'id': 22, 'polymorphic_ctype': 'workflow job', 'modified': '2021-03-17 17:47:11.654874+00:00', 'description': 'TermFreedomSentenceStoryBedBeingPublicPhotoFoldDefinition㬋', 'created_by': 'User-SwimTheory144707247105912480326120585437031209205148812391537797713', 'modified_by': 'User-SwimTheory144707247105912480326120585437031209205148812391537797713', 'name': 'WorkflowJobTemplate - TunePermission♪', 'execution_environment': 'Default EE-1', 'old_pk': None, 'emitted_events': 0, 'unified_job_template': 'WorkflowJobTemplate - TunePermission♪-23', 'created': '2021-03-17 17:47:11.424617+00:00', 'launch_type': 'manual', 'schedule': None, 'execution_node': '', 'controller_node': '', 'cancel_flag': False, 'status': 'successful', 'failed': False, 'started': '2021-03-17 17:47:11.654733+00:00', 'dependencies_processed': False, 'finished': '2021-03-17 17:47:17.018853+00:00', 'canceled_on': None, 'elapsed': '5.364', 'job_args': '', 'job_cwd': '', 'job_env': '{}', 'job_explanation': '', 'start_args': 'hidden', 'result_traceback': '', 'celery_task_id': '14ba3022-7faf-4be2-af80-aa09e603c07a', 'instance_group': None, 'organization': None, 'installed_collections': '{}', 'unifiedjob_ptr': '2021-03-17 17:47:11.424617+00:00-22-successful', 'survey_passwords': '{}', 'webhook_service': '', 'webhook_credential': None, 'webhook_guid': '', 'inventory': None, 'char_prompts': '{}', 'extra_vars': '{}', 'allow_simultaneous': False, 'workflow_job_template': 'WorkflowJobTemplate - TunePermission♪-23', 'job_template': None, 'is_sliced_job': False}
```
if you check the `workflow_job_template` the value is `WorkflowJobTemplate - TunePermission♪-23`
the workflow_job_template is a reference to `wfjt` with the name and the id concatenated.

**Comment by jbradberry** [comment]
> Hey @amolgautam25, on this scenario - `normal job template launched by a workflow`
> 
> Why we had on the `job.ancestor_job.unified_job_template` the `workflow.name` concatenated with the `workflow.id`? Would it be better if has the name or the id only?
> 
> example:
> 
>     * `wjft` is a workflow representation that i will use to run a job.
> 
> 
> ```
> (Pdb) wfjt.name
> 'WorkflowJobTemplate - TunePermission♪'
> (Pdb) wfjt.id
> 23
> ```
> 
>     * the `job` is a variable that contains the result of a job execution.
> 
> 
> ```
> (Pdb) job.ancestor_job
> {'id': 22, 'polymorphic_ctype': 'workflow job', 'modified': '2021-03-17 17:47:11.654874+00:00', 'description': 'TermFreedomSentenceStoryBedBeingPublicPhotoFoldDefinition㬋', 'created_by': 'User-SwimTheory144707247105912480326120585437031209205148812391537797713', 'modified_by': 'User-SwimTheory144707247105912480326120585437031209205148812391537797713', 'name': 'WorkflowJobTemplate - TunePermission♪', 'execution_environment': 'Default EE-1', 'old_pk': None, 'emitted_events': 0, 'unified_job_template': 'WorkflowJobTemplate - TunePermission♪-23', 'created': '2021-03-17 17:47:11.424617+00:00', 'launch_type': 'manual', 'schedule': None, 'execution_node': '', 'controller_node': '', 'cancel_flag': False, 'status': 'successful', 'failed': False, 'started': '2021-03-17 17:47:11.654733+00:00', 'dependencies_processed': False, 'finished': '2021-03-17 17:47:17.018853+00:00', 'canceled_on': None, 'elapsed': '5.364', 'job_args': '', 'job_cwd': '', 'job_env': '{}', 'job_explanation': '', 'start_args': 'hidden', 'result_traceback': '', 'celery_task_id': '14ba3022-7faf-4be2-af80-aa09e603c07a', 'instance_group': None, 'organization': None, 'installed_collections': '{}', 'unifiedjob_ptr': '2021-03-17 17:47:11.424617+00:00-22-successful', 'survey_passwords': '{}', 'webhook_service': '', 'webhook_credential': None, 'webhook_guid': '', 'inventory': None, 'char_prompts': '{}', 'extra_vars': '{}', 'allow_simultaneous': False, 'workflow_job_template': 'WorkflowJobTemplate - TunePermission♪-23', 'job_template': None, 'is_sliced_job': False}
> ```
> 
> if you check the `workflow_job_template` the value is `WorkflowJobTemplate - TunePermission♪-23`
> the workflow_job_template is a reference to `wfjt` with the name and the id concatenated.

@yagomarques The `name` attribute is an actual field in the database.  What you are seeing is just a convention for how names for test objects are being constructed in the test suite, I think.

File: awx/main/models/unified_jobs.py [hunk 36]
```diff
@@ -1512,3 +1512,43 @@ def log_lifecycle(self, state, blocked_by=None):
...
+            return {
+                'id':obj.id,
+                'name':getattr(obj, 'name', None) or obj.username,
+                'type':get_type_for_model(obj),
+                'url':obj.get_absolute_url()    
+            }
+        return {}
+
+    @property
+    def ancestor_job(self):
```
amolgautam25, line 1510: In this function i have used ```get_workflow_job()``` function a lot. 
Let me know if i can reduce the number of function calls. 
From what i understand this a very resource heavy call for database.

amolgautam25: cc @jbradberry

ryanpetrello: This *could* get expensive given that we're nesting this call, and there's probably contexts where this is called in a list view (the `launched_by` property relies on it, and that's displayed in the list view).

That said, if we want to provide this data reliably, we probably don't have any other great options here.  Also, this should only be expensive for views that end up being populated primarily by jobs that spawned from workflows.

ryanpetrello: It we really wanted to know for sure, it might be useful to spawn ~50 workflow jobs, and then compare the time it takes to render `/api/v2/jobs/` *before* and *after* this change to make sure it's not _notably_ different.

ryanpetrello: We could probably also optimize this a bit without having to do so many joins every time by testing with `spawned_by_workflow`.  This is also a place where recursion might be a bit clearer than a loop.  I haven't tested this code, but maybe something like this is close?

```python
@property
def ancestor_job(self):
    if self.spawned_by_workflow:
        return self.get_workflow_job().ancestor_job
    return self
```

@jbradberry I'm curious on your thoughts on ☝️ too, as you've been paying closer attention to this PR than I have.

ryanpetrello: I _think_ this works; I tested this by making a Workflow called "Grandpa", which has one node: another workflow called "Dad".  The Dad workflow has one node, which is a Job Template.

Grandpa --> Dad --> Demo Job Template

```python
In [5]: j.name
Out[5]: 'Demo Job Template'

In [6]: j.ancestor_job.name
Out[6]: 'Grandpa'

In [7]: j.get_workflow_job().name
Out[7]: 'Dad'

In [8]: j.get_workflow_job().get_workflow_job().name
Out[8]: 'Grandpa'
```

amolgautam25: Changed it to a one line recursive function.

awx/main/models/unified_jobs.py
```diff
@@ -1519,3 +1519,38 @@ def log_lifecycle(self, state, blocked_by=None):
...
+        if ancestor_job.launch_type == "dependency":
+            return {
+                'id':None,
+                'name':'Generated by AWX',
+                'type':'Dependency',
+                'url':None
+            }
+
+        attr = {
+            "manual":"created_by",
```
jbradberry: There should be a space after the `:`.

-> `"manual": "created_by",`

awx/main/models/unified_jobs.py
```diff
@@ -1519,3 +1519,38 @@ def log_lifecycle(self, state, blocked_by=None):
...
+            "relaunch":"created_by",
+            "scheduled":"schedule" ,
```
jbradberry: There's an extraneous space before the comma.  Please remove it.

unlikelyzero:
https://github.com/ansible/awx/pull/9084 Is blocked by this PR

masbahnana (approved):

ryanpetrello:
@amolgautam25 would you mind rebasing this PR?  It looks like it needs a rebase and some conflict resolution now that our new linting tool (`black`) has been merged in?

PR #9205, opened 2021-02-02
Fix Pylint Error

beeankha:
##### SUMMARY
`pylint` doesn't like throwaway variables; this PR replaces the `_` in `test_build_notification_message_undefined` with a tuple index so that the string `'running'` can still be accessed without throwing an error.

##### AWX VERSION

```
17.0.0
```


Spredzy (approved):
LGTM

PR #9210, opened 2021-02-02
Prompts on schedules

AlexSCorey:
##### SUMMARY
This Addresses #7692.  Couple of unique items for this workflow because this is a form that uses both standard on screen form, and a wizard combo.
1)The save button should be disabled when the user must do some work inside the wizard because it is missing some required information like an inventory, or some survey values.
2) If the user cancels the modal in the middle of it, and then saves, it should not submit new values from the modal
3)Prompt button should not render for schedules that do not need it (ie. Inventory Source, Projects, and templates that do not have any prompt on launch values).

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
schedulePrompting

### DIFF
--- a/awx/ui_next/src/api/models/InventorySources.js
+++ b/awx/ui_next/src/api/models/InventorySources.js
@@ -10,6 +10,7 @@ class InventorySources extends LaunchUpdateMixin(
     super(http);
     this.baseUrl = '/api/v2/inventory_sources/';
 
+    this.createSchedule = this.createSchedule.bind(this);
     this.createSyncStart = this.createSyncStart.bind(this);
     this.destroyGroups = this.destroyGroups.bind(this);
     this.destroyHosts = this.destroyHosts.bind(this);
--- a/awx/ui_next/src/api/models/JobTemplates.js
+++ b/awx/ui_next/src/api/models/JobTemplates.js
@@ -10,6 +10,7 @@ class JobTemplates extends SchedulesMixin(
     super(http);
     this.baseUrl = '/api/v2/job_templates/';
 
+    this.createSchedule = this.createSchedule.bind(this);
     this.launch = this.launch.bind(this);
     this.readLaunch = this.readLaunch.bind(this);
     this.associateLabel = this.associateLabel.bind(this);
--- a/awx/ui_next/src/api/models/Projects.js
+++ b/awx/ui_next/src/api/models/Projects.js
@@ -16,6 +16,7 @@ class Projects extends SchedulesMixin(
     this.readPlaybooks = this.readPlaybooks.bind(this);
     this.readSync = this.readSync.bind(this);
     this.sync = this.sync.bind(this);
+    this.createSchedule = this.createSchedule.bind(this);
   }
 
   readAccessList(id, params) {
--- a/awx/ui_next/src/api/models/Schedules.js
+++ b/awx/ui_next/src/api/models/Schedules.js
@@ -14,6 +14,19 @@ class Schedules extends Base {
     return this.http.get(`${this.baseUrl}${resourceId}/credentials/`, params);
   }
 
+  associateCredential(resourceId, credentialId) {
+    return this.http.post(`${this.baseUrl}${resourceId}/credentials/`, {
+      id: credentialId,
+    });
+  }
+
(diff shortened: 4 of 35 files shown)


jakemcdermott (approved):

tiagodread:
Hey @AlexSCorey 
- I was not able to visit the schedule form for a JT with prompts and survey:

schedules

After disabling the survey the schedule add form was properly rendered 

- Repeat Frequency field isn't being properly persisted:

schedules1

tiagodread:
- If the JT contains survey (without default answer) enabled and the user attempts to enable/disable the schedule an error message is displayed:

schedules3

- If a JT contains a scheduler defined and then the user update the JT enabling any field as prompt (i.e inventory) the next run for this JT scheduled will fail and there's nothing to inform the user about it (that the JT has a scheduler enabled and it needs to be updated) The old UI has the same problem, but differently from the new UI the user can see why this job failed using `Explanation` not present in the new UI

schedules4

Old UI
image

In addition to this problem, if the user try to edit the schedule the message `TypeError: Cannot read property 'id' of undefined ` is displayed:

schedules5


- The same problem occurs for surveys, if the JT with scheduler was modified adding a survey without default answer, the job should fail right? because the scheduler is outdated (without the survey answer) but the job is succeeding and isn't respecting the survey answer and using an empty value by default:

image

- Create a JT without prompt, survey and schedule, create a workflow using this JT with schedule, modify the JT enabling prompt (i.e inventory) then this workflow starts failing and nothing is displayed to the user saying that the JT being edited is used by a workflow with schedule enabled and the schedule needs to be updated with the missing JT fields marked as prompt. OK, then, if the user try to edit the workflow schedule to add these missing jt prompt fields the edit page doesn’t show the prompt button with the wizard for JT prompts fields and the page crashes with the message `Unhandled Rejection (TypeError): (intermediate value)(intermediate value)(intermediate value) is not iterable if the user try to save.
`

image

tiagodread (changes requested):
awesome! only a couple of things to be addressed before moving forward

AlexSCorey:
@tiagodread I have resolved the issues with the page crashing, frequency persistence, the toggling, and the job detail error explanation.  For the issues involving changes to a job template that cause a schedule to fail I've filed an enhancement issue https://github.com/ansible/awx/issues/9294.

tiagodread:
@AlexSCorey about the page crashing: I was not able to save a schedule on workflow the Save button is always disabled.


- If any prompt field is required the `Save` button should be disabled: 
1. Create a JT
2. Create a schedule
3. Add a survey without default answer
4. Edit the schedule

image

- There’s two scenarios where the user can not toggle the schedule on/off:
1. Create a JT
2. Create a schedule
3. Add a survey without default answer
4. toggle the schedule on/off

image

1. Create a JT
2. Create a schedule
3. Update the JT enabling prompt for inventory (required field)
4. toggle the schedule on/off

image

Both cases, when the user visit the schedule edit page: `TypeError: Cannot read property 'id' of undefined at https://localhost:3001/static/js/main.chunk.js:29799:261 at async https://localhost:3001/static/js/main.chunk.js:124255:26`

Ok, i know that these failures are related to the necessity to update the schedule with prompt, imo this is more than an [enhancement](https://github.com/ansible/awx/issues/9294) because to be able to enable/disable the schedule, now the user have to guess that he needs to update the schedule in order to disable it. Thoughts on it @mabashian @AlexSCorey ?

- The credential password prompt is not being required on the schedule modal wizard:
1. Create a machine credential OR a vault credential with prompt enabled for password
2. Create a JT and mark Credentials as prompt
2. Create a schedule for this JT
3. On the Credentials step choose the credential created before (with prompt)

The step `Credential passwords` step to fill the credential password isn't displayed: 

schedules7

mdicerbo:
Are there any workaround for this problem?

AlexSCorey:
@mdicerbo What problem are you referring to?

tiagodread:
Filed a separate issue to keep track on: https://github.com/ansible/awx/issues/9366, https://github.com/ansible/awx/issues/9367

tiagodread (approved):
Awesome!

tiagodread:
regate

Issue #9229, opened 2021-02-04
[ui_next][security] Plain text passwords appearing in URL address bar for SSH credentials with user input


jitteryjuice:
##### ISSUE TYPE
 - Bug Report
 - Security Bug Report

##### SUMMARY
Launching a template that requires the user entering a password and the user hitting the enter key after typing in their password (without hitting next) will close the modal popup, and provide the credential name and user password in the URL bar in plain text.

##### ENVIRONMENT
* AWX version: 17.0.1
* AWX install method: Kubernetes
* Ansible version:  2.9.17
* Operating System: Ubuntu OS Base, AWX Container
* Web Browser: Chrome/Edge/Firefox

##### STEPS TO REPRODUCE

1. Launch a template that requires user credentials
2. Select Inventory
3. Select Credential that requires user input
4. Type in credential password
5. Hit enter

##### EXPECTED RESULTS

The users entered password is saved in the form, and the modal continues to "Other Prompts"

##### ACTUAL RESULTS

The modal popup closes, the entered password, and its associated username are shown in plain-text in the URL address bar

##### ADDITIONAL INFORMATION

![ssh_credential_name_address_url](https://user-images.githubusercontent.com/8851024/106836383-0650f180-66d4-11eb-89b0-a00a2dbbd468.png)
![ssh_credential_password_address_url](https://user-images.githubusercontent.com/8851024/106836389-081ab500-66d4-11eb-873c-20d68d1458f0.png)

jakemcdermott:
Thank you, @j-shade.

jakemcdermott:
@j-shade 

I've opened a pull request to fix this. https://github.com/ansible/awx/pull/9231

It will be fixed in the next community release of awx.

jitteryjuice:
> @j-shade
> 
> I've opened a pull request to fix this. #9231
> 
> It will be fixed in the next community release of awx.

Perfect! thanks for the prompt reply @jakemcdermott :)

unlikelyzero:
No unit tests were added as a part of the PR.

I believe that this https://github.com/ansible/awx/issues/9310 issue is a result of the current implementation. Although I can't be sure. I'm putting this back into devel

jakemcdermott:
They're not related. Credential form code wasn't modified in that pr.

PR #9231, opened 2021-02-04
Prevent url smashing from prompted inputs

jakemcdermott:
##### SUMMARY
See: https://github.com/ansible/awx/issues/9229

The launch prompts use forms on several steps. By default, the forms attempt to handle all Enter key-press events as a form submission. The default method for a raw html form is a GET request, so user input to the prompt fields will be injected into the url as query params when handling the key-press.

User input is collected by the launch prompts on multiple forms across several steps, so we should disable the default event handlers on the intermediate steps.

### DIFF
--- a/awx/ui_next/src/components/LaunchPrompt/steps/CredentialPasswordsStep.jsx
+++ b/awx/ui_next/src/components/LaunchPrompt/steps/CredentialPasswordsStep.jsx
@@ -84,7 +84,11 @@ function CredentialPasswordsStep({ launchConfig, i18n }) {
   }
 
   return (
-    <Form>
+    <Form
+      onSubmit={e => {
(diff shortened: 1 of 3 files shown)


awx/ui_next/src/components/LaunchPrompt/steps/CredentialPasswordsStep.jsx
```diff
@@ -84,7 +84,11 @@ function CredentialPasswordsStep({ launchConfig, i18n }) {
   }
 
   return (
-    <Form>
+    <Form
+      onSubmit={e => {
+        e.preventDefault();
```
marshmalien, line 89: 👍

marshmalien (approved):

PR #9239, opened 2021-02-04
Use Postgresql secret in openshift/k8s

jaymiracola:
##### SUMMARY

Removed plain text password from environment variables and added them to the secret that is created for postgresql. 

##### ISSUE TYPE

 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME

 - Installer

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/installer/roles/kubernetes/templates/postgresql-persistent.yml.j2
+++ b/installer/roles/kubernetes/templates/postgresql-persistent.yml.j2
@@ -159,13 +159,13 @@ parameters:
 - description: Username for PostgreSQL user that will be used for accessing the
     database.
   displayName: PostgreSQL Connection Username
-  from: user[A-Z0-9]{3}
+  from: {{ pg_username }}
   generate: expression
   name: POSTGRESQL_USER
   required: true
 - description: Password for the PostgreSQL connection user.
   displayName: PostgreSQL Connection Password
-  from: '[a-zA-Z0-9]{16}'
+  from: '{{ pg_password }}'
   generate: expression
   name: POSTGRESQL_PASSWORD
   required: true
@@ -173,4 +173,4 @@ parameters:
   displayName: PostgreSQL Database Name
   name: POSTGRESQL_DATABASE
   required: true
-  value: sampledb
+  value: {{ pg_database }}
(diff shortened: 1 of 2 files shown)


installer/roles/kubernetes/templates/postgresql-persistent.yml.j2
```diff
@@ -159,18 +159,18 @@ parameters:
 - description: Username for PostgreSQL user that will be used for accessing the
     database.
   displayName: PostgreSQL Connection Username
-  from: user[A-Z0-9]{3}
+  from: {{ pg_username }}
```
shanemcd, line 162: Does this actually fix the problem? Unless I'm missing something, these are still going to show up in the environment: https://github.com/ansible/awx/blob/b12d852e6d1cf491c77d44aff3fb62bf5b997003/installer/roles/kubernetes/templates/postgresql-persistent.yml.j2#L90-L94

jaymiracola: It does fix the problem. The problem was two fold. 1. the environment secure variable gets overwritten by the shell command that follows which has been removed. 2. Adding the Ansible variable to the jinja2 lets the secret get created correctly and remain. Today the postgresql secret sits there doing nothing with the default values.

jaymiracola: The variable set in the postgresql template you cited is inherited from the variable that I have changed.

shanemcd: Thanks for the explanation. I will pull this down and test it out, probably Monday, and merge if things go well.

shanemcd (approved):

rooftopcellist:
regate

ryanpetrello:
@jaymiracola sorry for the delay on this - we've since replaced this installer with an Operator (https://github.com/ansible/awx-operator https://groups.google.com/g/awx-project/c/47MjWSUQaOc/m/bCjSDn0eBQAJ) in the next major release of AWX, so I suspect this PR is obsolete now cc @rooftopcellist

jaymiracola:
Could still be useful to include in older versions. Or for those unwilling or unable to use the operator for whatever reason.

ryanpetrello:
@jaymiracola,

Potentially so.  That said, we don't generally maintain stable branches/minor releases of the AWX project.

Issue #9087, opened 2021-01-14
Update EE Add/List/Edit/Details with latest API changes


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Update EE Add/List/Edit/Details with latest API changes. There is a field name now, for instance. Evaluate the latest API changes, and update UI accordingly.

jakemcdermott:
Addressed in this PR: https://github.com/ansible/awx/pull/9179 via https://github.com/ansible/awx/pull/9179/commits/7d7a1e5095652e928f91eb5fe418db38f8ce304c

tiagodread:
@nixocio note that the field Pull has `No pull option has been selected` as a valid selection, is this the expected behaviour or this option is only the "label" for this field?

![image](https://user-images.githubusercontent.com/12464858/107686646-43912280-6c84-11eb-9d2d-59a95ec2b008.png)
![image](https://user-images.githubusercontent.com/12464858/107686674-4ab83080-6c84-11eb-9120-726299f4f872.png)

PR #9262, opened 2021-02-08
Migrate EE list to tables

nixocio:
Migrate EE list to tables.

See:https://github.com/ansible/awx/issues/7884

https://github.com/ansible/awx/issues/9087
image


jakemcdermott (approved):
:+1:

PR #9263, opened 2021-02-08
Update Job Detail container group variable

nixocio:
`is_containerized` was updated on the API side to be
`is_container_group`.

image


nixocio:
Those fields were modified in the devel branch, after those variables were renamed on the EE, that is why we missed them.

jakemcdermott (approved):

PR #9279, opened 2021-02-09
replace null unicode with nothing

chrismeyersfsu:
* Null characters "printed" in ansible are replaced by the unicode
literal "\u0000". For example command: echo -ne 'foo\0bar' in ansible
would result in the output "foo\u0000bar". AWX stores the string literal
without modification. This does not cause problems for most of the AWX
code base because AWX rarely asks Postgres to treat the TEXT field
event_data as json data. Instead, it pulls the data into memory and then
does JSON operations. The exception here is the analytics code.
* In the analytics code we ask postgres to convert the event_data text
field into json. At this point Postgres converts found literals \u0000
to the null character. This is a problem because Postgres also uses the
null character to terminate json strings. This causes postgres to blow
up.
* This changeset avoids the blowup when pulling the data out of postgres
by replacing \u0000 with '' before converting to json.

### DIFF
--- a/awx/main/analytics/collectors.py
+++ b/awx/main/analytics/collectors.py
@@ -282,32 +282,34 @@ def _copy_table(table, query, path):
 
 @register('events_table', '1.2', format='csv', description=_('Automation task records'), expensive=True)
 def events_table(since, full_path, until, **kwargs):
-    events_query = '''COPY (SELECT main_jobevent.id, 
+
+    safe_event_data = "replace(main_jobevent.event_data::text, '\\u0000', '')::json"
+    events_query = f'''COPY (SELECT main_jobevent.id,
                               main_jobevent.created,
                               main_jobevent.modified,
                               main_jobevent.uuid,
                               main_jobevent.parent_uuid,
-                              main_jobevent.event, 
-                              main_jobevent.event_data::json->'task_action' AS task_action,
+                              main_jobevent.event,
+                              {safe_event_data}->>'task_action' AS task_action,
                               (CASE WHEN event = 'playbook_on_stats' THEN event_data END) as playbook_on_stats,
-                              main_jobevent.failed, 
-                              main_jobevent.changed, 
-                              main_jobevent.playbook, 
+                              main_jobevent.failed,
+                              main_jobevent.changed,
+                              main_jobevent.playbook,
                               main_jobevent.play,
                               main_jobevent.task,
-                              main_jobevent.role, 
-                              main_jobevent.job_id, 
-                              main_jobevent.host_id, 
-                              main_jobevent.host_name
-                              , CAST(main_jobevent.event_data::json->>'start' AS TIMESTAMP WITH TIME ZONE) AS start,
-                              CAST(main_jobevent.event_data::json->>'end' AS TIMESTAMP WITH TIME ZONE) AS end,
-                              main_jobevent.event_data::json->'duration' AS duration,
-                              main_jobevent.event_data::json->'res'->'warnings' AS warnings,
-                              main_jobevent.event_data::json->'res'->'deprecations' AS deprecations
-                              FROM main_jobevent 
-                              WHERE (main_jobevent.created > '{}' AND main_jobevent.created <= '{}')
+                              main_jobevent.role,
+                              main_jobevent.job_id,
+                              main_jobevent.host_id,
+                              main_jobevent.host_name,
+                              CAST({safe_event_data}->>'start' AS TIMESTAMP WITH TIME ZONE) AS start,
+                              CAST({safe_event_data}->>'end' AS TIMESTAMP WITH TIME ZONE) AS end,
+                              {safe_event_data}->'duration' AS duration,
+                              {safe_event_data}->'res'->'warnings' AS warnings,
+                              {safe_event_data}->'res'->'deprecations' AS deprecations
+                              FROM main_jobevent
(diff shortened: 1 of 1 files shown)


chrismeyersfsu:
I can reason how this won't impact performance. We already cast the TEXT field `event_data` into JSON. This would require Postgres to read all of the data from disk. Similarly, the `replace` function would require passing over all of the data. But we already pay the penalty of pulling the data from disk because of the JSON cast.

Ideally, we would test the performance before and after.

Ladas (approved):
Lets add unit tests that is adding this bad char into the DB, then this looks great :+1: 

(I'll make sure this gets perf tested, probably on 3.7.x backport)

chrismeyersfsu:
The unit tests use sqlite. Sqlite doesn't support the json type. https://github.com/ansible/awx/blob/devel/awx/main/tests/functional/analytics/test_collectors.py#L49

I don't think we would get much value out of trying to test this in our unit/functional tests.

awx/main/analytics/collectors.py
```diff
@@ -282,32 +282,34 @@ def _copy_table(table, query, path):
 
 @register('events_table', '1.2', format='csv', description=_('Automation task records'), expensive=True)
 def events_table(since, full_path, until, **kwargs):
-    events_query = '''COPY (SELECT main_jobevent.id, 
+
+    safe_event_data = "replace(main_jobevent.event_data::text, '\\u0000', '')::json"
```
ryanpetrello, line 286: @chrismeyersfsu @Ladas is there any additional performance cost on doing a `replace()` on every row?

I'd be curious to see what an explain and timing looks like for a medium sized table with actual data here.

If there's an added cost, perhaps we can do what I suggested @chrismeyersfsu:

```
try:
    return _copy_table(...)
except ThatError as e:
    if e.errno == 'whatever':
        return _copy_table(query_with_replace)
```

chrismeyersfsu: Would the billion event workload be good for testing this?

ryanpetrello: I doubt you even need that many.

ryanpetrello:
```sql
awx=> \timing
awx=> SELECT CAST(replace(main_jobevent.event_data::text, '\\u0000', '')::json->>'start' AS TIMESTAMP WITH TIME ZONE) AS start FROM main_jobevent LIMIT 200000;
Time: 6738.434 ms (00:06.738)
awx=> SELECT CAST(main_jobevent.event_data::json->>'start' AS TIMESTAMP WITH TIME ZONE) AS start FROM main_jobevent LIMIT 200000;
Time: 5823.323 ms (00:05.823)
```

Ladas: so 15% increase and it might vary based on various sizes of the json I guess

That should be ok as sanity check perf-test (we should be loading a batch of events always, so hopefully not 1B per batch, although that can happen with the time based chunking)

ryanpetrello: I'm guessing how much it increases varies widely based on the size of `event_data`; for really large event data, the cost of a full text search could be pretty expensive, I think.  We really may want to consider avoiding doing this by default for *all* users.

Ladas: @ryanpetrello   well, we could try to catch the bad unicode failure and then try it again with this code as you say

chrismeyersfsu: Is the performance gain worth the added complexity? Specifically, when it comes to performance testing. Does this add another dimension that will always need to be verified or do we just sort of know "ok, that replace adds a 15% overhead"

Ladas: I'd like to believe that this is an outlier, in that case what @ryanpetrello says would help (we know we have some large json blobs, like playbook on stats, so the extra processing time can be much bigger)

But we don't have any metrics about the event collection failures at customer side, so it's just a guess. If this would appear very often, we'd make the collection much slower (depending how fast it'll fail first time, so up to like ~2x slower).

Would be good to identify which playbooks/roles is causing this in IT cluster, so see if it's something custom or widely used. (can we put together a query that'd print this? So Patrick can run it on the IT Cluster?)

chrismeyersfsu: `http://host/api/v2/job_events/?event_data__regex=\\u0000` **WARNING** This query will be **EXPENSIVE**
I know I used the above query. If `__contains=` instead of `__regex` works it will be more performant.

If we wanted to know about how many jobs something like https://tower.ansible.eng.rdu2.redhat.com/api/v2/jobs/?job_events__event_data__contains=\\u0000

Ladas: @chrismeyersfsu IT has a looots of events, the API query till probably timeout? We should be able to limit this to e.g. last few days? (or we can just prepare a SQL query, that will print the templates with their source, so we know here is this coming from)

chrismeyersfsu: Good idea on the date restriction @Ladas . https://tower.ansible.eng.rdu2.redhat.com/api/v2/jobs/?created__gte=2021-02-24T00:00:00.00Z&job_events__event_data__contains=%5C%5Cu0000

Salt the above to taste. Currently the date is set to jobs > `February 24, 2021`

Ladas:
@chrismeyersfsu @ryanpetrello looks like we forgot about this one? What is the status?

PR #9287, opened 2021-02-10
Disable inventory copy button when inventory has sources

mabashian:
##### SUMMARY
link #7679 

Screen Shot 2021-02-10 at 11 40 18 AM
Screen Shot 2021-02-10 at 11 40 25 AM

Had to do a small amount of refactoring because `<Tooltip>` components won't render when the direct child is a `<Button>` and the button is disabled.

Since the `<ActionItem>` component already handles tooltips I removed this logic from the copy button.  There was one remaining list where the copy button was used but we had not yet converted the list to tables (Notification List).  I had to wrap the button in a tooltip there.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/components/CopyButton/CopyButton.jsx
+++ b/awx/ui_next/src/components/CopyButton/CopyButton.jsx
@@ -3,7 +3,7 @@ import { withI18n } from '@lingui/react';
 import { t } from '@lingui/macro';
 import PropTypes from 'prop-types';
 
-import { Button, Tooltip } from '@patternfly/react-core';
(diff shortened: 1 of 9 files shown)


tiagodread (approved):
Verified that the following resources can be copied:
```
- job templates
- inventories
- notification templates
- credentials
- project 
```
Verified that inventory with source can't be copied.
🟢   signal from QE waiting for PR review

jakemcdermott (approved):

Issue #7062, opened 2020-05-18
Replace custom virtual environments with custom execution environments


ryanpetrello:
To support custom SDKs/libraries in playbook execution, AWX supports the notion of "custom virtual environments" today, whereby users:

1.  Build and deploy Python virtualenvs across their AWX infrastructure (in particular, on all `awx_task` nodes)
2.  AWX code _discovers_ these virtualenvs on the file system and exposes them as "options" for selection.
3.  Users associate these virtualenvs at the Organization, Inventory, or Job Template level so that they will be utilized for playbook runs or inventory updates.

In a future version of AWX, we'd like to **replace** this functionality with the custom execution environments described in https://github.com/ansible/awx/issues/7059

This has a few implications for migration:

1.  The process for defining these custom execution environments will be different (the thing you're producing is a container image, not a Python virtualenv), and as such it is possible to incorporate system-level dependencies and collection-based content.  We'll need good documentation on the new way of doing things, and will need to point users looking to migrate in the right direction.

2.  We may want to consider a process for aiding users in migration from "old style" virtual environments (perhaps via `pip freeze` and a generated `requirements.txt`) to the new execution environments process.

3.  In newer versions of AWX, we'll likely need a way to detect that you _have_ old virtual environments associated with Organizations/Inventory/Job Templates and warn you that you'll need to migrate to the new custom execution enviroment approach described in https://github.com/ansible/awx/issues/7059

    - In this scenario, what should the behavior be?  If you have an "old style" virtual environment, should we just use the *default* execution environment, or should AWX refuse to run the job until you've migrated away?

kdelee:
@rebeccahhh is this what you are planning on working on this first sprint?

rebeccahhh:
@kdelee yep!

rebeccahhh:
update, this is moving to a later sprint

kdelee:
@rebeccahhh where are we on this? When are we thinking we will remove it from the API? Is this feature the one that is tracking the tool that will do a pip freeze on your venvs?

rebeccahhh:
@kdelee good question, this is kind of an umbrella issue but the next thing on the list is doing the pip freeze tool. The tool is also one of the last things we have to do for this prior to this being completed for release 🎊 
Currently I have created a new API endpoint that spits out the pip freeze data for all the virtualenvs for the user to see. Still a bit buggy at the moment, but I'll have it up in a branch or draft PR today so it can be visible and others can poke it.

[...]

PR #9303, opened 2021-02-12
add copy ability to execution environments

rebeccahhh:
##### SUMMARY

Adding copy functionality/api endpoints to execution environments and wrote a test for creating an EE
this will tie into #7062 
originally also added the pull field from #9179 to AWXKit in a couple places and started work on an RBAC issue, but I moved that work into PRs: #9333 and #9346

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 17.0.1
```

### DIFF
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -710,6 +710,12 @@ class ExecutionEnvironmentJobTemplateList(SubListAPIView):
     relationship = 'unifiedjobtemplates'
 
 
+class ExecutionEnvironmentCopy(CopyAPIView):
+    
+    model = models.ExecutionEnvironment
+    copy_return_serializer_class = serializers.ExecutionEnvironmentSerializer
+
+
 class ExecutionEnvironmentActivityStreamList(SubListAPIView):
 
     model = models.ActivityStream
--- a/awx/main/tests/functional/test_execution_environments.py
+++ b/awx/main/tests/functional/test_execution_environments.py
@@ -1,0 +1,19 @@
+import pytest
+
+from awx.main.models import (ExecutionEnvironment)
+
+
+@pytest.mark.django_db
+def test_execution_environment_creation(execution_environment, organization):
+    execution_env = ExecutionEnvironment.objects.create(
+        name='Hello Environment', 
+        image='', 
+        organization=organization, 
+        managed_by_tower=False, 
+        credential=None,
+        pull='missing'
+    )
+    assert type(execution_env) is type(execution_environment)
+    assert execution_env.organization == organization
+    assert execution_env.name == 'Hello Environment'
+    assert execution_env.pull == 'missing'
--- a/awxkit/awxkit/api/pages/execution_environments.py
+++ b/awxkit/awxkit/api/pages/execution_environments.py
@@ -1,6 +1,6 @@
 import logging
 
-from awxkit.api.mixins import DSAdapter, HasCreate
+from awxkit.api.mixins import DSAdapter, HasCreate, HasCopy
 from awxkit.api.pages import (
     Credential,
     Organization,
@@ -15,7 +15,7 @@ from . import page
 log = logging.getLogger(__name__)
 
 
-class ExecutionEnvironment(HasCreate, base.Base):
+class ExecutionEnvironment(HasCreate, HasCopy, base.Base):
 
     dependencies = [Organization, Credential]
     NATURAL_KEY = ('name',)
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -1373,6 +1373,7 @@ class ExecutionEnvironmentSerializer(BaseSerializer):
         res.update(
             activity_stream=self.reverse('api:execution_environment_activity_stream_list', kwargs={'pk': obj.pk}),
             unified_job_templates=self.reverse('api:execution_environment_job_template_list', kwargs={'pk': obj.pk}),
+            copy=self.reverse('api:execution_environment_copy', kwargs={'pk': obj.pk}),
         )
         if obj.organization:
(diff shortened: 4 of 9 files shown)


awx/main/tests/functional/test_execution_environments.py
```diff
@@ -0,0 +1,15 @@
+import pytest
+from unittest import mock
+from rest_framework.exceptions import PermissionDenied
+
+from awx.main.models import (ExecutionEnvironment, Organization)
+
+
+@pytest.mark.django_db
+def test_execution_environment_creation(execution_environment, organization):
+    execution_env = ExecutionEnvironment.objects.create(name='Hello Environment', image='', organization=organization, managed_by_tower=False, credential=None, pull='missing')
```
nixocio: @rebeccahhh can one copy an EE `managed_by_tower`?

awx/api/views/__init__.py
```diff
@@ -708,6 +708,12 @@ class ExecutionEnvironmentJobTemplateList(SubListAPIView):
     relationship = 'unifiedjobtemplates'
 
 
+class ExecutionEnvironmentCopy(CopyAPIView):
+
```
nixocio, line 714: Are we updating `summary_fields` to add the copy field as part of the `user_capabilities`?

shanemcd:
Can we rename this PR to be a bit more descriptive?

awxkit/awxkit/api/pages/execution_environments.py
```diff
@@ -15,15 +15,15 @@
...
-class ExecutionEnvironment(HasCreate, base.Base):
+class ExecutionEnvironment(HasCreate, HasCopy, base.Base):
 
     dependencies = [Organization, Credential]
     NATURAL_KEY = ('name',)
 
-    # fields are name, image, organization, managed_by_tower, credential
-    def create(self, name='', image='quay.io/ansible/ansible-runner:devel', credential=None, **kwargs):
+    # fields are name, image, organization, managed_by_tower, credential, pull
+    def create(self, name='', image='quay.io/ansible/ansible-runner:devel', credential=None, pull='missing', **kwargs):
```
shanemcd: We should almost always use `None` as a default kwarg value. See this for more info: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

I see that we were not doing that here for `image`, we should fix that too.

awxkit/awxkit/api/pages/execution_environments.py
```diff
@@ -33,13 +33,13 @@ def create_payload(self, name='', organization=Organization, **kwargs):
         payload.ds = DSAdapter(self.__class__.__name__, self._dependency_store)
         return payload
 
-    def payload(self, name='', image=None, organization=None, credential=None, **kwargs):
+    def payload(self, name='', image=None, organization=None, credential=None, pull='missing', **kwargs):
```
shanemcd: Same as my comment above, the default value here should be `None`.

awx/main/tests/functional/test_execution_environments.py
rebeccahhh: @nixocio I'm not sure, @jbradberry do you know?

jbradberry: @nixocio there's no rbac logic that prevents it.  The expectation is that (some) global EEs are going to be the ones that are `managed_by_tower=True`, but there's nothing currently stopping users from creating global ones without it or Org-attached ones with it.  Anyway, for the purposes of rbac, my expectation is that if you can see it you can copy it, as long as you have sufficient privileges to do so in the current context (e.g. you have an Org Admin role).

... that raises an interesting question:  should an Org Admin be able to copy a global EE into their Org, and have that Org attachment happen automatically as part of the copy operation?

awx/api/views/__init__.py
jbradberry: Good call, we should do that.

awx/main/tests/functional/test_execution_environments.py
nixocio: Thanks for your explanation @jbradberry. Regarding the last question an Org Admin can be the admin of more than one Org, then automatically attaching the Org will not be a good fit, right? Right now for Org Admins to create an EE they must provide an Org.

awxkit/awxkit/api/pages/execution_environments.py
jbradberry: Python strings aren't mutable, and are perfectly fine as default function arguments.

rebeccahhh: pulled out into #9333

awx/api/views/__init__.py
rebeccahhh: done!

jbradberry (approved):
This looks reasonable to me.

shanemcd (approved):
Amazing!

awxkit/awxkit/api/pages/execution_environments.py
```diff
@@ -39,8 +39,6 @@ def payload(self, name='', image=None, organization=None, credential=None, pull=
             image=image or random_title(10),
             organization=organization.id if organization else None,
             credential=credential.id if credential else None,
-            pull=pull,
-            **kwargs
```
rebeccahhh: I think these 2 lines must have been deleted when I was cherry-picking commits

jbradberry:
Making sure to add this here, so it doesn't get missed -- we should allow for the copy of managed EEs, but copying such an EE should result in a new EE that has `managed_by_tower=False`.

shanemcd (approved):

Issue #9165, opened 2021-01-26
Cant sync ovh inventory on AWX 16.0.0 and 17.0.0 with region name


bdoublet91:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Try to configure openstack synchronisation with new AWX version 17.0.0. My region is GRA7 but only download servers from UK1 like version and get error 401 in logs
Same behavior with version 16.0.0

##### ENVIRONMENT
* AWX version: 17.0.0
* AWX install method: docker on linux
* Ansible version:  2.9.16
* Operating System: Ubuntu (Awx official docker image)
* Web Browser: Firefox

##### STEPS TO REPRODUCE

Create openstack credential and specify region GRA7
![image](https://user-images.githubusercontent.com/57101863/105751261-10d9f100-5f46-11eb-89c6-320ce63c57e0.png)
Username and password are the same when I connect through horizon portal, I have also application Key with secret and consumer key. Tried all

##### EXPECTED RESULTS

Get 20 servers from GRA7 region

##### ACTUAL RESULTS

Get 0 server from UK1 region in logs but tested with one server on UK1 and it worked

##### ADDITIONAL INFORMATION

Create a server in UK1 region and setup credential without region name
20 servers in GRA7region

**First test: **
No region name
Curl all regions and get 1 server from UK1 region and 0 server from GRA7
Get error 401 and warning

```
REQ: curl -g -i -X GET https://compute.uk1.cloud.ovh.net/v2.1 -H "Accept: application/json" -H "User-Agent: openstacksdk/0.37.0 keystoneauth1/3.18.0 python-requests/2.22.0 CPython/3.6.8"
RESP: [401] Content-Length: 114 Content-Type: application/json Date: Tue, 26 Jan 2021 13:29:13 GMT Strict-Transport-Security: max-age=15768000 Www-Authenticate: Keystone uri='https://auth.cloud.ovh.net:443/' X-Compute-Request-Id: req-7c7adec3-fea9-4c0c-aa1d-ffb8b31e4247 X-IPLB-Instance: 24841
RESP BODY: {"error": {"message": "The request you have made requires authentication.", "code": 401, "title": "Unauthorized"}}

Parsed /tmp/awx_25_so4g1bx_/openstack.yml inventory source with auto plugin
    5.714 INFO     Processing JSON output...
    5.716 DEBUG    Loaded group: all
    5.716 DEBUG    Loaded group: _nova
[...]

PR #8880, opened 2020-12-13
Related #5080 openstack region field missing

vguillard:
##### SUMMARY

- Add a 'Region Name' textfield for OpenStack credentials, as mentionned by [ansible-collections-openstack](https://github.com/openstack/ansible-collections-openstack/blob/8255ec4c80f186aa7851f023b85007a593b6f42f/scripts/inventory/openstack.yml#L14)
- Set English and French locales
- Add one specific test (test_openstack_client_config_generation_with_project_region_name)

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API
 - UI

##### AWX VERSION

```
awx: 16.0.0
```

##### ADDITIONAL INFORMATION

```
$ make test && make ui-test-ci && echo "OK"
make: Nothing to be done for 'ui-test-ci'.
OK

```

PR #9304, opened 2021-02-12
Fix Openstack credential region implementation.

wenottingham:
See https://github.com/ansible/awx/issues/9165

This fixes a bug introduced in https://github.com/ansible/awx/pull/8880/files when this code was added.

Not tested yet, but I assert that the prior code can't possibly have worked, as the injector/tests/UI were not using the same credential field name as the model (see https://github.com/ansible/awx/blob/devel/awx/main/models/credential/__init__.py#L823).

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API
 - UI

##### AWX VERSION
current

### DIFF
--- a/awx/main/models/credential/injectors.py
+++ b/awx/main/models/credential/injectors.py
@@ -92,8 +92,8 @@ def _openstack_data(cred):
         },
     }
 
-    if cred.has_input('project_region_name'):
-        openstack_data['clouds']['devstack']['region_name'] = cred.get_input('project_region_name', default='')
+    if cred.has_input('region'):
+        openstack_data['clouds']['devstack']['region_name'] = cred.get_input('region', default='')
 
     return openstack_data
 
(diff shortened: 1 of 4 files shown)


vguillard:
This PR fix my code

Inputs:
image

File generated :
```
sh-4.4$ while [ 1 ]; do cat /tmp/awx_*/tmp* ; sleep 2; done
cat: '/tmp/awx_*/tmp*': No such file or directory
cat: '/tmp/awx_*/tmp*': No such file or directory
clouds:
  devstack:
    auth:
      auth_url: foo_url.txt
      password: api_key
      project_name: my_tenant_name
      username: Foo_user
    private: true
    region_name: Some_regions
    verify: true
```

ryanpetrello (approved):
@wenottingham this seems pretty reasonable and straightforward to me if somebody has cycles to verify it cc @kdelee @elyezer @jneedle

PR #9306, opened 2021-02-13
Allow dependabot to check python dependencies

jlosito:
##### SUMMARY

This change allows dependabot to check python dependencies which this project uses and submit pull requests with version bumps in order to keep packages up-to-date.

##### ISSUE TYPE

 - Chore Pull Request

##### COMPONENT NAME

 - API
 - UI
 - Installer

##### ADDITIONAL INFORMATION

Here's a [link](https://docs.github.com/en/github/administering-a-repository/keeping-your-actions-up-to-date-with-dependabot) to dependabot's docuemntation.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/requirements"
+    schedule:
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,4 +4,4 @@ updates:
   - package-ecosystem: "pip"
     directory: "/requirements"
     schedule:
-      interval: "weekly"
+      interval: "monthly"


shanemcd:
Hello. Thanks for the PR. We use `pip-compile` to generate our `requirements.txt` files. (Note the file `requirements/requirements.in`). It looks like dependabot supports this: https://dependabot.com/blog/dependabot-now-supports-pip-compile/

Could you please include an entry for `package-ecosystem: "pip-compile"`, in addition to what you already have?

jlosito:
The dependabot documentation states that for pip-compile, the yaml value will still be pip. In other words, there doesn't exist a pip-compile package ecosystem value only pip which covers pip, pipenv, pip-compile, and poetry.

https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates

unlikelyzero:
@jlosito thanks for the PR!

Is there a specific dependency you think we should watch first?

My experience with dependabot is that it becomes quite noisy at first and can take some time to adjust. It might make sense to start small with a single dependency and ignore the rest.

In terms of UI dependencies, I think the patternfly dependencies might be a good place to start. 

What do you think?

jlosito:
Maybe the bot should be adjusted to a monthly basis so that things can converge more gradually. Some packages get updated much more frequently than others. A monthly basis would allow one to make sure the less frequently updated dependencies can be properly tested and integrated while ignoring the other packages that may have had several minor releases during the month. I agree that the bot can be quite noisy sometimes depending on how often it is scheduled to check.

I personally don't have a preference as to which component should be updated first. You might have much more insight than I have regarding that topic.

shanemcd:
I agree that weekly may be too noisy. We just don't have the capacity to look at, or test version bumps every week. 

One of the problems I have with this kind of approach is that it assumes all dependency upgrades are inherently valuable and more secure. From our experience in the open source community, this just isn't the case. 

We do monitor our dependencies, and we do address any CVEs that pop up pretty much immediately.

That being said, I'm not totally against this PR, but at least for now let's move it to monthly.

shanemcd (approved):

PR #9309, opened 2021-02-14
Tower workflow schema

sean-m-sullivan:
##### SUMMARY
Adds a way to add entire workflow node schemas to workflows. Either through the workflow schema module or the workflow job template module. 

This speeds up workflow creation vs the workflow node module by 3x. 
The model for the schemas is the format used by the tower_export module. 

The main difference between this and the workflow node module is that the loops are done in python. Traditionally if you have a workflow with 10 nodes, ansible tasks need to be invoked 19 times. 1x to create the workflow, 10 x to initially create the nodes, and then one time for each node that is not an endpoint in the schema. This removes the need to loop and invoke many times. 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
awx-collection

##### AWX VERSION
17.0.1


awx_collection/plugins/modules/tower_workflow_job_template_schema.py
```diff
@@ -0,0 +1,347 @@
+#!/usr/bin/python
+# coding: utf-8 -*-
+
+
+# (c) 2020, John Westcott IV <john.westcott.iv@redhat.com>, Sean Sullivan <ssulliva@redhat.com>
```
beeankha: Change :copyright: year to 2021

sean-m-sullivan: Updated.

awx_collection/plugins/modules/tower_workflow_job_template.py
```diff
@@ -142,11 +300,66 @@
...
+    name: example-workflow
+    inventory: Demo Inventory
+    extra_vars: {'foo': 'bar', 'another-foo': {'barz': 'bar2'}}
+    schema:
+      - identifier: node101
+        unified_job_template:
+          name: example-project
+          inventory:
+            organization:
+              name: Default
```
kdelee, line 347: is the indentation correct here? looks a bit wonky
I don't understand why the inventory looks like a sub-item of unified_job_template or why organization is a sub item of inventory

kdelee: :mag: ok I think I get it....you are trying to indicate the organization of the unified job template indirectly by saying use the JT with name `example-project` that has an inventory (not named) that is in the Default organization.

Is that right?

sean-m-sullivan: This is for an inventory sync update in a workflow. It's actually the direct output of tower_export and the way it keeps it in the api. The goal is to reduce looping to create nodes of a workflow, and it follows the model from tower_export, which follows how it keeps things in the API. The fact that organization is referenced differently for inventories vs Job templates is just the way current tower api works.

beeankha (approved):

john-westcott-redhat:
@sean-m-sullivan, How does tower_workflow_job_template_schema.py compare to the existing tower_workflow_job_template_node.py? Also, if there is a module for managing the schema why support it in tower_workflow_job_template?

sean-m-sullivan:
> How does tower_workflow_job_template_schema.py compare to the existing tower_workflow_job_template_node.py? Also, if there is a module for managing the schema why support it in tower_workflow_job_template?

TBH I added the separate module just to add options, it could be done with just the existing tower_workflow_job_template. 

In terms of comparison to tower_workflow_job_template_node, the difference is that node does a single node creation, link, OR modify existing settings/links, The code is much the same, except it adds the ability to loop over nodes, create or make sure they exist, then link them, in a single ansible module for an entire schema. As mentioned above, its meant to reduce 18 calls to tower_workflow_job_template_node and a call to tower_workflow_job_template to a single call,

ryanpetrello:
@beeankha @john-westcott-iv is this PR ready for another set of eyeballs, or is there still unaddressed feedback?

sean-m-sullivan:
@ryanpetrello I had talked with Beeankha about this, and we were waiting on Alan Coding to come back to do a review, in addition I found an bug where we didn't address approval nodes, and will try to get that in. In addition, the copy PR #9374 somehow got mixed in here, and I need to extract it.

sean-m-sullivan:
Regate

awx_collection/test/awx/test_completeness.py
```diff
@@ -38,7 +38,7 @@
     'tower_workflow_launch',
     'tower_job_cancel',
     'tower_workflow_template',
-    'tower_ad_hoc_command_wait',
+    'tower_workflow_job_template_schema' 'tower_ad_hoc_command_wait',
```
beeankha: Looks like there's a missing `,` and newline after `'tower_workflow_job_template_schema'`

awx_collection/test/awx/test_completeness.py
```diff
@@ -59,8 +59,8 @@
     # Organization is how we are looking up job templates, Approval node is for workflow_approval_templates
     'tower_workflow_job_template_node': ['organization', 'approval_node'],
     # Survey is how we handle associations
-    'tower_workflow_job_template': ['survey_spec'],
-    # ad hoc commands support interval and timeout since its more like tower_job_launch
+    'tower_workflow_job_template': ['survey_spec', 'destroy_current_schema'],
+    # ad hoc commands support interval and timeout since its more like tower_job_launc
```
beeankha: An `h` got deleted at the end of this comment line

john-westcott-iv:
With the changes you've made on this do we still need to introduce the new module tower_workflow_job_template_schema.py or can we just have tower_workflow_job_template? If we still need the second one, can you summarize the difference between them?

sean-m-sullivan:
@john-westcott-iv 
Part of it was that I created the tower_workflow_job_template_schema for testing, and its meant to mimic how people use the tower_workflow_job_template_node, where it only edits the schema. I am in favor of just removing it. I kept it, on the off chance it was decided to only have it as its own module.

We can just have tower_workflow_job_template changes. The goal was to make it functionally work like tower_workflow_template that we are depreciating, and to put the looping needed to create a schema in python rather then what is the current method of using ansible loops.

awx_collection/test/awx/test_completeness.py
```diff
@@ -38,6 +38,7 @@
     'tower_workflow_launch',
     'tower_job_cancel',
     'tower_workflow_template',
+    'tower_workflow_job_template_schema',
```
john-westcott-iv: This can be removed now right?

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -740,3 +743,179 @@ def wait_on_url(self, url, object_name, object_type, timeout=30, interval=10):
     def wait_output(self, response):
         for k in ('id', 'status', 'elapsed', 'started', 'finished'):
             self.json_output[k] = response['json'].get(k)
+
+    def create_schema_nodes(self, response, schema, workflow_id):
```
john-westcott-iv: Since we only have one module now can we move this logic into the module instead of the tower_api library?

john-westcott-iv (changes requested):

john-westcott-iv (approved):

Issue #9200, opened 2021-02-01
awx.awx.tower_job_template will update the limit to none if the parameter is not set


xding2020:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

tower_job_template ansible module will update the limit of job template to none if the parameter is not set, instead of keeping it unchanged

##### ENVIRONMENT
* Ansible version:  2.9.4

##### STEPS TO REPRODUCE
1. Assuming ansible tower is ready and there is a job template with limit parameter set
2. Run an ansible playbook that contains a task updating the job template (but not updating the limit). e.g.:

```yaml
  - name: update template with corresponding venv
    awx.awx.tower_job_template:
      name: "{{ template_name }}"
      custom_virtualenv: "/var/lib/awx/venv/venv-test"
```

##### EXPECTED RESULTS
The job template is updated with limit unchanged

##### ACTUAL RESULTS
The limit of job template is set to none

blomquisg:
@john-westcott-iv I think you've seen this type of issue before.  Can you give some feedback here, or point to another issue where you've talked about this?

john-westcott-iv:
In this case we have a legitimate bug. We have a default of '' on the field (there are a couple) and later we are checking to see if the field is not pythons `None`. This will never be the case because we gave a default of ''. So, if you don't specify limit (or one of the other fields) the module will kindly set the limit as '' instead of leaving it alone.

sean-m-sullivan:
I found the same behaviour happening within other modules and fields, Submitted PR to fix bug
```
plugins/modules/tower_credential_input_source.py:79:        description=dict(default=''),
plugins/modules/tower_credential_input_source.py:82:        source_credential=dict(default=''),
plugins/modules/tower_job_template.py:343:        description=dict(default=''),
plugins/modules/tower_job_template.py:349:        credential=dict(default=''),
plugins/modules/tower_job_template.py:350:        vault_credential=dict(default=''),
plugins/modules/tower_job_template.py:357:        job_tags=dict(default=''),
plugins/modules/tower_job_template.py:359:        skip_tags=dict(default=''),
plugins/modules/tower_job_template.py:360:        start_at_task=dict(default=''),
plugins/modules/tower_project.py:232:        scm_branch=dict(default=''),
plugins/modules/tower_project.py:233:        scm_refspec=dict(default=''),
plugins/modules/tower_ad_hoc_command.py:121:        module_args=dict(default=""),
```

sean-m-sullivan:
[...]

PR #9313, opened 2021-02-14
Fix defaults that were set to empty string in collections.

sean-m-sullivan:
##### SUMMARY
Fix several places where an empty string was set as default instead of nothing. This is related to #9200 . Currently if user does not set the field, but it was set in tower, it would overwrite the value with an empty string. With this change the original value stays unless specified. 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - awx-collection

##### AWX VERSION
17.0.1


awx_collection/plugins/modules/tower_ad_hoc_command.py
```diff
@@ -118,7 +118,7 @@ def main():
         limit=dict(),
         credential=dict(required=True),
         module_name=dict(required=True),
-        module_args=dict(default=""),
+        module_args=dict(),
```
john-westcott-redhat, line 121: Can you confirm there is a test on this one? Down on like 148 module_args is used to build a post_data without a conditional `if module_args`

awx_collection/plugins/modules/tower_project.py
```diff
@@ -229,8 +229,8 @@ def main():
         scm_type=dict(choices=['manual', 'git', 'svn', 'insights'], default='manual'),
         scm_url=dict(),
         local_path=dict(),
-        scm_branch=dict(default=''),
-        scm_refspec=dict(default=''),
+        scm_branch=dict(),
+        scm_refspec=dict(),
```
john-westcott-redhat, line 233: Also make sure there are tests for these as well. Lines 322/323 are hard coding these into the project_files unconditionally.

john-westcott-redhat (changes requested):

awx_collection/plugins/modules/tower_project.py
sean-m-sullivan: Checked this one, and we normally don't reference them, so looks like tests passed. https://github.com/ansible/awx/blob/4e4811870418b17ec173bd7651b675003abb572a/awx_collection/tests/integration/targets/tower_project/tasks/main.yml#L78

awx_collection/plugins/modules/tower_ad_hoc_command.py
sean-m-sullivan: I'm working on this one, have to make a test for it.

sean-m-sullivan: Test added, does not look to affect the module.

beeankha (approved):

Issue #8852, opened 2020-12-10
[ui_next] Clicking on Text within a modal doesn't select but DOES issue api requests


unlikelyzero:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Clicking on Text within a modal doesn't select but DOES issue api requests.

##### ENVIRONMENT
* AWX version: 12/10

##### STEPS TO REPRODUCE
1. Open modal with checkboxes like Add Team Roles to User
2. Click on the name of the resource or object in the list
3. Object's checkbox not selected but API requests fired off.

##### EXPECTED RESULTS
Selected items should be selected

##### ACTUAL RESULTS
Not selected with checkbox

##### ADDITIONAL INFORMATION
<img width="630" alt="Screen Shot 2020-12-10 at 6 15 00 PM" src="https://user-images.githubusercontent.com/12446869/101841296-a72b9200-3b13-11eb-9756-d834140a9673.png">

unlikelyzero:
@tiagodread i'm positive that you've filed this one already but cannot find it

unlikelyzero:
Note, this is filed here: 
https://github.com/ansible/awx/issues/7220 
https://github.com/ansible/awx/issues/7510 
https://github.com/ansible/awx/issues/7508

and was reported in each bug bash.

I'm going to keep this one open and close the others

tiagodread:
PR already merged working as expected!

PR #9316, opened 2021-02-15
Fix deadspots on multiselect items

jakemcdermott:
##### SUMMARY
for #8852

### DIFF
--- a/awx/ui_next/src/components/CheckboxListItem/CheckboxListItem.jsx
+++ b/awx/ui_next/src/components/CheckboxListItem/CheckboxListItem.jsx
@@ -27,7 +27,7 @@ const CheckboxListItem = ({
       aria-labelledby={`check-action-item-${itemId}`}
       id={`${itemId}`}
     >
-      <DataListItemRow>
+      <DataListItemRow onClick={isSelected ? onDeselect : onSelect}>
         <CheckboxRadio
           aria-label={`check-action-item-${itemId}`}
           aria-labelledby={`check-action-item-${itemId}`}
@@ -44,7 +44,6 @@ const CheckboxListItem = ({
             <DataListCell key="name">
               <label
                 id={`check-action-item-${itemId}`}
-                htmlFor={`selected-${itemId}`}
                 className="check-action-item"
               >
                 <b>{label}</b>


awx/ui_next/src/components/CheckboxListItem/CheckboxListItem.jsx
```diff
@@ -27,7 +27,7 @@ const CheckboxListItem = ({
       aria-labelledby={`check-action-item-${itemId}`}
       id={`${itemId}`}
     >
-      <DataListItemRow>
+      <DataListItemRow onClick={isSelected ? onDeselect : onSelect}>
```
jakemcdermott, line 30: for deadspot between label and checkbox

awx/ui_next/src/components/CheckboxListItem/CheckboxListItem.jsx
```diff
@@ -44,7 +44,6 @@ const CheckboxListItem = ({
             <DataListCell key="name">
               <label
                 id={`check-action-item-${itemId}`}
-                htmlFor={`selected-${itemId}`}
```
jakemcdermott, line 47: for deadspot on label itself

jakemcdermott, line 47: cc @tiagodread I think removing this would be fine for both a11y and test selectors but I wanted to double-check.

jakemcdermott:
recheck

awx/ui_next/src/components/ResourceAccessList/__snapshots__/ResourceAccessListItem.test.jsx.snap
```diff
@@ -840,6 +840,7 @@ exports[`<ResourceAccessListItem /> initially renders succesfully 1`] = `
                                                         isClosable={false}
                                                         numChips={5}
                                                         onClick={[Function]}
+                                                        onOverflowChipClick={[Function]}
```
jakemcdermott: May not be related to the (now resolved) zuul failures. Nevertheless, `jest` thinks these snapshots should be updated so I did it.

AlexSCorey (approved):
👍

awx/ui_next/src/components/CheckboxListItem/CheckboxListItem.jsx
tiagodread: @jakemcdermott from the e2e tests we're good, for a11y, I'm not _master_ (I'm still learning) but I think doing so isn't a good idea because:

> The label text is not only visually associated with its corresponding text input; it is programmatically associated with it too. This means that, for example, a screen reader will read out the label when the user is focused on the form input, making it easier for an assistive technology user to understand what data should be entered.

and 

> You can click the associated label to focus/activate the input, as well as the input itself. This increased hit area provides an advantage to anyone trying to activate the input, including those using a touch-screen device.

Another thing to consider is about these [Accessibility concerns](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#accessibility_concerns)

> Don't place interactive elements such as anchors or buttons inside a label. Doing so makes it difficult for people to activate the form input associated with the label.

Running the axe tool against a modal for example was found 2 issues none of them related to `labels`:

image

image

tiagodread (commented):
@jakemcdermott added a comment about htmlFor

Issue #9329, opened 2021-02-16
Log correlation IDs for requests


JonasKs:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Implement [`django-guid`](https://github.com/snok/django-guid) to attach a correlation ID for all logs related to a request.
I think it's best described in the README of the package, but attaching a Correlation ID to every log message makes debugging easier. We implement it manually for our running AWX container, but I think everyone would benefit from this.  

Disclaimer: I'm the author of the package. 

If accepted, I would love to raise a PR.

ryanpetrello:
Hey @JonasKs - this sounds really interesting - if you've got an idea of what a PR would look like, I'd be very interested in looking it over and trying it out.

ryanpetrello:
@JonasKs I'm curious what your approach would look like, but I've managed to get something minimal working (that I could turn into a PR) by tinkering a bit - including integrating w/ our task management code:

![image](https://user-images.githubusercontent.com/214912/108098701-ad794580-7051-11eb-99c2-13b1f9227d64.png)

JonasKs:
Thanks @ryanpetrello! I'll submit a draft this weekend/early next week and tag you in it. 😊

ryanpetrello:
@JonasKs I'd love your feedback on this (still testing a bit and making sure it works w/ our external logging aggregator support):

https://github.com/ansible/awx/pull/9332

JonasKs:
Hey that's awesome @ryanpetrello ! Sorry I didn't see the last reply from you - I forgot to refresh before I wrote a reply. 

I don't really have anything in mind, so you're welcome to implement a PR yourself!

ryanpetrello:
Hey @JonasKs, this is a good idea, and this library is really simple to integrate with 😄, so I gave it a stab.

I'd love your feedback on the approach in that PR.

JonasKs:
I'll have a look first thing in the morning when I'm on a computer again! Thanks for the kind feedback. 🎉 

I recommend checking out this [integration](https://django-guid.readthedocs.io/en/latest/integrations.html#integration-settings) setting if you haven't already. 😄

kdelee:
This is working as expected, thanks!

```
2021-03-26 19:58:58,811 WARNING  [afc7bda11fb14b89b735d1b18eb3b3b0] awx.main.dispatch scaling down worker pid:2129587
[...]

PR #9332, opened 2021-02-16
add a per-request GUID and log as it travels through background services

ryanpetrello:
see: https://github.com/ansible/awx/issues/9329

cc @gamuniz @tchellomello

### DIFF
--- a/awx/main/dispatch/periodic.py
+++ b/awx/main/dispatch/periodic.py
@@ -6,6 +6,7 @@ from multiprocessing import Process
 from django.conf import settings
 from django.db import connections
 from schedule import Scheduler
+from django_guid.middleware import GuidMiddleware
 
 from awx.main.dispatch.worker import TaskWorker
 
@@ -35,6 +36,7 @@ class Scheduler(Scheduler):
                         # If the database connection has a hiccup, re-establish a new
                         # connection
                         conn.close_if_unusable_or_obsolete()
+                    GuidMiddleware.set_guid(GuidMiddleware._generate_guid())
                     self.run_pending()
                 except Exception:
                     logger.exception(
--- a/awx/main/dispatch/pool.py
+++ b/awx/main/dispatch/pool.py
@@ -16,6 +16,7 @@ from queue import Full as QueueFull, Empty as QueueEmpty
 from django.conf import settings
 from django.db import connection as django_connection, connections
 from django.core.cache import cache as django_cache
+from django_guid.middleware import GuidMiddleware
 from jinja2 import Template
 import psutil
 
@@ -445,6 +446,8 @@ class AutoscalePool(WorkerPool):
             return super(AutoscalePool, self).up()
 
     def write(self, preferred_queue, body):
+        if 'guid' in body:
+            GuidMiddleware.set_guid(body['guid'])
         try:
             # when the cluster heartbeat occurs, clean up internally
             if isinstance(body, dict) and 'cluster_node_heartbeat' in body['task']:
--- a/awx/main/dispatch/publish.py
+++ b/awx/main/dispatch/publish.py
@@ -5,6 +5,7 @@ import json
 from uuid import uuid4
 
 from django.conf import settings
+from django_guid.middleware import GuidMiddleware
 
 from . import pg_bus_conn
 
@@ -83,6 +84,9 @@ class task:
                     'kwargs': kwargs,
                     'task': cls.name
                 }
+                guid = GuidMiddleware.get_guid()
+                if guid:
+                    obj['guid'] = guid
                 obj.update(**kw)
                 if callable(queue):
                     queue = queue()
--- a/awx/main/dispatch/worker/callback.py
+++ b/awx/main/dispatch/worker/callback.py
@@ -9,6 +9,7 @@ from django.conf import settings
 from django.utils.timezone import now as tz_now
 from django.db import DatabaseError, OperationalError, connection as django_connection
 from django.db.utils import InterfaceError, InternalError
+from django_guid.middleware import GuidMiddleware
 
 import psutil
 
@@ -152,6 +153,8 @@ class CallbackBrokerWorker(BaseWorker):
 
                 if body.get('event') == 'EOF':
                     try:
+                        if 'guid' in body:
+                            GuidMiddleware.set_guid(body['guid'])
                         final_counter = body.get('final_counter', 0)
                         logger.info('Event processing is finished for Job {}, sending notifications'.format(job_identifier))
                         # EOF events are sent when stdout for the running task is
@@ -176,6 +179,8 @@ class CallbackBrokerWorker(BaseWorker):
                             handle_success_and_failure_notifications.apply_async([uj.id])
                     except Exception:
                         logger.exception('Worker failed to emit notifications: Job {}'.format(job_identifier))
+                    finally:
+                        GuidMiddleware.set_guid('')
                     return
 
                 event = cls.create_from_data(**body)
--- a/awx/main/dispatch/worker/task.py
+++ b/awx/main/dispatch/worker/task.py
@@ -6,6 +6,8 @@ import traceback
 
 from kubernetes.config import kube_config
 
+from django_guid.middleware import GuidMiddleware
+
 from awx.main.tasks import dispatch_startup, inform_cluster_of_shutdown
 
 from .base import BaseWorker
@@ -52,6 +54,8 @@ class TaskWorker(BaseWorker):
         uuid = body.get('uuid', '<unknown>')
         args = body.get('args', [])
         kwargs = body.get('kwargs', {})
+        if 'guid' in body:
+            GuidMiddleware.set_guid(body.pop('guid'))
         _call = TaskWorker.resolve_callable(task)
         if inspect.isclass(_call):
             # the callable is a class, e.g., RunJob; instantiate and
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -34,6 +34,7 @@ from django.contrib.auth.models import User
 from django.utils.translation import ugettext_lazy as _, gettext_noop
 from django.core.cache import cache
 from django.core.exceptions import ObjectDoesNotExist
+from django_guid.middleware import GuidMiddleware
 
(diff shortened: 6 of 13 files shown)


ryanpetrello:
```shell
~/dev/awx curl -sik "https://awx.example.org/api/v2/job_templates/7/launch/" -X POST | grep "Request-Id"
X-API-Request-Id: f6f7827e250d439dadea3e95cdaa5032
Access-Control-Expose-Headers: X-API-Request-Id
```

image

(later...)

image

awx/main/dispatch/periodic.py
```diff
@@ -41,6 +42,7 @@ def run():
                         'encountered an error while scheduling periodic tasks'
                     )
                 time.sleep(idle_seconds)
+                GuidMiddleware.get_guid()
```
ryanpetrello, line 45: Every time the periodic background scheduler wakes up, get a GUID so it's available in the current thread when `.apply_async()` is called to schedule work.

awx/main/dispatch/publish.py
```diff
@@ -83,6 +84,11 @@ def apply_async(cls, args=None, kwargs=None, queue=None, uuid=None, **kw):
                     'kwargs': kwargs,
                     'task': cls.name
                 }
+                correlation_id = GuidMiddleware.get_guid()
```
ryanpetrello, line 87: When `.apply_async()` is called, get the latest GUID from the current thread, *or* generate a new one on the fly (in case it's called from some other non-HTTP request context, like the task manager).

awx/main/dispatch/worker/task.py
```diff
@@ -52,6 +54,8 @@ def run_callable(self, body):
         uuid = body.get('uuid', '<unknown>')
         args = body.get('args', [])
         kwargs = body.get('kwargs', {})
+        if 'correlation_id' in body:
+            GuidMiddleware.set_guid(body.pop('correlation_id'))
```
ryanpetrello: When a dispatcher child process deserializes some message, if there's a GUID set, apply it to the current thread so all log messages in this process inherit it.

awx/main/utils/filters.py
```diff
@@ -364,3 +367,11 @@ def query_from_string(cls, filter_string):
             return res[0].result
 
         raise RuntimeError("Parsing the filter_string %s went terribly wrong" % filter_string)
+
+
+
+class DefaultCorrelationId(CorrelationId):
+
+    def filter(self, record):
+        record.correlation_id = GuidMiddleware.get_guid() or '-'
```
ryanpetrello: Slight modification - I didn't like that the default filter printed `[None]` in various places.

ryanpetrello:
Here it is connected to our external log aggregator support (I'm running a local Flask server that just prints the POST body):

image

awx/main/dispatch/periodic.py
```diff
@@ -35,7 +36,11 @@ def run():
                         # If the database connection has a hiccup, re-establish a new
                         # connection
                         conn.close_if_unusable_or_obsolete()
-                    self.run_pending()
+                    GuidMiddleware.set_guid(GuidMiddleware._generate_guid())
```
ryanpetrello, line 39: Every time the periodic background scheduler wakes up, set a new GUID so it's available in the current thread when `.apply_async()` is called to schedule work.  In this way, we can map the periodic timer w/ tasks that get scheduled (so you can see that a task was initiated via our periodic task *or* that it originated in some HTTP request)

fosterseth:
maybe we can consider lowering the UUID length for settings/development.py
```
DJANGO_GUID = {
    'UUID_LENGTH': 32,
}
```

JonasKs:
I don't know the AWX codebase very well, but this seems very slick to me. :tada: Very fun to see you jump on this so quickly 😄 Thanks for commenting your own code here, it made it very easy for me to read through and understand.

I agree with lowering the length of the ID in development settings, and personally feel like 6-8 is a good length. 

Pulp implemented Django-GUID ([issue](https://pulp.plan.io/issues/4689)), but thought the `INFO`-logs generated a bit too much noise ([issue](https://pulp.plan.io/issues/7982)). I would recommend deciding on whether you want the `INFO` logs or not.


[Example PR from Pulp](https://github.com/pulp/pulpcore/pull/1057) or from the [docs](https://django-guid.readthedocs.io/en/latest/README_PYPI.html?highlight=warning#django-guid-logger-optional):  
```py
LOGGING = {
    ...
    'loggers': {
        'django_guid': {
            'handlers': ['<myhandler>'],
            'level': 'WARNING',
            'propagate': False,
        }
    }
}
````

**Comment by tchellomello** [comment]
@ryanpetrello well done!!  This feature will enhance a lot the troubleshooting efforts as the message will be logged in all nodes and we can see a reference from beginning to end a process. 

Having the ability to see a related task in the same tasks is neat! A simple `grep` and that is!

We will put together a lunch and learn so everyone in the team knows how to track it stuff from now on!

**Comment by ryanpetrello** [comment]
@JonasKs

I tried that out, but it yelled at me; I think the version we're pinned to maybe doesn't support that setting (and we're not ready to jump to Django 3.x yet):

```python
bash-4.4$ awx-python
Python 3.6.8 (default, Aug 24 2020, 17:57:11)
[GCC 8.3.1 20191121 (Red Hat 8.3.1-5)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import django_guid
>>> django_guid.__version__
'2.2.0'
```

```python
DJANGO_GUID['UUID_LENGTH'] = 8
```

image

JonasKs:
@ryanpetrello , you're right. [That wasn't included ](https://github.com/snok/django-guid/blob/2.2/docs/settings.rst) in the 2x releases. 

This is a setting I gladly will backport for you guys, if it's wanted. (PRs to the `2.2` branch also welcome 😄)

awx/main/dispatch/pool.py
```diff
@@ -445,6 +446,8 @@ def up(self):
             return super(AutoscalePool, self).up()
 
     def write(self, preferred_queue, body):
+        if 'guid' in body:
+            GuidMiddleware.set_guid(body['guid'])
```
ryanpetrello, line 450: When the (parent) dispatcher process receives a message over the bus, set it in this thread so logs in the parent process include the guid

JonasKs:
@tchellomello , wish I could be the fly on the wall for that! Sounds fun 🙌

awx/main/dispatch/publish.py
```diff
@@ -83,6 +84,9 @@ def apply_async(cls, args=None, kwargs=None, queue=None, uuid=None, **kw):
                     'kwargs': kwargs,
                     'task': cls.name
                 }
+                guid = GuidMiddleware.get_guid()
+                if guid:
+                    obj['guid'] = guid
```
ryanpetrello, line 89: When messages are *published* to the message bus, include the guid of the current thread (this is how the guid passes from e.g., an HTTP request or the task manager to the dispatcher)

awx/main/dispatch/worker/callback.py
```diff
@@ -152,6 +153,8 @@ def perform_work(self, body):
 
                 if body.get('event') == 'EOF':
                     try:
+                        if 'guid' in body:
+                            GuidMiddleware.set_guid(body['guid'])
```
ryanpetrello, line 157: When we get the EOF event, set the guid from the message so that messages about job event processing include the guid

awx/main/dispatch/worker/task.py
```diff
@@ -52,6 +54,8 @@ def run_callable(self, body):
         uuid = body.get('uuid', '<unknown>')
         args = body.get('args', [])
         kwargs = body.get('kwargs', {})
+        if 'guid' in body:
+            GuidMiddleware.set_guid(body.pop('guid'))
```
ryanpetrello, line 58: When a (child) dispatcher process begins working on a task, set the guid to the value from the message that came across the bus.

awx/main/tasks.py
```diff
@@ -1310,6 +1311,7 @@ def finished_callback(self, runner_obj):
         event_data = {
             'event': 'EOF',
             'final_counter': self.event_ct,
+            'guid': GuidMiddleware.get_guid(),
```
ryanpetrello: Pass the guid in the EOF message so the callback receiver can access its value

ryanpetrello:
@JonasKs @fosterseth here, I cheated:

```diff
--- a/awx/main/utils/filters.py
+++ b/awx/main/utils/filters.py
@@ -18,6 +18,7 @@ from django.conf import settings
 from django_guid.log_filters import CorrelationId
 from django_guid.middleware import GuidMiddleware

+from awx import MODE
 from awx.main.constants import LOGGER_BLOCKLIST
 from awx.main.utils.common import get_search_fields

@@ -373,5 +374,8 @@ class SmartFilter(object):
 class DefaultCorrelationId(CorrelationId):

     def filter(self, record):
-        record.guid = GuidMiddleware.get_guid() or '-'
+        guid = GuidMiddleware.get_guid() or '-'
+        if MODE == 'development':
+            guid = guid[:8]
+        record.guid = guid
         return True
```

image

JonasKs:
Hehe, that saves me some backport work, so I'm content with that implementation. 🥇

fosterseth (approved):

JonasKs:
As a sidenote, I would recommend providing the correlation ID to the end user together with the request error. (Maybe in [here](https://github.com/ansible/awx/blob/devel/awx/ui_next/src/util/useRequest.js#L15)?) Makes it easy to troubleshoot issues that happens for those who don't have log access 😃

matburt:
This is pretty great and I love the idea... a year or so ago we were throwing around some ideas on this with https://opentracing.io/ and I think this will be a huge boon to supportability.

blomquisg:
Nice

ryanpetrello:
@tchellomello @gamuniz in terms of a practical example, here's a Job Template I launched:

image

...and here it is all hooked up to an ELK setup w/ `level=DEBUG`:

image

...and some sample logs:

image
image

ryanpetrello:
This also works with job events (across related project syncs and actual playbook runs):

image

ELK:

image
image

PR #9333, opened 2021-02-16
set pull option default and add to AWXKit

rebeccahhh:
##### SUMMARY

set default option for pulling the container to none, this does not change the 3 available options for this field (which are still `Always`, `Never`, or `Missing`). This does set the default for the model and the DB to None.
This PR also adds the pull field into into the AWXKit api pages model

##### ISSUE TYPE

 - Bughancement Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 17.0.1
```


awx/main/models/execution_environments.py
```diff
@@ -44,7 +44,7 @@ class Meta:
     pull = models.CharField(
         max_length=1024,
         choices=PULL_CHOICES,
-        default='missing',
+        default=None,
```
shanemcd: I believe you also want to add 
```        
blank=True,
null=True,
```

See here for another example: https://github.com/ansible/awx/blob/devel/awx/main/models/ha.py#L198-L200

awxkit/awxkit/api/pages/execution_environments.py
```diff
@@ -33,12 +33,13 @@ def create_payload(self, name='', organization=Organization, **kwargs):
...
         return payload
 
-    def payload(self, name='', image=None, organization=None, credential=None, **kwargs):
+    def payload(self, name='', image=None, organization=None, credential=None, pull=None, **kwargs):
         payload = PseudoNamespace(
             name=name or "EE - {}".format(random_title()),
             image=image or random_title(10),
             organization=organization.id if organization else None,
             credential=credential.id if credential else None,
+            pull=pull if pull else None,
```
shanemcd: I think this can just be `pull=pull` since it _is_ `None` by default.

rebeccahhh: credential is none by default as well, I just was matching up, but I can change it

shanemcd: It's passing along the _id_ of the credential if there is one:

```
credential=credential.id if credential else None
```

Here you are just passing along `pull`, so this conditional is unnecessary.

rebeccahhh: done

shanemcd (approved):
👍

awx/main/models/execution_environments.py
```diff
@@ -44,7 +44,9 @@ class Meta:
     pull = models.CharField(
         max_length=1024,
```
jbradberry: I'd go for something much shorter than this.

awx/main/models/execution_environments.py
```diff
@@ -42,9 +42,11 @@ class Meta:
         on_delete=models.SET_NULL,
     )
     pull = models.CharField(
-        max_length=1024,
+        max_length=16,
         choices=PULL_CHOICES,
-        default='missing',
+        blank=True,
+        null=True,
```
jbradberry: We should also ditch `null=True`.

rebeccahhh: I'm going to remove that and interactive rebase, we don't need this many commits.

jbradberry (approved):

PR #9335, opened 2021-02-16
Detect EE Collections Name and Version Info

beeankha:
##### SUMMARY

Connected to the work being done in this Ansible Runner PR: https://github.com/ansible/ansible-runner/pull/629

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 17.0.1

```


awx/main/tasks.py
```diff
@@ -1179,11 +1179,17 @@ def final_run_hook(self, instance, status, private_data_dir, fact_modification_t
         instance.log_lifecycle("finalize_run")
         job_profiling_dir = os.path.join(private_data_dir, 'artifacts/playbook_profiling')
         awx_profiling_dir = '/var/log/tower/playbook_profiling/'
+        collections_info = os.path.join(private_data_dir, 'artifacts/')
```
shanemcd:
```suggestion
        collections_info = os.path.join(private_data_dir, 'artifacts', 'collections.json')
```

awx/main/tasks.py
```diff
@@ -1179,11 +1179,17 @@ def final_run_hook(self, instance, status, private_data_dir, fact_modification_t
...
+
         if not os.path.exists(awx_profiling_dir):
             os.mkdir(awx_profiling_dir)
         if os.path.isdir(job_profiling_dir):
             shutil.copytree(job_profiling_dir, os.path.join(awx_profiling_dir, str(instance.pk)))
-
+        if os.path.exists(collections_info):
+            with open(collections_info + 'collections.json') as ee_json_info:
```
shanemcd: With my suggestion above this would become: 
```suggestion
            with open(collections_info) as ee_json_info:
```

awx/main/tasks.py
shanemcd: The `/artifacts` directory will _always_ exist, but `collections.json` might not. So we should look for that.

beeankha: Yep, makes sense!

ryanpetrello (approved):
Nice!

awx/main/migrations/0128_unifiedjob_installed_collections.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('main', '0127_reset_pod_spec_override'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='unifiedjob',
+            name='installed_collections',
+            field=awx.main.fields.JSONBField(blank=True, default=dict, editable=False, help_text='The Collections names and versions installed in the execution environment.'),
```
ryanpetrello: @wenottingham @shanemcd @beeankha should we go ahead and update the analytics code to include this new data in what we ship to Analytics?

https://github.com/ansible/awx/blob/devel/awx/main/analytics/collectors.py#L316-L338

wenottingham: Yes, although cc @Ladas for updating the processor.

shanemcd: Yep, I agree that we should go ahead and do this here.

beeankha:
recheck

awx/main/analytics/collectors.py
```diff
@@ -334,7 +334,8 @@ def unified_jobs_table(since, full_path, until, **kwargs):
                                  main_unifiedjob.finished, 
                                  main_unifiedjob.elapsed, 
                                  main_unifiedjob.job_explanation, 
-                                 main_unifiedjob.instance_group_id
+                                 main_unifiedjob.instance_group_id,
```
ryanpetrello, line 337: @Ladas do we need to increment the version in the `@register()` decorator above?  What's the protocol for when we want to make/additions changes to this schema?

shanemcd (approved):

awx/main/tasks.py
```diff
@@ -1179,11 +1179,17 @@ def final_run_hook(self, instance, status, private_data_dir, fact_modification_t
...
+        collections_info = os.path.join(private_data_dir, 'artifacts/', 'collections.json')
+
         if not os.path.exists(awx_profiling_dir):
             os.mkdir(awx_profiling_dir)
         if os.path.isdir(job_profiling_dir):
             shutil.copytree(job_profiling_dir, os.path.join(awx_profiling_dir, str(instance.pk)))
-
+        if os.path.exists(collections_info):
+            with open(collections_info) as ee_json_info:
+                ee_collections_info = ee_json_info.read()
```
chrismeyersfsu: I _think_ we might need a `json.loads(ee_json_info.read())` here. In the demo you gave it "looked" like the data might be stored as a string in the database. The reason this works is that a top-level string is considered valid json in some circles and postgres is in that circle.

Below is an example of what it looks like when valid json is stored in a `jsonb` field vs. a string being stored.

https://asciinema.org/a/S451G5hKPoHMuywzm9QMxNtJM

awx/main/analytics/collectors.py
Ladas: @ryanpetrello  Yeah, the version increment looks good.

There isn't any official protocol right now, we're keeping track of all changes and test them on analytics side. :-) (and we're trying to create automated tests, that would test all supported Tower/Ansible versions)

Issue #9310, opened 2021-02-14
[ui_next] On Credential Edit Form, input values are not populated


unlikelyzero:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Previously, when editing a Cred Type Edit Form, the user would see their previous edits. This is no longer the case.

https://dashboard.cypress.io/projects/ui2mq7/runs/4411/test-results/6dbc5299-7a10-4b0d-b1ec-61756cadd307/video?utm_source=Dashboard&utm_medium=Share+URL&utm_campaign=Video

##### STEPS TO REPRODUCE
1. Create a GCE Credential with by uploading a valid json file
2. Save
2. Edit the same credential

##### EXPECTED RESULTS
Form elements to be populated with previous values including the $encrypted$ text

##### ACTUAL RESULTS
Form elements not pre populated with previous values

##### ADDITIONAL INFORMATION
https://dashboard.cypress.io/projects/ui2mq7/runs/4411/test-results/6dbc5299-7a10-4b0d-b1ec-61756cadd307/video?utm_source=Dashboard&utm_medium=Share+URL&utm_campaign=Video

PR #9336, opened 2021-02-16
Fix bug where credential inputs were not filled on edit

mabashian:
##### SUMMARY
link #9310 

Just needed to set the initial input values if they are present

Screen Shot 2021-02-16 at 6 51 39 PM

You don't need to upload the gce file to recreate this.  Just create a GCE cred and then try to edit.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/screens/Credential/CredentialEdit/CredentialEdit.jsx
+++ b/awx/ui_next/src/screens/Credential/CredentialEdit/CredentialEdit.jsx
@@ -116,7 +116,7 @@ function CredentialEdit({ credential, me }) {
         },
       ] = await Promise.all([
         CredentialTypesAPI.read({ page_size: 200 }),
-        CredentialsAPI.readInputSources(credId, { page_size: 200 }),
+        CredentialsAPI.readInputSources(credId),
       ]);
       const credTypes = data.results;
       if (data.next && data.next.includes('page=2')) {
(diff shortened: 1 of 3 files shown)


awx/ui_next/src/screens/Credential/CredentialEdit/CredentialEdit.jsx
```diff
@@ -116,7 +116,7 @@ function CredentialEdit({ credential, me }) {
         },
       ] = await Promise.all([
         CredentialTypesAPI.read({ page_size: 200 }),
-        CredentialsAPI.readInputSources(credId, { page_size: 200 }),
```
mabashian, line 119: This method only takes one param: https://github.com/ansible/awx/blob/devel/awx/ui_next/src/api/models/Credentials.js#L23

It defaults to a page size of 200

unlikelyzero: @mabashian for the record, there's a hidden performance problem here that I can't see.

unlikelyzero:
@mabashian is there a way to add a simple unit test?

unlikelyzero:
@mabashian i think you might want to add a unit test. Also: https://dashboard.cypress.io/projects/ui2mq7/runs/4460 is green

jakemcdermott (approved):

unlikelyzero (approved):

Issue #9259, opened 2021-02-08
[db_partitioning] adapt job runs to use new partitioned tables


jladdjr:
In order to create partitions, the job_event tables will need to have a new, denormalized column. Update job runs to ensure that when we're saving events, our saves / inserts take the new column into account.

jladdjr:
I _thought_ that what I was doing [here](https://github.com/jladdjr/awx/blob/1ed34380bb4c1e7354e0431a3371be789d04a08a/awx/main/migrations/0124_event_partitions.py#L43) would work, but after doing the raw sql update to add a new column, Django is left in the dark about the new column's existence.

![image](https://user-images.githubusercontent.com/4440360/107272193-31a04d00-6a02-11eb-9019-d851a452075b.png)


Chatted with @jbradberry who looked at looked at https://github.com/jladdjr/awx/blob/db_partition/awx/main/migrations/0113_v370_event_bigint.py and spotted how that handles things correctly. I knew the bigint migration made things more _performant_ but didn't realize it was handling the row addition differently. Studying that now to see how I can adapt that. May just adapt the whole approach since we want this to be performant anyways.

jladdjr:
yep, I need this:
https://github.com/jladdjr/awx/blob/db_partition/awx/main/migrations/0113_v370_event_bigint.py#L77-L82

jladdjr:
think we want something like this:

![image](https://user-images.githubusercontent.com/4440360/107285895-a8464600-6a14-11eb-8c9c-fcf99c4695d3.png)

Using FakeAdd instead of FakeAlter. Tweaked the model's args to what I think should be the appropriate values.

jladdjr:
This tells Django about the field's existence, but doesn't actually insert the job_created value when we're defining events. That will be the next thing to figure out if this works.

jladdjr:
reverting awx to test changes - rough notes:

![image](https://user-images.githubusercontent.com/4440360/107291143-8bae0c00-6a1c-11eb-914f-cdd27396a7ab.png)

---

.. as [script](https://gist.github.com/jladdjr/f0a842a7e24ac63cda87f461a8f13fae)

jladdjr:
Looks like after doing a partition I can still see events from an old job:

![image](https://user-images.githubusercontent.com/4440360/107292541-e21c4a00-6a1e-11eb-9091-25291bff6a53.png)

[...]

PR #9341, opened 2021-02-17
awx db partitioning

jladdjr:
closing this PR, opened [new PR](https://github.com/ansible/awx/pull/9839) using branch under ansible org to make it easier for @chrismeyersfsu and I to work on together


jladdjr:
cc @Ladas - for some reason the 'request review' feature isn't finding your name. cc-ing you for awareness

Ladas:
> I guess that's true - if someone did run a job immediately after the migration, the new migration that was created would overlap with the old partition, but any events created would just go in the new partition.
> 
> okay, cool, so this isn't needed

Seem like you figured it out, so just a note: the partitions can't overlap. Postgre should fail if you'd try to create overlapping partition. Same as it will fail if no partition is available for the timestamp.

awx/main/migrations/0124_event_partitions.py
```diff
@@ -0,0 +1,112 @@
...
+    #
+    # The initial partion will be a unique case. After
+    # the migration is completed, awx should create
+    # new partitions on an hourly basis, as needed.
+    # All events for a given job should be placed in
+    # a partition based on the job's _created time_.
+
+    # Only partitioning main_jobevent on first pass
+    #
+    #for tblname in (
```
ryanpetrello: @jladdjr before you continue onward with with problems related to analytics, job event cleanup, and large-scale migrations, it might be good to pause and update this PR with some changes that introduce partitions for *all* event types - looking at the layout of this code and the comments here, I think you're pretty close 😄, and having that is probably a first pass of what QE and perfscale folks need to actually stand up and test this from a functional/correctness perspective.

awx/main/utils/common.py
```diff
@@ -1079,3 +1082,47 @@ def truncate_stdout(stdout, size):
...
+    end_timestamp = str(end)
+
+    if not partition_label:
+        if minutely:
+            partition_label = start.strftime('%Y%m%d_%H%M')
+        else:
+            partition_label = start.strftime('%Y%m%d_%H')
+
+    with connection.cursor() as cursor:
+        # Only partitioning main_jobevent on first pass
```
ryanpetrello: Similar to the "make it work for all use cases" comment above, you could probably lose this loop, and just pass `tblname` _directly into_ this function so that the proper parent table is operated on depending on the context from which this function is called (either in the upgrade-time migration, or via the task manage as new jobs are scheduled).

awx/main/scheduler/task_manager.py
```diff
@@ -255,6 +256,11 @@ def start_task(self, task, rampart_group, dependent_tasks=None, instance=None):
                              rampart_group.name, task.log_format))
                 return
 
+        # Before task leaves pending state, ensure that job_event partitions exist
+        # TODO: Currently, only creating partition for jobs. Drop contiditional once
+        # there are partitions for all job event types
+        if task_actual['type'] == 'job':
+            create_partition()
```
ryanpetrello: I *do* think this is a reasonable place to put this function call, because it's within the scope of the global task manager advisory lock (so we don't have to worry about competing jobs creating a partition twice in close succession), and it's *right before* we anticipate we'll need the partition.

This doesn't need to happen now, but we should think about what it would mean if this function called *threw some exception*.  How would an uncaught exception behave in this portion of the task manager?  Would we see an error?  Would the task manager exit gracefully and "try again" in 30s?  Is that okay?

cc @chrismeyersfsu @Ladas 

This is a complicated and sensitive part of the overall system, so we should start pondering on it early 😄

ryanpetrello: It might even make more sense to put it at the top of the "post commit" function below where we actually *are positive* we're about to run an `.apply_async()` call:

https://github.com/ansible/awx/pull/9341/files#diff-a37220424979b0075fa6e25bf8c309d671f30641e64830c27cd046272c73a703R333

That way if the job fails `.pre_start()` (https://github.com/ansible/awx/pull/9341/files#diff-a37220424979b0075fa6e25bf8c309d671f30641e64830c27cd046272c73a703R268) or any of the other code directly below _this_ section of the code, and we *don't* actually end up passing a message to the dispatcher to perform work, we won't have needlessly created a partition.

Doing it that way will also ensure that we don't create partitions for `WorkflowJob`s, which don't have events.

awx/main/migrations/0124_event_partitions.py
```diff
@@ -0,0 +1,107 @@
...
+            cursor.execute(
+                f'ALTER TABLE ONLY {tblname} '
+                f'ADD CONSTRAINT {tblname}_pkey_new PRIMARY KEY (id, job_created);'
+            )
+
+            current_time = now()
+
+            # .. as well as initial partition containing all existing events
+            epoch = datetime(2000, 1, 1, 0, 0)
+            create_partition(epoch, current_time, 'old_events')
```
ryanpetrello: `create_partition(tblname, ...)` ?

awx/main/scheduler/task_manager.py
ryanpetrello: Sounds like we're also ready to call this function for _all_ event types, not just job events?

awx/main/tasks.py
```diff
@@ -1388,6 +1393,10 @@ def run(self, pk, **kwargs):
         if self.instance.spawned_by_workflow:
             self.parent_workflow_job_id = self.instance.get_workflow_job().id
 
+        # TODO: can we count on instance always having created?
```
ryanpetrello: Yes.  If we get to this point where we're creating an actual event, the related job already exists in the database, and definitely has a `created` value.

ryanpetrello: Also, this column is not nullable:

```sql
                                             Table "public.main_unifiedjob"
         Column          |           Type           | Collation | Nullable |                   Default
-------------------------+--------------------------+-----------+----------+---------------------------------------------
 id                      | integer                  |           | not null | nextval('main_unifiedjob_id_seq'::regclass)
 polymorphic_ctype_id    | integer                  |           |          |
 created                 | timestamp with time zone |           | not null |
```

awx/main/tasks.py
```diff
@@ -1246,6 +1247,10 @@ def event_handler(self, event_data):
                 event_data.pop('parent_uuid', None)
         if self.parent_workflow_job_id:
             event_data['workflow_job_id'] = self.parent_workflow_job_id
+        # Do we have to check if the field exists? if it doesn't
+        # how will be eventually store the event in the db?
+        if self.job_created:
```
ryanpetrello: Yep, I'd say we don't want this if statement, because if this value isn't set, the event won't be saved (because it's what the partitioning is based on).

If you get to this part of the codebase and `self.job_created` isn't set, there's a bug (so let's expose it).

awx/api/views/__init__.py
```diff
@@ -3826,7 +3826,7 @@ class JobJobEventsList(BaseJobEventsList):
     def get_queryset(self):
         job = self.get_parent_object()
         self.check_parent_access(job)
-        qs = job.job_events.select_related('host').order_by('start_line')
+        qs = job.job_events.select_related('host').filter(job_created=job.created).order_by('start_line')
```
ryanpetrello: We should copy this approach for all of the event list endpoints e.g.,

`/api/v2/project_updates/N/events/`
`/api/v2/inventory_updates/N/events/`
`/api/v2/system_job_updates/N/events/`
`/api/v2/ad_hoc_commands/N/events/`

awx/main/migrations/0124_event_partitions.py
jladdjr: yep, I got pulled off before I got everything updated / pushed / tested

awx/main/tasks.py
jladdjr: The db as the single source of truth on existence of data - that's cool / makes sense 👍 

Also, makes sense that the job _better_ have a created value. It was just that as I was going through keying off of things I wondered about taking for granted the existence of fields. Good to know ^

awx/main/tasks.py
jladdjr: makes sense - dropped comment / conditional

awx/api/views/__init__.py
jladdjr: If we wanted to put this method _everywhere_ seemed like maybe we wanted it in BaseJobEventsList
.. but not all of the other classes are children of that so 🤷 
.. but maybe there's another reason why we actually want this copy/pasted in the other classes?

jladdjr: oh, and it looks like in some cases the qs is formed a bit differently (`HostJobEventsList(BaseJobEventsList)`)

jladdjr: pushed https://github.com/ansible/awx/pull/9341/commits/1f746733cfeea99a98208a4fda4703e9109c7f3b - I'll need to test this out

awx/main/scheduler/task_manager.py
jladdjr: yep, got pulled off before I was able to make this change 👍

awx/main/models/events.py
```diff
@@ -426,13 +426,32 @@ def create_from_data(cls, **kwargs):
         except (KeyError, ValueError):
             kwargs.pop('created', None)
 
+        # same as above, for job_created
+        # TODO: if this approach, identical to above, works, can convert to for loop
```
ryanpetrello, line 428: I think doing this explicitly in this way is fine without a loop

awx/main/utils/common.py
```diff
@@ -1079,3 +1082,38 @@ def truncate_stdout(stdout, size):
...
 
     return stdout + u'\u001b[0m' * (set_count - reset_count)
+
+def create_partition(tblname, start=None, end=None, partition_label=None, minutely=False):
+    """Creates new partition table for events.
+    - start defaults to beginning of current hour
+    - end defaults to end of current hour
+    - partition_label defaults to YYYYMMDD_HH
+
+    - minutely will create partitions that span _a single minute_ for testing purposes
```
ryanpetrello, line 1056: This is neat and helpful for testing and prototyping; let's make a mental note to remove the per-minute support prior to merge, however (having this accidentally called at some point in the future would be bad, because postgres isn't going to like having this many partitions over time.  Also, we're likely to have other logic (event cleanup) that assumes partitions are "per hour").

awx/main/migrations/0124_event_partitions.py
```diff
@@ -0,0 +1,133 @@
...
+            cursor.execute(
+                f'ALTER TABLE {tblname} RENAME TO {tblname}_old'
+            )
+
+            # drop primary key constraint; in a partioned table
+            # constraints must include the partition key itself
+            # TODO: do more generic search for pkey constraints
+            # instead of hardcoding this one that applies to main_jobevent
+            cursor.execute(
+                f'ALTER TABLE {tblname}_old DROP CONSTRAINT {tblname}_pkey1'
```
ryanpetrello: Reading pg docs, this is a little unclear to me.  Does the *primary key* have to include the partition column, or just **any** indexed constraint?  There's almost certainly going to be situations where we want to search for a specific event based on ID only e.g., `/api/v2/jobs/N/job_events/Y/` and it's going to be super slow if this column isn't the key.

e.g.,

Does this still work if we **keep** `id` as the primary key, and *instead* do:

```sql
ALTER TABLE ONLY {tblname} ADD CONSTRAINT {tblname}_unique_id_job_created UNIQUE (id, job_created);
```

If this works w/ partitions, I'd prefer it.

jladdjr: > Does this still work if we keep id as the primary key

No, this is not an option; postgres refuses to let you claim a field as being unique _unless_ you use a composite primary key and include the partition field (job_created)

jladdjr: okay, found my comment where I posted my findings on this:
https://github.com/ansible/awx/issues/9259#issuecomment-776325677

jladdjr: What I _think_ pg will try to do using the tables exactly the way they are today is:
* run a parallel search of each of the partitions associated with the parent job event table (literally spinning up different workers to handle each table partition)
* for each partition, use _that table's_ index for id / job_created

I tried running a query with `explain analyze` just now, but I hardly have any events in my db. It's showing a sequential scan, but my sense is that's probably because the postgres planner knew we hardly had any events and didn't bother with the index.

cc @Ladas for fact-checking because this is a pretty wild guess on my part

jladdjr:
rebased.. this PR has now joined 2021

awx/main/migrations/0124_event_partitions.py
```diff
@@ -0,0 +1,117 @@
...
+
+            # create parent table
+            cursor.execute(
+                f'CREATE TABLE {tblname} '
+                f'(LIKE _unpartitioned_{tblname} INCLUDING ALL, job_created TIMESTAMP WITH TIME ZONE NOT NULL) '
+                f'PARTITION BY RANGE(job_created);'
+            )
+
+            # let's go ahead and add and subtract a few indexes while we're here
+            cursor.execute(f'CREATE INDEX {tblname}_modified_idx ON {tblname} (modified);')
```
ryanpetrello: cc @Ladas

awx/main/utils/common.py
```diff
@@ -161,6 +165,14 @@ def memoize_delete(function_name):
     return cache.delete(function_name)
 
 
+@memoize(ttl=3600 * 24)  # in practice, we only need this to load once at process startup time
+def get_event_partition_epoch():
+    from django.db.migrations.recorder import MigrationRecorder
+    return MigrationRecorder.Migration.objects.filter(
+        app='main', name='0124_event_partitions'
```
ryanpetrello: @jladdjr if we end up reordering this migration (I'll bet exec envs beat us), we'll have to correct this label.

awx/main/migrations/0124_event_partitions.py
Ladas: @ryanpetrello I'm fairly sure this will work for analytics. We'll probably need to limit the query with the partition key (created), so e.g. we scan max 24 partitions/h back. But I think the index should be still used in this case

awx/main/migrations/0124_event_partitions.py
Ladas: @ryanpetrello @jladdjr so for query like `/api/v2/jobs/N/job_events/Y/` you will still need to use the partition key `job_created`, otherwise PG will scan all partitions and that will always be slow.

So the rule of thumb is using the partitioning key in all queries, when using partitions. (Well there are e.g. cases where you'd want to scan all the data in parallel, but postgres does pretty good parallel seq scan even without partitions.)

For this specific usecase PG requires unique constraint/primary key to include the partitioning key. And it should use the index on id with `(id, job_created)`

-------

@jladdjr yeah, the index/seq scan is dependent on what cost will PG compute. Which in simplified terms takes into account what percentage of the table PG thinks you'll be looking at, based on statistics it keeps. (very simply said, index should be used when querying ~5% of the table). 

Not sure if you're using default costs for PG. The most important is usually `random_page_cost`, which is by default 4.0, which is good for magnetic disks (index access is expensive). I set it 1.1 for ssd. Do you recommend PG config to users in docs? Or this is up to users?

awx/main/migrations/0124_event_partitions.py
ryanpetrello: So one question @Ladas is how we're *actually* going to make use of this in a forward-facing way that's *also* backwards-compatible given https://github.com/ansible/awx/pull/9358/files#r584917954

jladdjr:
rebased and ran through basic sanity check, confirmed things are still looking good

awx/main/tasks.py
```diff
@@ -740,22 +746,16 @@ def update_host_smart_inventory_memberships():
...
-    with advisory_lock(f'bigint_migration_{tblname}', wait=False) as acquired:
+    with advisory_lock(f'partition_migration_{tblname}', wait=False) as acquired:
         if acquired is False:
             return
         chunk = settings.JOB_EVENT_MIGRATION_CHUNK_SIZE
 
         def _remaining():
             try:
-                cursor.execute(f'SELECT MAX(id) FROM _old_{tblname};')
+                cursor.execute(f'SELECT MAX(id) FROM _unpartitioned_{tblname};')
```
kdelee: what if we just look if there is at least 1 record
```
select id from _unpartitioned_main_jobevent limit 1;
```

Ladas: @jladdjr  each of the MAX(id), INSERT and DELETE queries is causing sequential scan now, so I think the index on id is needed, if you want to use the current way 

(although there should be a way to move the data with DB procedure without index, in one sequential pass of the table, using a DB cursor :-), but it's better to stay with working algorithm as @ryanpetrello  says :-) )

------

I'd say `copy original schema to TMP table` -> `change TMP table schema(remove the primary key)` -> `create partition parent from TMP table schema`  is probably the simplest (one line change, or 2 lines, deleting also the TMP table :slightly_smiling_face: )

So keeping the primary key on the original events table and reusing the algorithm sounds the simplest :-)

awx/main/models/events.py
```diff
@@ -426,13 +426,28 @@ def create_from_data(cls, **kwargs):
...
         host_map = kwargs.pop('host_map', {})
 
         sanitize_event_keys(kwargs, cls.VALID_KEYS)
         workflow_job_id = kwargs.pop('workflow_job_id', None)
         event = cls(**kwargs)
         if workflow_job_id:
             setattr(event, 'workflow_job_id', workflow_job_id)
+        # shouldn't job_created _always_ be present?
+        # if it's not, how could we save the event to the db?
+        job_created = kwargs.pop('job_created', None)
```
Ladas, line 446: I'd say that inserting None will fail, since it won't match any partition

awx/main/tasks.py
jladdjr: > I'd say copy original schema to TMP table -> change TMP table schema(remove the primary key) -> create partition parent from TMP table schema is probably the simplest (one line change, or 2 lines, deleting also the TMP table 🙂 )

makes sense - I'm writing this up now and will kick the tires a bit. if things look good we can try to patch things in the scale testing context

jladdjr: https://github.com/jladdjr/awx/commit/91b08176884690ef6f3ec69a3cfada884f597571 creates a tmp table that we modify / use as a basis for the partitioned table's schema

jladdjr: I made a small tweak to that ^, but after that was able to confirm that the migration went through and that we still had an index on the old jobevents table

jladdjr: Going to try to run through a migration w/ a larger set of data and see if we hit any performance issues at a bigger scale (before we go back and try this at true scale)

awx/main/models/events.py
ryanpetrello: Probably better to just do `kwargs['job_created']` here - if we were ever missing this key at this point in the code, it's a bug and we want to know.

jladdjr:
rebased.. and bumped migration

jladdjr:
rebasing again - another migration dropped in devel

jladdjr:
builds are now passing..

murakobardenas (approved):

jladdjr:
closing this PR, opened [new PR](https://github.com/ansible/awx/pull/9839) using branch under ansible org to make it easier for @chrismeyersfsu and I to work on together

Issue #5546, opened 2019-12-19
Handle Deletion Warnings


marshmalien:
### Summary
Original AWX Feature Issue: https://github.com/ansible/awx/issues/275

In the delete modal, we show a warning if the resource is currently used by other resources.  We retrieve the "count" by firing off N number of requests per dependency-- i.e., deleting one credential will fire 4 GET requests to retrieve project, job template, inventory, and inventory source counts. 

If we apply the same UI/UX implementation to the multi-delete modal, we could end up firing >200 requests to the api just to **_display_** the delete modal. Before coming to a solution, this issue will require some discussion and UX assistance @trahman73 

**Resources and dependencies**:
- Credentials
    - Project
    - Job Template
    - Inventory
    - Inventory Source
- Credential Type
    - Credential
- Inventory
    - Job Template
    - Workflow Job Template
- Inventory Script
    - Inventory Source
- Inventory Source
    - Workflow Job Template Node
- Job Template
    - Workflow Job Template Node
- Organizations
    - Credential
- Project
    - Job Template
    - Workflow Job Template Node
    - Inventory Source
- Unified Job Template
    - Workflow Job Template Node

#### Current mockup:
_AWX-PF Mockup_: 
![image (3)](https://user-images.githubusercontent.com/15881645/71124195-54462580-21b2-11ea-9d5b-0707561f6a04.png)

#### Current AWX behavior for reference:
_Credential Delete Modal_
<img width="859" alt="Screen Shot 2019-12-18 at 4 25 23 PM" src="https://user-images.githubusercontent.com/15881645/71188469-03d0d580-224f-11ea-83a9-a116b911656f.png">

_Template Delete Modal_
<img width="721" alt="Screen Shot 2019-12-18 at 11 14 49 AM" src="https://user-images.githubusercontent.com/15881645/71188502-1519e200-224f-11ea-93bb-961647bd9959.png">

dsesami:
spoke with @marshmalien, this isn't implemented yet. moving to needs_devel.

trahman73:
Updated mockup with latest PF4 styling:
![image](https://user-images.githubusercontent.com/13351591/89209496-667fa300-d58c-11ea-8623-e68e6021afb4.png)

unlikelyzero:
Consider this case: https://github.com/ansible/awx/issues/8795

unlikelyzero:
coverage added

PR #9345, opened 2021-02-17
Adds delete warnings on lists

AlexSCorey:
##### SUMMARY
This addresses #5546.  
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
Screen Shot 2021-02-17 at 12 19 18 PM
Screen Shot 2021-02-17 at 3 14 35 PM


awx/ui_next/src/util/useDeleteDetails.js
```diff
@@ -0,0 +1,23 @@
...
+        Object.values(requests).map(request => request())
+      );
+      Object.keys(requests).forEach((key, index) => {
+        results[key] = calls[index].data.count;
+      });
+    } catch (err) {
+      error = err;
+    }
+  }
+  const hasCount = Object.values(results).some(value => value > 0);
```
keithjgrant: I think `count` and `hasCount` variables are a bit confusing because they refer to different counts. Maybe rename `count` to `numItemsToDelete` or something a little more specific

awx/ui_next/src/util/useDeleteDetails.js
```diff
@@ -0,0 +1,23 @@
+export default async function getDeleteDetails(requests, count) {
+  const results = {};
+  let error = null;
+
+  if (count === 1) {
```
keithjgrant: what if count > 1?

awx/ui_next/src/components/TemplateList/TemplateList.jsx
```diff
@@ -168,6 +169,13 @@ function TemplateList({ defaultParams, i18n }) {
     <AddDropDownButton key="add" dropdownItems={addDropDownButton} />
   );
 
+  const deleteDetailsRequests = {
+    [i18n._(t`Workflow Job Template Node`)]: async () =>
```
keithjgrant: it feels a little weird using translated strings as object keys here. Maybe consider restructuring it a bit to something like:

```
{
  template_node: {
    fetch: async () => {...},
    label: i18n._(t`Workflow Job Template Node`),
  }
}
```

(and maybe even change it into an array of those objects rather than an object of nested objects

awx/ui_next/src/util/useDeleteDetails.js
AlexSCorey: @keithjgrant  if the count is >1 we don't want this function to do anything  But, it might make more sense to do that check inside the component that calls this helper function.

If count is > 1 we just show this:

Screen Shot 2021-02-17 at 12 18 39 PM

keithjgrant: In that case, I'd re-structure to: (cuts down indentation/visual complexity)

```
if (count > 1) {
  return { error };
}
```

awx/ui_next/src/components/DeleteButton/DeleteButton.jsx
```diff
@@ -12,48 +23,98 @@ function DeleteButton({
...
-          >
-            {i18n._(t`Cancel`)}
-          </Button>,
-        ]}
-      >
-        {i18n._(t`Are you sure you want to delete:`)}
-        <br />
-        <strong>{name}</strong>
-      </AlertModal>
+      {!deleteMessageError && (
```
keithjgrant: this check seems unnecessary since you've already returned on linke 47 if `deleteErrorMessage` is set

awx/ui_next/src/components/DeleteButton/DeleteButton.jsx
```diff
@@ -12,48 +23,98 @@ function DeleteButton({
...
+          <strong>{name}</strong>
+          {Object.values(deleteDetails).length > 0 && (
+            <WarningMessage
+              variant="warning"
+              isInline
+              title={
+                <div>
+                  <div aria-label={deleteMessage}>{deleteMessage}</div>
+                  <br />
+                  {Object.entries(deleteDetails).map(([key, value]) => (
```
keithjgrant: 👍🏻

awx/ui_next/src/components/AlertModal/AlertModal.jsx
```diff
@@ -77,7 +77,7 @@ function AlertModal({
       aria-label={label || i18n._(t`Alert modal`)}
       aria-labelledby="alert-modal-header-label"
       isOpen={Boolean(isOpen)}
-      variant="small"
+      variant="medium"
```
nixocio: This change will affect the whole app, perhaps we could pass this one as a prop.

awx/ui_next/src/components/TemplateList/TemplateList.jsx
```diff
@@ -236,6 +242,11 @@ function TemplateList({ defaultParams, i18n }) {
                   onDelete={handleTemplateDelete}
                   itemsToDelete={selected}
                   pluralizedItemName={i18n._(t`Templates`)}
+                  deleteDetailsRequests={deleteDetailsRequests}
+                  deleteMessage={i18n._(
+                    '{numItemsToDelete, plural, one {This template is currently being used by some workflow nodes. Are you sure you want to delete it?} other {Deleting these templates could impact some workflow nodes that rely on them. Are you sure you want to delete anyway?}}',
```
nixocio, line 247: cool usage of plural 👍

awx/ui_next/src/screens/CredentialType/CredentialTypeDetails/CredentialTypeDetails.jsx
```diff
@@ -82,6 +88,10 @@ function CredentialTypeDetails({ credentialType, i18n }) {
               modalTitle={i18n._(t`Delete credential type`)}
               onConfirm={deleteCredentialType}
               isDisabled={isLoading}
+              deleteDetailsRequests={deleteDetailsRequests}
+              deleteMessage={i18n._(
+                t`This credential type is currently being used by some credentials. Are you sure you want to delete it?`
```
nixocio: We are showing an alert that a credential type is being used by some credentials. But the API does not allow this action to happen. Perhaps we can handle this one in a different way from the others, or even in another issue. Since the user will not be able to delete the credential type.

image


image

awx/ui_next/src/components/PaginatedDataList/ToolbarDeleteButton.jsx
```diff
@@ -66,20 +80,44 @@ function ToolbarDeleteButton({
...
+
+  const [deleteMessageError, setDeleteMessageError] = useState();
   const handleDelete = () => {
     onDelete();
     toggleModal();
   };
 
-  const toggleModal = () => {
-    setIsModalOpen(!isModalOpen);
+  const toggleModal = async isOpen => {
```
nixocio, line 112: ~We can wrap this function on `useCallback` to avoid to be recreated every re-render.~

nixocio (commented):
We are rendering an empty alert on the schedules page.

image

Also on 

image

awx/ui_next/src/screens/CredentialType/CredentialTypeDetails/CredentialTypeDetails.jsx
AlexSCorey: @nixocio good point.  I think the best thing to do here is, if the credential type is associated with an existing credential

AlexSCorey:
recheck

unlikelyzero:
## Concerns:
1. Primary concern is that there is a resource which has a delete warning... which doesnt.
2. Not an exhaustive list. May need more investigation.
3. What are the EE dependencies?

## What's been tested so far:
**bug in bold**
** follow up with 3.8 to see how we did it

#### Credentials List
- User-owned credential. If used by JT, What happens when you delete the user? Answer: No additional warning needed
- Credentials used by inventory sources (i.e VMware) should display the notification warning on delete
- Cred types used by credentials should display the notification warning on delete

### Job templates
- Job templates used by workflows should display the notification warning on delete
- Workflows used by workflow should display the notification warning on delete
- Inventories
- Projects

### Projects

### Teams
- Organizations

### Notification templates
- Organizations

### Schedules
- has not dependency

### Instance groups

- [ ] An instance group can be used by organizations, Inventories, JTs... would be good to warn the user before deleting 

### Applications

- [ ] An app can be used to create a new token for a user, if the app was removed the to token also is removed, would be good to warn

#### User List

### EE
- Job templates/WFJT
- Organizations
- Inventories
- Inventory sync

#### Organization
-  Deleting Many Orgs should warn have general warning
- [ ] -  Deleting an Org with an attached team should warn that the Team will be deleted
-  Deleting an Org with an attached credential should warn that the Credential will be deleted
-  Deleting an Org with an attached user should NOT warn that the user will be deleted. The user becomes orphaned
-  Deleting an Org with multiple AH Credentials will only warn of the non-Galaxy API credentials
-  Deleting an Org with an attached Notification Template should warn that the Notification Template will be deleted
- [ ] - Organization used by projects or inventories should display the warning on delete and also have correct message:
image

#### External Secret Credentials
-  **Deleting an external Secret Credentials which is used by other credentials should have a general warning about dependencies

tiagodread:
10 e2e tests broken in this branch

akus062381:
@AlexSCorey we are going to update the tests tomorrow and get this into a mergeable state, we should be able to merge tomorrow depending developer feedback.

Follow up with @jakemcdermott : the 3.8 parity for externals credentials should not block merge

unlikelyzero:
regate

jakemcdermott:
@AlexSCorey @keithjgrant  @nixocio 


Is this one ready for another look?

nixocio (approved):

PR #9346, opened 2021-02-17
allow org members to add EEs to an org and superusers to change EEs in an org

rebeccahhh:
##### SUMMARY

allow org members to add EEs to an org and superusers to change EEs in an org and address #9359 

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 17.0.1
```


awx/main/access.py
```diff
@@ -1332,11 +1332,13 @@ def filtered_queryset(self):
     @check_superuser
     def can_add(self, data):
         if not data:  # So the browseable API will work
-            return Organization.accessible_objects(self.user, 'execution_environment_admin_role').exists()
+            return Organization.accessible_objects(self.user, 'member_role').exists()
         return self.check_related('organization', Organization, data)
 
     @check_superuser
     def can_change(self, obj, data):
+        if self.user.is_superuser:
```
jbradberry, line 1342: `@check_superuser` already does this

awx/main/access.py
```diff
@@ -1332,11 +1332,13 @@ def filtered_queryset(self):
     @check_superuser
```
jbradberry, line 1332: In the `filtered_queryset` method right above this point, there's a query using `'execution_environment_admin_role'` that needs to change to use `'member_role'`.  This is the query that gets used to see if a user can view something.

awx/main/access.py
rebeccahhh: 🤦🏻 you right

awx/main/access.py
```diff
@@ -1332,11 +1332,13 @@ def filtered_queryset(self):
...
+            return Organization.accessible_objects(self.user, 'member_role').exists()
         return self.check_related('organization', Organization, data)
 
     @check_superuser
     def can_change(self, obj, data):
+        if self.user.is_superuser:
+            return True
         if obj and obj.organization_id is None:
             raise PermissionDenied
         if self.user not in obj.organization.execution_environment_admin_role:
```
jbradberry, line 1346: I suspect that this needs to check that the user is not in `obj.organization.admin_role` as well.

awx/main/access.py
```diff
@@ -1325,14 +1325,14 @@ class ExecutionEnvironmentAccess(BaseAccess):
...
-            Q(organization__in=Organization.accessible_pk_qs(self.user, 'execution_environment_admin_role')) |
+            Q(organization__in=Organization.accessible_pk_qs(self.user, 'member_role')) |
             Q(organization__isnull=True)
         ).distinct()
 
     @check_superuser
     def can_add(self, data):
         if not data:  # So the browseable API will work
-            return Organization.accessible_objects(self.user, 'execution_environment_admin_role').exists()
+            return Organization.accessible_objects(self.user, 'member_role').exists()
```
jbradberry: I _think_ that this change isn't correct, but am willing to be shown wrong.

kdelee: I agree this looks like any member can create one

rebeccahhh: it sure does, and I tested it just to verify. I'll pull it out

awx/main/access.py
```diff
@@ -1325,7 +1325,7 @@ class ExecutionEnvironmentAccess(BaseAccess):
 
     def filtered_queryset(self):
         return ExecutionEnvironment.objects.filter(
-            Q(organization__in=Organization.accessible_pk_qs(self.user, 'execution_environment_admin_role')) |
+            Q(organization__in=Organization.accessible_pk_qs(self.user, 'member_role')) |
```
jbradberry: Probably should be `'read_role'` now that I've looked more closely again at the `Organization` model roles.  Also, I think we need to add `'execution_environment_admin_role'` to `Organization.read_role`'s parent roles.

awx/main/access.py
```diff
@@ -1339,7 +1339,7 @@ def can_add(self, data):
     def can_change(self, obj, data):
         if obj and obj.organization_id is None:
             raise PermissionDenied
-        if self.user not in obj.organization.execution_environment_admin_role:
+        if self.user not in obj.organization.execution_environment_admin_role and self.user not in obj.organization.admin_role:
```
jbradberry: It turns out that this change isn't necessary:

    In [15]: pprint([(u, u in o.execution_environment_admin_role) for u in User.obje
        ...: cts.all()])
    [(<User: admin>, True),
     (<User: auditor>, False),
     (<User: org-admin>, True),
     (<User: org-ee-admin>, True),
     (<User: org-member>, False),
     (<User: ordinary>, False)]

awx/main/access.py
```diff
@@ -1325,7 +1325,7 @@ class ExecutionEnvironmentAccess(BaseAccess):
 
     def filtered_queryset(self):
         return ExecutionEnvironment.objects.filter(
-            Q(organization__in=Organization.accessible_pk_qs(self.user, 'execution_environment_admin_role')) |
+            Q(organization__in=Organization.accessible_pk_qs(self.user, 'member_role')) |
             Q(organization__isnull=True)
         ).distinct()
```
jbradberry, line 1331: Leaving this here, since `can_add` is outside of the commentable lines of code -- it turns out that `check_related` takes an optional `mandatory` flag that I think does what we want wrt Org Admins / EE Admins not being able to create global EEs.

awx/main/access.py
rebeccahhh: `read_role` replaced member role and added to the orgs read_role list

awx/main/access.py
rebeccahhh: alright, removed

yagomarques (approved):

awx/api/serializers.py
```diff
@@ -1362,6 +1362,7 @@ def validate(self, attrs):
 
 class ExecutionEnvironmentSerializer(BaseSerializer):
     show_capabilities = ['edit', 'delete']
+    managed_by_tower = serializers.ReadOnlyField()
```
shanemcd, line 1365: 👍

shanemcd (approved):

rebeccahhh:
@shanemcd @jbradberry I rebased with the new changes in the main EE branch, I think we're good to `mergit`

awx/main/access.py
```diff
@@ -1308,6 +1308,54 @@ def can_unattach(self, obj, sub_obj, relationship, *args, **kwargs):
...
+    @check_superuser
+    def can_add(self, data):
+        if not data:  # So the browseable API will work
+            return Organization.accessible_objects(self.user, 'execution_environment_admin_role').exists()
+        return self.check_related('organization', Organization, data, mandatory=True,
+                                  role_field='execution_environment_admin_role')
+
+    @check_superuser
+    def can_change(self, obj, data):
+        if obj.managed_by_tower is True:
```
jbradberry: This should probably be changed to `if obj.managed_by_tower:`.  Let's not rely on Python inlining all instances of `True` to the same object.

awx/main/access.py
```diff
@@ -1308,6 +1308,54 @@ def can_unattach(self, obj, sub_obj, relationship, *args, **kwargs):
...
+    prefetch_related = ('organization__admin_role', 'organization__execution_environment_admin_role')
+
+    def filtered_queryset(self):
+        return ExecutionEnvironment.objects.filter(
+            Q(organization__in=Organization.accessible_pk_qs(self.user, 'read_role')) |
+            Q(organization__isnull=True)
+        ).distinct()
+
+    @check_superuser
+    def can_add(self, data):
```
jbradberry, line 1333: We should probably do a similar `managed_by_tower` check here as well.

awx/main/access.py
rebeccahhh: got it :)

awx/main/access.py
```diff
@@ -1325,31 +1325,35 @@ class ExecutionEnvironmentAccess(BaseAccess):
...
+            raise PermissionDenied
+        if self.user.is_superuser:
+            return True
+        return self.check_related('organization', Organization, data, mandatory=True,
+                                  role_field='execution_environment_admin_role')
 
-    @check_superuser
     def can_change(self, obj, data):
+        if obj.managed_by_tower:
+            raise PermissionDenied
```
jbradberry, line 1341: We should also check, I think, whether `'managed_by_tower'` is being attempted to be set in the data dict.  Maybe.

jbradberry: never mind, we set it to a read-only field :facepalm:

shanemcd (approved):

rebeccahhh:
regate

PR #9348, opened 2021-02-17
don't allow X-API-Request-Id to be specified in the request headers

ryanpetrello:
see: https://github.com/ansible/awx/pull/9332

### DIFF
--- a/awx/main/middleware.py
+++ b/awx/main/middleware.py
@@ -15,6 +15,7 @@ from django.apps import apps
 from django.utils.deprecation import MiddlewareMixin
 from django.utils.translation import ugettext_lazy as _
 from django.urls import reverse, resolve
+from django_guid.middleware import GuidMiddleware
 
 from awx.main.utils.named_url_graph import generate_graph, GraphNode
 from awx.conf import fields, register
@@ -23,6 +24,14 @@ from awx.main.utils.profiling import AWXProfiler
 
 logger = logging.getLogger('awx.main.middleware')
 perf_logger = logging.getLogger('awx.analytics.performance')
+
+
+class AWXGuidMiddleware(GuidMiddleware):
+
+    def _get_id_from_header(self, request):
+        # never accept the correlation ID from the HTTP response header
+        request.correlation_id = self._generate_guid()
+        return request.correlation_id
 
 
 class TimingMiddleware(threading.local, MiddlewareMixin):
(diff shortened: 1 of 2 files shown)


awx/main/middleware.py
```diff
@@ -25,6 +26,14 @@
 perf_logger = logging.getLogger('awx.analytics.performance')
 
 
+class AWXGuidMiddleware(GuidMiddleware):
+
+    def _get_id_from_header(self, request):
+        # never accept the correlation ID from the HTTP response header
+        request.correlation_id = self._generate_guid()
```
ryanpetrello, line 33: `django-guid` lets you specify a GUID via a header in the HTTP request; we don't want this behavior

ingvaldlorentzen: Just wondering why you don't want this behaviour?  
If triggering a job from another API or service that uses a similar correlation ID system, this is a great way to track a request/job/call across multiple systems.

shanemcd:
recheck

JonasKs:
@ingvaldlorentzen already asked for a reason behind this, and I'm curious too, since I disagree with the change.

Let's say I have a correlation ID in my backend(which we do), and my backend triggers a task or does API calls to AWX. I want the logs from AWX to have the same correlation ID as my backend, so when a user reports an error to me, I only need one correlation ID to find _all_ logs, through all systems. Being unable to set a correlation ID to AWX would break the chain.

ryanpetrello:
@ingvaldlorentzen @JonasKs yea, this is a valid point, and I may actually back away from this PR.  On one hand, I see how this could be useful from an integration perspective with third-party systems, but on the other hand, I could see some users attaching some stigma to "a non-privileged user can pass in a request header, and it ends up showing up in all sort of logs".  That said, it looks like there's validation that ensures the provided string is UUID-like, so perhaps this isn't really that much of an issue:

https://github.com/snok/django-guid/blob/2.2/django_guid/middleware.py#L136
https://github.com/snok/django-guid/blob/2.2/django_guid/middleware.py#L112

```python
>>> uuid.UUID('not a uuid 🐉', version=4)
  File "/usr/lib64/python3.8/uuid.py", line 169, in __init__
    raise ValueError('badly formed hexadecimal UUID string')
ValueError: badly formed hexadecimal UUID string
```

ryanpetrello:
Yep, I think this is a non-issue.  Thanks for the feedback.

PR #9358, opened 2021-02-18
Gather job event analytics by last pk instead of created datetime

jbradberry:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
awx: 17.0.1
```

### DIFF
--- a/awx/main/analytics/collectors.py
+++ b/awx/main/analytics/collectors.py
@@ -1,11 +1,12 @@
 import io
+import json
 import os
 import os.path
 import platform
 import distro
 
 from django.db import connection
-from django.db.models import Count, Max
+from django.db.models import Count, Max, Min
 from django.conf import settings
 from django.utils.timezone import now, timedelta
 from django.utils.translation import ugettext_lazy as _
@@ -295,11 +296,13 @@ def _copy_table(table, query, path):
 def events_slicing(key, since, until):
     from awx.conf.models import Setting
 
+    pk_values = models.JobEvent.objects.filter(created__gte=since, created__lte=until).aggregate(Min('pk'), Max('pk'))
+
     step = 100000
     last_entries = Setting.objects.filter(key='AUTOMATION_ANALYTICS_LAST_ENTRIES').first()
-    last_entries = last_entries.value if last_entries is not None else {}
-    previous_pk = last_entries.get(key) or 0
-    final_pk = models.JobEvent.objects.filter(created_lte=until).aggregate(Max('pk'))['pk__max']
+    last_entries = json.loads((last_entries.value if last_entries is not None else '') or '{}')
+    previous_pk = last_entries.get(key) or pk_values['pk__min'] or 0
+    final_pk = pk_values['pk__max']
 
     for start in range(previous_pk, final_pk + 1, step):
         yield (start, min(start + step, final_pk))
--- a/awx/main/analytics/collectors.py
+++ b/awx/main/analytics/collectors.py
@@ -8,16 +8,16 @@ import distro
 from django.db import connection
 from django.db.models import Count, Max, Min
 from django.conf import settings
+from django.contrib.sessions.models import Session
 from django.utils.timezone import now, timedelta
 from django.utils.translation import ugettext_lazy as _
 
 from awx.conf.license import get_license
 from awx.main.utils import get_awx_version, get_custom_venv_choices, camelcase_to_underscore, datetime_hook
 from awx.main import models
-from django.contrib.sessions.models import Session
 from awx.main.analytics import register
 
-'''
+"""
 This module is used to define metrics collected by awx.main.analytics.gather()
 Each function is decorated with a key name, and should return a data
 structure that can be serialized to JSON
@@ -31,7 +31,7 @@ All functions - when called - will be passed a datetime.datetime object,
 `since`, which represents the last time analytics were gathered (some metrics
 functions - like those that return metadata about playbook runs, may return
 data _since_ the last report date - i.e., new data in the last 24 hours)
-'''
+"""
 
 
 def trivial_slicing(key, since, until):
@@ -74,13 +74,13 @@ def events_slicing(key, since, until):
     horizon = until - timedelta(weeks=4)
 
     lower = since or last_gather or horizon
-    if last_entries.get(key):
+    if not since and last_entries.get(key):
         lower = horizon
     pk_values = models.JobEvent.objects.filter(created__gte=lower, created__lte=until).aggregate(Min('pk'), Max('pk'))
 
-    previous_pk = pk_values['pk__min'] or 0
-    if last_entries.get(key):
-        previous_pk = max(last_entries[key] + 1, previous_pk)
+    previous_pk = pk_values['pk__min'] - 1 if pk_values['pk__min'] is not None else 0
+    if not since and last_entries.get(key):
+        previous_pk = max(last_entries[key], previous_pk)
     final_pk = pk_values['pk__max'] or 0
 
     step = 100000
--- a/awx/main/analytics/collectors.py
+++ b/awx/main/analytics/collectors.py
@@ -12,7 +12,7 @@ from django.utils.timezone import now, timedelta
 from django.utils.translation import ugettext_lazy as _
 
 from awx.conf.license import get_license
-from awx.main.utils import get_awx_version, get_custom_venv_choices, camelcase_to_underscore
+from awx.main.utils import get_awx_version, get_custom_venv_choices, camelcase_to_underscore, datetime_hook
 from awx.main import models
 from django.contrib.sessions.models import Session
 from awx.main.analytics import register
@@ -42,7 +42,7 @@ def trivial_slicing(key, since, until):
 
     horizon = until - timedelta(weeks=4)
     last_entries = Setting.objects.filter(key='AUTOMATION_ANALYTICS_LAST_ENTRIES').first()
-    last_entries = json.loads((last_entries.value if last_entries is not None else '') or '{}')
+    last_entries = json.loads((last_entries.value if last_entries is not None else '') or '{}', object_hook=datetime_hook)
     last_entry = max(last_entries.get(key) or settings.AUTOMATION_ANALYTICS_LAST_GATHER or horizon, horizon)
     return [(last_entry, until)]
 
@@ -55,7 +55,7 @@ def four_hour_slicing(key, since, until):
 
         horizon = until - timedelta(weeks=4)
         last_entries = Setting.objects.filter(key='AUTOMATION_ANALYTICS_LAST_ENTRIES').first()
-        last_entries = json.loads((last_entries.value if last_entries is not None else '') or '{}')
+        last_entries = json.loads((last_entries.value if last_entries is not None else '') or '{}', object_hook=datetime_hook)
         last_entry = max(last_entries.get(key) or settings.AUTOMATION_ANALYTICS_LAST_GATHER or horizon, horizon)
 
     start, end = last_entry, None
@@ -70,7 +70,7 @@ def events_slicing(key, since, until):
 
     last_gather = settings.AUTOMATION_ANALYTICS_LAST_GATHER
     last_entries = Setting.objects.filter(key='AUTOMATION_ANALYTICS_LAST_ENTRIES').first()
-    last_entries = json.loads((last_entries.value if last_entries is not None else '') or '{}')
+    last_entries = json.loads((last_entries.value if last_entries is not None else '') or '{}', object_hook=datetime_hook)
     horizon = until - timedelta(weeks=4)
 
     lower = since or last_gather or horizon
--- a/awx/main/analytics/collectors.py
+++ b/awx/main/analytics/collectors.py
@@ -34,12 +34,58 @@ data _since_ the last report date - i.e., new data in the last 24 hours)
 '''
 
 
+def trivial_slicing(key, since, until):
+    if since is not None:
+        return [(since, until)]
+
+    from awx.conf.models import Setting
+
+    horizon = until - timedelta(weeks=4)
+    last_entries = Setting.objects.filter(key='AUTOMATION_ANALYTICS_LAST_ENTRIES').first()
+    last_entries = json.loads((last_entries.value if last_entries is not None else '') or '{}')
+    last_entry = max(last_entries.get(key) or settings.AUTOMATION_ANALYTICS_LAST_GATHER or horizon, horizon)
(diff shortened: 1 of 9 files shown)


awx/main/analytics/collectors.py
```diff
@@ -282,32 +283,43 @@ def _copy_table(table, query, path):
 
 @register('events_table', '1.2', format='csv', description=_('Automation task records'), expensive=True)
 def events_table(since, full_path, until, **kwargs):
-    events_query = '''COPY (SELECT main_jobevent.id, 
+    # The events_table collector will no longer respect the value of
+    # `since`, as that would potentially cause gaps in the output
+    # under circumstances where continuous output is desired.
+
+    truncate = now() - timedelta(weeks=4)
+    truncate_pk = models.JobEvent.objects.filter(created__lt=truncate).aggregate(Max('pk'))['pk__max']
```
wenottingham: So when this lands, we will automatically send 4 weeks of data regardless?

awx/main/analytics/collectors.py
```diff
@@ -282,32 +283,43 @@ def _copy_table(table, query, path):
...
+    previous_pk = Setting.objects.filter(key='AUTOMATION_ANALYTICS_LAST_EVENT').first()  # Bypass cache
+    previous_pk = max(previous_pk or 0, truncate_pk or 0)
+    final_pk = models.JobEvent.objects.filter(id__gt=previous_pk, created_lte=until).aggregate(Max('pk'))['pk__max']
+    settings.AUTOMATION_ANALYTICS_LAST_EVENT = final_pk
```
wenottingham: This implies the copy is successful, and will cause a skip of data if it's not.

awx/main/analytics/collectors.py
jbradberry: No.  The caller is still doing the 4-hour time slicing, which is being passed in here as `until`.  This is getting converted into a relevant pk value on line 294, as `final_pk`.  Still just 4 hours at a time.

awx/main/analytics/collectors.py
jbradberry: True enough.  I'll see if I can work around that.

awx/main/analytics/collectors.py
jbradberry: What the above line accomplishes is making sure that if our starting point is more than 4 weeks in the past, it will bring it forward to the 4-week mark.  It's needed because since we are dealing with pk values instead of timestamps, the equivalent truncate-at-4-weeks logic back in the caller doesn't do anything with this collector.

awx/main/analytics/collectors.py
jbradberry: I've brought it forward after the `_copy_table` call.

wenottingham: That may not be sufficient if we fail to send the data.

awx/main/analytics/collectors.py
```diff
@@ -282,33 +283,48 @@ def _copy_table(table, query, path):
...
+                              main_jobevent.host_name,
+                              CAST(main_jobevent.event_data::json->>'start' AS TIMESTAMP WITH TIME ZONE) AS start,
                               CAST(main_jobevent.event_data::json->>'end' AS TIMESTAMP WITH TIME ZONE) AS end,
                               main_jobevent.event_data::json->'duration' AS duration,
                               main_jobevent.event_data::json->'res'->'warnings' AS warnings,
                               main_jobevent.event_data::json->'res'->'deprecations' AS deprecations
-                              FROM main_jobevent 
-                              WHERE (main_jobevent.created > '{}' AND main_jobevent.created <= '{}')
+                              FROM main_jobevent
+                              WHERE (main_jobevent.id > {} AND main_jobevent.id <= {})
```
ryanpetrello, line 370: cc @Ladas @jladdjr now that the partitioning PR is taking form, we may want to meet in the middle here and decide how we actually want to land these two things together.  The reason being that in @jladdjr's PR, (as @Ladas pointed out here: ) the primary key of this table is actually now compound (`id`, `job_created`):

https://github.com/ansible/awx/pull/9341#discussion_r584604124

Early on, we had discussed *this* PR being the "first pass" at the problem, but I think the viability of partitioning may be closer than we realize, and we _might_ be able to integrate the two sooner than we thought.  @jladdjr @jbradberry we may want to have a short call and team up to figure out how to integrate these two PRs in `devel`.

Ladas: @ryanpetrello ok, so for backportable version, we will need to include at least `WHERE main_jobevent.job_created > X` , otherwise this would scan all partitions I believe. But finding the right time window will be a hard, to ensure we send 100% of the events.

Maybe it'll be ok doing index scan of all partitions in this case (given we need this to be reliable). Given we order the index as (id, job_created), PG should use the index if we query just by id (it will just do it in cycle for every table, should be a parallel scan, if that's allowed in the PG config). (@jladdjr good thing we've reordered the index, since (job_created, id) wouldn't be used in this case :-) )

ryanpetrello: Yea @Ladas I think unfortunately we're going to have to do a multi-partition index scan here; there's not really any other way I can think of to be sure we don't miss events if e.g., you don't run the collector for a week.

ryanpetrello: @Ladas is there any sort of duplication detection on the analytics side of things?  Is it an issue if we end up sending an event that's already been received?

Ladas: @ryanpetrello yeah, duplicates are handled on Analytics side (everything is atomic upsert based on primary keys)

ryanpetrello: So @Ladas given this, would it be better in the `devel` branch to just keep our code as-is and switch to `modified`, and for @jbradberry's PK-based approach to be something we *only* utilize in backported versions?

I'll admit, just doing `s/created/modified` is a *lot* simpler.  If we had a `modified` index in `devel` as part of @jladdjr's work, our query could just become:

```sql
...
WHERE (
    main_jobevent.job_created > '{since}' AND
    main_jobevent.modified > '{since}' AND main_jobevent.modified <= '{until}'
)
```

If older versions were using pk-based tracking, and then you upgraded, and we went back to the date range based on the combination of the partition key and indexed modified column, and _accidentally_ had some overlap which resulted in us re-sending some stuff we'd already sent, then 🤷 ?

Ladas: @ryanpetrello I think that for this case, it might not matter if we use id or modified? E.g. ^ is the same if for id, if we have events coming 10h after the job was created, they will still be stored in the bucket 10H ago, so the query ^ would miss them.

So we would still be guessing, e.g. limiting `main_jobevent.job_created > '{since - 48h}'` . (But it would miss events for jobs running for more than 2 days)

jbradberry:
Some things to note about this latest version:

- Management of the interval slicing for expensive collectors has been folded in to `analytics.gather`.  This is partly to avoid putting the burden on callers to match up subsets of collectors with their appropriate interval types, and partly to allow alternate entry points (such as the management command) to not gratuitously break.
- For expensive collectors, every interval-sliced _and_ filesize-sliced output file gets their own tarball, which is shipped as each is produced.
- Calls to `analytics.ship` and updates of the settings variables that track the last shipped items/times have also been folded into `analytics.gather`, but will only be invoked if it is not a dry-run.
- Interval slicing code is being attached to expensive collectors using the `expensive` parameter in the `@register` decorator, which is no longer expected to be a boolean value.  This allows such code to be written nearby to the collector for which it is relevant, and hopefully makes such code easy to update when circumstances change (e.g. supporting @jladdjr 's Job Event table partitioning).

awx/main/analytics/core.py
```diff
@@ -82,145 +61,152 @@ def decorate(f):
...
+
+                    if not files:
+                        continue
+                    for fpath in files:
+                        tgzfile = package(dest.parent, {filename: (fpath, func.__awx_analytics_version__)}, until)
+                        if tgzfile is not None:
+                            tarfiles.append(tgzfile)
+
+                            if collection_type != 'dry-run':
+                                ship(tgzfile)
```
Ladas: @jbradberry just making sure, we need to send `config.json` with every payload, since it contains the cluster uuid. Can't see it being added here (I can be missing it though :-) )

awx/main/analytics/core.py
```diff
@@ -82,145 +61,152 @@ def decorate(f):
...
+                                with disable_activity_stream():
+                                    last_entries[key] = end
+                                    settings.AUTOMATION_ANALYTICS_LAST_ENTRIES = last_entries
```
Ladas: change to how we did this before:

before: if there was collection error we've skipped the 4h interval, we moved the time (and there is issue about sending report about this)
current: we'll get stuck on the error (so we'll receive no more events), since we're not moving the id marker?

Tbh. not sure what is better, feels like the way it was before is slightly better, although if we're receiving data with holes in it periodically, it's unusable. But at least the recovery doesn't require a DB change? @ryanpetrello thoughts?

awx/main/analytics/collectors.py
```diff
@@ -280,7 +288,20 @@ def _copy_table(table, query, path):
...
 
-@register('events_table', '1.2', format='csv', description=_('Automation task records'), expensive=True)
+def events_slicing(key, since, until):
+    from awx.conf.models import Setting
+
+    step = 100000
+    last_entries = Setting.objects.filter(key='AUTOMATION_ANALYTICS_LAST_ENTRIES').first()
+    last_entries = last_entries.value if last_entries is not None else {}
+    previous_pk = last_entries.get(key) or 0
+    final_pk = models.JobEvent.objects.filter(created_lte=until).aggregate(Max('pk'))['pk__max']
```
Ladas: Should we still limit the `previous_pk` to max 4 weeks back like the others? If tower can take it, I don't mind sending the whole gap (e.g. if customer stops the collection for a year and then starts it when they'll see some new useful feature)

But if we'd be sending the whole gap, we'll need to send also the other collections. E.g. analytics will not use events that can't be associated to jobs (since we're grouping by job attrs like org, inventory, etc.). 

@ryanpetrello thoughts?

awx/main/analytics/core.py
```diff
@@ -82,145 +61,152 @@ def decorate(f):
...
+                if isinstance(item, str):
+                    info = f.gettarinfo(item, arcname=name)
+                    f.addfile(info)
+                else:
+                    info = tarfile.TarInfo(name)
+                    fileobj = io.BytesIO(json.dumps(item).encode('utf-8'))
+                    fileobj.size = len(fileobj.getvalue())
+                    f.addfile(info, fileobj=fileobj)
+
+            info = tarfile.TarInfo('manifest.json')
```
Ladas: @jbradberry this might be a good place to add the config.json? So it's ensured it's part of every payload?

awx/main/analytics/core.py
jbradberry: No, you are right.  I'll update to ensure it is included.

awx/main/analytics/core.py
jbradberry: My understanding of the issue was that this behavior was what we are aiming for.  cc @wenottingham

wenottingham: I'd prefer to not update the ID if we don't have confirmed submission of the data.

Ladas: @jbradberry @wenottingham ok, that sounds good, just stating the side effect:

- in a case of error we won't be sending any more events (e.g. like we've seen, when we can't read some unicode chars), until the customer updates their Tower with a fix

And I see we're doing this just for events? The rest is using the common last_run. Should we align all collections like this?

---------

Then with sending a report of failures/successes, this is probably good.

awx/main/analytics/collectors.py
Ladas: (also if we're not setting last 4 weeks, we should at least set MIN(id), instead of 0, or this will run for a while from 0. If e.g. the customer produces couple of B events every week and deletes them. )

jbradberry: Ok, I've updated this.

rooftopcellist (approved):

jladdjr:
Started a draft PR where we can tinker with a post-db partitioning version of this change:
https://github.com/jladdjr/awx/pull/17

Ladas (approved):
:+1: looks great

awx/main/analytics/collectors.py
```diff
@@ -33,6 +34,14 @@ def something(since):
 '''
 
 
+def four_hour_slicing(key, since, until):
```
ryanpetrello, line 50: I like this 👍

ryanpetrello (approved):
I really like the way this refactor turned out, great work on this 👍

awx/main/analytics/collectors.py
jladdjr: I was going to say the same thing - was cool to see this generator 👍

shanemcd (approved):

jbradberry:
@Ladas it might be worth taking a look at this PR again.

Ladas (approved):
@jbradberry looks great :+1: Now it's nicely consistent, so it became also a great refactoring of this code :-)

@jbradberry @ryanpetrello so now we keep last timestamp/id per collection, so we should collect all gaps. But I think we'll still need to send a status like described in https://github.com/ansible/tower/issues/4793 , since if the gap is longer than cleanup period, we'll still miss some data?

jbradberry:
@Ladas while reporting the status and collected interval seems like a useful and important thing to do, it seems like that can be a follow-on issue.  Especially since, do we even have anything on the analytics side that would do anything with this information?  There's also some design issues that should be discussed -- csv vs. json, what the reporting should look like when sending up a too-large file that has been split into chunks, whether it truly makes sense to convert pks or other non-datetime intervals back into datetimes when reporting, etc.

Ladas:
@jbradberry ah yes sorry, it should definitely be a follow up, just stating that this PR changes it. :-) (that the gap can happen only after the cleanup kicks in, analytics will use it to check if data are complete)

beeankha (approved):
:+1:

jladdjr:
@jbradberry things with db partitioning have settled down enough again to where I'm looking at this again in the 4.0 context. I'm going to do a hard reset on my `db_partition_analytics` branch so that it points to the head of `db_partition`, plus the commits from here.

rebeccahhh (approved):
🎊  I dig it

ryanpetrello (approved):

jladdjr (approved):

beeankha (approved):

PR #9360, opened 2021-02-18
create performance logger to log api response time

kdelee:
this middleware allready existed, and we were trying to log this
data but it was not working.

Hope is these logs will be able to be shipped via external logging
and we could use kibana to track response time of different endpoints

Just adding logger to `defaults.py` made it print out 
```
awx_1            | 2021-02-18 22:51:17,509 INFO     [2d049242] awx.analytics.performance api response times
``` 

so it seems the "Extra" keyword is just used by rsyslog (I know because i took it out, and rsyslog started throwing errors)

Now it prints this to console:

```
awx_1            | 2021-02-18 23:22:32,908 INFO     [540fd291] awx.analytics.performance request: <WSGIRequest: GET '/api/v2/job_events/'>, response_time: 0.593s
awx_1            | 172.18.0.1 GET /api/v2/job_events/ - HTTP/1.1 200
```

Still need to check if 

```
    "LOG_AGGREGATOR_LOGGERS": [
        "awx",
        "activity_stream",
        "job_events",
        "system_tracking",
        "performance"
    ],
    "LOG_AGGREGATOR_LEVEL": "DEBUG",
```

successfully causes this to get shipped to external logger

### DIFF
--- a/awx/main/middleware.py
+++ b/awx/main/middleware.py
@@ -45,7 +45,10 @@ class TimingMiddleware(threading.local, MiddlewareMixin):
         response['X-API-Total-Time'] = '%0.3fs' % total_time
         if settings.AWX_REQUEST_PROFILE:
             response['X-API-Profile-File'] = self.prof.stop()
-        perf_logger.info('api response times', extra=dict(python_objects=dict(request=request, response=response)))
+        perf_logger.info(
+            f'request: {request}, response_time: {response["X-API-Total-Time"]}',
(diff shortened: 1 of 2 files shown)


kdelee:
I can get what I need from just making a custom conf file in `/etc/nginx/conf.d/` and then a custom config in `/var/lib/awx/rsyslog/conf.d/timing.conf`

It may still be a good idea to fix this "logger to nowhere"

ryanpetrello (approved):
@kdelee I'm fine merging this whenever you're ready 👍

ryanpetrello:
regate

Issue #8464, opened 2020-10-27
CLI import of surveys does not work


AnnSopVier:
##### ISSUE TYPE
 - Bug Report

##### ENVIRONMENT
* AWX version: 3.7.3.
* AWX install method: docker on linux
* Ansible version:  2.9.14
* Operating System: CentOS
* Web Browser: Firefox

##### STEPS TO REPRODUCE

Install Ansible and AWX and also CLI.  Copy the inventory and the survey.json file in /var/liv/awx/projects/00_NewCustomer of the docker container awx_web.
Enter the following commands via the CLI on the docker container awx_web:

- awx organizations create --name "trustedgate" 
awx users create --username "User" --is_superuser false --is_system_auditor false --password "password"
awx inventory create --name "00_NewCustomerSurvey_Inventory" --organization "trustedgate" 
awx projects create --name "00_NewCustomerSurvey_Project" --local_path @/var/lib/awx/projects/00_NewCustomer --organization "trustedgate"
awx job_templates create --name "00_NewCustomerSurvey_Template" --job_type run --inventory "00_NewCustomerSurvey_Inventory" --project "00_NewCustomerSurvey_Project" --playbook "00_NewCustomerSurvey.yaml" --survey_enabled true

All of these steps work and I can see the templates in the UI. 

If I now try to add a survey to the job_template (same result for both ways)  the survey is not added in the UI and I get the following feedback on the console.

awx job_templates delete --name "00_NewCustomerSurvey_Template" 

awx job_templates create --name "00_NewCustomerSurvey_Template" --job_type run --inventory "00_NewCustomerSurvey_Inventory" --project "00_NewCustomerSurvey_Project" --playbook "00_NewCustomerSurvey.yaml" --
survey_spec '{
  'description': 'A survey',
  'name': 'my survey'
  'spec': [{
      'index': 0,
      'question_name': 'What is your department',
      'required': False,
      'variable': 'department',
      'type': 'password',
      'max': 3
    }],
}' --survey_enabled=true

awx job_templates delete --name "00_NewCustomerSurvey_Template" 

awx job_templates modify --id "00_NewCustomerSurvey_Template" --survey_spec @/var/lib/awx/projects/00_NewCustomer/survey.json --survey_enable=true

The content of survey.json is:

{
  'description': 'A survey',
  'spec': [{
      'index': 0,
      'question_name': 'What is your department',
      'required': False,
      'variable': 'department',
      'type': 'password',
      'max': 3
    }],
[...]

PR #9362, opened 2021-02-19
added survey_spec to JobTemplate API endpoints

pnhowe:
##### SUMMARY
added "survey_spec" to the fields of the JobTemplate model, this enables the awx cli to be able to upload survey specs.

related #8464

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
17.0.1
```

##### ADDITIONAL INFORMATION
This is a PR to implement the change as mentioned in "CLI import of surveys does not work #8464", there was a request for a PR, so here it is.  I have tried it out my self and it seems to function as it should.

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -2921,8 +2921,8 @@ class JobTemplateSerializer(JobTemplateMixin, UnifiedJobTemplateSerializer, JobO
             'ask_variables_on_launch', 'ask_limit_on_launch', 'ask_tags_on_launch',
             'ask_skip_tags_on_launch', 'ask_job_type_on_launch', 'ask_verbosity_on_launch',
             'ask_inventory_on_launch', 'ask_credential_on_launch', 'survey_enabled',
-            'become_enabled', 'diff_mode', 'allow_simultaneous', 'custom_virtualenv',
-            'job_slice_count', 'webhook_service', 'webhook_credential',
+            'survey_spec', 'become_enabled', 'diff_mode', 'allow_simultaneous',
+            'custom_virtualenv', 'job_slice_count', 'webhook_service', 'webhook_credential',
         )
 
     def get_related(self, obj):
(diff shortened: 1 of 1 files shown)


ryanpetrello (changes requested):
Hey @pnhowe,

Thanks for this PR.  While this is simple, it's probably the wrong approach w/ the current API.  There's a survey-spec specific API endpoint `/api/v2/job_templates/N/survey_spec/` that implements a bunch of survey-specific validation; the CLI probably needs code to interface with that for managing the survey spec.

pnhowe:
Not the PR you were looking for eh @ryanpetrello ;-)  ... While pushing those changes in I also noticed "JobTemplateWithSpecSerializer", which seems to be JobTemplateSerializer plus the survey_spec, was this another way to modify the spec that the API designers had in mind?  Anything we can do to light a fire under the CLI wizards to get the CLI updated to use a better interface?

ryanpetrello:
> light a fire under the CLI wizards to get the CLI updated to use a better interface

Find them more hours in a day 😄 

This is an area we're definitely looking for volunteers.  I'll also admit, however, that this is very much a special case where the existing API doesn't match the same RESTful patterns as other top-level endpoints.

If I imagined how this would work, there'd be some new code (in the CLI) such that `awx job_templates modify N survey_spec=...` would just transparently make the correct API requests.

jbradberry:
Closing this, since we need to approach the issue a bit differently.

earchibald-lv:
OK, this is closed.

Will we EVER be able to at least **read** the survey spec without hacking the CLI? This is ridiculous.

jbradberry:
@earchibald-lv it's admittedly a bit of a workaround, but reading and writing of `survey_spec` is supported using the `awx import` and `awx export` sub-commands.

PR #9368, opened 2021-02-19
Add support for replace/revert on secret credential fields

mabashian:
##### SUMMARY
link #7256 

Note that this only applies to editing an existing credential.  You should not see this button on fields when adding a new credential.

When editing an existing credential the replace button should show up on fields where `secret` is true _and_ the field has an existing value _that is not an external credential_.  Examples:

Screen Shot 2021-02-19 at 2 09 54 PM
Screen Shot 2021-02-19 at 2 10 43 PM
Screen Shot 2021-02-19 at 2 12 07 PM

Fields with external credentials should look the same:

Screen Shot 2021-02-19 at 2 12 59 PM

Initially the button tooltip should say `Replace`.  Clicking `Replace` will clear out the previously saved value and enable the form field:

replace_revert

The tooltip will change to `Revert`.  Clicking `Revert` will take the field back to it's original state.

I also noticed a race condition which would result in the input fields (subform) not being populated due to the form rendering before the request(s) were completed.  I fixed this.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/screens/Credential/CredentialAdd/CredentialAdd.jsx
+++ b/awx/ui_next/src/screens/Credential/CredentialAdd/CredentialAdd.jsx
@@ -127,7 +127,7 @@ function CredentialAdd({ me }) {
       </PageSection>
     );
   }
-  if (isLoading && !result) {
+  if (isLoading) {
     return (
       <PageSection>
         <Card>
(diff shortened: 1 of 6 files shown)


nixocio (approved):

tiagodread:
This is currently failing our visual tests cc @unlikelyzero

tiagodread (approved):
Related to this change we had a broken e2e test which I'll be working on it and will identify cov improvements there.

Also, filed this [issue](https://github.com/ansible/awx/issues/9635) while verifying this PR

Issue #7010, opened 2020-05-13
awx collection tower_group: append to existing host list


fxmick:
**SUMMARY**
When using hosts parameter to add hosts to a group, all existing hosts are deleted.

**ISSUE TYPE**
Enhancement

**COMPONENT NAME**
tower_group

**ADDITIONAL INFORMATION**
Add an overwrite parameter to prevent deletion of existing hosts.

ryanpetrello:
@AlanCoding is this always how the old module worked, or is this a regression in the new collection?

AlanCoding:
the old module never supported group hosts at all. This is all new territory. We need to take our time and carefully discuss any implications of the interface design. My initial feeling is that we should allow "magic" words in the list like:

```yaml
tower_group:
  name: foo
  hosts:
    - "<existing>"
    - hostB
```

That would still be idempotent, so I'm not against the general concept.

This would incur a some validation burden, but I would prefer it to new parameters. The number of modules & parameters is too high, it's difficult to manage.

fxmick:
This "magic" word trick would be great !
I currently managed this with a mix of api calls to get the the current list of hosts and add new one(s) with the collection.

john-westcott-iv:
Will this "existing" pattern apply to all parameters that accept multiple things? i.e. `extra vars` in job_templates, notification fields, labels, etc. (I'd like to think yes)

From an implementation perspective it might be easier to have a a flag like `keep_existing_hosts: True` otherwise we will have to modify any resolution loops which are attempting to lookup IDs by name to handle the case of our keyword.

AlanCoding:
> i.e. `extra vars` in job_templates, notification fields, labels, etc. (I'd like to think yes)

There are many problems with dict-like fields. The API inconsistently accepts YAML strings, JSON strings, or dict objects. Right now the collection only has to worry about the type that it _gives to_ the API, and if you added logic to combine with existing, then you would have to worry about the type that the API _gives back_.

> otherwise we will have to modify any resolution loops which are attempting to lookup IDs by name to handle the case of our keyword.

Not every string is a valid hostname, see #3384. Provided that the keyword has at least 1 character that would make it unusable as an _actual_ Ansible hostname, conflicts should not be a reasonable concern for this particular case.

However, I see this as being implemented generally for all many-to-many relationships, so it may be a valid concern in other cases. I still think the benefits outweigh the risks.

planeturban:
[...]

Issue #7008, opened 2020-05-13
Add copy option to awx collection modules


fxmick:
##### ISSUE TYPE
 - AWX collection modules  only support from scratch creation

##### SUMMARY
Add an option in the module parameters : 
from: existingItemName

This feature already exists for all kind of items using REST API :
POST
/api/v2/job_templates/{id}/copy/

Fail if "from item" does not exist

sean-m-sullivan:
Been mulling this over as a feature for most modules, since usually they lookup existing item, we add a field for copy_from_existing, and have them give the name/id of the object in tower they want to copy and just like now we do an existing item lookup and maybe use the fields from that? we'd need to remove the id field, but it may be an elegant way of implementing this.

sean-m-sullivan:
Wanted to note here, in case I or someone else develops this, that ansible has a combine vars function
```
from ansible.utils.vars import combine_vars
```

PR #9374, opened 2021-02-22
Add Copy to option to awx collection modules.

sean-m-sullivan:
##### SUMMARY
related #7010 #7008 

This adds the copy from option using tower endpoints. This allows a user to specify an object to copy from. Any provided parameters will then be pushed to update the new object. 

In addition this allows for the preservation of hosts and children when updating groups. It retrieves exiting groups or children appends to those being added and pushes all of them as part of the change. Current behavior overwrites all groups and children when making changes. This allows for either children or groups to be idempotent if changing one and not the other. 

Endpoints used
/api/v2/projects/{id}/copy/
/api/v2/notification_templates/{id}/copy/
/api/v2/credentials/{id}/copy/
/api/v2/inventories/{id}/copy/
/api/v2/job_templates/{id}/copy/
/api/v2/workflow_job_templates/{id}/copy/
/api/v2/groups/{id}/all_hosts/
/api/v2/groups/{id}/children/

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - awx collection

##### AWX VERSION
17.0.1


awx_collection/tests/integration/targets/tower_credential/tasks/main.yml
```diff
@@ -369,7 +392,7 @@
       - result is failed
       - "'Unable to create credential {{ vault_cred_name2 }}' in result.msg"
       - "'Additional properties are not allowed' in result.msg"
-      - "'\\'vault_password\\' was unexpected' in result.msg"
+      - "'\\'vault_********\\' was unexpected' in result.msg"
```
beeankha: This change needs to be reverted (otherwise the test playbook fails).

sean-m-sullivan: I changed this to work locally and thought assumed it would do the same on the regular test, Reverted the change.

awx_collection/plugins/modules/tower_credential.py
```diff
@@ -274,6 +282,12 @@
       vault_password: 'new_password'
       vault_id: 'My ID'
 
+- name: Copy Credential
+  tower_credential:
+    name: Copy password
+    copy_from: Example password
+    credential_type: Vault
```
john-westcott-redhat, line 289: We should modify these params so that credential_type is not required. It could be confusing if you are copying a machine credential and specify a vault type.

awx_collection/plugins/modules/tower_credential.py
```diff
@@ -385,6 +401,28 @@ def main():
...
+            module.fail_json(msg="A credential with the name {0} already exists.".format(name))
+        else:
+            # Lookup existing credential.
+            copy_lookup_data = {
+                'credential_type': cred_type_id,
+            }
+            copy_from_lookup = module.get_one('credentials', name_or_id=copy_from, **{'data': copy_lookup_data})
+            if copy_from_lookup is None:
+                module.fail_json(msg="A credential with the name {0} was not able to be found.".format(copy_from))
+            else:
```
john-westcott-redhat: This seems like a redundant pattern. Can we move this logic of checking the target name and checking that the source item exist into the copy function itself to reduce code in modules?

awx_collection/plugins/modules/tower_inventory.py
```diff
@@ -119,6 +160,9 @@ def main():
         # If the state was absent we can let the module delete it if needed, the module will handle exiting from this
         module.delete_if_needed(inventory)
 
+    # Reset Org id to push in case copy_from was used.
+    org_id = module.resolve_name_to_id('organizations', organization)
```
john-westcott-redhat: If you move the logic into the copy_item function this may not be needed.

awx_collection/plugins/modules/tower_project.py
```diff
@@ -254,22 +271,14 @@ def main():
 
     # Extract our parameters
     name = module.params.get('name')
-    description = module.params.get('description')
```
john-westcott-redhat, line 255: Why did a bunch of these lines go away?

awx_collection/plugins/modules/tower_workflow_job_template.py
```diff
@@ -208,6 +224,34 @@ def main():
...
+            copy_from_lookup = module.get_one('workflow_job_templates', name_or_id=copy_from)
+            if copy_from_lookup is None:
+                module.fail_json(msg="A workflow job template with the name {0} was not able to be found.".format(copy_from))
+            else:
+                workflow_copy_get_check = module.get_endpoint(copy_from_lookup['related']['copy'])
+                if workflow_copy_get_check['status_code'] in [200]:
+                    if (workflow_copy_get_check['json']['can_copy'] and workflow_copy_get_check['json']['can_copy_without_user_input'] and
+                        not workflow_copy_get_check['json']['templates_unable_to_copy'] and not workflow_copy_get_check['json']['credentials_unable_to_copy']
+                            and not workflow_copy_get_check['json']['inventories_unable_to_copy']):
+                        # Because the initial copy will keep its organization, this can be different then the specified one.
```
john-westcott-redhat: In this instance you are checking the can_copy response from the copy endpoint but you don't for other objects. Perhaps this can become common if the code is centralized in copy_item?

awx_collection/test/awx/test_completeness.py
```diff
@@ -30,7 +30,7 @@
 
 # Global module parameters we can ignore
 ignore_parameters = [
-    'state', 'new_name', 'update_secrets'
+    'state', 'new_name', 'update_secrets', 'copy_from'
```
john-westcott-redhat, line 33: I'd like to see if we can actually inspect endpoints for related/copy URLs and then report if the module should support copy or not. But that can be a different PR.

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -410,6 +410,27 @@ def modify_associations(self, association_endpoint, new_association_list):
             else:
                 self.fail_json(msg="Failed to associate item {0}".format(response['json'].get('detail', response['json'])))
 
+    def copy_item(self, existing_item, name, item_type='unknown'):
```
john-westcott-redhat: See comments below about merging more code in here but some cleanup on the variable names might be helpful. i.e. existing_item => item_to_copy, name => new_item_name, etc.

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -410,6 +410,27 @@ def modify_associations(self, association_endpoint, new_association_list):
...
+            item_url = response['json']['url']
+        else:
+            if 'json' in response and '__all__' in response['json']:
+                self.fail_json(msg="Unable to create {0} {1}: {2}".format(item_type, name, response['json']['__all__'][0]))
+            elif 'json' in response:
+                self.fail_json(msg="Unable to create {0} {1}: {2}".format(item_type, name, response['json']))
+            else:
+                self.fail_json(msg="Unable to create {0} {1}: {2}".format(item_type, name, response['status_code']))
+        last_data = response['json']
+        return last_data
```
john-westcott-redhat: This response is never referenced anywhere so it can be removed.

john-westcott-redhat (commented):
Thanks for the copy feature, good idea. See my comments below.

awx_collection/plugins/modules/tower_project.py
sean-m-sullivan: They moved to line 352 to only get them if the value is not None, Otherwise sometimes they would get overwritten on the copy command as it was set to None. 
With the way it was originally, if description was not supplied it would become "", because description returned None here.

awx_collection/test/awx/test_completeness.py
sean-m-sullivan: agreed, that would be a good feature, will think about it :)

awx_collection/plugins/modules/tower_workflow_job_template.py
sean-m-sullivan: In this case, only the workflows have this check. 
[project copy endpoint](https://docs.ansible.com/ansible-tower/latest/html/towerapi/api_ref.html#/Projects/Projects_projects_copy_list)
[workflow copy endpoint](https://docs.ansible.com/ansible-tower/latest/html/towerapi/api_ref.html#/Workflow_Job_Templates/Workflow_Job_Templates_workflow_job_templates_copy_list)
If the can_copy options were added to other endpoints moving this to the method would be good.

awx_collection/plugins/modules/tower_credential.py
sean-m-sullivan: Credentials requires Kind or Credential type atm, regardless of copy, and I believe we are phasing out kind, where credential type will then be required, This was more just not messing with the basic functioning of the module, and credential_type can be machine as it is on line 210 of the tower_credential/tasks/main.yml test.

awx_collection/plugins/modules/tower_credential.py
sean-m-sullivan: I can try and do that, was debating it myself, but some of the items were specific to each one. Will think about this and see about getting the code simplified into the function.

awx_collection/plugins/module_utils/tower_api.py
```diff
@@ -410,6 +410,47 @@ def modify_associations(self, association_endpoint, new_association_list):
             else:
                 self.fail_json(msg="Failed to associate item {0}".format(response['json'].get('detail', response['json'])))
 
+    def copy_item(self, existing_item, copy_from_name_or_id, new_item_name, endpoint=None, item_type='unknown', copy_lookup_data={}):
+
+        if existing_item is not None:
```
john-westcott-iv, line 415: One other o'dark 30 thought I had was that this check makes your playbooks not idempotent.... i.e. if you had a playbook with a copy_from you could run it once but the second run causes a failure in the module halting your playbook. You could get around this by making your task with the copy_from wrapped in a when clause but then you have to do the lookup yourself out side. I am questioning if we want this to be a failure or if want an existing item to just nullify the copy from? i.e. if the target object already exists we just do a return here and the copy never takes place but the rest of the module runs. We could document this behavior and then a playbook can be run over and over w/o a when condition on the task. Thoughts?

sean-m-sullivan: Definitely doable, as a return would just return back to the module to continue as normal, IF we do this, We should make changed = false, with the rest of the module determining if it then gets changed to true, and a "copied" = false if it doesn't, and true if it does. That way the user can parse the results easier. Additionally maybe change this line to self.warn?

Went through and updated the code, Reason for both changed and copied to be different is that its possible for it to Not copy, and then make additional changes, so this helps keeps track of which function was executed.

sean-m-sullivan: Made said changes and updated test logic.

awx_collection/plugins/module_utils/tower_api.py
sean-m-sullivan: Changed so it is useful with now that other changes have been implemented

awx_collection/plugins/module_utils/tower_api.py
sean-m-sullivan: Cleaned

awx_collection/plugins/modules/tower_inventory.py
sean-m-sullivan: Moved, and your right

awx_collection/plugins/modules/tower_credential.py
sean-m-sullivan: Done

beeankha (approved):

Issue #9376, opened 2021-02-22
Disable fields name, surname and email due to GDPR requirements


lucas-benedito:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
As per GDPR requirement the data name, surname and email cannot be stored in the systems but those fields cannot be disabled in Ansible Tower UI at the moment.
Suggestions:
- Create toggle option for GDPR where you can disable specific fields
- Create user role in order to disable those fields if not system administrator

xstasi:
FWIW I'm not entirely convinced that this is not a bug: if a system is not GDPR compliant when it holds personal data that users may not be able to retain control of, it should not be possible for unprivileged users to "force" the system out of compliance.

This is even more true when authentication is done externally, so tower should be holding virtually no user data.

Rather than disabling specific fields, I think that there should be an optional role that allows users to self-edit. It could be enabled by default for retrocompatibility.

wenottingham:
> As per GDPR requirement the data name, surname and email cannot be stored in the systems

While I don't want to speak for their auditor, on the face of it that's either an overly restrictive corporate policy that goes well above and beyond, or an incorrect reading of the standard.

For a deployment with internal users only to that company, it's *definitely* incorrect - all of the username fields are already covered under the normal legal basis for data collection. Even with external users it's likely covered under whatever agreement that they're granting them accounts to the automation system under. and the GDPR then just specifies how that data is protected.

In any case, they're all optional fields regardless in the API, and in AWX most of them are now optional in the UI.

xstasi:
@wenottingham the customer's use case is to address users with "self-harming" behaviour, who go and input their personal data in their Tower account. By doing so, the users turn the tower node to a "GDPR-relevant" machine which it would otherwise not be.

That is why they would like users to not be able to self edit, and I believe it makes sense: it should not be obvious that users are in charge of defining their own account parameters. Being optional in this case is, while appreciated, not enough.

wenottingham:
By default, fields that come from an IDP (LDAP, etc) are replaced by the IDP data on every login; beyond that, there are service accounts and tokens if they are not interested in having user accounts tied to individuals.

xstasi:
They want to have individual accounts tied to each users, they just don't want any personal data other than the login to be recorded.

[...]

PR #9381, opened 2021-02-22
Make email not required for a user in the UI.

wenottingham:
It's already not required in the API.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

See https://github.com/ansible/awx/issues/9376

### DIFF
--- a/awx/ui_next/src/screens/User/shared/UserForm.jsx
+++ b/awx/ui_next/src/screens/User/shared/UserForm.jsx
@@ -11,7 +11,7 @@ import FormField, {
   FormSubmitError,
 } from '../../../components/FormField';
 import OrganizationLookup from '../../../components/Lookup/OrganizationLookup';
-import { required, requiredEmail } from '../../../util/validators';
+import { required } from '../../../util/validators';
 import { FormColumnLayout } from '../../../components/FormLayout';
 
 function UserFormFields({ user, i18n }) {
@@ -76,8 +76,7 @@ function UserFormFields({ user, i18n }) {
         id="user-email"
         label={i18n._(t`Email`)}
         name="email"
-        validate={requiredEmail(i18n)}
-        isRequired
+        type="text"
       />
       {!ldapUser && !(socialAuthUser && externalAccount) && (
         <>
--- a/awx/ui_next/src/types.js
+++ b/awx/ui_next/src/types.js
@@ -258,7 +258,7 @@ export const User = shape({
   username: string,
   first_name: string,
   last_name: string,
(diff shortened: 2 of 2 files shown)


awx/ui_next/src/screens/User/shared/UserForm.jsx
```diff
@@ -11,7 +11,7 @@ import FormField, {
   FormSubmitError,
 } from '../../../components/FormField';
 import OrganizationLookup from '../../../components/Lookup/OrganizationLookup';
-import { required, requiredEmail } from '../../../util/validators';
```
ryanpetrello, line 14: Is there some validator that's "not required, but if you fill it in, we'll yell if it doesn't look email-like" ?

ryanpetrello (approved):

ryanpetrello:
This looks fine to me, but I defer to @jakemcdermott and @mabashian, two people who *actually* know JavaScript.

awx/ui_next/src/screens/User/shared/UserForm.jsx
jakemcdermott: From what I can see, this PR does that:

useremail

jakemcdermott (approved):
Looks great, thanks

awx/ui_next/src/screens/User/shared/UserForm.jsx
jakemcdermott: ^ Actually, that message is coming from the server. It's just so darn fast I couldn't tell the difference :sweat_smile:

jakemcdermott: ... Anyway, if you wanted to do something like this (with a client-side validator to avoid a request) you _totally_ could.

You could do it by copying the `requiredEmail` validator, renaming it, and removing this line: https://github.com/ansible/awx/blob/devel/awx/ui_next/src/util/validators.jsx#L50-L52

unlikelyzero:
@wenottingham this change is causing the tests to fail here: https://dashboard.cypress.io/projects/ui2mq7/runs/4547/overview

awx/ui_next/src/screens/User/shared/UserForm.jsx
ryanpetrello: Eh, I think it's fine if it's enforced on the API side.

unlikelyzero:
https://dashboard.cypress.io/projects/ui2mq7/runs/4555

wenottingham:
regate

PR #9382, opened 2021-02-22
make sure we use built in credential type

kdelee:
this way we can pass kind="registry" to akit create method and
we get the correct built in type

@rebeccahhh @yagomarques


yagomarques (approved):

rebeccahhh (approved):
wfhparty

shanemcd (approved):

jbradberry (approved):

Issue #9056, opened 2021-01-11
Observability and metrics - metrics dashboard


chrismeyersfsu:
We want to view the metrics we expose. Ideally, we don't require there to be a Prometheus instance gathering the prometheus data that we expose in order to get a basic view of Tower metrics. We want a view of graphs that can display the basic prometheus data types: Counter, Gauge, Histogram over time. Over time means that when the dashboard is opened, the metrics will be gathered either from the exposed prometheus endpoint or from a special websocket group that we add. No data will be loaded from the past.  If the dashboard is refreshed, the metric history is lost and starts over.

The dashboard will be useful for debugging a problem or suspected problem live.

Once https://github.com/ansible/awx/issues/9012 and https://github.com/ansible/awx/issues/9019 land the UI will have something to consume.

unlikelyzero:
@chrismeyersfsu can you help me understand when work might begin on the ui component

mabashian:
@unlikelyzero I'll try to explain as best as I understand it.  @chrismeyersfsu can probably provide more color.  The basic idea is that there will be a `metrics` endpoint that the UI can periodically query.  The ask here is for us to build a single page that displays a graph.  The data used to populate that graph will come from polling that metrics endpoint on some interval.  It'll be our responsibility on the client to store the necessary data and update the graph.  There's no storage mechanism on the backend.  If the user refreshes the page, they start with a fresh, empty graph.  As we poll that endpoint, the graph will update.  The user also needs the ability to select which field on that endpoint they want to see graphed so there will probably be a dropdown somewhere (akin to the dashboard graph dropdowns).

wenottingham:
Copying some discussion from the meeting:
- This needs to be somewhere in the UI, and RBAC'd so only system admins see it (it's not part of any other user persona)
- Logically, this is part of managing the cluster. The other part of the UI that fits here is Instance Groups, so it fits alongside/with that.

cc @trahman73 for ideas on how to organize this, as either another menu item or a new combo menu item.

trahman73:
This makes sense to live under the Admin menu subset as a stand alone page. The page can contain a filter toolbar component with the graph component underneath.

chrismeyersfsu:
@mabashian On top of basic plotting. It's looking like we are going to need at least a `division` operator. For example, we have two related metrics `events_inserted` and `time_to_insert_events_total`. We want to be able to visualize events inserted per second over time, which would be `events_inserted` / `time_to_insert_events_total`.

[...]

Issue #9019, opened 2021-01-06
Observability and metrics - Surface metrics prometheus friendly


chrismeyersfsu:
Websockets contain an [example](https://github.com/ansible/awx/blob/devel/awx/main/analytics/broadcast_websocket.py#L121) of Prometheus data structures serialized, deserialized & [exposed via a Prometheus endpoint](https://github.com/chrismeyersfsu/awx/commit/b58c71bb74bd5f5c836f88daa3d7806ff68cea47#diff-497de0f3f551df8d0333aa7a3d2cae353a08d4e2b500ca122195bafda82f6d3b).  We want to do the same for our generalized metrics. Note that websockets didn't have the potential performance needs that this metrics tracking system will need. I'm unsure of the performance of storing the metrics in Prometheus objects and _then_ serializing them. Also, the websockets stats objects persist in memory for a long time. There are some metrics workloads where the metric object will be short lived in memory. So there will be many deserialize, modify, serialize operations. For this reason, we may have our own minimal data structure that we operate on (i.e. serialize, increment, deserialize) and only convert it to a Prometheus style data-structure when rendering in the API.

"Normal" prometheus usage is to have a `/metrics` endpoint per-node that you are monitoring. We are deviating from the norm and centralizing information about multiple nodes. It would be nice to have a filter so that you get the best of both worlds, centralized metrics and the ability to use existing tooling. To this end, a filter that supports filtering by-node would be useful i.e. `/metrics?node=node1.awx.mydomain.com`

We also have to consider performance. This new set of metrics should render very quickly. Our existing metrics can be slow since they require database queries. We should use filtering or a different endpoint to allow users to avoid the pre-existing DB query metrics.

kdelee:
- add testing for filtering in tower-qa
- add new attributes to existing tests for existence/correctness
  - basic correctness,example: num events monotonically increasing
- in cluster installs, verify data is propagated across cluster
  - example, if job is run on other node, show on other node that job count increased on job running node, and that event count increased
- add baseline pass/fail API response time benchmark test to nightly tests
  - if it takes more than a second, fail no if ands or buts about it

NEED from @fosterseth @chrismeyersfsu 

- need list of each metric, and which ones are multi-node metrics are
- discuss each metric and reasonable test case for each we want a automated correctness test

chrismeyersfsu:
Current thoughts below. Will add other subsystems to this table.

[...]

Issue #9012, opened 2021-01-05
Observability and metrics - Save to Redis


chrismeyersfsu:
TL;DR Store metrics in Redis so they are accessible to the API.

Metrics live in various process spaces but we want to display/view them via other process memory spaces. For example, metrics about the task manager come from a dispatcher process that runs task manager code. The dispatcher processes space may be short lived. We want to present or view, in the API, the metrics that occur in the task manager, but the API is a separate process. We need some IPC mechanism to accomplish this. We can use Redis.

Another consumer of the metrics data would be a periodic task that outputs the values to a file on disk. This file can be collected by sosreport for offline reconstruction of past performance.

Yet another consumer of this data could be a command line tool to aid in debugging performance live.

### Requirements
1. Metric data stored in Redis should be convertible to Prometheus data types: Counter, Gauge, Histogram, Summary https://prometheus.io/docs/concepts/metric_types/ . We should serialize the metrics into JSON (look for a fast JSON serialize/serialize lib)

2. Metrics should be stored efficiently. Maybe metrics about metrics overhead.

We can use the Redis hash data type to store key -> value metrics https://redis.io/commands#hash possible organization:

Single Hash
```
HMSET awx_metrics task_manager_pending_total 2
HMSET awx_metrics task_manager_running_total 5
HMSET awx_metrics api_jobs_created 100
HMSET awx_metrics api_workflow_jobs_created 5
HMSET awx_metrics callback_receiver_events_processed 10000
```

Multiple Hashes
```
HMSET awx_metrics_task_manager pending_total 2
HMSET awx_metrics_task_manager running_total 5
HMSET awx_metrics_api jobs_created 100
HMSET awx_metrics_api workflow_jobs_created 5
HMSET awx_metrics_callback_receiver events_processed 10000
```

#### Notes
There will be many cases were many metrics need to be updated for a subsystem per "loop". It can be costly to perform multiple redis operations per-loop if we are in a critical section (i.e. the callback receiver). We could try to use redis async but this can complicate the code. Instead, we should use redis pipeline feature. You can think of the pipeline feature as batching or a transaction.

unlikelyzero:
@kdelee has some testing notes and will add here

kdelee:
we will end up testing this once it is served to the metrics endpoint https://github.com/ansible/awx/issues/9019#issuecomment-771926550

ryanpetrello:
Closing, cc @kdelee @fosterseth

PR #9383, opened 2021-02-22
Add subsystem metrics that propagate through Redis

fosterseth:
##### SUMMARY

#9019 -- list of metrics and their purpose / description
#9012
#9056
#8629

Use Redis to store metrics pertaining to the performance and health of subsystems such as the callback receiver and task manager. It is thread / multiprocess safe and should be fast enough to handle a high volume of data.

Each time a metric is updated, we trigger a `send_metrics()` method that uses websockets to send this data to other tower instances. Once received, other instances call `store_metrics()` to record that data in their own Redis storage.

This data shows up at the `/api/v2/metrics` endpoint
You can filter down nodes using `/api/v2/metrics/?subsystemonly=1&node=awx-1`

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 17.0.1
```


awx/api/views/metrics.py
```diff
@@ -37,6 +38,10 @@ class MetricsView(APIView):
     def get(self, request):
         ''' Show Metrics Details '''
         if (request.user.is_superuser or request.user.is_system_auditor):
-            return Response(metrics().decode('UTF-8'))
+            metrics_to_show = ''
+            if 'redisonly' not in request.query_params:
```
kdelee: nitpick, but wonder if we don't want o directly mention redis here in the query because 1) its an implementation detail maybe users don't know and 2) maybe we'll change out redis someday 

Is there some other way to describe these metrics like....`callback_receiver` since that is what ll the metrics are referencing?

so it would be like `/api/v2/metrics?callback_receiver`

kdelee:
@fosterseth it would be cool if an OPTIONS request to `api/v2/metrics/` described the query params in the `description` like we do for `api/v2/job_templates/` that gets templated out through a few more layers than I was able to follow, but one way or another it is set a the template here https://github.com/ansible/awx/tree/devel/awx/api/templates/api
`job_job_plays_list.md` for example defines custom info about filtering

fosterseth:
renamed branch, new PR here #9461

PR #9289, opened 2021-02-10
Replace Local Docker Install with Community Docker-Compose

rooftopcellist:
Issue: https://github.com/ansible/awx/issues/9125

Initial WIP PR for Replacing the Local Docker installation with a docker-compose method close to what we do for the development environment.  

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - Installer

##### AWX VERSION

```
AWX 50.0.0
```

##### ADDITIONAL INFORMATION

```

```

PR #9387, opened 2021-02-22
Mount SECRET_KEY into dev env & document it

rooftopcellist:
##### SUMMARY
As part of https://github.com/ansible/awx/pull/9289, we need to actually mount the SECRET_KEY file into the awx container to be able to use the migrated data.  

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - Installer

##### AWX VERSION

```
devel
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/settings/defaults.py
+++ b/awx/settings/defaults.py
@@ -1,6 +1,7 @@
 # Copyright (c) 2015 Ansible, Inc.
 # All Rights Reserved.
 
+import base64
 import os
 import re  # noqa
 import sys
@@ -148,7 +149,10 @@ SCHEDULE_MAX_JOBS = 10
 SITE_ID = 1
 
 # Make this unique, and don't share it with anybody.
-SECRET_KEY = open('/etc/tower/SECRET_KEY', 'rb').read().strip()
+if os.path.exists('/etc/tower/SECRET_KEY'):
+    SECRET_KEY = open('/etc/tower/SECRET_KEY', 'rb').read().strip()
+else:
+    SECRET_KEY = base64.encodebytes(os.urandom(32)).decode().rstrip()
 
 # Hosts/domain names that are valid for this site; required if DEBUG is False
 # See https://docs.djangoproject.com/en/dev/ref/settings/#allowed-hosts
--- a/awx/settings/defaults.py
+++ b/awx/settings/defaults.py
@@ -148,7 +148,7 @@ SCHEDULE_MAX_JOBS = 10
 SITE_ID = 1
 
 # Make this unique, and don't share it with anybody.
-SECRET_KEY = 'p7z7g1ql4%6+(6nlebb6hdk7sd^&fnjpal308%n%+p^_e6vo1y'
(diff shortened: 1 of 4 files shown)


shanemcd (changes requested):
In the migration steps you have:

```
$ ansible-playbook migrate.yml -e "migrate_local_docker=true" -e "postgres_data_dir=~/.awx/pgdocker" -e "old_docker_compose_dir=~/.awx/awxcompose"
```

This does not point to the inventory file (which includes the secret_key).

We need something like:

```
$ ansible-playbook -i tools/docker-compose/inventory tools/docker-compose/ansible/migrate.yml \
    -e "migrate_local_docker=true" \
    -e "postgres_data_dir=~/.awx/pgdocker" \
    -e "old_docker_compose_dir=~/.awx/awxcompose"
```

shanemcd (approved):

rooftopcellist:
regate

ryanpetrello (approved):

ryanpetrello:
recheck

awx/settings/defaults.py
```diff
@@ -148,7 +148,7 @@ def IS_TESTING(argv=None):
 SITE_ID = 1
 
 # Make this unique, and don't share it with anybody.
-SECRET_KEY = 'p7z7g1ql4%6+(6nlebb6hdk7sd^&fnjpal308%n%+p^_e6vo1y'
+SECRET_KEY = open('/etc/tower/SECRET_KEY', 'rb').read().strip()
```
ryanpetrello: Makes tests fail.  Should we put this line into `settings/development.py` instead of `defaults.py`?  cc @shanemcd

ryanpetrello: We could also wrap this in a `try...except...` and just generate a random value in the except block (so it'll just make a random one for tests)

rooftopcellist: The SECRET_KEY isn't getting mounted into the container at `/etc/tower` anymore for some reason, even though it is being rendered in the docker-compose.yml.

rooftopcellist: I was able to reproduce this locally, it looks like what is happening is that an old tools_awx_1 container was still around and was being used by the docker-compose, so the new volume-mount wasn't being created.  

After removing all containers and re-running `make docker-compose`, the secret_key was properly mounted into the container.

rooftopcellist: I could probably get tests passing by adding a `--force-recreate` flag in the Makefile.

ryanpetrello: Okay, that works too if we can get tests passing.

shanemcd: I'll take another look at this. I _really_ don't think we should have a default value for the secret key!

shanemcd: @rooftopcellist I don't think `--force-recreate` is going to help here. Can you back out that change please?

ryanpetrello (approved):

PR #9395, opened 2021-02-23
fix dev env migrate.yml path in docs

rooftopcellist:
##### SUMMARY
Correct small path typo.

### DIFF
--- a/tools/docker-compose/docs/data_migration.md
+++ b/tools/docker-compose/docs/data_migration.md
@@ -1,7 +1,7 @@
 # Migrating Data from Local Docker
 
 If you are migrating data from a Local Docker installation (17.0.1 and prior), you can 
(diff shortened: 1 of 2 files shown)


tools/docker-compose/docs/data_migration.md
```diff
@@ -13,7 +13,7 @@ If you had a custom pgdocker or awxcompose location, you will need to set the `p
 
 1. Run the [migrate playbook](./ansible/migrate.yml) to migrate your data to the new postgresql container and convert the data directory to a volume mount.
```
ryanpetrello: This link on line 14 doesn't work.

ryanpetrello (commented):
@rooftopcellist On line 8, we should change it to be explicit, and say:

```
First, in the [inventory file](../inventory)
```

ryanpetrello (approved):

PR #9438, opened 2021-03-01
Update EE breadcrumb to use name instead of image

nixocio:
Update EE breadcrumb to use name instead of image to be consistent with
the other screens.

See: https://github.com/ansible/awx/issues/9087


jakemcdermott (approved):

nixocio:
regate

jakemcdermott:
recheck

PR #9442, opened 2021-03-01
No longer delete local_settings.py

rooftopcellist:
##### SUMMARY
Removes task in docker-compose sources role that removes the old local_settings.py file. 

If you want to add custom dev settings, create a file called `local_whatever.py` in `awx/settings/`, then put your custom settings there.  

 * this assumes people have already migrated from the old dev env.
 * new custom dev settings can be set in /etc/tower/conf.d/local_settings.py
 * this new settings file is bind-mounted from tools/docker-compose/_sources/local_settings.py

##### COMPONENT NAME

 - dev
 - Installer

##### AWX VERSION

```
devel
```

### DIFF
--- a/tools/docker-compose/ansible/roles/sources/tasks/main.yml
+++ b/tools/docker-compose/ansible/roles/sources/tasks/main.yml
@@ -45,11 +45,6 @@
     - "database.py"
     - "websocket_secret.py"
 
-- name: Delete old local_settings.py
-  file:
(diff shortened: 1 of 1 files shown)


shanemcd (approved):

rooftopcellist:
We are not going to remove this task because it may trip up people when switching between old and new branches.  

If you want to add custom dev settings, create a file called local_whatever.py in awx/settings/, then put your custom settings there.

Issue #8356, opened 2020-10-08
Upgrade lingui library


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Upgrade lingui library to latest stable version. 

This will accomplish two goals:
1. Allow for native pseudolocalization support. More about that [here](https://github.com/ansible/awx/issues/9285)
2. Get rid of the console error warnings that is causing so much noise that dev and QE ignore valuable information

<img width="872" alt="image" src="https://user-images.githubusercontent.com/9053044/95508753-78f9cc80-0981-11eb-87a8-e61ccd7b6123.png">

unlikelyzero:
Before this merges, we should try to measure console log output from cypress between branches

AlexSCorey:
@unlikelyzero I was also thinking that we should do visual regression testing on it as well.

mabashian:
https://lingui.js.org/releases/migration-3.html

rooftopcellist:
This will also allow us to run `lingui extract-template` to create `messages.pot` for the UI strings.  This is needed for translations.  

I have tested this out with `@lingui/cli@3.4.0`.  There are other changes needed to get the build/complile targets working, but it worked for running `extract-template`.  

In case it is useful to whoever works on this feature, my diff was:
```
[chadams@chadams-work awx]$ git diff awx/ui_next/package.json
diff --git a/awx/ui_next/package.json b/awx/ui_next/package.json
index 0d470a17ca..7b2f350a0d 100644
--- a/awx/ui_next/package.json
+++ b/awx/ui_next/package.json
@@ -32,7 +32,7 @@
   "devDependencies": {
     "@babel/polyfill": "^7.8.7",
     "@cypress/instrument-cra": "^1.4.0",
-    "@lingui/cli": "^2.9.2",
+    "@lingui/cli": "^3.4.0",
     "@lingui/macro": "^2.9.1",
     "@nteract/mockument": "^1.0.4",
     "babel-core": "^7.0.0-bridge.0",
@@ -64,6 +64,7 @@
     "lint": "eslint --ext .js --ext .jsx .",
     "add-locale": "lingui add-locale",
     "extract-strings": "lingui extract",
+    "extract-template": "lingui extract-template",
     "compile-strings": "lingui compile",
```

Then I installed it and ran the extract-template command:
```
npm install @lingui/cli@3.4.0
npm --prefix awx/ui_next --loglevel warn run extract-template
```

rooftopcellist:
[...]

PR #9443, opened 2021-03-02
Update translations targets

rooftopcellist:
##### SUMMARY
Update make targets for extracting strings for translation.  

This also removes the step to extract strings from the ui build/compile targets, which should save some time and confusion.  

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API
 - UI


unlikelyzero:
@rooftopcellist i believe that these commands will change with update of https://github.com/ansible/awx/issues/8356. Might be worth combining that with this PR

Makefile
```diff
@@ -655,3 +619,49 @@ awx-kube-dev-build: Dockerfile.kube-dev
...
+	    language=`grep '^"Language:' "$$po" | grep '_'`; \
+	    if test "x$$language" != x ; then \
+	        right_language=`echo $$language | sed -e 's/_/-/'`; \
+	        echo "ERROR: WRONG $$language CORRECTION: $$right_language in $$po"; \
+	        exit -1; \
+	    fi; \
+	done;
+else
+	@echo No PO files
+endif
```
jakemcdermott: @rooftopcellist We'll want to configure the formatter to skip the locale files. (This should also get the zuul tests passing)

```shell
cat << EOF > awx/ui_next/.prettierignore
build
src/locales
EOF
```

jakemcdermott (changes requested):

Makefile
```diff
@@ -445,39 +445,6 @@ bulk_data:
 	fi; \
 	$(PYTHON) tools/data_generators/rbac_dummy_data_generator.py --preset=$(DATA_GEN_PRESET)
 
-# l10n TASKS
```
rooftopcellist, line 448: I am removing this `check-po` target because I can't find a place where we actually use it anymore.  It looks like it's purpose was to check the Language header in the .po files to make sure the valid locale for Chinese is set.  

Original commit: https://github.com/ansible/awx/commit/40b94a38a86aa0d2a4cc38861c3d1bd64e3feb3e

jakemcdermott (approved):
lgtm - if all the tests, e2e, etc. are happy then I am happy

Makefile
```diff
@@ -655,3 +619,20 @@ awx-kube-dev-build: Dockerfile.kube-dev
...
 	    -t $(DEV_DOCKER_TAG_BASE)/awx_kube_devel:$(COMPOSE_TAG) .
+
+
+# Translation TASKS
+# --------------------------------------
+
+# generate UI .pot
+pot: $(UI_BUILD_FLAG_FILE)
+	$(NPM_BIN) --prefix awx/ui_next --loglevel warn run extract-strings
+	$(NPM_BIN) --prefix awx/ui_next --loglevel warn run extract-template
```
rooftopcellist, line 630: This exists as a placeholder until we upgrade lingui to v3.  If we don't upgrade, we will need to figure out another way to extract strings.

Issue #8505, opened 2020-10-30
 Log rsyslog failures with a counter


mrmibrown:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
This is a request to log rsyslog failures with a counter to provide more transparency when a message is dropped.

To help avoid the silent dropping of messages it would be good to have metrics keep a counter that is incremented every time a log message is dropped. Maybe have a "dropped external logger message" category that has sub counts for messages dropped because of blocked I/O and connection refused.

Related PR: https://github.com/ansible/awx/pull/7292

wenottingham:
Where is this expected to count? When rsyslog refuses the message? When rsyslog can't send it to wherever?

I'm skeptical this is practical in the current architecture.

ryanpetrello:
I tend to agree; if the rsyslogd daemon is unresponsive over a local domain socket, 🤷

blomquisg:
This is essentially a dup of https://github.com/ansible/awx/issues/8576.  We don't see a compelling reason to include this.

suryarao67:
Hello,

If there is no counter and no messages indicating messages are dropped, how do support teams even become aware of this? Bear in mind that these logs often may contain critical data which cannot be lost. Silent dropping is not good. There must be an indication when messages are being dropped..somewhere.

wenottingham:
That would be in the rsyslog error log in /var/log/tower.

wenottingham:
That would be in the rsyslog error log inside the container.

suryarao67:
If rsyslog itself was unresponsive, the Tower process would drop the external log messages being sent to rsyslog to avoid blocking. In that case I don't think there will be anything inside rsyslog error log. Please confirm?

wenottingham:
rsyslog has not been unresponsive in our testing - if this is a continuous issue in some environment, that is something that we would need details about the AWX deployment.

As for logging failures, I don't know that it makes sense to log a failure when you can't log, but perhaps we can expose a count of failed logs via our prometheus metrics that are being enhanced to store some per-node metrics for a future release. CC'ing @fosterseth .

suryarao67:
> rsyslog has not been unresponsive in our testing - if this is a continuous issue in some environment, that is something that we would need details about the AWX deployment.

If rsyslog has never been observed to be unresponsive, then I am confused as to why the need arose to drop messages as part of PR# [7292](https://github.com/ansible/awx/pull/7292)

[...]

PR #9452, opened 2021-03-02
make rsyslogd socket emit failures a bit less verbose (but still write to stderr)

ryanpetrello:
see: https://github.com/ansible/awx/issues/8505

### DIFF
--- a/awx/main/utils/external_logging.py
+++ b/awx/main/utils/external_logging.py
@@ -32,7 +32,7 @@ def construct_rsyslog_conf_template(settings=settings):
         '$IncludeConfig /var/lib/awx/rsyslog/conf.d/*.conf',
         f'main_queue(queue.spoolDirectory="{spool_directory}" queue.maxdiskspace="{max_disk_space}g" queue.type="Disk" queue.filename="awx-external-logger-backlog")',  # noqa
         'module(load="imuxsock" SysSock.Use="off")',
-        'input(type="imuxsock" Socket="' + settings.LOGGING['handlers']['external_logger']['address'] + '" unlink="on")',
+        'input(type="imuxsock" Socket="' + settings.LOGGING['handlers']['external_logger']['address'] + '" unlink="on" RateLimit.Burst="0")',
         'template(name="awx" type="string" string="%rawmsg-after-pri%")',
     ])
 
--- a/awx/main/utils/handlers.py
+++ b/awx/main/utils/handlers.py
@@ -3,7 +3,8 @@
 
 # Python
 import logging
-import os.path
+import sys
+import traceback
 
 # Django
 from django.conf import settings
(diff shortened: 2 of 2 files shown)


awx/main/utils/external_logging.py
```diff
@@ -32,7 +32,7 @@ def construct_rsyslog_conf_template(settings=settings):
         '$IncludeConfig /var/lib/awx/rsyslog/conf.d/*.conf',
         f'main_queue(queue.spoolDirectory="{spool_directory}" queue.maxdiskspace="{max_disk_space}g" queue.type="Disk" queue.filename="awx-external-logger-backlog")',  # noqa
         'module(load="imuxsock" SysSock.Use="off")',
-        'input(type="imuxsock" Socket="' + settings.LOGGING['handlers']['external_logger']['address'] + '" unlink="on")',
+        'input(type="imuxsock" Socket="' + settings.LOGGING['handlers']['external_logger']['address'] + '" unlink="on" RateLimit.Burst="0")',
```
ryanpetrello, line 35: Noticed this while I was reading rsyslogd docs - the _default_ suggests it's basically 200 per second per pid, but I wasn't able to reproduce that.  To be on the safe side, I'm gonna set this to "no limit".

awx/main/utils/handlers.py
```diff
@@ -24,24 +25,22 @@ def _connect_unixsocket(self, address):
...
+            # a configuration change) unfortunately, we can't log that
+            # because...rsyslogd is down (and would just put us back down this
+            # code path)
+            # as a fallback, it makes the most sense to just write the
+            # messages to sys.stderr (which will end up in supervisord logs,
+            # and in containerized installs, cascaded down to pod logs)
+            # because the alternative is blocking the
+            # socket.send() in the Python process, which we definitely don't
+            # want to do)
+            sys.stderr.write(e)
```
ryanpetrello: I'm working on testing this, and trying to come up with a reproducible way to actually _hit_ `ConnectionRefusedError` or `BlockingIOError`, but so far I haven't been able to throw enough load at rsyslogd to get it to behave this way.

ryanpetrello: From https://rsyslog.readthedocs.io/en/latest/configuration/modules/imuxsock.html:

```Unix log sockets can be flow-controlled. That is, if processing queues fill up, the unix socket reader is blocked for a short while. This may be useful to prevent overruning the queues (which may cause excessive disk-io where it actually would not be needed).```

So this might be one way to accomplish it, but this behavior is *off* by default in rsyslogd.

Comment (ryanpetrello) [reply awx/main/utils/handlers.py]: I tested this by running this logger in a tight loop:

```python
for _ in range(100000): a.error('test')
```

After stopping my configured log server (i.e., my ELK stack) I **was** able to get rsyslogd to start blocking, but only if I broke the outbound logger and left it broken for about a minute:

image
image

ryanpetrello:
image

ryanpetrello:
Here's an example of this PR working w/ a _less_ verbose error/warning (in this case it was warning because rsyslogd got restarted, which unsets the socket file temporarily):

image

awx/main/utils/handlers.py
```diff
@@ -21,27 +22,31 @@ def _connect_unixsocket(self, address):
...
-        if not os.path.exists(settings.LOGGING['handlers']['external_logger']['address']):
-            return
-        try:
-            return super(RSysLogHandler, self).emit(msg)
-        except ConnectionRefusedError:
-            # rsyslogd has gone to lunch; this generally means that it's just
-            # been restarted (due to a configuration change)
-            # unfortunately, we can't log that because...rsyslogd is down (and
-            # would just us back ddown this code path)
-            pass
```
ryanpetrello, line 36: We'll never actually enter these two blocks, because the `emit()` call on L30 actually has its *own* exception handling that calls self.handleError()`.

awx/main/utils/handlers.py
```diff
@@ -21,27 +22,31 @@ def _connect_unixsocket(self, address):
...
+        # a configuration change) unfortunately, we can't log that
+        # because...rsyslogd is down (and would just put us back down this
+        # code path)
+        # as a fallback, it makes the most sense to just write the
+        # messages to sys.stderr (which will end up in supervisord logs,
+        # and in containerized installs, cascaded down to pod logs)
+        # because the alternative is blocking the
+        # socket.send() in the Python process, which we definitely don't
+        # want to do)
+        msg = f'{record.asctime} ERROR rsyslogd was unresponsive: '
```
ryanpetrello, line 37: By default, this function in cpython `logging.Handler.handleError()` prints a full traceback and is **extremely** noisy.  This version cuts the failure reporting down to 1-2 lines per emit.

ryanpetrello: https://github.com/python/cpython/blob/v3.8.5/Lib/logging/__init__.py#L987-L1032

awx/main/utils/handlers.py
```diff
@@ -21,27 +22,31 @@ def _connect_unixsocket(self, address):
         super(RSysLogHandler, self)._connect_unixsocket(address)
         self.socket.setblocking(False)
 
+    def handleError(self, record):
```
shanemcd, line 25: Just curious... why is this camel case?

ryanpetrello: Whoever wrote this module in Python liked camel-case 🤷 

https://github.com/python/cpython/blob/master/Lib/logging/__init__.py#L1000

camelCase isn't too uncommon in older parts of cpython's source code.

ryanpetrello: Better answer:  this Python module probably ~predates PEP 8 (cir. 2001), or at least, before PEP8 was widely implemented:

https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables
https://github.com/python/cpython/commit/57102f861d506b6c2d2215d100dac9143574fa77#diff-6b781e2294c88b6a18e9be5d536be2afb74f22593ec5be3de3bc3b507d955755R3

chrismeyersfsu:
Before this patch. When we were seeing blockingio tracebacks. Did those tracebacks actually get thrown up the stack or were they  just being caught and printed? If they were previously thrown up the stack, does this PR remove that? If it does eat the exception, I am in favor of that.

ryanpetrello:
@chrismeyersfsu 

before this PR: tracebacks caught via `logging.Handler.handleError` and printed (very verbosely): https://github.com/ansible/awx/pull/9452#issuecomment-789790161

this PR: tracebacks caught by **our** overridden `RSysLogHandler.handleError` and printed in less lines: https://github.com/ansible/awx/pull/9452#issuecomment-789792330

chrismeyersfsu (approved):

Issue #6528, opened 2020-04-01
WF viz needs error handling


dsesami:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Create a template where one of the nodes is a broken JT, missing an inventory that got deleted or something.

I made one that looks like this:
![Screenshot from 2020-04-01 16-26-30](https://user-images.githubusercontent.com/19942822/78183359-c5de7280-7435-11ea-99c5-e70bbed09dc5.png)

The launch button is disabled, but we aren't told why. Clicking save sends a DELETE and then a POST, which is a bit odd, but the first one 404s and the second one 400s. 

We should:

- [ ] Have error handling that indicates why the job won't launch.
- [ ] Indicate the broken node in the WF viz somehow. 
- [ ] Maybe think of a way to make it more obvious that you need to save the workflow before launching.

mabashian:
OK so, this issue was addressed by [this commit](https://github.com/ansible/awx/commit/ca1e597a4dec9f9ef70dbe7a198bca4601b0cd8d) of [this PR](https://github.com/ansible/awx/pull/8399).

We didn't seem to add a lot of unit test coverage though so I'm going to add some in a separate PR.

> Indicate the broken node in the WF viz somehow

This is being tracked by https://github.com/ansible/awx/issues/6805

> Create a template where one of the nodes is a broken JT, missing an inventory that got deleted or something.

Soon this will not be possible with https://github.com/ansible/awx/pull/9439 and https://github.com/ansible/awx/issues/8921

mabashian:
cc @unlikelyzero @tiagodread @one-t @akus062381 

Here's a PR with unit tests to cover the different scenarios where we want to show an error on workflow save:

https://github.com/ansible/awx/pull/9460

Feel free to add to those cases as you see fit.

Otherwise, I believe the development work here is done.

unlikelyzero:
@mabashian we were unable to reproduce the save errors identified in the unit tests. 

We will have to add an e2e test for:
Error shown when user selects JT which is deleted before Save action occurs. @akus062381 please reach out to me for an explanation

akus062381:
This issue has been fixed in the UI and the unit tests provide sufficient coverage.

PR #9460, opened 2021-03-02
Adds tests for workflow save error handling

mabashian:
##### SUMMARY
Also removes unnecessary code that was attempting to remove credentials from a new node.

This is what was left of #6528 

The actual error handling was added when we added support for workflow node prompting.

The way the tests work is that we set up the reducer to always force a specific state and then hit the Save button and expect an error to be shown.

### DIFF
--- a/awx/ui_next/src/screens/Template/WorkflowJobTemplateVisualizer/Visualizer.jsx
+++ b/awx/ui_next/src/screens/Template/WorkflowJobTemplateVisualizer/Visualizer.jsx
@@ -64,10 +64,10 @@ const getAggregatedCredentials = (
         templateDefaultCred.credential_type === overrideCred.credential_type
       ) {
         if (
-          (!templateDefaultCred.vault_id && !overrideCred.inputs.vault_id) ||
+          (!templateDefaultCred.vault_id && !overrideCred.inputs?.vault_id) ||
           (templateDefaultCred.vault_id &&
-            overrideCred.inputs.vault_id &&
-            templateDefaultCred.vault_id === overrideCred.inputs.vault_id)
+            overrideCred.inputs?.vault_id &&
+            templateDefaultCred.vault_id === overrideCred.inputs?.vault_id)
         ) {
           credentialHasOverride = true;
         }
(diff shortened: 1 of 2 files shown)


awx/ui_next/src/screens/Template/WorkflowJobTemplateVisualizer/Visualizer.jsx
```diff
@@ -405,16 +405,7 @@ function Visualizer({ template, i18n }) {
                   failure_nodes: [],
                   always_nodes: [],
                 };
-                if (node.promptValues?.removedCredentials?.length > 0) {
```
mabashian, line 408: When writing the tests, I realized this code was specific to editing a node and isn't needed in this block.  It was probably just copied over.  We only need to remove _overriding_ credentials.  Since this is a new node, there are no overriding credentials to remove (only overriding credentials to add)

jakemcdermott (approved):

unlikelyzero:
We just need the list of failure modes in the form of stubs. 3 have been identified in the unit tests

awx/ui_next/src/screens/Template/WorkflowJobTemplateVisualizer/Visualizer.test.jsx
```diff
@@ -219,6 +248,633 @@ describe('Visualizer', () => {
...
+    wrapper.update();
+    expect(
+      WorkflowJobTemplateNodesAPI.associateCredentials
+    ).toHaveBeenCalledTimes(1);
+    expect(
+      wrapper.find('AlertModal[title="Error saving the workflow!"]').length
+    ).toBe(1);
+  });
+
   test('Error shown to user when error thrown fetching workflow nodes', async () => {
```
unlikelyzero, line 878: @mabashian how can we reproduce this error?

unlikelyzero:
Errors:
- Error shown when saving fails due to node add error
- Error shown when saving fails due to node edit error
- Error shown when saving fails due to approval template add error
- Error shown when saving fails due to approval template edit error
- Error shown when saving fails due to node disassociate failure
- Error shown when saving fails due to node associate failure
- Error shown when saving fails due to credential disassociate failure
- Error shown when saving fails due to credential associate failure
- Error shown to user when error thrown fetching workflow nodes

Error not shown when selecting a JT which is missing resources
Error shown when user selects JT which is deleted before Save action occurs

PR #9461, opened 2021-03-02
Add subsystem metrics that propagate through Redis

fosterseth:
##### SUMMARY

#9019 -- list of metrics and their purpose / description
#9012
#9056
#8629

Use Redis to store metrics pertaining to the performance and health of subsystems such as the callback receiver and task manager. It is thread / multiprocess safe and should be fast enough to handle a high volume of data.

This data shows up at the `/api/v2/metrics` endpoint
You can filter down nodes using `/api/v2/metrics/?subsystemonly=1&node=awx-1`
You can also filter down to a specific metric,
`/api/v2/metrics/?subsystemonly=1&metrics=callback_receiver_events_insert_db_seconds&node=awx-1`

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 17.0.1
```

### DIFF
--- a/awx/api/views/metrics.py
+++ b/awx/api/views/metrics.py
@@ -14,6 +14,7 @@ from rest_framework.exceptions import PermissionDenied
 
 # AWX
 # from awx.main.analytics import collectors
+import awx.main.analytics.subsystem_metrics as s_metrics
 from awx.main.analytics.metrics import metrics
 from awx.api import renderers
 
@@ -33,5 +34,10 @@ class MetricsView(APIView):
     def get(self, request):
         ''' Show Metrics Details '''
         if request.user.is_superuser or request.user.is_system_auditor:
-            return Response(metrics().decode('UTF-8'))
+            metrics_to_show = ''
+            if not request.query_params.get('subsystemonly', "0") == "1":
+                metrics_to_show += metrics().decode('UTF-8')
+            if not request.query_params.get('dbonly', "0") == "1":
+                metrics_to_show += s_metrics.metrics(request)
+            return Response(metrics_to_show)
         raise PermissionDenied()
--- a/awx/main/analytics/analytics_tasks.py
+++ b/awx/main/analytics/analytics_tasks.py
@@ -1,0 +1,14 @@
+# Python
+import logging
+
+# AWX
+from awx.main.analytics.subsystem_metrics import Metrics
+from awx.main.dispatch.publish import task
+from awx.main.dispatch import get_local_queuename
+
+logger = logging.getLogger('awx.main.scheduler')
+
+
+@task(queue=get_local_queuename)
+def send_subsystem_metrics():
+    Metrics().send_metrics()
--- a/awx/main/analytics/subsystem_metrics.py
+++ b/awx/main/analytics/subsystem_metrics.py
@@ -1,0 +1,304 @@
+import redis
+import json
+import time
+import logging
+
+from django.conf import settings
+from django.apps import apps
+from awx.main.consumers import emit_channel_notification
+
+root_key = 'awx_metrics'
+logger = logging.getLogger('awx.main.wsbroadcast')
+
+
+class BaseM:
+    def __init__(self, field, help_text):
+        self.field = field
+        self.help_text = help_text
+        self.current_value = 0
+
+    def clear_value(self, conn):
+        conn.hset(root_key, self.field, 0)
+        self.current_value = 0
+
+    def inc(self, value):
+        self.current_value += value
+
+    def set(self, value):
+        self.current_value = value
+
+    def decode(self, conn):
+        value = conn.hget(root_key, self.field)
+        return self.decode_value(value)
+
+    def to_prometheus(self, instance_data):
+        output_text = f"# HELP {self.field} {self.help_text}\n# TYPE {self.field} gauge\n"
+        for instance in instance_data:
+            output_text += f'{self.field}{{node="{instance}"}} {instance_data[instance][self.field]}\n'
+        return output_text
+
+
+class FloatM(BaseM):
+    def decode_value(self, value):
+        if value is not None:
+            return float(value)
+        else:
+            return 0.0
+
+    def store_value(self, conn):
+        conn.hincrbyfloat(root_key, self.field, self.current_value)
+        self.current_value = 0
+
+
+class IntM(BaseM):
+    def decode_value(self, value):
+        if value is not None:
+            return int(value)
+        else:
+            return 0
+
+    def store_value(self, conn):
+        conn.hincrby(root_key, self.field, self.current_value)
+        self.current_value = 0
+
+
+class SetIntM(BaseM):
+    def decode_value(self, value):
+        if value is not None:
+            return int(value)
+        else:
+            return 0
+
+    def store_value(self, conn):
+        # do not set value if it has not changed since last time this was called
+        if self.current_value is not None:
+            conn.hset(root_key, self.field, self.current_value)
+            self.current_value = None
+
+
+class SetFloatM(SetIntM):
+    def decode_value(self, value):
+        if value is not None:
+            return float(value)
+        else:
+            return 0
+
+
+class HistogramM(BaseM):
+    def __init__(self, field, help_text, buckets):
+        self.buckets = buckets
+        self.buckets_to_keys = {}
+        for b in buckets:
+            self.buckets_to_keys[b] = IntM(field + '_' + str(b), '')
+        self.inf = IntM(field + '_inf', '')
+        self.sum = IntM(field + '_sum', '')
+        super(HistogramM, self).__init__(field, help_text)
+
+    def clear_value(self, conn):
+        conn.hset(root_key, self.field, 0)
+        self.inf.clear_value(conn)
+        self.sum.clear_value(conn)
+        for b in self.buckets_to_keys.values():
+            b.clear_value(conn)
+        super(HistogramM, self).clear_value(conn)
+
+    def observe(self, value):
+        for b in self.buckets:
+            if value <= b:
+                self.buckets_to_keys[b].inc(1)
+                break
+        self.sum.inc(value)
+        self.inf.inc(1)
+
+    def decode(self, conn):
+        values = {'counts': []}
+        for b in self.buckets_to_keys:
+            values['counts'].append(self.buckets_to_keys[b].decode(conn))
+        values['sum'] = self.sum.decode(conn)
+        values['inf'] = self.inf.decode(conn)
+        return values
+
+    def store_value(self, conn):
+        for b in self.buckets:
+            self.buckets_to_keys[b].store_value(conn)
+        self.sum.store_value(conn)
+        self.inf.store_value(conn)
+
+    def to_prometheus(self, instance_data):
+        output_text = f"# HELP {self.field} {self.help_text}\n# TYPE {self.field} histogram\n"
+        for instance in instance_data:
+            for i, b in enumerate(self.buckets):
+                output_text += f'{self.field}_bucket{{le="{b}",node="{instance}"}} {sum(instance_data[instance][self.field]["counts"][0:i+1])}\n'
+            output_text += f'{self.field}_bucket{{le="+Inf",node="{instance}"}} {instance_data[instance][self.field]["inf"]}\n'
+            output_text += f'{self.field}_count{{node="{instance}"}} {instance_data[instance][self.field]["inf"]}\n'
+            output_text += f'{self.field}_sum{{node="{instance}"}} {instance_data[instance][self.field]["sum"]}\n'
+        return output_text
+
+
+class Metrics:
+    def __init__(self, auto_pipe_execute=True):
+        self.pipe = redis.Redis.from_url(settings.BROKER_URL).pipeline()
+        self.conn = redis.Redis.from_url(settings.BROKER_URL)
+        self.last_pipe_execute = time.time()
+        # track if metrics have been modified since last saved to redis
+        # start with True so that we get an initial save to redis
+        self.metrics_have_changed = True
+        self.pipe_execute_interval = settings.SUBSYSTEM_METRICS_INTERVAL_SAVE_TO_REDIS
+        self.send_metrics_interval = settings.SUBSYSTEM_METRICS_INTERVAL_SEND_METRICS
+        # auto pipe execute will commit transaction of metric data to redis
+        # at a regular interval (pipe_execute_interval). If set to False,
+        # the calling function should call .pipe_execute() explicitly
+        self.auto_pipe_execute = auto_pipe_execute
+        Instance = apps.get_model('main', 'Instance')
+        self.instance_name = Instance.objects.me().hostname
+
+        # metric name, help_text
+        METRICSLIST = [
+            SetIntM('callback_receiver_events_queue_size_redis', 'Current number of events in redis queue'),
+            IntM('callback_receiver_events_popped_redis', 'Number of events popped from redis'),
+            IntM('callback_receiver_events_in_memory', 'Current number of events in memory (in transfer from redis to db)'),
+            IntM('callback_receiver_batch_events_errors', 'Number of times batch insertion failed'),
+            FloatM('callback_receiver_events_insert_db_seconds', 'Time spent saving events to database'),
+            IntM('callback_receiver_events_insert_db', 'Number of events batch inserted into database'),
+            HistogramM(
+                'callback_receiver_batch_events_insert_db', 'Number of events batch inserted into database', settings.SUBSYSTEM_METRICS_BATCH_INSERT_BUCKETS
+            ),
+            FloatM('subsystem_metrics_pipe_execute_seconds', 'Time spent saving metrics to redis'),
+            IntM('subsystem_metrics_pipe_execute_calls', 'Number of calls to pipe_execute'),
+            FloatM('subsystem_metrics_send_metrics_seconds', 'Time spent sending metrics to other nodes'),
+        ]
+        # turn metric list into dictionary with the metric name as a key
+        self.METRICS = {}
+        for m in METRICSLIST:
+            self.METRICS[m.field] = m
+
+        # track last time metrics were sent to other nodes
+        self.previous_send_metrics = SetFloatM('send_metrics_time', 'Timestamp of previous send_metrics call')
+
+    def clear_values(self):
+        for m in self.METRICS.values():
+            m.clear_value(self.conn)
+        self.metrics_have_changed = True
+        self.conn.delete(root_key + "_lock")
+
+    def inc(self, field, value):
+        if value != 0:
+            self.METRICS[field].inc(value)
+            self.metrics_have_changed = True
+            if self.auto_pipe_execute is True and self.should_pipe_execute() is True:
+                self.pipe_execute()
+
+    def set(self, field, value):
+        self.METRICS[field].set(value)
+        self.metrics_have_changed = True
+        if self.auto_pipe_execute is True and self.should_pipe_execute() is True:
+            self.pipe_execute()
+
+    def observe(self, field, value):
+        self.METRICS[field].observe(value)
+        self.metrics_have_changed = True
+        if self.auto_pipe_execute is True and self.should_pipe_execute() is True:
+            self.pipe_execute()
+
+    def serialize_local_metrics(self):
+        data = self.load_local_metrics()
+        return json.dumps(data)
+
+    def load_local_metrics(self):
+        # generate python dictionary of key values from metrics stored in redis
+        data = {}
+        for field in self.METRICS:
+            data[field] = self.METRICS[field].decode(self.conn)
+        return data
+
+    def store_metrics(self, data_json):
+        # called when receiving metrics from other instances
+        data = json.loads(data_json)
+        if self.instance_name != data['instance']:
+            logger.debug(f"{self.instance_name} received subsystem metrics from {data['instance']}")
+        self.conn.set(root_key + "_instance_" + data['instance'], data['metrics'])
+
+    def should_pipe_execute(self):
+        if self.metrics_have_changed is False:
+            return False
+        if time.time() - self.last_pipe_execute > self.pipe_execute_interval:
+            return True
+        else:
+            return False
+
+    def pipe_execute(self):
+        if self.metrics_have_changed is True:
+            duration_to_save = time.perf_counter()
+            for m in self.METRICS:
+                self.METRICS[m].store_value(self.pipe)
+            self.pipe.execute()
+            self.last_pipe_execute = time.time()
+            self.metrics_have_changed = False
+            duration_to_save = time.perf_counter() - duration_to_save
+            self.METRICS['subsystem_metrics_pipe_execute_seconds'].inc(duration_to_save)
+            self.METRICS['subsystem_metrics_pipe_execute_calls'].inc(1)
+
+            duration_to_save = time.perf_counter()
+            self.send_metrics()
+            duration_to_save = time.perf_counter() - duration_to_save
+            self.METRICS['subsystem_metrics_send_metrics_seconds'].inc(duration_to_save)
+
+    def send_metrics(self):
+        # more than one thread could be calling this at the same time, so should
+        # get acquire redis lock before sending metrics
+        lock = self.conn.lock(root_key + '_lock', thread_local=False)
+        if not lock.acquire(blocking=False):
+            return
+        try:
+            current_time = time.time()
+            if current_time - self.previous_send_metrics.decode(self.conn) > self.send_metrics_interval:
+                payload = {
+                    'instance': self.instance_name,
+                    'metrics': self.serialize_local_metrics(),
+                }
+                # store a local copy as well
+                self.store_metrics(json.dumps(payload))
+                emit_channel_notification("metrics", payload)
+                self.previous_send_metrics.set(current_time)
+                self.previous_send_metrics.store_value(self.conn)
+        finally:
+            lock.release()
+
+    def load_other_metrics(self, request):
+        # data received from other nodes are stored in their own keys
+        # e.g., awx_metrics_instance_awx-1, awx_metrics_instance_awx-2
+        # this method looks for keys with "_instance_" in the name and loads the data
+        # also filters data based on request query params
+        # if additional filtering is added, update metrics_view.md
+        instances_filter = request.query_params.getlist("node")
+        # get a sorted list of instance names
+        instance_names = [self.instance_name]
+        for m in self.conn.scan_iter(root_key + '_instance_*'):
+            instance_names.append(m.decode('UTF-8').split('_instance_')[1])
+        instance_names.sort()
+        # load data, including data from the this local instance
+        instance_data = {}
+        for instance in instance_names:
+            if len(instances_filter) == 0 or instance in instances_filter:
+                instance_data_from_redis = self.conn.get(root_key + '_instance_' + instance)
+                # data from other instances may not be available. That is OK.
+                if instance_data_from_redis:
+                    instance_data[instance] = json.loads(instance_data_from_redis.decode('UTF-8'))
+        return instance_data
+
+    def generate_metrics(self, request):
+        # takes the api request, filters, and generates prometheus data
+        # if additional filtering is added, update metrics_view.md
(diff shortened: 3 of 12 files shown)


fosterseth:
sample output

image

awx/main/dispatch/worker/base.py
```diff
@@ -58,6 +58,7 @@ def __init__(self, name, worker, queues=[], pool=None):
             self.pool = WorkerPool()
         self.pool.init_workers(self.worker.work_loop)
         self.redis = redis.Redis.from_url(settings.BROKER_URL)
+        self.redis.client_setname("awx_consumer_base")
```
ryanpetrello: Could we use `self.redis.client_setname(name)` instead?  This value would change depending on if it's the dispatcher or callback receiver.

awx/main/dispatch/worker/callback.py
```diff
@@ -48,24 +48,32 @@ def __init__(self):
         self.buff = {}
         self.pid = os.getpid()
         self.redis = redis.Redis.from_url(settings.BROKER_URL)
+        self.redis.client_setname("callbackbrokerworker")
```
ryanpetrello: ...and maybe this should match https://github.com/ansible/awx/pull/9461/files#r586609091 for the callback worker process?

awx/main/dispatch/worker/base.py
ryanpetrello: This is going to be either `"callback_receiver"` or `"dispatcher"` if `name` is used.

awx/main/dispatch/worker/callback.py
```diff
@@ -48,24 +48,32 @@ def __init__(self):
...
 
     def read(self, queue):
         try:
-            res = self.redis.blpop(settings.CALLBACK_QUEUE, timeout=1)
-            if res is None:
-                return {'event': 'FLUSH'}
-            self.total += 1
-            return json.loads(res[1])
+            with self.subsystem_metrics.conn.pipeline() as pipe:
+                res = self.redis.blpop(settings.CALLBACK_QUEUE, timeout=1)
```
ryanpetrello: This *could* get prohibitively expensive doing this per-read.  That said, I can't think of an alternative here that won't result in potentially inaccurate data.

I'd be curious to see what happens performance-wise here if we attempt to throw really high event load at this PR.

ryanpetrello: @chrismeyersfsu @fosterseth

Here's how I generally test this performance-wise for bottlenecks that can result in regressions.

First, I spun up a `devel` install, and stopped the callback receiver:

```shell
~ supervisorctl stop tower-processes:awx-receiver
```

Next, I launched some Job Template and noted the job ID (`1`), and I ran a Python script in `~ awx-manage shell_plus` to fill up the redis callback queue (this takes a minute or two):

```python
In [1]: from awx.main.queue import CallbackQueueDispatcher

In [2]: import uuid

In [3]: d = CallbackQueueDispatcher()

In [4]: for i in range(1000000):
   ...:     d.dispatch({ 'uuid': str(uuid.uuid4()), 'stdout': 'Line %d' % i, 'job_id': 1, })
   ...:
```

```shell
~ docker exec -it tools_redis_1 bash
root@0a52a640e1d2:/data# redis-cli -s /var/run/redis/redis.sock
redis /var/run/redis/redis.sock> llen callback_tasks
(integer) 1000000
```

Finally, I ran `~ awx-manage callback_stats` and restarted the callback receiver w/ `~ supervisorctl stop tower-processes:awx-receiver`, and I watched to see how high my throughput got.  On my local Docker-based install I got pretty far:

```shell
bash-4.4$ awx-manage callback_stats
main_jobevent
↳  last minute 156992
```

(so about 8 to 9M inserts per hour)

awx/main/queue.py
```diff
@@ -30,6 +30,9 @@ def __init__(self):
         self.queue = getattr(settings, 'CALLBACK_QUEUE', '')
         self.logger = logging.getLogger('awx.main.queue.CallbackQueueDispatcher')
         self.connection = redis.Redis.from_url(settings.BROKER_URL)
+        self.subsystem_metrics = s_metrics.Metrics(conn=self.connection)
+        self.connection.client_setname("callbackqueuedispatcher")
 
     def dispatch(self, obj):
         self.connection.rpush(self.queue, json.dumps(obj, cls=AnsibleJSONEncoder))
+        self.subsystem_metrics.inc('callback_receiver_events_insert_redis', 1)
```
ryanpetrello: In my testing, this per-event inc *really* seems to drastically slow down event insertion.

awx/main/dispatch/worker/callback.py
ryanpetrello: Next, I stopped all my containers, and re-ran this entire exercise using _this_ branch.  Unfortunately, these constant `incr()` calls really seem to _drastically_ affect write throughput:

```shell
bash-4.4$ awx-manage callback_stats
main_jobevent
↳  last minute 76796
```

(so about 4 to 5M inserts per hour, or about half the speed)

ryanpetrello: We would probably make up a lot of this performance loss by *not* doing the `incr()` calls per `.read()`, but instead at the end of each batch.

awx/main/queue.py
ryanpetrello: We might consider just getting rid of this metric; it seems really close to `llen(callback_tasks)`, and doing this additional `inc()` per-write is just plain expensive.

ryanpetrello (commented):
@fosterseth This PR is looking great so far.  If we can rethink some of our per-`rpush` and per-`blpop` operations in a way that doesn't cause us to lose much write throughput, then this will be perfect 😄

ryanpetrello:
image

This is awesome.

fosterseth:
sample of the JSON formatted data, to be consumed by UI

https://gist.github.com/fosterseth/b713c72384c6a3f7c5de2955d0e34a80

awx/api/views/metrics.py
```diff
@@ -37,6 +38,10 @@ class MetricsView(APIView):
     def get(self, request):
         ''' Show Metrics Details '''
         if (request.user.is_superuser or request.user.is_system_auditor):
-            return Response(metrics().decode('UTF-8'))
+            metrics_to_show = ''
+            if not request.query_params.get('subsystemonly', "0") == "1":
```
chrismeyersfsu, line 38: :+1: verified, with this filter the endpoint is fast.

chrismeyersfsu:
One thing I didn't thing about, but observed while testing. From an observational perspective (i.e. the /api/metrics endpoint), counter values can decrease. This can happen when the `/api/metrics` request is serviced by the node that has the most-up-to-date data vs. querying a node with "dated" data. Without some sort of census algorithm I don't know if we can do much about this. If a user cares about this they can subvert the aggregation and ask each node for their values.

### Steps to Recreate
* Run chatty_tasks.yml w/ `num_messages: 10000`
* Continuously refresh the metrics in your browser http://localhost:8013/api/v2/metrics/?subsystemonly=1
* watch `callback_receiver_events_insert_db_seconds` 

### Expected
* for `callback_receiver_events_insert_db_seconds` to never go down

### Actual
* Sometimes `callback_receiver_events_insert_db_seconds` will be less than previous observations.

chrismeyersfsu:
```
callback_receiver_batch_events_insert_db_bucket{le="10",node="awx_1"} 712
callback_receiver_batch_events_insert_db_bucket{le="50",node="awx_1"} 1119
callback_receiver_batch_events_insert_db_bucket{le="150",node="awx_1"} 1119
callback_receiver_batch_events_insert_db_bucket{le="350",node="awx_1"} 1119
callback_receiver_batch_events_insert_db_bucket{le="650",node="awx_1"} 1119
callback_receiver_batch_events_insert_db_bucket{le="1000",node="awx_1"} 1119
callback_receiver_batch_events_insert_db_bucket{le="+Inf",node="awx_1"} 1119
callback_receiver_batch_events_insert_db_count{node="awx_1"} 1119
callback_receiver_batch_events_insert_db_sum{node="awx_1"} 10148
```
Trying to wrap my head around the above.

```
callback_receiver_events_in_memory{node="awx_1"} 52
callback_receiver_events_in_memory{node="awx_2"} 0
callback_receiver_events_in_memory{node="awx_3"} 0
```
My system is idle now and there are events in memory. I'll dig and see if I can see why it's not reset to 0.

fosterseth:
Here are some changes I did to improve performance

1. `inc` and `set` no longer immediately write to redis. Instead these methods update and maintain variables in the Metrics() object (so in memory only, which should be fast)
2. Every so often, we need to commit these to redis. By default the Metrics() class will do so automatically every 2 seconds. It does this by opening a redis pipe (a db transaction), looping through each metric and incrementing the right redis key. We can also have more fined-grained control of when these transactions take place, which I do in callback.py with the `pipe_execute` command.

awx/main/analytics/subsystem_metrics.py
```diff
@@ -0,0 +1,246 @@
...
+                 'Current number of events in memory (in transfer from redis to db)'),
+            IntM('callback_receiver_batch_events_errors',
+                 'Number of times batch insertion failed'),
+            FloatM('callback_receiver_events_insert_db_seconds',
+                   'Time spent saving events to database'),
+            IntM('callback_receiver_events_insert_db',
+                 'Number of events batch inserted into database'),
+            HistogramM('callback_receiver_batch_events_insert_db',
+                       'Number of events batch inserted into database',
+                       [10, 50, 150, 350, 650, 1000]),
```
fosterseth: should this be a `setting`? i.e. batch inserts might vary depending on environment

awx/main/analytics/analytics_tasks.py
```diff
@@ -0,0 +1,16 @@
...
+from awx.main.dispatch.publish import task
+from awx.main.dispatch import get_local_queuename
+
+logger = logging.getLogger('awx.main.scheduler')
+
+
+@task(queue=get_local_queuename)
+def send_subsystem_metrics():
+    logger.debug("Sending subsystem metrics.")
+    Metrics().send_metrics()
```
chrismeyersfsu, line 14: I like that the local node stats live in a different key and that this part basically creates a copy on the local node into a different hash & sends to remove nodes.

chrismeyersfsu:
Verified performance of devel vs. Seth's implicit pipeline using timeout vs. explicit timeout PR that I have submitted.

## Experiment

We want to ensure the metrics feature doesn't effect the callback receiver performance.

* chatty_tasks.yml `10` hosts `10` forks `10,000` events resulting in =~ `100,000` events
* stop callback receiver, run [chatty_tasks.yml](https://github.com/ansible/test-playbooks/blob/master/chatty_tasks.yml) w/ `count=100000`
* start callback receiver, process `100,000` events
* Query the DB to ask events processed per minute in the past hour

### devel

Baseline.

```
SELECT * FROM (
SELECT to_char(start_time, 'YYYY-MM-DD HH24:MI'), count(e.modified) AS events, count(e.modified) / 1 as events_per_minute
FROM   generate_series(localtimestamp - interval '1 hours'
                     , localtimestamp
                     , interval '1 min') g(start_time)
LEFT   JOIN main_jobevent e ON e.modified >= g.start_time
                   AND e.modified <  g.start_time + interval '1 min'
GROUP  BY start_time
ORDER  BY start_time ) as s WHERE events != 0;
```

```
 2021-03-12 21:21 |  18321 |             18321
 2021-03-12 21:22 |  77112 |             77112
 2021-03-12 21:23 |   4591 |              4591
```

So about **77k** per minute in the busiest minute.



### feat_metrics_via_redis

Devel + Seth's metrics changes.



Per 30-seconds

```
SELECT * FROM (
SELECT to_char(start_time, 'YYYY-MM-DD HH24:MI'), count(e.modified) AS events, count(e.modified) / 1 as events_per_10_seconds
FROM   generate_series(localtimestamp - interval '1 hours'
                     , localtimestamp
                     , interval '30 seconds') g(start_time)
LEFT   JOIN main_jobevent e ON e.modified >= g.start_time
                   AND e.modified <  g.start_time + interval '30 seconds'
GROUP  BY start_time
ORDER  BY start_time ) as s WHERE events != 0;
```

```
 2021-03-12 22:00 |  19888 |                 19888
 2021-03-12 22:01 |  38515 |                 38515
 2021-03-12 22:01 |  35817 |                 35817
 2021-03-12 22:02 |   5804 |                  5804
```

38515+35817 =~ **74k** per minute

```
callback_receiver_events_queue_size_redis{node="awx_1"} 0
# HELP callback_receiver_events_popped_redis Number of events popped from redis
# TYPE callback_receiver_events_popped_redis gauge
callback_receiver_events_popped_redis{node="awx_1"} 110110
# HELP callback_receiver_events_in_memory Current number of events in memory (in transfer from redis to db)
# TYPE callback_receiver_events_in_memory gauge
callback_receiver_events_in_memory{node="awx_1"} 0
# HELP callback_receiver_batch_events_errors Number of times batch insertion failed
# TYPE callback_receiver_batch_events_errors gauge
callback_receiver_batch_events_errors{node="awx_1"} 0
# HELP callback_receiver_events_insert_db_seconds Time spent saving events to database
# TYPE callback_receiver_events_insert_db_seconds gauge
callback_receiver_events_insert_db_seconds{node="awx_1"} 69.368788
# HELP callback_receiver_events_insert_db Number of events batch inserted into database
# TYPE callback_receiver_events_insert_db gauge
callback_receiver_events_insert_db{node="awx_1"} 110105
# HELP callback_receiver_batch_events_insert_db Number of events batch inserted into database
# TYPE callback_receiver_batch_events_insert_db histogram
callback_receiver_batch_events_insert_db_bucket{le="10",node="awx_1"} 1370
callback_receiver_batch_events_insert_db_bucket{le="50",node="awx_1"} 1696
callback_receiver_batch_events_insert_db_bucket{le="150",node="awx_1"} 1707
callback_receiver_batch_events_insert_db_bucket{le="350",node="awx_1"} 1946
callback_receiver_batch_events_insert_db_bucket{le="650",node="awx_1"} 2024
callback_receiver_batch_events_insert_db_bucket{le="2000",node="awx_1"} 2024
callback_receiver_batch_events_insert_db_bucket{le="+Inf",node="awx_1"} 2024
callback_receiver_batch_events_insert_db_count{node="awx_1"} 2024
callback_receiver_batch_events_insert_db_sum{node="awx_1"} 110105
```

### Pipeline Metrics

Explicit pipeline calls.

per-20 seconds

```
SELECT * FROM (
SELECT to_char(start_time, 'YYYY-MM-DD HH24:MI'), count(e.modified) AS events, count(e.modified) / 1 as events_per_10_seconds
FROM   generate_series(localtimestamp - interval '1 hours'
                     , localtimestamp
                     , interval '20 seconds') g(start_time)
LEFT   JOIN main_jobevent e ON e.modified >= g.start_time
                   AND e.modified <  g.start_time + interval '20 seconds'
GROUP  BY start_time
ORDER  BY start_time ) as s WHERE events != 0;
```



```
 2021-03-13 18:10 |  28668 |                 28668
 2021-03-13 18:11 |  29365 |                 29365
 2021-03-13 18:11 |  30812 |                 30812
```

28668+29365+30812 =~ **88k** per minute



```
# HELP callback_receiver_events_queue_size_redis Current number of events in redis queue
# TYPE callback_receiver_events_queue_size_redis gauge
callback_receiver_events_queue_size_redis{node="awx_1"} 0
# HELP callback_receiver_events_inserted_redis Number of playbook events inserted into redis
# TYPE callback_receiver_events_inserted_redis gauge
callback_receiver_events_inserted_redis{node="awx_1"} 100105
# HELP callback_receiver_events_popped_redis Number of events popped from redis
# TYPE callback_receiver_events_popped_redis gauge
callback_receiver_events_popped_redis{node="awx_1"} 100105
# HELP callback_receiver_events_in_memory Current number of events in memory (in transfer from redis to db)
# TYPE callback_receiver_events_in_memory gauge
callback_receiver_events_in_memory{node="awx_1"} 0
# HELP callback_receiver_batch_events_errors Number of times batch insertion failed
# TYPE callback_receiver_batch_events_errors gauge
callback_receiver_batch_events_errors{node="awx_1"} 0
# HELP callback_receiver_events_insert_db_seconds Time spent saving events to database
# TYPE callback_receiver_events_insert_db_seconds gauge
callback_receiver_events_insert_db_seconds{node="awx_1"} 37.328108
# HELP callback_receiver_events_insert_db Number of events batch inserted into database
# TYPE callback_receiver_events_insert_db gauge
callback_receiver_events_insert_db{node="awx_1"} 100102
# HELP callback_receiver_batch_events_insert_db Number of events batch inserted into database
# TYPE callback_receiver_batch_events_insert_db histogram
callback_receiver_batch_events_insert_db_bucket{le="10",node="awx_1"} 4
callback_receiver_batch_events_insert_db_bucket{le="50",node="awx_1"} 6
callback_receiver_batch_events_insert_db_bucket{le="150",node="awx_1"} 267
callback_receiver_batch_events_insert_db_bucket{le="350",node="awx_1"} 647
callback_receiver_batch_events_insert_db_bucket{le="650",node="awx_1"} 647
callback_receiver_batch_events_insert_db_bucket{le="2000",node="awx_1"} 647
callback_receiver_batch_events_insert_db_bucket{le="+Inf",node="awx_1"} 647
callback_receiver_batch_events_insert_db_count{node="awx_1"} 647
callback_receiver_batch_events_insert_db_sum{node="awx_1"} 100102
```

Zokormazo:
I've been playing a bit with this... just a question (and I'm not sure if this is related to this PR):

Almost all the metrics appear with `node=` tags for all the nodes on the cluster.

But I noticed this ones:

```
awx_instance_launch_type_total{launch_type="dependency",node="node01"} 2.0
awx_instance_launch_type_total{launch_type="manual",node="node01"} 2.0
# HELP awx_instance_status_total Status of Job launched
# TYPE awx_instance_status_total gauge
awx_instance_status_total{node="node01",status="running"} 1.0
awx_instance_status_total{node="node01",status="successful"} 4.0
```

And the other nodes? Are not appearing because their value is 0? Will prometheus like that?

fosterseth:
@chrismeyersfsu @ryanpetrello 

testing setup
- push 100,000 events into redis
- then enable and profile callback receiver
- I found that counting number of events inserted into db every 20 seconds was very unreliable for some reason. I would get drastically different results, run to run. I think the reason is that the flush() method is mostly unchanged, and this is where the bulk of the code spends its time. Instead, I think the better method is to profile directly, so we can see how changing the read() method affects performance.

### devel
image

`read()` time **36 seconds**

### my branch
(omitting flame graph, looks the same as above)

`read()` time **36 seconds**

### cmeyers pipeline change
image

`read()` time **81 seconds**

I really liked the idea of combining redis pop operation and the metrics writing operations into the same pipeline. However I'm finding there is non-negligible overhead with doing a pipeline operations in general. For example, doing just doing a single redis pop in a pipeline nearly doubles the execution time, compared to without using a pipeline.

here is a script that should simulate what is happening in `read` with cmeyer's pipeline PR
```python
import time
import redis
e = "======================================= this is an event ================================================================"
connection = redis.Redis.from_url(settings.BROKER_URL)
connection_pipe = redis.Redis.from_url(settings.BROKER_URL).pipeline()
connection.flushdb()
for _ in range(100000):
    connection.rpush("sbf_test", e)
x = time.time()
for _ in range(100000):
    connection_pipe.blpop("sbf_test")
    connection_pipe.llen("sbf_test")
    for _ in range(10):
        connection_pipe.incr("sbf_inc", 1)
    connection_pipe.execute()
print("pipe pop, len, inc: ", time.time() - x)
connection.flushdb()
for _ in range(100000):
    connection.rpush("sbf_test", e)
x = time.time()
for _ in range(100000):
    connection.blpop("sbf_test")
print("no pipe just pop: ", time.time() - x)
```
```
pipe pop, len, inc:  31.561394214630127
no pipe just pop:  5.687417030334473
```

## What effect does subsystem metrics have on the callback receiver?
image
For processing 100,000 events, subsystem metrics only took 0.26 seconds (0.06% of the total execution time)
So, very little overhead

## My alternative way to fix the negative memory stats
Chris found a bug where occasionally there would be negative value for # of events in memory. I've fixed it here. Now the increment operation occurs as soon as we pop from redis, instead of being delayed for 2 seconds. That way, if during this delay, the decrement operation occurs, the total value will still be accurate. 

https://github.com/ansible/awx/commit/6d5fa1eb4ea666b90d003e5efbbfb5ffd7bea1c0#diff-915de8c0ee78894addb6ce93fd72bda6adc7b9a0bf2ba54ae9e9f1a6c0d8f011L64

awx/settings/defaults.py
```diff
@@ -467,6 +470,10 @@ def IS_TESTING(argv=None):
         'schedule': timedelta(seconds=60),
         'options': {'expires': 50,}
     },
+    'send_subsystem_metrics': {
+        'task': 'awx.main.analytics.analytics_tasks.send_subsystem_metrics',
+        'schedule': timedelta(seconds=3),
```
fosterseth: @ryanpetrello this feels unconventional. is using the dispatcher scheduler overkill for a simple 'run this command every 3 seconds" sort of task?

ryanpetrello: Yea, this is probably the wrong way to do this - if the queue gets backed up, this is probably gonna cause it to grow needlessly and run the same thing over and over (you wouldn't want to queue up 100 of these and then do the work 100 times).  Generally in the past, we've used this mechanism for things we want to run every ~30-60+ seconds.

ryanpetrello: The overhead of beat --> message bus --> fork a dispatcher process --> ship data is pretty high for something we'll be doing very regularly.  If we want to do this at this interval, perhaps it could be something that happens as part of the callback process (maybe we could call this function on every `.read()`, but skip it if N seconds have elapsed?)

cc @chrismeyersfsu for other ideas here

another possibility is treating this as some additional persistent process that just runs and sleeps in a loop?

fosterseth:
@ryanpetrello @chrismeyersfsu 

Instead of sending metrics to other nodes via dispatcher, `send_metrics()` is being evoked each time `pipe_execute()` is called, which should occur once every 2-3 seconds.

`pipe_execute()` is called from a variety of threads, but we don't want to send metrics in each of these threads, so we can check a timestamp that is stored in redis (that is shared with all threads) called "send_metrics_time". If it has been > 2 seconds since this last call, then send metrics.

Another optimization is that if data hasn't changed since the last save to redis, we skip the save. That way an idle system isn't doing any unnecessary work.

I've also added two new metrics to capture data about the subsystem_metrics itself (the last two metrics in this image)

image

So the `pipe_execute()` only took about 0.75 seconds to process 400,000 events over 4.5 minutes

ryanpetrello (approved):
I tried this latest branch out with a large amount of queue back pressure, and I'm seeing really great write performance/throughput.

To test the callback worker throughput, I stopped the callback workers, filled the `callback_tasks` in redis:

image

```
root@025b34e32408:/data# redis-cli -s /var/run/redis/redis.sock
redis /var/run/redis/redis.sock> llen callback_tasks
(integer) 1000000
```

...and then restarted the workers and watched the metrics in the API and CLI:

```
bash-4.4$ awx-manage callback_stats
main_jobevent
↳  last minute 144572

main_inventoryupdateevent
↳  last minute 0

main_projectupdateevent
↳  last minute 0

main_adhoccommandevent
↳  last minute 0
```

```shell
~ curl -sk "https://ryan:ryan@yossarian:8043/api/v2/metrics/?subsystemonly=1" | ack "insert_db\{"
callback_receiver_events_insert_db{node="awx_1"} 460875
```

I let this run for ~6-7 minutes and it finished processing all the events 😄.

image

I want to make sure other API reviewers have time to weigh in on this PR, but my testing of this looks like it works pretty well (and it's performant, too!)

Thanks for your hard work on this one, @fosterseth - great job 😄

awx/main/analytics/subsystem_metrics.py
```diff
@@ -0,0 +1,306 @@
...
+
+    def store_metrics(self, data_json):
+        # called when receiving metrics from other instances
+        data = json.loads(data_json)
+        if self.instance_name != data['instance']:
+            logger.debug(f"{self.instance_name} received subsystem metrics from {data['instance']}")
+        self.conn.set(root_key + "_instance_" + data['instance'], data['metrics'])
+
+    def should_pipe_execute(self):
+        if time.time() - self.last_pipe_execute > self.pipe_execute_interval:
```
chrismeyersfsu, line 224: I don't love using a timeout as a solution to reduce the number of metrics request to redis. We can have no loss in performance if we piggyback with pipelining onto the redis callback receiver pop()

chrismeyersfsu: Maybe this is OK. Since these are atomic incr and dec operations.

I'm not sure if this pattern for pipelining will work well when used in a process that isn't a long-lived one that continuously loops i.e. the task manager. In that case, we want to use pipelining to form a single transaction and "commit" it at the end of each task manager run; but we can cross that bridge when we get to it.

awx/main/dispatch/worker/callback.py
```diff
@@ -66,8 +72,19 @@ def read(self, queue):
...
         finally:
             self.record_statistics()
+            self.record_read_metrics()
+
         return {'event': 'FLUSH'}
 
+    def record_read_metrics(self):
+        if self.queue_pop == 0:
+            return
+        if self.subsystem_metrics.should_pipe_execute() is True:
```
chrismeyersfsu, line 80: I think there have been some changes to the code since I last looked. Before, I was observing negative stats for the events in memory statistic. I think it can still happen here if the decrement operation below always "takes" the pipeline execution for the time window. The addition here could get starved while the system is busy, resulting in a negative number.

awx/main/analytics/subsystem_metrics.py
fosterseth: @chrismeyersfsu
> We can have no loss in performance if we piggyback with pipelining onto the redis callback receiver pop()

I found performance loss when introducing pipelining on every `read()` loop. I outlined the details here, but I think it got lost in the "hidden" section of this PR
https://github.com/ansible/awx/pull/9461#issuecomment-800580082

you can run this code snippet in awx-manage shell to get an idea of the overhead

```python
import time
import redis
e = "======================================= this is an event ================================================================"
connection = redis.Redis.from_url(settings.BROKER_URL)
connection_pipe = redis.Redis.from_url(settings.BROKER_URL).pipeline()
connection.flushdb()
for _ in range(100000):
    connection.rpush("sbf_test", e)
x = time.time()
for _ in range(100000):
    connection_pipe.blpop("sbf_test")
    connection_pipe.llen("sbf_test")
    for _ in range(10):
        connection_pipe.incr("sbf_inc", 1)
    connection_pipe.execute()
print("pipe pop, len, inc: ", time.time() - x)
connection.flushdb()
for _ in range(100000):
    connection.rpush("sbf_test", e)
x = time.time()
for _ in range(100000):
    connection.blpop("sbf_test")
print("no pipe just pop: ", time.time() - x)
connection.flushdb()
for _ in range(100000):
    connection.rpush("sbf_test", e)
x = time.time()
for _ in range(100000):
    connection_pipe.blpop("sbf_test")
    connection_pipe.execute()
print("pipe just pop: ", time.time() - x)
```

```
pipe pop, len, inc:  19.783135890960693
no pipe just pop:  4.3485107421875
pipe just pop:  6.442138433456421
```

I do agree that for task manager we'll want a way to do a single transaction at the end. Currently in this PR, `inc` and `set` only modify object memory values -- they are not doing pipeline operations. So in the task manager we could just accumulate the values in a Metrics() object, then do a `pipe_execute()` whenever we wish at the end.

awx/main/dispatch/worker/callback.py
fosterseth: from my understanding, "read" is always called before "flush" since it runs in the same thread/loop. Therefore, the decrement in "flush" will always occur after the increment in "read". The reason we were getting negative values before was that we were delaying the increment operation in read. And during that delay, it was possible to decrement in flush.

chrismeyersfsu: Yes, but we are short circuiting the increment in "read" by checking if `should_pipe_execute`, which will return `False` because the pipe in decrement is taking the window.

chrismeyersfsu (approved):

kdelee (approved):
This is good as far as I'm concerned. 

@Zokormazo do you want to do any further verification?

Issue #9479, opened 2021-03-04
Job Template loads twice if autopopulate is triggered


nixocio:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Job Template loads twice if the project auto populate is triggered. 

Right now, if there is just one project available - the auto populate will be triggered on the the Job Template Add screen, and the page will load twice.

This issue is being sorted out.

unlikelyzero:
Note: this is showing up in the visual tests

tiagodread:
Verified manually and all e2e tests are green

PR #9488, opened 2021-03-04
Fix extra re-render for Job Template

nixocio:
Fix extra re-render for Job Template.

Also, update a few unit-tests.

See: https://github.com/ansible/awx/issues/9479

### DIFF
--- a/awx/ui_next/src/components/Lookup/ExecutionEnvironmentLookup.jsx
+++ b/awx/ui_next/src/components/Lookup/ExecutionEnvironmentLookup.jsx
@@ -1,11 +1,11 @@
 import React, { useCallback, useEffect } from 'react';
-import { string, func, bool } from 'prop-types';
+import { string, func, bool, oneOfType, number } from 'prop-types';
 import { useLocation } from 'react-router-dom';
 import { withI18n } from '@lingui/react';
 import { t } from '@lingui/macro';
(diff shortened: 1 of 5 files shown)


nixocio:
Currently running e2e tests. I ran the tests running github actions and got 🟢

awx/ui_next/src/components/Lookup/ExecutionEnvironmentLookup.jsx
```diff
@@ -26,15 +26,38 @@ function ExecutionEnvironmentLookup({
...
   organizationId,
   popoverContent,
+  projectId,
   tooltip,
   value,
-  onBlur,
 }) {
   const location = useLocation();
 
+  const {
```
jakemcdermott, line 39: I guess we needed to move it into the lookup component - that makes sense.

nixocio: I originally had this change like that, but I tried to reduce the number of props, since we want to filter using the organization id. Attempting to fetch the project before, and just pass the related org, but that generated extra requests/re-renders. Not terrible, but now the component has an extra prop related to the project.

jakemcdermott (approved):

tiagodread (approved):

PR #9498, opened 2021-03-05
remove custom_virtualenv support from the AWX collection and docs

ryanpetrello:
cc @rebeccahhh @shanemcd @jbradberry 

Until we address tooling for identification/migration of these (later: https://github.com/ansible/awx/issues/7062), I thought it might make sense to remove this from our docs and collection to discourage people from using these (since the API still technically allows you to set the field).


shanemcd (approved):
🙌 thanks!

Issue #7603, opened 2020-07-10
Enable html in custom login text


jakemcdermott:
### summary
see: https://github.com/ansible/awx/pull/7585 --> https://github.com/ansible/awx/issues/7600

- Could try to use `__dangerouslySetInnerHTML` along with `html-entities` to sanitize (like we do for job events: https://github.com/ansible/awx/blob/devel/awx/ui_next/src/screens/Job/JobOutput/JobEvent.jsx#L55)

- If that doesn't work we'd probably need to first put in a request to pf-react to explicitly enable passing a `React.ReactNode` to the LoginPage component for the `textContent` prop like they do for the `helperText` prop: https://patternfly-react.surge.sh/documentation/react/components/loginpage

mabashian:
@jakemcdermott bumping this up to high as it seems like something we supported in the old UI.  If you think differently feel free to readjust.

tiagodread:
@one-t can we close this?

PR #9519, opened 2021-03-08
Adds support for html in custom login text

mabashian:
##### SUMMARY
link #7603 

I couldn't come up with a way to do this without breaking up the component and discontinuing use of the LoginPage PF component.  This is because LoginPage expects the `textContent` component (what we use to display the custom login text) to be a string.  By using the underlying LoginPage components I reconstructed the login page and got more control over that prop.

The custom message in the old UI supported both strings and HTML:

Screen Shot 2021-03-08 at 1 25 31 PM

So we need to support rendering HTML but we need to do it in a safe way.  Our solution to that was https://docs.angularjs.org/api/ngSanitize.  React doesn't seem to have anything like this built in so I went looking for outside help.  html-entities is already included in our project but as best as I can tell that lib is mainly focused on swapping special characters out for html entities.  I wanted something that was going to strip the HTML of bits that could be exploited by a malicious actor.

I settled on https://www.npmjs.com/package/sanitize-html because it was a) small and b) actively maintained.  The API was simple and let me sanitize the HTML before setting it using `dangerouslySetInnerHTML`.  If we need to tweak the configuration away from the default values then we can certainly do that.

Screen Shot 2021-03-08 at 1 37 07 PM

Screen Shot 2021-03-08 at 1 37 31 PM

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,6 @@
 
 This is a list of high-level changes for each release of AWX. A full list of commits can be found at `https://github.com/ansible/awx/releases/tag/<version>`.
 
-# TBD
-- Added support for html in custom login text https://github.com/ansible/awx/pull/9519
-
 # (March 19, 2021)
 
 **IMPORTANT INSTALL AND UPGRADE NOTES**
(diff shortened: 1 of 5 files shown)


mabashian:
@jakemcdermott curious to see what you think of this approach.  You had mentioned `html-entities` in the original issue and I want to make sure I'm not missing a possible solution that involves just that library.

jakemcdermott (approved):

one-t:
just adding here, allowable tags (according to defaults for sanitizeHTML are
```
allowedTags: [
  "address", "article", "aside", "footer", "header", "h1", "h2", "h3", "h4",
  "h5", "h6", "hgroup", "main", "nav", "section", "blockquote", "dd", "div",
  "dl", "dt", "figcaption", "figure", "hr", "li", "main", "ol", "p", "pre",
  "ul", "a", "abbr", "b", "bdi", "bdo", "br", "cite", "code", "data", "dfn",
  "em", "i", "kbd", "mark", "q", "rb", "rp", "rt", "rtc", "ruby", "s", "samp",
  "small", "span", "strong", "sub", "sup", "time", "u", "var", "wbr", "caption",
  "col", "colgroup", "table", "tbody", "td", "tfoot", "th", "thead", "tr"
],
```

one-t:
Hey, @mabashian - can we add `ouia-id`s to the logo and footer elements?

PR #9533, opened 2021-03-09
Reduce the log level for some of the more spammy sources

jbradberry:
##### SUMMARY
The log levels for certain things cause the docker terminal logs to be filled with lots of messages about the scheduler and task workers, potentially pushing out of view error log messages that the developer should pay attention to.

Also, let's put in the setting change that allows Django Debug Toolbar to work properly.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API
 - Installer

##### AWX VERSION
```
awx: 17.0.1
```

### DIFF
--- a/awx/settings/development.py
+++ b/awx/settings/development.py
@@ -53,6 +53,10 @@ LOGGING['loggers']['awx.isolated.manager.playbooks']['propagate'] = True  # noqa
 
 # celery is annoyingly loud when docker containers start
 LOGGING['loggers'].pop('celery', None)  # noqa
+# avoid awx.main.dispatch WARNING-level scaling worker up/down messages
+LOGGING['loggers']['awx.main.dispatch']['level'] = 'ERROR'  # noqa
+# suppress the spamminess of the awx.main.scheduler and .tasks loggers
+LOGGING['loggers']['awx']['level'] = 'INFO'  # noqa
 
 ALLOWED_HOSTS = ['*']
 
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -5,13 +5,13 @@ Django Debug Toolbar (DDT)
 ----------------
 This is a useful tool for examining SQL queries, performance, headers, requests, signals, cache, logging, and more.  
 
-To enable DDT, you need to set your `INTERNAL_IPS` to the IP address of your load balancer.  This can be overriden in `local_settings`.  
+To enable DDT, you need to set your `INTERNAL_IPS` to the IP address of your load balancer.  This can be overridden by creating a new settings file beginning with `local_` in `awx/settings/` (e.g. `local_overrides.py`).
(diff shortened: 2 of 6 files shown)


awx/settings/development.py
```diff
@@ -53,7 +53,13 @@
 
 # celery is annoyingly loud when docker containers start
 LOGGING['loggers'].pop('celery', None)  # noqa
+# avoid awx.main.dispatch WARNING-level scaling worker up/down messages
+LOGGING['loggers']['awx.main.dispatch']['level'] = 'ERROR'  # noqa
+# suppress the spamminess of the awx.main.scheduler and .tasks loggers
+LOGGING['loggers']['awx']['level'] = 'INFO'  # noqa
 
+# allow Django Debug Toolbar to work in the local dev environment
+INTERNAL_IPS = ('127.0.0.1', '172.19.0.1', '172.18.0.1', '192.168.100.1')
```
rooftopcellist: Do we really want to have DDT enabled by default?  iirc, it adds a fair bit of overhead to requests.

jbradberry: I obviously strongly think so, but let's see if more people are opposed.

jbradberry: All right, pruning it out based on @matburt 's very strong opinion on the matter.

docs/debugging.md
```diff
@@ -5,13 +5,13 @@ Django Debug Toolbar (DDT)
 ----------------
 This is a useful tool for examining SQL queries, performance, headers, requests, signals, cache, logging, and more.  
 
-To enable DDT, you need to set your `INTERNAL_IPS` to the IP address of your load balancer.  This can be overriden in `local_settings`.  
+To enable DDT, you need to set your `INTERNAL_IPS` to the IP address of your load balancer.  This can be overriden by creating a new settings file beginning with `local_` in `awx/settings/` (e.g. `local_overrides.py`).
```
rooftopcellist: nitpick, `overridden` instead of `overriden`.

rooftopcellist (approved):

docs/debugging.md
jbradberry: fine fine, even though that wasn't me...

jbradberry: updated.

ryanpetrello (approved):

awx/settings/development.py
```diff
@@ -53,6 +53,10 @@
 
 # celery is annoyingly loud when docker containers start
 LOGGING['loggers'].pop('celery', None)  # noqa
+# avoid awx.main.dispatch WARNING-level scaling worker up/down messages
+LOGGING['loggers']['awx.main.dispatch']['level'] = 'ERROR'  # noqa
```
AlanCoding, line 57: This has a pretty severely silencing effect on the dispatcher logs, and I'm interested to see if I can return to it to restore those messages.

PR #9535, opened 2021-03-09
combine database.py and local_settings.py to fix settings import orde…

rooftopcellist:
…r issue

##### SUMMARY
@jbradberry noticed that the unit & functional tests are failing because they are being inadvertently pointed at the postgres db instead of sqlite.  

This is because of an import order issue cause by the introduction of `/etc/tower/conf.d/database.py` from the sources role.  

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - Installer

##### AWX VERSION

```
devel
```


tools/docker-compose/ansible/roles/sources/tasks/main.yml
```diff
@@ -42,7 +42,7 @@
     dest: "{{ sources_dest }}/{{ item }}"
     mode: '0600'
   with_items:
-    - "database.py"
+    - "local_settings.py"
```
jbradberry, line 45: I mentioned this elsewhere, but I'm not a fan of naming this setting file `local_settings.py`.  It has a conventional meaning for Django projects that I don't think this is matching.

tools/docker-compose/ansible/roles/sources/templates/local_settings.py.j2
```diff
@@ -20,6 +20,19 @@
 # INTERNAL_IPS = ('172.19.0.1', '172.18.0.1', '192.168.100.1')
 # ALLOWED_HOSTS = ['*']
 
+
```
jbradberry, line 23: Along those same lines, this file still has a lot of things in it that are based around the assumption that it is going to be used as a developer's local settings overrides, with values to be commented and uncommented as circumstances warrant.

PR #9536, opened 2021-03-09
consolidate conditional pytest sqlite3 db settings

rooftopcellist:
##### SUMMARY
@jbradberry noticed that the unit & functional tests are failing because they are being inadvertently pointed at the postgres db instead of sqlite.  

This is because of an import order issue cause by the introduction of `/etc/tower/conf.d/database.py` from the sources role.  

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - Installer

##### AWX VERSION

```
devel
```

### DIFF
--- a/awx/settings/development.py
+++ b/awx/settings/development.py
@@ -21,13 +21,6 @@ from split_settings.tools import optional, include
 # Load default settings.
 from .defaults import *  # NOQA
 
-if "pytest" in sys.modules:
-    CACHES = {
-        'default': {
-            'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
-            'LOCATION': 'unique-{}'.format(str(uuid.uuid4())),
-        },
-    }
 
 # awx-manage shell_plus --notebook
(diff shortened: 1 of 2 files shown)


awx/settings/development.py
```diff
@@ -166,6 +159,27 @@
     traceback.print_exc()
     sys.exit(1)
 
+# Use SQLite for unit tests instead of PostgreSQL.  If the lines below are
+# commented out, Django will create the test_awx-dev database in PostgreSQL to
+# run unit tests.
+if "pytest" in sys.modules:
```
rooftopcellist, line 165: Note: it is import that this be after the includes for the settings files here: https://github.com/ansible/awx/pull/9536/files#diff-bd08a936903e8f4f8830bf6d0caa6ff3520d75ec50a62c594b32c4a33e800912R140-R141

jbradberry (approved):

ryanpetrello (approved):

PR #9537, opened 2021-03-09
Updates Ling UI

AlexSCorey:
##### SUMMARY
This PR updates Ling ui. One of the reasons for updating this dependency was that they were deprecating `withI18n()`.  They changed their minds on that so we didn't need to urgently remove all those HOCs.  Thus, we can now make that conversion a bit slower a couple of files at a time.

One other thing:  When we are changing the string based on a count of an item (ie. Cancel Job vs. Cancel Jobs) we should use ling ui's `<Plural>` [component.](https://lingui.js.org/ref/macro.html?highlight=plural#id1) However, in order to show the update strings passed to that component the developer will have to run `npm run extract-strings` each time they are changed to render the updated strings properly. More info [here](https://lingui.js.org/ref/cli.html#extract).
##### ISSUE TYPE
 -dependency  upgrade

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION

```

```


awx/ui_next/src/App.jsx
```diff
@@ -29,7 +37,7 @@ const ProtectedRoute = ({ children, ...rest }) =>
   );
 
 function App() {
-  const catalogs = { en, ja };
+  const catalogs = { en: english, ja: japanese, zu: zulu, fr: french };
```
nixocio: Do we need to add the other just added languages also? 🇪🇸 🇨🇳 🇳🇱

AlexSCorey: based on the convo we had today at dev sync I am trying to figure out the best approach here. In production, yes I think we will need to add all the language catalogs, except `zu`.

akus062381:
@AlexSCorey  marking is blocked because I know that this current implementation cannot work with pluralization.

AlexSCorey:
Trying to wait until we get some guidance here https://github.com/lingui/js-lingui/issues/1003

awx/ui_next/src/App.jsx
```diff
@@ -39,43 +55,56 @@ function App() {
   const match = useRouteMatch();
   const { hash, search, pathname } = useLocation();
 
+  i18n.loadLocaleData('en', { plurals: en });
```
semoal: I would recommend just loading the initial catalog (for example english) and then load the other catalogs dynamically. Check out https://lingui.js.org/ref/loader.html

awx/ui_next/src/App.jsx
```diff
@@ -39,43 +55,56 @@ function App() {
...
-                        </ProtectedRoute>
-                      )}
-                  </Switch>
-                </AppContainer>
-              </ProtectedRoute>
-            </Switch>
-          </Background>
-        )}
-      </I18n>
+    <I18nProvider i18n={i18n} catalogs={catalogs}>
```
semoal: Catalogs isn't a property of the provider anymore since v3.0 :)

awx/ui_next/src/components/JobList/JobListCancelButton.jsx
```diff
@@ -54,35 +53,32 @@ function JobListCancelButton({ i18n, jobsToCancel, onCancel }) {
...
     if (numJobsToCancel > 0) {
-      return i18n._(
-        '{numJobsToCancel, plural, one {Cancel selected job} other {Cancel selected jobs}}',
-        {
-          numJobsToCancel,
-        }
+      return (
+        <Plural
+          value={numJobsToCancel}
+          one={i18n._(t`Cancel selected job`)}
```
semoal, line 79: Since v3 you don't need i18n._() when using the t`` macro.
Also in the Plural component you don't need to wrap one,other... with the t. just strings :)

awx/ui_next/src/components/Pagination/Pagination.jsx
```diff
@@ -19,26 +19,22 @@ const AWXPagination = styled(PFPagination)`
...
-          paginationTitle: i18n._(t`Pagination`),
-        }}
-        dropDirection={DropdownDirection.up}
-        {...props}
-      />
-    )}
-  </I18n>
+  <AWXPagination
+    titles={{
+      items: i18n._(t`items`),
```
semoal, line 24: As I said you can use directly the t`` macro

awx/ui_next/src/screens/Inventory/shared/InventoryGroupsDeleteModal.jsx
```diff
@@ -18,12 +18,8 @@ const ListItem = styled.li`
...
 `;
 
-const InventoryGroupsDeleteModal = ({
-  onAfterDelete,
-  isDisabled,
-  groups,
-  i18n,
-}) => {
+const InventoryGroupsDeleteModal = ({ onAfterDelete, isDisabled, groups }) => {
+  const { i18n } = useLingui();
```
semoal, line 22: are you using this for anything?

awx/ui_next/src/screens/Job/JobOutput/JobOutput.jsx
```diff
@@ -685,161 +685,157 @@ function JobOutput({ job, eventRelatedSearchableKeys, eventSearchableKeys }) {
...
+            onDelete={deleteJob}
+            isDeleteDisabled={isDeleting}
+          />
+        </OutputHeader>
+        <HostStatusBar counts={job.host_status_counts} />
+        <SearchToolbar
+          id="job_output-toolbar"
+          clearAllFilters={handleRemoveAllSearchTerms}
+          collapseListedFiltersBreakpoint="lg"
+          clearFiltersButtonText={i18n._(t`Clear all filters`)}
```
semoal, line 715: same :)

semoal (changes requested):

awx/ui_next/src/components/JobList/JobListCancelButton.jsx
AlexSCorey: @semoal you mentioned the Plural component.  We are having some trouble with this component described here https://github.com/lingui/js-lingui/issues/1003.  Are you able to offer and guidance on how to solve that problem?

awx/ui_next/src/screens/Inventory/shared/InventoryGroupsDeleteModal.jsx
AlexSCorey: we are, but it seems that we don't need to anymore.  I'll keep it in to follow our pattern for now and we will write an issue to update/remove the unnecessary i18n in the future.

awx/ui_next/src/App.jsx
```diff
@@ -29,53 +30,51 @@ const ProtectedRoute = ({ children, ...rest }) =>
...
     // preferred language, default to one that has strings.
     language = 'en';
   }
   const match = useRouteMatch();
   const { hash, search, pathname } = useLocation();
-
+  useEffect(() => {
+    dynamicActivate(language);
+  }, [language]);
+  i18n.activate(language);
```
semoal: why do you need this line? ^^

awx/ui_next/src/components/JobList/JobListCancelButton.jsx
semoal: If you need to remove all this strings I have a cool regex to do this in vscode:
```
i18n._\(t`(.*?)`\)
```

You copy this regex and will find all the occurrences with i18n._(tànystring), and then in your replace bar of the vscode you add 
```
t`$1`
```

Screenshot 2021-04-07 at 11 47 59

To remove the withI18n() hoc:

```
withI18n\(\)\((.*?)\)
```
to

```
$1
```

awx/ui_next/src/App.jsx
AlexSCorey: @semoal I get an error and it doesn't compile without that
`I18nProvider did not render. A call to i18n.activate still needs to happen or forceRenderOnLocaleChange must be set to false.`

awx/ui_next/src/components/JobList/JobList.jsx
```diff
@@ -1,7 +1,7 @@
 import React, { useState, useEffect, useCallback } from 'react';
 import { useLocation } from 'react-router-dom';
 import { withI18n } from '@lingui/react';
-import { t } from '@lingui/macro';
+import { t, plural } from '@lingui/macro';
```
nixocio, line 4: I noticed import of `Plural` and `plural`. That does not seem to matter. Right?

nixocio:
I triggered an e2e tests.  I am going to read more about the latest changes on the library.
A few things that reading the docs was unclear to me:
* Do we need to update any babel settings in order to use the dynamic import? or CRA setup is good enough?
* Do we need to setup `extractBabelOptions` since we are using CRA? https://lingui.js.org/ref/conf.html#extractbabeloptions

Overall great job @AlexSCorey !

semoal:
> I triggered an e2e tests. I am going to read more about the latest changes on the library.
> A few things that reading the docs was unclear to me:
> 
> * Do we need to update any babel settings in order to use the dynamic import? or CRA setup is good enough?
> * Do we need to setup `extractBabelOptions` since we are using CRA? [lingui.js.org/ref/conf.html#extractbabeloptions](https://lingui.js.org/ref/conf.html#extractbabeloptions)
> 
> Overall great job @AlexSCorey !

Hey, CRA setup is good enough to support dynamic imports, also there's no need to setup anything related to babel + lingui together with the cra default options. :)

Everything should work fine

awx/ui_next/src/App.jsx
semoal: Because you need to add that to the dynamicActivate function, just behind i18n.load(), extracting this out of the useEffect you're re-running it on each re-render

nixocio:
Thank you @semoal

nixocio (approved):
🚢

Issue #9559, opened 2021-03-11
Projects Unable to Sync Properly


beeankha:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

When attempting to sync/update a Project, the following error is output in the logs:
```
awx_1_1     | 2021-03-11 14:17:18,927 ERROR    [b3409386] awx.main.dispatch Worker failed to run task awx.main.tasks.RunProjectUpdate(*[10], **{}
awx_1_1     | Traceback (most recent call last):
awx_1_1     |   File "/var/lib/awx/venv/awx/lib/python3.6/site-packages/django/db/backends/base/base.py", line 235, in _cursor
awx_1_1     |     return self._prepare_cursor(self.create_cursor(name))
awx_1_1     |   File "/var/lib/awx/venv/awx/lib/python3.6/site-packages/django/db/backends/postgresql/base.py", line 223, in create_cursor
awx_1_1     |     cursor = self.connection.cursor()
awx_1_1     | psycopg2.InterfaceError: connection already closed
awx_1_1     | 
awx_1_1     | The above exception was the direct cause of the following exception:
awx_1_1     | 
awx_1_1     | Traceback (most recent call last):
awx_1_1     |   File "/awx_devel/awx/main/dispatch/worker/task.py", line 90, in perform_work
awx_1_1     |     result = self.run_callable(body)
awx_1_1     |   File "/awx_devel/awx/main/dispatch/worker/task.py", line 66, in run_callable
awx_1_1     |     return _call(*args, **kwargs)
awx_1_1     |   File "/awx_devel/awx/main/tasks.py", line 824, in _wrapped
awx_1_1     |     return f(self, *args, **kwargs)
awx_1_1     |   File "/awx_devel/awx/main/tasks.py", line 1340, in run
awx_1_1     |     start_args='')  # blank field to remove encrypted passwords
awx_1_1     |   File "/awx_devel/awx/main/tasks.py", line 868, in update_model
awx_1_1     |     instance.save(update_fields=update_fields)
awx_1_1     |   File "/awx_devel/awx/main/models/projects.py", line 655, in save
awx_1_1     |     return super(ProjectUpdate, self).save(*args, **kwargs)
awx_1_1     |   File "/awx_devel/awx/main/models/unified_jobs.py", line 890, in save
awx_1_1     |     self._update_parent_instance()
awx_1_1     |   File "/awx_devel/awx/main/models/projects.py", line 544, in _update_parent_instance
[...]

PR #9561, opened 2021-03-11
clean stale dispatcher connections closer to post-fork

ryanpetrello:
Note that relocating the post-fork cleanup code *here* is actually the *top frame* of the stack post-fork (this function is the actual `multiprocessing.Process` target we point at):

```python
awx_1_1     | Traceback (most recent call last):
awx_1_1     |   File "/awx_devel/awx/main/dispatch/worker/task.py", line 90, in perform_work
awx_1_1     |     result = self.run_callable(body)  <--------------------------
awx_1_1     |   File "/awx_devel/awx/main/dispatch/worker/task.py", line 66, in run_callable
awx_1_1     |     return _call(*args, **kwargs)
awx_1_1     |   File "/awx_devel/awx/main/tasks.py", line 824, in _wrapped
awx_1_1     |     return f(self, *args, **kwargs)
```

see: https://github.com/ansible/awx/issues/9559

### DIFF
--- a/awx/conf/settings.py
+++ b/awx/conf/settings.py
@@ -352,7 +352,6 @@ class SettingsWrapper(UserSettingsHolder):
         self.cache.set_many(settings_to_cache, timeout=SETTING_CACHE_TIMEOUT)
 
     def _get_local(self, name, validate=True):
-        self.__clean_on_fork__()
         self._preload_cache()
         cache_key = Setting.get_cache_key(name)
         try:
--- a/awx/main/dispatch/worker/task.py
+++ b/awx/main/dispatch/worker/task.py
@@ -6,6 +6,7 @@ import traceback
 
 from kubernetes.config import kube_config
 
+from django.conf import settings
 from django_guid.middleware import GuidMiddleware
 
 from awx.main.tasks import dispatch_startup, inform_cluster_of_shutdown
@@ -85,6 +86,7 @@ class TaskWorker(BaseWorker):
             'task': u'awx.main.tasks.RunProjectUpdate'
         }
         '''
+        settings.__clean_on_fork__()
         result = None
         try:
             result = self.run_callable(body)


awx/conf/settings.py
```diff
@@ -352,7 +352,6 @@ def _preload_cache(self):
         self.cache.set_many(settings_to_cache, timeout=SETTING_CACHE_TIMEOUT)
 
     def _get_local(self, name, validate=True):
-        self.__clean_on_fork__()
```
ryanpetrello, line 355: This function gets called **a lot**, and I think the frequency of this call explains the racy nature of this bug.

awx/main/dispatch/worker/task.py
```diff
@@ -85,6 +86,7 @@ def perform_work(self, body):
             'task': u'awx.main.tasks.RunProjectUpdate'
         }
         '''
+        settings.__clean_on_fork__()
```
ryanpetrello, line 89: In reality, we should only need to call this *once* in the forked child process, and this function is what `multiprocessing.Process`'s target is pointed at for the dispatcher.

We also don't care about this cleanup for the callback receiver, because unlike the dispatcher, it doesn't auto scale up/down work processes, so there isn't a fork cleanup race.

chrismeyersfsu:
Just to leave a trail of breadcrumbs in the form of random thoughts.

I think I like this fix.

* This doesn't solve the general problem of forking Django. Specifically, around the database connection. However, our dispatcher is the only sub-system with this constant fork() pattern.
* Our callback receiver forks but only on init. If a user hit this for the callback receiver then wouldn't the parent process just restart it and we go about our business?
* uWSGI it multi-process but it forks before Django loads.

chrismeyersfsu (approved):

beeankha (approved):

ryanpetrello:
regate

Issue #9207, opened 2021-02-02
[ExecEnvs] Surface missing-resource warnings on list items for UJTs that were using custom virtualenvs 


jakemcdermott:
##### SUMMARY
After upgrade, job templates that were using virtualenvs will need to be linked to an execution environment. We should show a "missing resource" warning (like we do for deleted projects, etc.) on list items to help inform the user about the actions they need to take in order to make the template runnable.

Potential implementation details: We can add a new api field to indicate warning status or the ui can just derive the state by looking at which job templates have a value for the old virtualenv field but don't have an EE yet.

AlanCoding:
Related to other EE migration issues https://github.com/ansible/awx/issues/7062 that @rebeccahhh has been working on

and less-so https://github.com/ansible/awx/issues/9146

tiagodread:
@one-t can we close this? I think this was tested here https://github.com/ansible/awx/pull/9570 right?

Issue #9190, opened 2021-01-28
Remove the custom venv fields from the UI


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Remove the custom venv fields from the UI. Every place that shows a custom virtual env should display the execution environment. For instance, Job Details output.

AlanCoding:
Ping @rebeccahhh and @shanemcd, there was talk of having some placeholder tooltips (this may not be the right word) for how to migrate these fields. I want to get you in the loop for that discussion.

unlikelyzero:
@mabashian this is required for EE environments

tiagodread:
@one-t could we close this issue?

PR #9570, opened 2021-03-11
Remove custom virtual env

nixocio:
Remove custom virtual from the UI.

Also, surface missing-resource warnings on list items for UJTs that were using
custom virtualenvs. And related details page.

See: https://github.com/ansible/awx/issues/9190
Also: https://github.com/ansible/awx/issues/9207


nixocio:
@tiagodread @unlikelyzero this still a working in progress, but I will run the e2e tests. I believe this one will disrupt a lot tests.

nixocio:
Adding more unit-tests.

one-t:
Just some testing updates-
I gave this a spin by adding a custom venv to a job template via psql commands
```
UPDATE main_jobtemplate
SET custom_virtualenv = '/opt/my-venvs/custom-venv';
```
image

I did find that the custom venv paths option is still present in system settings:
image

Otherwise, this looks good to me. Once that is removed from the system settings page we should be good to merge.

awx/ui_next/src/components/ExecutionEnvironmentDetail/ExecutionEnvironmentDetail.jsx
```diff
@@ -0,0 +1,82 @@
...
+    return (
+      <Detail
+        label={label}
+        value={
+          <>
+            {i18n._(t`Missing resource`)}
+            <span>
+              <Tooltip
+                content={i18n._(
+                  t`Custom virtual environment ${virtualEnvironment} must be replaced by an execution environment.`
```
ryanpetrello, line 54: cc @shanemcd @wenottingham @tvo318 we may want to come up with some instructions (at some URL) for doing so - I have to imagine we'll have some type of knowledge base article or documentation that describes the transition process - do we want to file an issue to track this work so we don't forget it later?

jakemcdermott (approved):

nixocio:
> Just some testing updates-
> I gave this a spin by adding a custom venv to a job template via psql commands
> 
> ```
> UPDATE main_jobtemplate
> SET custom_virtualenv = '/opt/my-venvs/custom-venv';
> ```
> 
> image
> 
> I did find that the custom venv paths option is still present in system settings:
> image
> 
> Otherwise, this looks good to me. Once that is removed from the system settings page we should be good to merge.

image

image

image

I updated the settings. Good find @one-t. If you want to take another look.

one-t:
That looks good- I do have a couple other points to mention, though they may be less common.

- [x]  Custom Venvs can be defined in three places, with increasing precedence. (Org, Project, JT) Do we want to display warnings on job templates and projects that inherit their custom venv from one of these resources?
- [x] This warning on a org or project with a venv defined should possibly be updated to indicate that there is a custom venv defined rather than saying it's missing an EE since it is inherited from the org.
image
  - this one is kind of tricky because you can't unset the custom venv in the UI.

Since custom venvs are already a power-user thing, maybe we just add some documentation on how to remove the custom venvs via awx-manage? 

Anyway, not really bugs at all, just potential UX issues.

nixocio:
Hey @one-t, we do show warnings about missing EE on lists for - Orgs, Projects, Inventory Sources, and Job Templates. We also show warnings on details pages for those resources.

image


As we talked, based on your feedback I will make the message on the list to have the custom virtual env to be replaced as shown on the details page. That is a good suggestion.

nixocio:
@one-t, updated. Now the same message is displayed on the lists and details.  I triggered the e2e tests again to verify the latest changes.

one-t (approved):

Issue #8907, opened 2020-12-15
Malformed yaml can crash the launch/workflow node prompt wizard


mabashian:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
This should actually be addressed by https://github.com/ansible/awx/issues/8906.  If we had validation on the extra variables field then the user would address this there.

This bug won't make it's way into devel until https://github.com/ansible/awx/pull/8399 merges.

##### STEPS TO REPRODUCE
1. Create a JT that prompts for extra variables
2. Launch the JT
3. Enter the following text in the extra variables field:

```
---
foo: bar
foobar
```

4. Go to the preview step

##### EXPECTED RESULTS
Error was shown before going to preview step

##### ACTUAL RESULTS
Page crashes when we attempt to transform the extra vars.

mabashian:
Here's the actual error that gets thrown:

`YAMLException: can not read a block mapping entry; a multiline key may not be an implicit key at line 4, column 1:`

mabashian:
@keithjgrant you think this is covered by the ace editor PR?  If so can we retest?

keithjgrant:
Yes, my PR should prevent this from happening. Is it worth adding additional handling to the page in case any pre-existing bad data gets fetched?

tiagodread:
The PR was merged, bug fixed.

Issue #8906, opened 2020-12-15
Add proper validation to CodeMirrorField component


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
It doesn't look like this was ever really implemented.  There's a prop on the component where you _could_ pass in a validate function but we don't actually do that in any of the places.

The idea here is that we want to validate the field for malformed yaml/json and _at a minimum_ show an error beneath the field.

The reason why I think this is going to be difficult is because we have to deal with formik.  By default, formik is going to run validation on _change_ and on _blur_.  In the case of these codemirror fields we definitely don't want to validate on change (each keystroke) since this would show errors constantly.  We probably only want to validate on blur.  This can be configured at the _form_ level with a prop passed to `<Formik ... ` but that would mean _all_ the fields in the form would only validate on blur.  We may have to come up with a clever solution here to only validate these codemirror fields on blur but validate the rest of the form on change and blur.

This is particularly important for places with prompts.  We manipulate/combine extra vars and survey values later on during the process and having unvalidated data makes error handling difficult.

This may also tie in to https://github.com/ansible/awx/issues/8905

jakemcdermott:
May also be related to: https://github.com/ansible/awx/issues/8898#issuecomment-744838149 ?

Not really sure, but I think adding form-like field validation might help there?

unlikelyzero:
and also related to #8816

akus062381:
waiting on implementation
https://github.com/ansible/awx/issues/8905

keithjgrant:
> The reason why I think this is going to be difficult is because we have to deal with formik. By default, formik is going to run validation on _change_ and on _blur_. In the case of these codemirror fields we definitely don't want to validate on change (each keystroke) since this would show errors constantly.

Our normal tactic is to wait until the first onBlur event to begin validating a field. Is there a reason that approach wouldn't work here?

unlikelyzero:
@one-t is workin' on it

PR #9586, opened 2021-03-15
Add VariablesField YAML/JSON validation

keithjgrant:
##### SUMMARY
Validates YAML or JSON syntax in the variables field on blur. Prevents formik from submitting the form while this error is present.

addresses #8906, #8907

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
Screen Shot 2021-03-15 at 11 31 19 AM

### DIFF
--- a/awx/ui_next/src/components/CodeEditor/CodeEditor.jsx
+++ b/awx/ui_next/src/components/CodeEditor/CodeEditor.jsx
@@ -73,6 +73,8 @@ function CodeEditor({
   id,
   value,
   onChange,
+  onFocus,
+  onBlur,
   mode,
   readOnly,
   hasErrors,
@@ -140,6 +142,8 @@ function CodeEditor({
           theme="github"
           onChange={debounce(onChange, 250)}
           value={value}
+          onFocus={onFocus}
+          onBlur={onBlur}
           name={id || 'code-editor'}
           editorProps={{ $blockScrolling: true }}
           fontSize={16}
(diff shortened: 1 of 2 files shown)


keithjgrant:
NOTE: this relies on the browser's JSON parser and our YAML library's parser for error messages, so I don't think there's a viable strategy for translation here 😕

nixocio (approved):

unlikelyzero:
@keithjgrant can you ensure that these strings are i18n'd ?

one-t:
Hey, @keithjgrant - I found that this validates there are no duplicate keys for YAML, but it does not prevent saving JSON with repeated values.
image

keithjgrant:
> Hey, @keithjgrant - I found that this validates there are no duplicate keys for YAML, but it does not prevent saving JSON with repeated values.

@one-t Hm. Technically, that is valid JSON (and most standard JSON validators will accept it. e.g. https://codebeautify.org/jsonvalidator). I talked a bit with @ryanpetrello and @chrismeyersfsu about this... we don't think the old UI checked for this, so it's not necessary for feature parity. It would also be pretty difficult to implement, since the browser's native JSON parser (and basically every other JSON parser lib) quietly accept duplicate keys.

jakemcdermott (approved):

one-t:
regate

Issue #9592, opened 2021-03-16
Unable to select non-global execution environments for organizations


one-t:
If you have an execution environment that has organizations assigned to it, you can't choose that EE from the UI when assigning EEs to an organization.

Steps to reproduce:
1. Create a new EE 
  a. Use `quay.io/ansible/awx-ee` for the image
  b. assign it to an organzation
2. Navigate to that organization's page
3. Try to assign that new EE to the organization.

It won't show up in the list. I've verified that you can assign it via the API.

PR #9596, opened 2021-03-16
Allow one to select non-global execution environments for organizations

nixocio:
Allow one to select non-global EE when editing an Organization.

See: https://github.com/ansible/awx/issues/9592

All those EE should be present as a choice when editing the `Default` organization.

image

Editing `Default` organization.

image


jakemcdermott (approved):

tiagodread (approved):
Added a test to cover this

Issue #9473, opened 2021-03-03
Add ansible version to collected EE data


wenottingham:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
https://github.com/ansible/awx/pull/9335 adds support for pulling collections and their versions from an EE.

This should be enhanced to also pass the ansible (core) version somehow.

michaeltcoelho:
PR https://github.com/ansible/tower-qa/pull/6172

PR #9598, opened 2021-03-16
Enable Ansible version to be collected from EEs

beeankha:
##### SUMMARY

Connecting issue https://github.com/ansible/awx/issues/9473

This PR, along with this [Ansible-Runner PR](https://github.com/ansible/ansible-runner/pull/647), enables us to obtain the Ansible (core) version for each execution environment that is utilized. This info can be gathered from the new `ansible_version` column on the `main_unifiedjobs` table.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 17.0.1
```
##### ADDITIONAL INFORMATION
Screenshot/example of the DB output:
Screenshot from 2021-03-16 16-57-32


awx/main/models/unified_jobs.py
```diff
@@ -728,6 +728,13 @@ class Meta:
         editable=False,
         help_text=_("The Collections names and versions installed in the execution environment."),
     )
+    ansible_version = models.CharField(
+        max_length=50,
```
ryanpetrello: Probably better to err on this side of making this larger, in case some future version of Ansible prints a really lengthy version string e.g.,

```python
>>> len('ansible automation engine [core enterprise edition 4.0.1<githash>]')
66
```

Maybe 255 🤷 ?

ryanpetrello: https://www.youtube.com/watch?v=7ffj8SHrbk0

awx/main/analytics/collectors.py
```diff
@@ -311,7 +311,7 @@ def events_table(since, full_path, until, **kwargs):
     return _copy_table(table='events', query=events_query, path=full_path)
 
 
-@register('unified_jobs_table', '1.2', format='csv', description=_('Data on jobs run'), expensive=True)
```
ryanpetrello: @beeankha @Ladas have we actually shipped a version with 1.2 yet (does this need to actually change - I thought the 1.2 bump was for `installed_collections`, which is probably not technically out in the wild yet?)

awx/main/migrations/0133_unifiedjob_ansible_version.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('main', '0132_instancegroup_is_container_group'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='unifiedjob',
+            name='ansible_version',
+            field=models.CharField(blank=True, default='', editable=False, help_text='The version of Ansible Core installed in the execution environment.', max_length=50),
```
ryanpetrello: @beeankha the `50` in this migration should be updated to match too

ryanpetrello (approved):
A few minor comments, otherwise this looks great.

Thanks @beeankha 😄

awx/main/migrations/0133_unifiedjob_ansible_version.py
beeankha: Oop, sorry to miss that! Fixed!

awx/main/analytics/collectors.py
Ladas: I am fine with both :-)

(we're not actually using the version for anything :-) , we will really need to do that only if we change e.g. the format of the field, in a way that would require conditional processing based on version)

beeankha: I'll change it back to `1.2`!

awx/main/tasks.py
```diff
@@ -1195,6 +1196,10 @@ def final_run_hook(self, instance, status, private_data_dir, fact_modification_t
                 ee_collections_info = json.loads(ee_json_info.read())
                 instance.installed_collections = ee_collections_info
                 instance.save(update_fields=['installed_collections'])
+            with open(ansible_version_file) as ee_ansible_info:
```
shanemcd, line 1200: I think it would be best to explicitly check for the existence of this file.

shanemcd (approved):

Ladas (approved):
:+1:

beeankha:
regate

PR #9599, opened 2021-03-16
refactor payload construction for awxkit

kdelee:
This fixes container_group creation to allow passing
"is_container_group" and "credential" to the "create" method
on instance groups, and refactors other page objects
to use a common utility function to eliminate copy-pasted code

This will help us update to set `is_container_group` correctly as is now needed since https://github.com/ansible/awx/commit/de52adedef525b3c67a42f218f2a736aa1e8c7db


unlikelyzero:
@kdelee we are using this construct in our ui tests. Please ensure that they are up to date along with the api tests

ryanpetrello (approved):

PR #9605, opened 2021-03-17
fix port conflict in dev cluster

fosterseth:
##### SUMMARY

problem: loop adds 100 to ports 7899 and 7999, which would yield 7999 to 8099 on the next iteration, so the 7999 is conflicting
fix: add 1000 instead
Also, haproxy was being defined twice, now it renders once.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 17.0.1
```

### DIFF
--- a/tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
+++ b/tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
@@ -3,8 +3,8 @@ version: '2'
 services:
 {% for i in range(cluster_node_count|int) %}
   {% set container_postfix = loop.index %}
-  {% set awx_sdb_port_start = 7899 + (loop.index0*100) | int %}
-  {% set awx_sdb_port_end = 7999 + (loop.index0*100) | int %}
+  {% set awx_sdb_port_start = 7899 + (loop.index0*1000) | int %}
+  {% set awx_sdb_port_end = 7999 + (loop.index0*1000) | int %}
   # Primary AWX Development Container
   awx_{{ container_postfix }}:
     user: "{{ ansible_user_uid }}"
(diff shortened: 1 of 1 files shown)


tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
```diff
@@ -37,27 +37,13 @@ services:
     privileged: true
     tty: true
     ports:
-      - "6899:6899"
       - "{{ awx_sdb_port_start }}-{{ awx_sdb_port_end }}:{{ awx_sdb_port_start }}-{{ awx_sdb_port_end }}"  # sdb-listen
 {% if cluster_node_count|int == 1 %}
+      - "6899:6899"
```
fosterseth, line 42: not sure what this port does

tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
```diff
@@ -37,27 +37,13 @@ services:
...
       - "8080:8080"  # unused but mapped for debugging
       - "8888:8888"  # jupyter notebook
       - "8013:8013"  # http
       - "8043:8043"  # https
-{% else %}
```
fosterseth, line 47: pull this out of the main for loop, we only want to render the haproxy def once

tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
ryanpetrello: This is the port used to advertise connections to `sdb-listen` w/ UDP.  Whenever `set_trace()` is called, a target port is chosen, and then it's sent to `0.0.0.0:6899` (which is what `sdb-listen` listens on):

https://github.com/ryanpetrello/sdb/blob/master/sdb.py#L150-L154

When `sdb-listen` receives these UDP messages, they contain "port numbers to connect to", and it tosses them into a queue of ports to connect to:

https://github.com/ryanpetrello/sdb/blob/master/sdb.py#L399-L407
https://github.com/ryanpetrello/sdb/blob/master/sdb.py#L415-L424

shanemcd (approved):

PR #9613, opened 2021-03-18
Prevent users from selecting credentials that prompt for passwords on workflow nodes and schedules

mabashian:
##### SUMMARY
link #8921 

If a user selects a job template with default credentials that prompt for passwords (but does not prompt for credentials) then the user should not be allowed to create the node and a different JT must be selected:

Screen Shot 2021-03-18 at 9 33 14 AM

If a user selects a credential that prompts for passwords when creating/editing a workflow node or schedule then we show this error:

Screen Shot 2021-03-18 at 9 48 43 AM

If a user removes a credential that exists in the default collection of credentials on the JT then it must be replaced.  This is the error we show:

Screen Shot 2021-03-18 at 9 49 49 AM

If a user attempts to create a schedule for a job template with default credentials that prompt (but does not prompt for credentials) then the API responds with this error:

Screen Shot 2021-03-18 at 9 45 06 AM

I believe this UX is consistent with the old UI but I am double checking that now.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/components/LaunchButton/LaunchButton.jsx
+++ b/awx/ui_next/src/components/LaunchButton/LaunchButton.jsx
@@ -36,6 +36,7 @@ function LaunchButton({ resource, i18n, children, history }) {
   const [showLaunchPrompt, setShowLaunchPrompt] = useState(false);
   const [launchConfig, setLaunchConfig] = useState(null);
   const [surveyConfig, setSurveyConfig] = useState(null);
+  const [resourceCredentials, setResourceCredentials] = useState([]);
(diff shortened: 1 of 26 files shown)


nixocio (approved):
👍

tiagodread (approved):
Verified https://github.com/ansible/awx/issues/8921#issuecomment-781470001 and https://github.com/ansible/awx/issues/8921#issuecomment-788134372 did some exploratory testing around this feature and worked as expect.

PR #9627, opened 2021-03-19
fix the large file parsing in project sync

jainnikhil30:
##### SUMMARY
Fixes the issue of scm update stuck on reading large files because of islice trying to read the whole file.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/utils/ansible.py
+++ b/awx/main/utils/ansible.py
@@ -6,7 +6,6 @@ import codecs
 import re
 import os
 import logging
-from itertools import islice
 from configparser import ConfigParser
 
 # Django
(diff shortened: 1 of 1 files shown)


awx/main/utils/ansible.py
```diff
@@ -87,19 +87,25 @@ def could_be_inventory(project_path, dir_path, filename):
...
             inventory_path,
-            'r',
-            encoding='utf-8',
-            errors='ignore'
+            encoding='utf-8'
         ) as inv_file:
-            for line in islice(inv_file, 10):
-                if not valid_inventory_re.match(line):
-                    return None
+            for line in inv_file:
```
ryanpetrello: Instead of tracking a counter, you can do:

```python
for i, line in enumerate(inv_file):
    if i > 10:
        break
    if valid_inventory_re.match(line):
        matched = True
        break
```

awx/main/utils/ansible.py
```diff
@@ -87,19 +87,25 @@ def could_be_inventory(project_path, dir_path, filename):
...
     try:
         # only read through first 10 lines for performance
-        with codecs.open(
+        counter = 0
+        with open(
             inventory_path,
-            'r',
-            encoding='utf-8',
-            errors='ignore'
+            encoding='utf-8'
```
ryanpetrello: We should keep `errors='ignore'`

jainnikhil30: yup just added that

awx/main/utils/ansible.py
ryanpetrello: *or* you could probably do this instead:

```python
for line in inv_file[:10]:
    if valid_inventory_re.match(line):
        matched = True
        break
```

jainnikhil30: Thanks Ryan. I used your first suggestion and removed the counter.

ryanpetrello (approved):
Looks good to me - thanks @jainnikhil30!

ryanpetrello:
Thanks @jainnikhil30!

Issue #9628, opened 2021-03-19
Filter by container registry credential to be associated with EE


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Filter by container registry credential to be associated with EE

ryanpetrello:
Additionally, at the API level, we want to make sure that *only* container registry credentials can be associated with an EE cc @rebeccahhh

tiagodread:
The lookup field is filtering by container registry:

![image](https://user-images.githubusercontent.com/12464858/119715944-f1db6500-be3a-11eb-80b8-99c86856bc9c.png)

PR #9639, opened 2021-03-19
Filter credential type by Container Registry for EE

nixocio:
Filter credential type by Container Registry for EE form.

See: https://github.com/ansible/awx/issues/9628

### DIFF
--- a/awx/ui_next/src/screens/ExecutionEnvironment/shared/ExecutionEnvironmentForm.jsx
+++ b/awx/ui_next/src/screens/ExecutionEnvironment/shared/ExecutionEnvironmentForm.jsx
@@ -23,7 +23,9 @@ function ExecutionEnvironmentFormFields({
   options,
   executionEnvironment,
 }) {
-  const [credentialField] = useField('credential');
+  const [credentialField, credentialMeta, credentialHelpers] = useField(
+    'credential'
+  );
   const [organizationField, organizationMeta, organizationHelpers] = useField({
     name: 'organization',
     validate:
@@ -124,8 +126,15 @@ function ExecutionEnvironmentFormFields({
 
       <CredentialLookup
         label={i18n._(t`Registry credential`)}
+        credentialTypeKind="registry"
+        helperTextInvalid={credentialMeta.error}
+        isValid={!credentialMeta.touched || !credentialMeta.error}
+        onBlur={() => credentialHelpers.setTouched()}
         onChange={onCredentialChange}
         value={credentialField.value}
+        tooltip={i18n._(
+          t`Credential to authenticate with a protected container registry.`
+        )}
       />
     </>
   );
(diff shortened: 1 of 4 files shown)


awx/ui_next/src/screens/ExecutionEnvironment/shared/ExecutionEnvironmentForm.jsx
```diff
@@ -124,8 +126,15 @@ function ExecutionEnvironmentFormFields({
 
       <CredentialLookup
         label={i18n._(t`Registry credential`)}
+        credentialTypeKind="registry"
```
jakemcdermott, line 129: :+1: Nice. Didn't know we already had a param for this!

awx/ui_next/src/screens/ExecutionEnvironment/shared/ExecutionEnvironmentForm.jsx
```diff
@@ -124,8 +126,15 @@ function ExecutionEnvironmentFormFields({
...
+        helperTextInvalid={credentialMeta.error}
+        isValid={!credentialMeta.touched || !credentialMeta.error}
+        onBlur={() => credentialHelpers.setTouched()}
         onChange={onCredentialChange}
         value={credentialField.value}
+        tooltip={i18n._(
+          t`Credential to authenticate with a protected container registry. Must be of type "Container Registry".`
```
jakemcdermott: @nixocio

Since we're _only_ showing container registry credentials, do we really need to warn the user about other types of credentials? Presumably they wouldn't even be able to select other types.

jakemcdermott (approved):
lgtm - but see comments / questions

awx/ui_next/src/screens/ExecutionEnvironment/shared/ExecutionEnvironmentForm.jsx
nixocio: @jakemcdermott, I looked at the container groups tooltip, there we also filter by a specific type of credential, and display the following message.

```Credential to authenticate with Kubernetes or OpenShift. Must be of type "Kubernetes/OpenShift API Bearer Token". If left blank, the underlying Pod's service account will be used.```

We can easily update this one if we decide that is redundant. Just let me know.

awx/ui_next/src/screens/ExecutionEnvironment/shared/ExecutionEnvironmentForm.jsx
nixocio: We do also have a parameter to filter by credential namespace. Credential lookup is a bit more complex due to those flags AFAIK.

tiagodread (approved):
Just a warn: We are filtering by kind `registry` on UI, if this change in the future on the API We'll need update the front as well.

awx/ui_next/src/screens/ExecutionEnvironment/shared/ExecutionEnvironmentForm.jsx
jakemcdermott: Let's leave the tooltips in but remove the "Must be of type" warning.

nixocio: @jakemcdermott, done. I will merge this one, since@tiagodread already approved. Thank you both.

nixocio:
regate

Issue #9088, opened 2021-01-14
Add option to select a default global EE on settings


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Add option to select default global EE. Evaluate best place to add/edit this feature. 

Take RBAC in consideration.

PR #9653, opened 2021-03-22
Add EE to the settings page

nixocio:
Allow a system admin to set the global default execution environment.

See: https://github.com/ansible/awx/issues/9088

This PR is also addressing the issue: #9669

Edit:

image

image

Details

image

### DIFF
--- a/awx/ui_next/src/screens/Setting/MiscSystem/MiscSystemDetail/MiscSystemDetail.jsx
+++ b/awx/ui_next/src/screens/Setting/MiscSystem/MiscSystemDetail/MiscSystemDetail.jsx
@@ -9,7 +9,7 @@ import ContentError from '../../../../components/ContentError';
 import ContentLoading from '../../../../components/ContentLoading';
 import { DetailList } from '../../../../components/DetailList';
 import RoutedTabs from '../../../../components/RoutedTabs';
-import { SettingsAPI } from '../../../../api';
+import { SettingsAPI, ExecutionEnvironmentsAPI } from '../../../../api';
 import useRequest from '../../../../util/useRequest';
 import { useConfig } from '../../../../contexts/Config';
 import { useSettings } from '../../../../contexts/Settings';
@@ -23,7 +23,15 @@ function MiscSystemDetail({ i18n }) {
   const { isLoading, error, request, result: system } = useRequest(
     useCallback(async () => {
       const { data } = await SettingsAPI.readCategory('all');
-
+      let DEFAULT_EXECUTION_ENVIRONMENT = '';
+      if (data.DEFAULT_EXECUTION_ENVIRONMENT) {
+        const {
+          data: { name },
+        } = await ExecutionEnvironmentsAPI.readDetail(
+          data.DEFAULT_EXECUTION_ENVIRONMENT
+        );
+        DEFAULT_EXECUTION_ENVIRONMENT = name;
+      }
       const {
         OAUTH2_PROVIDER: {
           ACCESS_TOKEN_EXPIRE_SECONDS,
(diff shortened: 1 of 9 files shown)


awx/ui_next/src/screens/Setting/MiscSystem/MiscSystemDetail/MiscSystemDetail.jsx
```diff
@@ -20,6 +20,25 @@ function MiscSystemDetail({ i18n }) {
...
+      if (!executionEnvironmentId) {
+        return '';
+      }
+      const { data } = await ExecutionEnvironmentsAPI.readDetail(
+        executionEnvironmentId
+      );
+      return data;
+    }, [executionEnvironmentId])
+  );
+
```
nixocio: Since the settings API does not have `summary_fields`. It was necessary an extra request for Details/Edit screens in order to fetch the EE and display the proper name.

nixocio:
running e2e tests.

awx/ui_next/src/screens/Setting/MiscSystem/MiscSystemDetail/MiscSystemDetail.jsx
nixocio: Also, I did not add a `Revert` for the EE lookup. Since the lookup is different from the other components rendered on that screen.

tiagodread:
Hey @nixocio some feedback, check if this make sense:

- [x] We are not displaying the helper popover for Execution Environment field on edit
- [x] The help popover on details is empty
image
image
- [x] We are calling `Execution Environment` field on edit form and the same field as `Global default execution environment` on details. I'm ok with `Global default execution environment` for both to avoid misunderstanding.

nixocio:
@tiagodread, please, take another look.

Details page.

image

Edit page.

image

awx/ui_next/src/screens/Setting/shared/SettingDetail.jsx
```diff
@@ -88,6 +88,8 @@ export default withI18n()(
         );
         break;
       case 'choice':
+      case 'field':
+      case 'string':
```
marshmalien, line 92: Great idea!

awx/ui_next/src/screens/Setting/shared/data.allSettingOptions.json
```diff
@@ -7091,6 +7099,15 @@
                   "read_only": false
               }
           },
+          "DEFAULT_EXECUTION_ENVIRONMENT": {
+            "type": "field",
+            "required": false,
+            "label": "Global default execution environment",
+            "help_text": ".",
```
marshmalien: "help_text" should match the changes added in `/main/conf.py`

awx/ui_next/src/screens/Setting/shared/SharedFields.jsx
```diff
@@ -76,6 +77,32 @@ const SettingGroup = withI18n()(
   )
 );
 
+const ExecutionEnvironmentLookupField = withI18n()(({ i18n }) => {
```
marshmalien: Since this lookup is used in only one place (MiscSystemEdit) within Settings, it probably doesn't need to be in SharedFields and instead we could import the lookup directly from /components in MiscSystemEdit.

nixocio:
@marshmalien, updated as per your comments.

awx/ui_next/src/screens/Setting/MiscSystem/MiscSystemDetail/MiscSystemDetail.test.jsx
```diff
@@ -110,6 +121,74 @@ describe('<MiscSystemDetail />', () => {
...
+            DEFAULT_EXECUTION_ENVIRONMENT: null,
+          }}
+        >
+          <MiscSystemDetail />
+        </SettingsProvider>
+      );
+    });
+    await waitForElement(newWrapper, 'ContentLoading', el => el.length === 0);
+
+    assertDetail(newWrapper, 'Access Token Expiration', '1 seconds');
```
marshmalien: We could probably remove some of these detail label and value assertions since we check them in the previous test.

marshmalien (approved):
LGTM! Tests are great and I like how you cleaned up the shared setting details.

tiagodread (approved):

PR #9654, opened 2021-03-22
Adding host_metrics for customers who don't have automation analytics

amolgautam25:
##### SUMMARY

I am trying to accomplish the following: 

- [x] Create a new model that stores details about hostname
- [x] create corresponding migrations for it
- [x] Update the table (main_hostmetrics) after playbook execution
- [x] retrieve the data with custom ```awx-manage``` command : ``` awx-manage host_metrics --since <date> --until <date>```

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 18.0.0

```

##### ADDITIONAL INFORMATION

```

```


amolgautam25:
I was able to create a model and corresponding migrations for it. Here is the table : 
```
                                           Table "public.main_hostmetrics"
      Column      |           Type           | Collation | Nullable | Default | Storage  | Stats target | Description 
------------------+--------------------------+-----------+----------+---------+----------+--------------+-------------
 hostname         | character varying(30)    |           | not null |         | extended |              | 
 first_automation | timestamp with time zone |           | not null |         | plain    |              | 
 last_automation  | timestamp with time zone |           | not null |         | plain    |              | 
Indexes:
    "main_hostmetrics_hostname_da8dac94_pk" PRIMARY KEY, btree (hostname)
    "main_hostmetrics_hostname_da8dac94_uniq" UNIQUE CONSTRAINT, btree (hostname)
    "main_hostmetrics_hostname_da8dac94_like" btree (hostname varchar_pattern_ops)
Access method: heap

File: awx/main/models/host_metrics.py [hunk 1]
```diff
@@ -0,0 +1,6 @@
+from django.db import models
+
+class HostMetrics(models.Model):
```
ryanpetrello: I'd put this model in `awx.main.models.inventory` near the `class Host` definition since it's pretty similar.

awx/main/models/host_metrics.py
```diff
@@ -0,0 +1,6 @@
...
+    hostname = models.CharField(primary_key=True,max_length=30)
```
ryanpetrello: I'd expand this `max_length` to 512 to match the related `Host` model; it's very likely that people will have hostnames over 30 characters:

https://github.com/ansible/awx/blob/devel/awx/main/models/inventory.py#L443
https://github.com/ansible/awx/blob/devel/awx/main/models/base.py#L359

awx/main/models/host_metrics.py
```diff
@@ -0,0 +1,6 @@
...
+    first_automation = models.DateTimeField()
```
ryanpetrello: It might make sense to just have a default on `first_automation` that gets set on insert e.g., `default=django.utils.timezone.now`

awx/main/models/host_metrics.py
amolgautam25: okay , i will change that.

awx/main/models/host_metrics.py
amolgautam25: I have updated it

awx/main/models/host_metrics.py
amolgautam25: I have updated it

awx/main/models/host_metrics.py
amolgautam25: I have updated it

awx/main/management/commands/host_metrics.py
```diff
@@ -0,0 +1,34 @@
...
+        parser.add_argument('--until', dest='until', action='store', help='End date for data')
+
+    def handle(self, *args, **options):
+        opt_since = options.get('since')
+        opt_until = options.get('until')
+        if opt_since:
+            since = parser.parse(opt_since)
+            since = since.replace(tzinfo=timezone.utc)
+        else:
+            since = None
```
amolgautam25: Shall i set ```since``` to earliest date possible in Postgres SQL, or display an error message to user stating that this date is not valid ?

awx/main/models/events.py
```diff
@@ -535,16 +536,19 @@ def _update_host_summary_from_stats(self, hostnames):
...
+            updated_hosts_list = []
             for h in all_hosts:
                 # if the hostname *shows up* in the playbook_on_stats event
                 if h.name in hostnames:
                     h.last_job_id = job.id
                     updated_hosts.add(h)
+                    updated_hosts.append(h.name)
                 if h.id in host_mapping:
                     h.last_job_host_summary_id = host_mapping[h.id]
                     updated_hosts.add(h)
```
amolgautam25: I am not sure if i should make another list and append ```h.id``` is that. I only need ```h.name``` ( hostname ) in my new table/model. 
Let me know.

awx/main/models/inventory.py
```diff
@@ -803,6 +803,21 @@ def _get_related_jobs(self):
         return UnifiedJob.objects.non_polymorphic().filter(Q(job__inventory=self.inventory) | Q(inventoryupdate__inventory_source__groups=self))
 
 
+class HostMetrics(models.Model):
+    hostname = models.CharField(primary_key=True, max_length=512)
+    first_automation = models.DateTimeField(auto_now_add=True, null=False)
+    last_automation = models.DateTimeField()
+
+    def host_metrics_update(self, hostname):
```
amolgautam25: I have created some helper functions here. 
My original logic was to put ```bulk_create()``` and ```bulk_update()``` code in these helper function. 
But later i realized that i can write them down in the ```awx/main/models/events.py``` file itself.

Let me know what you think is a better approach. Writing the ```bulk_create``` and ```bulk_update``` query in ```    def _update_host_summary_from_stats(self, hostnames) ``` or writing those query in ```class HostMetrics```

I will make changes accordingly.

amolgautam25:
I have made some changes to the schema as well. Here is the new updated schema.

```awx=# \d main_hostmetrics
                       Table "public.main_hostmetrics"
      Column      |           Type           | Collation | Nullable | Default 
------------------+--------------------------+-----------+----------+---------
 hostname         | character varying(512)   |           | not null | 
 first_automation | timestamp with time zone |           | not null | 
 last_automation  | timestamp with time zone |           | not null | 
Indexes:
    "main_hostmetrics_pkey" PRIMARY KEY, btree (hostname)
    "main_hostmetrics_hostname_da8dac94_like" btree (hostname varchar_pattern_ops)

awx=# 
```

awx/main/models/events.py
```diff
@@ -535,16 +536,19 @@ def _update_host_summary_from_stats(self, hostnames):
...
 
             Host.objects.bulk_update(list(updated_hosts), ['last_job_id', 'last_job_host_summary_id'], batch_size=100)
+            HostMetric.objects.bulk_create([HostMetric(hostname=hostname, last_automation=dt) for hostname in updated_hosts_list], ignore_conflicts=True)
```
amolgautam25: So here are two things that need to be done: 
--- If the ```hostname``` already exists in the table , update the field ```last_automation```
--- If the ```hostname``` does not already exist in the table, update these two fields ```first_automation``` and ```last_automation```. In this case both of these values ```first_automation``` and ```last_automation``` will be the same. 

Now, I have some queries on that : 
1) Am i supposed to write one query that does these 2 things at once. OR 
2) Am i supposed to write two queries ? One for bulk create and one for bulk update?


Relevant comment : https://github.com/ansible/tower/issues/4687#issuecomment-807378884

awx/main/management/commands/host_metrics.py
```diff
@@ -0,0 +1,34 @@
...
+        if opt_until:
+            until = parser.parse(opt_until)
+            until = until.replace(tzinfo=timezone.utc)
+        else:
+            until = now()
+
```
amolgautam25: Shall i set ```until``` to earliest ```now()```, or display an error message to user stating that this date is not valid ?

awx/main/management/commands/host_metrics.py
```diff
@@ -0,0 +1,34 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument('--since', dest='since', action='store', help='Start date for data')
+        parser.add_argument('--until', dest='until', action='store', help='End date for data')
+
+    def handle(self, *args, **options):
+        opt_since = options.get('since')
+        opt_until = options.get('until')
+        if opt_since:
+            since = parser.parse(opt_since)
+            since = since.replace(tzinfo=timezone.utc)
```
amolgautam25: I had to add ```tzinfo=timezone.utc``` to make the time timezone aware. The ```datetime``` field in postgres SQL is time aware.

awx/main/models/inventory.py
```diff
@@ -803,6 +803,21 @@ def _get_related_jobs(self):
         return UnifiedJob.objects.non_polymorphic().filter(Q(job__inventory=self.inventory) | Q(inventoryupdate__inventory_source__groups=self))
 
 
+class HostMetrics(models.Model):
+    hostname = models.CharField(primary_key=True, max_length=512)
+    first_automation = models.DateTimeField(auto_now_add=True, null=False)
```
chrismeyersfsu: We will want an index on this field.

awx/main/models/inventory.py
```diff
@@ -803,6 +803,21 @@ def _get_related_jobs(self):
...
+    last_automation = models.DateTimeField()
```
chrismeyersfsu: and this field.

awx/main/management/commands/host_metrics.py
chrismeyersfsu: earliest date, unix epoch. Or just don't include the `__gte` part of the query if `since` isn't set. You can pass a dict into the `.filter()`.

```
filter_kwargs = {
  "last_automation__lte": until
}
if since:
  filter_kwargs["last_automation__gte"] = since

HostMetrics.objects.filter(**filter_kwargs)
```

awx/main/management/commands/host_metrics.py
chrismeyersfsu: If the user specifies a time, and it is in the wrong format or not valid, throw an error. If the user does not specify a time, I prefer to omit the corresponding time limit from the `.filter()`

awx/main/models/events.py
chrismeyersfsu: 2 queries. 1 for bulk create 1 for bulk update.

awx/main/models/inventory.py
chrismeyersfsu: I like the queries in events.py The reason being that the events.py code is in our callback receiver code path, which needs to be performant. I'd rather have any database queries very clearly in the code path rather than "hidden" away in another class.

chrismeyersfsu (commented):
This PR has all the pieces I would expect for this feature, new model, migration, management command, and code in events.py to capture the hosts :+1:

awx/main/management/commands/host_metrics.py
jbradberry: I wouldn't do this unconditionally.  A string that gets parsed by `parser.parse` might have timezone information encoded in it, in which case the resulting datetime object will be timezone aware.  Blowing that away to become UTC would be a problem then in that case.

awx/main/management/commands/host_metrics.py
jbradberry: I like Chris's don't-include suggestion.

awx/main/models/events.py
```diff
@@ -535,16 +536,19 @@ def _update_host_summary_from_stats(self, hostnames):
             # in single queries
             host_mapping = dict((summary['host_id'], summary['id']) for summary in JobHostSummary.objects.filter(job_id=job.id).values('id', 'host_id'))
             updated_hosts = set()
+            updated_hosts_list = []
             for h in all_hosts:
                 # if the hostname *shows up* in the playbook_on_stats event
                 if h.name in hostnames:
                     h.last_job_id = job.id
                     updated_hosts.add(h)
+                    updated_hosts.append(h.name)
```
chrismeyersfsu: I'm not sure if this is where we want to get the hostname from. It's a good start. This will capture any hosts that are in the AWX inventory. However, there are potentially another set of hosts, that are never in the AWX inventory. There could be hosts that are added to Ansible's known hosts and automation ran against those. This is a common use case of Ansible. For example, use the ec2 modules to spin up a NEW vm. This host may never exist in the AWX inventory, but Ansible job events will have a record or running against this transient cloud host.

awx/main/management/commands/host_metrics.py
jbradberry: something like, `if since.tzinfo is None:`

awx/main/models/events.py
```diff
@@ -510,6 +510,7 @@ def _update_host_summary_from_stats(self, hostnames):
             job = self.job
 
             from awx.main.models import Host, JobHostSummary  # circular import
+            from awx.main.models import HostMetrics  # circular import
```
jbradberry: You should combine this with the preceding line's imports.

awx/main/models/events.py
jbradberry: Agreed.  I don't believe it's feasible to do it as a single query.  Though I think the 2nd query can and should be just a `.filter(...).update(...)`, with suitable parameters.  With `.bulk_update(...)`, you would have to already have the object instances.

jbradberry: Actually, I take that back.  It _is_ possible to do it all as one query.

awx/main/models/events.py
jbradberry: I don't believe you should be managing a new list or lists at all here.  `updated_hosts` is already a set that is getting the full `Host` objects.  In your line 551 below, you could and should be constructing your list comprehension off of that.

awx/main/management/commands/host_metrics.py
amolgautam25: I have made appropriate changes

awx/main/models/events.py
wenottingham: In analytics, I believe we pull hostnames for automation either directly from playbook_on_stats, or from the runner_on_* task events. Ideallly, we would match that here. (cc @Ladas )

awx/main/models/events.py
```diff
@@ -546,6 +546,17 @@ def _update_host_summary_from_stats(self, hostnames):
...
             Host.objects.bulk_update(list(updated_hosts), ['last_job_id', 'last_job_host_summary_id'], batch_size=100)
 
+            updated_hosts_list = list()
+            for host in updated_hosts:
+                updated_hosts_list.append(host.name)
+
+            # bulk-create
+            current_time = now()
+            HostMetrics.objects.bulk_create(
+                [HostMetrics(hostname=hostname, last_automation=current_time) for hostname in updated_hosts_list], ignore_conflicts=True
```
ryanpetrello: We might also want to take advantage of the Django ORM's support for batching these inserts *and* updates:

https://docs.djangoproject.com/en/3.1/ref/models/querysets/#django.db.models.query.QuerySet.bulk_create

`batch_size=100`, maybe?

In other places, this is what we've done:

https://github.com/ansible/awx/blob/devel/awx/main/models/events.py#L547

awx/main/models/events.py
```diff
@@ -546,6 +546,17 @@ def _update_host_summary_from_stats(self, hostnames):
 
             Host.objects.bulk_update(list(updated_hosts), ['last_job_id', 'last_job_host_summary_id'], batch_size=100)
 
+            updated_hosts_list = list()
+            for host in updated_hosts:
+                updated_hosts_list.append(host.name)
```
ryanpetrello: Do we need this data structure, or can we just do `hostname__in=updated_hosts`.

If it *must* be a `list`, perhaps we could just do:

`hostname__in=list(updated_hosts)` instead of doing the loop here?

awx/main/models/inventory.py
```diff
@@ -803,6 +803,14 @@ def _get_related_jobs(self):
         return UnifiedJob.objects.non_polymorphic().filter(Q(job__inventory=self.inventory) | Q(inventoryupdate__inventory_source__groups=self))
 
 
+class HostMetrics(models.Model):
```
ryanpetrello: This is a little nitpicky, but for models like this `Host`, `Inventory`, etc... we generally keep this name singular i.e.,

`class HostMetric` since *each one* represents _one_ metric.  I think we should follow that here for consistency.

awx/main/management/commands/host_metrics.py
```diff
@@ -0,0 +1,69 @@
...
+            until = parser.parse(opt_until)
+            if until.tzinfo is None:
+                until = until.replace(tzinfo=timezone.utc)
+
+        if (since is None) and (until is not None):
+            filter_kwargs = {"last_automation__lte": until}
+        elif (since is not None) and (until is None):
+            filter_kwargs = {"last_automation__gte": since}
+        else:
+            filter_kwargs = {"last_automation__gte": since, "last_automation__lte": until}
```
ryanpetrello: Maybe this checking of cases could simplified by just having a set of conditions and adding them individually:

```python
filter_kwargs = {}
if since is not None:
    filter_kwargs['last_automation__gte'] = since
if until is not None:
    filter_kwargs['last_automation__lte'] = until
```

awx/main/management/commands/host_metrics.py
```diff
@@ -0,0 +1,69 @@
...
+
+        result = HostMetrics.objects.filter(**filter_kwargs)
+
+        # if --json flag is set, output the result in json format
+        if options['json']:
+            list_of_queryset = list(result.values('hostname', 'first_automation', 'last_automation'))
+            for item in list_of_queryset:
+                item['first_automation'] = item['first_automation'].strftime('%Y-%m-%d')
```
ryanpetrello: If people run this on a weeks/months interval, this could end being a *very* large dataset to keep around in memory.

@wenottingham do you *care* about printing a full audit of *all* the materialized automated hosts (in tabular format), or do you want just a *total number* printed.

If we want this ability, maybe we could print the full list with an _optional_ flag to the CLI, and by default just show the count?

awx/main/models/events.py
chrismeyersfsu: We want to pull the `updated_hosts_list` from `hostnames` that is passed into the function https://github.com/ansible/awx/blob/devel/awx/main/models/events.py#L518 

This list potentially contains **more** hosts than the host summary fields list because it can contain host names that aren't in the AWX database.  Below is an add_host example you can use to test on your current version vs. a version that uses `hostnames`

```
# main.yml
- hosts: all
  gather_facts: false
  tasks:
    - name: "Do something"
      shell: ls

    - add_host:
        name: "boo1"
        ansible_connection: "local"

- hosts: all
  gather_facts: false
  tasks:

    - name: "Do something else"
      shell: ls
```

```
# Make an awx inventory that uses two hosts
[all]
fake1 ansible_connection=local
fake2 ansible_connection=local
```


```
ansible-playbook -i inventory main.yml

PLAY [all] ************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************

TASK [Do something] ***************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************
[DEPRECATION WARNING]: Distribution ubuntu 20.04 on host fake1 should use /usr/bin/python3, but is using /usr/bin/python for backward compatibility with prior Ansible releases. A future Ansible release will default to using the discovered platform python for this host. See https://docs.ansible.com/ansible/2.9/reference_appendices/interpreter_discovery.html for more information. This feature will be removed in
version 2.12. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
changed: [fake1]
[DEPRECATION WARNING]: Distribution ubuntu 20.04 on host fake2 should use /usr/bin/python3, but is using /usr/bin/python for backward compatibility with prior Ansible releases. A future Ansible release will default to using the discovered platform python for this host. See https://docs.ansible.com/ansible/2.9/reference_appendices/interpreter_discovery.html for more information. This feature will be removed in
version 2.12. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
changed: [fake2]

TASK [add_host] *******************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************
changed: [fake1]

PLAY [all] ************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************

TASK [Do something else] **********************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************
changed: [fake2]
changed: [fake1]
[DEPRECATION WARNING]: Distribution ubuntu 20.04 on host boo1 should use /usr/bin/python3, but is using /usr/bin/python for backward compatibility with prior Ansible releases. A future Ansible release will default to using the discovered platform python for this host. See https://docs.ansible.com/ansible/2.9/reference_appendices/interpreter_discovery.html for more information. This feature will be removed in
version 2.12. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
changed: [boo1]

PLAY RECAP ************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************
boo1                       : ok=1    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
fake1                      : ok=3    changed=3    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
fake2                      : ok=2    changed=2    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```
In the above output, `boo1` is the host dynamically added to the ansible runtime inventory. The "play recap" is basically equivalent to the `playbook_on_stats`.

Ladas: @ryanpetrello not sure if we're using `django-postgres-extra` anywhere, but with that we can do 1 query that will insert or update on conflict like: https://django-postgres-extra.readthedocs.io/en/master/conflict_handling.html#bulk

awx/main/models/events.py
Ladas: Right now analytics is taking the host_names from `runner_on_* task` events and I am now doing the processing of `playbook_on_stats` event. 

@wenottingham  Also we're using the job's created timestamp for timestamp of first/last automation, so it might be good to align on this too?

-----

@wenottingham @chrismeyersfsu Is there a case possible when hostname would appear in `runner_on_* task` but not in `playbook_on_stats` ? (maybe when job ends in `error` state?)

awx/main/models/events.py
ryanpetrello: 😉 

https://github.com/ansible/tower/issues/4687#issuecomment-807375828

We decided to just take this approach instead of introducing a new dependency.

awx/main/models/inventory.py
amolgautam25: I have changed it to ```HostMetric```

awx/main/models/events.py
amolgautam25: ```hostname__in``` has to be list of strings. It has to be list of string because the schema has a column ```hostname``` which is a string. From what my understanding is , i can only filter it against list of strings. 

```hostname__in=list(updated_hosts)``` would make ```hostname__in``` a list of ```Host``` ,and cant be used to filter the table.

File: awx/main/management/commands/host_metrics.py (hunk 17 above)
Comment (amolgautam25) [reply awx/main/management/commands/host_metrics.py]: I have made the necessary changes.

File: awx/main/models/events.py (hunk 14 above)
Comment (amolgautam25) [reply awx/main/models/events.py]: I have made the required changes.

Comment (jbradberry) [reply awx/main/models/events.py]: @Ladas we are not.  `django-postgres-extra` came up in another context recently, but we decided against bringing it in as a new dependency.

File: awx/main/management/commands/host_metrics.py (hunk 18 above)
Comment (wenottingham) [reply awx/main/management/commands/host_metrics.py]: A customer may need to/want to reconcile certain hosts that are "decommissioned on January 31st" vs just "not automated on since January 31st"

To do that, we would need to dump the full list. It doesn't need to be the default, but it does need to at least be an option.

**Comment by amolgautam25** [comment]
recheck

File: awx/main/models/events.py (hunk 14 above)
Comment (amolgautam25) [reply awx/main/models/events.py]: I have made the required changes

File: awx/main/models/events.py (hunk 12 above)
Comment (chrismeyersfsu) [reply awx/main/models/events.py]: Yeah @Ladas from my observations that would be one case. Or maybe even a `failed` case if the Tower system crashes in the middle of a run, the reaper reaps the Job, and the `playbook_on_stats` event never exists for that job.

Maybe another case would be through some fancy use of Ansible `delegate_to` ?

File: awx/main/models/events.py [hunk 19]
```diff
@@ -546,6 +548,13 @@ def _update_host_summary_from_stats(self, hostnames):
 
             Host.objects.bulk_update(list(updated_hosts), ['last_job_id', 'last_job_host_summary_id'], batch_size=100)
 
+            # bulk-create
```
ryanpetrello, line 552: This looks about right to me.

A good place to add some tests for this logic would be:

`awx/main/tests/functional/models/test_events.py`

There's similar tests in here that validate the creation of the `JobHostSummary` objects.

amolgautam25: I have made the required changes. I have added a unit test case in the file : 
```awx/main/tests/functional/models/test_host_metric.py```

**Comment by amolgautam25** [comment]
recheck

**Comment by amolgautam25** [comment]
recheck

File: awx/main/management/commands/host_metric.py [hunk 20]
```diff
@@ -0,0 +1,66 @@
...
+from awx.main.models.inventory import HostMetric
+import json
+
+
+class Command(BaseCommand):
+
+    help = 'This is for offline licensing usage'
+
+    def add_arguments(self, parser):
+        parser.add_argument('--since', dest='since', action='store', help='Start date for data')
```
ryanpetrello: Now that we're on Python3.8, you could probably do something like this instead of relying on `dateutil.parse`.  It should also warn about things that are passed in that aren't in the correct format:

`parser.add_argument('--since', type=datetime.date.fromisoformat)`

If you do it this way, you can probably change the code in `def handle()` to just call `.replace(tzinfo=...)` as you're doing now with the assumption that the inputs to the function are already date objects.

ryanpetrello:
```~ python test.py
{'since': None}
~ python test.py --since 2021-01-01
{'since': datetime.date(2021, 1, 1)}
~ python test.py --since 2021-01-dog
usage: test.py [-h] [--since SINCE]
test.py: error: argument --since: invalid fromisoformat value: '2021-01-dog'```

File: awx/main/management/commands/host_metric.py [hunk 21]
```diff
@@ -0,0 +1,66 @@
...
+        opt_until = options.get('until')
+
+        # user gives no arguments
+        if opt_since is None and opt_until is None:
+            print("No Arguments received")
+            return None
+
+        since = None
+        if opt_since:
+            since = parser.parse(opt_since)
```
ryanpetrello: This current approach of just calling `parser.parse` *probably* throws an exception if you pass in a string that doesn't look like a date.  By using the approach described above, you'll should get error messages like the one I described for free.

awx/main/management/commands/host_metric.py
```diff
@@ -0,0 +1,66 @@
...
+        if until is not None:
+            filter_kwargs['last_automation__lte'] = until
+
+        result = HostMetric.objects.filter(**filter_kwargs)
+
+        # if --json flag is set, output the result in json format
+        if options['json']:
+            list_of_queryset = list(result.values('hostname', 'first_automation', 'last_automation'))
+            for item in list_of_queryset:
+                item['first_automation'] = item['first_automation'].strftime('%Y-%m-%d')
```
ryanpetrello: If you're doing this because the date objects aren't JSON-serializable, there's another way to accomplish this by providing `json.dumps()` a hint about how to serialize dates (that way you don't have to loop over this data structure and replace the values in-place):

```python
from django.core.serializers.json import DjangoJSONEncoder
if options['json']:
    print(
        json.dumps(
            result.values('hostname', 'first_automation', 'last_automation')
        ), cls=DjangoJSONEncoder
    )
```

https://docs.djangoproject.com/en/3.2/topics/serialization/#djangojsonencoder

awx/main/models/inventory.py
```diff
@@ -803,6 +803,14 @@ def _get_related_jobs(self):
         return UnifiedJob.objects.non_polymorphic().filter(Q(job__inventory=self.inventory) | Q(inventoryupdate__inventory_source__groups=self))
 
 
+class HostMetric(models.Model):
+    hostname = models.CharField(primary_key=True, max_length=512)
+    first_automation = models.DateTimeField(
+        auto_now_add=True, null=False, db_index=True, help_text=_('DateTime ( with timestamp ) , when the object was first created')
```
ryanpetrello: `When the object was first created` -> `When the host was first automated against.`
`When the object was last modified` -> `When the host was last automated against.`

awx/main/management/commands/host_metric.py
```diff
@@ -0,0 +1,66 @@
...
+                item['last_automation'] = item['last_automation'].strftime('%Y-%m-%d')
+
+            json_result = json.dumps(list_of_queryset)
+            print(json_result)
+
+        # --json flag is not set, output in plain text
+        else:
+            print("Total Number of hosts automated : {hosts_automated}".format(hosts_automated=len(result)))
```
ryanpetrello: This isn't _necessary_, but just as an FYI, a really nice feature of newer Pythons is `f-strings`, which make this sort of string templating a lot less verbose than `.format()`:

```python
print(f"Total Number of hosts automated: {len(result)}")
```

```python
>>> result = range(10)
>>> print(f"Total Number of hosts automated: {len(result)}")
Total Number of hosts automated: 10
```

awx/main/tests/functional/models/test_host_metric.py
```diff
@@ -0,0 +1,22 @@
...
+    # 1) if all the objects were created or not
+    assert HostMetric.objects.count() == len(hostnames)
+
+    # 2) Match the hostnames stored in DB with the one passed in bulk_create
+    assert sorted([s.hostname for s in HostMetric.objects.all()]) == sorted(hostnames)
+
+    # 3) Make sure that first_automation attribute is today's date
+    date_today = now().strftime('%Y-%m-%d')
+    result = HostMetric.objects.filter(first_automation__startswith=date_today).count()
+    assert result == len(hostnames)
```
ryanpetrello, line 22: Nice, I like this test!

ryanpetrello (approved):
Left a few minor comments, but otherwise this is looking great.

Nice work!

awx/main/models/inventory.py
ryanpetrello: Out of my feedback, this is probably the one item I feel strongly about addressing before merging.

awx/main/management/commands/host_metric.py
amolgautam25: I have made the changes.

awx/main/management/commands/host_metric.py
amolgautam25: I have made the required changes.

awx/main/models/inventory.py
amolgautam25: I have made the required changes

awx/main/management/commands/host_metric.py
amolgautam25: I am doing this because I was wondering Customer would not want that second level granularity. :sweat_smile:.
I think he would look at dates ```"last_automation": "2021-04-13"```  rather than ```"last_automation": "2021-04-13T16:08:56.841Z"``` in the JSON

amolgautam25: I have made the changes but let me know if i need to revert back.

awx/main/models/events.py
```diff
@@ -546,6 +549,13 @@ def _update_host_summary_from_stats(self, hostnames):
...
+            current_time = now()
```
Ladas, line 553: @wenottingham @ryanpetrello  so just to verify, we're using time `now()` here, which will lead to slightly different numbers than in analytics, is that ok?

In analytics, we're currently exposing `job_created` timestamp, then timestamp of first/last finished task on host and then timestamp of first/last task with changed=True on the host.

jbradberry: @Ladas this version of `now()` is timezone aware, not naive.  I would think it should be fine, unless analytics itself is not using aware timestamps.

jbradberry: ... unless I've misunderstood the objection, which upon second reading I think I may have.

awx/main/models/__init__.py
```diff
@@ -12,7 +12,16 @@
...
+from awx.main.models.inventory import (  # noqa
+    CustomInventoryScript,
+    Group,
+    Host,
+    HostMetric,
+    Inventory,
+    InventorySource,
+    InventoryUpdate,
+    SmartInventoryMembership,
+)  # noqa
```
chrismeyersfsu: Remove this `# noqa`, the `# noqa` on line 15 does what is wanted.

amolgautam25: I have made the changes.

awx/main/models/events.py
Ladas: @jbradberry right, yeah I am basically saying this list will produce different timestamps than when you'll fetch this list from AA. (so trying to confirm, if the listing of the hosts like this from Tower and AA should produce identical lists or not :-) )

In here we're taking `now()` somewhere around processing of playbook on stats, right?

In AA we have multiple timestamps, but always taking the existing `created` timestamps of job/events, so not creating another timestamp.

ryanpetrello: I'm interested in how much @wenottingham cares about this being an exact behavioral match.  My understanding is that this feature solely exists as an alternative offline reporting mechanism for folks who *don't* use AA and want an idea of total automation count (unique hosts) over a date range.

ryanpetrello: If we wanted to be *super* particular or improve this slightly, _maybe_ instead of using `now()` we could just use `self.created` (given that `self` is a `playbook_on_stats` `JobEvent`).

wenottingham: Ideally, they would match. 

In the long run, it probably doesn't matter too much whether or not they do, as the difference over the course of the job isn't that meaningful. This won't catch playbooks that abort before the end, and the AA approach will, as I understand it, which is another differene.

Ladas: @ryanpetrello `created` timestamp of the `playbook_on_stats` event should be close enough to the `max(created)` of host tasks we're storing, so that should be ok I think. (we're storing timestamp of last task per host, so it might differ more for hosts processed early and later)

appuk:
tested the command, everything works as expected. also ran a slowyo and there's no regression. Amol is going to open up a PR for adding test to tower-qa
@amolgautam25 looks good for me to merge

amolgautam25:
I have added the relevant test in ```tower-qa``` PR : https://github.com/ansible/tower-qa/pull/6237

amolgautam25:
recheck

jakemcdermott:
recheck

amolgautam25:
recheck

jbradberry:
@wenottingham @Ladas how do we feel about this now?  Looking at the code, it looks ready to me.

amolgautam25:
Blocker for this PR : 

- [x] Waiting for performance tests to be fixed
- [x] Have a performance baseline after DB partition PR is merged 
- [x] Have another performance baseline for this PR
- [x] If there is no major performance penalty , then we are good to merge the PR

awx/main/models/events.py
```diff
@@ -510,12 +510,15 @@ def _update_host_summary_from_stats(self, hostnames):
...
             from awx.main.models import Host, JobHostSummary  # circular import
+            from awx.main.models import Host, JobHostSummary, HostMetric
 
             all_hosts = Host.objects.filter(pk__in=self.host_map.values()).only('id')
             existing_host_ids = set(h.id for h in all_hosts)
 
             summaries = dict()
+            updated_hosts_list = list()
             for host in hostnames:
+                updated_hosts_list.append(host)
```
AlanCoding, line 521: how is `updated_hosts_list` any different from `hostnames`?

awx/main/models/events.py
```diff
@@ -546,6 +549,13 @@ def _update_host_summary_from_stats(self, hostnames):
 
             Host.objects.bulk_update(list(updated_hosts), ['last_job_id', 'last_job_host_summary_id'], batch_size=100)
 
+            # bulk-create
+            current_time = now()
+            HostMetric.objects.bulk_create(
+                [HostMetric(hostname=hostname, last_automation=current_time) for hostname in updated_hosts_list], ignore_conflicts=True, batch_size=100
+            )
+            HostMetric.objects.filter(hostname__in=updated_hosts_list).update(last_automation=current_time)
```
AlanCoding, line 557: why is this bulk update needed and what's a metric for how many rows it will touch

awx/main/models/events.py
AlanCoding: I'm increasingly thinking that this shouldn't be needed.

awx/main/models/events.py
AlanCoding: Answering my own question, there should be as many updates as the number of hosts the job affected, because of the `ignore_conflicts=True` of the bulk_create.... and I think the intent is that these are truly unique on hostname.

awx/main/models/events.py
chrismeyersfsu: I think because `hostnames` comes in as a `set`

chrismeyersfsu (approved):

awx/main/models/events.py
AlanCoding: The Django ORM is just as happy with a `set` as a `list` (and the ordering is equally arbitrary in both cases). The introduction of `updated_hosts_list` should be removed, and the later use of it should simply be replaced with `hostnames`.

Issue #9636, opened 2021-03-19
Editing a credential for moment will display the id of credential type


nixocio:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Editing a credential for moment will display the id of credential type, then display the proper name.

https://user-images.githubusercontent.com/9053044/111824737-f6285680-88bc-11eb-8b28-4693aeb125d1.mov

tiagodread:
@jakemcdermott @nixocio I tried this on devel on dev or production build and with network throotling enabled/disabled and I couldn't reproduce this issue 

![chrome-capture](https://user-images.githubusercontent.com/12464858/113922332-8cb1ae80-97bd-11eb-926b-6347b93746b9.gif)

nixocio:
This issue was already addressed as part of https://github.com/ansible/awx/pull/9662 - that is probably why we can not recreate anymore on devel.

Issue #8828, opened 2020-12-10
[ui_next] User should not be able to change credential types for existing credentials


unlikelyzero:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Generally, giving the user the ability to switch between credential types after a credential is created causes lots of complexity with very little benefit

##### ENVIRONMENT
* AWX version: 12/10

##### STEPS TO REPRODUCE
1. Create a credential
2. Modify credential type
Note: dragons and form errors

##### EXPECTED RESULTS

I think we should remove this functionality until we can map out the intent.

##### ACTUAL RESULTS
Dragons

mabashian:
_yes_

unlikelyzero:
@wenottingham i think we should prioritize this for the next release so that we can dramatically reduce the complexity in testing this

yagomarques:
![image](https://user-images.githubusercontent.com/18562328/114591798-a12ff400-9c60-11eb-9d95-439882e55bec.png)
it's working, the field are locked to modify.

PR #9662, opened 2021-03-23
Fix cred type display on form initialization

jakemcdermott:
##### SUMMARY
For #9636 and https://github.com/ansible/awx/issues/8828

### DIFF
--- a/awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx
+++ b/awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx
@@ -1,4 +1,5 @@
 import React, { useCallback, useEffect, useState } from 'react';
+import { useLocation } from 'react-router-dom';
 import { shape } from 'prop-types';
 import { Formik, useField, useFormikContext } from 'formik';
 import { withI18n } from '@lingui/react';
@@ -11,6 +12,7 @@ import {
   Select as PFSelect,
   SelectOption as PFSelectOption,
   SelectVariant,
+  Tooltip,
 } from '@patternfly/react-core';
 import styled from 'styled-components';
 import FormField, { FormSubmitError } from '../../../components/FormField';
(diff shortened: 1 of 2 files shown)


nixocio:
hey @jakemcdermott, I got this code locally and I noticed the following.

1 - Editing a credential type it will trigger the form validation.

image

2 - Attempting to save a edited credential type.

image

On devel the the form validation is not triggered once the credential type is modified, and the following error is raised once one attempt to save a modified credential type.

save_validation

jakemcdermott:
@nixocio 
> 2 - Attempting to save a edited credential type.

This should be fixed in the latest commit - can you try it again with the latest? I'll check on the other two.

nixocio:
> @nixocio
> 
> > 2 - Attempting to save a edited credential type.
> 
> This should be fixed in the latest commit - can you try it again with the latest? I'll check on the other two.

I just tested this. It is fixed.

unlikelyzero:
just visual tests on this one

jakemcdermott:
cc @nixocio @unlikelyzero 

re: issues 1 and 3 mentioned [here](https://github.com/ansible/awx/pull/9662#issuecomment-804886235)

I added a commit to this PR that disables the credential-type select on the credential edit forms. Allowing the field value to be changed invites a lot of complexity and bugs. Moreover, there doesn't seem to be a good reason to let users change it in the first place given that [you can't actually modify a credential's type after it's been created.](https://github.com/ansible/awx/blob/devel/awx/api/serializers.py#L2675) Users will _always_ get an error when they try to do this

Screenshot from 2021-03-24 12-24-13

awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx
```diff
@@ -27,6 +29,7 @@ const Select = styled(PFSelect)`
   ul {
     max-width: 495px;
   }
+  ${props => (props.isDisabled ? `cursor: not-allowed` : null)}
```
nixocio, line 32: 👍

nixocio (approved):
I like this new UX! and the reported issue is fixed.

unlikelyzero:
We need a single new visual test for the hover-over disabled 'edit' component

jakemcdermott:
regate

PR #9664, opened 2021-03-23
Fix service reference error for cred plugin dev env

jakemcdermott:
##### SUMMARY
Intended to fix the error I see when running `COMPOSE_TAG=devel make docker-compose-credential-plugins`:

Screenshot from 2021-03-23 08-30-24


ryanpetrello (approved):
Thanks @jakemcdermott

Issue #4539, opened 2019-08-20
Cannot create a custom notification message with {{ job.extra_vars }} in message


unlikelyzero:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Cannot create a custom notification message with {{ job.extra_vars }} in message

##### STEPS TO REPRODUCE
1. Attempt to create a custom notification Start Message with `{{ job.extra_vars }}` in the body

##### EXPECTED RESULTS
Can save and use NT

##### ACTUAL RESULTS
400 Error on Save

##### ADDITIONAL INFORMATION
![Screen Shot 2019-08-20 at 7 14 03 PM](https://user-images.githubusercontent.com/12446869/63390917-b155da00-c37e-11e9-8109-384e7758b70f.png)

unlikelyzero:
![Screen Shot 2019-08-20 at 7 15 28 PM](https://user-images.githubusercontent.com/12446869/63390986-eb26e080-c37e-11e9-963f-f996c7ff3dfb.png)

jladdjr:
@wenottingham - any concerns about extra_vars potentially including sensitive data? If not, we can add this back in

wenottingham:
Extra vars includes password survey types. As long as we are returning the API-view where these have been replaced by "$encrypted$", then it should be OK.

ryanpetrello:
Using `job.extra_vars` will _likely_ contain things like `$encrypted$UTF8$AESCBC$<nonsense>`.

jladdjr:
While being able to show survey values sounds interesting, I've been trying to stay away from anything that would surface sensitive information (even an encrypted or hashed version). I'd vote to not surface this unless we've got some strong reasons / interest in doing so. We _could_ try to strip out sensitive info, but I think it'd be hard to come up with a filter that was guaranteed to catch everything, which makes me nervous.

TLDR: Vote to close b/c of potential to surface sensitive info.

ryanpetrello:
https://github.com/ansible/awx/issues/4539#issuecomment-523321867

I agree with this on the first pass.  If we want to do this at some point, I'd rather treat this as an additional enhancement (after all, we never promised the templated notification feature would give you access to *every* field on a job, just many of the useful ones so you can craft a meaningful custom message), and not a requirement to merge https://github.com/ansible/awx/pull/4291

I'd also consider this a low-ish priority in relation to other open issues in the current release.

unlikelyzero:
[...]

PR #9668, opened 2021-03-23
surface jobs.extra_vars in custom notification templates

jladdjr:
We're already surfacing (a filtered version of) extra_vars in `{{ job_metadata }}` for webhooks. This surfaces the same information in `job.extra_vars`.

Related:
https://github.com/ansible/awx/issues/4539

---

Given job with extra vars set to:
```
---
key_for_value: simple_value
key_for_list:
- list_item1
- list_item2
- list_item3
key_for_dict:
  subkey1: subvalue1
  subkey2: subvalue2
  subkey3: subvalue3
```

And custom NT with message set to:
`Job #{{ job.id }} extra vars: {{ job.extra_vars }}`

Delivers notification:
`Job #15 extra vars: {"key_for_value": "simple_value", "key_for_list": ["list_item1", "list_item2", "list_item3"], "key_for_dict": {"subkey1": "subvalue1", "subkey2": "subvalue2", "subkey3": "subvalue3"}}`

### DIFF
--- a/awx/main/tests/functional/models/test_notifications.py
+++ b/awx/main/tests/functional/models/test_notifications.py
@@ -120,7 +120,6 @@ class TestJobNotificationMixin(object):
     @pytest.mark.django_db
     def test_schedule_context(self, job_template, admin_user):
         schedule = Schedule.objects.create(name='job-schedule', rrule='DTSTART:20171129T155939z\nFREQ=MONTHLY', unified_job_template=job_template)
-        kwargs = {}
         job = Job.objects.create(name='fake-job', launch_type='workflow', schedule=schedule, job_template=job_template, extra_vars={'foo': 'bar'})
 
         job_serialization = UnifiedJobSerializer(job).to_representation(job)
--- a/awx/main/tests/functional/models/test_notifications.py
+++ b/awx/main/tests/functional/models/test_notifications.py
@@ -35,6 +35,7 @@ class TestJobNotificationMixin(object):
             'launch_type': str,
             'limit': str,
             'modified': datetime.datetime,
+            'extra_vars': dict,
             'name': str,
             'playbook': str,
             'scm_branch': str,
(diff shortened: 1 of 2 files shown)


AlanCoding:
I think it would be good to clarify behavior in https://github.com/ansible/awx/blob/devel/docs/notification_system.md

But given this wording

> The notification message can include templated fields written using Jinja templates. The templates may reference a set of white-listed fields found in the associated job's serialization.

With the most honest reading of that documentation, this patch looks right.

jladdjr:
@AlanCoding - what kind of clarification were you thinking? mentioning that extra_vars gets filtered?

jladdjr:
Tried `nt.messages = {'started': {'message': '{{ job.extra_vars.foo }}'}}` and got a traceback when the job ran; there's more work to do here

awx/main/tests/functional/models/test_notifications.py
```diff
@@ -117,7 +120,8 @@ def test_context(self, JobClass, sqlite_copy_expert, project, inventory_source):
     @pytest.mark.django_db
     def test_schedule_context(self, job_template, admin_user):
         schedule = Schedule.objects.create(name='job-schedule', rrule='DTSTART:20171129T155939z\nFREQ=MONTHLY', unified_job_template=job_template)
-        job = Job.objects.create(name='fake-job', launch_type='workflow', schedule=schedule, job_template=job_template)
+        kwargs = {}
```
AlanCoding: I'm not sure why kwargs is set back to empty dict here

jladdjr: neither am I, deleting

AlanCoding:
@jladdjr is this ready for another look, or does it still need some sort of revision?

jladdjr:
@AlanCoding still needs revision - have been working on other bugs, but thanks for the reminder, would love to circle back to this

PR #9670, opened 2021-03-23
remove unnecessary references to bwrap, bubblewrap, and proot

ryanpetrello:


awx/ui_next/src/screens/Setting/Jobs/JobsEdit/data.defaultJobSettings.json
```diff
@@ -28,7 +28,6 @@
   "AWX_ISOLATED_HOST_KEY_CHECKING": false,
   "AWX_ISOLATED_LAUNCH_TIMEOUT": 600,
   "AWX_PROOT_BASE_PATH": "/tmp",
-  "AWX_PROOT_ENABLED": true,
```
jakemcdermott, line 31: edit: ([I think?](https://github.com/ansible/awx/pull/9670/files#r599948754)) There's more to delete in
```
screens/Host/data.hostFacts.json
screens/Job/shared/data.job.json
screens/Inventory/shared/data.hostFacts.json
screens/Setting/shared/data.allSettings.json
screens/Setting/shared/data.jobSettings.json
screens/Setting/shared/data.allSettingOptions.json
screens/Setting/Jobs/JobsEdit/data.defaultJobSettings.json
screens/Setting/Jobs/JobsEdit/JobsEdit.jsx
```

awx/main/conf.py
```diff
@@ -232,15 +232,6 @@
...
-    'AWX_PROOT_ENABLED',
-    field_class=fields.BooleanField,
-    label=_('Enable job isolation'),
-    help_text=_('Isolates an Ansible job from protected parts of the system to prevent exposing sensitive information.'),
-    category=_('Jobs'),
-    category_slug='jobs',
-)
-
 register(
     'AWX_PROOT_BASE_PATH',
```
jakemcdermott: Should we delete `AWX_PROOT_BASE_PATH` and anything else with `PROOT`, too?

ryanpetrello: @jakemcdermott I think *technically* we still lay these files down in `/tmp` before runner is invoked, and there may be cases where customers want to put them on another volume.  So I suspect these settings - while slightly esoteric - probably should continue to exist for now.

cc @shanemcd

jakemcdermott: Sounds good. Please check the [list](https://github.com/ansible/awx/pull/9670#discussion_r599948246) above if anything else needs deleting from the ui.

wenottingham: Does the container code still use SHOW_PATHS to map volumes from the host into the container?

docs/process_isolation.md
```diff
@@ -4,14 +4,9 @@ In older versions of Ansible Tower, we used a system called `proot` to isolate T
...
-Tower 3.5 and later uses the process isolation feature in Ansible runner to achieve process isolation.
+Tower 3.5 through 3.8 use the process isolation feature in Ansible runner to achieve process isolation.
 
-
-### Activating Process Isolation
-
-`bubblewrap` is enabled by default; it can be turned off via Tower Config or from a Tower settings file:
-
-    AWX_PROOT_ENABLED = False
+More modern versions of AWX and Tower achieve isolation by running every playbook in a container (which cannot be disabled).
```
wenottingham: This maybe could be just rewriting/removing this docs file entirely.

awx/main/conf.py
ryanpetrello: Yes:

https://github.com/ansible/awx/blob/devel/awx/main/tasks.py#L911-L914

We could probably get rid of `HIDE_PATHS`, though, because...it's a container (we don't have to itemize all the stuff we want to hide).

awx/conf/migrations/0009_rename_proot_settings.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+from django.db import migrations
+from awx.conf.migrations import _rename_setting
+
+
+def rename_proot_settings(apps, schema_editor):
+    _rename_setting.rename_setting(apps, schema_editor, old_key='AWX_PROOT_BASE_PATH', new_key='AWX_ISOLATION_BASE_PATH')
```
shanemcd, line 8: I don't think we actually use this anywhere. Can we just delete this setting?

shanemcd: Never mind... I read the rest of the PR.

But now I'm wondering if this can get rolled into the isolated code removal and we just ditch this setting all together. I guess it doesn't really matter.

ryanpetrello: Ignoring isolated nodes, we *do* still lay down a private data dir on the controller-side file system (in `/tmp`) *before* jobs run, right?

ryanpetrello:
@wenottingham to answer your question, the `AWX_ISOLATION_SHOW_PATHS` **does** still work:

image

Moments after adding `/tmp` to the list `AWX_ISOLATION_SHOW_PATHS`:

image
image

shanemcd (approved):

PR #9671, opened 2021-03-23
Fix api/v2/metrics data displaying incorrect value

fosterseth:
### How to reproduce bug
1. Disable all instances
2. Queue up 5 jobs, (if using same JT, enable concurrent jobs)
3. Enable instance so jobs will start running
4. Refresh `/api/v2/metrics` endpoint. You should see a metric that says `awx_status_total{status="pending"} 5.0`. Once you see this, don't refresh again.
5. Wait for all jobs to finish.
6. Now start refreshing the `api/v2/metrics` endpoint. Every once in a while you will see this pending jobs metric show (`awx_status_total{status="pending"} 5.0`) even though there are no jobs in pending state

### Fix
Use a locally defined prometheus registry instead of a global registry. Each time the endpoint is refreshed, a new, local registry is set up with prometheus objects (Gauge, Info). Since we aren't actually incrementing these metrics across api calls, we should be safe. That is to say, we just lookup the values from the database and set the prometheus values explicitly.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 18.0.0
```


ryanpetrello (approved):

fosterseth:
regate

Issue #7066, opened 2020-05-18
Determine how to handle registry authentication for execution environments


shanemcd:
This is a subtask of a larger enhancement/initiative surrounding execution environments: #5157

Execution Environments may require a credential to pull from a protected registry. Different execution environments may come from different registries. We need a way to handle the authentication process that does not involve a shared global config. Whatever we do should be secure in terms of file permissions.

ryanpetrello:
This work will likely encompass new data model and API work to define credential types for interacting with registries, and associating the necessary credentials to defined execution environments (see: https://github.com/ansible/awx/issues/7064)

kdelee:
Right now errors when pulling images for ContainerGroups can be rather opaque/difficult to realize that is the problem. When we switched to a registry for default images downstream this came up https://github.com/ansible/tower/issues/4402.

Would be good to figure out a less opaque way of surfacing problems when pulling images.

AlanCoding:
@jbradberry https://github.com/ansible/awx/pull/8113 seems like it's starting toward this. It defines some credential inputs, but I don't know if it sets up injectors? And we're a ways away from this being testable I think

AlanCoding:
This is looking more and more difficult to me. The `podman` executable on traditional installs is invoked by `ansible-runner worker`. That means that we need to send the credentials over receptor before we can apply them. These credentials are not for the `ansible-playbook` process, so that means it needs to be a part of the ansible-runner interface. It can't just be added onto the environment variables for the job.

Then, the OCP deployments will need its own way to handle this.

AlanCoding:
I would start with quay.io (which may also work the same for dockerhub).

Although tokens would be preferable, https://docs.quay.io/api/ establishes that the pattern may be the same as username/pass, with "$oauthtoken" as the username.

It may be expecting this gets written to `~/.docker/config.json`, and how they can be implemented in the credential type injector, need to get pinned down.

kdelee:
@rebeccahhh @shanemcd whats the status on this? I know at this point we can associate a credential with an EE, but I never got caught up on whether or not we are applying the credential to the actual container pull step

ryanpetrello:
`state:needs_devel` is accurate here (I don't think the work has started yet to hook the credential into the container pull step)

kdelee:
@yagomarques @rebeccahhh some things to try out during development
[...]

PR #3257, opened 2019-02-19
define native CredentialType inputs/injectors in code, not in the DB

ryanpetrello:
This has a few benefits:

1.  It makes adding new fields to built-in CredentialTypes _much_
    simpler.  In the past, we've had to write a migration every time we
    want to modify an existing type (changing a label/help text,
    changing options like the recent become_method changes) or
    when adding a new field entirely

2.  It paves the way for third party credential plugins support, where
    importable libraries will define their own source code-based schema

PR #9683, opened 2021-03-24
Pull with credentials from protected registries

rebeccahhh:
##### SUMMARY

relates to #7066
if a credential is associated with an EE this will create a JSON authfile that is then passed with the pull request to the host of the registry

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 18.0.0
```

----
#### TODOs
- [x] Remove separate `token` field from the registry credential. 
- [x] Rename the existing password field to say "password/token"
- [x] Ensure only registry credentials can be associated with an EE https://github.com/ansible/awx/issues/9628
- [x] Write out the `auth.json` file to the `pdd_wrapper_` directory. https://github.com/ansible/awx/pull/9683#issuecomment-808254962
- [x] Use secure permissions for `auth.json` https://github.com/ansible/awx/pull/9683#discussion_r600645653

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -1412,6 +1412,11 @@ class ExecutionEnvironmentSerializer(BaseSerializer):
             res['credential'] = self.reverse('api:credential_detail', kwargs={'pk': obj.credential.pk})
         return res
 
+    def validate_credential(self, value):
+        if value and value.kind != 'registry':
+            raise serializers.ValidationError(_('Only Container Registry credentials can be associated with an Execution Environment'))
+        return value
+
     def validate(self, attrs):
         # prevent changing organization of ee. Unsetting (change to null) is allowed
         if self.instance:
--- a/awx/main/models/credential/__init__.py
+++ b/awx/main/models/credential/__init__.py
@@ -1106,7 +1106,7 @@ ManagedCredentialType(
             },
             {
                 'id': 'password',
-                'label': ugettext_noop('Password'),
+                'label': ugettext_noop('Password or Token'),
                 'type': 'string',
                 'secret': True,
                 'help_text': ugettext_noop('A password or token used to authenticate with'),
--- a/awx/main/models/credential/__init__.py
+++ b/awx/main/models/credential/__init__.py
@@ -297,8 +297,11 @@ class Credential(PasswordFieldsModel, CommonModelNameNotUnique, ResourceMixin):
 
     def has_inputs(self, field_names=()):
         for name in field_names:
-            if name not in self.inputs:
-                return False
+            if name in self.inputs:
+                if self.inputs[name] in ('', None):
+                    return False
+            else:
+                raise ValueError('{} is not an input field'.format(name))
         return True
 
     def _get_dynamic_input(self, field_name):
--- a/awx/main/models/credential/__init__.py
+++ b/awx/main/models/credential/__init__.py
@@ -295,6 +295,12 @@ class Credential(PasswordFieldsModel, CommonModelNameNotUnique, ResourceMixin):
             return True
         return field_name in self.inputs and self.inputs[field_name] not in ('', None)
 
+    def has_inputs(self, field_names=()):
+        for name in field_names:
+            if name not in self.inputs:
+                return False
+        return True
+
     def _get_dynamic_input(self, field_name):
         for input_source in self.input_sources.all():
             if input_source.input_field_name == field_name:
@@ -1096,11 +1102,11 @@ ManagedCredentialType(
                 'type': 'string',
             },
             {
-                'id': 'password/token',
-                'label': ugettext_noop('Password/Token'),
+                'id': 'password',
+                'label': ugettext_noop('Password'),
                 'type': 'string',
                 'secret': True,
-                'help_text': ugettext_noop('A token to use to authenticate with. ' 'This should not be set if username/password are being used.'),
+                'help_text': ugettext_noop('A password or token used to authenticate with'),
             },
         ],
         'required': ['host'],
--- a/awx/main/models/credential/__init__.py
+++ b/awx/main/models/credential/__init__.py
@@ -1096,14 +1096,8 @@ ManagedCredentialType(
                 'type': 'string',
             },
             {
-                'id': 'password',
-                'label': ugettext_noop('Password'),
-                'type': 'string',
-                'secret': True,
-            },
-            {
-                'id': 'token',
-                'label': ugettext_noop('Access Token'),
+                'id': 'password/token',
+                'label': ugettext_noop('Password/Token'),
                 'type': 'string',
                 'secret': True,
                 'help_text': ugettext_noop('A token to use to authenticate with. ' 'This should not be set if username/password are being used.'),
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -834,7 +834,7 @@ class BaseTask(object):
         """
         return os.path.abspath(os.path.join(os.path.dirname(__file__), *args))
 
-    def build_execution_environment_params(self, instance):
+    def build_execution_environment_params(self, instance, private_data_dir):
         if settings.IS_K8S:
             return {}
 
@@ -854,7 +854,7 @@ class BaseTask(object):
         if instance.execution_environment.credential:
             cred = instance.execution_environment.credential
             if cred.has_inputs(field_names=('host', 'username', 'password')):
-                path = self.build_private_data_dir(instance)
+                path = os.path.split(private_data_dir)[0]
                 with open(path + '/auth.json', 'w') as authfile:
                     host = cred.get_input('host')
                     username = cred.get_input('username')
@@ -866,7 +866,7 @@ class BaseTask(object):
                 os.chmod(authfile.name, stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR)
                 params["container_options"].append(f'--authfile={authfile.name}')
             else:
-                logger.exception('Please recheck that your host, username, and password fields are all filled.')
+                raise RuntimeError('Please recheck that your host, username, and password fields are all filled.')
 
         pull = instance.execution_environment.pull
         if pull:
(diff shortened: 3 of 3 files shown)


awx/main/tasks.py
```diff
@@ -904,6 +904,16 @@ def build_execution_environment_params(self, instance):
             "container_options": ['--user=root'],
         }
 
+        if instance.execution_environment.credential:
+            with open('/tmp/auth.json', 'w') as authfile:
+                image = instance.execution_environment.image
```
ryanpetrello: We should probably explicitly do something like this so this file isn't readable by non-awx users (as it contains secrets).

```python
os.chmod(authfile.name, stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR)
```

awx/main/tasks.py
```diff
@@ -904,6 +904,16 @@ def build_execution_environment_params(self, instance):
...
+                credentials = instance.execution_environment.credential.inputs
```
ryanpetrello: `.inputs` contains encrypted values:

```python
In [2]: cred = Credential.objects.get(pk=4)

In [3]: cred.inputs
Out[3]:
{'host': 'https://quay.io/foo/bar',
 'password': '$encrypted$UTF8$AESCBC$Z0FBQUFBQmdXMmV3YkhuSjVYajhLdzRDOENpZDZZb2o5S2lvc2JkR2VwR1hrTl9ub3ZlajhhZkx2RXRyVDhGbHJ4N2c3eE9QN1MwSjFlNnpuWUNGYi1hMmQySklMSkJOSnc9PQ==',
 'username': 'ryan'}
```

You might want to explicitly build a dictionary and call `.get_input()`, which will perform the decryption for the password:

```python
In [4]: {
   ...:     'username': cred.get_input('username'),
   ...:     'password': cred.get_input('password'),
   ...: }
Out[4]: {'password': 'testing', 'username': 'ryan'}
```

ryanpetrello: Or maybe something like this, which **won't** populate keys for credentials that don't have a user/pass set:

```python
>>> cred.inputs
{'host': 'https://quay.io/foo/bar', 'username': ''}
>>> dict(
    (k, cred.get_input(k))
    for k in cred.inputs if k != 'host' and cred.get_input(k)
)
{}
```

awx/main/tasks.py
```diff
@@ -904,6 +904,23 @@ def build_execution_environment_params(self, instance):
...
 
+        if instance.execution_environment.credential:
+            with open('/tmp/auth.json', 'w') as authfile:
+                host = instance.execution_environment.credential.get_input('host')
+                username = instance.execution_environment.credential.get_input('username')
+                password = instance.execution_environment.credential.get_input('password')
+                token = "{}:{}".format(username,password)
+                auth_data = {'auths': 
+                    {host: 
+                        {'auth': b64encode(token.encode('ascii')).decode()}
```
ryanpetrello: Nice!

awx/main/tasks.py
```diff
@@ -904,6 +904,23 @@ def build_execution_environment_params(self, instance):
             "container_options": ['--user=root'],
         }
 
+        if instance.execution_environment.credential:
+            with open('/tmp/auth.json', 'w') as authfile:
```
ryanpetrello: @rebeccahhh @shanemcd should we randomize this file path w/ `tempfile.NamedTemporaryFile`?  If we don't, it seems like parallel jobs would step on eachother all editing this file.

shanemcd: Yup I think that's the plan and this is just a WIP

ryanpetrello: Actually, is this something that belongs in the private data dir itself?

shanemcd:
A few outstanding items:

- [ ] Remove separate `token` field from the registry credential. 
- [ ] Rename the existing password field to say "password/token"
- [ ] Ensure only registry credentials can be associated with an EE https://github.com/ansible/awx/issues/9628

awx/main/tasks.py
ryanpetrello: Okey-doke, sorry if I'm being noisy 😄

shanemcd: Probably not, because we don't want the user in the playbook to see the registry auth info.

Maybe it could go in the new wrapper directory we added?

awx/main/tasks.py
rebeccahhh: better now?

awx/main/models/credential/__init__.py
```diff
@@ -1092,14 +1092,8 @@ def create(self):
...
-                'id': 'password',
-                'label': ugettext_noop('Password'),
-                'type': 'string',
-                'secret': True,
-            },
-            {
-                'id': 'token',
-                'label': ugettext_noop('Access Token'),
+                'id': 'password/token',
+                'label': ugettext_noop('Password/Token'),
```
ryanpetrello: This might break AWX 18.0.0 -> 18.next upgrades if anybody made a credential in the old format (`inputs.password` or `inputs.token` set).  That might be acceptable, though, considering this functionality doesn't actually have any effect yet.

Also, I don't think that `id` is allowed to have `/` in it; I think it has to be a valid Python identifier.

ryanpetrello: For backwards compatibility, maybe we should just pick one of the prior `id` values e.g., `token`.  This would mean if you previously set `password`, you just have to do it again after upgrading past AWX 18.0.0

shanemcd: Hmmm. Does this not require a migration?

shanemcd:
Added one more item to https://github.com/ansible/awx/pull/9683#issue-599794103

Basically, the private data directory looks like this: `/tmp/pdd_wrapper_81_jhx26rx_/awx_81_8qpl47iz`.

We don't want the `auth.json` to be visible to the playbook / job, so we should write this file inside of `/tmp/pdd_wrapper_81_jhx26rx_`

awx/main/models/credential/__init__.py
rebeccahhh: I arbitrarily chose token, but will change it if need be. Since I combined the two fields into one it may require a migration even if I do just choose an existing id string

ryanpetrello: @shanemcd I'm pretty sure that the definitions for credential type inputs and injectors no longer reside in the database, so this won't actually require a data migration:

https://github.com/ansible/awx/pull/3257

That said, if somebody sets up a **Credential** w/ the *old* schema and sets a `password` value in 18.0.0 in its `.inputs`, it'll just never be accessed again after this code change in 18.next.  Post-upgrade, they'll have to come fill this field back in.  I think that's okay, though.

awx/api/serializers.py
```diff
@@ -1395,6 +1395,11 @@ def get_related(self, obj):
             res['credential'] = self.reverse('api:credential_detail', kwargs={'pk': obj.credential.pk})
         return res
 
+    def validate_credential(self, value):
+        if value and value.kind != 'registry':
+            raise serializers.ValidationError(_('Only Container Registry credentials can be associated with an Execution Environment'))
```
ryanpetrello, line 1417: 👍 nice!

awx/main/models/credential/__init__.py
shanemcd: Agreed. Hopefully the number of people who have done this already is fairly minimal.

awx/api/serializers.py
shanemcd: 😎

awx/main/models/credential/__init__.py
rebeccahhh: double agreed ^^

awx/main/tasks.py
```diff
@@ -904,6 +904,20 @@ def build_execution_environment_params(self, instance):
             "container_options": ['--user=root'],
         }
 
+        if instance.execution_environment.credential:
+            path = self.build_private_data_dir(instance)
```
shanemcd: I don't _think_ we want to re-build the private data directory here, but rather we want to reference the existing one that should have gotten created earlier in the call stack.

Also this seems like it would put this file directly in the private data directory, but what we actually want is to put it in the `pdd_wrapper` directory.

rebeccahhh: ope, I'll add that post-lunch

awx/main/tasks.py
```diff
@@ -906,6 +906,23 @@ def build_execution_environment_params(self, instance):
...
+                    username = cred.get_input('username')
+                    password = cred.get_input('password')
+                    token = "{}:{}".format(username, password)
+                    auth_data = {'auths': {host: {'auth': b64encode(token.encode('ascii')).decode()}}}
+                    authfile.write(json.dumps(auth_data, indent=4))
+                authfile.close()
+                os.chmod(authfile.name, stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR)
+                params["container_options"].append(f'--authfile={authfile.name}')
+            else:
+                logger.exception('Please recheck that your host, username, and password fields are all filled.')
```
kdelee: is this printed to the job stdout/stderr?

rebeccahhh: oh shoot no it's not, but it definitely should

rebeccahhh: I've updated this so it's in stdout and it's now visible in the play output in the UI, but if there's something that would look better I'll change it.

AlanCoding (approved):

shanemcd (approved):

beeankha (approved):

awx/main/tasks.py
```diff
@@ -906,6 +906,23 @@ def build_execution_environment_params(self, instance):
             "container_options": ['--user=root'],
         }
 
+        if instance.execution_environment.credential:
+            cred = instance.execution_environment.credential
+            if cred.has_inputs(field_names=('host', 'username', 'password')):
```
ryanpetrello, line 856: Isn't `host` required?

awx/main/tasks.py
```diff
@@ -906,6 +906,23 @@ def build_execution_environment_params(self, instance):
...
+                path = os.path.split(private_data_dir)[0]
+                with open(path + '/auth.json', 'w') as authfile:
+                    host = cred.get_input('host')
+                    username = cred.get_input('username')
+                    password = cred.get_input('password')
+                    token = "{}:{}".format(username, password)
+                    auth_data = {'auths': {host: {'auth': b64encode(token.encode('ascii')).decode()}}}
```
ryanpetrello, line 863: Does this work if the password/token contains unicode characters, e.g., `Iñtërnâtiônàlizætiøn` ?

ryanpetrello (approved):
Some minor comments, but I think this looks great and is ready to merge 😄

awx/main/tasks.py
rebeccahhh: yes, I was just being abundantly cautious

awx/main/tasks.py
rebeccahhh: the pass is defined by quay, so I haven't explicitly tested that, I definitely can though.

masbahnana (approved):

PR #9684, opened 2021-03-24
add a helpful error message for people who want to run `make black`

ryanpetrello:


shanemcd:
Just a thought. Maybe we could just create a simple script under `tools` called `tools/run_black.sh` or something, and call that from here. Would help keep the Makefile a little cleaner.

shanemcd (approved):

PR #9691, opened 2021-03-25
build requirements with python3.8

ryanpetrello:


requirements/requirements.txt
```diff
@@ -148,22 +148,12 @@ hyperlink==20.0.1
     # via
     #   autobahn
     #   twisted
-idna-ssl==1.1.0
```
ryanpetrello, line 151: This is unnecessary in Python3.7+

https://github.com/aio-libs/aiohttp/issues/949
https://github.com/python/cpython/pull/3010

requirements/requirements.txt
```diff
@@ -148,22 +148,12 @@ hyperlink==20.0.1
...
-    # via aiohttp
 idna==2.9
     # via
     #   hyperlink
-    #   idna-ssl
     #   requests
     #   twisted
     #   yarl
-importlib-metadata==1.5.0
```
ryanpetrello, line 160: These are unnecessary in Python3.8

requirements/requirements.txt
```diff
@@ -391,8 +381,6 @@ twisted[tls]==20.3.0
     #   daphne
 txaio==20.12.1
     # via autobahn
-typing-extensions==3.7.4.1
```
ryanpetrello, line 394: `typing` is no longer provisional in 3.8, this polyfill library is unnecessary in 3.8

requirements/requirements.txt
```diff
@@ -409,10 +397,6 @@ xmlsec==1.3.3
     # via python3-saml
 yarl==1.4.2
     # via aiohttp
-zipp==3.1.0
```
ryanpetrello, line 412: This isn't needed because its parent dependencies are gone

requirements/requirements.txt
ryanpetrello:
```shell
bash-4.4$ python3.8 -c 'from importlib import metadata'
bash-4.4$ python3.6 -c 'from importlib import metadata'
  File "<string>", line 1, in <module>
ImportError: cannot import name 'metadata'

bash-4.4$ python3.8 -c 'from importlib import resources'
bash-4.4$ python3.6 -c 'from importlib import resources'
  File "<string>", line 1, in <module>
ImportError: cannot import name 'resources'
```

requirements/requirements.txt
ryanpetrello:
```
bash-4.4$ python3.6 -c 'from typing import Protocol'
  File "<string>", line 1, in <module>
ImportError: cannot import name 'Protocol'
bash-4.4$ python3.8 -c 'from typing import Protocol'
```

Spredzy (approved):

Issue #9472, opened 2021-03-03
awx.main.utils.common.get_ansible_version breaks with ansible-core


elyezer:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

The `awx.main.utils.common.get_ansible_version` helper breaks with new `ansible --version` when ansible-core is installed. This also changes the format the `api/v2/config` returns for the `ansible_version` field. What breaks tools that relies on this field to check the current ansible version used by AWX.

##### ENVIRONMENT
* AWX version: devel
* AWX install method: vm
* Ansible version:  ansible-core 2.11
* Operating System: RHEL8.3
* Web Browser:

##### STEPS TO REPRODUCE

Install AWX with ansible-core and perform a GET request to `/api/v2/config` and check the `ansible_version` field format.

##### EXPECTED RESULTS

The ansible version format should be something like `X.Y.Z` (format capable to be parsed by tools like `distutils.version.LooseVersion`)

##### ACTUAL RESULTS

Returning something like `"ansible_version": "[core 2.11.0.dev0]"`

##### ADDITIONAL INFORMATION

Output of `ansible --version` for ansible-core 2.11:

```console
$ ansible --version
ansible [core 2.11.0b1]
  config file = /home/elyezer/code/ansible/ansible-core/ansible.cfg
  configured module search path = ['/home/elyezer/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/elyezer/.virtualenvs/ansible-core/lib/python3.9/site-packages/ansible
  ansible collection location = /home/elyezer/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/elyezer/.virtualenvs/ansible-core/bin/ansible
  python version = 3.9.0 (default, Oct  6 2020, 00:00:00) [GCC 10.2.1 20200826 (Red Hat 10.2.1-3)]
  jinja version = 2.11.3
  libyaml = True
```

For ansible 2.9:

```
ansible 2.9.16
  config file = /home/elyezer/code/ansible/ansible/ansible.cfg
  configured module search path = ['/home/elyezer/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/elyezer/.virtualenvs/ansible/lib/python3.9/site-packages/ansible
  executable location = /home/elyezer/.virtualenvs/ansible/bin/ansible
[...]

PR #9697, opened 2021-03-25
remove ansible_version from the API config and metrics endpoints

ryanpetrello:
AWX no longer includes Ansible on the control plane and there is no
"global" version of Ansible aside from what's configured at the
Execution Environment level

see: https://github.com/ansible/awx/issues/9472

### DIFF
--- a/awx/main/analytics/collectors.py
+++ b/awx/main/analytics/collectors.py
@@ -11,7 +11,7 @@ from django.utils.timezone import now
 from django.utils.translation import ugettext_lazy as _
 
 from awx.conf.license import get_license
-from awx.main.utils import get_awx_version, get_ansible_version, get_custom_venv_choices, camelcase_to_underscore
+from awx.main.utils import get_awx_version, get_custom_venv_choices, camelcase_to_underscore
 from awx.main import models
 from django.contrib.sessions.models import Session
 from awx.main.analytics import register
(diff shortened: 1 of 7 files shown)


awx/main/analytics/collectors.py
```diff
@@ -52,7 +52,6 @@ def config(since, **kwargs):
         'instance_uuid': settings.SYSTEM_UUID,
         'tower_url_base': settings.TOWER_URL_BASE,
         'tower_version': get_awx_version(),
-        'ansible_version': get_ansible_version(),
```
ryanpetrello, line 55: cc @Ladas

ryanpetrello: @beeankha even though this isn't reported here anymore, it's now part of the analytics payload we ship for jobs (per-job) 🎉 https://github.com/ansible/awx/pull/9598

Ladas (approved):
:+1: makes sense, I'll add issue on AA side, so we deal with this change

Spredzy (approved):
Looks good, (Minus test failing)

beeankha (approved):
LGTM! :+1:

jakemcdermott (approved):
The code removal lgtm & I ran this locally to do some light testing in the places we used to reference or indirectly reference this variable. I didn't find anything notable.

Let's kick off an e2e just to be sure then :+1: from me.

cc @unlikelyzero @tiagodread @akus062381

jakemcdermott (approved):
[branch results](https://dashboard.cypress.io/projects/ui2mq7/runs/5065/overview) match [devel results](https://dashboard.cypress.io/projects/ui2mq7/runs/5062/overview)

PR #9703, opened 2021-03-25
Avoid double LDAP updates

tchellomello:
##### SUMMARY

cc: @ryanpetrello  @gamuniz  @jamesmarshall24

When performing an API call using LDAP auth, the PostgreSQL table `auth_user` is updated twice on each request.

In some environments due to the size of the database, customers opt-in to disable the django-ldap attribute `AUTH_LDAP_ALWAYS_UPDATE_USER` (see  https://django-auth-ldap.readthedocs.io/en/latest/reference.html?highlight=AUTH_LDAP_ALWAYS_UPDATE_USER#auth-ldap-always-update-user) to avoid LDAP attributes being pushed into the database on every authentication. 

Having said this, this patch solves the following scenarios:

1. `auth_table` gets updated 2 times when `AUTH_LDAP_ALWAYS_UPDATE_USER=True` (default)
2. `auth_table` still gets update 1 time when `AUTH_LDAP_ALWAYS_UPDATE_USER=False` and `first_name` or `last_name` is complaint with the max_chars policy

Here is an example of how the query impacts a huge environment

*SQL statement example*
```sql
UPDATE "auth_user" SET "password" = 'pbkdf2_sha256xxxxxxxxxxx', "last_login" = NULL, "is_superuser" = true, 
  "username" = 'admin', "first_name" = 'ryan', "last_name" = '', "email" = 'admin@example.com', 
  "is_staff" = true, "is_active" = true, "date_joined" = '2020-10-20T16:42:36.220849+00:00'::timestamptz 
 WHERE "auth_user"."id" = 1
```

```
$ grep 'statement: UPDATE "auth_user" SET "password"' postgresql-Thu.log  | grep '$Jojk0nPYz2d<SECRET>' | grep "2021-02-18 03:" | wc -l 
3773

--- just yesterday
$ grep 'statement: UPDATE "auth_user" SET "password"' postgresql-Thu.log  | grep  '$Jojk0nPYz2d<SECRET>' | grep "2021-02-18" | wc -l 
6256


$ grep 'statement: UPDATE' postgresql-Thu.log | awk -F" ms " '{ print $1 }' | awk -F'duration:' '{ print $2 }' | sort -n -r | head -n20 
 63108.957
 51544.792
 49495.762
 48576.775
 46365.636
 44387.892
 43147.609
 42522.390
 41821.763
 41497.775
 41309.086
 41056.887
 40278.850
 39570.201
 39256.742
 38831.457
 38775.066
 38706.352
 38152.821
 38012.462
```

**NOTE:** On this patch, I tried to conserve the same approach used currently. During the tests, when the `first_name` has more than 30 chars, it seems the `django-ldap` already truncates the field, so, the second `user.save()` would not be required, however, the patch still will perform the 2nd `update()` in that scenario. 

```
ldapsearch  -ZZ  -x  -H ldap://ldap.tatu.home -D "uid=root,cn=users,dc=ldap,dc=tatu,dc=home" -b "dc=ldap,dc=tatu,dc=home" -W cn=ldap-user -LLLL| grep gecos                                                                                                                                                                  
Enter LDAP Password: 
gecos: long_string_with_more_30_chars____
```

So in this case, you will see the 2 updates still (with AUTH_LDAP_ALWAYS_UPDATE_USER=True)
```
< 2021-02-24 18:29:38.089 UTC localhost-24592-uwsgi >LOG:  duration: 0.533 ms  statement: UPDATE "auth_user" SET "password" = '!8JWTuyWlNrUwltCcYnixrB3YBQFL3a7gNweIauBr', "last_login" = '2021-02-24T17:13:32.598709+00:00'::timestamptz, "is_superuser" = false, "username" = 'ldap-user', "first_name" = 'long_string_with_more_30_chars', "last_name" = 'ldap-user', "email" = 'tchello.mello@gmail.com', "is_staff" = false, "is_active" = true, "date_joined" = '2021-02-23T05:50:48.882136+00:00'::timestamptz WHERE "auth_user"."id" = 8

< 2021-02-24 18:29:38.100 UTC localhost-24592-uwsgi >LOG:  duration: 0.233 ms  statement: UPDATE "auth_user" SET "password" = '!8JWTuyWlNrUwltCcYnixrB3YBQFL3a7gNweIauBr', "last_login" = '2021-02-24T17:13:32.598709+00:00'::timestamptz, "is_superuser" = false, "username" = 'ldap-user', "first_name" = 'long_string_with_more_30_chars', "last_name" = 'ldap-user', "email" = 'tchello.mello@gmail.com', "is_staff" = false, "is_active" = true, "date_joined" = '2021-02-23T05:50:48.882136+00:00'::timestamptz WHERE "auth_user"."id"
```

Related issue: https://github.com/ansible/tower/issues/4558

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
17.1.0
```

##### ADDITIONAL INFORMATION

Maybe QE can evaluate if we should trust the `django-ldap` behavior in truncating the field directly, and then we can completely remove the second `user.update()` completely.

### DIFF
--- a/awx/sso/backends.py
+++ b/awx/sso/backends.py
@@ -352,6 +352,10 @@ def on_populate_user(sender, **kwargs):
     ldap_user = kwargs['ldap_user']
     backend = ldap_user.backend
 
+    # Boolean to determine if we should force an user update
+    # to avoid duplicate SQL update statements
+    force_user_update = False
+
     # Prefetch user's groups to prevent LDAP queries for each org/team when
     # checking membership.
     ldap_user._get_groups().get_group_dns()
@@ -362,6 +366,7 @@ def on_populate_user(sender, **kwargs):
         field_len = len(getattr(user, field))
         if field_len > max_len:
             setattr(user, field, getattr(user, field)[:max_len])
+            force_user_update = True
             logger.warn('LDAP user {} has {} > max {} characters'.format(user.username, field, max_len))
 
     # Update organization membership based on group memberships.
@@ -391,8 +396,9 @@ def on_populate_user(sender, **kwargs):
         _update_m2m_from_groups(user, ldap_user, team.member_role.members, users_opts, remove)
 
     # Update user profile to store LDAP DN.
-    user.save()
-    profile = user.profile
-    if profile.ldap_dn != ldap_user.dn:
-        profile.ldap_dn = ldap_user.dn
-        profile.save()
+    if force_user_update:
+        user.save()
+        profile = user.profile
+        if profile.ldap_dn != ldap_user.dn:
+            profile.ldap_dn = ldap_user.dn
+            profile.save()


ryanpetrello (approved):
Thanks for digging into this @tchellomello 😄 

If @kdelee et al are happy w/ this from an LDAP integration perfective, then I say 👍

felipe4334:
Good job breaking ldap Marcelo 😂

PR #9705, opened 2021-03-25
remove awx.ui

ryanpetrello:
@jakemcdermott this has been bugging me


awx/ui/apps.py
```diff
@@ -1,9 +0,0 @@
-# Django
```
jakemcdermott, line 1: This couples with the configuration system in a few ways. You may end up needing to bring one or both of these files over as well.

jakemcdermott (approved):
I plan to move all of the `ui_next` src back under `ui` once we get some other work items out of the way.

That said, I'm all for moving some of those modules over to the other folder in the meanwhile if it is bugging you :) 

lgtm

ryanpetrello:
> I plan to move all of the ui_next src back under ui once we get some other work items out of the way.

Oh.  Then I'm gonna let you do this correctly when you're ready then 😄

Issue #6663, opened 2020-04-08
Organization field on WFJT form should be required for all non-superusers


dsesami:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Org admin user can't make workflow job templates.

##### STEPS TO REPRODUCE
Copied from the steps to reproduce a different bug:

1. Create OrgA & OrgB
2. Add org admin users CuA & CuB respectively in the Orgs.
3. Create a jt in OrgA
4. Give "execute" permission on the jt for CuB.
5. Login as CuB.
~~6. Try creating a wfjt using the jt.~~
6. Try creating a wfjt without specifying an organization

##### EXPECTED RESULTS
Can get to the workflow add form.

##### ACTUAL RESULTS
Permission denied.

mabashian:
Ah, I see what's going on here.  If you follow all of these steps and try to create a workflow _without_ specifying an org then the API will respond with a 403.  This is because the org field _should_ be required for this user.  I'm going to repurpose this issue to describe that bug.

mabashian:
Here's the relevant lines in the old app: https://github.com/ansible/awx/blob/devel/awx/ui/client/src/templates/workflows.form.js#L64-L66

mabashian:
The steps to reproduce this can probably be simplified tbh.  All you need is an org admin user.  If you go to the WFJT form and don't specify an org you should see this error.

tiagodread:
Verified that:

sys admin still can create wfjt without org
wfjt org field is required for org admins

PR #9710, opened 2021-03-25
Make Org a required field for org admins on WFJT

nixocio:
Make Org a required field for org admins on WFJT.

See: https://github.com/ansible/awx/issues/6663


nixocio:
running e2e tests.

AlexSCorey (approved):
This looks good to me.  Well done!

tiagodread (approved):
Verified that:

- sys admin still can create wfjt without org
- wfjt org field is required for org admins

Issue #9701, opened 2021-03-25
Inventory file dropdown adds slash to filename


lukasertl:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

Creating an inventory - Sourced from a project.  The dropdown for the inventory file actually shows the correct file (and others), but if I choose it there's a slash added to the end of the filename.  The same thing happens if I freely type a filename into the textbox.

##### ENVIRONMENT
* AWX version: 18.0.0
* AWX install method: openshift (using AWX operator)

kladiv:
+1 same issue

larueli:
I had the same issue on a fresh AWX 18.

The fact is the "runner" is choosig the wrong folder.

When i ran into the issue desribed above, I used the api to edit manually the value and remove the trailing slash.
You can do it easily by opening the dev perspective in firefox and the network tab before hitting the save button while editing a inventory source, click the save button, check the put operation in the network tab, right-click on it and edit, then edit the request parameters (source_path) and set it to the correct value and resend it.

But... it didn't work either (maybe it's a different issue ? - edit, it is : #9704). Here is my log :

<details>
<summary>Logs</summary>

```
ansible-inventory [core 2.11.0b3.post0] 
  config file = /tmp/pdd_wrapper_37_xa9qdvcg/awx_37_o4gmbvbm/project/ansible.cfg
  configured module search path = ['/tmp/pdd_wrapper_37_xa9qdvcg/awx_37_o4gmbvbm/project/plugins/modules']
  ansible python module location = /usr/local/lib/python3.8/site-packages/ansible
  ansible collection location = /runner/requirements_collections:/tmp/pdd_wrapper_37_xa9qdvcg/awx_37_o4gmbvbm/project/collections:/home/runner/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible-inventory
  python version = 3.8.3 (default, Aug 31 2020, 16:03:14) [GCC 8.3.1 20191121 (Red Hat 8.3.1-5)]
  jinja version = 2.10.3
  libyaml = True
Using /tmp/pdd_wrapper_37_xa9qdvcg/awx_37_o4gmbvbm/project/ansible.cfg as config file
[WARNING]: Ignoring invalid path provided to plugin path:
'/runner/project/inventory' is not a directory
host_list declined parsing /runner/project/inventory/hosts as it did not pass its verify_file() method
Skipping due to inventory source not existing or not being readable by the current user
[...]

PR #9713, opened 2021-03-25
Don't append a slash to file paths

jakemcdermott:
for #9701

### DIFF
--- a/awx/ui_next/src/screens/Inventory/shared/InventorySourceSubForms/SCMSubForm.jsx
+++ b/awx/ui_next/src/screens/Inventory/shared/InventorySourceSubForms/SCMSubForm.jsx
@@ -126,6 +126,7 @@ const SCMSubForm = ({ autoPopulateProject, i18n }) => {
           }}
           aria-label={i18n._(t`Select source path`)}
           placeholder={i18n._(t`Select source path`)}
+          createText={i18n._(t`Set source path to`)}
           isCreatable
           onCreateOption={value => {
             value.trim();
--- a/awx/ui_next/src/screens/Inventory/shared/InventorySourceSubForms/SCMSubForm.jsx
+++ b/awx/ui_next/src/screens/Inventory/shared/InventorySourceSubForms/SCMSubForm.jsx
@@ -122,9 +122,6 @@ const SCMSubForm = ({ autoPopulateProject, i18n }) => {
           onSelect={(event, value) => {
             setIsOpen(false);
             value = value.trim();
-            if (!value.endsWith('/')) {
-              value += '/';
-            }
             sourcePathHelpers.setValue(value);
           }}
           aria-label={i18n._(t`Select source path`)}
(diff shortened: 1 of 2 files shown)


awx/ui_next/src/screens/Inventory/shared/InventorySourceSubForms/SCMSubForm.jsx
```diff
@@ -122,20 +122,13 @@ const SCMSubForm = ({ autoPopulateProject, i18n }) => {
...
           }}
           aria-label={i18n._(t`Select source path`)}
           placeholder={i18n._(t`Select source path`)}
           isCreatable
           onCreateOption={value => {
             value.trim();
-
-            if (!value.endsWith('/')) {
-              value += '/';
-            }
```
jakemcdermott, line 138: This _may_ be addressing a different use case or bug that I am not aware of...

jakemcdermott: @AlexSCorey Can these custom inventories ever be whole directories? :thinking: (Trying to make sure I don't miss something by removing this)

shanemcd (approved):

awx/ui_next/src/screens/Inventory/shared/InventorySourceSubForms/SCMSubForm.jsx
jakemcdermott: I feel like removing this code is _probably_ fine? Although I can't think of a use case off the top of my head, if someone does need to append a `/` they can still do it manually:

inv_slash

cc @AlexSCorey - thoughts?

AlexSCorey (approved):
LGTM

awx/ui_next/src/screens/Inventory/shared/InventorySourceSubForms/SCMSubForm.jsx
```diff
@@ -126,6 +126,7 @@ const SCMSubForm = ({ autoPopulateProject, i18n }) => {
           }}
           aria-label={i18n._(t`Select source path`)}
           placeholder={i18n._(t`Select source path`)}
+          createText={i18n._(t`Set source path to`)}
```
jakemcdermott, line 129: The default text shown when manually setting an inventory source is kind of weird:

Screenshot from 2021-03-25 18-32-23

"Create foo/bar" makes it seem like a file or directory would be created somewhere when it wouldn't be. 

---

When we type in a path like this, we're really just specifying a file location that may or may not exist. Our component library has a [parameter for customizing the text](https://www.patternfly.org/2020.04/documentation/react/components/select) so I changed it:

Screenshot from 2021-03-25 18-30-54

AlexSCorey: I like this phrasing

jakemcdermott:
note: [branch results](https://dashboard.cypress.io/projects/ui2mq7/runs/5081/overview) currently match [devel results](https://dashboard.cypress.io/projects/ui2mq7/runs/5077/overview)

PR #9726, opened 2021-03-26
Hide commands that are being run for `make black`

shanemcd:
Before:

```
$ make black
command -v black >/dev/null 2>&1 || { echo "could not find black on your PATH, you may need to \`pip install black\`, or set AWX_IGNORE_BLACK=1" && exit 1; }
could not find black on your PATH, you may need to `pip install black`, or set AWX_IGNORE_BLACK=1
make: *** [Makefile:275: black] Error 1
```

After:
```
$ make black
could not find black on your PATH, you may need to `pip install black`, or set AWX_IGNORE_BLACK=1
make: *** [Makefile:275: black] Error 1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -272,12 +272,12 @@ reports:
 	mkdir -p $@
 
 black: reports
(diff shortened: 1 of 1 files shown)


ryanpetrello (approved):
Yes, thank you.

Issue #8349, opened 2020-10-07
Explore ways to make modules project_update and inventory_source_update idempotent


nesanton:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Modules tower_project_update and tower_inventory_source_update both seem to report `changed: true` unconditionally. 
I checked that `modified` does not change on sync and `last_updated` changes every time you hit the sync button even if no change came in. So this might as well be not possible.
But at least for scm projects we could use `scp_revision` before and after to make just this case idempotent.
Or am I late to this discussion?

sean-m-sullivan:
@nesanton Put in a PR to do this when wait is called on the project_update, because that is the situation where the user would be "waiting" on the result to get the information back, otherwise we assume it is changed. 

In terms of Inventory Update, last_updated is updated every time a sync is run, regardless if it updated hosts or not. Modified only keep track of when values on the inventory source gui settings are changed, not when hosts are changed. 

I did a direct comparison of before and after of a non-updating sync from the api results, and also on the update job, and was unable to find a value that would indicate changed, the closest I came to was host count, and that can be misleading. 

We would need a new variable in the API that tracks if the inventory was "changed" or not as a result of the update to implement that function in the module.

nesanton:
> We would need a new variable in the API that tracks if the inventory was "changed" or not as a result of the update to implement that function in the module.

Or perhaps propose to change the behavior of last_updated.

AlanCoding:
> We would need a new variable in the API that tracks if the inventory was "changed" or not as a result of the update to implement that function in the module.

I have a fairly ancient RFE for this at https://github.com/ansible/awx/issues/364. The logging from the update it perfectly clear about changed/not-changed status through all of the stages of the update, so making the evaluation would be to trivially save that fact whenever a stage of the update made a change to the inventory content.

I'm still trying to tease out the immediate objectives of https://github.com/ansible/awx/pull/13928

PR #9729, opened 2021-03-29
Update changed logic on project update

sean-m-sullivan:
##### SUMMARY
Related to #8349
Found the inventory source does not have a way to track it, so did not do that part.

However scm revision is tracked for project update, as we already retrieve this data when looking for the existing item and waiting, put in logic to compare the two values when wait is true. 

Also double checked the integration tests, and believe this should not change them, as its looking for result is success not changed in all but the first update. 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
- awx_collection

##### AWX VERSION
```
18.0.0
```


nesanton:
LGTM! With scm-based inventories and "update on project update" this would cover most of my use cases.

beeankha (approved):
:+1:

beeankha (approved):

Issue #7669, opened 2020-07-20
Add support for convergence any/all on workflow nodes


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
In old tower we added this field:

<img width="1673" alt="Screen Shot 2020-07-20 at 1 18 32 PM" src="https://user-images.githubusercontent.com/9889020/87966675-99ee0800-ca8b-11ea-8d50-95041afa2e41.png">

We'll need to add the same field to the workflow node wizard in ui_next.

mabashian:
Probably best to do this after #8399 lands to avoid some nasty conflicts.

jlmitch5:
This has been verified cc @unlikelyzero

PR #9737, opened 2021-03-29
workflow convergence

jlmitch5:
Satisfies https://github.com/ansible/awx/issues/7669

This adds the ability to set convergence from any nodes (default) to all nodes to the workflow feature.  Specifically:

There is an additional convergence dropdown located on the bottom of the first "node type" step for all node types:

Screen Shot 2021-03-29 at 2 09 54 PM

This field defaults to "Any" on add node and whatever the api has the field set to on edit node.  It resets to any if you change the node type dropdown (even on edit when changing and then changing back to the original type...I can update that point depending on what UX is preferred).

tvo created a new link explicitly to the explanation in documentation of what the convergence setting does [here](https://docs.ansible.com/ansible-tower/latest/html/userguide/workflow_templates.html#convergence-node), and I link to it in the help popover shown in the screenshot below.

Consistent with the old UI, When "All" is selected, a small tab is displayed with the label "ALL" in the node visualizer.  "Any" nodes do not get any sort of tab.

Screen Shot 2021-03-29 at 2 15 08 PM

A slight tweak compared to the old ui...I set the tab's border and background to be the same color as the border of the node to create a consistent look for the node across various states and confirmed this behavior was good with @trahman73 

OLD:

Screen Shot 2021-03-29 at 2 19 18 PM
Screen Shot 2021-03-29 at 2 19 06 PM
Screen Shot 2021-03-29 at 2 18 53 PM

NEW:

Screen Shot 2021-03-29 at 2 20 40 PM
Screen Shot 2021-03-29 at 2 20 19 PM
Screen Shot 2021-03-29 at 2 20 00 PM

### DIFF
--- a/awx/ui_next/src/screens/Template/WorkflowJobTemplateVisualizer/Modals/NodeModals/NodeTypeStep/NodeTypeStep.jsx
+++ b/awx/ui_next/src/screens/Template/WorkflowJobTemplateVisualizer/Modals/NodeModals/NodeTypeStep/NodeTypeStep.jsx
@@ -234,7 +234,8 @@ function NodeTypeStep({ i18n }) {
                 setIsConvergenceOpen(false);
               }}
               aria-label={i18n._(t`Convergence select`)}
-              id="convergence-select"
+              className="convergenceSelect"
+              ouiaId="convergenceSelect"
             >
               <SelectOption key="any" value="any">
                 {i18n._(t`Any`)}
--- a/awx/ui_next/src/screens/Template/WorkflowJobTemplateVisualizer/Modals/NodeModals/NodeTypeStep/NodeTypeStep.jsx
+++ b/awx/ui_next/src/screens/Template/WorkflowJobTemplateVisualizer/Modals/NodeModals/NodeTypeStep/NodeTypeStep.jsx
@@ -237,10 +237,10 @@ function NodeTypeStep({ i18n }) {
               className="convergenceSelect"
               ouiaId="convergenceSelect"
             >
-              <SelectOption key="any" value="any">
+              <SelectOption key="any" value="any" id="select-option-any">
                 {i18n._(t`Any`)}
               </SelectOption>
-              <SelectOption key="all" value="all">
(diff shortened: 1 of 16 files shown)


jakemcdermott (approved):

unlikelyzero:
### Feedback

#### To be Addressed in this PR

- [x]  Node Details should be informative and describe if a Convergence is Any or All

 Screen Shot 2021-03-31 at 2 11 41 PM

- [x] Need access to https://github.com/ansible/awx/pull/9737#discussion_r605264816

- [x] [Legend/Key should be updated](https://github.com/ansible/awx/issues/9791)


#### To be filed as issues
- It would be easier to understand how workflows converge if the links were highlighted with hovering over a Node
- min and sec are not formatted well on the add/edit node wizard
Screen Shot 2021-03-31 at 2 41 11 PM

- [Many websocket connections on WFJT Output page](https://github.com/ansible/awx/issues/9795)
- [Z-depth of links prevents edit link functionality in certain WF Arrangements](https://github.com/ansible/awx/issues/9796)
Screen Shot 2021-03-31 at 2 56 55 PM

awx/ui_next/src/screens/Template/WorkflowJobTemplateVisualizer/Modals/NodeModals/NodeTypeStep/NodeTypeStep.jsx
```diff
@@ -129,61 +145,107 @@ function NodeTypeStep({ i18n }) {
...
+              variant={SelectVariant.single}
+              isOpen={isConvergenceOpen}
+              selections={convergenceField.value}
+              onToggle={setIsConvergenceOpen}
+              onSelect={(event, selection) => {
+                convergenceFieldHelpers.setValue(selection);
+                setIsConvergenceOpen(false);
+              }}
+              aria-label={i18n._(t`Convergence select`)}
+              id="convergence-select"
```
unlikelyzero: I'm not able to access this id

PR #9744, opened 2021-03-29
Fix k8s credentials that use a custom ca cert

shanemcd:


kdelee (approved):

awx/main/models/credential/injectors.py
```diff
@@ -115,6 +115,6 @@ def kubernetes_bearer_token(cred, env, private_data_dir):
         with os.fdopen(handle, 'w') as f:
             os.chmod(path, stat.S_IRUSR | stat.S_IWUSR)
             f.write(cred.get_input('ssl_ca_cert'))
-        env['K8S_AUTH_SSL_CA_CERT'] = path
+        env['K8S_AUTH_SSL_CA_CERT'] = os.path.join('/runner', os.path.basename(path))
```
kdelee, line 118: do we want to check any other credentials that use a CA cert?

PR #9745, opened 2021-03-29
Add uwsgi stats to metrics endpoint

fosterseth:
##### SUMMARY

image
Adds new stats to the api/metrics endpoint such as number of workers busy, number of requests, and request response times

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 18.0.0
```

### DIFF
--- a/awx/main/dispatch/worker/base.py
+++ b/awx/main/dispatch/worker/base.py
@@ -17,6 +17,7 @@ from django.conf import settings
 
 from awx.main.dispatch.pool import WorkerPool
 from awx.main.dispatch import pg_bus_conn
+import awx.main.analytics.subsystem_metrics as s_metrics
 
 if 'run_callback_receiver' in sys.argv:
     logger = logging.getLogger('awx.main.commands.run_callback_receiver')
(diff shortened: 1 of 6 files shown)


awx/main/dispatch/worker/callback.py
```diff
@@ -75,9 +75,8 @@ def read(self, queue):
         return {'event': 'FLUSH'}
 
     def record_read_metrics(self):
-        if self.queue_pop == 0:
```
fosterseth: since we are piggy backing off the read method, we don't want to update uwsgi stats only when jobs are being processed, so we remove this condition.

awx/main/dispatch/worker/base.py
```diff
@@ -127,10 +128,20 @@ class AWXConsumerRedis(AWXConsumerBase):
...
         super(AWXConsumerRedis, self).run(*args, **kwargs)
         self.worker.on_start()
-
+        subsystem_metrics = s_metrics.Metrics(auto_pipe_execute=False)
+        last_debug_time = 0
+        previous_requests = 0
         while True:
-            logger.debug(f'{os.getpid()} is alive')
-            time.sleep(60)
+            requests = subsystem_metrics.update_uwsgi_stats()
```
fosterseth, line 135: we need somewhere to call this `update_uwsgi_stats()` method. The problem with doing it in `callback.py` is that there are mulitiple forks of the callback receiver, and calling this method in each of those could be expensive and wasteful.

So I found this `while True` loop that runs the in the parent thread of all those callback receiver workers. Not ideal, but it does work.

fosterseth:
prom

Averaging response times across workers can be a little deceiving because some workers aren't really active, but have might have high response times, driving the average up.

Therefore I have a metric called `uwsgi_busy_response_time` that only averages across "busy" workers to get a better sense of response times under load.

shanemcd:
@fosterseth Do we still want this?

fosterseth:
These could be helpful stats, but I really don't like hijacking a random while True loop to do it, going to close this for now.

fosterseth:
@kdelee some work to track wsgi worker activity, we could resurrect it if we want

PR #9749, opened 2021-03-30
add the fallback satellite_instance_var_id

jainnikhil30:
##### SUMMARY
With https://github.com/ansible/awx/pull/9037, we fixed the issue to use the foreman_* because we had an migration which was converting the vars as foreman_*. Here:

https://github.com/ansible/tower/blob/devel/awx/main/migrations/_inventory_source_vars.py#L663

But this migration only happens on upgrade. But for new users this migration wont run and hence the variable would not be in foreman_* format. This produces a warning like following:

```
6.240 WARNING  Host "example.com" has no "foreman_id" variable(s)
```
We should have a fallback on the new var of foreman.id with foreman_id in there too. This PR adds that.
 
##### ISSUE TYPE

  - Bugfix Pull Request
 

##### COMPONENT NAME

 - API

##### ADDITIONAL INFORMATION

Thanks @gamuniz for finding the bug and pointing the migration.

### DIFF
--- a/awx/settings/defaults.py
+++ b/awx/settings/defaults.py
@@ -706,10 +706,10 @@ TOWER_INSTANCE_ID_VAR = 'remote_tower_id'
 # ---------------------
 # ----- Foreman -----
 # ---------------------
-SATELLITE6_ENABLED_VAR = 'foreman_enabled'
+SATELLITE6_ENABLED_VAR = 'foreman_enabled, foreman.enabled'
 SATELLITE6_ENABLED_VALUE = 'True'
 SATELLITE6_EXCLUDE_EMPTY_GROUPS = True
-SATELLITE6_INSTANCE_ID_VAR = 'foreman_id'
+SATELLITE6_INSTANCE_ID_VAR = 'foreman_id, foreman.id'
 # SATELLITE6_GROUP_PREFIX and SATELLITE6_GROUP_PATTERNS defined in source vars
 
 # ---------------------


sugitk:
I tried to your PR on my side (Tower 3.8.2).

1) 
When changed the source variables in inventory source side, the warnings were not shown.
```
vars_prefix: 'foreman_'
legacy_hostvars: false
```

2) 
Reverted inventory source side to the default.
```
#vars_prefix: 'foreman_'
legacy_hostvars: true
```

/etc/tower/conf.d/custom.py
```
SATELLITE6_ENABLED_VAR = 'foreman_enabled, foreman.enabled'
SATELLITE6_INSTANCE_ID_VAR = 'foreman_id, foreman.id'
```

It caused an error:
```
usage: awx-manage inventory_import [-h] [--inventory-name n]
                                   [--inventory-id i] [--venv VENV]
                                   [--overwrite] [--overwrite-vars]
                                   [--keep-vars] [--custom] [--source s]
                                   [--enabled-var v] [--enabled-value v]
                                   [--group-filter regex]
                                   [--host-filter regex]
                                   [--exclude-empty-groups]
                                   [--instance-id-var v] [--version]
                                   [-v {0,1,2,3}] [--settings SETTINGS]
                                   [--pythonpath PYTHONPATH] [--traceback]
                                   [--no-color] [--force-color]
awx-manage inventory_import: error: unrecognized arguments: foreman.enabled
```

3) 
Changed /etc/tower/conf.d/custom.py
```
SATELLITE6_ENABLED_VAR = 'foreman_enabled'
SATELLITE6_INSTANCE_ID_VAR = 'foreman_id, foreman.id'
```

It still shows warnings.
```
    6.538 WARNING  Host "ip-10-0-0-11.ca-central-1.compute.internal" has no "foreman_id, foreman.id" variable(s)
    6.538 WARNING  Host "sheri-magan.example.com" has no "foreman_id, foreman.id" variable(s)
    6.538 WARNING  Host "virt-who-abcdef01.example.com" has no "foreman_id, foreman.id" variable(s)
```

4)
Changed again /etc/tower/conf.d/custom.py
```
SATELLITE6_ENABLED_VAR = 'foreman_enabled'
SATELLITE6_INSTANCE_ID_VAR = 'foreman.id'
```

It worked without warnings.

### results

When `legacy_hostvars: true`, the sync will parse the JSON data from the Satellite API and construct it as a structure under `foreman` tree.  On the other hand `legacy_hostvars: false`, the data will be stored as a flat.  Then it may need `vars_prefix: 'foreman_'` and it should be parsed as `foreman_id`.

AlanCoding:
Wouldn't you want to do:

```python
SATELLITE6_INSTANCE_ID_VAR = 'foreman_id,foreman.id'
```

(with no space)

It looks like there was a CLI argument processing issue, which is odd, because there shouldn't be any need to use non-standard characters.

In code, I also saw that we don't do any stripping of whitespace.

john-westcott-iv:
Closing in favor of #13851

PR #9756, opened 2021-03-30
Adding back in missing functionality of Custom Docker Image Builds

james-crowley:
##### SUMMARY
With the switch to 18.0, the removal of local docker install was removed. During this shuffle of code, it looks like some functionality was dropped. This PR aims to add back in that functionality. Plus add unofficial support for ppc64le like there is for arm64.  

##### ISSUE TYPE
 - Feature/Bug Pull Request

##### COMPONENT NAME
 - Installer

##### AWX VERSION

```
awx: 18.0.0
```

##### ADDITIONAL INFORMATION
The functionality that was delete/shifted was:

- Allowing for custom build images to be pushed to remote registry
- Allowing for multi-arch builds via rendering the right download paths

New Stuff Added:

- Unofficial support for build ppc64le
- Support for custom base images(aka if a certain arch does not support `centos:8`, you could swap it for `fedora` or something similar)

### DIFF
--- a/tools/ansible/roles/dockerfile/defaults/main.yml
+++ b/tools/ansible/roles/dockerfile/defaults/main.yml
@@ -4,3 +4,8 @@ kube_dev: false
 dockerfile_dest: '../..'
 dockerfile_name: 'Dockerfile'
 template_dest: '_build'
+
+# Helper vars to construct the proper multi-arch configuration
+tini_architecture: '{{ { "x86_64": "amd64", "aarch64": "arm64", "armv7": "arm", "ppc64le": "ppc64le" }[ansible_facts.architecture] }}'
+kubectl_architecture: '{{ { "x86_64": "amd64", "aarch64": "arm64", "armv7": "arm", "ppc64le": "ppc64le" }[ansible_facts.architecture] }}'
+base_container: '{{ { "x86_64": "centos:8", "aarch64": "centos:8", "armv7": "centos:8", "ppc64le": "centos:8" }[ansible_facts.architecture] }}'
--- a/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
+++ b/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
@@ -5,7 +5,7 @@
 ###
 
 # Build container
-FROM centos:8 as builder
+FROM {{ base_container | default('centos:8') }}
 
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US:en
@@ -70,7 +70,7 @@ RUN make sdist && \
 {% endif %}
 
 # Final container(s)
-FROM centos:8
+FROM {{ base_container | default('centos:8') }}
 
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US:en
--- a/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
+++ b/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
@@ -5,7 +5,7 @@
 ###
 
 # Build container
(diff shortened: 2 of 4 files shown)


james-crowley:
@vielmetti If you get the chance, do you mind confirming that `arm64` support is back in for custom built images in my PR? I think #9289 broke it in the reshuffling of code.

tools/ansible/roles/dockerfile/templates/Dockerfile.j2
```diff
@@ -5,7 +5,7 @@
 ###
 
 # Build container
-FROM centos:8 as builder
+FROM {{ base_container | default('centos:8') }}
```
jakemcdermott: I think you want to keep the `as builder` directive, as other parts of the container file [expect a build stage by that name](https://github.com/ansible/awx/blob/devel/tools/ansible/roles/dockerfile/templates/Dockerfile.j2#L168).

jakemcdermott (changes requested):

tools/ansible/roles/dockerfile/defaults/main.yml
```diff
@@ -4,3 +4,8 @@ kube_dev: false
 dockerfile_dest: '../..'
 dockerfile_name: 'Dockerfile'
 template_dest: '_build'
+
+# Helper vars to construct the proper multi-arch configuration
+tini_architecture: '{{ { "x86_64": "amd64", "aarch64": "arm64", "armv7": "arm", "ppc64le": "ppc64le" }[ansible_facts.architecture] }}'
+kubectl_architecture: '{{ { "x86_64": "amd64", "aarch64": "arm64", "armv7": "arm", "ppc64le": "ppc64le" }[ansible_facts.architecture] }}'
+base_container: '{{ { "x86_64": "centos:8", "aarch64": "centos:8", "armv7": "centos:8", "ppc64le": "centos:8" }[ansible_facts.architecture] }}'
```
jakemcdermott, line 11: For `base_container`, why use a mapping if all keys point to the same container image?

tools/ansible/roles/dockerfile/templates/Dockerfile.j2
james-crowley: @jakemcdermott Good call. I'll fix that now.

tools/ansible/roles/dockerfile/defaults/main.yml
james-crowley: @jakemcdermott This is to allow for other architectures that do not have `centos:8` available, notably `s390x`. In my testing, I was able to add an entry into the map here that looked something like:

```
"s390x": "fedora:32"
```

Which thus allowed me to build custom images for `s390x`. The reason I did not add `s390x` into the existing map is it requires other jinja edits in the Dockerfile template. I was going to do some more testing and do a PR later to add in unofficial support for `s390x`. But I thought adding in the `base_container` variable is useful right now for others.

james-crowley: You can see some of the smaller edits I backed ported into `17.1.0` to add `s390x` support [here](https://github.com/ansible/awx/compare/17.1.0...LinuxCoC:17.1.0_multi_arch#diff-4d685a626988eae11d1c9ada073e0926b3518704d8ae3a5e1869631f9edbf3a4).

james-crowley: @jakemcdermott Are you fine with this addition?

shanemcd:
Hello. If you could please rebase and resolve the conflict, we can merge this.

shanemcd:
Closing due to lack of response.

Issue #9187, opened 2021-01-28
The user can try to delete a running job


tiagodread:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Follow up bug from https://github.com/ansible/awx/pull/8970#issuecomment-767863581 
The user can select a running job and the delete button appears enabled

##### STEPS TO REPRODUCE

1. Select a running job on Job list
2. Click on delete button

##### EXPECTED RESULTS

The delete button shouldn’t be enabled

##### ACTUAL RESULTS

![image](https://user-images.githubusercontent.com/12464858/106160916-1e20f680-6165-11eb-8355-2c44ba1dedfe.png)

tiagodread:
Verified that the user can only delete finished jobs 👍🏽

PR #9758, opened 2021-03-30
Do not allow user to attempt to delete a running job

nixocio:
Do not allow user to attempt to delete a running job

See: https://github.com/ansible/awx/issues/9187

image

### DIFF
--- a/awx/ui_next/src/components/JobList/JobList.jsx
+++ b/awx/ui_next/src/components/JobList/JobList.jsx
@@ -152,6 +152,8 @@ function JobList({ i18n, defaultParams, showTypeColumn = false }) {
     }
   };
 
+  const cannotDeleteItems = selected.filter(job => isJobRunning(job.status));
+
   return (
     <>
       <Card>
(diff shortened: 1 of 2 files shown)


awx/ui_next/src/components/JobList/JobList.jsx
```diff
@@ -238,6 +238,13 @@ function JobList({ i18n, defaultParams, showTypeColumn = false }) {
                   onDelete={handleJobDelete}
                   itemsToDelete={selected}
                   pluralizedItemName={i18n._(t`Jobs`)}
+                  cannotDelete={item =>
+                    item.status === 'running' ||
```
jakemcdermott: There's probably more than one status we'll need to check for. Let's use the `isJobRunning` util: https://github.com/ansible/awx/blob/devel/awx/ui_next/src/util/jobs.js#L10

nixocio: I will also try to use the plural for the error message.

nixocio: @jakemcdermott, updated.

jakemcdermott (approved):

awx/ui_next/src/components/JobList/JobList.jsx
```diff
@@ -238,6 +240,19 @@ function JobList({ i18n, defaultParams, showTypeColumn = false }) {
                   onDelete={handleJobDelete}
                   itemsToDelete={selected}
                   pluralizedItemName={i18n._(t`Jobs`)}
+                  cannotDelete={item =>
```
AlexSCorey, line 243: Since pluralization can be complicated its best to use this syntax. https://github.com/ansible/awx/blob/705881123c3e8e923ecd45236b989bbdc23449c1/awx/ui_next/src/screens/Inventory/InventoryList/InventoryList.jsx#L222

Once we upgrade Lingui we will adopt `<Plural/>`

nixocio: @AlexSCorey, I saw this syntax. And I attempted to use it. But the string reported during unit-tests was not sane. I will take a better look if I missed something.

tiagodread (approved):
Worked as expected 👍🏽

PR #9759, opened 2021-03-30
Make tower_license module idempotent

ghjm:
##### SUMMARY

Currently, the tower_license module always installs a license, without checking if Tower is already licensed.  In general, Ansible modules are supposed to be idempotent.  This PR updates the module to check whether Tower is already licensed, and do nothing in that case.  A new parameter `force` is provided so the user can always install the license, or override an existing license with a new one.

##### ISSUE TYPE

Bughancement

##### COMPONENT NAME

awx_collection

##### AWX VERSION

awx: 18.0.0
(but really Tower 3.8.2)

### DIFF
--- a/awx_collection/plugins/modules/tower_license.py
+++ b/awx_collection/plugins/modules/tower_license.py
@@ -59,6 +59,7 @@ def main():
         argument_spec=dict(
             manifest=dict(type='str', required=True),
             eula_accepted=dict(type='bool', required=True),
+            force=dict(type='bool', required=False),
         ),
     )
 
--- a/awx_collection/plugins/modules/tower_license.py
+++ b/awx_collection/plugins/modules/tower_license.py
@@ -33,7 +33,7 @@ options:
     force:
       description:
         - By default, the license manifest will only be applied if Tower is currently
-          unlicensed or trial licensed.  When force=true, the license is always applied. 
+          unlicensed or trial licensed.  When force=true, the license is always applied.
       required: True
       type: bool
       default: 'False'
@@ -62,7 +62,7 @@ def main():
         ),
     )
 
-    json_output = {}
+    json_output = {'changed': False}
 
     if not module.params.get('eula_accepted'):
         module.fail_json(msg='You must accept the EULA by passing in the param eula_accepted as True')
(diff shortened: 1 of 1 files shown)


shanemcd (approved):
Well look who it is 😄

ghjm:
Can I get a backport to the next 3.8.x plzthx 😁    
Edit: or a merge to wherever the certified ansible.tower collection lives, is what I really mean

shanemcd:
Please contact your Red Hat support rep 😉

ghjm:
> Please contact your Red Hat support rep 😉

Don't think I won't, now that I can 😂

rooftopcellist (approved):

awx_collection/plugins/modules/tower_license.py
```diff
@@ -55,7 +62,7 @@ def main():
         ),
     )
 
-    json_output = {'changed': True}
+    json_output = {}
```
AlanCoding: Reading through this diff, I see a scenario where the "changed" key is not populated. Instead of deleting its initial value, I suggest that you should change the initial value to `False`.

If `perform_install` is evaluated to `False` (say, it already has a license, non-force), and it's not running check mode, then it won't set changed=False right now

ghjm: @AlanCoding Updated to make sure `changed` is always set.  I see that many other modules in this directory rely on an unset changed meaning not changed, but I'm happy to be explicit about it.

I also cast `perform_install` as a bool, because when `force` is undefined, `module.params.get('force')` will return None, not False, and maybe there's some world in which setting `changed: None` causes some problem downstream of this code.

AlanCoding (approved):

AlanCoding (dismissed):
ansible-test failure

```
Error Message
The test `ansible-test sanity --test validate-modules` failed with 1 error:
Stacktrace
plugins/modules/tower_license.py:0:0: nonexistent-parameter-documented: Argument 'force' is listed in DOCUMENTATION.options, but not accepted by the module argument_spec
```

beeankha (approved):

ghjm:
Good catch @AlanCoding, fixed

awx_collection/plugins/modules/tower_license.py
ghjm: Good catch @AlanCoding, fixed

ghjm:
Good catch @AlanCoding, fixed

beeankha (approved):

jakemcdermott:
regate

jakemcdermott:
regate

PR #9776, opened 2021-03-31
Observability Metrics

AlexSCorey:
##### SUMMARY
This adds the chart for Observability Metrics (#9019).  To see the chart you need to navigate to /metrics.  Also, its best if you run a build that has multiple instances.  This will do that for you `COMPOSE_TAG=devel CLUSTER_NODE_COUNT=2 make docker-compose`.  

When this feature loads the user has to select an instance (1, or all) and a metric to render data on the graph.  Once they select those items, the chart appears and we start to make requests to the api every 3 seconds to get the data. (Currently the api does not support web sockets for this feature) If the user changes the values for either of the drop down items the chart resets.  The chart also only show the last 50 data points.  

There is a "tooltip" that is rendered at the bottom left hand side.  I decided to put it there, instead of on the chart itself because this chart could get quite crowded depending the number of data points rendered and the number of instances rendering lines.  

The X axis is sort of meaningless.  The values below simply render the number of api requests. This isn't a value of time. Since the main goal of this feature is to show significant changes instead of tryin to pinpoint when the change occurs I felt that showing a time stamp on this axis would crowd the axis as well.  

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
metrics


unlikelyzero:
@chrismeyersfsu can you give us some info on what sort of data you think will be the most interesting? We can mock/stub any API data for our tests

awx/ui_next/src/screens/ObservabilityMetrics/ObservabilityMetrics.jsx
```diff
@@ -0,0 +1,216 @@
...
+      setRenderedData(countRestrictedData);
+      return data[metric].help_text;
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [instance, metric, count, instances]),
+    ''
+  );
+
+  // polls api every 3 seconds for data
+
+  // useInterval(fetchData, 3000, instance, metric);
```
nixocio:
```suggestion
  useInterval(fetchData, 3000, instance, metric);
```

awx/ui_next/src/screens/ObservabilityMetrics/ObservabilityMetrics.jsx
```diff
@@ -0,0 +1,216 @@
...
+        });
+      });
+      let countRestrictedData = rendered;
+      if (count.length > 49) {
+        countRestrictedData = rendered.map(({ values, name }) => ({
+          name,
+          values: values.slice(-50),
+        }));
+      }
+      console.log(data[metric].help_text);
```
nixocio:
```suggestion
```

awx/ui_next/src/App.jsx
```diff
@@ -18,6 +18,7 @@ import ja from './locales/ja/messages';
 import en from './locales/en/messages';
 import { isAuthenticated } from './util/auth';
 import { getLanguageWithoutRegionCode } from './util/language';
+import ObservabilityMetrics from './screens/ObservabilityMetrics';
```
jakemcdermott:
```suggestion
```

awx/ui_next/src/App.jsx
```diff
@@ -65,6 +66,9 @@ function App() {
                         </ProtectedRoute>
                       ))
                       .concat(
+                        <Route exact path="/metrics">
+                          <ObservabilityMetrics />
+                        </Route>,
```
jakemcdermott:
```suggestion
```

Please import the screen component and add the route to [ui_next/src/routeConfig.js](https://github.com/ansible/awx/blob/devel/awx/ui_next/src/routeConfig.js#L136-L137)

awx/ui_next/src/screens/ObservabilityMetrics/LineChart.jsx
```diff
@@ -0,0 +1,272 @@
+import React, { useEffect, useCallback } from 'react';
+import * as d3 from 'd3';
+import { t } from '@lingui/macro';
+import { withI18n } from '@lingui/react';
+import { PageContextConsumer } from '@patternfly/react-core';
+
+function LineChart({ data, i18n, helpText }) {
+  console.log(data, 'data');
```
jakemcdermott:
```suggestion
```

awx/ui_next/src/screens/ObservabilityMetrics/LineChart.jsx
```diff
@@ -0,0 +1,272 @@
...
+    return () => window.removeEventListener('resize', handleResize);
+  }, [draw]);
+
+  return <div id="chart" />;
+}
+
+const withPageContext = Component => {
+  return function contextComponent(props) {
+    return (
+      <PageContextConsumer>
```
jakemcdermott: What is `PageContextConsumer` for? 

(Just to see what would happen, I removed it from the code and as far as I can tell everything still works just fine without it?)

awx/ui_next/src/screens/ObservabilityMetrics/ObservabilityMetrics.jsx
```diff
@@ -0,0 +1,239 @@
...
+                    {instances.map(inst => (
+                      <SelectOption value={inst} key={inst} />
+                    ))}
+                  </Select>
+                </ToolbarItem>
+                <ToolbarItem>{i18n._(t`Metric`)}</ToolbarItem>
+                <ToolbarItem>
+                  <Select
+                    ouiaId="Metric-select"
+                    placeholderText={i18n._(t`Select an metric`)}
```
jakemcdermott:
```suggestion
                    placeholderText={i18n._(t`Select a metric`)}
```

awx/ui_next/src/screens/ObservabilityMetrics/LineChart.jsx
```diff
@@ -0,0 +1,272 @@
...
+          .duration(duration)
+          .attr('r', circleRadius);
+      });
+
+    /* Add Axis into SVG */
+    const xAxis = d3
+      .axisBottom(xScale)
+      .ticks(data[0].values.length > 5 ? data[0].values.length : 5);
+    const yAxis = d3.axisLeft(yScale).ticks(5);
+
```
jakemcdermott: Screenshot from 2021-04-06 09-29-03

 The "values" and "count" labels are confusing to me. "Count" isn't really accurate for the horizontal axis and, if anything, "counts" is more suitable for the vertical axis because it's typically showing counts of the event types received. I think we should either remove the axis labels or change "count" to "time" and "values" to "count".

awx/ui_next/src/screens/ObservabilityMetrics/ObservabilityMetrics.jsx
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    const id = setInterval(tick, delay);
+    return () => {
+      clearInterval(id);
+    };
+  }, [callback, delay, instance, metric]);
+  return { count };
+}
+function ObservabilityMetrics({ i18n }) {
```
jakemcdermott:
```suggestion
function Metrics({ i18n }) {
```

As far as the code is concerned, I think we should just call this screen "Metrics".

jakemcdermott (changes requested):
@AlexSCorey This is great! Nice work on this. I left a few suggestions in the review comments for you to look at.

A few additional notes / general questions:

- I really like how the chart behaves when it reaches "steady state": Once there's 60 data points to show on the horizontal axis, the scaling stays consistent and older data points are removed as new ones are added :thumbsup:

  That said, I do think the _initial_ behavior (before 60 data points are received) is a little strange. We show the first data point in the center and then progressively subdivide the axis until 60 data points are received. Instead of rescaling the axis every time, can we make it so the graph initializes with the final scale and we just add data to it from the right as it's received? I feel like that would make the graph updates nicer in the beginning.

- need to add breadcrumb

awx/ui_next/src/screens/ObservabilityMetrics/LineChart.jsx
AlexSCorey: Yeah I struggled with the names here since, especially the x-axis, isn't really anything of value.  I'm on board with removing them.

awx/ui_next/src/screens/ObservabilityMetrics/ObservabilityMetrics.jsx
```diff
@@ -0,0 +1,239 @@
...
+    if (instance && metric) {
+      fetchData();
+    }
+  }, [fetchData, instance, metric]);
+
+  useEffect(() => {
+    fetchInstances();
+  }, [fetchInstances]);
+  if (fetchInitialError || updateError) {
+    console.log('here');
```
jakemcdermott:
```suggestion
```

awx/ui_next/src/screens/ObservabilityMetrics/ObservabilityMetrics.jsx
```diff
@@ -0,0 +1,239 @@
...
+    return (
```
jakemcdermott: Breadcrumbs are missing:

Screenshot from 2021-04-06 09-52-37

awx/ui_next/src/App.jsx
jakemcdermott: talked about this w/ @AlexSCorey - we actually don't want this in the navbar right now so this is fine as is :+1:

unlikelyzero:
@AlexSCorey can you add unique IDs to the items in the select components?

Is it possible to convert it to a select component? As we see it, it looks like  a dropdown select with individual buttons in a list

one-t:
just a quick access control comment- this is currently viewable by non-admin users. Do we want to restrict it to admins?

one-t:
When left open the memory usage in chrome appears to balloon- this has been open for about an hour.
image

unlikelyzero:
@AlexSCorey are you going to address the potential memory leaks?

AlexSCorey:
> just a quick access control comment- this is currently viewable by non-admin users. Do we want to restrict it to admins?

@wenottingham Whats your thoughts here?

wenottingham:
Are the metrics world-viewable in the API?

chrismeyersfsu:
Nope, must be authed to view.

one-t:
> Are the metrics world-viewable in the API?

They're viewable by any authenticated user, but not publicly exposed

one-t:
After discussing with Alex and experimenting with the memory usage, we determined that the dev tools being open in the tab was the cause of the excessive usage. it will continually balloon when they are open but once closed it drops right back down to the normal 200-300 M range

wenottingham:
> They're viewable by any authenticated user, but not publicly exposed

I would not restrict them if the API isn't.

awx/ui_next/src/screens/ObservabilityMetrics/ObservabilityMetrics.jsx
one-t: Hey. @jakemcdermott , @AlexSCorey - just wanted to see if there's an update on this and the other requested changes.

jakemcdermott (approved):

one-t (approved):

Issue #9788, opened 2021-04-01
Add ability to set instance groups on a job template in the tower_job_template module in the collection


kdelee:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Right now to associate an instance group with a JT there is no way to do it from the `tower_job_template` module.

Only way I've figured out how to do it is with a `uri` module task after creating the job template, and this is not necessarily as nice as I'd like because it is not really able to ensure that this is the only instance group set and is not idempotent, because it does not check what instance groups are set first. It would take several `uri` tasks to implement checking the current instance groups, deleting others, and setting the desired ones and only doing so if necessary.

## Current way w/ URI:

```
    - name: Create job templates
      tower_job_template:
        name: "{{ jt_name }}"
        organization: "{{ organization_name }}"
        project: "{{ proj_name }}"
        playbook: example.yml
        job_type: run
        forks: 16
        allow_simultaneous: true
        inventory: "example_inventory"
        state: present
      register: jt_result

    - name: associate the instance group with the job template
      uri:
        url: https://{{ tower_host }}/api/v2/job_templates/{{ item.id }}/instance_groups/
        force_basic_auth: yes
        user: admin
        password: "{{ admin_password }}"
        method: POST
        body_format: json
        body:
          "{{ {'id': instance_group_id|int }|to_json }}"
        status_code: 204
        validate_certs: no
      loop: "{{ jt_results.result}}"
```

## Ideal way w/ module
```
    - name: Create job templates
      tower_job_template:
        name: "{{ jt_name }}"
        organization: "{{ organization_name }}"
        project: "{{ proj_name }}"
        playbook: example.yml
        job_type: run
        instance_groups: [1, 2, 5] # list of id's of instance groups. Even slicker if it can take names instead
        forks: 16
        allow_simultaneous: true
        inventory: "example_inventory"
        state: present
      register: jt_result
```

sean-m-sullivan:
should be an association that we can add to our many association we have.

one-t:
Tested, and this is working as expected. instance groups are replaced when changed in the playbook.

PR #9804, opened 2021-04-03
instance groups options for modules

sean-m-sullivan:
##### SUMMARY
Add ability to set instance groups in organization, job templates, and inventories. 

#9788 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
- awx-collection

##### AWX VERSION
```
18.0.0
```


awx_collection/tests/integration/targets/tower_inventory/tasks/main.yml
```diff
@@ -25,17 +26,30 @@
         that:
           - "result is changed"
 
+    - name: Create an Instance Group
+      tower_instance_group:
+        name: "{{ group_name1 }}"
+        state: present
+      register: result
```
beeankha: This needs to be something like `instance_result`, otherwise the `insights_credential: "{{ result.id }}"` line in the following task can't find the proper Insights credential

beeankha (approved):
Left a comment; after that change is addressed, we are good to go! :+1:

PR #9810, opened 2021-04-05
Fix docker-compose up options

jogleasonjr:
##### SUMMARY
Should the `up` options be after the `up` command? I'm assuming `COMPOSE_UP_OPTS` exists so we can run detached like so:

`COMPOSE_UP_OPTS=-d make docker-compose`

If I am misunderstanding the purpose of `COMPOSE_UP_OPTS`, is there other guidance on how to autostart AWX detached?

##### ISSUE TYPE
 - Bugfix Pull Request
 

##### COMPONENT NAME
 - Installer

##### AWX VERSION
```
awx: 18.0.0
```


shanemcd (approved):
Thanks!

Issue #9817, opened 2021-04-05
CSP violation errors on job details and job settings views


jakemcdermott:
##### ISSUE TYPE
 - Bug Report

##### ENVIRONMENT
- dev environment, `make ui-devel`, port `8043`

##### STEPS TO REPRODUCE
- run `make ui-devel` to create a production bundle of the ui
- using port `8043` in the dev environment, navigate to the job details panel of a job template run or go to the job settings page

##### EXPECTED RESULTS
- no CSP errors in console

##### ACTUAL RESULTS
- CSP error in console

##### ADDITIONAL INFORMATION
After some debugging by @jbradberry @ryanpetrello , appears to be related to [this code](https://github.com/ajaxorg/ace/blob/94422a4a892495564c56089af85019a8f8f24673/lib/ace/worker/worker_client.js#L46) cc @ryanpetrello @keithjgrant

tiagodread:
I can't see these CSP errors in console anymore

PR #9818, opened 2021-04-05
Configure ace editor to not use blob worker

jakemcdermott:
##### SUMMARY
Intended to address #9817

### DIFF
--- a/awx/ui_next/src/components/CodeEditor/CodeEditor.jsx
+++ b/awx/ui_next/src/components/CodeEditor/CodeEditor.jsx
@@ -1,5 +1,7 @@
 import React, { useEffect, useRef, useCallback } from 'react';
 import { oneOf, bool, number, string, func, oneOfType } from 'prop-types';
+import { config } from 'ace-builds';
+
 import ReactAce from 'react-ace';
 import 'ace-builds/src-noconflict/mode-json';
 import 'ace-builds/src-noconflict/mode-javascript';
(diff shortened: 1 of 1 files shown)


awx/ui_next/src/components/CodeEditor/CodeEditor.jsx
```diff
@@ -11,6 +13,8 @@ import { t } from '@lingui/macro';
 import styled from 'styled-components';
 import debounce from '../../util/debounce';
 
+config.set('loadWorkerFromBlob', false);
+
```
jakemcdermott, line 17: For more context, see:

- https://github.com/ansible/awx/issues/9817
- [worker_client.js#L47](https://github.com/ajaxorg/ace/blob/master/lib/ace/worker/worker_client.js#L47)
- https://github.com/ajaxorg/ace/issues/3846
- https://github.com/ajaxorg/ace/pull/3862/files

awx/ui_next/src/components/CodeEditor/CodeEditor.jsx
```diff
@@ -11,6 +13,8 @@ import { t } from '@lingui/macro';
 import styled from 'styled-components';
 import debounce from '../../util/debounce';
 
+config.set('loadWorkerFromBlob', false);
```
jakemcdermott, line 16: This seems to address the CSP error. My only concern is unknowingly breaking some functionality w/ the code editors somewhere (although I think we're _probably_ fine).

cc @keithjgrant

ryanpetrello (approved):

unlikelyzero:
@jakemcdermott just needs e2e 🟢

PR #9819, opened 2021-04-05
Update pod reaper to work with receptor launched pods

shanemcd:


ryanpetrello (approved):
Yep, this makes sense to me.

AlanCoding (approved):
Thanks, I can clearly see the AWX_CONTAINER_GROUP_DEFAULT_IMAGE fix, as well as some other code structure changes and these all look like solid improvements

shanemcd:
Thanks for the reviews. Going to run this through some tests and merge before we release tomorrow.

Issue #7775, opened 2020-07-30
Remove Inventory Scripts


unlikelyzero:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
We are removing inventory scripts from awx. This ticket tracks the API work necessary to do that for docs and QE.

##### Related Issues
Related ui_next issue: https://github.com/ansible/awx/pull/7577
Related docs issue referenced below
Related tests issue referenced below

unlikelyzero:
@AlexSCorey @mabashian This was on the list of todo's for feature parity. Do we want to keep this ticket https://github.com/ansible/awx/issues/9267 open?

AlanCoding:
@wenottingham are we thinking about any migration solution for this?

wenottingham:
If practical: dump the script to the local filesystem somewhere just as a reference. No migration of sources; they will fail to update on upgrade.

AlanCoding:
@jbradberry what should expectations be for import/export where a user made use of custom inventory scripts in a prior version?

jbradberry:
@AlanCoding to my best recollection, inventory scripts are not even supported in awx-cli import/export.  Do we care enough about these things to leave an API endpoint active (and marked as deprecated) for a release cycle, and make them available to export?  I'm thinking that the answer is probably 'no', but if we do, I don't imagine it would be difficult to support the export.

AlanCoding:
The PR has been updated with this approach for the migration:

 - Inventory sources of the "custom" type are deleted in the migration, and hosts/groups from them remain and can be manually edited
 - The `CustomInventoryScript` model & table will be kept, but with ForeignKeys removed
   - Users will be able to export a `.tar` file with the scrips with an `awx-manage` command (name `export_custom_scripts` for now)

elyezer:
All verification was done pre-merge. Also tests were updated to make sure we can test both 4.0+ or <4.0 versions. I waited a few days to make sure we would not have issues after the merge of the related tower-qa PR.

There is one test failure that I suspect that may be related to this, I will double check and close this after addressing it.

elyezer:
Final checks were made and we are good to get this one closed.

PR #9822, opened 2021-04-06
Remove custom inventory scripts from the API

AlanCoding:
##### SUMMARY
Connect https://github.com/ansible/awx/issues/7775

##### AWX VERSION
```
18.0.0
```


elyezer:
@AlanCoding with these changes I am getting:

```awxkit.exceptions.BadRequest: Bad Request (400) received - {'source_script': ["If 'source' is 'custom', 'source_script' must be provided."]}```

When using awxkit and passing source as `custom` when creating an inventory source. I was expecting to get it as an invalid field with these changes, should we consider that here?

**Comment by AlanCoding** [comment]
Yes, sorry, that part just isn't finished yet.

File: awx/main/tests/functional/analytics/test_counts.py [hunk 1]
```diff
@@ -12,7 +12,6 @@ def test_empty():
         "active_sessions": 0,
         "active_host_count": 0,
         "credential": 0,
-        "custom_inventory_script": 0,
```
AlanCoding, line 15: @rooftopcellist I'm not yet sure if what I've done here works with analytics. It would be trivial to keep these keys but return a count of 0 every time (so that the schema doesn't change). I just need to figure out how to test and assure that it still works.

AlanCoding:
recheck

awx/main/management/commands/export_custom_scripts.py
```diff
@@ -0,0 +1,36 @@
...
+
+                for cis in CustomInventoryScript.objects.all():
+                    # naming convention similar to project paths
+                    slug_name = slugify(str(cis.name)).replace(u'-', u'_')
+                    script_filename = u'_%d__%s' % (int(cis.pk), slug_name)
+                    script_path = os.path.join(tmpdirname, script_filename)
+
+                    with open(script_path, 'w') as f:
+                        f.write(cis.script)
+                    os.chmod(script_path, stat.S_IRWXU)
```
AlanCoding, line 33: @wenottingham Important bit here, this was modeled after your suggestions. But `oct(stat.S_IRWXU) == 0o700`, as opposed to 600, because we need the executable bit set.

What this looks like in the test container:

```
$ tar -tvf my_scripts.tar 
-rwx------ awx/root        912 2021-04-19 12:57 _17__inventory_script_bidstory
...
```

jbradberry (approved):
Everything that I can think to look for appears covered here.

beeankha (approved):

jbradberry:
recheck

jbradberry:
regate

PR #9847, opened 2021-04-07
Fix elapsed time on job showing incorrect value

fosterseth:
##### SUMMARY

Elapsed time would always stay at zero

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 18.0.0
```

### DIFF
--- a/awx/main/models/unified_jobs.py
+++ b/awx/main/models/unified_jobs.py
@@ -851,7 +851,7 @@ class UnifiedJob(
         # out the time that elapsed, do so.
         if self.started and self.finished and self.elapsed == 0.0:
             td = self.finished - self.started
-            self.elapsed = (td.microseconds + (td.seconds + td.days * 24 * 3600) * 10 ** 6) / (10 ** 6 * 1.0)
+            self.elapsed = str((td.microseconds + (td.seconds + td.days * 24 * 3600) * 10 ** 6) / (10 ** 6 * 1.0))
             if 'elapsed' not in update_fields:
                 update_fields.append('elapsed')
 
--- a/awx/main/models/unified_jobs.py
+++ b/awx/main/models/unified_jobs.py
@@ -4,6 +4,7 @@
 # Python
 from io import StringIO
 import datetime
+import decimal
 import codecs
 import json
 import logging
@@ -842,8 +843,9 @@ class UnifiedJob(
             if 'finished' not in update_fields:
                 update_fields.append('finished')
 
+        dq = decimal.Decimal('1.000')
         if self.elapsed is None:
-            self.elapsed = 0.0
+            self.elapsed = decimal.Decimal(0.0).quantize(dq)
             if 'elapsed' not in update_fields:
                 update_fields.append('elapsed')
 
@@ -851,7 +853,8 @@ class UnifiedJob(
         # out the time that elapsed, do so.
         if self.started and self.finished and self.elapsed == 0.0:
             td = self.finished - self.started
-            self.elapsed = str((td.microseconds + (td.seconds + td.days * 24 * 3600) * 10 ** 6) / (10 ** 6 * 1.0))
+            elapsed = decimal.Decimal(td.total_seconds())
+            self.elapsed = elapsed.quantize(dq)
             if 'elapsed' not in update_fields:
                 update_fields.append('elapsed')
 
--- a/awx/main/models/unified_jobs.py
+++ b/awx/main/models/unified_jobs.py
@@ -846,8 +846,6 @@ class UnifiedJob(
         dq = decimal.Decimal('1.000')
         if self.elapsed is None:
             self.elapsed = decimal.Decimal(0.0).quantize(dq)
-            if 'elapsed' not in update_fields:
-                update_fields.append('elapsed')
 
         # If we have a start and finished time, and haven't already calculated
         # out the time that elapsed, do so.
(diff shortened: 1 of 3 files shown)


jakemcdermott:
nc

awx/main/models/unified_jobs.py
```diff
@@ -842,15 +842,16 @@ def save(self, *args, **kwargs):
...
+            if 'elapsed' not in update_fields:
+                update_fields.append('elapsed')
+
         # If we have a start and finished time, and haven't already calculated
         # out the time that elapsed, do so.
-        if self.started and self.finished and not self.elapsed:
+        if self.started and self.finished and self.elapsed == 0.0:
             td = self.finished - self.started
-            elapsed = (td.microseconds + (td.seconds + td.days * 24 * 3600) * 10 ** 6) / (10 ** 6 * 1.0)
-        else:
```
fosterseth, line 850: this line is the culprit

the issue is that on a job run we call `.save()` multiple times

when `started` and `finished` is defined and elapsed is 0.0, we set it (e.g. 5 seconds) and correctly save this field to the database

now on the very next `.save()`, the `not self.elapsed` condition fails, and we set elapsed back to 0.0

the `if self.elapsed != elapsed` condition is now true, `self.elapsed` (which is 5 seconds) is not equal to 0.0, so we save 0.0 to the database

fosterseth: And I suppose an additional call to .save() would set it back to 5 seconds. So it would toggle between 0.0 and 5 seconds on each save.

ryanpetrello: @chrismeyersfsu do you think this is related to the job parentage perf optimizations you made recently?  I can't think of any other change that would explain a side effect like this.

awx/main/models/unified_jobs.py
```diff
@@ -842,15 +843,18 @@ def save(self, *args, **kwargs):
             if 'finished' not in update_fields:
                 update_fields.append('finished')
 
+        dq = decimal.Decimal('1.000')
+        if self.elapsed is None:
+            self.elapsed = decimal.Decimal(0.0).quantize(dq)
+            if 'elapsed' not in update_fields:
+                update_fields.append('elapsed')
```
AlanCoding: I don't understand why you would want to add elapsed to update fields here. You don't get into this unless `self.elapsed` is None, but why would that happen? If the job isn't finished, I don't want to update the field, even if it's setting it to 0.

Maybe I could see why you'd want to initialize it to a `decimal.Decimal` object, but why do so only for the case where it's unset?

fosterseth: yeah I think update_fields is not needed in the `self.elapsed` `None` case. The `elapsed` DecimalField is non-null, so I think the column will be updated in this case regardless? Not sure what the consequence of having `update_fields` here, but it works without it so I'll remove it.

> Maybe I could see why you'd want to initialize it to a decimal.Decimal object, but why do so only for the case where it's unset?

I cast `self.elapsed` to Decimal in both cases -- where it's initialized to 0.0, and below where it's calculated based on `started` and `finished`

fosterseth: but really converting elapsed to Decimal isn't going to prevent the awx_collections tests from breaking as I had hoped. If people would rather return a float here I'm okay with that too

awx/main/models/unified_jobs.py
```diff
@@ -842,15 +843,16 @@ def save(self, *args, **kwargs):
...
-        if self.started and self.finished and not self.elapsed:
+        if self.started and self.finished and self.elapsed == 0.0:
             td = self.finished - self.started
-            elapsed = (td.microseconds + (td.seconds + td.days * 24 * 3600) * 10 ** 6) / (10 ** 6 * 1.0)
-        else:
-            elapsed = 0.0
-        if self.elapsed != elapsed:
-            self.elapsed = str(elapsed)
+            elapsed = decimal.Decimal(td.total_seconds())
+            self.elapsed = elapsed.quantize(dq)
```
AlanCoding, line 855: I would only have 1 minor tweak to suggest - before hitting this line, add another conditional line like

```python
if elapsed != self.elapsed:
```

Otherwise, it looks like every save after a job finished would add this `elapsed` field to the payload for updating the row. It would be preferable and intuitive that it updates the field exactly once, on the first save after the `finished` field is filled in.

AlanCoding (approved):
I'm on board now. Thanks

awx/main/models/unified_jobs.py
fosterseth: I think `and self.elapsed == 0.0` in the conditional should cover this. Once we hit this code block, self.elapsed will be non-zero, and any subsequent calls to `.save()` should skip this code.

To confirm this, we can add log messages for when we save and when we update elapsed. You can see it saves three times, but updates elapsed once.

```bash
[4bcbe8ea] awx.main.models.unified_jobs ====== in save
[4bcbe8ea] awx.main.models.unified_jobs ===== updating elapsed
[4bcbe8ea] awx.analytics.job_lifecycle job-752 finalize run
[4bcbe8ea] awx.main.models.unified_jobs ====== in save
[4bcbe8ea] awx.main.models.unified_jobs ====== in save

```

jbradberry (approved):

Issue #9039, opened 2021-01-07
Implement PostgreSQL event partitioning for better performance and manageability


ryanpetrello:
Users of awx are storing increasingly large amounts of event data (which are emitted any time an Ansible playbook, inventory, or any other automation unit is run).  Currently, events are stored in a set of tables by category, which by default grow in a fairly unbounded way:

```sql
main_adhoccommandevent
main_inventoryupdateevent
main_projectupdateevent
main_systemjobevent
main_jobevent
```

Over time, our largest users have begun to encounter growing pains managing this data.  A number of critical components of AWX interact with these tables.  For example:

1.  Parallel playbook runs write into these tables potentially at very high rates.  As table size grows over time, it becomes increasingly time and disk consumptive to maintain very large B-Tree indexes for important columns on these tables.

2. A number of internal ORM properties utilize these tables to determine if a Job is "finished" (e.g., `UnifiedJob.event_processing_finished`).  Even though we employ indexes for many of these queries, performance is affected when you venture into the realm of billions of rows, and if you're running these queries with any regularity (which can happen on certain stdout views).

3. Various APIs are available, and used by the UI to drive the display of historical job output.  This is almost exclusively the type of access that is indexed by a specific Job identifier (i.e., show me all the events for Job XYZ, sorted by e.g., line number).  This too can be extremely expensive at large scale: https://github.com/ansible/awx/issues/7405.
 
4. Event data can optionally be collected (and shipped to third party services) via AWX's analytics gathering framework.  This is generally done in large chunks, which can be expensive to read from large, long-lived tables.

5.  Cleaning up this data is also expensive and time consuming, because it involves indexed queries and bulk updates against tables that are actively being written to.  Additionally, deleting large chunks of singular tables requires periodic table vacuuming, which is also very expensive at this scale.

Moving to PostgreSQL 12 will *help*, because there have been some notable optimizations, especially in index utilization and query planning.  However, that's likely just a band-aid.  All of these operations become increasingly more expensive and time consuming as these tables continue to scale in an unbounded way into billions or tens of billions of rows.

This data is very much _write-once_ (in other words, it's almost never modified after the fact).  To alleviate some of these concerns, we'd like to explore introducing PostgreSQL table partitioning for these tables, so that we can segment this data into more manageable, logical partitions:

https://www.postgresql.org/docs/12/ddl-partitioning.html

[...]

PR #9856, opened 2021-04-08
Prevent analytics collector from missing events when tower is under stress

jladdjr:
Adapts https://github.com/ansible/awx/pull/9358#issuecomment-797747525 to work in the context of awx using partitioned job events tables (see https://github.com/ansible/awx/issues/9039 and https://github.com/ansible/awx/pull/9341).

### DIFF
--- a/awx/main/analytics/collectors.py
+++ b/awx/main/analytics/collectors.py
@@ -109,10 +109,9 @@ def events_slicing(key, since, until):
     # and all job events are stored in partitions,
     # slice by partition
     if lower == horizon and lower > partition_epoch:
-        # get list of partitions in this time frame
-        # .. and yeld them
         for partition in get_partitions():
-            yield (partition, None)
+            if partition_within_horizon(partition, horizon):
+                yield (partition, None)
         return
 
     # JobEvent.modified index was created at the partition epoch
--- a/awx/main/analytics/collectors.py
+++ b/awx/main/analytics/collectors.py
@@ -347,29 +347,29 @@ def _copy_table(table, query, path):
 @register('events_table', '1.2', format='csv', description=_('Automation task records'), expensive=events_slicing)
 def events_table(since, full_path, until, **kwargs):
     def query(event_data):
-        return f'''COPY (SELECT main_jobevent.id, 
+        return f'''COPY (SELECT main_jobevent.id,
                          main_jobevent.created,
                          main_jobevent.modified,
                          main_jobevent.uuid,
                          main_jobevent.parent_uuid,
-                         main_jobevent.event, 
+                         main_jobevent.event,
                          {event_data}->'task_action' AS task_action,
                          (CASE WHEN event = 'playbook_on_stats' THEN event_data END) as playbook_on_stats,
-                         main_jobevent.failed, 
-                         main_jobevent.changed, 
-                         main_jobevent.playbook, 
+                         main_jobevent.failed,
+                         main_jobevent.changed,
+                         main_jobevent.playbook,
                          main_jobevent.play,
                          main_jobevent.task,
-                         main_jobevent.role, 
-                         main_jobevent.job_id, 
-                         main_jobevent.host_id, 
+                         main_jobevent.role,
+                         main_jobevent.job_id,
+                         main_jobevent.host_id,
                          main_jobevent.host_name,
                          CAST({event_data}->>'start' AS TIMESTAMP WITH TIME ZONE) AS start,
                          CAST({event_data}->>'end' AS TIMESTAMP WITH TIME ZONE) AS end,
                          {event_data}->'duration' AS duration,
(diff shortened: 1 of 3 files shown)


jladdjr:
cc @chrismeyersfsu @Ladas @jbradberry @kdelee

awx/main/analytics/collectors.py
```diff
@@ -4,6 +4,8 @@
 import os.path
 import platform
 import distro
+import pytz
```
jbradberry, line 7: You should probably use the `utc` from django.utils.timezone instead of importing pytz.

awx/main/analytics/collectors.py
```diff
@@ -78,7 +80,51 @@ def events_slicing(key, since, until):
...
+    def partition_within_horizon(partition, horizon):
+        dt_str = partition[14:]
+        dt = datetime.datetime.strptime(dt_str, '%Y%m%d_%H').replace(tzinfo=pytz.UTC)
+        return dt > horizon
+
+    # TODO: Make default once unpartitioned tables are no longer supported
+    # if we are collecting job events starting at the horizon
+    # and all job events are stored in partitions,
+    # slice by partition
+    if lower == horizon and lower > partition_epoch:
```
jbradberry: I don't agree with limiting this optimization to the case where the lower bound is exactly at the 4-week limit.

jladdjr: agreed, rewriting this 👍

awx/main/analytics/collectors.py
```diff
@@ -78,16 +104,26 @@ def events_slicing(key, since, until):
...
     if not since and last_entries.get(key):
         lower = horizon
-    pk_values = models.JobEvent.objects.filter(created__gte=lower, created__lte=until).aggregate(Min('pk'), Max('pk'))
+    partition_epoch = get_event_partition_epoch()
 
-    previous_pk = pk_values['pk__min'] - 1 if pk_values['pk__min'] is not None else 0
-    if not since and last_entries.get(key):
-        previous_pk = max(last_entries[key], previous_pk)
-    final_pk = pk_values['pk__max'] or 0
+    # Stage 1: Process any job events in unpartitioned job event table
```
jladdjr, line 109: @jbradberry this is still a work in progress. We still need to get some data on the (1) only collect data by scooping up partitions approach vs the (2) get partitions when going up to the horizon, otherwise collect by pk / modified field.

This is written to follow the second approach, but I'd actually like to have it so that we can toggle between both options to make testing at scale easier.

awx/main/analytics/core.py
```diff
@@ -52,7 +52,7 @@ def all_collectors():
     }
 
 
-def register(key, version, description=None, format='json', expensive=None):
+def register(key, version, description=None, format='json', expensive=None, compare_key=None):
```
jladdjr, line 55: @jbradberry I'm extending this a bit so that, when needed, we can customize how we compare entries. The key function for job events is [here](https://github.com/ansible/awx/pull/9856/files#diff-5e7230fef85166a18967e2216535cd734d66de8b26983c303981e3c82b6fa1afR383).

awx/main/tests/functional/analytics/test_collectors.py
```diff
@@ -72,6 +73,29 @@ def write_stdout(self, sql, fd):
     return path
 
 
+@pytest.mark.django_db
```
jladdjr, line 76: I'm still working on the unit tests, but want to get several of these in so that we can make sure we're collecting job events the way we expect.

awx/main/analytics/collectors.py
```diff
@@ -344,39 +380,56 @@ def _copy_table(table, query, path):
     return file.file_list()
 
 
-@register('events_table', '1.2', format='csv', description=_('Automation task records'), expensive=events_slicing)
+def jobevent_lastentry_cmp(lastentry):
```
jladdjr, line 397: When `analytics/core.py` uses `max` to compare last entries for job events, we pass this key function in to ensure that pk-based entries come before partition table entries.

awx/main/analytics/core.py
```diff
@@ -243,7 +246,7 @@ def gather(dest=None, module=None, subset=None, since=None, until=None, collecti
                     if not files:
                         if collection_type != 'dry-run':
                             with disable_activity_stream():
-                                last_entries[key] = max(last_entries[key], end) if last_entries.get(key) else end
+                                last_entries[key] = max(last_entries[key], end, key=compare_key) if last_entries.get(key) else end
```
jbradberry, line 249: The support for `key=None` was only added in Python 3.8.

jbradberry (changes requested):
I don't see anything here that addresses the concern of what should happen with dirty partitions that have been marked as already shipped.

Issue #7662, opened 2020-07-20
Support expanded rows on Templates List


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Here's a link to the mockups: https://tower-mockups.testing.ansible.com/patternfly/templates/templates-expand/

unlikelyzero:
@trahman73 @mabashian is this necessary for launch?

mabashian:
This already exists with the new tables:

<img width="1488" alt="Screen Shot 2021-02-03 at 4 31 36 PM" src="https://user-images.githubusercontent.com/9889020/106812223-59c43080-663d-11eb-811b-8deba632fef5.png">

MrBones757:
Hello!
Appreciate the improvement!
Is the idea here is to replicate the old "Expanded" view from the old ui to ui_next?
If possible also showing credentials, and a list of labels in line with the previous ui would be awesome to see! (as shown in the mockup)
If this is not in scope for this PR, id be happy to look at adding this as a separate PR but would need some guidance / advice.

unlikelyzero:
Putting this back into parity and devel so that we can track the missing fields

mabashian:
Fields in the old UI:

```
<div class="at-Row-container">
                        <at-row-item
                            header-value="{{ template.name }}"
                            header-state="templates.editJobTemplate({job_template_id:{{template.id}}})"
                            header-tag="{{ vm.templateTypes[template.type] }}"
                            ng-if="template.type === 'job_template'">
                        </at-row-item>
                        <at-row-item
                            header-value="{{ template.name }}"
                            header-state="templates.editWorkflowJobTemplate({workflow_job_template_id:{{template.id}}})"
                            header-tag="{{ vm.templateTypes[template.type] }}"
                            ng-if="template.type === 'workflow_job_template'">
                        </at-row-item>
                        <at-row-item
                            smart-status="template">
                        </at-row-item>
                    </div>
                    <div class="at-Row-actions">
                        <at-launch-template template="template"
                            ng-show="template.summary_fields.user_capabilities.start">
                        </at-launch-template>
                        <at-row-action icon="fa-copy" ng-click="vm.copyTemplate(template)"
                            ng-show="!vm.isPortalMode && template.summary_fields.user_capabilities.copy"
                            tooltip="{{:: vm.strings.get('listActions.COPY', vm.getType(template)) }}">
                        </at-row-action>
[...]

PR #9858, opened 2021-04-08
Add missing template list expanded section details

marshmalien:
##### SUMMARY
Issue: #7662

* Add the following detail items:
  * Description
  * Organization and link to organization details
  * Credentials
* Move Credential and Label details to the bottom of the expanded section and make them stretch the entire width of the row. 

Screen Shot 2021-04-07 at 1 58 51 PM

##### COMPONENT NAME

 - UI

### DIFF
--- a/awx/ui_next/src/components/TemplateList/TemplateListItem.jsx
+++ b/awx/ui_next/src/components/TemplateList/TemplateListItem.jsx
@@ -231,11 +231,13 @@ function TemplateListItem({
                 value={template.description}
                 dataCy={`template-${template.id}-description`}
               />
-              <Detail
-                label={i18n._(t`Activity`)}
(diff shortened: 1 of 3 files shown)


nixocio (approved):
@marshmalien 

I triggered the e2e tests. 

I noticed that we are displaying empty `activity`. In general we do not display empty fields, but for the activity one make sense. Right?
Also the EE field would be a good one to be displayed as part of the expand section details. It was introduced later on.

marshmalien:
@nixocio We probably didn't show an empty "Activity" detail intentionally, so now we should check for recent_jobs before rendering "Activity". I also added Execution Environment detail to the expanded content and fixed a bug where short-circuit falsy values rendered 0 instead of nothing. 
Screen Shot 2021-04-09 at 3 36 38 PM

tiagodread (approved):

PR #9862, opened 2021-04-09
Use built-in suppress from contextlib

jakemcdermott:
In python3, we use the [built-in suppress](https://docs.python.org/3/library/contextlib.html#contextlib.suppress) from contextlib.


ryanpetrello (approved):
I like it, and I like you.

Issue #9769, opened 2021-03-31
EE RBAC violation when EE org changes


chrismeyersfsu:
##### STEPS TO REPRODUCE

* Create two orgs "Org 1" and "Org 2"
* Create a global EE
* Create a JT in "Org 1"
* Assign the global EE to the JT
* Change the EE, assign "Org 2" to the EE
* Run the JT

##### EXPECTED RESULTS

* For the JT to fail to launch because the associated EE is in a different Org than the JT or maybe for the JT to use the default EE since it doesn't have permission to the explicitly defined one.

##### ACTUAL RESULTS

* JT runs

This problem generalizes to ASSIGN time permissions/RBAC vs. runtime permissions/RBAC.

#### User Story
For example. Say I am a company and I have created a global EE, call it "My Networking Toolbox". A few JT explicitly use this EE. My company is growing, we start to create multiple Organizations, we now have an "IT" org and a "Networking" org. I now want to restrict the usage of "My Networking Toolbox" to the "Networking" org so I update the EE accordingly. This is where the problem comes in. Existing JT that are NOT in the "Networking" org still point at the "My Networking Toolbox" EE. New JT living outside of the "Networking" org are NOT allowed to point at "My Networking Toolbox".

#### Ideas
* Do nothing. Tell the user to create a new EE record if they want to do this.
* Check Jobs <-> EE RBAC at runtime and block running if violation
* Do not allow EE to change Org relationship
* Do not allow EE to change Org relationship if assigned to at least 1 job (similar behavior CredentialType)

jakemcdermott:
> Do not allow EE to change Org relationship

+1

tiagodread:
Verified that:

- EEs with org associated can be updated and become global
- EEs globals can not be updated to set an organization
- EEs with org A can not be updated to use org B

PR #9874, opened 2021-04-09
Prevent execution environment from being assigned to a new organization

fosterseth:
##### SUMMARY

related https://github.com/ansible/awx/issues/9769
- ee organization can be changed to null (less restrictive)
- if organization is null, cannot be assigned to org (more restrictive)
- if org is assigned, it cannot be set to a different org

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.0.0
```

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -1412,6 +1412,14 @@ class ExecutionEnvironmentSerializer(BaseSerializer):
             res['credential'] = self.reverse('api:credential_detail', kwargs={'pk': obj.credential.pk})
         return res
 
+    def validate(self, attrs):
+        # prevent changing organization of ee. Unsetting (change to null) is allowed
+        if self.instance:
+            org = attrs.get('organization', None)
+            if org and org.pk != self.instance.organization_id:
+                raise serializers.ValidationError({"organization": _("Cannot change the organization of an execution environment")})
+        return super(ExecutionEnvironmentSerializer, self).validate(attrs)
+
 
 class ProjectSerializer(UnifiedJobTemplateSerializer, ProjectOptionsSerializer):
 


nixocio:
I tested the following scenarios  - logged in as system admin.

1. Being able to assign an Org to a Globally Available EE. 
2. Not being able to remove an Org from an EE - to make the EE globally available.
3. Not being able to modify the Org already assigned to an EE.

I will work on the UI side to mimic those changes.

nixocio (approved):

awx/api/serializers.py
```diff
@@ -1412,6 +1412,18 @@ def get_related(self, obj):
...
         return res
 
+    def validate(self, attrs):
+        # if ee already has organization, cannot change organization
+        # or set it to null
+        if self.instance and self.instance.organization_id is not None:
+            if 'organization' in attrs:
+                org = attrs.get('organization', None)
+                if not org:
+                    raise serializers.ValidationError({"organization": _("Cannot remove the execution environment from the organization")})
```
chrismeyersfsu: Setting it to NULL should be OK. In this case you are moving from more restrictive to less restrictive.

Anyone can use EE w/ org Null.

jbradberry: I concur.

fosterseth: we currently allow going from null to non-null. So in essence a user could change from org-3 to org-2 by first setting to null, then setting to org-2.

chrismeyersfsu: We shouldn't allow going from NULL to ORG1. Resources that point to the EE when it was NULL that are outside of ORG1 are now using the EE, outside of the org they are allowed to.

fosterseth: thanks, changed it to prevent going from null to an org. But can go from org to null

chrismeyersfsu (approved):

jbradberry (approved):

PR #9878, opened 2021-04-11
Add workflow approval and node wait modules

sean-m-sullivan:
##### SUMMARY
Add a module to find a workflow approval node and approve or deny it, based on Issue #8013. 

Add a module to wait for a specific workflow node to complete and return information on it. 

Both of these are based on tests I have been creating for testing workflows. 
Scenario
Launch workflow
Wait for A node in the workflow to finish, compare output to expected output. 
If it matches, approve the approval node, otherwise deny the approval node.
Workflow completes. 

Even used in concert I've added the wait feature to both of these so a user can wait on either to appear. 
This does require a workflow to use unique names on the job nodes they are waiting on, As the job # is created on the fly, it would be difficult for user to specify, A future update could explore searching for a specific identifier among a workflow template and then finding that job created by that identifier. 

Currently without the modules this depends on generous use of the uri module, with until and retry coupled together. 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
awx-collection

##### AWX VERSION

```
19.0.0
```


shanemcd:
The diff is really noisy with all of the quoting changes. Is there a reason why that was done?

shanemcd:
We want to avoid these kinds of commits: 
image

Please use `git rebase` instead of merging the upstream remote into your branch. Given the name of the remote/branch in the commit message above, you probably want to do something like this:

```
$ git fetch ansible
$ git rebase ansible/devel
```

sean-m-sullivan:
@shanemcd I think its noisy due to the I ran black8 on the tower_api file, I could try and redo this into another branch/pr, but I did the direct PR into my base dev branch, so not sure how to do the rebase now without deleting my repo clone, which would destroy the other PR

awx_collection/plugins/modules/tower_workflow_approval.py
```diff
@@ -0,0 +1,127 @@
+#!/usr/bin/python
+# coding: utf-8 -*-
+
+# (c) 2017, Wayne Witzel III <wayne@riotousliving.com>
```
beeankha: Year should be 2021, name should be Sean Sullivan :smile:

Issue #8641, opened 2020-11-19
Cannot disable a schedule for a JT with a missing project (and it keeps running)


AlanCoding:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
It's fairly easy to get into a state where a schedule is enabled with an invalid JT, and the schedule cannot be turned off, due to a validation error on the JT.

Even though the schedule is invalid, it keeps running on the schedule interval, and its jobs fail obviously.

##### ENVIRONMENT
* AWX version: 15.0.1

##### STEPS TO REPRODUCE

Create a JT

Create a 1-minute schedule for that JT

Delete the project used by the JT (or its inventory)

Try to disable the schedule in the UI

##### EXPECTED RESULTS

Schedule turns off

##### ACTUAL RESULTS

400 response

```json
{"resources_needed_to_start":["Job Template project is missing or undefined.","Job Template inventory is missing or undefined."]}
```

This is what you get hitting the disable button in the UI, which performs a PATCH to the schedule, setting the "enabled" field to False, and it includes the rest of the data.

The issue is on the API here, because the request made by the UI seems reasonable, and should be expected to succeed. Perhaps the API should only error on this type of validation if the schedule is being _enabled_.

##### ADDITIONAL INFORMATION

Similar to https://github.com/ansible/awx/issues/7521 by @chrismeyersfsu, seen in test runs, and creates background processing since the 1-minute schedule continues to run for the duration that the server is up.

kdelee:
This is fixed now, and does the reasonable thing -- allows schedule for the broken JT to be disabled, but blocks you from re-enabling it. Closing as verified

PR #9890, opened 2021-04-12
allow schedules to be disabled even if the associated UJT isn't valid

ryanpetrello:
see: https://github.com/ansible/awx/issues/8641


rooftopcellist (approved):
LGTM

jbradberry (approved):

PR #9892, opened 2021-04-12
record the receptor unit ID on the job record for debugging purposes

ryanpetrello:
see: https://github.com/ansible/awx/issues/8641

image

### DIFF
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -1394,6 +1394,7 @@ class BaseTask(object):
                 )
             else:
                 receptor_job = AWXReceptorJob(self, params)
+                self.unit_id = receptor_job.unit_id
                 res = receptor_job.run()
 
                 if not res:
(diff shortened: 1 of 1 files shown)


AlanCoding:
This won't be in the actual job environment, will it?

AlanCoding:
I think I've confirmed that (via https://github.com/ansible/test-playbooks/pull/153)

My thought was that a new `UnifiedJob` field would be a less confusing way to present that. Aside from system jobs, workflow jobs, and approval jobs, I think that all job types should have this non-blank.

ryanpetrello:
@AlanCoding it won't, no.  This JSON blob was just a convenient place to store the data without having to resort to a new column on a table that tends to be very large (schema changes on this table tend to be quite expensive).

If you all don't like this approach, that's fine, though.  I figured this PR was a cheap way to get us what we needed for debugging.

shanemcd (approved):
We will eventually probably need this in its own database column, but I think this is fine for now.

PR #9899, opened 2021-04-13
metadata: Fixes invalid model classes field.

### LINKED ISSUE #9441: Advanced search for type returns 400
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

Trying to do an advanced search by `type` the results are always `ExecutionEnvironment has no field named 'type'`:

image

### LINKED ISSUE #9898: Invalid `type` field in key select filter column in advanced search
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

Some model's classes do not contain `type` attributes or fields. Thus, filterable `type` should not be displayed in the key select filter in the advanced search.

Affected classes are as follows:
1. ActivityStream
2. CredentialType
3. ExecutionEnvironment
4. Host
5. InstanceGroup
6. Inventory
7. NotificationTemplate
8. OAuth2Application
9. Organization
10. Schedule
11. Team
12. Users

##### ENVIRONMENT
* AWX version: `devel`

##### STEPS TO REPRODUCE
1. Go to `schedules` section
2. Select advanced search
3. Select `type` key from the key select filter column
4. Type in any characters in the text input field
5. Click search

##### EXPECTED RESULTS
Should only display filterable key in the key select filter column

##### ACTUAL RESULTS
Display non-filterable key in the key select filter column

##### ADDITIONAL INFORMATION
Invalid_field_bug

#### Potential Solution
Set the `type` field's filterable as `True` only if the model class consists of the `type` attribute.

seiwailai:
Fixes #9441.

##### SUMMARY

Set `type`' field's filterable as `True` only if the model classes contain `type` field such as `UnifiedJob`, `WorkflowApproval`, `UnifiedJobTemplate`, `Project` and `SystemJobTemplate`

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API - metadata.py

##### AWX VERSION

```
awx: 19.0.0
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/api/metadata.py
+++ b/awx/api/metadata.py
@@ -26,6 +26,9 @@ from awx.main.fields import JSONField, ImplicitRoleField
 from awx.main.models import NotificationTemplate
 from awx.main.tasks import AWXReceptorJob
 
+# Polymorphic
+from polymorphic.models import PolymorphicModel
+
 
 class Metadata(metadata.SimpleMetadata):
     def get_field_info(self, field):
@@ -78,7 +81,9 @@ class Metadata(metadata.SimpleMetadata):
                 field_info['help_text'] = field_help_text[field.field_name].format(verbose_name)
 
             if field.field_name == 'type':
-                field_info['filterable'] = True
+                # Only include model classes with `type` field.
+                if issubclass(serializer.Meta.model, PolymorphicModel):
+                    field_info['filterable'] = True
             else:
                 for model_field in serializer.Meta.model._meta.fields:
                     if field.field_name == model_field.name:


shanemcd (changes requested):
Hello, thanks for the PR!

I believe we should be able to introspect the objects directly rather than constructing a hardcoded list like this. This should work:


```python
# snip...
from polymorphic.models import PolymorphicModel
# snip...

            if field.field_name == 'type':
                if issubclass(serializer.Meta.model, PolymorphicModel):
                    field_info['filterable'] = True
```

seiwailai:
> Hello, thanks for the PR!
> 
> I believe we should be able to introspect the objects directly rather than constructing a hardcoded list like this. This should work:
> 
> ```python
> # snip...
> from polymorphic.models import PolymorphicModel
> # snip...
> 
>             if field.field_name == 'type':
>                 if issubclass(serializer.Meta.model, PolymorphicModel):
>                     field_info['filterable'] = True
> ```

@shanemcd Thanks for your review!

awx/api/metadata.py
```diff
@@ -23,9 +23,12 @@
 # AWX
 from awx.api.fields import ChoiceNullField
 from awx.main.fields import JSONField, ImplicitRoleField
-from awx.main.models import NotificationTemplate
+from awx.main.models import NotificationTemplate, UnifiedJob, WorkflowApproval, UnifiedJobTemplate, Project, SystemJobTemplate
```
shanemcd: Can you revert this part? I don't think we need it anymore after your last commit.

seiwailai: Sure, thanks for your reminder!

shanemcd (approved):
Thank you! 😎

jakemcdermott:
regate

Issue #8777, opened 2020-12-07
awx collection workflow approval with success node fail


chris93111:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
apply success node to approval_node fail because success_nodes not exists in /api/v2/workflow_approval_templates/

"msg": "The requested object could not be found at /api/v2/workflow_approval_templates/280/success_nodes/."

##### ENVIRONMENT
* AWX version: 15.0.1
* AWX install method: docker
* Ansible version:  2.8.7
* Operating System: redhat
* Web Browser: chrome

##### STEPS TO REPRODUCE

approval_node with success_node

##### EXPECTED RESULTS

ansible tasks changed

##### ACTUAL RESULTS

ansible fail

ryanpetrello:
Hey @chris93111,

Can you give some clearer step-by-step instructions on how to reproduce this?  I'm having trouble understanding how to reproduce the issue.

ryanpetrello:
Also @chris93111,

Could you give this a shot in the latest AWX (16.0.0)?  We've recently rewritten the entire workflow editor interface, and if this is a bug in the old interface, it's possible that it's been resolved in a newer AWX version.

wenottingham:
/api/v2/workflow_approval_templates - > workflow_approval_templates is the wrong toplevel endpoint (it doesn't exist)

MaximilianMeister:
I also ran into this, even in a later version (19.0.0) and the only way I could make it happen was to first create all the workflow nodes and link them later, e.g.:

```yaml
- name: Create workflow with 2 Job Templates and an approval node in between
  block:
  - name: Create a workflow job template
    awx.awx.tower_workflow_job_template:
      name: my-workflow-job-template
      ask_scm_branch_on_launch: true
      organization: Default

  - name: Create 1st node
    awx.awx.tower_workflow_job_template_node:
      identifier: my-first-node
      workflow_job_template: my-workflow-job-template
      unified_job_template: some_job_template
      organization: Default

  - name: Create 2nd approval node
    awx.awx.tower_workflow_job_template_node:
      identifier: my-second-approval-node
      workflow_job_template: my-workflow-job-template
      organization: Default
      approval_node:
        description: "Do this?"
        name: my-second-approval-node
        timeout: 3600

  - name: Create 3rd node
    awx.awx.tower_workflow_job_template_node:
      identifier: my-third-node
      workflow_job_template: my-workflow-job-template
      unified_job_template: some_other_job_template
[...]

PR #9904, opened 2021-04-14
Document usage of workflow job template nodes with approvals

MaximilianMeister:
##### SUMMARY

I'm not sure if this procedure is mandatory, but it was the only way it worked for me.

It's already documented in [one sentence](https://github.com/ansible/awx/blob/devel/awx_collection/plugins/modules/tower_workflow_job_template_node.py#L24-L26) that you can create all workflow template nodes first and then link them later. However in the Examples it has been missing to make it explicit, as it's not really intuitive when you create the nodes and link them at the same time, it won't work and you run into https://github.com/ansible/awx/issues/8777

##### ISSUE TYPE

 - Docs Pull Request

##### AWX VERSION

```
awx: 19.0.0
```

### DIFF
--- a/awx_collection/plugins/modules/tower_workflow_job_template_node.py
+++ b/awx_collection/plugins/modules/tower_workflow_job_template_node.py
@@ -174,6 +174,54 @@ EXAMPLES = '''
     organization: Default
     success_nodes:
       - my-first-node
+
+- name: Create workflow with 2 Job Templates and an approval node in between
+  block:
+  - name: Create a workflow job template
+    tower_workflow_job_template:
+      name: my-workflow-job-template
+      ask_scm_branch_on_launch: true
+      organization: Default
+
+  - name: Create 1st node
+    tower_workflow_job_template_node:
+      identifier: my-first-node
+      workflow_job_template: my-workflow-job-template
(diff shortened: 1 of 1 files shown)


awx_collection/plugins/modules/tower_workflow_job_template_node.py
```diff
@@ -174,6 +174,54 @@
     organization: Default
     success_nodes:
       - my-first-node
+
+- name: Create workflow with 2 Job Templates and an approval node in between
+  block:
+  - name: Create a workflow job template
+    awx.awx.tower_workflow_job_template:
```
shanemcd: Can you drop the FQCNs? It should be consistent with the examples above, ie just `tower_workflow_job_template:`

MaximilianMeister: done, thanks for the hint

MaximilianMeister:
> Build failed.

can't say why it failed here, tried to rebase to devel but there are no changes. test was already green before

MaximilianMeister:
rebased, hopefully will fix the CI

AlanCoding (approved):
I put this into a playbook and ran it locally, everything checks out

Issue #9212, opened 2021-02-02
[execution-environments] Image cleanup


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Steps:

 - User creates a new EE that references a new custom image
 - Runs jobs with this EE
 - Deletes the EE through the API

Over time, the user accumulates bulky images, on the scale of 100s of Gb.

The API needs to provide some means of deleting both

 - Images no longer referenced by an execution environment in the API
 - Dangling images

If not, some users can run out of space.

AlanCoding:
Locally, I find that the podman prune command isn't thorough enough, I get by that running

```
podman rmi $(podman images --filter "dangling=true" -q) -f
```

and I have to run this multiple times, unfortunately. I don't know what to tell you for the AWX solution here.

jbradberry:
cc @unlikelyzero

unlikelyzero:
Might need to be a management job

shanemcd:
Some thoughts:

- This management job should only exist on installs where `settings.IS_K8S = False`.
- The command Alan mentioned in https://github.com/ansible/awx/issues/9212#issuecomment-771964900 should be good enough for us.

shanemcd:
@fosterseth You can see an example of how the original management jobs were created here: https://github.com/ansible/awx/blob/devel/awx/main/migrations/0002_squashed_v300_release.py#L32-L53

AlanCoding:
My for testing here is that we check to see if any unexpected images are present at the end of an integration test run.

Zokormazo:
Tested manually, the cleaner works.

https://github.com/ansible/tower-qa/issues/6497 for automation debt

PR #9907, opened 2021-04-14
Add cleanup_images scheduled task

fosterseth:
##### SUMMARY

https://github.com/ansible/awx/issues/9212
Removes dangling podman images on the system

`[70e6cc8a] awx.main.tasks Cleanup execution environment images: deleting quay.io/fosterseth/awx-ee:v1, 643 MB`

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.0.0
```


awx/main/management/commands/cleanup_images.py
```diff
@@ -0,0 +1,79 @@
...
+                    self.logger.info(f"{self.delete_prefix} {i}: {image_size:.0f} MB")
+                    if not self.dry_run:
+                        subprocess.run(['podman', 'rmi', i, '-f'], stdout=subprocess.DEVNULL)
+
+    def cleanup_images(self):
+        self.images_in_use = [ee.image for ee in ExecutionEnvironment.objects.all()]
+        self.logger.info(f"Execution environment images in use: {self.images_in_use}")
+        self.deleted = []
+        # find and remove unused images
+        images_system = subprocess.run("podman images -a --format json".split(" "), capture_output=True)
```
fosterseth: `podman images -a` is going to grab everything regardless if the images are related to EE. Right now it seems that podman is only being used to manage EE images (?), but if we start having other images on the system, this is not going to work.

ryanpetrello: It looks like podman has an RESTful API and a Python SDK:

https://github.com/containers/podman-py

This stuff may be too bleeding edge at the moment, though, (the git history suggests it's actively being developed) and I'm not clear on how you expose it via a local domain socket.

awx/main/management/commands/cleanup_images.py
```diff
@@ -0,0 +1,82 @@
...
+class Command(BaseCommand):
+    """
+    Management command to cleanup unused execution environment images.
+    """
+
+    help = 'Remove unused execution environment images'
+
+    def init_logging(self):
+        log_levels = dict(enumerate([logging.ERROR, logging.INFO, logging.DEBUG, 0]))
+        self.logger = logging.getLogger('awx.main.commands.cleanup_sessions')
```
ryanpetrello: `cleanup_session` copy pasta

awx/main/management/commands/cleanup_images.py
```diff
@@ -0,0 +1,82 @@
...
+            self.delete_images(images_system)
+        if not self.deleted:
+            self.logger.info("Did not find unused images to remove")
+
+    def handle(self, *args, **options):
+        self.verbosity = int(options.get('verbosity', 1))
+        self.init_logging()
+        self.dry_run = bool(options.get('dry_run', False))
+        if self.dry_run:
+            self.delete_prefix = "Would delete"
```
ryanpetrello: It's a little weird that this is an instance variable that's only used in a different method - it might just make sense to make that a conditional local variable over in `delete_images()`

fosterseth: fixed!

awx/main/management/commands/cleanup_images.py
fosterseth: i've been caught red hatted! fixed

awx/main/management/commands/cleanup_images.py
```diff
@@ -0,0 +1,83 @@
...
+                self.logger.info(f"\t{i}")
+        self.deleted = []
+        # find and remove unused images
+        images_system = subprocess.run("podman images -a --format json".split(" "), capture_output=True)
+        if len(images_system.stdout) > 0:
+            images_system = json.loads(images_system.stdout)
+
+            self.delete_images(images_system)
+        # find and remove dangling images
+        images_system = subprocess.run('podman images -a --filter "dangling=true" --format json'.split(" "), capture_output=True)
```
ryanpetrello: It's a little nitpick, but given that you're doing the same thing twice, you could probably clean this up a little with a loop i.e.,

```python
for cmd in (
    'podman images -a --format json',
    'podman images -a --filter "dangling=true" --format json',
):
    ...

File: awx/main/management/commands/cleanup_images.py [hunk 5]
```diff
@@ -0,0 +1,83 @@
...
+        parser.add_argument('--dry-run', dest='dry_run', action='store_true', default=False, help='Dry run mode (show items that would ' 'be removed)')
+
+    def delete_images(self, images_json):
+        if self.dry_run:
+            delete_prefix = "Would delete"
+        else:
+            delete_prefix = "Deleting"
+        for e in images_json:
+            if 'Names' in e:
+                image_names = e['Names']
```
ryanpetrello: It looks like `Names` can be `null` ?

image

Also, my keys are lowercased.  Does this differ based on podman version?

```
~ podman --version
podman version 1.8.2
```

awx/main/management/commands/cleanup_images.py
fosterseth: good idea. although from my testing it seems the first command includes all the images from the second command. So we might only need a single command to grab the images we care about. If we do need more than one command, I'll loop it like you've demonstrated

awx/main/management/commands/cleanup_images.py
```diff
@@ -0,0 +1,83 @@
...
+            else:
+                image_names = [e["Id"]]
+            image_size = e['Size'] / 1e6
+            for i in image_names:
+                if i not in self.images_in_use and i not in self.deleted:
+                    self.deleted.append(i)
+                    self.logger.info(f"{delete_prefix} {i}: {image_size:.0f} MB")
+                    if not self.dry_run:
+                        subprocess.run(['podman', 'rmi', i, '-f'], stdout=subprocess.DEVNULL)
```
ryanpetrello: _Just in case_ something goes wrong, it might be nice to have e.g.,

```python
process = subprocess.run(...)
if process.returncode != 0:
   logging_something()
```

awx/main/migrations/0136_cleanup_ee_images.py
```diff
@@ -0,0 +1,32 @@
...
+from django.db import migrations
+from django.utils.timezone import now
+
+
+def create_cleanup_ee_images(apps, schema_editor):
+    SystemJobTemplate = apps.get_model('main', 'SystemJobTemplate')
+    ContentType = apps.get_model('contenttypes', 'ContentType')
+    sjt_ct = ContentType.objects.get_for_model(SystemJobTemplate)
+    now_dt = now()
+    sjt, created = SystemJobTemplate.objects.get_or_create(
```
ryanpetrello: Do (or should) these run by default on a schedule, or do we want the user to do this themselves on demand?  If so, how often cc @shanemcd @wenottingham @kdelee

Also, if I recall, the ability view and launch system jobs in the *old* UI required manual code per template to implement.  Is the new UI more flexible/dynamic enough to just materialize this new system job and do the right thing cc @jakemcdermott?

ryanpetrello (approved):
I've mostly got nitpicky feedback to offer on this; I think the approach is pretty reasonable, and looks like it'll solve the ask.

ryanpetrello (changes requested):
@wenottingham brings up a good point - this being a system job, it's not going to "fanout" to all control plane nodes.

Also, down the line when we eventually have execution-only nodes, we'll have to brainstorm about how *that* cleanup would actually work.

For those types of tasks, in the past we've generally had "broadcast tasks" defined in `awx.main.tasks`, and we've set up a periodic schedule that runs them in the AWX settings.  That may be a better fit with the current architecture than a System Job.

awx/main/management/commands/cleanup_images.py
fosterseth: in `tools_awx_1`
```bash
bash-4.4$ podman --version
podman version 2.2.1
```

awx/settings/defaults.py
```diff
@@ -439,6 +439,7 @@ def IS_TESTING(argv=None):
     'task_manager': {'task': 'awx.main.scheduler.tasks.run_task_manager', 'schedule': timedelta(seconds=20), 'options': {'expires': 20}},
     'k8s_reaper': {'task': 'awx.main.tasks.awx_k8s_reaper', 'schedule': timedelta(seconds=60), 'options': {'expires': 50}},
     'send_subsystem_metrics': {'task': 'awx.main.analytics.analytics_tasks.send_subsystem_metrics', 'schedule': timedelta(seconds=20)},
+    'cleanup_images': {'task': 'awx.main.tasks.cleanup_execution_environment_images', 'schedule': timedelta(hours=8)},
```
fosterseth: @shanemcd what are your thoughts on this interval?

awx/main/management/commands/cleanup_images.py
fosterseth: fixed

awx/main/tasks.py
```diff
@@ -396,6 +398,27 @@ def purge_old_stdout_files():
...
+    images_system = subprocess.run("podman images -a --format json".split(" "), capture_output=True)
+    if len(images_system.stdout) > 0:
+        images_system = json.loads(images_system.stdout)
+        for e in images_system:
+            if 'Names' in e:
+                image_name = e['Names'][0]
+            else:
+                image_name = e["Id"]
+            image_size = e['Size'] / 1e6
+            if image_name not in images_in_use:
```
AlanCoding: What if my EE has set `quay.io/project-receptor/receptor` but the reported output from `podman images -a --format json` gives:

```json
    {
        "Id": "bf4de35112248c7c5b1db46c43f87346f31e2717850831433af1fee6dfd0c2b4",
        "ParentId": "",
        "Size": 387129820,
        "Labels": {
            "license": "ASL2",
            "name": "receptor",
            "vendor": "project-receptor",
            "version": "0.9.1"
        },
        "Names": [
            "quay.io/project-receptor/receptor:latest"
        ],
        "Digest": "sha256:11d9a25f07ea27146d94772eb3c6f2faf5d5c75d9c3313fcad7390c7e417c32a",
        "Digests": [
            "sha256:11d9a25f07ea27146d94772eb3c6f2faf5d5c75d9c3313fcad7390c7e417c32a"
        ],
        "Created": 1602187764,
        "CreatedAt": "2020-10-08T20:09:24Z"
    }
```

I think it's valid to not include `:latest` so I'm wondering if it might confuse this code.

It's also going to be tricky to validate this. If it incorrectly deletes images and layers that it shouldn't, it will be quite difficult for QE to track down what was _successfully_ deleted, but shouldn't have been (triggering an expensive download on the next run).

fosterseth: Good observation. We could assume that image names without a tag are using `latest` and check for those images on the system. The trick then is to find a reliable regex or something similar to parse the image name.

AlanCoding: https://docs.docker.com/engine/reference/commandline/tag/

> A tag name must be valid ASCII and may contain lowercase and uppercase letters, digits, underscores, periods and dashes. A tag name may not start with a period or a dash and may contain a maximum of 128 characters.

You'd need to rely on what you know is valid for the image URL & tag. I think it could work in exactly the way you say. It would be good to do some manual testing of different ways of specifying an image and making sure that this code still works as intended.

fosterseth: to complicate matters, users could also pull using a digest

`quay.io/fosterseth/awx-ee@sha256:cbb1256806ea16762ea157bfe52578070b900b607780eca196f57838dd2d7eff`

shanemcd (approved):

PR #9912, opened 2021-04-15
Return distinct labels for normal Users

fosterseth:
##### SUMMARY

1. Create a single label and add it to 2 or more JTs
2. Visit `/api/v2/labels` as superuser and you'll see a single label
3. Create a normal user and give admin role to each JT in step 1
4. Visit `/api/v2/labels` as this normal user and you will see duplicate entries for the label. Number of entries will be equal to the number of JTs in step 1

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.0.0
```

### DIFF
--- a/awx/main/access.py
+++ b/awx/main/access.py
@@ -2627,7 +2627,7 @@ class LabelAccess(BaseAccess):
         return self.model.objects.filter(
             Q(organization__in=Organization.accessible_pk_qs(self.user, 'read_role'))
             | Q(unifiedjobtemplate_labels__in=UnifiedJobTemplate.accessible_pk_qs(self.user, 'read_role'))
-        )
+        ).distinct()
 
     @check_superuser
     def can_add(self, data):


awx/main/access.py
```diff
@@ -2627,7 +2627,7 @@ def filtered_queryset(self):
         return self.model.objects.filter(
             Q(organization__in=Organization.accessible_pk_qs(self.user, 'read_role'))
             | Q(unifiedjobtemplate_labels__in=UnifiedJobTemplate.accessible_pk_qs(self.user, 'read_role'))
-        )
+        ).distinct()
```
fosterseth, line 2630: the second Q filter involves a Many To Many relationship, which has the possibility of returning duplicate label entries. We can call `distinct()` to get unique labels.

jbradberry (approved):

PR #9922, opened 2021-04-15
Always overwrite the contents of .git/hooks/pre-commit

jbradberry:
##### SUMMARY
Set the `make` command to always write into the pre-commit hook file, even if it already exists.  This will allow this file to be updated when changes are made without the developer having to jump through hoops or remember that it is a thing.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION

```

```


ryanpetrello (approved):
Seems reasonable to me 👍

PR #9908, opened 2021-04-14
Add Session context and redirects

marshmalien:
##### SUMMARY
Issues:
#8272 #9197 #8155 #9311

* When user is logged out due to inactivity, redirect to the Login Page with a warning alert stating that their session expired.
* When a user logs in on one tab, redirect from login page to home on the other tabs. 
* Clear session state when it expires and/or when a user logs out.
* Redirect users to direct link destination after authentication

~~@AlexSCorey is looking at possibly rolling [#9915](https://github.com/ansible/awx/issues/9115) (SSO login redirect override URL is not working) into this PR~~

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - UI
 
session

PR #9943, opened 2021-04-19
Fixes SSO Redirect

### LINKED ISSUE #9115: SSO login redirect override URL is not working
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
After upgrading to 16.0.0, SSO Login redirect override URL is no longer working.

"Settings->System->Miscellaneous System settings->Details" correctly shows "Login redirect override URL" setting (for example: /sso/login/saml/?idp=MYIDP), but editing is not possible and users are no longer automatically redirected.

##### ENVIRONMENT
* AWX version: 16.0.0
* AWX install method: docker on linux
* Operating System: any
* Web Browser: any

##### STEPS TO REPRODUCE
1. go to Settings->System->Miscellaneous System settings->Details
2. click on Edit
3. "Login redirect override URL" edit field is missing

##### EXPECTED RESULTS
1. go to Settings->System->Miscellaneous System settings->Details
2. click on Edit
3. enter data for "Login redirect override URL"
4. users automatically redirected to SSO

##### ACTUAL RESULTS
1. it's not possible to add/edit SSO override URL ("Login redirect override URL")
2. users are not redirected to SSO

##### ADDITIONAL INFORMATION

Also, would it be possible, when using SSO, to have SSL (SAML) icon and text dominantly displayed inside the login box? Currently, there is no text and the icon (UserCircleIcon) is too small. For SSO (SAML) users it is difficult to figure out that in order to login one needs to click on that tiny-microscopic-icon.

AlexSCorey:
##### SUMMARY
This fixes #9115 by simply checking if there is a redirect url in and then replacing it with the existing url in history, navigating the user to the correct login url.
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
redirect

### DIFF
--- a/awx/ui_next/src/screens/Login/Login.jsx
+++ b/awx/ui_next/src/screens/Login/Login.jsx
@@ -42,12 +42,18 @@ function AWXLogin({ alt, i18n, isAuthenticated }) {
     isLoading: isCustomLoginInfoLoading,
     error: customLoginInfoError,
     request: fetchCustomLoginInfo,
-    result: { brandName, logo, loginInfo, socialAuthOptions },
+    result: {
+      brandName,
+      logo,
+      loginInfo,
+      socialAuthOptions,
+      loginRedirectOverride,
+    },
   } = useRequest(
     useCallback(async () => {
       const [
         {
-          data: { custom_logo, custom_login_info },
+          data: { custom_logo, custom_login_info, login_redirect_override },
         },
         {
           data: { BRAND_NAME },
@@ -61,11 +67,13 @@ function AWXLogin({ alt, i18n, isAuthenticated }) {
       const logoSrc = custom_logo
         ? `data:image/jpeg;${custom_logo}`
         : loginLogoSrc;
+
       return {
         brandName: BRAND_NAME,
         logo: logoSrc,
         loginInfo: custom_login_info,
         socialAuthOptions: authData,
+        loginRedirectOverride: login_redirect_override,
       };
     }, []),
     {
@@ -84,7 +92,6 @@ function AWXLogin({ alt, i18n, isAuthenticated }) {
   useEffect(() => {
     fetchCustomLoginInfo();
   }, [fetchCustomLoginInfo]);
-
   const {
     isLoading: isAuthenticating,
     error: authenticationError,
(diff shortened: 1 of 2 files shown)


AlexSCorey:
recheck

tiagodread:
We are gonna have the https://github.com/ansible/awx/pull/9908 in first

awx/ui_next/src/screens/Login/Login.jsx
```diff
@@ -61,13 +61,16 @@ function AWXLogin({ alt, i18n, isAuthenticated }) {
       const logoSrc = custom_logo
         ? `data:image/jpeg;${custom_logo}`
         : loginLogoSrc;
+      if (login_redirect_override && !isAuthenticated(document.cookie)) {
+        return window.location.replace(login_redirect_override);
```
jakemcdermott: @AlexSCorey 

I like the simplicity. That said, burying this deep within an async callback (and conditionally changing its return structure) makes me a little uneasy.  To me, it's an unexpected thing to do in an asynchronous api response handler.

We have other conditional redirects further down in the component definition that use the typical `react-router` semantics and, if possible, I think we should do something similar.

I think all you'd need to do is add:

```jsx
  if (login_redirect_override) {
    return <Redirect to={login_redirect_override} />;
  }
```

to [this line](https://github.com/ansible/awx/blob/devel/awx/ui_next/src/screens/Login/Login.jsx#L115)

jakemcdermott (changes requested):

awx/ui_next/src/screens/Login/Login.jsx
jakemcdermott: or, if react-router doesn't like external urls:

```jsx
  if (login_redirect_override) {
    window.location.replace(login_redirect_override);
    return null;
  }
```
to [this line](https://github.com/ansible/awx/blob/devel/awx/ui_next/src/screens/Login/Login.jsx#L115)

AlexSCorey: React router does not like redirecting outside of the app.  Using option 1

jakemcdermott (approved):

one-t:
recheck

one-t:
Hey, @AlexSCorey - once the check failure is fixed I can take a look at this. I'll be out next week so we may need to reassign if I can't get it hammered out

one-t:
I've verified this is working correctly. SSO verification is a manual process currently (though I'd like to automate it in the future).

Issue #106, opened 2017-09-11
Allow for different role requirements file names/locations


geerlingguy:
### Summary

I was trying to figure out a way for AWX to automatically install role requirements from Ansible Galaxy prior to running my playbook in a new Project Template, but didn't see any way to do it.

My current project structure is:

```
project/
  requirements.yml #role requirements
  main.yml # playbook
```

But it looks like [according to the documentation](http://docs.ansible.com/ansible-tower/latest/html/userguide/projects.html#ansible-galaxy-support), Ansible Galaxy requirements will _only_ be installed if there is a file named `requirements.yml` inside a specifically-named `roles` directory.

### Environment

* AWX version: 1.0.0-[latest as of yesterday]
* Ansible version:  2.3.1.0
* Operating System: macOS 10.12
* Web Browser: Safari

### Steps To Reproduce:

  1. Create a project with the structure as shown above.
  1. Add a Template in AWX for this project.
  1. Run a new Job with the Template.

### Expected Results:

AWX should find my role requirements file in any location inside the repository, maybe as specified in the template itself.

### Actual Results:

```
ERROR! the role 'geerlingguy.firewall' was not found in /var/lib/awx/projects/_6__my_project/roles:/etc/ansible/roles:/var/lib/awx/projects/_6__my_project
```

My root `requirements.yml` file was not used to install Galaxy roles.

### Additional Information:

Another note—for some projects, I have an `ansible.cfg` in the project root that defines a specific roles path (which may not always be `./roles/`). It would be especially nice if AWX could figure out a way to parse that setting if present and download the roles into that location... though that could prove fruitless since that setting could be a set of many paths :-/

matburt:
We have definitely been opinionated in the past about where things like this get picked up. I'm fine with someone tackling this but it's not a priority for us.

We do see some *extremely* large directory trees in projects on user's systems so an exhaustive traversal of the entire project hierarchy will be rejected.

wenottingham:
To be fair, extremely large project directory trees are going to break with isolated nodes anyway.

geerlingguy:
To avoid the large directory tree issue, it would be nice to maybe add a directive to `ansible.cfg` for 'role_requirements_file' or something like that, or to add a field in the template form to allow setting a specific path. Automatically discovering it would definitely be a pain.

wenottingham:
If you do need the naive version of this, it's something like this: 

```
[...]

PR #9945, opened 2021-04-19
Move to fileglob loop for yml requirement locations

AlanCoding:
##### SUMMARY
Overdue followup on https://github.com/ansible/awx/pull/8312

I knew there was more research needed, but didn't know what form it would take until I got my hands dirty.

This entails a _look and feel_ change.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
Screenshot from 2021-04-19 15-46-47

It has a warning... but I don't dislike it.

### DIFF
--- a/awx/playbooks/project_update.yml
+++ b/awx/playbooks/project_update.yml
@@ -161,22 +161,17 @@
   tasks:
 
     - block:
-        - name: detect roles/requirements.(yml/yaml)
-          stat:
-            path: "{{project_path|quote}}/roles/requirements{{ item.ext }}"
-          with_items: "{{ yaml_exts }}"
-          register: doesRequirementsExist
-
         - name: fetch galaxy roles from requirements.(yml/yaml)
           command: >
-            ansible-galaxy role install -r {{ item.stat.path }}
+            ansible-galaxy role install -r {{ item }}
             --roles-path {{projects_root}}/.__awx_cache/{{local_path}}/stage/requirements_roles
             {{ ' -' + 'v' * ansible_verbosity if ansible_verbosity else '' }}
           args:
             chdir: "{{project_path|quote}}"
           register: galaxy_result
-          with_items: "{{ doesRequirementsExist.results }}"
-          when: item.stat.exists
+          with_fileglob:
+            - "{{project_path|quote}}/roles/requirements.yaml"
+            - "{{project_path|quote}}/roles/requirements.yml"
           changed_when: "'was installed successfully' in galaxy_result.stdout"
           environment:
             ANSIBLE_FORCE_COLOR: false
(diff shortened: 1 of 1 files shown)


awx/playbooks/project_update.yml
```diff
@@ -187,22 +182,19 @@
...
           args:
             chdir: "{{project_path|quote}}"
           register: galaxy_collection_result
-          with_items: "{{ doesCollectionRequirementsExist.results }}"
-          when: item.stat.exists
+          with_fileglob:
+            - "{{project_path|quote}}/collections/requirements.yaml"
+            - "{{project_path|quote}}/collections/requirements.yml"
+            - "{{project_path|quote}}/requirements.yaml"
+            - "{{project_path|quote}}/requirements.yml"
```
AlanCoding, line 193: @geerlingguy with https://github.com/ansible/awx/issues/106 in mind, am I getting hotter or colder?

Did Ansible itself roll out a combined `ansible-galaxy install` command that handled both collections and roles? If it did, then using that for the top-level requirements would be preferable.

thedoubl3j (approved):
look good to me @AlanCoding bringing it down to 2 tasks will make future edits easier and create less bloat.

awx/playbooks/project_update.yml
AlanCoding: Was thinking about https://github.com/ansible/ansible/pull/67843 by @jborean93 

and after reading back over that, I believe there's nothing we can really use from it. We still support Ansible 2.9 (which doesn't have the combined command), and we have to specify the roles path, which isn't compatible with the combined roles & collections install command.

So I think this answers the question for me - we have to keep using `ansible-galaxy` `role`/`collection` `install` forever.

Now I'm questioning if we should install roles from `{{project_path|quote}}/requirements.yaml` or not.

fourstepper:
Is this backwards compatible? I am having some issues with a project that has the requirements in the `roles/` directory

AlanCoding:
yeah, it should be completely backward compatible. If you have some details we'll be interested.

fourstepper:
I am getting the following warning when fetching the requirements, could it be related?:

https://paste.sr.ht/~fourstepper/8dd481a03cc7389350409efdef37f5908661cdcf

AlanCoding:
No, that's expected. The aesthetics of it are a drawback of this method. All its saying is that you don't have the `collections/` folder, and it was in the list that was searched for. If you have stuff in `roles/`, that task isn't even relevant to you. There's a different task that is expected to install roles.

fourstepper:
Yeah, my bad. Everything is working fine

PR #9949, opened 2021-04-19
Completes the process of migrating to ling ui v3.

AlexSCorey:
##### SUMMARY
This removes all `withI18n()` and the injected `i18n`.  It also converts all of our strings to be marked for translation via `t'string'` instead of `i18n._(string)`.  
##### ISSUE TYPE
-dependency upgrade

##### COMPONENT NAME
 - UI


jakemcdermott:
@AlexSCorey 
When you have a moment, will you please rebase this pr? (Looks like it's a ways behind current devel)

unlikelyzero:
@AlexSCorey i think this goes without saying, but I would definitely run the tests against this PR

PR #9951, opened 2021-04-19
Add option for ignoring tls on Container Registry credentials

shanemcd:


awx/main/tasks.py
```diff
@@ -1049,6 +1049,29 @@ def build_env(self, instance, private_data_dir, isolated, private_data_files=Non
...
+                with open(registries_conf_path, 'w') as registries_conf:
+                    os.chmod(registries_conf.name, stat.S_IRUSR | stat.S_IWUSR)
+
+                    lines = [
+                        '[[registry]]',
+                        'location = "{}"'.format(host),
+                        'insecure = true',
+                    ]
+
+                    registries_conf.write('\n'.join(lines))
```
jakemcdermott, line 1082: Random remark after poking around in the code: At first, I had a concern that it might be possible for this file to be unexpectedly reused by future job runs after the `verify_ssl` flag gets set back to `true`.

After some testing and thinking, I realized that isn't a problem because the job runs are isolated and these directories aren't reused

Screenshot from 2021-04-19 19-40-53


... Then, after some more thinking, I realized that _even if these directories were reused_, it _still_ wouldn't be a problem because `env['CONTAINERS_REGISTRIES_CONF']` is only set when `verify_ssl` is false :sweat_smile:

awx/main/models/credential/__init__.py
```diff
@@ -1111,6 +1111,12 @@ def create(self):
                 'secret': True,
                 'help_text': ugettext_noop('A password or token used to authenticate with'),
             },
+            {
+                'id': 'verify_ssl',
+                'label': ugettext_noop('Verify SSL'),
+                'type': 'boolean',
+                'default': True,
```
jakemcdermott, line 1118: While testing this, I found a bug w/ the ui form code. The bug isn't directly related to this PR but anyone that tries to change the value of this flag will most likely encounter it. https://github.com/ansible/awx/issues/9955

awx/main/tasks.py
```diff
@@ -859,14 +859,14 @@ def build_execution_environment_params(self, instance, private_data_dir):
...
+                    os.chmod(authfile.name, stat.S_IRUSR | stat.S_IWUSR)
+
                     host = cred.get_input('host')
                     username = cred.get_input('username')
                     password = cred.get_input('password')
                     token = "{}:{}".format(username, password)
                     auth_data = {'auths': {host: {'auth': b64encode(token.encode('ascii')).decode()}}}
                     authfile.write(json.dumps(auth_data, indent=4))
-                authfile.close()
-                os.chmod(authfile.name, stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR)
```
jakemcdermott, line 888: :+1:

jakemcdermott (approved):

awx/main/models/credential/__init__.py
jakemcdermott: https://github.com/ansible/awx/pull/9960

PR #9957, opened 2021-04-20
Isolated removal

jbradberry:
##### SUMMARY
Removal of the isolated nodes feature.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```

```


awx/main/management/commands/register_queue.py
```diff
@@ -95,11 +80,6 @@ def register(self):
                 elif not created:
                     print("Instance Group already registered {}".format(ig.name))
 
-                if self.controller:
-                    (ig_ctrl, changed2) = self.update_instance_group_controller(ig)
```
AlanCoding, line 99: You should get rid of the other uses of `changed2` here. It should work as it is, but we can get rid of the now-useless variable.

jbradberry: This is all very much a work in progress.  Reviews are premature.

awx/main/models/jobs.py
```diff
@@ -759,7 +755,7 @@ def display_artifacts(self):
 
     @property
     def can_run_containerized(self):
-        return any([ig for ig in self.preferred_instance_groups if ig.is_container_group])
+        return True
```
AlanCoding, line 758: Calling out a notable change here. I agree with it, but will make sure to follow the use of the property to double-check.

AlanCoding: Finished looking, I feel pretty confident in this.

jbradberry: Grep shows 5 instances:

```
./awx/main/scheduler/task_manager.py:282:                elif task.can_run_cont>
./awx/main/scheduler/task_manager.py:485:                if task.can_run_contai>
./awx/main/models/jobs.py:757:    def can_run_containerized(self):
./awx/main/models/ad_hoc_commands.py:154:    def can_run_containerized(self):
./awx/main/models/unified_jobs.py:741:    def can_run_containerized(self):
```

3 are the defs, 1 is the one I know I moved this `any()` check to.  The other one (`./awx/main/scheduler/task_manager.py:485`) seems to be doing this same `.is_container_group` logic in a more complex loop.

AlanCoding (approved):

elyezer (approved):

beeankha (approved):

Issue #9910, opened 2021-04-14
Components with aria-label marked to be translated with Plural being displayed as [object Object]


tiagodread:
##### ISSUE TYPE
 - Bug Report

#### Note the same problem may be happening in different places...

While fixing e2e tests I realize that some components is showing `aria-label="[object Object]"` because is using a variable using the <Plural> component.

##### ENVIRONMENT
* Devel

##### STEPS TO REPRODUCE

1. Go to jobs list
2. inspect the button `Cancel Job`

##### EXPECTED RESULTS

That the correct text is displayed for aria-labels not object

##### ACTUAL RESULTS

![image](https://user-images.githubusercontent.com/12464858/114781911-b5065380-9d4f-11eb-92ec-7b94dab932fc.png)

tiagodread:
Fixed

PR #9958, opened 2021-04-20
Fixes object rendered on dom in aria-label

AlexSCorey:
##### SUMMARY
This addresses #9910. Aria-labels that use `<Plural/>` won't work properly.  In this case I used aria-labelledby and passed it the id of the button.  I tested it with a screen reader and that fixed it.  I also fixed some `JobListCancelButton` tooltips

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
Screen Shot 2021-04-20 at 11 59 49 AM


awx/ui_next/src/components/JobList/JobListCancelButton.jsx
```diff
@@ -58,15 +58,29 @@ function JobListCancelButton({ i18n, jobsToCancel, onCancel }) {
                 one="You do not have permission to cancel the following job:"
                 other="You do not have permission to cancel the following jobs:"
               />
+              {cannotCancelPermissions.map((job, i) =>
+                i === cannotCancelPermissions.length - 1 ? (
+                  <strong> {job}</strong>
+                ) : (
+                  <strong> {job},</strong>
+                )
+              )}
```
jakemcdermott, line 67:
```suggestion
              {cannotCancelPermissions.join(',').map(text => <strong>{text}</strong>)}
```

You'll want to use [`join`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join) for this kind of problem. 

Screenshot from 2021-04-20 12-13-15

awx/ui_next/src/components/JobList/JobListCancelButton.jsx
```diff
@@ -58,15 +58,29 @@ function JobListCancelButton({ i18n, jobsToCancel, onCancel }) {
...
+                one="You cannot cancel the following job because it is not running:"
+                other="You cannot cancel the following jobs because they are not running:"
               />
+              {cannotCancelNotRunning.map((job, i) =>
+                i === cannotCancelNotRunning.length - 1 ? (
+                  <strong> {job}</strong>
+                ) : (
+                  <strong> {job},</strong>
+                )
+              )}
```
jakemcdermott, line 83:
```suggestion
{cannotCancelNotRunnning.join(',').map(text => <strong>{text}</strong>)}
```

jakemcdermott (changes requested):

awx/ui_next/src/components/JobList/JobListCancelButton.jsx
jakemcdermott: Actually, nevermind. This won't work for what we're doing here.

jakemcdermott (approved):

unlikelyzero:
@AlexSCorey this may have some pretty serious consequences for the UI Tests. We unfortunately focused too much on aria-labels in the beginning. I personally spent 3 weeks removing all the aria-labels from the visual tests. I would expect the same amount of effort here

AlexSCorey:
@unlikelyzero I think the issue is actually quite unique because we don't have many, and this this might be the only place, where we use lingui's `<Plural/>` to generate an aria label, which was causing the issue

AlexSCorey:
clean cypress test except for 1 failure that doesn't look related [cypress tests](https://dashboard.cypress.io/projects/ui2mq7/runs?branches=%5B%7B%22label%22%3A%229910-PluralAriaLabel%22%2C%22suggested%22%3Afalse%2C%22value%22%3A%229910-PluralAriaLabel%22%7D%5D&committers=%5B%5D&flaky=%5B%5D&page=1&status=%5B%5D&tags=%5B%5D&timeRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D)

tiagodread (approved):
The original issue was fixed, e2e tests are good.

Issue #9955, opened 2021-04-20
Boolean fields on credential inputs sometimes display as true even when they're false


jakemcdermott:
##### ISSUE TYPE
 - Bug Report

##### STEPS TO REPRODUCE
- check out [this pr](https://github.com/ansible/awx/pull/9951)
- make a container registry credential
- after making the cred, edit it and change the tls flag to false and save again
- Note that when you return to the form it'll still display as a checked box but the api value is actually false
- Observation: if you log the value of credential.inputs [in the form component](https://github.com/ansible/awx/blob/devel/awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx#L240), it changes to true somehow during initialization.

mabashian:
You can reproduce this bug on devel.  Just create a Tower credential and leave the Verify SSL option unchecked.  After saving, attempt to edit the credential and you'll see Verify SSL checked.

tiagodread:
The bug was fixed by #9960, now the UI is properly displaying the current state of the API data for options. We had some e2e broken tests, however, none of them is related to this change.

PR #9960, opened 2021-04-20
Fixes bug where credential form checkboxes were erroneously checked

mabashian:
##### SUMMARY
link #9955 

A couple of things going on here.

1) Updating `initialValues.inputs` was also modifying `credential.inputs` because `initialValues.inputs` was originally a reference to `credential.inputs`.  I changed this so that `initialValues.inputs` now starts off as a clone of `credential.inputs`.
2) https://github.com/ansible/awx/compare/devel...mabashian:9955-cred-bool?expand=1#diff-db8df3eaf3e3b3117f845786dea77451fba53e9d6a3f49ae367a26137039fc35L256 <- this line was erroneously evaluating to `false` when the value of an input was `false`.  We actually just want to make sure the key exists in the object before dropping in to this block.  This is what actually fixes the bug.  Before this change we would fall in to https://github.com/ansible/awx/compare/devel...mabashian:9955-cred-bool?expand=1#diff-db8df3eaf3e3b3117f845786dea77451fba53e9d6a3f49ae367a26137039fc35R268 where the checkbox value would be erroneously set.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx
+++ b/awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx
@@ -244,7 +244,7 @@ function CredentialForm({
     description: credential.description || '',
     organization: credential?.summary_fields?.organization || null,
     credential_type: credentialTypes[initialTypeId]?.name || '',
-    inputs: credential?.inputs || {},
+    inputs: { ...credential?.inputs },
     passwordPrompts: {},
     isOrgLookupDisabled: isOrgLookupDisabled || false,
   };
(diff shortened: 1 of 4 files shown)


awx/ui_next/src/screens/Credential/shared/CredentialForm.jsx
```diff
@@ -244,7 +244,7 @@ function CredentialForm({
     description: credential.description || '',
     organization: credential?.summary_fields?.organization || null,
     credential_type: credentialTypes[initialTypeId]?.name || '',
-    inputs: credential?.inputs || {},
+    inputs: { ...credential?.inputs },
```
jakemcdermott, line 247: :zap:, there it is! nice find

jakemcdermott (approved):

tiagodread (approved):
The bug was fixed, now the UI is properly displaying the current state of the API data for options. We had some e2e broken tests, however, none of them is related to this change.

Issue #9852, opened 2021-04-08
error when removing single Source from inventory


balonik:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
It is not possible to delete single Source from Inventory. When trying to delete single inventory source following error occurs

```
Error: TypeError: Cannot read property 'count' of undefined at https://awx.example.com/static/js/main.2594bb50.chunk.js:1:148664 at c (https://awx.example.com/static/js/2.89b6cada.chunk.js:2:1469770) at Generator._invoke (https://awx.example.com/static/js/2.89b6cada.chunk.js:2:1469523) at Generator.next (https://awx.example.com/static/js/2.89b6cada.chunk.js:2:1470129) at r (https://awx.example.com/static/js/2.89b6cada.chunk.js:2:729) at s (https://awx.example.com/static/js/2.89b6cada.chunk.js:2:932)
```

##### ENVIRONMENT
* AWX version: 19.0.0
* AWX install method: awx-operator 0.8.0 on RKE
* Ansible version:  X.Y.Z
* Operating System:
* Web Browser:

##### STEPS TO REPRODUCE

Create Inventory
Create Inventory source
Delete Inventory source

##### EXPECTED RESULTS

Source is deleted.

##### ACTUAL RESULTS

Source is not deleted. See error above.

##### ADDITIONAL INFORMATION

Currently as a workaround it is possible to create a dummy Source and select it as well. Multiple Source deletion is working. The warning message is broken, tho.

```
{numItemsToDelete, plural, one {This inventory source is currently being used by other resources that rely on it. Are you sure you want to delete it?} other {Deleting these inventory sources could impact other resources that rely on them. Are you sure you want to delete anyway?}}
```

mabashian:
Slight tweak to the reproduction steps:

Create Inventory
Create _two_ Inventory sources
Delete _one of the_ Inventory sources

If you only have one inventory source on the inventory then you can delete it without issue.

tiagodread:
PR merged, added a e2e test case for it

PR #9975, opened 2021-04-21
Fixes bug where users could not delete a single inventory source

mabashian:
##### SUMMARY
link #9852 

Also fixes a bug that I came across with deletion warnings.  The deletion warning was showing a count for workflow nodes that referenced any inventory source with the same parent inventory.  For example:

Create an inventory
Create two inventory sources invsrc1 and invsrc2
Create a workflow with a node that syncs invsrc1
Attempt to delete invsrc2

The warning will indicate that there's 1 workflow node that uses the inventory source but that's actually not true.  There should be no deletion warning in this case.

This PR addresses ^^

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI


jakemcdermott (approved):

tiagodread:
Ok, manually tested and both problems were fixed:
- Now if the user try to delete one or more inv sources they can.
- The deletition warning is only being displayed for inv sources used by any workflow.

I'll Add a e2e test case for it

tiagodread (approved):

PR #9980, opened 2021-04-21
Rename awx_collection modules

fosterseth:
##### SUMMARY

- drop "tower_" prefix from modules
  - tower_meta will become controller_meta
- provide redirects so that old playbooks can still run
- change ansible.tower to ansible.controller for downstream collection
- dropping deprecated modules tower_send, tower_receive, tower_workflow_template

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.0.0
```

##### ADDITIONAL INFO
Tower references remaining

- Ansible Tower credential type
- URLs like https://www.ansible.com/tower
- tower-cli package, tower_cli.cfg
- tower_inventory.yml
- environment variables like TOWER_HOST, TOWER_USERNAME
- module options like tower_host, tower_username
- tower_token result output from the token module


awx_collection/README.md
```diff
@@ -32,7 +32,7 @@ Installing the `tar.gz` involves no special instructions.
 ## Running
 
 Non-deprecated modules in this collection have no Python requirements, but
-may require the official [AWX CLI](https://docs.ansible.com/ansible-tower/latest/html/towercli/index.html)
+may require the official [AWX CLI](https://docs.ansible.com/ansible-controller/latest/html/controllercli/index.html)
```
fosterseth: @tvo318  do we plan to change urls? if not I need to clean up all urls

awx_collection/plugins/modules/controller_user.py
```diff
@@ -71,43 +71,43 @@
...
-  tower_user:
+- name: Add controller user as a system administrator
+  controller_user:
     username: jdoe
     password: foobarbaz
     email: jdoe@example.org
     superuser: yes
     state: present
-    tower_config_file: "~/tower_cli.cfg"
+    controller_config_file: "~/tower_cli.cfg"
```
fosterseth: should we change to `controller_cli.cfg`?

fosterseth:
locally, I'm experiencing the following unit test failures
```
awx_collection/test/awx/test_completeness.py::test_completeness ERROR
awx_collection/test/awx/test_credential_type.py::test_changed_false_with_api_changes FAILED
```

and integration test failures,

```
controller_instance_group
    fatal: [testhost]: FAILED! => {"changed": false, "msg": "Unable to create instance_group AWX-Collection-tests-controller_instance_group-group2-wMBnrxYvxQGHRxhz: {'is_container_group': ['is_container_group must be True when associating a credential to an Instance Group']}"}
controller_credential
    fatal: [testhost]: FAILED! => {"changed": false, "msg": "Request to /api/v2/credential_types/?name=Ansible+Controller returned 0 items, expected 1", "query": {"name": "Ansible Controller"}, "response": {"json": {"count": 0, "next": null, "previous": null, "results": []}, "status_code": 200}, "total_results": 0}
controller_ad_hoc_command_cancel
    fatal: [testhost]: FAILED! => {"changed": false, "msg": "Ad Hoc Command is not running"}
controller_job_template
    fatal: [testhost]: FAILED! => {"changed": false, "msg": "Project update failed"}
controller_job_cancel
    fatal: [testhost]: FAILED! => {"changed": false, "msg": "Failed to launch job, see response for details", "response": {"json": {"playbook": ["Missing a revision to run due to failed project update."]}, "status_code": 400}}
```

not sure if these are related to my changes

awx_collection/README.md
```diff
@@ -16,7 +16,7 @@ This Ansible collection allows for easy interaction with an AWX server via Ansib
 
 This source for this collection lives in the `awx_collection` folder inside of the
 AWX source.
-The previous home for this collection was inside the folder [lib/ansible/modules/web_infrastructure/ansible_tower](https://github.com/ansible/ansible/tree/stable-2.9/lib/ansible/modules/web_infrastructure/ansible_tower) in the Ansible repo,
+The previous home for this collection was inside the folder [lib/ansible/modules/web_infrastructure/ansible_controller](https://github.com/ansible/ansible/tree/stable-2.9/lib/ansible/modules/web_infrastructure/ansible_controller) in the Ansible repo,
```
beeankha: Should all instances of "Ansible Controller" actually be "Automation Controller"? I might not know exactly what was agreed upon for this particular renaming, however, in other places I've seen `Ansible Tower` `--->` `Automation Controller`.

AlanCoding:
I don't see a redirect for the inventory plugin or lookup plugins. You renamed the plugins themselves, so I'd expect this would break all playbooks that use them. I don't know for sure if fixing these is possible.

Also please keep in mind that we have hard-coded redirects in Ansible core for the old names.

https://github.com/ansible/ansible/blob/ca448f7c350fae8080dcfec648342d2fc8837da0/lib/ansible/config/ansible_builtin_runtime.yml#L6561

I would definitely want to be sure that still works with this change.

awx_collection/meta/runtime.yml
```diff
@@ -1,46 +1,123 @@
...
-    - tower_role
-    - tower_schedule
-    - tower_settings
-    - tower_team
-    - tower_token
-    - tower_user
-    - tower_workflow_job_template_node
-    - tower_workflow_job_template
-    - tower_workflow_launch
+  controller:
```
AlanCoding, line 3: The use case for this looks like:

```yaml
---
- hosts: localhost
  gather_facts: false
  connection: local
  collections:
    - awx.awx
  module_defaults:
    awx.awx.controller:
      tower_host: https://foo.invalid
      organization: Default
  tasks:
    - name: ...
```

If you're going to be vague but count on FQCN use, you might as well name it "default" or something. If you're trying to use a name such that people recognize the product, I think "controller" is _too_ vague and borderline self-contradictory sitting next to `awx.awx`. Maybe "ansible_controller". Even "tower" was always too vague of an abbreviation when put into its proper use context IMO.

awx_collection/plugins/doc_fragments/auth.py
```diff
@@ -10,52 +10,52 @@
...
-    - Whether to allow insecure connections to Tower or AWX.
+    - Whether to allow insecure connections to Controller or AWX.
     - If C(no), SSL certificates will not be validated.
     - This should only be used on personally controlled sites using self-signed certificates.
-    - If value not set, will try environment variable C(TOWER_VERIFY_SSL) and then config files
+    - If value not set, will try environment variable C(CONTROLLER_VERIFY_SSL) and then config files
     type: bool
-    aliases: [ tower_verify_ssl ]
-  tower_config_file:
+    aliases: [ controller_verify_ssl ]
```
AlanCoding: If you are doing this, then `tower_verify_ssl` should be an alias alongside `controller_verify_ssl`.

Repeat for every other option.

Even options that don't currently have aliases should get a new alias to the old name.

awx_collection/plugins/modules/controller_user.py
AlanCoding: the modules in this collection predate AWX itself. When they were created, Ansible Tower existed and nothing else.

Several things here resisted the name change from Tower->AWX, due to risks of breaking people's content. The config file is possibly the most painful example of that. It uses the location and formats from the `tower-cli` project, which was replaced with the `awx-cli`. The CLI itself has no notion of a config file itself, but the modules (which became the collection) still respects the naming and format from 2014

john-westcott-iv:
If we are doing a rename anyway, are there going to be other collections for the platform? I.e. should our collection namespace become something like ansible.platform.controller?

john-westcott-iv:
Also, can we take this opportunity to actually remove the deprecated modules form the collection?

awx_collection/plugins/doc_fragments/auth.py
```diff
@@ -10,52 +10,52 @@
...
 
-    # Ansible Tower documentation fragment
+    # Ansible Controller documentation fragment
     DOCUMENTATION = r'''
 options:
-  tower_host:
+  controller_host:
     description:
-    - URL to your Tower or AWX instance.
-    - If value not set, will try environment variable C(TOWER_HOST) and then config files
```
john-westcott-iv, line 19: These variables matched what was coming from Tower for the "Ansible Tower" credential type. Are those also changing? Technically, if they are aliased it won't matter.

awx_collection/meta/runtime.yml
fosterseth: I believe the idea is to rename the collections to `ansible.controller`, I need to update the code to reflect that

awx_collection/README.md
wenottingham: We should (or move the docs to access.redhat.c0m)

wenottingham: Though in this case it would be the automation-controller, under the Automation Platform. It is *not* called the "ansible controller"

awx_collection/plugins/modules/controller_user.py
wenottingham: see /etc/tower/SECRET_KEY. This is probably not worth changing.

awx_collection/README.md
wenottingham: Correct, it's automation controller.

awx_collection/meta/runtime.yml
AlanCoding: An `ansible.*` name is for the downstream. This `awx.awx` name is for the upstream.

We can't have split names for things _inside_ the collection, and we really shouldn't have had 2 different names for the upstream / downstream collection in the first place. The framework for collections works very bad. We template code in some places.

If you use the recent ansible-core and run `ansible-test` without our ignore list, you will find it littered with errors because all of the warnings and deprecations are version-aware, but we are supporting two different versioning systems. This is broken, and it was never fixed. We have skated by because we largely failed to test with the recent Ansible versions.

While a name might sound superficial, it's actually a monster of a problem to go from the intended naming scheme to something that works and actually confirms to present standards for collections.

AlanCoding:
> Also, can we take this opportunity to actually remove the deprecated modules form the collection?

Maybe we could do it in a separate PR. It's already a lot of work to finish up what is being attempted here.

awx_collection/plugins/doc_fragments/auth.py
AlanCoding: This comment may have been on an outdated diff.

It looks like the environment variables are remaining the same for now. I will raise this point in an issue. Thanks for bringing it up.

We _have_ in the past aliased an environment variable for the config.

https://github.com/ansible/awx/blob/eec4f8dcc27428cd1aeb1c251d43052a3ea780e8/awxkit/awxkit/cli/format.py#L19

but that might have just been for the CLI.

awx_collection/plugins/inventory/controller.py
```diff
@@ -104,12 +104,12 @@ def parse(self, inventory, loader, path, cache=True):
```
AlanCoding, line 115: On line 93 of this module

```python
return path.endswith(('tower_inventory.yml', 'tower_inventory.yaml', 'tower.yml', 'tower.yaml'))
```

You should add some options to this for the change to be coherent with the new naming scheme. What do you intend inventory files to be named (forward-looking)? Should it be `platform_controller.yml`? `controller.yml`?

Don't remove options like `tower.yml` for compatibility's sake. The `verify_file` method is often very very annoying for the user, and being maximally permissive is the best policy.

fosterseth:
```python3
def verify_file(self, path):
    if path.endswith('@controller_inventory') or path.endswith('@tower_inventory'):
        self.no_config_file_supplied = True
        return True
    elif super(InventoryModule, self).verify_file(path):
        return path.endswith(('controller_inventory.yml', 'controller_inventory.yaml', 'controller.yml', 'controller.yaml',
                              'tower_inventory.yml', 'tower_inventory.yaml', 'tower.yml', 'tower.yaml'))
    else:
        return False
```

I'm thinking that

AlanCoding: yeah that looks good.

awx_collection/plugins/doc_fragments/auth.py
```diff
@@ -10,52 +10,57 @@
 
 class ModuleDocFragment(object):
 
-    # Ansible Tower documentation fragment
+    # Automation Platform Controller documentation fragment
     DOCUMENTATION = r'''
 options:
-  tower_host:
+  controller_host:
```
AlanCoding, line 16: I still have hesitations about the use of bare "controller" in variable names, because it's a very overloaded term in the Ansible world. An Ansible controller is a generic concept, and we haven't been vetting this with the wider community, who may balk at the naming. The alternatives which I can see a strong justification for are:

1. `awx_host`
2. `platform_controller_host`

If you want a reason why this is named `tower_host`, I'll tell you it is because these modules were created before AWX existed. That's the only reason why the AWX naming was not used. In the case of the CLI, created after AWX existed, the entrypoint remains `awx` for the downstream, product, version. If you think about this in the Ansible Platform context, this holds up well against env vars like `ANSIBLE_GALAXY_SERVER`.

The option (2) is the same idea but specific enough.

awx_collection/tools/vars/associations.yml
```diff
@@ -4,10 +4,13 @@ associations:
...
+      required: False
   groups:
     - related_item: hosts
       endpoint: hosts
       description: "The hosts associated with this group"
+      required: False
     - related_item: groups
       endpoint: children
       description: "The hosts associated with this group"
+      required: False
```
beeankha: All instances of `False` in the above need to be lower case (`false`)

awx_collection/plugins/doc_fragments/auth.py
rebeccahhh: I am a huge fan of `awx_host`, it fits our existing infrastructure well and is shorter, which to me is a win.

shanemcd (approved):

beeankha (approved):

PR #9981, opened 2021-04-22
[devel] calculate 4 week limit based on since, when since is provided

jladdjr:
If `since` is provided, we should be calculating the horizon starting from that point and moving forward
.. instead of starting from the end and working backwards.

### DIFF
--- a/awx/main/analytics/core.py
+++ b/awx/main/analytics/core.py
@@ -154,9 +154,9 @@ def gather(dest=None, module=None, subset=None, since=None, until=None, collecti
         until = None if until is None else min(until, _now)
         since = None if since is None else min(since, _now)
 
-        if since and not until:
-            # If `since` is explicit but not `until`, `since` should be used to calculate the 4-week limit
-            until = min(since + timedelta(weeks=4), _now)
+        if since:
+            # If `since` is explicit, the four-week limit should be calculated from `since`
+            until = min(since + timedelta(weeks=4), _now if until is None else until)
         else:
             until = _now if until is None else until
 
--- a/awx/main/analytics/core.py
+++ b/awx/main/analytics/core.py
@@ -156,6 +156,8 @@ def gather(dest=None, module=None, subset=None, since=None, until=None, collecti
 
         if since:
             # If `since` is explicit, the four-week limit should be calculated from `since`
+            if until and until > since + timedelta(weeks=4):
+                logger.warning(f'collection interval exceeds four weeks, reducing interval to {since} - {since + timedelta(weeks=4)}')
             until = min(since + timedelta(weeks=4), _now if until is None else until)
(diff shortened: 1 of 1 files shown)


chrismeyersfsu:
As I understand is, the PR here implements the below logic. The only case I question is the last one. I'm fine if periodic analytics gather always limits the time window to 4 weeks. But I question limiting an explicit command-line invocation where `--until` is specified but, under the hood, we limit the gather to 4 weeks.

I'd be fine if somehow told the user "hey we are limiting you".

```
User gives us --since

--------------------------------------------------------------------------------------
                 ^ since

We set until = since + 4 weeks

                   4 weeks
                 |------------------|
-------------------------------------------------------------------------------------
                 ^ since            ^ until
```

```
User gives us --since and since is close to now

                                                                                 4 weeks
                                                                           |------------------|
--------------------------------------------------------------------------------------
                                                                           ^ since

We set until = now

                                                                                 4 weeks
                                                                           |------------------|
--------------------------------------------------------------------------------------
                                                                           ^ since    ^until

```

```
User gives us --since and --until

                                                                4 weeks
                                                           |------------------|
--------------------------------------------------------------------------------------
                                                           ^ since  ^ until

We take the values as they are
```

```
User gives us --since and --until

                                                               4 weeks
                                                           |------------------|
--------------------------------------------------------------------------------------
                                                           ^ since                        ^ until

We set until = since + 4 weeks

                                                               4 weeks
                                                           |------------------|
--------------------------------------------------------------------------------------
                                                           ^ since           ^ until
```

<hr>

jbradberry:
@chrismeyersfsu it's a good point.  The analytics gathering feature has always tried to make sure to limit the amount of data given to 4 weeks, but it would be good to add some logging statements notifying that it is doing so.

jladdjr:
image

when running `awx-manage gather_analytics` logger warnings are surfaced to stdout by default, so this logging statement should cover both analytics collection via the task manager and cli

awx/main/analytics/core.py
```diff
@@ -154,9 +154,11 @@ def gather(dest=None, module=None, subset=None, since=None, until=None, collecti
         until = None if until is None else min(until, _now)
         since = None if since is None else min(since, _now)
 
-        if since and not until:
-            # If `since` is explicit but not `until`, `since` should be used to calculate the 4-week limit
-            until = min(since + timedelta(weeks=4), _now)
+        if since:
+            # If `since` is explicit, the four-week limit should be calculated from `since`
+            if until and until > since + timedelta(weeks=4):
```
chrismeyersfsu, line 159: We want this message to trigger even when `until` is not specified. i.e. if a user specifies `--since <10 years ago>` we will move the date to `since + 4weeks` and we should let them know.

Issue #9163, opened 2021-01-25
Admins cannot access Workflow Approvals page


djj106:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

After upgrading from 15.0.1 to 17.0.0, admin users cannot access the Workflow Approvals page. Non-admin users can access the page as expected. Relevant error from the browser:

`TypeError: Cannot read property 'username' of undefined
    at dI (WorkflowApprovalStatus.jsx:50)
    at Xo (react-dom.production.min.js:153)
    at vs (react-dom.production.min.js:261)
    at cl (react-dom.production.min.js:246)
    at sl (react-dom.production.min.js:246)
    at Qs (react-dom.production.min.js:239)
    at react-dom.production.min.js:123
    at t.unstable_runWithPriority (scheduler.production.min.js:19)
    at Hi (react-dom.production.min.js:122)
    at Gi (react-dom.production.min.js:123)`

##### ENVIRONMENT
* AWX version: 17.0.0
* AWX install method: docker on linux
* Ansible version:  X.Y.Z
* Operating System:
* Web Browser:

djj106:
I believe i've narrowed down the issue. It would appear that the UI is attempting to display a username that has been deleted, but the workflow approval record still exists tied to the deleted user. Since admins see all approval records, they would run into the issue where a non-admin user may not.

thibveni:
same issue here

tiagodread:
@one-t I think this can be closed with the merge of #9988 ?

PR #9988, opened 2021-04-22
Fixes bug where workflow approvals page would crash if deleted username was referenced

mabashian:
##### SUMMARY
link #9163 

The tooltip on the status labels for completed workflow approvals attempts to show the username of the user that acted on it.  If that user has been deleted then the key is not present in the object.  This would cause the page to crash.

To fix this, I conditionally show a different string when the username is not available.  If a user wants to see which deleted user acted on the approval they'll have to go to the activity stream.

Screen Shot 2021-04-22 at 9 59 23 AM

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/screens/WorkflowApproval/shared/WorkflowApprovalStatus.jsx
+++ b/awx/ui_next/src/screens/WorkflowApproval/shared/WorkflowApprovalStatus.jsx
@@ -1,5 +1,4 @@
 import React from 'react';
-
 import { t } from '@lingui/macro';
 import { Label, Tooltip } from '@patternfly/react-core';
 import { CheckIcon, InfoCircleIcon } from '@patternfly/react-icons';
@@ -24,9 +23,15 @@ function WorkflowApprovalStatus({ workflowApproval }) {
   if (workflowApproval.status === 'failed' && workflowApproval.failed) {
     return (
       <Tooltip
-        content={t`Denied by ${
-          workflowApproval.summary_fields.approved_or_denied_by.username
-        } - ${formatDateString(workflowApproval.finished)}`}
+        content={
+          workflowApproval.summary_fields?.approved_or_denied_by?.username
+            ? t`Denied by ${
+                workflowApproval.summary_fields.approved_or_denied_by.username
+              } - ${formatDateString(workflowApproval.finished)}`
+            : t`Denied - ${formatDateString(
+                workflowApproval.finished
+              )}.  See the Activity Stream for more information.`
+        }
         position="top"
       >
         <Label variant="outline" color="red" icon={<InfoCircleIcon />}>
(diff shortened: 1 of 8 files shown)


awx/ui_next/src/screens/WorkflowApproval/shared/WorkflowApprovalStatus.jsx
```diff
@@ -28,11 +28,19 @@ function WorkflowApprovalStatus({ workflowApproval, i18n }) {
...
     return (
       <Tooltip
-        content={i18n._(
-          t`Denied by ${
-            workflowApproval.summary_fields.approved_or_denied_by.username
-          } - ${formatDateString(workflowApproval.finished)}`
-        )}
+        content={
+          workflowApproval.summary_fields?.approved_or_denied_by?.username
+            ? i18n._(
```
AlexSCorey: Remove the `i18n._` here in favor of  ```t`string` ```

If you can get rid of the `i18n` object and `withI18n` in this file as well that would be good to do.

awx/ui_next/src/screens/WorkflowApproval/shared/WorkflowApprovalStatus.jsx
```diff
@@ -28,11 +28,19 @@ function WorkflowApprovalStatus({ workflowApproval, i18n }) {
...
+                t`Denied by ${
+                  workflowApproval.summary_fields.approved_or_denied_by.username
+                } - ${formatDateString(workflowApproval.finished)}`
+              )
+            : i18n._(
```
AlexSCorey: same

awx/ui_next/src/screens/WorkflowApproval/shared/WorkflowApprovalStatus.jsx
```diff
@@ -45,11 +53,19 @@ function WorkflowApprovalStatus({ workflowApproval, i18n }) {
...
     return (
       <Tooltip
-        content={i18n._(
-          t`Approved by ${
-            workflowApproval.summary_fields.approved_or_denied_by.username
-          } - ${formatDateString(workflowApproval.finished)}`
-        )}
+        content={
+          workflowApproval.summary_fields?.approved_or_denied_by?.username
+            ? i18n._(
```
AlexSCorey: same

awx/ui_next/src/screens/WorkflowApproval/shared/WorkflowApprovalStatus.jsx
```diff
@@ -45,11 +53,19 @@ function WorkflowApprovalStatus({ workflowApproval, i18n }) {
...
+                t`Approved by ${
+                  workflowApproval.summary_fields.approved_or_denied_by.username
+                } - ${formatDateString(workflowApproval.finished)}`
+              )
+            : i18n._(
```
AlexSCorey: same

AlexSCorey (changes requested):
Looks good.  Small changes.

mabashian:
@AlexSCorey I think I got all those cleaned up for ya

tiagodread:
@mabashian could you rebase this one?

tiagodread:
@mabashian could you rebase this one?

one-t:
Just mentioning here, this PR may resolve some issues we're seeing with YOLO runs that include API tests. These seem to leave behind orphaned objects that cause the page to crash.

I'll see if I can figure out the exact reproducer(s). We want the UI to handle this but we also don't want orphaned objects floating around in the DB.

one-t:
Ok, I've confirmed, the exact situation is this:
1) Create a workflow job template with an approval node
2) Create a user with the rights to approve
3) Launch workflow job and approve as the user
4) Delete the user
5) Page fails to render as described in the related issue.

I don't think any changes will be needed for the API tests, we already don't clean up jobs.

one-t:
Confirmed, this does fix the issue and the page now renders when the approving user is deleted.

one-t (approved):
Jenkins run demonstrated that this fixes our issue with the API tests leaving behind orphaned approvals. :shipit:

one-t:
!regate

AlexSCorey (approved):
Looks good well done!

PR #9991, opened 2021-04-22
Upgrade Patternfly packages

nixocio:
* Upgrade @patternfly/patternfly and @patternfly/react-core to the
latest available version.

### DIFF
--- a/awx/ui_next/package-lock.json
+++ b/awx/ui_next/package-lock.json
@@ -2265,18 +2265,18 @@
       "dev": true
     },
     "@patternfly/patternfly": {
-      "version": "4.96.2",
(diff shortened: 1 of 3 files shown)


AlexSCorey (approved):
LGTM pending e2e tests.

jakemcdermott (changes requested):
The diffs on the lockfile shouldn't blow up for an update like this. When I update these packages using the steps in our readme, I get [this](https://github.com/jakemcdermott/awx/commit/68bd7179560c4748f6426090e97a17ef1c39d4aa)

We'll probably want to standardize on doing these updates in the development container. It's a topic I'd like to visit in the near future. In the meantime, try with the exact steps in our readme and if that doesn't work for you, use [this commit](https://github.com/jakemcdermott/awx/commit/68bd7179560c4748f6426090e97a17ef1c39d4aa) please.

Issue #9997, opened 2021-04-22
Visual differences on the same component between dev and production builds


tiagodread:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Looking at the visual tests diffs we realized that there is a visual diff between the same component on dev and prod builds
This may or may not be introduced by https://github.com/ansible/awx/pull/9649

##### ENVIRONMENT
* Devel

##### STEPS TO REPRODUCE

Run the dev server and production build and compare the component above

Before:
![image](https://user-images.githubusercontent.com/12464858/115785371-bbbb4900-a395-11eb-95e8-ef8e4f944c20.png)

After:
![image (1)](https://user-images.githubusercontent.com/12464858/115785367-bb22b280-a395-11eb-89a3-ab0308fd49d2.png)

Note the missing border of the select component

nixocio:
Related to patternfly/patternfly-react#5882

Also see: patternfly/patternfly-react#5650

tiagodread:
fixed!

PR #9998, opened 2021-04-22
Update patternfly packages

nixocio:
Update patternfly packages


jakemcdermott (approved):

nixocio:
I triggered another e2e test run, and visual. Evaluating failures to see if they are related to see this change.

tiagodread (approved):
I went through the e2e results including visual tests, there are some visual differences mostly related to this problem https://github.com/ansible/awx/issues/9997 because we render the Select component in almost every single page across the app. I didn't see any big failure/differences so far.

Issue #8686, opened 2020-11-25
AWX Project module should allow forcing update


kedark3:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

We need to have ability to force project update in our tower using where we are using infrastructure as code. I am aware of the fact that we have `project_update` but Sean and I talked and he has initial code https://github.com/ansible/awx/compare/devel...sean-m-sullivan:project_update here.

Opening this issue to add visibility. 

Adding this enhancement will make https://github.com/redhat-cop/tower_configuration/tree/devel/roles/tower_role more useful for us(my team) as regular Tower Infra-as-Code (IaC) users.

Thanks @sean-m-sullivan

sean-m-sullivan:
@wenottingham This feature has been added and merged, this can now be closed.

yagomarques:
I noticed the project update is only triggering if we change something else

We use it to test.
```
    - name: Create a Project
      tower_project:
        name: "New Project - no update"
        description: "foobar"
        organization: "New Org"
        timeout: 10
        update_project: no
        scm_type: git
        scm_url: https://github.com/one-t/verification_repo.git
```

sean-m-sullivan:
What do you mean by change something? in the previous example the update project is no, so it shouldn't trigger, and the timeout would be ignored.

I was looking at this with a 19.0.0 awx. and with this code
```
    - name: Create a Project
      awx.awx.tower_project:
        name: "New Project - no update"
        description: "foobar"
        organization: Default
        timeout: 10
        scm_type: git
        scm_url: https://github.com/one-t/verification_repo.git
        update_project: yes
```
The key here is update_project: yes , if you put no/false, it will behave like it did in the past, and not update the project, and its working as expected. If this is not the case can you go into more detail?
@yagomarques

one-t:
Hi, Sean-

What we noticed in testing this is that the project update only occurs if the task reports a Changed status. 

Steps to reproduce:

1. Create a project using the `tower_project` task with `update_project` enabled
2. Push a new commit to the repository
3. Run the same task again

Expected result: Project update is triggered

Actual result: Project update is not triggered

Then,
1. Change something about the task- for example, change the `timeout` value to `30`
2. Run the task again

Result- timeout value is changed and the project update is triggered.

sean-m-sullivan:
@one-t @yagomarques 
Found the issues, update is #10000

yagomarques:
[...]

PR #10000, opened 2021-04-23
Fix project update

sean-m-sullivan:
##### SUMMARY
Fixing bug found in #8686 
Found that the on change was not triggering due to no actual changes in the Post, Set so when project update is set to true, that the wait for update is Always triggered. Also added logic from project_update to set changed status depending on if refspec changes.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
- awx_collection
##### AWX VERSION
19.0.0

### DIFF
--- a/awx_collection/plugins/modules/tower_project.py
+++ b/awx_collection/plugins/modules/tower_project.py
@@ -204,6 +204,7 @@ def wait_for_project_update(module, last_request):
     wait = module.params.get('wait')
     timeout = module.params.get('timeout')
     interval = module.params.get('interval')
+    scm_revision_original = last_request['scm_revision']
 
     if 'current_update' in last_request['summary_fields']:
         running = True
@@ -229,9 +230,14 @@ def wait_for_project_update(module, last_request):
         start = time.time()
 
         # Invoke wait function
-        module.wait_on_url(
+        result_final = module.wait_on_url(
             url=result['json']['url'], object_name=module.get_item_name(last_request), object_type='Project Update', timeout=timeout, interval=interval
         )
+
+        # Set Changed to correct value depending on if hash changed Also output refspec comparision
+        module.json_output['changed'] = True
+        if result_final['json']['scm_revision'] == scm_revision_original:
+            module.json_output['changed'] = False
 
     module.exit_json(**module.json_output)
 
@@ -380,9 +386,20 @@ def main():
         on_change = wait_for_project_update
 
     # If the state was present and we can let the module build or update the existing project, this will return on its own
-    module.create_or_update_if_needed(
-        project, project_fields, endpoint='projects', item_type='project', associations=association_fields, on_create=on_change, on_update=on_change
+    response = module.create_or_update_if_needed(
+        project,
+        project_fields,
+        endpoint='projects',
+        item_type='project',
+        associations=association_fields,
+        on_create=on_change,
+        on_update=on_change,
+        auto_exit=not update_project,
     )
+
+    if update_project:
+        wait_for_project_update(module, response)
+    module.exit_json(**module.json_output)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


awx_collection/tests/integration/targets/tower_project/tasks/main.yml
```diff
@@ -9,245 +9,245 @@
...
+        scm_type: git
+        scm_url: https://github.com/ansible/test-playbooks
+        wait: true
+      register: result
+
+    - assert:
+        that:
+          - result is changed
+
+    - name: Recreate the project to validate not changed, currenty reports inaccurate warning
```
beeankha: This task can be renamed to just `Recreate the project to validate not changed`

awx_collection/tests/integration/targets/tower_project/tasks/main.yml
```diff
@@ -9,245 +9,245 @@
...
+        name: "{{ project_name1 }}"
+        organization: Default
+        scm_type: git
+        scm_url: https://github.com/ansible/test-playbooks
+        wait: false
+      register: result
+      ignore_errors: true
+
+    - debug:
+        var: result
```
beeankha: Debug lines should be removed here

awx_collection/tests/integration/targets/tower_project/tasks/main.yml
```diff
@@ -9,245 +9,245 @@
...
+    - assert:
+        that:
+          - result is changed
+          - "'The field scm_track_submodules of project' in result.warnings[0]"
+          - "'inaccurately report task is changed.' in result.warnings[0]"
```
beeankha: Result is `not changed` and the below assertions no longer apply :heavy_check_mark:

awx_collection/plugins/modules/tower_project.py
```diff
@@ -229,10 +230,15 @@ def wait_for_project_update(module, last_request):
...
 
         # Invoke wait function
-        module.wait_on_url(
+        result_final = module.wait_on_url(
             url=result['json']['url'], object_name=module.get_item_name(last_request), object_type='Project Update', timeout=timeout, interval=interval
         )
 
+        # Set Changed to correct value depending on if hash changed Also output refspec comparision
+        module.json_output['changed'] = True
+        if result_final['json']['scm_revision'] == scm_revision_original:
```
john-westcott-iv, line 239: Food for thought..... I think on a technical level it can be more complicated that this. A project update can include a requirements.yml. if a new version of a collection/role was added the project hash itself may not change but that new version of the collection/role could have been downloaded changing Tower.

sean-m-sullivan: Except in this case the proejct is https://github.com/ansible/test-playbooks without a collections/roles requirements files.

sean-m-sullivan: At least considering The test, but In this case its not always going to work, but its was a requested feature to have it not changed when the scm does not change. I hadn't thought of the collections going in.

sean-m-sullivan: Ok did test, Synced to 
https://github.com/sean-m-sullivan/verification_repo
and it synced with scm_revision 48214f29aa737c8ce093c339d6f4db0cbd74a59b
which matches   repo scm hash 48214f29aa737c8ce093c339d6f4db0cbd74a59b
and upon collection update got  48214f29aa737c8ce093c339d6f4db0cbd74a59b 
even though the tower_utilities version moved from 1.22 to 1.23

AlanCoding: Yes, but the API is vague in these cases. It is intended that a project update can pull in new versions of stuff from Galaxy (otherwise you would have to add new commits to rev it, yuck).

I've believed that the inevitable end-game is to track a checksum for the collections & roles folders specific to that project, and show that in the API.

In the absence of the API solving the problem, I'm not going to worry about the collection having false-negatives.

AlanCoding (approved):

beeankha (approved):

awx_collection/plugins/modules/tower_project.py
john-westcott-iv: Doesn't Ansible usually air on the side of caution and report false changes instead of false "OKs"? i.e. the command module returns changed by default.

PR #10003, opened 2021-04-23
Log adjustments made to the analytics interval due to the 4-week limit

jbradberry:
##### SUMMARY
If the user passes in the `since` and/or `until` parameters, sometimes they will be truncated to limit the data to only 4 weeks.  Emit a log message when this happens.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```

```

### DIFF
--- a/awx/main/analytics/core.py
+++ b/awx/main/analytics/core.py
@@ -116,6 +116,51 @@ def package(target, data, timestamp):
         return None
 
 
+def calculate_collection_interval(since, until):
+    _now = now()
+
+    # Make sure that the endpoints are not in the future.
+    if until is not None and until > _now:
+        until = _now
+        logger.warning(f"End of the collection interval is in the future, setting to {_now}.")
+    if since is not None and since > _now:
+        since = _now
+        logger.warning(f"Start of the collection interval is in the future, setting to {_now}.")
+
+    # The value of `until` needs to be concrete, so resolve it.  If it wasn't passed in,
+    # set it to `now`, but only if that isn't more than 4 weeks ahead of a passed-in
+    # `since` parameter.
+    if since is not None:
+        if until is not None:
+            if until > since + timedelta(weeks=4):
+                until = since + timedelta(weeks=4)
+                logger.warning(f"End of the collection interval is greater than 4 weeks from start, setting end to {until}.")
+        else:  # until is None
+            until = min(since + timedelta(weeks=4), _now)
+    elif until is None:
+        until = _now
+
+    if since and since >= until:
+        logger.warning("Start of the collection interval is later than the end, ignoring request.")
+        raise ValueError
+
+    # The ultimate beginning of the interval needs to be compared to 4 weeks prior to
+    # `until`, but we want to keep `since` empty if it wasn't passed in because we use that
+    # case to know whether to use the bookkeeping settings variables to decide the start of
+    # the interval.
+    horizon = until - timedelta(weeks=4)
+    if since is not None and since < horizon:
+        since = horizon
+        logger.warning(f"Start of the collection interval is more than 4 weeks prior to {until}, setting to {horizon}.")
+
+    last_gather = settings.AUTOMATION_ANALYTICS_LAST_GATHER or horizon
+    if last_gather < horizon:
+        last_gather = horizon
+        logger.warning(f"Last analytics run was more than 4 weeks prior to {until}, using {horizon} instead.")
+
+    return since, until, last_gather
+
+
 def gather(dest=None, module=None, subset=None, since=None, until=None, collection_type='scheduled'):
     """
     Gather all defined metrics and write them as JSON files in a .tgz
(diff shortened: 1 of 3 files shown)


chrismeyersfsu:
It would be great to refactor this code so that all the since and until logic could be unit tested in isolation.

```
def calculate_since_until(since, until, last_gather, window=timedelta(weeks=4)):
    ....
    return since, until
```

jbradberry:
> It would be great to refactor this code so that all the since and until logic could be unit tested in isolation.
> 
> ```
> def calculate_since_until(since, until, last_gather, window=timedelta(weeks=4)):
>     ....
>     return since, until
> ```

Early on in the `core.gather()` function is not the only place where this type of limiting calculation occurs.  By necessity, the slicers have to manage this calculation themselves (e.g. the job events slicer is doing an apples-to-oranges comparison of timestamps vs pks).  I'm trying to imagine what a broken-out time calculation function would wind up looking like if it had to be used in all of those places, and I'm imagining it would wind up being only the timedelta part, with the comparisons of `since` with `LAST_GATHER` and `LAST_ENTRIES` having to take place outside the new function.

jladdjr (approved):

awx/main/tests/unit/analytics/test_collection_interval.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+class TestIntervalWithSinceAndUntil:
+    @pytest.mark.parametrize('gather', [None, 2, 6])
+    def test_ok(self, caplog, gather):
+        _now = now()
+        _prior = _now - timedelta(weeks=gather) if gather is not None else None
+        until = _now
+        since = until - timedelta(weeks=3)
+        settings.AUTOMATION_ANALYTICS_LAST_GATHER = _prior
```
chrismeyersfsu, line 22: This can result in the setting changing in other tests. There is a `settings` fixture that can be used that handles this.

jbradberry: ok, should be fixed now

awx/main/tests/unit/analytics/test_collection_interval.py
```diff
@@ -0,0 +1,192 @@
...
+
+        new_since, new_until, last_gather = calculate_collection_interval(since, until)
+        assert new_since == since
+        assert new_until == until
+
+        # 1 log message, if LAST_GATHER was more than 4 weeks ago
+        expected = 1 if gather and gather > 4 else 0
+        assert len(caplog.records) == expected
```
chrismeyersfsu, line 30: I'm liking this `caplog` sorcery

chrismeyersfsu (approved):

awx/main/analytics/core.py
```diff
@@ -116,6 +116,51 @@ def package(target, data, timestamp):
         return None
 
 
+def calculate_collection_interval(since, until):
```
jladdjr, line 119: reading through this again, after spending a few days away from it, it reads well / makes a lot of sense. this feels solid 👍

jladdjr: .. and it's awesome to see all this logic refactored into its own method

awx/main/tests/unit/analytics/test_collection_interval.py
jladdjr: > This can result in the setting changing in other tests. There is a `settings` fixture that can be used that handles this.

good catch ^

awx/main/tests/unit/analytics/test_collection_interval.py
```diff
@@ -0,0 +1,191 @@
...
+
+from django.utils.timezone import now, timedelta
+
+from awx.main.analytics.core import calculate_collection_interval
+
+# This is some bullshit.
+logger = logging.getLogger('awx.main.analytics')
+logger.propagate = True
+
+epsilon = timedelta(minutes=1)
```
jladdjr, line 12: advanced calc flashback 😄

awx/main/tests/unit/analytics/test_collection_interval.py
```diff
@@ -0,0 +1,191 @@
...
+        assert sum(1 for msg in caplog.messages if "later than the end" in msg) == 1
+
+    def test_until_in_future(self, caplog):
+        _now = now()
+        since = _now - timedelta(weeks=1)
+        until = _now + timedelta(weeks=1)
+
+        new_since, new_until, _ = calculate_collection_interval(since, until)
+        assert new_since == since
+        assert abs(_now - new_until) < epsilon  # `until` gets truncated to now()
```
jladdjr, line 56: man, that is practical. nice 👍

jladdjr (approved):
💯 

so, so good to see this logic nailed down w/ unit tests. calendar math = 🐉 🐉 🐉

awx/main/tests/unit/analytics/test_collection_interval.py
jbradberry: I _am_ still a mathematician at heart.

Issue #9371, opened 2021-02-19
Project Update fails with no output


hjkatz:
##### ISSUE TYPE
 - Bug Report

##### ENVIRONMENT
* AWX version: 17.0.1
* AWX install method: kubernetes
* Ansible version:  2.9

##### STEPS TO REPRODUCE

Launch a Job Template with a custom SCM branch. Have that SCM branch been recently updated via a `git commit --amend` and `git push --force`, therefore forcing a rebase or a `git pull --force` to take place.

##### EXPECTED RESULTS

The project should update properly and the job should run.

##### ACTUAL RESULTS

No output in the Job screen: 
![image](https://user-images.githubusercontent.com/2234516/108566967-4e2b6780-72d5-11eb-925a-08a0b30a6ee7.png)

Some api information from `/api/v2/jobs/3340/`
```
    "job_explanation": "Previous Task Failed: {\"job_type\": \"project_update\", \"job_name\": \"Shared Ansible Repo\", \"job_id\": \"3341\"}",
    "execution_node": "awx-job-3340",
    "controller_node": "awx-65b9694b58-7chbw",
    "result_traceback": "Traceback (most recent call last):\n  File \"/var/lib/awx/venv/awx/lib/python3.6/site-packages/awx/main/tasks.py\", line 1385, in run\n    self.pre_run_hook(self.instance, private_data_dir)\n  File \"/var/lib/awx/venv/awx/lib/python3.6/site-packages/awx/main/tasks.py\", line 1944, in pre_run_hook\n    sync_task.run(local_project_sync.id)\n  File \"/var/lib/awx/venv/awx/lib/python3.6/site-packages/awx/main/tasks.py\", line 816, in _wrapped\n    return f(self, *args, **kwargs)\n  File \"/var/lib/awx/venv/awx/lib/python3.6/site-packages/awx/main/tasks.py\", line 1557, in run\n    raise AwxTaskError.TaskError(self.instance, rc)\nException: project_update 3341 (failed) encountered an error (rc=2), please see task stdout for details.\n",
    "event_processing_finished": true,
    "job_template": 35,
```

And from `/api/v2/project_updates/3341/`

```
[...]

PR #10005, opened 2021-04-23
Update with force option if branch_override

AlanCoding:
Addresses a reported issue where force-pushing to feature branches did not work.

Connect https://github.com/ansible/awx/issues/9371#issuecomment-784261042

### DIFF
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -1766,6 +1766,7 @@ class RunJob(BaseTask):
             )
             if branch_override:
                 sync_metafields['scm_branch'] = job.scm_branch
+                sync_metafields['scm_clean'] = True  # to accomidate force pushes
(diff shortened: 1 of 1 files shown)


wenottingham:
Does this not fail when it's a non-branch_override force push?

If it's required to fix it, fine - I'd just expect it would fail in more cases.

AlanCoding:
> Does this not fail when it's a non-branch_override force push?

I don't fully understand the question. If it's non-branch_override, then we don't set this. That case is not changed here.

This is only used in the case where we're doing a project sync in advance of a job that is overriding the branch. This might be intuitive because this is the only case where a project sync might change from run-to-run (if job specifies a branch, then tip of that branch changes over time).

wenottingham (approved):

PR #10007, opened 2021-04-23
temporary API endpoint for custom virtual env data

rebeccahhh:
##### SUMMARY
related to: #7062 
This creates a new API endpoint for custom virtual environments. It spits out the pip freeze data for the custom virtual environments so that data can be used to create an execution environment with the same requirements, should that be desired.

This API endpoint will not be long lived and will only be used to help with the transition from Custom Venvs to Execution Environments.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.0.0
```

### DIFF
--- a/awx/api/urls/custom_venvs.py
+++ b/awx/api/urls/custom_venvs.py
@@ -1,0 +1,14 @@
+# Copyright (c) 2017 Ansible, Inc.
+# All Rights Reserved.
+
+from django.conf.urls import url
+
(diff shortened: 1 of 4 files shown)


rebeccahhh:
closing this in favor of new work that will shift this logic into an awx-manage command, see #7062 for full history

PR #10011, opened 2021-04-23
Remove 'custom' as a Source Option from inventory_source Module

beeankha:
These changes are related to PR https://github.com/ansible/awx/pull/9822


AlanCoding (approved):
Yeah, sorry I totally missed this

cc @elyezer we got most other removals in the collection

PR #10023, opened 2021-04-26
Db partition analytics cmeyers2

chrismeyersfsu:
* Keep old primary key based analytics gathering for unpartitioned
tables.
* Use created time on new partitioned tables.

<p align="center"><b>80 million partitioned + 1.5 million unpartitioned Events</b></p>

| Query                                                        | awx-manage gather_analytics --dry-run Time | Micro Benchmark Query Time* | Query Only Time** |
| ------------------------------------------------------------ | ------------------------------------------ | --------------------------- | ----------------- |
| sequential index scan, multiple `::json` casts, 100,000 event batches | 102m7.836s                                 | 6s                          | 80 minutes        |
| sequential index scan, optimized json cast, 100,000 event batches | 48m9.276s                                  | 2.2s                        | 30.4 minutes      |
| sequential index scan, optimized json cast, 1,00,000 event batches | 39m35.094s                                 | 10s                         | 13.3 minutes      |
| sequential **table** scan, optimized json cast, per-partition batch 600,000 *** | 36m42.081s                                 | 11.5s                       | 25.5 minutes      |

*micro benchmarking consists of simply copying a query, running it manually, and observing the runtime.
**micro benchmark time x (80 million / batch size)
**Note that this testing does NOT include the extra modified range query that is needed for correctness. We _expect_ this to be quite fast and is only needed to catch edge case events.


awx/main/management/commands/cleanup_jobs.py
```diff
@@ -36,6 +164,43 @@ def add_arguments(self, parser):
...
+        return (delete_meta.jobs_no_delete_count, delete_meta.jobs_to_delete_count)
+
+    def _cascade_delete_job_events(self, model, pk_list):
+        if len(pk_list) > 0:
+            with connection.cursor() as cursor:
+                tblname = unified_job_class_to_event_table_name(model)
+
+                pk_list_csv = ','.join(map(str, pk_list))
+                rel_name = model().event_parent_key
+                cursor.execute(f"DELETE FROM _unpartitioned_{tblname} WHERE {rel_name} IN ({pk_list_csv})")
```
Ladas, line 202: @chrismeyersfsu are we dropping the table somewhere after everything is deleted? Since it won't return the storage space to OS until we drop it (when deleted, the space can be only reused by this table)

If we need the table here because of some queries checking it, we can create new _unpartitioned_{tblname}, switch names and drop the old one

chrismeyersfsu:
recheck

kdelee (approved):

jladdjr:
thx @kdelee!

PR #10030, opened 2021-04-27
Add ouiaId to toolbar delete button

akus062381:
In fixing UI tests, I came across an element that was missing testability. I added specific ouiaIds to this element, to override the automatically generated ones. This will make testability much easier for QE.

image


AlexSCorey:
recheck

PR #10031, opened 2021-04-27
Assorted renaming and string changes

wenottingham:
##### SUMMARY

Goal was to hit user-facing strings and docs.

Didn't try and catch all comments, although did a few when in those files.


awx/main/notifications/webhook_backend.py
```diff
@@ -75,7 +72,7 @@ def send_messages(self, messages):
                 "{}".format(m.recipients()[0]),
                 auth=auth,
                 data=json.dumps(m.body, ensure_ascii=False).encode('utf-8'),
-                headers=self.headers,
+                headers=get_awx_http_client_headers(),
```
wenottingham, line 75: Actual code change, but should do the right thing here (didn't want to cut & paste the user agent templating)

wenottingham:
Note: this does not do anything with the name of the inventory source. That requires more thought.

AlexSCorey (changes requested):
I pulled down this code locally.  The strings themselves seem to be fine, but you have 1 unit test failure that needs to be addressed `MiscSystemDetail.test.jsx`

wenottingham:
> I pulled down this code locally. The strings themselves seem to be fine, but you have 1 unit test failure that needs to be addressed `MiscSystemDetail.test.jsx`

Fixed, I think. (Also fixed prettier lint failure.)

AlexSCorey (approved):
Looks good!

PR #10067, opened 2021-04-30
Unit test warning cleanup

mabashian:
##### SUMMARY
These commits target the warnings that look like:

`Can't perform a React state update on an unmounted component`

The underlying problem here is that we have network requests that are being made by components that are subsequently being unmounted.  When the network request resolves, we attempt to update some state but the component is no longer mounted and the warning is triggered.  To address this I consolidated a lot of `isMounted` code into a single hook which can be used across the app to check to see whether the component in question is still mounted before attempting to update state inside of a `useEffect`.  This primarily applies to network requests.

I think this points to a larger issue which is that we sometimes mount components prematurely.  For example, when the job template edit component is mounted we actually mount:

1) JobTemplateForm (briefly)
2) ContentLoading
3) JobTemplateForm

Network requests triggered by the first mount of JobTemplateForm are suscepitble to attempting to update state on an unmounted component.  I believe this pattern exists in many places across the app but I haven't tried to solve this in this PR.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/util/useIsMounted.js
+++ b/awx/ui_next/src/util/useIsMounted.js
@@ -1,0 +1,12 @@
+import { useEffect, useRef } from 'react';
+
+export default function useIsMounted() {
+  const isMounted = useRef(null);
+  useEffect(() => {
+    isMounted.current = true;
+    return () => {
+      isMounted.current = false;
+    };
+  });
+  return isMounted;
+}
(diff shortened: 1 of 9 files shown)


nixocio:
I just triggered an e2e test run. [🟢 ](https://dashboard.cypress.io/projects/ui2mq7/runs/5783/overview)

awx/ui_next/src/util/useIsMounted.js
```diff
@@ -0,0 +1,12 @@
+import { useEffect, useRef } from 'react';
+
+export default function useIsMounted() {
```
marshmalien, line 3: 👍

marshmalien (approved):

nixocio (approved):
I got this change locally ran the test, the mentioned warnings were removed.

PR #10068, opened 2021-04-30
Have the project update playbook warn if role/collection syncing is disabled

wenottingham:
In recent AWX, a galaxy credential (even a blank one for galaxy.ansible.com) is required to sync role/collection content. This is done so that server precedence can be properly set, and so that it can be configured to pull only from a private content host.

This does lead to bug reports where the credentials are not set, and users don't understand why their content is not syncing. This makes that more clear.

Screenshot from 2021-04-30 16-04-52

### DIFF
--- a/awx/playbooks/project_update.yml
+++ b/awx/playbooks/project_update.yml
@@ -156,6 +156,21 @@
   name: Install content with ansible-galaxy command if necessary
   tasks:
 
+    - name: Check content sync settings
+      debug:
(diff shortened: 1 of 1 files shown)


wenottingham:
cc @AlanCoding @shanemcd

AlanCoding:
I wish there was anything we could do better for this, like a skip_reason.

We could split the enable settings into 2, for the global setting vs. the org credential presence. This would be pretty minor, and not worth it.

For the UX, it's just completely unclear that removing the credential (or just migrating your data) means doing an org-specific disabling of pulling content.

AlanCoding (approved):

Issue #10008, opened 2021-04-23
awx.awx.tower_job_launch tags are not properly passed to API


mwester117:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
If you are using awx.awx.tower_job_launch to launch a job-/workflowtemplate in AWX and you pass tags to the job launch they will be ignored. Against expectations all tasks are executed not only those with the provided tags 

##### ENVIRONMENT
* AWX version: Tower 3.8.2
* Ansible version:   Ansible 2.9.13 
* Operating System: RHEL 8.3

##### STEPS TO REPRODUCE
Create a playbook similar to this:
```
- hosts: localhost
  gather_facts: no
  environment:
    TOWER_USERNAME: "<<INPUT TOWER USERNAME HERE>>"
    TOWER_HOST: "<<INPUT TOWER HOST HERE>>"
    TOWER_PASSWORD: "<<INPUT TOWER USER PASSWORD HERE>>"
  vars:
    ansible_connection: local

  tasks:
  - name: Launch a job template with a tag on remote Tower instance
    awx.awx.tower_job_launch:
      job_template: "test_template"
      inventory: "test_inventory"
      tags:
        - example_tag
```
Create a second playbook named "test_template" with two random tasks. One needs to be tagged with "example_tag"
 

##### EXPECTED RESULTS

 Only the task tagged with "example_tag" will be executed

##### ACTUAL RESULTS

All tasks will be executed

##### ADDITIONAL INFORMATION

 The error occurs because [tower_job_launch.py](https://github.com/ansible/awx/blob/devel/awx_collection/plugins/modules/tower_job_launch.py) passes the wrong post data to the api. The tags have to be provided under key "job_tags". This can be found in [API documentation](https://docs.ansible.com/ansible-tower/latest/html/towerapi/api_ref.html#/Job_Templates/Job_Templates_job_templates_launch_read). Certainly the module passes the tags with the key "tags" which is wrong

I think the follwing code needs to be adjusted:
https://github.com/ansible/awx/blob/devel/awx_collection/plugins/modules/tower_job_launch.py#L184

The correct line should be: 
`    optional_args['job_tags'] = module.params.get('tags')`

I will test the suggestion on monday and report the result.

sean-m-sullivan:
While we are fixing skip_tags also does not seem happy with either string or list. 
Here is a good demo playbook with tags of always, never, sometimes, maybe
[...]

PR #10077, opened 2021-05-02
Fix tower_job_launch tags being passed to API

sean-m-sullivan:
##### SUMMARY
#10008
Fix tower_job_launch tags being passed to API
Previously the wrong field was being used on the api to pass tags, also while the module accepts lists, the API does not, so it would error. Take a users list, and converts to comma separated string, in order to maintain status quo. 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - awx_collection

##### AWX VERSION
```
19.1.0
```

### DIFF
--- a/awx_collection/plugins/modules/tower_job_launch.py
+++ b/awx_collection/plugins/modules/tower_job_launch.py
@@ -176,21 +176,33 @@ def main():
     optional_args = {}
     # Extract our parameters
     name = module.params.get('name')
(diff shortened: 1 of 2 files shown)


beeankha (approved):
Thank you for this work, @sean-m-sullivan !

rebeccahhh (approved):

AlanCoding (approved):

PR #10090, opened 2021-05-03
add a new awx-manage command `custom_venvs`

rebeccahhh:
add an awx-manage command that gets pip freeze data from custom_venv and outputs to command line stdout

##### SUMMARY

part of #7062  - this command is a glorified pip freeze + some extra stuff, people could navigate to each of their custom virtual environments themselves and run a pip freeze, but this allows them to not, and everyone likes their life to be easier. The extra stuff allows users to see the connections that their existing virtual envs have in awx to things like organizations, jobs, inventory updates, and projects.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.1.0
```

##### ADDITIONAL INFORMATION

This is built off of existing code and there is a line that gets custom venv paths from the settings module, that line does not seem to be working. I have written around that but want to make a note of it.


awx/api/views/__init__.py
```diff
@@ -295,6 +296,7 @@ def get(self, request, format=None):
         data['teams'] = {'url': reverse('api:team_list', request=request), 'total': team_list.count()}
         data['credentials'] = {'url': reverse('api:credential_list', request=request), 'total': credential_list.count()}
         data['job_templates'] = {'url': reverse('api:job_template_list', request=request), 'total': job_template_list.count()}
+        data['custom_venvs'] = {'url': reverse('api:custom_venv_list', request=request)}
```
AlanCoding: should this still be here?

awx/main/management/commands/custom_venvs.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def handle(self, *args, **options):
+        super(Command, self).__init__()
+        if options.get('path'):
+            pip_data = get_custom_venv_pip_freeze(options.get('path'))
+            print(pip_data)
+        else:
+            venvs = get_custom_venv_choices()
+            for venv in venvs:
+                print(venv)
```
AlanCoding: how does the user know what line corresponds to what custom venv?

awx/api/views/__init__.py
rebeccahhh: oh no it should not, thanks

awx/main/management/commands/custom_venvs.py
rebeccahhh: only one venv path can be passed at a time

rebeccahhh: ah, sorry I misunderstood, the path is the only info shared here, I don't know that there's a good way to demonstrate any more than that which path corresponds to which venv identity

AlanCoding: oh right, yeah that flew past me.

So this usage doesn't show any pip freeze data.

rebeccahhh: if you pass nothing/no path in the command it spits out a list of paths for you to view, if you pass one of those paths in it will spit out the pip freeze data

awx/main/utils/common.py
```diff
@@ -887,30 +888,38 @@ def get_current_apps():
...
                 )
         except Exception:
             logger.exception("Encountered an error while discovering custom virtual environments.")
     return custom_venv_choices
 
 
+def get_custom_venv_pip_freeze(venv_path):
+    try:
+        freeze_data = subprocess.run([f"{venv_path}/pip", "freeze"], capture_output=True)
+        pip_data = (freeze_data.stdout).decode('ascii')
```
jbradberry: I'm a bit uncomfortable with assuming they are guaranteed to be ASCII.

jbradberry: I tried looking around to see if the output of `pip freeze` would ever contain non-ASCII characters but don't see anything.  The closest I see is that pip supports a UTF-8 encoding for requirements files: https://pip.pypa.io/en/stable/cli/pip_install/#requirements-file-format

rebeccahhh: subprocess outputs to byte format by default, so I have to decode. I picked ascii arbitrarily so I'm not particularly tied to it, that being said, I _highly_ doubt non-ascii characters would ever be present. 
I could decode to utf-8 if we think it's ever going to be problematic

jbradberry: yeah, hard to say.  I would mostly agree, but haven't come up with any hard reason why a package couldn't be named something with some non-ascii characters.

rebeccahhh: will there ever be any `'`? 😂 cause then we've got trouble.

awx/main/utils/common.py
```diff
@@ -887,30 +888,38 @@ def get_current_apps():
...
+def get_custom_venv_choices():
     from django.conf import settings
 
-    custom_paths = custom_paths or settings.CUSTOM_VENV_PATHS
-    all_venv_paths = [settings.BASE_VENV_PATH] + custom_paths
+    all_venv_paths = []
+
+    all_venv_paths.append(settings.BASE_VENV_PATH)  # get paths from settings
+
+    for root, dir, files in os.walk('../awx_devel/var/lib/awx/venv/'):  # get paths on machine
```
AlanCoding: what's going on with `/awx_devel` here? Looks like something leftover from debugging in the development environment.

awx/main/utils/common.py
```diff
@@ -887,30 +888,38 @@ def get_current_apps():
     return current_apps
 
 
-def get_custom_venv_choices(custom_paths=None):
+def get_custom_venv_choices():
     from django.conf import settings
 
-    custom_paths = custom_paths or settings.CUSTOM_VENV_PATHS
```
AlanCoding, line 896: I don't see how we get around the need for `CUSTOM_VENV_PATHS` which is removed here. Without that, I don't see how this command could make sense.

awx/main/management/commands/custom_venvs.py
```diff
@@ -0,0 +1,30 @@
...
+class Command(BaseCommand):
+    """Returns either a list of custom venv paths or outputs the pip freeze from the path passed in the argument"""
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--path',
+            dest='path',
+            type=str,
+            default='',
+            help='run without arguments to see a list of paths, run with one of those paths as an argument and see the pip freeze data',
```
AlanCoding: As a matter of wording:

> run without arguments to see a list of paths

I think that should be left for the general command help text to establish (as opposed to the help text about the option).

Will this take either an absolute path or a relative path?

awx/main/utils/common.py
rebeccahhh: that command actually doesn't return anything anyways 🤷🏻

rebeccahhh: oh my gosh @AlanCoding I actually take that back. I'm so glad you said this because I swapped this call and the one on line 896 which is why that command doesn't return anything. I'll update 🎊

awx/main/utils/common.py
rebeccahhh: swapped to utf-8

awx/main/management/commands/custom_venvs.py
rebeccahhh: it's an absolute path, I'll work on some wording, maybe with @tvo318 so it matches with docs and is helpful to users

awx/main/management/commands/custom_venvs.py
```diff
@@ -0,0 +1,28 @@
...
+from awx.main.utils.common import get_custom_venv_choices, get_custom_venv_pip_freeze
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    """Returns either a list of custom venv paths or outputs the pip freeze from the path passed in the argument"""
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--path',
```
shanemcd: Since we only have one argument, what do you think about just making it positional?

rebeccahhh: I feel great about that

awx/main/utils/common.py
rebeccahhh: yep! It's gone now :)

awx/main/management/commands/custom_venvs.py
rebeccahhh: I feel less great now, the way that parser works requires the flags for the variable to be optional, so it would have to be separated into 2 different commands if I want it to be just positional

rebeccahhh:
just bumped the collector version, logic remains the same

awx/main/management/commands/custom_venvs.py
rebeccahhh: for future reference it is possible, I was just doing it wrong :)

rebeccahhh:
split this into 2 commands instead of one for clarity `export_custom_venv` and `list_custom_venvs`

awx/main/management/commands/export_custom_venv.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class Command(BaseCommand):
+    """Returns either the pip freeze from the path passed in the argument"""
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'path',
+            type=str,
+            nargs='?',
```
shanemcd: I think we should drop this now.

rebeccahhh: the whole thing or just `nargs`?

AlanCoding: I think the `nargs`?

kdelee:
~Whatever happened with being able to figure out what venv each JT/org/etc used in the past, so the user could have a clue to what new replacement EE they should use?~
~Is that just staying on the JT/org/etc model but not showing in the UI? can't rememeber~

We discussed and will be adding a 3rd command to list where a venv is used

awx/main/management/commands/export_custom_venv.py
```diff
@@ -0,0 +1,32 @@
...
+        )
+
+    def handle(self, *args, **options):
+        super(Command, self).__init__()
+        if options.get('path'):
+            pip_data = get_custom_venv_pip_freeze(options.get('path'))
+            if pip_data:
+                print(pip_data)
+                msg = [
+                    'To list all available custom virtual environments run:',
```
kdelee: Probably should rephrase this...The virtual environments are present, but not "available" for use.

Maybe:
`'To list all (now deprecated) custom virtual environments run:'`

awx/main/management/commands/list_custom_venvs.py
```diff
@@ -0,0 +1,33 @@
...
+# All Rights Reserved
+import sys
+
+from awx.main.utils.common import get_custom_venv_choices
+from django.core.management.base import BaseCommand
+from django.conf import settings
+
+
+class Command(BaseCommand):
+    """Returns either a list of custom venv paths from the path passed in the argument"""
```
kdelee: remove `either` has it doesn't make sense in the sentence

kdelee:
@rebeccahhh the extra text is helpful to the first time user, but will cause problems to people who might want to use these commands in a small script or redirect the output

For example, if we could just get the raw output without the help text by passing `-q`, I could use bash to loop over the venvs and cat the pip freeze to a file

```
$ for path in $(awx-manage list_custom_venvs -q); do awx-manage export_custom_venv -q $path > $path/freeze.txt; done
```

rebeccahhh:
moving this back to in progress since we've decided to make this tool a little more robust with a third command

awx/main/management/commands/get_custom_venv_associations.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def handle(self, *args, **options):
+        # look organiztions and unified job templates (which include JTs, workflows, and Inventory updates)
+        super(Command, self).__init__()
+        results = {}
+        if options.get('path'):
+            # sanity check here - is path in list
+            path = options.get('path')
+            orgs = [{"name": org.name, "id": org.id} for org in Organization.objects.filter(custom_virtualenv=path)]
+            ujts = [{"name": org.name, "id": org.id} for org in Organization.objects.filter(custom_virtualenv=path)]
```
shanemcd: This says `ujts` but it looks to be a list of `Organization`s?

rebeccahhh: oh that was me copy and pasting and moving too quickly last night!

awx/main/management/commands/get_custom_venv_associations.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class Command(BaseCommand):
+    """Returns the pip freeze from the path passed in the argument"""
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'path',
+            type=str,
+            nargs='?',
```
shanemcd: It seems like this should be a required argument, I think? The `?` here makes it optional. Not sure that's what you want.

rebeccahhh: I currently handle that in the handle function but it's probably better to handle it here.

awx/main/management/commands/get_custom_venv_associations.py
```diff
@@ -0,0 +1,35 @@
...
+            default='',
+            help='run this with a path to a virtual environment as an argument to see the pip freeze data',
+        )
+
+    def handle(self, *args, **options):
+        # look organiztions and unified job templates (which include JTs, workflows, and Inventory updates)
+        super(Command, self).__init__()
+        results = {}
+        if options.get('path'):
```
shanemcd: You can ditch this conditional once you make the argparse change.

awx/main/management/commands/get_custom_venv_associations.py
```diff
@@ -0,0 +1,35 @@
...
+            # sanity check here - is path in list
+            path = options.get('path')
+            orgs = [{"name": org.name, "id": org.id} for org in Organization.objects.filter(custom_virtualenv=path)]
+            ujts = [{"name": org.name, "id": org.id} for org in UnifiedJobTemplate.objects.filter(custom_virtualenv=path)]
+            invsrc = [{"name": inv.name, "id": inv.id} for inv in InventorySource.objects.filter(custom_virtualenv=path)]
+            results["Organizations"] = orgs
+            results["Unified Job Templates"] = ujts
+            results["Inventory Sources"] = invsrc
+            print(results)
```
shanemcd: Would it be more readable to print yaml here?

rebeccahhh: 100%. This whole thing needs a bit of polish, I think yaml would be the most readable

awx/main/management/commands/get_custom_venv_associations.py
```diff
@@ -0,0 +1,35 @@
...
+class Command(BaseCommand):
+    """Returns the pip freeze from the path passed in the argument"""
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'path',
+            type=str,
+            nargs=1,
+            default='',
+            help='run this with a path to a virtual environment as an argument to see the pip freeze data',
```
kdelee: update copy pasta

awx/main/management/commands/get_custom_venv_associations.py
```diff
@@ -0,0 +1,35 @@
...
+        results = {}
+        if options.get('path'):
+            # sanity check here - is path in list
+            path = options.get('path')[0]
+            orgs = [{"name": org.name, "id": org.id} for org in Organization.objects.filter(custom_virtualenv=path)]
+            jts = [{"name": jt.name, "id": jt.id} for jt in JobTemplate.objects.filter(custom_virtualenv=path)]
+            proj = [{"name": proj.name, "id": proj.id} for proj in Project.objects.filter(custom_virtualenv=path)]
+            invsrc = [{"name": inv.name, "id": inv.id} for inv in InventorySource.objects.filter(custom_virtualenv=path)]
+            results["Organizations"] = orgs
+            results["Job Templates"] = ujts
```
kdelee: undefined. use same variable as defined on line 28...in this case, `jts`

awx/main/management/commands/get_custom_venv_associations.py
```diff
@@ -0,0 +1,35 @@
...
+            help='run this with a path to a virtual environment as an argument to see the pip freeze data',
+        )
+
+    def handle(self, *args, **options):
+        # look organiztions and unified job templates (which include JTs, workflows, and Inventory updates)
+        super(Command, self).__init__()
+        results = {}
+        if options.get('path'):
+            # sanity check here - is path in list
+            path = options.get('path')[0]
```
kdelee: I would add validation that this is in list of custom venvs, otherwise we can get tripped up --especially with trailing slash

```
# works as expected
# awx-manage get_custom_venv_associations /tmp/elijahfoo
{'Organizations': [], 'Job Templates': [{'name': 'Demo Job Template', 'id': 7}], 'Projects': [], 'Inventory Sources': []}

# does not
# awx-manage get_custom_venv_associations /tmp/elijahfoo/
{'Organizations': [], 'Job Templates': [], 'Projects': [], 'Inventory Sources': []}
```

AlanCoding:
what does the use case for the newer venv associations command? How do users go from there to EE images?

awx/main/management/commands/custom_venv_associations.py
```diff
@@ -0,0 +1,37 @@
...
+                path = options.get('path')[0]
+                orgs = [{"name": org.name, "id": org.id} for org in Organization.objects.filter(custom_virtualenv=path)]
+                jts = [{"name": jt.name, "id": jt.id} for jt in JobTemplate.objects.filter(custom_virtualenv=path)]
+                proj = [{"name": proj.name, "id": proj.id} for proj in Project.objects.filter(custom_virtualenv=path)]
+                invsrc = [{"name": inv.name, "id": inv.id} for inv in InventorySource.objects.filter(custom_virtualenv=path)]
+                results["Organizations"] = orgs
+                results["Job Templates"] = jts
+                results["Projects"] = proj
+                results["Inventory Sources"] = invsrc
+                print(results)
```
AlanCoding: an `else:` here would be helpful.

If the user gives a path (don't they have to in order to run the command?) and it's not a valid path, then it would be more sensible to error rather than do nothing and exit silently.

rebeccahhh: I'll add it

awx/main/management/commands/list_custom_venvs.py
```diff
@@ -0,0 +1,33 @@
...
+        venvs = get_custom_venv_choices()
+        if venvs:
+            print('# {}'.format("Discovered virtual environments:"))
+            for venv in venvs:
+                print(venv)
+
+            msg = [
+                '',
+                'To export the contents of a (deprecated) virtual environment, ' 'run the following command while supplying the path as an argument:',
+                'awx-manage export_custom_venv /path/to/venv',
```
AlanCoding: Also mention the command to get the uses of that venv, and give the syntax that will work for that.

awx/main/management/commands/custom_venv_associations.py
AlanCoding: this may already be covered by the `nargs=1`, I just want to set the expectation that giving no arguments doesn't do nothing and return

AlanCoding (approved):

rebeccahhh:
recheck

rebeccahhh:
@kdelee we've got a third command now and a -q added as well

rebeccahhh:
recheck

shanemcd (approved):

Issue #10055, opened 2021-04-29
Proposal, Add back Custom virtual env fields in modules


sean-m-sullivan:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Proposal:
Return the code to have custom virtual environments in their applicable modules that were removed in the update to 17 -> 18. 

The reason being is that there are tower users that use awx.awx because it contains the latest updates and feature improvements, the vast majority still work on Tower 3.7.x,3.8.x.

These would updated and noted in the docs that it is meant for older versions of awx, and will fail on newer. 

The expectation is that the options when used with the wrong version would fail, and it would be on the users to use the right one. I also thing putting depreciated on custom virtual environment options would be appropriate.

The idea is to make it so that users can use the same features and upgrades done to the collection regardless of the version they are using, and put the onus on them to use the correct ones that correspond to their instance. 

I would do the PR for this, but wanted to gather feedback before submitting it.

weiyentan:
Hmm with execution enviroments I can add the collection version I want. And ansible runner will use that...ansible builder can do that. I supply a list of collections I want to pull and then ansible-builder bundles them up and when the job template runs, it executes in the environment using the correct collection version.

At the moment I have a execution environment with Ansible 2.9 (the one that goes bundled is 2.11) tower-configuration (i can choose version) and other stuff... 

Have I misunderstood what you are after?

sean-m-sullivan:
This is for Older versions of Tower/Awx that don't have execution environments, making the modules compatible with both at the same time while people transition from one to the other.

shanemcd:
I'd be interested in learning if we have a pattern in place for backwards compatibility.

PR #10097, opened 2021-05-04
add custom_venv for backward compatibility

sean-m-sullivan:
##### SUMMARY
As per #10055 , this is the proposal to add back custom virtual environments to the modules. With the understanding that they will not be tested. But to allow users of Ansible Tower 3.8 the ability to use newer features and bugfixes in awx.awx as applicable, without restorting to backporting these removals. Would expect these to remain for a time after Tower moves to Execution environments as teams/companies adopt the next version of tower. 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - awx_collection

##### AWX VERSION
```
19.1.0
```

### DIFF
--- a/awx_collection/plugins/modules/tower_inventory_source.py
+++ b/awx_collection/plugins/modules/tower_inventory_source.py
@@ -72,6 +72,12 @@ options:
       description:
         - Execution Environment to use for the source.
       type: str
+    custom_virtualenv:
+      description:
+        - Local absolute file path containing a custom Python virtualenv to use.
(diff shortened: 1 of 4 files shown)


AlanCoding (approved):

john-westcott-iv:
Just asking here, why can't people just use the version of the collection that ships with their version of the product? Supporting old features is why the collection will forever have tower-cli.cfg references in it.

beeankha (approved):

Issue #8942, opened 2020-12-17
Http requests /me and /config been made twice on refresh


tiagodread:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Some pages, when you refresh the page the /me and /config requests are being made twice

##### STEPS TO REPRODUCE

1. Navigate to a page
2. Refresh the page reloading or force reload

##### EXPECTED RESULTS

The request for /me and /config not being executed twice

##### ACTUAL RESULTS

![image](https://user-images.githubusercontent.com/12464858/102522716-608aea00-4075-11eb-8b5c-ed9397c4243c.png)

mabashian:
This should be addressed by https://github.com/ansible/awx/pull/10133

akus062381:
this is fixed.

PR #10133, opened 2021-05-06
Adds support for pendo initialization across the app

mabashian:
##### SUMMARY
We were already bootstrapping pendo as part of the subscription code I just moved that code to a more general place.  When the app container mounts (after login or on refresh) we check to see if the pendo flag is turned on.  If it is, we initialize pendo.  If it's not then we do nothing.  If a user goes into settings and manually changes the pendo tracking setting then we trigger a hard reload of the browser tab (to take the new setting into account and either initialize or not).  This functionality existed in the old UI as well.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI


mabashian:
link #8942

nixocio:
I triggered an e2e test run.

awx/ui_next/src/screens/Setting/Subscription/SubscriptionEdit/SubscriptionEdit.jsx
```diff
@@ -178,6 +163,18 @@ function SubscriptionEdit() {
...
+      const [
+        { data },
+        {
+          data: {
+            results: [me],
+          },
+        },
+      ] = await Promise.all([ConfigAPI.read(), MeAPI.read()]);
+      const newConfig = { ...data, me };
+      setConfig(newConfig);
```
nixocio: Minor suggestion. Could we also just import the request function from ```const { request, isLoading } = useConfig();``` and call the  request function? If we modify the API calls inside the Config, we will need to duplicate them also here to avoid to be overwritten.

awx/ui_next/src/components/AppContainer/AppContainer.jsx
```diff
@@ -138,6 +139,13 @@ function AppContainer({ navRouteConfig = [], children }) {
     }
   }, [handleLogout, timeRemaining]);
 
+  useEffect(() => {
+    if ('analytics_status' in config) {
+      issuePendoIdentity(config, 'foobar');
```
mabashian: Need to pull the real pendo key and pass it along here

nixocio (approved):

awx/ui_next/src/components/AppContainer/AppContainer.jsx
mabashian: Actually, I forgot that I pushed the logic to grab the api key down to the utility function so this extra param is not needed.

awx/ui_next/src/screens/Setting/Subscription/SubscriptionEdit/SubscriptionEdit.jsx
mabashian: @nixocio good idea, I think we do this in other places.  I'll look in to this.

mabashian: Actually we weren't doing this anywhere else but I pushed a commit that should achieve something similar.

tiagodread (approved):
I verified that the data is being sent successfully to pendo when the flag User Analytics is enabled in different scenarios.

PR #10141, opened 2021-05-07
Fix prop typo

nixocio:
Fix prop typo.

See: https://www.patternfly.org/2020.04/documentation/react/components/select#props


AlexSCorey (approved):
cool

Issue #7506, opened 2020-07-01
YAML code editor evals the YAML


chrismeyersfsu:
##### SUMMARY
The codemirror box "evals" the yaml references and expands the references. This is undesirable.

##### ENVIRONMENT
* AWX version: X.Y.Z
* AWX install method: openshift, minishift, docker on linux, docker for mac, boot2docker
* Ansible version:  X.Y.Z
* Operating System:
* Web Browser:

##### STEPS TO REPRODUCE

```
---
a: &aa [a,b,c]
b: *aa
```
Create a job template with the above extra vars in the box.

##### EXPECTED RESULTS

![image](https://user-images.githubusercontent.com/722880/86271958-d53b9c00-bb9b-11ea-913a-d34c3f68a822.png)

##### ACTUAL RESULTS

![image](https://user-images.githubusercontent.com/722880/86272042-f3090100-bb9b-11ea-808b-140caf515eec.png)

keithjgrant:
Behavior may have changed slightly here... I can only reproduce this now by toggling the variables to JSON and back to YAML. This makes sense on some level, since JSON doesn't support some of the expressions that YAML does (like &/* repeated nodes). @chrismeyersfsu are you still able to produce this bug without clicking the JSON toggle?

Regarding the toggle: Should we take steps to prevent this behavior? First thing that comes to mind is to keep track whether the value gets edited in JSON mode, and if it isn't, revert to the original YAML to preserve the special expressions

PR #10148, opened 2021-05-07
Add Variables field validation in launch prompt

keithjgrant:
##### SUMMARY
* Adds validation for the Variables field during launch prompts.
* Prevents unexpected expanding of YAML expressions in the default value (similar to #7506) — this was caused by the default value being parsed into a JavaScript object and then re-translated back to YAML when mounting the "Other Prompts" step

addresses #8907 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI


jakemcdermott (approved):

tiagodread (approved):
This bug is now fixed, we are showing this error message when the Variables validation has erros and the launch button is also disabled.

image

PR #10157, opened 2021-05-10
minikube instructions adjustments from debugging

AlanCoding:
##### SUMMARY
The awx-operator examples expect the `tower_image_version` as a separate var, and can error without that.

Also, there's still an issue getting going with `ansible/deploy-operator.yml`'s Deploy Operator task. I'm still looking into that, the associated hack might change.

##### ISSUE TYPE
 - Docs Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
19.1.0
```


shanemcd (approved):

AlanCoding:
I took the hack off because it wasn't necessary.

jakemcdermott:
regate

Issue #10144, opened 2021-05-07
Display error when modifying allow_override on projects


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Display error when modifying allow_override on projects being used by job templates.

Previous implementation:

<img width="1346" alt="image" src="https://user-images.githubusercontent.com/9053044/117497831-fc46c500-af46-11eb-82bb-51bd22b3c15b.png">

Current implementation:

<img width="1792" alt="image" src="https://user-images.githubusercontent.com/9053044/117497949-2304fb80-af47-11eb-8ce9-091ec44b4875.png">

akus062381:
all set

PR #10161, opened 2021-05-10
Display error for boolean fields

nixocio:
Display error when modifying Allow Override Branch on Projects.

See: https://github.com/ansible/awx/issues/10144

image

### DIFF
--- a/awx/ui_next/src/components/FormField/sortErrorMessages.js
+++ b/awx/ui_next/src/components/FormField/sortErrorMessages.js
@@ -52,6 +52,9 @@ function parseFieldErrors(obj, formValues) {
       }
       formErrors = formErrors.concat(parsed.formErrors);
     }
+    if (typeof formValues[key] === 'boolean') {
+      formErrors = formErrors.concat(value);
+    }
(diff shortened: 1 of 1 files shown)


nixocio:
I triggered an e2e test run.

awx/ui_next/src/components/FormField/sortErrorMessages.js
```diff
@@ -52,6 +52,9 @@ function parseFieldErrors(obj, formValues) {
       }
       formErrors = formErrors.concat(parsed.formErrors);
     }
+    if (typeof formValues[key] === 'boolean') {
```
keithjgrant, line 55: I wonder if this should just be an `else` block to catch any and all "other" types that could occur

nixocio: I tried the catch all - else, but the error was not displayed.

keithjgrant (approved):

akus062381:
@nixocio this is checked and verified, ok as far as QE is concerned!

Issue #10123, opened 2021-05-06
AWX 19.1.0 - Job settings cannot be saved (/worker-json.js returns 404)


fourstepper:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

Job settings cannot be saved, probably due to the `worker-json.js` returning 404.

##### ENVIRONMENT
* AWX version: 19.1.0
* AWX install method: minikube
* OS: Debian
* Browser: Chromium: 90.0.4430.93 (Official Build) (64-bit)

##### STEPS TO REPRODUCE

- Go the Settings > Jobs settings > Edit
- Tweak a setting
- Hit Save

##### EXPECTED RESULTS

- The settings are saved succesfully

##### ACTUAL RESULTS

- The settings never get updated, the save button doesn't do anything

##### ADDITIONAL INFORMATION

I am not sure what debug information would help, but I will be happy to provide any if required

wenottingham:
We'd need more detail on the error from the browser console.

tiagodread:
When the edit page is loaded for Job settings, there are some errors:

```Warning: Failed prop type: Invalid prop `groupTitle` of type `object` supplied to `NavExpandableGroup`, expected `string`.```

```Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of FindRefWrapper which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://fb.me/react-strict-mode-find-node```

```Warning: Failed prop type: The prop `config` is marked as required in `BooleanField`, but its value is `undefined`.```

```Warning: Failed prop type: The prop `config` is marked as required in `InputField`, but its value is `undefined`.```

Trying to save the form nothing happens.

**Comment by tiagodread** [comment]
This is a UI only problem I could patch the Job settings by the API

**Comment by fourstepper** [comment]
This is what shows up in the browser console during the whole operation

```
2.fe3515b6.chunk.js:2 I18nProvider did not render. A call to i18n.activate still needs to happen or forceRenderOnLocaleChange must be set to false.
worker-json.js:1 GET https://example.com/worker-json.js 404
worker-json.js:1 GET https://example.com/worker-json.js 404
worker-json.js:1 GET https://example.com/worker-json.js 404
worker-json.js:1 GET https://example.com/worker-json.js 404
worker-json.js:1 GET https://example.com/worker-json.js 404
worker-json.js:1 GET https://example.com/worker-json.js 404
[...]

PR #10171, opened 2021-05-11
Remove all AWX_ISOLATED* fields

jakemcdermott:
I believe this should address https://github.com/ansible/awx/issues/10123. The issue mentions a `worker-json.js` 404 and, while we should fix that at some point, I don't think it's actually related to the `save` not working.

##### SUMMARY
I noticed the job settings form wasn't saving on devel. 

After some debugging, I found that removing the `AWX_ISOLATED_CHECK_INTERVAL` and `AWX_ISOLATED_LAUNCH_TIMEOUT` made the form saveable again.

It doesn't seem like we use _any_ of the `AWX_ISOLATED*` fields anymore (not to be confused with `AWX_ISOLATION*` fields, which we do use) so this PR removes them all. 

cc @wenottingham @shanemcd @marshmalien  @gamuniz @tiagodread @jbradberry


jbradberry:
The settings I removed in #9957 are:

  - AWX_ISOLATED_PUBLIC_KEY
  - AWX_ISOLATED_PRIVATE_KEY
  - AWX_ISOLATED_KEY_GENERATION
  - AWX_ISOLATED_HOST_KEY_CHECKING
  - AWX_ISOLATED_USERNAME
  - AWX_ISOLATED_CONNECTION_TIMEOUT
  - AWX_ISOLATED_LAUNCH_TIMEOUT
  - AWX_ISOLATED_PERIODIC_CHECK
  - AWX_ISOLATED_CHECK_INTERVAL

awx/ui_next/src/screens/Host/data.hostFacts.json
```diff
@@ -112,7 +112,6 @@
       "ANSIBLE_FORCE_COLOR": "True",
       "ANSIBLE_CACHE_PLUGIN": "jsonfile",
       "AWX_PRIVATE_DATA_DIR": "/tmp/awx_13_r1ffeqze",
-      "AWX_ISOLATED_DATA_DIR": "/tmp/awx_13_r1ffeqze/artifacts/13",
```
jbradberry, line 115: This one should not be removed.

jbradberry: Never mind, something else must have removed this.

wenottingham (approved):

tiagodread (approved):
Verified that the user can save the form successfully.

Issue #9952, opened 2021-04-20
Problem with symbolic links in ansible role to run a job


aldelsa:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Recently I upgrade my awx version from 17.0 to 19.0 using awx-operator and migrating database such as is described on official docu. Everything is okay but when I run a job I had an error in container awx-task.

##### ENVIRONMENT
* AWX version: 19.0.0
* AWX install method: awx-operator in k8s
* Ansible version: 2.10
* Operating System: kubernetes running on ubuntu 18

##### STEPS TO REPRODUCE
Have a broken symbolic link in an ansible role repo and update from 17.0.0 to 19.0.0 using awx-operator

##### EXPECTED RESULTS
Jobs running successful even with a broken symbolic link like in the 17.0.0 version.

##### ACTUAL RESULTS
I am getting this error in awx-task container:
```
2021-04-20 11:35:50,723 DEBUG    [-] awx.main.commands.run_callback_receiver ProjectUpdateEvent.objects.bulk_create(4)
2021-04-20 11:35:50,899 DEBUG    [fb4b1880602c4f4881d883bc995d279d] awx.analytics.job_lifecycle job-15549 running playbook
Exception in thread Thread-2:
Traceback (most recent call last):
  File "/usr/lib64/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib64/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/tasks.py", line 3018, in transmit
    ansible_runner.interface.run(streamer='transmit', _output=_socket.makefile('wb'), **self.runner_params)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/ansible_runner/interface.py", line 243, in run
    r.run()
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/ansible_runner/streaming.py", line 53, in run
    self._output.write(utils.stream_dir(self.private_data_dir))
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/ansible_runner/utils.py", line 95, in stream_dir
    archive.write(os.path.join(dirpath, fname), arcname=os.path.join(relpath, fname))
  File "/usr/lib64/python3.8/zipfile.py", line 1740, in write
    zinfo = ZipInfo.from_file(filename, arcname,
[...]

PR #10172, opened 2021-05-11
created TransmitterThread class to spawn TransmitJobs as threads

sarabrajsingh:
##### SUMMARY

Surfaced this bug while fixing issue #9952. 

`Receptor-ctl` would hang trying to read from a socket pair that is never closed, when the `transmit` job was executed yet threw an exception. This anomaly occurred when using a k8s cluster (Openshift in our case) as the execution environment.

We moved the transmit function to execute in its own thread, and rejoin the calling function (`_run_internal`) when it finishes transmitting, and bubbling up exceptions if any occurred in the thread.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.1.0
```
##### ADDITIONAL INFORMATION

Steps to reproduce bug:

1. Create an ansible project with some files (or roles) and commit to a git repository
2. Purposely commit a broken symlink on any file in the project files to git
2. Create a credential set that authenticates to your k8s cluster in AWX
3. Create a container instance group that leverages the credentials from Step 3 in AWX
4. Create a project in AWX that leverages the ansible project/git repo from Step 1
5. Create a job template that leverages the AWX project from Step 4 and set your EE to the container instance group from Step 3
6. Execute the job from Step 5


awx/main/tasks.py
```diff
@@ -1369,7 +1370,12 @@ def run(self, pk, **kwargs):
             if isinstance(self.instance, SystemJob):
                 cwd = self.build_cwd(self.instance, private_data_dir)
                 res = ansible_runner.interface.run(
-                    project_dir=cwd, event_handler=self.event_handler, finished_callback=self.finished_callback, status_handler=self.status_handler, **params
+                    project_dir=cwd,
+                    event_handler=self.event_handler,
+                    finished_callback=self.finished_callback,
+                    status_handler=self.status_handler,
+                    debug=True,
```
shanemcd: 👀

fosterseth (approved):

AlanCoding (approved):

awx_collection/plugins/modules/tower_export.py
```diff
@@ -112,6 +112,18 @@
...
+        teams=dict(type='str'),
+        credential_types=dict(type='str'),
+        credentials=dict(type='str'),
+        execution_environments=dict(type='str'),
+        notification_templates=dict(type='str'),
+        inventory_sources=dict(type='str'),
+        inventory=dict(type='str'),
+        projects=dict(type='str'),
+        job_templates=dict(type='str'),
+        workflow_job_templates=dict(type='str'),
```
AlanCoding, line 126: this was expected as part of separate work, are you having some branch management problems? Let me know if you want some help.

PR #10175, opened 2021-05-11
Add more fields to .only since they get referenced

AlanCoding:
##### SUMMARY
Fixes a performance bottleneck when saving playbook_on_stats for jobs that use `--limit` against a small number of hosts while it has a large number in the inventory.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
'tis the problem with `.only()`, that _only_ the developer who wrote the queryset remembers that the other fields cannot be referenced except at the cost of another query

https://github.com/ansible/awx/pull/7352/files

### DIFF
--- a/awx/main/models/events.py
+++ b/awx/main/models/events.py
@@ -511,7 +511,7 @@ class JobEvent(BasePlaybookEvent):
 
             from awx.main.models import Host, JobHostSummary  # circular import
 
-            all_hosts = Host.objects.filter(pk__in=self.host_map.values()).only('id', 'name', 'last_job_id', 'last_job_host_summary_id')
+            all_hosts = Host.objects.filter(pk__in=self.host_map.values()).only('id', 'name')
             existing_host_ids = set(h.id for h in all_hosts)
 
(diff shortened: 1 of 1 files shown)


awx/main/models/events.py
```diff
@@ -511,7 +511,7 @@ def _update_host_summary_from_stats(self, hostnames):
 
             from awx.main.models import Host, JobHostSummary  # circular import
 
-            all_hosts = Host.objects.filter(pk__in=self.host_map.values()).only('id')
+            all_hosts = Host.objects.filter(pk__in=self.host_map.values()).only('id', 'name', 'last_job_id', 'last_job_host_summary_id')
```
chrismeyersfsu: Can we do without `last_job_id` and `last_job_host_summary_id` ? We only seem to be setting these fields, not reading them.

AlanCoding: yes

chrismeyersfsu (approved):

Issue #9126, opened 2021-01-21
login page ui tweaks


jlmitch5:
![Screen Shot 2021-01-20 at 12 54 28 PM](https://user-images.githubusercontent.com/1342624/105369911-0380d780-5bd1-11eb-9be8-02855c75ac1e.png)

- We should make sure logo is the white text version (not sure exactly how that branding gets added, may need to be a separate issue)
- We should move "Please sign in." to secondary text (and it should be sentence cased).  like this shows https://www.patternfly.org/v4/components/login-page/react/basic/

jakemcdermott:
Also:

"Sign in"
"Login"

(Should pick one, or at least not show both on the same page)

tiagodread:
The login page with the white brand logo 

![image](https://user-images.githubusercontent.com/12464858/119698839-06fac880-be28-11eb-8bf1-1b4bb9934fe5.png)

PR #10179, opened 2021-05-12
Update login message to avoid duplicate brand name

nixocio:
Returned brandName already contain the name brand Ansible. Update to
avoid name brand repetition.

Also update brandName on document title.

See: https://github.com/ansible/awx/issues/9126

### DIFF
--- a/awx/ui_next/src/screens/Login/Login.jsx
+++ b/awx/ui_next/src/screens/Login/Login.jsx
@@ -147,7 +147,7 @@ function AWXLogin({ alt, isAuthenticated }) {
     <Login header={Header} footer={Footer}>
       <LoginMainHeader
         data-cy="login-header"
-        title={brandName ? t`Welcome to Ansible ${brandName}!` : ''}
+        title={brandName ? t`Welcome to ${brandName}!` : ''}
         subtitle={t`Please log in`}
       />
       <LoginMainBody>
--- a/awx/ui_next/src/index.jsx
+++ b/awx/ui_next/src/index.jsx
@@ -5,7 +5,7 @@ import '@patternfly/react-core/dist/styles/base.css';
 import App from './App';
 import { BrandName } from './variables';
 
-document.title = `Ansible ${BrandName}`;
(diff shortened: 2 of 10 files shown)


awx/ui_next/src/screens/Login/Login.jsx
```diff
@@ -147,7 +147,7 @@ function AWXLogin({ alt, isAuthenticated }) {
     <Login header={Header} footer={Footer}>
       <LoginMainHeader
         data-cy="login-header"
-        title={brandName ? t`Welcome to Ansible ${brandName}!` : ''}
```
nixocio, line 150: @jakemcdermott, I also looked for other parts of the code that we are using `brandName` and the only other one similar case that I found was on the `document.title` - updated accordingly.

AlexSCorey (approved):
Looks good.  

We now have some dead strings in the `.po` files like this [one](https://github.com/ansible/awx/pull/10179/files#diff-53a6bf78d750453126d950e9fdf791cd43c9c40777d12a870310665d47990e3aR9524) but I have a meeting on the calendar with Christian to go through them to clean up all the dead strings before we push for translation services.

nixocio:
> We now have some dead strings in the .po

@AlexSCorey, what is the root cause of this issue? - I updated the strings using the `npm run  extract-strings`.

awx/ui_next/src/screens/Login/Login.jsx
jakemcdermott: sgtm

AlexSCorey:
The dead strings are related to when you update existing strings, which are now on a new line in the file.  Ling ui sees them as new string entirely instead of updated existing strings.  It's not a big deal since there is a cli command to remove them.  Also, I believe we won't have this problem very soon because it looks like ling ui is going to fix the issue that forces us to run that `extract-strings` so frequently https://github.com/lingui/js-lingui/issues/1003

keithjgrant (approved):

tiagodread (approved):

PR #10182, opened 2021-05-12
Use development container when updating dependencies

jakemcdermott:
##### GOAL
When adding, updating, or removing a package, the resultant changeset to the `package.json` and `package-lock.json` files that we check into devel should be consistent and reproducible, regardless of who is doing the update or what machine they're doing the update on.

##### SUMMARY OF CHANGES
* Even though we usually run the ui development tooling outside of the awx development container, we should standardize on using this environment to update packages to ensure some consistency.
* Use `npm ci` instead of `npm install` as the default package installation command, since we generally want to modify package files only when updating deps
* You can run into memory limits from docker or node when building in the container, so increase the node memory limit and add a note to the docs that you may need to increase the container memory limit in your docker-for-mac settings (or similar tool, if you use it)

### DIFF
--- a/awx/ui_next/README.md
+++ b/awx/ui_next/README.md
@@ -92,6 +92,9 @@ docker exec -it tools_awx_1 bash
 # add the updated package.json and package-lock.json files to scm
 git add awx/ui_next_next/package.json awx/ui_next_next/package-lock.json
 ```
+#### Note:
+- Building the ui can use up a lot of resources. If you're running docker for mac or similar
+virtualization, the default memory limit may not be enough and you should increase it.
 
 ## Building for Production
 ```shell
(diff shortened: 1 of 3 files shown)


jakemcdermott:
kicked off an e2e and visual

awx/ui_next/README.md
```diff
@@ -62,26 +62,39 @@ npm --prefix awx/ui_next run test -- --coverage
...
+
+# exit the container
+(tools_awx_1)$ exit
+
+# add the updated package.json and package-lock.json files to scm
+git add awx/ui_next_next/package.json awx/ui_next_next/package-lock.json
 ```
+#### Note:
+- Building the ui can use up a lot of resources. If you're running docker for mac or similar
+virtualization, the default memory limit may not be enough and you should increase it.
```
nixocio, line 97: 👍

nixocio (approved):

keithjgrant (approved):

tiagodread (approved):
The e2e tests broken was not related to this change, I don't expect this to cause any regression in the app

PR #10184, opened 2021-05-12
Break out the pre-commit hook into an explicit script

jbradberry:
##### SUMMARY

This means that

- we don't have to be always updating the underlying .git/hooks/pre-commit file
- updates to the logic will just work automatically
- the logic of the conditional invocation of black has been fixed so that AWX_IGNORE_BLACK=1 should work correctly now

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/bash
+
+if [ -z $AWX_IGNORE_BLACK ]
+then
+        black --check $(git diff --cached --name-only --diff-filter=AM | grep -E '\.py') || \
+        (echo 'To fix this, run `make black` to auto-format your code prior to commit, or set AWX_IGNORE_BLACK=1' && \
+        exit 1)
+fi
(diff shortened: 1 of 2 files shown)


pre-commit.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/bash
+
+if [ -z $AWX_IGNORE_BLACK ]
+then
+        black --check $(git diff --cached --name-only --diff-filter=AM | grep -E '\.py') || \
+        (echo 'To fix this, run `make black` to auto-format your code prior to commit, or set AWX_IGNORE_BLACK=1' && \
+        exit 1)
```
rebeccahhh, line 7: should we mention `git add` needs to be run after the `make black`?

AlanCoding (approved):
nice

pre-commit.sh
jbradberry: I lean towards probably not, since it is going to be explicitly run by the user and the fact that it will change files is (hopefully) understood.

Issue #10170, opened 2021-05-11
Execution Environment Image Name confusing.


gamuniz:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

The image name implies to add in the name of your image from the registery. 

##### ENVIRONMENT
* AWX version: devel
* AWX install method: awx-operator
* Ansible version:  2.11
* Operating System: RHEL8
* Web Browser: n/a

##### STEPS TO REPRODUCE
1. create registry credential
2. create execution environment with image name
3. profit

##### EXPECTED RESULTS
successfully use image from EE to run job template.

##### ACTUAL RESULTS
error pulling image

```
Error: 3 errors occurred while pulling:
 * Error initializing source docker://registry.access.redhat.com/my_first_ee_image:latest: Error reading manifest latest in registry.access.redhat.com/my_first_ee_image: name unknown: Repo not found
 * Error initializing source docker://registry.redhat.io/my_first_ee_image:latest: unable to retrieve auth token: invalid username/password: unauthorized: Please login to the Red Hat Registry using your Customer Portal credentials. Further instructions can be found here: https://access.redhat.com/RegistryAuthentication
 * Error initializing source docker://my_first_ee_image:latest: Error reading manifest latest in docker.io/library/my_first_ee_image: errors:
denied: requested access to the resource is denied
unauthorized: authentication required
```

jakemcdermott:
We're more likely to change the displayed name of the field than the function of the field. I'm thinking something like "path" or "url" might make more sense rather than name.

Thoughts? cc @gamuniz @nixocio

gamuniz:
@jakemcdermott I was thinking the same thing Image Url probably makes the most sense.

nixocio:
+1 to change the displayed name. Something like `Container Image URL`

jakemcdermott:
@gamuniz @beeankha  @nixocio 

In other places, we just call the field "Image".

![Screenshot from 2021-05-12 13-59-09](https://user-images.githubusercontent.com/9753817/118022515-705be100-b32a-11eb-8517-41a4e2ba9991.png)
![Screenshot from 2021-05-12 13-58-56](https://user-images.githubusercontent.com/9753817/118022528-7225a480-b32a-11eb-9004-d10da4b4cb2d.png)

I'm inclined to follow this convention. I'm thinking the confusion can probably be cleared up just by dropping the `name` from the input label so that is what I did. See: https://github.com/ansible/awx/pull/10185

[...]

PR #10185, opened 2021-05-12
Drop the word 'name' from image field, add examples

jakemcdermott:
for #10170 

We call this field "Image" in other places, so we want it to match. Also, "Image name" can be confusing because people may think they just need to provide the image name instead of the full image location, which includes registry and tag version.

cc @gamuniz @nixocio @beeankha @tiagodread 

Screenshot from 2021-05-12 15-39-13


nixocio (approved):
I liked the comment 

> the full image location, which includes registry and tag version.

We could add this as part of the tooltip, and API help text.

nixocio:
I triggered an e2e test run.

jakemcdermott:
@nixocio sounds good thx

beeankha (approved):

Zokormazo (approved):

PR #10191, opened 2021-05-13
close db and cache connection in new threads

chrismeyersfsu:
It's not safe to share or re-use django db connections nor django cache connections across new threads nor processes.


AlanCoding (approved):
Makes sense to me. Make sure to run tests and see that things are stable.

shanemcd (approved):

Issue #9377, opened 2021-02-22
Workflow only uses toplevel limit


mikecurrancs:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Since workflows make use of job templates, I would expect that the individual job template has the master limit for each job being run through the workflow. Instead the limit at the top level of the workflow is the only one that is being honored. If we leave that top-level blank, then the job run against every machine in our inventory instead of the ones listed only in the job template being run.

##### ENVIRONMENT
* AWX version: 17.0.1
* AWX install method: docker on linux
* Ansible version:  2.9.17
* Operating System: Ubuntu 20.04
* Web Browser: Any

##### STEPS TO REPRODUCE

1. Create a workflow 
2. Assign job templates in the visualizer
3. Set your limit in the job templates only
4. Save your workflow
##### EXPECTED RESULTS

I expect the limit in the job templates of the visualizer to be the master source for the limit, not the workflow. 

##### ACTUAL RESULTS

The job template limit is ignored, and each job template task is run against every server in the inventory

##### ADDITIONAL INFORMATION

Will provide any more information as requested.

wenottingham:
As noted in the docs, the workflow limit "provide[s] a host pattern to further constrain the list of hosts that will be managed or affected by the workflow. This limit is applied to all job template nodes that prompt for a limit."

Do your job templates allow prompting for a limit? If so, they'll use whatever limit is set on the workflow.

mikecurrancs:
This becomes problematic with our customer/server models and now we have to potentially duplicate every set of job templates to create individual workflows to ensure someone doesn't miss properly updating the limit in the job template. Defeats the idea of generic job templates that can be run against a multitude of servers just by asking for a limit. Now you have to always edit the limit and if someone runs one, thinking they did just that, they could take down the wrong customer. Automation to fix or not, this becomes a production incident and outage. 
 
The prompting of the job templates in the workflow visualizer should be the master source for the limit, not the workflow itself. If left blank, it should be ignored instead of defaulting to the entire inventory.  

This design seems like a step backwards. Especially since upgrades of our application have a mix of windows servers that are application only and sql only , and not even necessary to be run only in a workflow.

wenottingham:
> If left blank

If you look at the workflow in the API, is the limit `""`, or `null`?

If you PATCH it to be `null`, does it work like you expect?

wenottingham:
[...]

PR #10195, opened 2021-05-13
Fix null issue for limit

nixocio:
Fix null issue for limit.

See: https://github.com/ansible/awx/issues/9377


nixocio:
I triggered an e2e test run.

jakemcdermott (approved):

one-t:
I've verified this is now working as expected.

one-t:
!regate

Issue #8998, opened 2021-01-05
UI: Broken link: Clicking on "User Details" gives 404 Page Not Found


gmtime:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Clicking on "User Details" in the upper right page corner gives 404 Page Not Found.

##### ENVIRONMENT
* AWX version: 16.0.0
* AWX install method: docker on linux
* Operating System: Linux
* Web Browser: any

##### STEPS TO REPRODUCE

1. login into AWX
2. click on the user name in the top right corner
3. select User Details from the dropdown menu

##### EXPECTED RESULTS

User Details 

##### ACTUAL RESULTS

404 Page Not Found

unlikelyzero:
closing https://github.com/ansible/awx/issues/8478 as this ticket is more readily discoverable

mabashian:
Looks like we need to change this line: https://github.com/ansible/awx/blob/devel/awx/ui_next/src/components/AppContainer/PageHeaderToolbar.jsx#L93-L95

to:

```
                  href={
                    loggedInUser
                      ? `/#/users/${loggedInUser.id}/details`
                      : '/#/home'
                  }
```

@AlexSCorey

tiagodread:
Fixed.
Covered by https://github.com/ansible/tower-qa/pull/5914

layer3lv:
how to fix this error still in 17.1.0

PR #10202, opened 2021-05-14
Use non-root path links

tiagodread:
Resolves https://github.com/ansible/tower/issues/3420

Use non-root path for links


jakemcdermott (approved):

jakemcdermott:
@tiagodread lgtm. I verified on dev + prod environments and did a quick regression check on https://github.com/ansible/awx/issues/8998

PR #10203, opened 2021-05-14
Add default error boundary around screens

jakemcdermott:
##### SUMMARY
Catch any unhandled non-async errors with an error boundary around screens. This will show a generic error message instead
of crashing the page.

[The react docs decently summarize the general concept of error boundaries](https://reactjs.org/docs/error-boundaries.html). The example code is informative, but there appears to be a few more edge cases that need to be dealt with. Rather than re-implement solutions for those, I decided to pull in a [simple and well-tested library](https://github.com/bvaughn/react-error-boundary).

##### ADDITIONAL INFORMATION
**before**
error-boundary-before

**after**
error-boundary-after

### DIFF
--- a/awx/ui_next/package-lock.json
+++ b/awx/ui_next/package-lock.json
@@ -16826,6 +16826,14 @@
         "prop-types-extra": "^1.1.0"
       }
     },
+    "react-error-boundary": {
+      "version": "3.1.3",
+      "resolved": "https://registry.npmjs.org/react-error-boundary/-/react-error-boundary-3.1.3.tgz",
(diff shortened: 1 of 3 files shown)


keithjgrant (approved):
Nice!

nixocio (approved):

tiagodread:
@jakemcdermott I tried to run this locally in dev or prod build without success

image

image

jakemcdermott:
@tiagodread you'll want to run  `make clean-ui` before the reinstall

tiagodread (approved):
Tested locally without the fix for workflow approvals to force the error, and now we are not displaying the white screen and the UX is better with this PR

image

PR #10204, opened 2021-05-14
Container groups registry creds

### LINKED ISSUE #10114: Container Registry credentials don't work with Container Groups
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
This is a reopen of #9928. #9928 was closed stating #9623 addresses the issue and would be resolved in the next release. However #9623 was included in release 19.1.0 and the issue still persists.

Execution environments can't be pulled from private registries as the authentication credentials are not being passed to the pod spec.

##### ENVIRONMENT
* AWX version: 19.1.0
* AWX install method: minikube
* Ansible version:  ?
* Operating System: Debian
* Web Browser: Chrome

##### STEPS TO REPRODUCE

image

image

image

```

# kubectl describe pod awx-job-53-cvh88
Name:         awx-job-53-cvh88
Namespace:    default
Priority:     0
Node:         minikube/192.168.49.2
Start Time:   Wed, 05 May 2021 14:06:06 +0000
Labels:       ansible-awx=f34e7217-6215-41db-b716-970e1b09be7d
              ansible-awx-job-id=53
Annotations:  <none>
Status:       Pending
IP:           172.17.0.6
IPs:
  IP:  172.17.0.6
Containers:
  worker:
    Container ID:  
    Image:         gitlab.example.com:5443/myproject/ansible-runner
    Image ID:      
    Port:          <none>
    Host Port:     <none>
    Args:
      ansible-runner
      worker
      --private-data-dir=/runner
    State:          Waiting
      Reason:       ImagePullBackOff
    Ready:          False
    Restart Count:  0
    Environment:    <none>
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-qfwv8 (ro)
Conditions:
  Type              Status
  Initialized       True 
  Ready             False 
  ContainersReady   False 
  PodScheduled      True 
Volumes:
  default-token-qfwv8:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  default-token-qfwv8
    Optional:    false
QoS Class:       BestEffort
Node-Selectors:  <none>
Tolerations:     node.kubernetes.io/not-ready:NoExecute op=Exists for 300s
                 node.kubernetes.io/unreachable:NoExecute op=Exists for 300s
Events:
  Type     Reason     Age                    From               Message
  ----     ------     ----                   ----               -------
  Normal   Scheduled  44m                    default-scheduler  Successfully assigned default/awx-job-53-cvh88 to minikube
  Normal   Pulling    43m (x4 over 44m)      kubelet            Pulling image "gitlab.example.com:5443/myproject/ansible-runner"
  Warning  Failed     43m (x4 over 44m)      kubelet            Failed to pull image "gitlab.example.com:5443/myproject/ansible-runner": rpc error: code = Unknown desc = Error response from daemon: Head https://gitlab.example.com:5443/v2/myproject/ansible-runner/manifests/latest: denied: access forbidden
  Warning  Failed     43m (x4 over 44m)      kubelet            Error: ErrImagePull
  Normal   BackOff    9m25s (x155 over 44m)  kubelet            Back-off pulling image "gitlab.example.com:5443/myproject/ansible-runner"
  Warning  Failed     4m25s (x177 over 44m)  kubelet            Error: ImagePullBackOff
```

##### EXPECTED RESULTS

Image to download & run successfully

##### ACTUAL RESULTS

Image download responds with `access forbidden` because of lack of credentials.

rooftopcellist:
##### SUMMARY
Fixes https://github.com/ansible/awx/issues/10114

In VM-based installs, the user provides image pull creds to us, then we put them in an authfile and give it to podman via `--authfile`.
This is not so simple with ContainerGroups because we need to use the k8s API to apply a podspec to create containers in this paradigm.  

Currently, the EE pod gets created, but errors when pulling the custom EE in from the private repo:

image

This work will modify the __init__() for the AWXReceptorJob class to create a k8s secret in the given namespace, then specify that secret name in the pod-spec as an `imagePullSecret`

Also, the imagePullPolicy was not being enforced when running JT's in EE's using container groups, this is because the `imagePullPolicy` nevery got set on the pod spec.  

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
devel
```

##### ADDITIONAL INFORMATION

Issues that this solves:
* Image pull secret now gets created in the cluster namespace specified by the user for that container group.  
* imagePullSecret name gets set on the pod spec
* If the pull secret already exists in the namespace, delete it, then create it.  (`kube_api.replace_namespaced_secret` did not work for this case...)
* Enforce `imagePullPolicy` for EE's in container groups
* Basic error handling

### DIFF
--- a/awx/main/scheduler/kubernetes.py
+++ b/awx/main/scheduler/kubernetes.py
@@ -2,6 +2,7 @@ import collections
 import json
 import logging
 from base64 import b64encode
+from urllib import parse as urlparse
 
 from django.conf import settings
 from kubernetes import client, config
@@ -52,19 +53,26 @@ class PodManager(object):
 
         return pods
 
-    @classmethod
     def create_secret(self, job):
-        task = collections.namedtuple('Task', 'id instance_group')(id='', instance_group=job.instance_group)
-        pm = PodManager(task)
         registry_cred = job.execution_environment.credential
-        host = registry_cred.get_input('host').split('/')[0]
+        host = registry_cred.get_input('host')
+        scheme = 'https'
+        # urlparse requires '//' to be provided if scheme is not specified
+        original_parsed = urlparse.urlsplit(host)
+        if (not original_parsed.scheme and not host.startswith('//')) or original_parsed.hostname is None:
+            host = '%s://%s' % (scheme, host)
+        parsed = urlparse.urlsplit(host)
+        host = parsed.hostname
+        if parsed.port:
+            host = "{0}:{1}".format(host, parsed.port)
+
         username = registry_cred.get_input("username")
         password = registry_cred.get_input("password")
 
         # Construct container auth dict and base64 encode it
-        token = b64encode("{}:{}".format(username, password).encode('ascii')).decode()
+        token = b64encode("{}:{}".format(username, password).encode('UTF-8')).decode()
         auth_dict = json.dumps({"auths": {host: {"auth": token}}}, indent=4)
-        auth_data = b64encode(str(auth_dict).encode('ascii')).decode()
+        auth_data = b64encode(str(auth_dict).encode('UTF-8')).decode()
 
         # Construct Secret object
         secret = client.V1Secret()
@@ -77,7 +85,7 @@ class PodManager(object):
         # Check if secret already exists
         secrets = None
         try:
-            secrets = pm.kube_api.list_namespaced_secret(namespace=pm.namespace)
+            secrets = self.kube_api.list_namespaced_secret(namespace=self.namespace)
         except client.rest.ApiException:
             error_msg = 'Invalid openshift or k8s cluster credential'
             logger.exception(error_msg)
@@ -90,20 +98,21 @@ class PodManager(object):
             for s in secrets_dict:
                 if s['metadata']['name'] == secret_name:
                     secret_exists = True
+                    break
             if secret_exists:
                 try:
                     # Try to replace existing secret
-                    pm.kube_api.delete_namespaced_secret(name=secret.metadata.name, namespace=pm.namespace)
-                    pm.kube_api.create_namespaced_secret(namespace=pm.namespace, body=secret)
+                    self.kube_api.delete_namespaced_secret(name=secret.metadata.name, namespace=self.namespace)
+                    self.kube_api.create_namespaced_secret(namespace=self.namespace, body=secret)
                 except Exception:
-                    error_msg = 'Failed to create imagePullSecret for container group {}'.format(task.instance_group.name)
+                    error_msg = 'Failed to create imagePullSecret for container group {}'.format(job.instance_group.name)
                     logger.exception(error_msg)
                     job.cancel(job_explanation=error_msg)
                     raise
             else:
                 # Create an image pull secret in namespace
                 try:
-                    pm.kube_api.create_namespaced_secret(namespace=pm.namespace, body=secret)
+                    self.kube_api.create_namespaced_secret(namespace=self.namespace, body=secret)
                 except client.rest.ApiException as e:
                     if e.status == 401:
                         error_msg = 'Failed to create imagePullSecret: {}. Check that openshift or k8s credential has permission to create a secret.'.format(
@@ -112,7 +121,7 @@ class PodManager(object):
                         logger.exception(error_msg)
                         # let job run for the case that the secret exists but the cluster cred doesn't have permission to create a secret
                 except Exception:
-                    error_msg = 'Failed to create imagePullSecret for container group {}'.format(task.instance_group.name)
+                    error_msg = 'Failed to create imagePullSecret for container group {}'.format(job.instance_group.name)
                     logger.exception(error_msg)
                     job.cancel(job_explanation=error_msg)
 
--- a/awx/main/scheduler/kubernetes.py
+++ b/awx/main/scheduler/kubernetes.py
@@ -1,4 +1,5 @@
 import collections
+import json
 import logging
 from base64 import b64encode
 
@@ -50,6 +51,72 @@ class PodManager(object):
             logger.exception('Failed to list pods for container group {}'.format(instance_group))
 
         return pods
+
+    @classmethod
+    def create_secret(self, job):
+        task = collections.namedtuple('Task', 'id instance_group')(id='', instance_group=job.instance_group)
+        pm = PodManager(task)
+        registry_cred = job.execution_environment.credential
+        host = registry_cred.get_input('host').split('/')[0]
+        username = registry_cred.get_input("username")
+        password = registry_cred.get_input("password")
+
+        # Construct container auth dict and base64 encode it
+        token = b64encode("{}:{}".format(username, password).encode('ascii')).decode()
+        auth_dict = json.dumps({"auths": {host: {"auth": token}}}, indent=4)
+        auth_data = b64encode(str(auth_dict).encode('ascii')).decode()
+
+        # Construct Secret object
+        secret = client.V1Secret()
+        secret_name = "automation-{0}-image-pull-secret-{1}".format(settings.INSTALL_UUID[:5], job.execution_environment.credential.id)
+        secret.metadata = client.V1ObjectMeta(name="{}".format(secret_name))
+        secret.type = "kubernetes.io/dockerconfigjson"
+        secret.kind = "Secret"
+        secret.data = {".dockerconfigjson": auth_data}
+
+        # Check if secret already exists
+        secrets = None
+        try:
+            secrets = pm.kube_api.list_namespaced_secret(namespace=pm.namespace)
+        except client.rest.ApiException:
+            error_msg = 'Invalid openshift or k8s cluster credential'
+            logger.exception(error_msg)
+            job.cancel(job_explanation=error_msg)
+            raise
+
+        if secrets:
+            secret_exists = False
+            secrets_dict = secrets.to_dict().get('items', [])
+            for s in secrets_dict:
+                if s['metadata']['name'] == secret_name:
+                    secret_exists = True
+            if secret_exists:
+                try:
+                    # Try to replace existing secret
+                    pm.kube_api.delete_namespaced_secret(name=secret.metadata.name, namespace=pm.namespace)
+                    pm.kube_api.create_namespaced_secret(namespace=pm.namespace, body=secret)
+                except Exception:
+                    error_msg = 'Failed to create imagePullSecret for container group {}'.format(task.instance_group.name)
+                    logger.exception(error_msg)
+                    job.cancel(job_explanation=error_msg)
+                    raise
+            else:
+                # Create an image pull secret in namespace
+                try:
+                    pm.kube_api.create_namespaced_secret(namespace=pm.namespace, body=secret)
+                except client.rest.ApiException as e:
+                    if e.status == 401:
+                        error_msg = 'Failed to create imagePullSecret: {}. Check that openshift or k8s credential has permission to create a secret.'.format(
+                            e.status
+                        )
+                        logger.exception(error_msg)
+                        # let job run for the case that the secret exists but the cluster cred doesn't have permission to create a secret
+                except Exception:
+                    error_msg = 'Failed to create imagePullSecret for container group {}'.format(task.instance_group.name)
+                    logger.exception(error_msg)
+                    job.cancel(job_explanation=error_msg)
+
+        return secret.metadata.name
 
     @property
     def namespace(self):
--- a/awx/main/scheduler/kubernetes.py
+++ b/awx/main/scheduler/kubernetes.py
@@ -83,47 +83,68 @@ class PodManager(object):
         secret.data = {".dockerconfigjson": auth_data}
 
         # Check if secret already exists
-        secrets = None
+        replace_secret = False
         try:
-            secrets = self.kube_api.list_namespaced_secret(namespace=self.namespace)
-        except client.rest.ApiException:
-            error_msg = 'Invalid openshift or k8s cluster credential'
-            logger.exception(error_msg)
-            job.cancel(job_explanation=error_msg)
-            raise
-
-        if secrets:
-            secret_exists = False
-            secrets_dict = secrets.to_dict().get('items', [])
-            for s in secrets_dict:
-                if s['metadata']['name'] == secret_name:
-                    secret_exists = True
-                    break
-            if secret_exists:
-                try:
-                    # Try to replace existing secret
-                    self.kube_api.delete_namespaced_secret(name=secret.metadata.name, namespace=self.namespace)
-                    self.kube_api.create_namespaced_secret(namespace=self.namespace, body=secret)
-                except Exception:
-                    error_msg = 'Failed to create imagePullSecret for container group {}'.format(job.instance_group.name)
-                    logger.exception(error_msg)
-                    job.cancel(job_explanation=error_msg)
-                    raise
+            existing_secret = self.kube_api.read_namespaced_secret(namespace=self.namespace, name=secret_name)
+            if existing_secret.data != secret.data:
+                replace_secret = True
+            secret_exists = True
+        except client.rest.ApiException as e:
+            if e.status == 404:
+                secret_exists = False
             else:
-                # Create an image pull secret in namespace
-                try:
-                    self.kube_api.create_namespaced_secret(namespace=self.namespace, body=secret)
-                except client.rest.ApiException as e:
-                    if e.status == 401:
-                        error_msg = 'Failed to create imagePullSecret: {}. Check that openshift or k8s credential has permission to create a secret.'.format(
-                            e.status
+                error_msg = _('Invalid openshift or k8s cluster credential')
+                if e.status == 403:
+                    error_msg = _(
+                        'Failed to create secret for container group {} because the needed service account roles are needed.  Add get, list, create and delete roles for secret resources for your cluster credential.'.format(
+                            job.instance_group.name
                         )
-                        logger.exception(error_msg)
-                        # let job run for the case that the secret exists but the cluster cred doesn't have permission to create a secret
-                except Exception:
-                    error_msg = 'Failed to create imagePullSecret for container group {}'.format(job.instance_group.name)
+                    )
+                full_error_msg = '{0}: {1}'.format(error_msg, str(e))
+                logger.exception(full_error_msg)
+                job.job_explanation = error_msg
+                job.save()
+                raise PermissionError(full_error_msg)
+
+        if replace_secret:
+            try:
+                # Try to replace existing secret
+                self.kube_api.delete_namespaced_secret(name=secret.metadata.name, namespace=self.namespace)
+                self.kube_api.create_namespaced_secret(namespace=self.namespace, body=secret)
+            except client.rest.ApiException as e:
+                error_msg = _('Invalid openshift or k8s cluster credential')
+                if e.status == 403:
+                    error_msg = _(
+                        'Failed to delete secret for container group {} because the needed service account roles are needed.  Add create and delete roles for secret resources for your cluster credential.'.format(
+                            job.instance_group.name
+                        )
+                    )
+                full_error_msg = '{0}: {1}'.format(error_msg, str(e))
+                logger.exception(full_error_msg)
+                job.job_explanation = error_msg
+                job.save()
+                # let job continue for the case where secret was created manually and cluster cred doesn't have permission to create a secret
+            except Exception as e:
+                error_msg = 'Failed to create imagePullSecret for container group {}'.format(job.instance_group.name)
+                logger.exception('{0}: {1}'.format(error_msg, str(e)))
+                raise RuntimeError(error_msg)
+        elif secret_exists and not replace_secret:
+            pass
+        else:
+            # Create an image pull secret in namespace
+            try:
+                self.kube_api.create_namespaced_secret(namespace=self.namespace, body=secret)
+            except client.rest.ApiException as e:
+                if e.status == 403:
+                    error_msg = _(
+                        'Failed to create imagePullSecret: {}. Check that openshift or k8s credential has permission to create a secret.'.format(e.status)
+                    )
                     logger.exception(error_msg)
-                    job.cancel(job_explanation=error_msg)
+                    # let job continue for the case where secret was created manually and cluster cred doesn't have permission to create a secret
+            except Exception:
+                error_msg = 'Failed to create imagePullSecret for container group {}'.format(job.instance_group.name)
+                logger.exception(error_msg)
+                job.cancel(job_explanation=error_msg)
 
         return secret.metadata.name
 
--- a/awx/main/scheduler/kubernetes.py
+++ b/awx/main/scheduler/kubernetes.py
@@ -7,6 +7,7 @@ from urllib import parse as urlparse
 from django.conf import settings
 from kubernetes import client, config
 from django.utils.functional import cached_property
+from django.utils.translation import ugettext_lazy as _
 
 from awx.main.utils.common import parse_yaml_or_json
 from awx.main.utils.execution_environments import get_default_pod_spec
@@ -56,11 +57,10 @@ class PodManager(object):
     def create_secret(self, job):
         registry_cred = job.execution_environment.credential
         host = registry_cred.get_input('host')
-        scheme = 'https'
         # urlparse requires '//' to be provided if scheme is not specified
         original_parsed = urlparse.urlsplit(host)
         if (not original_parsed.scheme and not host.startswith('//')) or original_parsed.hostname is None:
-            host = '%s://%s' % (scheme, host)
+            host = 'https://%s' % (host)
         parsed = urlparse.urlsplit(host)
         host = parsed.hostname
         if parsed.port:
@@ -102,8 +102,6 @@ class PodManager(object):
                     )
                 full_error_msg = '{0}: {1}'.format(error_msg, str(e))
                 logger.exception(full_error_msg)
-                job.job_explanation = error_msg
-                job.save()
                 raise PermissionError(full_error_msg)
 
         if replace_secret:
@@ -121,8 +119,6 @@ class PodManager(object):
                     )
                 full_error_msg = '{0}: {1}'.format(error_msg, str(e))
                 logger.exception(full_error_msg)
-                job.job_explanation = error_msg
-                job.save()
                 # let job continue for the case where secret was created manually and cluster cred doesn't have permission to create a secret
             except Exception as e:
                 error_msg = 'Failed to create imagePullSecret for container group {}'.format(job.instance_group.name)
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -3077,7 +3077,20 @@ class AWXReceptorJob:
         # Enforce EE Pull Policy
         pull_options = {"always": "Always", "missing": "IfNotPresent", "never": "Never"}
         if self.task and self.task.instance.execution_environment:
-            pod_spec['spec']['containers'][0]['imagePullPolicy'] = pull_options[self.task.instance.execution_environment.pull]
+            if self.task.instance.execution_environment.pull:
+                pod_spec['spec']['containers'][0]['imagePullPolicy'] = pull_options[self.task.instance.execution_environment.pull]
+
+        if self.task and self.task.instance.is_container_group_task:
+            # If EE credential is passed, create an imagePullSecret
+            if self.task.instance.execution_environment and self.task.instance.execution_environment.credential:
+                # Create pull secret in k8s cluster based on ee cred
+                from awx.main.scheduler.kubernetes import PodManager  # prevent circular import
+
+                pm = PodManager(self.task.instance)
+                secret_name = pm.create_secret(job=self.task.instance)
+
+                # Inject secret name into podspec
+                pod_spec['spec']['imagePullSecrets'] = [{"name": secret_name}]
 
         if self.task:
             pod_spec['metadata'] = deepmerge(
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -842,7 +842,7 @@ class BaseTask(object):
                     username = cred.get_input('username')
                     password = cred.get_input('password')
                     token = "{}:{}".format(username, password)
-                    auth_data = {'auths': {host: {'auth': b64encode(token.encode('ascii')).decode()}}}
+                    auth_data = {'auths': {host: {'auth': b64encode(token.encode('UTF-8')).decode('UTF-8')}}}
                     authfile.write(json.dumps(auth_data, indent=4))
                 params["container_options"].append(f'--authfile={authfile.name}')
             else:
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -3074,6 +3074,11 @@ class AWXReceptorJob:
         pod_spec['spec']['containers'][0]['image'] = ee.image
         pod_spec['spec']['containers'][0]['args'] = ['ansible-runner', 'worker', '--private-data-dir=/runner']
 
+        # Enforce EE Pull Policy
+        pull_options = {"always": "Always", "missing": "IfNotPresent", "never": "Never"}
+        if self.task and self.task.instance.execution_environment:
+            pod_spec['spec']['containers'][0]['imagePullPolicy'] = pull_options[self.task.instance.execution_environment.pull]
+
         if self.task:
             pod_spec['metadata'] = deepmerge(
                 pod_spec.get('metadata', {}),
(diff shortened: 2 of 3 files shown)


awx/main/scheduler/kubernetes.py
```diff
@@ -51,6 +53,44 @@ def list_active_jobs(self, instance_group):
...
+                logger.exception('Failed to create imagePullSecret, check openshift or k8s credential')
+            elif e.status == 409:
+                try:
+                    # Try to replace existing secret
+                    pm.kube_api.delete_namespaced_secret(name=secret.metadata.name, namespace=pm.namespace)
+                    pm.kube_api.create_namespaced_secret(namespace=pm.namespace, body=secret)
+                except Exception as e:
+                    logger.exception('Failed to replace imagePullSecret for container group {}'.format(instance_group))
+        except Exception as e:
+            print("here in exception")
```
AlanCoding: can lose this line

awx/main/tasks.py
```diff
@@ -2895,6 +2895,13 @@ def __init__(self, task=None, runner_params=None):
         if self.task and not self.task.instance.is_container_group_task:
             execution_environment_params = self.task.build_execution_environment_params(self.task.instance, runner_params['private_data_dir'])
             self.runner_params['settings'].update(execution_environment_params)
+        elif self.task and self.task.instance.is_container_group_task:
+            pass
+            # look to see if there is a credential on the EE
+            # use the kubernetes python lib to create  an image pull secret
+            # inject `imagePullSecret` into pod definition.  maybe do this inside the pod_definition(),
+            #   or in a function called by pod_definition, which will already only be called when.
+            #   will be a list of diction
```
AlanCoding: Is this still a work in progress, or is it accomplished later in the diff here?

rooftopcellist: That can be removed now, I'll do that now, thanks.

rooftopcellist:
While I had this set up, I also tested out 
* specifying a different namespace in the podspec, which worked no problem.  :heavy_check_mark: 
* checking `Verify SSL` on the Quay credential and the OCP4 cluster :heavy_check_mark:

awx/main/scheduler/kubernetes.py
```diff
@@ -51,6 +52,43 @@ def list_active_jobs(self, instance_group):
 
         return pods
 
+    @classmethod
+    def create_secret(self, instance_group, credential):
```
AlanCoding: Specifically... this has to be a registry credential?

awx/main/scheduler/kubernetes.py
```diff
@@ -51,6 +52,43 @@ def list_active_jobs(self, instance_group):
...
+                logger.exception('Failed to create imagePullSecret, check openshift or k8s credential')
+            elif e.status == 409:
+                try:
+                    # Try to replace existing secret
+                    pm.kube_api.delete_namespaced_secret(name=secret.metadata.name, namespace=pm.namespace)
+                    pm.kube_api.create_namespaced_secret(namespace=pm.namespace, body=secret)
+                except Exception:
+                    logger.exception('Failed to replace imagePullSecret for container group {}'.format(instance_group))
+        except Exception:
+            logger.exception('Failed to create imagePullSecret for container group {}'.format(instance_group))
```
AlanCoding: shouldn't we fail the job in these error cases?

It failed to set up the registry pull secrets. That should lead to job failure in any sane situation.

awx/main/scheduler/kubernetes.py
```diff
@@ -51,6 +52,43 @@ def list_active_jobs(self, instance_group):
...
+        secret.metadata = client.V1ObjectMeta(name="{}-image-pull-secret".format(task.instance_group.name.lower()))
+        secret.type = "kubernetes.io/dockerconfigjson"
+        secret.kind = "Secret"
+        secret.data = {".dockerconfigjson": auth_data}
+
+        # Create pullImageSecret in namespace, delete and create if it already exists
+        try:
+            pm.kube_api.create_namespaced_secret(namespace=pm.namespace, body=secret)
+        except client.rest.ApiException as e:
+            if e.status == 401:
```
jbradberry: How likely is it to happen that the status isn't 401 or 409?  Should we have some kind of last-ditch catch here to handle it?

awx/main/scheduler/kubernetes.py
AlanCoding: Yeah, it might be good to either docstring that info, or give it some more specific naming. This is really only for registry credentials, and that's not natural to follow with how you've named things so far.

AlanCoding:
:+1: this looks like really good work, all my comments are pretty minor

nice work closing the feature gaps on this.

rooftopcellist:
Thanks, it looks like I also need to add some safeguards on the secret name (which is based on the name of the container group):

image

That's not a valid secret name as far as k8s is concerned, so the secret never gets created.

rooftopcellist:
I think [this Issue](https://github.com/ansible/awx/issues/7066) can probably also be closed after this PR merges.  

Work to be done:
~- [ ] Replace all non-alphanumeric characters from secret name (execpt `-` and `.`).~
- [x] Try to find a good way to surface the job_explanation as an error to the UI if the job is canceled.  
- [x] Re-arrange this to check if the secret exists first so that we can better catch the two 401 scenarios possible depending on the access the cluster cred has.  
- [x] change the name of the secret to be created to be based on the Install ID and the registry cred ID.  For example: `awx-00000-image-pull-secret-cred-3`

rooftopcellist:
There are now 4 cases to consider here:
1. Check if secret exists.  If 401, cancel job and raise error (invalid cluster cred)
2. Check if secret exists.  If not, create it
3. Check if secret exists.  If yes, delete it, get 401 --> then continue and try to run the job (cluster cred without access to create secret)
4. Check if secret exists.  If yes, delete and re-create secret

Secrets created by AWX will now have a naming scheme similar to the following:
```
>>> secret_name
'awx-51661-image-pull-secret-cred-3'
```

We might want to consider taking out `awx` and replacing it with `automation`, or something like that.

Zokormazo:
> Secrets created by AWX will now have a naming scheme similar to the following:
> 
> ```
> >>> secret_name
> 'awx-51661-image-pull-secret-cred-3'
> ```
> 
clusterId + credId right ? This should give as the required immutability and uniqueness even with two instances on the same namespace. Works for me.

> We might want to consider taking out `awx` and replacing it with `automation`, or something like that.

`automation` for consistency: https://github.com/ansible/awx/pull/10100

rooftopcellist:
@Zokormazo yep, specifically the `INSTALL_UUID` and the ID for the registry credential.  

Secrets are now named with the `automation` prefix:

```
$ oc -n default get secrets
NAME                                                   TYPE                                  DATA   AGE
automation-51661-image-pull-secret-cred-3              kubernetes.io/dockerconfigjson        1      34s
```

awx/main/scheduler/kubernetes.py
```diff
@@ -51,6 +52,72 @@ def list_active_jobs(self, instance_group):
...
+        password = registry_cred.get_input("password")
+
+        # Construct container auth dict and base64 encode it
+        token = b64encode("{}:{}".format(username, password).encode('ascii')).decode()
+        auth_dict = json.dumps({"auths": {host: {"auth": token}}}, indent=4)
+        auth_data = b64encode(str(auth_dict).encode('ascii')).decode()
+
+        # Construct Secret object
+        secret = client.V1Secret()
+        secret_name = "automation-{0}-image-pull-secret-cred-{1}".format(settings.INSTALL_UUID[:5], job.execution_environment.credential.id)
```
shanemcd: Can you drop the `cred` part? I don't think it's necessary.

awx/main/scheduler/kubernetes.py
```diff
@@ -51,6 +52,72 @@ def list_active_jobs(self, instance_group):
...
+        password = registry_cred.get_input("password")
+
+        # Construct container auth dict and base64 encode it
+        token = b64encode("{}:{}".format(username, password).encode('ascii')).decode()
+        auth_dict = json.dumps({"auths": {host: {"auth": token}}}, indent=4)
+        auth_data = b64encode(str(auth_dict).encode('ascii')).decode()
+
+        # Construct Secret object
+        secret = client.V1Secret()
+        secret_name = "automation-{0}-image-pull-secret-{1}".format(settings.INSTALL_UUID[:5], job.execution_environment.credential.id)
```
chrismeyersfsu, line 79: :+1: If I read the RFC right I think using the first 5 bytes is right piece of the UUID because it's the timestamp part.

 https://datatracker.ietf.org/doc/html/rfc4122#section-4.1.2

awx/main/tasks.py
```diff
@@ -3076,6 +3076,24 @@ def pod_definition(self):
...
+                pod_spec['spec']['containers'][0]['imagePullPolicy'] = pull_options[self.task.instance.execution_environment.pull]
+
+        if self.task and self.task.instance.is_container_group_task:
+            # If EE credential is passed, create an imagePullSecret
+            if self.task.instance.execution_environment and self.task.instance.execution_environment.credential:
+                # Create pull secret in k8s cluster based on ee cred
+                from awx.main.scheduler.kubernetes import PodManager  # prevent circular import
+
+                pm = PodManager(self.task.instance)
+                secret_name = pm.create_secret(job=self.task.instance)
```
chrismeyersfsu, line 3090: Does an execution environment require a credential or it optional? If it's optional then we need to sprinkle in some if's.

awx/main/scheduler/kubernetes.py
```diff
@@ -51,6 +52,72 @@ def list_active_jobs(self, instance_group):
...
+    def create_secret(self, job):
+        task = collections.namedtuple('Task', 'id instance_group')(id='', instance_group=job.instance_group)
+        pm = PodManager(task)
+        registry_cred = job.execution_environment.credential
+        host = registry_cred.get_input('host').split('/')[0]
+        username = registry_cred.get_input("username")
+        password = registry_cred.get_input("password")
+
+        # Construct container auth dict and base64 encode it
+        token = b64encode("{}:{}".format(username, password).encode('ascii')).decode()
```
chrismeyersfsu:
```
>>> "hello_🐉".encode('ascii')
UnicodeEncodeError: 'ascii' codec can't encode character '\U0001f409' in position 6: ordinal not in range(128)
```
I don't think we want to `.encode('ascii')`. We wouldn't support unicode characters in the username/password then.

awx/main/tasks.py
AlanCoding: this is already inside of `if self.task.instance.execution_environment and self.task.instance.execution_environment.credential`, so that looks fine to me

awx/main/scheduler/kubernetes.py
```diff
@@ -51,6 +52,72 @@ def list_active_jobs(self, instance_group):
 
         return pods
 
+    @classmethod
+    def create_secret(self, job):
+        task = collections.namedtuple('Task', 'id instance_group')(id='', instance_group=job.instance_group)
+        pm = PodManager(task)
+        registry_cred = job.execution_environment.credential
+        host = registry_cred.get_input('host').split('/')[0]
```
chrismeyersfsu: Wouldn't someone commonly enter `http://my_host/` for the `host` field of the credential ? The split would return `http:` in this case.

awx/main/scheduler/kubernetes.py
```diff
@@ -51,6 +52,72 @@ def list_active_jobs(self, instance_group):
 
         return pods
 
+    @classmethod
+    def create_secret(self, job):
+        task = collections.namedtuple('Task', 'id instance_group')(id='', instance_group=job.instance_group)
+        pm = PodManager(task)
```
AlanCoding: why would you create `pm` in this hacky way as opposed to just using `self`?

rooftopcellist: Good point, I was modeling it off of [this](https://github.com/ansible/awx/blob/devel/awx/main/scheduler/kubernetes.py#L37-L39).  I don't see a good reason for it though, I'll update both spots.

awx/main/scheduler/kubernetes.py
rooftopcellist: True, I'll use urlparse to just grab the host.

awx/main/scheduler/kubernetes.py
rooftopcellist: :thinking: we'll probably want to change this for when we lay down the docker auth file too then.  https://github.com/ansible/awx/blob/devel/awx/main/tasks.py#L845

awx/main/scheduler/kubernetes.py
shanemcd: I think was was so that we could get a reference to the Kubernetes API object that is only instantiated inside of PodManager.

awx/main/scheduler/kubernetes.py
```diff
@@ -51,6 +52,72 @@ def list_active_jobs(self, instance_group):
...
+        secret.type = "kubernetes.io/dockerconfigjson"
+        secret.kind = "Secret"
+        secret.data = {".dockerconfigjson": auth_data}
+
+        # Check if secret already exists
+        secrets = None
+        try:
+            secrets = pm.kube_api.list_namespaced_secret(namespace=pm.namespace)
+        except client.rest.ApiException:
+            error_msg = 'Invalid openshift or k8s cluster credential'
```
chrismeyersfsu: Any time we interact with a 3rd party API I think. "How can this thing fail?". Could have a DNS error, timeout, partial response, etc. So two things (1) I'm not sure how the library we are using surfaces those failures, are they all `client.rest.ApiException`? (2) It would be REALLY nice to tell users more about the failure. I do see that we have a logger.exception, that is helpful but requires digging in the logs when there is a failure. It's a balancing act though, it's tempting to just toss the entire exception into `job_explanation` but the response maybe contains sensitive information (I doubt it does, but maybe?). We should look at a few failure scenarios and see if we can do something nice for the user. For example https://github.com/ansible/awx/blob/devel/awx/api/views/__init__.py#L1677

awx/main/scheduler/kubernetes.py
chrismeyersfsu: Get the port too

awx/main/scheduler/kubernetes.py
```diff
@@ -51,6 +52,72 @@ def list_active_jobs(self, instance_group):
...
+            secrets_dict = secrets.to_dict().get('items', [])
+            for s in secrets_dict:
+                if s['metadata']['name'] == secret_name:
+                    secret_exists = True
+            if secret_exists:
+                try:
+                    # Try to replace existing secret
+                    pm.kube_api.delete_namespaced_secret(name=secret.metadata.name, namespace=pm.namespace)
+                    pm.kube_api.create_namespaced_secret(namespace=pm.namespace, body=secret)
+                except Exception:
```
chrismeyersfsu: :+1: A lot can go wrong, capturing all Exception's here is good.

awx/main/scheduler/kubernetes.py
rooftopcellist: @Zokormazo can we add an integration test for this?

awx/main/scheduler/kubernetes.py
```diff
@@ -51,6 +52,72 @@ def list_active_jobs(self, instance_group):
...
+                    logger.exception(error_msg)
+                    job.cancel(job_explanation=error_msg)
+                    raise
+            else:
+                # Create an image pull secret in namespace
+                try:
+                    pm.kube_api.create_namespaced_secret(namespace=pm.namespace, body=secret)
+                except client.rest.ApiException as e:
+                    if e.status == 401:
+                        error_msg = 'Failed to create imagePullSecret: {}. Check that openshift or k8s credential has permission to create a secret.'.format(
```
chrismeyersfsu: This is nice. This is what we want to do for our users.

awx/main/scheduler/kubernetes.py
```diff
@@ -51,6 +52,72 @@ def list_active_jobs(self, instance_group):
...
+            logger.exception(error_msg)
+            job.cancel(job_explanation=error_msg)
+            raise
+
+        if secrets:
+            secret_exists = False
+            secrets_dict = secrets.to_dict().get('items', [])
+            for s in secrets_dict:
+                if s['metadata']['name'] == secret_name:
+                    secret_exists = True
```
chrismeyersfsu: I'm being picky, I like to put a `break` here. You have found the thing you are looking for, short-circuit the loop.

awx/main/scheduler/kubernetes.py
rooftopcellist: @AlanCoding I chatted with @shanemcd about this, the reason for this approach is because without an instantiated PodManager object, the `kube_api` cached property won't be created, which is needed for authing to the cluster.  

We could re-organize this so that the kube_api auth has it's own separate class, which gets instantiated in the init of PodManager, or something along those lines.  But I'd want to tackle that in a separate PR.

shanemcd: @rooftopcellist I see what Alan was saying now. There's no reason for also making this a `@classmethod`. Ditch that, and just instantiate `PodManager` before you call this method in tasks.py

awx/main/scheduler/kubernetes.py
AlanCoding: You could stringify the exception and save it to `job_explanation`

but if you mean to present a traceback to the user, `result_traceback` would be the most appropriate (the UI needs to surface it, but they will). That might be accomplished by the `raise` already.

awx/main/scheduler/kubernetes.py
```diff
@@ -51,6 +53,78 @@ def list_active_jobs(self, instance_group):
...
         return pods
 
+    def create_secret(self, job):
+        registry_cred = job.execution_environment.credential
+        host = registry_cred.get_input('host')
+        scheme = 'https'
+        # urlparse requires '//' to be provided if scheme is not specified
+        original_parsed = urlparse.urlsplit(host)
+        if (not original_parsed.scheme and not host.startswith('//')) or original_parsed.hostname is None:
+            host = '%s://%s' % (scheme, host) if scheme else '//%s' % host
```
AlanCoding:
```python
if scheme else '//%s' % host
```

does not make sense to me, because `scheme` is statically defined. Perhaps you meant `original_parsed.scheme`? But I'm not sure.

rooftopcellist: @AlanCoding You are right, that last bit shouldn't be needed.  I'll update this [here](https://github.com/ansible/awx/blob/devel/awx/main/utils/external_logging.py#L54) too.  

Explanation of why this bit is here in the first place: the issue is that if `https://` is not in the URL, `urllib.parse` will not correctly grab the hostname (netloc).  It won't set `scheme`, and sees the host as post of the path:

```
In [1]: from urllib import parse as urlparse
   ...: 
   ...: host = 'quay.io/ansible/awx-ee'
   ...: scheme = 'https'
   ...: # urlparse requires '//' to be provided if scheme is not specified
   ...: original_parsed = urlparse.urlsplit(host)

In [2]: original_parsed
Out[2]: SplitResult(scheme='', netloc='', path='quay.io/ansible/awx-ee', query='', fragment='')
```

-----------------
This is now fixed.

awx/main/scheduler/kubernetes.py
rooftopcellist: @chrismeyersfsu we show job_explanation in the UI now as part of the job status, so we probably don't want a super long traceback there.  

image

awx/main/scheduler/kubernetes.py
```diff
@@ -51,6 +53,81 @@ def list_active_jobs(self, instance_group):
...
         return pods
 
+    def create_secret(self, job):
+        registry_cred = job.execution_environment.credential
+        host = registry_cred.get_input('host')
+        scheme = 'https'
+        # urlparse requires '//' to be provided if scheme is not specified
+        original_parsed = urlparse.urlsplit(host)
+        if (not original_parsed.scheme and not host.startswith('//')) or original_parsed.hostname is None:
+            host = '%s://%s' % (scheme, host)
```
AlanCoding: could you remove `scheme` and just do

```python
host = 'https://%s' % host
```

It feels distracting to have a static variable definition, which then gets used once.

Zokormazo:
I'm playing a bit with this with a k8s service account that does not have access to secrets.
After running a job (it should fail) I have this:
```
    "status": "canceled",
    "job_explanation": "Invalid openshift or k8s cluster credential",
    "execution_node": "",
    "controller_node": "localhost",
    "result_traceback": "Invalid openshift or k8s cluster credential: (403)\nReason: Forbidden\nHTTP response headers: HTTPHeaderDict({'Audit-Id': 'c5a1162e-cbe4-43da-8d66-b2c694bba864', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'X-Content-Type-Options': 'nosniff', 'Date': 'Tue, 25 May 2021 11:02:46 GMT', 'Content-Length': '315'})\nHTTP response body: {\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"secrets is forbidden: User \\\"system:serviceaccount:y1aro06u6g:serviceaccount-y1aro06u6g\\\" cannot list resource \\\"secrets\\\" in API group \\\"\\\" in the namespace \\\"y1aro06u6g\\\"\",\"reason\":\"Forbidden\",\"details\":{\"kind\":\"secrets\"},\"code\":403}\n\n",
    "event_processing_finished": true,
```

`job_explanation` is wrong. My openshift credential is valid and enough to run jobs with public ees. We need a better explanation of what's happening.
I don't like `canceled` as status. This was not canceled, it's an error due to lack of permissions. `failed` ?

Zokormazo:
Now I'm playing with a serviceaccount with all verbs granted for `secrets`

I can see a `automation-e5c74-image-pull-secret-102` secret on it, seems sane, but the job failed:
```
(Pdb) job
{
    "id": 85,
    "type": "job",
    "url": "/api/v2/jobs/85/",
    "related": {
        "created_by": "/api/v2/users/1/",
        "labels": "/api/v2/jobs/85/labels/",
        "inventory": "/api/v2/inventories/37/",
        "project": "/api/v2/projects/75/",
        "organization": "/api/v2/organizations/134/",
        "credentials": "/api/v2/jobs/85/credentials/",
        "unified_job_template": "/api/v2/job_templates/76/",
        "stdout": "/api/v2/jobs/85/stdout/",
        "execution_environment": "/api/v2/execution_environments/32/",
        "job_events": "/api/v2/jobs/85/job_events/",
        "job_host_summaries": "/api/v2/jobs/85/job_host_summaries/",
        "activity_stream": "/api/v2/jobs/85/activity_stream/",
        "notifications": "/api/v2/jobs/85/notifications/",
        "create_schedule": "/api/v2/jobs/85/create_schedule/",
        "job_template": "/api/v2/job_templates/76/",
        "cancel": "/api/v2/jobs/85/cancel/",
        "relaunch": "/api/v2/jobs/85/relaunch/"
    },
    "summary_fields": {
        "organization": {
            "id": 134,
            "name": "Organization - EventLeg\u35fb",
            "description": "TennisStudioMatterPositiveLevelGirlCancerShopScrewOutcome\ufffd"
        },
        "inventory": {
            "id": 37,
            "name": "Inventory - EaseFish\u0dd9",
            "description": "GeneralLinkLuckDefinitionMomStorePartnerKeyIncidentClaim\ubc84",
            "has_active_failures": false,
            "total_hosts": 1,
            "hosts_with_active_failures": 0,
            "total_groups": 0,
            "has_inventory_sources": false,
            "total_inventory_sources": 0,
            "inventory_sources_with_failures": 0,
            "organization_id": 134,
            "kind": ""
        },
        "execution_environment": {
            "id": 32,
            "name": "EE - RegretSource\ufffd",
            "description": "",
            "image": "private.registry/awx-ee"
        },
        "project": {
            "id": 75,
            "name": "Project - RelationStomach\ufdb9",
            "description": "WayEscapeTimeDriverBeachReplyTargetCharacterSalaryObjective\ufffd",
            "status": "successful",
            "scm_type": "git",
            "allow_override": false
        },
        "job_template": {
            "id": 76,
            "name": "JobTemplate - ClimateSpecial\ufffd",
            "description": "NaturalBuildingSelectionPoolMiddleClassMuscleCharacterNorthLaw\u46b2"
        },
        "unified_job_template": {
            "id": 76,
            "name": "JobTemplate - ClimateSpecial\ufffd",
            "description": "NaturalBuildingSelectionPoolMiddleClassMuscleCharacterNorthLaw\u46b2",
            "unified_job_type": "job"
        },
        "instance_group": {
            "id": 35,
            "name": "ContainerGroup - fscope-upa7jpq1z5",
            "is_container_group": true
        },
        "created_by": {
            "id": 1,
            "username": "admin",
            "first_name": "",
            "last_name": ""
        },
        "user_capabilities": {
            "delete": true,
            "start": true
        },
        "labels": {
            "count": 0,
            "results": []
        },
        "credentials": [
            {
                "id": 103,
                "name": "Credential - TieExperience\u284b",
                "description": "DepressionMindWalkBedroomGovernmentTrickTourQualityFacePhysical\ufffd",
                "kind": "ssh",
                "cloud": false
            }
        ]
    },
    "created": "2021-05-25T13:32:19.934025Z",
    "modified": "2021-05-25T13:32:20.194858Z",
    "name": "JobTemplate - ClimateSpecial\ufffd",
    "description": "NaturalBuildingSelectionPoolMiddleClassMuscleCharacterNorthLaw\u46b2",
    "job_type": "run",
    "inventory": 37,
    "project": 75,
    "playbook": "ping.yml",
    "scm_branch": "",
    "forks": 0,
    "limit": "",
    "verbosity": 0,
    "extra_vars": "{}",
    "job_tags": "",
    "force_handlers": false,
    "skip_tags": "",
    "start_at_task": "",
    "timeout": 0,
    "use_fact_cache": false,
    "organization": 134,
    "unified_job_template": 76,
    "launch_type": "manual",
    "status": "error",
    "execution_environment": 32,
    "failed": true,
    "started": "2021-05-25T13:32:20.273485Z",
    "finished": "2021-05-25T13:32:24.021624Z",
    "canceled_on": null,
    "elapsed": 3.748,
    "job_args": "",
    "job_cwd": "",
    "job_env": {
        "RECEPTOR_UNIT_ID": "jTYCf9zQ"
    },
    "job_explanation": "",
    "execution_node": "",
    "controller_node": "localhost",
    "result_traceback": "Traceback (most recent call last):\n  File \"/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/tasks.py\", line 1382, in run\n    res = receptor_job.run()\n  File \"/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/tasks.py\", line 2924, in run\n    return self._run_internal(receptor_ctl)\n  File \"/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/tasks.py\", line 2987, in _run_internal\n    raise RuntimeError(detail)\nRuntimeError: Pod Running\n",
    "event_processing_finished": true,
    "launched_by": {
        "id": 1,
        "name": "admin",
        "type": "user",
        "url": "/api/v2/users/1/"
    },
    "job_template": 76,
    "passwords_needed_to_start": [],
    "allow_simultaneous": false,
    "artifacts": {},
    "scm_revision": "5df2d0f34c1c5bd45a3eb52a342773a128c1d893",
    "instance_group": 35,
    "diff_mode": false,
    "job_slice_number": 0,
    "job_slice_count": 1,
    "webhook_service": "",
    "webhook_credential": null,
    "webhook_guid": "",
    "host_status_counts": {},
    "playbook_counts": {
        "play_count": 0,
        "task_count": 0
    },
    "custom_virtualenv": "/var/lib/awx/venv/ansible"
}
```

on task_system.log I have this:
```
2021-05-25 13:57:31,190 ERROR    [865fe3123f054fba89ba139109e808af] awx.main.scheduler Failed to list pods for container group ContainerGroup - fscope-0tllbtg7tu-27
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/scheduler/kubernetes.py", line 44, in list_active_jobs
    for pod in pm.kube_api.list_namespaced_pod(pm.namespace, label_selector='ansible-awx={}'.format(settings.INSTALL_UUID)).to_dict().get('items', []):
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/kubernetes/client/api/core_v1_api.py", line 12803, in list_namespaced_pod
    (data) = self.list_namespaced_pod_with_http_info(namespace, **kwargs)  # noqa: E501
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/kubernetes/client/api/core_v1_api.py", line 12891, in list_namespaced_pod_with_http_info
    return self.api_client.call_api(
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/kubernetes/client/api_client.py", line 340, in call_api
    return self.__call_api(resource_path, method,
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/kubernetes/client/api_client.py", line 172, in __call_api
    response_data = self.request(
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/kubernetes/client/api_client.py", line 362, in request
    return self.rest_client.GET(url,
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/kubernetes/client/rest.py", line 237, in GET
    return self.request("GET", url,
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/kubernetes/client/rest.py", line 231, in request
    raise ApiException(http_resp=r)
kubernetes.client.rest.ApiException: (401)
Reason: Unauthorized
HTTP response headers: HTTPHeaderDict({'Audit-Id': 'b89df4fc-c8a4-47af-a07c-36ed01aef41c', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'Www-Authenticate': 'Basic realm="kubernetes-master"', 'Date': 'Tue, 25 May 2021 13:57:31 GMT', 'Content-Length': '129'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Unauthorized","reason":"Unauthorized","code":401}
```

I removed the execution environment from the job template, and ran it using this service account with the default ee. It works

Zokormazo:
Just checked, The secret is good, the serviceaccount I'm using is good, I can deploy this with my serviceaccount using the pull secret created by controller:

```
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  containers:
  - name: private-reg-container
    image: private.registry/awx-ee
  imagePullSecrets:
  - name: automation-e5c74-image-pull-secret-133
```

But controller can't run the job

awx/main/scheduler/kubernetes.py
rooftopcellist: After discussion, I switch these back to raising exception rather than canceling the job.  As a result, the status message in the UI will be "Error"

awx/main/scheduler/kubernetes.py
rooftopcellist: @AlanCoding good point.  That will be cleaner.

shanemcd (approved):

rooftopcellist:
@Zokormazo thank you for the feedback.  I have changed the rules for the service account generated by the installer role in the awx-operator to include get/create/delete permission on secrets.  

I can now run jobs via the controller in an awx-operator deployed AWX.  To do this, I built an AWX image of my branch, then deployed with an awx-operator including [my changes](https://github.com/ansible/awx-operator/pull/334).

image

Then, I ran the demo job, and it runs to completion.  

image


Re: Error handling:
It was advised to raise a runtime error instead of cancelling the job.  As a result, there is no way to pass a job_explanation, so the resulting job status message in the UI will look like this:

image

This isn't ideal, but changing it would require overhauling a good some exception handling logic that would end up being a big can of worms... which should be handled in another PR if at all.  

-----------------------------------------
Also, @Zokormazo re:

> job_explanation is wrong. My openshift credential is valid and enough to run jobs with public ees. We need a better explanation of what's happening.
I don't like canceled as status. This was not canceled, it's an error due to lack of permissions. failed ?

I changed the status to Failed (by raising and error instead), and put a more specific error message for 403's.  ([here](https://github.com/ansible/awx/pull/10204/files#diff-e47089fc124c2bf4fc4bf47dde61fe91888e31331c3e067e27652858f6ae0fc6R99) and [here](https://github.com/ansible/awx/pull/10204/files#diff-e47089fc124c2bf4fc4bf47dde61fe91888e31331c3e067e27652858f6ae0fc6R136))

awx/main/scheduler/kubernetes.py
```diff
@@ -51,6 +54,96 @@ def list_active_jobs(self, instance_group):
...
+                replace_secret = True
+            secret_exists = True
+        except client.rest.ApiException as e:
+            if e.status == 404:
+                secret_exists = False
+            else:
+                error_msg = _('Invalid openshift or k8s cluster credential')
+                if e.status == 403:
+                    error_msg = _(
+                        'Failed to create secret for container group {} because the needed service account roles are needed.  Add get, list, create and delete roles for secret resources for your cluster credential.'.format(
```
shanemcd, line 99: Just noticed this but you can drop `list` from this error message now.

rooftopcellist: Thanks, here is the PR to fix that.  https://github.com/ansible/awx/pull/10289

Issue #10181, opened 2021-05-12
Host Groups Associate Modals Closes on Sort


AlexSCorey:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
The Associate modal closes when you sort the Host Groups names.

##### STEPS TO REPRODUCE

Click on Hosts --> Groups Tab --> Associate ---> Sort by name

##### EXPECTED RESULTS

Group names are returned in the modal in reverse alpha order

##### ACTUAL RESULTS

Modal Closes

##### ADDITIONAL INFORMATION
![modalcloses](https://user-images.githubusercontent.com/39280967/117996216-f6b6f980-b30f-11eb-9330-f17c546a5d53.gif)

akus062381:
this issue is resolved

Issue #9218, opened 2021-02-03
[ui_next] Paging credentials closes the launch popup


jitteryjuice:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
After upgrading AWX from 14.1.0 (to 15.0.1 to 16.0.0) to 17.0.1 it has been noticed that if you launch a job template from the templates menu (that has been found via filtering, or via pagination) using the "Launch Template" button, and paginating the credential list results in the Modal popup closing and the URL that was filtered returning to the root templates URL.

If Launching the template from within the Template itself, the issue is not apparent. In summary, paginating the credentials in the popup modal from the templates list causes the root URL to reset and the filtering/pagination to be lost.

##### ENVIRONMENT
* AWX version: 17.0.0.1
* AWX install method: Kubernetes
* Ansible version:  2.9.17
* Operating System: Container host - Ubuntu Server, using awx container
* Web Browser: Chrome/Edge/Firefox/Brave

##### STEPS TO REPRODUCE

- Install AWX 17.0.1 using Kubernetes
- Configure multiple templates (to exceed default 20 per page)
- Paginate template list to 2nd page or higher
- Click "Launch Template" from one of the templates in the list
- Paginate the credential list
- Watch modal close, and the search results be reset

##### EXPECTED RESULTS

- The credential is selected and applied to the launch template

##### ACTUAL RESULTS

- The launch template modal is closed, search filter emptied and the progress lost
- The URL in the address bar is reset prior to launch template is clicked

##### ADDITIONAL INFORMATION

- The time between modal closing and URL change is too fast for screen capture
- Can upload additional photos upon request
- URL Changes Observed
  - URL in Template List before launching template: https://awx.my.domain/#/templates?template.page=2
  - URL after paginating credential list: https://awx.my.domain/#/templates?credential.page=2
- Happy to provide a PR, however I am not too familiar with react/the code base surrounding the issue. If pointed in the right direction, id be happy to do further investigation into the issue.

MrBones757:
I will also be able to provide info here in addition to @j-shade as we have access to the same instance / environment

mabashian:
Here's the bug in action

![cred_page](https://user-images.githubusercontent.com/9889020/106796630-9b4ae080-6629-11eb-89a6-d0e9a50e40c9.gif)

MrBones757:
@j-shade and I attempted to have a look at this issue today locally, but our lack of react experience wasn't particularly helpful and we didn't get very far.

[...]

Issue #10193, opened 2021-05-13
Job Launch prompt window closes when trying to launch a job


supremedacos:
##### ISSUE TYPE
 - Bug Report

##### ENVIRONMENT
* AWX version: 19.1.0
* AWX install method: k8s docker on linux via awx-operator
* Ansible version:  2.10.8
* Operating System: CentOS Linux release 8.3.2011
* Web Browser: Chrome & Safari

##### STEPS TO REPRODUCE

1. From main web UI page select `Templates`
2. In the search field, search for a template with a name that exists
  The URL will change to something like this - `https://awx.../#/templates?template.name__icontains=sleep`
3. Click the launch rocket button on the right for a template that has an Inventory `Prompt on Launch` checked
4. When the new prompt window is displayed, click on the single `>` button to navigate to the next page of inventories
  The URL will change to something like this - `https://awx.../#/templates?inventory.page=2`
5. Wait 5 seconds

##### EXPECTED RESULTS

The prompt window does not disappear and either I get the final screen to launch the job or more prompt screens.

##### ACTUAL RESULTS

The prompt window disappears.

##### ADDITIONAL INFORMATION

awx-web logs when attempting to launch a job
```
2021-05-13 17:00:29,112 INFO     [bb849c73529441499d799a6435f6652f] awx.analytics.performance request: <WSGIRequest: OPTIONS '/api/v2/unified_job_templates/'>, response_time: 0.401s
[pid: 29|app: 0|req: 823/2886] 10.43.237.14 () {74 vars in 3359 bytes} [Thu May 13 17:00:28 2021] OPTIONS /api/v2/unified_job_templates/ => generated 9598 bytes in 403 msecs (HTTP/1.1 200) 14 headers in 571 bytes (1 switches on core 0)
10.43.237.14 - - [13/May/2021:17:00:29 +0000] "OPTIONS /api/v2/unified_job_templates/ HTTP/1.1" 200 9598 "https://awx.<redacted>/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 11_2_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/89.0.4389.90 Safari/537.36" "10.145.194.10"
2021-05-13 17:00:29,155 INFO     [47c5d4d8111747d4b617281c2bb775ee] awx.analytics.performance request: <WSGIRequest: OPTIONS '/api/v2/workflow_job_templates/'>, response_time: 0.445s
[...]

PR #10215, opened 2021-05-17
fix namespaced url params

keithjgrant:
##### SUMMARY
Fixes issues when multiple lists are on the page (generally only occurs now when on a list page with an open modal that contains a second list) — Navigating within the modal list currently wipes out URL parameters for the main list, causing the page to reload, which closes the modal. The fix prevents changes to one set of namespaced URL parameters from wiping out URL parameters from another namespace

* Refactors query string utils to consolidate a lot of repeated logic from components into a new util, `updateQueryString`. Use this function to modify query string parameters while maintaining params for multiple namespaces at the same time.
* QS utils that are no longer needed have been deleted: `replaceParams` and `encodeNonDefaultQueryString`

Addresses #10181 and #9218 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/components/ListHeader/ListHeader.jsx
+++ b/awx/ui_next/src/components/ListHeader/ListHeader.jsx
@@ -73,6 +73,8 @@ class ListHeader extends React.Component {
     Object.keys(oldParams).forEach(key => {
       oldParams[key] = null;
     });
+    delete oldParams.page_size;
+    delete oldParams.order_by;
     const qs = replaceNamespacedParams(qsConfig, location.search, oldParams);
     this.pushHistoryState(qs);
   }
--- a/awx/ui_next/src/components/ListHeader/ListHeader.jsx
+++ b/awx/ui_next/src/components/ListHeader/ListHeader.jsx
@@ -9,7 +9,7 @@ import {
   parseQueryString,
   mergeParams,
   removeParams,
-  replaceNamespacedParams,
+  updateQueryString,
 } from '../../util/qs';
 import { QSConfig, SearchColumns, SortColumns } from '../../types';
 
@@ -38,7 +38,7 @@ class ListHeader extends React.Component {
   handleSearch(key, value) {
     const { location, qsConfig } = this.props;
     const params = parseQueryString(qsConfig, location.search);
-    const qs = replaceNamespacedParams(qsConfig, location.search, {
+    const qs = updateQueryString(qsConfig, location.search, {
       ...mergeParams(params, { [key]: value }),
       page: 1,
     });
@@ -47,7 +47,7 @@ class ListHeader extends React.Component {
 
   handleReplaceSearch(key, value) {
     const { location, qsConfig } = this.props;
-    const qs = replaceNamespacedParams(qsConfig, location.search, {
+    const qs = updateQueryString(qsConfig, location.search, {
       [key]: value,
     });
     this.pushHistoryState(qs);
@@ -59,11 +59,7 @@ class ListHeader extends React.Component {
     const updatedParams = removeParams(qsConfig, oldParams, {
       [key]: value,
     });
-    const qs = replaceNamespacedParams(
-      qsConfig,
-      location.search,
-      updatedParams
-    );
+    const qs = updateQueryString(qsConfig, location.search, updatedParams);
     this.pushHistoryState(qs);
   }
 
@@ -75,13 +71,13 @@ class ListHeader extends React.Component {
     });
     delete oldParams.page_size;
     delete oldParams.order_by;
-    const qs = replaceNamespacedParams(qsConfig, location.search, oldParams);
+    const qs = updateQueryString(qsConfig, location.search, oldParams);
     this.pushHistoryState(qs);
   }
 
   handleSort(key, order) {
     const { location, qsConfig } = this.props;
-    const qs = replaceNamespacedParams(qsConfig, location.search, {
+    const qs = updateQueryString(qsConfig, location.search, {
       order_by: order === 'ascending' ? key : `-${key}`,
       page: null,
     });
--- a/awx/ui_next/src/components/ListHeader/ListHeader.jsx
+++ b/awx/ui_next/src/components/ListHeader/ListHeader.jsx
@@ -6,11 +6,10 @@ import { Toolbar, ToolbarContent } from '@patternfly/react-core';
 import DataListToolbar from '../DataListToolbar';
 
 import {
-  encodeNonDefaultQueryString,
   parseQueryString,
   mergeParams,
-  replaceParams,
   removeParams,
+  replaceNamespacedParams,
 } from '../../util/qs';
 import { QSConfig, SearchColumns, SortColumns } from '../../types';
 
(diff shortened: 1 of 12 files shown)


keithjgrant:
e2e tests started

awx/ui_next/src/util/qs.js
```diff
@@ -246,3 +247,22 @@ export function replaceParams(oldParams, newParams) {
...
+
+/**
+ * Update namespaced param(s), returning a new query string. Leaves params
+ * from other namespaces unaltered
+ * @param {object} qs config object for namespacing params, filtering defaults
+ * @param {string} the url query string to update
+ * @param {object} namespaced params to add or update
+ * @return {string} url query string
+ */
+export function replaceNamespacedParams(config, queryString, newParams) {
```
keithjgrant: wondering if I should rename this to `updateNamespacedParams`, since it can also delete or add new ones

awx/ui_next/src/components/ListHeader/ListHeader.jsx
```diff
@@ -38,60 +37,59 @@ class ListHeader extends React.Component {
...
+    const qs = replaceNamespacedParams(qsConfig, location.search, {
+      [key]: value,
+    });
+    this.pushHistoryState(qs);
   }
 
   handleRemove(key, value) {
     const { location, qsConfig } = this.props;
-    let oldParams = parseQueryString(qsConfig, location.search);
-    if (parseInt(value, 10)) {
```
jlmitch5, line 56: are you sure you don't need to do this conditional anymore?  I would double check that number-based tag removal works correctly.

awx/ui_next/src/components/ListHeader/ListHeader.jsx
```diff
@@ -38,60 +37,59 @@ class ListHeader extends React.Component {
...
+    const qs = replaceNamespacedParams(
+      qsConfig,
+      location.search,
+      updatedParams
+    );
+    this.pushHistoryState(qs);
   }
 
   handleRemoveAll() {
-    // remove everything in oldParams except for page_size and order_by
```
jlmitch5, line 65: so far as I can tell, this function no longer maintains whatever page_size and order_by when remove all is clicked.  this means that clicking remove all not only removes all the filter tags, but also resets the sort and page size dropdown (which it shouldn't do).

awx/ui_next/src/components/ListHeader/ListHeader.jsx
```diff
@@ -6,11 +6,10 @@ import { Toolbar, ToolbarContent } from '@patternfly/react-core';
 import DataListToolbar from '../DataListToolbar';
 
 import {
-  encodeNonDefaultQueryString,
   parseQueryString,
   mergeParams,
-  replaceParams,
   removeParams,
+  replaceNamespacedParams,
```
jlmitch5: I don't see a similar tweak to mergeParams here (what is used for certain searches, single select if I remember correct...I'm not sure if that is needed or not, I would test a single-select based search to be sure.

FWIW, I don't understand the point of the name change, just seems like more to type, it doesn't add any clarification in terms of the API, IMO.  I would keep the exported function as replaceParams (as it is clear), and change the original replaceParams to something else (and make it a non-exported helper function

awx/ui_next/src/components/ListHeader/ListHeader.jsx
```diff
@@ -38,60 +37,59 @@ class ListHeader extends React.Component {
...
+    const { history } = this.props;
     const { pathname } = history.location;
-    const nonNamespacedParams = parseQueryString({}, history.location.search);
-    const encodedParams = encodeNonDefaultQueryString(
-      qsConfig,
-      params,
-      nonNamespacedParams
-    );
-    history.push(encodedParams ? `${pathname}?${encodedParams}` : pathname);
+    history.push(queryString ? `${pathname}?${queryString}` : pathname);
```
jlmitch5, line 90: the idea here was to encode only those params that were non-default.  things set as default params (both key and value) would be stripped out of the URL.  And finally, there was trickiness with encoding that things needed to be done a certain way to prevent double encoding (which could take things like spaces and encode them twice, resulting in something invalid).  On the latter point, I would manually test that adding keys with spaces is not broken.

All that being said, I think you moved all of this functionality to replacedNamespacedParams, so I think all of it should be working smoothly.

awx/ui_next/src/components/ListHeader/ListHeader.test.jsx
```diff
@@ -71,7 +71,7 @@ describe('ListHeader', () => {
     expect(history.location.search).toEqual(query);
     const toolbar = wrapper.find('DataListToolbar');
     toolbar.prop('clearAllFilters')();
-    expect(history.location.search).toEqual('?item.page_size=5');
+    expect(history.location.search).toEqual('');
```
jlmitch5: this was testing that custom page size would not be stripped out when selecting the clear all button...the fact it needed to be changed does lead me to believe there may be a bug with the refactor?

awx/ui_next/src/util/qs.test.js
```diff
@@ -810,4 +818,66 @@ describe('qs (qs.js)', () => {
...
+    // This fix needed after we're confident refactoring components
+    // to use replaceNamespacedParams provides equivalent functionality
+    test.skip('should not alter params of other namespaces', () => {
+      const query =
+        'template.name__icontains=workflow&template.page=2&credential.page=3';
+      const newParams = {
+        page: 3,
+      };
+      expect(replaceNamespacedParams(config, query, newParams)).toEqual(
+        'template.name__icontains=workflow&template.page=3&credential.page=3'
```
jlmitch5: On first glance, the updates to replace params including the null-valued keys which should be removed is a bug...but it seems you are testing that with this new replaceNamespacedParams and it seems to be working fine.  That again leads me to believe you should just make the old function non-exported and remove those tests as what they were intending to cover is done with these new ones.

I also feel that you should have a test which makes sure default params are stripped out and not encoded (that's more of a concern at the pushHistoryState level, which was done where qs was used, but now that stuff is within the qs lib itself.

jlmitch5 (commented):
This looks good!

The biggest overall change seems to be to pull in the encoding that was happening at the pushHistoryState level within the qs library...which is a better separation of concerns in my opinion.

There were a couple conditionals that were removed that give me some slight pause...I put why they were done the way they were so that you could manually confirm they aren't still needed.  There was also the list header test tweak that I think should be looked at to make sure it is not a bug.

Finally, I feel like it would be beneficial to go all the way and remove the old replaceParams (or rather, rename your new function to that and change the name of the old one to something else).  It's kind of confusing to export both from the qs API when I think really, it's the new one that is used in all the places now, I think.

Great stuff, I'm glad to see this work get a facelift :)

awx/ui_next/src/components/ListHeader/ListHeader.jsx
keithjgrant: yeah, this all moved into the new util function. I confirmed queries with a space still behave (though both before and after these changes, spaces appear in the URL as a space instead of the encoded `%20`, which might be worth fixing at some point

awx/ui_next/src/util/qs.test.js
keithjgrant: Yeah, I agree about not exporting the old function. It's still used in `ActivityStream` though, so I need to see if a different more focused util would do the job there. I also use the old utils inside `replaceNamespacedParams`, so I will keep it and its tests around, but will see about removing the export

AlexSCorey:
@keithjgrant 

It looks like you might still be working on this but I wanted to share some things that I see and some thoughts.  Notice the url after the modal closes.  `updateQueryString` should be called when this modal closes to remove the queries associated with the modal.  Also, there could be more than 1 query associated with the modal in a wizard.  For example, when the user searches for an inventory in step 1 of the POL wizard, and then navigates to the credential step, should the inventory step `qs` be cleared?  


params

keithjgrant:
@AlexSCorey 

> Notice the url after the modal closes. `updateQueryString` should be called when this modal closes to remove the queries associated with the modal.

I wondered about this, but this is the current behavior of the app in devel, so I left it as is. I can see an argument for it working both ways. We can certainly follow up in another PR if we want to fix this

> Also, there could be more than 1 query associated with the modal in a wizard. For example, when the user searches for an inventory in step 1 of the POL wizard, and then navigates to the credential step, should the inventory step `qs` be cleared?

Also a good question... This PR should make it easy for any particular page or modal to "clean up after itself" when it's done with a namespace. I'll punt that to another PR if we want to change it

*edit* I created a separate issue for this: #10240

AlexSCorey (approved):
I pulled this down and tested it. It resolves the two issues described.  I think this is ready. Well done Keith!

akus062381:
@keithjgrant @AlexSCorey I manually tested this against Keith's branch, and I am still seeing the errors in question from https://github.com/ansible/awx/issues/9218 and https://github.com/ansible/awx/issues/10181

Is anybody available to take a look at this with me?

akus062381:
The first issue, referenced in the beginning of this PR, is still manifesting itself. @keithjgrant and I looked at it together- he is going to fix and send back to me.

keithjgrant:
@akus062381 Ready for you to take another look

AlexSCorey:
This will also probably address https://github.com/ansible/awx/issues/10193

akus062381:
@keithjgrant awesome! these issues are resolved

PR #10217, opened 2021-05-18
Updates LingUI Again

AlexSCorey:
##### SUMMARY
This updates ling ui to bring in their latest release so that we don't have to run `npm extract-strings` every time we add/update a `<Plural>` or a string with a variable inside it.

##### ISSUE TYPE
-dependency upgrade

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/package-lock.json
+++ b/awx/ui_next/package-lock.json
@@ -2169,9 +2169,9 @@
       }
     },
     "@lingui/core": {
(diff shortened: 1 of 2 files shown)


AlexSCorey:
E2E test started

tiagodread:
No broken tests due this change, this won't require any aditional tests.

jakemcdermott (approved):

PR #10233, opened 2021-05-19
Pin awxkit version for compatibility with Tower collection

davidtluong:
##### SUMMARY

Addresses problem with Tower Collection not having a pinned awxkit version

Seems execution environments was added in awxkit version 18.0
https://github.com/ansible/awx/blob/18.0.0/awxkit/awxkit/api/pages/api.py#L26

##### ISSUE TYPE

 - Bugfix Pull Request
Solves #4941 on tower repo (not this one) (potentially)
 
##### COMPONENT NAME

 - Tower Collection

##### AWX VERSION

```
Tower 3.8.x
```

##### ADDITIONAL INFORMATION

This will _break_ upstream, please only target this for the Tower collection, not the awx collection.  I was requested to open up this here instead of the other repository.

### DIFF
--- a/awx_collection/requirements.txt
+++ b/awx_collection/requirements.txt
@@ -1,3 +1,3 @@
 pytz  # for tower_schedule_rrule lookup plugin
 python-dateutil>=2.7.0  # tower_schedule_rrule
-awxkit  # For import and export modules
+awxkit awxkit >=9.3.0,<=17.1.0  # For import and export modules


AlanCoding (changes requested):
This could be done as a part of the templating in the release process (leaving open questions of downstream vs upstream)

https://github.com/ansible/awx/blob/devel/awx_collection/tools/roles/template_galaxy/tasks/main.yml

but clearly we cannot put in static version numbers.

AlanCoding:
Not dismissing the issue report. We have some options of a different form, and those will need to be discussed.

PR #10237, opened 2021-05-20
[WIP] Fixes issue where firefox users could not create schedule

### LINKED ISSUE #10152: AWX 19.1: management jobs can not be scheduled
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
I'm trying to edit the schedules for the management jobs, but the (existing) schedules are prefilled with values that can't be saved??

##### ENVIRONMENT
* AWX version: 19.1.0
* AWX install method: k3s on debian buster, deployed with awx-operator 0.9.0
* Ansible version:  n/a at this point
* Operating System: whatever is used inside the containers?
* Web Browser: any

##### STEPS TO REPRODUCE

1. click "management jobs"
2. click any one job of the four
3. edit or create a schedule
4. observe "start time" is prefilled with "NaN-NaN-NaNTNaN:NaN:NaN" even if the job had been edited and/or run before already
5. observe how it is impossible to save the schedule unless you manually change the start time to some value that makes sense

##### EXPECTED RESULTS

I'd expect the "start time" to be either 01-01-1970T00:00:00 or the time the AWX instance was deployed before its first run, then the time&date of the first run

##### ACTUAL RESULTS

Schedule can not be saved.

AlexSCorey:
##### SUMMARY
This resolves #10152.

I examined using PF date and time picker but realized pretty quickly those components need some work.  They are essentially text inputs combined with a calendar/ time widget and as such the user could conceivably try to input a date or time in any format, which would make formatting/validation a nightmare.  As such I elected to use native html elements that are supported by chrome and firefox and they provide a lot more structure reducing the possible formatting and validation issues.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
Fire fox
Screen Shot 2021-05-20 at 11 38 56 AM
Chrome (yes it is in French.  I was investigation how this would impact I18n)
Screen Shot 2021-05-20 at 11 40 47 AM


jakemcdermott:
woahh

jakemcdermott:
Seems like a simple way to solve the related bug, which I really like. I understand that this PR is still WIP so I'll take another look when it's ready.

awx/ui_next/src/components/Schedule/shared/ScheduleForm.jsx
```diff
@@ -110,22 +119,37 @@ function ScheduleFormFields({ hasDaysToKeepField, zoneOptions }) {
...
         label={t`Start date/time`}
       >
-        <input
-          className="pf-c-form-control"
-          type="datetime-local"
-          id="schedule-start-datetime"
-          step="1"
-          {...startDateTime}
-        />
+        <DateTimeGroup>
```
jakemcdermott, line 139: I could _maybe_ see it being useful to move all of this out to it's own component if it makes it simpler to test, but I'll leave that up to you.

@AlexSCorey Does the API expect a `datetime` for these or separate `date` and `time`?  

* If the api expects a `datetime`, we'd probably want some extra test coverage on whatever code we're using to stitch the input values together before presenting them to the api.
* If the api expects separate `date` and `time` values, I like this PR even more :)

AlexSCorey:
Closing in favor of https://github.com/ansible/awx/pull/10291

PR #10251, opened 2021-05-24
tweak message about image pull option

kdelee:
##### SUMMARY

I'm not sure if it is OK to change the migration in this way.

This message in the UI didn't make sense to me. If you all prefer the existing message, that is fine...I just think it is weird and doesn't tell you what it does.

If I don't care, I can leave it on `---------` in the UI which is "do whatever the default is"
If I do care, its weird to choose an option saying I don't choose and option.

Screenshot from 2021-05-24 17-16-28


beeankha (approved):

jbradberry (approved):
@kdelee it is ok to change migrations in this way since the choices text is purely programmatic, it doesn't have anything to do with what is in the database.

Issue #6073, opened 2020-02-25
[Feedback] User should be able to type value in lookups


unlikelyzero:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
A web ui user should not have to click on the magnifying glass icon in order to perform a lookup. Example: the Organization search field under the Create New User form only works if I click on the magnifying glass

In legacy awx, you could type the name of the resource

##### ENVIRONMENT
* AWX version: 9.next

##### ADDITIONAL INFORMATION
<img width="614" alt="Screen Shot 2020-02-25 at 5 16 25 PM" src="https://user-images.githubusercontent.com/12446869/75292637-92834500-57f2-11ea-80a9-2fb93e3fd68b.png">

jlmitch5:
as this enhancement is for lookup fields inside forms (which we refer to as lookups) and not list search bar (which does allow you to type and click enter to create the search), I've updated the text of the issue.

mabashian:
I think this is worth looking in to for an initial release.  I do agree that this would be very nice UX behavior.

mabashian:
link https://github.com/ansible/awx/issues/6089 - there's a lot of overlap between these two issues IMO.  We could potentially address both with one set of changes.

tiagodread:
Verified that:

single select lookup can be typed and find results properly
multiple select lookup keep working as is
all e2e tests are passing

PR #10257, opened 2021-05-25
Adds support for typing values into single select lookups

mabashian:
##### SUMMARY
link #6073 

This adds support for typing values into single select lookups.  Multi-select lookups (credentials on the JT form, instance groups, etc) will remain unchanged.

An input is now rendered next to the lookup button which will allow users to type a value in.

Here's the button (which opens the modal):

Screen Shot 2021-05-25 at 9 37 52 AM

Here's the input (which lets users type a value in):

Screen Shot 2021-05-25 at 9 37 52 AM

There's a debounce on the input so that we only make a request to check to see if the name is valid after a second of no typing.

The tricky part of this implementation was handling validation on the lookup fields.  If a user types in a string that does not exactly match a value that they can use then we want to show an error indicating that no matching value was found:

Screen Shot 2021-05-25 at 9 41 25 AM

but this was tricky because some of these fields are required so the validation functions would need to change a bit.

Since the typed text and the actual underlying value of the lookup are different the validation function would need to have access to both values whenever it runs for this to work.  This meant either storing the typed text up at the form level _or_ pushing the validation down into the lookup(s).  After talking with @keithjgrant we decided to go down the route of pushing the validation down to the lookups.

You can now pass an optional validation function down to the various lookups that will get combined with the typeahead validation via the useField hook.

Here's the whole thing in action:

typeahead

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/components/Lookup/CredentialLookup.jsx
+++ b/awx/ui_next/src/components/Lookup/CredentialLookup.jsx
@@ -114,31 +114,32 @@ function CredentialLookup({
 
   const checkCredentialName = useCallback(
     async name => {
-      if (name && name !== '') {
-        try {
-          const typeIdParams = credentialTypeId
-            ? { credential_type: credentialTypeId }
-            : {};
-          const typeKindParams = credentialTypeKind
-            ? { credential_type__kind: credentialTypeKind }
-            : {};
-          const typeNamespaceParams = credentialTypeNamespace
-            ? { credential_type__namespace: credentialTypeNamespace }
-            : {};
+      if (!name) {
+        onChange(null);
+        return;
+      }
 
-          const {
-            data: { results: nameMatchResults, count: nameMatchCount },
-          } = await CredentialsAPI.read({
-            name,
-            ...typeIdParams,
-            ...typeKindParams,
-            ...typeNamespaceParams,
-          });
-          onChange(nameMatchCount ? nameMatchResults[0] : null);
-        } catch {
-          onChange(null);
-        }
-      } else {
+      try {
+        const typeIdParams = credentialTypeId
+          ? { credential_type: credentialTypeId }
+          : {};
+        const typeKindParams = credentialTypeKind
+          ? { credential_type__kind: credentialTypeKind }
+          : {};
+        const typeNamespaceParams = credentialTypeNamespace
+          ? { credential_type__namespace: credentialTypeNamespace }
+          : {};
+
+        const {
+          data: { results: nameMatchResults, count: nameMatchCount },
+        } = await CredentialsAPI.read({
+          name,
+          ...typeIdParams,
+          ...typeKindParams,
+          ...typeNamespaceParams,
+        });
+        onChange(nameMatchCount ? nameMatchResults[0] : null);
+      } catch {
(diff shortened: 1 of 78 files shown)


nixocio:
I triggered an e2e test run.

nixocio (approved):

awx/ui_next/src/components/Lookup/CredentialLookup.jsx
```diff
@@ -111,6 +112,39 @@ function CredentialLookup({
     }
   );
 
+  const checkCredentialName = useCallback(
+    async name => {
+      if (name && name !== '') {
```
keithjgrant: To un-nest this all a bit, you could handle the `else` case first and return early:

```
if (!name) {
  onChange(null);
  return;
}
...
```

Also, the empty string check is redundant, so probably not necessary

awx/ui_next/src/components/Lookup/InventoryLookup.jsx
```diff
@@ -73,6 +74,24 @@ function InventoryLookup({
     }
   );
 
+  const checkInventoryName = useCallback(
+    async name => {
+      if (name && name !== '') {
```
keithjgrant: same here with returning early to un-nest the logic one level

awx/ui_next/src/components/Lookup/Lookup.jsx
```diff
@@ -44,9 +45,23 @@ function Lookup(props) {
...
+    fieldName,
+    validate,
   } = props;
+  const [typedText, setTypedText] = useState('');
+  const debounceRequest = useDebounce(onDebounce, 1000);
+
+  useField({
+    name: fieldName,
+    validate: val => {
+      if (!multiple && !val && typedText && typedText !== '') {
```
keithjgrant, line 59: empty string check is redundant

awx/ui_next/src/components/Lookup/OrganizationLookup.jsx
```diff
@@ -69,6 +69,24 @@ function OrganizationLookup({
     }
   );
 
+  const checkOrganizationName = useCallback(
+    async name => {
+      if (name && name !== '') {
```
keithjgrant: same

awx/ui_next/src/components/Lookup/ProjectLookup.jsx
```diff
@@ -72,6 +73,24 @@ function ProjectLookup({
     }
   );
 
+  const checkProjectName = useCallback(
+    async name => {
+      if (name && name !== '') {
```
keithjgrant: same

keithjgrant (approved):
looks good! just a few minor readability suggestions

awx/ui_next/src/components/Lookup/CredentialLookup.jsx
mabashian: Good call - I like that

tiagodread (approved):
Verified that:
- single select lookup can be typed and find results properly
- multiple select lookup keep working as is
- all e2e tests are passing

PR #10258, opened 2021-05-25
allow singular execution_environment for cli

kdelee:
this makes "execution_environments" or "execution_environment" work
which is helpful

Now these both work:
```
awx execution_environments create --name foobar --image quay.io/ansible/awx-ee:latest
```
```
awx execution_environment create --name foobar1 --image quay.io/ansible/awx-ee:latest
```


beeankha (approved):

jbradberry:
Wait, do we actually want to add brand-new things to the deprecated list?

jbradberry (commented):
Wait, do we actually want to add brand-new things to the deprecated list?

kdelee:
That's reasonable if you don't want this...I just think its kinda nice to accept plural or singular

PR #10261, opened 2021-05-25
Bump receptor and receptorctl

shanemcd:


requirements/requirements.txt
```diff
@@ -297,7 +297,7 @@ pyyaml==5.4.1
     #   djangorestframework-yaml
     #   kubernetes
     #   receptorctl
-receptorctl==0.9.7
+receptorctl==1.0.0.0a2
```
shanemcd, line 300: Yes this looks weird, no the extra 0 is not a mistake. PBR seems to have strong opinions about how alpha versions should be formatted 🤷

Spredzy (approved):

Issue #9715, opened 2021-03-26
Clarify what it means for project updates to have a default execution environment


balonik:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Execution environment is not used when executing (manual) project sync.

Project sync is an Ansible Playbook run as well, it is expected to honor the Execution Environment setting?

##### ENVIRONMENT
* AWX version: 18.0.0
* AWX install method: AWX Operator
* Ansible version:  X.Y.Z
* Operating System:
* Web Browser:

##### STEPS TO REPRODUCE

Change the `Instance Group -> tower` definition to:

```
apiVersion: v1
kind: Pod
metadata:
  namespace: awx
spec:
  containers:
    - image: my_custom_awx-ee:latest
      name: worker
      args:
        - ansible-runner
        - worker
```

Create new SCM project and try to sync.
The embedded (default) `awx-ee` container in `awx` pod is still used for the execution of Project Sync.

Add new Execution Environment in AWX.
Edit the SCM project and define Default Execution Environment to the newly created Execution Environment.
The embedded (default) `awx-ee` container in `awx` pod is still used for the execution of Project Sync.

##### EXPECTED RESULTS

Project Sync job will use the `Instance Group -> tower` setting if set.
Project Sync job will use the Default Execution Environment configured for that Project.

##### ACTUAL RESULTS

The embedded (default) `awx-ee` container in `awx` pod is still used for the execution of Project Sync.

shanemcd:
This is by design. There is an EE inside of the control plane pod used for Project Updates. Some projects can be really big, and the overhead of cloning them inside of a remote EE and streaming the project back to the control plane would be a major performance impact. You can override the EE used inside of the control plane by supplying `tower_ee_image` on your CR like this: https://github.com/ansible/awx-operator/blob/4eb4abf4f54d84857149ece35bedbaaf907fbaaa/deploy/crds/awx_v1beta1_cr.yaml#L18

shanemcd:
I just noticed that we _are_ allowing you to provide an EE in the UI for Projects. This is a bug. Will discuss with the team.

jbradberry:
This is not the intended meaning of "Default Execution Environment".  The default environment is one that gets inherited by jobs that make use of this project (or organization or whatever other objects we added a Default Execution Environment field to), when the job does not have an explicitly configured environment of its own.  As noted, project updates are forced to use the system execution environment, but this is not a bug.  I recommend that we close this issue.

[...]

PR #10278, opened 2021-05-26
Update message about Default Execution Environment

nixocio:
Update message about Default Execution Environment

Organizations

image

Projects

image

See: https://github.com/ansible/awx/issues/9715

### DIFF
--- a/awx/ui_next/src/screens/Organization/shared/OrganizationForm.jsx
+++ b/awx/ui_next/src/screens/Organization/shared/OrganizationForm.jsx
@@ -93,7 +93,7 @@ function OrganizationFormFields({
         onBlur={() => executionEnvironmentHelpers.setTouched()}
         value={executionEnvironmentField.value}
         onChange={value => executionEnvironmentHelpers.setValue(value)}
-        popoverContent={t`Select the default execution environment for this organization.`}
+        popoverContent={t`The execution environment that will be used for jobs inside of this organization. This will be used a fallback when an execution environment has not been explicitly assigned at the project, job template or workflow level.`}
         globallyAvailable
         organizationId={organizationId}
         isDefaultEnvironment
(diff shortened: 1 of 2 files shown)


shanemcd:
Should we say something like "jobs that _use_ this project"?

shanemcd:
Maybe something like this:

> The Execution Environment that will be used for jobs that use this project. This will be used a fallback when an EE has not been explicitly assigned at the Job Template or Workflow level.

Will defer to @wenottingham and @tvo318 about the capitalization of resource types here.

jbradberry:
I think, barring feedback from @wenottingham and @tvo318 otherwise, that @shanemcd 's last suggestion (https://github.com/ansible/awx/pull/10278#issuecomment-848783949) should be used.

nixocio:
@jbradberry @shanemcd updated as per suggestions.

nixocio:
regate

awx/ui_next/src/screens/Organization/shared/OrganizationForm.jsx
```diff
@@ -93,7 +93,7 @@ function OrganizationFormFields({
         onBlur={() => executionEnvironmentHelpers.setTouched()}
         value={executionEnvironmentField.value}
         onChange={value => executionEnvironmentHelpers.setValue(value)}
-        popoverContent={t`Select the default execution environment for this organization.`}
+        popoverContent={t`The Execution Environment that will be used for jobs inside of this organization. This will be used a fallback when an EE has not been explicitly assigned at the Project, Job Template or Workflow level.`}
```
wenottingham: In general, Execution Environment is not capitalized... but neither would be Project, Job Template, or Workflow. So... meh.

I would not use 'EE' - would spell it out.

nixocio: @wenottingham updated.

wenottingham (approved):

PR #10290, opened 2021-05-26
Add Thycotic DevOps Secrets Vault support

amigus:
##### SUMMARY
Add a Credential Plugin for Thycotic DevOps Secrets Vault 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API
 - UI

##### AWX VERSION
```
awx: 19.1.0
```

### DIFF
--- a/awx/main/credential_plugins/dsv.py
+++ b/awx/main/credential_plugins/dsv.py
@@ -13,6 +13,13 @@ dsv_inputs = {
             'help_text': _('The tenant e.g. "ex" when the URL is https://ex.secretservercloud.com'),
             'type': 'string',
         },
+        {
+            'id': 'tld',
+            'label': _('Top-level Domain (TLD)'),
+            'help_text': _('The TLD of the tenant e.g. "com" when the URL is https://ex.secretservercloud.com'),
+            'choices': ['ca', 'com', 'com.au', 'com.sg', 'eu'],
+            'default': 'com',
+        },
         {'id': 'client_id', 'label': _('Client ID'), 'type': 'string'},
         {
             'id': 'client_secret',
@@ -28,12 +35,6 @@ dsv_inputs = {
             'type': 'string',
             'help_text': _('The secret path e.g. /test/secret1'),
         },
-        {
-            'id': 'expr',
-            'label': _('Attribute Expression'),
-            'type': 'string',
-            'help_text': _('A Python expression on the dict representation of the secret e.g. "data.username"'),
-        },
     ],
     'required': ['tenant', 'client_id', 'client_secret', 'path'],
 }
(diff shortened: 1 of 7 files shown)


awx/main/credential_plugins/dsv.py
```diff
@@ -0,0 +1,71 @@
...
+            'default': 'com',
+            'choices': ['ca', 'com', 'com.au', 'com.sg', 'eu'],
+        }
+    )
+
+
+dsv_plugin = CredentialPlugin(
+    'Thycotic DevOps Secrets Vault',
+    dsv_inputs,
+    lambda **kwargs: eval(
```
shanemcd: Do we really need to use `eval` here?

amigus: Yes and no. There has to be some mechanism to allow users to "select" the field or fields they want from the secret data. Eval was the simplest. At one point, I had a TODO to constrain it with a safety check. Now that you've raised the issue, I'll look at that and make some changes.

The alternative is to allow for the use of "JSONpath," which is the pseudo-equivalent of XPath, but I opted against that because I don't see much adoption and/or a prevalent library that does it.

awx/main/credential_plugins/dsv.py
```diff
@@ -0,0 +1,71 @@
...
+    dsv_inputs['fields'].append(
+        {
+            'id': 'url_template',
+            'label': _('URL template'),
+            'type': 'string',
+            'default': 'https://{}.secretsvaultcloud.{}/v1',
+        }
+    )
+
+if to_locale(get_language()) != 'en_US':
```
matburt: Can you explain what's going on here?

awx/main/credential_plugins/dsv.py
shanemcd: @amigus Any luck with that? This will definitely need to be addressed before we can consider this PR.

amigus: Understood and no disagreement. I'll resolve it ASAP as there are people waiting on the plugin.

awx/main/credential_plugins/dsv.py
amigus: Why, a dubious assumption, of course! 😁

The TLD reflects where the cloud instance lives and it defaults to '.com' which is a US deployment.

The assumption is that the user will need to use a non-default TLD for DSV only when AWX is running in a non-US region.

I originally had an override in settings.py but I removed it.

amigus:
Customers want the plugin ASAP so I'm going to remove the conditionals and swap the eval for a simple match. Give me a day or two.

awx/main/credential_plugins/dsv.py
matburt: It just seems strange, does it not make more sense to just have them enter the base url on its own? If the tld or, god forbid, the url structure needs to change then it would require a code change that wouldn't be picked up until the next Tower release. Seems like something that could get broken pretty easily?

amigus:
Welp, I tried to go for the gusto and implement expression evaluation using pyjq but I couldn't get the `make docker-compose-build` to succeed once I added pyjq to requirements.in and `./updater.sh`. I got:

> #24 68.93 ERROR: Could not build wheels for pyjq which use PEP 517 and cannot be installed directly

If anyone has a fix for that LMK otherwise I'm stuck implementing the minimum. 😒

amigus:
This version no longer has an `eval` nor does it check the locale. Reasonable?

matburt:
I think you might have messed up the rebase on this

image

amigus:
Argh! So just rebase to devel and push?

amigus:
> Argh! So just rebase to devel and push?

Pardon my ignorance but do I just need to rebase the branch onto devel and push it?

beeankha:
Hey @amigus , hopefully the following will help:

1) Run `git remote -v` and note the name of the remote that's the upstream AWX repo
2) While on your working branch (the one for this PR), rebase with `devel` from the name of that upstream remote via `git pull --rebase [name_of_upstream_remote] devel`

amigus:
Did I do it right? Are we ready to merge?

matburt (approved):
This all looks reasonable to me

shanemcd (approved):
Thanks for seeing this through!

amigus:
Thanks a bunch, fellas!  FYI, I'll be submitting a 'strikingly similar' one for Thycotic Secret Server, shortly.

Issue #6214, opened 2020-03-06
Swap input type=date element for PF Date Picker 


marshmalien:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Tracking PF Issues
React: https://github.com/patternfly/patternfly-react/issues/2010
Core: https://github.com/patternfly/patternfly-next/issues/180

This issue is blocked until PF React component merges upstream

mabashian:
I believe this is ready to go - going to move to needs_devel

marshmalien:
PatternFly Status Update:

* They released a beta date picker component upstream - [date picker docs](https://www.patternfly.org/v4/components/date-picker/react)
* Their next priority is the [date component cleanup issue](https://github.com/patternfly/patternfly-react/issues/5057) with target date of 11.18.2020. 
* The PF React [time picker issue](https://github.com/patternfly/patternfly-react/issues/4731) is scheduled for 12.09.2020 
* The PF React [date range picker issue](https://github.com/patternfly/patternfly-react/issues/4732) is scheduled for 12.09.2020

PF-React Date and Time component tracking issue: https://github.com/patternfly/patternfly-react/issues/4711

mabashian:
AFAIK, there are only two places currently where we display date form fields:

https://github.com/ansible/awx/blob/devel/awx/ui_next/src/components/Schedule/shared/FrequencyDetailSubform.jsx#L564-L569

https://github.com/ansible/awx/blob/devel/awx/ui_next/src/components/Schedule/shared/ScheduleForm.jsx#L114-L120

IIRC, this PF component doesn't support time selection so if/when we do this we may need to split the field into date & time.

mabashian:
@AlexSCorey I believe you're current working on this in conjunction with the schedule bug(s).  If that's the case I'll update the assignee/status.

mabashian:
See https://github.com/ansible/awx/issues/10152 and https://github.com/ansible/awx/pull/10237

tiagodread:
This is now implemented with #10291

PR #10291, opened 2021-05-26
Resolves issue where FireFox users could not schedule a job template

AlexSCorey:
##### SUMMARY
This addresses #10152 and https://github.com/ansible/awx/issues/6214

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
Screen Shot 2021-05-26 at 3 47 16 PM

### DIFF
--- a/awx/ui_next/src/components/Schedule/shared/DateTimePicker.jsx
+++ b/awx/ui_next/src/components/Schedule/shared/DateTimePicker.jsx
@@ -1,0 +1,67 @@
+import React from 'react';
+import { t } from '@lingui/macro';
+import { useField } from 'formik';
+import {
+  DatePicker,
+  isValidDate,
+  yyyyMMddFormat,
+  TimePicker,
+  FormGroup,
+} from '@patternfly/react-core';
+import styled from 'styled-components';
+import { required, validateTime, combine } from '../../../util/validators';
+
+const DateTimeGroup = styled.span`
+  display: flex;
+`;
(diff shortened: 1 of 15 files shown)


awx/ui_next/src/components/Schedule/shared/DateTimePicker.jsx
```diff
@@ -0,0 +1,66 @@
...
+      label={`${label} date/time`}
+    >
+      <DateTimeGroup>
+        <DatePicker
+          aria-label={t`${label} date`}
+          {...dateField}
+          value={dateField.value.split('T')[0]}
+          onChange={onDateChange}
+        />
+        <TimePicker
```
mabashian, line 54: I think the default time picker placeholder is a bit confusing since it doesn't actually match up character for character with the format.  I wonder if we change it to something like `hh:mm AM/PM` if that would help folks know what the format is (if they're typing it in).  This probably isn't a huge deal for us since we always have a default value set but I found this confusing the first time I played with this component.  I thought it was expecting time in a 24 hour clock syntax like `21:45` instead of `09:45 PM`.

mabashian (approved):

tiagodread (approved):
Now the datepicker component can be successfully rendered in Chrome or Firefox.

Verified that:
- Users can edit schedules created by the old component
- Users can create/edit schedules
- Jobs can be launched by a schedule

image

Issue #8824, opened 2020-12-10
[ui_next] [UX] When adding role to team multiple times state is erroneously remembered


unlikelyzero:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
We currently preserve the last known configuration on the Add user workflow. Also

```Dissociating team roles is not really working. I’ve created a role to admin an inventory and then deleted it. Then I dissociated that and the list was clean. Then I tried to add an org role and the inventory one came with it. This seems related to the not resetting the form above.```

##### ENVIRONMENT
* AWX version: 12/10

##### STEPS TO REPRODUCE
1. Add a resource type permissions to a user from the team page.
2. Add another resource type permission with the 'Add' button
Note: refilled form. Including Admin rights

##### EXPECTED RESULTS
Some permission state remembered. Not the Admin permission role.

##### ADDITIONAL INFORMATION
![Screen Shot 2020-12-10 at 1 30 22 PM](https://user-images.githubusercontent.com/12446869/101814256-e2b26600-3aeb-11eb-9336-c55f23b11e98.png)

unlikelyzero:
May need to split this up based on:

```Dissociating team roles is not really working. I’ve created a role to admin an inventory and then deleted it. Then I dissociated that and the list was clean. Then I tried to add an org role and the inventory one came with it. This seems related to the not resetting the form above.```

mabashian:
This sounds much more like a bug than an enhancement

mabashian:
@unlikelyzero this definitely sounds like two separate issues.  I'm not able to reproduce the first issue but it does sound concerning.

>Dissociating team roles is not really working. I’ve created a role to admin an inventory and then deleted it. Then I dissociated that and the list was clean. Then I tried to add an org role and the inventory one came with it. This seems related to the not resetting the form above.

This part is a bit confusing to me.  Do you have concrete steps to repro?

mabashian:
Here's my tldr; version:

1) Go to a team's roles list
2) Click add and associate a job template admin role (the job template doesn't matter just make sure it's one that's not already associated)
3) Without refreshing/navigating away click add again
(Note that the state for the previous add is still present - this is not correct, it should be back to it's original form)
4) Attempt to associate a _different_ job template admin role

Duplicate requests will get made.  One request will successfully associate the role, the other will result in a 400.

We should only be making the one request.

Another slightly different situation that is probably the exact same bug:

1) Go to a team's roles list
[...]

PR #10292, opened 2021-05-26
Fixes bug where user/team role add modal state is not cleared on close

mabashian:
##### SUMMARY
link #8824

I modeled these changes after the pattern that existed between ResourceAccessList and AddResourceRole.  The key to fixing this bug is that the component that renders the wizard needs to be unmounted when the wizard closes so that the state, etc get cleared out before the next time the wizard is opened.  In order to achieve that I needed to decouple the add button from the wizard.

The sort of weird part of this pattern (and this exists in the ResourceAccessList as well) is error handling.  We pass the error back and set that to state before rendering the modal which isn't quite as clean as having the request made out at the list level and leveraging our hooks for error handling but I decided to just get in and get out and not worry about trying to refactor too much.

Here it is in action:

role_modal

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/components/AddRole/SelectResourceStep.jsx
+++ b/awx/ui_next/src/components/AddRole/SelectResourceStep.jsx
@@ -101,7 +101,9 @@ function SelectResourceStep({
         headerRow={
           <HeaderRow qsConfig={QS_Config(sortColumns)}>
             {sortColumns.map(({ name, key }) => (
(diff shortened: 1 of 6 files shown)


nixocio (approved):

nixocio:
Very good explanation of the solution @mabashian. I triggered an e2e test run.

akus062381:
@mabashian this works great and fixes the bug. Thank you!

PR #10304, opened 2021-05-27
Adds ouiaId to YAML and JSON buttons in Variables Detail

akus062381:
small change to add a better locator to these buttons. if we do not do this, our locators for our visual tests are not good and will eventually break tests


mabashian:
Talked with @akus062381 about this one.  I think we'll need a slightly more complicated solution since this will result in duplicate ouiaId's on pages where we have multiple variable details (like the cred type detail for example).

jakemcdermott (approved):

akus062381:
@AlexSCorey this isn't ready for test yet- per Mike's message above, this provides a locator that will be duplicated in certain sections of the app, which would break our tests. a bit more dev work needs to be done here, and it is above my React skill level
@jakemcdermott said he'd look at it later today

PR #10310, opened 2021-05-28
Update tower_notification module routing

tadeboro:
##### SUMMARY

When the name of the module was updated, a redirect that was added did not use a fully-qualified collection name as the redirection target.

When we try to use the old name, Ansible reports the following error:

    $ ansible -m awx.awx.tower_notification localhost
    localhost | FAILED! => {
      "msg": "The module awx.awx.tower_notification was redirected to
        tower_notification_template, which could not be loaded."
    }

Changes in this commit add the namespace and collection name to the redirect.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - awx.awx.tower_notification_template

### DIFF
--- a/awx_collection/meta/runtime.yml
+++ b/awx_collection/meta/runtime.yml
@@ -43,4 +43,4 @@ plugin_routing:
         removal_date: TBD
         warning_text: see plugin documentation for details
     tower_notification:
-      redirect: tower_notification_template
+      redirect: awx.awx.tower_notification_template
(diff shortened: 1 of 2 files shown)


fosterseth:
We've made some changes to the awx.awx collection, which includes a proper redirect

https://github.com/ansible/awx/blob/b10eb6f4c2f6f3f65869c487808d1413bfc4473a/awx_collection/meta/runtime.yml#L97

you'll also notice we've dropped the `tower_` prefix from module names

tadeboro:
Closing this as the underlying issue was solved in a different way.

Issue #10287, opened 2021-05-26
Sysadmins can edit/delete managed_by_tower=true EE but can't then create new ones


kdelee:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

@jbradberry has more background on this.

Now that sysadmins are allowed to edit and delete `managed_by_tower=true` EE, it stands to reason that they should be able to modify the `managed_by_tower` attribute via the API.

This is necessary to fix your project updates if an overzelous sysadmin deletes all their `managed_by_tower=true` EEs....because if you do that there is no EE left to run project updates with.

##### STEPS TO REPRODUCE

Delete all your `managed_by_tower=true` EE
Run a project update
Everything is broken
Regret what you did
Try and recreate the default EE and set managed_by_tower=true

##### EXPECTED RESULTS

Able to fix mistake

##### ACTUAL RESULTS

Unable to modify `managed_by_tower` attribute via the API -- only way to do it is in `shell_plus`

nixocio:
It would nice to add those changes to the UI also.

shanemcd:
Should we allow this through the API? Or just tell people to re-run `awx-manage register_default_execution_environments`?

jbradberry:
A good question.

jbradberry:
related #10321 #9851

jbradberry:
I think #10378 finishes this up.

nixocio:
The final decision was to not allow the user to delete managed by tower EE. Then there is no need to update the UI.
<img width="1468" alt="image" src="https://user-images.githubusercontent.com/9053044/121432090-3b9e7200-c948-11eb-9486-be3bf469de8a.png">

kdelee:
Now sysadmins can restore the control plane EE which is the only managed_by_tower=True EE via an awx-manage command. Going to close this as we have some other issues tracking other aspects of this

Issue #9851, opened 2021-04-08
Instance Group or Job fails with AttributeError: 'NoneType' object has no attribute 'image'


balonik:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
In UI `Instance Groups` ends up with Error 500
It is not possible to run ANY Job.

`api/v2/instance_groups/1/` returns

```
{
    "id": 1,
    "type": "instance_group",
    "url": "/api/v2/instance_groups/1/",
    "related": {
        "named_url": "/api/v2/instance_groups/tower/",
        "jobs": "/api/v2/instance_groups/1/jobs/",
        "instances": "/api/v2/instance_groups/1/instances/"
    },
    "name": "tower",
    "created": "2020-12-03T09:59:18.467551Z",
    "modified": "2021-04-08T08:56:34.198408Z",
    "capacity": 0,
    "committed_capacity": 0,
    "consumed_capacity": 0,
    "percent_capacity_remaining": 0.0,
    "jobs_running": 0,
    "jobs_total": 446,
    "instances": 0,
    "controller": null,
    "is_controller": false,
    "is_isolated": false,
    "is_container_group": true,
    "credential": null,
    "policy_instance_percentage": 100,
    "policy_instance_minimum": 0,
    "policy_instance_list": [],
    "pod_spec_override": "apiVersion: v1\nkind: Pod\nmetadata:\n  namespace: awx\nspec:\n  containers:\n    - image: registry.local/awx-ee:latest\n      name: worker\n      args:\n        - ansible-runner\n        - worker",
    "summary_fields": {
        "user_capabilities": {
            "edit": true,
            "delete": true
        }
    }
}
```

AWX spec snippet

```
apiVersion: awx.ansible.com/v1beta1
kind: AWX
metadata:
  name: awx
  namespace: awx
spec:
  tower_ee_images:
    - name: my default
      image: registry.local/awx-ee:latest
...
```

##### ENVIRONMENT
* AWX version: 19.0.0
* AWX install method: awx-operator 0.8.0 on RKE
* Ansible version:  X.Y.Z
* Operating System:
* Web Browser:

##### STEPS TO REPRODUCE

Deploy AWX 19.0.0 using awx-operator 0.8.0 with custom `tower_ee_images` defined in YAML file

##### EXPECTED RESULTS

Working

##### ACTUAL RESULTS

HTTP 500 or Job Error

##### ADDITIONAL INFORMATION

UI error logs from `awx-web` container

```
[...]

PR #10321, opened 2021-06-01
Create default EE in seperate awx-managment cmd

kdelee:
Create EE at a seperate time and also attach a registry credential if
auth information provided

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -2603,8 +2603,6 @@ class CredentialSerializer(BaseSerializer):
         return summary_dict
 
     def validate(self, attrs):
-        if self.instance and self.instance.managed_by_tower:
-            raise PermissionDenied(detail=_("Modifications not allowed for managed credentials"))
         return super(CredentialSerializer, self).validate(attrs)
 
     def get_validation_exclusions(self, obj=None):
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -2603,6 +2603,8 @@ class CredentialSerializer(BaseSerializer):
         return summary_dict
 
     def validate(self, attrs):
+        if self.instance and self.instance.managed_by_tower:
+            raise PermissionDenied(detail=_("Modifications not allowed for managed credentials"))
         return super(CredentialSerializer, self).validate(attrs)
 
     def get_validation_exclusions(self, obj=None):
--- a/awx/main/management/commands/register_default_execution_environments.py
+++ b/awx/main/management/commands/register_default_execution_environments.py
@@ -1,6 +1,8 @@
 # Copyright (c) 2015 Ansible, Inc.
 # All Rights Reserved
 import sys
+from distutils.util import strtobool
+from argparse import RawTextHelpFormatter
 
 from django.core.management.base import BaseCommand
 from django.conf import settings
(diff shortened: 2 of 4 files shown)


jbradberry:
@kdelee do we have any associated PRs that make use of this new command during install or upgrade?

jbradberry (approved):

jbradberry:
This PR addresses the problem in #10287.  Also, #9851, but we need to change the packaging to invoke this command.

kdelee:
@jbradberry I've got  a PR for the downstream traditional installer, but we will need to make a PR to the operator to use this, except in the operator we may need to some extra finess around the secrets because we won't be able to directly pass secrets to this command

for operator:
 - need to call the register_default_execution_environment commands
 - for ee that need auth, we need to pull secret from an openshift secret, see use of k8s_info for other secrets like the SECRET_KEY
 - https://github.com/ansible/awx-operator/blob/devel/roles/installer/tasks/broadcast_websocket_configuration.yml#L10-L15

Zokormazo:
Since `DEFAULT_EXECUTION_ENVIRONMENTS` is a list, I went with 2 ees there and ended with two credentials named as `Default Execution Environment Registry Credential` . Since we are not allowing to set a credential per image, could we just reuse the first generated one?

kdelee:
@Zokormazo thats strange, did you call the command twice? I was hoping using `update_or_create` on the credential would prevent that. Let me try on my side again

kdelee:
@Zokormazo you were right :D I fixed my use of `update_or_create` and it should work now. Can you try it out?

awx/main/management/commands/register_default_execution_environments.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class Command(BaseCommand):
+    """Create default execution environments, intended for new installs"""
+
+    help = """
+    Creates or updates the execution environments set in settings.DEFAULT_EXECUTION_ENVIRONMENTS if they are not yet created.
+    Optionally provide authentication details to create or update a container registry credential that will be set on all of these default execution environments.
+    Note that settings.DEFAULT_EXECUTION_ENVIRONMENTS is and ordered list, the first in the list will be used for project updates and system jobs.
+    """
```
kdelee, line 19: @shanemcd @jbradberry I updated this help text, what do you think?

shanemcd (approved):

awx/api/serializers.py
```diff
@@ -2603,8 +2603,6 @@ def get_summary_fields(self, obj):
         return summary_dict
 
     def validate(self, attrs):
-        if self.instance and self.instance.managed_by_tower:
-            raise PermissionDenied(detail=_("Modifications not allowed for managed credentials"))
```
kdelee: @jbradberry mentioned I should revert this

awx/main/management/commands/register_default_execution_environments.py
jbradberry: "... is an ordered list"

rebeccahhh (approved):
I like this a lot

awx/main/management/commands/register_default_execution_environments.py
```diff
@@ -0,0 +1,104 @@
...
+        )
+        parser.add_argument(
+            "--registry-password",
+            type=str,
+            default="",
+            help="Password or token for CLI authentication with the container registry",
+        )
+        parser.add_argument(
+            "--verify-ssl",
+            type=lambda x: bool(strtobool(str(x))),
```
kdelee, line 48: nice

Issue #10210, opened 2021-05-17
Expose the output of the EE resolver in the job template detail API endpoints


jbradberry:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

It seems like it would be useful and would mitigate some confusion to expose which EE would be used when executing a template that doesn't have an EE explicitly assigned.

**Clarification**: template detail endpoints.

shanemcd:
Note: Meyers had some performance concerns here. Chat with him about that before implementing this.

jbradberry:
Meyers' concern was the obvious one about not exposing potentially expensive calculated fields in the list views.

thenets:
We made some tests and it's working.

Those are the coverage scenarios:

- **job_template without EE specified**
  - Endpoints
    - related.unified_job_template
    - job_templates
  - Result:
    - {'id': 5, 'name': 'tests @ 3:25:21 PM', 'description': '', 'image': 'test'}

- **job_template with explicit EE associated** (ee.id == 17)
  - Endpoints
    - related.unified_job_template
    - job_templates
  - Result: {'id': 17, 'name': 'EE - AppearanceAddress', 'description': '', 'image': 'quay.io/ansible/ansible-runner:devel'}

We also updated the automated tests.

PR #10322, opened 2021-06-01
Expose the EE resolver to the API

jbradberry:
##### SUMMARY

This change exposes the output of `.resolve_execution_environment()` for job templates as a new summary field in the API.  Note that this is only exposed for the detail views, not the list views.  Also, there is the caveat for job templates that may be run under workflows that the workflow EE might override the results of this value.

related #10210

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```

```

### DIFF
--- a/awx/main/models/mixins.py
+++ b/awx/main/models/mixins.py
@@ -477,11 +477,11 @@ class ExecutionEnvironmentMixin(models.Model):
             if wf_template.execution_environment is not None:
                 return wf_template.execution_environment
             wf_node = getattr(wf_node.workflow_job, 'unified_job_node', None)
-        if getattr(self, 'project', None) and self.project.default_environment is not None:
+        if getattr(self, 'project_id', None) and self.project.default_environment is not None:
             return self.project.default_environment
-        if getattr(self, 'organization', None) and self.organization.default_environment is not None:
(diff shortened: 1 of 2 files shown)


shanemcd (approved):

rooftopcellist (approved):

awx/main/models/mixins.py
```diff
@@ -464,15 +464,24 @@ class Meta:
...
+        Return the execution environment that should be used when executing a job.
         """
         if self.execution_environment is not None:
             return self.execution_environment
+        template = getattr(self, 'unified_job_template', None)
+        if template is not None and template.execution_environment is not None:
+            return template.execution_environment
+        wf_node = getattr(self, 'unified_job_node', None)
+        while wf_node is not None:
+            wf_template = wf_node.workflow_job.workflow_job_template
```
chrismeyersfsu: Can't quite articulate when/why but feel like `workflow_job` and `workflow_job_template` _could_ be null.

jbradberry: I believe that's further addressed in #10351

jbradberry: (also, it's part of the already merged PR)

PR #10323, opened 2021-06-01
Add additional controller directory for collections for inventory update

wenottingham:
##### SUMMARY

This may be present in some scenarios with additional collections for inventory usage.

### DIFF
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -2391,6 +2391,12 @@ class RunInventoryUpdate(BaseTask):
                         paths = [config_values[config_setting]] + paths
             paths = [os.path.join(CONTAINER_ROOT, folder)] + paths
             env[env_key] = os.pathsep.join(paths)
+        if 'ANSIBLE_COLLECTIONS_PATHS' in env:
+            paths = env['ANSIBLE_COLLECTIONS_PATHS'].split(':')
+        else:
+            paths = ['~/.ansible/collections', '/usr/share/ansible/collections']
+        paths.append('/usr/share/automation-controller/collections')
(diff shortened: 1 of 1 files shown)


AlanCoding:
I could be wrong, but I don't think this works. We still want `/usr/share/ansible/collections` to work.

https://github.com/ansible/ansible/blob/devel/lib/ansible/config/base.yml

```yaml
COLLECTIONS_PATHS:
  name: ordered list of root paths for loading installed Ansible collections content
  description: >
    Colon separated paths in which Ansible will search for collections content.
    Collections must be in nested *subdirectories*, not directly in these directories.
    For example, if ``COLLECTIONS_PATHS`` includes ``~/.ansible/collections``,
    and you want to add ``my.collection`` to that directory, it must be saved as
    ``~/.ansible/collections/ansible_collections/my/collection``.
  default: ~/.ansible/collections:/usr/share/ansible/collections
  type: pathspec
  env:
  - name: ANSIBLE_COLLECTIONS_PATHS  # TODO: Deprecate this and ini once PATH has been in a few releases.
  - name: ANSIBLE_COLLECTIONS_PATH
    version_added: '2.10'
  ini:
  - key: collections_paths
    section: defaults
  - key: collections_path
    section: defaults
    version_added: '2.10'
```

This is a list-valued setting. If the setting is not present in the existing `env`, then you have to go and append the defaults, or else it will no longer consider those default locations part of the collection paths.

wenottingham:
@AlanCoding updated

AlanCoding (approved):
I'm still interested to find some related tracking info and documentation for this, but given what I know of the new folder, this PR looks like it integrates it alright.

PR #10324, opened 2021-06-01
Introduce distinct controlplane instance group

shanemcd:


awx/main/conf.py
```diff
@@ -177,6 +177,24 @@
     read_only=True,
 )
 
+register(
+    'DEFAULT_CONTROL_PLANE_QUEUE_NAME',
+    field_class=fields.CharField,
+    label=_('The instance group where control plane tasks run'),
```
AlanCoding, line 183: The name of the instance group where control plane tasks run?

The default EE uses a ForeignKey instead of name, this isn't consistent. We've never had settings reference objects before, which is why we're facing it with this release cycle.

If the user changes the value, what is the expected behavior? It won't create a new instance group with that name right? So would they have to make a corresponding change to their installer (or installer variables)?

shanemcd: Did you miss the `read_only=True` part?

matburt (approved):

Spredzy (approved):

beeankha (approved):

PR #10325, opened 2021-06-01
Add a field for hosts automated across to the subscription info

wenottingham:
##### SUMMARY

This is populated by the new table we've added.
Update the subs check to check against this, not imported hosts.

Screenshot from 2021-06-01 13-30-09

##### ISSUE TYPE
 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME
 - API
 - UI

### DIFF
--- a/awx/main/models/events.py
+++ b/awx/main/models/events.py
@@ -552,7 +552,7 @@ class JobEvent(BasePlaybookEvent):
             summaries = dict()
             updated_hosts_list = list()
             for host in hostnames:
-                updated_hosts_list.append(host)
+                updated_hosts_list.append(host.lower())
                 host_id = self.host_map.get(host, None)
                 if host_id not in existing_host_ids:
                     host_id = None
--- a/awx/main/utils/licensing.py
+++ b/awx/main/utils/licensing.py
@@ -35,7 +35,7 @@ from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
 
 # AWX
-from awx.main.models import Host
+from awx.main.models import Host, HostMetric, Instance
 
 MAX_INSTANCES = 9999999
 
@@ -383,12 +383,20 @@ class Licenser(object):
             current_instances = Host.objects.active_count()
         else:
             current_instances = 0
+        license_date = int(attrs.get('license_date', 0) or 0)
+        automated_instances = HostMetric.objects.count()
+        first_host = HostMetric.objects.only('first_automation').order_by('first_automation').first()
+        if first_host:
+            automated_since = int(first_host.first_automation.timestamp())
+        else:
+            automated_since = int(Instance.objects.order_by('id').first().created.timestamp())
         instance_count = int(attrs.get('instance_count', 0))
         attrs['current_instances'] = current_instances
-        free_instances = instance_count - current_instances
+        attrs['automated_instances'] = automated_instances
+        attrs['automated_since'] = automated_since
+        free_instances = instance_count - automated_instances
         attrs['free_instances'] = max(0, free_instances)
 
-        license_date = int(attrs.get('license_date', 0) or 0)
         current_date = int(time.time())
         time_remaining = license_date - current_date
         attrs['time_remaining'] = time_remaining
--- a/awx/ui_next/src/components/DetailList/NumberSinceDetail.jsx
+++ b/awx/ui_next/src/components/DetailList/NumberSinceDetail.jsx
@@ -1,0 +1,29 @@
+import React from 'react';
+import { node, string } from 'prop-types';
+import { t } from '@lingui/macro';
+import styled from 'styled-components';
+import { formatDateString } from '../../util/dates';
+import _Detail from './Detail';
+
+const Detail = styled(_Detail)`
+  word-break: break-word;
+`;
+
+function NumberSinceDetail({ label, number, date, dataCy = null }) {
+  const dateStr = formatDateString(date);
+
+  return (
+    <Detail
+      label={label}
+      dataCy={dataCy}
+      value={t`${number} since ${dateStr}`}
+    />
+  );
+}
+NumberSinceDetail.propTypes = {
+  label: node.isRequired,
+  number: string.isRequired,
+  date: string.isRequired,
+};
+
+export default NumberSinceDetail;
--- a/awx/main/management/commands/inventory_import.py
+++ b/awx/main/management/commands/inventory_import.py
@@ -36,20 +36,20 @@ from awx.main.utils.pglock import advisory_lock
 logger = logging.getLogger('awx.main.commands.inventory_import')
 
 LICENSE_EXPIRED_MESSAGE = '''\
-License expired.
-See http://www.ansible.com/renew for license extension information.'''
+Subscription expired.
+Contact us (https://www.redhat.com/contact) for subscription extension information.'''
 
 LICENSE_NON_EXISTANT_MESSAGE = '''\
-No license.
-See http://www.ansible.com/renew for license information.'''
+No subscription.
+Contact us (https://www.redhat.com/contact) for subscription information.'''
 
 LICENSE_MESSAGE = '''\
-Number of licensed instances exceeded, would bring available instances to %(new_count)d, system is licensed for %(instance_count)d.
-See http://www.ansible.com/renew for license extension information.'''
+%(new_count)d instances have been automated, system is subscribed for %(instance_count)d.
+Contact us (https://www.redhat.com/contact) for upgrade information.'''
 
 DEMO_LICENSE_MESSAGE = '''\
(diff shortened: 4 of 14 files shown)


wenottingham:
@trahman73 @mabashian I'd love UI/UX review on if there are nicer ways to display that 'Hosts automated" field, either in wording or in adjusting how it's shown.

trahman73:
@wenottingham "Hosts automated" section looks good to me.

awx/main/models/events.py
```diff
@@ -518,7 +518,7 @@ def _update_host_summary_from_stats(self, hostnames):
             summaries = dict()
             updated_hosts_list = list()
             for host in hostnames:
-                updated_hosts_list.append(host)
+                updated_hosts_list.append(host.lower())
```
wenottingham, line 555: This changes the host metrics table to fold mixed-case hosts into a single host. cc @amolgautam25

awx/main/utils/licensing.py
```diff
@@ -383,9 +384,18 @@ def validate(self):
...
+            automated_since = first_host.first_automation
+        else:
+            automated_instances = 0
+            automated_since = now()
         instance_count = int(attrs.get('instance_count', 0))
         attrs['current_instances'] = current_instances
-        free_instances = instance_count - current_instances
+        attrs['automated_instances'] = automated_instances
+        attrs['automated_since'] = automated_since
+        free_instances = instance_count - automated_instances
```
wenottingham, line 397: This changes the definition of `free_instances` to be based on the HostMetric automated instances.

awx/main/models/events.py
amolgautam25: Got it. 👍🏻

amolgautam25 (approved):

wenottingham:
recheck

awx/ui_next/src/components/DetailList/NumberSinceDetail.jsx
```diff
@@ -0,0 +1,28 @@
...
+`;
+
+function NumberSinceDetail({ label, number, date, dataCy = null }) {
+  const dateStr = formatDateString(date);
+
+  return (
+    <Detail
+      label={label}
+      dataCy={dataCy}
+      value={`${number} since ${dateStr}`}
```
AlexSCorey:
```suggestion
      value={t`${number} since ${dateStr}`}
```

awx/ui_next/src/components/DetailList/NumberSinceDetail.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { node, string } from 'prop-types';
+import styled from 'styled-components';
+import { formatDateString } from '../../util/dates';
+import _Detail from './Detail';
+
```
AlexSCorey, line 7:
```suggestion
import {t} from '@lingui/macro'
```

AlexSCorey (changes requested):
The UI looks good, just 2 minor i18n changes

awx/ui_next/src/components/DetailList/NumberSinceDetail.jsx
wenottingham:
```diff --git a/awx/ui_next/src/components/DetailList/NumberSinceDetail.jsx b/awx/ui_next/src/components/DetailList/NumberSinceDetail.jsx
index 617e7273b4..495e91b13f 100644
--- a/awx/ui_next/src/components/DetailList/NumberSinceDetail.jsx
+++ b/awx/ui_next/src/components/DetailList/NumberSinceDetail.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { node, string } from 'prop-types';
+import {t} from '@lingui/macro';
 import styled from 'styled-components';
 import { formatDateString } from '../../util/dates';
 import _Detail from './Detail';
@@ -15,7 +16,7 @@ function NumberSinceDetail({ label, number, date, dataCy = null }) {
     <Detail
       label={label}
       dataCy={dataCy}
-      value={`${number} since ${dateStr}`}
+      value={t`${number} since ${dateStr}`}
     />
   );
 }
```
Screenshot from 2021-06-02 16-21-08

@AlexSCorey Did I miss something somewhere?

AlexSCorey: haha, probably not.  We updated a dependency to solve that issue last week so, depending on how you are building you likely need to update dependencies.

AlexSCorey (approved):
looks good to me.

tiagodread:
running e2e tests

tiagodread:
I got the changes and for some reason, the values `{number} since {dateStr}` are not being displayed in a production view:

image

awx/main/utils/licensing.py
```diff
@@ -383,12 +384,24 @@ def validate(self):
             current_instances = Host.objects.active_count()
         else:
             current_instances = 0
+        license_date = int(attrs.get('license_date', 0) or 0)
+        if HostMetric:
```
chrismeyersfsu: The `HostMetric` model from the import should always exist. Won't this always be True?

awx/main/utils/licensing.py
```diff
@@ -383,12 +384,24 @@ def validate(self):
...
+            automated_instances = HostMetric.objects.count()
+            first_host = HostMetric.objects.order_by('first_automation').first()
```
chrismeyersfsu:
```    "main_hostmetric_pkey" PRIMARY KEY, btree (hostname)
    "main_hostmetric_first_automation_b747a0a1" btree (first_automation)
    "main_hostmetric_hostname_87ac3c1f_like" btree (hostname varchar_pattern_ops)
    "main_hostmetric_last_automation_11010683" btree (last_automation)
``` 

Index exists on `first_automation`, so this is great, if you wanted to get real fancy you could do `first_host = HostMetric.objects.only('first_automation').order_by('first_automation').first()` so the query could be fully satisfied by the index & not have to read the record from disk.

awx/main/utils/licensing.py
```diff
@@ -383,12 +384,24 @@ def validate(self):
...
+            if first_host:
+                automated_since = int(first_host.first_automation.timestamp())
+            else:
+                automated_since = license_date
+        else:
+            automated_instances = 0
+            automated_since = now()
```
chrismeyersfsu: I think this else will go away. If it stays though, `automated_since` becomes a mix of date and seconds since epoch i.e. 

```>>> str(datetime.datetime.now())
'2021-06-07 13:26:27.456871'
```

awx/main/utils/licensing.py
```diff
@@ -383,12 +384,24 @@ def validate(self):
             current_instances = Host.objects.active_count()
         else:
             current_instances = 0
+        license_date = int(attrs.get('license_date', 0) or 0)
+        if HostMetric:
+            automated_instances = HostMetric.objects.count()
```
chrismeyersfsu: After doing the db partitioning work counting any tables makes me twitch a little. I don't love memoizing this value either. Computed fields have bitten us too many times in the past also. 

Don't have a better solution currently so this is probably fine. There are probably bigger fish to fry on the conf endpoint.

awx/main/utils/licensing.py
wenottingham: Yea, this needs fixed, incoming.

awx/main/utils/licensing.py
wenottingham: Updated

awx/main/utils/licensing.py
```diff
@@ -383,12 +383,20 @@ def validate(self):
             current_instances = Host.objects.active_count()
         else:
             current_instances = 0
+        license_date = int(attrs.get('license_date', 0) or 0)
+        automated_instances = HostMetric.objects.count()
+        first_host = HostMetric.objects.only('first_automation').order_by('first_automation').first()
+        if first_host:
+            automated_since = int(first_host.first_automation.timestamp())
+        else:
+            automated_since = int(Instance.objects.order_by('id').first().created.timestamp())
```
wenottingham, line 392: The prior version of this used `license_date` for the since date if there are no automated hosts... which is the license *end* date. Oops.

This uses the nearest analog I can think of for "install date". @chrismeyersfsu do you see any problems with this?

chrismeyersfsu: Instance created date is better than I would have thought.

Below is from our dogfood server that has been alive for a long time (has migrated servers and everything).
image

vs.

image

 `int(Instance.objects.order_by('created').first().created.timestamp())` <-- maybe find the oldest `Instance`. Id is _pretty_ much that so it's fine as is.

wenottingham:
@tiagodread can you recheck?

tiagodread (approved):
After pushing the .po files the strings was properly displayed in prod build
image

PR #10328, opened 2021-06-01
Allow the ExecutionEnvironment.managed_by_tower field to be editable …

jbradberry:
##### SUMMARY

…again

This reverses the last of the restrictions put in place by #9346

related #10287 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION

```

```

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -1396,7 +1396,6 @@ class ProjectOptionsSerializer(BaseSerializer):
 
 class ExecutionEnvironmentSerializer(BaseSerializer):
     show_capabilities = ['edit', 'delete', 'copy']
-    managed_by_tower = serializers.ReadOnlyField()
 
     class Meta:
         model = ExecutionEnvironment


rebeccahhh (approved):

shanemcd:
Should we allow this through the API? Or just tell people to re-run awx-manage register_default_execution_environments?

rebeccahhh:
I have the same question, originally we planned to explicitly not allow this through the API @jbradberry

awx/api/serializers.py
```diff
@@ -1396,7 +1396,6 @@ def validate(self, attrs):
 
 class ExecutionEnvironmentSerializer(BaseSerializer):
     show_capabilities = ['edit', 'delete', 'copy']
-    managed_by_tower = serializers.ReadOnlyField()
```
rebeccahhh, line 1399: we should keep this, imo. removing it will cause headaches for us later down the line and cause the field to be too flexible

rebeccahhh (commented):
we should keep this field read only

jbradberry:
All right, after discussion it seems like we don't want to be _that_ permissive.  Closing.

Issue #7509, opened 2020-07-01
Select all checkbox is tied to tightly to items per page


chrismeyersfsu:
##### STEPS TO REPRODUCE

* check 5 resources (1 checked should not be the first 5)
* set the page size to 5

##### EXPECTED RESULTS

* the select all checkbox to not be checked

##### ACTUAL RESULTS

* the select all box is checked

![image](https://user-images.githubusercontent.com/722880/86274225-56e0f900-bb9f-11ea-8845-9c6f62dde1c0.png)

![image](https://user-images.githubusercontent.com/722880/86274248-62342480-bb9f-11ea-994c-71fc7d2e6b47.png)

mabashian:
link #6853 - I think there's some overlap between these two

mabashian:
Marking this as blocked until #6853 is completed

akus062381:
this is resolved.

Issue #6853, opened 2020-04-28
Delete Only Items On Screen


AlexSCorey:
##### ISSUE TYPE
-Enhancement
##### SUMMARY
Currently when we selectAll and then delete items it is unclear as to what is exactly being deleted.  This feature needs to be improved upon so that only the items that are rendered on the screen are deleted. Things to consider:
-Deleting all items on page
-Deleting filter, sorted, searched items
-Navigation after deleting bulk delete.  See #6734 
This solution should also fix the bug described in #6734 
-In the case where a user select n number of items to be deleted, and then filters the list, all selected items should be unselected and the user will have to recheck the items they want to delete.
##### ENVIRONMENT
* AWX version: 11.1
* AWX install method: docker for ma
* Ansible version:  2.10
* Operating System: Mojave
* Web Browser: Chrome

##### STEPS TO REPRODUCE

 See #6734

AlexSCorey:
see the work done https://github.com/ansible/awx/pull/6784 that might be helpful

mabashian:
@AlexSCorey  As I understand it, clicking the Select All checkbox selects all the rows currently listed on the page and throws them into state.  If you then filter the list down to a subset and click delete, we'll still delete all of the originally selected rows (?).  Is that the problem we're trying to solve here?  If so, we may want to make the title a bit more specific.

As far as #6734 goes, I'm not sure that the delete action should be handling whether or not a previous page of contents still exists.  In order to do that, we'd have to make a request to see if the previous page has rows _before_ we redirect.  I feel like it'd be much better to have the list itself redirect to page 1 when a user attempts to access a page that does not exist.

akus062381:
@keithjgrant this looks great!

akus062381:
this is resolved.

PR #10329, opened 2021-06-02
Clear list selections on pagination/search/sort

keithjgrant:
##### SUMMARY
Updates nearly every list* so that URL param changes (pagination, search, or sort) clear the selection. This prevents the list of selected items in state from including items that may no longer appear on screen — preventing the user from accidentally deleting or otherwise altering an item they may not realize they still have selected.

This also updates the `useSelected` hook to provide `selectAll` and `clearSelected` functions. Any lists that weren't yet already using this hook have been updated to do so.

Addresses #6853 and #7509

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION

*Lists that do not include this change are modals where the user is expected to paginate through screens and make several selections along the way (e.g. Multi Credential select modal), and lists that still use PaginatedDataList and are yet to be converted to PaginatedTable

Note: I originally wanted to make the `clearSelected` prop on PaginatedTable required, so any lists that don't have this fix applied would fail loudly. Unfortunately that wasn't possible, as there were a few lists that should not have this behavior, so I had to leave it as an optional prop.

### DIFF
--- a/awx/ui_next/src/components/DataListToolbar/DataListToolbar.jsx
+++ b/awx/ui_next/src/components/DataListToolbar/DataListToolbar.jsx
@@ -37,7 +37,6 @@ function DataListToolbar({
   onExpand,
   onSelectAll,
   additionalControls,
-
   qsConfig,
   pagination,
 }) {
--- a/awx/ui_next/src/components/JobList/JobList.jsx
+++ b/awx/ui_next/src/components/JobList/JobList.jsx
@@ -90,7 +90,6 @@ function JobList({ defaultParams, showTypeColumn = false }) {
     selected,
     isAllSelected,
     handleSelect,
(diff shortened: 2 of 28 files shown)


AlexSCorey (commented):
I think this is a really easy and clean way of solving this problem.  We probably want to write some tests for the unselecting of list items on search, pagination etc, at the very least, inside `useSelected.test.jsx`

jakemcdermott (approved):
:+1:

akus062381:
@keithjgrant this looks good!

akus062381:
@keithjgrant can you please rebase and fix the conflict? once you are able to do that, we can add the mergeit label

PR #10334, opened 2021-06-02
Adds ability to refresh project revision on sync'd rows in the project list

mabashian:
##### SUMMARY
This PR also adds the revision to the project details view as well as handles updating the revision on the project details view when the project is done syncing.

Since the project revision is _not_ included in the websocket payload, when a project is done syncing the displayed revision _may_ be out of date.  As such, we wanted to expose that information to the user and give them the ability to "refresh" and fetch the new revision.

Here's what that flow looks like:

project_list_sync

When a particular row finishes syncing the user should see this in place of the revision:

Screen Shot 2021-06-02 at 9 57 10 AM

Clicking on that refresh button goes out and fetches the updated project (and with it the potentially updated revision).

We don't do this automatically on the projects list (and force the user to click on the refresh button) is due to issues we've had in the past with the UI triggering API calls based on websocket events.

The flow when a user is on the project details view is a little different because I wasn't as worried about spamming the API with requests.

When a project finishes syncing and the user is viewing the details I do go ahead and automatically refresh the project data (and with it, the revision).  Here's what that looks like:

project_detail_sync

A few other notes:

1) @tiagodread @akus062381 @one-t I'm almost certain this is going to break some tests because I removed the ClipboardCopyButton in favor of PF's [ClipboardCopy component](http://patternfly-react.surge.sh/components/clipboard-copy).  This component looks and behaves slightly differently from our home grown solution.  Additionally, the PF ClipboardCopy button does not expose the ouiaId prop so I had to add a `data-cy` on that component.  You'll likely have to use that identifier to then grab the button inside in order to test out the clipboard copy functionality.
2) Source Control Revision is a net-new detail in the project details.  I think this was simply missed on our initial build of this page.

Here are the identifiers on the various bits:

Screen Shot 2021-06-02 at 10 10 24 AM
Screen Shot 2021-06-02 at 10 11 21 AM
Screen Shot 2021-06-02 at 10 10 48 AM

Note that the identifiers on the project rows have the `id` of the project appended to them for uniqueness.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI


awx/ui_next/src/screens/Project/ProjectDetail/ProjectDetail.jsx
```diff
@@ -146,6 +156,37 @@ function ProjectDetail({ project }) {
...
+                hoverTip={t`Copy full revision to clipboard.`}
+                onCopy={event => {
+                  const clipboard = event.currentTarget.parentElement;
+                  const el = document.createElement('textarea');
+                  el.value = scm_revision.toString();
+                  clipboard.appendChild(el);
+                  el.select();
+                  document.execCommand('copy');
+                  clipboard.removeChild(el);
+                }}
```
jakemcdermott: `execCommand` is deprecated ([apparently](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand))

... But I think we can do this now:

```suggestion
                onCopy={() => navigator.clipboard.writeText(scm_revision.toString())}
```

awx/ui_next/src/screens/Project/ProjectList/ProjectList.jsx
```diff
@@ -212,13 +251,14 @@ function ProjectList() {
             )}
             renderRow={(project, index) => (
               <ProjectListItem
-                fetchProjects={fetchProjects}
+                fetchUpdatedProjects={fetchUpdatedProjects}
```
jakemcdermott:
```suggestion
                fetchProject={fetchUpdatedProjects}
```

 [The list item](https://github.com/ansible/awx/blob/devel/awx/ui_next/src/screens/Project/ProjectList/ProjectListItem.jsx#L45)  still expects this to be called `fetchProject`.

fetchProjectProp

jakemcdermott (changes requested):

akus062381:
@mabashian - I am testing this manually, and I am seeing the same source control revision number for every project. Is this what you intend?

image

akus062381:
stubs added for this on https://github.com/ansible/tower-qa/pull/6393/files

awx/ui_next/src/screens/Project/ProjectDetail/ProjectDetail.jsx
mabashian: 👍 will check that out.  I actually just yoinked this code from the PF React repo: https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/ClipboardCopy/ClipboardCopy.tsx#L13-L21.  Seems like they might need an update too.

awx/ui_next/src/screens/Project/ProjectList/ProjectList.jsx
mabashian: Oof - good catch

jakemcdermott (approved):

akus062381 (approved):
@mabashian I manually tested this, all is working as expected. good to merge!

Issue #8596, opened 2020-11-17
[ui_next] [Meta-issue] Failed to execute 'Send' on Websocket: Still in CONNECTING state.


unlikelyzero:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Multiple ways to trigger Failed to execute 'Send' on Websocket: Still in CONNECTING state.
1. Toggling between Job Output and Detail causes invalidstateerror. 
2. Relaunching Jobs

##### ENVIRONMENT
* AWX version: 10664d1931d789c281e2c9d3d2f14c0a69cb9e13

##### STEPS TO REPRODUCE
1. Launch a Job
2. Navigate to the Output Page and toggle between Details and Output tabs

##### EXPECTED RESULTS
No errors

##### ACTUAL RESULTS
Unhandled error

##### ADDITIONAL INFORMATION
![Screen Recording 2020-11-17 at 08 50 37 AM](https://user-images.githubusercontent.com/12446869/99398198-17971880-28b2-11eb-8ba4-c798a90fe95c.gif)

mabashian:
Yea, I hit this by spamming the relaunch button 5-10 times

akus062381:
**Important Note: this same error is appearing when you click the Relaunch button from the job output screen.**

tiagodread:
This bug was fixed with #10339, there are still some places where the websocket is being closed:

- Host tabs: Moving from Groups tab to Jobs tab
- Project list after clicking on a project

may be fixed with https://github.com/ansible/awx/issues/10192

my2ter:
Old topic and people have probably moved away and use latest AWX but for those still stuck in 17.1 - or around that version - with similar error, the reason for the Job Events panel to not reload is due to the **jobSocketCounter** not being updated properly...

The websocket works, you can see in the Network tab that it gets updated and receives messages (if you don't then your error is not this issue and you should look at the nginx proxy conf - that might be the problem for you)

Othwerwise for this see below:

https://github.com/ansible/awx/blob/17.1.0/awx/ui_next/src/screens/Job/JobOutput/JobOutput.jsx#L259

In here update to:

```
componentDidMount() {
    const { job } = this.props;
    this._isMounted = true;
    this.loadJobEvents();
    this.jobSocketCounter = 0;

    connectJobSocket(job, data => {
      if (data.group_name === 'job_events') {
        if (data.counter && data.counter > this.jobSocketCounter) {
          this.jobSocketCounter = data.counter;
        }
      }
[...]

PR #10339, opened 2021-06-02
Only attempt to reconnect socket if connection wasn't closed cleanly

mabashian:
##### SUMMARY
link https://github.com/ansible/awx/issues/8596

After some investigation, I believe that this error is caused by the reconnect logic that we have.  When the component that references the ws hook unmounts, we disconnect the socket.  There's some logic in the `onclose` method that attempts to reconnect the socket but if we've disconnected cleanly we shouldn't try to reconnect (because we probably meant to disconnect in the first place).

This _should_ clean up the console errors that we've been seeing about the socket already being in a connected state since we won't have timers running past the lifecycle of the component.

cc @keithjgrant does this sound good to you?

Here's the spec for the disconnect event https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent#properties

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/util/useWebsocket.js
+++ b/awx/ui_next/src/util/useWebsocket.js
@@ -33,9 +33,11 @@ export default function useWebsocket(subscribeGroups) {
     ws.current.onclose = e => {
       // eslint-disable-next-line no-console
       console.debug('Socket closed. Reconnecting...', e);
-      setTimeout(() => {
-        connect();
-      }, 1000);
+      if (e.code !== 1000) {
+        setTimeout(() => {
+          connect();
+        }, 1000);
+      }
     };
 
     ws.current.onerror = err => {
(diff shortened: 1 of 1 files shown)


keithjgrant:
Interesting. This makes sense to me. I'm wondering if this also explains why we've seen duplicate open websockets on some pages?

awx/ui_next/src/util/useWebsocket.js
```diff
@@ -33,9 +33,11 @@ export default function useWebsocket(subscribeGroups) {
     ws.current.onclose = e => {
       // eslint-disable-next-line no-console
       console.debug('Socket closed. Reconnecting...', e);
```
keithjgrant: We should probably put this log statement inside the `if`, since it doesn't try to reconnect otherwise

keithjgrant (approved):

tiagodread:
@mabashian I navigated through pages manually with console opened and this is fixed for almos all cases, running the e2e tests I searched in the logs this error and this can be also found in:

- Host tabs: Moving from `Groups` tab to `Jobs` tab
image

- Project list after clicking on a project

tiagodread (approved):

Issue #9905, opened 2021-04-14
Unable to cancel job - method POST not allowed - awx19


phemmer:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
I had an inventory sync job that was stuck. Upon trying to cancel it, I got an error saying `Method "POST" not allowed`

This is a refile of #9714 as the issue is still not fixed.

##### ENVIRONMENT
* AWX version: 19.0.0
* AWX install method: minishift
* Ansible version:  
* Operating System: Debian 10
* Web Browser: Chrome

##### STEPS TO REPRODUCE

1. Get a stuck job?
2. Try to cancel job.

##### EXPECTED RESULTS

job to cancel

##### ACTUAL RESULTS

```
POST /api/v2/inventory_updates/11/cancel/ 405
Method "POST" not allowed.
```
![image](https://user-images.githubusercontent.com/1826947/114759279-f7b63480-9d2b-11eb-96ff-2d5669e4c4a1.png)

phemmer:
Ok, so this looks to be a UI issue. The first time you click cancel, nothing visible happens. Then you click it again and you get the error. However the first cancel actually succeeded, and the second cancel is failing.

shanemcd:
Need to handle the case where the user tries to cancel an already canceled job.

phemmer:
While possibly true, I think making the "cancel" dialog go away when cancelled would be a higher priority.

jakemcdermott:
On which screen are you attempting to cancel the job? Is it a job list, the details panel for the job, the output page, or somewhere else?

phemmer:
The job output page.

akus062381:
this is resolved.

PR #10346, opened 2021-06-03
Disable cancel button on http 405

jakemcdermott:
##### SUMMARY
for #9905 

When a 405 is received, it means the job is already cancelled. Treat the
request as a success and don't show an error modal.

We disable the button to handle a rare scenario where we receive the 405 long before
the job status is properly updated to "cancelled" over websockets. We want to prevent
more cancel requests when this happens. We're disabling instead of hiding the button
because, assuming the status hasn't changed over sockets, the buttons that usually
replace the cancel button on the toolbar won't be shown yet and we don't want to
needlessly flicker and shift button positions around by rapidly loading and unloading
a different number of buttons onto the bar.

### DIFF
--- a/awx/ui_next/src/components/JobCancelButton/JobCancelButton.jsx
+++ b/awx/ui_next/src/components/JobCancelButton/JobCancelButton.jsx
@@ -27,11 +27,14 @@ function JobCancelButton({
     cancelError
   );
 
+  const isAlreadyCancelled = cancelError?.response?.status === 405;
+
   return (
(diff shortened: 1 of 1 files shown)


keithjgrant (approved):

akus062381 (approved):
@jakemcdermott Tiago and I worked on this together. We tried to reproduce this error manually without success. We will work on an automated test to try to reproduce the 405 behavior. This looks good to be merged.

PR #10359, opened 2021-06-04
Remove uses of ansible_virtualenv_path

AlanCoding:
Run a job with recent AWX, then inspect:

```
In [2]: Job.objects.order_by('-created').first().custom_virtualenv
Out[2]: '/var/lib/awx/venv/ansible'
```

We shouldn't do this. The job was ran after the `custom_virtualenv` field has become entirely informative / non-functional. It is very arguable whether _jobs_ should keep this field. For pre-migration jobs, it is an accurate record of fact, so I'm not going out of my way to argue for removal.

With this change, new jobs will record a blank value for `custom_virtualenv`, which is more correct, because we don't do this.

I have carefully looked over @rebeccahhh's https://github.com/ansible/awx/pull/10090, and I don't see any way that `ansible_virtualenv_path` is used. Model use is reported for _templates_, so that doesn't block removing the field on _jobs_ either.

### DIFF
--- a/awx/main/management/commands/inventory_import.py
+++ b/awx/main/management/commands/inventory_import.py
@@ -66,13 +66,9 @@ class AnsibleInventoryLoader(object):
         /usr/bin/ansible/ansible-inventory -i hosts --list
     """
 
-    def __init__(self, source, venv_path=None, verbosity=0):
+    def __init__(self, source, verbosity=0):
         self.source = source
         self.verbosity = verbosity
-        if venv_path:
-            self.venv_path = venv_path
-        else:
-            self.venv_path = settings.ANSIBLE_VENV_PATH
 
     def get_base_args(self):
         bargs = ['podman', 'run', '--user=root', '--quiet']
@@ -131,7 +127,6 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         parser.add_argument('--inventory-name', dest='inventory_name', type=str, default=None, metavar='n', help='name of inventory to sync')
         parser.add_argument('--inventory-id', dest='inventory_id', type=int, default=None, metavar='i', help='id of inventory to sync')
-        parser.add_argument('--venv', dest='venv', type=str, default=None, help='absolute path to the AWX custom virtualenv to use')
         parser.add_argument('--overwrite', dest='overwrite', action='store_true', default=False, help='overwrite the destination hosts and groups')
         parser.add_argument('--overwrite-vars', dest='overwrite_vars', action='store_true', default=False, help='overwrite (rather than merge) variables')
         parser.add_argument('--keep-vars', dest='keep_vars', action='store_true', default=False, help='DEPRECATED legacy option, has no effect')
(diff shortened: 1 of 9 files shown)


awx/main/management/commands/inventory_import.py
```diff
@@ -131,7 +127,6 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         parser.add_argument('--inventory-name', dest='inventory_name', type=str, default=None, metavar='n', help='name of inventory to sync')
         parser.add_argument('--inventory-id', dest='inventory_id', type=int, default=None, metavar='i', help='id of inventory to sync')
-        parser.add_argument('--venv', dest='venv', type=str, default=None, help='absolute path to the AWX custom virtualenv to use')
```
AlanCoding, line 134: @wenottingham is removing a CLI option here fine with you?

awx/settings/development.py
```diff
@@ -95,7 +95,6 @@
 include(optional('/etc/tower/conf.d/*.py'), scope=locals())
 
 BASE_VENV_PATH = "/var/lib/awx/venv/"
-ANSIBLE_VENV_PATH = os.path.join(BASE_VENV_PATH, "ansible")
```
AlanCoding, line 98: After searching around the current code base, I figured we need to bite the bullet here.

This location doesn't exist. Keeping around the setting can only cause trouble.

rebeccahhh (commented):
I think the only potential conflict with my PR would be the changes in inventory.py or jobs.py. 
The association between a job and a project and a custom virtual env is used in one of the commands for the migration tool.
If this impacts that, we need to keep it in until the next release in the name of migration. I can run local tests to see if those two changes are compatible.

AlanCoding:
> The association between a job and a project and a custom virtual env is used in one of the commands for the migration tool.

I don't see this, could you point it out more specifically?

https://github.com/ansible/awx/pull/10090/files#diff-47596aff3e3aaf33cb7d2a4d5b459215dbbac36b9edb81e8dbba3a1f40587056

There are _job templates_ given in the output. I don't see your PR doing anything with jobs at all. Pre-migration jobs can have a non-blank custom venv, but I do not recall anyone making an argument for preserving that.

awx/main/management/commands/inventory_import.py
wenottingham: it doesn't do anything now, so yes, it's fine.

Also, isn't there the other issue as to whether this command needs removed entirely for the OCP use case?

AlanCoding: yes, that OCP issue for the command is hard. I figure, I won't lose sleep over dropping a minor option when it needs major work in this other area.

awx/settings/development.py
rebeccahhh: I agree, let's rip it out now

rebeccahhh (approved):
as long as QE is cool with it, let's do it

shanemcd (approved):

Issue #10355, opened 2021-06-04
Wrong text in Missing Resource tooltip for Job Details/Jobs List


AlexSCorey:
##### ISSUE TYPE
 - Bug Report

##### ENVIRONMENT
* AWX version: 19.1
* AWX install method: docker for mac
* Operating System:Mojave
* Web Browser:Chrome

##### STEPS TO REPRODUCE

Create a job template the uses an execution environment.  Run job.  delete execution environment

##### EXPECTED RESULTS

See below.
The `Missing resource` text appears but with the correct tooltip

##### ADDITIONAL INFORMATION

![image (1)](https://user-images.githubusercontent.com/39280967/120804756-05f62500-c513-11eb-96a1-b79a1166c634.png)

akus062381:
https://github.com/ansible/tower-qa/pull/6423
issue resolved

PR #10361, opened 2021-06-04
Update message for missing EE for Job pages

nixocio:
Update message for missing EE for Job pages

See: https://github.com/ansible/awx/issues/10355

image

image

### DIFF
--- a/awx/ui_next/src/components/JobList/JobListItem.jsx
+++ b/awx/ui_next/src/components/JobList/JobListItem.jsx
@@ -203,8 +203,8 @@ function JobListItem({
                 />
               )}
               <ExecutionEnvironmentDetail
-                virtualEnvironment={job.custom_virtualenv}
                 executionEnvironment={execution_environment}
+                verifyMissingVirtualEnv={false}
               />
               {credentials && credentials.length > 0 && (
                 <Detail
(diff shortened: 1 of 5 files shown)


awx/ui_next/src/components/JobList/JobListItem.jsx
```diff
@@ -202,9 +209,29 @@ function JobListItem({
                   dataCy={`job-${job.id}-project`}
                 />
               )}
-              <ExecutionEnvironmentDetail
-                virtualEnvironment={job.custom_virtualenv}
-                executionEnvironment={execution_environment}
+              <Detail
```
keithjgrant: Is there any reason we can't move this logic/tooltip into the ExecutionEnvironmentDetail and trigger it via a prop? It's seems strange to not use that component, especially when we have to repeat this code again in the JobDetail page

nixocio: @keithjgrant, I will take a look.

nixocio: @keithjgrant updated per suggestion.

mabashian (approved):

keithjgrant (approved):

akus062381:
https://github.com/ansible/tower-qa/pull/6423

akus062381 (commented):
all good!

PR #9879, opened 2021-04-12
WIP: implement an initial development environment for receptor-based clusters

ryanpetrello:
see: https://github.com/ansible/awx/issues/5151

This PR is a proof of concept for a development environment that provides an out-of-the-box multi-hop receptor cluster, including support for dynamic discovery and registration (and removal) of receptor mesh members.  This is *not* ready to merge, and does _not_ solve several of the open questions necessary to stabilize this (such as how this will integrate with the recently introduced operator, and how the generation and management of TLS certificates will be handled in an automated fashion).

That said, it does work with plain TCP, and you can use it to spin up a fairly simple cluster mesh, which acquires per-node capacity, and allows you to run jobs across a simple receptor mesh:

image

```
~ make docker-compose-cluster
```

```
~ docker ps -a
CONTAINER ID        IMAGE                            COMMAND                  CREATED             STATUS              PORTS                                                                                                                                                              NAMES
777f83b4bbbf        quay.io/ansible/awx-ee           "entrypoint receptor…"   21 minutes ago      Up 21 minutes                                                                                                                                                                          tools_receptor_3
aed6f4bea010        quay.io/ansible/awx-ee           "entrypoint receptor…"   21 minutes ago      Up 21 minutes                                                                                                                                                                          tools_receptor_2
be7d6b36b587        quay.io/ansible/awx-ee           "entrypoint receptor…"   21 minutes ago      Up 21 minutes                                                                                                                                                                          tools_receptor_1
74547e1f9a59        quay.io/ansible/awx-ee           "entrypoint receptor…"   21 minutes ago      Up 21 minutes       0.0.0.0:5555->5555/tcp                                                                                                                                             tools_receptor_hop
55d426d6274a        quay.io/awx/awx_devel:receptor   "/entrypoint.sh laun…"   21 minutes ago      Up 21 minutes       0.0.0.0:6899->6899/tcp, 0.0.0.0:7899-7999->7899-7999/tcp, 0.0.0.0:8013->8013/tcp, 0.0.0.0:8043->8043/tcp, 0.0.0.0:8080->8080/tcp, 22/tcp, 0.0.0.0:8888->8888/tcp   tools_awx_1
3179aa92453d        postgres:12                      "docker-entrypoint.s…"   21 minutes ago      Up 21 minutes       5432/tcp                                                                                                                                                           tools_postgres_1
[...]

PR #10365, opened 2021-06-05
Add receptor_node field to instance_group

ghjm:
##### SUMMARY

This PR adds a field to the instance_group model called receptor_node.  When set, the value of this field will be passed to `receptor_ctl.submit_work`, which will cause the local Receptor node to send the job to the named remote Receptor node instead of running it locally.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.2.0
```

##### ADDITIONAL INFORMATION

To try this in a single node Docker development environment, update your `tools/docker-compose/receptor.conf` and replace the `local-only` line with:

```
- tcp-peer:
    address: 172.18.0.1:2222
    redial: true
```
(This assumes you have the standard Docker setup where your host-only network is 172.18.0.0/24.)

Then, outside the container, run a Receptor instance with the following config:

```
---
- node:
    id: node1

- log-level: debug

- control-service:
    service: control
    filename: /tmp/receptor.sock

- tcp-listener:
    port: 2222

- work-command:
    worktype: local
    command: ansible-runner
    params: worker
    allowruntimeparams: true

- work-kubernetes:
    worktype: kubernetes-runtime-auth
    authmethod: runtime
    allowruntimeauth: true
    allowruntimepod: true
    allowruntimeparams: true

- work-kubernetes:
    worktype: kubernetes-incluster-auth
    authmethod: incluster
    allowruntimeauth: true
    allowruntimepod: true
    allowruntimeparams: true
```

Start the AWX development environment and you should see the the internal Receptor node `awx-1` connect and establish routing.  Create an instance group through the UI, let's say named `test`, and add the awx-1 instance to it.  Then inside the awx container, run `awx-manage dbshell` and set the Receptor node for the instance group as follows:

```
update main_instancegroup set receptor_node='node1' where name='test';
```

Now set a job template to run on this instance group and launch it.  It should run on your host, outside the container, via the node1 Receptor instance.

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -4885,6 +4885,7 @@ class InstanceGroupSerializer(BaseSerializer):
             "policy_instance_minimum",
             "policy_instance_list",
             "pod_spec_override",
+            "receptor_node",
             "summary_fields",
         )
 
--- a/awx/main/migrations/0145_instance_group_receptor_node.py
+++ b/awx/main/migrations/0145_instance_group_receptor_node.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('main', '0144_event_partitions'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='instancegroup',
+            name='receptor_node',
+            field=models.CharField(
+                blank=True,
+                default=None,
+                help_text='Receptor node to send all jobs to from this instance group',
+                max_length=40,
+                null=True,
+            ),
+        ),
+    ]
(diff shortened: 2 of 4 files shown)


ghjm:
Ping @wenottingham @shanemcd @chrismeyersfsu 

My question here is whether you would consider merging something like this.  I understand there are a bunch of concerns before this can be productized, like supportability, dealing with certificates, how the Receptor configuration gets written to the Tower instances, etc, etc.  But if there was a database-only field like this, it would give an extremely basic enablement of multi-node Receptor to people willing to live outside the support boundary.

chrismeyersfsu:
Don't love the single receptor-node per-instance group restriction.

@ryanpetrello wrote a POC that uses and abuses the per-instance version field to denote receptor-only nodes. We will likely go with an implementation that adds a field to the `Instance` record or introduce a new model. https://github.com/ansible/awx/pull/9879/files#diff-6b5e2004df5844765d0ce3cb503debb4fb980d19fd436c4d9516b49a6685fd06R146

PR #10343, opened 2021-06-03
add docs on debugging job event perf

chrismeyersfsu:
related to https://github.com/ansible/tower/issues/4985

PR #10373, opened 2021-06-07
Add OPTIONS documentation for new job limit feature

AlanCoding:
Looking at the docs and stuff from #10023

I'm sure this is somewhere else too, but this is the place that users should naturally expect it to be.


AlanCoding:
Looking over #10343, I don't think this mentions the limit query parameter.

I tested `/api/v2/organizations/?limit=5` and it does not seem to apply to the other views. I have not yet looked into the technical implementation of this.

chrismeyersfsu (approved):
This pagination works for job events for all unified job i.e. system job, project updates, inventory updates, ad hoc jobs, and jobs.

AlanCoding:
Those view names look like

https://github.com/ansible/awx/blob/7832639c25bd3fdbc3782d9337cd638994aa7bce/awx/api/urls/project_update.py#L24

So this needs to add files and do an include with those names too...

AlanCoding:
more view names

https://github.com/ansible/awx/blob/7832639c25bd3fdbc3782d9337cd638994aa7bce/awx/api/urls/inventory_update.py#L24

https://github.com/ansible/awx/blob/7832639c25bd3fdbc3782d9337cd638994aa7bce/awx/api/urls/ad_hoc_command_event.py#L10

https://github.com/ansible/awx/blob/7832639c25bd3fdbc3782d9337cd638994aa7bce/awx/api/urls/system_job.py#L14

AlanCoding:
wait a minute

https://github.com/ansible/awx/blob/7832639c25bd3fdbc3782d9337cd638994aa7bce/awx/api/urls/ad_hoc_command.py#L23

If this is the ad hoc command events view, then what's the other ad hoc command events view I linked?????????

AlanCoding:
Not wrapping _that_ bug into this https://github.com/ansible/awx/issues/10392

Using the name `ad_hoc_command_ad_hoc_command_events_list` for this work, as it is the consistently structured view.

AlanCoding:
Okay, I updated the PR and manually verified that the new text shows up for stuff like the project update events list.

chrismeyersfsu (approved):

PR #10375, opened 2021-06-07
Remove reload on manual logout

jakemcdermott:
##### SUMMARY
https://github.com/ansible/awx/issues/10383

We initiate a reload on manual logout, resulting in requests to `/config` that always return a 400 error due to being unauthenticated. 

From what I can tell, we might not actually need to do the reload since the `login` redirect will still happen once the logout request resolves.

### DIFF
--- a/awx/ui_next/src/contexts/Session.jsx
+++ b/awx/ui_next/src/contexts/Session.jsx
@@ -73,15 +73,12 @@ function SessionProvider({ children }) {
   const [authRedirectTo, setAuthRedirectTo] = useState('/');
 
   const logout = useCallback(async () => {
-    if (!isSessionExpired.current) {
-      history.replace('/');
-    }
     await RootAPI.logout();
     setSessionTimeout(0);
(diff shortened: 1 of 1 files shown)


awx/ui_next/src/contexts/Session.jsx
```diff
@@ -73,15 +73,12 @@ function SessionProvider({ children }) {
   const [authRedirectTo, setAuthRedirectTo] = useState('/');
 
   const logout = useCallback(async () => {
-    if (!isSessionExpired.current) {
-      history.replace('/');
-    }
```
jakemcdermott, line 78: talked w/ @marshmalien 

This code makes sure you land on the expected page when signing back in.

jakemcdermott:
https://github.com/ansible/awx/pull/10375#discussion_r646905240

jakemcdermott:
closing this PR in lieu of https://github.com/ansible/awx/pull/10379

Issue #10360, opened 2021-06-04
Activity Stream for setting deletion does not tell what setting was deleted


AlanCoding:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
If you delete a setting, then a user cannot reasonably determine what happened from the activity stream because no form of the setting key or label is reported.

##### STEPS TO REPRODUCE

Go to the API, set a setting to a test value.

Then issue a DELETE request to that setting's category.

##### EXPECTED RESULTS

I can see what happened in activity stream

##### ACTUAL RESULTS

Setting creation entry looks good

deletion entry looks like this:

![Screenshot from 2021-06-04 10-55-17](https://user-images.githubusercontent.com/1385596/120821679-b1a77100-c523-11eb-82c2-df3d791f005b.png)

##### ADDITIONAL INFORMATION

This primarily on the API. See `/api/v2/activity_stream/?order_by=-timestamp&page_size=2`

```json
{
    "count": 1451,
    "next": "/api/v2/activity_stream/?order_by=-timestamp&page=2&page_size=2",
    "previous": null,
    "results": [
        {
            "id": 1451,
            "type": "activity_stream",
            "url": "/api/v2/activity_stream/1451/",
            "related": {
                "actor": "/api/v2/users/2/"
            },
            "summary_fields": {
                "actor": {
                    "id": 2,
                    "username": "alan",
                    "first_name": "",
                    "last_name": ""
                }
            },
            "timestamp": "2021-06-04T14:54:58.073945Z",
            "operation": "delete",
            "changes": {
                "value": "[\"test_value\"]",
                "id": 142
            },
            "object1": "setting",
            "object2": "",
            "object_association": "",
            "action_node": "awx_1",
            "object_type": ""
        },
        {
            "id": 1450,
            "type": "activity_stream",
            "url": "/api/v2/activity_stream/1450/",
            "related": {
                "actor": "/api/v2/users/2/",
                "setting": "/api/v2/settings/jobs/"
            },
            "summary_fields": {
                "actor": {
                    "id": 2,
                    "username": "alan",
                    "first_name": "",
                    "last_name": ""
                },
                "setting": [
                    {
                        "category": "jobs",
                        "name": "AWX_ANSIBLE_CALLBACK_PLUGINS"
                    }
                ]
            },
            "timestamp": "2021-06-04T14:54:55.180050Z",
            "operation": "create",
            "changes": {
                "value": "[\"test_value\"]",
                "id": 142
            },
[...]

PR #10376, opened 2021-06-07
Changes settings revert all to send DELETE on individual endpoint rather than PATCHing

### LINKED ISSUE #3831: The UI will PATCH to all instead of DELETE when reverting settings category to defaults
##### ISSUE TYPE
 - Bug Report

##### COMPONENT NAME
 - UI

##### SUMMARY
The UI does not follow the intended API contract when reverting settings

##### ENVIRONMENT
* AWX version: all
* AWX install method: openshift, minishift, docker on linux, docker for mac, boot2docker
* Ansible version:  X.Y.Z
* Operating System:
* Web Browser:

##### STEPS TO REPRODUCE

go to the UI page `/#/configuration/system`

navigate to the logging sub-category

click "REVERT ALL TO DEFAULT"

##### EXPECTED RESULTS

DELETE to `/api/v2/settings/logging/`

##### ACTUAL RESULTS

PATCH to `/api/v2/settings/all/` with all of this data

https://gist.github.com/AlanCoding/034f9ec18fa7eb5037f221fa015937cc

This creates an enormous swarm of activity stream entries.

It also reverts all non-logging settings, which isn't what the user expects based on the dialog.

mabashian:
##### SUMMARY
Resolves https://github.com/ansible/awx/issues/3831

https://github.com/ansible/awx/issues/3831#issuecomment-501690495 is a good explanation of the state of the new UI as well as the old.  The Misc System settings were the most problematic part of this since that form pulled fields from both `/api/v2/settings/system` and `/api/v2/settings/authentication`.  In order to revert these fields with a DELETE request I needed to split this form up into Miscellaneous System and Miscellaneous Authentication.  The Activity Stream Settings also needed to be absorbed by the Miscellaneous System Settings.

All settings forms (except noted below) will now issue a DELETE request to revert all the fields for a particular settings subsection.

There was one form that I did not change the Revert All functionality on and that was the LDAP Settings form(s).  The UI splits these fields out into 5 different subforms and reverting one (issuing a DELETE on `/api/v2/settings/ldap`) would revert all the ldap settings which I think is undesirable.

This fix has the added benefit of cleaning up future activity stream entries for reverting an entire settings category.

I also consulted with @wenottingham and @gamuniz on some fields that were present in the api but missing in the ui and I added those.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI


AlanCoding:
I checked out this branch and poked around. Starting out I mucked with the jobs settings. Changing some, then reverting all.

Screenshot from 2021-06-08 10-17-16

Screenshot from 2021-06-08 10-17-23

This looks nice. I think there are good reasons to go with this.

The one qualifier I want to make here is that the PATCH will include all settings in the category ("jobs" category here). This leads to "creation" entries like this:

Screenshot from 2021-06-08 10-22-03

The default value of the "Default Project Update Timeout" is 0. So this is a junk entry. I do feel opinionated that we should address this. However, I think the proper place for the fix is API-side. I think we can get by without it for now, but we should file and issue for that if this merges.

Confirmed that LDAP categories behave like they always have. Those should be pretty stable and I'm not too worried about it.

I looked over the activity stream after mucking with some settings categories. There are a lot of entries, due to the default value issue I mentioned. Also, the deletion issues are pretty ugly, due to an open issue, #10360

AlanCoding (approved):

AlanCoding:
(you should get UI reviews too, I'm just considering the requests being made)

jakemcdermott (approved):

tiagodread (approved):
We are now making DELETE requests for all settings subforms properly and these events are being displayed as deleted in the activity stream.

PR #10377, opened 2021-06-07
Break out control plane EE as its own thing

kdelee:
We want only the control plane EE to be managed_by_tower=True, which will be the only one used by project updates

Jeff is going to do some more follow up things w/ RBAC and resolving the right EE for project updates

### DIFF
--- a/awx/main/management/commands/register_default_execution_environments.py
+++ b/awx/main/management/commands/register_default_execution_environments.py
@@ -68,23 +68,28 @@ class Command(BaseCommand):
                 sys.stderr.write("No registry credential type found")
                 sys.exit(1)
 
-            registry_cred, cred_created = Credential.objects.get_or_create(
-                name="Default Execution Environment Registry Credential", managed_by_tower=True, credential_type=registry_cred_type[0]
-            )
-
-            if cred_created:
-                print("'Default Execution Environment Credential' registered.")
-
             inputs = {
                 "host": options.get("registry_url"),
                 "password": options.get("registry_password"),
                 "username": options.get("registry_username"),
                 "verify_ssl": options.get("verify_ssl"),
             }
+            registry_cred, cred_created = Credential.objects.get_or_create(
+                name="Default Execution Environment Registry Credential",
+                managed_by_tower=True,
+                credential_type=registry_cred_type[0],
+                defaults={'inputs': inputs},
+            )
+
+            if cred_created:
+                changed = True
+                print("'Default Execution Environment Credential' registered.")
+
             for key, value in inputs.items():
-                if registry_cred.get_input(key) != value:
+                if not registry_cred.inputs.get(key) or registry_cred.get_input(key) != value:
                     registry_cred.inputs[key] = value
                     changed = True
+
             if changed:
                 registry_cred.save()
                 print("'Default Execution Environment Credential' updated.")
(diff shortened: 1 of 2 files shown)


awx/main/management/commands/register_default_execution_environments.py
```diff
@@ -84,21 +84,28 @@ def handle(self, *args, **options):
...
+                print("Updated 'Default Execution Environment Credential'")
+
+        # Create default globally available Execution Environments
+        for ee in reversed(settings.GLOBAL_JOB_EXECUTION_ENVIRONMENTS):
+            _, ee_created = ExecutionEnvironment.objects.update_or_create(name=ee["name"], image=ee["image"], credential=registry_cred)
+            if ee_created:
+                changed.append(True)
+                print(f"'{ee['name']}' Default Execution Environment registered.")
+
+        # Create the control plane execution environment that is used for project updates and system jobs
```
AlanCoding, line 114: not used for system jobs, right?

but is used for inventory updates, for all the default cloud sources like ec2, azure, etc.

awx/main/management/commands/register_default_execution_environments.py
```diff
@@ -84,21 +84,28 @@ def handle(self, *args, **options):
...
-            changed = True
-            print("Default Execution Environment(s) registered.")
+            changed.append(True)
+
+            if not cred_created:
+                print("Updated 'Default Execution Environment Credential'")
+
+        # Create default globally available Execution Environments
+        for ee in reversed(settings.GLOBAL_JOB_EXECUTION_ENVIRONMENTS):
+            _, ee_created = ExecutionEnvironment.objects.update_or_create(name=ee["name"], image=ee["image"], credential=registry_cred)
```
AlanCoding: so here, we lose the `managed_by_tower` flag for the list of EEs from the pre-creation list in settings.

but let's keep in mind that this is how the default EE gets grabbed:

https://github.com/ansible/awx/blob/68f214c2bec56af99129c3c8971cb7eaab8bd9ca/awx/main/utils/execution_environments.py#L9-L12

I thought we said that the list would grow some means of declaring 1 out of the list as the job default (`DEFAULT_EXECUTION_ENVIRONMENT`). Maybe that's still a TODO for @jbradberry

awx/main/management/commands/register_default_execution_environments.py
kdelee: yeah I'll fix that comment....
but FWIW  currently the project update resolve EE method is the ONLY one that keys on managed_by_tower=True

awx/main/management/commands/register_default_execution_environments.py
```diff
@@ -84,21 +84,28 @@ def handle(self, *args, **options):
...
+            if ee_created:
+                changed.append(True)
+                print(f"'{ee['name']}' Default Execution Environment registered.")
+
+        # Create the control plane execution environment that is used for project updates and system jobs
+        control_plane_ee = settings.CONTROL_PLANE_EXECUTION_ENVIRONMENT
+        _cp_ee, cp_created = ExecutionEnvironment.objects.update_or_create(
+            name="Control Plane Execution Environment", defaults={'image': control_plane_ee, 'managed_by_tower': True, 'credential': registry_cred}
```
kdelee: @AlanCoding you missed this one-- this is the new and ONLY managed_by_tower=True EE that will be used for project updates

awx/main/management/commands/register_default_execution_environments.py
AlanCoding: You're right in that project update goes straight to the global jobs default

https://github.com/ansible/awx/blob/2ce276455ad8bf252607759f8070b9f5f8f10838/awx/main/models/projects.py#L192

but all the other models fallback to the same thing

https://github.com/ansible/awx/blob/3e5ee9d57a2bad3f78141e5a17842fd9483141f5/awx/main/models/mixins.py#L489

if there's not an EE specifically set (which I would consider to be the default mode of operation)

awx/main/management/commands/register_default_execution_environments.py
AlanCoding: This may not have been your intent, but I think this would have _jobs_ use the EE named "Control Plane Execution Environment" as their default.

And I like that, we should go with that. What's the harm in jobs using the control plane EE?

awx/main/management/commands/register_default_execution_environments.py
```diff
@@ -68,35 +68,54 @@ def handle(self, *args, **options):
...
-                print('Updated Default Execution Environment Credential')
-
-        for ee in reversed(settings.DEFAULT_EXECUTION_ENVIRONMENTS):
-            _, created = ExecutionEnvironment.objects.update_or_create(
-                name=ee["name"], defaults={"image": ee["image"], "managed_by_tower": True, "credential": registry_cred}
-            )
-
-        if created:
+            print("Control Plane Execution Environment registered.")
+        elif _cp_ee.image != control_plane_ee_image or (registry_cred and _cp_ee.credential_id != registry_cred.id):
```
kdelee: @shanemcd I think we need to think about what it means to re-run the command except w/o any credentials...does that mean you want us to null out any existing credential? I think I need another clause here and above if `registry_cred` is null but `_cp_ee.credential` is not null

awx/main/management/commands/register_default_execution_environments.py
AlanCoding: Scratch my last comment, project/job/inventory default is in flux with #10378

awx/main/management/commands/register_default_execution_environments.py
jbradberry: To copy over info discussed in #10378, the most recently created EE is the one that gets prioritized.  Keep that in mind when specifying the order of EEs in the variable.

awx/main/management/commands/register_default_execution_environments.py
jbradberry: Based on my understanding of what @shanemcd said, Project Updates will continue to be the only job type that gets set to use the control plane EE.

awx/main/management/commands/register_default_execution_environments.py
```diff
@@ -68,35 +68,54 @@ def handle(self, *args, **options):
...
+                _this_ee.image = ee["image"]
+                _this_ee.credential = registry_cred
+                _this_ee.save()
+                changed = True
+                print(f"'{ee['name']}' Default Execution Environment updated.")
+
+        # Create the control plane execution environment that is used for project updates and system jobs
+        control_plane_ee_image = settings.CONTROL_PLANE_EXECUTION_ENVIRONMENT
+        _cp_ee, cp_created = ExecutionEnvironment.objects.get_or_create(
+            name="Control Plane Execution Environment", defaults={'image': control_plane_ee_image, 'managed_by_tower': True, 'credential': registry_cred}
```
AlanCoding: > Control Plane Execution Environment

What about string translations?

kdelee: @shanemcd and I think we'll just have to deal with that at a later time

shanemcd (approved):

Issue #10363, opened 2021-06-04
Default Execution Environment not being update on settings page after EE deletion


nixocio:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Default Execution Environment not being update on settings page after EE deletion. 

##### STEPS TO REPRODUCE

1. Create a new EE
2. Go to settings page - Set as Default Execution Environment
3. Delete the EE from the step 1.
4. Id of the EE remains on the settings page.

##### EXPECTED RESULTS

Settings to be updated as per the EE deletion.

##### ACTUAL RESULTS

![image](https://user-images.githubusercontent.com/9053044/120859402-242f4580-c552-11eb-96ac-a20b87b20cd0.png)

![image](https://user-images.githubusercontent.com/9053044/120859429-2e514400-c552-11eb-9222-082c0d8e1745.png)

yagomarques:
It's working, tested manually   :heavy_check_mark: 

Before delete: 
![image](https://user-images.githubusercontent.com/18562328/121940551-9e1ab800-cd24-11eb-9a95-f23b8e303f62.png)

After delete:
![image](https://user-images.githubusercontent.com/18562328/121940810-e20dbd00-cd24-11eb-8c03-d6917b3be59e.png)

PR #10378, opened 2021-06-07
Control plane EE

jbradberry:
##### SUMMARY

Per discussion, we are now going to have privileged execution environments for use on the control plane, and these will not be allowed to be edited or deleted.

related https://github.com/ansible/tower/issues/5016

##### TODO

- [x] reinstate the restrictive RBAC for `managed_by_tower=True` EEs
- [x] convert any EEs automatically installed prior into `managed_by_tower=False` EEs
- [x] change the resolver so that ordinary jobs do not get a managed EE, and project updates only get a managed EE
- [x] allow sysadmin users to edit the `pull` field for managed EEs 
- [x] automatically disassociate EEs that get deleted from the `DEFAULT_EXECUTION_ENVIRONMENT` setting #10363

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API


AlanCoding:
Could we merge this with #10377 from @kdelee and go from there?

jbradberry:
I don't see why this needs to be merged with another PR before landing.  These changes are pretty distinct.

awx/main/utils/execution_environments.py
```diff
@@ -6,10 +6,14 @@
 from awx.main.models.execution_environments import ExecutionEnvironment
 
 
+def get_control_plane_execution_environment():
+    return ExecutionEnvironment.objects.filter(organization=None, managed_by_tower=True).first()
```
AlanCoding, line 10: Perhaps it's already established as a TODO, but since the pre-load EEs are concretely going to be a list with >1 elements, we need a selection criteria other than `.first()`

awx/main/models/projects.py
```diff
@@ -185,11 +185,11 @@ def clean_credential(self):
 
     def resolve_execution_environment(self):
         """
-        Project updates, themselves, will use the default execution environment.
+        Project updates, themselves, will use the control plane execution environment.
         Jobs using the project can use the default_environment, but the project updates
         are not flexible enough to allow customizing the image they use.
         """
-        return get_default_execution_environment()
+        return get_control_plane_execution_environment()
```
AlanCoding, line 192: This makes it so project updates run with the "control plane EE".

Should ec2 inventory updates also run with the "control plane EE"? Why or why not?

awx/main/utils/execution_environments.py
jbradberry: Based on the ordering criteria I imposed on the EE model way back when, `.first()` translates to 'most recent', which I think is the correct priority we should give to installed EEs.

awx/main/models/projects.py
jbradberry: I don't have a strong opinion about it, but presumably there was a reason we hadn't had that in place previously.  @shanemcd ?

awx/main/utils/execution_environments.py
AlanCoding: intended use (near-term) is that someone passes a list of 4 EEs to create in the installer

the `awx-manage` command loops over those and creates them

what you're saying would indicate that jobs will use the _final_ EE in the list as the default. This is high-value information, so much so that I will @Spredzy since he may be implementing the installer side of this.

jbradberry: Yes, though the idea originally was that newer EEs installed when updating would be preferentially used over prior ones.

awx/main/models/projects.py
shanemcd: - Inventory updates don't run on the control plane. 
- Users are able to choose which EE they want to use at the inventory source level.

awx/api/views/__init__.py
```diff
@@ -695,17 +695,30 @@ class TeamAccessList(ResourceAccessList):
...
 
+    def update(self, request, *args, **kwargs):
+        instance = self.get_object()
+        fields_to_check = ['name', 'description', 'organization', 'image', 'credential']
+        if instance.managed_by_tower and request.user.can_access(models.ExecutionEnvironment, 'change', obj):
+            for field in fields_to_check:
+                left = getattr(instance, field, None)
+                right = request.data.get(field, None)
+                if left is not None and right is not None and left != right:
+                    raise PermissionDenied(_("Only the 'pull' field can be edited for managed execution environments."))
```
shanemcd, line 719: ❤️

awx/main/signals.py
```diff
@@ -623,6 +624,12 @@ def deny_orphaned_approvals(sender, instance, **kwargs):
         approval.deny()
 
 
+@receiver(pre_delete, sender=ExecutionEnvironment)
```
shanemcd, line 627: 🙌

shanemcd (approved):

awx/api/views/__init__.py
```diff
@@ -695,17 +695,30 @@ class TeamAccessList(ResourceAccessList):
...
     swagger_topic = "Execution Environments"
 
+    def update(self, request, *args, **kwargs):
+        instance = self.get_object()
+        fields_to_check = ['name', 'description', 'organization', 'image', 'credential']
+        if instance.managed_by_tower and request.user.can_access(models.ExecutionEnvironment, 'change', instance):
+            for field in fields_to_check:
+                left = getattr(instance, field, None)
+                right = request.data.get(field, None)
+                if left is not None and right is not None and left != right:
```
jbradberry: This might be too much, I may need to do just `left != right` in order to avoid things like setting the Org.

fosterseth (approved):

awx/main/signals.py
jbradberry: Turns out that this isn't sufficient :(

beeankha (approved):

awx/main/signals.py
jbradberry: fixed.

nixocio:
I got this code locally, and we still allow the managed by tower EE to be deleted by system admin as per summary fields.

image

PR #10379, opened 2021-06-07
Avoid reload on manual logout

jakemcdermott:
##### SUMMARY
https://github.com/ansible/awx/issues/10383

Instead of reloading, set the `authRedirectTo` context variable to "/logout" and handle it as a special case when routing


marshmalien (approved):
Looks good! Tested this out locally with a few login redirect flows using a regular user and a social auth user:
1. Session expiration  => Redirect to last URL
2. Manual logout via the session warning or toolbar => Redirect home
3. Debounced direct URL access  => Redirect to direct URL
4. Regular login => Redirect home

akus062381:
@jakemcdermott nice job, this issue is fixed. all good to merge.

Issue #2817, opened 2018-11-28
cannot specify "not__" host filter in smart inventory


hiyokotaisa:
##### ISSUE TYPE
 - Bug Report

##### COMPONENT NAME

 - API
 - UI

##### SUMMARY

Normally, I can specify the host filter to exclude the specific host in API as below:
```
 http://<AWX_URL>/api/v2/hosts/?not__name__icontains=test
```
But when I created the smart inventory, we cannot specify the `not.name.icontains:test` filter.
Is it the normal behavior of smart inventory? Is there any way to exclude the specific host on smart inventory?

##### ENVIRONMENT
* AWX version: 2.1.0

##### STEPS TO REPRODUCE

1. Register hosts like "test1", "test2" and "test3".
2. Create smart inventory and open smart host filter.
3. set `not.name.icontains:test3` but it shows `No records matched your search..`

##### EXPECTED RESULTS
"test1" and "test2" will appear on smart inventory.

##### ACTUAL RESULTS
No records matched.

mabashian:
I spent a few moments looking into this one.  I don't think that operators other than `equal` are supported by the api when specifying a host filter on a smart inventory.  Additionally, the UI doesn't currently support any keys that come before the field in any capacity.  For example, in a regular (non host filter) search something like `not.name:foo` is not currently supported by the UI.  In this case, the field would be `name` and since `not` comes before it, it's not currently supported.  The UI will convert that to `?search=not.name:foo` which is generally not useful.

There are two things that would need to happen in order to support this:

1) The api would need to add support for operators other than equal wrt the smart inventory host filter so that POSTing something like:

```
{
    "name": "test inv",
    "description": "",
    "organization": 1,
    "kind": "smart",
    "host_filter": "not__name__icontains=test",
    "variables": "",
    "insights_credential": null
}
```

would be valid.

2) The UI will need to add support for operators like `or` and `not` which appear before the field in the search string

All that being said, I'd say this is an enhancement to search since none of this was supported in the initial implementation.

jdh239:
How long until this is available in a Tower general release?

boaz-raz:
Any updates on this please?

chrismeyersfsu:
This would be doable on "classic" django sort of queries, we support it as a generic filter.  Due to performance reasons, it's not possible to support this on ansible facts (json type).

LaurencelleJ:
[...]

PR #10380, opened 2021-06-07
Do not show `not` as choice for Advanced Search on Smart Inventory

nixocio:
Do not show `not` set type as a  choice for Advanced Search on Smart Inventory since
this feature is not implemented on the API side yet.

image

See: https://github.com/ansible/awx/issues/2817

### DIFF
--- a/awx/ui_next/src/components/Lookup/HostFilterLookup.jsx
+++ b/awx/ui_next/src/components/Lookup/HostFilterLookup.jsx
@@ -111,13 +111,13 @@ const buildSearchColumns = () => [
 
 function HostFilterLookup({
   helperTextInvalid,
-
   isValid,
   isDisabled,
   onBlur,
   onChange,
   organizationId,
   value,
+  enableNegativeFiltering,
 }) {
   const history = useHistory();
   const location = useLocation();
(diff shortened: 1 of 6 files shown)


awx/ui_next/src/components/Lookup/HostFilterLookup.jsx
```diff
@@ -111,13 +111,13 @@ const buildSearchColumns = () => [
...
 function HostFilterLookup({
   helperTextInvalid,
-
   isValid,
   isDisabled,
   onBlur,
   onChange,
   organizationId,
   value,
+  displaySetTypeNot,
```
jakemcdermott: I don't love the naming, as I'm not sure I'd be able to guess (correctly) what it did without looking at the code.

Maybe `enableNegativeFiltering` or `negativeFilteringEnabled`?

jakemcdermott (approved):

awx/ui_next/src/components/Lookup/HostFilterLookup.jsx
nixocio: I did not love that name either. Thanks. I will update.

nixocio:
I triggered and e2e test run.

tiagodread (approved):
The users will still be able to filter hosts using `not__` from hosts list and then, try to create a smart inventory which won't work. This PR is avoiding users to filter by `not__` only from the smart inventory lookup filter view while this feature is not fully implemented yet by API.

PR #10385, opened 2021-06-07
Adds version context processor back in to fix api browser doc link

mabashian:
##### SUMMARY
Here's what it looks like on devel now (note the URL in the bottom left):

Screen Shot 2021-06-07 at 7 54 28 PM

Here's what it looks like after the change (note the URL in the bottom left):

Screen Shot 2021-06-07 at 7 53 00 PM

I dropped this in as it was before the removal of the old UI.  I believe the new UI needs access to some of these variables as well to force assets to be refetched after upgrade.

Also of note: I have no idea what I'm doing with django so please help me to become educated if I've done something silly here.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/settings/defaults.py
+++ b/awx/settings/defaults.py
@@ -270,6 +270,7 @@ TEMPLATES = [
                 'django.template.context_processors.tz',
                 'django.contrib.messages.context_processors.messages',
                 'awx.ui.context_processors.csp',
+                'awx.ui.context_processors.version',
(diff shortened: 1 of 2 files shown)


jakemcdermott (approved):
Ah, looks like this context processor is still used by the dynamic api docs.  :+1: lgtm

jakemcdermott:
regate

PR #10394, opened 2021-06-08
Pull brand name from default.strings.json

mabashian:
##### SUMMARY
We were pulling this string from `variables.js`.

There were several places where the brand name was not being pulled dynamically - this PR addresses that.  This changes the way we set the document title a little bit.  Since we needed access to a dynamic string, I went down the JS route to set this.

Screen Shot 2021-06-08 at 2 44 14 PM

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/public/index.html
+++ b/awx/ui_next/public/index.html
@@ -37,7 +37,6 @@
       name="description"
       content="AWX"
     />
-    <title>AWX</title>
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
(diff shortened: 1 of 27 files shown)


jakemcdermott (approved):

akus062381:
@mabashian - I built an ec2 instance using your branch. I see your change in all applicable spots, but all of them look like this:
image

Can you clarify? I realize this is dynamic- but shouldn't there be a name there? Thanks! :-)

tiagodread:
recheck

akus062381 (approved):
verified and looks good!

PR #10405, opened 2021-06-09
Renaming tower inventory source and credentials [WIP]

amolgautam25:
##### SUMMARY

Added Migration for Renaming Field in the following Model : `credential, credentialtype , InventorySource , InventoryUpdate`

##### ISSUE TYPE

 - Bugfix Pull Request
 

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.2.0
```

##### ADDITIONAL INFORMATION

What Did i Do : 
--- Renamed `Ansible Tower` with `Red Hat Ansible Platform` in `inventorysource` and `inventoryupdate` model.
--- Renamed `Ansible Tower` with `Red Hat Ansible Platform` in help_text references
--- Added `CONTROLLER*` variables in Credential Initialization. Refer to https://github.com/ansible/tower/issues/5002
--- Added `CONTROLLER*` environment variables in `awx/main/tests/data/inventory/plugins/tower/env.json`
--- Refactored `AWXKIT` code with the new `CONTROLLER*` variables
--- Modified the test cases with respect to the above change


awx/api/serializers.py
```diff
@@ -1409,11 +1409,11 @@ def validate(self, attrs):
 
 class ExecutionEnvironmentSerializer(BaseSerializer):
     show_capabilities = ['edit', 'delete', 'copy']
-    managed_by_tower = serializers.ReadOnlyField()
+    managed_by_controller = serializers.ReadOnlyField()
```
wenottingham: From a practical perspective, this would need to merge after @jbradberry's changes for EEs.

awx/main/migrations/0145_auto_20210609_1652.py
```diff
@@ -0,0 +1,63 @@
...
+                    ('file', 'File, Directory or Script'),
+                    ('scm', 'Sourced from a Project'),
+                    ('ec2', 'Amazon EC2'),
+                    ('gce', 'Google Compute Engine'),
+                    ('azure_rm', 'Microsoft Azure Resource Manager'),
+                    ('vmware', 'VMware vCenter'),
+                    ('satellite6', 'Red Hat Satellite 6'),
+                    ('openstack', 'OpenStack'),
+                    ('rhv', 'Red Hat Virtualization'),
+                    ('controller', 'Ansible Controller'),
```
wenottingham: This should be 'Red Hat Ansible Automation Platform'. While it is talking to the controller API, it is the inventory for the platform deployment..

Also changing the choice name has larger API fallout (collection, tests, etc.) - cc @AlanCoding

awx/main/migrations/0145_auto_20210609_1652.py
```diff
@@ -0,0 +1,63 @@
...
```
wenottingham: Same here - this should be Red Hat Ansible Automation Platform, and 'controller' has downstream implications.

awx/main/models/credential/__init__.py
```diff
@@ -1004,15 +1004,15 @@ def create(self):
 ManagedCredentialType(
     namespace='tower',
```
amolgautam25: Any thoughts on this , if I should change the `namespace` value ?

awx/main/models/credential/__init__.py
```diff
@@ -1039,11 +1039,11 @@ def create(self):
...
-            'TOWER_HOST': '{{host}}',
-            'TOWER_USERNAME': '{{username}}',
-            'TOWER_PASSWORD': '{{password}}',
-            'TOWER_VERIFY_SSL': '{{verify_ssl}}',
-            'TOWER_OAUTH_TOKEN': '{{oauth_token}}',
+            'CONTROLLER_HOST': '{{host}}',
+            'CONTROLLER_USERNAME': '{{username}}',
+            'CONTROLLER_PASSWORD': '{{password}}',
+            'CONTROLLER_VERIFY_SSL': '{{verify_ssl}}',
+            'CONTROLLER_OAUTH_TOKEN': '{{oauth_token}}',
```
AlanCoding, line 1057: Why not just repeat these with the `TOWER_*` names? Avoids any versioning problems. These are injectors, it does us no harm to put more stuff in.

awx/main/migrations/0145_auto_20210609_1652.py
AlanCoding: > and 'controller' has downstream implications.

Would "awx" have been a better choice? Myself, I think so, yes.

But very soon, users are going to be very frustrated with knowing what to use between tower / awx / controller.

Making any change is going to bring pain. But changing some things but not others, or worse, changing inconsistently, will bring even greater pain. There were valid arguments about the messaging, but I think that discussion has closed because we've already merged other user-facing stuff that has made a selection. To use another selection now would be bad.

wenottingham: There's two questions:
- do we change the API namespace/setting from `tower`
- if so, what to

Assuming we *do* change, it, the standard for inventory and credential types is to use product names - satellite6, rhv, openstack, etc. So the change would be to 'aap'.

awx/main/models/credential/__init__.py
amolgautam25: I have made the required changes.

awx/main/models/credential/__init__.py
```diff
@@ -1008,21 +1008,23 @@ def create(self):
...
-                'help_text': ugettext_noop('The Ansible Tower base URL to authenticate with.'),
+                'help_text': ugettext_noop('Red Hat Ansible Automation Platform base URL to authenticate with.'),
             },
             {
                 'id': 'username',
                 'label': ugettext_noop('Username'),
                 'type': 'string',
-                'help_text': ugettext_noop('The Ansible Tower user to authenticate as.' 'This should not be set if an OAuth token is being used.'),
+                'help_text': ugettext_noop(
+                    'Red Hat Ansible Automation Platform user to authenticate as.' 'This should not be set if an OAuth token is being used.'
```
rebeccahhh: I would change this to "user id" for clarity

amolgautam25: Made the changes, will be pushed in next commit.

amolgautam25:
recheck

awx/main/models/credential/__init__.py
wenottingham: maybe `username`? Don't want to specify a PK here. (I don't think anyone would do that, but you never know...)

amolgautam25:
recheck

amolgautam25:
recheck

awx/main/models/credential/__init__.py
rebeccahhh: that's totally fair ^ I'm not tied to User ID in any way, I vote for clarity

amolgautam25: I have changed it , will be pushed in the next commit.

amolgautam25:
recheck

AlanCoding (approved):

awx/main/models/credential/__init__.py
AlanCoding: Other discussion dives further into this question

https://github.com/ansible/awx/pull/10405#discussion_r649351719

I don't think it should be changed to "controller", since that doesn't align with the big-picture direction. If possible, it might be best to leave it alone until we can introduce a new platform-wide Credential Type to phase it out. I'd rather not change it, just to change it again.

wenottingham: In general, this should only be changed with renaming the credential 'kind' and the inventory source type at the same time.

wenottingham:
Looks generally good if we are holding off on renaming the API types themselves. Still will require docs updates.

rebeccahhh (commented):
I would squash your four commits into one detailed commit that explains the change

awx/main/tests/functional/test_credential.py
```diff
@@ -74,12 +74,15 @@
...
+    import pdb
+
+    pdb.set_trace()
     assert sorted(CredentialType.defaults.keys()) == [
         'aim',
         'aws',
         'azure_kv',
         'azure_rm',
-        'centrify_vault_kv',
+        #        'centrify_vault_kv',
```
amolgautam25: I had to comment out this line for the test to pass. Any pointers as to why this might be causing the issue ?

awx/main/tests/functional/test_credential.py
```diff
@@ -96,11 +99,12 @@ def test_default_cred_types():
         'satellite6',
         'scm',
         'ssh',
-        'thycotic_dsv',
+        #        'thycotic_dsv',
```
amolgautam25: I had to comment out this line as well for the test to pass. Any pointers as to why this might be causing the issue ?

amolgautam25:
recehck

amolgautam25:
recheck

awx/main/migrations/0147_auto_20210610_2314.py
```diff
@@ -0,0 +1,55 @@
+# Generated by Django 2.2.16 on 2021-06-10 23:14
```
rooftopcellist: @amolgautam25 We should rename this migration, maybe something like `0147_rename_inv_sources.py` or something along those lines

amolgautam25: I have made the required changes , will be pushed in next commit.

fosterseth:
I suppose we need to change this too? 

https://github.com/ansible/awx/blob/b92c5076a25589c9195fb201100b6daccc141872/awx_collection/plugins/modules/credential.py#L295

also if we make that change, can you please change this too? Seems I overlooked it during my collections renaming PR

https://github.com/ansible/awx/blob/b92c5076a25589c9195fb201100b6daccc141872/awx_collection/tools/roles/template_galaxy/templates/README.md.j2#L15

AlanCoding:
Let me echo what I think I heard from conversations with @wenottingham (which carries a great amount of extrapolation / speculation)

If we are going to rename the `InventorySource.source` field `tower` -> `controller`, then there is a strong expectation that the credential type `namespace` field is also renamed `tower` -> `controller`.

In other words, change both of these, or don't change either.

In that case, I agree we should change them all. We don't want to key off "tower". This is an API change, but this release carries lot of other minor API changes like that.

Changing the namespace of the credential type does pose technical issues, but it can be done with some novel migration code.

kdelee:
we are going to need this for our tests, I'm driving blind here but 99% sure this should be it if the kind is still `tower`
```
diff --git a/awxkit/awxkit/api/pages/credentials.py b/awxkit/awxkit/api/pages/credentials.py
index 5f78b63f8b..e809fd22b2 100644
--- a/awxkit/awxkit/api/pages/credentials.py
+++ b/awxkit/awxkit/api/pages/credentials.py
@@ -79,7 +79,7 @@ credential_type_name_to_config_kind_map = {
     'amazon web services': 'aws',
     'container registry': 'registry',
     'ansible galaxy/automation hub api token': 'galaxy',
-    'ansible tower': 'tower',
+    'red hat ansible automation platform': 'tower',
     'google compute engine': 'gce',
     'insights': 'insights',
     'openshift or kubernetes api bearer token': 'kubernetes',
```

jbradberry:
I believe @kdelee is correct.

amolgautam25:
recheck

awxkit/awxkit/cli/format.py
```diff
@@ -11,30 +11,30 @@ def add_authentication_arguments(parser, env):
     auth = parser.add_argument_group('authentication')
     auth.add_argument(
         '--conf.host',
-        default=env.get('TOWER_HOST', 'https://127.0.0.1:443'),
+        default=env.get('CONTROLLER_HOST', 'https://127.0.0.1:443'),
```
kdelee: @AlanCoding you mentioned that we would still be accepting the old vars, and I see you are still going to lay them down in the injector for the credential. But here I see us ONLY looking for the new vars

rooftopcellist: Let's cherry-pick [this commit](https://github.com/ansible/awx/commit/3c34242eeb2e144a46471c268dc4648aef69a395) onto this PR.  It checks for both the `TOWER_` and `CONTROLLER_` env vars and makes tests pass.

amolgautam25: I have made the required changes.

amolgautam25:
recheck

amolgautam25:
I have made this change : https://github.com/ansible/awx/pull/10405#issuecomment-862625145 
@kdelee @jbradberry

amolgautam25:
recheck

amolgautam25:
recheck

awx/main/migrations/0148_rename_inv_sources_inv_updates.py
```diff
@@ -0,0 +1,55 @@
...
+                    ('openstack', 'OpenStack'),
+                    ('rhv', 'Red Hat Virtualization'),
+                    ('tower', 'Red Hat Ansible Automation Platform'),
+                    ('insights', 'Red Hat Insights'),
+                ],
+                default=None,
+                max_length=32,
+            ),
+        ),
+    ]
```
AlanCoding, line 90: This still needs the data migrations we talked about, right?

amolgautam25: I have updated the migration. 
Let me know if any changes are to be done.

AlanCoding:
Conflicts are because a version of this already merged, closing.

Issue #10400, opened 2021-06-08
Advanced search by groups not working in hosts list


tiagodread:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
While working on https://github.com/ansible/awx/issues/9232 I realized that the user can not use the OR operator with the key `Groups` the host_filter is not adding `or__`  I've found this in host lists, however, this may happen in other places with different keys.

##### STEPS TO REPRODUCE

1. Go to host list
2. Do an advanced search using the key `Groups` type the name of the group

##### EXPECTED RESULTS

To list all hosts within the group

##### ACTUAL RESULTS

The OR operator is not defined and the user can't filter by groups using advanced search

![image](https://user-images.githubusercontent.com/12464858/121255680-12b3a980-c882-11eb-8c13-036a8eee8e43.png)

this issue was reported here initially https://github.com/ansible/awx/issues/9389
Note that https://github.com/ansible/awx/pull/10347 is not addressing this.

mabashian:
OK so this is pretty complicated and touches https://github.com/ansible/awx/issues/7834.  I think I'd like to propose a simplification of this interface under certain circumstances.  In the case of the host list there are a number of `related_search_fields` that are listed in the `Lookup Type` dropdown.  These fields can be searched on but only in two ways:

1) Directly `groups=1`
2) Via the search modifier `groups__search=somestring`

As best as I can tell, these keys do not support a prefix (and/or/not) when using the search modifier _and_ we don't currently support directly querying these fields (see item 1).

So, my proposal is to clear/disable the `Set Type` and `Lookup Type` dropdowns whenever one of these  `related_search_fields` is selected.  This way, users don't think that they can apply modifiers to these search keys (the UI is basically ignoring these fields for these fields at this point - it just uses the search modifier every time):

![adv_search](https://user-images.githubusercontent.com/9889020/121391645-9d48e700-c91c-11eb-9027-cf989475152a.gif)

Also of note: it seems like the default behavior is for these values to be `or`'d when you have multiple:
`groups__search=foo&groups_search=bar` returns hosts in `foo` or in `bar`.

So here's what the interface will look like when you select a related search key:

<img width="1393" alt="Screen Shot 2021-06-09 at 1 20 03 PM" src="https://user-images.githubusercontent.com/9889020/121400298-6b884e00-c925-11eb-9da2-9fcfc345f585.png">

but these fields will be enabled for search keys which can use them (ie non related search keys).

mabashian:
[...]

PR #10416, opened 2021-06-10
Addresses bug where advanced search by groups wasn't working on host list

mabashian:
##### SUMMARY
link #10400 

OK so this got a bit more complicated than I wanted it to but I think on the whole this is a nice improvement.  In the initial bug report it was noted that we were ignoring the `or`/`not` operators when searching on a related key (in this case it was groups).  We were also ignoring the modifier at the end (icontains, exact, startswith, etc).  Since we were always defaulting to `__search` for these types of keys this was actually valid.  The API doesn't like us attempting to do something like `?or__groups__search=foo` or `?groups__search__icontains` because they aren't valid.  So, I changed the UX a little bit.  A user can still do a fuzzy search on a related key but the prefix is disabled in this case:

Screen Shot 2021-06-10 at 2 10 02 PM

I changed the third dropdown (specifically for related keys) to contain the following options: `search`, `id`, `name__icontains` which I think would be the three most general cases.  `id` and `name__icontains` do allow for prefixing (`or`, `not`, `and`).

This should make searching on related keys a little bit easier and a little bit more flexible since name and id weren't possible in the UI before.

Once place where this is a little different is the host filter in the smart inventory form.  Using `__search` currently throws an error when the user attempts to save.  I believe this is an API bug and will file it as such but for now I prevent users from attempting to use `__search` by removing it as an option for this particular list.  When this bug gets fixed in the API we can remove this logic.

I also noticed that there was a bug where `or__` search terms were being converted to `and` when using the Smart Inventory button on the host list so I fixed that.

Here's the fix in action:

host_search

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/components/DataListToolbar/DataListToolbar.jsx
+++ b/awx/ui_next/src/components/DataListToolbar/DataListToolbar.jsx
@@ -40,6 +40,7 @@ function DataListToolbar({
   qsConfig,
   pagination,
   enableNegativeFiltering,
+  enableRelatedFuzzyFiltering,
 }) {
   const showExpandCollapse = onCompact && onExpand;
   const [isKebabOpen, setIsKebabOpen] = useState(false);
@@ -92,6 +93,7 @@ function DataListToolbar({
               onShowAdvancedSearch={onShowAdvancedSearch}
               onRemove={onRemove}
               enableNegativeFiltering={enableNegativeFiltering}
+              enableRelatedFuzzyFiltering={enableRelatedFuzzyFiltering}
             />
           </ToolbarItem>
           {sortColumns && (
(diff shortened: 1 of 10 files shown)


mabashian:
recheck

nixocio (approved):
💯 This PR looks great.

tiagodread:
Running e2e

nixocio:
> Running e2e

I forgot to mention that I also triggered an e2e test run. @tiagodread

tiagodread (approved):
Verified that:

- basic search filters works
- advanced search type `and/or/not` works for `default `keys
- advanced search type `and/or/not` works for `related_search_fields` keys when `lookup type` is `name__icontains` or `id`
- type is disabled and tooltip ins displayed if fuzzy search is enabled in `related_search_fields` keys
- smart inv button is disabled and contains help text in tolltip if the fuzzy search is enabled 
- smart inv can be created using or operator in hosts list or in smart inv form

Issue #10213, opened 2021-05-17
Give better error details when receptor drops a job


AlanCoding:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Errors from the receptor network (the most obvious ones, details below) are surfaced in the API in a way that gives no really useful information.

##### ENVIRONMENT
* AWX version: X.Y.Z
* AWX install method: openshift, minishift, docker on linux, docker for mac, boot2docker
* Ansible version:  X.Y.Z
* Operating System:
* Web Browser:

##### STEPS TO REPRODUCE

Start a job with a sleep task.

Shell into the task container and run `receptorctl work cancel` manually.

Inspect the API detail of those jobs

##### EXPECTED RESULTS

Some kind of `job_explanation`, but failing that, some information in `result_traceback` which is minimally readable and actionable.

##### ACTUAL RESULTS

Get error and this traceback.

```
Traceback (most recent call last):
  File "/awx_devel/awx/main/tasks.py", line 1380, in run
    res = receptor_job.run()
  File "/awx_devel/awx/main/tasks.py", line 2911, in run
    return self._run_internal(receptor_ctl)
  File "/awx_devel/awx/main/tasks.py", line 2969, in _run_internal
    raise RuntimeError(detail)
RuntimeError: Running: PID 6122
```

snooyen:
I've been trying to play with container groups in 19.2.2, but my jobs have been failing with no easily diagnosable error logs. At best I've been able to find things like


`
awx-ee: ERROR2021/08/12 16:04:25 Read error in control service: read unix /var/run/receptor/receptor.sock-u003e@: use of closed network connection
awx-task: WARNING  [556237baf394484a904f40f1558db897] awx.main.dispatch job 18 (error) encountered an error (rc=None), please see task stdout for details.
`

AlanCoding:
Since this was filed we have added a lot more exception handling around this spot, we have a test for what is described, and expectations may change around this anyway.

PR #10420, opened 2021-06-11
Add reaper task to release old receptor work units

### LINKED ISSUE #9154: [execution-environments] In some cases, AWX fails to release job that had error
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Run a job that had an error due to some receptor configuration problem (more detailed specifics TBD).

Make sure there are no jobs running in the system.

Now shell into the server and check `receptorctl work list`

Expected result: nothing listed

Actual result:

```
bash-4.4$ receptorctl work list
{'2ntorWZp': {'Detail': 'Error creating pod: At least one container must be '
                        'named worker',
              'ExtraData': {'Command': '',
                            'Image': '',
                            'KubeConfig': '',
                            'KubeNamespace': 'eon9jxmrgo',
                            'KubePod': '',
                            'KubeTLSCAData': '',
                            'KubeVerifyTLS': True,
                            'Params': '',
                            'PodName': ''},
              'State': 3,
              'StateName': 'Failed',
              'StdoutSize': 0,
              'WorkType': 'ocp'},
...
```

The `result_traceback` in the job reports the error. However, as of https://github.com/ansible/awx/pull/8598, the code that throws the error isn't wrapped in a try-except such that the release command will still be ran.

fosterseth:
related https://github.com/ansible/awx/issues/9154

##### SUMMARY

Adds `work_unit_id` field to UnifiedJob. When a job is launched via receptor, this field will be populated

```python3
result = receptor_ctl.submit_work(worktype=self.work_type, payload=sockout.makefile('rb'), params=self.receptor_params)
self.unit_id = result['unitid']
self.task.update_model(self.task.instance.pk, work_unit_id=result['unitid'])
```

Adds a reaper task that runs once per minute:

1. get a current list of receptor work units by running `work list`
2. get a list of awx jobs that have a `work_unit_id` in that list, and with status not in `ACTIVE_STATES`
3. release those work units via `work release`

Drops the `RECEPTOR_UNIT_ID` from job.job_env

##### SIDE NOTE
I'm unsure of the expectation that all work units must be cleaned up. For now, work units in receptor are tied directly to an awx job, and these receptor work units should be released after an awx job completes. However, it's also possible that we launch receptor work that aren't associated with awx jobs -- e.g. to gather meta data / capacity of the execution node. 

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.2.0
```


awx/main/tasks.py
```diff
@@ -472,6 +472,33 @@ def cluster_node_heartbeat():
                 logger.exception('Error marking {} as lost'.format(other_inst.hostname))
 
 
+@task(queue=get_local_queuename)
+def awx_receptor_workunit_reaper():
+    """
+    When an AWX job is launched via receptor, files such as status, stdin, and stdout are created
+    in a specific receptor directory. This directory on disk is an random 8 character string, e.g. qLL2JFNT
+    This is also called the work Unit ID in receptor, and is used in various receptor commands,
+    e.g. "work results qLL2JFNT"
```
AlanCoding, line 481: Such good docs! Let's make sure we have this content in docs.

awx/main/tasks.py
```diff
@@ -472,6 +472,33 @@ def cluster_node_heartbeat():
...
+    This task will call "work release" on each of these work units to clean up the files on disk.
+    """
+    if not settings.RECEPTOR_RELEASE_WORK:
+        return
+    logger.debug("Checking for unreleased receptor work units")
+    receptor_ctl = get_receptor_ctl()
+    receptor_work_list = receptor_ctl.simple_command("work list")
+
+    unit_ids = [id for id in receptor_work_list]
+    jobs_with_unreleased_receptor_units = UnifiedJob.objects.filter(receptor_unit_id__in=unit_ids).exclude(status__in=ACTIVE_STATES)
```
AlanCoding: Would it be bad if someone else was using receptor in the cluster to do something irrelevant and we killed their jobs?

awx/main/tasks.py
```diff
@@ -472,6 +472,33 @@ def cluster_node_heartbeat():
...
+    if AWX crashes during a job's execution, the work release command is never issued to begin with.
+    As such, this periodic task will obtain a list of all receptor work units, and find which ones
+    belong to AWX jobs that are in a completed state (status is cancelled, error, or succeeded).
+    This task will call "work release" on each of these work units to clean up the files on disk.
+    """
+    if not settings.RECEPTOR_RELEASE_WORK:
+        return
+    logger.debug("Checking for unreleased receptor work units")
+    receptor_ctl = get_receptor_ctl()
+    receptor_work_list = receptor_ctl.simple_command("work list")
```
AlanCoding, line 493: would it be better to filter this to inactive / finished tasks?

awx/main/tasks.py
```diff
@@ -472,6 +472,33 @@ def cluster_node_heartbeat():
...
+
+    unit_ids = [id for id in receptor_work_list]
+    jobs_with_unreleased_receptor_units = UnifiedJob.objects.filter(receptor_unit_id__in=unit_ids).exclude(status__in=ACTIVE_STATES)
+    for job in jobs_with_unreleased_receptor_units:
+        logger.debug("{} is not active, reaping receptor work unit".format(job.log_format))
```
AlanCoding: The job is marked as inactive, but if there's a bug or network issue, the remote process could still be active. In that case, we would prefer to tell receptor to kill the remote work. But I'm okay if you want to avoid active tasks for this instead.

awx/main/models/unified_jobs.py
```diff
@@ -717,6 +717,9 @@ class Meta:
         editable=False,
         help_text=_("The version of Ansible Core installed in the execution environment."),
     )
+    receptor_unit_id = models.CharField(
+        max_length=255, blank=True, default=None, editable=False, null=True, help_text=_("The Receptor work unit ID associated with this job.")
```
AlanCoding, line 721: :+1:

awx/main/tasks.py
fosterseth: good question. I am kind of using the awx side as the source of truth for the state of things -- if awx says the job is done, then I am fine just killing/removing whatever is on receptor side. This might be helpful too in cases where a receptor work unit is hanging out in pending

awx/main/tasks.py
fosterseth: This reaper should only be touching receptor work units that are associated with awx jobs, so even if receptor is also being used for non-awx related stuff, those files should be untouched.

fosterseth:
recheck

awx/main/tasks.py
fosterseth: > we would prefer to tell receptor to kill the remote work

yes the reaper will do this. This reaper will only consider inactive AWX jobs, but it will kill ACTIVE or INACTIVE receptor work units

awx/settings/defaults.py
```diff
@@ -427,6 +427,7 @@ def IS_TESTING(argv=None):
     'gather_analytics': {'task': 'awx.main.tasks.gather_analytics', 'schedule': timedelta(minutes=5)},
     'task_manager': {'task': 'awx.main.scheduler.tasks.run_task_manager', 'schedule': timedelta(seconds=20), 'options': {'expires': 20}},
     'k8s_reaper': {'task': 'awx.main.tasks.awx_k8s_reaper', 'schedule': timedelta(seconds=60), 'options': {'expires': 50}},
+    'receptor_reaper': {'task': 'awx.main.tasks.awx_receptor_workunit_reaper', 'schedule': timedelta(seconds=60)},
```
AlanCoding, line 430: once per minute seems aggressive. Won't this run on every node in the cluster each time? I worry about scaling up to large numbers of nodes, like 30 nodes.

AlanCoding (approved):

awx/main/migrations/0147_unifiedjob_receptor_unit_id.py
```diff
@@ -0,0 +1,20 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('main', '0146_add_insights_inventory'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='unifiedjob',
+            name='receptor_unit_id',
```
chrismeyersfsu: I have this lurking fear that receptor will get renamed in the future. Maybe we can call it `work_unit_id` ? (I'll let you make the call, the rename wound is fresh)

awx/main/tasks.py
chrismeyersfsu: include the work unit id in this debug log message. This message is going to be super helpful in the future.

awx/main/tasks.py
```diff
@@ -472,6 +472,33 @@ def cluster_node_heartbeat():
...
+        receptor_ctl.simple_command(f"work release {job.receptor_unit_id}")
```
chrismeyersfsu: I'm guessing this is a fire and forget command. Otherwise, I'd say log the "return" value.

awx/settings/defaults.py
chrismeyersfsu: I don't think we support that many control nodes. When we switch to having receptor execution nodes, this logic will need to be expanded then we can reassess. At 30 seconds, that matches the k8s reaper so I am fine with it.

chrismeyersfsu (approved):
I like this targeted solution. We may still have to deal with jobs that think they are `running` in awx but have no work unit in receptor. (i.e. a receptor node dies & the tower control node dies too). But that can be done in another PR.

awx/settings/defaults.py
AlanCoding: We won't need to change anything as we introduce purely execution nodes, will we?

My entire concern was that these `receptorctl` actions apply to the whole network. If you only ran _one_ instance of this task once per period that would accomplish the objective just fine (it's not some queue specific to this control node, if I understand). We just have no mechanism to have a whole-cluster periodic task.

AlanCoding:
> We may still have to deal with jobs that think they are running in awx but have no work unit in receptor.

We do have a "receptorctl kill" test case, where we cancel the job via receptorctl while the task is in progress. That's where #10213 came from.

awx/settings/defaults.py
chrismeyersfsu: The scheduled jobs function + the analytics functionality are cluster-wide periodic tasks. We trigger a periodic beat per-node, use a lock to make sure the exec is exclusive, update a database timestamp to ensure only 1 execution during the desired period.

> We won't need to change anything as we introduce purely execution nodes, will we?

It's subtle, but we rely on local information from the dispatcher to rectify execution. What this means is that the reaper must run on each control node. More specifically, the reaper can only reap jobs that have the same control node that the reaper is running on. 
`receptorctl work list` must target a single receptor execution node. For a single control node, we will have to make multiple calls to `receptorctl work list` to get all of the `exexecution_node` work units.

We are also thinking about dropping the need for a single-specific control node "babysitter". That actually might simplify the reaping process. I'm unsure.

AlanCoding: okay, you're talking about a call `receptorctl work list --node=<node_id>` call. Yes, that makes it self-obvious that it's a per-node call.

PR #10425, opened 2021-06-11
introduced a pre-flight check for postgres 12

sarabrajsingh:
##### SUMMARY

Introduced a pre-flight check to check the version of Postgres 12 before the application starts - https://github.com/ansible/tower/issues/4940

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.2.0
```

### DIFF
--- a/awx/__init__.py
+++ b/awx/__init__.py
@@ -34,6 +34,7 @@ else:
     from django.db.backends.base import schema
     from django.db.models import indexes
     from django.db.backends.utils import names_digest
+    from django.db import connection
 
 
 if HAS_DJANGO is True:
@@ -148,6 +149,11 @@ def manage():
     # Now run the command (or display the version).
     from django.conf import settings
     from django.core.management import execute_from_command_line
+
+    # enforce the postgres version is equal to 12. if not, then terminate program with exit code of 1
+    if (connection.pg_version // 10000) < 12:
+        sys.stderr.write("Postgres version 12 is required\n")
+        sys.exit(1)
 
     if len(sys.argv) >= 2 and sys.argv[1] in ('version', '--version'):  # pragma: no cover
         sys.stdout.write('%s\n' % __version__)
--- a/awx/__init__.py
+++ b/awx/__init__.py
@@ -151,9 +151,10 @@ def manage():
     from django.core.management import execute_from_command_line
 
     # enforce the postgres version is equal to 12. if not, then terminate program with exit code of 1
-    if (connection.pg_version // 10000) < 12:
-        sys.stderr.write("Postgres version 12 is required\n")
-        sys.exit(1)
+    if not MODE == 'development':
+        if (connection.pg_version // 10000) < 12:
+            sys.stderr.write("Postgres version 12 is required\n")
+            sys.exit(1)
 
     if len(sys.argv) >= 2 and sys.argv[1] in ('version', '--version'):  # pragma: no cover
         sys.stdout.write('%s\n' % __version__)


wenottingham:
This is likely the wrong place for this - you'd end up with a system that upgraded and then failed with no back out.

The goal would be to catch this at installation/upgrade time, not in the application itself.

sarabrajsingh:
@wenottingham I am introducing another PR for a check at install-time in the [ansible/automation-platform-collection](https://github.com/ansible/automation-platform-collection) repo. 

@shanemcd and I decided to put a check in the uwsgi bootstrap to have a check at the application level. the use case would be an awx pod in k8s that is pointed to a postgres lower than v12, or a botched operator installation where postgres wasn't upgraded for an instance of awx. what do you think?

wenottingham:
> @wenottingham I am introducing another PR for a check at install-time in the ansible/automation-platform-collection repo.

:+1: 

As for this, it could be OK, but I'm still worried about how this will manifest and how the user will recover. Will this lead to a reconciliation loop that never ends in the operator?

sarabrajsingh:
@wenottingham the reconciliation loop wouldn't occur at the operator instance, but rather at the pod instance level, as the program exits with an exit code of 1 upon failure. the operator will try to maintain the pod instance state (that is, `n` number of pods at any given time), therefore the end-user should look at the pod logs if they're stuck in a CrashLoopBackOff state. what do you think?

sarabrajsingh:
recheck

awx/__init__.py
```diff
@@ -149,6 +150,13 @@ def manage():
     from django.conf import settings
     from django.core.management import execute_from_command_line
 
+    # enforce the postgres version is equal to 12. if not, then terminate program with exit code of 1
+    version = connection.cursor().connection.server_version
```
AlanCoding: Might we prefer `pg_version` because it's cached?

https://github.com/django/django/blob/ca9872905559026af82000e46cde6f7dedc897b6/django/db/backends/postgresql/base.py#L335

AlanCoding: which looks to be `connection.pg_version`

wenottingham:
> @wenottingham the reconciliation loop wouldn't occur at the operator instance, but rather at the pod instance level, as the program exits with an exit code of 1 upon failure. the operator will try to maintain the pod instance state (that is, `n` number of pods at any given time), therefore the end-user should look at the pod logs if they're stuck in a CrashLoopBackOff state. what do you think?

That works. How does that manifest if you try to connect to the instance - hanging connection, or connection refused?

awx/__init__.py
sarabrajsingh: done. i think that'll fix the unit tests from failing as well

sarabrajsingh:
recheck

awx/__init__.py
```diff
@@ -149,6 +150,12 @@ def manage():
     from django.conf import settings
     from django.core.management import execute_from_command_line
 
+    # enforce the postgres version is equal to 12. if not, then terminate program with exit code of 1
```
rooftopcellist, line 153: This comment should probably say `equal to or greater than` to reflect the code.  Newer versions of postgresql for external db's shouldn't be an issue, though it isn't strictly supported.

rooftopcellist:
@sarabrajsingh I tend to think that this db version check in the code is unnecessary overhead.  If anywhere, I think this check could be in the installer, in this case the `awx-operator`.

shanemcd:
@rooftopcellist That would be a rather poor experience. Read the other comments on this PR for context.

shanemcd (approved):

Issue #10399, opened 2021-06-08
Workflow template- details tab shows missing EE resource when not required


akus062381:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
When creating a workflow job template, the only field required is the name field. But when the workflow is launched, and the user views the details tab of that job, the EE shows as a missing resource for the WFJT. I then went back to my WFJT, and edited and assigned a specific EE. I then launched the WFJT again. In the job details page, the EE still showed as a missing resource, even though when the user looks at the template itself, it shows the EE that you assigned when you edited the template.

##### STEPS TO REPRODUCE

1. Create a WFJT. Only fill in required fields, leave all others blank.
2. Access the visualizer tab of the WFJT and add one node with a JT. Save.
3. Access the schedules tab of that same WFJT and create a schedule to run the WFJT. 
4. Access the Jobs list and wait until the job is triggered and finishes. Expand that job's hidden details line. You will see the warning icon next to the EE saying it is missing. Then click into that job's details page, and next to EE you will see Missing Resource. 

##### EXPECTED RESULTS

The WFJT should have the default EE assigned to it if the user doesn't pick one. Especially with it not being a required field.

##### ACTUAL RESULTS

User has the option not to associate an EE when creating a WFJT, and when they look at the job results after that template launches a job, it shows there is a missing resource.
Even if the user does designate a specific EE for the WFJT, the job details page still shows the EE as a missing resource.
##### ADDITIONAL INFORMATION

![image](https://user-images.githubusercontent.com/63685497/121250949-530c2b00-c874-11eb-88a7-28e6b6f7f9d0.png)
Launch this workflow, and visit the details screen for that job:
![image](https://user-images.githubusercontent.com/63685497/121251083-7931cb00-c874-11eb-9198-f57dc75d5acf.png)

nixocio:
There is a new field `resolved_environment` that will display what execution environment will be used to run this job, even the execution environment was not explicit selected when creating the workflow, for example. This one seems to a replacement for this missing execution environment.

```
        "resolved_environment": {
            "id": 2,
            "name": "Control Plane Execution Environment",
            "description": "",
            "image": "quay.io/ansible/awx-ee:0.3.0"
        },

```

See: https://github.com/ansible/awx/issues/10327

I will attempt to solve both issues.

nixocio:
[...]

PR #10426, opened 2021-06-11
Make sure that the resolved EE gets attached to new workflow jobs

jbradberry:
##### SUMMARY
Make sure that the resolved EE gets attached to new workflow jobs.  This is solely for future reference and UI consistency, since workflows are not actually executed in EE images.

related #10399 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/models/workflow.py
+++ b/awx/main/models/workflow.py
@@ -525,7 +525,12 @@ class WorkflowJobTemplate(UnifiedJobTemplate, WorkflowJobOptions, SurveyJobTempl
         )
 
     def create_unified_job(self, **kwargs):
+        from awx.main.signals import disable_activity_stream
+
         workflow_job = super(WorkflowJobTemplate, self).create_unified_job(**kwargs)
+        with disable_activity_stream():
+            workflow_job.execution_environment = workflow_job.resolve_execution_environment()
+            workflow_job.save()
         workflow_job.copy_nodes_from_original(original=self)
         return workflow_job
 


jbradberry:
recheck

jbradberry:
recheck

jbradberry:
recheck

AlanCoding:
Would this result in the `default_environment` from the organization of the WFJT overriding the EE set directly on JTs in the WFJT?

That seems like wrong behavior.

jbradberry:
@AlanCoding no, it wouldn't.  Or at least it shouldn't.

jbradberry:
I've double-checked, this does not cause the behavior that you were worrying about.

Setup:
- "Demo Job Template", explicit EE = "AWX EE 0.3.0"
- "Workflow Template", contains one node that points to "Demo Job Template", no explicit EE
- "Second-Order Workflow Template", contains one node that points to "Workflow Template", explicit EE = "Foo"

Result:
Screenshot_2021-06-11 AWX

jbradberry:
If the explicit EE is unset on "Demo Job Template", all three jobs correctly wind up with EE "Foo" used.

jbradberry:
To be clearer, the workflow job isn't part of the resolver hierarchy, it's the workflow job template (and only when the EE is explicitly set on the workflow template).

AlanCoding:
What about

 - "Demo Job Template", no explicit EE
 - "Workflow Template", contains one node that points to "Demo Job Template", no explicit EE, in "foo" organization
 - The "foo" organization has a "foo-org-EE"

Launch the "Workflow Template", would it run with "foo-org-EE"?

> To be clearer, the workflow job isn't part of the resolver hierarchy, it's the workflow job template (and only when the EE is explicitly set on the workflow template).

Yeah, I would like that the WFJT organization doesn't come into play. If this is the intention (just that workflow job gets EE from its respective WFJT) then can you just add `execution_environment` to this list?

https://github.com/ansible/awx/blob/3063073395143a019ad0918e397ab7a30f740c82/awx/main/models/workflow.py#L385-L391

These are fields copied over from the WFJT to the workflow job on launch.

jbradberry:
> Launch the "Workflow Template", would it run with "foo-org-EE"?

That's pretty unambiguously 'yes'.  But to confirm:

Screenshot_2021-06-11 AWX(1)

jbradberry:
> Yeah, I would like that the WFJT organization doesn't come into play. If this is the intention (just that workflow job gets EE from its respective WFJT) then can you just add `execution_environment` to this list?

I'm having a bit of trouble parsing what you actually mean by this.  Let's chat.

jbradberry:
Based on https://github.com/ansible/awx/issues/10399#issuecomment-859828641, it seems like this isn't the direction to go in.  Closing.

PR #10429, opened 2021-06-11
Make ExecutionEnvironment awxkit class able to handle null values in …

AlanCoding:
This is how the `Project` class does it, and I can confirm it makes `factories.execution_environment(organization=None)` work correctly.


jbradberry (approved):

AlanCoding:
recheck

PR #10436, opened 2021-06-11
Prefer installer defaults over user-defined global EEs

AlanCoding:
The installer knows better than the user

### DIFF
--- a/awx/main/tests/functional/utils/test_execution_environments.py
+++ b/awx/main/tests/functional/utils/test_execution_environments.py
@@ -1,0 +1,46 @@
+import pytest
+
+from django.conf import settings
+from django.test.utils import override_settings
+
+from awx.main.models.execution_environments import ExecutionEnvironment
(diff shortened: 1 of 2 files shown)


jbradberry (approved):
I wonder if we should use the name field instead, but sure.

AlanCoding:
You know, we should also add `organization=None, managed_by_tower=False` to this filter.

beeankha (approved):

kdelee (approved):
This is good...we can't know that it is still around, but if it is, we prefer it

AlanCoding:
As kind of a followup to this, I filed https://github.com/ansible/awx/issues/10442. That's not something for the near-term, but as we manage multiple versions in future releases this will become more important.

ping @wenottingham

AlanCoding:
saw that a release was in progress https://github.com/ansible/awx/pull/10441

PR #10443, opened 2021-06-14
Execution environments are meaningless for workflows

jbradberry:
##### SUMMARY

Execution environments are meaningless for workflows, so, remove them from the API endpoints for workflows.  Also, tear out
the WFJT.execution_environment step in the resolver.  If we want that to be a thing, it ought to be a .default_environment instead.

related #10399

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -729,11 +729,12 @@ class UnifiedJobTemplateSerializer(BaseSerializer):
 
         if self.is_detail_view:
             resolved_ee = obj.resolve_execution_environment()
-            summary_fields['resolved_environment'] = {
-                field: getattr(resolved_ee, field, None)
-                for field in SUMMARIZABLE_FK_FIELDS['execution_environment']
-                if getattr(resolved_ee, field, None) is not None
-            }
+            if resolved_ee is not None:
+                summary_fields['resolved_environment'] = {
+                    field: getattr(resolved_ee, field, None)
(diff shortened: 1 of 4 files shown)


awx/api/serializers.py
```diff
@@ -3515,6 +3519,7 @@ class Meta:
 
     def get_related(self, obj):
         res = super(WorkflowJobSerializer, self).get_related(obj)
+        res.pop('execution_environment', None)  # EEs aren't meaningful for workflows
```
AlanCoding, line 3522: It should be impossible to set this anyway, but sure.

AlanCoding:
I'll bet we need awx collection changes too.

awx/api/serializers.py
jbradberry: I want to say that is so, but for workflow jobs intermediate ones might get it anyway due to being triggered within tasks.py.

AlanCoding (approved):

wenottingham:
@jbradberry @AlanCoding what is preventing merging this? Are there UI fixes needed?

nixocio:
> @jbradberry @AlanCoding what is preventing merging this? Are there UI fixes needed?

@wenottingham I will send a PR to remove the UI part of this now. I just tested this one locally

jbradberry:
@nixocio I should reopen this as a PR against the release branch instead.

jbradberry:
The release PR has merged.  Closing.

Issue #10137, opened 2021-05-07
Not able to import inventory to AWX from host


trinesh17:
##### ISSUE TYPE
 - Bug Report

##### COMPONENT NAME

 - API
 - UI
 - Installer

##### SUMMARY
I installed AWX using awx-operator. 
Issue1: I didn't get what password to give inside the awx-task container when prompts for sudo password.
Issue2: Getting this error FileNotFoundError: [Errno 2] No such file or directory: 'podman' while importing the inventory from host file
##### ENVIRONMENT
* AWX version: 19
* AWX install method: operator, developer environment
* AWX deployment target: kubernetes
* Operating System: Centos 7.9
* Web Browser:

##### STEPS TO REPRODUCE

Login to the awx-task container and run the below command
awx-manage inventory_import --inventory-name inventory --source /var/lib/awx/projects/inventory
My file is the source given and the inventory is already created in AWX

##### EXPECTED RESULTS

Inventory is imported successfully

##### ACTUAL RESULTS

    1.390 INFO     Updating inventory 4: inventory
2021-05-07 07:32:35,559 DEBUG    [-] awx.analytics.job_lifecycle inventoryupdate-21 created
    1.469 INFO     Reading Ansible inventory source: /var/lib/awx/projects/inventory
Traceback (most recent call last):
  File "/usr/bin/awx-manage", line 8, in <module>
    sys.exit(manage())
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/__init__.py", line 164, in manage
    execute_from_command_line(sys.argv)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/management/commands/inventory_import.py", line 872, in handle
[...]

Issue #1823, opened 2018-04-30
[Feature Request] Allow to create inventory source from a manual project


jean-christophe-manciot:
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
 - web

##### SUMMARY
As of 1.0.6, it is not possible to import a local hosts file and use it as an inventory.

##### ENVIRONMENT
* AWX version: tag 1.0.6
* AWX install method: docker/docker-compose on linux
* Ansible version: v2.5.2 (from PyPi)
* Operating System: Ubuntu 17.10
* docker-ce ```18.04.0~ce~3-0~ubuntu``` (from official docker PPA)
* docker==3.3.0
* docker-compose==1.21.1
* requests==2.18.1

##### STEPS TO REPRODUCE

- Add a project folder below the projects base folder containing a hosts file such as:
```
[all:vars]
nms_mgt_ip_address=172.21.0.1
remote_domain_name=actionmystique.net

[firewalls]
ASAv1 ansible_host=172.21.154.231

[spines]
NX_OSv_9k_Spine_31 ansible_host=172.21.114.153 
NX_OSv_9k_Spine_32 ansible_host=172.21.114.153

[mgt]
IOSv_L2_10 ansible_host=172.21.100.30

[nx_osv_9k:children]
spines

[asav:children]
firewalls

[cloud:children]
firewalls
spines
mgt
```
- Add this manual project in the GUI with Projects/+ADD.
- Add inventory in the GUI with Inventories/+ADD
- Add source --> Sourced from a project
- Choose an inventory file --> /(project root)

##### EXPECTED RESULTS
It should be possible to import a local hosts file and use it as an inventory.

##### ACTUAL RESULTS
![error loading hosts file from manual project](https://user-images.githubusercontent.com/13176858/39424242-f036cfe4-4c75-11e8-8dd5-24f7ab0b8220.png)

AlanCoding:
I don't think this would be a major problem to allow, with simple qualifiers:

 - we will not try to detect changes in the source tree on the file system
 - user will be responsible for defining the update behavior on their own (like Update on Launch)

PsychoSid:
Did this not used to be allowed ? Sure I used this prior to v1.0.5.

AlanCoding:
I don't think so, but I don't think it would have worked. There is more than 1 place where we need to add some special `if` clauses, because ordinarily we need a project sync before running the inventory update, and in this case we specifically need to not do that.

[...]

PR #10445, opened 2021-06-14
Make the inventory_import command respect the EE credential

jbradberry:
##### SUMMARY
related #10137 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/management/commands/inventory_import.py
+++ b/awx/main/management/commands/inventory_import.py
@@ -2,12 +2,10 @@
 # All Rights Reserved.
 
 # Python
-from base64 import b64encode
 import json
 import logging
 import os
 import re
-import stat
 import subprocess
 import sys
 import time
@@ -79,24 +77,11 @@ class AnsibleInventoryLoader(object):
             bargs.extend(['-e', '{0}={1}'.format(key, value)])
         ee = get_default_execution_environment()
         bargs.extend([ee.image])
-        cred = ee.credential
-        if cred.has_inputs(field_names=('host', 'username', 'password')):
-            path = functioning_dir(self.source)
-            with open(path + '/auth.json', 'w') as authfile:
-                os.chmod(authfile.name, stat.S_IRUSR | stat.S_IWUSR)
-
-                host = cred.get_input('host')
-                username = cred.get_input('username')
-                password = cred.get_input('password')
-                token = "{}:{}".format(username, password)
-                auth_data = {'auths': {host: {'auth': b64encode(token.encode('utf-8')).decode('utf-8')}}}
-                authfile.write(json.dumps(auth_data, indent=4))
-            bargs.append(f'--authfile={authfile.name}')
-        else:
-            raise RuntimeError('Please recheck that your host, username, and password fields are all filled.')
-        pull = ee.pull
-        if pull:
-            bargs.append(f'--pull={pull}')
+        authpath = ee.build_authfile(functioning_dir(self.source))
+        if authpath is not None:
+            bargs.append(f'--authfile={authpath}')
+        if ee.pull:
+            bargs.append(f'--pull={ee.pull}')
 
         bargs.extend(['ansible-inventory', '-i', self.source])
         bargs.extend(['--playbook-dir', functioning_dir(self.source)])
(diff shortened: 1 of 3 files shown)


awx/main/management/commands/inventory_import.py
```diff
@@ -75,7 +76,27 @@ def get_base_args(self):
...
+                password = cred.get_input('password')
+                token = "{}:{}".format(username, password)
+                auth_data = {'auths': {host: {'auth': b64encode(token.encode('utf-8')).decode('utf-8')}}}
+                authfile.write(json.dumps(auth_data, indent=4))
+            bargs.append(f'--authfile={authfile.name}')
+        else:
+            raise RuntimeError('Please recheck that your host, username, and password fields are all filled.')
+        pull = ee.pull
+        if pull:
+            bargs.append(f'--pull={pull}')
```
kdelee: are you fearful of refactoring where-ever else we do this similar thing to use a function so we can re-use it here? Is the other one mangled up with other code that its hard to pick it out?

kdelee (commented):
left a question....if its too much of a pain to re-use where ever else we do a similar thing I can approve

awx/main/management/commands/inventory_import.py
jbradberry: Sure, I think it could be refactored a bit.

rebeccahhh: yeah if you want to refactor with the stuff I wrote we could probably consolidate nicely

awx/main/management/commands/inventory_import.py
```diff
@@ -75,7 +75,14 @@ def get_base_args(self):
         bargs.extend(['-v', '{0}:{0}:Z'.format(self.source)])
         for key, value in STANDARD_INVENTORY_UPDATE_ENV.items():
             bargs.extend(['-e', '{0}={1}'.format(key, value)])
-        bargs.extend([get_default_execution_environment().image])
+        ee = get_default_execution_environment()
+        bargs.extend([ee.image])
+        authpath = ee.build_authfile(functioning_dir(self.source))
```
AlanCoding: really going to need some cleanup.

I have found `tempfile.NamedTemporaryFile` to be preferable. Normally, this would not be an option because uWSGI processes are somewhat long-lived, but in this case `awx-manage` invocations will exit when the work is done, so cleanup would happen at the right time.

jbradberry: This is functionality shared with code in tasks.py.  Would a temp file be appropriate there as well?  Should the temp file still wind up in the relevant directory in each case, or /tmp/ or somewhere else like that?  And what are the implications of trying to do that in a container filesystem?

AlanCoding: `self.source` is provided by the user by the `--source` option, right? That is quite different from `tasks.py`.

The user may not want you to mess with the directory where their inventory is.

I don't see any need to put the authfile anywhere near the inventory file. You probably want to _avoid_ putting it in the container mounted directories, if possible.

Zokormazo:
What happens on k8s? Will this try to exec podman inside the pod?

AlanCoding:
@Zokormazo that's tracked in https://github.com/ansible/awx/issues/10137

This PR is attempting to address private registries.

Zokormazo:
> @Zokormazo that's tracked in #10137
> 
> This PR is attempting to address private registries.

My concern is more around inventory updates and https://github.com/ansible/awx/blob/983082ade231343a031bbdcd7f93e3a0cec0d54b/awx/main/management/commands/inventory_import.py#L883-L884 this comment

AlanCoding:
That method `perform_update` isn't named well. It has `data` passed into it, which is the output of `ansible-inventory`. This comes from the remote executor in k8s. It doesn't run the inventory plugin logic, it only takes the data and saves it to the database.

Zokormazo:
thanks for the clarification, and sorry for the noise

shanemcd (approved):

AlanCoding:
@wenottingham as long as this isn't making it in now, or is incomplete, we may want to re-consider priority of https://github.com/ansible/awx/issues/1823, since it would be the next most obvious way to import a static file.

jbradberry:
Closing to re-do against the release branch.

Issue #10099, opened 2021-05-04
Demo Job Template run fails; source control update error: 'overlay' is not supported over overlayfs


lo78cn:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Demo Job Template run fails; source control update error: 'overlay' is not supported over overlayfs 

##### ENVIRONMENT
* AWX version: 19.1.0
* AWX install method: docker on linux
* Ansible version:  2.9.14
* Operating System: Debian 10.9
* Web Browser: Brave version 1.23.75 Chromium: 90.0.4430.93 (Official Build) (x86_64)

##### STEPS TO REPRODUCE
* git clone repo
* apply fix #9552 (nginx user missing)
* make docker-compose-build
* control + c
* start Docker containers
* make docker-compose
* docker exec tools_awx_1 make clean-ui ui-devel
* GUI: login: admin user
* GUI: resources > templates > Demo Job Template > launch
* GUI: views > jobs > 3 — Demo Project

##### EXPECTED RESULTS

successful job

##### ACTUAL RESULTS

job status failed

```
WARN[0000] Found deprecated file /etc/containers/libpod.conf, please remove. Use /etc/containers/containers.conf to override defaults. 
WARN[0000] Found deprecated file /etc/containers/libpod.conf, please remove. Use /etc/containers/containers.conf to override defaults. 
Error: 'overlay' is not supported over overlayfs, a mount_program is required: backing file system is unsupported for this graph driver
```

##### ADDITIONAL INFORMATION

Valvos420:
Hello, which version of docker do you use ?

lo78cn:
```
docker version
Client: Docker Engine - Community
Version:           20.10.6

docker-compose version
docker-compose version 1.29.1, build c34c88b2

**Comment by garnser** [comment]
Running into the same issue, using the following versions

```
docker-ce-20.10.1-3.el8.x86_64
docker-compose==1.27.4
```

benapetr:
this issue can be worked over by uncommenting mount_program in /etc/containers/containers.conf

noesberger:
> this issue can be worked over by uncommenting mount_program in /etc/containers/containers.conf

I'm running Centos 8.3 (CentOS Linux release 8.3.2011)
With Docker (Docker version 20.10.6, build 370c289)

I have the overlay error, but I don't have a folder /etc/containers. So how I can fix the issue on my centos system?

benapetr:
[...]

PR #10463, opened 2021-06-16
Update Dockerfile.j2

mayasd:
##### SUMMARY

Jobs unable to start because podman trying to use systemd cgroup manager. See error below :

```
WARN[0000] Failed to add conmon to systemd sandbox cgroup: dial unix /run/systemd/private: connect: no such file or directory
Error: OCI runtime error: systemd cgroup flag passed, but systemd support for managing cgroups is not available
```

related #10099 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION

```
awx: 19.0.0
```

##### ADDITIONAL INFORMATION

* According to this PR https://github.com/containers/podman/pull/7009, podman switch references from libpod.conf to containers.conf.
* According to containers.conf man (https://github.com/containers/common/blob/main/docs/containers.conf.5.md), configuration file is a TOML file but engine section declaration is missing.

thanks to @Siorde too :+1:

### DIFF
--- a/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
+++ b/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
@@ -153,7 +153,7 @@ RUN dnf --enablerepo=debuginfo -y install python3-debuginfo || :
 
 {% if build_dev|bool %}
 RUN dnf install -y podman
-RUN echo -e 'cgroup_manager = "cgroupfs"\nevents_logger = "file"' > /etc/containers/libpod.conf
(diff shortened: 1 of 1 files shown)


AlanCoding:
@chrismeyersfsu Would this properly merge with

```toml
[engine]
runtime = "crun"
```

As a file in the `~/.config/containers/containers.conf` location?

chrismeyersfsu:
I don't care too much if `runtime = "crun"` is in the per-user config for the developer container since the container isn't a shared tenant. I think the `crun` is a separate issue. I think this change rightfully belongs in the global system config because, as I understand it, this would effect all users.

chrismeyersfsu (approved):

shanemcd (approved):
Thank you!

PR #10464, opened 2021-06-16
Update and expand docs/ folder for EEs

AlanCoding:
This has some new content about EE precedence, which I don't think we've documented anywhere else, thinking of @tvo318 here, content was developed by @jbradberry 

(I think the numbers 2 and 3 in the global job default EE may still be subject to revision, just a warning)

@shanemcd https://github.com/ansible/awx/pull/10324/ is incorporated into this

This mentions @rebeccahhh's venv migration stuff, but I'm trying to write the absolute minimum possible while still mentioning migration.

### DIFF
--- a/docs/clustering.md
+++ b/docs/clustering.md
@@ -181,13 +181,21 @@ If an Instance Group is configured but all instances in that group are offline o
 
 #### Project Synchronization Behavior
 
-Project updates behave differently than they did before. Previously they were ordinary jobs that ran on a single instance. It's now important that they run successfully on any instance that could potentially run a job. Projects will sync themselves to the correct version on the instance immediately prior to running the job. If the needed revision is already locally checked out and Galaxy or Collections updates are not needed, then a sync may not be performed.
+It is important that project updates run on the instance which prepares the ansible-runner private data directory.
+This is accomplished by a project sync which is done by the dispatcher control / launch process.
+The sync will update the source tree to the correct version on the instance immediately prior to transmitting the job.
+If the needed revision is already locally checked out and Galaxy or Collections updates are not needed, then a sync may not be performed.
 
 When the sync happens, it is recorded in the database as a project update with a `launch_type` of "sync" and a `job_type` of "run". Project syncs will not change the status or version of the project; instead, they will update the source tree _only_ on the instance where they run. The only exception to this behavior is when the project is in the "never updated" state (meaning that no project updates of any type have been run), in which case a sync should fill in the project's initial revision and status, and subsequent syncs should not make such changes.
 
+All project updates run with container isolation (like jobs) and volume mount to the persistent projects folder.
+
 #### Controlling Where a Particular Job Runs
 
-By default, a job will be submitted to the `tower` queue, meaning that it can be picked up by any of the workers.
+By default, a job will be submitted to the default queue (formerly the `tower` queue).
+To see the name of the queue, view the setting `DEFAULT_EXECUTION_QUEUE_NAME`.
+Administrative actions, like project updates, will run in the control plane queue.
+The name of the control plane queue is surfaced in the setting `DEFAULT_CONTROL_PLANE_QUEUE_NAME`.
 
 
 ##### How to Restrict the Instances a Job Will Run On
--- a/docs/execution_environments.md
+++ b/docs/execution_environments.md
@@ -23,7 +23,7 @@ These can become the global job default in certain circumstances.
 Installers should run the `awx-manage register_default_execution_environments` command to pre-populate
 the system with some EEs defined in settings. This will create:
 
- - a control plane EE - corresponding to the `CONTROL_PLANE_EXECUTION_ENVIRONMENT` setting
+ - the control plane EE - corresponding to the `CONTROL_PLANE_EXECUTION_ENVIRONMENT` setting
  - global job EEs - all images enumerated in the `GLOBAL_JOB_EXECUTION_ENVIRONMENTS` setting
 
 These EEs are critical for system function, so this command must be ran for AWX to be functional.
@@ -31,21 +31,21 @@ All EEs created by this command are global EEs.
 
 ### Project Update EE Precedence
 
-Project updates will always use the control plane execution environment.
+Project updates will always use the control plane EE.
 
 ### Job, Ad Hoc Commands, and inventory update EE Precedence
 
 Jobs will use the first available execution environment in this list:
 
-1. an EE from the template the spawned the job
+1. the `execution_environment` defined on the template (job template or inventory source) that spawned the job
 2. the `default_environment` defined on the project that the job uses
-3. the `default_environment` defined on the organization of the job (a direct API field)
+3. the `default_environment` defined on the organization of the job
 4. the `default_environment` defined on the organization of the inventory the job uses
-5. the current `DEFAULT_EXECUTION_ENVIRONMENT` setting
-6. Any images from the `GLOBAL_JOB_EXECUTION_ENVIRONMENTS` setting
-7. Any other global EEs (null organization)
+5. the current `DEFAULT_EXECUTION_ENVIRONMENT` setting (configurable at `/api/v2/settings/jobs/`)
+6. Any image from the `GLOBAL_JOB_EXECUTION_ENVIRONMENTS` setting
+7. Any other global EE
 
-If more than one EE fits a criteria (applies for 5 and 6), then the most recently created one will be used.
+If more than one EE fits a criteria (applies for 6 and 7), then the most recently created one will be used.
 
 ## Migrating from Custom Virtual Environments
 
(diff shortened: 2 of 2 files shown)


docs/execution_environments.md
```diff
@@ -0,0 +1,61 @@
...
+2. Any images from the `GLOBAL_JOB_EXECUTION_ENVIRONMENTS` setting
+3. Any other global EEs (null organization)
+
+If more than 1 EE fits these criteria, then the most recently created one will be used.
+
+## Migrating from Custom Virtual Environments
+
+If you have installed dependencies inside of custom virtual environments in
+a prior release, then have a look at this series of commands for help migrating
+dependencies out of the venvs and into EEs.
```
rebeccahhh, line 54: maybe throw a sentence in about the `-q` option here?

AlanCoding: what's the -q option?

docs/clustering.md
```diff
@@ -181,13 +181,21 @@ If an Instance Group is configured but all instances in that group are offline o
 
 #### Project Synchronization Behavior
 
-Project updates behave differently than they did before. Previously they were ordinary jobs that ran on a single instance. It's now important that they run successfully on any instance that could potentially run a job. Projects will sync themselves to the correct version on the instance immediately prior to running the job. If the needed revision is already locally checked out and Galaxy or Collections updates are not needed, then a sync may not be performed.
+It is important that project updates run on the instance which prepares the ansible-runner private data directory.
```
jbradberry, line 184: Super clear.  I love it.

docs/execution_environments.md
```diff
@@ -0,0 +1,61 @@
...
+For more information, see [ansible-runner EE docs](https://ansible-runner.readthedocs.io/en/latest/execution_environments.html)
+for information on how container isolation works, and see [ansible-builder docs](https://ansible-builder.readthedocs.io/en/latest/index.html)
+for instructions on how to build them.
+
+The Execution Environment model has an `image` field for the image identifier which will be used by jobs.
+The job details view will link to the execution environment that the job uses.
+
+## Creating and using EEs
+
+Users with and organization's `execution_environment_admin_role` can create new EEs in that organization.
```
jbradberry: and -> an

docs/execution_environments.md
```diff
@@ -0,0 +1,61 @@
...
+The RBAC rules follow standard rules for org-scoped resources.
+
+Only superusers can create global EEs (null organization).
```
jbradberry: Do we consistently use 'superuser' or 'system administrator' for this user role?

docs/execution_environments.md
```diff
@@ -0,0 +1,61 @@
...
+### Pre-created EEs
+
+Installers should run the `awx-manage register_default_execution_environments` command to pre-populate
+the system with some EEs defined in settings. This will create:
+
+ - a control plane EE - corresponding to the `CONTROL_PLANE_EXECUTION_ENVIRONMENT` setting
+ - global job EEs - all images enumerated in the `GLOBAL_JOB_EXECUTION_ENVIRONMENTS` setting
+
+These EEs are critical for system function, so this command must be ran for AWX to be functional.
+All EEs created by this command are global (do not belong to an organization).
```
jbradberry: This is the second time this information is given in a parenthetical remark.  Perhaps would should devote a sentence or two to describe the concept explicitly.

docs/execution_environments.md
```diff
@@ -0,0 +1,61 @@
...
+
+Jobs will use the first available execution environment in this list:
+
+1. an EE from the template the spawned the job
+2. the `default_environment` defined on the project that the job uses
+3. the `default_environment` defined on the organization of the job (a direct API field)
+4. the `default_environment` defined on the organization of the inventory the job uses
+5. the global job default EE
+
+This global default job default EE will be the first available out of:
```
jbradberry: I'm not sure I see the value in splitting these out into two separate lists like this.

docs/execution_environments.md
```diff
@@ -0,0 +1,61 @@
...
+
+1. the `DEFAULT_EXECUTION_ENVIRONMENT` setting
+2. Any images from the `GLOBAL_JOB_EXECUTION_ENVIRONMENTS` setting
```
jbradberry: maybe "Any images that were installed from the most recent list contained in the `GLOBAL_JOB_EXECUTION_ENVIRONMENTS` setting"?

docs/execution_environments.md
```diff
@@ -0,0 +1,61 @@
...
+3. Any other global EEs (null organization)
+
+If more than 1 EE fits these criteria, then the most recently created one will be used.
```
jbradberry: maybe 1 -> one?

docs/execution_environments.md
jbradberry: I think running the commands with `-q` allows you to just redirect the output to a file that you can immediately use?

docs/execution_environments.md
AlanCoding: hrmm...

```
(env) [alancoding@alan-red-hat awx]$ grep -r "superuser" docs/
docs/auth/ldap.md:"is_superuser": "cn={{superusergroupname}},cn=groups,cn=accounts,dc=example,dc=com"
docs/credentials/custom_credential_types.md:* Superusers (and only superusers) should be able to define custom `Credential
docs/workflow.md:  * Workflow job templates can only be accessible by superusers ---- system admin, admin of the same organization and system auditor and auditor of the same organization with read permission only.
docs/workflow.md:  * Workflow jobs can be deleted by superusers and org admins of the organization of its associated workflow job template, and no one else.
docs/execution_environments.md:Only superusers can create global EEs (null organization).
(env) [alancoding@alan-red-hat awx]$ 
(env) [alancoding@alan-red-hat awx]$ 
(env) [alancoding@alan-red-hat awx]$ grep -r "system admin" docs/
docs/workflow.md:  * Workflow job templates can only be accessible by superusers ---- system admin, admin of the same organization and system auditor and auditor of the same organization with read permission only.
(env) [alancoding@alan-red-hat awx]$ 
(env) [alancoding@alan-red-hat awx]$ grep -r "System admin" docs/
docs/auth/oauth.md:- System admins will be able to see and manipulate all applications in the system;
(env) [alancoding@alan-red-hat awx]$ grep -r "System Admin" docs/
docs/auth/oauth.md:- The System Administrator is able to see and manipulate every token in the system;
docs/clustering.md:Instance Groups can be created by posting to `/api/v2/instance_groups` as a System Administrator.
docs/rbac.md:want a System Administrator to have access to everything that an Organization
docs/prompting.md:but only with Organization or System Administrator privileges.
```

Prefer to keep "superuser" because it will be easiest to hit for a search & replace, should we make it consistent later.

docs/execution_environments.md
AlanCoding: Agree, I'm rewording a prior section to read more like a terminology introduction

```
EEs without an organization (value is null in the API) are global EEs.
Only superusers can create global EEs.
These can become the global job default in certain circumstances.
```

And then I'll reference "global EEs" without further specification after that.

docs/execution_environments.md
AlanCoding: Yeah, I thought I was going to have a separate section for inventory / ad hoc

but then I didn't, so I'll make then into one list.

docs/execution_environments.md
jbradberry: I personally prefer System Administrator for user-facing docs, but not enough to do more than raise the question.  Carry on.

docs/execution_environments.md
AlanCoding: "most recent list"

I think I know what you're saying, but since this is in the initial release, I'm honestly not sure that behavior will hold once we actually get there.

Even then, I expect communication challenges, because there are 2 constructs here. Say that in version 21.0.0 the defaults are `[A, B]`, then in version 25.0.0 the defaults are `[C, D]` (or was it already self-obvious?). Firstly, we establish that we will pick from either `C` or `D`. Then, we need to somehow establish which we pick between the two. Will it be C, or will it be D?

I really want to have more of a process to answer this completely when the time comes.

docs/execution_environments.md
AlanCoding: gotcha, attempted.

docs/execution_environments.md
AlanCoding: Alright, I added a "current" to this wording, so that hints at what's to come, but I don't want to say too much.

jbradberry (approved):

AlanCoding:
whoops, had a few errors after implementing the prior updates. Hopefully it should be good now.

docs/execution_environments.md
rebeccahhh: That is not correct (but it's a good idea), it allows these to to be run without any output except the results from the functions (no tips or explanations, which are included in the default commands). This is specifically useful if you want to run them in combination or in a scripting style workup

AlanCoding: Thanks, updated the wording to be correct.

AlanCoding:
regate

PR #10477, opened 2021-06-18
Symlink collection modules for Ansible 2.9 compatibility

AlanCoding:
I need test results here.

This fixes use like `awx.awx.tower_host`, for the module name in the task.


shanemcd (approved):

AlanCoding:
Actually, I think I will close this for the sake of simplicity of next downstream merge. Sorry about that. Too many things going on.

PR #10484, opened 2021-06-21
only get active long running queries

chrismeyersfsu:
Getting all connections that have been alive > 5 minutes is noisy. Instead, we want only queries that are active and have been alive for more than 5 minutes. This will filter out any idle connections.


shanemcd (approved):

shanemcd:
recheck

shanemcd:
recheck

chrismeyersfsu:
recheck

Issue #7524, opened 2020-07-01
JT Survey fancy characters in answer variable name breaks the survey


chrismeyersfsu:
##### SUMMARY
Special characters in the `Answer variable name` field do not allow for the field to be edited later. Feels like it has something to do with the routing URL.

##### STEPS TO REPRODUCE

* Create a JT with a survey
* Enter `/hello_world` into `Answer variable name`
* Save
* Edit the survey variable you just created

##### EXPECTED RESULTS

To be shown the edit view

##### ACTUAL RESULTS

Edit view fails to render (see below)

![image](https://user-images.githubusercontent.com/722880/86285283-530aa200-bbb2-11ea-960a-9f0fae4685fa.png)

![image](https://user-images.githubusercontent.com/722880/86285469-ada3fe00-bbb2-11ea-9b9b-9d7dab900409.png)

```
main.ccc609b2.chunk.js:1 TypeError: Cannot read property 'variable' of undefined
    at main.ccc609b2.chunk.js:1
    at l (runtime.js:45)
    at Generator._invoke (runtime.js:274)
    at Generator.forEach.e.<computed> [as next] (runtime.js:97)
    at r (asyncToGenerator.js:3)
    at s (asyncToGenerator.js:25)
    at asyncToGenerator.js:32
    at new Promise (<anonymous>)
    at asyncToGenerator.js:21
    at main.ccc609b2.chunk.js:1
```

`dfdsafds<b>bold</b>\"><br>` as `Answer variable name` also fails. But in a different way. Editing the variable shows a blank edit page. The url is https://localhost:8043/next/templates/job_template/8/survey/edit/dfdsafds%3Cb%3Ebold%3C/b%3E/%22%3E%3Cbr%3E
![image](https://user-images.githubusercontent.com/722880/86285586-de843300-bbb2-11ea-99a2-507457ed371b.png)

nixocio:
This issue seems to be fixed as side effect of other work on the surveys.

Create Survey 

![image](https://user-images.githubusercontent.com/9053044/122264203-422b6d00-cea5-11eb-97b9-80cd6cac13bd.png)

Edit Survey

![image](https://user-images.githubusercontent.com/9053044/122264449-8d458000-cea5-11eb-9f52-e66358f20237.png)

Launch Job Template

![image](https://user-images.githubusercontent.com/9053044/122264561-abab7b80-cea5-11eb-94af-8e8d034196d7.png)

tiagodread:
[...]

PR #10504, opened 2021-06-23
Encode/decode question variables

nixocio:
Encode/decode question variables

See: https://github.com/ansible/awx/issues/7524


AlexSCorey (approved):
I tested this locally and everything checks out as expected.

tiagodread (approved):

PR #10509, opened 2021-06-24
Remove warning about error is null

nixocio:
Just render `ErrorDetail` if there is an error. Avoiding warning about
`error` being a required field.


jakemcdermott (approved):

keithjgrant (changes requested):
I think a better approach is to check for the error inside the `AlertModal` itself, since we clearly need to do that check pretty much every time. Then we don't have to remember to do a null check everywhere. I actually already made that change in #10614

nixocio:
Closing this one in favor of the changes here: https://github.com/ansible/awx/pull/10614

PR #10512, opened 2021-06-24
Add job status on expand row for hosts

nixocio:
Add job status on expand row for hosts

image

See: https://github.com/ansible/awx/issues/5236

### DIFF
--- a/awx/ui_next/src/screens/Host/HostList/HostListItem.jsx
+++ b/awx/ui_next/src/screens/Host/HostList/HostListItem.jsx
@@ -1,62 +1,97 @@
 import 'styled-components/macro';
-import React from 'react';
+import React, { useState } from 'react';
 import { string, bool, func } from 'prop-types';
 
 import { t } from '@lingui/macro';
 import { Button } from '@patternfly/react-core';
-import { Tr, Td } from '@patternfly/react-table';
+import { Tr, Td, ExpandableRowContent } from '@patternfly/react-table';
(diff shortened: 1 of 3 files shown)


awx/ui_next/src/screens/Host/HostList/HostListItem.jsx
```diff
@@ -1,62 +1,99 @@
...
 
 function HostListItem({ host, isSelected, onSelect, detailUrl, rowIndex }) {
   const labelId = `check-action-${host.id}`;
+  const [isExpanded, setIsExpanded] = useState(false);
+
+  const {
+    summary_fields: { recent_jobs: recentJobs = [] },
+  } = host;
+
+  const recentPlaybookJobs = recentJobs.map(job => ({ ...job, type: 'job' }));
```
jakemcdermott, line 24: Are we replacing the `type` value or adding it in to each object?

jakemcdermott: Are we replacing the type value or adding it in to each object?

nixocio: @jakemcdermott I was adding the type for every object. However this data is already present on the API endpoint.

image

jakemcdermott (approved):

nixocio:
e2e test triggered.

tiagodread (approved):

Issue #10454, opened 2021-06-15
Add testing around the execution environment collections module


fosterseth:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Need some basic testing around creating EE using modules

PR #10524, opened 2021-06-25
Add Integration Test Playbook for execution_environment Collections Module

beeankha:
##### SUMMARY

Adressing Issue https://github.com/ansible/awx/issues/10454


AlanCoding:
won't we need this downstream as well?

AlanCoding (approved):

PR #10543, opened 2021-06-29
Update last name for Resource Access List

### LINKED ISSUE #10533: Org Access Tab- First Name Displayed in Last Name Column
Bug Report

Issue: When a user is on the Access tab of an organization, and users are shown there, the first name of the user is also printed in the last name column (instead of the last name being there).

Expected results: the Org Access page name columns would show the first name and last name properly.

Actual Results: The user's first name is displayed in the first name column and the last name column.

image

nixocio:
Update last name for Resource Access List

image

Closes: https://github.com/ansible/awx/issues/10533


marshmalien (approved):
Looks good!

tiagodread (approved):

Issue #10568, opened 2021-07-01
Unable to customize args or image directly from the container group configuration


AlexisDuf:
On 19.0.0 we were able to customize the image and the args.

And now(on 19.2.2) it's kind of hard coded 
```python
        pod_spec['spec']['containers'][0]['image'] = ee.image
        pod_spec['spec']['containers'][0]['args'] = ['ansible-runner', 'worker', '--private-data-dir=/runner']
```
https://github.com/ansible/awx/blob/2afa406b7f16971ea83150c326f8e04717b5c14d/awx/main/tasks.py#L3075

Could we add a `if` condition if image is already set ? 

About `args` the default command in the image is `ansible-runner worker --private-data-dir=/runner` (cf https://github.com/ansible/awx-ee/blob/devel/run.sh) so we can't overload this part

shanemcd:
Curious - what do you need to change the command to? Anything else will likely cause AWX not to work properly. I did this to stop people from shooting themselves in the foot.

AlexisDuf:
Hi @shanemcd, 

Yes, I customized awx-ee by executing some script before running the command `ansible-runner worker --private-data-dir=/runner`.  And yes I noticed that AWX can't start a job properly 

As workaround I tried to use an `initContainers` in the pod spec but it's not working too. 

And I don't have any output to see what's going wrong

shanemcd:
I think `initContainers` is probably the more elegant solution here. We'll look into why that isn't working.

nmaupu:
Same here, I would like to be able to customize the image args to configure stuff before job is ran.
I could do all the stuff using ansible itself but it's not very a nice solution because this ansible task would need to be included in all playbooks...
I need some line of bash to generate a vault access token prior to run all the jobs. `initContainers` is not the best way imho because it makes the custom pod spec complicated for nothing (initContainers specs + volumes mounting for file sharing between containers vs. a path to a custom entrypoint).

As a workaround, I tried to make a script named `ansible-runner` with `chmod +x` containing:

```
#!/bin/bash

## My stuff here

/usr/local/bin/ansible-runner $@
```

This can be executed instead of the default args if you put it in the `PATH` before `ansible-runner` binary.

But the job does not start and no logs are output in `awx-task` container. As of now, I can't figure out why...

It would be much better to let user choose what they want to do... I mean we are all adults here ;)

nmaupu:
[...]

PR #10569, opened 2021-07-01
Allow custom image and args

AlexisDuf:
##### SUMMARY
Being able to customize Image and Args from the container group pod specification 

related #10568 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME

 - API (awx/main/tasks.py)

##### AWX VERSION

```
awx: 19.2.2
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -3088,8 +3088,8 @@ class AWXReceptorJob:
             pod_spec_override = parse_yaml_or_json(self.task.instance.instance_group.pod_spec_override)
         pod_spec = {**default_pod_spec, **pod_spec_override}
 
-        pod_spec['spec']['containers'][0]['image'] = ee.image
-        pod_spec['spec']['containers'][0]['args'] = ['ansible-runner', 'worker', '--private-data-dir=/runner']
+        if pod_spec['spec']['containers'][0]['image'] is None or pod_spec['spec']['containers'][0]['image'] == "":
+            pod_spec['spec']['containers'][0]['image'] = ee.image
 
         # Enforce EE Pull Policy
         pull_options = {"always": "Always", "missing": "IfNotPresent", "never": "Never"}


awx/main/tasks.py
```diff
@@ -3088,8 +3088,8 @@ def pod_definition(self):
             pod_spec_override = parse_yaml_or_json(self.task.instance.instance_group.pod_spec_override)
         pod_spec = {**default_pod_spec, **pod_spec_override}
 
-        pod_spec['spec']['containers'][0]['image'] = ee.image
-        pod_spec['spec']['containers'][0]['args'] = ['ansible-runner', 'worker', '--private-data-dir=/runner']
+        if pod_spec['spec']['containers'][0]['image'] is None or pod_spec['spec']['containers'][0]['image'] == "":
```
shanemcd, line 3091: This feels wrong to me. 

I think the right thing to do would change the `image:` field in the default pod spec to say something like:

 ```
image: <placeholder> # Value comes from the resolved Execution Environment at runtime
```

File: awx/main/tasks.py [hunk 2]
```diff
@@ -3088,8 +3088,8 @@ def pod_definition(self):
             pod_spec_override = parse_yaml_or_json(self.task.instance.instance_group.pod_spec_override)
         pod_spec = {**default_pod_spec, **pod_spec_override}
 
-        pod_spec['spec']['containers'][0]['image'] = ee.image
-        pod_spec['spec']['containers'][0]['args'] = ['ansible-runner', 'worker', '--private-data-dir=/runner']
```
shanemcd, line 3092: -1. See https://github.com/ansible/awx/issues/10568#issuecomment-872314993

shanemcd (changes requested):
Comments inline.

awx/main/tasks.py
nmaupu: If I override the pod spec's image, I would think my configuration would be taken into account. Changing the EE to actually change the image is confusing.

awx/main/tasks.py
nmaupu: It's fine to put a default value if not set but I feel it's wrong when it's silently replaced.
User should be able to override to whatever value they'd like

shanemcd:
The default pod spec now contains the command that gets run.

Issue #790, opened 2017-12-11
Add UI support for management jobs in workflows


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
 - UI

##### SUMMARY
The UI allows 3 types of templates in workflows, this issue is to add the option to add system job templates.

##### ADDITIONAL INFORMATION

Creating a break-off issue from #779.

The API for doing this works fine. You can add a System Job Template to a workflow, and also specify `extra_vars`, which are needed for many of them.

The difficult determination that needs to be made is how the UI would present the options for the extra_vars that go into the system job without making it way too complicated.

mabashian:
Exposing system jobs in that list would be trivial.  We'd just need to tweak the GET request to include system jobs in addition to job templates and workflows.  There'd be some other UX considerations for sure like adding a S icon like the P (project), I (inventory) and W (workflow) icons that we currently have.

As Alan pointed out, the tricky part would be exposing the form fields:

<img width="1366" alt="screen shot 2018-11-26 at 3 16 09 pm" src="https://user-images.githubusercontent.com/9889020/49039606-416b4800-f18e-11e8-8c3a-e3685ed93019.png">

in the workflow node form.  We could potentially leverage the PROMPT modal for this and make them required fields.  Alternatively we could just expose those fields under the list when a system job is selected.

wenottingham:
At some point I think system jobs should probably be a real project with real survey specs, etc. But that's probably an unpleasant migration.

tiagodread:
This was added by #10572

PR #10572, opened 2021-07-01
Add UI support for management jobs in workflows

mabashian:
##### SUMMARY
link https://github.com/ansible/awx/issues/790

This has been supported via the API for a long time.  The tricky part was handling the `days` (days to keep) for some of the management jobs.  That value needs to flow in/out of extra_data so it's a bit different from other workflow node fields.

Here it is in action:

mgt_job_wf

I elected to just show the variables in the details view for simplicity.  If it's important we could pull `days` out of the object but more code brings added risk of bugs 🤷 .

Screen Shot 2021-07-01 at 11 23 33 AM

@wenottingham if you're not really into this as a feature that's fine - we can close this and reference it later.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/util/validators.js
+++ b/awx/ui_next/src/util/validators.js
@@ -57,6 +57,12 @@ export function minLength(min) {
 
 export function minMaxValue(min, max) {
   return (value) => {
+    if (!Number.isFinite(min) && value > max) {
+      return t`This field must be a number and have a value less than ${max}`;
+    }
+    if (!Number.isFinite(max) && value < min) {
+      return t`This field must be a number and have a value greater than ${min}`;
+    }
     if (value < min || value > max) {
       return t`This field must be a number and have a value between ${min} and ${max}`;
     }
--- a/awx/ui_next/src/api/models/SystemJobTemplates.js
+++ b/awx/ui_next/src/api/models/SystemJobTemplates.js
@@ -10,12 +10,6 @@ class SystemJobTemplates extends Mixins {
     this.baseUrl = '/api/v2/system_job_templates/';
   }
 
-  readDetail(id) {
-    const path = `${this.baseUrl}${id}/`;
-
-    return this.http.get(path).then(({ data }) => data);
-  }
-
   launch(id, data) {
     return this.http.post(`${this.baseUrl}${id}/launch/`, data);
   }
--- a/awx/ui_next/src/components/PromptDetail/PromptDetail.js
+++ b/awx/ui_next/src/components/PromptDetail/PromptDetail.js
@@ -1,18 +1,15 @@
 import 'styled-components/macro';
 import React from 'react';
 import { shape } from 'prop-types';
-
 import { t, Trans } from '@lingui/macro';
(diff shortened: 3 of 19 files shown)


AlexSCorey (changes requested):
Some things I found:
1) if you type/edit the days field it then shows up as a string in the variable details view after clicking on the `i` icon when hovering a node.  However, when you edit that same node, but don't edit the days field the number appears as an integer  in the variables detail after clicking on the `i` icon.  There doesn't seem to be any negative impact on functionality, just some ux inconsistency.

2) We don't seem to show the convergence data in the detail modal.  Should we?

3) `Days` vs `Days to keep`
Screen Shot 2021-07-06 at 10 19 09 AM
Screen Shot 2021-07-06 at 10 19 20 AM

4) Create a `Clean up job details` node and connect it to a `Cleanup expired sessions` node.  Then click on the `Cleanup expired sessions` node `i` icon.  Page crashes.
Screen Shot 2021-07-06 at 10 22 57 AM
Screen Shot 2021-07-06 at 10 23 08 AM

mabashian:
> if you type/edit the days field it then shows up as a string in the variable details view after clicking on the i icon when hovering a node. However, when you edit that same node, but don't edit the days field the number appears as an integer in the variables detail after clicking on the i icon. There doesn't seem to be any negative impact on functionality, just some ux inconsistency.

Will include a fix for this

> We don't seem to show the convergence data in the detail modal. Should we?

Was probably just missed when we did the initial work.  Let's file an issue for this.

> Days vs Days to keep

The api expects the key to be `days` in `extra_data` so that can't change.  Labelling the field `Days` is ambiguous to me so I don't think we should change it to that.  Here's what the field looks like when you launch a management job:

Screen Shot 2021-07-12 at 12 01 24 PM

I'd propose changing the string to something like:

Screen Shot 2021-07-12 at 12 03 23 PM

so that it more closely matches the other interface.

> Create a Clean up job details node and connect it to a Cleanup expired sessions node. Then click on the Cleanup expired sessions node i icon. Page crashes.

I'll look at this

AlexSCorey (approved):
This looks good to me.  The unit tests pass, and the issues described earlier are resolved.  I'm triggering e2e tests now.

tiagodread:
@mabashian could you rebase this one?

awx/ui_next/src/util/validators.js
```diff
@@ -57,6 +57,12 @@ export function minLength(min) {
 
 export function minMaxValue(min, max) {
   return (value) => {
+    if (!Number.isFinite(min) && value > max) {
+      return t`This field must be a number and have a value less than ${max}`;
+    }
+    if (!Number.isFinite(max) && value < min) {
+      return t`This field must be a number and have a value greater than ${min}`;
```
tiagodread, line 64: This is how the ${min} is being displayed.
 
image

tiagodread (approved):
Verified that mgmt jobs can be added as nodes to a workflows and run it, also tested the scheduled workflow with mgmt jobs.

Issue #10571, opened 2021-07-01
awx_collection: Adding inventory source with duplicated project name fails in 'multi-tenant' like environment 


Tompage1994:
##### ISSUE TYPE
 - Bug Report

##### COMPONENT NAME
 - awx_collection

##### SUMMARY
Adding an inventory source of type `scm` fails when the source_project has the same name as another project in another organization. This therefore fails in a multi-tenant environment.

##### ENVIRONMENT
* AWX collection version: 19.2.2 

##### STEPS TO REPRODUCE

- Create 2 orgs
- Create a project and copy it so there is one in each org with the same name. 
- Create an inventory in one org
- Use the awx.awx.inventory_source as below:
```
- awx.awx.inventory_source:
        name: is1
        source: scm
        source_project: Demo Project
        organization: Default
        inventory: Demo Inventory
```

##### EXPECTED RESULTS

It should use the organization which is used for the inventory by default, but you should also be able to specify the source_project's org.

##### ACTUAL RESULTS

```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Request to /api/v2/projects/?name=Demo+Project returned 2 items, expected 1", "query": {"name": "Demo Project"} ...
```

shanemcd:
Probably just need to add the organization to the query params here.

john-westcott-iv:
With the merged PR solving the crash and the ability to look up items with the `lookup` module we are going to close this issue. For reference, as mentioned in the PR, once can lookup objects using the lookup module like:
```
  - name: create INVSOURCE
    inventory_source:
      name: inv1
      source: scm
      source_project: "{{ lookup('awx.awx.controller_api', 'projects', query_params={'name': 'Demo Project', 'organization__name':'org1'}, return_ids=True )}}"
      organization: Default
      inventory: Demo Inventory
  ```

PR #10573, opened 2021-07-01
Default source_project to organization for inventory source

Tompage1994:
##### SUMMARY

related #10571 

Adds the defaulting of the source_project (when given) to be looked up under the `organization` selected in the same way the inventory is looked up.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
- awx_collection

##### AWX VERSION
awx_collection v19.2.2

### DIFF
--- a/awx_collection/plugins/modules/inventory_source.py
+++ b/awx_collection/plugins/modules/inventory_source.py
@@ -256,7 +256,10 @@ def main():
     if ee is not None:
         inventory_source_fields['execution_environment'] = module.resolve_name_to_id('execution_environments', ee)
     if source_project is not None:
-        inventory_source_fields['source_project'] = module.resolve_name_to_id('projects', source_project)
+        source_project_object = module.get_one('projects', name_or_id=source_project, data=lookup_data)
+        if not source_project_object:
+            module.fail_json(msg='The specified source project, {0}, was not found.'.format(lookup_data))
+        inventory_source_fields['source_project'] = source_project_object['id']
 
     OPTIONAL_VARS = (
         'description',
--- a/awx_collection/tests/integration/targets/inventory_source_update/tasks/main.yml
+++ b/awx_collection/tests/integration/targets/inventory_source_update/tasks/main.yml
@@ -28,6 +28,14 @@
         scm_url: https://github.com/ansible/test-playbooks
         wait: true
 
+    - name: Create a git project with same name, different org
+      project:
+        name: "{{ project_name }}"
+        organization: Default
+        scm_type: git
+        scm_url: https://github.com/ansible/test-playbooks
+        wait: true
+
     - name: Create an Inventory
       inventory:
         name: "{{ inv_name }}"


fosterseth:
this can be achieved more generally using our lookup plugin (albeit more verbose)
e.g.

```
tasks:
  - name: get project in org1
    set_fact:
        proj1: "{{ lookup('awx.awx.controller_api', 'projects', query_params={'name': 'Demo Project', 'organization__name':'org1'}, return_ids=True )}}"

  - name: create INVSOURCE
    inventory_source:
      name: inv1
      source: scm
      source_project: "{{ proj1 }}"
      organization: Default
      inventory: Demo Inventory
```

Tompage1994:
> this can be achieved more generally using our lookup plugin (albeit more verbose)
> e.g.
> 
> ```
> tasks:
>   - name: get project in org1
>     set_fact:
>         proj1: "{{ lookup('awx.awx.controller_api', 'projects', query_params={'name': 'Demo Project', 'organization__name':'org1'}, return_ids=True )}}"
> 
>   - name: create INVSOURCE
>     inventory_source:
>       name: inv1
>       source: scm
>       source_project: "{{ proj1 }}"
>       organization: Default
>       inventory: Demo Inventory
> ```

I agree you can, but I think its much nicer to have it as part of the module as its something that fundamentally breaks at the moment if there are multiple projects under the same name without having to do the second call as you describe (This is also something you have to know to do and isn't necessarily intuitive to work out). 

In much the same way as there is an `organization` field on this module to ensure the lookup of the inventory's org is correct, it surely makes sense to do the same thing for the project too?

Tompage1994:
Have updated and removed the extra option and made it so we just default the project lookup to the `organization`

beeankha:
Hi @Tompage1994 , would you mind adding a task or two to the integration test playbook for the  Inventory Source module? The test playbook can be found here: https://github.com/ansible/awx/blob/devel/awx_collection/tests/integration/targets/inventory_source/tasks/main.yml

john-westcott-iv:
When initially designed the collection was assuming that things would be defined together and that the returned id from one object could be used in the second object. For example:
```
tasks:
  - name: Create Organization
    tower_organization:
      name: Default
    register: my_default_org

- name: Create inventory
  tower_inventory:
    name: Demo Inventory
    organization: "{{ my_default_org.id }}"
```
In practice,  there are times where things are not all defined at the same time or in the same playbook. So we added the ability to resolve by name but that introduced issues with non-unique instance of an objects. This is why we built the lookup plugin, so that you can resolve an objects name to its ID if the name is non-unique.

The problem with adding a new filed like this is that any object that has a dependency on another object could also require similar fields. For example take job_template, if we added fields like this we would need the following fields:
    - credential_organization
    - inventory_organization
    - project_organization
    - credentials_organization (this would somehow have to be in the credentials hash)
  
If I recall correctly there are one or two objects in Tower where a name and an org may not provide a unique ID as they can be a child of another non-unique object. So while it may be trivial in some modules, adding fields to identify uniqueness of Tower objects can be a very large task. 

Also, as an FYI, it can be combined into a single step rather than two:
```
tasks:
  - name: create INVSOURCE
    inventory_source:
      name: inv1
      source: scm
      source_project: "{{ lookup('awx.awx.controller_api', 'projects', query_params={'name': 'Demo Project', 'organization__name':'org1'}, return_ids=True )}}"
      organization: Default
      inventory: Demo Inventory
```

Tompage1994:
@john-westcott-iv I completely take on board what you say here and I do agree entirely.

I've now changed this PR to address the fact that the module does fail if there are two projects of the same name. Now the project org will default to `organization` in the same way the inventory does. Hopefully this is something more useful. Speaking to @sean-m-sullivan, this functionality exists elsewhere.

I've also added something to the integration tests..

beeankha (approved):

PR #10588, opened 2021-07-06
Workflow slice aggregation

sean-m-sullivan:
##### SUMMARY
This module is built to Aggregate artifacts from a sliced job template. This was meant to be used in the next node of a workflow after a sliced job template to replicate set_stats in a workflow. I believe this is a long requested feature in tower, and this is a best effort module to add that feature. 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - awx_collection

##### AWX VERSION

```
19.2.2
```

### DIFF
--- a/awx_collection/plugins/module_utils/controller_api.py
+++ b/awx_collection/plugins/module_utils/controller_api.py
@@ -707,7 +707,7 @@ class ControllerAPIModule(ControllerModule):
         else:
             return True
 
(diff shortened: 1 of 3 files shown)


wenottingham:
This is the wrong place to solve this.

You're making a controller-only feature depend on a specific module in the awx.awx collection depending on how it's run.

This should be solved in the controller, not by making users rewrite playbooks.

Issue #2032, opened 2018-06-29
Aliases for templates/jobs within workflow


sumkincpp:
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
 - API
 - UI

##### SUMMARY

Would be great to have job aliases for templates/jobs within workflow which will be shown on workflow pipeline.

Currently only template name is shown. If it is reusable template with many configurable options it's name might not fully describe exact template action performed within workflow.

wenottingham:
Assume this would apply to schedule configurations as well.

AlanCoding:
This is a duplicate of https://github.com/ansible/awx/issues/1422, although I had no traction for that one and I'll be happy to close that in favor of this one.

> Assume this would apply to schedule configurations as well.

I don't follow that. Schedules have a `name` field already. Yes, schedules and workflow nodes have a commonality that they may contain promotable fields, but this lack of an identifier/name problem is unique to the case of workflow nodes.

unlikelyzero:
Note there is a similar request here https://github.com/ansible/awx/issues/5824

AlanCoding:
Status update: merging https://github.com/ansible/awx/pull/6291 for adding the `identifier` field, which does something like this, but targeted toward API clients / the CLI / the Ansible modules.

As a UI request, this still stands.

I would encourage the UI to leverage the `identifier` field, but the auto-generation of UUIDs when it's blank could give it some challenges (which gets parity between the UI and playbook-created content). I understand if the UI instead goes its own way and adds a different description-like field.

Anyway, I'm taking the API label off because it's not sure to require API changes, and it is oriented towards UI-next.

gpiucco:
+1. without this I have to pragmatically created thousands of JT just so my users can understand what the workflow is doing

Gianlu:
Hello,
do you have any news about this issue? It would be very useful.

Thanks

AlexSCorey:
@Gianlu and @gpiucco no news yet.  Discussions about adding this feature to the new ui have started but nothing definitive has been set.    Follow this issue for updates, if you aren't already doing so.

tiagodread:
The aliasing in workflow nodes was added by #10592, closing this ticket.

PR #10592, opened 2021-07-06
Adds support for workflow node aliasing via identifier field

mabashian:
##### SUMMARY
link #2032 

cc @AlanCoding since you may have opinions here based on your comments in the original issue.

The implementation leverages the `identifier` field on a workflow node in order to allow users to specify an alias to be displayed on the node instead of the resource name.  This would allow users to uniquely identify nodes that have the same underlying unified job template.  If the string is a uuid (generated by the server) then we still display the resource name.

Here it is in action:

node_alias

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui/src/components/Workflow/WorkflowNodeHelp.js
+++ b/awx/ui/src/components/Workflow/WorkflowNodeHelp.js
@@ -38,11 +38,17 @@ function WorkflowNodeHelp({ node }) {
   const unifiedJobTemplate =
     node?.fullUnifiedJobTemplate ||
     node?.originalNodeObject?.summary_fields?.unified_job_template;
-  const identifier =
-    node?.identifier ||
-    (!stringIsUUID(node?.originalNodeObject?.identifier)
-      ? node.originalNodeObject.identifier
-      : null);
+  let identifier = null;
+  if (node?.identifier) {
+    ({ identifier } = node);
+  } else if (
+    node?.originalNodeObject?.identifier &&
+    !stringIsUUID(node.originalNodeObject.identifier)
+  ) {
+    ({
+      originalNodeObject: { identifier },
+    } = node);
+  }
   if (unifiedJobTemplate || job) {
     const type = unifiedJobTemplate
       ? unifiedJobTemplate.unified_job_type || unifiedJobTemplate.type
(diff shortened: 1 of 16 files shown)


mabashian:
cc @wenottingham - if this is a bad idea let me know and we'll close this one

wenottingham:
@mabashian no, this is neat. Is it plumbed through the job detail screen as well if you cllick into it on a workflow run?

AlanCoding:
> If the string is a uuid (generated by the server) then we still display the resource name.

That's pretty critical behavior. I think I agree with your decisions here.

Disclosure: I haven't actually looked at the code yet.

How are you distinguishing UUID type identifiers from user-given ones? Some pattern matching?

mabashian:
> How are you distinguishing UUID type identifiers from user-given ones? Some pattern matching?

yea right now it's just a regex match

mabashian:
> Is it plumbed through the job detail screen as well if you cllick into it on a workflow run?

I'll have to double check on this one

mabashian:
@wenottingham so if you click on a running node to view the details I don't know that we have the info we need via the api to display the `identifier`.  We'd need something that pointed back to the node in the summary_fields for the job.  Maybe @AlanCoding knows if that's possible or not.  If it was there we could certainly display it in the details.

AlanCoding:
We already show the workflow job in the list view. I'd bet this query is poor performance, and I see no UI need for it in the _list view_. I digress.

We fully load the job's _node_ before we can load the _workflow_. So as long as we are properly using the cache (we likely aren't), there is no performance impact for giving node details. `identifier` is a direct node field, so no worries there.

Answering the question, no technical blockers, but we should subject the view's performance to scrutiny (using the normal Django toolbar stuff) as we implement this.

keithjgrant (approved):

tiagodread:
The `Node Alias` is not available on job details but this is really cool, on hover over the node the user can see the `Node Alias` in addition to `Resource Name`

Screenshot from 2021-08-03 13-16-11

tiagodread (approved):
@mabashian once the rebase is done we can merge this one

PR #10614, opened 2021-07-08
Test cleanup

keithjgrant:
##### SUMMARY
* Makes a unit test fail if any errors or warnings are logged (see setupTest.js)
* Updates/fixes all tests that were logging errors
* Fixes a key cause of test flake caused by Jobs.test.jsx (I have still seen a rogue "Network error" cause a flaky failure once; which I will hopefully take care of in a subsequent PR)
* Removes all unnecessary calls to `wrapper.unmount()` (this also surfaced several instances of tests leaving asynchronous code running after test completion, which are now fixed)

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui_next/src/components/ErrorDetail/ErrorDetail.jsx
+++ b/awx/ui_next/src/components/ErrorDetail/ErrorDetail.jsx
@@ -36,6 +36,10 @@ function ErrorDetail({ error }) {
   const { response } = error;
   const [isExpanded, setIsExpanded] = useState(false);
 
+  if (!error) {
+    return null;
+  }
+
   const handleToggle = () => {
     setIsExpanded(!isExpanded);
   };
@@ -84,7 +88,10 @@ function ErrorDetail({ error }) {
 }
 
 ErrorDetail.propTypes = {
-  error: PropTypes.instanceOf(Error).isRequired,
+  error: PropTypes.instanceOf(Error),
+};
+ErrorDetail.defaultProps = {
+  error: null,
 };
 
 export default ErrorDetail;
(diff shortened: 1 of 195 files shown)


awx/ui_next/src/components/ErrorDetail/ErrorDetail.jsx
```diff
@@ -84,7 +88,10 @@ function ErrorDetail({ error }) {
 }
 
 ErrorDetail.propTypes = {
-  error: PropTypes.instanceOf(Error).isRequired,
+  error: PropTypes.instanceOf(Error),
+};
+ErrorDetail.defaultProps = {
+  error: null,
```
nixocio, line 94: With the change to `fail tests if warning is logged` could we keep this a required prop? 

See: https://github.com/ansible/awx/pull/10509

nixocio (approved):
This is an excellent improvement.

@keithjgrant from your findings updating the tests is there something that we should document here https://github.com/ansible/awx/blob/devel/awx/ui_next/CONTRIBUTING.md to prevent future issues?

awx/ui_next/src/components/ErrorDetail/ErrorDetail.jsx
keithjgrant: Oh nice, I hadn't seen that PR. Though I figured since we obviously need that conditional check everywhere, we might as well do it once inside the component instead of all over the place

jakemcdermott (approved):
Looks great!

keithjgrant:
> from your findings updating the tests is there something that we should document here https://github.com/ansible/awx/blob/devel/awx/ui_next/CONTRIBUTING.md to prevent future issues?

This is a good point. I think we should probably discuss at some point if we have any clear guidelines when to use `shallow` vs `mount`, so we can offer guidance in CONTRIBUTING.md

tiagodread (approved):

tiagodread:
regate

keithjgrant:
hmm, I'll check if my changes don't like something since my last rebase

PR #10617, opened 2021-07-09
Run flake8 on the AWX collection in CI

AlanCoding:
We kind of assumed that linter issues like this would be picked up by `ansible-test sanity`, but they're not.

This seems to provide meaningful value. Unused variables are reflecting a lot of history of changes, and it's good to move on to leave those in the past.

### DIFF
--- a/awx_collection/test/awx/test_instance_group.py
+++ b/awx_collection/test/awx/test_instance_group.py
@@ -5,7 +5,6 @@ __metaclass__ = type
 import pytest
 
 from awx.main.models import InstanceGroup, Instance
-from awx.main.tests.functional.conftest import kube_credential, credentialtype_kube
 
 
 @pytest.mark.django_db
--- a/tox.ini
+++ b/tox.ini
@@ -9,8 +9,9 @@ setenv =
   BLACK_ARGS = --check
 commands =
   make black
-  flake8 awx awxkit
+  flake8 awx awxkit awx_collection
   yamllint -s .
 
 [flake8]
 select = F401,F402,F821,F823,F841
+exclude = awx/ui_next/node_modules,awx/ui/node_modules,env,awx_collection_build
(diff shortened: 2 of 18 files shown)


tox.ini
```diff
@@ -9,8 +9,9 @@ setenv =
   BLACK_ARGS = --check
 commands =
   make black
-  flake8 awx awxkit
+  flake8 awx awxkit awx_collection
   yamllint -s .
 
 [flake8]
 select = F401,F402,F821,F823,F841
+exclude = awx/ui_next/node_modules,awx/ui/node_modules,env,awx_collection_build
```
AlanCoding, line 17: This is so that I can run

```
flake8 .
```

and it will finish in a reasonable amount of time. If it hits some kind of site packages, it's not useful.

awx_collection/test/awx/test_instance_group.py
```diff
@@ -5,7 +5,6 @@
 import pytest
 
 from awx.main.models import InstanceGroup, Instance
-from awx.main.tests.functional.conftest import kube_credential, credentialtype_kube
```
beeankha, line 8: We still need this line, but can take out `credentialtype_kube`

AlanCoding: We don't need `kube_credential` either do we? It's passed as a fixture. Or do we need to import so that pytest loads the `conftest.py`? That doesn't make total sense to me. I think mucking with the pytest config might be the right way to fix it.

beeankha: Yes, I think we need it for the loading of `conftest.py`; your idea of changing the pytest config sounds like a good way to possibly fix this.

beeankha (approved):

awx_collection/test/awx/test_instance_group.py
AlanCoding: I tried a little bit, but it's going to be a good deal more work to change the general pattern, and we have similar fixtures so I added this along with them.

AlanCoding:
Integration tests look good, so I'm going to go with this.

PR #10623, opened 2021-07-09
Adds functionality for expand all 

### LINKED ISSUE #4208: Expand/collapse all rows v. single row
There are some lists (Jobs and Job Templates in particular) where it may not make a lot of sense for the user to expand/collapse all of the rows at the same time.  It may make more sense to allow the user to expand/collapse a single row.  I'm thinking that this could be a feature _in addition to_ the list level expand/collapse.  Ping @trahman73 for thoughts.  Ping @jakemcdermott because you mentioned something along these lines in our meeting.

link #47 because it's loosely related

AlexSCorey:
##### SUMMARY
Closes #4208.  Adds basically the same work as useSelected, in a useExpanded hook.
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
expand

### DIFF
--- a/awx/ui_next/src/components/DataListToolbar/DataListToolbar.jsx
+++ b/awx/ui_next/src/components/DataListToolbar/DataListToolbar.jsx
@@ -4,6 +4,7 @@ import styled from 'styled-components';
 
 import { t } from '@lingui/macro';
 import {
+  Button,
   Checkbox,
   Toolbar,
   ToolbarContent as PFToolbarContent,
@@ -14,7 +15,11 @@ import {
   DropdownPosition,
   KebabToggle,
 } from '@patternfly/react-core';
-import { SearchIcon } from '@patternfly/react-icons';
+import {
+  AngleDownIcon,
+  AngleRightIcon,
+  SearchIcon,
+} from '@patternfly/react-icons';
 import ExpandCollapse from '../ExpandCollapse';
 import Search from '../Search';
 import Sort from '../Sort';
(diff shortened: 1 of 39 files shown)


awx/ui_next/src/components/DataListToolbar/DataListToolbar.jsx
```diff
@@ -28,6 +33,9 @@ const ToolbarContent = styled(PFToolbarContent)`
 `;
 
 function DataListToolbar({
+  isAllExpanded,
+  showExpandAll = false,
+  onExpandAll,
```
keithjgrant: I wonder if we can reduce the number of props here a bit and get rid of `showExpandAll` by checking if `onExpandAll` is falsy (and same for `showSelectAll`/`onSelectAll`)

awx/ui_next/src/components/DataListToolbar/DataListToolbar.jsx
```diff
@@ -79,6 +86,26 @@ function DataListToolbar({
       clearFiltersButtonText={t`Clear all filters`}
     >
       <ToolbarContent>
+        {showExpandAll && (
```
keithjgrant: use `{ showExpandAll ? (...) : null }`

keithjgrant (approved):

nixocio:
e2e test triggered.

akus062381 (approved):

Issue #5000, opened 2019-10-11
Allow custom Ansible project structures


Timoses:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Since Ansible offers a lot of flexibility in the way that projects can be set up, it would be great if the `project_update.yml` would be exposed for user customization.

EDIT: As [stated](https://github.com/ansible/awx/issues/5000#issuecomment-541098848) `project_update.yml` is not the only place affected by this feature request.

Some examples why this may be required:
* We use an `Ansible` playbook which installs Ansible in various flavours (CLI, Dev workplace, AWX, ...). The `ansible.cfg` for each setup differs. Thus for a project to be set up it is required to run the `Ansible` play or at least use the Ansible `template` module to place the ansible.cfg in the right place (in case of AWX in the project root path).
* `requirements.yml` may not be set up as required by AWX currently. In our case we have a requirements.yml in each `plays/<play>/requirements.yml`. ansible-galaxy should be executed to place the required roles into `plays/<play>/roles/`.
* Custom plugin usage (mainly driven through ansible.cfg configuration)
* ...

As it is currently, AWX forces users to (some examples):
* place roles into `roles/requirements.yml`
* no bootstrapping options to e.g. pull content, modules, ... (Addition of collections probably does help at this point)

There are so many ways to use Ansible. Forcing users to one way only disrupts usability.

Currently, we will probably move towards overriding the `project_update.yml` in `/var/lib/awx/venv/awx/lib/python3.6/site-packages/awx/playbooks/project_update.yml` of the `awx_task` container.
EDIT: Unfortunately `awx/main/tasks.py` contains further logic which makes above approach unfeasible.

Any other way to achieve this? Any documentation on this I have missed?

Timoses:
I just found out that projects are copied (checked out with git) to `/tmp` directory and only the role requirements from `roles/requirements.yml` are used in `awx/main/tasks.py` (rendering all the `plays/<play>/requirements.yml` files as ignored).

This makes AWX quite unusable with custom Ansible project structures. Any way to circumvent this and simply "copy" the project path to the job temporary directory?

wenottingham:
If we start going down the road of managing execution environments, this wouldn't be done by hooking into project_update.yml; it would be done at a higher level.

Timoses:
Thanks @wenottingham . I adjusted the issue title to better reflect the issue at hand.

ffirg:
[...]

PR #10626, opened 2021-07-12
Allow requirements.yml/yaml files to be anywhere in project_path related #5000

dsglaser:
##### SUMMARY

This change searches the project_path for all instances of requirements.yml/yaml and adds them to a list. This list is then used to select which roles/requirements.yml/yaml or collections/requirements.yml/yaml are added. This removes the hardcoding of the requirements.yml/yaml in project_update.yml and allows users to have them in multiple locations.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API
 - UI - project.update.yml

##### AWX VERSION

```
19.2.2
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/playbooks/project_update.yml
+++ b/awx/playbooks/project_update.yml
@@ -192,7 +192,7 @@
           environment:
             ANSIBLE_FORCE_COLOR: false
             GIT_SSH_COMMAND: "ssh -o StrictHostKeyChecking=no"
-          loop: "{{ requirements_files.files | selectattr('path','match','.*/roles/requirements.yml$') | list }}"
+          loop: "{{ requirements_files.files | selectattr('path','match','.*/roles/requirements.ya?ml$') | list }}"
           loop_control:
             label: "{{ item.path }}"
       when: roles_enabled|bool
@@ -216,7 +216,7 @@
             # Put the local tmp directory in same volume as collection destination
             # otherwise, files cannot be moved accross volumes and will cause error
             ANSIBLE_LOCAL_TEMP: "{{projects_root}}/.__awx_cache/{{local_path}}/stage/tmp"
-          loop: "{{ requirements_files.files | selectattr('path','match','.*/collections/requirements.yml$') | list }}"
+          loop: "{{ requirements_files.files | selectattr('path','match','.*/collections/requirements.ya?ml$') | list }}"
           loop_control:
             label: "{{ item.path }}"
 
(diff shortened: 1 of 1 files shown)


wenottingham (changes requested):
This breaks caching as soon as any two paths have conflicting requirements.

wenottingham:
See also https://github.com/ansible/awx/issues/106#issuecomment-878289010 (cc @AlanCoding).

Given the move to execution environments, there is some level of diminishing returns to supporting arbitrary project layout.

dsglaser:
@wenottingham A way of prioritizing requirements would have to be built in then, yes? 

Even with execution environments, a project could have requirements in a different location in a project, right? Is the processing of requirements.yml being handled differently when execution environments are being used?

This PR came from a customer and we were requested to submit it.

wenottingham:
> @wenottingham A way of prioritizing requirements would have to be built in then, yes?

The caching code would need rewritten to cache differently, or ansible-galaxy would need to be smarter.

> Even with execution environments, a project could have requirements in a different location in a project, right? Is the processing of requirements.yml being handled differently when execution environments are being used?

Downloaded-at-runtime requirements would be done the same way. Built-in collections in an execution environment would be installed globally.

Rather than having content contstantly downloaded at runtime, for relatively static content, we'd want users to just embed it in an EE used for those runs.

> This PR came from a customer and we were requested to submit it.

Yes but that doesn't mean we will take it if it breaks use cases of the platform, and it does.

Take an example project:

- a/requirements.yml - specifies collection foo-1.2.3
- a/b/c/requirements.yml - specifies foo-2.4.5
- a/d/requirements.yml - specifies foo-0.9.7

This patch would populate the cache with foo-1.2.3, which would be used for *all* runs from the project. Things in a/b/c and /a/d would run with the wrong version of the collection.

dsglaser:
@wenottingham Understood. I've invited the actual contacts to contribute to this discussion. Obviously the PR as written won't work, but perhaps it is a discussion point for further development. 

In the current code, there is a fileglob for requirements.yml and requirements.yaml. I didn't see a check to make sure that both of these didn't exist or the same issue could be run into. Or perhaps it's somewhere else.

AlanCoding:
> In the current code, there is a fileglob for requirements.yml and requirements.yaml. I didn't see a check to make sure that both of these didn't exist or the same issue could be run into. Or perhaps it's somewhere else.

What's the concern with this? The fileglob is a means of looping, and if no matching files are found then it's like looping over an empty list - it doesn't do anything and that's the intended behavior for projects with no requirement files.

dsglaser:
Removing this PR after discussing with PM

PR #10636, opened 2021-07-13
Incorporates Minikube to devel environment

tchellomello:
##### SUMMARY

Incorporates Minikube to devel environment

This PR allows creating a smooth devel environment deploying `Minikube` connected to the AWX devel network interface using Docker. 

Furthermore, both `CredentialType`  and `Container Group` gets created automatically. 

Steps performed by the target `make docker-compose-container-group`

1. Downloads Minikube (currently supported Linux and MacOS)
2. Downloads `kubectl` (currently supported Linux and MacOS)
3. Starts Minikube using `docker` as backend
4. Creates a `Service Account`, `Role` and `RoleBinding` on Minikube
5. Exports Minikube `ServiceAccount` token to be used on AWX Credential
6. Creates Minikube AWX Credential
7. Creates Container Group to use Minikube backend
8. Starts regular AWX devel instances and connected with Minikube network

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
devel
```

##### ADDITIONAL INFORMATION
Quick step-by-step to use this PR:

```sh
$ make docker-compose-container-group-clean  # this will remove your database and any old Minikube already deployed
$ make docker-compose-container-group
```

Verifying some work:
```yaml

$ kubectl get serviceaccount awx-devel
NAME        SECRETS   AGE
awx-devel   1         130m

$ kubectl get roles
NAME        CREATED AT
awx-devel   2021-07-15T00:47:37Z

$ kubectl get rolebindings
NAME        ROLE             AGE
awx-devel   Role/awx-devel   131m
```
*Credential*
image

*Container Instance Group*
image

*After updating a Job Template and assigning to the Container Instance Group*
```bash
kubectl get pods -w
NAME                     READY   STATUS    RESTARTS   AGE
automation-job-6-w6rmh   0/1     Pending   0          0s
automation-job-6-w6rmh   0/1     Pending   0          0s
automation-job-6-w6rmh   0/1     ContainerCreating   0          0s
automation-job-6-w6rmh   1/1     Running             0          5s
automation-job-6-w6rmh   1/1     Terminating         0          6s
```

### DIFF
--- a/tools/docker-compose-minikube/minikube/tasks/main.yml
+++ b/tools/docker-compose-minikube/minikube/tasks/main.yml
@@ -1,0 +1,81 @@
+---
+- name: Create _sources directory
+  file:
+    path: "{{ sources_dest }}"
+    state: 'directory'
+    mode: '0700'
+
+# Linux block
+- block:
+    - name: Download Minikube
+      get_url:
+        url: "{{ minikube_url_linux }}"
+        dest: "{{ sources_dest }}/minikube"
+        mode: 0755
+
+    - name: Download Kubectl
+      get_url:
+        url: "{{ kubectl_url_linux }}"
+        dest: "{{ sources_dest }}/kubectl"
+        mode: 0755
+  when:
+    - ansible_architecture == "x86_64"
+    - ansible_system == "Linux"
+
+# MacOS block
+- block:
+    - name: Download Minikube
+      get_url:
+        url: "{{ minikube_url_macos }}"
+        dest: "{{ sources_dest }}/minikube"
+        mode: 0755
+
+    - name: Download Kubectl
+      get_url:
+        url: "{{ kubectl_url_macos }}"
+        dest: "{{ sources_dest }}/kubectl"
+        mode: 0755
+  when:
+    - ansible_architecture == "x86_64"
+    - ansible_system == "Darwin"
+
+- name: Starting Minikube
+  shell: "{{ sources_dest }}/minikube start --driver={{ driver }} --install-addons=true --addons=ingress"
+
+- name: Create ServiceAccount and clusterRoleBinding
+  k8s:
+    apply: true
+    definition: "{{ lookup('template', 'rbac.yml.j2') }}"
+
+- name: Retrieve serviceAccount secret name
+  k8s_info:
+    kind: ServiceAccount
+    namespace: '{{ minikube_service_account_namespace }}'
+    name: '{{ minikube_service_account_name }}'
+  register: service_account
+
+- name: Register serviceAccount secret name
+  set_fact:
+    _service_account_secret_name: '{{ service_account["resources"][0]["secrets"][0]["name"] }}'
+  when:
+    - service_account["resources"][0]["secrets"] | length
+    - '"name" in service_account["resources"][0]["secrets"][0]'
+
+- name: Retrieve bearer_token from serviceAccount secret
+  k8s_info:
+    kind: Secret
+    namespace: '{{ minikube_service_account_namespace }}'
+    name: '{{ _service_account_secret_name }}'
+  register: _service_account_secret
+
+- name: Load Minikube Bearer Token
+  set_fact:
+    service_account_token: '{{ _service_account_secret["resources"][0]["data"]["token"] | b64decode }}'
+  when:
+    - _service_account_secret["resources"][0]["data"] | length
+
+- name: Render minikube credential JSON template
+  template:
+    src: bootstrap_minikube.py.j2
+    dest: "{{ sources_dest }}/bootstrap_minikube.py"
+    mode: '0600'
--- a/tools/docker-compose-minikube/minikube/templates/bootstrap_minikube.py.j2
+++ b/tools/docker-compose-minikube/minikube/templates/bootstrap_minikube.py.j2
@@ -1,0 +1,42 @@
+# Create Openshift/Kubernetes credential for Minikube
+# This script gets called by the bootstrap_development process
+# awx-manage shell_plus --quiet < bootstrap_minikube.py
+
+from awx.main.utils.encryption import encrypt_field
+
+NAME = 'Minikube'
+
+POD_SPEC = """apiVersion: v1
+kind: Pod
+metadata:
+  namespace: {{ minikube_service_account_namespace }}
+spec:
+  containers:
+    - image: 'quay.io/ansible/awx-ee:devel'
+      name: worker
+      args:
+        - ansible-runner
+        - worker
+        - '--private-data-dir=/runner'"""
+
+# Creates Minikube credential
+if not Credential.objects.filter(name=NAME).count():
+    cred = Credential()
+    cred.name = NAME
+    cred.credential_type = CredentialType.objects.get(name='OpenShift or Kubernetes API Bearer Token')
+    cred.description = 'Minikube Devel'
+    cred.inputs['host'] =  'https://minikube:8443'
+    cred.inputs['verify_ssl'] =  False
+    cred.inputs['bearer_token'] = '{{ service_account_token }}'
+    encrypt_field(cred, 'bearer_token', secret_key=settings.SECRET_KEY)
+    cred.save()
+
+# Create Container Group for Minikube
+if not InstanceGroup.objects.filter(name=NAME).count():
+    ccgrp = InstanceGroup()
+    ccgrp.name = NAME
+    ccgrp.credential = cred
+    ccgrp.pod_spec_override = POD_SPEC
+    ccgrp.is_container_group = True
+    ccgrp.save()
+
--- a/tools/docker-compose-minikube/minikube/templates/bootstrap_minikube.py.j2
+++ b/tools/docker-compose-minikube/minikube/templates/bootstrap_minikube.py.j2
@@ -1,8 +1,10 @@
 # Create Openshift/Kubernetes credential for Minikube
 # This script gets called by the bootstrap_development process
-# awx-manage shell_plus --quiet < bootstrap_minikube.py
+# awx-manage shell < bootstrap_minikube.py
 
 from awx.main.utils.encryption import encrypt_field
+from awx.main.models import Credential, CredentialType, InstanceGroup
+from django.conf import settings
 
 NAME = 'Minikube'
 
--- a/tools/docker-compose-minikube/minikube/templates/rbac.yml.j2
+++ b/tools/docker-compose-minikube/minikube/templates/rbac.yml.j2
@@ -1,0 +1,38 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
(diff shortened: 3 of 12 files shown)


tools/docker-compose/supervisor.conf
```diff
@@ -95,6 +95,16 @@ redirect_stderr=true
 stdout_events_enabled = true
 stderr_events_enabled = true
 
+[program:minikube-credentials]
+command = make minikube-credentials
```
shanemcd: This feels very wrong 😄 Let's chat more tomorrow about what you're trying to accomplish here.

tools/docker-compose-minikube/minikube/tasks/main.yml
```diff
@@ -0,0 +1,86 @@
+---
+- name: Set home binary path directory
+  set_fact:
+    home_bin: "{{ lookup('env','HOME') }}/bin"
+
+- name: Create home binary path directory
+  file:
+    path: "{{ home_bin }}"
```
shanemcd: Not a huge fan of putting things in people's home directory like this. Consider putting in under a ignored / hidden directory in the root of the repo.

tools/docker-compose-minikube/minikube/templates/api_credential.json.j2
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "Minikube",
+    "credential_type": "16",
```
shanemcd: This seems likely to break / not be consistent. We'll probably want to query the API to get the ID.

tools/docker-compose-minikube/minikube/templates/rbac.yml.j2
```diff
@@ -0,0 +1,19 @@
...
+  namespace: default
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ minikube_service_account_name }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
```
shanemcd: Not a fan of this. Let's be very explicit and mindful about which permissions we need, even in the dev env. This prevents surprises later. https://github.com/ansible/awx-operator/blob/devel/roles/installer/templates/service_account.yaml.j2

tools/docker-compose-minikube/minikube/tasks/main.yml
```diff
@@ -0,0 +1,86 @@
...
+    minikube_ca_cert: '{{ _service_account_secret["resources"][0]["data"]["ca.crt"] | b64decode }}'
+    service_account_token: '{{ _service_account_secret["resources"][0]["data"]["token"] | b64decode }}'
+  when:
+    - _service_account_secret["resources"][0]["data"] | length
+
+- name: Render minikube credential JSON template
+  template:
+    src: api_credential.json.j2
+    dest: "api_credential.json"
+    mode: '0600'
```
kdelee, line 81: maybe all this regarding the credential can be carved out into a separate playbook and then you use the awx.awx collection to create the credential, and then have the playbook run after the awx API is up or just leave it as a tool someone can run given they have minikube and awx up and running. Using the awx.awx.credential module would implicate deleting tools/docker-compose-minikube/minikube/templates/api_credential.json.j2 and the service definition you have in tools/docker-compose/supervisor.conf 
https://docs.ansible.com/ansible/latest/collections/awx/awx/credential_module.html#ansible-collections-awx-awx-credential-module

tchellomello: Having a separate tool to be executed once the AWX API is up will facilitate a lot of things. If we are good with that extra step, I can adjust this into a separate play.

tools/docker-compose-minikube/minikube/templates/rbac.yml.j2
kdelee: yeah we used the default SA and gave it cluster-admin because we just wanted to try something fast, but creating a dedicated SA would be ideal. if it is not in the default namespace we have to modify the ContainerGroup podspec.

tchellomello: Yes, that makes sense. Thanks @shanemcd

tools/docker-compose/supervisor.conf
tchellomello: Yes, it does :) I actually took a different approach which I believe will simplify the process and work better. I'll update the PR shortly

tools/docker-compose-minikube/minikube/templates/api_credential.json.j2
tchellomello: Yes, the new approach addresses this ugly part ;)

tools/docker-compose-minikube/minikube/tasks/main.yml
tchellomello: OK, it makes sense. Created a new space under the repo and ignored it on git.

tools/docker-compose-minikube/minikube/tasks/main.yml
tchellomello: So to avoid an extra step, decided to call it directly on the `boostrap_development.py` script which works transparently.  Let me know what you guys think. 

```bash
if [[ -n "$MINIKUBE_UP" ]]; then
    awx-manage shell_plus --quiet < /awx_devel/tools/docker-compose-minikube/_sources/bootstrap_minikube.py
fi
```

tools/docker-compose-minikube/minikube/templates/bootstrap_minikube.py.j2
```diff
@@ -0,0 +1,42 @@
...
+metadata:
+  namespace: default
+spec:
+  containers:
+    - image: 'quay.io/ansible/awx-ee:devel'
+      name: worker
+      args:
+        - ansible-runner
+        - worker
+        - '--private-data-dir=/runner'"""
```
kdelee, line 20: Is there a particular reason why you are overriding the podspec? You can just take the default, the only thing we really would want to override here would be the `namespace` if we didn't want to use an SA in the default namespace.

kdelee: or is it just because you are directly creating the instace group here in the shell that the default podspec you normally get through the API is not populated yet?

tools/docker-compose-minikube/minikube/templates/bootstrap_minikube.py.j2
```diff
@@ -0,0 +1,42 @@
...
+# awx-manage shell_plus --quiet < bootstrap_minikube.py
+
+from awx.main.utils.encryption import encrypt_field
+
+NAME = 'Minikube'
+
+POD_SPEC = """apiVersion: v1
+kind: Pod
+metadata:
+  namespace: default
```
kdelee: If you want to be very consistent, you could figure out how to pass in what the `namespace` should be here based on `minikube_service_account_namespace` in the playbook, because if it is something other than default the user will get permission denied because here is where we specify what namespace to deploy the pod in

tchellomello: Yes, that is a good point. :)  Thank you

tools/docker-compose/bootstrap_development.sh
```diff
@@ -34,3 +34,8 @@ awx-manage register_default_execution_environments
 mkdir -p /awx_devel/awx/public/static
 mkdir -p /awx_devel/awx/ui/static
 mkdir -p /awx_devel/awx/ui_next/build/static
+
+# Create resource entries when using Minikube
+if [[ -n "$MINIKUBE_UP" ]]; then
+    awx-manage shell_plus --quiet < /awx_devel/tools/docker-compose-minikube/_sources/bootstrap_minikube.py
```
shanemcd: 😎

tools/docker-compose/README.md
```diff
@@ -213,3 +213,18 @@ Certain features or bugs are only applicable when running a cluster of AWX nodes
 `CLUSTER_NODE_COUNT` is configurable and defaults to 1, effectively a non-clustered AWX.
 
 Note that you may see multiple messages of the form `2021-03-04 20:11:47,666 WARNING  [-] awx.main.wsbroadcast Connection from awx_2 to awx_5 failed: 'Cannot connect to host awx_5:8013 ssl:False [Name or service not known]'.`. This can happen when you bring up a cluster of many nodes, say 10, then you bring up a cluster of less nodes, say 3. In this example, there will be 7 `Instance` records in the database that represent AWX instances. The AWX development environment mimics the VM deployment (vs. kubernetes) and expects the missing nodes to be brought back to healthy by the admin. The warning message you are seeing is all of the AWX nodes trying to connect the websocket backplane. You can manually delete the `Instance` records from the database i.e. `Instance.objects.get(hostname='awx_9').delete()` to stop the warnings.
+
+### Start with Minikube
+
+To bring up a 1 node AWX + minikube that is accessible from AWX run the following.
+
+```bash
+(host)$ make docker-compose-minikube
```
shanemcd: I am not a fan of the name of this make target. I find it confusing given we already have https://github.com/ansible/awx/blob/devel/docs/development/minikube.md and this is doing something different.

tools/docker-compose/README.md
```diff
@@ -213,3 +213,18 @@ Certain features or bugs are only applicable when running a cluster of AWX nodes
...
+```
+
+Alternatively, you can set the env var `MINIKUBE_UP=true` to use the default dev env bring up. his way you can use other env flags like the cluster node count.
```
shanemcd: Instead of `MINIKUBE_UP`, consider something like `MINIKUBE_CONTAINER_GROUP`.

tools/docker-compose/README.md
tchellomello: Renamed it to `make docker-compose-container-group`

tools/docker-compose/README.md
tchellomello: Done!

tools/docker-compose-minikube/minikube/tasks/main.yml
```diff
@@ -0,0 +1,81 @@
+---
+- name: Create _sources directory
+  file:
+    path: "{{ sources_dest }}"
+    state: 'directory'
+    mode: '0700'
+
```
sarabrajsingh, line 7: need to pip install the openshift module

tchellomello: Yes, it's required however I did not include on the playbook because we will not know if the devel wants to use a venv or global python, so left for the devel to hand the python stuff, but if we think installing is important, we can add it

shanemcd (approved):

beeankha (approved):

PR #10639, opened 2021-07-14
Fix inaccurate translation

weidongkl:
##### SUMMARY

Fix some inaccurate translation in zh meesage.po

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - UI

##### AWX VERSION

```
devel branch
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/ui_next/src/locales/zh/messages.po
+++ b/awx/ui_next/src/locales/zh/messages.po
@@ -42,7 +42,7 @@ msgstr "/ (project root)"
 #: screens/Template/JobTemplateDetail/JobTemplateDetail.jsx:106
 #: screens/Template/shared/JobTemplateForm.jsx:211
 msgid "0 (Normal)"
-msgstr "0（普通）"
+msgstr "0（正常）"
 
 #: screens/Inventory/InventorySourceDetail/InventorySourceDetail.jsx:102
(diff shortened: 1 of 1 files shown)


keithjgrant:
@rooftopcellist Could you review this PR?

awx/ui_next/src/locales/zh/messages.po
```diff
@@ -8843,11 +8843,11 @@ msgstr "周"
 
 #: components/Schedule/shared/FrequencyDetailSubform.jsx:449
 msgid "Weekday"
-msgstr "周中日"
+msgstr "周内"
```
rooftopcellist: @weidongkl for the translation of "Weekday", perhaps it might be better to translated as "工作日" or "周中"?  This string is describing the days during the work week (Monday - Friday).  This was the feedback from one of our internal translators.  What do you think?


 if it means "Monday to Friday"

weidongkl: I agree that  "工作日" are the better  translation

rooftopcellist (approved):

awx/ui_next/src/locales/zh/messages.po
rooftopcellist: Ok, great.  Thank you @weidongkl!

PR #10658, opened 2021-07-15
Updates Several Dev Dependencies

AlexSCorey:
##### SUMMARY
This PR updates ESLint, its related dependencies and Prettier.
 
Some noted changes:
1)ESLint no longer links when we spread props (ie. `{...props}`) but I've ignored that rule in the code base. [you can read more here](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-props-no-spreading.md)
2) ESLint does not like using `<Fragment>` and wants us to use `<>` instead.
3) ESLint doesn't like explicit return statements in arrow functions
4) Prettier now wants us to use parentheses around arguments in functions, even if there is only 1.  (ie. `const functionName = (oneArgument) => (do stuff here)`
5) I've also deleted `ExpandCollapse.jx` and associated files/lines of code because we don't use it and we have expand collapse functionality on lists already.

##### ISSUE TYPE
dependency bump

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION


AlexSCorey:
e2e tests triggered

awx/ui_next/.eslintrc
```diff
@@ -12,7 +12,6 @@
   "extends": [
     "airbnb",
     "prettier",
-    "prettier/react",
```
nixocio, line 15: 👍

nixocio (approved):
Good job @AlexSCorey, I was able to re-generate the same `package.json` and `package.lock.json`. 

See: https://github.com/nixocio/awx/tree/ui_dep_update

And the other changes seem to be adequate to the new lint version.

nixocio:
e2e tests triggered.

tiagodread (approved):

PR #10664, opened 2021-07-15
Update docker-compose

beeankha:
Connecting https://github.com/ansible/awx/issues/10645

Deploys 1 control and 1 execution node!

### DIFF
--- a/tools/docker-compose-cluster/awx-1-receptor-standalone.conf
+++ b/tools/docker-compose-cluster/awx-1-receptor-standalone.conf
@@ -11,7 +11,7 @@
 #    rootcas: /etc/receptor/certs/ca.crt
 
 - tcp-peer:
-    address: tools_awx_1:4444
+    address: awx_1:4444
     redial: true
     #tls: mutual-tls-client
 
--- a/tools/docker-compose-cluster/awx-1-receptor-standalone.conf
+++ b/tools/docker-compose-cluster/awx-1-receptor-standalone.conf
@@ -1,0 +1,25 @@
+---
+- node:
+    id: receptor_1
+
+- log-level: info
+
+#- tls-client:
+#    name: mutual-tls-client
+#    cert: /etc/receptor/certs/receptor_1.crt
+#    key: /etc/receptor/certs/receptor_1.key
+#    rootcas: /etc/receptor/certs/ca.crt
+
+- tcp-peer:
+    address: tools_awx_1:4444
+    redial: true
+    #tls: mutual-tls-client
+
+- work-command:
+    worktype: ansible-runner
+    command: ansible-runner
+    params: worker
+    allowruntimeparams: true
+
+- control-service:
+    service: control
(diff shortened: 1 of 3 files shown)


jladdjr:
Think I may have found the issue preventing our jobs from running:
```
awx-dispatcher stdout | 2021-07-19 23:46:09,765 DEBUG    [9e3a04d9] awx.analytics.job_lifecycle job-1 needs capacity
```

jladdjr:
I'm wondering if there's something funky going on with how we're getting memory info for the receptor execution nodes - here maybe:
https://github.com/beeankha/awx/blob/49fe58807da4d31fc207acc2c0919276e7e09fe7/awx/main/tasks.py#L3066

The execution node has cpu capacity but _zero_ mem capacity. (Meanwhile, the awx node has both).

From what I can tell, this doesn't have anything to do with the actual specs of the containers. `docker inspect` for awx and the receptor nodes shows similar info, as do the /etc/{cpu,mem}info files on both containers (2 cpus, ~6GB memory in both cases).

jladdjr:
Hrmm, maybe not. Just set an sdb trace there and saw mem info.

tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
```diff
@@ -27,6 +27,7 @@ services:
     links:
       - postgres
       - redis_{{ container_postfix }}
+      - receptor_{{ container_postfix }}
```
jladdjr: `receptor_1`

jladdjr (approved):

tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
```diff
@@ -62,6 +63,14 @@ services:
     entrypoint: ["redis-server"]
     command: ["/usr/local/etc/redis/redis.conf"]
 {% endfor %}
+receptor_1:
+  image: quay.io/ansible/awx-ee:devel
```
AlanCoding: when the other merge happens, we'll want to change this to the main awx image.

This is because you can't really have the code compatible with both. The awx-ee is missing `podman`, which means that you can't run with isolation. The awx image lacks `ansible`, which means that you can't run anything _outside_ of the container.

tools/docker-compose-cluster/awx-1-receptor-standalone.conf
```diff
@@ -0,0 +1,25 @@
...
+    #tls: mutual-tls-client
+
+- work-command:
+    worktype: ansible-runner
+    command: ansible-runner
+    params: worker
+    allowruntimeparams: true
+
+- control-service:
+    service: control
```
AlanCoding: This file should definitely not go in the `tools/docker-compose-cluster` folder, because it's for docker-compose, which has its own folder.

If we're moving it anyway, it might also be good to name it something other than "receptor". We have historically called these "receptor nodes", but in planning we veered away from that, and "execution nodes" became more of the nomenclature. I would like a little more clarity on the language we use, but that's more the responsibility of the people working on adding the `node_type` field.

So we'll have 1 control node and 1 execution node. This is the receptor.conf for the execution node. The control node has the `tools/docker-compose/receptor.conf` file, which is statically ADDed to the Dockerfile. I don't know if that makes sense. Clearly the control node conf file won't have `tcp-peer`, but it seems like it doesn't need that?

AlanCoding (approved):

PR #10688, opened 2021-07-20
Run capacity checks with container isolation

AlanCoding:
This requires swapping out the container images
  for the execution nodes from awx-ee to the awx image

For completeness, the hop node image is switched to the raw
  receptor image

A few outright bugs are fixed here
  memory calculation just was not right at all
  the execution_capacity calculation was reverse of intention

Drop in a few TODOs about error handling from debugging

----

I have this running a real job in the development environment, so that's as complete as I'm going for at the moment.

### DIFF
--- a/awx/main/models/ha.py
+++ b/awx/main/models/ha.py
@@ -207,7 +207,7 @@ class InstanceGroup(HasPolicyEditsMixin, BaseModel, RelatedJobsMixin):
     @property
     def execution_capacity(self):
         # TODO: update query to exclude based on node_type field
-        return sum([inst.capacity for inst in self.instances.exclude(version__startswith='ansible-runner-')])
+        return sum([inst.capacity for inst in self.instances.filter(version__startswith='ansible-runner-')])
 
     @property
     def jobs_running(self):
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -59,7 +59,6 @@ from receptorctl.socket_interface import ReceptorControl
 from dateutil.parser import parse as parse_date
 
 # AWX
-from awx import MODE
 from awx import __version__ as awx_application_version
 from awx.main.constants import PRIVILEGE_ESCALATION_METHODS, STANDARD_INVENTORY_UPDATE_ENV, MINIMAL_EVENTS
 from awx.main.access import access_registry
@@ -103,6 +102,7 @@ from awx.main.utils.common import (
     cleanup_new_process,
     create_partition,
     get_cpu_capacity,
+    get_mem_capacity,
     get_system_task_capacity,
 )
 from awx.main.utils.execution_environments import get_default_execution_environment, get_default_pod_spec, CONTAINER_ROOT, to_container_path
@@ -466,6 +466,8 @@ def cluster_node_heartbeat():
         if inst.hostname == settings.CLUSTER_HOST_ID:
             this_inst = inst
             instance_list.remove(inst)
+        elif inst.version.startswith('ansible-runner'):  # TODO: use proper field when introduced
+            continue
         elif inst.is_lost(ref_time=nowtime):
             lost_instances.append(inst)
             instance_list.remove(inst)
@@ -1773,7 +1775,6 @@ class RunJob(BaseTask):
                 ]
             )
 
-        params['process_isolation'] = False if MODE == 'development' else True
         return params
 
     def pre_run_hook(self, job, private_data_dir):
@@ -2847,11 +2848,6 @@ class RunAdHocCommand(BaseTask):
         d[r'Password:\s*?$'] = 'ssh_password'
         return d
 
-    def build_execution_environment_params(self, instance, private_data_dir):
-        params = super(RunAdHocCommand, self).build_execution_environment_params(instance, private_data_dir)
-        params['process_isolation'] = False if MODE == 'development' else True
-        return params
-
 
 @task(queue=get_local_queuename)
 class RunSystemJob(BaseTask):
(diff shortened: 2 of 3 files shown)


tools/docker-receptor.yml
```diff
@@ -17,7 +17,7 @@ services:
     volumes:
       - "./docker-compose-cluster:/etc/receptor"
   receptor-hop:
-    image: quay.io/ansible/awx-ee:devel
+    image: quay.io/project-receptor/receptor:latest
```
AlanCoding, line 20: @fosterseth you should find this interesting.

The jump nodes _really don't need anything_ other than receptor. It's the academically correct choice, then, to use the pure receptor image. This gets us much closer to what we're going for. You can't run jobs on this node, or use podman inside of it.

And most importantly, it actually works too.

tools/docker-receptor.yml
```diff
@@ -29,32 +29,38 @@ services:
...
+    user: "1000"
     container_name: tools_receptor_1
     hostname: receptor-1
     command: 'receptor --config /etc/receptor/awx-1-receptor.conf'
     links:
       - receptor-hop
     volumes:
       - "../../docker-compose-cluster:/etc/receptor"
+      - "/sys/fs/cgroup:/sys/fs/cgroup"
+    privileged: true
```
AlanCoding, line 42: @shanemcd Reusing your hacks to run podman inside of the container here.

I know that having the execution nodes on `quay.io/awx/awx_devel:devel` is non-ideal, and this is the next step we've planned:

https://github.com/ansible/awx/issues/10691

These execution nodes need a _subset_ of what's in the `quay.io/awx/awx_devel:devel` image. So we can get by without introducing a new image. However, I can see us refactoring this whole image ecosystem, and if we are doing that, we could start out doing things the right way.

awx/main/tasks.py
```diff
@@ -466,6 +466,8 @@ def cluster_node_heartbeat():
         if inst.hostname == settings.CLUSTER_HOST_ID:
             this_inst = inst
             instance_list.remove(inst)
+        elif inst.version.startswith('ansible-runner'):  # TODO: use proper field when introduced
+            continue
```
AlanCoding, line 470: Jobs won't really work in the main branch, and this is the necessary fix to resolve that.

Execution nodes won't update capacity on a regular basis whatsoever. This set capacity of other nodes to 0 if they their heartbeat did not work. Execution nodes have no heartbeat, so that's the reason for excluding them here.

awx/main/tasks.py
```diff
@@ -3085,19 +3088,26 @@ def update_model(self, *args, **kw):
...
                 instance.mem_capacity,
             )
             instance.version = task.version
             instance.save(update_fields=['capacity', 'version', 'cpu', 'memory', 'cpu_capacity', 'mem_capacity'])
+            logger.info(f'Updated capacity of {node} to cpu: {instance.cpu_capacity} mem: {instance.mem_capacity}')
+        else:
+            # TODO: error handling like we do with jobs
+            # receptorctl work results
+            # receptorctl work list
+            logger.info(f'Capacity check not successful for execution node {node}')
```
AlanCoding, line 3110: should probably be a warn

tools/docker-receptor.yml
fosterseth: perfection

jladdjr: ditto, this makes sense

awx/main/models/ha.py
```diff
@@ -207,7 +207,7 @@ def capacity(self):
     @property
     def execution_capacity(self):
         # TODO: update query to exclude based on node_type field
-        return sum([inst.capacity for inst in self.instances.exclude(version__startswith='ansible-runner-')])
```
jladdjr, line 210: oh man, no wonder things weren't working

jladdjr: good catch

awx/main/tasks.py
jladdjr: If they don’t have a heartbeat, how do we know that they are still healthy?
Also, won’t they use up capacity as they execute jobs?

awx/main/tasks.py
```diff
@@ -3033,7 +3034,7 @@ def check_heartbeat(cls, node):
             'settings': {
                 "container_image": get_default_execution_environment().image,
                 "container_options": ['--user=root'],
-                "process_isolation": False if MODE == 'development' else True,
+                "process_isolation": True,
```
jladdjr, line 3037: @shanemcd Do we want to make this configurable at some point or are we thinking that execution nodes will always spin up containers locally? Guessing the latter, but wasn't 100% sure.

awx/main/tasks.py
```diff
@@ -3085,19 +3088,26 @@ def update_model(self, *args, **kw):
         res = receptor_job.run(work_type='ansible-runner')
         if res.status == 'successful':
             cpu = get_cpu_capacity(task.cpus)
-            mem = get_cpu_capacity(task.mem_mb)
+            mem = get_mem_capacity(task.mem_mb * 1000000)
```
jladdjr, line 3091: how.. did I miss this. good catch.

jladdjr (approved):

beeankha (approved):

jladdjr (approved):

jladdjr:
Requested auto-merge. All checks are pending atm. I'm still coming up-to-speed on the new git actions, so not sure when everything will go through.

AlanCoding:
I think that the normal checks may not run when against a non-default branch. That's probably set in some configuration somewhere.

AlanCoding:
hmm..

```
[alancoding@alan-red-hat awx]$ git push awx mesh_code
Enumerating objects: 17, done.
Counting objects: 100% (17/17), done.
Delta compression using up to 8 threads
Compressing objects: 100% (9/9), done.
Writing objects: 100% (9/9), 2.37 KiB | 2.37 MiB/s, done.
Total 9 (delta 8), reused 0 (delta 0), pack-reused 0
remote: Resolving deltas: 100% (8/8), completed with 8 local objects.
remote: error: GH006: Protected branch update failed for refs/heads/mesh_code.
remote: error: 6 of 6 required status checks are expected.
To https://github.com/ansible/awx.git
 ! [remote rejected]       mesh_code -> mesh_code (protected branch hook declined)
error: failed to push some refs to 'https://github.com/ansible/awx.git'
```
looks like it won't let me just push manually. So I guess we're going to have to actually fix this.

Issue #10663, opened 2021-07-15
When creating a new host auto-populate inventory field is not working


nixocio:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

When creating a new host auto-populate inventory field is not working.

### AWX version

devel

### Installation method

docker for mac

### Modifications

no

### Steps to reproduce

1 - Attempt to create a new host - when there is just one inventory available

<img width="1497" alt="image" src="https://user-images.githubusercontent.com/9053044/125836821-12a80c51-250c-4e00-8a91-b7634bddb644.png">

<img width="1484" alt="image" src="https://user-images.githubusercontent.com/9053044/125836862-1d65b6ae-17fb-4948-b201-c8228c3c2457.png">

### Expected results

Inventory lookup field to be auto-populated

### Actual results

Inventory lookup field is not auto-populated

tiagodread:
PR already merged

PR #10692, opened 2021-07-20
Add auto-populate inventory host form

nixocio:
Add auto-populate inventory host form

closes:https://github.com/ansible/awx/issues/10663

### DIFF
--- a/awx/ui_next/src/components/Lookup/InventoryLookup.js
+++ b/awx/ui_next/src/components/Lookup/InventoryLookup.js
@@ -5,6 +5,7 @@ import { t } from '@lingui/macro';
 import { InventoriesAPI } from 'api';
 import { Inventory } from 'types';
 import useRequest from 'hooks/useRequest';
+import useAutoPopulateLookup from 'hooks/useAutoPopulateLookup';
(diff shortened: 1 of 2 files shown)


awx/ui_next/src/components/Lookup/InventoryLookup.js
```diff
@@ -78,7 +86,7 @@ function InventoryLookup({
           Boolean(actionsResponse.data.actions.POST) || isOverrideDisabled,
       };
       // eslint-disable-next-line react-hooks/exhaustive-deps
```
AlexSCorey, line 88: I think we can remove this comment now.

AlexSCorey (approved):
This looks great.  I'm approving, but we should remove that old comment turning off es ling exhaustive dependencies array.

awx/ui_next/src/components/Lookup/InventoryLookup.js
nixocio: That comment was added because of `isOverrideDisabled` and we will need to keep it.

keithjgrant (approved):

tiagodread (approved):

Issue #8586, opened 2020-11-13
StatusLabel component text isn't translated


mabashian:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
https://github.com/ansible/awx/blob/devel/awx/ui_next/src/components/StatusLabel/StatusLabel.jsx#L49

If we take this approach of splitting the status string as it comes back via the API then this label will always be in English.

I think we're going to have to build a map of all possible labels that are marked for translation instead.

PR #10696, opened 2021-07-20
Create map of strings to be translated

nixocio:
Create map of strings to be translated to StatusLabel

closes:https://github.com/ansible/awx/issues/8586


AlexSCorey (changes requested):
Looks good but we might want to have a test that ensures it returns the right status.

nixocio:
@AlexSCorey we already had tests verifying that string - label, I expanded a missing case. Thanks.

AlexSCorey (approved):
Good to go

akus062381 (approved):

PR #10710, opened 2021-07-21
[WIP] add e2e tests to ci.yml

tiagodread:
##### SUMMARY
Closes https://github.com/ansible/tower-qa/issues/6642

Add e2e test execution in CI job


.github/workflows/ci.yml
```diff
@@ -171,3 +171,36 @@ jobs:
         run: |
           docker run -u $(id -u) --rm -v ${{ github.workspace}}:/awx_devel/:Z \
             --workdir=/awx_devel ghcr.io/${{ github.repository_owner }}/awx_devel:${{ github.base_ref }} make ui-test
+
+  e2e-test:
```
shanemcd: Unless we want this to run on every PR, I think this should be moved into its own workflow rather than get added here.

tiagodread: Ok, I think the idea is run on every PR right? Any reason why we shouldn't run for every PR?

.github/workflows/ci.yml
```diff
@@ -171,3 +171,36 @@ jobs:
...
+      - uses: actions/checkout@v2
+
+      - name: Checkout tower-qa
+        uses: actions/checkout@v2
+        with:
+          path: tower-qa
+
+      - name: Log in to gcr.io
+        run: |
+          echo "${{ secrets.GCR_PULL_SECRET }}" | docker login gcr.io -u _json_key --password-stdin
```
tiagodread: @shanemcd I think we'll need add the secret `secrets.GCR_PULL_SECRET`

.github/workflows/ci.yml
tiagodread: Ok, moved this to another workflow and will run when the label `state:needs_test` is applied to the PR.

PR #10711, opened 2021-07-21
Install boto to ci ansible python env

jakemcdermott:
##### SUMMARY
Addresses a post-merge [schema upload failure](https://github.com/ansible/awx/runs/3118168665) on devel.

I verified [the approach](https://github.com/ansible/awx/blob/pip-boto/.github/workflows/upload_schema.yml#L39) by [reproducing the failure](https://github.com/ansible/awx/actions/runs/1054176726/workflow#L21-L23) with a temporary workflow that ran on this PR and then confirming that the error was [resolved](https://github.com/ansible/awx/actions/runs/1054150052/workflow#L20-L22) with another temporary workflow that ran on this pr.

### DIFF
--- a/.github/workflows/upload_schema.yml
+++ b/.github/workflows/upload_schema.yml
@@ -36,7 +36,7 @@ jobs:
           AWS_SECRET_KEY: ${{ secrets.AWS_SECRET_KEY }}
           AWS_REGION: 'us-east-1'
         run: |
-          pip install boto3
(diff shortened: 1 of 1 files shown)


chrismeyersfsu (approved):

AlanCoding:
> Addresses a post-merge [schema upload failure](https://github.com/ansible/awx/runs/3118168665) on devel.

Too bad we can't follow because all the results are expired...

PR #10712, opened 2021-07-21
Making it possible to list workflow jobs

ranmses:
##### SUMMARY
Had the need to list workflow jobs, but couldn't with the resources provided by the job_list.py file. This one does that.
Please feel free to accept or reject it, I'm just sharing the solution that worked for me in case it can be useful for others.
I also recommend to review and test it, since my AWX version is old (9.3.0).

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awx_collection/plugins/modules/workflow_job_list.py
+++ b/awx_collection/plugins/modules/workflow_job_list.py
@@ -1,0 +1,126 @@
+#!/usr/bin/python
+# coding: utf-8 -*-
+
+# (c) 2021, Ranmsés <ranmses@gmail.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+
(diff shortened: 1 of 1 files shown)


shanemcd:
```
N/A                                  | workflow_job_list          | all_pages                                          |     |   X    | Failed, why does this module have no endpoint
614
N/A                                  | workflow_job_list          | page                                               |     |   X    | Failed, why does this module have no endpoint
615
N/A                                  | workflow_job_list          | query                                              |     |   X    | Failed, why does this module have no endpoint
616
N/A                                  | workflow_job_list          | status                                             |     |   X    | Failed, why does this module have no endpoint
```

AlanCoding:
I'm fairly strongly opposed to this, because the use can can be satisfied by the lookup plugin. We have gone this direction for a number of problems.

https://github.com/ansible/awx/blob/devel/awx_collection/plugins/lookup/controller_api.py

We appreciate the contribution, and I hope the lookup plugin can help develop a clean solution to your problem.

PR #10727, opened 2021-07-22
Code changes to support execution-only nodes in receptor mesh

AlanCoding:
##### SUMMARY
This adds support for auto-discovery of receptor nodes through a capacity check that is spawned out of the periodic heartbeat task.

Testing has been in the development environment so far, `make docker-compose` and `make docker-compose-cluster`.

If a job (through instance group assignment) is sent to one of the auto-discovered execution node, then it is transmitted by ansible-runner over receptor to that node, and results are streamed back to the control node.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API
 - UI

##### AWX VERSION
```
19.2.2
```


awx/settings/development.py
```diff
@@ -125,6 +125,8 @@
         }
     }
 
+CELERYBEAT_SCHEDULE['cluster_heartbeat']['schedule'] = timedelta(seconds=10)
```
AlanCoding: This is not really necessary, and sooner or later I'd like to delete it.

awx/api/views/__init__.py
```diff
@@ -370,15 +370,6 @@ class InstanceDetail(RetrieveUpdateAPIView):
...
     serializer_class = serializers.InstanceSerializer
 
-    def update(self, request, *args, **kwargs):
-        r = super(InstanceDetail, self).update(request, *args, **kwargs)
-        if status.is_success(r.status_code):
-            obj = self.get_object()
-            obj.refresh_capacity()
-            obj.save()
-            r.data = serializers.InstanceSerializer(obj, context=self.get_serializer_context()).to_representation(obj)
-        return r
```
AlanCoding, line 380: Do we still need to run this for control / hybrid nodes? Maybe.

AlanCoding: No... on the serializer that can't make any sense. It only runs locally.

Makefile
```diff
@@ -173,7 +173,15 @@ init:
...
-	$(MANAGEMENT_COMMAND) register_queue --queuename=controlplane --instance_percent=100;
+	$(MANAGEMENT_COMMAND) register_queue --queuename=controlplane --instance_percent=100;\
+	if [ ! -f /etc/receptor/certs/awx.key ]; then \
+		rm -f /etc/receptor/certs/*; \
+		receptor --cert-init commonname="AWX Test CA" bits=2048 outcert=/etc/receptor/certs/ca.crt outkey=/etc/receptor/certs/ca.key; \
+		for node in $(RECEPTOR_MUTUAL_TLS); do \
+			receptor --cert-makereq bits=2048 commonname="$$node test cert" dnsname=$$node nodeid=$$node outreq=/etc/receptor/certs/$$node.csr outkey=/etc/receptor/certs/$$node.key; \
+			receptor --cert-signreq req=/etc/receptor/certs/$$node.csr cacert=/etc/receptor/certs/ca.crt cakey=/etc/receptor/certs/ca.key outcert=/etc/receptor/certs/$$node.crt verify=yes; \
+		done; \
+	fi; \
```
shanemcd:
```suggestion
	fi
```

beeankha (approved):

awx/main/scheduler/task_manager.py
```diff
@@ -75,6 +75,8 @@ def after_lock_init(self):
         instances_partial = [
             SimpleNamespace(
                 obj=instance,
+                version=instance.version,
```
AlanCoding: We should remove this `version` field, assuming we're not using it somewhere.

shanemcd (approved):

PR #10738, opened 2021-07-23
Fix docker-clean-volumes target when not using minikube

shanemcd:
Without this patch, you get:

```
tools/docker-compose-minikube/_sources/minikube delete
make: tools/docker-compose-minikube/_sources/minikube: No such file or directory
```


tchellomello (approved):
Thanks @shanemcd

rooftopcellist (approved):

Issue #10745, opened 2021-07-23
[Docs] Delete All the Unnecessary Info in awx/docs/clustering.md and awx/docs/tasks.md


beeankha:
(related to https://github.com/ansible/awx/issues/10695)

- [x] In `tasks.md`, we mention "Reaping Receptor Work Units" but don't explain what Receptor work units are; we should explain this concept before explaining how to reap them
- [x] `clustering.md` needs to be thoroughly updated/have passages be purged
- [x] "Provisioning and Deprovisioning Instances and Groups" section in `clustering.md` needs to be re-written with help from the Mesh Installer team
- [x] Remove `receptor_mesh.md` file

Issue #10695, opened 2021-07-20
[Docs] Write developer documentation (receptor conf, auto-provisioning instructions)


jladdjr:

AlanCoding:
@tvo318 Heads up, we merged this, it will be kind of a living document through this release:

https://github.com/ansible/awx/blob/mesh_code/docs/receptor_mesh.md

Please feel free to read it, and we can confer, and change anything if needed.

PR #10747, opened 2021-07-23
Add developer docs for incoming receptor mesh features

AlanCoding:


AlanCoding:
The issue https://github.com/ansible/awx/issues/10745 has been filed to remove or re-org old and conflicting information.

docs/receptor_mesh.md
```diff
@@ -0,0 +1,67 @@
...
+After the initial receptor integration, but before the control plane and execution plane split, all job types ran as the "local" work type on the same node as the control node.
+
+If ran in a container_group, user-space jobs run as the "kubernetes-runtime-auth" work type.
+
+### Auto-discovery of execution nodes
+
+Instances in control plane must be registered by the installer via `awx-manage`
+commands like `awx-manage register_queue` or `awx-manage register_instance`.
+
+Execution-only nodes are automatically discovered when they appear in the receptor mesh.
```
kdelee: They need to declare a control plane node as a peer for this to work, right? or is it any other node that is eventually connected to the control plane?

AlanCoding: Yes, we should clarify that this is all after the receptor setup.

docs/receptor_mesh.md
```diff
@@ -0,0 +1,68 @@
...
+> NOTE: user-space jobs are what carry out the user's Ansible automation. These job types run inside of the designated execution environment so that the needed content is available.
+
+### Split of Control Plane versus Execution Plane
+
+Instances in the control plane run persistent AWX services (like the web server, task dispatcher, etc.), project updates, and management jobs.
+
+Hybrid instances in the control plane may also run user-space jobs.
+
+The unified jobs API reports `controller_node` and `execution_node` fields.
+The execution node is where the job runs. The controller node prepares the private_data_dir for the job, and processes the receptor output stream.
```
jbradberry: `private_data_dir` should probably get back-ticked

docs/receptor_mesh.md
```diff
@@ -0,0 +1,68 @@
...
+```
+- work-command:
+    worktype: ansible-runner
+    command: ansible-runner
+    params: worker
+    allowruntimeparams: true
+```
+
+Control (and hybrid) nodes need to advertise the "local" work type.
+Project updates are submitted as this work type.
```
jbradberry, line 53: Ah, thanks for this clarification.  :+1:

docs/receptor_mesh.md
```diff
@@ -0,0 +1,68 @@
...
+
+```
+- work-command:
+    worktype: local
+    command: ansible-runner
+    params: worker
+    allowruntimeparams: true
+```
+
+If user-space jobs run on a hybrid node, they will also run as the "local" work type.
```
jbradberry, line 54: Wait, then how is a hybrid node different from a control-only node?  Purely the database record?  That seems off to me.

docs/receptor_mesh.md
```diff
@@ -0,0 +1,68 @@
...
+
+After the initial receptor integration, but before the control plane and execution plane split, all job types ran as the "local" work type on the same node as the control node.
+
+If ran in a container_group, user-space jobs run as the "kubernetes-runtime-auth" work type.
```
jbradberry: Is the underscore intentional here?  Is this a convention for a group name (and if so, Ansible group or InstanceGroup)?

docs/receptor_mesh.md
```diff
@@ -0,0 +1,68 @@
+## Receptor Mesh
+
+AWX uses a Receptor mesh to transmit "user-space" unified jobs:
+ - jobs
+ - ad hoc commands
+ - inventory updates
+
+to the node where they run.
+
+https://github.com/ansible/receptor
```
beeankha: I think it would be best to remove this standalone link from here and convert the word `Receptor` in that very first sentence to `[Receptor](https://github.com/ansible/receptor)`

docs/receptor_mesh.md
AlanCoding: control-only nodes will not be sent user-space jobs by the task manager. The receptor configuration, however, will be indistinguishable between the two types in this proposal.

I will try to find some place to use more specific wording to distinguish the two types.

docs/receptor_mesh.md
AlanCoding: It's an instance group

https://docs.ansible.com/ansible-tower/3.8.3/html/installandreference/glossary.html#term-Container-Groups

these docs seems to use lowercase in some cases. I find this grammatically hard to parse, but we do that in server logs, so that's the pattern I should continue. Will change.

docs/receptor_mesh.md
```diff
@@ -0,0 +1,66 @@
...
+
+> NOTE: user-space jobs are what carry out the user's Ansible automation. These job types run inside of the designated execution environment so that the needed content is available.
+
+> NOTE: The word "node" corresponds to entries in the `Instance` database model, or the `/api/v2/instances/` endpoint, and is a machine participating in the cluster / mesh.
+
+### Split of Control Plane versus Execution Plane
+
+Instances in the control plane run persistent AWX services (like the web server, task dispatcher, etc.), project updates, and management jobs.
+
+Hybrid instances in the control plane may also run user-space jobs.
```
rebeccahhh: I think maybe move this below the next paragraph so the order is
- Control
- Execution
- Both (aka hybrid)

docs/receptor_mesh.md
```diff
@@ -0,0 +1,66 @@
...
+    params: worker
+    allowruntimeparams: true
+```
+
+Control (and hybrid) nodes advertise the "local" work type instead.
+So the entry is the same as above, except that it has `worktype: local`.
+Project updates are submitted as this work type.
+If user-space jobs run on a hybrid node, they will also run as the "local" work type.
+
+After the initial receptor integration, but before the control plane and execution plane split, all job types ran as the "local" work type on the same node as the control node.
```
rebeccahhh: ran -> run

beeankha (approved):

docs/receptor_mesh.md
AlanCoding: This paragraph is a comment about historical behavior. Honestly that's too confusing to keep up with, because we avoid referencing version numbers. It's just too hard to context switch from description of behavior to past behavior. I was already thinking about deleting this paragraph, and I think I'm just going to do that.

docs/receptor_mesh.md
jbradberry: Ok, good to know.

docs/receptor_mesh.md
AlanCoding: I took your advice to heart, but it kind of ballooned into a larger rewrite. New commit incoming.

beeankha:
Connecting issue https://github.com/ansible/awx/issues/10695

PR #10757, opened 2021-07-26
add a node_type field to the instance model

rebeccahhh:
##### SUMMARY

add a field to the instance model for the type of node the instance is, control, execution, or hybrid

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.2.2
```


awx/main/migrations/0152_instance_node_type.py
```diff
@@ -0,0 +1,22 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('main', '0151_rename_managed_by_tower'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='instance',
+            name='node_type',
```
chrismeyersfsu, line 15: thank you for not just calling this `type`

awx/main/migrations/0152_instance_node_type.py
```diff
@@ -0,0 +1,22 @@
...
+            field=models.CharField(
+                choices=[('control', 'Control plane node'), ('execution', 'Execution plane node'), ('hybrid', 'Controller and execution')],
```
chrismeyersfsu, line 17: hybrid, it's now a thing. Let it be known!

rebeccahhh: it's better on the gas mileage.

awx/api/serializers.py
```diff
@@ -4799,6 +4799,7 @@ class Meta:
             "mem_capacity",
             "enabled",
             "managed_by_policy",
+            "node_type",
```
chrismeyersfsu, line 4802: Should be read-only.

AlanCoding:
Could we go ahead and add this to `awx-manage provision_instance`?

I'd like to minimize surprises as much as possible, so `--node-type` with a default of `hybrid` would be great with me. Choices could be `control` and `hybrid`. I would _not_ include `execution`, because those should be auto-discovered, not registered. Unless anyone has any disagreements.

If we had that, we could go ahead and start using it with `make docker-compose`

awx/api/serializers.py
jbradberry: If so, how can we convert a hybrid node that was created in 4.0 into a control-only node?

AlanCoding: what if we set editable=False on the field, and then `awx-manage provision_instance` could still change it to reflect value in the inventory file?

rebeccahhh:
something like that most recent commit @AlanCoding ?

AlanCoding (approved):

awx/main/management/commands/provision_instance.py
```diff
@@ -18,11 +18,12 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument('--hostname', dest='hostname', type=str, help='Hostname used during provisioning')
+        parser.add_argument('--node_type', type=str, default="hybrid", choices=["control", "execution", "hybrid"], help='Instance Node type')
```
AlanCoding, line 21: @chrismeyersfsu I guess maybe execution nodes have to be registered by the installer to be added to a group?

In that case, our heartbeat tasks won't be adding the instance, they will be marking its capacity as non-zero, as if it had went dark previously.

awx/api/serializers.py
jbradberry: We made the provision_instance command be able to change the type.  Resolved.

AlanCoding:
@wenottingham @tvo318 Do you have any objections to the `node_type` phrasing?

When writing the developer docs page for the [receptor mesh code changes](https://github.com/ansible/awx/blob/mesh_code/docs/receptor_mesh.md), I realized it was difficult how I always used the word "node" in my writing, but had to switch to "instance" when touching back to the model presented by AWX.

The most obvious alternative would be `instance_type`, but perhaps there are other options.

I'm asking now, as changing it will become more difficult after it merges, and become increasingly difficult thereafter as other developers adopt this.

wenottingham:
If they are under /api/v2/instances, and grouped into instance_groups, then maybe `instance_type` is better for consistency.

AlanCoding:
A counter-argument is that we have the `controller_node` and `execution_node` fields for where the job runs.

wenottingham:
True. I leave it up to the team's judgement, I don't have strong feelings, and can see both sides.

jbradberry:
I'm not a big fan of `instance_type`.

shanemcd (approved):

beeankha (approved):

rebeccahhh:
I prefer node_type specifically because it connects to `execution node`, `hybrid node`, and `control node` which we do use elsewhere.
While the term switching is not ideal I agree, the one that makes the most sense to me given our current naming convention is `node_type` (which may or may not be a given, since I wrote this).
I completely agree that if we want to establish a better term we need to do it now, but I can't think of anything better than `node_type`

AlanCoding (approved):

Issue #10755, opened 2021-07-26
Make awx_x nodes fully connected in docker-compose environment


AlanCoding:

AlanCoding:
Pasting in notes I had on this:

 - delete `tools/docker-compose/awx-1-receptor-standalone.conf`
 - move `tools/docker-compose/receptor.conf` into the docker-compose templates, alongside the other 2 receptor config files
 - in the template for `awx_i`, add peers for `i+1` to `N`, or, add peer from `1` to `i-1`

AlanCoding:
Demo command for later reference

```
CONTROL_PLANE_NODE_COUNT=2 EXECUTION_NODE_COUNT=3 COMPOSE_TAG=devel make docker-compose
```

PR #10758, opened 2021-07-26
Make the AWX nodes fully connected in the development environment

AlanCoding:
Connect https://github.com/ansible/awx/issues/10755

This is making the development environment more like what the final production environment is now expected to be like.

All control nodes will be connected, but only connected in 1 direction. That means that, for node `i` out of `N` nodes, it will connect to nodes from `i+1`, to `N`.

WIP - still testing.


AlanCoding:
This does seem to be able to run jobs. I might ask some other people to test, but otherwise I'm pretty happy with this.

fosterseth (approved):

beeankha (approved):
I tested this and it ran a job successfully :+1:

Issue #10594, opened 2021-07-06
Dependency upgrade Meta Issue


AlexSCorey:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

We should upgrade dependencies.  Below is a list of dependencies that are at least 1 major release behind.  

- [ ] eslint-config-airbnb
- [ ] eslint-plugin-react-hooks
- [ ] has-ansi
- [ ] html entities
- [ ] http-proxy-middleware
- [ ] js-yaml
- [ ] prettier
- [ ] react
- [ ] react-dom
- [ ] styled-components
- [ ] react-router (this will be much easier/faster once the backwards-compat package is available: https://reactrouterdotcom.fly.dev/docs/en/v6/upgrading/v5#backwards-compatibility-package)

Mostly this issue is to discuss which, if any, of these we want to upgrade now, and which can wait.  We should create other issues to track upgrading of individual/groups of dependencies and link those issues to this 1.  

### AWX version

19.2

### Installation method

docker for mac

### Modifications

no

### Steps to reproduce

run `npm outdated`

### Expected results

some deps will be outdated

### Actual results

see screen shot below

### Additional information

![Screen Shot 2021-07-06 at 2 14 28 PM](https://user-images.githubusercontent.com/39280967/124647893-89819980-de64-11eb-82f0-9b8b59cce168.png)

AlexSCorey:
We have made really good progress in updating many dependencies that were behind. Some of them were behind by several major releases.  Now, all dependencies have been updated to the latest or are no more than 1 major release behind.  We looked into upgrading React, but with React 18 coming out soon we decided to upgrade react after it is released.

![Screen Shot 2021-08-06 at 1 10 54 PM](https://user-images.githubusercontent.com/39280967/128547522-aa92f339-daa2-4336-adb1-c24e672ae7e6.png)

AlexSCorey:
I've reassigned this to myself to see about updating React to v17 as discussed yesterday during dev sync.

AlexSCorey:
related https://github.com/ansible/awx/issues/11105

AlexSCorey:
[...]

PR #10772, opened 2021-07-28
Upgrade formik

nixocio:
Upgrade formik

See: https://github.com/ansible/awx/issues/10594


keithjgrant (approved):

nixocio:
I used the steps described here  https://github.com/ansible/awx/blob/devel/awx/ui_next/README.md#updating-dependencies to upgrade this dependency.  Even doing that the `awx/ui_next/package-lock.json` was massively altered.

PR #10779, opened 2021-07-28
[WIP] [Automated-Installer] first iteration of ansible plugin for generating and validating user-defined mesh topology

sarabrajsingh:
Upstream issue: https://github.com/ansible/tower-packaging/issues/1381

# TODO
- [ ] Nodes in `[automationcontroller]` automatically get connected with a one-way link between each pair
- [x] Nodes in `[automationcontroller]` are default of type `'hybrid'` and nodes in `[execution_nodes]` are default of type `'execution'` https://github.com/sarabrajsingh/awx/pull/1
- [x] The `peers` hostvar gets resolved into a list of hosts in the output, groups fully expanded, and comma-separated strings vs yaml lists dealt with
- [x] Output the structure to be used in the `jinja2` templates. (list of dicts  - one per host node) `[{'name': 'p70', 'peers': [...], 'node_type': 'hybrid', ...}, {...}, ...]`
- [ ] With an optional flag, the module dumps out a dot graph format file
- [ ] Preflight checks
  - [x] Ensure hosts cannot be members of `[automationcontroller]` and `[execution_nodes]` at the same time
  - [x] Ensure all hosts from `[automationcontroller]` and `[execution_nodes]` have a valid `node_type`
  - [ ] Whenever a node has the `deprovision` action, must have at least 2 hosts in the `[automationcontroller]` group 
- [ ] Create a skeleton module that contains documentation and usage examples

### DIFF
--- a/awx_collection/plugins/action/generate_mesh.py
+++ b/awx_collection/plugins/action/generate_mesh.py
@@ -90,7 +90,7 @@ class ActionModule(ActionBase):
 
         return None
 
-    def preflight_verify_node_type(self, task_vars=None, group_name=None, valid_types=None,):
+    def assert_verify_node_type(self, task_vars=None, group_name=None, valid_types=None,):
         """
         Members of given group_name must have a valid node_type.
         """
@@ -104,7 +104,7 @@ class ActionModule(ActionBase):
         return
 
 
-    def preflight_unique_group(self, task_vars=None):
+    def assert_unique_group(self, task_vars=None):
         """
         A given host cannot be part of the automationcontroller and execution_nodes group.
         """
@@ -125,10 +125,9 @@ class ActionModule(ActionBase):
         if task_vars is None:
             task_vars = dict()
 
-        # Series of preflight that must pass
-        self.preflight_unique_group(task_vars)
-        self.preflight_verify_node_type(task_vars, group_name='automationcontroller', valid_types=CONTROLLER_NODE_TYPES)
-        self.preflight_verify_node_type(task_vars, group_name='execution_nodes', valid_types=EXECUTION_NODE_TYPES)
+        self.assert_unique_group(task_vars)
+        self.assert_verify_node_type(task_vars, group_name='automationcontroller', valid_types=CONTROLLER_NODE_TYPES)
+        self.assert_verify_node_type(task_vars, group_name='execution_nodes', valid_types=EXECUTION_NODE_TYPES)
 
         result = super(ActionModule, self).run(tmp, task_vars)
         # module_args = self._task.args.copy()
--- a/awx_collection/plugins/action/generate_mesh.py
+++ b/awx_collection/plugins/action/generate_mesh.py
@@ -7,6 +7,8 @@ __metaclass__ = type
 from ansible.errors import AnsibleError
 from ansible.plugins.action import ActionBase
 from datetime import datetime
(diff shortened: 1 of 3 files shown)


awx_collection/plugins/action/generate_mesh.py
```diff
@@ -0,0 +1,121 @@
...
+# Make coding more python3-ish, this is required for contributions to Ansible
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+from ansible.plugins.action import ActionBase
+from datetime import datetime
+
+
+class ActionModule(ActionBase):
```
chrismeyersfsu: Ansible action module  :muscle:

awx_collection/plugins/action/generate_mesh.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def assert_unique_group(self, task_vars=None):
+        """
+        A given host cannot be part of the automationcontroller and execution_nodes group.
+        """
+        automation_group = task_vars.get('groups').get('automationcontroller')
+        execution_nodes = task_vars.get('groups').get('execution_nodes')
+
+        if automation_group and execution_nodes:
+            intersection = list(set(automation_group) & set(execution_nodes))
```
jbradberry: This doesn't need to be cast to a `list`

awx_collection/plugins/action/generate_mesh.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        Members of given group_name must have a valid node_type.
+        """
+        if 'node_type' not in vars.keys():
+            return valid_types[group_name]['default_type']
+
+        if vars['node_type'] not in valid_types[group_name]['types']:
+            raise AnsibleError(
+                'The host %s must have one of the valid node_types: %s' %
+                (host, ', '.join(str(_) for _ in list(valid_types[group_name]['types'])))
```
jbradberry: By convention, `_` as a loop variable means that you are throwing it away, which we aren't doing here.

awx_collection/plugins/action/generate_mesh.py
```diff
@@ -0,0 +1,157 @@
...
+            return
+        for key, value in local_rhs.items():
+            if key in local_lhs:
+                local_lhs[key] = local_lhs[key] + value
+            else:
+                local_lhs[key] = value
+
+        return local_lhs
+
+    def generate_dot_syntax_from_dict(dict=None):
```
jbradberry: Let's not override the `dict` builtin accidentally.

awx_collection/plugins/action/generate_mesh.py
```diff
@@ -0,0 +1,157 @@
...
+                (host, ', '.join(str(_) for _ in list(valid_types[group_name]['types'])))
+            )
+        return vars['node_type']
+
+
+    def assert_unique_group(self, task_vars=None):
+        """
+        A given host cannot be part of the automationcontroller and execution_nodes group.
+        """
+        automation_group = task_vars.get('groups').get('automationcontroller')
```
jbradberry: Defensively using `.get()` is a good idea, but chaining like this could wind up with `None` being the result of the first call and thus an error for the second.  We probably want `.get('groups', {}).get(...)`

jbradberry:
Closing, since this work was ported to a different collection.

Issue #10587, opened 2021-07-06
Scheduled Managed Job Time not saved properly


AlexSCorey:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

Edit the time of day for a schedule on a managed job.  The selected time is not properly reflected in the details view.

### AWX version

devel

### Installation method

docker for mac

### Modifications

no

### Steps to reproduce

Click on Managed Jobs.  Click on schedule.  Edit the schedule.  Change the start date.  Save the schedule. 

### Expected results

The time of day is properly reflected in the details view

### Actual results

The selected time of day is not properly reflected in the details view

### Additional information

![schedule](https://user-images.githubusercontent.com/39280967/124621117-02bec380-de48-11eb-86d7-4988f7870a95.gif)

fosterseth:
your selection is in UTC, but the details page will convert the time to your local timezone (note, not the Local Time Zone field on the edit page, but your actual local time)

shanemcd:
@wenottingham thinks something is up here. Worth investigating.

kialam:
Changing the timezone dropdown to anything but `UTC` and changing the start date also causes some weird behavior with displaying the date as well:

![date-timezone](https://user-images.githubusercontent.com/2293210/125652895-53d8a126-7a0f-439a-b4ad-173691329ec5.gif)

Not sure if this is expected behavior, but it looks like we're showing the next scheduled run for the managed job (e.g. 8/2) instead of the newly patched start date (e.g 7/26).  This form field directly correlates to the `dtstart` value returned from the API.

kialam:
@shanemcd @fosterseth is this intended behavior that the `dtstart` field returns the _next_ scheduled job run? Currently in the UI a user can select a start date that doesn't correspond to the "on days" selection, resulting in some discrepancy. For example, if you select a date that doesn't land on "Monday", a schedule that runs ever Monday will then return a `dtstart` date of the following Monday.

kialam:
Note that toggling the end date in the UI does not create this issue mentioned above, because we use the `until` param which correctly logs the user selected end date/time. Should we have another field the API returns called `begin` or something similar that logs the user selected start date/time?

[...]

PR #10783, opened 2021-07-29
Fix 10587 managed job time

kialam:
##### SUMMARY

This PR modifies the Schedules section (list, edit, detail, etc) to use a `luxon.js` to support displaying timezones.

Previously, schedules shown in the Management Jobs detail and edit page were converted to local time instead of the timezone of the selected schedule.

Related #10587 
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - UI

##### AWX VERSION

```
19.2.2
```

### DIFF
--- a/awx/ui/package-lock.json
+++ b/awx/ui/package-lock.json
@@ -21,6 +21,7 @@
         "has-ansi": "4.0.0",
         "html-entities": "2.3.2",
         "js-yaml": "^3.13.1",
+        "luxon": "^2.0.1",
         "prop-types": "^15.6.2",
         "react": "^16.13.1",
         "react-ace": "^9.3.0",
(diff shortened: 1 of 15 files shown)


keithjgrant (approved):
Code looks good... though the package-lock file shows a huge number of deletions, which seems odd since we're adding a dependency. Might be worth rebasing and then installing luxon afterwards?

nixocio:
We have been adding/updating dependencies from inside the container to make `package-lock.json` consistent with different environments. See: https://github.com/ansible/awx/blob/devel/awx/ui_next/README.md#updating-dependencies

nixocio (approved):
Good job @kialam! 👍

akus062381:
Adding the needs_test label - UI QE should have a look at this.

tiagodread (approved):
This looks good, we can merge after fixing the conflict

PR #10797, opened 2021-07-30
Update workflow to use pull_request_target

tiagodread:
This update the github workflow to run on `pull_request_target` instead of `pull_request`


nixocio (approved):

nixocio:
See: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target

shanemcd (approved):

Issue #9222, opened 2021-02-03
Passing a bad lookup type modifier in api request triggers 500 error


mabashian:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Send a request like `/api/v2/jobs/?id__iexact=1`

<img width="1680" alt="Screen Shot 2021-02-03 at 12 28 26 PM" src="https://user-images.githubusercontent.com/9889020/106785392-56b84880-661b-11eb-9288-fa50845636f7.png">

`iexact` isn't supported on the `id` field.  What I'm hoping for is a nice 4xx error that says that modifier is not supported.

The impetuous for this issue came from https://github.com/ansible/awx/issues/7834.  The advanced search in the UI allows the user to select from a static list of modifiers in the `Advanced` section.  I left that issue open as a UI enhancement to attempt to trim that list down to only modifiers that will work with a particular field.

AlanCoding:
@sarabrajsingh are you still working on this right now? If now I'd like to keep it in the On Deck issues for anyone to pick up as able.

rebeccahhh:
There is a theory that this may change with the in progress Django upgrade. 
@jbradberry I believe recommended this wait until that's complete to recreate/resolve.

AlanCoding:
Marked as blocked

AlanCoding:
Confirmed that `/api/v2/jobs/?id__iexact=1` gives a 500 with current devel. I believe this is ready to be worked on.

jay-steurer:
PR merged. Closing it.

Issue #7834, opened 2020-08-05
Advanced search: certain lookups (suffixes) invalid for certain types of keys.


jlmitch5:
![89440610-de2f0880-d719-11ea-8b09-d29297428d2e](https://user-images.githubusercontent.com/1342624/89447526-16d3df80-d724-11ea-880e-e8bbaffeed39.png)

In the above example, iexact is invalid for the "id" key, because id expects a number.  This enhancement would filter the list of lookups based on the key selected's type.

my guess is that it's doable...we could probably filter that list based on what "type" the key they select is.  but that's a pretty sizeable enhancement (we need to get that from the options request, pass it to advanced search, check to see what type of key it is upon selection and then filter the possible options list)

I feel that we should triage this ask to see if we really want to add this for 4.0

mabashian:
The enhancement that @jlmitch5 outlines is definitely worthwhile so I think we should keep this issue open.

The fact that we can trigger a 500 by doing this is not great so I'm going to open up an api issue to see if they can return a 4xx error instead to give the user a little better message.

mabashian:
Link https://github.com/ansible/awx/issues/9222

unlikelyzero:
still accurage

nixocio:
This one seems to be related to https://github.com/ansible/awx/issues/9956

PR #10798, opened 2021-07-30
Only allow legal/logical match types in advanced search

keithjgrant:
##### SUMMARY
Limits the match type options in Advanced Search based on the selected search key:
- iexact, contains, and icontains matching only allowed on string fields
- startswith and endswith options removed when searching on datetime fields
- greater than/less then options removed when searching a json field

As far as I can tell, invalid `icontains` searches were the only ones that actually resulted in errors from the API, but since some of these other methods didn't really seem to make sense to me, I removed options that seemed nonsensical or not useful (I'm open to suggestions on any changes here)

Addresses #7834, #9222

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI
 
##### ADDITIONAL INFORMATION
This change involved restructuring the `searchableKeys` prop from an array of strings to an array of objects, each containing the search key and its data type. I added a util function to make it easier to extract this from the ACTIONS response and updated everywhere that did this work to use the new util


nixocio:
hello @keithjgrant, goob job on this PR.

~I spent some time with this code, and I noticed this one.~

image

~Is that a case that we want to avoid with your changes?~

I was not able to reproduce this behavior again.

keithjgrant:
@nixocio good catch. I'll take a look

nixocio (approved):

akus062381:
I am adding the needs_test label because I feel UI-QE should verify these changes.

akus062381:
@keithjgrant question- I noticed that the "elapsed" field had lookup type options that the date/time fields did not. I know that elasped is only time, but I wanted to make sure this was your intention?

keithjgrant:
@akus062381 Yes, that was my intention. I make the decision purely based on the `type` the API expects for a field. Since "elapsed" is type "decimal" some of the filters that don't make sense for dates might still make sense for filtering on elapsed # of seconds

akus062381 (approved):

Issue #10364, opened 2021-06-04
Allow job_explanation to be set for jobs that hit an exception


rooftopcellist:
##### ISSUE TYPE
 - Bug Report

#### SUMMARY

Inspired by conversation on this ticket: https://github.com/ansible/awx/issues/10114#issuecomment-853953369

Currently, if a job_explanation is set on a job, if that job hits an exception/error, the job_explanation gets cleared.  This is not the case for canceled jobs.  

###### Use case

For container groups, when a job is run, if invalid k8s/openshift credentials are provided or the service account doesn't have sufficient permissions, the job will fail with a raised exception.  If we could set the `job_explanation` and have it persist, a useful message would show in the UI next to the status of the job, in place of the "Error" in the screenshot below. 

![image](https://user-images.githubusercontent.com/11698892/120867096-39f73780-c55f-11eb-9981-9a1858885fbf.png)

More background on that in [this comment](https://github.com/ansible/awx/pull/10204#issuecomment-848244192).  

#### ENVIRONMENT
* AWX version: 19.0.0
* Feature: Container Groups & Error Handling for Jobs

#### STEPS TO REPRODUCE

1. Create OCP cluster cred with a service account with no rules for secrets
2. Create an awx-ee and push it to a protected registry
3. Create a registry cred in AWX
4. Create an EE in AWX pointing to your custom ee in the protected registry, add your registry cred to it
5. Create a Container Group with your OCP cluster cred
6. Create a JT, configure the Container Group and EE.  
7. Run a job and see that is fails with no job_explanation set

#### EXPECTED RESULTS

When you add:
```
job.job_explanation(error_msg)
job.save()
```

In the code here: https://github.com/ansible/awx/pull/10204/files#diff-e47089fc124c2bf4fc4bf47dde61fe91888e31331c3e067e27652858f6ae0fc6R125

the job_explanation should be set on the job, no matter what happens next.  
#### ACTUAL RESULTS

The job_explanation gets cleared when the job hits an error in execution and the user is left without a descriptive error (though they do have a stacktrace in the stdout of the job).

jeffkala:
[...]

PR #10805, opened 2021-08-03
Filling job_explanation field if an exception occurred during the job execution

x0r3ax3ax:
##### SUMMARY

The job_explanation field is filled if an exception occurred during the job execution

#10364 

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
19.2.2
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/scheduler/kubernetes.py
+++ b/awx/main/scheduler/kubernetes.py
@@ -7,7 +7,7 @@ from urllib import parse as urlparse
 from django.conf import settings
 from kubernetes import client, config
 from django.utils.functional import cached_property
-from django.utils.translation import ugettext_lazy as _
+from django.utils.translation import ugettext_lazy as _, gettext_noop
 
 from awx.main.utils.common import parse_yaml_or_json
 from awx.main.utils.execution_environments import get_default_pod_spec
@@ -102,7 +102,7 @@ class PodManager(object):
                     )
                 full_error_msg = '{0}: {1}'.format(error_msg, str(e))
                 logger.exception(full_error_msg)
-                raise PermissionError(full_error_msg)
+                raise PermissionError(gettext_noop(error_msg))
 
         if replace_secret:
             try:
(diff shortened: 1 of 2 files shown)


AlanCoding:
The outer context already has exception handling

https://github.com/ansible/awx/blob/9af9e9a4e4b05531ff076733a79883630ba46db3/awx/main/tasks.py#L1422-L1427

I don't want this block to be re-implemented for the receptor job class. The `result_traceback` is also the most appropriate place for the data to go. If you thought that `job_explanation` should also be filled in for traceback cases (from `str(e)`), we could manage that.

It doesn't look like this would help with issues like https://github.com/ansible/awx/issues/10213 either.

x0r3ax3ax:
> The outer context already has exception handling
> 
> https://github.com/ansible/awx/blob/9af9e9a4e4b05531ff076733a79883630ba46db3/awx/main/tasks.py#L1422-L1427
> 
> I don't want this block to be re-implemented for the receptor job class. The `result_traceback` is also the most appropriate place for the data to go. If you thought that `job_explanation` should also be filled in for traceback cases (from `str(e)`), we could manage that.
> 
> It doesn't look like this would help with issues like #10213 either.

Hi!
Yes you are right. The current solution is bad. I will move the filling of job_explanation to the part you provided.

As for issue #10213, I was unable to reproduce that problem. In my case i get a traceback like this
```python
  File "/awx_devel/awx/main/tasks.py", line 1406, in run
    res = receptor_job.run()
  File "/awx_devel/awx/main/tasks.py", line 2936, in run
    return self._run_internal(receptor_ctl)
  File "/awx_devel/awx/main/tasks.py", line 3005, in _run_internal
    raise RuntimeError(detail)
RuntimeError: Stream error running pod: context canceled
```
Please, could you send me an ansible-playbook with the task you are running on the mail specified in my profile?

AlanCoding:
I'll leave a note that I need to re-test that issue, thanks for the info.

AlanCoding:
With the recent push, I don't think it would expose any new information, and the UI will get mixed up about which to show between the two fields.

PR #10810, opened 2021-08-03
Bump npm version inside container

nixocio:
Bump npm version inside container

See: https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json#lockfileversion

With this change
```
base) ➜  ~ docker exec -it tools_awx_1 bash
bash-4.4$ npm -v
7.20.3
bash-4.4$ node -v
v14.15.1
bash-4.4$```

### DIFF
--- a/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
+++ b/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
@@ -141,7 +141,7 @@ RUN dnf -y install \
     wget \
     diffutils \
     unzip && \
-    npm install -g n && n 14.15.1 && dnf remove -y nodejs
+    npm install -g n && n 14.15.1 && npm install -g npm@7.20.3 && dnf remove -y nodejs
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tools/ansible/roles/dockerfile/templates/Dockerfile.j2 [hunk 1]
```diff
@@ -141,7 +141,7 @@ RUN dnf -y install \
     wget \
     diffutils \
     unzip && \
-    npm install -g n && n 14.15.1 && dnf remove -y nodejs
+    npm install -g n && n 14.15.1 && npm install -g npm@latest && dnf remove -y nodejs
```
jakemcdermott: @nixocio I think we should pin to a version, if possible.

nixocio: @jakemcdermott, suggestion accepted, tested and updated.

AlexSCorey (approved):
this looks good to me.

Issue #8474, opened 2020-10-28
Clear All Filters in Search should disappear


AlexSCorey:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
After searching it is expected that Clear all filters should be removed after clearing all of the search chips.  

##### ENVIRONMENT
* AWX version: 14.1
* AWX install method:  docker for mac
* Ansible version:  2.9.13
* Operating System: Mojave
* Web Browser: Chrome

##### STEPS TO REPRODUCE
1) Go to a list
2) Search for something on that list that generates 1 chip.
3) Notice that a chip is generated under the search input
4) Click on that chip (all chips) to remove it.

##### EXPECTED RESULTS
the chip is removed and `Clear all filters` is also removed
##### ACTUAL RESULTS

The chip is removed but `Clear all filters` is not removed. 

##### ADDITIONAL INFORMATION
![weird filter](https://user-images.githubusercontent.com/39280967/97449126-bc53b500-1907-11eb-97aa-59aa61beaa04.gif)

unlikelyzero:
This is a dupe of https://github.com/ansible/awx/issues/7830

akus062381:
I am still seeing this happening

Steps to take:
1. run a playbook job
2. go to output page
3. select advanced search and search for a valid host for that job by name
4. once results are shown, click Clear All Filters

You will notice that the filters are cleared, but the Clear All Filters link remains

PR #10819, opened 2021-08-05
Remove Clear All button in search when filters cleared

keithjgrant:
##### SUMMARY
Fixes "Clear All Filters" button from lingering around after clearing filters.

Addresses #8474

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
This was caused because the labels on our search filters changed depending whether or not the filter was present in the URL params. This object was built by a helper function in the component; I moved the helper to its own file and wrapped some tests around it, then ensured the filter label was the same at all times.


AlexSCorey (approved):
I tested this locally.  It looks good to me.  Well done!

Issue #10740, opened 2021-07-23
Add expand all for hosts lists


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Add expand all for hosts lists.

<img width="1484" alt="image" src="https://user-images.githubusercontent.com/9053044/126792020-4b0ad09c-1284-49c4-9f68-773a51d569b4.png">

See: https://github.com/ansible/awx/issues/4208

akus062381:
implemented and works great

Issue #10739, opened 2021-07-23
Convert Survey List to Tables


AlexSCorey:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
This table should not be sortable since the order of the survey questions matter.  The user can manage the question ordering by using the up and down button on the left side of the screen.  Also need to remove the inline labels (ie. Type)

Also consider doing this work too when doing this https://github.com/ansible/awx/issues/10654

akus062381:
implemented and looks great

Issue #10654, opened 2021-07-15
Update SurveyList to use draggable data list as per mock ups


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Mockup:

<img width="1502" alt="image" src="https://user-images.githubusercontent.com/9053044/125801248-3285764c-41f0-4bf1-89c6-de0e1881addc.png">

This patternfly component was added after the first implementation of this screen.

https://www.patternfly.org/v4/components/data-list/#draggable

Current implementation:

<img width="1491" alt="image" src="https://user-images.githubusercontent.com/9053044/125801431-7e985daf-92e6-43ac-bdd7-697d7ed77abc.png">

AlexSCorey:
After conversation with @trahman73 we think it makes the most sense to re-utilize the preview modal.  Currently that modal shows that same data as the list itself.  
With the new functionality we could rename the preview button to be `Edit Order` and the user would click on that to be able to reorder the survey questions using the drag and drop functionality.  This looks better, improves UX, and it means the user no longer has to be in `isLoading` each time they move a survey question up or down 1 row.  They can move all of the questions as many rows as they want, save the order, and see it reflected in the new list after they save.

trahman73:
Updated flow:
List view - https://tower-mockups.testing.ansible.com/patternfly/templates/templates-survey/
Edit order modal - https://tower-mockups.testing.ansible.com/patternfly/templates/templates-survey-preview/

akus062381:
implemented and looks great

PR #10820, opened 2021-08-05
Host List Expand All and Survey List -> Tables

AlexSCorey:
##### SUMMARY
This PR addresses #10654 #10739 and #10740.  The biggest change here is the way the use will edit the Survey question order.  We wanted to add drag functionality and @trahman73 decided that the best way to do it was to repurpose the Preview modal(which had the same data on it as the list itself) to be a order edit modal that allows the user to drag their questions where they want.  This also means that the user does not have to wait for the UI to POST the new order and fetch the new order from the API every time they change a single question.  They can do all of there changes at 1 time and then save those changes.
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API
 - UI

##### ADDITIONAL INFORMATION
host expand
SurveyChanges


awx/ui/src/screens/Template/Survey/SurveyList.js
```diff
@@ -129,36 +104,47 @@ function SurveyList({
...
-            question={question}
-            isChecked={selected.some((q) => q.variable === question.variable)}
-            onSelect={() => handleSelect(question)}
-            onMoveUp={moveUp}
-            onMoveDown={moveDown}
-            canEdit={canEdit}
-          />
-        ))}
+      <>
+        <TableComposable>
```
nixocio: We could add datalabel, ids, or ouiaId to help identifying all those table elements. Ease testing.

nixocio (changes requested):
This is a nice UX improvement.

A few suggestions.

1. Add a note about drag and drop here.
image

Something like:

image

2. Disable Edit Order button if there is just one survey

image

image

3. Add a tooltip to `Edit Order` button.
4. Add a `Answer variable name` `blah` and use the drag and drop - click on the survey, and it will redirect away from the app.

nixocio (approved):
This a great UX improvement. Good job @AlexSCorey.

akus062381:
@AlexSCorey

Would you please add better locators (OUIAID, data-cy, ID) to the expanded list items in the Host table, as shown below?
image

Can you please add better locators for the places I have circled?
image

Thank you!

akus062381 (approved):

nixocio:
Since there were more change since the last time we ran the e2e tests, I will apply this label again.

PR #10823, opened 2021-08-06
Make status command in error handling cleaner

AlanCoding:
Just addressing the TODO in the code here, while I'm learning some more of the receptor interaction patterns.

PR #10825, opened 2021-08-06
Use the ansible-runner worker --worker-info to perform execution node capacity checks

AlanCoding:
What we had before was really more of a proof-of-concept than something polished. This still needs more polishing around the edges, but the overall flow is something I'm vastly more happy with.

This makes use of https://github.com/ansible/ansible-runner/pull/762

Advantages of this approach:

 - No image download to get capacity, so won't take 5 minutes, actually finishes in about 1 second
 - No temporary files
 - No processing of stuff in callback methods
 - More clearly establishing the pattern of interacting with receptor_ctl, which I hope can help influence job error handling too (see https://github.com/ansible/awx/pull/10823, but that's just scratching the surface), and eventually I'd like to move a lot of the functions performed here upstream to receptorctl

I have manually installed the `devel` branch of ansible-runner to node `receptor-1`. With that, a demo here of running this:

```
tools_awx_1     | awx-dispatcher stdout | 2021-08-06 13:36:04,763 WARNING  [1cb3cb74] awx.main.tasks Failed to find capacity of execution node receptor-2, errors:
tools_awx_1     | awx-dispatcher stdout | Old version (2.0.1 or earlier) of ansible-runner on node receptor-2 without --worker-info
tools_awx_1     | awx-dispatcher stdout | 2021-08-06 13:36:04,768 WARNING  [1cb3cb74] awx.main.tasks Failed to find capacity of execution node receptor-3, errors:
tools_awx_1     | awx-dispatcher stdout | Old version (2.0.1 or earlier) of ansible-runner on node receptor-3 without --worker-info
tools_awx_1     | awx-dispatcher stdout | 2021-08-06 13:36:04,785 INFO     [1cb3cb74] awx.main.tasks Set capacity of execution node receptor-1 to 290.00, worker info data:
tools_awx_1     | awx-dispatcher stdout | {
tools_awx_1     | awx-dispatcher stdout |   "Errors": [],
tools_awx_1     | awx-dispatcher stdout |   "transmit_timing": 0.011393308639526367,
tools_awx_1     | awx-dispatcher stdout |   "run_timing": 1.0031061172485352,
tools_awx_1     | awx-dispatcher stdout |   "CPU Capacity": 8,
tools_awx_1     | awx-dispatcher stdout |   "Memory Capacity": 32619828,
tools_awx_1     | awx-dispatcher stdout |   "Version": "2.0.0.0a4.dev90"
tools_awx_1     | awx-dispatcher stdout | }
```

### DIFF
--- a/awx/main/models/ha.py
+++ b/awx/main/models/ha.py
@@ -131,6 +131,13 @@ class Instance(HasPolicyEditsMixin, BaseModel):
         grace_period = 120
         return self.modified < ref_time - timedelta(seconds=grace_period)
 
+    def mark_offline(self, on_good_terms=False):
+        self.cpu = self.cpu_capacity = self.memory = self.mem_capacity = self.capacity = 0
+        update_fields = ['capacity', 'cpu', 'memory', 'cpu_capacity', 'mem_capacity']
+        if on_good_terms:
+            update_fields.append('modified')
+        self.save()
+
     def refresh_capacity(self):
         cpu = get_cpu_capacity()
         mem = get_mem_capacity()
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -404,7 +404,11 @@ def cleanup_execution_environment_images():
 
 @task(queue=get_local_queuename)
 def check_heartbeat(node):
-    instance = Instance.objects.get(hostname=node)
+    try:
+        instance = Instance.objects.get(hostname=node)
+    except Instance.DoesNotExist:
+        logger.warn(f'Instance record for {node} missing, could not check capacity.')
+        return
     data = worker_info(node)
 
     if data['Errors']:
@@ -426,7 +430,7 @@ def check_heartbeat(node):
             instance.mem_capacity,
         )
         instance.version = 'ansible-runner-' + data['Version']
-        instance.save(update_fields=['capacity', 'version', 'cpu', 'memory', 'cpu_capacity', 'mem_capacity', 'modified'])
+        instance.save(update_fields=['capacity', 'version', 'cpu', 'memory', 'cpu_capacity', 'mem_capacity'])
         logger.info('Set capacity of execution node {} to {}, worker info data:\n{}'.format(node, instance.capacity, json.dumps(data, indent=2)))
 
 
@@ -441,31 +445,30 @@ def discover_receptor_nodes():
             continue
         (changed, instance) = Instance.objects.register(hostname=hostname, node_type='execution')
         was_lost = instance.is_lost(ref_time=nowtime)
+        last_seen = parse_date(ad['Time'])
+        if instance.modified == last_seen:
+            continue
+        instance.modified = last_seen
+        if instance.is_lost(ref_time=nowtime):
+            # if the instance hasn't advertised in awhile, don't save a new modified time
+            # this is so multiple cluster nodes do all make repetitive updates
+            continue
+
+        instance.save(update_fields=['modified'])
         if changed:
             logger.warn("Registered execution node '{}'".format(hostname))
             check_heartbeat.apply_async([hostname])
-        else:
-            last_seen = parse_date(ad['Time'])
-            if instance.modified == last_seen:
-                continue
-            instance.modified = last_seen
-            if instance.is_lost(ref_time=nowtime):
-                # if the instance hasn't advertised in awhile, don't save a new modified time
-                # this is so multiple cluster nodes do all make repetitive updates
-                continue
-
-            instance.save(update_fields=['modified'])
-            if was_lost:
-                # if the instance *was* lost, but has appeared again,
-                # attempt to re-establish the initial capacity and version
-                # check
-                logger.warn(f'Execution node attempting to rejoin as instance {hostname}.')
-                check_heartbeat.apply_async([hostname])
-            elif instance.capacity == 0:
-                # Periodically re-run the health check of errored nodes, in case someone fixed it
-                # TODO: perhaps decrease the frequency of these checks
-                logger.debug(f'Restarting health check for execution node {hostname} with known errors.')
-                check_heartbeat.apply_async([hostname])
+        elif was_lost:
+            # if the instance *was* lost, but has appeared again,
+            # attempt to re-establish the initial capacity and version
+            # check
+            logger.warn(f'Execution node attempting to rejoin as instance {hostname}.')
+            check_heartbeat.apply_async([hostname])
+        elif instance.capacity == 0:
+            # Periodically re-run the health check of errored nodes, in case someone fixed it
+            # TODO: perhaps decrease the frequency of these checks
+            logger.debug(f'Restarting health check for execution node {hostname} with known errors.')
+            check_heartbeat.apply_async([hostname])
 
 
 @task(queue=get_local_queuename)
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -102,7 +102,7 @@ from awx.main.utils.common import (
     get_mem_effective_capacity,
     get_system_task_capacity,
 )
-from awx.main.utils.execution_environments import get_default_execution_environment, get_default_pod_spec, CONTAINER_ROOT, to_container_path
+from awx.main.utils.execution_environments import get_default_pod_spec, CONTAINER_ROOT, to_container_path
 from awx.main.utils.ansible import read_ansible_config
 from awx.main.utils.external_logging import reconfigure_rsyslog
 from awx.main.utils.safe_yaml import safe_dump, sanitize_jinja
@@ -180,8 +180,7 @@ def dispatch_startup():
 def inform_cluster_of_shutdown():
     try:
         this_inst = Instance.objects.get(hostname=settings.CLUSTER_HOST_ID)
-        this_inst.capacity = 0  # No thank you to new jobs while shut down
-        this_inst.save(update_fields=['capacity', 'modified'])
+        this_inst.mark_offline(on_good_terms=True)  # No thank you to new jobs while shut down
         try:
             reaper.reap(this_inst)
         except Exception:
@@ -410,7 +409,11 @@ def check_heartbeat(node):
 
     if data['Errors']:
         formatted_error = "\n".join(data["Errors"])
-        logger.warn(f'Failed to find capacity of execution node {node}, errors:\n{formatted_error}')
+        if instance.capacity:
+            logger.warn(f'Health check marking execution node {node} as lost, errors:\n{formatted_error}')
+        else:
+            logger.info(f'Failed to find capacity of new or lost execution node {node}, errors:\n{formatted_error}')
+        instance.mark_offline()
     else:
         # TODO: spin off new instance method from refresh_capacity that calculates derived fields
         instance.cpu = data['CPU Capacity']  # TODO: rename field on runner side to not say "Capacity"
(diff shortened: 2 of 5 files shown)


awx/main/tasks.py
```diff
@@ -407,7 +405,26 @@ def cleanup_execution_environment_images():
...
+    instance = Instance.objects.get(hostname=node)
+    data = worker_info(node)
+
+    if data['Errors']:
+        formatted_error = "\n".join(data["Errors"])
+        logger.warn(f'Failed to find capacity of execution node {node}, errors:\n{formatted_error}')
+    else:
+        # TODO: spin off new instance method from refresh_capacity that calculates derived fields
+        instance.cpu = data['CPU Capacity']  # TODO: rename field on runner side to not say "Capacity"
+        instance.cpu_capacity = get_cpu_effective_capacity(instance.cpu)
```
AlanCoding, line 424: @beeankha see how we have the split `cpu` / `cpu_capacity` fields? What we want from runner is "raw" units, like "number of CPUs", so because of that, I think "CPU Capacity" is not a good name. I'd like to go through a rename, and also dump the capital first letters.

awx/main/tasks.py
```diff
@@ -407,7 +405,26 @@ def cleanup_execution_environment_images():
...
 def check_heartbeat(node):
-    AWXReceptorJob.check_heartbeat(node)
+    instance = Instance.objects.get(hostname=node)
+    data = worker_info(node)
+
+    if data['Errors']:
+        formatted_error = "\n".join(data["Errors"])
+        logger.warn(f'Failed to find capacity of execution node {node}, errors:\n{formatted_error}')
+    else:
+        # TODO: spin off new instance method from refresh_capacity that calculates derived fields
```
AlanCoding, line 422: @jladdjr I was trying to talk to you about a `refresh_capacity` refactor. The concept is there is a system check which produces in fields `cpu` and `memory`, which are in some kind of natural units, like "number of CPUs (varies by counting method)" or "bytes".

Then we need another method which is clearly logically separated to go from natural units to `cpu_capacity`, `mem_capacity`, and obviously `capacity`. These are in a form of arbitrary units. They're something the task manager uses to count stuff so it figures out where to put jobs. I want these to always appear in code in two separate processes, even if control plane nodes do them both in one go.

awx/main/tasks.py
```diff
@@ -407,7 +405,26 @@ def cleanup_execution_environment_images():
 
 @task(queue=get_local_queuename)
 def check_heartbeat(node):
-    AWXReceptorJob.check_heartbeat(node)
+    instance = Instance.objects.get(hostname=node)
+    data = worker_info(node)
+
+    if data['Errors']:
+        formatted_error = "\n".join(data["Errors"])
+        logger.warn(f'Failed to find capacity of execution node {node}, errors:\n{formatted_error}')
```
AlanCoding: Need to zero out capacity in this case. I have not done that yet.

awx/main/utils/receptor.py
```diff
@@ -0,0 +1,77 @@
...
+    data['transmit_timing'] = run_start - transmit_start
+    data['run_timing'] = 0.0
+
+    try:
+
+        resultfile = receptor_ctl.get_work_results(unit_id)
+
+        stdout = ''
+
+        while data['run_timing'] < 20.0:
```
fosterseth, line 34: maybe leave a comment that `ttl` should probably be >= this value. I can seem someone coming by to increase this timeout, but forget about the ttl in the work submission command

or 
```python3
timeout_seconds = 20.0
ttl = "{:.0f}s".format(timeout_seconds)
```

awx/main/utils/common.py
```diff
@@ -739,16 +758,8 @@ def get_mem_capacity(raw_mb=None):
...
-    elif settings_forkmem:
-        forkmem = int(settings_forkmem)
-    else:
-        forkmem = 100
-
-    if raw_mb is None:
-        raw_mb = psutil.virtual_memory().total
-    return (raw_mb, max(1, ((raw_mb // 1024 // 1024) - 2048) // forkmem))
+    mem = psutil.virtual_memory().total
+    return (mem, get_mem_effective_capacity(mem))
```
AlanCoding, line 762: @shanemcd I believe this still respects the SYSTEM_TASK_FORKS_MEM / SYSTEM_TASK_FORKS_CPU settings the same way it did before.

This will not respect SYSTEM_TASK_ABS_MEM / SYSTEM_TASK_ABS_CPU. My thinking is that it doesn't make sense for execution nodes. I could be wrong, so I wanted to ping you.

I still can't find any example of these being used for the life of me. Can you link me to anything?

awx/main/utils/receptor.py
AlanCoding: hmm.. I'm actually not totally sure if I understand what you're saying. Is it okay if both timeouts are 20 seconds?

What is most _reasonable_? Would it be better for transmit timeout to be 20 seconds, and run timeout to be 40 seconds?

awx/main/tasks.py
```diff
@@ -407,7 +404,30 @@ def cleanup_execution_environment_images():
...
+        instance.cpu_capacity = get_cpu_effective_capacity(instance.cpu)
+        instance.memory = data['Memory Capacity'] * 1000  # TODO: double-check the multiplier here
+        instance.mem_capacity = get_mem_effective_capacity(instance.memory)
+        instance.capacity = get_system_task_capacity(
+            instance.capacity_adjustment,
+            instance.cpu_capacity,
+            instance.mem_capacity,
+        )
+        instance.version = 'ansible-runner-' + data['Version']
+        instance.save(update_fields=['capacity', 'version', 'cpu', 'memory', 'cpu_capacity', 'mem_capacity', 'modified'])
```
AlanCoding: I've realized that I shouldn't have `modified` listed here.

awx/main/tasks.py
```diff
@@ -422,23 +442,29 @@ def discover_receptor_nodes():
         (changed, instance) = Instance.objects.register(hostname=hostname, node_type='execution')
         was_lost = instance.is_lost(ref_time=nowtime)
         if changed:
-            logger.info("Registered execution node '{}'".format(hostname))
+            logger.warn("Registered execution node '{}'".format(hostname))
             check_heartbeat.apply_async([hostname])
         else:
```
AlanCoding: I think this shouldn't be an `else:`. If it's a new execution node, then we should still set `modified` to the last-seen date as viewed by receptor.

awx/main/utils/receptor.py
fosterseth: yeah it's fine if both are 20 seconds. By transmit timeout you mean `ttl`?

After the receptor job expires due to ttl (20 seconds), it will have a Fail state in receptor. Then in your code the `if state_name` conditional hits and we break the while loop. Therefore, increasing the `while data['run_timing'] < 20.0:` to say, 500 seconds, would still have no real effect because we'll still break the loop after 20 seconds due to the ttl.

As such, it's probably best if we make sure ttl matches this `while data['run_timing']` timeout

AlanCoding:
This now has all the really important updates on the agenda, and the rest of discussions will be in another issue.

awx/main/models/ha.py
```diff
@@ -131,6 +131,13 @@ def is_lost(self, ref_time=None):
         grace_period = 120
         return self.modified < ref_time - timedelta(seconds=grace_period)
 
+    def mark_offline(self, on_good_terms=False):
```
jladdjr, line 134: I'm still reading through the PR, so don't have the full context yet, but maybe `verified` or `by_request` instead of `on_good_terms`?

awx/main/models/ha.py
```diff
@@ -131,6 +131,13 @@ def is_lost(self, ref_time=None):
...
+        self.cpu = self.cpu_capacity = self.memory = self.mem_capacity = self.capacity = 0
```
jladdjr, line 135: should we also mark the instance as disabled (`self.enabled = False` iirc)?

awx/main/utils/common.py
```diff
@@ -699,12 +699,25 @@ def parse_yaml_or_json(vars_str, silent_failure=True):
     return vars_dict
 
 
-def get_cpu_capacity(raw=None):
+def get_cpu_effective_capacity(cpu_count):
```
jladdjr, line 702: These env vars are still so confusing to me, but I'm glad to see some of this refactored out a helper method

awx/main/utils/receptor.py
```diff
@@ -0,0 +1,77 @@
...
+
+        res = receptor_ctl.simple_command(f"work release {unit_id}")
+        if res != {'released': unit_id}:
+            logger.warn(f'Could not confirm release of receptor work unit id {unit_id} from {node_name}, data: {res}')
+
+        receptor_ctl.close()
+
+    if state_name.lower() == 'failed':
+        work_detail = status.get('Detail', '')
+        if not work_detail.startswith('exit status'):
```
jladdjr, line 57: I wonder if some of this scraping is going to get us into trouble. Seems like someone could change some of the verbiage in receptor and this section may start to silently fail / start following a different codepath? Maybe it's more stable than I realize. I also realize that there may not be a clear alternative atm.

awx/main/utils/receptor.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+logger = logging.getLogger('awx.main.utils.receptor')
+
+
+def get_receptor_ctl():
+    return ReceptorControl('/var/run/receptor/receptor.sock')
+
+
+def worker_info(node_name):
```
jladdjr, line 15: This method is focused on the `--worker-info`, but there's a lot of great error handling here that it seems like we should use whenever we invoke `ansible-runner` w/ receptor. Could we turn this into a more generic helper for that?

awx/main/utils/receptor.py
```diff
@@ -0,0 +1,77 @@
+import logging
```
jladdjr, line 1: Really like having this new util library.

awx/main/models/ha.py
AlanCoding: the distinction is whether it is marked offline due to judgement of non-compliance by a peer, or the node marks itself offline as it conducts a normal shutdown.

So I mean to express that it's an amicable breakup from the cluster in one case. I don't like any of the other namings quite as much.

awx/main/models/ha.py
AlanCoding: No, the `enabled` field should be totally separate. It may actually come-and-go while being disabled, because the enabled flag is something the user will have control over. These fields should become the sole domain of the health checks.

awx/main/utils/common.py
AlanCoding: yeah, it's still not great. I expect these will have continued revisions as we keep going.

awx/main/utils/receptor.py
AlanCoding: some of it could be pushed upstream, and we've absolutely talked about that.

The question asked here is a very reasonable request - did you get an exit code from the work unit I told you to run?

If you're interested in refining this, we get a status dictionary earlier in the code, and you can print that out and compare the different cases. For this case, I don't remember seeing any data that was more straight to the point.

If we moved this functionality upstream, it could be something like `receptor_ctl.get_exit_code()` and return None / 0 / some other int.

awx/main/utils/receptor.py
AlanCoding: I do hope this inspires some additional error handling for jobs, but it's hard to see code sharing. Here, we start a fire-and-forget job and then poll until it's finished, and then read the entire standard out. That's way more simple than what jobs do

beeankha (approved):

Issue #10628, opened 2021-07-12
No Convergence Data in Workflow Node Info Modal


AlexSCorey:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

We need to include convergence data in a workflow node info modal

### AWX version

devel

### Installation method

docker for mac

### Modifications

no

### Web browser

Chrome

### Steps to reproduce

Create a workflow node.  Hover over the node and click on the `i` icon.  

### Expected results

See all the data about the node including the convergence data.

### Actual results

No convergence data included

### Additional information

![Screen Shot 2021-07-12 at 3 33 43 PM](https://user-images.githubusercontent.com/39280967/125345467-94d23a80-e326-11eb-8f28-a9dd4f7d078b.png)
![Screen Shot 2021-07-12 at 3 33 38 PM](https://user-images.githubusercontent.com/39280967/125345473-96036780-e326-11eb-97cb-edf49d49402b.png)

akus062381:
upon testing this, found a bug when clicking on the info button of a node. Wambugu is working on filing a bug.

AlexSCorey:
@akus062381 Can we close this issue?

akus062381:
@AlexSCorey because of the bug, I am unable to verify whether this particular issue has been solved, so i'd say we are blocked until https://github.com/ansible/awx/issues/11061 is fixed

akus062381:
Duplicate: https://github.com/ansible/awx/issues/10945

InnocentK:
Issue causing the block (#11061) was resolved

InnocentK:
Successfully resolved

Issue #10549, opened 2021-06-29
Auditors can try to associate a Host with a Group


tiagodread:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

The Add button is enable to auditors users 

##### ENVIRONMENT
* AWX version: X.Y.Z
* AWX install method: openshift, minishift, docker on linux, docker for mac, boot2docker
* Ansible version:  X.Y.Z
* Operating System:
* Web Browser:

##### STEPS TO REPRODUCE

1. Login as auditor
2. Click on hosts
3. Go to Groups tab

##### EXPECTED RESULTS

Add button to be disabled
##### ACTUAL RESULTS

![image](https://user-images.githubusercontent.com/12464858/123857917-4c178c00-d8f9-11eb-8348-5b1fd0159ba2.png)

akus062381:
@AlexSCorey - Upon investigation, Tiago and I noticed that the Associate button on the Groups tab inside of a host is no longer present in the DOM, even for an Admin user. I am reopening this so it can be addressed.

AlexSCorey:
it seems that the auditor can attempt to associate a group, but the api isn't actually allowing it if you try.

AlexSCorey:
Filling this as an api bug as it seems that the api is misreporting the permissions.  If I hit `/hosts/id/groups` both super user and system auditor have associate and disassociate permissions and if I hit `/hosts/is/all_groups` neither can associate/disassociate.  However, if I actually try to associate a group as a system auditor  by hitting `/hosts/id/groups` the api will not allow it and sends a 403.

PR #10828, opened 2021-08-06
Addresses permission problem on host group list and adds convergence data to Prompt Detail

AlexSCorey:
##### SUMMARY

Addresses #10628 and #10549.

Re: 10549-We essentially have 2 end points we can hit with an OPTIONS request to get the permissions data. They are `/hosts/id/groups` where both super users and system administrator erroneously have POST permissions, or `hosts/id/all_groups` where neither have POST permissions, which is also an error.  However, when hitting `/hosts/id/groups` a when associating or disassociating the api returns a 403 for system auditor.

##### ISSUE TYPE
 - Bugfix Pull Request
##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
Screen Shot 2021-08-06 at 11 07 27 AM
Screen Shot 2021-08-06 at 11 07 22 AM
Screen Shot 2021-08-06 at 11 07 10 AM

### DIFF
--- a/awx/ui/src/components/PromptDetail/PromptDetail.js
+++ b/awx/ui/src/components/PromptDetail/PromptDetail.js
@@ -87,7 +87,12 @@ function omitOverrides(resource, overrides, defaultConfig) {
   return clonedResource;
 }
 
-function PromptDetail({ resource, launchConfig = {}, overrides = {} }) {
+function PromptDetail({
+  resource,
+  launchConfig = {},
+  overrides = {},
+  workflowNode = false,
+}) {
   const VERBOSITY = {
     0: t`0 (Normal)`,
     1: t`1 (Verbose)`,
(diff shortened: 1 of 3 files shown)


awx/ui/src/components/PromptDetail/PromptDetail.js
```diff
@@ -87,7 +87,12 @@ function omitOverrides(resource, overrides, defaultConfig) {
   return clonedResource;
 }
 
-function PromptDetail({ resource, launchConfig = {}, overrides = {} }) {
+function PromptDetail({
+  resource,
+  launchConfig = {},
+  overrides = {},
+  workflowNode,
```
nixocio: Do we need add a default value for `workflowNode`? Since PromptDetail is used in different places.

nixocio (approved):

PR #10834, opened 2021-08-07
fixed module name to FQCN in example

akira6592:
##### SUMMARY

Fixed the example module name in the documentation to FQCN to avoid confusing for example  `awx.awx.user` and `ansible.builtin.user`.

##### ISSUE TYPE

 - Docs Pull Request

##### COMPONENT NAME

 - awx.awx collection

##### AWX VERSION

```
awx: 19.2.2
```

##### ADDITIONAL INFORMATION

```

```


AlanCoding (changes requested):
This isn't actually true in all cases, because the same source is vendored as the `ansible.controller` collection on Automation Hub.

akira6592:
I understood. and  close this pr.

Issue #10610, opened 2021-07-08
Introduce option for WorkflowApprovalTemplates so users cannot approve their own workflow job


davidburkhart:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
At the moment the right to allow approvals is only dependent on the approval permission or the super user role. Therefore a user who can start a workflow job and also has the right to approve, can approve their own jobs.

Sometimes teams want to apply something like _"two-man rule"_ or _"4-eye principle"_, where everybody in the team has the right to start a job, but a second member of the team should approve.

###### Suggestion
Add a setting on the level of WorkflowApprovalTemplate named e.g. "Self approval forbidden", which is false by default to not break existing approval templates.

When enabled, this setting enforces the approver to be somebody else than the person who started the workflow job.

In `awx/main/access.py` the right to approve is checked in the method WorkflowApprovalAccess#can_approve_or_deny. The new code could look something like this:

```python
    def can_approve_or_deny(self, obj):
        
        # new code starting here
        if obj.approval_template and obj.approval_template.cannotApproveSelf:
            return self.user != obj.created_by
        # new code ending here

        if (obj.workflow_job_template and self.user in obj.workflow_job_template.approval_role) or self.user.is_superuser:
            return True
```

GitGeshizzle:
This would be really great feature as the current setup only works with two separate teams which is kind of the opposite as most „DevOps Teams“ work.

Another approach to solve that problem would be to configure an "approval counter“ where we could configure something like „this template needs at least two approvals“ where obviously the approvers should be unique accounts.

AlexSCorey:
@shanemcd @jakemcdermott @AlanCoding @jbradberry @rebeccahhh 
The current PR for this issue takes the approach of allowing the user to set the approval permissions on each node of the workflow.  This has some merit and allows granularity but I do think we should also consider whether we want to take a different approach and allow the user to set this permission at the Organization level or at the server level.  

I tend to lean to the server level, or to allow approval setting on individual nodes.  What do you all think?

rebeccahhh:
I think a setting like this would be best enabled/disabled at a higher level. I like the feature idea, I think it would be a nice thing to have as a user.

AlanCoding:
Good points about the design of the functionality, a larger perspective about what users need is appropriate, and I wouldn't want to move on this without involving @ffirg

PR #10847, opened 2021-08-10
Added setting for disallow to approve their own workflows

x0r3ax3ax:
##### SUMMARY

Added setting for disallow users to approve their own workflow jobs

#10610 

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 - UI

##### AWX VERSION

```
19.2.2
```

##### ADDITIONAL INFORMATION

```

```


AlexSCorey:
@oweel. This is really great work. Thank you for submitting this.  I pulled this code down and went through the steps of creating 2 workflow approval nodes.  One of them was set up to allow self approval of the node, and the other was set up to not allow self approval.  I was able to self approve the 1 that was set up to not allow self approval.  We need to add some permission.  We should add some tests for this use case as well.  
Screen Shot 2021-08-12 at 11 18 26 AM

x0r3ax3ax:
@AlexSCorey. Thank you for your feedback!
 
Maybe you do it under the admin user? Is this case -  workflow approve is allow for superusers.
https://github.com/ansible/awx/blob/61a3074800ba479c1a7d05dea2af5d0a918d7b96/awx/main/access.py#L2859-L2866

AlexSCorey:
@oweel For my 2nd run through of this work I created 2 users that have admin permission on the workflow job template, but are not super users.  User 1 created the approval node and launched the job.  When that user goes to approve the job the approval button is active (it should be disabled), but when that user presses the button they get an api error (as I would expect saying the user doesn't have permission to approve their own workflow).  

Then, I signed in as User 2.  When this user goes to approve the node the Approval button is active (as expected) but when they click the approve button they also get an api error saying  the user doesn't have permission to approve their own workflow.

x0r3ax3ax:
> @oweel For my 2nd run through of this work I created 2 users that have admin permission on the workflow job template, but are not super users. User 1 created the approval node and launched the job. When that user goes to approve the job the approval button is active (it should be disabled), but when that user presses the button they get an api error (as I would expect saying the user doesn't have permission to approve their own workflow).
> 
> Then, I signed in as User 2. When this user goes to approve the node the Approval button is active (as expected) but when they click the approve button they also get an api error saying the user doesn't have permission to approve their own workflow.

@AlexSCorey Sorry, i fix them. Thank you for the bug found! 
Also, I made the "Approve" button inactive if the user does not have permission to.

AlexSCorey:
@oweel can you rebase this PR so that I can review it?  I think it is a few commits behind devel

x0r3ax3ax:
@AlexSCorey All is ready :)

beeankha:
Hi @oweel , could you please renumber your migration file (to `0156`) as well as rename it  (currently named `awx/main/migrations/0153_auto_20210804_0938.py`)? 

You can see the pattern of numbering and naming in this directory: https://github.com/ansible/awx/tree/devel/awx/main/migrations 

`0156_self_approve_wfjt.py` might be something that works, but feel free to choose a different name.

AlexSCorey:
@oweel I think this needs another rebase

AlexSCorey:
Not a priority

AlexSCorey (approved):

PR #10856, opened 2021-08-10
Update awxkit random_title setting non_ascii to False

tiagodread:
##### SUMMARY

Working on Bumping Cypress version to latest, I realized that there is an [issue](https://github.com/cypress-io/cypress/issues/17055) with `cy.intercept()` and special chars.

I'm updating `random_title` and setting `non_ascii` default to False, based on the comment, this was because a bug related to selenium webdriver, however, Cypress works in a different way and this is not necessary.

### DIFF
--- a/awxkit/awxkit/utils/__init__.py
+++ b/awxkit/awxkit/utils/__init__.py
@@ -277,7 +277,7 @@ def random_utf8(*args, **kwargs):
 
 def random_title(num_words=2, non_ascii=True):
     if os.getenv('AWXKIT_FORCE_ONLY_ASCII', False):
-        non_ascii=False
+        non_ascii = False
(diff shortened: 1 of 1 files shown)


jakemcdermott (approved):

kdelee (commented):
Sorry forgot to finish this review. This will solve your problem and is the simplest solution, I wonder though if it would be possible to keep the default True as that does provide some coverage in other use cases than your own. But maybe we could have a setting or an env var that overrides it. like

```
def random_title(num_words=2, non_ascii=True):
    if os.getenv('AWXKIT_FORCE_ONLY_ASCII', False):
        non_ascii=False
```

kdelee (approved):

PR #10858, opened 2021-08-11
Install receptor and ansible-runner from devel branch

rooftopcellist:
Signed-off-by: Christian M. Adams <rooftopcellist@gmail.com>

##### SUMMARY
To keep pace with changes in receptor and ansible-runner during development, we are switching to installing directly from the devel branch for each.  

##### COMPONENT NAME

 - Install

##### AWX VERSION

```
19.2.0
```

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -4,8 +4,7 @@ aiohttp==3.6.2
     # via -r /awx_devel/requirements/requirements.in
 aioredis==1.3.1
     # via channels-redis
-ansible-runner==2.0.1
-    # via -r /awx_devel/requirements/requirements.in
+    # via -r /awx_devel/requirements/requirements_git.txt
 ansiconv==1.0.0
     # via -r /awx_devel/requirements/requirements.in
 asciichartpy==1.5.25
@@ -303,8 +302,7 @@ pyyaml==5.4.1
     #   djangorestframework-yaml
     #   kubernetes
     #   receptorctl
-receptorctl==1.0.0
-    # via -r /awx_devel/requirements/requirements.in
+    # via -r /awx_devel/requirements/requirements_git.txt
 redis==3.4.1
     # via
     #   -r /awx_devel/requirements/requirements.in
--- a/awx/main/tests/functional/test_licenses.py
+++ b/awx/main/tests/functional/test_licenses.py
@@ -60,6 +60,8 @@ def test_python_and_js_licenses():
                     (name, version) = reqt.link.filename.split('@', 1)
                     if name.endswith('.git'):
                         name = name[:-4]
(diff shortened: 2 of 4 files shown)


requirements/requirements.txt
```diff
@@ -4,8 +4,7 @@ aiohttp==3.6.2
     # via -r /awx_devel/requirements/requirements.in
 aioredis==1.3.1
     # via channels-redis
-ansible-runner==2.0.1
-    # via -r /awx_devel/requirements/requirements.in
+    # via -r /awx_devel/requirements/requirements_git.txt
```
shanemcd, line 7: This line seems wrong.

requirements/requirements.txt
```diff
@@ -303,8 +302,7 @@ pyyaml==5.4.1
     #   djangorestframework-yaml
     #   kubernetes
     #   receptorctl
-receptorctl==1.0.0
-    # via -r /awx_devel/requirements/requirements.in
+    # via -r /awx_devel/requirements/requirements_git.txt
```
shanemcd, line 305: This line seems wrong too.

rooftopcellist:
We should rely on updater.sh to put all python deps in `requirements.txt` and use `requirements_git.in` (along with `requirements.in` as a source file for creating requirements.txt.  

  - updater.sh should be run whenever requirements are changed
  - pip-compile (run by updater.sh) will merge the requirements_git.in
    into requirements.txt

@shanemcd I'd love your opinion on this, I'm happy to back out the last commit if it doesn't make sense.

rooftopcellist:
It turns out separating these `_git` requirements out into a separate file is needed for downstream packaging reasons, _reverting_.

requirements/requirements.txt
rooftopcellist: This is an unfortunate side effect of the `pip-compile` part of the updater.sh script, and the fact that we remove all `git+` dependencies it adds to `requirements.txt`.  

The pattern so far has just been to leave this in (that's what we did for certifi it looks like).

AlanCoding: oh... so it gets installed by a separate `pip install -r ...` command?

rooftopcellist: @AlanCoding yes, this happens here upstream - https://github.com/ansible/awx/blob/devel/Makefile#L139

kdelee:
also necessary to get all the newest bits everywhere:

https://github.com/ansible/awx-operator/pull/495
https://github.com/ansible/awx/pull/10861
https://github.com/ansible/awx-ee/pull/84
https://github.com/ansible/awx-ee/pull/85

Some of these need approval, most I think I don't have power to merge

shanemcd (approved):

PR #10859, opened 2021-08-11
Remove warning for SSO session when logging in

nixocio:
Remove warning for SSO session when logging in

See: #10860

### DIFF
--- a/awx/ui/src/contexts/Session.js
+++ b/awx/ui/src/contexts/Session.js
@@ -6,6 +6,7 @@ import React, {
   useCallback,
 } from 'react';
 import { useHistory } from 'react-router-dom';
+import { DateTime } from 'luxon';
 import { RootAPI, MeAPI } from 'api';
 import { isAuthenticated } from 'util/auth';
 import { SESSION_TIMEOUT_KEY } from '../constants';
@@ -66,8 +67,8 @@ SessionContext.displayName = 'SessionContext';
 function SessionProvider({ children }) {
   const history = useHistory();
   const isSessionExpired = useRef(false);
-  const sessionTimeoutId = useRef();
-  const sessionIntervalId = useRef();
+  const sessionTimeoutId = useRef(null);
+  const sessionIntervalId = useRef(null);
   const [sessionTimeout, setSessionTimeout] = useStorage(SESSION_TIMEOUT_KEY);
   const [sessionCountdown, setSessionCountdown] = useState(0);
   const [authRedirectTo, setAuthRedirectTo] = useState('/');
@@ -89,8 +90,15 @@ function SessionProvider({ children }) {
       return () => {};
     }
 
-    const calcRemaining = () =>
-      parseInt(sessionTimeout, 10) - new Date().getTime();
+    const calcRemaining = () => {
+      if (sessionTimeout) {
+        return Math.max(
+          parseInt(sessionTimeout, 10) - DateTime.now().toMillis(),
+          0
+        );
+      }
+      return 0;
+    };
 
     const handleSessionTimeout = () => {
       let countDown = SESSION_WARNING_DURATION;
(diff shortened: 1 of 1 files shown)


awx/ui/src/contexts/Session.js
```diff
@@ -89,21 +89,28 @@ function SessionProvider({ children }) {
...
     }
 
-    const calcRemaining = () =>
-      parseInt(sessionTimeout, 10) - new Date().getTime();
+    const calcRemaining = () => {
+      const remainingTime = parseInt(sessionTimeout, 10) - new Date().getTime();
+      if (remainingTime < 0) {
+        return 0;
+      }
+      return remainingTime;
```
jakemcdermott:
```suggestion
    const calcRemaining = () =>
      Math.max(0, parseInt(sessionTimeout, 10) - new Date().getTime());
```

awx/ui/src/contexts/Session.js
```diff
@@ -89,21 +89,28 @@ function SessionProvider({ children }) {
...
+
+        sessionIntervalId.current = setInterval(() => {
+          if (countDown > 0) {
+            setSessionCountdown(--countDown);
+          } else {
+            isSessionExpired.current = true;
+            logout();
+          }
+        }, 1000);
+      }
```
jakemcdermott:
```suggestion
      if (sessionCountdown <= 0) {
        return;
      } 
      setSessionCountdown(countDown);
      sessionIntervalId.current = setInterval(() => {
        if (countDown > 0) {
          setSessionCountdown(--countDown);
        } else {
          isSessionExpired.current = true;
          logout();
        }
      }, 1000);
```

awx/ui/src/contexts/Session.js
nixocio: @jakemcdermott neat.

awx/ui/src/contexts/Session.js
```diff
@@ -89,11 +89,15 @@ function SessionProvider({ children }) {
       return () => {};
     }
 
-    const calcRemaining = () =>
-      parseInt(sessionTimeout, 10) - new Date().getTime();
+    const calcRemaining = () => {
+      Math.max(0, parseInt(sessionTimeout, 10) - new Date().getTime());
```
kialam: let's use luxon for current time

awx/ui/src/contexts/Session.js
```diff
@@ -110,10 +110,12 @@ function SessionProvider({ children }) {
     clearTimeout(sessionTimeoutId.current);
     clearInterval(sessionIntervalId.current);
 
+    const calcTimeOut = calcRemaining() - SESSION_WARNING_DURATION * 1000;
+
     isSessionExpired.current = false;
     sessionTimeoutId.current = setTimeout(
       handleSessionTimeout,
-      Math.min(calcRemaining() - SESSION_WARNING_DURATION * 1000, MAX_TIMEOUT)
```
nixocio, line 116: I tested this again. We were using a MAX_TIMEOUT to compare Math.min, and then the other value was always being returned - since it was always smaller. If I missed something let me know.

awx/ui/src/contexts/Session.js
nixocio: @kialam, I can update to use luxon, but this is how we dealt with the same conversion on `Base.js`. This is basically the opposite operation. 

```
defaultHttp.interceptors.response.use((response) => {
  const timeout = response?.headers['session-timeout'];
  if (timeout) {
    const timeoutDate = new Date().getTime() + timeout * 1000;
    updateStorage(SESSION_TIMEOUT_KEY, String(timeoutDate));
  }
  return response;
});
```

kialam: Yup, we are using a date time library so I think it makes sense to start using it moving forward. The luxon equivalent is `Datetime.now().milliseconds`.

awx/ui/src/contexts/Session.js
jakemcdermott: Make sure to consider multiple tabs, etc. Is there a specific reason we're removing this?

tiagodread (approved):
Verified that the modal is not being displayed anymore, only when necessary, also added a test case to cover several scenarios on 6744

awx/ui/src/contexts/Session.js
```diff
@@ -89,8 +90,15 @@ function SessionProvider({ children }) {
       return () => {};
     }
 
-    const calcRemaining = () =>
-      parseInt(sessionTimeout, 10) - new Date().getTime();
+    const calcRemaining = () => {
```
kialam, line 93: Suggestion: you could also write this as a ternary expression:

```js
const remainingTime = sessionTimeout ? Math.max(...) : 0;
```

kialam (approved):
👍

awx/ui/src/contexts/Session.js
nixocio: Thanks @kialam.

PR #10861, opened 2021-08-11
Use awx-ee:latest to get latest receptor and runner

kdelee:
We are updating the requirements to get the latest receptor and runner in the task container,
we should also have the latest in the EE

@simaishi @rooftopcellist 

Still need to do this in awx-operator and also make sure awx-ee devel branch is getting rebuilt more often and pushed to latest tag


kdelee:
related to https://github.com/ansible/awx/pull/10858 and https://github.com/ansible/awx-operator/pull/495

rooftopcellist (approved):

simaishi (approved):

PR #10870, opened 2021-08-12
Model changes for instance last_seen field to replace modified

AlanCoding:
The problem is that `modified` field logic wasn't working for execution nodes.

We tried to save the receptor "Last seen" value to the field, but every time it just saved the current time, which isn't correct, and we can't mark lost nodes offline with this method.

So this adds a new field so that we can actually solve the problem. There's still more work to do as of creating this PR.


awx/main/migrations/0153_instance_last_seen.py
```diff
@@ -0,0 +1,27 @@
...
+            field=models.DateTimeField(
+                editable=False,
+                help_text='Last time instance ran its heartbeat task for main cluster nodes. Last known connection to receptor mesh for execution nodes.',
+                null=True,
+            ),
+        ),
+        migrations.AlterField(
+            model_name='instance',
+            name='memory',
+            field=models.BigIntegerField(default=0, editable=False, help_text='Total system memory of this instance in bytes.'),
```
jladdjr, line 25: Was wondering what changed here. Looks like it was just the `help_text`?

awx/main/models/ha.py
```diff
@@ -126,17 +137,24 @@ def choose_online_control_plane_node():
         return random.choice(Instance.objects.filter(enabled=True).filter(node_type__in=['control', 'hybrid']).values_list('hostname', flat=True))
 
     def is_lost(self, ref_time=None):
```
jladdjr, line 139: 👍

awx/main/migrations/0153_instance_last_seen.py
AlanCoding: yeah, exactly. None of these `Instance` fields have help text. I started out thinking I would write help_text for each one of them (since I was writing a migration anyway). But when I started on the first one, I realized that the values were wrong and filed https://github.com/ansible/ansible-runner/issues/788

AlanCoding:
This is made against the `mesh_code` branch, which has a PR at https://github.com/ansible/awx/pull/10727

awx/settings/defaults.py
```diff
@@ -420,10 +420,18 @@ def IS_TESTING(argv=None):
 # Set default ports for live server tests.
 os.environ.setdefault('DJANGO_LIVE_TEST_SERVER_ADDRESS', 'localhost:9013-9199')
 
+# heartbeat period can factor into some forms of logic, so it is maintained as a setting here
+CLUSTER_NODE_HEARTBEAT_PERIOD = 60
+RECEPTOR_SERVICE_ADVERTISEMENT_PERIOD = 60  # https://github.com/ansible/receptor/blob/aa1d589e154d8a0cb99a220aff8f98faf2273be6/pkg/netceptor/netceptor.go#L34
```
AlanCoding, line 424: from discussing with @fosterseth 

If you follow the use of this setting, you can paint a picture of worst-case timing. The execution node is seen by the receptor mesh at 1 second after 1pm, so 13:00:01, which just missed the heartbeat. The control node runs its next heartbeat at 13:01:00.

But there might be a timing issue where control node 1 and control node 2 run almost at the same time. Anyway, we add these numbers together because we want to be _sure_ that the node is lost before marking its capacity to 0.

awx/api/views/__init__.py
```diff
@@ -370,6 +370,16 @@ class InstanceDetail(RetrieveUpdateAPIView):
     model = models.Instance
     serializer_class = serializers.InstanceSerializer
 
+    def update(self, request, *args, **kwargs):
+        r = super(InstanceDetail, self).update(request, *args, **kwargs)
+        if status.is_success(r.status_code):
+            obj = self.get_object()
+            if obj.capacity != 0:
+                obj.set_capacity_value()
+                obj.save()
```
AlanCoding: @matburt the prior state of code here is that we called `obj.refresh_capacity()` when someone adjusted the capacity adjustment slider. That method would actually run `psutil.virtual_memory().total` on the node the request was made to. Realistically, all nodes probably have the same memory/CPU specs for ordinary clusters. Moving forward with execution nodes, however, I'd expect them to be very different machines.

So now, the capacity adjustment changes the `capacity` field and that's it. I just realized I should add `update_fields` to make that super clear.

AlanCoding:
Here's a link to some motivating issues for this:

https://gist.github.com/AlanCoding/4a929a8cd16738c111121bb7cf8a0b47

Some of this could be reused for developer docs, but not all.

awx/main/models/ha.py
```diff
@@ -61,24 +62,34 @@ class Instance(HasPolicyEditsMixin, BaseModel):
         max_length=50,
         unique=True,
     )
+    # Auto-fields, implementation is different from BaseModel
     created = models.DateTimeField(auto_now_add=True)
     modified = models.DateTimeField(auto_now=True)
+    # Fields defined in health check or heartbeat
```
beeankha, line 68: Yay for helpful comments!

beeankha (approved):

AlanCoding:
Sorry for the churn, test results found a bug in the implementation. I pushed a commit that addresses this and I confirmed that tests pass after that change.

awx/api/views/__init__.py
matburt: 👍🏻 that works for me.

Issue #10808, opened 2021-08-03
Add worker thread settings to external_logging.py to help combat performance issues


msilmser:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Some users may run into performance issues shipping logs to external loggers due to default rsyslog settings. Adding worker thread settings in the rsyslog.conf file can help increase that performance which enables the external loggers to receive logs significantly faster. This could be done by making these thread values configurable in the API similar to the queue.spoolDirectory and queue.maxdisksize values.

Making the following change to increase those worker thread values enabled a user to be able to access these logs on their external service close to two hours earlier during job runs with a large number (~250k) of events:

```
f'main_queue(queue.spoolDirectory="{spool_directory}" queue.maxdiskspace="{max_disk_space}g" queue.type="Disk" queue.filename="awx-external-logger-backlog" queue.workerthreadminimummessages="100" queue.workerthreads="50")',
```

Before the change, the user was seeing the jobs take upwards of 2 hours to populate externally while after this change they populated in ~10sec. In addition to this change, they also increased the max_disk_space value from 1g to 10g but this value is already easily configurable from the API.

ckabalan:
This solved our syslog queueing problem. With the defaults we were consistently seeing large jobs take hours to output their logs to Sumo Logic (one of the preset log providers). We monitored this by monitoring the queue file sizes and saw them growing to ~130mb and then stairstep down.

Default behavior (~250k event job):

![image](https://user-images.githubusercontent.com/6820600/128759306-cf310cf3-acdf-429f-83ed-26c43978cf82.png)

With threading tweaks set to 100 thread minimum and 50 workers we saw only 14 workers created in the rsyslog debug log, but sub-10-second log output and a much more manageable queue pattern like this (note the much smaller Y axis and sharp drop after job completion). Based on my reading of the rsyslog documenation this would lead me to believe we only had ~1,400 events in flight at any given time, which is probably reasonable for our environment.

![image](https://user-images.githubusercontent.com/6820600/128759320-34e97ebb-ade2-4124-bd86-783adb53417f.png)

I'm happy to provide any more insight into this issue if there are questions.

PR #10874, opened 2021-08-12
Added worker thread settings to external_logging  for optimize performance

x0r3ax3ax:
##### SUMMARY

#10808 

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API
 - UI

##### AWX VERSION

```
19.2.2
```

##### ADDITIONAL INFORMATION

```

```


wenottingham:
Is just setting a number of threads sufficient? When would a user change the default for the messages per thread and not just add another thread?

x0r3ax3ax:
@wenottingham i set the default to 0 for these settings and made them optional in the rsyslog config builder. Now users will control these settings themselves or will leave the default values that are calculated in rsyslog. I think this behavior will be more transparent.

AlexSCorey:
Not a priority

PR #10886, opened 2021-08-16
Update path to reflect correct version

nntrn:
##### SUMMARY
Update pg path to correct version in `tools/ansible/roles/dockerfile/templates/Dockerfile.j2`

##### AWX VERSION

```
19.3.0
```


shanemcd:
Hello - can you please rebase this so the tests will run?

Issue #10718, opened 2021-07-22
Editing a managed job schedule can return null fields


kialam:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

It is possible to edit a managed job schedule and have the API return `null` for requisite fields.

Discovered when looking into #10587

![time_warp](https://user-images.githubusercontent.com/2293210/126654476-c840f34b-db6c-4d89-b27e-26b7e2952b89.gif)

### AWX version

AWX 19.2.2

### Installation method

docker for mac

### Modifications

no

### Steps to reproduce

1. Go to Management Jobs > Edit Job Schedule
2. Edit the start date of a management job
3. Select "On date" for when a management job ends
4. Enter end date
5. Observe that the UI shows "1970-01-01" for the start date section

Request payload is as follows:

```js
daysToKeep: "120"
description: "Automatically Generated Schedule"
extra_data: {days: "120"}
days: "120"
name: "Cleanup Job Schedule"
occurrences: 1
rrule: "DTSTART;TZID=Europe/Berlin:20210802T000100 RRULE:INTERVAL=1;FREQ=WEEKLY;BYDAY=SU;UNTIL=20210803T000600"
runOnTheOccurrence: 1
```

Response payload from API:
```js
created: "2021-07-06T20:50:04.085408Z"
description: "Automatically Generated Schedule"
diff_mode: null
dtend: null
dtstart: null
enabled: true
extra_data: {days: "120"}
id: 1
inventory: null
job_tags: null
job_type: null
limit: null
modified: "2021-07-22T14:09:24.333570Z"
name: "Cleanup Job Schedule"
next_run: null
related: {modified_by: "/api/v2/users/1/", credentials: "/api/v2/schedules/1/credentials/",…}
rrule: "DTSTART;TZID=Europe/Berlin:20210802T000100 RRULE:INTERVAL=1;FREQ=WEEKLY;BYDAY=SU;UNTIL=20210802T220600Z"
scm_branch: null
skip_tags: null
summary_fields: {,…}
timezone: "Europe/Berlin"
type: "schedule"
unified_job_template: 1
until: "2021-08-03T00:06:00"
url: "/api/v2/schedules/1/"
verbosity: null
[...]

PR #10898, opened 2021-08-17
Validate that start/end datetime creates at least 1 schedule.

kialam:
##### SUMMARY

Related #10718 

Include client-side validation for when a user is about to create a schedule set that has no entries. E.g. a schedule that repeats every Sunday however the start and end dates are Mon-Tues will result in an empty schedules list. We now show them a form error if such a scenario is encountered:

rrule_validation

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - UI

##### AWX VERSION

```
19.3.0
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/ui/package.json
+++ b/awx/ui/package.json
@@ -29,7 +29,7 @@
     "react-error-boundary": "^3.1.3",
     "react-router-dom": "^5.1.2",
     "react-virtualized": "^9.21.1",
-    "rrule": "^2.6.4",
+    "rrule": "2.6.4",
     "sanitize-html": "2.4.0",
     "styled-components": "5.3.0"
   },
--- a/awx/ui/src/components/Schedule/shared/ScheduleForm.js
+++ b/awx/ui/src/components/Schedule/shared/ScheduleForm.js
@@ -30,6 +30,13 @@ import {
 import FrequencyDetailSubform from './FrequencyDetailSubform';
 import SchedulePromptableFields from './SchedulePromptableFields';
 import DateTimePicker from './DateTimePicker';
+import buildRuleObj from './buildRuleObj';
+
+const NUM_DAYS_PER_FREQUENCY = {
+  week: 7,
+  month: 31,
+  year: 365,
+};
 
 const generateRunOnTheDay = (days = []) => {
   if (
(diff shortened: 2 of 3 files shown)


awx/ui/package.json
```diff
@@ -29,7 +29,7 @@
     "react-error-boundary": "^3.1.3",
     "react-router-dom": "^5.1.2",
     "react-virtualized": "^9.21.1",
-    "rrule": "^2.6.4",
+    "rrule": "2.6.4",
```
kialam, line 32: Pinning `rrule` due to a known bug with v2.6.6 and higher: https://github.com/jakubroztocil/rrule/issues/427

nixocio (approved):

awx/ui/src/components/Schedule/shared/ScheduleForm.js
```diff
@@ -562,14 +563,25 @@ function ScheduleForm({
...
 
+            if (
+              end === 'onDate' &&
+              DateTime.fromISO(endDate)
+                .diff(DateTime.fromISO(startDate), 'days')
+                .toObject().days < 7
+            ) {
+              const rule = new RRule(buildRuleObj(values));
+              if (rule.all().length === 0) {
+                errors.startDate = t`Please select a valid start and end date/time.`;
```
keithjgrant: Can we make the error message a little more informative? Maybe something like "Selected frequency does not occur between start and end dates". Thoughts on this one, @trahman73 ?

keithjgrant: It would also be nice to add this one as a form-level error, rather than field-level on the startDate... though I don't remember the specifics of how we normally do that... ?

keithjgrant (approved):
nice work!

tiagodread (approved):
This is now showing the error message properly when there is no occurrence between start/end date avoiding the user to save the form:

Screenshot from 2021-08-20 14-29-11

**Note:** for users who saved the schedule previusly in this scenario will see an error on console until they update the schedule dates.

```
parseoptions.ts:19 Uncaught (in promise) Error: Invalid options: dtstart
```

PR #10904, opened 2021-08-17
do not collect artifact_data when gathering analytics

jladdjr:
`artifact_data` can include sensitive information that should be omitted


rebeccahhh (approved):
👍🏻

Issue #10829, opened 2021-08-03
Editing idle_timeout and job_timeout for ansible-runner environment settings


felipe4334:
So currently I'm having an issue were our windows patching playbook is failing because the working Kubernetes Pod gets terminated before the patching is completed. We believe this is because of the idle_timeout set inside ansible-runner environment settings

settings:
---
idle_timeout: 600
job_timeout: 3600
pexpect_timeout: 10

I'm currently building a custom AWX-EE image based out of quay.io/ansible/awx-ee:0.5.0, and that original image is being built from quay.io/ansible/ansible-runner:latest. I understand that I can build ansible/ansible-runner myself and edit the "ansible-runner/demo/env/settings" file and then build my own custom ansible/awx-ee:0.5.0 from that but is there a way to do this using awx-operator. I understand that awx-operator recommends customizing the Pod spec for custom EE but I dont see what specs I can provide the custom ansible/awx-ee:0.5.0 image since that image is not the one that gets idle_timeout: environment variable assigned.

![image](https://user-images.githubusercontent.com/13802820/128064041-3690072a-35d8-46db-a82a-7653e28dda53.png)

So to make the question above simpler, can I set the idle_timeout environment variable for ansible-runner inside the spec of a custom awx-ee image?

Or maybe include the idle_timeout ansible-runner envrionemnt variable when build the custom awx-ee image based out of ansible/awx-ee:0.5.0?

```
FROM quay.io/ansible/awx-ee:0.5.0

# Install Ansible requirements.
COPY requirements.yml ${HOME}/requirements.yml
RUN ansible-galaxy collection install -r ${HOME}/requirements.yml

# Version: 1.3.1
RUN pip3 install pandas

ENV idle_timeout=5000
```

wenottingham:
Moving to the awx repo. The right way for this to work would be for AWX to pass this down via some setting, not via overriden env vars in a podspec or EE.

tiagodread:
Verified setting `DEFAULT_JOB_IDLE_TIMEOUT` to 10s and running the playbook:
```
---

- name: 'Test playbook to sleep for a specified interval'
  hosts: all
  gather_facts: false
  vars:
    sleep_interval: 30

  tasks:
    - name: sleep for a specified interval
      command: sleep '{{ sleep_interval }}'
```

![image](https://user-images.githubusercontent.com/12464858/137484611-64760ec9-9c79-4229-a236-9bebaea92922.png)

felipe4334:
[...]

PR #10906, opened 2021-08-18
Added idle_timeout setting to job settings

x0r3ax3ax:
##### SUMMARY

#10829 
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API
 - UI

##### AWX VERSION

```
19.2.2
```

##### ADDITIONAL INFORMATION

```

```


nixocio (approved):
UI changes are sane. Thanks @oweel.

awx/main/conf.py
```diff
@@ -408,6 +408,18 @@
     unit=_('seconds'),
 )
 
+register(
+    'DEFAULT_JOB_IDLE_TIMEOUT',
+    field_class=fields.IntegerField,
+    min_value=0,
+    default=600,
```
jakemcdermott, line 415: Is this also the default value used by runner for `idle_timeout`?

If we add this setting using a `default` of `600`, will impact the existing behavior of preexisting job templates after upgrading?

jakemcdermott: Is this also the default value used by runner for idle_timeout?

Will the behavior of preexisting job templates be impacted after upgrading?

x0r3ax3ax: @jakemcdermott thank you for your feedback! 
Yes, you are right. I set default value is 0 for prevent overriding idle_timeout in preexisting job templates.

felipe4334:
@shanemcd It would be great if you could please review this pull request.

felipe4334:
Hello @shanemcd

felipe4334:
@nixocio Can this please be merged?

felipe4334:
@oweel it looks like idle_timeout might not be working too well.

Even though idle_timeout is set to 0 its still timing out in 5-6 minutes.
image

image

- pause:
      minutes: 20

If idle_timeout is set to something like 10 seconds then it works.
image
image


And if idle_timeout is set to like 9999 it will still timeout in 5-6 minutes:
image
image

Issue #10872, opened 2021-08-12
Approval Node Allows Saving without Required Name Field Value 


AlexSCorey:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

A workflow job template node has an active save button even with the name field being empty.  The User can save this node, and the modal will close and show a DELETED text on the node inside the visualizer

### AWX version

devel

### Installation method

docker for mac

### Modifications

no

### Web browser

Chrome

### Steps to reproduce

Create a workflow job template.
Go to the visualizer and create a node.  Select Approval node.
Leave name field blank and click save

### Expected results

The save button should be disabled without a name field.  

### Actual results

The save button is enabled and the user can save, and then closed with wizard

tiagodread:
Fixed with #10907

PR #10907, opened 2021-08-18
properly validates node

AlexSCorey:
##### SUMMARY
This addresses #10872.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
approval node

### DIFF
--- a/awx/ui/src/screens/Template/WorkflowJobTemplateVisualizer/Modals/NodeModals/NodeTypeStep/useNodeTypeStep.js
+++ b/awx/ui/src/screens/Template/WorkflowJobTemplateVisualizer/Modals/NodeModals/NodeTypeStep/useNodeTypeStep.js
@@ -61,7 +61,7 @@ function getStep(
       (nodeTypeField.value !== 'workflow_approval_template' &&
         nodeResourceField.value === null) ||
       (nodeTypeField.value === 'workflow_approval_template' &&
-        approvalNameField.value === undefined) ||
+        approvalNameField.value === '') ||
       formError
     ) {
       return false;


awx/ui/src/screens/Template/WorkflowJobTemplateVisualizer/Modals/NodeModals/NodeTypeStep/useNodeTypeStep.js
```diff
@@ -61,7 +61,7 @@ function getStep(
       (nodeTypeField.value !== 'workflow_approval_template' &&
         nodeResourceField.value === null) ||
       (nodeTypeField.value === 'workflow_approval_template' &&
-        approvalNameField.value === undefined) ||
+        approvalNameField.value === '') ||
```
jakemcdermott, line 64: @AlexSCorey That it was previously `undefined` makes me think there could be a scenario where the value _is_ `undefined` (This can happen on the survey form, if I recall). 

I'm wondering if we should add the empty string check to the existing conditional:
```suggestion
        (approvallNameField.value === 'undefined' || approvalNameField.value === '')) ||
```

jakemcdermott: discussed w/ @AlexSCorey - we're reasonably sure this value can never be `undefined`, so leaving as is.

jakemcdermott (approved):

PR #10910, opened 2021-08-18
Fix overwrite bug where hosts with no instance ID var are re-created

AlanCoding:
##### SUMMARY
Tests ran without this fix produce these errors:

https://gist.github.com/AlanCoding/81582dd60c727c35c5649ae0d6443779

Leaving as a draft until we have another independent reproducer.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/management/commands/inventory_import.py
+++ b/awx/main/management/commands/inventory_import.py
@@ -268,13 +268,13 @@ class Command(BaseCommand):
         """
         self.db_instance_id_map = {}
         if self.instance_id_var:
-            host_qs = self.inventory_source.hosts.all()
-            host_qs = host_qs.filter(instance_id='', variables__contains=self.instance_id_var.split('.')[0])
+            host_qs = self.inventory_source.hosts.all().only('id', 'instance_id', 'variables')
             for host in host_qs:
                 instance_id = self._get_instance_id(host.variables_dict)
-                if not instance_id:
-                    continue
-                self.db_instance_id_map[instance_id] = host.pk
+                if instance_id:
+                    self.db_instance_id_map[instance_id] = host.id
+                elif host.instance_id:
+                    self.db_instance_id_map[host.instance_id] = host.id
 
     def _build_mem_instance_id_map(self):
         """
--- a/awx/main/management/commands/inventory_import.py
+++ b/awx/main/management/commands/inventory_import.py
@@ -268,12 +268,12 @@ class Command(BaseCommand):
         """
         self.db_instance_id_map = {}
         if self.instance_id_var:
-            host_qs = self.inventory_source.hosts.all().only('id', 'instance_id', 'variables')
+            host_qs = self.inventory_source.hosts.all().only('id', 'instance_id', 'variables').iterator()
             for host in host_qs:
                 instance_id = self._get_instance_id(host.variables_dict)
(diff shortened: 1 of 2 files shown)


AlanCoding:
I feel pretty confident that I can demonstrate the fix with a real inventory source now.

 - Run an azure inventory sync that imports 3 hosts
 - Set overwrite on that source, update it again
   - Observe that it persists the host (no bug)
 - Now change the file-based setting `AZURE_RM_INSTANCE_ID_VAR` to something not there
 - Update it again

In the output, you can see:

```
    5.229 WARNING  Host "<redacted>" has no "alan_messed_with_diz_id" variable(s)
    5.273 WARNING  host deletions took 36 queries for 3 hosts
```

And in this case, it failed to recognize the host.

awx/main/management/commands/inventory_import.py
```diff
@@ -268,13 +268,18 @@ def _build_db_instance_id_map(self):
         """
         self.db_instance_id_map = {}
         if self.instance_id_var:
-            host_qs = self.inventory_source.hosts.all()
-            host_qs = host_qs.filter(instance_id='', variables__contains=self.instance_id_var.split('.')[0])
+            host_qs = self.inventory_source.hosts.all().only('id', 'instance_id', 'variables').iterator()
```
AlanCoding: I think we need an alternative to this for performance reasons.

To do that, I don't see easy option other than to do a loop of `instance_id_var.split(',')` (probably reversed), and then `.split('.')[0]` inside that loop.

awx/main/management/commands/inventory_import.py
```diff
@@ -269,12 +270,13 @@ def _build_db_instance_id_map(self):
         self.db_instance_id_map = {}
         if self.instance_id_var:
             host_qs = self.inventory_source.hosts.all()
-            host_qs = host_qs.filter(instance_id='', variables__contains=self.instance_id_var.split('.')[0])
-            for host in host_qs:
-                instance_id = self._get_instance_id(host.variables_dict)
-                if not instance_id:
-                    continue
-                self.db_instance_id_map[instance_id] = host.pk
+            for instance_id_part in reversed(self.instance_id_var.split(',')):
```
jladdjr, line 273: I'm still reading through everything here, so don't have the full context yet, but had two questions come up here (mostly just for my own edification):

* Based on `self.instance_id_var.split(',')` it looks like we just weren't respecting the fact that `self.instance_id_var` could be a _list_ of potential id vars?
* Why `reversed` here? Does the last id var have highest precedence?

jladdjr (approved):
Man, tricky stuff. Thanks for dealing with such a thorny issue.

awx/main/management/commands/inventory_import.py
AlanCoding: First point - exactly, yes.

Second point - the _first_ id has the highest precedence. The loop isn't checking if the dict has the key before assigning it. So if both id vars match, it'll save the 2nd value first, but then get overwritten by the 1st.

jladdjr: ah, makes sense. 👍

beeankha (approved):

PR #10918, opened 2021-08-20
Attempt to surface streaming errors that were being eaten

shanemcd:


kdelee:
I manually applied this diff and now I'm getting better error messages:
https://github.com/ansible/tower/issues/5251#issuecomment-902840339

kdelee (approved):

awx/main/tasks.py
```diff
@@ -3077,7 +3081,12 @@ def _run_internal(self, receptor_ctl, work_type=None):
                     return res
 
                 if not self.task.instance.result_traceback:
-                    raise RuntimeError(detail)
+                    try:
+                        resultsock = receptor_ctl.get_work_results(self.unit_id, return_sockfile=True)
+                        lines = resultsock.readlines()
+                        self.task.instance.result_traceback = b"".join(lines).decode()
```
AlanCoding, line 3084: I'm somewhat confused by the fact that this doesn't raise (which would save the text to `result_traceback` ultimately) and doesn't call `self.task.save(update_fields=['result_traceback'])`. Not saying it doesn't work, it could, but it's not easy to follow.

AlanCoding:
I'm also increasingly feeling like this data shouldn't be considered a traceback. Yes, it might contain a python traceback, but it's just a command that errored. It could be anything.

AlanCoding (approved):

AlanCoding:
Also on the subject of error handling - https://github.com/ansible/awx/pull/10823

This is really difficult stuff, I'm worried about it getting out of control, and I'd really like to draw down the complexity in any way I can.

beeankha (approved):

PR #10919, opened 2021-08-20
awx aint the only user in town

kdelee:
##### SUMMARY

Building off of https://github.com/ansible/awx/pull/10918 , I wan to see if this makes me able to run a project update

### DIFF
--- a/awx/main/models/credential/__init__.py
+++ b/awx/main/models/credential/__init__.py
@@ -506,7 +506,7 @@ class CredentialType(CommonModelNameNotUnique):
             _, path = tempfile.mkstemp(dir=os.path.join(private_data_dir, 'env'))
             with open(path, 'w') as f:
                 f.write(data)
-            os.chmod(path, stat.S_IRUSR | stat.S_IWUSR)
+            os.chmod(path, stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IWGRP)
             container_path = to_container_path(path, private_data_dir)
 
             # determine if filename indicates single file or many
@@ -539,7 +539,7 @@ class CredentialType(CommonModelNameNotUnique):
                 f = os.fdopen(handle, 'w')
                 f.write(safe_dump(vars))
                 f.close()
-                os.chmod(path, stat.S_IRUSR)
+                os.chmod(path, stat.S_IRUSR | stat.S_IRGRP)
                 return path
 
             if extra_vars:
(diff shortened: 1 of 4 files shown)


awx/main/models/credential/injectors.py
```diff
@@ -31,7 +31,7 @@ def gce(cred, env, private_data_dir):
     f = os.fdopen(handle, 'w')
     json.dump(json_cred, f, indent=2)
     f.close()
-    os.chmod(path, stat.S_IRUSR | stat.S_IWUSR)
+    os.chmod(path, stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IWGRP)
```
AlanCoding, line 34: Seeing this repeated so many times, it makes it seem obvious that it should become a shared constant, and an easy location for this is `awx.main.constants`. We might have multiple variety of file permissions for executable files, and some others.

awx/main/models/credential/__init__.py
```diff
@@ -539,7 +539,7 @@ def build_extra_vars_file(vars, private_dir):
                 f = os.fdopen(handle, 'w')
                 f.write(safe_dump(vars))
                 f.close()
-                os.chmod(path, stat.S_IRUSR)
+                os.chmod(path, stat.S_IRUSR | stat.S_IRGRP)
```
AlanCoding, line 542: I cannot articulate why the extra vars file should not be writable when the others are.

awx/main/models/credential/injectors.py
kdelee: +1

awx/main/tasks.py
```diff
@@ -966,6 +967,10 @@ def build_private_data_files(self, instance, private_data_dir):
         private_data = self.build_private_data(instance, private_data_dir)
         private_data_files = {'credentials': {}}
         if private_data is not None:
+            artifact_dir = os.path.join(private_data_dir, 'artifacts', str(self.instance.id))
+            if not os.path.exists(artifact_dir):
+                os.makedirs(artifact_dir)
+                os.chmod(artifact_dir, stat.S_IRUSR | stat.S_IXUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IXGRP | stat.S_IWGRP)
```
AlanCoding, line 973: Why did you have to move artifact dir handling and change the pattern for it?

kdelee:
Abandoning this approach

PR #10926, opened 2021-08-23
Corresponding AWX changes for runner --worker-info schema update

AlanCoding:
Update for https://github.com/ansible/ansible-runner/pull/802

### DIFF
--- a/awx/main/models/ha.py
+++ b/awx/main/models/ha.py
@@ -3,6 +3,7 @@
 
 from decimal import Decimal
 import random
+import logging
 
 from django.core.validators import MinValueValidator
 from django.db import models, connection
@@ -25,6 +26,8 @@ from awx.main.utils.common import measure_cpu, get_corrected_cpu, get_cpu_effect
 from awx.main.models.mixins import RelatedJobsMixin
 
 __all__ = ('Instance', 'InstanceGroup', 'TowerScheduleState')
+
+logger = logging.getLogger('awx.main.models.ha')
 
 
 class HasPolicyEditsMixin(HasEditsMixin):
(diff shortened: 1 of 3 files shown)


awx/main/utils/receptor.py
```diff
@@ -71,7 +71,12 @@ def worker_info(node_name):
...
             error_list.append(f'Remote node {node_name} --worker-info output is not a YAML dict, output:{stdout}')
         else:
-            error_list.extend(remote_data.pop('Errors'))  # merge both error lists
+            error_list.extend(remote_data.pop('errors', []))  # merge both error lists
             data.update(remote_data)
 
+    # see tasks.py usage of keys
+    missing_keys = set(('runner_version', 'mem_in_bytes', 'cpu_count')) - set(data.keys())
+    if missing_keys:
+        data['errors'].append('Worker failed to return keys {}'.format(' '.join(missing_keys)))
```
AlanCoding, line 80: This was an error case we were missing before - if the remote executor gives no relevant keys. We mark it lost if there are errors, but there was no "errors" key returned, so it was saved as a working instance with a memory capacity of 1. That's not good, but is resolved here to my satisfaction.

beeankha (approved):

awx/main/models/ha.py
```diff
@@ -174,13 +177,19 @@ def refresh_capacity_fields(self):
...
+    def save_health_data(self, version, cpu, memory, uuid=None, last_seen=None, has_error=False):
         update_fields = []
 
         if last_seen is not None and self.last_seen != last_seen:
             self.last_seen = last_seen
             update_fields.append('last_seen')
 
+        if uuid is not None and self.uuid != uuid:
+            if self.uuid is not None:
+                logger.warn(f'Self-reported uuid of {self.hostname} changed from {self.uuid} to {uuid}')
```
jladdjr, line 189: Know this isn't the heart of the PR, but love that you thought to surface this info in the logger 👍 . Can see this kind of thing (uuid-change) being responsible for some kind of esoteric bug down the line.

jladdjr (approved):

Issue #10853, opened 2021-08-10
Disallow delete or edit of Control Instances


AlexSCorey:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
In 4.1 the Mesh code POD will introduce different types of instances to include Controller, Executioner, and Hybrid( which will do both).  Control nodes are automatically added to the controlplane instance group when they come online.  We should not allow them to be deleted or edited. 

We will also need API validation.

jbradberry:
https://github.com/ansible/awx/pull/10929 has merged.

CFSNM:
Opened PR to add automated API tests for this feature: https://github.com/ansible/tower-qa/pull/6800

CFSNM:
This issue has been tested from the API perspective. Need to test from the UI perspective as well. FYI @tiagodread @akus062381 @InnocentK

akus062381:
resolved

CFSNM:
Thanks @akus062381

Issue #10830, opened 2021-08-06
Prevent control-only instances from being added to user-defined instance groups.


AlanCoding:
Users can now add nodes to the cluster that are not allowed to run jobs (they can still run project updates and some other things).

These instances should be added to the `controlplane` instance group, but they should not be allowed to be added to any other instance group.

---

Jotting down some acceptance criteria:

- control and hybrid instances should automatically be added to `controlplane` IG (this should happen during awx installation)
- user attempts to add/remove instance from `controlplane` IG => 400
- user attempts to _delete_ hybrid / control instance altogether => 400
- (in general, seems like `controlplane` IG's membership should be immutable after install - do we want to go so far as to disable POST/PUT/DELETE for instancegroups/<control plane IG>/instances endpoint? seems like maybe we don't want to go quite this far)
- deleting an execution node is maybe allowed?

tchellomello:
PR https://github.com/ansible/automation-platform-collection/pull/128 should cover this

AlanCoding:
@tchellomello instances can _also_ be added to instance groups through the API. If that is attempted, we should give a 400 error, and that's the primary thing we'd like to accomplish with this issue.

tchellomello:
@AlanCoding alright, I will mention this for the team tomorrow so we can add that as well to this PR. Thanks for the heads up

AlanCoding:
I intended this to be done by the mesh code pod, because it's on the server code side of things. You already did the installer part.

But if anyone wants to get this knocked out, go ahead. It would be several weeks out for us because it's not urgent.

AlanCoding:
https://github.com/ansible/awx/pull/10985 will prevent them from being added by the policy rules. Once @jbradberry merges we should be done with this.

jbradberry:
https://github.com/ansible/awx/pull/10929 has merged.

jladdjr:
@thenets let us know if you need any additional context on this for testing.

tiagodread:
Verified that:
1- User cannot change instance group membership of control-only type:
```
{
    "msg": "Cannot change instance group membership of control-only node: ec2-34-239-101-88.compute-1.amazonaws.com."
}
```
although the user can at lest try to do so via UI

https://user-images.githubusercontent.com/12464858/137398251-22e7836f-25e9-416f-addd-94d37ef5e1b9.mp4


2- Hybrid and control-only are added to controlplane automatically
3- Users are not allowed to remove hybrid/control instances

[...]

Issue #10841, opened 2021-08-09
System Auditors can check list items


AlexSCorey:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

On most, if not all, lists a system auditor can click the check box on the list items.  The delete button is still disabled so they can't really do anything with the button but it would be better UX to disabled the individual check boxes as well.

### AWX version

devel

### Installation method

docker for mac

### Modifications

no

### Steps to reproduce

Create a system auditor and sign in as that user. Go to any list.  

### Expected results

Check boxes on the tool bar, and the list items should be disabled.

### Actual results

List item check boxes are not disabled, and in some cases the toolbar check box  (select all) is not disabled.

BhaveshAnandpara:
Hey @AlexSCorey I am interested in working on this issue is this still open?

AlexSCorey:
@BhaveshAnandpara yes. Go for it

BhaveshAnandpara:
@AlexSCorey I need a bit of a context on what is the problem so if there is any image ??

PR #10927, opened 2021-08-23
Disable checkbox for instances with node type control.

kialam:
##### SUMMARY

Related #10853
Related #10830

Requires: https://github.com/ansible/awx/pull/10929

On the Instances Group > Instances tab, instances with a `node_type` of `control` should not be editable, deleted, disassociated, associated, etc. The API will validate this on their end. For the UI, we have now disabled the checkbox next to the instance to disableany user actions with control type nodes.

Screen Shot 2021-08-23 at 1 56 19 PM

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### AWX VERSION

```
19.3.0
```

### DIFF
--- a/awx/ui/src/screens/InstanceGroup/Instances/InstanceListItem.js
+++ b/awx/ui/src/screens/InstanceGroup/Instances/InstanceListItem.js
@@ -106,7 +106,7 @@ function InstanceListItem({
         <Td
           select={{
             rowIndex,
-            isSelected,
+            isSelected: isSelected && instance.node_type !== 'control',
             onSelect,
             disable: instance.node_type === 'control',
           }}
--- a/awx/ui/src/screens/InstanceGroup/Instances/InstanceListItem.js
+++ b/awx/ui/src/screens/InstanceGroup/Instances/InstanceListItem.js
@@ -108,7 +108,7 @@ function InstanceListItem({
             rowIndex,
             isSelected,
             onSelect,
-            disable: false,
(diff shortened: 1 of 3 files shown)


awx/ui/src/screens/InstanceGroup/Instances/InstanceListItem.js
```diff
@@ -108,7 +108,7 @@ function InstanceListItem({
             rowIndex,
             isSelected,
             onSelect,
-            disable: false,
+            disable: instance.node_type === 'control',
```
jbradberry, line 111: I take it that this is the list of instances that are members of a particular instance group?  Do we have an equivalent screen file for the list of instance groups that an instance is a member of?  In doing the back-end code, I realized that both needed to be dealt with.

nixocio:
@kialam as we talked we will also need to verify the proper UX for the select all for disabled checkboxes.

AlexSCorey (commented):
This does look like it disables the checkbox for `node_type === 'control'`. We should also probably disabled the select all button if all instances are `node_type === 'control'`.  Also check out [this issue](https://github.com/ansible/awx/issues/10841) to see if it can also be addressed during this work.

nixocio (approved):
LGTM. I would suggest to merge the API changes first then we can better test those changes.

jbradberry:
Ok, https://github.com/ansible/awx/pull/10929 has merged.

PR #10930, opened 2021-08-24
Clean up added work_type processing for mesh_code branch

AlanCoding:
This undos some more changes from the original PoC that are no longer relevant.

still testing.

### DIFF
--- a/awx/main/managers.py
+++ b/awx/main/managers.py
@@ -10,6 +10,7 @@ from django.conf import settings
 
 from awx.main.utils.filters import SmartFilter
 from awx.main.utils.pglock import advisory_lock
+from awx.main.utils.common import get_capacity_type
 from awx.main.constants import RECEPTOR_PENDING
 
 ___all__ = ['HostManager', 'InstanceManager', 'InstanceGroupManager', 'DeferJobCreatedManager']
@@ -207,6 +208,8 @@ class InstanceGroupManager(models.Manager):
         if name not in graph:
             graph[name] = {}
         graph[name]['consumed_capacity'] = 0
+        for capacity_type in ('execution', 'control'):
+            graph[name][f'consumed_{capacity_type}_capacity'] = 0
         if breakdown:
             graph[name]['committed_capacity'] = 0
             graph[name]['running_capacity'] = 0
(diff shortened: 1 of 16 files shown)


awx/main/managers.py
```diff
@@ -160,7 +160,10 @@ def get_or_register(self):
             from awx.main.management.commands.register_queue import RegisterQueue
 
             pod_ip = os.environ.get('MY_POD_IP')
-            registered = self.register(ip_address=pod_ip)
+            if settings.IS_K8S:
+                registered = self.register(ip_address=pod_ip, node_type='control')
+            else:
+                registered = self.register(ip_address=pod_ip)
```
AlanCoding, line 167: @shanemcd this split may not be necessary if `AWX_AUTO_DEPROVISION_INSTANCES` is 100% synonymous with `IS_K8S`... and maybe it is, but I don't have a firm understanding of all of these knobs.

shanemcd: I think it technically is at this point, but I can see a scenario in the future where we want to use `AWX_AUTO_DEPROVISION_INSTANCES` in VM-land.

beeankha (approved):
:+1:

awx/main/scheduler/task_manager.py
```diff
@@ -87,7 +87,21 @@ def after_lock_init(self):
         instances_by_hostname = {i.hostname: i for i in instances_partial}
 
         for rampart_group in InstanceGroup.objects.prefetch_related('instances'):
-            self.graph[rampart_group.name] = dict(graph=DependencyGraph(), capacity_total=rampart_group.execution_capacity, consumed_capacity=0, instances=[])
+            self.graph[rampart_group.name] = dict(
+                graph=DependencyGraph(),
+                execution_capacity=0,
+                control_capacity=0,
+                consumed_capacity=0,
```
AlanCoding, line 94: We don't count total available capacity, but we do count total consumed capacity. This is only used for writing log messages, which I don't want to muck with.

In the API, we do report total capacity, which is also something I don't want to mess with, because it should be the same as execution capacity in all cases except for the control plane group.

PR #10932, opened 2021-08-24
Obtain receptor sockfile from the receptor.conf file

AlanCoding:
This should allow installers to change the location of the receptor sockfile without having to make other additional changes.


tchellomello (approved):
Looks good to me @AlanCoding Thanks!!

shanemcd (approved):

beeankha (approved):

Issue #10925, opened 2021-08-23
Wrong Delete Warning Modal


AlexSCorey:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

Wrong delete warning modal when removing role from `/organizations/id/access`.  

### AWX version

devel

### Installation method

docker for mac

### Modifications

no

### Web browser

Chrome

### Steps to reproduce

Create an organization.  Click on the access tab within that organization.  Remove 1 of the roles for one of the members of the org.  

### Expected results

Some confirmation modal to pop up that does not refer to `The Team`.

### Actual results

The modal text references the team and asks the user to remove the permission in `/teams`.  Also, the first sentence seems incomplete.

### Additional information

![Screen Shot 2021-08-23 at 1 23 24 PM](https://user-images.githubusercontent.com/39280967/130490932-bb2832fd-312a-4bca-9456-45ee1b22065c.png)

PR #10934, opened 2021-08-24
Fixes issue where the wrong text appeared in modal

AlexSCorey:
##### SUMMARY
Addresses #10925 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION

ModalText

### DIFF
--- a/awx/ui/src/components/ResourceAccessList/DeleteRoleConfirmationModal.js
+++ b/awx/ui/src/components/ResourceAccessList/DeleteRoleConfirmationModal.js
@@ -8,10 +8,9 @@ import { Role } from 'types';
 import AlertModal from '../AlertModal';
 
 function DeleteRoleConfirmationModal({ role, username, onCancel, onConfirm }) {
-  const isTeamRole = () =>
+  const sourceOfRole = () =>
     typeof role.team_id !== 'undefined' ? t`Team` : t`User`;
-
-  const title = t`Remove ${isTeamRole()} Access`;
+  const title = t`Remove ${sourceOfRole()} Access`;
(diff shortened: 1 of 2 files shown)


awx/ui/src/components/ResourceAccessList/DeleteRoleConfirmationModal.js
```diff
@@ -38,7 +37,7 @@ function DeleteRoleConfirmationModal({ role, username, onCancel, onConfirm }) {
         </Button>,
       ]}
     >
-      {isTeamRole() ? (
+      {isTeamRole() === 'Team' ? (
```
jakemcdermott: I think we should rename this function since it returns a string. By convention, we typically use `is` for boolean variable names and (rarely) functions that return booleans. The line being changed even seems like it could have been a usage error from expecting the function to work this way, since the result of the function call was being used to conditionally render even though the result is always a (truthy) string.

awx/ui/src/components/ResourceAccessList/DeleteRoleConfirmationModal.test.js
```diff
@@ -14,7 +14,7 @@ const role = {
 };
 
 describe('<DeleteRoleConfirmationModal />', () => {
-  test('should render initially', () => {
+  test('shold render Team confirmation modal', () => {
```
jakemcdermott:
```suggestion
  test('should render Team confirmation modal', () => {
```

jakemcdermott (approved):

PR #10935, opened 2021-08-24
Remove sleep from tests

jakemcdermott:
##### SUMMARY
- Remove sleeps

### DIFF
--- a/awx/ui/src/components/LaunchButton/LaunchButton.test.js
+++ b/awx/ui/src/components/LaunchButton/LaunchButton.test.js
@@ -10,7 +10,6 @@ import {
   WorkflowJobTemplatesAPI,
 } from 'api';
 import { mountWithContexts } from '../../../testUtils/enzymeHelpers';
-import { sleep } from '../../../testUtils/testUtils';
 
 import LaunchButton from './LaunchButton';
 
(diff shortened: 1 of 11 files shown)


awx/ui/src/components/LaunchButton/LaunchButton.test.js
```diff
@@ -128,7 +125,7 @@ describe('LaunchButton', () => {
     WorkflowJobTemplatesAPI.launch.mockImplementation(async () => {
       // return asynchronously so isLaunching isn't set back to false in the
       // same tick
-      await sleep(10);
+      await new Promise((resolve) => setTimeout(resolve, 10));
```
jakemcdermott, line 128: Unsure if this is still needed.

AlexSCorey: The test passes locally without `new Promise` and `sleep`.  However, the test, which is supposed to check if the button is disabled, never asserts that `disabled === true`, so I'm not really sure what this is actually testing.

AlexSCorey (changes requested):
This all seem good to go except for that 1 line in question in the `LaunchButton` test.  If in fact it is not actually testing what it says it is we should either remove it (not my first choice) or update it.

awx/ui/src/components/LaunchButton/LaunchButton.test.js
jakemcdermott: Good find @AlexSCorey

AlexSCorey (approved):
Looks good

jakemcdermott:
@AlexSCorey - Looks like the commit to address the launch button test resulted in some new test failures. I agree that we should update that test, but since the issue wasn't introduced by this PR I'm going to leave that commit out for now and we can follow up on updating the launch button test with a separate issue: https://github.com/ansible/awx/issues/11433

Issue #10746, opened 2021-07-23
Add auto-discovered instances to a (to-be-created) "default" instance group.


AlanCoding:
The `default` instance group currently only exists on OCP installs.

We will need to add logic to create this `default` instance group on other install types as well.

Any newly discovered execution nodes in the mesh should go into the default group, disabled. (NOTE: the `disabled` state may still need more discussion) There could maybe be a setting that auto-enables the instances by default.

The process of enabling and putting into another group can be a day 2 operation.

AlanCoding:
Putting on the "small" label, for a near-term focus on these objectives:

 - enabling the "default" group for all install types
 - adding auto-provisioned nodes to this group
 - probably mention this in `docs/receptor_mesh.md`

Discussions and revisions can still happen after that, but these items are clear and relatively easy.

AlanCoding:
I had a mis-understanding. The installer _will_ run `awx-manage provision_instance` for execution nodes listed in the inventory file. This command can be responsible for making the nodes enabled if they were specified by the user. So add to the work items:

 - Make auto-discovered nodes disabled by default

AlanCoding:
We should have a meeting with people in the other pod and a few high-level people to establish expectations about group memberships. Where memberships are set, what expectations of the groups are, and so on. There is a lot more work than what is outlined here that we need to do.

For example, we need to stop using the `controlplane` as a fallback for jobs.

AlexSCorey:
Related: https://github.com/ansible/awx/issues/10853

jladdjr:
> The default instance group currently only exists on OCP installs.
> We will need to add logic to create this default instance group on other install types as well.

Just wanted to note that I confirmed with @tchellomello yesterday that the `default` IG is created in trad installs as well.

jladdjr:
@AlanCoding I _think_ https://github.com/ansible/awx/pull/10937 does what we want. Need to add some tests for this, but ad hoc testing confirmed the following changes:

- [x] `default` instance group created
- [x] auto-discovered nodes marked as disabled by default
- [x] .. and assigned to `default` instance group (by way of `policy_instance_list` )

tchellomello:
> > The default instance group currently only exists on OCP installs.
> > We will need to add logic to create this default instance group on other install types as well.
> 
> Just wanted to note that I confirmed with @tchellomello yesterday that the `default` IG is created in trad installs as well.

[...]

PR #10937, opened 2021-08-25
Add auto discovered instances to default ig (also, mark as disabled)

jladdjr:
see https://github.com/ansible/awx/issues/10746

---

Need to add some tests for this, but ad hoc testing confirmed the following changes:

- [x] `default` instance group created
- [x] auto-discovered nodes marked as disabled by default
- [x] .. and assigned to `default` instance group (by way of `policy_instance_list` )

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,6 +174,7 @@ init:
 	fi; \
 	$(MANAGEMENT_COMMAND) provision_instance --hostname=$(COMPOSE_HOST); \
 	$(MANAGEMENT_COMMAND) register_queue --queuename=controlplane --instance_percent=100;\
+	$(MANAGEMENT_COMMAND) register_queue --queuename=default;
 	if [ ! -f /etc/receptor/certs/awx.key ]; then \
 		rm -f /etc/receptor/certs/*; \
 		receptor --cert-init commonname="AWX Test CA" bits=2048 outcert=/etc/receptor/certs/ca.crt outkey=/etc/receptor/certs/ca.key; \
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -429,47 +429,57 @@ def execution_node_health_check(node):
 
 
 def inspect_execution_nodes(instance_list):
-    node_lookup = {}
-    for inst in instance_list:
-        if inst.node_type == 'execution':
-            node_lookup[inst.hostname] = inst
+    with advisory_lock('inspect_execution_nodes_lock', wait=True):
+        node_lookup = {}
+        for inst in instance_list:
+            if inst.node_type == 'execution':
+                node_lookup[inst.hostname] = inst
 
-    ctl = get_receptor_ctl()
-    connections = ctl.simple_command('status')['Advertisements']
-    nowtime = now()
-    for ad in connections:
-        hostname = ad['NodeID']
-        commands = ad.get('WorkCommands') or []
-        if 'ansible-runner' not in commands:
-            continue
-        changed = False
-        if hostname in node_lookup:
-            instance = node_lookup[hostname]
-        else:
-            defaults = dict(enabled=False)
-            (changed, instance) = Instance.objects.register(hostname=hostname, node_type='execution', defaults=defaults)
-        was_lost = instance.is_lost(ref_time=nowtime)
-        last_seen = parse_date(ad['Time'])
+        ctl = get_receptor_ctl()
+        connections = ctl.simple_command('status')['Advertisements']
+        nowtime = now()
+        for ad in connections:
+            hostname = ad['NodeID']
+            commands = ad.get('WorkCommands') or []
+            if 'ansible-runner' not in commands:
+                continue
+            changed = False
+            if hostname in node_lookup:
+                instance = node_lookup[hostname]
+            else:
+                defaults = dict(enabled=False)
+                (changed, instance) = Instance.objects.register(hostname=hostname, node_type='execution', defaults=defaults)
+                logger.warn("Registered execution node '{}' (marked as disabled by default)".format(hostname))
 
-        if instance.last_seen and instance.last_seen >= last_seen:
-            continue
-        instance.last_seen = last_seen
-        instance.save(update_fields=['last_seen'])
+                default_ig = InstanceGroup.objects.get(name='default')
+                if instance.hostname not in default_ig.policy_instance_list:
+                    default_ig.policy_instance_list += [instance.hostname]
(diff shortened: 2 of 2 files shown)


awx/main/tasks.py
```diff
@@ -429,46 +429,57 @@ def execution_node_health_check(node):
 
 
 def inspect_execution_nodes(instance_list):
-    node_lookup = {}
-    for inst in instance_list:
-        if inst.node_type == 'execution':
-            node_lookup[inst.hostname] = inst
+    with advisory_lock('inspect_execution_nodes_lock', wait=True):
```
jladdjr: Added lock to avoid instance group policy change race conditions. I made the scope of the lock much narrower initially, then figured that really the whole operation here should probably be atomic. I went back and forth on this, though, and could see a good case for going either way.

awx/main/tasks.py
```diff
@@ -429,46 +429,57 @@ def execution_node_health_check(node):
...
+        for ad in connections:
+            hostname = ad['NodeID']
+            commands = ad.get('WorkCommands') or []
+            if 'ansible-runner' not in commands:
+                continue
+            changed = False
+            if hostname in node_lookup:
+                instance = node_lookup[hostname]
+            else:
+                defaults = dict(enabled=False)
```
jladdjr, line 450: ignoring the rest of the diff (which is due to the indentation change), this is where I actually made some tweaks:
* disable the instance by default
* .. and add the instance to the `default` IG by way of the `policy_instance_list` (after fishing around a bunch, I determined this seemed like the 'right' way to assign the instance to the group -- tell the IG what we want, and let our `apply_cluster_membership_policies()` task be responsible for making it so.)

Makefile
```diff
@@ -174,6 +174,7 @@ init:
 	fi; \
 	$(MANAGEMENT_COMMAND) provision_instance --hostname=$(COMPOSE_HOST); \
 	$(MANAGEMENT_COMMAND) register_queue --queuename=controlplane --instance_percent=100;\
+	$(MANAGEMENT_COMMAND) register_queue --queuename=default;
```
jladdjr, line 177: Creating instance groups for the dev env

awx/main/tasks.py
AlanCoding: I'm not against using an advisory lock here, it is, after all, an action global to all execution nodes. But I am very strongly in disagreement with `wait=True`, because as we scale up clusters, more and more time will be spent with a worker tied up running this task. More dangerous, it delays the further execution of the cluster_node_heartbeat, making data in the `instance_list` data structure more and more outdated.

We only _want_ this to run once per period _for the entire cluster_, so `wait=False` would have no downside.

The lock is also a little distracting. The cluster coordination problem here is _mainly_ solved by using the `last_seen` timestamp, which returns the same value throughout the receptor mesh, and updates minutely. That is how we prevent running health checks more than once per minute, per execution node.

awx/main/tasks.py
```diff
@@ -429,46 +429,57 @@ def execution_node_health_check(node):
...
+                (changed, instance) = Instance.objects.register(hostname=hostname, node_type='execution', defaults=defaults)
+                logger.warn("Registered execution node '{}' (marked as disabled by default)".format(hostname))
+
+                default_ig = InstanceGroup.objects.get(name='default')
+                if instance.hostname not in default_ig.policy_instance_list:
+                    default_ig.policy_instance_list += [instance.hostname]
+                    default_ig.save()
```
AlanCoding: would be good to use `update_fields=['policy_instance_list']` here

awx/main/tasks.py
```diff
@@ -429,46 +429,57 @@ def execution_node_health_check(node):
...
+                continue
+            changed = False
+            if hostname in node_lookup:
+                instance = node_lookup[hostname]
+            else:
+                defaults = dict(enabled=False)
+                (changed, instance) = Instance.objects.register(hostname=hostname, node_type='execution', defaults=defaults)
+                logger.warn("Registered execution node '{}' (marked as disabled by default)".format(hostname))
+
+                default_ig = InstanceGroup.objects.get(name='default')
```
AlanCoding: if this doesn't exist, we would like to catch the error. An uncaught error has bad consequences.

AlanCoding: this should also only happened if `changed` is True.

awx/main/tasks.py
jladdjr: `wait=False` totally makes sense, ty for that catch.

I'm having a harder time seeing what you mentioned re: `last_seen` as helping with the coordination problem. That comes into play after the `default_ig` is retrieved and its `policy_instance_list` is updated, one of the key things that seemed like it needed to be atomic (otherwise, there's a risk two nodes could go to update this same field and one of the updates would win out over the other one).

awx/main/tasks.py
```diff
@@ -429,46 +429,60 @@ def execution_node_health_check(node):
...
-            logger.warn(f'Execution node attempting to rejoin as instance {hostname}.')
-            execution_node_health_check.apply_async([hostname])
-        elif instance.capacity == 0:
-            # Periodically re-run the health check of errored nodes, in case someone fixed it
-            # TODO: perhaps decrease the frequency of these checks
-            logger.debug(f'Restarting health check for execution node {hostname} with known errors.')
-            execution_node_health_check.apply_async([hostname])
+            if changed:
+                try:
+                    default_ig = InstanceGroup.objects.get(name='default')
```
AlanCoding: there's a setting for this that should be preferred over the string `'default'`

awx/main/tasks.py
AlanCoding: > otherwise, there's a risk two nodes could go to update this same field and one of the updates would win out over the other one

Yeah, the lock will make this atomic, but we also need to realize that not all things _need_ to be atomic.

The last_seen field comes from receptor, and all control/hybrid nodes see the same value until the node checks in again (once per minute). They will not update the field unless they get a timestamp greater than the current value. In clustered situations, what happens is that they commonly compare the same value and do nothing because of that. There's never a time when 1 node would be updating with a value that contradicts what another node is trying to update with.

That _can_ be an issue for appending onto the `instance_policy_list` field. However, for that, there can be conflicts with other code. For that, the only true solution is `select_for_update`, which we do in other cases for that field. It's rare that we need extreme measures like that, but appending onto a list carries a very high risk that parallel updates may _erase_ an update from another process.

AlanCoding (approved):

PR #10949, opened 2021-08-26
Always use controlplane as project update backup IG

AlanCoding:
To recap the scenario:

 - no IG in project update preferred instance groups have any "control" type capacity (they are all execution-only nodes)
 - project update gets stuck in the waiting state forever

These execution-only instance groups equate fairly well to the existing container groups. In both cases, the IG can run a job, but not run a project update. We do this check:

```python
not any([not group.is_container_group for group in selected_groups])
```

If you use boolean algebra on that, I think it's the same thing as saying "if any groups in selected_groups are container groups".

At this point, I think it's easier / simpler to just say that control plane should always be a backup for project updates.

### DIFF
--- a/awx/main/models/projects.py
+++ b/awx/main/models/projects.py
@@ -628,7 +628,7 @@ class ProjectUpdate(UnifiedJob, ProjectOptions, JobNotificationMixin, TaskManage
         selected_groups = template_groups + organization_groups
 
         controlplane_ig = self.control_plane_instance_group
-        if controlplane_ig[0] not in selected_groups:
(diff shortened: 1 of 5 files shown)


AlanCoding:
Looks like we need to pull in a controlplane instance fixture for a few tests, since we are already effectively assuming that it exists.

beeankha (approved):

AlanCoding:
Surprisingly, this didn't fail any of the integration tests. So after unit tests are passing, I might look at merging.

AlanCoding:
I filed a docs ticket for @tvo318 

I will go back over the developer docs, but it may or may not be changed by this.

PR #10954, opened 2021-08-26
Use ansible-runner imports for cpu and memory calculation

AlanCoding:
We delayed this to make sure the images were updated, but now it seems there are no blockers for doing it.

### DIFF
--- a/awx/main/utils/common.py
+++ b/awx/main/utils/common.py
@@ -710,6 +710,14 @@ def parse_yaml_or_json(vars_str, silent_failure=True):
 def get_cpu_effective_capacity(cpu_count):
     from django.conf import settings
 
+    settings_abscpu = getattr(settings, 'SYSTEM_TASK_ABS_CPU', None)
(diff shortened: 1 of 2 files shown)


awx/main/utils/common.py
```diff
@@ -761,20 +770,14 @@ def get_mem_effective_capacity(mem_mb):
...
 def get_corrected_memory(memory):
     from django.conf import settings
 
     settings_absmem = getattr(settings, 'SYSTEM_TASK_ABS_MEM', None)
     env_absmem = os.getenv('SYSTEM_TASK_ABS_MEM', None)
 
-    if env_absmem is not None:
-        return 0, int(env_absmem)
-    elif settings_absmem is not None:
-        return 0, int(settings_absmem)
```
AlanCoding, line 777: I'm going to be honest here, this was a mistake I made. You can clearly see it returns a tuple, and then a single value later. The intent is accurately expressed in the name of the method, but the refactor wasn't followed through on completely.

beeankha (approved):
:+1:

Issue #10557, opened 2021-06-30
Notification Add/Edit Form Validation Issues


akus062381:
The form to add a notification has the field validation on the Recipient List field, but the edit form does not have field validation for that same field (and it is required).

This same error happens for the Destination Channels or Users field under the IRC Notification type.
This same error happens for the Destination Channels field under the Slack Notification type.

Type Mattermost- field Target URL- does not validate whether the email address entered is in email format.
Type Twilio- Source phone number field- does not validate whether a phone number is entered- it accepts alphabet characters.

![notification](https://user-images.githubusercontent.com/63685497/123979836-0efdd900-d98f-11eb-93ed-b43deed330e7.gif)

PR #10958, opened 2021-08-27
Fixes validation issues associated with the Notification Form

AlexSCorey:
##### SUMMARY
This addresses #10557 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
Email validation
Email Edit

Twilio phone number validation
twilio phone number

### DIFF
--- a/awx/ui/src/screens/NotificationTemplate/shared/TypeInputsSubForm.js
+++ b/awx/ui/src/screens/NotificationTemplate/shared/TypeInputsSubForm.js
@@ -22,6 +22,7 @@ import {
   requiredEmail,
   url,
   minMaxValue,
+  twilioPhoneNumber,
 } from 'util/validators';
 import { NotificationType } from 'types';
 
@@ -398,7 +399,7 @@ function TwilioFields() {
         label={t`Source phone number`}
         name="notification_configuration.from_number"
         type="text"
-        validate={required(null)}
+        validate={combine([required(null), twilioPhoneNumber()])}
         isRequired
         tooltip={t`Enter the number associated with the "Messaging
           Service" in Twilio in the format +18005550199.`}
@@ -408,10 +409,10 @@ function TwilioFields() {
         label={t`Destination SMS number(s)`}
         name="notification_configuration.to_numbers"
         type="textarea"
-        validate={required(null)}
+        validate={combine([required(null), twilioPhoneNumber()])}
         isRequired
         tooltip={t`Enter one phone number per line to specify where to
-          route SMS messages.`}
+          route SMS messages. Phone numbers should be formatted +11231231234. For more information see Twilio documentation`}
       />
       <FormField
         id="twilio-account-sid"
--- a/awx/ui/src/util/validators.js
+++ b/awx/ui/src/util/validators.js
@@ -1,4 +1,4 @@
-import { t } from '@lingui/macro';
+import { plural, t } from '@lingui/macro';
 import { isValidDate } from '@patternfly/react-core';
 
 export function required(message) {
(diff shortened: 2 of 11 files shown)


awx/ui/src/util/validators.js
```diff
@@ -128,6 +128,20 @@ export function number() {
   };
 }
 
+export function twilioPhoneNumber() {
+  return (value) => {
+    let error;
+    if (!error) {
+      value.map((v) => {
+        if (!/^\s*(?:\+?(\d{1,3}))?[-. (]*(\d{3})(\d{3})(\d{4})/.test(v)) {
+          error = t`Please enter a valid phone number.  Phone numbers must begin with + and contain the country code, area code and subscriber number with no spaces, dashes or other characters. For example: +112312341234.  See your Twilio documentation for more information.`;
```
nixocio: This error message is really helpful, but long. What about we move some of this text to the tooltip? We could show the expected number format there, for example.
image

awx/ui/src/util/validators.js
```diff
@@ -128,6 +128,20 @@ export function number() {
   };
 }
 
+export function twilioPhoneNumber() {
+  return (value) => {
+    let error;
+    if (!error) {
+      value.map((v) => {
+        if (!/^\s*(?:\+?(\d{1,3}))?[-. (]*(\d{3})(\d{3})(\d{4})/.test(v)) {
```
nixocio: I looked at twilio docs about phone number validations, and I found.

https://www.twilio.com/docs/glossary/what-e164#regex-matching-for-e164

What are we taking in consideration international numbers as well?

```

RegEx Matching for E.164

According to the official ITU E.164 recommendation the format must be a number up to fifteen digits in length starting with a ‘+’. You can also exclude 0 as the first character since there are no country codes that start with 0. Here is a sample regular expression:

^\+[1-9]\d{1,14}$```

**Review by nixocio — COMMENTED** [review]
Good job @AlexSCorey . Just to confirm details with you.

File: awx/ui/src/util/validators.js (hunk 2 above)
Comment (AlexSCorey) [reply awx/ui/src/util/validators.js]: Yes, the regex does take international numbers into consideration.

File: awx/ui/src/screens/NotificationTemplate/shared/TypeInputsSubForm.js [hunk 3]
```diff
@@ -408,10 +409,21 @@ function TwilioFields() {
...
         type="textarea"
-        validate={required(null)}
+        validate={combine([required(null), twilioPhoneNumber()])}
         isRequired
-        tooltip={t`Enter one phone number per line to specify where to
-          route SMS messages.`}
+        tooltip={
+          <>
+            {t`Enter one phone number per line to specify where to
+          route SMS messages. Phone numbers should be formatted +11231231234. For more information see `}
```
nixocio: This message is very clear. 👍

awx/ui/src/screens/NotificationTemplate/shared/TypeInputsSubForm.js
```diff
@@ -408,10 +409,21 @@ function TwilioFields() {
...
+            <a
+              target="_blank"
+              href="https://www.twilio.com/docs/verify/migrating-1x-2x#e164-phone-number-formatting-for-sms-and-voice"
```
nixocio: If the third party docs link changes, we will have a broken link in our application - I would suggest not adding a link to a certain URL.

awx/ui/src/util/validators.js
nixocio: @AlexSCorey thanks.

I noticed that when I press `enter` after entering a valid number the validation is being triggered.

image

nixocio (changes requested):

akus062381:
@AlexSCorey - I am still seeing that these two fields (source phone number and destination SMS number) in Twilio accept alpha characters....
image

awx/ui/src/util/validators.js
```diff
@@ -1,4 +1,4 @@
-import { t } from '@lingui/macro';
+import { plural, t } from '@lingui/macro';
```
nixocio, line 1: Hey @AlexSCorey do you want to import lower case plural, instead of Plural component?

akus062381 (approved):
Looks good from a QE standpoint.

nixocio (approved):

PR #10964, opened 2021-08-29
Automate release process and changelog generation

shanemcd:
## Summary
This is the result of a weekend project that turned into something useful.

There is still more work to do in order to move the entire release process to GHA, but this PR:

- Automates the publishing of releases to https://github.com/ansible/awx/releases.
- Provides a solution for automating changeling generation, with the goal of:
  - Not making a changelog entry a hard requirement for each PR
  - Not requiring any additional files or commits
  - Allowing the changelog entry to be modified after a PR merges
  - Making the release process as frictionless as possible

How does this work?

### Release Publication

There is now a new "Release AWX" workflow:

awx 2021-08-29 13-09-55

Running this will create a new draft release at https://github.com/ansible/awx/releases.

### Changelog Generation

All new PR descriptions (this one included) now start with this HTML comment:

```html
<!--- changelog-entry
# Fill in 'msg' below to have an entry automatically added to the next release changelog.
# Leaving 'msg' blank will not generate a changelog entry for this PR.
# Please ensure this is a simple (and readable) one-line string.
---
msg: "Automated changelog generation and GitHub Release publication."
-->
```

When the new release workflow runs, it will look for all PRs merged since the last release that include a non-empty changelog message:

https://github.com/ansible/awx/blob/16fbde887572e4c4f91b927697177d9d87248cf2/tools/ansible/roles/release/tasks/main.yml#L12-L29

These messages will be used to construct a changelog in the description of the release itself:

image

 Once verified, the release can be published via the GitHub UI.


shanemcd:
To see this in action:

image

shanemcd:
image

tools/ansible/release.yml
```diff
@@ -0,0 +1,23 @@
+---
+- hosts: localhost
+  connection: local
+  tasks:
+    - name: Prepare Release Data
+      set_fact:
+        payload:
```
samdoran: This would be done in `vars` at the play level.

.github/workflows/release.yml
```diff
@@ -0,0 +1,56 @@
...
+
+      - name: Generate changelog
+        uses: shanemcd/simple-changelog-generator@v1
+        id: changelog
+        with:
+          repo: "${{ github.repository }}"
+
+      - name: Write changelog to file
+        run: |
+          cat << 'EOF' > /tmp/changelog
```
chrismeyersfsu, line 44: :fire: To anyone who stops by here in the future. This usage of input redirection instead of just `echo "${{ steps.changelog.outputs.changelog }}" > /tmp/changelog` prevents malicious shell injection.

chrismeyersfsu (approved):
Hot

Issue #10883, opened 2021-08-16
LDAP users not getting reflected in AWX 19.3.0


satbirsp:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

Hi Team 

After doing update/new install for 19.3.0 , Once LDAP new user login it will not show LDAP in front 
![awx](https://user-images.githubusercontent.com/50193939/129516092-eee9cc89-77dd-44af-9543-b263d6b446e7.png)

### AWX version

19.3.0

### Installation method

minikube

### Modifications

no

### Operating system

Ubuntu 20.04

### Web browser

Chrome

### Steps to reproduce

Install/update version 19.3.0 and login via new LDAP user who never log in to AWX . Change LDAP users to SystemAdmin group now new LDAP user will not able to login 

### Expected results

New LDAP user will able to login after changing to systemadmin 

### Actual results

New LDAP user not able to login after changing to systemadmin

lumarel:
I also ran in the exact same problem on a kubernetes cluster.
This is how the login looks in the log:
```
2021-08-17 08:24:57,426 DEBUG    [17b6e01d0c72477e88772d0b15d412c3] awx.analytics.performance request: <WSGIRequest: GET '/api/login/'>, response_time: 0.036s
[pid: 58|app: 0|req: 329/706] <ip-address> () {60 vars in 2482 bytes} [Tue Aug 17 08:24:57 2021] GET /api/login/ => generated 5710 bytes in 37 msecs (HTTP/1.1 200) 10 headers in 483 bytes (1 switches on core 0)
<ip-address> - - [17/Aug/2021:08:24:57 +0000] "GET /api/login/ HTTP/1.1" 200 5710 "https://<fqdn>/" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:78.0) Gecko/20100101 Firefox/78.0" "<ip-address>"
2021-08-17 08:24:57,514 WARNING  [13b719369dce43eaa20cc5e32bdbffc3] awx.api.generics Login failed for user <ldap-username> from <ip-address>
[...]

PR #10965, opened 2021-08-30
Associates ldap_dn on a first User() login

### LINKED ISSUE #507: LDAP login works only once per user
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
After configuring LDAP, I'm able to login. Attributes and groups seem to sync correctly, After logging out however I can no longer login with that same username. The only error in the `awx-task` log is `awx.api.generics Login failed for user`.

##### ENVIRONMENT
* AWX version: 19.3.0
* Operator version: 0.12.0
* Kubernetes version: 1.20.0-1089
* AWX install method: OKD 4.7

##### STEPS TO REPRODUCE

* Install AWX and configure LDAP authentication as below.
* Login with an LDAP user
* Logout
* Login again

##### EXPECTED RESULTS

User should be able to login again successfully.

##### ACTUAL RESULTS

User can only login once.

##### ADDITIONAL INFORMATION

Redacted config as follows:
```
LDAP Server URI
ldap://ldap-server.example.com/
LDAP Bind DN
uid=svc_awx,cn=sysaccounts,cn=etc,dc=example,dc=com
LDAP Bind Password
Encrypted
LDAP User DN Template
uid=%(user)s,cn=users,cn=accounts,dc=example,dc=com
LDAP Group Type
GroupOfNamesType
LDAP Require Group
cn=awx-users,cn=groups,cn=accounts,dc=example,dc=com
LDAP Deny Group
Not configured
LDAP Start TLS
On

LDAP User Search
[
  "cn=users,cn=accounts,dc=example,dc=com",
  "SCOPE_SUBTREE",
  "(uid=%(user)s)"
]

LDAP Group Search
[
  "cn=groups,cn=accounts,dc=example,dc=com",
  "SCOPE_SUBTREE",
  "(objectClass=nestedGroup)"
]

LDAP User Attribute Map
{
  "first_name": "givenName",
  "last_name": "sn",
  "email": "mail"
}

LDAP User Flags By Group
{
  "is_superuser": [
    "cn=awx-admins,cn=groups,cn=accounts,dc=example,dc=com"
  ]
}
```

Deleting the LDAP user using the admin account seems to allow logging in again.

tchellomello:
##### SUMMARY
Related: https://github.com/ansible/awx/issues/10883
Fixes: https://github.com/ansible/awx-operator/issues/507

Related: https://github.com/ansible/awx/pull/9703

Fix the missing ldap_dn attribute for new LDAP users.

To avoid calling the user.save() on every single login (PR#9703)
we can check if the user.profile is available. For new users,
accessing the user.profile throws an ValueError exception which
is capture on this fix.

* Example:
```
>>> _ = user.profile
*** ValueError: save() prohibited to prevent data loss due to unsaved related object 'user'.
>>> User.objects.filter(username=user.username).count()
0
```

This way, the user.save() gets called for brand users and will get the
ldap_dn associated as expected.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
19.3.0
```

##### ADDITIONAL INFORMATION
image

### DIFF
--- a/awx/sso/backends.py
+++ b/awx/sso/backends.py
@@ -395,10 +395,17 @@ def on_populate_user(sender, **kwargs):
         remove = bool(team_opts.get('remove', True))
         _update_m2m_from_groups(user, ldap_user, team.member_role.members, users_opts, remove)
 
+    # Check if user.profile is available, otherwise force user.save()
+    try:
+        _ = user.profile
+    except ValueError:
+        force_user_update = True
+    finally:
+        if force_user_update:
+            user.save()
+
     # Update user profile to store LDAP DN.
-    if force_user_update:
-        user.save()
-        profile = user.profile
-        if profile.ldap_dn != ldap_user.dn:
-            profile.ldap_dn = ldap_user.dn
-            profile.save()
+    profile = user.profile
+    if profile.ldap_dn != ldap_user.dn:
+        profile.ldap_dn = ldap_user.dn
+        profile.save()


rooftopcellist:
I checked this PR out and I can confirm that I can:

1. Configure LDAP server (ldap.forumsys.com)
2. Log in with an LDAP user

image

3. Make them a sys admin
image

4. Log back in as the LDAP user

image

rooftopcellist (approved):
LGTM

awx/sso/backends.py
```diff
@@ -395,10 +395,17 @@ def on_populate_user(sender, **kwargs):
         remove = bool(team_opts.get('remove', True))
         _update_m2m_from_groups(user, ldap_user, team.member_role.members, users_opts, remove)
 
+    # Check if user.profile is available, otherwise force user.save()
+    try:
+        _ = user.profile
+    except ValueError:
+        force_user_update = True
+    finally:
+        if force_user_update:
```
chrismeyersfsu, line 404:
```
>>> try:
...   raise Exception
... except ValueError:
...   print("hello")
... finally:
...   print("good bye")
...
good bye
  File "<stdin>", line 2, in <module>
Exception
```

You can end up in the `finally` due to some exception other than `ValueError` & `force_user_update` will be an undefined variable. Set `force_user_update = False` before the `try:` to be defensive so that if a programmer comes here in the future they are less likely to add code to the `try:` that might increase the chance of raising some other uncaught exception.

TL;DR set `force_user_update = False` before the `try:`.

awx/sso/backends.py
```diff
@@ -395,10 +395,17 @@ def on_populate_user(sender, **kwargs):
...
+            user.save()
+
     # Update user profile to store LDAP DN.
-    if force_user_update:
-        user.save()
-        profile = user.profile
-        if profile.ldap_dn != ldap_user.dn:
-            profile.ldap_dn = ldap_user.dn
-            profile.save()
+    profile = user.profile
```
chrismeyersfsu, line 408: Can this `user.profile` result in a `ValueError`? Or will the `save()` above ensure that `user.profile` gets set?

chrismeyersfsu (approved):

awx/sso/backends.py
tchellomello: @chrismeyersfsu the `force_user_update = False` is already set before the `try:`  statement at https://github.com/ansible/awx/blob/devel/awx/sso/backends.py#L357

awx/sso/backends.py
tchellomello: @chrismeyersfsu  during my tests, the only time I was able to get a `ValueError` was for a brand new user. For users already present in the database,  `user.profile`  worked as expected.

chrismeyersfsu (approved):

damluji:
Any updates when a release with this fix is coming out please?

PR #10980, opened 2021-08-30
Change the location of the receptor socket

jbradberry:
to /var/run/awx-receptor, to match what the installer is currently doing.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
+++ b/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
@@ -210,7 +210,7 @@ RUN for dir in \
       /var/log/nginx \
       /var/lib/postgresql \
       /var/run/supervisor \
-      /var/run/receptor \
+      /var/run/awx-receptor \
       /var/lib/nginx ; \
     do mkdir -m 0775 -p $dir ; chmod g+rw $dir ; chgrp root $dir ; done && \
     for file in \
--- a/tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
+++ b/tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
@@ -18,7 +18,7 @@ services:
       SDB_PORT: {{ awx_sdb_port_start }}
       AWX_GROUP_QUEUES: tower
       MAIN_NODE_TYPE: "${MAIN_NODE_TYPE:-hybrid}"
-      RECEPTORCTL_SOCKET: /var/run/receptor/receptor.sock
+      RECEPTORCTL_SOCKET: /var/run/awx-receptor/receptor.sock
 {% if loop.index == 1 %}
       RUN_MIGRATIONS: 1
 {% endif %}
--- a/tools/docker-compose/ansible/roles/sources/templates/receptor-awx.conf.j2
+++ b/tools/docker-compose/ansible/roles/sources/templates/receptor-awx.conf.j2
@@ -22,7 +22,7 @@
 
 - control-service:
     service: control
-    filename: /var/run/receptor/receptor.sock
+    filename: /var/run/awx-receptor/receptor.sock
 
 - work-command:
     worktype: local


beeankha (approved):

rebeccahhh (approved):

AlanCoding:
Also
 - `tools/docker-compose/ansible/roles/sources/templates/receptor-hop.conf.j2`
 - `tools/docker-compose/ansible/roles/sources/templates/receptor-worker.conf.j2`

AlanCoding (approved):
okay, I think this will work, but haven't tested.

AlanCoding:
I think this is what's not adding up for me.

https://github.com/ansible/awx/blob/ad7a5beb321931f1b66ca2c506a8aa776812e74d/tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2#L116

This is for the `tools_receptor_i` nodes, in other words, the execution nodes. They use the same image as the AWX nodes. They use a different receptor socket location.

You changed the permission stuff you did in the Dockerfile, but you didn't change the socket file location in the execution nodes. That's not adding up.

AlanCoding:
You also missed this:

https://github.com/ansible/awx/blob/eec4f8dcc27428cd1aeb1c251d43052a3ea780e8/tools/docker-compose/supervisor.conf#L88

which is the actual receptor command that gets run.

AlanCoding (changes requested):
I take it back - none of this works like either of us thought it did

jbradberry:
@AlanCoding yeah, I didn't feel very confident about this after discussion yesterday afternoon.  I'm inclined to throw this PR away entirely.

AlanCoding:
I had to call in for help on this. Here is what I've tested after discussing with @fosterseth 

```diff
diff --git a/tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2 b/tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
index 3781f21888..d31296926b 100644
--- a/tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
+++ b/tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
@@ -114,6 +114,8 @@ services:
     container_name: tools_receptor_{{ loop.index }}
     hostname: receptor-1
     command: 'receptor --config /etc/receptor/receptor.conf'
+    environment:
+      RECEPTORCTL_SOCKET: /var/run/receptor/receptor.sock
     links:
       - receptor-hop
     volumes:
diff --git a/tools/docker-compose/ansible/roles/sources/templates/receptor-worker.conf.j2 b/tools/docker-compose/ansible/roles/sources/templates/receptor-worker.conf.j2
index ef31f4d443..03b8573f3e 100644
--- a/tools/docker-compose/ansible/roles/sources/templates/receptor-worker.conf.j2
+++ b/tools/docker-compose/ansible/roles/sources/templates/receptor-worker.conf.j2
@@ -16,3 +16,4 @@
 
 - control-service:
     service: control
+    filename: /var/run/receptor/receptor.sock
```

The AWX and receptor nodes are both using the same container image. Since the Dockerfile does relevant (file permission) stuff, if you move it for 1 of these nodes, you should probably move it for both. You can't move it for the receptor node, because a location isn't specified.

As far as I can tell, the real installer will use the `awx-receptor` directory for execution nodes as well as control nodes. In fact, it uses the same template for both, which we could also do with more work.

So, with the new directory added to the above places, :+1: that should work well. Since these are templated files, you can put it in defaults.yml so that it only gets typed out once.

AlanCoding (approved):
thinking about this more, I suspect it won't break anything to merge as-is, and we can do what I mention as followup work

AlanCoding:
merged the replacement of this PR

thanks!

PR #10929, opened 2021-08-23
Validate that control-only Instance nodes cannot change IG membership

jbradberry:
##### SUMMARY
Backend part of #10853 

related #10830 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.3.0
```

PR #10985, opened 2021-08-31
Exclude control-only nodes from IG policy calculations

AlanCoding:
Debugging with @thenets, we discovered a bug where (in the development environment) it kept adding `receptor-1` and `receptor-2` to the `controlplane` instance group.

```json
    "instance_groups": [
        {
            "name": "controlplane",
            "capacity": 74,
            "instances": [
                "awx_1",
                "execution_node_1",
                "receptor-1",
                "receptor-2"
            ]
        },
        {
            "name": "default",
            "capacity": 37,
            "instances": [
                "receptor-1",
                "receptor-2"
            ]
        },
```

In this case, `awx_1` is a control node and the receptor nodes are execution nodes. The receptor nodes are absolutely _not_ a part of the control plane, and they were incorrectly added.

----

There's another side to this, where we have agreed that control nodes shouldn't be added to user-defined IGs, which includes the `default` instance group.

This diff prevents the policy rules from adding control nodes to user-controlled IGs, which is a goal shared with @jbradberry's https://github.com/ansible/awx/pull/10929. In that PR, it's preventing this association via the API. This is enforcing the same rule via the policy rules.

### DIFF
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -250,6 +250,8 @@ def apply_cluster_membership_policies():
         for g in sorted(actual_groups, key=lambda x: len(x.instances)):
             exclude_type = 'execution' if g.obj.name == settings.DEFAULT_CONTROL_PLANE_QUEUE_NAME else 'control'
             candidate_pool_ct = len([i for i in actual_instances if i.obj.node_type != exclude_type])
+            if not candidate_pool_ct:
+                continue
             policy_per_added = []
             for i in sorted(actual_instances, key=lambda x: len(x.groups)):
                 if i.obj.node_type == exclude_type:
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -229,8 +229,11 @@ def apply_cluster_membership_policies():
         actual_instances = [Node(obj=i, groups=[]) for i in all_instances if i.managed_by_policy]
         logger.debug("Total instances: {}, available for policy: {}".format(total_instances, len(actual_instances)))
         for g in sorted(actual_groups, key=lambda x: len(x.instances)):
+            exclude_type = 'execution' if g.obj.name == settings.DEFAULT_CONTROL_PLANE_QUEUE_NAME else 'control'
             policy_min_added = []
             for i in sorted(actual_instances, key=lambda x: len(x.groups)):
+                if i.obj.node_type == exclude_type:
+                    continue  # never place execution instances in controlplane group or control instances in other groups
                 if len(g.instances) >= g.obj.policy_instance_minimum:
                     break
                 if i.obj.id in g.instances:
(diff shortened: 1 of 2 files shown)


AlanCoding:
Cluster status with this fix in place:

```json
    "instance_groups": [
        {
            "name": "controlplane",
            "capacity": 298,
            "instances": [
                "awx_1"
            ]
        },
        {
            "name": "default",
            "capacity": 0,
            "instances": [
                "receptor-1",
                "receptor-2"
            ]
        }
    ]
}
```

AlanCoding:
@thenets I want to talk out some expectations. If you're interested, you could use these as some test cases to get rolling with integration tests.

 - the `controlplane` and `default` instance groups should always exist, this is true regardless of install method (VMs, OCP)
 - if control type nodes exist, they should always be present in the controlplane group, but not in the default group
 - if execution type nodes exist, they should always be present in the default group, but not in the controlplane group
 - if you create a new instance group and give it a policy_instance_percentage of 100, then it should basically obtain the same instances as the default group (hybrid and execution nodes)

beeankha (approved):

awx/main/tasks.py
```diff
@@ -245,13 +248,17 @@ def apply_cluster_membership_policies():
...
             policy_per_added = []
             for i in sorted(actual_instances, key=lambda x: len(x.groups)):
+                if i.obj.node_type == exclude_type:
+                    continue
                 if i.obj.id in g.instances:
                     # If the instance is already _in_ the group, it was
                     # applied earlier via a minimum policy or policy list
                     continue
-                if 100 * float(len(g.instances)) / len(actual_instances) >= g.obj.policy_instance_percentage:
+                if 100 * float(len(g.instances)) / candidate_pool_ct >= g.obj.policy_instance_percentage:
```
jladdjr, line 263: super corner case:
* you deploy awx with control nodes only (intending to deploy execution nodes dynamically later perhaps)
* you create a new IG
* `candidate_pool_ct` will be zero => divide by zero error

awx/main/tests/functional/test_instances.py
```diff
@@ -243,6 +243,40 @@ def test_policy_instance_list_explicitly_pinned(instance_factory, instance_group
...
+    of available execution nodes (1 out of 2), as opposed to 50 percent
+    of all available nodes (2 out of 4) which include unusable control nodes
+    """
+    ig = InstanceGroup.objects.create(name='bar', policy_instance_percentage=50)
+    for i in range(2):
+        Instance.objects.create(hostname=f'foo-{i}', node_type='control')
+    for i in range(2):
+        Instance.objects.create(hostname=f'bar-{i}', node_type='execution')
+    apply_cluster_membership_policies()
+    assert set(['bar-0']) == set([inst.hostname for inst in ig.instances.all()])
```
jladdjr: if apply_cluster_membership_policies gets tweaked later on, seems like `bar-1` could get chosen instead and test could get flaky. Maybe just assert that there's one instance that begins with `bar-`?

jladdjr:
Came across two things that are minor, but probably worth tweaking. Otherwise, looks awesome - great fix!

awx/main/tests/functional/test_instances.py
AlanCoding: I know that the task does `sorted()` to the lists precisely to make this deterministic. But we could relax the assertion too, yes

awx/main/tasks.py
AlanCoding: if there are no candidate instances, we shouldn't ever be getting here, because that means we're considering whether to add an instance. We could add a `continue` in that case

Issue #9013, opened 2021-01-05
Notify user that a sync has started and completed


AlexSCorey:
On the details view for Inventory Sources the user can click a sync button.  There is no notification to the user of the start, success, or failure of that action.  We should notify the user some how.  Consider adding websockets to this view and rendering the JobCancelButton when syncing is in progress.  https://github.com/ansible/awx/pull/10113

##### ISSUE TYPE
 - Feature Idea

unlikelyzero:
Marking as enhancement instead of feature as this is not required for feature parity

john-westcott-iv:
What if clicking the sync button on the details page took you to the job screen? At least then you would know that Tower is "doing something" once the button is pushed.

mabashian:
Before anyone works on this lets discuss with Bill/Taufique

wenottingham:
We do now have very minimal feedback that the sync button was presed, you do still have to go to another page to see the status.

nixocio:
After following up with @mabashian about this issue, in order to be consistent we should add the same UX already implemented on project details page. Not allowing the user to start multiple `syncs` from the details page.

<img width="1497" alt="image" src="https://user-images.githubusercontent.com/9053044/130455353-440c81dc-b5b7-4e6a-b370-668e4c382023.png">

akus062381:
resolved

PR #10987, opened 2021-08-31
Add websockets to Inventory Source Details

nixocio:
Add websockets to Inventory Source Details

image

See: https://github.com/ansible/awx/issues/9013

### DIFF
--- a/awx/ui/src/screens/Inventory/InventorySourceDetail/InventorySourceDetail.js
+++ b/awx/ui/src/screens/Inventory/InventorySourceDetail/InventorySourceDetail.js
@@ -7,22 +7,27 @@ import {
   TextListItem,
   TextListVariants,
   TextListItemVariants,
+  Tooltip,
 } from '@patternfly/react-core';
 import AlertModal from 'components/AlertModal';
-import { CardBody, CardActionsRow } from 'components/Card';
-import { VariablesDetail } from 'components/CodeEditor';
 import ContentError from 'components/ContentError';
 import ContentLoading from 'components/ContentLoading';
 import CredentialChip from 'components/CredentialChip';
 import DeleteButton from 'components/DeleteButton';
-import ExecutionEnvironmentDetail from 'components/ExecutionEnvironmentDetail';
-import { DetailList, Detail, UserDateDetail } from 'components/DetailList';
 import ErrorDetail from 'components/ErrorDetail';
+import ExecutionEnvironmentDetail from 'components/ExecutionEnvironmentDetail';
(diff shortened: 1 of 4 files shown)


AlexSCorey (commented):
On the projects page, when we create a new project and click save a sync is started as we move to the details view, and that change is reflected in the sync status icon.    Seems like something we should implement here.  Also, I'd expect the name of the inventory source to always be the first thing displayed on the details view instead of the sync status icon being displayed first.

awx/ui/src/screens/Inventory/InventorySourceDetail/InventorySourceDetail.js
```diff
@@ -226,16 +275,26 @@ function InventorySourceDetail({ inventorySource }) {
...
+            <InventorySourceSyncButton source={inventorySource} icon={false} />
+          ))}
         {user_capabilities?.delete && (
           <DeleteButton
             name={name}
             modalTitle={t`Delete inventory source`}
             onConfirm={handleDelete}
             deleteDetailsRequests={deleteDetailsRequests}
             deleteMessage={t`This inventory source is currently being used by other resources that rely on it. Are you sure you want to delete it?`}
+            isDisabled={job?.status === 'running'}
```
mabashian, line 297: I can't remember - can you delete jobs that are in `new`, `pending`, or `waiting`?

mabashian (approved):
Code looks good to me - I think this is a big improvement and great for consistency.

awx/ui/src/screens/Inventory/InventorySourceDetail/InventorySourceDetail.js
nixocio: This change will disable the possibility to delete the inventory source when the sync is running.  Same pattern that we have for Project Details. 

The one to cancel the job itself is here - https://github.com/ansible/awx/blob/edceb0274291f3095a0b2e53bf20bd1256defa0d/awx/ui/src/screens/Inventory/InventorySourceDetail/InventorySourceDetail.js#L279

nixocio:
> On the projects page, when we create a new project and click save a sync is started as we move to the details view, and that change is reflected in the sync status icon. Seems like something we should implement here. Also, I'd expect the name of the inventory source to always be the first thing displayed on the details view instead of the sync status icon being displayed first.

I updated the name location.  I also looked at the previous UI how the sync of the inventory source was implemented. It does not automatically sync like project does, we can file an issue for a future enhancement.

AlexSCorey (approved):

PR #10988, opened 2021-08-31
List dir where receptor socket should be

kdelee:
This is for adding more info to the sos report


AlanCoding (approved):

AlanCoding:
it would also be good to add `receptorctl status` to the commands.

kdelee:
@AlanCoding going to merge and then follow up, not sure if I should use full path to `receptorctl` and if it is in the awx venv or elsewhere

Issue #10966, opened 2021-08-30
Jobs and traceback output filtering, 'Filter By' event, needs to be marked for translations


sshedmake:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

When in ja google-chrome, while Jobs and traceback output filtering, 'Filter By' event needs to be marked for translations

### AWX version

19.3.0

### Installation method

docker on linux

### Modifications

yes

### Steps to reproduce

1. Launch UI in ja google-chrome
2. Click Jobs from nav bar
3. Launch a simple 'ping' job
4. Click the Stdout drop down
5. Click or select the option 'Event' button
6. Click or select 'Filter By' Event button
7. Observe 'Filter By' translation

### Expected results

100% Translated (into Japanese)

### Actual results

Appears in English

### Additional information

1. When in ja google-chrome, to find strings obviously in English, just swapped the ~/ja/messages.js with dummy ja translation and rebuild the UI
2. Screenshot 1 highlighting the unmarked string when in ja google-chrome
![Screenshot from 2021-08-27 11-54-50_1](https://user-images.githubusercontent.com/12151785/131330055-81654d38-18b1-4519-807d-753427274c95.png)
3. Screenshot 1 highlighting the unmarked string when in ja google-chrome
![Screenshot from 2021-08-29 21-50-56_1](https://user-images.githubusercontent.com/12151785/131330207-7831fae9-8af1-462a-9e0e-09d1b1a6f9b0.png)

PR #10994, opened 2021-08-31
Add string `Filter By` to be translated

nixocio:
Add string `Filter By` to be translated

See:https://github.com/ansible/awx/issues/10966

image


AlexSCorey (approved):
Looks good

Issue #10990, opened 2021-08-31
awx reports identical UUIDs for instances


jladdjr:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

Fresh deployment of awx with three hybrid nodes shows that all three instances have the same UUID.

Related ansible-runner mechanism for generating uuid:
https://github.com/ansible/ansible-runner/blob/devel/ansible_runner/utils/capacity.py#L24

### AWX version

devel

### Installation method

docker on linux

### Modifications

no

### Steps to reproduce

out-of-the box deployment with 3 hybrid nodes

### Expected results

Unique UUIDs for each instance.

### Actual results

Identical UUIDs:
https://gist.github.com/jladdjr/375a2207a5be9599798eec6384660ef6

jladdjr:
Think I see what's going on here:

1. Hybrid instance is provisioned using `awx-manage provision_instance`
2. If a uuid is not provided, it [retrieves it from `settings.SYSTEM_UUID`](https://github.com/ansible/awx/blob/9083425c24a42090e0b8eba5d668f847ea8112a7/awx/main/management/commands/provision_instance.py#L37). _This will always be the_ local system's _UUID_.

In short, `awx-manage provision_instance` will stamp all new instances with its own UUID.

In the deployment where this was noticed all instances were registered from the same host.

jladdjr:
https://github.com/ansible/awx/pull/10995 updates awx to generate a fresh UUID instead of defaulting to the system UUID. It also adds a new optional argument to pass in a UUID.

jladdjr:
Resolved by #10995 . Follow-on work for OCP being tracked with https://github.com/ansible/awx/issues/11019.

PR #10995, opened 2021-08-31
awx-manage provision_instance should not default to using local system's UUID

jladdjr:
##### SUMMARY

Per https://github.com/ansible/awx/issues/10990, `awx-manage provision_instance` defaults to using _the current system's_ UUID. If one is provisioning multiple instances from a single instance, they will all end up sharing the same UUID. 

Patch updates the command to instead generate a new uuid. It also adds a new optional argument to pass in a UUID.

related https://github.com/ansible/awx/issues/10990

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API


jladdjr:
One thing not currently addressed is the existence of `Settings.SYSTEM_UUID`. Seems like we have two sources of truth: what's in the db and what's on disk. Trying to gather some background on this (and also see if we can just get down to a single source of truth if possible).

AlanCoding:
I agree that `awx-manage provision_instance` using `settings.SYSTEM_UUID` is fundamentally flawed.

> Seems like we have two sources of truth: what's in the db and what's on disk

I think the db is the red herring here. The entire point of a system UUID is that it's available on disk.

It's not unreasonable that `awx-manage provision_instance` _leaves uuid blank_. We independently developed this pattern for execution nodes, and some version of this might be the original intent. First, we get a clue that the instance (mainly keyed on hostname) exists, so we create the record. Later, when we have the ability to run health check code, we save the uuid.

Why not just back out any attempt to obtain the uuid independently in the `register` method or the provision_instance command. Then, declare that there will only be 1 place where the UUID is set, that would be:

 - In `Instance.local_health_check` called from cluster_node_heartbeat for control / hybrid nodes
 - the remote health check data payload for execution nodes

The core principle of the local health check is that you are physically on the instance that it is running for, making the file settings valid. There would only be a minor gap of time during initial install when it gets left blank, and no one will care.

tchellomello:
Great finding. In the case of the current behavior is kept, once the PR lands, and **if you decide** to continue using the `ha.py` as a source of truth, we can modify the installer to pass the `UUID`. As we can see, the task that creates the `/etc/tower/conf.d/ha.py` already creates a register  (see https://github.com/ansible/automation-platform-collection/blob/devel/roles/misc/tasks/main.yml#L2-L5) 
```yaml
- name: If Ansible does not provide a system UUID, get a random one.
  shell: "echo 'import uuid; print(uuid.uuid4())' | awx-python"
  register: system_uuid
  changed_when: False
```

So we basically need to modify the registration task for the following:

```yaml
- name: Register the node instance in the database.
  shell: awx-manage provision_instance --hostname="{{ receptor_host_identifier }}" --node_type="{{ node_type }}" --uid="{{ hostvars[inventory_hostname]['system_uuid']['stdout'] }}"
  args:
    executable: /bin/bash
  become_user: awx
  become: True
  register: provision_execution_node
  changed_when: "'changed: True' in provision_execution_node.stdout"
  delegate_to: "{{ groups['automationcontroller'][0] }}"
  when: node_type in receptor_workable_types
```

jladdjr:
That all makes sense 👍 

> There would only be a minor gap of time during initial install when it gets left blank, and no one will care.

This ^ is what I was wondering about. I'm wondering if we might actually key off of the UUID for something (surely we don't just have the UUID just to have it). But we could probably do something as a stop gap (dummy UUID or something). We may need to tweak analytics collectors, which may expect the UUID, too, so they don't get junk data.

But in general, I think you're right - we've already got a pattern for this, and it'd be good to move toward making the SYSTEM_UUID _the_ single source of truth.

jladdjr:
fwiw, checked to make sure that the SYSTEM_UUID file (ha.py) does not get overwritten on subsequent runs of the installer. Looks like it shouldn't:
https://github.com/ansible/automation-platform-collection/blob/devel/roles/misc/tasks/main.yml#L15

awx/main/managers.py
```diff
@@ -115,7 +116,7 @@ def me(self):
 
     def register(self, uuid=None, hostname=None, ip_address=None, node_type='hybrid', defaults=None):
         if not uuid:
-            uuid = settings.SYSTEM_UUID
```
jladdjr, line 118: oh, wait, we said we wanted to just let this field resolve to the default value if not specified

jladdjr: alas, it doesn't have a default, will add that

jladdjr: done

awx/main/managers.py
```diff
@@ -148,12 +145,15 @@ def register(self, uuid=None, hostname=None, ip_address=None, node_type='hybrid'
                     return (False, instance)
 
             # Create new instance, and fill in default values
-            create_defaults = dict(capacity=0, uuid=uuid)
+            create_defaults = dict(capacity=0)
             if defaults is not None:
                 create_defaults.update(defaults)
+            uuid_option = {}
+            if uuid is not None:
+                uuid_option = dict(uuid=uuid)
```
AlanCoding, line 153: you could just throw this into `create_defaults` as opposed to creating another dict. The double `**` seems to work, but I don't see any need for it here.

AlanCoding (approved):
the code looks well-considered and tested on all the functional elements.

awx/main/managers.py
jladdjr: I avoided `create_defaults` mostly because it wasn't technically a default. But if we're not too picky about that, then yep, sounds great.

AlanCoding: It's a default in the way that Django's [get_or_create](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#get-or-create) method referrers to defaults. If there's an existing object, the value will be ignored, but for a new object, it will be used.

beeankha (approved):

AlanCoding:
@chrismeyersfsu has added 1 more request of this patch - that we when OpenShift nodes are auto-registered (different kind of auto discovery) that it should pass the SYSTEM_UUID setting in its call to `register`

jladdjr:
> @chrismeyersfsu has added 1 more request of this patch - that we when OpenShift nodes are auto-registered (different kind of auto discovery) that it should pass the SYSTEM_UUID setting in its call to register

looking into it 👍

jladdjr:
@AlanCoding @chrismeyersfsu 

I'm trying to deploy to OCP right now (it's been a while). In the meantime, I'm trying to infer from the code where a freshly deployed OCP instance (would be a pod, I believe) would get registered.

Would this happen as part of the cluster_node_heartbeat task?
https://github.com/ansible/awx/blob/324ca7fe7248378dc8d83043d32866562589d17e/awx/main/tasks.py#L511

AlanCoding:
@jladdjr yes, and that method `get_or_register` has a code path specific to OCP (`AWX_AUTO_DEPROVISION_INSTANCES` is equivalent with OCP) where `register` is called, and all that needs to happen is that `uuid` gets added to those call kwargs

jladdjr:
OCP follow-up work is being tracked here:
https://github.com/ansible/awx/issues/11019

Issue #10737, opened 2021-07-23
Handle unresponsive instances (when joining, when trying to run a job in that instance, and coming back online)


AlanCoding:
I want to note that, as of now, when I test running against a node which has gone off the receptor network, then it gets stuck in the running state.

The dispatcher should never attempt to send a job to a node which `receptorctl status` does not currently recognize. Instead, we should mark the instance as lost and provide details.

jladdjr:
Studied how we run heartbeats, handle reaping etc.

Starting to make some updates here:
https://github.com/ansible/awx/compare/mesh_code...mesh_code_healthchecks?expand=1

AlanCoding:
Let me just echo that I still hit the case mentioned in this issue. I run against an instance that has a receptor name which is not in the mesh. The job will eventually have:

> Task was marked as running but was not present in the job queue, so it has been marked as failed.

This might be a pre-existing issue, but it becomes important with execution nodes. It may very well be the case that receptorctl is giving some useful feedback, but we're not recognizing it, so we just keep trying to read from the pipe. The whole transmit function is probably putting it in a pipe-to-nowhere. There's got to be some useful debugging we can do for this.

It's probably overkill to check the network before starting a job. There is almost certainly some useful feedback from receptor that we're _not seeing_.

For this feature, however, we also need to add some special handling for the instance when this occurs. We need error handling, and then maybe submit the task to re-run the instance's health check. That would make a ton of sense.

jladdjr:
> I run against an instance that has a receptor name which is not in the mesh. The job will eventually have 'Task was marked as running but was not present in the job queue, so it has been marked as failed.' ... It may very well be the case that receptorctl is giving some useful feedback, but we're not recognizing it, so we just keep trying to read from the pipe.

> The whole transmit function is probably putting it in a pipe-to-nowhere. There's got to be some useful debugging we can do for this.

@AlanCoding Reading through `awx.main.tasks.py:AWXReceptorJob._run_internal()` I think I see a spot where awx may be vulnerable to the kind of hanging that you're describing. When we transmit the job, we have a join() that can potentially wait forever. If we're trying to transmit to a node that doesn't exist, perhaps this is where we're hanging.

I'm wondering if we should use a timeout here. Something like this:

https://github.com/ansible/awx/compare/mesh_code...mesh_code_healthchecks?expand=1#diff-6fb7bd98930ec6a05c1ae9296b41b63921e1813dcee53895f16769f910917c38R3144

[...]

PR #11002, opened 2021-09-01
Allow manually running a health check, and make other adjustments to the health check trigger

AlanCoding:
##### SUMMARY
This is intended to close out the remaining API work on https://github.com/ansible/awx/issues/10737

Things being done here:

 - Introduced a new endpoint that the user can use to manually kick off a health check, anticipated that the UI and testing will make use of this
 - Frequency of health checks for execution nodes with known errors is reduced to once per 10 minutes by default
 - If a job fails by some unclear error in AWXReceptorJob, then this kicks off a health check so that we don't keep sending jobs to an instance that has an error

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API


awx/api/views/__init__.py
```diff
@@ -403,6 +404,32 @@ class InstanceInstanceGroupsList(InstanceGroupMembershipMixin, SubListCreateAtta
     relationship = 'rampart_groups'
 
 
+class InstanceHealthCheck(GenericAPIView):
+
+    name = _('Instance Health Check')
+    model = models.Instance
+    serializer_class = serializers.InstanceHealthCheckSerializer
+    permission_classes = (IsSuperUser,)
```
AlexSCorey: Is this saying that only super users will be able to retrieve health check data with a GET request.  System Auditors will also have access to the instance groups and instances list in the UI so it might be helpful to allow them to retrieve this data as well.

awx/settings/defaults.py
```diff
@@ -422,6 +422,7 @@ def IS_TESTING(argv=None):
 # heartbeat period can factor into some forms of logic, so it is maintained as a setting here
 CLUSTER_NODE_HEARTBEAT_PERIOD = 60
 RECEPTOR_SERVICE_ADVERTISEMENT_PERIOD = 60  # https://github.com/ansible/receptor/blob/aa1d589e154d8a0cb99a220aff8f98faf2273be6/pkg/netceptor/netceptor.go#L34
+EXECUTION_NODE_REMEDIATION_CHECKS = 60 * 10  # once every 10 minutes check if an execution node errors have been resolved
```
AlexSCorey, line 425: Is this something we should render in the UI Settings?

AlexSCorey (commented):
Just some questions asked below.

awx/main/tasks.py
```diff
@@ -409,22 +409,26 @@ def cleanup_execution_environment_images():
...
 
     instance.save_health_data(
-        version='ansible-runner-' + data.get('runner_version', '???'),
+        version=('ansible-runner-' + data.get('runner_version', '???')) if instance.node_type == 'execution' else None,
         cpu=data.get('cpu_count', 0),
         memory=data.get('mem_in_bytes', 0),
-        uuid=data.get('uuid'),
-        has_error=bool(data.get('errors')),
+        uuid=data.get('uuid') if instance.node_type == 'execution' else None,  # TODO, have --worker-info read from settings file
+        errors='\n'.join(data.get('errors', [])),
```
AlanCoding, line 450: This just keeps getting uglier, and I'm leaning towards redesigning this to use the dispatcher system as opposed to the receptor mesh for main cluster nodes.

awx/api/views/__init__.py
AlanCoding: That's valid.

I will plan to allow superusers to GET and POST, and allow system auditors to only GET.

awx/settings/defaults.py
AlanCoding: Out of these 3 settings, this is the only one that would be valid to implement as a CTiT setting. It will not require a restart in order to take effect.

I don't want to volunteer to do this, because I don't know if it's worth it. But yes, we could put this in UI settings.

awx/main/tasks.py
rebeccahhh: I actually really like this idea, the dispatcher system is where I would expect this to be rather than over the mesh.

AlanCoding:
Tested in a real deployment and the endpoint is working for all the node types. Still some tests to write, but everything looks functional.

awx/main/tasks.py
```diff
@@ -3028,12 +3055,17 @@ def run(self):
...
-            return self._run_internal(receptor_ctl)
+            res = self._run_internal(receptor_ctl)
+            return res
         finally:
             # Make sure to always release the work unit if we established it
             if self.unit_id is not None and settings.RECEPTOR_RELEASE_WORK:
                 receptor_ctl.simple_command(f"work release {self.unit_id}")
+            # If an error occured without the job itself failing, it could be a broken instance
+            if self.work_type == 'ansible-runner' and res is None or getattr(res, 'rc', None) is None:
+                execution_node_health_check(self.task.instance.execution_node)
```
AlanCoding, line 3068: @shanemcd I'd like for you to think about this.

For healthy execution nodes _we don't regularly re-run health checks_. But what if one of them was working, but is no longer working? We shouldn't keep sending jobs there. So we have 2 options, we can run checks on a regular basis, or we can re-run health checks when a job on that instance gets a suspicious error.

We don't want to do this when the ansible-playbook sub-process errors (have an `rc`), because that's on the execution layer, not the instance configuration layer (as far as we know).

awx/api/templates/api/instance_health_check.md
```diff
@@ -0,0 +1,33 @@
...
+
+{% ifmeth POST %}
+# Manually Initiate a Health Check
+For purposes of error remediation or debugging, a health check can be
+manually initiated by making a POST request to this endpoint.
+
+This will submit the work unit to the target node through the receptor mesh and wait for it to finish.
+The model will be updated with the result.
+Up-to-date values of the fields will be returned in the response data.
+{% endifmeth %}
```
AlanCoding, line 33: @tvo318 these are docs for the API browser, but it might be a good overview of what's happening here in general.

awx/main/tasks.py
thenets: Yeah. The output didn't work for me on dev env:

```
{
    "uuid": "00000000-0000-0000-0000-000000000000",
    "hostname": "awx_1",
    "version": "19.2.2",
    "last_health_check": "2021-09-03T14:37:08.587927Z",
    "errors": "",
    "cpu": 8,
    "memory": 6030827520,
    "cpu_capacity": 32,
    "mem_capacity": 37,
    "capacity": 37
}
```

AlanCoding: @thenets that output is perfectly fine. What did you think is wrong with it?

thenets: @AlanCoding I thought the `uuid` would be something different than zeros. If this was expected, perfect :)

thenets (approved):

awx/main/tasks.py
AlanCoding: right, the dev environment just never set that up. Someone could do it, but it's not specific to the current feature work.

AlanCoding:
I expect my last commit will be the last code change to this before merge.

I'm working on preparing some more writeups from my notes to hand over to other people right now, and then I'll see about merging.

AlanCoding:
Added some more info here https://gist.github.com/AlanCoding/33f79993fcf08f3b2dac47e5c5ea45cc#file-health_check_wrapup-md

awx/api/views/__init__.py
```diff
@@ -408,6 +409,54 @@ def is_valid_relation(self, parent, sub, created=False):
...
+        if obj.node_type == 'execution':
+            from awx.main.tasks import execution_node_health_check
+
+            runner_data = execution_node_health_check(obj.hostname)
+            obj.refresh_from_db()
+            data = self.get_serializer(data=request.data).to_representation(obj)
+            # Add in some extra unsaved fields
+            for extra_field in ('transmit_timing', 'run_timing'):
+                if extra_field in runner_data:
+                    data[extra_field] = runner_data[extra_field]
```
AlanCoding, line 436: We don't have to add these fields. I was on the fence about it, they don't make sense to me on the serializer, but they could be helpful.

beeankha (approved):

Makefile
```diff
@@ -175,7 +175,7 @@ init:
 	fi; \
 	$(MANAGEMENT_COMMAND) provision_instance --hostname=$(COMPOSE_HOST) --node_type=$(MAIN_NODE_TYPE); \
 	$(MANAGEMENT_COMMAND) register_queue --queuename=controlplane --instance_percent=100;\
-	$(MANAGEMENT_COMMAND) register_queue --queuename=default;
+	$(MANAGEMENT_COMMAND) register_queue --queuename=default --instance_percent=100;
```
AlanCoding, line 178: It seems that this was preventing hybrid nodes from getting added to the "default" instance group, so I thought I would go ahead and fix it.

awx/api/templates/api/instance_health_check.md
```diff
@@ -0,0 +1,33 @@
+{% ifmeth GET %}
```
jladdjr, line 1: These docs look great. Love that this info is being surfaced. Can take some real splunking otherwise to see how this actually works for awx.

awx/api/permissions.py
```diff
@@ -236,13 +236,18 @@ def check_post_permissions(self, request, view, obj=None):
...
-class IsSuperUser(permissions.BasePermission):
+class IsSystemAdminOrAuditor(permissions.BasePermission):
     """
-    Allows access only to admin users.
+    Allows write access only to system admin users.
+    Allows read access only to system auditor users.
     """
 
     def has_permission(self, request, view):
-        return request.user and request.user.is_superuser
```
jladdjr, line 245: 👍

awx/api/views/__init__.py
```diff
@@ -408,6 +409,54 @@ def is_valid_relation(self, parent, sub, created=False):
...
+        data = self.get_serializer(data=request.data).to_representation(obj)
+        return Response(data, status=status.HTTP_200_OK)
+
+    def post(self, request, *args, **kwargs):
+        obj = self.get_object()
+
+        if obj.node_type == 'execution':
+            from awx.main.tasks import execution_node_health_check
+
+            runner_data = execution_node_health_check(obj.hostname)
```
jladdjr, line 430: I'm wondering if this request is ever going to timeout waiting for this to complete. I guess `ansible-runner worker --worker-info` doesn't require pulling an image, so it should be pretty quick to respond (ditto for hybrid / control nodes - should be fast). So, probably not an issue.

awx/main/tests/functional/test_instances.py
```diff
@@ -324,6 +324,23 @@ def test_instance_group_capacity(instance_factory, instance_group_factory):
...
+def test_health_check_clears_errors():
+    instance = Instance.objects.create(hostname='foo-1', enabled=True, capacity=0, errors='something went wrong')
+    data = dict(version='ansible-runner-4.2', cpu=782, memory=int(39e9), uuid='asdfasdfasdfasdfasdf', errors='')
+    instance.save_health_data(**data)
+    for k, v in data.items():
+        assert getattr(instance, k) == v
+
+
+@pytest.mark.django_db
+def test_health_check_oh_no():
```
jladdjr, line 337: https://www.youtube.com/watch?v=e-eGqk6htfE

jladdjr (approved):

awx/api/views/__init__.py
AlanCoding: For the _execution node_ case, the maximum time it could take is practically 20 seconds, because _that method_ enforces timeouts, and will return errors if timeout happens.

But you reminded me that I didn't handle the timeout case for control nodes... which should be _exceedingly_ rare but I will still add something really quick here.

Issue #10971, opened 2021-08-30
Schedule with prompts Template, while input of job tag, string 'No results found', needs to be marked for translations


sshedmake:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

When in ja google-chrome, while editing 'Schedule with prompts' Template, and input of job tag, string 'No results found', needs to be marked for translations

### AWX version

19.3.0

### Installation method

docker on linux

### Modifications

yes

### Steps to reproduce

1. Launch UI in ja google-chrome
2. Click 'Templates' from nav bar
3. Click (or create) the said Job Template titled 'Tower4 - new schedule with prompts'
4. Click on the 'Edit' button
5. Input a Job Tag
6. Observe translation of string 'No results found'

### Expected results

100% Translated (into Japanese)

### Actual results

Appears in English

### Additional information

1. When in ja google-chrome, to find strings obviously in English, just swapped the ~/ja/messages.js with dummy ja translation and rebuild the UI
2. Screenshot 1 highlighting the unmarked string when in ja google-chrome
![Screenshot from 2021-08-30 18-24-00_1](https://user-images.githubusercontent.com/12151785/131342306-0ddc23ec-23cd-4a02-b245-17e9a92e0247.png)

AlexSCorey:
This is a default string that comes from Patternfly.  We can take advantage of the `noResultsFound` [prop](https://www.patternfly.org/v4/components/select#select) to utilize this string and to translate it properly.  Whoever picks up this issue should audit the rest of the `Select` components and fix them as needed.  I'll bet this is an issue in many places in awx.

PR #11011, opened 2021-09-02
Add strings to be translated

nixocio:
Add strings to be translated

See: https://github.com/ansible/awx/issues/10971

### DIFF
--- a/awx/ui/src/screens/Dashboard/DashboardGraph.js
+++ b/awx/ui/src/screens/Dashboard/DashboardGraph.js
@@ -103,6 +103,7 @@ function DashboardGraph() {
             }}
             selections={periodSelection}
             isOpen={isPeriodDropdownOpen}
+            noResultsFoundText={t`No results found`}
           >
             <SelectOption key="month" value="month">
               {t`Past month`}
--- a/awx/ui/src/screens/Metrics/Metrics.js
+++ b/awx/ui/src/screens/Metrics/Metrics.js
@@ -190,7 +190,7 @@ function Metrics() {
                         setRenderedData([]);
                       }}
                       selections={instance}
-                      placeholderText={t`Select a instance`}
+                      placeholderText={t`Select an instance`}
                     >
                       {instances.map((inst) => (
                         <SelectOption value={inst} key={inst} />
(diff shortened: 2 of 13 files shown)


awx/ui/src/screens/Metrics/Metrics.js
```diff
@@ -191,6 +191,7 @@ function Metrics() {
                       }}
                       selections={instance}
                       placeholderText={t`Select a instance`}
+                      noResultsFoundText={t`No results found`}
```
AlexSCorey: I don't think we need this prop here because it isn't an input/select typeahead

awx/ui/src/screens/Metrics/Metrics.js
```diff
@@ -191,6 +191,7 @@ function Metrics() {
                       }}
                       selections={instance}
                       placeholderText={t`Select a instance`}
```
AlexSCorey:
```suggestion
                      placeholderText={t`Select an instance`}
```

keithjgrant (approved):

awx/ui/src/screens/Dashboard/DashboardGraph.js
```diff
@@ -129,6 +130,7 @@ function DashboardGraph() {
             }}
             selections={jobTypeSelection}
             isOpen={isJobTypeDropdownOpen}
+            noResultsFoundText={t`No results found`}
```
AlexSCorey: We probably don't need this here.

awx/ui/src/screens/Metrics/Metrics.js
```diff
@@ -211,6 +212,7 @@ function Metrics() {
                       }}
                       onToggle={setMetricIsOpen}
                       selections={metric}
+                      noResultsFoundText={t`No results found`}
```
AlexSCorey: same as above

awx/ui/src/screens/Template/Survey/SurveyReorderModal.js
```diff
@@ -136,6 +136,7 @@ function SurveyReorderModal({
             typeAheadAriaLabel={t`Multiple Choice`}
             placeholderText={q.default}
             onToggle={() => {}}
+            noResultsFoundText={t`No results found`}
```
AlexSCorey: We don't need this here because this is always going to be a disabled select.

AlexSCorey (commented):
There are some `<Select/>` that don't need the `noResultsFoundText` string because they are not typeahead, or typeaheadmulti and thus do not have an input.  I don't think we should pass unused/unneeded props to those components.

AlexSCorey (approved):
Looks good.  Well done!

Issue #10977, opened 2021-08-30
Job Output details page briefly showing invalid date


nixocio:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

Job Output details page briefly showing invalid date

### AWX version

devel

### Installation method

docker for mac

### Modifications

no

### Steps to reproduce

1 - Launch a job
2 - Go to job details

### Expected results

`Finished` to be shown once there is a valid date.

### Actual results

See `Finished`

<img width="1492" alt="image" src="https://user-images.githubusercontent.com/9053044/131387942-914d7f0c-a619-4291-8d88-eef12afc5216.png">

PR #11013, opened 2021-09-02
Display finished date once the job is finished

nixocio:
Display finished date once the job is finished.

See: https://github.com/ansible/awx/issues/10977


AlexSCorey (approved):
Looks good.

Issue #11019, opened 2021-09-03
Explicitly pass in UUID for new OCP instances


jladdjr:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

In https://github.com/ansible/awx/pull/10995, `awx-manage provision_instance` was updated to support a new option, `--uuid`. This allows one instance to register a separate instance, citing the new UUID it should have. Previously, the `provision_instance` command would erroneously default to using _its own_ UUID when registering an instance.

This ticket is for some follow-up work that needs to happen for OCP-specifically, where instance registration follows a different path (see https://github.com/ansible/awx/pull/10995#issuecomment-912608796).

### AWX version

devel

### Installation method

openshift

### Modifications

no

### Steps to reproduce

On OCP, run a new job, using an (OCP) container group to execute the job.

### Expected results

When the instance is registered, it should receive a unique UUID.

### Actual results

I'm actually unable to verify the current behavior (we're hitting some roadblocks with provisioning OCP instances, atm). This might actually work already, but we want to make sure that we're _explicitly_ passing in expected uuid via the new support for passing in UUID directly (both to `provision_instance` as well as to the under-lying helper methods involved - for OCP, this is `get_or_register`).

jladdjr:
Marking as blocked given that jobs for deploying OCP are hitting build issues atm.

AlanCoding:
Can we not just do this?

```diff
diff --git a/awx/main/managers.py b/awx/main/managers.py
index eaee03ed7e..d5386a1d2a 100644
--- a/awx/main/managers.py
+++ b/awx/main/managers.py
@@ -162,9 +162,9 @@ class InstanceManager(models.Manager):
 
             pod_ip = os.environ.get('MY_POD_IP')
             if settings.IS_K8S:
-                registered = self.register(ip_address=pod_ip, node_type='control')
+                registered = self.register(ip_address=pod_ip, node_type='control', uuid=settings.SYSTEM_UUID)
             else:
-                registered = self.register(ip_address=pod_ip)
+                registered = self.register(ip_address=pod_ip, uuid=settings.SYSTEM_UUID)
[...]

PR #11021, opened 2021-09-03
Explicitly pass in UUID to get_or_register

jladdjr:
see https://github.com/ansible/awx/issues/11019


AlanCoding (approved):

jladdjr:
Note: Deploying OCP is having some hiccups, atm. Will verify once that is cleared up.

PR #11022, opened 2021-09-03
Set python to ansible_playbook_python on hosts

kdelee:
see 9d000a7

This change works around the fact that the presumed correct python3 for rhel8 (which the EE is based on)
is not the python3 that ansible-playbook is using, and is not where the python dependencies are installed.


matburt (approved):
👍🏻

AlanCoding (approved):

PR #11034, opened 2021-09-07
Assure consistent ordering with default IG first

AlanCoding:
##### SUMMARY
I have been hitting a mismatch between expectation  / outcome

 - expectation: jobs run on "default" instance group by default
 - outcome: jobs run on "controlplane" instance group by default

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
For traditional installs, we only recently started creating the default instance group to begin with.

In many cases, the "default" and "controlplane" will have the same instance groups.

I cannot tell what the InstanceGroup default ordering should be by looking at this code @jbradberry do you know?

### DIFF
--- a/awx/main/models/unified_jobs.py
+++ b/awx/main/models/unified_jobs.py
@@ -1438,7 +1438,7 @@ class UnifiedJob(
         default_instance_groups = list(InstanceGroup.objects.filter(name__in=default_instance_group_names))
 
         # assure deterministic precedence by making sure the default group is first
-        if default_instance_groups[0].name != settings.DEFAULT_EXECUTION_QUEUE_NAME:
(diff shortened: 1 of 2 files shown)


jbradberry:
It appears to me that the `InstanceGroup` model does not have a default ordering imposed upon it.  The inner `Meta` class only has an `app_label` assigned.

That said, I think this code would be conceptually clearer if you just did individual `InstanceGroup.objects.get(name=...)` queries and pushed each onto the list as you get them.

AlanCoding:
I'm hesitant to add more queries because this code will be called for _most_ jobs. Worse than that, it will be called for each pending job in the task manager. That's very nearly the metric on which @chrismeyersfsu worked so hard to optimize the task manager. This can't hit a cache on a related manager or anything like that, and any time this executes _at all_ it's a redundant query, because all IGs have already been loaded into the in-memory graph in the task manager.

jbradberry (approved):

Issue #11036, opened 2021-09-07
Pool max_workers based on host memory when containerized instead of container memory limit


kimbernator:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

I noticed that when a large number of jobs were submitted in rapid succession, my awx-task container keeps spinning up worker processes to run jobs and eventually has more running than the memory limits on the container can handle, causing workers to die off. Logs indicate that jobs handled by them would be re-queued, but they never are (presumably the system has no memory to do that kind of thing). Jobs randomly start getting reaped, then they get marked with the status discussed in https://github.com/ansible/awx/issues/9594. I don't think this is the sole reason for that error, but in my case it appears to be one possible source of it.

The problem comes from the fact that `AutoscalePool` attempts to calculate what it believes a good upper limit on workers would be based on the result of `psutil.virtual_memory().total`. This is a problem because within containers, this returns the total virtual memory for the host that the container is running on. So if my container has a limit of 2G of memory and the host has 8G, AWX will run `(psutil.virtual_memory().total >> 30) + 1` and get 9GB (from rounding), then multiply that by 5 to get a max of 45 worker processes. Based on what I've seen, each will take a minimum of 150MB (but can go higher) which results in a minimum of (45*150MB) or 6.7GB of memory budgeted for these workers.

There is the option of defining `SYSTEM_TASK_ABS_MEM` to override this calculation, but either I missed the documentation suggesting that or it's not well-documented.

### AWX version

19.3.0

### Installation method

kubernetes

### Modifications

no

### Ansible version

latest

### Operating system

not applicable, running in the official awx container.

### Steps to reproduce

Set a memory limit for the awx-task container then submit a lot of jobs at once

### Expected results

jobs get queued and respect the resource limits assigned to the containers

### Actual results

many more jobs than would be appropriate attempt to run at once, crashing queue workers and losing jobs.

ffirg:
This should be fixed via https://github.com/ansible/awx/issues/11640 so closing. Please re-open if still a problem.

PR #11038, opened 2021-09-07
Check cgroup memory limit when determining max queue workers

kimbernator:
##### SUMMARY

related to https://github.com/ansible/awx/issues/11036

This checks for the existence of and reads from /sys/fs/cgroup/memory/memory.limit_in_bytes when determining the upper limit on how many workers can be running at once.  As stated in the related issue, `psutil.virtual_memory().total` will read the total virtual memory of the host running the container instead of the limit assigned to the container itself.

Since it's not just containers that have /sys/fs/cgroup/memory/memory.limit_in_bytes, this checks both sources and takes the lowest of the two. It's possible that I have a blind spot here and that this isn't sufficient to ensure a good outcome, but it should be applicable for anyone running AWX in kubernetes.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- main/dispatch/pool.py

##### AWX VERSION

```
awx: 19.3.0
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/dispatch/pool.py
+++ b/awx/main/dispatch/pool.py
@@ -321,7 +321,14 @@ class AutoscalePool(WorkerPool):
             if settings_absmem is not None:
                 total_memory_gb = int(settings_absmem)
             else:
-                total_memory_gb = (psutil.virtual_memory().total >> 30) + 1  # noqa: round up
+                # Get the total virtual memory of the host
+                memory_limits = [psutil.virtual_memory().total]
+                if os.path.isfile('/sys/fs/cgroup/memory/memory.limit_in_bytes'):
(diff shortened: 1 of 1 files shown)


kdelee:
@shanemcd can correct me if I'm wrong, But I think this is now not needed since https://github.com/ansible/awx/pull/11725 merged. You can now either set resource limits on the task container or set the setting `SYSTEM_TASK_ABS_MEM` to achieve this purpose. We decided this would be less prone problems with changes in how these files are laid out in `/sys/fs/cgroup/memory/` which we observed some subtle differences depending on your k8s cluster versions etc.

PR #11062, opened 2021-09-09
Collection version change

john-westcott-iv:
##### SUMMARY

When pointing the collection to older versions of AWX an error can occur if the version header is not present:
```
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'awx.awx.controller_api'. Error was a <class 'AttributeError'>, original message: 'LooseVersion' object has no attribute 'version'"}
```

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - Collection
##### AWX VERSION

```
awx: 19.3.0
```

##### ADDITIONAL INFORMATION

Before Change:
```
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'awx.awx.controller_api'. Error was a <class 'AttributeError'>, original message: 'LooseVersion' object has no attribute 'version'"}
```

After Change:
```
[WARNING]: You are using the 19.2.2 version of this collection but connecting to a controller that did not return a version
```

### DIFF
--- a/awx_collection/plugins/module_utils/controller_api.py
+++ b/awx_collection/plugins/module_utils/controller_api.py
@@ -263,9 +263,9 @@ class ControllerAPIModule(ControllerModule):
             parsed_collection_version = Version(self._COLLECTION_VERSION).version
             if not controller_version:
                 self.warn(
-                  "You are using the {0} version of this collection but connecting to a controller that did not return a version".format(
-                    self._COLLECTION_VERSION
-                  )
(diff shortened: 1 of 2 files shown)


AlanCoding (approved):
Nice, thank you very much.

I would also like to properly handle the migration-in-progress page (mention because it is a similar problem in a way). This should be a huge improvement.

beeankha (approved):

Issue #10442, opened 2021-06-14
Add some kind of flag to indicate the global default EEs


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
See the UI for Execution Environments, it's a list

![Screenshot from 2021-06-14 13-59-58](https://user-images.githubusercontent.com/1385596/121937634-e16f2980-cd18-11eb-9a4e-58edb82f9248.png)

We should have a concrete visual indicator (other than the name) for the EE which will be used for project updates, and the EE that will be the default for jobs.

The system administrator can change the default for jobs, so the flag should be responsive to this.

The API does not return data sufficient for this yet.

wenottingham:
@AlanCoding @mabashian what is editiable for the control plane EE? Should it be shown in the list at all?

AlanCoding:
Project updates show a link to their EE, which would be the control plane EE. I would be careful about "hiding" it, because people can still click through to see it.

Could the UI make the control plane EE non-editable? That, plus a badge, would make this the most clear, IMO.

nixocio:
@wenottingham @AlanCoding changes related to the controlplane on the UI side were addressed here - https://github.com/ansible/awx/issues/10396. 

API RBAC for those EEs. 

<img width="977" alt="image" src="https://user-images.githubusercontent.com/9053044/121953569-b93cf600-cd2b-11eb-922b-553d0cba4103.png">

 Right now we do not allow one to delete the EE that has summary fields delete set as false -

<img width="1466" alt="image" src="https://user-images.githubusercontent.com/9053044/121954376-ad9dff00-cd2c-11eb-9554-8f1ef12c7e8e.png">

 When attempting to modify `name` or `image` the following error is raised to EE managed by tower.

<img width="1452" alt="image" src="https://user-images.githubusercontent.com/9053044/121954133-64e64600-cd2c-11eb-912a-369c77c73457.png">

AlanCoding:
Yes, those are solutions to some of those problems.

The point of this issue is more - how do I know that "AWX EE 0.3.0" is the functional one that jobs default to using, as opposed to "AWX EE 0.4.0"

AlexSCorey:
Seems like a great decision for @trahman73 to help us with.

trahman73:
One solution could be that we add an additional column to the table (and possibly the filter) to distinguish the different types of EEs.

AlexSCorey:
[...]

PR #11063, opened 2021-09-09
Add is_global_default to Execution Environment API

AlanCoding:
##### SUMMARY
Connect https://github.com/ansible/awx/issues/10442

This is intended to solve that without degrading performance of the `/api/v2/execution_environments/` endpoint. A lot of what's going on here are optimizations for that endpoint.

The intent is that the UI will make use of this field to indicate in the list views which EE is the default.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API


awx/main/managers.py
```diff
@@ -271,3 +273,19 @@ def capacity_values(self, qs=None, tasks=None, breakdown=False, graph=None):
...
+        if settings.DEFAULT_EXECUTION_ENVIRONMENT is not None:
+            return settings.DEFAULT_EXECUTION_ENVIRONMENT
+        installed_default = self.model.objects.filter(
+            image__in=[ee['image'] for ee in settings.GLOBAL_JOB_EXECUTION_ENVIRONMENTS], organization=None, managed=False
+        ).first()
+        if installed_default:
+            return installed_default
+        return self.model.objects.filter(organization=None, managed=False).first()
+
+    default_execution_environment = cached_property(get_default_execution_environment, name='default_execution_environment')
```
AlanCoding, line 291: @jbradberry Picking up the conversation from:

https://github.com/ansible/awx/pull/11034#issuecomment-914723421

This is the pattern I would prefer to use for these static, global, objects. We can use the cached property with abandon, because it will never accumulate queries. This is good if the method we're writing is going to be used incidentally deep inside the task manager.

jbradberry: I think that using `cached_property` is going to be too "sticky".  What happens when the sysadmin changes `settings.DEFAULT_EXECUTION_ENVIRONMENT` or makes other changes that would affect which EE is considered the global default?

Also, this cache is just an attribute stashed on an instantiation of the custom manager.  This value isn't going to be shared amongst different server processes, and even within a single process may go out of scope and need to be re-queried unexpectedly.

I'm not sure which of these potential problems would be the bigger annoyance, or whether they would wind up mitigating each other in practice.

AlanCoding: So far, the only other caching option I've thought about is stashing it in the view context (so list views won't re-query same object 25x). That's going to get messy across multiple view, and it's not quite as aggressive as I'd like. If the task manager grabbed this once for all pending jobs, I'd like all but 1 to use a cache.

It might be best to allow it fall back to the actual cache via the settings. But I don't really know what the ForeignKey mechanics are for that, and the `settings.DEFAULT_EXECUTION_ENVIRONMENT` is only one of the code paths that it might ultimately end up on.

jbradberry: We could use Django's caching system directly, and stash the pk of the resolved global EE in whatever our backing store is.  A signal handler on EEs and on the `settings.DEFAULT_EXECUTION_ENVIRONMENT` value could then invalidate the cache when something changes.

AlexSCorey (commented):
This looks good to me.  To confirm the flag the UI will use to key the icon on the list will is `is_global_default`, right?

AlanCoding:
@AlexSCorey right

AlexSCorey (approved):
Looks good to me

AlanCoding:
Closing for someone else to revive as per advice:

https://github.com/ansible/awx/pull/11063#discussion_r706389940

Issue #10944, opened 2021-08-26
Fix EEs from protected registries on execution-only nodes


shanemcd:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

![image](https://user-images.githubusercontent.com/773186/130968122-86616704-a64a-4836-92c0-f10117834c18.png)

### AWX version

devel sometime on 8/25

### Installation method

openshift

### Modifications

no

### Steps to reproduce

Try to run a job using an EE from an auth-protected registry.

### Expected results

💥 (see screenshot above)

### Actual results

🚀 😎

AlanCoding:
Changes on the AWX side are code-complete.

We have 1 straggler on the runner side https://github.com/ansible/ansible-runner/pull/829, _but_, I don't think that affects any use case that integration testing currently hits (I only know of a means to _manually_ test). Anyway, it is expected that that should land by EOD today.

That's a bit of qualifier, but nonetheless moving to needs_test

kdelee:
@AlanCoding That PR is still hanging out. I also know we are waiting on a few more PR's going into runner. Whats up with that?

beeankha:
@kdelee we are working on getting in-progress Runner PRs merged in by hopefully the middle of next week! The main Runner PR related to this issue (https://github.com/ansible/ansible-runner/pull/817) has been merged; the remaining WIP stuff is not related to this particular "protected EE registries" issue.

Zokormazo:
This is working properly now

PR #11064, opened 2021-09-09
Enable Jobs to Run on Execution-Only Nodes Via EEs from Protected Registries

beeankha:
##### SUMMARY

Connecting issue https://github.com/ansible/awx/issues/10944

This PR updates the way auth info is passed to Ansible Runner when pulling Execution Environments from protected registries.
Related Ansible Runner PR: https://github.com/ansible/ansible-runner/pull/817

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.3.0
```

### DIFF
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -939,23 +939,17 @@ class BaseTask(object):
         params = {
             "container_image": image,
             "process_isolation": True,
+            "process_isolation_executable": "podman",  # need to provide, runner enforces default via argparse
             "container_options": ['--user=root'],
         }
 
         if instance.execution_environment.credential:
             cred = instance.execution_environment.credential
             if cred.has_inputs(field_names=('host', 'username', 'password')):
-                path = os.path.split(private_data_dir)[0]
-                with open(path + '/auth.json', 'w') as authfile:
-                    os.chmod(authfile.name, stat.S_IRUSR | stat.S_IWUSR)
-
-                    host = cred.get_input('host')
-                    username = cred.get_input('username')
-                    password = cred.get_input('password')
-                    token = "{}:{}".format(username, password)
-                    auth_data = {'auths': {host: {'auth': b64encode(token.encode('UTF-8')).decode('UTF-8')}}}
-                    authfile.write(json.dumps(auth_data, indent=4))
-                params["container_options"].append(f'--authfile={authfile.name}')
+                host = cred.get_input('host')
+                username = cred.get_input('username')
+                password = cred.get_input('password')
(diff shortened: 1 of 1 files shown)


AlanCoding (approved):

awx/main/tasks.py
```diff
@@ -3049,7 +3043,7 @@ def __init__(self, task, runner_params=None):
 
         if self.task and not self.task.instance.is_container_group_task:
             execution_environment_params = self.task.build_execution_environment_params(self.task.instance, runner_params['private_data_dir'])
-            self.runner_params['settings'].update(execution_environment_params)
+            self.runner_params.update(execution_environment_params)
```
jbradberry, line 3047: Is this ok?  We're changing the level in the data structure where we're injecting the EE params.

AlanCoding: Yeah, a real critical non-obvious thing here is:

https://github.com/ansible/ansible-runner/blob/205c4940e44ae0cdbe418c2b4bcf3d2f312388eb/ansible_runner/config/_base.py#L185

I'm not sure if allowing either was a stroke of genius or complete confusion.

jladdjr: Instead of having ansible-runner collect parameters as if they were settings, we're making use of the runner's python API _directly_:
https://github.com/beeankha/ansible-runner/blob/a991e968be08f31182700b0390541ae4e054be5c/ansible_runner/config/_base.py#L66

> I'm not sure if allowing either was a stroke of genius or complete confusion.

I echo this 100%. I guess runner's python API accepts both a `settings` argument as well as the individual settings as kwargs in order to distinguish where they came from?

jladdjr (approved):

awx/main/tasks.py
```diff
@@ -939,23 +939,17 @@ def build_execution_environment_params(self, instance, private_data_dir):
...
-                    username = cred.get_input('username')
-                    password = cred.get_input('password')
-                    token = "{}:{}".format(username, password)
-                    auth_data = {'auths': {host: {'auth': b64encode(token.encode('UTF-8')).decode('UTF-8')}}}
-                    authfile.write(json.dumps(auth_data, indent=4))
-                params["container_options"].append(f'--authfile={authfile.name}')
+                host = cred.get_input('host')
+                username = cred.get_input('username')
+                password = cred.get_input('password')
+                params['container_auth_data'] = {'host': host, 'username': username, 'password': password}
```
AlanCoding: If we merge https://github.com/ansible/ansible-runner/pull/829, then we need to add `verify_ssl` to this.

PR #11073, opened 2021-09-13
fix workflow_job_template related field type specification

kimbernator:
##### SUMMARY

The documentation currently states that the workflow_job_template schema `related` field should be a dict which contains 4 possible keys (success_nodes/failure_nodes/credentials/always_nodes) with values that are arrays of strings, despite the example on the same page being an array of dicts:

image

This is further validated by looking at the logic in the module itself: https://github.com/ansible/awx/blob/devel/awx_collection/plugins/modules/workflow_job_template.py#L616 and seeing that it does indeed check for a key in the value instead of reading the value directly.

##### ISSUE TYPE

 - Docs Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
N/A
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx_collection/plugins/modules/workflow_job_template.py
+++ b/awx_collection/plugins/modules/workflow_job_template.py
@@ -277,42 +277,46 @@ options:
                 - Nodes that will run after this node completes.
                 - List of node identifiers.
               type: list
+              elements: dict
               suboptions:
                 identifier:
                   description:
                     - Identifier of Node that will run after this node completes given this option.
-                  elements: str
+                  type: str
             success_nodes:
               description:
                 - Nodes that will run after this node on success.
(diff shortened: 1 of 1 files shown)


AlanCoding:
Ping @sean-m-sullivan, can I hit you up for a review?

sean-m-sullivan (approved):
This looks right, looks like an oversight on my part when constructing the documentation. It's supposed to be a mirror of what the export module exports for workflows, which in itself is a mirror of the node entries in the API.

djyasin:
Thank you so much for contributing!! We are so sorry it took us so long to get back to you. 

We are happy to confirm this has been resolved.  Thank you again for taking the time to contribute!

PR #11074, opened 2021-09-13
prevent duplicate UUIDs from being created and allow users to update hostnames based on uuid

rebeccahhh:
##### SUMMARY

if the user provides a uuid and it exists, allow that to tie to the instance, which allows the user to update the instance based on the UUID (includeding updating the hostname) should they choose to do so.

##### ISSUE TYPE

 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.3.0
```

##### ADDITIONAL INFORMATION

```
awx-manage provision_instance
```
##### TODO
---
- [ ] Modify the installer to make the `system_uuid` consistent (see discussion below)

### DIFF
--- a/awx/main/managers.py
+++ b/awx/main/managers.py
@@ -129,24 +129,25 @@ class InstanceManager(models.Manager):
                         other_inst.save(update_fields=['ip_address'])
                         logger.warning("IP address {0} conflict detected, ip address unset for host {1}.".format(ip_address, other_hostname))
 
-            # get the instance based on the hostname
-            instance = self.filter(hostname=hostname)
-
-            # Check or update the existing uuid
-            if uuid is not None and uuid != UUID_DEFAULT:
-                if self.filter(uuid=uuid).exists():
-                    instance = self.filter(uuid=uuid)
+            # Return existing instance that matches hostname or UUID (default to UUID)
+            if uuid is not None and uuid != UUID_DEFAULT and self.filter(uuid=uuid).exists():
+                instance = self.filter(uuid=uuid)
+            else:
+                # if instance was not retrieved by uuid and hostname was, use the hostname
+                instance = self.filter(hostname=hostname)
 
             # Return existing instance
             if instance.exists():
-                instance = instance.get()
+                instance = instance.first()  # in the unusual occasion that there is more than one, only get one
                 update_fields = []
+                # if instance was retrieved by uuid and hostname has changed, update hostname
                 if instance.hostname != hostname:
+                    logger.warning("passed in hostname {0} is different from the original hostname {1}, updating to {0}".format(hostname, instance.hostname))
                     instance.hostname = hostname
                     update_fields.append('hostname')
+                # if any other fields are to be updated
                 if instance.ip_address != ip_address:
                     instance.ip_address = ip_address
-                    update_fields.append('ip_address')
                 if instance.node_type != node_type:
                     instance.node_type = node_type
                     update_fields.append('node_type')
--- a/awx/main/managers.py
+++ b/awx/main/managers.py
@@ -12,9 +12,10 @@ from awx.main.utils.pglock import advisory_lock
 from awx.main.utils.common import get_capacity_type
 from awx.main.constants import RECEPTOR_PENDING
 
-___all__ = ['HostManager', 'InstanceManager', 'InstanceGroupManager', 'DeferJobCreatedManager']
+___all__ = ['HostManager', 'InstanceManager', 'InstanceGroupManager', 'DeferJobCreatedManager', 'UUID_DEFAULT']
 
 logger = logging.getLogger('awx.main.managers')
+UUID_DEFAULT = '00000000-0000-0000-0000-000000000000'
 
 
 class DeferJobCreatedManager(models.Manager):
@@ -105,7 +106,7 @@ class InstanceManager(models.Manager):
         """Return the currently active instance."""
         # If we are running unit tests, return a stub record.
         if settings.IS_TESTING(sys.argv) or hasattr(sys, '_called_from_test'):
-            return self.model(id=1, hostname=settings.CLUSTER_HOST_ID, uuid='00000000-0000-0000-0000-000000000000')
+            return self.model(id=1, hostname=settings.CLUSTER_HOST_ID, uuid=UUID_DEFAULT)
 
         node = self.filter(hostname=settings.CLUSTER_HOST_ID)
         if node.exists():
@@ -115,6 +116,7 @@ class InstanceManager(models.Manager):
     def register(self, uuid=None, hostname=None, ip_address=None, node_type='hybrid', defaults=None):
         if not hostname:
             hostname = settings.CLUSTER_HOST_ID
+
         with advisory_lock('instance_registration_%s' % hostname):
             if settings.AWX_AUTO_DEPROVISION_INSTANCES:
                 # detect any instances with the same IP address.
@@ -127,11 +129,21 @@ class InstanceManager(models.Manager):
                         other_inst.save(update_fields=['ip_address'])
                         logger.warning("IP address {0} conflict detected, ip address unset for host {1}.".format(ip_address, other_hostname))
 
-            # Return existing instance that matches hostname
+            # get the instance based on the hostname
             instance = self.filter(hostname=hostname)
+
+            # Check or update the existing uuid
+            if uuid is not None and uuid != UUID_DEFAULT:
+                if self.filter(uuid=uuid).exists():
+                    instance = self.filter(uuid=uuid)
+
+            # Return existing instance
             if instance.exists():
                 instance = instance.get()
                 update_fields = []
+                if instance.hostname != hostname:
+                    instance.hostname = hostname
+                    update_fields.append('hostname')
                 if instance.ip_address != ip_address:
                     instance.ip_address = ip_address
                     update_fields.append('ip_address')
--- a/awx/main/models/ha.py
+++ b/awx/main/models/ha.py
@@ -18,7 +18,7 @@ from solo.models import SingletonModel
 
 from awx import __version__ as awx_application_version
 from awx.api.versioning import reverse
-from awx.main.managers import InstanceManager, InstanceGroupManager
+from awx.main.managers import InstanceManager, InstanceGroupManager, UUID_DEFAULT
 from awx.main.fields import JSONField
 from awx.main.models.base import BaseModel, HasEditsMixin, prevent_search
 from awx.main.models.unified_jobs import UnifiedJob
@@ -59,7 +59,7 @@ class Instance(HasPolicyEditsMixin, BaseModel):
     objects = InstanceManager()
 
     # Fields set in instance registration
-    uuid = models.CharField(max_length=40, default='00000000-0000-0000-0000-000000000000')
+    uuid = models.CharField(max_length=40, default=UUID_DEFAULT)
     hostname = models.CharField(max_length=250, unique=True)
     ip_address = models.CharField(
         blank=True,


awx/main/managers.py
```diff
@@ -101,11 +101,13 @@ class InstanceManager(models.Manager):
     instance or role.
     """
 
+    UUID_DEFAULT = '00000000-0000-0000-0000-000000000000'
```
AlanCoding: you should combine this with

https://github.com/ansible/awx/blob/6a17e5b65bf9421c2e650d23817c79ce81248c6f/awx/main/models/ha.py#L62

awx/main/managers.py
```diff
@@ -127,11 +130,21 @@ def register(self, uuid=None, hostname=None, ip_address=None, node_type='hybrid'
                         other_inst.save(update_fields=['ip_address'])
                         logger.warning("IP address {0} conflict detected, ip address unset for host {1}.".format(ip_address, other_hostname))
 
-            # Return existing instance that matches hostname
+            # get the instance based on the hostname
             instance = self.filter(hostname=hostname)
+
+            # Check or update the existing uuid
+            if uuid != None and uuid != self.UUID_DEFAULT:
```
jbradberry: this should be written as `uuid is not None...`

awx/main/managers.py
rebeccahhh: you're right, good call.

awx/main/managers.py
rebeccahhh: I kept it this way to match the existing if statements below, but I'm not set on that.

sarabrajsingh (approved):

awx/main/managers.py
sarabrajsingh: i conditionally approved, pending these changes!

awx/main/managers.py
```diff
@@ -127,11 +130,21 @@ def register(self, uuid=None, hostname=None, ip_address=None, node_type='hybrid'
...
+                if self.filter(uuid=uuid).exists():
+                    instance = self.filter(uuid=uuid)
```
AlanCoding, line 139: Someone gives both hostname and uuid in the provision_instance command.

There is a hit based on hostname, but there is not a hit based on uuid. First, `instance` is defined to be a non-empty queryset. Then, it is redefined to be an empty queryset.

Later in the code, we find `if instance.exists()` is False and because of that, run `instance = self.create`. The Instance table is still, absolutely, unique on hostname, so it would give an IntegrityError if I read it right.

I could be wrong, but a unit test would be good. I don't know what your expectation is, but that has tie-ins with how the installer is written.

AlanCoding:
@jladdjr has merged related installer changes and I'd like him to make sure this is compatible with what that does.

tchellomello:
To provide context, please assume the following dataset
```
>>> Instance.objects.values('hostname', 'uuid')
<QuerySet [{'hostname': '192.168.111.108', 'uuid': '5ffbd38ae5734483882edbb68ee1a43a'},
 {'hostname': '192.168.111.115', 'uuid': '2b9b8b03-8de8-466a-8f9f-49b0283079f2'}, 
{'hostname': 'controller_p70', 'uuid': '241c61d9-14ca-45fb-be38-ca8d55894e08'},
 {'hostname': 'controller_t470n1', 'uuid': 'e98da768-6fb0-4195-b47d-c85aba7155df'},
 {'hostname': 'node_p70', 'uuid': 'e82814c6df6e4db6a93beb6c041f76a5'}]>
>>> Instance.objects.values('hostname', 'uuid').count()
5

```
Then testing:

```
[root@126-addr ~]# awx-manage  provision_instance --hostname='controller_p70' --node_type='control' --uuid='241c61d9-14ca-45fb-be38-ca8d55894e08' 
Instance already registered controller_p70
[root@126-addr ~]# awx-manage  provision_instance --hostname='controller_p70' --node_type='control' --uuid='241c61d9-14ca-45fb-be38-ca8d55NEW' 
Instance already registered controller_p70
[root@126-addr ~]# awx-manage  provision_instance --hostname='controller_p70_renamed' --node_type='control' --uuid='241c61d9-14ca-45fb-be38-ca8d55894e08' 
Successfully registered instance controller_p70_renamed
(changed: True)
[root@126-addr ~]# awx-manage  provision_instance --hostname='controller_p70_renamed' --node_type='control'
Instance already registered controller_p70_renamed
[root@126-addr ~]# awx-manage  provision_instance --hostname='controller_no_uuid' --node_type='control'
Successfully registered instance controller_no_uuid
(changed: True)
```

Which provides:
```
>>> Instance.objects.values('hostname', 'uuid').count()
6
>>> Instance.objects.values('hostname', 'uuid')
<QuerySet [{'hostname': '192.168.111.108', 'uuid': '5ffbd38ae5734483882edbb68ee1a43a'},
 {'hostname': '192.168.111.115', 'uuid': '2b9b8b03-8de8-466a-8f9f-49b0283079f2'}, 
{'hostname': 'controller_no_uuid', 'uuid': '00000000-0000-0000-0000-000000000000'}, 
{'hostname': 'controller_p70_renamed', 'uuid': '241c61d9-14ca-45fb-be38-ca8d55894e08'},
 {'hostname': 'controller_t470n1', 'uuid': 'e98da768-6fb0-4195-b47d-c85aba7155df'}, 
{'hostname': 'node_p70', 'uuid': 'e82814c6df6e4db6a93beb6c041f76a5'}]>
```

awx/main/managers.py
rebeccahhh: is that along the lines of what you were thinking @AlanCoding ?

awx/main/managers.py
rebeccahhh: updated, cheers y'all

awx/main/managers.py
```diff
@@ -127,11 +129,21 @@ def register(self, uuid=None, hostname=None, ip_address=None, node_type='hybrid'
                         other_inst.save(update_fields=['ip_address'])
                         logger.warning("IP address {0} conflict detected, ip address unset for host {1}.".format(ip_address, other_hostname))
 
-            # Return existing instance that matches hostname
+            # get the instance based on the hostname
             instance = self.filter(hostname=hostname)
+
+            # Check or update the existing uuid
```
jladdjr: It seems like the basic logic here is "default to looking up instance by uuid, if one is provided. otherwise, look up by hostname". Could make that more explicit with:

```
            # attempt to find instance by uuid
            if uuid is not None and uuid != UUID_DEFAULT and self.filter(uuid=uuid).exists():
                instance = self.filter(uuid=uuid)
            else:
                # otherwise, rely on hostname
                instance = self.filter(hostname=hostname)
```

jladdjr: `update the existing uuid` doesn't seem accurate, since we're not updating the uuid here.

awx/main/managers.py
```diff
@@ -127,11 +129,21 @@ def register(self, uuid=None, hostname=None, ip_address=None, node_type='hybrid'
...
+            if uuid is not None and uuid != UUID_DEFAULT:
+                if self.filter(uuid=uuid).exists():
+                    instance = self.filter(uuid=uuid)
+
+            # Return existing instance
             if instance.exists():
                 instance = instance.get()
                 update_fields = []
+                if instance.hostname != hostname:
```
jladdjr, line 144: Would add something like this to call out what's happening here:
`# if instance was retrieved by uuid and hostname has changed, update hostname`

jladdjr: Would add a `logger.warn` here so that the logs reflect the hostname change happened.

jladdjr (approved):
Makes sense to make the `provision_instance` code a bit more defensive, so that `uuid` wins over the `hostname` and we update the hostname if a new one is provided 👍

awx/main/managers.py
AlanCoding: what if there is a match based on both hostname & uuid.

In that case, you would prefer the uuid, but... you would change the hostname, and instances are unique on hostname. This would probably error.

This pattern (of creating fallback lookup fields in the midst of unique fields) and the errors that come with it are familiar to me from the inventory import `instance_id` field. If the precedence ever matters, then you're left with an IntegrityError.

awx/main/managers.py
```diff
@@ -127,11 +129,21 @@ def register(self, uuid=None, hostname=None, ip_address=None, node_type='hybrid'
...
             instance = self.filter(hostname=hostname)
+
+            # Check or update the existing uuid
+            if uuid is not None and uuid != UUID_DEFAULT:
+                if self.filter(uuid=uuid).exists():
+                    instance = self.filter(uuid=uuid)
+
+            # Return existing instance
             if instance.exists():
                 instance = instance.get()
```
jbradberry: I think I had started to say something about this previously but got distracted.  This should probably be written a bit more defensively, since it will still get here and will error if there are multiple `Instance` objects pulled in by the queryset.  The most obvious option would be to use `.first()` instead of `.get()`, but we could also do something like with the conflicting ip address detection towards the top of this method instead.

awx/main/managers.py
rebeccahhh: so currently this will change the hostname if it's different and both a UUID and hostname are provided, I can see why that would be problematic, the solution would be to alert the user and verify that they want to change the hostname, but I can also see that getting very annoying very quickly.

awx/main/managers.py
AlanCoding: oh yeah, looks great, sorry for the delayed reply.

awx/main/managers.py
```diff
@@ -127,14 +129,23 @@ def register(self, uuid=None, hostname=None, ip_address=None, node_type='hybrid'
                         other_inst.save(update_fields=['ip_address'])
                         logger.warning("IP address {0} conflict detected, ip address unset for host {1}.".format(ip_address, other_hostname))
 
-            # Return existing instance that matches hostname
-            instance = self.filter(hostname=hostname)
+            # Return existing instance that matches hostname or UUID (default to UUID)
+            if uuid is not None and uuid != UUID_DEFAULT and self.filter(uuid=uuid).exists():
+                instance = self.filter(uuid=uuid)
```
AlanCoding, line 134: Before I approve, I want to echo a concern for which discussions happened in other channels.

My reading of the installer before this point was that they would generate a new uuid every run (different from real uuid if an existing host exists). I would like to be sure that doesn't happen for re-runs of the installer, otherwise I get nervous. What makes the _most_ sense to me is for the installer to read the `ha.py` file's SYSTEM_UUID value if it's an existing instance, and generate one only if it's a new instance.

At this moment in code, if `self.filter(hostname=hostname).exists()`, then it will result in an error. I'm fine with that because it should throw an error in that case.

AlanCoding (approved):

AlanCoding:
My current understanding is that this is needed for a fix in conjunction with some other installer changes, and I'd like to see it move along.

awx/main/managers.py
tchellomello: @AlanCoding yes, that makes sense. 

As you mentioned, the installer does generate a new `system_uuid` on every run, but it will only be used if the `ha.py` does not exist, **however**, at the task https://github.com/ansible/automation-platform-collection/blob/devel/roles/receptor/tasks/main.yml#L156-L159, we consume the `system_uuid` which is where the problem could happen. 

We will modify the installer to set the `system_uuid` from the `ha.py` and when `ha.py` then generate it fresh.

AlanCoding: Yeah, if reading from `ha.py` isn't needed for the current fix, and is just a matter of code cleanup, then feel free to file and issue to rework it later and get the currently needed fixes in. I think we're fully on the same page now.

jladdjr (approved):

awx/main/managers.py
tchellomello: @rebeccahhh++  opened the PR https://github.com/ansible/automation-platform-collection/pull/179 which should take of this

awx/main/managers.py
rebeccahhh: added line 145 should address that @jladdjr :)

beeankha (approved):

PR #11077, opened 2021-09-14
Install receptorctl from new nightly url

shanemcd:
We ran into problems with our offline builds with our usage of PBR + subdirectory


simaishi:
@shanemcd Thank you for making the change!

simaishi (approved):

PR #11084, opened 2021-09-15
Mark execution node error strings for translation, avoid unnecessary mark_lost

AlanCoding:
This is a random thing on my TODO list from prior discussions with @AlexSCorey 

The `Instance.errors` field may be surfaced in the UI for errored nodes. Because of that, it is expected that it will be translated for the errors that we control. Projects of receptor & ansible-runner do not have translated versions, so errors from those cannot reasonably be translated. They'll just be put in as-is.

### DIFF
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -452,7 +452,8 @@ class InstanceHealthCheck(GenericAPIView):
                     else:
                         time.sleep(1.0)
                 else:
-                    obj.mark_offline(errors=_('Health check initiated by user determined this instance to be unresponsive'))
+                    if obj.capacity > 0:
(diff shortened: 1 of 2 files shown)


awx/api/views/__init__.py
```diff
@@ -452,7 +452,8 @@ def post(self, request, *args, **kwargs):
                     else:
                         time.sleep(1.0)
                 else:
-                    obj.mark_offline(errors=_('Health check initiated by user determined this instance to be unresponsive'))
+                    if obj.capacity > 0:
+                        obj.mark_offline(errors=_('Health check initiated by user determined this instance to be unresponsive'))
```
AlanCoding, line 456: We also don't want to change `errors` if the instance already started out offline. This is a tricky case, but should be rarely relevant.

PR #11089, opened 2021-09-16
Make some needed updates to docker-refresh target

AlanCoding:
Couple of things were happening:

 - when we removed containers, we removed `tools_awx_x` but now `tools_receptor_x` will use the same image, so if we don't remove those, we can't remove its respective image
 - when you ran `make docker-refresh` two times it a row, it would error due to getting no arguments, added an `if` statement to handle this
 - remove orphans so that I only have `tools_receptor_1`, not 1, 2, 3, etc.


beeankha (approved):
:+1:

Issue #8826, opened 2020-12-10
[ui_next] [UX] Inline form validation is applied when canceling an edit instead of <back> action


unlikelyzero:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
The inline form validation on an add/edit form is applied when hitting the cancel button. This prevents the user from being navigated back on first click of cancel button.

##### ENVIRONMENT
* AWX version: 12/10

##### STEPS TO REPRODUCE
1. Create an organization
2. Click on the 'Name' field
3. Click on the 'Cancel' button
Note: user has to click cancel button twice because form validation is applied initially

##### EXPECTED RESULTS
First cancel press should navigate the user back to previous page

##### ACTUAL RESULTS
inline form validation applied
User navigated back to previous page

##### ADDITIONAL INFORMATION
![Screen Shot 2020-12-10 at 1 39 43 PM](https://user-images.githubusercontent.com/12446869/101815188-2d80ad80-3aed-11eb-8f7a-28c83e304339.png)

AlexSCorey:
Putting this in needs test because I don't think this is a valid issue any more.

tiagodread:
@AlexSCorey this is still a valid bug, look:


https://user-images.githubusercontent.com/12464858/128750953-8c996e9a-3905-420f-a627-85c954e3e180.mp4

PR #11104, opened 2021-09-21
Prevents form validation on cancel button click

AlexSCorey:
##### SUMMARY
This addresses #8826.  There is an argument to be made that the bug is not really a bug but rather it is `Formik `doing what `Formik` does.  The bug only seems to present itself when the user clicks on a required field without entering a value, and then clicking the cancel button. This causes `Formik` to fire off validation because the required field has been touched and it has no value.  

By adding the `onMouseDown` event handler we are rearranging the order that we can prevent the onBlur function to execute via `preventDefault()` inside `onMouseDown`

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATIO
cancel
N

### DIFF
--- a/awx/ui/src/components/FormActionGroup/FormActionGroup.js
+++ b/awx/ui/src/components/FormActionGroup/FormActionGroup.js
@@ -23,6 +23,7 @@ const FormActionGroup = ({ onCancel, onSubmit, submitDisabled }) => (
         aria-label={t`Cancel`}
         variant="link"
         type="button"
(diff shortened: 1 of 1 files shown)


akus062381:
@AlexSCorey , I just built a new local environment in order to test this, and I am not seeing your fix?

akus062381 (approved):

nixocio (approved):

Issue #10785, opened 2021-07-29
System Settings- Default EE- Revert button missing


akus062381:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

When a user accesses the System settings page, and clicks Edit, there are several fields available to make changes to.
Every single field has its own Revert button, except for Global Default Execution Environment. 
So, it may leave the user wondering whether reverting the change is possible at all...
Until the user clicks the "Revert All to Default" button on the Edit screen- and that does, in fact, revert any change made to the Global Default Execution Environment field.
If a user can revert this option- shouldn't it also have its own revert button?

![image](https://user-images.githubusercontent.com/63685497/127525651-33733a01-2f04-4d59-9d80-e013c63bacd6.png)

### AWX version

4.0

### Installation method

docker for mac

### Modifications

no

### Steps to reproduce

Click Settings.
Click Miscellaneous System Settings.
Click Edit.
Assign a Global Default EE, click Save.
Note the change.
Click edit again. Note the absence of a Revert button by Global Default EE.
Click Revert All to Default at the bottom of the page. 
Note all changes reverted (which a user may not want).

### Expected results

A user would be able to individually Revert a change made to the Global Default EE field.

### Actual results

User is only able to revert all changes on that settings page- if they want the change to the Global Default EE to be reverted.

akus062381:
resolved

PR #11107, opened 2021-09-22
Wrap ExecutionEnv Lookup in SettingGroup component.

kialam:
##### SUMMARY

Related #10785 

This PR adds a Revert Button to the Execution Environment lookup in the Misc. Settings page.

ee_revert

##### ISSUE TYPE

 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME

 - UI

##### AWX VERSION

```
19.3.0
```

### DIFF
--- a/awx/ui/src/components/Lookup/Lookup.js
+++ b/awx/ui/src/components/Lookup/Lookup.js
@@ -54,14 +54,15 @@ function Lookup(props) {
   } = props;
   const [typedText, setTypedText] = useState('');
   const debounceRequest = useDebounce(onDebounce, 1000);
-
   useField({
     name: fieldName,
     validate: (val) => {
-      if (!multiple && !val && typedText && typedText !== '') {
-        return t`That value was not found. Please enter or select a valid value.`;
-      }
-      return validate(val);
+      setTimeout(() => {
+        if (!multiple && !val && typedText && typedText !== '') {
+          return t`That value was not found. Please enter or select a valid value.`;
+        }
+        return validate(val);
+      });
     },
   });
 
@@ -79,6 +80,8 @@ function Lookup(props) {
     dispatch({ type: 'SET_VALUE', value });
     if (value?.name) {
       setTypedText(value.name);
+    } else {
+      setTypedText('');
     }
   }, [value, multiple]);
 
--- a/awx/ui/src/components/Lookup/Lookup.js
+++ b/awx/ui/src/components/Lookup/Lookup.js
@@ -57,12 +57,10 @@ function Lookup(props) {
   useField({
     name: fieldName,
     validate: (val) => {
-      setTimeout(() => {
-        if (!multiple && !val && typedText && typedText !== '') {
-          return t`That value was not found. Please enter or select a valid value.`;
-        }
-        return validate(val);
-      });
+      if (!multiple && !val && typedText && typedText !== '') {
+        return t`That value was not found. Please enter or select a valid value.`;
+      }
+      return validate(val);
     },
   });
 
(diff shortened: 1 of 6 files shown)


awx/ui/src/components/Lookup/Lookup.js
```diff
@@ -54,14 +54,15 @@ function Lookup(props) {
...
   const debounceRequest = useDebounce(onDebounce, 1000);
-
   useField({
     name: fieldName,
     validate: (val) => {
-      if (!multiple && !val && typedText && typedText !== '') {
-        return t`That value was not found. Please enter or select a valid value.`;
-      }
-      return validate(val);
+      setTimeout(() => {
```
kialam: Use [`setTimeout`](https://github.com/formium/formik/issues/2083#issuecomment-694265824) to avoid race condition between formik validator and React hooks throwing validation errors.

AlexSCorey: I don't think this is necessary.  I think the issue is that we are calling `setTouched` inside the `onChange` call [here](https://github.com/ansible/awx/pull/11107/files#diff-edaa37419860b1661e289b954b771a9f5e10dfc50976c26d459f193ecc2221e4R253) when in reality we only want to validate that look up(never really, because it isn't required anywhere) when we blur from the field, rather on onChange.

awx/ui/src/screens/Setting/shared/SharedFields.js
```diff
@@ -229,6 +230,39 @@ EncryptedField.propTypes = {
...
+      popoverContent={config.help_text}
+      validated={isValid ? 'default' : 'error'}
+      isDisabled={field.value === null}
+    >
+      <ExecutionEnvironmentLookup
+        onBlur={() => helpers.setTouched(true)}
+        value={field.value}
+        onChange={(value) => {
+          helpers.setValue(value);
+          helpers.setTouched(true);
```
AlexSCorey: I think we should remove the setTouched here in favor of the one [above](https://github.com/ansible/awx/pull/11107/files#diff-edaa37419860b1661e289b954b771a9f5e10dfc50976c26d459f193ecc2221e4R249)

AlexSCorey (changes requested):
Re: Validating the ExecutionEnvironmentLookup
I don't think we need to validate that look up because it isn't required anywhere in the project from what I can tell.  However, if you do want to have validation on it I have made some suggestions below that seem to work and allow us to remove the `setTimeout` inside the `<Lookup>` component.

akus062381 (approved):
QE approval!

awx/ui/src/screens/Setting/shared/SharedFields.js
```diff
@@ -229,6 +230,35 @@ EncryptedField.propTypes = {
...
+      isDisabled={field.value === null}
+      onRevertCallback={() => helpers.setValue(config.default)}
+    >
+      <ExecutionEnvironmentLookup
+        value={field.value}
+        onChange={(value) => {
+          helpers.setValue(value, false);
+        }}
+        overrideLabel
+        fieldName={name}
```
nixocio, line 252: hello @kialam, good job. 

We need to keep those 2 props.

```js
                  popoverContent={t`The Execution Environment to be used when one has not been configured for a job template.`}
                  isGlobalDefaultEnvironment
```

nixocio (changes requested):

awx/ui/src/screens/Setting/shared/SharedFields.js
nixocio: I see the changes that you added about `overrideLabel` 👍 and relying on the data returned by the API.

nixocio (approved):

AlexSCorey (approved):
Looks good to me.

PR #11129, opened 2021-09-24
Delete wording that we have reversed position on

AlanCoding:
In current development work, this statement is no longer true. Without refactoring the whole docs, I want to remove this one statement to avoid ambiguity.


chrismeyersfsu (approved):

tvo318 (commented):
ok that's consistent with your doc PR 👍

Issue #6302, opened 2020-03-16
Wrapping on lists for long names does not seem to be in place


appuk:
##### ISSUE TYPE
 - Bug Report

##### STEPS TO REPRODUCE
1. Create an inventory
2. Go to the hosts tab
3. Add a host with a really long name
4. Go to the Host List View

##### ADDITIONAL INFORMATION
Verified on Hosts and Schedules( generic issue for schedules: https://github.com/ansible/awx/issues/4959) , but need to check for all the pages

##### EXPECTED RESULTS
The name is wrapped properly

##### ACTUAL RESULTS
The name is not wrapped and the buttons go out of the page

<img width="1669" alt="Screen Shot 2020-03-16 at 3 58 02 PM" src="https://user-images.githubusercontent.com/11617351/76796171-426c2280-67a1-11ea-9312-89149cc311fb.png">

unlikelyzero:
@mabashian is there a good way to unit test this?

mabashian:
I'm curious to see how this shakes out when we move to tables.  This problem may go away.

mabashian:
I don't think anyone should work on this until tables land.

mabashian:
Just checked this out in tables:

<img width="1516" alt="Screen Shot 2021-02-04 at 11 44 28 AM" src="https://user-images.githubusercontent.com/9889020/106925964-985df780-66de-11eb-81db-ec0d47ebf1ba.png">

There doesn't seem to be any word breaking so you can make the table really wide requiring horizontal scrolling.

I think this still needs to remain open

akus062381:
Additional issue identified with regard to wrapping in the Schedules Tab: When the browser window is made small and the schedule names are long, Next Run wraps the date/time stamp in an odd way.

PR #11132, opened 2021-09-24
Break lines for long strings on main lists

nixocio:
Break lines for long strings on main lists to avoid horizontal
scrolling.

Main goal of this change is to avoid actions items to be hidden on the
main lists.

With this change
image

Without this change

image

See: https://github.com/ansible/awx/issues/6302


AlexSCorey (approved):
Looks good.  Well done!

akus062381 (approved):
great addition

PR #11133, opened 2021-09-24
AWX dev environment changes for receptor work signing feature

amolgautam25:
##### SUMMARY

Made dev environment changes to AWX docker install.
Added RSA keys for signing receptor work
Changed jinja config for the ```awx_tools``` and ```tools_receptor```

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.3.0
```

##### ADDITIONAL INFORMATION

### DIFF
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -478,7 +478,10 @@ def inspect_execution_nodes(instance_list):
         for ad in connections:
             hostname = ad['NodeID']
             commands = ad.get('WorkCommands') or []
-            if 'ansible-runner' not in commands:
+            worktypes = []
+            for c in commands:
+                worktypes.append(c["WorkType"])
+            if 'ansible-runner' not in worktypes:
                 continue
             changed = False
             if hostname in node_lookup:
@@ -3053,7 +3056,7 @@ class AWXReceptorJob:
             use_stream_tls = get_conn_type(_kw['node'], receptor_ctl).name == "STREAMTLS"
             _kw['tlsclient'] = get_tls_client(use_stream_tls)
 
-        result = receptor_ctl.submit_work(worktype=self.work_type, payload=sockout.makefile('rb'), params=self.receptor_params, **_kw)
+        result = receptor_ctl.submit_work(worktype=self.work_type, payload=sockout.makefile('rb'), params=self.receptor_params, signwork=True, **_kw)
         self.unit_id = result['unitid']
         self.task.update_model(self.task.instance.pk, work_unit_id=result['unitid'])
 
--- a/awx/main/utils/receptor.py
+++ b/awx/main/utils/receptor.py
@@ -70,6 +70,7 @@ def worker_info(node_name, work_type='ansible-runner'):
 
     kwargs = {}
     kwargs['tlsclient'] = get_tls_client(use_stream_tls)
+    kwargs['signwork'] = True
     if work_type != 'local':
         kwargs['ttl'] = '20s'
     result = receptor_ctl.submit_work(worktype=work_type, payload='', params={"params": f"--worker-info"}, node=node_name, **kwargs)
--- a/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
+++ b/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
@@ -166,7 +166,7 @@ COPY --from=builder /var/lib/awx /var/lib/awx
 RUN ln -s /var/lib/awx/venv/awx/bin/awx-manage /usr/bin/awx-manage
 
 {%if build_dev|bool %}
-COPY --from=quay.io/project-receptor/receptor:1.0.0a2 /usr/bin/receptor /usr/bin/receptor
+COPY --from=quay.io/ansible/receptor:devel /usr/bin/receptor /usr/bin/receptor
 RUN openssl req -nodes -newkey rsa:2048 -keyout /etc/nginx/nginx.key -out /etc/nginx/nginx.csr \
         -subj "/C=US/ST=North Carolina/L=Durham/O=Ansible/OU=AWX Development/CN=awx.localhost" && \
     openssl x509 -req -days 365 -in /etc/nginx/nginx.csr -signkey /etc/nginx/nginx.key -out /etc/nginx/nginx.crt && \
--- a/tools/docker-compose/ansible/roles/sources/tasks/main.yml
+++ b/tools/docker-compose/ansible/roles/sources/tasks/main.yml
@@ -79,6 +79,16 @@
     awx_image_tag: "{{ lookup('file', playbook_dir + '/../../../VERSION') }}"
   when: awx_image_tag is not defined
 
+- name: Generate Private RSA key for signing work
+  command: openssl genrsa -out {{ work_sign_private_keyfile }} {{ receptor_rsa_bits }}
+  args:
+    creates: "{{ work_sign_private_keyfile }}"
+
+- name: Generate public RSA key for signing work
+  command: openssl rsa -in {{ work_sign_private_keyfile }} -out {{ work_sign_public_keyfile }} -outform PEM -pubout
+  args:
+    creates: "{{ work_sign_public_keyfile }}"
+
 - name: Render Docker-Compose
   template:
     src: docker-compose.yml.j2
--- a/tools/docker-compose/ansible/roles/sources/templates/receptor-awx.conf.j2
+++ b/tools/docker-compose/ansible/roles/sources/templates/receptor-awx.conf.j2
@@ -1,11 +1,22 @@
 ---
 - node:
     id: awx_{{ item }}
+    firewallrules:
+      - action: "reject"
+        tonode: awx_{{ item }}
+        toservice: "control"
 
 - log-level: info
 
 - tcp-listener:
     port: 2222
+
+- work-signing:
+    privatekey: /etc/receptor/work_private_key.pem
+    tokenexpiration: 1m
+
+- work-verification:
+    publickey: /etc/receptor/work_public_key.pem
 
 {% for i in range(item | int + 1, control_plane_node_count | int + 1) %}
 - tcp-peer:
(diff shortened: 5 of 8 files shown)


awx/main/utils/receptor.py
```diff
@@ -37,6 +37,7 @@ def worker_info(node_name, work_type='ansible-runner'):
     kwargs = {}
     if work_type != 'local':
         kwargs['ttl'] = '20s'
+        kwargs['signwork'] = True
```
fosterseth: you'll need signwork=True in local case as well. So you can just put signwork=True in the submit_work down below

tools/docker-compose/ansible/roles/sources/tasks/main.yml
```diff
@@ -8,6 +8,7 @@
   loop:
     - secrets
     - receptor
+    - work_sign_keys
```
fosterseth: we may just want to put this stuff under the receptor, or in a subfolder within receptor. that way people know these keys are a receptor thing

awx/main/utils/receptor.py
amolgautam25: I have made the required changes

tools/docker-compose/ansible/roles/sources/templates/receptor-worker.conf.j2
```diff
@@ -8,12 +8,17 @@
...
     worktype: ansible-runner
     command: ansible-runner
     params: worker
     allowruntimeparams: true
+    verifysignature: true
 
 - control-service:
     service: control
     filename: {{ receptor_socket_file }}
+    verifysignature: true
```
fosterseth: not on control-service :)

tools/docker-compose/ansible/roles/sources/tasks/main.yml
amolgautam25: I have made the required change.

awx/main/tasks.py
```diff
@@ -475,7 +475,10 @@ def inspect_execution_nodes(instance_list):
         for ad in connections:
             hostname = ad['NodeID']
             commands = ad.get('WorkCommands') or []
-            if 'ansible-runner' not in commands:
+            worktypes = []
+            for c in commands:
+                worktypes.append(c["WorkType"])
+            if 'ansible-runner' not in worktypes:
                 continue
```
AlanCoding, line 485: JSON format of the current dev environment

```json
{
  "Advertisements": [
    {
      "NodeID": "awx_1",
      "Service": "control",
      "Time": "2021-09-27T17:57:47.642621671Z",
      "ConnType": 1,
      "Tags": null,
      "WorkCommands": [
        "local",
        "kubernetes-runtime-auth",
        "kubernetes-incluster-auth"
      ]
    },
    {
      "NodeID": "receptor-2",
      "Service": "control",
      "Time": "2021-09-27T17:56:58.552544691Z",
      "ConnType": 1,
      "Tags": null,
      "WorkCommands": [
        "ansible-runner"
      ]
    },
    {
      "NodeID": "receptor-1",
      "Service": "control",
      "Time": "2021-09-27T17:56:58.696644071Z",
      "ConnType": 1,
      "Tags": null,
      "WorkCommands": [
        "ansible-runner"
      ]
    }
  ],
  "Connections": [
    {
      "NodeID": "receptor-hop",
      "Cost": 1
    }
  ],
  "KnownConnectionCosts": {
    "awx_1": {
      "receptor-hop": 1
    },
    "receptor-1": {
      "receptor-hop": 1
    },
    "receptor-2": {
      "receptor-hop": 1
    },
    "receptor-hop": {
      "awx_1": 1,
      "receptor-1": 1,
      "receptor-2": 1
    }
  },
  "NodeID": "awx_1",
  "RoutingTable": {
    "receptor-1": "receptor-hop",
    "receptor-2": "receptor-hop",
    "receptor-hop": "receptor-hop"
  },
  "SystemCPUCount": 8,
  "SystemMemoryMiB": 31855,
  "Version": "1.0.0.0a2"
}
```

It might be nice if you could share `receptorctl status --json` of the environment you have running with these changes so we can see what we're dealing with.

Otherwise, it's kind of a shot in the dark as to whether this logic is robust or not.

I'm tempted to say that it should work with either type of configuration (i.e. it should correctly parse the output I pasted). In that case, it looks like we would need to branch based on the type of elements in `commands` being dict / string.

tools/ansible/roles/dockerfile/templates/Dockerfile.j2
```diff
@@ -166,7 +166,7 @@ COPY --from=builder /var/lib/awx /var/lib/awx
 RUN ln -s /var/lib/awx/venv/awx/bin/awx-manage /usr/bin/awx-manage
 
 {%if build_dev|bool %}
-COPY --from=quay.io/project-receptor/receptor:1.0.0a2 /usr/bin/receptor /usr/bin/receptor
+COPY --from=quay.io/ansible/receptor:devel /usr/bin/receptor /usr/bin/receptor
```
AlanCoding, line 169: well now, that's important to update :+1:

shanemcd (approved):

tools/docker-compose/ansible/roles/sources/templates/receptor-awx.conf.j2
```diff
@@ -7,6 +7,13 @@
 - tcp-listener:
     port: 2222
 
+- work-signing:
+    privatekey: /etc/receptor/work_private_key.pem
+    tokenexpiration: 1m
```
AlanCoding, line 16: Could you help us out by just talking out what the selection of 1m means and why you chose it?

AlanCoding (changes requested):
This looks really good, but I think I want to double-down on my request that `inspect_execution_nodes` should parse what's under "WorkCommands" whether it has this stuff turned on or off. This should be a plain python modification so let me know if there's any confusion about my request.

and the rest of it looks really good!

tools/docker-compose/ansible/roles/sources/tasks/main.yml
fosterseth: i think you still need to remove this - work_sign_keys, the folder is still be created

awx/main/tasks.py
fosterseth: below is the new format. what is the reason for handling the old format?

``` json
{
  "Advertisements": [
    {
      "NodeID": "awx_1",
      "Service": "control",
      "Time": "2021-09-28T19:47:01.210495323Z",
      "ConnType": 1,
      "Tags": {
        "type": "Control Service"
      },
      "WorkCommands": [
        {
          "WorkType": "local",
          "Secure": true
        },
        {
          "WorkType": "kubernetes-runtime-auth",
          "Secure": true
        },
        {
          "WorkType": "kubernetes-incluster-auth",
          "Secure": true
        }
      ]
    },
    {
      "NodeID": "receptor-2",
      "Service": "control",
      "Time": "2021-09-28T19:46:33.802188673Z",
      "ConnType": 1,
      "Tags": {
        "type": "Control Service"
      },
      "WorkCommands": [
        {
          "WorkType": "ansible-runner",
          "Secure": true
        }
      ]
    },
    {
      "NodeID": "receptor-1",
      "Service": "control",
      "Time": "2021-09-28T19:46:33.806122556Z",
      "ConnType": 1,
      "Tags": {
        "type": "Control Service"
      },
      "WorkCommands": [
        {
          "WorkType": "ansible-runner",
          "Secure": true
        }
      ]
    }
  ],
  "Connections": [
    {
      "NodeID": "receptor-hop",
      "Cost": 1
    }
  ],
  "KnownConnectionCosts": {
    "awx_1": {
      "receptor-hop": 1
    },
    "receptor-1": {
      "receptor-hop": 1
    },
    "receptor-2": {
      "receptor-hop": 1
    },
    "receptor-hop": {
      "awx_1": 1,
      "receptor-1": 1,
      "receptor-2": 1
    }
  },
  "NodeID": "awx_1",
  "RoutingTable": {
    "receptor-1": "receptor-hop",
    "receptor-2": "receptor-hop",
    "receptor-hop": "receptor-hop"
  },
  "SystemCPUCount": 8,
  "SystemMemoryMiB": 15865,
  "Version": "-0.git20210928."
}
```

fosterseth:
maybe we should add our same firewall rules too, for complete parity between dev and installer setups

fosterseth:
Informally tested this -- can launch jobs successfully and not seeing any errors in the log output. Unfortunately we can't run yolo at the moment due to receptor RPM not being up to date. 

Receptor RPM should be built roughly the same time this is merged, since they co-depend on each other

tools/docker-compose/ansible/roles/sources/templates/receptor-awx.conf.j2
fosterseth: when awx_1 makes a connection to receptor_1, it issues a payload with the "work submit" command. This payload includes a json web token that expires after 1 minute. So, that payload has 1 whole minute to get to receptor_1 where it will be verified.

It only generates the token at the very last second, after it has already made a valid connection to receptor_1. Therefore, even if receptor_1 is down temporarily, it should only start counting when it comes back online.

keeping the token expiration short should help mitigate replay attacks.

awx/main/tasks.py
AlanCoding: okay, so then we just need to be extra diligent about updating images for this merge.

AlanCoding (approved):
my issue was addressed by @fosterseth

beeankha (approved):

tools/docker-compose/ansible/roles/sources/tasks/main.yml
amolgautam25: I have made the changes.

tools/docker-compose/ansible/roles/sources/templates/receptor-awx.conf.j2
AlanCoding: Okay, so this payload is just telling the execution node "run `ansible-runner worker`", right? _After_ that, the controller node and the execution node will be chatting it up over the socket. This will not require that the streaming of the job directory from the controller node to the execution node has finished in 1 minute, right? Because it will almost certainly take longer in cases.

Issue #11127, opened 2021-09-24
Jobs List alert- EE is a missing resource- error


akus062381:
##### ISSUE TYPE
 - Bug Report

##### COMPONENT NAME
 - UI

##### SUMMARY
If an instance is disabled, and a job is run against that instance, it gets stuck in a pending state. If the user cancels the job, the UI shows "Missing Resource" next to EE. In this case, the job didn't run because the instance was disabled, but this error/alert may make the user think that they are missing a resource and go down a "rabbit hole" trying to fix something that isn't the problem.  Please see GIF for more detail:
![jobs-list-error](https://user-images.githubusercontent.com/63685497/134601204-df01caa3-60bf-4a22-8d30-de4ce89b8c05.gif)

**Important to **Note--**** this same error occurrs if a job is canceled in the jobs list.

##### STEPS TO REPRODUCE
1. Visit the controlplace instance and click the instances tab.
2. Toggle the switch to disable the instance.
3. Access the Demo Job Template and click Launch.
4. When the output screen is shown, click the link to take you to the jobs list.
5. Cancel the pending jobs. You will see the missing resource alert next to EE.
6. Go back to controlplane and enable the instance.
7. Go back to Demo Job Template and click launch.
8. Cancel the job before it has time to finish.
9. Take note of the missing resource alert next to EE.

##### EXPECTED RESULTS

The field next to EE would either be blank or simply indicate "Job Canceled"

##### ACTUAL RESULTS

The UI shows "Missing Resource" next to EE

akus062381:
resolved

PR #11134, opened 2021-09-24
Do not display EE if a job was canceled

nixocio:
Do not display EE if a job was canceled. Since the API is returning null
for the value of the EE for this particular scenario.

See: https://github.com/ansible/awx/issues/11127


AlexSCorey (approved):
This looks good to me

akus062381 (approved):
yay, awesome

Issue #10822, opened 2021-08-05
Clean up react warnings


nixocio:
##### SUMMARY
Clean up react warnings.  Go to console.

See:
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    in InventorySourceAdd (at InventorySources.js:12)
    in Route (at InventorySources.js:11) index.js:1```
    
```
```
Warning: React does not recognize the `isValid` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `isvalid` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    in ul (created by Context.Consumer)
    in SelectMenu (created by ForwardRef)
    in ForwardRef (created by Select)
    in div (created by Select)
    in GenerateId (created by Select)
    in Select (at SCMSubForm.js:113)
    in div (created by FormGroup)
    in div (created by FormGroup)
    in FormGroup (at SCMSubForm.js:94)
    in SCMSubForm (at InventorySourceForm.js:204)
    in div (created by styled.div)
    in styled.div (at InventorySourceForm.js:155)
    in div (created by styled.div)
    in styled.div (at InventorySourceForm.js:151)
    in InventorySourceFormFields (at InventorySourceForm.js:298)
    in div (created by styled.div)
    in styled.div (at InventorySourceForm.js:297)
```

One more - Logged out and see 

``` 
Warning: Cannot update a component (`SessionProvider`) while rendering a different component (`ProtectedRoute`). To locate the bad setState() call inside `ProtectedRoute`, follow the stack trace as described in https://fb.me/setstate-in-render
    in ProtectedRoute (at App.js:166)
    in Switch (at App.js:156)
    in SessionProvider (at App.js:155)
    in Unknown (at App.js:154)
    in I18nProvider (at App.js:153)
    in App (at App.js:180)
    in Router (created by HashRouter)
    in HashRouter (at App.js:179)
    in Unknown (at src/index.js:10)
    in StrictMode (at src/index.js:9)
```

AlexSCorey:
Some of the warnings have been cleaned up but there are still a few related to logging out.  I'm going to remove myself from this so someone else can pick it up.  Whoever picks this up feel free to reach out to me

PR #11147, opened 2021-09-27
Wrap `setAuthRedirectTo` in useEffect.

kialam:
##### SUMMARY

Related #10822

[Introduced in React v16.3.0](https://reactjs.org/blog/2020/02/26/react-v16.13.0.html#warnings-for-some-updates-during-render), this warning fires if `setState` is called from within a different component than the one being rendered. As documented, the fix is to wrap the `setState` call in a `useEffect` hook.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - UI

##### AWX VERSION

```
19.3.0
```

### DIFF
--- a/awx/ui/src/App.js
+++ b/awx/ui/src/App.js
@@ -102,6 +102,10 @@ const ProtectedRoute = ({ children, ...rest }) => {
   const { authRedirectTo, setAuthRedirectTo } = useSession();
   const { pathname } = useLocation();
 
+  useEffect(() => {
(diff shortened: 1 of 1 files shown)


AlexSCorey (approved):
Holy cow.  I'm so happy you fixed this...and with such a simple change.  Well done!

nixocio:
The test failures on e2e were not related to this change.

Issue #11113, opened 2021-09-22
slack URL uses /system/ instead of /management/ 404 (not found)


nbari:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

When enabling slack notifications and running the management jobs, the URL parameter used is `/system/` instead of `/management/`

Slacks `#/jobs/system/77/output`

instead of `#/jobs/management/77/output`

### AWX version

19.3.0

### Installation method

kubernetes

### Modifications

no

### Operating system

linux

### Web browser

Chrome

### Steps to reproduce

Enable notifications for any management jobs

### Expected results

https:/<domai>/#/jobs/**management**/77/output

### Actual results

https:/<domai>/#/jobs/**system**/77/output

PR #11149, opened 2021-09-27
Redirect system path to management on jobs URL

nixocio:
When user attempts to access `/jobs/system/66` redirect to
`/jobs/management/66`.

This will catch management jobs notifications, for instance, and redirect to
the proper URL.

See:https://github.com/ansible/awx/issues/11113


AlexSCorey (approved):
This looks good to me.

PR #11154, opened 2021-09-28
Fix rpm 'offline' build (add pbr to venv and change receptorctl wheel file name)

simaishi:
Requires: https://github.com/ansible/receptor/pull/427

There are currently 2 issues with rpm 'offline' build:

- The build fails tying to get version info from ansible-runner due to missing `pbr`. Since ansible-runner requires pbr to get version info when it comes from git branch, adding pbr to venv for now.  When we move ansible-runner from requirements_git.txt to requirements.in, pbr should be removed as well.

- receptorctl nightly wheel file name `receptorctl-latest-py3-none-any.whl` was considered invalid and needed to change the version identifier to numeric.

### DIFF
--- a/awx/main/tests/functional/test_licenses.py
+++ b/awx/main/tests/functional/test_licenses.py
@@ -48,6 +48,7 @@ def test_python_and_js_licenses():
 
     def read_api_requirements(path):
         ret = {}
+        skip_pbr_license_check = False
         for req_file in ['requirements.txt', 'requirements_git.txt']:
             fname = '%s/%s' % (path, req_file)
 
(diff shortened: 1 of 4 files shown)


kdelee (approved):

kdelee:
Looks like the test that is failing is because of this pbr addition:
```
_________________________ test_python_and_js_licenses __________________________
[gw0] linux -- Python 3.8.6 /var/lib/awx/venv/awx/bin/python3.8
  File "/awx_devel/awx/main/tests/functional/test_licenses.py", line 125, in test_python_and_js_licenses
    raise Exception('Included licenses not consistent with requirements:\n%s' % '\n'.join(errors))
Exception: Included licenses not consistent with requirements:
 license for requirement pbr is missing
 ```

we either need to skip pbr somehow or add the license it looks like :expressionless:

kdelee:
here is the license https://opendev.org/openstack/pbr/src/branch/master/LICENSE

simaishi:
I changed the license test code to skip pbr if ansible-runner isn't a released version.

shanemcd (approved):

PR #11157, opened 2021-09-28
Redirect system path to management on jobs URL

nixocio:
When user attempts to access `/jobs/system/66` redirect to
`/jobs/management/66`.

This will catch management jobs notifications, for instance, and redirect to
the proper URL.

See:#11113

### DIFF
--- a/awx/ui/src/screens/Job/Job.js
+++ b/awx/ui/src/screens/Job/Job.js
@@ -28,6 +28,7 @@ export const JOB_URL_SEGMENT_MAP = {
   playbook: 'job',
   project: 'project_update',
   management: 'system_job',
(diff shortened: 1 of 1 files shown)


AlexSCorey (approved):
Looks good to me

akus062381:
@nixocio I just tested this, and what I am seeing is that the redirect only works if the user is not already on the output screen. So for example- if a user is on /#/jobs/management/86/output, and they manually delete the word 'management' in the address bar, and replace it with system, that the url redirect does not work.
I am not sure if you see this as an issue, but I wanted to bring it to your attention.

akus062381 (approved):

Issue #10948, opened 2021-08-26
Background job to cleanup private_data_dir files left by interrupted job


jladdjr:
When a job runs successfully on awx, the private_data_dir should be cleaned up, both on the controller instance where it was created and on the execution instance. This work should already be addressed by https://github.com/ansible/awx/issues/10687.

If a job fails abnormally (e.g. due to a network failure, tower services restarting unexpectedly, etc), we need to ensure that all copies of the private_data_dir are cleaned up.

## Acceptance Criteria

Create a new periodic task that
- [ ] prunes private_data_dirs  for inactive jobs on remote execution nodes
- [ ] prunes private_data_dirs  for inactive jobs on controller instances

shanemcd:
A _little_ more context here would be good.

jladdjr:
@shanemcd 👍  see updated description

jladdjr:
We could see this also incorporating the task of cleaning up unused images - see https://github.com/ansible/awx/issues/10701

AlanCoding:
I've created this doc to express what I'm trying to do

https://gist.github.com/AlanCoding/99de3b85031cae3b7f2c7c8f81560359#file-cleanup_demo-md

AlanCoding:
> prunes private_data_dirs for inactive jobs on controller instances

We already have a mechanism for this and have no intention of changing anything with it.

AlanCoding:
Recap of active PRs:

 - awx change: https://github.com/ansible/awx/pull/11160
 - runner change: https://github.com/ansible/ansible-runner/pull/846

jladdjr:
@thenets - just checking in, do you need any info for testing this?

Issue #10702, opened 2021-07-21
cleanup unused execution environment images job not working as intended


fosterseth:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

original PR here https://github.com/ansible/awx/pull/9907

This cleanup task only removes dangling (untagged) podman images. If a user removes an EE, those images left on the system will still be tagged, but ideally they would be removed too.

We need to revisit this.

### AWX version

awx: 19.2.2

### Installation method

docker on linux

### Modifications

no

### Steps to reproduce

- change the celerybeat_schedule in defaults.py for `cleanup_images` to run once per 30 seconds or so
- create EE images
- Associate them with EEs in awx
- Remove the EE in awx
- Wait 30 seconds to make sure cleanup job runs (will produce debug statement log as well)
- podman images -a will still show the disassociated images on the system

### Expected results

Images should be removed

### Actual results

not removed

Zokormazo:
Seems fixed now

Issue #10701, opened 2021-07-21
Cleanup unused images in execution nodes


AlanCoding:
We have this task:

https://github.com/ansible/awx/blob/68e309ee3281eb725f38b88ffa84959e1c007937/awx/main/tasks.py#L384-L399

That has a couple of problems when we introduce the control plane / execution plane split.

 - This task should not do anything if it is ran on a control-only node. Depending on installer changes, those nodes may not have podman installed.
 - This won't run at all on execution-only nodes.

AlanCoding:
Marking as blocked, because we may hold off pending solution for https://github.com/ansible/awx/issues/10702

jladdjr:
We could see this tying in to the work to clean up the private data dir via a background task - see https://github.com/ansible/awx/issues/10948

jladdjr:
Wanted to note that this may have already be an issue with awx and hybrid nodes (before execution nodes were introduced).

jladdjr:
We need to check on the existing behavior for this and make sure that at the very least we haven't regressed on this. On the latest release do we have containers / images that get left behind? What do we see happening today?

AlanCoding:
You can't say we regressed, because we never removed images from deleted EEs to begin with.

I'm leaving the door open to addressing this in conjunction with https://github.com/ansible/ansible-runner/pull/846, but I'm not committing to it.

PR #11160, opened 2021-09-28
Consolidate cleanup actions under new `ansible-runner worker cleanup` command

AlanCoding:
##### SUMMARY
This is the corresponding AWX change for the ansible-runner PR https://github.com/ansible/ansible-runner/pull/846

These issues I link here should be fully closed when this is merged and tested:

Connect https://github.com/ansible/awx/issues/10948
Connect https://github.com/ansible/awx/issues/10701
Connect https://github.com/ansible/awx/issues/10702

This is a little bit of an aggressive refactor that grew to envelop several issues, because this involves designing a new interface, and I don't want to introduce that sort of thing unless it is thought-out and works for all the cases we planned.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
19.3.3
```

##### ADDITIONAL INFORMATION
This might help people to grok what's going on here:

```
$ ansible-runner worker cleanup --help
usage: ansible-runner worker cleanup [-h] [--file-pattern FILE_PATTERN] [--exclude-strings [EXCLUDE_STRINGS ...]]
                                     [--remove-images [REMOVE_IMAGES ...]] [--grace-period GRACE_PERIOD] [--image-prune]
                                     [--process-isolation-executable PROCESS_ISOLATION_EXECUTABLE]

optional arguments:
  -h, --help            show this help message and exit
  --file-pattern FILE_PATTERN
                        A file glob pattern to find private_data_dir folders to remove. Example: --file-
                        pattern=/tmp/.ansible-runner-*
  --exclude-strings [EXCLUDE_STRINGS ...]
                        A comma separated list of keywords in directory name or path to avoid deleting.
  --remove-images [REMOVE_IMAGES ...]
                        A comma separated list of podman or docker tags to delete. This may not remove the corresponding
                        layers, use the image-prune option to assure full deletion. Example: --remove-
                        images=quay.io/user/image:devel quay.io/user/builder:latest
  --grace-period GRACE_PERIOD
                        Time (in minutes) after last modification to exclude a folder from deletion for. This is to avoid
                        deleting folders that were recently created, or folders not started via the start command. Value of
                        0 indicates that no folders will be excluded based on modified time.
  --image-prune         If specified, will run docker / podman image prune --force. This will only run after untagging.
  --process-isolation-executable PROCESS_ISOLATION_EXECUTABLE
                        The container image to clean up images for (default=podman)
```

Everything here is glue to use this. For the control / hybrid nodes, we import it and just directly call it. For execution nodes, we build the CLI args and run those over receptor in the same way as we do for the `--worker-info` health checks.

There's still a lot of tests and docs to write, but this has developed enough that I don't expect any additional major code revisions.

### DIFF
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -27,7 +27,6 @@ import socket
 import threading
 import concurrent.futures
 from base64 import b64encode
-import subprocess
 import sys
 
 # Django
@@ -51,13 +50,14 @@ from gitdb.exc import BadName as BadGitName
 
 # Runner
 import ansible_runner
+import ansible_runner.cleanup
 
 # dateutil
 from dateutil.parser import parse as parse_date
 
 # AWX
 from awx import __version__ as awx_application_version
-from awx.main.constants import PRIVILEGE_ESCALATION_METHODS, STANDARD_INVENTORY_UPDATE_ENV, MINIMAL_EVENTS
+from awx.main.constants import PRIVILEGE_ESCALATION_METHODS, STANDARD_INVENTORY_UPDATE_ENV, MINIMAL_EVENTS, JOB_FOLDER_PREFIX
 from awx.main.access import access_registry
 from awx.main.redact import UriCleaner
 from awx.main.models import (
@@ -106,7 +106,7 @@ from awx.main.utils.safe_yaml import safe_dump, sanitize_jinja
 from awx.main.utils.reload import stop_local_services
 from awx.main.utils.pglock import advisory_lock
 from awx.main.utils.handlers import SpecialInventoryHandler
-from awx.main.utils.receptor import get_receptor_ctl, worker_info, get_conn_type, get_tls_client
+from awx.main.utils.receptor import get_receptor_ctl, worker_info, get_conn_type, get_tls_client, worker_cleanup
 from awx.main.consumers import emit_channel_notification
 from awx.main import analytics
 from awx.conf import settings_registry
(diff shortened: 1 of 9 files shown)


awx/main/tasks.py
```diff
@@ -390,22 +392,42 @@ def purge_old_stdout_files():
...
+    with StringIO() as buffer:
+        with redirect_stdout(buffer):
+            ansible_runner.cleanup.run_cleanup(runner_cleanup_kwargs)
+            stdout = buffer.getvalue()
+    if '(changed: True)' in stdout:
+        logger.info(f'Performed local cleanup with kwargs {kwargs}, output:\n{stdout}')
+
+    # if we are the first instance alphabetically, then run cleanup on execution nodes
+    checker_instance = Instance.objects.filter(node_type__in=['hybrid', 'control'], enabled=True, capacity__gt=0).order_by('-hostname').first()
+    if checker_instance and this_inst.hostname == checker_instance.hostname:
```
AlanCoding, line 410: @chrismeyersfsu it feels like we have done everything under the sun in order to choose an arbitrary node out of the cluster to run some logic. We have used a Django solo model, we have used fields to track timestamps, select-for-update, and decide to run based on the time delta.

This is possibly the laziest option in the universe to solve that problem. And I feel no shame for it.

AlanCoding:
I don't think it will be possible to fix the unit tests here until the ansible-runner PR is merged.

awx/main/tasks.py
AlexSCorey: I think this comment should live in the code instead of behing hidden in github

beeankha (approved):

rebeccahhh:
pretend I approved this, because I was about to 😄 looks good y'all

PR #11166, opened 2021-09-29
get receptorctl status for sosreport

kdelee:
I presume the logs also get collected from journalctl but I'm not sure


fosterseth (approved):

beeankha (approved):

tools/sosreport/controller.py
```diff
@@ -24,6 +24,7 @@
     "ls -ll /etc/tower",
     "ls -ll /var/run/awx-receptor", # list contents of dirctory where receptor socket should be
     "ls -ll /etc/receptor",
+    "receptorctl --socket /var/run/awx-receptor/receptor.sock status", # Get invormation about the status of the mesh
```
kdelee:
```suggestion
    "receptorctl --socket /var/run/awx-receptor/receptor.sock status", # Get information about the status of the mesh
```

Issue #11169, opened 2021-09-29
Inventory Source AAP subform not being displayed


nixocio:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

Inventory Source AAP subform not being displayed

### AWX version

devel

### Installation method

docker for mac

### Modifications

no

### Steps to reproduce

1 - Attempt to create an inventory source using `Red Hat Ansible Automation Platform`.
2 - Subform not being displayed

<img width="1492" alt="image" src="https://user-images.githubusercontent.com/9053044/135331900-b537e935-bec1-4df4-86ea-419efdf514c5.png">

### Expected results

Subform being displayed

### Actual results

Subform not being displayed

PR #11171, opened 2021-09-29
Fix subform for AAP for inventory source

nixocio:
Fix subform for AAP for inventory source

image

See: https://github.com/ansible/awx/issues/11169

### DIFF
--- a/awx/ui/src/screens/Inventory/shared/InventorySourceForm.js
+++ b/awx/ui/src/screens/Inventory/shared/InventorySourceForm.js
@@ -22,7 +22,7 @@ import {
   OpenStackSubForm,
   SCMSubForm,
   SatelliteSubForm,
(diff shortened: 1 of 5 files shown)


awx/ui/src/screens/Inventory/shared/InventorySourceForm.js
```diff
@@ -207,7 +207,7 @@ const InventorySourceFormFields = ({
                     }
                   />
                 ),
-                tower: (
+                controller: (
                   <TowerSubForm
```
AlexSCorey: Should we change the name of the form?  I feel like at some point we should remove references to tower.  Maybe a backlog issue or something

AlexSCorey (approved):
This looks good to me.

akus062381 (approved):

PR #11173, opened 2021-09-29
Adds support for pre-filling EE add form name, description, and image from query params

mabashian:
##### SUMMARY
These changes allow links to be built that will take a navigating user straight to the EE add form with query params that will pre-fill some EE add form values:

Screen Shot 2021-09-29 at 5 23 49 PM

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui/src/App.js
+++ b/awx/ui/src/App.js
@@ -100,7 +100,7 @@ const AuthorizedRoutes = ({ routeConfig }) => {
 
 const ProtectedRoute = ({ children, ...rest }) => {
   const { authRedirectTo, setAuthRedirectTo } = useSession();
-  const { pathname } = useLocation();
+  const location = useLocation();
 
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


AlexSCorey (approved):
I pulled this down and tested it locally.  It all works as expected.  This is cool.  Well done!

tiagodread:
@marshmalien @mabashian this looks good, I added a test case to cover this feature in 6865.

I found just one thing when navigating back/forward the query string changes but the page isn't reloading with the correct prefilled information

https://user-images.githubusercontent.com/12464858/136041503-17fe3900-97eb-4395-8b9a-4eb872a60723.mp4

tiagodread (approved):
Verified that:

- EE can be created using name, image description via query string
- IF any query string is missing (or all), the form will just not populate the field
- Users can use an EE in hub using IP or subdomain (tested with ec2 URL)

PR #11175, opened 2021-09-30
don't collect keys with sosreport

kdelee:
##### SUMMARY

I realized we should exclude these files after talking with @tchellomello as they are not suitable for sharing.


tchellomello (approved):
Thanks @kdelee

PR #11180, opened 2021-09-30
update locator for UserDateDetail

akus062381:
Tiago and I worked on this so that we would have better locators for dates shown on the Detail screens.

### DIFF
--- a/awx/ui/src/components/DetailList/UserDateDetail.js
+++ b/awx/ui/src/components/DetailList/UserDateDetail.js
@@ -11,13 +11,13 @@ const Detail = styled(_Detail)`
   word-break: break-word;
 `;
 
-function UserDateDetail({ label, date, user, dataCy = null }) {
+function UserDateDetail({ label, date, user }) {
   const dateStr = formatDateString(date);
   const username = user ? user.username : '';
   return (
     <Detail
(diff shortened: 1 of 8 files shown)


awx/ui/src/components/DetailList/UserDateDetail.js
```diff
@@ -11,13 +11,13 @@ const Detail = styled(_Detail)`
...
 
-function UserDateDetail({ label, date, user, dataCy = null }) {
+function UserDateDetail({ label, date, user }) {
   const dateStr = formatDateString(date);
   const username = user ? user.username : '';
   return (
     <Detail
       label={label}
-      dataCy={dataCy}
+      dataCy="date-detail"
```
nixocio:
```suggestion
      dataCy="user-date-detail"
```

nixocio: This change will possibly generate duplicate `data-cy` on the same page. If that is ok.

tiagodread (approved):
Looks good to me, if you end up updating the locators to `user-date-detail`don't forget to update the tests too.

akus062381:
closing in favor of https://github.com/ansible/awx/pull/11180

PR #11204, opened 2021-10-05
Always set project sync execution_node to current host

AlanCoding:
@tchellomello made this exact change in commit 6d4b4cac3721e459ca77fdd2bb05038197ac91a5, but that was only for project syncs running before a job starts. We do the same thing for inventory updates, and this updates that other point in code with the same diff.


kdelee (approved):

tchellomello:
Got late for the party but thanks so much for fixing it @AlanCoding++

PR #11206, opened 2021-10-05
Upgrade a few ui dependencies

nixocio:
Upgrade axios, and ansi-to-html.


AlexSCorey (approved):
Noice!

tiagodread (approved):
I ran the tests and the same failures is happening on devel right now, so we can go ahead and just merge.

Issue #7741, opened 2020-07-28
Update advanced search key options to be grouped


jlmitch5:
https://github.com/patternfly/patternfly-react/pull/4622 -- blocked until this PR merges and we update to get that code

Currently, there are two sets of keys displayed in the options for this typeahead--the direct keys of the resource, and the related search fields. Since there is currently a bug with PF support of select groups with typeaheads, I have to concat and dedup these two list of options together.

Related to this work, the related search fields are special in that when no prefix or lookup is specified, we want to allow the user to access the fuzzy search for these related resources, so I append the __search lookup when that is the case. Because there are some cases where a related key is both a direct key (usually the id of which resource it is connected to), as well as a related search key with which you can search the fields of that resource, I do not do the __search appending if it is in both sets of options.

So the work involved here is:
- stop concat/deduping the list into allKeys https://github.com/ansible/awx/pull/7739/files#diff-c0fc4cf1aeb8368f537c5f44b71be4d7R35 and update the key options to be grouped.
- update the handleAdvancedSearch function to be aware of which group the key came from, so that we can do the __search lookup add if it was selected from the related group https://github.com/ansible/awx/pull/7739/files#diff-c0fc4cf1aeb8368f537c5f44b71be4d7R54

unlikelyzero:
@jlmitch5 @mabashian @wenottingham removing state:blocked as the upstream pf PR component has merged

tiagodread:
This was addressed by #11208

PR #11208, opened 2021-10-05
Groups Advanced Search Keys

AlexSCorey:
##### SUMMARY
This addresses #7741 and it also addresses a bug that I don't think has an issue, where the Clear all filters button was not being removed, when there were no search filters.  
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
searchClear
Screen Shot 2021-10-05 at 3 57 26 PM
Screen Shot 2021-10-05 at 3 57 21 PM

### DIFF
--- a/awx/ui/src/components/Search/AdvancedSearch.js
+++ b/awx/ui/src/components/Search/AdvancedSearch.js
@@ -5,8 +5,10 @@ import { t } from '@lingui/macro';
 import {
   Button,
   ButtonVariant,
+  Divider,
   InputGroup,
   Select,
+  SelectGroup,
   SelectOption,
   SelectVariant,
   TextInput,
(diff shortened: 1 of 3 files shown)


awx/ui/src/components/Search/AdvancedSearch.js
```diff
@@ -41,15 +43,9 @@ function AdvancedSearch({
...
-  // we might want to revisit that.
-  const allKeys = [
-    ...new Set([
-      ...(searchableKeys.map((k) => k.key) || []),
-      ...(relatedSearchableKeys || []),
-    ]),
-  ];
+  const relatedKeys = relatedSearchableKeys.filter(
+    (sKey) => !searchableKeys.map(({ key }) => key).includes(sKey)
+  );
```
nixocio, line 48: This a nice function to remove duplicate keys.

nixocio (changes requested):
Hey @AlexSCorey, good job.

Overall those changes look good, I just noticed from time to time it will display a division without populate the select groups.

Like:

image

Next update the keys were shown. It seems that we could hide the divider and groups if we do not have keys. To be explored.

nixocio (approved):

tiagodread (changes requested):
@AlexSCorey I can not type inside key select anymore, when I try to do so, I get this error:

image

2 broken e2e tests is failing

akus062381:
@Alex, I found this bug in a different place (as we talked about). In this case, the user not only needs to be able to type, but they also need to be able to create a tag entry.
wfjt-error

AlexSCorey:
@akus062381 and @tiagodread that bug seems to be related to a PF upgrade that happened a few days ago.  They made a change that seems to have broken things for us but those changes are not related to this PR. I think we need to file an issue for it, pleasse

nixocio:
@AlexSCorey, please, can you rebase this PR? Since we downgraded PF, and verify that reported error does not remain.

tiagodread (approved):
After reverting PF4 upgrade, the tests are looking better now and the previous failures aren't related to this change.

Issue #10272, opened 2021-05-25
Error when deleting multiple inventory source sync jobs at one time


akus062381:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
This must be done with a jobs list that has data already populated. 
When syncing a source inside of an inventory multiple times (I did 4 in my manual test), and going to the Jobs List page, after waiting for them to finish syncing, click the check mark next to all 4 jobs and click delete. Instead of returning the remaining job list history, it returns a page saying No Jobs Found

##### STEPS TO REPRODUCE
![inventory-source-issue](https://user-images.githubusercontent.com/63685497/119571303-330e3f00-bd7f-11eb-8009-2d4f6552fe64.gif)

##### EXPECTED RESULTS

After deleting the sync jobs, the UI would return the remaining list of jobs

##### ACTUAL RESULTS

the UI returns a message saying No Jobs Found. Upon refresh, the jobs then appear.

AlexSCorey:
I wasn't able to reproduce this consistently.  I did get this error on occasion when deleting inventory source sync jobs, but also a when deleting a mixture of different kinds of jobs.

nixocio:
~Possible related to: https://github.com/ansible/awx/issues/9013~

nixocio:
The issue here seems to be a performance issue. Once the Job list has several jobs running/pending/successful attempt to delete several jobs. The list does not re-render to the initial amount of jobs.

1  - Create a Workflow with several nodes, 10 nodes - job templates.
2  - Trigger this Workflow run
3 - Trigger other jobs, inventory source, for example.
4 - Attempt to delete a few successful jobs.
5 - List should render back displaying all remaining jobs, sometimes this does not happen.

![Job](https://user-images.githubusercontent.com/9053044/134189117-e74b8f60-d44e-4b2e-8aa9-43966c60fac4.gif)

InnocentK:
Resolved in #11209

PR #11209, opened 2021-10-05
Pass configurable qs to fetchJobsById function.

kialam:
##### SUMMARY

This one was a bit tricky to pin down, but the issue is that on occasion, if there were multiple jobs running and a user attempted to delete a few completed jobs, the Job List would sometimes show an empty list of jobs. Thanks @nixocio for doing most of the legwork here.

I've traced this down to the following key events that is causing this bug to fire:
1. User kicks off a multitude of jobs and visits the Job List page.
2. Job List page uses a `useWSJobs` hook that throttles (e.g. limits) the number of job updates to request to the API (one call every 5 seconds).
3. User deletes a completed job on the Job List page.
4. A throttle event is released - e.g. the Job List page is asked to query for example, a job with an `id` of `1101`.
5. The `useDeleteItems` hook takes a callback function, `fetchItems`, that gets triggered after a `POST` request is made to delete the job.
6. `fetchItems` now fires, and because of the throttle event, is asked to query for a specific job.
7. `fetchItems` makes a request to the API with the following parameters:
```javascript
https://localhost:3001/api/v2/unified_jobs/?id__in=1101&not__launch_type=sync&order_by=-finished&page=1&page_size=20
```
*Note the `id__in=1101&not__launch_type=sync&order_by=-finished&page=1&page_size=20` in the query string. This effectively returns `0` results, which clears out the jobs list page because there are zero jobs to display. 

The fix is to make the `fetchJobsById` function accept a configurable query string. The `useWsJobs` hook takes care of stacking and ordering the jobs the Job List page needs to query the API. The correct call the Job List page needs to be making is `https://localhost:3001/api/v2/unified_jobs/?id__in=1101`.

Incidentally, this also fixes a bug I was seeing where if a user had multiple jobs running and they would paginate away from the first page of the Job List, after a while they would start to see a string of `404` requests from the network tab, because `id__in=1101&not__launch_type=sync&order_by=-finished&page=2&page_size=20` returns an `invalid page` response from the API.

Screen Shot 2021-10-05 at 1 29 31 PM

Screen Shot 2021-10-05 at 10 22 13 AM

Related #10272 
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - UI

##### AWX VERSION

```
19.3.0
```

### DIFF
--- a/awx/ui/src/components/JobList/useWsJobs.js
+++ b/awx/ui/src/components/JobList/useWsJobs.js
@@ -31,7 +31,7 @@ export default function useWsJobs(initialJobs, fetchJobsById, qsConfig) {
         return;
       }
       setJobsToFetch([]);
-      const newJobs = await fetchJobsById(throttledJobsToFetch, {});
+      const newJobs = await fetchJobsById(throttledJobsToFetch);
       const deduplicated = newJobs.filter(
         (job) => !jobs.find((j) => j.id === job.id)
       );
--- a/awx/ui/src/components/JobList/useWsJobs.js
+++ b/awx/ui/src/components/JobList/useWsJobs.js
@@ -31,7 +31,7 @@ export default function useWsJobs(initialJobs, fetchJobsById, qsConfig) {
         return;
       }
       setJobsToFetch([]);
-      const newJobs = await fetchJobsById(throttledJobsToFetch);
+      const newJobs = await fetchJobsById(throttledJobsToFetch, {});
       const deduplicated = newJobs.filter(
         (job) => !jobs.find((j) => j.id === job.id)
       );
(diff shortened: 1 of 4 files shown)


AlexSCorey (approved):
This looks really good.  Well done!

awx/ui/src/components/JobList/useWsJobs.js
```diff
@@ -31,7 +31,7 @@ export default function useWsJobs(initialJobs, fetchJobsById, qsConfig) {
         return;
       }
       setJobsToFetch([]);
-      const newJobs = await fetchJobsById(throttledJobsToFetch);
+      const newJobs = await fetchJobsById(throttledJobsToFetch, {});
```
nixocio: @kialam just a minor suggestion we can rely on `qs` default value of `fetchJobsById`.
```suggestion
      const newJobs = await fetchJobsById(throttledJobsToFetch);
```

nixocio (approved):
@kialam great job, and excellent explanation. 🙇

InnocentK (approved):
Excellent work!

Issue #11114, opened 2021-09-22
Validate/Enforce Runner Version in Execution Nodes


beeankha:
Due to the work that was done to fix https://github.com/ansible/awx/issues/10944, we will need to make sure that a compatible version of Runner is installed on execution nodes.

rooftopcellist:
# Meeting notes (Sept 23)

## Where does ansible-runner get installed in AAP 2.1 2.0-2.1+?

* Alongside receptor on the execution node
* Inside the Execution Environment (EE)

## Problem
We can potentially end up with different versions of ansible-runner in those 2 places which could be incompatible.

## Solution
We add validation before using that EE that it will work with the runner version on the execution node.
We also decided we want to pin ansible-runner to the major version in the places where it is consumed
* pin version in awx/requirements.in
* pin in tower-packaging group_vars to match
* pin major version in Dockerfile for the EE's

**Runner Release Policy**: no breaking changes on minor or patch releases for Ansible-runner

We also talked about releasing ansible-runner 3.0.0 - before we do this, we need to talk to all stakeholders before making this decision.

rooftopcellist:
We should also ensure that execution nodes and controller nodes have compatible versions of ansible-runner.

PR #11213, opened 2021-10-06
Add additional validation against hard-coded ansible-runner version

AlanCoding:
##### SUMMARY
Connect https://github.com/ansible/awx/issues/11114

With this, we will _always_ write a warning to the log if an execution node's runner version differs from the node doing the health check.

The idea here is that users should be able to bump runner versions as needed, but if they do, a log warning should be a comforting validation of the customization they have, themselves, configured.

Second thing here - we hard-code an ansible-runner version in the settings to be a strict requirement. If the execution node does not match the major/minor numbers we fail, and we also require at z-stream or above, because there might be CVEs or bug fixes that we care absolutely need to have in the future.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/utils/receptor.py
+++ b/awx/main/utils/receptor.py
@@ -1,8 +1,13 @@
 import logging
 import yaml
 import time
+from distutils.version import LooseVersion as Version
 
 from receptorctl.socket_interface import ReceptorControl
+
+from ansible_runner.__main__ import VERSION as local_runner_version
+
+from django.conf import settings
 
 from enum import Enum, unique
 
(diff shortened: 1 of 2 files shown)


awx/main/utils/receptor.py
```diff
@@ -152,12 +157,24 @@ def worker_info(node_name, work_type='ansible-runner'):
...
-            data['errors'].append('Worker failed to return keys {}'.format(' '.join(missing_keys)))
+    # If we have errors, we should not go on to do further stricter validation
+    if data['errors']:
+        return data
+
+    # see tasks.py usage of keys
+    missing_keys = set(('runner_version', 'mem_in_bytes', 'cpu_count')) - set(data.keys())
+    if missing_keys:
+        data['errors'].append('Worker failed to return keys {}'.format(' '.join(missing_keys)))
+        return data
```
AlanCoding, line 168: I'm now thinking that this block should come after the version validation block (but obviously don't run the version check if "runner_version" key isn't there)

AlanCoding:
I'm unclear if this remains necessary.

PR #11226, opened 2021-10-08
Changed Work Submission parameter for K8s work

amolgautam25:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/utils/receptor.py
+++ b/awx/main/utils/receptor.py
@@ -3,6 +3,7 @@ import yaml
 import time
 
 from receptorctl.socket_interface import ReceptorControl
+from django.conf import settings
 
 from enum import Enum, unique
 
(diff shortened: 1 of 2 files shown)


shanemcd (approved):

awx/main/utils/receptor.py
```diff
@@ -78,7 +79,8 @@ def run_until_complete(node, timing_data=None, **kwargs):
     kwargs.setdefault('payload', '')
 
     transmit_start = time.time()
-    result = receptor_ctl.submit_work(worktype='ansible-runner', node=node, **kwargs)
+    sign_work = False if settings.IS_K8S else True
```
shanemcd, line 81: look up on line 77... probably want to delete that line now.

AlanCoding (approved):

AlanCoding:
I would like to see some docs explaining why this is done. Let me know if you want some specific ideas of where / how to write it up.

fosterseth (approved):

fosterseth:
@AlanCoding 

There is no established receptor mesh in the k8s setup. Receptor is used to launch pods, and that is about it. As such, signed work is disabled for k8s installations.

https://github.com/ansible/awx-operator/blob/21fe2646b433537538490a2cd6de104bf366de1d/roles/installer/templates/config.yaml.j2#L242

i.e., no verifySignature, work-signing, work-verification, etc.

This could change one day if we ever allow VM installations to coexist in the same mesh as k8s setups.

PR #11242, opened 2021-10-12
Add awx-operator CI check

shanemcd:
I was in the process of rewriting the automation around our release process and this fell out as a separate thing that would be nice to have.

My goal was to have this run faster than our current slowest CI check. It takes about ~15 minutes:

image

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,3 @@
 awx/ui/node_modules
 Dockerfile
+.git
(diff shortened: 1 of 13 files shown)


rebeccahhh (approved):

beeankha (approved):

rooftopcellist (approved):
:tada: thank you for doing this!

PR #11252, opened 2021-10-18
Update/Add Targets that Acquire AWX Version

beeankha:
Changing the Makefile so that:
1) Python 3 vs stricter 3.8 is used
2) Hiding stderr on `make VERSION` target
3) Creating a new target for Collections version since we can't build collections using version output in the format of something like `awx: 19.4.1.dev84+g00faf7a2f1`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ CHROMIUM_BIN=/tmp/chrome-linux/chrome
 GIT_BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD)
 MANAGEMENT_COMMAND ?= awx-manage
 VERSION := $(shell $(PYTHON) setup.py --version)
+COLLECTION_VERSION := $(shell $(PYTHON) setup.py --version | cut -d . -f 1-3)
 
 # NOTE: This defaults the container image version to the branch that's active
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,12 +1,13 @@
-PYTHON ?= python3.8
```
shanemcd: This is most likely why CI is failing.

tools/ansible/roles/dockerfile/templates/Dockerfile.j2
```diff
@@ -36,16 +36,16 @@ RUN dnf -y update && \
     patch \
     @postgresql:12 \
     postgresql-devel \
-    python38-devel \
-    python38-pip \
-    python38-psycopg2 \
-    python38-setuptools \
+    python3-devel \
```
shanemcd: I'm pretty sure that this will resolve to python3.6 on centos 8, which we definitely dont want.

shanemcd (changes requested):
I think almost everything here needs to get reverted. 

But I also see the problem you were trying to solve, and can help you look for a more surgical fix.

shanemcd (approved):

Issue #11012, opened 2021-09-02
Login redirect not working


AlexSCorey:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

When on v2.0EA, the Login redirect override URL is set, the /#/login page can't be reached anymore. Even when a direct URL in put in the browser like https:///#/login, the page redirects to the SAML login page as defined in the redirect override.

This is issue always happens when the Login redirect override URL is set.

Users are unable to login to AAP as a non-social user like the admin user.

### AWX version

devel

### Installation method

docker for mac

### Modifications

no

### Web browser

Chrome

### Steps to reproduce

When on v2.0EA, the Login redirect override URL is set, the /#/login page can't be reached anymore. Even when a direct URL in put in the browser like https:///#/login, the page redirects to the SAML login page as defined in the redirect override.

This issue always happens when the Login redirect override URL is set.

### Expected results

Users are able to login to AAP as a non-social user like the admin user.

### Actual results

Users are unable to login to AAP as a non-social user like the admin user.

nixocio:
See: https://github.com/ansible/awx/pull/5479

tiagodread:
@one-t can we close this issue?

nixocio:
@tiagodread @one-t  this is a legit bug, I already worked on this, I will circle back to this one later.

PR #5479, opened 2019-12-10
Configurable login redirect

ghjm:
##### SUMMARY

Allow users to specify an internal or external URL to which unauthorized users should be redirected.  This will allow SAML and other social auth users to provide a more seamless login experience.  See https://github.com/ansible/awx/issues/1886.

PR #11271, opened 2021-10-21
Fix login redirect

nixocio:
Allows the user to visit login page when the login redirect url is set.

Also, redirects to login page once logging out and there is session from
a SAML available.

See: https://github.com/ansible/awx/issues/11012

### DIFF
--- a/awx/ui/src/App.js
+++ b/awx/ui/src/App.js
@@ -98,8 +98,13 @@ const AuthorizedRoutes = ({ routeConfig }) => {
   );
 };
 
-const ProtectedRoute = ({ children, ...rest }) => {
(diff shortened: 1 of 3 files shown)


nixocio:
I would suggest testing the following scenarios:

https://github.com/ansible/awx/pull/5479#issuecomment-564740716

Also, make sure that the  `SOCIAL_AUTH_USER_FIELDS` is set to `null` when testing this PR. When updating the LOGIN_REDIRECT_OVERRIDE to SAML provide - there is another bug that is setting `SOCIAL_AUTH_USER_FIELDS` to `[]` use the API and modify to `null`.


image

PR #11273, opened 2021-10-22
Added support to import manual projects

eitanp82:
msg: ""

##### SUMMARY

Import manual project marked as failed even when is completed successfully.
This issue may cause to wait until timeout expired.

Here's an example output:
```
Object import failed: Project-13 has status of ok, which is not in ['successful', 'failed', 'error', 'canceled'].
TIME WHEN STATUS WAS FOUND: 2021-10-21 12:59:21.217321 (UTC)
```
The suggested solution based on the following code from the [awx/main/models/unified_jobs.py](https://github.com/ansible/awx/blob/devel/awx/main/models/unified_jobs.py) module:
```
    PROJECT_STATUS_CHOICES = COMMON_STATUS_CHOICES + [
        ('ok', _('OK')),  # Project is not configured for SCM and path exists.
        ('missing', _('Missing')),  # Project path does not exist.
    ]
```

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
awx: 19.4.1.dev91+g8a7cd911aa.d20211022
```

### DIFF
--- a/awxkit/awxkit/api/pages/api.py
+++ b/awxkit/awxkit/api/pages/api.py
@@ -240,7 +240,8 @@ class ApiV2(base.Base):
                         # When creating a project, we need to wait for its
                         # first project update to finish so that associated
                         # JTs have valid options for playbook names
-                        _page.wait_until_completed()
+                        status = _page.completed_statuses + ['ok']
+                        _page.wait_until_completed(status=status)
                 else:
                     _page = _page.put(post_data)
                     changed = True
--- a/awxkit/awxkit/api/mixins/has_status.py
+++ b/awxkit/awxkit/api/mixins/has_status.py
@@ -34,9 +34,9 @@ class HasStatus(object):
             self.assert_status(status)
         return self
 
-    def wait_until_completed(self, interval=5, timeout=60, **kwargs):
+    def wait_until_completed(self, interval=5, timeout=60, *, status=None, **kwargs):
         start_time = datetime.utcnow()
-        HasStatus.wait_until_status(self, self.completed_statuses, interval=interval, timeout=timeout, **kwargs)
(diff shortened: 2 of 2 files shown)


awxkit/awxkit/api/pages/api.py
```diff
@@ -240,7 +240,8 @@ def _import_list(self, endpoint, assets):
                         # When creating a project, we need to wait for its
                         # first project update to finish so that associated
                         # JTs have valid options for playbook names
-                        _page.wait_until_completed()
+                        status = _page.completed_statuses + ['ok']
```
AlanCoding, line 243: maybe it would be cleaner to just add "ok" to the list of `completed_statuses`?

eitanp82: I agree and I actually did it in my code as workaround since I'm using this patch for import manual projects only.

I didn't do it here because 'ok' is not part of either [JOB_STATUS_CHOICES](https://github.com/ansible/awx/blob/devel/awx/main/models/unified_jobs.py#L72) or [COMMON_STATUS_CHOICES](https://github.com/ansible/awx/blob/devel/awx/main/models/unified_jobs.py#L83).
This change will affect on all operations and may cause to unexpected result when using a different operation (e.g., launching a job template).

If it's ok by you, I will update the PR with your suggestion.  

Thanks
Eitan

djyasin:
Hi!

Thank you very much for your submission to AWX. It means a lot to us that you have taken the time to contribute.

On this PR, changes were requested but it has been some time since then. We think this PR has merit but without the requested changes we are unable to merge it. At this time we are closing your PR. If you get time to address the changes you are welcome to open another PR or we can reopen this PR upon request if you contact us by using any of the communication methods listed in the page below:

https://github.com/ansible/awx/#get-involved

Thank you once again for this and your interest in AWX!

PR #11274, opened 2021-10-22
Allow configuring sources_dest value other than ../_sources

notok:
Currently it is not possible to configure sources_dest to the value other than `../_sources` because the value is hard-coded in defaults file and docker-compose template file.
This PR makes it possible by using variables instead of hard-coding, and remove unnecessary paths.

### DIFF
--- a/tools/docker-compose/ansible/roles/sources/defaults/main.yml
+++ b/tools/docker-compose/ansible/roles/sources/defaults/main.yml
@@ -13,6 +13,6 @@ receptor_image: quay.io/ansible/receptor:devel
 # Keys for signing work
 receptor_rsa_bits: 4096
 receptor_work_sign_reconfigure: false
-work_sign_key_dir: '../_sources/receptor'
+work_sign_key_dir: '{{ sources_dest }}/receptor'
 work_sign_private_keyfile: "{{ work_sign_key_dir }}/work_private_key.pem"
 work_sign_public_keyfile: "{{ work_sign_key_dir }}/work_public_key.pem"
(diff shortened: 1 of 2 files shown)


john-westcott-iv:
Thank you for this contribution. We are going back over old PRs and found that this was never responded to, our apologies for taking so long. What was your rationale for trying to variablize the _sources directory? We are having a hard time justifying changing this because its for the development process and there may be other areas where the _sources directory was referenced.

notok:
Thank you for your response.
My first intention was to change `sources_dest` to another directory just for my test. (i.e. create docker-compose.yml and other files to another directory, not `../_sources`.)

(I understand these tools are only for development and other developers usually don't change _sources path.)

In the process of trying that, I found followings and fixed.

1. The variable `sources_dest` is not used for value including `'../_sources'`.
```diff
-work_sign_key_dir: '../_sources/receptor'
+work_sign_key_dir: '{{ sources_dest }}/receptor'
```
2. unnecessary paths (`../../docker-compose/_sources` is the directory where docker-compose.yml exists.)
```diff
-      - "../../docker-compose/_sources/database.py:/etc/tower/conf.d/database.py"
+      - "./database.py:/etc/tower/conf.d/database.py"
```
3. relative paths that does not work when _sources path was changed.
```diff
-      - "../../../:/awx_devel"
+      - "{{ (playbook_dir + '/../../../') | realpath() }}:/awx_devel"
```

I think 1. and 2. is acceptable for you as a refactoring (I think it does not affect dev process), but 3. may be a bit controversial because the path will be full path.
Comments are welcome.


As formerly mentioned, I understand these are for development,
And if you don't want to change these not to break dev process, I will close this.
Or, I will re-check PR and resolve conflicts.

djyasin:
Thank you so much for your contribution! We are closing this particular PR and work will continue here: 
https://github.com/ansible/awx/pull/13682

PR #11282, opened 2021-10-26
Upgrade d3 to v7.

kialam:
##### SUMMARY

Upgrade D3 to v7.

##### COMPONENT NAME

 - UI

### DIFF
--- a/awx/ui/package.json
+++ b/awx/ui/package.json
@@ -101,6 +101,9 @@
     "coveragePathIgnorePatterns": [
       "<rootDir>/src/locales",
       "index.js"
+    ],
+    "transformIgnorePatterns": [
+      "<rootDir>/node_modules/(?!d3)/"
     ]
   }
 }
--- a/awx/ui/package.json
+++ b/awx/ui/package.json
@@ -15,7 +15,7 @@
     "axios": "0.22.0",
     "babel-plugin-macros": "^3.0.1",
     "codemirror": "^5.47.0",
-    "d3": "6.7.0",
+    "d3": "7.1.1",
     "dagre": "^0.8.4",
     "formik": "2.2.9",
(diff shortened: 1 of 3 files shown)


awx/ui/package.json
```diff
@@ -101,6 +101,9 @@
     "coveragePathIgnorePatterns": [
       "<rootDir>/src/locales",
       "index.js"
+    ],
+    "transformIgnorePatterns": [
+      "<rootDir>/node_modules/(?!d3)/"
```
kialam, line 106: It wasn't easy to track this down, however following a trail of release logs eventually led me to the solution on how to get jest to work again.

In short, [d3 v7](https://github.com/d3/d3/releases/tag/v7.0.0) introduces a new version of their sub packages (d3-array, d3-hierarchy, etc). [v3](https://github.com/d3/d3-array/releases/tag/v3.0.0) of these packages have adopted a [new module type](https://github.com/d3/d3/issues/3501). These packages are now [ESM modules ](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#im-having-problems-with-esm-and-jest)instead of CommonJS. [Jest ships with experimental support of ESM modules](https://github.com/facebook/jest/blob/64de4d7361367fd711a231d25c37f3be89564264/docs/ECMAScriptModules.md), and in order for the tests to properly import the packages when the tests are ran is to disable code transformation from within the d3 directory.

nixocio:
Good job @kialam, the e2e test failures are also present on devel. Not related to this change.

nixocio (approved):

nixocio:
See: https://github.com/ansible/awx/issues/10594

tiagodread (approved):
I compared the broken tests with the broken tests in devel and they are the same, they are also not failing on downstream build.

Issue #11326, opened 2021-11-09
In Survey Spec if the choices is an array of strings the UI fails


mkanoor:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

In the AAP UI when choices are added for a multi select field they are stored as a string with \n as the delimiter. e.g
`choices: "NJ\nPA\OK"`
The REST API allows for the choices to be stored as an array of strings e.g.
```
"choices": [
        "NJ",
        "PA",
        "OK"
      ]
```

So if someone uses a REST API to add a Survey spec and specifies an array/list of strings the REST API call succeeds and stores the value but when the UI tries to display it fails.

### AWX version

AWX 4.0.0

### Installation method

openshift

### Modifications

no

### Operating system

Linux

### Web browser

Chrome

### Steps to reproduce

Create a Job Template 
Add a Survey Spec for the Job Template using a REST API with the following payload

```
POST https://{{TowerIP}}/api/v2/job_templates/{{job_template_id}}/survey_spec/
BODY

  "name": "",
  "description": "",
  "spec": [
    {
      "question_name": "State",
      "question_description": "The state where you live",
      "required": true,
      "type": "multiplechoice",
      "variable": "state",
      "min": 0,
      "max": 1024,
      "new_question": false,
      "default": "",
      "choices": [
        "NJ",
        "PA",
        "OK"
      ]
    }
  ]
}
```

After the Survey Spec has been created click on the Survey Spec "state" field and the UI gets an error
<img width="1229" alt="Screen Shot 2021-11-09 at 4 15 49 PM" src="https://user-images.githubusercontent.com/6452699/141007197-4f6d27e2-7280-4cb6-89a8-6e417b740073.png">

Survey Spec 
<img width="1211" alt="Screen Shot 2021-11-09 at 4 16 19 PM" src="https://user-images.githubusercontent.com/6452699/141007271-ddab072b-d008-4a04-ae08-842783ecae9b.png">

### Expected results

The UI should be able to display the survey spec item or reject the REST API call when arrays are used for choices.

### Actual results

The REST API call succeeds for both GET and POST  but the UI fails.

jbradberry:
[...]

PR #11328, opened 2021-11-10
fix UI failure if `choices` in Survey Spec is of type `list`

joshiayush:
Handle type `list` when using a REST API to add a Survey Spec

This commit fixes issue #11326

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API


jbradberry:
@joshiayush Looks good, but it could use a test or two I think.  Probably in here: https://github.com/ansible/awx/blob/devel/awx/main/tests/functional/api/test_survey_spec.py

jbradberry (changes requested):
As noted, it needs a test or two.

jbradberry:
Author is unresponsive, closing.

PR #11342, opened 2021-11-14
Allow for running AWX at non-root path (URL prefixing)

shanemcd:
##### SUMMARY
These changes allow for running AWX at a non-root path, like `https://example.com/awx/`:

Screen Shot 2021-11-14 at 12 38 15 PM

Must be merged along with https://github.com/ansible/awx-operator/pull/652

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -391,7 +391,20 @@ class BaseSerializer(serializers.ModelSerializer, metaclass=BaseSerializerMetacl
         view = self.context.get('view', None)
         if view and (hasattr(view, 'retrieve') or view.request.method == 'POST') and type(obj) in settings.NAMED_URL_GRAPH:
             original_url = self.get_url(obj)
-            res['named_url'] = self._generate_named_url(original_url, obj, settings.NAMED_URL_GRAPH[type(obj)])
+
+            # If the app is running at a location other than /, temporarily remove the
+            # prefix. This is to avoid changing the code in _generate_named_url where
+            # it is assumed the ID of the resource is at url_units[4].
+            url_prefix = view.request.META.get('SCRIPT_NAME', '')
+            if url_prefix:
+                original_url = removeprefix(original_url, url_prefix)
+
+            named_url = self._generate_named_url(original_url, obj, settings.NAMED_URL_GRAPH[type(obj)])
+
+            if url_prefix:
+                named_url = url_prefix + named_url
+
+            res['named_url'] = named_url
         if getattr(obj, 'created_by', None):
             res['created_by'] = self.reverse('api:user_detail', kwargs={'pk': obj.created_by.pk})
         if getattr(obj, 'modified_by', None):
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -379,32 +379,22 @@ class BaseSerializer(serializers.ModelSerializer, metaclass=BaseSerializerMetacl
     def _get_related(self, obj):
         return {} if obj is None else self.get_related(obj)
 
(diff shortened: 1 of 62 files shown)


shanemcd:
Added instructions on how to test at https://github.com/ansible/awx-operator/pull/652#issue-780017026

pabelanger:
\o/

sean-m-sullivan:
So I looked at this, and it looks brilliant, but I wonder how its going to affect the awx.awx collection, does it effect how the url is generated there?

AlanCoding:
If a client does splitting on the `/` characters, it should split starting from the right. However, we still have testing we should still run for this.

awx/api/serializers.py
```diff
@@ -390,7 +391,20 @@ def get_related(self, obj):
         view = self.context.get('view', None)
         if view and (hasattr(view, 'retrieve') or view.request.method == 'POST') and type(obj) in settings.NAMED_URL_GRAPH:
             original_url = self.get_url(obj)
-            res['named_url'] = self._generate_named_url(original_url, obj, settings.NAMED_URL_GRAPH[type(obj)])
+
+            # If the app is running at a location other than /, temporarily remove the
+            # prefix. This is to avoid changing the code in _generate_named_url where
+            # it is assumed the ID of the resource is at url_units[4].
```
AlanCoding: this seems like a hacky way to solve the problem, as opposed to modifying `_generate_named_url` itself.

tools/ansible/roles/dockerfile/templates/supervisor.conf.j2
```diff
@@ -28,7 +28,7 @@ directory = /awx_devel
 environment =
     UWSGI_DEV_RELOAD_COMMAND='supervisorctl -c /etc/supervisord_task.conf restart all; supervisorctl restart tower-processes:daphne tower-processes:wsbroadcast'
 {% else %}
-command = /var/lib/awx/venv/awx/bin/uwsgi --socket 127.0.0.1:8050 --module=awx.wsgi:application --vacuum --processes=5 --harakiri=120 --no-orphans --master --max-requests=1000 --master-fifo=/var/lib/awx/awxfifo --lazy-apps -b 32768
+command = /var/lib/awx/venv/awx/bin/uwsgi /etc/tower/uwsgi.ini
```
AlanCoding, line 31: oh this is very nice

awx/api/serializers.py
shanemcd: Suggestions? We don't have access to the request object there. Sending the index in as an argument seems equally weird to me.

shanemcd: @AlanCoding Please take a look at my latest commit.

rooftopcellist:
It looks like removing this import is what is needed to make the linters happy:

```
awx/api/serializers.py:100:1: F401 'awx.main.utils.removeprefix' imported but unused
```

AlanCoding (approved):

tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
```diff
@@ -56,6 +56,7 @@ services:
       - "8013:8013"  # http
       - "8043:8043"  # https
       - "2222:2222"  # receptor foo node
+      - "3001:3001"  # used by the UI dev env
```
jakemcdermott:
```suggestion
      - "3000:3001"  # used by the UI dev env
```

jakemcdermott (approved):

adpavlov:
It seems that Mesh.js also needs to be changed to remove leading / in

`this.baseUrl = '/api/v2/mesh_visualizer/';`

@shanemcd

PR #11345, opened 2021-11-15
Added an example for credential_type

Akasurde:
##### SUMMARY
Added an example of custom credential type using inputs
as YAML spec.

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

##### ISSUE TYPE

 - Docs Pull Request

##### COMPONENT NAME

 - UI

##### AWX VERSION

```
devel
```

### DIFF
--- a/awx_collection/plugins/modules/credential_type.py
+++ b/awx_collection/plugins/modules/credential_type.py
@@ -71,6 +71,26 @@ EXAMPLES = '''
     state: present
     validate_certs: false
 
+- name: Custom Credential type with inputs as a YAML
+  credential_type:
+      name: Nexus
+      description: Credentials type for Nexus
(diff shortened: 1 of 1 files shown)


awx_collection/plugins/modules/credential_type.py
```diff
@@ -71,6 +71,26 @@
     state: present
     validate_certs: false
 
+- name: Custom Credential type with inputs as a YAML
+  credential_type:
+      name: Nexus
```
shanemcd, line 76: The indentation looks off here. Please update this to be consistent with the other examples in this file.

shanemcd:
Looks like this fell through the cracks and abandoned. Feel free to reopen if you get around to fixing it.

Issue #11352, opened 2021-11-16
Link to instance group from job template details page


akus062381:
I was doing some instance group testing as it related to job templates. I wanted to click on the name of the associated instance group, to link to that IG, but the name is not a clickable link on the job template details page.

![image](https://user-images.githubusercontent.com/9053044/142062518-b6b6cf67-9f19-421f-a723-e348c4a99293.png)

PR #11355, opened 2021-11-16
Linkify instance/container groups job template details

nixocio:
Linkify instance/container groups job template

See: https://github.com/ansible/awx/issues/11352

### DIFF
--- a/awx/ui/src/screens/Template/JobTemplateDetail/JobTemplateDetail.js
+++ b/awx/ui/src/screens/Template/JobTemplateDetail/JobTemplateDetail.js
@@ -167,6 +167,11 @@ function JobTemplateDetail({ template }) {
     );
   };
 
+  const buildLinkURL = (instance) =>
+    instance.is_container_group
+      ? '/instance_groups/container_group/'
+      : '/instance_groups/';
+
   if (instanceGroupsError) {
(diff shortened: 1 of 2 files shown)


awx/ui/src/screens/Template/JobTemplateDetail/JobTemplateDetail.js
```diff
@@ -167,6 +167,11 @@ function JobTemplateDetail({ template }) {
     );
   };
 
+  const buildLinkURL = (instance) =>
```
AlexSCorey, line 170: I think my preference would be to just build the whole path here instead of building some of it here, and the rest in the jsx.  Sort of like we do with `detailsURL` in other places.

AlexSCorey (changes requested):
Just a nit picky change.

awx/ui/src/screens/Template/JobTemplateDetail/JobTemplateDetail.js
nixocio: We already have this function in other parts of the code. Since this function is already being tested. Lets keep consistent.

akus062381 (approved):
Looks good!

AlexSCorey (approved):
Looks good

Issue #10337, opened 2021-06-02
job event optimization - Do not null job event relationship when deleting


chrismeyersfsu:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
There is a relationship from job_event -> host. This relationship is nulled when a host is deleted. The relationship is also updated when inventory is deleted. This can be an expensive operation. We should try and remove it. Related https://github.com/ansible/awx/issues/10331

### TODO Investigate
- [ ] From a technical perspective. Does this relationship need to be consistent?
- [ ] From a product perspective. What user-facing (i.e. UI or even API) features rely on this relationship being consistent? For example, do we do any queries that count or look at the last failed event for an inventory that changes when/if a related host' job events are deleted?

https://github.com/ansible/awx/blob/devel/awx/main/tasks.py#L741 Consider can/how to remove this. Even if we don't remove this, it seems inefficient to do this the way we are. Surely we do not need to loop over each host to perform the update.
https://github.com/ansible/awx/blob/devel/awx/main/models/events.py#L467 `on_delete=models.SET_NULL,` -> do nothing
https://github.com/ansible/awx/blob/devel/awx/main/models/events.py#L743 `on_delete=models.SET_NULL,` -> do nothing

AlanCoding:
I believe this would address what happened here as well https://github.com/ansible/awx/issues/12176

Would also address slow host deletions described in https://github.com/ansible/awx/issues/12277... probably

I think I filed a duplicate issue with this, I'm looking for it.

AlanCoding:
I wrote down some notes on the error related to this at https://gist.github.com/AlanCoding/b9919c1b1ebbb9ee6b5be2a584362200

AlanCoding:
I keep failing to accurately search for this issue, because the exact syntax of models.DO_NOTHING was not used. Hopefully that is resolved now.

edit: DO_NOTHING

AlanCoding:
Closing as @john-westcott-iv's merge appears to address this.

PR #11367, opened 2021-11-18
Add unique host target retention

bhavenst:
New main_uniquehosts db table and config flag

##### SUMMARY
Addition of UniqueHost model and db table with optional flag.  Ensures target execution hosts are retained when jobs are deleted.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - DB

### DIFF
--- a/awx/main/migrations/0156_unique_host.py
+++ b/awx/main/migrations/0156_unique_host.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('main', '0155_improved_health_check'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='UniqueHost',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('created', models.DateTimeField(default=None, editable=False)),
+                ('modified', models.DateTimeField(default=None, editable=False)),
+                ('host_name', models.CharField(default='', max_length=1024, editable=False)),
+                (
+                    'host',
+                    models.ForeignKey(
+                        related_name='unique_hosts', on_delete=django.db.models.deletion.SET_NULL, default=None, editable=False, to='main.Host', null=True
+                    ),
+                ),
+                (
+                    'job_host_summary',
+                    models.ForeignKey(
+                        related_name='unique_hosts',
+                        on_delete=django.db.models.deletion.SET_NULL,
+                        default=None,
+                        editable=False,
+                        to='main.JobHostSummary',
+                        null=True,
+                    ),
+                ),
+                ('recorded_date', models.DateTimeField(default=None, editable=False, null=True)),
+            ],
+            options={
+                'ordering': ('-pk',),
+                'verbose_name_plural': 'unique hosts',
+            },
+        ),
+    ]
--- a/awx/main/models/events.py
+++ b/awx/main/models/events.py
@@ -544,6 +544,9 @@ class JobEvent(BasePlaybookEvent):
             from awx.main.models import Host, JobHostSummary  # circular import
             from awx.main.models import Host, JobHostSummary, HostMetric
 
+            if settings.UNIQUE_HOST_RETENTION_ENABLED:
+                from awx.main.models import UniqueHost
+
             all_hosts = Host.objects.filter(pk__in=self.host_map.values()).only('id', 'name')
             existing_host_ids = set(h.id for h in all_hosts)
 
@@ -578,6 +581,13 @@ class JobEvent(BasePlaybookEvent):
                 if h.id in host_mapping:
                     h.last_job_host_summary_id = host_mapping[h.id]
                     updated_hosts.add(h)
+
+                # Populate unique host retention table if enabled
+                if settings.UNIQUE_HOST_RETENTION_ENABLED:
+                    # If host already exists, update it, else create it
+                    UniqueHost.objects.update_or_create(
+                        host_name=h.name, defaults={'created': now(), 'modified': now(), 'job_host_summary_id': h.last_job_host_summary_id, 'host_id': h.id}
+                    )
 
             Host.objects.bulk_update(list(updated_hosts), ['last_job_id', 'last_job_host_summary_id'], batch_size=100)
 
--- a/awx/main/models/jobs.py
+++ b/awx/main/models/jobs.py
@@ -1088,6 +1088,53 @@ class JobHostSummary(CreatedModifiedModel):
         super(JobHostSummary, self).save(*args, **kwargs)
 
 
+class UniqueHost(CreatedModifiedModel):
+    """
+    Retention of unique host targets (optional)
+    """
+
+    class Meta:
+        app_label = 'main'
+        constraints = [models.UniqueConstraint(name='ensure_unique_hosts', fields=['host_name'])]
+        verbose_name_plural = _('unique hosts')
+        ordering = ('-pk',)
+
+    job_host_summary = models.ForeignKey(
+        'JobHostSummary',
+        related_name='unique_hosts',
+        null=True,
+        default=None,
+        on_delete=models.SET_NULL,
+        editable=False,
+    )
+
+    host = models.ForeignKey(
+        'Host',
+        related_name='unique_hosts',
+        null=True,
+        default=None,
+        on_delete=models.SET_NULL,
+        editable=False,
+    )
+
+    host_name = models.CharField(
+        max_length=1024,
+        default='',
+        editable=False,
+    )
+
+    recorded_date = models.DateTimeField(
+        default=None,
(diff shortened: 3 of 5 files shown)


awx/main/models/events.py
```diff
@@ -579,6 +582,13 @@ def _update_host_summary_from_stats(self, hostnames):
                     h.last_job_host_summary_id = host_mapping[h.id]
                     updated_hosts.add(h)
 
+                # Populate unique host retention table if enabled
+                if settings.UNIQUE_HOST_RETENTION_ENABLED:
+                    # If host already exists, update it, else create it
+                    UniqueHost.objects.update_or_create(
```
shanemcd, line 588: This is really critical part of the application in terms of performance. If you see below, we do a `bulk_update` because we found individual insertions to be problematic.

shanemcd:
@matburt Can you help Bryan work through this while I'm out on vacation? One idea he had was to do create a database trigger that fires when new host records are inserted. I am not a database expert, but it seems like that might be more performant than handling this synchronously in the [danger zone](https://github.com/ansible/awx/blob/f60f68af8fe360435d002d8a9a86c5fc40d5e7a5/awx/main/models/events.py#L408).

bhavenst:
> 

Hi, alternate implementation using trigger is here:
https://github.com/ansible/ansible-automation-platform-azure-billing/blob/main/azure_billing/db.py

matburt:
In the general sense we don't use triggers, so if we can stay away from those it'll work best.

bhavenst:
OK, Matt says no triggers and Shane says no queries in AWX. :)  I did consider bulk_update, but since there is a unique constraint on the host (it is unique hosts after all) I didn't go with it.  But I suppose it would be possible to load the list of current unique hosts and bulk save any new ones not found in that list.  The initial query could return a lot of hosts, but maybe that is better than a query per host..

awx/main/models/events.py
```diff
@@ -579,6 +582,13 @@ def _update_host_summary_from_stats(self, hostnames):
...
+                        host_name=h.name, defaults={'created': now(), 'modified': now(), 'job_host_summary_id': h.last_job_host_summary_id, 'host_id': h.id}
```
matburt, line 589: In this case isn't `created` and `modified` automatically applied from the base class? Is there a reason we need to call  `now()` explicitly at this level?

awx/main/models/jobs.py
```diff
@@ -1088,6 +1088,53 @@ def save(self, *args, **kwargs):
...
 
 
+class UniqueHost(CreatedModifiedModel):
+    """
+    Retention of unique host targets (optional)
+    """
+
+    class Meta:
+        app_label = 'main'
+        constraints = [models.UniqueConstraint(name='ensure_unique_hosts', fields=['host_name'])]
```
matburt, line 1098: Is this preferredd rather than setting it on the `host` field below as `unique=True` ?

```
host = models.ForeignKey(
  'Host',
  [...]
  unique=True,
  [...]
```

awx/main/models/events.py
matburt: I *think* the explicit act of calling save is all you need to do?

https://github.com/ansible/awx/blob/ba324c73ceefa22d516d872527649d090d3bc0cc/awx/main/models/base.py#L166

or is that not called explicitly when it's happening in bulk?

matburt:
I think this is a good start, I'd like to understand how this fits into the larger solution that we're providing. I'm not so sure about wrapping this up under a setting that turns it on. It seems like that's not really saving much.

awx/main/models/events.py
bhavenst: Apparently so, removed the created/modified.  Thanks

awx/main/models/jobs.py
bhavenst: The reason for the new table to begin with is that the hosts and jobs can be deleted anytime, at which point they are removed entirely from the DB.  This table does have links to the host and job summary tables, but just for information.  When those hosts/jobs are deleted, the links are removed but the rows remain in this table until specifically cleaned up.  So yes that would work, but it wouldn't meet the goal.  The unique constraint is placed on the hostname field since it is independent of the foreign keys..

bhavenst:
I'll fix the repeated query issue and the now() calls and make it non-optional (I think that was suggested by Shane, but I'm fine with removing it).

awx/main/models/jobs.py
matburt: Honesty, I was asking more about the *semantics* of defining the unique constraint there rather than at the table definition

matburt:
The other discussion we might should have is.... should we *ever* clean this table up? In our original discussion we were talking about the `JobHotSummary` table which might need to be cleaned up at some point. This one might not need to be, but it might end up growing forever if you have a large transient pool of infrastructure.

Some of my earlier thoughts on this had this being cleaned up after a year or so... or a few years.

awx/main/migrations/0156_unique_host.py
```diff
@@ -0,0 +1,47 @@
...
+                    'host',
+                    models.ForeignKey(
+                        related_name='unique_hosts', on_delete=django.db.models.deletion.SET_NULL, default=None, editable=False, to='main.Host', null=True
+                    ),
+                ),
+                (
+                    'job_host_summary',
+                    models.ForeignKey(
+                        related_name='unique_hosts',
+                        on_delete=django.db.models.deletion.SET_NULL,
```
chrismeyersfsu, line 33: Even setting this field to NULL on delete can become expensive. Can we instead deal with this value being "wrong" ?

See https://github.com/ansible/awx/issues/10337

awx/main/migrations/0156_unique_host.py
```diff
@@ -0,0 +1,47 @@
...
+            name='UniqueHost',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('created', models.DateTimeField(default=None, editable=False)),
+                ('modified', models.DateTimeField(default=None, editable=False)),
+                ('host_name', models.CharField(default='', max_length=1024, editable=False)),
+                (
+                    'host',
+                    models.ForeignKey(
+                        related_name='unique_hosts', on_delete=django.db.models.deletion.SET_NULL, default=None, editable=False, to='main.Host', null=True
```
chrismeyersfsu, line 26: Same as below. When a host is deleted, setting this relationship to NULL can be expensive because .. django.

https://github.com/ansible/awx/issues/10337

chrismeyersfsu:
How many `UniqueHost` entries are we building for?

If I had to throw out a number I'd say 100,000 is a good "large" and I could see looking into the performance of 1,000,000

I'd like to see how the callback receiver performs with the numbers that we chose to build for. I think this will point out quickly, for example, the need for an index on `host_name`. I'm creating a tech debt issue card for the callback receiver so hopefully I can link you to how to observe the performance of the callback receiver while developing.

This PR shows the creation of the per-host automation history data .. but what does the other side look like? What kind of "reporting" queries will be ran against this data? I presume `recorded_date` will be a key range to pivot on, it's probably going to need an index too. Will it need to be a limit offset style query as to not blow out the python memory space? I could see it even being a more raw SQL COPY style query. Regardless, we should understand what it looks like so that we can performance test it.

awx/main/models/jobs.py
bhavenst: Yeah I'm not sure what the difference is.  I'll add it to the field definition instead..  But also based on the subsequent comments and risk of impacting performance, I'm thinking maybe just removing the links to the host and job summary tables entirely is the way to go.  In that case it will just be host_name and recorded date, which don't have any dependency on other tables.

awx/main/migrations/0156_unique_host.py
bhavenst: I think I can just remove the links to the job host summary and host tables entirely as they aren't really providing any value other than temporary traceability.

bhavenst:
The requirement is to retain the records for a month.  I'm not sure how many hosts could be accumulated by large customers in a month though.  Cleanup of the data after a month has elapsed can be done or not done as desired.  There is no particular requirement to immediately remove it or retain it long term, so whatever makes sense.  I was planning to add an API for cleaning the table (probably with a "remove older than x days or records before y date" params).  The "reporting" query is just checking for entries with null for recorded date, reporting them and then setting the recorded date.  That is it.  I'll add indices for date and host_name as you suggest, thanks.  New commit coming soon..

matburt:
We have a few different cleanup commands that run periodically to tidy up things like this, for example https://github.com/ansible/awx/blob/fb1c97cdc1c2134c0ba849725ed1fd0ce74d7e8c/awx/main/management/commands/cleanup_sessions.py

bhavenst:
Hi!  I'll abandon this PR.  The new approach can be found here:
https://github.com/ansible/ansible-automation-platform-azure-billing/pull/4

Please take a look at that one if you could and let me know if you see anything worrying.  It accesses the controller DB in a read-only way and copies/stores the billing data in a separate DB, so I don't think it should be objectionable..  In the longer term, changes can be made to retain the data on the controller side since this approach still allows for job deletion between execution and billing time..  Thanks!

PR #11376, opened 2021-11-19
Consume job_explanation we sometimes set in ansible-runner

AlanCoding:
##### SUMMARY
Verified with tests

https://github.com/ansible/ansible-runner/blob/ef8edc3832a02786d309941b72ed805f90902b7e/ansible_runner/streaming.py#L113

Screenshot from 2021-11-06 21-20-31

This was encountered in https://github.com/ansible/tower/issues/5423, but is an error handling issue

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/tasks/callback.py
+++ b/awx/main/tasks/callback.py
@@ -203,12 +203,12 @@ class RunnerCallback:
                 if os.path.exists(key_data_file) and stat.S_ISFIFO(os.stat(key_data_file).st_mode):
                     os.remove(key_data_file)
         elif status_data['status'] == 'error':
-            result_traceback = status_data.get('result_traceback', None)
-            if result_traceback:
-                from awx.main.signals import disable_activity_stream  # Circular import
-
-                with disable_activity_stream():
-                    self.instance = self.update_model(self.instance.pk, result_traceback=result_traceback)
+            updates = {}
+            for potential_field in ('result_traceback', 'job_explanation'):
+                if status_data.get(potential_field, None):
+                    updates[potential_field] = status_data[potential_field]
+            if updates:
+                self.instance = self.update_model(self.instance.pk, **updates)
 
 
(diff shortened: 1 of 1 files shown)


AlanCoding:
Looking at this more, I think we need to append to `job_explanation` in other places, in case it got set multiple times.

awx/main/tasks.py
```diff
@@ -1394,12 +1394,12 @@ def status_handler(self, status_data, runner_config):
             with disable_activity_stream():
                 self.instance = self.update_model(self.instance.pk, job_args=json.dumps(runner_config.command), job_cwd=runner_config.cwd, job_env=job_env)
         elif status_data['status'] == 'error':
-            result_traceback = status_data.get('result_traceback', None)
-            if result_traceback:
-                from awx.main.signals import disable_activity_stream  # Circular import
-
-                with disable_activity_stream():
```
jbradberry: Do we want to capture this kind of automated update in the activity stream?

AlanCoding: the activity stream won't capture entries that modify fields with editable=False, which was the case for both of these.

Really, all unified job fields should be non-editable, but the _original_ implementation of the API was that you did a POST to create the job record, and then you did a POST to the `/api/v2/jobs/N/start/` endpoint. Because of that, you could still edit the job fields while status=new, meaning that every field related to the user values or the job template were editable. We don't allow this anymore, but the fields are what they are. Anyway, neither `result_traceback` or `job_explanation` would have fit that prior definition of editability, because the user wouldn't be able to set them.

I tested it and it did seem to not produce an activity stream entry. But you know, a unit test would be good for this.

shanemcd:
cherry-picked this into https://github.com/ansible/awx/pull/12089

PR #11377, opened 2021-11-19
 Update how ui tests are invoked on CI

nixocio:
Update how ui tests are invoked on CI as an attempt to speed up test
run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -418,7 +418,7 @@ ui-lint:
 
 ui-test:
 	$(NPM_BIN) --prefix awx/ui install
-	$(NPM_BIN) run --prefix awx/ui test -- --coverage --maxWorkers=4 --watchAll=false
+	$(NPM_BIN) run --prefix awx/ui test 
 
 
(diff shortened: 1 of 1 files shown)


keithjgrant:
have you done any benchmarking here to see if this helps? When I did test perf work this summer, I found it worked best if I left off the maxWorkers flag and let it determine automatically -- but that was for local test runs, so it could be totally different in CI

nixocio:
@keithjgrant I did several test runs with different combinations, let me try this one that you mentioned.

nixocio:
@keithjgrant not setting the number of workers generated the best results so far.

image

image

AlexSCorey (approved):
Looks good

jakemcdermott (approved):

nixocio:
Currently we are not enforcing any rule related to coverage and we can bring back if necessary.  Also, we to be mindful of the changes to code coverage with local runs. Removal of coverage on CI seems to help improve tests speed.

PR #11378, opened 2021-11-19
Update example command for running test container

jakemcdermott:
Quick update to the docs so that the example command references the right network.


AlexSCorey (approved):
Looks good

PR #11381, opened 2021-11-22
Consolidate github workflow logic for CI

AlanCoding:
##### SUMMARY
This is modeled after docs here:

https://docs.github.com/en/actions/learn-github-actions/managing-complex-workflows

```yaml
jobs:
  setup:
    runs-on: ubuntu-latest
    steps:
      - run: ./setup_server.sh
  build:
    needs: setup
    runs-on: ubuntu-latest
    steps:
      - run: ./build_server.sh
  test:
    needs: build
    runs-on: ubuntu-latest
    steps:
      - run: ./test_server.sh
```

This is just trying to eliminate duplication in the workflow logic.

I still intend to look into reusing workflows to share this across multiple workflows (as opposed to ci.yml, which is _one_ workflow) using docs here:

https://docs.github.com/en/actions/learn-github-actions/reusing-workflows

Right now, I'm still in the crawl-before-you-walk phase.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API
 - UI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ env:
 on:
   pull_request:
 jobs:
-  api-test:
+  build-dev-container:
     runs-on: ubuntu-latest
     permissions:
       packages: write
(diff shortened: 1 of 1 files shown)


AlanCoding:
Looks like workflow reuse might be the only solution, and there are still questions with that.

Issue #8771, opened 2020-12-07
Job done isn't always correct


chrismeyersfsu:
After a job is complete we do post-things like send websocket notification as to the status of the job, process the fact cache, send notifications, and maybe more. There are 5 cases that are not completely correct.

Notifications are triggered when the events processes == events emitted from ansible for all job types other than Job.

1. Notifications for Job's are triggered via playbook_on_stats because we need data from the playbook_on_stats event. I think we only do this for convenience sake. We could trigger the notification after all events are processed and just get the `playbook_on_stats` event from the DB. This would reduce the deviation in call paths. Note that it's "kind of" safer to do it on `playbook_on_stats` because the number of events emitted does not have to equal number of events processed. This requirement _could_ lead to notifications never being sent if an event is lost between creation from ansible and saving to the db.
1. `EOF`. We detect when `events processes == events emitted` and create a special `EOF` event at the time we push events into Redis. When this `EOF` event is gotten in the callback receiver, we trigger notification for all jobs other than a Job Template Job. The problem here is that when we process `EOF` in the callback receiver, not all other events for the job may be processed. This is because we have multiple callback receiver threads. This basically makes the `EOF` event useless.
1. Job timeout. Tower has a feature of job timeout. If a job runs longer than the timeout the job will be forcefully killed. Combine that with sending notification on `playbook_on_stats` events and you _could_ run into a case where notifications are double sent.
1. Notification errors. We don't ensure that notifications are sent (no retry)
1. Job finishes 5+ seconds after all events have been emitted and saved to the database.

### Ideal Solution(s)

Detect `events processed == events emitted` in the callback receiver (see below with an example of how). If information is needed from `playbook_on_stats`, query the record from the database. Alternatively, cache the `playbook_on_stats` info in memory until `events processed == events emitted` and pass along with the notification.

We also want to account for sending events for jobs when `events processed` never equals `events emitted`. This is tricky. How do we know if the callback receiver is just slow to process events or that the event is lost?

As for the job timeout double send case. Any time notifications are to be sent, we should check to see if notifications have already been sent and short-circuit if so. This solution can be applied to the reaper and the job finishing 5+ seconds after all events are processed.

### Detect events processed == events emitted in callback receiver
[...]

PR #11384, opened 2021-11-23
Make sure to fire off failure notifications on error

jbradberry:
##### SUMMARY

where the error is unrelated to Ansible, thus is not caught by the
usual methods.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -1542,9 +1542,11 @@ class BaseTask(object):
             status = res.status
             rc = res.rc
 
-            if status == 'timeout':
-                self.instance.job_explanation = "Job terminated due to timeout"
-                status = 'failed'
+            if status in ('timeout', 'error'):
+                self.instance.job_explanation = f"Job terminated due to {status}"
+                if status == 'timeout':
+                    status = 'failed'
+
                 extra_update_fields['job_explanation'] = self.instance.job_explanation
                 # ensure failure notification sends even if playbook_on_stats event is not triggered
                 handle_success_and_failure_notifications.apply_async([self.instance.job.id])


jbradberry:
@chrismeyersfsu The one thing I'm worried about in this PR is if there are circumstances where an error result can cause the notification templates fire off by this path _and_ the more usual path.  Any thoughts?

chrismeyersfsu:
That's a good thing to be worried about @jbradberry . Do we call `handle_success_and_failure_notifications` anywhere else?

sarabrajsingh:
> That's a good thing to be worried about @jbradberry . Do we call `handle_success_and_failure_notifications` anywhere else?

@chrismeyersfsu the other calls to `handle_success_and_failure_notifications` are done in `callback.py` and `events.py` but aren't directly involved in firing off notifications when jobs fail. this is done at the `tasks.py` level

AlanCoding:
I didn't understand https://github.com/ansible/awx/issues/8771 before, but I might be seeing it now.

Either way, this seems fairly correct on its own. The speculated problem here seems to be the mutual exclusivity of the error status and the EOF message. While it could possibly happen, that doesn't sound highly credible to me.

AlanCoding (approved):

rebeccahhh (approved):

awx/main/tasks.py
```diff
@@ -1542,9 +1542,11 @@ def run(self, pk, **kwargs):
             status = res.status
             rc = res.rc
 
-            if status == 'timeout':
-                self.instance.job_explanation = "Job terminated due to timeout"
-                status = 'failed'
+            if status in ('timeout', 'error'):
+                self.instance.job_explanation = f"Job terminated due to {status}"
```
AlanCoding, line 1546: I just realized that we don't want `job_explanation` to read

> Job terminated due to error

This is particularly bad in the cases where an error occurred and `job_explanation` was set to something in response to that. This code would overwrite that useful message with this vague one.

PR #11386, opened 2021-11-23
Remove dev-only log filters and downgrade periodic logs

AlanCoding:
##### SUMMARY
I strongly prefer to share the logging settings between development and production to whatever extent possible.

With this, the global setting defaults to DEBUG. If we want INFO instead, we can talk about what is needed to make that happen. We do have the CTiT setting for the log level which applies to all logs.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
Without this, the dispatcher is effectively silenced. That's troublesome to me because you don't know what tasks are running. Yes, periodic tasks are annoying, but we should ask more general questions like - "will DEBUG or INFO include logs from periodic actions?"

### DIFF
--- a/awx/main/dispatch/pool.py
+++ b/awx/main/dispatch/pool.py
@@ -248,7 +248,7 @@ class WorkerPool(object):
         except Exception:
             logger.exception('could not fork')
         else:
-            logger.warn('scaling up worker pid:{}'.format(worker.pid))
+            logger.debug('scaling up worker pid:{}'.format(worker.pid))
         return idx, worker
 
     def debug(self, *args, **kwargs):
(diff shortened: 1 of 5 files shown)


john-westcott-iv (approved):

AlanCoding:
> With this, the global setting defaults to DEBUG

I don't know why I said that. I spun up things again and it's on INFO as expected. This is what I would prefer. The entirety of the developer team would need to know about that setting and how to change it to DEBUG, but this seems acceptable.

jbradberry (approved):

PR #11394, opened 2021-11-29
Load ansible.cfg from the branch specified on job template

notok:
Load ansible.cfg from the branch specified on job template (i.e. the same branch that the playbook exists), not from the branch set in the "project".

##### SUMMARY

Currently, `ansible.cfg` is loaded from the project directory on "Source Control Branch" set in the **"Project"**.
But when overriding project's branch by job template, the branch of ansible.cfg and
the branch of playbook to run does not match.
Users usually assume that the playbook will run with `ansible.cfg` on the same branch, 
so it should be loaded from "Source Control Branch" on the **"Job Template"**.

I'm trying to fix this in the [similar way as `RunInventoryUpdate` loads `ansible.cfg`](https://github.com/ansible/awx/blob/devel/awx/main/tasks.py#L2548).

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.5.1.dev10+gd6e4582c9c
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -1708,7 +1708,7 @@ class RunJob(BaseTask):
             ('ANSIBLE_ROLES_PATH', 'roles_path', 'requirements_roles', '~/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles'),
         )
 
-        config_values = read_ansible_config(job.project.get_project_path(), list(map(lambda x: x[1], path_vars)))
+        config_values = read_ansible_config(os.path.join(private_data_dir, 'project'), list(map(lambda x: x[1], path_vars)))
 
         for env_key, config_setting, folder, default in path_vars:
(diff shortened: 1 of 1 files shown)


AlanCoding (approved):
yeah, all looks valid.

AlanCoding:
@notok would you mind doing a rebase against the current `devel` branch to unblock some testing for me?

And while you're at it, would you mind going ahead with doing a `--signoff` on the commit?

https://github.com/ansible/awx/blob/eec4f8dcc27428cd1aeb1c251d43052a3ea780e8/DCO_1_1.md#L4

thanks!

notok:
@AlanCoding Thank you for your review!
I rebased against current `devel`, and added `--signoff` on commit.

PR #11410, opened 2021-12-01
Add setting for optionally disabling ansible-runner process isolation

enpaul:
##### SUMMARY

Currently the `process_isolation` parameter for Ansible Runner is [hardcoded to `True`](https://github.com/ansible/awx/blob/devel/awx/main/tasks.py#L975). This update makes this setting configurable via a Django setting called `RUNNER_PROCESS_ISOLATION` to allow a user to disable containerized process isolation if they want to (the default of `True` remains unchanged). This is intended to support development, testing, and small-scale environments where using containerized execution environments is either not possible or unnecessary.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.5.1.dev26+g3232b6b005
```

### DIFF
--- a/awx/main/tasks.py
+++ b/awx/main/tasks.py
@@ -969,6 +969,9 @@ class BaseTask(object):
         if settings.IS_K8S:
             return {}
 
+        if not settings.RUNNER_PROCESS_ISOLATION:
+            return {"process_isolation": False}
+
         image = instance.execution_environment.image
         params = {
             "container_image": image,
--- a/awx/main/utils/execution_environments.py
+++ b/awx/main/utils/execution_environments.py
@@ -21,6 +21,12 @@ def get_default_execution_environment():
     return ExecutionEnvironment.objects.filter(organization=None, managed=False).first()
 
 
+# this is the root of the private data dir as seen from inside
+# of the container running a job. if process isolation is disabled
+# then this should just be host project root directory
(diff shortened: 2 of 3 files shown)


awx/main/tasks.py
```diff
@@ -969,6 +969,9 @@ def build_execution_environment_params(self, instance, private_data_dir):
         if settings.IS_K8S:
             return {}
 
+        if not settings.RUNNER_PROCESS_ISOLATION:
+            return {"process_isolation": False}
```
AlanCoding, line 973: I can't imagine this would work right because project updates and other job types use `container_volume_mounts`... and I don't think those are compatible with setting this to False, but I'm not sure what the actual ansible-runner behavior would be in that case.

enpaul: Hm, I believe you're correct that these options are incompatible with each. I'll check that now.

However, I think this can still work if the `container_volume_mounts` option is removed as well when this is false, since the job will be running directly under the parent process and so will have access to the project directory. I'll do some more testing to make sure this is the case though.

enpaul: @AlanCoding I did confirm that this does work as I expect. The one thing to note is that the default value for `AWX_ISOLATION_BASE_PATH` doesn't work when this setting is set to `False` because container's don't share a `/tmp` location. However, by setting `AWX_ISOLATION_BASE_PATH` to a shared volume mounted in both containers this new setting worked exactly as intended.

PR #11414, opened 2021-12-02
Upgrade react

AlexSCorey:
##### SUMMARY
This upgrade the UI to React 17
##### COMPONENT NAME
 - UI

##### AWX VERSION


akus062381:
Triggering a Jenkins run against this branch

awx/ui/package.json
```diff
@@ -40,6 +40,7 @@
     "@lingui/loader": "^3.8.3",
     "@lingui/macro": "^3.7.1",
     "@nteract/mockument": "^1.0.4",
+    "@wojtekmaj/enzyme-adapter-react-17": "0.6.5",
```
keithjgrant, line 43: the fact that enzyme seems to be falling behind is a little concerning. I wonder if we're going to have to evaluate the switch to react testing library at some point 😖

nixocio (approved):

awx/ui/package.json
nixocio: @keithjgrant enzyme adapter still does not have support for react 17. It seems like react testing is being kept in sync with react versions. Something to be evaluated IMO before we upgrade to react 18.
https://testing-library.com/docs/react-testing-library/migrate-from-enzyme/

akus062381 (approved):

PR #11415, opened 2021-12-03
Fix: honor update_secrets in awx.awx.user

nbr23:
##### SUMMARY

Fix awx.awx.user in awx_collection to honor the update_secrets parameter when set

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - collection

##### AWX VERSION

```
```

##### ADDITIONAL INFORMATION

In the current awx.awx.user module, even if `update_secrets` is false, the secrets will get updated.
This fixes this issue, making the parameter effective.

### DIFF
--- a/awx_collection/plugins/modules/user.py
+++ b/awx_collection/plugins/modules/user.py
@@ -163,6 +163,10 @@ def main():
     if password is not None:
         new_fields['password'] = password
 
+    # Prevent password reset when the user exists and update_secrets is False
+    if existing_item and existing_item.get('password') is None and not module.update_secrets:
+      del new_fields['password']
+
     # If the state was present and we can let the module build or update the existing item, this will return on its own
     module.create_or_update_if_needed(existing_item, new_fields, endpoint='users', item_type='user')
 


awx_collection/plugins/modules/user.py
```diff
@@ -163,6 +163,10 @@ def main():
     if password is not None:
         new_fields['password'] = password
 
+    # Prevent password reset when the user exists and update_secrets is False
+    if existing_item and existing_item.get('password') is None and not module.update_secrets:
+      del new_fields['password']
```
AlanCoding, line 168: looks like indentation is off from prior lines in this file. I'd like to have that consistent.

AlanCoding:
Looks good to me aside from my whitespace comment

Could you do the `--signoff` while you're at it?

https://github.com/ansible/awx/blob/eec4f8dcc27428cd1aeb1c251d43052a3ea780e8/DCO_1_1.md#L4

john-westcott-iv:
Can you provide some replication steps for the problem? The update_secrets should be handled by the controller_api library automatically. In `update_if_needed` (called at the end of user.py) we call `self.objects_could_be_different` which should be checking the value of all of the fields in the existing object vs the new fields and then considering the `update_secrets` parameter to determine if we need to update the object. If the user module is not respecting this properly perhaps this is a bug in `has_encrypted_values` or somewhere else in that code?

AlanCoding:
^ yes, @john-westcott-iv was being sharper than I was. If you give a task and what's going wrong with it, then we might be able to better speculate

nbr23:
>In `update_if_needed` (called at the end of user.py) we call `self.objects_could_be_different` which should be checking the value of all of the fields in the existing object vs the new fields and then considering the `update_secrets` parameter to determine if we need to update the object.

Yes, and this behavior is working fine. The problem is that the AWX instance does not return a `password` value at all (which I believe make sense from a security standpoint, a GET on a user should not allow the requester to get a password, or even a hash). So `existing_item.get('password')` is None, and shows different when compared with the new password value.

The high level call trace is the following:
```
awx.awx.user:main()
->create_or_update_if_needed
	->update_if_needed
		->objects_could_be_different   -> True
			-> fields_could_be_same(old_field=None, new_field="Newpassword") -> False
```

`fields_could_be_same` needs one of the values to be set to  `$encrypted$` in order to report a difference.

Let me know if you see a more adequate fix for this (perhaps setting the existing item's password to `$encrypted$` instead of deleting the field entirely?), or if I am not making sense.

Cheers!

AlanCoding:
Oh, interesting, valid point.

https://github.com/ansible/awx/blob/cdf7fd64b2d8564f0ede2251f50f8fc05aee5cea/awx_collection/plugins/module_utils/controller_api.py#L864

How risky would it be to do `old_field = old.get(field, "$encrypted$")` here? Open question.

john-westcott-iv:
I was thinking of something like this in `objects_could_be_different` (untested):
```
            if 'username' in new and field is 'password':
                old_field = self.ENCRYPTED_STRING
```

But I am also wondering if the API itself should be changed to return a more standard field for a user entry like this:
```
"password": "$encrypted$",
```
That would match other endpoints and maybe auto-magically fix this issue.

nbr23:
@john-westcott-iv fixing it API side would definitely be ideal... But, we'll still have the issue with older (<=current) AWX versions.

For those versions, my logic in fixing it in the user module directly is that this is not the cleanest fix, so I thought it might be better to contain it to this module. Something for the password field `controller_api` might have side effects I wouldn't know of.

john-westcott-iv:
So this is an interesting point with the awx.awx collection. Its versioned corresponding with the version of the awx that it was built for. There is no guarantee that using using a mismatch of awx and the collection will work. We also don't typically "back port" fixes to the collection either. In theory it is possible but typically we just put fixes in and move forward. Therefore we couldn't normally consider dealing with older(<=current) versions.

Fixing this the way you did in user.py actually changes the way the parameter is intended to work (see the help docs on line 60). So if we wanted to fix the issue in user.py there we need a much more elaborate solution that would make that parameter work as advertised. For example, we could do one update for "everything else" and then do an update for the password if the last update made a change and the parameter was set. However, a "two step write" like this might come with its own drawbacks (i.e. what happens if the first write succeeds and the password update fails, at that point do you roll back or just error out with the change in place?). 

As far as changing controller_api.py goes, there is a fairly large testing environment around the collection. There are integration tests in  awx_collection/tests/integration/targets and unit tests in awx_collection/test/awx so I would be less worried about having adverse affects as those tests should bring any issues forward. In fact, adding a test or two to validate this functionality would be good so that we would know if we ever regressed around this issue.

nbr23:
@john-westcott-iv good point about not matching the module doc.

I checked a bit further on the AWX API docs, there's no mention of a `password` value ever being returned. And I cannot think of a use case other than ours. I am not sure our case justifies the addition (unless I am mistaken, the returned value for 'password' would *always* be `"$encrypted$"`, making it not so interesting), but if you think it does, then I will close this PR and open an issue so someone more familiar with AWX itself can look into it.

If you think it's worth fixing in controller_api, I'm happy to try something around what you drafted and add some test scenarios. The doc will also need to be updated in this case, to reflect the API behavior and how we handle it.

Thanks!

john-westcott-iv:
@AlanCoding Any thoughts on the api returning `"password": "$encrypted$",` for the user object?

AlanCoding:
I don't think I have any objections.

https://github.com/ansible/awx/blob/481f6435eeaa90c05bc605f34fe42bcafbc653a2/awx/api/serializers.py#L939

It was intentional how it was done... but I don't have any reason to defend this. It was introduced in 9916a5eefbbd56239cc16e91852aaa592574326c, which is truly ancient and there wasn't any other precedent of encrypted fields at that time.

john-westcott-iv:
API fix: https://github.com/ansible/awx/pull/13704

PR #11417, opened 2021-12-03
Improve module collection AWX (job_launch and workflow_launch)

paytroff:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - job_launch
 - workflow_launch

### DIFF
--- a/awx_collection/plugins/modules/job_launch.py
+++ b/awx_collection/plugins/modules/job_launch.py
@@ -102,6 +102,11 @@ options:
         - If waiting for the job to complete this will abort after this
           amount of seconds
       type: int
+    check_prompted:
+      description:
+        - Check if field prompted are send by module
+      type: bool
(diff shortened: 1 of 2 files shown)


TheRealHaoLiu:
@paytroff Hi there, thank you for your contribution to AWX and welcome to the community!

From the content of the PR we understand that you are trying to add a check on the job template and workflow job template modules to confirm that the information being passed into the module is prompted for in the template on AWX.

The philosophy we used when writing the modules was to pass gathered information back to the API and let the API raise any errors on its own. That way, if the API behavior ever changes we don't need to change any checks like this in the modules.

For these reasons, the AWX team decided to not accept this PR.

If you like to discuss further come join us on our Matrix chatroom https://matrix.to/#/#awx:ansible.com or join us live on our Community Office Hour!

PR #11427, opened 2021-12-07
Merge Tower release_4.1 into awx/devel

AlexSCorey:
##### SUMMARY
This merges Tower_4.1 into awx/devel


chrismeyersfsu (approved):
Looks good. Just need to resolve that Spanish translation fix and then merge away.

.github/workflows/upload_schema.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - devel
+      - release_4.1
```
kdelee: I'm guessing we don't want that on awx
```suggestion
```

awx/main/tasks.py
```diff
@@ -3251,10 +3256,21 @@ def work_type(self):
...
     @cleanup_new_process
     def cancel_watcher(self, processor_future):
+        receptor_ctl = get_receptor_ctl()
         while True:
             if processor_future.done():
                 return processor_future.result()
 
+            # cancel job if receptor no longer knows about work item
+            try:
+                receptor_ctl.simple_command(f'work status {self.unit_id}')
```
AlanCoding, line 3266: This was reverted. The diff creates a new connection to receptor in a thread, which is dangerous and it definitely broke stuff.

I think the issue here is that previous downstream merges into `devel` picked up both this commit and its revision commit. Then that branch was either modified or rebased.

If you're really set of getting the branches fully synced, then you could just add another reversion commit to the tip of the branch

PR #11428, opened 2021-12-07
Make awx-python script available in k8s app images

rooftopcellist:
##### SUMMARY
Currently, the `awx-python` script is not making it into the AWX k8s container, which means that awx-manage cannot run. 

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API


rooftopcellist:
@kdelee and I just built an awx image from this branch and deployed it with the awx-operator and can now run the awx-manage command, which allows us to run migrations.

rooftopcellist:
We should consider adding a health check to the awx-operator.  This could hit an API endpoint and update a status on the AWX object.  Then we could add a verification step to the AWX PR check which would check that this status is true: https://github.com/ansible/awx/blob/devel/.github/workflows/ci.yml#L213

kdelee (approved):
Confirmed this makes things actually work :1st_place_medal:

PR #11430, opened 2021-12-07
SAML user flags 5303

### LINKED ISSUE #5303: RFE: Flagging users as superusers and auditors in SAML integration
##### ISSUE TYPE
Feature request for flagging users as superusers and auditors in SAML integration, similar to LDAP implementation.

john-westcott-iv:
##### SUMMARY

Modifies SAML auth providers to allow for setting of is_superuser is_systemauditor flags as well as enabling org mapping to now include adding users as org admins. This is addresses related #5303. 

##### ISSUE TYPE

 - Feature Pull Request
 
##### COMPONENT NAME

 - API
 - UI

##### AWX VERSION

```
awx: 1.0.7.dev13754+g41569d5532.d20211207
```

##### ADDITIONAL INFORMATION

```

```


tools/docker-compose/ansible/plumb_keycloak.yml
```diff
@@ -0,0 +1,83 @@
...
+        # We add the extra \\ in here so that when jinja is templating out the files we end up with \n in the strings.
+        public_key_content: "{{ lookup('file', public_key_file) | regex_replace('\n', '\\\\n') }}"
+        private_key_content: "{{ lookup('file', private_key_file) | regex_replace('\n', '\\\\n') }}"
+
+    - debug:
+        msg:
+          - "{{ existing_saml }}"
+
+    - pause:
+        prompt: "Continuing to run this will replace your existing saml settings (displayed above). They will all be captured except for your private key. Be sure that is backuped before continuing"
```
rebeccahhh: backed up

tools/docker-compose/README.md
```diff
@@ -190,6 +190,76 @@ awx_1        |   Applying auth.0001_initial... OK
 ...
 ```
 
+##### Keycloak Integration
+Keycloak is a SAML provider and can be used to test AWX social auth. This section describes how to build a reference Keycloak instance and plumb it with AWX for testing purposes.
+
+First, be sure that you have the awx.awx collection installed by running `make install_collection`.
+Next, make sure you have your containers running by running `make docker-compose`.
+
+Note: The following instructions assume we are using the built in postgres database container. If you are not using the internal database you can use this guide as a reference, updating the database fields as required for your connection.
```
AlanCoding: might want to hyphen "built-in" for clarity

tools/docker-compose/README.md
```diff
@@ -190,6 +190,76 @@ awx_1        |   Applying auth.0001_initial... OK
 ...
 ```
 
+##### Keycloak Integration
```
AlanCoding: I'd prefer to put this under the "Extras" section, similar to the "Start with Minikube" section.

tools/docker-compose/README.md
```diff
@@ -190,6 +190,76 @@ awx_1        |   Applying auth.0001_initial... OK
...
+
+The first one time command will be creating a Keycloak database in your postgres database by running:
+```bash
+docker exec tools_postgres_1 /usr/bin/psql -U awx --command "create database keycloak with encoding 'UTF8';"
+```
+
+The second one time command will be to start a Keycloak container to build our admin user; be sure to set pg_username and pg_password to work for you installation. Note: the command below set the username as admin with a password of admin, you can change this if you want.
+```bash
+docker run -e KEYCLOAK_USER=admin -e KEYCLOAK_PASSWORD=admin --net=_sources_default \
+           -e DB_VENDOR=postgres -e DB_ADDR=postgres -e DB_DATABASE=keycloak -e DB_USER=<pg_username> -e DB_PASSWORD=<pg_password> \
```
AlanCoding: don't they need instructions to get the pg_password? You can get it from `awx-manage shell_plus`. And don't we know the intended pg_username?

tools/docker-compose/README.md
```diff
@@ -190,6 +190,76 @@ awx_1        |   Applying auth.0001_initial... OK
...
+Once you see a message like: `WFLYSRV0051: Admin console listening on http://127.0.0.1:9990` you can stop the container.
+
+Now that we have performed the one time setup anytime you want to run a Keycloak instance alongside AWX we can start docker-compose with the KEYCLOAK option to get a Keycloak instance with the command:
+```bash
+KEYCLOAK=true make docker-compose
+```
+
+Go ahead and stop your existing docker-compose run and restart with Keycloak before proceeding to the next steps.
+
+Once the containers come up a new port (8443) should be exposed and the Keycloak interface should be running on that port. Connect to this interface and select the "Administration console" link. Log into the UI the username/password set in the previous `docker run` command to validate Keycloak is working.
```
AlanCoding: Can you spell this out a little? Is this https? So can you give a URL to copy + paste? Presumably https://localhost:8443?

tools/docker-compose/README.md
```diff
@@ -190,6 +190,76 @@ awx_1        |   Applying auth.0001_initial... OK
...
+* Create a realm in Keycloak with a client for AWX and 3 users.
+* Backup and configure the SMAL adapter in AWX. NOTE: the private key of any existing SAML adapters can not be backed up through the API, you need a DB backup to recover this.
+
+Before we can run the playbook we need to understand that SAML works by sending redirects between AWX and Keycloak through the browser. Because of this we have to tell both AWX and Keycloak how they will construct the redirect URLs. On the Keycloak side, this is done within the realm configuration and on the AWX side its done through the SAML settings. The playbook requires a variable called `container_reference` to be set. The container reference variable needs to be how your browser will be able to talk to the running containers.  Here are some examples of how to choose a proper container reference.
+* If you develop on a mac which runs a Fedora VM which has AWX running within that and the browser you use to access AWX runs on the mac. The the VM with the container has its own IP that is mapped to a name like `tower.home.net`. In this scenario your "container reference" could be either the IP of the VM or the tower.home.net friendly name.
```
AlanCoding: nit pick, but I'd prefer to write `"container reference"` as the `container_reference` variable because people will ctl+F to follow along

tools/docker-compose/README.md
```diff
@@ -190,6 +190,76 @@ awx_1        |   Applying auth.0001_initial... OK
...
+```bash
+ansible-playbook tools/docker-compose/ansible/plumb_keycloak.yml -e container_reference=<your container reference here>
+```
+
+Once the playbook is done running SAML should now be setup in your development environment. This realm has three users with the following username/passwords:
+1. awx_unpriv:unpriv123
+2. awx_admin:admin123
+3. awx_auditor:audit123
+
+The first account is a normal user. The second account has the attribute is_superuser set in Keycloak so will be a super user in AWX. The third account has the is_system_auditor attribute in Keycloak so it will be a system auditor in AWX. To log in with one of these Keycloak users go to the AWX login screen and click the small "Sign In With SAML Keycloak" button at the bottom of the login box.
```
AlanCoding: This is pretty awesome to give this level of reproducibility of the feature you developed.

AlanCoding:
Since `keycloak.awx.realm.json.j2` is such a massive file, I'd like it if you could give some instructions for re-generation of that file, in the case that something breaks in the future. Very loose instructions would be fine, giving someone a starting point to fix unforeseen future problems would be really good to have.

awx/sso/pipeline.py
```diff
@@ -229,3 +229,70 @@ def update_user_teams_by_saml_attr(backend, details, user=None, *args, **kwargs)
...
+    role_setting = user_flags_settings.get(is_role_key, None)
+    if role_setting:
+        # We do a 2 layer check here so that we don't spit out the else message if there is no role defined
+        if role_setting in attributes.get('Role', []):
+            logger.debug("User %s has %s role %s" % (user.username, flag, role_setting))
+            new_flag = True
+        else:
+            logger.debug("User %s is missing the %s role %s" % (user.username, flag, role_setting))
+
+    # Next, check to see if we are respecting an attribute; this will take prio over the role if its defined
```
AlanCoding: prio --> priority

awx/sso/pipeline.py
```diff
@@ -229,3 +229,70 @@ def update_user_teams_by_saml_attr(backend, details, user=None, *args, **kwargs)
...
+    attr_setting = user_flags_settings.get(is_attr_key, None)
+    if attr_setting and attributes.get(attr_setting, None):
+        # Do we have a required value for the attribute
+        if user_flags_settings.get(is_value_key, None):
+            # If so, check and see if the value of the attr matches the required value
+            attribute_value = attributes.get(attr_setting, None)
+            if isinstance(attribute_value, (list, tuple)):
+                attribute_value = attribute_value[0]
```
AlanCoding, line 258: Why is this sometimes a list, and why is the first element desirable?

awx/sso/pipeline.py
```diff
@@ -229,3 +229,70 @@ def update_user_teams_by_saml_attr(backend, details, user=None, *args, **kwargs)
...
+
+
+def update_user_flags(backend, details, user=None, *args, **kwargs):
+    if not user:
+        return
+
+    from django.conf import settings
+
+    user_flags_settings = settings.SOCIAL_AUTH_SAML_USER_FLAGS_BY_ATTR
+    logger.debug(user_flags_settings)
```
AlanCoding: probably a development log here to delete

awx/sso/tests/functional/test_pipeline.py
```diff
@@ -357,3 +357,73 @@ def test_galaxy_credential_auto_assign(self, users, kwargs, galaxy_credential, m
...
+            'email': ['noone@nowhere.com'],
+            'last_name': ['Westcott'],
+            'is_superuser': ['true'],
+            'username': ['test_id'],
+            'first_name': ['John'],
+            'Role': ['test-role-1'],
+            'name_id': 'test_id',
+        }
+
+        assert expected == _check_flag(user, 'superuser', attributes, user_flags_settings)
```
AlanCoding, line 429: is it worth testing the auditor flag too?

Nice test btw

john-westcott-iv: The _check_flag should take either superuser or system_auditor and run the same logic on them so I figured testing one is indirectly testing the other.

awx/sso/pipeline.py
john-westcott-iv: I had intentionally left this in as a way to validate that SAML is actually sending you the attributes you expect. It will log information about the user if DEBUG is on but I believe LDAP may be doing the same.

awx/sso/pipeline.py
john-westcott-iv: This should be a function of how the IdP is sending the data to the SP. It's possible for an IdP to send multiple values for attributes or a single attribute. The SAML adapter already used the prescience of just picking the first item if it was a list so I followed suit.

tools/docker-compose/README.md
john-westcott-iv: Thanks for the suggestion, this command was kind of a pain and assuming awx user and grabbing the password will make it easier. I modified the note to include a comment that if you aren't using the defaults you need to change them in this command.

tools/docker-compose/README.md
john-westcott-iv: Thanks!

awx/sso/pipeline.py
AlanCoding: If that's intentional, then it might be better to add some context around it with an f-string or something. Someone may not be able to tell what it is just looking at the data. Like `f"Current value of SOCIAL_AUTH_SAML_USER_FLAGS_BY_ATTR is:\n{user_flags_settings}"`. Otherwise someone may not draw the line from the data to its meaning.

john-westcott-iv: After looking at this close, this one I think can be removed, I was thinking this was the lower debug that says what is coming back from SAML but this was just settings. An admin should know what was set in the SAML adapter screen.

tools/docker-compose/README.md
```diff
@@ -311,3 +314,75 @@ If you want to clean all things once your are done, you can do:
...
+First, be sure that you have the awx.awx collection installed by running `make install_collection`.
+Next, make sure you have your containers running by running `make docker-compose`.
+
+Note: The following instructions assume we are using the built-in postgres database container. If you are not using the internal database you can use this guide as a reference, updating the database fields as required for your connection.
+
+We are now ready to run two one time commands to build and pre-populate the Keycloak database.
+
+The first one time command will be creating a Keycloak database in your postgres database by running:
+```bash
+docker exec tools_postgres_1 /usr/bin/psql -U awx --command "create database keycloak with encoding 'UTF8';"
```
AlanCoding, line 330: It might be helpful to share what the expected output looks like. For me it just says "CREATE DATABASE"

tools/docker-compose/README.md
```diff
@@ -311,3 +314,75 @@ If you want to clean all things once your are done, you can do:
...
+```
+
+The second one time command will be to start a Keycloak container to build our admin user; be sure to set pg_username and pg_password to work for you installation. Note: the command below set the username as admin with a password of admin, you can change this if you want. Also, if you are using your own container or have changed the pg_username please update the command accordingly.
+```bash
+PG_PASSWORD=`cat tools/docker-compose/_sources/secrets/pg_password.yml  | cut -f 2 -d \'`
+docker run -e KEYCLOAK_USER=admin -e KEYCLOAK_PASSWORD=admin --net=_sources_default \
```
AlanCoding, line 341: After you ctl+c it, this will leave the stopped container hanging around. I _think_ adding `--rm` may take care of that. I tend to use a formula like `docker run --rm --tty --interactive` to start with, but some of that may or may not be desirable in this case.

tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
```diff
@@ -79,6 +79,22 @@ services:
   {% set container_postfix = loop.index %}
     - "awx_{{ container_postfix }}"
   {% endfor %}
+{% endif %}
+{% if enable_keycloak|bool %}
+  keycloak:
+    image: quay.io/keycloak/keycloak:15.0.2
```
AlanCoding, line 85: I'd like to add `container_name`, because I don't like the default name it gets of `sources_keycloak_1`.

I think maybe `tools_keycloak` would be best, but ultimately up to you.

awx/settings/defaults.py
```diff
@@ -535,6 +536,7 @@ def IS_TESTING(argv=None):
 
 SOCIAL_AUTH_SAML_ORGANIZATION_ATTR = {}
 SOCIAL_AUTH_SAML_TEAM_ATTR = {}
+SOCIAL_AUTH_SAML_USER_FLAGS_BY_ATTR = {}
```
AlanCoding, line 539: We don't have any other `USER_FLAGS_BY_ATTR` settings, is this the only external auth that supports this so-far?

john-westcott-iv: LDAP has a similar concept with the field `LDAP User Flags by Group` (`AUTH_LDAP_1_USER_FLAGS_BY_GROUP`). Initially I was just doing attributes but I also added groups. We could change the name of the property to be more reflective of that. Maybe just shorten it to`SOCIAL_AUTH_SAML_USER_FLAGS`?

awx/sso/tests/unit/test_fields.py
```diff
@@ -115,6 +115,48 @@ def test_internal_value_invalid(self, data, expected):
...
+                    'junk': 'something',
+                    'is_superuser_value': 'value',
+                    'is_superuser_role': 'my_peeps',
+                    'is_system_auditor_attr': 'else',
+                    'is_system_auditor_value': 'value2',
+                    'is_system_auditor_role': 'other_peeps',
+                },
+                {'junk': ['Invalid field.']},
+            ),
+        ],
```
AlanCoding, line 168: Maybe you meant to add another example later, but `parametrize` with only 1 entry isn't a good general practice.

Are these keys like `is_superuser_value` required? If you have 2 junk keys, should they both show as invalid? It wouldn't be hard to brainstorm a 2nd test case.

tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
john-westcott-iv: I'd like to keep the `_1` to match the naming convention of `tools_postgres_1` and `tools_redis_1` but I like changing it from `sources_` to `tools_`.

awx/ui/src/screens/Setting/shared/data.allSettingOptions.json
```diff
@@ -3706,6 +3706,28 @@
...
+          "is_system_auditor_attr": "saml_attr",
+          "is_system_auditor_value": "value",
+          "is_system_auditor_role": "saml_role"
+        },
+        "default": {},
+        "child": {
+          "type": "field",
+          "required": true,
+          "read_only": false
+        }
```
AlanCoding, line 3730: I'm adding @nixocio as a reviewer to this PR because you're touching some of the configure-tower-in-tower UI code here.

Do you even need the "placeholder" if it can come from the API options?

tools/docker-compose/README.md
john-westcott-iv: Let me know if I am thinking about this wrong but this should be ok because the next command should pretty much restart the container when you run the `KEYCLOAK=true make docker-compose`

tools/docker-compose/ansible/plumb_keycloak.yml
```diff
@@ -0,0 +1,83 @@
...
+        existing_saml: "{{ lookup('awx.awx.controller_api', 'settings/saml', host=awx_host, password=awx_pass, username=awx_user, verify_ssl=false) }}"
+        new_saml: "{{ lookup('template', 'saml_settings.json.j2') }}"
+      vars:
+        # We add the extra \\ in here so that when jinja is templating out the files we end up with \n in the strings.
+        public_key_content: "{{ lookup('file', public_key_file) | regex_replace('\n', '\\\\n') }}"
+        private_key_content: "{{ lookup('file', private_key_file) | regex_replace('\n', '\\\\n') }}"
+
+    - debug:
+        msg:
+          - "{{ existing_saml }}"
```
AlanCoding, line 36: eh, it would be nice to pad this with some explanation.

tools/docker-compose/README.md
```diff
@@ -311,3 +314,75 @@ If you want to clean all things once your are done, you can do:
...
+    awx_user: admin
+    awx_pass: admin
+    cert_subject:  "/C=US/ST=NC/L=Durham/O=awx/CN="
+```
+
+* keycloak_(user|pass) need to change if you modified the user when starting the initial container above.
+* awx_(user|pass) are credentials to get you into your AWX instance so you can read and write the SAML settings.
+* cert_subject will be the subject line of the certificate shared between AWX and keycloak you can change this if you like or just use the defaults.
+
+To override any of the variables above you can add more `-e` arguments to the playbook run below. For example, if you simply need to change the `awx_pass` add the argument `-r awx_pass=my_secret_pass` to the next command.
```
AlanCoding: You also need to authenticate for the collection with `~/.tower-cli.cfg` or the env vars. Given the flow of these instructions, I say you should put them host/username/password for `awx.awx` collection as env vars before `ansible-playbook` in the snippet. That will make this _real_ smooth for someone to follow through.

AlanCoding (approved):
Screenshot from 2021-12-17 15-05-25

I followed the instructions, logged in as the auditor user, and then logged in as the admin user. Confirmed that I have admin privileges as the admin user, and other users show up as they log in.

This is outstanding work, and I have a high level of comfort with this diff myself. After today, I'll be out until January. I've left a few minor improvement suggestions, all of these are non-blocking as the core feature is :+1:

tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
AlanCoding: we have a `tools_receptor_hop` container, that's where I was coming from here. But the `_1` is fine too.

EDIT: I think now I agree this should be named similarly to postgres and redis.

awx/ui/src/screens/Setting/shared/data.allSettingOptions.json
nixocio: This data is just used on tests. You are right values come from the API options.

nixocio (approved):
The UI changes are per standards already present for other fields.

image


I also triggered the UI e2e tests.

AdamVB:
Maybe we're doing something wrong,

but it seems like the the sso pipeline is only evaluating the first value in the SAML attribute.


Our SAML response looks like this:

```
<saml:Attribute Name="groups" NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:basic">
  <saml:AttributeValue xsi:type="xs:string">qaadmins</saml:AttributeValue>
  <saml:AttributeValue xsi:type="xs:string">jira-software-users</saml:AttributeValue>
  <saml:AttributeValue xsi:type="xs:string">awx-frontend-admins</saml:AttributeValue>
......
```

and we get an error message like this:

```
awx.sso.pipeline Refusing superuser for AdamVB because attr groups (qaadmins) did not match value 'awx-frontend-admins' 
```

the awx attribute mapping looks like this:

```
{
  "is_superuser_attr": "groups",
  "is_superuser_role": "is_superuser",
  "is_superuser_value": "awx-frontend-admins",
  "is_system_auditor_attr": "groups",
  "is_system_auditor_role": "is_system_auditor",
  "is_system_auditor_value": "awx-frontend-auditors"
}
```

the login with ldap grants the superuser role, but logging in with SAML (which has the same LDAP as backend) results in the above error.

To me it looks like not all values in the attribute "group" are checked/iterated over, can somebody confirm this?

john-westcott-iv:
@AdamVB that is correct, it picks the first one if the attribute is a list. Can you map your group into a role? Or can your IdP send a single attribute to AWX based on that group membership?

AdamVB:
@john-westcott-iv we run a very simple IdP that pretty much mirrors our ldap structure.

we could modify it to send a single attribute for AWX, but then AWX would be the only application fow which we'd need to take such a step, not really an elegant solution.


You mentioned in your PR:

> The SAML adapter already used the prescience of just picking the first item if it was a list so I followed suit.

could you think of any issues if we changed that to iterate over the list?


that would make it work similarly to the ldap mapping, and make it more admin friendly, as SAML Attributes can always be lists, and as of now it's not documented anywhere that it only takes the first list entry.


I will test this and open up a PR  when its done, should be a quick fix

PR #11431, opened 2021-12-07
Modify Instance and introduce InstanceLink

jbradberry:
##### SUMMARY
Allow us to capture and serve back up the mesh topology.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

TODO:
- [x] ~~undo exclusion of hop nodes from API endpoints?~~
  - ~~how much UI work is going to be needed to deal with this?~~
- [x] directionality
  - ~~make links symmetric, or~~
  - ~~make all entry points consistent about direction choice, e.g. sort by hostname, or~~
  - ~~make sure `receptorctl` grows a means by which the direction of links can be inspected, or~~
  - rip out the auto-discovery feature (**EDIT: we have decided to go this route**)
- [x] make the dev environment properly register the instances and links
- [x] add the ability to de-register links


awx/main/models/receptor.py
```diff
@@ -0,0 +1,28 @@
...
+
+from awx.main.models.base import BaseModel
+from awx.main.models.ha import Instance
+
+
+class ReceptorNode(BaseModel):
+    NODE_TYPE_CHOICES = [
+        ('control', "Control plane node"),
+        ('hybrid', "Controller and execution"),
+        ('execution', "Execution plane node"),
```
AlanCoding: The hop node doesn't say "execution plane node". To avoid definition problems I'm prefer to avoid this language and just refer to the fact that it's execution-only functionality wise.

awx/main/models/receptor.py
```diff
@@ -0,0 +1,28 @@
...
+        ('hop', "Hop node"),
+    ]
+    NODE_STATE_CHOICES = [
+        ('healthy', "Healthy"),
+        ('unhealthy', "Unhealthy"),
+        ('disabled', "Disabled"),
```
AlanCoding: The disabled state would be duplicated with the instance model. I can maybe see through to the intention with the unhealthy state being a receptor construct, but this is also duplicated with the `last_seen` attribute on the instance model. While it's true that doesn't record the `last_seen` for hop nodes, hop nodes can't be disabled... so this starts to look like a little bit of a data model mess.

AlanCoding:
How will you deal with nodes that are removed from the receptor mesh intentionally?

More basically, where will you be getting this information from? I see 2 options - the installer sets it all mostly transferring its inventory contents to this table, or by the output of `receptorctl status`. I would want that to be abundantly clear before we commit a schema change.

If we're setting this by the inventory file contents, then we will still need to deal with the problem of removing nodes. Would this be done by the de-provisioning hostvar, or would it be done by comparing the database list to the inventory list of nodes? If it's the former, then that still leaves gaps I don't like, because I don't want to assume that users do things correctly.

jbradberry:
@AlanCoding As I see it, there were only 2 paths forward

1. Overload the existing `Instance` model.  The downside to this is that semantically these objects are intended to represent units of execution capacity, which would no longer be the case if we start mapping them one-to-one to receptor nodes, particularly hop nodes.  How much machinery would have to be adjusted to take this change into consideration?
2. Create new models that directly represent the mesh.  The downside to this is, as you quickly noticed, that there is a lot of overlap with the `Instance` model and we would have to keep them in sync.

Our group ultimately preferred option 2, since it is lower risk.  The intention is that we are going to ingest the information from the installer in much the same way that we currently do for `Instance`s and `InstanceGroup`s, by creating a couple of new management commands that the installer can then invoke.  Things that dynamically happen to `Instance`s within the app itself can probably be reasonably dealt with via Django signals.

awx/main/models/receptor.py
jbradberry: These choices (except the hop node type) were copied directly from the choices for the `Instance` model.  If it is a problem here it's a problem there too.

awx/main/models/receptor.py
jbradberry: There is definitely something to what you say, but this is information that @kialam wants to include in the visualizer.  Do we want the API list view for `ReceptorNode`s to prefetch the `Instance`s, or is it ultimately acceptable for this to be denormalized in the database the way I have it here.

AlanCoding:
I get the direction you were going with the (2) option, and I'm fine with that assuming that we can have strict rules about what the data means. If this is a table meant to reflect the installer's inventory file, then that's defensible and that's what I heard pitched.

Allow me to take point (2) as gospel and then outline review comments here:
 - remove the `ReceptorNode.instance` field and the `ReceptorNode.node_state` field
 - delete the `ReceptorLink` model, track peers as a JSON field which is a list, because this is a more literal reflection of the inventory file data structure, _and_ because this is what the `policy_instance_list` already does

I want to avoid a little bit of option (1) and a little bit of option (2). If we're doing option (2), then I don't think Django signals should come into the picture. Edits to the `Inventory` model can happen in any arbitrary context, which is the opposite of what we want, we want this to only be defined in setup.

I've wanted to make the API have a means of vendoring its own "cluster inventory". Here, this is the dynamic inventory for the `make docker-compose` development environment:

```python
#!/usr/bin/env python3

import subprocess
import json
import sys

p = subprocess.run("docker ps --format '{{.Names}}'", stdout=subprocess.PIPE, shell=True)
container_names = p.stdout.decode('utf-8').strip()

awx_nodes = []
execution_nodes = []
hop_nodes = []

out = {
    'local': ['127.0.0.1'],
    'tower': [],
    'automationcontroller': [],
    'execution_nodes': [],
    "_meta": {
        "hostvars": {
            "127.0.0.1": {
                'ansible_connection': 'local',
                'ansible_python_interpreter': sys.executable,  # mainly needed for collection tests
            }
        }
    }
}

def add_node(hostname, instance_groups, node_type):
    for ig in instance_groups:
        out[ig].append(hostname)

    # hybrid, execution nodes use quay.io/awx/awx_devel
    # while hop nodes use quay.io/project-receptor/receptor
    out['_meta']['hostvars'][hostname] = {'ansible_connection': 'community.docker.docker',
                                          'ansible_user': 'root' if node_type == 'hop' else '0'}
    if node_type != 'hybrid':
        out['_meta']['hostvars'][hostname]['node_type'] = node_type


for container_name in container_names.split('\n'):
    if container_name.startswith('tools_awx_'):
        add_node(container_name, ['tower', 'automationcontroller'], 'hybrid')
    elif container_name.startswith('tools_receptor_'):
        node_type = 'hop' if 'hop' in container_name else 'execution'
        add_node(container_name, ['execution_nodes'], node_type)

for ig in ('tower', 'automationcontroller', 'execution_nodes'):
    out[ig] = sorted(out[ig])

print(json.dumps(out))
```

I'd love to have a universal thing for "here's a URL, give me its cluster inventory" that works for all install methods, including OpenShift. Nothing we have now is either universal or sufficiently correct.

----

But on the subject of (1)

> The downside to this is that semantically these objects are intended to represent units of execution capacity, which would no longer be the case if we start mapping them one-to-one to receptor nodes, particularly hop nodes. How much machinery would have to be adjusted to take this change into consideration?

Possibly no adjustments for _functionality_. See references to `capacity_type` in the task manager module. This tracks whether something is a control plane thing or an execution thing. Whenever we have to select a node to do something, you can count on us filtering on `node_type` in some way.

There would still be additional work, we would have to add  lot of validation cases. You shouldn't be able to add a hop node to an instance group... you really can't even do a health check. So there would have to be meaningful _polishing_.

AlanCoding:
Here is where nodes get registered in openshift

https://github.com/ansible/awx/blob/e5dbb592fa159688bda1f7fff0430f041a3ab4ef/awx/main/managers.py#L177-L180

Implication is that when this happens, it will have to get added to this table as well. If you look at the code I linked there, I don't know if there is any case where the `else:` gets hit, or it might be a really really unsupported case.

This leads to a bigger picture question - what _is_ the receptor network when on an OpenShift install? Do nodes even have peers?

https://github.com/ansible/awx-operator/blob/ebe5d1e2f4afd7fd329c18420947ff5840670293/roles/installer/templates/config.yaml.j2#L231

The answer appears to be "no". The diagram for the receptor network may be just a collection of dots with no links.

awx/main/models/ha.py
```diff
@@ -54,6 +54,11 @@ def has_policy_changes(self):
         return self._values_have_edits(new_values)
 
 
+class InstanceLink(BaseModel):
+    source = models.ForeignKey('Instance', on_delete=models.CASCADE, related_name='+')
+    target = models.ForeignKey('Instance', on_delete=models.CASCADE, related_name='reverse_peers')
```
AlanCoding, line 59: So we have 2 relations, `instance.peers` and `instance.reverse_peers`? I think I read that right, and that looks good.

AlanCoding (approved):

awx/main/models/ha.py
jbradberry: Yep, you got it.

awx/api/views/__init__.py
```diff
@@ -364,11 +364,14 @@ class InstanceList(ListAPIView):
...
 
+    def get_queryset(self):
+        return super().get_queryset().exclude(node_type='hop')
+
 
 class InstanceDetail(RetrieveUpdateAPIView):
 
     name = _("Instance Detail")
-    model = models.Instance
+    queryset = models.Instance.objects.exclude(node_type='hop')
```
AlanCoding, line 374: Part of me wants to say this shouldn't be necessary. I think it's a weak argument to say that users who can see execution nodes shouldn't be able to see hop nodes.

But to clarify that, if we did show hop nodes in this list, that would leave the burden on the UI to filter then in the instance group instance selector menus. What you have now could be shippable without further work.

To be honest, I'm not a fan of filtering, because it can cause user frustration. Greying out is often a better option.

awx/main/management/commands/register_peers.py
```diff
@@ -0,0 +1,30 @@
+from django.core.management.base import BaseCommand, CommandError
```
fosterseth, line 1: how do we intend to use this command?

jbradberry: I'm imagining this command being used every place the existing `provision_instance` command is.  So, primarily the installer.  I know we just went over how we do have auto-discovery machinery active in `awx/main/tasks.py`, but the setting that enables it can be toggled off and some means of getting the data into the database is still required.  Or, that machinery can be preempted by having the data already entered in.

fosterseth (approved):

awx/api/views/__init__.py
shanemcd: I think I agree with Alan here. Or at least share the confusion. Why do we want to exclude hop nodes?

jbradberry: Recording this here for posterity, but the exclusions of hop nodes I've added are purely for existing endpoints so that users don't get confused by seeing additional hosts that will never be part of their job execution capacity.  After offline discussion, it seemed like you were both fine with this, since the endpoint brought in with #11454 _will_ include the hop nodes.

awx/main/tasks.py
```diff
@@ -494,29 +495,22 @@ def execution_node_health_check(node):
 
 def inspect_execution_nodes(instance_list):
     with advisory_lock('inspect_execution_nodes_lock', wait=False):
-        node_lookup = {}
-        for inst in instance_list:
-            if inst.node_type == 'execution':
-                node_lookup[inst.hostname] = inst
+        # {inst.hostname: inst for inst in instance_list}
+        node_lookup = Instance.objects.in_bulk(field_name='hostname')
```
AlanCoding: What I don't like about this is that it would (presumably) incur an additional query. Normally, 1 query isn't something I should raise a stink about. But this code path is the heartbeat, which has very poor scaling for clusters with lots of nodes and just 1 database node. That's why this code was kind of awkward to begin with, this is a special case where performance is a priority.

awx/main/tasks.py
```diff
@@ -494,29 +495,22 @@ def execution_node_health_check(node):
...
-        for ad in connections:
+        workers = mesh_status['Advertisements']
+        for ad in workers:
             hostname = ad['NodeID']
-            commands = ad.get('WorkCommands') or []
-            worktypes = []
-            for c in commands:
-                worktypes.append(c["WorkType"])
-            if 'ansible-runner' not in worktypes:
+            if not any(cmd['WorkType'] == 'ansible-runner' for cmd in ad['WorkCommands']):
```
AlanCoding: I'd rather not risk `KeyError` exceptions here. You could get assurances from the receptor team that "WorkCommands" will always be there, or you could just change it to a `.get()` with a fallback (the cheap solution)

tools/docker-compose/ansible/roles/sources/templates/receptor-hop.conf.j2
```diff
@@ -10,3 +10,6 @@
 
 - tcp-listener:
     port: 5555
+
+- control-service:
+    service: control
```
AlanCoding, line 15: Does receptorctl really work on the hop node? Have you been able to get it to work?

awx/main/managers.py
```diff
@@ -188,7 +188,7 @@ def get_or_register(self):
 
     def active_count(self):
         """Return count of active Tower nodes for licensing."""
-        return self.all().count()
+        return self.exclude(node_type='hop').count()
```
AlanCoding, line 191: You've been very thorough with these queryset updates. I don't know if we sufficiently asked the question about whether execution nodes should be included in this.

awx/main/management/commands/register_peers.py
```diff
@@ -0,0 +1,31 @@
...
+
+class Command(BaseCommand):
+    """
+    Internal tower command.
+    Register the peers of a receptor node.
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument('source', type=str, help="")
+        parser.add_argument('--peers', type=str, nargs='+', required=True, help="")
```
AlanCoding: Help text here would be ideal to clear up some important expectations about behavior.

If I run

```
awx-manage register_peers foo --peers bar1 bar2
```

(or whatever the intended syntax is)

If "foo" is already a peer of "bar3", what will happen? Does this command just connect peers, or does it make the list absolutely true?

In other similar commands, we had a lot of race conditions as _lists_ of nodes (like the policy instance list) would conflict as multiple commands were run in parallel. This tended to get solved with the `advisory_lock`. I expected a similar thing to happen here. But I'm not sure if I expected the list to be removing unlisted instances from the peers relation...

awx/main/management/commands/list_instances.py
```diff
@@ -13,7 +13,7 @@ class Ungrouped(object):
 
     @property
     def instances(self):
-        return Instance.objects.filter(rampart_groups__isnull=True)
+        return Instance.objects.filter(rampart_groups__isnull=True).exclude(node_type='hop')
```
AlanCoding, line 16: I don't think an IG can have a hop node assigned to begin with. To me, it seems reasonable that the command would list them, although maybe in their own category. It might be good to file an issue instead of tacking it right now.

awx/api/views/__init__.py
```diff
@@ -406,13 +409,15 @@ class InstanceInstanceGroupsList(InstanceGroupMembershipMixin, SubListCreateAtta
...
+        if parent.node_type == 'hop':
+            return {'msg': _(f"Cannot change instance group membership of hop node: {parent.hostname}.")}
         return None
 
 
 class InstanceHealthCheck(GenericAPIView):
 
     name = _('Instance Health Check')
-    model = models.Instance
+    queryset = models.Instance.objects.exclude(node_type='hop')
```
AlanCoding, line 420: Right, you can't run a health check on a hop node.

It might be good to have a test case that you attempt to visit `/api/v2/instances/2/health_check/` and you get an error code. I don't know what error code it should be, but I want to double-check with you that this is working as intended.

Also, the serializer shouldn't link to this if it's a hop node. I don't see that in this diff

awx/main/tasks.py
jbradberry: As it turns out, it is sometimes `None` (when it is a hop node), which I entirely forgot.  I've added a fallback of `or []`.

tools/docker-compose/ansible/roles/sources/templates/receptor-hop.conf.j2
jbradberry: Yes, it does.  @fosterseth and I worked that out in a pairing session, to find out how we could get hop nodes to show up at all in the output of `receptorctl status`.  Since this PR is no longer relying on that behavior for auto-discovery this could technically be dropped, but I believe we had found that the installer and aap-qe does have this turned on for hops so we should probably keep it for consistency.

fosterseth:
Do we need a mechanism to "unlink" peers if someone uses deprovision_instance.py?

awx/api/views/__init__.py
AlanCoding: tested, and this case gives a 404.

It's consistent behavior, I'm totally fine with the solution here.

awx/main/tasks.py
jbradberry: You are right.  Rolled back to a dict comprehension.

awx/api/views/__init__.py
jbradberry: Excellent.

jbradberry:
> Do we need a mechanism to "unlink" peers if someone uses deprovision_instance.py?

@fosterseth No, since the `InstanceLink` model has fks into `Instance`, if one of those goes away the link gets cascade-deleted.

jbradberry:
_But_, we may want a command to be able to manually remove a link that has been cut by other means.

awx/main/management/commands/register_peers.py
jbradberry: Recording here what was said offline: no, it does not make the list absolutely true (i.e. all links _not_ in the `peers` parameter would be deleted).  I would oppose such behavior, since it would make it impossible to incrementally add links.

I absolutely should add help text, though.  :facepalm:

AlanCoding (approved):

PR #11436, opened 2021-12-08
Pulling in upstream changes

AlexSCorey:
##### SUMMARY
This is a sync of upstream with devel
 https://github.com/ansible/awx/pull/11427


awx/main/tasks.py
```diff
@@ -3251,10 +3256,21 @@ def work_type(self):
...
 
+            # cancel job if receptor no longer knows about work item
+            try:
+                receptor_ctl.simple_command(f'work status {self.unit_id}')
+            except RuntimeError:
+                self.task.instance.result_traceback = traceback.format_exc()
+                self.task.instance.save(update_fields=['result_traceback'])
+
+                result = namedtuple('result', ['status', 'rc'])
+                return result('error', 1)
```
AlanCoding, line 3272: We should either remove or revert this commit, because it was reverted downstream and git just messed it up somehow.

rebeccahhh: do we need to revert again downstream?

AlexSCorey: Revert the whole commit, or just this line of code?

AlanCoding: the whole commit 2a11bb4f3bc41feab9c563e2c653f44fb365a8a2. Doesn't matter to me how you do the git process.

PR #11440, opened 2021-12-09
Adds metrics for cloud deployments

benthomasson:
msg: "Adds cloud metrics endpoint"

##### SUMMARY

In order to collect metrics for cloud installations we require an unauthenticated endpoint.   This endpoint should not be exposed outside of the pod when running in openshift/k8s and should not be exposed at all for bare metal installations. 

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION

```
awx: 19.5.1.dev29+g868e0e9af4.d20211209
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/api/views/cloud_metrics.py
+++ b/awx/api/views/cloud_metrics.py
@@ -1,0 +1,42 @@
+# Copyright (c) 2018 Red Hat, Inc.
+# All Rights Reserved.
+
+# Python
+import logging
+
+# Django
+from django.utils.translation import ugettext_lazy as _
+
+# Django REST Framework
+from rest_framework.response import Response
+
+# AWX
+# from awx.main.analytics import collectors
+import awx.main.analytics.subsystem_metrics as s_metrics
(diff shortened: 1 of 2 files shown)


shanemcd:
> This endpoint should not be exposed outside of the pod when running in openshift/k8s and should not be exposed at all for bare metal installations.

It looks like this PR _does_ expose this endpoint unconditionally, so we'll need to identify how we are going to achieve ^ before we merge this.

fosterseth:
is the goal to create an _additional_ metrics endpoint to live alongside the current one we have? Right now this endpoint is going to clobber with our current metrics endpoint defined here https://github.com/ansible/awx/blob/709c439afc962b2526cb143813ba9623f69f5dfb/awx/api/urls/urls.py#L90

PabloHiro:
Closing this PR as this feature has been implemented already.

PR #11441, opened 2021-12-09
Remove unnecessary DEBUG logger level settings

AlanCoding:
##### SUMMARY
I'm just trying to remove some code that doesn't have any true effect here so that our logging config is easier to read.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
This does change a numerical value, before:

```
In [29]: import logging

In [30]: logger = logging.getLogger('awx')

In [31]: logger.level
Out[31]: 10

```

after:

```
In [2]: import logging

In [3]: logger = logging.getLogger('awx')

In [4]: logger.level
Out[4]: 0

```

But as I understand it, the distinction between 0 and 10 only matters if you use custom (lower) log levels. We certainly don't. Even if we did, that should be something that can be handled by the `LOG_AGGREGATOR_LEVEL` setting, and a filter in the logger definition is not where we want it.


rebeccahhh (approved):

AlanCoding:
the awx-operator check failure is concerning, but hard to read.

Issue #11442, opened 2021-12-09
Workflow Visualizer making endless API requests with no result


akus062381:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

When a user creates a workflow and saves and is redirected to the visualizer screen, API requests happen endlessly, with no result.

Note: I am able to recreate this on an ec2 instance as well as a local build.

### AWX version

devel

### Installation method

docker on linux

### Modifications

no

### Operating system

Fedora 35

### Web browser

Chrome

### Steps to reproduce

1. access the templates page.
2. create a new workflow template.
3. when the redirect to the visualizer happens, you will see the problem.

https://user-images.githubusercontent.com/63685497/145466510-0b5de590-ee87-446e-9b58-058ce7bca611.mp4

### Expected results

user would be able to create, save, and launch a workflow.

### Actual results

the visualizer is not working properly and submits endless API requests.

PR #11443, opened 2021-12-09
Fix extra requests when creating WorkFlowJobTemplate

nixocio:
Fix extra requests when creating WorkFlowJobTemplate.

I used useRef to store the el - div, but it was becoming more complex to do the clean up. The useState solved the issue. 

See: https://github.com/ansible/awx/issues/11442

### DIFF
--- a/awx/ui/src/components/AppendBody/AppendBody.js
+++ b/awx/ui/src/components/AppendBody/AppendBody.js
@@ -1,8 +1,8 @@
-import { useEffect } from 'react';
+import { useEffect, useState } from 'react';
 import ReactDOM from 'react-dom';
 
 function AppendBody({ children }) {
-  const el = document.createElement('div');
+  const [el] = useState(document.createElement('div'));
 
   useEffect(() => {
     document.body.appendChild(el);


nixocio:
I triggered the e2e tests.

keithjgrant:
It feels weird putting an element in state here... is that an example you found somewhere? Why does it fix the issue?

Most examples I've seen of portals use an existing element in the DOM... Maybe instead of useState and adding/removing the div from the DOM, we can pass in a unique id prop and get a `div` by id (creating it if it's not found) and use that?

nixocio:
@keithjgrant the solution involving `useRef` would be. Main issue is the `el` is being recreated over and over.

With useRef the clean up is a bit more complex. See: https://stackoverflow.com/questions/67069827/cleanup-ref-issues-in-react

Good article about portals and hooks: https://www.jayfreestone.com/writing/react-portals-with-hooks/

```js
import { useEffect, useRef } from 'react';
import ReactDOM from 'react-dom';

function AppendBody({ children }) {

  const el = useRef(document.createElement('div'));

  useEffect(() => {
    let elRefValue = null;
    if (el.current) {
      elRefValue = el.current;
    }
    document.body.appendChild(el.current);
    return () => {
      if (elRefValue) {
        document.body.removeChild(elRefValue);
      }
    };
  }, []);

  return ReactDOM.createPortal(children, el.current);
}

export default AppendBody;
```

keithjgrant (approved):

tiagodread (approved):

tiagodread:
The broken tests related to this branch is also failing on origin/devel @nixocio feel free to merge it, I'll be working on these failures

PR #11454, opened 2021-12-13
Creates end point and beginning of serializer for receptor mesh

AlexSCorey:
##### SUMMARY
This PR adds an end point for Receptor Mesh, and begins to create the visualizer for it. The end goal here is to have the api view compile data from the Instance Model and the newly created Instance Link model into a response such that the UI only has to make 1 api call to get what it needs to build the mesh visualizer.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -57,6 +57,7 @@ from awx.main.models import (
     Host,
     Instance,
     InstanceGroup,
+    InstanceLink,
     Inventory,
     InventorySource,
     InventoryUpdate,
@@ -4765,6 +4766,28 @@ class ScheduleSerializer(LaunchConfigurationBaseSerializer, SchedulePreviewSeria
         if self.context['request'].method == 'PATCH' and attrs == {'enabled': False}:
             return attrs
         return super(ScheduleSerializer, self).validate(attrs)
+
+
+class InstanceLinkSerializer(BaseSerializer):
+    class Meta:
+        model = InstanceLink
+        fields = ('source', 'target')
+
+    source = serializers.SlugRelatedField(slug_field="hostname", read_only=True)
+    target = serializers.SlugRelatedField(slug_field="hostname", read_only=True)
+
+
+class InstanceNodeSerializer(BaseSerializer):
+    class Meta:
+        model = Instance
+        fields = ('id', 'hostname', 'node_type', 'node_state')
+
+    node_state = serializers.SerializerMethodField()
+
+    def get_node_state(self, obj):
+        if not obj.enabled:
+            return "disabled"
+        return "unhealthy" if obj.errors else "healthy"
 
 
 class InstanceSerializer(BaseSerializer):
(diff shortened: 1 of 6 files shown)


awx/api/templates/api/mesh_visualizer.md
```diff
@@ -0,0 +1 @@
+Make a GET request to this resource to obtain a list all Receptor Nodes and their links.
```
shanemcd, line 1: Are we planning to fill this out a little more? If not I would lean towards just deleting this.

shanemcd (approved):

awx/api/templates/api/mesh_visualizer.md
AlexSCorey: I think this was created with the understanding that this component would increase in functionality over time.  But, as that happens we can update this file....or delete it now and create this file later.

awx/api/serializers.py
```diff
@@ -4767,6 +4768,28 @@ def validate(self, attrs):
         return super(ScheduleSerializer, self).validate(attrs)
 
 
+class InstanceLinkSerializer(BaseSerializer):
+    class Meta:
+        model = InstanceLink
+        fields = ('source', 'target')
+
+    source = serializers.SlugRelatedField(slug_field="hostname", read_only=True)
+    target = serializers.SlugRelatedField(slug_field="hostname", read_only=True)
```
AlanCoding, line 4777: I don't see any other uses of this field type in the file. What do the OPTIONS for the endpoint using this look like?

AlanCoding: well I guess OPTIONS won't be useful for anything, since it's not employed in a real list view.

awx/api/serializers.py
```diff
@@ -4767,6 +4768,28 @@ def validate(self, attrs):
...
+    class Meta:
+        model = Instance
+        fields = ('id', 'hostname', 'node_type', 'node_state')
+
+    node_state = serializers.SerializerMethodField()
+
+    def get_node_state(self, obj):
+        if not obj.enabled:
+            return "disabled"
+        return "unhealthy" if obj.errors else "healthy"
```
AlanCoding, line 4790: We don't check for the health of hop nodes. Seeing this, I realize that we should.

To outline that in better clarity, if the `receptorctl` data shows that a hop node has failed to check in within the configured receptor broadcast frequency (practically, 60 seconds), then it should have some errors filled in. Likewise, the errors should be cleared if it appears on the mesh afterwards.

Could you file an issue for that and add to the project so that we make sure to cover it? The purpose of doing this would be strictly to make the mesh graph data accurate. Thanks.

AlanCoding (approved):
Seems thoughtful but not overly-complicated.

awx/api/serializers.py
AlexSCorey: OPTIONS for this end point look exactly the same as OPTIONS for `/job_templates/labels`

PR #11474, opened 2021-12-16
adding event handler specific to when awx-rsyslog throws PROCESS_LOG_STDERR

sarabrajsingh:
this code change introduces a handler to handle a log trace that contains a 4XX error in the awx-rsyslog logs. as mentioned in the support case, the 4XX event causes the `syslog` to hang while seeming operational in supervisord. this handler sends a SIGTERM signal to the process, and supervisord will attempt to reboot the service.

### DIFF
--- a/tools/scripts/failure-event-handler
+++ b/tools/scripts/failure-event-handler
@@ -2,24 +2,72 @@
 import sys
 import os
 import signal
+import datetime
+
+from datetime import timezone
+
 
 def write_stdout(s):
     sys.stdout.write(s)
     sys.stdout.flush()
 
+
 def write_stderr(s):
     sys.stderr.write(s)
     sys.stderr.flush()
 
+
 def main():
     while 1:
-        write_stdout('READY\n')
+        write_stdout("READY\n")
+
+        # read header line and print it to stderr
         line = sys.stdin.readline()
-        headers = dict([ x.split(':') for x in line.split() ])
-        headers.update(dict([ x.split(':') for x in sys.stdin.read(int(headers['len'])).split()]))
-        if headers['eventname'] == 'PROCESS_STATE_FATAL':
-            os.kill(os.getppid(), signal.SIGTERM)
-        write_stdout('RESULT 2\nOK')
+
+        # parse header line and decide what to do with it
+        try:
+            headers = dict([x.split(":") for x in line.split()])
+            data = sys.stdin.read(int(headers["len"]))
+        except ValueError as e:
+            write_stderr(str(e))
+
+        # now decide what do to based on eventnames
+        if headers["eventname"] == "PROCESS_STATE_FATAL":
+            headers.update(
+                dict(
+                    [x.split(":") for x in sys.stdin.read(int(headers["len"])).split()]
+                )
+            )
+
+            try:
+                # incoming event that produced PROCESS_STATE_FATAL will have a PID. SIGTERM it!
+                write_stderr(
+                    f"{datetime.datetime.now(timezone.utc)} - sending SIGTERM to proc={headers} with data={headers}\n"
+                )
+                os.kill(headers["pid"], signal.SIGTERM)
+            except Exception as e:
+                write_stderr(str(e))
+
+        # awx-rsyslog PROCESS_LOG_STDERR handler
+        if headers["eventname"] == "PROCESS_LOG_STDERR":
+            # pertinent data to process that produced PROCES_LOG_STDERR is in the first line of the data payload; so lets extract it
+            proc_details = dict([x.split(":") for x in (data.split("\n")[0]).split()])
+
+            if proc_details["processname"] == "awx-rsyslogd":
+                log_message = "".join(data.split("\n")[1:])
+
+                # look for a 4XX HTTP CODE in the log message. if found, issue a sigkill
+                if any(str(x) in log_message.split() for x in range(400, 420)):
+                    try:
+                        write_stderr(
+                            f"{datetime.datetime.now(timezone.utc)} - sending SIGTERM to proc=[{proc_details['processname']}] with pid=[{int(proc_details['pid'])}] due to log_message=[{log_message}]\n"
+                        )
+                        os.kill(int(proc_details["pid"]), signal.SIGTERM)
+                    except Exception as e:
+                        write_stderr(str(e))
+
+        write_stdout("RESULT 2\nOK")
+
 
 if __name__ == "__main__":
     main()


AlanCoding:
I don't quite understand the objective of this new behavior. Doesn't the 4XX event come from the external log collector? What if restarting rsyslog doesn't change that behavior (it probably won't), will it be stuck in a reboot loop forever? Or have I misunderstood the meaning of the 4XX error?

sarabrajsingh:
@AlanCoding you're correct in assuming that there could be a possibility of a 4XX error causing an infinite restart loop for the `awx-rsyslog` service. i think we should let end-users handle this scenario.

this code change was an ask from a customer support ticket, and i guess what they were trying to achieve is a handler for intermittent/random 4XX errors that would cause the `awx-rsyslog` service to malfunction. one 4XX error returned to `awx-rsyslog` would cause the logging service to stop. this could have been caused by an extraneous event in a user's environment. one step closer to total fault-tolerance :)

AlanCoding:
Gotcha, yes, it is a completely reasonable expectation that we would get some 4XX errors and then be able to recover. I just immediately assumed that getting those errors would be due to a permanent configuration error.

AlanCoding:
> the 4XX event causes the syslog to hang while seeming operational in supervisord.

This point caused me some confusion.

The way this works is that the script is specified in the supervisor conf under `[eventlistener:exit-event-listener]`. So yes, the process stops, but it's not completely unrecognized by supervisor. It feeds info about the exit to this script.

This framing makes the potential for an infinite restart seem less bad to me. This would only happen in the event that the rsyslog process exits.

If you feel confident from testing this, it appears reasonable to me.

EDIT: looks like I may be wrong, and this is triggering a restart when the process writes to stderr, maybe.

tools/scripts/failure-event-handler
```diff
@@ -2,24 +2,75 @@
...
+
+        # parse header line and decide what to do with it
+        try:
+            headers = dict([x.split(":") for x in line.split()])
+            data = sys.stdin.read(int(headers["len"]))
+        except ValueError as e:
+            write_stderr(str(e))
+
+        # now decide what do to based on eventnames
+        if headers["eventname"] == "PROCESS_STATE_FATAL":
```
AlanCoding, line 35: Why bother initializing `headers = None` if you're going to assume it's a dict here?

tools/scripts/failure-event-handler
```diff
@@ -2,24 +2,75 @@
...
+                    [x.split(":") for x in sys.stdin.read(int(headers["len"])).split()]
+                )
+            )
+
+            try:
+                # incoming event that produced PROCESS_STATE_FATAL will have a PID. SIGTERM it!
+                write_stderr(
+                    f"{datetime.datetime.now(timezone.utc)} - sending SIGTERM to proc={headers} with data={headers}\n"
+                )
+                os.kill(headers["pid"], signal.SIGTERM)
```
AlanCoding, line 47: So is this a change in behavior from before when it was

```python
os.kill(os.getppid(), signal.SIGTERM)
```

It's true that yours looks better, but it causes me some concerns about the exception cases. If the headers doesn't have a "len" value, but the "eventname" is PROCESS_STATE_FATAL, maybe we really do want to SIGTERM `os.getppid()`

tools/scripts/failure-event-handler
```diff
@@ -2,24 +2,75 @@
...
+            except Exception as e:
+                write_stderr(str(e))
+
+        # awx-rsyslog PROCESS_LOG_STDERR handler
+        if headers["eventname"] == "PROCESS_LOG_STDERR":
+            # pertinent data to process that produced PROCES_LOG_STDERR is in the first line of the data payload; so lets extract it
+            proc_details = dict([x.split(":") for x in (data.split("\n")[0]).split()])
```
AlanCoding, line 54: A performance concern here, does this get ran _every_ time any process writes anything to stderr? If something is going crazy, I wonder if that would end up processing a lot of text (even though what it's doing is fairly simple).

AlanCoding:
For whatever changes this is expecting to be made in the supervisor conf, I'd like at least some of those to be reflected in the dev environment:

https://github.com/ansible/awx/blob/devel/tools/docker-compose/supervisor.conf

We could hypothetically connect the failure-event-handler script there, so that it could be tested in this context. Don't have to.

AlanCoding:
From the get-go, I had a specific gut reaction to this - that we should fix this with the `rsyslog.conf` config file, as opposed to altering the exit handler. I see now that this goes beyond just editing the event handler, and really changes the role of the script to _both_ be an exit handler and watch for breadcrumbs in stderr. This makes me want to double-down on that intuition I had.

It's hard to follow, so let me link to where the `rsyslog.conf` is built:

https://github.com/ansible/awx/blob/f0129c857b01d9949a68492a6934b908fe032982/awx/main/utils/external_logging.py#L31-L37

This will have the details of the external log aggregator in it (like the hostname to send stuff to).

We actually set `action.resumeRetryCount="-1"`, and I'm curious if we could just change this value and maybe some other values to fix this issue.

AlanCoding (approved):
After reading more of the history, it looks like other people have found rsyslog to have a pretty big shortcoming here, and the better option is likely to not pan out.

AlanCoding:
Among these settings:

https://github.com/ansible/awx/blob/f0129c857b01d9949a68492a6934b908fe032982/awx/main/utils/external_logging.py#L71-L79

I thought `action.reportSuspension` looked the closest to what we would want to indicate "don't stop sending messages after a 400 error". But docs are terribly unclear on this.

https://www.rsyslog.com/doc/master/configuration/actions.html

sarabrajsingh:
> For whatever changes this is expecting to be made in the supervisor conf, I'd like at least some of those to be reflected in the dev environment:
> 
> https://github.com/ansible/awx/blob/devel/tools/docker-compose/supervisor.conf
> 
> We could hypothetically connect the failure-event-handler script there, so that it could be tested in this context. Don't have to.

I can definitely introduce these changes into the `dev` supervisor.conf file

sarabrajsingh:
> For whatever changes this is expecting to be made in the supervisor conf, I'd like at least some of those to be reflected in the dev environment:
> 
> https://github.com/ansible/awx/blob/devel/tools/docker-compose/supervisor.conf
> 
> We could hypothetically connect the failure-event-handler script there, so that it could be tested in this context. Don't have to.

I can definitely introduce these changes into the `dev` supervisor.conf file

sarabrajsingh:
> > For whatever changes this is expecting to be made in the supervisor conf, I'd like at least some of those to be reflected in the dev environment:
> > https://github.com/ansible/awx/blob/devel/tools/docker-compose/supervisor.conf
> > We could hypothetically connect the failure-event-handler script there, so that it could be tested in this context. Don't have to.
> 
> I can definitely introduce these changes into the `dev` supervisor.conf file

@AlanCoding after taking a look at our `dev` supervisor.conf file, looks like @shanemcd already wrote some code to redirect all `stdout_events` and `stderr_events` generated by any supervisord-managed proc to a stdout listener. I think this is sufficient enough as-is. what do you think?

PR #11481, opened 2021-12-19
more time-out resilient project-update playbook

velzend:
##### SUMMARY
In our environment/ setup we switched from using Git to our own HTTP(S) download service which uses a on-premise S3 as the back-end to store and pull Ansible Projects/ S3 objects. This migration made project updates much faster, and the S3 service is very durable, reliable and always up... but sometimes it coops with some latency (issues) that would exceed the default 10-seconds download time-out for the get_url and ansible-galaxy update roles and collection tasks.

We've added some retry parameters to ensure the download of the project archive and the task which executes Ansible Galaxy to fetch Ansible collections and Ansible roles does not immediately fail when time-outs occur. Any other error like [403] Authentication issues, or [404] File not found or [5xx] errors will make the task fail immediately. 

We have this change already implemented in our production environment, which resulted in proper handling of time-out/ latency issues, and no automatic incidents being created of the very few (<0,001%) project updates that would have been failed. Now these project updates do finish after one or two retries... 

##### ISSUE TYPE
 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
awx: 14.1.1.dev4206+gddc428532f.d20211218
```

##### ADDITIONAL INFORMATION
Added below parameters to the "Get archive from url" task:
```
          retries: 12
          delay: 10
          until: >
            get_archive is succeeded or
            get_archive.status_code is defined 
```

Added below parameters to the fetch galaxy roles and collections tasks:
```
retries: 12
          delay: 10
          until: >
            galaxy_collection_result is succeeded or
            'The read operation timed out' not in galaxy_collection_result.stderr

```

### DIFF
--- a/awx/playbooks/project_update.yml
+++ b/awx/playbooks/project_update.yml
@@ -103,6 +103,9 @@
             dest: "{{ project_path|quote }}/.archive"
             state: directory
 
+        # the status_code is only defined if the request was handled
+        # all status_codes >= 400 are marked as failed
+        # 10s timeout 1st attempt + 10s delay * 12 retries == 250s
         - name: Get archive from url
           get_url:
             url: "{{ scm_url|quote }}"
@@ -111,6 +114,11 @@
             url_password: "{{ scm_password|default(omit) }}"
             force_basic_auth: true
           register: get_archive
+          retries: 12
+          delay: 10
+          until: >
+            get_archive is succeeded or
+            get_archive.status_code is defined 
 
         - name: Unpack archive
           project_archive:
@@ -172,6 +180,10 @@
         - install_collections
 
     - block:
+
+        # when a timeout occurs ansible-galaxy will result in:
+        # rc: 1 and "[ERROR]: failed to download the file: The read operation timed out"
+        # approx. 10s timeout first run + 10s delay * 12 retries == 250s
         - name: fetch galaxy roles from requirements.(yml/yaml)
           command: >
             ansible-galaxy role install -r {{ item }}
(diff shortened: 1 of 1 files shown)


awx/playbooks/project_update.yml
```diff
@@ -111,6 +114,11 @@
             url_password: "{{ scm_password|default(omit) }}"
             force_basic_auth: true
           register: get_archive
+          retries: 12
+          delay: 10
+          until: >
+            get_archive is succeeded or
+            get_archive.status_code is defined
```
AlanCoding, line 121: This would increase the time it takes for any project update to fail due to unrelated problems, and this would make the user experience worse. You've done some investigating into what the timeout conditions look like, and I would like to see the `until` condition reflect those specifics. Like, could it retry until the status_code is not one of (504, 408, maybe others?). That would still allow for quick fails in the event that something is misconfigured.

AlanCoding: You said

> Any other error like [403] Authentication issues, or [404] File not found or [5xx] errors will make the task fail immediately.

But I'm not following how. If status_code is 403, why wouldn't this retry?

velzend: It will continue if succeeded which is "almost always" for the get_url task or it will retry until the status code is defined... A status code is not defined if the request times out.

velzend: But let me run and share some test results...

AlanCoding:
Sorry I haven't given this more attention. I'm a little too nervous to add this for the ansible-galaxy task, but for the archive type SCM this could make sense. I'm going to close it for now, but you are free to re-open if you want to revisit.

piellick:
Hi  @velzend  @AlanCoding same problem for us, random timeout with our git project. A way to try this patch on AWX 19.5.0 ? How to tweak this playbook throught the operator ?

PR #11501, opened 2022-01-04
Add mesh viz

kialam:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - UI

Screen Shot 2021-12-10 at 11 22 26 AM
Screen Shot 2021-12-17 at 9 15 07 AM


awx/ui/src/screens/TopologyView/MeshGraph.js
```diff
@@ -0,0 +1,434 @@
...
+      .force('center', d3.forceCenter(width / 2, height / 2));
+
+    const link = svg
+      .append('g')
+      .attr('class', `links`)
+      .selectAll('path')
+      .data(graph.links)
+      .enter()
+      .append('path')
+      .attr('class', (d, i) => `link-${i}`)
```
tiagodread: We'll need something more explicit here in order to be able to verify the link between nodes in UI tests, something like `${hostSource}-${hostTarget}` or `${hostSourceId}-${hostTargetId}`

awx/ui/src/screens/TopologyView/MeshGraph.js
```diff
@@ -0,0 +1,434 @@
...
+      .style('fill', 'none')
+      .style('stroke', '#ccc')
+      .style('stroke-width', '2px')
+      .attr('pointer-events', 'none')
+      .on('mouseover', function (event, d) {
```
tiagodread: Is it possible to set an id to the tooltip on `mouseover` event so that we can assert the value inside?

AlanCoding:
It looks like we need this API change to be patched into this for the hop nodes to work correctly

https://github.com/ansible/awx/pull/11714

I have steps to test adding and killing off hop nodes if you need. The criteria I meant to hand off to the UI was that any instance with non-blank `errors` field should be shown as non-functioning / dead / grey.

Let me know what you want to do with that, if you want to cherry pick it into here I can just close it there.

kialam:
@AlanCoding thanks for that. The UI keys off the `node_state` value returned from the `/mesh_visualizer` endpoint for node "health" attribute so as long as that logic accounts for the changes above (e.g. non-empty error field) then all should be well. cc @AlexSCorey

AlanCoding:
Yes, that has the correct criteria:

https://github.com/ansible/awx/blob/750e1bd80a822127ac3f403c755834ade4787560/awx/api/serializers.py#L4793

kialam:
@AlexSCorey @jbradberry Can we change the node state from 'unhealthy' to 'error'? The reason I ask is that the UI uses the [StatusLabel](https://github.com/ansible/awx/blob/c1b2428b5da9c881f0189949ef34a9da3f0658d2/awx/ui_next/src/components/StatusLabel/StatusLabel.js#L92) component to render node state. There is no 'unhealthy' label and we should be using 'error' instead.

kialam:
Unit tests will be in a separate PR. I will link to it once it's up.

awx/ui/src/screens/TopologyView/MeshGraph.js
```diff
@@ -0,0 +1,292 @@
...
+    const link = mesh
+      .append('g')
+      .attr('class', `links`)
+      .attr('data-cy', 'links')
+      .selectAll('line')
+      .data(graph.links)
+      .enter()
+      .append('line')
+      .attr('class', (_, i) => `link-${i}`)
+      .attr('data-cy', (d) => `${d.source}-${d.target}`)
```
tiagodread: image
```suggestion
      .attr('data-cy', (d) => `${d.source.hostname}-${d.target.hostname}`)
```

awx/ui/src/screens/TopologyView/ContentLoading.js
```diff
@@ -0,0 +1,42 @@
...
+        style={{ fontWeight: 'bold', color: 'black' }}
+      >
+        {t`Please wait until the topology view is populated...`}
+      </Text>
+    </TextContent>
+    <EmptyStateIcon variant="container" component={Spinner} />
+  </EmptyState>
+);
+
+export { ContentLoading as _ContentLoading };
```
nixocio:
```suggestion

```
This export is an old standard used to export components to test. We have not removed from other places, but new components do not require this one.

awx/ui/src/screens/TopologyView/utils/helpers.js
```diff
@@ -0,0 +1,89 @@
...
+export function renderNodeIcon(selectedNode) {
+  if (selectedNode) {
+    const { node_type: nodeType } = selectedNode;
+    return NODE_TYPE_SYMBOL_KEY[nodeType];
+  }
+
+  return false;
+}
+
+export async function redirectToDetailsPage(selectedNode, history) {
```
nixocio:
```suggestion
export function redirectToDetailsPage(selectedNode, history) {
```

awx/ui/src/screens/TopologyView/MeshGraph.js
```diff
@@ -0,0 +1,274 @@
...
+import ContentLoading from './ContentLoading';
+import {
+  renderStateColor,
+  renderLabelText,
+  renderNodeType,
+  renderNodeIcon,
+  redirectToDetailsPage,
+  getHeight,
+  getWidth,
+  // generateRandomNodes,
```
nixocio:
```suggestion
```

awx/ui/src/screens/TopologyView/MeshGraph.js
```diff
@@ -0,0 +1,274 @@
...
+  // getRandomInt,
```
nixocio:
```suggestion

```

awx/ui/src/screens/TopologyView/MeshGraph.js
```diff
@@ -0,0 +1,274 @@
...
+} from './constants';
+
+const Loader = styled(ContentLoading)`
+  height: 100%;
+  position: absolute;
+  width: 100%;
+  background: white;
+`;
+function MeshGraph({ data, showLegend, zoom }) {
+  // function MeshGraph({ showLegend, zoom }) {
```
nixocio:
```suggestion

```

awx/ui/src/screens/TopologyView/MeshGraph.js
```diff
@@ -0,0 +1,274 @@
...
+  const [isNodeSelected, setIsNodeSelected] = useState(false);
+  const [selectedNode, setSelectedNode] = useState(null);
+  const [nodeDetail, setNodeDetail] = useState(null);
+  const history = useHistory();
+
+  // const data = generateRandomNodes(getRandomInt(4, 50));
```
nixocio:
```suggestion
```

awx/ui/src/screens/TopologyView/MeshGraph.js
```diff
@@ -0,0 +1,274 @@
...
+      .attr('r', DEFAULT_RADIUS)
+      .attr('class', (d) => d.node_type)
+      .attr('class', (d) => `id-${d.id}`)
+      .attr('fill', DEFAULT_NODE_COLOR)
+      .attr('stroke', DEFAULT_NODE_LABEL_TEXT_COLOR);
+
+    // node type labels
+    node
+      .append('text')
+      .text((d) => renderNodeType(d.node_type))
```
nixocio: Would possible to place the text at center of the node?

image

awx/ui/src/screens/TopologyView/MeshGraph.js
```diff
@@ -0,0 +1,274 @@
+import React, { useEffect, useState } from 'react';
+import { useHistory } from 'react-router-dom';
+import styled from 'styled-components';
+import debounce from 'util/debounce';
+// import { t } from '@lingui/macro';
```
nixocio:
```suggestion

```

nixocio (commented):
Hey @kialam, excellent job. I left a couple of minor suggestions. 

I'm still looking through.

awx/ui/src/screens/TopologyView/MeshGraph.js
tiagodread: This is happening on Firefox browser

tiagodread:
Large inventories can take a while to be rendered, might be interesting to let the user know about it changing the text in the loading state from `Please wait until the topology view is populated...`  by `Please wait until the topology view is populated (it can take more time for large inventories)...` 

Since the toolbar buttons (zoom/fit etc) can move the graph and recauculate everything, I'd sugest disabling those buttons while the graph is been loaded

awx/ui/src/screens/TopologyView/MeshGraph.js
kialam: This should now be resolved with the latest [commit](https://github.com/ansible/awx/pull/11501/commits/fee47fe347124e0201c88bddd9c0a752e577635c).

kialam:
@nixocio thank you for your review. I've pushed up the latest changes to address your feedback. 👍

tiagodread:
@kialam on a production build running `make ui-devel` the loading just hangs on `0%` with some console errors:

image

The Mesh viz is never rendered.

kialam:
> @kialam on a production build running `make ui-devel` the loading just hangs on `0%` with some console errors:
> 
> image
> 
> The Mesh viz is never rendered.

@tiagodread thanks for finding this. The latest changes should fix the issue.

awx/ui/src/util/simulationWorker.js
```diff
@@ -0,0 +1,34 @@
...
+    )
+    .force('collide', d3.forceCollide(62))
+    .force('forceX', d3.forceX(0))
+    .force('forceY', d3.forceY(0))
+    .stop();
+
+  for (
+    let i = 0,
+      n = Math.ceil(
+        Math.log(simulation.alphaMin()) / Math.log(1 - simulation.alphaDecay())
```
nixocio, line 24: 🥇

nixocio (approved):
@kialam, excellent job!

Even though this a complex feature, you broke down in small pieces that make it more digestible. 🙇 .

Also, do we have plans to squash those commits?

I read a bit more about web workers, I wonder if we will need to update CSP as well later. 

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/worker-src

kialam:
@nixocio great point about the CSP. I will open up a separate PR to add d3 as an allowable web worker source.

Issue #1236, opened 2018-02-13
Log response message along with API failure codes


wenottingham:
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
 - API

##### SUMMARY

A typical error log from AWX is:

```
2018-02-12 21:41:29,559 WARNING  awx.api.generics status 400 received by user SuperCoolUser attempting to access /api/v2/job_templates/42/extra_credentials/ from 127.0.0.12
```

It would help debugging API failures (especially provisioning callbacks) if the JSON error returned to the user in this case was also included in the log.

##### ENVIRONMENT
* AWX version: current

AlanCoding:
Log message in question:

https://github.com/ansible/awx/blob/fd99b366c29bf3ebcae713711053eadce160ecc4/awx/api/generics.py#L211

I imagine that this can obviously be accomplished by adding `response.data` in that. That might give long log messages...

CFSNM:
This issue has been validated manually on QE side. Once the PR https://github.com/ansible/awx/pull/11527 is merged, this issue can be closed

PR #11527, opened 2022-01-12
Api 4XX error msg customization 1236

john-westcott-iv:
##### SUMMARY

Give ability to customize the the 4XX error messages that come out of the API in the tower.log file. 

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API
 - UI

##### AWX VERSION

```
1.0.7.dev13817+gf73bb0e6a7.d20220112
```

##### ADDITIONAL INFORMATION

Addresses #1236.

```

```


awx/api/generics.py
```diff
@@ -208,19 +209,37 @@ def finalize_response(self, request, response, *args, **kwargs):
...
+                    logger.error("Unable to format API_400_ERROR_LOG_FORMAT setting, defaulting log message")
+                status_msg = (
+                    settings_registry.get_setting_field('API_400_ERROR_LOG_FORMAT')
+                    .get_default()
+                    .format(
+                        status_code=response.status_code,
+                        user_name=request.user,
+                        url_path=request.path,
+                        remote_addr=request.META.get('REMOTE_ADDR', None),
+                        error=response.data.get('error', None),
```
AlanCoding: So you don't repeat yourself, I'd prefer to dump these parameters into a dict before-hand. The `error` value is different in the 2 cases.

AlanCoding:
Surfacing the setting in the UI seems a _little_ bit like overkill to me (I defer to @ffirg for that decision), but even if we backed that out, I would keep this as a file-based setting (and the rest of the implementation here). I certainly foresee _some_ future case where a user needs to reduce the line length of this log.

awx/api/tests/functional/test_api_generics.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+
+# We have to tell flake 8 to ignore these since they are fixtures we need to import
+from awx.conf.tests.functional.conftest import api_request, admin  # noqa: F401; pylint: disable=unused-variable
```
AlanCoding: The linter isn't happy about the unused `api_request`, and maybe I kind of agree. Why do you think they should be imported? Pytest handles fixtures for you, something in the fixture list doesn't need an import.

john-westcott-iv: I could be off base here but I thought the conftest was only picked up if its in the same directory as the test. To avoid code duplication I am pulling in fixtures from awx/conf/tests whereas this test is in awx/api/tests. So without the import `py.test awx/api/tests/functional/test_api_generics.py` yields:
```
_______________________________________________________________ ERROR at setup of test_custom_400_error_log ________________________________________________________________
file /awx_devel/awx/api/tests/functional/test_api_generics.py, line 22
  @pytest.mark.django_db
  def test_custom_400_error_log(caplog, api_request, settings):
E       fixture 'api_request' not found
>       available fixtures: _dj_autoclear_mailbox, _django_clear_site_cache, _django_db_helper, _django_db_marker, _django_set_urlconf, _django_setup_unittest, _fail_for_invalid_template_variable, _live_server_helper, _template_string_if_invalid_marker, admin_client, admin_user, async_client, async_rf, cache, capfd, capfdbinary, caplog, capsys, capsysbinary, client, cov, db, django_assert_max_num_queries, django_assert_num_queries, django_capture_on_commit_callbacks, django_db_blocker, django_db_createdb, django_db_keepdb, django_db_modify_db_settings, django_db_modify_db_settings_parallel_suffix, django_db_modify_db_settings_tox_suffix, django_db_modify_db_settings_xdist_suffix, django_db_reset_sequences, django_db_serialized_rollback, django_db_setup, django_db_use_migrations, django_mail_dnsname, django_mail_patch_dns, django_test_environment, django_user_model, django_username_field, doctest_namespace, live_server, mailoutbox, mock, mocker, monkeypatch, no_cover, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, rf, settings, testrun_uid, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory, transactional_db, worker_id
>       use 'pytest --fixtures [testpath]' for help on them.
```

awx/api/tests/functional/test_api_generics.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@pytest.mark.django_db
+def test_change_400_error_log(caplog, api_request, settings):
+    settings.API_400_ERROR_LOG_FORMAT = 'Test'
```
AlanCoding: I think it's best to use `overwrite_settings` like this

https://github.com/ansible/awx/blob/eec4f8dcc27428cd1aeb1c251d43052a3ea780e8/awx/main/tests/functional/commands/test_secret_key_regeneration.py#L34

awx/api/tests/functional/test_api_generics.py
john-westcott-iv: Resolved by moving tests.

john-westcott-iv:
> Surfacing the setting in the UI seems a _little_ bit like overkill to me (I defer to @ffirg for that decision), but even if we backed that out, I would keep this as a file-based setting (and the rest of the implementation here). I certainly foresee _some_ future case where a user needs to reduce the line length of this log.

My main thought behind this was that an admin could add the `{error}` temporarily for troubleshooting by saving the settings page vs having to change the file-based settings and restarting the service. The other option I thought about would be to key off `LOG_AGGREGATOR_LEVEL` and providing a fixed format message but I already feel like that is misused for LDAP/SAML debugging and it wouldn't allow for the user to reduce the length of the msg.

@ffirg any thoughts on this?

awx/main/tests/functional/test_api_generics.py
```diff
@@ -0,0 +1,26 @@
...
+
+from django.test.utils import override_settings
+from awx.api.versioning import reverse
+
+
+@pytest.mark.django_db
+def test_change_400_error_log(caplog, post, admin_user):
+    with override_settings(API_400_ERROR_LOG_FORMAT='Test'):
+        post(url=reverse('api:setting_logging_test'), data={}, user=admin_user, expect=409)
+        assert 'Test' in caplog.text
```
AlanCoding, line 11: Very nice use of `caplog` btw!

AlanCoding (approved):

ffirg:
I'd prefer the file based approach over the UI setting, as it's likely that a user would turn this on for debugging and likely leave it on 'by mistake'. Can't we just make this the default for a file based solution as well?

ffirg (approved):

nixocio:
@john-westcott-iv I updated the UI failing test, and triggered the UI e2e tests.

john-westcott-iv:
Acceptance Criteria:
- [X] Users should be able to change the format of the API 4XX error messages.
- [X] Users should be able to display the error returned from the API if desired.
- [X] After changes the logs should reflect the new format.

PR #11528, opened 2022-01-12
Splitting 'tasks.py' file  into a module

amolgautam25:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 16.0.1.dev3147+gd31c74fe50
```

##### ADDITIONAL INFORMATION

```
--- Changed tasks.py to a directory module containing 3 files
--- Changed the reference to awx.main.tasks to proper awx.main.tasks.<sub-module> 
--- Changed the corresponding unit tests on awx repo
```


shanemcd:
It looks like the CI checks for awx-operator are failing. If you search the logs for `Re-emit failure`, you'll see that either the demo job template or project update failed. Unfortunately the actual error is hidden. When I added that CI check, I intended for the errors to be printed out under the task `Get all job and project details`... but that isn't happening.

Maybe try changing this line: https://github.com/ansible/awx/blob/devel/.github/workflows/ci.yml#L210

to this:

```
KUSTOMIZE_PATH=$(readlink -f bin/kustomize) molecule -v test -s kind
```

awx/main/tasks/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2015 Ansible, Inc.
+# All Rights Reserved.
+
+# Python
+import logging
+
+logger = logging.getLogger('awx.main.tasks')
```
AlanCoding: It doesn't make sense to get the logger in this file because it's not being used. Should be a safe delete.

awx/main/tasks/receptor.py
```diff
@@ -0,0 +1,318 @@
...
+from awx.main.utils.common import (
+    deepmerge,
+    parse_yaml_or_json,
+    cleanup_new_process,
+)
+
+
+# AWX
+from awx.main.utils.receptor import get_receptor_ctl, get_conn_type, get_tls_client
+from awx.main.tasks import logger
```
AlanCoding: In these cases, I would advocate defining the logger in _this_ file, and using the "awx.main.tasks.receptor" logger. I welcome any other opinions.

awx/main/tasks/receptor.py
```diff
@@ -0,0 +1,318 @@
...
+import ansible_runner.cleanup
+from awx.main.utils.common import (
+    deepmerge,
+    parse_yaml_or_json,
+    cleanup_new_process,
+)
+
+
+# AWX
+from awx.main.utils.receptor import get_receptor_ctl, get_conn_type, get_tls_client
```
AlanCoding: It's likely possible to move code from `awx.main.utils.receptor` into this file, and that would make the most sense to me.

tools/ansible/roles/dockerfile/files/settings.py
```diff
@@ -55,6 +55,9 @@ def get_secret():
 LOGGING['loggers']['awx']['handlers'] = ['console', 'external_logger']
 LOGGING['loggers']['awx.main.commands.run_callback_receiver']['handlers'] = ['console']
 LOGGING['loggers']['awx.main.tasks']['handlers'] = ['console', 'external_logger']
+
+# not sure how to change this
```
AlanCoding: I don't envision any changes to the `LOGGING` definition, because `awx.main.tasks` will still be the parent logger of the new python modules. Let me know if anything is unclear.

awx/main/tasks/__init__.py
```diff
@@ -0,0 +1,6 @@
+# __init__.py
+
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2015 Ansible, Inc.
```
AlanCoding: We don't really need these headers. An empty file would be fine with me.

awx/main/tasks/jobs.py
```diff
@@ -1,77 +1,52 @@
-# -*- coding: utf-8 -*-
-
-# Copyright (c) 2015 Ansible, Inc.
-# All Rights Reserved.
+# jobs.py
```
AlanCoding: I think line is still to-be-deleted?

awx/main/tasks/jobs.py
```diff
@@ -1,77 +1,52 @@
...
+# Runner
+import ansible_runner
 
 # GitPython
 import git
 from gitdb.exc import BadName as BadGitName
 
-# Runner
-import ansible_runner
-import ansible_runner.cleanup
```
AlanCoding, line 55: It's good to see groups of imports like this go away, to me, it helps justify that this makes sense.

awx/main/tasks/receptor.py
```diff
@@ -0,0 +1,536 @@
+# receptor.py
```
AlanCoding: again, I assume this line gets deleted

awx/main/tasks/system.py
```diff
@@ -0,0 +1,915 @@
...
+from django.utils.translation import gettext_noop
+from django.core.cache import cache
+from django.core.exceptions import ObjectDoesNotExist
+
+# Django-CRUM
+from crum import impersonate
+
+
+# Runner
+import ansible_runner
```
AlanCoding: I don't see any use of `ansible_runner` in this file which is not `ansible_runner.cleanup`. Looks like a special case the linter might not catch.

awx/main/tasks/receptor.py
```diff
@@ -0,0 +1,536 @@
...
+import threading
+import time
+import yaml
+
+# Django
+from django.conf import settings
+
+# Runner
+import ansible_runner
+import ansible_runner.cleanup
```
AlanCoding: I think this import of `ansible_runner.cleanup` can be deleted.

awx/main/tasks/system.py
```diff
@@ -0,0 +1,915 @@
...
+from awx.main import analytics
+from awx.conf import settings_registry
+from awx.main.analytics.subsystem_metrics import Metrics
+
+from rest_framework.exceptions import PermissionDenied
+
+logger = logging.getLogger('awx.main.tasks.system')
+
+__all__ = [
+    'RunJob',
```
AlanCoding: wait a minute, `RunJob` isn't in this file here. Maybe this is just justification that the utility of the `__all__` doesn't really apply here, and I'd be fine to delete the whole thing.

awx/main/tasks/system.py
```diff
@@ -0,0 +1,915 @@
...
+    'handle_work_error',
+    'handle_work_success',
+    'apply_cluster_membership_policies',
+    'update_inventory_computed_fields',
+    'update_host_smart_inventory_memberships',
+    'send_notifications',
+    'purge_old_stdout_files',
+]
+
+HIDDEN_PASSWORD = '**********'
```
AlanCoding: This variable `HIDDEN_PASSWORD` isn't used anywhere in this file. Obviously we should not keep it.

But maybe this hints that we need stricter linter rules to guide a transition like this. We are using `black` as a replacement of these rules, but clearly some things are being missed. I would like to see some experimenting with resetting linter rules, and you could just use that to discover more rules we should put back in.

I mean, we whitelist rules right now:

https://github.com/ansible/awx/blob/d33a0d5ddeefc534e5c70f2bc95746479002cd17/tox.ini#L16

If you delete these, and re-run the linter, that should tell you more things. If anything it tells you is genuinely good feedback, then we should consider adding those rules to the allow list.

AlanCoding:
I looked further into the flake8 status of things, running with my preference of rules, I got this:

```
$ flake8 awx/main/tasks/
awx/main/tasks/jobs.py:65:1: F811 redefinition of unused 'safe_dump' from line 42
awx/main/tasks/jobs.py:78:1: F811 redefinition of unused 'logger' from line 2
```

Before you merge, I would like to address these two items. Separately, I will look into adding the rule. I realize that it's not appropriate to add the rule here, because it may introduce failures from other files.

AlanCoding:
I put up https://github.com/ansible/awx/pull/11544 about that.

AlanCoding (approved):
I feel great about the direction this is going in, thanks for taking care of it!

jbradberry:
A huge number of line changes in this PR are unnecessary, and could have been dealt with by doing imports from the new submodules into `main/tasks/__init__.py`.

PR #11530, opened 2022-01-13
Don't expose serviceAccount token on default pod spec

tchellomello:
##### SUMMARY
We don't need to expose the `serviceAccount` token for the `default` SA  in containers group

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
devel
```

##### ADDITIONAL INFORMATION

image

So even in a special case where the `spec` override does not include the `serviceAccount`, we can em `deepmerge()` and get it as well

```yaml
>>> pod_spec_override
{'apiVersion': 'v1',
 'kind': 'Pod',
 'metadata': {'labels': {'testing': 'mmello'}, 'namespace': 'ansible-devel'},
 'spec': {'containers': [''{'args': ['''ansible-runner',
                                   'worker',
                                   '--private-data-dir=/runner'],
                          'image': 'quay.io/ansible/awx-ee:latest',
                          'name': 'worker',
                          'volumeMounts': [''{'mountPath': '/shared_data',
                                            'name': 'awx-devel-test-claim'}]}],
          'volumes': [''{'name': 'awx-devel-test-claim',
                       'persistentVolumeClaim': {'claimName': 'awx-devel-test-claim'}}]}}
>>> pod_spec
{'apiVersion': 'v1',
 'kind': 'Pod',
 'metadata': {'labels': {'testing': 'mmello'}, 'namespace': 'ansible-devel'},
 'spec': {'automountServiceAccountToken': False,
          'containers': [''{'args': ['''ansible-runner',
                                   'worker',
                                   '--private-data-dir=/runner'],
                          'image': 'quay.io/ansible/awx-ee:latest',
                          'name': 'worker',
                          'volumeMounts': [''{'mountPath': '/shared_data',
                                            'name': 'awx-devel-test-claim'}]}],
          'serviceAccountName': 'default',
          'volumes': [''{'name': 'awx-devel-test-claim',
                       'persistentVolumeClaim': {'claimName': 'awx-devel-test-claim'}}]}}

```

```yaml
kubectl get  pod -n ansible-devel automation-job-mmello2-102-54tdr -o yaml                                                                                 23:43:12
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: "2022-01-13T04:42:50Z"
  generateName: automation-job-mmello2-102-
  labels:
    ansible-awx: 14676f40-a717-4322-bd4c-0874d2157566
    ansible-awx-job-id: "102"
    testing: mmello
  name: automation-job-mmello2-102-54tdr
  namespace: ansible-devel
  resourceVersion: "69985425"
  uid: b11fafb8-18f6-43ff-8b82-52b07d457a51
spec:
  automountServiceAccountToken: false   <=== here
  containers:
  - args:
    - ansible-runner
    - worker
    - --private-data-dir=/runner
    image: quay.io/ansible/awx-ee:latest
    imagePullPolicy: Always
    name: worker
    resources: {}
    stdin: true
    stdinOnce: true
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    volumeMounts:
    - mountPath: /shared_data
      name: awx-devel-test-claim
  dnsPolicy: ClusterFirst
  enableServiceLinks: true
  nodeName: p50
  priority: 0
  restartPolicy: Never
  securityContext: {}
  serviceAccount: default  
  serviceAccountName: default   <==== here
  tolerations:
  - effect: NoExecute
    key: node.kubernetes.io/not-ready
    operator: Exists
    tolerationSeconds: 300
  - effect: NoExecute
    key: node.kubernetes.io/unreachable
    operator: Exists
    tolerationSeconds: 300
  volumes:
  - name: awx-devel-test-claim
    persistentVolumeClaim:
      claimName: awx-devel-test-claim
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2022-01-13T04:42:50Z"
    status: "True"
    type: Initialized
  - lastProbeTime: null
    lastTransitionTime: "2022-01-13T04:42:54Z"
    status: "True"
    type: Ready
  - lastProbeTime: null
    lastTransitionTime: "2022-01-13T04:42:54Z"
    status: "True"
    type: ContainersReady
  - lastProbeTime: null
    lastTransitionTime: "2022-01-13T04:42:50Z"
    status: "True"
    type: PodScheduled
  containerStatuses:
  - containerID: cri-o://45306576c7cb4275a81a559f0c56e0c3233201db07793830a4156e9d01a8a06d
    image: quay.io/ansible/awx-ee:latest
    imageID: quay.io/ansible/awx-ee@sha256:05d2ddbb53a1cbfe12eb655b0c69c0bc72a712ec35458c30651d0d2582078e30
    lastState: {}
    name: worker
    ready: true
    restartCount: 0
    started: true
    state:
      running:
        startedAt: "2022-01-13T04:42:53Z"
  hostIP: 192.168.111.50
  phase: Running
  podIP: 10.233.67.184
  podIPs:
  - ip: 10.233.67.184
  qosClass: BestEffort
  startTime: "2022-01-13T04:42:50Z"
```

```

```

### DIFF
--- a/awx/main/utils/execution_environments.py
+++ b/awx/main/utils/execution_environments.py
@@ -31,6 +31,8 @@ def get_default_pod_spec():
         "kind": "Pod",
         "metadata": {"namespace": settings.AWX_CONTAINER_GROUP_DEFAULT_NAMESPACE},
         "spec": {
+            "serviceAccountName": "default",
+            "automountServiceAccountToken": False,
             "containers": [
                 {
                     "image": ee.image,


shanemcd:
It looks like the awx-operator CI checks are failing. I noticed this in the logs:

```
Error creating pod: v1.Pod.Spec: v1.PodSpec.AutomountServiceAccountToken: ReadBool: expect t or f, but found \", error found in #10 byte of ...
```

I'm also not sure how I feel about adding `automountServiceAccountToken` to the default pod spec. What was the reasoning for that? It's possible that folks have created role bindings against that user, and are using the `k8s` module to orchestrate things inside the EE... I know I have done that.

tchellomello:
@shanemcd thanks for pointing the CI error out. I hear you and I myself also have created a dedicated RBAC to allow playbooks to interact with the k8s cluster itself.  Tests are green now!

The idea for this patch is to provide the **least** possible access, as the default `serviceAccount` provides access read-only access to some objects which we could avoid out of the box. Since we don't also specify the `serviceAccountName` in the pod spec, that means we will use the default `serviceAccount` for that.  If the user needs to extend the RBAC to allow more granular permissions on the k8s cluster itself, it should be using a dedicated SA and not extending the default SA, as it most likely will expose access to any other pod running on that namespace under the very same SA. 

Having said this, the PR above we can knock this out as we determine the SA is default and we don't expose credentials by default. Any user that wants to modify it, still is possible, by creating their own SA and RBAC as usual. 

Please let me know if this provides more context about it. 

@Zokormazo @rooftopcellist  any thoughts?

shanemcd:
> the default serviceAccount provides access read-only access to some objects which we could avoid out of the box.

What kinds of things can it do that you're worried about? I didnt think it had access to much of anything. The [docs say](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#service-account-permissions):

> Default RBAC policies grant scoped permissions to control-plane components, nodes, and controllers, but grant no permissions to service accounts outside the kube-system namespace (beyond discovery permissions given to all authenticated users).

I would like to keep the default pod spec as minimal as possible. I dont see a reason to go against the defaults here. 

Also, you dont have to create a different service account, you can create role bindings against the default service account or even change the default role binding.

tchellomello:
@shanemcd the idea was to make the pod spec as minimalist and secure as possible by default.  For example, assuming an RBAC gets applied to the `serviceaccount:default:default` then the launched pod will inherit that RBAC. In other words, for AWX users that want their pods to have access to their k8s API, they would need to extend the pod spec and enable it. 

```
kubectl auth can-i --list --as=system:serviceaccount:default:default
Resources                                       Non-Resource URLs                     Resource Names   Verbs
selfsubjectaccessreviews.authorization.k8s.io   []                                    []               [create]
selfsubjectrulesreviews.authorization.k8s.io    []                                    []               [create]
                                                [/.well-known/openid-configuration]   []               [get]
                                                [/api/*]                              []               [get]
                                                [/api]                                []               [get]
                                                [/apis/*]                             []               [get]
                                                [/apis]                               []               [get]
                                                [/healthz]                            []               [get]
                                                [/healthz]                            []               [get]
                                                [/livez]                              []               [get]
                                                [/livez]                              []               [get]
                                                [/openapi/*]                          []               [get]
                                                [/openapi]                            []               [get]
                                                [/openid/v1/jwks]                     []               [get]
                                                [/readyz]                             []               [get]
                                                [/readyz]                             []               [get]
                                                [/version/]                           []               [get]
                                                [/version/]                           []               [get]
                                                [/version]                            []               [get]
                                                [/version]                            []               [get]
```

Another alternative would be to create a dedicated SA to avoid undesired and mistakenly RBACs naively attached to the default SA. 

At the same time, I understand this introduces a new step for the AWX admin (like anything security-wise).  Assuming that an RBAC was intentionally and purposely created to the `serviceaccount:default`, then asking the user to modify the pod spec to expose the token is an extra step.  Therefore, we can close this for now and revisit it later if needed.

shanemcd (approved):
I think I've come around to the idea.

PR #11546, opened 2022-01-14
Remove the Instance.objects.active_count() method

jbradberry:
##### SUMMARY
Literally nothing uses it.  The similar Host.objects.active_count()
method seems to be what is actually important for licensing.

This came up in the initial modeling work to support the mesh visualizer.

related #11431

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/managers.py
+++ b/awx/main/managers.py
@@ -186,10 +186,6 @@ class InstanceManager(models.Manager):
         else:
             return (False, self.me())
 
-    def active_count(self):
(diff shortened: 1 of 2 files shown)


AlanCoding:
If you like this patch, you might also like https://github.com/ansible/awx/pull/11503

AlanCoding (approved):

PR #11548, opened 2022-01-15
Revert "Make awx-python script available in k8s app images"

shanemcd:
See comment on https://github.com/ansible/awx-operator/issues/689


rooftopcellist:
I just built an image from the awx devel branch with that commit reverted and deployed it with the latest awx-operator code and sure enough there were no hiccups this time.  @kdelee and I must have been hitting something else.  Sorry for the confusion.

rooftopcellist (approved):

PR #11551, opened 2022-01-18
[WIP] Let container group jobs stay in Pending, allowing for pods to queue for resources in k8s

kdelee:
##### SUMMARY

Rely on the receptor status that already reflects when pod was successfully created, but the containers in it have not yet started. This allows us to better use resource requests to make sure workloads we schedule don't disrupt the stability of the AWX control plane which also runs in the same namespace on kubernetes.

I'm open to suggestions, ideas and discussion about this implementation, and this needs some good corner case testing regarding failure and job reaping scenarios.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
devel
```

##### ADDITIONAL INFORMATION

Currently, we don't by default make any resource requests in the pod spec. This has facilitated the generally successful behavior we currently have, which is to immediately put container group jobs into running after doing the work submit to receptor, providing a pretty short 5 min as a timeout to fail the job if the job's worker container doesn't actually get running in less than 5 min. This works whenever the pod can scheduled fast. Given that the podspec we've generally been using has no resource requests, the should get scheduled fast up to any limit on how many pods can be running on a node or in a namespace, which is generally much higher than is actually reasonable to assume we could run jobs and successfully deal with the output.

For example, if an automation job takes just 250m cpu (1/4 a CPU), by the time I have 80 jobs running on my 5 worker node cluster with 4vCPUs, my cluster CPUs are for all intents and purposes fully commited, and that's just with the jobs, not the control plane. To cope, OCP/k8s can take advantage of the fact that CPU is "compressible", e.g. throttle-able , but this is bad news especially for the control plane, because job event processing is very CPU intensive, and if we fall behind on job event processing, that can spin Redis into an OOM state if its queue of unprocessed events grows.
 
All this is a lot to say, that the way things are, using ContainerGroups with no resource requests is a foot gun, where its easy to start enough jobs that can cause the whole deployment to go sideways, or if you DO use resource requests and jobs do go into pending in OCP/k8s as they queue for resources, AWX can cascade fail jobs that are pending too long. So its a lose-lose situation.

### DIFF
--- a/awx/main/dispatch/reaper.py
+++ b/awx/main/dispatch/reaper.py
@@ -5,7 +5,7 @@ from django.db.models import Q
 from django.utils.timezone import now as tz_now
 from django.contrib.contenttypes.models import ContentType
 
-from awx.main.models import Instance, UnifiedJob, WorkflowJob, InstanceGroup
+from awx.main.models import Instance, UnifiedJob, WorkflowJob
 from awx.main.tasks.receptor import receptor_work_status, RECEPTOR_AWX_STATE_MAP
 
 logger = logging.getLogger('awx.main.dispatch')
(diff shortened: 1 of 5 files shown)


awx/main/dispatch/reaper.py
```diff
@@ -14,6 +15,17 @@ def reap_job(j, status):
...
         # just in case, don't reap jobs that aren't running
         return
+    if j.instance_group.is_container_group:
+        status, detail = receptor_work_status(j.work_unit_id)
+        if RECEPTOR_AWX_STATE_MAP[status] != j.status:
+            # Sometimes the job has exited already, but we had not had chance to update the
+            # job status with one of these terminal states.
+            # Update the status and return early.
+            # Otherwise the job truly is gone but still marked as running, in which case it should
+            # rightfully be reaped
```
kdelee, line 28: This workaround fixed some jobs that were getting reaped when they actually were healthy. But I do wonder if this is necessary because I'm missing something earlier in the lifecycle of the job.

kdelee:
This is a can of worms, and I'm not sure following the pattern of resubmitting jobs w/o any backoff is wise in any case.
Perhaps we could introduce ANOTHER job state, but I'm not sure I want to go that route either.

Looking at https://github.com/ansible/awx/pull/11559 as an alternative

PR #11559, opened 2022-01-18
Add resource requests to default podspec

kdelee:
Extend the timeout, assuming that we want to let the kubernetes scheduler
start containers when it wants to start them. This allows us to make
resource requests knowing that when some jobs queue up waiting for
resources, they will not get reaped in as short of a
timeout.

this is an alternative to https://github.com/ansible/awx/pull/11551
How is this different:
  * doesn't replicate what is already a bad pattern w/ the ResourceQuota workaround
  * don't get pretty messages in awx UI, it does show "running" the whole time

How does this achieve similar goals:
* DO get expected behavior that jobs queue up in kubernetes and complete eventually

Problem with this approach:
Reveals that image pull errors are not well reflected in receptor https://github.com/ansible/receptor/issues/521, for which I'm working on https://github.com/ansible/receptor/pull/522

### DIFF
--- a/awx/main/tasks/receptor.py
+++ b/awx/main/tasks/receptor.py
@@ -381,7 +381,7 @@ class AWXReceptorJob:
                             self.task.instance.result_traceback = detail
                             self.task.instance.save(update_fields=['result_traceback'])
                         else:
-                            logger.warn(f'No result details or output from {self.task.instance.log_format}, status:\n{unit_status}')
+                            logger.warn(f'No result details or output from {self.task.instance.log_format}, status:\n{state_name}')
                     except Exception:
                         raise RuntimeError(detail)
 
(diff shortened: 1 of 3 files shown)


kdelee:
It would be nice to have this,because if we use the setting from https://github.com/ansible/awx/pull/11395 for items in the actual container entry, we override everything about the `worker` container. That said, as is pointed out in https://github.com/ansible/awx/pull/10569/files -- the image and args always get ignored. 

Also, `DEFAULT_EXECUTION_QUEUE_POD_SPEC_OVERRIDE` only sets the resource requests for the `default` contianer group pod spec, and I think giving the baseline request is a good pattern to establish in the default podspec.

Alternatives I can think of are, yet another setting "DEFAULT_CONTAINER_GROUP_CONTAINER_RESOURCES", or work on the `deepmerge` function that will merge the list items in `containers` based on if they share names.

shanemcd:
I think I am -1 on this idea, it's likely to cause more problems than it is to help. I would also encourage the usage of a `LimitRange` rather than putting this in the pod spec itself.

shanemcd:
I spoke too soon. For whatever reason I read this and thought you were setting limits.

shanemcd (approved):

PR #11562, opened 2022-01-18
Avoid duplicated entries when calling create_preload_data

tchellomello:
##### SUMMARY
When deleting the `Default Organization`, re-running the `create_preload_data` causes the database integrity issue as some `read-only` resources will get be duplicated in the database

#### Steps to reproduce
1. Delete the `Default Org from the weUI
2. Run the `create_preload_data` script 3 times. 
```bash
```
 
image

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
devel
```

##### Testing after patching the code
1. Cleaning database
```bash
$ awx-manage shell_plus
JobTemplate.objects.all().delete()
Credential.objects.all().delete()
Project.objects.all().delete()
Inventory.objects.all().delete()
Organization.objects.all().delete()
```
2. Run multiple times the `awx-manage create_preload_data` 

3. Things should look good :)

### DIFF
--- a/awx/main/management/commands/create_preload_data.py
+++ b/awx/main/management/commands/create_preload_data.py
@@ -23,44 +23,50 @@ class Command(BaseCommand):
         with impersonate(superuser):
             with disable_computed_fields():
                 if not Organization.objects.exists():
-                    o = Organization.objects.create(name='Default')
+                    o, _ = Organization.objects.get_or_create(name='Default')
 
-                    p = Project(
+                    p, _ = Project.objects.get_or_create(
                         name='Demo Project',
                         scm_type='git',
                         scm_url='https://github.com/ansible/ansible-tower-samples',
                         scm_update_on_launch=True,
                         scm_update_cache_timeout=0,
-                        organization=o,
                     )
-                    p.save(skip_update=True)
+                    p.organization = o
+                    p.save()
 
                     ssh_type = CredentialType.objects.filter(namespace='ssh').first()
-                    c = Credential.objects.create(
+                    c, _ = Credential.objects.get_or_create(
                         credential_type=ssh_type, name='Demo Credential', inputs={'username': superuser.username}, created_by=superuser
                     )
 
                     c.admin_role.members.add(superuser)
 
-                    public_galaxy_credential = Credential(
+                    public_galaxy_credential, _ = Credential.objects.get_or_create(
                         name='Ansible Galaxy',
                         managed=True,
                         credential_type=CredentialType.objects.get(kind='galaxy'),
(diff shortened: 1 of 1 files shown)


awx/main/management/commands/create_preload_data.py
```diff
@@ -23,44 +23,50 @@ def handle(self, *args, **kwargs):
...
                         name='Ansible Galaxy',
                         managed=True,
                         credential_type=CredentialType.objects.get(kind='galaxy'),
                         inputs={'url': 'https://galaxy.ansible.com/'},
                     )
-                    public_galaxy_credential.save()
                     o.galaxy_credentials.add(public_galaxy_credential)
 
-                    i = Inventory.objects.create(name='Demo Inventory', organization=o, created_by=superuser)
+                    i, _ = Inventory.objects.get_or_create(name='Demo Inventory', organization=o, created_by=superuser)
```
shanemcd, line 53: Do we not have the same problem here as we did with the project? Will this create a duplicate inventory associated with the new org?

tchellomello: I was not able to get the `Inventory` duplicated as the other resources

shanemcd (approved):

awx/main/management/commands/create_preload_data.py
shanemcd: Looked into it - that's because the inventory gets deleted when you delete the org.

awx/main/management/commands/create_preload_data.py
```diff
@@ -23,44 +23,50 @@ def handle(self, *args, **kwargs):
...
                 if not Organization.objects.exists():
-                    o = Organization.objects.create(name='Default')
+                    o, _ = Organization.objects.get_or_create(name='Default')
 
-                    p = Project(
+                    p, _ = Project.objects.get_or_create(
                         name='Demo Project',
                         scm_type='git',
                         scm_url='https://github.com/ansible/ansible-tower-samples',
                         scm_update_on_launch=True,
```
AlanCoding, line 32: Personally, I frequently go into the UI and disable `scm_update_on_launch` for the demo project. If I read this right, that would cause subsequent runs of this script to error. Because of my change, a project with these attributes won't exist, so between get & create, it will create, but a project with the (name, organization) identity tuple does exist, thus, probably an error.

So I would prefer to eventually move to the documented solution that the non-identity fields get passed in the `defaults` kwarg. For an example

https://github.com/ansible/awx/blob/2afa406b7f16971ea83150c326f8e04717b5c14d/awx/main/management/commands/register_default_execution_environments.py#L77-L82

PR #11571, opened 2022-01-20
Added new class for  Ansible Runner Callbacks

amolgautam25:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 16.0.1.dev3184+g2f0e80f113
```

##### ADDITIONAL INFORMATION

```

```

Things to do: 

- [x] Figure out a way to update the ```instance``` from db in the new class and propagate it back to the BaseTask
- [x] Fix unit tests

### DIFF
--- a/awx/main/tasks/callback.py
+++ b/awx/main/tasks/callback.py
@@ -1,0 +1,257 @@
+import json
+import time
+import logging
+from collections import deque
+import os
+import stat
+
+# Django
+from django.utils.timezone import now
+from django.conf import settings
+from django_guid.middleware import GuidMiddleware
+
+# AWX
+from awx.main.redact import UriCleaner
+from awx.main.constants import MINIMAL_EVENTS
+from awx.main.utils.update_model import update_model
+from awx.main.queue import CallbackQueueDispatcher
+
+logger = logging.getLogger('awx.main.tasks.callback')
+
+
+class RunnerCallback:
+    event_data_key = 'job_id'
+
+    def __init__(self, model=None):
+        self.parent_workflow_job_id = None
+        self.host_map = {}
+        self.guid = GuidMiddleware.get_guid()
+        self.job_created = None
+        self.recent_event_timings = deque(maxlen=settings.MAX_WEBSOCKET_EVENT_RATE)
+        self.dispatcher = CallbackQueueDispatcher()
+        self.safe_env = {}
+        self.event_ct = 0
+        self.model = model
+
+    def update_model(self, pk, _attempt=0, **updates):
+        return update_model(self.model, pk, _attempt=0, **updates)
+
+    def event_handler(self, event_data):
+        #
+        # ⚠️  D-D-D-DANGER ZONE ⚠️
+        # This method is called once for *every event* emitted by Ansible
+        # Runner as a playbook runs.  That means that changes to the code in
+        # this method are _very_ likely to introduce performance regressions.
+        #
+        # Even if this function is made on average .05s slower, it can have
+        # devastating performance implications for playbooks that emit
+        # tens or hundreds of thousands of events.
+        #
+        # Proceed with caution!
+        #
+        """
+        Ansible runner puts a parent_uuid on each event, no matter what the type.
+        AWX only saves the parent_uuid if the event is for a Job.
+        """
+        # cache end_line locally for RunInventoryUpdate tasks
+        # which generate job events from two 'streams':
+        # ansible-inventory and the awx.main.commands.inventory_import
+        # logger
+
+        if event_data.get(self.event_data_key, None):
+            if self.event_data_key != 'job_id':
+                event_data.pop('parent_uuid', None)
+        if self.parent_workflow_job_id:
+            event_data['workflow_job_id'] = self.parent_workflow_job_id
+        event_data['job_created'] = self.job_created
+        if self.host_map:
+            host = event_data.get('event_data', {}).get('host', '').strip()
+            if host:
+                event_data['host_name'] = host
+                if host in self.host_map:
+                    event_data['host_id'] = self.host_map[host]
+            else:
+                event_data['host_name'] = ''
+                event_data['host_id'] = ''
+            if event_data.get('event') == 'playbook_on_stats':
+                event_data['host_map'] = self.host_map
+
+        if isinstance(self, RunnerCallbackForProjectUpdate):
+            # need a better way to have this check.
+            # it's common for Ansible's SCM modules to print
+            # error messages on failure that contain the plaintext
+            # basic auth credentials (username + password)
+            # it's also common for the nested event data itself (['res']['...'])
+            # to contain unredacted text on failure
+            # this is a _little_ expensive to filter
+            # with regex, but project updates don't have many events,
+            # so it *should* have a negligible performance impact
+            task = event_data.get('event_data', {}).get('task_action')
+            try:
+                if task in ('git', 'svn'):
+                    event_data_json = json.dumps(event_data)
+                    event_data_json = UriCleaner.remove_sensitive(event_data_json)
+                    event_data = json.loads(event_data_json)
+            except json.JSONDecodeError:
+                pass
+
+        if 'event_data' in event_data:
+            event_data['event_data']['guid'] = self.guid
+
+        # To prevent overwhelming the broadcast queue, skip some websocket messages
+        if self.recent_event_timings:
+            cpu_time = time.time()
+            first_window_time = self.recent_event_timings[0]
+            last_window_time = self.recent_event_timings[-1]
+
+            if event_data.get('event') in MINIMAL_EVENTS:
+                should_emit = True  # always send some types like playbook_on_stats
+            elif event_data.get('stdout') == '' and event_data['start_line'] == event_data['end_line']:
+                should_emit = False  # exclude events with no output
+            else:
+                should_emit = any(
+                    [
+                        # if 30the most recent websocket message was sent over 1 second ago
+                        cpu_time - first_window_time > 1.0,
+                        # if the very last websocket message came in over 1/30 seconds ago
+                        self.recent_event_timings.maxlen * (cpu_time - last_window_time) > 1.0,
+                        # if the queue is not yet full
+                        len(self.recent_event_timings) != self.recent_event_timings.maxlen,
+                    ]
+                )
+
+            if should_emit:
+                self.recent_event_timings.append(cpu_time)
+            else:
+                event_data.setdefault('event_data', {})
+                event_data['skip_websocket_message'] = True
+
+        elif self.recent_event_timings.maxlen:
+            self.recent_event_timings.append(time.time())
+
+        event_data.setdefault(self.event_data_key, self.instance.id)
+        self.dispatcher.dispatch(event_data)
+        self.event_ct += 1
+
+        '''
+        Handle artifacts
+        '''
+        if event_data.get('event_data', {}).get('artifact_data', {}):
+            self.instance.artifacts = event_data['event_data']['artifact_data']
+            self.instance.save(update_fields=['artifacts'])
+
+        return False
+
+    def cancel_callback(self):
+        """
+        Ansible runner callback to tell the job when/if it is canceled
+        """
+        unified_job_id = self.instance.pk
+        self.instance.refresh_from_db()
+        if not self.instance:
+            logger.error('unified job {} was deleted while running, canceling'.format(unified_job_id))
+            return True
+        if self.instance.cancel_flag or self.instance.status == 'canceled':
+            cancel_wait = (now() - self.instance.modified).seconds if self.instance.modified else 0
+            if cancel_wait > 5:
+                logger.warn('Request to cancel {} took {} seconds to complete.'.format(self.instance.log_format, cancel_wait))
+            return True
+        return False
+
+    def finished_callback(self, runner_obj):
+        """
+        Ansible runner callback triggered on finished run
+        """
+        event_data = {
+            'event': 'EOF',
+            'final_counter': self.event_ct,
+            'guid': self.guid,
+        }
+        event_data.setdefault(self.event_data_key, self.instance.id)
+        self.dispatcher.dispatch(event_data)
+
+    def status_handler(self, status_data, runner_config):
+        """
+        Ansible runner callback triggered on status transition
+        """
+        if status_data['status'] == 'starting':
+            job_env = dict(runner_config.env)
+            '''
+            Take the safe environment variables and overwrite
+            '''
+            for k, v in self.safe_env.items():
+                if k in job_env:
+                    job_env[k] = v
+            from awx.main.signals import disable_activity_stream  # Circular import
+
+            with disable_activity_stream():
+                self.instance = self.update_model(self.instance.pk, job_args=json.dumps(runner_config.command), job_cwd=runner_config.cwd, job_env=job_env)
+        elif status_data['status'] == 'failed':
+            # For encrypted ssh_key_data, ansible-runner worker will open and write the
+            # ssh_key_data to a named pipe. Then, once the podman container starts, ssh-agent will
+            # read from this named pipe so that the key can be used in ansible-playbook.
+            # Once the podman container exits, the named pipe is deleted.
+            # However, if the podman container fails to start in the first place, e.g. the image
+            # name is incorrect, then this pipe is not cleaned up. Eventually ansible-runner
+            # processor will attempt to write artifacts to the private data dir via unstream_dir, requiring
+            # that it open this named pipe. This leads to a hang. Thus, before any artifacts
+            # are written by the processor, it's important to remove this ssh_key_data pipe.
+            private_data_dir = self.instance.job_env.get('AWX_PRIVATE_DATA_DIR', None)
+            if private_data_dir:
+                key_data_file = os.path.join(private_data_dir, 'artifacts', str(self.instance.id), 'ssh_key_data')
+                if os.path.exists(key_data_file) and stat.S_ISFIFO(os.stat(key_data_file).st_mode):
+                    os.remove(key_data_file)
+        elif status_data['status'] == 'error':
+            result_traceback = status_data.get('result_traceback', None)
+            if result_traceback:
+                from awx.main.signals import disable_activity_stream  # Circular import
+
+                with disable_activity_stream():
+                    self.instance = self.update_model(self.instance.pk, result_traceback=result_traceback)
+
+
+class RunnerCallbackForProjectUpdate(RunnerCallback):
+
+    event_data_key = 'project_update_id'
+
+    def __init__(self, *args, **kwargs):
+        super(RunnerCallbackForProjectUpdate, self).__init__(*args, **kwargs)
+        self.playbook_new_revision = None
+        self.host_map = {}
+
+    def event_handler(self, event_data):
+        super_return_value = super(RunnerCallbackForProjectUpdate, self).event_handler(event_data)
+        returned_data = event_data.get('event_data', {})
+        if returned_data.get('task_action', '') == 'set_fact':
+            returned_facts = returned_data.get('res', {}).get('ansible_facts', {})
+            if 'scm_version' in returned_facts:
+                self.playbook_new_revision = returned_facts['scm_version']
+        return super_return_value
+
+
+class RunnerCallbackForInventoryUpdate(RunnerCallback):
+
+    event_data_key = 'inventory_update_id'
+
+    def __init__(self, *args, **kwargs):
+        super(RunnerCallbackForInventoryUpdate, self).__init__(*args, **kwargs)
+        self.end_line = 0
+
+    def event_handler(self, event_data):
+        self.end_line = event_data['end_line']
+
+        return super(RunnerCallbackForInventoryUpdate, self).event_handler(event_data)
+
+
+class RunnerCallbackForAdHocCommand(RunnerCallback):
+
+    event_data_key = 'ad_hoc_command_id'
+
+    def __init__(self, *args, **kwargs):
+        super(RunnerCallbackForAdHocCommand, self).__init__(*args, **kwargs)
+        self.host_map = {}
+
+
+class RunnerCallbackForSystemJob(RunnerCallback):
+
+    event_data_key = 'system_job_id'
--- a/awx/main/tasks/jobs.py
+++ b/awx/main/tasks/jobs.py
@@ -1,5 +1,5 @@
 # Python
-from collections import deque, OrderedDict
+from collections import OrderedDict
 from distutils.dir_util import copy_tree
 import errno
 import functools
@@ -19,10 +19,8 @@ from uuid import uuid4
 
 
 # Django
-from django_guid.middleware import GuidMiddleware
 from django.conf import settings
-from django.db import transaction, DatabaseError
-from django.utils.timezone import now
+from django.db import transaction
 
 
 # Runner
@@ -40,11 +38,9 @@ from awx.main.dispatch import get_local_queuename
 from awx.main.constants import (
     PRIVILEGE_ESCALATION_METHODS,
     STANDARD_INVENTORY_UPDATE_ENV,
-    MINIMAL_EVENTS,
     JOB_FOLDER_PREFIX,
     MAX_ISOLATED_PATH_COLON_DELIMITER,
 )
-from awx.main.redact import UriCleaner
 from awx.main.models import (
     Instance,
     Inventory,
@@ -61,7 +57,13 @@ from awx.main.models import (
     SystemJobEvent,
     build_safe_env,
 )
-from awx.main.queue import CallbackQueueDispatcher
+from awx.main.tasks.callback import (
+    RunnerCallback,
+    RunnerCallbackForAdHocCommand,
+    RunnerCallbackForInventoryUpdate,
+    RunnerCallbackForProjectUpdate,
+    RunnerCallbackForSystemJob,
+)
 from awx.main.tasks.receptor import AWXReceptorJob
 from awx.main.exceptions import AwxTaskError, PostRunError, ReceptorNodeNotFound
 from awx.main.utils.ansible import read_ansible_config
@@ -76,6 +78,7 @@ from awx.main.utils.common import (
 from awx.conf.license import get_license
 from awx.main.utils.handlers import SpecialInventoryHandler
 from awx.main.tasks.system import handle_success_and_failure_notifications, update_smart_memberships_for_inventory, update_inventory_computed_fields
+from awx.main.utils.update_model import update_model
 from rest_framework.exceptions import PermissionDenied
 from django.utils.translation import ugettext_lazy as _
 
@@ -105,46 +108,14 @@ class BaseTask(object):
     model = None
     event_model = None
     abstract = True
+    callback_class = RunnerCallback
 
     def __init__(self):
         self.cleanup_paths = []
-        self.parent_workflow_job_id = None
-        self.host_map = {}
-        self.guid = GuidMiddleware.get_guid()
-        self.job_created = None
-        self.recent_event_timings = deque(maxlen=settings.MAX_WEBSOCKET_EVENT_RATE)
+        self.runner_callback = self.callback_class(model=self.model)
 
     def update_model(self, pk, _attempt=0, **updates):
-        """Reload the model instance from the database and update the
-        given fields.
-        """
-        try:
-            with transaction.atomic():
-                # Retrieve the model instance.
-                instance = self.model.objects.get(pk=pk)
-
-                # Update the appropriate fields and save the model
-                # instance, then return the new instance.
-                if updates:
-                    update_fields = ['modified']
-                    for field, value in updates.items():
-                        setattr(instance, field, value)
-                        update_fields.append(field)
-                        if field == 'status':
-                            update_fields.append('failed')
-                    instance.save(update_fields=update_fields)
-                return instance
-        except DatabaseError as e:
-            # Log out the error to the debug logger.
-            logger.debug('Database error updating %s, retrying in 5 ' 'seconds (retry #%d): %s', self.model._meta.object_name, _attempt + 1, e)
-
-            # Attempt to retry the update, assuming we haven't already
-            # tried too many times.
-            if _attempt < 5:
-                time.sleep(5)
-                return self.update_model(pk, _attempt=_attempt + 1, **updates)
-            else:
-                logger.error('Failed to update %s after %d retries.', self.model._meta.object_name, _attempt)
+        return update_model(self.model, pk, _attempt=0, **updates)
 
     def get_path_to(self, *args):
         """
@@ -350,7 +321,7 @@ class BaseTask(object):
         script_data = instance.inventory.get_script_data(**script_params)
         # maintain a list of host_name --> host_id
         # so we can associate emitted events to Host objects
-        self.host_map = {hostname: hv.pop('remote_tower_id', '') for hostname, hv in script_data.get('_meta', {}).get('hostvars', {}).items()}
+        self.runner_callback.host_map = {hostname: hv.pop('remote_tower_id', '') for hostname, hv in script_data.get('_meta', {}).get('hostvars', {}).items()}
         json_data = json.dumps(script_data)
         path = os.path.join(private_data_dir, 'inventory')
         fn = os.path.join(path, 'hosts')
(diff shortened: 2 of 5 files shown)


awx/main/tasks/jobs.py
```diff
@@ -76,6 +76,177 @@
...
+        self.instance = self.update_model(unified_job_id)
+        if not self.instance:
+            logger.error('unified job {} was deleted while running, canceling'.format(unified_job_id))
+            return True
+        if self.instance.cancel_flag or self.instance.status == 'canceled':
+            cancel_wait = (now() - self.instance.modified).seconds if self.instance.modified else 0
+            if cancel_wait > 5:
+                logger.warn('Request to cancel {} took {} seconds to complete.'.format(self.instance.log_format, cancel_wait))
+            return True
+        return False
```
AlanCoding: With the other PR, this will be handled in a different class, but I think we will be free of conflicts if we just leave this method unchanged in this diff. We can work on an elegant combination of the 2 in another unit of work.

awx/main/tasks/jobs.py
```diff
@@ -76,6 +76,177 @@
 logger = logging.getLogger('awx.main.tasks.jobs')
 
 
+class RunnerCallback:
```
AlanCoding: I don't see this used anywhere?

You have references like `self.event_data_key` in here. In that, `self` is supposed to be a subclass of `BaseTask` and that's not true when you split it into a different class. So you can process these in an `__init__` method, and if it makes sense to you, save it right to `self` again. This clarifies the data flow between the task class and the callback class.

And obviously, you'll have to initialize it somewhere and replace the references to the methods.

amolgautam25: @AlanCoding , thank you.
I have tried to make the required changes.

awx/main/tasks/jobs.py
```diff
@@ -82,6 +82,190 @@
...
+    def __init__(self, BaseTaskObject):
+        self.event_data_key = BaseTaskObject.event_data_key
+        self.parent_workflow_job_id = BaseTaskObject.parent_workflow_job_id
+        self.job_created = BaseTaskObject.job_created
+        self.host_map = BaseTaskObject.host_map
+        self.guid = BaseTaskObject.guid
+        self.recent_event_timings = BaseTaskObject.recent_event_timings
+        self.instance = BaseTaskObject.instance
```
AlanCoding: style-wise, `BaseTaskObject` looks like it should be a class, I'd prefer a style like `task_object` so it's more obviously an instantiated object.

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,188 @@
...
+        """
+        Ansible runner puts a parent_uuid on each event, no matter what the type.
+        AWX only saves the parent_uuid if the event is for a Job.
+        """
+        # cache end_line locally for RunInventoryUpdate tasks
+        # which generate job events from two 'streams':
+        # ansible-inventory and the awx.main.commands.inventory_import
+        # logger
+        if isinstance(self, RunInventoryUpdate):
+            self.end_line = event_data['end_line']
```
amolgautam25: Right now there are variables such as ```end_line``` which belong to class ```BaseTask```. I am not sure if these variables are supposed to be moved to ```RunnerCallback``` class or left as is.

awx/main/tasks/receptor.py
```diff
@@ -405,15 +405,20 @@ def transmit(self, _socket):
             # Socket must be shutdown here, or the reader will hang forever.
             _socket.shutdown(socket.SHUT_WR)
 
+    def attach_runner_callback(self):
+        from awx.main.tasks.jobs import RunnerCallback
```
amolgautam25: To avoid circular imports.

awx/main/tasks/jobs.py
```diff
@@ -444,164 +442,8 @@ def final_run_hook(self, instance, status, private_data_dir, fact_modification_t
...
-                self.instance = self.update_model(self.instance.pk, job_args=json.dumps(runner_config.command), job_cwd=runner_config.cwd, job_env=job_env)
-        elif status_data['status'] == 'error':
-            result_traceback = status_data.get('result_traceback', None)
-            if result_traceback:
-                from awx.main.signals import disable_activity_stream  # Circular import
-
-                with disable_activity_stream():
-                    self.instance = self.update_model(self.instance.pk, result_traceback=result_traceback)
+    def attach_callback(self):
+        self.runner_callback = RunnerCallback()
```
amolgautam25: The two classes ```BaseTak``` and ```RunnerCallback``` are coupled here.

awx/main/tasks/callback.py
amolgautam25: https://github.com/ansible/awx/pull/11571/files#r793020460

awx/main/tasks/jobs.py
amolgautam25: If we look at this coupling we can easily do something like this :

--- change a value inside of a callback instance
example : self.runner_callback.some_variable = new_value

What I am stuck at is the following:
1) if something  changes inside the runner_callback ( which is kind of an inside class, how would that change be propagated to  outside class, in this case ```BaseTask``` . 
I can go ahead and implement something like this: https://stackoverflow.com/questions/2024566/how-to-access-outer-class-from-an-inner-class , but it seems to overcomplicate things. 

2) Where do I invoke this function ? There has to be a place where a runner_callback is attached to the BaseTask.

awx/main/tasks/jobs.py
```diff
@@ -107,12 +105,15 @@ class BaseTask(object):
     abstract = True
```
AlanCoding, line 110: Right here I would prefer to do

```python
    callback_class = RunnerCallback
```

Then in RunProjectUpdate, put in the respective callback class

awx/main/tasks/jobs.py
```diff
@@ -107,12 +105,15 @@ class BaseTask(object):
...
 
     def __init__(self):
+        self.runner_callback = RunnerCallback()
```
AlanCoding: Corresponding to prior comment, change this to 

```python
        self.runner_callback = self.callback_class()
```

awx/main/tasks/jobs.py
```diff
@@ -107,12 +105,15 @@ class BaseTask(object):
...
-        self.host_map = {}
-        self.guid = GuidMiddleware.get_guid()
-        self.job_created = None
-        self.recent_event_timings = deque(maxlen=settings.MAX_WEBSOCKET_EVENT_RATE)
+        self.runner_callback.parent_workflow_job_id = None
+        self.runner_callback.host_map = {}
+        self.runner_callback.guid = GuidMiddleware.get_guid()
+        self.runner_callabck.job_created = None
+        self.runner_callback.recent_event_timings = deque(maxlen=settings.MAX_WEBSOCKET_EVENT_RATE)
+        self.runner_callback.instance.id = self.instance.id
```
AlanCoding: could easily add

```python
     self.runner_callback.event_data_key = self.event_data_key
```

awx/main/tasks/jobs.py
```diff
@@ -1203,18 +1048,11 @@ class RunProjectUpdate(BaseTask):
 
     def __init__(self, *args, job_private_data_dir=None, **kwargs):
         super(RunProjectUpdate, self).__init__(*args, **kwargs)
-        self.playbook_new_revision = None
+        self.runner_callback = RunnerCallbackForProjectUpdate()
+        self.runner_callback.playbook_new_revision = None
```
AlanCoding: Setting `playbook_new_revision` to None would be better done in `RunnerCallbackForProjectUpdate.__init__`

awx/main/tasks/jobs.py
amolgautam25: I have removed the ```attach callback()``` function

awx/main/tasks/jobs.py
amolgautam25: I have made the required changes.

awx/main/tasks/jobs.py
amolgautam25: I have added that.

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,203 @@
...
+        """
+        Ansible runner puts a parent_uuid on each event, no matter what the type.
+        AWX only saves the parent_uuid if the event is for a Job.
+        """
+        # cache end_line locally for RunInventoryUpdate tasks
+        # which generate job events from two 'streams':
+        # ansible-inventory and the awx.main.commands.inventory_import
+        # logger
+        if isinstance(self, RunInventoryUpdate):
+            # Will removing this 'if'statement from here have any performance impact ?
```
amolgautam25: @AlanCoding In this new file this line ``` if isinstance(self, RunInventoryUpdate):``` would always return false. 
Would it be wise to just remove this line? 
Any thoughts / suggestion on this ?

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,203 @@
...
+                event_data['host_name'] = host
+                if host in self.host_map:
+                    event_data['host_id'] = self.host_map[host]
+            else:
+                event_data['host_name'] = ''
+                event_data['host_id'] = ''
+            if event_data.get('event') == 'playbook_on_stats':
+                event_data['host_map'] = self.host_map
+
+        if isinstance(self, RunProjectUpdate):
```
amolgautam25: @AlanCoding following up on https://github.com/ansible/awx/pull/11571/files#r795575348 , i can see a similar problem here. 
Can I do something like this : 
```python
 if isinstance(self,RunnercallbackForProjectUpdate)
```

This seems logical to me. I think it would work here.

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,203 @@
...
+            self.instance.save(update_fields=['artifacts'])
+
+        return False
+
+    def cancel_callback(self):
+        """
+        Ansible runner callback to tell the job when/if it is canceled
+        """
+        unified_job_id = self.instance.pk
+        self.instance = self.update_model(unified_job_id)
```
amolgautam25: ```self.instance and self.update_model``` are both not accessible here. 
I have a bit of idea how we can change the model from the database. ( ```refresh_from_db()``` ) , but i am not sure how we  can use that here  to update the ```instance```.

On top of that ```instance``` belongs to the ```BaseTask``` class. I have not yet removed the reference of that from callback class. 

@AlanCoding  any thoughts on this one ?

awx/main/tasks/jobs.py
amolgautam25: I have made the required change.

awx/main/tasks/callback.py
AlanCoding: Just like you introduced a class `RunnerCallbackForProjectUpdate`, you can cover this elegantly by introducing `RunnerCallbackForInventoryUpdate`, then move the logic there, and then delete it here.

You could keep it here and then change references in `jobs.py` to `self.runner_callback.end_line`, if that makes it easier to get to a state of functionality.

awx/main/tasks/callback.py
AlanCoding: Yes, I agree.

awx/main/tasks/callback.py
AlanCoding: Yes, the current cancel_callback is fundamentally dependent on having the instance available and accessing the database. Until that is refactored, `self.instance` will need to be kept on the callback class. To do that, you have to assign it in `__init__` which shouldn't be a problem since it receives the `BaseTask` derived object to begin with.

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,203 @@
...
+logger = logging.getLogger('awx.main.tasks.jobs')
+
+
+class RunnerCallback:
+    def __init__(self):
+        self.runner_callback.parent_workflow_job_id = None
+        self.runner_callback.host_map = {}
+        self.runner_callback.guid = GuidMiddleware.get_guid()
+        self.runner_callback.job_created = None
+        self.runner_callback.recent_event_timings = deque(maxlen=settings.MAX_WEBSOCKET_EVENT_RATE)
```
AlanCoding: Pattern-wise, `self.runner_callback.parent_workflow_job_id = None` --> `self.parent_workflow_job_id = None`

This is possibly the most important thing to clear up so that you can get running with testing this.

amolgautam25: I have made the changes.

awx/main/tasks/jobs.py
```diff
@@ -105,14 +103,12 @@ class BaseTask(object):
...
 
     def __init__(self):
+        self.runner_callback = self.callback_class()
         self.cleanup_paths = []
-        self.parent_workflow_job_id = None
-        self.host_map = {}
-        self.guid = GuidMiddleware.get_guid()
-        self.job_created = None
-        self.recent_event_timings = deque(maxlen=settings.MAX_WEBSOCKET_EVENT_RATE)
+        self.runner_callback.event_data_key = self.event_data_key
```
AlanCoding: As I search this file for `event_data_key`, I don't see any reason the task class should own it. I feel that it should be moved entirely to the callback class. So, I picture:

```python
class RunnerCallbackForProjectUpdate(RunnerCallback):
    event_data_key = 'project_update_id'
```

That means that you would have to introduce a new class for every job type that has an event_data_key defined. To me, that feels like the right solution. That will clean up a lot of the work in `__init__` methods here.

awx/main/tasks/jobs.py
```diff
@@ -723,16 +561,16 @@ def run(self, pk, **kwargs):
                 if not params[v]:
                     del params[v]
 
-            self.dispatcher = CallbackQueueDispatcher()
+            self.runner_callback.dispatcher = CallbackQueueDispatcher()
```
AlanCoding: I can't see any reason for `CallbackQueueDispatcher` to be in this file. I think it should be moved to the callback module. It should work in the `__init__` method.

awx/main/tasks/jobs.py
```diff
@@ -802,6 +640,9 @@ class RunJob(BaseTask):
     event_model = JobEvent
     event_data_key = 'job_id'
 
+    def __init__(self):
+        self.runner_callback.event_data_key = self.event_data_key
```
AlanCoding: I expect this will break inventory update runs, because you didn't defer to the `super` method. The proper way to override `__init__` in a subclass is that you call the super's method (like `super(self, RunJob).__init__(*args, **kwargs)`).

But if you take my advice from a prior comment, you shouldn't need this method at all.

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,206 @@
...
+                from awx.main.signals import disable_activity_stream  # Circular import
+
+                with disable_activity_stream():
+                    self.instance = self.update_model(self.instance.pk, result_traceback=result_traceback)
+
+
+class RunnerCallbackForProjectUpdate(RunnerCallback):
+    def __init__(self):
+        super(RunnerCallbackForProjectUpdate, self).__init__()
+        self.playbook_new_revision = None
```
AlanCoding, line 219: Perfect, this looks great :+1:

awx/main/tasks/receptor.py
amolgautam25: This is no longer relevant

awx/main/tasks/jobs.py
amolgautam25: I have made the required changes.

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,299 @@
...
+        """
+        Ansible runner puts a parent_uuid on each event, no matter what the type.
+        AWX only saves the parent_uuid if the event is for a Job.
+        """
+        # cache end_line locally for RunInventoryUpdate tasks
+        # which generate job events from two 'streams':
+        # ansible-inventory and the awx.main.commands.inventory_import
+        # logger
+        if isinstance(self, RunnerCallbackForInventoryUpdate):
+            self.end_line = event_data['end_line']
```
AlanCoding: It would be quite elegant to keep this line, lose the `if` and then defer to return the result from the super call.

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,299 @@
...
+        returned_data = event_data.get('event_data', {})
+        if returned_data.get('task_action', '') == 'set_fact':
+            returned_facts = returned_data.get('res', {}).get('ansible_facts', {})
+            if 'scm_version' in returned_facts:
+                self.playbook_new_revision = returned_facts['scm_version']
+
+
+class RunnerCallbackForInventoryUpdate(RunnerCallback):
+    def __init__(self):
+        super(RunnerCallbackForInventoryUpdate, self).__init__()
```
AlanCoding: shouldn't need `__init__` here.

amolgautam25: I have removed it.

awx/main/tasks/callback.py
amolgautam25: I have made the changes

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,232 @@
...
+        Ansible runner puts a parent_uuid on each event, no matter what the type.
+        AWX only saves the parent_uuid if the event is for a Job.
+        """
+        # cache end_line locally for RunInventoryUpdate tasks
+        # which generate job events from two 'streams':
+        # ansible-inventory and the awx.main.commands.inventory_import
+        # logger
+        self.end_line = event_data['end_line']
+
+        return super(RunnerCallbackForInventoryUpdate, self).event_handler(event_data)
```
AlanCoding, line 243: This looks great! As an aesthetic comment, I would lose the long code comment because it's covered by the parent class.

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,232 @@
...
+    def event_handler(self, event_data):
+        super(RunnerCallbackForProjectUpdate, self).event_handler(event_data)
+        returned_data = event_data.get('event_data', {})
+        if returned_data.get('task_action', '') == 'set_fact':
+            returned_facts = returned_data.get('res', {}).get('ansible_facts', {})
+            if 'scm_version' in returned_facts:
+                self.playbook_new_revision = returned_facts['scm_version']
+
+
+class RunnerCallbackForInventoryUpdate(RunnerCallback):
```
AlanCoding, line 232: This looks perfectly structured to add the `event_data_key` here right below the class declaration

awx/main/tasks/jobs.py
```diff
@@ -622,21 +458,21 @@ def run(self, pk, **kwargs):
         status, rc = 'error', None
         extra_update_fields = {}
         fact_modification_times = {}
-        self.event_ct = 0
+        self.runner_callback.event_ct = 0
```
AlanCoding, line 437: I think it would be best to move this to the `__init__` of the callback class, since it's just setting it to 0. That will also help with cases that we later reference `runner_callback.event_ct`, prevent attribute errors if this hasn't happened yet.

awx/main/tasks/jobs.py
```diff
@@ -622,21 +458,21 @@ def run(self, pk, **kwargs):
...
 
         '''
         Needs to be an object property because status_handler uses it in a callback context
         '''
-        self.safe_env = {}
+        self.runner_callback.safe_env = {}
```
AlanCoding: Again, I would say move this to the callback class `__init__`.

awx/main/tasks/jobs.py
```diff
@@ -669,15 +505,15 @@ def run(self, pk, **kwargs):
...
+            self.runner_callback.safe_env = build_safe_env(env)
 
             credentials = self.build_credentials_list(self.instance)
 
             for credential in credentials:
                 if credential:
                     credential.credential_type.inject_credential(credential, env, self.safe_cred_env, args, private_data_dir)
 
-            self.safe_env.update(self.safe_cred_env)
+            self.runner_callback.safe_env.update(self.safe_cred_env)
```
AlanCoding, line 494: You have a few things that need to be copied over from the task class to the callback class, but they're spread out throughout the `run` method. It would make the code easier to read if you consolidated them. Here would work. Those assignments look to be:
 - instance
 - parent_workflow_job_id
 - job_created
 - safe_env (assignment and then an update)

Just putting them all in 1 place would be nice. Maybe that will suggest that a helper method would help, but maybe not.

awx/main/tasks/callback.py
amolgautam25: I have removed that, will be pushed in the next commit.

awx/main/tasks/jobs.py
amolgautam25: I have made the changes , will be pushed in the next commit.

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,213 @@
...
+
+class RunnerCallback:
+
+    parent_workflow_job_id = None
+    host_map = {}
+    guid = GuidMiddleware.get_guid()
+    job_created = None
+    recent_event_timings = deque(maxlen=settings.MAX_WEBSOCKET_EVENT_RATE)
+    dispatcher = CallbackQueueDispatcher()
+    safe_env = {}
```
AlanCoding: These are mutable data structures, and when you put them here (below the `class`), they are attached to the _class_. This is perfectly fine for immutable constants because they never change. This is a bad idea for anything that will be modified as you go. For those, it's better to set initial values in `__init__`. All of these look good, they just need to be saved to `self` in the init method so that they have a clearly defined scope of the lifetime of the RunnerCallback _object_, not the class. Classes are kind of immortal.

awx/main/tasks/jobs.py
amolgautam25: Done , will be pushed in the next commit

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,247 @@
...
+
+        return super(RunnerCallbackForInventoryUpdate, self).event_handler(event_data)
+
+
+class RunnerCallbackForAdHocCommand(RunnerCallback):
+
+    event_data_key = 'ad_hoc_command_id'
+
+    def __init__(self):
+        super(RunnerCallbackForJob, self).__init__()
```
AlanCoding, line 239: You don't have the right class named passed here, it would be `RunnerCallbackForAdHocCommand` in this case, but it's better to delete these `__init__` methods that don't do anything other than call the super.

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,247 @@
...
+                with disable_activity_stream():
+                    self.instance = self.update_model(self.instance.pk, result_traceback=result_traceback)
+
+
+class RunnerCallbackForJob(RunnerCallback):
+
+    event_data_key = 'job_id'
+
+    def __init__(self):
+        super(RunnerCallbackForJob, self).__init__()
```
AlanCoding: In this class, I would like to set `self.host_map` to `{}`. Same for the ad-hoc command class. This avoids throwing an AttributeError when referencing it before it has been assigned.

awx/main/tasks/jobs.py
```diff
@@ -444,171 +437,20 @@ def final_run_hook(self, instance, status, private_data_dir, fact_modification_t
...
         """
         self.instance = self.model.objects.get(pk=pk)
+        self.runner_callback.instance = self.instance
+
+        # store a reference to the parent workflow job (if any) so we can include
+        # it in event data JSON
+        if self.instance.spawned_by_workflow:
+            self.runner_callback.parent_workflow_job_id = self.instance.get_workflow_job().id
+
+        self.runner_callback.job_created = str(self.instance.created)
```
AlanCoding: We need to avoid logic that might fail before the `try:` block starts. This looks good, all of these lines are good logic, I just don't want them _here_. It should work equally well around line 515.

That will help further consolidate all operations that add something to `self.runner_callback`. it's better to have them all in the same place as opposed to scattered about.

awx/main/tasks/jobs.py
```diff
@@ -1199,21 +1040,13 @@ class RunProjectUpdate(BaseTask):
...
         self.job_private_data_dir = job_private_data_dir
-
-    def event_handler(self, event_data):
-        super(RunProjectUpdate, self).event_handler(event_data)
-        returned_data = event_data.get('event_data', {})
-        if returned_data.get('task_action', '') == 'set_fact':
-            returned_facts = returned_data.get('res', {}).get('ansible_facts', {})
-            if 'scm_version' in returned_facts:
-                self.playbook_new_revision = returned_facts['scm_version']
+        self.runner_callback = self.callback_class()
```
AlanCoding: You shouldn't need this line, because it is done in `super(RunProjectUpdate, self).__init__(*args, **kwargs)`. That runs the `BaseTask` version of the init method.

When `BaseTask.__init__` runs, it will do this, `self.runner_callback = self.callback_class()`, and when it does that `self` is an instance of `RunProjectUpdate`, so it'll pick up the class property of `callback_class` so it should correctly make the project update version of the callback object.

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,243 @@
...
+from django.utils.timezone import now
+from django.conf import settings
+from django_guid.middleware import GuidMiddleware
+
+
+# AWX
+from awx.main.redact import UriCleaner
+from awx.main.constants import (
+    MINIMAL_EVENTS,
+)
```
AlanCoding: syntax, put in single line.

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,243 @@
...
+
+from awx.main.queue import CallbackQueueDispatcher
+
+logger = logging.getLogger('awx.main.tasks.jobs')
```
AlanCoding: change logger jobs -> callback

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,243 @@
...
+        self.playbook_new_revision = None
+        self.host_map = {}
+
+    def event_handler(self, event_data):
+        super(RunnerCallbackForProjectUpdate, self).event_handler(event_data)
+        returned_data = event_data.get('event_data', {})
+        if returned_data.get('task_action', '') == 'set_fact':
+            returned_facts = returned_data.get('res', {}).get('ansible_facts', {})
+            if 'scm_version' in returned_facts:
+                self.playbook_new_revision = returned_facts['scm_version']
```
AlanCoding, line 228: this should return the same value as the super class, `r = super(...` and then `return r` at the end

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+class RunnerCallbackForInventoryUpdate(RunnerCallback):
+
+    event_data_key = 'inventory_update_id'
+
+    def __init__(self):
+        super(RunnerCallbackForInventoryUpdate, self).__init()
```
AlanCoding: Here I would set `self.end_line` to 0

awx/main/tasks/jobs.py
```diff
@@ -2148,7 +1984,10 @@ class RunSystemJob(BaseTask):
 
     model = SystemJob
     event_model = SystemJobEvent
-    event_data_key = 'system_job_id'
+    callback_class = RunnerCallbackForSystemJob
+
+    def __init__(self):
+        self.runner_callback = self.callback_class()
```
AlanCoding: Highest priority still TODO items seems like it would be to delete these `__init__` methods like this.

awx/main/tasks/jobs.py
```diff
@@ -800,7 +640,9 @@ class RunJob(BaseTask):
 
     model = Job
     event_model = JobEvent
-    event_data_key = 'job_id'
+
+    def __init__(self):
+        super(RunJob, self).__init__()
```
AlanCoding: shouldn't need this `__init__`, there's also another init method in the callback module that you can lose

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,240 @@
...
+            for k, v in self.safe_env.items():
+                if k in job_env:
+                    job_env[k] = v
+            from awx.main.signals import disable_activity_stream  # Circular import
+
+            with disable_activity_stream():
+                self.instance.job_args = json.dumps(runner_config.command)
+                self.instance.job_cwd = runner_config.cwd
+                self.instance.job_env = job_env
+                self.instance.save()
```
AlanCoding: for safety / performance, I do want to look into passing `update_fields` here. I think the updated fields are just these, right? job_args, job_cwd, job_env

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,240 @@
...
+        elif status_data['status'] == 'error':
+            result_traceback = status_data.get('result_traceback', None)
+            if result_traceback:
+                from awx.main.signals import disable_activity_stream  # Circular import
+
+                with disable_activity_stream():
+                    self.instance = self.update_model(self.instance.pk, result_traceback=result_traceback)
```
AlanCoding, line 210: whoops! we don't have `self.update_model`. You want to follow the same `self.instance.save()` pattern, but also probably include update fields, which would just be `result_traceback` here

awx/main/tasks/callback.py
amolgautam25: Yes , correct. 
I am looking at the original code and it seems only these 3 parameters are being changed.

awx/main/tasks/callback.py
fosterseth: @AlanCoding 

do we need this `update_model` method at all, or can we scrap it? It's old (8 years +)

it has some retry logic in case of db errors, not sure how relevant that is anymore.

AlanCoding: It's tied into how the unit tests work, and there is extra stuff about how the `modified` field is handled which we probably still need. I agree about the DB access. The philosophy is that the job prep and transmission should use the DB just like normal, but after receptor has received the job, the control process should maintain no active DB connections whatsoever. That's still a ways off, and I don't want to rush into refactoring the unit tests because I don't think we're ready for it.

awx/main/tasks/callback.py
```diff
@@ -0,0 +1,257 @@
...
+        event_data.setdefault(self.event_data_key, self.instance.id)
+        self.dispatcher.dispatch(event_data)
+        self.event_ct += 1
+
+        '''
+        Handle artifacts
+        '''
+        if event_data.get('event_data', {}).get('artifact_data', {}):
+            self.instance.artifacts = event_data['event_data']['artifact_data']
+            self.instance.save(update_fields=['artifacts'])
```
AlanCoding, line 141: I feel like we should be consistent and if we're using update_model at all, we should use it here.

awx/main/tasks/utils.py
```diff
@@ -0,0 +1,40 @@
...
+        # Log out the error to the debug logger.
+        logger.debug('Database error updating %s, retrying in 5 ' 'seconds (retry #%d): %s', model._meta.object_name, _attempt + 1, e)
+
+        # Attempt to retry the update, assuming we haven't already
+        # tried too many times.
+        if _attempt < 5:
+            time.sleep(5)
+            return update_model(model, pk, _attempt=_attempt + 1, **updates)
+        else:
+            logger.error('Failed to update %s after %d retries.', model._meta.object_name, _attempt)
```
AlanCoding: @jbradberry I wanted to ask if you had any input on where a simple static method like this should go.

AlanCoding (approved):

awx/main/tasks/callback.py
amolgautam25: Got it. I will go ahead and change it. However, this ```save``` method isn't put by me. It was there even before i started refactoring the code. So i thought to leave it as is.

awx/main/tasks/utils.py
shanemcd: It might be better to create a new file under https://github.com/ansible/awx/tree/devel/awx/main/utils

awx/main/tasks/utils.py
```diff
@@ -0,0 +1,40 @@
...
+                for field, value in updates.items():
+                    setattr(instance, field, value)
+                    update_fields.append(field)
+                    if field == 'status':
+                        update_fields.append('failed')
+                instance.save(update_fields=update_fields)
+            return instance
+    except DatabaseError as e:
+        # Log out the error to the debug logger.
+        logger.debug('Database error updating %s, retrying in 5 ' 'seconds (retry #%d): %s', model._meta.object_name, _attempt + 1, e)
```
jbradberry: We have some scar tissue in this string from the Black conversion from a while back.  `5 ' 'seconds` -> `5 seconds`

awx/main/tasks/utils.py
jbradberry: I think I agree with Shane.

awx/main/tasks/callback.py
jbradberry: Red flag.  This should _not_ be making use of a function that is going to query a fresh copy out of the database.

Issue #11588, opened 2022-01-21
Add top-level Instances List Page


kialam:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Similar to how we interpret Hosts and Inventories, we have top level list pages for both views as well as tabbed list views to represent their relationship. We would like to apply this to the Instances Groups and Instances page.

☐ Create top-level instances list page (between Instance Groups and Applications)
  ☐ Add search bar, health check buttons
  ☐ Promote "Node Type" detail to top level column view
  ☐ Remove "Running Jobs", "Total Jobs" from top level column view and put in expand/collapse detail area
  ☐ Include "hop" nodes/instances in Instances list page
  ☐ "Hop" nodes cannot be enabled or disabled via the toggle
  ☐ "Hop" nodes cannot be selected via the checkbox to be disassociated/health check
  ☐ User can see instance details page by clicking on an instance from the Instances List page
  ☐ Mesh viz will redirect users to Instances > Instance details page
  ☐ Only admins and auditors can see Instances list page

  Future scope
  ☐ add "associate" button to top level Instances page

AlexSCorey:
related https://github.com/ansible/awx/issues/11425

Issue #11575, opened 2022-01-20
Server error doing a GET to /api/v2/instances/N/


AlanCoding:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

Detail view of an instance gives a 500 doing a GET request as an organization admin

### AWX version

19.5.0

### Installation method

docker on linux

### Modifications

no

### Ansible version

devel

### Operating system

N/A

### Web browser

Chrome

### Steps to reproduce

 - create a new user
 - make user admin of an organization
 - create an instance group and add an instance
 - add the instance group to the org's instance groups
 - try to view the instance in that IG as the org admin

I might have missed some details, this is from `tests/api/rbac/test_instance_group_rbac.py::TestInstanceGroupRBAC::test_org_admin` test.

### Expected results

works

### Actual results

500 server error

### Additional information

logs

```
2022-01-20 20:51:08,268 ERROR    [c25d4a7f57904ceabe3945e152cdb51a] django.request Internal Server Error: /api/v2/instances/1/
Traceback (most recent call last):
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/lib64/python3.8/contextlib.py", line 75, in inner
    return func(*args, **kwds)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/api/generics.py", line 332, in dispatch
[...]

PR #11591, opened 2022-01-21
Turn off the filtering of hop nodes from the Instance endpoints

jbradberry:
##### SUMMARY

This change re-enables hop nodes being accessible in all API endpoints except for the health check.  This enables having a UI detail page that users can reach from the mesh visualizer.

As a side-effect, this deals with the problem in #11575.

related #11588 
related #11575 

##### ISSUE TYPE

 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/tasks/system.py
+++ b/awx/main/tasks/system.py
@@ -437,10 +437,15 @@ def inspect_execution_nodes(instance_list):
         for ad in workers:
             hostname = ad['NodeID']
             changed = False
+
             if hostname in node_lookup:
                 instance = node_lookup[hostname]
             else:
                 logger.warn(f"Unrecognized node advertising on mesh: {hostname}")
+                continue
+
+            # Control-plane nodes are dealt with via local_health_check instead.
+            if instance.node_type in ('control', 'hybrid'):
                 continue
 
             was_lost = instance.is_lost(ref_time=nowtime)
@@ -451,8 +456,8 @@ def inspect_execution_nodes(instance_list):
             instance.last_seen = last_seen
             instance.save(update_fields=['last_seen'])
 
-            # Make sure that hop nodes don't fall through and have the execution_node_health_check task applied
-            if not any(cmd['WorkType'] == 'ansible-runner' for cmd in ad['WorkCommands'] or []):
+            # Only execution nodes should be dealt with by execution_node_health_check
+            if instance.node_type == 'hop':
                 continue
 
             if changed:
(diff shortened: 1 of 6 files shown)


rebeccahhh (approved):

AlanCoding (approved):
I still see at least 2 things we need to do:
1. in the InstanceSerializer, avoid linking to the health check if it's a hop node
2. in the cluster_node_heartbeat, stop filtering out the hop nodes

doing (2) might involve adding some more exceptions, because the idea is that we don't run health checks, but we can still mark it as error if it hasn't updated its `last_seen` timestamp in ages.

Anyway, these are some good stability-oriented bug fixes and I say :+1:

awx/main/tasks/system.py
```diff
@@ -454,6 +451,10 @@ def inspect_execution_nodes(instance_list):
             instance.last_seen = last_seen
             instance.save(update_fields=['last_seen'])
 
+            # Make sure that hop nodes don't fall through and have the execution_node_health_check task applied
+            if not any(cmd['WorkType'] == 'ansible-runner' for cmd in ad['WorkCommands'] or []):
+                continue
```
AlanCoding, line 461: yeah, this is probably not needed anymore, but it looks good as a defensive coding measure.

awx/main/tasks/system.py
```diff
@@ -482,7 +483,6 @@ def cluster_node_heartbeat():
     for inst in instance_list:
         if inst.hostname == settings.CLUSTER_HOST_ID:
             this_inst = inst
-            instance_list.remove(inst)
```
AlanCoding, line 485: Removing this, I would like to avoid `inst` getting in the `lost_instances` list. By definition, if you are running this code, you are not a lost instance. Any issues will be handled in `this_isnt.local_health_check()` which is a separate block of code. I'm a little worried that the order of operations might jerk its error status back and fourth.

AlanCoding (approved):
Left one minor logic change request, otherwise everything looks good to go :+1:

awx/main/tasks/system.py
jbradberry: Marking this for posterity: we decided on a clearer, more-readable pattern for these checks.

awx/main/tasks/system.py
jbradberry: Dealt with, for now.  A real cleanup of this code seems in order, though.

Issue #11017, opened 2021-09-03
Unable to use ansible_facts for smart inventory source


ivantomica:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

One older article describes how ansible_facts can be used to create smart inventory within AWX/Ansible tower: https://www.ansible.com/blog/ansible-tower-advanced-smart-inventory-usage

Unfortunately this feature seems to be broken or not properly documented, as API doesn't seem to be recognizing field ansible_facts.

GET to /api/v2/hosts/?ansible_facts.ansible_distribution_version=18.04
```
HTTP 400 Bad Request
Allow: GET, POST, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept
X-API-Node: awx-6bd564b87c-26bv6
X-API-Product-Name: AWX
X-API-Product-Version: 19.3.0
X-API-Time: 0.005s

{
    "error": "Host has no field named 'ansible_facts.ansible_distribution_version'"
}
```

### AWX version

19.3.0

### Installation method

kubernetes

### Modifications

no

### Web browser

Firefox

### Steps to reproduce

1. Go to Inventory
2. Add - Smart Inventory
3. Try to filter using "Smart host filter" field

Alternative:
1. Go to /api/v2/hosts/
2. try using query string with desired field (eg /api/v2/hosts/?ansible_facts.ansible_distribution_version=18.04)

### Expected results

API queries facts and displays hosts that only match those facts

### Actual results

API complains that the field is incorrect

### Additional information

Using any top-level field works normally:

eg. /api/v2/hosts/?id=302 (if host with ID 302 in AWX exists) returns properties for that host.

It would appear that since the ansible_facts is just a reference to additional page that the API is not able to query it.

chrismeyersfsu:
`GET to /api/v2/hosts/?ansible_facts.ansible_distribution_version=18.04` should be `ansible_facts__ansible_distribution_version=18.04` see these examples https://docs.ansible.com/ansible-tower/latest/html/userguide/inventories.html#smart-host-filter

ivantomica:
Thank you @chrismeyersfsu! I totally missed that part of docs. It helps a lot.

[...]

Issue #7850, opened 2020-08-06
Add smart inventory host filter lookup text input


marshmalien:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Expose a simple text input that a user could opt in to when creating a smart inventory host filter where a complete host_filter string could be supplied. This would serve 2 purposes:
1. Users wanting to specify complex/non-ui supported host filter strings could do so in the UI and wouldn't have to go through the API browser
2. If the UI detects a string that we can't parse into consumable search tags then we just show the string in the input.

Example of a host filter string that the user could type into the input:
```
(name=bengal and groups__name=Cat) or (name=vizsla and groups__name=Dog)
```

To retrieve the corresponding list of hosts and display them in the modal, the same filter string above must be url-encoded:
```
/hosts/?host_filter=(name%3Dbengal%20and%20groups__name%3D%22Cat%22)%20or%20(name%3Dvizsla%20and%20groups__name%3DDog)
```

##### ADDITIONAL INFO
Smart host filter docs: https://docs.ansible.com/ansible-tower/latest/html/userguide/inventories.html#smart-host-filter
Initial host filter lookup PR:  https://github.com/ansible/awx/pull/7644

Host filter lookup mockup: https://tower-mockups.testing.ansible.com/patternfly/inventories/inventories-smart-edit-filter/
<img width="1304" alt="Screen Shot 2020-08-06 at 12 37 42 PM" src="https://user-images.githubusercontent.com/15881645/89557986-a390a300-d7e1-11ea-91c3-144e05f997cc.png">

Image below is a mockup of what it could put in the host filter lookup or advanced search
<img width="889" alt="Screen Shot 2020-08-06 at 12 31 09 PM" src="https://user-images.githubusercontent.com/15881645/89557399-be164c80-d7e0-11ea-823e-73131215fa91.png">

nixocio:
See also: https://github.com/ansible/awx/issues/11985

cloneluke:
I would love this feature, right now we can only filter on one ansible_fact at a time, but we brought in all of our aws tags as ansible facts and are limited here.

PR #11592, opened 2022-01-21
Modify usage of ansible_facts on advanced search

nixocio:
Modify usage of ansible_facts on advanced search, once `ansible_facts`
key is selected render a text input allowing the user to type special
query expected for ansible_facts.

This change will add more flexibility to the usage of ansible_facts when
creating a smart inventory.

In order to filter by ansible_facts, the API requires the `host_filter=ansible_facts__query`. Adjust to cover this case. 

This PR is not addressing the compound search, but it is being tracked as another issue: https://github.com/ansible/awx/issues/7850

See: https://github.com/ansible/awx/issues/11017

image

image

image

image

### DIFF
--- a/awx/ui/src/components/ListHeader/ListHeader.js
+++ b/awx/ui/src/components/ListHeader/ListHeader.js
@@ -1,5 +1,5 @@
 /* eslint-disable react/jsx-no-useless-fragment */
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { useHistory, useLocation } from 'react-router-dom';
 import styled from 'styled-components';
@@ -27,6 +27,7 @@ const EmptyStateControlsWrapper = styled.div`
 `;
 function ListHeader(props) {
   const { search, pathname } = useLocation();
+  const [isFilterCleared, setIsFilterCleared] = useState(false);
   const history = useHistory();
   const {
     emptyStateControls,
@@ -73,6 +74,7 @@ function ListHeader(props) {
     delete oldParams.page_size;
     delete oldParams.order_by;
     const qs = updateQueryString(qsConfig, search, oldParams);
+    setIsFilterCleared(true);
     pushHistoryState(qs);
   };
 
@@ -120,6 +122,7 @@ function ListHeader(props) {
             clearAllFilters: handleRemoveAll,
             qsConfig,
             pagination,
+            isFilterCleared,
           })}
         </>
       )}
--- a/awx/ui/src/components/Lookup/HostFilterLookup.js
+++ b/awx/ui/src/components/Lookup/HostFilterLookup.js
@@ -6,6 +6,7 @@ import styled from 'styled-components';
 import { t } from '@lingui/macro';
 import { SearchIcon } from '@patternfly/react-icons';
 import {
+  Alert as PFAlert,
   Button,
   ButtonVariant,
   Chip,
@@ -16,6 +17,8 @@ import {
 } from '@patternfly/react-core';
 import { HostsAPI } from 'api';
 import { getQSConfig, mergeParams, parseQueryString } from 'util/qs';
+import getDocsBaseUrl from 'util/getDocsBaseUrl';
+import { useConfig } from 'contexts/Config';
 import useRequest, { useDismissableError } from 'hooks/useRequest';
 import ChipGroup from '../ChipGroup';
 import Popover from '../Popover';
@@ -33,7 +36,14 @@ import {
   toHostFilter,
   toQueryString,
   toSearchParams,
+  modifyHostFilter,
 } from './shared/HostFilterUtils';
+
+const Alert = styled(PFAlert)`
+  && {
+    margin-bottom: 8px;
+  }
+`;
 
 const ChipHolder = styled.div`
   && {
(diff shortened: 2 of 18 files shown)


awx/ui/src/components/Lookup/HostFilterLookup.js
```diff
@@ -140,6 +153,10 @@ function HostFilterLookup({
     return searchFields;
   };
 
+  const handleIsAnsibleFactsSelected = (v) => {
+    setIsAnsibleFactsSelected(v);
+  };
```
keithjgrant: This is as unnecessary wrapper function. Just pass through `setIsAnsibleFactsSelected` below

awx/ui/src/components/Search/AdvancedSearch.js
```diff
@@ -86,7 +118,12 @@ function AdvancedSearch({
       const actualSearchKey = [actualPrefix, keySelection, lookupSelection]
         .filter((val) => !!val)
         .join('__');
-      onSearch(actualSearchKey, searchValue);
+      if (keySelection === 'ansible_facts') {
+        const ansibleFactValue = [actualSearchKey, searchValue].join('__');
```
keithjgrant: why not use `${actualSearchKey}__${searchValue}` in a string? Seems a little more direct than using `join`

awx/ui/src/components/Search/AdvancedSearch.js
```diff
@@ -137,17 +174,78 @@ function AdvancedSearch({
     </Select>
   );
 
-  return (
-    <AdvancedGroup>
-      {lookupSelection === 'search' ? (
+  const renderLookupType = () => {
+    if (keySelection !== 'ansible_facts') {
```
keithjgrant: Use a "return early" pattern here to un-indent the whole `if` block

```
if (keySelection === 'ansible_facts') {
  return null;
}
...
```

awx/ui/src/components/Search/AdvancedSearch.js
```diff
@@ -137,17 +174,78 @@ function AdvancedSearch({
...
+        isDisabled={!keySelection}
+        value={(!keySelection && t`First, select a key`) || searchValue}
+        onChange={setSearchValue}
+        onKeyDown={handleAdvancedTextKeyDown}
+      />
+    );
+  };
+
+  const renderLookupSelection = () => {
+    if (keySelection !== 'ansible_facts') {
```
keithjgrant: you can return early here too

```
if (keySelection === 'ansible_facts') {
  return null;
}
...
```

awx/ui/src/components/Lookup/HostFilterLookup.js
```diff
@@ -363,6 +381,26 @@ function HostFilterLookup({
...
+                    target="_blank"
+                    rel="noopener noreferrer"
+                  >
+                    {t`documentation`}
+                  </a>{' '}
+                  {t`for more info.`}
+                </>
+              }
+            />
+          )}
```
keithjgrant: Ideally, we'd render this inside of AdvancedSearch, so the `handleIsAnsibleFactsSelected` prop wouldn't need to be drilled down through all the intermediate components. I don't think there's a way to do that and get the layout we want though, unfortunately

keithjgrant (commented):
I'm still looking through, but here are some initial thoughts...

awx/ui/src/components/Lookup/HostFilterLookup.js
```diff
@@ -378,12 +424,15 @@ function HostFilterLookup({
             }
             renderRow={(item) => <HostListItem key={item.id} item={item} />}
             renderToolbar={(props) => (
-              <DataListToolbar
-                {...props}
-                fillWidth
-                enableNegativeFiltering={enableNegativeFiltering}
-                enableRelatedFuzzyFiltering={enableRelatedFuzzyFiltering}
-              />
+              <>
```
marshmalien: We could probably remove this fragment.

awx/ui/src/components/Lookup/shared/HostFilterUtils.js
```diff
@@ -159,3 +160,32 @@ export function removeDefaultParams(config, obj = {}) {
...
   return clonedObj;
 }
+
+/**
+ * Helper function to update host_filter value
+ * @param {string} value A string with host_filter value from querystring
+ * @param {object} obj An object returned by toSearchParams - in which the
+ * host_filter value was partially removed.
+ * @return {object} An object with the value of host_filter modified
+ */
```
marshmalien, line 170: 💯

awx/ui/src/screens/Inventory/shared/utils.js
```diff
@@ -0,0 +1,10 @@
+const parseHostFilter = (value) => {
```
marshmalien, line 1: We should add a unit test for this helper function.

marshmalien (approved):
Great work on this @nixocio!! 🎉

awx/ui/src/components/ListHeader/ListHeader.js
```diff
@@ -56,6 +57,11 @@ function ListHeader(props) {
   };
 
   const handleRemove = (key, value) => {
+    if (key === 'ansible_facts') {
+      value = `${key}__${value}`;
+      key = 'host_filter';
+    }
+
```
keithjgrant: I think this logic would make more sense inside Search, since that's where we do the reverse work when we call parseQueryString(). I think the fewer components that need to be aware of this special logic, the better

keithjgrant (approved):

nixocio:
I triggered the e2e tests.

akus062381:
@nixocio I am seeing this screen when i follow the documentation link:

image

Other than that, this feature works well

akus062381 (approved):

nixocio:
hey @akus062381, I verified the document link, and that is a generic function. Locally I was redirect to 
image

I will go ahead and merge this PR. But add a note to verify this link on the final build. Thanks.

Issue #11584, opened 2022-01-21
Project Sync Error Due to Missing Resource


akus062381:
I built a new ec2 instance against AWX devel. the initial project sync ended in Error status due to missing resource:
![image](https://user-images.githubusercontent.com/63685497/150571400-c23bef8b-357d-4cbf-851d-56f1ffc555c4.png)
![image](https://user-images.githubusercontent.com/63685497/150571421-8181d0d4-73a1-42f7-82b7-323868cc7891.png)

Subsequent project sync attempts were successful.

Linked issue:
https://github.com/ansible/awx/issues/11583

nixocio:
I also started an ec2 instance using awx/devel and got the same error. This is not an UI issue. Same behavior as above.

![image](https://user-images.githubusercontent.com/9053044/150804396-7b62624a-4998-4259-bed0-7a11d406e4ab.png)

EE presents on the system

![image](https://user-images.githubusercontent.com/9053044/150805556-35d7406c-cff2-4e42-8c17-297d069403ba.png)

tchellomello:
@shanemcd wasn't this error that we were getting during the `awx-operator` testing our the last PR?

tchellomello:
@akus062381 @nixocio this issue should be fixed in devel now. Please let know if that still occurs for you guys.

PR #11602, opened 2022-01-24
Avoid Project..get_or_create() in create_preload_data

tchellomello:
Django ORM method get_or_create() does not call save() directly, but it calls the create() [1].

https://github.com/django/django/blob/0f6946495a8ec955b471ca1baaf408ceb53d4796/django/db/models/query.py#L678

The create method ignores the `skip_update=True` option, which then will trigger a project update, however, the EE was not yet created in the database.

To avoid this problem, we just check the existence of the default project and create it with save(skip_update=True) manually.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
devel
```

Related: https://github.com/ansible/awx/issues/11584

##### ADDITIONAL INFORMATION
Testing it:
```
awx-manage shell_plus --quiet
In [2]: JobTemplate.objects.all().delete()
   ...: Credential.objects.all().delete()
   ...: Project.objects.all().delete()
   ...: Inventory.objects.all().delete()
   ...: Organization.objects.all().delete()
   ...: ExecutionEnvironment.objects.all().delete()
   ...: 

Out[2]: 
(4,
 {'main.ActivityStream_execution_environment': 2,
  'main.ExecutionEnvironment': 2})
bash-4.4$ awx-manage create_preload_data
Default organization added.
Demo Credential, Inventory, and Job Template added.
(changed: True)
bash-4.4$ awx-manage register_default_execution_environments
'AWX EE (latest)' Default Execution Environment registered.
'AWX EE (latest)' Default Execution Environment updated.
Control Plane Execution Environment registered.
(changed: True)
```

No project updates were triggered 
```
In [1]: ProjectUpdate.objects.count()
Out[1]: 0

In [2]: Project.objects.all()
Out[2]: <PolymorphicQuerySet [<Project: Demo Project-25>]>

```

### DIFF
--- a/awx/main/management/commands/create_preload_data.py
+++ b/awx/main/management/commands/create_preload_data.py
@@ -25,13 +25,17 @@ class Command(BaseCommand):
                 if not Organization.objects.exists():
                     o, _ = Organization.objects.get_or_create(name='Default')
 
-                    p, _ = Project.objects.get_or_create(
-                        name='Demo Project',
-                        scm_type='git',
-                        scm_url='https://github.com/ansible/ansible-tower-samples',
-                        scm_update_on_launch=True,
-                        scm_update_cache_timeout=0,
-                    )
+                    # Avoid calling directly the get_or_create() to bypass project update
+                    p = Project.objects.filter(name='Demo Project', scm_type='git').first()
+                    if not p:
+                        p = Project(
+                            name='Demo Project',
+                            scm_type='git',
+                            scm_url='https://github.com/ansible/ansible-tower-samples',
+                            scm_update_on_launch=True,
+                            scm_update_cache_timeout=0,
+                        )
+
                     p.organization = o
                     p.save(skip_update=True)
 


kdelee (approved):
thanks, I hit this today too

awx/main/management/commands/create_preload_data.py
```diff
@@ -25,15 +25,17 @@ def handle(self, *args, **kwargs):
...
+                    if not Project.objects.filter(name='Demo Project', scm_type='git').count():
+                        p = Project(
+                            name='Demo Project',
+                            scm_type='git',
+                            scm_url='https://github.com/ansible/ansible-tower-samples',
+                            scm_update_on_launch=True,
+                            scm_update_cache_timeout=0,
+                        )
+                        p.organization = o
+                        p.save(skip_update=True)
```
rebeccahhh: just to verify, we don't need to worry about the organization not being saved here when the `if` is not called, correct?

tchellomello: Talking with @shanemcd we decided not to override since the user can do it manually as the project is not a `read-only` entry. So if that already exists, then we don't update the Org which will be only for the cases where the `default org` has been deleted.

tchellomello:
Tested locally and enabled auto-merge once tests passed.

awx/main/management/commands/create_preload_data.py
tchellomello: Thinking about it, we can fix this issue and also we use the `project` later so  I'll fix it.

PR #11606, opened 2022-01-25
Revert "Avoid Project..get_or_create() in create_preload_data"

akus062381:
Reverts ansible/awx#11602

checking to see if this is what introduced the UI failures.


AlanCoding:
https://github.com/ansible/awx/pull/11602 was expected to fix it and is merged now

PR #11619, opened 2022-01-27
Iterate over SAML attribute values

AdamVB:
Signed-off-by: abenesh <Adam-Daniel.Benesh@t-systems.com>

##### SUMMARY

The SSO Pipeline only evaluates the first value in a list of values that are part of a SAML attribute. Granting superuser or auditor permissions would fail if the first list value is not the expected one.

For example you have a SAML attribute "groups" that includes a list of the users group memberships:

```
<saml:Attribute Name="groups" NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:basic">
  <saml:AttributeValue xsi:type="xs:string">qaadmins</saml:AttributeValue>
  <saml:AttributeValue xsi:type="xs:string">jira-software-users</saml:AttributeValue>
  <saml:AttributeValue xsi:type="xs:string">awx-frontend-admins</saml:AttributeValue>
......
```

This attribute mapping would not work:

```
{
  "is_superuser_attr": "groups",
  "is_superuser_role": "is_superuser",
  "is_superuser_value": "awx-frontend-admins",
  "is_system_auditor_attr": "groups",
  "is_system_auditor_role": "is_system_auditor",
  "is_system_auditor_value": "awx-frontend-auditors"
}
```

Because ```awx-frontend-admins``` is not the first entry in the list.

Related to #5303

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME

 - SSO

##### AWX VERSION

```
awx: 19.5.2.dev41+g1b1b2fd4a7
```

### DIFF
--- a/awx/sso/pipeline.py
+++ b/awx/sso/pipeline.py
@@ -255,17 +255,18 @@ def _check_flag(user, flag, attributes, user_flags_settings):
             # If so, check and see if the value of the attr matches the required value
             attribute_value = attributes.get(attr_setting, None)
             if isinstance(attribute_value, (list, tuple)):
-                attribute_value = attribute_value[0]
-            if attribute_value == user_flags_settings.get(is_value_key):
-                logger.debug("Giving %s %s from attribute %s with matching value" % (user.username, flag, attr_setting))
-                new_flag = True
-            # if they don't match make sure that new_flag is false
-            else:
-                logger.debug(
-                    "Refusing %s for %s because attr %s (%s) did not match value '%s'"
-                    % (flag, user.username, attr_setting, attribute_value, user_flags_settings.get(is_value_key))
-                )
-                new_flag = False
+                for attribute_value in attribute_value:            
+                    if attribute_value == user_flags_settings.get(is_value_key):
+                        logger.debug("Giving %s %s from attribute %s with matching value" % (user.username, flag, attr_setting))
+                        new_flag = True
+                        break
+                    # if they don't match make sure that new_flag is false
+                    else:
+                        logger.debug(
+                            "Refusing %s for %s because attr %s (%s) did not match value '%s'"
+                            % (flag, user.username, attr_setting, attribute_value, user_flags_settings.get(is_value_key))
+                        )
+                        new_flag = False
         # If there was no required value then we can just allow them in because of the attribute
         else:
             logger.debug("Giving %s %s from attribute %s" % (user.username, flag, attr_setting))


awx/sso/pipeline.py
```diff
@@ -255,17 +255,18 @@ def _check_flag(user, flag, attributes, user_flags_settings):
             # If so, check and see if the value of the attr matches the required value
             attribute_value = attributes.get(attr_setting, None)
             if isinstance(attribute_value, (list, tuple)):
```
john-westcott-iv, line 257: Do we have documentation that attributes will always be a list? If so we don't need this if condition. If not, we should also account for strings here.

awx/sso/pipeline.py
```diff
@@ -255,17 +255,18 @@ def _check_flag(user, flag, attributes, user_flags_settings):
...
-                logger.debug("Giving %s %s from attribute %s with matching value" % (user.username, flag, attr_setting))
-                new_flag = True
-            # if they don't match make sure that new_flag is false
-            else:
-                logger.debug(
-                    "Refusing %s for %s because attr %s (%s) did not match value '%s'"
-                    % (flag, user.username, attr_setting, attribute_value, user_flags_settings.get(is_value_key))
-                )
-                new_flag = False
+                for attribute_value in attribute_value:
```
john-westcott-iv, line 258: Can we use different variable names for the loop item and what we are looping over?

awx/sso/pipeline.py
```diff
@@ -255,17 +255,18 @@ def _check_flag(user, flag, attributes, user_flags_settings):
...
+                        logger.debug("Giving %s %s from attribute %s with matching value" % (user.username, flag, attr_setting))
+                        new_flag = True
+                        break
+                    # if they don't match make sure that new_flag is false
+                    else:
+                        logger.debug(
+                            "Refusing %s for %s because attr %s (%s) did not match value '%s'"
+                            % (flag, user.username, attr_setting, attribute_value, user_flags_settings.get(is_value_key))
+                        )
+                        new_flag = False
```
john-westcott-iv, line 269: This else block was previously run outside of a loop. I don't think we would want to display that debug message for each of the values in an array (especially if one of them passes). Perhaps we could move this out of the for loop and do something like:
```
new_flag = False
<for loop>
    <has attribute if block>
    
if not new_flag:
    logger.debug(
        "Refusing %s for %s because attr %s (%s) did not match value '%s'"
        % (flag, user.username, attr_string, ', '.join(attribute_values, user_flags_settings.get(is_value_key))
    )
```

john-westcott-iv (changes requested):
There is also updated documentation with a table as to how this logic should be working with all of the 6 settings available at https://github.com/ansible/awx/blob/devel/docs/auth/saml.md.  We want to make sure any changes we make in this PR do not affect that table. (I don't think we are at this point but want to make sure we don't with the requested changes)

john-westcott-iv:
Closing in favor of https://github.com/ansible/awx/pull/11798

Issue #10694, opened 2021-07-20
Consider capacity for controller_node selection


AlanCoding:
edited by cmeyers:

Today, we randomly choose an online controller node https://github.com/ansible/awx/blob/b5fac4157df2f8c949361a7ca26452f47fc3baf5/awx/main/models/ha.py#L155

We should do "something better" (see comments below for ideas)

kdelee:
We need to:
  - consider the instance capacity when selecting the control node
  - deduct from the instance capacity when we select node as controller
  - release capacity/return capacity to the instance when job is done (when it is done with event processing??)

AlanCoding:
> deduct from the instance capacity when we select node as controller

We talked about deducting 1 unit. That should be sufficient for our purposes. We'd need more data on nodes with control-only nodes to further refine this.

> release capacity/return capacity to the instance when job is done (when it is done with event processing??)

This is pretty much automatic. We compute available capacity for the entire cluster based on the state of jobs using the instances. So when the job goes from pending/running to failed/successful, then it will be removed from consideration.

jladdjr:
We're thinking this is more of a "nice-to-have" item, and will probably get pushed back a bit.

kdelee:
update on current state of PR:

The capacity "impact" of being a controller node is default 1, but is configurable via a setting
In order to be selected to be a control node for a system job/project update, that means we need a minimum of 2 capacity. If a cluster has less than 2 capacity project updates will never run. But for a node to have less than 2 capacity it would need to have such minuscule amount of CPU and RAM...that it practically has 0.
If the control nodes in a cluster have less that 2 available capacity, any jobs in pending will stay in pending. A log to the job_lifecycle logger should indicate the job is waiting on capacity.

The "jobs_running" attribute on instances and instance groups does NOT count controlling jobs as "running jobs" -- only jobs that use node as the execution node/instance group.

The "consumed_capacity" attribute on instances DOES account for controlling jobs

kdelee:
https://github.com/ansible/awx/pull/11665 merged, some docs updates were added to https://github.com/ansible/awx/blob/devel/docs/capacity.md#impact-of-job-types-in-awx and https://github.com/ansible/awx/blob/devel/docs/capacity.md#selecting-the-right-settingsawx_control_node_task_impact

AlanCoding:
If tests are passing in a day or two then this should be a pretty easy close.

thenets:
So far, looks stable. Closing it.

PR #11620, opened 2022-01-27
[Draft] Consume control node capacity

kdelee:
I've tested with container groups in a local minikube cluster, but it generally results in desired behavior
of only starting container group jobs when we have a control node with
capacity to run them. 

Read @beeankha 's comment below to explore what remains to implement the same thing completely for all jobs that have a separate control node.

##### SUMMARY

This patch addresses https://github.com/ansible/awx/issues/10694

For container groups in particular, there is currently no controller-side mechanism to throttle how many jobs get launched. This can lead to overwhelmed control nodes, and control of jobs unevenly spread across cluster (e.g., you have 3 control pods but 1 sits idle while the other two are overtaxed).

Even before we start to see bad things like control pod eviction from resource consumption, things start to happen like jobs sitting in `'waiting'` for over a min, which can lead to the jobs getting "reaped" but then running and succeeding anyway! By counting capacity consumption for controlling jobs, we at least have the "capacity slider" as a tool to have _some_ way to manage how many jobs a control pod will be controlling/proccesing events for at any given time.

TODO:
  - [ ] add some tests
  - [ ] do some exploratory testing in a production k8s deployment
  - [ ] decide if we want to fully implement same thing for all jobs w/ separate control node in this PR, or in a follow up PR

##### ISSUE TYPE

 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/managers.py
+++ b/awx/main/managers.py
@@ -244,7 +244,7 @@ class InstanceGroupManager(models.Manager):
             # TODO: dock capacity for isolated job management tasks running in queue
             impact = t.task_impact
             control_groups = []
-            if t.controller_node:
+            if hasattr(t, 'controller_node') and t.controller_node:
                 # TODO: what happens if a control node comes online after graph was built?
                 # In k8s nodes need to be able to come/go more often
                 control_groups = instance_ig_mapping.get(t.controller_node, [])
--- a/awx/main/managers.py
+++ b/awx/main/managers.py
@@ -243,7 +243,11 @@ class InstanceGroupManager(models.Manager):
         for t in tasks:
             # TODO: dock capacity for isolated job management tasks running in queue
             impact = t.task_impact
-            if t.status == 'waiting' or not t.execution_node:
+            control_groups = []
+            if t.controller_node:
+                control_groups = instance_ig_mapping[t.controller_node]
+
+            if t.status == 'waiting' or (not t.execution_node and not t.is_container_group_task):
                 # Subtract capacity from any peer groups that share instances
                 if not t.instance_group:
                     impacted_groups = []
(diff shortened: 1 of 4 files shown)


rebeccahhh:
**Potential problems with our code:**
- awx/main/managers.py::250 - we saw some situations where the control node was not found in the mapping of the nodes to instance group. 
    * It must have appeared _after_ the mapping was made, this means we have more capacity than we originally planned for.
    * Consequence is that we will not attribute consumed capacity to the instance group that it must belong to.
    * We added a warning log, we want to test in production k8s deploy to see if we can make this occur in a prod environment as opposed to a dev env.

**How we think it works:**
1. task manager inits and gets the lock
2. the graph is an empty dict - then on line 89 of task_managers.py in the post init the graph is initialized with keys that are the names of insatnce groups and values that are dummy values (i.e. 0)
3. On line 104 it seems like we annotate their total potential capacity (nothing consumed so far)
4. We see the graph next in process_tasks on line 187 where it is implicitly passed in calculate_capacity_consumed to capacity_values from manager.py along with the list of all running and waiting task_managers.py
5. in capacity_values we zero out the graph. Why did we even do `#3` if we were just going to do this again.
6. then we loop over the tasks and decide how much capacity is being consumed in each instance group by the tasks and update the graph (lines 267 and 293).
    * originally we were only tracking the task impact on the execution instance group, but now we track the task impact on the instance groups that the control node belongs to too.
    * our changes will affect all jobs - at the beginning of the task maanger
7. For jobs that are using get_and_consume_capacity_on_control_node_with_sufficient_capacity to select and record the task_impact for the their control node we update the graph.
8. container group jobs will only leave pending once they have capacity on a control node, otherwise they will stay in pending.
9. for all non-container group jobs on line 568 in task_manager.py we finally check the graph by calling get_remaining_capacity in which we consult the graph (line 677).


**Remaining questions:**
- what's up with `#5`?? why do we initialize and then zero out the graph twice.
- A risk/to be figured out: we know the task manager has problems when there are too many jobs to process, we are intentionally putting a bottleneck on jobs leaving pending and will be collected and looped over by the task manger and may stress the start_task_limit

**Things we claim this PR does:**
- With these changes container group jobs fully consider and record their task impact on control nodes.
- All other nodes are impacted by our changes when the graph is initially built but still don't consider their impact on control nodes when transitioning from pending into waiting.
- If we wanted to fully apply consideration of capacity for selection of control node for _non-container_ group jobs we need to select the control node at some point between line 568 and line 597 (i.e. during process_pending_tasks)

kdelee:
Replacing with https://github.com/ansible/awx/pull/11651/commits

PR #11638, opened 2022-01-28
Only display usable inventories for launch prompt

jakemcdermott:
##### SUMMARY
Prompted job launch attempts fail when the user only has read or update access to the selected inventory. This PR updates the inventory prompt list request parameters so that only usable inventories are displayed.


nixocio (approved):
I got this code locally, and this failure here is fixed by this PR changes. 

I also triggered the e2e tests.

image

tiagodread (approved):

Issue #11560, opened 2022-01-18
AWX believes facts are modified even when gather_facts=false


AlanCoding:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

Jobs are no longer accurately determining if the fact cache should be updated. It is updated every time.

### AWX version

19.5.0

### Installation method

docker on linux

### Modifications

no

### Ansible version

devel

### Steps to reproduce

Run this playbook with extra vars

```yaml
set_fact_cacheable: true
```

https://github.com/ansible/test-playbooks/blob/master/gather_facts.yml

then run:

https://github.com/ansible/test-playbooks/blob/master/ping.yml

Inspect the hosts that this playbook targeted, and then look at the field `ansible_facts_modified`

### Expected results

The datetime value of `ansible_facts_modified` does not change with the second playbook run, because it does not affect facts

### Actual results

Every time every playbook is ran against any host `ansible_facts_modified` is updated to that job's time.

### Additional information

This isn't a very big problem for the case of jobs that don't gather facts.

It's a bigger problem for users who have large inventories and target 1 or a few hosts with the `limit` field, and have the fact cache enabled.

In these cases, we would spend a large amount of time updating facts for all hosts in the inventory, because there are a lot of hosts, and for each host, there are a lot of facts.

AlanCoding:
The proposed solution is that we fix: https://github.com/ansible/ansible-runner/issues/964, which @jbradberry went over in general terms.

In addition to this, I believe we should create some dummy files to help us keep track of time stamps. At both the start and finish of `start_job_fact_cache` we should touch a file in the artifacts directory. By doing this, we can compare the host facts files modification time to the reference files. This will tell us whether that file was created from the cached DB facts or not.

@amolgautam25 I got really excited about this because it could greatly help to improve code structure of `tasks/jobs.py`. We pass this data structure `fact_modification_times` around at the highest level. If we did the reference files approach, we would no longer need this at all. Plus... it's not working to begin with.

AlanCoding:
The ansible-runner PR is merged, unblocking this.

[...]

PR #11641, opened 2022-01-30
Fix fact clearing and updating of fact modification times

AlanCoding:
##### SUMMARY
Connect https://github.com/ansible/awx/issues/11560

This requires https://github.com/ansible/ansible-runner/pull/966 to work completely, which means you need to be using ansible-runner `devel` for now.

Upon creating this as a draft PR, I'm pretty confident in it, but I still haven't addressed unit tests, kicked off final test runs, and gone through a final polishing.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
My original draft of this involved backdating the files to a modification time of 2 days ago. However, I found that this doesn't work with `meta: clear_facts`, and I have documented why in https://github.com/ansible/ansible/issues/76882. Ansible core may or may not care to fix that, but I would not wait on a fix anyway. I just changed the backdating from 2 days to 30 seconds. It only needs to be greater than 2 seconds to fix timing bugs due to the python ZipInfo data rounding to the nearest 2nd second due to a [crazy old standard](https://stackoverflow.com/questions/64048499/zipfile-lib-weird-behaviour-with-seconds-in-modified-time). I could have kept it at 2 days ago if I set `ANSIBLE_CACHE_PLUGIN_TIMEOUT`, but I feel that should be the domain of the user.

This addresses a few articulable bugs, but the main reason I want to do this is to refactor the fact cache start / finished logic to be more self-contained without bleeding over into the main task code. I like the idea that someone could look at the artifacts folder and make a conclusion about why we did or did not update a host's facts.

### DIFF
--- a/awx/main/models/jobs.py
+++ b/awx/main/models/jobs.py
@@ -798,7 +798,9 @@ class Job(UnifiedJob, JobOptions, SurveyJobMixin, JobNotificationMixin, TaskMana
             return []
         return self.inventory.hosts.only(*only)
 
-    def start_job_fact_cache(self, destination, modification_times, timeout=None):
+    def start_job_fact_cache(self, private_data_dir, timeout=None):
+        artifacts_dir = os.path.join(private_data_dir, 'artifacts', str(self.id))
+        destination = os.path.join(artifacts_dir, 'fact_cache')
         self.log_lifecycle("start_job_fact_cache")
         os.makedirs(destination, mode=0o700)
         hosts = self._get_inventory_hosts()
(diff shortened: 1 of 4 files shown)


AlanCoding:
I've come around to a mostly finalized state, but I have a few tricky questions that deal with subtle behaviors which are hard to finalize a decision about.

1. What should the backdate value be? If your answer is less than 2 seconds, I implore you to review the evidence. Also, should it be a setting?
2. Should the Ansible setting `ANSIBLE_CACHE_PLUGIN_TIMEOUT` be set by us when fact gathering is turned on?
3. To deal with timezone and computer time clock sync issues, should we change the modified time comparision from `>` to `!=`?

the diff for the 3rd point

```diff
diff --git a/awx/main/models/jobs.py b/awx/main/models/jobs.py
index 00394f7d33..52c476853c 100644
--- a/awx/main/models/jobs.py
+++ b/awx/main/models/jobs.py
@@ -844,7 +844,7 @@ class Job(UnifiedJob, JobOptions, SurveyJobMixin, JobNotificationMixin, TaskMana
             if os.path.exists(filepath):
                 # If the file changed since we wrote it pre-playbook run...
                 modified = os.path.getmtime(filepath)
-                if modified > backdate:
+                if modified != backdate:
                     with codecs.open(filepath, 'r', encoding='utf-8') as f:
                         try:
                             ansible_facts = json.load(f)
```

kdelee (approved):

AlanCoding:
Have a newer PR that hits similar code.

PR #11642, opened 2022-01-30
Fix newly-added black rules

AlanCoding:
I upgraded black on my computer and saw that the upgrade introduced new reformatting behavior.

Implication is that black made a change and we must follow suit.

The new rule will fail checks on all other PRs until we merge this change.


jbradberry (approved):

kdelee (approved):
I saw this on my PR check too, thanks!

Issue #11640, opened 2022-01-28
control pods in k8s think they have capacity of whole k8s node


kdelee:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

Control pods are reporting as if each one has the capacity of the entire underlying k8s node it is on.

For example, I have one node in my minikube cluster:
```
kubectl get nodes -o json | jq '.items[].status.capacity'
{
  "cpu": "8",
  "ephemeral-storage": "71711864Ki",
  "hugepages-1Gi": "0",
  "hugepages-2Mi": "0",
  "memory": "32475736Ki",
  "pods": "110"
}
```

And in my awx deployment I have 3 control pods, each one reporting the entire CPU and memory of the underlying node:
```
{
    "count": 3,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": 11,
            "type": "instance",
            "url": "/api/v2/instances/11/",
            "related": {
                "jobs": "/api/v2/instances/11/jobs/",
                "instance_groups": "/api/v2/instances/11/instance_groups/",
                "health_check": "/api/v2/instances/11/health_check/"
            },
            "uuid": "adaab148-5636-4984-b36e-be901eeff068",
            "hostname": "awx-56fbfbb6c8-bw2ms",
            "created": "2022-01-28T22:42:02.019666Z",
            "modified": "2022-01-28T22:42:02.019770Z",
            "last_seen": "2022-01-28T22:43:02.187845Z",
            "last_health_check": "2022-01-28T22:43:02.187845Z",
            "errors": "",
            "capacity_adjustment": "1.00",
            "version": "12.0.0",
            "capacity": 296,
            "consumed_capacity": 0,
            "percent_capacity_remaining": 100.0,
            "jobs_running": 0,
            "jobs_total": 0,
            "cpu": 8,
            "memory": 33255153664,
            "cpu_capacity": 32,
            "mem_capacity": 296,
            "enabled": true,
            "managed_by_policy": true,
            "node_type": "control"
        },
        {
            "id": 7,
            "type": "instance",
            "url": "/api/v2/instances/7/",
            "related": {
                "jobs": "/api/v2/instances/7/jobs/",
[...]

Issue #11306, opened 2021-11-03
Task Manager Ideas


chrismeyersfsu:
This is to capture thoughts about the task manager over the years.

# Small - Optimization - Store Decision(s)

I think this was @fosterseth original idea.

The task manager decides to transition a job from `pending` to `waiting` when two conditions are met.

1. No blocking business logic
2. Capacity available

Today, the task manager checks to see if a job is blocked for business logic reasons, possibly multiple times. Instead, we could store when the decision has been made that a job is no longer blocked for business logic reasons (i.e. as a field on the job or even in a node-local cache). This, effectively, moves the job further through the pipeline. Note that if a job is not blocked on business logic and there is capacity, then the only reason to record the decision is for posterity. But, if a job is not blocked on business logic & there is no capacity, this is exactly when we want to record the decision made. The next time the task manager is ran, the marked jobs can avoid running the business logic check(s).

We should measure the performance before and after this optimization. We still need to put the Job from the DB to make the second decision but maybe we need to pull less dependency information i.e. for a job we wouldn't need to pull the project update.

# Large -  Incremental Update

Today, the task manager gets the "state of the world" from the database each time it runs. Technically, a little less than the state of the world. For example, we need all the jobs in the `pending` state. Let's say that we have job `J1`. We would need to know the status of the last project update for `J1`, but we don't need to know about ALL project updates. We don't _need_ to get this state from the database each iteration. Instead, we could keep an up-to-date "state of the world" in cache that is piece-wise updated as events occur.

"As events occur" isn't something we have historically been good at. For example, a developer may add code to a reaper for a new case. Any code that changes job `status` would need need to update this in cache view of the "state of the world". A developer modifying reaper code may not be aware that they need to update this cache. Since we know we aren't historically good at communicating this need this system design will need to be such that a developer need not know this cache exists. As much as I hate code that lives in model `save()`, it's the location in code that comes to mind to capture any `status` transition without the developer having to explicitly do so.

Let's also require that we can rebuild the "state of the world" from persistent storage because crashes happen and it's a nice feature. 

We will also need to propagate the state between all Controller control nodes and stay in-sync. We can use websockets or postgres LISTEN/NOTIFY. We also want to include some kind of hash of the "state of the world" so we can check if we are out of sync and/or corrupt.

[...]

PR #11651, opened 2022-01-31
[draft] Consume control node capacity

kdelee:
##### SUMMARY

This PR both consumes capacity on control nodes as well as considers their capacity when scheduling jobs.

replaces https://github.com/ansible/awx/pull/11620
addresses https://github.com/ansible/awx/issues/10694

##### ISSUE TYPE

 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.5.2.dev72+gf7b073ba5d
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/models/ha.py
+++ b/awx/main/models/ha.py
@@ -12,6 +12,7 @@ from django.dispatch import receiver
 from django.utils.translation import ugettext_lazy as _
 from django.conf import settings
 from django.utils.timezone import now, timedelta
+from django.db.models import Q
 
 import redis
 from solo.models import SingletonModel
@@ -146,9 +147,12 @@ class Instance(HasPolicyEditsMixin, BaseModel):
     @property
     def consumed_capacity(self):
         # TODO refactor, we can probably get this done in one query
-        control_capacity_consumed = sum(x.task_impact for x in UnifiedJob.objects.filter(controller_node=self.hostname, status__in=('running', 'waiting')))
-        execution_capacity_consumed = sum(x.task_impact for x in UnifiedJob.objects.filter(execution_node=self.hostname, status__in=('running', 'waiting')))
-        return control_capacity_consumed + execution_capacity_consumed
+        capacity_consumed = sum(
+            x.task_impact
+            for x in UnifiedJob.objects.filter(Q(controller_node=self.hostname) | Q(execution_node=self.hostname) & Q(status__in=('running', 'waiting')))
+        )
+        logger.debug("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA")
+        return capacity_consumed
 
     @property
     def remaining_capacity(self):
--- a/awx/main/models/ha.py
+++ b/awx/main/models/ha.py
@@ -146,10 +146,11 @@ class Instance(HasPolicyEditsMixin, BaseModel):
 
     @property
     def consumed_capacity(self):
-        capacity_consumed = sum(
-            x.task_impact
-            for x in UnifiedJob.objects.filter(Q(controller_node=self.hostname) | Q(execution_node=self.hostname) & Q(status__in=('running', 'waiting')))
-        )
+        tasks = [
+            x for x in UnifiedJob.objects.filter(Q(controller_node=self.hostname) | Q(execution_node=self.hostname) & Q(status__in=('running', 'waiting')))
+        ]
+        capacity_consumed = sum(x.task_impact for x in tasks)
+        logger.debug(f"tasks on the host: {[(task.log_format, task.task_impact) for task in tasks]}")
         logger.debug(f"{capacity_consumed} capacity consumed on {self.hostname}")
         return capacity_consumed
 
--- a/awx/main/models/ha.py
+++ b/awx/main/models/ha.py
@@ -146,11 +146,10 @@ class Instance(HasPolicyEditsMixin, BaseModel):
 
     @property
     def consumed_capacity(self):
-        tasks = [
-            x for x in UnifiedJob.objects.filter((Q(controller_node=self.hostname) | Q(execution_node=self.hostname)) & Q(status__in=('running', 'waiting')))
-        ]
-        capacity_consumed = sum(x.task_impact for x in tasks)
-        logger.debug(f"tasks on the host: {[(task.log_format, task.task_impact) for task in tasks]}")
+        capacity_consumed = sum(
+            x.task_impact
+            for x in UnifiedJob.objects.filter((Q(controller_node=self.hostname) | Q(execution_node=self.hostname)) & Q(status__in=('running', 'waiting')))
+        )
         logger.debug(f"{capacity_consumed} capacity consumed on {self.hostname}")
         return capacity_consumed
 
--- a/awx/main/models/ha.py
+++ b/awx/main/models/ha.py
@@ -178,9 +178,20 @@ class Instance(HasPolicyEditsMixin, BaseModel):
         )
 
     @staticmethod
-    def choose_control_plane_node_with_sufficient_capacity(task):
-        """Returns the control plane node with most capacity or None if there is none with capacity"""
-        instances = Instance.objects.filter(enabled=True, capacity__gte=task.task_impact).filter(node_type__in=['control', 'hybrid']).all()
+    def choose_control_plane_node_with_sufficient_capacity(task=None):
+        """Return an enabled control node with sufficient capacity.
+
+        Optionally provide a task that will inform the minimum amount of capacity needed,
+        otherwise an instance with remaining capacity greater than or equal to 0 will meet
+        criteria for selection.
+
+        If no instance meets criteria, return None.
+        """
+        # TODO: Replace most uses of choose_online_control_plane_node with this method.
+        impact = 0
+        if task:
+            impact = task.task_impact
+        instances = Instance.objects.filter(enabled=True, capacity__gte=impact).filter(node_type__in=['control', 'hybrid']).all()
         if instances:
             return max(instances, key=lambda i: i.remaining_capacity)
         return None
--- a/awx/main/models/ha.py
+++ b/awx/main/models/ha.py
@@ -147,7 +147,7 @@ class Instance(HasPolicyEditsMixin, BaseModel):
     @property
     def consumed_capacity(self):
         tasks = [
-            x for x in UnifiedJob.objects.filter(Q(controller_node=self.hostname) | Q(execution_node=self.hostname) & Q(status__in=('running', 'waiting')))
+            x for x in UnifiedJob.objects.filter((Q(controller_node=self.hostname) | Q(execution_node=self.hostname)) & Q(status__in=('running', 'waiting')))
         ]
         capacity_consumed = sum(x.task_impact for x in tasks)
         logger.debug(f"tasks on the host: {[(task.log_format, task.task_impact) for task in tasks]}")
--- a/awx/main/models/ha.py
+++ b/awx/main/models/ha.py
@@ -145,7 +145,10 @@ class Instance(HasPolicyEditsMixin, BaseModel):
 
     @property
     def consumed_capacity(self):
-        return sum(x.task_impact for x in UnifiedJob.objects.filter(execution_node=self.hostname, status__in=('running', 'waiting')))
+        # TODO refactor, we can probably get this done in one query
+        control_capacity_consumed = sum(x.task_impact for x in UnifiedJob.objects.filter(controller_node=self.hostname, status__in=('running', 'waiting')))
+        execution_capacity_consumed = sum(x.task_impact for x in UnifiedJob.objects.filter(execution_node=self.hostname, status__in=('running', 'waiting')))
+        return control_capacity_consumed + execution_capacity_consumed
 
     @property
     def remaining_capacity(self):
--- a/awx/main/models/ha.py
+++ b/awx/main/models/ha.py
@@ -146,12 +146,11 @@ class Instance(HasPolicyEditsMixin, BaseModel):
 
     @property
     def consumed_capacity(self):
-        # TODO refactor, we can probably get this done in one query
         capacity_consumed = sum(
             x.task_impact
             for x in UnifiedJob.objects.filter(Q(controller_node=self.hostname) | Q(execution_node=self.hostname) & Q(status__in=('running', 'waiting')))
         )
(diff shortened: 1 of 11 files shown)


awx/main/models/ha.py
```diff
@@ -145,7 +146,12 @@ def get_absolute_url(self, request=None):
 
     @property
     def consumed_capacity(self):
-        return sum(x.task_impact for x in UnifiedJob.objects.filter(execution_node=self.hostname, status__in=('running', 'waiting')))
+        capacity_consumed = sum(
+            x.task_impact
+            for x in UnifiedJob.objects.filter((Q(controller_node=self.hostname) | Q(execution_node=self.hostname)) & Q(status__in=('running', 'waiting')))
```
AlanCoding: Do you intend the filter here to return duplicates? Because if a job is on a hybrid node, it _should_ be double-counted. It should be counted by `task_impact` for the execution part, and the constant value for the control part, so at first glance I don't see how this is going to work.

What is the intended capacity consumption for the control nodes? I'd like to start with some docs about that.

awx/main/scheduler/task_manager.py
```diff
@@ -36,9 +35,8 @@
 from awx.main.utils import get_type_for_model, task_manager_bulk_reschedule, schedule_task_manager
 from awx.main.utils.common import create_partition
 from awx.main.signals import disable_activity_stream
-from awx.main.scheduler.dependency_graph import DependencyGraph
```
AlanCoding, line 39: Wouldn't this leave this class orphaned? No uses would remain. If you do this, you should delete that class (probably the whole module), but I might have more questions before jumping into that.

kdelee: The class does get used in the new file `awx/main/scheduler/task_manager_instances.py`

awx/main/models/ha.py
kdelee: Good call on double counting for hybrid nodes. I think I need to work on that, I originally was just working on container group jobs where that is never the case.

I had started with just counting the task impact as the same on the control node as the execution node, because that was simpler.

Problem with going with same control node task impact as execution:
  - we'll be able to run less jobs. But maybe for good reason, if you really are running a job against 1,000 hosts, the events coming back from that job are very likely MUCH more than a job running against 1 host. So scaling the control node impact w/ `task.task_impact` seems good.

Problem with going with small, constant impact for control nodes:
  - we'll not put much of a throttle on controlling jobs
  - doesn't scale the task impact in same way as we scale it for execution node capacity impact
  - in k8s, the whole way we are calculating capacity of control nodes is wrong https://github.com/ansible/awx/issues/11640 so a tiny constant impact won't show much improvement re; not overloading conttrol plane jobs

kdelee: I have addressed this on hybrid nodes in https://github.com/ansible/awx/pull/11651/commits/1fe442270db44279fe3c64cbcdcbd1e20b959709

awx/main/models/ha.py
```diff
@@ -145,7 +146,19 @@ def get_absolute_url(self, request=None):
...
+            execution_capacity_consumed = sum(
+                x.task_impact for x in UnifiedJob.objects.filter(Q(execution_node=self.hostname) & Q(status__in=('running', 'waiting')))
+            )
+        if self.node_type in ('hybrid', 'control'):
+            control_capacity_consumed = sum(
+                settings.AWX_CONTROL_NODE_TASK_IMPACT
+                for x in UnifiedJob.objects.filter(Q(controller_node=self.hostname) & Q(status__in=('running', 'waiting')))
+            )
+        logger.debug(f"{execution_capacity_consumed} execution capacity and {control_capacity_consumed} control capacity consumed on {self.hostname}")
+        return execution_capacity_consumed + control_capacity_consumed
```
AlanCoding, line 160: Okay, I think this looks valid now. I need to think about performance, but this at least seems correct.

awx/main/scheduler/task_manager_instances.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+class InstanceGroupCapacityGraph:
+    def __init__(self, task_manager_instances):
+        self.ig_capacity_graph = dict()
+        breakdown = False
+        for rampart_group in InstanceGroup.objects.prefetch_related('instances'):
+            InstanceGroupManager.zero_out_group(self.ig_capacity_graph, rampart_group.name, breakdown)
+            # Didn't move the init of DependencyGraph to InstanceGroupManager because of circular import
+            self.ig_capacity_graph[rampart_group.name]['dependency_graph'] = DependencyGraph()
```
AlanCoding: I'm trying to figure out, what is a dependency graph, and why does an instance group need to have one attached?

I can see why there is a global dependency graph for jobs in general. This dictates blocking logic, which I see breadcrumbs of. But why put it here? What connection does dependency / blocking logic have to instance groups?

awx/main/scheduler/task_manager_instances.py
```diff
@@ -0,0 +1,213 @@
...
+            for instance in rampart_group.instances.filter(enabled=True).order_by('hostname'):
```
AlanCoding: I'm never entirely clear on behavior with experimenting, but if you do `rampart_group.instances.filter(enabled=True).order_by('hostname')`, do you hit the cache populated by `InstanceGroup.objects.prefetch_related('instances')`?

With the complexity here... I don't think so. If you want enabled instances ordered by hostname, then you need to do that in python. The thing I know you can safely reference without performance penalties is `rampart_group.instances.all()`

AlanCoding:
I feel like this is getting to be kind of nuts. I regret that I introduced things like `consumed_control_capacity`, and seeing the lengths gone to here, I want to remove it.

You need to assign a pending job's instances, `control_node` in particular. What I would like to see happen is:
 - from whatever in-memory data structure you have, get the "controlplane" group. This is the group you select the node from.
 - based on available capacity of instances in that group, pick the best node, or do not pick any node

awx/main/scheduler/task_manager.py
```diff
@@ -511,39 +444,83 @@ def process_pending_tasks(self, pending_tasks):
                 continue
 
             for rampart_group in preferred_instance_groups:
```
AlanCoding: Even before this line, I think we can assign `controller_node`. I mean, not even in the loop. We don't need to consider the preferred instance groups, that's just over-complicating the problem. All jobs, project updates, container group jobs, regular jobs, all need a control node. If a control node can't be found, you can't dispatch _any_ jobs. You might as well stop processing pending tasks in general.

awx/main/scheduler/task_manager_instances.py
```diff
@@ -0,0 +1,236 @@
...
+
+        We can add/remove the node to update object in PrioritizedNodes to have right capacity.
+        """
+        self.nodes.remove(node)
+        self.nodes.add(node)
+
+    def best_node(self):
+        heap = list(self.nodes)
+        heapq.heapify(heap)
+        return heapq.nlargest(1, heap).pop()
```
AlanCoding: I would really prefer to keep the paradigms we have of

1. try to find a node where the job "fits"
2. find largest idle instance

I don't see why that won't work for control nodes as well. You don't need to reuse the existing logic for this, you could write new methods that do this, but the selection algorithm is something I would prefer to not alter in this refacoring.

awx/main/scheduler/task_manager_instances.py
```diff
@@ -0,0 +1,236 @@
...
+    def __init__(self):
+        realinstances = Instance.objects.filter(hostname__isnull=False, enabled=True).exclude(node_type='hop').prefetch_related('rampart_groups')
+        self.instances_partial = dict()
+        self.control_nodes = PrioritizedNodes()
+
+        for instance in realinstances:
+            self.instances_partial[instance.hostname] = SimpleNamespace(
+                obj=instance,
+                node_type=instance.node_type,
+                remaining_capacity=instance.remaining_capacity,
```
AlanCoding: There are going to be :dragon: s here

https://github.com/ansible/awx/blob/2de5ffc8d94d592498c32b8c3d68a82470d57763/awx/main/models/ha.py#L146-L152

You're using a property on the Instance model which is _very_ inefficient. It queries the list of jobs anew, which is really really counter to the purpose of the task manager. The task manager has complex logic to load instances, instance groups, and jobs. It is _most_ important that we don't re-load jobs multiple times. You can't just create a new data structure for instance capacity because the hard part is integrating it with the list of jobs that the task manager fetches.

shanemcd:
I am growing a bit concerned with the scope of this PR. It's not clear to me which parts are necessary, and what is just refactoring. 

This is a very critical part of the application. Any new patterns or large refactoring should be done in isolation so that we can more easily observe the impact on performance.

I would like to take a step back and reassess the problem we are trying to solve and see if there is a more direct path that does not involve such a large diff.

awx/main/scheduler/task_manager.py
```diff
@@ -509,50 +442,91 @@ def process_pending_tasks(self, pending_tasks):
...
+            controller_node = self.task_manager_instances.best_control_node_candidate()
+            task.controller_node = controller_node
+            if not controller_node:
+                found_acceptable_queue = False
+                # There is not enough control capacity, we can't run any more tasks. Stop processing pending tasks.
+                logger.warning("No control plane nodes available, will stop processing pending tasks")
+                # TODO: if we break here, we don't get to do
+                # task.log_lifecycle("needs_capacity")
+                # for all the tasks
+                break
```
AlanCoding, line 456: I know I said you should `break` here, but you can't because it would halt processing of workflow jobs. Best to `continue` instead.

awx/main/scheduler/task_manager.py
```diff
@@ -509,50 +442,91 @@ def process_pending_tasks(self, pending_tasks):
...
+
+            if isinstance(task, ProjectUpdate):
+                controller_node = self.task_manager_instances.assign_task_to_control_node(task, self.ig_capacity_graph, controller_node)
```
AlanCoding: You already picked this, I don't see any need to pick it again. Can't you just delete this line?

awx/main/scheduler/task_manager_instances.py
```diff
@@ -0,0 +1,252 @@
...
+        return largest_instance
+
+    @staticmethod
+    def fit_task_to_most_remaining_capacity_instance(task, instances):
+        instance_most_capacity = None
+        for i in instances:
+            if i.node_type not in (task.capacity_type, 'hybrid'):
+                continue
+            if i.remaining_capacity >= task.task_impact and (
+                instance_most_capacity is None or i.remaining_capacity > instance_most_capacity.remaining_capacity
```
AlanCoding: Instead of considering `i.remaining_capacity >= task.task_impact` as the criteria here, I think you should consider the criteria to be `i.remaining_capacity >= task.task_impact + settings.AWX_CONTROL_NODE_TASK_IMPACT` for hybrid nodes.

This should be actually pretty easy to do, although you'll be adding a little bit more python code to this method. This avoids any situation where we "fit" something but it won't actually fit because the control capacity consumption pushes it over.

This is a key change that I alluded to but probably didn't communicate effectively. If it doesn't make sense, let me know. It's really important for what needs to happen here.

awx/main/scheduler/task_manager.py
```diff
@@ -597,27 +571,23 @@ def reap_jobs_from_orphaned_instances(self):
...
+            # to compare against a list of current instance hostnames
+            if (
+                settings.IS_K8S
+                and j.is_container_group_task
+                and not isinstance(j, WorkflowJob)
+                and not j.controller_node
+                and j.status in ['waiting', 'running']
+            ):
+                logger.error(f'Job {j.id} has no control node, but is in status {j.status}. The control node pod must have been lost; reaping {j.log_format}')
+                reap_job(j, 'error')
```
AlanCoding, line 571: Reaping a job at this location seems very suspect to me. How could the control node have gone away? You just ran in a loop to select it. This doesn't add up to me at all.

awx/main/scheduler/task_manager_instances.py
rebeccahhh: Is there a technical reason for your preference?

awx/main/scheduler/task_manager_instances.py
rebeccahhh: This replaces the graphing on line 106 of `awx/main/scheduler/task_manager.py` (the call to this is on line 60).
https://github.com/ansible/awx/blob/devel/awx/main/scheduler/dependency_graph.py <- file the function comes from
All that to say, this paradigm existed before this refactor.

awx/main/scheduler/task_manager_instances.py
kdelee: I really don't think the order matters, this is just what the code was previously doing. I'll experiment w/o the order_by

awx/main/scheduler/task_manager.py
kdelee: doing this now

awx/main/scheduler/task_manager_instances.py
```diff
@@ -0,0 +1,204 @@
...
+                    best_node = self.nodes[nodes][instance_group].best_node()
+                    if best_node:
+                        break
+            if best_node and best_node.jobs_running == 0:
+                largest_instace = best_node
+        if largest_instance:
+            return largest_instance.hostname
+        return None
+
+    def fit_task_to_most_remaining_capacity_instance(self, task, instance_group, capacity_type, impact=None):
```
AlanCoding, line 149: After tinkering with this a bit today, I have to say that the _old_ `fit_task_to_most_remaining_capacity_instance` method was _beautifully_ isolated and easy to read. It took in 2 arguments - a job, and a list of instances. It was a _static_ method, meaning that it didn't require initializing a class or anything. It got the `capacity_type` from the task itself.

Was this method _in the right place_? No. Attaching it to the manager was confusing. IMO, a better place would be a new module with shared methods, like `awx/main/scheduler/fitting.py`, where it has nothing other than shared static methods.

I'm pro-refactoring, but I'm reading down into this method and seeing new loop structures, and it doesn't seem to be getting any more simple. A list of instances to consider is simple.

kdelee:
Closing this since the refactor is overshadowing the main point, which is to select the control node earlier and based on capacity and track consumed capacity for control nodes.

Replacement: https://github.com/ansible/awx/pull/11665

AlanCoding:
Our main issue for task manager refactoring is https://github.com/ansible/awx/issues/11306

I think the sub-components of that issue are still wildly unfocused but we have discussed several concrete actions that we _should_ and _need_ to take here. As I follow up from this, my goal is to address those quality points @kdelee discovered through our process so that we don't lose the important things in a sea of non-specific issues or giant code diffs.

Issue #11425, opened 2021-12-06
Control node inside instance group- enable ability for user to do health check from list view


akus062381:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

When a user visits the instance tab inside of an instance group, and there is a control node listed there, the user cannot click the check box to run a health check. 

### AWX version

devel

### Installation method

docker on linux

### Modifications

no

### Steps to reproduce

Create a build with at least one control node. Access the instance tab inside of the control instance group. Make sure that control node is associated with that instance group, and if it isn't, click associate to do so.
On the list view, you will notice that you cannot click the check box.

### Expected results

user would visit the instance tab inside of an instance group and would be able to click the check box to be able to run a health check from the list view. (Disassociate button would remain grayed out for this node type).

### Actual results

The checkbox for the control node is grayed out for all available button actions. The user has to access the details page of the instance if they want to run a health check.

### Additional information

![image](https://user-images.githubusercontent.com/63685497/144908480-c383ba41-e297-478c-84e1-b3d39fd3f715.png)

PR #11653, opened 2022-01-31
Adds top level instances list

AlexSCorey:
msg: "Adds a top level instances list for system admins to view."

##### SUMMARY
This addresses #11588 and #11425

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
Instances

### DIFF
--- a/awx/ui/src/components/HealthCheckButton/HealthCheckButton.js
+++ b/awx/ui/src/components/HealthCheckButton/HealthCheckButton.js
@@ -1,0 +1,62 @@
+import React from 'react';
+import { Plural, t } from '@lingui/macro';
+import { Button, DropdownItem, Tooltip } from '@patternfly/react-core';
+import { useKebabifiedMenu } from 'contexts/Kebabified';
+
+function HealthCheckButton({ isDisabled, onClick, selectedItems }) {
+  const { isKebabified } = useKebabifiedMenu();
+  const hopNodeSelected =
+    selectedItems.filter((instance) => instance.node_type === 'hop').length > 0;
+  const hasSelectedItems = selectedItems.length > 0;
+
+  const buildTooltip = () => {
+    if (hopNodeSelected) {
+      return (
+        <Plural
+          value={hopNodeSelected}
+          one="Cannot run health check on a hop node.  Deselect the hop node to run a health check."
+          other="Cannot run health check on hop nodes.  Deselect the hop nodes to run health checks."
+        />
+      );
+    }
+    return selectedItems.length ? (
+      <Plural
+        value={selectedItems.length}
+        one="Click to run a health check on the selected instance."
+        other="Click to run a health check on the selected instances."
+      />
+    ) : (
+      t`Select an instance to run a health check.`
+    );
+  };
+
+  if (isKebabified) {
+    return (
+      <Tooltip data-cy="healthCheckTooltip" content={buildTooltip()}>
+        <DropdownItem
+          key="approve"
+          isDisabled={hopNodeSelected || isDisabled || !hasSelectedItems}
+          component="button"
+          onClick={onClick}
+          ouiaId="health-check"
+        >
+          {t`Health Check`}
+        </DropdownItem>
+      </Tooltip>
+    );
+  }
+  return (
+    <Tooltip data-cy="healthCheckTooltip" content={buildTooltip()}>
+      <div>
+        <Button
+          isDisabled={hopNodeSelected || isDisabled || !hasSelectedItems}
+          variant="secondary"
+          ouiaId="health-check"
+          onClick={onClick}
+        >{t`Health Check`}</Button>
+      </div>
+    </Tooltip>
+  );
+}
+
+export default HealthCheckButton;
--- a/awx/ui/src/components/HealthCheckButton/HealthCheckButton.js
+++ b/awx/ui/src/components/HealthCheckButton/HealthCheckButton.js
@@ -5,37 +5,26 @@ import { useKebabifiedMenu } from 'contexts/Kebabified';
 
 function HealthCheckButton({ isDisabled, onClick, selectedItems }) {
   const { isKebabified } = useKebabifiedMenu();
-  const hopNodeSelected =
-    selectedItems.filter((instance) => instance.node_type === 'hop').length > 0;
-  const hasSelectedItems = selectedItems.length > 0;
 
-  const buildTooltip = () => {
-    if (hopNodeSelected) {
-      return (
-        <Plural
-          value={hopNodeSelected}
-          one="Cannot run health check on a hop node.  Deselect the hop node to run a health check."
-          other="Cannot run health check on hop nodes.  Deselect the hop nodes to run health checks."
-        />
-      );
-    }
-    return selectedItems.length ? (
+  const selectedItemsCount = selectedItems.length;
+
+  const buildTooltip = () =>
+    selectedItemsCount ? (
       <Plural
-        value={selectedItems.length}
+        value={selectedItemsCount}
         one="Click to run a health check on the selected instance."
         other="Click to run a health check on the selected instances."
       />
     ) : (
       t`Select an instance to run a health check.`
     );
-  };
 
   if (isKebabified) {
     return (
       <Tooltip data-cy="healthCheckTooltip" content={buildTooltip()}>
         <DropdownItem
           key="approve"
-          isDisabled={hopNodeSelected || isDisabled || !hasSelectedItems}
+          isDisabled={isDisabled || !selectedItemsCount}
           component="button"
           onClick={onClick}
           ouiaId="health-check"
@@ -49,7 +38,7 @@ function HealthCheckButton({ isDisabled, onClick, selectedItems }) {
     <Tooltip data-cy="healthCheckTooltip" content={buildTooltip()}>
       <div>
         <Button
-          isDisabled={hopNodeSelected || isDisabled || !hasSelectedItems}
+          isDisabled={isDisabled || !selectedItemsCount}
           variant="secondary"
           ouiaId="health-check"
           onClick={onClick}
--- a/awx/ui/src/routeConfig.js
+++ b/awx/ui/src/routeConfig.js
@@ -8,6 +8,7 @@ import Credentials from 'screens/Credential';
 import Dashboard from 'screens/Dashboard';
 import ExecutionEnvironments from 'screens/ExecutionEnvironment';
 import Hosts from 'screens/Host';
+import Instances from 'screens/Instances';
 import InstanceGroups from 'screens/InstanceGroup';
 import Inventory from 'screens/Inventory';
 import ManagementJobs from 'screens/ManagementJob';
@@ -130,6 +131,11 @@ function getRouteConfig(userProfile = {}) {
           title: <Trans>Instance Groups</Trans>,
           path: '/instance_groups',
           screen: InstanceGroups,
+        },
+        {
+          title: <Trans>Instances</Trans>,
+          path: '/instances',
+          screen: Instances,
         },
         {
           title: <Trans>Applications</Trans>,
--- a/awx/ui/src/screens/InstanceGroup/Instances/InstanceList.js
+++ b/awx/ui/src/screens/InstanceGroup/Instances/InstanceList.js
@@ -150,6 +150,7 @@ function InstanceList() {
         mergeParams(params, {
           ...{ not__rampart_groups__id: instanceGroupId },
           ...{ not__node_type: 'control' },
+          ...{ not__node_type: 'hop' },
         })
       ),
     [instanceGroupId]
--- a/awx/ui/src/screens/InstanceGroup/Instances/InstanceList.js
+++ b/awx/ui/src/screens/InstanceGroup/Instances/InstanceList.js
@@ -88,9 +88,13 @@ function InstanceList() {
     useCallback(async () => {
       await Promise.all(selected.map(({ id }) => InstancesAPI.healthCheck(id)));
       fetchInstances();
-      clearSelected();
-    }, [selected, clearSelected, fetchInstances])
+    }, [selected, fetchInstances])
   );
+
+  const handleHealthCheck = async () => {
+    await fetchHealthCheck();
+    clearSelected();
+  };
 
   const {
     isLoading: isDisassociateLoading,
@@ -231,7 +235,7 @@ function InstanceList() {
               />,
               <HealthCheckButton
                 isDisabled={!canAdd}
-                onClick={fetchHealthCheck}
+                onClick={handleHealthCheck}
                 selectedItems={selected}
               />,
             ]}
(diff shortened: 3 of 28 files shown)


nixocio (changes requested):
@AlexSCorey, if we are using the same code on `InstanceDetail.js` as the ones as part of the instance groups, we should move the code for a common place, and import here, and there.
Right now,  we will have to keep those 2 components in sync, if there is a bug, that will lead to extra fixes.
We already use this pattern for a couple of components, see the `JobList` as an example.

tiagodread:
- Health check on kebabified menu showing as a button
image

- Breadcumb and Health Check tooltip translation not showing
- The fields `Policy Type`, `Running Jobs` and `Total Jobs` doesn't make sense for this kind of node (hop) (we show again in expanded view)
image

- Clicking on Activity Stream button from the top level instance list navigates to `/#/activity_stream?type=instances` with the message **No Events Found** even having events generated, clicking on  **Dashboard (all activity)** it's possible to see them. The correct URL should be `/#/activity_stream?type=instance`
image

- In the API the field `   "last_health_check": null,` is not set for hop nodes. Which information are we consuming to show `Healthy` in the UI?
image

AlexSCorey:
> @AlexSCorey, if we are using the same code on `InstanceDetail.js` as the ones as part of the instance groups, we should move the code for a common place, and import here, and there. Right now, we will have to keep those 2 components in sync, if there is a bug, that will lead to extra fixes. We already use this pattern for a couple of components, see the `JobList` as an example.

I did consider refactoring the existing Instance List that lives inside Instance Groups, but this new list is quite different.  Some differences are:

- New List shows all node types, which add a whole layer of complexity that Instance Groups list does not have to deal with.
- New List doesn't allow associate/disassociate to an instance group.
- The new list shows node type on the list item, where as old list shows it in the expanded view
- Functionality on this list is likely to continue to diverge from the nested instances list in future releases.


I also decided not to refactor following the existing pattern of Inventories and nested Inventories, and following the WET (write everything twice) practice.  

To your concern about keeping the lists in sync:  These lists are already a bit out of sync by design because they have different use cases.
@nixocio

nixocio:
Thanks for clarifying @AlexSCorey

AlexSCorey:
> * Health check on kebabified menu showing as a button
>   image.  Good find.  This is also an issue on the instances list inside the instance group.  I'll fix it there too.
> * Breadcumb and Health Check tooltip translation not showing

This should not have been marked for translation as that value is the host name-generated by the customer
> * The fields `Policy Type`, `Running Jobs` and `Total Jobs` doesn't make sense for this kind of node (hop) (we show again in expanded view)
>   image

Fair point
> * Clicking on Activity Stream button from the top level instance list navigates to `/#/activity_stream?type=instances` with the message **No Events Found** even having events generated, clicking on  **Dashboard (all activity)** it's possible to see them. The correct URL should be `/#/activity_stream?type=instance`
>   image

Activity stream is not available for instances right now so I'll remove the button there
> * In the API the field `   "last_health_check": null,` is not set for hop nodes. Which information are we consuming to show `Healthy` in the UI?
>   image

For the last health check date/time value we use the field you show.  For the status label we check for `errors:""`.

tiagodread:
@AlexSCorey 
- The Activity stream events for instance are available in here `/#/activity_stream?type=instance`, so I think we should keep the shortcut

- Health Check tooltip translation not showing
Captura de tela de 2022-02-07 17-22-59

- Hitting the Health check, there is no way to know that a request is happening under de hood, we should display the loading spinner when the request for `/health_check/` start

- Sortting the instances by `Status` or `Capacity Adjustment` then, toogling the field `Enabled/Disabled` for one row, is making other change as well

https://user-images.githubusercontent.com/12464858/152869116-8efcd252-8353-4a80-9cc0-cb6ab580066e.mp4

The error is showing on console:
```
React does not recognize the `ouiaId` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `ouiaid` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```
and
```
Unchecked runtime.lastError: The message port closed before a response was received.
```

awx/ui/src/components/HealthCheckButton/HealthCheckButton.js
```diff
@@ -0,0 +1,62 @@
...
+  const hopNodeSelected =
+    selectedItems.filter((instance) => instance.node_type === 'hop').length > 0;
+  const hasSelectedItems = selectedItems.length > 0;
+
+  const buildTooltip = () => {
+    if (hopNodeSelected) {
+      return (
+        <Plural
+          value={hopNodeSelected}
+          one="Cannot run health check on a hop node.  Deselect the hop node to run a health check."
```
nixocio, line 17:
```suggestion
          one="Cannot run health check on a hop node. Deselect the hop node to run a health check."
```

nixocio:
```suggestion
          one="Cannot run health check on a hop node. Deselect the hop node to run a health check."
```

awx/ui/src/components/HealthCheckButton/HealthCheckButton.js
```diff
@@ -0,0 +1,62 @@
...
+          other="Cannot run health check on hop nodes.  Deselect the hop nodes to run health checks."
```
nixocio:
```suggestion
          other="Cannot run health check on hop nodes. Deselect the hop nodes to run health checks."
```

awx/ui/src/components/HealthCheckButton/HealthCheckButton.js
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+import { Plural, t } from '@lingui/macro';
+import { Button, DropdownItem, Tooltip } from '@patternfly/react-core';
+import { useKebabifiedMenu } from 'contexts/Kebabified';
+
+function HealthCheckButton({ isDisabled, onClick, selectedItems }) {
+  const { isKebabified } = useKebabifiedMenu();
+  const hopNodeSelected =
+    selectedItems.filter((instance) => instance.node_type === 'hop').length > 0;
```
nixocio:
```js
const hopNodeSelected = selectedItems.filter((instance) => instance.node_type === 'hop')
const isHopNodeSelected = hopNodeSelected.length > 0;
```

then you could pass `hopNodeSelected.length` to the value on `Plural`.  Names on suggestion could be better.

awx/ui/src/components/HealthCheckButton/HealthCheckButton.js
```diff
@@ -0,0 +1,62 @@
...
+  const hasSelectedItems = selectedItems.length > 0;
+
+  const buildTooltip = () => {
+    if (hopNodeSelected) {
+      return (
+        <Plural
+          value={hopNodeSelected}
```
nixocio:
```suggestion
          value={hopNodeSelected}
```

hopNodeSelected is a boolean,

awx/ui/src/components/HealthCheckButton/HealthCheckButton.js
```diff
@@ -0,0 +1,62 @@
...
+          one="Cannot run health check on a hop node.  Deselect the hop node to run a health check."
+          other="Cannot run health check on hop nodes.  Deselect the hop nodes to run health checks."
+        />
+      );
+    }
+    return selectedItems.length ? (
```
nixocio: 👍

awx/ui/src/screens/InstanceGroup/Instances/InstanceList.js
```diff
@@ -83,11 +80,16 @@ function InstanceList() {
...
+    error: healthCheckError,
+    request: fetchHealthCheck,
+    isLoading: isHealthCheckLoading,
+  } = useRequest(
     useCallback(async () => {
       await Promise.all(selected.map(({ id }) => InstancesAPI.healthCheck(id)));
       fetchInstances();
-    }, [selected, fetchInstances])
+      clearSelected();
+    }, [selected, clearSelected, fetchInstances])
```
nixocio: Is there a particular reason for `clearSelected` be part of dependency array? I looked at other lists and in general we have this pattern.

```js 
const handleHealthCheck = async () => {

await fetchHealthCheck();
clearSelected();

}

File: awx/ui/src/screens/Instances/Instance.js [hunk 7]
```diff
@@ -0,0 +1,53 @@
...
+  return (
+    <PageSection>
+      <Card>
+        <RoutedTabs tabsArray={tabsArray} />
+        <Switch>
+          <Redirect from="/instances/:id" to="/instances/:id/details" exact />
+          <Route path="/instances/:id/details" key="details">
+            <InstanceDetail setBreadcrumb={setBreadcrumb} />
+          </Route>
+          ,
```
nixocio:
```suggestion
         
```

awx/ui/src/screens/Instances/InstanceList/InstanceList.js
```diff
@@ -0,0 +1,185 @@
...
+    isLoading: isHealthCheckLoading,
+  } = useRequest(
+    useCallback(async () => {
+      await Promise.all(
+        selected
+          .filter(({ node_type }) => node_type !== 'hop')
+          .map(({ id }) => InstancesAPI.healthCheck(id))
+      );
+      fetchInstances();
+      clearSelected();
```
nixocio: Same comment as above.

nixocio (changes requested):
Hey @AlexSCorey,

I'm still looking through, but here are some initial thoughts.

awx/ui/src/components/HealthCheckButton/HealthCheckButton.js
AlexSCorey: This is a good catch, but I actually need to pass the selectedItems.length to the plural here instead of the `hopeNodeSelected.length`. I've refactored this work a bit to hopefully make it more clean.

awx/ui/src/components/HealthCheckButton/HealthCheckButton.js
AlexSCorey: Good catch.  This should be selectedItems.length

awx/ui/src/components/HealthCheckButton/HealthCheckButton.js
```diff
@@ -0,0 +1,62 @@
...
+import { useKebabifiedMenu } from 'contexts/Kebabified';
+
+function HealthCheckButton({ isDisabled, onClick, selectedItems }) {
+  const { isKebabified } = useKebabifiedMenu();
+  const hopNodeSelected =
+    selectedItems.filter((instance) => instance.node_type === 'hop').length > 0;
+  const selectedItemsCount = selectedItems.length;
+
+  const buildTooltip = () => {
+    if (hopNodeSelected) {
```
nixocio: hey @AlexSCorey, if one hop node is selected will display the tooltip, but we can have one hop node selected an many others of different types what leads `selectedItemsCount` > 1 and display `other` message on Plural. Even with only one hop node selected.

awx/ui/src/components/HealthCheckButton/HealthCheckButton.js
```diff
@@ -0,0 +1,62 @@
...
+      return (
+        <Plural
+          value={selectedItemsCount}
```
nixocio: This should be the amount of hop nodes selected, right?

awx/ui/src/routeConfig.js
```diff
@@ -131,6 +132,11 @@ function getRouteConfig(userProfile = {}) {
           path: '/instance_groups',
           screen: InstanceGroups,
         },
+        {
+          title: <Trans>Instances</Trans>,
+          path: '/instances',
```
nixocio, line 137: What user profile RBAC should see this list on the side bar? I logged in as Org Admin, and I got back an empty list. 

I could not find `instance_admin` as a role. Just raising the question.

image

awx/ui/src/screens/InstanceGroup/Instances/InstanceListItem.js
```diff
@@ -197,7 +196,8 @@ function InstanceListItem({
         <Td colSpan={7}>
           <ExpandableRowContent>
             <DetailList>
-              <Detail label={t`Node Type`} value={instance.node_type} />
+              <Detail value={instance.jobs_running} label={t`Running Jobs`} />
```
nixocio: `data-cy`for Detail?

awx/ui/src/screens/Instances/Instance.js
```diff
@@ -0,0 +1,52 @@
...
+              )}
+            </ContentError>
+          </Route>
+        </Switch>
+      </Card>
+    </PageSection>
+  );
+}
+
+export { Instance as _Instance };
```
nixocio:
```suggestion

```

awx/ui/src/screens/Instances/InstanceList/InstanceListItem.js
```diff
@@ -0,0 +1,248 @@
...
+`;
+
+const SliderForks = styled.div`
+  flex-grow: 1;
+  margin-right: 8px;
+  margin-left: 8px;
+  text-align: center;
+`;
+
+function computeForks(memCapacity, cpuCapacity, selectedCapacityAdjustment) {
```
nixocio: This function could be moved an utils file, since it is even defined outside the components in which is used.

awx/ui/src/screens/Instances/InstanceList/InstanceListItem.js
```diff
@@ -0,0 +1,248 @@
...
+      {!isHopNode && (
+        <Tr
+          ouiaId={`instance-row-${instance.id}-expanded`}
+          isExpanded={isExpanded}
+        >
+          <Td colSpan={2} />
+          <Td colSpan={7}>
+            <ExpandableRowContent>
+              <DetailList>
+                <Detail value={instance.jobs_running} label={t`Running Jobs`} />
```
nixocio: It would be nice to add `data-cy` for all those details.

nixocio (changes requested):
hey @AlexSCorey, good job.

Minor comments.

awx/ui/src/routeConfig.js
AlexSCorey: Will follow up issue for this and mesh visualizer, and instance groups to address RBAC

nixocio: @AlexSCorey https://github.com/ansible/awx/issues/11721

nixocio (approved):

nixocio (approved):
Good job @AlexSCorey

tiagodread:
> @AlexSCorey
> 
> * The Activity stream events for instance are available in here `/#/activity_stream?type=instance`, so I think we should keep the shortcut

Solved!
> * Health Check tooltip translation not showing

Not solved
>   Captura de tela de 2022-02-07 17-22-59
> * Hitting the Health check, there is no way to know that a request is happening under de hood, we should display the loading spinner when the request for `/health_check/` start
> * Sortting the instances by `Status` or `Capacity Adjustment` then, toogling the field `Enabled/Disabled` for one row, is making other change as well
> 
>  instances-2022-02-07_17.44.25.mp4 

Not solved

tiagodread:
@AlexSCorey was the intention to show the tooltip in the `name` column?
image

tiagodread (approved):
Thanks @AlexSCorey for addressing all feedback, the broken tests aren't related to this change and I'll be fixing them

PR #11654, opened 2022-01-31
Django 3.2 upgrade

jbradberry:
##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API


awx/main/tasks/system.py
```diff
@@ -121,6 +128,105 @@ def inform_cluster_of_shutdown():
...
+            for colname in columns:
+                f = model._meta.get_field(colname)
+                _, _, args, kwargs = f.deconstruct()
+                kwargs['null'] = True
+                new_f = f.__class__(*args, **kwargs)
+                new_f.set_attributes_from_name(f'_{colname}')
+
+                schema_editor.add_field(model, new_f)
+
+                # Create a trigger to make sure new data automatically gets put in both fields.
```
shanemcd: This is cool 👍

rebeccahhh (approved):

awx/main/models/__init__.py
```diff
@@ -97,6 +98,93 @@
 User.add_to_class('accessible_objects', user_accessible_objects)
 
 
+def convert_jsonfields_to_jsonb():
```
AlanCoding, line 101: Where does this method get called from? The presentation of the large diff is messing me up.

And then after this gets called, it delays another method.

AlanCoding:
Conflict was introduced by https://github.com/ansible/awx/pull/11855

fosterseth (approved):

awx/main/tasks/system.py
```diff
@@ -78,6 +81,9 @@
 
 def dispatch_startup():
     startup_logger = logging.getLogger('awx.main.tasks')
+
+    convert_jsonfields_to_jsonb()
```
AlanCoding, line 85: Is this going to be okay to run on clusters? This will be running multiple versions of the task all at the same time right? Is that okay?

jbradberry: I _think_ this is fine.  If you look at the `migrate_json_fields` task, which is ultimately what is called, I set an advisory lock in the database and short-circuit out if the lock is already taken.

AlanCoding: oh right :100:

AlanCoding (approved):

Zokormazo (approved):

AlanCoding:
awx-operator check's last fail appears to be a 502, so flake.

> #3 ERROR: unexpected status code [manifests stream9]: 502 Bad Gateway", "------", " > [internal] load metadata for quay.io/centos/centos:stream9:", "------", "failed to solve with frontend dockerfile.v0: failed to create LLB definition: unexpected status code [manifests stream9]: 502 Bad Gateway"

awx/conf/apps.py
```diff
@@ -12,6 +14,9 @@ class ConfConfig(AppConfig):
 
     def ready(self):
         self.module.autodiscover()
-        from .settings import SettingsWrapper
 
-        SettingsWrapper.initialize()
+        if not set(sys.argv) & {'migrate', 'check_migrations'}:
```
AlanCoding, line 18: If you're filing follow issues, then let's get one to remove some hacks in `awx/conf/settings.py`. With this in place, I don't feel like that should ever be checking `sys.argv`. I quickly searched, and didn't find any sign of that file in the current diff.

jbradberry: I'm not sure I know what hacks you're referring to.

AlanCoding: https://github.com/ansible/awx/blob/4e97492ed624eeb847c96e09d65478726eb81036/awx/conf/settings.py#L84

jbradberry: Right, gotcha.  Yep, I'll file the ticket.

Issue #11600, opened 2022-01-24
Allow isolated path to be exposed in ocp/k8s as HostPath


tchellomello:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
This issue is complimentary of the PR https://github.com/ansible/awx/pull/11566 which allows to user to specify which mount point the isolated path will be mounted. 

Currently, there is no such mechanism to expose this path list into k8s/ocp.

tchellomello:
Started some work on this branch https://github.com/ansible/awx/tree/expose_isolate_path_k8s

Assuming this configuration:

![image](https://user-images.githubusercontent.com/809840/150865522-ca400703-c4a0-416c-8599-0c5756eee260.png)
![image](https://user-images.githubusercontent.com/809840/150865577-ceacd758-6072-4d5d-8ce4-44b7cf7edc73.png)

Then once the playbook runs, we should see a POD spec like below:

```yaml
kubectl get pods -o yaml -n ansible-devel automation-job-40-h9qn4                                                                                         16:10:57
apiVersion: v1
kind: Pod
metadata:
  annotations:
    cni.projectcalico.org/containerID: 9208adcbfa5ae0410111b6f2199f04cb171205b0a6e343091309377420fd725e
    cni.projectcalico.org/podIP: 10.233.100.17/32
    cni.projectcalico.org/podIPs: 10.233.100.17/32
  creationTimestamp: "2022-01-24T21:13:26Z"
  generateName: automation-job-40-
  labels:
    ansible-awx: 2ab77134-e715-4458-af58-24eb187d2772
    ansible-awx-job-id: "40"
  name: automation-job-40-h9qn4
  namespace: ansible-devel
  resourceVersion: "512525"
  uid: 603a0115-5ac9-4097-82ee-5ac567a84e0f
spec:
  automountServiceAccountToken: false
  containers:
  - args:
    - ansible-runner
    - worker
    - --private-data-dir=/runner
    image: quay.io/ansible/awx-ee:latest
    imagePullPolicy: Always
    name: worker
    resources:
      requests:
        cpu: 250m
        memory: 100Mi
    stdin: true
    stdinOnce: true
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    volumeMounts:
    - mountPath: /mnt2
      name: volume-0
      readOnly: true
    - mountPath: /mnt3
      name: volume-1
      readOnly: true
    - mountPath: /mnt4
      name: volume-2
      readOnly: true
[...]

PR #11659, opened 2022-02-01
Allow isolated paths as hostPath volume @ k8s/ocp/container groups

tchellomello:
##### TODO
* [x] - Add UI test (@marshmalien++)

##### SUMMARY
Allow isolated paths as hostPath volume @ k8s/ocp/container groups

related: https://github.com/ansible/awx/issues/11600

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
devel
```

##### ADDITIONAL INFORMATION
image

```yaml
# kubectl -n ansible-devel -o yaml get pod automation-job-27-n7fwb
apiVersion: v1
kind: Pod
metadata:
  annotations:
    cni.projectcalico.org/containerID: deae40ebe06e0749ca2534d7cdddeac57df94f45c7d5b43acd0ed17834eb75dc
    cni.projectcalico.org/podIP: 10.233.100.199/32
    cni.projectcalico.org/podIPs: 10.233.100.199/32
  creationTimestamp: "2022-02-01T18:16:50Z"
  generateName: automation-job-27-
  labels:
    ansible-awx: 6c9f8789-d573-4910-bcc6-0eabc88e7d1d
    ansible-awx-job-id: "27"
  name: automation-job-27-n7fwb
  namespace: ansible-devel
  resourceVersion: "2752667"
  uid: 4498a195-3163-44a5-9c68-334160463e86
spec:
  automountServiceAccountToken: false
  containers:
  - args:
    - ansible-runner
    - worker
    - --private-data-dir=/runner
    image: quay.io/ansible/awx-ee:latest
    imagePullPolicy: Always
    name: worker
    resources:
      requests:
        cpu: 250m
        memory: 100Mi
    stdin: true
    stdinOnce: true
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    volumeMounts:
    - mountPath: /mnt2
      name: volume-0
    - mountPath: /mnt
      name: volume-1
      readOnly: true
    - mountPath: /mnt3
      name: volume-2
      readOnly: true
  dnsPolicy: ClusterFirst
  enableServiceLinks: true
  nodeName: t470n1
  priority: 0
  restartPolicy: Never
  securityContext: {}
  serviceAccount: default
  serviceAccountName: default
  tolerations:
  - effect: NoExecute
    key: node.kubernetes.io/not-ready
    operator: Exists
    tolerationSeconds: 300
  - effect: NoExecute
    key: node.kubernetes.io/unreachable
    operator: Exists
    tolerationSeconds: 300
  volumes:
  - hostPath:
      path: /data-awx2
      type: ""
    name: volume-0
  - hostPath:
      path: /data-awx
      type: ""
    name: volume-1
  - hostPath:
      path: /not_valid
      type: ""
    name: volume-2
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2022-02-01T18:16:50Z"
    status: "True"
    type: Initialized
  - lastProbeTime: null
    lastTransitionTime: "2022-02-01T18:16:52Z"
    status: "True"
    type: Ready
  - lastProbeTime: null
    lastTransitionTime: "2022-02-01T18:16:52Z"
    status: "True"
    type: ContainersReady
  - lastProbeTime: null
    lastTransitionTime: "2022-02-01T18:16:50Z"
    status: "True"
    type: PodScheduled
  containerStatuses:
  - containerID: cri-o://00681d042fa22b3e423341e865bd12ecfed14b2ca9ee12acd0e77f2756feb646
    image: quay.io/ansible/awx-ee:latest
    imageID: quay.io/ansible/awx-ee@sha256:4c39d7bf4ad569b388a005fec5ca737a0519be153722b84c0107273f207d135d
    lastState: {}
    name: worker
    ready: true
    restartCount: 0
    started: true
    state:
      running:
        startedAt: "2022-02-01T18:16:51Z"
  hostIP: 192.168.111.34
  phase: Running
  podIP: 10.233.100.199
  podIPs:
  - ip: 10.233.100.199
  qosClass: Burstable
  startTime: "2022-02-01T18:16:50Z"
```


awx/main/conf.py
```diff
@@ -334,6 +334,19 @@
     category_slug='jobs',
 )
 
+register(
+    'AWX_MOUNT_ISOLATED_PATHS_ON_K8S',
```
AlanCoding, line 338: Is this _just_ on K8s?

Can't traditional installs also make use of container group jobs? Would this be usable in those situations as well?

tchellomello: This should work with container groups too. Maybe let's rename it to make more inclusive

fosterseth (approved):

awx/main/conf.py
tchellomello: At the tooltip, you can see the user will see the `Container Groups` text. Since this feature works for K8S installs and traditional installs with Container Groups, do you think we should rename it or doc string already should be sufficient?

AlanCoding: yeah, that works

awx/main/constants.py
```diff
@@ -88,7 +88,8 @@
 
 # :z option tells Podman that two containers share the volume content with r/w
 # :O option tells Podman to mount the directory from the host as a temporary storage using the overlay file system.
+# :ro or :rw option to mount a volume in read-only or read-write mode, respectively. By default, the volumes are mounted read-write.
 # see podman-run manpage for further details
 # /HOST-DIR:/CONTAINER-DIR:OPTIONS
-CONTAINER_VOLUMES_MOUNT_TYPES = ['z', 'O']
+CONTAINER_VOLUMES_MOUNT_TYPES = ['z', 'O', 'ro', 'rw']
```
shanemcd, line 94: Is there a reason we dont want to support [all of these](https://docs.podman.io/en/latest/markdown/podman-run.1.html#volume-v-source-volume-host-dir-container-dir-options)?

```
              • rw|ro

              • z|Z

              • [r]shared|[r]slave|[r]private[r]unbindable

              • [r]bind

              • [no]exec

              • [no]dev

              • [no]suid

              • [O]

              • [U]
```

awx/main/tasks/jobs.py
```diff
@@ -193,6 +194,13 @@ def build_execution_environment_params(self, instance, private_data_dir):
                 # Uppercase Z restricts access (in weird ways) to 1 container at a time
                 if this_path.count(':') == MAX_ISOLATED_PATH_COLON_DELIMITER:
                     src, dest, scontext = this_path.split(':')
+
+                    # scontext validation via performed via API, but since this can be overriden via settings.py
+                    # let's ensure scontext is one that we support
+                    if scontext not in CONTAINER_VOLUMES_MOUNT_TYPES:
```
shanemcd: I dont think `scontext` is accurate. Should this be something like `mount_option`?

awx/main/constants.py
tchellomello: I believe we are just extending support to the options with minimal impact or configuration required on the host itself and `harmless`  possible. 

For example, if we allow other types of propagations, changes like `mount --make-share <DIR>` might be required on the host itself. 

If we believe it's important to cover other scenarios, I would suggest that we can visit this on a 2nd pass and explore the pros and cons from configuration and security angles for each option.

awx/main/tasks/jobs.py
tchellomello: You got it

sarabrajsingh (approved):

rebeccahhh (approved):

shanemcd (approved):
+1, but let's get QE to look at this before merging.

tchellomello:
@CFSNM @Zokormazo  can you get yours :eyes:  on this?

ffirg (approved):

CFSNM (approved):

CFSNM (approved):
Launched YOLO job to run container group tests: (removed)

CFSNM:
The YOLO job has 2 failing tests, but they are related to job stucked running forever. I dont think this is related to this PR

Zokormazo:
I'm having hard time trying to figure out how is this going to work on OCP. We are allowing to bind a hostpath, do our customers have to sync a host dir between immutable rhcos nodes? Is this something supported? How is this going to work in real life? iirc, you don't even have ssh access to those nodes on a default ipi installation, you need to inyect the debug ssh key on the nodes

Won't be better to allow customers to define a RWX pvc for this mounts and get rid of being locked to worker nodes? I would say we want to continue running jobs on any of the k8s nodes and allowing to share this mount between those jobs at the same time and a hostpath does not achieve that if you don't sync the content between nodes. A PVC would not solve the issue between different container groups, but at least the jobs running on the same container group would be able to share the same mount point independently of the worker node where the job was scheduled

tchellomello:
> I'm having hard time trying to figure out how is this going to work on OCP. We are allowing to bind a hostpath, do our customers have to sync a host dir between immutable rhcos nodes? Is this something supported? How is this going to work in real life? iirc, you don't even have ssh access to those nodes on a default ipi installation, you need to inyect the debug ssh key on the nodes
> 
> Won't be better to allow customers to define a RWX pvc for this mounts and get rid of being locked to worker nodes? I would say we want to continue running jobs on any of the k8s nodes and allowing to share this mount between those jobs at the same time and a hostpath does not achieve that if you don't sync the content between nodes. A PVC would not solve the issue between different container groups, but at least the jobs running on the same container group would be able to share the same mount point independently of the worker node where the job was scheduled

@Zokormazo  this solution is not the ideal for `OCP/k8s` and I agree with you that we have to introduce a way to provide and mount the PVCs so the filesystem consistency happens across the board. 

This solution will only allow some cases like exposing some node-local files to the container itself and not much for the work data artifacts. For that, the current solution is to manually add the PVC and `mountPath` to the container `SPEC` but +1 we should allow the user to compose this via UI in the future.

tchellomello:
rebased

PR #11663, opened 2022-02-01
Update saml.md

srinathman:
- Updated link to python documentation
- Added instructions for superadmin permissions

- replace invalid parameters with new ones

Signed-off-by: Srinath M <srinath_man@yahoo.com>

##### ISSUE TYPE

 - Docs Pull Request


john-westcott-iv (approved):
@srinathman thanks for starting this. I reverted some of the changes you made to the SAML mapping fields around `saml_attr` and `saml_admin_attr` as they are intended to be place holders. You are correct that they will not work as in but what they should be replaced with depends on your IdP implementation. In addition, for the new SAML auditor/admin you can also do this mapping via IdP roles and also require a specific value on the attributes. I've added some more details around those and added a logic table for how the option behave together. Thank you again for your contribution, I'll get the PR merged shortly.

PR #11665, opened 2022-02-02
Consume control capacity

kdelee:
##### SUMMARY

Addresses https://github.com/ansible/awx/issues/10694
Replaces https://github.com/ansible/awx/pull/11651 

##### ISSUE TYPE

 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.5.2.dev67+gbcba14e53e
```

##### ADDITIONAL INFORMATION

This PR focuses exclusively on implementing https://github.com/ansible/awx/issues/10694

This implementation enforces the requirement for having all control and hybrid nodes be members of the `'controlplane'` instance group.

The tests did not previously comply with this requirement so I'm having to update a number of them. I have all but about 5 passing now. 

This PR:
  - selects a candidate controller node before we loop over instance groups
  - For non-container group jobs, if we end up selecting a hybrid node with enough capacity to do both the job and the control task, we prefer that node to do both
  - introduces setting `AWX_CONTROL_NODE_TASK_IMPACT` that is a constant integer (I use 5) amount of "task_impact" for when a node is the `controller_node` of a job
  - Selection of `task.execution_node` and `task.controller_node` and consumption of capacity in the in-memory capacity tracking happen before we go into `start_task`
- skip looping over `preferred_instance_groups` for project updates and system jobs. 

This PR does not:
- do any refactor of the in-memory capacity tracking mechanism.

### DIFF
--- a/awx/main/managers.py
+++ b/awx/main/managers.py
@@ -243,7 +243,13 @@ class InstanceGroupManager(models.Manager):
         for t in tasks:
             # TODO: dock capacity for isolated job management tasks running in queue
             impact = t.task_impact
-            if t.status == 'waiting' or not t.execution_node:
+            control_groups = []
+            if t.controller_node:
+                control_groups = instance_ig_mapping.get(t.controller_node, [])
+                if not control_groups:
+                    logger.warn(f"No instance group found for {t.controller_node}, capacity consumed may be innaccurate.")
+
+            if t.status == 'waiting' or (not t.execution_node and not t.is_container_group_task):
                 # Subtract capacity from any peer groups that share instances
                 if not t.instance_group:
                     impacted_groups = []
@@ -260,6 +266,12 @@ class InstanceGroupManager(models.Manager):
                     graph[group_name][f'consumed_{capacity_type}_capacity'] += impact
                     if breakdown:
                         graph[group_name]['committed_capacity'] += impact
+                for group_name in control_groups:
+                    if group_name not in graph:
+                        self.zero_out_group(graph, group_name, breakdown)
+                    graph[group_name][f'consumed_control_capacity'] += settings.AWX_CONTROL_NODE_TASK_IMPACT
+                    if breakdown:
+                        graph[group_name]['committed_capacity'] += settings.AWX_CONTROL_NODE_TASK_IMPACT
             elif t.status == 'running':
                 # Subtract capacity from all groups that contain the instance
                 if t.execution_node not in instance_ig_mapping:
@@ -271,6 +283,7 @@ class InstanceGroupManager(models.Manager):
                         impacted_groups = []
                 else:
                     impacted_groups = instance_ig_mapping[t.execution_node]
+
                 for group_name in impacted_groups:
                     if group_name not in graph:
                         self.zero_out_group(graph, group_name, breakdown)
@@ -279,6 +292,12 @@ class InstanceGroupManager(models.Manager):
                     graph[group_name][f'consumed_{capacity_type}_capacity'] += impact
                     if breakdown:
                         graph[group_name]['running_capacity'] += impact
+                for group_name in control_groups:
+                    if group_name not in graph:
+                        self.zero_out_group(graph, group_name, breakdown)
+                    graph[group_name][f'consumed_control_capacity'] += settings.AWX_CONTROL_NODE_TASK_IMPACT
+                    if breakdown:
+                        graph[group_name]['running_capacity'] += settings.AWX_CONTROL_NODE_TASK_IMPACT
             else:
                 logger.error('Programming error, %s not in ["running", "waiting"]', t.log_format)
         return graph
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -286,8 +286,10 @@ class TaskManager:
             # at this point we already have control/execution nodes selected for the following cases
             else:
                 task.instance_group = rampart_group
-                queue_submitted_to = task.execution_node if task.execution_node else rampart_group.name
-                logger.debug(f'Submitting job {task.log_format} to queue {queue_submitted_to} controlled by {task.controller_node}.')
+                execution_node_msg = f' and execution node {task.execution_node}' if task.execution_node else ''
+                logger.debug(
+                    f'Submitting job {task.log_format} controlled by {task.controller_node} to instance group {rampart_group.name}{execution_node_msg}.'
+                )
             with disable_activity_stream():
                 task.celery_task_id = str(uuid.uuid4())
                 task.save()
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -502,7 +502,7 @@ class TaskManager:
                 control_impact = settings.AWX_CONTROL_NODE_TASK_IMPACT
             control_instance = InstanceGroup.fit_task_to_most_remaining_capacity_instance(
                 task, self.graph[settings.DEFAULT_CONTROL_PLANE_QUEUE_NAME]['instances'], impact=control_impact, capacity_type='control'
-            ) or InstanceGroup.find_largest_idle_instance(self.graph[settings.DEFAULT_CONTROL_PLANE_QUEUE_NAME]['instances'], capacity_type='control')
+            )
             if not control_instance:
                 self.task_needs_capacity(task, tasks_to_update_job_explanation)
                 logger.debug(f"Skipping task {task.log_format} in pending, not enough capacity left on controlplane to control new tasks")
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -284,12 +284,10 @@ class TaskManager:
                 logger.debug('Transitioning %s to running status.', task.log_format)
                 schedule_task_manager()
             # at this point we already have control/execution nodes selected for the following cases
-            elif rampart_group.is_container_group:
-                task.instance_group = rampart_group
-                logger.debug('Submitting containerized {} to queue {}.'.format(task.log_format, task.execution_node))
             else:
                 task.instance_group = rampart_group
-                logger.debug('Submitting job {} to queue {} controlled by {}.'.format(task.log_format, task.execution_node, task.controller_node))
+                queue_submitted_to = task.execution_node if task.execution_node else rampart_group.name
+                logger.debug(f'Submitting job {task.log_format} to queue {queue_submitted_to} controlled by {task.controller_node}.')
             with disable_activity_stream():
                 task.celery_task_id = str(uuid.uuid4())
                 task.save()
@@ -461,6 +459,7 @@ class TaskManager:
     def process_pending_tasks(self, pending_tasks):
         running_workflow_templates = {wf.unified_job_template_id for wf in self.get_running_workflow_jobs()}
         tasks_to_update_job_explanation = []
+        controlplane_ig = InstanceGroup.objects.get(name='controlplane')
         for task in pending_tasks:
             if self.start_task_limit <= 0:
                 break
@@ -497,31 +496,30 @@ class TaskManager:
 
             task.controller_node = control_instance.hostname
 
-            for rampart_group in preferred_instance_groups:
-                if task.capacity_type == 'execution' and rampart_group.is_container_group:
-                    self.graph[rampart_group.name]['graph'].add_job(task)
-                    self.start_task(task, rampart_group, task.get_jobs_fail_chain(), None)
-                    found_acceptable_queue = True
+            # All task.capacity_type == 'control' jobs should run on control plane, no need to loop over instance groups
+            if task.capacity_type == 'control':
+                # This is a task that actually needs to run in the controlplane
+                is_idle = control_instance.jobs_running == 0
+                has_capacity_for_task_impact_and_control = control_instance.remaining_capacity >= (task.task_impact + settings.AWX_CONTROL_NODE_TASK_IMPACT)
+                if not has_capacity_for_task_impact_and_control and not is_idle:
+                    # As in other places, we accept an idle instance if the node with most capacity (the control node we already selected)
+                    # does not have enough capacity.
+                    logger.debug(f"Not enough control capacity on {control_instance} to run {task.log_format}")
                     break
+                task.execution_node = control_instance.hostname
+                control_instance.remaining_capacity = max(0, control_instance.remaining_capacity - (task.task_impact + settings.AWX_CONTROL_PLANE_TASK_IMPACT))
+                control_instance.jobs_running += 1
+                self.graph['controlplane']['graph'].add_job(task)
+                execution_instance = self.real_instances[control_instance.hostname]
+                self.start_task(task, controlplane_ig, task.get_jobs_fail_chain(), execution_instance)
+                found_acceptable_queue = True
+                continue
 
-                if task.capacity_type != 'execution' and rampart_group.is_container_group:
-                    # This is a task that actually needs to run in the controlplane, like a project update for a container_group job
-                    if (
-                        not control_instance.remaining_capacity >= task.task_impact + settings.AWX_CONTROL_NODE_TASK_IMPACT
-                        and not control_instance.jobs_running == 0
-                    ):
-                        # As in other places, we accept an idle instance if the node with most capacity (the control node we already selected)
-                        # does not have enough capacity.
-                        logger.debug(f"Not enough control capacity on {control_instance} to run {task.log_format}")
-                        break
-                    task.execution_node = control_instance
-                    control_instance.remaining_capacity = max(
-                        0, control_instance.remaining_capacity - (task.task_impact + settings.AWX_CONTROL_PLANE_TASK_IMPACT)
-                    )
+            for rampart_group in preferred_instance_groups:
+                if rampart_group.is_container_group:
                     control_instance.jobs_running += 1
                     self.graph['controlplane']['graph'].add_job(task)
-                    execution_instance = self.real_instances[control_instance.hostname]
-                    self.start_task(task, 'controlplane', task.get_jobs_fail_chain(), execution_instance)
+                    self.start_task(task, rampart_group, task.get_jobs_fail_chain(), None)
                     found_acceptable_queue = True
                     break
 
@@ -529,12 +527,14 @@ class TaskManager:
                 if settings.IS_K8S and task.capacity_type == 'execution':
                     logger.debug("Skipping group {}, task cannot run on control plane".format(rampart_group.name))
                     continue
-
+                # at this point we know the instance group is NOT a container group
+                # because if it was, it would have started the task and broke out of the loop.
                 execution_instance = InstanceGroup.fit_task_to_most_remaining_capacity_instance(
                     task, self.graph[rampart_group.name]['instances']
                 ) or InstanceGroup.find_largest_idle_instance(self.graph[rampart_group.name]['instances'], capacity_type=task.capacity_type)
 
-                if execution_instance or rampart_group.is_container_group:
+                if execution_instance:
+                    task.execution_node = execution_instance.hostname
                     # If our execution instance is a hybrid, prefer to do control tasks there as well.
                     if (
                         execution_instance.node_type == 'hybrid'
@@ -543,22 +543,19 @@ class TaskManager:
                         control_instance = execution_instance
                         task.controller_node = execution_instance.hostname
 
-                    if not rampart_group.is_container_group:
-                        control_instance.remaining_capacity = max(0, control_instance.remaining_capacity - settings.AWX_CONTROL_PLANE_TASK_IMPACT)
-                        task.log_lifecycle("controller_node_chosen")
-                        if control_instance != execution_instance:
-                            control_instance.jobs_running += 1
-                        execution_instance.remaining_capacity = max(0, execution_instance.remaining_capacity - task.task_impact)
-                        execution_instance.jobs_running += 1
-                        task.log_lifecycle("execution_node_chosen")
-                        logger.debug(
-                            "Starting {} in group {} instance {} (remaining_capacity={})".format(
-                                task.log_format, rampart_group.name, execution_instance.hostname, remaining_capacity
-                            )
+                    control_instance.remaining_capacity = max(0, control_instance.remaining_capacity - settings.AWX_CONTROL_PLANE_TASK_IMPACT)
+                    task.log_lifecycle("controller_node_chosen")
+                    if control_instance != execution_instance:
+                        control_instance.jobs_running += 1
+                    execution_instance.remaining_capacity = max(0, execution_instance.remaining_capacity - task.task_impact)
+                    execution_instance.jobs_running += 1
+                    task.log_lifecycle("execution_node_chosen")
+                    logger.debug(
+                        "Starting {} in group {} instance {} (remaining_capacity={})".format(
+                            task.log_format, rampart_group.name, execution_instance.hostname, execution_instance.remaining_capacity
                         )
-
-                    if execution_instance:
-                        execution_instance = self.real_instances[execution_instance.hostname]
+                    )
+                    execution_instance = self.real_instances[execution_instance.hostname]
                     self.graph[rampart_group.name]['graph'].add_job(task)
                     self.start_task(task, rampart_group, task.get_jobs_fail_chain(), execution_instance)
                     found_acceptable_queue = True
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -70,6 +70,7 @@ class TaskManager:
         """
         instances = Instance.objects.filter(hostname__isnull=False, enabled=True).exclude(node_type='hop')
         self.real_instances = {i.hostname: i for i in instances}
+        self.controlplane_ig = None
 
         instances_partial = [
             SimpleNamespace(
@@ -86,6 +87,8 @@ class TaskManager:
         instances_by_hostname = {i.hostname: i for i in instances_partial}
 
         for rampart_group in InstanceGroup.objects.prefetch_related('instances'):
+            if rampart_group.name == settings.DEFAULT_CONTROL_PLANE_QUEUE_NAME:
+                self.controlplane_ig = rampart_group
             self.graph[rampart_group.name] = dict(
                 graph=DependencyGraph(),
                 execution_capacity=0,
@@ -298,7 +301,12 @@ class TaskManager:
             if rampart_group is not None:
                 self.consume_capacity(task, rampart_group.name, instance=instance)
             if task.controller_node:
-                self.consume_capacity(task, 'controlplane', instance=self.real_instances[task.controller_node], impact=settings.AWX_CONTROL_NODE_TASK_IMPACT)
+                self.consume_capacity(
+                    task,
+                    settings.DEFAULT_CONTROL_PLANE_QUEUE_NAME,
+                    instance=self.real_instances[task.controller_node],
+                    impact=settings.AWX_CONTROL_NODE_TASK_IMPACT,
+                )
 
         def post_commit():
             if task.status != 'failed' and type(task) is not WorkflowJob:
@@ -461,7 +469,6 @@ class TaskManager:
     def process_pending_tasks(self, pending_tasks):
         running_workflow_templates = {wf.unified_job_template_id for wf in self.get_running_workflow_jobs()}
         tasks_to_update_job_explanation = []
-        controlplane_ig = InstanceGroup.objects.get(name='controlplane')
         for task in pending_tasks:
             if self.start_task_limit <= 0:
                 break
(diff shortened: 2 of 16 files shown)


awx/settings/development.py
```diff
@@ -21,6 +21,8 @@
 # Load default settings.
 from .defaults import *  # NOQA
 
+# How much capacity controlling a task costs a node
+AWX_CONTROL_NODE_TASK_IMPACT = 5
```
shanemcd: What's the difference between this variable and `AWX_CONTROL_PLANE_TASK_IMPACT` in `defaults.py`?

awx/main/scheduler/task_manager.py
```diff
@@ -600,16 +626,17 @@ def reap_jobs_from_orphaned_instances(self):
...
         logger.debug(
             '{} consumed {} capacity units from {} with prior total of {}'.format(
-                task.log_format, task.task_impact, instance_group, self.graph[instance_group]['consumed_capacity']
+                task.log_format, impact, instance_group, self.graph[instance_group]['consumed_capacity']
             )
         )
-        self.graph[instance_group]['consumed_capacity'] += task.task_impact
+        self.graph[instance_group]['consumed_capacity'] += impact
         for capacity_type in ('control', 'execution'):
             if instance is None or instance.node_type in ('hybrid', capacity_type):
```
AlanCoding, line 631: If you are consuming capacity for controller_node selection, then `capacity_type` should be "control", not whatever type the unified job is.

Otherwise your approach to call consume_capacity twice is valid and matches my expectations.

awx/main/scheduler/task_manager.py
```diff
@@ -283,38 +283,12 @@ def start_task(self, task, rampart_group, dependent_tasks=None, instance=None):
                 task.send_notification_templates('running')
                 logger.debug('Transitioning %s to running status.', task.log_format)
                 schedule_task_manager()
+            # at this point we already have control/execution nodes selected for the following cases
             elif rampart_group.is_container_group:
                 task.instance_group = rampart_group
```
AlanCoding: you might as well remove the `elif` because there's no difference between that and the `else`

awx/settings/development.py
kdelee: I agree this should not be necessary, I just did this because one test was failing on not picking up the setting in `defaults.py`. I can revert this part and try and get to the bottom of that

awx/main/scheduler/task_manager.py
```diff
@@ -510,13 +487,49 @@ def process_pending_tasks(self, pending_tasks):
                 self.start_task(task, None, task.get_jobs_fail_chain(), None)
                 continue
 
+            # Determine if ther is control capacity for the task
+            remaining_control_capacity = self.get_remaining_capacity('controlplane', capacity_type='control')
+            if remaining_control_capacity < settings.AWX_CONTROL_PLANE_TASK_IMPACT:
+                logger.debug(f"Skipping task {task.log_format} in pending, not enough capacity left on controlplane to control new tasks")
+                continue
```
AlanCoding: I'd vote to delete this condition. I was making notes myself to delete the similar condition for `execution_node` selection. If the IG doesn't have capacity, then the `fit_task_to_most_remaining_capacity_instance` / `find_largest_idle_instance` should fail to find a node. I know I asked you to back out your large refactor for now, but I want to hold onto the important critical thinking that came from it.

awx/main/scheduler/task_manager.py
```diff
@@ -510,13 +487,49 @@ def process_pending_tasks(self, pending_tasks):
...
+
+            control_instance = InstanceGroup.fit_task_to_most_remaining_capacity_instance(
+                task, self.graph['controlplane']['instances']
+            ) or InstanceGroup.find_largest_idle_instance(self.graph['controlplane']['instances'], capacity_type='control')
+            if not control_instance:
+                logger.debug(f"Skipping task {task.log_format} in pending, not enough capacity left on controlplane to control new tasks")
+                continue
+
+            task.controller_node = control_instance.hostname
```
AlanCoding, line 511: I have some concern about assigning here. I thought maybe a local variable for `controller_node` would be best, and then assign right before `start_task` is called. You could still do this, and it's not necessarily a bad idea, but it may force some surprising follow-on changes because I think it could get saved when _execution_ capacity falls short.

AlanCoding: Next comment, if you get to this line and `task.capacity_type == 'control'`, then you should assign `execution_node` to the same value as `controller_node`, put it in the `controlplane` IG, start task, and then `continue`. Project updates and system jobs should not search over preferred instance groups.

awx/main/scheduler/task_manager.py
```diff
@@ -510,13 +487,49 @@ def process_pending_tasks(self, pending_tasks):
...
+
             for rampart_group in preferred_instance_groups:
                 if task.capacity_type == 'execution' and rampart_group.is_container_group:
```
AlanCoding: if you accept my prior comment, `task.capacity_type == 'execution'` is always true, and can be deleted in cases like these.

awx/main/scheduler/task_manager.py
kdelee: When we call this we pass the controller_node as the `instance`, so it will be hybrid or control. We don't reference the task.capacity_type here, rather we make decision based on the instance.node_type.

For hybrid node:
  - it will update:
    - `consumed_capacity`
    -  `consumed_execution_capacity`  
    -  `consumed_control_capacity`
    
for control node:
  - it will update:
    -  `consumed_capacity`
    -  `consumed_control_capacity`

Is that not right for hybrid nodes?
If that is not right for hybrid nodes, then are we also doing the wrong thing for jobs? Because anytime this is called w/ a hybrid node we'll deduct from both.

awx/main/scheduler/task_manager.py
```diff
@@ -510,13 +487,49 @@ def process_pending_tasks(self, pending_tasks):
...
+                    task.execution_node = control_instance
+                    control_instance.remaining_capacity = max(
+                        0, control_instance.remaining_capacity - (task.task_impact + settings.AWX_CONTROL_PLANE_TASK_IMPACT)
+                    )
+                    control_instance.jobs_running += 1
+                    self.graph['controlplane']['graph'].add_job(task)
+                    execution_instance = self.real_instances[control_instance.hostname]
+                    self.start_task(task, 'controlplane', task.get_jobs_fail_chain(), execution_instance)
+                    found_acceptable_queue = True
+                    break
```
AlanCoding, line 530: Again, if you accept my prior comment about `capacity_type` being control, then the `if` condition here becomes unreachable and all of this block is deleted.

awx/main/scheduler/task_manager.py
AlanCoding: you're right. I was thinking that the `task` capacity type played into this, but it doesn't, because this is instance-oriented.

awx/main/scheduler/task_manager.py
```diff
@@ -532,9 +543,22 @@ def process_pending_tasks(self, pending_tasks):
                 ) or InstanceGroup.find_largest_idle_instance(self.graph[rampart_group.name]['instances'], capacity_type=task.capacity_type)
 
                 if execution_instance or rampart_group.is_container_group:
```
AlanCoding: If it's a container group, then this code is unreachable, so delete `or rampart_group.is_container_group`

awx/main/scheduler/task_manager.py
```diff
@@ -510,13 +485,49 @@ def process_pending_tasks(self, pending_tasks):
...
                 continue
 
+            # Determine if ther is control capacity for the task
+            remaining_control_capacity = self.get_remaining_capacity('controlplane', capacity_type='control')
+            if remaining_control_capacity < settings.AWX_CONTROL_PLANE_TASK_IMPACT:
+                logger.debug(f"Skipping task {task.log_format} in pending, not enough capacity left on controlplane to control new tasks")
+                continue
+
+            control_instance = InstanceGroup.fit_task_to_most_remaining_capacity_instance(
+                task, self.graph['controlplane']['instances']
```
AlanCoding: It looks like this needs something else. `fit_task_to_most_remaining_capacity_instance` will search for a suitable instance for `task.task_impact`, but what you need here is the controller_node impact instead (constant of 5 for most jobs).

awx/main/scheduler/task_manager.py
kdelee: done

awx/main/scheduler/task_manager.py
kdelee: did the second thing already. I will look at setting `task.controller_node` later

I _do_ think that is a bit paranoid though, because we dont call `task.save()` until the bottom of `start_task` ....which we never get to if we don't have execution capacity

AlanCoding: What I was thinking of was at the end if this method we do `UnifiedJob.objects.bulk_update(tasks_to_update_job_explanation, ['job_explanation'])`, but now that I paste that I realize it won't update other fields, lol.

awx/main/scheduler/task_manager.py
kdelee: ok I've done this

awx/main/scheduler/task_manager.py
```diff
@@ -510,9 +486,39 @@ def process_pending_tasks(self, pending_tasks):
...
+            # All task.capacity_type == 'control' jobs should run on control plane, no need to loop over instance groups
+            if task.capacity_type == 'control':
+                # This is a task that actually needs to run in the controlplane
+                is_idle = control_instance.jobs_running == 0
+                has_capacity_for_task_impact_and_control = control_instance.remaining_capacity >= (task.task_impact + settings.AWX_CONTROL_NODE_TASK_IMPACT)
+                if not has_capacity_for_task_impact_and_control and not is_idle:
+                    # As in other places, we accept an idle instance if the node with most capacity (the control node we already selected)
+                    # does not have enough capacity.
+                    logger.debug(f"Not enough control capacity on {control_instance} to run {task.log_format}")
+                    break
```
kdelee: @AlanCoding I think I should acutally call `continue` here and add
```
                task.log_lifecycle("needs_capacity")
                job_explanation = gettext_noop("This job is not ready to start because there is not enough available control capacity.")
```
what do you think of that

awx/main/managers.py
```diff
@@ -243,7 +243,15 @@ def capacity_values(self, qs=None, tasks=None, breakdown=False, graph=None):
         for t in tasks:
             # TODO: dock capacity for isolated job management tasks running in queue
             impact = t.task_impact
-            if t.status == 'waiting' or not t.execution_node:
+            control_groups = []
+            if hasattr(t, 'controller_node') and t.controller_node:
```
shanemcd: The use of `hasattr` is a little suspicious here. Should we not just define a default property on `UnifiedJob`?

shanemcd: It looks like it's a database field on UnifiedJob: https://github.com/ansible/awx/blob/ecf9a0827dc29338493ea7431393913029785458/awx/main/models/unified_jobs.py#L586

Why do we need this `hasattr`?

awx/main/scheduler/task_manager.py
shanemcd: I think you are correct, otherwise we're ejecting too early out of the `for task in pending_tasks` loop.

AlanCoding: I don't think you need any of L501-L508. What question is this asking that wasn't already asked above when `control_instance` was set?

If the control plane doesn't have capacity for the control action (rewording your variable name), then `control_instance` will be `None` and you will have hit the prior `continue`. If all nodes had less than 5 capacity (however unlikely) but there were idle nodes, then the logic to define `control_instance` handles that too.

kdelee: basically this catches the case when a node has 5 capacity but not 6 :laughing: maybe not worth it

awx/main/managers.py
kdelee: I ran into an exception in this area when working on this in my previous PR...but probably just me thrashing. I'll just call `t.controller_node`

kdelee: done

awx/main/scheduler/task_manager.py
AlanCoding: Well, it wouldn't be worth it, but there's a better way to do that

```python
            control_instance = InstanceGroup.fit_task_to_most_remaining_capacity_instance(
                task, self.graph['controlplane']['instances'], impact=settings.AWX_CONTROL_NODE_TASK_IMPACT, capacity_type='control'
            ) or InstanceGroup.find_largest_idle_instance(self.graph['controlplane']['instances'], capacity_type='control')
```

You are specifically managing the impact here, `impact=settings.AWX_CONTROL_NODE_TASK_IMPACT`. If you have a "control" type task, just add 1 to that impact number you pass in.

awx/main/managers.py
```diff
@@ -243,7 +243,13 @@ def capacity_values(self, qs=None, tasks=None, breakdown=False, graph=None):
         for t in tasks:
             # TODO: dock capacity for isolated job management tasks running in queue
             impact = t.task_impact
-            if t.status == 'waiting' or not t.execution_node:
+            control_groups = []
+            if t.controller_node:
```
kdelee, line 247: Ok @shanemcd I remember why I was doing the `hasattr(t, 'controller_node')`

Its because in the tests of this class, the Job objects don't have a `controller_node` attribute for some reason
```
FAILED awx/main/tests/unit/test_capacity.py::test_committed_capacity - AttributeError: 'Job' object has no attribute 'controller_node'
FAILED awx/main/tests/unit/test_capacity.py::test_running_capacity - AttributeError: 'Job' object has no attribute 'controller_node'
FAILED awx/main/tests/unit/test_capacity.py::test_offline_node_running - AttributeError: 'Job' object has no attribute 'controller_node'
FAILED awx/main/tests/unit/test_capacity.py::test_offline_node_waiting - AttributeError: 'Job' object has no attribute 'controller_node'
FAILED awx/main/tests/unit/test_capacity.py::test_RBAC_reduced_filter - AttributeError: 'Job' object has no attribute 'controller_node'
```

Looks like it is a mock class, so i'll update the tests

awx/main/scheduler/task_manager.py
```diff
@@ -521,28 +529,35 @@ def process_pending_tasks(self, pending_tasks):
...
-                        logger.debug(
-                            "Starting {} in group {} instance {} (remaining_capacity={})".format(
-                                task.log_format, rampart_group.name, execution_instance.hostname, remaining_capacity
-                            )
+                if execution_instance:
+                    task.execution_node = execution_instance.hostname
+                    # If our execution instance is a hybrid, prefer to do control tasks there as well.
+                    if (
+                        execution_instance.node_type == 'hybrid'
+                        and execution_instance.remaining_capacity >= task.task_impact + settings.AWX_CONTROL_PLANE_TASK_IMPACT
```
AlanCoding: I think `and execution_instance.remaining_capacity >= task.task_impact + settings.AWX_CONTROL_PLANE_TASK_IMPACT` should be deleted.

Let me know if my comments haven't been making sense. My proposition is that control tasks (like project updates) and the extra penalty for hybrid nodes can fairly easily be handled in `fit_task_to_most_remaining_capacity_instance`, and I have example code for it in https://github.com/ansible/awx/compare/devel...AlanCoding:control_node_fast?expand=1#diff-6b5e2004df5844765d0ce3cb503debb4fb980d19fd436c4d9516b49a6685fd06

awx/main/scheduler/task_manager.py
```diff
@@ -521,28 +529,35 @@ def process_pending_tasks(self, pending_tasks):
...
+                    ):
+                        control_instance = execution_instance
+                        task.controller_node = execution_instance.hostname
+
+                    control_instance.remaining_capacity = max(0, control_instance.remaining_capacity - settings.AWX_CONTROL_PLANE_TASK_IMPACT)
+                    task.log_lifecycle("controller_node_chosen")
+                    if control_instance != execution_instance:
+                        control_instance.jobs_running += 1
```
AlanCoding, line 552: is `jobs_running` ever used for anything? I don't see why we should care about updating it for controller_node functions, but even if we did, this feels like the wrong thing to do. You could introduce a `jobs_controlling`, but just don't? Can we delete this?

EDIT: discussed, and this is the best option for satisfying the idle node check, given the tradeoffs

awx/main/scheduler/task_manager.py
```diff
@@ -510,9 +486,41 @@ def process_pending_tasks(self, pending_tasks):
...
+            if task.capacity_type == 'control':
+                # This is a task that actually needs to run in the controlplane
+                is_idle = control_instance.jobs_running == 0
+                has_capacity_for_task_impact_and_control = control_instance.remaining_capacity >= (task.task_impact + settings.AWX_CONTROL_NODE_TASK_IMPACT)
+                if not has_capacity_for_task_impact_and_control and not is_idle:
+                    # As in other places, we accept an idle instance if the node with most capacity (the control node we already selected)
+                    # does not have enough capacity.
+                    logger.debug(f"Not enough control capacity on {control_instance} to run {task.log_format}")
+                    self.task_needs_capacity(task, tasks_to_update_job_explanation)
+                    continue
```
AlanCoding: still want to get this `if` deleted, let me know if you want me to try and contribute a patch or anything.

awx/main/scheduler/task_manager.py
```diff
@@ -485,6 +459,7 @@ def generate_dependencies(self, undeped_tasks):
     def process_pending_tasks(self, pending_tasks):
         running_workflow_templates = {wf.unified_job_template_id for wf in self.get_running_workflow_jobs()}
         tasks_to_update_job_explanation = []
+        controlplane_ig = InstanceGroup.objects.get(name='controlplane')
```
AlanCoding: We're closing on finishing this up, so let me start with more nit-picky style stuff.

We should use `DEFAULT_CONTROL_PLANE_QUEUE_NAME` instead of the string "controlplane".

Also, I think you should hack this by adding onto `self` in `after_lock_init`. It does a loop over all instance groups, and you can just save this if the name matches this setting. Then you can reference `self.controlplane_ig` or whatever.

awx/main/scheduler/task_manager.py
```diff
@@ -283,46 +283,20 @@ def start_task(self, task, rampart_group, dependent_tasks=None, instance=None):
...
-                        task.log_lifecycle("controller_node_chosen")
-                    except IndexError:
-                        logger.warning("No control plane nodes available to manage {}".format(task.log_format))
-                        return
-                else:
-                    # control plane nodes will manage jobs locally for performance and resilience
-                    task.controller_node = task.execution_node
-                    task.log_lifecycle("controller_node_chosen")
-                logger.debug('Submitting job {} to queue {} controlled by {}.'.format(task.log_format, task.execution_node, task.controller_node))
+                queue_submitted_to = task.execution_node if task.execution_node else rampart_group.name
```
AlanCoding: This smells funny. Instance group names aren't queues. All unified jobs (except for workflows and approval JTs, ugh) are submitted to the controller_node queue.

I think you should just rewrite this log

> Submitting {task.log_format} to control node {task.controller_node}

You could expand this to include the instance group name, and optionally the execution node, but execution node is blank for openshift, so pick your poison.

awx/main/scheduler/task_manager.py
kdelee: this is used when determining if an instance is idle, which is one of the possible ways it can be chosen

awx/main/scheduler/task_manager.py
kdelee: you addressed this in the patch I pushed

awx/main/scheduler/task_manager.py
kdelee: we addressed this in the patch you helped me with

awx/main/scheduler/task_manager.py
```diff
@@ -510,9 +488,36 @@ def process_pending_tasks(self, pending_tasks):
                 self.start_task(task, None, task.get_jobs_fail_chain(), None)
                 continue
 
+            # Determine if ther is control capacity for the task
```
AlanCoding: ther -> there

awx/main/scheduler/task_manager.py
kdelee: thanks, modified

awx/main/scheduler/task_manager.py
kdelee: done

awx/settings/development.py
kdelee: finally realized these were just two different words :D. fixed/deduplicated now

AlanCoding (approved):
As I approve, let me acknowledge that the managers.py code getting ugly. There are 2 spinoff issues from this, and that messy code will become a candidate for outright deletion as those are worked on.

shanemcd (approved):

AlanCoding:
the test failure looks like it only needs a rebase

kdelee:
example of a job that waits on capacity for control then ends up running (Taken from a "hybrid-standalone" type deploy)
```
{"type": "job", "task_id": 1261, "state": "created", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:07.120251+00:00"}
{"type": "job", "task_id": 1261, "state": "acknowledged", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:07.974560+00:00"}
{"type": "job", "task_id": 1261, "state": "needs_capacity", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:08.146437+00:00"}
{"type": "job", "task_id": 1261, "state": "needs_capacity", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:08.529294+00:00"}
{"type": "job", "task_id": 1261, "state": "needs_capacity", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:10.112022+00:00"}
{"type": "job", "task_id": 1261, "state": "needs_capacity", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:17.020540+00:00"}
{"type": "job", "task_id": 1261, "state": "needs_capacity", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:19.669164+00:00"}
{"type": "job", "task_id": 1261, "state": "controller_node_chosen", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "controller_node": "ec2-54-152-200-151.compute-1.amazonaws.com", "time": "2022-02-04T18:01:21.850781+00:00"}
{"type": "job", "task_id": 1261, "state": "execution_node_chosen", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "execution_node": "ec2-54-152-200-151.compute-1.amazonaws.com", "time": "2022-02-04T18:01:21.850976+00:00"}
{"type": "job", "task_id": 1261, "state": "waiting", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:21.883302+00:00"}
{"type": "job", "task_id": 1261, "state": "pre_run", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:22.250789+00:00"}
{"type": "job", "task_id": 1261, "state": "preparing_playbook", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:22.380333+00:00"}
{"type": "job", "task_id": 1261, "state": "running_playbook", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:22.470447+00:00"}
{"type": "job", "task_id": 1261, "state": "work_unit_id_received", "work_unit_id": "vD1UnMq9", "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:22.540987+00:00"}
{"type": "job", "task_id": 1261, "state": "work_unit_id_assigned", "work_unit_id": "vD1UnMq9", "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:22.591757+00:00"}
{"type": "job", "task_id": 1261, "state": "post_run", "work_unit_id": "vD1UnMq9", "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:25.701226+00:00"}
{"type": "job", "task_id": 1261, "state": "finalize_run", "work_unit_id": "vD1UnMq9", "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:25.781308+00:00"}
```

kdelee:
example of a job that waits on capacity for control then ends up running (Taken from a "hybrid-standalone" type deploy)
```
{"type": "job", "task_id": 1261, "state": "created", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:07.120251+00:00"}
{"type": "job", "task_id": 1261, "state": "acknowledged", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:07.974560+00:00"}
{"type": "job", "task_id": 1261, "state": "needs_capacity", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:08.146437+00:00"}
{"type": "job", "task_id": 1261, "state": "needs_capacity", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:08.529294+00:00"}
{"type": "job", "task_id": 1261, "state": "needs_capacity", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:10.112022+00:00"}
{"type": "job", "task_id": 1261, "state": "needs_capacity", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:17.020540+00:00"}
{"type": "job", "task_id": 1261, "state": "needs_capacity", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:19.669164+00:00"}
{"type": "job", "task_id": 1261, "state": "controller_node_chosen", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "controller_node": "my-hybrid-awx-instance.com", "time": "2022-02-04T18:01:21.850781+00:00"}
{"type": "job", "task_id": 1261, "state": "execution_node_chosen", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "execution_node": "my-hybrid-awx-instance.com", "time": "2022-02-04T18:01:21.850976+00:00"}
{"type": "job", "task_id": 1261, "state": "waiting", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:21.883302+00:00"}
{"type": "job", "task_id": 1261, "state": "pre_run", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:22.250789+00:00"}
{"type": "job", "task_id": 1261, "state": "preparing_playbook", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:22.380333+00:00"}
{"type": "job", "task_id": 1261, "state": "running_playbook", "work_unit_id": null, "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:22.470447+00:00"}
{"type": "job", "task_id": 1261, "state": "work_unit_id_received", "work_unit_id": "vD1UnMq9", "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:22.540987+00:00"}
{"type": "job", "task_id": 1261, "state": "work_unit_id_assigned", "work_unit_id": "vD1UnMq9", "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:22.591757+00:00"}
{"type": "job", "task_id": 1261, "state": "post_run", "work_unit_id": "vD1UnMq9", "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:25.701226+00:00"}
{"type": "job", "task_id": 1261, "state": "finalize_run", "work_unit_id": "vD1UnMq9", "template_name": "JobTemplate - SurveyParking\ubbea", "time": "2022-02-04T18:01:25.781308+00:00"}
```

we can see the hybrid node is quite busy both running and controlling jobs:
```
{
    "id": 1,
    "type": "instance",
    "url": "/api/v2/instances/1/",
    "related": {
        "named_url": "/api/v2/instances/my-hybrid-awx-instance.com/",
        "jobs": "/api/v2/instances/1/jobs/",
        "instance_groups": "/api/v2/instances/1/instance_groups/",
        "health_check": "/api/v2/instances/1/health_check/"
    },
    "uuid": "5b81710c-7518-42f4-b63d-7b13c0e90ab1",
    "hostname": "my-hybrid-awx-instance.com",
    "created": "2022-02-04T16:00:46.603459Z",
    "modified": "2022-02-04T16:00:46.603543Z",
    "last_seen": "2022-02-04T18:07:29.323864Z",
    "last_health_check": "2022-02-04T18:07:29.323864Z",
    "errors": "",
    "capacity_adjustment": "1.00",
    "version": "4.2.0",
    "capacity": 54,
    "consumed_capacity": 50,
    "percent_capacity_remaining": 7.41,
    "jobs_running": 6,
    "jobs_total": 356,
    "cpu": 2,
    "memory": 7835545600,
    "cpu_capacity": 8,
    "mem_capacity": 54,
    "enabled": true,
    "managed_by_policy": true,
    "node_type": "hybrid"
}
```

Also, from this same system running a number of jobs/project updates/workflows/inventory updates etc I have seen no tracebacks from any related changes in this PR (nothing from scheduler etc)

kdelee:
@AlanCoding I'm not seeing a fix for this in devel:
```
/usr/lib64/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
awx/main/tests/functional/test_licenses.py:10: in <module>
    from pip.req import parse_requirements
E   ModuleNotFoundError: No module named 'pip.req'
```
That is what is causing my test failure

I'm not having luck re-creating that failure locally. As far as I can tell the import above this must be failing, and so its falling back to this old way to import it (pip < 10) 
https://github.com/pypa/pip/blob/ec8edbf5df977bb88e1c777dd44e26664d81e216/src/pip/_internal/commands/uninstall.py#L11
Looks like pip on main is doing the same way as us :confused:

kdelee:
figured out the bug...https://github.com/pytest-dev/pytest/pull/9609

I imagine we'll pin in devel and I'll rebase

AlanCoding:
Aside from rebasing, I think this is ready to merge.

thenets (approved):

kdelee:
Merging!

rebeccahhh (approved):
Approving even though I worked on this a bit, it looks great

PR #11676, opened 2022-02-03
Automate labels with GHA

shanemcd:
My goal here is to replace awxbot (an unmaintained project) with a set of simple GitHub Actions.

- awxbot used to automatically label our issues, but hasn't since we migrated to using an [issue form](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms).
- The action for [labeling PRs](https://github.com/actions/labeler) looks at the files that were changed, not the description of the PR. This allows us to automate the `component:` label, but we lose the ability to automate the `type:` label for PRs. I couldn't find an action for PRs that behaves the same way as awxbot does.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,8 +1,6 @@
 ---
 name: Bug Report
 description: Create a report to help us improve
-labels:
-  - bug
 body:
   - type: markdown
     attributes:
@@ -38,6 +36,15 @@ body:
       description: What version of AWX are you running?
     validations:
       required: true
+
+  - type: checkboxes
+    id: components
+    attributes:
+      label: Select the relevant components
+      options:
+        - label: UI
+        - label: API
+        - label: Docs
 
   - type: dropdown
     id: awx-install-method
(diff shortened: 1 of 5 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -1,8 +1,6 @@
 ---
 name: Bug Report
 description: Create a report to help us improve
-labels:
-  - bug
```
shanemcd, line 5: This wasn't working to begin with because there is no `bug` label.

Rather than changing this to `type:bug` here, I opted to keep everything consistent and include it in the `.github/issue_labeler.yml` config

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -39,6 +37,15 @@ body:
     validations:
       required: true
 
+  - type: checkboxes
+    id: components
```
shanemcd, line 41: I added this so we could automate the `component:` label for new issues.

jakemcdermott (approved):

Issue #11413, opened 2021-12-02
Set a non-default value for Django's `SESSION_COOKIE_NAME`


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
See 

https://docs.djangoproject.com/en/3.2/ref/settings/#session-cookie-name

> Default: 'sessionid'

> The name of the cookie to use for sessions. This can be whatever you want (as long as it’s different from the other cookie names in your application).

Originally attempted in https://github.com/ansible/awx/pull/11348 and then reverted in https://github.com/ansible/awx/pull/11412. The biggest problem that we have is that something is not persisting right between multiple runs of integration tests, probably related to the awxkit connection object.

In addition to that, the current value of "sessionid" is hard-coded in many places by clients:

https://github.com/ansible/awx/search?q=sessionid

Those can be trivially updated, and would obviously need to be for merging this change. We also have clients in a few other places that would need to change. We would also need to very strongly broadcast and release-note this change, as it could break automation that users have built around some custom clients.

john-westcott-iv:
The session ID is used in 2 places in awxkit: ws and the client.

The client is performing an interactive login and we can set the "default" value to be the old sessionid (in case you are connecting to an older instance) and we can have AWX send a special header to indicate what the cookie name is when the session is first established.

The WS is more of a programatic interface and we can modify that to include a session_cookie_name variable but we will default this variable to the new session cookie name because WS may not have a connection object before it attempts to make a connection. So, WS will assume you are connecting to something new and if you need to connect WS to an older instance you will need to change your code to use the old `sessionid` cookie name.

CFSNM:
The validation of this item is done here: https://github.com/ansible/tower-qa/pull/7195#issuecomment-1046633787

CFSNM:
PR merged. Closing it -...

tvo318:
For docs: https://github.com/ansible/product-docs/pull/1651/commits/217be6fd4ffe7e409c510a47c886a22b66bd8d82

PR #11679, opened 2022-02-03
Changing session cookie name and added a way for clients to know what the name is #11413

john-westcott-iv:
##### SUMMARY

This changes the name of the cookie which holds the users session from `sessionid` to `awx_sessionid` to help distinguish the AWX session cooke from other django applications. The name of the cookie is defined in `awx/settings/defaults.py`. In case this gets changed we have also modified the login logic to set a header `X-API-Session-Cookie-Name` which will have a value of the session cookie name. This way, if a integration point is referencing the cookie, they remote integration will be informed of the name of the cooke on a successful login. In addition we have updated awxkit to reference this new session cookie as well as the header. In awxkit there is the interactive client as well as a WebSocket (WS) programatic interface. The WS code has a new parameter for specifying the cookie name. However, the WS interface does not have access to a Connection object before the session cookie name may be needed so if you are using the WS interface to an AWX instance and change the default cookie name (or use the latest WS to connect to an older AWX instance) you must be sure to add the cookie name parameter for a proper connection.

WARNING: If you have a client that depends on the sessionid cookie you will need to modify the client.

Related #11413

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API
- awxkit

##### AWX VERSION

```
awx: 19.5.2.dev75+gd62bc1a798
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/api/generics.py
+++ b/awx/api/generics.py
@@ -99,6 +99,7 @@ class LoggedLoginView(auth_views.LoginView):
             current_user = smart_text(JSONRenderer().render(current_user.data))
             current_user = urllib.parse.quote('%s' % current_user, '')
             ret.set_cookie('current_user', current_user, secure=settings.SESSION_COOKIE_SECURE or None)
+            ret.setdefault('X-API-Session-Cookie-Name', getattr(settings, 'SESSION_COOKIE_NAME', 'awx_sessionid'))
 
             return ret
         else:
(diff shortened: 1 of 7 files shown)


jbradberry (approved):

CFSNM:
Hey @john-westcott-iv Is this PR ready to me merged? It has the Ok from QE side

shanemcd (approved):

AlanCoding (approved):

Issue #11634, opened 2022-01-28
Can't input the name starts with "default" in "Create new container group" screen


kurokobo:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

In the "Create new container group" screen, if I enter "default" in the "Name" field, the field will be immediately disabled and cannot be edited.

![Animation](https://user-images.githubusercontent.com/2920259/151577165-52794190-527f-44e0-bafb-a121696d3f23.gif)

This is a problem when entering a name that starts with "default", such as "default instance for hoge team".

This will only happen if the string in the field is an exact match for "default".

This can be avoided as a workaround by typing "instance for hoge team" first and then adding "default" to the beginning, or by pasting the full name from the clipboard.

### AWX version

19.5.0

### Installation method

kubernetes

### Modifications

no

### Ansible version

2.11.8rc1.post0 on quay.io/ansible/awx-ee:latest

### Operating system

CentOS 8.2.2004

### Web browser

Firefox, Chrome

### Steps to reproduce

1. Open "Instance Groups" > "Add" page.
2. Type "default" by hand in the "Name" field.

### Expected results

The field remains editable.

### Actual results

The field immediately disabled and cannot be edited.

### Additional information

Not tested on Safari or Edge.

mabashian:
https://github.com/ansible/awx/blob/devel/awx/ui/src/screens/InstanceGroup/shared/ContainerGroupForm.js#L49.  We probably only want to disable this field when the value is `default` and the field is clean (untouched).

@kurokobo easiest workaround until we can get this fixed is to type out the name of your container group somewhere else and then paste it in the field.

InnocentK:
Resolved

PR #11691, opened 2022-02-07
Fixes erroneous disabling of name input field on container and instance group forms

AlexSCorey:
##### SUMMARY
This addresses #11634 
Instance Groups and Container Groups have names, (`default`, `controlplane`) that are not available to the user when creating/editing an instance or container group.  This PR fixes a case where a user would go to type `default` and the name input would immediately disable which would prevent them from typing something like `default Instance Group`.  

To resolve that both forms do a check to see if the `initialValues === default || controlplane` (which would mean we are in edit mode, and these are protected instance groups so the name input should be disabled. Doing it this way prevents the field from going disabled as soon as the use types `default` or `controlplane` giving them the opportunity to type more than just those names.  I also introduce a validators to throw a form error when the user tries to create IG/CGs of those protected names
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
Instancegroups

### DIFF
--- a/awx/ui/src/util/validators.js
+++ b/awx/ui/src/util/validators.js
@@ -186,3 +186,8 @@ export function regExp() {
     return undefined;
   };
 }
+
+export function protectedResourceName(message, names = []) {
+  return (value) =>
+    names.some((name) => value.trim() === `${name}`) ? message : undefined;
+}
--- a/awx/ui/src/screens/InstanceGroup/ContainerGroup.js
+++ b/awx/ui/src/screens/InstanceGroup/ContainerGroup.js
@@ -30,13 +30,16 @@ function ContainerGroup({ setBreadcrumb }) {
     isLoading,
     error: contentError,
     request: fetchInstanceGroups,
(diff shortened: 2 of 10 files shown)


awx/ui/src/util/validators.js
```diff
@@ -186,3 +186,8 @@ export function regExp() {
     return undefined;
   };
 }
+
+export function protectedResourceName(names = [], message) {
```
nixocio:
```suggestion
export function protectedResourceName(message, names = []) {
```

nixocio (changes requested):
Good job @AlexSCorey, minor suggestion.

nixocio (approved):

InnocentK (approved):
The above E2E tests showing failure are not caused by the work done in this PR and so it should be safe to merge

InnocentK:
From QE: Good to merge once all the required CLI tests pass

PR #11692, opened 2022-02-07
Workflow changes

john-westcott-iv:
Modifying workflows to install python for make commands
Squashing CI tasks
Modifying pre-commit.sh to not fail if there are no python file changes

##### SUMMARY

The workflow tests that run on the ubuntu-latest containers are having a soft failure because python 3.9 is specified by the make file but not present within the container. This change modifies the GitHub workflows to query the python version from the Makefile and the install said python version. In addition, a change the pre-commit.sh file was added to that if there are no python files change the black command will not be run. This was causing a failure on some OSes such as OS X.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - GitHub Workflows
 
##### AWX VERSION

```
awx: 19.5.2.dev85+gdd7a3c6066
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,13 +5,47 @@ env:
 on:
   pull_request:
 jobs:
-  api-test:
+  common_tests:
+    name: ${{ matrix.tests.name }}
     runs-on: ubuntu-latest
     permissions:
       packages: write
       contents: read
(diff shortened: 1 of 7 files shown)


shanemcd (approved):
Great work!

PR #11693, opened 2022-02-07
Fixes case sensitive host count

AlexSCorey:
##### SUMMARY
Reopens https://github.com/ansible/awx/pull/7621

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION

### DIFF
--- a/awx/main/managers.py
+++ b/awx/main/managers.py
@@ -6,7 +6,7 @@ import logging
 import os
 from django.db import models
 from django.conf import settings
-
+from django.db.models.functions import Lower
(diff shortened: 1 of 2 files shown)


AlanCoding (approved):

AlanCoding:
In the old PR, there was no performance discussion. I don't expect there would be any _major_ performance impact, but it is a possibility we should keep in the backs of our minds.

nixocio (dismissed):
hey @AlexSCorey, there are several extra commits on this PR.

PR #11702, opened 2022-02-08
Do not mount /etc/redhat-access-insights into EEs

tchellomello:
##### SUMMARY
Sharing the /etc/redhat-access-insights is no longer
required for EEs. Furthermore, this fixes a SELinux issue
when launching multiple jobs with concurrency and fact_caching enabled.

```
i.e.:
lsetxattr /etc/redhat-access-insights: operation not permitted
```

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION

```
devel
```

### DIFF
--- a/awx/main/tasks/jobs.py
+++ b/awx/main/tasks/jobs.py
@@ -854,24 +854,6 @@ class RunJob(BaseTask):
                 d[r'Vault password \({}\):\s*?$'.format(vault_id)] = k
         return d
 
-    def build_execution_environment_params(self, instance, private_data_dir):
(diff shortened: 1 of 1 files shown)


fosterseth (approved):

AlanCoding (changes requested):
I think the entire method of `RunJob.build_execution_environment_params` should be deleted. The IS_K8S exception is handled in the parent class

AlanCoding (approved):
Thanks!

Issue #11532, opened 2022-01-13
Allow Hop Nodes to execute and expose health check data


AlexSCorey:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Hope nodes are being rendered on the Mesh Visualizer, but  there are never any health checks that occur on them.  For consistency and clarity we should execute health checks  on Hop nodes in the event that a hope node does not 'check in', at the same frequency that health checks occur on other node types.  We should also clear errors on hop nodes as they begin to check in on the expected frequency.
See this [comment](https://github.com/ansible/awx/pull/11454#discussion_r783434257)

AlanCoding:
Health checks work by running the ansible-runner health check command, and hop nodes do not have that installed. In that comment, I meant to say that we would put errors on those nodes if they went missing from the mesh (more or less).

shanemcd:
@AlanCoding Thoughts on installing runner on the hop nodes for the purposes of running the health check?

shanemcd:
Maybe for hop nodes we dont care about capacity, but it would be good to know if it is on or off.

AlanCoding:
^ receptor already does everything we need. The only reason we run the health check is for information related to capacity which is irrelevant for hop nodes.

However, hop nodes _can_ have `errors`, a field we populated in the API. This only matters for AWX mirroring the logic that receptor reports. If a node isn't reachable, then receptor will discover this fact through its own clustering rules. That is reported from `receptorctl status` which AWX regularly checks.

The way I would phrase this work item - if receptor reports that a hop node has become unreachable, this should be reflected in the mesh visualizer in the UI.

AlexSCorey:
@AlanCoding when a hop node becomes unreachable will it be declared on the `errors` field or should we be tracking some other field to show a hop node isn't reachable?  Just seeking clarity here
@kialam just FYI

AlanCoding:
@AlexSCorey that's right, but we may need to check this in the API to assure we're not skipping the hop nodes. I'm not sure if @jbradberry has covered this or not. It shouldn't be hard to check in the dev environment - start up once with execution nodes, and then start up again with 0 execution nodes, so it won't create the hybrid node either. All the execution plane should show as not working in the visualizer.

tiagodread:
Based on https://github.com/ansible/awx/pull/11653#issuecomment-1027250278 after having the health check for hop nodes, @AlexSCorey, maybe, we'll need to stop looking on `errors` field to look at `last_health_check` instead or other.

[...]

PR #11714, opened 2022-02-09
Fill in `errors` for hop nodes when Last Seen is out of date, and clear them when not

AlanCoding:
Connect https://github.com/ansible/awx/issues/11532

### DIFF
--- a/awx/main/tasks/system.py
+++ b/awx/main/tasks/system.py
@@ -436,7 +436,6 @@ def inspect_execution_nodes(instance_list):
         workers = mesh_status['Advertisements']
         for ad in workers:
             hostname = ad['NodeID']
-            changed = False
 
             if hostname in node_lookup:
                 instance = node_lookup[hostname]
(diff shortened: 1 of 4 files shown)


AlanCoding:
From discussions I had, it sounds like this is doing the expected thing.

awx/main/utils/common.py
```diff
@@ -756,6 +756,9 @@ def get_mem_effective_capacity(mem_mb):
 
 
 def get_corrected_memory(memory):
+    if memory == 0:
+        return memory
```
AlanCoding: This might go better in `get_mem_effective_capacity`

jbradberry (approved):

AlanCoding:
I relaunched some test results and they all look pretty good.

awx/main/tasks/system.py
```diff
@@ -458,11 +457,11 @@ def inspect_execution_nodes(instance_list):
 
             # Only execution nodes should be dealt with by execution_node_health_check
             if instance.node_type == 'hop':
+                logger.warn(f'Hop node {hostname}, has rejoined the receptor mesh')
```
jbradberry: The `.warn()` method is deprecated and throws a warning.  Please use `.warning()` instead.

AlanCoding: d'oh, I should have known better.

PR #11717, opened 2022-02-10
Add metric for number of events emitted over websocket broadcast

fosterseth:
##### SUMMARY

Add metric to show how many times emit_event_detail is being called by the callback receivers.

This is an important metric to track as it can have an impact on websocket broadcasting performance.

e.g. after running some very chatty playbooks.
```
# HELP callback_receiver_events_insert_db Number of events batch inserted into database
# TYPE callback_receiver_events_insert_db gauge
callback_receiver_events_insert_db{node="awx_1"} 50160

# HELP callback_receiver_events_broadcast Number of events broadcast to other control plane nodes
# TYPE callback_receiver_events_broadcast gauge
callback_receiver_events_broadcast{node="awx_1"} 19676
```

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 19.5.2.dev110+g3993aa9524
```

### DIFF
--- a/awx/main/dispatch/worker/callback.py
+++ b/awx/main/dispatch/worker/callback.py
@@ -119,6 +119,7 @@ class CallbackBrokerWorker(BaseWorker):
             bulk_events_saved = 0
             singular_events_saved = 0
             metrics_events_batch_save_errors = 0
+            emitted_events = 0
             for cls, events in self.buff.items():
                 logger.debug(f'{cls.__name__}.objects.bulk_create({len(events)})')
                 for e in events:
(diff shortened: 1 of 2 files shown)


awx/main/dispatch/worker/callback.py
```diff
@@ -143,6 +144,7 @@ def flush(self, force=False):
                 duration_to_save = time.perf_counter() - duration_to_save
                 for e in events:
                     if not getattr(e, '_skip_websocket_message', False):
+                        emitted_events += 1
```
AlanCoding: that's not a super descriptive variable name.

fosterseth: yeah I changed it, and I added `metrics_` to all of the other variables whose purpose is to track data for metrics.

AlanCoding (approved):

PR #11720, opened 2022-02-10
Move runner params logic into the awx.main.tasks.jobs module

AlanCoding:
Following up from @amolgautam25 refactoring work, it seems pretty clear that ownership of the ansible-runner parameters should be in `BaseTask` derived code. This is an odd-one-out, so I'd like to get it taken care of as we continue to further analyze the code.


AlanCoding:
needs some more work

PR #11725, opened 2022-02-11
Fixup conversion of memory and cpu settings to support k8s resource request format

kdelee:
##### SUMMARY

This setting has not been getting set by default, and needed some fixing
up to be compatible with setting the memory in the same way as we set it
in the operator, as well as with other changes from last year which
assume that ansible runner is returning memory in bytes.

This way we can start setting this setting in the operator, and get a
more accurate reflection of how much memory is available to the control
pod in k8s.

Once this is in, I can set the memory in the operator and that should address the memory part of https://github.com/ansible/awx/issues/11640

This is an alternative to https://github.com/ansible/awx/pull/11038. @shanemcd and I both think using this setting may be a little more fool proof, as we don't have to worry about something changing in cgroups. Just keep it simple and take whatever we set on the resource request/limit for k8s.

##### ISSUE TYPE

 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

Its important to note that we were using this function and passing bytes as we get it from ansible runner https://github.com/ansible/awx/blob/6f9d4d89cde79e47e9716b27b3bb515041417598/awx/main/models/ha.py#L282 which ends up in https://github.com/ansible/awx/blob/6f9d4d89cde79e47e9716b27b3bb515041417598/awx/main/models/ha.py#L225-L264
so I updated the variables to be more clearly named, as we are dealing in bytes not mb, and not gb. So this unifies all the places that look at this setting to assume its result is in bytes.

I've added a test to ensure we get the expected memory conversion and resulting forks.
See https://github.com/ansible/awx/blob/devel/docs/capacity.md#memory-relative-capacity for docs on the arithmetic there, but this basically translates into "Take memory in bytes, convert to megabytes, subtract 2048 megabytes, divide by setting that determines megabytes per fork".

Do we want to revisit that algorithm? Maybe but I'm just respecting whats there now.


AlanCoding:
> in the operator, as well as with other changes from last year which
> assume that ansible runner is returning memory in bytes.

So you're talking about the behavior of the operator, or of something from the past?

https://github.com/ansible/awx/blob/2de5ffc8d94d592498c32b8c3d68a82470d57763/awx/main/models/ha.py#L89-L92

If I check my API locally I find

```json
"memory": 33391198208,
```

which is ~31 Gb, not ~31 Mb (in which case I would not be able to type this in Chrome). So hopefully we're on the same page that we're consistent in units from runner to AWX, because it is documented along all that way. See runner's behavior:

```
bash-5.1$ ansible-runner worker --worker-info
{cpu_count: 8, errors: [], mem_in_bytes: 33391198208, runner_version: 2.1.1.dev24,
  uuid: 126cf42a-9bd1-408d-87d4-80fd0250d289}
```

I'm unclear exactly how the bug here is documented. What values are getting put in? We have the expectation set correctly, it should be an integer number of bytes. You're not getting that, somewhere, for some reason?

kdelee:
@AlanCoding So couple things going on here:
  - SYSTEM_TASK_ABS_MEM used be used by a now deprecated openshift installer
  - SYSTEM_TASK_ABS_MEM used to be assumed to be Gi (gigabytes)
  - now I want to assume SYSTEM_TASK_ABS_MEM is a string that conforms to k8s accepted ways of defining memory resource request/limit https://kubernetes.io/docs/tasks/configure-pod-container/assign-memory-resource/#memory-units
  - with this PR, I can do https://github.com/ansible/awx-operator/pull/791 
  - I agree that runner returns bytes, I just updated the local variable "memory_mb" to be "memory_bytes" to match that
  - I updated the `awx/main/dispatch/pool.py` to consider the memory in bytes instead of Gi, so we can use same conversion of SYSTEM_TASK_ABS_MEM to bytes as in the other place

So, there is nothing wrong with how runner is returning memory -- I'm just updating it so if we set the env vars, we get the right behavior and everything is in bytes so it works the same as the value when we get it from runner instead of a setting.

AlanCoding (approved):
all depends on openshift, sure, but I don't have a problem with how it appears here.

shanemcd (approved):

kdelee:
@shanemcd pushed new changes to migrate CPU to a decimal field and also to not deduct the 2GB memory penalty on k8s that we do when all the services are colocated

Here is a control pod on a minikube deployment with operator built from https://github.com/ansible/awx-operator/pull/791
```
{
    "id": 3,
    "type": "instance",
    "url": "[/api/v2/instances/3/](http://192.168.49.2:31406/api/v2/instances/3/)",
    "related": {
        "named_url": "[/api/v2/instances/awx-7d6b447ccc-tqw9h/](http://192.168.49.2:31406/api/v2/instances/awx-7d6b447ccc-tqw9h/)",
        "jobs": "[/api/v2/instances/3/jobs/](http://192.168.49.2:31406/api/v2/instances/3/jobs/)",
        "instance_groups": "[/api/v2/instances/3/instance_groups/](http://192.168.49.2:31406/api/v2/instances/3/instance_groups/)",
        "health_check": "[/api/v2/instances/3/health_check/](http://192.168.49.2:31406/api/v2/instances/3/health_check/)"
    },
    "uuid": "cbfbfcca-55f6-4ca3-83fb-cd05cfba9089",
    "hostname": "awx-7d6b447ccc-tqw9h",
    "created": "2022-02-14T17:51:10.153995Z",
    "modified": "2022-02-14T17:51:10.154129Z",
    "last_seen": "2022-02-14T18:13:26.629703Z",
    "last_health_check": "2022-02-14T18:13:26.629703Z",
    "errors": "",
    "capacity_adjustment": "1.00",
    "version": "19.5.2.dev111+g9db07d9d7b.d20220214",
    "capacity": 10,
    "consumed_capacity": 0,
    "percent_capacity_remaining": 100.0,
    "jobs_running": 0,
    "jobs_total": 0,
    "cpu": "0.5",
    "memory": 1073741824,
    "cpu_capacity": 2,
    "mem_capacity": 10,
    "enabled": true,
    "managed_by_policy": true,
    "node_type": "control"
}
```
the request on the pod (this is default for awx-operator right now)
```
  awx-task:
....omitted...
    Requests:
      cpu:     500m
      memory:  1Gi
```

shanemcd (approved):

rebeccahhh (approved):

PR #11742, opened 2022-02-14
drop unused logic in task manager

kdelee:
There is no current need or use to keep a seperate dependency graph for
each instance group. In the interest of making it clearer what the
current code does, eliminate this complication.

We are no longer ever referencing any accounting of instance group
capacity, instead we only look
at capacity on intances.

##### SUMMARY

Follow up from https://github.com/ansible/awx/pull/11665, this deletes some more code that maintains data around instance group capacity, but that is never referenced.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

`InstanceGroup.objects.capacity_values` is still used outside of this code, so I did not delete it or attempt a refactor, though it may be possible to refactor that code too.

### DIFF
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -71,6 +71,7 @@ class TaskManager:
         instances = Instance.objects.filter(hostname__isnull=False, enabled=True).exclude(node_type='hop')
         self.real_instances = {i.hostname: i for i in instances}
         self.controlplane_ig = None
+        self.dependency_graph = DependencyGraph()
 
         instances_partial = [
             SimpleNamespace(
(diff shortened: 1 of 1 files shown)


rebeccahhh (approved):

awx/main/scheduler/task_manager.py
```diff
@@ -90,32 +91,18 @@ def after_lock_init(self):
...
-                    self.graph[rampart_group.name]['instances'].append(instances_by_hostname[instance.hostname])
 
     def job_blocked_by(self, task):
         # TODO: I'm not happy with this, I think blocking behavior should be decided outside of the dependency graph
         #       in the old task manager this was handled as a method on each task object outside of the graph and
         #       probably has the side effect of cutting down *a lot* of the logic from this task manager class
-        for g in self.graph:
-            blocked_by = self.graph[g]['graph'].task_blocked_by(task)
-            if blocked_by:
-                return blocked_by
```
AlanCoding, line 118: Every time I see this, I get deadlocked in a stare. How did anyone ever think this made sense? I don't know.

AlanCoding (approved):

PR #11756, opened 2022-02-16
Enable Podman ipv6 support by default

shanemcd:


CFSNM:
https://main-jenkins-csb-aap.apps.ocp-c1.prod.psi.redhat.com/job/AAPQA/job/Test_Tower_Yolo_Express/301/

AlanCoding:
What if someone uses `--network` options with the `DEFAULT_CONTAINER_RUN_OPTIONS` setting? In that case, I would argue for removing our own option, and as these cases proliferate, that seems like the most sustainable.

AlanCoding:
Here's what I'm thinking:

https://github.com/ansible/awx/compare/devel...AlanCoding:podman_options?expand=1

To me, this seems like a much more maintainable approach to the problem. There are good and bad ways to set up settings values, and this is the best approach I can think of. If other people have better ideas, I'd like to hear them.

tchellomello:
> Here's what I'm thinking:
> 
> https://github.com/ansible/awx/compare/devel...AlanCoding:podman_options?expand=1
> 
> To me, this seems like a much more maintainable approach to the problem. There are good and bad ways to set up settings values, and this is the best approach I can think of. If other people have better ideas, I'd like to hear them.

+1 I like the way @AlanCoding  proposed to use the `settings.DEFAULT_CONTAINER_RUN_OPTIONS` for this option. I believe this way allows better control or customization for customers.

shanemcd:
@AlanCoding I dont think we should include `--user=root` in that variable. People will try removing it and it will break the system. I'm cool with the rest of your idea.

AlanCoding (approved):

Issue #8460, opened 2020-10-26
Job template survey float doesn’t work 


tiagodread:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Trying to create a job template survey type Float when hit save the API return `{"error":"Default value 2 in survey question 0 expected to be float."}`. This happens on the old and new UI

See more: https://github.com/ansible/awx/pull/8419
Also related: https://github.com/ansible/awx/issues/8222

##### ENVIRONMENT
* AWX version: 15.0.0
* AWX install method: docker for mac
* Ansible version:  2.9.3
* Operating System: Catalina
* Web Browser: Chrome

##### STEPS TO REPRODUCE

1. Navigate to Job templates
2. Click on Job template
3. Click on Survey tab
4. Create a `Float` survey type like question: 1.0+1.0 default: 2.0

OR use this payload on the `/api/v2/job_templates/<id>/survey_spec/` to create a new survey

```
{
    "description": "",
    "name": "",
    "spec": [
        {
            "choices": "",
            "default": 2,
            "max": 1024,
            "min": 0,
            "new_question": true,
            "question_description": "",
            "question_name": "1.0+1.0",
            "required": true,
            "type": "float",
            "variable": "float"
        }
    ]
}
```

##### EXPECTED RESULTS

The API can successfully save 2 as float (2.0) even passing as integer

##### ACTUAL RESULTS

The API returns `{"error":"Default value 2 in survey question 0 expected to be float."}`

##### ADDITIONAL INFORMATION

![survey](https://user-images.githubusercontent.com/12464858/97207605-fc485a00-1798-11eb-8b4a-99cbfe757b37.gif)

unlikelyzero:
Adding `component:ui` as it does not function there, either. May not have ever worked so not using `qe:regression`label

shanemcd:
When someone picks this up, we just need to review and merge the PR above, and let QE decide if they want to add tests.

InnocentK:
Resolved

PR #11757, opened 2022-02-16
Api issue float

nixocio:
Add SURVEY_TYPE_MAPPING to constants, and replace usage in a couple of
files.

See: https://github.com/ansible/awx/issues/8460

### DIFF
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -2468,7 +2468,7 @@ class JobTemplateSurveySpec(GenericAPIView):
     obj_permission_type = 'admin'
     serializer_class = serializers.EmptySerializer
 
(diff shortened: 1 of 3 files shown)


AlanCoding:
This doesn't make any functional changes yet, right? So this just needs to be adopted in another file to address the issue.

nixocio:
@AlanCoding, no functional changes. Exactly.

AlanCoding:
huh, well then it seems surprising that the test would be passing, because that covers what the issue reported.

AlanCoding (approved):
I was going too fast earlier, I see this adds the int type to the `'float'` question type, this looks good.

InnocentK (approved):
Great job!

InnocentK:
I verified that the above failing test is not caused by this work so its good to go

Issue #8674, opened 2020-11-24
Show copy resource success toast on all pages where resources can be copied


tiagodread:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Success toast isn’t displayed when copying resources

##### STEPS TO REPRODUCE

1. Navigate to inventories
2. Copy an inventory

##### EXPECTED RESULTS

The user should be able to see on the top right of the screen a success toast (like the old UI)

##### ACTUAL RESULTS

The resource is copied but the toast isn’t displayed 

##### ADDITIONAL INFORMATION

- Need to verify others parts that this toast is displayed

![toast](https://user-images.githubusercontent.com/12464858/100152892-26696680-2e82-11eb-90e5-85f12cc6dcbb.gif)

keithjgrant:
mockup for a page with a similar alert here: https://tower-mockups.testing.ansible.com/patternfly/exec-env/exec-env-detail-test/

InnocentK:
Resolved

PR #11758, opened 2022-02-16
Add Toast messages when resources are copied

keithjgrant:
##### SUMMARY
* Adds `Toast` component to manage displaying toast messages
* Adds toast success messages when copying an inventory, credential, template, EE, or project

related #8674 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui/src/components/TemplateList/TemplateList.js
+++ b/awx/ui/src/components/TemplateList/TemplateList.js
@@ -12,6 +12,7 @@ import useSelected from 'hooks/useSelected';
 import useExpanded from 'hooks/useExpanded';
 import { getQSConfig, parseQueryString } from 'util/qs';
 import useWsTemplates from 'hooks/useWsTemplates';
+import useToast, { AlertVariant } from 'hooks/useToast';
 import { relatedResourceDeleteRequests } from 'util/getRelatedResourceDeleteDetails';
 import AlertModal from '../AlertModal';
 import DatalistToolbar from '../DataListToolbar';
@@ -41,6 +42,8 @@ function TemplateList({ defaultParams }) {
   );
 
   const location = useLocation();
+  const { addToast, Toast, toastProps } = useToast();
+
   const {
     result: {
       results,
@@ -121,6 +124,18 @@ function TemplateList({ defaultParams }) {
       allItemsSelected: isAllSelected,
       fetchItems: fetchTemplates,
     }
+  );
+
+  const handleCopy = useCallback(
+    (newTemplateId) => {
+      addToast({
+        id: newTemplateId,
+        title: t`Template copied successfully`,
+        variant: AlertVariant.success,
+        hasTimeout: true,
+      });
+    },
+    [addToast]
   );
 
   const handleTemplateDelete = async () => {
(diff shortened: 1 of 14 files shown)


awx/ui/src/components/TemplateList/TemplateList.js
```diff
@@ -123,6 +126,18 @@ function TemplateList({ defaultParams }) {
     }
   );
 
+  const handleCopy = useCallback(
+    (newTemplateId) => {
+      addToast({
+        id: newTemplateId,
+        title: t`Template copied successfully`,
+        variant: AlertVariant.success,
```
AlexSCorey, line 134: What if we don't export the AlertVariant from the hook and just pass a string for the `variant:` value here.  

it would look something like `addToast({
        id: newTemplateId,
        title: t'Template copied successfully',
        variant: 'success`,
        hasTimeout: true
        })`

awx/ui/src/components/TemplateList/TemplateListItem.js
```diff
@@ -52,15 +53,19 @@ function TemplateListItem({
   )}/html/upgrade-migration-guide/upgrade_to_ees.html`;
 
   const copyTemplate = useCallback(async () => {
+    let response;
```
AlexSCorey, line 56: It doesn't look like there is error handling for these api request? If not, we should file an issue. In the mean time since we are only show Toast on success we should not pass the variants around and just default to alert.

awx/ui/src/hooks/useToast.js
```diff
@@ -0,0 +1,80 @@
...
+          ouiaId={`toast-message-${toast.id}`}
+        >
+          {toast.message}
+        </Alert>
+      ))}
+    </AlertGroup>
+  );
+}
+
+const ToastType = shape({
```
AlexSCorey: We could store this in `/types` to stay consistent with the pattern.

AlexSCorey (commented):
This is awesome feature that we can use in Health Check also.  Consider whether you want to add that work to this PR.

I've attached a gif to show behavior for slow internet connections. I think we either need to go right into loading state, or some other thing to show...."hey we are working on it."  
slowinternettoast

awx/ui/src/components/TemplateList/TemplateList.js
keithjgrant: I went with the object because it prevents typo issues... but I suppose prop-types would also throw an error if we typo, so 🤷🏻‍♂️

awx/ui/src/components/TemplateList/TemplateListItem.js
keithjgrant: all the pages I changed were already catching errors and displaying them in a modal, so I left that behavior in place

awx/ui/src/components/TemplateList/TemplateList.js
AlexSCorey: For me, this is 6 of one, half-dozen of the other, I was just trying to figure out ways to reduce the number of items being passed back and forth.

AlexSCorey:
Have you had a chance to consider whether you want to add the toast to Health Checks for instances in /instances and /instances_groups/id/instance? 

How about the loading state situation?

keithjgrant:
> Have you had a chance to consider whether you want to add the toast to Health Checks for instances in /instances and /instances_groups/id/instance?
> 
> How about the loading state situation?

This are probably good use cases to consider for the Toast, but I'd like to keep the scope on this PR down to the original issue for now. Should we open new issues for those? Did you have any specific pages in mind where we need that?

AlexSCorey (approved):

InnocentK (approved):
Looks good. Great work Keith!

PR #11762, opened 2022-02-16
Display roles for organization listed when using non-English web browser

nixocio:
Display roles for organization listed when using non-English web browser

### DIFF
--- a/awx/ui/src/components/ResourceAccessList/ResourceAccessList.js
+++ b/awx/ui/src/components/ResourceAccessList/ResourceAccessList.js
@@ -56,31 +56,29 @@ function ResourceAccessList({ apiModel, resource }) {
 
       let orgRoles;
       if (location.pathname.includes('/organizations')) {
-        const {
-          data: { results: roles },
-        } = await RolesAPI.read({ content_type__isnull: true });
-        const sysAdmin = roles.filter(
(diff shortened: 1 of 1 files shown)


awx/ui/src/components/ResourceAccessList/ResourceAccessList.js
```diff
@@ -56,31 +56,29 @@ function ResourceAccessList({ apiModel, resource }) {
...
-        });
+        const [
+          {
+            data: { results: systemAdmin },
+          },
+          {
+            data: { results: systemAuditor },
+          },
+        ] = await Promise.all([
+          RolesAPI.read({ singleton_name: 'system_administrator' }),
```
marshmalien, line 67: Nice! I didn't realize we could query this directly - this lifts some of the complexity from the UI code.

marshmalien (approved):

awx/ui/src/components/ResourceAccessList/ResourceAccessList.js
nixocio: That was @AlanCoding magic. 🧙

nixocio:
I triggered the e2e tests.

InnocentK (approved):
Fix looks good. Running some E2E tests and can merge after verifying those results

PR #11763, opened 2022-02-16
Modifying Makefile so you can specify an alternate branch for runner

john-westcott-iv:
##### SUMMARY

With this chance you can modify the version of ansible-runner built into the awx_tools_1 container with a variable to `make docker-compose-build` like:
```
CUSTOM_RUNNER=git+https://github.com/<your fork>/ansible-runner.git@<your branch>#egg=ansible-runner make docker-compose-build
```

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - Makefile
 
##### AWX VERSION

```
awx: 19.5.2.dev135+g3ec0861ab0.d20220216
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ DEV_DOCKER_TAG_BASE ?= quay.io/awx
 DEVEL_IMAGE_NAME ?= $(DEV_DOCKER_TAG_BASE)/awx_devel:$(COMPOSE_TAG)
 
 RECEPTOR_IMAGE ?= quay.io/ansible/receptor:devel
+CUSTOM_RUNNER ?= 
 
 # Python packages to install only from source (not from binary wheels)
 # Comma separated list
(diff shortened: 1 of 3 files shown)


AlanCoding (changes requested):
I don't love this for a couple of reasons

I don't think it would have an effect on execution nodes?

It's going to take extra time on every startup.

And really, you probably want an in-place install with `-e`, which this won't support. From some prior tinkering I had some action items that I think you could make headway with, but never resolved. I'll send you some suggestions.

shanemcd:
Yeah, -1 on how this is written.

What I'd really love to see is an environment variable like `ANSIBLE_RUNNER_REPO_PATH`, that when set, mounts your checkout of ansible-runner into development environment. Then the startup scripts for the container would run something like this before starting the services:

```
$ /var/lib/awx/venv/awx/bin/pip install -e /path/to/mounted/ansible-runner
```

PR #11776, opened 2022-02-17
Use ternary rather than &&

nixocio:
Use ternary rather than && to avoid display 0.

With this change

image

Without this change

image


nixocio:
This article is pretty good about this problem. https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

marshmalien (approved):
🚀 

_From the article ^_
> I'm well aware that we could've solved the contact problem by using !!contacts.length && ... or contacts.length > 0 && ... or even Boolean(contacts.length) && ..., but I still prefer not abusing the logical AND operator for rendering. I prefer being explicit by using a ternary.

kialam:
with great power comes great responsibility

Issue #11586, opened 2022-01-21
AWX CLI installation instructions don't work as described


Throckmortra:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

following the docs here https://github.com/ansible/awx/blob/19.5.0/awxkit/awxkit/cli/docs/source/install.rst

I get the following error(s):

```
LookupError: setuptools-scm was unable to detect version for /private/var/folders/xp/33zq06rn0m3gr912fmh0ph3w0000gn/T/pip-install-n7cypdcn/awxkit_77096e0915ed4df3ae4de4f71d700f0b.

  Make sure you're either building from a fully intact git repository or PyPI tarballs. Most other sources (such as GitHub's tarballs, a git checkout without the .git folder) don't contain the necessary metadata and will not work.

  For example, if you're using pip, instead of https://github.com/user/proj/archive/master.zip use git+https://github.com/user/proj.git#egg=proj
  ----------------------------------------
WARNING: Discarding https://github.com/ansible/awx/archive/19.5.1.tar.gz#egg=awxkit&subdirectory=awxkit. Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
ERROR: Could not find a version that satisfies the requirement awxkit (unavailable) (from versions: 13.0.0, 14.0.0, 14.1.0, 15.0.0, 15.0.1, 16.0.0, 17.0.0, 17.0.1, 17.1.0, 18.0.0, 19.0.0, 19.1.0, 19.2.0, 19.2.1, 19.2.2, 19.3.0, 19.4.0)
ERROR: No matching distribution found for awxkit (unavailable)
```

Python 3.9.3, OSX

### AWX version

19.5.0

### Installation method

kubernetes

### Modifications

no

### Ansible version

2.11.3

### Operating system

OSX

### Steps to reproduce

Attempt to install AWX CLI with instructions in docs

### Expected results

`awx --help` has output

### Actual results

Failure to install

Throckmortra:
this command works `pip install "git+https://github.com/ansible/awx@19.5.0#egg=awxkit&subdirectory=awxkit"`

Sharnoth:
[...]

PR #11780, opened 2022-02-18
Added version file back to awxkit to resolve installation issues

Throckmortra:
Signed-off-by: Aaron Throckmorton <rickthrockmorton@gmail.com>

##### SUMMARY

fixing https://github.com/ansible/awx/issues/11586

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - awxkit

##### AWX VERSION

```
awx: 20.0.1.dev17+gd8e6c2a6ed
```


Throckmortra:
per https://github.com/ansible/awx/issues/11586

shanemcd (changes requested):
Hi, thanks for the PR. Unfortunately we can't accept this change. We've worked hard to automate our release process and we can not justify spending time on bumping versions in files.

Installs from pypi work fine, and your initial comment that suggested the use of `pip install "git+` is a viable workaround. I understand it's not as fast as installing from a tarball, but I do not believe we can justify the overhead this would add to our release process.

altbrace:
> we can not justify spending time on bumping versions in files.
> Installs from pypi work fine, and your initial comment that suggested the use of `pip install "git+` is a viable workaround. I understand it's not as fast as installing from a tarball, but I do not believe we can justify the overhead this would add to our release process.

@shanemcd But why should users have to find workarounds when it can be fixed so easily? I’m trying to build an Ansible Tower Execution Environment, and the ansible-builder utility that builds the image, relies on a VERSION file in the repo root, not even in /awxkit. There is no workaround. Should ansible-builder maintainers introduce changes to their code, or should you add simple VERSION files?

Issue #10281, opened 2021-05-26
Job output host filter- event type search should include operators


akus062381:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
In the job output page of a job that failed with multiple hosts, and for example, one of the hosts failed and the other two hosts were ok, the event type search filter does not give the user the ability to filter and see multiple search types (for example, there is no option for the user to change the operator from OR to AND, etc)

##### ENVIRONMENT
Google Chrome 90.0.4430.93 on Fedora 34

##### STEPS TO REPRODUCE

When accessing the job output screen of a failed job run with multiple hosts, and at least one host failed and one or more hosts had a different status-
using the Event type filter-
The user is not able to change the operator from AND to OR, for example, which means that if the user checks the boxes of more than one event type, the search filter defaults AND and will show no results.
EXAMPLE: two filters are chosen, and one filter applies to one of the hosts, and one filter applies to the other host, no results will populate.

##### EXPECTED RESULTS

When on the job output page, and the job has multiple hosts of different event types, the user should be able to check the boxes next to more than one filter and see results. This is not happening because the default is set to AND, and since a host cannot manifest  with multiple event types, no results show up. This filter should default to OR.

![image](https://user-images.githubusercontent.com/63685497/119683811-f8a1b200-be11-11eb-8001-ee1b20666d68.png)

mabashian:
Let's coordinate on the automated test coverage here and make sure we've got it covered somewhere

akus062381:
@AlexSCorey in testing this, I see that the filters do not clear properly- the job output screen does not refresh properly unless the user clicks the refresh button after clearing the filter(s).

To recreate:

Access the Demo Inventory. In addition to localhost, add two more hosts to this inventory.

Create a new job template and use this inventory, and choose the playbook
image

Run the job template. On the output screen, you will notice that 1 host was ok and 2 were unreachable.

In the search, choose Event, and then in the Filter by Event dropdown menu, select Host Ok, and then Host Unreachable. The filters work correctly.

Now go back to the Filter by Event dropdown and unselect Host Ok and Host Unreachable. You will notice that the output screen does not properly show the output unless you click the refresh button.

mabashian:
Looks like we created a follow up issue for https://github.com/ansible/awx/issues/10281#issuecomment-1068389526.  Can we close this @akus062381?

akus062381:
@mabashian yes thank you!

PR #11788, opened 2022-02-21
Fixes search drop down items for job output search

AlexSCorey:
##### SUMMARY
This addresses #10281 
This adds the `or` operator to the Job Output Event filter as the option items are mutually exclusive.  Also fixes an undiscovered bug where someone would try to filter events on an inventory_update job or a system_job and the page would crash.  This work here removes the Event option from the search drop down when `job.type === inventory_update || job.type === system_job`
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION

jobfiltering

### DIFF
--- a/awx/ui/src/screens/Job/JobOutput/JobOutputSearch.js
+++ b/awx/ui/src/screens/Job/JobOutput/JobOutputSearch.js
@@ -93,9 +93,12 @@ function JobOutputSearch({
       key: 'stdout__icontains',
       isDefault: true,
     },
-    {
+  ];
+
+  if (job.type !== 'system_job' && job.type !== 'inventory_update') {
+    columns.push({
       name: t`Event`,
-      key: 'event',
+      key: 'or__event',
       options: [
         ['runner_on_failed', t`Host Failed`],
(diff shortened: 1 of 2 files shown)


keithjgrant (approved):

akus062381:
@AlexSCorey in testing this, I see that the filters do not clear properly- the job output screen does not refresh properly unless the user clicks the refresh button after clearing the filter(s).

To recreate:
1. Access the Demo Inventory. In addition to localhost, add two more hosts to this inventory.

2. Create a new job template and use this inventory, and choose the playbook 
image

3. Run the job template. On the output screen, you will notice that 1 host was ok and 2 were unreachable.

4. In the search, choose Event, and then in the Filter by Event dropdown menu, select Host Ok, and then Host Unreachable. The filters work correctly.

5. Now go back to the Filter by Event dropdown and unselect Host Ok and Host Unreachable. You will notice that the output screen does not properly show the output unless you click the refresh button.

akus062381 (approved):

Issue #11712, opened 2022-02-09
Use Select_related on mesh visualizer


AlexSCorey:
##### ISSUE TYPE
Use select_related on queries for mesh_vis end point.

PR #11768, opened 2022-02-17
Fix TypeError when running a command on a host in a smart inventory

nixocio:
Fix TypeError when running a command on a host in a smart inventory

image

See: https://github.com/ansible/awx/issues/11611

PR #11790, opened 2022-02-22
Use select_related on db queries to reduce db calls

AlexSCorey:
##### SUMMARY
Addresses #11712.  Without use `select_related('target', 'source')` we were making `n+1` sequel queries which is a performance issue.  This change reduces that number to just `n`.
##### ISSUE TYPE
-enhancement

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/api/views/mesh_visualizer.py
+++ b/awx/api/views/mesh_visualizer.py
@@ -19,7 +19,7 @@ class MeshVisualizer(APIView):
 
         data = {
             'nodes': InstanceNodeSerializer(Instance.objects.all(), many=True).data,
-            'links': InstanceLinkSerializer(InstanceLink.objects.all(), many=True).data,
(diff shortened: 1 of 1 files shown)


nixocio (changes requested):
This code here https://github.com/ansible/awx/pull/11768 is already merged and being added again as part of this PR.

AlexSCorey:
> This code here #11768 is already merged and being added again as part of this PR.

That is weird I'm not sure it it looks I added that code.

AlanCoding:
@nixocio it doesn't look like this has already been merged. It seems valid as far as I can tell.

AlanCoding (approved):

nixocio (approved):
UI duplicated code was removed.

AlanCoding:
This looks ready to merge!

Issue #11713, opened 2022-02-09
Can disassociate a hybrid node from Control Plane instance group


kialam:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

A user can disassociate a hybrid node from Control Plane instance group.

### AWX version

AWX 19.5.2

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs

### Installation method

docker for mac

### Modifications

no

### Steps to reproduce

1. Go to Instance Groups.
2. Click on controlplane instance group.
3. Select hybrid node and click Disassociate button.
4. Observe that the controlplane IG is now empty.

### Expected results

A user should not be able to have an empty controlplane IG.

### Actual results

Hybrid node is removed from controlplane IG and the group is now empty.

jbradberry:
cc @AlanCoding @shanemcd

AlanCoding:
@kialam is correct, someone can throw the logic in here:

https://github.com/ansible/awx/blob/709c439afc962b2526cb143813ba9623f69f5dfb/awx/api/views/__init__.py#L510-L514

This is a fast API fix. The control plane memberships should not be modified by the user, if any disagrees let me know. Putting it on-deck.

akus062381:
resolved

PR #11791, opened 2022-02-22
Prevents disassociate hybrid node on controlplane instance group

AlexSCorey:
##### SUMMARY
Addresses #11713. Prevents the user from being able to disassociate a hybrid node from the control plane instance group. In the UI inside Instance Groups --> controlplane instance group --> instances select hybrid instance.  Disassociate button should be disabled.

Then, click on the hybrid node.  In the details view the disassociate button should also be disabled.
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API
 - UI

### DIFF
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -408,7 +408,15 @@ class InstanceInstanceGroupsList(InstanceGroupMembershipMixin, SubListCreateAtta
         if parent.node_type == 'control':
             return {'msg': _(f"Cannot change instance group membership of control-only node: {parent.hostname}.")}
         if parent.node_type == 'hop':
-            return {'msg': _(f"Cannot change instance group membership of hop node: {parent.hostname}.")}
+            return {'msg': _(f"Cannot change instance group membership of hop node : {parent.hostname}.")}
+        return None
+
+    def is_valid_removal(self, parent, sub):
+        res = self.is_valid_relation(parent, sub)
+        if res:
+            return res
+        if sub.name == settings.DEFAULT_CONTROL_PLANE_QUEUE_NAME and parent.node_type == 'hybrid':
+            return {'msg': _(f"Cannot disassociate hybrid instance {parent.hostname} from {sub.name}.")}
         return None
 
 
@@ -510,7 +518,15 @@ class InstanceGroupInstanceList(InstanceGroupMembershipMixin, SubListAttachDetac
         if sub.node_type == 'control':
             return {'msg': _(f"Cannot change instance group membership of control-only node: {sub.hostname}.")}
         if sub.node_type == 'hop':
-            return {'msg': _(f"Cannot change instance group membership of hop node: {sub.hostname}.")}
+            return {'msg': _(f"Cannot change instance group membership of hop node : {sub.hostname}.")}
+        return None
+
+    def is_valid_removal(self, parent, sub):
+        res = self.is_valid_relation(parent, sub)
+        if res:
+            return res
+        if sub.node_type == 'hybrid' and parent.name == settings.DEFAULT_CONTROL_PLANE_QUEUE_NAME:
+            return {'msg': _(f"Cannot disassociate hybrid node {sub.hostname} from {parent.name}.")}
         return None
 
 
--- a/awx/ui/src/screens/InstanceGroup/InstanceDetails/InstanceDetails.js
+++ b/awx/ui/src/screens/InstanceGroup/InstanceDetails/InstanceDetails.js
@@ -275,10 +275,11 @@ function InstanceDetails({ setBreadcrumb, instanceGroup }) {
           </Tooltip>
           {me.is_superuser && instance.node_type !== 'control' && (
             <DisassociateButton
-              verifyCannotDisassociate={!me.is_superuser}
+              verifyCannotDisassociate={instanceGroup.name === 'controlplane'}
               key="disassociate"
               onDisassociate={disassociateInstance}
               itemsToDisassociate={[instance]}
+              isProtectedInstanceGroup={instanceGroup.name === 'controlplane'}
               modalTitle={t`Disassociate instance from instance group?`}
             />
           )}
--- a/awx/ui/src/screens/InstanceGroup/Instances/InstanceList.js
+++ b/awx/ui/src/screens/InstanceGroup/Instances/InstanceList.js
@@ -31,7 +31,7 @@ const QS_CONFIG = getQSConfig('instance', {
   order_by: 'hostname',
 });
 
-function InstanceList() {
+function InstanceList({ instanceGroup }) {
   const [isModalOpen, setIsModalOpen] = useState(false);
   const location = useLocation();
   const { id: instanceGroupId } = useParams();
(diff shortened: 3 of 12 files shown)


AlanCoding:
Why only hybrid nodes? That's how the issue was reported, but that's not how I thought the fix would go. It doesn't make sense to disassociate a "control" node from the controlplane either, and execution nodes shouldn't be associated to begin with. I think you should be able to drop that condition.

awx/api/views/__init__.py
```diff
@@ -405,11 +405,23 @@ class InstanceInstanceGroupsList(InstanceGroupMembershipMixin, SubListCreateAtta
...
+        if sub.name == 'controlplane' and parent.node_type == 'hybrid':
+            data = dict({'msg': _(f"Cannot disassociate hybrid node {parent.hostname} from controlplane.")})
+            res = Response(data, status=status.HTTP_400_BAD_REQUEST)
+            return (sub_id, res)
+        return (sub_id, res)
+
     def is_valid_relation(self, parent, sub, created=False):
         if parent.node_type == 'control':
-            return {'msg': _(f"Cannot change instance group membership of control-only node: {parent.hostname}.")}
+            return {'msg': _(f"Cannot change instance group membership of control-only node <<<<<<<<<<<<<<<<<<<<<<<<: {parent.hostname}.")}
```
AlanCoding: looks like git conflict artifacts.

awx/api/views/__init__.py
```diff
@@ -405,11 +405,23 @@ class InstanceInstanceGroupsList(InstanceGroupMembershipMixin, SubListCreateAtta
     parent_model = models.Instance
     relationship = 'rampart_groups'
 
+    def unattach_validate(self, request):
+        sub_id, res = super().unattach_validate(request)
+        if res:
+            return (sub_id, res)
+        parent = self.get_parent_object()
```
AlanCoding: I don't think there's any need to obtain `parent` at all. All of these queries will be duplicates too.

AlanCoding: This is why I suggested overriding `is_valid_relation`

nixocio:
hey @AlexSCorey as we discussed, see PR https://github.com/ansible/awx/pull/11839 may have some overlapping with the work on this PR.

awx/ui/src/screens/InstanceGroup/InstanceDetails/InstanceDetails.js
```diff
@@ -275,10 +275,13 @@ function InstanceDetails({ setBreadcrumb, instanceGroup }) {
           </Tooltip>
           {me.is_superuser && instance.node_type !== 'control' && (
             <DisassociateButton
-              verifyCannotDisassociate={!me.is_superuser}
+              verifyCannotDisassociate={
+                !me.is_superuser || instanceGroup.name === 'controlplane'
```
nixocio: From line 276, we just show this button if it is a super_user. We could simplify.
```suggestion
               instanceGroup.name === 'controlplane'
```

awx/ui/src/screens/InstanceGroup/Instances/InstanceList.js
```diff
@@ -224,13 +224,15 @@ function InstanceList() {
                   ]
                 : []),
               <DisassociateButton
-                verifyCannotDisassociate={selected.some(
-                  (s) => s.node_type === 'control'
-                )}
+                verifyCannotDisassociate={
+                  selected.some((s) => s.node_type === 'control') ||
+                  instanceGroup.name === 'controlplane'
```
nixocio, line 229: Do we need to verify 'hybrid' as well? To be similar to be other related logic. Did I misread this?

https://github.com/ansible/awx/blob/53a2679e3a77d90ed0f64ee38d2b2824e4751bba/awx/ui/src/components/DisassociateButton/DisassociateButton.js#L43

nixocio (commented):
@AlexSCorey, good job. I left a couple of questions related to UI changes.

awx/ui/src/screens/InstanceGroup/Instances/InstanceList.js
AlexSCorey: We only need to prevent disassociation of an hybrid instance if the instance group it belongs to is a protected instance group.  IIRC my thought process was to put that logic here so as to help keep the logic in the instancegroup/instances/instancelist from being too complicated.

nixocio: Thanks for clarifying @AlexSCorey

nixocio (approved):
Approve the UI changes.

akus062381 (approved):

akus062381:
can be merged as soon as @AlanCoding approves API changes :-)

AlanCoding:
I understand why @AlexSCorey did everything he did in this PR, but I'm not happy with the direction it moves the API in. I suggest another change in https://github.com/AlexSCorey/awx/pull/22

awx/api/views/__init__.py
```diff
@@ -405,10 +405,8 @@ class InstanceInstanceGroupsList(InstanceGroupMembershipMixin, SubListCreateAtta
     relationship = 'rampart_groups'
 
     def is_valid_relation(self, parent, sub, created=False):
-        if parent.node_type == 'control':
-            return {'msg': _(f"Cannot change instance group membership of control-only node: {parent.hostname}.")}
-        if parent.node_type == 'hop':
-            return {'msg': _(f"Cannot change instance group membership of hop node: {parent.hostname}.")}
+        if parent.node_type == 'control' or parent.node_type == 'hop' or (sub.name == 'controlplane' and parent.node_type == 'hybrid'):
+            return {'msg': _(f"Cannot change instance group membership of {parent.node_type} node: {parent.hostname}.")}
```
AlanCoding: This message will say the wrong thing for hybrid nodes. You _can_ change memberships of hybrid nodes, but not for the control plane. @jbradberry along the lines of your recent comments.

The sneaky thing here is that instance-instance-group endpoints use `is_valid_relation` symmetrically, on attach and unattach, and no other types of views do this.

It may still be possible to resolve the issue with this approach, but need the wording to be right, and use the setting instead of hard-coding "controlplane"

awx/api/views/__init__.py
```diff
@@ -506,11 +514,19 @@ class InstanceGroupInstanceList(InstanceGroupMembershipMixin, SubListAttachDetac
     relationship = "instances"
     search_fields = ('hostname',)
 
-    def is_valid_relation(self, parent, sub, created=False):
+    def is_valid_relation(self, sub, created=False):
```
AlanCoding: Even if you don't use it, you still need to keep `parent` in the arg list, because it's a position argument. I hope that clears out the test failures, and then this will be ready to go.

awx/api/views/__init__.py
```diff
@@ -404,11 +404,19 @@ class InstanceInstanceGroupsList(InstanceGroupMembershipMixin, SubListCreateAtta
     parent_model = models.Instance
     relationship = 'rampart_groups'
 
-    def is_valid_relation(self, parent, sub, created=False):
+    def is_valid_relation(self, parent, created=False):
```
AlanCoding: I expect you'll need to add back in `sub` as well

awx/api/views/__init__.py
```diff
@@ -408,7 +408,15 @@ def is_valid_relation(self, parent, sub, created=False):
         if parent.node_type == 'control':
             return {'msg': _(f"Cannot change instance group membership of control-only node: {parent.hostname}.")}
         if parent.node_type == 'hop':
-            return {'msg': _(f"Cannot change instance group membership of hop node: {parent.hostname}.")}
+            return {'msg': _(f"Cannot change instance group membership of hop node : {parent.hostname}.")}
+        return None
+
+    def is_valid_removal(self, parent, sub):
+        res = self.is_valid_relation(parent)
```
AlanCoding: need to pass `parent, sub` here

AlanCoding (approved):
thanks for working through this, looks great.

Issue #8321, opened 2020-10-05
RFE: Job view list should have 'caller' schedule name column


swatinarvekar:
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
- UI

##### SUMMARY
The job list doesn't show who ran each job. Desire to have the schedule name (in case trigger by schedule) that made the call or have schedule type like scheduled with schedule name / manual with user name.

Customer Explanation: 
- We use Tower to schedule templates, each schedule performs a different function for the same template. However all jobs and notifications are showing only the template name, so all look the same. How can we customize the output so that we see the name of the schedule?
- Instead of opening every single job in the UI to see the schedule name (LAUNCHED BY)  we need to be able to see the scheduler within the main view, or even with the extended details view.
- Having either the schedule name added to custom notifications and job views, or having labels added to schedules (I believe they can only be set at the template level currently), or even better both, as that'll help with searching too.

##### STEPS TO REPRODUCE
1. Create one schedule named **My Schedule 1** for your job template.
2. Create another schedule named **My Schedule 2** for the same job template.
3. Navigate to the Jobs page.

##### EXPECTED RESULTS
To have a way to differentiate the caller of the jobs (My Schedule 1/ My Schedule 2)

##### ACTUAL RESULTS
Can't figure out the caller of the template (same applies to  project and Inventory source)

ADDITIONAL INFORMATION

Can't easily find which scheduler triggers the template. See the screenshot below:

![compact_view](https://user-images.githubusercontent.com/30921032/95072930-a3e9e380-0729-11eb-8454-a4912fe61697.png)
![expanded_view](https://user-images.githubusercontent.com/30921032/95072936-a51b1080-0729-11eb-954b-38191dbc6248.png)

Here job 207 was started as per schedule while job 208 was started manually.

jakemcdermott:
This sounds good to me. Schedule information available to us in the rest api in the job list view (summary fields) so I consider this something we can add in the new ui once we've established a pattern for "expanded" lists:

![Screenshot from 2020-10-07 14-48-44](https://user-images.githubusercontent.com/9753817/95374341-3619f500-08ac-11eb-9ff6-3d128738a2d0.png)

mabashian:
We've now got the expanded row view in the jobs list:

[...]

PR #11792, opened 2022-02-22
Add schedule detail to job list expanded view

marshmalien:
##### SUMMARY
Issue: https://github.com/ansible/awx/issues/8321

Screen Shot 2022-02-22 at 3 29 10 PM

##### ISSUE TYPE

 - Enhancement Pull Request

##### COMPONENT NAME

 - UI

### DIFF
--- a/awx/ui/src/components/DetailList/LaunchedByDetail.js
+++ b/awx/ui/src/components/DetailList/LaunchedByDetail.js
@@ -1,45 +1,16 @@
 import React from 'react';
 import { Link } from 'react-router-dom';
 import { t } from '@lingui/macro';
+import getScheduleUrl from 'util/getScheduleUrl';
 import Detail from './Detail';
 
-function getScheduleURL(template, scheduleId, inventoryId = null) {
-  let scheduleUrl;
-
-  switch (template.unified_job_type) {
-    case 'inventory_update':
-      scheduleUrl =
-        inventoryId &&
-        `/inventories/inventory/${inventoryId}/sources/${template.id}/schedules/${scheduleId}/details`;
-      break;
-    case 'job':
-      scheduleUrl = `/templates/job_template/${template.id}/schedules/${scheduleId}/details`;
-      break;
-    case 'project_update':
-      scheduleUrl = `/projects/${template.id}/schedules/${scheduleId}/details`;
-      break;
-    case 'system_job':
-      scheduleUrl = `/management_jobs/${template.id}/schedules/${scheduleId}/details`;
-      break;
-    case 'workflow_job':
-      scheduleUrl = `/templates/workflow_job_template/${template.id}/schedules/${scheduleId}/details`;
-      break;
-    default:
-      break;
-  }
-
-  return scheduleUrl;
-}
-
-const getLaunchedByDetails = ({ summary_fields = {}, launch_type }) => {
+const getLaunchedByDetails = (job) => {
   const {
     created_by: createdBy,
     job_template: jobTemplate,
-    unified_job_template: unifiedJT,
     workflow_job_template: workflowJT,
-    inventory,
     schedule,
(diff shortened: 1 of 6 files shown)


awx/ui/src/components/DetailList/LaunchedByDetail.js
```diff
@@ -1,45 +1,16 @@
 import React from 'react';
 import { Link } from 'react-router-dom';
 import { t } from '@lingui/macro';
+import getScheduleUrl from 'util/getScheduleUrl';
```
AlexSCorey, line 4: I like the idea of moving out this of this file.

awx/ui/src/components/JobList/JobListItem.js
```diff
@@ -167,6 +168,17 @@ function JobListItem({
                   />
                 )}
               <LaunchedByDetail job={job} />
+              {job.launch_type === 'scheduled' && (
+                <Detail
+                  dataCy="job-schedule"
+                  label={t`Schedule`}
+                  value={
+                    <Link to={getScheduleUrl(job)}>
+                      {job.summary_fields.schedule.name}
```
AlexSCorey: You could add `schedule` to the destructed job object [here](https://github.com/ansible/awx/pull/11792/files#diff-04d85be33263d77d4c6a8ba191660b3524a19d0b96d788ded224e23484693551R45-R54)

awx/ui/src/components/JobList/JobListItem.test.js
```diff
@@ -51,7 +60,11 @@ describe('<JobListItem />', () => {
 
   test('initially renders successfully', () => {
     expect(wrapper.find('JobListItem').length).toBe(1);
+  });
```
AlexSCorey, line 63: Do we assert anywhere that the Scheduled value does not show up when it shouldn't, if not we probably should.

AlexSCorey (commented):
This looks really good to me.  My only real concern is that we write tests to assert the schedule does not show up when it shouldn't.  Everything else is just my thoughts.

awx/ui/src/util/getScheduleUrl.js
```diff
@@ -0,0 +1,32 @@
+export default function getScheduleUrl(job) {
+  const templateId = job.summary_fields.unified_job_template.id;
+  const scheduleId = job.summary_fields.schedule.id;
```
nixocio, line 3: hey @marshmalien, good job. I added 2 schedules to job template, and the job details showed the schedule what triggered the job. However if I delete those schedules, I got the following error:

image

Once the error above is fixed. We get an empty schedule.  For other items deleted, we add a warning - `Resource deleted`. Perhaps we could add some similar.

image

nixocio, line 3:
```suggestion
  const scheduleId = job.summary_fields?.schedule.id;
```

awx/ui/src/components/JobList/JobListItem.js
nixocio:
```suggestion
                      {job.summary_fields.schedule?.name}
```

awx/ui/src/components/JobList/JobListItem.test.js
marshmalien: 👍 Added a new test that checks for this case "should not show schedule detail in expanded view".

awx/ui/src/util/getScheduleUrl.js
marshmalien: Updated PR to handle Deleted schedules.
Screen Shot 2022-02-28 at 3 51 07 PM

nixocio (approved):

AlexSCorey (approved):
Well done.

akus062381 (approved):

PR #11794, opened 2022-02-23
Allow more than 400 credential types in drop down while adding new credential

jainnikhil30:
##### SUMMARY

This PR fixes the issue of credential type drop down while adding a new credential. Previously the drop down would only show 400 values, because of which a lot of credential types were not showing up (for example: machine, cloud credential etc) but with this PR it will show all credential types even if there are more than 400 values.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - UI

##### ADDITIONAL INFORMATION

Before this PR:

credentila-type

After this PR:
credentialTypesfix


nixocio (approved):
I triggered the e2e tests.

tiagodread (approved):

PR #11797, opened 2022-02-23
Add a leading plus to the refspec

AlanCoding:
Speculative patch based on https://github.com/ansible/awx/issues/9371#issuecomment-888002348

### DIFF
--- a/awx/main/tasks/jobs.py
+++ b/awx/main/tasks/jobs.py
@@ -1177,7 +1177,7 @@ class RunProjectUpdate(BaseTask):
             extra_vars['scm_refspec'] = project_update.scm_refspec
         elif project_update.project.allow_override:
             # If branch is override-able, do extra fetch for all branches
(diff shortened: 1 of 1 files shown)


AlanCoding:
Reported to not solve the issue, interest is waning

junrrein:
Sorry for commenting on a closed MR.

I would like to ask: is there a reason for AWX having a default refspec that is different from Git's default?  
I couldn't find an explanation for this either at the code level (i.e. comments) or at the commit level (with `git blame`).

The reason I'm asking is that in my org we are hitting the referenced issue, and the ops team would like to know the reason for AWX's default. In other words, if there are any (negative) consequences of changing the refspec to allow non-fast-forward updates on `git fetch`.

AlanCoding:
I'll be happy to revisit if I can get a test case that it will fix. I failed to develop one myself.

junrrein:
Thanks. I will try to create a reproducible test case and get back to you.

PR #11809, opened 2022-02-24
WIP disable database settings during migration

AlanCoding:
##### SUMMARY
As of raising the pull request, I want go get some general readings on CI and how happy people would be with this approach.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/conf/disablement.py
+++ b/awx/conf/disablement.py
@@ -1,0 +1,16 @@
+from contextlib import contextmanager
+import threading
+
+
+_conf_settings = threading.local()
+
+
+@contextmanager
+def default_settings():
+    """A context manager that will shut down database settings temporarily, for things like migrations"""
+    try:
+        previous_value = getattr(_conf_settings, 'db_settings_disabled', False)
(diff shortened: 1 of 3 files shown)


jbradberry (commented):
At a first glance I don't love trying to override the `migrate` command like this.  Maybe I'll grow to like it though, I need to give it some thought.

AlanCoding:
@jbradberry problem is that other solutions don't seem to fit my requirements. I don't want to be putting out fires eternally in a bunch of different places. If you raise an exception in a migration python script, you see that the call stack is Django all the way down until you're in the migration file. See the activity stream disabling logic... it's not disabled in all migrations. Our migrations continue to be susceptible to errors from signals and model state. We need a place to do intercept and turn stuff off before migrations start, and without this, I don't see it happening.

I found it very nice that `awx-manage --help` does the right thing automatically. It moves `migrate` out of the Django group and into the main group, which is what I would have wanted.

Issue #10787, opened 2021-07-29
Expose host trusted cert store to EEs by default


wenottingham:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Mostly relevant outside of container groups, but we should consider adding
- /etc/pki/ca-trust/
- /usr/share/pki/ca-trust-source
(or whatever the appropriate paths are)

to the default for AWX_ISOLATION_SHOW_PATHS. Not everyone will want to rebuild an EE just to add certs, and inheriting from the host out of the box does make some sense.

derhoeppi:
This feature is very important. Collections like community.vmware could not be used in a secure way with custom CA's. The only workaround is to ignore every cert validation or create an own ee image. Please prioritize the topic because i don't want to maintain an own image because of certificates.

andrinbr:
It would be great if the EE container would use the "bundle_cacert_secret" from the awx-operator spec like the other deployed containers do.

xx4h:
Just in case anyone else does not want to fork the awx-ee image but still needs a workaround.

My preferred workaround for now is creating a new instance group in AWX (based on default), naming it like "default-with-certs" and adding the same ```initContainers``` logic to this instance group which is used when adding certificates over the AWX spec.

This can be looked up by executing
```
kubectl get deployments.apps awx -o yaml
```

The final custom pod spec should look like
```yaml
apiVersion: v1
kind: Pod
metadata:
  namespace: awx
spec:
  containers:
    - image: 'quay.io/ansible/awx-ee:0.5.0'
      name: worker
      args:
        - ansible-runner
        - worker
        - '--private-data-dir=/runner'
      volumeMounts:
      - mountPath: /etc/pki/ca-trust/extracted
        name: ca-trust-extracted
      - mountPath:
        name: awx-bundle-cacert
        readOnly: true
        subPath: bundle-ca.crt
  initContainers:
    - command:
      - /bin/sh
      - -c
      - |
        mkdir -p /etc/pki/ca-trust/extracted/{java,pem,openssl,edk2}
        update-ca-trust
      image: quai.io/centos/centos:8
      name: init
      imagePullPolicy: IfNotPresent
      resources: {}
      terminationMessagePath: /dev/termination-log
      terminationMessagePolicy: File
      volumeMounts:
      - mountPath: /etc/pki/ca-trust/extracted
        name: ca-trust-extracted
      - mountPath: /etc/pki/ca-trust/source/anchors/bundle-ca.crt
[...]

PR #11812, opened 2022-02-25
Automatically mount the system trust store when invoking EEs with podman

shanemcd:
Related: https://github.com/ansible/awx/issues/10787

Note: This change is only relevant for the downstream product where we support running on virtual machines.

### DIFF
--- a/awx/settings/production.py
+++ b/awx/settings/production.py
@@ -91,3 +91,8 @@ except IOError:
 DATABASES.setdefault('default', dict()).setdefault('OPTIONS', dict()).setdefault(
     'application_name', f'{CLUSTER_HOST_ID}-{os.getpid()}-{" ".join(sys.argv)}'[:63]
 )  # noqa
+
+DEFAULT_CONTAINER_VOLUME_MOUNTS = [
+    '/etc/pki/ca-trust:/etc/pki/ca-trust:O',
+    '/usr/share/pki:/usr/share/pki:O',
+]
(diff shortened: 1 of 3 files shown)


AlanCoding:
I wish we could have just set this for `AWX_ISOLATION_SHOW_PATHS`. If you set a value for that in settings, then it will be the database-setting default value. That allows us to have different default values for production vs. other deployments. That would also give users visibility to this and the ability to remove them through the normal mechanism, if needed. But I understand this would involve migration work, and the need for an immediate band aid is stronger than the long-term maintainability.

awx/settings/production.py
```diff
@@ -91,3 +91,8 @@
 DATABASES.setdefault('default', dict()).setdefault('OPTIONS', dict()).setdefault(
     'application_name', f'{CLUSTER_HOST_ID}-{os.getpid()}-{" ".join(sys.argv)}'[:63]
 )  # noqa
+
+DEFAULT_CONTAINER_VOLUME_MOUNTS = [
+    '/etc/pki/ca-trust:/etc/pki/ca-trust:O',
+    '/usr/share/pki:/usr/share/pki:O',
```
AlanCoding, line 97: I still think that the `AWX_ISOLATION_SHOW_PATHS` setting should disclose (in its help text if nothing else) that additional mounts may or may not be added, and that there's no indication in advance of running jobs.

AlanCoding (approved):

AlanCoding (changes requested):
I think this was trumped by conflicting changes you made.

PR #11816, opened 2022-02-28
Add resolved_action to analytics event data

AlanCoding:
Depends on https://github.com/ansible/ansible-runner/pull/995

This adds the Ansible FQCN (like `ansible.builtin.ping`) to the data analytics gathers about job events.

### DIFF
--- a/awx/main/analytics/collectors.py
+++ b/awx/main/analytics/collectors.py
@@ -337,6 +337,7 @@ def _events_table(since, full_path, until, tbl, where_column, project_job_create
                           {tbl}.parent_uuid,
                           {tbl}.event,
                           task_action,
+                          resolved_action,
                           -- '-' operator listed here:
                           -- https://www.postgresql.org/docs/12/functions-json.html
                           -- note that operator is only supported by jsonb objects
@@ -356,7 +357,7 @@ def _events_table(since, full_path, until, tbl, where_column, project_job_create
                           x.duration AS duration,
                           x.res->'warnings' AS warnings,
                           x.res->'deprecations' AS deprecations
-                          FROM {tbl}, jsonb_to_record({event_data}) AS x("res" json, "duration" text, "task_action" text, "start" text, "end" text)
+                          FROM {tbl}, jsonb_to_record({event_data}) AS x("res" json, "duration" text, "task_action" text, "resolved_action" text, "start" text, "end" text)
                           WHERE ({tbl}.{where_column} > '{since.isoformat()}' AND {tbl}.{where_column} <= '{until.isoformat()}')) TO STDOUT WITH CSV HEADER'''
         return query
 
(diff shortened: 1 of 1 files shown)


AlanCoding:
Integration tests for this are passing now.

jbradberry (changes requested):
If I'm not mistaken, this will require a bump in the version number of this collector.

AlanCoding:
I guess maybe 2 collector versions

https://github.com/ansible/awx/blob/eb520956706df22f86a0c2ed7c43f311f6cbbfcf/awx/main/analytics/collectors.py#L369-L376

Do I just bump it to 1.4?

jbradberry (approved):

Ladas (approved):
Looks good :+1:

slemrmartin:
@AlanCoding the same question like in #11835 . What is db type?

AlanCoding:
@slemrmartin I'm less clear on how to answer this. It's a text value inside of a jsonb field.

Ladas:
@AlanCoding do we document somewhere what the FQCN notation is? Is it:

`<namespace>.<collection>.<role|module>`
or just 
`<namespace>.<collection>.<module>`
?

AlanCoding:
I'd have to run something and check to find out. This came from core, and then runner, and then here. Aren't there 2 scenarios with roles here? A collection can have a role, or you could be running a role next to your playbook. In the latter case, that role can be using a module from a collection.

Ladas:
@AlanCoding I think so, it'd make sense for it being just <namespace>.<collection>.<module>, to always point where the module is coming from. 

There is already separate {tbl}.role in the data, but that is just a role name, so we're missing the info for parent collection of the the role? (if it's role under collection, not standalone)

Ladas:
@AlanCoding I've spoken to @sivel and it looks like the format of resolved_action is:
`{namespace}.{name}.{dotted_path?}{resource}`

Where resource for tasks should be `module`. I don't know if we'll be able to see a plugin usage in the event data somewhere, we'd need to try.

I am creating new issue for runner, including FQCN for roles (or separate namespace and collection name for each role which should be easier to implement) here https://github.com/ansible/ansible-runner/issues/1019 . Since if playbook won't point to the role with FQCN, but use e.g. the `collections:` we won't have access to role's FQCN. (role name is now stored as event.role)

AlanCoding:
What is `dotted_path?`?

I would have liked to go through this before the ansible-runner change was merged. I want to be defensive about the schema of event_data data structure that ansible-runner produces because historically it has felt like a dumping ground, so I always start with a negative gut reaction to adding another field. Writing test cases in ansible-runner is a very cheap thing to do, and we can have a much better discussion with the scenarios I mentioned in my prior comment.

I may re-open the issue for this so I can spend time on those test cases.

sivel:
> What is `dotted_path?`?

Sorry I didn't describe that here.  Basically it was the intention to show that an FQCN can have more than 3 segments, such as:

```
community.general.cloud.packet.packet_device
```

The resource is not always the 3rd segment, it may live in a nested directory under the plugin type.

The above corresponds to the following path within the `community.general` collection:

```
plugins/modules/cloud/packet/packet_device.py
```

PR #11819, opened 2022-02-28
Reimplement transmitter thread as future

shanemcd:
This avoids the need for an explicit `.join()`, and removes the need for the TransmitterThread wrapper class.

### DIFF
--- a/awx/main/tasks/receptor.py
+++ b/awx/main/tasks/receptor.py
@@ -7,8 +7,6 @@ import logging
 import os
 import shutil
 import socket
-import sys
-import threading
 import time
 import yaml
 
@@ -247,16 +245,6 @@ def worker_cleanup(node_name, vargs, timeout=300.0):
     return stdout
 
 
-class TransmitterThread(threading.Thread):
-    def run(self):
-        self.exc = None
-
-        try:
-            super().run()
-        except Exception:
-            self.exc = sys.exc_info()
-
-
 class AWXReceptorJob:
     def __init__(self, task, runner_params=None):
         self.task = task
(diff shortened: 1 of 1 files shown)


awx/main/tasks/receptor.py
```diff
@@ -296,41 +284,45 @@ def _run_internal(self, receptor_ctl):
...
+            # work_unit_id_assigned event then this case may have occured.
+            self.task.instance.work_unit_id = result['unitid']  # Set work_unit_id in-memory only
+            self.task.instance.log_lifecycle("work_unit_id_received")
+            self.task.update_model(self.task.instance.pk, work_unit_id=result['unitid'])
+            self.task.instance.log_lifecycle("work_unit_id_assigned")
+
+        try:
+            transmitter_future.result()
+        except Exception as exc:
+            logger.exception(f'An error was encountered while transmitting the payload for {self.unit_id}')
```
AlanCoding: I don't agree with using `logger.exception` if it's just going to raise the exception afterwards, because logs wind up getting duplicated due to the `raise`. I would suggest that we _do not log at all_. In another issue, we are hunting an issue with the setting local cache, because it's not thread-safe.

AlanCoding:
Do you expect this will make a functional difference, or is it primary a code structure change? There are a lot of bugs related to the threads here, and I am trying to proceed methodically, clearly identifying bugs and fixes before landing patches.

awx/main/tasks/receptor.py
```diff
@@ -296,41 +284,43 @@ def _run_internal(self, receptor_ctl):
...
-        self.task.instance.work_unit_id = result['unitid']  # Set work_unit_id in-memory only
-        self.task.instance.log_lifecycle("work_unit_id_received")
-        self.task.update_model(self.task.instance.pk, work_unit_id=result['unitid'])
-        self.task.instance.log_lifecycle("work_unit_id_assigned")
-
-        sockin.close()
-        sockout.close()
-
-        if transmitter_thread.exc:
-            raise transmitter_thread.exc[1].with_traceback(transmitter_thread.exc[2])
```
shanemcd, line 331: @AlanCoding There was a race condition here... We were looking to see if there was an exception in the transmitter before calling `.join()`. @nitzmahone found this while reading the code. AFAIK this hasn't bitten us yet, but it's pretty obviously a problem.

AlanCoding (approved):
I have a laundry list of risks I want to address close to this code, but these changes are completely orthogonal to the concerns I have, and I see the potential error this would fix.

PR #11830, opened 2022-02-28
Add unique row id to subscription modal list items

marshmalien:
##### SUMMARY

Fix bug where subscription modal shows a console warning when subscriptions have duplicate `pool_ids`: 
```
index.js:1 Warning: Encountered two children with the same key, `row-8a85f99a7e6ed231017e8d77ebf21d70`. 
Keys should be unique so that components maintain their identity across updates. 
Non-unique keys may cause children to be duplicated and/or omitted — 
the behavior is unsupported and could change in a future version.
```

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - UI


AlexSCorey (approved):
The work here to resolve the issue is good to go for me.  Marly and I discussed utilizing formik to handle the values when the modal is opened/closed instead of having a state variable, but that can be addressed another time.

tiagodread (approved):

Issue #11821, opened 2022-02-28
Data from `set_stats` not getting saved to job `artifacts` field as a flaky occurrence


AlanCoding:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

Sometimes the job's `artifacts` field of a job is not getting populated with data from `set_stats` from within a playbook.

The playbook content I'm using to test this comes from https://github.com/ansible/test-playbooks/pull/153

After running that and confirming the job is fully finished, I find that artifacts is `{}`.

Nothing relevant appears in the logs.

This happens in clusters and control-standalone scenarios. Those are the only ones which have been tested.

### AWX version

devel

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs

### Installation method

docker on linux

### Modifications

no

### Ansible version

2.12

### Operating system

Fedora / centos / RHEL

### Web browser

Chrome

### Steps to reproduce

```python
        jt = factories.job_template(
            project=project_from_PR(153, 'Playbook to put environment in artifacts'),
            playbook='print_environment.yml',
            allow_simultaneous=True
        )
        jt.ds.inventory.add_host()
        jobs = [jt.launch() for i in range(16)]
        for job in jobs:
            job.wait_until_completed()
            assert job.artifacts, f'Failed to get artifacts from job {job.id}, output: {job.result_stdout}'
            assert 'printenv' in job.artifacts
```

To summarize, use the test-playbooks content to make a project, create a JT with allow_simultaneous, launch 16 jobs right away. After they have all been launched, wait for completion. After they complete, inspect artifacts field

### Expected results

The playbook content will populate the artifacts with the discovered environment variables and other stuff.

### Actual results

In job details page, `/api/v2/jobs/N/`

```json
    "artifacts": {},
```

AlanCoding:
I re-ran the script here and was not able to detect any failures, so it seems this was fixed by some unknown changes.

PR #11832, opened 2022-02-28
Delay update of artifacts until final job save

AlanCoding:
##### SUMMARY
Connect https://github.com/ansible/awx/issues/11821

The test I had that was failing with that is passing with this patch.

EDIT: Since I raised this PR, it appears that the artifacts issue that was failing tests before was fixed by some other change. I still advocate for this code change for the tech debt aspect of things.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
I don't know exactly what was happening, but some process was clearly doing a `.save` on the job which was regressing the `artifacts` field back to `{}`. By delaying until the final save, that actually leaves us in a safer position, because it's less likely for us to hit one of these unqualified saves after that point, and less likely for one of those saves to have a stale version of the job.

This might need some unit tests added.

I hope it also addresses the concerns that others brought up about the artifacts save in the callback class. It really shouldn't be touching the database.

### DIFF
--- a/awx/main/tasks/callback.py
+++ b/awx/main/tasks/callback.py
@@ -33,6 +33,7 @@ class RunnerCallback:
         self.model = model
         self.update_attempts = int(settings.DISPATCHER_DB_DOWNTOWN_TOLLERANCE / 5)
         self.wrapup_event_dispatched = False
+        self.extra_update_fields = {}
 
     def update_model(self, pk, _attempt=0, **updates):
         return update_model(self.model, pk, _attempt=0, _max_attempts=self.update_attempts, **updates)
@@ -44,6 +45,21 @@ class RunnerCallback:
     @cached_property
     def event_data_key(self):
         return self.instance.event_class.JOB_REFERENCE
+
+    def delay_update(self, skip_if_already_set=False, **kwargs):
+        """Save fields to update after the job finishes."""
+        for key, value in kwargs.items():
+            if key in self.extra_update_fields and skip_if_already_set:
+                continue
+            elif key in self.extra_update_fields and key in ('job_explanation', 'result_traceback'):
+                # In the case of these fields, we do not want to lose any prior information, so combine values
+                self.extra_update_fields[key] = '\n'.join([str(self.extra_update_fields[key]), str(value)])
+            else:
+                self.extra_update_fields[key] = value
+
+    def get_extra_update_fields(self):
+        self.extra_update_fields['emitted_events'] = self.event_ct
+        return self.extra_update_fields
 
     def event_handler(self, event_data):
         #
@@ -149,8 +165,7 @@ class RunnerCallback:
         Handle artifacts
         '''
         if event_data.get('event_data', {}).get('artifact_data', {}):
-            self.instance.artifacts = event_data['event_data']['artifact_data']
-            self.instance.save(update_fields=['artifacts'])
+            self.delay_update(artifacts=event_data['event_data']['artifact_data'])
 
         return False
 
@@ -218,10 +233,7 @@ class RunnerCallback:
         elif status_data['status'] == 'error':
             result_traceback = status_data.get('result_traceback', None)
             if result_traceback:
-                from awx.main.signals import disable_activity_stream  # Circular import
-
-                with disable_activity_stream():
-                    self.instance = self.update_model(self.instance.pk, result_traceback=result_traceback)
+                self.delay_update(result_traceback=result_traceback)
 
 
 class RunnerCallbackForProjectUpdate(RunnerCallback):
--- a/awx/main/tasks/callback.py
+++ b/awx/main/tasks/callback.py
@@ -13,7 +13,7 @@ from django.utils.functional import cached_property
 
 # AWX
 from awx.main.redact import UriCleaner
-from awx.main.constants import MINIMAL_EVENTS
+from awx.main.constants import MINIMAL_EVENTS, ANSIBLE_RUNNER_NEEDS_UPDATE_MESSAGE
 from awx.main.utils.update_model import update_model
 from awx.main.queue import CallbackQueueDispatcher
 
@@ -52,6 +52,8 @@ class RunnerCallback:
             if key in self.extra_update_fields and skip_if_already_set:
                 continue
             elif key in self.extra_update_fields and key in ('job_explanation', 'result_traceback'):
+                if str(value) in self.extra_update_fields.get(key, ''):
+                    continue  # if already set, avoid duplicating messages
                 # In the case of these fields, we do not want to lose any prior information, so combine values
                 self.extra_update_fields[key] = '\n'.join([str(self.extra_update_fields[key]), str(value)])
             else:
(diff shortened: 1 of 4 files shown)


awx/main/tasks/callback.py
```diff
@@ -32,10 +34,26 @@ def __init__(self, model=None):
         self.safe_env = {}
         self.event_ct = 0
         self.model = model
+        self.extra_update_fields = {}
 
     def update_model(self, pk, _attempt=0, **updates):
         return update_model(self.model, pk, _attempt=0, **updates)
 
+    def delay_update(self, **kwargs):
```
jbradberry: By caching artifact updates on the in-memory object like this, won't we be making more pain for ourselves in the future when we try to move to an async style of processing jobs?

jbradberry: If we want to wait until the end of the job, is it possible to do a final sweep over the job events for that job and accumulate the data that way?  Or do we lose this data?

AlanCoding: We can save things on the job record and job events in one of two places
 - the control process
 - the callback receiver

The problem with the callback receiver is that it can (notoriously) be delayed as the events processing is the long-tail action, where a large queue can build up because we don't have the throughput to save at the rate we can produce events. The job _status_ can flip long before all of its events come in. This may have nothing to do with the output of that job. Events may come in at a delayed time before the _last job_ before that job produced more events than could be immediately handled, so all new events have to wait in line.

We cannot have artifacts wait in line with other events. The status changes, and dependent jobs in workflows will start with or without the events saved, and these can use artifacts. Likewise, the event_ct is needed to figure out if all events have been saved in the first place.

AlanCoding:
I realize that we have other saves in `receptor.py`, which should align with this mechanism. I haven't done that, and I probably should before this is a complete and coherent diff.

awx/main/tasks/receptor.py
```diff
@@ -364,26 +369,19 @@ def _run_internal(self, receptor_ctl):
                     logger.warn(f"Could not launch pod for {log_name}. Exceeded quota.")
                     self.task.update_model(self.task.instance.pk, status='pending')
                     return
-                # If ansible-runner ran, but an error occured at runtime, the traceback information
-                # is saved via the status_handler passed in to the processor.
-                if state_name == 'Succeeded':
-                    return res
```
AlanCoding, line 371: There's an in-congruence here. Yes, if the status_handler received this information the receptor output should be this... but it's an imperfect metric. These two things are associated, but other events could cause one to be true but not the other.

so I forced it by doing this

```diff
diff --git a/awx/main/tasks/callback.py b/awx/main/tasks/callback.py
index ab27ce9f36..fd720ac985 100644
--- a/awx/main/tasks/callback.py
+++ b/awx/main/tasks/callback.py
@@ -220,6 +220,8 @@ class RunnerCallback:
             result_traceback = status_data.get('result_traceback', None)
             if result_traceback:
                 self.delay_update(result_traceback=result_traceback)
+        elif status_data['status'] == 'successful':
+            self.delay_update(result_traceback='alan set this!')
 
 
 class RunnerCallbackForProjectUpdate(RunnerCallback):
```

And this works as I expected. The new `if` condition is accurate. I like this also because it avoids running additional receptor commands when they are not necessary. That's something the `if` here was accomplishing before, and I didn't want to lose that.

shanemcd:
Please ping me for another review of this once it's ready to go.

AlanCoding:
Sorry, I had neglected to pick up a test fixture change that came from the Django upgrade. That should be fixed with the latest commit, otherwise I was happy with this and the integration testing looked good.

phill-holbrook:
Hi all - we've been encountering this issue since upgrading to AWX 20 earlier this week, and it's caused us a fair amount of trouble with several workflows. Are there any workarounds we can implement while waiting for the review to be completed?

AlanCoding:
Trivial conflicts resolved.

No, I don't know of any other easy workaround. It was flaky, but pretty easy to hit the error at least once out of 5 attempts. This patch seemed like the easiest possible fix.

awx/main/tasks/callback.py
```diff
@@ -33,10 +33,26 @@ def __init__(self, model=None):
...
+        for key, value in kwargs.items():
+            if key in self.extra_update_fields and skip_if_already_set:
+                continue
+            elif key in self.extra_update_fields and key in ('job_explanation', 'result_traceback'):
+                # In the case of these fields, we do not want to lose any prior information, so combine values
+                self.extra_update_fields[key] = '\n'.join([str(self.extra_update_fields[key]), str(value)])
+            else:
+                self.extra_update_fields[key] = value
+
+    def get_extra_update_fields(self):
```
john-westcott-iv: What about changing this name to match the delay_update function? i.e. `get_delayed_update_fields`.

awx/main/tasks/jobs.py
```diff
@@ -570,18 +567,20 @@ def run(self, pk, **kwargs):
...
         except Exception:
             logger.exception('{} Post run hook errored.'.format(self.instance.log_format))
 
         # We really shouldn't get into this one but just in case....
-        if 'got an unexpected keyword argument' in extra_update_fields.get('result_traceback', ''):
-            extra_update_fields['result_traceback'] = "{}\n\n{}".format(extra_update_fields['result_traceback'], ANSIBLE_RUNNER_NEEDS_UPDATE_MESSAGE)
+        original_traceback = self.runner_callback.get_extra_update_fields().get('result_traceback', '')
+        if 'got an unexpected keyword argument' in original_traceback:
+            if ANSIBLE_RUNNER_NEEDS_UPDATE_MESSAGE not in original_traceback:
+                self.runner_callback.delay_update(result_traceback=ANSIBLE_RUNNER_NEEDS_UPDATE_MESSAGE)
```
john-westcott-iv: One of the things I never really loved was this code being somewhat duplicated in here and receptor.py could we just move this into an if statement in delay_update?

john-westcott-iv (commented):
My only other thought is could this induce a timing issue. i.e. I launch a job its almost done, I click cancel its cancels but then this delayed update thread "finishes" and it changes from canceled to some other status? I haven't really thought through if that would be possible or not.

awx/main/tasks/jobs.py
AlanCoding: Probably. What I think that would look like - if `delay_update` gets `job_explanation` or `result_traceback` and the given string is already in that field, then it does nothing. So we would never repeat the same string twice. Seems reasonable but I should add a test.

AlanCoding:
> My only other thought is could this induce a timing issue. i.e. I launch a job its almost done, I click cancel its cancels but then this delayed update thread "finishes" and it changes from canceled to some other status?

We never call `delay_update` with "status". That would confuse the domains of responsibility. ansible-runner returns a result which has status, and this is what it's set to. If ansible-runner fires the status_callback with one status, but returns the final object with another status, then that would be a contradiction, but AWX would choose the final object.

It could be the case that the job is on its way to error, but a cancel is processed before it finalizes. In that case, we would get a result_traceback or job_explanation in conjunction with the canceled status. That sounds like desired behavior. I would want that information if I was the user.

AlanCoding:
@john-westcott-iv In the last commit, I've come up with a better thought-out answer for how to handle your special ansible-runner error message.

I can't bear adding more code into the main (monolithic) code path for running jobs. So the special case is moved to `get_extra_update_fields` which is a method called right before we update the final status. This can accumulate whatever post-processing of the error fields we need without looking too messy IMO.

shanemcd (approved):

PR #11835, opened 2022-03-01
Add job execution environment image to analytics data

AlanCoding:
##### SUMMARY
It is desired to track what image a job used. That's what this does.

There's still a pending request that we add the image for JTs. This makes less sense to me, but it may be added to this still.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/analytics/collectors.py
+++ b/awx/main/analytics/collectors.py
@@ -419,7 +419,7 @@ def unified_jobs_table(since, full_path, until, **kwargs):
     return _copy_table(table='unified_jobs', query=unified_job_query, path=full_path)
 
 
-@register('unified_job_template_table', '1.0', format='csv', description=_('Data on job templates'))
+@register('unified_job_template_table', '1.1', format='csv', description=_('Data on job templates'))
 def unified_job_template_table(since, full_path, **kwargs):
     unified_job_template_query = '''COPY (SELECT main_unifiedjobtemplate.id,
                                  main_unifiedjobtemplate.polymorphic_ctype_id,
--- a/awx/main/analytics/collectors.py
+++ b/awx/main/analytics/collectors.py
@@ -376,13 +376,14 @@ def events_table_partitioned_modified(since, full_path, until, **kwargs):
     return _events_table(since, full_path, until, 'main_jobevent', 'modified', project_job_created=True, **kwargs)
 
 
-@register('unified_jobs_table', '1.2', format='csv', description=_('Data on jobs run'), expensive=four_hour_slicing)
+@register('unified_jobs_table', '1.3', format='csv', description=_('Data on jobs run'), expensive=four_hour_slicing)
 def unified_jobs_table(since, full_path, until, **kwargs):
     unified_job_query = '''COPY (SELECT main_unifiedjob.id,
                                  main_unifiedjob.polymorphic_ctype_id,
                                  django_content_type.model,
                                  main_unifiedjob.organization_id,
                                  main_organization.name as organization_name,
+                                 main_executionenvironment.image as execution_environment_image,
                                  main_job.inventory_id,
                                  main_inventory.name as inventory_name,
                                  main_unifiedjob.created,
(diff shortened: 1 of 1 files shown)


AlanCoding:
Testing looks good, this is ready for a look.

awx/main/analytics/collectors.py
```diff
@@ -376,13 +376,14 @@ def events_table_partitioned_modified(since, full_path, until, **kwargs):
...
 
-@register('unified_jobs_table', '1.2', format='csv', description=_('Data on jobs run'), expensive=four_hour_slicing)
+@register('unified_jobs_table', '1.3', format='csv', description=_('Data on jobs run'), expensive=four_hour_slicing)
 def unified_jobs_table(since, full_path, until, **kwargs):
     unified_job_query = '''COPY (SELECT main_unifiedjob.id,
                                  main_unifiedjob.polymorphic_ctype_id,
                                  django_content_type.model,
                                  main_unifiedjob.organization_id,
                                  main_organization.name as organization_name,
+                                 main_executionenvironment.image as execution_environment_image,
```
Ladas, line 386: @AlanCoding  What metadata do we track in `main_executionenvironment`? I wonder if it'd make sense to also collect the `main_executionenvironment.id` and also collect full table content of `main_executionenvironment`. ( in a similar way how we collect unified_job_template_table, so a full content of the table)

AlanCoding: the direct fields appear to be these:

```
id
created
modified
description
image
managed
created_by_id
credential_id
modified_by_id
organization_id
name
pull
```

A lot of these are generic to most types. The `pull` field is the only other one that I would say could affect what content a job uses. However, it won't actually tell you whether the image is pulled or not, it's just the pull policy.

> and also collect full table content of main_executionenvironment.

Like a new `@register` method inside of this collectors.py module? So that it could produce a separate csv file just for execution environments.

Ladas: @AlanCoding  Yeah it would be a new `@register` method. I guess it might be worth to pull the image table if we'd add more data to it. (and removing anything sensitive, e.g. description). Are we able to fetch which registry (Hub or I think we support also other registries?) it comes from? Is that captured in the credential?

AlanCoding: > Are we able to fetch which registry (Hub or I think we support also other registries?) it comes from?

Not remotely close. There is no line of visibility into what podman does. We don't have a structured way to report _if_ the image was pulled. Because runner has a minimal set of dependencies, it's hard to see where you could even start to develop these abilities. Picture doing a `podman run` command on the CLI, that's what normal jobs do.

Ladas: ok, so we don't nee to complicate this, lets just keep here the image name (with tag I assume) like you have it now :-)

Ladas (approved):
Looks good :+1:

awx/main/analytics/collectors.py
```diff
@@ -422,6 +424,7 @@ def unified_job_template_table(since, full_path, **kwargs):
     unified_job_template_query = '''COPY (SELECT main_unifiedjobtemplate.id,
                                  main_unifiedjobtemplate.polymorphic_ctype_id,
                                  django_content_type.model,
+                                 main_executionenvironment.image as execution_environment_image,
```
Ladas, line 427: let also bump a version to 1.1 here

AlanCoding: Oh, I didn't catch that. Thanks.

jbradberry (approved):

awx/main/analytics/collectors.py
slemrmartin: @AlanCoding what is database type of the main_executionenvironment.image?
Is some actual db schema available?

AlanCoding: Django CharField

https://github.com/ansible/awx/blob/2afa406b7f16971ea83150c326f8e04717b5c14d/awx/main/models/execution_environments.py#L31

If you want to see specifically what this results in, from dbshell:

```
awx=# select column_name, data_type from information_schema.columns where table_name = 'main_executionenvironment';
   column_name   |        data_type         
-----------------+--------------------------
 id              | integer
 created         | timestamp with time zone
 modified        | timestamp with time zone
 description     | text
 image           | character varying
 managed         | boolean
 created_by_id   | integer
 credential_id   | integer
 modified_by_id  | integer
 organization_id | integer
 name            | character varying
 pull            | character varying
(12 rows)
```

PR #11837, opened 2022-03-02
Move model and settings operations out of threaded code

AlanCoding:
##### SUMMARY
We have some documentation on situations where we hit errors like those described here: https://github.com/tkem/cachetools/issues/90

Most of our code is not thread-safe. We need to be careful to do as little as possible while we have this `ThreadPoolExecutor` active. Several of the kwargs can easily wind up referencing `settings`, and this is a known hazard. Easiest thing to do is build all of these type of params before doing any threads, and I can't see any argument against it.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/tasks/receptor.py
+++ b/awx/main/tasks/receptor.py
@@ -286,19 +286,18 @@ class AWXReceptorJob:
         # reading.
         sockin, sockout = socket.socketpair()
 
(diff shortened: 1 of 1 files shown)


shanemcd (approved):
Awesome

Issue #11718, opened 2022-02-10
Error on Instance Group List based on RBAC


nixocio:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

Error on Instance Group List based on RBAC. There is a request for settings being done at the `InstanceGroupList`.

https://github.com/ansible/awx/blob/6f9d4d89cde79e47e9716b27b3bb515041417598/awx/ui/src/screens/InstanceGroup/InstanceGroupList/InstanceGroupList.js#L71

That leads to an error based on user RBAC.

![image](https://user-images.githubusercontent.com/9053044/153441625-1f664f96-8bf0-4910-98c4-b7db1c735543.png)

### AWX version

devel

### Select the relevant components

- [X] UI
- [ ] API
- [ ] Docs

### Installation method

docker for mac

### Modifications

no

### Steps to reproduce

1 - Login as Org Admin
2 - Go to Instance Group tab on side bar

### Expected results

1 - Display the instance groups list

### Actual results

Error due to lack of permissions to read settings.

tiagodread:
Fixed with the merge of https://github.com/ansible/awx/pull/11839

Issue #11130, opened 2021-09-24
Disable editing more fields of `controlplane` group


AlanCoding:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

Users are able to modify several fields of the `controlplane` instance group, which they probably shouldn't be allowed to edit (these edits will break things).

### AWX version

19.2.2

### Installation method

docker on linux

### Modifications

no

### Ansible version

devel

### Operating system

N/A

### Web browser

Chrome

### Steps to reproduce

PATCH to `/api/v2/instance_groups/N/` where that record has the `name` of "controlplane"

try giving all the fields (individually) a non-default value:

```json
    "name": "controlplane",
    "is_container_group": false,
    "credential": null,
    "policy_instance_percentage": 98,
    "policy_instance_minimum": 0,
    "policy_instance_list": [],
    "pod_spec_override": ""
```

### Expected results

Most of these should give a 400 error

### Actual results

Only modifying `name` gives a 400

Out of the others, particularly problematic seeming ones are:

 - I can set `is_container_group` to true... that doesn't sound good
 - I can set `policy_instance_percentage` to a value lower than what was set by the installer, which is probably not desired (maybe arguable? but I doubt it)
 - I can add execution nodes to `policy_instance_list`, which is absolutely not okay

AlanCoding:
existing name validation

https://github.com/ansible/awx/blob/6a17e5b65bf9421c2e650d23817c79ce81248c6f/awx/api/serializers.py#L4935

AlanCoding:
It would be really easy to deal with this along with:

https://github.com/ansible/awx/issues/11622

tiagodread:
Fixed with the merge of #11839

PR #11839, opened 2022-03-02
Add several changes to Instance Groups

nixocio:
Add several changes to API and UI related to Instance Groups.

* Update summary_fields for DEFAULT_CONTROL_PLANE_QUEUE_NAME, and
  DEFAULT_EXECUTION_QUEUE_NAME. Rely on API validation for those fields.

* Fix Instance Group list RBAC

* Add validation for a couple of fields on the Instance Groups endpoint
	1. is_container_group
	2. policy_instance_percentage
	3. policy_instance_list

See: https://github.com/ansible/awx/issues/11130
Also: https://github.com/ansible/awx/issues/11718

### DIFF
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -105,7 +105,6 @@ from awx.api.permissions import (
     ProjectUpdatePermission,
     InventoryInventorySourcesUpdatePermission,
     UserPermission,
-    InstanceGroupTowerPermission,
     VariableDataPermission,
     WorkflowApprovalPermission,
     IsSystemAdminOrAuditor,
@@ -480,7 +479,6 @@ class InstanceGroupDetail(RelatedJobsPreventDeleteMixin, RetrieveUpdateDestroyAP
     name = _("Instance Group Detail")
     model = models.InstanceGroup
     serializer_class = serializers.InstanceGroupSerializer
-    permission_classes = (InstanceGroupTowerPermission,)
 
     def update_raw_data(self, data):
         if self.get_object().is_container_group:
--- a/awx/ui/src/screens/InstanceGroup/ContainerGroupDetails/ContainerGroupDetails.js
+++ b/awx/ui/src/screens/InstanceGroup/ContainerGroupDetails/ContainerGroupDetails.js
@@ -15,7 +15,7 @@ import { jsonToYaml, isJsonString } from 'util/yaml';
 import { InstanceGroupsAPI } from 'api';
 import { relatedResourceDeleteRequests } from 'util/getRelatedResourceDeleteDetails';
 
-function ContainerGroupDetails({ instanceGroup, defaultExecution }) {
+function ContainerGroupDetails({ instanceGroup }) {
   const { id, name } = instanceGroup;
 
   const history = useHistory();
@@ -99,8 +99,7 @@ function ContainerGroupDetails({ instanceGroup, defaultExecution }) {
               {t`Edit`}
             </Button>
           )}
-        {name !== defaultExecution &&
-          instanceGroup.summary_fields.user_capabilities &&
+        {instanceGroup.summary_fields.user_capabilities &&
           instanceGroup.summary_fields.user_capabilities.delete && (
             <DeleteButton
               ouiaId="container-group-detail-delete-button"
--- a/awx/ui/src/screens/InstanceGroup/InstanceGroupEdit/InstanceGroupEdit.test.js
+++ b/awx/ui/src/screens/InstanceGroup/InstanceGroupEdit/InstanceGroupEdit.test.js
@@ -55,10 +55,7 @@ describe('<InstanceGroupEdit>', () => {
     history = createMemoryHistory();
     await act(async () => {
       wrapper = mountWithContexts(
-        <InstanceGroupEdit
-          defaultControlPlane="controlplane"
-          instanceGroup={instanceGroupData}
-        />,
+        <InstanceGroupEdit instanceGroup={instanceGroupData} />,
         {
           context: { router: { history } },
         }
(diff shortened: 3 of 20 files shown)


awx/api/views/__init__.py
```diff
@@ -480,7 +479,6 @@ class InstanceGroupDetail(RelatedJobsPreventDeleteMixin, RetrieveUpdateDestroyAP
     name = _("Instance Group Detail")
     model = models.InstanceGroup
     serializer_class = serializers.InstanceGroupSerializer
-    permission_classes = (InstanceGroupTowerPermission,)
```
AlanCoding, line 483: I believe this was intended to case a 405 error with a DELETE request. Have you tested that a DELETE request will still give some 400-level error after this change? Because I don't see how. If you don't use the permissions class (which I think is fine... personally), then you probably need to handle it in the view, probably via the `perform_destroy` method.

nixocio: With the current change we get a 403.

```HTTP/1.1 403 Forbidden
Server: nginx
Date: Wed, 02 Mar 2022 19:52:44 GMT
Content-Length: 63
Connection: keep-alive
Vary: Accept, Accept-Language, Origin, Cookie
Allow: GET, PUT, PATCH, DELETE, HEAD, OPTIONS
X-API-Product-Version: 20.0.1.dev35+gf65fe0de86
X-API-Product-Name: AWX
X-API-Node: awx_1
X-API-Time: 0.113s
X-API-Query-Count: 3
X-API-Query-Time: 0.003s
Content-Language: en
X-API-Total-Time: 0.174s
X-API-Request-Id: 98dfbeea1de44a8db050893b2b2b8da2
Access-Control-Expose-Headers: X-API-Request-Id
```

```bash
{"detail":"You do not have permission to perform this action."}
```
I can revert those changes if we need to.

AlanCoding: Okay, as long as it doesn't actually delete, I have no preference, myself

jbradberry: I think 403 is preferable in this circumstance, where the decision to allow delete is dynamic.  405 strikes me as meaning that the endpoint doesn't support the verb in general.

jbradberry (approved):

AlanCoding (approved):

awx/ui/src/screens/InstanceGroup/ContainerGroupDetails/ContainerGroupDetails.js
```diff
@@ -99,8 +99,7 @@ function ContainerGroupDetails({ instanceGroup, defaultExecution }) {
               {t`Edit`}
             </Button>
           )}
-        {name !== defaultExecution &&
-          instanceGroup.summary_fields.user_capabilities &&
+        {instanceGroup.summary_fields.user_capabilities &&
```
AlexSCorey, line 102: Should we make this a ternary?

awx/ui/src/screens/InstanceGroup/InstanceGroupEdit/InstanceGroupEdit.test.js
```diff
@@ -70,27 +67,6 @@ describe('<InstanceGroupEdit>', () => {
     jest.clearAllMocks();
   });
 
-  test('controlplane instance group name can not be updated', async () => {
```
AlexSCorey, line 73: This seems like a good use case for an integration test to confirm that the api throws a validation error when trying to change this @tiagodread @akus062381 @InnocentK

awx/ui/src/screens/InstanceGroup/InstanceGroups.js
```diff
@@ -16,31 +17,28 @@ import ContainerGroup from './ContainerGroup';
...
+    result: { isKubernetes },
   } = useRequest(
     useCallback(async () => {
       const {
-        data: {
-          IS_K8S,
-          DEFAULT_CONTROL_PLANE_QUEUE_NAME,
-          DEFAULT_EXECUTION_QUEUE_NAME,
-        },
+        data: { IS_K8S },
```
AlexSCorey, line 31: What happens here if an org admin accesses instance groups list on a kubernetes deployment?

awx/ui/src/screens/InstanceGroup/shared/InstanceGroupForm.js
```diff
@@ -1,49 +1,25 @@
...
-    ),
-  ]);
-
+function InstanceGroupFormFields() {
   return (
     <>
       <FormField
         name="name"
-        helperText={
-          isProtected
```
AlexSCorey, line 37: I wish we didn't have to remove this but I think the refactor you've done here outweighs this little bit of lost UX.  I know you've done a lot of investigating into how to keep this and short of making some api changes that the api group doesn't support (for valid reason) I'm not real sure what else we can do.

AlexSCorey (approved):

awx/ui/src/screens/InstanceGroup/InstanceGroups.js
nixocio: This user profile can not read settings endpoint. This variable is used used to enable or not the add button for `instance groups` and `container groups` and this user will not have the privilege to add those not matter what. Then `IS_K8S` will default to initial value to false, and we do not make an extra requests to settings, and also do not display the add button.

tiagodread:
@nixocio @AlanCoding some feedback regard RBAC:

- ~A system auditor can try to associate/disassociate an instance on a IG and execute the `health check` because we allow them to do so, the API is blocking the action:~

image

image


- ~An Org admin can not associate/disassociate instances or run Health check on instances groups associated with their org, they can only see it~

~List of access the normal user has in the org~

image

~Health check~
image

~Disassociate~
image

~Visit Instance details~
image

nixocio:
hey @tiagodread, Thanks for your report. This PR did not address changes to the `instances` but `instance groups`. This behavior reported related to RBAC is also present on AWX/devel. I would suggest for us to file a dedicated issue related to that.

AlanCoding:
Yeah, those comments make sense. We need a full-stack discussion about what new type of information the UI needs to prevent those cases.

tiagodread:
Opened #11910 to address these issues

tiagodread (approved):

PR #11840, opened 2022-03-02
load job meta vars after JT extra vars

fosterseth:
##### SUMMARY

Currently users who have JT edit privileges can override the meta variables we inject into each job run

example, user can override the username by setting this JT extra var

```yaml
awx_user_name: bob
```

Is there any good reason to allow this?
This change would load in meta variables _after_ JT extra variables.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 7.0.1.dev8986+g01b41531c3
```

### DIFF
--- a/awx/main/tasks/jobs.py
+++ b/awx/main/tasks/jobs.py
@@ -823,11 +823,12 @@ class RunJob(BaseTask):
         return job.playbook
 
     def build_extra_vars_file(self, job, private_data_dir):
-        # Define special extra_vars for AWX, combine with job.extra_vars.
-        extra_vars = job.awx_meta_vars()
-
+        extra_vars = dict()
+        # load in JT extra vars
         if job.extra_vars_dict:
             extra_vars.update(json.loads(job.decrypted_extra_vars()))
+        # load in meta vars, overriding any variable set in JT extra vars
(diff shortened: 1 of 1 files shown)


AlanCoding (changes requested):
Change needs to be repeated for adhoc commands.

awx/main/tasks/jobs.py
```diff
@@ -1885,14 +1886,6 @@ def build_args(self, ad_hoc_command, private_data_dir, passwords):
         if ad_hoc_command.verbosity:
             args.append('-%s' % ('v' * min(5, ad_hoc_command.verbosity)))
 
-        extra_vars = ad_hoc_command.awx_meta_vars()
```
fosterseth, line 1888: this is dead code

fosterseth:
I noticed that extra vars supplied to ad hoc commands were marked as "unsafe"
```
!unsafe 'this_var': 123
```

whereas in regular jobs they were marked as safe

fosterseth:
if a user tries to override, we may considering doing some validation and throwing an error when they save JT, or if set at launch, when user clicks Launch. That would probably be a better user experience

AlanCoding (approved):

PR #11844, opened 2022-03-03
Adopt changes to AWX_ISOLATION_SHOW_PATHS for trust store

AlanCoding:
##### SUMMARY
Porting downstream changes of @shanemcd 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/tests/functional/api/test_settings.py
+++ b/awx/main/tests/functional/api/test_settings.py
@@ -38,7 +38,10 @@ def test_jobs_settings(get, put, patch, delete, admin):
     data.pop('AWX_ANSIBLE_CALLBACK_PLUGINS')
     put(url, user=admin, data=data, expect=200)
     response = get(url, user=admin, expect=200)
-    assert response.data['AWX_ISOLATION_SHOW_PATHS'] == []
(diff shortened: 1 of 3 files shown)


AlanCoding:
Would replace https://github.com/ansible/awx/pull/11812, not yet sure if this is the right thing to do.

shanemcd (approved):

shanemcd (changes requested):
Actually wait. I think this will break the development environment.

shanemcd:
Yeah, I get:

```
Error: mounting overlay failed "/usr/share/pki/": chown /var/lib/awx/.local/share/containers/storage/overlay-containers/ba7243af6a0b1510caaffcf1e848721ec384513532009fc63487bc803fdea0bf/userdata/overlay/3107658967/upper: invalid argument
```

I think we just want this in `production.py`.

shanemcd (approved):

AlanCoding:
This still leaves us needing to sync up other branches with the production / development settings. I'm not sure how yet, still thinking on it.

PR #11859, opened 2022-03-08
Add a CI check for the development environment

shanemcd:
This is a dedicated test that ensures the development environment is able to run the Demo Project and Demo Job Template. 

Our existing CI checks do not start any of our ancillary services, nor do they exercise the container-in-container functionality we rely on in the dev env.

### DIFF
--- a/tools/docker-compose/ansible/smoke-test.yml
+++ b/tools/docker-compose/ansible/smoke-test.yml
@@ -1,0 +1,77 @@
+---
+#
+# This is used by a CI check in GitHub Actions and isnt really
+# meant to be run locally.
+#
+# The development environment does some unfortunate things to
+# make rootless podman work inside of a docker container.
+# The goal here is to essentially tests that the awx user is
+# able to run `podman run`.
+#
+- name: Test that the development environment is able to launch a job
+  hosts: localhost
+  tasks:
+    - name: Boot the development environment
+      command: |
+        make docker-compose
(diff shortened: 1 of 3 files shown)


tools/docker-compose/ansible/smoke-test.yml
```diff
@@ -0,0 +1,78 @@
...
+      until: _result.status == 200
+      retries: 120
+      delay: 5
+
+    - name: Reset admin password
+      shell: |
+        docker exec -i tools_awx_1 bash <<EOSH
+          awx-manage createsuperuser --noinput --username=admin --email=admin@localhost
+          awx-manage update_password --username=admin --password=password
+          awx-manage create_preload_data
```
AlanCoding, line 35: This doesn't seem to be working 100% from the output of the check

```
"stderr": "CommandError: Error: That username is already taken."
```

I'm not sure _which_ of the lines this is coming from.

AlanCoding: but since the rest of it works, I think that's merely telling us that something here isn't necessary. I think the `update_password` is absolutely necessary, this wouldn't work without that.

shanemcd: Yeah, since this is a shell script, only the last return code matters. But I can take out that line.

AlanCoding (approved):

Issue #11842, opened 2022-03-03
HashiCorp Vault Signed SSH broken since update to 20.0.0


al-lac:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

We used a signed ssh key for accessing some of ours hosts. For this we use a credential of the type **HashiCorp Vault Signed SSH**. Since upgrading from 19.4.0 to 20.0.0 we cannot access hosts that need a signed ssh key anymore. 

Before we would get this in the Job Logs:
```bash
Identity added: /runner/artifacts/6919/ssh_key_data (/runner/artifacts/6919/ssh_key_data)
Certificate added: /runner/artifacts/6919/ssh_key_data-cert.pub (vault-token-AWX-<ID>)
```

Now we only get the following:
```bash
Identity added: /runner/artifacts/6928/ssh_key_data (/runner/artifacts/6928/ssh_key_data)
```

### AWX version

20.0.0

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs

### Installation method

kubernetes

### Modifications

no

### Steps to reproduce

- Start a Job Template with a Credential of the Category **HashiCorp Vault Signed SSH**
- Wait for the connection to the host

### Expected results

Connection to the host should be established

### Actual results

Error when trying to connect to a host:

```
Failed to connect to the host via ssh: user@10.10.0.1: Permission denied (publickey,gssapi-keyex,gssapi-with-mic)."
```

nilsding:
I looked a bit into it, and to me looks like the `shutil.rmtree(artifact_dir)` introduced in  #11472 inside `awx/main/tasks/receptor.py` is the culprit here... the cert file is still created inside the artifact dir, but it's deleted immediately afterwards.  Removing that `rmtree` call makes the jobs pass again

AlanCoding:
Yes, I am already aware of problems introduced by that `shutil.rmtree` as @nilsding and I suspect it is the issue here.

That messes things up for `node_type='hybrid'` nodes. In AWX, all nodes should be `node_type='control'`. Anyway, we need a condition to not run that block of code if it's running locally. I had that change in some PR but it got mixed in with other work and tied up.

al-lac:
[...]

PR #11860, opened 2022-03-08
Do not remove artifacts for local work

AlanCoding:
##### SUMMARY
This is cherry-picking a fix from https://github.com/ansible/awx/pull/11641 because that work got stranded and this is still not fixed.

Connect https://github.com/ansible/awx/issues/11842, which describes (predictably) unrelated fallout which is a consequence of this.

What's going on - `only_transmit_kwargs` is used as a parameter to ansible-runner when running local work. This means that the private_data_dir is not streamed to the receptor process, and the one on disk is used in-place. However, this does not turn off streaming of artifacts back from receptor (maybe it should). The artifacts directory is _inside_ of the private_data_dir. This line deletes the artifacts directory so that any file deletions on the remote node will be picked up. But this doesn't work when it's running locally. It "streams" the artifacts back, but... from the same directory it started in. So deleting stuff before the job run deletes the stuff in the context where the job runs.

But this issue has also made it clear that we don't have test coverage for certs... which we should also do.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/tasks/receptor.py
+++ b/awx/main/tasks/receptor.py
@@ -326,7 +326,7 @@ class AWXReceptorJob:
         # Artifacts are an output, but sometimes they are an input as well
         # this is the case with fact cache, where clearing facts deletes a file, and this must be captured
         artifact_dir = os.path.join(self.runner_params['private_data_dir'], 'artifacts')
-        if os.path.exists(artifact_dir):
+        if self.work_type != 'local' and os.path.exists(artifact_dir):
             shutil.rmtree(artifact_dir)
 
         resultsock, resultfile = receptor_ctl.get_work_results(self.unit_id, return_socket=True, return_sockfile=True)


shanemcd:
> But this issue has also made it clear that we don't have test coverage for certs... which we should also do.

Let's make sure that we capture this somewhere. cc @Zokormazo

AlanCoding:
This is a fairly bad logical flaw and shouldn't be left to just sit here. I will put it on my agenda to start work on a testing strategy to launch a job that will do a simple `ssh-add -l` command and inspect the output. This was already being discussed to add coverage of other things, and it may contain information related to certs.

al-lac (approved):
We tested this change on multiple AWX servers and it solved our issue of SSH Signing not working.

nixocio (approved):
@AlanCoding explained me the change. LGTM.

john-westcott-iv (approved):

PR #11865, opened 2022-03-08
Add user-defined environment variables to ansible-galaxy commands

AlanCoding:
##### SUMMARY
Allows the user to use a proxy server for `ansible-galaxy` commands by letting them set whatever environment variables they want. We already have a set of environment variables we use for this command, and I don't care if the user changes them, so those are absorbed into this new setting.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION
Here's what OPTIONS look like for this under PUT

```json
            "GALAXY_TASK_ENV": {
                "type": "nested object",
                "required": true,
                "label": "Environment Variables for Galaxy Commands",
                "help_text": "Additional environment variables set for invocations of ansible-galaxy within project updates. Useful if you must use a proxy server for ansible-galaxy but not git.",
                "category": "Jobs",
                "category_slug": "jobs",
                "placeholder": {
                    "HTTP_PROXY": "myproxy.local:8080"
                },
                "default": {
                    "ANSIBLE_FORCE_COLOR": "false",
                    "GIT_SSH_COMMAND": "ssh -o StrictHostKeyChecking=no"
                },
                "child": {
                    "type": "string",
                    "required": true,
                    "read_only": false
                }
```

Here's what the UI looks like:

Screenshot from 2022-03-08 10-32-25

I've tested this by setting the HTTPS_PROXY to an invalid URl, and it fails exactly as expected.

### DIFF
--- a/awx/ui/src/screens/Setting/shared/data.allSettingOptions.json
+++ b/awx/ui/src/screens/Setting/shared/data.allSettingOptions.json
@@ -232,25 +232,25 @@
         }
       },
       "GALAXY_TASK_ENV": {
-              "type": "nested object",
-               "required": true,
-              "label": "Environment Variables for Galaxy Commands",
-               "help_text": "Additional environment variables set for invocations of ansible-galaxy within project updates. Useful if you must use a proxy server for ansible-galaxy but not git.",
-               "category": "Jobs",
-               "category_slug": "jobs",
-               "placeholder": {
-                    "HTTP_PROXY": "myproxy.local:8080"
-                },
-               "default": {
-                   "ANSIBLE_FORCE_COLOR": "false",
-                    "GIT_SSH_COMMAND": "ssh -o StrictHostKeyChecking=no"
-              },
-                "child": {
-                  "type": "string",
-                    "required": true,
-                  "read_only": false
-              }
-             },
+        "type": "nested object",
+         "required": true,
+        "label": "Environment Variables for Galaxy Commands",
+         "help_text": "Additional environment variables set for invocations of ansible-galaxy within project updates. Useful if you must use a proxy server for ansible-galaxy but not git.",
+         "category": "Jobs",
+         "category_slug": "jobs",
+         "placeholder": {
+              "HTTP_PROXY": "myproxy.local:8080"
+          },
+         "default": {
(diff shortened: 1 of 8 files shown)


AlanCoding:
@nixocio I wanted to ask for some UI help making an assessment here. The setting shows up, but perhaps some more logic is needed on the UI. This is almost a direct repeat of the `AWX_TASK_ENV` setting, which has some supporting UI code:

https://github.com/ansible/awx/blob/a988ad0c4ed592c9ded902298ff84fdda4d82cf6/awx/ui/src/screens/Setting/Jobs/JobsEdit/JobsEdit.js#L87

I could repeat the pattern with this setting, but I would like more info about what the function is. I told @mabashian that this didn't _appear_ to need UI work, but that might not be the whole story.

nixocio:
@AlanCoding, I believe we will need to add some changes to `JobsEdit.js`. I am looking at this now.

AlanCoding:
I looked up when that was introduced for AWX_TASK_ENV, and it didn't make it clear what function that was playing. But I feel like it had something to do with correctly reformatting YAML or something like that. Anyway, these changes look like what I would expect following the patterns of that prior settings :+1:

AlanCoding:
I'm completely happy with this patch, and the only thing blocking is the need for review.

nixocio (approved):
I added a commit to add change to the JobsEdit as well. I verified UI changes.

awx/ui/src/screens/Setting/shared/data.allSettingOptions.json
```diff
@@ -3942,6 +3962,26 @@
           "type": "string"
         }
       },
+      "GALAXY_TASK_ENV": {
+              "type": "nested object",
+               "required": true,
+              "label": "Environment Variables for Galaxy Commands",
```
jbradberry: you've got some odd indents going on

jbradberry (approved):

AlanCoding:
I rebuilt the UI and tinkered with it. Everything about the setting is working just fine (only takes JSON, but is the same as other similar settings). Merging now.

Issue #11622, opened 2022-01-27
Hop nodes instances can be modified


tiagodread:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

After merging https://github.com/ansible/awx/pull/11591 the hop nodes are now exposed in `/api/v2/instances/`.

Example:

```
 {
      "id": 4,
      "type": "instance",
      "url": "/api/v2/instances/4/",
      "related": {
          "jobs": "/api/v2/instances/4/jobs/",
          "instance_groups": "/api/v2/instances/4/instance_groups/"
      },
      "uuid": "00000000-0000-0000-0000-000000000000",
      "hostname": "receptor-hop",
      "created": "2022-01-26T19:15:07.439405Z",
      "modified": "2022-01-26T19:27:58.447951Z",
      "last_seen": "2022-01-27T13:36:46.163205Z",
      "last_health_check": null,
      "errors": "",
      "capacity_adjustment": "1.00",
      "version": "",
      "capacity": 0,
      "consumed_capacity": 0,
      "percent_capacity_remaining": 0.0,
      "jobs_running": 0,
      "jobs_total": 0,
      "cpu": 0,
      "memory": 0,
      "cpu_capacity": 0,
      "mem_capacity": 0,
      "enabled": false,
      "managed_by_policy": false,
      "node_type": "hop"
}
```

and now is possible to `PUT` or `PATCH` the values through API

```
{
    "capacity_adjustment": "1.00",
    "enabled": false,
    "managed_by_policy": false
}
```
although setting capacity to 0 enabld to false is not making the job to faul, @jbradberry do we want allow users to patch/put on hop node instances?

### AWX version

devel

### Installation method

docker on linux

### Modifications

no

### Web browser

Chrome

### Steps to reproduce

1. Go to /api/v2/instances/<id>/
2. patch any value:
```
{
    "capacity_adjustment": 1.0,
    "enabled": true,
    "managed_by_policy": true
}
```

### Expected results

These fields to be not modifyable for hop nodes?

### Actual results

1. The user can change those values
2. This is not making the job to fail
3. The hop nodes instances are showing as `disabled` in the mesh view (may confuse users)

jbradberry:
[...]

PR #11882, opened 2022-03-09
Add validation for hop nodes

nixocio:
Add validation for hop nodes

See: https://github.com/ansible/awx/issues/11622


AlanCoding (approved):
This should get some test coverage, but it would be fine to do that on the integration test side.

Does the UI need any more information to know not to show the capacity slider or the enable/disable button? I'm sure that could be done with the node_type field, so I don't expect we need anything here.

nixocio:
hey @AlanCoding, we do not display edit button or slider for hop nodes.

image

tiagodread (approved):

Issue #11881, opened 2022-03-09
awxkit export/import should allow exporting of applications


john-westcott-iv:
##### ISSUE TYPE
 - Feature Idea

thenets:
@john-westcott-iv this can be closed, right?

PR #11887, opened 2022-03-10
Enabling export of applications through awxkit

john-westcott-iv:
##### SUMMARY

This change modifies the api to add organization to the related field on the applications view.
It also enables the export of applications through awxkit.

Addresses https://github.com/ansible/awx/issues/11881

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API
 - CLI

##### AWX VERSION

```
awx: 19.5.2.dev216+gad16f4545b
```

##### ADDITIONAL INFORMATION

```
awx export --applications
{
     "applications": [
          {
               "name": "My App",
               "description": "",
               "client_type": "public",
               "redirect_uris": "",
               "authorization_grant_type": "password",
               "skip_authorization": false,
               "organization": {
                    "name": "Default",
                    "type": "organization"
               },
               "natural_key": {
                    "organization": {
                         "name": "Default",
                         "type": "organization"
                    },
                    "name": "My App",
                    "type": "o_auth2_application"
               }
          }
     ]
}

```

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -865,7 +865,7 @@ class UnifiedJobSerializer(BaseSerializer):
         if 'elapsed' in ret:
             if obj and obj.pk and obj.started and not obj.finished:
                 td = now() - obj.started
-                ret['elapsed'] = (td.microseconds + (td.seconds + td.days * 24 * 3600) * 10**6) / (10**6 * 1.0)
+                ret['elapsed'] = (td.microseconds + (td.seconds + td.days * 24 * 3600) * 10 ** 6) / (10 ** 6 * 1.0)
             ret['elapsed'] = float(ret['elapsed'])
         # Because this string is saved in the db in the source language,
         # it must be marked for translation after it is pulled from the db, not when set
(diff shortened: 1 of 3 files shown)


awx/api/serializers.py
```diff
@@ -865,7 +865,7 @@ def to_representation(self, obj):
         if 'elapsed' in ret:
             if obj and obj.pk and obj.started and not obj.finished:
                 td = now() - obj.started
-                ret['elapsed'] = (td.microseconds + (td.seconds + td.days * 24 * 3600) * 10**6) / (10**6 * 1.0)
+                ret['elapsed'] = (td.microseconds + (td.seconds + td.days * 24 * 3600) * 10 ** 6) / (10 ** 6 * 1.0)
```
john-westcott-iv: This was strictly a black reformatting change in order to commit to this file.

jbradberry (approved):

awx/api/serializers.py
```diff
@@ -1261,6 +1261,7 @@ def get_related(self, obj):
             dict(
                 tokens=self.reverse('api:o_auth2_application_token_list', kwargs={'pk': obj.pk}),
                 activity_stream=self.reverse('api:o_auth2_application_activity_stream_list', kwargs={'pk': obj.pk}),
+                organization=self.reverse('api:organization_detail', kwargs={'pk': obj.organization.pk}),
```
jbradberry: This is a nullable field, so you need to check if it exists first then add it to the dict.  Also, I'd recommend using `obj.organization_id` to avoid the potential query to get the organization object.

jbradberry (changes requested):

awx/api/serializers.py
john-westcott-iv: Fixed.

jbradberry (approved):

PR #11905, opened 2022-03-15
Simple patches to make jobs robust to database restarts

AlanCoding:
##### SUMMARY
The goal of this is to make it so that a small amount of downtime from postgres (less than 5 seconds) will not fail long-running jobs (multiple hours).

One time in the distant past, this was a design goal, and this goal is obvious if you read the code of `update_model`. Systems have shifted significantly since that time, and several novel failure points have been introduced.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
My current testing method looks like this:

```bash
echo "select pg_terminate_backend(pid) from pg_stat_activity where datname='awx';" | awx-manage dbshell
```

This is not the _only_ type of database hiccup that might happen, and we should test other things (like a simple restart), this is just my first stop. I have some manual steps, which I will consolidate into an integration test:

 - Launch a job that sleeps for 60 seconds
 - Wait for first lines of output to appear
 - Run the above command to terminate database connections
 - Wait for the rest of the minute to finish
   - **before** this patch the job is marked as failed by the reaper
   - **after** this patch the job completes successfully.

The logs during the event are a little beat up, but the main dispatcher processes remaining running through the whole thing.

### DIFF
--- a/awx/main/dispatch/worker/base.py
+++ b/awx/main/dispatch/worker/base.py
@@ -136,9 +136,10 @@ class AWXConsumerRedis(AWXConsumerBase):
 class AWXConsumerPG(AWXConsumerBase):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.pg_down_time = 0.0
-        self.pg_is_down = False
         self.pg_max_wait = settings.DISPATCHER_DB_DOWNTOWN_TOLLERANCE
+        # if no successful loops have ran since startup, then we should fail right away
+        self.pg_is_down = True  # set so that we fail if we get database errors on startup
+        self.pg_down_time = time.time() - self.pg_max_wait  # allow no grace period
 
     def run(self, *args, **kwargs):
         super(AWXConsumerPG, self).run(*args, **kwargs)
@@ -163,6 +164,7 @@ class AWXConsumerPG(AWXConsumerBase):
                 logger.warning("Stale Postgres message bus connection, reconnecting")
                 continue
             except (db.DatabaseError, psycopg2.OperationalError):
+                # If we have attained stady state operation, tolerate short-term database hickups
                 if not self.pg_is_down:
                     logger.exception(f"Error consuming new events from postgres, will retry for {self.pg_max_wait} s")
                     self.pg_down_time = time.time()
--- a/awx/main/dispatch/worker/base.py
+++ b/awx/main/dispatch/worker/base.py
@@ -138,7 +138,7 @@ class AWXConsumerPG(AWXConsumerBase):
         super().__init__(*args, **kwargs)
         self.pg_down_time = 0.0
         self.pg_is_down = False
-        self.pg_max_wait = 60.0
+        self.pg_max_wait = settings.DISPATCHER_DB_DOWNTOWN_TOLLERANCE
 
     def run(self, *args, **kwargs):
         super(AWXConsumerPG, self).run(*args, **kwargs)
@@ -156,9 +156,9 @@ class AWXConsumerPG(AWXConsumerBase):
                         init = True
                     for e in conn.events():
                         self.process_task(json.loads(e.payload))
+                        self.pg_is_down = False
                     if self.should_stop:
                         return
-                    self.pg_is_down = False
             except psycopg2.InterfaceError:
                 logger.warning("Stale Postgres message bus connection, reconnecting")
                 continue
@@ -175,6 +175,8 @@ class AWXConsumerPG(AWXConsumerBase):
                     if self.should_stop:
                         return
                     time.sleep(0.1)
+                for conn in db.connections.all():
+                    conn.close_if_unusable_or_obsolete()
 
 
 class BaseWorker(object):
--- a/awx/main/dispatch/worker/base.py
+++ b/awx/main/dispatch/worker/base.py
@@ -138,6 +138,7 @@ class AWXConsumerPG(AWXConsumerBase):
         super().__init__(*args, **kwargs)
         self.pg_down_time = 0.0
         self.pg_is_down = False
+        self.pg_max_wait = 60.0
 
     def run(self, *args, **kwargs):
         super(AWXConsumerPG, self).run(*args, **kwargs)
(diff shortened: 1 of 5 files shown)


awx/main/dispatch/worker/base.py
```diff
@@ -152,9 +157,18 @@ def run(self, *args, **kwargs):
...
                         return
+                    self.pg_is_down = False
             except psycopg2.InterfaceError:
                 logger.warn("Stale Postgres message bus connection, reconnecting")
                 continue
+            except (db.DatabaseError, psycopg2.OperationalError):
+                if not self.pg_is_down:
+                    logger.exception("Postgres error processing tasks")
+                    self.pg_down_time = time.time()
+                    self.pg_is_down = True
```
AlanCoding, line 171: After this, I believe it would be best to pause for a second or two, so that we're not pummeling a recovering database with failing requests. However, while this is happening we probably need to have a sub-loop checking `self.should_stop` at a faster rate.

AlanCoding, line 168: After this, I believe it would be best to pause for a second or two, so that we're not pummeling a recovering database with failing requests. However, while this is happening we probably need to have a sub-loop checking `self.should_stop` at a faster rate.

awx/main/dispatch/worker/base.py
```diff
@@ -152,9 +157,18 @@ def run(self, *args, **kwargs):
...
+                if time.time() - self.pg_down_time > 60.0:
```
AlanCoding: We should make the 60 configurable, and maybe we share that constant with `update_model`. But I'm worried about doing any settings references in that cancel_callback (due to threading issues).

AlanCoding, line 169: We should make the 60 configurable, and maybe we share that constant with `update_model`. But I'm worried about doing any settings references in that cancel_callback (due to threading issues).

rcdlinux:
It would really be great to have this improvement into AWX and Ansible Tower!  A customer is testing some heavy jobs with Tower 3.8 and he has stumbled with this 'issue' with jobs failing after database restarts (or database recovering from a down node to an up node when in a Postgresql cluster).

Congratulations and Thank you very much for providing this patch @AlanCoding !  I hope it can be merged to Awx/Tower soon as possible.

awx/main/dispatch/worker/base.py
```diff
@@ -150,11 +156,27 @@ def run(self, *args, **kwargs):
...
+                    self.pg_down_time = time.time()
+                    self.pg_is_down = True
+                if time.time() - self.pg_down_time > self.pg_max_wait:
+                    logger.warning(f"Postgres event consumer has not recovered in {self.pg_max_wait} s, exiting")
+                    return
+                # Wait for a second before next attempt, but still listen for any shutdown signals
+                for i in range(10):
+                    if self.should_stop:
+                        return
+                    time.sleep(0.1)
```
AlanCoding, line 179: This additional logic may be unnecessary because the signal handler ends with `SystemExit`. That may also be making the other checks of `self.should_stop` dead code as well. For defensiveness, I currently opt to keep it.

brunowcs:
Hi @AlanCoding I´m test this code with Patroni HA PostgreSQL with faiolver between node, but only finished my job for me when change time.sleep(0.1) to time.sleep(1).

Job finished, but status is "Task was marked as running but was not present in the job queue, so it has been marked as failed."

I think something here about log_lifecycle.

2022-03-25 20:20:41,349 ERROR    [bc5446138bd44a158a5db67d0962c3a7] awx.main.dispatch Worker failed to run task awx.main.tasks.RunJob(*[15482], **{}
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/tasks.py", line 1575, in run
    self.final_run_hook(self.instance, status, private_data_dir, fact_modification_times)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/tasks.py", line 1968, in final_run_hook
    super(RunJob, self).final_run_hook(job, status, private_data_dir, fact_modification_times)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/tasks.py", line 1227, in final_run_hook
    instance.log_lifecycle("finalize_run")
AttributeError: 'NoneType' object has no attribute 'log_lifecycle'

awx/main/dispatch/worker/base.py
```diff
@@ -150,11 +157,28 @@ def run(self, *args, **kwargs):
...
                 continue
+            except (db.DatabaseError, psycopg2.OperationalError):
+                # If we have attained stady state operation, tolerate short-term database hickups
+                if not self.pg_is_down:
+                    logger.exception(f"Error consuming new events from postgres, will retry for {self.pg_max_wait} s")
+                    self.pg_down_time = time.time()
+                    self.pg_is_down = True
+                if time.time() - self.pg_down_time > self.pg_max_wait:
+                    logger.warning(f"Postgres event consumer has not recovered in {self.pg_max_wait} s, exiting")
+                    return
```
AlanCoding: oh this should be a `raise`, not a `return`

AlanCoding:
@brunowcs that error just means to me that it failed to recover in the allotted amount of time. Was database access gone for over 40 seconds? That could explain the behavior you saw.

brunowcs:
> @brunowcs that error just means to me that it failed to recover in the allotted amount of time. Was database access gone for over 40 seconds? That could explain the behavior you saw.

Hi Alan, Thanks for this feature. 

my db downtime is very fast (5s until 15s) ,my job print output success, but status not.

2022-03-28 11:47:42,787 ERROR    [ca793dfae4014d94b83d80e368334b6e] awx.main.dispatch Error consuming new events from postgres, will retry for 120 s
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/dispatch/worker/base.py", line 158, in run
    for e in conn.events():
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/dispatch/__init__.py", line 39, in events
    self.conn.poll()
psycopg2.OperationalError: server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.

.............
...............
2022-03-28 11:47:45,319 DEBUG    [b552bdc8089c40c2ba98e54d7959e806] awx.analytics.performance request: <WSGIRequest: GET '/'>, response_time: 0.022s
2022-03-28 11:47:47,085 INFO     [-] awx.main.wsbroadcast Active instance with hostname host-001 is registered.
2022-03-28 11:47:47,098 WARNING  [-] awx.main.wsbroadcast Adding {'host-002', 'host-003'} to websocket broadcast list

brunowcs:
tower-error

shanemcd (approved):

CMPluto:
Hi @AlanCoding  & @CFSNM 
On Which AWX release this change has been implemented ?
I am using Awx 21.5.0 but still facing the same issue.

Job finished, but status is "Task was marked as running but was not present in the job queue, so it has been marked as failed."

pls suggest

AlanCoding:
^ that message is a catch-all for way too many problems. Honestly we're moving towards architecture changes so this won't happen at all. In these cases, what we firstly seek to know is the receptor status of the job, which is probably killed due to some reason but it is still a struggle to collect that error properly right now.

CMPluto:
@AlanCoding 
How can I bring the solution on my existing AWX?
Shall I upgrade the AWX Version from 21.5.0  to latest 21.13.0 version ? 
or the new upcoming AWX version will have the above change in architecture ?

Issue #10243, opened 2021-05-20
Unable to use awx collection with LDAP user, due to token creation


jhampson-dbre:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

When trying to schedule a job with awx.awx.tower_schedule, I am unable to authenticate to Tower using username/password with LDAP authentication provider. We are using Ansible Tower, but based on the module error message, this appears to be a problem with awx.awx collection attempting to create an Oauth token instead of using basic auth for the request.

##### ENVIRONMENT
* AWX version: Tower 3.7.1
* AWX install method: Bare metal
* Ansible version:  2.9.21
* Operating System:  CentOS Linux release 7.9.2009
* Web Browser: N/A
* awx.awx collection version: 19.1.0

##### STEPS TO REPRODUCE

Attempt to schedule a job using awx.awx.tower_schedule with username/password authentication for LDAP user. Instead the task errors with `OAuth2 Tokens cannot be created by users associated with an external authentication provider`.  We are using Ansible Tower, but based on the module error stack trace the problem appears to be with how authentication is implemented for awx.awx.

Example playbook below.

##### EXPECTED RESULTS

The schedule is created successfully in Tower

##### ACTUAL RESULTS

awx.awx.tower_schedule module fails with the following error:
```
    "msg": "Failed to get token: HTTP Error 403: Forbidden",
    "response": "{\"detail\":\"(access_denied) OAuth2 Tokens cannot be created by users associated with an external authentication provider (ldap)\"}"
```

##### ADDITIONAL INFORMATION

Playbook used for testing
```yaml
- hosts: localhost
  connection: local
  gather_facts: false
  tasks:
    - name: Schedule a job test
      awx.awx.tower_schedule:
        name: "Schedule a job test"
        state: present
        unified_job_template: "Demo Job"
        rrule: "{{ query('awx.awx.tower_schedule_rrule', 'week', start_date='2021-12-19 13:05:51') }}"
        tower_host: https://tower.company.example.com/#/login
        tower_username: myuser
        tower_password: mypassword
      register: result
```

Output from failed playbook run
```
(demo_venv) [vagrant@a9d8b915b6ea jobs_local]$ ansible-playbook ./scheduler_demo.yml -vvv
ansible-playbook 2.9.21
  config file = /home/vagrant/.ansible.cfg
  configured module search path = ['/home/vagrant/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
[...]

PR #11906, opened 2022-03-15
Add force_basic_auth to allow basic authentication without creating t…

rchicoli:
##### SUMMARY

This is "related https://github.com/ansible/awx/issues/10243"

The code changes add the `controller_force_basic_auth` option to allow basic authentication without the need of creating tokens.
Additionally the token feature could be disabled, which breaks the functionality of this collection at the moment.

**Reasons**
* Depending on the implementation, if a user is blocked in the external authentication system (e.g. LDAP), the token would still be valid. For that reason, sometimes there are policies against using tokens.
* Besides other ansible modules and the AWX API itself already support basic authentications.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - Collection

##### AWX VERSION

```
20.0.2.dev52+g70dfb68976
```

##### ADDITIONAL INFORMATION

It is my first time contributing to this project, so please, let me know, if I am missing something

### DIFF
--- a/awx_collection/plugins/doc_fragments/auth.py
+++ b/awx_collection/plugins/doc_fragments/auth.py
@@ -32,6 +32,12 @@ options:
     - If value not set, will try environment variable C(CONTROLLER_PASSWORD) and then config files
     type: str
     aliases: [ tower_password ]
+  controller_force_basic_auth:
+    description:
+    - Force the sending of the Basic authentication header.
+    - This can be used only if C(controller_oauthtoken) is not set, otherwise it will be ignored.
+    type: bool
+    aliases: [ tower_force_basic_auth ]
   controller_oauthtoken:
     description:
     - The OAuth token to use.
--- a/awx_collection/plugins/module_utils/controller_api.py
+++ b/awx_collection/plugins/module_utils/controller_api.py
@@ -3,7 +3,7 @@ from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
 from ansible.module_utils.basic import AnsibleModule, env_fallback
-from ansible.module_utils.urls import Request, SSLValidationError, ConnectionError
+from ansible.module_utils.urls import Request, SSLValidationError, ConnectionError, basic_auth_header
 from ansible.module_utils.parsing.convert_bool import boolean as strtobool
 from ansible.module_utils.six import PY2
 from ansible.module_utils.six import raise_from, string_types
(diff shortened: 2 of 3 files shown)


fosterseth:
@rchicoli Hi! 
Thanks for taking the time to open a PR. We're taking a look at this, can you please elaborate more on the use case here? thank you

rchicoli:
As you might guess, I do not have admin permissions to AWX/Tower to allow the external users to create oauth2 tokens.
No matter how I try to convince the security team or even if this is a short-lived token.

https://github.com/ansible/awx/blob/devel/awx_collection/plugins/module_utils/controller_api.py#L592,L601
If I get it right, the first basic authentication step will request the /api/v2/tokens/ for generating an oauth token to be used for the running session. That is the problem!

In order to circumvent that issue, we could provide an extra parameter like controller_force_basic_auth, which would not try to create a token, but send an authorization header together with each request.

john-westcott-iv:
@rchicoli  our apologies but this PR needs a rebase, is that something you could do? Please let us know if you need help or no longer have time to work on this.

rchicoli:
I did rebase the commit and force pushed the code again, though I haven't spent too much time  to check what has changed since last year. If the CI fails I could invest some time to fix it next week. Thanks

john-westcott-iv:
Thanks, all tests are successful!

awx_collection/plugins/doc_fragments/auth.py
```diff
@@ -32,6 +32,12 @@ class ModuleDocFragment(object):
     - If value not set, will try environment variable C(CONTROLLER_PASSWORD) and then config files
     type: str
     aliases: [ tower_password ]
+  controller_force_basic_auth:
```
TheRealHaoLiu, line 35: like the CONTROLLER_USERNAME and CONTROLLER_PASSWORD we might want to be able to use this as a environment variable

gundalow:
@rchicoli Thanks for this PR, what do you think of @TheRealHaoLiu's comment above?

awx_collection/plugins/doc_fragments/auth.py
rchicoli: it is the case already:
* https://github.com/ansible/awx/blob/6e0cff669d53688b6122d9b67d2ef85508bda137/awx_collection/plugins/module_utils/controller_api.py#L67

Maybe the description could be extended.

rchicoli:
> @rchicoli Thanks for this PR, what do you think of @TheRealHaoLiu's comment above?

The env option was implemented already, let me know if this PR should be merged, then i could rebase it again and extend the description.
Though I will be able to update this PR only next week.

AlexM566:
is this merged anywhere ?  the controller_force_basic_auth is not available in the awz module ...

PR #11914, opened 2022-03-16
add ID as default filter if no other filtering criteria is provided

### LINKED ISSUE #11740: Instances endpoint returning different payload order when sort is applied

### Summary

This issue came while working on https://github.com/ansible/awx/pull/11653

basically, whenever you sort the instance list by `Status` `/#/instances?instance.order_by=errors` and then, trying to disable/enable the instance, the payload order changes all the time.

### AWX version

devel

- [X] API

### Installation method

docker on linux

### Modifications

no

### Ansible version

_No response_

### Operating system

Ubuntu LTS

### Web browser

Chrome

### Steps to reproduce

1. Go to top level instances list
2. Sort by Status
3. toggle the instance on the list (with the sort applied)

### Expected results

After patching to enable/disable the instance, the order of the list should respect the sorting applied and always bring the results in the same order

### Actual results

After toggling the instance it shows the instance in a different order

### Additional information

Uploading 152869116-8efcd252-8353-4a80-9cc0-cb6ab580066e.mp4…

rebeccahhh:
have instances be filtered by ID in case of no filtering criteria passed in per the sorting issues present in https://github.com/ansible/awx/issues/11740

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 20.0.2
```


shanemcd:
@rebeccahhh Are you looking into the test failures here?

rebeccahhh:
@shanemcd yes, changes incoming to prevent other things from being affected.

awx/api/filters.py
```diff
@@ -395,11 +395,11 @@ def filter_queryset(self, request, queryset, view):
                         order_by = value.split(',')
                     else:
                         order_by = (value,)
-            if order_by is None:
-                order_by = self.get_default_ordering(view)
+            default_order_by = self.get_default_ordering(view)
+            # glue the order by and default order by together so that the default is the backup option
+            order_by = list(order_by or []) + list(default_order_by or [])
```
AlanCoding, line 403: With the change to the view ordering, I'm not sure if this change is necessary.  I also don't like that it would be changing the filter class behavior in the case that `order_by` is `None`.

jbradberry: This change is absolutely necessary.  The original problem is when someone has clicked on the 'status' header, the sort is unstable.  This is because the usual case is that the value is the empty string.  Our goal here is to have the parameters passed to Django's `.order_by()` be `['errors', 'id']`, so that there is a fall-back field resulting in a stable sort.

jbradberry: Previously when the UI passes in _any_ parameter for `order_by`, the default ordering specified by the view gets thrown away entirely.  Now it will be appended to the end of any parameters passed in by the UI as a fall back.

jbradberry: There _may_ be instances where this changes the actual ordering that a view results in, but it can only come up when a user-requested sort is used and _only_ as the final fall-back field, so I think it will be quite rare and unobtrusive.

AlanCoding: This is going to impact _all_ list views. I see what you're saying - the prior behavior was that when `order_by` was None it used the view's default ordering (seems unnecessary, but not something this is changing). My last comment was wrong.

But I would like to see a test or two, because this could be a relatively high-impact change. While the issue filed was for instances, I'd expect that similar bugs can be seen in other parts of the UI (order projects by scm_type, right?)

jbradberry: > But I would like to see a test or two, because this could be a relatively high-impact change. While the issue filed was for instances, I'd expect that similar bugs can be seen in other parts of the UI (order projects by scm_type, right?)

Probably.

rebeccahhh: I am currently writing some tests. If there are any particular scenarios you're concerned about let me know so I can account for them specifically.

fosterseth (approved):

CFSNM:
Created PR in tower-qa to add some automated test to cover this issue: https://github.com/ansible/tower-qa/pull/7346

awx/api/filters.py
AlanCoding: @rebeccahhh Create a decent number (more than 5) projects with randomly generated names. Use the test fixtures to simulate a GET to the projects list endpoint, and loop over the results, assuring that they have the default ordering applied, which is by id, which comes from the model definition.

awx/api/views/__init__.py
```diff
@@ -364,6 +364,7 @@ class InstanceList(ListAPIView):
     model = models.Instance
     serializer_class = serializers.InstanceSerializer
     search_fields = ('hostname',)
+    ordering = ('id',)
```
AlanCoding, line 368: The default ordering of the model is by hostname

https://github.com/ansible/awx/blob/56f51eebce7004630394c1c76a15eaebd652072e/awx/main/models/ha.py#L142

I don't see any reason to introduce this ordering by id.

sarabrajsingh: doesn't this need to be `id` so that the sorting works with other drop-downs/views?

CFSNM:
Hi, I left this comment in the QE PR. Can anybody take a look?
https://github.com/ansible/tower-qa/pull/7346#issuecomment-1083263071

awx/api/views/__init__.py
jbradberry: @AlanCoding the issue is when the user overrides sorting by clicking on the table headers.  The default model ordering is thrown out in that case, and the reason this has been brought up is for situations where the user is trying to sort by something that has a lot of indistinguishable values, e.g. the errors field.  We absolutely need this view default ordering, because that is what we are using in the changes in this PR to build up a multi-field set of sorting criteria.  An argument could be made that maybe we should try to do something clever and fold in _all_ sources of default ordering, but that would have been even more complex and potentially fragile.

CFSNM (approved):
Merged PR https://github.com/ansible/tower-qa/pull/7346 to add QE coverage. OK to merge this one

rebeccahhh:
we already have testing for ordering surrounding instances and instance groups, I don't think there's much more to cover here, anyone disagreeing ping me so I can add more testing to cover it (also interactive rebased so that there are only 2 commits, one for the bugfix and one for the tests).

awx/main/tests/functional/test_projects.py
```diff
@@ -408,3 +408,113 @@ def test_project_delete(delete, organization, admin_user):
...
+    project_ids = []
+    # TODO: ask for an order by here that doesn't apply
+    results = get(reverse('api:project_list'), objects.superusers.admin, QUERY_STRING='order_by=%s' % order_by).data['results']
+    for x in range(len(results)):
+        project_names.append(results[x]['name'])
+        project_ids.append(results[x]['id'])
+    assert project_names == expected_names and project_ids == expected_ids
+
+
+@pytest.mark.parametrize('order_by', ('name', '-name'))
```
rebeccahhh, line 438: @AlanCoding I think the below test covers what you were hoping for it to cover.

awx/main/tests/functional/test_projects.py
```diff
@@ -408,3 +408,46 @@ def test_project_delete(delete, organization, admin_user):
...
+@pytest.mark.django_db
+def test_project_list_ordering_with_duplicate_names(get, order_by, organization_factory):
```
kialam, line 440: 👍

awx/main/tests/functional/test_projects.py
```diff
@@ -408,3 +408,46 @@ def test_project_delete(delete, organization, admin_user):
...
+@pytest.mark.parametrize(
+    'order_by, expected_names, expected_ids',
+    [
+        ('name', ['alice project', 'bob project', 'shared project'], [1, 2, 3]),
+        ('-name', ['shared project', 'bob project', 'alice project'], [3, 2, 1]),
+    ],
+)
+@pytest.mark.django_db
+def test_project_list_ordering_by_name(get, order_by, expected_names, expected_ids, organization_factory):
+    'ensure sorted order of project list is maintained correctly when the requested order is invalid or not applicable'
```
kialam, line 422: 👍

PR #11923, opened 2022-03-18
Adding ability to run user level pre-commit hooks

john-westcott-iv:
##### SUMMARY

This change allows users to place additional pre-commit hook scripts in a directory called pre-commit-user. These files will be ignored by git.

##### ISSUE TYPE

 - Feature Pull Request

##### AWX VERSION

```
awx: 19.5.2.dev308+g74198d26f0.d20220318
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -6,3 +6,11 @@ if [ -z $AWX_IGNORE_BLACK ] ; then
         	exit 1)
 	fi
 fi
+
+if [ -z $AWX_IGNORE_USER ] ; then
+	if [ -d ./pre-commit-user ] ; then
+		for SCRIPT in `find ./pre-commit-user -name "*.sh" -executable` ; do
+			$SCRIPT || exit 1
+		done
+	fi
+fi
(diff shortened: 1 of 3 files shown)


AlanCoding:
Seems fine, but a _little_ bit more in the way of docs / hand-holding would be appropriate.

pre-commit.sh
```diff
@@ -6,3 +6,11 @@ if [ -z $AWX_IGNORE_BLACK ] ; then
         	exit 1)
 	fi
 fi
+
+if [ -z $AWX_IGNORE_USER ] ; then
+	if [ -d ./pre-commit-user ] ; then
+		for SCRIPT in `find ./pre-commit-user -name "*.sh" -executable` ; do
```
shanemcd: A lot of the times I don't add a `.sh` extension to my scripts. It would also be cool if we could run any executable we put in this directory.

AlanCoding (approved):

pre-commit.sh
john-westcott-iv: It will now execute any executable file that is found.

PR #11925, opened 2022-03-18
Adding awx_ as well as tower_ variable names for webhooks

john-westcott-iv:
Adding utility to ease testing webhooks from command line

Modifying all variables to use a constants list of variable names

##### SUMMARY

Modify webhook variables passed to jobs to include both the tower_ and awx_ prefixes. Added a script in tools/scripts to allow posting of webhooks from command line. Modified all variables with tower_/awx_ to pull prefixes from constants.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```
awx: 19.5.2.dev308+geeeef21029
```


tools/scripts/post_webhook.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python
+
+# You can lookup sample webhook events online.
+#    GitLab - https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html
+#    GitHub - https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads
+#
+
+# User modified variables (maybe params?)
```
AlanCoding: > (maybe params?)

old comment? Looks like they are implemented as CLI params now.

Also might be nice to reformat these example values into the CLI use.

AlanCoding (approved):
Everything looks solid.

tools/scripts/post_webhook.py
john-westcott-iv: Good catch, that comment should have been completely removed; they were my actual variables before I added click. But, I do like the idea of adding example calls so I added that to the --help. I also moved the comment about where to get more files to the help.

InnocentK (approved):
Looks good!

InnocentK:
No issues (specific to this work) found during QE testing. Should be good to merge (see approval above)

PR #11927, opened 2022-03-21
Fix subsystem metrics sync-only operation in async context

fosterseth:
## The Issue
I see this log

`django.core.exceptions.SynchronousOnlyOperation: You cannot call this from an async context - use a thread or sync_to_async.`

this is the culprit line

`self.instance_name = Instance.objects.me().hostname`

This particular instance of subsystem metrics is being called in an async context from the websocket broadcast manager.

I assume this was brought on by the django 3 upgrade

## The Fix
Just pass the hostname into init of Metrics(), so we can avoid calling this code.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 20.0.2.dev97+gc9853f0359

```

### DIFF
--- a/awx/main/analytics/subsystem_metrics.py
+++ b/awx/main/analytics/subsystem_metrics.py
@@ -136,7 +136,7 @@ class HistogramM(BaseM):
 
 
 class Metrics:
-    def __init__(self, auto_pipe_execute=True):
+    def __init__(self, auto_pipe_execute=True, instance_name=None):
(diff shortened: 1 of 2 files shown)


jbradberry:
Note: this will have to be backported or folded into the still-open PRs for the other upgrade branches

AlanCoding (approved):
This is one of those bits of code that should never have had access to the database in the first place

Issue #11765, opened 2022-02-17
Job output is not displayed fully


guliaka:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

Some lines is grayed out in job output results

![image](https://user-images.githubusercontent.com/4406405/154390078-a315359e-89ab-48a3-baa2-9093156ff1e5.png)

### AWX version

19.5.1

### Select the relevant components

- [X] UI
- [ ] API
- [ ] Docs

### Installation method

openshift

### Modifications

no

### Ansible version

2.9

### Operating system

Win10

### Web browser

Firefox, Chrome, Edge

### Steps to reproduce

Run job with many line output

![image](https://user-images.githubusercontent.com/4406405/154390318-2a0b015e-6c7d-4ffd-834e-aa8b6d8be4ba.png)

### Expected results

See all output lines

### Actual results

Many lines, specially toward end of the job just gray rectangles

![image](https://user-images.githubusercontent.com/4406405/154390430-20ed7968-e465-446b-a685-ccbdc37e1b5d.png)
  
or
  
![image](https://user-images.githubusercontent.com/4406405/154390762-cfc0fc00-6bd9-4455-8ba8-f262d6de3de9.png)

per-lind:
I can see the same thing in the job log, some times it will not scroll at all.
Just upgraded to version 20.0.0 in the hope that it would get better, but it feels like it is happening even more often now.

![awx-1](https://user-images.githubusercontent.com/37022025/154955888-1c0fae05-4b2f-4580-bf54-a559953d4b48.png)
Can see lots of request going from the browser, requests get fired nonstop even without any scrolling or interaction.
![awx-2](https://user-images.githubusercontent.com/37022025/154956025-1189b90d-3dc0-4a64-9571-8d546003c02a.png)

In the AWX pod this is from the log (URL changed to a dummy version)
[...]

PR #11928, opened 2022-03-21
Add JobJobEventsChildrenSummary endpoint

fosterseth:
##### SUMMARY

Adds a special endpoint for the purposes of optimizing the Job output page in the UI

Events are hierarchical

```

E1
- E2
- - E3
- - E4
- E5

```

- E1 has 4 total children
- E2 has 2 total children
- E2 and E5 are siblings

The UI needs to know how many children (and grandchildren, etc.) belong to a particular event. Currently the UI does a couple of API calls to get this information, _per_ nested event branch. These extraneous networks calls were leading to issues

https://github.com/ansible/awx/issues/11765#issuecomment-1066985723

This new endpoint is job specific, and returns a json object where each key is the event counter, and the value is the total number of children under that event. The idea is that the UI can get this data once, then use it as a lookup table any time it needs to know the number of children that belong to an event.

for the above event structure it would be
```json
{
1: 4,
2: 2,
}
```
Note, events with no children are not included. Also, it only returns data if the job is in a completed state.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 20.0.2.dev101+g0d3af6ec4a
```


awx/api/views/__init__.py
```diff
@@ -3824,6 +3824,82 @@ def get_queryset(self):
...
+    '''
+    Special view to facilitate processing job output in the UI.
+    In order to collapse events and their children, the UI needs to know how
+    many children exist for a given event.
+    This view returns a JSON object where the key is the event counter, and the value
+    is the number of children (and grandchildren) events.
+    Only events with children are included in the output.
+
+    e.g. Demo Job Template job
+    given:
```
AlanCoding: By "given", you just mean that this is the event data for the Demo Job Template?

awx/api/urls/job.py
```diff
@@ -27,6 +28,7 @@
     re_path(r'^(?P<pk>[0-9]+)/create_schedule/$', JobCreateSchedule.as_view(), name='job_create_schedule'),
     re_path(r'^(?P<pk>[0-9]+)/job_host_summaries/$', JobJobHostSummariesList.as_view(), name='job_job_host_summaries_list'),
     re_path(r'^(?P<pk>[0-9]+)/job_events/$', JobJobEventsList.as_view(), name='job_job_events_list'),
+    re_path(r'^(?P<pk>[0-9]+)/job_events/children_summary/$', JobJobEventsChildrenSummary.as_view(), name='job_job_event_children_summary'),
```
AlanCoding: Does the UI really need this specific to jobs, or would it be easier to apply it to all unified jobs, including project updates and ad-hoc commands? Your logic (like `job.get_event_queryset()`) could mostly work for all job types. Best to ask these questions before making the schema change.

awx/api/views/__init__.py
```diff
@@ -3824,6 +3824,82 @@ def get_queryset(self):
...
+                return
+            map_uuid_children_tally[parent] += 1
+            update_grandparents(parent)
+
+        for counter, uuid, puuid in events:
+            if not uuid:
+                continue
+            map_uuid_counter[uuid] = counter
+            if not puuid:
+                continue
```
AlanCoding, line 3911: For these cases, it would be better to filter out empty/null values when creating the original queryset. You're already at risk of performance problems, and that _could_ help.

awx/api/views/__init__.py
fosterseth: exactly

awx/api/views/__init__.py
fosterseth: good call

awx/api/views/__init__.py
```diff
@@ -3824,6 +3824,82 @@ def get_queryset(self):
...
+        # key is uuid, value is event counter (the event order)
+        map_uuid_counter = dict()
+        # key is uuid, value is parent uuid. Used as a quick lookup
+        map_uuid_puuid = dict()
+        # ultimately we want to return dict(counter: tally), not dict(uuid: tally),
+        # so we build up this dict from the map_uuid_children_tally
+        map_counter_children_tally = dict()
+
+        def update_grandparents(uuid):
+            # recursive function that updates the tally for grandparent events
```
AlanCoding: I'm not a die-hard purist against recursive methods, but in this case it looks like it would be pretty simple to re-write this as a nested loop and avoid the recursion.

awx/api/views/__init__.py
```diff
@@ -3824,6 +3824,82 @@ def get_queryset(self):
...
+        (7, '39a49992-5ca4-4b6c-b178-e56d0b0333da', 'aae0d189-e3cb-102a-9f00-000000000008')
+        (8, '504f3b28-3ea8-4f6f-bd82-60cf8e807cc0', 'aae0d189-e3cb-102a-9f00-000000000008')
+        (9, 'a242be54-ebe6-4021-afab-f2878bff2e9f', '4598d19e-93b4-4e33-a0ae-b387a7348964')
+
+    output:
+        {
+        "1": 8,
+        "2": 6,
+        "3": 2,
+        "6": 2
```
AlanCoding: Even if this method can run in a reasonable amount of time for a high-output job, the _number of keys_ is going to make it difficult to return in the response. If there are a million events, will it even be able to _transmit_ the response before the uWSGI worker dies? As a whole-job endpoint, I find it doubtful that this endpoint will survive the test of real users.

AlanCoding: Then again, I get where you're coming from excluding events with _no_ children. Assuming the multi-million or multi-billion events are because of number of hosts... maybe. How do loops figure into this? If I run one host but loop a task a million times, what fraction of the events have children?

fosterseth: yeah this is where things aren't clear me to me. The output here only includes events with children, which I am hoping is a small enough subset to reliably do this kind of thing. We may need to find a realistic, high output job to test this on.

AlanCoding: Running chatty_tasks with 1 host

```
In [6]: parents = set(event.parent_uuid for event in events)

In [7]: len(events)
Out[7]: 56

In [8]: len([event for event in events if event.uuid in parents])
Out[8]: 3
```

So seems fairly valid to cut out the bulk of data from loops.

awx/api/urls/job.py
fosterseth: @keithjgrant is this only specific to Jobs? Would it be best to have children_summary endpoint for other types of jobs?

ProjectUpdateEvent does not have a parent_uuid field

awx/api/views/__init__.py
fosterseth: changed this to a while loop

awx/api/urls/job.py
keithjgrant: Currently the UI only shows the collapsible view for playbook jobs. If we want that functionality for any other job types, I wouldn't be opposed (assuming it makes sense to), but they would also need their parent_uuid's populated, which we don't have now

awx/api/views/__init__.py
```diff
@@ -3824,6 +3824,75 @@ def get_queryset(self):
...
+class JobJobEventsChildrenSummary(APIView):
+    renderer_classes = [JSONRenderer]
+    '''
+    Special view to facilitate processing job output in the UI.
+    In order to collapse events and their children, the UI needs to know how
+    many children exist for a given event.
+    The UI also needs to know the order of the event (0 based index), which
+    usually matches the counter, but not always.
+    This view returns a JSON object where the key is the event counter, and the value
+    includes the number of children (and grandchildren) events.
```
AlanCoding: value includes something else now

awx/api/views/__init__.py
```diff
@@ -3824,6 +3824,75 @@ def get_queryset(self):
...
+        (9, 'a242be54-ebe6-4021-afab-f2878bff2e9f', '4598d19e-93b4-4e33-a0ae-b387a7348964')
+
+    output:
+        {
+        "1": [0, 8],
+        "2": [1, 6],
+        "3": [2, 2],
+        "6": [5, 2],
+        }
+    counter 1 is event 0, and has 8 children
```
AlanCoding: what does "is event 0" mean? that 0 is the primary key?

awx/api/views/__init__.py
```diff
@@ -3826,6 +3826,174 @@ def get_queryset(self):
         return job.get_event_queryset().select_related('host').order_by('start_line')
 
 
+class JobJobEventsChildrenSummary(APIView):
+    '''
```
fosterseth: if people feel this doc string is too long, we can add this as an entry in docs/

awx/api/views/__init__.py
```diff
@@ -3826,6 +3826,168 @@ def get_queryset(self):
...
+    and the value is the assigned nested uuid.
+    '''
+
+    renderer_classes = [JSONRenderer]
+    meta_events = ('debug', 'verbose', 'warning', 'error', 'system_warning', 'deprecated')
+
+    def get(self, request, **kwargs):
+        job = get_object_or_404(models.Job, pk=kwargs['pk'])
+        if not job.event_processing_finished:
+            return Response({})
```
AlanCoding: If this won't run when event processing isn't finished, then why do we need to bother with the rowNumber? It should be the same as counter minus 1.

keithjgrant: @AlanCoding does that mean there are never gaps in the counter values once processing is done? If that's the case, we can sort that out in the UI.

It also means I could probably look for ways to simplify the UI logic, as it does extra work to handle cases where there's a counter gap. Can there still be a gap while the job is running/processing?

AlanCoding: There can't reasonably be a gap when `job.event_processing_finished` is True. Consider the logic from that.

https://github.com/ansible/awx/blob/6c1adade250020fcf900ecfc6d33e6f8274c8e0a/awx/main/models/unified_jobs.py#L1022

The way this works:
 - Control process receives output from receptor, and it does minimal processing of this and submits the data to receptor
   - This process counts the number of events that it sends, and saves that to `emitted_events`
 - Callback receiver pulls the data off the redis queue and then saves it to database

If `event_processing_finished` never transitions to `True`, then that's an API bug. Doesn't matter what type of job or what it does. This is accounting - the callback receiver should always save the same number of events that the control process submitted.

If the job is actually _running_ then this endpoint returns `{}` anyway

fosterseth: there are 3 numbers to consider here

- event_qs.count()
- emitted_events
- the `counter` field on the event itself (from ansible-runner)

the first 2 are pretty much guaranteed to match if event_processing_finished==true, as Alan outlined above.

What about `counter` field on the event itself? if there are 100 events, are we guaranteed that we have events with counters 1 through 100? Looking at code I don't see why not. I'm wondering if the previous reports of missing counters was because of not checking whether events had been fully processed or not.

We could make the decision to rely on `counter`, simplifying this endpoint a bit. We could just throw an error or warning if this code finds some missing counters.

AlanCoding: Right after this line, I think you should go ahead and write a conditional to give a 400 with too many events. You don't need to load anything, use `job.emitted_events`, already in memory. If that number is greater than your limit, then bail.

How do you know what the limit is? Use the firehose script or whatever you want to get the time taken to load some events. Divide to get seconds per event, and then use that to find out how many events will take 10 seconds to load and make that the limit. This is simple enough that it would be best to do that now and avoid more costly iteration in the future.

rebeccahhh (approved):
I like these updates a lot!

PR #11933, opened 2022-03-22
Add setting for notification job status retry loop

AlanCoding:
This change has already been made downstream, and this is adopting it for `devel`.


apollo13:
@AlanCoding Any advice on how this setting should be chosen? What is the actual underlying issue?

AlanCoding:
This is not attempting to address the underlying issue, this is to maintain consistency across versions so that this setting will continue to be respected if someone set it in the past.

When the underlying issue is addressed, it is expected that this setting will be removed because it will no longer be needed.

shanemcd (approved):

PR #11939, opened 2022-03-22
Fix search toolbar clear all filters

### LINKED ISSUE #8474: Clear All Filters in Search should disappear

marshmalien:
### SUMMARY

Issue: https://github.com/ansible/awx/issues/8474

Dynamically removing the ToolbarFilter component threw the Toolbar component’s inner `filterInfo` state out of sync. The filters were responsible for notifying the Toolbar’s context of changes, so by removing them prematurely, the Toolbar was unaware that it needed to remove filters from its state. This PR persists the ToolbarFilter components, which allows the Toolbar to manage showing and hiding the filter list. 

filter

### ISSUE TYPE

- Bugfix Pull Request

### COMPONENT NAME

- UI

### DIFF
--- a/awx/ui/src/components/Search/Search.js
+++ b/awx/ui/src/components/Search/Search.js
@@ -1,5 +1,5 @@
 import 'styled-components/macro';
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 
 import { t } from '@lingui/macro';
@@ -65,6 +65,26 @@ function Search({
   const [searchValue, setSearchValue] = useState('');
   const [isFilterDropdownOpen, setIsFilterDropdownOpen] = useState(false);
 
+  const params = parseQueryString(qsConfig, location.search);
+  if (params?.host_filter) {
+    params.ansible_facts = params.host_filter.substring(
+      'ansible_facts__'.length
+    );
+    delete params.host_filter;
+  }
+
+  const searchChips = getChipsByKey(params, columns, qsConfig);
+  const [chipsByKey, setChipsByKey] = useState(
+    JSON.parse(JSON.stringify(searchChips))
+  );
+
+  useEffect(() => {
+    Object.keys(chipsByKey).forEach((el) => {
+      chipsByKey[el].chips = [];
+    });
+    setChipsByKey({ ...chipsByKey, ...searchChips });
+  }, [location.search]); // eslint-disable-line react-hooks/exhaustive-deps
+
   const handleDropdownSelect = ({ target }) => {
     const { key: actualSearchKey } = columns.find(
       ({ name }) => name === target.innerText
@@ -97,15 +117,6 @@ function Search({
       onRemove(key, actualValue);
     }
   };
-
-  const params = parseQueryString(qsConfig, location.search);
-  if (params?.host_filter) {
-    params.ansible_facts = params.host_filter.substring(
-      'ansible_facts__'.length
-    );
-    delete params.host_filter;
-  }
-  const chipsByKey = getChipsByKey(params, columns, qsConfig);
 
   const { name: searchColumnName } = columns.find(
     ({ key }) => key === searchKey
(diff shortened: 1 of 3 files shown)


awx/ui/src/components/Search/Search.js
```diff
@@ -65,6 +65,26 @@ function Search({
...
+    params.ansible_facts = params.host_filter.substring(
+      'ansible_facts__'.length
+    );
+    delete params.host_filter;
+  }
+
+  const searchChips = getChipsByKey(params, columns, qsConfig);
+  const [chipsByKey, setChipsByKey] = useState(
+    JSON.parse(JSON.stringify(searchChips))
+  );
```
AlexSCorey, line 79:
```suggestion
  const [chipsByKey, setChipsByKey] = useState(
getChipsByKey(params, columns, qsConfig)
  );
```

awx/ui/src/components/Search/Search.js
```diff
@@ -65,6 +65,26 @@ function Search({
...
+
+  useEffect(() => {
+    Object.keys(chipsByKey).forEach((el) => {
+      chipsByKey[el].chips = [];
+    });
+    setChipsByKey({ ...chipsByKey, ...searchChips });
+  }, [location.search, searchChips.length]); // eslint-disable-line react-hooks/exhaustive-deps
```
AlexSCorey:
```suggestion
  useEffect(() => {
    Object.keys(chipsByKey).forEach((el) => {
      chipsByKey[el].chips = [];
    });
    setChipsByKey(getChipsByKey(params, columns, qsConfig));
  }, [location.search]);
```

searchChips is an object so it looks like searchChips.length is always going to be undefined.  if we can remove that dependency from the array then I think we can also remove the commented `eslint-disable-line...`

Finally, since we want to keep chipsByKey in sync with searchChips lets just use the getChipsByKey function to give us those values here.

AlexSCorey (changes requested):
I did a cursory look at this and found some things that we could improve on.  If there are cases where my requested changes wouldn't work, like were searchChips is an array and not an object, then I'm good with the code as is.

awx/ui/src/components/Search/Search.js
marshmalien: I agree that `searchChips.length` is unnecessary here and should be removed from the dependency array since it is a js object. 

```
const [chipsByKey, setChipsByKey] = useState(
    JSON.parse(JSON.stringify(searchChips))
);
 ````
Initializing the `chipsByKey` state (above) is key to removing any object reference to `searchChips` and to persisting the original filters. The goal of this PR was to build up the list of filters and maintain them even after the params change. When the user deletes a filter, instead of removing it from the `chipsByKey` object, we pass that specific filter's chips property an empty array. Keeping the original filter key around for PatternFly's Toolbar component was the trick to fixing the bug. 

```
setChipsByKey(getChipsByKey(params, columns, qsConfig));
```
The line above would overwrite the list of filters and bring back the original bug. 

```
setChipsByKey({ ...chipsByKey, ...searchChips });
```
By spreading the original `chipsByKey` and the current `searchChips` into the setter, I'm able to update the state with new search filters from `searchChips` while preserving the older search filter keys.

AlexSCorey (approved):
We talked offline.  We are going to make patternfly aware of their toolbar dynamic filtering of the toolbar.  This code works well for us.  Good work!

akus062381 (approved):

Issue #11818, opened 2022-02-28
Most of output is greyed out since upgrading to 20.0.0


nicolaibaralmueller:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

Output not showing full result since upgrading from 19.4 to 20.0. Browser windows also becomes unresponsive sometimes. 

### AWX version

20.0.0

### Select the relevant components

- [X] UI
- [ ] API
- [ ] Docs

### Installation method

minikube

### Modifications

no

### Ansible version

quay.io/ansible/awx-ee:0.6.0

### Operating system

Centos

### Web browser

Chrome, Edge

### Steps to reproduce

I don't know

### Expected results

Output scrolling should show all output

### Actual results

Most of the lines grayed out

stanislav-zaprudskiy:
Is this a duplicate of https://github.com/ansible/awx/issues/11765?

nicolaibaralmueller:
> Is this a duplicate of #11765?

Somewhat, yes. But in my output is completely greyed out after a certain point.

radek-sprta:
I have the same issue. But I am actually on 19.5.1.

MrBones757:
Its actually a duplicate of this :) https://github.com/ansible/awx/issues/11647
We had the same symptoms and did some digging and this is what we came up with.

kimbernator:
Seeing the same issue on version 20

nicolaibaralmueller:
> Its actually a duplicate of this :) #11647 We had the same symptoms and did some digging and this is what we came up with.

Thanks. This makes sense. We are hosting AWX in Azure and the VM (4vCores/16GiB) crashes periodically when many jobs are executed at the same time. Might need to downgrade to 19.4 again if this is not resolved asap.

chris93111:
hi same issue on 20.0.0 

<img width="1172" alt="image" src="https://user-images.githubusercontent.com/27526373/157115614-685102db-e37e-4de5-8660-61d53e4928b3.png">

download output is complete is only on UI

with ?job_output.not__stdout__startswith=[WARNING] all output is appear

[...]

Issue #11647, opened 2022-01-31
Upgrade AWX 19.2.2 -> 19.5.1 - UI Job Output Causing Postgres DB to hit 100% CPU - AWX Becomes Unresponsive


MrBones757:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

Following the upgrade to AWX 19.5.1 from 19.2.2, streaming job logs or scrolling though job logs either while the job is running or after the job has finished causes Postgres DB CPU Usage to reach 100% on 4 CPU cores.
Streaming job output via API (cli) does not yield the same issue.
The issue only seems to occur when the job output is longer than what fits on the initial output window size.

REVERTING TO 19.5.0 fixes the issue - the issue is with 19.5.1

### AWX version

19.5.1

### Installation method

kubernetes

### Modifications

yes

### Ansible version

EEs W/ 2.9.27 & 2.11.7

### Operating system

Ubuntu as k8s Hosts

### Web browser

Chrome

### Steps to reproduce

1) Upgrade an instance from 19.2.2 - 19.5.1 (optional?)
2) Run a job where the job output exceeds one loaded page in the UI
3) observe results not load (taking upto minutes to render)
    observe postgres node cpu usage spiking extremely high while the load is occuring, coming from postgres processes (see screenshot)

### Expected results

Fast STDout rendering, with low CPU overhead as in the API and previous versions.

### Actual results

Extreme CPU usage on DB node.
Many select processes showing when using top / htop on the physical host that has the postgres DB on it.
![CPU Screenshot](https://user-images.githubusercontent.com/22608473/151782708-d2226ced-4086-47f4-a3ea-67c34dc152b8.png)
We were also able to observe a large number of uncaught promises when scrolling in the job output UI, and the operation its self is causing a 500 internal server error, which is causing the uncaught promise error.
![DevTools](https://user-images.githubusercontent.com/22608473/151785681-e395eddf-f666-45ca-86eb-141a0f869167.png)

Manually navigating to one of the affected URLS:
https://awx.my.cool.company/api/v2/job_events/15472448/children/?counter__gt=4&order_by=counter&page_size=1
[...]

PR #11944, opened 2022-03-23
Use new children-summary endpoint data to traverse job event tree

keithjgrant:
##### SUMMARY
Updates UI to use the new children-summary endpoint (#11928) to determine number of child events & position in the tree. This greatly reduces the number of network requests made by the job output page for playbook jobs while expand/collapse is enabled.

This also addresses certain job events (such as Warnings) that do not fit into the parent/child tree structure of job events the way normal events do, which was breaking the UI's event lookup algorithm. The problem is these events don't have an `parent_uuid` field even though they might occur between two other events that share a parent. The children-summary endpoint provides a "meta_event_nested_uuid" object to place these events under the correct parent node so the lookup algorithm can work as expected.

~~This depends on #11928 and should not be merged until that PR is also merged.~~

Addresses #11818, #11765, #11647 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui/src/screens/Job/JobOutput/JobOutput.test.js
+++ b/awx/ui/src/screens/Job/JobOutput/JobOutput.test.js
@@ -1,4 +1,3 @@
-/* eslint-disable max-len */
 import React from 'react';
 import { act } from 'react-dom/test-utils';
 import { JobsAPI, JobEventsAPI } from 'api';
@@ -26,14 +25,9 @@ const applyJobEventMock = (mockJobEvents) => {
     };
   };
   JobsAPI.readEvents = jest.fn().mockImplementation(mockReadEvents);
-  JobEventsAPI.readChildren = jest.fn().mockResolvedValue({
+  JobsAPI.readChildrenSummary = jest.fn().mockResolvedValue({
     data: {
-      results: [
-        {
-          counter: 20,
-          uuid: 'abc-020',
-        },
-      ],
+      1: [0, 100],
     },
   });
 };
--- a/awx/ui/src/screens/Job/JobOutput/useJobEvents.js
+++ b/awx/ui/src/screens/Job/JobOutput/useJobEvents.js
@@ -11,16 +11,18 @@ const initialState = {
   // events with parent events that aren't yet loaded.
   // arrays indexed by parent uuid
   eventsWithoutParents: {},
+  // tuple arrays in the form [rowNumber, numChildren] for parent nodes, indexed by counter
+  childrenSummary: {},
   isAllCollapsed: false,
 };
 export const ADD_EVENTS = 'ADD_EVENTS';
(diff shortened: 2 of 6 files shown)


nixocio (approved):
hey @keithjgrant, I got this code locally. I did several tests. 

I was not able to see the errors reported on the issues. Excellent job.

Also there was a big improvement on the number of requests being made.

Number of requests for job output using awx/devel.

image

Number of requests for job output using code.

image

awx/ui/src/screens/Job/JobOutput/JobOutput.test.js
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable max-len */
 import React from 'react';
 import { act } from 'react-dom/test-utils';
 import { JobsAPI, JobEventsAPI } from 'api';
```
marshmalien, line 3: We could remove the JobEventsAPI import from this test since we're no longer using it. What are your thoughts on removing the JobEventsAPI from our models completely? It doesn't look like we use that model now that we have this new `/jobs/:id/job_events/children_summary/` endpoint.

nixocio:
I triggered the e2e tests.

awx/ui/src/screens/Job/JobOutput/useJobEvents.js
```diff
@@ -42,6 +47,31 @@ export default function useJobEvents(callbacks, isFlatMode) {
     });
   }, [actionQueue]);
 
+  useEffect(() => {
+    if (isFlatMode) {
+      return;
+    }
+
+    callbacks
+      .fetchChildrenSummary()
```
marshmalien, line 56: Do we want to avoid calling this whenever the `job.type` is not `job`? I noticed that the API returns a 404 response from `/jobs/:id/job_events/children_summary` whenever it is a `project_update` or `system_job` job.

Screen Shot 2022-04-05 at 10 56 53 AM

keithjgrant: good idea. I chatted with @akus062381 and we'll look to clean this up in a follow-up PR to avoid changes this late in testing. I can delete the unused JobEventsAPI at the same time

keithjgrant: #12011 filed

InnocentK (approved):
Looks good! The above failed tests are unrelated to this work so it should be fine to merge

ryanmerolle:
Thanks for the fix!  Do we know when a new release with this PR will be cut?

AlexSCorey:
The release cadence of AWX varies, however we try to release a new version of AWX every few weeks.  If you feel comfortable running the HEAD of awx/devel then these changes will be immediately available.

Issue #11811, opened 2022-02-25
main_projectupdateevent already exists


chris93111:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

Hello

With many job execution at same time, some job crash with error django.db.utils.ProgrammingError: relation "main_projectupdateevent_xxxxxxx" already exists

awx 20.0.0
rhel 8.4

4 instance of awx in same kubernetes cluster

### AWX version

20.0.0

### Select the relevant components

- [X] UI
- [X] API
- [ ] Docs

### Installation method

kubernetes

### Modifications

no

### Ansible version

2.9

### Operating system

redhat 8.4

### Web browser

Chrome

### Steps to reproduce

launch many same job with allow override branch scm 

### Expected results

launch without error

### Actual results

some job crash with django.db.utils.ProgrammingError: relation "main_projectupdateevent_xxxxxxx" already exists

### Additional information

worked in 15.0.0

nixocio:
hey @chris93111, could you add the full error traceback?

chris93111:
Hi @nixocio , they are no trace back only error 
<img width="1148" alt="image" src="https://user-images.githubusercontent.com/27526373/155787447-77776069-1ab7-40c6-af16-b6ad99013121.png">

chris93111:
@nixocio found un result job
```
Traceback (most recent call last):
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/db/backends/utils.py", line 82, in _execute
    return self.cursor.execute(sql)
psycopg2.errors.DuplicateTable: relation "main_projectupdateevent_20220226_12" already exists


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/jobs.py", line 449, in run
    self.pre_run_hook(self.instance, private_data_dir)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/jobs.py", line 946, in pre_run_hook
    create_partition(local_project_sync.event_class._meta.db_table, start=local_project_sync.created)
[...]

PR #11945, opened 2022-03-23
Handle error for create_partition

fosterseth:
##### SUMMARY

https://github.com/ansible/awx/issues/11811

`create_partition` should already skip creating a table if it already exists.

```python3
    cursor.execute(
        f'CREATE TABLE IF NOT EXISTS {tblname}_{partition_label} '
        f'PARTITION OF {tblname} '
        f'FOR VALUES FROM (\'{start_timestamp}\') to (\'{end_timestamp}\');'
    )
```
However, occasionally parallel operations collide and we get an error.

I can recreate the scenario with

```python3
from awx.main.utils.common import create_partition
from django.utils import timezone
from multiprocessing import Pool

def x(i):
    create_partition('main_projectupdateevent', start=timezone.now())

with Pool(30) as p:
    print(p.map(x, range(30)))
```
 

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 20.0.2.dev121+g26c183b870

```

### DIFF
--- a/awx/main/utils/common.py
+++ b/awx/main/utils/common.py
@@ -21,7 +21,7 @@ from django.core.exceptions import ObjectDoesNotExist, FieldDoesNotExist
 from django.utils.dateparse import parse_datetime
 from django.utils.translation import gettext_lazy as _
 from django.utils.functional import cached_property
-from django.db import connection
+from django.db import connection, transaction, ProgrammingError
 from django.db.models.fields.related import ForeignObjectRel, ManyToManyField
 from django.db.models.fields.related_descriptors import ForwardManyToOneDescriptor, ManyToManyDescriptor
 from django.db.models.query import QuerySet
@@ -1141,12 +1141,16 @@ def create_partition(tblname, start=None, end=None, partition_label=None, minute
         else:
             partition_label = start.strftime('%Y%m%d_%H')
 
-    with connection.cursor() as cursor:
-        cursor.execute(
-            f'CREATE TABLE IF NOT EXISTS {tblname}_{partition_label} '
-            f'PARTITION OF {tblname} '
-            f'FOR VALUES FROM (\'{start_timestamp}\') to (\'{end_timestamp}\');'
-        )
+    try:
+        with transaction.atomic():
+            with connection.cursor() as cursor:
+                cursor.execute(
+                    f'CREATE TABLE IF NOT EXISTS {tblname}_{partition_label} '
+                    f'PARTITION OF {tblname} '
+                    f'FOR VALUES FROM (\'{start_timestamp}\') to (\'{end_timestamp}\');'
+                )
+    except ProgrammingError as e:
+        logger.debug(f'Caught known error due to existing partition: {e}')
 
 
 def cleanup_new_process(func):


AlanCoding (changes requested):
There's 1 more call to this method in `tasks/jobs.py`, and even though it's already in a try-except, that does not accomplish the intention of this. Following the current pattern, I think you should add another _nested_ try-except in that case, so that an error here does not prevent the inventory update from running.

Or you could put this in the `create_partition` method itself. Up to you.

jbradberry (changes requested):
I think I have to throw a red flag here.  A failed write to the database here will likely cause further attempts to write to the database in completely unrelated code to also fail.  I strongly suspect that we need transactions and rollbacks to do this right.

AlanCoding:
transactions have their own way of making error handing worse, but Seth's script looks like a good test for the solution, whatever that may be.

awx/main/utils/common.py
```diff
@@ -1141,12 +1141,16 @@ def create_partition(tblname, start=None, end=None, partition_label=None, minute
...
+    try:
+        with transaction.atomic():
+            with connection.cursor() as cursor:
+                cursor.execute(
+                    f'CREATE TABLE IF NOT EXISTS {tblname}_{partition_label} '
+                    f'PARTITION OF {tblname} '
+                    f'FOR VALUES FROM (\'{start_timestamp}\') to (\'{end_timestamp}\');'
+                )
+    except Exception as e:
+        logger.warning(e)
```
fosterseth: not sure of the best log level here -- exception and error is going to show a bunch of scary red text, and we probably don't want users to panic or open issues if they see it.

I think warning is okay-ish, but I could be convinced of debug or even just passing with no logging.

fosterseth:
with django's autocommit mode, we might not even need a `with transactions.atomic()` wrapper around the cursor.execute() call.

```
try:
    cursor.execute(..)
except:
    pass
```

this would work under normal conditions. If cursor.execute hits an error, it will be rolled back automatically and we can carry on with subsequent queries.

However, if we are already in a transaction block by the time we hit this code (i.e. a `with transaction.atomic():` higher in the call stack), then this try / except block isn't enough. Once we hit the error, nothing is rolled back, and then try to run more queries in the same thread. This would lead to the "An error occurred in the current transaction. You can't execute queries until the end of the 'atomic' block." error.

So for that, wrapping the transaction.atomic() is safer, or we could consider adding the `durable=True` flag to ensure it's the outer transaction block.

#### Demonstration

To demonstrate that we can make queries right after the errors,

```python3
from awx.main.utils.common import create_partition
from django.utils import timezone
from multiprocessing import Pool

def x(i):
  try:
    create_partition('main_projectupdateevent', start=timezone.now())
  except:
    pass
  return Job.objects.count()

def y(i):
  with transaction.atomic():
    return x(i)

with Pool(30) as p:
  print(p.map(y, range(30)))

```

Note: the outer transaction.atomic() is necessary to demonstrate the error. On devel, calling create_partition() will run the cursor.execute() in its own transaction (and thus will auto rollback), so we would never see the "ignored until end of transaction block" error.

##### output on current devel branch:

`InternalError: current transaction is aborted, commands ignored until end of transaction block`

##### on this branch:

```
2022-03-28 17:14:29,236 WARNING  [-] awx.main.utils relation "main_projectupdateevent_20220328_17" already exists
..
2022-03-28 17:14:29,296 WARNING  [-] awx.main.utils relation "main_projectupdateevent_20220328_17" already exists
2022-03-28 17:14:29,294 WARNING  [-] awx.main.utils relation "main_projectupdateevent_20220328_17" already exists
2022-03-28 17:14:29,300 WARNING  [-] awx.main.utils relation "main_projectupdateevent_20220328_17" already exists
[51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51, 51]
```

51 is the output from `Job.objects.count()`

awx/main/utils/common.py
jbradberry: Since the situation is not actually very exceptional, I'd go for no higher than `info`, possibly `debug`.

AlanCoding: `info` or `debug` would be just fine if you were sure that you were catching the error you think you're catching. With a blanket `except Exception`, something could have gone wrong that you didn't expect. If you could narrow the error type, then a low-level log would be appropriate. If you can't do that, then you should use `logger.exception`

jbradberry (approved):

awx/main/utils/common.py
fosterseth: The "already exists" error is a ProgrammingError, which seems narrow enough to catch on

https://peps.python.org/pep-0249/#programmingerror

```
Exception raised for programming errors, e.g. table not found or already exists, syntax error in the SQL statement, wrong number of parameters specified, etc. It must be a subclass of [DatabaseError](https://peps.python.org/pep-0249/#databaseerror).
```

I'm thinking we should just catch the ProgrammingError and let all other errors raise an exception.

awx/main/utils/common.py
```diff
@@ -1141,12 +1141,16 @@ def create_partition(tblname, start=None, end=None, partition_label=None, minute
...
+    try:
+        with transaction.atomic():
+            with connection.cursor() as cursor:
+                cursor.execute(
+                    f'CREATE TABLE IF NOT EXISTS {tblname}_{partition_label} '
+                    f'PARTITION OF {tblname} '
+                    f'FOR VALUES FROM (\'{start_timestamp}\') to (\'{end_timestamp}\');'
+                )
+    except ProgrammingError as e:
+        logger.debug(e)
```
AlanCoding: A little more context would be appreciated, because you don't get line numbers or anything. If a log statement looks absurdly redundant, you're doing it right. Like `f'Caught known error due to existing partition: {e}'`

AlanCoding (approved):
great research on this

rebeccahhh (approved):
Loving the research and changes. +1

fosterseth:
thanks team, going to make sure tests pass, then I'll merge it in :)

Issue #11930, opened 2022-03-21
Credential Search Filter- Request Ability to Filter by Type


akus062381:
On the list of Credentials, the Type name is shown in the list next to each credential. However, a user cannot filter and just see Amazon Web Services, or Insights, or HashiCorp Vault Secret Lookup, for example. I think that this would be a helpful addition from a user experience perspective.

![image](https://user-images.githubusercontent.com/63685497/159350649-effa369b-ad68-43ff-a3f6-4b8c794afcf3.png)

nixocio:
The API provides `"credential_type__search"` on the credentials endpoint.

PR #11946, opened 2022-03-23
Add credential type filter

nixocio:
Add credential type filter

image

See: https://github.com/ansible/awx/issues/11930


akus062381 (approved):
awesome addition, thanks!

AlexSCorey (approved):
Looks good!

PR #11950, opened 2022-03-24
collection: update bindep.txt py39 and el9 support

Spredzy:
Updating bindep.txt to add support for py39 runtime (for el8) and add suppor for el9.


AlanCoding (approved):
:ship:

PR #4829, opened 2019-09-27
Fix error with rejoining node to cluster after lost connection to postgres

ryanpetrello:
see: https://github.com/ansible/awx/issues/4294
context: https://github.com/ansible/awx/issues/4294#issuecomment-535912978

PR #11955, opened 2022-03-24
Increase resiliency when application crashes

shanemcd:
##### SUMMARY
Prior to this change, the control plane pod would remain running if a service crashed.

The commits should be self-explanatory, but at a high-level, this PR:

- Adds a new event listener that causes the pod to exit if any process crashes.
- Fixes a race condition where the callback receiver would crash if code in the dispatcher didnt run first.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

### DIFF
--- a/tools/ansible/roles/dockerfile/files/launch_awx.sh
+++ b/tools/ansible/roles/dockerfile/files/launch_awx.sh
@@ -24,4 +24,4 @@ cat << EOF > /var/lib/awx/rsyslog/rsyslog.conf
 action(type="omfile" file="/dev/null")
 EOF
 
-supervisord -c /etc/supervisord.conf
+exec supervisord -c /etc/supervisord.conf
--- a/tools/ansible/roles/dockerfile/files/launch_awx.sh
+++ b/tools/ansible/roles/dockerfile/files/launch_awx.sh
@@ -17,4 +17,11 @@ set -e
 
 wait-for-migrations
 
+# This file will be re-written when the dispatcher calls reconfigure_rsyslog(),
+# but it needs to exist when supervisor initially starts rsyslog to prevent the
+# container from crashing. This was the most minimal config I could get working.
+cat << EOF > /var/lib/awx/rsyslog/rsyslog.conf
+action(type="omfile" file="/dev/null")
+EOF
+
 supervisord -c /etc/supervisord.conf
--- a/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
+++ b/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
@@ -120,11 +120,7 @@ RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master
     chmod 700 get_helm.sh && \
     ./get_helm.sh
 
-# Install tini
-RUN curl -L -o /usr/bin/tini https://github.com/krallin/tini/releases/download/v0.19.0/tini-{{ tini_architecture | default('amd64') }} && \
-    chmod +x /usr/bin/tini
-
-RUN pip3 install virtualenv supervisor
+RUN pip3 install virtualenv supervisor dumb-init
 
 RUN rm -rf /root/.cache && rm -rf /tmp/*
 
@@ -293,7 +289,7 @@ CMD ["/bin/bash"]
 USER 1000
 EXPOSE 8052
 
-ENTRYPOINT ["/usr/bin/tini", "--"]
+ENTRYPOINT ["dumb-init", "--"]
 CMD /usr/bin/launch_awx.sh
 VOLUME /var/lib/nginx
 VOLUME /var/lib/awx/.local/share/containers
--- a/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
+++ b/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
@@ -194,6 +194,7 @@ RUN mkdir -p /etc/containers/registries.conf.d/ && echo "unqualified-search-regi
 # Create default awx rsyslog config
 ADD tools/ansible/roles/dockerfile/files/rsyslog.conf /var/lib/awx/rsyslog/rsyslog.conf
 ADD tools/ansible/roles/dockerfile/files/wait-for-migrations /usr/local/bin/wait-for-migrations
+ADD tools/ansible/roles/dockerfile/files/stop-supervisor /usr/local/bin/stop-supervisor
 
 ## File mappings
 {% if build_dev|bool %}
@@ -264,9 +265,9 @@ RUN for dir in \
     for file in \
       /etc/containers/containers.conf \
       /var/lib/awx/.config/containers/containers.conf \
-      /var/lib/shared/overlay-images/images.lock \ 
+      /var/lib/shared/overlay-images/images.lock \
       /var/lib/shared/overlay-layers/layers.lock \
-      /var/lib/shared/vfs-images/images.lock \ 
+      /var/lib/shared/vfs-images/images.lock \
       /var/lib/shared/vfs-layers/layers.lock \
       /var/run/nginx.pid \
       /var/lib/awx/venv/awx/lib/python3.9/site-packages/awx.egg-link ; \
--- a/tools/ansible/roles/dockerfile/templates/supervisor.conf.j2
+++ b/tools/ansible/roles/dockerfile/templates/supervisor.conf.j2
@@ -96,18 +96,13 @@ stdout_logfile_maxbytes=0
 programs=nginx,uwsgi,daphne,wsbroadcast,awx-rsyslogd
 priority=5
 
-# TODO: Exit Handler
-
-{% if kube_dev | bool %}
-[eventlistener:awx-config-watcher]
-command=/usr/bin/config-watcher
+[eventlistener:superwatcher]
+command=stop-supervisor
+events=PROCESS_STATE_FATAL,PROCESS_STATE_EXITED
 stderr_logfile=/dev/stdout
 stderr_logfile_maxbytes=0
 stdout_logfile=/dev/stdout
 stdout_logfile_maxbytes=0
-events=TICK_60
-priority=0
-{% endif %}
 
 [unix_http_server]
 file=/var/run/supervisor/supervisor.web.sock
--- a/tools/ansible/roles/dockerfile/templates/supervisor.conf.j2
+++ b/tools/ansible/roles/dockerfile/templates/supervisor.conf.j2
@@ -12,9 +12,10 @@ directory = /awx_devel
 {% else %}
 command = nginx -g "daemon off;"
 {% endif %}
-autostart = true
 autorestart = true
-stopwaitsecs = 5
+startsecs = 30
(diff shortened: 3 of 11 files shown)


tools/ansible/roles/dockerfile/templates/Dockerfile.j2
```diff
@@ -292,7 +289,7 @@ CMD ["/bin/bash"]
 USER 1000
 EXPOSE 8052
 
-ENTRYPOINT ["/usr/bin/tini", "--"]
+ENTRYPOINT ["dumb-init", "--"]
```
rooftopcellist, line 292: We should use the full path here: `/usr/bin/dumb-init`

rooftopcellist (approved):
Other than this [small nitpick](https://github.com/ansible/awx/pull/11955/files#r836679890), LGTM.

tools/ansible/roles/dockerfile/files/launch_awx.sh
```diff
@@ -17,4 +17,13 @@ set -e
 
 wait-for-migrations
 
-supervisord -c /etc/supervisord.conf
+awx-manage provision_instance
```
AlanCoding: I want @jbradberry eyes on this

tools/ansible/roles/dockerfile/files/launch_awx.sh
```diff
@@ -17,4 +17,13 @@ set -e
...
+
+# This file will be re-written when the dispatcher calls reconfigure_rsyslog(),
```
AlanCoding, line 20: Can you paste the error from that?

tools/ansible/roles/dockerfile/templates/Dockerfile.j2
shanemcd: @rooftopcellist Why tho?

tools/ansible/roles/dockerfile/files/launch_awx.sh
jbradberry: At a first glance I think I like it.

jbradberry: I never liked `--hostname` being a mandatory "optional" flag anyway.

tools/ansible/roles/dockerfile/files/launch_awx.sh
shanemcd: Pretty straight forward to test it out:

```
bash-5.1$ rsyslogd -n -f /var/lib/awx/rsyslog/rsyslog-doesnt-exist-yet.conf
rsyslogd: could not open config file '/var/lib/awx/rsyslog/rsyslog-doesnt-exist.conf': No such file or directory [v8.2102.0-101.el9 try https://www.rsyslog.com/e/2104 ]
rsyslogd: run failed with error -2104 (see rsyslog.h or try https://www.rsyslog.com/e/2104 to learn what that number means)
```

tools/ansible/roles/dockerfile/templates/Dockerfile.j2
rooftopcellist: I just noticed we had previously done it with tini and @pabelanger   did it with dumb-init in the Dockerfile for python-base (used for the EE containers).
I was thinking it was best practice because installing it a different way (like via pip instead of the rpm) might cause it to favor the wrong path. 
 
But in this case, we control how it is installed, and will only be installing with pip upstream and rpm downstream... So it's a non issue.

tools/ansible/roles/dockerfile/files/launch_awx.sh
AlanCoding: The other thing I'm chewing on is this being in an entry-point kind of script, but I now realize this pattern is more similar to the development environment

https://github.com/ansible/awx/blob/devel/tools/docker-compose/bootstrap_development.sh

I still feel like something's suspiciously missing. Isn't this a function of the `AWX_AUTO_DEPROVISION_INSTANCES` setting? If we register in this script, what are we doing _here_?

https://github.com/ansible/awx/blob/ac6a82eee41feb041ff3e4d16459d4b1a774175f/awx/main/managers.py#L172-L173

Already, that raises another point that if we do register here, we should register as a control node. Also with the correct uuid.

Also, what's the reason for moving it here to begin with?

AlanCoding: After this change, instances can still get registered here, and I'm unsure if it should be removed (but lean toward removing)

https://github.com/ansible/awx/blob/f5c176701b0551e179bb4ef663b57fd3a6ba6f0a/awx/main/dispatch/reaper.py#L38

AlanCoding: What's the behavior if you run `awx-manage provision_instance` on a non-K8S environment? Will it give a sensible error?

tools/ansible/roles/dockerfile/files/launch_awx.sh
rooftopcellist: @shanemcd if we are now creating a dummy file (that will be overwritten) in the launch script, then we may not need to copy in a dummy file in the Dockerfile anymore - https://github.com/ansible/awx/blob/1de9dddd21b0bd7144137f8a8a2956f4f097bccd/tools/ansible/roles/dockerfile/templates/Dockerfile.j2#L195

shanemcd: @rooftopcellist I think that's still used by the development environment... But I can do the same thing in `tools/docker-compose/launch_awx.sh` as I did here.

tools/ansible/roles/dockerfile/files/launch_awx.sh
shanemcd: > After this change, instances can still get registered here, and I'm unsure if it should be removed (but lean toward removing)

I lean towards keeping it. If for some reason the instance gets deprovisioned while the pod is running, the system would not recover. I am worried about reintroducing the bug that was getting solved by https://github.com/ansible/awx/pull/4829. I dont think it's going hurt anything by keeping it.

shanemcd: > What's the behavior if you run awx-manage provision_instance on a non-K8S environment? Will it give a sensible error?

I dont see a reason to throw an error. But I have updated the help text to reflect what will happen.

AlanCoding: This fix appears to be very strongly related to https://github.com/ansible/awx/pull/11927

I think this needs more careful thought about expectations. If a missing instance is going to result in an error (as one would expect if registration is done by the entrypoint or installer) then fine, that's not a bad decision. But care needs to be taken to carry that out in a way where other changes won't be fighting it.

AlanCoding:
https://github.com/shanemcd/awx/pull/60

tools/ansible/roles/dockerfile/files/launch_awx.sh
shanemcd: I tried doing that and it made the tests blow up... I'm going to keep this, it shouldn't hurt anything.

tools/ansible/roles/dockerfile/templates/supervisor.conf.j2
```diff
@@ -12,9 +12,10 @@ directory = /awx_devel
 {% else %}
 command = nginx -g "daemon off;"
 {% endif %}
-autostart = true
```
AlanCoding, line 15: removed because it's the default... just want to get that stated as a sanity check.

tools/docker-compose/supervisor.conf
```diff
@@ -5,79 +5,75 @@ nodaemon=true
 
 [program:awx-dispatcher]
 command = make dispatcher
-autostart = true
 autorestart = true
-stopwaitsecs = 1
-stopsignal=KILL
+startsecs = 30
```
AlanCoding, line 9: Looks like the default for `stopwaitsecs` is 10. This would be my preference.

The dispatcher with running jobs may not respond well to SIGTERM right now. This is something I packed into https://github.com/ansible/awx/pull/11745, along with a supervisor config change which is almost identical to the change you make here (for stopwaitsecs and stopsignal).

But that's getting kicked out until next release. If you want, I could look into a targeted change to _just_ process SIGTERM signals to cancel a job so that it's not left running orphaned on the execution node.

AlanCoding:
I'd like to make a note to port changes to `startsecs` (at least) to the traditional installer.

AlanCoding (approved):

AlanCoding:
This behavior could still use some cleanup.

```
$ awx-manage provision_instance
  File "/awx_devel/awx/__init__.py", line 170, in manage
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/management/__init__.py", line 413, in execute
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/management/base.py", line 354, in run_from_argv
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/management/base.py", line 398, in execute
  File "/usr/lib64/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/awx_devel/awx/main/management/commands/provision_instance.py", line 37, in handle
    self._register_hostname(options.get('hostname'), options.get('node_type'), options.get('uuid'))
  File "/awx_devel/awx/main/management/commands/provision_instance.py", line 25, in _register_hostname
    (changed, instance) = Instance.objects.get_or_register()
  File "/awx_devel/awx/main/managers.py", line 182, in get_or_register
    return (False, self.me())
  File "/awx_devel/awx/main/managers.py", line 109, in me
    raise RuntimeError("No instance found with the current cluster host id")
RuntimeError: No instance found with the current cluster host id
```

I'm tempted to just not run this in non-OCP environments. This seems the "safest" from a maintainability mindset.

```diff
diff --git a/awx/main/management/commands/provision_instance.py b/awx/main/management/commands/provision_instance.py
index b873cb2fd5..aeae496306 100644
--- a/awx/main/management/commands/provision_instance.py
+++ b/awx/main/management/commands/provision_instance.py
@@ -1,8 +1,9 @@
 # Copyright (c) 2015 Ansible, Inc.
 # All Rights Reserved
 
-from django.core.management.base import BaseCommand
+from django.core.management.base import BaseCommand, CommandError
 from django.db import transaction
+from django.conf import settings
 
 from awx.main.models import Instance
 
@@ -22,6 +23,8 @@ class Command(BaseCommand):
 
     def _register_hostname(self, hostname, node_type, uuid):
         if not hostname:
+            if not settings.AWX_AUTO_DEPROVISION_INSTANCES:
+                raise CommandError('Registering with values from settings only intended for use in K8s installs')
             (changed, instance) = Instance.objects.get_or_register()
         else:
             (changed, instance) = Instance.objects.register(hostname=hostname, node_type=node_type, uuid=uuid)
```

AlanCoding:
^ that might be better done in followup work, because it could cascade to some unused code deletions.

PR #11956, opened 2022-03-24
setting environment variables for credential plugins

amolgautam25:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

We have several parts in the code where we apply the AWX_TASK_ENV before making requests to external systems like this:
``` 
with set_environ(**settings.AWX_TASK_ENV): 
     validated = get_licenser().validate_rh(user, pw) 
```

There is no indication that this was ever done for credential lookup plugins. This PR does exactly the same, so that credential lookup ( while the job is being prepared ) can redirect request to proxy server.


AlanCoding (changes requested):
I'd rather not change the contract for credential plugins, and instead use this wrapper in the callers that invokes these.

AlanCoding (approved):

PR #11957, opened 2022-03-24
fixed hardcode tmp ha.py

lucas-benedito:
##### SUMMARY
Fixing the hardcoded /tmp in ha.py to use AWX_ISOLATION_BASE_PATH values instead

##### ISSUE TYPE
 - Bugfix Pull Request

### DIFF
--- a/awx/main/models/ha.py
+++ b/awx/main/models/ha.py
@@ -187,7 +187,7 @@ class Instance(HasPolicyEditsMixin, BaseModel):
         """
         vargs = dict()
         if settings.AWX_CLEANUP_PATHS:
-            vargs['file_pattern'] = '/tmp/{}*'.format(JOB_FOLDER_PREFIX % '*')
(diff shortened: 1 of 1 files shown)


awx/main/models/ha.py
```diff
@@ -187,7 +187,7 @@ def get_cleanup_task_kwargs(self, **kwargs):
         """
         vargs = dict()
         if settings.AWX_CLEANUP_PATHS:
-            vargs['file_pattern'] = '/tmp/{}*'.format(JOB_FOLDER_PREFIX % '*')
+            vargs['file_pattern'] = '{}/{}*'.format(settings.AWX_ISOLATION_BASE_PATH, JOB_FOLDER_PREFIX % '*')
```
AlanCoding: We should use `os.path.join`, because we don't know if the user's value for the setting has a trailing `/`

```python
>>> os.path.join('/home/alan/', 'awx_123')
'/home/alan/awx_123'
```

So you see that solves the problem of removing the extra `/`, whether or not I add it.

AlanCoding (approved):
lgtm

PR #11961, opened 2022-03-24
Respect system configured tmp directory

shanemcd:
We found this when debugging https://github.com/ansible/awx/pull/11957

The only gotcha is that after changing this setting you need to either:

- Run this in `awx-manage shell_plus`:
  - `Setting.objects.filter(key='AWX_ISOLATION_BASE_PATH').delete()`
- Click "Revert" for the setting in the UI under `/#/settings/jobs/edit`

### DIFF
--- a/awx/settings/defaults.py
+++ b/awx/settings/defaults.py
@@ -5,6 +5,7 @@ import base64
 import os
 import re  # noqa
 import sys
+import tempfile
 from datetime import timedelta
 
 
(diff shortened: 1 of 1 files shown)


AlanCoding:
I heard that @Loicavenel was advocating for changing the default to `/var/lib/awx/tmp` (or something like that). I don't believe this will accomplish that, although it could help users re-configure it to do so. I want @ffirg to see this, because I thought he had noted this problem (but I could be mis-remembering).

AlanCoding:
Could also use a release note.

AlanCoding (approved):
pretty clear improvement, :+1: from me

Loicavenel:
Customers having issues with RH using /tmp, which in some cases is too small and then 100% of /tmp file system is used.. there is no recommendation ...

ffirg:
We need to make this user configurable. Often /var is also part of the / filesystem so that might not necessarily solve the issue. We would also have other issues for controller if /var filled up, making the potential problem worse.

I think leaving /tmp as the default but allowing a user configurable override is the way to go. We would probably want to manifest this as a var back in the installer as well so they can set it at install time.

AlanCoding:
@ffirg this setting is already configurable in the UI. Sound like you might have some kind of other issue that I don't have the details for.

PR #11964, opened 2022-03-25
Escape name__regex and name__iregex

nixocio:
Escacpe name__regex and name__iregex. Escaping the value for those
keys when creating a smart inventory is a work around for the
pyparsing code on the API side for special characters. This will just
display an extra escape when showing the host_filter on details page.

This PR is a collaboration with @john-westcott-iv.

Details

image

Smart Inventory host filter being applied

image


AlanCoding (approved):

marshmalien (approved):

InnocentK (approved):
Looks good! The above failing test is unrelated so this is safe to merge

PR #11965, opened 2022-03-25
Small style update

nixocio:
Small style update to make a bit more consistent.


AlexSCorey (approved):
Looks good

nixocio:
I looked at the cypress test failure, and it is not related to this change.

PR #11968, opened 2022-03-28
Minor tweaks to ansible-runner cleanup task arguments

AlanCoding:
##### SUMMARY
Fixes a bug reported multiple times, I don't have AWX issue links ready right now.

I believe this was caused by 2 errors both making each other worse. One, `grace_period` not being provided was supposed to use the default of 60 minutes. It didn't. The programmatic interface was not considered deeply here, and the result is that it resulted in _no_ grace period. This should still only surface as very infrequent race conditions.

The other issue is that job ids were not returned if it was using the instance as a controller node. Both the execution and control nodes have the same directory.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
This is still tricky to reproduce, because the task needs to be ran during the period between being dispatched and sending the job off to receptor, which has no explicitly blocking steps. This might be best accomplished by some aggressive unit tests.

### DIFF
--- a/awx/main/models/ha.py
+++ b/awx/main/models/ha.py
@@ -186,10 +186,12 @@ class Instance(HasPolicyEditsMixin, BaseModel):
         returns a dict that is passed to the python interface for the runner method corresponding to that command
         any kwargs will override that key=value combination in the returned dict
         """
-        vargs = dict(grace_period=60)  # grace period of 60 minutes, need to set because CLI default will not take effect
+        vargs = dict()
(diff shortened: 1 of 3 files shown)


AlanCoding:
I added tests here, and I believe they represent the best available means of regression testing for this. Without both the changes here, they fail.

With the current state of this code, I'm totally happy with this.

shanemcd (approved):

PR #11971, opened 2022-03-29
Add setuptools_scm[toml] and pkgconfig to requirements file

simaishi:
Follow up for https://github.com/ansible/awx/pull/11654

As a part of the django update, xmlsec was updated. The new version added setup_requires and offline rpm build is now failing becuse those dependencies are missing in the requirements file.

I'm not sure where in the file those build deps should be added, so I added toward the end of the file where pip/setuptools are.


shanemcd (approved):
Thanks @simaishi !

shanemcd:
Oops. API tests are failing because we are missing a license file.

jbradberry:
Sounds like this change will need to be added to the other branches as well, then.

jbradberry (approved):
This seems fine, aside from the missing license file.

simaishi:
Added the license files.

Issue #6807, opened 2020-04-22
Optimize fact cache for scale


wenottingham:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
When the fact cache is utilized in AWX, we query the database for the fact blobs at job start for each host, write them to the execution bundle alongside the project, and send it off for execution.

This can lead to a notable startup penalty when a) the facts for a host are large b) the size of the inventory is large.

We should investigate what can be done to optimize this, both tactically at the in terms of optimizing queries, and strategically in terms of potentially moving to a different mode of fact storage.

chrismeyersfsu:
The meat of the implementation is in [start_job_fact_cache](https://github.com/ansible/awx/blob/devel/awx/main/models/jobs.py#L830) and [finish_job_fact_cache](https://github.com/ansible/awx/blob/devel/awx/main/models/jobs.py#L855) To populate the Ansible fact cache we pull the per-host fact records out of the database and write them to individual `.json` files. AWX then enables the Ansible [jsonfile](https://docs.ansible.com/ansible/2.8/plugins/cache/jsonfile.html) fact cache plugin.

Updating/pushing facts back into the database is handled in `finish_job_fact_cache`. About the only "optimization" that we do is check to the see if the modification timestamp has changed on the per-host fact cache. If so, overwrite the old facts in the database for that host with the fact-cache version from the file.

#### Thoughts:
* Facts need to live in the database for the smart inventory feature.
* How slow is our current system? Why is it slow? Make an inventory with 10,000 hosts and default Ansible facts or 10x and 100x fact size.
  * Maybe there isn't an index on our search `hosts = hosts.filter(ansible_facts_modified__gte=timeout)`
  * Maybe getting all the fields for the host (i.e. the host vars) is slowing things down. Try `.values()`
  * Maybe making a file per-host and json dumping the data is slow?
  * Maybe Ansible loads the fact cache on init?
* edit: removed Redis idea. With the distributed nature of AWX and the direction we are going, we aren't going to able to lean on Redis.

shanemcd:
Maybe helpful: https://docs.ansible.com/ansible/latest/dev_guide/developing_plugins.html#cache-plugins

chrismeyersfsu:
[...]

Issue #5567, opened 2019-12-23
Performance Issue with large inventory with large cache and concurrent jobs


dnguyen9:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

When launching a job template with fact cache enabled concurrently (200-400 jobs), AWX will fetch retrieve the entire inventory of host facts, for each jobs and stores them into AWX file system and memcached.  This becomes problematic when each hosts in the inventory has a larger facts cached on the database.

For the main_host table size of 300MB,  we see repeatability the following queries made to the database by running SQL Profiler

![image](https://user-images.githubusercontent.com/16950383/71367375-dea3d600-2572-11ea-99f7-f02194c14ddc.png)

Code:

https://github.com/ansible/awx/blob/c6dc69c68b9bb3a3dfcdced9b6f774039451c97f/awx/main/tasks.py#L1252-L1256

https://github.com/ansible/awx/blob/c6dc69c68b9bb3a3dfcdced9b6f774039451c97f/awx/main/models/jobs.py#L780-L802

##### ENVIRONMENT
* AWX version: 5.0
* AWX install method: docker on linux
* Ansible version:  2.8
* Operating System: RHEL 7
* Web Browser: Chrome/ Firefox

##### STEPS TO REPRODUCE

1. create a medium-ish inventory ( 2000 hosts)
2. populate each host with 50KB of  ansible_facts data ( we can just postgres sql to do this)
3. create a job template and enable concurrent jobs and enable fact facting
4. launch 200-400 jobs

##### EXPECTED RESULTS

Job response time  should be clustered and within less than 5 seconds. this is without fact caching enable:

![image](https://user-images.githubusercontent.com/16950383/71367621-9c2ec900-2573-11ea-8752-e573f36f8e8c.png)

##### ACTUAL RESULTS

jobs should be executed quick but response time for a job is a stepper function not to mention about 300MB x 200 = 6GB ansible_facts of data transferred. Note a lot of jobs was harakiri-ed  if the response took longer than 120second. 

![image](https://user-images.githubusercontent.com/16950383/71367564-76a1bf80-2573-11ea-85fa-c03b4c307577.png)

ryanpetrello:
@dnguyen9,

> For the main_host table size of 300MB,  we see repeatability the following queries made to the database by running SQL Profiler

![image](https://user-images.githubusercontent.com/16950383/71367375-dea3d600-2572-11ea-99f7-f02194c14ddc.png)

[...]

PR #11974, opened 2022-03-29
Limit fact cache for server targeted in limits field

dnguyen9:
##### SUMMARY

Modify the behavior of fact cache so it will pull for servers that are identified in the limit field. this helps to remediate scaling problems when the host tables for a specific inventory is large. 

related #6807
related #5567

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```


AlanCoding (changes requested):
The logic to process a limit lives in ansible-core

https://github.com/ansible/ansible/blob/94b73d66d53ca0df9911d1dd412bc1d3c9181b1b/lib/ansible/inventory/manager.py#L589

Somehow porting that logic over to AWX in order to internally parse limits is something I would be willing to consider (with applications in several parts of the app). This approach would be brittle and undocumented.

AlanCoding:
The general possibility still remains open, but closing this for now

PR #11979, opened 2022-03-30
Move AWX_MOUNT_ISOLATED_PATHS_ON_K8S to be closer to other booleans

nixocio:
Move AWX_MOUNT_ISOLATED_PATHS_ON_K8S to be closer to other booleans on
the layout.

With this change

image

Without

image


marshmalien (approved):
💯

PR #12004, opened 2022-04-04
Downgrade min required node LTS

nixocio:
Downgrade min required node LTS


shanemcd:
What about this one?

https://github.com/ansible/awx/blob/b7fa24ddea7fc600256be4337e1a9e0c3d2c9c1d/tools/ansible/roles/dockerfile/templates/Dockerfile.j2#L70

kialam (approved):
This is good to merge as long as tests pass.

nixocio:
I verified with @akus062381 and the test failures are not related to this PR changes.

PR #12028, opened 2022-04-11
Add saved replies dir and default reply

nixocio:
Add save replies dir and default reply


AlexSCorey:
Will this work create auto replies for people that file issues on awx?

nixocio:
hey @AlexSCorey I do not think so, it was @shanemcd suggestion to have a place to store common replies - to help during triage.

rebeccahhh:
👍🏻 thanks for starting the process. I'll follow up today with some more content I've been gathering over the last few months.

rebeccahhh (approved):

Issue #11671, opened 2022-02-02
[Tech debt] Task manager fetches resources (instances, jobs, etc.) twice


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Running the task manager, it repeats itself doing the same thing twice. Documenting partial call stack here. Both start with the main call path to the task manager:

https://github.com/ansible/awx/blob/bff49f2a5fd8929808b2890bc2d16279c523cc53/awx/main/scheduler/task_manager.py#L674

Both of these ultimately do the same queries, and the only reason we have two is because multiple developers wrote them, probably mutually unaware of the other data structures that were built, or an unwillingness to modify the code in order to combine efforts.

We should combine them. An attempt at this problem was started in https://github.com/ansible/awx/pull/11651

**first fetch**

https://github.com/ansible/awx/blob/bff49f2a5fd8929808b2890bc2d16279c523cc53/awx/main/scheduler/task_manager.py#L634

https://github.com/ansible/awx/blob/bff49f2a5fd8929808b2890bc2d16279c523cc53/awx/main/scheduler/task_manager.py#L71

This will then fetch _jobs_ indirectly as `instance.remaining_capacity` is referenced.

**second fetch**

https://github.com/ansible/awx/blob/bff49f2a5fd8929808b2890bc2d16279c523cc53/awx/main/scheduler/task_manager.py#L662

https://github.com/ansible/awx/blob/bff49f2a5fd8929808b2890bc2d16279c523cc53/awx/main/scheduler/task_manager.py#L620

https://github.com/ansible/awx/blob/bff49f2a5fd8929808b2890bc2d16279c523cc53/awx/main/scheduler/task_manager.py#L601

https://github.com/ansible/awx/blob/bff49f2a5fd8929808b2890bc2d16279c523cc53/awx/main/managers.py#L237

This case attempts to avoid _fetching the jobs list_ by reusing the main list that the task manager operates on.

Consider this is a sub-issue of https://github.com/ansible/awx/issues/11306

kdelee:
looks like we also fetch jobs indirectly with 
https://github.com/ansible/awx/blob/bff49f2a5fd8929808b2890bc2d16279c523cc53/awx/main/scheduler/task_manager.py#L80

https://github.com/ansible/awx/blob/bff49f2a5fd8929808b2890bc2d16279c523cc53/awx/main/models/ha.py#L155-L162

AlanCoding:
[...]

PR #12030, opened 2022-04-11
drop call that queries all running and waiting jobs in task manager

kdelee:
this is to fix one more place in the task manager where we end up
querying all running and waiting jobs.

Partial fix for https://github.com/ansible/awx/issues/11671

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/models/ha.py
+++ b/awx/main/models/ha.py
@@ -390,7 +390,7 @@ class InstanceGroup(HasPolicyEditsMixin, BaseModel, RelatedJobsMixin):
         for i in instances:
             if i.node_type not in (capacity_type, 'hybrid'):
                 continue
-            if i.jobs_running == 0:
+            if (hasattr(i, 'jobs_running') and i.jobs_running == 0) or i.remaining_capacity == i.capacity:
                 if largest_instance is None:
                     largest_instance = i
                 elif i.capacity > largest_instance.capacity:
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -79,7 +79,6 @@ class TaskManager:
                 node_type=instance.node_type,
                 remaining_capacity=instance.remaining_capacity,
                 capacity=instance.capacity,
-                jobs_running=instance.jobs_running,
                 hostname=instance.hostname,
             )
             for instance in instances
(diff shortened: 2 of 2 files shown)


awx/main/models/ha.py
```diff
@@ -390,7 +390,7 @@ def find_largest_idle_instance(instances, capacity_type='execution'):
         for i in instances:
             if i.node_type not in (capacity_type, 'hybrid'):
                 continue
-            if i.jobs_running == 0:
+            if (hasattr(i, 'jobs_running') and i.jobs_running == 0) or i.remaining_capacity == i.capacity:
```
kdelee, line 393: I figure `jobs_running == 0` is equivalent to `remaining_capacity == total_capacity` and both indicate an "idle" instance

awx/main/scheduler/task_manager.py
```diff
@@ -79,7 +79,6 @@ def after_lock_init(self):
                 node_type=instance.node_type,
                 remaining_capacity=instance.remaining_capacity,
```
AlanCoding, line 80: I can't see how `jobs_running` would be any more expensive than `remaining_capacity`, and following the logic, it seems that it isn't.

I want to voice hesitation here, because this may be an optimization without looking at the profiling data first. This will get us some performance, yes, but as long as we have other similar fields calculated it will be a drop in the bucket.

I also worry whether this PR is going in the right direction, because `jobs_running` is somewhat easy to track. If we remove a field that's easy to track, but leave other fields that competed for the same purpose, that could leave us in a worse spot if the fields we leave are messy and difficult to maintain.

However, that doesn't seem to be the case, because after initialization we do appear to track changes to `remaining_capacity` appropriately.

AlanCoding (approved):

AlanCoding:
Following up from my comment... what can we do about this?

Assuming we are not throwing away the entire task manage at this point and wish to further optimize, `remaining_capacity` could optionally take the task manager jobs list. You'd need to avoid using it as a `@property`, so call it `get_remaining_capacity`. Then this would have an optional kwarg, which would be a full listing of all active jobs. The task manager has such a listing.

This implementation would be similar to how we pass in tasks to capacity_values, in `self.graph = InstanceGroup.objects.capacity_values(tasks=tasks, graph=self.graph)`. The idea is that these methods will need to be able to compute values based on the in-memory list of tasks, to avoid re-fetching.

kdelee:
test results looking good. Saw some flake with websockets, but this disappeared on re-run. Doesn't seem related so I'm going to merge.

PR #12037, opened 2022-04-12
calculate remaining capacity in static method 

kdelee:
this is to avoid additional queries when we allready have all
the active jobs fetched in the task manager

Additional follow up from https://github.com/ansible/awx/issues/11671

##### SUMMARY

By calling `remaining_capacity` on all the instances, we end up running many queries for data we already have. Avoid fetching all running/waiting jobs an additional time by calculating remaining capacity with the already fetched instance and job data.

##### ISSUE TYPE

 - Bugfix Pull Request (techdebt)

##### COMPONENT NAME

 - API

##### ADDITIONAL INFORMATION

Still working on using the django debug toolbar to test this. But it passes tests and "works" as far as, the task manager is still running and scheduling jobs and tabulating remaining capacity.

### DIFF
--- a/awx/main/models/ha.py
+++ b/awx/main/models/ha.py
@@ -161,6 +161,25 @@ class Instance(HasPolicyEditsMixin, BaseModel):
     def remaining_capacity(self):
         return self.capacity - self.consumed_capacity
 
+    @staticmethod
+    def update_remaining_capacity(instances, jobs):
+        """Takes mapping of hostname to SimpleNamespace instance like objects and a list of jobs.
+
+        Computes remaining capacity for all the instances based on currently running and waiting jobs.
+
+        No return value, updates the "remaining_capacity" field on the SimpleNamespace instance like object in place.
+        For use in the task manager to avoid refetching jobs from the database.
+        """
+        for job in jobs:
+            if job.status not in ['waiting', 'running']:
+                continue
+            control_instance = instances.get(job.controller_node, '')
+            execution_instance = instances.get(job.execution_node, '')
+            if execution_instance and execution_instance.node_type in ('hybrid', 'execution'):
+                instances[job.execution_node].remaining_capacity -= job.task_impact
+            if control_instance and control_instance.node_type in ('hybrid', 'control'):
+                instances[job.controller_node].remaining_capacity -= settings.AWX_CONTROL_NODE_TASK_IMPACT
+
     @property
     def jobs_running(self):
         return UnifiedJob.objects.filter(
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -64,7 +64,7 @@ class TaskManager:
 
         self.time_delta_job_explanation = timedelta(seconds=30)
 
-    def after_lock_init(self):
+    def after_lock_init(self, all_sorted_tasks):
         """
         Init AFTER we know this instance of the task manager will run because the lock is acquired.
         """
@@ -77,7 +77,7 @@ class TaskManager:
             SimpleNamespace(
                 obj=instance,
                 node_type=instance.node_type,
-                remaining_capacity=instance.remaining_capacity,
+                remaining_capacity=instance.capacity,  # Updated with Instance.update_remaining_capacity by looking at all active tasks
                 capacity=instance.capacity,
                 hostname=instance.hostname,
             )
@@ -85,6 +85,9 @@ class TaskManager:
         ]
 
         instances_by_hostname = {i.hostname: i for i in instances_partial}
+
+        # updates remaining capacity value based on currently running and waiting tasks
+        Instance.update_remaining_capacity(instances_by_hostname, all_sorted_tasks)
 
         for rampart_group in InstanceGroup.objects.prefetch_related('instances'):
             if rampart_group.name == settings.DEFAULT_CONTROL_PLANE_QUEUE_NAME:
@@ -603,7 +606,7 @@ class TaskManager:
         finished_wfjs = []
         all_sorted_tasks = self.get_tasks()
 
-        self.after_lock_init()
+        self.after_lock_init(all_sorted_tasks)
 
         if len(all_sorted_tasks) > 0:
             # TODO: Deal with


awx/main/scheduler/task_manager.py
```diff
@@ -302,6 +304,8 @@ def post_commit():
         connection.on_commit(post_commit)
 
     def process_running_tasks(self, running_tasks):
+        # update remaining capacity on instances
+        self.instances_by_hostname = Instance.get_remaining_capacity(self.instances_by_hostname, running_tasks)
```
AlanCoding: I don't like the location of this, specifically, I think the call should happen earlier next to more relevant code.

Maybe I get where you're coming from, in that, you want to only consider running tasks for capacity usage.

I would outright pass `all_sorted_tasks` into `after_lock_init` and then you could run it at any point after you created `self.instances_by_hostname`.

awx/main/models/ha.py
```diff
@@ -161,6 +161,24 @@ def consumed_capacity(self):
     def remaining_capacity(self):
         return self.capacity - self.consumed_capacity
 
+    @staticmethod
+    def get_remaining_capacity(instances={}, jobs=[]):
```
AlanCoding: Name isn't good anymore. I don't even like that this is on the `Instance` model, because the `instances` data structure is not populated with `Instance`s. No, the `instances` values are full of `types.SimpleNamespace`s, which are informally referred to as "partial instances".

This more-or-less starts to do the same thing as `capacity_values`. That's okay by me, because our current direction should lead to deleting the `capacity_values` method.

awx/main/models/ha.py
```diff
@@ -161,6 +161,24 @@ def consumed_capacity(self):
...
+        """Takes mapping of hostname to SimpleNamespace instance like objects and a list of jobs.
+
+        Computes remaining capacity for all the instances.
+
+        Updates the "remaining_capacity" field on the SimpleNamespace instance like object.
+        For use in the task manager to avoid refetching jobs from the database.
+        """
+        for job in jobs:
```
AlanCoding, line 173: You should `continue` if the job's status is not something that counts toward capacity usage. You should double-check, but I believe that only "running" and "waiting" jobs count.

AlanCoding:
Testing with Django debug toolbar, in a quiet system with 3 nodes (control, hop, execution).

Before this, I get 17 queries, including

```
/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django_guid/middleware.py in middleware(75)
  response = get_response(request)
  
/awx_devel/./awx/sso/middleware.py in process_request(39)
  return self.get_response(request)
  
/var/lib/awx/venv/awx/lib64/python3.9/site-packages/crum/__init__.py in __call__(97)
  response = self.get_response(request)
  
/usr/lib64/python3.9/contextlib.py in inner(79)
  return func(*args, **kwds)
  
/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/views/decorators/csrf.py in wrapped_view(54)
  return view_func(*args, **kwargs)
  
/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/views/generic/base.py in view(70)
  return self.dispatch(request, *args, **kwargs)
  
/awx_devel/./awx/api/generics.py in dispatch(340)
  return super(APIView, self).dispatch(request, *args, **kwargs)
  
/var/lib/awx/venv/awx/lib64/python3.9/site-packages/rest_framework/views.py in dispatch(506)
  response = handler(request, *args, **kwargs)
  
/awx_devel/./awx/api/views/root.py in get(60)
  TaskManager().schedule()
  
/awx_devel/./awx/main/scheduler/task_manager.py in schedule(652)
  self._schedule()
  
/awx_devel/./awx/main/scheduler/task_manager.py in _schedule(612)
  self.after_lock_init()
  
/awx_devel/./awx/main/scheduler/task_manager.py in after_lock_init(76)
  instances_partial = [
  
/awx_devel/./awx/main/scheduler/task_manager.py in <listcomp>(82)
  jobs_running=instance.jobs_running,
  
/awx_devel/./awx/main/models/ha.py in jobs_running(166)
  return UnifiedJob.objects.filter(
```

I applied this change and my hacks for testing with DDT.

The number of queries seems to drop down to 13, and the queries for the job list look pretty clean. I think you're on track here. I do want to address those comments I left. I don't think the changes I asked for will affect the performance.

awx/main/scheduler/task_manager.py
kdelee: ok I can do that

awx/main/models/ha.py
kdelee: yeah I'll do that when I start passing all the jobs (I had just passed already selected running/waiting jobs)

awx/main/models/ha.py
jbradberry: @kdelee be careful about using mutable default parameters like this.  I think probably we should make these mandatory, with no defaults.

awx/main/scheduler/task_manager.py
kdelee: done

awx/main/models/ha.py
AlanCoding: I agree with @jbradberry 

I stand by my point that this method should be renamed.

I retract my request to move the method, as it is marked as `@staticmethod` and operates on partial instances. A pattern for this already exists, and a refactor of that should be handled separately from your present work.

kdelee: ok will move it back, rename it, and remove the defaults

kdelee: yeah the task manager is not using `capacity_values` anymore, but the API is...you're deprecation of `committed_capacity` should help move us in that direction. `consumed_capacity` also uses it though, you should maybe think about if we can deprecate that too (unless we want to do some other re-implementation w/o the capacity_values function)

kdelee:
@AlanCoding @jbradberry  I think I've addressed all your comments

awx/main/models/ha.py
AlanCoding: The latter, we can move the serializer to use the method you wrote, and then purge references to the old `capacity_values`, and delete it. This is assuming this goes in, and that the field deprecation is merged too.

AlanCoding (approved):

kdelee:
yolo results looking good. Going to merge this

PR #12039, opened 2022-04-12
Adds popover help text to project details, and unifies those strings …

AlexSCorey:
##### SUMMARY
This addresses #11863.  It does 2 new things:
1) Introduces a string file where the help text that is used by the form and the details view will be stored.  If changes need to be made they only have to be made in 1 place.  I considered naming that file `project.strings.js` but `project.helptext.js` and I didn't because I wanted to stick with out current convention.  I could be convinced otherwise.
2) Added a prop `isDetailItem` to popover components so that they are grey on load and turn black onHover after consulting with @trahman73 and Tina Yip.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

##### Additional Information
hover

### DIFF
--- a/awx/ui/src/components/Popover/Popover.js
+++ b/awx/ui/src/components/Popover/Popover.js
@@ -10,6 +10,8 @@ const PopoverButton = styled.button`
   padding: var(--pf-global--spacer--xs);
   margin: -(var(--pf-global--spacer--xs));
   font-size: var(--pf-global--FontSize--sm);
+  --pf-c-form__group-label-help--Color: var(--pf-global--Color--200);
+  --pf-c-form__group-label-help--hover--Color: var(--pf-global--Color--100);
 `;
 
 function Popover({ ariaLabel, content, header, id, maxWidth, ...rest }) {
--- a/awx/ui/src/screens/Project/ProjectDetail/ProjectDetail.test.js
+++ b/awx/ui/src/screens/Project/ProjectDetail/ProjectDetail.test.js
@@ -20,6 +20,12 @@ jest.mock('react-router-dom', () => ({
     url: '/projects/1/details',
   }),
 }));
+jest.mock('hooks/useBrandName', () => ({
+  __esModule: true,
+  default: () => ({
+    current: 'AWX',
+  }),
+}));
 describe('<ProjectDetail />', () => {
   const mockProject = {
     id: 1,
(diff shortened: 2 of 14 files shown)


keithjgrant (approved):
I like this approach a lot 👍🏻

awx/ui/src/screens/Project/ProjectDetail/ProjectDetail.test.js
```diff
@@ -20,6 +20,18 @@ jest.mock('react-router-dom', () => ({
     url: '/projects/1/details',
   }),
 }));
+jest.mock('hooks/useBrandName', () => ({
+  __esModule: true,
+  default: () => ({
+    current: 'AWX',
+  }),
+}));
+jest.mock('hooks/useBrandName', () => ({
```
nixocio: We could remove one of those mocks.

awx/ui/src/screens/Project/ProjectDetail/ProjectDetail.test.js
```diff
@@ -125,17 +137,20 @@ describe('<ProjectDetail />', () => {
     expect(dateDetails.at(1).prop('date')).toEqual(
       '2019-10-10T01:15:06.780490Z'
     );
+    console.log(wrapper.debug());
```
kialam: remove line

awx/ui/src/screens/Project/shared/ProjectHelpTextStrings.js
```diff
@@ -0,0 +1,142 @@
+import React from 'react';
+import { t } from '@lingui/macro';
+import getDocsBaseUrl from 'util/getDocsBaseUrl';
+import { useConfig } from 'contexts/Config';
+import useBrandName from 'hooks/useBrandName';
+
+const ProjectHelpTextStrings = () => ({
+  execution_environment: t`The execution environment that will be used for jobs that use this project. This will be used as fallback when an execution environment has not been explicitly assigned at the job template or workflow level.`,
```
kialam: use camelCase instead of snake_case

awx/ui/src/screens/Project/shared/ProjectSubForms/ArchiveSubForm.js
```diff
@@ -14,25 +14,7 @@ const ArchiveSubForm = ({
...
-            <li>
-              <code>
-                https://github.com/username/project/archive/v0.0.2.zip
-              </code>
-            </li>
-          </ul>
-        </span>
-      }
-    />
+    <UrlFormField tooltip={ProjectHelpTextStrings().archiveUrl} />
```
kialam: do we want to use the `const projectHelpStrings = ProjectHelpTextStrings();` convention here?

kialam: alternatively, could just destructure out what's needed: `const { archiveUrl } = ProjectHelpText()`.

kialam (approved):
There is a stray console.log and I left some comments about implementation details but otherwise nice job on this!

awx/ui/src/screens/Project/shared/ProjectSubForms/ArchiveSubForm.js
AlexSCorey: These are good finds.  I've elected not to destructure the object because in some files (if we were to be consistent with this format and I think we should be) it would add a bunch of lines to the file and not really provide any clarity or make it easier to maintain/read

awx/ui/src/components/Popover/Popover.js
```diff
@@ -10,9 +10,27 @@ const PopoverButton = styled.button`
...
+
+  ${(props) =>
+    props.isDetailItem &&
+    `
+    && {
+      color: #6a6e73;
+      &:hover {
+        color: var(--pf-c-check__label--Color);
+      }
+    }`}
```
marshmalien: If we add 
```
  --pf-c-form__group-label-help--Color: var(--pf-global--Color--200);
  --pf-c-form__group-label-help--hover--Color: var(--pf-global--Color--100);
  ``` 
  instead, we could get rid of the `isDetailItem` prop.

AlexSCorey:
@marshmalien I like that a lot. Thats much cleaner! Thanks.

InnocentK (approved):
Great work!

InnocentK:
The one failing test shown above is unrelated to this work so should be good to merge

PR #12045, opened 2022-04-13
Only fetch fields we need in task manager

kdelee:
By using .only we select fewer columns, avoiding potentially large
fields that we never reference.

Also, small tweak to eliminate what was a duplicate dictionary of
hostname:instance, because we don't need build and carry two copies of
the same data.

##### SUMMARY

Jobs have lots of fields, but we only need a few of them for the task manager.

at @AlanCoding 's suggestion and after reading docs from Django and others, this looks like a better practice because NOT using .only or .defer we end up with what is essentially a `SELECT * FROM table`.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### ADDITIONAL INFORMATION

docs:
https://docs.djangoproject.com/en/4.0/ref/models/querysets/#only

other links:
https://redwerk.com/blog/top-8-mistakes-when-working-with-django-orm/
https://testdriven.io/tips/815d1399-b284-4732-893e-deedd2aa8d49/

### DIFF
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -68,8 +68,7 @@ class TaskManager:
         """
         Init AFTER we know this instance of the task manager will run because the lock is acquired.
         """
-        instances = Instance.objects.filter(hostname__isnull=False, enabled=True).exclude(node_type='hop')
-        self.real_instances = {i.hostname: i for i in instances}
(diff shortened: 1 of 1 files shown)


jbradberry (approved):
I enthusiastically endorse.

kdelee:
downstream tests looking good, going to merge. thanks @jbradberry !

PR #12050, opened 2022-04-13
move static methods used by task manager

kdelee:
These static methods were being used to act on Instance-like objects
that were SimpleNamespace objects with the necessary attributes.

This change introduces dedicated classes to replace the SimpleNamespace
objects and moves the formerlly staticmethods to a place where they are
more relevant instead of tacked onto models to which they were only
loosly related.

##### SUMMARY

follow up from conversation with @AlanCoding  for some cleanup in the task manager to prepare for more changes down the road.

##### ISSUE TYPE

 - Bugfix Pull Request
 
##### COMPONENT NAME

 - API
 
##### ADDITIONAL INFORMATION

need to test this some more

### DIFF
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -6,7 +6,6 @@ from datetime import timedelta
 import logging
 import uuid
 import json
-from types import SimpleNamespace
 
 # Django
 from django.db import transaction, connection
@@ -19,7 +18,6 @@ from awx.main.dispatch.reaper import reap_job
 from awx.main.models import (
     AdHocCommand,
     Instance,
-    InstanceGroup,
     InventorySource,
     InventoryUpdate,
     Job,
@@ -37,6 +35,8 @@ from awx.main.utils import get_type_for_model, task_manager_bulk_reschedule, sch
 from awx.main.utils.common import create_partition
 from awx.main.signals import disable_activity_stream
 from awx.main.scheduler.dependency_graph import DependencyGraph
+from awx.main.scheduler.task_manager_models import TaskManagerInstances
+from awx.main.scheduler.task_manager_models import TaskManagerInstanceGroups
 from awx.main.utils import decrypt_field
 
 
@@ -54,49 +54,22 @@ class TaskManager:
         The NOOP case is short-circuit logic. If the task manager realizes that another instance
         of the task manager is already running, then it short-circuits and decides not to run.
         """
-        self.graph = dict()
         # start task limit indicates how many pending jobs can be started on this
         # .schedule() run. Starting jobs is expensive, and there is code in place to reap
         # the task manager after 5 minutes. At scale, the task manager can easily take more than
         # 5 minutes to start pending jobs. If this limit is reached, pending jobs
         # will no longer be started and will be started on the next task manager cycle.
         self.start_task_limit = settings.START_TASK_LIMIT
-
         self.time_delta_job_explanation = timedelta(seconds=30)
 
     def after_lock_init(self, all_sorted_tasks):
         """
         Init AFTER we know this instance of the task manager will run because the lock is acquired.
         """
-        instances = Instance.objects.filter(hostname__isnull=False, enabled=True).exclude(node_type='hop')
-        self.real_instances = {i.hostname: i for i in instances}
-        self.controlplane_ig = None
         self.dependency_graph = DependencyGraph()
-
-        instances_partial = [
-            SimpleNamespace(
-                obj=instance,
-                node_type=instance.node_type,
-                remaining_capacity=instance.capacity,  # Updated with Instance.update_remaining_capacity by looking at all active tasks
-                capacity=instance.capacity,
-                hostname=instance.hostname,
-            )
-            for instance in instances
-        ]
-
-        instances_by_hostname = {i.hostname: i for i in instances_partial}
-
-        # updates remaining capacity value based on currently running and waiting tasks
-        Instance.update_remaining_capacity(instances_by_hostname, all_sorted_tasks)
-
-        for rampart_group in InstanceGroup.objects.prefetch_related('instances'):
-            if rampart_group.name == settings.DEFAULT_CONTROL_PLANE_QUEUE_NAME:
-                self.controlplane_ig = rampart_group
-            self.graph[rampart_group.name] = dict(
-                instances=[
-                    instances_by_hostname[instance.hostname] for instance in rampart_group.instances.all() if instance.hostname in instances_by_hostname
-                ],
-            )
+        self.instances = TaskManagerInstances(all_sorted_tasks)
+        self.instance_groups = TaskManagerInstanceGroups(instances_by_hostname=self.instances)
+        self.controlplane_ig = self.instance_groups.controlplane_ig
 
     def job_blocked_by(self, task):
         # TODO: I'm not happy with this, I think blocking behavior should be decided outside of the dependency graph
@@ -244,7 +217,7 @@ class TaskManager:
                     schedule_task_manager()
         return result
 
-    def start_task(self, task, rampart_group, dependent_tasks=None, instance=None):
+    def start_task(self, task, instance_group, dependent_tasks=None, instance=None):
         self.start_task_limit -= 1
         if self.start_task_limit == 0:
             # schedule another run immediately after this task manager
@@ -277,10 +250,10 @@ class TaskManager:
                 schedule_task_manager()
             # at this point we already have control/execution nodes selected for the following cases
             else:
-                task.instance_group = rampart_group
+                task.instance_group = instance_group
                 execution_node_msg = f' and execution node {task.execution_node}' if task.execution_node else ''
                 logger.debug(
-                    f'Submitting job {task.log_format} controlled by {task.controller_node} to instance group {rampart_group.name}{execution_node_msg}.'
+                    f'Submitting job {task.log_format} controlled by {task.controller_node} to instance group {instance_group.name}{execution_node_msg}.'
                 )
             with disable_activity_stream():
                 task.celery_task_id = str(uuid.uuid4())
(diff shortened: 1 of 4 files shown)


awx/main/scheduler/task_manager_models.py
```diff
@@ -0,0 +1,110 @@
...
+            if task.status not in ['waiting', 'running']:
+                continue
+            control_instance = self.instances_by_hostname.get(task.controller_node, '')
+            execution_instance = self.instances_by_hostname.get(task.execution_node, '')
+            if execution_instance and execution_instance.node_type in ('hybrid', 'execution'):
+                self.instances_by_hostname[task.execution_node].remaining_capacity -= task.task_impact
+            if control_instance and control_instance.node_type in ('hybrid', 'control'):
+                self.instances_by_hostname[task.controller_node].remaining_capacity -= settings.AWX_CONTROL_NODE_TASK_IMPACT
+
+    def fit_task_to_most_remaining_capacity_instance(self, task, impact=None, capacity_type=None, add_hybrid_control_cost=False):
```
kdelee: TODO need to accept the instance group name we want it to be in or move this to the TaskManagerInstanceGroups class

awx/main/scheduler/task_manager_models.py
```diff
@@ -0,0 +1,110 @@
...
+            would_be_remaining = i.remaining_capacity - impact
+            # hybrid nodes _always_ control their own tasks
+            if add_hybrid_control_cost and i.node_type == 'hybrid':
+                would_be_remaining -= settings.AWX_CONTROL_NODE_TASK_IMPACT
+            if would_be_remaining >= 0 and (instance_most_capacity is None or would_be_remaining > most_remaining_capacity):
+                instance_most_capacity = i
+                most_remaining_capacity = would_be_remaining
+        return instance_most_capacity
+
+    def find_largest_idle_instance(self, capacity_type='execution'):
```
kdelee: TODO need to accept the instance group name we want it to be in or move this to the TaskManagerInstanceGroups class

awx/main/tests/unit/models/test_ha.py
```diff
@@ -59,15 +61,17 @@ class TestInstanceGroup(object):
...
 
-        instance_picked = ig.fit_task_to_most_remaining_capacity_instance(task, instances)
+        instance_picked = tm_igs.fit_task_to_most_remaining_capacity_instance(task, instances=instances)
 
         if instance_fit_index is None:
             assert instance_picked is None, reason
         else:
             assert instance_picked == instances[instance_fit_index], reason
 
+    @pytest.mark.django_db
```
AlanCoding: Anything marked with django_db should be in the awx/main/tests/unit folder. I just want to give you the general principal here. I'm generally find converting unit tests to functional tests as the need arises based on how things are written.

kdelee: should _not_ be inthe unit test folder? e.g. move it to the functional tests?

awx/main/scheduler/task_manager_models.py
```diff
@@ -0,0 +1,120 @@
...
+        self._update_remaining_capacity(active_tasks)
+        self.instances_by_hostname = {i.hostname: i for i in instance_list}
+
+    def _update_remaining_capacity(self, tasks):
+        """Takes a list of tasks and updates the remaining capacity on our TaskManagerInstance list.
+
+        Computes remaining capacity for all the instances based on currently running and waiting tasks.
+        """
+        if self._remaining_capacity_updated:
+            # Only want to do this once
```
AlanCoding: Yeah, I just about add a comment in your last PR criticizing this, but you're wrapping that loose end up here.

awx/main/scheduler/task_manager_models.py
```diff
@@ -0,0 +1,120 @@
...
+    InstanceGroup,
+)
+
+logger = logging.getLogger('awx.main.scheduler')
+
+
+class TaskManagerInstance:
+    """A class representing minimal data the task manager needs to represent an Instance."""
+
+    def __init__(self, obj, node_type, capacity, hostname):
```
AlanCoding: I'd prefer kwargs instead of position args, and you're already kind of invoking it this way.

awx/main/scheduler/task_manager_models.py
```diff
@@ -0,0 +1,120 @@
...
+                continue
+            control_instance = self.instances_by_hostname.get(task.controller_node, '')
+            execution_instance = self.instances_by_hostname.get(task.execution_node, '')
+            if execution_instance and execution_instance.node_type in ('hybrid', 'execution'):
+                self.instances_by_hostname[task.execution_node].remaining_capacity -= task.task_impact
+            if control_instance and control_instance.node_type in ('hybrid', 'control'):
+                self.instances_by_hostname[task.controller_node].remaining_capacity -= settings.AWX_CONTROL_NODE_TASK_IMPACT
+
+    def __getitem__(self, hostname):
+        return self.instances_by_hostname.get(hostname)
```
AlanCoding, line 48: Just wait until I create an instance group with the name of `_remaining_capacity_updated`, lol

awx/main/scheduler/task_manager.py
```diff
@@ -54,49 +54,22 @@ def __init__(self):
...
-        instances_by_hostname = {i.hostname: i for i in instances_partial}
-        self.instances_by_hostname = instances_by_hostname
-
-        # updates remaining capacity value based on currently running and waiting tasks
-        Instance.update_remaining_capacity(instances_by_hostname, all_sorted_tasks)
-
-        for rampart_group in InstanceGroup.objects.prefetch_related('instances').only('name', 'instances'):
-            if rampart_group.name == settings.DEFAULT_CONTROL_PLANE_QUEUE_NAME:
-                self.controlplane_ig = rampart_group
-            self.graph[rampart_group.name] = dict(
```
AlanCoding, line 95: interesting... so `self.graph` goes away with this change. I guess that was only used as a means to reference the (partial) instance groups. This does feel much better how you've done it.

awx/main/scheduler/task_manager_models.py
```diff
@@ -0,0 +1,120 @@
...
+        return self.instance_groups.get(instance_group_name)
+
+    def fit_task_to_most_remaining_capacity_instance(
+        self, task, instance_group_name=None, impact=None, capacity_type=None, add_hybrid_control_cost=False, instances=[]
+    ):
+        impact = impact if impact else task.task_impact
+        capacity_type = capacity_type if capacity_type else task.capacity_type
+        instance_most_capacity = None
+        most_remaining_capacity = -1
+        assert instances or instance_group_name, "Need to provide either the name of an instance group or a list of instances"
```
AlanCoding: I'm a little thrown off by this method signature. It looks like we could get rid of the `instances` keyword argument. Also, other people will jump on your case for using a mutable default of `[]`. This parameter is a liability that we would do better to get rid of.

It seems that the tests use it, and modifying them is challenging without database access. Removing `instances` would require some re-configuration so that we can probably initialize everything in this file with data from outside the database. This is generally a good idea anyway.

To start with `TaskManagerInstances.__init__` will need to optionally accept a _list_ of instances as an alternative to fetching the queryset. That's the right way to do things. I think it's a good idea. Then in the unit test, you construct all of these objects from basic python.

awx/main/scheduler/task_manager_models.py
```diff
@@ -0,0 +1,120 @@
...
+    def __init__(self, obj, node_type, capacity, hostname):
+        self.obj = obj
+        self.node_type = node_type
+        self.remaining_capacity = capacity
+        self.capacity = capacity
+        self.hostname = hostname
+
+
+class TaskManagerInstances:
+    def __init__(self, active_tasks=[]):
```
AlanCoding: Another mutable default `[]`. If it's really _required_ then you could have it as a positional argument. Positional arguments aren't bad when there's only 1. Or you could allow `None`, in which case you don't fill in capacities.

awx/main/scheduler/task_manager_models.py
AlanCoding: It looks like you only use this to get the controlplane IG. If that's true, then I would lean towards axing this because it carries with it risk and awkward questions. Instead you could have a `get_control_plane` method or something.

AlanCoding: Wait, I was getting confused. At issue is the line

```python
self.controlplane_ig = self.instance_groups.controlplane_ig
```

This doesn't make sense, this would work if you override getattr, but both of these dict-like class (instance list and IG list) define _getitem_. I like the getitem, and use pattern like:

```python
self.instances[control_instance.hostname].obj
```

That looks really good. Still, if you don't use it, I wouldn't put it on `TaskManagerInstanceGroups`, and you still probably need a separate function to get the control plane group.

AlanCoding:
I think I've finished with this round of comments. Let me know if anything I wrote is not immediately actionable. Fixing these should be relatively fast keystrokes, and I can help if any turn out to be something more than that.

awx/main/scheduler/task_manager_models.py
kdelee: Great ideas! Yes this is why I was accepting the list, because of the tests

kdelee:
@AlanCoding thanks for the review. i will get back to this, got dragged into a firefighting scenario Thursday and I'm crawling my way out

kdelee:
ready for a fresh round of review

awx/main/scheduler/task_manager_models.py
```diff
@@ -0,0 +1,121 @@
...
+        return self.instances_by_hostname.get(hostname)
+
+    def __contains__(self, hostname):
+        return hostname in self.instances_by_hostname
+
+
+class TaskManagerInstanceGroups:
+    """A class representing minimal data the task manager needs to represent an InstanceGroup."""
+
+    def __init__(self, instances_by_hostname=None, instance_groups=None):
```
AlanCoding, line 57: super minor nit about the kwargs, if someone uses the defaults like `TaskManagerInstanceGroups()` then I think that would give an error due to iterating over the `None` value. But in this case, making it required may be bad, because someone could give `instance_groups` by itself.

AlanCoding (approved):
Looks great, thanks

kdelee:
Tests looking good. Talked with @AlanCoding , going to merge

PR #12055, opened 2022-04-14
Revert "Only fetch fields we need in task manager"

kdelee:
This reverts commit 868e811b3f3743b79c41186ef2617df6ecef6fe0.

Turns out this does not play well with polymorphic models.

Will try again with .defer()

Testing w/ django debug toolbar showed MANY more queries because it was doing some wild stuff with hidden polymorphic stuff.

Now that @AlanCoding has got me up to speed on how to do that debugging will try again with more information

### DIFF
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -68,7 +68,8 @@ class TaskManager:
         """
         Init AFTER we know this instance of the task manager will run because the lock is acquired.
         """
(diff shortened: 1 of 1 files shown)


AlanCoding:
Informing @jbradberry of this incoming change.

AlanCoding (approved):

jbradberry:
Well, foo.  I have some opinions about polymorphic stuff from things I had to do during the upgrade.  I'd like to jump in and help out on this.

Issue #11068, opened 2021-09-10
Standard Out tab empty even when there is related data on the API


nixocio:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

Standard Out tab empty even when there is related data on the API

### AWX version

devel

### Installation method

docker for mac

### Modifications

no

### Steps to reproduce

1. Create a Project with https://github.com/ansible/test-playbooks
2. Create a Job Template using the above, choose playbook `test_angular_var_bind.yml`
3. Launch the job template - Task expected to fail
4. Inspect fail by host

![image](https://user-images.githubusercontent.com/9053044/132908813-ac718d8c-11f7-4672-8dc9-57da9fbf2be5.png)

Job events endpoint could be queried to get the missing data.
```
api/v2/jobs/7/job_events/?id=19&
```

```json
{
   "count":1,
   "next":null,
   "previous":null,
   "results":[
      {
         "id":19,
         "type":"job_event",
         "url":"/api/v2/job_events/19/",
         "related":{
            "job":"/api/v2/jobs/7/",
            "children":"/api/v2/job_events/19/children/",
            "host":"/api/v2/hosts/1/"
         },
         "summary_fields":{
            "host":{
               "id":1,
               "name":"localhost",
               "description":""
            },
            "job":{
               "id":7,
               "name":"Demo Job Template@2:38:34 PM",
               "description":"",
               "status":"failed",
               "failed":true,
               "elapsed":2.606,
               "type":"job",
               "job_template_id":9,
               "job_template_name":"Demo Job Template@2:38:34 PM"
            },
            "role":{
               
            }
         },
         "created":"2021-09-10T18:40:49.363388Z",
         "modified":"2021-09-10T18:40:49.367787Z",
         "job":7,
         "event":"runner_on_failed",
         "counter":8,
         "event_display":"Host Failed",
         "event_data":{
            "playbook":"test_angular_var_bind.yml",
            "playbook_uuid":"7b2ce1dd-e1c2-445b-850a-ca0a44f12f1c",
            "play":"Test that curly braces \"{\" in task names",
            "play_uuid":"12d7eca1-6d21-ae0c-8025-000000000006",
            "play_pattern":"all",
[...]

PR #12064, opened 2022-04-18
Remove output/stderr tabs from host detail modals when not present

keithjgrant:
##### SUMMARY
* Fixes issues in Host Detail modal where output was sometimes not displayed.
* Hides the Output and Standard Out tabs from the modal when they are empty

Screen Shot 2022-04-18 at 9 23 47 AM

Screen Shot 2022-04-18 at 9 24 25 AM

addresses #11068 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI


marshmalien (approved):
Tested out locally and I like the improvements you've added to `getStdOutValue`. Thanks for cleaning up the bug where we only displayed the first item in the results array. Great work!

InnocentK (approved):
Looks good!

PR #12068, opened 2022-04-18
Fix the job event partition alignment

jbradberry:
##### SUMMARY
it really should be always aligned to the hour, so that real job events don't slip through the cracks.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API


fosterseth (approved):

AlanCoding (approved):

kdelee (approved):
:tada:

PR #12073, opened 2022-04-19
SCM inv source should trigger project update

fosterseth:
##### SUMMARY

scm based inventory sources should launch project updates prior to running inventory updates for that source (if update_on_launch is True for the project)

#### Implementation

The key change is to allow dependencies to have their own dependencies. A number of changes are needed to make that possible.

1. Turn off symmetry for `dependent_jobs` M2M field
It's better that this field capture some directionality of the relationship. This will help when we need to determine whether a job is blocked because a dependency has yet to run.

With change,
```python
job.dependent_jobs.add(proj)

job.dependent_jobs.all() # returns [proj]
proj.dependent_jobs.all() # returns []
proj.unifiedjobs_blocked_by.all() # returns [job]
```

2. Split up `generate_dependencies()` for Job and InventoryUpdate
This keeps the code cleaner and easier to read 

```python
if type(task) is Job:
    created_dependencies += self.gen_dep_for_job(task)
elif type(task) is InventoryUpdate:
    created_dependencies += self.gen_dep_for_inventory_update(task)
```

3. Call `generate_dependencies()` twice
Now that dependencies can have their own dependencies, we need to call generate_dependencies() a second time in the task manager loop. This need not be recursive, because the depth is only going to be a max of 2 (JobA > Inventory UpdateA > Project Update A)

Importantly, this all applies if you start an inventory source update manually. It doesn't have to be in the context of of a job launch. If a project `update_on_launch` is True, the project will be updated if corresponding job or inventory source is launched.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 20.0.2.dev249+g38ab7fc48c
```


awx/main/scheduler/dependency_graph.py
```diff
@@ -72,7 +72,12 @@ def project_update_blocked_by(self, job):
         return self.get_item(self.PROJECT_UPDATES, job.project_id)
 
     def inventory_update_blocked_by(self, job):
-        return self.get_item(self.INVENTORY_SOURCE_UPDATES, job.inventory_source_id)
+        inventory_source_block = self.get_item(self.INVENTORY_SOURCE_UPDATES, job.inventory_source_id)
+        project_block = self.get_item(self.PROJECT_UPDATES, job.inventory_source.source_project_id)
+        return inventory_source_block or project_block
+
+    def inventory_source_update_blocked_by(self, job):
+        return self.get_item(self.PROJECT_UPDATES, job.source.project_id)
```
AlanCoding: I'm not reading this coherently. Is `job` here an inventory update, or a job?

Also, the `InventoryUpdate` field that you want would be `source_project` IIRC, not `source`, so I'd expect to see reference to `.source_project_id`, but not totally sure what the thought here is.

awx/main/scheduler/dependency_graph.py
```diff
@@ -72,7 +72,12 @@ def project_update_blocked_by(self, job):
         return self.get_item(self.PROJECT_UPDATES, job.project_id)
 
     def inventory_update_blocked_by(self, job):
-        return self.get_item(self.INVENTORY_SOURCE_UPDATES, job.inventory_source_id)
+        inventory_source_block = self.get_item(self.INVENTORY_SOURCE_UPDATES, job.inventory_source_id)
+        project_block = self.get_item(self.PROJECT_UPDATES, job.inventory_source.source_project_id)
```
AlanCoding: You should check if `job.inventory_source` is null before referencing `job.inventory_source.source_project_id`, because at this point it's not clear that it's SCM type, or we could be in a deletion cascade or something.

I'm also unsure if this condition is a good idea. Your other logic already covers creating and linking a project update dependency. This looks like it would block the inventory update if another update started after its dependency... which I don't really want.

awx/main/scheduler/task_manager.py
```diff
@@ -412,7 +414,7 @@ def generate_dependencies(self, undeped_tasks):
                 continue
             # TODO: Can remove task.project None check after scan-job-default-playbook is removed
             if task.project is not None and task.project.scm_update_on_launch is True:
-                latest_project_update = self.get_latest_project_update(task)
+                latest_project_update = self.get_latest_project_update(task.project_id)
```
AlanCoding: looks like you still need to change the method signature in some other places as well.

awx/main/scheduler/dependency_graph.py
fosterseth: I mistakenly left this method in, the only method we need to update is `inventory_update_blocked_by`

awx/main/scheduler/task_manager.py
```diff
@@ -432,6 +434,12 @@ def generate_dependencies(self, undeped_tasks):
                     continue
                 latest_inventory_update = self.get_latest_inventory_update(inventory_source)
                 if self.should_update_inventory_source(task, latest_inventory_update):
+                    if inventory_source.source == "scm":
+                        latest_src_project_update = self.get_latest_project_update(inventory_source.source_project_id)
+                        if self.should_update_related_project(task, latest_src_project_update):
```
AlanCoding: This could still need more work to avoid spawning the same update twice when the playbook comes from the same project. There's also an optimization issue here, because if the inventory update was just spawned, you would already have the update in memory, but this would re-fetch it.

AlanCoding: but I'm not sure if the optimization part of this needs fixing.

awx/main/scheduler/dependency_graph.py
```diff
@@ -72,7 +72,12 @@ def project_update_blocked_by(self, job):
         return self.get_item(self.PROJECT_UPDATES, job.project_id)
 
     def inventory_update_blocked_by(self, job):
-        return self.get_item(self.INVENTORY_SOURCE_UPDATES, job.inventory_source_id)
+        inventory_source_block = self.get_item(self.INVENTORY_SOURCE_UPDATES, job.inventory_source_id)
+        if job.inventory_source:
+            project_block = self.get_item(self.PROJECT_UPDATES, job.inventory_source.source_project_id)
+        else:
+            project_block = None
```
AlanCoding: Still not sure if I agree with this. If the _generated update_ is still running, it will be blocked by

https://github.com/ansible/awx/blob/6b5aa1ec74630b27fd0125a4e28ce5c9f08530bb/awx/main/scheduler/task_manager.py#L109-L112

awx/main/scheduler/dependency_graph.py
fosterseth: the task manager links the project update as a dependency to the _job_, not the inventory update. Maybe this should be a dependency off of the inventory update itself

AlanCoding:
Thanks for the implementation notes. The CI check is telling you that you need to rebase and bump the migration number.

I'll try to get a good look at this tomorrow, and will want to sync up about the status of testing.

awx/main/models/unified_jobs.py
```diff
@@ -575,7 +575,8 @@ class Meta:
     dependent_jobs = models.ManyToManyField(
         'self',
         editable=False,
-        related_name='%(class)s_blocked_jobs+',
+        related_name='%(class)s_blocked_jobs',
+        symmetrical=False,
```
jbradberry, line 579: :fire:

awx/main/models/mixins.py
```diff
@@ -429,19 +426,28 @@ class TaskManagerUpdateOnLaunchMixin(TaskManagerUnifiedJobMixin):
...
 
 class TaskManagerInventoryUpdateMixin(TaskManagerUpdateOnLaunchMixin):
     class Meta:
         abstract = True
 
+    def get_jobs_fail_chain(self):
+        return list(self.unifiedjob_blocked_jobs.all())
+
+    def dependent_jobs_finished(self):
+        for j in self.dependent_jobs.all():
```
jbradberry: This could be written more compactly

```
return not any(j.status in ['pending', 'waiting', 'running'] for j in self.dependent_jobs.all())
```

awx/main/scheduler/task_manager.py
```diff
@@ -302,14 +304,10 @@ def create_inventory_update(self, task, inventory_source_task):
         # self.process_inventory_sources(inventory_sources)
         return inventory_task
 
-    def capture_chain_failure_dependencies(self, task, dependencies):
```
fosterseth, line 305: this was kind of misnamed -- our mixins have the method `get_jobs_fail_chain()` that is unique to types of tasks (ProjectUpdate, InventoryUpdate, Job, etc).

in essence all this method does is **add dependencies** with activity stream off, so that is what we should call it.

awx/main/scheduler/task_manager.py
```diff
@@ -376,45 +374,69 @@ def should_update_related_project(self, job, latest_project_update):
...
+                dependencies.append(project_task)
+            else:
+                dependencies.append(latest_project_update)
+
+        # Inventory created 2 seconds behind job
+        try:
+            start_args = json.loads(decrypt_field(task, field_name="start_args"))
+        except ValueError:
+            start_args = dict()
+        for inventory_source in [invsrc for invsrc in self.all_inventory_sources if invsrc.inventory == task.inventory]:
```
jbradberry: I'd turn this list comprehension into a generator expression (square brackets into parentheses)

awx/main/scheduler/task_manager.py
```diff
@@ -376,45 +374,69 @@ def should_update_related_project(self, job, latest_project_update):
...
+                continue
+            latest_inventory_update = self.get_latest_inventory_update(inventory_source)
+            if self.should_update_inventory_source(task, latest_inventory_update):
+                inventory_task = self.create_inventory_update(task, inventory_source)
+                created_dependencies.append(inventory_task)
+                dependencies.append(inventory_task)
+            else:
+                dependencies.append(latest_inventory_update)
+
+        if len(dependencies) > 0:
```
jbradberry: How about just `if dependencies:`?

jbradberry (approved):

awx/main/models/mixins.py
AlanCoding: Could also use the `ACTIVE_STATES` constant.

awx/main/scheduler/task_manager.py
```diff
@@ -302,14 +304,10 @@ def create_inventory_update(self, task, inventory_source_task):
...
         return inventory_task
 
-    def capture_chain_failure_dependencies(self, task, dependencies):
+    def add_dependencies(self, task, dependencies):
         with disable_activity_stream():
             task.dependent_jobs.add(*dependencies)
 
-            for dep in dependencies:
-                # Add task + all deps except self
-                dep.dependent_jobs.add(*([task] + [d for d in dependencies if d != dep]))
```
AlanCoding, line 311: Probably related to the migration, but could you explain how you're able to remove this?

fosterseth: yeah this change here _could_ have implications

consider this setup (JobA and JobB both share the projUpdate dependency. But only JobA depends on invUpdate)

```
            JobA                    JobB
            |  |                      |
            |  |                      |
            |  |                      |
            |  |                      |
invUpdate <-+  +--->  projUpdate <----+
````

Currently on devel, invUpdate and projUpdate are linked (because of the above code snippet that is in question). If one fails, it brings down the other. If invUpdate fails, it causes projUpdate to fail, and thus JobB would fail, even though JobB never depended on invUpdate in the first place.

I can't think of why this behavior is needed.

Because of the nature of sharing dependencies, if a project update or inventory update fails, it should only fail the jobs that directly depend on it, not other stuff.

Comment (fosterseth) [reply awx/main/scheduler/task_manager.py:311]: the dependent_jobs field plays a role in determining which jobs fail, as I mention above.

The other role for dependent_jobs is to determine whether a job is blocked from running.

in task_manager.py
```python
if not task.dependent_jobs_finished():
    blocked_by = task.dependent_jobs.first()
```
On devel this really only comes into play for Jobs

type Job
```python
def dependent_jobs_finished(self):
    for j in self.dependent_jobs.all():
        if j.status in ['pending', 'waiting', 'running']:
            return False
    return True
```

But for ProjectUpdate and InventoryUpdate, this always returned True
```python
def dependent_jobs_finished(self):
    return True
```

Therefore, the removed code in the diff doesn't really have an effect here. Jobs still have the same dependent_jobs as before. The only difference is that InventoryUpdate and ProjectUpdate do not have each other as dependencies, but this doesn't affect the `is_blocked` behavior.

awx/main/models/mixins.py
```diff
@@ -415,33 +415,41 @@ class TaskManagerJobMixin(TaskManagerUnifiedJobMixin):
...
 
 class TaskManagerInventoryUpdateMixin(TaskManagerUpdateOnLaunchMixin):
     class Meta:
         abstract = True
 
+    def get_jobs_fail_chain(self):
+        blocked_jobs = list(self.unifiedjob_blocked_jobs.all())
+        other_updates = []
+        if blocked_jobs:
+            for dep in blocked_jobs[0].dependent_jobs.all():
```
AlanCoding, line 453: I'm a little bit suspicious of this. I don't know that the `blocked_jobs` list is ordered... it might be, but that's really non-obvious looking at it, and the 0 reference, `blocked_jobs[0]` always feels suspicious. If this is well-established, then that's fine, but some code comments could help. What _is_ the first entry of the blocked jobs?

If the list is always length 1, then maybe you should just loop instead of using the first one?

AlanCoding: oh that's probably the job. This is making more sense to me. It will fail its job if it fails... and any other unified jobs that is waiting on.

awx/main/scheduler/task_manager.py
```diff
@@ -376,45 +374,71 @@ def should_update_related_project(self, job, latest_project_update):
...
+            else:
+                dependencies.append(latest_project_update)
+
+        # Inventory created 2 seconds behind job
+        try:
+            start_args = json.loads(decrypt_field(task, field_name="start_args"))
+        except ValueError:
+            start_args = dict()
+        # generator for inventory sources related to this task
+        task_inv_sources = (invsrc for invsrc in self.all_inventory_sources if invsrc.inventory == task.inventory)
```
AlanCoding: Especially when looping over a global jobs list like `self.all_inventory_sources`, I would like to stick with `*_id` fields for performance reasons (even if it may or may not make a difference). So here, I would make the conditional be `if invsrc.inventory_id == task.inventory_id`

AlanCoding: well I get that you're just moving the code

awx/main/scheduler/task_manager.py
```diff
@@ -572,6 +596,8 @@ def process_tasks(self, all_sorted_tasks):
         pending_tasks = [t for t in all_sorted_tasks if t.status == 'pending']
         undeped_tasks = [t for t in pending_tasks if not t.dependencies_processed]
         dependencies = self.generate_dependencies(undeped_tasks)
+        deps_of_deps = self.generate_dependencies(dependencies)
+        dependencies += deps_of_deps
```
AlanCoding, line 612: I remember you mentioning this, but now I get it. Basically, `deps_of_deps` can only be the project updates generated for inventory updates. Because you know it only goes this-many levels deep it's written like this and not iterated indefinitely.

AlanCoding (changes requested):
I have 2 requests remaining 
1. that we add some code comments to the effect that `blocked_jobs[0]` is just a back-reference to the job and
2. change related object references to the `*_id` field in a case, this will make future optimization work go more smoothly

after these, I'm happy to approve

awx/main/scheduler/task_manager.py
fosterseth: exactly

awx/main/models/mixins.py
fosterseth: yeah `blocked_jobs` are all the jobs that depend on this inventory update succeeding.

But, how can we fail the _other_ inventory updates that were started too (to retain the same behavior as devel)? Well, we can get that by looking at one of the blocked_jobs and getting its dependent_jobs. These dependent jobs might include project updates, so we need to filter it down further by checking its type

`if type(dep) is type(self) and dep.id != self.id:`

We can really grab _any_ of the blocked_jobs for this, as they all are guaranteed to have the same inventory, and thus the same set of inventory sources.

awx/main/scheduler/task_manager.py
fosterseth: made the change

awx/main/models/mixins.py
fosterseth: add some comments around this @AlanCoding

AlanCoding (approved):

awx/main/models/mixins.py
AlanCoding: sounds good.

Ultimately, looks like these feed into `handle_work_error` via the dispatcher errback. That part of the design isn't bad, it's pretty sensible to handle the triggers this way. On the other hand, `handle_work_error` looks even worse than the reaper in terms for changing job status from underneath a running process, and not even bothering to send a cancel signal. Problem for another day...

fosterseth:
my latest commit addresses an issue where the task manager can start a job even if a dependency has failed.
`job_blocked_by` checks if a dependency is in ACTIVE_STATES and if not, proceeds to run. Well that isn't good because the dependency could be a failed or error state.

The job would run and probably fail during execution because it detects that the last project update/inventory update has failed, stopping it in its tracks. But we can prevent the dispatcher from even getting that far by failing the task as soon as we know we should (in the task manager).

@AlanCoding please take a look at my most recent commit to see how I handled the above ^

https://github.com/ansible/awx/pull/12073/commits/b94b1ef430866a0764462067cee0b282210fa7d0

note: this is a current issue in devel as well

awx/main/scheduler/task_manager.py
```diff
@@ -79,10 +80,23 @@ def job_blocked_by(self, task):
...
+        for dep in task.dependent_jobs.all():
+            if dep.status in ACTIVE_STATES:
+                return dep
+            # if we detect a failed or error dependency, go ahead and fail this
+            # task. The errback on the dependency takes some time to trigger,
+            # and we don't want the task to enter running state if its
+            # dependency has failed or errored.
+            elif dep.status in ("error", "failed"):
+                task.status = 'failed'
+                task.job_explanation = 'Previous Task Failed: {"job_type": "%s", "job_name": "%s", "job_id": "%s"}' % (
```
AlanCoding, line 92: My only note is that I'd like this message to be _distinguishable_ from the similar message in the handle_work_error task - for ease of debugging. Any changing of the wording whatsoever would be sufficient to accomplish that.

Issue #12058, opened 2022-04-14
Clean up React Warnings


nixocio:
##### SUMMARY
Clean up React Warnings.

I noticed a couple of react warnings on the dev env.

- [x] Go to https://localhost:3001/#/inventories/inventory/1/sources/add

![image](https://user-images.githubusercontent.com/9053044/163439307-7e0597b0-df92-4ba5-9513-2861bd25d355.png)

![image](https://user-images.githubusercontent.com/9053044/163439489-302ac7fe-79e6-4814-9bbe-5fbe79d5623c.png)

kialam:
The first error should go away once we upgrade React to [v18](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#react). The other warnings should be fixed.

AlexSCorey:
I am going to break this into 2 issues.  1 for the invalid prop type and 1 for React upgrade. Reason: We have to wait for Patternfly to get to react 18 before we can update.

Issue #12011, opened 2022-04-05
Only fetch child summary when job type is job


akus062381:
This issue is a spin off from https://github.com/ansible/awx/pull/11944#pullrequestreview-932128826. 
Due to being so close to code freeze, we decided to make this a separate issue to handle later.

Let's avoid calling this whenever the job.type is not job. The API returns a 404 response from /jobs/:id/job_events/children_summary whenever it is a project_update or system_job job.

![image](https://user-images.githubusercontent.com/63685497/161842776-764d15dd-3b13-4801-97c8-b48bcbeefc9c.png)

keithjgrant:
Looks like Alex is addressing this in https://github.com/ansible/awx/pull/12078

PR #12078, opened 2022-04-20
Cleans up some console warnings.

AlexSCorey:
##### SUMMARY
Addresses https://github.com/ansible/awx/issues/12011 and  #12058 for the most part.  It does not address the:
image
because to do so would require an update to react 18, which we cannot do until Patternfly updates as well.  
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui/src/screens/Credential/CredentialList/CredentialList.js
+++ b/awx/ui/src/screens/Credential/CredentialList/CredentialList.js
@@ -49,9 +49,8 @@ function CredentialList() {
         CredentialsAPI.readOptions(),
       ]);
       const searchKeys = getSearchableKeys(credActions.data.actions?.GET);
-      const item = searchKeys.indexOf('type');
-      if (item) {
-        searchKeys[item] = 'credential_type__kind';
+      if (credActions.data.actions?.GET.type) {
+        searchKeys.push({ key: 'credential_type__kind', type: 'string' });
       }
       return {
         credentials: creds.data.results,
--- a/awx/ui/src/screens/Inventory/InventorySources/InventorySourceList.js
+++ b/awx/ui/src/screens/Inventory/InventorySources/InventorySourceList.js
@@ -14,6 +14,7 @@ import PaginatedTable, {
   ToolbarAddButton,
   ToolbarDeleteButton,
   ToolbarSyncSourceButton,
+  getSearchableKeys,
 } from 'components/PaginatedTable';
 import useSelected from 'hooks/useSelected';
 import DatalistToolbar from 'components/DataListToolbar';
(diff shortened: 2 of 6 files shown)


awx/ui/src/screens/Credential/CredentialList/CredentialList.js
```diff
@@ -49,9 +49,8 @@ function CredentialList() {
         CredentialsAPI.readOptions(),
       ]);
       const searchKeys = getSearchableKeys(credActions.data.actions?.GET);
-      const item = searchKeys.indexOf('type');
-      if (item) {
-        searchKeys[item] = 'credential_type__kind';
+      if (credActions.data.actions?.GET.type) {
```
AlexSCorey, line 52: When `getSearchableKeys` was introduced we lost the ability to search by credential kind in advanced search.  This reimplements that functionality.

awx/ui/src/screens/Inventory/InventorySources/InventorySourceList.js
```diff
@@ -57,9 +58,7 @@ function InventorySourceList() {
         sourceCount: results[0].data.count,
         sourceChoices: results[1].data.actions.GET.source.choices,
         sourceChoicesOptions: results[1].data.actions,
-        searchableKeys: Object.keys(results[1].data.actions?.GET || {}).filter(
-          (key) => results[1].data.actions?.GET[key].filterable
-        ),
+        searchableKeys: getSearchableKeys(results[1].data.actions?.GET),
```
AlexSCorey, line 61: This addresses the second item in #12058

awx/ui/src/screens/Job/JobOutput/useJobEvents.js
```diff
@@ -25,7 +25,12 @@ export const REBUILD_TREE = 'REBUILD_TREE';
 export const TOGGLE_COLLAPSE_ALL = 'TOGGLE_COLLAPSE_ALL';
 export const SET_CHILDREN_SUMMARY = 'SET_CHILDREN_SUMMARY';
 
-export default function useJobEvents(callbacks, jobId, isFlatMode) {
+export default function useJobEvents(
+  callbacks,
+  jobId,
+  hasChildrenSummary,
```
AlexSCorey: This addresses a failed API request for `children_summary` on job types that do not have that end point.  @keithjgrant

keithjgrant: `hasChildSummary` shouldn't be necessary, since it will always be in flat mode for job types that don't have child summaries

awx/ui/src/screens/Job/JobOutput/JobOutput.js
```diff
@@ -139,6 +139,7 @@ function JobOutput({ job, eventRelatedSearchableKeys, eventSearchableKeys }) {
       setJobTreeReady: () => setIsTreeReady(true),
     },
     job.id,
+    job.related?.children_summary,
```
keithjgrant: the child summary parameter shouldn't be necessary, since it will always be in flat mode for job types that don't have child summaries

keithjgrant:
If `isFlatMode` is true, `setJobTreeReady()` is never getting called, so it looks like the events are stuck in a loading state. You can fix this in useJobEvents.js, line 52:

```
    if (isFlatMode) {
      callbacks.setJobTreeReady();
      return;
    }
```

keithjgrant (approved):

PR #12086, opened 2022-04-21
Remove committed_capacity field, delete supporting code

AlanCoding:
##### SUMMARY
Concluding what was started with https://github.com/ansible/awx/pull/12041

This is now unblocked after the merge of https://github.com/ansible/awx/pull/12050

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/scheduler/task_manager_models.py
+++ b/awx/main/scheduler/task_manager_models.py
@@ -18,9 +18,19 @@ class TaskManagerInstance:
     def __init__(self, obj):
         self.obj = obj
         self.node_type = obj.node_type
-        self.remaining_capacity = obj.capacity
+        self.consumed_capacity = 0
         self.capacity = obj.capacity
         self.hostname = obj.hostname
+
+    def consume_capacity(self, impact):
+        self.consumed_capacity += impact
+
+    @property
+    def remaining_capacity(self):
+        remaining = self.capacity - self.consumed_capacity
+        if remaining < 0:
+            return 0
+        return remaining
 
 
 class TaskManagerInstances:
@@ -40,9 +50,9 @@ class TaskManagerInstances:
             control_instance = self.instances_by_hostname.get(task.controller_node, '')
             execution_instance = self.instances_by_hostname.get(task.execution_node, '')
             if execution_instance and execution_instance.node_type in ('hybrid', 'execution'):
-                self.instances_by_hostname[task.execution_node].remaining_capacity -= task.task_impact
+                self.instances_by_hostname[task.execution_node].consume_capacity(task.task_impact)
             if control_instance and control_instance.node_type in ('hybrid', 'control'):
-                self.instances_by_hostname[task.controller_node].remaining_capacity -= settings.AWX_CONTROL_NODE_TASK_IMPACT
+                self.instances_by_hostname[task.controller_node].consume_capacity(settings.AWX_CONTROL_NODE_TASK_IMPACT)
 
     def __getitem__(self, hostname):
         return self.instances_by_hostname.get(hostname)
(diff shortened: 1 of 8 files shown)


awx/main/scheduler/task_manager_models.py
```diff
@@ -18,10 +18,20 @@ class TaskManagerInstance:
...
+    def consume_capacity(self, impact):
+        self.consumed_capacity += impact
+
+    @property
+    def remaining_capacity(self):
+        remaining = self.capacity - self.consumed_capacity
+        if remaining < 0:
+            return 0
+        return remaining
+
```
kdelee, line 34: :tada:

awx/main/scheduler/task_manager_models.py
```diff
@@ -70,6 +82,14 @@ def __init__(self, instances_by_hostname=None, instance_groups=None):
...
                 )
 
+    def get_remaining_capacity(self, group_name):
+        instances = self.instance_groups[group_name]['instances']
+        return sum(inst.remaining_capacity for inst in instances)
+
+    def get_consumed_capacity(self, group_name):
+        instances = self.instance_groups[group_name]['instances']
+        return sum(inst.consumed_capacity for inst in instances)
+
```
kdelee, line 92: :100:

kdelee (approved):
Great! Excellent continuation. More lines deleted than added! :1st_place_medal:

awx/main/scheduler/task_manager_models.py
```diff
@@ -18,10 +18,20 @@ class TaskManagerInstance:
...
         self.hostname = obj.hostname
 
+    def consume_capacity(self, impact):
+        self.consumed_capacity += impact
+
+    @property
+    def remaining_capacity(self):
+        remaining = self.capacity - self.consumed_capacity
+        if remaining < 0:
+            return 0
```
john-westcott-iv, line 32: Do we not want to return a negative number here to show that the instance is over capacity?

awx/main/scheduler/task_manager_models.py
```diff
@@ -54,14 +64,16 @@ def __contains__(self, hostname):
 class TaskManagerInstanceGroups:
     """A class representing minimal data the task manager needs to represent an InstanceGroup."""
 
-    def __init__(self, instances_by_hostname=None, instance_groups=None):
+    def __init__(self, instances_by_hostname=None, instance_groups=None, ig_qs=None):
```
john-westcott-iv: As someone unfamiliar with this part of the code I'm not really sure what `ig_qs` is representing. The `ig` part I'm assuming is instance group but what is the `qs` part, queue_size? Can we expand that part of the variable to be words to be clearer?

AlanCoding: queryset of instance groups. I guess maybe `instance_groups_queryset`?

awx/main/scheduler/task_manager_models.py
AlanCoding: This is shared between the task manager and the API. From the API's perspective, `remaining_capacity` should probably get set to 0 when over-capacity for ease of understanding. From the perspective of the task manager's logic, the two "selection" methods have to work correctly, and I'm not worried about that, it handles 0 remaining capacity fine, I don't think there's any case where it would send work to a node that has 0 remaining capacity.

awx/main/scheduler/task_manager_models.py
john-westcott-iv: That'd be great, thanks!

john-westcott-iv (approved):

PR #12089, opened 2022-04-22
Consume job_explanation we sometimes set in ansible-runner

shanemcd:
Replaces / updates https://github.com/ansible/awx/pull/11376

### DIFF
--- a/awx/main/tasks/jobs.py
+++ b/awx/main/tasks/jobs.py
@@ -546,6 +546,11 @@ class BaseTask(object):
             status = res.status
             rc = res.rc
 
+            # We call this to get the current values from the database, in case update_model was called
+            # within the threadpools inside of AWXReceptorJob. We use update_model instead of
+            # refresh_from_db here because it contains retry logic that is resilient to database failures.
+            self.instance = self.update_model(self.instance.pk)
+
             if status in ('timeout', 'error'):
                 job_explanation = f"Job terminated due to {status}"
                 self.instance.job_explanation = self.instance.job_explanation or job_explanation
@@ -580,7 +585,6 @@ class BaseTask(object):
         if 'got an unexpected keyword argument' in extra_update_fields.get('result_traceback', ''):
             extra_update_fields['result_traceback'] = "{}\n\n{}".format(extra_update_fields['result_traceback'], ANSIBLE_RUNNER_NEEDS_UPDATE_MESSAGE)
 
-        self.instance = self.update_model(pk)
         self.instance = self.update_model(pk, status=status, emitted_events=self.runner_callback.event_ct, **extra_update_fields)
 
         try:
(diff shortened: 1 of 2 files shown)


AlanCoding:
Regarding the part of this that I wrote, what data would this get us?

https://github.com/ansible/ansible-runner/blob/d01304455c3c25669fba70d9983487ffdb85f1c5/ansible_runner/streaming.py#L101

> Failed to JSON parse a line from transmit stream

> Failed to extract private data directory on worker

These are pretty vague messages. If I got them by themselves, I wouldn't know what to do. I'm not against surfacing them to the user, indeed, I think that's a very good idea. However, this save will overwrite any `job_explanation` that was set in another part of the code or by another process. This case, in particular, is non-exclusive with getting additional error details. The most highly valued error details tends to get what comes from `AWXReceptorJob`, reading details from the work unit, or the work unit output.

Because of that, I have cold feet about doing this without stacking it on top of https://github.com/ansible/awx/pull/11832, which fixed tests that were actively failing (we just forget because they get set to skip). If we can get this in a way that stacks on top of any other error details, then that's the ideal, that would be perfect.

shanemcd:
Right now jobs fail with an even more vague explanation: `Job terminated due to error`.

This at least tells _us_ the last thing that went wrong.

awx/main/tasks/jobs.py
```diff
@@ -566,6 +566,11 @@ def run(self, pk, **kwargs):
             status = res.status
             rc = res.rc
 
+            # We call this to get the current values from the database, in case update_model was called
+            # within the threadpools inside of AWXReceptorJob. We use update_model instead of
+            # refresh_from_db here because it contains retry logic that is resilient to database failures.
+            self.instance = self.update_model(self.instance.pk)
```
AlanCoding, line 552: I think this should replace the `self.instance = self.update_model(pk)` on line 599 / 604. You can just read the code, and know for absolute sure that does nothing helpful. It's called right before another `update_model` call, and the first thing _that_ does is get the object again. Swapping this for that will leave it more functional without adding more noise.

jbradberry: I think I agree.

AlanCoding (approved):

jbradberry (approved):

awx/main/tasks/jobs.py
```diff
@@ -546,6 +546,11 @@ def run(self, pk, **kwargs):
...
```
AlanCoding, line 552: Ping @sarabrajsingh, as he and I discussed error logs that came from `post_run_hook` due to not doing this.

Summarizing here...

Each thread has its own connection object. When connections are dropped, that object is stale. If it's not hitting one particular corner case, it will correct itself after 1 failed query. So post_run_hook errors, but then code after it starts to work again, so it's not easy to observe unless you check the logs.

AlanCoding:
I've modified adjacent code, so here's what I think a rebase should look like:

https://github.com/ansible/awx/compare/devel...AlanCoding:capture-runner-error?expand=1

EDIT: to resolve conflicts.

AlanCoding:
A replacement PR exists

Issue #11698, opened 2022-02-08
[tech debt] Overly-aggressive unified job save method errors when another process changes model


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
This traceback paints a very complete picture of an error case that we should address.

```
2022-02-08 07:55:56,634 ERROR    [dc755b9d534d46fe80d2413561ce1af4] awx.main.tasks.jobs job 3758 (running) Exception occurred while running task
Traceback (most recent call last):
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/tasks/receptor.py", line 401, in _run_internal
    self.task.instance.save(update_fields=['result_traceback'])
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/models/unified_jobs.py", line 879, in save
    self._update_parent_instance()
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/models/unified_jobs.py", line 801, in _update_parent_instance
    parent_instance.save(update_fields=update_fields)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/models/jobs.py", line 330, in save
    return super(JobTemplate, self).save(*args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/models/unified_jobs.py", line 280, in save
    super(UnifiedJobTemplate, self).save(*args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/polymorphic/models.py", line 93, in save
    return super(PolymorphicModel, self).save(*args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/models/base.py", line 329, in save
    super(PrimordialModel, self).save(*args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/models/base.py", line 176, in save
    super(CreatedModifiedModel, self).save(*args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/models/mixins.py", line 566, in save
    super().save(*args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/db/models/base.py", line 743, in save
    self.save_base(using=using, force_insert=force_insert,
[...]

PR #12090, opened 2022-04-22
Avoid parent instance update when status was unchanged

AlanCoding:
##### SUMMARY
Connect https://github.com/ansible/awx/issues/11698

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/models/unified_jobs.py
+++ b/awx/main/models/unified_jobs.py
@@ -813,7 +813,8 @@ class UnifiedJob(
 
         # If this job already exists in the database, retrieve a copy of
         # the job in its prior state.
-        if self.pk:
+        # If update_fields are given without status, then that indicates no change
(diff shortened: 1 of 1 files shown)


AlanCoding:
@fosterseth @kdelee as much as we're looking into deep updates to the task manager, I think we need to clear out some really basic sanity fixes like this.

We have plenty of times we do `.save(update_fields=["something"])` in the task manager, and this triggers completely unrelated queries. For the task manager, most of those are updating status, and so will necessarily have to do this other code, so it won't make a world of difference.

kdelee (approved):

fosterseth (approved):

AlanCoding:
:crossed_fingers: this should be a fairly substantial performance gain, similar to the activity stream optimization we merged, I wouldn't even be surprised if this can crawl back some more performance on `.create_unified_job` actions.

Issue #12057, opened 2022-04-14
Have job_wait module set a good example and do exponetial backoff


kdelee:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

AWX is sensitive to DDOS attacks by clients polling too aggressively. We should set a good example in our module and implement longer interval/some kind of backoff scheme.

See https://github.com/ansible/awx/blob/devel/awx_collection/plugins/modules/job_wait.py

sean-m-sullivan:
The default value for wait/interval were set to timeout=30, interval=1 effectivly
I've made an update to double it, if we want to adopt this. Should help, but keep things quick. 
@kdelee Thoughts?

kdelee:
@sean-m-sullivan that's a good start, though I still think its too much if people are using this at a larger scale.
As an alternative to a backoff scheme, I had it suggested to me that folks should subscribe via websocket for updates. awxkit already has some hooks for subscribing to websockets https://github.com/ansible/awx/blob/677187a43edcac82ee6f1af92386fca7005caa1e/awxkit/awxkit/ws.py#L28-L50

that may be a good pattern to try out

sean-m-sullivan:
Was going to say, the Redhat_cop collection recently started using asyncs with 1 second delays on most items, and have not run into issues, but it does retry every 1 second. The websocket might be a way to solve both

PR #12094, opened 2022-04-22
update awx collection wait interval to 2

sean-m-sullivan:
##### SUMMARY
update awx collection wait interval to 2
As per #12057 we should set a better example for modules by defaulting to a higher timeout.
A higher timeout and users will likely start to complain about speed, but this doubles the default, and helps prevent DDOS of the controller. 

This also adds uniformity across the wait intervals as the python methods were set to 10, and the modules set to 1.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - Collection

##### AWX VERSION
```
20.1.0
```

### DIFF
--- a/awx_collection/plugins/module_utils/controller_api.py
+++ b/awx_collection/plugins/module_utils/controller_api.py
@@ -994,7 +994,7 @@ class ControllerAPIModule(ControllerModule):
         else:
             return True
 
-    def wait_on_url(self, url, object_name, object_type, timeout=30, interval=10):
+    def wait_on_url(self, url, object_name, object_type, timeout=30, interval=2):
         # Grab our start time to compare against for the timeout
         start = time.time()
         result = self.get_endpoint(url)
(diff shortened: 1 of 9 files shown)


kdelee (approved):
good start

sean-m-sullivan:
@kdelee @fosterseth Do we want to put this into implementation, or close it and wait on the websockets update. I'm fine closing this if we don't want to do this as stopgap. just thinking to clean up my own and the AWX Open PR list :)

fosterseth (approved):

fosterseth:
definitely in favor of this change

john-westcott-iv:
I'm wondering if this is important enough to say something about it in the collections README.md? Like maybe under the `Release and Upgrade Notes` section. 
Any thoughts?

sean-m-sullivan:
It might be, I did put a note for the changelog though
changelog-entry
---
msg: "update awx collection wait interval to 2"
-->

PR #12098, opened 2022-04-25
Fix broken job WFJT details when related JT is deleted

nixocio:
Fix broken job WFJT details when related JT is deleted


AlexSCorey (approved):
Looks good

Issue #11422, opened 2021-12-06
Notifications are not sent


apollo13:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

Currently notifications for (some) jobs are not sent due to a race condition (?):
```
awx_1    | 2021-12-06 13:39:14,446 WARNING  [a1a65966262c4559afd398df32b5f445] awx.main.dispatch scaling down worker pid:14114
awx_1    | 2021-12-06 13:39:14,448 DEBUG    [a1a65966262c4559afd398df32b5f445] awx.main.dispatch task 4648c5ba-e5bb-49c4-8040-0e98e1d4975b starting awx.main.tasks.handle_success_and_failure_notifications(*[4021])
awx_1    | 2021-12-06 13:39:14,450 WARNING  [a1a65966262c4559afd398df32b5f445] awx.main.dispatch scaling down worker pid:14116
awx_1    | 2021-12-06 13:39:14,451 WARNING  [a1a65966262c4559afd398df32b5f445] awx.main.dispatch worker exiting gracefully pid:14114
awx_1    | 2021-12-06 13:39:14,453 WARNING  [a1a65966262c4559afd398df32b5f445] awx.main.dispatch worker exiting gracefully pid:14116
awx_1    | 2021-12-06 13:39:14,813 INFO     [949fbc302a95418e85b3b941b7bb551d] awx.main.commands.run_callback_receiver Event processing is finished for Job 4021, sending notifications
awx_1    | 2021-12-06 13:39:14,813 INFO     [949fbc302a95418e85b3b941b7bb551d] awx.main.commands.run_callback_receiver Event processing is finished for Job 4021, sending notifications
awx_1    | INFO 2021/12/06 13:39:15 Stdout complete - closing channel for: aCroB8Bf 
awx_1    | ERROR 2021/12/06 13:39:15 Read error in control service: read unix /var/run/awx-receptor/receptor.sock->@: use of closed network connection
awx_1    | INFO 2021/12/06 13:39:15 Client disconnected from control service @
awx_1    | ERROR 2021/12/06 13:39:15 Error closing connection: close unix /var/run/awx-receptor/receptor.sock->@: use of closed network connection
[...]

PR #12110, opened 2022-04-26
Fix notification timing issue by sending in the latter of 2 events

### LINKED ISSUE #8771: Job done isn't always correct

AlanCoding:
##### SUMMARY
Connect https://github.com/ansible/awx/issues/11422 as the primary issue for this.

As a secondary issue, I would suggest we close https://github.com/ansible/awx/issues/8771 if we merge this. This PR adds the basic conflict management between event processing and status change. This does not attempt to assure that all events are processed, only that the "stats" event in processed, which is the "playbook_on_stats" event for playbooks, and EOF for other types.

See prior work at https://github.com/ansible/awx/compare/devel...chrismeyersfsu:job_done_done?expand=1 --> this tried to tackle the problem of a "job done" hook in conjunction with making the events queue durable. I'm not attacking the problem of event queue durability here.

I first implemented this fix in a different way here --> https://github.com/ansible/awx/compare/devel...AlanCoding:send_notifications?expand=1, which added a `notifications_processed` field, but that was probably poorly named. You could do what I'm doing here with a boolean like `stats_event_processed`, but that feels like it has no connection to other use cases, and would be useless for any other than this very narrow problem.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
WIP upon opening because:
 - unit tests need to be updated
 - integration tests still need to be verified
 - I removed tracking of the stats event dispatch in the callback class, which I think is still needed to handle error events, and I'll probably add that back in and deal with that problem


awx/main/dispatch/worker/callback.py
```diff
@@ -26,6 +26,15 @@
 logger = logging.getLogger('awx.main.commands.run_callback_receiver')
 
 
+EVENT_MAP = {
+    'job_id': JobEvent,
+    'ad_hoc_command_id': AdHocCommandEvent,
+    'project_update_id': ProjectUpdateEvent,
+    'inventory_update_id': InventoryUpdateEvent,
+    'system_job_id': SystemJobEvent,
+}
```
john-westcott-iv: If you inverted this map could you get rid of your new for loop on 185-189?

awx/main/dispatch/worker/callback.py
```diff
@@ -165,47 +207,33 @@ def perform_work(self, body):
...
                 job_identifier = 'unknown job'
-                for key, cls in event_map.items():
+                for key, cls in EVENT_MAP.items():
                     if key in body:
                         job_identifier = body[key]
                         break
 
                 self.last_event = f'\n\t- {cls.__name__} for #{job_identifier} ({body.get("event", "")} {body.get("uuid", "")})'  # noqa
 
+                trigger_event_type = 'playbook_on_stats' if cls in (JobEvent, ProjectUpdateEvent, AdHocCommandEvent) else 'EOF'
```
john-westcott-iv: (assuming you can invert EVENT_MAP from previous comment) 
What about changing event map to include a field for this and then just doing:
`trigger_event_map = EVENT_MAP[cls]['event_type']`

nixocio (approved):
Approve the UI changes.

awx/main/dispatch/worker/callback.py
AlanCoding: I can dig that. Still need to consider the original loop that used `event_map` - if we made the values a dictionary, then that could look like

```python
                for cls, props in EVENT_MAP.items():
                    if props['job_ref'] in body:
                        job_identifier = body[props['job_ref']]
                        break
```

There's a little bit of indirection going on there, but this data structure could be reused more broadly than what I have now. I would probably move it into another module.

AlanCoding:
All test results are looking good for this, so I'm pretty serious about it at this point.

AlanCoding:
All test results are looking good for this, so I'm pretty serious about it at this point.

awx/main/dispatch/worker/callback.py
AlanCoding: Pushed a commit addressing this.

AlanCoding: But I still missed this exact line, pushed another commit which should wrap it up.

awx/main/tasks/system.py
```diff
@@ -716,25 +716,6 @@ def handle_work_error(task_id, *args, **kwargs):
...
-    while retries < settings.AWX_NOTIFICATION_JOB_FINISH_MAX_RETRY:
-        if uj.finished:
-            uj.send_notification_templates('succeeded' if uj.status == 'successful' else 'failed')
-            return
-        else:
-            # wait a few seconds to avoid a race where the
-            # events are persisted _before_ the UJ.status
-            # changes from running -> successful
-            retries += 1
-            time.sleep(1)
```
shanemcd, line 732: 🔥

AlanCoding:
Tried to write up some details if someone wants to read. https://gist.github.com/AlanCoding/f7ae3d10ef04692c254c17507ef16b59

awx/main/dispatch/worker/callback.py
```diff
@@ -113,6 +113,32 @@ def work_loop(self, *args, **kw):
             signal.signal(signal.SIGUSR1, self.toggle_profiling)
         return super(CallbackBrokerWorker, self).work_loop(*args, **kw)
 
+    @staticmethod
```
shanemcd: I only see you calling this as a method on `self`, can we drop this decorator?

fosterseth (approved):

jbradberry (approved):

awx/main/dispatch/worker/callback.py
AlanCoding: Pushed a commit to try and address this comment. I get what you're saying, I was calling `self.job_stats_wrapup` which isn't great.

PR #12114, opened 2022-04-26
update awx collection workflow module

sean-m-sullivan:
##### SUMMARY
I was going through making documentation about workflow schemas. The awxkit export and api uses the term workflow_nodes, however the module uses schema. Back when first created this was in line with the old workflow job template module, however we should probably update it to use the same thing as used in the export and API.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - Collection

##### AWX VERSION

```
21.0.0
```

### DIFF
--- a/awx_collection/tests/integration/targets/workflow_job_template/tasks/main.yml
+++ b/awx_collection/tests/integration/targets/workflow_job_template/tasks/main.yml
@@ -233,8 +233,7 @@
     that:
       - "result is changed"
 
-# Node actions do what this schema command used to do
-# schema: [{"success": [{"job_template": "{{ jt1_name }}"}], "job_template": "{{ jt2_name }}"}]
+# Node actions do what the schema command used to do
 - name: Create leaf node
   workflow_job_template_node:
     identifier: leaf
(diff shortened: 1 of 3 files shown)


awx_collection/tests/integration/targets/workflow_job_template/tasks/main.yml
```diff
@@ -233,8 +233,8 @@
     that:
       - "result is changed"
 
-# Node actions do what this schema command used to do
-# schema: [{"success": [{"job_template": "{{ jt1_name }}"}], "job_template": "{{ jt2_name }}"}]
+# Node actions do what this workfow_nodes command used to do
```
fosterseth: minor spelling error on this line and next

sean-m-sullivan: Updated, decided to leave the old text for the one line, and remove the schema line as I believe that is so old that no one remembers what its trying to say.

fosterseth (approved):

fosterseth:
a nice change

AlanCoding (approved):

Issue #10838, opened 2021-08-09
"Cleanup Job Schedule" fails with external DB, schema other than public


tobias-oe:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

When you use an external postgres database for AWX and don't use the default schema "public", then the cleanup job "Cleanup Job Schedule" fails, because there is a fixed call for the table "public.main_jobevent".
This table cannot exist, if you use another schema e.g. "awx".

Please honor the environment variables passed by deployment scripts, so you don't run in errors if using not the default schema name.

### AWX version

19.2.2

### Installation method

kubernetes

### Modifications

no

### Steps to reproduce

You need an installation of AWX configured with an external postgres DB using a schema not named "public".

1. Run the scheduled job "Cleanup Job Schedule"
2. Check the output of this job

### Expected results

The cleanup job finishes successfully.

### Actual results

The cleanup job fails with the following exception:

	Traceback (most recent call last):
	  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/db/backends/utils.py", line 82, in _execute
		return self.cursor.execute(sql)
	psycopg2.errors.UndefinedTable: relation "public.main_jobevent" does not exist
	LINE 1: ...ild FROM pg_catalog.pg_inherits WHERE inhparent = 'public.ma...
																 ^
	The above exception was the direct cause of the following exception:
	Traceback (most recent call last):
	  File "/usr/bin/awx-manage", line 8, in <module>
		sys.exit(manage())
	  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/__init__.py", line 171, in manage
		execute_from_command_line(sys.argv)
	  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
		utility.execute()
	  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/__init__.py", line 375, in execute
		self.fetch_command(subcommand).run_from_argv(self.argv)
	  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/django/core/management/base.py", line 323, in run_from_argv
[...]

PR #12119, opened 2022-04-27
Remove hardcoded public schema in cleanup_jobs.py

kimbernator:
##### SUMMARY

related https://github.com/ansible/awx/issues/10838

This removes the hardcoded reference to the public schema when looking for table partitions to clean up. I'm -pretty- sure this works fine, and I have successfully tested it on my AWX instance which uses an external DB with a  schema named 'awx' instead of 'public'.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - Management Job

##### AWX VERSION

```
awx: 19.3.1.dev1076+g3d5f302d10
```

##### ADDITIONAL INFORMATION

```

```


jbradberry (approved):
I personally think that this makes sense.

shanemcd (approved):

PR #12122, opened 2022-04-27
Fixing write location of ssh_key-data-cert.pub

john-westcott-iv:
##### SUMMARY

Fixing regression from [#12083](https://github.com/ansible/awx/pull/12083) where the ssh_key-data-cert.pub files were not properly being written into the artificats/<job id> folder. 

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API
 
##### AWX VERSION

```
awx: 20.1.1.dev103+ga9639c4831
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/tasks/jobs.py
+++ b/awx/main/tasks/jobs.py
@@ -119,12 +119,11 @@ class BaseTask(object):
     def update_model(self, pk, _attempt=0, **updates):
         return update_model(self.model, pk, _attempt=0, _max_attempts=self.update_attempts, **updates)
 
-    def write_private_data_file(self, private_data_dir, file_name, data, sub_dir=None, permissions=0o600):
+    def write_private_data_file(self, private_data_dir, file_name, data, sub_dir=None, file_permissions=0o600):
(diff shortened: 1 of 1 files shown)


AlanCoding:
https://github.com/ansible/awx/blob/16e2bfa89e93f9b65d084074117ba76e57d081a0/awx/main/tasks/jobs.py#L127

Does this not need to be `os.makedirs`? Or a loop or something..

john-westcott-iv:
> https://github.com/ansible/awx/blob/16e2bfa89e93f9b65d084074117ba76e57d081a0/awx/main/tasks/jobs.py#L127
> 
> Does this not need to be `os.makedirs`? Or a loop or something..

I think the previous assumption was that "something else" was creating the directories. But I like this idea, especially with a sub_dir option that can be multiple directories.

AlanCoding (approved):

InnocentK (approved):
From my testing this seems to have resolved the issue. Great work!

PR #12135, opened 2022-05-02
Avoid running jobs that have already been reaped

AlanCoding:
##### SUMMARY
We may reap jobs in the waiting status:

https://github.com/ansible/awx/blob/b6362a63ccc8338b48860cf710be7739c84f6fc5/awx/main/dispatch/reaper.py#L45-L51

It is extremely likely that a job reaped by the second condition there will still be picked up by the dispatcher and will start to run. That's not good.

We should still consider modifying that reaping condition, but this will avoid the self-contradictory situation generally. If a job was moved out of active status by another process, we assume that was done for a reason.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/tasks/jobs.py
+++ b/awx/main/tasks/jobs.py
@@ -397,6 +397,12 @@ class BaseTask(object):
         Run the job/task and capture its output.
         """
         self.instance = self.model.objects.get(pk=pk)
+        if self.instance.status != 'canceled' and self.instance.cancel_flag:
+            self.instance = self.update_model(self.instance.pk, status='canceled')
+        if self.instance.status not in ACTIVE_STATES:
+            # Prevent starting the job if it has been reaped or handled by another process.
+            logger.warning(f'Not starting {self.instance.status} task pk={pk}')
+            return
 
         if self.instance.execution_environment_id is None:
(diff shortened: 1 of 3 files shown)


awx/main/tasks/jobs.py
```diff
@@ -400,6 +400,11 @@ def run(self, pk, **kwargs):
         Run the job/task and capture its output.
         """
         self.instance = self.model.objects.get(pk=pk)
+        if self.instance.status != 'canceled' and self.instance.cancel_flag:
+            self.instance = self.update_model(self.instance.pk, start_args='', status='canceled')
+        if self.instance.status not in ACTIVE_STATES:
+            # Prevent starting the job if it has been reaped or handled by another process.
+            raise RuntimeError(f'Not starting {self.instance.status} task pk={pk}')
```
kdelee: maybe include 
```suggestion
            raise RuntimeError(f'Not starting {self.instance.status} task pk={pk} because {self.instance.status} is not in a valid active state')
```

kdelee (approved):

AlanCoding:
Absorbing into the "waiting" reaper issue, because there is a hazard with merging this on its own.

PR #12145, opened 2022-05-02
Add SCM Type detail to job detail view

### LINKED ISSUE #4057: Project update detail page in UI could use some more properties
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Here is a project with `scm_type` of hg:

Screen Shot 2019-06-12 at 1 11 57 PM

It would be nice if the side panel would show at least the scm type, and preferably the label from the options:

```json
                "choices": [
                    [
                        "",
                        "Manual"
                    ],
                    [
                        "git",
                        "Git"
                    ],
                    [
                        "hg",
                        "Mercurial"
                    ],
                    [
                        "svn",
                        "Subversion"
                    ],
                    [
                        "insights",
                        "Red Hat Insights"
                    ]
                ]
```

marshmalien:
##### SUMMARY
Resolves https://github.com/ansible/awx/issues/4057
Screen Shot 2022-05-02 at 4 00 02 PM

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - UI

### DIFF
--- a/awx/ui/src/screens/Job/JobDetail/JobDetail.js
+++ b/awx/ui/src/screens/Job/JobDetail/JobDetail.js
@@ -75,6 +75,14 @@ function JobDetail({ job, inventorySourceLabels }) {
     workflow_job: t`Workflow Job`,
   };
 
+  const scmTypes = {
+    '': t`Manual`,
+    git: t`Git`,
+    svn: t`Subversion`,
+    insights: t`Red Hat Insights`,
+    archive: t`Remote Archive`,
+  };
+
   const deleteJob = async () => {
     try {
       await getJobModel(job.type).destroy(job.id);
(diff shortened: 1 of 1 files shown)


awx/ui/src/screens/Job/JobDetail/JobDetail.js
```diff
@@ -241,6 +249,11 @@ function JobDetail({ job, inventorySourceLabels }) {
           label={t`Job Type`}
           value={jobTypes[job.type]}
         />
+        <Detail
+          dataCy="source-control-type"
+          label={t`Source Control Type`}
+          value={scmTypes[job.scm_type]}
```
keithjgrant, line 255: does this need to be `scmTypes[job.scm_type] || scmTypes['']`, or will `job.scm_type` always be an empty string if it's not one of the four specific types?

keithjgrant (approved):

awx/ui/src/screens/Job/JobDetail/JobDetail.js
marshmalien: I don't think so. Not every `job` will have a `scm_type`, so in those cases we wouldn't want to display the Source Control Type detail.

Issue #11862, opened 2022-03-08
Add help text popovers to /#/credentials details fields 


mabashian:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Spawned out of https://github.com/ansible/awx/issues/9215

See /#/settings for how this should work. If you look the details view for a particular settings group you'll see help text popovers next to many (if not all) of the fields. The help text matches up with the fields in the form. We want to bring all the forms/details to parity in this regard. If you see form fields that are missing help text, feel free to add it.

I'm not going to list out all the possible fields across all the different credential types.  Given that custom credential types are a thing, we'll need to do this dynamically.

PR #12158, opened 2022-05-03
Add help text popovers to /#/credentials details fields

nixocio:
Add help text popovers to /#/credentials details fields

See: https://github.com/ansible/awx/issues/11862

For the credentials we get most of the strings from the backend. I needed to surface this data on the details page.

I created one credential for each type available on the API and any of them that displayed a text help_text during add/edit now display the same help_text on the details page.

To show a few:

image

image

image

image

image

image

### DIFF
--- a/awx/ui/src/screens/Credential/CredentialDetail/CredentialDetail.js
+++ b/awx/ui/src/screens/Credential/CredentialDetail/CredentialDetail.js
@@ -100,12 +100,19 @@ function CredentialDetail({ credential }) {
 
   const { error, dismissError } = useDismissableError(deleteError);
 
-  const renderDetail = ({ id, label, type, ask_at_runtime }) => {
+  const renderDetail = ({
+    id,
+    label,
+    type,
+    ask_at_runtime,
+    help_text = '',
+  }) => {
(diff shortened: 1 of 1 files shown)


kialam:
Looks good so far - do we handle cases where `help_text` is `null` or `undefined`?

nixocio:
> Looks good so far - do we handle cases where `help_text` is `null` or `undefined`?

I added a default value to take of null or undefined. And `Detail` component just `help_text` it there is one.

awx/ui/src/screens/Credential/CredentialDetail/CredentialDetail.js
```diff
@@ -100,12 +100,19 @@ function CredentialDetail({ credential }) {
 
   const { error, dismissError } = useDismissableError(deleteError);
 
-  const renderDetail = ({ id, label, type, ask_at_runtime }) => {
+  const renderDetail = ({
+    id,
+    label,
+    type,
+    ask_at_runtime,
+    help_text = '',
```
kialam, line 108: 👍

kialam (approved):

PR #12161, opened 2022-05-03
Adding popover for details is showing breaking of words

nixocio:
Now that we are adding popovers for details pages, I noticed a couple of
strings wrapping in odd places, update css to avoid that.

Also `word-break: break-word` was deprecated.

Without this change

image

With this change

image


marshmalien (approved):
👍 IMO this change improves readability. Nice!

PR #12175, opened 2022-05-05
change SELinux label for EE volume mount

TheRealHaoLiu:

##### SUMMARY

- The `z` option indicates that the bind mount content is shared among multiple containers.
- The `Z` option indicates that the bind mount content is private and unshared.

If multiple container attempt to mount the same directory `Z` option will cause a raise condition where only the last container started will have access to the file.

Ref: https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.0.1.dev51+gd52d2af4b4
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/tasks/jobs.py
+++ b/awx/main/tasks/jobs.py
@@ -1451,8 +1451,8 @@ class RunProjectUpdate(BaseTask):
         params.setdefault('container_volume_mounts', [])
         params['container_volume_mounts'].extend(
             [
-                f"{project_path}:{project_path}:Z",
-                f"{cache_path}:{cache_path}:Z",
+                f"{project_path}:{project_path}:z",
+                f"{cache_path}:{cache_path}:z",
             ]
         )
         return params


awx/main/tasks/jobs.py
```diff
@@ -1451,8 +1451,8 @@ def build_execution_environment_params(self, instance, private_data_dir):
         params.setdefault('container_volume_mounts', [])
         params['container_volume_mounts'].extend(
             [
-                f"{project_path}:{project_path}:Z",
-                f"{cache_path}:{cache_path}:Z",
+                f"{project_path}:{project_path}:z",
```
Zokormazo, line 1454: Can we make this a setting? I would love to keep `:Z` in those platforms that work with `:Z`

shanemcd: @Zokormazo I think this is ok. It's not about the platform... this becomes a problem when we try to containerized the control plane services. `:Z` is the same as `:z` except it only allows 1 container to be accessing a directory.

TheRealHaoLiu: `Z` may cause unintended race condition?

if multiple container try to mount the same dir/file only the last container will retain access to the dir/file and rest of the container will get permission deny when try to read from and write to the file

TheRealHaoLiu: @Zokormazo are u ok with the change?

shanemcd (approved):

awx/main/tasks/jobs.py
shanemcd: Followed up with Julen in Slack to resolve this.

PR #12178, opened 2022-05-05
feat: add token description column

Tioborto:
##### SUMMARY

- Add "description" column to user token list page
- Correction on french vocabulary

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui/src/components/PaginatedTable/HeaderRow.js
+++ b/awx/ui/src/components/PaginatedTable/HeaderRow.js
@@ -7,6 +7,7 @@ import { parseQueryString, updateQueryString } from 'util/qs';
 
 const Th = styled(PFTh)`
   --pf-c-table--cell--Overflow: initial;
+  --pf-c-table--cell--MaxWidth: none;
 `;
 
 export default function HeaderRow({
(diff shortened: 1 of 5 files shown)


Tioborto:
Hello, 
I propose to you to add a new column in user token list page. 
I've just a problem with array headers style. How to correct it ? 
image

akus062381 (approved):

nixocio (changes requested):
> Hello, I propose to you to add a new column in user token list page. I've just a problem with array headers style. How to correct it ? image

@Tioborto, Thanks for your PR.

On the `HeaderRow.js` - Modify

```
const Th = styled(PFTh)`
  --pf-c-table--cell--Overflow: initial;
  --pf-c-table--cell--MaxWidth: none;
`;
```

I believe that should fix the header overlapping that you described.

Tioborto:
Indeed, thank you @nixocio. It works and fix header row style

nixocio:
hey @Tioborto, thanks for updating that. Would you mind adding a different PR to address the translations? In case we need to revert, etc.

nixocio (approved):

PR #12198, opened 2022-05-09
Prevents the api label from being added to UI only PRs

AlexSCorey:
##### SUMMARY
This change prevents the API label from being applied to issues that are ui only

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
-github workflows

##### AWX VERSION


shanemcd (approved):
Nice!

PR #12213, opened 2022-05-11
Add details related workflow job on the workflow approval details

nixocio:
Add details related workflow job on the work flow approval details

Remove not used prop isLoading, fix, and expand unit-tests related to
workflow approval details.

This UX was created with @trahman73 feedback.

image

### DIFF
--- a/awx/ui/src/components/AdHocCommands/AdHocCommandsWizard.test.js
+++ b/awx/ui/src/components/AdHocCommands/AdHocCommandsWizard.test.js
@@ -368,7 +368,7 @@ describe('<AdHocCommandsWizard/>', () => {
         response: {
           config: {
             method: 'get',
-            url: '/api/v2/credentals',
+            url: '/api/v2/credentials',
           },
           data: 'An error occurred',
           status: 403,
--- a/awx/ui/src/screens/WorkflowApproval/WorkflowApprovalDetail/WorkflowApprovalDetail.js
+++ b/awx/ui/src/screens/WorkflowApproval/WorkflowApprovalDetail/WorkflowApprovalDetail.js
@@ -1,14 +1,28 @@
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 
 import { t } from '@lingui/macro';
 import { Link, useHistory, useParams } from 'react-router-dom';
+import styled from 'styled-components';
+import {
+  Divider as PFDivider,
+  Title as PFTitle,
+  Chip,
+} from '@patternfly/react-core';
 import AlertModal from 'components/AlertModal';
-import { CardBody, CardActionsRow } from 'components/Card';
+import ChipGroup from 'components/ChipGroup';
+import ContentError from 'components/ContentError';
+import ContentLoading from 'components/ContentLoading';
 import DeleteButton from 'components/DeleteButton';
-import { Detail, DetailList, UserDateDetail } from 'components/DetailList';
 import ErrorDetail from 'components/ErrorDetail';
+import { CardBody, CardActionsRow } from 'components/Card';
+import { Detail, DetailList, UserDateDetail } from 'components/DetailList';
+import { VariablesDetail } from 'components/CodeEditor';
 import { formatDateString, secondsToHHMMSS } from 'util/dates';
-import { WorkflowApprovalsAPI, WorkflowJobsAPI } from 'api';
+import {
+  WorkflowApprovalsAPI,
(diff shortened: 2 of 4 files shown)


awx/ui/src/screens/WorkflowApproval/WorkflowApprovalDetail/WorkflowApprovalDetail.js
```diff
@@ -194,6 +258,88 @@ function WorkflowApprovalDetail({ workflowApproval }) {
...
+            label={t`Source Control Branch`}
+            value={workflowJob.scm_branch}
+            dataCy="wa-detail-source-job-scm"
+          />
+        ) : null}
+        {workflowJob?.summary_fields?.inventory ? (
+          <Detail
+            label={t`Inventory`}
+            value={
+              workflowJob.summary_fields?.inventory ? (
```
marshmalien: I don't think we need this additional check for `workflowJob.summary_fields?.inventory` since we check for the value in the ternary above.

awx/ui/src/components/AdHocCommands/AdHocCommandsWizard.test.js
```diff
@@ -368,7 +368,7 @@ describe('<AdHocCommandsWizard/>', () => {
         response: {
           config: {
             method: 'get',
-            url: '/api/v2/credentals',
+            url: '/api/v2/credentials',
```
marshmalien, line 371: 👍

awx/ui/src/screens/WorkflowApproval/WorkflowApprovalDetail/WorkflowApprovalDetail.test.js
```diff
@@ -1,6 +1,12 @@
 import React from 'react';
 import { act } from 'react-dom/test-utils';
-import { WorkflowApprovalsAPI } from 'api';
+import { Route } from 'react-router-dom';
+import { createMemoryHistory } from 'history';
```
marshmalien: These imports are unused.

marshmalien (approved):

awx/ui/src/screens/WorkflowApproval/WorkflowApprovalDetail/WorkflowApprovalDetail.test.js
```diff
@@ -12,18 +16,278 @@ import mockWorkflowApprovals from '../data.workflowApprovals.json';
...
+  });
+
+  test('should render Details', async () => {
+    let wrapper;
+    await act(async () => {
+      wrapper = mountWithContexts(
+        <WorkflowApprovalDetail workflowApproval={workflowApproval} />
+      );
+    });
+    waitForElement(wrapper, 'WorkflowApprovalDetail', (el) => el.length > 0);
```
kialam, line 290: RTL has a built-in function `waitFor` that we can leverage here. An example of its usage would be: `await waitFor(() => screen.getByText(/Workflow job details/i));`

awx/ui/src/screens/WorkflowApproval/WorkflowApprovalDetail/WorkflowApprovalDetail.test.js
```diff
@@ -1,6 +1,10 @@
...
 import { act } from 'react-dom/test-utils';
-import { WorkflowApprovalsAPI } from 'api';
+import {
+  WorkflowApprovalsAPI,
+  WorkflowJobTemplatesAPI,
+  WorkflowJobsAPI,
+} from 'api';
 import { formatDateString } from 'util/dates';
 import {
   mountWithContexts,
```
kialam, line 10: If we're writing unit tests, would it make sense to convert this to RTL instead of using Enzyme?

kialam (changes requested):
Good job here. I think the tests should be written using RTL instead of Enzyme. If you want help mounting contextualized components (documentation can be found [here](https://testing-library.com/docs/example-react-context/)), let me know.

awx/ui/src/screens/WorkflowApproval/WorkflowApprovalDetail/WorkflowApprovalDetail.test.js
nixocio: hey @kialam , I basically fixed several uni-tests. Once I added more requests to the workflowApprovalDetails, it was required to wrap them in `act` -  I agree with you that we should aim to write brand new tests using RTL.

kialam (approved):

akus062381 (approved):

Issue #12199, opened 2022-05-09
[Performance] Apply START_TASK_LIMIT to queries of pending jobs in the task manager


kdelee:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

In the theme of small changes to do to the task manager for scalability.

We introduced START_TASK_LIMIT as a way to prevent the task manager from timing out. Problem is, we only respect this on the "back end" of what the task manager is doing -- when it is starting the jobs 
https://github.com/ansible/awx/blob/a86740c3c9eaf9a551e850341d8adec5a3962dd5/awx/main/scheduler/task_manager.py#L220-L224

We still are exposed to the risk that everything on the "front end" may be slow (initializing data). Specifically, START_TASK_LIMIT is attempting to  limit how many `pending` jobs we process in a single run of the task manager.

By that logic, we should limit how many pending jobs we fetch from the database. For the moment, while we continue to have a single task manager with a global lock we must collect ALL running and waiting jobs. But we can limit how many pending jobs we collect, and I argue that although this  means 2 sets of queries (one for pending jobs, with a limit of START_TASK_LIMIT) and one for running/waiting jobs, we will get a net win for users with many  jobs (who are  the ones that need this feature)

This is because we KNOW that right now the max number of running/waiting jobs pratically is 1024, because that is the maximum number of databse connections we tend to configure. Of course Controller doens't actually know this is a practical limit but that is a different bug.

BUT...there can be many thousands of pending jobs. And it is in that particular case that we are exposed to the risk that START_TASK_LIMIT will not sufficiently limit how slow the task manager is, and customers can end up with the task manager timing out anyway.

### Implementation notes
We query the pending, waiting, and running jobs in 
https://github.com/ansible/awx/blob/a86740c3c9eaf9a551e850341d8adec5a3962dd5/awx/main/scheduler/task_manager.py#L89-L101

The approach we want to try is to split this into 2 sets of queries, one that does pending jobs with a limit https://docs.djangoproject.com/en/4.0/topics/db/queries/#limiting-querysets-1
We probably want to make sure we get the oldest pending jobs, e.g. order by created (ascending).

The other set of queries for running/waiting jobs would remain essentially unchanged. 

### Notes
[...]

PR #12214, opened 2022-05-11
apply start task limit to queries of pending jobs

kdelee:
The idea behind this is that it will limit exactly how slow the task
manager will get as the number of pending jobs grows.

For https://github.com/ansible/awx/issues/12199

##### SUMMARY

Apply start task limit also to how many pending jobs we consider in the first place. This is to prevent performance degredation when the number of pending jobs baloons.
See https://github.com/ansible/awx/issues/12199

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -122,19 +122,50 @@ class TaskManager:
 
         return None
 
+    def get_tasks_with_limit(self, status_list):
+        limit = None
+        order_by_list = []
+        tasks = []
+
+        if status_list == ['pending']:
+            limit = self.start_task_limit
+            # order by job_explanation and created gets the oldest jobs with no job explanation
+            # we set job explanation for jobs that are blocked on not having capacity
+            # this way we will consider them last
+            order_by_list = ['job_explanation', 'created']
+
+        jobs_qs = Job.objects.filter(status__in=status_list).prefetch_related('instance_group').order_by(*order_by_list)[:limit]
+        inventory_updates_qs = (
+            InventoryUpdate.objects.filter(status__in=status_list)
+            .exclude(source='file')
+            .prefetch_related('inventory_source', 'instance_group')
+            .order_by(*order_by_list)[:limit]
+        )
+        # Notice the job_type='check': we want to prevent implicit project updates from blocking our jobs.
+        project_updates_qs = (
+            ProjectUpdate.objects.filter(status__in=status_list, job_type='check').prefetch_related('instance_group').order_by(*order_by_list)[:limit]
+        )
+        system_jobs_qs = SystemJob.objects.filter(status__in=status_list).prefetch_related('instance_group').order_by(*order_by_list)[:limit]
+        ad_hoc_commands_qs = AdHocCommand.objects.filter(status__in=status_list).prefetch_related('instance_group').order_by(*order_by_list)[:limit]
+        workflow_jobs_qs = WorkflowJob.objects.filter(status__in=status_list).order_by(*order_by_list)[:limit]
+
+        # unpack query sets to a list because that is the return type expected
+        for task_qs in [jobs_qs, inventory_updates_qs, project_updates_qs, system_jobs_qs, ad_hoc_commands_qs, workflow_jobs_qs]:
+            tasks.extend([t for t in task_qs])
+
+        tasks = sorted(tasks, key=lambda task: task.created)
+
+        return tasks
+
     @timeit
     def get_tasks(self, status_list=('pending', 'waiting', 'running')):
-        jobs = [j for j in Job.objects.filter(status__in=status_list).prefetch_related('instance_group')]
-        inventory_updates_qs = (
-            InventoryUpdate.objects.filter(status__in=status_list).exclude(source='file').prefetch_related('inventory_source', 'instance_group')
-        )
-        inventory_updates = [i for i in inventory_updates_qs]
-        # Notice the job_type='check': we want to prevent implicit project updates from blocking our jobs.
-        project_updates = [p for p in ProjectUpdate.objects.filter(status__in=status_list, job_type='check').prefetch_related('instance_group')]
-        system_jobs = [s for s in SystemJob.objects.filter(status__in=status_list).prefetch_related('instance_group')]
-        ad_hoc_commands = [a for a in AdHocCommand.objects.filter(status__in=status_list).prefetch_related('instance_group')]
-        workflow_jobs = [w for w in WorkflowJob.objects.filter(status__in=status_list)]
-        all_tasks = sorted(jobs + project_updates + inventory_updates + system_jobs + ad_hoc_commands + workflow_jobs, key=lambda task: task.created)
+        all_tasks = []
+
+        if 'pending' in status_list:
+            all_tasks.extend(self.get_tasks_with_limit(['pending']))
+
+        all_tasks.extend(self.get_tasks_with_limit(tuple(s for s in status_list if s != 'pending')))
+
         return all_tasks
 
     def get_running_workflow_jobs(self):


awx/main/scheduler/task_manager.py
```diff
@@ -41,6 +41,20 @@
...
+perf_logger = logging.getLogger('awx.analytics.performance')
+
+
+def perf_logger_timer(func):
+    def inner(*args, **kwargs):
+        start = tz_now()
+        result = func(*args, **kwargs)
+        cls = args[0]
+        perf_logger.debug(
+            f'{{"func": task_manager.{func.__name__}, "duration": {tz_now() - start}, "pending_task_count": {cls.pending_task_count}, "running_task_count": {cls.running_task_count}}}'
```
AlanCoding: I don't love the handling of this in debug logger statements, because this requires a lot of work for later extraction of the data. I'd rather have us collect the timings in a more complete package, which is bound around the `TaskManager` class, as you're kind of leaning towards here. Ultimately, you want something that can be made into a pie chart at the end of it.

@fosterseth has done some work for efficient collection of stats in subsystem metrics. Since that's how we measure the callback receiver, I think we should use the same tool to measure the task manager timing breakdown.

Either way, it's going to go smoothest if we separate it from the rest of your work here.

kdelee: ok, so move the logger out

kdelee:
@AlanCoding confirmed my hypothesis that in general, there may be cases in which the old way would be faster when there are very few jobs to schedule. But the new way does not slow down when there are many jobs to schedule. 

Without change:
```
2022-05-12 18:00:48,541 DEBUG    [5ae2fec737f0453fa2ad007c97e4dc36] awx.analytics.performance {"func": task_manager._schedule, "duration": 0:00:55.748403, "pending_task_count": 1900, "running_task_count": 56}
2022-05-12 18:02:24,343 DEBUG    [e95bfa9f587f40beae174aa9285dbcca] awx.analytics.performance {"func": task_manager._schedule, "duration": 0:01:06.934849, "pending_task_count": 1820, "running_task_count": 77}
2022-05-12 18:07:56,484 DEBUG    [74eb9d3fc6424fba9c41a4a065446a91] awx.analytics.performance {"func": task_manager._schedule, "duration": 0:00:44.633093, "pending_task_count": 1541, "running_task_count": 56}
2022-05-12 18:13:12,674 DEBUG    [74eb9d3fc6424fba9c41a4a065446a91] awx.analytics.performance {"func": task_manager._schedule, "duration": 0:00:23.992480, "pending_task_count": 1208, "running_task_count": 44}
2022-05-12 18:18:40,890 DEBUG    [1e43735ce58146c1a6b82169b0ce904b] awx.analytics.performance {"func": task_manager._schedule, "duration": 0:00:21.407403, "pending_task_count": 872, "running_task_count": 44}
2022-05-12 18:20:19,397 DEBUG    [1e43735ce58146c1a6b82169b0ce904b] awx.analytics.performance {"func": task_manager._schedule, "duration": 0:00:14.573423, "pending_task_count": 780, "running_task_count": 20}
2022-05-12 18:21:43,857 DEBUG    [1e43735ce58146c1a6b82169b0ce904b] awx.analytics.performance {"func": task_manager._schedule, "duration": 0:00:34.385078, "pending_task_count": 680, "running_task_count": 86}
```

With change:
```
2022-05-12 18:39:06,529 DEBUG    [e4be777fd31a412395423537589cd3e1] awx.analytics.performance {"func": task_manager._schedule, "duration": 0:00:16.628090, "pending_task_count": 100, "running_task_count": 45}
2022-05-12 18:40:26,820 DEBUG    [e4be777fd31a412395423537589cd3e1] awx.analytics.performance {"func": task_manager._schedule, "duration": 0:00:31.576494, "pending_task_count": 100, "running_task_count": 78}
2022-05-12 18:41:14,491 DEBUG    [71e09893708a4c649b5af6207649075d] awx.analytics.performance {"func": task_manager._schedule, "duration": 0:00:31.936805, "pending_task_count": 100, "running_task_count": 58}
2022-05-12 18:42:21,225 DEBUG    [71e09893708a4c649b5af6207649075d] awx.analytics.performance {"func": task_manager._schedule, "duration": 0:00:30.022297, "pending_task_count": 100, "running_task_count": 69}
2022-05-12 18:43:17,365 DEBUG    [71e09893708a4c649b5af6207649075d] awx.analytics.performance {"func": task_manager._schedule, "duration": 0:00:31.434857, "pending_task_count": 100, "running_task_count": 60}
2022-05-12 18:45:14,772 DEBUG    [71e09893708a4c649b5af6207649075d] awx.analytics.performance {"func": task_manager._schedule, "duration": 0:00:28.618118, "pending_task_count": 100, "running_task_count": 67}
2022-05-12 18:46:16,076 DEBUG    [4d07a677032d41029e8779e9289b755f] awx.analytics.performance {"func": task_manager._schedule, "duration": 0:00:35.602123, "pending_task_count": 100, "running_task_count": 69}
2022-05-12 18:47:10,462 DEBUG    [0ca773cdaf074da8ab5350f8f3c207eb] awx.analytics.performance {"func": task_manager._schedule, "duration": 0:00:27.404066, "pending_task_count": 100, "running_task_count": 66}
```

I think perhaps I'm missing the part where it reschedules itself if there are still pending jobs though, because 2 min pass between the runs.

Maybe I should fetch START_TASK_LIMIT * 2 or START_TASK_LIMIT + 1 pending jobs to pick up https://github.com/ansible/awx/blob/cfdba959dda7680712fd01ecdaca237d0041c05c/awx/main/scheduler/task_manager.py#L224

awx/main/scheduler/task_manager.py
```diff
@@ -87,17 +87,32 @@ def job_blocked_by(self, task):
...
-            InventoryUpdate.objects.filter(status__in=status_list).exclude(source='file').prefetch_related('inventory_source', 'instance_group')
-        )
-        inventory_updates = [i for i in inventory_updates_qs]
-        # Notice the job_type='check': we want to prevent implicit project updates from blocking our jobs.
-        project_updates = [p for p in ProjectUpdate.objects.filter(status__in=status_list, job_type='check').prefetch_related('instance_group')]
-        system_jobs = [s for s in SystemJob.objects.filter(status__in=status_list).prefetch_related('instance_group')]
-        ad_hoc_commands = [a for a in AdHocCommand.objects.filter(status__in=status_list).prefetch_related('instance_group')]
-        workflow_jobs = [w for w in WorkflowJob.objects.filter(status__in=status_list)]
-        all_tasks = sorted(jobs + project_updates + inventory_updates + system_jobs + ad_hoc_commands + workflow_jobs, key=lambda task: task.created)
+        all_tasks = []
```
AlanCoding, line 162: `all_tasks` becomes misleading as a variable name now.

While I don't want to dive off the abyss of refactors, this does call into question what all the other task manager methods are expecting, like the suspicious `get_inventory_source_tasks`. Should we maybe categorize which methods require what statuses? Maybe. I don't feel totally comfortable with this merge until I have at least some cursory understanding of that.

AlanCoding:
This is going to be a large performance enhancement, but I want to hold off for some time period until we can measure it better. That means:

 - first merging https://github.com/ansible/awx/pull/12240 to get dev environment dashboards from metrics, which is the intended use case
 - add the task manager metrics https://github.com/ansible/awx/pull/12235
 - touch up the default dashboard, and include some task manager metrics

If these are sufficient to demonstrate the gain we get from this, then that will be enough. However, I have my doubts. I don't think we will easily be able to measure a difference without hacking. This hacking will be to prevent the normal task manager from running so we're _able_ to build up a large queue of jobs.

After this, I want to further develop some hacking so that we can switch the jobs query to use the UnifiedJob model.

awx/main/scheduler/task_manager.py
```diff
@@ -122,19 +122,51 @@ def job_blocked_by(self, task):
...
+        for task_qs in [jobs_qs, inventory_updates_qs, project_updates_qs, system_jobs_qs, ad_hoc_commands_qs, workflow_jobs_qs]:
+            tasks.extend([t for t in task_qs])
+
+        tasks = sorted(tasks, key=lambda task: task.created)
+
+        # update modified on pending tasks so we loop over all pending tasks in a window "limit" size, making sure we don't get stuck re-processing
+        # possibly unstartable pending tasks. E.g. we get the least recently modified each time, but each time we update modified so the next
+        # round of the task manager will pick up a new set of pending jobs (in the case that there are jobs stuck in pending due to capacity or other issues)
+        if status_list == ['pending'] and tasks:
+            UnifiedJob.objects.non_polymorphic().filter(pk__in=[t.pk for t in tasks]).update(modified=now)
```
AlanCoding: This is a pretty heavy update here. In severe situations, we're talking about 100 updates due to the task start limit. This is committed as a part of the overall task manager transaction, which can take up to 5 minutes, but commonly in the 10s of seconds. I'm less concerned about the number of 100, as opposed to the fact that it will happen over and over again to the same jobs. The task manager will run way more than once per 20 seconds in heavy load, and this will result in a gob-smacking amount of activity.

It also seems at least half-way unnecessary. Let's say we go into this with 30 pending jobs that are blocked on a resource or capacity, and 30 pending jobs that will be transitioned here. This will update the `modified` of all 60 jobs, but only 30 of them were necessary to update for your goal. It would be possible to, instead, track the blocked / un-started pending jobs and make the modified update at the end, meaning only 30 rows would need to be updated.

This also kind of violates our ideas of processing order. We sort tasks in order of `created`, but if (number of pending jobs) > (task start limit) it won't really be functioning that way. Jobs started later will jump the queue in front of other jobs started earlier.

I dislike the direction this is going as well. I want us to be doing fewer things in the task manager, this adds more, and it will be hard to get rid of.

kdelee: dropped the update

awx/main/scheduler/task_manager.py
```diff
@@ -122,19 +122,48 @@ def job_blocked_by(self, task):
...
-        jobs = [j for j in Job.objects.filter(status__in=status_list).prefetch_related('instance_group')]
+    def get_tasks_with_limit(self, status_list):
+        limit = None
+        if status_list == ['pending']:
+            limit = self.start_task_limit
+        tasks = []
+        # order by job_explanation and modified gets least recently modified with no job explanation
+        # we set job explanation for jobs that are blocked on not having capacity
+        # this way we will consider them last
+        order_by_list = ['job_explanation', 'created']
```
AlanCoding: I don't see any need to order_by for running jobs. The performance concerns would argue for not doing that.

AlanCoding:
I have to make the point that `job_explanation` is not indexed.

https://github.com/ansible/awx/blob/237402068ca3b4f763d7529a74c70d3ce1b9d7d6/awx/main/models/unified_jobs.py#L663

For small queries, I don't think this will matter, and we can rely on postgres to handle it gracefully. I am tempted to say we should merge it as it is right now.

kdelee:
@AlanCoding Ok I'll look into removing the job_explanation for order_by when looking at running/waiting jobs

kdelee:
Going to table this for now. There are some corner cases surrounding getting stalled out on "unstartable jobs". We're going to move forward on https://github.com/ansible/awx/pull/12437 and re-address the core issue after that

PR #12216, opened 2022-05-11
add param all_pages to method export_assets

jangel97:
Hi,

I am using the new awx cli (awxkit python package) `export_assets` method to export my AWX/AAP objects. 

I have realized that when trying to export a particular object using name, export_assets method will only return first 25 results (first page). In my case I have a credential with same name 49 times (this credential is repeated in some of my organisations and it has same name in all of them), I would need to be able to export those 49 credentials.

I tested the fork in my environment and it worked. Let me know if you need any other kind of verification

Does the MR look good to you? I made the `all_pages` method optional and default to False so it does not break any project. 

Let me know what you think.

Thanks!

### DIFF
--- a/awxkit/awxkit/api/pages/api.py
+++ b/awxkit/awxkit/api/pages/api.py
@@ -186,7 +186,7 @@ class ApiV2(base.Base):
             return endpoint.get(id=int(value))
         options = self._cache.get_options(endpoint)
         identifier = next(field for field in options['search_fields'] if field in ('name', 'username', 'hostname'))
-        return endpoint.get(**{identifier: value},all_pages=True)
+        return endpoint.get(**{identifier: value}, all_pages=True)
 
     def export_assets(self, **kwargs):
         self._cache = page.PageCache()
--- a/awxkit/awxkit/api/pages/api.py
+++ b/awxkit/awxkit/api/pages/api.py
@@ -186,7 +186,7 @@ class ApiV2(base.Base):
             return endpoint.get(id=int(value))
         options = self._cache.get_options(endpoint)
         identifier = next(field for field in options['search_fields'] if field in ('name', 'username', 'hostname'))
-        return endpoint.get(**{identifier: value})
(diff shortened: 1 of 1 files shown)


jbradberry (changes requested):
@jangel97 nice catch, but I don't think there is a need for preferring a single page of exports.  Let's drop the parameter from the export methods and just always use `all_pages=True` when doing the underlying `.get()` calls.  I see that we were already doing it in `PageCache.get_page()`, we just missed it for this case.

jangel97:
Hi @[jbradberry](https://github.com/jbradberry),
I just set the default value for `all_pages` in get method to true.
Let me know if you have any other issues.
Kind regards,

jbradberry:
@jangel97 not quite, we don't want to change the underlying behavior of the library for non-import/export uses.  I meant just use `all_pages=True` in the one spot in your original PR where you were doing an actual `.get()` call.  Here, I think:

https://github.com/ansible/awx/blob/devel/awxkit/awxkit/api/pages/api.py#L189

awxkit/awxkit/api/pages/api.py
```diff
@@ -186,7 +186,7 @@ def _filtered_list(self, endpoint, value):
             return endpoint.get(id=int(value))
         options = self._cache.get_options(endpoint)
         identifier = next(field for field in options['search_fields'] if field in ('name', 'username', 'hostname'))
-        return endpoint.get(**{identifier: value})
+        return endpoint.get(**{identifier: value},all_pages=True)
```
jbradberry: That's the ticket, but the lack of a space is going to cause the linting check to fail

jangel97: My bad, pushed it again with proper indentation

jbradberry (approved):

jangel97:
Hi,

I just set the `all_pages` parameter in the _filtered_list method, where it is invoked. 

Thanks you! Let me know if you think there is anything else. :)

PR #12215, opened 2022-05-11
Falling back to project.status if the last project sync job was deleted

john-westcott-iv:
##### SUMMARY

Co authored with @nixocio 

Modify the project list UI so that its status will be displayed in the UI if the last project sync job is missing. 
Pre change:
image

Post change:
image

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - UI

##### AWX VERSION

```
awx: 21.0.1.dev89+g3976c4fedb
```

##### ADDITIONAL INFORMATION

```

```

PR #12219, opened 2022-05-11
Change Demo Project status to successful

AlanCoding:
##### SUMMARY
This requires UI changes from https://github.com/ansible/awx/pull/12215, and should not be fully considered until after that has landed.

This will change the user experience for the "Demo Project" created from the `create_preload_data` management command. These changes will avoid kicking off an initial project update, and instead directly populate the `scm_revision` and `playbook_files`. Populating these fields plus the `status` will make this project usable for the "Demo Job Template". For a new install, it will perform a project "sync" for the first launch of the Demo JT, to pull the known revision from github, at that time (it also might pull the controlplane EE).

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
21.0.0
```

### DIFF
--- a/awx/main/management/commands/create_preload_data.py
+++ b/awx/main/management/commands/create_preload_data.py
@@ -32,8 +32,10 @@ class Command(BaseCommand):
                             name='Demo Project',
                             scm_type='git',
                             scm_url='https://github.com/ansible/ansible-tower-samples',
-                            scm_update_on_launch=True,
                             scm_update_cache_timeout=0,
+                            status='successful',
+                            scm_revision='347e44fea036c94d5f60e544de006453ee5c71ad',
+                            playbook_files=['hello_world.yml'],
                         )
 
                     p.organization = o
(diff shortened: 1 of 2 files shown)


john-westcott-iv (approved):

AlanCoding:
I was able to confirm this was working after rebasing on top of that UI change.

jbradberry:
I am not comfortable with this change (reasons forthcoming...)

jbradberry:
Ok, after trying this out in a completely fresh dev environment, I do confirm that running the Demo job template doesn't break.  My mental model of this was suggesting that it would attempt to run without having a checkout of the repo, so would fail in some horrible way.

However, it did take a weirdly long amount of time to run.

AlanCoding:
Curious if it took a long time to run because it was pulling the controlplane image?

nixocio:
@AlanCoding I also got this code locally and I also noticed that Demo Job Template took longer than usual to run.

image

Should we expect a Project Sync to be triggered as well - once the Job template was launched? Because that was not the case.

image


image

We still do not have data about the job related to project sync.

AlanCoding:
> I also got this code locally and I also noticed that Demo Job Template took longer than usual to run.

Yes, because it's pulling the image.

> Should we expect a Project Sync to be triggered as well - once the Job template was launched? Because that was not the case.

Inexact wording could be tricky here. When the job template runs, it should run a project "sync", as in `launch_type` has value of "sync". This does not count as something which can update the _project revision_. So the message in the UI is accurate, in that there is no project update to point to which updates the revision. There was a project sync to pull the source tree before the job ran.

jbradberry (approved):

PR #12222, opened 2022-05-12
Add help text popovers to /#/applications details fields

nixocio:
Add help text popovers to /#/applications details fields

See: https://github.com/ansible/awx/issues/11873

### DIFF
--- a/awx/ui/src/screens/Application/shared/Application.helptext.js
+++ b/awx/ui/src/screens/Application/shared/Application.helptext.js
@@ -1,0 +1,9 @@
+import { t } from '@lingui/macro';
+
+const applicationHelpTextStrings = {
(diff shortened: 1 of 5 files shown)


awx/ui/src/screens/Application/shared/Application.helptext.js
```diff
@@ -0,0 +1,9 @@
+import { t } from '@lingui/macro';
+
+const applicationHelpTextStrings = () => ({
```
kialam: Since the translations here are just strings, there's no need to make this an anonymous function. An object literal would be preferred.

kialam (changes requested):

kialam (approved):

InnocentK (approved):
This should be good to go once the CI checks pass

Issue #4481, opened 2019-08-14
RFE: Using set_stats module to pass a variable value from nested WT to JT


theansibler:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

On creating a Workflow template like below, we can able to pass the variable value from JT to WT: 

Job Template(passes the value of var1) >>On Success>> Nested Workflow(can use the value of var1)

But currently we cannot pass the variable value from WT to JT: 

Nested Workflow(passes the value of var1) >>On Success>> Job Template(can use the value of var1)

The purpose of this RFE is to pass variable value from a nested WT to JT/WT.

![wf](https://user-images.githubusercontent.com/53525271/63020258-435f6d80-bebb-11e9-95c2-34cffd1a56dd.png)

Siegfrost:
This enhancement would also be highly beneficial for our usage where we're trying to combine and nest WT created by different teams.
Hopefully this RFE can be prioritized!

JacobCallahan:
Since #6806 is considered a dupe of this issue, I'll add on our requests from that one.

##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Ansible Tower API RFE

Background:
Our team uses nested workflows. Each of which passes down information from 
the other using set_stats. At the end of the execution flows, the information 
is accumulated and formatted in our desired format. Our usage of Ansible Tower API is almost entirely for identifying and executing workflows.
This is expected to be at a high level, with most workflows having nested 
workflows and/or templates.

Problem:
With the current implementation of AT's API, set_stats can only flow down.
This means that, after the entire flow has finished executing, we need to 
find the last job that ran and pull the stats from it.

Desired Solution:
Implement a way to flow information back up to parent jobs and workflows. 
This way an API user can kick off a workflow or job execution and get all 
generated information from it once all child processes have completed.

Possible implementation:

Because workflows in general have arbitrary number of leaf nodes and we can not anticipate what will be the leaf nodes (for example,  some paths may not be taken depending on what happens at run time),  and it would be impossible to guess the preference of how to merge the different sets of artifacts and with what preference, as a first approach we propose setting the artifacts set by a job of a workflow node on the workflow node itself.

E.g. if a child job of a workflow job node has the `artifacts` property, set that on the workflow node.

As a second step, on the workflow, if workflow nodes in the workflow have the `artifacts` property set, save this on the workflow job itself in the form of a dictionary, where the key is some identifier for the workflow node, and the value is the artifacts set on that workflow node.

[...]

PR #12223, opened 2022-05-12
Pass combined artifacts from nested workflows into downstream nodes

AlanCoding:
##### SUMMARY
Connect https://github.com/ansible/awx/issues/4481

The issue is a reasonable request, and I am trying not to overthink the problem.

As of putting up this PR, I am doing this without a schema change. This trades lower memory use for more processing cost. Because of that, I've tried to write the method `get_combined_artifacts` relatively efficiently, although you can come up with cases where it will slog.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

##### AWX VERSION
```
21.0.0
```

##### ADDITIONAL INFORMATION
I've done two test cases
 - like the issue states, a workflow with (sliced job) --> (receiving job), and verify that receiving job gets artifacts
 - a workflow inside a workflow with normal job, and receiving job downstream in outer workflow, verify receiving job gets artifacts

Both test cases fail in `devel`, finding that the receiving job does not get the artifacts.


AlanCoding:
test results look good, ready for review.

AlanCoding:
having thought about it, I realize I do need to put in some recursion protection.

awx/main/models/workflow.py
```diff
@@ -57,6 +57,20 @@
 logger = logging.getLogger('awx.main.models.workflow')
 
 
+def get_artifacts_from_unified_job(job):
```
jbradberry: Instead of this standalone function, why not just have different implementations of `.get_combined_artifacts()` on the different `UnifiedJob` subtypes?

awx/main/models/workflow.py
```diff
@@ -682,6 +695,22 @@ def get_ancestor_workflows(self):
...
+    def get_combined_artifacts(self):
+        """
+        For downstream jobs of a workflow nested inside of a workflow,
+        we send aggregated artifacts from the nodes inside of the nested workflow
+        """
+        artifacts = {}
+        job_queryset = (
+            UnifiedJob.objects.filter(unified_job_node__workflow_job=self)
+            .defer('job_args', 'job_cwd', 'start_args', 'result_traceback')
+            .order_by('status', 'finished', 'id')
```
jbradberry: I'm a bit iffy on ordering by `status`.  This will wind up being alphabetical, which feels like it might have unintended consequences.

AlanCoding: haha, I was kind of fishing for a comment like this. The status (s)uccessful comes after (f)ailed and (e)rror, thus, the successful jobs will have their artifacts take precedence.

In practice, I'm iffy on whether error or failed jobs should contribute artifacts in the first place. Still unsure on what the best thing to do here is.

jbradberry: `.filter(status='successful')` instead, then?

This brings to mind a thing that I was enthusiastic about when I noticed it early on in the Django upgrades, but that got buried under the weight of dealing with the difficult bits: Django 3 supports actual enumerated choices for choice fields.  https://docs.djangoproject.com/en/3.2/ref/models/fields/#enumeration-types

It'd be nice to have these, and avoid having to hard-code choice strings.  Not urgent, just kind of pleasant.

awx/main/models/workflow.py
AlanCoding: Took me a while to process, but I wound up going with exactly this suggestion. That allowed getting rid of this method altogether.

awx/main/models/workflow.py
```diff
@@ -682,6 +681,27 @@ def get_ancestor_workflows(self):
...
+            .iterator()
+        )
+        if parents_set is None:
+            parents_set = set()
+        new_parents_set = parents_set | set([self.id])
```
jbradberry: I'd like to start getting away from Python 2-isms like `set([...])`.

awx/main/models/workflow.py
```diff
@@ -682,6 +681,28 @@ def get_ancestor_workflows(self):
             wj = wj.get_workflow_job()
         return ancestors
 
+    def get_effective_artifacts(self, parents_set=None):
```
jbradberry: Can we just make this signature `(self, **kwargs)` as well, and get `parents_set` out of the `kwargs` dict?  It will make things just a bit nicer if we wind up adding more things later on.

awx/main/models/workflow.py
jbradberry: Is it even possible to have a cycle in the jobs?  The job templates, yes, I could see that potentially happening, but even if the same JT is used in multiple places inside a nested workflow won't it just spawn distinct new jobs (thus having new pks)?

awx/main/models/workflow.py
```diff
@@ -318,8 +318,7 @@ def get_job_kwargs(self):
         for parent_node in self.get_parent_nodes():
             is_root_node = False
             aa_dict.update(parent_node.ancestor_artifacts)
-            if parent_node.job and hasattr(parent_node.job, 'artifacts'):
-                aa_dict.update(parent_node.job.artifacts)
+            aa_dict.update(parent_node.job.get_effective_artifacts(parents_set=set([self.workflow_job_id])))
```
jbradberry: :eyes: 

I need to think through this.  This is the part that I worry is going to cause a bunch of repeated queries.

jbradberry: Never mind, I think this isn't a problem.

I still dislike `set([...])`, though.

awx/main/models/workflow.py
AlanCoding: No, as structured I don't think it would have any cycles. The map of workflow jobs would always be a tree, since a workflow job node can't (in practice) link to the same workflow job as another job.

I kept coming back to questioning if we should track and skip duplicate _workflow job templates_, which is a much stronger criteria. A workflow can use the same WFJT inside of it multiple times, or it can be nested deeper. This would cut down on queries, but I'm not sure if users might actually want to combine them because multiple runs of the same WFJT could give different artifacts.

jbradberry (approved):

Issue #9014, opened 2021-01-05
User can add multiple Vault Credentials with same Vault ID to job template


AlexSCorey:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
A user can end up with a job template that has multiple vault credentials that have the same vault ID.

##### ENVIRONMENT
* AWX version: 14.1
* AWX install method:  docker for mac
* Ansible version:  2.9.16
* Operating System:Mojave
* Web Browser: Chrome

##### STEPS TO REPRODUCE

Create 2 vault credentials with different vault IDs.  Add them both to a job template.  Save job template.  Edit 1 of the vault credentials so that it has the same vault id as the other.  Go back to the job template form or details view.

##### EXPECTED RESULTS

Not sure.  Some sort of warning or icon or something.  @trahman73 what would you expect the UX to be here?

##### ACTUAL RESULTS
See that those credentials are still attached to that job template

##### ADDITIONAL INFORMATION

![vault](https://user-images.githubusercontent.com/12464858/103770751-45bced00-5005-11eb-8584-739f9f4e7328.gif)

jakemcdermott:
Discussed w/ the api team. Current plan is to make the vault-id non-editable, like we do for credential type.

harshakhmk:
Can I work on this issue?

rebeccahhh:
@harshakhmk we welcome contributions at any time. :)

PR #12227, opened 2022-05-12
Prevent edit of  vault ID once credential is created.

djyasin:
##### SUMMARY

related #9014 

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.0.1
```

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -2664,6 +2664,13 @@ class CredentialSerializer(BaseSerializer):
 
         return credential_type
 
+    def validate_inputs(self, inputs):
+        if self.instance and self.instance.credential_type.kind == "vault":
+            if 'vault_id' in inputs and inputs['vault_id'] != self.instance.inputs['vault_id']:
+                raise ValidationError(_('We do not permit Vault IDs to be changed after they have been created.'))
+
+        return inputs
+
 
 class CredentialSerializerCreate(CredentialSerializer):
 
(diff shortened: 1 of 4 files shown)


AlanCoding:
Might want to avoid the use of "We". Phrasing in a more ambiguous tense is better.

You might also want to consider allowing changing of the vault_id if the credential isn't in use by anything. So you could figure out the reverse relationship(s) and check `.exists()`. This might be nice to have for some users, but up to you.

djyasin:
Thank you, Alan! I will review the tone of our other messages and get this rewritten accordingly. I will absolutely bear that in mind moving forward!

fosterseth (approved):

john-westcott-iv:
Looking good. Do you want to mention something in this doc about the new limitation? https://github.com/ansible/awx/blob/devel/docs/credentials/multi_credential_assignment.md#specifying-multiple-vault-credentials

Also, were you able to think about @AlanCoding's comment about allowing the vault_id to change if the credential is not in use by any job template? Or will that be counter productive to the issue you are solving?

Also, along those lines this looks like this change would prevent me from adding a vault ID (i.e. changing from "" to a value). Should we allow for that? I'm thinking this should also be excluded but wanted to double check.

AlanCoding (approved):

AlexSCorey (approved):
The UI work is great. Well done, and thanks for contributing.

CFSNM (approved):

Issue #11167, opened 2021-09-29
Logout as User A and Login as User B redirects to User A's last location


AlexSCorey:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](http://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I am not entitled to status updates or other assurances.

### Summary

If I have a user, we will call it User A that is a System Administrator, that is using AWX, and logs out when they are done.  Then I log in as User B, who is a normal user,  User B is directed to the last location of User A.  

### AWX version

devel

### Installation method

docker for mac

### Modifications

no

### Steps to reproduce

Create 2 users.  Login as User A, and then log out.  Login as User B

### Expected results

User B is sent to the home screen

### Actual results

User B is sent to the last location of User A

### Additional information

![loginout](https://user-images.githubusercontent.com/39280967/135306478-b4b62135-ba1f-4b2d-87fd-7ab47fd14830.gif)

nixocio:
PR was approved and merged.

InnocentK:
Resolved

PR #12229, opened 2022-05-12
Persistent list filters

keithjgrant:
##### SUMMARY
Keeps any list search/pagination filters in place when clicking the "Back to XXX" button. Applies to all top-level list pages except Schedules page.

This also updates `RoutedTabs` to use the proper `href` prop on each tab, so URLs show as expected in the browser while hovering mouse over the tabs.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui/src/components/PersistentFilters/PersistentFilters.js
+++ b/awx/ui/src/components/PersistentFilters/PersistentFilters.js
@@ -1,0 +1,32 @@
+import { useEffect } from 'react';
+import { useLocation, useHistory } from 'react-router';
+
+const STORAGE_KEY = 'persistentFilter';
+
+export default function PersistentFilters({ pageKey, children }) {
+  const location = useLocation();
+  const history = useHistory();
+
+  useEffect(() => {
+    if (!location.search.includes('restoreFilters=true')) {
+      return;
+    }
+
+    const filterString = sessionStorage.getItem(STORAGE_KEY);
+    const filter = filterString ? JSON.parse(filterString) : { qs: '' };
+
+    if (filter.pageKey === pageKey) {
+      history.replace(`${location.pathname}${filter.qs}`);
+    }
+  }, [history, location, pageKey]);
+
+  useEffect(() => {
+    const filter = {
+      pageKey,
+      qs: location.search,
+    };
+    sessionStorage.setItem(STORAGE_KEY, JSON.stringify(filter));
+  }, [location.search, pageKey]);
+
+  return children;
+}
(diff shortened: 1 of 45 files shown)


awx/ui/src/components/RoutedTabs/RoutedTabs.js
```diff
@@ -39,7 +43,7 @@ function RoutedTabs({ tabsArray }) {
           aria-label={typeof tab.name === 'string' ? tab.name : null}
           eventKey={tab.id}
           key={tab.id}
-          link={tab.link}
+          href={`#${tab.link}`}
```
keithjgrant, line 46: This is a bit quirky... help the URL appear correctly while mousing over the link.

Ideally, I'd like to remove the hash from URLs entirely to make this unnecessary, but it may be too late for that

AlexSCorey (commented):
I'm not sure this is expected behavior.  The filtered list should load with I click `Back to templates`, right?
filters

awx/ui/src/components/PersistentFilters/PersistentFilters.js
```diff
@@ -0,0 +1,34 @@
+import { useEffect } from 'react';
+import { useLocation, useHistory } from 'react-router';
+
+const STORAGE_KEY = 'persistentFilter';
```
nixocio: We could move this one to the constant file. Also, the other constants used on session storage starts with awx to make ease to identify them, and avoid name collision.

awx/ui/src/screens/Template/Templates.js
```diff
@@ -78,7 +79,9 @@ function Templates() {
         </Route>
         <Route path="/templates">
           <PageSection>
-            <TemplateList />
+            <PersistentFilters pageKey="templates">
```
nixocio, line 82: I noticed a delay in certain pages to filter be evaluated using firefox, on chrome seems to be more responsive.



https://user-images.githubusercontent.com/9053044/168847315-a927759b-94d9-4f87-86ee-a4ac94e11aae.mov

nixocio (commented):
hey @keithjgrant, excellent work.  There is an issue on awx/devel once an user, lets say User A is logged out, another User b logging in back the User B will be redirect to where the URL where User A left. This issue was not introduced by this PR, but saving the filters will redirect to the previous saved filter.  This issue could be addressed by https://github.com/ansible/awx/issues/11167

AlexSCorey (approved):
All of my concerns have been addressed. Well done Keith!

InnocentK (approved):
After further investigating, it seems the above failed tests are unrelated. Great job on this!

PR #12235, opened 2022-05-13
Subsystem metrics for task manager

fosterseth:
##### SUMMARY

Adds a handful of subsystem metrics for the task manager. It tracks data from the last-ran task manager cycle only The metrics are saved only at the very end of the task manager `schedule()` call. . A consequence of that useful task manager data may be quickly overridden before prometheus can scrape the endpoint.

imagine this timeline of events

> prometheus scrapes /api/v2/metrics 
> queue a bunch of jobs
> task manager runs and processes all these jobs (maybe takes a long time to run)
> another task manager is scheduled and processes no jobs (takes ~0 seconds to run)
> prometheus scrapes /api/v2/metrics 

prometheus would have missed the first task manager cycle that did a lot of processing, and only "see" the data from the task manager that ran 0 jobs.

To combat this, the task manager will track each time it records metrics. If the last time the metrics were written is less than 15s ( `settings.SUBSYSTEM_METRICS_TASK_MANAGER_RECORD_INTERVAL`), the task manager will not record metrics to redis. This will give prometheus enough time to scrape the metrics endpoint and capture a snapshot of the task manager run.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.0.1.dev103+gff74e538cd
```

### DIFF
--- a/awx/main/analytics/subsystem_metrics.py
+++ b/awx/main/analytics/subsystem_metrics.py
@@ -8,7 +8,7 @@ from django.apps import apps
 from awx.main.consumers import emit_channel_notification
 
 root_key = 'awx_metrics'
-logger = logging.getLogger('awx.main.wsbroadcast')
+logger = logging.getLogger('awx.main.analytics')
 
 
 class BaseM:
@@ -16,16 +16,22 @@ class BaseM:
         self.field = field
         self.help_text = help_text
         self.current_value = 0
+        self.metric_has_changed = False
 
-    def clear_value(self, conn):
+    def reset_value(self, conn):
         conn.hset(root_key, self.field, 0)
         self.current_value = 0
 
     def inc(self, value):
         self.current_value += value
+        self.metric_has_changed = True
 
     def set(self, value):
         self.current_value = value
+        self.metric_has_changed = True
+
+    def get(self):
+        return self.current_value
 
     def decode(self, conn):
         value = conn.hget(root_key, self.field)
@@ -34,7 +40,7 @@ class BaseM:
     def to_prometheus(self, instance_data):
         output_text = f"# HELP {self.field} {self.help_text}\n# TYPE {self.field} gauge\n"
         for instance in instance_data:
-            output_text += f'{self.field}{{node="{instance}"}} {instance_data[instance].get(self.field, -1)}\n'  # TODO: fix because this -1 is neccessary when dealing with old instances (ex. you didn't clean up your database)
+            output_text += f'{self.field}{{node="{instance}"}} {instance_data[instance][self.field]}\n'
         return output_text
 
 
@@ -46,8 +52,10 @@ class FloatM(BaseM):
             return 0.0
 
     def store_value(self, conn):
-        conn.hincrbyfloat(root_key, self.field, self.current_value)
-        self.current_value = 0
+        if self.metric_has_changed:
+            conn.hincrbyfloat(root_key, self.field, self.current_value)
+            self.current_value = 0
+            self.metric_has_changed = False
 
 
 class IntM(BaseM):
@@ -58,8 +66,10 @@ class IntM(BaseM):
             return 0
 
     def store_value(self, conn):
-        conn.hincrby(root_key, self.field, self.current_value)
-        self.current_value = 0
+        if self.metric_has_changed:
+            conn.hincrby(root_key, self.field, self.current_value)
+            self.current_value = 0
+            self.metric_has_changed = False
 
 
 class SetIntM(BaseM):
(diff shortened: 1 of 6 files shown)


awx/main/scheduler/task_manager.py
```diff
@@ -620,4 +658,14 @@ def schedule(self):
                 logger.debug("Starting Scheduler")
                 with task_manager_bulk_reschedule():
                     self._schedule()
+                if not settings.IS_TESTING(sys.argv):
```
kdelee: not sure I understand this one setting

awx/main/analytics/subsystem_metrics.py
```diff
@@ -170,6 +170,17 @@ def __init__(self, auto_pipe_execute=True, instance_name=None):
...
+            FloatM('task_manager_start_task_seconds', 'Time spent starting task'),
+            FloatM('task_manager_process_running_tasks_seconds', 'Time spent processing running tasks'),
+            FloatM('task_manager_process_pending_tasks_seconds', 'Time spent processing pending tasks'),
+            FloatM('task_manager_generate_dependencies_seconds', 'Time spent generating dependencies for pending tasks'),
+            FloatM('task_manager_spawn_workflow_graph_jobs_seconds', 'Time spent spawning workflow jobs'),
+            FloatM('task_manager__schedule_seconds', 'Time spent in running the entire _schedule'),
+            FloatM('task_manager_tasks_started', 'Number of tasks started'),
+            FloatM('task_manager_running_processed', 'Number of running tasks processed'),
+            FloatM('task_manager_pending_processed', 'Number of pending tasks processed'),
+            FloatM('task_manager_tasks_blocked', 'Number of tasks blocked from running'),
```
kdelee: :tada:

kdelee (approved):

awx/main/scheduler/task_manager.py
kdelee: ok I get it now that I found...don't collect these metrics if tests are running https://github.com/ansible/awx/blob/78660ad0a22798ace9210b37ba5be9429603e49d/awx/settings/defaults.py#L16-L31

awx/main/scheduler/task_manager.py
```diff
@@ -657,5 +712,8 @@ def schedule(self):
                     return
                 logger.debug("Starting Scheduler")
                 with task_manager_bulk_reschedule():
+                    # if sigterm due to timeout, still record metrics
+                    signal.signal(signal.SIGTERM, self.record_aggregate_metrics)
```
fosterseth: @kdelee as discussed, this will ensure we record the metrics in the event the dispatcher kills the task manager due to the 5 minute timeout.

it will also record a `task_manager_total_runs` metric that increments each time the task manager runs

awx/main/analytics/subsystem_metrics.py
```diff
@@ -70,10 +80,9 @@ def decode_value(self, value):
...
     def store_value(self, conn):
-        # do not set value if it has not changed since last time this was called
-        if self.current_value is not None:
+        if self.metric_has_changed:
             conn.hset(root_key, self.field, self.current_value)
-            self.current_value = None
+            self.metric_has_changed = False
 
 
 class SetFloatM(SetIntM):
```
kdelee, line 88: Could you add a doc string to the SetFloatM and SetIntM classes to help in the future distinguish between when one would be used vs the FloatM and IntM classes? The names don't communicate enough context. From what I gather w/ @rebeccahhh's PR the SetFloatM and SetIntM are necessary to use if you plan to completely wipeout a metric each time you do push it to redis. The IntM and FloatM `.set` methods we behaving differently when we were experimenting but I never got whole gist of whats going on (they were incrementing periodically in a weird way)

fosterseth: my plan is to write up docs/subsystem_metrics.md to explain how the subsystem metrics works

kdelee:
@fosterseth @AlanCoding whats between this and merging?

awx/main/scheduler/task_manager.py
```diff
@@ -648,6 +683,28 @@ def _schedule(self):
...
+            if time_last_recorded > settings.SUBSYSTEM_METRICS_TASK_MANAGER_RECORD_INTERVAL:
+                logger.debug(f"recording metrics, last recorded {time_last_recorded} seconds ago")
+                self.subsystem_metrics.set("task_manager_recorded_timestamp", current_time)
+                self.subsystem_metrics.pipe_execute()
+            else:
+                logger.debug(f"skipping recording metrics, last recorded {time_last_recorded} seconds ago")
+
+    def record_aggregate_metrics_and_exit(self, *args):
+        self.record_aggregate_metrics()
+        sys.exit(1)
```
AlanCoding, line 706: what does this `sys.exit(1)` do to the transaction? I think it rolls it back right? I assume that would be the same as the current behavior. Right? Probably.

AlanCoding (approved):

PR #12249, opened 2022-05-17
use qs params when fetching new/updated jobs to preserve filters

keithjgrant:
##### SUMMARY
Applies active search filters when adding/updating jobs in the job list. This prevents sync jobs or other misc jobs that don't meet the criteria from appearing (and then disappearing if the page is refreshed)

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - UI


nixocio (approved):
👍

nixocio:
I triggered the e2e tests.

Issue #10409, opened 2021-06-10
Jobs should report image SHA256 and EEs should give latest value


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
If someone runs a job with `quay.io/ansible/awx-ee:devel`, that doesn't give traceability. There are a great multitude of ways that could have ran with something other than what was expected (expectation isn't even _clear_). Even the tags like `:0.3.0` are not ironclad.

If a job gave `quay.io/ansible/awx-ee:devel`, `9c55e8c5d023`, then I could find what that maps to in the build history.

https://quay.io/repository/ansible/awx-ee?tab=history

Likewise, jobs & projects give a git hash.

We don't have traceability around roles & collections from `collections/requirements.yml`, but this is being de-emphasized in favor of EEs. So EEs should at least have a mechanism of real audit / traceability.

nixocio:
We could also show that hash value on the UI.

Issue #3780, opened 2019-04-29
Support Direct Role Execution


matburt:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Allow creating and executing Roles (possibly via the ad-hoc interface) and via Workflows.

Where Roles are possibly sourced via Collections.

mabashian:
OK so for role execution in a workflow the UI/UX will depend a little bit on https://github.com/ansible/awx/issues/3776.  If we decide to go down the `collections are a subset of projects` then I think this would mean new rows showing up here:

<img width="421" alt="Screen Shot 2019-04-30 at 4 22 37 PM" src="https://user-images.githubusercontent.com/9889020/56990863-32852900-6b64-11e9-8ca2-e67003318c14.png">

but I'm not sure that `Project Sync` would accurately describe that list anymore.  Alternatively we could have a new list of `Collections` that the user could choose from.  When the user does select a `Collection` are there going to be other inputs needed?  What might those fields look like?

We probably want to add a new a new node type:

<img width="209" alt="Screen Shot 2019-04-30 at 4 24 25 PM" src="https://user-images.githubusercontent.com/9889020/56990939-737d3d80-6b64-11e9-8301-215a2d132222.png">

for collections...

If we go down the route of making components separate from projects then a lot of the above still applies but projects/collections will likely not be mixed confirming the necessity for a new tab.

I'm not sure if this makes sense for this particular issue but I know that https://github.com/ansible/awx/issues/2253 has been thrown out there in the past.  Would something like that make the most sense for supporting role execution in a workflow node?

matburt:
My thought here, is that Roles *could* actually come from projects even if we were to decide on a top level collections resources (I'm inclined to think these should all go under Projects, and multi-project job templates need to be a thing anyway).

Role execution I think will look a lot like ad-hoc execution from the inventory screen anyway. Then being able to add ad-hoc and role jobs to a workflow.

mabashian:
Updated UI plan:

- This will not impact projects/project sync nodes in any way
- A new node type will be supported "role execution"
- Creating a role execution node will mirror creating other nodes
- UI for creating nodes in general may need to change from tabs (Templates, Projects, Inventories) to a dropdown now that we'll have a fourth type (and maybe more types to come).
- First step when creating a role execution node will be to select a collection
- After selecting a collection, the user will either type in a role or select a role from a list.  This implementation detail depends on whether or not the api introspects the collection.
[...]

PR #12250, opened 2022-05-18
Experimental direct role execution

matburt:
##### SUMMARY

See this old issue: https://github.com/ansible/awx/issues/3780

This is work towards adding Direct Role execution capabilities into AWX. Since we're moving towards Execution Environments as the mode of execution and Collections as one of the primary delivery pipelines adding support for this gives another vector for delivering automation directly to Controller itself.

Having direct role execution also enables more targeted use of Role Arg Spec as a capability within Ansible core itself, potentially being able to inform Survey semantics. This also leads towards being able to select playbooks from Collections as a source of automation.

The functionality provided here may also lead to directly publishing content from GalaxyNG and Dev Tools where those systems have a deep understanding of the source and provenance of the automation content but don't have the capability to create and publish a source control repo. Imagine a scenario where a developer is testing a role or a playbook and has the ability to push it directly into AWX organization specifically confined for automation testing.

This change also encourages users and organizations to standardize on automation content publishing without needing to expose source control semantics to the automation system itself.

The way this PR is implemented is that it removes the hard requirement that a project and playbook be defined and instead allows either a Project/Playbook to be used OR the FQCN+Role be specified. This will trigger different behavior in Ansible Runner.

This PR is missing some things that would make it generally more useful for an integration:

- [x] Tests
- [ ] Documentation
- [ ] Collection and Role indexing from the Execution Environment itself. This is currently done in a partial way from the Job runtime and should be changed so that the information is also stored adjacent to the EE.
- [ ] UI/UX -- The UI is deeply confused and concerned that there's no project or playbook assigned 
- [ ] We can probably take care of https://github.com/ansible/awx/issues/10409 at the same time, as right now there's just a placeholder value for the SCM revision.
- [ ] Direct Playbook Execution from a Collection
(More that I'm sure will come up as part of discussion)

There are some adjacent things I'm using as part of the validation while working on this, namely:

- https://github.com/ansible/ansible-tower-samples/pull/26 is a PR on the demo git repo that allows generating a collection from its content
- The EE definition that makes use of that change to produce an EE containing that collections: https://gist.github.com/matburt/94c5410a4ba7da20c9cf09ce07091c73
- The EE (quay.io/matburt/ansible-sample-ee:latest) that I've published to my quay account that is the manifestation of both of those things: https://quay.io/repository/matburt/ansible-sample-ee?tab=tags

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API
 - UI
 - Collection

##### AWX VERSION

```
HEAD
```

##### ADDITIONAL INFORMATION

directro8

```

```


awx/api/serializers.py
```diff
@@ -2778,6 +2778,7 @@ class Meta:
             'inventory',
             'project',
             'playbook',
+            'fqcn_role',
```
matburt, line 2781: Not sure how I feel about the name here... "role" is mostly already defined for the rbac system.

awx/api/serializers.py
```diff
@@ -2828,25 +2829,27 @@ def to_representation(self, obj):
         return ret
 
     def validate(self, attrs):
-        if 'project' in self.fields and 'playbook' in self.fields:
+        if 'project' in self.fields and 'playbook' in self.fields and 'fqcn_role' in self.fields:
```
matburt, line 2832: I feel like a lot of these conditionals should be reworked, but for a first stab at it... these work.

awx/api/serializers.py
```diff
@@ -2828,25 +2829,27 @@ def to_representation(self, obj):
...
-            )
-            if playbook_not_found:
-                raise serializers.ValidationError({'playbook': _('Playbook not found for project.')})
-            if project and not playbook:
-                raise serializers.ValidationError({'playbook': _('Must select playbook for project.')})
-            if scm_branch and not project.allow_override:
-                raise serializers.ValidationError({'scm_branch': _('Project does not allow overriding branch.')})
-            if ask_scm_branch_on_launch and not project.allow_override:
-                raise serializers.ValidationError({'ask_scm_branch_on_launch': _('Project does not allow overriding branch.')})
+            if fqcn_role in (None, ''):
```
matburt, line 2838: The potential duality of testing these makes these conditional checks really uncomfortable.

awx/api/serializers.py
```diff
@@ -2828,25 +2829,27 @@ def to_representation(self, obj):
...
+                if not project:
+                    raise serializers.ValidationError({'project': _('This field is required.')})
```
matburt, line 2840: We don't ... that I'm aware ... have a way to represent something like this in an error "Either do this OR do this other thing"

awx/main/models/jobs.py
```diff
@@ -291,14 +297,13 @@ def validation_errors(self):
...
-                _("Job Templates must have a project assigned."),
-            ]
+            if self.fqcn_role == '' or self.fqcn_role is None:
+                validation_errors['project'] = [_("Job Templates must have a role or project assigned")]
         return validation_errors
 
     @property
     def resources_needed_to_start(self):
-        return [fd for fd in ['project', 'inventory'] if not getattr(self, '{}_id'.format(fd))]
+        return [fd for fd in ['inventory'] if not getattr(self, '{}_id'.format(fd))]
```
matburt: This is a mostly unacceptable change, but it gets us past the check itself. There's a complex layer of validation on these properties and it feels like we need another representation. At least at this level.

awx/main/tasks/jobs.py
```diff
@@ -10,6 +10,7 @@
 from pathlib import Path
 import shutil
 import stat
+from xxlimited import foo
```
matburt: Not sure where this came from, I'll remove it

awx/api/serializers.py
```diff
@@ -2970,13 +2973,14 @@ def get_field_from_model_or_attrs(fd):
...
+        fqcn_role = get_field_from_model_or_attrs('fqcn_role')
 
         if get_field_from_model_or_attrs('host_config_key') and not inventory:
             raise serializers.ValidationError({'host_config_key': _("Cannot enable provisioning callback without an inventory set.")})
 
         prompting_error_message = _("Must either set a default value or ask to prompt on launch.")
-        if project is None:
-            raise serializers.ValidationError({'project': _("Job Templates must have a project assigned.")})
+        if project is None and fqcn_role == '':
+            raise serializers.ValidationError({'project': _("Job Templates must have a project or role assigned.")})
```
AlanCoding, line 2983: For validation - should `execution_environment` be required if `fqcn_role` is set?

Otherwise you would be using the project / default EE. While yes, that might conceptually work, in a way, it doesn't make sense. Providing the FQCN path is creating a hard dependency on the execution environment. Ultimately (even if not now), the user experience for doing this should start with the EE.

matburt: Possibly... I think my question would be, is there a scenario where the user overrides the default EE and thus expects that default EE to be capable of role execution without the EE being supplied to the Job Template?

awx/main/tasks/jobs.py
```diff
@@ -845,48 +849,52 @@ def pre_run_hook(self, job, private_data_dir):
...
-        # Galaxy requirements are not supported for manual projects
-        if job.project.scm_type and ((not has_cache) or branch_override):
-            sync_needs.extend(['install_roles', 'install_collections'])
+            has_cache = os.path.exists(os.path.join(job.project.get_cache_path(), job.project.cache_id))
+            # Galaxy requirements are not supported for manual projects
+            if job.project.scm_type and ((not has_cache) or branch_override):
+                sync_needs.extend(['install_roles', 'install_collections'])
+        else:
+            sync_needs = []
+            job_revision = "collection"
```
AlanCoding, line 897: We're here if `job.fqcn_role` is not blank...

Would this allow running from a project's collection requirements? In that case you'd have both a project and a role path. This stuff could get weird.

awx/api/serializers.py
matburt: Collecting details about whether or not a collection and role is available in the selected EE is probably worth persuing as an indication to the user that this job might not run with the currently selected EE. It is entirely possible that the EE could materialize the collection and role at runtime, though.

awx/main/tasks/jobs.py
matburt: This is why I make this an either/or thing.... maybe I could be persuaded against that but I'm drawing a line here that says: No project + FQCN. The alternative is to have this be entirely vs a selected playbook but that removes some of the utility of just being able to selected an EE sans project.

AlanCoding:
I had imagined direct role execution as more of a special case of ad hoc commands, as opposed to job templates. Although I'm still thinking through that, if the underlying command is `ansible`, I think it should be an ad hoc command. If the underlying command is `ansible-playbook`, what you're doing here would be preferable.

matburt:
Honestly, I'd like to get away from the way we do ad-hoc as a concept. It's a dark pattern and hidden behind the inventory interface unnecessarily.

AlanCoding:
https://docs.ansible.com/ansible/latest/user_guide/intro_adhoc.html

https://ansible-runner.readthedocs.io/en/stable/standalone/?highlight=role#running-roles-directly

Well, it's true that direct role execution is not a canonical Ansible core construct, and I see that ansible-runner writes a playbook on the fly (sorry this was not obvious to me at the start). Content being:

```python
       play = [{'hosts': vargs.get('hosts') if vargs.get('hosts') is not None else "all",
                 'gather_facts': not vargs.get('role_skip_facts'),
                 'roles': [role]}]
```

These implementation details have a lot of impact on the user experience.

AlanCoding:
I'm going to close this, as I would like to revive some still-workable related issues, like the ability to run playbooks from a collection.

PR #12252, opened 2022-05-18
Double escape all unicode escape sequences in job events data

jbradberry:
##### SUMMARY
Double escape all unicode escape sequences in job events data when collecting it for analytics.  This will have the following advantages:

- there is no longer a need to stop and try again when a null character sequence is hit
- bad unicode sequences other than the null character will no longer result in a confusing traceback that includes the message about the null character

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/analytics/collectors.py
+++ b/awx/main/analytics/collectors.py
@@ -12,8 +12,6 @@ from django.contrib.sessions.models import Session
 from django.utils.timezone import now, timedelta
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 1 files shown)


jbradberry:
cc @Ladas

Ladas:
@jbradberry I believe we wrote tests for this last time? If it works then this looks great. :-) And I think it shouldn't affect analytics at all. :+1:

AlanCoding (approved):

PR #12255, opened 2022-05-19
Adding fields to job_metadata for workflows and approval nodes

john-westcott-iv:
##### SUMMARY

Adding extra vars for workflow approval notifications.
Adding the following for workflow templates:
  * inventory
  * limit
  * extra_vars

Added empty extra_vars to the content_stub.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.0.1.dev123+g6587e95783
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/models/workflow.py
+++ b/awx/main/models/workflow.py
@@ -659,6 +659,13 @@ class WorkflowJob(UnifiedJob, WorkflowJobOptions, SurveyJobMixin, JobNotificatio
                 node_job_description = 'job #{0}, "{1}", which finished with status {2}.'.format(node.job.id, node.job.name, node.job.status)
             str_arr.append("- node #{0} spawns {1}".format(node.id, node_job_description))
         result['body'] = '\n'.join(str_arr)
+        result.update(
+            dict(
+                inventory=self.inventory.name if self.inventory else None,
+                limit=self.limit,
+                extra_vars=self.display_extra_vars(),
+            )
+        )
         return result
 
     @property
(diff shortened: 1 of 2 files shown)


awx/main/models/workflow.py
```diff
@@ -885,3 +892,12 @@ def workflow_job_template(self):
     @property
     def workflow_job(self):
         return self.unified_job_node.workflow_job
+
+    def notification_data(self):
+        result = super(WorkflowApproval, self).notification_data()
+        result.update(
+            dict(
+                extra_vars=self.workflow_job.extra_vars,
```
AlanCoding: wait a minute, you used `self.display_extra_vars()` for workflow notification data, but then `self.workflow_job.extra_vars` for the approval node notification data. Shouldn't this be `self.workflow_job.display_extra_data()`?

john-westcott-iv: Good catch, we want display_ here since stuff could bleed through the workflow.

AlanCoding (approved):

AlanCoding:
This still might need some :brain: power for docs, but it seems to solve the ask.

We should first figure out what content we already have for notification templating rules for workflows & approval nodes. If that's scantly documented to begin with, communicating this could be a challenge.

awx/main/models/workflow.py
```diff
@@ -659,6 +659,13 @@ def notification_data(self):
                 node_job_description = 'job #{0}, "{1}", which finished with status {2}.'.format(node.job.id, node.job.name, node.job.status)
             str_arr.append("- node #{0} spawns {1}".format(node.id, node_job_description))
         result['body'] = '\n'.join(str_arr)
+        result.update(
```
jbradberry, line 662: `dict.update()` does take kwargs, so constructing another inner `dict` isn't necessary

jbradberry: As an aside, I want to advocate for our team to get away from the style of using manual `dict()` calls in favor of `dict` literals.  It's less flexible.

jbradberry (approved):

lucab85:
can we merge?

lucab85 (approved):

jbradberry:
I believe we can merge, yes.

Issue #12268, opened 2022-05-23
Support gitlab event status on other webhook events than Merge Requests


HampusLundqvist:
##### ISSUE TYPE
 - Support gitlab webhook events status callback for other events than Merge Request events.

##### SUMMARY

We would like to get feedback to gitlab in the master/main branch when we decide to Merge the MR into our master branch. 

Current support for only Merge Request events, which means when "A merge request is created, updated, merged, or closed, or a commit is added in the source branch."

We would like this integration to work when we decide only to merge to our protected main/master branch. 

In gitlab the webhook is configured as a push event webhook, with a filter containing the master/main branch.

I did a POC of this myself just now,  by editing:
 awx/api/views/webhooks.py

in  def get_event_status_api(self) function i deleted:
```
        if self.get_event_type() != 'Merge Request Hook':
            return
```

Restarted the services, and upon a merge of a MR into my master branch, the resulting job status showed up as expected in my gitlab main branch, with a working link to the job.

fosterseth:
@jbradberry might be interested in this

jbradberry:
@fosterseth I've been responding to the attached PR, #12271

jbradberry:
The attached PR has been merged.

PR #12271, opened 2022-05-23
return event_status on push, tag push, and merge gitlab webhook events

HampusLundqvist:
##### SUMMARY

Enables webhook event status to be visible in gitlab on e.g. push events so that we can view the status from the job launch in gitlab.

related: #12268 

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.0.1.dev130+gf02212b1fe

```

##### ADDITIONAL INFORMATION

Simply removed these two lines from webhook.py:
```
 if self.get_event_type() != 'Merge Request Hook':
            return
```
<!--- Paste verbatim command output below, e.g. before and after your change --

### DIFF
--- a/awx/api/views/webhooks.py
+++ b/awx/api/views/webhooks.py
@@ -204,8 +204,6 @@ class GitlabWebhookReceiver(WebhookReceiverBase):
         return h.hexdigest()
 
     def get_event_status_api(self):
-        if self.get_event_type() != 'Merge Request Hook':
-            return
         project = self.request.data.get('project', {})
         repo_url = project.get('web_url')
         if not repo_url:
--- a/awx/api/views/webhooks.py
+++ b/awx/api/views/webhooks.py
@@ -204,6 +204,8 @@ class GitlabWebhookReceiver(WebhookReceiverBase):
         return h.hexdigest()
 
     def get_event_status_api(self):
+        if self.get_event_type() not in self.ref_keys.keys():
+            return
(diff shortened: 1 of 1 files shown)


jbradberry (commented):
I am very hesitant to allow this with webhooks in the state they are currently in.  Currently there is no way to filter out events coming in to AWX, so everything that matches a job template is going to run a job.  Similarly, there is no machinery in place to allow event post-backs to be sent for some cases and not others.

Because of this, I think making this change would allow for situations where the Gitlab server would get spammed.

HampusLundqvist:
> I am very hesitant to allow this with webhooks in the state they are currently in. Currently there is no way to filter out events coming in to AWX, so everything that matches a job template is going to run a job. Similarly, there is no machinery in place to allow event post-backs to be sent for some cases and not others.
> 
> Because of this, I think making this change would allow for situations where the Gitlab server would get spammed.

Hi, thanks for your reply @jbradberry.

Sure, it makes sense to not do it for every webhook event type.

I have updated to to use the event types defined in the ref_keys variable
( ref_keys contains same as https://docs.gitlab.com/ee/integration/jenkins.html#configure-a-jenkins-integration-recommended  seems to support)

jbradberry (approved):
Since the only supported events for Gitlab are push, tag push, and merge, this change is probably ok.

fosterseth (approved):

Issue #12233, opened 2022-05-12
Error in job log output


ybucci:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Some outputs don't load

##### ENVIRONMENT
* AWX version: 21.0.0
* Operator version: 0.21.0
* Kubernetes version: 1.21.10
* AWX install method: Kubernetes

##### STEPS TO REPRODUCE

Run job

##### EXPECTED RESULTS

Correctly show the output log

##### ACTUAL RESULTS

Can't se logs without download them

##### ADDITIONAL INFORMATION

![image](https://user-images.githubusercontent.com/52977138/168065673-259049c9-28e8-4af2-9051-9d84016e3a38.png)

fosterseth:
possible duplicate of https://github.com/ansible/awx/issues/12194

fosterseth:
@ybucci 

do you mind sharing the first 10 events of your job output? please remove any sensitive information beforehand

also, can you share the output of `api/v2/jobs/2119/job_events/children_summary`

ybucci:
Hi @fosterseth 

**First 10 events:**

`No config file found; using defaults

PLAY [iis_sincro] **************************************************************

TASK [Gathering Facts] *********************************************************
ok: [OMITED]
<OMITED> Running win_updates - round 1

TASK [wu : Checking for Available Windows Updates] *****************************
ok: [OMITED] => {"changed": false, "failed_update_count": 0, "filtered_updates": {}, "found_update_count": 0, "installed_update_count": 0, "reboot_required": false, "updates": {}}

TASK [netscaler_disable_server : Call API Nitro DISABLE server PDVHavan in LoadBalancer] ***
skipping: [OMITED] => (item=OMITED)  => {"ansible_loop_var": "item", "changed": false, "item": "OMITED", "skip_reason": "Conditional result was False"}

TASK [netscaler_disable_server : Call API Nitro DISABLE server PDVHavan in LoadBalancer] ***
skipping: [OMITED] => (item=OMITED)  => {"ansible_loop_var": "item", "changed": false, "item": "OMITED", "skip_reason": "Conditional result was False"}

TASK [netscaler_disable_server : Call API Nitro DISABLE server Sistema in LoadBalancer] ***
skipping: [OMITED] => (item=OMITED)  => {"ansible_loop_var": "item", "changed": false, "item": "OMITED", "skip_reason": "Conditional result was False"}

[...]

Issue #12194, opened 2022-05-09
Missing job output sections


sooslaca:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

Not streaming logs, but when I open the output of the job later I have missing sections in the output.
`/#/jobs/playbook/XXXXXX/output`
I cannot find any correlation with the content and between missing sections. (eg not the same part missing even if I include the same role/task)
I cannot find console logs in developer tools, neither any network resources fail to load.
Attaching a screen showing the SAME include (line 95 and 112), but NOT same parts are missing.
Also weird that line numbers are missing next to those empty lines.

Has anyone have an idea how to debug it further or anyone else already affected by this issue?

### AWX version

AWX 21.0.0

### Select the relevant components

- [X] UI
- [ ] API
- [ ] Docs

### Installation method

minikube

### Modifications

no

### Ansible version

awx-ee:latest

### Operating system

linux

### Web browser

Chrome

### Steps to reproduce

Run playbook, I don't have a test playbook, just prod ones I cannot share.

### Expected results

No interruption in job output

### Actual results

Missing parts in job output

### Additional information

![awx_issue](https://user-images.githubusercontent.com/26524201/167415718-9c94cd23-63a9-49b9-877c-8777f41776f3.png)

keithjgrant:
Hi, a couple questions to help me narrow this down:

- Is it always the same lines missing if you refresh the page?
- Is it always the same lines missing if you re-run the same job?
- Does the issue resolve itself if you add any query parameters to the page URL? (e.g. `/jobs/playbook/{id}/output?x`) (this will disable the expand/collapse arrows on the left, which I suspect are part of the issue)

sooslaca:
Hi @keithjgrant ,

1) yes, same lines are missing
2) yes, same part is missing if I rerun the same job
3) and yes, this helped :) We can live with this for now.
As soon as I add a query parameter, arrows on the left are disappear and page loads fully !

Thanks for the hint ;) I hope it will be fixed sometimes...

lals1:
I am having exactly same issue with AWX 20.0.1. 

Adding query parameter in the URL does display all the lines in the job output.

keithjgrant:
[...]

PR #12287, opened 2022-05-26
detect if job events are tree-like and collapsible

fosterseth:
##### SUMMARY

related https://github.com/ansible/awx/issues/12194 and https://github.com/ansible/awx/issues/12233

shoutout to @ybucci for helping us discover the root of the bug!

if the ansible playbook is run with a strategy like free or host_pinned, then the events may no longer follow a tree-like hierarchy

e.g.
```
 E1
    E2
 E3
    E4  <- parent is E3
    E5  <- parent is E1
```
in the above, there is no clear way to collapse E1, because E5 comes after E3, which occurs after E1. Thus the tree view should be disabled.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API

##### AWX VERSION

```
21.0.1.dev144+g783e170e6a
```

### DIFF
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -3850,7 +3850,7 @@ class JobJobEventsChildrenSummary(APIView):
     meta_events = ('debug', 'verbose', 'warning', 'error', 'system_warning', 'deprecated')
 
     def get(self, request, **kwargs):
-        resp = dict(children_summary={}, meta_event_nested_uuid={}, event_processing_finished=False)
+        resp = dict(children_summary={}, meta_event_nested_uuid={}, event_processing_finished=False, is_tree=True)
         job = get_object_or_404(models.Job, pk=kwargs['pk'])
         if not job.event_processing_finished:
             return Response(resp)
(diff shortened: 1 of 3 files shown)


awx/api/views/__init__.py
```diff
@@ -3870,13 +3870,34 @@ def get(self, request, **kwargs):
...
+        # and return a response with is_tree: False
+        level_current_uuid = [None, None, None, None]
+
         prev_non_meta_event = events[0]
         for i, e in enumerate(events):
             if not e['event'] in JobJobEventsChildrenSummary.meta_events:
                 prev_non_meta_event = e
             if not e['uuid']:
                 continue
+            level_current_uuid[models.JobEvent.LEVEL_FOR_EVENT[e['event']]] = e['uuid']
```
keithjgrant: should we ensure lower-level uuids are set to None here? e.g. if the current event's level is 1, then`level_current_uuid[2] = None` and `level_current_uuid[3] = None` to prevent uuids of prior leaf nodes from lingering in the array

fosterseth: good point, changed

awx/api/views/__init__.py
```diff
@@ -3870,13 +3870,41 @@ def get(self, request, **kwargs):
...
+
         prev_non_meta_event = events[0]
         for i, e in enumerate(events):
             if not e['event'] in JobJobEventsChildrenSummary.meta_events:
                 prev_non_meta_event = e
             if not e['uuid']:
                 continue
+
+            if not e['event'] in JobJobEventsChildrenSummary.meta_events:
+                level = models.JobEvent.LEVEL_FOR_EVENT[e['event']]
```
AlanCoding, line 3897: I thought you said that `JobEvent.LEVEL_FOR_EVENT` was unreliable?

fosterseth: the event types we get back from ansible are reliable, but the ordering is not always preserved.

akus062381 (approved):

AlexSCorey (approved):
This looks good to me.

Issue #9363, opened 2021-02-19
Show host description attribute on more places


Commifreak:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Please add the Host description attribute to at least one (or all?) places:

* Job output: There, ansible only shows the IP address (if no real hostname given) - maybe not possible here but:
* Show the host description inside the host task result modal, next to its name:

![grafik](https://user-images.githubusercontent.com/87267/108463917-899b4700-727f-11eb-8abc-081346b53980.png)

That would be great :)

wenottingham:
The first of these cannot happen - it would be making an extra REST request for every line of playbook output, which really isn't practical.

The second of these could potentially be possible (although it's still an extra request when loading the event details.)

InnocentK:
Resolved

Issue #3348, opened 2019-03-01
Display Host description along with its name


Commifreak:
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME

 - UI

##### SUMMARY
In all views regarding selecting/displaying Hosts, the description field is not shown.
To be honest, I added all my hosts by IP, so the UI always displays only the IP when selecting hosts for example.

I put the hostname in the description field.

I dont have DNS entries for all hostname, thats the reason why I use IPs. I dont want to update my hosts file everytime I add a host.

Is this a good idea or a bad practice of mine?

AlexSCorey:
Host descriptions will be added in future releases.

lijok:
Ran into this yesterday as I was dealing with the same issue
What you can do is set the hostname as whatever, and add a variable "ansible_host" with the actual ip/hostname of the host

dsesami:
@mabashian question for you -- when I move the inventories/demo inventory/HOSTS page to half-width, the description vanishes. Is this intentional? Seems odd that the field vanishes when the page is at half width.

**Half width:**
![Screenshot from 2019-07-24 13-56-08](https://user-images.githubusercontent.com/19942822/61816694-e708b100-ae1a-11e9-8983-f3ef8c5f1fd2.png)

**Full width of same page:**
![Screenshot from 2019-07-24 13-56-24](https://user-images.githubusercontent.com/19942822/61816711-ec65fb80-ae1a-11e9-9cd4-e284e851dd1d.png)

(note that the description field vanishes with no way to see it)

dsesami:
followed up w/ @mabashian , this design is intentional. closing :+1: host description appears with names.

Commifreak:
Since my upgrade to 17.0.1 this is gone again:

![grafik](https://user-images.githubusercontent.com/87267/108367360-ac831800-71f9-11eb-96f2-f59d4b891933.png)

Any config option here?

Commifreak:
@mabashian @dsesami Can you confirm, that this issue is back since AWX 17? Thanks in advance!

AlexSCorey:
This is an issue in the new UI.  Reopening

VladoPortos:
Can confirm its gone :-/

Commifreak:
@mabashian or others: Any update on this?

InnocentK:
Related to the work in https://github.com/ansible/awx/issues/9363

InnocentK:
Resolved

PR #12292, opened 2022-05-27
Add host description in a couple of screens

nixocio:
Add host description in a couple of screens

See: https://github.com/ansible/awx/issues/3348
Also: https://github.com/ansible/awx/issues/9363

image

image


awx/ui/src/components/Lookup/HostFilterLookup.js
```diff
@@ -419,6 +419,7 @@ function HostFilterLookup({
             headerRow={
               <HeaderRow qsConfig={QS_CONFIG} isSelectable={false}>
                 <HeaderCell sortKey="name">{t`Name`}</HeaderCell>
+                <HeaderCell sortKey="description">{t`Description`}</HeaderCell>
```
AlexSCorey, line 422: When a user sorts by description does it just sort alphabetically by the first letter in the description?

AlexSCorey (approved):
This looks good to me.  Awesome job updating/adding tests.

awx/ui/src/components/Lookup/HostFilterLookup.js
nixocio: Exactly

InnocentK (approved):
Looks good!

Issue #12171, opened 2022-05-04
[Metrics&Observability]  Track (recent/sliding window) event processing lag in metrics


kdelee:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

The amount of time between when events are created in ansible and they get written to the database is a measure of how far behind the callback receiver is. Monitoring this metric can give administrators insight if they need to horizontally scale the control plane to get more event processing power/spread jobs over more control instances (right now we do not dynamically scale the callback receiver workers, there are always 4 per control instance).

I think there are two ways to do this. One is to query the database, relying on the fact that this should now be less atrocious since the events table is now partitioned. The other is to hook into what the subsystem metrics does.

### Option 1: query database

In this case we would add it to the "traditional" metrics  in awx/main/analytics/metrics.py or awx/main/analytics/collectors.py where we currently populate most of the "normal" metrics in we can add a function like:

```
# we need to import this above
from django.db.models import Avg, ExpressionWrapper, DurationField, F

def avg_event_processing_delta(lookback_seconds=60, event_limit=100):
    since = datetime.now() - timedelta(seconds=lookback_seconds)
    avg_delta = (
        JobEvent.objects.annotate(delta=ExpressionWrapper(F('modified') - F('created'), output_field=DurationField()))
        .filter(modified__gte=since)
        .only('modified', 'created')[:event_limit]
        .aggregate(Avg('delta'))
    )
    return avg_delta['delta__avg']
```

### Option 2: hook into subsystem metrics
I'm a bit more hazy on this one. But i think it would look like calculating the average delta/lag in each batch of events like we do with 
https://github.com/ansible/awx/blob/e3fe9010b73e613646be91f1618d4d1806ae1431/awx/main/dispatch/worker/callback.py#L145
which we just increment during this loop where we also actually compute `modified` -- so this would be a great time to just keep a running average of the delta between modified and created.
https://github.com/ansible/awx/blob/e3fe9010b73e613646be91f1618d4d1806ae1431/awx/main/dispatch/worker/callback.py#L149-L155
Right there we could capture the delta between the two.

Then we could bubble it up in the same way we do the other metrics
https://github.com/ansible/awx/blob/e3fe9010b73e613646be91f1618d4d1806ae1431/awx/main/dispatch/worker/callback.py#L179-L188

@fosterseth would be probably be a good resource on figuring out how to fit into this pattern

fosterseth:
[...]

PR #12308, opened 2022-06-01
Metrics for callback receiver job event lag

rebeccahhh:
##### SUMMARY

Spawning out of https://github.com/ansible/awx/issues/12171

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - API/Metrics

##### AWX VERSION

```
awx: 21.0.1
```

##### Additional Information
view of Grafana graph example:
callback_event_processing_seconds

### DIFF
--- a/awx/main/analytics/subsystem_metrics.py
+++ b/awx/main/analytics/subsystem_metrics.py
@@ -34,7 +34,7 @@ class BaseM:
     def to_prometheus(self, instance_data):
         output_text = f"# HELP {self.field} {self.help_text}\n# TYPE {self.field} gauge\n"
         for instance in instance_data:
-            output_text += f'{self.field}{{node="{instance}"}} {instance_data[instance][self.field]}\n'
+            output_text += f'{self.field}{{node="{instance}"}} {instance_data[instance].get(self.field, -1)}\n'  # TODO: fix because this -1 is neccessary when dealing with old instances (ex. you didn't clean up your database)
         return output_text
 
 
@@ -161,7 +161,7 @@ class Metrics:
             IntM('callback_receiver_events_popped_redis', 'Number of events popped from redis'),
             IntM('callback_receiver_events_in_memory', 'Current number of events in memory (in transfer from redis to db)'),
             IntM('callback_receiver_batch_events_errors', 'Number of times batch insertion failed'),
-            FloatM('callback_receiver_events_insert_db_seconds', 'Time spent saving events to database'),
+            FloatM('callback_receiver_events_insert_db_seconds', 'Total time spent saving events to database'),
             IntM('callback_receiver_events_insert_db', 'Number of events batch inserted into database'),
             IntM('callback_receiver_events_broadcast', 'Number of events broadcast to other control plane nodes'),
             HistogramM(
@@ -170,6 +170,7 @@ class Metrics:
             FloatM('subsystem_metrics_pipe_execute_seconds', 'Time spent saving metrics to redis'),
             IntM('subsystem_metrics_pipe_execute_calls', 'Number of calls to pipe_execute'),
             FloatM('subsystem_metrics_send_metrics_seconds', 'Time spent sending metrics to other nodes'),
+            SetFloatM('callback_receiver_event_processing_avg_seconds', 'Average processing time per event per callback receiver batch'),
         ]
         # turn metric list into dictionary with the metric name as a key
         self.METRICS = {}
@@ -179,9 +180,11 @@ class Metrics:
         # track last time metrics were sent to other nodes
         self.previous_send_metrics = SetFloatM('send_metrics_time', 'Timestamp of previous send_metrics call')
 
-    def clear_values(self):
-        for m in self.METRICS.values():
-            m.clear_value(self.conn)
+    def clear_values(self, fields=None):
+        if not fields:
+            fields = self.METRICS.keys()
+        for m in fields:
+            self.METRICS[m].clear_value(self.conn)
         self.metrics_have_changed = True
         self.conn.delete(root_key + "_lock")
 
--- a/awx/main/dispatch/worker/callback.py
+++ b/awx/main/dispatch/worker/callback.py
@@ -4,6 +4,7 @@ import os
 import signal
 import time
 import traceback
+import datetime
 
 from django.conf import settings
 from django.utils.functional import cached_property
@@ -151,12 +152,17 @@ class CallbackBrokerWorker(BaseWorker):
             metrics_singular_events_saved = 0
             metrics_events_batch_save_errors = 0
             metrics_events_broadcast = 0
+            metrics_events_missing_created = 0
+            metrics_total_job_event_processing_seconds = datetime.timedelta(seconds=0)
             for cls, events in self.buff.items():
                 logger.debug(f'{cls.__name__}.objects.bulk_create({len(events)})')
                 for e in events:
+                    e.modified = now  # this can be set before created because now is set above on line 149
                     if not e.created:
                         e.created = now
-                    e.modified = now
+                        metrics_events_missing_created += 1
+                    else:  # only calculate the seconds if the created time already has been set
+                        metrics_total_job_event_processing_seconds += e.modified - e.created
                 metrics_duration_to_save = time.perf_counter()
                 try:
                     cls.objects.bulk_create(events)
@@ -189,6 +195,11 @@ class CallbackBrokerWorker(BaseWorker):
                 self.subsystem_metrics.observe('callback_receiver_batch_events_insert_db', metrics_bulk_events_saved)
                 self.subsystem_metrics.inc('callback_receiver_events_in_memory', -(metrics_bulk_events_saved + metrics_singular_events_saved))
                 self.subsystem_metrics.inc('callback_receiver_events_broadcast', metrics_events_broadcast)
+                self.subsystem_metrics.set(
+                    'callback_receiver_event_processing_avg_seconds',
+                    metrics_total_job_event_processing_seconds.total_seconds()
+                    / (metrics_bulk_events_saved + metrics_singular_events_saved - metrics_events_missing_created),
+                )
             if self.subsystem_metrics.should_pipe_execute() is True:
                 self.subsystem_metrics.pipe_execute()
 


awx/main/analytics/subsystem_metrics.py
```diff
@@ -170,6 +170,7 @@ def __init__(self, auto_pipe_execute=True, instance_name=None):
             FloatM('subsystem_metrics_pipe_execute_seconds', 'Time spent saving metrics to redis'),
             IntM('subsystem_metrics_pipe_execute_calls', 'Number of calls to pipe_execute'),
             FloatM('subsystem_metrics_send_metrics_seconds', 'Time spent sending metrics to other nodes'),
+            FloatM('callback_receiver_event_processing_avg_seconds', 'Time delay of events being created and then saved to the database'),
```
kdelee: "Average processing time per event per callback receiver batch"

awx/main/dispatch/worker/callback.py
```diff
@@ -151,12 +151,14 @@ def flush(self, force=False):
             metrics_singular_events_saved = 0
             metrics_events_batch_save_errors = 0
             metrics_events_broadcast = 0
+            metrics_job_lag = 0
```
kdelee: probably want a `metrics_total_job_event_processing_seconds` that we add to as we loop over events

awx/main/dispatch/worker/callback.py
```diff
@@ -151,12 +151,14 @@ def flush(self, force=False):
...
             for cls, events in self.buff.items():
                 logger.debug(f'{cls.__name__}.objects.bulk_create({len(events)})')
                 for e in events:
                     if not e.created:
                         e.created = now
                     e.modified = now
+                    # metrics_job_lag = e.modified - e.created  # get the delta between modified and created. TODO: currently being overwritten
```
kdelee: we're going to want to `+=` add to `metrics_total_job_event_processing_seconds` here -- may need wrangle datetime timedelta to seconds in float

awx/main/dispatch/worker/callback.py
```diff
@@ -182,13 +184,15 @@ def flush(self, force=False):
             self.buff = {}
             self.last_flush = time.time()
             # only update metrics if we saved events
+            metrics_job_lag = 1  # TODO: currently a placeholder value because currently it is doing wacky stuff, a la maybe it's incrementing?
             if (metrics_bulk_events_saved + metrics_singular_events_saved) > 0:
```
kdelee: when we compute average probably safer to do it in this if block to avoid division by zero
```suggestion
            if (metrics_bulk_events_saved + metrics_singular_events_saved) > 0:
                metrics_avg_event_processing_seconds = metrics_total_job_event_processing_seconds / (metrics_bulk_events_saved + metrics_singular_events_saved)
```

awx/main/analytics/subsystem_metrics.py
```diff
@@ -34,7 +34,7 @@ def decode(self, conn):
     def to_prometheus(self, instance_data):
         output_text = f"# HELP {self.field} {self.help_text}\n# TYPE {self.field} gauge\n"
         for instance in instance_data:
-            output_text += f'{self.field}{{node="{instance}"}} {instance_data[instance][self.field]}\n'
+            output_text += f'{self.field}{{node="{instance}"}} {instance_data[instance].get(self.field, -1)}\n'  # TODO: fix because this -1 is neccessary when dealing with old instances (ex. you didn't clean up your database)
```
rebeccahhh, line 37: @fosterseth do you have any idea why this would have to be? I had some old instances around that necessitated adding a default because they didn't have the new key (ie they didn't exist and weren't running but still got caught in this).

awx/main/dispatch/worker/callback.py
```diff
@@ -189,6 +192,12 @@ def flush(self, force=False):
                 self.subsystem_metrics.observe('callback_receiver_batch_events_insert_db', metrics_bulk_events_saved)
                 self.subsystem_metrics.inc('callback_receiver_events_in_memory', -(metrics_bulk_events_saved + metrics_singular_events_saved))
                 self.subsystem_metrics.inc('callback_receiver_events_broadcast', metrics_events_broadcast)
+                self.subsystem_metrics.set(
+                    'callback_receiver_event_processing_avg_seconds',
+                    metrics_total_job_event_processing_seconds.total_seconds() / (metrics_bulk_events_saved + metrics_singular_events_saved),
+                )  # (metrics_total_job_event_processing_seconds )
```
kdelee: drop old comment

awx/main/dispatch/worker/callback.py
```diff
@@ -189,6 +192,12 @@ def flush(self, force=False):
...
+            else:
+                self.subsystem_metrics.set('callback_receiver_event_processing_avg_seconds', 0)  # if there is no lag the avg is 0.
```
kdelee:
```suggestion
                self.subsystem_metrics.set('callback_receiver_event_processing_avg_seconds', 0)  # if we wrote no events the the avg processing time is 0.
```

awx/main/dispatch/worker/callback.py
```diff
@@ -151,12 +152,14 @@ def flush(self, force=False):
...
             metrics_events_batch_save_errors = 0
             metrics_events_broadcast = 0
+            metrics_total_job_event_processing_seconds = datetime.timedelta(seconds=0)
             for cls, events in self.buff.items():
                 logger.debug(f'{cls.__name__}.objects.bulk_create({len(events)})')
                 for e in events:
                     if not e.created:
                         e.created = now
                     e.modified = now
+                    metrics_total_job_event_processing_seconds += e.modified - e.created
```
AlanCoding: I'd rather this not be saved if the `if not e.created:` above is hit. In that case, you would erroneously record no lag, which isn't theoretically possible. I realize that it would mess up your denominator, so you'd have to track another thing.

awx/main/dispatch/worker/callback.py
AlanCoding: I'm only just now catching up on much of what subsystem metrics is doing, and I am very worried that it's saving too many things (already, even before this). I don't agree that 0 is representative of this situation, it's more akin to `0/0`. I would prefer to not write in the `else:` condition here, and let the old value hang out.

kdelee: k my concern is it would never go back down again on a quiet system. but I do see where you're coming from and I'm not against dropping this condition because downside of this is we may have some idle callback receivers pushing this metric while others are busy

awx/main/dispatch/worker/callback.py
kdelee: something like
```
                    e.modified = now
                    if not e.created:
                        e.created = now
                    else:
                        metrics_total_job_event_processing_seconds += e.modified - e.created
```
?

awx/main/dispatch/worker/callback.py
kdelee: so +1 to dropping whole else condition

rebeccahhh: I'll throw it through some manual testing without the else, should work just fine.

awx/main/analytics/subsystem_metrics.py
```diff
@@ -179,9 +180,11 @@ def __init__(self, auto_pipe_execute=True, instance_name=None):
         # track last time metrics were sent to other nodes
         self.previous_send_metrics = SetFloatM('send_metrics_time', 'Timestamp of previous send_metrics call')
 
-    def clear_values(self):
-        for m in self.METRICS.values():
-            m.clear_value(self.conn)
+    def clear_values(self, fields=None):
+        if not fields:
+            fields = self.METRICS.keys()
```
shanemcd, line 185: This seems a little weird. What's the reasoning behind this conditional? Might be worth a comment.

awx/main/dispatch/worker/callback.py
AlanCoding: To articulate more specifically what @kdelee is thinking:

I expect that performance testing will launch a large number of jobs all at the same time. These will finish significantly faster than the event processing can finish, because that's by design. Then at the time the last event is saved, the event lag will be high. At the time of processing, the redis queue is effectively empty, but the measured event lag will be high. Since it's the last event of a long burst, the callback system will be dead silent after that.

In any case, the graph would be discontinuous. If you do what I'm asking for here, it will remain at a relatively high value after the burst has finished, and drop down to a low (not zero) value upon the first events of the next job that starts. With what this was doing, it would drop it down to zero right after the burst finishes processing.

awx/main/analytics/subsystem_metrics.py
rebeccahhh: I may or may not have stolen this from @fosterseth 's open PR 😅 
What I _believe_ is that it's resetting the values at the end of each partition in preparation for the next cycle.

awx/main/dispatch/worker/callback.py
rebeccahhh: I dislike the idea that modified could ever be _before_ created (even if only by a couple milliseconds), but the intent I definitely get behind.

kdelee: we cleared up that now() allready got called so they are they same

awx/main/dispatch/worker/callback.py
```diff
@@ -189,6 +195,11 @@ def flush(self, force=False):
                 self.subsystem_metrics.observe('callback_receiver_batch_events_insert_db', metrics_bulk_events_saved)
                 self.subsystem_metrics.inc('callback_receiver_events_in_memory', -(metrics_bulk_events_saved + metrics_singular_events_saved))
                 self.subsystem_metrics.inc('callback_receiver_events_broadcast', metrics_events_broadcast)
+                self.subsystem_metrics.set(
+                    'callback_receiver_event_processing_avg_seconds',
+                    metrics_total_job_event_processing_seconds.total_seconds()
+                    / (metrics_bulk_events_saved + metrics_singular_events_saved - metrics_events_missing_created),
+                )
             if self.subsystem_metrics.should_pipe_execute() is True:
```
kdelee, line 203: @rebeccahhh @AlanCoding 
...so rebeccah sent me examples of how this looks now that we dropped the `else` condition and we never zero out the metric.
image
image (1)





TBH I think it is much worse, and would make it difficult/harder to use in a sane way because it does exactly what alan suggested which is leave it pegged up high. Seems like there was some concern that hypothetically having the else condition would cause this situation where the the metric would oscillate between zero and some number because there would be one idle worker while others were posting a value. We never actually have seen that happen though. So to me, its the lesser of the evils because with leaving the metric up high you impede usefulness of any computation of avg, mean, and stdev. And if we later do what @AlanCoding wants to to which is have the parent process be the one that really pushes the metrics, this won't be a concern at all.

rebeccahhh: to further add onto this, this will actually never zero out (I tested it overnight, lag hovered just above 1 second for the whole night, having taken the average from many hours earlier and running with it and it's lack of change) which I find further problematic in the short term. I added the else statement back on because as @kdelee mentioned, this won't be an issue at all in a future PR. I left a comment so we know to adjust that section of the code when the time comes.

rebeccahhh: Here's a visual example where you can see from yesterday afternoon until about 8am it never changes. It's just not useful information at that point in my opinion. Since we're taking an average I would rather have some zeros included when appropriate vs. this situation.
image

awx/main/dispatch/worker/callback.py
rebeccahhh: @AlanCoding We left a couple more comments below but I wanted to leave on here too. I would prefer in the short term we keep the else and actually drop down since in the future we'll be changing how this behaves anyways. I think it has a better delivery of meaningful information with the included else block but it's not a deal breaker for me if you think we should drop it in order to get this in.

awx/main/dispatch/worker/callback.py
AlanCoding: Compromising _performance itself_ for the visuals of performance metrics is like throwing the baby out with the bathwater. The current `else:` strikes me as a categorically bad decision.

If you wish to set it back down to 0, then you should not continue to update it to that value on every flush. This wouldn't be a problem if the subsystem metrics utils, themselves, compared new values to old values and avoided unnecessary system calls when nothing changed. That doesn't appear to be what it does.

https://github.com/ansible/awx/blob/d9f26a411ecd893725372b93b9698b017a40b739/awx/main/analytics/subsystem_metrics.py#L195-L197

If you still wish to set it down to 0 in this event, you should track a variable on `self` to remember whether or not the last value is already 0, and don't keep updating it if it is.

rebeccahhh: fair enough ^
As I said I'm not staunchly committed to this so I'll pull the else out and when we do an overhaul of the rest of this we'll just know some changes surrounding how this graph behaves may or may not come with that.
cc @kdelee

awx/main/analytics/subsystem_metrics.py
AlanCoding: I think the best thing to do would be to revert the changes here before merge. As it stands, it's like adding mystery code.

rebeccahhh: I have validated that without this the average seconds just continues to rise. vs. being refreshed with the partitionining. I'll leave it in so this is functional but full credit to @fosterseth for it. Grafana of jobs over approximately 25m to demonstrate if it's not included.
image

rebeccahhh:
did a little rebase to get this into one commit it's passed internal tests on my end as well.

awx/main/analytics/subsystem_metrics.py
AlanCoding: oh, I didn't realize it was fixing something real. I guess it's in the nuance of how it interacts with the python library?

AlanCoding (approved):

PR #12312, opened 2022-06-02
Update logout/login redirect for different users

nixocio:
* Logout as User A and Login as User B redirects to `/home`
* Logout as User A and Login as User A redirects to `/home`
* Allow session to timeout as User A and Login as User A redirects to User A's last location

In order to test this change session to expire in `Idle Time Force Log Out` to be 61s. Create 2 users, A and B.

image

See: https://github.com/ansible/awx/issues/11167

### DIFF
--- a/awx/ui/src/screens/Login/Login.js
+++ b/awx/ui/src/screens/Login/Login.js
@@ -1,5 +1,5 @@
 /* eslint-disable react/jsx-no-useless-fragment */
-import React, { useCallback, useEffect } from 'react';
+import React, { useCallback, useEffect, useRef } from 'react';
 import { Redirect, withRouter } from 'react-router-dom';
 
 import { t } from '@lingui/macro';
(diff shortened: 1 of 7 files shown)


awx/ui/src/util/auth.js
```diff
@@ -6,3 +6,34 @@ export function isAuthenticated(cookie) {
...
+        );
+        break;
+      }
+    }
+  }
+  return userId;
+}
+
+function parseUserId(decodedUserData) {
+  const items = decodedUserData.split(',');
```
keithjgrant: The decodedUserData from the cookie is just JSON. you should be able to do

```
const userData = JSON.parse(decodedUserData);
return userData.id;
```

awx/ui/src/screens/Login/Login.js
```diff
@@ -112,8 +115,29 @@ function AWXLogin({ alt, isAuthenticated }) {
...
+        window.sessionStorage.getItem(SESSION_USER_ID)
+      );
+      if (previousUserId === null) {
+        return true;
+      }
+      return currentUserId.toString() !== previousUserId.toString();
+    };
+    isNewUser.current = verifyIsNewUser();
+    hasVerifiedUser.current = true;
+    window.sessionStorage.setItem(
```
keithjgrant: this is only storing the numerical user id, right? we definitely don't want to store the username into session storage where it can be easily accessed

nixocio: yes, just user id.

awx/ui/src/util/auth.js
nixocio: hey @keithjgrant, great suggestion. Thanks.

akus062381 (approved):
looks great

keithjgrant (approved):

PR #12313, opened 2022-06-02
Only do substitutions for container path conversions with resolved paths

AlanCoding:
##### SUMMARY
We have a report that, on some systems, there is a symlink `/var/run -> ../run`

I've added tests here that create a symlink for case-in-point like `/tmp/dst -> /tmp/src`

Before this change, running jobs with `AWX_ISOLATION_BASE_PATH` set to a symlink of this pattern would error like:

```
  File "/awx_devel/awx/main/tasks/jobs.py", line 469, in run
  File "/awx_devel/awx/main/models/credential/__init__.py", line 507, in inject_credential
    container_path = to_container_path(path, private_data_dir)
  File "/awx_devel/awx/main/utils/execution_environments.py", line 62, in to_container_path
    raise RuntimeError(f'Cannot convert path {path} unless it is a subdir of {private_data_dir}')
RuntimeError: Cannot convert path /tmp/foobar/awx_117_j_ds0t3g/env/tmpw093dey9 unless it is a subdir of /tmp/foobar/awx_117_j_ds0t3g
```

After this change, that works fine.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/tests/unit/test_tasks.py
+++ b/awx/main/tests/unit/test_tasks.py
@@ -4,6 +4,7 @@ import json
 import os
 import shutil
 import tempfile
+from pathlib import Path
 
 import fcntl
 from unittest import mock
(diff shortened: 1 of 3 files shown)


AlanCoding:
I'm pretty happy with this patch. It takes some thought, but this code seems clearly safer.

I moved the method `to_host_path` because it was only used in the tests. If our method to convert from host->container path needs updating for these reason, then it would stand to reason this method would need updating too. But we don't use it except for in the tests, so it would be pointless to do so.

jbradberry (approved):

PR #12320, opened 2022-06-03
Pre-fill project for job template from query params

nixocio:
Pre-fill project when creating JT from Project -> Job Templates
List

image


AlexSCorey (approved):
This looks good to me.  It works as expected and fails if wrong params are passed into the url as expected.  Well done!

jay-steurer:
Tested and Approved

PR #12323, opened 2022-06-03
Removes Sanatize html in favor of dom purify library

AlexSCorey:
##### SUMMARY
After some investigation it was determined that a user could not customize their login page footer due to some issues with Sanitize HTML library.  It turns out that library is really meant for projects where the UI is server side rendered and therefore was not able to work properly in our project.  I decided to remove that library and replace it with [DomPurify](https://github.com/cure53/DOMPurify), a well supported, active alternative.  
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME

 - UI

##### ADDITIONAL INFORMATION
Screen Shot 2022-06-03 at 2 38 41 PM


awx/ui/src/screens/Login/Login.js
```diff
@@ -5,7 +5,8 @@ import { Redirect, withRouter } from 'react-router-dom';
 import { t } from '@lingui/macro';
 import { Formik } from 'formik';
 import styled from 'styled-components';
-import sanitizeHtml from 'sanitize-html';
+import DOMPurify from 'dompurify';
+// import sanitizeHtml from 'sanitize-html';
```
kialam: remove comment

kialam (approved):
👍

akus062381 (approved):

PR #12331, opened 2022-06-06
Allow worker files to be loaded as blob objects.

### LINKED ISSUE #10192: worker-json 404 and various warnings on job settings page
(from @tiagodread, which we found while addressing what ended up being an unrelated issue with saves)

> 
> When the edit page is loaded for Job settings, there are some errors:
> 
> Warning: Failed prop type: Invalid prop `groupTitle` of type `object` supplied to `NavExpandableGroup`, expected `string`.
> 
> Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of FindRefWrapper which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://fb.me/react-strict-mode-find-node
> 
> Warning: Failed prop type: The prop `config` is marked as required in `BooleanField`, but its value is `undefined`.
> 
> Warning: Failed prop type: The prop `config` is marked as required in `InputField`, but its value is `undefined`.

Also, when loading the page there's a `worker-json` 404 error that we need to remove. see: https://github.com/ansible/awx/issues/10123#issuecomment-836303358

kialam:
##### SUMMARY

Fixes #10192 

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - UI

### DIFF
--- a/awx/ui/public/index.html
+++ b/awx/ui/public/index.html
@@ -24,7 +24,7 @@
       </script>
       <meta
         http-equiv="Content-Security-Policy"
-        content="default-src 'self'; connect-src 'self' ws: wss:; style-src 'self' 'unsafe-inline'; script-src 'self' 'nonce-{{ csp_nonce }}' *.pendo.io https://d3js.org; img-src 'self' *.pendo.io data:; worker-src 'self';"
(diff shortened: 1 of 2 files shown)


AlexSCorey (approved):
Looks good to me

jay-steurer:
tested and approved

PR #12335, opened 2022-06-07
import/export error codes when something bad happens

sarabrajsingh:
a simple boolean flag in the ApiV2 Object gets flipped to `True` if a `log.error()` event was triggered, and this bubbles back up to the calling main() function as a thrown exception in the form `raise SystemExit(0 if not error else 1)`.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - CLI

##### AWX VERSION

```
awx: 21.0.1.dev154+gbfc8989c1a.d20220607
```

### DIFF
--- a/awxkit/awxkit/api/pages/api.py
+++ b/awxkit/awxkit/api/pages/api.py
@@ -90,6 +90,7 @@ class ApiV2(base.Base):
             return None
         if post_fields is None:  # Deprecated endpoint or insufficient permissions
             log.error("Object export failed: %s", _page.endpoint)
+            self._has_error = True
             return None
 
         # Note: doing _page[key] automatically parses json blob strings, which can be a problem.
@@ -110,6 +111,7 @@ class ApiV2(base.Base):
                             pass
                     if resource is None:
                         log.error("Unable to infer endpoint for %r on %s.", key, _page.endpoint)
+                        self._has_error = True
                         continue
                     related = self._filtered_list(resource, _page.json[key]).results[0]
                 else:
@@ -119,12 +121,14 @@ class ApiV2(base.Base):
             if rel_endpoint is None:  # This foreign key is unreadable
                 if post_fields[key].get('required'):
                     log.error("Foreign key %r export failed for object %s.", key, _page.endpoint)
+                    self._has_error = True
                     return None
                 log.warning("Foreign key %r export failed for object %s, setting to null", key, _page.endpoint)
                 continue
             rel_natural_key = rel_endpoint.get_natural_key(self._cache)
             if rel_natural_key is None:
                 log.error("Unable to construct a natural key for foreign key %r of object %s.", key, _page.endpoint)
+                self._has_error = True
                 return None  # This foreign key has unresolvable dependencies
             fields[key] = rel_natural_key
 
@@ -169,6 +173,7 @@ class ApiV2(base.Base):
         natural_key = _page.get_natural_key(self._cache)
         if natural_key is None:
             log.error("Unable to construct a natural key for object %s.", _page.endpoint)
+            self._has_error = True
             return None
         fields['natural_key'] = natural_key
 
@@ -260,6 +265,7 @@ class ApiV2(base.Base):
             except (exc.Common, AssertionError) as e:
                 identifier = asset.get("name", None) or asset.get("username", None) or asset.get("hostname", None)
                 log.error(f"{endpoint} \"{identifier}\": {e}.")
+                self._has_error = True
                 log.debug("post_data: %r", post_data)
                 continue
 
@@ -294,6 +300,7 @@ class ApiV2(base.Base):
             pass
         except exc.Common as e:
             log.error("Role assignment failed: %s.", e)
+            self._has_error = True
             log.debug("post_data: %r", {'id': role_page['id']})
 
     def _assign_membership(self):
@@ -324,17 +331,21 @@ class ApiV2(base.Base):
                 for item in related_set:
                     rel_page = self._cache.get_by_natural_key(item)
                     if rel_page is None:
-                        continue  # FIXME
+                        log.error("Could not find matching object in Tower for imported relation, item: %r", item)
+                        self._has_error = True
+                        continue
                     if rel_page['id'] in existing:
                         continue
                     try:
                         post_data = {'id': rel_page['id']}
                         endpoint.post(post_data)
                         log.error("endpoint: %s, id: %s", endpoint.endpoint, rel_page['id'])
+                        self._has_error = True
                     except exc.NoContent:  # desired exception on successful (dis)association
                         pass
                     except exc.Common as e:
                         log.error("Object association failed: %s.", e)
+                        self._has_error = True
                         log.debug("post_data: %r", post_data)
             else:  # It is a create set
                 self._cache.get_page(endpoint)
--- a/awxkit/awxkit/cli/resource.py
+++ b/awxkit/awxkit/cli/resource.py
@@ -144,6 +144,8 @@ class Import(CustomCommand):
         client.authenticate()
         client.v2.import_assets(data)
 
+        self._has_error = getattr(client.v2, '_has_error', False)
+
         return {}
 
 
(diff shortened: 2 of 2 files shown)


sarabrajsingh:
to do - unit tests :)

awxkit/awxkit/api/pages/api.py
```diff
@@ -355,7 +367,10 @@ def import_assets(self, data):
         self._assign_membership()
         self._assign_roles()
 
-        return changed
+        return changed, self.get_error()
+
+    def get_error(self):
```
jbradberry: I don't think that this is really needed.  It would have been ok to make the attribute semi-private (`self._has_error`) and just use it directly, in my opinion.

awxkit/awxkit/api/pages/api.py
```diff
@@ -325,10 +335,12 @@ def _assign_related(self):
                         post_data = {'id': rel_page['id']}
                         endpoint.post(post_data)
                         log.error("endpoint: %s, id: %s", endpoint.endpoint, rel_page['id'])
+                        self.__has_error = True
```
jbradberry: Dropping this comment here because the actual line is currently out of scope...

There is a line a few lines above here with a bare `FIXME` comment.  Not having a `log.error()` call there is the cause of the problem tracked in https://github.com/ansible/tower/issues/5889.  That line will eventually need a log call plus your error attribute, and it probably makes sense to just fold that change in here in your PR.

awxkit/awxkit/api/pages/api.py
```diff
@@ -355,7 +367,10 @@ def import_assets(self, data):
         self._assign_membership()
         self._assign_roles()
 
-        return changed
+        return changed, self.get_error()
```
jbradberry: Other code directly makes use of this method (see awx_collection/plugins/modules/import.py).  Changing the output is going to mess with those other callers, some of which we can't anticipate.  I haven't 100% thought this through yet, but I suspect what we actually want to do is to raise a `SystemExit` or do a `sys.exit` call right above here.  YMMV.

jbradberry: Ditto with other places you changed the return signature to include the error bool, most likely.

jbradberry (changes requested):

awxkit/awxkit/api/pages/api.py
sarabrajsingh: what should the error message read?

`log.error("failed to get natural key from cache: %s, id: %s", endpoint.endpoint, rel_page['id'])` ?

awxkit/awxkit/api/pages/api.py
sarabrajsingh: this seems to be safe and is the equivalent of doing

r, _ = command.handle(client, parser)

so as long as the ordering of the return statement in `handle` is preserved, the other files consuming it (like `awx_collection/plugins/modules/import.py`) should be fine. again another case where unit testing would work well.

awxkit/awxkit/api/pages/api.py
sarabrajsingh: just had a thought - our integration tests in internal jenkins which have a collections test-set should pick up and possible errors with `import.py` in the collections

jbradberry: > what should the error message read?
> 
> `log.error("failed to get natural key from cache: %s, id: %s", endpoint.endpoint, rel_page['id'])` ?

No, the error messaging around natural keys seems to be confusing for users, and in this case there is a more straightforward explanation -- there doesn't seem to be a matching object on this Tower instance for the imported relation.

awxkit/awxkit/api/pages/api.py
jbradberry: I'm saying this is
- unnecessary
- and will involve **all** consuming code having to be adjusted from getting a single returned object to needing to destructure a tuple

Please don't do this.  Just raise a `SystemExit` inside this method and the `export_assets` method.

jbradberry: To be clearer, the awx collection directly uses the `import_assets` and `export_assets` methods, without going through the `CustomCommand` interfaces provided for awx-cli.  We do not want to break the contract for these methods for applications and other consumers that make use of awxkit _as a library_, instead of invoking the CLI wrapper.

awxkit/awxkit/api/pages/api.py
```diff
@@ -318,17 +331,21 @@ def _assign_related(self):
                 for item in related_set:
                     rel_page = self._cache.get_by_natural_key(item)
                     if rel_page is None:
-                        continue  # FIXME
+                        log.error("Could not find matching object in Tower for imported relation")
```
jbradberry: I think we should add information about what the attempted match was.

awxkit/awxkit/api/pages/api.py
```diff
@@ -355,6 +372,9 @@ def import_assets(self, data):
         self._assign_membership()
         self._assign_roles()
 
+        if self._has_error:
+            raise SystemExit(1)
```
jbradberry: On second thought, we probably don't want to use an exception type that doesn't derive from `Exception` here.  That would bypass most try/except blocks by consumers of the library.

awxkit/awxkit/api/pages/api.py
```diff
@@ -80,6 +80,9 @@ class Api(base.Base):
 
 
 class ApiV2(base.Base):
+    def __init__(self, connection=None, *a, **kw):
+        super().__init__(connection, *a, **kw)
+        self._has_error = False
```
jbradberry: Since this attribute is only used for `import_assets` and `export_assets`, I'd recommend moving setting this there instead of adding this overridden `__init__`

awxkit/awxkit/api/pages/api.py
jbradberry: Doing a simple `raise Exception("oops!")` at the top of `export_assets`, I'm able to trace that `awxkit/cli/__init__.py` line 25 is covered by a try/except that ultimately does a `sys.exit(1)`.

```
  File "/home/jrb/ansible/awx/awxkit/awxkit/cli/__init__.py", line 25, in run
    cli.parse_resource()
  File "/home/jrb/ansible/awx/awxkit/awxkit/cli/client.py", line 152, in parse_resource
    self.resource = parse_resource(self, skip_deprecated=skip_deprecated)
  File "/home/jrb/ansible/awx/awxkit/awxkit/cli/resource.py", line 212, in parse_resource
    response = command.handle(client, parser)
  File "/home/jrb/ansible/awx/awxkit/awxkit/cli/resource.py", line 177, in handle
    return client.v2.export_assets(**kwargs)
  File "/home/jrb/ansible/awx/awxkit/awxkit/api/pages/api.py", line 198, in export_assets
    raise Exception("oops!")
```

One problem I do see making the change here to just an `Exception`, the awx collection will exit early without doing output if we do this.  I might have steered you wrong.
:disappointed: 

@john-westcott-iv got any opinions about this?

jbradberry (changes requested):

awxkit/awxkit/api/pages/api.py
sarabrajsingh: in light of way that errors from the ApiV2 level should bubble back up to calling methods in `resource.py` - I believe that we should leave the initialization of `self._has_error` in the constructor

sarabrajsingh: done

awxkit/awxkit/cli/resource.py
```diff
@@ -210,6 +220,10 @@ def parse_resource(client, skip_deprecated=False):
         parser = client.subparsers[resource]
         command = CustomCommand.registry[resource]()
         response = command.handle(client, parser)
+
+        if hasattr(command, '_has_error') and command._has_error:
```
jbradberry: this can be written more compactly with `getattr()`

jbradberry: The other two places as well, assign `getattr()` with a fallback of `False`

awxkit/awxkit/cli/resource.py
```diff
@@ -221,7 +235,7 @@ def parse_resource(client, skip_deprecated=False):
                 connection = None
             formatted = format_response(Page.from_json(response, connection=connection), fmt=client.get_config('format'), filter=_filter)
             print(formatted)
-        raise SystemExit()
+        raise SystemExit(_system_exit)
```
jbradberry, line 236: :+1:

jbradberry (approved):

CFSNM:
Launched a SLOWYO (PR https://github.com/ansible/awx/pull/12408) using the CLI collection: https://main-jenkins-csb-aap.apps.ocp-c1.prod.psi.redhat.com/job/AAPQA/job/Test_Tower_Yolo_Express/1925/

iirc. we have several import/export tests that check some command errors.

Issue #12164, opened 2022-05-03
API not saving timezones correctly


kialam:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

Certain timezones do not save correctly in the API.
Steps to reproduce:
1. Visit the Schedules tab from a Job Template
2. Attempt to save a schedule with `Europe/Skopje` as the timezone
3. Observe the `PATCH` is successful, however inspect the corresponding `GET` request and see:
```
rrule: "DTSTART;TZID=Europe/Skopje:20220707T230000 RRULE:INTERVAL=1;COUNT=1;FREQ=MINUTELY"
timezone: "Europe/Belgrade"
```

### AWX version

21.0.1

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs

### Installation method

docker for mac

### Modifications

no

### Steps to reproduce

See above

### Expected results

API saves correct timezone

### Actual results

Certain timezones get saved as different timezones

jbradberry:
This ultimately stems from recent controversy about how the timezone database (that is used by all languages, not just Python) is maintained.  Some details can be found here: [A fork for the time-zone database?](https://lwn.net/Articles/870478/).  The gist though is that some historical timezones have had their information merged with other more standard ones, and the historical name is now just a link.

My off-the-cuff response to this was to think that we should only expose the standardized non-link timezones now, but upon reflection I think that would be bad UX.  We should probably do something a bit more involved that lets the user know that they can select their old timezone, but standardized timezone X is really what will be used.

jbradberry:
@hesmithrh I no longer think that this is ready to be on deck yet.  There needs to be some discussion on the UI side (led by @kialam) over what to do about this.

john-westcott-iv:
@nixocio and I discussed some potential options:

1. Modify the schedule endpoint to contain both a "ui_selected_tz" and "timezone". The current "timezone" field would not be able to be changed through the UI and would become a computed filed and the new ui_selected_tz would contain the TZ the user actually selected. This field would be used to populate the Local time zone dropdown on edit. 

2. Modify the API to:
    * Return not only the timezone names but also the linked time zone.
[...]

PR #12340, opened 2022-06-07
Add documentation around schedule timezone change

john-westcott-iv:
##### SUMMARY

Timezones can be links to other timezones. For example the UTC timezone is a link to Etc/UCT timezone.
When a user selects UTC in the schedule UI and saves the schedule time UI will then display the Etc/UCT timezone.
This behavior was confusing.
This PR modifies the UI to display a warning message that a selected timezone will be replaced with a different timezone on save to warn the user of the behavior:
image

In addition, the API has been updated to give information about the timezone and until fields (from the options page):
```
            "timezone": {
                "type": "field",
                "label": "Timezone",
                "help_text": "The timezone this schedule runs in. This field is extracted from the RRULE. If the timezone in the RRULE is a link to another timezone this field will show the links name.",
                "filterable": false
            },
            "until": {
                "type": "field",
                "label": "Until",
                "help_text": "The date this schedule will end. This field is computed from the RRULE. If the schedule does not end an emptry string will be returned",
                "filterable": false
            }
```

Addresses #12164 

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - API
 - UI

##### AWX VERSION

```
awx: 21.0.1.dev192+g96eb122aef
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -4683,7 +4683,7 @@ class ScheduleSerializer(LaunchConfigurationBaseSerializer, SchedulePreviewSeria
 
     timezone = serializers.SerializerMethodField(
         help_text=_(
-            'The timezone this schedule runs in. This field is extracted from the RRULE. If the timezone in the RRULE is a link to another timezone this field will show the links name.'
+            'The timezone this schedule runs in. This field is extracted from the RRULE. If the timezone in the RRULE is a link to another timezone, the link will be reflected in this field.'
         ),
     )
     until = serializers.SerializerMethodField(
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -4681,8 +4681,16 @@ class SchedulePreviewSerializer(BaseSerializer):
 class ScheduleSerializer(LaunchConfigurationBaseSerializer, SchedulePreviewSerializer):
     show_capabilities = ['edit', 'delete']
 
-    timezone = serializers.SerializerMethodField()
-    until = serializers.SerializerMethodField()
+    timezone = serializers.SerializerMethodField(
+        help_text=_(
+            'The timezone this schedule runs in. This field is extracted from the RRULE. If the timezone in the RRULE is a link to another timezone this field will show the links name.'
+        ),
+        read_only=True,
+    )
+    until = serializers.SerializerMethodField(
+        help_text=_('The date this schedule will end. This field is computed from the RRULE. If the schedule does not end an emptry string will be returned'),
+        read_only=True,
+    )
 
     class Meta:
         model = Schedule
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -4685,11 +4685,9 @@ class ScheduleSerializer(LaunchConfigurationBaseSerializer, SchedulePreviewSeria
         help_text=_(
             'The timezone this schedule runs in. This field is extracted from the RRULE. If the timezone in the RRULE is a link to another timezone this field will show the links name.'
         ),
-        read_only=True,
     )
     until = serializers.SerializerMethodField(
         help_text=_('The date this schedule will end. This field is computed from the RRULE. If the schedule does not end an emptry string will be returned'),
-        read_only=True,
     )
 
     class Meta:
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -578,8 +578,7 @@ class ScheduleZoneInfo(APIView):
     swagger_topic = 'System Configuration'
 
     def get(self, request):
-        zone_info = models.Schedule.get_zoneinfo_with_links()
-        return Response(zone_info)
+        return Response({'zones': models.Schedule.get_zoneinfo(), 'links': models.Schedule.get_zoneinfo_links()})
 
 
 class LaunchConfigCredentialsBase(SubListAttachDetachAPIView):
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -578,8 +578,8 @@ class ScheduleZoneInfo(APIView):
     swagger_topic = 'System Configuration'
 
     def get(self, request):
-        zones = [{'name': zone} for zone in models.Schedule.get_zoneinfo()]
-        return Response(zones)
+        zone_info = models.Schedule.get_zoneinfo_with_links()
+        return Response(zone_info)
 
 
 class LaunchConfigCredentialsBase(SubListAttachDetachAPIView):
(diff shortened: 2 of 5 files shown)


awx/api/serializers.py
```diff
@@ -4681,8 +4681,16 @@ def validate_rrule(self, value):
 class ScheduleSerializer(LaunchConfigurationBaseSerializer, SchedulePreviewSerializer):
     show_capabilities = ['edit', 'delete']
 
-    timezone = serializers.SerializerMethodField()
-    until = serializers.SerializerMethodField()
+    timezone = serializers.SerializerMethodField(
+        help_text=_(
+            'The timezone this schedule runs in. This field is extracted from the RRULE. If the timezone in the RRULE is a link to another timezone this field will show the links name.'
+        ),
+        read_only=True,
```
AlanCoding: I'm pretty sure that `read_only` isn't needed because it defaults to True for the SerializerMethodField

help text look great!

awx/api/views/__init__.py
```diff
@@ -578,8 +578,8 @@ class ScheduleZoneInfo(APIView):
     swagger_topic = 'System Configuration'
 
     def get(self, request):
-        zones = [{'name': zone} for zone in models.Schedule.get_zoneinfo()]
-        return Response(zones)
+        zone_info = models.Schedule.get_zoneinfo_with_links()
+        return Response(zone_info)
```
AlanCoding: Does this change the returned JSON schema of the view?

awx/main/models/schedules.py
```diff
@@ -88,6 +88,15 @@ class Meta:
     def get_zoneinfo(self):
         return sorted(get_zonefile_instance().zones)
 
+    @classmethod
+    def get_zoneinfo_with_links(self):
```
AlanCoding: Instead of returning a dict like `{"zones": {...}, "links": {...}}`, it might be more clear to return a tuple of `(zones, links)`. That way the specific dict structure is more view-specific. This is something devs are can be pretty opinionated about, so I offer this as a very weak suggestion.

awx/api/views/__init__.py
AlanCoding: Interestingly, I don't _think_ this view is captured by the api-schema check. We could add it to the validated schema by writing a unit test that made a GET to the view. This is an unusual case where that may actually be helpful (unless the time zone library changes frequently).

awx/api/serializers.py
```diff
@@ -4681,8 +4681,16 @@ def validate_rrule(self, value):
 class ScheduleSerializer(LaunchConfigurationBaseSerializer, SchedulePreviewSerializer):
     show_capabilities = ['edit', 'delete']
 
-    timezone = serializers.SerializerMethodField()
-    until = serializers.SerializerMethodField()
+    timezone = serializers.SerializerMethodField(
+        help_text=_(
+            'The timezone this schedule runs in. This field is extracted from the RRULE. If the timezone in the RRULE is a link to another timezone this field will show the links name.'
```
akus062381: My suggested text:
'Timezone for the schedule. This field is extracted from the RRULE. If the RRULE's timezone is linked to another timezone, the link will reflect that.

akus062381 (approved):
I just wanted to make a suggestion!

awx/main/models/schedules.py
```diff
@@ -88,6 +88,15 @@ class Meta:
...
+        zone_instance = get_zonefile_instance()
+        return_val = {'zones': sorted(zone_instance.zones), 'links': {}}
```
jbradberry: I second Alan's comment.  This data structure seems awkward compared to, say, a list of objects like `{'zone': zone, 'link': link}`.

jay-steurer:
Tested and approved.

awx/ui/src/components/Schedule/shared/ScheduleForm.js
```diff
@@ -100,6 +100,24 @@ function ScheduleFormFields({ hasDaysToKeepField, zoneOptions }) {
   });
   const [{ name: dateFieldName }] = useField('startDate');
   const [{ name: timeFieldName }] = useField('startTime');
+  const [timezoneMessage, setTimezoneMessage] = useState('');
+  const warnLinkedTZ = (event, selectedValue) => {
+    if (zoneLinks[selectedValue]) {
+      setTimezoneMessage(
+        `Warning: ${selectedValue} is a link to ${zoneLinks[selectedValue]} and will be saved as that.`
```
kialam, line 107: To mark this string for translation you'll need to import the `t` macro from `lingui` library and wrap it, e.g. 

```
t`my string for translation`
```

kialam:
In the future, PRs that involve UI changes should include at least one UI approval before merging.

awx/ui/src/components/Schedule/shared/ScheduleForm.js
```diff
@@ -124,17 +142,17 @@ function ScheduleFormFields({ hasDaysToKeepField, zoneOptions }) {
...
-        }
+        validated={timezoneValidatedStatus}
         label={t`Local time zone`}
+        helperText={timezoneMessage}
       >
         <AnsibleSelect
           id="schedule-timezone"
           data={zoneOptions}
           {...timezone}
+          onChange={warnLinkedTZ}
```
kialam, line 155: Best practice here is to name the function prop you are passing to the onChange event handler as `handleOnChange` rather than something as specific as `warnLinkedTZ`. In actuality, you're doing more than just warning the user that a user has selected a linked time zone (which depends on their selection) - there's a series of validations that occur or there may be other bits of logic that need to slot into place.

awx/ui/src/components/Schedule/shared/ScheduleForm.js
```diff
@@ -100,6 +100,24 @@ function ScheduleFormFields({ hasDaysToKeepField, zoneOptions }) {
...
+  const [timezoneMessage, setTimezoneMessage] = useState('');
+  const warnLinkedTZ = (event, selectedValue) => {
+    if (zoneLinks[selectedValue]) {
+      setTimezoneMessage(
+        `Warning: ${selectedValue} is a link to ${zoneLinks[selectedValue]} and will be saved as that.`
+      );
+    } else {
+      setTimezoneMessage('');
+    }
+    timezone.onChange(event, selectedValue);
```
kialam, line 112: ~~I would not call this event handler function here - the user is not providing any input. If you want to change the field value, you can call a [helper](https://formik.org/docs/api/useField#fieldhelperprops) method from formik called `setValue`.~~ <- onChange works because it's tied to the actual onChange prop in the dropdown. I was thrown off by the handler function name :)

kialam (commented):
Some of these changes I have requested should be done as part of a follow up PR to this one.

Issue #12334, opened 2022-06-06
Selecting management jobs as non system admin raises error as node type


nixocio:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

Selecting management jobs as non system admin raises error as node type

<img width="1124" alt="image" src="https://user-images.githubusercontent.com/9053044/172234815-633a7b1b-c11a-455c-ab65-9656de270ccd.png">

### AWX version

devel

### Select the relevant components

- [X] UI
- [ ] API
- [ ] Docs

### Installation method

docker for mac

### Modifications

no

### Steps to reproduce

1 - Create a user a normal user, user A.
2 - Give the user A - admin permissions on a workflow job template.
3 - Login as user A.
4 - Attempt to add an extra node - select management job as node type.

### Expected results

Not be able to select management jobs as node type

### Actual results

Able to attempt to select management jobs as node type

nixocio:
See: https://github.com/ansible/awx/pull/10572

InnocentK:
Resolved

PR #12341, opened 2022-06-07
Hide management job for non system admin as node choice

nixocio:
Hide management job for non system admin as node type choice. Also, fix
related uni-tests related to this change.

See: https://github.com/ansible/awx/issues/12334
Also: https://github.com/ansible/awx/pull/10572

System Admin

image

Non System Admin
image


kialam (approved):
👍

InnocentK (approved):
Looks good!

Issue #11987, opened 2022-03-31
Project status in jobs it's wrong


chris93111:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

Hello 

when job fail with project update error , the status of sync project is wrong in UI 

In output job
```
Traceback (most recent call last):
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/jobs.py", line 449, in run
    self.pre_run_hook(self.instance, private_data_dir)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/jobs.py", line 955, in pre_run_hook
    sync_task.run(local_project_sync.id)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/jobs.py", line 92, in _wrapped
    return f(self, *args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/jobs.py", line 604, in run
    raise AwxTaskError.TaskError(self.instance, rc)
Exception: project_update 443522 (failed) encountered an error (rc=None), please see task stdout for details.
```

In job details ui 
<img width="211" alt="image" src="https://user-images.githubusercontent.com/27526373/161085818-6ecc265c-7ac8-4df4-8ab6-3b708d7dc688.png">

in project ui 
<img width="307" alt="image" src="https://user-images.githubusercontent.com/27526373/161085668-e85bd62d-009e-4276-984c-34305b835a68.png">

### AWX version

20.0.0

### Select the relevant components

- [X] UI
- [x] API
- [ ] Docs

### Installation method

kubernetes

### Modifications

no

### Ansible version

2.9.9

### Operating system

redhat

### Web browser

Chrome

### Steps to reproduce

launch job with sync project or override branch scm

### Expected results

good status in job details

### Actual results

wrong status

fosterseth:
@chris93111 what happens when you refresh the page, does the status change to failed on the job detail page?

also what does the `/api/v2/project_updates/{id}/` show for that project update? also what does `/api/v2/jobs/{id}/` for the job that was launched show?

chris93111:
[...]

PR #12348, opened 2022-06-08
Update project status to reflect project sync related to job template

nixocio:
Update project status to reflect project update sync related to job
template that was launched with branch override.

We were displaying status of project sync itself, not from the project
update job as expected.

Also, rename `Project Status` to be `Project Update Status`.

Launching a Job Template using a Project that allows branch override using a non valid branch name.

Before this change.

`Project Status` was different from the job that is linked to.

image

After clicking `Project Status`

image

After this change

Status for `Project Update Status` is the same as the job that is linked to.

image

After clicking `Project Update Status`

image

See: https://github.com/ansible/awx/issues/11987

### DIFF
--- a/awx/ui/src/screens/Job/Job.helptext.js
+++ b/awx/ui/src/screens/Job/Job.helptext.js
@@ -18,6 +18,7 @@ const jobHelpText = {
   jobTags: t`Tags are useful when you have a large playbook, and you want to run a specific part of a play or task. Use commas to separate multiple tags. Refer to the documentation for details on the usage of tags.`,
   skipTags: t`Skip tags are useful when you have a large playbook, and you want to skip specific parts of a play or task. Use commas to separate multiple tags. Refer to the documentation for details on the usage of tags.`,
   sourceControlBranch: t`Select a branch for the workflow. This branch is applied to all job template nodes that prompt for a branch.`,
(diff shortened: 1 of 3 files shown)


AlanCoding:
Can I go through a few questions here?

Will this require any additional requests to render the page? It appears the project update is already in summary fields.

```json
        "project_update": {
            "id": 309,
            "name": "Demo Project",
            "description": "",
            "status": "successful",
            "failed": false
        },
```

So I don't think this change should need to make anything slower.

Next, what happens when there is no related project update? This is common. Run with 1 hybrid node, and most jobs will not run one, because the relevant revision is already present on disk.

nixocio:
Good questions @AlanCoding .

There is no extra request, we already read that information from summary_fields as you stated.
```js

    labels,
    project,
    project_update: projectUpdate,
    source_workflow_job,
    execution_environment: executionEnvironment,
  } = job.summary_fields;
```

 There is a conditional to just display that info if there is `projectUpdate` otherwise we do not show that information. We already linked to the right job, but we were displaying the status related to project and not project_update. This PR update this.

AlanCoding (approved):
Looks good as far as I'm concerned with the communication of status to the user.

kialam (approved):
nice work

akus062381 (approved):

Issue #12092, opened 2022-04-22
Provide feedback when saving a Large Workflow Template 


nixocio:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Provide feedback when saving a large workflow template. 

Right now we do not provide any feedback when saving a large workflow job template, this is specially noticeable with a large number of nodes and connections. This leads to several HTTP requests, and the user can click the save button several times during this process.

We could either disable the button once the user clicks on it, until we redirect the user to the details page.

PR #12349, opened 2022-06-08
Add loading state when saving a visualizer

nixocio:
Add loading state when saving a visualizer

See: https://github.com/ansible/awx/issues/12092


AlexSCorey (approved):
This looks good.  Consider adding a test for loading state.

akus062381 (approved):

PR #12351, opened 2022-06-08
Adds import export to awx cli for schedules as a top level object

AlexSCorey:
##### SUMMARY
This works allows an awx cli user to import and export Schedules as top level objects
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/awxkit/awxkit/api/pages/schedules.py
+++ b/awxkit/awxkit/api/pages/schedules.py
@@ -1,6 +1,7 @@
 from contextlib import suppress
 
-from awxkit.api.pages import UnifiedJob
+from awxkit.api.pages import SystemJobTemplate
+from awxkit.api.mixins import HasCreate
 from awxkit.api.resources import resources
 from awxkit.config import config
 import awxkit.exceptions as exc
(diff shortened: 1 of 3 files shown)


awxkit/awxkit/api/pages/schedules.py
```diff
@@ -25,7 +26,7 @@ def silent_delete(self):
             pass
 
 
-page.register_page([resources.schedule, resources.related_schedule], Schedule)
+page.register_page([resources.schedule, resources.related_schedule, (resources.schedule, 'post')], Schedule)
```
jbradberry: I think you intended `(resources.schedules, 'post')` here.  You are trying to say, "when I do a post to the schedules list page, the resulting object is a schedule".

jbradberry: What you currently have is like saying, "when I post changes to a schedule detail page, the resulting object is a schedule", and is already implied by the first `resources.schedule` entry in the list.

AlexSCorey: This was left over from my debugging and I did not mean to commit it.  Thanks for catching it.

jbradberry (approved):

PR #12355, opened 2022-06-09
Make awx-autoreloader work faster for large code changes

AlanCoding:
##### SUMMARY
I found that if I switched from a large branch to `devel` and back again, this would kick off _multiple_ supervisor reload commands. They would all queue up, and you would have to wait for it to complete that number of restarts (imagine, ballpark of 20 restarts, for 20 file changes) before you could use the server again.

This tracks the time of the last code reload, so that if it was less than a second ago, we don't issue another reload command. Initial testing seems to show it working well.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
 - API

### DIFF
--- a/tools/docker-compose/awx-autoreload
+++ b/tools/docker-compose/awx-autoreload
@@ -6,10 +6,15 @@ if [ $# -lt 2 ]; then
     exit 1
 fi
 
+last_reload=`date +%s`
+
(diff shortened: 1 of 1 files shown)


shanemcd (approved):
Dude!

PR #12356, opened 2022-06-09
Replace git shallow clone with shutil.copytree

AlanCoding:
##### SUMMARY
This is a rebase of https://github.com/ansible/awx/pull/7367 (but actually re-applied by hand)

As of opening this:
 - I have some tests running, and don't want to take further action until looking at those
 - I would like to break off a new method `build_project_dir`, for a couple of reasons
   - so we don't erroneously mix `pre_run_hook` error handling into the project error handling
   - due to this change, we need to more clearly demarcate where the `project` folder needs to be created (copytree requires that it does not yet exist)
   - this is more descriptive and move more code out of the catch-all `*_hook` methods

Because of the heavy tie-ins to this change, I believe it's a good time to do that minor refactor after the basic functionality is validated.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
I talked before about removing the GitPython library. For the moment, I am considering that too complex and risky. We have two very narrow uses for that library that remain. The objective here is just to fix the git-related bugs.

### DIFF
--- a/awx/main/tasks/jobs.py
+++ b/awx/main/tasks/jobs.py
@@ -1,6 +1,5 @@
 # Python
 from collections import OrderedDict
-from distutils.dir_util import copy_tree
 import errno
 import functools
 import fcntl
@@ -15,7 +14,6 @@ import tempfile
 import traceback
 import time
 import urllib.parse as urlparse
-from uuid import uuid4
 
 # Django
 from django.conf import settings
@@ -211,13 +209,21 @@ class BaseTask(object):
         os.chmod(path, stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR)
         if settings.AWX_CLEANUP_PATHS:
             self.cleanup_paths.append(path)
-        # Ansible runner requires that project exists,
-        # and we will write files in the other folders without pre-creating the folder
-        for subfolder in ('project', 'inventory', 'env'):
+        # We will write files in these folders later
+        for subfolder in ('inventory', 'env'):
             runner_subfolder = os.path.join(path, subfolder)
             if not os.path.exists(runner_subfolder):
                 os.mkdir(runner_subfolder)
         return path
+
+    def build_project_dir(self, instance, private_data_dir):
+        """
+        Create the ansible-runner project subdirectory. In many cases this is the source checkout.
+        In cases that do not even need the source checkout, we create an empty dir to be the workdir.
+        """
+        project_dir = os.path.join(private_data_dir, 'project')
+        if not os.path.exists(project_dir):
+            os.mkdir(project_dir)
 
     def build_private_data_files(self, instance, private_data_dir):
         """
@@ -354,6 +360,52 @@ class BaseTask(object):
             expect_passwords[k] = passwords.get(v, '') or ''
         return expect_passwords
 
+    def release_lock(self, project):
+        try:
+            fcntl.lockf(self.lock_fd, fcntl.LOCK_UN)
+        except IOError as e:
+            logger.error("I/O error({0}) while trying to release lock file [{1}]: {2}".format(e.errno, project.get_lock_file(), e.strerror))
+            os.close(self.lock_fd)
+            raise
+
+        os.close(self.lock_fd)
+        self.lock_fd = None
+
+    def acquire_lock(self, project, unified_job_id=None):
+        if not os.path.exists(settings.PROJECTS_ROOT):
+            os.mkdir(settings.PROJECTS_ROOT)
+
+        lock_path = project.get_lock_file()
+        if lock_path is None:
+            # If from migration or someone blanked local_path for any other reason, recoverable by save
+            project.save()
+            lock_path = project.get_lock_file()
+            if lock_path is None:
+                raise RuntimeError(u'Invalid lock file path')
+
+        try:
+            self.lock_fd = os.open(lock_path, os.O_RDWR | os.O_CREAT)
+        except OSError as e:
+            logger.error("I/O error({0}) while trying to open lock file [{1}]: {2}".format(e.errno, lock_path, e.strerror))
+            raise
+
+        start_time = time.time()
+        while True:
+            try:
+                fcntl.lockf(self.lock_fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                break
+            except IOError as e:
+                if e.errno not in (errno.EAGAIN, errno.EACCES):
+                    os.close(self.lock_fd)
+                    logger.error("I/O error({0}) while trying to aquire lock on file [{1}]: {2}".format(e.errno, lock_path, e.strerror))
+                    raise
+                else:
+                    time.sleep(1.0)
+        waiting_time = time.time() - start_time
+
+        if waiting_time > 1.0:
+            logger.info(f'Job {unified_job_id} waited {waiting_time} to acquire lock for local source tree for path {lock_path}.')
+
     def pre_run_hook(self, instance, private_data_dir):
         """
         Hook for any steps to run before the job/task starts
@@ -424,6 +476,7 @@ class BaseTask(object):
             self.instance.send_notification_templates("running")
             private_data_dir = self.build_private_data_dir(self.instance)
             self.pre_run_hook(self.instance, private_data_dir)
+            self.build_project_dir(self.instance, private_data_dir)
             self.instance.log_lifecycle("preparing_playbook")
             if self.instance.cancel_flag or signal_callback():
                 self.instance = self.update_model(self.instance.pk, status='canceled')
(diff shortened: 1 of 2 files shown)


awx/main/tasks/jobs.py
```diff
@@ -1324,62 +1334,23 @@ def clear_project_cache(cache_dir, keep_value):
                     except OSError:
                         logger.warning(f"Could not remove cache directory {old_path}")
 
+    def make_local_copy_with_lock(self, project, job_private_data_dir):
+        self.acquire_lock(project)
+        try:
+            self.make_local_copy(project, job_private_data_dir)
+        finally:
+            self.release_lock(project)
```
AlanCoding: I'm still not happy with this one part. It's superficial and doesn't really work.

We need to obtain the lock before we check the revision, and we need to keep holding it until after the folder copy is done.

```python
git_repo = git.Repo(project_path)
if job_revision == git_repo.head.commit.hexsha:
  ...
```

Otherwise there's not really any point in this. I didn't really want to make a context manager out of this, but maybe that's what will happen.

AlanCoding:
@fosterseth after looking into the other option, I believe PR is the right path forward. I will wait for your work to land and then rebase this.

awx/main/tasks/jobs.py
```diff
@@ -856,98 +1045,14 @@ def pre_run_hook(self, job, private_data_dir):
...
-            # Project update does not copy the folder, so copy here
-            RunProjectUpdate.make_local_copy(job.project, private_data_dir, scm_revision=job_revision)
-
         if job.inventory.kind == 'smart':
             # cache smart inventory memberships so that the host_filter query is not
             # ran inside of the event saving code
             update_smart_memberships_for_inventory(job.inventory)
 
+    def build_project_dir(self, job, private_data_dir):
+        self.sync_and_copy(job.project, private_data_dir, scm_branch=job.scm_branch)
```
john-westcott-iv, line 1061: Do you not need the "if manual" check here like on line 1564? If we do end up using that in more than one place could we just pass in the project type to sync_and_copy and let that method return if the type is manual?

awx/main/tasks/jobs.py
```diff
@@ -586,8 +639,144 @@ def run(self, pk, **kwargs):
...
+                    )
+                    raise
+                self.instance.refresh_from_db()
+                if self.instance.cancel_flag:
+                    return
+        else:
+            # Case where a local sync is not needed, meaning that local tree is
+            # up-to-date with project, job is running project current version
+            if isinstance(self.instance, Job):
+                self.instance = self.update_model(self.instance.pk, scm_revision=project.scm_revision)
```
john-westcott-iv, line 752: Do you not need the else clause here like you did on line 716?

awx/main/tasks/jobs.py
```diff
@@ -586,8 +639,144 @@ def run(self, pk, **kwargs):
...
+                if isinstance(self.instance, Job):
+                    self.instance = self.update_model(self.instance.pk, scm_revision=local_project_sync.scm_revision)
+            except Exception:
+                local_project_sync.refresh_from_db()
+                if local_project_sync.status != 'canceled':
+                    self.instance = self.update_model(
+                        self.instance.pk,
+                        status='failed',
+                        job_explanation=(
+                            'Previous Task Failed: {"job_type": "%s", "job_name": "%s", "job_id": "%s"}'
```
john-westcott-iv: Why do a sub on 'project_update' vs just having it in the string?

awx/main/tasks/jobs.py
```diff
@@ -353,6 +359,52 @@ def create_expect_passwords_data_struct(self, password_prompts, passwords):
...
+                raise RuntimeError(u'Invalid lock file path')
+
+        try:
+            self.lock_fd = os.open(lock_path, os.O_RDWR | os.O_CREAT)
+        except OSError as e:
+            logger.error("I/O error({0}) while trying to open lock file [{1}]: {2}".format(e.errno, lock_path, e.strerror))
+            raise
+
+        start_time = time.time()
+        while True:
```
john-westcott-iv, line 393: Do you want a max_wait_time counter in here to prevent an infinite loop?

awx/main/tasks/jobs.py
```diff
@@ -353,6 +359,52 @@ def create_expect_passwords_data_struct(self, password_prompts, passwords):
...
+        if not os.path.exists(settings.PROJECTS_ROOT):
+            os.mkdir(settings.PROJECTS_ROOT)
+
+        lock_path = project.get_lock_file()
+        if lock_path is None:
+            # If from migration or someone blanked local_path for any other reason, recoverable by save
+            project.save()
+            lock_path = project.get_lock_file()
+            if lock_path is None:
+                raise RuntimeError(u'Invalid lock file path')
```
john-westcott-iv, line 384: Should we do this logic inside get_lock_file method?

AlanCoding: This isn't really code I'm trying to change, I'm just moving it.

I think a problem with `get_lock_file` raising an exception is that it would preclude the approach here, where it does a `.save()` as a last-ditch fix if it initially returns `None`. On the other hand, I would not want a method named `get_lock_file` to cause a save as a side-effect. So that's my 2-second defense of the structure here, although I'm committed to defending it, I don't have a complete idea of how to restructure it.

awx/main/tasks/jobs.py
AlanCoding: But what would we set it to? From the logger down below, we know it's common in real environments to wait for a long time. We've seen 2 hours of waiting happen. This isn't great, but it's going to be hard to avoid with slow and problematic project and Ansible content.

awx/main/tasks/jobs.py
AlanCoding: ha, the joys of old code. I'll push a commit to modernize.

awx/main/tasks/jobs.py
AlanCoding: Unfortunately no, this is why I filed https://github.com/ansible/awx/issues/12413

There are a couple of model changes that would make a lot of sense to do for consistency's sake, but we need corresponding UI changes for those. This PR exists to solve the problem of nested submodules in projects, and it got very big so I'm trying to punt on further changes.

We _should_ add the `scm_revision` field _to inventory updates_ and there will then be no `if` here, because the same will work for either a `Job` or `InventoryUpdate`

awx/main/tasks/jobs.py
AlanCoding: You bring up an interesting point, and it looks like the comment on ln 1566 about manual projects is misleading. The method `sync_and_copy` should already handle manual projects correctly. The super method `build_project_dir` doesn't copy anything, it just creates an empty project directory (for purposes of mounting, workdir, and such).

AlanCoding:
After this merges, we should revisit https://github.com/ansible/awx/issues/1823, as it may become trivial.

gundalow:
@AlanCoding What's needed to get this reviewed & merged

awx/main/tasks/jobs.py
john-westcott-iv: What about a setting like AWX_MAX_PROJECT_LOCK_WAIT_TIME_IN_SECONDS, set to something like 6h as a a starting point?

gundalow:
> @fosterseth after looking into the other option, I believe PR is the right path forward. I will wait for your work to land and then rebase this.

Believe this is referring to https://github.com/ansible/awx/pull/12366 which has been merged.

gundalow:
Seth has kindly offered to review this so we can get it merged.

AlanCoding:
I've been working with @jay-steurer to develop complete integration testing of nested git submodules, which we've never had before. We currently have testing of a few permutations of submodules, but these are all one layer deep. Nested submodules requires a different approach, because you can't hack it together with test-playbooks, you need another github repo, or a different approach.

fosterseth (approved):

awx/main/tasks/jobs.py
```diff
@@ -1244,11 +1291,12 @@ def pre_run_hook(self, instance, private_data_dir):
             shutil.rmtree(stage_path)
         os.makedirs(stage_path)  # presence of empty cache indicates lack of roles or collections
 
+    def build_project_dir(self, instance, private_data_dir):
         # the project update playbook is not in a git repo, but uses a vendoring directory
         # to be consistent with the ansible-runner model,
         # that is moved into the runner project folder here
         awx_playbooks = self.get_path_to('../../', 'playbooks')
-        copy_tree(awx_playbooks, os.path.join(private_data_dir, 'project'))
+        shutil.copytree(awx_playbooks, os.path.join(private_data_dir, 'project'))
```
shanemcd, line 1306: @branic had some valid feedback that it'd be good to catch errors that happen here and make sure they are surfaced clearly to the user.

shanemcd (changes requested):

awx/main/tasks/jobs.py
branic: There are a couple of other places copytree is used (lines 1324 and 1337) that should have errors caught as well.

gundalow: Is that the last thing before we can get this merged?
Do we require further reviews or testing?

AlanCoding: This is run inside of the main try-except block in the `run` method (pre_run_hook previously, another name here). Errors there are saved directly to the `result_traceback` of the job. This shows in the output tab.

If their run involved a project sync, then the traceback may be visible on the project sync (not the job), and this is a common cause of user confusion, but it's still in the UI, and broader UX issues I would like to keep separate from this.

shanemcd (approved):

kdelee (approved):

PR #12358, opened 2022-06-10
Hide add access button based on the user profile for credentials

nixocio:
* Show add access button if it is a system admin
* Hide access button if the user is credential admin, org admin, but the
  credential does not belong to any org.
* Show access button if the user is a credential admin, org admin, and
  the credential is associated to an org.
* Show access button if the user is an org admin and the credential is
  associated to the org.

All those permutations are allowed by the API RBAC.
This PR update UX to not allow the user to attempt to perform any
action that will raise an error when modifying access to the
credentials.

### DIFF
--- a/awx/ui/src/api/models/Organizations.js
+++ b/awx/ui/src/api/models/Organizations.js
@@ -77,6 +77,10 @@ class Organizations extends InstanceGroupsMixin(NotificationsMixin(Base)) {
       disassociate: true,
     });
   }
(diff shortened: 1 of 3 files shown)


AlexSCorey (approved):
This looks good. Really good test coverage.  Just 1 nit pick pick that isn't a blocker.

akus062381 (approved):
looks great

Issue #10093, opened 2021-05-03
UX Feedback: Allow for refreshing data in lookup modals


shanemcd:
A lot of the times when i'm using the UI, I get to a point where i need to create a credential. I'll open up a new tab, create the credential, then go back to my first tab. The old UI used to do do the api call when you clicked the little magnifying glass, so it would pick up the newly created credential.

Not prescribing a solution here, just describing the problem. Maybe a little manual "refresh" button / icon would do the trick?

AlexSCorey:
@shanemcd are you saying that the newly created credential does not appear in a credential look up when you open it?
![Screen Shot 2021-05-04 at 10 06 24 AM](https://user-images.githubusercontent.com/39280967/117016524-cabcc800-acc0-11eb-8c41-6ced8c0de8e8.png)

shanemcd:
Correct.

AlexSCorey:
Ok, this is probably a problem with other look ups as well.

mabashian:
Yes, we make the requests to initially load the data when the form loads and not when the lookup is opened.  I agree that the behavior in the old application is preferable as it would cut down on the total number of requests made (only making requests when the user wants to see more data) and it would solve the problem described initially.

There's been some time that's passed since this pattern initially emerged.  Perhaps there was a good reason why we did it this way but I don't remember.

My gut tells me that the fix here would be to initialize the fetching of the rows when the modal is opened rather than when the component mounts.

Pretty sure this applies to _all_ lookups in all forms.

keithjgrant:
> Perhaps there was a good reason why we did it this way but I don't remember.

My guess would be we did it for the performance/perceived performance of having the list pre-fetched before the Lookup is opened. We could probably try to do some intelligent re-fetching if the lookup gets opened after a long delay or something, but a manual refresh button might just make more sense

In some cases, the lookup might be able to subscribe to websocket events for new objects, but that seems rather heavy-handed

hesmithrh:
potentially related to https://github.com/ansible/awx/issues/11346 (per UI dev convo today)

PR #12361, opened 2022-06-10
Allow lookup modals to refresh when opened.

kialam:
##### SUMMARY

Related #10093 

##### ISSUE TYPE

 - Feature Pull Request
 
##### COMPONENT NAME

 - UI


AlexSCorey:
This looks good so far.  My only question is, are we still fetching the lookup data when the page loads in addition to when the user clicks to open the look up?

kialam:
Some of them yes because they require the field to be auto populated with the selected org, cred, etc.

nixocio (approved):
@kialam I like the simplicity of this solution. I got the code locally, and work as expected. Great job.

jay-steurer:
Tested and approved.

PR #12363, opened 2022-06-13
Modify position of tooltip for management job list

nixocio:
Modify position of tooltip for management job list. Also, remove
duplicated tooltip.

### DIFF
--- a/awx/ui/src/screens/ManagementJob/ManagementJobList/LaunchManagementPrompt.js
+++ b/awx/ui/src/screens/ManagementJob/ManagementJobList/LaunchManagementPrompt.js
@@ -29,7 +29,7 @@ function LaunchManagementPrompt({
   const [dataRetention, setDataRetention] = useState(defaultDays);
   return (
     <>
-      <Tooltip content={t`Launch management job`} position="top">
+      <Tooltip content={t`Launch management job`} position="left">
         <Button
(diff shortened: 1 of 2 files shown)


awx/ui/src/components/PaginatedTable/ActionItem.js
```diff
@@ -2,7 +2,13 @@ import 'styled-components/macro';
 import React from 'react';
 import { Tooltip } from '@patternfly/react-core';
 
-export default function ActionItem({ column, tooltip, visible, children }) {
+export default function ActionItem({
+  column,
+  tooltip,
+  visible,
+  position,
```
kialam: This prop isn't being used anywhere in the code - is this change meant for future-scoped work?

kialam (changes requested):
Changes to `ActionItem.js` seem extraneous.

awx/ui/src/components/PaginatedTable/ActionItem.js
nixocio: @kialam, yeah. It would be.

But I can remove that one.

nixocio: @kialam, removed.

kialam (approved):
👍

akus062381 (approved):

Issue #12206, opened 2022-05-10
Remove already-deprecated `update_on_project_update` field


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

(feature removal?)

##### SUMMARY
This field was deprecated in https://github.com/ansible/awx/pull/11989

See the comment chain here:

https://github.com/ansible/awx/issues/11336#issuecomment-970321742

This is filing that as a separate issue. We now have https://github.com/ansible/awx/pull/12073, which resolves what was discussed in https://github.com/ansible/awx/issues/11336 and more. We will want to encourage people to use this new, simpler, system.

To do that, we should add a migration which deletes the old field `update_on_project_update`.

How do we do that without breaking functionality? I would argue for the following migration script:

 - Loop over all inventory sources with `update_on_project_update=True`
 - Change the inventory source `update_on_launch` to `True`
 - For every one of those, check related `source_project`
 - If that project has `update_on_launch=False`, change its value to `True`

As users continue to make use of AWX, this will result in more timely inventory updates.

Only downside is that it may run _more updates_ than what the user wants. I find this to be a minor inconvenience, and is a trade-off that we should be willing to make.

fosterseth:
also we can safely remove the scm_last_revision field on the inventory source model

john-westcott-iv:
Blocked pending merge of #12073

AlanCoding:
@fosterseth I saw this was marked as in-progress, are you working on it?

hesmithrh:
@fosterseth can we close this since we are tracking the docs work in https://github.com/ansible/product-docs/issues/1722 (linked above) thanks!

hesmithrh:
@fosterseth can we close this since we have a separate docs issue? thanks

PR #12366, opened 2022-06-13
Remove deprecated field update_on_project_update

fosterseth:
##### SUMMARY

remove deprecated fields `update_on_project_update` and `scm_last_revision` for InventorySource
related https://github.com/ansible/awx/issues/12206

todo: fix up / remove QA tests

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.0.1.dev219+g0df3c655b9
```


0164_remove_inventorysource_update_on_project_update.py
```diff
@@ -0,0 +1,29 @@
...
+
+def forwards(apps, schema_editor):
+    InventorySource = apps.get_model('main', 'InventorySource')
+    sources = InventorySource.objects.filter(update_on_project_update=True)
+    for src in sources:
+        if src.update_on_launch == False:
+            src.update_on_launch = True
+            src.save(update_fields=['update_on_launch'])
+        proj = src.source_project
+        if proj and proj.update_on_launch == False:
```
AlanCoding: Linter probably wants `is False`

0164_remove_inventorysource_update_on_project_update.py
```diff
@@ -0,0 +1,29 @@
...
+            proj.update_on_launch = True
+            proj.save(update_fields=['update_on_launch'])
```
AlanCoding: This is the only change that might be somewhat consequential to the user. I would probably put a warning level log here. This is also a big item for the release notes.

AlanCoding:
We may be able to remove the SCM launch type

https://github.com/ansible/awx/blob/0df3c655b9df68b6a60636cd1b0ea3ae2f7f7134/awx/main/models/unified_jobs.py#L536

If not, I would advocate that we change it to deprecated. Inventory updates from before the migration could still validly have this status.

AlanCoding:
Even if we do nothing on that `launch_type` option, we can still purge references to the "scm" value. In particular, these lines in the `RunInventorySource` model are complementary to logic in the `_update_dependent_inventories` which is already being deleted.

https://github.com/ansible/awx/blob/0df3c655b9df68b6a60636cd1b0ea3ae2f7f7134/awx/main/tasks/jobs.py#L1604-L1606

AlanCoding:
A more minor deletion would be here:

https://github.com/ansible/awx/blob/a3e29317c5d4220fffe28370ec73c73802255246/awx/main/models/inventory.py#L1307

The condition `self.launch_type != 'scm'` should always resolve to true after this. So only checking `if self.source_project_update_id:` would be the most efficient.

awx/main/tasks/jobs.py
```diff
@@ -1655,8 +1591,6 @@ def pre_run_hook(self, inventory_update, private_data_dir):
                 sync_task = project_update_task(job_private_data_dir=private_data_dir)
                 sync_task.run(local_project_sync.id)
                 local_project_sync.refresh_from_db()
-                inventory_update.inventory_source.scm_last_revision = local_project_sync.scm_revision
-                inventory_update.inventory_source.save(update_fields=['scm_last_revision'])
```
AlanCoding, line 1659: This looks like it may need something added back in. Jobs have an `scm_revision` field to track which revision it ran. We may want to add back in a similar field for inventory updates.

fosterseth: would it be better to have `InventorySource.scm_revision` to be consistent with `Job.scm_revision`?

AlanCoding: I'm hitting some similar questions in https://github.com/ansible/awx/pull/12356. Jobs and inventory updates logically should have a multitude of things in common (like this) but it was never fully thought out. My short answer is "yes", it should be consistent, but the broader UI impact and things probably need to get some thought.

awx/main/models/inventory.py
```diff
@@ -1302,11 +1260,7 @@ def preferred_instance_groups(self):
         return selected_groups
 
     def cancel(self, job_explanation=None, is_chain=False):
-        res = super(InventoryUpdate, self).cancel(job_explanation=job_explanation, is_chain=is_chain)
-        if res:
-            if self.launch_type != 'scm' and self.source_project_update:
-                self.source_project_update.cancel(job_explanation=job_explanation)
-        return res
+        return super(InventoryUpdate, self).cancel(job_explanation=job_explanation, is_chain=is_chain)
```
AlanCoding: no need to keep this method, you can delete the entire `def` as the super class implementation is fine now.

fosterseth: oh right, of course, thanks!

awx/main/migrations/0164_remove_inventorysource_update_on_project_update.py
```diff
@@ -0,0 +1,39 @@
...
+logger = logging.getLogger("awx")
+
+
+def forwards(apps, schema_editor):
+    InventorySource = apps.get_model('main', 'InventorySource')
+    sources = InventorySource.objects.filter(update_on_project_update=True)
+    for src in sources:
+        if src.update_on_launch == False:
+            src.update_on_launch = True
+            src.save(update_fields=['update_on_launch'])
```
AlanCoding, line 15: it seems a little in-congruent to not have a log here (like for the project change). Even if it's a lower level, like `logger.info`, that would be good to add.

AlanCoding (approved):

PR #12367, opened 2022-06-13
Forcing an unbind for a django-auth-ldap sticky session to the LDAP server

john-westcott-iv:
##### SUMMARY

django-auth-ldap will use a sticky session if the BIND_DN and BIND_PASS are specified in the LDAP config. This change will modify the behavior to have AWX close the sticky session to free up the uwsgi thread.

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.1.1.dev64+gd132e9ec2a
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/sso/backends.py
+++ b/awx/sso/backends.py
@@ -124,9 +124,9 @@ class LDAPBackend(BaseLDAPBackend):
                 logger.debug("Forcing LDAP connection to close")
                 try:
                     ldap_user.ldap_user._connection.unbind_s()
-                except Exception as unbind_e:
-                    logger.error("Got unexpected exception when forcing disconnect")
-                    logger.error(unbind_e)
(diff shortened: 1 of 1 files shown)


john-westcott-iv:
This should have some testing to make sure a uwsgi thread that had an LDAP connection does not error on an attempted reuse of an LDAP connection that was forced closed. We should also see if we can do a "user login load test" to see what the performance difference is between a sticky and regular connection.

AlanCoding (approved):
assuming the units tests are worked out, this looks fine to me in principle

jbradberry (approved):

dsglaser:
Is this planning on being backported into any of the Tower 3.8.x stream?

ffirg:
No. 3.8 is now in maintenance support phase 2.

Issue #11795, opened 2022-02-23
Organization as a column in the Templates view?


jonli662:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Is it possible to add organization as a column in the Templates view in AWX?
Currently, we need to prefix all template names to easily separate ownership.

PR #12370, opened 2022-06-13
Add column org to template list

nixocio:
Add column org to template list

This change was added based on the feedback from @trahman73 .

image

See: https://github.com/ansible/awx/issues/11795


AlexSCorey (approved):
Looks good!

PR #12376, opened 2022-06-14
work around for 500 on /api/v2/metrics when upgrading

kdelee:
##### SUMMARY

we've observed this bug in development and some users have reported experiencing 500's on /api/v2/metrics because of a key error here where a metric is missing from a certain instance

##### ISSUE TYPE

 - Bug or Docs Fix

##### COMPONENT NAME

 - API

##### AWX VERSION

```

```
devel, though it seems this is present for a while

##### ADDITIONAL INFORMATION

```

```
I'm not sure if just skipping the metric is best, or if we should write it but with some null/dummy value.


fosterseth (approved):

fosterseth:
note to self, will probably need to be backported to 2.2

AlanCoding (approved):

PR #12378, opened 2022-06-14
Making verbosity list and options a constant and adding WinRM debug 

john-westcott-iv:
##### ISSUE TYPE

 - Bug or Docs Fix

##### COMPONENT NAME

 - UI

##### AWX VERSION

```
awx: 21.1.1.dev84+gf042274bc5
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/ui/src/components/AdHocCommands/AdHocCommandsWizard.js
+++ b/awx/ui/src/components/AdHocCommands/AdHocCommandsWizard.js
@@ -78,7 +78,6 @@ const FormikApp = withFormik({
 FormikApp.propTypes = {
   onLaunch: PropTypes.func.isRequired,
   moduleOptions: PropTypes.arrayOf(PropTypes.array).isRequired,
-  // verbosityOptions: PropTypes.arrayOf(PropTypes.object).isRequired,
   onCloseWizard: PropTypes.func.isRequired,
   credentialTypeId: PropTypes.number.isRequired,
 };
--- a/awx/ui/src/components/AdHocCommands/AdHocCommandsWizard.js
+++ b/awx/ui/src/components/AdHocCommands/AdHocCommandsWizard.js
@@ -3,13 +3,13 @@ import { t } from '@lingui/macro';
 import { withFormik, useFormikContext } from 'formik';
 import PropTypes from 'prop-types';
 
+import { VERBOSITY } from 'components/VerbositySelectField';
 import Wizard from '../Wizard';
 import useAdHocLaunchSteps from './useAdHocLaunchSteps';
 
 function AdHocCommandsWizard({
   onLaunch,
   moduleOptions,
-  verbosityOptions,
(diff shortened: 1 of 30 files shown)


awx/ui/src/constants.js
```diff
@@ -12,3 +14,18 @@ export const SESSION_TIMEOUT_KEY = 'awx-session-timeout';
 export const SESSION_REDIRECT_URL = 'awx-redirect-url';
 export const PERSISTENT_FILTER_KEY = 'awx-persistent-filter';
 export const SESSION_USER_ID = 'awx-session-user-id';
+
+export const VERBOSITY = {
```
kialam: Good example of standardization here. If you wanted to just use one constant to represent Verbosity and it's permutations in the UI, I could see creating the object literal containing the `key`, `value`, and `label` keys and then calling a function to return the appropriate label when not used as an options dropdown.

```javascript
export const getVerbosity = (key) => VERBOSITY.filter(v => v.key === key)
const [{ label }] = getVerbosity(verbosityKey);
```

You could probably just save both the constant and the function as one object literal itself.

kialam (approved):
This works as-is, however if you wanted to use just one constant variable instead of having to maintain two I would suggest creating a mapper function to return the value (`label`) you want for the details views.

nixocio (dismissed):
Those strings are not being selected to be translated.

I followed up with @john-westcott-iv offline.

awx/ui/src/components/AdHocCommands/AdHocDetailsStep.js
```diff
@@ -296,7 +280,7 @@ function AdHocDetailsStep({ verbosityOptions, moduleOptions }) {
 
 AdHocDetailsStep.propTypes = {
   moduleOptions: PropTypes.arrayOf(PropTypes.array).isRequired,
-  verbosityOptions: PropTypes.arrayOf(PropTypes.object).isRequired,
+  // verbosityOptions: PropTypes.arrayOf(PropTypes.object).isRequired,
```
keithjgrant: delete?

awx/ui/src/components/AdHocCommands/AdHocCommandsWizard.js
```diff
@@ -79,7 +78,7 @@ const FormikApp = withFormik({
 FormikApp.propTypes = {
   onLaunch: PropTypes.func.isRequired,
   moduleOptions: PropTypes.arrayOf(PropTypes.array).isRequired,
-  verbosityOptions: PropTypes.arrayOf(PropTypes.object).isRequired,
+  // verbosityOptions: PropTypes.arrayOf(PropTypes.object).isRequired,
```
keithjgrant: delete?

keithjgrant (approved):

akus062381 (approved):

PR #12391, opened 2022-06-16
Do the grafana thing in docker-compose templating itself

AlanCoding:
##### SUMMARY
I'm following the pattern started by @john-westcott-iv here, and I find it vastly better quality-of-life than the existing idea of starting up independent containers.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API
 - Other

##### ADDITIONAL INFORMATION
So far tested and works great

```
PROMETHEUS=true GRAFANA=true MAIN_NODE_TYPE=hybrid EXECUTION_NODE_COUNT=0 COMPOSE_TAG=devel make docker-compose
```


kdelee (approved):

kdelee:
One issue I've had is tearing down gafana/prometheus to start over if I've changed the scrape config. not sure we have a make docker-compose-clean target for awx either though, so its not like that is a regression

AlanCoding:
I pushed another change that adds the volumes to `make docker-clean-volumes`, so I think this should address that concern.

rebeccahhh (approved):
with that addition I say :shipit:

tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
```diff
@@ -135,6 +135,37 @@ services:
...
+    links:
+      - awx_1:awx1
+{% endif %}
+{% if enable_grafana|bool %}
+  grafana:
+    image: grafana/grafana-enterprise:latest
+    container_name: tools_grafana_1
+    hostname: splunk
+    ports:
+      - "3001:3000"
```
fosterseth, line 158: for those wondering, I found that 3000 was conflicting with docker engine on my local machine, but which is why 3001 was chosen.

fosterseth (approved):

PR #12397, opened 2022-06-20
add state to awx license module

sean-m-sullivan:
##### SUMMARY
The API allows for removal of license, and the collections should reflect what the API allows. In that vein, adding a state to the license module. While the manifest option is no longer required, it should still fail if not supplied when state is present. 

[This issue was raised in this blog.](https://tekunabe.hatenablog.jp/entry/2022/06/20/181408)
[English translation link](https://tekunabe-hatenablog-jp.translate.goog/entry/2022/06/20/181408?_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=en)

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - Collection

##### AWX VERSION
```
21.1.0
```

##### ADDITIONAL INFORMATION
Testing was not added, as I was not sure how that would effect the testing environment. Can add it if requested. It has been tested.

### DIFF
--- a/awx_collection/plugins/modules/license.py
+++ b/awx_collection/plugins/modules/license.py
@@ -31,6 +31,12 @@ options:
           unlicensed or trial licensed.  When force=true, the license is always applied.
       type: bool
       default: 'False'
+    state:
+      description:
+        - Desired state of the resource.
+      default: "present"
+      choices: ["present", "absent"]
+      type: str
 extends_documentation_fragment: awx.awx.auth
 '''
 
@@ -40,6 +46,10 @@ EXAMPLES = '''
 - name: Set the license using a file
   license:
     manifest: "/tmp/my_manifest.zip"
+
+- name: Remove license
+  license:
+    state: absent
 '''
 
 import base64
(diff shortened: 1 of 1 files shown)


awx_collection/plugins/modules/license.py
```diff
@@ -50,13 +60,20 @@ def main():
 
     module = ControllerAPIModule(
         argument_spec=dict(
-            manifest=dict(type='str', required=True),
+            manifest=dict(type='str', required=False),
             force=dict(type='bool', default=False),
+            state=dict(choices=['present', 'absent'], default='present'),
```
john-westcott-iv, line 65: If state = present and I don't provide manifest (since its optional now) what happens? Should we include a test file now that we have combinations of options?

sean-m-sullivan: Right now it fails on 77-81 as we have a check for the manifest file.

sean-m-sullivan: I wasn't sure on the test file, which is why I made a comment about adding a test

john-westcott-iv: In Ansible there is a way to do "required_if". See workflow_job_template_node.py. That way, if state is present you can force manifest to be required. Right now the is probably pretty ugly. If we add that I'd say its functionally the same as it was before and no need to add a test at this time.

sean-m-sullivan: Agreed, Had not remembered that required if was a thing, and it gives a good error message on what wrong, so thanks for reminding me!, I think this looks good now.

john-westcott-iv (approved):

Issue #12406, opened 2022-06-22
Having multiple schedules of same name causes schedule module to fail


Tompage1994:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

Using the `awx.awx.schedule` module, having multiple schedules of the same name but for different unified job templates causes the module to fail even if you specify the `unified_job_template`

### AWX version

21.0.0

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [X] Collection
- [ ] CLI
- [ ] Other

### Installation method

N/A

### Modifications

no

### Ansible version

2.12.1

### Operating system

RHEL 8

### Web browser

Chrome

### Steps to reproduce

Create two schedules with the same name but for different UJTs. In this example, name them 'my_sched'
Delete one of the schedules using the module as below

```
    - awx.awx.schedule:
        name: my_sched
        unified_job_template: my_workflow
        state: absent
```

### Expected results

Playbook runs correctly and deletes the schedule for 'my_workflow' but leaves the other schedule

### Actual results

Fails to delete. An error message of `returned 2 items` comes back. Clearly it hasn't done the lookup for the UJT.

Full error here:

```
[...]

PR #12407, opened 2022-06-22
Allow modification of schedule if there are two of the same name

Tompage1994:
##### SUMMARY

Fixes issue where you couldn't modify or delete a schedule if two schedules of the same name exist.
related #12406  

##### ISSUE TYPE

 - Bug or Docs Fix

##### COMPONENT NAME

 - Collection
 
##### AWX VERSION

```
21.2.0
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx_collection/plugins/modules/schedule.py
+++ b/awx_collection/plugins/modules/schedule.py
@@ -105,7 +105,7 @@ options:
         - 5
     unified_job_template:
       description:
-        - Name of unified job template to schedule.
+        - Name of unified job template to schedule. Used to look up an already existing schedule.
       required: False
       type: str
     organization:
@@ -158,6 +158,12 @@ EXAMPLES = '''
         every: 1
         on_days: 'sunday'
         include: False
+
+- name: Delete 'my_schedule' schedule for my_workflow
+  schedule:
+    name: "my_schedule"
+    state: absent
+    unified_job_template: my_workflow
 '''
 
 from ..module_utils.controller_api import ControllerAPIModule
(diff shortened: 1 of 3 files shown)


awx_collection/plugins/modules/schedule.py
```diff
@@ -214,14 +214,16 @@ def main():
...
         inventory_id = module.resolve_name_to_id('inventories', inventory)
     search_fields = {}
+    sched_search_fields = {}
     if organization:
         search_fields['organization'] = module.resolve_name_to_id('organizations', organization)
     unified_job_template_id = None
     if unified_job_template:
         search_fields['name'] = unified_job_template
         unified_job_template_id = module.get_one('unified_job_templates', **{'data': search_fields})['id']
+        sched_search_fields['unified_job_template'] = unified_job_template_id
```
john-westcott-iv, line 230: Could we either add information in the docs (line 107) and/or add an example which states that the unified_job_template field will be used to search for the schedule if provided. For the use case of deleting a schedule where there are 2 with the same name that might be useful.

john-westcott-iv (changes requested):
Thanks for finding and fixing this issue!

Could we also add a test in `awx/awx_collection/tests/integration/targets/schedule/tasks` where you create two schedules with the same name and then try to delete one by name and assert that fails and then delete by name/unified_job_template and assert that works?

Tompage1994:
> 

I can do, however that wouldn't necessarily have failed in the way you expect with the current code.

Right now what happens if you create two of the same name using the modules it will just override the first with the second. Obviously the checking for a failure when no UJT is set will help us detect regression here though.

john-westcott-iv (approved):
Looks good from a code perspective, will wait for QE to add coverage before merging

Tompage1994:
@john-westcott-iv I managed to get my pytest working properly and have added a unit test which captures the issue and should prevent regression

PR #12414, opened 2022-06-22
Handle git submodules with recursive kwarg

AlanCoding:
##### SUMMARY
This is an alternative to https://github.com/ansible/awx/pull/12356 suggested by @sugitk 

I don't have a very clear opinion of this yet, but for now I thought I would share the diff an run related tests.

##### ISSUE TYPE
 - Bug or Docs Fix

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/tasks/jobs.py
+++ b/awx/main/tasks/jobs.py
@@ -1354,14 +1354,8 @@ class RunProjectUpdate(BaseTask):
                 branch=source_branch,
                 depth=1,
                 single_branch=True,  # shallow, do not copy full history
+                recursive=True,
             )
-            # submodules copied in loop because shallow copies from local HEADs are ideal
-            # and no git clone submodule options are compatible with minimum requirements
-            for submodule in git_repo.submodules:
-                subrepo_path = os.path.abspath(os.path.join(project_path, submodule.path))
-                subrepo_destination_folder = os.path.abspath(os.path.join(destination_folder, submodule.path))
-                subrepo_uri = Path(subrepo_path).as_uri()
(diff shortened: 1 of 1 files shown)


AlanCoding:
A case this fails on is https://github.com/ansible/test-playbooks/pull/85, and it seems to be at least somewhat legitimate.

```
fatal: '/var/lib/awx/projects/test-playbooks' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: clone of 'file:///var/lib/awx/projects/test-playbooks' into submodule path '/tmp/awx_325_7sxjnmaj/project/test-playbooks' failed
Failed to clone 'test-playbooks'. Retry scheduled
Cloning into '/tmp/awx_325_7sxjnmaj/project/test-playbooks'...
fatal: '/var/lib/awx/projects/test-playbooks' does not appear to be a git repository
fatal: Could not read from remote repository.
```

This definitely seems to be coming from the copy-via-clone into the job's project directory. We start with a project path `/var/lib/awx/projects/_68__project_appearancebowl`. It is true that, relative to that, taking a relative URL `../test-playbooks` will give the obviously bogus path of `/var/lib/awx/projects/test-playbooks`.

It seems that this was likely a valid bug that caused us to copy each submodule individually. We're just not doing the same thing as what this git `recursive=True` is trying to do. Indeed, I expect it may reach out to the _remote_ during the process, and that is the wrong thing to do.

AlanCoding:
Closing as this appears to be unworkable.

Issue #12400, opened 2022-06-21
awx intermittently reports non-complete playbook as failed and stops outputing log


tree4125:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

Shortly after starting a playbook using the AWX UI it is marked as failed. log output is incomplete and the job is still running in the background.

### AWX version

21.0.0

### Select the relevant components

- [X] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [X] Other

### Installation method

kubernetes

### Modifications

no

### Steps to reproduce

This has been happening randomly for the last few versions of AWX.

### Expected results

Playbook should be marked as successful and have full log output in AWX.

### Actual results

Log output incomplete or empty and playbook is marked as failed.

### Additional information

2022-06-09 17:55:18,024 ERROR    [7fffc6e497e7415c90df446bd2ff7720] awx.main.dispatch Worker failed to run task awx.main.tasks.jobs.RunJob(*[21317], **{}
Traceback (most recent call last):
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/jobs.py", line 587, in run
    self.final_run_hook(self.instance, status, private_data_dir, fact_modification_times)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/jobs.py", line 962, in final_run_hook
    super(RunJob, self).final_run_hook(job, status, private_data_dir, fact_modification_times)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/jobs.py", line 376, in final_run_hook
    instance.log_lifecycle("finalize_run")
AttributeError: 'NoneType' object has no attribute 'log_lifecycle'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/dispatch/worker/task.py", line 90, in perform_work
    result = self.run_callable(body)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/dispatch/worker/task.py", line 65, in run_callable
    return _call(*args, **kwargs)
[...]

PR #12415, opened 2022-06-22
Raise exception if update_model never recovers from problem

AlanCoding:
##### SUMMARY
Connect https://github.com/ansible/awx/issues/12400

This converts a return value of `None` to bubbling back up the exception. We believe that some exceptions are being suppressed and this fix is intended to make those discover-able somewhere.

##### ISSUE TYPE
 - Bug or Docs Fix

##### COMPONENT NAME
 - API


AlanCoding:
Testing seems to show this is working pretty well.

AlanCoding:
Too many other things are editing the same code, replacing with https://github.com/ansible/awx/pull/12527, which includes more error handling stuff.

Issue #12375, opened 2022-06-14
Invoking setup.py directly is deprecated


jbradberry:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

The Makefile in our development environment exposes a number of commands that make use of `python setup.py ...`.  This raises a deprecation warning in recent enough versions of Python and setuptools, and presumably will start breaking at some point.

See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html for more information.

For example, issuing `make VERSION`,

```
/usr/local/lib/python3.10/site-packages/setuptools/installer.py:27: SetuptoolsDeprecationWarning: setuptools.installer is deprecated. Requirements should be satisfied by a PEP 517 installer.
  warnings.warn(
/usr/local/lib/python3.10/site-packages/setuptools/installer.py:27: SetuptoolsDeprecationWarning: setuptools.installer is deprecated. Requirements should be satisfied by a PEP 517 installer.
  warnings.warn(
awx: 21.1.1.dev77+gc231e4d05e
```

### AWX version

N/A

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

docker development environment

### Modifications

no

### Steps to reproduce

N/A

### Expected results

N/A

### Actual results

N/A

PR #12428, opened 2022-06-24
Updated setup.py --version to python3 -m setuptools_scm.

djyasin:
##### SUMMARY

Invoking setup.py directly is deprecated #12375

##### ISSUE TYPE

 - Bug Fix PR

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.0.1
```


djyasin:
This is a PR in progress. Since these changes will touch multiple areas and require additional dependencies, we have decided to incrementally make commits as we update each term. This PR specifically updates: setup.py --version

djyasin:
Updated setup.py --version in this WIP PR. Given the scope of this change, we have opted to incrementally commit updates here as we find replacements for the deprecated terminology.

djyasin:
Can view full issue notes to date here: https://docs.google.com/document/d/18D46qJBMaPYi2mT5c7CJa7u0lfa8PaCjlj1Y5N3VC-U/edit

requirements/requirements_dev.txt
```diff
@@ -1,3 +1,4 @@
+build
```
Zokormazo, line 1: do we need this to build the rpm, or just for dev env?

tools/ansible/roles/dockerfile/templates/Dockerfile.j2
```diff
@@ -43,7 +43,7 @@ RUN dnf -y update && dnf install -y 'dnf-command(config-manager)' && \
     xmlsec1-devel \
     xmlsec1-openssl-devel
 
-RUN pip3 install virtualenv
+RUN pip3 install virtualenv setuptools_scm build
```
Zokormazo, line 46: The rest of the pip calls are using `python3 -m pip` . Should we align and do the same here?

Zokormazo (commented):
Makefile code allows to modify the python interpreter using `$PYTHON` , but the rest of the changes are hardcoding `python3` . Should we allow to modify the interpreter in playbooks and Dockerfiles?

requirements/requirements_dev.txt
shanemcd: This is only required for building the sdist, which we only do for the upstream builds.

tools/ansible/roles/dockerfile/templates/Dockerfile.j2
shanemcd: Meh

shanemcd:
@Zokormazo The Makefile is used within the Dockerfile to create /var/lib/awx/venv, so we know it will always get created with the intended version. For the build playbook - I think python3 is fine there because it's only grabbing the version. I dont see a reason to tie this to a particular python version. For the Dockerfile - I think it's fine because we know within the context of EL9 that will always be 3.9.

awx/__init__.py
```diff
@@ -6,9 +6,40 @@
...
+def get_version_from_file():
+    vf = version_file()
+    if vf:
+        with open(vf, 'r') as file:
+            return file.read().strip()
+
+
+def version_file():
+    current_dir = os.path.dirname(os.path.abspath(__file__))
+    version_file = os.path.join(current_dir, '..', 'VERSION')
```
AlanCoding, line 30: What lays down this version file and what's the expectation for using the development environment? Will it have a version file or not?

rooftopcellist (approved):

awx/__init__.py
rooftopcellist: The `VERSION` file is only expected to be present downstream.  The development environment should use setuptools_scm.get_version(), which will get the version based on the most recent git tag.

shanemcd (approved):

Zokormazo (approved):

Issue #11786, opened 2022-02-21
Allow <Detail/> to handle cases where there is no value


AlexSCorey:
##### ISSUE TYPE
Task

##### SUMMARY
Along the same lines of this PR https://github.com/ansible/awx/pull/11776 I think we should pass a key to <Detail/> so that it can handle whether it renders. Doing so would allow us to remove alot of `&& ( <Detail/>` phrases in our detail views and make those more readable.

shubhamranjan12:
Hello,
Can I work on this issue please ?

InnocentK:
Resolved

PR #12432, opened 2022-06-27
Enhanced detail component

vedaperi:
##### SUMMARY
Enhanced detail component to handle cases with no values, and refactored components that use detail component.
Addresses #11786

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui/src/components/DetailList/Detail.js
+++ b/awx/ui/src/components/DetailList/Detail.js
@@ -41,11 +41,16 @@ const Detail = ({
   className,
   dataCy,
   alwaysVisible,
+  isEmpty,
   helpText,
   isEncrypted,
   isNotConfigured,
 }) => {
   if (!value && typeof value !== 'number' && !alwaysVisible) {
+    return null;
+  }
+
+  if (isEmpty && !alwaysVisible) {
     return null;
   }
 
--- a/awx/ui/src/components/PromptDetail/PromptJobTemplateDetail.js
+++ b/awx/ui/src/components/PromptDetail/PromptJobTemplateDetail.js
@@ -26,7 +26,7 @@ function PromptJobTemplateDetail({ resource }) {
     extra_vars,
     forks,
     host_config_key,
-    instance_groups,
+    instance_groups = [],
     job_slice_count,
     job_tags,
     job_type,
@@ -94,9 +94,11 @@ function PromptJobTemplateDetail({ resource }) {
 
   return (
     <>
-      {summary_fields.recent_jobs?.length > 0 && (
-        <Detail value={<Sparkline jobs={recentJobs} />} label={t`Activity`} />
-      )}
+      <Detail
+        label={t`Activity`}
+        value={<Sparkline jobs={recentJobs} />}
+        isEmpty={summary_fields.recent_jobs?.length === 0}
+      />
       <Detail label={t`Job Type`} value={toTitleCase(job_type)} />
       {summary_fields?.organization ? (
         <Detail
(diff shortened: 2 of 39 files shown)


awx/ui/src/components/DetailList/Detail.js
```diff
@@ -49,6 +49,16 @@ const Detail = ({
     return null;
   }
 
+  if (typeof value == 'object' && typeof value.props.jobs == 'object' &&
+      value.props.jobs.length == 0) {
```
kialam: I'm not sure this logic works as intended. Javascript object length property is not accessed via dot notation and it would appear that this conditional never runs.

awx/ui/src/components/PromptDetail/PromptJobTemplateDetail.js
```diff
@@ -272,7 +266,7 @@ function PromptJobTemplateDetail({ resource }) {
             </ChipGroup>
           }
         />
-      )}
+      )}
```
kialam: stray space :)

awx/ui/src/screens/Job/JobDetail/JobDetail.js
```diff
@@ -428,6 +427,7 @@ function JobDetail({ job, inventorySourceLabels }) {
                 ))}
               </ChipGroup>
             }
+            isEmpty={credentials.length === 0}
```
keithjgrant, line 430: you should be able to use something like `isEmpty={!credentials || credentials.length === 0}` and get rid of the `credentials &&` check on line 408

keithjgrant: ...and similar for some of the other Details in this file

awx/ui/src/screens/Template/WorkflowJobTemplateDetail/WorkflowJobTemplateDetail.js
```diff
@@ -202,26 +201,25 @@ function WorkflowJobTemplateDetail({ template }) {
...
+              ouiaId="workflow-job-template-detail-label-chips"
+            >
+              {summary_fields.labels.results.map((l) => (
+                <Chip key={l.id} ouiaId={`${l.name}-label-chip`} isReadOnly>
+                  {l.name}
+                </Chip>
+              ))}
+            </ChipGroup>
+          }
+          isEmpty={summary_fields.labels?.results?.length === 0}
```
keithjgrant: if `summary_fields.labels` is null/undefined, this won't work correctly since `undefined === 0` is false. It's probably better to do `isEmpty={!summary_fields.labels?.results?.length}`

awx/ui/src/screens/WorkflowApproval/WorkflowApprovalDetail/WorkflowApprovalDetail.js
```diff
@@ -309,25 +309,24 @@ function WorkflowApprovalDetail({ workflowApproval }) {
...
+              ouiaId="wa-detail-label-chips"
+            >
+              {workflowJob.summary_fields.labels.results.map((label) => (
+                <Chip key={label.id} isReadOnly>
+                  {label.name}
+                </Chip>
+              ))}
+            </ChipGroup>
+          }
+          isEmpty={workflowJob?.summary_fields?.labels?.results?.length === 0}
```
keithjgrant: same as above

keithjgrant (approved):
Nice work!

InnocentK (approved):
Looks good!

PR #12435, opened 2022-06-28
Give more specific messages if a job was killed due to SIGTERM or SIGKILL signals

AlanCoding:
##### SUMMARY
The current behavior is that a dispatcher worker with a running job will not exit on SIGTERM. This means that it runs out the clock for its supervisor grace period and then gets a SIGKILL. Eventually the job still gets reaped anyway, but with potentially missing events and a vague message (it's not self-consistent after a SIGKILL, lost information).

The intent of this change is that we get snappy SIGTERM exits of the dispatcher, writes new messages for the different cases, and does some other minor things to make debugging easier.

##### ISSUE TYPE
 - Bug or Docs Fix

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
This is pulling in particular parts of https://github.com/ansible/awx/pull/11745, but not doing the architectural change of changing our canceling mechanism. This processes the SIGTERM and SIGINT signals and cancels jobs with a descriptive message. Those signals can come from a `supervisorctl stop` kind of command, or something else on the machine going rogue doing it by itself.

I've tried to apply everything I've learned in an appropriate manner. PR 11745 has some awkward code because it's passing around the objects related to signal handling. Fundamentally, the python `signal` library works with python globals, so we should also do so, but, take action to limit the context in which they apply. We also have a highly specific an opinionated use case with project syncs before job runs. If you look at `this_is_outermost_caller`, this is some additional work to ensure that nested applications of the decorator work with the same state.

### DIFF
--- a/awx/main/dispatch/reaper.py
+++ b/awx/main/dispatch/reaper.py
@@ -22,7 +22,7 @@ def startup_reaping():
         job_ids.append(j.id)
         j.status = 'failed'
         j.start_args = ''
-        j.job_explanation += 'Task was marked as running on system startup, so it has been marked as failed.'
+        j.job_explanation += 'Task was marked as running at system start up. The system must have previously not shut down properly, so it has been marked as failed.'
         j.save(update_fields=['status', 'start_args', 'job_explanation'])
         if hasattr(j, 'send_notification_templates'):
             j.send_notification_templates('failed')
--- a/awx/main/dispatch/reaper.py
+++ b/awx/main/dispatch/reaper.py
@@ -22,7 +22,7 @@ def startup_reaping():
         job_ids.ids.append(j.id)
         j.status = 'failed'
         j.start_args = ''
-        j.job_explanation += 'Task was stopped due to a service disruption.'
+        j.job_explanation += 'Task was marked as running on system startup, so it has been marked as failed.'
         j.save(update_fields=['status', 'start_args', 'job_explanation'])
         if hasattr(j, 'send_notification_templates'):
             j.send_notification_templates('failed')
--- a/awx/main/dispatch/reaper.py
+++ b/awx/main/dispatch/reaper.py
@@ -10,6 +10,24 @@ from awx.main.models import Instance, UnifiedJob, WorkflowJob
 logger = logging.getLogger('awx.main.dispatch')
 
 
+def startup_reaping():
+    """
+    If this particular instance is starting, then we know that any running jobs are invalid
+    so we will reap those jobs as a special action here
+    """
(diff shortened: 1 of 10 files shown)


kdelee (approved):

awx/main/dispatch/reaper.py
```diff
@@ -10,6 +10,27 @@
...
+    so we will reap those jobs as a special action here
+    """
+    me = Instance.objects.me()
+    jobs = UnifiedJob.objects.filter(status='running', controller_node=me.hostname)
+    job_ids = []
+    for j in jobs:
+        job_ids.append(j.id)
+        j.status = 'failed'
+        j.start_args = ''
+        j.job_explanation += 'Task was marked as running on system startup, so it has been marked as failed.'
```
kdelee:
```suggestion
        j.job_explanation += 'Task was marked as running at system start up. The system must have previously not shut down properly, so it has been marked as failed.'
```

AlanCoding: I might strike out the word "previously" but otherwise I'll take it.

awx/main/dispatch/reaper.py
```diff
@@ -10,6 +10,27 @@
...
+    so we will reap those jobs as a special action here
+    """
+    me = Instance.objects.me()
+    jobs = UnifiedJob.objects.filter(status='running', controller_node=me.hostname)
+    job_ids = []
+    for j in jobs:
+        job_ids.append(j.id)
+        j.status = 'failed'
+        j.start_args = ''
+        j.job_explanation += 'Task was marked as running at system start up. The system must have not shut down properly, so it has been marked as failed.'
```
shanemcd, line 25: 🍰

awx/main/dispatch/worker/base.py
```diff
@@ -169,8 +169,9 @@ def run(self, *args, **kwargs):
                     logger.exception(f"Error consuming new events from postgres, will retry for {self.pg_max_wait} s")
                     self.pg_down_time = time.time()
                     self.pg_is_down = True
-                if time.time() - self.pg_down_time > self.pg_max_wait:
-                    logger.warning(f"Postgres event consumer has not recovered in {self.pg_max_wait} s, exiting")
+                current_downtime = time.time() - self.pg_down_time
+                if current_downtime > self.pg_max_wait:
+                    logger.exception(f"Postgres event consumer has not recovered in {current_downtime} s, exiting")
```
shanemcd, line 174: 👍

shanemcd (approved):

PR #12437, opened 2022-06-28
[not devel] split into 3 parts - TaskManager, DependencyManager, and WorkflowManager

fosterseth:
##### SUMMARY

TaskManager - considers blocked jobs and capacity, and starts tasks
DependencyManager - spawns hard dependencies
WorkflowManager - analyzes running workflow jobs and spawns node jobs if ready to run


awx/main/utils/common.py
```diff
@@ -862,11 +862,13 @@ def ignore_inventory_computed_fields():
...
 def _schedule_task_manager():
-    from awx.main.scheduler.tasks import run_task_manager
+    from awx.main.scheduler.tasks import task_manager, dependency_manager, workflow_manager
     from django.db import connection
 
     # runs right away if not in transaction
-    connection.on_commit(lambda: run_task_manager.delay())
+    connection.on_commit(lambda: task_manager.delay())
+    connection.on_commit(lambda: dependency_manager.delay())
+    connection.on_commit(lambda: workflow_manager.delay())
```
AlanCoding, line 871: I could see some concern about getting into an infinite scheduling loop, or one that' takes a long time to resolve. I could see a safety argument for breaking `_schedule_task_manager` into 3 separate ones.

awx/main/scheduler/task_manager.py
```diff
@@ -53,103 +53,86 @@ def inner(*args, **kwargs):
...
-        system_jobs = [s for s in SystemJob.objects.filter(status__in=status_list).prefetch_related('instance_group')]
-        ad_hoc_commands = [a for a in AdHocCommand.objects.filter(status__in=status_list).prefetch_related('instance_group')]
-        workflow_jobs = [w for w in WorkflowJob.objects.filter(status__in=status_list)]
+        project_updates = [p for p in ProjectUpdate.objects.filter(**filter_args).filter(job_type='check').prefetch_related('instance_group')]
+        system_jobs = [s for s in SystemJob.objects.filter(**filter_args).prefetch_related('instance_group')]
+        ad_hoc_commands = [a for a in AdHocCommand.objects.filter(**filter_args).prefetch_related('instance_group')]
+        workflow_jobs = [w for w in WorkflowJob.objects.filter(**filter_args)]
         all_tasks = sorted(jobs + project_updates + inventory_updates + system_jobs + ad_hoc_commands + workflow_jobs, key=lambda task: task.created)
+        logger.debug(f"{self.prefix} {all_tasks}")
         return all_tasks
```
AlanCoding: So get_tasks is still shared between the subclasses? That seems like a lot of unnecessary stuff, as the workflow manager will only use workflows, etc.

awx/main/scheduler/task_manager.py
```diff
@@ -53,103 +53,86 @@ def inner(*args, **kwargs):
...
-                    dep.id,
-                )
-                task.save(update_fields=['status', 'job_explanation'])
-                task.websocket_emit_status('failed')
-                return dep
-
-        return None
+    def get_running_workflow_jobs(self):
+        graph_workflow_jobs = [wf for wf in WorkflowJob.objects.filter(status='running')]
+        return graph_workflow_jobs
```
kdelee, line 75: re: get_tasks being used in WorkflowManager
....from my reading all the workflow functions currently acutally use the workflows fetched from this function. so I think the WorkflowManager may  not need get_tasks re: what alan said

awx/main/scheduler/task_manager.py
fosterseth: currently the WorkflowManager doesn't call get_tasks, instead calls `get_running_workflow_jobs`

we could, for consistency, give WorkflowManager its own `get_tasks` which just returns get_running_workflow_jobs

fosterseth: each of the managers have a different _schedule() method, some of them call get_tasks, but WorkflowManager does not

awx/main/utils/common.py
fosterseth: `schedule_task_manager` is called in about 13 places in the code, and it's a little challenging to infer exactly which manager we should be invoking in those 13 calls. That is why I wrapped it up to call all three at once. But I see your concern regarding infinite loop, although in my local testing I don't see looping.

awx/settings/defaults.py
```diff
@@ -442,7 +442,9 @@ def IS_TESTING(argv=None):
         'options': {'expires': 50},
     },
     'gather_analytics': {'task': 'awx.main.tasks.system.gather_analytics', 'schedule': timedelta(minutes=5)},
-    'task_manager': {'task': 'awx.main.scheduler.tasks.run_task_manager', 'schedule': timedelta(seconds=20), 'options': {'expires': 20}},
+    'task_manager': {'task': 'awx.main.scheduler.tasks.task_manager', 'schedule': timedelta(seconds=20), 'options': {'expires': 20}},
+    'dependency_manager': {'task': 'awx.main.scheduler.tasks.dependency_manager', 'schedule': timedelta(seconds=20), 'options': {'expires': 20}},
+    'workflow_manager': {'task': 'awx.main.scheduler.tasks.workflow_manager', 'schedule': timedelta(seconds=20), 'options': {'expires': 20}},
```
AlanCoding, line 447: In the past, I've had proof-of-concepts that could pass all the integration tests without the task_manager schedule, itself. I don't expect it would be hard to remove the schedule for the workflow_manager, in particular. The task_manager could schedule the workflow_manager based on the existence of any active workflow jobs. We should eventually also have triggers from RunJob when a workflow job finishes, and things like that.

awx/main/scheduler/task_manager.py
```diff
@@ -255,72 +238,52 @@ def process_finished_workflow_jobs(self, workflow_jobs):
...
+                task.status = 'failed'
+                task.send_approval_notification('timed_out')
+                task.websocket_emit_status(task.status)
+                task.job_explanation = timeout_message
+                task.save(update_fields=['status', 'job_explanation', 'timed_out'])
 
-        dependent_tasks = dependent_tasks or []
+    @timeit
+    def get_tasks(self):
+        return self.get_running_workflow_jobs()
```
AlanCoding: Let me volunteer that, before you started on this, `get_running_workflow_jobs` was an atrocious method. Running workflow jobs were already fetched by `get_tasks` and this is one of those duplicate queries that @kdelee has discussed.

With this as a separate class, it could make more sense, but it still gets called in `process_pending_tasks`, which might be because you haven't finished yet. But I wanted to call it out.

awx/main/scheduler/task_manager.py
```diff
@@ -53,103 +53,86 @@ def inner(*args, **kwargs):
...
+                    # if sigterm due to timeout, still record metrics
+                    signal.signal(signal.SIGTERM, self.record_aggregate_metrics_and_exit)
+                    self._schedule()
+                    self.record_aggregate_metrics()
+                logger.debug(f"Finishing {self.prefix} Scheduler")
+
+
+class WorkflowManager(TaskBase):
+    def __init__(self):
+        self.prefix = "workflow_manager"
```
AlanCoding, line 122: I would prefer to define `prefix` on the class level. You don't even need to define `__init__` here.

awx/main/scheduler/task_manager.py
```diff
@@ -255,72 +238,52 @@ def process_finished_workflow_jobs(self, workflow_jobs):
...
-        task_actual = {
-            'type': get_type_for_model(type(task)),
-            'id': task.id,
-        }
-        dependencies = [{'type': get_type_for_model(type(t)), 'id': t.id} for t in dependent_tasks]
+    @timeit
+    def _schedule(self):
+        running_workflow_tasks = self.get_tasks()
+        if len(running_workflow_tasks) > 0:
+            self.process_finished_workflow_jobs(running_workflow_tasks)
```
AlanCoding: We hit on this topic a little bit, but may not be totally clear yet.

It would make the most sense to have `WorkflowManager` process pending workflow JTs and approval JTs. That involves moving it out of the process pending tasks method of the task manager class and introducing a new method here. That will allow this to re-use the same objects for all purposes (and hopefully axe some existing re-loading), and then the task manager and the workflow manager load (at least with get_tasks) a mutually exclusive set of objects.

awx/main/scheduler/task_manager.py
```diff
@@ -255,72 +238,52 @@ def process_finished_workflow_jobs(self, workflow_jobs):
...
-                task.instance_group = instance_group
-                execution_node_msg = f' and execution node {task.execution_node}' if task.execution_node else ''
-                logger.debug(
-                    f'Submitting job {task.log_format} controlled by {task.controller_node} to instance group {instance_group.name}{execution_node_msg}.'
-                )
-            with disable_activity_stream():
-                task.celery_task_id = str(uuid.uuid4())
-                task.save()
-                task.log_lifecycle("waiting")
+            self.spawn_workflow_graph_jobs(running_workflow_tasks)
```
AlanCoding, line 267: You know, @kdelee if you really want a hack with the `modified` time stamp that would help performance, I'd argue that you could update the `modified` time of a workflow job whenever a job inside of it finishes, and then this could only consider workflows with modified in the last 20 seconds for spawning or finishing. That would get rid of a lot of expensive reprocessing.

But I personally think the _most_ ideal solution would be for the job wrapup logic to trigger the workflow manager, _and pass in its workflow job id_ so you basically know you need to do something interesting, as opposed to running the graph logic on all workflows... every loop... which does nothing 99% of the time.

awx/main/scheduler/task_manager.py
```diff
@@ -481,6 +450,145 @@ def generate_dependencies(self, undeped_tasks):
 
         return created_dependencies
 
+    def process_tasks(self, all_sorted_tasks):
```
AlanCoding, line 441: seems like this `process_tasks` method isn't worth keeping around anymore.

awx/main/scheduler/task_manager.py
```diff
@@ -481,6 +450,145 @@ def generate_dependencies(self, undeped_tasks):
...
+        self.generate_dependencies(all_sorted_tasks)
+        self.subsystem_metrics.inc(f"{self.prefix}_pending_processed", len(all_sorted_tasks))
+
+    @timeit
+    def _schedule(self):
+        all_sorted_tasks = self.get_tasks(dict(status__in=["pending"], dependencies_processed=False))
+
+        if len(all_sorted_tasks) > 0:
+            self.all_inventory_sources = self.get_inventory_source_tasks(all_sorted_tasks)
```
AlanCoding, line 450: For performance, but also for code sanity, I think it would make most sense to create coherent in-memory data structures to lookup inventory sources... and repeat that pattern for projects as well (kind of a custom prefetch_related). This particular assignment is a big throw-my-computer-out-the-window kind of moment for me.

awx/main/scheduler/task_manager.py
```diff
@@ -481,6 +450,145 @@ def generate_dependencies(self, undeped_tasks):
...
+        # .schedule() run. Starting jobs is expensive, and there is code in place to reap
+        # the task manager after 5 minutes. At scale, the task manager can easily take more than
+        # 5 minutes to start pending jobs. If this limit is reached, pending jobs
+        # will no longer be started and will be started on the next task manager cycle.
+        self.start_task_limit = settings.START_TASK_LIMIT
+        self.time_delta_job_explanation = timedelta(seconds=30)
+        self.prefix = "task_manager"
+        super().__init__()
+
+    def after_lock_init(self, all_sorted_tasks):
```
AlanCoding, line 476: We could stand to rename some methods...

This is the method to initialize in-memory data structures.

awx/main/scheduler/task_manager.py
```diff
@@ -481,6 +450,145 @@ def generate_dependencies(self, undeped_tasks):
...
+                    errbacks=[{'task': handle_work_error.name, 'args': [task.celery_task_id], 'kwargs': {'subtasks': [task_actual] + dependencies}}],
+                )
+
+        task.websocket_emit_status(task.status)  # adds to on_commit
+        connection.on_commit(post_commit)
+
+    @timeit
+    def process_running_tasks(self, running_tasks):
+        for task in running_tasks:
+            self.dependency_graph.add_job(task)
```
AlanCoding, line 570: This, also, doesn't feel like it deserves to be a method. I also don't think it has the potential to consume any significant amount of time (as its just adding stuff to dicts). Correct me if I'm wrong.

awx/main/scheduler/task_manager.py
kdelee: @AlanCoding I had almost exactly the same thought last night! I think I know where we need to do this

PR #12445, opened 2022-06-30
Complex schedules UI

keithjgrant:
##### SUMMARY
Adds ability to set more complex patterns in schedule - multiple recurring patterns can be built together (e.g. an hourly AND a daily recurrence)

- ability to add exceptions is forthcoming (e.g. run daily EXCEPT on certain dates) (#12542)

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - UI


awx/ui/src/components/Schedule/shared/ScheduleForm.test.js
```diff
@@ -1096,40 +1233,64 @@ describe('<ScheduleForm />', () => {
...
+          .find('input#schedule-days-of-week-sat-frequencyOptions-week')
+          .prop('checked')
       ).toBe(false);
       expect(
         wrapper.find('DatePicker[aria-label="End date"]').prop('value')
       ).toBe('2021-01-01');
       expect(
         wrapper.find('TimePicker[aria-label="End time"]').prop('value')
-      ).toBe('1:00 AM');
+      ).toBe('12:00 AM');
```
keithjgrant, line 1293: I think this was a timezone/DST bug in the previous version...? America/New_York should be UTC-5:00 in January

keithjgrant:
~~This has some minor cleanup remaining, but~~ I'd like to get some eyes on it now. The logic to add "Exceptions" fields is still TODO in a follow-up PR (and a little bit of it is here now, just commented out)

AlexSCorey:
@keithjgrant I've started to look at this.  First, this work is amazing and is a great leap forward in helping customers automate more.  

I did want to note 1 thing:

- [ ] If you create and turn on a survey and create and turn on a schedule, then return to the survey to turn it off, and then return to edit the schedule you cannot edit the schedule without clicking on prompt.  This is a problem, because the survey is off and should not impede the ability to save/edit a schedule
-[ ] The repeat frequency field keeps track of how many items the user has clicked on, but now which items those are.  Shouldn't we show chips there? 
Screen Shot 2022-07-20 at 2 13 05 PM
- [x] Also looks like Repeat Frequency label is not marked for translated
Screen Shot 2022-07-20 at 2 22 29 PM
- [ ] I'd be cautious about using abbreviations for the days of the week because they may not translate properly
- [x] This frequency schedule (weekly and monthly, with repeats on both) results in a `Repeat frequency` of `yearly` on the details view, with the first and next run fields containing the wrong value.
Screen Shot 2022-07-20 at 2 24 39 PM
Screen Shot 2022-07-20 at 2 22 29 PM

akus062381:
feedback after manual testing:

- [ ] some time zones are showing the following text upon being selected on the form: `Warning: {selectedValue} is a link to {0} and will be saved as that.`
- [x] some time zones show up one way when selected in the form but change after the user saves the form. Example: America/Toronto saves and shows up as Canada/Eastern
- [x] consider providing more explicit information on the schedules detail page about what the user stipulated on the schedule form
- [x] in the form field on the following snippet, I feel like a better user experience would be for actual names of intervals to show up (minute, hour, etc)  instead of numbers 
image

kialam:
I'm seeing some errors in the console that we'll probably want to resolve. When trying to edit a management job (5, Cleanup Expired OAuth 2 Tokens https://localhost:3001/#/management_jobs/5/schedules/5)

1. When you select the repeat frequency dropdown I get this error:
Screen Shot 2022-07-20 at 2 26 25 PM

2. I cannot select a day when trying to create a weekly frequency and get this error:
Screen Shot 2022-07-20 at 2 24 21 PM

awx/ui/src/components/Schedule/shared/ScheduleFormFields.js
```diff
@@ -0,0 +1,194 @@
...
+      <FormGroup
+        name="frequency"
+        fieldId="schedule-frequency"
+        helperTextInvalid={frequencyMeta.error}
+        validated={
+          !frequencyMeta.touched || !frequencyMeta.error ? 'default' : 'error'
+        }
+        label={t`Repeat frequency`}
+      >
+        <FrequencySelect
```
kialam, line 113: When selecting multiple frequencies the `Repeat Frequency` detail in the Schedule Detail page may be a bit misleading:
Screen Shot 2022-07-20 at 3 18 19 PM

I have a schedule that repeats every min and hour, however the `Repeat Frequency` detail reads "every year."

keithjgrant:
Changes up and ready for review. I've updated the Details page with a full schedule breakdown for run frequencies. Some other thoughts/replies to specific comments below:

> * If you create and turn on a survey and create and turn on a schedule, then return to the survey to turn it off, and then return to edit the schedule you cannot edit the schedule without clicking on prompt.  This is a problem, because the survey is off and should not impede the ability to save/edit a schedule

@AlexSCorey This appears to be an existing bug in devel. I filed #12585 to track it

> * Also looks like Repeat Frequency label is not marked for translated

@AlexSCorey it is marked, but I haven't run extract-strings and compile-strings, so this should get picked up when we run them for the next release

> * The repeat frequency field keeps track of how many items the user has clicked on, but now which items those are.  Shouldn't we show chips there?

@AlexSCorey I tend to agree, though the current approach matches mockups & patternfly component for multiselect. We can definitely discuss/revisit later


> * some time zones are showing the following text upon being selected on the form: `Warning: {selectedValue} is a link to {0} and will be saved as that.`

@akus062381 This is a translation/string replacement issue. These are replacing normally in my local environment, though. Any idea why they would be different in your instance?


> I'm seeing some errors in the console that we'll probably want to resolve. When trying to edit a management job (5, Cleanup Expired OAuth 2 Tokens https://localhost:3001/#/management_jobs/5/schedules/5)

@kialam I couldn't reproduce these, though I suspect they are resulting from a particular format of a schedule created in previous version. I updated those two lines to handle undefined/null values.

mabashian (approved):
I'm going to approve this based on my manual testing.  I think the UX is definitely a vast improvement and I'm excited to be expanding the scheduler functionality for the community to cover more cases.  I'll defer to @AlexSCorey and @kialam on the actual code changes.

InnocentK (approved):
Great feature Keith! This looks much better now with the fixes you pushed. From QE side I think this is good to go now. Ignore those failing tests--a lot of those are just updates QE needs to make to our E2E tests to accommodate the changes from this PR. We should be pushing fixes to those failures in the near future

akus062381 (approved):

kialam (approved):
lgtm

Issue #12277, opened 2022-05-24
Critical - Inventory update blocks AWX


Tioborto:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

When we launch an update of an inventory source, after a while, our AWX instance is totally blocked. 
In other words, it is no longer possible to launch any job

Groups: 166
Hosts: 8614
Options: Overwrite enable

```
ansible-inventory 2.9.27.post0
  config file = None
  configured module search path = ['/home/runner/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.8/site-packages/ansible
  executable location = /usr/local/bin/ansible-inventory
  python version = 3.8.12 (default, Sep 21 2021, 00:10:52) [GCC 8.5.0 20210514 (Red Hat 8.5.0-3)]
No config file found; using defaults
setting up inventory plugins
host_list declined parsing /runner/project/inventory.py as it did not pass its verify_file() method
Parsed /runner/project/inventory.py inventory source with script plugin

   30.068 INFO     Processing JSON output...
   30.069 DEBUG    Loaded group: all
   30.069 DEBUG    Loaded group: subgroup
   30.070 DEBUG    Adding child group subgroup parent all
   ...
   37.861 DEBUG    Adding child group subgroup2 to parent all
   37.867 INFO     Loaded 166 groups, 8614 hosts
   217.170 DEBUG    Deleted host "h1"
   383.312 DEBUG    Deleted host "h2"
   537.445 DEBUG    Deleted host "h3"
   715.064 DEBUG    Deleted host "h4"
   886.845 DEBUG    Deleted host "h5"
  1059.130 DEBUG    Deleted host "h6"
```

Everything goes well when adding hosts (7 seconds for all hosts) but when deleting hosts that no longer exist, it takes a long time (around 150 seconds per host)

### AWX version

21.0.0

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs

### Installation method

kubernetes

### Modifications

no

### Ansible version

2.9

### Operating system

k8s v1.21.10

### Web browser

Safari

### Steps to reproduce

Source a large number of hosts in inventory with variables

### Expected results

All works fine :

[...]

Issue #12176, opened 2022-05-05
Deleting a host gives internal server error (500)


lutskevich:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

Can't delete hosts from awx inventories

### AWX version

21.0.0

### Select the relevant components

- [X] UI
- [X] API
- [ ] Docs

### Installation method

minikube

### Modifications

no

### Web browser

Chrome

### Steps to reproduce

Go to Resources > Host page
Create a host
After creating the host and being on the details page for that host, hit delete.

### Expected results

Host deleted and returned to host list

### Actual results

![image](https://user-images.githubusercontent.com/1433297/166867756-e6c35078-1c49-4929-b3d3-804eea68214e.png)

lutskevich:
deleting inventories with hosts too
![image](https://user-images.githubusercontent.com/1433297/166868386-4c3034fe-a8d7-44fa-b0a7-43d809b97848.png)

AlanCoding:
The server logs should have a traceback, and we really need to see that to go anywhere with issues like these.

lutskevich:
@AlanCoding 
```
2022-05-06 13:59:52,122 ERROR    [ab2af7283a484d05bd1dab6a90fba492] django.request Internal Server Error: /api/v2/hosts/2596/
Traceback (most recent call last):
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.OutOfMemory: out of shared memory
HINT:  You might need to increase max_locks_per_transaction.


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/lib64/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
[...]

PR #12447, opened 2022-06-30
Altering events relationship to hosts to increase performance

john-westcott-iv:
##### SUMMARY

Altering ForeignKey to an integer field for JobEvents.
Dropping cascade set null to do nothing on AdHocCommandEvent

There were reported issues around cascading on lockups and DB issues when deleting a host that was related to many events. This used to be a ForeignKey constraint that was "lost" at the DB layer when we partitioned the table. However, there was still a Django action to set the event hosts to null on an inventory delete.

This change should help prevent issue like that because there will be no cascading or setting null. Instead, we will just let the DB be referentially inconsistent. 

Because of the lack of foreign key, in order to keep the job_event detail page to have the host in its relations we had to add a lookup in the serializer and remove the related field from the access model as well as some other changes.

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.1.1.dev135+gdda69ca2d7.d20220630
```

##### ADDITIONAL INFORMATION

```

```


john-westcott-iv:
May address #12277 #12176 #10337

john-westcott-iv:
The failed api-schema is because host has changed from (id) to (integer).

awx/main/migrations/0165_alter_jobevent_host.py
```diff
@@ -0,0 +1,26 @@
...
+            model_name='adhoccommandevent',
+            name='host',
+            field=models.ForeignKey(
+                default=None, editable=False, null=True, on_delete=django.db.models.deletion.DO_NOTHING, related_name='ad_hoc_command_events', to='main.host'
+            ),
+        ),
+        migrations.AlterField(
+            model_name='jobevent',
+            name='host',
+            field=models.IntegerField(default=None, editable=False, null=True),
```
shanemcd: This has alarm bells going off for me. Does this operation happen at the table level or does it need to touch each row? Modifying the job event table can potentially affect millions of rows so we need to be extremely careful.

Would this help solve the problem without converting the field type? https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.ForeignKey.db_constraint

awx/main/models/events.py
```diff
@@ -794,12 +803,16 @@ class Meta:
         editable=False,
         db_index=False,
     )
+    # We need to keep this as an FK in the model because AdHocCommand uses a ManyToMany field
+    #   to hosts through adhoc_events. But in https://github.com/ansible/awx/pull/8236/ we
```
AlanCoding, line 802: Do we really need to keep that relationship at all? I can't actually find any places where it is used, and if it were to be used, there would likely be an alternative.

AlanCoding:
One thing that might be useful here is that you can change some Django settings to print the SQL it runs as it migrates. That might be helpful here, because the event table is massively huge in many deployments, and we want to identify anything that will affect the whole thing. Field re-definitions don't always result a database schema, so it's kind of hard to tell. We probably need to get some estimates of how long it will take as well.

shanemcd:
Can you make the PR title a little more descriptive? Remember that these end up in our release notes.

jbradberry (commented):
This approach feels wrong to me.  As pointed out by others, the job event tables are huge (and the fact that they are partitioned makes things even more complicated), so we probably want to avoid doing a database migration that will **certainly** lock the table for a long period of time.

It also seems weird that our historical modeling only has the relationship of hosts to ad hoc commands through the events table.  I assume that this is needed because you need to specify hosts to run the command on when firing it off, but haven't unraveled all of this to be certain.  I think as @AlanCoding seems to be suggesting that we should move away from that.

awx/main/migrations/0165_alter_jobevent_host.py
john-westcott-iv: In the migration sql it looks like this should only be renaming a column:
```
BEGIN;
--
-- Alter field host on adhoccommandevent
--
--
-- Alter field host on jobevent
--
DROP INDEX IF EXISTS "main_jobevent_host_id_idx1";
ALTER TABLE "main_jobevent" RENAME COLUMN "host_id" TO "host";
COMMIT;
```

That being said we could try leaving this as a ForeignKey and just applying the no-cascade option along with the db_contraint and that should give an identical effect (and likely less code changes).

awx/main/models/events.py
john-westcott-iv: For clarification, you don't see the Hosts field on AdHocCommand being used?

AlanCoding:
> It also seems weird that our historical modeling only has the relationship of hosts to ad hoc commands through the events table.

Why would this seem weird? We don't parse the provided `limit` field, thus, we only know experimentally whether an ad hoc command will operate on a host. This is kind of true for jobs too, but those are linked by doing extra work at the end of a job, creating the JobHostSummary objects. If an ad hoc command can't loop (which makes sense), then the extra intermediary of host summaries should be unnecessary, so this all makes sense to me.

Also, this PR has been changed substantially since the last round of comments. The `db_constraint` approach is interesting.

https://docs.djangoproject.com/en/4.0/ref/models/fields/

> ForeignKey.db_constraint
> 
> Controls whether or not a constraint should be created in the database for this foreign key. The default is True, and that’s almost certainly what you want; setting this to False can be very bad for data integrity. That said, here are some scenarios where you might want to do this:
> - You have legacy data that is not valid.
> - You’re sharding your database.
> 
> If this is set to False, accessing a related object that doesn’t exist will raise its DoesNotExist exception.

At a quick reading, those exceptions seem plausibly valid for our situation, and could be promising. We should look for some confirmation this is working from integration testing.

jay-steurer:
Tested and approved by QE
Clean yolo runs.

jay-steurer (approved):

awx/main/models/events.py
```diff
@@ -794,13 +798,18 @@ class Meta:
...
+    # We need to keep this as a FK in the model because AdHocCommand uses a ManyToMany field
+    #   to hosts through adhoc_events. But in https://github.com/ansible/awx/pull/8236/ we
+    #   removed the nulling of the field in case of a host going away before an event is saved
+    #   so this needs to stay SET_NULL on the ORM level
     host = models.ForeignKey(
         'Host',
         related_name='ad_hoc_command_events',
         null=True,
         default=None,
         on_delete=models.SET_NULL,
```
jbradberry, line 810: Why is this one still `SET_NULL`?

john-westcott-iv: I think initially, when I was going for IntegerFields, I was just going to leave this field alone because it had to remain a ForeignKey and its adhoc so I'd think there are less circumstances where people have a large number of events tied to a single host, and it was less impact for testing. But since we've altered paths with the `db_constraint` field I'd be ok with also changing this to SET_NULL we think its warranted.

AlanCoding:
@john-westcott-iv one last question, just to make sure. Did you test the case where you run a job, delete a host it touched, and then check that the output still works? If that's good, then I'm totally in favor of this.

AlanCoding:
Oh, sorry, I see that you've already documented that. Looks good.

AlanCoding (approved):

jbradberry (approved):

jainnikhil30:
@john-westcott-iv Just an update, I am working on testing this in our Perf and Scale environment with following number of job events in the DB:

- 10 million
- 50 million
- 1 billion 

Once I have DB populated, I am doing the migration with this change. So far 10 and 50 million looks fine: It takes about ~6-10 seconds to do the migration:

```
TASK [ansible.automation_platform_installer.automationcontroller : Migrate the database schema (may take awhile when upgrading).] ***
^[[0;33mchanged: [128.168.128.123] => {"changed": true, "cmd": ["awx-manage", "migrate", "--noinput"], "delta": "0:00:04.226824", "end": "2022-07-13 07:16:23.824669", "msg": "", "rc": 0, "start": "2022-07-13 07:16:19.597845", "stderr": "", "stderr_lines": [], "stdout": "Operations to perform:\n  Apply all migrations: auth, conf, contenttypes, main, oauth2_provider, sessions, sites, social_django, sso, taggit\nRunning migrations:\n  Applying main.0160_alter_schedule_rrule... OK\n  Applying main.0161_unifiedjob_host_status_counts... OK\n  Applying main.0162_alter_unifiedjob_dependent_jobs... OK\n  Applying main.0163_convert_job_tags_to_textfield... OK\n  Applying main.0164_remove_inventorysource_update_on_project_update... OK\n  Applying main.0165_alter_jobevent_host... OK", "stdout_lines": ["Operations to perform:", "  Apply all migrations: auth, conf, contenttypes, main, oauth2_provider, sessions, sites, social_django, sso, taggit", "Running migrations:", "  Applying main.0160_alter_schedule_rrule... OK", "  Applying main.0161_unifiedjob_host_status_counts... OK", "  Applying main.0162_alter_unifiedjob_dependent_jobs... OK", "  Applying main.0163_convert_job_tags_to_textfield... OK", "  Applying main.0164_remove_inventorysource_update_on_project_update... OK", "  Applying main.0165_alter_jobevent_host... OK"]}^[[0m 
```

john-westcott-iv:
When this PR is merged and released ask https://github.com/ansible/awx/issues/12520 to test.

jainnikhil30:
@john-westcott-iv 

Seems like we have an issue. This migration doesn't work now. 

`
TASK [ansible.automation_platform_installer.automationcontroller : Migrate the database schema (may take awhile when upgrading).] ***
fatal: [128.168.128.214]: FAILED! => {"changed": true, "cmd": ["awx-manage", "migrate", "--noinput"], "delta": "0:00:01.997372", "end": "2022-08-16 06:58:14.104424", "msg": "non-zero return code", "rc": 1, "start": "2022-08-16 06:58:12.107052", "stderr": "CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0165_alter_jobevent_host, 0165_task_manager_refactor in main).\nTo fix them run 'python manage.py makemigrations --merge'", "stderr_lines": ["CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0165_alter_jobevent_host, 0165_task_manager_refactor in main).", "To fix them run 'python manage.py makemigrations --merge'"], "stdout": "", "stdout_lines": []}
`


Looks like there was another 0165 (https://github.com/ansible/tower/blob/devel/awx/main/migrations/0165_task_manager_refactor.py) migration that went into devel. You probably have to rename this to 0166 ?

kdelee (approved):

kdelee:
Talking with @jainnikhil30 and he says he's tested with up to 500 million events and that he sees no problem with this

kdelee:
maybe you need to rebase? 
```
django.db.migrations.exceptions.NodeNotFoundError: Migration main.0166_alter_jobevent_host dependencies reference nonexistent parent node ('main', '0165_task_manager_refactor.py')
```

jainnikhil30:
With 500 million following is what we get:

Running migrations:
  Applying main.0165_task_manager_refactor... OK (0.359s)
  Applying main.0166_alter_jobevent_host... OK (0.212s)


This looks good to me. I am good with merging this.

Issue #1084, opened 2018-01-30
Feature idea: easier review of workflow output


wenottingham:
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
 - UI

##### SUMMARY

Right now, if you want to view the output of a workflow beyond success/failure at the job template level, the flow is:

1. Find workflow run in list
2. Click on it
3. Click on individual step
4. Read output
5. Hit back/up
6. Go to step 3
<loop until done>

This is a clunky workflow and requires many clicks to to progress through a full workflow. 
An alternate/different interface may make this easier.

##### ENVIRONMENT
* AWX version: any
* AWX install method: N/A
* Ansible version:  N/A
* Operating System: N/A
* Web Browser: N/A

wenottingham:
As part of this, we should have a better way to show the full workflow graph when it's large (download/print).

seansabour:
Would be nice to have a popup when you click on the details for that template that shows the output, instead of redirecting to another page.

AlexSCorey:
There is also a bug in this workflow.  When the user clicks on the node it navigates them to the Job Template Details view when it should navigate Job Template Output page.

trahman73:
Initial mockup idea. Need to iterate further...

![image](https://user-images.githubusercontent.com/13351591/175659107-06804ce7-73cb-4e02-b19a-17b3738ec3c9.png)

InnocentK:
@akus062381 We will need to add some stubs to improve test coverage to include this feature for our E2E tests

InnocentK:
Resolved

PR #12459, opened 2022-07-01
Easier review workflow output

AlexSCorey:
##### SUMMARY
Addresses #1084

##### ISSUE TYPE 
 - New or Enhanced Feature

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
OutputNav

### DIFF
--- a/awx/ui/src/components/RoutedTabs/RoutedTabs.js
+++ b/awx/ui/src/components/RoutedTabs/RoutedTabs.js
@@ -1,14 +1,24 @@
 import React from 'react';
 import { shape, string, number, arrayOf, node, oneOfType } from 'prop-types';
-import { Tab, Tabs, TabTitleText } from '@patternfly/react-core';
+import {
+  Tab as PFTab,
+  Tabs as PFTabs,
+  TabTitleText,
+} from '@patternfly/react-core';
 import { useHistory, useLocation } from 'react-router-dom';
 import styled from 'styled-components';
 
-const Wrapper = styled.div`
-  display: flex;
+const Tabs = styled(PFTabs)`
+  & > ul {
+    flex-grow: 1;
+  }
 `;
 
-function RoutedTabs({ tabsArray, isWorkflow, children }) {
+const Tab = styled(PFTab)`
+  ${(props) => props.hasstyle && `${props.hasstyle}`}
+`;
+
+function RoutedTabs({ tabsArray }) {
   const history = useHistory();
   const location = useLocation();
 
@@ -36,28 +46,7 @@ function RoutedTabs({ tabsArray, isWorkflow, children }) {
       history.push(link);
     }
   };
-  return children ? (
-    <Wrapper>
-      <Tabs
-        activeKey={getActiveTabId()}
-        onSelect={handleTabSelect}
-        ouiaId="routed-tabs"
-      >
-        {tabsArray.map((tab) => (
-          <Tab
-            aria-label={typeof tab.name === 'string' ? tab.name : null}
-            eventKey={tab.id}
-            key={tab.id}
-            href={`#${tab.link}`}
-            title={<TabTitleText>{tab.name}</TabTitleText>}
-            aria-controls=""
-            ouiaId={`${tab.name}-tab`}
-          />
-        ))}
-      </Tabs>
-      {children}
-    </Wrapper>
-  ) : (
+  return (
     <Tabs
       activeKey={getActiveTabId()}
       onSelect={handleTabSelect}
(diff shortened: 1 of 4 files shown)


awx/ui/src/components/WorkflowOutputNavigation/WorkflowOutputNavigation.js
```diff
@@ -0,0 +1,120 @@
...
+      }}
+      selections={filterBy}
+      onSelect={(e, v) => {
+        if (v !== 'Failed' && v !== 'Successful') return;
+        handleFilter(v);
+      }}
+      isOpen={isOpen}
+      isGrouped
+      hasInlineFilter
+      placeholderText={t`Workflow Job 1/${relevantResults.length}`}
```
keithjgrant, line 76: should this read 1/6, 2/6, 3/6 etc based on which job is currently being viewed?

awx/ui/src/screens/Job/Job.js
```diff
@@ -147,8 +168,11 @@ function Job({ setBreadcrumb }) {
 
   return (
     <PageSection>
-      <Card>
-        <RoutedTabs tabsArray={tabsArray} />
+      <Card id="jobOuputCard">
```
keithjgrant:
```suggestion
      <Card id="jobOutputCard">
```

awx/ui/src/screens/Job/Job.js
```diff
@@ -147,8 +168,11 @@ function Job({ setBreadcrumb }) {
...
+        <RoutedTabs
+          isWorkflow={match.url.startsWith('/jobs/workflow')}
```
keithjgrant: Do we need to pass this `isWorkflow` prop? It looks like it isn't used in `RoutedTabs`

awx/ui/src/components/RoutedTabs/RoutedTabs.js
```diff
@@ -43,10 +57,11 @@ function RoutedTabs({ tabsArray }) {
           aria-label={typeof tab.name === 'string' ? tab.name : null}
           eventKey={tab.id}
           key={tab.id}
-          href={`#${tab.link}`}
+          href={!tab.hasstyle && `#${tab.link}`}
```
keithjgrant, line 60: seems a little odd to hide/nullify the href if `hasstyle` is truthy. Maybe just do it based on whether `tab.link` is truthy instead?

awx/ui/src/components/WorkflowOutputNavigation/WorkflowOutputNavigation.js
```diff
@@ -0,0 +1,120 @@
...
+      chipGroupComponent={
+        <ChipGroup numChips={1} totalChips={1}>
+          <Chip key={filterBy} onClick={() => handleFilter(filterBy)}>
+            {[filterBy]}
+          </Chip>
+        </ChipGroup>
+      }
+    >
+      {[getOptions()]}
```
keithjgrant: not sure why this needs to be returned from a function. Can we just inline the `SelectGroup` JSX here in-place?

awx/ui/src/components/WorkflowOutputNavigation/WorkflowOutputNavigation.js
```diff
@@ -0,0 +1,120 @@
...
+  project_update: 'project',
+  system_job: 'management',
+  system: 'system_job',
+  inventory_update: 'inventory',
+  workflow_job: 'workflow',
+};
+
+function WorkflowOutputNavigation({ relatedJobs }) {
+  const { id } = useParams();
+  const parent = document.getElementById('jobOuputCard');
```
keithjgrant: this feels wrong, using getElementById, especially with a hard-coded id string (i.e. this component is dependent on particular internals of a parent component and will break if those change). Can this be done by passing in a ref instead?

keithjgrant (changes requested):
This is very cool! Nice work. I noted a few items for cleanup, though most are tiny nit-picky things apart from the card-coded id/ref idea

awx/ui/src/components/WorkflowOutputNavigation/WorkflowOutputNavigation.js
AlexSCorey: I went around several times with this, and I landed on this pattern because workflows are not always linear so how does someone tell if the output being reviewed is #3 in a workflow, or some other number?  I landed on `1/n` pattern simply as a way to tell the user you are looking a 1 node output, and there are `n` nodes total.

awx/ui/src/components/WorkflowOutputNavigation/WorkflowOutputNavigation.js
AlexSCorey: hadn't thought of the ref.  I'll explore

mabashian:
@AlexSCorey I haven't pulled this down so apologies if I'm describing how this already works but I wanted to ask about node aliases.  In that dropdown, it looks like a user can select a node.  Nodes can have an alias (field is `identifier`).  Are we showing the alias if it's available?  I think that would be extremely useful.  You can reference the workflow node component to see some logic around when we show the alias and when we don't.  That field is filled with a uuid even when an alias is not provided so there's some additional logic needed to detect that.

keithjgrant (approved):

InnocentK (approved):
Looks good! Great feature!

PR #12470, opened 2022-07-06
[task_manager_refactor] Time out 3 managers

fosterseth:
##### SUMMARY

have dispatcher timeout each of the 3 managers after 5 minutes

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/dispatch/pool.py
+++ b/awx/main/dispatch/pool.py
@@ -401,13 +401,15 @@ class AutoscalePool(WorkerPool):
                 # the task manager to never do more work
                 current_task = w.current_task
                 if current_task and isinstance(current_task, dict):
-                    if current_task.get('task', '').endswith('tasks.run_task_manager'):
+                    endings = ['tasks.task_manager', 'tasks.dependency_manager', 'tasks.workflow_manager']
+                    current_task_name = current_task.get('task', '')
+                    if any([current_task_name.endswith(e) for e in endings]):
                         if 'started' not in current_task:
                             w.managed_tasks[current_task['uuid']]['started'] = time.time()
                         age = time.time() - current_task['started']
                         w.managed_tasks[current_task['uuid']]['age'] = age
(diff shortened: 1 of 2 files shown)


AlanCoding (approved):
This is the obvious fix. It's mainly needed for the workflow_manager.

Longer-term, to really address issues users see with the task manager, I would like to change this so that the timeout causes it to gracefully exit and commit its work in-progress. This would have to be done at the code level in the task manager itself, and we may still want to keep the timeouts at a higher threshold.

kdelee (approved):

awx/main/dispatch/pool.py
```diff
@@ -401,13 +401,15 @@ def cleanup(self):
...
+                    endings = ['tasks.task_manager', 'tasks.dependency_manager', 'tasks.workflow_manager']
+                    current_task_name = current_task.get('task', '')
+                    if any([current_task_name.endswith(e) for e in endings]):
                         if 'started' not in current_task:
                             w.managed_tasks[current_task['uuid']]['started'] = time.time()
                         age = time.time() - current_task['started']
                         w.managed_tasks[current_task['uuid']]['age'] = age
                         if age > (settings.TASK_MANAGER_TIMEOUT + settings.TASK_MANAGER_TIMEOUT_GRACE_PERIOD):
-                            logger.error(f'run_task_manager has held the advisory lock for {age}, sending SIGTERM to {w.pid}')  # noqa
+                            logger.error(f'{current_task_name} has held the advisory lock for >5m, sending SIGTERM to {w.pid}')  # noqa
```
AlanCoding: I don't see why you wouldn't keep the `{age}`.

If someone is debugging the situation, they can manually run `settings.TASK_MANAGER_TIMEOUT + settings.TASK_MANAGER_TIMEOUT_GRACE_PERIOD`, but only the log can tell them how long it actually ran at this point.

fosterseth: yeah merge conflict mistake, i fixed it up!

Issue #12259, opened 2022-05-19
Animate in new jobs in jobs list


keithjgrant:
When a new job is added to the jobs list based on websocket activity, it would be nice to animate/expand the row into place rather than jump the rest of the list downward in an instant. This could prevent potential mis-clicks.

kialam:
This would be a great example to try out React's [ReactCSSTransitionGroup](https://reactjs.org/docs/animation.html#high-level-api-reactcsstransitiongroup). What's nice about this API is that React handles the logic needed to determine whether something has been added or removed from the DOM and assigns the relevant class names we can then target with CSS.

PR #12474, opened 2022-07-06
Animate new jobs in jobs list

kialam:
##### SUMMARY

Related #12259 

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - UI
Animate Jobs List

### DIFF
--- a/awx/ui/src/components/JobList/JobList.js
+++ b/awx/ui/src/components/JobList/JobList.js
@@ -1,4 +1,5 @@
-import React, { useEffect, useCallback } from 'react';
+import React, { useEffect, useCallback, useRef } from 'react';
+import { TransitionGroup } from 'react-transition-group';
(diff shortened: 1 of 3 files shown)


mabashian:
@kialam I'm trying to figure out what the expected behavior here is.  My assumption was that when a new row is added, it would get animated but once a row is rendered, only the status would change.  I'm not 100% sure what's happening but I see rows re-animating when there's a large-ish number of jobs coming in.  Best way to set this up would be with a workflow (probably set up with a schedule).  Here's what my workflow looks like:

Screen Shot 2022-08-01 at 11 15 16 AM

and here's what the jobs list looks like while the jobs are coming in/being updated:

animated_rows

kialam:
Closing for now as the behavior we expect isn't quite there.

PR #12475, opened 2022-07-07
[not devel][task_manager_refactor] fixup starting pending workflow jobs

kdelee:
@AlanCoding noticed that we forgot to send notifications, websocket messages, or setting job explanation.

This PR addresses:

- [x] send notifications and websocket messages for starting workflows
- [x] drop unused logic in dependency graph regarding workflows and allow_simultaneous
- [x] filter out workflows and workflow approvals from UnifiedJob query using ctype
- [x] set job explanation on wfs blocked by allow_simultaneous

### DIFF
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -15,6 +15,7 @@ from django.db import transaction, connection
 from django.utils.translation import gettext_lazy as _, gettext_noop
 from django.utils.timezone import now as tz_now
 from django.conf import settings
+from django.contrib.contenttypes.models import ContentType
 
 # AWX
 from awx.main.dispatch.reaper import reap_job
@@ -78,14 +79,20 @@ class TaskBase:
             return True
         return False
 
-    def get_running_workflow_jobs(self):
-        graph_workflow_jobs = [wf for wf in WorkflowJob.objects.filter(status='running')]
-        return graph_workflow_jobs
-
     @timeit
     def get_tasks(self, filter_args):
-        qs = UnifiedJob.objects.filter(**filter_args).exclude(launch_type='sync').order_by('created').prefetch_related('instance_group')
-        return [task for task in qs if not type(task) is WorkflowJob]
+        # We exclude workflows (both from WorkflowJobTemplates as well as sliced jobs)
+        workflow_ctype_id = ContentType.objects.get_for_model(WorkflowJob).id
+        workflow_approval_ctype_id = ContentType.objects.get_for_model(WorkflowApproval).id
+        qs = (
+            UnifiedJob.objects.filter(**filter_args)
+            .exclude(launch_type='sync')
+            .exclude(polymorphic_ctype=workflow_ctype_id)
+            .exclude(polymorphic_ctype=workflow_approval_ctype_id)
+            .order_by('created')
+            .prefetch_related('instance_group')
+        )
+        return [task for task in qs]
 
     def record_aggregate_metrics(self, *args):
         if not settings.IS_TESTING():
@@ -267,7 +274,16 @@ class WorkflowManager(TaskBase):
     def get_tasks(self):
         workflow_jobs_running = [wf for wf in WorkflowJob.objects.filter(status='running')]
         workflow_jobs_pending = [wf for wf in WorkflowJob.objects.filter(status='pending')]
+        return workflow_jobs_running, workflow_jobs_pending
+
+    def start_pending_workflows(self, workflow_jobs_running, workflow_jobs_pending):
+        """Take in list of currently pending and running workflows.
+
+        Start the workflows that we can and send any notifications or websockets needed.
+        Return list of the now running workflows.
+        """
         workflow_to_start = []
+        workflow_blocked = []
         running_wfjt_ids = {wf.unified_job_template_id for wf in workflow_jobs_running}
         for wf in workflow_jobs_pending:
             if wf.allow_simultaneous or wf.unified_job_template_id not in running_wfjt_ids:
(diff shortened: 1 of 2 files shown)


awx/main/scheduler/task_manager.py
```diff
@@ -249,20 +266,44 @@ def timeout_approval_node(self):
...
         for wf in workflow_jobs_pending:
-            if wf.allow_simultaneous or wf.pk not in running_workflow_pk:
+            if wf.allow_simultaneous or wf.unified_job_template_id not in running_wfjt_ids:
                 wf.status = 'running'
                 workflow_to_start.append(wf)
+                running_wfjt_ids.add(wf.unified_job_template_id)
+                logger.debug('Transitioning %s to running status.', wf.log_format)
+                self.start_task_limit -= 1
+                if self.start_task_limit == 0 or self.timed_out():
+                    break
```
shanemcd, line 299: May be be prudent to split out this condition and add logging.

awx/main/scheduler/task_manager.py
```diff
@@ -230,6 +243,10 @@ def timeout_approval_node(self):
         workflow_approvals = WorkflowApproval.objects.filter(status='pending')
         now = tz_now()
         for task in workflow_approvals:
+            if self.timed_out():
```
shanemcd: Consider adding logging either here or in timed_out

awx/main/scheduler/task_manager.py
```diff
@@ -66,17 +67,31 @@ def __init__(self):
...
+        # We exclude workflows (both from WorkflowJobTemplates as well as sliced jobs)
+        workflow_ctype_id = ContentType.objects.get_for_model(WorkflowJob).id
+        qs = (
+            UnifiedJob.objects.filter(**filter_args)
+            .exclude(launch_type='sync')
+            .exclude(polymorphic_ctype=workflow_ctype_id)
+            .order_by('created')
+            .prefetch_related('instance_group')
+        )
         return [task for task in qs if not type(task) is WorkflowJob]
```
fosterseth: is this conditional on WorkflowJob necessary if we do `.exclude(polymorphic_ctype=workflow_ctype_id)` above?

awx/main/scheduler/task_manager.py
```diff
@@ -66,17 +67,31 @@ def __init__(self):
         # ensures each task manager metric will be overridden when pipe_execute
         # is called later.
         self.subsystem_metrics = s_metrics.Metrics(auto_pipe_execute=False)
+        self.start_time = tz_now()
```
AlanCoding: Why would you use timezone time as opposed to epoch time? For measuring time differences locally, the `time.time()` should be faster, and it's pretty simple by comparision.

kdelee: ok cool

awx/main/scheduler/task_manager.py
```diff
@@ -249,20 +272,47 @@ def timeout_approval_node(self):
     def get_tasks(self):
         workflow_jobs_running = [wf for wf in WorkflowJob.objects.filter(status='running')]
         workflow_jobs_pending = [wf for wf in WorkflowJob.objects.filter(status='pending')]
+        return workflow_jobs_running, workflow_jobs_pending
+
+    def start_pending_workflows(self, workflow_jobs_running, workflow_jobs_pending):
+        """Take in list of currently pending and running workflows.
+
+        Start the workflows that we can and send any notifications or websockets needed.
+        Return list of the now running workflows.
```
AlanCoding, line 283: Something isn't clicking for me here. We both saw that this problem is solved twice, here and by the dependency graph. What I don't understand is why you would keep _this_ logic, and remove workflows from the dependency_graph construct.

Maybe that's not clear enough... so re-framing...

Why should the `WorkflowManager` process "pending" workflows? Wouldn't this be better addressed in the main task manager, and have the workflow manager be specialized to spawning workflow jobs and processing finished workflows? The blocking logic for pending jobs is very much the domain of the main task manager.

Consider that sliced jobs probably don't respect project `update_on_launch` correctly. That introduces a dependency between pending workflow jobs and project updates.. or at least it would if we did it correctly. By having a separate class process pending workflow jobs, you're closing the door on fixing that competently.

awx/main/scheduler/task_manager.py
```diff
@@ -249,20 +272,47 @@ def timeout_approval_node(self):
...
+                running_wfjt_ids.add(wf.unified_job_template_id)
+                logger.debug('Transitioning %s to running status.', wf.log_format)
+                self.start_task_limit -= 1
+                if self.start_task_limit == 0:
+                    break
+                if self.timed_out():
+                    logger.warning(f"Workflow manager has reached time out processing pending workflows, exiting loop early")
+                    break
+            else:
+                logger.debug('Workflow %s staying in pending, blocked by another running workflow from the same workflow job template', wf.log_format)
```
AlanCoding: Likewise, this doesn't correctly set `job_explanation` with a blurb about the blocking condition. We _currently_ do that, so by choosing this solution over the dependency graph, you introduce a regress unless you add more logic back in. I'd rather not do that.

awx/main/scheduler/task_manager.py
```diff
@@ -249,20 +272,47 @@ def timeout_approval_node(self):
...
 
         WorkflowJob.objects.bulk_update(workflow_to_start, ['status'])
+
+        # We could do the following in the previous loop
+        # Doing it in a loop after the bulk_update preserves previous behavior
+        # of not sending the notifications until after the status of running has been saved.
+        for wf in workflow_to_start:
+            wf.send_notification_templates('running')
+            wf.websocket_emit_status(wf.status)
```
AlanCoding, line 313: This is done in a transaction, and you can have race conditions if these actions aren't delayed as `on_commit` methods. But again, this is starting to feel like re-inventing the wheel when the task manager already does this stuff more correctly.

awx/main/scheduler/task_manager.py
kdelee: My reasoning was:
 - Do all workflow stuff in the WorkflowManager
 - update_on_launch behavior gets resolved when actual jobs of the sliced job workflow get created -- as you said, this is how it works now, nothing in current logic considers this. To me, that makes sense. Workflows are just a container for actual jobs. No need to update_on_launch to start the container, the inner nodes are the ones that matter.
 - only dependency ever considered in the DependencyGraph was allow_simultaneous
 - We were not build the DependencyGraph in this class so I retained the allow_simultaneous logic that was previously scattered through process_pending_jobs and start_task
 
My main goal was to have  a KISS solution where only the WorkflowManager managed workflows. I'm not beyond convincing.

kdelee:
we've decided that we will not handle pending workflow jobs in the WorkflowManager. Will open a new PR with the desired changes next week

PR #12480, opened 2022-07-07
Translate contents of Hosts Automated field as a single string

mabashian:
##### SUMMARY
Translate the entire string rather than just 'since'.  This is necessary for languages where the date should come before the number of hosts.

Here's what it looks like in practice.  I don't know Japanese so I just translated 'since' to 'sometranslatedtext':

hosts_automated

Here's what the sample .po looks like:

Screen Shot 2022-07-07 at 12 42 06 PM

##### ISSUE TYPE
 - Bug or Docs Fix

##### COMPONENT NAME
 - UI


keithjgrant (approved):

InnocentK (approved):
Looks good!

PR #12489, opened 2022-07-07
Remove external script call to D3.js.

kialam:
##### SUMMARY

We are seeing this bug in our production builds: https://github.com/facebook/create-react-app/issues/12503. Removing `homepage: "."` from `package.json` correctly fixes the script routing and properly loads D3.js from our vendor bundle.

##### ISSUE TYPE

 - Bug or Docs Fix

##### COMPONENT NAME

 - UI


awx/ui/src/util/simulationWorker.js
```diff
@@ -1,21 +1,22 @@
-/* eslint-disable no-undef */
-importScripts('https://d3js.org/d3-collection.v1.min.js');
```
gundalow, line 2: In a follow up PR (so we don't block this fix) would we want CI to error if it finds `importScripts`?

keithjgrant (approved):

kialam:
Worked with @shanemcd on this - it looks like removing `homepage: '.'` may not be an option as that disables running awx from a non-root path. I will need to look at workarounds to my original solution.

awx/ui/src/util/simulationWorker.js
kialam: No because you can use importScripts to pull in local JS files, which is part of the fix in this PR.

PR #12490, opened 2022-07-08
[task_manager_refactor] add schedule dependency and workflow managers

fosterseth:
##### SUMMARY

change all of the `schedule_task_manager` calls in the code to invoke the one of the 3 managers

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### Notes

the bulk reschedule seems to work fine inside of the class

here is a script I ran in awx shell to demonstrate that the flags are set correctly outside and inside the context manager

```
In [13]: print(ScheduleTaskManager().manager_threading_local.__dict__)
{'bulk_reschedule': False, 'needs_scheduling': False}

In [14]: with ScheduleTaskManager().task_manager_bulk_reschedule():
    ...:     print(ScheduleTaskManager().manager_threading_local.__dict__)
    ...:     ScheduleTaskManager().schedule()
    ...:     print(ScheduleTaskManager().manager_threading_local.__dict__)
    ...: 
{'bulk_reschedule': True, 'needs_scheduling': False}
2022-07-08 16:16:03,432 DEBUG    [-] awx.main.utils returning
{'bulk_reschedule': True, 'needs_scheduling': True}
```

### DIFF
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -93,7 +93,7 @@ from awx.main.utils import (
     get_object_or_400,
     getattrd,
     get_pk_from_dict,
-    schedule_task_manager,
+    ScheduleWorkflowManager,
     ignore_inventory_computed_fields,
 )
 from awx.main.utils.encryption import encrypt_value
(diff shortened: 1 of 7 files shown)


AlanCoding (approved):

awx/api/views/__init__.py
```diff
@@ -3391,7 +3391,7 @@ def post(self, request, *args, **kwargs):
         obj = self.get_object()
         if obj.can_cancel:
             obj.cancel()
-            schedule_task_manager()
+            ScheduleWorkflowManager().schedule()
```
kdelee, line 3394: did you import the right thing in this file? I see `ScheduleTaskManager` imported here

kdelee (approved):

PR #12500, opened 2022-07-11
Task manager refactor revert pending wfs change

kdelee:
##### SUMMARY

We agreed last week that we want the TaskManager to deal with all pending tasks to keep all "soft blocking" logic contained to the DependencyGraph.

This does:
- [x] fix blocked_by logic for sliced jobs in depedency graph that was looking at wrong template ID
- [x] cherry-pick a fix @AlanCoding has for a fix to adding sliced jobs to the DepedencyGraph
- [x] move logic for moving pending workflows to running back to task manager
- [x] filter WorkflowApproval jobs out of the main get_tasks query so we are not grabbing them in DepdencyManager and TaskManager 
- [x] make get_tasks and related member functions set the lists of tasks as attributes on the classes and stop passing around the list of tasks
- [x] move calls to add newly running jobs to DepedencyGraph to start_task so we do it the same way for everything.

What I've tested (manually):
- [x] WorkflowJobs respect allow_simultaneous
- [x] SlicedJobs respect allow_simultaneous
- [x] regular jobs still start correctly
- [x] project update_on_launch still working

### DIFF
--- a/awx/main/scheduler/dependency_graph.py
+++ b/awx/main/scheduler/dependency_graph.py
@@ -7,6 +7,11 @@ from awx.main.models import (
     WorkflowJob,
 )
 
+import logging
+
+
+logger = logging.getLogger('awx.main.scheduler.dependency_graph')
+
 
 class DependencyGraph(object):
     PROJECT_UPDATES = 'project_updates'
@@ -36,6 +41,9 @@ class DependencyGraph(object):
         self.data[self.WORKFLOW_JOB_TEMPLATES_JOBS] = {}
 
     def mark_if_no_key(self, job_type, id, job):
+        if id is None:
+            logger.warning(f'Null dependency graph key from {job}, could be integrity error or bug, ignoring')
+            return
         # only mark first occurrence of a task. If 10 of JobA are launched
         # (concurrent disabled), the dependency graph should return that jobs
         # 2 through 10 are blocked by job1
@@ -66,7 +74,10 @@ class DependencyGraph(object):
         self.mark_if_no_key(self.JOB_TEMPLATE_JOBS, job.job_template_id, job)
 
     def mark_workflow_job(self, job):
-        self.mark_if_no_key(self.WORKFLOW_JOB_TEMPLATES_JOBS, job.workflow_job_template_id, job)
+        if job.workflow_job_template_id:
+            self.mark_if_no_key(self.WORKFLOW_JOB_TEMPLATES_JOBS, job.workflow_job_template_id, job)
+        elif job.unified_job_template_id:  # for sliced jobs
+            self.mark_if_no_key(self.WORKFLOW_JOB_TEMPLATES_JOBS, job.unified_job_template_id, job)
 
     def project_update_blocked_by(self, job):
         return self.get_item(self.PROJECT_UPDATES, job.project_id)
--- a/awx/main/scheduler/dependency_graph.py
+++ b/awx/main/scheduler/dependency_graph.py
@@ -96,7 +96,11 @@ class DependencyGraph(object):
 
     def workflow_job_blocked_by(self, job):
         if job.allow_simultaneous is False:
-            return self.get_item(self.WORKFLOW_JOB_TEMPLATES_JOBS, job.workflow_job_template_id)
+            if job.workflow_job_template_id:
+                return self.get_item(self.WORKFLOW_JOB_TEMPLATES_JOBS, job.workflow_job_template_id)
+            elif job.unified_job_template_id:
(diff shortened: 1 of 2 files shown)


awx/main/scheduler/dependency_graph.py
```diff
@@ -85,7 +96,11 @@ def job_blocked_by(self, job):
 
     def workflow_job_blocked_by(self, job):
         if job.allow_simultaneous is False:
-            return self.get_item(self.WORKFLOW_JOB_TEMPLATES_JOBS, job.workflow_job_template_id)
+            if job.workflow_job_template_id:
+                return self.get_item(self.WORKFLOW_JOB_TEMPLATES_JOBS, job.workflow_job_template_id)
+            elif job.unified_job_template_id:
+                # Sliced jobs are WorkflowJob type but do not have a workflow_job_template_id
+                return self.get_item(self.WORKFLOW_JOB_TEMPLATES_JOBS, job.unified_job_template_id)
```
AlanCoding, line 103: Oh wow, good catch here. That makes the exclusivity rules complete, I just assumed it would be harder than this.

Let me just note - we don't have any testing for this, and it is a very notable absence.

awx/main/scheduler/task_manager.py
```diff
@@ -272,34 +276,13 @@ def timeout_approval_node(self):
...
-        workflow_jobs_running.extend(workflow_to_start)
-        return workflow_jobs_running
+        self.all_tasks = [wf for wf in WorkflowJob.objects.filter(status='running')]
 
     @timeit
     def _schedule(self):
-        running_workflow_tasks = self.get_tasks()
-        if len(running_workflow_tasks) > 0:
-            self.spawn_workflow_graph_jobs(running_workflow_tasks)
+        self.get_tasks()
```
AlanCoding: I want to say this without taking a position on it:

It would be ideal that all subclasses of `TaskBase` have the same call pattern for `get_tasks()`. This may be the best, it may not be.

awx/main/scheduler/task_manager.py
```diff
@@ -623,6 +615,13 @@ def process_pending_tasks(self, pending_tasks):
                         tasks_to_update_job_explanation.append(task)
                 continue
 
+            if isinstance(task, WorkflowJob):
+                # Previously we were tracking allow_simultaneous blocking both here and in DependencyGraph.
+                # Double check that using just the DependencyGraph works for Workflows and Sliced Jobs.
+                self.dependency_graph.add_job(task)
```
AlanCoding: Not a critique of your PR, but we should consider moving the call to `self.dependency_graph.add_job` into the `start_task` method, as I cannot find a case where we call `start_task` without adding it to the dependency graph. I am aware that the `start_task` method does a lot of things, and it's extra mental energy to question whether this line in appropriate or not, when it should always be changing from pending to running.

awx/main/scheduler/task_manager.py
```diff
@@ -572,10 +555,19 @@ def start_task(self, task, instance_group, dependent_tasks=None, instance=None):
...
             # at this point we already have control/execution nodes selected for the following cases
-            task.instance_group = instance_group
-            execution_node_msg = f' and execution node {task.execution_node}' if task.execution_node else ''
-            logger.debug(f'Submitting job {task.log_format} controlled by {task.controller_node} to instance group {instance_group.name}{execution_node_msg}.')
+            else:
+                task.instance_group = instance_group
+                execution_node_msg = f' and execution node {task.execution_node}' if task.execution_node else ''
+                logger.debug(
+                    f'Submitting job {task.log_format} controlled by {task.controller_node} to instance group {instance_group.name}{execution_node_msg}.'
+                )
```
AlanCoding, line 571: Just marking for notes - this seems to bring us back to parity with `devel`.

The decisions here have significant system consequences and are non-obvious, so this is best. For example, from the cancel rework stuff, I know that workflows are assigned a `celery_task_id`. Because of that, I have to special-case workflows to not look for the process running under the dispatcher service. Nonetheless, that's accepted canonical behavior, so we shouldn't change it.

awx/main/scheduler/task_manager.py
kdelee: agree

awx/main/scheduler/task_manager.py
kdelee: I think I addressed your concern, checkout latest commit

AlanCoding (approved):
This is great work. Down to the very small detail, I'm in strong agreement here.

fosterseth (approved):

PR #12502, opened 2022-07-11
[not devel] Move approval node expiration logic into queryset

AlanCoding:
##### SUMMARY
Instead of loading all active approval nodes into memory, this only loads in the expired approval nodes from the start.

The obvious benefit is that the task manager does not spend a ton of time looping over approval nodes that have not expired, or that will never expire. Since the goal is to pause workflows, we should expect a large number of pending / running approval node / workflow jobs but the task manager is poorly written to handle this.

This passes the integration test I have on the subject, but may not pass unit tests at the moment.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API


kdelee (approved):

kdelee (approved):
The expires thing does make this way easier to read. I'm for it

fosterseth (approved):

PR #12503, opened 2022-07-11
[task manager refactor] Cache preferred instance groups when creating job

fosterseth:
##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/scheduler/task_manager_models.py
+++ b/awx/main/scheduler/task_manager_models.py
@@ -67,6 +67,7 @@ class TaskManagerInstanceGroups:
     def __init__(self, instances_by_hostname=None, instance_groups=None, instance_groups_queryset=None):
         self.instance_groups = dict()
         self.controlplane_ig = None
+        self.pk_ig_map = dict()
 
         if instance_groups is not None:  # for testing
             self.instance_groups = instance_groups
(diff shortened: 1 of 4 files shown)


awx/main/scheduler/task_manager_models.py
```diff
@@ -121,3 +123,16 @@ def find_largest_idle_instance(self, instance_group_name, capacity_type='executi
                 elif i.capacity > largest_instance.capacity:
                     largest_instance = i
         return largest_instance
+
+    def get_instance_groups_from_task_cache(self, task):
+        igs = []
+        for pk in task.preferred_instance_groups_cache:
```
AlanCoding: Imagine that you have a job from before the migration. In that case `task.preferred_instance_groups_cache` may be `None` because of the default you set. The _last_ thing we want is an uncaught exception here, so I would advocate for further, aggressive, data sanitization. For another pathway, someone could create a job with `awx-manage shell_plus`, and you don't want to grind to a halt because it didn't take the canonical creation path. In all of these cases, there's likely very little care for the specific IG / instance selection. A job went rogue, and the user just wants you to move it along and let the work happen.

AlanCoding (approved):

AlanCoding:
left nit-pick comment, can be handled in followup work.

awx/main/scheduler/task_manager_models.py
fosterseth: good call, made a change to check for task.preferred_instance_groups_cache before looping over it

PR #12504, opened 2022-07-11
Modifying SAML adapter to not auto-add default galaxy creds to orgs on login

john-westcott-iv:
##### SUMMARY
The SAML adapter would attempt to add the default galaxy credential every time a user logged into an organization. With this change the cred will only be added if the organization is created.

##### ISSUE TYPE

 - Bug or Docs Fix

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.2.1.dev62+gedc3a242bd
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/models/organization.py
+++ b/awx/main/models/organization.py
@@ -114,13 +114,6 @@ class Organization(CommonModel, NotificationFieldsModel, ResourceMixin, CustomVi
     def _get_related_jobs(self):
         return UnifiedJob.objects.non_polymorphic().filter(organization=self)
 
-    def create_default_galaxy_credential(self):
-        from awx.main.models import Credential
-
-        public_galaxy_credential = Credential.objects.filter(managed=True, name='Ansible Galaxy').first()
-        if public_galaxy_credential is not None and public_galaxy_credential not in self.galaxy_credentials.all():
-            self.galaxy_credentials.add(public_galaxy_credential)
-
 
 class OrganizationGalaxyCredentialMembership(models.Model):
 
--- a/awx/sso/pipeline.py
+++ b/awx/sso/pipeline.py
@@ -77,6 +77,15 @@ def _update_m2m_from_expression(user, related, expr, remove=True):
         related.remove(user)
 
 
+def get_or_create_with_default_galaxy_cred(**kwargs):
+    from awx.main.models import Organization
+
+    (org, org_created) = Organization.objects.get_or_create(kwargs)
+    if org_created:
+        org.create_default_galaxy_credential()
+    return org
+
+
 def _update_org_from_attr(user, related, attr, remove, remove_admins, remove_auditors, backend):
     from awx.main.models import Organization
     from django.conf import settings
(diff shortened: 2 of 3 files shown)


awx/sso/pipeline.py
```diff
@@ -77,6 +77,15 @@ def _update_m2m_from_expression(user, related, expr, remove=True):
         related.remove(user)
 
 
+def get_or_create_with_default_galaxy_cred(**kwargs):
+    from awx.main.models import Organization
+
+    (org, org_created) = Organization.objects.get_or_create(kwargs)
```
AlanCoding: I mentioned this in our side-bar, but this probably needs to be `**kwargs`, and I believe that could be why unit tests are failing. I'm happy to approve this once unit tests are passing, because this is a solid fix.

awx/main/models/organization.py
```diff
@@ -114,13 +114,6 @@ def get_absolute_url(self, request=None):
     def _get_related_jobs(self):
         return UnifiedJob.objects.non_polymorphic().filter(organization=self)
 
-    def create_default_galaxy_credential(self):
```
john-westcott-iv, line 117: @AlanCoding are you good with moving this function into the pipeline since its only used there?

awx/sso/pipeline.py
```diff
@@ -77,6 +77,21 @@ def _update_m2m_from_expression(user, related, expr, remove=True):
...
+    from awx.main.models import Organization, Credential
+
+    (org, org_created) = Organization.objects.get_or_create(**kwargs)
+    if org_created:
+        logger.debug("Created org {} (id {}) from {}".format(org.name, org.id, kwargs))
+        public_galaxy_credential = Credential.objects.filter(managed=True, name='Ansible Galaxy').first()
+        if public_galaxy_credential is not None and public_galaxy_credential not in org.galaxy_credentials.all():
+            org.galaxy_credentials.add(public_galaxy_credential)
+            logger.debug("Added default Ansible Galaxy credential to org")
+        else:
```
AlanCoding, line 90: You might want to modify the `else:` to `elif public_galaxy_credential is not None:`

Because otherwise, the log isn't very clear. It _could_ find the default Ansible Galaxy credential, but it is already associated with the organization. If this is the case, then I would suggest that no log would be preferable, because no action has been taken, and no anomaly has happened.

john-westcott-iv: I also thought about just taking that check off because we only get in here if we created an org. So this would only be triggered if someone altered the creation of an org to auto-add the galaxy credential.

AlanCoding: Oh I see that now. So you might be able to delete `and public_galaxy_credential not in org.galaxy_credentials.all()`, because you have more information than the old method, which is that it's a new organization, so the galaxy_credential list should necessarily be empty. That would be better performance-wise as well.

awx/sso/tests/functional/test_pipeline.py
```diff
@@ -92,8 +92,13 @@ def test_update_user_orgs(self, org, backend, users, galaxy_credential):
         assert Organization.objects.get(name="Default_Alias") is not None
 
         for o in Organization.objects.all():
-            assert o.galaxy_credentials.count() == 1
-            assert o.galaxy_credentials.first().name == 'Ansible Galaxy'
+            if o.name == 'Default':
+                # The default org was already created and should not have a galaxy credential
+                assert o.galaxy_credentials.count() == 0
```
AlanCoding, line 97: It seems like the old assertion was really asserting on a bug...

AlanCoding (approved):
great work with this

awx/sso/pipeline.py
john-westcott-iv: Removed.

PR #12508, opened 2022-07-12
Optimize object creation by getting fewer empty relationships

AlanCoding:
##### SUMMARY
Making changes to objects unintentionally winds up being a very heavy operation in this app.

The first reason for this is that an activity stream entry has to be created. This requires 1 query at a minimum, but we will probably not bother with optimization if it comes in under around 10 queries.

Unexpectedly, the _main_ performance bottleneck when modifying an object is that we run the `ActivityStreamSerializer` in order to send to the `analytics_logger`, which very few users are going to have turned on in the first place. This winds up being notoriously inefficient, as it maintains a many-to-many relationship with virtually every model type that exists, on the order of ~25. To fully serialize, we make requests to all of those, even though only 3 at most will ever be used.

This manifests in unexpected places, like `create_unified_job`, which has caught my interest because https://github.com/ansible/awx/pull/12503 will add more operations to that method, decreasing the POST response times.

This is an offer to pay down some of our performance debt as a counter-balance to those additions.

##### ISSUE TYPE
 - Bug or Docs Fix

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
I must address the difficult question "is this correct?"

I ran this script over all 3147 activity stream entries that I had:

```python
for o in ActivityStream.objects.all():
  for t in ActivityStreamSerializer(o)._local_summarizable_fk_fields:
    if not hasattr(o, t[0]):
      continue
    rel_items = list(getattr(o, t[0]).all())
    if not rel_items:
      continue
    if t[0] not in (o.object1, o.object2):
      print(f'Found counter example {o}, {t[0]}, {o.object1}, {o.object2}')
```

From that, I was able to find that roles were the only exception to the rule that `object1` or `object2` will correspond to the name of a many-to-many object manager that is populated with something.

Confirm for yourself, that our signals add to the relationship named by the object1 or 2 fields:

https://github.com/ansible/awx/blob/bff49f2a5fd8929808b2890bc2d16279c523cc53/awx/main/signals.py#L450

https://github.com/ansible/awx/blob/bff49f2a5fd8929808b2890bc2d16279c523cc53/awx/main/signals.py#L478

https://github.com/ansible/awx/blob/bff49f2a5fd8929808b2890bc2d16279c523cc53/awx/main/signals.py#L560-L561

https://github.com/ansible/awx/blob/bff49f2a5fd8929808b2890bc2d16279c523cc53/awx/main/signals.py#L574

the one possible exception there is:

https://github.com/ansible/awx/blob/bff49f2a5fd8929808b2890bc2d16279c523cc53/awx/main/signals.py#L571-L572

But if this didn't match the other field name, normal (non-RBAC) entries wouldn't show up right either. Filtering on `/api/v2/activity_stream/?object_relationship_type__endswith=_role`, this does seem to work right, with each entry summarizing 4 items typically.

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -29,7 +29,6 @@ from django.utils.translation import gettext_lazy as _
 from django.utils.encoding import force_str
 from django.utils.text import capfirst
 from django.utils.timezone import now
-from django.utils.functional import cached_property
 
 # Django REST Framework
 from rest_framework.exceptions import ValidationError, PermissionDenied
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -5008,8 +5008,7 @@ class ActivityStreamSerializer(BaseSerializer):
     object_association = serializers.SerializerMethodField(help_text=_("When present, shows the field name of the role or relationship that changed."))
     object_type = serializers.SerializerMethodField(help_text=_("When present, shows the model on which the role or relationship was defined."))
 
-    @cached_property
-    def _local_summarizable_fk_fields(self):
+    def _local_summarizable_fk_fields(self, obj):
         summary_dict = copy.copy(SUMMARIZABLE_FK_FIELDS)
         # Special requests
         summary_dict['group'] = summary_dict['group'] + ('inventory_id',)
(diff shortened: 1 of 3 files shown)


AlanCoding:
For a sense of the difference:

Queries for activity stream global list
Before: 403
After: 55

Rough ballpark for a single serialization for the on_commit hook
Before: 46
After: 6

comparing the create_unified_job method (weirness due to DDT and on_commit)
Before: 24 / 79
After: 24 / 40

kdelee (approved):

fosterseth (approved):

jbradberry (approved):

AlanCoding:
I ran the full tests and they came back looking good. We also now have up https://github.com/ansible/awx/pull/12545, and some others will be coming soon, which are the motivators for this change. So I will go ahead with merging.

PR #12511, opened 2022-07-12
set default dependencies_processed for each task type

fosterseth:
##### SUMMARY

WorkflowJob, SystemJob, AdHocCommand, ProjectUpdate should all start with `dependencies_processed = True`

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/models/jobs.py
+++ b/awx/main/models/jobs.py
@@ -1213,6 +1213,9 @@ class SystemJob(UnifiedJob, SystemJobOptions, JobNotificationMixin):
 
     extra_vars_dict = VarsDictProperty('extra_vars', True)
 
+    def _set_default_dependencies_processed(self):
+        self.dependencies_processed = True
+
     @classmethod
     def _get_parent_field_name(cls):
         return 'system_job_template'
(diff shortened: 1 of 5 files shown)


awx/main/models/jobs.py
```diff
@@ -1213,6 +1213,9 @@ class Meta:
 
     extra_vars_dict = VarsDictProperty('extra_vars', True)
 
+    def _set_default_dependencies_processed(self):
+        self.dependencies_processed = True
```
AlanCoding, line 1217: this can't be right

fosterseth: for SystemJob?

kdelee: think I'm on same page as seth

kdelee (approved):

awx/main/models/jobs.py
AlanCoding: oh sorry, so the `pass` is what gets used as the default for the normal type jobs. In other words, it uses the normal default, because the method doesn't do anything.

AlanCoding (approved):

PR #12522, opened 2022-07-14
Adding remove_superuser and remove_system_auditors to the SAML user attribute map

### LINKED ISSUE #12519: Alter SAML adapter to not remove system admin and system auditor flags

### Feature Summary

In the awx setting SOCIAL_AUTH_SAML_USER_FLAGS_BY_ATTR add a property like `remove_superuser` which would behave like the `remove_admins` in the team mapping; preventing users granted the superuser role through the UI from having that flag removed from their user. A smilier option for `remove_auditors` would also be present. 

- [ ] UI
- [X] API
- [X] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

john-westcott-iv:
##### SUMMARY

In the SAML Adapter user attribute map we are adding to remove properties. These will function like the org/team maps remove function and not auto-remove the system_admin or system_auditor flags. This will also the SMAL adapter to work similar to how it used to before the user attribute map was put in place.

Fixes https://github.com/ansible/awx/issues/12519

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 
##### AWX VERSION

```
awx: 21.2.1.dev70+g2039b20b7b
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/sso/pipeline.py
+++ b/awx/sso/pipeline.py
@@ -245,6 +245,7 @@ def _check_flag(user, flag, attributes, user_flags_settings):
     is_role_key = "is_%s_role" % (flag)
     is_attr_key = "is_%s_attr" % (flag)
     is_value_key = "is_%s_value" % (flag)
+    remove_setting = "remove_%ss" % (flag)
 
     # Check to see if we are respecting a role and, if so, does our user have that role?
     role_setting = user_flags_settings.get(is_role_key, None)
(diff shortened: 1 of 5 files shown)


awx/sso/pipeline.py
```diff
@@ -276,7 +277,7 @@ def _check_flag(user, flag, attributes, user_flags_settings):
             # if they don't match make sure that new_flag is false
             else:
                 logger.debug(
-                    "Refusing %s for %s because attr %s (%s) did not match value '%s'"
+                    "For %s on %s attr %s (%s) did not match expected value '%s'"
```
gundalow, line 280: For flag on user.username attr...

Is `For` the right word here?

john-westcott-iv: I'm opened to suggestions but this would resolve to something like:
`For superuser on awx_unpriv attr username (['awx_unpriv']) did not match expected value 'a'`

I used username as the attribute which is why you get that `['awx_unpriv']`.

AlanCoding (approved):

jay-steurer (approved):

fosterseth (approved):

PR #12533, opened 2022-07-18
Add state fields to Instance and InstanceLink

jbradberry:
##### SUMMARY
In order to support scaling the Receptor mesh from within Collector, we need to track the state.

New states:
- provisioning: marks the new instance as one that needs to have Receptor and ansible-runner and so forth installed
- provision-fail: a provisioning job attempted to run on this node and failed
- installed: this instance already has the packages installed (perhaps because the admin manually installed them), and we are just waiting for the health check to bring it fully into the mesh
- deprovisioning: this instance needs to have the bits uninstalled.  Once that succeeds, the instance can be removed from the database.
- deprovision-fail: a de-provisioning job attempted to run on this node and failed.

The states of 'unavailable' and 'ready' correspond to existing ideas about whether or not an instance has errors.

Additionally, we are adding a `link_state` field to `InstanceLink`.  This allows us to specify peer relationships that should be added or removed, even if the instances on either side are not having any other operations performed on them.

- adding
- established
- removing

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API


jbradberry:
closing in favor of https://github.com/ansible/awx/pull/12587

PR #12534, opened 2022-07-18
[task_manager_refactor] Fix unit / functional tests

fosterseth:
##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.3.1.dev26+g52ca75f9d0
```

### DIFF
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -37,7 +37,6 @@ from awx.main.utils.pglock import advisory_lock
 from awx.main.utils import (
     get_type_for_model,
     ScheduleTaskManager,
-    ScheduleDependencyManager,
     ScheduleWorkflowManager,
 )
 from awx.main.utils.common import create_partition, task_manager_bulk_reschedule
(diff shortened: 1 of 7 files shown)


awx/main/scheduler/task_manager.py
```diff
@@ -432,7 +432,8 @@ def generate_dependencies(self, undeped_tasks):
         return created_dependencies
 
     def process_tasks(self):
-        self.generate_dependencies(self.all_tasks)
+        deps = self.generate_dependencies(self.all_tasks)
+        deps_of_deps = self.generate_dependencies(deps)
```
AlanCoding: so, do you need to process `deps_of_deps` for update-on-launch behavior? Otherwise that's your linter failure

awx/main/tests/functional/task_management/__init__.py
```diff
@@ -0,0 +1,6 @@
+def create_job(jt, dependencies_processed=True):
+    job = jt.create_unified_job()
+    job.status = "pending"
```
AlanCoding, line 3: I think the canonical way to do this would be `signal_start` with some mocks on top. I'm not trying to change anything now, just want to fill in any gaps in understanding.

AlanCoding (approved):

PR #12537, opened 2022-07-18
[task_manager_refactor] move consume_capacity to start_task

fosterseth:
##### SUMMARY

move consume capacity logic to start_task so that it is centralized and not called in various places elsewhere.

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.3.1.dev26+g1b5e3ccec4

```

### DIFF
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -512,6 +512,12 @@ class TaskManager(TaskBase):
             ScheduleTaskManager().schedule()
         from awx.main.tasks.system import handle_work_error, handle_work_success
 
+        # update capacity for control node and execution node
+        if task.controller_node:
+            self.instances[task.controller_node].consume_capacity(settings.AWX_CONTROL_NODE_TASK_IMPACT)
+        if task.execution_node:
+            self.instances[task.execution_node].consume_capacity(task.task_impact)
+
         dependent_tasks = dependent_tasks or []
 
         task_actual = {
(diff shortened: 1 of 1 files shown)


awx/main/scheduler/task_manager.py
```diff
@@ -600,13 +606,9 @@ def process_pending_tasks(self, pending_tasks):
...
             preferred_instance_groups = self.instance_groups.get_instance_groups_from_task_cache(task)
-            # Determine if there is control capacity for the task
-            if task.capacity_type == 'control':
-                control_impact = task.task_impact + settings.AWX_CONTROL_NODE_TASK_IMPACT
-            else:
-                control_impact = settings.AWX_CONTROL_NODE_TASK_IMPACT
+
             control_instance = self.instance_groups.fit_task_to_most_remaining_capacity_instance(
-                task, instance_group_name=settings.DEFAULT_CONTROL_PLANE_QUEUE_NAME, impact=control_impact, capacity_type='control'
+                task, instance_group_name=settings.DEFAULT_CONTROL_PLANE_QUEUE_NAME, impact=settings.AWX_CONTROL_NODE_TASK_IMPACT, capacity_type='control'
```
AlanCoding: I could be wrong, but I think you should back out changes to these lines, but leave everything else. Without this, control actions (like project updates) can put a node over capacity.

fosterseth: no I think you're right -- for capacity type "control", we should be fitting with the task_impact + control_node_task_impact

PR #12545, opened 2022-07-19
Cache task_impact

fosterseth:
##### SUMMARY

task_impact is now a field on the database
It is calculated and set during `create_unified_job`

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.3.1.dev28+gcde87c065b

```

### DIFF
--- a/awx/main/migrations/0165_task_manager_refactor.py
+++ b/awx/main/migrations/0165_task_manager_refactor.py
@@ -27,4 +27,9 @@ class Migration(migrations.Migration):
                 blank=True, default=None, editable=False, help_text='A cached list with pk values from preferred instance groups.', null=True
             ),
         ),
+        migrations.AddField(
+            model_name='unifiedjob',
+            name='task_impact',
+            field=models.PositiveIntegerField(default=0, editable=False),
+        ),
     ]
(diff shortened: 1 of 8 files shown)


AlanCoding:
It would be good to go ahead and lock-in some performance enhancements in `awx/main/models/ha.py`, for example. Also, the serializers (like instance serializer) do some gross stuff that can go away fairly easily now that this is in place. But this will still be a substantial speedup without that and if you want to move ahead as it is, I support that.

AlanCoding (approved):

rebeccahhh (approved):
I particularly like the `or 0` in there

AlanCoding (approved):

sdktr:
Is this property a tradeoff between Compute resources (calculate based on other (cached?) properties vs. fetch from database (and calculate when changing)?

If so: is the tradeoff worthwhile? Sacrifice a db lookup for less compute time?

AlanCoding:
Yes, it is worth it. This property is never changed, so it is computed once on creation. Adding a db lookup on creation avoids a db lookup by the task manager when considering a pending job. The task manager considers all pending jobs on every run, and as pending jobs pile up, the system fails to keep up, which becomes an outright bug where it seizes up and is unable to process any further work.

PR #12546, opened 2022-07-19
Fix bug where node alias is not remaining after changing the template on a wf node

mabashian:
##### SUMMARY
Changing the underlying unified job template on a node would wipe the alias out.  I fixed this along with another bug where clearing out the alias `''` would result in the old alias still being displayed on the node (rather than the UJT name).

##### ISSUE TYPE
 - Bug or Docs Fix

##### COMPONENT NAME
 - UI


akus062381 (approved):
QE tested and approved

AlexSCorey (approved):
II tested it out.  Looks good to me

Issue #12512, opened 2022-07-12
Update Approval node toolbar buttons


AlexSCorey:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature Summary

The Approve, Deny, Cancel workflow for the Workflow Approval List is not super clear.  We should abandon the dropdown button here and display Approve, Deny, Cancel buttons is the selected items are still pending action. If the selected items are not pending action we should display the Delete Button.

### Select the relevant components

- [X] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

PR #12548, opened 2022-07-20
Refactors and redesigns workflow approval to impove UX

AlexSCorey:
##### SUMMARY
Addresses #12512.  Moves approve, deny, cancel buttons to be in the row items instead of in the toolbar.  This puts us more inline with UXD recommendations

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - UI


mabashian:
@AlexSCorey I noticed that after I approved a workflow the approve and cancel buttons got greyed out but the deny button was still enabled.  It's showing the correct tooltip.

Screen Shot 2022-07-29 at 3 03 02 PM

The same thing happens in reverse if I deny an approval, the approve button remains enabled but with a tooltip saying that it's already been denied.

mabashian:
@AlexSCorey I think there's something funny going on with cancelling an approval:

workflow_approval_cancel

The same thing happens when trying to delete from the workflow approval detail.

kialam:
I don't see the wf approval details page refresh after clicking the "Cancel" button. This could be a local issue on my end. @akus062381 @InnocentK @jay-steurer are you able to reproduce the issue as well?
nipkwE94Hk

kialam:
This compiler warning should be fixed before merging:

```
src/components/JobCancelButton/JobCancelButton.js
  Line 32:8:  React Hook useCallback has a missing dependency: 'onCancelWorkflow'. Either include it or remove the dependency array. If 'onCancelWorkflow' changes too often, find the parent component that defines it and wrap that definition in useCallback  react-hooks/exhaustive-deps
```

kialam (approved):
Good to merge once warnings are fixed.

PR #12550, opened 2022-07-20
enable feature branch api-schema test

TheRealHaoLiu:
##### SUMMARY
when submitting a PR against a feature branch of AWX the api-schema check in CI workflow will always fail due to missing schema file on the s3 bucket for the feature branch

this PR adds the following to the github action
- enable schema upload to s3 bucket for feature branch
- add workflow to delete schema from s3 bucket when feature branch is deleted

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - Other

##### AWX VERSION

```
awx: 21.3.1.dev30+g6db693ad62
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/.github/workflows/feature_branch_deletion.yml
+++ b/.github/workflows/feature_branch_deletion.yml
@@ -1,0 +1,24 @@
+---
+name: Feature branch deletion cleanup
+on:
+  delete:
+    branches:
+      - feature_**
+jobs:
+  push:
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+      contents: read
+    steps:
+      - name: Delete API Schema
+        env:
+          AWS_ACCESS_KEY: ${{ secrets.AWS_ACCESS_KEY }}
+          AWS_SECRET_KEY: ${{ secrets.AWS_SECRET_KEY }}
+          AWS_REGION: 'us-east-1'
+        run: |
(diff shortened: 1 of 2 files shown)


john-westcott-iv (approved):

AlanCoding (approved):

.github/workflows/feature_branch_deletion.yml
```diff
@@ -0,0 +1,24 @@
+---
+name: Feature branch deletion cleanup
+on:
+  delete:
+    branches:
+      - feature_**
```
AlanCoding, line 6: Elsewhere, you named your branch `feature-project-integrity`, but this syntax implies a canonical naming of feature branches with `feature_*`, which creates a conflict of dashes (-) versus underscores (_), and you should declare a winner.

TheRealHaoLiu: `>.<` lol

AlanCoding:
This still looks fine to me. If we get it in soon, it's sure to be tested in the next round of feature development.

TheRealHaoLiu:
Need to change to - don't merge

TheRealHaoLiu:
actually its ok to do _ i guess to follow the release branch schema 
the reason why i wanted `-` is bc i can create automation that deploy the feature branch in to a namespace in kube cluster but `_` is not allowed in the namespace name

so the automation would just have to strip the _ is and replace with - is all

PR #12552, opened 2022-07-20
Implement Generic OIDC Provider

### LINKED ISSUE #1967: [RFE] Add OIDC generic integration to AWX
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
 - API
 - UI

##### SUMMARY
As social auth2 for Google or Github have an integration method on WebUI, we need another for OIDC connect to provide a generic solution for auth/authz into AWX.

whitej6:
##### SUMMARY

Updates social-auth-core to 4.3.0 and implemented a generic OIDC SSO provider. #1967 

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 - UI
 - SSO
##### AWX VERSION

```
awx: 9.0.2.dev8705+gb26a515497
```

##### ADDITIONAL INFORMATION

Tested with a generic development provider. Only implemented with minimal set of variables required by social-auth-core.

```

```

### DIFF
--- a/awx/sso/conf.py
+++ b/awx/sso/conf.py
@@ -1216,6 +1216,54 @@ register(
 )
 
 ###############################################################################
+# Generic OIDC AUTHENTICATION SETTINGS
+###############################################################################
+
+register(
+    'SOCIAL_AUTH_OIDC_KEY',
+    field_class=fields.CharField,
+    allow_null=False,
+    default=None,
+    label=_('OIDC Key'),
+    help_text='The OIDC key (Client ID) from your IDP.',
+    category=_('Generic OIDC'),
+    category_slug='oidc',
+)
+
+register(
+    'SOCIAL_AUTH_OIDC_SECRET',
+    field_class=fields.CharField,
+    allow_blank=True,
+    default='',
+    label=_('OIDC Secret'),
+    help_text=_('The OIDC secret (Client Secret) from your IDP.'),
+    category=_('Generic OIDC'),
+    category_slug='oidc',
+    encrypted=True,
+)
+
+register(
+    'SOCIAL_AUTH_OIDC_OIDC_ENDPOINT',
+    field_class=fields.CharField,
+    allow_blank=True,
+    default='',
+    label=_('OIDC Provider URL'),
+    help_text=_('The URL for your OIDC provider including the path up to /.well-known/openid-configuration'),
+    category=_('Generic OIDC'),
+    category_slug='oidc',
+)
+
+register(
+    'SOCIAL_AUTH_OIDC_VERIFY_SSL',
+    field_class=fields.BooleanField,
+    default=True,
+    label=_('Verify OIDC Provider Certificate'),
+    help_text=_('Verify the OIDV provider ssl certificate.'),
+    category=_('Generic OIDC'),
+    category_slug='oidc',
+)
+
+###############################################################################
 # SAML AUTHENTICATION SETTINGS
 ###############################################################################
 
--- a/awx/ui/src/screens/Setting/OIDC/OIDCEdit/OIDCEdit.js
+++ b/awx/ui/src/screens/Setting/OIDC/OIDCEdit/OIDCEdit.js
@@ -1,0 +1,147 @@
+import React, { useCallback, useEffect } from 'react';
+import { useHistory } from 'react-router-dom';
+import { Formik } from 'formik';
+import { Form } from '@patternfly/react-core';
+import { CardBody } from 'components/Card';
+import ContentError from 'components/ContentError';
+import ContentLoading from 'components/ContentLoading';
+import { FormSubmitError } from 'components/FormField';
+import { FormColumnLayout } from 'components/FormLayout';
+import { useSettings } from 'contexts/Settings';
+import useModal from 'hooks/useModal';
+import useRequest from 'hooks/useRequest';
+import { SettingsAPI } from 'api';
+import { RevertAllAlert, RevertFormActionGroup } from '../../shared';
+import {
+  EncryptedField,
+  InputField,
+  BooleanField,
+} from '../../shared/SharedFields';
+
+function OIDCEdit() {
+  const history = useHistory();
+  const { isModalOpen, toggleModal, closeModal } = useModal();
+  const { PUT: options } = useSettings();
+
+  const {
+    isLoading,
+    error,
+    request: fetchOIDC,
+    result: OIDC,
+  } = useRequest(
+    useCallback(async () => {
+      const { data } = await SettingsAPI.readCategory('oidc');
+      const mergedData = {};
+      Object.keys(data).forEach((key) => {
+        if (!options[key]) {
+          return;
+        }
+        mergedData[key] = options[key];
+        mergedData[key].value = data[key];
+      });
+      return mergedData;
+    }, [options]),
+    null
+  );
+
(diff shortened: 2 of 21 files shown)


jmcgill298:
This also fixes an issue with using a load-balancer with backend servers using different ports

https://github.com/python-social-auth/social-core/issues/609

whitej6:
@shanemcd I resolved linting and licensing errors in tests, schema I test looks to be pulling from S3 so I am guessing that is expected to fail since I am adding new endpoints and appears to be a stead state schema export. 

The only test I was unable to resolve was screen tests. My background is python and some hints if possible would be greatly appreciated as I am happy to have all pipelines passing before thorough review. I copied the same patterns from AzureAD but still get failures an am unsure why.

whitej6:
@shanemcd was able to fix my issues with screen tests. Now all tests that should pass do pass and is ready for review. Only test expected to fail is API schema due to registering a new endpoint to settings.

awx/ui/src/screens/Setting/OIDC/OIDCEdit/OIDCEdit.js
```diff
@@ -0,0 +1,142 @@
...
+              <FormColumnLayout>
+                <InputField
+                  name="SOCIAL_AUTH_OIDC_KEY"
+                  config={OIDC.SOCIAL_AUTH_OIDC_KEY}
+                />
+                <EncryptedField
+                  name="SOCIAL_AUTH_OIDC_SECRET"
+                  config={OIDC.SOCIAL_AUTH_OIDC_SECRET}
+                />
+                <ObjectField
```
john-westcott-iv: Why an object field here instead of just an InputField? And, actually, I think we have a URL validator field. @keithjgrant may be able to point you towards that.

awx/sso/conf.py
```diff
@@ -1215,6 +1215,44 @@ def _register_ldap(append=None):
...
+    encrypted=True,
+)
+
+register(
+    'SOCIAL_AUTH_OIDC_OIDC_ENDPOINT',
+    field_class=fields.CharField,
+    allow_blank=True,
+    default='',
+    label=_('OIDC Provider URL'),
+    help_text=_('The URL for your OIDC provider, e.g.: http(s)://hostname/.'),
```
john-westcott-iv: Can we change this help_test to be something like `The URL for your OIDC provider including the path up to /.well-known/openid-configuration`

awx/sso/conf.py
```diff
@@ -1215,6 +1215,44 @@ def _register_ldap(append=None):
...
+    category=_('Generic OIDC'),
+    category_slug='oidc',
+)
+
```
john-westcott-iv, line 1265: There is one other setting we may want to think about adding here:
```
register(
    'SOCIAL_AUTH_OIDC_VERIFY_SSL',
    field_class=fields.BooleanField,
    default=True,
    label=_('Verify OIDC Provider Certificate'),
    help_text=_('Verify the OIDV provider ssl certificate.'),
    category=_('Generic OIDC'),
    category_slug='oidc',
)
```
I'm a little torn because we don't do this for other providers but we also occasionally get questions about how to connect to a provider with an untrusted cert. This field may be a quick and easy way for an admin to do some testing. @AlanCoding any thoughts on adding this field?

john-westcott-iv (commented):
Thanks again for the submission! Could you rebase this change on the latest devel branch along with some minor teaks mentioned below?

awx/ui/src/screens/Setting/OIDC/OIDCEdit/OIDCEdit.js
whitej6: Likely a copy pasta oversight, will move to correct field type.

awx/sso/conf.py
whitej6:
```suggestion
    help_text=_('The URL for your OIDC provider including the path up to /.well-known/openid-configuration'),
```

awx/sso/conf.py
whitej6: To play devil's advocate, OIDC _should_ always be SSL and you should _never_ do an implicit trusting of any cert but instead you should install the cert so you can do an explicit trusting of said cert.

Happy to add it or happy to leave as is, will wait on AlanCoding's response.

awx/ui/src/screens/Setting/OIDC/OIDCEdit/OIDCEdit.js
whitej6: Reading through SharedField I believe I just needed to change to InputField and `type="url"`

whitej6:
@john-westcott-iv I fixed the two issues, pending to head if ssl verify is needed. Also rebased, got a little wonky in the process but I think everything is good. Let me know if there is anything else

awx/ui/src/screens/Setting/OIDC/OIDCEdit/OIDCEdit.js
john-westcott-iv: This is all set.

awx/sso/conf.py
john-westcott-iv: I 100% agree that OIDC should always be SSL and trusted in production. But people also use AWX for testing/development and personal use. We also allow credentials and other items to ignore SSL verification. And the stack trace when SSL is not trusted is pretty massive. I talked with an architect this morning and they are good with us adding this. If we wanted to we could make this just settable via the API and not have a UI field on the form.

awx/ui/src/screens/Setting/OIDC/OIDCEdit/OIDCEdit.js
```diff
@@ -0,0 +1,146 @@
...
+import useRequest from 'hooks/useRequest';
+import { SettingsAPI } from 'api';
+import { RevertAllAlert, RevertFormActionGroup } from '../../shared';
+import {
+  EncryptedField,
+  InputField,
+<<<<<<< HEAD
+=======
+  ObjectField,
+>>>>>>> 3505092cf8d127bc75dfe83d39f196a889478890
```
john-westcott-iv: Looks like there was an unresolved merge issue here. I think these lines can just be removed.

john-westcott-iv (changes requested):

awx/ui/src/screens/Setting/OIDC/OIDCEdit/OIDCEdit.js
whitej6: Sorry about that. Fixed.

awx/sso/conf.py
john-westcott-iv: Let's ignore this for now in this PR and I can add this as an option later as needed.

whitej6: @john-westcott-iv I just added it..... 😆 Want me to revert?

john-westcott-iv: If you got it in there, its totally fine; thanks!
We'll get this over to our QE team and hopefully we can merge soon.

john-westcott-iv (approved):

jay-steurer:
Hi @whitej6. Could you rebase your PR? Therefore I can properly check it. Thank you!

whitej6:
@jay-steurer rebased again 😃

jay-steurer:
Verified by QE.

PR #12554, opened 2022-07-20
bulk reschedule context should track each of the 3 managers

fosterseth:
##### SUMMARY

I realized my original try at fixing up the bulk reschedule is incorrect. Right now the context only tracks calls to a single scheduler, but in the task manager, we may be calling _other_ schedulers. For example, we call ScheduleWorkflowManager from within the TaskManager. Therefore, the bulk reschedule context in the TaskManager should be sensitive to each of the 3 types of Schedule managers.

Also, we move the bulk reschedule to _outside_ of the atomic lock, that way if a reschedule happens, that new dispatched manager will be able to grab the lock.

 

##### ISSUE TYPE

 - Bug or Docs Fix

##### COMPONENT NAME

 - API

##### ADDITIONAL INFORMATION

demonstration of the bulk rescheduler tracking each of the 3 managers appropriately

See that TaskManager and DependencyManager are going to be rescheduled, but WorkflowManager is not.

```
In [17]: def print_managers():
    ...:     for m in [ScheduleTaskManager(), ScheduleDependencyManager(), ScheduleWorkflowManager()]:
    ...:         print(m.__class__, m.manager_threading_local.__dict__)
    ...:     print()
    ...: 

In [18]: with task_manager_bulk_reschedule():
    ...:     print_managers()
    ...:     ScheduleTaskManager().schedule()
    ...:     print_managers()
    ...:     ScheduleDependencyManager().schedule()
    ...:     print_managers()
    ...: 
<class 'awx.main.utils.common.ScheduleTaskManager'> {'bulk_reschedule': True, 'needs_scheduling': False}
<class 'awx.main.utils.common.ScheduleDependencyManager'> {'bulk_reschedule': True, 'needs_scheduling': False}
<class 'awx.main.utils.common.ScheduleWorkflowManager'> {'bulk_reschedule': True, 'needs_scheduling': False}

<class 'awx.main.utils.common.ScheduleTaskManager'> {'bulk_reschedule': True, 'needs_scheduling': True}
<class 'awx.main.utils.common.ScheduleDependencyManager'> {'bulk_reschedule': True, 'needs_scheduling': False}
<class 'awx.main.utils.common.ScheduleWorkflowManager'> {'bulk_reschedule': True, 'needs_scheduling': False}

<class 'awx.main.utils.common.ScheduleTaskManager'> {'bulk_reschedule': True, 'needs_scheduling': True}
<class 'awx.main.utils.common.ScheduleDependencyManager'> {'bulk_reschedule': True, 'needs_scheduling': True}
<class 'awx.main.utils.common.ScheduleWorkflowManager'> {'bulk_reschedule': True, 'needs_scheduling': False}
```

### DIFF
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -40,7 +40,7 @@ from awx.main.utils import (
     ScheduleDependencyManager,
     ScheduleWorkflowManager,
 )
-from awx.main.utils.common import create_partition
+from awx.main.utils.common import create_partition, task_manager_bulk_reschedule
 from awx.main.signals import disable_activity_stream
 from awx.main.constants import ACTIVE_STATES
 from awx.main.scheduler.dependency_graph import DependencyGraph
@@ -121,23 +121,22 @@ class TaskBase:
 
     def schedule(self):
         # Lock
-        with advisory_lock(f"{self.prefix}_lock", wait=False) as acquired:
-            with transaction.atomic():
-                if acquired is False:
-                    logger.debug(f"Not running {self.prefix} scheduler, another task holds lock")
-                    return
-                logger.debug(f"Starting {self.prefix} Scheduler")
-                with self.schedule_manager.task_manager_bulk_reschedule():
+        with task_manager_bulk_reschedule():
+            with advisory_lock(f"{self.prefix}_lock", wait=False) as acquired:
+                with transaction.atomic():
+                    if acquired is False:
+                        logger.debug(f"Not running {self.prefix} scheduler, another task holds lock")
+                        return
+                    logger.debug(f"Starting {self.prefix} Scheduler")
                     # if sigterm due to timeout, still record metrics
                     signal.signal(signal.SIGTERM, self.record_aggregate_metrics_and_exit)
                     self._schedule()
                     self.record_aggregate_metrics()
-                logger.debug(f"Finishing {self.prefix} Scheduler")
+                    logger.debug(f"Finishing {self.prefix} Scheduler")
 
 
 class WorkflowManager(TaskBase):
     def __init__(self):
-        self.schedule_manager = ScheduleWorkflowManager()
         super().__init__(prefix="workflow_manager")
 
     @timeit
@@ -146,7 +145,7 @@ class WorkflowManager(TaskBase):
         for workflow_job in self.all_tasks:
             if self.timed_out():
                 logger.warning("Workflow manager has reached time out while processing running workflows, exiting loop early")
-                self.schedule_manager.schedule()
+                ScheduleWorkflowManager().schedule()
                 # Do not process any more workflow jobs. Stop here.
                 # Maybe we should schedule another WorkflowManager run
                 break
@@ -263,7 +262,6 @@ class WorkflowManager(TaskBase):
 
 class DependencyManager(TaskBase):
     def __init__(self):
-        self.schedule_manager = ScheduleDependencyManager()
         super().__init__(prefix="dependency_manager")
 
     def create_project_update(self, task, project_id=None):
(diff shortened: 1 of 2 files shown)


awx/main/scheduler/task_manager.py
```diff
@@ -146,7 +145,7 @@ def spawn_workflow_graph_jobs(self):
         for workflow_job in self.all_tasks:
             if self.timed_out():
                 logger.warning("Workflow manager has reached time out while processing running workflows, exiting loop early")
-                self.schedule_manager.schedule()
+                WorkflowManager().schedule()
```
AlanCoding: elsewhere, in this same method you do `ScheduleWorkflowManager().schedule()`, and that looks like the correct pick to me.

awx/main/utils/common.py
```diff
@@ -851,6 +851,25 @@ def get_mem_effective_capacity(mem_bytes):
 _workflow_manager = threading.local()
 
 
+@contextlib.contextmanager
+def task_manager_bulk_reschedule():
+    managers = [ScheduleTaskManager(), ScheduleWorkflowManager(), ScheduleDependencyManager()]
+    """Context manager to avoid submitting task multiple times."""
+    try:
+        for m in managers:
+            previous_flag = getattr(m.manager_threading_local, 'bulk_reschedule', False)
```
AlanCoding, line 860: When you reference `m.manager_threading_local`, that looks like it points to the global object `_task_manager`, because in the subclasses, like `ScheduleTaskManager`, that's what it passes in. So to rephrase this line, it would set `_task_manager.bulk_reschedule` to `True`, which would sacrifice the object-specific variable that I think you want.

You could unit test this and find out if I'm right about that or not.

`previous_flag` and `previous_value` have some scope confusion too. If you want, you could keep the `bulk_reschedule` on a global `threading.local()` object (that is, `_task_manager`), but importantly, `needs_scheduling` would need to be specific to the manager. There's considerable thought involved to get that right, and the core fix should be much more simply, which I'll point out in a subsequent suggestion.

awx/main/utils/common.py
```diff
@@ -851,6 +851,25 @@ def get_mem_effective_capacity(mem_bytes):
...
+            previous_value = getattr(m.manager_threading_local, 'needs_scheduling', False)
+            m.manager_threading_local.bulk_reschedule = True
+            m.manager_threading_local.needs_scheduling = False
+        yield
+    finally:
+        for m in managers:
+            m.manager_threading_local.bulk_reschedule = previous_flag
+            if m.manager_threading_local.needs_scheduling:
+                m.schedule()
+            m.manager_threading_local.needs_scheduling = previous_value
```
AlanCoding: The scope of `previous_value` and `previous_flag` doesn't make any sense to me.

```suggestion
            m.previous_flag = getattr(m.manager_threading_local, 'bulk_reschedule', False)
            m.previous_value = getattr(m.manager_threading_local, 'needs_scheduling', False)
            m.manager_threading_local.bulk_reschedule = True
            m.manager_threading_local.needs_scheduling = False
        yield
    finally:
        for m in managers:
            m.manager_threading_local.bulk_reschedule = m.previous_flag
            if m.manager_threading_local.needs_scheduling:
                m.schedule()
            m.manager_threading_local.needs_scheduling = m.previous_value
            del m.previous_value
            del m.previous_flag
```

AlanCoding: This would be a low-consequence bug because we don't ever nest use of these anyway, but for logical consistency this is what makes sense to me.

awx/main/scheduler/task_manager.py
fosterseth: yes I meant ScheduleWorkflowManager, thanks for catching that

awx/main/utils/common.py
fosterseth: yep I see it, fixed

AlanCoding (approved):

PR #12559, opened 2022-07-21
[task_manager_refactor] ensure system can complete work without periodic scheduler

fosterseth:
##### SUMMARY

I disabled the periodic schedulers listed in defaults.py and launched a job. The system _should_
 be able to run the job (and its dependencies) to completion. Of course we'll keep the periodic schedulers in, but it is nice to have the system be responsive on its own and progress work forward.

This commit adds Schedule*Manager.schedule() calls in appropriate places to ensure the system can keep propelling forward. 

Tested with Job and dependencies project update A,  inventory update,  project update B
Tested with workflows

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.3.1.dev34+g92739e2d9e

```

### DIFF
--- a/awx/settings/defaults.py
+++ b/awx/settings/defaults.py
@@ -449,7 +449,6 @@ CELERYBEAT_SCHEDULE = {
     'gather_analytics': {'task': 'awx.main.tasks.system.gather_analytics', 'schedule': timedelta(minutes=5)},
     'task_manager': {'task': 'awx.main.scheduler.tasks.task_manager', 'schedule': timedelta(seconds=20), 'options': {'expires': 20}},
     'dependency_manager': {'task': 'awx.main.scheduler.tasks.dependency_manager', 'schedule': timedelta(seconds=20), 'options': {'expires': 20}},
-    'workflow_manager': {'task': 'awx.main.scheduler.tasks.workflow_manager', 'schedule': timedelta(seconds=20), 'options': {'expires': 20}},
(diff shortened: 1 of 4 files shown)


awx/settings/defaults.py
```diff
@@ -449,7 +449,6 @@ def IS_TESTING(argv=None):
     'gather_analytics': {'task': 'awx.main.tasks.system.gather_analytics', 'schedule': timedelta(minutes=5)},
     'task_manager': {'task': 'awx.main.scheduler.tasks.task_manager', 'schedule': timedelta(seconds=20), 'options': {'expires': 20}},
     'dependency_manager': {'task': 'awx.main.scheduler.tasks.dependency_manager', 'schedule': timedelta(seconds=20), 'options': {'expires': 20}},
-    'workflow_manager': {'task': 'awx.main.scheduler.tasks.workflow_manager', 'schedule': timedelta(seconds=20), 'options': {'expires': 20}},
```
fosterseth, line 452: I think we can get rid of this periodic scheduler - the task manager will schedule a workflow manager to run if it sees a running workflow job

AlanCoding: Yes, I was going to suggest doing a version of exactly this, but you did it even better than what I was thinking. Since the queryset of WorkflowManager is exactly the running workflow jobs, we're pretty safely covered.

AlanCoding (approved):

PR #12563, opened 2022-07-23
Enable django debug toolbar by default in development environment

shanemcd:
This PR: 

- Enables Django Debug Toolbar by default in the development environment. 
  - Is there any reason not to do this?
- Uses an Ansible [vars plugin](https://docs.ansible.com/ansible/latest/plugins/vars.html) to automatically set the `INTERNAL_IPS` setting (used by DDT)

### DIFF
--- a/tools/docker-compose/ansible/roles/sources/tasks/main.yml
+++ b/tools/docker-compose/ansible/roles/sources/tasks/main.yml
@@ -56,9 +56,9 @@
     path: "{{ playbook_dir }}/../../../awx/settings/local_settings.py"
     state: absent
 
-- name: Copy local_settings.py
-  copy:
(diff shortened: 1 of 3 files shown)


AlanCoding:
Personally, I would turn this off right away. The performance hit is truly _profound_, and I'm not on particularly bad hardware.

The toolbar is inconsistently usable in the actual UI. It only makes sense to me to use it via the API browser. If we could turn it off for UI requests, that wouldn't be so bad, but without that this makes normal testing much more of a chore, and I wouldn't recommend it.

This could still be useful to insert a commented-out `INTERNAL_IPS` which includes the correct host name. Myself, I've never had trouble with the default hard-coded values, but that may not work on different systems.

Issue #12483, opened 2022-07-07
[UI] Allow user to configure content verification when creating and editing projects


djyasin:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature Summary

As a user I want to be able to configure the content verification when creating and editing projects in the UI.

### Select the relevant components

- [X] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

djyasin:
No longer relevant

PR #12566, opened 2022-07-25
12483 configure content verification

vedaperi:
##### SUMMARY
Added the "Content Signature Validation Credential" field to the Project Form page in the Git subform as a credential lookup field. The field only shows users the applicable credentials by filtering by GPG Public Key. Once the form is saved, the selection is reflected in the Project Details view.
Addresses #12483 

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
There is a bug in the UI that requires the Source Control Credential to be interacted with in order to submit the form with a Content Signature Verification Credential selected. For instance, if a user wants to submit the form with a Content Signature Verification Credential but not a Source Control Credential, they will still need to open the Source Control Credential lookup and explicitly empty the field before being able to submit. This bug is being tracked in #12589.

### DIFF
--- a/awx/ui/src/components/PromptDetail/PromptProjectDetail.js
+++ b/awx/ui/src/components/PromptDetail/PromptProjectDetail.js
@@ -125,6 +125,21 @@ function PromptProjectDetail({ resource }) {
           }
         />
       )}
+      {summary_fields?.signature_validation_credential?.id && (
+        <Detail
+          label={t`Content Signature Validation Credential`}
+          dataCy={`${prefixCy}-content-signature-validation-credential`}
+          value={
+            <CredentialChip
+              key={resource.summary_fields.signature_validation_credential.id}
+              credential={
+                resource.summary_fields.signature_validation_credential
+              }
+              isReadOnly
+            />
+          }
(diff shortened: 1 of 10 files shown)


awx/main/models/projects.py
```diff
@@ -284,6 +284,21 @@ class Meta:
         help_text=_('Allow changing the SCM branch or revision in a job template ' 'that uses this project.'),
     )
 
+    # indicate if content signature validation is enabled
+    signature_validation = models.BooleanField(
+        default=False,
+        help_text=_('Enable signature validation.'),
+    )
```
AlanCoding: I feel obligated to ask the question - why should `signature_validation` be independently configurable from `signature_validation_credential`? If a signature validation credential is associated, wouldn't you _always_ want to enable signature validation? Should the need to disable the validation arise, it would probably need to happen specific to certain playbooks, which is better associated with the template, rather than the project.

TheRealHaoLiu: good point, im thinking over that as well...

PR #12568, opened 2022-07-25
Fix awx collection project module

TheRealHaoLiu:
##### SUMMARY
- added input arg for `signature_validation`
- added input arg for `signature_validation_credential`

##### ISSUE TYPE
 - Bug or Docs Fix

##### COMPONENT NAME
 - Collection

##### AWX VERSION
```
awx: 21.3.1.dev41+g44e64d9401
```

##### ADDITIONAL INFORMATION
created a playbook to tested the following scenario 
- adding project without `signature_validation` enabled (validate project is created as expected)
- added project with `signature_validation` enabled without `signature_validation_credential `(validated project creation failed due to missing credential)
- added project with both `signature_validation` and `signature_validation_credential` (validated project created successfully with both value set correctly)

### DIFF
--- a/awx_collection/plugins/modules/project.py
+++ b/awx_collection/plugins/modules/project.py
@@ -169,6 +169,18 @@ options:
       required: False
       default: 2
       type: float
+    signature_validation:
+      description:
+        - Whether to validate the signature during project update.
+      required: False
+      default: False
+      type: bool
+    signature_validation_credential:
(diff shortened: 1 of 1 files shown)


djyasin (approved):
I can confirm that Hao and I tested this together and it does seem to be behaving properly to correct the AWX Collection issue.

PR #12569, opened 2022-07-25
[FieldLookupBackend] limit iexact to string fields

### LINKED ISSUE #9222: Passing a bad lookup type modifier in api request triggers 500 error

relrod:
Change:
- Case-insensitive search only makes sense on strings, so check the
  type of the field we are searching and ensure it is a string field
  (TextField, CharField, or some subclass thereof).

- This prevents a 500 error when a user uses iexact on, e.g., an
  integer field. Now, a 400 Bad Request is returned instead.

Test Plan:
- Added simple unit tests for iexact

Tickets:
- Fixes #9222

### DIFF
--- a/awx/api/filters.py
+++ b/awx/api/filters.py
@@ -232,6 +232,9 @@ class FieldLookupBackend(BaseFilterBackend):
                 re.compile(value)
             except re.error as e:
                 raise ValueError(e.args[0])
+        elif new_lookup.endswith('__iexact'):
+            if not isinstance(field, (CharField, TextField)):
+                raise ValueError(f'{field.name} is not a text field and cannot be filtered by case-insensitive search')
         elif new_lookup.endswith('__search'):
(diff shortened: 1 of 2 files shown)


AlanCoding (approved):
Great unit test! :100:

fosterseth (approved):

PR #12573, opened 2022-07-26
Submit job to dispatcher as part of transaction

AlanCoding:
##### SUMMARY
This is incomplete, because it creates a UX problem with the websockets. This was expected.

Screenshot from 2022-07-26 14-19-40

The UI status obtains the "waiting" status while the job is supposed to be shown as "running".

The crux of this PR is that we really _need_ to make this change because a very large fraction of the time the task manager runs is spent in the on_commit method. The two unresolved issues are:
 - [x] the "waiting" websocket status message, leading to the bug in the screenshot
 - [ ] the need to do git webhook requests... which I _think_ can be moved in the same way `create_partition` is here.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -11,7 +11,7 @@ import sys
 import signal
 
 # Django
-from django.db import transaction, connection
+from django.db import transaction
 from django.utils.translation import gettext_lazy as _, gettext_noop
 from django.utils.timezone import now as tz_now
 from django.conf import settings
@@ -39,7 +39,7 @@ from awx.main.utils import (
     ScheduleTaskManager,
     ScheduleWorkflowManager,
 )
-from awx.main.utils.common import create_partition, task_manager_bulk_reschedule
+from awx.main.utils.common import task_manager_bulk_reschedule
 from awx.main.signals import disable_activity_stream
 from awx.main.constants import ACTIVE_STATES
 from awx.main.scheduler.dependency_graph import DependencyGraph
@@ -556,22 +556,23 @@ class TaskManager(TaskBase):
                 task.save()
                 task.log_lifecycle("waiting")
 
-        def post_commit():
-            if task.status != 'failed' and type(task) is not WorkflowJob:
-                # Before task is dispatched, ensure that job_event partitions exist
-                create_partition(task.event_class._meta.db_table, start=task.created)
-                task_cls = task._get_task_class()
-                task_cls.apply_async(
-                    [task.pk],
-                    opts,
-                    queue=task.get_queue_name(),
-                    uuid=task.celery_task_id,
-                    callbacks=[{'task': handle_work_success.name, 'kwargs': {'task_actual': task_actual}}],
-                    errbacks=[{'task': handle_work_error.name, 'args': [task.celery_task_id], 'kwargs': {'subtasks': [task_actual] + dependencies}}],
-                )
+        # apply_async does a NOTIFY to the channel dispatcher is listening to
+        # postgres will treat this as part of the transaction, which is what we want
+        if task.status != 'failed' and type(task) is not WorkflowJob:
+            task_cls = task._get_task_class()
+            task_cls.apply_async(
+                [task.pk],
+                opts,
+                queue=task.get_queue_name(),
+                uuid=task.celery_task_id,
+                callbacks=[{'task': handle_work_success.name, 'kwargs': {'task_actual': task_actual}}],
+                errbacks=[{'task': handle_work_error.name, 'args': [task.celery_task_id], 'kwargs': {'subtasks': [task_actual] + dependencies}}],
+            )
 
-        task.websocket_emit_status(task.status)  # adds to on_commit
-        connection.on_commit(post_commit)
+        # In exception cases, like a job failing pre-start checks, we send the websocket status message
+        # for jobs going into waiting, we omit this because of performance issues, as it should go to running quickly
+        if task.status != 'waiting':
+            task.websocket_emit_status(task.status)  # adds to on_commit
 
     @timeit
     def process_running_tasks(self, running_tasks):
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -11,7 +11,7 @@ import sys
 import signal
 
 # Django
-from django.db import transaction, connection
+from django.db import transaction
 from django.utils.translation import gettext_lazy as _, gettext_noop
 from django.utils.timezone import now as tz_now
 from django.conf import settings
@@ -565,7 +565,6 @@ class TaskManager(TaskBase):
                 opts,
                 queue=task.get_queue_name(),
                 uuid=task.celery_task_id,
-                new_connection=False,  # we want the message to be contingent on transaction commit
                 callbacks=[{'task': handle_work_success.name, 'kwargs': {'task_actual': task_actual}}],
                 errbacks=[{'task': handle_work_error.name, 'args': [task.celery_task_id], 'kwargs': {'subtasks': [task_actual] + dependencies}}],
             )
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -11,7 +11,7 @@ import sys
 import signal
 
 # Django
-from django.db import transaction
+from django.db import transaction, connection
 from django.utils.translation import gettext_lazy as _, gettext_noop
 from django.utils.timezone import now as tz_now
 from django.conf import settings
@@ -565,6 +565,7 @@ class TaskManager(TaskBase):
                 opts,
                 queue=task.get_queue_name(),
                 uuid=task.celery_task_id,
+                db_conn=connection,
                 callbacks=[{'task': handle_work_success.name, 'kwargs': {'task_actual': task_actual}}],
                 errbacks=[{'task': handle_work_error.name, 'args': [task.celery_task_id], 'kwargs': {'subtasks': [task_actual] + dependencies}}],
             )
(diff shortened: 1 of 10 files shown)


AlanCoding:
Oh noes, I just realized that this sets `conn.set_session(autocommit=True)`, which undermines the whole point. Maybe I pass a connection in, I'm not sure.

kdelee (approved):

awx/main/scheduler/task_manager.py
```diff
@@ -552,22 +552,23 @@ def start_task(self, task, instance_group, dependent_tasks=None, instance=None):
...
+                callbacks=[{'task': handle_work_success.name, 'kwargs': {'task_actual': task_actual}}],
+                errbacks=[{'task': handle_work_error.name, 'args': [task.celery_task_id], 'kwargs': {'subtasks': [task_actual] + dependencies}}],
+            )
 
-        task.websocket_emit_status(task.status)  # adds to on_commit
-        connection.on_commit(post_commit)
+        # In exception cases, like a job failing pre-start checks, we send the websocket status message
+        # for jobs going into waiting, we omit this because of performance issues, as it should go to running quickly
+        if task.status != 'waiting':
+            task.websocket_emit_status(task.status)  # adds to on_commit
```
kdelee, line 575: I am 100% behind this. In a healthy system, the ui flashing the waiting state is near useless because it should almost immediately go into running. In an unhealthy system where jobs are hanging out in waiting, they will get the update on a hard page load. So to me cost > benefit and I support dropping sending this webhook message for normal jobs in waiting.

AlanCoding:
The current state of this PR doesn't do exactly what I thought it does, but perhaps my current understanding is more accurate.

To do `.apply_async`, it _establishes a new connection_ and notifies. This means that it won't abide by the transaction rules of the active transaction. Should it fail to send the message due to max_connections, then that will error and the task manager's other changes will be aborted. This would reverse our state inconsistency - messages sent when the status has not been flipped. However... we can just have the job abort when loading it and finding the status is "pending". I have not done that yet, but it could be a full solution.

Alternatively, I am trying to get it to work without establishing a new connection. It seems that I'm not the first person to beat their heads on this problem, and that bodes poorly for an eventual resolution.

AlanCoding:
I have gone into the history, and now believe there's no reason this updated version _can't_ work. Let me link things here.

The original introduction of this `PubSub` class goes back to https://github.com/ansible/awx/pull/6034, and within that, the commit https://github.com/ansible/awx/commit/2a2c34f56779e7ccd8337141efc3bef8f6695c3e

But if you look into what that commit's doing, it moves and rewrites [an archaic module](https://github.com/ansible/awx/blob/558e92806b0e692939faf113029ad322960c7369/awx/main/dispatch/worker/basepg.py) from the original work-in-progress. The old form imported from a library, https://pypi.org/project/pgpubsub/, and the update moved the `PubSub` class definition here. There were some interesting minor changes that happened in this conversion, like introducing a `@contextmanager`, which seems to follow directly from the library's documentation, where they made it clear that `PubSub` objects are not generally safe to pass around between threads (or to be long-lived _at all_). This is because they establish a new connection all on their own, and here is where we get the problem.

By establishing a new connection, that library is able to circumvent the Django transaction rules... which ultimately defer to the postgres transaction rules. Specific to pg_notify, see the documentation:

https://www.postgresql.org/docs/current/sql-notify.html

> Firstly, if a NOTIFY is executed inside a transaction, the notify events are not delivered until and unless the transaction is committed. This is appropriate, since if the transaction is aborted, all the commands within it have had no effect, including NOTIFY.

...that would be exactly what I want. It seems that the root problem here is that the author of `pgpubsub` went out of their way to avoid this behavior, and have it always send messages right away, transactions be damned. That would be consistent with the library's stated goal, which is to mimic the behavior of similar redis libraries. We got what they wanted, but that's not what _we_ should want for AWX.

So instead of continuing to cargo-cult this, I feel obligated to at least attempt the proper change, even if it is bold.

Now, `with pg_bus_conn() as conn:` will not establish a new connection. It will, quite simply, use the existing Django connection. If the connection is closed, it will connect, just as any query would. My only (very minor) gripe is possibly the naming, as `conn` should probably be renamed to `pubsub` or something, indicating that it's an object for interacting with pg_notify, not something that follows or defines a connection in a way unlike any other ORM actions. So far, this surprisingly, seems to show no difference from prior behavior, but I will run some more tests.

awx/main/scheduler/task_manager.py
```diff
@@ -552,22 +552,23 @@ def start_task(self, task, instance_group, dependent_tasks=None, instance=None):
                 task.save()
                 task.log_lifecycle("waiting")
 
-        def post_commit():
-            if task.status != 'failed' and type(task) is not WorkflowJob:
-                # Before task is dispatched, ensure that job_event partitions exist
-                create_partition(task.event_class._meta.db_table, start=task.created)
```
AlanCoding, line 562: @fosterseth here `create_partition` is moved into the task code when the job starts. This, plus removing the "waiting" websocket message, are both changes that remove `on_commit` logic from the task manager.

kdelee:
> Firstly, if a NOTIFY is executed inside a transaction, the notify events are not delivered until and unless the transaction is committed. This is appropriate, since if the transaction is aborted, all the commands within it have had no effect, including NOTIFY.

I agree this is what we want. Thank you for looking into this! I think this will help us on multiple fronts, making sure messages are not sent unless the transaction succeeds (so job state is consistent), as well as the move of create_partition and removal of the websocket message for waiting will also carve off a big expensive chunk off of the task manager that we knew was problematic.

fosterseth (approved):

AlanCoding:
Let me give an update - before the last 2 commits here I had a bug that made it so that _listeners_ would no longer work. Functionally, this means tasks didn't run, jobs stayed in pending, etc.

The ultimate culprit seems to come down to this, which is found here as a side effect of `self.pool.up()`.

https://github.com/ansible/awx/blob/38a0950f4640569b7164bed9560dd39404738ceb/awx/main/dispatch/pool.py#L243

This completely torpedoes the idea that a listener could use the existing Django connection. Because that connection might later be closed, that will disconnect you from the channel (see postgres docs). _Even if_ you were to re-connect, that leaves a gap that would open you up to lost messages, because you won't be listening for a time. So as the dispatcher is very frequently scaling workers up and down, and closing connections each time, it _must not_ close the connection on which it is listening. Unavoidably, this means that it has to use a separate connection for listening, versus making ORM queries.

Before, I was unsure if listeners should make a new connection or not, but this settles it. This still does nothing to the argument that _publishers_ (err.. notifiers) should use their existing connection. They should.

I don't know why I remember this was working for me before. Maybe the pool wasn't scaling at the time I tested it? I have confirmed the bug and confirmed that the last commits fix it.

I should also note that https://github.com/ansible/awx/pull/12656 is very relevant to this, and I do want to test them in conjunction.

AlanCoding:
Since this is removing the "waiting" status websocket, and we've now added timings for the on_commit methods, I wanted to compare timings to devel.

### devel as reference

Screenshot from 2022-08-15 21-26-11

### this PR

Screenshot from 2022-08-15 21-40-57

The main thing that matters is the scale of the orange compared to the red. This change here seems to make virtually no difference. That is surprising, and I'm not totally sure what the time is occupied with now (maybe the actual database transaction commit??)

Anyway, this change still necessitates removing the "waiting" websocket message, because if not, it's guaranteed that the "running" message would outrun the "waiting" message, leaving the UI with the wrong status. Unfortunately, we can't claim any task manager timing benefits along with this.

AlanCoding:
The one big item left to do here is check with the UI if removing the "waiting" status message is okay with them (doesn't mean there is any UI change needed, it seems to work now).

One thing I found through the tests is that the "waiting" websocket message provides the "instance_group_name" field. This is suspicious, considering that the task manager also sets other fields like the "execution_node" and we don't do anything about that. Also, the UI doesn't seem to use this field anywhere. For now, I'm going to _move_ the field to the "running" message, but it's arguable that deleting it would be better.

AlanCoding:
Oh I found one more real nuanced thing that comes with this.

```
tools_postgres_1 | 2022-08-17 02:29:17.246 UTC [105] NOTICE:  relation "main_projectupdateevent_20220817_02" already exists, skipping
tools_postgres_1 | 2022-08-17 02:29:17.246 UTC [105] STATEMENT:  CREATE TABLE IF NOT EXISTS main_projectupdateevent_20220817_02 PARTITION OF main_projectupdateevent FOR VALUES FROM ('2022-08-17 02:00:00+00:00') to ('2022-08-17 03:00:00+00:00');
```

This is a completely non-fatal event, but it's log noise. This is due to the fact that `create_partition` was moved to the main task code path, so we have extra calls to it after this is done.

This is easily fixed by the patch:

```diff
diff --git a/awx/main/tasks/jobs.py b/awx/main/tasks/jobs.py
index 9b11589547..fc08864241 100644
--- a/awx/main/tasks/jobs.py
+++ b/awx/main/tasks/jobs.py
@@ -918,7 +918,6 @@ class RunJob(BaseTask):
             local_project_sync = job.project.create_project_update(_eager_fields=sync_metafields)
             local_project_sync.log_lifecycle("controller_node_chosen")
             local_project_sync.log_lifecycle("execution_node_chosen")
-            create_partition(local_project_sync.event_class._meta.db_table, start=local_project_sync.created)
             # save the associated job before calling run() so that a
             # cancel() call on the job can cancel the project update
             job = self.update_model(job.pk, project_update=local_project_sync)
@@ -1587,7 +1586,6 @@ class RunInventoryUpdate(BaseTask):
             )
             local_project_sync.log_lifecycle("controller_node_chosen")
             local_project_sync.log_lifecycle("execution_node_chosen")
-            create_partition(local_project_sync.event_class._meta.db_table, start=local_project_sync.created)
             # associate the inventory update before calling run() so that a
             # cancel() call on the inventory update can cancel the project update
             local_project_sync.scm_inventory_updates.add(inventory_update)
```

_However_, I would like to punt on this, because it would introduce major git conflicts with the "copytree" PR. A very similar diff can be done after rebasing on top of that without conflicts.

...actually we seem to get this anyway, so the change hardly has any effect. It's still code we should prefer to get rid of.

PR #12582, opened 2022-07-27
Move reaper logic into worker, avoiding bottlenecks

AlanCoding:
##### SUMMARY
I'm hoping that this can be the final piece of the puzzle to address all the reports of bad behavior by the reaper with jobs in the "waiting" status.

Another piece is up at https://github.com/ansible/awx/pull/12573

With this change, we will still sometimes see slowness as `self.pool.up()` is called to expand the worker pool, but we will remove the risky behavior of running (potentially slow) database queries periodically from the main dispatcher process.

It is believed that slowness in the main dispatcher process can cause a pileup of incoming messages, which is what leaves jobs in the waiting state.

This moves the `.cleanup()` logic out of the main process and into a worker process, and runs at the same period it did before. Local reaper will be a side-effect of the cluster_node_heartbeat task, that was true before and will stay the same here. Instead of having the main dispatcher process _run_ the logic, it will instead attach a number of necessary parameters, so it will can be ran in the main body of the task.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
This will have conflicts as other merges happen. That's fine, there shouldn't be any problem resolving them.

### DIFF
--- a/awx/main/tasks/system.py
+++ b/awx/main/tasks/system.py
@@ -10,6 +10,7 @@ from contextlib import redirect_stdout
 import shutil
 import time
 from distutils.version import LooseVersion as Version
+from datetime import datetime
 
 # Django
 from django.conf import settings
@@ -482,8 +483,8 @@ def inspect_execution_nodes(instance_list):
                     execution_node_health_check.apply_async([hostname])
 
 
-@task(queue=get_local_queuename)
-def cluster_node_heartbeat():
+@task(queue=get_local_queuename, bind_kwargs=['dispatch_time', 'active_task_ids'])
+def cluster_node_heartbeat(dispatch_time=None, active_task_ids=None):
     logger.debug("Cluster node heartbeat task.")
     nowtime = now()
     instance_list = list(Instance.objects.all())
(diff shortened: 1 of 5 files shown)


awx/main/tasks/system.py
```diff
@@ -475,8 +476,8 @@ def inspect_execution_nodes(instance_list):
                     execution_node_health_check.apply_async([hostname])
 
 
-@task(queue=get_local_queuename)
-def cluster_node_heartbeat():
+@task(queue=get_local_queuename, bind_kwargs=['dispatch_time', 'active_task_ids'])
+def cluster_node_heartbeat(dispatch_time=None, active_task_ids=None):
```
AlanCoding: I use the "bind" wording because Celery has a `bind` kwarg

https://docs.celeryq.dev/en/stable/userguide/calling.html

In celery, this attaches a worker instance. In our case, the parameters we really want are from a separate process, so we can't just pass python objects. They need to go through serialization and through the socket communication.

I also imagine that "publish_time" could be added to this easily. I don't have a solid use case for that, other than some logging.

kdelee (approved):

AlanCoding:
This has been working relatively well, the only caveat is that there are non-trivial git conflicts with other branches.

AlanCoding:
There was some logic changes required by the rebase, but I feel confident this doesn't compromise anything now.

PR #12587, opened 2022-07-28
Allow for adding external execution nodes via API

jbradberry:
##### SUMMARY
In order to support scaling the Receptor mesh from within Collector, we need to track the state.

New states:

- provisioning: marks the new instance as one that needs to have Receptor and ansible-runner and so forth installed
- provision-fail: a provisioning job attempted to run on this node and failed
- installed: this instance already has the packages installed (perhaps because the admin manually installed them), and we are just waiting for the health check to bring it fully into the mesh
- deprovisioning: this instance needs to have the bits uninstalled. Once that succeeds, the instance can be removed from the database.
- deprovision-fail: a de-provisioning job attempted to run on this node and failed.

The states of 'unavailable' and 'ready' correspond to existing ideas about whether or not an instance has errors.

Additionally, we are adding a `link_state` field to `InstanceLink`. This allows us to specify peer relationships that should be added or removed, even if the instances on either side are not having any other operations performed on them.

- adding
- established
- removing

This PR will add

- [x] new state fields to the `Instance` and `InstanceLink`
- [x] new `Instance.listener_port` field
- [x] external provisioners (via `provision_instance`) will result in a state of Installed
- [x] the state fields will be exposed in the existing API endpoints
- [x] peer links will be exposed in the instance detail endpoint
- [x] the `enabled` boolean field will be exposed to the mesh visualizer
  - we decided to keep `enabled` as a separate field instead of capturing it as a new state, since its use for the user intention of allowing jobs to run or not run is orthogonal to the states
- [x] the health checks will now capture changes using the state field
- [x] jobs will now only use instances in the Ready state
- [x] ~~the `instances/` endpoint will allow creation of new to-be-provisioned instances~~  **kicked into #12616**
- [x] ~~the `instances/<id>/` endpoint will allow changing the state of existing instances, within certain parameters~~  **kicked into #12616**
- [x] ~~changes to peer links will be postable~~  **descoped**
- [x] ~~changes to instance groups will be postable~~  **descoped**

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API


awx/main/migrations/0165_node_and_link_state.py
```diff
@@ -0,0 +1,49 @@
...
+
+def forwards(apps, schema_editor):
+    # All existing InstanceLink objects need to be in the state
+    # 'Established', which is the default, so nothing needs to be done
+    # for that.
+
+    Instance = apps.get_model('main', 'Instance')
+    for instance in Instance.objects.all():
+        instance.node_state = 'ready' if not instance.errors else 'unavailable'
+        instance.save(update_fields=['node_state'])
```
AlanCoding, line 15: this will be a no-op in most cases... but the table size is small in the worst of cases, so I'm fine with all of this.

awx/main/models/ha.py
```diff
@@ -126,13 +133,25 @@ class Instance(HasPolicyEditsMixin, BaseModel):
...
+    class States(models.TextChoices):
+        PROVISIONING = 'provisioning', _('Provisioning')
+        PROVISION_FAIL = 'provision-fail', _('Provisioning Failure')
+        INSTALLED = 'installed', _('Installed')
+        READY = 'ready', _('Ready')
+        UNAVAILABLE = 'unavailable', _('Unavailable')
+        DEPROVISIONING = 'deprovisioning', _('De-provisioning')
+        DEPROVISION_FAIL = 'deprovision-fail', _('De-provisioning Failure')
+
+    node_state = models.CharField(choices=States.choices, default=States.READY, max_length=16)
```
AlanCoding: It might be good to put in `help_text` for these new fields because I love love that as a means of self-documentation.

To clarify, any states related to provisioning are only intended for nodes managed by this new whatever-you-call-it dynamic mesh management feature.

TheRealHaoLiu (approved):
yolo

PR #12592, opened 2022-07-29
Use in-memory data structures to correctly reap orphaned tasks

AlanCoding:
##### SUMMARY
This is very much a WIP, because to complete it, we need a good deal of other things to merge first.

I believe this will help a lot to avoid some real reports from the community that OCP deployments can leave jobs stuck in running/waiting forever. Our reaper just wasn't working because it didn't account for the current job types that we support (like execution node jobs, container group jobs). I'm pretty happy with this overall structure.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
My single hesitation here is that running/waiting jobs with a cached `task_impact` may not be something that we want to fetch forever. Capacities could be calculated with a Django annotation instead. In that case, we would need a different kind of hook to run this logic. We may have 2 different loops in different places, one for pending jobs and another for running/waiting. Even if we go this route, we should have _correct_ reaping logic, which I think this offers.

### DIFF
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -20,7 +20,6 @@ from django.contrib.contenttypes.models import ContentType
 # AWX
 from awx.main.dispatch.reaper import reap_job
 from awx.main.models import (
-    Instance,
     InventorySource,
     InventoryUpdate,
     Job,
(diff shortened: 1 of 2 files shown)


kdelee (approved):
:tada: :exploding_head: Yes. This is good.

I presume you made it to the task_manager_refactor branch to avoid rebase issues.

AlanCoding:
Yes, that's exactly why. I am assuming that a merge of `task_manager_refactor` branch is imminent and that most things in this file have moved.

awx/main/scheduler/task_manager.py
```diff
@@ -688,16 +687,26 @@ def task_needs_capacity(self, task, tasks_to_update_job_explanation):
...
+
+            if task.execution_node:
+                if task.execution_node not in self.instances:
+                    self.all_tasks.remove(task)
+                    reap_job(task, 'failed')
+            elif task.instance_group_id:
+                # for container group jobs
+                if task.instance_group_id not in self.instance_groups.pk_ig_map:
+                    self.all_tasks.remove(task)
+                    reap_job(task, 'failed')
```
AlanCoding: one more nit pick - in either of these cases, we should probably do `task.cancel()`. Maybe in a separate task? I know, it's a weird idea.

kdelee (approved):

AlanCoding:
The one big item left to do here is check with the UI if removing the "waiting" status message is okay with them (doesn't mean there is any UI change needed, it seems to work now).

One thing I found through the tests is that the "waiting" websocket message provides the "instance_group_name" field. This is suspicious, considering that the task manager also sets other fields like the "execution_node" and we don't do anything about that. Also, the UI doesn't seem to use this field anywhere. For now, I'm going to _move_ the field to the "running" message, but it's arguable that deleting it would be better.

awx/main/scheduler/task_manager.py
AlanCoding: I'm getting cold feet about the idea of a container group reaper.

https://github.com/ansible/awx/blob/cc1e349ea8ecfb25cc21127751522a0be6acc284/awx/main/models/unified_jobs.py#L699

The field is SET_NULL, so the event of `elif task.instance_group_id:` without the instance in the in-memory mapping should only happen due to race conditions, and isn't a good metric for reaping. Perhaps we could do this if `task.status == 'running'` and the instance group is _Null_, but that just buys us into the task of requesting canceling, as mentioned.

AlanCoding:
From manual testing:

Screenshot from 2022-08-17 14-45-59

But this quickly raised some new questions - in particular, if I were to _disable_ the node awx_1, the job should _not_ be reaped, and that was a gap in my implementation. We normally _don't fetch_ these instances, limiting our query to the nodes that are enabled and capacity>0. Because of that, I had to add some additional logic and fetch the full hostlist (only once!) to make this complete.

Reading back over the related issue, I see that the case of reaping jobs in deleted container_groups was never mentioned directly - we only know this happened from getting the controller_node deleted. So it still seems like an appropriate stop-gap to go with the updated implementation here.

AlanCoding:
No, this still has hazards, because if `execution_node` is dead, then we need to actually cancel the job (or else it may flip right back to successful/error). Perhaps, for now, the only thing we can do is to enact this reaping if the `controller_node` is dead.

AlanCoding:
Closing b/c this is too risky and not developed enough, and not possible to finish near-term.

Issue #12594, opened 2022-07-30
awx collection job template and execution environment fail to properly lookup some items using organization field.


sean-m-sullivan:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

awx collection job template and execution environment fail to properly lookup some items using organization field.
 
For job template it is credentials, and inventories, for execution environments its credentials. 

Generally its assumed if you are applying an organization its looking for things with the organizations, its an issue when the same object name is duplicated in different orgs. In other cases we have been using search fields to resolve this, or require the user to supply the ID. 

I'll be putting in a PR with tests to fix this.

### AWX version

21.3.0

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [X] Collection
- [ ] CLI
- [ ] Other

### Installation method

minikube

### Modifications

no

### Steps to reproduce

Run the module

### Expected results

It assigns the inventory/credential in the org

### Actual results

2 items are returned from two different orgs.

PR #12595, opened 2022-07-30
update execution environment and job template modules to use organizatioin lookup when looking for sub objects when supplied

sean-m-sullivan:
##### SUMMARY
Fix lookup of objects to refer to helper org.
#12594

##### ISSUE TYPE
 - Bug or Docs Fix

##### COMPONENT NAME
 - Collection

##### AWX VERSION
```
21.3.0
```

### DIFF
--- a/awx_collection/plugins/modules/execution_environment.py
+++ b/awx_collection/plugins/modules/execution_environment.py
@@ -115,12 +115,14 @@ def main():
 
     # Attempt to look up the related items the user specified (these will fail the module if not found)
     organization = module.params.get('organization')
+    search_fields = {}
     if organization:
-        new_fields['organization'] = module.resolve_name_to_id('organizations', organization)
+        search_fields['organization'] = new_fields['organization'] = module.resolve_name_to_id('organizations', organization)
 
     credential = module.params.get('credential')
     if credential:
-        new_fields['credential'] = module.resolve_name_to_id('credentials', credential)
+        credential_object =  module.get_one('credentials', name_or_id=credential, **{'data': search_fields})
+        new_fields['credential'] = credential_object['id']
 
     module.create_or_update_if_needed(existing_item, new_fields, endpoint='execution_environments', item_type='execution_environment')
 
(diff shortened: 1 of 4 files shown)


awx_collection/plugins/modules/execution_environment.py
```diff
@@ -115,12 +115,14 @@ def main():
...
+    search_fields = {}
     if organization:
-        new_fields['organization'] = module.resolve_name_to_id('organizations', organization)
+        search_fields['organization'] = new_fields['organization'] = module.resolve_name_to_id('organizations', organization)
 
     credential = module.params.get('credential')
     if credential:
-        new_fields['credential'] = module.resolve_name_to_id('credentials', credential)
+        credential_object =  module.get_one('credentials', name_or_id=credential, **{'data': search_fields})
+        new_fields['credential'] = credential_object['id']
```
john-westcott-iv, line 125: This seems like it would be an anti-pattern to the lookup module we have going on. For example, f I want my EE in org1 to use a cred from org2. i.e. 
```
- execution_environmnet:
    organization: "My EE org"
    credential: "{{ lookup('awx.awx.tower_api', 'credential', { 'name': 'My Cred Name', 'organization': 'The org my cred belongs to'}) }}"
```

 I've already done the cred lookup with its org but now you are looking up the cred with an overridden org specified by the EE.
 
 Do you agree or am I thinking about this wrong?

sean-m-sullivan: I think you are right, I think I may have just gone to far on this one. After writing it, and testing it, I was like, Maybe this is to far, enforcing this. I think it was more that was trying to tie things in, I am almost thinking a rewrite where you do things like
```
- execution_environmnet:
    organization: "My EE org"
    credential: 
        name: EE1
        organization: Org2
```
and lists for those that do that, so a different org can be supplied for every sub object.

john-westcott-iv: That could be its own rabbit hole as we would have to what pieces of information are required too lookup which objects.

Issue #12598, opened 2022-07-31
update role plugin to provide better error messages


sean-m-sullivan:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

Currently if the role plugin fails to find an object to assign a role to, it fails with a python error. If you are submitting for a team and 100 job templates and workflows, this is nigh unworkable to find out which one is missing. This should be fixed. 

### AWX version

21.3.0

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [X] Collection
- [ ] CLI
- [ ] Other

### Installation method

minikube

### Modifications

no

### Steps to reproduce

Run roles module with a team/job template/workflow, something that doesn't exist. 
```
    - name: Set user to admin over inventory
      awx.awx.role:
        team: Service Provisioning
        role: execute
        lookup_organization: Default
        state: present
        job_templates:
          - Azure
        workflows:
          - Azure Nothing
```

### Expected results

 "Unable to find workflows with name: Azure Nothing"

### Actual results

An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: 'NoneType' object is not subscriptable
fatal: [localhost]: FAILED! => {"changed": false, "module_stderr": "Traceback (most recent call last):
  File \"/home/excalibrax/.ansible/tmp/ansible-tmp-1659290592.3939743-111457-226432510653785/AnsiballZ_role.py\", line 107, in <module>
    _ansiballz_main()
  File \"/home/excalibrax/.ansible/tmp/ansible-tmp-1659290592.3939743-111457-226432510653785/AnsiballZ_role.py\", line 99, in _ansiballz_main
    invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)
  File \"/home/excalibrax/.ansible/tmp/ansible-tmp-1659290592.3939743-111457-226432510653785/AnsiballZ_role.py\", line 47, in invoke_module
    runpy.run_module(mod_name='ansible_collections.awx.awx.plugins.modules.role', init_globals=dict(_module_fqn='ansible_collections.awx.awx.plugins.modules.role', _modlib_path=modlib_path),
[...]

PR #12599, opened 2022-07-31
Update role module to provide better error message when an object does not exist

sean-m-sullivan:
##### SUMMARY
Currently if the role plugin fails to find an object to assign a role to, it fails with a python error. If you are submitting for a team and 100 job templates and workflows, this is nigh unworkable to find out which one is missing. This should be fixed.

Follows up issue  #12598

##### ISSUE TYPE
 
##### COMPONENT NAME
 - Collection

##### AWX VERSION
```
21.3.0
```

### DIFF
--- a/awx_collection/plugins/modules/role.py
+++ b/awx_collection/plugins/modules/role.py
@@ -241,6 +241,7 @@ def main():
     # Lookup actor data
     # separate actors from resources
     actor_data = {}
+    missing_items = []
     for key in ('user', 'team'):
         if key in resources:
             if key == 'user':
@@ -248,9 +249,14 @@ def main():
             else:
                 lookup_data_populated = lookup_data
             # Attempt to look up project based on the provided name or ID and lookup data
-            actor_data[key] = module.get_one('{0}s'.format(key), name_or_id=resources[key], data=lookup_data_populated)
-            resources.pop(key)
-
+            data = module.get_one('{0}s'.format(key), name_or_id=resources[key], data=lookup_data_populated)
+            if data is None:
+                module.fail_json(
+                    msg='Unable to find {0} with name: {1}'.format(key, resources[key]), changed=False
+                )
+            else:
+                actor_data[key] = module.get_one('{0}s'.format(key), name_or_id=resources[key], data=lookup_data_populated)
+                resources.pop(key)
     # Lookup Resources
     resource_data = {}
     for key, value in resources.items():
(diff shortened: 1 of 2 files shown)


shanemcd:
Hey Sean - thanks for all the good work. One request - we use PR titles to generate the release notes for AWX. Can you reword the titles on all the PRs you have open so that they make sense in that context?

sean-m-sullivan:
> 

@shanemcd I have updated all 3, is there any prefix or anything else you would like me to add?

awx_collection/plugins/modules/role.py
```diff
@@ -261,8 +266,13 @@ def main():
                     lookup_data_populated = {}
                 else:
                     lookup_data_populated = lookup_data
-
-            resource_data.setdefault(key, []).append(module.get_one(key, name_or_id=resource, data=lookup_data_populated))
+            data = module.get_one(key, name_or_id=resource, data=lookup_data_populated)
+            if data is None:
+                module.fail_json(
+                    msg='Unable to find {0} with name: {1}'.format(key, resource), changed=False
```
john-westcott-iv: This comment applies for both loops here.

In the case where >1 item was missing this would just give you the last message for an item that it failed to find. Can we make an array like `missing_items = []` and append these messages to the array. After the loops, before the module changes things we could exit early in this scenario to prevent a partial update with something like:
```
if len(missing_items) > 0:
    msg='There were {} missing items, see missing_items for more details'.format(len(missing_items))
    module.fail_json(msg, missing_items, changed=False)
```

(I don't remember if ^ is that exact syntax for exiting with another parameter or not)

john-westcott-iv (changes requested):

awx_collection/plugins/modules/role.py
john-westcott-iv: Also, if we do this add a test for >1 error as well.

sean-m-sullivan: In practice its usually one item, but I can see about doing it in loop to catch more.

sean-m-sullivan: Did testing around this, for the user/team keys it fails before you get to the apply to list so I am leaving it as my ofiginal change, as there is only one item provided from the user, and it fails if the thing to be granted permissions to act on lists of objects does not exist.

CFSNM (approved):

john-westcott-iv (approved):

PR #12620, opened 2022-08-03
Restrict files that trigger a reload

jbradberry:
##### SUMMARY
Restrict files that trigger a reload to files explicitly ending in '.py' that do not start with a dot.
This will avoid Emacs lockfiles from triggering the restart.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION

old style:
```
$ file='.#serializers.py'
$ if [[ "$file" =~ .*py$ ]]; then echo "true"; else echo "false"; fi
true
```

new style:
```
$ file='.#serializers.py'
$ if [[ "$file" =~ ^[^.].*\.py$ ]]; then echo "true"; else echo "false"; fi
false
[jrb@fedora awx]$ file2=foo.py
[jrb@fedora awx]$ if [[ "$file2" =~ ^[^.].*\.py$ ]]; then echo "true"; else echo "false"; fi
true
```


shanemcd (approved):

relrod (approved):
lgtm

john-westcott-iv (approved):

Issue #11423, opened 2021-12-06
Workflow Tags Prompt on Launch


exp-hc:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Workflows could accept parameters like `ask_job_tags_on_launch` and `ask_skip_tags_on_launch`. This would allow you to apply tags (skip tags) at the workflow level during launch without creating 2 separate workflow to use tagging.

AlanCoding:
Link https://github.com/ansible/awx/issues/7859

@ffirg we have added these workflow "prompt" fields one-by-one, and my argument is that it will take considerably less time to do a refactor where we have workflow prompts be at full parity with JT prompts.

AlanCoding:
Other recent prompting requests
 - prompting for EE (not tracked)
 - prompting for instance group or instance https://github.com/ansible/awx/issues/5711

It would be best to handle the workflow prompting as a separate issue from these two, which apply to jobs. Unless, of course, it is expected that the new job prompts would be available on workflows. In that case, it's most efficient to first add the prompting to jobs and then apply it to workflows.

ffirg:
possibly related? https://github.com/ansible/awx/issues/10719

mabashian:
Addressed by https://github.com/ansible/awx/pull/12875

PR #12622, opened 2022-08-04
add fields and features to Workflow Templates

sarabrajsingh:
##### SUMMARY

Originating GH Issue - https://github.com/ansible/awx/issues/11423

This PR introduces many features/options to Workflow Templates.

1. - [X] `prompt-on-launch` option to Labels field 
2. - [X] `ask_jobs_tags_on_launch` to the Workflow form (#11423)
3. - [X] `ask_skip_tags_on_launch` to the Workflow form (#11423)

##### prompt-on-launch field for WFJT
image

##### ask_skip_tags_on_launch with skip_tags field and prompt-on-launch on WFJT form
image

##### ask_job_tags_on_launch with job_tags field and prompt-on-launch on WFJT form
image

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 - UI

##### AWX VERSION

```
21.4.1.dev1+gacec9ad51e
```


AlexSCorey (commented):
This is a really good start.  I don't see anywhere the user can add labels at launch time, though.  Should that be part of this work?

Also, we probably want to add an assertion [here](https://github.com/ansible/awx/blob/devel/awx/ui/src/screens/Template/shared/WorkflowJobTemplateForm.test.js#L250)

awx/main/models/workflow.py
```diff
@@ -441,9 +441,18 @@ class Meta:
         blank=True,
         default=False,
     )
-    notification_templates_approvals = models.ManyToManyField("NotificationTemplate", blank=True, related_name='%(class)s_notification_templates_for_approvals')
-
-    admin_role = ImplicitRoleField(parent_role=['singleton:' + ROLE_SINGLETON_SYSTEM_ADMINISTRATOR, 'organization.workflow_admin_role'])
+    ask_labels_on_launch = AskForField(
+        blank=True,
+        default=False,
+    )
```
AlanCoding: What's the history of the request for `ask_labels_on_launch` here? I don't see this field applied to job templates, either here or previously.

Workflows should be essentially mirroring the `ask_` fields from job templates. Applying those prompts to launch JTs in the workflow is the point.

I was talking with @john-westcott-iv regarding new prompt-able fields he's adding, and we would prefer to not add it in multiple places. Instead of adding a new `ask_` field to workflows, I'd advocate that we should _start_ with creating an abstract parent class to house all of those fields, to enforce standardization between the models that define prompts (WFJT, JT), we have a model `LaunchConfigBase` to be the abstract parent of all models that _contain_ prompts, and that is already used here.

AlanCoding: Looking back over this, we already have a base model that is used strictly by JTs and WFJTs, which is `SurveyJobTemplateMixin`. This makes sense, because surveys are a kind of prompt.

I'd advocate prompting all the `ask_` fields up the inheritance chain from JT to this model. You may want to rename it as well. This would be the abstract base for any "askable" type of model. Then the form to fill out prompts should essentially be the same in all cases, but practically, this means JTs and WFJTs.  Let me know if you see any reason this should not hold true.

john-westcott-iv: Nice call out on the `SurveyJobTemplateMixin`. Once we get the JT code in this feature branch we should migrate common ask fields into that class.

sarabrajsingh: @john-westcott-iv and myself will merge our JT and WFJT template changes into a common branch and work on the refactor there (https://github.com/ansible/awx/tree/feature-prompt-on-launch-on-templates) as well as work out the failing tests and generate a monolithic migration there.

john-westcott-iv:
From a quick API test it looks like the job_tags and skip job tags aren't saving on the WFJT. Also, can we fix the lint/swagger checks and potentally the awx-collection as well?

jbradberry:
While I'm generally not a proponent of squashing or other major edits, no one wants to see a bunch of non-descriptive typo and "fix" commits in the history.  Please make use of `git commit --amend` more often to keep this under control.

awx/api/serializers.py
```diff
@@ -3417,6 +3420,10 @@ class Meta:
             'webhook_service',
             'webhook_credential',
             '-execution_environment',
+            'ask_labels_on_launch',
+            'ask_skip_tags_on_launch',
```
john-westcott-iv, line 3426: What about `ask_job_tags_on_launch`? Are we not going to be prompting for them (this option seems to be missing in other spots as well)

awx/api/serializers.py
```diff
@@ -3455,6 +3462,7 @@ def get_related(self, obj):
     def validate_extra_vars(self, value):
         return vars_validate_or_raise(value)
 
+    # posting
```
john-westcott-iv, line 3468: Remove or elaborate?

awx/api/views/__init__.py
```diff
@@ -3197,13 +3197,17 @@ def update_raw_data(self, data):
                 data['extra_vars'] = extra_vars
             modified_ask_mapping = models.WorkflowJobTemplate.get_ask_mapping()
             modified_ask_mapping.pop('extra_vars')
-            for field_name, ask_field_name in obj.get_ask_mapping().items():
```
john-westcott-iv, line 3200: Was this a bug that it wasn't using the modified_ask_mapping?

awx/main/migrations/0168_wfjt_prompt_all.py
```diff
@@ -0,0 +1,39 @@
+# Generated by Django 3.2.13 on 2022-08-21 01:59
+
+import awx.main.fields
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('main', '0167_jt_prompt_everything_on_launch'),
```
john-westcott-iv: We should merge the two of these in the feature branch along with any refactoring we do.

awx/main/models/workflow.py
```diff
@@ -466,6 +475,24 @@ class Meta:
...
+    skip_tags = models.CharField(
+        max_length=1024,
+        blank=True,
+        default='',
+    )
+    ask_tags_on_launch = AskForField(
+        blank=True,
+        default=False,
+        allows_field='job_tags',
+    )
```
john-westcott-iv: Stylistically, can we move these up with the other ask options? Also labels is not here because they are a related field correct?

awx/main/tests/factories/fixtures.py
```diff
@@ -210,7 +210,7 @@ def mk_workflow_job_template(name, extra_vars='', spec=None, organization=None,
     if extra_vars:
         extra_vars = json.dumps(extra_vars)
 
-    wfjt = WorkflowJobTemplate(name=name, extra_vars=extra_vars, organization=organization, webhook_service=webhook_service)
+    wfjt = WorkflowJobTemplate.objects.create(name=name, extra_vars=extra_vars, organization=organization, webhook_service=webhook_service)
```
john-westcott-iv, line 213: Can we add some details in the conversation history as to why we need to switch too .objects.create.

john-westcott-iv:
Job and skip tags don't seem to save on a patch to api/v2/workflow_job_templates/ID/. Could be my setup, do they work for you?

john-westcott-iv:
Limit comes out as null on /api/v2/workflow_job_templates/ID/ which can't be submitted without getting:
```
{
    "limit": [
        "This field may not be null."
    ]
}
```

Not sure if this was something that we changed or not. If so, maybe we could give it a default of `''` so that you don't have to specify to launch.

john-westcott-iv:
After submitting a WFJT launch I see the labels I added but the specified `skip_tags` were not shown in the result. This can also be seen on api/v2/workflow_jobs/<job ID>/

john-westcott-iv:
When launching the WFJT with labels/skip tags those do not appear to have propagated down to the JT the WFJT ran.

awx/api/serializers.py
sarabrajsingh: removed.

sarabrajsingh:
just a development note -

we had to change the way that WFJTs were being instantiated, factoried (for lack of a better term) and/or created in the unit and functional tests.

##### old way:
```
wfjt = WorkflowJobTemplate(...)
```
##### new way:
```
wfjt = WorkflowJobTemplate.objects.create(...)
```

leveraging django's built in **.objects.create() method for Django-y objects was necessary to avoid failing unit and functional tests. this change correctly creates WFJT objects in the database, which can then be consumed in later checks.

awx/ui/src/components/LaunchPrompt/LaunchPrompt.test.js
```diff
@@ -150,20 +154,21 @@ describe('LaunchPrompt', () => {
...
+    expect(steps[5].name.props.children).toEqual('Survey');
+    expect(steps[6].name.props.children).toEqual('Preview');
     expect(wizard.find('WizardHeader').prop('title')).toBe('Launch | Foobar');
     expect(wizard.find('WizardHeader').prop('description')).toBe(
       'Foo Description'
     );
   });
 
-  test('should add inventory step', async () => {
+  test.only('should add inventory step', async () => {
```
keithjgrant:
```suggestion
  test('should add inventory step', async () => {
```

awx/main/models/workflow.py
```diff
@@ -366,6 +366,32 @@ class WorkflowJobOptions(LaunchTimeConfigBase):
...
+    ask_skip_tags_on_launch = AskForField(
+        blank=True,
+        default=False,
+    )
```
AlanCoding: I don't see any reason we can't move these to the `SurveyJobTemplateMixin` abstract model, which is specifically intended for reuse between JTs and WFJTs. These aren't new fields, JTs already have them, so instead of copying them, can't we just move them up the class inheritance structure?

awx/main/migrations/0168_wfjt_prompt_all.py
```diff
@@ -0,0 +1,59 @@
...
+        ),
+        migrations.AddField(
+            model_name='workflowjob',
+            name='ask_skip_tags_on_launch',
+            field=awx.main.fields.AskForField(blank=True, default=False),
+        ),
+        migrations.AddField(
+            model_name='workflowjob',
+            name='ask_tags_on_launch',
+            field=awx.main.fields.AskForField(blank=True, default=False),
```
AlanCoding: Why should `WorkflowJob` get `ask_tags_on_launch`? I don't think that reflects the requirements/needs here.

awx/main/migrations/0168_wfjt_prompt_all.py
```diff
@@ -0,0 +1,59 @@
+# Generated by Django 3.2.13 on 2022-08-31 17:30
```
AlanCoding: I would prefer that the feature branches keep it limited to a single migration file. These are ultimately user-facing, but the reason for separating them is the notion of keeping AWX releasable, and if that really is the logic, then feature branches should not have any problem combining migration files.

awx/main/models/workflow.py
```diff
@@ -588,6 +616,12 @@ def node_prompts_rejected(self):
                 node_list.append(node.pk)
         return node_list
 
+    def save(self, *args, **kwargs):
+
+        self.validate_unique()
+
+        super(WorkflowJobTemplate, self).save(*args, **kwargs)
```
AlanCoding: This `validate_unique` is only used to enforce `SOFT_UNIQUE_TOGETHER`, and it's in the base model. I really don't think this should be necessary. I can't think of a good reason we shouldn't delete this `save` method definition here.

awx/main/tests/functional/models/test_workflow.py
```diff
@@ -269,13 +269,17 @@ def test_always_success_failure_creation(self, wfjt, admin, get):
...
+        wfjt1 = WorkflowJobTemplate.objects.create(name='foo', organization=organization)
         wfjt1.save()
-        wfjt2 = WorkflowJobTemplate(name='foo', organization=organization)
         with pytest.raises(ValidationError):
-            wfjt2.validate_unique()
-        wfjt2 = WorkflowJobTemplate(name='foo', organization=None)
-        wfjt2.validate_unique()
+            WorkflowJobTemplate.objects.create(name='foo', organization=organization)
+        wfjt2 = WorkflowJobTemplate.objects.create(name='foo', organization=None)
+        wfjt2.save()
```
AlanCoding: I dug into what's happening here, and I think I've got a pulse on what's going on.

The `full_clean` method calls the `validate_unique` method. But importantly, that is _not_ called inside of the model code. Even if you read the Django code itself, it's clear that this is done as a part of the form validation, not ran for every model save. Should we add the call to `validate_unique`, that will run the logic _twice_ for real requests in production, and this adds cruft which makes it more difficult to maintain due to the confusion introduced.

The structure of the framework dictates that the _serializer_ has the responsibility for running this validation. This test _was confused_ already. It wasn't a good test. We could either directly call the methods in all cases, or call the serializer in all cases. Backing out to get the full picture, I ran this:

```python
    def test_wfjt_unique_together_with_org(self, organization, admin, post):
        '''
        uniqueness on a workflow object is determined by the (name, organization) tuple.
        and leverages the SOFT_UNIQUE_TOGETHER variable defined on the given model; as a check to uniqueness
        '''
        post(
            url=reverse('api:workflow_job_template_list'),
            data=dict(name='foo', organization=organization.id),
            user=admin, expect=201
        )
        r = post(
            url=reverse('api:workflow_job_template_list'),
            data=dict(name='foo', organization=organization.id),
            user=admin
        )
        assert r.status_code == 400
        assert r.data
        post(
            url=reverse('api:workflow_job_template_list'),
            data=dict(name='foo', organization=None),
            user=admin, expect=201
        )
```

This makes it pass even when the new WFJT validate_unique is deleted. That articulates the direction I'd like to go in here.

AlanCoding (approved):

adworjan:
Are there plans to also include credential prompt on launch capability down the line for workflows?

PR #12623, opened 2022-08-04
Adds functionality to provision or add an instance

AlexSCorey:
##### SUMMARY
This adds functionality to provision/add an mesh instance.  You have to have a kubernetes deployment, or override the IS_K8S variable [here](https://github.com/ansible/awx/compare/feature-mesh-scaling...AlexSCorey:6036-t-AddInstance?expand=1#diff-828a318d5bf2548b3b7ebc88148b7cf48ae20fe6062afea8ddcef439f20c9938R48)

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui/src/screens/Instances/InstanceAdd/InstanceAdd.js
+++ b/awx/ui/src/screens/Instances/InstanceAdd/InstanceAdd.js
@@ -1,0 +1,45 @@
+import React, { useState } from 'react';
+import { useHistory } from 'react-router-dom';
+import { Card, PageSection } from '@patternfly/react-core';
+import { InstancesAPI } from 'api';
+import InstanceForm from '../Shared/InstanceForm';
+
+function InstanceAdd() {
+  const history = useHistory();
+  const [formError, setFormError] = useState();
+  const handleSubmit = async (values) => {
+    const { instanceGroups, executionEnvironment } = values;
+    values.execution_environment = executionEnvironment?.id;
+
+    try {
+      const {
+        data: { id },
+      } = await InstancesAPI.create();
+
+      for (const group of instanceGroups) {
+        await InstancesAPI.associateInstanceGroup(id, group.id);
+      }
+      history.push(`/instances/${id}/details`);
+    } catch (err) {
+      setFormError(err);
+    }
+  };
+
+  const handleCancel = () => {
+    history.push('/instances');
+  };
+
+  return (
+    <PageSection>
+      <Card>
+        <InstanceForm
+          submitError={formError}
+          handleSubmit={handleSubmit}
+          handleCancel={handleCancel}
+        />
+      </Card>
+    </PageSection>
+  );
+}
+
+export default InstanceAdd;
--- a/awx/ui/src/screens/Instances/InstanceAdd/InstanceAdd.js
+++ b/awx/ui/src/screens/Instances/InstanceAdd/InstanceAdd.js
@@ -8,17 +8,11 @@ function InstanceAdd() {
   const history = useHistory();
   const [formError, setFormError] = useState();
   const handleSubmit = async (values) => {
-    const { instanceGroups, executionEnvironment } = values;
-    values.execution_environment = executionEnvironment?.id;
-
     try {
       const {
         data: { id },
-      } = await InstancesAPI.create();
+      } = await InstancesAPI.create(values);
 
-      for (const group of instanceGroups) {
-        await InstancesAPI.associateInstanceGroup(id, group.id);
-      }
       history.push(`/instances/${id}/details`);
     } catch (err) {
       setFormError(err);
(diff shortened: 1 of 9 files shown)


awx/ui/src/screens/Instances/InstanceList/InstanceList.js
```diff
@@ -135,6 +140,15 @@ function InstanceList() {
                 onExpandAll={expandAll}
                 qsConfig={QS_CONFIG}
                 additionalControls={[
+                  ...(isK8
```
kialam: we should also add a check for if the logged in user is a SysAdmin

awx/ui/src/screens/Instances/Shared/InstanceForm.js
```diff
@@ -0,0 +1,130 @@
...
+import {
+  ExecutionEnvironmentLookup,
+  InstanceGroupsLookup,
+} from 'components/Lookup';
+
+// This is hard coded because the API does not have the ability to send us a list that contains
+// only the types of instances that can be added.  Control and Hybrid instances cannot be added.
+
+const INSTANCE_TYPES = [
+  { id: 2, name: t`Execution`, value: 'execution' },
```
kialam: `id` here is used by `AnsibleSelect` to set the `select` element id attribute. We're not passing it as part of the payload and should make this a string value and not an integer.

awx/ui/src/screens/Instances/InstanceAdd/InstanceAdd.js
```diff
@@ -0,0 +1,45 @@
...
+    const { instanceGroups, executionEnvironment } = values;
+    values.execution_environment = executionEnvironment?.id;
+
+    try {
+      const {
+        data: { id },
+      } = await InstancesAPI.create(values);
+
+      for (const group of instanceGroups) {
+        await InstancesAPI.associateInstanceGroup(id, group.id);
```
kialam: spoke with @jbradberry and I think we want to allow default behavior here. E.g. the user should be able to create an Instance and not associate it with any instance group. They can always associate it later in the InstancesGroup list page.

awx/ui/src/screens/Instances/Shared/InstanceForm.js
```diff
@@ -0,0 +1,130 @@
...
+            label: type.name,
+            isDisabled: false,
+          }))}
+          value={instanceType.value}
+          onChange={(e, opt) => {
+            instanceTypeHelpers.setValue(opt);
+          }}
+        />
+      </FormGroup>
+      <InstanceGroupsLookup
```
kialam: Same as the comment earlier, we probably don't need this lookup.

awx/ui/src/api/models/Instances.js
```diff
@@ -1,6 +1,7 @@
 import Base from '../Base';
+import InstanceGroupsMixin from '../mixins/InstanceGroups.mixin';
 
-class Instances extends Base {
+class Instances extends InstanceGroupsMixin(Base) {
```
kialam: Not sure we need this.

awx/ui/src/screens/Instances/InstanceList/InstanceList.js
AlexSCorey: I thought about that...and I'm happy to do that, but this section of the app is only available to sys admins

awx/ui/src/screens/Instances/InstanceList/InstanceList.js
```diff
@@ -135,6 +142,15 @@ function InstanceList() {
                 onExpandAll={expandAll}
                 qsConfig={QS_CONFIG}
                 additionalControls={[
+                  ...(isK8 && me.is_superuser
```
jbradberry, line 145: minor terminology pick: K8S is short for K + 8 characters (ubernete) + S.  You should probably keep the S.

awx/ui/src/screens/Instances/InstanceAdd/InstanceAdd.test.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import { act } from 'react-dom/test-utils';
+import { createMemoryHistory } from 'history';
+import { InstancesAPI } from 'api';
+import {
+  mountWithContexts,
```
kialam, line 6: For net new unit tests we should be using react testing library and not enzyme, however we can merge the PR for now with a follow up ticket to convert later.

awx/ui/src/screens/Instances/InstanceList/InstanceList.js
kialam, line 145: <3

kialam (approved):

PR #12626, opened 2022-08-04
AAP jobs get stuck  and do not progress

jangel97:
Hello there,

I was having issues with some jobs not progressing in Ansible Automation Platform. After checking with Red Hat support team, they said 
```
Looking at the shared awx-dbshell_last-seen.txt and also in the sosreport command awx-manage list_instances we can see a huge discrepancy in the nodes timestamp.
It looks like our list_instances.py which is performing the heartbeat check is looking for the modified timestamp and not last_seen.
```
In order to tackle this Bug we believe that heartbeat in `awx list_instances` command should look for the last_seen attribute of the node and not the `modified`. This was a Breaking Change

Please, let me know if this makes sense.

Thank you!


Zokormazo (approved):

AlanCoding (approved):

AlanCoding:
Can you do the `git commit --signoff` before we merge this? thank!

https://github.com/ansible/awx/blob/eec4f8dcc27428cd1aeb1c251d43052a3ea780e8/DCO_1_1.md

PR #12629, opened 2022-08-05
Task manager refactor

fosterseth:
#### Split TaskManager into 3 separate managers that can run in parallel
- DependencyManager processes tasks that might need dependencies spawned, and flips `dependencies_processed` to True after resolving any dependencies
- TaskManager looks at pending tasks with `dependencies_processed` True. It is responsible for fitting the task to a control instance and execution instance, and submitting the task to the dispatcher.
- WorkflowManager looks only at running workflows and decides whether it can spawn new jobs

#### Save progress before timing out manager
- After 5 minutes, the task manager PID will be killed. However, there is now a grace period that allows the managers to bail out and save whatever progress it has made before it is killed.

#### Cache preferred_instance_groups and task_impact
- These two fields on UnifiedJob are fairly expensive, and were being calculated repeatedly across task manager runs. Now, these fields are calculated and set in `.create_unified_job`, thus no extra DB hits accrue when accessing these fields.

the parent issue tracker can be found here https://github.com/ansible/tower/issues/5943

- [x] fix up unit / functional tests
  - PR https://github.com/ansible/awx/pull/12534 
- [x] fix up integration / collections tests
- [x] green on test suite
- [x] performance and scale team results
- [x] merge it

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.3.1.dev72+g9c74678c61
```


awx/main/dispatch/pool.py
```diff
@@ -401,13 +401,15 @@ def cleanup(self):
                 # the task manager to never do more work
                 current_task = w.current_task
                 if current_task and isinstance(current_task, dict):
-                    if current_task.get('task', '').endswith('tasks.run_task_manager'):
+                    endings = ['tasks.task_manager', 'tasks.dependency_manager', 'tasks.workflow_manager']
+                    current_task_name = current_task.get('task', '')
+                    if any([current_task_name.endswith(e) for e in endings]):
```
jbradberry: FYI, nesting a list comprehension in like this isn't necessary.  `any(current_task_name.endswith(e) for e in endings)` will work just fine.

awx/main/models/inventory.py
```diff
@@ -337,9 +337,14 @@ def update_computed_fields(self):
         else:
             active_inventory_sources = self.inventory_sources.filter(source__in=CLOUD_INVENTORY_SOURCES)
         failed_inventory_sources = active_inventory_sources.filter(last_job_failed=True)
+        total_hosts = active_hosts.count()
+        if total_hosts != self.total_hosts:
```
jbradberry: or just `update_task_impact = (total_hosts != self.total_hosts)`

awx/main/scheduler/task_manager.py
```diff
@@ -228,99 +162,106 @@ def process_finished_workflow_jobs(self, workflow_jobs):
...
+                            logger.info(
+                                'Refusing to start recursive workflow-in-workflow id={}, wfjt={}, ancestors={}'.format(
+                                    job.id, spawn_node.unified_job_template.pk, [wa.pk for wa in workflow_ancestors]
+                                )
+                            )
+                            display_list = [spawn_node.unified_job_template] + workflow_ancestors
+                            job.job_explanation = gettext_noop(
+                                "Workflow Job spawned from workflow could not start because it "
+                                "would result in recursion (spawn order, most recent first: {})"
+                            ).format(', '.join(['<{}>'.format(tmp) for tmp in display_list]))
```
jbradberry: the `.join()` call also does not need the parameter to be a list comprehension.  An implicit generator expression (i.e. just remove the `[` and `]`) will work just fine, like with the `any()` call I mentioned previously.

awx/main/scheduler/task_manager.py
```diff
@@ -228,99 +162,106 @@ def process_finished_workflow_jobs(self, workflow_jobs):
...
+                        else:
+                            logger.debug(
+                                'Starting workflow-in-workflow id={}, wfjt={}, ancestors={}'.format(
+                                    job.id, spawn_node.unified_job_template.pk, [wa.pk for wa in workflow_ancestors]
+                                )
+                            )
+                    if not job._resources_sufficient_for_launch():
+                        can_start = False
+                        job.job_explanation = gettext_noop(
+                            "Job spawned from workflow could not start because it " "was missing a related resource such as project or inventory"
```
jbradberry: Looks like black reformatted this string from a multi-line string.

awx/main/analytics/subsystem_metrics.py
```diff
@@ -184,19 +184,28 @@ def __init__(self, auto_pipe_execute=False, instance_name=None):
...
             IntM('task_manager_schedule_calls', 'Number of calls to task manager schedule'),
             SetFloatM('task_manager_recorded_timestamp', 'Unix timestamp when metrics were last recorded'),
             SetIntM('task_manager_tasks_started', 'Number of tasks started'),
             SetIntM('task_manager_running_processed', 'Number of running tasks processed'),
             SetIntM('task_manager_pending_processed', 'Number of pending tasks processed'),
             SetIntM('task_manager_tasks_blocked', 'Number of tasks blocked from running'),
+            SetFloatM('dependency_manager_get_tasks_seconds', 'Time spent in loading tasks from db'),
+            SetFloatM('dependency_manager_generate_dependencies_seconds', 'Time spent generating dependencies for pending tasks'),
+            SetFloatM('dependency_manager__schedule_seconds', 'Time spent in running the entire _schedule'),
+            IntM('dependency_manager_schedule_calls', 'Number of calls to task manager schedule'),
```
AlanCoding: This help text doesn't clarify the metric sufficiently for me. Does this count the total number of times the dependency manager acquired the lock on ran on the local instance?

awx/main/analytics/subsystem_metrics.py
```diff
@@ -184,19 +184,28 @@ def __init__(self, auto_pipe_execute=False, instance_name=None):
...
+            SetFloatM('dependency_manager_recorded_timestamp', 'Unix timestamp when metrics were last recorded'),
+            SetIntM('dependency_manager_pending_processed', 'Number of pending tasks processed'),
+            SetFloatM('workflow_manager__schedule_seconds', 'Time spent in running the entire _schedule'),
+            IntM('workflow_manager_schedule_calls', 'Number of calls to task manager schedule'),
+            SetFloatM('workflow_manager_recorded_timestamp', 'Unix timestamp when metrics were last recorded'),
+            SetFloatM('workflow_manager_spawn_workflow_graph_jobs_seconds', 'Time spent spawning workflow jobs'),
+            SetFloatM('workflow_manager_get_tasks_seconds', 'Time spent in loading tasks from db'),
```
AlanCoding: maybe say "time spent in loading workflow jobs from db"

awx/main/models/unified_jobs.py
```diff
@@ -693,6 +699,17 @@ class Meta:
...
+    preferred_instance_groups_cache = models.JSONField(
+        blank=True,
+        null=True,
+        default=None,
+        editable=False,
+        help_text=_("A cached list with pk values from preferred instance groups."),
+    )
+    task_impact = models.PositiveIntegerField(
+        default=0,
+        editable=False,
```
AlanCoding: I don't see why we don't add help_text to this (consistent with other new field). If you are not sure if it's taking effect, see GET listed in the OPTIONS data. It's mundane work, but those who work with our API will appreciate it.

awx/main/analytics/subsystem_metrics.py
fosterseth: yeah these are calls to `_schedule` specifically, not `schedule`, so only after the lock

fosterseth: oh nvm, I realize the metric does not emphasize _schedule, let me make this more clear

awx/main/analytics/subsystem_metrics.py
fosterseth: fixed

awx/main/dispatch/pool.py
fosterseth: nice catch, fixed

awx/main/models/inventory.py
fosterseth: fixed

awx/main/scheduler/task_manager.py
```diff
@@ -481,16 +428,166 @@ def generate_dependencies(self, undeped_tasks):
 
         return created_dependencies
 
+    def process_tasks(self):
+        deps = self.generate_dependencies(self.all_tasks)
+        self.generate_dependencies(deps)
+        self.subsystem_metrics.inc(f"{self.prefix}_pending_processed", len(self.all_tasks) + len(deps))
+
+    @timeit
```
AlanCoding, line 436: I'm still really bothered that we are putting timings on `_schedule` and not `schedule`, because this misses timings of on_commit actions, which we know are a constant point of failure. Sure, this could be handled by filing a followup issue, but I really don't want it to get lost or slip into the next release, because it undermines the very point of the task manager metrics. We can still hit the _dispatcher_ timeout limits, with no forewarning by the metrics. Until we fix this, the metrics are not really to be believed.

just the other day, I was able to see this by a new mechanism:

```
tools_awx_1 | 2022-08-08 17:40:42,465 ERROR    [252b9dcc] awx.main.dispatch Worker failed to run task awx.main.scheduler.tasks.run_task_manager(*[], **{}
tools_awx_1 |   File "/awx_devel/awx/main/dispatch/worker/task.py", line 102, in perform_work
tools_awx_1 |     result = self.run_callable(body)
tools_awx_1 |   File "/awx_devel/awx/main/dispatch/worker/task.py", line 77, in run_callable
tools_awx_1 |     return _call(*args, **kwargs)
tools_awx_1 |   File "/awx_devel/awx/main/scheduler/tasks.py", line 15, in run_task_manager
tools_awx_1 |     TaskManager().schedule()
tools_awx_1 |   File "/awx_devel/awx/main/scheduler/task_manager.py", line 657, in schedule
tools_awx_1 |     start_time = time.time()
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/db/transaction.py", line 290, in __exit__
tools_awx_1 |     connection.set_autocommit(True)
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/db/backends/base/base.py", line 420, in set_autocommit
tools_awx_1 |     self.run_and_clear_commit_hooks()
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/db/backends/base/base.py", line 653, in run_and_clear_commit_hooks
tools_awx_1 |     func()
tools_awx_1 |   File "/awx_devel/awx/main/scheduler/task_manager.py", line 294, in post_commit
tools_awx_1 |     task_cls.apply_async(
tools_awx_1 |   File "/awx_devel/awx/main/dispatch/publish.py", line 87, in apply_async
tools_awx_1 |     with pg_bus_conn() as conn:
tools_awx_1 |   File "/usr/lib64/python3.9/contextlib.py", line 119, in __enter__
tools_awx_1 |     return next(self.gen)
tools_awx_1 |   File "/awx_devel/awx/main/dispatch/__init__.py", line 50, in pg_bus_conn
tools_awx_1 |     conn = psycopg2.connect(dbname=conf['NAME'], host=conf['HOST'], user=conf['USER'], password=conf['PASSWORD'], port=conf['PORT'], **conf.get("OPTIONS", {}))
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/psycopg2/__init__.py", line 126, in connect
tools_awx_1 |     conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
tools_awx_1 | psycopg2.OperationalError: FATAL:  sorry, too many clients already
```

This shows that the `on_commit` logic is even _more_ vulnerable than thought, because it requires establishing a new database connection. When approaching connection limits, this guarantees we leave jobs stuck in waiting. That may or may not be exasperated failing to close unused connections, not sure yet.

AlanCoding (approved):

PR #12638, opened 2022-08-09
Adds a github workflow that updates dependabot prs 

AlexSCorey:
##### SUMMARY
We have a github workflow [file ](https://github.com/ansible/awx/blob/devel/.github/workflows/pr_body_check.yml) that checks to see if all prs have 1 of a set of strings in them to automate release notes.  Dependabot-created-prs do not have any of those strings, and the dependabot file cannot be configured for custom pr body.  Ergo this PR. This workflow checks for workflows with labels ('dependencies' and 'component:ui') which are automatically added to dependabot prs and updates that body of those prs to add ` Bug, Docs Fix or other nominal change` so it will pass that aformentioned workflow

##### ISSUE TYPE 
 - New or Enhanced Feature

##### COMPONENT NAME
 - Other

##### AWX VERSION

### DIFF
--- a/awx_collection/plugins/module_utils/controller_api.py
+++ b/awx_collection/plugins/module_utils/controller_api.py
@@ -12,7 +12,7 @@ from ansible.module_utils.six.moves.http_cookiejar import CookieJar
 from ansible.module_utils.six.moves.urllib.parse import urlparse, urlencode
 from ansible.module_utils.six.moves.configparser import ConfigParser, NoOptionError
 from distutils.version import LooseVersion as Version
-from socket import gethostbyname
+from socket import getaddrinfo, IPPROTO_TCP
 import time
 import re
 from json import loads, dumps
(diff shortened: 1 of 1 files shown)


.github/workflows/update_dependabot_prs.yml
```diff
@@ -0,0 +1,25 @@
+---
+name: Customize PR body 1
+on:
+  pull_request:
+    types: [labeled, opened, reopened]
+
+jobs:
+  pr-check:
+    name: Scan PR description for semantic versioning keywords
```
shanemcd: Copy pasta?

.github/workflows/update_dependabot_prs.yml
```diff
@@ -0,0 +1,25 @@
+---
+name: Customize PR body 1
```
shanemcd: Can we make this something more descriptive?

.github/workflows/update_dependabot_prs.yml
AlexSCorey: You caught me

.github/workflows/update_dependabot_prs.yml
AlexSCorey: yes

.github/workflows/update_dependabot_prs.yml
```diff
@@ -0,0 +1,25 @@
...
+
+      - name: Update PR Body
+        env:
+            GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+            OWNER: ${{ github.repository_owner }}
+            REPO: ${{ github.event.repository.name }}
+            BRANCH: ${{github.event.pull_request.head.ref}}
+        run: |
+          gh pr checkout ${{ env.BRANCH }}
+          gh pr edit --body "Bug, Docs Fix or other nominal change"
```
john-westcott-iv: It looks like this is just going to replace the body per https://cli.github.com/manual/gh_pr_edit

But dependabot seems to put some potentially useful stuff in the body i.e. https://github.com/ansible/awx/pull/12606

Should we see if we can do like a `gh pr view` to get the body and then append our string to it and set the new body as the result?

PR #12642, opened 2022-08-10
fix use of distinct on query that UI uses

kdelee:

When on the screen in the UI that loads the job events, the ui includes
a filter to exclude job events where stdout = ''. Because this is a
TextField and was not in the allow list, we were applying DISTINCT to
the query. This made it very unperformant for large jobs, especially
on the query that gets the count and cannot put a LIMIT on the query.

Also correctly prefetch the related job_template data on the view to
cut down the number of queries we make from around 50 to under 10.

We need to analyze other similar views for other prefetch type
optimizations we should make.

going to build this so @jainnikhil30 can take a look and see if it fixes the bug he was reproducing with jobs with large amounts of events not loading in the UI.

### DIFF
--- a/awx/api/filters.py
+++ b/awx/api/filters.py
@@ -157,7 +157,7 @@ class FieldLookupBackend(BaseFilterBackend):
 
     # A list of fields that we know can be filtered on without the possiblity
     # of introducing duplicates
-    NO_DUPLICATES_ALLOW_LIST = (CharField, IntegerField, BooleanField)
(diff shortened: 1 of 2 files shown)


AlanCoding (approved):

kdelee:
confirmed with @akus062381 that the UI test result is getting fixed elsewhere, unrelated to this PR

kdelee:
YOLO and Nikhil's testing looking good, going to merge this now

PR #12649, opened 2022-08-12
Added more context to subscription details and rearrange the order of some of the fields

mabashian:
##### SUMMARY
Adds static help text to subscription details depending on status.  Moved a number of compliance related fields up to the beginning of the details view.  See the original issue for more details.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui/src/screens/Setting/Subscription/SubscriptionDetail/SubscriptionDetail.js
+++ b/awx/ui/src/screens/Setting/Subscription/SubscriptionDetail/SubscriptionDetail.js
@@ -1,8 +1,13 @@
 import React from 'react';
 import { Link } from 'react-router-dom';
-
 import { t, Trans } from '@lingui/macro';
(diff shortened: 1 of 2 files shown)


keithjgrant:
I'm curious what the thinking is on using static help text here when we use Popovers in all other Details. (There is certainly an accessibility argument for using static text everywhere.) `<Detail>` has a `helpText` prop that uses a Popover... I'm wondering if we should use that

mabashian:
@keithjgrant - Yea I'll defer to @ffirg but I think the idea is to put something right in the users face.  If we put this in a tooltip/popover it might be too subtle.

AlexSCorey (approved):
This looks good to me.

AlexSCorey:
Is there an issue tracking this work?

InnocentK (approved):
Looks good! (ignore the failing tests above--those are unrelated to these changes)

PR #12650, opened 2022-08-12
Fixed a bug where the initial form value of verbosity isn't respected

### LINKED ISSUE #12575: Attempting to run an adhoc command results in error

### Bug Summary

The adhoc modal appears to be doing something strange with a verbosity level of 0. When attempting to launch with verbosity `0 (Normal)` another modal pops up complaining about an error. Other verbosity levels work fine. 

When selecting level 0 the POST to /api/v2/inventories/X/ad_hoc_commands has the full label `0 (Normal)`:
image

While when selecting others we just pass the numeric value:
image

I also tried selecting something other than 0 and then back to 0 but still the full string was passed.

### AWX version

latest

- [X] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

N/A

### Modifications

no

### Ansible version

N/A

### Operating system

N/A

### Web browser

Chrome

### Steps to reproduce

Create an inventory, add a host.
Run an ad-hoc command against the host (ping module will work).
Select EE, select cred, launch.
You will then get the error:
image

### Expected results

Ad Hoc should launch properly after UI passed the value 0 to the API.

### Actual results

image

### Additional information

_No response_

matburt:
##### SUMMARY

It looks like https://github.com/ansible/awx/pull/12378/commits/ef8d4e73ae08b6ea10c495db28d9ede0ddb8a114 from PR https://github.com/ansible/awx/pull/12378 removed this reference to `.value`. I don't fully understand the implications of the commit but putting this value reference back in allows the form to work in it's default state again.  This would seem to fix https://github.com/ansible/awx/issues/12575

##### ISSUE TYPE

##### COMPONENT NAME

 - UI

##### AWX VERSION

```
awx: 21.4.1.dev50+g1ae1da3f9c
```

##### ADDITIONAL INFORMATION

This is currently broken on the latest AWX release 

```

```

### DIFF
--- a/awx/ui/src/components/AdHocCommands/AdHocCommandsWizard.js
+++ b/awx/ui/src/components/AdHocCommands/AdHocCommandsWizard.js
@@ -2,8 +2,6 @@ import React from 'react';
 import { t } from '@lingui/macro';
 import { withFormik, useFormikContext } from 'formik';
 import PropTypes from 'prop-types';
-
-import { VERBOSITY } from 'components/VerbositySelectField';
 import Wizard from '../Wizard';
 import useAdHocLaunchSteps from './useAdHocLaunchSteps';
 
@@ -62,7 +60,7 @@ const FormikApp = withFormik({
       limit: adHocItemStrings || 'all',
       credentials: [],
       module_args: '',
-      verbosity: VERBOSITY()[0].value,
+      verbosity: 0,
       forks: 0,
       diff_mode: false,
       become_enabled: '',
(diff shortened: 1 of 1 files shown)


mabashian:
Fix looks good, I'll address the test failure

mabashian:
Under the hood, verbosity is tracked by an integer.  0 corresponds to 0 (Normal), 1 corresponds to 1 (Verbose) and so on...

For adhoc jobs we always default to 0 (Normal).  @matburt's original fix worked because `VERBOSITY()[0].value` was evaluating as undefined (same as 0) but the unit tests caught that.  Since we know the default initial value is 0 lets just set it to 0.

mabashian (approved):

akus062381 (approved):

PR #12653, opened 2022-08-15
Fix job canceling when getting stuck reading or writing receptor socket

shanemcd:
##### SUMMARY
On a recent customer case they reported not being able to cancel a running job. After looking at the code, there were 2 additional places (`submit_work` and `get_work_results`) where it is possible to get stuck on a read or write with the receptor socket. This patch introduces a new pattern that will allow us to eject whenever we might be stuck.

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/tasks/receptor.py
+++ b/awx/main/tasks/receptor.py
@@ -290,7 +290,7 @@ class AWXReceptorJob:
         # This ThreadPoolExecutor runs for the duration of the job.
         # The cancel_func pattern is intended to guard againsnt any situation where we may
         # end up stuck while reading or writing from the receptor socket.
-        with concurrent.futures.ThreadPoolExecutor(max_workers=3) as executor:
+        with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
             # Launch transmitter and submitter at the same time so the payload
             # can be read by receptor while it's still being written.
             transmitter_future = executor.submit(self.transmitter, payload_writer)
@@ -310,8 +310,7 @@ class AWXReceptorJob:
                 self.receptor_ctl._sockfile.close()
 
             # This will return either when the job is canceled, or when the transmitter and submitter futures return.
-            cancel_watcher_future = executor.submit(self.cancel_watcher, [transmitter_future, submitter_future], cancel_func)
-            res = cancel_watcher_future.result()
+            res = self.cancel_watcher([transmitter_future, submitter_future], cancel_func)
             if res and res.status == "canceled":
                 return res, None
 
@@ -328,9 +327,8 @@ class AWXReceptorJob:
                 self.receptor_ctl._socket.close()
                 self.receptor_ctl._sockfile.close()
 
-            cancel_watcher_future = executor.submit(self.cancel_watcher, [work_results_future], cancel_func)
+            res = self.cancel_watcher([work_results_future], cancel_func)
             # This will return either when the job is canceled, or when the worker_results future returns.
-            res = cancel_watcher_future.result()
             if res and res.status == "canceled":
                 return res, unit_id
 
@@ -343,7 +341,7 @@ class AWXReceptorJob:
                 resultsock.shutdown(socket.SHUT_RDWR)
                 resultfile.close()
 
-            cancel_watcher_future = executor.submit(self.cancel_watcher, [processor_future], cancel_func)
+            res = self.cancel_watcher([processor_future], cancel_func)
             # This will return either when the job is canceled, or when the processor future returns.
             if res and res.status == "canceled":
                 return res, unit_id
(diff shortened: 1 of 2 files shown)


AlanCoding:
In that case, was it confirmed that the job process was really running? It would be good to get strace information to find out what specifically the control process is hanging on. I agree with your assessments and this is adding in good general defense to code points that might result in hanging, but there are other scenarios for jobs stuck in running.

awx/main/tasks/receptor.py
```diff
@@ -266,139 +267,119 @@ def __init__(self, task, runner_params=None):
...
+    def transmitter(self, _socket):
         try:
             ansible_runner.interface.run(streamer='transmit', _output=_socket.makefile('wb'), **self.runner_params)
         finally:
             # Socket must be shutdown here, or the reader will hang forever.
             _socket.shutdown(socket.SHUT_WR)
 
+    @cleanup_new_process
+    def submitter(self, payload_reader):
+        # Prepare the submit_work kwargs before creating threads, because references to settings are not thread-safe
```
AlanCoding, line 367: comment doesn't work, because you are already in a thread at this point.

awx/main/tasks/receptor.py
```diff
@@ -266,139 +267,119 @@ def __init__(self, task, runner_params=None):
...
+                self.handle_error(unit_id)
+
             return res
         finally:
             # Make sure to always release the work unit if we established it
-            if self.unit_id is not None and settings.RECEPTOR_RELEASE_WORK:
+            if unit_id is not None and settings.RECEPTOR_RELEASE_WORK:
                 try:
-                    receptor_ctl.simple_command(f"work release {self.unit_id}")
+                    self.receptor_ctl.simple_command(f"work release {unit_id}")
```
AlanCoding, line 283: If the concept is the, generally, receptor commands are not safe from hangs, then I expect that would apply to `receptorctl work release` as well. I believe I've had it hang here before.

awx/main/tasks/receptor.py
```diff
@@ -266,139 +267,119 @@ def __init__(self, task, runner_params=None):
...
+                self.receptor_ctl._socket.close()
+                self.receptor_ctl._sockfile.close()
+
+            cancel_watcher_future = executor.submit(self.cancel_watcher, [work_results_future], cancel_func)
+            # This will return either when the job is canceled, or when the worker_results future returns.
+            res = cancel_watcher_future.result()
+            if res and res.status == "canceled":
+                return res, unit_id
+
+            resultsock, resultfile = work_results_future.result()
```
AlanCoding, line 335: All `work_results` does is return the socket. It seems somewhat extraordinary that _it_ would be hanging.

https://github.com/ansible/receptor/blob/ba3ed4532509e9d92a4d0bd89c0a284ee13f58a0/receptorctl/receptorctl/socket_interface.py#L248

It does read from the general communication socket to receptor, so it absolutely could hang.

awx/main/tasks/receptor.py
```diff
@@ -266,139 +267,119 @@ def __init__(self, task, runner_params=None):
...
+
+            def cancel_func():
+                payload_writer.shutdown(socket.SHUT_RDWR)
+                payload_writer.close()
+                payload_reader.shutdown(socket.SHUT_RDWR)
+                payload_reader.close()
+                payload_reader_file.close()
+                self.receptor_ctl._socket.shutdown(socket.SHUT_RDWR)
+                self.receptor_ctl._socket.close()
+                self.receptor_ctl._sockfile.close()
```
AlanCoding, line 310: Somewhat of a tangent - I have many thought it would be better to call `self.receptor_ctl.close()`, to replace the 3 lines above here. It's written defensively enough, and would be more clear / DRY.

fosterseth: `self.receptor_ctl._socket.shutdown(socket.SHUT_RDWR)` is not call in receptorctl's close, so we could keep that line. But the other 2 lines could be replaced

awx/main/tasks/receptor.py
```diff
@@ -266,139 +267,119 @@ def __init__(self, task, runner_params=None):
...
-                    logger.exception(f"Error releasing work unit {self.unit_id}.")
+                    logger.exception(f"Error releasing work unit {unit_id}.")
+
+    def _run(self):
+        payload_writer, payload_reader = socket.socketpair()
+
+        # This ThreadPoolExecutor runs for the duration of the job.
+        # The cancel_func pattern is intended to guard againsnt any situation where we may
+        # end up stuck while reading or writing from the receptor socket.
+        with concurrent.futures.ThreadPoolExecutor(max_workers=3) as executor:
```
AlanCoding: I checked out your branch and ran graphs

Screenshot from 2022-08-16 08-32-44

This is for running 5 jobs. The increase in database connections is 30-10 = 20. Then 20 / 5 = 4 threads per job, as expected. This is because `max_workers=3` doesn't include the main thread, so 4 threads in total, and without connection pooling between threads, a new connection is created for each. Currently it's 3 per job, and particularly if we're backporting, increasing this number by 1 will be disastrous for some large (or poorly balanced) deployments.

Really, it should be either 1 or 0. But for now, I'm going to pull out some of my tricks to suggest to mitigate what you're doing here, because watching for cancels over the transmit phase is something I know we were lacking.

AlanCoding:
Link https://github.com/shanemcd/awx/pull/75, I may still have more changes I want to see after this.

At this stage I might also try to run some tests.

Issue #12682, opened 2022-08-18
Grafana Notification => ValueError: invalid literal for int() with base 10: ''


manuel-held:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

Administration => Notification - Type Grafana.

Test failed when "ID of the dashboard (optional)" or/and "ID of the panel (optional)" is empty. 

`2022-08-18 10:11:57,916 ERROR    [9170e756] awx.main.tasks.system Send Notification Failed invalid literal for int() with base 10: ''
Traceback (most recent call last):
  File "/awx_devel/awx/main/tasks/system.py", line 290, in send_notifications
    sent = notification.notification_template.send(notification.subject, notification.body)
  File "/awx_devel/awx/main/models/notifications.py", line 182, in send
    backend_obj = self.notification_class(**notification_configuration)
  File "/awx_devel/awx/main/notifications/grafana_backend.py", line 58, in __init__
    self.panelId = int(panelId) if panelId is not None else None
ValueError: invalid literal for int() with base 10: ''`

and

`2022-08-18 10:09:46,302 ERROR    [c7de9089] awx.main.tasks.system Send Notification Failed invalid literal for int() with base 10: ''
Traceback (most recent call last):
  File "/awx_devel/awx/main/tasks/system.py", line 290, in send_notifications
    sent = notification.notification_template.send(notification.subject, notification.body)
  File "/awx_devel/awx/main/models/notifications.py", line 182, in send
    backend_obj = self.notification_class(**notification_configuration)
  File "/awx_devel/awx/main/notifications/grafana_backend.py", line 57, in __init__
    self.dashboardId = int(dashboardId) if dashboardId is not None else None
ValueError: invalid literal for int() with base 10: ''`

The reason i think is that '' is not None so '' is passed to int() and triggers the Error

AWX Version: Ansible AWX 21.4.0
Grafana Version: v9.0.7

### AWX version

21.4.0

### Select the relevant components

- [X] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

docker development environment

### Modifications

no

### Ansible version

2.10.8

### Operating system

Ubuntu 22.04.1 LTS

[...]

PR #12684, opened 2022-08-18
Fixes invalid literal for int() with base 10 in the Grafana Notification 

manuel-held:
##### ISSUE TYPE

 - Fix Bug #12682

##### COMPONENT NAME

 - UI

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/notifications/grafana_backend.py
+++ b/awx/main/notifications/grafana_backend.py
@@ -53,8 +53,8 @@ class GrafanaBackend(AWXBaseEmailBackend, CustomNotificationBase):
     ):
         super(GrafanaBackend, self).__init__(fail_silently=fail_silently)
         self.grafana_key = grafana_key
-        self.dashboardId = int(dashboardId) if dashboardId is not None else None
-        self.panelId = int(panelId) if panelId is not None else None
+        self.dashboardId = int(dashboardId) if dashboardId is not None and len(dashboardId) else None
+        self.panelId = int(panelId) if panelId is not None and len(panelId) else None
         self.annotation_tags = annotation_tags if annotation_tags is not None else []
         self.grafana_no_verify_ssl = grafana_no_verify_ssl
         self.isRegion = isRegion


john-westcott-iv:
Hello, thank you so much for taking the time to contribute to AWX. You have a failing test, CI/api-test, that we need to get working before we can proceed with accepting this PR. Please let us know if you need additional help to get this test in working order.

awx/main/notifications/grafana_backend.py
```diff
@@ -54,8 +54,8 @@ def __init__(
     ):
         super(GrafanaBackend, self).__init__(fail_silently=fail_silently)
         self.grafana_key = grafana_key
-        self.dashboardId = int(dashboardId) if dashboardId is not None else None
-        self.panelId = int(panelId) if panelId is not None else None
+        self.dashboardId = int(dashboardId) if dashboardId is not None and len(dashboardId) else None
```
djyasin, line 56: We would prefer a type check rather than using len() here to ensure the cast to int will be successful.

AlanCoding: I followed the reproducer in the issue, and it is called out that the value is empty string, `""`. The more standard python way of handling that would be

```diff
diff --git a/awx/main/notifications/grafana_backend.py b/awx/main/notifications/grafana_backend.py
index 0d7dff9e05..7879449526 100644
--- a/awx/main/notifications/grafana_backend.py
+++ b/awx/main/notifications/grafana_backend.py
@@ -54,8 +54,8 @@ class GrafanaBackend(AWXBaseEmailBackend, CustomNotificationBase):
     ):
         super(GrafanaBackend, self).__init__(fail_silently=fail_silently)
         self.grafana_key = grafana_key
-        self.dashboardId = int(dashboardId) if dashboardId is not None else None
-        self.panelId = int(panelId) if panelId is not None else None
+        self.dashboardId = int(dashboardId) if dashboardId else None
+        self.panelId = int(panelId) if panelId else None
         self.annotation_tags = annotation_tags if annotation_tags is not None else []
         self.grafana_no_verify_ssl = grafana_no_verify_ssl
         self.isRegion = isRegion
```

So you use the value as if it were a boolean, it handles false-y values of all kinds, which includes None and empty string.

dmzoneill:
This is fixed in a newer PR

Issue #12556, opened 2022-07-20
Job output scroll position is off after events refresh


keithjgrant:
##### ISSUE TYPE
 - Bug or Docs Fix

##### COMPONENT NAME
 - UI

##### SUMMARY
When a job finishes running, the events on the job output page are refreshed from the API to ensure displayed output is fully in sync. However, after this happens, the output generally doesn't scroll back to the same location the user was viewing beforehand.

We can either:
a) keep better track of the scroll position prior to the refresh and ensure we scroll back to that position and stay there as events load, or
b) re-think this full refresh idea, and only fetch specific events that may be missing in the browser view

Option _b)_ could probably be done in conjunction with #12543

InnocentK:
Resolved via option _b)_ (along with https://github.com/ansible/awx/issues/12543) by the work completed in https://github.com/ansible/awx/pull/12689

Issue #12543, opened 2022-07-19
UI should poll job endpoint to determine exactly when events are done processing


mabashian:
Peeling this off of https://github.com/ansible/awx/issues/12103

After some discussion, this feels like the best path forward in the short term.  Longer term it would be ideal if the UI got an event via the websocket that indicates that we're done processing events but this requires more discussion.

The work here should include triggering a series of GET requests to the job endpoint looking at the `event_processing_finished` field.  One idea to avoid dos'ing the backend would be to iteratively increase the interval that we poll this endpoint.  Maybe start out doing it every few seconds and then extend it out to every 30 seconds after some number of iterations.

keithjgrant:
as I’ve been thinking about it, we may just want to change the strategy. It really only refreshes job events in case the websocket events skipped some (and to enable expand/collapse if relevant) (edited) 

so we probably only need to poll for a minute or two, and if it doesn’t finish processing, just leave the UI as is… After a long delay, I don’t think the user will want to see output disappear and then reload into the screen “randomly”

it might even be worth simply displaying a message instead. “AWX has finished processing events for this job. Click here to refresh”

keithjgrant:
see also #12556 — if we re-think the way we refresh events at job completion, we can probably knock out both these issues together

mabashian:
@keithjgrant wondering if https://github.com/ansible/awx/pull/12555 addresses this issue as well.  I don't see this issue listed in the PR but they seem sorta kinda related.

InnocentK:
Resolved by https://github.com/ansible/awx/pull/12689

PR #12689, opened 2022-08-18
poll for events processing completion

keithjgrant:
##### SUMMARY
Polls API after job completion to detect when job events processing is complete. When completion is detected, a banner is added informing the user and providing them with a button to reload the job output:

Screen Shot 2022-08-18 at 9 01 07 AM

Addresses #12543, #12556

Note: Previously, the job output refreshed automatically after job completion. I have disabled this, because it feels more intrusive to the user experience if the output disappears and refreshes without warning. Additionally, I was not able to get the scroll position to line up correctly with the previous position after refresh (as live-run job events are stored differently in the UI compared to events loaded after job completion).

##### ISSUE TYPE

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui/src/screens/Job/JobOutput/JobOutput.js
+++ b/awx/ui/src/screens/Job/JobOutput/JobOutput.js
@@ -10,7 +10,7 @@ import {
   InfiniteLoader,
   List,
 } from 'react-virtualized';
-import { Button } from '@patternfly/react-core';
+import { Button, Alert } from '@patternfly/react-core';
 
 import AlertModal from 'components/AlertModal';
 import { CardBody as _CardBody } from 'components/Card';
@@ -99,6 +99,7 @@ function JobOutput({ job, eventRelatedSearchableKeys, eventSearchableKeys }) {
   const scrollHeight = useRef(0);
   const history = useHistory();
   const eventByUuidRequests = useRef([]);
+  const eventsProcessedDelay = useRef(250);
 
   const fetchEventByUuid = async (uuid) => {
     let promise = eventByUuidRequests.current[uuid];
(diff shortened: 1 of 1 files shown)


awx/ui/src/screens/Job/JobOutput/JobOutput.js
```diff
@@ -99,6 +99,7 @@ function JobOutput({ job, eventRelatedSearchableKeys, eventSearchableKeys }) {
   const scrollHeight = useRef(0);
   const history = useHistory();
   const eventByUuidRequests = useRef([]);
+  const eventsProcessedDelay = useRef(250);
```
AlexSCorey, line 102: why not just use a regular `const` or `let` here instead of a useRef hook?

awx/ui/src/screens/Job/JobOutput/JobOutput.js
```diff
@@ -196,14 +198,28 @@ function JobOutput({ job, eventRelatedSearchableKeys, eventSearchableKeys }) {
     rebuildEventsTree();
   }, [isFlatMode]); // eslint-disable-line react-hooks/exhaustive-deps
 
+  const pollForEventsProcessed = useCallback(async () => {
```
AlexSCorey, line 201: Does this function poll at increasing longer intervals?  Like it starts polling at 250, and then doubles its interval each time until it gets to 1000 * 60 * 5 (5 minutes)?

keithjgrant: yes, exactly

awx/ui/src/screens/Job/JobOutput/JobOutput.js
keithjgrant: Since we edit this timeout in the `pollForEventsProcessed`, we need the updated value to persist in a ref

kialam (approved):
LGTM. I pulled the changes down and tested locally. Nicely done.

InnocentK (approved):
The above failing tests are unrelated to these changes so this should be good to go

PR #12693, opened 2022-08-18
[content signing] 12687 remove enable field

djyasin:
##### SUMMARY

Based on the new functionality of the content signing process we determined we no longer needed the Signature_validation_enable field on the model and in the corresponding serializer. 

Related issue: #12523
##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```
4.3
```

##### ADDITIONAL INFORMATION

```
            "default_environment": null,
            "signature_validation_credential": null,
            "last_update_failed": false,
```


awx/main/migrations/0167_remove_project_signature_validation.py
```diff
@@ -0,0 +1,17 @@
...
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('main', '0166_task_manager_refactor'),
+    ]
+
+    operations = [
```
TheRealHaoLiu: Modify our original migration rather than generate new migration

awx/main/migrations/0165_added_contentsigning_to_project_model.py
```diff
@@ -37,4 +37,13 @@ class Migration(migrations.Migration):
             ),
         ),
         migrations.RunPython(setup_tower_managed_defaults),
+        migrations.RemoveField(
+            model_name='project',
+            name='signature_validation',
+        ),
```
TheRealHaoLiu: this field should never exist after this PR so not Remove just deleted these line

awx/main/migrations/0167_added_contentsigning_to_project_model.py
```diff
@@ -37,4 +37,13 @@ class Migration(migrations.Migration):
...
         migrations.RunPython(setup_tower_managed_defaults),
+        migrations.RemoveField(
+            model_name='project',
+            name='signature_validation',
+        ),
+        migrations.AlterField(
+            model_name='project',
+            name='signature_validation_credential',
+            field=models.ForeignKey(blank=True, default=None, null=True, on_delete=django.db.models.deletion.SET_NULL, to='main.credential'),
+        ),
```
TheRealHaoLiu: this doesn't look right. 

delete file and regenerate migration

awx_collection/plugins/modules/project.py
```diff
@@ -291,15 +285,14 @@ def main():
...
         signature_validation_credential=dict(type='str'),
     )
 
     # Create a module for ourselves
     module = ControllerAPIModule(
         argument_spec=argument_spec,
         required_if=[
-            ("signature_validation", True, ["signature_validation_credential"]),
+            (signature_validation_credential is not None),
         ],
```
TheRealHaoLiu: i think we just need to remove the whole required_if section

awx_collection/plugins/modules/project.py
```diff
@@ -169,12 +169,6 @@
       required: False
       default: 2
       type: float
-    signature_validation:
-      description:
-        - Whether to validate the signature during project update.
-      required: False
-      default: False
-      type: bool
     signature_validation_credential:
```
TheRealHaoLiu, line 172: description need to be changed 

no longer have a IF condition and also setting this will imply enabling signature validation

awx/main/models/projects.py
```diff
@@ -284,15 +284,9 @@ class Meta:
...
 
-    # indicate if content signature validation is enabled
-    signature_validation = models.BooleanField(
-        default=False,
-        help_text=_('Enable signature validation.'),
-    )
     # credential (keys) used to validate content signature
     signature_validation_credential = models.ForeignKey(
         'Credential',
-        related_name='%(class)ss_signature_validation',
```
TheRealHaoLiu: i guess this was not correct before and should still be
`related_name='%(class)ss`

awx_collection/plugins/modules/project.py
```diff
@@ -169,16 +169,10 @@
       required: False
       default: 2
       type: float
-    signature_validation:
```
relrod, line 172: I think line 390 needs to be dropped now too.

relrod (changes requested):

awx_collection/plugins/modules/project.py
djyasin: I believe this is resolved as well now!

awx_collection/plugins/modules/project.py
```diff
@@ -169,16 +169,10 @@
...
-      description:
-        - Whether to validate the signature during project update.
-      required: False
-      default: False
-      type: bool
     signature_validation_credential:
       description:
         - Name of the credential to use for signature validation.
-        - Required if C(signature_validation=True)
+        - If signature valdiation credential is provided, signature validation will be enabled.
```
relrod:
```suggestion
        - If signature validation credential is provided, signature validation will be enabled. 
```

PR #12694, opened 2022-08-18
Shortcut Instance.objects.me when possible

AlanCoding:
##### SUMMARY
The only time we actually need to get the instance record for the local node is when we're doing something with capacity. Otherwise, generally, services shouldn't require any information from that to function correctly. The only thing required is that the installer sets up the CLUSTER_HOST_ID setting, which happens much earlier and is much more static and reliable.

Cases of `Instance.objects.me().hostname` don't make any sense to me at all. So I want to look into whether there are any ill-effects from this, and if not, I don't see what is keeping us from using the simpler and less error-prone approach.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/analytics/subsystem_metrics.py
+++ b/awx/main/analytics/subsystem_metrics.py
@@ -166,11 +166,7 @@ class Metrics:
         elif settings.IS_TESTING():
             self.instance_name = "awx_testing"
         else:
-            try:
-                self.instance_name = Instance.objects.me().hostname
-            except Exception as e:
-                self.instance_name = settings.CLUSTER_HOST_ID
-                logger.info(f'Instance {self.instance_name} seems to be unregistered, error: {e}')
+            self.instance_name = Instance.objects.my_hostname()
 
         # metric name, help_text
         METRICSLIST = [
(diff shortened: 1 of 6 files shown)


shanemcd (approved):
🔥

AlanCoding:
I ran the testing for this, and it actually looks pretty good, but I will avoid merging it the Friday before I go on vacation.

jbradberry (approved):

jbradberry:
@AlanCoding since the mesh scaling work has merged, and I believe we made use of `.me()` in there, you might want to fold those changes into this PR as well.

AlanCoding:
I see it in `write_receptor_config`, yes. But the point of that method is that it takes the receptor config from the database and writes it. If the instance isn't present, then clearly you don't have a valid receptor configuration available in the database.

If someone later tries to really add that to the dispatcher startup... then I guess we would have a problem. Many functions of the dispatcher can proceed without access to receptor, so we may have some philosophical arguments to work out there. Anything related to capacity for running any types of jobs requires receptor... but if receptor is not available, then the dispatcher needs to still function in some other ways in order to report that it is not available to accept jobs (health check).

PR #12699, opened 2022-08-19
Remove need for settings.py during image build

shanemcd:
##### SUMMARY

I ran into this when doing https://github.com/ansible/awx-operator/pull/1026.

This avoids needing to duplicate that code ^ in `tools/ansible/roles/dockerfile/files/settings.py` (deleted completely in this PR).

##### ISSUE TYPE

##### COMPONENT NAME

 - API


awx/settings/defaults.py
```diff
@@ -1043,3 +1044,6 @@ def IS_TESTING(argv=None):
 # Number of seconds after a container group job finished time to wait
 # before the awx_k8s_reaper task will tear down the pods
 K8S_POD_REAPER_GRACE_PERIOD = 60
+
+# This is overridden downstream via /etc/tower/conf.d/cluster_host_id.py
+CLUSTER_HOST_ID = socket.gethostname()
```
shanemcd, line 1049: @AlanCoding Does this look ok to you? I'm asking because of https://github.com/ansible/awx/pull/11385

jbradberry (approved):

PR #12704, opened 2022-08-19
Consolidate and refactor logging configuration code

shanemcd:
##### SUMMARY
Once this merges, we can also delete the _new_ code that was added in https://github.com/ansible/awx-operator/pull/1026

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/settings/defaults.py
+++ b/awx/settings/defaults.py
@@ -821,85 +821,24 @@ LOGGING = {
         'dispatcher': {'format': '%(asctime)s %(levelname)-8s [%(guid)s] %(name)s PID:%(process)d %(message)s'},
         'job_lifecycle': {'()': 'awx.main.utils.formatters.JobLifeCycleFormatter'},
     },
+    # Extended below based on install scenario. You probably don't want to add something directly here.
+    # See 'handler_config' below.
     'handlers': {
         'console': {
             '()': 'logging.StreamHandler',
             'level': 'DEBUG',
-            'filters': ['require_debug_true_or_test', 'dynamic_level_filter', 'guid'],
+            'filters': ['dynamic_level_filter', 'guid'],
             'formatter': 'simple',
         },
         'null': {'class': 'logging.NullHandler'},
         'file': {'class': 'logging.NullHandler', 'formatter': 'simple'},
         'syslog': {'level': 'WARNING', 'filters': ['require_debug_false'], 'class': 'logging.NullHandler', 'formatter': 'simple'},
+        'inventory_import': {'level': 'DEBUG', 'class': 'logging.StreamHandler', 'formatter': 'timed_import'},
         'external_logger': {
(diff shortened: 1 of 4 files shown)


awx/settings/defaults.py
```diff
@@ -821,86 +821,25 @@ def IS_TESTING(argv=None):
...
-            'formatter': 'dispatcher',
-        },
-        'wsbroadcast': {
-            # don't define a level here, it's set by settings.LOG_AGGREGATOR_LEVEL
-            'class': 'logging.handlers.WatchedFileHandler',
-            'filters': ['require_debug_false', 'dynamic_level_filter', 'guid'],
-            'filename': os.path.join(LOG_ROOT, 'wsbroadcast.log'),
-            'formatter': 'simple',
-        },
-        'celery.beat': {'class': 'logging.StreamHandler', 'level': 'ERROR'},  # don't log every celerybeat wakeup
```
shanemcd, line 868: This was unused so I just deleted it.

awx/settings/defaults.py
```diff
@@ -821,86 +821,25 @@ def IS_TESTING(argv=None):
...
-            'formatter': 'simple',
-        },
-        'management_playbooks': {
-            'level': 'DEBUG',
-            'class': 'logging.handlers.WatchedFileHandler',
-            'filters': ['require_debug_false'],
-            'filename': os.path.join(LOG_ROOT, 'management_playbooks.log'),
-            'formatter': 'simple',
-        },
-        'system_tracking_migrations': {
```
shanemcd, line 884: This also seemed to be unused so I deleted it.

awx/settings/defaults.py
```diff
@@ -821,86 +821,25 @@ def IS_TESTING(argv=None):
...
-        },
-        'rbac_migrations': {
-            'level': 'WARNING',
-            'class': 'logging.handlers.WatchedFileHandler',
-            'filters': ['require_debug_false'],
-            'filename': os.path.join(LOG_ROOT, 'tower_rbac_migrations.log'),
-            'formatter': 'simple',
-        },
-        'job_lifecycle': {
-            'level': 'DEBUG',
```
shanemcd, line 899: We were explicitly setting debug here, but almost all of our other handlers have a comment that says "don't define a level here, it's set by settings.LOG_AGGREGATOR_LEVEL". @fosterseth Do you know if this was intentional? If so, do you remember the reason?

fosterseth: can't remember, but I do know the job lifecycle logging uses debug level, so maybe setting `level` here was to override log_aggregator_level.

We can (and probably should) just change this to `info`
```python
        logger_job_lifecycle.debug(msg, extra=extra)
```
https://github.com/fosterseth/awx/blob/devel/awx/main/models/unified_jobs.py#L1552

AlanCoding (approved):

PR #12709, opened 2022-08-22
Update node details and legend in Topology View.

kialam:
##### SUMMARY

Update node details and legend in Topology View.
Screen Shot 2022-08-21 at 8 16 29 PM

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - UI


awx/ui/src/screens/TopologyView/Legend.js
```diff
@@ -70,9 +71,7 @@ function Legend() {
       <DescriptionList isHorizontal isFluid>
         <DescriptionListGroup>
           <DescriptionListTerm>
-            <Button variant="primary" isSmall>
-              {t`C`}
-            </Button>
+            <Button isSmall>{t`C`}</Button>
```
AlexSCorey, line 73: Just thinking about how this is going to be translated.  I think we should add an aria-label here to help with accessibility.

awx/ui/src/screens/TopologyView/Tooltip.js
```diff
@@ -71,36 +184,130 @@ function Tooltip({
...
+                          </div>
+                          <Slider
+                            areCustomStepsContinuous
+                            max={1}
+                            min={0}
+                            step={0.1}
+                            value={instanceDetail.capacity_adjustment}
+                            onChange={handleChangeValue}
+                            isDisabled={!instanceDetail.enabled}
+                            // isDisabled={!me?.is_superuser || !instance.enabled}
```
AlexSCorey: Remove?

AlexSCorey (approved):
This looks good.  I asked a question offline about a translation question.  I don't really think its 
 a blocker but more of something we need to consider.  Just remove the commented code and we are good to go.

PR #12726, opened 2022-08-23
Fix deadlocks when processing playbook_on_stats

shanemcd:
##### SUMMARY
When doing some prototyping, I discovered the potential for deadlocks in this code. I think we are now somewhat protected from (or covering up) this problem by the fact that we have a small number of callback receivers. When running many of these queries at once, you start to see stuff like this:

image (16)

image (17)

I ran this patch through some performance tests and am no longer seeing this 🎉 

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/models/events.py
+++ b/awx/main/models/events.py
@@ -6,7 +6,7 @@ from collections import defaultdict
 
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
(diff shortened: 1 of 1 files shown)


shanemcd:
Well... Never mind. After clearing out my hosts and re-running tests, I'm now seeing deadlocks again.

PR #12728, opened 2022-08-24
Adding prevent_instance_group_fallback

john-westcott-iv:
##### SUMMARY

Adds a `prevent_instnace_group_fallback` field to job templates and inventories and refactors `preferred_instance_groups` too "short circuit" if it finds a flag in either location.

WIP status: Needs unit test updates.

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 - UI

##### AWX VERSION

```
awx: 21.5.1.dev2+g1585f02bff.d20220824
```

##### ADDITIONAL INFORMATION

```

```


AlexSCorey:
This is really great work.  I realize this is a WIP but wanted to provide some feedback.  We still need to update the details view for Job Templates, and for Inventory, along with the unit tests to reflect these new fields.

john-westcott-iv:
Thanks, the checkbox on inventories is a little wonky and I'll probably need some help getting it aligned a little better.

awx/main/models/inventory.py
```diff
@@ -175,6 +175,13 @@ class Meta:
         related_name='inventory_labels',
         help_text=_('Labels associated with this inventory.'),
     )
+    prevent_instance_group_fallback = models.BooleanField(
+        default=False,
+        help_text=(
+            "If enabled, the job template will prevent adding any inventory or organization "
+            "instance groups to the list of preferred instances groups to run on."
```
AlanCoding: This help text isn't adding up for me b/c it's a field on `Inventory` model here (I think). I'd like some clarification of what this means in the context of setting it on inventory.

awx/main/models/jobs.py
```diff
@@ -767,19 +774,14 @@ def is_container_group_task(self):
...
-        else:
-            template_groups = []
-        selected_groups = template_groups + inventory_groups + organization_groups
+        selected_groups = []
+        for obj_type in ['job_template', 'inventory', 'organization']:
+            if getattr(self, obj_type) is not None:
+                for instance_group in getattr(self, obj_type).instance_groups.all():
+                    selected_groups.append(instance_group)
+                if getattr(getattr(self, obj_type), 'prevent_instance_group_fallback', False):
+                    logger.error("Breaking in preferred instance group at {}".format(obj_type))
```
AlanCoding: Since this is intended, I don't think a log would be appropriate. Even if you do a different lower-level log, there's contextual info we'd need, like the job id.

awx/main/models/jobs.py
```diff
@@ -268,6 +268,13 @@ class Meta:
             'admin_role',
         ],
     )
+    prevent_instance_group_fallback = models.BooleanField(
+        default=False,
+        help_text=(
+            "If enabled, the job template will prevent adding any inventory or organization "
+            "instance groups to the list of preferred instances groups to run on."
```
AlanCoding, line 281: Isn't the value of `JT.prevent_instance_group_fallback = False` inconsistent with `ask_instance_groups_on_launch`? Perhaps this could be enforced with validation logic...

I guess you could say someone would _want_ the behavior that the job uses org / inventory IGs but only uses the provided ones if the user passes those when launching. But it is a big stretch to imagine that the user would _understand_ the behavior. I don't want to underestimate how big of a documentation and messaging challenging this would be.

awx/main/models/inventory.py
john-westcott-iv: help text updated.

awx/main/models/jobs.py
john-westcott-iv: Good catch, this was intended as debug but I bumped it up for testing. I'll move this back to debug and add a job_id and add more info like job.id for clarity.

john-westcott-iv: Actually, I can't add the job ID here because we are still in the creation of the job. We could say something like:
```
For job template <name> (<id>) breaking in preferred instance group at <obj_type> when launching.
``` 
Thoughts? Should we just get rid of the log message?

awx/main/models/jobs.py
john-westcott-iv: `JT.prevent_instance_group_fallback` drops you out of the loop in preferred_instance_groups.
`ask_instance_groups_on_launch` (if passed in) [actually prevents us from running perferred_instance_groups](https://github.com/ansible/awx/blob/feature-prompt-on-launch-on-templates/awx/main/models/unified_jobs.py#L385). 

So a non-blank ask_instance_groups_on_launch actually supersedes this option all together. 
What if we add a note in the prompt on launch like:
```
Note setting instance groups from a prompt on launch will set the jobs instances groups to that value and not allow for the default fall back behavior from job template to inventory to organization.
```

awx/main/models/jobs.py
AlanCoding: Yeah, that's an option. You could even do the JT id as a less risk-prone option. But I'd also say that getting rid of the log would be a good idea.

john-westcott-iv: Debug removed per our conversation.

awx/main/models/jobs.py
john-westcott-iv: I modified the help text in this PR to note that setting this new option and providing an empty list of instance groups will result in the global instance groups being selected.

AlexSCorey:
Needs QE tests to be created.

awx/ui/src/screens/Template/shared/JobTemplateForm.js
```diff
@@ -545,6 +545,12 @@ function JobTemplateForm({
                     label={t`Enable Fact Storage`}
                     tooltip={helpText.enableFactStorage}
                   />
+                  <CheckboxField
+                    id="option-prevent-instance-group-fallback"
+                    name="prevent_instance_group_fallback"
+                    label={t`Prevent Instance Group Fallback`}
+                    tooltip={helpText.preventInstanceGroupFallback}
```
AlexSCorey, line 604: Just be cognizant of this [PR](https://github.com/ansible/awx/pull/12788). As it does impact the execution of gathering the help text strings just slightly.

AlexSCorey (changes requested):
We need to add these Details to Inventory Detail and Job Template Details view as well

AlexSCorey:
The only test case for integration tests is whether the `Prevent fallback....` value shows up on the details page after save.

AlanCoding (approved):

awx/ui/src/screens/Template/shared/JobTemplateForm.js
john-westcott-iv: That caused me some heartache in a rebase but it should be squared away now and I updated the details view for the options. Take a look when you can and we can get this moved into testing.

awx/ui/src/screens/Inventory/InventoryDetail/InventoryDetail.js
```diff
@@ -104,6 +125,22 @@ function InventoryDetail({ inventory }) {
...
+        )}
+        {renderOptionsField && (
+          <Detail
+            fullWidth
+            label={t`Enabled Options`}
+            value={renderOptions}
+            dataCy="jt-detail-enabled-options"
+            helpText={helpText.enabledOptions}
+          />
+        )}
```
keithjgrant, line 143: can these two conditionals be grouped into one, since `renderOptionsField === prevent_instance_group_fallback` (assigned on line 65)

keithjgrant (approved):

awx/ui/src/screens/Inventory/InventoryDetail/InventoryDetail.js
john-westcott-iv: We could but I kept them separate to mimic how JTs work and to reduce work if we ever added another inventory option.

AlanCoding:
There are several other implementations of this method `preferred_instance_groups`, and I don't think we can call this complete without updating those. For example, for ad hoc commands.

https://github.com/ansible/awx/blob/85a5b58d1882c683341573390b9213dc292f97ff/awx/main/models/ad_hoc_commands.py#L230-L242

If I set this flag for my inventory, I would expect that jobs will not fallback to organization instance groups. But why should I expect otherwise when running ad hoc commands against that inventory?

I think that's a pretty critical change needed here before merge. Otherwise I don't have any major objections. You could re-apply the pattern you have to those methods, or maybe delete some of the non-unified versions of that method if what you have is general enough.

john-westcott-iv:
@AlanCoding good feedback. I modified ad_hoc and inventory source updates as well. Please have a look and let me know what you think.

awx/main/models/ad_hoc_commands.py
```diff
@@ -228,15 +228,14 @@ def add_to_update_fields(name):
...
-            organization_groups = []
+        selected_groups = []
         if self.inventory is not None:
-            inventory_groups = [x for x in self.inventory.instance_groups.all()]
-        else:
-            inventory_groups = []
-        selected_groups = inventory_groups + organization_groups
+            for instance_group in self.inventory.instance_groups.all():
+                selected_groups.append(instance_group)
+            if not getattr(self.inventory, 'prevent_instance_group_fallback', False) and self.inventory.organization is not None:
```
AlanCoding: I don't think `getattr` is necessary here. Since you know the type, you can safely write `self.inventory.prevent_instance_group_fallback`

awx/main/models/jobs.py
```diff
@@ -274,6 +274,15 @@ class Meta:
             'admin_role',
         ],
     )
+    prevent_instance_group_fallback = models.BooleanField(
```
AlanCoding, line 277: maybe you need to add it to the list of FIELDS_TO_PRESERVE_AT_COPY or something like that. Minor detail, but best to wrap it up now.

AlanCoding (approved):
Pretty targeted enhancement, all my blocker comments are addressed, fairly low-level of concern for merging.

AlanCoding:
One of the big next steps I think we should take in this area (and notifications too) is to express the search criteria used in the UI... in some special place the user can drill down into, and have this self-document. Same for notification templates and other organization-fallback kind of things.

AlexSCorey (approved):
I pulled this down and everything is working as expected.

jbradberry (approved):
I think I'm ok with this once the side PR folds into here.

AlanCoding (approved):

jay-steurer:
verified by QE

jay-steurer (approved):

Issue #12675, opened 2022-08-17
`make messages` fail with `invalid locale en-us, did you mean en_us?`


TheRealHaoLiu:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

```
bash-5.1$ make messages
invalid locale en-us, did you mean en_us?
```

### AWX version

awx: 21.4.1.dev61+gd2217bc61d.d20220817

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [X] Other

### Installation method

docker development environment

### Modifications

no

### Steps to reproduce

- standup docker-compose development environment using `make docker-compose`
- validate that development environment come up cleanly by visiting `https://localhost:8043/api/v2` and login 
- exec into docker container `docker exec -it tools_awx_1 bash`
- run make target `make messages` in `/awx_devel`

### Expected results

US English local file should be generated

### Actual results

encounter error
```
bash-5.1$ make messages
invalid locale en-us, did you mean en_us?
```

fosterseth:
want to make sure this isn't an issue before sending off for translations

rooftopcellist:
We override it when called by translation automation.  So it hasn't really been an issue when pushing strings.  Here is a PR to fix the make target so it works out of the box.  The lang ref was just wrong.  

https://github.com/ansible/awx/pull/12731

AlexSCorey:
@TheRealHaoLiu This may be a make target to remove.  Sounds like tower-packaging doesn't use it.  I'm not aware of others that use it, but we'd want to confirm that before removing it.

rooftopcellist:
@AlexSCorey it is used here https://github.com/ansible/awx/blob/9f8b3948e1aad75620bec818f1965b1bb1acbf6e/tools/scripts/l18n/pre_translation.sh#L5

we need to keep the make target.

rooftopcellist:
Fixed in linked PR.

PR #12731, opened 2022-08-24
Fix make target for compiling api strings

rooftopcellist:
##### SUMMARY
Up until recently, our translation automation had been overriding this `--lang` flag with `en_us`, which is why this issue got overlooked previously.  This lang ref should be updated to fix the Makefile.  Addresses https://github.com/ansible/awx/issues/12675

##### ISSUE TYPE

##### COMPONENT NAME

 - Other


rooftopcellist:
Testing:
```
bash-5.1$ make messages
invalid locale en-us, did you mean en_us?

bash-5.1$ vim Makefile      # Edit make target to use `en_us`
bash-5.1$ make messages
processing locale en_us
```

AlexSCorey (approved):
Looks good to me.

PR #12732, opened 2022-08-24
Adding EE/IG/labels/forks/timeout/job_slice_count to schedules

john-westcott-iv:
##### SUMMARY

TODO:
- [x] AWX Collection Changes
- [x] New unit tests
- [x] Test WFJT Nodes

##### ISSUE TYPE

 - New or Enhanced Feature
 
##### COMPONENT NAME

 - API
 
##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```


awx/api/views/__init__.py
```diff
@@ -2985,6 +3049,38 @@ class WorkflowJobTemplateNodeCredentialsList(LaunchConfigCredentialsBase):
...
+            if existing.exists():
+                existing = existing[0]
+                request.data['id'] = existing.id
+                del request.data['name']
+                del request.data['organization']
+        if models.Label.objects.filter(workflowjobtemplatenode_labels=self.kwargs['pk']).count() > 100:
+            return Response(
+                dict(msg=_('Maximum number of labels for {} reached.'.format(self.parent_model._meta.verbose_name_raw))), status=status.HTTP_400_BAD_REQUEST
+            )
+        return super(WorkflowJobTemplateNodeLabelsList, self).post(request, *args, **kwargs)
```
AlanCoding: since this has already repeated 3 times (that you added), this would be a good candidate to consolidate into a shared method.

john-westcott-iv: Done.

awx/api/views/__init__.py
```diff
@@ -209,6 +209,26 @@ def api_exception_handler(exc, context):
     return exception_handler(exc, context)
 
 
+class LabelList(DeleteLastUnattachLabelMixin, SubListCreateAttachDetachAPIView):
```
AlanCoding, line 212: class name is a little confusing, would be better to say `LabelSubList` or something /nit

awx/api/views/__init__.py
```diff
@@ -209,6 +209,26 @@ def api_exception_handler(exc, context):
...
+    def post(self, request, *args, **kwargs):
+        # If a label already exists in the database, attach it instead of erroring out
+        # that it already exists
+        if not getattr(self, 'label_filter', None):
```
AlanCoding, line 216: If it's a reverse ForeignKey then `parent_key` is the back-reference. It seems that we don't have something similar for many-to-many relationship sublists.

But it seems that we could get it in an algorithmic way.

```
In [6]: Schedule._meta.get_field('labels').remote_field.name
Out[6]: 'schedule_labels'

In [7]: JobTemplate._meta.get_field('labels').remote_field.name
Out[7]: 'unifiedjobtemplate_labels'

In [8]: WorkflowJobNode._meta.get_field('labels').remote_field.name
Out[8]: 'workflowjobnode_labels'
```

It's not that I'd prefer this generally, but here, it would avoid the need for the `return Response(...)` part, which is not good practice and also fairly unreachable by testing.

awx/main/access.py
```diff
@@ -1923,19 +1923,91 @@ def can_change(self, obj, data):
...
+                return False
+            return True
+
+        if isinstance(sub_obj, InstanceGroup) and relationship == 'instance_groups':
+            if not sub_obj in self.user.get_queryset(InstanceGroup):
+                logger.debug(
+                    "User {} not allowed access to instance_group {} for {} {} ({})".format(self.user.username, sub_obj.name, obj.__class__, obj_name, obj.id)
+                )
+                return False
+            return True
```
AlanCoding, line 1951: I get the need for some divergences from the parent class here:

https://github.com/ansible/awx/blob/9203062cd64b929cbc6d692066aec7d2e58ab63f/awx/main/access.py#L284-L288

But I'd like to sketch out the goals here from the first principles. In this change, you added `instance_groups` to WFJT nodes, not just schedules. This is evidenced by the introduction of the view `WorkflowJobTemplateNodeInstanceGroupsList`, likewise with schedules.

But WFJT node access class makes no reference to the access method here. So why bother? Or is it broken?

There's also a high degree of repetition here. If the node identifier is a big problem, then fix it in `BaseAccess.can_attach`, and it's really hard for me to see the need for the subclass to handle much... other than not checking the primary object can_change (because it's a supplemental check), but this makes it hard to justify keeping this method at all, since the `BaseAccess` implementation will probably be fine as it is manifested for `ScheduleAccess.can_attach` and the like.

AlanCoding (approved):
LGTM into the feature branch, recognizing that the RBAC logic still needs some work and testing.

PR #12735, opened 2022-08-24
[sig validation] better error for job template run

### LINKED ISSUE #12690: improve job launch failed error when due to content signature validation failure

### Feature Summary

As a user I want to be able to see succinct and informative messaging in the event that content signature validation fails. This failure will reside in the job failure messaging and will be the same output for UI, API, and CLI.

Also update what we are calling task (right now says project integrity  vs Content Signature Validation. This nomenclature needs to be updated.)

- [X] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [X] CLI
- [ ] Other

Image

relrod:
##### SUMMARY
When launching a job template, if the last project update failed due to
signature validation, show an error that actually says that.

**NOTE**
I have *no* idea if this is the right way to do this, but if nothing else it acts as a proof of concept and shows that it's possible.
I am tagging @AlanCoding as a reviewer for feedback.

Potentially fixes #12690

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 - UI

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -4201,7 +4201,7 @@ class JobLaunchSerializer(BaseSerializer):
             if latest_update is not None and latest_update.failed:
                 failed_validation_tasks = latest_update.project_update_events.filter(
                     event='runner_on_failed',
-                    play="Validate project content integrity",
+                    play="Perform project signature/checksum verification",
                 )
                 if failed_validation_tasks:
                     errors['playbook'] = _("Last project update failed due to signature validation failure.")
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -4197,6 +4197,15 @@ class JobLaunchSerializer(BaseSerializer):
         elif template.project.status in ('error', 'failed'):
             errors['playbook'] = _("Missing a revision to run due to failed project update.")
 
+            latest_update = template.project.project_updates.last()
+            if latest_update is not None and latest_update.failed:
+                failed_validation_tasks = latest_update.project_update_events.filter(
+                    event='runner_on_failed',
+                    play="Validate project content integrity",
+                )
+                if failed_validation_tasks:
+                    errors['playbook'] = _("Last project update failed due to signature validation failure.")
+
         # cannot run a playbook without an inventory
         if template.inventory and template.inventory.pending_deletion is True:
             errors['inventory'] = _("The inventory associated with this Job Template is being deleted.")


awx/api/serializers.py
```diff
@@ -4197,6 +4197,16 @@ def validate(self, attrs):
         elif template.project.status in ('error', 'failed'):
             errors['playbook'] = _("Missing a revision to run due to failed project update.")
 
+            latest_update = template.project.project_updates.last()
+            if latest_update is not None and latest_update.failed:
+                # TODO: How slow is this query when there are tons of other events?
```
TheRealHaoLiu: there shouldnt be much event in project_update so i dont think this is much of a concern

relrod: Ok, cool :+1:

relrod: I dug into this a bit more because it worried me. I'd like @AlanCoding's feedback on this whenever he is back and has time. Maybe there's a better approach.

That said, I am not a DB expert, but I think the query might actually be okay, here. (The query below is what Django generates plus a few extra quotes around strings, to make it valid SQL).

```
awx=# EXPLAIN ANALYZE SELECT "main_projectupdateevent"."id", "main_projectupdateevent"."event", "main_projectupdateevent"."event_data", "main_projectupdateevent"."failed", "main_projectupdateevent"."changed", "main_projectupdateevent"."uuid", "main_projectupdateevent"."playbook", "main_projectupdateevent"."play", "main_projectupdateevent"."role", "main_projectupdateevent"."task", "main_projectupdateevent"."counter", "main_projectupdateevent"."stdout", "main_projectupdateevent"."verbosity", "main_projectupdateevent"."start_line", "main_projectupdateevent"."end_line", "main_projectupdateevent"."created", "main_projectupdateevent"."modified", "main_projectupdateevent"."project_update_id" FROM "main_projectupdateevent" WHERE ("main_projectupdateevent"."project_update_id" = 5 AND "main_projectupdateevent"."event" = 'runner_on_failed' AND "main_projectupdateevent"."play" = 'Validate project content integrity') ORDER BY "main_projectupdateevent"."id" ASC;
                                                                                                     QUERY PLAN                                                                                                     
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=16.35..16.35 rows=2 width=1839) (actual time=0.042..0.043 rows=1 loops=1)
   Sort Key: main_projectupdateevent_20220824_21.id
   Sort Method: quicksort  Memory: 27kB
   ->  Append  (cost=0.14..16.34 rows=2 width=1839) (actual time=0.036..0.037 rows=1 loops=1)
         ->  Index Scan using main_projectupdateevent_20220_project_update_id_job_create_idx2 on main_projectupdateevent_20220824_21  (cost=0.14..8.17 rows=1 width=772) (actual time=0.009..0.009 rows=0 loops=1)
               Index Cond: (project_update_id = 5)
               Filter: (((event)::text = 'runner_on_failed'::text) AND ((play)::text = 'Validate project content integrity'::text))
         ->  Index Scan using main_projectupdateevent_20220_project_update_id_job_create_idx5 on main_projectupdateevent_20220825_18  (cost=0.14..8.16 rows=1 width=2908) (actual time=0.024..0.025 rows=1 loops=1)
               Index Cond: (project_update_id = 5)
               Filter: (((event)::text = 'runner_on_failed'::text) AND ((play)::text = 'Validate project content integrity'::text))
               Rows Removed by Filter: 25
 Planning Time: 0.183 ms
 Execution Time: 0.066 ms
```

This index (`main_projectupdateevent_20220_project_update_id_job_create_idx5`) (`idx2` is the same, just on a different partition) is a multicolumn B-tree index:

```
awx=# select indexdef from pg_indexes where indexname = 'main_projectupdateevent_20220_project_update_id_job_create_idx5';
                                                                                     indexdef                                                                                     
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 CREATE INDEX main_projectupdateevent_20220_project_update_id_job_create_idx5 ON public.main_projectupdateevent_20220825_18 USING btree (project_update_id, job_created, counter)
(1 row)
```

and since we are filtering on `project_update_id` (the leftmost column of the index), this should avoid us having to scan the entire table to look for these events - the search space should be only the rows specific to this project update, which should be few and quick.

Again, I am definitely not a database/indexing expert, so I might be missing something, but this is my understanding.

AlanCoding (approved):

awx/api/serializers.py
AlanCoding: Yeah, looks totally fine to have this :+1: 

It only happens for error cases, so it's not slowing down the usage in the happy path. Plus, project updates can only produce a limited number of events (because it's our own playbook), and the filters have the table partitions to keep them fast.

djyasin (approved):
This one looks great to me as well!

AlanCoding (approved):

PR #12744, opened 2022-08-25
[feature] Ability to add execution nodes at runtime

TheRealHaoLiu:
##### SUMMARY
This PR delivers the ability to add and peer out to remote execution for AWX deployed on Kuberentes cluster.

**API**
- introducing `node_state` to `Instance`
- introducing `link_state` to `InstanceLink`
- `Instance` and `InstanceLink` states will automatically change based on events in the system
- restricted job execution to `Enabled` and `Ready` instances
- added `listener_port` field to Instance for customization of receptor `listener_port`
- added ability to auto generate receptor configuration change for control node receptor instances to peer out to remote execution nodes
- added `instance_install_bundle` endpoint to Instance for generating install bundle needed for initializing remote execution node 
- `instance_healthcheck` endpoint now kicks off a async task to execute the health check.

**UI**
- enable creation of `InstanceGroup` on AWX deployed on Kubernetes cluster on the UI
- added forum to `Instance` UI for adding new remote execution instances.
- added ability to remove of remote execution node in the UI
- improved mesh topology visualizer to display node and link state and directionality of link 
- Instance health check updated to support async instance health check behavior

**Collection**
- added `instance` module to support instance related operations

**AWXKIT**
- support install bundle endpoint

**Other**
- created https://github.com/ansible/receptor-collection to support installation of receptor
- updated https://github.com/ansible/awx-operator to support runtime modification of receptor.conf

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 - UI
 - Collection
 - CLI
 - Docs
 - Other

##### AWX VERSION

```
awx: 21.6.1.dev66+gd1d9a1e01f
```

##### ADDITIONAL INFORMATION
https://github.com/ansible/awx/pull/12910 documentation PR for the feature


TheRealHaoLiu:
https://github.com/ansible/awx/pull/12910

PR containing the documentation

awx/api/serializers.py
```diff
@@ -4840,10 +4851,51 @@ def get_percent_capacity_remaining(self, obj):
...
+
+        return value
+
+    def validate_hostname(self, value):
+        if self.instance and self.instance.hostname != value:
+            raise serializers.ValidationError("Cannot change hostname.")
+
+        return value
+
+    def validate_listener_port(self, value):
```
john-westcott-iv, line 4988: If I add some nodes and then realize I got something wrong do I need to delete it and and readd? If so, would it make sense to add an additional check and if the state is installed (just started) to allow changes to hostname/port?

awx/api/templates/instance_install_bundle/install_receptor.yml
```diff
@@ -0,0 +1,18 @@
...
+---
+- hosts: all
+  become: yes
+  tasks:
+    - name: Create the receptor user
+      user:
+        name: "{{ receptor_user }}"
+        shell: /bin/bash
+    - name: Enable Copr repo for Receptor
+      command: dnf copr enable ansible-awx/receptor -y
```
john-westcott-iv, line 11: Do we need to think about offline installs or people with satellite or anything else around enabling repos?

awx/api/views/__init__.py
```diff
@@ -173,6 +173,7 @@
 from awx.api.pagination import UnifiedJobEventPagination
 from awx.main.utils import set_environ
 
+from awx.api.views.instance_install_bundle import InstanceInstallBundle  # noqa
```
john-westcott-iv: Nit picky, can we move this up into the other awx.api.views import statements?

awx/api/views/instance_install_bundle.py
```diff
@@ -0,0 +1,188 @@
...
+
+    name = _('Install Bundle')
+    model = models.Instance
+    serializer_class = serializers.InstanceSerializer
+    permission_classes = (IsSystemAdminOrAuditor,)
+
+    def get(self, request, *args, **kwargs):
+        instance_obj = self.get_object()
+
+        # if the instance is not a hop or execution node than return 400
```
john-westcott-iv: Comment needs update?

awx/api/views/instance_install_bundle.py
```diff
@@ -0,0 +1,188 @@
...
+            return Response(
+                data=dict(msg=_('Install bundle can only be generated for execution nodes.')),
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        with io.BytesIO() as f:
+            with tarfile.open(fileobj=f, mode='w:gz') as tar:
+                # copy /etc/receptor/tls/ca/receptor-ca.crt to receptor/tls/ca in the tar file
+                tar.add(
+                    os.path.realpath('/etc/receptor/tls/ca/receptor-ca.crt'), arcname=f"{instance_obj.hostname}_install_bundle/receptor/tls/ca/receptor-ca.crt"
```
john-westcott-iv, line 60: We reuse `{instance_obj.hostname}_install_bundle` several times, would it make sense to variablize it incase we ever needer/wanted to change it?

awx/api/serializers.py
```diff
@@ -4840,10 +4851,51 @@ def get_percent_capacity_remaining(self, obj):
         else:
             return float("{0:.2f}".format(((float(obj.capacity) - float(obj.consumed_capacity)) / (float(obj.capacity))) * 100))
 
-    def validate(self, attrs):
-        if self.instance.node_type == 'hop':
-            raise serializers.ValidationError(_('Hop node instances may not be changed.'))
-        return attrs
+    def validate(self, data):
```
john-westcott-iv, line 4948: Do we want to do any validation on the name field here? It looks like the Name field is used for generating certs and other items so should it be allowed to contain spaces, special characters, etc?

awx/api/views/instance_install_bundle.py
```diff
@@ -0,0 +1,188 @@
...
+        )
+        .add_extension(
+            x509.SubjectAlternativeName(san_params),
+            critical=False,
+        )
+        .sign(key, hashes.SHA256())
+    )
+
+    # sign csr with the receptor ca key from /etc/receptor/ca/receptor-ca.key
+    with open('/etc/receptor/tls/ca/receptor-ca.key', 'rb') as f:
```
john-westcott-iv, line 153: Let me know if I am wrong but these are just the temp files that we are going to be placing into the zip file. If so, is there any chance of confusion if we write these new files out to these locations instead of a temp location?

awx/main/models/ha.py
```diff
@@ -59,6 +59,15 @@ class InstanceLink(BaseModel):
     source = models.ForeignKey('Instance', on_delete=models.CASCADE, related_name='+')
     target = models.ForeignKey('Instance', on_delete=models.CASCADE, related_name='reverse_peers')
 
+    class States(models.TextChoices):
+        ADDING = 'adding', _('Adding')
+        ESTABLISHED = 'established', _('Established')
+        REMOVING = 'removing', _('Removing')
+
+    link_state = models.CharField(
+        choices=States.choices, default=States.ESTABLISHED, max_length=16, help_text=_("Indicates the current life cycle stage of this peer link.")
```
john-westcott-iv, line 68: Should `max_length` be reduced here?

awx/main/models/ha.py
```diff
@@ -127,13 +136,33 @@ class Instance(HasPolicyEditsMixin, BaseModel):
...
-    ]
-    node_type = models.CharField(default='hybrid', choices=NODE_TYPE_CHOICES, max_length=16)
+
+    class Types(models.TextChoices):
+        CONTROL = 'control', _("Control plane node")
+        EXECUTION = 'execution', _("Execution plane node")
+        HYBRID = 'hybrid', _("Controller and execution")
+        HOP = 'hop', _("Message-passing node, no execution capability")
+
+    node_type = models.CharField(default=Types.HYBRID, choices=Types.choices, max_length=16, help_text=_("Role that this node plays in the mesh."))
```
john-westcott-iv, line 146: Same question about `max_length`

awx/main/models/ha.py
```diff
@@ -127,13 +136,33 @@ class Instance(HasPolicyEditsMixin, BaseModel):
...
+        DEPROVISIONING = 'deprovisioning', _('De-provisioning')
+        DEPROVISION_FAIL = 'deprovision-fail', _('De-provisioning Failure')
+
+    node_state = models.CharField(
+        choices=States.choices, default=States.READY, max_length=16, help_text=_("Indicates the current life cycle stage of this instance.")
+    )
+    listener_port = models.PositiveIntegerField(
+        blank=True,
+        default=27199,
+        validators=[MinValueValidator(1), MaxValueValidator(65535)],
```
john-westcott-iv, line 163: Since receptor runs as non-root do we want to set MinValueValidtor to 1024?

awx/main/tasks/receptor.py
```diff
@@ -574,3 +604,105 @@ def kube_config(self):
...
+        raise RuntimeError("Receptor reload failed")
+
+    links = InstanceLink.objects.filter(source=this_inst, target__in=instances, link_state=InstanceLink.States.ADDING)
+    links.update(link_state=InstanceLink.States.ESTABLISHED)
+
+
+@task(queue=get_local_queuename)
+def remove_deprovisioned_node(hostname):
+    InstanceLink.objects.filter(source__hostname=hostname).update(link_state=InstanceLink.States.REMOVING)
+    InstanceLink.objects.filter(target__hostname=hostname).update(link_state=InstanceLink.States.REMOVING)
```
john-westcott-iv, line 692: Is this in error?

awx/main/tasks/receptor.py
```diff
@@ -574,3 +604,105 @@ def kube_config(self):
...
+
+    node_jobs = UnifiedJob.objects.filter(
+        execution_node=hostname,
+        status__in=(
+            'running',
+            'waiting',
+        ),
+    )
+    while node_jobs.exists():
```
john-westcott-iv, line 701: Do we want to add a counter to prevent an infinite loop here? Also, do we need to update node_jobs in here or is that happening async? If it is an async thing can we add a comment so that someone reading this in the future will know?

awx/main/models/ha.py
```diff
@@ -127,13 +136,33 @@ class Instance(HasPolicyEditsMixin, BaseModel):
...
-        ("execution", "Execution plane node"),
-        ("hybrid", "Controller and execution"),
-        ("hop", "Message-passing node, no execution capability"),
-    ]
-    node_type = models.CharField(default='hybrid', choices=NODE_TYPE_CHOICES, max_length=16)
+
+    class Types(models.TextChoices):
+        CONTROL = 'control', _("Control plane node")
+        EXECUTION = 'execution', _("Execution plane node")
+        HYBRID = 'hybrid', _("Controller and execution")
```
john-westcott-iv, line 143: In the code we seem to have reference like `self.type is 'hybrid'` since we have these do you want to change those to Instances.Types.HYBRID? (applies to other types). This would probably be pretty a pretty big change and is likely out of scope.

awx/ui/src/screens/Instances/InstanceDetail/InstanceDetail.js
```diff
@@ -127,135 +149,229 @@ function InstanceDetail({ setBreadcrumb }) {
...
-                      isDisabled={!me?.is_superuser || !instance.enabled}
-                      data-cy="slider"
+    <>
+      {showHealthCheckAlert ? (
+        <HealthCheckAlert onSetHealthCheckAlert={setShowHealthCheckAlert} />
+      ) : null}
+      <CardBody>
+        <DetailList gutter="sm">
+          <Detail
+            label={t`Host Name`}
```
john-westcott-iv, line 191: Detail screen shows "Host Name" but other screens just use "Name". I would say this screen is correct and the others need updating.

awx/main/models/ha.py
```diff
@@ -59,6 +59,15 @@ class InstanceLink(BaseModel):
     source = models.ForeignKey('Instance', on_delete=models.CASCADE, related_name='+')
     target = models.ForeignKey('Instance', on_delete=models.CASCADE, related_name='reverse_peers')
 
+    class States(models.TextChoices):
```
john-westcott-iv, line 62: From a quick look it looks like there are 2 common combinations of states referenced:
* READY/UNAVAILABLE/INSTALLED
* UNAVAILABLE/INSTALLED

Should we make "meta" states or functions for these to reduce lines like:
`if self.node_state not in (Instance.States.READY, Instance.States.UNAVAILABLE, Instance.States.INSTALLED):`
To something like:
`if self.is_in_active_state():`
Or:
`if self.node_state in Instances.MetaStates.ACTIVE`

.yamllint
```diff
@@ -8,6 +8,8 @@ ignore: |
   awx/ui/test/e2e/tests/smoke-vars.yml
   awx/ui/node_modules
   tools/docker-compose/_sources
+  # django template files
+  awx/api/templates/instance_install_bundle/**
```
AlanCoding, line 12: The yamllint checks _seem_ useful at minimum for this file

```diff
diff --git a/awx/api/templates/instance_install_bundle/inventory.yml b/awx/api/templates/instance_install_bundle/inventory.yml
index 1124cae88e..665e38863c 100644
--- a/awx/api/templates/instance_install_bundle/inventory.yml
+++ b/awx/api/templates/instance_install_bundle/inventory.yml
@@ -2,8 +2,8 @@
 all:
   hosts:
     remote-execution:
-      ansible_host: {{ instance.hostname }}
-      ansible_user: <username> # user provided
+      ansible_host: "{{ instance.hostname }}"
+      ansible_user: <username>  # user provided
       ansible_ssh_private_key_file: ~/.ssh/id_rsa
       receptor_verify: true
       receptor_tls: true
@@ -21,7 +21,7 @@ all:
       receptor_group: awx
       receptor_protocol: 'tcp'
       receptor_listener: true
-      receptor_port: {{ instance.listener_port }}
+      receptor_port: "{{ instance.listener_port }}"
       receptor_dependencies:
         - podman
         - crun
```

Those are good changes, and it would be good to have the checks turned on for these.

On the other hand, `awx/api/templates/instance_install_bundle/install_receptor.yml` does _not_ give useful feedback, but that's because it's a jinja2 template, not a yaml file. Again, not _bad_ feedback, because maybe the `*.j2` extension would be better to have on that file.

awx/main/tasks/receptor.py
jbradberry: No.  If either side of a link is being deprovisioned, the link will be going away.

awx/main/tasks/system.py
```diff
@@ -81,6 +81,10 @@
 def dispatch_startup():
     startup_logger = logging.getLogger('awx.main.tasks')
 
+    # TODO: Enable this on VM installs
```
AlanCoding, line 84: Enable this, as in... have VMs write their receptor config on startup? Implication... stop writing the receptor config for usual installs with the installer? Is that right?

awx/main/tasks/system.py
```diff
@@ -118,11 +122,7 @@ def inform_cluster_of_shutdown():
     try:
         this_inst = Instance.objects.get(hostname=settings.CLUSTER_HOST_ID)
         this_inst.mark_offline(update_last_seen=True, errors=_('Instance received normal shutdown signal'))
-        try:
-            reaper.reap_waiting(this_inst, grace_period=0)
-        except Exception:
-            logger.exception('failed to reap waiting jobs for {}'.format(this_inst.hostname))
```
AlanCoding, line 124: Wait, why'd you get rid of this? Won't waiting jobs get stuck in waiting forever if we don't reap them on shutdown? We could probably give a specific message - "job was in the queue while instance shut down", that would be good. But not reaping here would be bad.

Maybe you moved it somewhere I haven't found yet?

.yamllint
TheRealHaoLiu: current problem is that this is not actually a jinja2 template its django template thats why we are hesitant to add the .j2 extension

awx/main/tasks/system.py
fosterseth: yeah exactly that, once we extend this feature to VM based installs, we'll need the ability to dynamically write receptor's conf

awx/main/tasks/system.py
jbradberry: This was a rebase conflict that was resolved inappropriately.  It should now be fixed.

awx/api/serializers.py
TheRealHaoLiu: follow up issue created to discuss this point https://github.com/ansible/awx/issues/12936

@jbradberry express hesitance on adding this and how it may affect existing clusters

awx/api/templates/instance_install_bundle/install_receptor.yml
TheRealHaoLiu: for special cases the user can modify their playbook to achieve their special cases

awx/api/views/__init__.py
TheRealHaoLiu: https://github.com/ansible/awx/pull/12945

awx/api/views/instance_install_bundle.py
TheRealHaoLiu: we can make the change when the need/want arise

awx/api/views/instance_install_bundle.py
TheRealHaoLiu: this is us reading the file that is mounted in to the container not the destination of the file

we write to BytesIO so no temp file was created

awx/main/models/ha.py
TheRealHaoLiu: ur right it is out of scope but we should clean it up eventually

awx/main/models/ha.py
TheRealHaoLiu: u do have a point, but lets leave that up to the user for now if we run into compiling evidence that this become a common mistake we can change it than

awx/main/tasks/receptor.py
TheRealHaoLiu: created a follow up issue to discuss https://github.com/ansible/awx/issues/12946

awx/main/models/ha.py
jbradberry: The `MetaStates` idea isn't workable given the current capabilities of Django/Python's enum types.

I wouldn't oppose a set of helper methods, but I feel like that's something that will take a bit of shakedown time to see what the groups ought to be and will provide only a marginal gain.

awx/main/models/ha.py
jbradberry: No, probably not.  Currently the enum values go up to a length of 11, so this gives us a little headroom if we add something.  Also, Postgres is pretty efficient about how it uses `varchar(n)` fields, so we're not really wasting the extra space.

AlanCoding:
> updated https://github.com/ansible/awx-operator to support runtime modification of receptor.conf

Is https://github.com/ansible/awx-operator/pull/1012 the correct link for that?

fosterseth (approved):

jay-steurer (approved):

CFSNM (approved):

akus062381 (approved):

rooftopcellist (approved):

relrod (approved):

tvo318:
Docs component addressed via PR https://github.com/ansible/product-docs/pull/1765.

PR #12745, opened 2022-08-25
Allow for projects to be GPG-verified against unexpected changes

relrod:
##### SUMMARY

This PR comprises the work of the Content Signing Working Group over the last several weeks. It includes a number of changes, including:

- the addition of a new default `GPG Public Key` credential type
- a new `Project` model field to make use of this credential type
- changes to the project update playbook and a new action plugin that make use of `ansible-sign` (as a library) for playbook/project signature validation
- necessary changes to add project update tags to make the above function
- UI changes to accommodate the new credential type and make use of it.
- small change to AWX collection to account for the new feature

Hats off to @djyasin @TheRealHaoLiu and @vedaperi for awesome work in making this come together. This PR is the culmination of much learning and great collaboration. It also marks the end of the `feature-project-integrity` branch, as once it is merged, further work will be done against `devel` instead.

To test this PR sufficiently:

- create a test project (possibly in a temporary repository on GitHub). It can be simple such as just an inventory file and a playbook, or more complex.
- use `ansible-sign` to sign the project with a GPG key (there is a [a rundown](https://ansible.github.io/ansible-sign/rundown.html) of how to use `ansible-sign`), and push your changes to the project repository.
- add your GPG public key to AWX as a credential (`gpg --export --armour <your pubkey fingerprint>` will render it).
- add the test project as an AWX project, and enable content signing on the project by selecting your GPG public key in the project edit page.
- sync the project and ensure that the content-verification tasks get run in the project sync job and are successful
- once that is done, modify the project in various ways and ensure that the module **fails** when expected.
  - Example:
    - You can edit a file, without re-signing the project, commit that change, sync the project, make sure it fails
    - Then revert that change so the project is valid again, edit `.ansible-sign/sha256sum.txt` (without re-signing anything), commit that change and ensure validation fails again.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 - UI
 - Collection


awx/playbooks/project_update.yml
```diff
@@ -153,6 +153,39 @@
...
+        - name: verify project integrity using GPG signature
+          playbook_integrity:
+            project_path: "{{ project_path | quote }}"
+            validation_type: gpg
+            gpg_pubkey: "{{ gpg_pubkey }}"
+          register: gpg_result
+
+        - name: GPG result
+          debug:
+            var: gpg_result
```
AlanCoding, line 171: I find the value of these debug statements kind of questionable. If you're in the UI, you can click on the "ok:" line in standard out.

Screenshot from 2022-08-30 12-52-31

Screenshot from 2022-08-30 12-52-38

So you're taking the return data from event data in one event and putting it into standard out in the next event, yes. But when you do that in a debug task, it's not much better for human readability. If I were using this on the regular, I'd think my preference would be to just have the one task, because the "ok:" output is probably sufficient at a high-level view.

Even if you do keep these debug tasks, I'd hope that the data could be narrowed down to _just_ echo the status or the fingerprint or something. This would be proportional to the debug task where we echo the repository version (which I also don't love). I think the JSON data is clearly too much, and my personal preference would be to delete the debug tasks entirely.

awx/main/models/credential/__init__.py
```diff
@@ -1171,6 +1171,25 @@ def create(self):
...
+                'id': 'gpg_public_key',
+                'label': gettext_noop('GPG Public Key'),
+                'type': 'string',
+                'secret': True,
+                'multiline': True,
+                'help_text': gettext_noop('GPG Public Key used to validate content signatures.'),
+            },
+        ],
+        'required': ['gpg_public_key'],
+    },
```
AlanCoding, line 1190: Related to your other TODO, I would prefer that we add to this:

```python
    injectors={
        'extra_vars': {
            "gpg_pubkey": "{{gpg_public_key}}",
        },
    },
```

Even if it's not used, because at least it articulates what it does. If, for whatever reason, someone used one of these credentials in a job, they could hypothetically employ the related plugins, use the same tasks, and have it work.

awx/playbooks/action_plugins/playbook_integrity.py
```diff
@@ -0,0 +1,115 @@
...
+            }
+
+        validation_type = self.params.get("validation_type")
+        if validation_type is None or validation_type not in VALIDATION_TYPES:
+            return {"failed": True, "msg": "validation_type must be one of: " + ', '.join(VALIDATION_TYPES)}
+        validation_method = getattr(self, f"validate_{validation_type}")
+        return validation_method()
+
+    def validate_gpg(self):
+        gpg_pubkey = self.params.get("gpg_pubkey")
```
AlanCoding, line 51: I would definitely want this in the DOCUMENTATION string, so that people could follow what parameters the action plugin expects. Someone from content might be able to help craft that. This would also take care of _some_ of the validation for us.

awx/main/tasks/jobs.py
```diff
@@ -1266,6 +1266,11 @@ def build_extra_vars_file(self, project_update, private_data_dir):
             # for raw archive, prevent error moving files between volumes
             extra_vars['ansible_remote_tmp'] = os.path.join(project_update.get_project_path(check_if_exists=False), '.ansible_awx', 'tmp')
 
+        # TODO: Is this the right way to do this? -relrod
+        if project_update.project.signature_validation_credential is not None:
+            pubkey = project_update.project.signature_validation_credential.get_input('gpg_public_key')
+            extra_vars['gpg_pubkey'] = pubkey
```
AlanCoding, line 1276: I get where this question is coming from now.

How else could we have done this? The history is that jobs used to have many credential ForeignKey relationships. It had `network_credential`, `cloud_credentials`, and others. Then we consolidated those into a single `credentials` relationship. Importantly, this is on the _unified job_ model, so even projects / project updates should have this arbitrary credentials relationship. If we tracked this new credential in that existing many-to-many, as opposed to introducing a new ForeignKey field, then we could define the `extra_vars` injector in the credential type definition, and there would be no code to write for it whatsoever.

However, I get why that's hard, weird, and non-obvious. For projects, we don't expose the `credentials` many-to-many  relationship in the UI _or_ the API. Exposing it would obligate us to write validation rules in the views and a bunch of other stuff. The UI would have to write more custom logic to specially help the user out so that they pick the right credential type (I guess they already add some stuff for this implementation). But most importantly, this would break the pattern we have with `scm_credential`. We could migrate that field into the generic relationship too, but then that creates even more work. This is a pre-existing inconsistency in the app.

Without getting into all of that, this is probably the most direct way to do the injection.

awx/main/models/projects.py
```diff
@@ -284,6 +284,16 @@ class Meta:
...
     )
 
+    # credential (keys) used to validate content signature
+    signature_validation_credential = models.ForeignKey(
+        'Credential',
+        related_name='%(class)ss_signature_validation',
+        blank=True,
+        null=True,
+        default=None,
+        on_delete=models.SET_NULL,
```
AlanCoding, line 294: I think it would be good to add `help_text` to the field definition here to make the OPTIONS more readable.

AlanCoding:
I've finished going through manually testing this and leaving comments. I think the implementation is great, the separate library, packaged in the EE, is perfect!

Wrapping up the comments I'd like to request a response on:

 - I would like the action plugin to add DOCUMENTATION, and reviewed by someone who works with maintaining Ansible content
 - I'd like to define `injectors` in the credential type definition
 - I'd like to remove the debug tasks and the `block:` syntax needed for them
 - add help_text for the new field

awx/playbooks/project_update.yml
relrod: :+1: on deleting the debug tasks

awx/main/models/credential/__init__.py
relrod: As we discussed in PM, I'd like to prohibit these from being used in jobs. It feels weird to allow it, like it's just leaking an implementation detail.

awx/playbooks/action_plugins/playbook_integrity.py
relrod: Yeah, we need a DOCUMENTATION block. I didn't view it as mission-critical since I don't anticipate anyone outside of AWX using this at least any time soon. But it's good to have.

awx/main/tasks/jobs.py
relrod: I will keep this and work to prevent the credential from being used on normal job. I'll remove the TODO.

awx/main/models/projects.py
relrod: @djyasin can you add that?

awx/main/tasks/jobs.py
relrod: Looking into this more, I am not sure how this credential type could be used in a normal job. If I try to apply it to a job via the API, I get:
```
Cannot assign a Credential of kind `cryptography`
```
from `JobTemplateCredentialsList#is_valid_relation`.

Is there anything to do here @AlanCoding?

relrod:
I am actually going to close this out, with intention to re-open it from the `feature-project-integrity` branch instead of a branch off of my own fork.

#12805 is a response to most of the feedback here except for [this question](https://github.com/ansible/awx/pull/12745#discussion_r960269281).

Once #12805 merges into `feature-project-integrity`, I will squash a few commits together, do a final rebase against devel, and open a new PR into devel from that branch.

PR #12757, opened 2022-08-26
Implement instance deprovisioning

jbradberry:
##### SUMMARY
- allow the node_state to be set to 'deprovisioning'
- set the links that touch the instance to 'removing'
- only allow on K8S
- only allow to be done to execution nodes
- kick off a task that waits for jobs to complete, then delete the node and broadcast the `write_receptor_config` task
- make `write_receptor_config` change any 'adding' links (for newly created nodes) to 'established' when complete

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/tasks/receptor.py
+++ b/awx/main/tasks/receptor.py
@@ -25,7 +25,7 @@ from awx.main.utils.common import (
     cleanup_new_process,
 )
 from awx.main.constants import MAX_ISOLATED_PATH_COLON_DELIMITER
-from awx.main.models import Instance
+from awx.main.models import Instance, InstanceLink, UnifiedJob
 from awx.main.dispatch.publish import task
 
 # Receptorctl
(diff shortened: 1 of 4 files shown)


awx/main/tasks/receptor.py
```diff
@@ -641,29 +641,33 @@ def kube_config(self):
...
+                receptor_ctl.simple_command("reload")
+                break
+            except ValueError:
+                logger.warning(f"Unable to reload Receptor configuration. {attempts-attempt} attempts left.")
+                time.sleep(backoff)
+                backoff += 1
+        else:
+            raise RuntimeError("Receptor reload failed")
+
+        InstanceLink.objects.filter(source=this_inst).exclude(target__in=instances).delete()
```
AlanCoding: Wait, why are you deleting these?

jbradberry: That line of code was intended to deal with links to a node that was just deprovisioned (we are excluding those links to nodes that are still there).  However, I ultimately decided to just let Django's cascade delete deal with them.

awx/main/tasks/receptor.py
```diff
@@ -641,29 +641,50 @@ def kube_config(self):
...
-            receptor_ctl.simple_command("reload")
-            break
-        except ValueError:
-            logger.warning(f"Unable to reload Receptor configuration. {attempts-attempt} attempts left.")
-            time.sleep(backoff)
-            backoff += 1
-    else:
-        raise RuntimeError("Receptor reload failed")
+        attempts = 10
+        for backoff in range(1, attempts + 1):
```
shanemcd, line 660: 🆒

AlanCoding (approved):

PR #12775, opened 2022-08-29
adding content into the install bundle

TheRealHaoLiu:
##### SUMMARY
install bundle now contain the following content 

```
├── install_receptor.yml
├── inventory.yml
├── receptor
│   ├── tls
│   │   ├── ca
│   │   │   └── receptor-ca.crt
│   │   ├── receptor.crt
│   │   └── receptor.key
│   └── work-public-key.pem
└── requirements.yml
```

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

##### AWX VERSION
```
awx: 21.5.1.dev26+g45e8712b7f.d20220829
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/api/views/instance_install_bundle.py
+++ b/awx/api/views/instance_install_bundle.py
@@ -1,19 +1,42 @@
 # Copyright (c) 2018 Red Hat, Inc.
 # All Rights Reserved.
 
+import datetime
+import io
 import os
 import tarfile
-import tempfile
+import ipaddress
 
+import asn1
 from awx.api import serializers
 from awx.api.generics import GenericAPIView, Response
 from awx.api.permissions import IsSystemAdminOrAuditor
 from awx.main import models
+from cryptography import x509
+from cryptography.hazmat.primitives import hashes, serialization
+from cryptography.hazmat.primitives.asymmetric import rsa
+from cryptography.x509 import DNSName, IPAddress, ObjectIdentifier, OtherName
+from cryptography.x509.oid import NameOID
+from django.http import HttpResponse
 from django.utils.translation import gettext_lazy as _
 from rest_framework import status
-from django.http import HttpResponse
+
+# Red Hat has an OID namespace (RHANANA). Receptor has its own designation under that.
+RECEPTOR_OID = "1.3.6.1.4.1.2312.19.1"
 
 # generate install bundle for the instance
+# install bundle directory structure
+# ├── install_receptor.yml (playbook)
+# ├── inventory.ini
+# ├── receptor
+# │   ├── vars.yml
+# │   ├── tls
+# │   │   ├── ca
+# │   │   │   └── receptor-ca.crt
+# │   │   ├── receptor.crt
+# │   │   └── receptor.key
+# │   └── work-public-key.pem
+# └── requirements.yml
 class InstanceInstallBundle(GenericAPIView):
 
     name = _('Install Bundle')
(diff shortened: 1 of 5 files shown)


awx/api/views/instance_install_bundle.py
```diff
@@ -31,24 +55,181 @@ def get(self, request, *args, **kwargs):
...
+                    os.path.realpath('/etc/receptor/tls/ca/receptor-ca.crt'), arcname=f"{instance_obj.hostname}_install_bundle/receptor/tls/ca/receptor-ca.crt"
+                )
+
+                # copy /etc/receptor/signing/work-public-key.pem to receptor/work-public-key.pem
+                tar.add('/etc/receptor/signing/work-public-key.pem', arcname=f"{instance_obj.hostname}_install_bundle/receptor/work-public-key.pem")
+
+                # generate and write the receptor key to receptor/tls/receptor.key in the tar file
+                key, cert = self.generate_receptor_tls()
+
+                key_tarinfo = tarfile.TarInfo(f"{instance_obj.hostname}_install_bundle/receptor/tls/receptor.key")
```
fosterseth, line 70: maybe create a loop for each item to reduce the duplicate operations

Makefile
```diff
@@ -564,6 +564,12 @@ awx-kube-dev-build: Dockerfile.kube-dev
 	    --cache-from=$(DEV_DOCKER_TAG_BASE)/awx_kube_devel:$(COMPOSE_TAG) \
 	    -t $(DEV_DOCKER_TAG_BASE)/awx_kube_devel:$(COMPOSE_TAG) .
 
+awx-kube-build: Dockerfile
```
TheRealHaoLiu: added this so that we can build test version of the kube build to deploy on cloud

awx/api/views/instance_install_bundle.py
TheRealHaoLiu: loop does not reduce duplicate 
and also each loop would be different so i would need to create a structure to get both the info 

at that point i think i would just be introducing "extra" structure unnecessarily

awx/api/views/instance_install_bundle.py
```diff
@@ -31,24 +55,183 @@ def get(self, request, *args, **kwargs):
...
+                requirements_yml_tarinfo.size = len(requirements_yml)
+                tar.addfile(requirements_yml_tarinfo, io.BytesIO(requirements_yml.encode('utf-8')))
+
+            # respond with the tarfile
+            f.seek(0)
+            response = HttpResponse(f.read(), status=status.HTTP_200_OK)
+            response['Content-Disposition'] = f"attachment; filename={instance_obj.hostname}_install_bundle.tar.gz"
+            return response
+
+    def generate_playbook(self):
```
jbradberry: These methods would be good candidates to be broken out into functions in one of the utilities modules.  They barely do anything with `self`, and what they _do_ use it for is to re-query for the object, causing additional unnecessary hits to the database.  Just pass in the already obtained instance as a parameter for those that need it.

jbradberry (changes requested):

awx/api/views/instance_install_bundle.py
```diff
@@ -31,24 +55,181 @@ def get(self, request, *args, **kwargs):
...
+        name: ansible.receptor.receptor
+    - name: Install ansible-runner
+      pip:
+        name: ansible-runner
+        executable: pip3.9
+"""
+
+
+def generate_requirements_yml():
+    return """
```
jbradberry: You missed one :P

awx/api/views/instance_install_bundle.py
```diff
@@ -31,24 +55,181 @@ def get(self, request, *args, **kwargs):
...
+                tar.addfile(key_tarinfo, io.BytesIO(key))
+
+                cert_tarinfo = tarfile.TarInfo(f"{instance_obj.hostname}_install_bundle/receptor/tls/receptor.crt")
+                cert_tarinfo.size = len(cert)
+                tar.addfile(cert_tarinfo, io.BytesIO(cert))
+
+                # generate and write install_receptor.yml to the tar file
+                playbook = generate_playbook()
+                playbook_tarinfo = tarfile.TarInfo(f"{instance_obj.hostname}_install_bundle/install_receptor.yml")
+                playbook_tarinfo.size = len(playbook)
```
jbradberry, line 81: This (and the similar patterns below) might cause problems in the presence of actual unicode characters, though at the moment I don't see that any can leak into these.  My recommendation would still be to do the `.encode('utf-8')` up front and pass the results to both `len()` and `io.BytesIO()`.

awx/api/views/instance_install_bundle.py
TheRealHaoLiu: ugg...

awx/api/views/instance_install_bundle.py
TheRealHaoLiu: fixed

jbradberry (approved):

PR #12784, opened 2022-08-30
Can deprovision an Instance from Instance List and Instance detail view

AlexSCorey:
##### SUMMARY
Can de-provision an instance from the list view and from the detail view.  To see these buttons the user must be a super user, on a k8 deployment

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui/src/api/models/Instances.js
+++ b/awx/ui/src/api/models/Instances.js
@@ -25,6 +25,12 @@ class Instances extends Base {
   readInstanceGroup(instanceId) {
     return this.http.get(`${this.baseUrl}${instanceId}/instance_groups/`);
   }
+
+  deprovisionInstance(instanceId) {
+    return this.http.post(`${this.baseUrl}${instanceId}`, {
+      node_state: 'deprovisioning',
+    });
+  }
 }
 
 export default Instances;
--- a/awx/ui/src/screens/Instances/InstanceDetail/InstanceDetail.js
+++ b/awx/ui/src/screens/Instances/InstanceDetail/InstanceDetail.js
@@ -1,6 +1,6 @@
 import React, { useCallback, useEffect, useState } from 'react';
 
-import { Link, useParams } from 'react-router-dom';
+import { Link, useHistory, useParams } from 'react-router-dom';
 import { t, Plural } from '@lingui/macro';
 import {
   Button,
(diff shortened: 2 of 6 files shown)


awx/ui/src/screens/Instances/InstanceDetail/InstanceDetail.js
```diff
@@ -288,8 +296,14 @@ function InstanceDetail({ setBreadcrumb }) {
           />
         )}
       </DetailList>
-      {!isHopNode && (
+      {!isHopNode && me.is_superuser && isK8s && (
```
kialam: `control` and `hybrid` types should also not be "deprovisionable." I would change this logic to just check for whether or not the `node type === 'execution'`

awx/ui/src/api/models/Instances.js
```diff
@@ -25,6 +25,12 @@ class Instances extends Base {
   readInstanceGroup(instanceId) {
     return this.http.get(`${this.baseUrl}${instanceId}/instance_groups/`);
   }
+
+  removeInstance(instanceId) {
```
kialam: The term `remove` is usually meant for `delete` actions when interfacing with REST. Since we are `POSTing` a field value, I would change this to `deprovisionInstance`

awx/ui/src/screens/Instances/InstanceDetail/InstanceDetail.js
kialam: Adding this check here would make it so that non-k8s users can't run Health Checks on their Instances.

awx/ui/src/screens/Instances/Shared/RemoveInstanceButton.js
```diff
@@ -0,0 +1,206 @@
...
+  }
+`;
+
+function RemoveInstanceButton({
+  itemsToRemove,
+  onRemove,
+  clearSelected,
+  isK8s,
+}) {
+  const { isKebabified, onKebabModalChange } = useContext(KebabifiedContext);
```
kialam, line 30: I haven't been able to get this to build locally, however I thought we had discussed with @trahman73 to not use kebabs this release and stick with our established design pattern of the delete button like on other lists?

kialam (changes requested):

awx/ui/src/screens/Instances/Shared/RemoveInstanceButton.js
AlexSCorey: This kebab is used in Advanced search mode.  It is not on the list items.

kialam:
This is looking good. There is a console error that should be fixed before merging:
Screen Shot 2022-09-02 at 12 46 04 PM

AlexSCorey:
@kialam give this another go when you get a chance

kialam (approved):
LGTM, nice job

PR #12787, opened 2022-08-30
Speed up image build when UI is pre-built on host

shanemcd:
##### SUMMARY
While this gives someone the chance to shoot themselves in the foot, it's worth it. In CI we'll always have a clean checkout. If you are building images locally, you need to `make clean-ui` to force a rebuild.

##### ISSUE TYPE

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,2 @@
-awx/ui/node_modules
 Dockerfile
 .git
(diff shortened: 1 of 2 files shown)


TheRealHaoLiu (approved):
on my laptop this speed up the awx image build from 600s to 230s 🥇 

and this wont affect the official awx image build since the `awx/ui/node_modules` is git ignored

PR #12790, opened 2022-08-30
Common Inventory slicing method for job slices

relrod:
##### SUMMARY

- Extract how slicing is done from Inventory#get_script_data and pull it
  into a new method, Inventory#get_sliced_hosts
- Make use of this method in Inventory#get_script_data
- Make use of this method in Job#_get_inventory_hosts (used by
  Job#start_job_fact_cache and Job#finish_job_fact_cache).

This fixes an issue (namely in Tower 4.1) where job slicing with fact
caching enabled doesn't save facts for all hosts.

This is written with backporting all the way to 4.1 in mind.

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/models/inventory.py
+++ b/awx/main/models/inventory.py
@@ -236,6 +236,12 @@ class Inventory(CommonModelNameNotUnique, ResourceMixin, RelatedJobsMixin):
             raise ParseError(_('Slice number must be 1 or higher.'))
         return (number, step)
 
+    def get_sliced_hosts(self, host_queryset, slice_number, slice_count):
+        if slice_count > 1 and slice_number > 0:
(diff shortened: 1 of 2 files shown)


AlanCoding (approved):
Code looks good to me, but I would like to make sure we have some good integration testing backing this up before merge.

relrod:
> Code looks good to me, but I would like to make sure we have some good integration testing backing this up before merge.

See ansible/tower-qa#7663 - if that looks good to you, I will merge that and this together. I did do a yolo run including that test and it passed.

djyasin (commented):
Thank you so much for tagging me on this!!

PR #12796, opened 2022-08-31
Prevent openldap from getting downgraded during build

shanemcd:
##### SUMMARY
We noticed [here](https://github.com/ansible/awx/runs/8118323342?check_suite_focus=true) that openldap was getting downgraded and caused our test suite to blow up.

image

##### ISSUE TYPE

##### COMPONENT NAME

 - Other


TheRealHaoLiu (approved):
lgtm

PR #12803, opened 2022-09-01
Fix an issue where default cleanup schedules only run once

### LINKED ISSUE #12802: Token and Session Expiration never run after the first time

### Bug Summary

Looks like when we implemented token and session cleanup way back in https://github.com/ansible/awx/pull/3856

We populated the recurrence rule incorrectly:

https://github.com/ansible/awx/blob/8a06ffbe15c9f8e68b1da86e5ca7daf5ecfd6da4/awx/main/migrations/_create_system_jobs.py#L39

This schedule will only ever run once due to `COUNT=1`.... we should omit that so that it will periodically run.

### AWX version

latest

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

minishift

### Modifications

no

### Ansible version

_No response_

### Operating system

_No response_

### Web browser

_No response_

### Steps to reproduce

install awx

### Expected results

periodic running of these management jobs

### Actual results

the jobs only run once

### Additional information

_No response_

matburt:
This looks like an oversight that has existed for a long time. We intend to run these on a pretty regular basis

This fixes #12802  only for new installations though. I think this is probably the reasonable thing to do so that we don't trigger a big schedule change that might cause performance implications if we were to change them in place.

### DIFF
--- a/awx/main/migrations/_create_system_jobs.py
+++ b/awx/main/migrations/_create_system_jobs.py
@@ -36,7 +36,7 @@ def create_clearsessions_jt(apps, schema_editor):
     if created:
         sched = Schedule(
             name='Cleanup Expired Sessions',
-            rrule='DTSTART:%s RRULE:FREQ=WEEKLY;INTERVAL=1;COUNT=1' % schedule_time,
+            rrule='DTSTART:%s RRULE:FREQ=WEEKLY;INTERVAL=1' % schedule_time,
             description='Cleans out expired browser sessions',
             enabled=True,
             created=now_dt,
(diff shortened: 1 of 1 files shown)


AlanCoding:
We could change existing installs in a migration that runs a quick script.

It appears that users can modify the schedules for other management jobs, so we could limit it to the case that it was using the default rrule. If we do that, I'm not particularly worried about other fallout.

However, I don't seem to be able to modify these schedule in the UI, as I get a console error. I will see if this needs a new issue in the morning.

matburt:
my concern would be if someone had an extremely large backlog of tokens and sessions... running this task after years of having this task not run might cause some performance issues. That might be a surprising change as part of an upgrade.... it might be better to remediate it once if it ever becomes an issue and then adjust the schedule then. I could go either way on that one.

AlanCoding (approved):

PR #12805, opened 2022-09-01
[feature-project-integrity] Cleanup, mostly around the action plugin and its use in the project update playbook

relrod:
##### SUMMARY

This implements most of the feedback from @AlanCoding's comments in #12745.

Note that this is a PR to merge this feature into `feature-project-integrity`, not `devel`.

- Nuke TODO, no longer needed
- Add `help_text` to new model field
- documentation for action plugin
- remove debug tasks and blocks from update playbook

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/migrations/0167_project_signature_validation_credential.py
+++ b/awx/main/migrations/0167_project_signature_validation_credential.py
@@ -2,6 +2,7 @@
 
 from django.db import migrations, models
 import django.db.models.deletion
+from django.utils.translation import gettext_lazy as _
 
 from awx.main.models import CredentialType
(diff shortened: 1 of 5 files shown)


AlanCoding (approved):
Thanks, looks really good! the CI is failing because it looks like migrations are not synced. If you add `help_text`, that change has to be picked up in the migration file, but it doesn't change the schema at all. That should fix the checks, and then you should be fine to merge.

djyasin (commented):
I think the verbiage for the help information and the documentation is clear and concise. This makes sense to me!

Issue #12739, opened 2022-08-25
enable health check to be run for remote execution node on kube deployment


TheRealHaoLiu:
## Summary
currently health check relies on the awx web process to be able to use receptorctl to connect to the receptor mesh to initiate a healthcheck for the remote execution node 

in kubernetes deployment the receptor socket is not mounted into the awx-web container and health check will never be successful 

while we work on allowing use to add remote execution node to the kubernetes deployment of awx we need enable user to run health check on the node 

ideally we would not want to mount the receptor socket into the awx-web container because it introduces additioanl "coupling" between awx-web and receptor

## ISSUE TYPE
New or Enhanced Feature

## COMPONENT NAME
API

CFSNM:
Validated from QE side. Closing

PR #12827, opened 2022-09-03
Run instance health check in task container

fosterseth:
##### SUMMARY

addresses https://github.com/ansible/awx/issues/12739

awx-web container does not have access to receptor socket, and the
execution node health check requires receptorctl in order to run.

This change runs the health check asynchronously in the task container.

Note -- We no longer spin in a loop waiting for the instance health check to complete. This was not working as intended anyways, as the apply_async inside of a django view does not start the task right away. This is because view are ran in a transaction, and pgnotify will only submit messages after the transaction is complete.

We'll need another method of updating the UI the data that returns from the health check (either websockets or refreshes)

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/models/ha.py
+++ b/awx/main/models/ha.py
@@ -243,20 +243,21 @@ class Instance(HasPolicyEditsMixin, BaseModel):
 
     def mark_offline(self, update_last_seen=False, perform_save=True, errors=''):
         if self.node_state not in (Instance.States.READY, Instance.States.UNAVAILABLE, Instance.States.INSTALLED):
-            return
+            return []
         if self.node_state == Instance.States.UNAVAILABLE and self.errors == errors and (not update_last_seen):
-            return
+            return []
         self.node_state = Instance.States.UNAVAILABLE
         self.cpu_capacity = self.mem_capacity = self.capacity = 0
         self.errors = errors
         if update_last_seen:
             self.last_seen = now()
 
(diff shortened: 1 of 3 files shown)


awx/main/models/ha.py
```diff
@@ -314,7 +315,8 @@ def save_health_data(self, version=None, cpu=0, memory=0, uuid=None, update_last
                 self.node_state = Instance.States.READY
                 update_fields.append('node_state')
         else:
-            self.mark_offline(perform_save=False, errors=errors)
+            fields_to_update = self.mark_offline(perform_save=False, errors=errors)
+            update_fields.extend(fields_to_update)
```
AlanCoding, line 319: duplicates in `update_fields` aren't a problem, right?

fosterseth: yeah I was wondering this too and checked the raw sql with and without duplicate update_fields. They yield the same sql. So I think django is handling duplicates somewhere along the way

AlanCoding:
Does the UI currently behave sanely for this scenario? I don't want to add more websocket communication, but just a "wait and refresh" message would be sufficient IMO, considering the nature of the action.

AlanCoding (approved):
From hybrid-cluster integration tests, we have evidence that `cluster_node_health_check` is probably _already_ buggy and needs to be debugged. Your changes probably won't affect that for better or worse. Going into the feature branch :+1: this is fine by me, but there's still more testing & UI followup work that needs to be on the agenda.

PR #12830, opened 2022-09-05
Fix LDAP volume conditional, better metrics interval

AlanCoding:
##### SUMMARY
First thing here - I find locally that `docker-volume ls` will show the ldap container, even when it shouldn't and I never asked for it. This fixes that.

Second thing - this appears to be a better way to set the graph interval. Grafana was deferring to what Prometheus was reporting, which was 15s for whatever reason. This still isn't templated, but it would be more reasonable for us to template later on.

##### ISSUE TYPE

##### COMPONENT NAME
 - API


john-westcott-iv (approved):
Nice find!

Issue #6753, opened 2020-04-20
Allow metrics endpoint to be retrieved anonymously


svenstaro:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
I think for some monitoring setups it might be helpful to have an option that would allow for retrieving `/api/v2/metrics` without being authenticated. It would be opt-in, though.

blomquisg:
Maybe make this configurable...

bryanasdev000:
+1

mamercad:
+1

texasbobs:
This would be amazing if it was a configurable option.

PR #12832, opened 2022-09-05
Allow metrics collection for anonymous users via settings

no-12:
Signed-off-by: Nico Ohnezat <nico@no-12.net>

##### SUMMARY
Some tools that collect metrics are not able to use authentication to poll metrics. related #6753

This PR will allow to disable the authentication of the metrics endpoint by setting ALLOW_METRICS_FOR_ANONYMOUS_USERS to 'true'.
The default value of ALLOW_METRICS_FOR_ANONYMOUS_USERS is 'false'

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/api/conf.py
+++ b/awx/api/conf.py
@@ -101,7 +101,7 @@ register(
     field_class=fields.BooleanField,
     default=False,
     label=_('Allow anonymous users to poll metrics'),
-    help_text=_('If false, only superusers and system auditors are allowed to poll metrics.'),
+    help_text=_('If true, anonymous users are allowed to poll metrics.'),
     category=_('Authentication'),
     category_slug='authentication',
 )
--- a/awx/api/conf.py
+++ b/awx/api/conf.py
@@ -96,6 +96,15 @@ register(
     category=_('Authentication'),
     category_slug='authentication',
 )
+register(
+    'ALLOW_METRICS_FOR_ANONYMOUS_USERS',
+    field_class=fields.BooleanField,
+    default=False,
+    label=_('Allow anonymous users to poll metrics'),
+    help_text=_('If false, only superusers and system auditors are allowed to poll metrics.'),
+    category=_('Authentication'),
+    category_slug='authentication',
+)
 
 
 def authentication_validate(serializer, attrs):
--- a/awx/api/views/metrics.py
+++ b/awx/api/views/metrics.py
@@ -5,9 +5,11 @@
 import logging
 
 # Django
+from django.conf import settings
 from django.utils.translation import gettext_lazy as _
 
 # Django REST Framework
+from rest_framework.permissions import AllowAny
 from rest_framework.response import Response
 from rest_framework.exceptions import PermissionDenied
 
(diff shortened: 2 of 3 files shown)


awx/api/conf.py
```diff
@@ -96,6 +96,15 @@
     category=_('Authentication'),
     category_slug='authentication',
 )
+register(
+    'ALLOW_METRICS_FOR_ANONYMOUS_USERS',
+    field_class=fields.BooleanField,
+    default=False,
+    label=_('Allow anonymous users to poll metrics'),
+    help_text=_('If false, only superusers and system auditors are allowed to poll metrics.'),
```
AlanCoding: This would read more clearly if you describe the "if true" case, because otherwise I would be tempted to think that a true value would still require authentication, which is not the case.

no-12: yep... that's true 😉

kdelee (commented):
This seems reasonable to me and makes it easier for folks to set up monitoring if they don't mind this info being available to anyone. I could see why some admins would want to lock it down, though, as it could provide some insight into the system they don't want to share

no-12:
> This seems reasonable to me and makes it easier for folks to set up monitoring if they don't mind this info being available to anyone. I could see why some admins would want to lock it down, though, as it could provide some insight into the system they don't want to share

You could block the access to the metrics endpoint in the ingress, and collect the metrics from inside the cluster without authentication.

no-12:
I just want to know if there's any progress in the review process of this pull request.
Maybe there is no prgress, because I messed up @AlanCoding review request to @ffirg ...sorry for that

TheRealHaoLiu:
@ffirg can u respond to this proposal? 

Although I do agree that there may be potential security attack vector and allow unauthenticated user to access metrics
Personally I think this is a good proposal. 

This proposal provides a intentional decision point to the  admin of the instance to expose metrics to unauthenticated user and so long as we validate the setting does what it's suppose to do I think the risk is limited.

AlanCoding:
@no-12 would you mind rebasing your PR to clear out the checks? The check config changed since this was opened and I think that's why it's not passing now. Otherwise I'm thinking there aren't any other blockers for this.

john-westcott-iv:
Thank you so much for the rebase. At this time we are still discussing things internally and will get back to you later as to whether or not we can merge this PR. We are sorry for the delay in acceptance.

ffirg:
I'm happy for this to go in as long as the default is off, the user has to actively turn this on, and there are big red warnings in any docs advising that setting this to true show be carefully evaluated by the customer.

shanemcd:
A rebase should fix the failing CI checks. Once that's done we'll get this merged.

shanemcd:
@AlanCoding If the code looks ok to you go ahead and approve / merge!

no-12:
The api-check fails. I will try to fix it

CFSNM (approved):

CFSNM:
@no-12 Hey, the api-schema precheck fails because a new parameter has been added. This happens everytime we add a new parameter or a new attribute, so we can go ahead a merge

no-12:
Ok, thanks. I haven't looked into the details yet

john-westcott-iv (approved):

PR #12833, opened 2022-09-06
Fix the `is_successful` assertion corner case

AlanCoding:
##### SUMMARY
This was breaking my attempts to write tests that did `job.assert_status(['successful', 'canceled'])`. Because "successful" is in the allowed status list, it won't pass on the "canceled" status because `is_successful` isn't true.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awxkit/awxkit/api/mixins/has_status.py
+++ b/awxkit/awxkit/api/mixins/has_status.py
@@ -63,7 +63,7 @@ class HasStatus(object):
             status_list = [status_list]
         if self.status in status_list:
             # include corner cases in is_successful logic
-            if 'successful' not in status_list or self.is_successful:
+            if self.status != 'successful' or self.is_successful:
                 return
         if msg is None:
             msg = ''


awxkit/awxkit/api/mixins/has_status.py
```diff
@@ -63,7 +63,7 @@ def assert_status(self, status_list, msg=None):
             status_list = [status_list]
         if self.status in status_list:
             # include corner cases in is_successful logic
-            if 'successful' not in status_list or self.is_successful:
+            if self.status != 'successful' or self.is_successful:
```
jbradberry, line 66: This doesn't seem useful, `is_successful` is literally defined as `return self.status == 'successful'`.  It seems to me that this line then will always be true.

If I follow what is going on, this changes the intent of this block.

jbradberry (changes requested):
To be honest, looking at this I realize that this sub-block should probably be taken out entirely.  If `self.status in status_list` is true, we ought to just bail out immediately.

jbradberry:
@AlanCoding having just tracked down the original PR that put that there, which indicates that `is_successful` may have different implementations, I am amending my opinion.  Something like this seems called for:

```
if ('successful' in status_list) == self.is_successful:
    return
```

This sub-block needs a bit better of a comment, though.  We shouldn't have to dig through a stale repo to find this out.

AlanCoding:
```python
if ('successful' in status_list) == self.is_successful:
    return
```

This still doesn't allow the fairly reasonable use of `job.assert_status(['successful', 'canceled'])`, because a canceled job is _not_ successful. That's why I think this needs to reference `self.status` in some way. The conditional is hard to read, that's true.

If I go through all the `is_successful` definition, I honestly don't find any of them helpful looking, and I wouldn't be opposed to deleting this block.

jbradberry:
Yeah, no, I just realized that my suggestion isn't workable either.  This is tricky, and I don't think anymore that getting rid of the inner block is the way to go either.

jbradberry:
Clearing out my review, in case this gets picked up again.

AlanCoding:
Clearly, this is challenging to think about. Although believed to be valid, I will close because unlikely to have time for it.

PR #12845, opened 2022-09-07
add alerting rule to grafana

kdelee:
##### SUMMARY

This rule alerts if the redis queue is larger than what the rolling average event insertion rate/second * 120. In other words, if the redis queue is larger than it appears we can process events in two minutes.

It appears it has to meet this condition for 60 seconds to start firing.

Future commits will address how to configure contact points like slack.

shout out to @jainnikhil30 and @rebeccahhh who figured this out in jam session this morning

##### ISSUE TYPE

##### COMPONENT NAME

 - Other


AlanCoding (approved):

kdelee:
I think I'm going to merge this as is, and come back with another PR if we want to automate setting up contact points -- which is going to have to involve templating a config file so we can add slack tokens at run time

PR #12848, opened 2022-09-07
Account for node state of 'unavailable' in the UI.

kialam:
##### SUMMARY

Screen Shot 2022-09-07 at 12 37 13 PM

##### ISSUE TYPE

##### COMPONENT NAME

 - UI

### DIFF
--- a/awx/ui/src/screens/TopologyView/utils/helpers.js
+++ b/awx/ui/src/screens/TopologyView/utils/helpers.js
@@ -44,6 +44,7 @@ export function renderLabelIcons(nodeState) {
     const nodeLabelIconMapper = {
       ready: 'checkmark',
       installed: 'clock',
+      unavailable: 'exclaimation',
       'provision-fail': 'exclaimation',
       'deprovision-fail': 'exclaimation',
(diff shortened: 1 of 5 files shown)


awx/ui/src/screens/TopologyView/utils/helpers.js
```diff
@@ -128,7 +130,8 @@ export const generateRandomNodes = (n) => {
       'installed',
       'provision-fail',
       'deprovision-fail',
-    ][getRandomInt(0, 5)];
+      'unavailable',
+    ][getRandomInt(0, 6)];
```
AlexSCorey, line 134: Just curious, what does the getRandomInt do here?

AlexSCorey (approved):

awx/ui/src/screens/TopologyView/utils/helpers.js
kialam: It just returns a random number between 0-6. So the state array randomly returns a state by calling it like so: `['ready', 'installed', etc.][2]`

PR #12851, opened 2022-09-08
Add support of update_on_project_update in inventory_source

Sispheor:
##### SUMMARY
The flag `update_on_project_update  `was missing from `inventory_source `module.

Api ref: https://docs.ansible.com/ansible-tower/latest/html/towerapi/api_ref.html#/Inventories/Inventories_inventories_inventory_sources_create

##### ISSUE TYPE
New or Enhanced Feature

##### COMPONENT NAME
 - Collection

### DIFF
--- a/awx_collection/plugins/modules/inventory_source.py
+++ b/awx_collection/plugins/modules/inventory_source.py
@@ -97,6 +97,10 @@ options:
       description:
         - Refresh inventory data from its source each time a job is run.
       type: bool
+    update_on_project_update:
+      description:
+        - Automatically update the inventory source on project update.
+      type: bool
     update_cache_timeout:
       description:
         - Time in seconds to consider an inventory sync to be current.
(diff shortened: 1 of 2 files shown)


Sispheor:
Pr updated.

- PR body updated
- Flag added to the E2E test

sean-m-sullivan (approved):
This matches the API, LGTM

john-westcott-iv (approved):

john-westcott-iv (changes requested):
Update on project update was actually removed from the API in https://github.com/ansible/awx/commit/052223389239392b32877a9798b5056a0048e049

john-westcott-iv:
Thank you so much for your contribution, it really means a lot to us. After further investigation, it looks like `update_on_project_update` was a field in the modules for quite some time but was recently removed in June of this year from the API and also the collection. At this time we don't want to add the parameter back in because the version of the collection should match the version of the API it was released with. Please let us know if you have a different opinion about this. If not, we will close this PR.

Sispheor:
Indeed it has been removed... And so this PR is not needed.
The new behavior based only on "update_no_launch" is not very good but I guess that we'll have to live with it.

Thanks for your time anyway guys. I close this one.

PR #12856, opened 2022-09-08
add firewall rules to control node

fosterseth:
##### SUMMARY

Adds firewall rules to prevent nodes on mesh from targeting the control service on control nodes. 

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API


TheRealHaoLiu (approved):
lgtm

PR #12865, opened 2022-09-09
Update Kind Docs

thedoubl3j:
##### SUMMARY

Update some formatting as well as updating some commands that I found along the way while testing on fedora. 

##### ISSUE TYPE

##### COMPONENT NAME

 - Docs

##### AWX VERSION

```
awx: 21.5.1.dev74+gabca78a893
```

##### ADDITIONAL INFORMATION

Also changed image URLs as they are not coming from GH but rather quay.

### DIFF
--- a/docs/development/kind.md
+++ b/docs/development/kind.md
@@ -1,7 +1,7 @@
 # Running Development Environment in Kubernetes using Kind Cluster
 
 ## Start Kind Cluster
-Note: This environment has only been tested on MacOS with Docker.
+Note: This environment has been tested on MacOS and Fedora with Docker. 
 
 If you do not already have Kind, install it from:
 https://kind.sigs.k8s.io/docs/user/quick-start/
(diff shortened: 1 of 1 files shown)


jbradberry (approved):

gundalow (approved):
Great to see docs updated when people run though them, thanks.

fosterseth:
@thedoubl3j thanks!

the `kind.md` and `minikube.md` have a lot of overlap. Personally I'd rather see a single document with both kind and minikube instructions. Otherwise I fear things will go stale in one or the other as we add more instructions.

however, if we plan to add more k8s cluster targets (like k3s), then having separate documents might be best.

thedoubl3j:
@fosterseth I am not opposed to condensing down into one. I know a lot of what was in the minikube one made it over into this one (some of which i was going to fix in another PR) but I am not against. can't speak on the future cluster targets but happy to go either way.

fosterseth (approved):

TheRealHaoLiu:
@fosterseth i dont think we need to be DRY with the docs 

i would prefer being succinct and context specific

PR #12866, opened 2022-09-12
Add unit tests to remove instance button component

AlexSCorey:
##### SUMMARY
This adds UI unit tests the the new Remove Instance Button component.

##### ISSUE TYPE

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui/src/screens/Instances/Shared/RemoveInstanceButton.test.js
+++ b/awx/ui/src/screens/Instances/Shared/RemoveInstanceButton.test.js
@@ -1,0 +1,133 @@
+import React from 'react';
+import { within, render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { InstanceGroupsAPI } from 'api';
+import RemoveInstanceButton from './RemoveInstanceButton';
+import { I18nProvider } from '@lingui/react';
+import { i18n } from '@lingui/core';
+import { en } from 'make-plural/plurals';
+import english from '../../../../src/locales/en/messages';
+
+jest.mock('api');
+
+const instances = [
+  {
+    id: 1,
+    type: 'instance',
+    url: '/api/v2/instances/1/',
+    capacity_adjustment: '0.40',
+    version: '13.0.0',
+    capacity: 10,
+    consumed_capacity: 0,
+    percent_capacity_remaining: 60.0,
+    jobs_running: 0,
+    jobs_total: 68,
+    cpu: 6,
+    node_type: 'execution',
+    node_state: 'ready',
+    memory: 2087469056,
+    cpu_capacity: 24,
+    mem_capacity: 1,
+    enabled: true,
+    managed_by_policy: true,
+  },
+  {
+    id: 2,
+    type: 'instance',
+    url: '/api/v2/instances/2/',
+    capacity_adjustment: '0.40',
+    version: '13.0.0',
+    capacity: 10,
+    consumed_capacity: 0,
+    percent_capacity_remaining: 60.0,
+    jobs_running: 0,
+    jobs_total: 68,
+    cpu: 6,
+    node_type: 'control',
+    node_state: 'ready',
+    memory: 2087469056,
+    cpu_capacity: 24,
+    mem_capacity: 1,
+    enabled: true,
+    managed_by_policy: false,
+  },
+];
+describe('<RemoveInstanceButtton />', () => {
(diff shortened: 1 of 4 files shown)


kialam (changes requested):
There are several things to point out:
1. Mixing Enzyme and RTL methods in `enzymeHelpers.js` seems counter-intuitive. If we need to create a context mapper/render using RTL since we are moving away from Enzyme, then that should be a separate method in a separate file. Maybe called `rtlHelpers.js` or something to that effect.
2. Components being used in Screens often do not need to be mounted with any React Context Providers and can (and should) be tested in isolation.
3. To move this forward I would remove the changes to `enzymeHelpers.js` and adjust the first test case to be rendered without Contexts.

AlexSCorey:
> There are several things to point out:
> 
> 1. Mixing Enzyme and RTL methods in `enzymeHelpers.js` seems counter-intuitive. If we need to create a context mapper/render using RTL since we are moving away from Enzyme, then that should be a separate method in a separate file. Maybe called `rtlHelpers.js` or something to that effect.

I've removed the changes in this file.  If we decide we want a help function for contexts with rtl tests we can add it again at a later time.  I do not suspect it will have a significant number of use cases.
> 2. Components being used in Screens often do not need to be mounted with any React Context Providers and can (and should) be tested in isolation.

The first test in that file does need the i18n context so that when the modal opens, in the test, the `<Plural/>` component can handle the strings. The other tests do not need it.  


> 3. To move this forward I would remove the changes to `enzymeHelpers.js` and adjust the first test case to be rendered without Contexts.

Removed.

awx/ui/src/screens/Instances/Shared/RemoveInstanceButton.test.js
```diff
@@ -0,0 +1,133 @@
...
+    );
+
+    const button = screen.getByRole('button');
+    await user.hover(button);
+    await waitFor(() =>
+      screen.getByText('You do not have permission to remove instances:')
+    );
+  });
+
+  test('Should handle error when fetching warning message details.', async () => {
```
kialam, line 105: If you want to test how we handle PATCH/POST errors after the user has clicked the button, then I would rewrite this test as the following:
1. Attempt to remove an instance (e.g. click the button)
2. Mock the `InstancesAPI` return value to give an Error
3. Check that the error shows up in the UI

awx/ui/src/screens/Instances/Shared/RemoveInstanceButton.test.js
```diff
@@ -0,0 +1,133 @@
...
+    await user.click(removeButton);
+
+    await waitFor(() => expect(onRemove).toBeCalled());
+  });
+
+  test('Should be disabled', async () => {
+    const user = userEvent.setup();
+    render(
+      <RemoveInstanceButton
+        isK8s={true}
```
kialam, line 92: I might be missing something here but if the remove button is disabled then shouldn't `isK8s === false`? Or is there some other condition that allows this test to pass?

awx/ui/src/screens/Instances/Shared/RemoveInstanceButton.test.js
```diff
@@ -0,0 +1,133 @@
...
+    node_state: 'ready',
+    memory: 2087469056,
+    cpu_capacity: 24,
+    mem_capacity: 1,
+    enabled: true,
+    managed_by_policy: false,
+  },
+];
+describe('<RemoveInstanceButtton />', () => {
+  test('Should open modal and deprovision node', async () => {
```
kialam, line 56: This is an excellent test case. I like how you simulated a user interaction with the component being tested and wrote your assertions to check for relevant component state updates and function callbacks.

awx/ui/src/screens/Instances/Shared/RemoveInstanceButton.test.js
AlexSCorey: In this case, the button is disabled because the selected instance to be removed is `node_type === control`

awx/ui/src/screens/Instances/Shared/RemoveInstanceButton.test.js
AlexSCorey: This specific test is ensuring that we handle errors when we are gathering info about instance_groups that use the instance to be deprovisioned.

kialam (approved):
These tests look good. Nice work.

PR #12869, opened 2022-09-13
UI handle async health checks

kialam:
##### SUMMARY

Adding UI toast notification for health checks. Also spoke with @trahman73 and UXD crew, we decided to simplify the Instances > Peers tab to avoid redundancy.

Screen Shot 2022-09-12 at 3 15 28 PM
Screen Shot 2022-09-12 at 3 15 36 PM
Screen Shot 2022-09-12 at 3 26 13 PM
Screen Shot 2022-09-12 at 6 33 39 PM

Simplified Peers tab:

Screen Shot 2022-09-12 at 6 32 02 PM

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - UI

### DIFF
--- a/awx/ui/src/components/HealthCheckAlert/HealthCheckAlert.js
+++ b/awx/ui/src/components/HealthCheckAlert/HealthCheckAlert.js
@@ -1,0 +1,26 @@
+import React from 'react';
+import { t } from '@lingui/macro';
+import { Alert, Button, AlertActionCloseButton } from '@patternfly/react-core';
+
+function HealthCheckAlert({ onSetHealthCheckAlert }) {
+  return (
+    <Alert
+      variant="default"
+      actionClose={
+        <AlertActionCloseButton onClose={() => onSetHealthCheckAlert(false)} />
+      }
+      title={
+        <>
+          {t`Health check request(s) submitted. Please wait and reload the page.`}{' '}
+          <Button
+            variant="link"
+            isInline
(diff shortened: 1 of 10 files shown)


awx/ui/src/components/HealthCheckAlert/HealthCheckAlert.js
```diff
@@ -0,0 +1,34 @@
...
+      actionClose={
+        <AlertActionCloseButton onClose={() => onSetHealthCheckAlert(false)} />
+      }
+      title={
+        <>
+          {t`Health check request(s) submitted. Please wait and reload the page.`}{' '}
+          <Button
+            variant="link"
+            isInline
+            onClick={() => window.location.reload(false)}
```
AlexSCorey:
```suggestion
            onClick={() => window.location.reload()}
```
Based on this [doc](https://developer.mozilla.org/en-US/docs/Web/API/Location/reload) a parameter is not supported here

AlexSCorey (approved):
I pulled this down and tested it out.  It works as expected.  The only suggestion I have, which isn't a blocker, would be to remove that argument on the window.location.reload().  Good work.

awx/ui/src/components/HealthCheckAlert/HealthCheckAlert.js
kialam: Good catch. I've removed it.

PR #12871, opened 2022-09-13
[feature-prompts] Standardize RBAC logic for new prompts

AlanCoding:
##### SUMMARY
This is mainly about doing DRY edits, but I hope there are a few bugs that will get swept up in this too.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/access.py
+++ b/awx/main/access.py
@@ -1947,14 +1947,14 @@ class WorkflowJobTemplateNodeAccess(UnifiedCredentialsMixin, BaseAccess):
         else:
             return self.user in obj.workflow_job_template.admin_role
 
-    def ujt_execute(self, obj):
+    def ujt_execute(self, obj, data=None):
         if not obj.unified_job_template:
             return True
-        return self.check_related('unified_job_template', UnifiedJobTemplate, {}, obj=obj, role_field='execute_role', mandatory=True)
+        return self.check_related('unified_job_template', UnifiedJobTemplate, data, obj=obj, role_field='execute_role', mandatory=True)
 
     def can_change(self, obj, data):
         # should not be able to edit the prompts if lacking access to UJT or WFJT
-        return self.ujt_execute(obj) and self.wfjt_admin(obj) and JobLaunchConfigAccess(self.user).can_change(obj, data)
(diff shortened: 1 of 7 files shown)


awx/main/access.py
```diff
@@ -328,6 +336,8 @@ def user_has_resource_access(resource):
             role = getattr(resource, role_field, None)
             if role is None:
                 # Handle special case where resource does not have direct roles
+                if role_field == 'read_role':
+                    return self.user.can_access(type(resource), 'read', resource)
```
john-westcott-iv, line 342: Why not just add an entry to the access_method_type field below?

awx/main/access.py
```diff
@@ -1833,184 +1839,60 @@ class JobLaunchConfigAccess(BaseAccess):
...
-                        permission_error = True
-                    else:
-                        logger.debug(
-                            "User {} does not have permissions to instance_group {} but its part of the template".format(self.user.username, an_ig.name)
-                        )
-        if permission_error:
-            return False
-        return self.check_related('inventory', Inventory, data, role_field='use_role')
+                    # If given model objects, only use the primary key from them
+                    sub_obj_pks = [cred.pk for cred in data[relationship]]
```
john-westcott-iv: Do we want to remove references to cred in here since it can really be any of the M2M items?

awx/main/access.py
AlanCoding: Because the method signature is different, b/c the "read" check does not take the `data` arg or kwarg.

awx/main/access.py
```diff
@@ -1833,184 +1841,63 @@ class JobLaunchConfigAccess(BaseAccess):
     In order to create a new object with a copy of this launch config, I need:
      - use access to related inventory (if present)
      - use role to many-related credentials (if any present)
-     - use role to Execution Environment (if present), unless the specified ee is already in the template
-     - use role to many-related labels (if any present), unless the specified label is already in the template
-     - use role to many-related instance groups (if any present), unless the specified instance group is already in the template
```
AlanCoding, line 1838: These two are not use-level access.

awx/main/access.py
AlanCoding: yes, thanks

john-westcott-iv (approved):

PR #12875, opened 2022-09-13
Feature prompt on launch on templates

john-westcott-iv:
##### SUMMARY

Enable prompting on launch of the following items in Job Templates:
* Execution Environment
* Labels
* Forks
* Job Slice Count
* Timeout
* Instance Groups

Enable prompting on launch for the following items in Workflow Job Templates:
* Labels
* Skip Tags
* Job Tags

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 - UI
 - Collection
 - CLI

##### AWX VERSION

```
awx: 21.5.1.dev132+g2dfdf504f8
```

##### ADDITIONAL INFORMATION

```

```


AlexSCorey:
Need QE re: @mabashian

awx/api/serializers.py
```diff
@@ -3035,6 +3041,9 @@ class Meta:
             'webhook_service',
             'webhook_credential',
             'webhook_guid',
+            # TODO: Do we want these here or just in the summary fields?
+            'labels',
+            'instance_groups',
```
AlanCoding: I think delete. I'm not even really sure of the effect right now.

awx/api/serializers.py
```diff
@@ -3061,6 +3070,11 @@ def get_related(self, obj):
                 res['project_update'] = self.reverse('api:project_update_detail', kwargs={'pk': obj.project_update.pk})
         except ObjectDoesNotExist:
             pass
+        try:
+            if obj.instance_groups:
+                res['instance_groups'] = self.reverse('api:job_instance_group_list', kwargs={'pk': obj.pk})
+        except ObjectDoesNotExist:
+            pass
```
AlanCoding: I don't think this is the right way to use `ObjectDoesNotExist`, because it's a relationship, and `obj.instance_groups` should always exist, so better to get rid of the try-except.

awx/api/serializers.py
```diff
@@ -3443,12 +3467,13 @@ def get_related(self, obj):
     def validate_extra_vars(self, value):
         return vars_validate_or_raise(value)
 
+    # posting
```
AlanCoding: delete comment

awx/api/serializers.py
```diff
@@ -3609,6 +3639,10 @@ class LaunchConfigurationBaseSerializer(BaseSerializer):
     skip_tags = serializers.CharField(allow_blank=True, allow_null=True, required=False, default=None)
     diff_mode = serializers.BooleanField(required=False, allow_null=True, default=None)
     verbosity = serializers.ChoiceField(allow_null=True, required=False, default=None, choices=VERBOSITY_CHOICES)
+    execution_environment = serializers.PrimaryKeyRelatedField(queryset=ExecutionEnvironment.objects.all(), required=False, allow_null=True, default=None)
```
AlanCoding: This follows the same pattern as `inventory`, and I don't think it needs to be defined on the serializer, because the DRF serializer does everything we need for this.

awx/api/serializers.py
```diff
@@ -3624,13 +3658,21 @@ class Meta:
...
 
     def get_related(self, obj):
         res = super(LaunchConfigurationBaseSerializer, self).get_related(obj)
         if obj.inventory_id:
             res['inventory'] = self.reverse('api:inventory_detail', kwargs={'pk': obj.inventory_id})
         res['credentials'] = self.reverse('api:{}_credentials_list'.format(get_type_for_model(self.Meta.model)), kwargs={'pk': obj.pk})
+        res['labels'] = self.reverse('api:{}_labels_list'.format(get_type_for_model(self.Meta.model)), kwargs={'pk': obj.pk})
+        res['instance_groups'] = self.reverse('api:{}_instance_groups_list'.format(get_type_for_model(self.Meta.model)), kwargs={'pk': obj.pk})
+        if obj.execution_environment_id:
+            res['execution_environment'] = self.reverse('api:execution_environment_detail', kwargs={'pk': obj.execution_environment_id})
```
AlanCoding, line 3664: the ordering is probably inconsistent. It would be good if we could formalize the order of the 5 - inventory, EE, labels, credentials, IGs

awx/api/serializers.py
```diff
@@ -3640,7 +3682,11 @@ def _build_mock_obj(self, attrs):
                 setattr(mock_obj, field.name, getattr(self.instance, field.name))
         field_names = set(field.name for field in self.Meta.model._meta.fields)
         for field_name, value in list(attrs.items()):
-            setattr(mock_obj, field_name, value)
+            if field_name == 'execution_environment':
```
AlanCoding: I still can't figure out why the EE should have an exception case here.

awx/api/views/__init__.py
```diff
@@ -208,6 +209,26 @@ def api_exception_handler(exc, context):
     return exception_handler(exc, context)
 
 
+class LabelList(DeleteLastUnattachLabelMixin, SubListCreateAttachDetachAPIView):
+    def post(self, request, *args, **kwargs):
+        # If a label already exists in the database, attach it instead of erroring out
+        # that it already exists
+        if not getattr(self, 'label_filter', None):
+            return Response(dict(msg=_('Class {} missing label filter.'.format(self.__class__.__name__))), status=status.HTTP_400_BAD_REQUEST)
```
AlanCoding: Again, I would like to touch this up as this is never an expected response.

awx/main/models/jobs.py
```diff
@@ -965,8 +1017,12 @@ class Meta:
             blank=True,
         )
     )
-    # Credentials needed for non-unified job / unified JT models
+    # Fields needed for non-unified job / unified JT models, because they are defined on unified models
     credentials = models.ManyToManyField('Credential', related_name='%(class)ss')
+    labels = models.ManyToManyField('Label', related_name='%(class)s_labels')
+    execution_environment = models.ForeignKey(
+        'ExecutionEnvironment', null=True, blank=True, default=None, on_delete=polymorphic.SET_NULL, related_name='%(class)s_as_prompt'
```
AlanCoding: As this is a `ForeignKey`, not a many-related field, it would be really nice to have help_text for OPTIONS, especially to echo the adherence to its `ask_` rule.

awx/main/tests/functional/models/test_job_launch_config.py
```diff
@@ -61,7 +62,42 @@ def test_char_field_only(self, config_factory):
...
+            'limit': 'foobar',
+            'inventory': inventory,
+            'credentials': [credential],
+            'execution_environment': ee,
+            'labels': [label],
+            'instance_groups': [default_instance_group],
+        }
         config = config_factory(prompts)
+        print(prompts)
+        print(config.prompts_dict())
```
AlanCoding: delete print statements

awx/main/models/jobs.py
john-westcott-iv: This field is on the JobLaunchConfig the JobTemplate endpoint gets the POST options from ExecutionEnvironmentMixin. Is there somewhere in the API that you can see job launch configs?

awx/ui/src/components/LaunchPrompt/steps/OtherPromptsStep.js
```diff
@@ -22,72 +24,105 @@ const FieldHeader = styled.div`
 `;
 
 function OtherPromptsStep({ launchConfig, variablesMode, onVarModeChange }) {
+  const helpTextSource = launchConfig.job_template_data
```
AlexSCorey, line 27: This is neat

awx/main/models/jobs.py
AlanCoding: Schedules and WFJT nodes _are_ `JobLaunchConfig`s. It is an abstract model, but that just means that it has content shared by other concrete models. Those I list are not the _only_ concrete model that subclasses this and inherits its fields, but they are the easiest examples because they are listed and editable in the API the same way other normal resources are.

awx/api/views/__init__.py
```diff
@@ -208,6 +209,26 @@ def api_exception_handler(exc, context):
     return exception_handler(exc, context)
 
 
+class LabelList(DeleteLastUnattachLabelMixin, SubListCreateAttachDetachAPIView):
```
AlanCoding: @john-westcott-iv this has a re-definition of `LabelList` at line 4501.

that should have been picked up by the linter, but we selectively enable rules b/c our first line of defense is black. But re-definition is a big one that we should capture.

awx/api/views/__init__.py
```diff
@@ -617,6 +638,23 @@ class ScheduleCredentialsList(LaunchConfigCredentialsBase):
     parent_model = models.Schedule
 
 
+class ScheduleLabelsList(LabelList):
```
AlanCoding: This view is a sublist-create-attach view. That means that you can create a label by

POST to `/api/v2/schedule/N/labels/`

@mabashian does the UI use this to create labels? Or does it go create them by POST to `/api/v2/labels/`? If the UI does not use it, we should change the views from a create-attach-sublist to just an attach-sublist type view. Otherwise we have this un-tested creation mechanism hanging around, and a multitude of things may be wrong about it.

~Same goes for the instance groups sublist.~

EDIT: instance groups sublist does not allow creating an instance group, so that's what we want. :+1: to that.

EDIT: the answer I am presenting in the related PR is that label sublists _should_ allow creating labels, and I am consolidating the view code for this so WFJT node & schedule label-lists work the same as other editable label-lists.

awxkit/awxkit/api/pages/workflow_job_templates.py
```diff
@@ -42,6 +42,9 @@ def payload(self, **kwargs):
             "ask_inventory_on_launch",
             "ask_scm_branch_on_launch",
             "ask_limit_on_launch",
+            "ask_labels_on_launch",
+            "ask_skip_tags_on_launch",
+            "ask_tags_on_launch",
             "limit",
             "scm_branch",
             "survey_enabled",
```
AlanCoding, line 50: Action item:

Also in this module, the WFJT model has a `add_label` method. The WFJT node should _also_ have this method, because this PR adds labels on it.

Same goes for labels on schedules, the awxkit Schedule model should get a new `add_label` method.

----

Instance groups was also an added endpoint, and `add_instance_group` is a method on some models. This method should be added to WFJT nodes and schedules as well.

obaranov: I'll do that

obaranov: add_label PR: https://github.com/ansible/awx/pull/12886

awx/api/serializers.py
john-westcott-iv: I think initially I was thinking incase an IG got deleted but since they are FK constraints that wouldn't matter, updated.

awx/api/serializers.py
john-westcott-iv: Removed.

awx/api/serializers.py
john-westcott-iv: Order updated.

awx/api/serializers.py
john-westcott-iv: Removed.

awxkit/awxkit/api/pages/workflow_job_templates.py
john-westcott-iv: Resolving from merge

awx/main/models/jobs.py
john-westcott-iv: Good call on the inheritance from schedules. The schedules endpoint was not display help text. That has been resolved.

obaranov:
I have test failures with /api/v2/jobs/142/create_schedule endpoint (test_prompt_ee_on_schedule test). I start job with an execution environment prompt. And then calling create_schedule produces the

`tools_awx_1     | TypeError: Object of type ExecutionEnvironment is not JSON serializable`

Here is simple pseudo code to reproduce: 
```

        jt = factories.job_template(ask_execution_environment_on_launch=True)
        job = jt.launch(dict(execution_environment=ee_copy.id)).wait_until_completed()
        create_schedule = job.related.create_schedule.get()
        assert create_schedule.prompts == dict(execution_environment=ee_copy.id)
        schedule = create_schedule.post()
```
The same happens for instance group prompt.

john-westcott-iv:
We need to review `is_detached`, `get_ophaned_labels` and `is_candidate_for_detach` from the Labels model to make sure they account for the new relations to schedules and wfjtnodes and any other objects.

sarabrajsingh:
https://github.com/ansible/awx/pull/12888 - needs to be scoped in here as well

awx/ui/src/components/LaunchButton/LaunchButton.js
```diff
@@ -60,6 +71,27 @@ function LaunchButton({ resource, children }) {
...
+        setLabels(allLabels);
+      }
+
+      if (launch.ask_instance_groups_on_launch) {
+        const {
+          data: { results },
+        } = await JobTemplatesAPI.readInstanceGroups(resource.id);
+
+        setInstanceGroups(results);
+      }
```
keithjgrant: These three requests (`readSurvey`, `readLabels` and `readInstanceGroups`) are made in serial. It would be more efficient to get both promises before awaiting the results (though it will require extra `if` blocks at the end before setting state vars).

awx/ui/src/components/LaunchPrompt/LaunchPrompt.js
```diff
@@ -53,6 +62,61 @@ function PromptModalForm({
...
+            organization: organizationId,
+          }).then(({ data }) => {
+            labelIds.push(data.id);
+          })
+        );
+      });
+
+      await Promise.all(labelRequests);
+
+      setValue('labels', labelIds);
```
keithjgrant: Consider moving most or all of this into a helper function to keep `handleSubmit` down to a scanable length

awx/main/models/jobs.py
```diff
@@ -314,10 +331,16 @@ def get_effective_slice_ct(self, kwargs):
         actual_inventory = self.inventory
         if self.ask_inventory_on_launch and 'inventory' in kwargs:
             actual_inventory = kwargs['inventory']
+        actual_slice_count = self.job_slice_count
+        if self.ask_job_slice_count_on_launch and 'job_slice_count' in kwargs:
+            actual_slice_count = kwargs['job_slice_count']
+            # Set the eager fields if its there as well
+            if '_eager_fields' in kwargs and 'job_slice_count' in kwargs['_eager_fields']:
+                kwargs['_eager_fields']['job_slice_count'] = actual_slice_count
```
AlanCoding: As a coding practice... obviously I would not like a `get_*` method to _set_ something like this.

But it's much more troublesome that there was some reason that caused this to need to get set here in the first place. This would seem to indicate that `job_slice_count` doesn't get copied on launch to the job. This should happen in `create_unified_job` irrelevant of the prompts.

```
In [8]: set(JobTemplate.get_ask_mapping().keys()) - set(WorkflowJobTemplate._get_unified_job_field_names())
Out[8]: 
{'credentials',
 'diff_mode',
 'execution_environment',
 'forks',
 'instance_groups',
 'job_slice_count',
 'job_type',
 'timeout',
 'verbosity'}
```

I expect some of these probably need to be added to the workflow `_get_unified_job_field_names`, because they are valid workflow fields.

awx/main/utils/common.py
```diff
@@ -531,7 +531,16 @@ def copy_m2m_relationships(obj1, obj2, fields, kwargs=None):
...
+                    if field_name == 'instance_groups':
+                        # instance_groups are a list but we need to preserve the order
+                        for ig_id in override_field_val:
+                            getattr(obj2, field_name).add(ig_id)
+                        continue
                     if isinstance(override_field_val, (set, list, QuerySet)):
+                        # Labels are additive so we are going to add any src labels in addition to the override labels
+                        if field_name == 'labels':
+                            for jt_label in src_field_value.all():
+                                getattr(obj2, field_name).add(jt_label.id)
```
AlanCoding, line 538: Merging of credentials in this process is handled here:

https://github.com/ansible/awx/blob/063b0d08343938295a3c8f6ccb5342ffd2421d40/awx/main/models/unified_jobs.py#L402-L409

Yes, all 3 relationships, instance_groups, labels, and credentials have distinct rules associated with them. But handing 1 relationship _before_ the method `copy_m2m_relationships` call, and 2 relationships _inside_ the method is going to make this much harder for anyone to ever follow in the future.

To pick one, I would say it should be done before the method call, because that can be adjusted to be more testable. That still leaves you with an exception case for _saving_ instance groups, which would have to remain here.

awx/main/models/unified_jobs.py
```diff
@@ -973,22 +977,42 @@ def create_config_from_prompts(self, kwargs, parent=None):
...
+                    getattr(config, field_name).add(*job_item)
+            else:
+                # Here we are doing a loop to make sure we preserve order in case this is a Ordered field
+                job_item = kwargs.get(field_name, [])
+                if job_item:
+                    parent_items = list(getattr(parent, field_name, []).all())
+                    for item in job_item:
+                        # Do not include this item in the config if its in the parent
+                        if item not in parent_items:
+                            getattr(config, field_name).add(item)
```
AlanCoding: The "Ordered field" block here can't be right - it contradicts the prompting rules we wrote for `instance_groups`.

If the launch-time list is not `[]`, then _the entire list_ constitutes the prompted value. This loop obtains any items _added relative to the JT_ and that's completely irrelevant of what it actually does. _Even if the entire list matches the JT_, if you provide it on launch, that gives a different meaning. The JT `instance_groups` says "use these groups, but fallback to inventory/org". The prompted list says "_only_ use these groups".

Putting it that way, it's also clear that the `default` value offered in the `JobLaunchSerializer` is wrong, because it returns the JT instance groups. Using that as-is would use that same list, but with a different meaning to it. So either the entire construct of the `default` should be declared invalid for this field, or the view should present `"default": []` or `"default": null`, because this is what obtains parity with the basic JT behavior.


While the prompts store is only used for the obscure purposes of relaunch or the "create_schedule", this makes obvious another more likely and severe bug. The method `create_unified_job` takes `kwargs` with an unclear meaning. Even if the UI gives `[]`, that doesn't mean we're off the hook because `copy_m2m_relationships` takes _any_ value in kwargs to be an _override_, including a falsy value, so it's not even clear that our our intended not-provided value will leave the JT values untouched.

I'm sure testing is at high risk of missing these things, because we are very light on testing (if at all) filling in JT values and then assuring launch works _without_ providing prompts, when prompts are accepted. Even if we did, integration tests use its own client, and it's not clear even what payload to the launch POST they are even _supposed_ to provide for this case, because in the request data, you can _exclude_ the key, set to _null_, set to `[]`, or set to the "default" values. I don't think we have ever touched that testing matrix.

AlanCoding: I don't want to leave this too vague...

I'm saying that, as you have already added special logic in `copy_m2m_relationships`, you need more special logic. Before you start the loop `for ig_id in override_field_val:` in the special IG logic, it needs to be checked that the data constitute an override in the first place.... like, indent that into a `if override_field_val:` conditional. But I'd hate to jump right into that without first characterizing relevant test cases.

AlanCoding: Oh, and this is actually impossible / impractical to test in integration tests. Only unit tests, because the resolved list is never reported by the API.

AlanCoding: Scratch some of that, don't take my prior advice. In the event that IGs are not provided, IIRC, we do not want to copy the JT values at all. Again, this is philosophically very problematic with `copy_m2m_relationships`, and it should probably be done before that, but using the existing code structure, that method should _always_ `continue` on the instance_groups field to avoid a copy of the JT list. I do need to articulate a specific test case for this.

awx/main/tests/functional/api/test_job_runtime_params.py
```diff
@@ -157,11 +189,28 @@ def test_job_accept_empty_tags(job_template_prompts, post, admin_user, mocker):
         with mocker.patch('awx.api.serializers.JobSerializer.to_representation'):
             post(reverse('api:job_template_launch', kwargs={'pk': job_template.pk}), {'job_tags': '', 'skip_tags': ''}, admin_user, expect=201)
             assert JobTemplate.create_unified_job.called
-            assert JobTemplate.create_unified_job.call_args == ({'job_tags': '', 'skip_tags': ''},)
+            assert JobTemplate.create_unified_job.call_args == ({'job_tags': '', 'skip_tags': '', 'forks': 1, 'job_slice_count': 0},)
```
AlanCoding: @john-westcott-iv going through tests, I was surprised by these numbers. This doesn't look desirable, I would prefer these files to default to `None` / not-provided.

awx/main/models/unified_jobs.py
AlanCoding: We discussed this and I am developing a potential patch.

https://github.com/ansible/awx/pull/12925

Needs work post-migration merge.

awx/main/tests/functional/api/test_job_runtime_params.py
AlanCoding: addressed in https://github.com/ansible/awx/pull/12917

awx/ui/src/components/LaunchPrompt/LaunchPrompt.js
mabashian: Good call - there's identical code in the Schedule Edit component so I'll pull this all out and put it somewhere where it can be imported.

awx/ui/src/components/LaunchButton/LaunchButton.js
mabashian: 👍

AlanCoding (approved):

AlanCoding:
https://github.com/ansible/awx/compare/devel...AlanCoding:squash_attempt3?expand=1

 if you want squashed

PR #12878, opened 2022-09-14
Prevent altering certain fields on Instance

fosterseth:
##### SUMMARY

- Prevents changing hostname, listener_port, or node_type for instances
that already exist
- API default node_type is execution
- API default node_state is installed

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -4778,12 +4778,12 @@ class InstanceSerializer(BaseSerializer):
         read_only_fields = ('ip_address', 'uuid', 'version')
         fields = (
             'id',
+            'hostname',
             'type',
             'url',
             'related',
             'summary_fields',
             'uuid',
-            'hostname',
             'created',
             'modified',
             'last_seen',
@@ -4807,6 +4807,7 @@ class InstanceSerializer(BaseSerializer):
             'ip_address',
             'listener_port',
         )
+        extra_kwargs = {'node_type': {'default': 'execution'}, 'node_state': {'default': 'installed'}}
 
     def get_related(self, obj):
         res = super(InstanceSerializer, self).get_related(obj)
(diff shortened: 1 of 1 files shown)


awx/api/serializers.py
```diff
@@ -4868,6 +4868,24 @@ def validate_node_state(self, value):
...
+        return value
+
+    def validate_listener_port(self, value):
+        if self.instance and self.instance.listener_port != value:
+            raise serializers.ValidationError("Cannot change listener port.")
+
+        return value
+
+
+class InstanceUpdateSerializer(InstanceSerializer):
```
jbradberry: This shouldn't be necessary, IMO.  `hostname` and `listener_port` are covered by your `validate_*` methods above, and `node_type` was already handled previously.

awx/main/migrations/0167_node_and_link_state.py
```diff
@@ -45,7 +45,7 @@ class Migration(migrations.Migration):
                     ('deprovisioning', 'De-provisioning'),
                     ('deprovision-fail', 'De-provisioning Failure'),
                 ],
-                default='ready',
+                default='installed',
```
jbradberry: I'm not sure we want to be doing this here.  If you look a few lines above, you'll see that this is where we bring in the field in the first place.  If we make the default `'installed'` here, people's existing infrastructure will get marked as such when they do an upgrade.  Maybe that's not such a bad thing, under most circumstances they'll transition normally once the advertisements propagate.  But then again, maybe not, and jobs now can't be run by nodes in any state other than `'ready'`.

I was leaning heavily towards just using `extra_kwargs` on the serializer to shim in a different default.

awx/api/serializers.py
fosterseth: fixed

awx/main/migrations/0167_node_and_link_state.py
fosterseth: k now I just do with extra_kwargs, db fields remain the same

jbradberry (approved):

PR #12884, opened 2022-09-14
[tech debt] Move the IS_TESTING method out of settings

AlanCoding:
##### SUMMARY
It is _really weird_ to find this function definition inside of settings, firstly. It also comes with some cost - as I looked into the logic for taking out _settings snapshot_ and `IS_TESTING` gets put into the snapshot by the `copy.deepcopy` and it is the only thing that isn't a plain int/string/bool type, or a list/dict of those types.

Doing some historical analysis, I found that the introduction corresponded to a special `postprocess.py` script for development.

https://github.com/ansible/awx/commit/60224cdbe4845f2e80b38486b62fd8f5feedfca0#diff-bd08a936903e8f4f8830bf6d0caa6ff3520d75ec50a62c594b32c4a33e800912

This makes sense, and it _would_ be why you would introduce a method like this in settings. However, no trace of that remains today.

All of the uses I see now seem to actually be relatively modern, as new code found this method and applied it in an _ordinary_ type place. These uses have no need for such an unorthodox definition inside of settings, and there are oodles of reasons to be concerned with all the layers of wrapping we have on top of settings.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
This is motivated by @shanemcd 's recent work to clean up settings.

For merging, one obvious criteria will be that the tests pass, because this is _mostly_ important as a metric for turning things off when tests are running. The other thing we want is to make sure that nothing obvious breaks.

### DIFF
--- a/awx/main/utils/common.py
+++ b/awx/main/utils/common.py
@@ -11,11 +11,12 @@ import os
 import subprocess
 import re
 import stat
+import sys
 import urllib.parse
 import threading
 import contextlib
 import tempfile
-from functools import reduce, wraps
+import functools
 
 # Django
 from django.core.exceptions import ObjectDoesNotExist, FieldDoesNotExist
@@ -73,6 +74,7 @@ __all__ = [
     'NullablePromptPseudoField',
     'model_instance_diff',
     'parse_yaml_or_json',
+    'is_testing',
     'RequireDebugTrueOrTest',
     'has_model_field_prefetched',
     'set_environ',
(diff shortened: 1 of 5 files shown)


AlanCoding:
I tried to measure how much speed boost this was compared to before (due to the deepcopy) and it was like 0.005 seconds. So nothing to get excited about.

awx/main/utils/common.py
```diff
@@ -144,6 +146,18 @@ def underscore_to_camelcase(s):
     return ''.join(x.capitalize() or '_' for x in s.split('_'))
 
 
+def is_testing(argv=None):
```
shanemcd, line 150: Could we use `@functools.cache` here?

AlanCoding: yes, that would be better. We use `cachetools` in other places, I don't have any examples where we use `functools.cache`, but I think it should be safe.

https://github.com/python/cpython/blob/f9433fff476aa13af9cb314fcc6962055faa4085/Lib/functools.py#L653

https://github.com/python/cpython/blob/f9433fff476aa13af9cb314fcc6962055faa4085/Lib/functools.py#L549

Since we won't ever _invalidate_ the cache, it could be thread-safe, or at least avoid the `del cache[key]` we hit before.

AlanCoding: I updated to use that, the import was reorganized a bit.

jbradberry (approved):
:fire: :fire: :fire:

PR #12886, opened 2022-09-15
Add new add_label method to the wfjt node and schedules awxkit models

obaranov:
Add add_label method to the awxkit models.

### DIFF
--- a/awxkit/awxkit/api/pages/schedules.py
+++ b/awxkit/awxkit/api/pages/schedules.py
@@ -48,10 +48,12 @@ class Schedules(page.PageList, Schedule):
             self.related.credentials.post(dict(id=cred.id, disassociate=True))
 
     def add_label(self, label):
-        if isinstance(label, page.Page):
-            label = label.json
         with suppress(exc.NoContent):
-            self.related.labels.post(label)
+            self.related.labels.post(dict(id=label.id))
+
+    def add_instance_group(self, instance_group):
+        with suppress(exc.NoContent):
+            self.related.instance_groups.post(dict(id=instance_group.id))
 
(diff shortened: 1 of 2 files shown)


awxkit/awxkit/api/pages/schedules.py
```diff
@@ -47,6 +47,12 @@ def remove_credential(self, cred):
         with suppress(exc.NoContent):
             self.related.credentials.post(dict(id=cred.id, disassociate=True))
 
+    def add_label(self, label):
+        if isinstance(label, page.Page):
```
obaranov, line 51: @AlanCoding what is the better pattern here? Use label object or id? I see for example add_credential method above expects the id.

awxkit/awxkit/api/pages/workflow_job_template_nodes.py
```diff
@@ -32,6 +32,12 @@ def payload(self, workflow_job_template, unified_job_template, **kwargs):
             'extra_data',
             'identifier',
             'all_parents_must_converge',
+            # prompt fields for JTs
+            'job_slice_count',
+            'labels',
```
AlanCoding: I don't think we want labels here, because it's not a direct field, so providing it to the factories won't have the desired effect anyway. Otherwise this all looks great.

AlanCoding (approved):

awxkit/awxkit/api/pages/workflow_job_template_nodes.py
obaranov: Fixed.

AlanCoding (approved):

PR #12887, opened 2022-09-15
[tech debt] Add new flake8 rules to do some meaningful corrections

AlanCoding:
##### SUMMARY
We have a select list of flake8 rules that we enforce. Every now and then, I believe we should revisit those. Here is my method here: I remove all our `select =` rules and run flake8. Then I create an `ignore =` list and add the rules I don't care about or find annoying. I iterate on those until every discovery is something that I really care about and want to fix.

Here are the things I'm adding here:
 - E265	block comment should start with #
 - E266	too many leading # for block comment
 - F541	fstring placeholder
 - W605	invalid escape sequence
 - E722	bare except
 - F822	undefined name
 - F523	format has unused arguments
 - W291	trailing whitespace
 - F405	variable may be undefined

The thing I really _wanted_ to do was find a rule that would find the re-definition of a class inside of https://github.com/ansible/awx/pull/12875, but I failed to do that. However, this found a good handful of cases where we had a reference to a variable that doesn't exist.

This fixes at least a few real bugs where an error message will not render correctly, but I'm guessing that no one hit those errors that that's why we don't otherwise know about these.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
I will keep this as draft for a while, and will probably block merging until other major feature branches have landed.


awx/playbooks/action_plugins/playbook_integrity.py
```diff
@@ -88,7 +87,7 @@ def validate_checksum_manifest(self):
         if not os.path.exists(manifest_file):
             return {
                 "failed": True,
-                "msg": f"Expected file not found: {path}",
+                "msg": f"Expected file not found: {manifest_file}",
```
AlanCoding: @relrod the `path` variable is referenced in other places in the file, but I don't think it's in scope here.

relrod: Yep, copypaste fail. Good catch!

AlanCoding: ok, I think that this PR won't make it in the current development sprint. If anything looks urgent enough that it needs fixing before we ship, you should consider making a minor PR that just makes those changes without the linting rules.

awx/main/migrations/_galaxy.py
```diff
@@ -44,7 +44,7 @@ def migrate_galaxy_settings(apps, schema_editor):
             credential_type=galaxy_type,
             inputs={'url': 'https://galaxy.ansible.com/'},
         )
-    except:
+    except Exception:
```
jbradberry, line 47: :100:

tox.ini
```diff
@@ -16,5 +16,5 @@ commands =
   yamllint -s .
 
 [flake8]
-select = F401,F402,F821,F823,F841,F811
+select = F401,F402,F821,F823,F841,F811,E265,E266,F541,W605,E722,F822,F523,W291,F405
```
jbradberry, line 19: Just to note what's going on here:

- E265: Block comment should start with `'# '`
- E266: Too many leading `'#'` for block comment
- F541: f-string without any placeholders
- W605: Invalid escape sequence 'x' (backslash escapes)
- E722: Do not use bare except, specify exception instead
- F822: undefined name name in `__all__`
- F523: `.format(...)` unused positional arguments
- W291: Trailing whitespace
- F405: Name may be undefined, or defined from star imports: module

I wholeheartedly endorse.

jbradberry (approved):

tox.ini
jbradberry: ... and then I see after the fact that you had already put this info in the main comment on this PR.  :man_facepalming:

AlanCoding: It sure would be nice to have these as a comment in this same file. But then I Google and find an RFE on that closed as wontfix

PR #12888, opened 2022-09-15
adding fix for labels pushdown on workflow job nodes

sarabrajsingh:
fixes labels and instance_group ManyToMany IDs not being pushed down to nodes that are spawned from parent templates.

### DIFF
--- a/awx/main/models/workflow.py
+++ b/awx/main/models/workflow.py
@@ -114,6 +114,9 @@ class WorkflowNodeBase(CreatedModifiedModel, LaunchTimeConfig):
             'credentials',
             'char_prompts',
             'all_parents_must_converge',
+            'labels',
+            'instance_groups',
+            'execution_environment',
         ]
 
     def create_workflow_job_node(self, **kwargs):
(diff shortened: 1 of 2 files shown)


AlanCoding (approved):

awx/main/models/workflow.py
```diff
@@ -114,6 +114,9 @@ def _get_workflow_job_field_names(cls):
             'credentials',
             'char_prompts',
             'all_parents_must_converge',
+            'labels',
```
john-westcott-iv, line 117: Are there any other objects in here that we need add fields to? WorkflowJobTemplateNode?

john-westcott-iv: We also have a failing test.

sarabrajsingh: test is fixed.

sarabrajsingh: > Are there any other objects in here that we need add fields to? WorkflowJobTemplateNode?

we need to think of any object in workflow.py that essentially executes a "copy". i will take a look.

PR #12889, opened 2022-09-15
Complete consolidation of the label views

AlanCoding:
##### SUMMARY
Followup from the comment originally here that `label_filter` was not necessary and the Response was not testable:

https://github.com/ansible/awx/pull/12732#discussion_r967320695

and then next comment here that `LabelList` had a re-definition:

https://github.com/ansible/awx/pull/12875#discussion_r971104275

The other major development here is that `InventoryLabelList` contained an outright copy of some logic, which we attempted to consolidate in the PR 12732.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
Testing looks good so far.

### DIFF
--- a/awx/api/views/labels.py
+++ b/awx/api/views/labels.py
@@ -1,0 +1,71 @@
+# AWX
+from awx.api.generics import SubListCreateAttachDetachAPIView, RetrieveUpdateAPIView, ListCreateAPIView
+from awx.main.models import Label
+from awx.api.serializers import LabelSerializer
+
+# Django
+from django.utils.translation import gettext_lazy as _
+
+# Django REST Framework
+from rest_framework.response import Response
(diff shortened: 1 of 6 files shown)


awx/api/views/labels.py
```diff
@@ -0,0 +1,69 @@
...
+
+    Subclasses need to define parent_model
+    """
+
+    model = Label
+    serializer_class = LabelSerializer
+    relationship = 'labels'
+
+    def unattach(self, request, *args, **kwargs):
+        (sub_id, res) = super(LabelSubListCreateAttachDetachView, self).unattach_validate(request)
```
AlanCoding: I should consistently follow the `super()` syntax here. Will push.

john-westcott-iv (approved):

PR #12893, opened 2022-09-15
Changing label functions to account for new relationships

john-westcott-iv:
##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/models/label.py
+++ b/awx/main/models/label.py
@@ -36,17 +36,6 @@ class Label(CommonModelNameNotUnique):
     def get_absolute_url(self, request=None):
         return reverse('api:label_detail', kwargs={'pk': self.pk}, request=request)
 
-    @staticmethod
-    def get_orphaned_labels():
-        return Label.objects.filter(
-            organization=None,
-            unifiedjobtemplate_labels__isnull=True,
-            inventory_labels__isnull=True,
-            schedule_labels__isnull=True,
-            workflowjobtemplatenode_labels__isnull=True,
-            workflowjobnode_labels=True,
-        )
-
     def is_detached(self):
         return Label.objects.filter(
             id=self.id,
--- a/awx/main/models/label.py
+++ b/awx/main/models/label.py
@@ -10,6 +10,8 @@ from awx.api.versioning import reverse
 from awx.main.models.base import CommonModelNameNotUnique
 from awx.main.models.unified_jobs import UnifiedJobTemplate, UnifiedJob
 from awx.main.models.inventory import Inventory
+from awx.main.models.schedules import Schedule
+from awx.main.models.workflow import WorkflowJobTemplateNode, WorkflowJobNode
 
 __all__ = ('Label',)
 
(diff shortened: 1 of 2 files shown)


awx/main/models/label.py
```diff
@@ -10,6 +10,8 @@
 from awx.main.models.base import CommonModelNameNotUnique
 from awx.main.models.unified_jobs import UnifiedJobTemplate, UnifiedJob
 from awx.main.models.inventory import Inventory
+from awx.main.models.schedules import Schedule
+from awx.main.models.workflow import WorkflowJobTemplateNode, WorkflowJobNode
```
AlanCoding, line 14: It always worries me to add more imports from these kind of sibling modules, because there's not a clear hierarchy / ordering of the models, and can lead to circular imports... or just make it more easy to create circular imports in the future.

awx/main/models/label.py
```diff
@@ -36,14 +38,31 @@ def get_absolute_url(self, request=None):
...
+            schedule_labels__isnull=True,
+            workflowjobtemplatenode_labels__isnull=True,
+            workflowjobnode_labels=True,
+        )
 
     def is_detached(self):
-        return Label.objects.filter(id=self.id, unifiedjob_labels__isnull=True, unifiedjobtemplate_labels__isnull=True, inventory_labels__isnull=True).exists()
+        return Label.objects.filter(
+            id=self.id,
+            unifiedjob_labels__isnull=True,
```
AlanCoding, line 42: All these kwargs are repeated in `get_orphaned_labels` except for this one. Is there are reason for that?

awx/main/tests/unit/models/test_label.py
```diff
@@ -37,39 +45,101 @@ def test_is_detached_not(self, mocker):
...
-        "jt_count,j_count,inv_count,expected",
+        "jt_count,j_count,inv_count,sched_count,wfnode_count,wfnodej_count,expected",
         [
-            (1, 0, 0, True),
-            (0, 1, 0, True),
-            (0, 0, 1, True),
-            (1, 1, 1, False),
+            (1, 0, 0, 0, 0, 0, True),
+            (0, 1, 0, 0, 0, 0, True),
+            (1, 1, 0, 0, 0, 0, False),
```
AlanCoding, line 65: It might be easier to read if you reordered these, so that the 1s for the `True` cases go in a straight line from the upper left to bottom right, like `\`

awx/main/models/label.py
john-westcott-iv: 100%, I tired to do a more generic import and got a circular reference.

awx/main/models/label.py
john-westcott-iv: I'm not sure why they left that out initially but as you pointed out, this method is never referenced anywhere so I chopped it out.

AlanCoding (approved):

awx/main/tests/unit/models/test_label.py
john-westcott-iv: They are in that order because I treated the 1's and 0's like a 6 digit binary number to make sure I didn't miss something.

AlanCoding: ok, sure, that's fine

PR #12906, opened 2022-09-19
[feature-mesh-scaling] [collection] Add 'instance' module

relrod:
##### SUMMARY

First go at awx_collection module for `instance`

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME

 - Collection

### DIFF
--- a/awx_collection/plugins/modules/instance.py
+++ b/awx_collection/plugins/modules/instance.py
@@ -1,0 +1,148 @@
+#!/usr/bin/python
+# coding: utf-8 -*-
+
+
+# (c) 2022 Red Hat, Inc.
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+
+ANSIBLE_METADATA = {'metadata_version': '1.1', 'status': ['preview'], 'supported_by': 'community'}
+
+DOCUMENTATION = '''
+---
+module: instance
+author: "Rick Elrod (@relrod)"
+version_added: "4.3.0"
+short_description: create, update, or destroy Automation Platform Controller instances.
+description:
+    - Create, update, or destroy Automation Platform Controller instances. See
+      U(https://www.ansible.com/tower) for an overview.
+options:
+    hostname:
+      description:
+        - Hostname of this instance.
+      required: True
+      type: str
+    capacity_adjustment:
+      description:
+        - Capacity adjustment (0 <= capacity_adjustment <= 1)
+      required: False
+      type: float
+    enabled:
+      description:
+        - If true, the instance will be enabled and used.
+      required: False
+      type: bool
+      default: True
+    managed_by_policy:
+      description:
+        - Managed by policy
+      required: False
+      default: True
+      type: bool
+    node_type:
+      description:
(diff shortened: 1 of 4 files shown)


awx_collection/plugins/modules/instance.py
```diff
@@ -0,0 +1,148 @@
...
+        - Role that this node plays in the mesh.
+      choices:
+        - control
+        - execution
+        - hybrid
+        - hop
+      required: False
+      type: str
+      default: execution
+    node_state:
```
john-westcott-iv, line 55: Couple questions here...
1) Normally ansible just uses state ([absent, present]) do we want to call this just state? (we could alter test_completness to accept this)
2) Should the module be allowed to set all of these states? i.e. should only the controller be allowed to put something into the -fail state/ready state/etc?
3) Should there be an additional "state" parameter to allow (someday) for deletion of the endpoint (or is that never planned).

john-westcott-iv (commented):
Do we need anything for downloading the zip file?

awx_collection/plugins/modules/instance.py
relrod: > Normally ansible just uses state ([absent, present]) do we want to call this just state? (we could alter test_completness to accept this)

I did consider adding a `state` field, I actually had one at first and was considering doing something like making it set `node_state=deprovisioning` when `state=absent`, but I think that gets confusing pretty quickly, and as Hao told me last week, there are some things that happen during the deprovisioning process other than just deleting the node. So I'm not sure it makes sense to combine them here or to rename `node_state` as `state`. @TheRealHaoLiu Do you have thoughts on this?

> Should the module be allowed to set all of these states? i.e. should only the controller be allowed to put something into the -fail state/ready state/etc?

Probably another question for @TheRealHaoLiu - is it safe to just limit `node_state` to `installing` and `deprovisioning`?

jbradberry: > Normally ansible just uses state ([absent, present]) do we want to call this just state?

@john-westcott-iv Definitely not.  This is not the equivalent of Ansible's assertions about a package being installed on the system, but rather a value that affects how a node behaves and can be in any of a half-dozen distinct states.

jbradberry: (that is to say, this value is a model field.)

jbradberry: > Should the module be allowed to set all of these states?

No, it shouldn't.  I'd limit it, at the current time, to `installed` and `deprovisioning`.

jbradberry: `provisioning` will be brought in later.

awx_collection/plugins/modules/instance.py
```diff
@@ -0,0 +1,148 @@
...
+      required: False
+      type: bool
+      default: True
+    managed_by_policy:
+      description:
+        - Managed by policy
+      required: False
+      default: True
+      type: bool
+    node_type:
```
jbradberry, line 47: At the moment, we are only allowing the creation of nodes of type `execution`.

jbradberry (approved):
These two fields need to have their choices restricted, but this otherwise seems fine.

jbradberry (approved):

TheRealHaoLiu (approved):

PR #12907, opened 2022-09-19
Add additional schedule fields for new prompts

obaranov:
Add new prompts fields to the Schedules model, so we can set those when creating schedules from the job templates.

### DIFF
--- a/awxkit/awxkit/api/pages/unified_job_templates.py
+++ b/awxkit/awxkit/api/pages/unified_job_templates.py
@@ -19,6 +19,10 @@ class UnifiedJobTemplate(HasStatus, base.Base):
         'job_type',
         'verbosity',
         'inventory',
+        'forks',
+        'timeout',
+        'job_slice_count',
+        'execution_environment',
     )
 
     def __str__(self):


awxkit/awxkit/api/pages/unified_job_templates.py
```diff
@@ -19,6 +19,10 @@ class UnifiedJobTemplate(HasStatus, base.Base):
         'job_type',
         'verbosity',
         'inventory',
+        'forks',
+        'timeout',
+        'job_slice_count',
+        'execution_environment',
```
john-westcott-iv, line 25: What about instance groups?

obaranov: we have a method to add instance_groups and labels: https://github.com/ansible/awx/blob/feature-prompt-on-launch-on-templates/awxkit/awxkit/api/pages/schedules.py#L50-L56
Methods use POST to associate labels and ig, and those fields are passed with the Schedules payload.

john-westcott-iv (approved):

AlanCoding (approved):

PR #12908, opened 2022-09-19
new example grafana alert rule

rebeccahhh:
##### SUMMARY

add new optional grafana alert rule for when error rate is over a certain rate, also fix typo in URL

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.5.1
```

### DIFF
--- a/tools/docker-compose/README.md
+++ b/tools/docker-compose/README.md
@@ -483,5 +483,11 @@ $ PROMETHEUS=yes GRAFANA=yes make docker-compose
 
 ### Alerts in Grafana
 
-We are configuring alerts in grafana using the provisioning files method. This feature is new in Grafana as of August, 2022. Documentation can be found: https://grafana.com/docs/grafana/latest/administration/provisioning/#alerting however it does not fully show all parameters to the config. One way to understand how to build rules is to build them in the UI and use chrometools to inspect the payload as you save the rules. It appears that the "data" portion of the payload for each rule is the same syntax as needed in the provisioning file config. To reload the alerts without restarting the container, from within the container you can send a POST with `curl -X POST http://admin:admin@localhost:3000/api/admin/provisioning/alerting/relo
-ad`. Keep in mind the grafana container does not default contain `curl` and you can get it with `apk add curl`. 
+We are configuring alerts in grafana using the provisioning files method. This feature is new in Grafana as of August, 2022. Documentation can be found: https://grafana.com/docs/grafana/latest/administration/provisioning/#alerting however it does not fully show all parameters to the config.
+
+One way to understand how to build rules is to build them in the UI and use chrometools to inspect the payload as you save the rules. It appears that the "data" portion of the payload for each rule is the same syntax as needed in the provisioning file config. To reload the alerts without restarting the container, from within the container you can send a POST with `curl -X POST http://admin:admin@localhost:3000/api/admin/provisioning/alerting/reload`. Keep in mind the grafana container does not contain `curl`. You can install it with the command `apk add curl`.
+
+Another way to export rules is explore the api.
+1. Get all the folders:  `GET` to `/api/folders`
+2. Get the rules `GET` to `/api/ruler/grafana/api/v1/rules/{{ Folder }}`
+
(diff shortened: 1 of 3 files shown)


kdelee (approved):

kdelee:
modified your comment to include the desired full sentence "Bug, Docs Fix or other nominal change"

tools/grafana/alerting/alerts.yml
```diff
@@ -6,6 +6,163 @@ groups:
...
     rules:
+      - condition: B
+        dashboardUid: awx
+        data:
+          - refId: A
+            queryType: ''
+            relativeTimeRange:
+              from: 900
+              to: 0
+            datasourceUid: PBFA97CFB590B2093
```
shanemcd: Where does this value come from? Will this change between different instances of Grafana?

kdelee: These Uid's are a bit mysterious. This feature of the "provisioning files" for alerting is relatively new so the docs are not great.

That said, we tested deleting all grafana containers and volumes and starting from scratch. The alerts come up and work, and appear to point to the correct data sources. I think we can leave this as it is and perhaps if we DO want to have control of what it is we need to set the Uid in tools/grafana/datasources/prometheus_source.yml

tools/docker-compose/README.md
```diff
@@ -483,5 +483,4 @@ $ PROMETHEUS=yes GRAFANA=yes make docker-compose
...
-ad`. Keep in mind the grafana container does not default contain `curl` and you can get it with `apk add curl`. 
+We are configuring alerts in grafana using the provisioning files method. This feature is new in Grafana as of August, 2022. Documentation can be found: https://grafana.com/docs/grafana/latest/administration/provisioning/#alerting however it does not fully show all parameters to the config. One way to understand how to build rules is to build them in the UI and use chrometools to inspect the payload as you save the rules. It appears that the "data" portion of the payload for each rule is the same syntax as needed in the provisioning file config. To reload the alerts without restarting the container, from within the container you can send a POST with `curl -X POST http://admin:admin@localhost:3000/api/admin/provisioning/alerting/reload`. Keep in mind the grafana container does not default contain `curl` and you can get it with `apk add curl`.
```
john-westcott-iv: If we are cleaning this up a bit already can we remove the random "default" in the last sentence?

kdelee:
```suggestion
We are configuring alerts in grafana using the provisioning files method. This feature is new in Grafana as of August, 2022. Documentation can be found: https://grafana.com/docs/grafana/latest/administration/provisioning/#alerting however it does not fully show all parameters to the config.

One way to understand how to build rules is to build them in the UI and use chrometools to inspect the payload as you save the rules. It appears that the "data" portion of the payload for each rule is the same syntax as needed in the provisioning file config. To reload the alerts without restarting the container, from within the container you can send a POST with `curl -X POST http://admin:admin@localhost:3000/api/admin/provisioning/alerting/reload`. Keep in mind the grafana container does not contain `curl`. You can install it with the command `apk add curl`.

Another way to export rules is explore the api.
1. Get all the folders:  `GET` to `/api/folders`
2. Get the rules `GET` to `/api/ruler/grafana/api/v1/rules/{{ Folder }}`

```

rebeccahhh: can we switch it to "by default"? the default isn't supposed to be random haha

tools/grafana/alerting/alerts.yml
```diff
@@ -6,10 +6,167 @@ groups:
...
+              maxDataPoints: 43200
+              reducer: max
+              refId: max_success
+              type: reduce
+          - refId: compare
+            queryType: ''
+            relativeTimeRange:
+              from: 0
+              to: 0
+            datasourceUid: '-100'
```
kdelee, line 120: @shanemcd we figured out how to give prometheus data source a deterministic uid
This -100 uid is for data that doesn't reference a true "datasource", rather variables previously defined in the alert.

fosterseth:
awesome work

fosterseth (approved):

tools/docker-compose/README.md
rebeccahhh: resolved regardless :)

rebeccahhh:
force pushed to 
1. fix my yaml indenting
and
2. get this into 1 commit.
no functionality changes :)

relrod (approved):

PR #12914, opened 2022-09-19
Fixing is_detached methods' filters

john-westcott-iv:


AlanCoding (approved):
need unit test changes, but otherwise looks good

AlanCoding (approved):

PR #12917, opened 2022-09-20
Validate job_slice_count in API

AlanCoding:
##### SUMMARY
Followup from a few comments I left on the main feature PR.

I was having a problem with the use of `_eager_fields`, and now I see the problem it was addressing, but I found it better to not allow us to get to that point in the first place. Host counts can change over time, so I get not enforcing this on the JT, but _on launch_ it seems like nonsense to accept these values.

Separately, removing integer defaults.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -4140,9 +4140,9 @@ class JobLaunchSerializer(BaseSerializer):
     verbosity = serializers.ChoiceField(required=False, choices=VERBOSITY_CHOICES, write_only=True)
     execution_environment = serializers.PrimaryKeyRelatedField(queryset=ExecutionEnvironment.objects.all(), required=False, write_only=True)
     labels = serializers.PrimaryKeyRelatedField(many=True, queryset=Label.objects.all(), required=False, write_only=True)
-    forks = serializers.IntegerField(required=False, write_only=True, min_value=0, default=1)
-    job_slice_count = serializers.IntegerField(required=False, write_only=True, min_value=0, default=0)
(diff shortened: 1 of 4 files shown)


john-westcott-iv (approved):

CFSNM (approved):

AlanCoding:
testing found a valid problem where providing `job_slice_count` of 1 would give a 400 when inventory had 1 host.. but obviously this is intended, not a 400 kind of error. Pushing quick fix for that.

obaranov:
@AlanCoding I fixed and rerun API tests. And I have one failure. for the relaunch() test. 

Here is the test: 

```
        jt, _ = jt_with_job_slice_factory(max_slices=3)
        job = jt.launch(dict(job_slice_count=3)).wait_until_completed()
        verify_job_slices(job, 3)

        # relaunch job and check that values are preserved
        job_relaunched = job.relaunch().wait_until_completed()
        verify_job_slices(job_relaunched, 3)
```

```
awxkit.exceptions.BadRequest: Bad Request (400) received - {'detail': 'Cannot relaunch sliced workflow job after slice count has changed.'}
```

Is that expected behavior? It seems like when I relaunch the sliced workflow job, it tries to use the initially configured slice values, and not the ones provides on prompt.

AlanCoding:
The criteria for that 400 error is really outdated, and I think it's failing because the JT slice count is 1 and the prompted slice count is 3. This is a valid issue, but it's not strongly tied to this PR, and we are saying that we will file an issue and delay a fix until later.

PR #12925, opened 2022-09-20
Treat instance_groups prompt as template-less

AlanCoding:
##### SUMMARY
Stems from conversations in the main branch. Avoids saving instance_groups to job, which are not meaningful.

Needs UI work, because the UI still presents the JT `instance_groups` related data. We need it to always give an empty list in the pre-fill data in the prompts form. The launch endpoint is now present `[]` in the defaults, but the UI reads from the JT related list, but this has a different meaning.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/models/unified_jobs.py
+++ b/awx/main/models/unified_jobs.py
@@ -985,37 +985,33 @@ class UnifiedJob(
         for field_name, value in kwargs.items():
             if field_name not in valid_fields:
                 raise Exception('Unrecognized launch config field {}.'.format(field_name))
-            if isinstance(getattr(self.__class__, field_name).field, models.ManyToManyField):
+            field = None
+            # may use extra_data as a proxy for extra_vars
+            if field_name in config.SUBCLASS_FIELDS and field_name != 'extra_vars':
+                field = config._meta.get_field(field_name)
+            if isinstance(field, models.ManyToManyField):
                 many_to_many_fields.append(field_name)
                 continue
-            if isinstance(getattr(self.__class__, field_name).field, (models.ForeignKey)):
-                if value:
-                    setattr(config, "{}_id".format(field_name), value.id)
-                continue
-            key = field_name
-            if key == 'extra_vars':
-                key = 'extra_data'
-            setattr(config, key, value)
+            if isinstance(field, (models.ForeignKey)) and (value is None):
+                continue  # the null value indicates not-provided for ForeignKey case
+            setattr(config, field_name, value)
         config.save()
 
         for field_name in many_to_many_fields:
-            if field_name == 'credentials':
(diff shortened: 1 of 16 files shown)


john-westcott-iv (approved):
Approved knowing that you need to update the migration.

AlanCoding:
We expect further incoming UI work from @mabashian for the UI in this pull request.

awx/main/models/unified_jobs.py
```diff
@@ -981,37 +985,33 @@ def create_config_from_prompts(self, kwargs, parent=None):
...
-                    parent_items = list(getattr(parent, field_name, []).all())
-                    for item in job_item:
-                        # Do not include this item in the config if its in the parent
-                        if item not in parent_items:
-                            getattr(config, field_name).add(item)
+                # Assuming this field merges prompts with parent, save just the diff
+                if field_name in [field.name for field in parent._meta.get_fields()]:
+                    prompted_items = set(prompted_items) - set(getattr(parent, field_name).all())
+                if prompted_items:
+                    getattr(config, field_name).add(*prompted_items)
```
AlanCoding, line 1014: @john-westcott-iv I should notify you that I got sucked into rewriting more of this method. Several of things are changes I wanted to make anyway, but to recap what was going wrong...
 - We can't do `getattr(self.__class__, field_name)` when `self` is a Job and the field is `instance_groups`, because that relation doesn't exist, but since this is about saving fields to the `config`, it seems more direct to inspect the fields of _that_ object
 - When saving `config.instance_groups` we don't want the check `if item not in parent_items:`, because we don't _want_ to diff with the parent object. This is reflected in the test change.

obaranov:
How in that case we can get the associated instance_group for the lunched job? 
Do we need to update awxkit?

obaranov (approved):

AlanCoding:
> How in that case we can get the associated instance_group for the lunched job?

You can't get the associated instance group _s_ because it's unclear what that would mean. The API actually has 2 related records of instance groups - one, the prompted value on launch, if a prompt was provided, and two, a fully resolved preference list to be used by the task manager in all cases. The latter isn't as clear as it seems, because between creation and current time, IGs could be deleted, etc. We could _re-compute_ the fully resolved preference list for a job, but we haven't even written a method for this because there is no use for it. The JobTemplate instance groups are only a single contribution to the fully resolved preference list.

However, a job has a well-defined singular _instance group_ associated with it, and this is a direct field in the `/api/v2/jobs/` API, shown in summary fields, and I think in the related links.

I don't think this requires any change in awxkit, because I don't think we ever updated the page registry to begin with.

AlanCoding:
@mabashian I tested the UI for prompting for instance groups and it looks good to me.

PR #12926, opened 2022-09-20
[proj signing] Fix error message, rename action

relrod:
##### SUMMARY
- Fix out of scope variable in error message in the action plugin
- Rename action plugin from playbook_integrity to verify_project

Refs #12887 which pointed out the out of scope variable

##### ISSUE TYPE

##### COMPONENT NAME

 - API


AlanCoding (approved):
I don't see any test references to "playbook_integrity", so right now you still have a free hand to rename.

PR #12937, opened 2022-09-21
Set initial value node_type and node_state

fosterseth:
##### SUMMARY

default doesn't do anything, probably because we supply a default on the model itself. 

instead we should use `initial` to give it the values we want.

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -4819,7 +4819,7 @@ class InstanceSerializer(BaseSerializer):
             'ip_address',
             'listener_port',
         )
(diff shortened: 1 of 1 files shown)


jbradberry:
I'm afraid not.

https://github.com/ansible/awx/blob/devel/awx/api/serializers.py#L549-L550

Maybe this will still work, if your code change is hit after this point.  But I'd want to see it.

fosterseth:
@jrb interesting. Fortunately think the extra_kwargs initial overrides all that. on my local dev setup
image

jbradberry (approved):
Ok, if it works it works.

Issue #12932, opened 2022-09-21
Cannot relaunch a job with the prompted job_slice_count value


obaranov:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

When the job_slice_count is configured to be prompted on launch and user tries relaunch this job template job, it will receive the 

```
 Bad Request (400) received - {'detail': 'Cannot relaunch sliced workflow job after slice count has changed.'}
```

### AWX version

4.3.0

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

docker development environment

### Modifications

no

### Steps to reproduce

1. Create Job Template. make sure the Job Slices is marked as "Prompt on launch". 
2. Launch job template with job_slic_count > 2 value
3. Wait unit job completion. 
4. Select the resulting workflow job and click Relaunch. 

### Expected results

Api returned 400: 

```
 Bad Request (400) received - {'detail': 'Cannot relaunch sliced workflow job after slice count has changed.'}
```

### Actual results

Job should be relaunched with the initially prompted number for slices.

AlanCoding:
I've said I'm going to write a test for this in the PR 12944, so assigning myself. That should cover this, and I will take it out of draft when that is ready.

AlanCoding:
Oh, I didn't see that @obaranov had already created a test for this! At `test_jobs_slices_preserved_on_job_relaunch`

PR #12944, opened 2022-09-23
[tech debt] Cleanup cruft from job slicing made possible by prompts model changes

AlanCoding:
##### SUMMARY
This is a technical debt patch that hinges on a workaround that is no longer needed:

When sliced jobs were introduced, there was an in-congruence between what prompts could be provided on launch, and the WFJT / WJ models. Because sliced jobs go (JT)->(workflow job)->(job), anything that the job needs to inherit from the original JT prompts needs to pass through the workflow job (not the JT properties itself, only prompts, because prompts are re-combined with the JT properties at the time of job spawn). But at the time, the workflow models didn't have all the fields it needed for this purpose. The workaround was that we saved prompts to the related launch config model, relative to the workflow job. Those were re-applied to the jobs that the workflow spawned.

This workaround wasn't good, because that's not otherwise the purpose or interpretation of the launch config. It should only come into play if the job is relaunched or something like that. The workaround also left the system over-determined, because the role of the properties on the workflow job, itself, for sliced jobs, was left _completely unclear_.

This change here moves the _sliced job prompts_ from the launch config onto the _workflow job_. This is re-affirming the identity of WFJT / WJ models _as a container of prompts_, even though those are more often not-provided than cases like nodes or schedules. This makes the sliced job workflow jobs a lot more like ordinary workflow jobs, mechanically. The remaining custom logic is more feature-specific to slicing.

When ready, this is intended to address https://github.com/ansible/awx/issues/12932

For now, this will be tabled as higher priority work gets in.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -3174,7 +3174,7 @@ class JobTemplateCallback(GenericAPIView):
             kv['extra_vars'] = extra_vars_redacted
         kv['_prevent_slicing'] = True  # will only run against 1 host, so no point
         with transaction.atomic():
-            job = job_template.create_job(**kv)
+            job = job_template.create_unified_job(**kv)
 
         # Send a signal to signify that the job should be started.
         result = job.signal_start(inventory_sources_already_updated=inventory_sources_already_updated)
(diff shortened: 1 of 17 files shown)


AlanCoding:
The full set of integration tests are good for this. I'm very close to taking this out of draft, but I do think I should write a test for the related issue I linked. I don't see any other notes I left for myself lying around.

AlanCoding:
I confirmed that the new test we have for the bug is fixed by this patch. So I'm moving it out of draft.

PR #12949, opened 2022-09-23
Refactor of LDAP backend to be more efficient

john-westcott-iv:
##### SUMMARY

Need engineering and QE testing badly

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/sso/backends.py
+++ b/awx/sso/backends.py
@@ -327,7 +327,7 @@ class SAMLAuth(BaseSAMLAuth):
         return super(SAMLAuth, self).get_user(user_id)
 
 
-def _update_m2m_from_groups(ldap_user_group_dns, opts, remove=True):
+def _update_m2m_from_groups(ldap_user, opts, remove=True):
     """
     Hepler function to evaluate the LDAP team/org options to determine if LDAP user should
       be a member of the team/org based on their ldap group dns.
@@ -347,8 +347,11 @@ def _update_m2m_from_groups(ldap_user_group_dns, opts, remove=True):
         if isinstance(opts, str):
             opts = [opts]
         # If any of the users groups matches any of the list options
-        if list(set.intersection(set(ldap_user_group_dns), set(opts))):
-            return True
+        for group_dn in opts:
+            if not isinstance(group_dn, str):
+                continue
+            if ldap_user._get_groups().is_member_of(group_dn):
+                return True
     return False
 
 
@@ -370,7 +373,7 @@ def on_populate_user(sender, **kwargs):
 
     # Prefetch user's groups to prevent LDAP queries for each org/team when
     # checking membership.
-    ldap_user_group_dns = ldap_user._get_groups().get_group_dns()
+    ldap_user._get_groups().get_group_dns()
 
     # If the LDAP user has a first or last name > $maxlen chars, truncate it
     for field in ('first_name', 'last_name'):
@@ -422,13 +425,13 @@ def on_populate_user(sender, **kwargs):
         admins_opts = org_opts.get('admins', None)
         remove_admins = bool(org_opts.get('remove_admins', remove))
         desired_org_states[org_name] = {}
-        desired_org_states[org_name]['admin_role'] = _update_m2m_from_groups(ldap_user_group_dns, admins_opts, remove_admins)
+        desired_org_states[org_name]['admin_role'] = _update_m2m_from_groups(ldap_user, admins_opts, remove_admins)
         auditors_opts = org_opts.get('auditors', None)
         remove_auditors = bool(org_opts.get('remove_auditors', remove))
-        desired_org_states[org_name]['auditor_role'] = _update_m2m_from_groups(ldap_user_group_dns, auditors_opts, remove_auditors)
+        desired_org_states[org_name]['auditor_role'] = _update_m2m_from_groups(ldap_user, auditors_opts, remove_auditors)
         users_opts = org_opts.get('users', None)
         remove_users = bool(org_opts.get('remove_users', remove))
-        desired_org_states[org_name]['member_role'] = _update_m2m_from_groups(ldap_user_group_dns, users_opts, remove_users)
+        desired_org_states[org_name]['member_role'] = _update_m2m_from_groups(ldap_user, users_opts, remove_users)
 
         # If everything returned None (because there was no configuration) we can remove this org from our map
         if (
@@ -445,7 +448,7 @@ def on_populate_user(sender, **kwargs):
             continue
         users_opts = team_opts.get('users', None)
         remove = bool(team_opts.get('remove', True))
-        state = _update_m2m_from_groups(ldap_user_group_dns, users_opts, remove)
+        state = _update_m2m_from_groups(ldap_user, users_opts, remove)
         if state is not None:
             desired_team_states[team_name] = state
 
--- a/awx/sso/backends.py
+++ b/awx/sso/backends.py
@@ -475,35 +475,45 @@ def on_populate_user(sender, **kwargs):
 def reconcile_users_org_team_mappings(user, desired_org_states, desired_team_states, source):
     from awx.main.models import Organization, Team
 
-    org_content_type = ContentType.objects.get_for_model(Organization)
-    team_content_type = ContentType.objects.get_for_model(Team)
+    content_types = []
+    reconcile_items = []
+    if desired_org_states:
+        content_types.append(ContentType.objects.get_for_model(Organization))
+        reconcile_items.append(('organization', desired_org_states, Organization))
+    if desired_team_states:
+        content_types.append(ContentType.objects.get_for_model(Team))
+        reconcile_items.append(('team', desired_team_states, Team))
 
-    # users_roles is a flat list of IDs
-    users_roles = list(user.roles.filter(content_type__in=[org_content_type, team_content_type]).values_list('pk', flat=True))
+    if not content_types:
+        # If both desired states were empty we can simply return because there is nothing to reconcile
+        return
 
-    for object_type, desired_states, model in [('organization', desired_org_states, Organization), ('team', desired_team_states, Team)]:
+    # users_roles is a flat set of IDs
+    users_roles = set(user.roles.filter(content_type__in=content_types).values_list('pk', flat=True))
+
+    for object_type, desired_states, model in reconcile_items:
         # Get all of the roles in the desired states for efficient DB extraction
         roles = []
         for sub_dict in desired_states.values():
             for role_name in sub_dict:
-                if sub_dict[role_name] == None:
+                if sub_dict[role_name] is None:
                     continue
                 if role_name not in roles:
                     roles.append(role_name)
 
-        # Get a set of named tuples for the org name plus all of the roles we got above
-        model_roles = getattr(model, 'objects').filter(name__in=desired_states.keys()).values_list(*(['name'] + roles), named=True)
+        # Get a set of named tuples for the org/team name plus all of the roles we got above
+        model_roles = model.objects.filter(name__in=desired_states.keys()).values_list('name', *roles, named=True)
         for row in model_roles:
             for role_name in roles:
                 desired_state = desired_states.get(row.name, {})
-                if desired_state[role_name] == None:
+                if desired_state[role_name] is None:
                     # The mapping was not defined for this [org/team]/role so we can just pass
                     pass
 
                 # If somehow the auth adapter knows about an items role but that role is not defined in the DB we are going to print a pretty error
                 # This is your classic safety net that we should never hit; but here you are reading this comment... good luck and Godspeed.
                 role_id = getattr(row, role_name, None)
-                if role_id == None:
+                if role_id is None:
                     logger.error("{} adapter wanted to manage role {} of {} {} but that role is not defined".format(source, role_name, object_type, row.name))
                     continue
 
--- a/awx/sso/backends.py
+++ b/awx/sso/backends.py
@@ -329,23 +329,26 @@ class SAMLAuth(BaseSAMLAuth):
 
 def _update_m2m_from_groups(ldap_user_group_dns, opts, remove=True):
     """
-    Hepler function to update m2m relationship based on LDAP group membership.
+    Hepler function to evaluate the LDAP team/org options to determine if LDAP user should
+      be a member of the team/org based on their ldap group dns.
+
+    Returns:
+        True - User should be added
+        False - User should be removed
+        None - Users membership should not be changed
     """
-    should_add = False
     if opts is None:
         return None
     elif not opts:
         pass
     elif isinstance(opts, bool) and opts is True:
-        should_add = True
+        return True
     else:
         if isinstance(opts, str):
             opts = [opts]
         # If any of the users groups matches any of the list options
         if list(set.intersection(set(ldap_user_group_dns), set(opts))):
-            should_add = True
-    if should_add:
-        return True
+            return True
     return False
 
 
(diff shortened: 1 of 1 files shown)


awx/sso/backends.py
```diff
@@ -423,3 +457,28 @@ def on_populate_user(sender, **kwargs):
     if profile.ldap_dn != ldap_user.dn:
         profile.ldap_dn = ldap_user.dn
         profile.save()
+
+    # Get all of the orgs listed in LDAP from the database
+    roles = ['admin_role', 'auditor_role', 'member_role']
+    prefetch_roles = ["{}__members".format(role_name) for role_name in roles]
+    orgs_tied_to_ldap = Organization.objects.filter(name__in=desired_org_states.keys()).prefetch_related(*prefecth_roles)
```
kdelee: this saves us a bunch of queries from when we do   `getattr(org, role).members`

awx/sso/backends.py
```diff
@@ -423,3 +457,28 @@ def on_populate_user(sender, **kwargs):
...
+            if desired_org_states[org.name][role] is None:
+                # If something got a none we just need to continue on
+                pass
+            elif desired_org_states[org.name][role]:
+                getattr(org, role).members.add(user)
+            elif user in getattr(org, role).members.all():
+                getattr(org, role).members.remove(user)
+
+    # Get all of the teams listed in LDAP from the database
+    teams_tied_to_ldap = Team.objects.filter(name__in=desired_team_states.keys()).prefetch_related('member_role__members')
```
kdelee: same thing here, doing the prefetch saves us a bunch of queries on ` team.member_role.members`

awx/sso/backends.py
```diff
@@ -383,31 +378,70 @@ def on_populate_user(sender, **kwargs):
...
     org_map = getattr(backend.settings, 'ORGANIZATION_MAP', {})
+    team_map = getattr(backend.settings, 'TEAM_MAP', {})
+
+    # Move this junk into save of the settings for performance later, there is no need to do that here
+    #    with maybe the exception of someone defining this in settings before the server is started?
+    # ==============================================================================================================
+    # Get all of the orgs in the DB by name and create any new org defined in LDAP
+
+    # IS THIS ACTUALLY GETTING JUST NAMES OR DOES IT GET ALL AND THEN GET NAMES FORM THEM ALL?
+    existing_org_names = list(Organization.objects.all().values_list('name', flat=True))
```
kdelee: might be able to do `Organization.objects.all().only('name')` though tbh I'm not 100% on difference

john-westcott-iv: I checked and this is doing a query specifically for 'name' field:
```
tools_postgres_1    | 2022-09-24 13:41:19.087 UTC [111] LOG:  statement: SELECT "main_organization"."name" FROM "main_organization" ORDER BY "main_organization"."name" ASC
```

AlanCoding: The only difference is that `.only()` still builds the python model objects with incomplete fields. If you're only going to use the name strings, then I think values_list is preferable.

awx/sso/backends.py
```diff
@@ -423,3 +460,34 @@ def on_populate_user(sender, **kwargs):
     if profile.ldap_dn != ldap_user.dn:
         profile.ldap_dn = ldap_user.dn
         profile.save()
+
+    # Get all of the orgs listed in LDAP from the database
+    roles = ['admin_role', 'auditor_role', 'member_role']
+    prefetch_roles = ["{}__members".format(role_name) for role_name in roles]
+    orgs_tied_to_ldap = Organization.objects.filter(name__in=desired_org_states.keys()).prefetch_related(*prefetch_roles)
```
AlanCoding: I realize that this is may be un-fixable in the abstract due to the business requirements, we should keep in mind that this is a heavy `prefetch_related` that could, at some point, introduce problems due to the quantity of data that it returns. The number of items it returns scales with the total number of admin/auditor/members in all the filtered organizations. But... these organizations are defined by the LDAP mapping.

You're still fetching more than what you technically need, as using a custom `Prefetch` object could avoid returning memberships for users other than the current user. That would be better handled as an annotate, since the expectations of prefetch get difficult to follow.

awx/sso/backends.py
```diff
@@ -383,31 +378,73 @@ def on_populate_user(sender, **kwargs):
...
-        _update_m2m_from_groups(user, ldap_user, org.auditor_role.members, auditors_opts, remove_auditors)
+        desired_org_states[org_name]['auditor_role'] = _update_m2m_from_groups(ldap_user_group_dns, auditors_opts, remove_auditors)
         users_opts = org_opts.get('users', None)
         remove_users = bool(org_opts.get('remove_users', remove))
-        _update_m2m_from_groups(user, ldap_user, org.member_role.members, users_opts, remove_users)
+        desired_org_states[org_name]['member_role'] = _update_m2m_from_groups(ldap_user_group_dns, users_opts, remove_users)
 
-    # Update team membership based on group memberships.
-    team_map = getattr(backend.settings, 'TEAM_MAP', {})
+        # If everything returned None (because there was no configuration) we can skip this host
```
AlanCoding: > we can skip this host

Host? Do you mean organization instead? "we can skip this organization"

awx/sso/backends.py
```diff
@@ -327,31 +327,26 @@ def get_user(self, user_id):
...
+        if list(set.intersection(set(ldap_user_group_dns), set(opts))):
+            should_add = True
     if should_add:
-        user.save()
-        related.add(user)
-    elif remove and user in related.all():
-        user.save()
-        related.remove(user)
+        return True
+    return False
```
AlanCoding, line 357: I don't see any case where `should_add` would not be a boolean, so it would be simpler to just `return should_add`. Before, this wasn't totally viable because it took these side-effect actions.

EDIT: you could also get rid of the `should_add` variable, with even greater condensing of code.

Also, what does this method do anymore? The docstring was probably incomplete before, but seems borderline incorrect now.

awx/sso/backends.py
john-westcott-iv: Good catch, updated.

awx/sso/backends.py
john-westcott-iv: Method comment updated, `should_add` removed.

AlanCoding: The comment I was thinking of was

> Hepler function to update m2m relationship based on LDAP group membership.

When I read that, I try to parse it into something that makes sense, and it just isn't happening.

awx/sso/backends.py
```diff
@@ -423,3 +463,34 @@ def on_populate_user(sender, **kwargs):
...
+            if desired_org_states[org.name][role] is None:
+                # If something got a none we just need to continue on
+                pass
+            elif desired_org_states[org.name][role]:
+                if user not in getattr(org, role).members.all():
+                    logger.debug("Adding LDAP user {} to {} as {}".format(user.username, org.name, role))
+                    getattr(org, role).members.add(user)
+            elif user in getattr(org, role).members.all():
+                logger.debug("Removing LDAP user {} from {} in {}".format(user.username, role, org.name))
+                getattr(org, role).members.remove(user)
```
AlanCoding: What you're doing is making the code easier to follow ultimately, which the primary data structures being `desired_org_states` and `desired_team_states`, that's good :+1: 

But you should allow yourself to modify the method signatures to reflect your improved code structure. This block, implemented in a separate method, would take `desired_org_states` and `user`. That could be tested independently to verify the correctness of the database queries and updates.

Name it like `update_database_to_LDAP_orgs` or `save_LDAP_desired_org_memberships`

awx/sso/backends.py
AlanCoding: just saw your update, resolved :+1:

awx/sso/backends.py
```diff
@@ -383,31 +381,73 @@ def on_populate_user(sender, **kwargs):
...
+
+    # Move this junk into save of the settings for performance later, there is no need to do that here
+    #    with maybe the exception of someone defining this in settings before the server is started?
+    # ==============================================================================================================
+
+    # Get all of the names of orgs in the DB and create any new org defined in LDAP that does not exist in the DB
+    existing_org_names = list(Organization.objects.all().values_list('name', flat=True))
+    for org_name in org_map.keys():
+        if org_name not in existing_org_names:
+            Organization.objects.get_or_create(name=org_name)
```
AlanCoding: If the org name isn't in `existing_org_names`, then deterministically, you create, right? Then I would opt for `create` instead of `get_or_create`, as it will be more performant and clear. If you are concerned about race conditions, this isn't the way to do things, you should do a try-except if you are really worried about it being created while in the loop so that you can later on make correct decisions.

awx/sso/backends.py
```diff
@@ -383,31 +381,73 @@ def on_populate_user(sender, **kwargs):
...
+            continue
+        # A team may have an org that we didn't previously evaluate
+        if team_opts['organization'] not in existing_org_names:
+            Organization.objects.get_or_create(name=team_opts['organization'])
+            # Append it to the list so that we don't try and create this again if its used by another team
+            existing_org_names.append(team_opts['organization'])
+        if team_name not in existing_team_names:
+            # Going a little inefficient here, we could prob get name/ids above and then set this by ID but I've got limited time r/n
+            #    and this code should only really be getting hit once if an LDAP config change adds a team with an org not built.
+            org, created = Organization.objects.get_or_create(name=team_opts['organization'])
```
AlanCoding: Again, the use of `get_or_create` here doesn't make any sense to me, since all the prior checks just validated that the organization _already exists_ so this should only be a get, and never a create.

awx/sso/backends.py
```diff
@@ -383,31 +381,73 @@ def on_populate_user(sender, **kwargs):
...
-        _update_m2m_from_groups(user, ldap_user, org.member_role.members, users_opts, remove_users)
+        desired_org_states[org_name]['member_role'] = _update_m2m_from_groups(ldap_user_group_dns, users_opts, remove_users)
 
-    # Update team membership based on group memberships.
-    team_map = getattr(backend.settings, 'TEAM_MAP', {})
+        # If everything returned None (because there was no configuration) we can remove this org from our map
+        if (
+            desired_org_states[org_name]['admin_role'] == None
+            and desired_org_states[org_name]['auditor_role'] == None
+            and desired_org_states[org_name]['member_role'] == None
```
AlanCoding: I'd prefer to tighten this up, like `if all(desired_org_states[org_name][role_field] is None for role_field in roles):`

I also thought that the linter got you in trouble for `== None`.

where the definition of `roles = ['admin_role', 'auditor_role', 'member_role']` is shared with its use later on for saving the data.

awx/sso/backends.py
john-westcott-iv: We can't technically guarantee that because there is no restriction that an organization tied to a team has to be in the org map. Consider an org map like:
```
{ "a", "b", "c"}
```
And a team map of:
```
{ "team1": { "org": "a" }, "team2": { "org": "d" } }
```
We would have never created "d" before this point.

AlanCoding: That would hit `if team_opts['organization'] not in existing_org_names:` in the conditional right before this. So in that scenario, we would hit line 407 and line 413, and it doesn't make sense to me that both would be `get_or_create`. In both cases, there should be no ambiguity about whether it's getting or creating. It's one or the other for each. It may create and then get.

awx/sso/backends.py
```diff
@@ -423,3 +468,52 @@ def on_populate_user(sender, **kwargs):
     if profile.ldap_dn != ldap_user.dn:
         profile.ldap_dn = ldap_user.dn
         profile.save()
+
+    reconsile_users_org_team_mappings(user, desired_org_states, desired_team_states, 'LDAP')
+
+
+def reconsile_users_org_team_mappings(user, desired_org_states, desired_team_states, source):
```
AlanCoding: spelling of "reconsile"

awx/sso/backends.py
```diff
@@ -423,3 +468,52 @@ def on_populate_user(sender, **kwargs):
...
+    from awx.main.models import Organization, Team
+
+    org_content_type = ContentType.objects.get_for_model(Organization)
+    team_content_type = ContentType.objects.get_for_model(Team)
+
+    # users_roles is a flat list of IDs
+    users_roles = list(user.roles.filter(content_type__in=[org_content_type, team_content_type]).values_list('pk', flat=True))
```
AlanCoding: Would be technically faster as `set`, and it looks like the code allows for that.

awx/sso/backends.py
```diff
@@ -423,3 +468,52 @@ def on_populate_user(sender, **kwargs):
     if profile.ldap_dn != ldap_user.dn:
         profile.ldap_dn = ldap_user.dn
         profile.save()
+
+    reconcile_users_org_team_mappings(user, desired_org_states, desired_team_states, 'LDAP')
```
AlanCoding, line 472: A big non-obvious performance tweak I'd like to have is to handle the case that `desired_org_states` and `desired_team_states` are effectively nothing, in practice, `{}`

That could come in the form of not calling this method when those are empty, or, inside of that method we could bail early. Because the `ContentType.objects.get_for_model` has a query cost and `user_roles` is potentially costly in terms of the volume of return data, and we just don't want to run those when there's nothing to reconcile.

This might already be handled upstream of this, and if it is, you can disregard this comment.

awx/sso/backends.py
```diff
@@ -423,3 +468,52 @@ def on_populate_user(sender, **kwargs):
...
+        # Get all of the roles in the desired states for efficient DB extraction
+        roles = []
+        for sub_dict in desired_states.values():
+            for role_name in sub_dict:
+                if sub_dict[role_name] == None:
+                    continue
+                if role_name not in roles:
+                    roles.append(role_name)
+
+        # Get a set of named tuples for the org name plus all of the roles we got above
```
AlanCoding: > org name

Could be org or team name.

awx/sso/backends.py
```diff
@@ -423,3 +468,52 @@ def on_populate_user(sender, **kwargs):
...
+        model_roles = getattr(model, 'objects').filter(name__in=desired_states.keys()).values_list(*(['name'] + roles), named=True)
```
AlanCoding: There's a lot of indirection going on here, so I'd prefer to clean it up a bit for readability. Even after someone gets past those readability issues, it uses Django feature that most people won't be familiar with.

So to reduce indirection, I'd suggest

```python
model_roles = model.objects.filter(name__in=desired_states.keys()).values_list('name', *roles, named=True)
```

awx/sso/backends.py
```diff
@@ -423,3 +468,52 @@ def on_populate_user(sender, **kwargs):
...
+        for row in model_roles:
+            for role_name in roles:
+                desired_state = desired_states.get(row.name, {})
+                if desired_state[role_name] == None:
```
AlanCoding: a standard python-ism is that we compare `is None` as opposed to the `==`. If you're worried that might affect your logic we can talk, but you can probably just drop that change in without any issues.

awx/sso/backends.py
```diff
@@ -383,31 +386,73 @@ def on_populate_user(sender, **kwargs):
...
+            except IntegrityError:
+                # Another thread must have created this org before we did so now we need to get it
+                new_org = Organization.objects.get(name=org_name)
+            # Add the org name to the existing orgs since we created it and we may need it to build the teams below
+            existing_orgs[org_name] = new_org.id
+
+    # Do the same for teams
+    existing_team_names = list(Team.objects.all().values_list('name', flat=True))
+    for team_name, team_opts in team_map.items():
+        if 'organization' not in team_opts:
```
AlanCoding: Having read the prior organization loop, I think this should be `if not team_opts.get('organization'):`, to handle the case that the team mapping has an organization that's blank or null.

AlanCoding (approved):
This work has matured very nicely. There are clearly delineated functional parts to this now, and I don't see anything that could easily be slower than what was there before.

jay-steurer (approved):

jay-steurer:
Verified by QE

kdelee (approved):

PR #12951, opened 2022-09-24
Adding make option to enable sql debugging

john-westcott-iv:
##### SUMMARY

Turns on SQL query logging and SQL headers if enabled.

##### ISSUE TYPE

##### COMPONENT NAME

 - Other

##### AWX VERSION

```
awx: 21.6.1.dev52+g05e3c29fe8
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/tools/docker-compose/ansible/roles/sources/tasks/sql_debug.yml
+++ b/tools/docker-compose/ansible/roles/sources/tasks/sql_debug.yml
@@ -1,0 +1,16 @@
+---
+- name: Enable SQL Logging in container
+  set_fact:
+    pg_log_statement: -c log_statement=all
+
+# This setting will add two headers to http responses about SQL:
+#   < X-API-Query-Count: 9
+#   < X-API-Query-Time: 0.006s
+# Technically, the way this is setup, if you disbale pg_debug after its enabled
+#  and don't rebuild your container the flag to set the headers will remain in the conf
+#  if I tried to remove it from here and someone had manually set it in the local_overrides files
+#  it would cause confustion as this code would strip it out of their local_overrides
+- name: Add SQL_DEBUG to local_settings.py
+  lineinfile:
+    path: "{{ sources_dest }}/local_settings.py"
+    line: "SQL_DEBUG = True"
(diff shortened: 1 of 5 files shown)


tools/docker-compose/ansible/roles/sources/tasks/sql_debug.yml
```diff
@@ -0,0 +1,16 @@
...
+#   < X-API-Query-Count: 9
+#   < X-API-Query-Time: 0.006s
+# Technically, the way this is setup, if you disbale pg_debug after its enabled
+#  and don't rebuild your container the flag to set the headers will remain in the conf
+#  if I tried to remove it from here and someone had manually set it in the local_overrides files
+#  it would cause confustion as this code would strip it out of their local_overrides
+- name: Add SQL_DEBUG to local_settings.py
+  lineinfile:
+    path: "{{ sources_dest }}/local_settings.py"
+    line: "SQL_DEBUG = True"
```
AlanCoding, line 16: It seems a little overkill to enable SQL logging in django _and_ in postgres. Wouldn't it be hard to see anything?

tools/docker-compose/ansible/roles/sources/tasks/sql_debug.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: Enable SQL Logging in container
+  set_fact:
+    pg_log_statement: -c log_statement=all
+
+# This setting will add two headers to http responses about SQL:
+#   < X-API-Query-Count: 9
+#   < X-API-Query-Time: 0.006s
```
AlanCoding, line 8: I thought these came with general DEBUG mode?

AlanCoding (changes requested):
This feels overly-specific to me. Users should already have a means to change the file-based settings. That changed some time ago for docker-compose, now it looks for local_overrides.py or something like that. The advice should be to set SQL_DEBUG there.

I also think that adding an option for extra postgres options would be fine, but handling each parameter individually doesn't strike me as a good approach.

djyasin:
@john-westcott-iv is this PR still relevant? Can we close this one?

Issue #12954, opened 2022-09-26
Better handling of Health Checks


kialam:
Now that health check tasks are performed asynchronously, there is very little feedback that the user currently receives that the task is running or is completed (ws future-scoped). Here is a list of follow up action items to improve the UX flow:

1. New API field in the Instances model called `health_check_started` to track when a user has submitted a health check
2. UI will show a Pending Health Check Status in the List view (expanded view), and details view based which timestamp is more recent (`health_check_started` vs `last_health_check`). Use clock icon next to timestamp.
3. UI will debounce requests to health check endpoint by greying out Health Check button. UI will add additional logic to skip instances that have pending health checks when making health check calls to the API.
4. `/health_check` endpoint will handle debouncing requests
5.  Instances endpoints will expose the new `health_check_started` field

Note: Toast notification will stay in place to indicate to the user that a refresh is still required to get the latest node state.

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 - UI

PR #12959, opened 2022-09-27
Add a new Instance.health_check_started field

jbradberry:
##### SUMMARY
This will enable us to provide more useful information for the user, now that all user-triggered health checks are async.

Also, de-bounce the health check endpoint to not allow additional health check tasks to be triggered when one is already in progress.

related #12954 

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API


jbradberry:
@AlanCoding what do you think about deprecating `Instance.last_health_check` in favor of a field named `Instance.health_check_completed`?

awx/main/tasks/system.py
```diff
@@ -485,13 +485,17 @@ def inspect_execution_nodes(instance_list):
...
+                instance.save(update_fields=['health_check_started'])
                 execution_node_health_check.apply_async([hostname])
             elif instance.capacity == 0 and instance.enabled:
                 # nodes with proven connection but need remediation run health checks are reduced frequency
                 if not instance.last_health_check or (nowtime - instance.last_health_check).total_seconds() >= settings.EXECUTION_NODE_REMEDIATION_CHECKS:
                     # Periodically re-run the health check of errored nodes, in case someone fixed it
                     # TODO: perhaps decrease the frequency of these checks
                     logger.debug(f'Restarting health check for execution node {hostname} with known errors.')
+                    instance.health_check_started = nowtime
+                    instance.save(update_fields=['health_check_started'])
```
fosterseth: maybe we can combine these saves with the last_seen save on line 474

jbradberry: I don't believe so, no.  That save is unconditional, these need to be conditional and to be tied to the launch of the health check tasks.  We don't want to make these unconditional and then not actually trigger a health check, because then the health check endpoint will be blocked for no good reason.

awx/main/models/ha.py
```diff
@@ -114,6 +114,11 @@ class Instance(HasPolicyEditsMixin, BaseModel):
         editable=False,
         help_text=_('Last time instance ran its heartbeat task for main cluster nodes. Last known connection to receptor mesh for execution nodes.'),
     )
+    health_check_started = models.DateTimeField(
+        null=True,
+        editable=False,
+        help_text=_("The last time a health check was initiated on this instance."),
```
AlanCoding, line 120: ...only for execution nodes

AlanCoding:
> Also, de-bounce the health check endpoint to not allow additional health check tasks to be triggered when one is already in progress.

...you mean that the UI will interpret a health check to be in progress if the "started" timestamp is after the "last_health_check" field?

awx/api/serializers.py
```diff
@@ -4893,6 +4893,7 @@ class Meta:
             'created',
             'modified',
             'last_seen',
+            'health_check_started',
```
AlanCoding, line 4897: Since this field is used so infrequently (and so specifically to user-initiated or execution-node triggers) it might be better to _hide_ this field by making it `editable=False`, not showing it in the serializer, and replacing it with a field like `health_check_in_progress`, which would only give `True` if a user request is still in progress, or an execution node check was in progress (probably in response to some user-initiated action).

Control plane nodes still have health checks, and it's unreasonable to apply any of this to that.

awx/main/models/ha.py
jbradberry: Why?  We can trigger the `cluster_node_health_check` from the endpoint as well.

awx/api/serializers.py
jbradberry: It is already `editable=False`.

jbradberry: > Control plane nodes still have health checks, and it's unreasonable to apply any of this to that.

@AlanCoding Why?  The instance health check endpoint can trigger the control plane health check.  In fact, it is the *only* entry point into `cluster_node_health_check` as far as I can see.

My intention is that any process, automated or manual, that would write to `last_health_check` ought to get a corresponding write beforehand to `health_check_started`.

jbradberry: > ... not showing it in the serializer, and replacing it with a field like `health_check_in_progress` ...

@kialam 's changes are effectively already doing this on the frontend, but my leanings are that if someone is peeping into the output of the API they actually care about the details and we shouldn't try to abstract it away into a less informative "field".

jbradberry:
> > Also, de-bounce the health check endpoint to not allow additional health check tasks to be triggered when one is already in progress.
> 
> ...you mean that the UI will interpret a health check to be in progress if the "started" timestamp is after the "last_health_check" field?

Yes.

awx/api/serializers.py
AlanCoding: > My intention is that any process, automated or manual, that would write to last_health_check ought to get a corresponding write beforehand to health_check_started.

That's a bad idea, because IIRC `cluster_node_heartbeat` updates `last_health_check` in a periodic task every 20 seconds, and you can't consolidate those saves (by statement of their intended function).

If you run a manual health check for a control node, it's not even clear whether the updated health check data came from the task you triggered _or the periodic task_, which is why this entire function is nonsense for the control plane and should be disabled for those node types. Nonetheless, you don't want to selectively hide fields in the serializer. At least if added `health_check_in_progress` as a boolean, it can be documented to only cover manual health checks and somewhat honestly return False for control plane nodes.

jbradberry: Ok, I have removed the writes to the new field for automated processes.  It only writes from the API endpoint now.

awx/main/tasks/system.py
jbradberry: Per the discussion yesterday, I've removed writes to this field by the automated processes.

awx/api/serializers.py
AlanCoding: you're still showing the field in the serializer, and I still think that could be confusing in cases.

AlanCoding (approved):
I take it there's no UI work to do anymore.

jbradberry:
@AlanCoding no, Kia is working on it.

Issue #12644, opened 2022-08-11
502 Bad Gateway Errors in AWX GUI browsing Jobs, Templates


2and3makes23:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

When browsing Jobs or other job related lists via UI or API, nginx produces 502 Bad Gateway errors, when certain Jobs are part of the result to be shown to the user.

Those jobs have in common that they ended in error state.

Log entries that might be of interest:
* Web container
```
2022/08/11 07:11:04 [error] 31#31: *5632 upstream prematurely closed connection while reading response header from upstream, client: <IP>, server: _, request: "GET /api/v2/unified_job_templates/?order_by=name&organization__name__icontains=<some_org>&page=1&page_size=20&type=job_template%2Cworkflow_job_template HTTP/1.1", upstream: "uwsgi://127.0.0.1:8050", host: "<awx_url>", referrer: "<awx_url>"
<IP> - - [11/Aug/2022:07:11:04 +0000] "GET /api/v2/unified_job_templates/?order_by=name&organization__name__icontains=<some_org>&page=1&page_size=20&type=job_template%2Cworkflow_job_template HTTP/1.1" 502 150 "<awx_url>" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:90.0) Gecko/20100101 Firefox/90.0" "<IP>, <IP>"
DAMN ! worker 1 (pid: 62) died, killed by signal 9 :( trying respawn ...
Respawned uWSGI worker 1 (new pid: 65)
mounting awx.wsgi:application on / 

...

[10/Aug/2022:08:12:02 +0000] "GET /websocket/broadcast/ HTTP/1.1" 502 150 "-" "Python/3.9 aiohttp/3.7.4" "-"
2022/08/10 08:12:02 [error] 31#31: *1 connect() failed (111: Connection refused) while connecting to upstream, client: <IP>, server: _, request: "GET /websocket/broadcast/ HTTP/1.1", upstream: "http://127.0.0.1:8051/websocket/broadcast/", host: "<IP>"
```

* Task container
```
[...]

PR #12961, opened 2022-09-27
Result_traceback should not include job stdout

fosterseth:
related https://github.com/ansible/awx/issues/12644
##### SUMMARY

If a job fails, we do receptor work results and put that output into result_traceback.

We should only do this if
1. Receptor unit has failed
2. Runner callback processed 0 events

Otherwise we risk putting too much data into this field.

Would require this receptor change https://github.com/ansible/receptor/pull/672

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.5.1.dev221+gaeb614e45d
```

### DIFF
--- a/awx/main/tasks/receptor.py
+++ b/awx/main/tasks/receptor.py
@@ -411,9 +411,11 @@ class AWXReceptorJob:
                     unit_status = receptor_ctl.simple_command(f'work status {self.unit_id}')
                     detail = unit_status.get('Detail', None)
                     state_name = unit_status.get('StateName', None)
+                    stdout_size = unit_status.get('StdoutSize', 0)
                 except Exception:
                     detail = ''
                     state_name = ''
+                    stdout_size = 0
                     logger.exception(f'An error was encountered while getting status for work unit {self.unit_id}')
 
                 if 'exceeded quota' in detail:
@@ -424,9 +426,16 @@ class AWXReceptorJob:
                     return
 
                 try:
-                    resultsock = receptor_ctl.get_work_results(self.unit_id, return_sockfile=True)
-                    lines = resultsock.readlines()
-                    receptor_output = b"".join(lines).decode()
+                    receptor_output = ''
+                    if state_name == 'Failed' and self.task.runner_callback.event_ct == 0:
+                        # if receptor work unit failed and no events were emitted, work results may
+                        # contain useful information about why the job failed. In case stdout is
+                        # massive, only ask for last 1000 bytes
+                        startpos = max(stdout_size - 1000, 0)
+                        resultsock, resultfile = receptor_ctl.get_work_results(self.unit_id, startpos=startpos, return_socket=True, return_sockfile=True)
+                        resultsock.setblocking(False)  # this makes resultfile reads non blocking
+                        lines = resultfile.readlines()
+                        receptor_output = b"".join(lines).decode()
                     if receptor_output:
                         self.task.runner_callback.delay_update(result_traceback=receptor_output)
                     elif detail:


awx/main/tasks/receptor.py
```diff
@@ -411,11 +413,16 @@ def _run_internal(self, receptor_ctl):
...
+                    if state_name == 'Failed' and self.task.runner_callback.event_ct == 0:
+                        # if receptor work unit failed and no events were emitted, work results may
+                        # contain useful information about why the job failed. In case stdout is
+                        # massive, only ask for last 1000 bytes
+                        startpos = max(stdout_size - 1000, 0)
+                        resultsock = receptor_ctl.get_work_results(self.unit_id, startpos=startpos, return_sockfile=True)
+                        lines = resultsock.readlines()
+                        receptor_output = b"".join(lines).decode()
+                        if receptor_output:
+                            self.task.runner_callback.delay_update(result_traceback=receptor_output)
```
AlanCoding: I'm concerned that in this case we may lose the info from `detail`, where `receptor_output` is empty

awx/main/tasks/receptor.py
```diff
@@ -411,11 +413,16 @@ def _run_internal(self, receptor_ctl):
...
-                    receptor_output = b"".join(lines).decode()
-                    if receptor_output:
-                        self.task.runner_callback.delay_update(result_traceback=receptor_output)
+                    if state_name == 'Failed' and self.task.runner_callback.event_ct == 0:
+                        # if receptor work unit failed and no events were emitted, work results may
+                        # contain useful information about why the job failed. In case stdout is
+                        # massive, only ask for last 1000 bytes
+                        startpos = max(stdout_size - 1000, 0)
+                        resultsock = receptor_ctl.get_work_results(self.unit_id, startpos=startpos, return_sockfile=True)
+                        lines = resultsock.readlines()
```
shanemcd: I'm worried about this hanging if the stream doesn't end with a `\n`. We may need to put the socket in non-blocking mode before doing this and just read the raw bytes.

awx/main/tasks/receptor.py
fosterseth: I fixed it up so that we still get Detail in the case where there are no results

AlanCoding (approved):
@shanemcd's comment may be a valid concern, but this doesn't seem to make it any worse than it was before in that respect, and should prevent some other disaster cases.

awx/main/tasks/receptor.py
fosterseth: I did some local testing and readlines() seems to be okay with files that don't end in \n. If it gets an EOF it just stops there.

we could also just do read() or read(1000)

as for non-blocking I added `resultsock.setblocking(False)`. Hard to test this exactly, but I will make sure this codepath doesn't throw errors

awx/main/tasks/receptor.py
```diff
@@ -411,9 +413,16 @@ def _run_internal(self, receptor_ctl):
...
-                    resultsock = receptor_ctl.get_work_results(self.unit_id, return_sockfile=True)
-                    lines = resultsock.readlines()
-                    receptor_output = b"".join(lines).decode()
+                    receptor_output = ''
+                    if state_name == 'Failed' and self.task.runner_callback.event_ct == 0:
+                        # if receptor work unit failed and no events were emitted, work results may
+                        # contain useful information about why the job failed. In case stdout is
+                        # massive, only ask for last 1000 bytes
+                        startpos = max(stdout_size - 1000, 0)
+                        resultsock, resultfile = receptor_ctl.get_work_results(self.unit_id, startpos=startpos, return_socket=True, return_sockfile=True)
```
AlanCoding, line 435: What's are the version compatibility issues with this? Does this requires an updated receptor, and does that make it unreasonable to backport? Is there any idea to work on a different fix for backports?

fosterseth: the receptor change is minimal, so backporting shouldn't be too hard. 

a more minimal version of this that doesn't require a receptor change is to not do get_work_results at all if stdout size > 1000.

AlanCoding: I see, so we could merge that and develop that solution you mention and test it separately. Fine by me.

AlanCoding:
I really want to get this merged!

PR #12968, opened 2022-09-27
Make sure to include field defaults for Instance node_type and node_state

jbradberry:
##### SUMMARY
Make sure to include field defaults for Instance node_type and node_state.

We want to make sure that if the user leaves off those fields from the payload that the expected defaults (type: execution, state: installed) get applied, instead of triggering the validation error.

##### ISSUE TYPE

##### COMPONENT NAME
 - API


relrod (approved):
Tested in Kind, works fine.

AlanCoding (approved):
logical

PR #12969, opened 2022-09-28
[tech debt] Remove imports and references that are no longer needed

AlanCoding:
##### SUMMARY
This is a code quality change, similar to https://github.com/ansible/awx/pull/12887, but with some different approaches.

Firstly, the `current_user` variable is cleanup from some recent merges. Then, I remove noqa marks on imports.

I believe that verifying that imports work, and that requests work should be sufficient to say this is working.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/api/generics.py
+++ b/awx/api/generics.py
@@ -6,7 +6,6 @@ import inspect
 import logging
 import time
 import uuid
-import urllib.parse
 
 # Django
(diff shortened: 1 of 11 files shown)


relrod (approved):
lgtm

Issue #12936, opened 2022-09-21
add validation for hostname field for the Instance object


TheRealHaoLiu:
current hostname field for the instance table have no validation and this allows user to put invalidate value into the field

TheRealHaoLiu:
`[a-z0-9]([-a-z0-9]*[a-z0-9])?` may be a good enough regex

TheRealHaoLiu:
this warrants additional discussion due to potential impact for existing install

FYI @sarabrajsingh

sarabrajsingh:
unit tests for supporting:
- [x] ipv4
- [x] ipv6
- [x] simple hostname
- [x] hostnames with domains or subdomains

TheRealHaoLiu:
conversation with @Zokormazo and @AlanCoding confirm that i is save to add this validation to the API serializer 
and we should add it 

FYI @jbradberry

AlanCoding:
I also want to link https://github.com/ansible/awx/issues/3384, as it may involve effectively the same or similar logic.

Issue #3384, opened 2019-03-07
Add validation for host names and group names


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
 - API

##### SUMMARY
See https://github.com/ansible/ansible/pull/52748

Ansible core is moving to a default behavior where group name characters which are not alphanumeric get automatically replaced with underscores.

In this world, we should not allow users to enter names with such characters.

Inventory sources, those should be covered by Ansible core behavior. Manually entered names should be validated.

##### ADDITIONAL INFORMATION

We have also hesitated to valid host names, even though many syntaxes will break playbooks. I figure I would throw that in here, but we can separate it.

unlikelyzero:
@AlanCoding should we have client-side validation in the UI as well?

AlanCoding:
Good question, normally I prefer that the corner cases can just be handled by the API, because it's not likely that the user will hit it and we maintain just 1 message.

However, this is both extremely probable to hit, and is extremely high-profile in terms of its position within the overall app. Existing paradigms would allow for client side validation here.

AlanCoding:
adding bug label because I expect it is possible to create hosts with duplicate / conflicting names which will lead to outright errors.

AlanCoding:
@sarabrajsingh thinking on this, I'm now leaning towards the idea that we should have a shared _serializer_ field class for this case, so it's not enforced on the model in any case.

PR #12979, opened 2022-09-29
hostname validation in InstanceSerializer

sarabrajsingh:
##### SUMMARY

Adding first iteration of hostname validation to InstanceSerializer
Addresses:

1. https://github.com/ansible/awx/issues/12936
2. https://github.com/ansible/awx/issues/3384

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 - UI
 - CLI
 - Docs

##### AWX VERSION

```
awx: 21.6.1.dev137+g5db777cfcc
```

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -29,6 +29,7 @@ from django.utils.translation import gettext_lazy as _
 from django.utils.encoding import force_str
 from django.utils.text import capfirst
 from django.utils.timezone import now
+from django.core.validators import RegexValidator, MaxLengthValidator
 
 # Django REST Framework
 from rest_framework.exceptions import ValidationError, PermissionDenied
@@ -119,6 +120,9 @@ from awx.main.validators import vars_validate_or_raise
 
 from awx.api.versioning import reverse
 from awx.api.fields import BooleanNullField, CharNullField, ChoiceNullField, VerbatimField, DeprecatedCredentialField
+
+# AWX Utils
+from awx.api.validators import HostnameRegexValidator
 
 logger = logging.getLogger('awx.api.serializers')
 
@@ -4921,6 +4925,18 @@ class InstanceSerializer(BaseSerializer):
         extra_kwargs = {
             'node_type': {'initial': Instance.Types.EXECUTION, 'default': Instance.Types.EXECUTION},
             'node_state': {'initial': Instance.States.INSTALLED, 'default': Instance.States.INSTALLED},
+            'hostname': {
+                'validators': [
+                    MaxLengthValidator(limit_value=255),
+                    RegexValidator(
+                        regex='^localhost$|^127(?:\.[0-9]+){0,2}\.[0-9]+$|^(?:0*\:)*?:?0*1$',
+                        flags=re.IGNORECASE,
+                        inverse_match=True,
+                        message="hostname cannot be localhost or 127.0.0.1",
+                    ),
+                    HostnameRegexValidator(),
+                ],
+            },
         }
 
     def get_related(self, obj):
@@ -4991,6 +5007,10 @@ class InstanceSerializer(BaseSerializer):
         return value
 
     def validate_hostname(self, value):
+        """
+        - Hostname cannot be "localhost" - but can be something like localhost.domain
+        - Cannot change the hostname of an-already instantiated & initialized Instance object
+        """
         if self.instance and self.instance.hostname != value:
             raise serializers.ValidationError("Cannot change hostname.")
 
--- a/awx/api/validators.py
+++ b/awx/api/validators.py
@@ -1,0 +1,55 @@
+import re
+
+from django.core.validators import RegexValidator, validate_ipv46_address
+from django.core.exceptions import ValidationError
+
+
+class HostnameRegexValidator(RegexValidator):
+    """
+    Fully validates a domain name that is compliant with norms in Linux/RHEL
+        - Cannot start with a hyphen
+        - Cannot begin with, or end with a "."
+        - Cannot contain any whitespaces
+        - Entire hostname is max 255 chars (including dots)
+        - Each domain/label is between 1 and 63 characters, except top level domain, which must be at least 2 characters
+        - Supports ipv4, ipv6, simple hostnames and FQDNs
+        - Follows RFC 9210 (modern RFC 1123, 1178) requirements
+
+    Accepts an IP Address or Hostname as the argument
+    """
+
+    regex = '^[a-z0-9][-a-z0-9]*$|^([a-z0-9][-a-z0-9]{0,62}[.])*[a-z0-9][-a-z0-9]{1,62}$'
+    flags = re.IGNORECASE
+
+    def __call__(self, value):
+        regex_matches, err = self.__validate(value)
+        invalid_input = regex_matches if self.inverse_match else not regex_matches
+        if invalid_input:
+            if err is None:
+                err = ValidationError(self.message, code=self.code, params={"value": value})
+            raise err
+
+    def __str__(self):
+        return f"regex={self.regex}, message={self.message}, code={self.code}, inverse_match={self.inverse_match}, flags={self.flags}"
+
+    def __validate(self, value):
+
+        if ' ' in value:
+            return False, ValidationError("whitespaces in hostnames are illegal")
+
+        """
+        If we have an IP address, try and validate it.
+        """
+        try:
+            validate_ipv46_address(value)
+            return True, None
+        except ValidationError:
+            pass
+
+        """
+        By this point in the code, we probably have a simple hostname, FQDN or a strange hostname like "192.localhost.domain.101"
+        """
+        if not self.regex.match(value):
+            return False, ValidationError(f"illegal characters detected in hostname={value}. Please verify.")
+
+        return True, None
--- a/awx/main/tests/functional/api/test_instance.py
+++ b/awx/main/tests/functional/api/test_instance.py
@@ -4,6 +4,8 @@ from awx.api.versioning import reverse
 from awx.main.models.activity_stream import ActivityStream
 from awx.main.models.ha import Instance
 
+from django.test.utils import override_settings
+
 
 INSTANCE_KWARGS = dict(hostname='example-host', cpu=6, memory=36000000000, cpu_capacity=6, mem_capacity=42)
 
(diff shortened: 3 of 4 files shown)


awx/main/tests/functional/api/test_instance.py
```diff
@@ -54,3 +54,27 @@ def test_health_check_usage(get, post, admin_user):
...
+
+
+def test_hostnames(post, admin_user):
+    url = reverse('api:instance_list')
+    for value in ["foobar", "foo.bar.baz"]:
+        post(url=url, user=admin_user, data={"hostname": value}, expect=201)
+    for value in ["--yoooo", "$3$@foobar@#($!@#*$"]:
+        post(url=url, user=admin_user, data={"hostname": value}, expect=400)
+    for value in ["localhost", "127.0.0.1"]:
+        post(url=url, user=admin_user, data={"hostname": value}, expect=400)
```
AlanCoding: I think these tests just need a `mocker.patch` to make it believe it's an OCP deployment... or maybe a use of `override_settings`

sarabrajsingh: `mock.patch('django.conf.settings')` didn't work but `with override_settings(IS_K8S=True)` worked!!!

AlanCoding: Yeah, mock is good to patch a simple method, and less good at patching data. Settings is more data than methods, balanced on top of a gazillion wrappers, hacks, and customizations.

awx/api/utils/common.py
```diff
@@ -0,0 +1,62 @@
+from django.core.validators import RegexValidator, validate_ipv46_address
```
AlanCoding: I'd prefer to add the file at awx/api/validators.py or something like that.

AlanCoding (approved):

awx/api/serializers.py
```diff
@@ -4980,9 +4983,18 @@ def validate_node_state(self, value):
...
     def validate_hostname(self, value):
+        """
+        - Hostname cannot be "localhost" - but can be something like localhost.domain
+        - Cannot change the hostname of an-already instantiated & initialized Instance object
+        """
         if self.instance and self.instance.hostname != value:
             raise serializers.ValidationError("Cannot change hostname.")
 
+        HostnameRegexValidator(regex="[a-z0-9]([-a-z0-9]*[a-z0-9])?", flags=re.IGNORECASE)(value)
+        HostnameRegexValidator(regex="^localhost$|^127.0.0.1$", flags=re.IGNORECASE, inverse_match=True, message="hostname cannot be localhost or 127.0.0.1")(
```
jbradberry: In my opinion we also need to exclude the IPv6 localhost.

awx/api/validators.py
```diff
@@ -0,0 +1,64 @@
+from django.core.validators import RegexValidator, validate_ipv46_address
+from django.core.exceptions import ValidationError
+
+
+class HostnameRegexValidator(RegexValidator):
+    """
+    Fully validates a domain name that is compliant with norms in Linux/RHEL
+        - Cannot start with a hypen
```
jbradberry: hyphen

awx/api/validators.py
```diff
@@ -0,0 +1,64 @@
...
+        - Cannot being with, or end with a "."
+        - Cannot contain any whitespaces
+        - Entire hostname is max 255 chars (including dots)
+        - Each domain/label is between 1 and 63 characters, except top level domain, which must be at least 2 characters
+        - Supports ipv4, ipv6, simple hostnames and FQDNs
+        - Follows RFC 9210 (modern RFC 1123, 1178) requirements
+
+    Excepts an IP Address or Hostname as the argument
```
jbradberry: Excepts -> Accepts?

awx/api/serializers.py
```diff
@@ -4980,9 +4983,18 @@ def validate_node_state(self, value):
...
 
     def validate_hostname(self, value):
+        """
+        - Hostname cannot be "localhost" - but can be something like localhost.domain
+        - Cannot change the hostname of an-already instantiated & initialized Instance object
+        """
         if self.instance and self.instance.hostname != value:
             raise serializers.ValidationError("Cannot change hostname.")
 
+        HostnameRegexValidator(regex="[a-z0-9]([-a-z0-9]*[a-z0-9])?", flags=re.IGNORECASE)(value)
```
jbradberry: I believe application of these validation objects should be left to `Meta.extra_kwargs`, IIRC

awx/api/serializers.py
jbradberry: Also, I suspect that we can get by with just a plain `RegexValidator` for these (and really just to handle the IPv6 localhost case).

awx/api/validators.py
```diff
@@ -0,0 +1,64 @@
...
+
+        if is_ip_addr:
+            if self.regex.match(value):
+                return True, None
+
+        """
+        By this point in the code, we probably have a simple hostname, FQDN or a strange hostname like "192.localhost.domain.101"
+        """
+        if len(value) > 255:
+            return False, ValidationError("hostname cannot be longer than 255 characters")
```
jbradberry: I believe we should lean on Django's `MaxLengthValidator` for this, as one of the list of validator objects applied via `extra_kwargs`.  https://docs.djangoproject.com/en/3.2/ref/validators/#maxlengthvalidator

awx/api/serializers.py
jbradberry: I think for this validator we should override the `regex` and `flags` in `__init__()` instead of requiring them to be passed in.  If as I suspect the localhost checks can be done separately with a plain `RegexValidator`, we won't need the flexibility to swap out different regexes.

awx/api/serializers.py
sarabrajsingh: this is true. I can definitely leverage the OOTB `RegexValidator` for `localhost` and `127.0.0.1`

`localhost` in ipv6 is `0000:0000:0000:0000:0000:0000:0000:0001` so it will pass both validators.

awx/api/validators.py
sarabrajsingh: **Expects - but yeah `Accepts` would be better

awx/api/serializers.py
sarabrajsingh: should we leave this flexibility in, in case this validator might be used elsewhere in `awx`?

awx/api/validators.py
```diff
@@ -0,0 +1,64 @@
...
+
+        for i, label in enumerate(value.split(".")):
```
jbradberry: I feel like most if not all of these conditions can be captured by the regex.  Off the top of my head, something like `([a-z0-9][-a-z0-9]{0,62}[.])*[a-z0-9][-a-z0-9]{1,}`.  Including the exclusion of spaces up on lines 32-33.

awx/api/validators.py
```diff
@@ -0,0 +1,64 @@
+from django.core.validators import RegexValidator, validate_ipv46_address
+from django.core.exceptions import ValidationError
+
+
+class HostnameRegexValidator(RegexValidator):
+    """
+    Fully validates a domain name that is compliant with norms in Linux/RHEL
+        - Cannot start with a hypen
+        - Cannot being with, or end with a "."
```
jbradberry: begin

awx/api/serializers.py
jbradberry: You can compress ipv6 localhost by replacing any combinations of the zeros with empty strings, so you can have a bunch of colons and `1`, or various combinations thereof.

awx/api/serializers.py
```diff
@@ -4913,7 +4917,19 @@ class Meta:
...
-        extra_kwargs = {'node_type': {'initial': 'execution'}, 'node_state': {'initial': 'installed'}}
+        extra_kwargs = {
+            'node_type': {'initial': 'execution'},
+            'node_state': {'initial': 'installed'},
+            'hostname': {
+                'validators': [
+                    MaxLengthValidator(limit_value=255),
+                    HostnameRegexValidator(),
+                    RegexValidator(
+                        regex='^localhost$|^127.0.0.1$', flags=re.IGNORECASE, inverse_match=True, message="hostname cannot be localhost or 127.0.0.1"
```
jbradberry: don't we still have to include the IPv6 localhost regex?

jbradberry: also I suspect that this one should go 2nd in the list, then the hostname validator

awx/api/validators.py
```diff
@@ -0,0 +1,69 @@
...
+        - Supports ipv4, ipv6, simple hostnames and FQDNs
+        - Follows RFC 9210 (modern RFC 1123, 1178) requirements
+
+    Accepts an IP Address or Hostname as the argument
+    """
+
+    def __init__(self, *args, **kwargs):
+
+        if 'regex' not in kwargs.keys():
+            self.regex = '^[a-z0-9][-a-z0-9]*$|^([a-z0-9][-a-z0-9]{0,62}[.])*[a-z0-9][-a-z0-9]{1,62}$'
```
jbradberry: I think the `super` call will wind up not playing well with this (and the flags below), at least under some circumstances.

Maybe just do a fresh dict with the regex and flags, then `new_kwargs.update(kwargs)`?

jbradberry: Well, ok, thinking it through a bit more I think it won't matter

awx/api/validators.py
```diff
@@ -0,0 +1,69 @@
...
+            if err is None:
+                err = ValidationError(self.message, code=self.code, params={"value": value})
+            raise err
+
+    def __str__(self):
+        return f"regex={self.regex}, message={self.message}, code={self.code}, inverse_match={self.inverse_match}, flags={self.flags}"
+
+    def __validate(self, value):
+
+        # import sdb; sdb.set_trace();
```
jbradberry: oops

awx/api/validators.py
```diff
@@ -0,0 +1,69 @@
...
+
+        if len(value.split(" ")) > 1:
```
jbradberry: if we are going to continue to make this flexible (which I think we shouldn't -- the semantics of valid hostnames aren't going to change between usages), this can be expressed more elegantly as `if ' ' in value:`.

awx/api/validators.py
```diff
@@ -0,0 +1,67 @@
+import re
+
+from django.core.validators import RegexValidator, MaxLengthValidator, validate_ipv46_address
```
jbradberry: `MaxLengthValidator` is unused here.

awx/api/validators.py
```diff
@@ -0,0 +1,67 @@
...
+        try:
+            validate_ipv46_address(value)
+            return True, None
+        except ValidationError:
+            pass
+
+        """
+        By this point in the code, we probably have a simple hostname, FQDN or a strange hostname like "192.localhost.domain.101"
+        """
+        for i, label in enumerate(value.split(".")):
```
jbradberry: I'm still thinking that this all logic has been folded into the regex.

awx/api/serializers.py
```diff
@@ -4921,6 +4925,18 @@ class Meta:
         extra_kwargs = {
             'node_type': {'initial': Instance.Types.EXECUTION, 'default': Instance.Types.EXECUTION},
             'node_state': {'initial': Instance.States.INSTALLED, 'default': Instance.States.INSTALLED},
+            'hostname': {
+                'validators': [
+                    MaxLengthValidator(limit_value=255),
+                    RegexValidator(
+                        regex='^localhost$|^127.0.0.1$|^0000:0000:0000:0000:0000:0000:0000:0001$',
```
jbradberry: https://stackoverflow.com/questions/8426171/what-regex-will-match-all-loopback-addresses has some ideas about how to capture the optional removal of various permutations of the zeros from the IPv6 localhost.

sarabrajsingh: this is a great find. will implement.

jbradberry (approved):

CFSNM (approved):

Issue #12946, opened 2022-09-23
improve waiting behavior for de-provisioning external remote execution node


TheRealHaoLiu:
currently in our implementation, de-provisioning a node starts a task that will wait indefinitely for all of the jobs to drain from the node 

we dont fully understand the negative path impact of this (if the job that's running is stuck)

this may potentially trigger some "timeout" causing the de-provisioning task to be reaped.

we might also want to introduce a mechanism for FORCE de-provisioning that bypass the checks

TheRealHaoLiu:
@AlanCoding what do you think about the current behavior

TheRealHaoLiu:
@AlanCoding agree that this can be address this issue post feature branch merge

AlanCoding:
Yep, feature branch in, :+1: definitely want to take our time and address this properly.

> we might also want to introduce a mechanism for FORCE de-provisioning that bypass the checks

I need to get some background information here - when you tell the system to de-provision a node, does it cancel the jobs on the instance? If not, I understand, but we need to consider the definition of "force" here.

Absent other information, I assume that the node we are de-provisioning can be communicated with. If we have communication, then de-provisioning is focusing on the symptom, not the sickness. The problem is that jobs are still running, so the first-level intervention is cancel the jobs. With recent changes, canceling sends a signal to the controller node's dispatcher, which sends SIGTERM to the control process, which issues the `receptorctl` command to cancel the work unit, which becomes a SIGTERM on the runner worker process.

If anything in this process is stalled, we need to debug that particular part. Maybe lack of response to SIGTERM needs to be followed by a SIGKILL at some point. We can do things like that. If you totally lose connection to an execution node, then at some point you may need to kick it out of the mesh, which would be de-provisioning, this should just be the absolute last resort.

AlanCoding:
Looking at this again

https://github.com/ansible/awx/blob/9c2185c68f25646f80fea5ce602540e7eb3f8e67/awx/main/tasks/receptor.py#L690

Confirming, this seems to have the intention to let the jobs follow their natural course until completion. So I would start with the idea of canceling them, and applying some appropriate follow-on wait steps.

TheRealHaoLiu:
so currently there are these cases we need to discuss 
- waiting for the job to finish with intend of removing the node when no job is running (deprovisioning)
- canceling all job with intend of removing the node when all job are canceled (this is something that customer can manually do already)
- just delete i dont care about the currently running job just delete (forced removal)

[...]

PR #12980, opened 2022-09-29
Handle finalization of node deprovisioning from within the task manager

jbradberry:
##### SUMMARY
Instead of doing a potentially infinite wait-loop looking for all of the jobs to complete on nodes marked as deprovisioning, instead check on this condition from within the task manager, which deals with this kind of thing already.

related #12946 

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -33,6 +33,7 @@ from awx.main.models import (
     WorkflowJobTemplate,
 )
 from awx.main.scheduler.dag_workflow import WorkflowDAG
+from awx.main.tasks.receptor import remove_deprovisioned_node
 from awx.main.utils.pglock import advisory_lock
 from awx.main.utils import (
     get_type_for_model,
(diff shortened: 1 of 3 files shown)


awx/main/scheduler/task_manager.py
```diff
@@ -686,14 +687,27 @@ def task_needs_capacity(self, task, tasks_to_update_job_explanation):
...
+        # Also, discover jobs that are in running state but aren't on an execution node
         # that we know about; this is a fairly rare event, but it can occur if you,
         # for example, SQL backup an awx install with running jobs and restore it
         # elsewhere
-        for j in UnifiedJob.objects.filter(
-            status__in=['pending', 'waiting', 'running'],
-        ).exclude(execution_node__in=Instance.objects.exclude(node_type='hop').values_list('hostname', flat=True)):
+
+        instances = Instance.objects.only('hostname', 'node_state')
+        jobs = UnifiedJob.objects.filter(status__in=['pending', 'waiting', 'running']).select_related('instance_group')
```
AlanCoding, line 698: This one line is my hangup. The last development cycle saw a massive improvement in the job lifecycle, and the time until the dispatcher starts a job. This code lines in that critical path, as any delay here is a delay submitting to pg_notify. Before this, the `.exclude` would at least prevent returning the bulk of the active jobs (essentially all of them if we're not in an error state). There are a lot of hazards.

jbradberry:
Closing.  I think this is going to need more careful consideration when we get around to doing this.

Issue #12981, opened 2022-09-29
Saving facts_cache performance enhancements


fosterseth:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

New Feature

### Feature Summary

Saving facts for many hosts (think 30k+) after a single job run can take a long time. We may need to reconsider how we doing this

- add log statements for every code path involved in reading the facts. Probably has to be debug level, but we should have them.
- add update_fields to that .save() in https://github.com/ansible/awx/blob/ff6fb3229720a6ba93f18ba287fbe99359a17a26/awx/main/models/jobs.py#L909
- consider batch bulk updates so that we address the root timing problem

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

na

### Current results

na

### Sugested feature result

na

AlanCoding:
Related testing content

https://github.com/AlanCoding/utility-playbooks/tree/master/AWX12981

I wanted a fact with a randomized file so I that I could be sure that it changed with every run.

AlanCoding:
This may be somewhat of a duplicate of https://github.com/ansible/awx/issues/6807

jay-steurer:
Feature implemented. Closing it.

PR #12998, opened 2022-10-04
Bulk save facts, and move to before status change

AlanCoding:
##### SUMMARY
Connect https://github.com/ansible/awx/issues/12981

Also connect https://github.com/ansible/awx/issues/6807

~Now also connect https://github.com/ansible/awx/issues/13131~

Initial data seems promising, can save 1k hosts in 0.6 seconds, down from about 30 seconds. This seems sufficient for the purpose.

With that speedup, I thought appropriate to move it to before the status change. This should fundamentally resolve the issue of a workflow node trying to use facts from an upstream node. It comes at the cost of a little delay to a "successful" status... but again... that is much reduced from before. For several 10s of 1,000s of hosts, it can still come to ~30 seconds, but only if all the facts are changing... in which case this doesn't seem disproportionate.

also expected to resolve AAP-3065, for internal referencing purposes

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/models/jobs.py
+++ b/awx/main/models/jobs.py
@@ -860,6 +860,7 @@ class Job(UnifiedJob, JobOptions, SurveyJobMixin, JobNotificationMixin, TaskMana
     @log_excess_runtime(logger, debug_cutoff=0.01, msg='Job {job_id} host facts prepared for {written_ct} hosts, took {delta:.3f} s', add_log_data=True)
     def start_job_fact_cache(self, destination, log_data, timeout=None):
         self.log_lifecycle("start_job_fact_cache")
+        log_data['written_ct'] = 0
         os.makedirs(destination, mode=0o700)
 
         if timeout is None:
--- a/awx/main/models/jobs.py
+++ b/awx/main/models/jobs.py
@@ -871,6 +871,8 @@ class Job(UnifiedJob, JobOptions, SurveyJobMixin, JobNotificationMixin, TaskMana
             hosts = self._get_inventory_hosts(ansible_facts_modified__gte=timeout)
         else:
             hosts = self._get_inventory_hosts()
+
+        last_filepath_written = None
         for host in hosts:
             filepath = os.sep.join(map(str, [destination, host.name]))
             if not os.path.realpath(filepath).startswith(destination):
--- a/awx/main/models/jobs.py
+++ b/awx/main/models/jobs.py
@@ -860,6 +860,7 @@ class Job(UnifiedJob, JobOptions, SurveyJobMixin, JobNotificationMixin, TaskMana
     @log_excess_runtime(logger, debug_cutoff=0.01, msg='Job {job_id} host facts prepared for {written_ct} hosts, took {delta:.3f} s', add_log_data=True)
     def start_job_fact_cache(self, destination, log_data, timeout=None):
         self.log_lifecycle("start_job_fact_cache")
+        log_data['job_id'] = self.id
         log_data['written_ct'] = 0
         os.makedirs(destination, mode=0o700)
 
--- a/awx/main/models/jobs.py
+++ b/awx/main/models/jobs.py
@@ -44,7 +44,7 @@ from awx.main.models.notifications import (
     NotificationTemplate,
     JobNotificationMixin,
 )
-from awx.main.utils import parse_yaml_or_json, getattr_dne, NullablePromptPseudoField, polymorphic
+from awx.main.utils import parse_yaml_or_json, getattr_dne, NullablePromptPseudoField, polymorphic, log_excess_runtime
 from awx.main.fields import ImplicitRoleField, AskForField, JSONBlob, OrderedManyToManyField
 from awx.main.models.mixins import (
     ResourceMixin,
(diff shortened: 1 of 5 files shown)


awx/main/models/jobs.py
```diff
@@ -869,15 +871,28 @@ def start_job_fact_cache(self, destination, modification_times, timeout=None):
...
                     os.chmod(f.name, 0o600)
                     json.dump(host.ansible_facts, f)
+                    written_ct += 1
             except IOError:
                 system_tracking_logger.error('facts for host {} could not be cached'.format(smart_str(host.name)))
                 continue
             # make note of the time we wrote the file so we can check if it changed later
             modification_times[filepath] = os.path.getmtime(filepath)
+        if written_ct:
+            dt = time.time() - start_time
```
fosterseth: this is interesting by itself. would be nice to elevate more logging from debugging to info based on performance criteria. If we do this kind of thing often we could build out a general method/system for it

AlanCoding: I started on a different kind of standard pattern for that here:

https://github.com/ansible/awx/blob/ff6fb3229720a6ba93f18ba287fbe99359a17a26/awx/main/utils/common.py#L1193

but that's insufficient for what I'm trying to do here, because I would really like to include the counts along with the time taken.

Maybe I could expand that method to take an optional log message, which would be the format statement. Then it would have to get the values from the return value... which here, would probably need to be a dictionary. This shouldn't be a problem because this never returned anything before.

AlanCoding: I developed this further, and put it into the decorator now.

awx/main/models/jobs.py
```diff
@@ -869,31 +881,46 @@ def start_job_fact_cache(self, destination, modification_times, timeout=None):
...
             except IOError:
                 system_tracking_logger.error('facts for host {} could not be cached'.format(smart_str(host.name)))
                 continue
-            # make note of the time we wrote the file so we can check if it changed later
-            modification_times[filepath] = os.path.getmtime(filepath)
-
-    def finish_job_fact_cache(self, destination, modification_times):
+        # make note of the time we wrote the last file so we can check if any file changed later
+        if last_filepath_written:
+            return os.path.getmtime(last_filepath_written)
```
AlanCoding, line 893: @jbradberry this is a scaled-back version of what I was trying to do in https://github.com/ansible/awx/pull/11641

We were passing this `modification_times` dict, which has a modtime for every host in the inventory with facts... so, ballpark that number at 35k keys... but really more (based on known case reports). Then that has to remain in memory for the duration of the job for comparison after it has finished.

Here, I replace the _dictionary_ with a _single value_. This fits with the goal of the rest of this PR - which is to addresses _scaling_ problems for facts with large inventories.

There's also no drawback I can see for this. We write all facts. Stop. We run the job. Stop. We read the facts. There's no way that we can _write the last fact_ before we _read the first fact_.

awx/main/models/jobs.py
```diff
@@ -844,22 +845,33 @@ def get_notification_templates(self):
...
         self.log_lifecycle("start_job_fact_cache")
         os.makedirs(destination, mode=0o700)
-        hosts = self._get_inventory_hosts()
         if timeout is None:
             timeout = settings.ANSIBLE_FACT_CACHE_TIMEOUT
         if timeout > 0:
             # exclude hosts with fact data older than `settings.ANSIBLE_FACT_CACHE_TIMEOUT seconds`
             timeout = now() - datetime.timedelta(seconds=timeout)
-            hosts = hosts.filter(ansible_facts_modified__gte=timeout)
+            hosts = hosts = self._get_inventory_hosts(ansible_facts_modified__gte=timeout)
```
jbradberry: hosts = hosts?  Looks like maybe a search and replace error.

AlanCoding: good catch, thank you

awx/main/tasks/jobs.py
```diff
@@ -1066,22 +1057,31 @@ def pre_run_hook(self, job, private_data_dir):
...
+        if job.use_fact_cache:
+            self.facts_write_time = self.instance.start_job_fact_cache(os.path.join(private_data_dir, 'artifacts', str(job.id), 'fact_cache'))
+
     def build_project_dir(self, job, private_data_dir):
         self.sync_and_copy(job.project, private_data_dir, scm_branch=job.scm_branch)
 
-    def final_run_hook(self, job, status, private_data_dir, fact_modification_times):
-        super(RunJob, self).final_run_hook(job, status, private_data_dir, fact_modification_times)
-        if not private_data_dir:
+    def post_run_hook(self, job, status):
```
AlanCoding, line 1068: @relrod I wanted to point this out to you because it changes the method interface here, and the way the variables are passed.

In the long run, I don't think there should be any python objects passed from the job preparation into the `post_run_hook` or `final_run_hook`. This still doesn't satisfy that because it saves the last modification time to onto the task (`BaseTask`) instance. My idea for that is that we save a "reference" file and obtain the modification time from that, instead of passing around in-memory.

This is important, because if we ever _de-couple_ job prep & starting from the job wrapup logic, we can't directly pass anything.

awx/main/models/jobs.py
jbradberry: I think that makes sense.

jbradberry (approved):

PR #13005, opened 2022-10-05
Fix undefined launchConfig error in schedule form

marshmalien:
##### SUMMARY

Addresses #13002 

##### ISSUE TYPE

##### COMPONENT NAME

 - UI

### DIFF
--- a/awx/ui/src/screens/ManagementJob/ManagementJob.js
+++ b/awx/ui/src/screens/ManagementJob/ManagementJob.js
@@ -180,8 +180,6 @@ function ManagementJob({ setBreadcrumb }) {
                 loadSchedules={loadSchedules}
                 loadScheduleOptions={loadScheduleOptions}
                 setBreadcrumb={setBreadcrumb}
-                launchConfig={{}}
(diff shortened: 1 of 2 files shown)


awx/ui/src/screens/ManagementJob/ManagementJob.js
```diff
@@ -180,8 +180,6 @@ function ManagementJob({ setBreadcrumb }) {
                 loadSchedules={loadSchedules}
                 loadScheduleOptions={loadScheduleOptions}
                 setBreadcrumb={setBreadcrumb}
-                launchConfig={{}}
-                surveyConfig={{}}
```
marshmalien, line 184: Removed these lines because it was causing an infinite re-render in the schedule form. We check `launchConfig` in ScheduleForm's useEffect dependency array and the empty object is causing hairy behavior because of the strict equivalency check.

mabashian (approved):
💯

akus062381 (approved):

keithjgrant (approved):

PR #13026, opened 2022-10-10
Install receptorctl in instance install playbook

TheRealHaoLiu:

##### SUMMARY
install receptorctl when installing receptor on remote execution node 

use for receptor reload and debugging

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### AWX VERSION

```
awx: 21.7.1.dev31+g074612f9e5
```

##### ADDITIONAL INFORMATION

```

```


jbradberry (approved):

awx/api/templates/instance_install_bundle/install_receptor.yml
```diff
@@ -15,4 +15,8 @@
       pip:
         name: ansible-runner
         executable: pip3.9
+    - name: Install receptorctl
+      pip:
+        name: receptorctl
```
relrod, line 22: This will be slightly faster if you do it as part of the previous task and pass a list instead of a string (which gets turned into a list anyway internally).

Something like:
```yaml
- name: Install dependencies from pip
  pip:
    name:
      - ansible-runner
      - receptorctl
    executable: pip3.9
```

jbradberry: Agreed.  I almost said something.

PR #13030, opened 2022-10-10
Adding GH action to lock the requirements_dev.txt file on release branch creation

john-westcott-iv:
##### ISSUE TYPE

##### COMPONENT NAME

 - Other

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

Here is an example of a commit created by this action: https://github.com/john-westcott-iv/github-file-test/commit/22ec93b648eaf9b6f468bdb78344ade5c0cc78a5

```

```

### DIFF
--- a/.github/lock_dev_requirements.py
+++ b/.github/lock_dev_requirements.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python
+
+# Run by workflows/lock_dev_requirements.yml
+
+import re
+
+pinned_versions = {}
+with open('./locked_requirements_dev.txt', 'r') as f:
+    for line in f.read().splitlines():
+        (package, version) = line.split('==')
+        pinned_versions[package] = version
+
+lines = []
+errors = False
+with open('./requirements_dev.txt', 'r') as f:
+    for line in f.read().splitlines():
+        comment = ''
+        if '#' in line:
+            package_version, comment = line.split('#', 2)
+            comment = ' # {}'.format(comment)
+            package_version = package_version.strip()
+        else:
+            package_version = line
+
+        if not package_version or package_version.startswith('git'):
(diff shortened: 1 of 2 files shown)


.github/workflows/lock_dev_requirements.yml
```diff
@@ -0,0 +1,40 @@
+---
+name: Lock Dev Requirements
+on:
+  create:
+    branches:
+      - release_**
```
AlanCoding, line 6: Since this only runs on creation of these branches I think this is fine?

The commit seems to land in the right branch in https://github.com/john-westcott-iv/github-file-test, so I think it'll work as intended. @shanemcd are you good with all this?

.github/lock_dev_requirements.py
```diff
@@ -0,0 +1,44 @@
...
+            #   then append the whole line back to the list of lines.
+            lines.append(line)
+            continue
+
+        package = re.split(r'[!=<>]+', package_version)[0]
+        if package not in pinned_versions:
+            print(f"I have failed you, the package {package} was not in the list of pinned versions.")
+            errors = True
+        else:
+            lines.append(f'{package}=={pinned_versions[package]}{comment}')
```
AlanCoding, line 36: I don't feel like you _needed_ to preserve comments. Saving the outright `pip freeze` output would kind of be enough. Half of our comments are outdated in the first place, and we're not even talking about changing `devel` anyway.

john-westcott-iv: The pip freeze was much more extensible because there are other modules from the python install in it and made it hard to line up the modules because the lines were so drastically different. https://github.com/john-westcott-iv/github-file-test/commit/16986a4b803265a181e3d782368d1c170959ac5b

Issue #13033, opened 2022-10-11
Attempting to relaunch adhoc command that requires password throws error


mabashian:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

![adhoc_relaunch](https://user-images.githubusercontent.com/9889020/195105789-da7a8989-4731-48cd-9831-7ab285ea28c7.gif)

### AWX version

devel

### Select the relevant components

- [X] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

docker development environment

### Modifications

no

### Steps to reproduce

1. Create machine credential with username. Set password to "prompt on launch".
2. Create a host in inventory to run the 'ping' module against.
3. Select Inventory -> Hosts -> Check the box next to target host.
4. Click "RUN COMMAND" button.
5. Select "ping" module.
6. Select "default execution environment"
7. Select machine credential.
8. Launch.
9. Relaunch.

### Expected results

User should be prompted for password on relaunch

### Actual results

JS error thrown

PR #13034, opened 2022-10-11
Fixes bug re-launching adhoc command with passwords required

mabashian:
##### SUMMARY
Link https://github.com/ansible/awx/issues/13033

Adhoc commands don't have a description field so we need to protect against attempting to access the length.

##### ISSUE TYPE

##### COMPONENT NAME
 - UI


mabashian:
I found another bug while testing this out - will push another commit shortly.

keithjgrant (approved):

akus062381 (approved):

PR #13037, opened 2022-10-11
[collection] remove module defaults where API defaults are the same

relrod:
##### SUMMARY

Providing defaults for API parameters where the API already provides defaults leads to some confusing scenarios, because we end up always sending those collection-defaulted fields in the request even if the field isn't provided by the user.

For example, we previously set the `scm_type` default to 'manual' and someone using the collection to update a project who does not explicitly include the `scm_type` every time they call the module, would inadvertently change the `scm_type` of the project back to 'manual' which is surprising behavior.

This change removes the collection defaults for API parameters, unless they differed from the API default. We let the API handle the defaults or otherwise ignore fields not given by the user so that the user does not end up changing unexpected fields when they use a module.

##### ISSUE TYPE

##### COMPONENT NAME
 - Collection

### DIFF
--- a/awx_collection/plugins/modules/execution_environment.py
+++ b/awx_collection/plugins/modules/execution_environment.py
@@ -84,6 +84,7 @@ def main():
         organization=dict(),
         credential=dict(default=''),
         state=dict(choices=['present', 'absent'], default='present'),
+        # NOTE: Default for pull differs from API (which is blank by default)
         pull=dict(choices=['always', 'missing', 'never'], default='missing'),
     )
 
--- a/awx_collection/plugins/modules/ad_hoc_command.py
+++ b/awx_collection/plugins/modules/ad_hoc_command.py
@@ -130,7 +130,7 @@ def main():
         diff_mode=dict(type='bool'),
         wait=dict(default=False, type='bool'),
         interval=dict(default=2.0, type='float'),
-        timeout=dict(default=None, type='int'),
+        timeout=dict(type='int'),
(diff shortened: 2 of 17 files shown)


john-westcott-iv (approved):
If the API has the same defaults I am fine removing these from the modules themselves. Once the tests are all passing this should be good to go.

AlanCoding:
I think the test failure you have is legit.

https://github.com/ansible/awx/blob/352c64b8c34979372843280dde34b46e08cdf03b/awx_collection/plugins/modules/project.py#L362-L369

It looks like these might be all the fields that previously had a default? In that case, it looks like they should be removed from `project_fields` assignment and put into the loop below that, where it ignores null values.

AlanCoding (approved):

awx_collection/plugins/modules/execution_environment.py
```diff
@@ -84,6 +84,7 @@ def main():
         organization=dict(),
         credential=dict(default=''),
         state=dict(choices=['present', 'absent'], default='present'),
+        # NOTE: Default for pull differs from API (which is blank by default)
```
john-westcott-iv, line 87: Should we add `''` as a choice?

john-westcott-iv (commented):
We should also leave a comment about this change in the README.md template.
I'd also say the PR has `Bug/Doc Fix` which indicates a z release but I think this should at least be a Y and maybe even a X (since it could break existing playbooks).

Issue #12966, opened 2022-09-27
Disable manual health check for control & hybrid nodes


AlanCoding:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

Enhancement to Existing Feature

### Feature Summary

We have an endpoint `/api/v2/instances/N/health_check/`, and this gives a 404 for hop nodes.

The link is also not shown in related links for hop nodes.

This issue argues that these should apply to control & hybrid nodes as well. They run health checks on a periodic basis so there is no need for the user to manually initiate it.

In other words, manually health checks should only be possible for execution nodes.

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

If you POST to that endpoint for these node types

### Current results

Right now it will submit a task to run a health check, but expectations are a little unclear, since the result is likely to come from the 20 second periodic task.

### Sugested feature result

The API will give a 404 for this endpoint for those instance types, and the UI will hide the button.

The health check button will only be shown for execution nodes.

AlexSCorey:
I believe this is already addressed in the UI

AlanCoding:
@AlexSCorey can this issue be closed now? Your PR addressed both the UI and the API.

PR #13041, opened 2022-10-11
Allows health checks on only execution nodes

AlexSCorey:
##### SUMMARY
Addresses #12966 

##### ISSUE TYPE 
 - New or Enhanced Feature

##### COMPONENT NAME
 - API
##### AWX VERSION

### DIFF
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -392,8 +392,8 @@ class InstanceHealthCheck(GenericAPIView):
     permission_classes = (IsSystemAdminOrAuditor,)
 
     def get_queryset(self):
+        return super().get_queryset().filter(node_type='execution')
         # FIXME: For now, we don't have a good way of checking the health of a hop node.
-        return super().get_queryset().exclude(node_type='hop')
 
     def get(self, request, *args, **kwargs):
         obj = self.get_object()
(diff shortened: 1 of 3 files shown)


awx/api/views/__init__.py
```diff
@@ -451,28 +456,12 @@ def post(self, request, *args, **kwargs):
                 if extra_field in runner_data:
                     data[extra_field] = runner_data[extra_field]
```
AlanCoding: You're not doing anything with `data` anymore.

You're also calling `execution_node_health_check(obj.hostname)`, not `execution_node_health_check.delay(obj.hostname)`, so it runs instantaneously. The point of `data` was to return the data _from_ the health check.

The response message _suggests_ that it is ran asynchronously, but you need to do a `.delay` call for that to be true. There should be an upper limit to how long it can take (20 or 40 seconds), which is why this was done synchronously to begin with, for background.

AlexSCorey: @AlanCoding I had done a bad rebase.  I have since fixed that and updated this PR.  Please re-review

awx/api/views/__init__.py
```diff
@@ -391,8 +391,10 @@ class InstanceHealthCheck(GenericAPIView):
     permission_classes = (IsSystemAdminOrAuditor,)
 
     def get_queryset(self):
+        obj = self.get_object()
+        if obj.node_type == 'execution':
+            return super().get_queryset()
```
AlanCoding: If the node type is not execution, this returns `None`, and I don't have any intuition about what happens in that case. Matters for a GET

awx/api/views/__init__.py
```diff
@@ -412,7 +412,10 @@ def post(self, request, *args, **kwargs):
 
             execution_node_health_check.apply_async([obj.hostname])
         else:
-            from awx.main.tasks.system import cluster_node_health_check
+            return Response(
+                {"error": f"Cannot run a health check on instances of type {obj.node_type}.  Health checks can only be run on execution nodes."},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
 
             cluster_node_health_check.apply_async([obj.hostname], queue=obj.hostname)
```
AlanCoding: This line (and maybe the method itself) now becomes unreachable and you should be free to remove it.

AlanCoding (approved):

Issue #13038, opened 2022-10-11
Topology view enabled/disabled node state resets when a user resizes browser window


kialam:
1. Enable/disable a node in the instances drawer in Topology view
2. Resize window.
3. Observe that the toggled node resets to its previous state

This behavior fixes itself on page refresh.

![Screen Shot 2022-10-12 at 7 48 44 AM](https://user-images.githubusercontent.com/2293210/195376531-fd290adc-28de-46d2-8c49-c65e3fb83fc9.png)
![node_enabed_resize](https://user-images.githubusercontent.com/2293210/195376968-1fe078d5-09b5-4579-83fd-251334919310.gif)

##### ISSUE TYPE

 - Bug, Docs Fix or other nominal change

##### COMPONENT NAME

 - UI

PR #13042, opened 2022-10-11
Fix enable/disable node state on browser resize.

kialam:
##### SUMMARY

Related #13038 

Pass `storedNodes` as a prop to MeshGraph. Update `storedNodes.current` within MeshGraph when a user toggles an instance. Update Topology View D3 graph by passing it the updated nodes to the web worker.

##### ISSUE TYPE

##### COMPONENT NAME

 - UI


akus062381 (approved):

AlexSCorey (approved):
Looks good to me.

PR #13044, opened 2022-10-12
[WIP] bugfix - allow edit hostname when org.max_hosts == active_count

sarabrajsingh:
##### SUMMARY

removed check that was disabling user from editing hostname of a host attached to an inventory when the max_hosts limit was reached for a particular organization.

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.6.1.dev183+g5e4590a2ed
```

##### TO-DO

- [x] functional tests for this functionality

### DIFF
--- a/awx/main/access.py
+++ b/awx/main/access.py
@@ -993,9 +993,6 @@ class HostAccess(BaseAccess):
         if data and 'name' in data:
             self.check_license(add_host_name=data['name'])
 
-            # Check the per-org limit
-            self.check_org_host_limit({'inventory': obj.inventory}, add_host_name=data['name'])
-
         # Checks for admin or change permission on inventory, controls whether
         # the user can edit variable data.
         return obj and self.user in obj.inventory.admin_role


awx/main/access.py
```diff
@@ -411,8 +415,8 @@ def check_org_host_limit(self, data, add_host_name=None):
                 )
             )
 
-        if add_host_name:
-            host_exists = Host.objects.filter(inventory__organization=org.id, name=add_host_name).exists()
+        if kwargs.get('add_host_name', None):
```
jbradberry: `.get(...)` automatically falls back to providing `None` when the key is not in the dict.

awx/main/access.py
```diff
@@ -411,8 +415,8 @@ def check_org_host_limit(self, data, add_host_name=None):
...
+            host_exists = Host.objects.filter(inventory__organization=org.id, name=kwargs.get('add_host_name')).exists()
```
jbradberry: Since the previous line proves that the key is there, it's ok to do the slightly shorter `kwargs['add_host_name']`

awx/main/access.py
```diff
@@ -994,7 +998,7 @@ def can_change(self, obj, data):
             self.check_license(add_host_name=data['name'])
 
             # Check the per-org limit
-            self.check_org_host_limit({'inventory': obj.inventory}, add_host_name=data['name'])
+            self.check_org_host_limit({'inventory': obj.inventory}, add_host_name=data['name'], action='change')
```
jbradberry: Since `check_org_host_limit` is short-circuiting out of doing any real work when using this kwarg by your changes, wouldn't it make sense to just remove this call entirely and not do the `action == 'change'` short-circuit, unless there is another usage I haven't gotten to yet?

sarabrajsingh: the only use case i could think of to justify keeping this code is when `active_count` > `org.max_hosts`. this should rarely happen and is an edge case, so i think the question becomes - do we need to keep this code around to encapsulate this edge case?

if not, then after thinking about it, i agree with you and we can just omit the call to `self.check_org_host_limit()`

jbradberry (approved):

PR #13048, opened 2022-10-12
fix: french vocabulary

grimlokason:
SUMMARY
- Correction on french vocabulary

ISSUE TYPE
- Incorect text in french

COMPONENT NAME
- UI


rooftopcellist:
FYI, we will likely not merge this PR directly. It will be communicated to our translators and will likely make it in to our next round of translations.  

Let's leave this PR open as a means of tracking it so that we remember to check the translations when they are pulled in.  

cc @AlexSCorey

AlexSCorey:
This work was included with https://github.com/ansible/awx/pull/13104

Issue #12999, opened 2022-10-04
Pg_notify / Psycop : InvalidParameterValue: payload string too long


Ghost0shell:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

I there,

We faced an issue on morning when having a huge amount of jobs pending in queue (next to a pending project update who has never started) . 

We noticed that when there are more than 500+ jobs in queue, Pg_notify is unable to push the payload in the database SELECT query because it's capped to 8000 bytes and we are over . 

Error log:
```
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/dispatch/worker/task.py", line 102, in perform_work
    result = self.run_callable(body)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/dispatch/worker/task.py", line 77, in run_callable
    return _call(*args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/scheduler/tasks.py", line 25, in task_manager
    run_manager(TaskManager, "task")
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/scheduler/tasks.py", line 20, in run_manager
    manager().schedule()
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/scheduler/task_manager.py", line 132, in schedule
    self._schedule()
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/scheduler/task_manager.py", line 58, in inner
    result = func(*args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/scheduler/task_manager.py", line 745, in _schedule
    self.process_tasks()
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/scheduler/task_manager.py", line 714, in process_tasks
    self.process_pending_tasks(pending_tasks)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/scheduler/task_manager.py", line 58, in inner
    result = func(*args, **kwargs)
[...]

PR #13061, opened 2022-10-13
Change role of scheduler DependencyManager, spawn fixed deps on creation

AlanCoding:
##### SUMMARY
This moves logic from `DependencyManager` into `create_unified_job`.

Also changes the _meaning_ of the unified job field `dependencies_processed` (but makes no schema change).
 - before: A value of `False` means the dependencies had not been spawned
 - now:  A value of `False` means the dependencies have not _finished_

Since spawning happens on creation, the state of "unspawned dependencies" is no longer a possible state (without errors)

I believe this is much more simple and has a lot of performance benefits, as even the `TaskManager` has to process fewer jobs each time.

Offshoot of tabled enhancements related to https://github.com/ansible/tower/issues/5943

##### Expected performance impacts

There is a drawback that a POST to the launch endpoint will take longer for jobs trigger update-on-launch behaviors. If a problem, I believe we could pay this back with some debugging and optimization, as we have done before.

This is balanced by greater scalability, although it only really matters inasmuch as someone uses `update_on_launch` features. In that case, this should speed up the `TaskManager` significantly. It leaves the `DependencyManager` with a finite set of queries (2) if none of them are erroring.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
Still trying to run more tests as of opening this.

### DIFF
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -2755,6 +2755,7 @@ class JobTemplateCallback(GenericAPIView):
         matching_hosts = self.find_matching_hosts()
         # If the host is not found, update the inventory before trying to
         # match again.
+        inventory_update_list = []
         inventory_sources_already_updated = []
         if len(matching_hosts) != 1:
             inventory_sources = job_template.inventory.inventory_sources.filter(update_on_launch=True)
@@ -2764,6 +2765,7 @@ class JobTemplateCallback(GenericAPIView):
                     # FIXME: Doesn't check for any existing updates.
                     inventory_update = inventory_source.create_inventory_update(**{'_eager_fields': {'launch_type': 'callback'}})
                     inventory_update.signal_start()
+                    inventory_update_list.append(inventory_update)
                     inventory_update_pks.add(inventory_update.pk)
             inventory_update_qs = models.InventoryUpdate.objects.filter(pk__in=inventory_update_pks, status__in=('pending', 'waiting', 'running'))
             # Poll for the inventory updates we've started to complete.
@@ -2802,10 +2804,13 @@ class JobTemplateCallback(GenericAPIView):
             kv['extra_vars'] = extra_vars_redacted
         kv['_prevent_slicing'] = True  # will only run against 1 host, so no point
         with transaction.atomic():
-            job = job_template.create_job(**kv)
+            if inventory_update_list:
+                job = job_template.create_job(available_deps=inventory_update_list, **kv)
+            else:
+                job = job_template.create_job(**kv)
 
         # Send a signal to signify that the job should be started.
-        result = job.signal_start(inventory_sources_already_updated=inventory_sources_already_updated)
+        result = job.signal_start()
         if not result:
             data = dict(msg=_('Error starting job!'))
             job.delete()
--- a/awx/main/analytics/subsystem_metrics.py
+++ b/awx/main/analytics/subsystem_metrics.py
@@ -196,8 +196,6 @@ class Metrics:
             SetIntM('task_manager_pending_processed', 'Number of pending tasks processed'),
             SetIntM('task_manager_tasks_blocked', 'Number of tasks blocked from running'),
             SetFloatM('task_manager_commit_seconds', 'Time spent in db transaction, including on_commit calls'),
-            SetFloatM('dependency_manager_get_tasks_seconds', 'Time spent loading pending tasks from db'),
-            SetFloatM('dependency_manager_generate_dependencies_seconds', 'Time spent generating dependencies for pending tasks'),
             SetFloatM('dependency_manager__schedule_seconds', 'Time spent in running the entire _schedule'),
             IntM('dependency_manager__schedule_calls', 'Number of calls to _schedule, after lock is acquired'),
             SetFloatM('dependency_manager_recorded_timestamp', 'Unix timestamp when metrics were last recorded'),
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -21,11 +21,6 @@ from django.contrib.contenttypes.models import ContentType
 from awx.main.dispatch.reaper import reap_job
 from awx.main.models import (
     Instance,
-    InventorySource,
-    InventoryUpdate,
-    Job,
-    Project,
-    ProjectUpdate,
     UnifiedJob,
     WorkflowApproval,
     WorkflowJob,
@@ -87,13 +82,7 @@ class TaskBase:
     @timeit
     def get_tasks(self, filter_args):
         wf_approval_ctype_id = ContentType.objects.get_for_model(WorkflowApproval).id
-        qs = (
-            UnifiedJob.objects.filter(**filter_args)
-            .exclude(launch_type='sync')
-            .exclude(polymorphic_ctype_id=wf_approval_ctype_id)
-            .order_by('created')
-            .prefetch_related('dependent_jobs')
-        )
+        qs = UnifiedJob.objects.filter(**filter_args).exclude(launch_type='sync').exclude(polymorphic_ctype_id=wf_approval_ctype_id).order_by('created')
         self.all_tasks = [t for t in qs]
 
     def record_aggregate_metrics(self, *args):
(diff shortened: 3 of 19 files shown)


awx/main/scheduler/task_manager.py
```diff
@@ -267,184 +262,17 @@ class DependencyManager(TaskBase):
...
     @timeit
     def _schedule(self):
-        self.get_tasks(dict(status__in=["pending"], dependencies_processed=False))
-
-        if len(self.all_tasks) > 0:
-            self.get_inventory_source_tasks()
-            self.process_tasks()
+        r = (
+            UnifiedJob.objects.filter(status='pending', dependencies_processed=False)
+            .exclude(dependent_jobs__status__in=ACTIVE_STATES)
```
AlanCoding: I realized that this might not actually be correct. Instead of `ACTIVE_STATES`, we only want `"successful"`, because canceled / error / failed dependencies should not progress a job to the next phase in the task manager kind of methods.

@fosterseth thinking back to https://github.com/ansible/awx/issues/12999, it wouldn't be hard to have the dependency manager absorb the work that the `handle_work_error` type of errback methods do now. We could do this update here for successful dependencies, and then do a second query where we get a list of all jobs that are both (1) in pending status and (2) have dependencies with failure status. Since you already have this running on a schedule, it would be a fool-proof way to mop up any cascading failures. Just an idea.

AlanCoding: I implemented what I mentioned here, but haven't removed the errbacks... following this through...

We pass dependent_jobs in `start_task`, but those basically come from `get_jobs_fail_chain` in the models.

In `TaskManagerInventoryUpdateMixin` and `TaskManagerProjectUpdateMixin` these pretty much equate to returning the reverse relationship objects - `unifiedjob_blocked_jobs`. The only exception is jobs giving related `project_update`, in which case I'm _sure_ we don't rely on this.

Having the `DependencyManager` run on a schedule is as airtight as we can get.

So to summarize, I'm ready to delete the use of `errbacks`, `handle_work_error`, and the vast network of inheritance starting from `TaskManagerUnifiedJobMixin`. This is a lot to nuke, so I'd like to check with others before fully carrying that out.

awx/main/scheduler/task_manager.py
```diff
@@ -267,185 +256,47 @@ class DependencyManager(TaskBase):
...
+            task.status = 'failed'
+            task.job_explanation = '\n'.join(messages)
+            task.save(update_fields=['status', 'job_explanation'])
+            task.websocket_emit_status('failed')
+            failed_job_ct += 1
+            already_failed.add(task.id)
+
+        advancing_job_ct = (
+            UnifiedJob.objects.filter(status='pending', dependencies_processed=False)
+            .exclude(dependent_jobs__status__in=ACTIVE_STATES + ('failed', 'error', 'canceled'))
```
fosterseth, line 291: is this the same as
`UnifiedJob.objects.filter(status='pending', dependencies_processed=False, dependent_jobs__status="successful")`

awx/main/scheduler/task_manager.py
fosterseth: green light from me

awx/main/scheduler/task_manager.py
AlanCoding: No, because of the case where a job has 2 dependencies (for example) and the 1st is successful and the 2nd is running (or something else). This says "exclude whenever a dependent job is running/failed/...", so we will correctly remove such a job from consideration (by virtue of its 2nd dep). But if you do the `.filter` suggested, that is saying "include jobs where any dependent job is successful", and would incorrectly include this example b/c of 1st dependency (successful). We need to gate on the status of all the parents, and although it's messier than I'd like, I don't see any easy alternative to this `.exclude`

awx/main/analytics/subsystem_metrics.py
```diff
@@ -196,8 +196,6 @@ def __init__(self, auto_pipe_execute=False, instance_name=None):
             SetIntM('task_manager_pending_processed', 'Number of pending tasks processed'),
             SetIntM('task_manager_tasks_blocked', 'Number of tasks blocked from running'),
             SetFloatM('task_manager_commit_seconds', 'Time spent in db transaction, including on_commit calls'),
-            SetFloatM('dependency_manager_get_tasks_seconds', 'Time spent loading pending tasks from db'),
-            SetFloatM('dependency_manager_generate_dependencies_seconds', 'Time spent generating dependencies for pending tasks'),
```
AlanCoding, line 200: Still need to remove from dashboard demo graphs.

awx/main/scheduler/task_manager.py
```diff
@@ -267,185 +256,47 @@ class DependencyManager(TaskBase):
...
+            # if we detect a failed or error dependency, go ahead and fail this
+            # task. The errback on the dependency takes some time to trigger,
+            # and we don't want the task to enter running state if its
+            # dependency has failed or errored.
+            task.status = 'failed'
+            task.job_explanation = '\n'.join(messages)
+            task.save(update_fields=['status', 'job_explanation'])
+            task.websocket_emit_status('failed')
+            failed_job_ct += 1
+            already_failed.add(task.id)
```
AlanCoding, line 287: Since I moved this stuff for failing jobs, we could separately track the timing and counts in metrics. This should _fairly rarely_ actually process anything, but it is very logically distinct from the other part (advancing jobs).

AlanCoding:
Status update: I'm pretty happy with this, but I just landed some big changes removing the subtasks in the `errback` method. I'll keep it in draft until the full tests finish against that. Before that, I saw 1 failure from the full API tests, but I understood it (have some related commits) and should be resolved by now.

awx/main/scheduler/task_manager.py
```diff
@@ -267,185 +256,47 @@ class DependencyManager(TaskBase):
...
-
     @timeit
     def _schedule(self):
-        self.get_tasks(dict(status__in=["pending"], dependencies_processed=False))
-
-        if len(self.all_tasks) > 0:
-            self.get_inventory_source_tasks()
-            self.process_tasks()
+        failing_job_qs = UnifiedJob.objects.filter(
+            status='pending', dependencies_processed=False, dependent_jobs__status__in=['failed', 'error']
```
AlanCoding, line 262: I want to call attention to the "canceled" status as a weird gap in queryset completeness here.

Here, we fail jobs with dependencies in the failed/error status. In the block after this, we advance pending jobs that don't have any dependencies in the failed/error/canceled status. What about canceled dependencies? This is saying that we don't fail them, and we don't advance them either.

That's right. Here, the `DependencyManager` doesn't do _anything_ with jobs that had a dependency get canceled. This is actually pre-existing behavior. It was moved and some lower-level details changed, but it was always the case that the task manager didn't do anything when dependencies got canceled. Why? Because that was/is handled in the `cancel` method, we propagate the cancel from the dependency to jobs that depend on it. That's what the (new name) `get_cancel_chain` loop does.

Would it be _easier_ to not propagate through the "hard" dependency chain in the `cancel` method, and do it here instead? Yes, I think it would be. It would be easier to follow (because this manages dependencies, seems only right to manage canceling through deps).

My only reason not to is to have a stopping place. I'm still maybe, kind of, considering it.

AlanCoding: Never mind, I've given up on this because it won't happen in a very satisfying way.

AlanCoding:
So @fosterseth brought up a concern that I had forgotten about - the chance that a burst of job launches could create multiple `update_on_launch` dependent updates. So a script to test that:

```python
        project = factories.project(scm_update_on_launch=True)
        jt = factories.job_template(project=project)
        project.wait_until_completed()

        initial_ct = project.get_related('project_updates').count

        number_of_jobs = 30
        threads = [threading.Thread(target=jt.launch, args=()) for i in range(0, number_of_jobs)]
        [t.start() for t in threads]
        [t.join() for t in threads]

        time.sleep(1)

        assert project.related.project_updates.get().count == initial_ct + 1
```

This confirms that this _is_ an issue unique to this branch, and was properly handled with the old `DependencyManager`. Logically, this is a lot easier to avoid if you do things while holding a global lock. We had discussion on this previously and I thought about a lot of options, so now I will test some of those. At least with initial testing, a simple `select_for_update` isn't working, and explaining why isn't easy. I still have a few more options to try, but having the request operate in a transaction actually seems to make this quite difficult.

AlanCoding:
^ these two concerns:

 - creation of duplicate updates
 - increased API timings

...may be fatal flaws of this approach.

I experimented with that script above, and I've found that it fails selectively for this approach for the reason that _some updates are created after others finish_ which is technically correct behavior, but misses the point. I don't want to do this as something that's slowing the system down, even in such a rare case.

I do like many other parts of this, and I'm thinking about a revised form where I don't spawn deps on creation, but still move task manager logic into the DependencyManager.

awx/api/views/__init__.py
```diff
@@ -2802,10 +2804,13 @@ def post(self, request, *args, **kwargs):
             kv['extra_vars'] = extra_vars_redacted
         kv['_prevent_slicing'] = True  # will only run against 1 host, so no point
         with transaction.atomic():
-            job = job_template.create_job(**kv)
+            if inventory_update_list:
+                job = job_template.create_job(available_deps=inventory_update_list, **kv)
```
kdelee, line 2808: I like this

AlanCoding:
Closing in favor of the more minimal PR

Issue #13015, opened 2022-10-07
Cached files allowing an inventory sync when the project sync failed


akus062381:
When a user creates an inventory sync using a project taken from a repo that has somehow cached the inventory files in that repo, BUT the project sync had subsequently failed, the inventory sync is still allowed to be created against that project. The inventory sync, of course, fails because the project was not synced successfully, and the failure will cause user confusion if this occurs.

This one is pretty tough to recreate. Discussions with UI devs resulted in two conclusions:
1. This is likely an issue on the API end and needs to be looked into further.
2. UI can bandaid this issue by preventing a user from being able to sync an inventory source off of a project that has a failed sync as its most recent sync.

For reference- in this screenshot, where '2hosts' appears, this is only appearing here because the files were cached. An unsuccessful project sync would only show /project_root in this field.
![image](https://user-images.githubusercontent.com/63685497/194575292-5aceff22-4d6f-4122-9a66-32b2ab79adfb.png)

relrod:
> The inventory sync, of course, fails because the project was not synced successfully

This behavior seems to depend on whether or not the Project has "Update revision on launch" set or not.

Here is what I think is happening after spending some time tracing bits and pieces of this code. I'm hoping maybe @AlanCoding can clear up anything I get wrong here.

If "Update revision on launch" is set on the project, when we trigger an inventory source update, we are actually launching _two_ `ProjectUpdate`s. We launch one (a 'check' job) because it gets added as a dependency job of the `InventoryUpdate`. I believe that in this case, when the PU dependency fails, the IU also bails out since the dependency wasn't "met". This is why in this case we see an IU failure. If the dependency *is* met, then we continue on, and we end up launching a second PU. I'll come back to that in a second.

When "Update revision on launch" is NOT set. We only launch the "second" PU mentioned above. There's no dependency added, so the IU won't fail due to an unmet dependency task.

That "second" `ProjectUpdate` is interesting. It's a 'run' job (as opposed to the dependency job which is a 'check' job, as mentioned above).

In `RunProjectUpdate#build_extra_vars_file`, we have this:

```python
        scm_branch = project_update.scm_branch
        if project_update.job_type == 'run' and (not project_update.branch_override):
            if project_update.project.scm_revision:
                scm_branch = project_update.project.scm_revision  # <--------
            elif not scm_branch:
                raise RuntimeError('Could not determine a revision to run from project.')
```

[...]

PR #13063, opened 2022-10-14
No InventoryUpdates when source Project is failed

relrod:
##### SUMMARY

Previously, in some cases, an InventoryUpdate sourced by an SCM project would still run and be successful even after the project it is sourced from failed to update. This would happen because the InventoryUpdate would revert the project back to its last working revision. This behavior is confusing and inconsistent with how we handle jobs (which just refuse to launch when the project is failed).

This change pulls out the logic that the job launch serializer and RunJob#pre_run_hook had implemented (independently) to check if the project is in a failed state, and puts it into a method on the Project model. This is then checked in the project launch serializer as well as the inventory update serializer, along with
SourceControlMixin#sync_and_copy as a fallback for things that don't run the serializer validation (such as scheduled jobs and WFJT jobs).

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -2222,8 +2222,7 @@ class InventorySourceUpdateView(RetrieveAPIView):
     def post(self, request, *args, **kwargs):
         obj = self.get_object()
         serializer = self.get_serializer(instance=obj, data=request.data)
-        if not serializer.is_valid():
-            return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+        serializer.is_valid(raise_exception=True)
         if obj.can_update:
             update = obj.update()
             if not update:
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -2221,6 +2221,9 @@ class InventorySourceUpdateView(RetrieveAPIView):
 
     def post(self, request, *args, **kwargs):
         obj = self.get_object()
+        serializer = self.get_serializer(instance=obj, data=request.data)
+        if not serializer.is_valid():
+            return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
         if obj.can_update:
             update = obj.update()
             if not update:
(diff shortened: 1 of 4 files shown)


relrod:
Tests in ansible/tower-qa#7767.


@jbradberry to review the serializer bits
@AlanCoding to review the tasks/jobs.py bits

awx/main/tasks/jobs.py
```diff
@@ -1056,18 +1059,18 @@ def pre_run_hook(self, job, private_data_dir):
...
             # cache smart inventory memberships so that the host_filter query is not
             # ran inside of the event saving code
             update_smart_memberships_for_inventory(job.inventory)
 
     def build_project_dir(self, job, private_data_dir):
-        self.sync_and_copy(job.project, private_data_dir, scm_branch=job.scm_branch)
+        try:
+            self.sync_and_copy(job.project, private_data_dir, scm_branch=job.scm_branch)
+        except RuntimeError as e:
+            self.update_model(job.pk, status='failed', job_explanation=str(e))
```
AlanCoding: Super nit pick comment, but I'd like to put the `update_model` call for job_explanation into the main `sync_and_copy` method for consistency between the two types.

relrod: done

AlanCoding (approved):
Looks great, thanks!

awx/api/views/__init__.py
```diff
@@ -2228,6 +2228,9 @@ class InventorySourceUpdateView(RetrieveAPIView):
 
     def post(self, request, *args, **kwargs):
         obj = self.get_object()
+        serializer = self.get_serializer(instance=obj, data=request.data)
+        if not serializer.is_valid():
```
jbradberry: I recommend making use of the `raise_exception=True` parameter instead of constructing your own response.

https://www.django-rest-framework.org/api-guide/serializers/#raising-an-exception-on-invalid-data
https://www.cdrf.co/3.13/rest_framework.generics/CreateAPIView.html#create

jbradberry (approved):

relrod:
Refs https://github.com/ansible/awx/issues/13015

PR #13069, opened 2022-10-17
Add back in the uniqueness validation on Instance.hostname

jbradberry:
##### SUMMARY
Our new explicitly defined validators on Instance.hostname missed the implied uniqueness validator that comes from the model field.  Add it back in.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -4931,7 +4931,8 @@ class InstanceSerializer(BaseSerializer):
             'node_state': {'initial': Instance.States.INSTALLED, 'default': Instance.States.INSTALLED},
             'hostname': {
                 'validators': [
-                    MaxLengthValidator(limit_value=255),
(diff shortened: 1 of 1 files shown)


CFSNM (approved):

awx/api/serializers.py
```diff
@@ -4931,7 +4931,8 @@ class Meta:
             'node_state': {'initial': Instance.States.INSTALLED, 'default': Instance.States.INSTALLED},
             'hostname': {
                 'validators': [
-                    MaxLengthValidator(limit_value=255),
```
AlanCoding, line 4934: why change from 255 to 250 max len here?

jbradberry: @AlanCoding because if you look at the definition of the model field, you'll see it is 250, not 255.  These need to match.

AlanCoding (approved):

Issue #13017, opened 2022-10-07
Race condition that can make jobs have "failed" status due to canceling it


AlanCoding:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

In some conditions, jobs will incorrectly have the following message applied to them:

> Task was canceled due to receiving a shutdown signal.

After the user does a POST to `/api/v2/jobs/N/cancel/`.

this appears to be fallout from recent development, where we landed fixes for some other things, and this was fallout from _those_ fixes.

### AWX version

devel

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

openshift

### Modifications

no

### Ansible version

devel

### Operating system

N/A

### Web browser

Chrome

### Steps to reproduce

Right now I don't have concrete steps to reproduce, other than to use this hack:

```python
diff --git a/awx/main/models/unified_jobs.py b/awx/main/models/unified_jobs.py
index a8ac64b2cc..35f63fe9ca 100644
--- a/awx/main/models/unified_jobs.py
+++ b/awx/main/models/unified_jobs.py
@@ -1481,8 +1481,12 @@ class UnifiedJob(
                     self.status = 'canceled'
                     cancel_fields.append('status')
 
+            import time
+            time.sleep(10)
+
             self.save(update_fields=cancel_fields)
 
+
         return self.cancel_flag
 
     @property

```

However, we have reports that this bug has been seen in real life in OCP deployments. Ping @CFSNM , the reporter.

### Expected results

A user cancel should always yield a simple "canceled" status.

### Actual results

![Screenshot from 2022-10-07 15-40-41](https://user-images.githubusercontent.com/1385596/194641008-4ef0fcb0-1ab4-4ea6-942c-60acbd864252.png)

### Additional information

To describe what happens...

In the request, a message is sent to the dispatcher to SIGTERM the controller process. The process receives that, stops processing data, does a `receptorctl` cancel of the work unit, then it checks the `cancel_flag` to determine if it should use the above message, or just a "cancelled" status.

[...]

PR #13072, opened 2022-10-18
Make canceling view non-atomic to fix 500 errors with job bursts

AlanCoding:
##### SUMMARY
This is designed to address two issues, the first is https://github.com/ansible/awx/issues/13017

The second issue is filed elsewhere, so I will give steps to reproduce here:
 - Set postgres max_connections setting to 100 (can be done in the docker-compose environment, thank you @kdelee!)
 - Launch 100 jobs really fast, using threading or something like that
 - After confirming pending or running status or something like that (unclear), use a python/requests script to cancel all those jobs
   - this can be done in a linear script, don't need threading here (I think)

Doing this, the canceling will get a 500 server error at some point. Kind of obvious... since there are 100 jobs and 100 available connections. For this to work, max workers needs to be over 100, but that shouldn't be hard, I have >300 in that environment myself. We do close connections when processing job output, but the simple act of _preparing_ the job data will obviously require a connection on a 1-to-1 basis. The real problem we have here, however, is that canceling _will create a new connection_, and when we're at max connections we get an error and that's where the server error is coming from.

We don't need to create a new connection to cancel a job. Using the existing one resolves this problem, because you already have that connection slot held by the uWSGI worker. That's what this does, although it requires making the view non-atomic.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION

Both of these bugs were introduced in https://github.com/ansible/awx/pull/12769, since this change explicitly changed it to create a new connection on cancel (which also moved up the message in the timeline, causing the first linked bug).

What about even older versions? The big design change that preceded 12769 was https://github.com/ansible/awx/pull/12573, however, this _did not introduce_ the cancel action communication with the dispatcher. Before that, we just didn't care if that message was successful or not. Now, we are a little more picky about the errors the control-with-reply communication produces and that's really the main divergence from the older behavior here.

### DIFF
--- a/awx/api/generics.py
+++ b/awx/api/generics.py
@@ -13,7 +13,7 @@ from django.contrib.auth import views as auth_views
 from django.contrib.contenttypes.models import ContentType
 from django.core.cache import cache
 from django.core.exceptions import FieldDoesNotExist
-from django.db import connection
+from django.db import connection, transaction
 from django.db.models.fields.related import OneToOneRel
 from django.http import QueryDict
 from django.shortcuts import get_object_or_404
@@ -64,6 +64,7 @@ __all__ = [
     'ParentMixin',
     'SubListAttachDetachAPIView',
     'CopyAPIView',
+    'GenericCancelView',
     'BaseUsersList',
 ]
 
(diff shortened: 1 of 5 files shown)


jbradberry (approved):
I like it.

AlanCoding:
I ran the full test suite with this, and that came back looking fine. I'm feeling confident enough, I'd like to get this out in AWX. Merging.

PR #13073, opened 2022-10-18
Fix dispatcher connection deadlock w scheduler and cleanup

AlanCoding:
##### SUMMARY
This is a fairly critical fix that address a complete system deadlock. Telling the story of what happens:
 - Deploy a system with too few max_connections in postgres, like 100
 - Launch a burst of jobs which will run out the max connections
 - [consequence] Due to some questionable reasons, the dispatcher periodic scheduler loses its connection and tries to create a new one
   - this fails because postgres has already exceeded max connections
   - the periodic scheduler remains in an error loop
   - obviously, no scheduled tasks are submitted anymore
  - [consequence] The dispatcher main process stops running the `.cleanup` method because _no new tasks are submitted_
    - the cleanup method is only ran as a side-effect of submitting the `cluster_node_heartbeat` task
    - because `.cleanup` isn't getting ran, workers are not getting scaled down
    - all those workers not getting scaled down are still keeping their database connections open
  - DEADLOCK - the dispatcher won't scale down workers, which won't release their connections, which prevent the periodic scheduler from submitting tasks, which prevents the dispatcher from scaling down workers

I believe this is a fairly conservative fix that uses an already-existing functionality of `yield_timeouts`. This makes it so that the dispatcher main process main loop - `for e in conn.events(yield_timeouts=True):` will always be active. In the deadlocked situation, we never got another `e` until infinity. This gives us one every 5 seconds. It was already going through the loop every 5 seconds deeper in the `PubSub`, the only thing this does is break us back out to the `AWXConsumerPG.run` loop in those events.

Then, while we have control back, we just do a minor re-org of the cleanup trigger. I've tested this, and with the reproducer we had, this allows us to saturate connections but then still scale down workers so we do not remain permanently deadlocked.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
Added some more logging, you might see how these are the places I _thought_ were causing the problem, but they were not. It will still be good to have these logs IMO, even though I've never seen them happen. I want to have the logs so that we rule out those scenarios in any future debugging.

### DIFF
--- a/awx/main/dispatch/worker/base.py
+++ b/awx/main/dispatch/worker/base.py
@@ -114,7 +114,6 @@ class AWXConsumerBase(object):
             queue = 0
         self.pool.write(queue, body)
         self.total_messages += 1
-        self.record_statistics()
 
     @log_excess_runtime(logger)
     def record_statistics(self):
@@ -156,6 +155,16 @@ class AWXConsumerPG(AWXConsumerBase):
         # if no successful loops have ran since startup, then we should fail right away
         self.pg_is_down = True  # set so that we fail if we get database errors on startup
         self.pg_down_time = time.time() - self.pg_max_wait  # allow no grace period
+        self.last_cleanup = time.time()
+
+    def run_periodic_tasks(self):
+        self.record_statistics()  # maintains time buffer in method
+
+        if time.time() - self.last_cleanup > 60:  # same as cluster_node_heartbeat
+            # NOTE: if we run out of database connections, it is important to still run cleanup
+            # so that we scale down workers and free up connections
+            self.pool.cleanup()
+            self.last_cleanup = time.time()
 
     def run(self, *args, **kwargs):
         super(AWXConsumerPG, self).run(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


awx/main/dispatch/worker/base.py
```diff
@@ -171,8 +180,10 @@ def run(self, *args, **kwargs):
                     if init is False:
                         self.worker.on_start()
                         init = True
-                    for e in conn.events():
-                        self.process_task(json.loads(e.payload))
+                    for e in conn.events(yield_timeouts=True):
```
fosterseth, line 183: wonderful!

fosterseth (approved):

jbradberry (approved):

awx/main/dispatch/worker/base.py
```diff
@@ -156,6 +155,16 @@ def __init__(self, *args, **kwargs):
         # if no successful loops have ran since startup, then we should fail right away
         self.pg_is_down = True  # set so that we fail if we get database errors on startup
         self.pg_down_time = time.time() - self.pg_max_wait  # allow no grace period
+        self.last_cleanup = time.time()
+
+    def run_periodic_tasks(self):
+        self.record_statistics()  # maintains time buffer in method
+
+        if time.time() - self.last_cleanup > 20:  # same as cluster_node_heartbeat
```
AlanCoding: So I was wrong on this, the cluster_node_heatbeat period is 60, and also has an associated setting, `CLUSTER_NODE_HEARTBEAT_PERIOD`. So we should use that 60 second frequency. But you could also argue that we should use the setting itself... but I don't know if I like that, because this should be relatively independent.

PR #13074, opened 2022-10-18
Customize application_name for different connections in dispatcher service

AlanCoding:
##### SUMMARY
I legitimately needed this to debug another issue I was working. I could not have gotten to the root of it without identifying which was the listener connection and which was another process.

With this change:

```
(env) [alancoding@alan-red-hat awx]$ docker exec -t tools_postgres_1 psql -d awx -U awx -c "select pid,application_name, state from pg_stat_activity;"
  pid  |                  application_name                  | state  
-------+----------------------------------------------------+--------
    30 |                                                    | 
    32 |                                                    | 
 10827 | 20204-awx_1-manage.py run_wsbroadcast              | idle
 10828 | 20204-awx_1-manage.py run_wsbroadcast              | idle
 10829 | 20206-listen-awx_1-manage.py run_dispatcher        | idle
 10830 | 20206-awx_1-manage.py run_dispatcher               | idle
 10879 | 20246-worker-awx_1-manage.py run_callback_receiver | idle
 10837 | 20238-periodic-awx_1-manage.py run_dispatcher      | idle
 10880 | 20248-worker-awx_1-manage.py run_callback_receiver | idle
 10839 | 20240-worker-awx_1-manage.py run_dispatcher        | idle
 10842 | 20242-worker-awx_1-manage.py run_dispatcher        | idle
 10882 | 20247-worker-awx_1-manage.py run_callback_receiver | idle
 10881 | 20249-worker-awx_1-manage.py run_callback_receiver | idle
 10896 | psql                                               | active
    28 |                                                    | 
    27 |                                                    | 
    29 |                                                    | 
(17 rows)
```

I'm putting this up as a draft because this does need more aesthetic and formatting input.

Problems I'm trying to address here:
 - Sometimes in prod the `CLUSTER_HOST_ID` is too long, and makes the other items (like pid) impossible to read
 - All the `run_dispatcher` names are totally indistinguishable from each other
 - There's no indicator of function - sometimes a single process will hold multiple connections, and debugging this is impossible

This makes it possible to debug a lot of things we couldn't before. It just needs some help to get it to the finish line.

Would still like to have:
 - short name as opposed to `sys.argv`, like dispatcher / callback / websocket
 - fixed width columns... at least for the first few?
 - insert a placeholder function when one is not given, like "main" as opposed to "listener"
 - a shorted version of the hostname. How? No idea, help @jbradberry 

##### ISSUE TYPE
 - Breaking Change 

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
Example of the host name swamping out the others:

```
[awx@ip-10-0-40-32 ~]$ echo "select pid,application_name, state from pg_stat_activity;" | awx-manage dbshell
  pid   |                        application_name                         |        state        
--------+-----------------------------------------------------------------+---------------------
  23186 |                                                                 | 
  23188 |                                                                 | 
  59762 | ec2-3-85-163-248.compute-1.amazonaws.com-59742-/usr/bin/awx-man | idle
  61893 | ec2-3-85-163-248.compute-1.amazonaws.com-61785-/usr/bin/awx-man | idle
  59765 | ec2-3-85-163-248.compute-1.amazonaws.com-59742-/usr/bin/awx-man | idle
 138280 | ec2-3-85-163-248.compute-1.amazonaws.com-61785-/usr/bin/awx-man | idle
 138194 | ec2-3-85-163-248.compute-1.amazonaws.com-61785-/usr/bin/awx-man | idle
 138221 | ec2-3-85-163-248.compute-1.amazonaws.com-61785-/usr/bin/awx-man | idle
 138005 | ec2-3-85-163-248.compute-1.amazonaws.com-61785-/usr/bin/awx-man | idle
 138219 | ec2-3-85-163-248.compute-1.amazonaws.com-61785-/usr/bin/awx-man | idle
  70185 | ec2-3-85-163-248.compute-1.amazonaws.com-59739-/usr/bin/awx-man | idle
 138226 | ec2-3-85-163-248.compute-1.amazonaws.com-61785-/usr/bin/awx-man | idle
 138308 | ec2-3-85-163-248.compute-1.amazonaws.com-59740-uwsgi            | idle
 138312 | ec2-3-85-163-248.compute-1.amazonaws.com-59740-uwsgi            | idle
 138307 | ec2-3-85-163-248.compute-1.amazonaws.com-59740-uwsgi            | idle in transaction
 138311 | ec2-3-85-163-248.compute-1.amazonaws.com-59741-/var/lib/awx/ven | idle
 138314 | psql                                                            | active
  23184 |                                                                 | 
  23183 |                                                                 | 
  23185 |                                                                 | 
(20 rows)
```

There, we get the process id here, but only barely. In many deployments we won't get this.

Now, some reference data in docker-compose environment

```
 pid  |           application_name           | state  
------+--------------------------------------+--------
   29 |                                      | 
   31 |                                      | 
 3597 | awx_1-3104-manage.py run_wsbroadcast | idle
 3598 | awx_1-3104-manage.py run_wsbroadcast | idle
 3600 | awx_1-3108-manage.py run_dispatcher  | idle
 3647 | awx_1-3108-manage.py run_dispatcher  | idle
 3648 | awx_1-3108-manage.py run_dispatcher  | idle
 3651 | awx_1-3108-manage.py run_dispatcher  | idle
 3650 | awx_1-3108-manage.py run_dispatcher  | idle
 3652 | awx_1-3108-manage.py run_dispatcher  | idle
 3645 | awx_1-3108-manage.py run_dispatcher  | idle
 3654 | psql                                 | active
   27 |                                      | 
   26 |                                      | 
   28 |                                      | 
(15 rows)
```

This tells me... very little. The dispatcher has a parent process, an associated pg_notify listener, and multiple workers. I would really like to make it so that I can see that.

### DIFF
--- a/awx/main/dispatch/periodic.py
+++ b/awx/main/dispatch/periodic.py
@@ -10,6 +10,7 @@ from django_guid import set_guid
 from django_guid.utils import generate_guid
 
 from awx.main.dispatch.worker import TaskWorker
+from awx.main.utils.db import set_connection_name
 
 logger = logging.getLogger('awx.main.dispatch.periodic')
 
@@ -21,6 +22,9 @@ class Scheduler(Scheduler):
         def run():
             ppid = os.getppid()
             logger.warning('periodic beat started')
+
+            set_connection_name(proc_function='periodic')  # set application_name to distinguish from other dispatcher processes
+
             while True:
                 if os.getppid() != ppid:
                     # if the parent PID changes, this process has been orphaned
--- a/awx/main/dispatch/periodic.py
+++ b/awx/main/dispatch/periodic.py
@@ -23,7 +23,7 @@ class Scheduler(Scheduler):
             ppid = os.getppid()
             logger.warning('periodic beat started')
 
-            set_connection_name(proc_function='periodic')  # set application_name to distinguish from other dispatcher processes
+            set_connection_name('periodic')  # set application_name to distinguish from other dispatcher processes
 
             while True:
                 if os.getppid() != ppid:
--- a/awx/settings/development.py
+++ b/awx/settings/development.py
@@ -105,8 +105,11 @@ AWX_CALLBACK_PROFILE = True
 AWX_DISABLE_TASK_MANAGERS = False
 # ======================!!!!!!! FOR DEVELOPMENT ONLY !!!!!!!=================================
 
-if 'sqlite3' not in DATABASES['default']['ENGINE']:  # noqa
-    DATABASES['default'].setdefault('OPTIONS', dict()).setdefault('application_name', f'{CLUSTER_HOST_ID}-{os.getpid()}-{" ".join(sys.argv)}'[:63])  # noqa
+from .application_name import set_application_name
+
+set_application_name(DATABASES, CLUSTER_HOST_ID)
+
+del set_application_name
 
 # If any local_*.py files are present in awx/settings/, use them to override
 # default settings for development.  If not present, we can still run using
--- a/awx/settings/production.py
+++ b/awx/settings/production.py
@@ -102,6 +102,6 @@ except IOError:
 
 from .application_name import set_application_name
 
-set_application_name(DATABASES, CLUSTER_HOST_ID)
+set_application_name(DATABASES, CLUSTER_HOST_ID)  # NOQA
 
 del set_application_name
--- a/awx/settings/production.py
+++ b/awx/settings/production.py
@@ -100,6 +100,8 @@ except IOError:
 
 # The below runs AFTER all of the custom settings are imported.
 
-DATABASES.setdefault('default', dict()).setdefault('OPTIONS', dict()).setdefault(
-    'application_name', f'{CLUSTER_HOST_ID}-{os.getpid()}-{" ".join(sys.argv)}'[:63]  # NOQA
-)  # noqa
+from .application_name import set_application_name
+
+set_application_name(DATABASES, CLUSTER_HOST_ID)
+
+del set_application_name
--- a/awx/main/dispatch/__init__.py
+++ b/awx/main/dispatch/__init__.py
@@ -85,7 +85,7 @@ def pg_bus_conn(new_connection=False):
     '''
 
     if new_connection:
-        conf = settings.DATABASES['default']
+        conf = settings.DATABASES['default'].copy()
         conf['OPTIONS'] = conf.get('OPTIONS', {}).copy()
         # Modify the application name to distinguish from other connections the process might use
         conf['OPTIONS']['application_name'] = get_application_name(settings.CLUSTER_HOST_ID, function='listener')
--- a/awx/main/dispatch/__init__.py
+++ b/awx/main/dispatch/__init__.py
@@ -4,6 +4,8 @@ import select
 
 from contextlib import contextmanager
 
+from awx.settings.application_name import get_application_name
+
 from django.conf import settings
 from django.db import connection as pg_connection
 
@@ -84,9 +86,10 @@ def pg_bus_conn(new_connection=False):
 
     if new_connection:
         conf = settings.DATABASES['default']
-        conn = psycopg2.connect(
-            dbname=conf['NAME'], host=conf['HOST'], user=conf['USER'], password=conf['PASSWORD'], port=conf['PORT'], **conf.get("OPTIONS", {})
-        )
+        conf['OPTIONS'] = conf.get('OPTIONS', {}).copy()
+        # Modify the application name to distinguish from other connections the process might use
+        conf['OPTIONS']['application_name'] = get_application_name(settings.CLUSTER_HOST_ID, function='listener')
+        conn = psycopg2.connect(dbname=conf['NAME'], host=conf['HOST'], user=conf['USER'], password=conf['PASSWORD'], port=conf['PORT'], **conf['OPTIONS'])
         # Django connection.cursor().connection doesn't have autocommit=True on by default
         conn.set_session(autocommit=True)
     else:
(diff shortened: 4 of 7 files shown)


kdelee:
we could use instance id instead of cluster host id? this would just be the pk of the host, though perhaps that would require a query :/

AlanCoding:
yeah, we have to apply the application_name _really_ early on in the python imports. This setting is used _by Django to create a connection_ and you have to have a connection to query the database.

jbradberry:
@AlanCoding what do you mean, a short form of the hostname?

AlanCoding:
@jbradberry see the example above, like

```
  59762 | ec2-3-85-163-248.compute-1.amazonaws.com-59742-/usr/bin/awx-man | idle
```

Both, due to the long hostname, and due to the `/usr/bin/awx-manage` parts, we fully lose what _service_ is being ran. We don't know if that was from the dispatcher or from the websockets service. This PR is going even further, trying to distinguish between the multiple connections that processes within the dispatcher service open. We have a set character limit, so there's no good way to do this.

Not attempted here - it should be workable to create a short name for each service we run. Like, have a value somewhere that is just "dispatcher" or "web". We would create these ourselves, and we would know the max length. The host names, on the other hand, would be fairly challenging to shorten.

jbradberry:
@AlanCoding I'm afraid I don't have much of an opinion on this.

AlanCoding:
I'm updating this so that the look is as follows:

```
bash-5.1$ echo "select pid,application_name, state from pg_stat_activity;" | awx-manage dbshell
 pid |              application_name               | state  
-----+---------------------------------------------+--------
  30 |                                             | 
  32 |                                             | 
 600 | awx-16827-heartbeet_listener-awx_1          | idle
 601 | awx-16824-rsyslog_configurer_listener-awx_1 | idle
 602 | awx-16824-rsyslog_configurer_listener-awx_1 | idle
 603 | awx-16825-wsrelay-awx_1                     | idle
 604 | awx-16812-dispatcher_listener-awx_1         | idle
 757 | awx-16950-dispatcher_worker-awx_1           | idle
 606 | awx-16825-wsrelay-awx_1                     | idle
  49 | awx-295-cache_clear-awx_1                   | idle
 609 | awx-16827-heartbeet_listener-awx_1          | idle
 620 | awx-16816-callback_receiver-awx_1           | idle
 727 | awx-16924-dispatcher_worker-awx_1           | idle
 756 | awx-16860-dispatcher_periodic-awx_1         | idle
 758 | awx-16918-dispatcher_worker-awx_1           | idle
 730 | awx-16952-dispatcher_worker-awx_1           | idle
 763 | awx-16850-callback_receiver_worker-awx_1    | idle
 764 | awx-16852-callback_receiver_worker-awx_1    | idle
 765 | awx-16851-callback_receiver_worker-awx_1    | idle
 766 | awx-16849-callback_receiver_worker-awx_1    | idle
 767 | psql                                        | active
  28 |                                             | 
  27 |                                             | 
  29 |                                             | 
(24 rows)
```

This has cut out the `awx-manage` text, but also added an "awx-" prefix in general. Nothing here is of unbounded length except for the node name.

awx/main/dispatch/periodic.py
```diff
@@ -21,6 +22,9 @@ def run_continuously(self):
         def run():
             ppid = os.getppid()
             logger.warning('periodic beat started')
+
+            set_connection_name('periodic')  # set application_name to distinguish from other dispatcher processes
```
AlanCoding, line 26: I should note - I don't ever see this. I believe that's because it _quickly_ closes the connection after sending off periodic tasks. Nonetheless, I believe it's working correctly, the service just cleans up after itself. This is a good thing.

awx/settings/production.py
```diff
@@ -100,6 +100,8 @@
...
 # The below runs AFTER all of the custom settings are imported.
 
-DATABASES.setdefault('default', dict()).setdefault('OPTIONS', dict()).setdefault(
-    'application_name', f'{CLUSTER_HOST_ID}-{os.getpid()}-{" ".join(sys.argv)}'[:63]  # NOQA
-)  # noqa
+from .application_name import set_application_name
+
+set_application_name(DATABASES, CLUSTER_HOST_ID)  # NOQA
+
+del set_application_name
```
fosterseth, line 107: can you explain the purpose of this delete?

awx/settings/development.py
```diff
@@ -105,8 +105,11 @@
 AWX_DISABLE_TASK_MANAGERS = False
 # ======================!!!!!!! FOR DEVELOPMENT ONLY !!!!!!!=================================
 
-if 'sqlite3' not in DATABASES['default']['ENGINE']:  # noqa
-    DATABASES['default'].setdefault('OPTIONS', dict()).setdefault('application_name', f'{CLUSTER_HOST_ID}-{os.getpid()}-{" ".join(sys.argv)}'[:63])  # noqa
+from .application_name import set_application_name
+
+set_application_name(DATABASES, CLUSTER_HOST_ID)
+
+del set_application_name
```
jessicamack, line 112: What's being deleted here?

AlanCoding: This module is what will result when you import Django settings, which you get doing

```python
```

The settings are pretty special compared to the rest of the code, and comes _way earlier_ in import order than most stuff. So you really can't import from other places in your app, because you will very quickly get circular imports. In this case, however, I have a static method that I want to share around. You can define methods / classes / logic in your settings, but you should try to avoid this as much as possible. My solution is that I introduce a new module inside of the settings folder that contains static methods. The ordinary settings use one of those static methods, but I'm not trying to vendor that static method inside of the settings.

In other words, I don't want someone to be able to do `settings.set_application_name()`. It's not that it would be invalid, but I'm doing the imports a different way, because my design preference is to not define methods inside of settings. Ideally, I want settings to contain _static_ content.

So deleting `set_application_name` removes that method from the local variables in settings, but the result of running it remains, in that it has altered the `DATABASES` local variable. I just want to be very careful about what I _add to_ settings, so I do this thing, but I delete the python resources I used in order to do it.

awx/settings/production.py
AlanCoding: Left a reply https://github.com/ansible/awx/pull/13074/files#r1165832158

fosterseth (approved):

Issue #13059, opened 2022-10-13
Gap in job output


keithjgrant:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

Gap in job output while job is running. Appears between events loaded when the page opens and those that are received via websockets after that point.

### AWX version

devel

### Select the relevant components

- [X] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

docker development environment

### Modifications

no

### Steps to reproduce

Launch a job. The bug generally occurs when the network requests on the job page are slow/delayed.

### Expected results

All job events display

### Actual results

Gap in job events and a gray loading bar appears in the gap, but never resolves loading.

PR #13075, opened 2022-10-18
Fix gap between API-loaded job events and WS-streamed events

keithjgrant:
##### SUMMARY
Fixes occasional bug where a gray loading bar appears between job events loaded via API and those streamed over websockets. When the first events stream over websockets, the page will now re-load the prior events from the API, updating them and the `remoteRowCount` to ensure accuracy.

I also fixed an off-by-one error that sometimes made the last API-fetched event not render even if it is already available in memory.

Addresses #13059

This bug generally only presents itself when the initial load of the job or its events via API takes too long to resolve, and already-completed events are missing from the fetched results. I was able to reproduce it with ~50% frequency by manually delaying the response when fetching the job object.

##### ISSUE TYPE

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui/src/screens/Job/JobOutput/JobOutput.js
+++ b/awx/ui/src/screens/Job/JobOutput/JobOutput.js
@@ -251,6 +251,9 @@ function JobOutput({ job, eventRelatedSearchableKeys, eventSearchableKeys }) {
         });
         const updated = oldWsEvents.concat(newEvents);
         jobSocketCounter.current = updated.length;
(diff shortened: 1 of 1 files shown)


AlexSCorey (approved):
This looks good to me.

akus062381 (commented):
The follow button is acting up. It won't stay clicked, but keeps turning itself off.

akus062381 (approved):

PR #13083, opened 2022-10-20
new machinery for webhook notifications to allow processing of redirects

sarabrajsingh:
##### SUMMARY

this change addresses a bug in webhook notifications, where if the location URL is a reference to an endpoint with a redirect, that the POST request would get turned into a GET request and unable to process the redirect.

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.6.1.dev237+gad39b5c88b.d20221020
```

### DIFF
--- a/awx/main/notifications/webhook_backend.py
+++ b/awx/main/notifications/webhook_backend.py
@@ -5,9 +5,6 @@ import json
 import logging
 import requests
 
-from django.utils.encoding import smart_str
-from django.utils.translation import gettext_lazy as _
-
 from awx.main.notifications.base import AWXBaseEmailBackend
 from awx.main.utils import get_awx_http_client_headers
 from awx.main.notifications.custom_notification_base import CustomNotificationBase
@@ -16,6 +13,8 @@ logger = logging.getLogger('awx.main.notifications.webhook_backend')
 
 
 class WebhookBackend(AWXBaseEmailBackend, CustomNotificationBase):
+
+    MAX_RETRIES = 5
 
     init_parameters = {
         "url": {"label": "Target URL", "type": "string"},
@@ -64,20 +63,67 @@ class WebhookBackend(AWXBaseEmailBackend, CustomNotificationBase):
         if self.http_method.lower() not in ['put', 'post']:
             raise ValueError("HTTP method must be either 'POST' or 'PUT'.")
         chosen_method = getattr(requests, self.http_method.lower(), None)
+
         for m in messages:
+
             auth = None
             if self.username or self.password:
                 auth = (self.username, self.password)
-            r = chosen_method(
-                "{}".format(m.recipients()[0]),
-                auth=auth,
-                data=json.dumps(m.body, ensure_ascii=False).encode('utf-8'),
-                headers=dict(list(get_awx_http_client_headers().items()) + list((self.headers or {}).items())),
-                verify=(not self.disable_ssl_verification),
-            )
-            if r.status_code >= 400:
-                logger.error(smart_str(_("Error sending notification webhook: {}").format(r.status_code)))
+
+            # the constructor for EmailMessage - https://docs.djangoproject.com/en/4.1/_modules/django/core/mail/message will turn an empty dictionary to an empty string
+            # sometimes an empty dict is intentional and we added this conditional to enforce that
+            if not m.body:
+                m.body = {}
+
+            url = str(m.recipients()[0])
+            data = json.dumps(m.body, ensure_ascii=False).encode('utf-8')
+            headers = {**(get_awx_http_client_headers()), **(self.headers or {})}
+
+            err = None
+
+            for retries in range(self.MAX_RETRIES):
+
+                # Sometimes we hit redirect URLs. We must account for this. We still extract the redirect URL from the response headers and try again. Max retires == 5
+                resp = chosen_method(
+                    url=url,
+                    auth=auth,
+                    data=data,
+                    headers=headers,
+                    verify=(not self.disable_ssl_verification),
+                    allow_redirects=False,  # override default behaviour for redirects
+                )
+
+                # either success or error reached if this conditional fires
+                if resp.status_code not in [301, 307]:
+                    break
+
+                # we've hit a redirect. extract the redirect URL out of the first response header and try again
+                logger.warning(
+                    f"Received a {resp.status_code} from {url}, trying to reach redirect url {resp.headers.get('Location', None)}; attempt #{retries+1}"
+                )
+
+                # take the first redirect URL in the response header and try that
+                url = resp.headers.get("Location", None)
+
+                if url is None:
+                    err = f"Webhook notification received redirect to a blank URL from {url}. Response headers={resp.headers}"
+                    break
+            else:
+                # no break condition in the loop encountered; therefore we have hit the maximum number of retries
+                err = f"Webhook notification max number of retries [{self.MAX_RETRIES}] exceeded. Failed to send webhook notification to {url}"
+
+            if resp.status_code >= 400:
+                err = f"Error sending webhook notification: {resp.status_code}"
+
+            # log error message
+            if err:
+                logger.error(err)
                 if not self.fail_silently:
-                    raise Exception(smart_str(_("Error sending notification webhook: {}").format(r.status_code)))
-            sent_messages += 1
+                    raise Exception(err)
+
+            # no errors were encountered therefore we successfully sent off the notification webhook
+            if resp.status_code in range(200, 299):
+                logger.debug(f"Notification webhook successfully sent to {url}. Received {resp.status_code}")
+                sent_messages += 1
+
         return sent_messages
--- a/awx/main/tests/functional/test_notifications.py
+++ b/awx/main/tests/functional/test_notifications.py
@@ -75,6 +75,7 @@ def test_encrypted_subfields(get, post, user, organization):
     url = reverse('api:notification_template_detail', kwargs={'pk': response.data['id']})
     response = get(url, u)
     assert response.data['notification_configuration']['account_token'] == "$encrypted$"
+
     with mock.patch.object(notification_template_actual.notification_class, "send_messages", assert_send):
         notification_template_actual.send("Test", {'body': "Test"})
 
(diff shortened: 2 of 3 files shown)


awx/main/notifications/webhook_backend.py
```diff
@@ -17,6 +17,8 @@
 
 class WebhookBackend(AWXBaseEmailBackend, CustomNotificationBase):
 
+    MAX_RETIRES = 5
```
jbradberry: retires -> retries

awx/main/notifications/webhook_backend.py
```diff
@@ -61,23 +63,60 @@ def format_body(self, body):
 
     def send_messages(self, messages):
         sent_messages = 0
+        retries = 0
         if self.http_method.lower() not in ['put', 'post']:
             raise ValueError("HTTP method must be either 'POST' or 'PUT'.")
         chosen_method = getattr(requests, self.http_method.lower(), None)
+        resp = None
+        '''
+        Sometimes we hit redirect URLs. We must account for this. We still extract the redirect URL from the response headers and try again. Max retires == 5
```
jbradberry: This is more of an explanatory comment than a docstring.

awx/main/notifications/webhook_backend.py
```diff
@@ -61,23 +63,60 @@ def format_body(self, body):
...
+                    verify=(not self.disable_ssl_verification),
+                    allow_redirects=False,  # override default behaviour for redirects
+                )
+
+                # we've hit a redirect. extract the redirect URL out of the first response header and try again
+                if resp.status_code in [301, 307]:
+                    # take the first redirect URL in the response header and try that
+                    url = resp.headers.get("Location", None)
+                    if url is None:
+                        raise Exception(smart_str(_("Recieved redirect to a blank URL. Response headers={}").format(resp_inner.history[0].headers)))
```
jbradberry: let's also do an error log for these

awx/main/notifications/webhook_backend.py
```diff
@@ -61,23 +63,60 @@ def format_body(self, body):
...
+                    continue
+
+                if resp.status_code in range(200, 299):
+                    sent_messages += 1
+                    break
+
+                if resp.status_code >= 400:
+                    logger.error(smart_str(_("Error sending notification webhook: {}").format(resp.status_code)))
+                    if not self.fail_silently:
+                        raise Exception(smart_str(_("Error sending notification webhook: {}").format(resp.status_code)))
```
jbradberry: and here

awx/main/notifications/webhook_backend.py
```diff
@@ -61,23 +63,60 @@ def format_body(self, body):
...
-            )
-            if r.status_code >= 400:
-                logger.error(smart_str(_("Error sending notification webhook: {}").format(r.status_code)))
-                if not self.fail_silently:
-                    raise Exception(smart_str(_("Error sending notification webhook: {}").format(r.status_code)))
-            sent_messages += 1
+
+            # convert emtpy string to empty dict for POST/PUT
+            if not m.body:
+                m.body = dict()
```
sarabrajsingh: @jbradberry just the preamble here for this line of code:

in the constructor for EmailMessage (https://docs.djangoproject.com/en/4.1/_modules/django/core/mail/message)

the django code is doing something like

```
self.body = body or ''
```

which is strange in the python-world i guess that an empty string `''` will be preferred over an empty `dict()` 

```python
Python 3.9.13 (main, Jul 25 2022, 00:00:00) 
[GCC 11.3.1 20220421 (Red Hat 11.3.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> body = {}
>>> body or ''
''
>>> 
```

this is why the email body was always coming through as `''` and not `{}` like we expected.

jbradberry: Though tbh I'd like to see us start using more modern-looking Python -- let's use `{}` instead of `dict()`.

jbradberry: And your description in this Github comment was terrific.  Can you capture that a bit better there in line 80?

awx/main/notifications/webhook_backend.py
```diff
@@ -61,23 +63,60 @@ def format_body(self, body):
...
+
+                # we've hit a redirect. extract the redirect URL out of the first response header and try again
+                if resp.status_code in [301, 307]:
+                    # take the first redirect URL in the response header and try that
+                    url = resp.headers.get("Location", None)
+                    if url is None:
+                        raise Exception(smart_str(_("Recieved redirect to a blank URL. Response headers={}").format(resp_inner.history[0].headers)))
+                    continue
+
+                if resp.status_code in range(200, 299):
```
jbradberry: I think the loop logic could be a bit more compact:

1. This block goes away entirely.
2. The redirect status code conditional above gets an `else:` clause with just a `break` in it.
3. The error status code conditional below goes outside of the loop.
4. A new 'skinny' redirect code conditional also goes outside of the loop, to catch the case where we exceeded the retries.

Right?  Both a successful request and an unsuccessful request are terminating conditions, so the loop is no longer necessary at that point.  And putting the error code check and a final redirect code check outside of the loop removes the need to track the message count.

Alternatively, you could use the admittedly difficult to understand `else:` clause of `for` loops to avoid the outer check for a redirect code:

```python
for attempt in range(self.RETRIES):
    # ...
else:
    # The `break` condition was never hit inside the loop, so we exceeded the retries.
    logger.error(...)
    raise Exception(...) 
```

awx/main/notifications/webhook_backend.py
sarabrajsingh: lines 108-110 are the happy path forward. do we need a `logger.error()` for that?

awx/main/notifications/webhook_backend.py
```diff
@@ -61,23 +63,60 @@ def format_body(self, body):
...
-                logger.error(smart_str(_("Error sending notification webhook: {}").format(r.status_code)))
-                if not self.fail_silently:
-                    raise Exception(smart_str(_("Error sending notification webhook: {}").format(r.status_code)))
-            sent_messages += 1
+
+            # convert emtpy string to empty dict for POST/PUT
+            if not m.body:
+                m.body = dict()
+
+            url = "{}".format(m.recipients()[0])
```
jbradberry: Python modernization: let's just do `str(m.recipients()[0])`.  Even the `str(...)` part is probably unnecessary, tbh.

awx/main/notifications/webhook_backend.py
```diff
@@ -61,23 +63,60 @@ def format_body(self, body):
...
+            data = json.dumps(m.body, ensure_ascii=False).encode('utf-8')
+
+            headers = dict(list(get_awx_http_client_headers().items()) + list((self.headers or {}).items()))
```
jbradberry: Python modernization:

```python
dict(**get_awx_http_client_headers(), **(self.headers or {}))
```

awx/main/notifications/webhook_backend.py
jbradberry: Suggestions:

1. I believe that `smart_str(...)` is redundant here and in the other places in this code.  It's a Django utility for forcing conversion from bytes to unicode strings, but everything we are dealing with here should _already_ be unicode before `smart_str` gets a chance to operate on it (stuff from the header may or may not be bytes, but we've already formatted them into a unicode `str`).
2. Python modernization: Let's use an f-string instead of `.format()`
3. Recieved -> Received

awx/main/notifications/webhook_backend.py
sarabrajsingh: i think it's safe to leave the cast

awx/main/notifications/webhook_backend.py
jbradberry: I suppose it depends on whether some caller upstream is catching the exception and doing some logging or whatever with it.  :man_shrugging:

sarabrajsingh: ok i can add a `logger.info()` here

awx/main/notifications/webhook_backend.py
```diff
@@ -64,20 +63,68 @@ def send_messages(self, messages):
...
+            if err:
+                logger.error(err)
                 if not self.fail_silently:
-                    raise Exception(smart_str(_("Error sending notification webhook: {}").format(r.status_code)))
-            sent_messages += 1
+                    raise Exception(err)
+
+            # no errors were encountered therefore we successfully sent off the notification webhook
+            if resp.status_code in range(200, 299):
+                logger.info(f"Notification webhook successfully sent to {url}. Recieved {resp.status_code}")
```
jbradberry: To be honest we should ditch this log call.  It feels like a debug at best.  (Also it still has 'Recieved' misspelled.)

awx/main/notifications/webhook_backend.py
```diff
@@ -64,20 +63,68 @@ def send_messages(self, messages):
...
+                    allow_redirects=False,  # override default behaviour for redirects
+                )
+
+                # either success or error reached if this conditional fires
+                if resp.status_code not in [301, 307]:
+                    break
+
+                # Sometimes we hit redirect URLs. We must account for this. We still extract the redirect URL from the response headers and try again. Max retires == 5
+                # we've hit a redirect. extract the redirect URL out of the first response header and try again
+                logger.error(
```
jbradberry: I think this one should downgrade to info or warning.  @AlanCoding ?

awx/main/notifications/webhook_backend.py
```diff
@@ -64,20 +63,68 @@ def send_messages(self, messages):
...
+            # the constructor for EmailMessage - https://docs.djangoproject.com/en/4.1/_modules/django/core/mail/message will turn an empty dictionary to an empty string
+            # sometimes an empty dict is intentional and we added this conditional to enforce that
+            if not m.body:
+                m.body = {}
+
+            url = str(m.recipients()[0])
+            data = json.dumps(m.body, ensure_ascii=False).encode('utf-8')
+            headers = {**(get_awx_http_client_headers()), **(self.headers or {})}
+
+            resp = None
```
jbradberry: This one is unneeded, we always set a new value at the start of every iteration of the loop below.

jbradberry: (I might have said something contrary previously, but seeing the structure of the code now I think it is unnecessary.)

awx/main/notifications/webhook_backend.py
jbradberry: I lean towards 'warning'.

awx/main/notifications/webhook_backend.py
```diff
@@ -64,20 +63,68 @@ def send_messages(self, messages):
...
+                    headers=headers,
+                    verify=(not self.disable_ssl_verification),
+                    allow_redirects=False,  # override default behaviour for redirects
+                )
+
+                # either success or error reached if this conditional fires
+                if resp.status_code not in [301, 307]:
+                    break
+
+                # Sometimes we hit redirect URLs. We must account for this. We still extract the redirect URL from the response headers and try again. Max retires == 5
```
jbradberry: Let's move this comment up right before the `chosen_method(...)` call.

awx/main/notifications/webhook_backend.py
sarabrajsingh: changed this to debug :) and fixed spelling mistake

awx/main/notifications/webhook_backend.py
sarabrajsingh: ok removed.

jbradberry (approved):

AlanCoding:
Quick looking into the subject

https://gist.github.com/cbare/4333273

They confirm what you're saying there, and from the requests library release notes, it looks like there has been some history around this, so it shouldn't be surprising to have this kind of workaround. In that gist they suggest `strict_mode` but the docs seem really unclear on that to me, and what you've done looks really well developed.

AlanCoding (approved):

jay-steurer:
Verified by QE. Merging it.

Issue #13084, opened 2022-10-20
awxkit api: import_asserts cannot import schedules


ArtsiomMusin:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

import_asserts from ApiV2 class cannot import schedules asserts as unified_job_template is not found:
```
/api/v2/schedules/ "myschedule": Bad Request (400) received - {'unified_job_template': ['This field may not be null.']}.

```

Seen this issue for latest versions awxkit 21.*. Maybe exists in prev releases as well

### AWX version

Ansible Tower/AAP

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [x] CLI
- [ ] Other

### Installation method

openshift

### Modifications

no

### Steps to reproduce

1. Install awxkit 21.7.0
1. Export a schedule
2. Import the schedule

### Expected results

A schedule is imported

### Actual results

Failed with 
/api/v2/schedules/ "myschedule": Bad Request (400) received - {'unified_job_template': ['This field may not be null.']}.

jay-steurer:
Closing this since issue was fixed by merged PR.

PR #13090, opened 2022-10-21
Fix expor/import for schedules

ArtsiomMusin:
Signed-off-by: Artsiom Musin <artyom.musin@gmail.com>

##### SUMMARY
related #13084 

Since I'm new to this code, please don't judge much :smile: 

Tried to fix the issue related to this error for importing schedules:
> /api/v2/schedules/ "myschedule": Bad Request (400) received - {'unified_job_template': ['This field may not be null.']}.

##### ISSUE TYPE

##### COMPONENT NAME
 - Collection

##### AWX VERSION
Ansible Tower 3.8.3
AAP 4.2.1

##### ADDITIONAL INFORMATION
 
**First issue**
Basically the initial error was related to the incorrect order of dependencies in
https://github.com/ansible/awx/blob/devel/awxkit/awxkit/api/pages/api.py#L386

`schedules` resource came before `inventory_sources`,`projects`,`job_templates` and that's why unified_job_template could not be found in the list of natural keys.

Added optional dependencies in `awxkit/awxkit/api/pages/schedules.py` to make the order correct

**Second issue**
After the first issue was fixed, I saw this error:
> /api/v2/schedules/ "myschedule": Bad Request (400) received - {'inventory': ['Field is not allowed on launch.']}.

This was related to the fact that inventory sources and job templates are setting inventory id. However, it's not possible to set that via API. I tried simple POST and PATCH and provided inventory id and always saw
image

I'm not sure if the inventory field in schedules is obsolete or something. Anyway I wasn't able to figure out where it's used and how to set it via POST/PATCH. Due to this strange behaviour, import failed as well.

To resolve that, I change the schedule export in `awxkit/awxkit/api/pages/api.py` to skip inventory id when searching in nattural keu. The same way as was done for `WorkflowApprovalTemplate`. After that import starts working fine.

### DIFF
--- a/awxkit/awxkit/api/pages/api.py
+++ b/awxkit/awxkit/api/pages/api.py
@@ -123,6 +123,9 @@ class ApiV2(base.Base):
             if rel_endpoint.__item_class__.__name__ == 'WorkflowApprovalTemplate':
                 continue
 
+            if rel_endpoint.__item_class__.__name__ == 'Inventory' and _page.__item_class__.__name__ == 'Schedule':
+                continue
+
             rel_natural_key = rel_endpoint.get_natural_key(self._cache)
             if rel_natural_key is None:
                 log.error("Unable to construct a natural key for foreign key %r of object %s.", key, _page.endpoint)
--- a/awxkit/awxkit/api/pages/schedules.py
+++ b/awxkit/awxkit/api/pages/schedules.py
@@ -1,6 +1,6 @@
 from contextlib import suppress
 
-from awxkit.api.pages import SystemJobTemplate
+from awxkit.api.pages import SystemJobTemplate, Projects, JobTemplates, InventorySources
 from awxkit.api.mixins import HasCreate
 from awxkit.api.resources import resources
 from awxkit.config import config
@@ -12,6 +12,7 @@ from . import base
 
 class Schedule(HasCreate, base.Base):
     dependencies = [SystemJobTemplate]
+    optional_dependencies = [Projects, JobTemplates, InventorySources]
     NATURAL_KEY = ('unified_job_template', 'name')
 
     def silent_delete(self):


TheRealHaoLiu:
thanks for your contribution @ArtsiomMusin !

i see that you have been pretty active in the AWX project lately that is awesome!

have you heard about our matrix chat server? lots of devs and contributor are hanging out in there! come and join us! https://matrix.to/#/#awx:ansible.com

AlanCoding:
I think this looks correct, just need to get some testing around it.

jbradberry (changes requested):
@ArtsiomMusin can you please provide a minimal reproducer import file?

My initial gut reaction is that there is something else going on here, and that messing with the order of operations in the way you are doing is just masking the problem.

ArtsiomMusin:
@jbradberry below is the example import file which shows this error:
> /api/v2/schedules/ "abc1_proj": Bad Request (400) received - {'unified_job_template': ['This field may not be null.']}.

```json
{
  "schedules": [
    {
      "rrule": "DTSTART;TZID=America/New_York:20221020T231500 RRULE:INTERVAL=1;COUNT=1;FREQ=MINUTELY",
      "name": "abc1_proj",
      "description": "",
      "extra_data": {},
      "inventory": null,
      "scm_branch": null,
      "job_type": null,
      "job_tags": null,
      "skip_tags": null,
      "limit": null,
      "diff_mode": null,
      "verbosity": null,
      "enabled": true,
      "unified_job_template": {
        "organization": {
          "name": "abc1",
          "type": "organization"
        },
        "name": "abc1_project_netcool",
        "type": "project"
      },
      "related": {
        "credentials": []
      },
      "natural_key": {
        "unified_job_template": {
          "organization": {
            "name": "abc1",
            "type": "organization"
          },
          "name": "abc1_project_netcool",
          "type": "project"
        },
        "name": "abc1_proj",
        "type": "schedule"
      }
    }
  ]
}

```

if you think it may hide some other different issue, let me know and I will try to fix the real root cause of that.

ArtsiomMusin:
one more note. Schedules are not imported as well when the import file has other asserts like below:
```text
{
  "organizations": [...],
  "credentials": [...],
  "credential_types": [...],
  "projects": [...],
  "schedules": [...],
  "inventory": [...],
  "job_templates": [...],
  "applications": [...]
}
```

jbradberry:
> one more note. Schedules are not imported as well when the import file has other asserts like below:
> 

On a fresh development environment, I get
```
(awx-cli) [jrb@fedora awx]$ awx export | jq "keys"
[
  "applications",
  "credential_types",
  "credentials",
  "execution_environments",
  "inventory",
  "inventory_sources",
  "job_templates",
  "notification_templates",
  "organizations",
  "projects",
  "schedules",
  "teams",
  "users",
  "workflow_job_templates"
]
(awx-cli) [jrb@fedora awx]$ awx export | jq ".schedules | length"
4
```

jbradberry:
> @jbradberry below is the example import file which shows this error:
> 
> > /api/v2/schedules/ "abc1_proj": Bad Request (400) received - {'unified_job_template': ['This field may not be null.']}.
> 
> if you think it may hide some other different issue, let me know and I will try to fix the real root cause of that.

Ok, so this is your own schedule that is attached to some ~~job template~~ project.  Did you import the relevant ~~job template~~ project, or does it otherwise exist on the system you are importing to?

jbradberry:
Testing it out.  I've created a new project and a schedule attached to it:

```
(awx-cli) [jrb@fedora awx]$ awx export --projects 8 --schedules 7 > /tmp/test.json
(awx-cli) [jrb@fedora awx]$ jq 'map_values(length)' /tmp/test.json 
{
  "projects": 1,
  "schedules": 1
}
```

Attempting to import does recreate the problem you describe.

jbradberry:
Ok, so here's the deal.  When we added support for import/export of schedules, we only put the system jobs in its dependency list.  Really it needs all of the unified job template subtypes.  I also had a problem with the project update timeout causing it not to be put into the page cache, so the schedule failed to pick it up on creation.  I've bumped the timeout and made it pass through the error catching logic.

Closing this PR in favor of #13132.

PR #13094, opened 2022-10-23
Draft: add config export

sean-m-sullivan:
##### SUMMARY
To improve on the export of objects from the controller, either as a supplement to the awx_export/cli or replacement.
 Idea is to add endpoints that are not there, resolve associations, and to allow for easier exporting of the endpoints to get people into configuration as code.

Included working playbook, with working filter, and a to do list of endpoints done and those with work needed. output file included for review during Draft. 

Mainly wanting some validation that this is a route to keep perusing and flushing out, or if you think this is a Horrible idea and that we should never include it in the collection, and/or feedback.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - Collection

### DIFF
--- a/awx_collection/all.yaml
+++ b/awx_collection/all.yaml
@@ -1,0 +1,1502 @@
+changed: false
+controller_credential_input_sources:
+  - created: '2022-10-22T17:42:56.159989Z'
+    description: Fill the gitlab credential from CyberArk
+    id: 1
+    input_field_name: password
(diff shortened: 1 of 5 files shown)


jbradberry (changes requested):
I very strongly recommend against taking the approach of re-implementing the import/export logic outside of awxkit.

Issue #12288, opened 2022-05-26
RFE: Max concurrent forks/jobs on container group


kdelee:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Right now ContainerGroups don't have any concept of capacity. They therefore are sinks for jobs if any container group is in the preferred instance group list when jobs are going through the task manager. This can cause problems of sending too many jobs to a container group. Even if you DO assign some kind of limit to pods in a namespace and kubernetes sends back a 4xx response, we just re-submit the job to pending and we'll try and schedule it there again the next time around.

A KISS solution is to do something like the jenkins kubernetes plugin does, which is allow users to set a cap on how many concurrent jobs to schedule on a container group.

We could add "max_concurrent_jobs" to InstanceGroups that are ContainerGroups (not sure on how to do this conditionally, but we must be able to since we don't have pod_spec_override on all the InstanceGroups)

At startup of the task manager, we can look at the `jobs_running`  for container groups https://github.com/ansible/awx/blob/0c4ddc7f6fe2e52c7a8b44ec7e576929ea861a66/awx/main/models/ha.py#L342-L344

Then, we probably need to extend the models in to store this value for ContainerGroups and keep a running tab of how many left we have before we hit "max_concurrent_jobs" https://github.com/ansible/awx/blob/devel/awx/main/scheduler/task_manager_models.py#L25-L26 to account for ContainerGroups in particular needing to "consume_capacity" in a slightly different way -- for CG just count jobs, not task_impact

adpavlov:
Limiting parallel jobs run levels out automation and ansible ideology.
I would state that It's real and dramatic degradation in performance comparing to AWX 15 and docker architecture.

Following my initially created issue #12088, I proceeded testing and experiencing very slow behavior. In my case few parallel jobs are running quite fast and fine.
However, if I need to launch more than 10 parallel jobs it will lead to 10 Pods/containers creation, which takes more than 2 minutes to create Pods and containers. I have a workflow with even 17 parallel templates and it's time increasing up to 7 minutes just for playbook to understand that no hosts matched:)

At the same time awx_task container memory usage goes to the sky up to 8Gb, server is lagging.
Sometimes jobs are going in Error state with different reasons like:

- Error creating pod: rpc error: code = Unknown desc = database is locked
- No output found for this job.
[...]

PR #13095, opened 2022-10-24
[wip] Add max concurrent jobs per container group

kdelee:
See https://github.com/ansible/awx/issues/12288
Tests are passing except one that is encountering more queries than expected.

TODO:
  * add tests
  * investigate additional queries
  * ensure validation is working in serializer so we don't set this on regular instance groups
  * add ui for setting the Max Concurrent Jobs

QUESTIONS:
  * we could probably have this on the model, but hide it in the api/ui for everything but container groups. Consider if we want to do that.

##### SUMMARY

Specifically for container groups, there are 2 pain points. 

* First, if there is control capacity, there is no way to throttle how many jobs get launched into a container group. 
* The second pain point is that it is useless to have 2 or more container groups assigned to a resource, because each container group has no limit on how many jobs we will launch into it, as long as we have control capacity.
 
By introducing a max number of concurrent jobs for a container group, we mitigate both pain points. It will now be possible to put in information so that your first 50 jobs will go to one container group, and the following 50 to a different container group, or simply to wait until there were less than $max_concurrent_jobs running on a container group to launch the next job.

related #12288
##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 - UI


fosterseth:
my gut instinct on this

1. worried about introducing yet another piece of logic to the task manager system, especially one that overlaps in functionality with another feature we have (capacity)
2. is there any way to leverage our notion of capacity for container groups?
   - maybe make capacity user-defined for CG, and just default to something reasonable
3.  if we don't do 2, what if we just round-robin the task assignment for an IG that consists of > 1 container group? That way both get filled up evenly

I might be missing some vital info here so forgive me if my suggestions make no sense :)

kdelee:
The overlap concern is valid. One thing we could do is like with policy_instance_percentage not being exposed on container groups etc, not expose this on regular instance groups, and not consider this field for any regular instance group, and only look at it for container groups.

The idea of setting max_forks on a container group and tracking how many forks are being used by jobs on a container group is interesting and I think worth further discussion. This feature when used with container groups is pretty cut and dry "only launch X jobs with Y pod spec". This other proposal is more dynamic, where we don't know what max number of jobs launched with Y pod spec will be, it will depend on the task_impact of the jobs.

I don't like the round-robin thing though, because that is only one pain point. The other pain point addressed by this is simply overloading a container group because we have no notion of max capacity of a container group.

AlanCoding:
Yeah, I also just kind of don't get why we would use the number of jobs metric here, as opposed to forks like we do in every other case.

Doing a schema change to add a field to instance groups is unavoidable given the problem, I would just like it to be a little more like a straightforward extension of the current system. It's true that we're using two different logic systems for the 2 types of instance groups, but it's the same model either way. I would advocate that we look at the new field as a _more restrictive_ capacity limit above and beyond what the instance capacity is. So to schedule a job, it needs to be on an instance group _and_ an instance with sufficient spare capacity. Who knows, maybe someone would like to use this for ordinary instance groups?

Say you have a cluster where the database is the limitation, not the hardware. Anyway, you're sharing the hardware with other apps, so the health check measurements aren't really useful to start with. Maybe you want to put a capacity limit on the default instance group, just to keep database usage below a certain value. That's not exactly possible tweaking individual instance capacities. Maybe just expose it for all instance groups and save yourself from writing additional exceptions?

kdelee:
For container groups my thought to go with max jobs was that each job gets a certain amount of CPU/memory requested per pod, doesn't matter how many forks/the task impact of the job. E.g. if you have 100 jobs with forks set to 1, we will attempt to schedule 100 pods if max_forks is set to 100 -- even if your namespace only actually has resources to satisfy the requests of 50 pods.

So its easier to map pod spec --> max # of pods your container group could sustain, and its something the administrator of tower would have more control over vs. the forks on the individual job templates, because the person setting up the container group likely knows how to access the kubernetes/openshift cluster namespace and can know "I can have X number of pods with this set of resource requests in this namespace" and configure it all on the container group. With max_forks its really a guessing game for the person setting up the container group how many forks the jobs assigned to it will consume.

Honestly I think we could actually implement both :smiling_imp: but I think that scares @fosterseth as far as making things complicated

kdelee:
closing in favor of  #13099 ...even if we don't decide to do both, I've added more tests/refactor over there so I'll just work off of that branch

Issue #12019, opened 2022-04-07
Consider task impact/resource requests for container group jobs on a per job basis


kdelee:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Right now when a container group job is created, we generate a pod spec based on the pod spec in the ContainerGroup. All jobs get the same resource requests/limits as is defined on the ContainerGroup podspec regardless of the calculated "task_impact" of the job.

Users can try and work around this by creating different container groups with different resource requests and/or using sliced jobs to try and split jobs into many smaller pods where the number of forks in the "slice" more closely matches the resources the pod has allocated to it.

To help users out, but not introduce breaking changes, we could

A) Allow a JT level "pod spec override"
B) Give a true/false attribute on the JT "dynamic resource requests for container group jobs" (better name needed) where we calculate the resource request based on the job's "task_impact"

##### Background
How do we limit how many forks attempt to execute in the job pod that gets launched into the container group?

Right now if you have an inventory of 1000 hosts and you set your max forks to 1000 on the job template, we will just launch it even if your underlying k8s cluster is on a laptop with 2GB of ram, because we also basically view that job pod as having unlimited resources/forks capacity (which is wrong). (really, we just don't consider capacity at all on the execution side of container groups, just the control side now since https://github.com/ansible/awx/pull/11665)

Currently we can customize the container group pod spec to add a resource request/limit, but right now that would only apply on the kubernetes level and we would STILL try and launch that 1000 fork job, though perhaps the pod will get throttled/reaped if we set a resource limit on memory/CPU.

A user can avoid trying to run a job with 1000 forks by doing some machinations with sliced jobs, but this needs to be managed by the user to make sure `inventory_hosts/slices <= max_forks_per_pod`.

### Questions/concerns

This still follows pattern of "launch the job and let kubernetes figure out if it can schedule it" -- but we still have trouble respecting Pending pods and we thrash the task manager if the namespace has resource quotas set because we get a 4XX from the Kube API and just put the job back in pending, trying to start it again on the next round of the task manager.

PR #13099, opened 2022-10-24
Allow setting max forks and max concurrent jobs per container/instance group

kdelee:
##### SUMMARY

Implements https://github.com/ansible/awx/issues/12288, which asked for "max jobs/container group" but in a more generic way, allowing users to set max_concurrent_jobs or max_forks, or both on any instance group.

For container groups, max_concurrent_jobs makes sense when you know what the pod resource requests are and what the global resource limit is on the kubernetes namespace, and you want a global limit of how many of these pods you will provision. However, it does not take into account the variability of how many forks and individual job will spawn of ansible. Allowing the setting of max_forks as well allows scenarios like:

max_concurrent_jobs: 10
max_forks: 50

If 10 jobs that use 5 forks each are run, an 11th job will wait until one of these finishes to run on that group (or be scheduled on a different group with capacity).
If 2 jobs are running with 20 forks each, then a 3rd job with a task_impact of 11 or more will wait until one of these finishes to run on that group (or be scheduled on a different group with capacity).

This does not implement the idea of dynamic resource requests on the pod as described in https://github.com/ansible/awx/issues/12019 -- but it DOES put us in the position where we can say "oh, max_forks on this container group is 100 and I have 5 jobs with forks set to 20, stop scheduling pods", which may be good enough! Especially if pod specs only set requests and not limits, the pods can hypothetically take as many resources as they want (as long as they don't surpass other more global limits that would cause kube to come reap the pod)

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### Additional Questions/Followup work

The next question is if we would want to be able to set this up from Day 0 with settings in the operator, at least for the default container group. That is probably feasible. But it gets me thinking, with all the work going into things like the awx.awx collection and the CasC collection https://github.com/redhat-cop/controller_configuration , if we instead of adding more an more options for that type of config to settings/the operator etc, if we want to have some other kind of first class way of having a post-install hook that can apply CasC config. :zap:


awx/main/scheduler/task_manager_models.py
```diff
@@ -33,9 +33,55 @@ def remaining_capacity(self):
...
+        self.max_concurrent_jobs = obj.max_concurrent_jobs
+        self.max_forks = obj.max_forks
+
+    def consume_capacity(self, task=None, control_impact=False):
+        if control_impact:
+            self.consumed_forks += settings.AWX_CONTROL_NODE_TASK_IMPACT
+            return
+        self.running_jobs += 1
+        if task and (task.execution_node in self.instance_hostnames or self.is_container_group):
+            self.consumed_forks += task.task_impact
```
kdelee: Ok, just remembered why this never worked in the old task manager, which tried to track this (but for no reason)

Instances can be members of multiple instance groups

This implementation is naive, we consume capacity when the instance group is assigned.
For container groups, this will always work, since they don't have instances.
But for regular instance groups, what about when an instance is in multiple instance groups? Do we have a reverse mapping of `instance: list of instance groups` and consume capacity on all of them? Suppose I could do that

kdelee: Resolved this now

awx/main/scheduler/task_manager.py
```diff
@@ -626,15 +625,21 @@ def process_pending_tasks(self, pending_tasks):
 
             # All task.capacity_type == 'control' jobs should run on control plane, no need to loop over instance groups
             if task.capacity_type == 'control':
+                if not self.instance_groups[self.controlplane_ig.name].has_remaining_capacity(control_impact=True):
+                    logger.debug(f"Skipping instance group {self.controlplane_ig.name}, already met max forks allowed to execute on Instance Group")
```
AlanCoding: This log statement could use some added context about `task`, otherwise it'll be a bunch of repeated and indecipherable logs.

awx/main/scheduler/task_manager_models.py
```diff
@@ -64,58 +133,67 @@ def __contains__(self, hostname):
...
 
     def get_consumed_capacity(self, group_name):
-        instances = self.instance_groups[group_name]['instances']
-        return sum(inst.consumed_capacity for inst in instances)
+        return self.instance_groups[group_name].get_consumed_instance_capacity()
+
+    def consume_capacity(self, task):
+        self.task_manager_instances.consume_capacity(task)
+        if task.instance_group is not None and task.instance_group.is_container_group:
+            self.instance_groups[task.instance_group.name].consume_capacity(task)
```
AlanCoding:
```python
        if task.instance_group_id:
             ig = self.pk_ig_map[task.instance_group_id]
            if ig.is_container_group:
                self.instance_groups[ig.name].consume_capacity(task)
```

for performance, I think this will allow dropping the prefetch of `instance_group`

awx/main/scheduler/task_manager_models.py
```diff
@@ -139,3 +228,20 @@ def get_instance_groups_from_task_cache(self, task):
...
+        self.instances = TaskManagerInstances(**kwargs)
+        self.instance_groups = TaskManagerInstanceGroups(task_manager_instances=self.instances, **kwargs)
+        if kwargs.get('init_consumed_capacity', False) or 'tasks' in kwargs:
+            tasks = kwargs.get('tasks', None)
+            if tasks is None:
+                from awx.main.models import UnifiedJob
+
+                tasks = UnifiedJob.objects.filter(status__in=['running', 'waiting']).only('task_impact', 'controller_node', 'execution_node', 'instance_group')
+            for task in tasks:
+                self.instance_groups.consume_capacity(task)
```
AlanCoding: I would like to better isolate `init_consumed_capacity` from the logic that runs in the task manager by a separate `@classmethod` constructor.

like

```python
@classmethod
def init_with_consumed_capacity(cls, **kwargs):
    tmm = cls(**kwargs)
    tasks = kwargs.get('tasks', None)
    if tasks is None:
        from awx.main.models import UnifiedJob

        tasks = UnifiedJob.objects.filter(status__in=['running', 'waiting']).only('task_impact', 'controller_node', 'execution_node', 'instance_group')
    for task in tasks:
        self.instance_groups.consume_capacity(task)
```

Then call by `TaskManagerModels.init_with_consumed_capacity()`

kdelee: sounds good, will do

kdelee: done

awx/main/scheduler/task_manager_models.py
```diff
@@ -64,61 +137,77 @@ def __contains__(self, hostname):
...
-        instances = self.instance_groups[group_name]['instances']
-        return sum(inst.consumed_capacity for inst in instances)
+        return self.instance_groups[group_name].get_consumed_instance_capacity()
+
+    def consume_capacity(self, task):
+        self.task_manager_instances.consume_capacity(task)
+        if task.instance_group_id:
+            ig = self.pk_ig_map[task.instance_group_id]
+            if ig.is_container_group:
+                self.instance_groups[ig.name].consume_capacity(task)
```
AlanCoding: I find it a little unexpected that `TaskManagerInstanceGroups.consume_capacity` deducts the capacity from the instance group as well as the instance(s) the job runs on.

There are 4 methods for `consume_capacity`, and this isn't a bad thing, but with that number it becomes hard to predict, so it might be best to follow expectations as closely as possible. As it stands, it might be more consistent to introduce a `TaskManagerModels.consume_capacity` method to deduct capacity from both the IGs and the instances, moving the `self.task_manager_instances.consume_capacity(task)` line from here to there. Maybe ask @fosterseth for his input on this matter as well. My suggestion here would require changing the task manager to keep the `TaskManagerModels` instance as opposed to the `TaskManagerInstanceGroups`, but that might obligate some other changes to the references between classes.

awx/main/scheduler/task_manager.py
```diff
@@ -626,15 +629,19 @@ def process_pending_tasks(self, pending_tasks):
...
+                execution_instance = self.instance_groups.task_manager_instances[control_instance.hostname].obj
                 task.log_lifecycle("controller_node_chosen")
                 task.log_lifecycle("execution_node_chosen")
                 self.start_task(task, self.controlplane_ig, task.get_jobs_fail_chain(), execution_instance)
                 found_acceptable_queue = True
                 continue
 
             for instance_group in self.instance_groups.get_instance_groups_from_task_cache(task):
+                if not self.instance_groups[instance_group.name].has_remaining_capacity(task):
+                    continue
```
AlanCoding, line 646: I do think we should have a log, like:

```python
                    logger.debug(
                        f"Instance group {instance_group.name} not available to run job {task.log_format} w/ capacity requirement {task.task_impact}"
                    )
```

AlanCoding:
This looks good to me from the API code design side of things :+1: 

There are a few more boxes to check, like testing and UI, but the main API development looks good.

awx/main/scheduler/task_manager.py
kdelee: I did add the log in the `has_remaining_capacity` though probably should update to use `task.log_format` i'll comment there to point it out

awx/main/scheduler/task_manager_models.py
kdelee: I agree its a bit weird to have the method on `TaskManagerInstanceGroups` but have it apply to groups and instances, I'll shop some ideas around with seth and we can move forward. I don't think any other place other than the task manager ever uses any `consume_capacity` function so it should be relatively small lift

awx/main/scheduler/task_manager_models.py
```diff
@@ -33,28 +36,98 @@ def remaining_capacity(self):
...
+            remaining_forks = self.max_forks - instance_consumed_forks - self.container_group_consumed_forks - task_impact
+
+        if remaining_jobs < 0 or remaining_forks < 0:
+            # A value less than zero means the task will not fit on the group
+            task_string = f"task {task.id}" if task else "control task"
+            if remaining_jobs < 0:
+                logger.debug(f"{task_string} cannot fit on instance group {self.name} with {remaining_jobs} remaining jobs")
+            if remaining_forks < 0:
+                impact_string = f"with impact {task_impact}"
+                logger.debug(f"{task_string} {impact_string} cannot fit on instance group {self.name} with {remaining_forks} remaining forks")
```
kdelee: here is where I'm logging if indeed the instance group does NOT have capacity, and it tells you why (either hit max forks or max concurrent jobs)

awx/main/scheduler/task_manager.py
kdelee: pointed out later where I do log in the case that we max_forks or max_concurrent_jobs is blocking (only at debug level)

awx/main/scheduler/task_manager_models.py
kdelee: implemented the suggestion

kdelee:
Screenshot from 2022-10-26 22-04-24
Screenshot from 2022-10-26 22-04-17
Screenshot from 2022-10-26 22-04-05
Screenshot from 2022-10-26 22-03-58

image

AlexSCorey (approved):
Looks good to me.

AlanCoding:
Summarizing remaining actions left to do

### Setting-based default

To make this fit the intended use in OCP, we need to have a setting which will give _the default value for the OCP `default` instance group_. This instance group is created in some bootstrap scripts by the `awx-manage register_queue` command, and we don't want to change that, it will be easier to have it get a default value from a setting.

### Docs

We need some some highly specific instructions for the primary intended use case. I find this very similar to `SYSTEM_TASK_ABS_MEM`, which is used to correct the capacity of control plane instances.

I'd like our docs approach to tackle the general problem of "how do I assure my OCP cluster is using accurate capacity for job distribution?" Then answer would be to add some things to `extra_settings` in the operator deployment where you set that old setting, as well as the new setting.

### Integration testing

It's expected that a serial test will be needed to exercise the basic functionality of this feature.

kdelee:
@AlanCoding ok added the settings for the default container group and the API tests/operator tests passing

CFSNM (approved):

awx/main/scheduler/task_manager.py
```diff
@@ -471,9 +475,8 @@ def after_lock_init(self):
         Init AFTER we know this instance of the task manager will run because the lock is acquired.
         """
         self.dependency_graph = DependencyGraph()
-        self.instances = TaskManagerInstances(self.all_tasks)
-        self.instance_groups = TaskManagerInstanceGroups(instances_by_hostname=self.instances)
-        self.controlplane_ig = self.instance_groups.controlplane_ig
+        self.tm_models = TaskManagerModels()
+        self.controlplane_ig = self.tm_models.instance_groups.controlplane_ig
```
AlanCoding, line 479: super duper minor comment, but I'd prefer to save `controlplane_ig` as a local variable in process_pending_tasks instead of `self.controlplane_ig`.

But the semantic of `self.tm_models` is an improvement from before, this is going in the right directions.

AlanCoding (approved):
I had already gone through this pretty completely before. I wanted to make sure the core feature testing was having success, and it is now. :100:

akus062381:
I was running into errors in the UI/API, and Elijah helped me diagnose that the error is related to this endpoint:
api/v2/job_templates/7/instance_groups/

akus062381:
this looks great

kdelee:
> I was running into errors in the UI/API, and Elijah helped me diagnose that the error is related to this endpoint: api/v2/job_templates/7/instance_groups/

We fixed this now! Thanks @akus062381

awx/main/scheduler/task_manager_models.py
```diff
@@ -139,3 +254,56 @@ def get_instance_groups_from_task_cache(self, task):
...
+            task_fields = kwargs.get('task_fields', ('task_impact', 'controller_node', 'execution_node', 'instance_group'))
+            from awx.main.models import UnifiedJob
+
+            if instance_group_queryset is not None:
+                logger.debug("******************INSTANCE GROUP QUERYSET PASSED -- FILTERING TASKS ****************************")
+                # Sometimes things like the serializer pass a queryset that looks at not all instance groups. in this case,
+                # we also need to filter the tasks we look at
+                tasks = UnifiedJob.objects.filter(status__in=task_status_filter_list, instance_group__in=[ig.id for ig in instance_group_queryset]).only(
+                    *task_fields
+                )
```
kdelee, line 285: @AlanCoding FYI @akus062381 found a bug in this when we were making requests to /api/v2/job_templates/<ID>/instance_groups, the serializer calls this with a instance group queryset already formed. So you could get a 500 if there were jobs running on a DIFFERENT instance group that wasn't in the queryset, because later in `consume_capacity` we assumed all the instance groups on the tasks should be in the instance group mapping

matburt:
Going into the merge review meeting I'd love to see details about this added to the right files in `docs/` : https://github.com/ansible/awx/tree/devel/docs

awx/main/scheduler/task_manager_models.py
AlanCoding: Oh, I get what you're saying about this but. Good find.

I'm on the fence about whether `instance_groups_queryset` is worth the effort we're putting in to maintain it. Think about the multiple types of optimizations. We have a single IG in a list view of IGs. The main optimization the serializer is trying to accomplish is to avoid re-computing capacity for every object in the list. Even if we incur 1 additional query, we can save those extra 20.

On the other hand, I see how we avoid tracking capacity for a _large number_ of instance groups if the user created a lot...

kdelee: yeah so alternative to this fix is to drop the instance_group_queryset argumetn and always compute everything for all groups and all tasks......

but there is a tiny part of me that is still dreaming of partitioning the task manager work on instance groups and making it where instances can only belong to 1 instance group :smiling_imp:

AlanCoding:
I can see how `docs/capacity.md` is an appropriate context for a summary of this feature. Looking at the doc now, I don't see anything outright wrong, unlike `docs/task_manager_system.md`, which needs substantial refactoring.

kdelee:
image
 bug! need to fix

kdelee:
@AlexSCorey helped me fix that form edit problem so we can successfully edit and set values back to 0

akus062381 (approved):

PR #13101, opened 2022-10-24
Adds info about git workflows for feature branches

AlexSCorey:
##### SUMMARY
No issue for this work.  This PR documents the new workflow for feature branches.  I put it in the contributing guide but if we feel it belongs somewhere else lets chat.
##### ISSUE TYPE

##### COMPONENT NAME
 - Docs

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -107,7 +107,7 @@ When you attempt to perform a `git commit` there will be a pre-commit hook that 
 
 While you can use environment variables to skip the pre-commit hooks GitHub will run similar tests and prevent merging of PRs if the tests do not pass.
 
- If you would like to add additional commit hooks for your own usage you can create a directory in the root of the repository called `pre-commit-user`. Any executable file in that directory will be executed as part of the pre-commit hooks. If any of the pre-commit checks fail the commit will be halted. For your convenience in user scripts, a variable called `CHANGED_FILES` will be set with any changed files present in the commit.
+If you would like to add additional commit hooks for your own usage you can create a directory in the root of the repository called `pre-commit-user`. Any executable file in that directory will be executed as part of the pre-commit hooks. If any of the pre-commit checks fail the commit will be halted. For your convenience in user scripts, a variable called `CHANGED_FILES` will be set with any changed files present in the commit.
 
 ## What should I work on?
 
@@ -121,8 +121,7 @@ If it has someone assigned to it then that person is the person responsible for 
 
 **NOTES**
 
-> Issue assignment will only be done for maintainers of the project. If you decide to work on an issue, please feel free to add a comment in the issue to let others know that you are working on it; but know that we will accept the first pull request from whomever is able to fix an issue. Once your PR is accepted we can add you as an assignee to an issue upon request. 
-
+> Issue assignment will only be done for maintainers of the project. If you decide to work on an issue, please feel free to add a comment in the issue to let others know that you are working on it; but know that we will accept the first pull request from whomever is able to fix an issue. Once your PR is accepted we can add you as an assignee to an issue upon request.
 
 > If you work in a part of the codebase that is going through active development, your changes may be rejected, or you may be asked to `rebase`. A good idea before starting work is to have a discussion with us in the `#ansible-awx` channel on irc.libera.chat, or on the [mailing list](https://groups.google.com/forum/#!forum/awx-project).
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -132,8 +131,39 @@ If it has someone assigned to it then that person is the person responsible for
...
+
+It is important that the feature branch be kept up-to-date with awx/devel to limit merge conflicts that might arise when merging the feature into devel. To do this someone with admin privileges on the AWX repo needs to remove branch protection and then follow these steps (In this example I will use origin as the remote name for AWX):
+
+- `git fetch origin`
+- `git merge origin/devel -ff-only` merges the branch you are on with awx/devel and fast forwards.
```
relrod:
```suggestion
- `git merge origin/devel --ff-only` merges the branch you are on with awx/devel and fast forwards.
```

(and similarly below)

CONTRIBUTING.md
```diff
@@ -132,8 +131,39 @@ If it has someone assigned to it then that person is the person responsible for
...
+
+### Common Git workflows
+
+In general there are two separate git workflows to follow when contributing code for a net new feature to AWX. One for keeping a feature branch up to date with devel, and another, used by individual contributors to keep their working branch up to date with the feature branch they are working on. When initiating work on a net new feature a new feature branch should be checked out from awx/devel and branch protection should be activated. Checking out the new branch, and activating branch protection must be done by someone with admin privileges on the AWX repo. This branch will be used by the developers working on the feature and merging work into the feature branch will allow the developers to work quickly without impacting devel. This also means that those working on the feature can decide what requirements need to pass before merging work into the feature branch. Merge meeting, QE sign off etc. can all be done at 1 time when the feature is ready to merge into awx/devel.
```
jbradberry, line 138: "When initiating work on a net new feature..."

I think this needs to be more specific, something like "on epic-level features".  I can imagine that we'll still occasionally have smaller features that are not going to need this long-lived feature branch workflow.

jbradberry: '1 time' -> 'one time'.

CONTRIBUTING.md
```diff
@@ -132,8 +131,39 @@ If it has someone assigned to it then that person is the person responsible for
...
+
+It is important that the feature branch be kept up-to-date with awx/devel to limit merge conflicts that might arise when merging the feature into devel. To do this someone with admin privileges on the AWX repo needs to remove branch protection and then follow these steps (In this example I will use origin as the remote name for AWX):
+
+- `git fetch origin`
+- `git merge origin/devel --ff-only` merges the branch you are on with awx/devel and fast forwards.
+
+If there are merge conflicts we recommend not using `git merge origin/devel -ff-only` and instead rebasing. Those steps are:
```
jbradberry, line 145: `--ff-only`

CONTRIBUTING.md
```diff
@@ -132,8 +131,39 @@ If it has someone assigned to it then that person is the person responsible for
...
+
+It is important that the feature branch be kept up-to-date with awx/devel to limit merge conflicts that might arise when merging the feature into devel. To do this someone with admin privileges on the AWX repo needs to remove branch protection and then follow these steps (In this example I will use origin as the remote name for AWX):
```
jbradberry, line 140: I'd recommend using `upstream` instead of `origin`.  I think the most typical setup for dev checkouts on our team has origin == the developer's fork and upstream == main fork.

CONTRIBUTING.md
```diff
@@ -132,8 +131,39 @@ If it has someone assigned to it then that person is the person responsible for
...
+
+- `git fetch origin`
+- `git merge origin/devel --ff-only` merges the branch you are on with awx/devel and fast forwards.
+
+If there are merge conflicts we recommend not using `git merge origin/devel -ff-only` and instead rebasing. Those steps are:
+
+- `git fetch origin` (you might have done this step above so it is not necessary to repeate it here)
```
jbradberry, line 147: -> repeat

CONTRIBUTING.md
```diff
@@ -132,8 +131,39 @@ If it has someone assigned to it then that person is the person responsible for
...
+- `git rebase -i origin/devel` (this will do an interactive rebase and make it easier to work through any merge conflicts)
 
+Then you are done and your feature branch is up-to-date with awx/devel. Just push your feature branch to `awx/feature-branch`.
```
jbradberry, line 150: We should be more specific.  I think you should refer to "the local copy of the feature branch" instead of "your feature branch", and you should probably also spell out the command to push in the same way you did the rest of the commands.

CONTRIBUTING.md
```diff
@@ -132,8 +131,39 @@ If it has someone assigned to it then that person is the person responsible for
 
 At this time we do not accept PRs for adding additional language translations as we have an automated process for generating our translations. This is because translations require constant care as new strings are added and changed in the code base. Because of this the .po files are overwritten during every translation release cycle. We also can't support a lot of translations on AWX as its an open source project and each language adds time and cost to maintain. If you would like to see AWX translated into a new language please create an issue and ask others you know to upvote the issue. Our translation team will review the needs of the community and see what they can do around supporting additional language.
 
-If you find an issue with an existing translation, please see the [Reporting Issues](#reporting-issues) section to open an issue and our translation team will work with you on a resolution. 
+If you find an issue with an existing translation, please see the [Reporting Issues](#reporting-issues) section to open an issue and our translation team will work with you on a resolution.
+
+### Common Git workflows
```
jbradberry, line 136: I just realized, this should be a level-2 header, not level-3.

jbradberry: And I think we should have two level-3 subheaders under here, for 1) the 'keeping the feature branch up to date' flow, and 2) the individual contributors' flow.

CONTRIBUTING.md
```diff
@@ -132,8 +131,39 @@ If it has someone assigned to it then that person is the person responsible for
...
+- `git fetch origin`
+- `git merge origin/devel --ff-only` merges the branch you are on with awx/devel and fast forwards.
+
+If there are merge conflicts we recommend not using `git merge origin/devel -ff-only` and instead rebasing. Those steps are:
+
+- `git fetch origin` (you might have done this step above so it is not necessary to repeate it here)
+- `git rebase -i origin/devel` (this will do an interactive rebase and make it easier to work through any merge conflicts)
 
+Then you are done and your feature branch is up-to-date with awx/devel. Just push your feature branch to `awx/feature-branch`.
+The second workflow should be used by individual contributors to keep their working branch up-to-date with the feature branch. Those steps are:
```
jbradberry, line 151: This isn't really describing what is happening below.  These commands seem to be more about initially creating a branch off of the feature branch to be used for the contributor's PR.

You'll also need to add a series of commands for keeping your PR branch up to date with respect to the feature branch.

CONTRIBUTING.md
```diff
@@ -132,8 +131,39 @@ If it has someone assigned to it then that person is the person responsible for
...
+
+- `git fetch upstream`
+- `git checkout origin/feature-branch`
```
jbradberry, line 154: I assume that the inclusion of the remote here is to avoid the situation where the local checkout of the feature branch becomes out of date.  If so, the remote ought to match the git fetch command above (I'd recommend going with `upstream` for both, per my earlier comment).  If not, just go with `git checkout feature-branch`, but you'll need to include a command to bring it up to date with the official version (just `git merge upstream/feature-branch` is what I do.  Include `--ff-only` if paranoid.)

CONTRIBUTING.md
```diff
@@ -132,8 +131,39 @@ If it has someone assigned to it then that person is the person responsible for
...
+- `git checkout working-branch-name` (Many people put the issue number in the branch name. ie. 12556-Working-branch-name)
```
jbradberry, line 155: I assume that creating a new PR branch is what is intended here, so you need to add `-b` to create the new branch.

CONTRIBUTING.md
```diff
@@ -132,8 +131,39 @@ If it has someone assigned to it then that person is the person responsible for
...
+
+Now you can start working on your issue. When ready to make a commit:
+
+- `git add .` Will stage all files that were touched. If you want to specify specific files to stage `git add file-path`
+- `git commit` You will be prompted to write a commit message.
+
+When ready to make a PR to the feature branch.
```
jbradberry, line 162: I strongly recommend calling out the need to select the target to be the feature branch when creating the PR on github.  People forget that all of the time.  Even I sometimes forget it.

jbradberry: ... oh, I see you did that on line 166.

jbradberry: Maybe include it in the list of steps to make it more obvious?

jbradberry:
@AlexSCorey actually, since this all is very specific to our team's workflow, I think that this should _not_ live in this document.

AlexSCorey:
Closing this in favor of adding to our internal docs.

PR #13102, opened 2022-10-25
Clean up ssh_key_data in timeout cases too

AlanCoding:
##### SUMMARY
Attempts to fix an issue where jobs sometimes were left in the "running" status forever.

Link original change this builds on https://github.com/ansible/awx/pull/11681

##### ISSUE TYPE
 - Bug, Docs Fix or other nominal change

##### COMPONENT NAME
 - API

PR #13114, opened 2022-10-25
Remove ssh_key_data fix, handled in runner now

AlanCoding:
##### SUMMARY
Making this PR for purposes of testing.

This does a complete 180 from https://github.com/ansible/awx/pull/13102, and removes the hack as opposed to expanding its scope. Since the runner fix is in `unstream_dir`, it should be active in the needed context.

This makes use of the fix https://github.com/ansible/ansible-runner/pull/1133

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/tasks/callback.py
+++ b/awx/main/tasks/callback.py
@@ -2,8 +2,6 @@ import json
 import time
 import logging
 from collections import deque
-import os
-import stat
 
 # Django
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


AlanCoding:
@fosterseth I haven't forgotten about this. Just got delayed due to flake when trying to run integration tests. Restarted just now :crossed_fingers:

AlanCoding:
tests are good

requirements/requirements_git.txt
```diff
@@ -1,6 +1,6 @@
 git+https://github.com/ansible/system-certifi.git@devel#egg=certifi
 # Remove pbr from requirements.in when moving ansible-runner to requirements.in
-git+https://github.com/ansible/ansible-runner.git@devel#egg=ansible-runner
+git+https://github.com/fosterseth/ansible-runner.git@fix_ssh_key_data_pipe_reading#egg=ansible-runner
```
AlanCoding: We can undo this when the PR is merged.

fosterseth (approved):

AlanCoding:
I un-did the change to our git requirements. The runner change merged a few days ago, so this should be workable now.

AlanCoding:
Finally confirmed that hang on production instances from before the runner fix, and confirmed it's still fixed with this change. Merging soon.

PR #13124, opened 2022-10-26
[performance] Reduce unnecessary queries for job launch and start process

AlanCoding:
##### SUMMARY
This focuses on inefficiencies seen in the process of:

```python
job = jt.launch()
job.signal_start()
```

This is done in the task manager for spawning dependencies, and separately in the workflow manager for workflow jobs. So this is, in some significant part, a task manager performance enhancement as well.

For some of this, I've tried to make similar performance improvements in the past. Here, I've tried to keep it fairly conservative.

I would like to be able to get this in to enable us to move more responsibilities into the launch request-response cycle as in https://github.com/ansible/awx/pull/13061

I measured 93 queries before and 55 queries afterwards for a particular scenario, for about a 40% improvement.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
This is all really low-hanging fruit. Like, if you do `job.save(update_fields=["status"])`, then you should need to update the RBAC structure, or obtain the related inventory from the database. Or, when apply credentials/labels to a job just created, there's not a need to fetch every related object in support of that...

### DIFF
--- a/awx/main/fields.py
+++ b/awx/main/fields.py
@@ -284,6 +284,28 @@ class ImplicitRoleField(models.ForeignKey):
         ct_id = ContentType_.objects.get_for_model(instance).id
 
         Model = utils.get_current_apps().get_model('main', instance.__class__.__name__)
+
+        update_fields = kwargs.get('update_fields', ())
+        is_rbac_update = False
+        if update_fields:
+            for field_name in update_fields:
+                try:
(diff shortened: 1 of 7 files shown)


AlanCoding:
These signals will not be hit if we move to the new RBAC system from django-ansible-base which is the current plan.

PR #13132, opened 2022-10-28
Expand the dependencies of the Schedule page type

jbradberry:
##### SUMMARY
Really these could get any of the unified job template types, not just system job templates, so importing e.g. a project with a schedule was doing them in the wrong order.

Also, bump the timeout of the project update and make sure that we stash it in the page cache even if it doesn't finish in 5 minutes.

related #13084
replaces #13090

##### ISSUE TYPE

##### COMPONENT NAME
 - CLI

### DIFF
--- a/awxkit/awxkit/api/pages/api.py
+++ b/awxkit/awxkit/api/pages/api.py
@@ -275,7 +275,13 @@ class ApiV2(base.Base):
                         # When creating a project, we need to wait for its
                         # first project update to finish so that associated
                         # JTs have valid options for playbook names
-                        _page.wait_until_completed()
+                        try:
+                            _page.wait_until_completed(timeout=300)
+                        except AssertionError:
+                            # If the project update times out, try to
+                            # carry on in the hopes that it will
+                            # finish before it is needed.
+                            pass
                 else:
                     # If we are an existing project and our scm_tpye is not changing don't try and import the local_path setting
                     if asset['natural_key']['type'] == 'project' and 'local_path' in post_data and _page['scm_type'] == post_data['scm_type']:
--- a/awxkit/awxkit/api/pages/schedules.py
+++ b/awxkit/awxkit/api/pages/schedules.py
@@ -1,6 +1,7 @@
 from contextlib import suppress
 
-from awxkit.api.pages import SystemJobTemplate
+from awxkit.api.pages import JobTemplate, SystemJobTemplate, Project, InventorySource
+from awxkit.api.pages.workflow_job_templates import WorkflowJobTemplate
 from awxkit.api.mixins import HasCreate
 from awxkit.api.resources import resources
 from awxkit.config import config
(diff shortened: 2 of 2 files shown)


AlanCoding (approved):

ArtsiomMusin:
@jbradberry one question about the second issue I mentioned in #13090 . After your changes do you see this error after importing a schedule?
> /api/v2/schedules/ "myschedule": Bad Request (400) received - {'inventory': ['Field is not allowed on launch.']}.

For me it wasn't enough to fix the order. Or maybe someone fixed that in the latest devel branch?

jbradberry:
@ArtsiomMusin I do not get that error, no.  My changes are sufficient to get at least my minimal reproducer case to work.

That message seems to be coming from here: https://github.com/ansible/awx/blob/e655e1dbc2925acea8e4f5632732eaa758948486/awx/main/models/unified_jobs.py#L471.  I think it is being called by the API validation, which means it is out of scope of changes to the awx cli.

ArtsiomMusin:
@jbradberry just for my understanding. I thought this code you mentioned is also used in awx cli. 
I don't use awx cli directly. I have just a simple python script which is doing this:
```python
import json
from awxkit.api.client import Connection
from awxkit.api.pages.api import ApiV2, EXPORTABLE_RESOURCES
from awxkit.api import get_registered_page

tower_url = '<tower url here>'
token = '<tower token here>'
awx_api_connection = Connection(tower_url, verify=False)
awx_api_connection.login(None, None, token=token)
awx_v2_index = get_registered_page('/api/v2/')(awx_api_connection).get()
awx_api = ApiV2(connection=awx_api_connection, **{'json': awx_v2_index})

with open("abc1.json","r") as f: org_data = f.read()
data = json.loads(org_data)
awx_api.import_assets(data)
``` 
where does this API validation happen?

ArtsiomMusin:
@jbradberry I tried awx cli to see how it works for this issue.

Before any changes, failed as expected:
```bash
$ awx -k --conf.host https://mytower --conf.token mytoken import < tower_org_backup/abc1.json 
/api/v2/schedules/ "abc1_schedule_inv_source": Bad Request (400) received - {'unified_job_template': ['This field may not be null.']}.
```

After injecting your fix, failed with the second issue I mentioned before:
```bash
$ awx -k --conf.host https://mytower --conf.token mytoken import < tower_org_backup/abc1.json 
/api/v2/schedules/ "abc1_schedule_inv_source": Bad Request (400) received - {'inventory': ['Field is not allowed on launch.']}.
```

So the second issue still remains for me. Any thoughts?

jbradberry:
@ArtsiomMusin That second issue is going to be a completely separate issue, one that you would (probably) hit by trying to do the same thing directly in the browseable api.  I suspect that if this is an actual bug, that it would be related to the recent expansion of prompt on launch functionality.

My recommendation would be to try to diagnose it using the api directly, and open up a new issue.  If you can provide a reproducer, that would be terrific.

ArtsiomMusin:
@jbradberry ok sure will open a new issue. Btw, maybe you didn't see this issue if you created a schedule for a project. Try to import schedules for inventory sources or job templates.

Issue #12576, opened 2022-07-26
UI sometimes gives "waiting" status while job is running


AlanCoding:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

Launching the Demo Job Template with current `devel`, I can relatively frequently observe (say 1 time in 4) a job staying in the "waiting" state in the UI while it is actually running. As the job finishes, it then flips from waiting to successful.

So the UI is not processing the "running" status message (which is sent by the API), or some other data incorrectly takes precedence over that.

### AWX version

devel

### Select the relevant components

- [X] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

docker development environment

### Modifications

no

### Ansible version

N/A

### Operating system

N/A

### Web browser

Chrome

### Steps to reproduce

Launch the Demo Job Template several times. It may take a handful of tries to get this behavior. I tend to use the relaunch button.

### Expected results

I expect that the job status will be running while I am getting events coming in.

### Actual results

Stays in the waiting status until it completes.

![Screenshot from 2022-07-26 16-53-15](https://user-images.githubusercontent.com/1385596/181110374-fe8e1149-b9b6-40b9-a0fb-509342c35a95.png)

### Additional information

I check the websocket network traffic and see that all status transitions are sent for the job in question - pending, waiting, running, and successful.

Interestingly, I do see that the UI makes multiple _API_ requests for the job here.

![Screenshot from 2022-07-26 16-52-45](https://user-images.githubusercontent.com/1385596/181110515-5b41cb7c-b897-44c8-912a-1b81d60b955c.png)

In the cases where the status gets stuck in waiting, I do see an API request returning the waiting status. It is entirely possible that the API reports this waiting status after the "running" websocket message has come in, due to comparative slowness in request processing.

I found this issue when researching for https://github.com/ansible/awx/pull/12573, where, bizarrely, completely removing the websocket message for the "waiting" message did not avoid this bug. That gives further evidence to back the idea that the UI gives a stale status based on an API request, which is extremely likely to predate the websocket data.

msilmser:
[...]

PR #13154, opened 2022-11-02
Fix running job showing "waiting" status

keithjgrant:
##### SUMMARY
Fixes scenario where job status sometimes shows as "waiting" even after the job has started.

Two things had to occur for the bug to present itself: first, the job had to load from the API with a status of "waiting", and second, the websocket message indicating a "running" status had to be received before the job finished loading. When this occurred, the page attempted to update the job in memory, but since no job was loaded, the message was essentially ignored. Then, as the job object finished loading, it showed as "waiting" on screen.

I addressed this by adding a `pendingMessages` queue to `useWsJob` when the job in memory is null. Then, when a job loads, this queue is flushed and any messages in it are processed as normal.

Addresses #12576

##### ISSUE TYPE

##### COMPONENT NAME
 - UI


AlanCoding (approved):
I am very familiar with the reproducer, so I checked this out and tried it. Can confirm that the reproducer no longer sees jobs go into "waiting" state when it should be running. The series of websocket subscriptions appears to otherwise be the same, which is expected based on your description of this change.

kialam (approved):
Pulled this down and tested locally. LGTM!

akus062381 (approved):

Issue #1782, opened 2018-04-18
Add yes/no checkbox options for user surveys


alexander-bauer:
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
 - API
 - UI

##### SUMMARY
I manage a number of playbooks designed for automating highly-branched tasks that used to be performed by operators. The nature of this branching logic is highly boolean, and so I *very* frequently find myself using `multiple choice` type user surveys with `[true, false]` options. This requires more effort to make the survey, more effort to complete it, and requires care in input sanitation (where inputs are not `true` and `false`, but `"true"` and `"false"`. The `bool` filter is a life-saver.)

User surveys should support checkbox type fields, with `true` and `false` defaults.

Perhaps they could be trinary checkboxes, with `true`, `false`, and `omitted`, to be more in-line with other non-required extra variables, but that strikes me as a potential UI headache.

##### ENVIRONMENT
* AWX version: 1.0.5
* AWX install method: docker on linux

wenottingham:
duplicate of https://github.com/ansible/awx/issues/553

Issue #1053, opened 2018-01-25
Boolean type for surveys


gfichera:
##### ISSUE TYPE

 - Feature Idea

##### COMPONENT NAME

 - API
 - UI

##### SUMMARY
It would be great if in survey editor if we could choose a boolean formular.

##### ENVIRONMENT
* AWX version: 1.0.2.285
* AWX install method: docker on linux, 
* Ansible version:  2.4.2.0
* Operating System: Linux
* Web Browser: chrome

##### STEPS TO REPRODUCE

In job_template > edit_survey > answer_type 
adding a boolean type.

wenottingham:
covered by https://github.com/ansible/awx/issues/553

PR #13157, opened 2022-11-04
Allow surveys to have boolean questions

relrod:
##### SUMMARY

A go at boolean survey questions.

Refs #553
Refs #1782
Refs #1053

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 - UI

### DIFF
--- a/awx/api/templates/api/job_template_survey_spec.md
+++ b/awx/api/templates/api/job_template_survey_spec.md
@@ -28,6 +28,7 @@ Within each survey item `type` must be one of:
 * float: For survey questions expecting a decimal number
 * multiplechoice: For survey questions where one option from a list is required
 * multiselect: For survey questions where multiple items from a presented list can be selected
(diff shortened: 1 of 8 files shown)


goldyfruit:
@relrod does this PR still in draft ?

AlexSCorey (approved):
This looks good to me. Well done on the UI work!

AlanCoding:
Your test at `awx/main/tests/unit/models/test_survey_models.py` is good, but that handles an error case, and we need _more_ testing than this for a high-level feature like this. I want to see the happy path tested, which is that `true` and `false` values are accepted... probably should verify this in both YAML / JSON. That can be done in the API tests with a mock request.

Beyond testing with the test database, we should also have full-cycle integration tests which test the happy path of valid survey answers and look for the value in a debug task that writes out hostvars (playbooks to accomplish this already exist).

AlanCoding (approved):

PR #13158, opened 2022-11-04
Removing dependency on asciichartpy

john-westcott-iv:
It was only used in one management command. We have left the command but exit with an error if the user does not have the library installed. Users can still choose to add the library into their system to use the commend.

##### ISSUE TYPE

##### COMPONENT NAME

 - Other

##### AWX VERSION

```
awx: 21.8.1.dev3+gadcdb4945f
```

##### ADDITIONAL INFORMATION

```

```


jbradberry (changes requested):
I'm going to veto this.

1. This is a great example of a library that doesn't need to change much.
2. Our usage of it is in a low risk area.
3. Perusal of the issues list of the github project indicates that there are not many open tickets, and none that seem to be critical + ignored.
4. While it has been a while since this library was _released_, the last commit at time of writing was Sep 7 this year.

awx/main/management/commands/graph_jobs.py
```diff
@@ -1,9 +1,14 @@
 # Python
-import asciichartpy as chart
 import collections
 import time
 import sys
 
+try:
+    import asciichartpy as chart
+except:
+    print("Error: this command requires asciichartpy library to be installed")
```
AlanCoding, line 9: Might be defunct already, but...

The `except` should be of an `ImportError`, also I think the message should give a `pip install ...` recommendation, because the _import name_ is often different from the library name, and unless it is spelled out exactly I may feel the need to go look it up. Also, maybe state that it has to be done by root? Into awx venv?

shanemcd (changes requested):
I agree with Jeff here.

PR #13162, opened 2022-11-07
update awx collection workflow module schema with new options

sean-m-sullivan:
##### SUMMARY
Added options for forks, job_slice_count, timeout, instance groups, labels, execution environment to workflow job template nodes in the workflow_nodes option. 

Also added ability to optionally specify the organization credentials and labels reside in.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - Collection

##### AWX VERSION
```
awx: 21.8.1
```

### DIFF
--- a/awx_collection/plugins/modules/workflow_job_template.py
+++ b/awx_collection/plugins/modules/workflow_job_template.py
@@ -208,6 +208,29 @@ options:
           description:
             - Limit to act on, applied as a prompt, if job template prompts for limit
           type: str
+        forks:
+          description:
+            - The number of parallel or simultaneous processes to use while executing the playbook, if job template prompts for forks
+          type: int
+        job_slice_count:
+          description:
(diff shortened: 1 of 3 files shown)


awx_collection/plugins/modules/workflow_job_template.py
```diff
@@ -474,11 +536,21 @@
             name: Default
           name: job template 2
           type: job_template
+        execution_environment:
+            name: testee
```
john-westcott-iv: Do you want to make these names a little more generic?

john-westcott-iv (approved):

john-westcott-iv:
We def should come up with a way to make the test_completness script look into this to ensure we don't run into this again later in the future.

awx_collection/plugins/modules/workflow_job_template.py
sean-m-sullivan: Updated Names to match other Documented examples.

jay-steurer (approved):

PR #13164, opened 2022-11-07
Fix for incorrect error messages on incomplete credentials.

djyasin:
##### SUMMARY

Making error messages better for incomplete credentials

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.8.1
```


awx/main/models/unified_jobs.py
```diff
@@ -1345,15 +1345,17 @@ def pre_start(self, **kwargs):
...
             defined_fields = credential.credential_type.defined_fields
             for required in credential.credential_type.inputs.get('required', []):
                 if required in defined_fields and not credential.has_input(required):
                     missing_credential_inputs.append(required)
+                    credential_name = credential.name
 
         if missing_credential_inputs:
             self.job_explanation = '{} cannot start because Credential {} does not provide one or more required fields ({}).'.format(
-                self._meta.verbose_name.title(), credential.name, ', '.join(sorted(missing_credential_inputs))
+                self._meta.verbose_name.title(), credential_name, ', '.join(sorted(missing_credential_inputs))
```
AlanCoding: This still doesn't handle the case where multiple credentials are missing inputs, and I think it would be easier to follow the code if it did. We do `missing_credential_inputs.append(required)`, if you were only going to give the error for _one_ credential, then `missing_credential_inputs = required` would be preferable (more correct). That's still not _best_, and I think the best would be to basically duplicate this message for every credential in the list that is missing inputs.

djyasin: Thank you so much! I will revisit this!

djyasin: @AlanCoding please let me know if there is anything else I should do for this PR!

AlanCoding: I do think this should be indented starting at `if missing_credential_inputs:`, because the message is specific to 1 credential, not all.

djyasin: Thank you! Please let me know if I I accomplished what you meant with this most recent push!

awx/main/models/unified_jobs.py
```diff
@@ -1345,18 +1345,20 @@ def pre_start(self, **kwargs):
...
-            self.save(update_fields=['job_explanation'])
-            return (False, None)
+                    credential_name = credential.name
+
+                if missing_credential_inputs:
+                    self.job_explanation = '{} cannot start because Credential {} does not provide one or more required fields ({}).'.format(
+                        self._meta.verbose_name.title(), credential_name, ', '.join(sorted(missing_credential_inputs))
+                    )
+                self.save(update_fields=['job_explanation'])
+                return (False, None)
```
AlanCoding, line 1359: This needs to be indented because if `missing_credential_inputs` doesn't have anything, we do not want to signal a validation error (which is what the `return` does).

Separately, I don't see a need to define `credential_name`. You can just delete that variable and use `credential.name` because `credential` is in scope as a loop variable here.

john-westcott-iv (approved):
This looks ok after the change @AlanCoding requested was completed.

AlanCoding (approved):

Issue #13055, opened 2022-10-13
awx.awx.tower_export - does't it export multiple assets for one type of asset


ArtsiomMusin:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

This is a question about what tower_export is capable of doing. 

In the doc https://docs.ansible.com/ansible/3/collections/awx/awx/tower_export_module.html#parameters
all parameters are set in plural form. Is that possible to export two or more assets like that:
```
awx.awx.tower_export: 
        projects: project1,project2
        validate_certs: no
```
I see the asset output contains an empty array for projects. So I suppose the export can export multiple assets. Or not?
```
ok: [localhost] => {
    "msg": {
        "assets": {
            "projects": []
        },
        "changed": false,
        "failed": false
    }
}

```

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [X] Collection
- [x] CLI
- [ ] Other

### Installation method

openshift

### Modifications

no

### Steps to reproduce

Try to export multiple assets for projects in CSV format

### Expected results

tower_export can export multiple assets of one type

### Actual results

nothing is exported

sean-m-sullivan:
@jbradberry I think this can be closed, its working and now in the collection proper?

jbradberry:
pinging @john-westcott-iv , since I'm not currently on the Controller team.

john-westcott-iv:
@ArtsiomMusin closing this from the merged PR. If this is still an issue please reach out and we can open it back up.

PR #13168, opened 2022-11-08
Add multiple asset export for awx cli

ArtsiomMusin:
##### SUMMARY

Related #13055.
I'm trying to add a new small feature for ansible collection/awx cli. This is a multiple export of certain assets. 

Why I added that? My scenario is to be able to export some assets at the same time in one run. Before this change, I have to export asset by asset by names. After that combine all asset json files into one common assets file which is used later to import back assets in one run.

This small new feature will reduce some extra work when exporting assets. And also it will sort of sync export and import as you can import all/some assets in one run but you cannot export some assets into one json file.
##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - Collection
 - CLI

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx_collection/plugins/modules/export.py
+++ b/awx_collection/plugins/modules/export.py
@@ -28,51 +28,51 @@ options:
       default: 'False'
     organizations:
       description:
-        - organization name to export
+        - organization names to export
       type: str
     users:
       description:
-        - user name to export
+        - user names to export
       type: str
     teams:
       description:
-        - team name to export
+        - team names to export
       type: str
     credential_types:
       description:
-        - credential type name to export
+        - credential type names to export
       type: str
     credentials:
       description:
-        - credential name to export
+        - credential names to export
       type: str
     execution_environments:
       description:
-        - execution environment name to export
+        - execution environment names to export
       type: str
     notification_templates:
       description:
-        - notification template name to export
+        - notification template names to export
       type: str
     inventory_sources:
       description:
-        - inventory soruce to export
+        - inventory soruces to export
       type: str
     inventory:
       description:
-        - inventory name to export
+        - inventory names to export
       type: str
     projects:
       description:
-        - project name to export
+        - project names to export
       type: str
     job_templates:
       description:
-        - job template name to export
+        - job template names to export
       type: str
     workflow_job_templates:
       description:
-        - workflow name to export
+        - workflow names to export
       type: str
 requirements:
   - "awxkit >= 9.3.0"
@@ -94,6 +94,10 @@ EXAMPLES = '''
   export:
     job_templates: "My Template"
     credential: 'all'
+
+- name: Export two or more inventories with a comma separated list of names
+  export:
+    inventory: "My Inventory1,My Inventory2"
 '''
 
 import logging
--- a/awx_collection/plugins/modules/export.py
+++ b/awx_collection/plugins/modules/export.py
@@ -29,51 +29,51 @@ options:
     organizations:
       description:
         - organization names to export
-      type: str
+      type: list
     users:
       description:
         - user names to export
-      type: str
+      type: list
     teams:
       description:
         - team names to export
-      type: str
+      type: list
     credential_types:
       description:
         - credential type names to export
-      type: str
+      type: list
     credentials:
       description:
         - credential names to export
-      type: str
+      type: list
     execution_environments:
       description:
         - execution environment names to export
-      type: str
+      type: list
     notification_templates:
       description:
         - notification template names to export
-      type: str
+      type: list
     inventory_sources:
       description:
         - inventory soruces to export
-      type: str
+      type: list
     inventory:
       description:
         - inventory names to export
-      type: str
+      type: list
     projects:
       description:
         - project names to export
-      type: str
+      type: list
     job_templates:
       description:
         - job template names to export
-      type: str
+      type: list
     workflow_job_templates:
       description:
         - workflow names to export
-      type: str
+      type: list
 requirements:
   - "awxkit >= 9.3.0"
 notes:
@@ -95,9 +95,9 @@ EXAMPLES = '''
     job_templates: "My Template"
     credential: 'all'
 
-- name: Export two or more inventories with a comma separated list of names
+- name: Export a list of inventories
   export:
-    inventory: "My Inventory1,My Inventory2"
+    inventory: ['My Inventory 1', 'My Inventory 2']
 '''
 
 import logging
@@ -115,24 +115,12 @@ except ImportError:
 def main():
     argument_spec = dict(
         all=dict(type='bool', default=False),
-        credential_types=dict(type='str'),
-        credentials=dict(type='str'),
-        execution_environments=dict(type='str'),
-        inventory=dict(type='str'),
-        inventory_sources=dict(type='str'),
-        job_templates=dict(type='str'),
-        notification_templates=dict(type='str'),
-        organizations=dict(type='str'),
-        projects=dict(type='str'),
-        teams=dict(type='str'),
-        users=dict(type='str'),
-        workflow_job_templates=dict(type='str'),
     )
 
     # We are not going to raise an error here because the __init__ method of ControllerAWXKitModule will do that for us
     if HAS_EXPORTABLE_RESOURCES:
         for resource in EXPORTABLE_RESOURCES:
-            argument_spec[resource] = dict(type='str')
+            argument_spec[resource] = dict(type='list')
 
     module = ControllerAWXKitModule(argument_spec=argument_spec)
 
--- a/awx_collection/plugins/modules/export.py
+++ b/awx_collection/plugins/modules/export.py
@@ -30,50 +30,62 @@ options:
       description:
         - organization names to export
       type: list
+      elements: str
     users:
       description:
         - user names to export
       type: list
+      elements: str
     teams:
       description:
         - team names to export
       type: list
+      elements: str
     credential_types:
       description:
         - credential type names to export
       type: list
+      elements: str
     credentials:
       description:
         - credential names to export
       type: list
+      elements: str
     execution_environments:
       description:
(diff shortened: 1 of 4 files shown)


ArtsiomMusin:
Looking forward to your comments @AlanCoding @jbradberry

jbradberry:
@ArtsiomMusin good catch

awxkit/awxkit/api/pages/api.py
```diff
@@ -218,6 +218,8 @@ def _filtered_list(self, endpoint, value):
             return endpoint.get(id=int(value))
         options = self._cache.get_options(endpoint)
         identifier = next(field for field in options['search_fields'] if field in ('name', 'username', 'hostname'))
+        if len(value.split(',')) > 0:
```
jbradberry: I think doing this here is fragile.

- it doesn't allow for the possibility of the user asking for multiple resources by id instead of name
- it presumes an unparsed value at a very late stage in the processing
- names might have commas in them

What I'm thinking instead is that this method should take a list if multiple items are requested, a string or int if only one.

jbradberry (changes requested):
I see value in fixing this, and I'm a bit surprised to find that I forgot to implement this originally.  However, I'd like to see changes that allow for this functionality to work robustly via multiple paths

- I think the argparse options should be adjusted to take multiple `nargs` (`*` instead of `?`).  The relevant module is `awxkit/awxkit/cli/resource.py`.  This will make it so that a list is always used when using the cli directly
- At the top of the `_filtered_list` method, you should check list vs non-list, then for each of those int vs non-int.  I think the `identifier += '__in'` ought to work for both list of int and list of non-int cases.
- Check to see if the collection module will accept a yaml list as well as a string for these parameters after the change to `_filtered_list`.  @john-westcott-iv, is there any changes that would need to be done there to support this?

jbradberry:
We might want to have logic that turns a list of one item back into just the one item by itself.

awxkit/awxkit/api/pages/api.py
john-westcott-iv: +1 to this

awx_collection/plugins/modules/export.py
```diff
@@ -111,24 +115,12 @@
...
-        projects=dict(type='str'),
-        teams=dict(type='str'),
-        users=dict(type='str'),
-        workflow_job_templates=dict(type='str'),
     )
 
     # We are not going to raise an error here because the __init__ method of ControllerAWXKitModule will do that for us
     if HAS_EXPORTABLE_RESOURCES:
         for resource in EXPORTABLE_RESOURCES:
-            argument_spec[resource] = dict(type='str')
```
ArtsiomMusin, line 131: removed lines above as this line is actually replacing and adding new assert types here

awxkit/awxkit/api/pages/api.py
```diff
@@ -213,11 +213,23 @@ def _export_list(self, endpoint):
...
+        if self._check_for_int(value):
             return endpoint.get(id=int(value))
+
         options = self._cache.get_options(endpoint)
         identifier = next(field for field in options['search_fields'] if field in ('name', 'username', 'hostname'))
+        if isinstance(value, list):
+            if all(self._check_for_int(item) for item in value):
+                identifier = 'or__id'
+            else:
+                identifier = 'or__' + identifier
```
ArtsiomMusin, line 231: I used `or__` here as `__in` no longer worked as `requests` is building query params with a list like that:
```
https://url/api/v2/projects/?name__in=name1&name__in=name2
```
and the result was empty. `or__` works better here I think

awx_collection/plugins/modules/export.py
ArtsiomMusin: it turned out that only this change is required here
Tested successgully on this task:
``` yaml
    - name: Export asserts
      awx.awx.tower_export:
        tower_host: url
        tower_oauthtoken: token
        projects: ['abc1_project_netcool', 'AIC_Project_Rockesh']
        #projects: abc1_project_netcool
        validate_certs: no
      register: data
     ```

File: awxkit/awxkit/api/pages/api.py (hunk 1 above)
Comment (ArtsiomMusin) [reply awxkit/awxkit/api/pages/api.py]: @jbradberry thanks for comments. I was thinking to use a list but ended with a csv string as it was easier to pass it in collection. Anyway I tired to resolve all your comments and left below some explanations for changes

**Comment by ArtsiomMusin** [comment]
after re-requesting for review, @john-westcott-iv was removed for some reason. Still want to have all your reviews :sweat_smile:

File: awx_collection/plugins/modules/export.py [hunk 4]
```diff
@@ -28,52 +28,52 @@
       default: 'False'
     organizations:
       description:
-        - organization name to export
-      type: str
+        - organization names to export
+      type: list
```
jbradberry, line 32: @john-westcott-iv do we have a convention for when a value could be a str or a list?

awx_collection/plugins/modules/export.py
```diff
@@ -111,24 +115,12 @@
 def main():
     argument_spec = dict(
         all=dict(type='bool', default=False),
-        credential_types=dict(type='str'),
```
jbradberry, line 114: @john-westcott-iv ditto for these, it worries me to completely remove parts of something called `argument_spec`.

awxkit/awxkit/api/pages/api.py
jbradberry: Good catch

awx_collection/plugins/modules/export.py
john-westcott-iv: From some testing it seems like ansible is ok with still getting a string here even though we specified it as a list. I was kind of hoping it would be like "I'll make a list with one string element" but that does not seem to be the case. So in the module code we may want to do a check and make it a list of one item if its just a string.

Also, we can add: `elements: str` to all of these to ensure that we are getting strings in the list (see other modules as examples).

awx_collection/plugins/modules/export.py
john-westcott-iv: As long as it shows up in `EXPORTABLE_RESOURCES` in awxkit these are ok to remove here because of lines 129-131.

awx_collection/plugins/modules/export.py
```diff
@@ -111,24 +115,12 @@
...
-        teams=dict(type='str'),
-        users=dict(type='str'),
-        workflow_job_templates=dict(type='str'),
     )
 
     # We are not going to raise an error here because the __init__ method of ControllerAWXKitModule will do that for us
     if HAS_EXPORTABLE_RESOURCES:
         for resource in EXPORTABLE_RESOURCES:
-            argument_spec[resource] = dict(type='str')
+            argument_spec[resource] = dict(type='list')
```
john-westcott-iv: Again, here I would add `elements='str'` to specify that we expect this list to only contain strings.

john-westcott-iv (changes requested):
Thanks for this great work. In addition to inline comments here are a couple more things.

Also, check out line 142 of the new file. If we got 'all' as the string for one of these elements we used to pass '' into awxkit so that it would export all of those asset types. Is that still valid or do we now need to pass `[]` or something else?

I would also like to add some test cases in `awx_collection/tests/integration/targets/export/tasks/main.yml` to make sure old syntax (just strings) and new syntax (lists) are working as expected.

awx_collection/plugins/modules/export.py
ArtsiomMusin: sure will add that

awx_collection/plugins/modules/export.py
ArtsiomMusin: I did two tests like that:

1. List
```yaml
     name: Export asserts
      awx.awx.tower_export:
        tower_host: url
        tower_oauthtoken: token
        projects: ['abc1_project_netcool', 'AIC_Project_Rockesh']
        validate_certs: no
      register: data
```
3. String
```yaml    
    name: Export asserts
      awx.awx.tower_export:
        tower_host: url
        tower_oauthtoken: token
        projects: abc1_project_netcool
        validate_certs: no
      register: data
 ```
In both cases projects are exported fine. So the list type can handle strings and lists without any issues it seems. Do we really need to create a list of one string? Later a sting or list will be handled correctly in `_filtered_list`

jbradberry:
> Also, check out line 142 of the new file. If we got 'all' as the string for one of these elements we used to pass '' into awxkit so that it would export all of those asset types. Is that still valid or do we now need to pass [] or something else?

@john-westcott-iv the way I asked @ArtsiomMusin to change the `_filter_list` method, it can now take both a single string/int _or_ a list of those.  It ought to be fine doing either `''` or `[]` to mean "export all", either way.

jbradberry (approved):
Pending @john-westcott-iv 's final approval.

john-westcott-iv:
@ArtsiomMusin if you're interested in chatting about any of my comments please feel free to reach out to us on the Matrix AWX channel https://matrix.to/#/#awx:ansible.com

ArtsiomMusin:
@john-westcott-iv updated integration tests for export. Here if you want to see is the results https://gist.github.com/ArtsiomMusin/6aac599a29be933f84ab8bc49f7e7969

For my test I commented out export all things, explained why in matrix :smiley:

awx_collection/plugins/modules/export.py
john-westcott-iv: Looks like we just need to add the `elements: str` to the docs section for all of the list types to coincide with the argument spec and then we should be good to go.

john-westcott-iv (approved):

awx_collection/plugins/modules/export.py
ArtsiomMusin: @jbradberry thanks for adding this small change instead of me 😃

PR #13177, opened 2022-11-09
Modify the update.sh script that manages dependencies

john-westcott-iv:
##### SUMMARY

Modify the readme file in requirements for usage of the updater.sh script.
Modify the update.sh script to be more explicit about where its run (inside the container) and how its run (don't just run the script unless "run" parameter is passed and also check for /awx_devel as a simple "am I in the container" check.

##### ISSUE TYPE

##### COMPONENT NAME

 - Other

##### AWX VERSION

```
awx: 21.8.1.dev26+gfe1b37afaf.d20221109
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/requirements/README.md
+++ b/requirements/README.md
@@ -1,25 +1,22 @@
 # Dependency Management
 
-The `requirements.txt` file is generated from `requirements.in`, using `pip-tools` `pip-compile`.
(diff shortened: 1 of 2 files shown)


AlanCoding (approved):

djyasin (approved):
I was able to check this out and run it with no issues!

jbradberry (approved):

PR #13180, opened 2022-11-09
initial commit for (hashcorp) vault machinery in docker stack

sarabrajsingh:
##### SUMMARY

Adding a Hashicorp Vault container to the docker-compose stack. Aiming to help other developers quickly debug Hashicorp Vault credential types.

Bring up a vault instance by issuing the following command from the root of the repository:
```bash
VAULT=true COMPOSE_TAG=<something> make docker-compose
```

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - Other

##### AWX VERSION

```
awx: 21.6.1.dev305+g02b7147803
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,8 @@ SPLUNK ?= false
 PROMETHEUS ?= false
 # If set to true docker-compose will also start a grafana instance
 GRAFANA ?= false
+# If set to true docker-compose will also start a hashicorp vault instance
+VAULT ?= false
 
 VENV_BASE ?= /var/lib/awx/venv
 
@@ -477,7 +479,8 @@ docker-compose-sources: .git/hooks/pre-commit
 	    -e enable_ldap=$(LDAP) \
 	    -e enable_splunk=$(SPLUNK) \
 	    -e enable_prometheus=$(PROMETHEUS) \
-	    -e enable_grafana=$(GRAFANA) $(EXTRA_SOURCES_ANSIBLE_OPTS)
+	    -e enable_grafana=$(GRAFANA) $(EXTRA_SOURCES_ANSIBLE_OPTS) \
+	    -e enable_vault=$(VAULT)
 
 
 
--- a/tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
+++ b/tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
@@ -220,6 +220,19 @@ services:
     privileged: true
   {% endfor %}
 {% endif %}
+{% if enable_vault|bool %}
+  vault:
+    image: vault:latest
+    container_name: tools_vault_1
+    hostname: vault
+    ports:
+      - "1234:1234"
+    environment:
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -477,7 +479,8 @@ docker-compose-sources: .git/hooks/pre-commit
 	    -e enable_ldap=$(LDAP) \
 	    -e enable_splunk=$(SPLUNK) \
 	    -e enable_prometheus=$(PROMETHEUS) \
-	    -e enable_grafana=$(GRAFANA) $(EXTRA_SOURCES_ANSIBLE_OPTS)
+	    -e enable_grafana=$(GRAFANA) $(EXTRA_SOURCES_ANSIBLE_OPTS) \
+		-e enable_vault=$(VAULT)
```
shanemcd: Looks like this should be 1 tab followed by 2 spaces.

tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
```diff
@@ -220,6 +220,19 @@ services:
...
 {% endif %}
+{% if enable_vault|bool %}
+  vault:
+    image: vault:latest
+    container_name: tools_vault_1
+    hostname: vault
+    ports:
+      - "1234:1234"
+    environment:
+      VAULT_DEV_ROOT_TOKEN_ID: "myroot"
```
shanemcd: Hardcoded credentials should be avoided, even in the dev env. We have a pattern for this:

https://github.com/ansible/awx/blob/1f939aa25e4ca32fa8a37a469ac929fc73ff9be3/tools/docker-compose/ansible/roles/sources/tasks/main.yml#L12-L37

sarabrajsingh: @shanemcd - i threw the vault password generation into it's own file, essentially copying alot of the existing machinery around secrets generation. I choose to do it this way because we don't want to generate a `vault_password` everytime `docker-compose` is being called and we're not leveraging `vault`. 

if you don't like this pattern, then i can change the logic to just extending the list in the code snippet you have above. i have no strong opinions about the best path forward.

shanemcd (approved):

AlanCoding:
It's a bit of a shame we didn't merge this earlier, but there has been flux around this stuff lately (name changes and such).

https://hub.docker.com/_/vault/

> Upcoming in Vault 1.14, we will stop publishing official Dockerhub images and publish only our Verified Publisher images. Users of Docker images should pull from [hashicorp/vault](https://hub.docker.com/r/hashicorp/vault) instead of [vault](https://hub.docker.com/_/vault). Verified Publisher images can be found at https://hub.docker.com/r/hashicorp/vault.

I do think this is a valid contribution, but in addition to conflicts, we need to see if the same approach will work for the current image which is:

https://hub.docker.com/r/hashicorp/vault

AlanCoding:
Moved to a new PR

PR #13181, opened 2022-11-09
Replace the querysets provided by django-qsstats-magic

jbradberry:
##### SUMMARY
Replace the querysets provided by django-qsstats-magic with ones that are constructed to do the same thing, without any extraneous functionality.

##### ISSUE TYPE

##### COMPONENT NAME
 - API


jbradberry:
The relevant logic in qsstats is here: https://github.com/PetrDlouhy/django-qsstats-magic/blob/master/qsstats/__init__.py#L87-L134

shanemcd (approved):

Issue #13036, opened 2022-10-11
Collection awx.awx.tower_import is slow


ArtsiomMusin:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

tower_import takes 6 minutes to import any object. Since there might be a lot of dependencies to import at one run, the entire job execution time may take hours to complete just for one organizations.

Attaching the screenshot below where it's seen one task takes 6 mins to complete: 
![image](https://user-images.githubusercontent.com/6213508/195107931-2eb2cea8-e1ee-4268-9f6f-b3c3904d68dc.png)
 

### AWX version

ansible tower 3.8.6

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [X] Collection
- [ ] CLI
- [ ] Other

### Installation method

openshift

### Modifications

no

### Web browser

Chrome

### Steps to reproduce

Just simple tower_import for organization or any other object

### Expected results

Some time as tower_export

### Actual results

6 mins for one tower_import

sarabrajsingh:
hi @ArtsiomMusin,

how big is your import? how many objects are you trying to import?

seem like you're running tower. there is a LDAP issue in `tower` if you're using basic authentication which causes an excessive amount of database traffic, which could be slowing down your import. 

please let us know,

AWX Team

ArtsiomMusin:
Hi @sarabrajsingh 

I'm trying to import one object per one tower_import task:
```yml
- name: import org
  awx.awx.tower_import:
    assets: "{{ lookup('file', file_name) | from_json() }}"
    tower_host: "{{ tower_url }}"
    tower_oauthtoken: "{{ lookup('env', 'TOWER_OAUTH_TOKEN') }}"
    validate_certs: no
```

The file content is:
```json
{"organizations": [{"name": "xxx", "description": "", "max_hosts": 0, "custom_virtualenv": null, "related": {"notification_templates": [{"organization": {"name": "xxx", "type": "organization"}, "name": "sid", "type": "notification_template"}], "notification_templates_started": [], "notification_templates_success": [], "notification_templates_error": [], "notification_templates_approvals": []}, "natural_key": {"name": "xxx", "type": "organization"}}]}
```

[...]

PR #13182, opened 2022-11-10
Improve performance for awx cli export 

ArtsiomMusin:
##### SUMMARY

Slightly related to #13036

I'd like to improve performance for awx cli export.

My scenario is to export 150 inventories with 50 hosts each inventory(7500 hosts in total). I discovered it takes 3 hours to do that and even 20 hours on another system which has slower connection to awx.

I noticed a performance bottleneck here:
https://github.com/ansible/awx/blob/4a30cc244f7c9413704e6d3af2a81447c52ab0ac/awxkit/awxkit/api/pages/api.py#L164

the issue is the endpoint value is always repeating here for inventories and hosts many time e.g.
```
/api/v2/inventory/123/hosts
/api/v2/hosts/1/groups
/api/v2/hosts/2/groups
...
```
The page cache is cashing by the full endpoint path and post fields are taken each time since a new endpoint is coming. But post fields are actually sort of constant things. So if I get posts fields from /api/v2/hosts/1/groups and /api/v2/hosts/2/groups, the  fields are the same. This leads to extra amount of options calls and we can really call here just once. 

I propose to add subcache for post fields options. This subcache is separate from the main page cache and it stores post fields just once by short key values.

Test results:
1. My initial tests with 150 invs and 50 hosts in each inv. The time reduced from 3 hours -> 1.5 hours. 
2. Tested on less amount of objects(2-3 invs, 10-20 JTs, 20 users, etc). Banchmarked with cProfile and the improvement is around 15-30%.
The more similar objects are exported the more perf improvement.

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - CLI

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```


ArtsiomMusin:
@jbradberry please take a look

jbradberry:
Ok, but if I don't get to it by tomorrow it'll be a couple of weeks.

jbradberry (changes requested):
I like the idea, but think that keying on the last segment of the url is too fragile.  The `URLRegistry` uses a table of regex patterns to turn urls into `Page` subclasses; I think that you should use the matching regex from the registry as the key here if you can figure out a nice way to do that.

Also, I'd prefer to see this regex-based table be of the `page.options()` results instead of the processed `post_fields`, if that can be done without getting stuck with a bad options object in the cache.

ArtsiomMusin:
> I like the idea, but think that keying on the last segment of the url is too fragile. The `URLRegistry` uses a table of regex patterns to turn urls into `Page` subclasses; I think that you should use the matching regex from the registry as the key here if you can figure out a nice way to do that.

ok I will take a look at those regex and how to use them.


> Also, I'd prefer to see this regex-based table be of the `page.options()` results instead of the processed `post_fields`, if that can be done without getting stuck with a bad options object in the cache.

I'm not sure why we need to store raw options instead of processed ones as this is reducing one more step - transforming options to post field each time we get options from this new cache. Any pros for storing  `page.options()` ?

jbradberry:
> I'm not sure why we need to store raw options instead of processed ones as this is reducing one more step - transforming options to post field each time we get options from this new cache. Any pros for storing `page.options()` ?

I'm not super attached to the idea, but my thinking on this is always that it is expensive doing the option request in the first place to populate the cache, so it's best to make the maximum use of that object instead of throwing it away in favor of a calculation result based on it.

ArtsiomMusin:
Back from vacation and ready to complete this one. 

I added just a few small changes here. I tried to use some regex matching strings from `URLRegistry` but that didn't work well here. Those regex strings look like:
```
re.compile('^/api/v2/hosts/17190/groups/(\\?.*)*$') for /api/v2/hosts/17190/groups/
re.compile('^/api/v2/hosts/17191/groups/(\\?.*)*$') for /api/v2/hosts/17191/groups/
re.compile('^/api/v2/hosts/17192/groups/(\\?.*)*$') 
```
so not good as keys. For that reaason I used an entire page object fetched based on url. I used `get_registered_page` from  `URLRegistry`.

@jbradberry regarding your comment: `my thinking on this is always that it is expensive doing the option request in the first place to populate the cache`. Sounds like one more improvement. :) Initially I just wanted to reduce an amount of options calls. But you are going further with even avoiding to call options and try to get those post fields from the object somehow... if I got you right... if it's possible, I can try to do that but I don't see any way of doing that as it seems options should be fetch at least once and first time... Let me know if I got it right or wrong

jbradberry:
@ArtsiomMusin `re.compile` objects have the pattern string on them as the attribute `.pattern`, which would hopefully work well as the key for this use.

```
In [1]: import re

In [2]: RE = re.compile('^([a-z0-9][-a-z0-9]{0,62}[.])*[a-z0-9][-a-z0-9]{1,}$')

In [3]: RE.pattern
Out[3]: '^([a-z0-9][-a-z0-9]{0,62}[.])*[a-z0-9][-a-z0-9]{1,}$'
```

jbradberry:
> with even avoiding to call options and try to get those post fields from the object somehow

No, that isn't possible.  Not every field on the `GET` object is going to be postable, and sometimes fields are going to be missing (even though I can't think of examples at the moment).

The options do need to be fetched at least once, and I believe every fetch of the options ought to be cached.

ArtsiomMusin:
> @ArtsiomMusin `re.compile` objects have the pattern string on them as the attribute `.pattern`, which would hopefully work well as the key for this use.
> 
> ```
> In [1]: import re
> 
> In [2]: RE = re.compile('^([a-z0-9][-a-z0-9]{0,62}[.])*[a-z0-9][-a-z0-9]{1,}$')
> 
> In [3]: RE.pattern
> Out[3]: '^([a-z0-9][-a-z0-9]{0,62}[.])*[a-z0-9][-a-z0-9]{1,}$'
> ```

ok I see. I've inspected the store of `URLRegistry` and what I see is there are multiple patterns for one single resource:
```
re.compile('^/api/v2/users/\\d+/(\\?.*)*$'): {
    re.compile('.*'): <class 'awxkit.api.pages.users.User'>
}, 
re.compile('^/api/v2/users/(\\?.*)*$'): {
    re.compile('post'): <class 'awxkit.api.pages.users.User'>, 
    re.compile('.*'): <class 'awxkit.api.pages.users.Users'>
}, 
re.compile('^/api/v2/organizations/\\d+/admins/(\\?.*)*$'): {
    re.compile('.*'): <class 'awxkit.api.pages.users.Users'>
}, 
re.compile('^/api/v2/\\w+/\\d+/users/(\\?.*)*$'): {
    re.compile('.*'): <class 'awxkit.api.pages.users.Users'>
}, 
re.compile('^/api/v2/credentials/\\d+/owner_users/(\\?.*)*$'): {
    re.compile('.*'): <class 'awxkit.api.pages.users.Users'>
}, 
re.compile('^/api/v2/users/\\d+/admin_of_organizations/(\\?.*)*$'): {
    re.compile('.*'): <class 'awxkit.api.pages.users.Users'>
}
```

We can get the pattern string where a post key exists but what if we don't have a post method?
Or we can store all pattern strings as keys but that doesn't sound as an efficient solution because all patterns are poiting to the same resource. 

So need some help to make the final decision...

jbradberry:
@ArtsiomMusin since all we really care about for the keying is what the url path is with the object-specific info washed out, I think the right thing to do is always pick the `.*` action.  The different http verbs in there are only to distinguish which page type to use with the results, usually list (get) vs detail (post) versions on the list endpoints.  Irrelevant for our use here.

jbradberry:
Er, that is to say, I don't think we need to dig into the registry values at all.  The key in the registry is the key we should use in this new cache.

TheRealHaoLiu:
@sean-m-sullivan can you help us review this PR?

sean-m-sullivan:
Initial testing shows no difference in output, and 20% performance improvement on the instances I was able to use it on. I like it!

sean-m-sullivan (approved):

ArtsiomMusin:
Hi guys, please let me know if anything pending from my side. I was trying to reach Jeff in element to discuss some details but looks like he is not there.

jjwatt:
I'm going to test this some more today. It seems fine so far. If there aren't any issues, we should be able to merge it.

jjwatt (approved):
Looks fine

awxkit/awxkit/api/pages/page.py
```diff
@@ -516,6 +517,26 @@ def get_options(self, page):
 
         return self.options.setdefault(url, options)
 
+    def get_post_fields(self, page):
+        url = page.endpoint if isinstance(page, Page) else str(page)
+        key = get_registered_page(url)
+        if key in self.post_fields:
+            return self.post_fields[key]
+
+        options_page = self.get_options(page)
```
jjwatt, line 526: You left out the check on `options_page` that was in the original! I know I already approved this, but can we fix this before merging? See L38 in the original. I'm concerned about possibly removing a null check in this updated code. I'm guessing it wasn't covered in the unit tests, either. Don't worry about that, though. I'll follow-up. Just add the check back in here if it still makes sense. If it doesn't make sense here anymore, let me know why.

jjwatt:
ohh already merged. Nevermind. I'll follow-up with a fix right now.

Issue #13189, opened 2022-11-11
Job traceback breaks job output


keithjgrant:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

If a job has a traceback error, it can break the UI's job events storage & lookup, causing gray loading bars and no visible output, along with an endless stream of network requests

### AWX version

devel

### Select the relevant components

- [X] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

docker development environment

### Modifications

no

### Steps to reproduce

open a job with a `result_traceback`

### Expected results

for traceback to display above job output

### Actual results

Gray loading bars

PR #13196, opened 2022-11-15
Fix job error traceback in job output

keithjgrant:
##### SUMMARY
Fixes the way tracebacks are displayed in job output so it doesn't break tree lookup for events in the event tree.

Addresses #13189 

##### ISSUE TYPE

##### COMPONENT NAME
 - UI


marshmalien (approved):
👍 lgtm

akus062381 (approved):

PR #13197, opened 2022-11-15
Fix /api/swagger endpoint (available only in development mode)

shanemcd:
##### SUMMARY

When trying to view `/api/swagger` in the dev env, you will currently see:

image (7)

https://github.com/marcgibbons/django-rest-swagger is no longer maintained and doesn't work with newer versions of Django.

I deleted code that did something around grouping, but not sure it really matters.

##### ISSUE TYPE

##### COMPONENT NAME

 - API
 - UI

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/api/views/root.py
+++ b/awx/api/views/root.py
@@ -20,6 +20,7 @@ from rest_framework import status
 
 import requests
 
+from awx import MODE
 from awx.api.generics import APIView
 from awx.conf.registry import settings_registry
 from awx.main.analytics import all_collectors
@@ -54,6 +55,8 @@ class ApiRootView(APIView):
         data['custom_logo'] = settings.CUSTOM_LOGO
         data['custom_login_info'] = settings.CUSTOM_LOGIN_INFO
         data['login_redirect_override'] = settings.LOGIN_REDIRECT_OVERRIDE
+        if MODE == 'development':
+            data['swagger'] = drf_reverse('api:schema-swagger-ui')
         return Response(data)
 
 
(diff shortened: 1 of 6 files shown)


shanemcd:
Now looks like this:

image

Alex-Izquierdo:
Hello @shanemcd  I understand that there is no any openapi spec available outside the development environment?

AlexSCorey:
It looks like we have default value of null, even if its a boolean.  It would be great if we can default to `false` on booleans.  It would help us keep out code clean on the UI side.

awx/api/views/root.py
```diff
@@ -54,6 +55,8 @@ def get(self, request, format=None):
         data['custom_logo'] = settings.CUSTOM_LOGO
         data['custom_login_info'] = settings.CUSTOM_LOGIN_INFO
         data['login_redirect_override'] = settings.LOGIN_REDIRECT_OVERRIDE
+        if MODE == 'development':
+            data['swagger'] = drf_reverse('api:schema-swagger-ui')
```
john-westcott-iv, line 59: @shanemcd I added a link from the root view.

john-westcott-iv (approved):
We should look at @AlexSCorey comment but it works.

awx/main/tests/docs/test_swagger_generation.py
```diff
@@ -7,7 +7,7 @@
 from django.utils.functional import Promise
 from django.utils.encoding import force_str
 
-from openapi_codec.encode import generate_swagger_object
+from drf_yasg.codecs import OpenAPICodecJson
```
john-westcott-iv, line 10: @shanemcd @AlanCoding I got the remaining tests working on this PR but was hoping one of you could take a look at how I changed this test to make it work and give a sign off on it.

PR #13200, opened 2022-11-15
Disable work signing by default in dev env

shanemcd:
##### SUMMARY

Certs are generated on the host and there is currently an issue due to openssl version mispatch between Fedora 36 and CentOS Stream 8 which causes:

```
tools_awx_1     | ERROR 2022/11/15 17:09:17 could not load signing key file: unknown block type PRIVATE KEY
tools_awx_1     | ERROR 2022/11/15 17:09:17 could not load signing key file: unknown block type PRIVATE KEY
```

##### ISSUE TYPE

##### COMPONENT NAME

 - Other

### DIFF
--- a/awx/api/templates/instance_install_bundle/group_vars/all.yml
+++ b/awx/api/templates/instance_install_bundle/group_vars/all.yml
@@ -7,7 +7,7 @@ receptor_work_commands:
     command: ansible-runner
     params: worker
     allowruntimeparams: true
-    verifysignature: true
+    verifysignature: {{ sign_work }}
 custom_worksign_public_keyfile: receptor/work-public-key.pem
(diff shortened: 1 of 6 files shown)


AlanCoding (approved):

awx/api/templates/instance_install_bundle/group_vars/all.yml
```diff
@@ -7,7 +7,7 @@ receptor_work_commands:
     command: ansible-runner
     params: worker
     allowruntimeparams: true
-    verifysignature: true
+    verifysignature: {{ sign_work }}
```
relrod, line 10: I think this particular instance of this change might have been unintended. This file isn't related to the dev env and I think this will break without a change to ansible/receptor-collection to go along with it.

PR #13201, opened 2022-11-15
Add management command to allow replaying job event processing

shanemcd:
##### SUMMARY
Added this to help @benthomasson who is trying to measure performance of job event processing.

Usage:

```
$ docker exec -ti tools_awx_1 awx-manage replay_job_events_processing 29
```

##### ISSUE TYPE

##### COMPONENT NAME

 - Other


awx/main/management/commands/replay_job_events_processing.py
```diff
@@ -0,0 +1,53 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument('job_id', type=int)
+
+    def handle(self, *args, **options):
+        runner_callback = RunnerCallback(model=Job)
+        job = Job.objects.get(pk=options['job_id'])
+        runner_callback.instance = job
+        runner_callback.job_created = str(job.created)
+
+        job.job_events.all().delete()
```
shanemcd, line 25: @AlanCoding @fosterseth What's the fastest way to delete job events for a particular job?

AlanCoding:
Pinging @relrod on this, because this is one subset of the debugging abilities we've been talking about planning. There might be some ideas about how to fit this into a larger system.

For a quick initial comment, I think help text should call it out that it would delete stuff in the database before doing that.

PR #13203, opened 2022-11-16
`make clean` docs and maybe refactor

thenets:
##### SUMMARY

I'm confused about the `make-clean` command. It does not actually clean everything in the environment.

This is my attempt to understand what's happening and how to make it obvious how to properly clean the environment.

So, at the end of this PR, I want to be able to:
- [ ] Understand what each clean command actually clean
- [ ] Update comments/descriptions for the clean commands
- [ ] Create a single command that actually cleans up everything
- [ ] Update the `CONTRIBUTING.md` with all the necessary info about that

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - Other (development environment)

##### ADDITIONAL INFORMATION

I don't want to necessarily move all the clean commands to a single section. Maybe the refactoring of one or two make targets would be enough. Maybe additional comments would also be great.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,47 +54,6 @@ I18N_FLAG_FILE = .i18n_built
 	VERSION PYTHON_VERSION docker-compose-sources \
 	.git/hooks/pre-commit
 
-clean-tmp:
-	rm -rf tmp/
-
-clean-venv:
-	rm -rf venv/
-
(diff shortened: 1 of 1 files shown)


thenets:
Too old. Closing it.

Issue #1999, opened 2018-06-19
Smart inventories don't contain groups


freehck:
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
 - API
 - UI

##### SUMMARY
Smart inventory contains now only hosts, but not their groups. This creates problems for roles that use group names.

##### ENVIRONMENT
* AWX version: 1.0.6.11 
* AWX install method: docker on linux
* Ansible version: 2.5.3 
* Operating System: Ubuntu 18.04 LTS

##### STEPS TO REPRODUCE

 - Make an inventory Inv with some number of hosts: Ha, Hb and Hc
 - Make a group Gtop
 - Make some subgroups Gsub1, Gsub2, ... in Gtop
 - Make new smart inventory Smrt with host filter by group Gtop
 - Make a job template, and set Smrt as its inventory
 - Run this job template

##### EXPECTED RESULTS

I expected to see groups Gtop, Gsub1, Gsub2 and so on: all the groups from Inv the filtered hosts in the Smrt are members of.

##### ACTUAL RESULTS

Only hosts are in the Smrt.

wenottingham:
Smart inventories don't have groups by design; this would be a large change to how they're constructed on the backend.

freehck:
@wenottingham so what are Smart Inventories needed for?
I mean, I can't imagine how anyone uses inventory without groups.

Testicule:
What freehck said.

What use are "Smart" Inventories, without groups?
They are useless.

Imagine the outrage if I said: "Oh, hey, I've got a great idea, lets remove groups from normal inventories, and just make them a flat list". People would be appalled. But this is where we are at now with "Smart Inventories".

If it is a large change, to implement, then it's a large change. But it is needed, urgently.

AlanCoding:
We are making some changes to the inventory script generation for job slicing, and I think this is a good opportunity to implement this request as well. The particular solution I have in mind I have discussed with @wwitzel3, and it comes down to a customized version of this query:

https://github.com/ansible/awx/blob/5910b8c5627d46ad52a41801c6998322973c650e/awx/main/models/inventory.py#L248

Specifically designed for smart inventories, so we don't filter on inventory, but instead `host__in=self.hosts`, or something like that.

[...]

Issue #371, opened 2017-10-05
Smart Inventory - Searching on host variables.


wtcross:
##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
- API
 - UI

##### SUMMARY
As an AWX user I should be able to create a smart inventory smart host filter based on host vars associated with a host. The behavior of this feature should be the same as how a user can create a smart host filter based on `ansible_facts`.

##### STEPS TO REPRODUCE

A dictionary called something like `host_vars` should be created that can be searched like this with a smart host filter:

`host_vars.my_cool_hostvar:foo`

##### ACTUAL RESULTS

I would expect this to only result in showing hosts that have a host variable `my_cool_hostvar` set to `foo`.

wtcross:
The motivation of this feature request it to be able to create smart inventories based on variables created via dynamic inventory scripts. Think integration with Puppet DB or the ServiceNow CMDB.

AlanCoding:
I don't see what the shortcoming is.

GET `/api/v2/hosts/?host_filter=variables__icontains=foobar`

When I do this, I see my host that I put the foobar variable into.

https://gist.github.com/AlanCoding/fc54f5557cc1ae19fdbf3582ac7d0de5

chrismeyersfsu:
This would absolutely be an awesome additional; applying the json `host_filter` syntax to host vars.

Three things to consider. 
1. We added a feature in tower like 2.2 to save comments in host vars. So we store them as strings basically. This is easily gotten around by continuing to store them the way we do currently and a new, shadow, json field. 
2. Database indexing size considerations (see below). 
3. Implicit host vars that come from groups and inventory. 
(3a.) differentiating from the want to search  inclusively through the inv+group+host vars or exclusively, only the host vars. 
(3b.) Being able to support the inclusive search.

<hr>

GIN index on 100,000 records w/ 2-4 KB average JSON size (typical ansible gather_facts payload).

Hosts | Table Size | Index Size
-- | -- | --
100,000 | 10 MB | 2.483 GB

wtcross:
@AlanCoding The issue with your example is that it's doing a simple search on a single word. A user will want to search for a specific variable-value combination. A good example is something as simple as the `ec2_state` host variable created by the ec2 inventory source.

`ec2_state: running`

You'd have to search for both the key and value. On top of that I'm not sure how good of an idea it is to allow searching for variables using `icontains` when the value of one variable could contain the target text. A comment could even include the searched text.

[...]

PR #13214, opened 2022-11-18
[demo] API side changes to allow running jobs against multiple inventories

AlanCoding:
##### SUMMARY
Allows running a job template against multiple inventories. Still need:

 - a related runner patch is necessary for this to run - https://github.com/ansible/ansible-runner/pull/1166
 - RBAC, creating one of this inventory type constitutes a delegation of "use_role" of source inventories
 - UI updates to make this work
 - some API work for back-links of stuff
   - show multi-inventory jobs in historical jobs for a constituent inventory
   - show a host list for the "multiple" kind of inventory combining the source inventories
   - some minor adjustments of related links and stuff

This is offered as an option to solve https://github.com/ansible/awx/issues/371 and https://github.com/ansible/awx/issues/1999, since users can simply use host patterns in their playbooks to target the hosts they want.

This is also a counter-point to more complex solutions to https://github.com/ansible/awx/issues/371 that still involve loading the full host set, per discussions with @nitzmahone @sivel and others. If a solution results in Ansible core loading all of the source inventory hosts, then it would appear inferior to this (simple) solution in every way I can think of. This multi-inventory approach is a good fallback if we're not able to implement the other proposals, but I also hope that it helps to cull obvious design dead-ends before we waste too much time on them.

Screenshot from 2022-11-18 14-21-32

This example runs on an execution node, where "alan" and "steve" are hosts from two separate inventories.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API


AlanCoding:
closing to put up another version with a more complete PoC

PR #13215, opened 2022-11-18
add podman newer policy

jangel97:
Hi,
Podman supports another pull policy besides from Always, Missing or Never, it is called "Newer".
Documentation from man:
   --pull=policy
       Pull image policy. The default is missing.

              · always: Always pull the image and throw an error if the pull fails.

              · missing: Pull the image only if it could not be found in the local containers storage.  Throw an  error  if  no
                image could be found and the pull fails.

              · never:  Never  pull  the  image  but use the one from the local containers storage.  Throw an error if no image
                could be found.

              **· newer: Pull if the image on the registry is newer than the one in the local containers storage.   An  image  is
                considered  to  be  newer  when the digests are different.  Comparing the time stamps is prone to errors.  Pull
                errors are suppressed if a local image was found.**
I suggest to add this to dropdown as it looks interesting.

I tested the change on my preprod AAP environment, apparently it is working:
image
image
image

Kind regards,


awx/main/models/execution_environments.py
```diff
@@ -17,6 +17,7 @@ class Meta:
         ('always', _("Always pull container before running.")),
         ('missing', _("Only pull the image if not present before running.")),
         ('never', _("Never pull container before running.")),
+        ('newer', _("Pull if the image on the registry is newer than the one in the local containers storage.")),
```
fosterseth: just to keep the description shorter, maybe this can be

`Only pull if a newer image is available in the registry.`

jangel97: Hey Seth! I just updated it, let me know if the PR looks good to you now!

fosterseth:
@jangel97 since you changed the django model field, you'll need to run `python manage.py makemigrations` and make sure the generated migration file has a reasonable name (e.g. `_add_the_newer_pull_option`)

git add that migration file and push commit

jangel97:
> @jangel97 since you changed the django model field, you'll need to run `python manage.py makemigrations` and make sure the generated migration file has a reasonable name (e.g. `_add_the_newer_pull_option`)
> 
> git add that migration file and push commit

I just added the file, let me know if you notice any issues. I had to run command `awx-manage migrate --noinput` after running the python command, otherwise it would get stuck like this 
image

I did test this in a small AAP instance

Kind regards,

awx/main/migrations/0174_executionenvironment_newer_pull_policy.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2.13 on 2022-12-13 15:48
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('main', '0159_deprecate_inventory_source_UoPU_field'),
```
fosterseth: this should point to the previous migration file
`0173_instancegroup_max_limits`

john-westcott-iv:
@jangel97 Do you happen to know which kubernetes version supports this option? It seems like it would work for our docker compose environment but unless kube supports it we can't merge this into AWX because it could cause issues.

fosterseth:
@jangel97 

EEs can be used for both container group jobs (via k8s cluster) and regular jobs (via podman)

the image pull policy needs to apply _both_ to podman and kubernetes. podman supports "newer", but unfortunately kubernetes does not have an `imagepullpolicy` option of "newer".

We could leave the option in, but update the description to indicate that newer only applies to non container group jobs

@shanemcd if there isn't a clean way to do this, we may need to scrap the idea, even if this is helpful for AAP VM deployments

jangel97:
IMHO I do not think Kubernetes will support Newer pull policy in the short term. Kubernetes supports many CRI container runtimes and most of them do not support Newer pull policy.

Is there a way to change PULL_CHOICES var depending on where AWX is running? For example, this is similar to ansible-runner which can run in different contexts (ansible-navigator and standalone) https://github.com/ansible/ansible-runner/blob/devel/ansible_runner/runner.py#L209 

Is there something similar in AWX so the API behaviour can change during runtime?

PR #13220, opened 2022-11-21
fix: french vocabulary/translation

grimlokason:
SUMMARY

Correction on french vocabulary
ISSUE TYPE

Incorect text in french
COMPONENT NAME

UI


fosterseth:
@rooftopcellist fyi here is a report of mistranslations, we may want to ping the translation team

@grimlokason thank you for taking the time to correct the files, but we have a process to build out our .po files, so we cannot accept PRs that modify them unfortunately.

PR #11439, opened 2021-12-09
Add shellcheck and codespell checks.

joeltenta:
Add shellcheck and codespell checks to the Makefile.
Spelling corrections in doc files.

##### SUMMARY

This PR adds shellcheck and codespell checks to the Makefile.
Double quotes array expansion tools/docker-compose/entrypoint.sh (SC2068).
Spelling corrections in doc files.

##### ISSUE TYPE

 - Docs Pull Request

##### COMPONENT NAME

 - Makefile
 - tools/docker-compose/entrypoint.sh

##### AWX VERSION

```
awx: 19.5.1.dev29+g99e2f7c86b
```

##### ADDITIONAL INFORMATION

```

```

PR #13221, opened 2022-11-21
Spelling corrections in markdown files

djyasin:
Spelling corrections in markdown files.

Co-Authored By: Lila Yasin <89486372+djyasin@users.noreply.github.com>

##### SUMMARY

Adding a few code spelling and spelling corrections. Cherry-picked from a community PR #11439

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - Docs
 - Other

### DIFF
--- a/ISSUES.md
+++ b/ISSUES.md
@@ -31,7 +31,7 @@ If your issue isn't considered high priority, then please be patient as it may t
 
 `state:needs_info` The issue needs more information. This could be more debug output, more specifics out the system such as version information. Any detail that is currently preventing this issue from moving forward. This should be considered a blocked state.
 
-`state:needs_review` The issue/pull request needs to be reviewed by other maintainers and contributors. This is usually used when there is a question out to another maintainer or when a person is less familar with an area of the code base the issue is for.
+`state:needs_review` The issue/pull request needs to be reviewed by other maintainers and contributors. This is usually used when there is a question out to another maintainer or when a person is less familiar with an area of the code base the issue is for.
 
 `state:needs_revision` More commonly used on pull requests, this state represents that there are changes that are being waited on.
 
--- a/awx/ui/CONTRIBUTING.md
+++ b/awx/ui/CONTRIBUTING.md
@@ -159,7 +159,7 @@ Inside these folders, the internal structure is:
 - **/hooks** - Custom react [hooks](https://reactjs.org/docs/hooks-custom.html)
 - **/locales** - [Internationalization](#internationalization) config and source files.
 - **/screens** - Based on the various routes of awx.
-  - **/shared** - Components that are meant to be used specifically by a particular route, but might be sharable across pages of that route. For example, a form component which is used on both add and edit screens.
+  - **/shared** - Components that are meant to be used specifically by a particular route, but might be shareable across pages of that route. For example, a form component which is used on both add and edit screens.
 - **/util** - Stateless helper functions that aren't tied to react.
 
 ### Patterns
--- a/awx/ui/SEARCH.md
+++ b/awx/ui/SEARCH.md
@@ -280,7 +280,7 @@ All of these functions act on the react-router history using the `pushHistorySta
 
 **a note on sort_columns and search_columns**
 
-We have split out column configuration into separate search and sort column array props--these are passed to the search and sort columns. Both accept an isDefault prop for one of the items in the array to be the default option selected when going to the page. Sort column items can pass an isNumeric boolean in order to chnage the iconography of the sort UI element. Search column items can pass type and if applicable choices, in order to configure the right-hand side of the search bar.
+We have split out column configuration into separate search and sort column array props--these are passed to the search and sort columns. Both accept an isDefault prop for one of the items in the array to be the default option selected when going to the page. Sort column items can pass an isNumeric boolean in order to change the iconography of the sort UI element. Search column items can pass type and if applicable choices, in order to configure the right-hand side of the search bar.
 
 ### FilterTags component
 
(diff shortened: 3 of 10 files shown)


Makefile
```diff
@@ -54,6 +54,23 @@ I18N_FLAG_FILE = .i18n_built
 	VERSION PYTHON_VERSION docker-compose-sources \
 	.git/hooks/pre-commit
 
+shellcheck:
```
fosterseth: @shanemcd  does this makefile diff look good to you? if so we'll merge this in soon

tools/docker-compose/entrypoint.sh
```diff
@@ -32,4 +32,4 @@ else
     export SDB_NOTIFY_HOST=$(ip route | head -n1 | awk '{print $3}')
 fi
 
-exec $@
+exec "$@"
```
shanemcd: This change looks suspicious to me and would expect it to cause problems for commands with more than one argument.

shanemcd:
I called this out on the original PR but, right now this seems to rely on someone manually running this command, which doesn't feel particularly useful. It's very trivial to add new CI checks and I think we should do that before merging this PR: https://github.com/ansible/awx/blob/a0b8215c06918260c4f3d0c481fb1adac6fa2613/.github/workflows/ci.yml#L17 The only additional thing you might need to do is ensure the correct package is installed, but only in the _devel_ image. There are other examples in the Dockerfile of us doing this.

shanemcd (changes requested):
Added some comments. The main thing that worries me right now is the change to the entry point. But I also don't see this being very useful without a CI check.

AlexSCorey (approved):
UI portion looks good to me.

Makefile
gundalow: How come we've got more than detect & fix?

Makefile
```diff
@@ -54,6 +54,23 @@ I18N_FLAG_FILE = .i18n_built
 	VERSION PYTHON_VERSION docker-compose-sources \
 	.git/hooks/pre-commit
```
gundalow: We will add these into CI at a later point?

gundalow (commented):
Do you plan on adding shellcheck & codespell into CI?

djyasin:
This PR had a bit of a mix-up. When we had initially discussed pulling these changes in, we decided not to pull in the Makefile changes and not to add to CI. That decision was made because we realized we would A) have to go back and retroactively fix all the existing spelling issues (I worked on that for a bit and realized it was very big endeavor to sift through) and B) it could cause PRs to fail since sometimes the spellchecks aren't 100% correct.

When I cherry-picked this commit initially, I didn't realize I had grabbed those changes as well.

I have now pulled out the makefile changes and am just keeping the spelling suggestions.

Let me know if you feel it should be done a different way- I've had quite a bit of conversation around this one and there are different opinions. I can definitely pivot!

gundalow:
> This PR had a bit of a mix-up. When we had initially discussed pulling these changes in, we decided not to pull in the Makefile changes and not to add to CI. That decision was made because we realized we would A) have to go back and retroactively fix all the existing spelling issues (I worked on that for a bit and realized it was very big endeavor to sift through) and B) it could cause PRs to fail since sometimes the spellchecks aren't 100% correct.
> 
> When I cherry-picked this commit initially, I didn't realize I had grabbed those changes as well.
> 
> I have now pulled out the makefile changes and am just keeping the spelling suggestions.
> 
> Let me know if you feel it should be done a different way- I've had quite a bit of conversation around this one and there are different opinions. I can definitely pivot!

Thanks for the explanation. 
This sounds good, though we may just need to accept things might regress without CI (though this PR still makes things better than they were).

djyasin:
> > This PR had a bit of a mix-up. When we had initially discussed pulling these changes in, we decided not to pull in the Makefile changes and not to add to CI. That decision was made because we realized we would A) have to go back and retroactively fix all the existing spelling issues (I worked on that for a bit and realized it was very big endeavor to sift through) and B) it could cause PRs to fail since sometimes the spellchecks aren't 100% correct.
> > When I cherry-picked this commit initially, I didn't realize I had grabbed those changes as well.
> > I have now pulled out the makefile changes and am just keeping the spelling suggestions.
> > Let me know if you feel it should be done a different way- I've had quite a bit of conversation around this one and there are different opinions. I can definitely pivot!
> 
> Thanks for the explanation. This sounds good, though we may just need to accept things might regress without CI (though this PR still makes things better than they were).

Thank you!! I will keep this in my notes in case we do decide to take the time to fix all of the existing spelling issues and add the CI check.

djyasin:
@shanemcd does this look okay now?

shanemcd:
I think it's good to fix these typos, but I also think it would be worthwhile and relatively trivial to add a CI check that runs the utility mentioned in https://github.com/ansible/awx/pull/11439. 

If we're not going to do that here, please open an issue for running spell checking as part of the PR checks. It should probably get added to https://github.com/ansible/awx/blob/devel/.github/workflows/ci.yml. 

You can see there is a pretty well established pattern there. As long as the dependency is installed in the devel image, it will just be a matter of adding a new item to the test [matrix](https://github.com/ansible/awx/blob/daf39dc77ecb75a97f7509ffa58ab76b18f4e8a9/.github/workflows/ci.yml#L19).

djyasin:
> It should probably get added to https://github.com/ansible/awx/blob/devel/.github/workflows/ci.yml.

Thank you so much for your feedback! I went ahead and made a new issue here: https://github.com/ansible/awx/issues/14039

shanemcd (approved):

Issue #13192, opened 2022-11-14
Tracking: AWX code changes for rsyslog decoupling


relrod:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

Enhancement to Existing Feature

### Feature Summary

For the [rsyslog decoupling proposal](https://github.com/ansible/awx/issues/13183) we will need to make some modifications to AWX.

Largely the changes come down to:

- [ ] Write configwatcher daemon that makes use of `construct_rsyslog_conf_template()` when it sees the pg_notify message mentioned above, to write out a new config and restart rsyslog using supervisord
- [ ] In `awx/main/utils/external_logging.py`, change `reconfigure_rsyslog()` to simply send a pg_notify messaage telling the new configwatcher to do its thing

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [X] Other

### Steps to reproduce

N/A

### Current results

N/A

### Sugested feature result

N/A

jay-steurer:
@CFSNM is working on the test coverage for this task as part of a new feature.

Issue #13183, opened 2022-11-10
Proposal: Decouple Rsyslog from web and task subsystem


TheRealHaoLiu:
### What

This is a proposal for the change in the rsyslog subsystem and its integration with AWX to allow decoupling of rsyslog configuration management from the web and task subsystems.

### Why

We are attempting to decouple the link between web container and task container within AWX. Decoupling the link between web and task containers will enable us to deploy and scale components of AWX in more flexible and resource-efficient ways.

In examining the interaction and connection between the web and task container we notice that the management of the rsyslog subsystem and its configuration is done in a way that requires a lot of shared files between the two containers.

This proposal will focus on discussing the linkage between web and task containers caused by the rsyslog subsystem and how we plan to sever it.

### Context
In the current deployment of AWX, rsyslog is a process that runs in the web container via supervisord. 

When a user changes the settings that are related to rsyslog, the web container commits the settings into the database and uses pg_notify to trigger the task subsystem to execute a task to generate new rsyslog configuration and restart rsyslog in the web container.

This design results in the following coupling:

- supervisord socket of the web container needs to be mounted into the task container.
- rsyslog config directory must be mounted by both the task and web container so that the task container can write to it and rsyslog that’s running in the web container can read from it.
- The rsyslog socket needs to be mounted into both web and task containers so logs can be written to rsyslog.

When deployed on Kubernetes, this coupling and sharing of files is done via `emptyDir` volumes. `emptyDir`s cannot be shared between containers in different deployments (pods) and therefore prevent task and web container to be deployed as separate deployments.

The following diagram shows the containers and `emptyDir` volumes within the AWX deployment with the rsyslog related `emptyDir` volumes highlighted in red.

<img width="681" alt="image" src="https://user-images.githubusercontent.com/44379968/201187736-2e8ae61d-e2d8-4aef-83d1-67c19d2eab9f.png">

#### Goals
- Eliminate the filesystem base coupling between web and task container caused by rsyslog
- Does not dramatically complicate VM/container base deployment

#### Non-Goals
- Couplings between web and task containers caused by other shared systems, such as redis. (This will be discussed in another follow up proposal)

#### Phase 1: rsyslog sidecar before splitting web and task
In order to eliminate the complicated cross interaction described in the context section we are proposing to create a new `configwatcher` process 

Responsibility of the rsyslog `configwatcher` process: 
- consumes pg_notify on configuration change
- handles the generation of rsyslog config
- reloads rsyslog when needed

[...]

PR #13222, opened 2022-11-21
AWX code changes for rsyslog decoupling

jessicamack:
##### SUMMARY

Work for #13192 

refs #13183 

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - Other

##### AWX VERSION

```
21.7.1
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/conf/views.py
+++ b/awx/conf/views.py
@@ -142,8 +142,7 @@ class SettingSingletonDetail(RetrieveUpdateDestroyAPIView):
             connection.on_commit(lambda: clear_setting_cache.delay(settings_change_list))
             if any([setting.startswith('LOG_AGGREGATOR') for setting in settings_change_list]):
                 # call notify to rsyslog. no data is need so payload is empty
-                with pg_bus_conn() as conn:
-                    conn.notify('rsyslog_configurer', "")
+                send_pg_notify('rsyslog_configurer', "")
 
         # When TOWER_URL_BASE is deleted from the API, reset it to the hostname
         # used to make the request as a default.
--- a/awx/conf/views.py
+++ b/awx/conf/views.py
@@ -26,10 +26,11 @@ from awx.api.generics import APIView, GenericAPIView, ListAPIView, RetrieveUpdat
 from awx.api.permissions import IsSystemAdminOrAuditor
 from awx.api.versioning import reverse
 from awx.main.utils import camelcase_to_underscore
-from awx.main.tasks.system import handle_setting_changes
+from awx.main.tasks.system import clear_setting_cache
 from awx.conf.models import Setting
 from awx.conf.serializers import SettingCategorySerializer, SettingSingletonSerializer
 from awx.conf import settings_registry
+from awx.main.dispatch import pg_bus_conn
 
 
 SettingCategory = collections.namedtuple('SettingCategory', ('url', 'slug', 'name'))
@@ -120,7 +121,11 @@ class SettingSingletonDetail(RetrieveUpdateDestroyAPIView):
                 setting.save(update_fields=['value'])
                 settings_change_list.append(key)
         if settings_change_list:
-            connection.on_commit(lambda: handle_setting_changes.delay(settings_change_list))
+            connection.on_commit(lambda: clear_setting_cache.delay(settings_change_list))
+            if any([setting.startswith('LOG_AGGREGATOR') for setting in settings_change_list]):
+                # call notify to rsyslog. no data is need so payload is empty
+                with pg_bus_conn() as conn:
+                    conn.notify('rsyslog_configurer', "")
 
     def destroy(self, request, *args, **kwargs):
         instance = self.get_object()
@@ -135,7 +140,7 @@ class SettingSingletonDetail(RetrieveUpdateDestroyAPIView):
             setting.delete()
             settings_change_list.append(setting.key)
         if settings_change_list:
-            connection.on_commit(lambda: handle_setting_changes.delay(settings_change_list))
+            connection.on_commit(lambda: clear_setting_cache.delay(settings_change_list))
 
         # When TOWER_URL_BASE is deleted from the API, reset it to the hostname
         # used to make the request as a default.
--- a/awx/conf/views.py
+++ b/awx/conf/views.py
@@ -141,6 +141,10 @@ class SettingSingletonDetail(RetrieveUpdateDestroyAPIView):
             settings_change_list.append(setting.key)
         if settings_change_list:
             connection.on_commit(lambda: clear_setting_cache.delay(settings_change_list))
+            if any([setting.startswith('LOG_AGGREGATOR') for setting in settings_change_list]):
+                # call notify to rsyslog. no data is need so payload is empty
+                with pg_bus_conn() as conn:
+                    conn.notify('rsyslog_configurer', "")
 
         # When TOWER_URL_BASE is deleted from the API, reset it to the hostname
         # used to make the request as a default.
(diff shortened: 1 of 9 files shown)


awx/main/tasks/system.py
```diff
@@ -115,7 +114,9 @@ def dispatch_startup():
     m.reset_values()
 
     # Update Tower's rsyslog.conf file based on loggins settings in the db
-    reconfigure_rsyslog()
+    # Paylod is empty string because the configurer already has the data
+    with pg_bus_conn() as conn:
+        conn.notify('rsyslog_configurer', "")
```
AlanCoding: Do we even need to change anything about this call? This is the startup hook for the dispatcher. I see no reason to send it to a separate service in order to run it. Doing so creates new problems, like if the dispatcher starts up faster than the rsyslog configurer, then it might be listening and the message could get dropped.

awx/main/tasks/system.py
```diff
@@ -254,7 +255,9 @@ def handle_setting_changes(setting_keys):
     cache.delete_many(cache_keys)
 
     if any([setting.startswith('LOG_AGGREGATOR') for setting in setting_keys]):
-        reconfigure_rsyslog()
+        # Paylod is empty string because the configurer already has the data
+        with pg_bus_conn() as conn:
+            conn.notify('rsyslog_configurer', "")
```
AlanCoding: By the nature of pg_notify, this is a broadcast-all kind of thing, because you have all nodes listening to the same "rsyslog_configurer" channel. So it's weird to broadcast the message inside of `handle_setting_changes`, which is already a broadcast message. Run with multiple nodes in the control plane and see how many times the method gets called for each node.

I think you'll find that this is better done from the submitter of this task.

https://github.com/ansible/awx/blob/ac6a82eee41feb041ff3e4d16459d4b1a774175f/awx/conf/views.py#L123

There, what we should want is logic to say "broadcast message to clear settings cache (`handle_setting_changes`) and then broadcast message to restart rsyslog"

awx/main/management/commands/run_rsyslog_configurer.py
```diff
@@ -0,0 +1,29 @@
...
+
+    help = 'Launch the rsyslog_configurer daemon'
+
+    def handle(self, *arg, **options):
+        try:
+            with pg_bus_conn(new_connection=True) as conn:
+                conn.listen("rsyslog_configurer")
+                for e in conn.events(yield_timeouts=True):
+                    if e is not None:
+                        reconfigure_rsyslog()
```
AlanCoding, line 34: Before you call this, you should clear all local caches for settings. Otherwise we could face some race conditions.

awx/main/tasks/system.py
TheRealHaoLiu: at the end of the day rsyslog wont be running in the same container as dispatcher so dispatcher wont have access to the rsyslog config file or the supervisord to restart rsyslog

running reconfigure_rsyslog in dispatcher will not work

awx/main/management/commands/run_rsyslog_configurer.py
TheRealHaoLiu: what is the local cache that ur referring to here?

awx/main/tasks/system.py
AlanCoding: true, I need to revise this comment. We should continue the existing practice of reconfiguring rsyslog on startup. But this should not be done by the dispatcher sending a message to the "rsyslog_configurer" channel. It can be done without race conditions by simply reconfiguring rsyslog on startup on the new service added here - the run_rsyslog_configurer service. So I'm now requesting that former L118 `reconfigure_rsyslog()` be moved from the dispatcher service startup to the new service startup (as opposed to task-ifying it)

awx/conf/views.py
```diff
@@ -135,7 +140,7 @@ def perform_destroy(self, instance):
             setting.delete()
             settings_change_list.append(setting.key)
         if settings_change_list:
-            connection.on_commit(lambda: handle_setting_changes.delay(settings_change_list))
+            connection.on_commit(lambda: clear_setting_cache.delay(settings_change_list))
```
AlanCoding, line 142: I think you should still broadcast the rsyslog_configer message here too, because a DELETE removes settings, and can functionally turn external logging off.

awx/main/management/commands/run_rsyslog_configurer.py
AlanCoding: it's implemented now, `cache` and `settings._awx_conf_memoizedcache`

AlanCoding:
:+1: this looks complete to me, in terms of getting the service running and connected correctly.

I have 1 outstanding fairly small comment about settings DELETE.

Before merging, we need corresponding installer changes, and in awx-operator. If those are not changed to add the new configwatcher service, then it will be impossible to change the external logging.

AlanCoding (approved):

awx/conf/views.py
```diff
@@ -120,7 +121,11 @@ def perform_update(self, serializer):
                 setting.save(update_fields=['value'])
                 settings_change_list.append(key)
         if settings_change_list:
-            connection.on_commit(lambda: handle_setting_changes.delay(settings_change_list))
+            connection.on_commit(lambda: clear_setting_cache.delay(settings_change_list))
+            if any([setting.startswith('LOG_AGGREGATOR') for setting in settings_change_list]):
```
relrod, line 125: I know this was just copied and modified from the original code in system.py, but is there any reason to build up a new list here and not use `any` with a generator? Like `any(setting.startswith('LOG_AGGREGATOR') for setting in settings_change_list)`

I guess the list of changed settings will be tiny enough in all cases that it doesn't matter much, though.

jessicamack: I don't think so. ```any``` with a generator gets us to the same result. It's only a list of 14 settings when I test and I don't think the number of settings would reach a size where it would have much impact.

awx/main/management/commands/run_rsyslog_configurer.py
```diff
@@ -0,0 +1,38 @@
...
+        try:
+            with pg_bus_conn(new_connection=True) as conn:
+                conn.listen("rsyslog_configurer")
+                # reconfigure rsyslog on start up
+                reconfigure_rsyslog()
+                for e in conn.events(yield_timeouts=True):
+                    if e is not None:
+                        logger.info("Change in logging settings found. Restarting rsyslogd")
+                        # clear the cache of relevant settings then restart
+                        setting_keys = [k for k in dir(settings) if k.startswith('LOG_AGGREAGTOR')]
```
fosterseth: spelling error,  `LOG_AGGREGATOR`

awx/conf/views.py
```diff
@@ -135,7 +139,11 @@ def perform_destroy(self, instance):
             setting.delete()
             settings_change_list.append(setting.key)
         if settings_change_list:
-            connection.on_commit(lambda: handle_setting_changes.delay(settings_change_list))
+            connection.on_commit(lambda: clear_setting_cache.delay(settings_change_list))
+            if any([setting.startswith('LOG_AGGREGATOR') for setting in settings_change_list]):
+                # call notify to rsyslog. no data is need so payload is empty
+                with pg_bus_conn() as conn:
+                    conn.notify('rsyslog_configurer', "")
```
relrod: I wasn't sure if this was missed or left intentionally - should this just call the new function too?

jessicamack: It was missed. Thought I had already changed this over, fixing it now.

TheRealHaoLiu (approved):

PR #13236, opened 2022-11-28
[tech debt] Avoid recursive include of DEFAULT_SETTINGS, sanity test

AlanCoding:
##### SUMMARY
I told @john-westcott-iv this was going to bother me, but I wasn't going to make a PR to change it until I saw this failure.

```
  File "/awx_devel/awx/main/tests/unit/test_settings.py", line 22, in test_default_settings
    raise Exception(f'Default setting for {k} does not match:\nsnapshot: {snapshot_val}\ndefault: {default_val}')
Exception: Default setting for DEFAULTS_SNAPSHOT does not match:
...
```

Wait, what? The snapshot value for `DEFAULTS_SNAPSHOT` doesn't match? That doesn't make any sense. We got that snapshot value _from_ the snapshot dictionary! Did this dict _save a copy of itself... inside itself_?

Yes.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/tests/unit/test_settings.py
+++ b/awx/main/tests/unit/test_settings.py
@@ -24,9 +24,8 @@ def test_default_settings():
     from django.conf import settings
 
     for k in dir(settings):
-        default_val = getattr(settings.default_settings, k, None)
         if k not in settings.DEFAULTS_SNAPSHOT or k in LOCAL_SETTINGS:
             continue
+        default_val = getattr(settings.default_settings, k, None)
         snapshot_val = settings.DEFAULTS_SNAPSHOT[k]
-        if default_val != snapshot_val:
-            raise Exception(f'Setting for {k} does not match shapshot:\nsnapshot: {snapshot_val}\ndefault: {default_val}')
+        assert default_val == snapshot_val, f'Setting for {k} does not match shapshot:\nsnapshot: {snapshot_val}\ndefault: {default_val}'
--- a/awx/main/tests/unit/test_settings.py
+++ b/awx/main/tests/unit/test_settings.py
@@ -1,8 +1,32 @@
 from split_settings.tools import include
 
 
+LOCAL_SETTINGS = (
+    'ALLOWED_HOSTS',
+    'BROADCAST_WEBSOCKET_PORT',
+    'BROADCAST_WEBSOCKET_VERIFY_CERT',
+    'BROADCAST_WEBSOCKET_PROTOCOL',
+    'BROADCAST_WEBSOCKET_SECRET',
+    'DATABASES',
+    'DEBUG',
+    'NAMED_URL_GRAPH',
+)
+
+
 def test_postprocess_auth_basic_enabled():
     locals().update({'__file__': __file__})
 
     include('../../../settings/defaults.py', scope=locals())
     assert 'awx.api.authentication.LoggedBasicAuthentication' in locals()['REST_FRAMEWORK']['DEFAULT_AUTHENTICATION_CLASSES']
+
+
+def test_default_settings():
+    from django.conf import settings
+
+    for k in dir(settings):
+        default_val = getattr(settings.default_settings, k, None)
+        if k not in settings.DEFAULTS_SNAPSHOT or k in LOCAL_SETTINGS:
+            continue
+        snapshot_val = settings.DEFAULTS_SNAPSHOT[k]
+        if default_val != snapshot_val:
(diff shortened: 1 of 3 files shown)


awx/settings/development.py
```diff
@@ -105,8 +92,22 @@
...
+# customizable config files.
+this_module = sys.modules[__name__]
+local_vars = dir(this_module)
+DEFAULTS_SNAPSHOT = {}
+for setting in local_vars:
+    if setting == setting.upper():
+        DEFAULTS_SNAPSHOT[setting] = copy.deepcopy(getattr(this_module, setting))
+
+del local_vars
+del this_module
```
AlanCoding, line 105: I wanted to `del` these because we do use `dir` on the settings module in our settings wrapper:

https://github.com/ansible/awx/blob/devel/awx/conf/settings.py#L474

and I just don't want to pollute that namespace with more variables that should be internal to logic inside of settings. It would be impractical to add an `__ALL__` to this module. Also, the original recursive copy here was because of lack of awareness of what things hanging out in locals(). We avoid copying lowercase vars in the loop, but I prefer to explicitly exclude them from the namespace, as they should not be public.

AlanCoding:
Integration tests are passing, so I'm happy with this myself now.

awx/main/tests/unit/test_settings.py
```diff
@@ -1,8 +1,32 @@
...
     assert 'awx.api.authentication.LoggedBasicAuthentication' in locals()['REST_FRAMEWORK']['DEFAULT_AUTHENTICATION_CLASSES']
+
+
+def test_default_settings():
+    from django.conf import settings
+
+    for k in dir(settings):
+        default_val = getattr(settings.default_settings, k, None)
+        if k not in settings.DEFAULTS_SNAPSHOT or k in LOCAL_SETTINGS:
+            continue
```
john-westcott-iv, line 28: Stylistically, I'd move this above line 27.

awx/main/tests/unit/test_settings.py
```diff
@@ -1,8 +1,32 @@
...
+        snapshot_val = settings.DEFAULTS_SNAPSHOT[k]
+        if default_val != snapshot_val:
+            raise Exception(f'Setting for {k} does not match shapshot:\nsnapshot: {snapshot_val}\ndefault: {default_val}')
```
john-westcott-iv: I think the answer to this is no but instead of raising here could we just emit a warning and set the setting to what we want it to be?

awx/settings/development.py
```diff
@@ -105,8 +92,28 @@
...
+###############################################################################################
+#
+#  Any settings defined after this point will be marked as as a read_only database setting
+#
+################################################################################################
+this_module = sys.modules[__name__]
+local_vars = dir(this_module)
+DEFAULTS_SNAPSHOT = {}  # define after we save local_vars so we do not snapshot the snapshot
+for setting in local_vars:
+    if setting == setting.upper():
```
john-westcott-iv: Why not just `if setting.isupper():`?

awx/settings/development.py
```diff
@@ -105,8 +92,28 @@
...
-if 'sqlite3' not in DATABASES['default']['ENGINE']:  # noqa
-    DATABASES['default'].setdefault('OPTIONS', dict()).setdefault('application_name', f'{CLUSTER_HOST_ID}-{os.getpid()}-{" ".join(sys.argv)}'[:63])  # noqa
+# Store a snapshot of default settings at this point before loading any
+# customizable config files.
+#
+###############################################################################################
+#
+#  Any settings defined after this point will be marked as as a read_only database setting
+#
+################################################################################################
```
john-westcott-iv: The ordering of these comments back to back feels strange. Would it make sense to put the banner after the completion of the snapshot?

AlanCoding: I don't love the wording here, because everything "after" should be the includes of other files. What it really should _mean_ is that no ordinary settings should come after this point. After this, only logic for includes should be done. Then anything defined by the includes will result in read-only settings.

awx/main/tests/unit/test_settings.py
AlanCoding: This is a test - so changed `raise` to `assert` for clarity.

AlanCoding:
@john-westcott-iv I believe I updated based on all your review comments. How does it look now?

john-westcott-iv (approved):

PR #13247, opened 2022-11-29
Fix high severity vulnerabilities.

kialam:
##### SUMMARY

Fix some Dependabot security warnings.
Cut down the number of high security vulnerabilities to `6`.
Note: the one high security vulnerability for the `nth-check` dependency lives within a dev dependency (`react-scripts`) and is unlikely to pose a real threat to our builds. There is also an [open PR](https://github.com/facebook/create-react-app/pull/12172) to address the security concern on their end, it is currently under review.

##### ISSUE TYPE

##### COMPONENT NAME

 - UI


keithjgrant:
We should probably bump these versions in package.json too

kialam:
@keithjgrant Done. I was unable to upgrade `d3` as it's already at the latest version, however the package-lock change should update the dev dep `d3-color `to what we need.

AlexSCorey (approved):

AlexSCorey:
@kialam just need to add some licenses for get the api-test to pass
```license file babel-helper-get-function-arity.txt does not correspond to an existing requirement; it should be removed.
 license file source-map.txt does not correspond to an existing requirement; it should be removed.
 license for requirement babel-helper-string-parser is missing
 license for requirement jridgewell-gen-mapping is missing
 license for requirement jridgewell-resolve-uri is missing
 license for requirement jridgewell-set-array is missing
 license for requirement jridgewell-sourcemap-codec is missing
 license for requirement jridgewell-trace-mapping is missing

keithjgrant (approved):

akus062381 (approved):

PR #13250, opened 2022-11-30
Fix a problem with migration main/0164

jbradberry:
##### SUMMARY
Since the original version of the migration a) invoked the .save() method, and b) involved a model with a custom field that had a post_save handler attached, this migration had a side-effect that caused the codebase's version of the model to be used when the table involved wasn't yet up to date.  This triggers an UndefinedColumn error.

This change works around the problem by making use of queryset .update() methods instead, which should avoid the post_save signal trigger.

##### ISSUE TYPE

##### COMPONENT NAME
 - API


awx/main/migrations/0164_remove_inventorysource_update_on_project_update.py
```diff
@@ -1,24 +1,13 @@
 # Generated by Django 3.2.13 on 2022-06-21 21:29
 
 from django.db import migrations
-import logging
-
-logger = logging.getLogger("awx")
```
john-westcott-iv, line 6: If someone had a large number of inventory sources are there any concerns that the logging might slow down the migration?

awx/main/migrations/0164_remove_inventorysource_update_on_project_update.py
```diff
@@ -1,24 +1,13 @@
...
-            src.update_on_launch = True
-            src.save(update_fields=['update_on_launch'])
-            logger.info(f"Setting update_on_launch to True for {src}")
-        proj = src.source_project
-        if proj and proj.scm_update_on_launch is False:
-            proj.scm_update_on_launch = True
-            proj.save(update_fields=['scm_update_on_launch'])
-            logger.warning(f"Setting scm_update_on_launch to True for {proj}")
+    sources.update(update_on_launch=True)
+    sources.filter(source_project__isnull=False).update(scm_update_on_launch=True)
```
AlanCoding: For my purposes of reading this, let me sub in the prior definition of the `sources` QuerySet

```python
InventorySource.objects.filter(update_on_project_update=True).update(scm_update_on_launch=True)
```

This is attempting to replace `proj.save(update_fields=['scm_update_on_launch'])`... and put in those terms the problem becomes kind of obvious. This was making a change to _projects_, but you're using an InventorySource QuerySet. I think you need to start the QuerySet over, starting from `Project.objects`

awx/main/migrations/0164_remove_inventorysource_update_on_project_update.py
jbradberry: Possibly, though my main concerns with removing it are,

1. our loggers make use of the database, except under circumstances that I don't entirely trust
2. migrations are usually run unattended, so it feels like this is just going to be spamming the logs

awx/main/migrations/0164_remove_inventorysource_update_on_project_update.py
jbradberry: Oops.  Good catch.

awx/main/migrations/0164_remove_inventorysource_update_on_project_update.py
jbradberry: And of course the obvious: if we are attempting to avoid the side-effects involved with executing `.save()` on each object, we no longer really have an opportunity to individually log each object.

awx/main/migrations/0164_remove_inventorysource_update_on_project_update.py
jbradberry: @AlanCoding ok, I think I have it correct now, let me know.

awx/main/migrations/0164_remove_inventorysource_update_on_project_update.py
AlanCoding: I thought we were pretty confident that migrations use the "default" (meaning file-based) settings. This might not respect external logging settings, but it's something better than nothing.

you can still log the return value of bulk updates.

```python
In [2]: Project.objects.filter(name='foobar').update(scm_update_on_launch=True)
Out[2]: 0
```

I have done this in several migrations before. I value the feedback logging gives, but this is more of a style topic than anything else, and I wouldn't ever say it's a blocker.

AlanCoding (approved):

AlanCoding:
> and b) involved a model with a custom field that had a post_save handler attached

Could you post some more details about this? I would like to know what custom field it is, and if anything in that logic might be something we can clean up.

jbradberry:
@AlanCoding it's the `ImplicitRoleField`.

https://github.com/ansible/awx/blob/devel/awx/main/fields.py#L220
https://github.com/ansible/awx/blob/devel/awx/main/fields.py#L283-L314

Line 289 is the one that broke in the logs I looked over.

Normally migrations will not trigger signal handlers, but in this case the signal handler is intimately connected to a custom field which has to be used in the migration model.

jbradberry:
I'd definitely be in favor of decoupling the signal handler from the field definition.

fosterseth (approved):

CFSNM (approved):

AlanCoding:
Starting from Line 289, if a simple `.objects.get(id=pk)` errors due to a missing field, then we must question the version of the `Model` we are using. It gets that from an app state returned from `get_current_apps()`, and in this case "current" _should_ mean apps from the migration file that's running.

But how would it know what the apps are in the current migration? Of course it doesn't. From memory and the looks of things, I believe this is relying on an undocumented and archaic practice that the migration code calls `set_current_apps()` before doing anything related to the RBAC system.

As cringe-worthy as this may be, you can find lots of examples of this system in use.

https://github.com/ansible/awx/blob/ac6a82eee41feb041ff3e4d16459d4b1a774175f/awx/main/migrations/_inventory_source.py#L83

Because we deprecated and deleted a field from a major RBAC model (InventorySource) we may currently be at very high risk of seeing similar migration errors, but this isn't as bad as it seems, if throwing around more calls to `set_current_apps` is a truly durable solution, and I think it probably is.

jbradberry:
@AlanCoding unfortunately, that is not how it works.  If business logic in a custom field does its own query based on imports from our models, you have broken out of the migration sandbox and are making use of the direct model from our side of the codebase.  Django does no monkey patching during migrations that I am aware of, you are just expected to make sure you always use the `apps.get_model()` method to obtain a clean model class to do queries with.  No amount of sprinkling in `set_current_apps()` can get around that, as far as I'm aware.

AlanCoding:
yes, that's exactly what I thought set/get current apps was trying to work around.

Issue #13234, opened 2022-11-28
AWX receptor issue with verifysignature. 


janorn:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

Trying todo a healthcheck on the remote execution node after upgrading from 21.8.0 -> 21.9.0.
Got this error in the UI:
```
Remote error: ERROR: could not parse response: ERROR: could not verify signature: signature is empty
```
Did a reinstall of the remote execution node. After which it failed to start.
```
# /usr/bin/receptor -c /etc/receptor/receptor.conf
Error: error loading config file: error setting field verifysignature in command work-command: value cannot be nil
```
Finally i found out the root cause:
```
# grep verifysignature /etc/receptor/receptor.conf
    verifysignature:
```
Removing this config makes everything working again. Setting it to true makes the node happy but not AWX. 

### AWX version

AWX 21.9.0  

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [X] Other

### Installation method

kubernetes

### Modifications

no

### Ansible version

ansible [core 2.12.2]

### Operating system

RHEL 8.6

### Web browser

Firefox

### Steps to reproduce

Trying todo a healthcheck on the remote execution node after upgrading from 21.8.0 -> 21.9.0.
Got this error in the UI:
```
Remote error: ERROR: could not parse response: ERROR: could not verify signature: signature is empty
```
Did a reinstall of the remote execution node. After which it failed to start.
```
# /usr/bin/receptor -c /etc/receptor/receptor.conf
Error: error loading config file: error setting field verifysignature in command work-command: value cannot be nil
```
Finally i found out the root cause:
```
# grep verifysignature /etc/receptor/receptor.conf
    verifysignature:
```
Removing this config makes everything working again. Setting it to true makes the node happy but not AWX. 

### Expected results

Remote execution node turning up green in UI.

### Actual results

Red

### Additional information

receptor.log
```
Error: error loading config file: error setting field verifysignature in command work-command: value cannot be nil
Error: error loading config file: error setting field verifysignature in command work-command: value cannot be nil
Error: error loading config file: error setting field verifysignature in command work-command: value cannot be nil
[...]

PR #13253, opened 2022-11-30
Revert verifysignature variablization in bundle

relrod:
##### SUMMARY

In #13200 the dev env was changed to make `verifysignature` optional, dependent on a variable set before ansible gets run to set up the `docker-compose` environment.

However along with that change, a change to the execution node install bundle slipped in, which is seemingly unrelated to the dev env change and is breaking some installs: #13234, ansible/awx-operator#1132.

I think this change was unintentional as it would at least require another change in ansible/receptor-collection and maybe a change in ansible/awx-operator as well.

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/api/templates/instance_install_bundle/group_vars/all.yml
+++ b/awx/api/templates/instance_install_bundle/group_vars/all.yml
@@ -7,7 +7,7 @@ receptor_work_commands:
     command: ansible-runner
     params: worker
     allowruntimeparams: true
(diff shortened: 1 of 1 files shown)


shanemcd (approved):

shanemcd:
Oops! Thanks for fixing this.

AlanCoding (approved):
Great find.

PR #13254, opened 2022-11-30
bugfix: add scm_branch to optional_args for workflow_launch

dnguyen9:
##### SUMMARY

scm_branch argument is not processed by the workflow_launch. small change in how optional_args are processed to be consistent with job_launch collection. since extra_vars param is already enforced by the AnsibleModule as a dict, we do not need to do json.dumps.

##### ISSUE TYPE

##### COMPONENT NAME

 - Collection

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```


john-westcott-iv:
Thanks for this contribution. We are getting some testing done around this and will figure out next steps.

john-westcott-iv:
It looks like there is a failing linting test now that we no longer us `json.dumps`, is this something you can fix? Let me know if you need any help.

dnguyen9:
> It looks like there is a failing linting test now that we no longer us `json.dumps`, is this something you can fix? Let me know if you need any help.

I think remove 'json' from import should fix it

john-westcott-iv:
Apologies, but we will need to rebase this issue. There was a bug we hit with rsyslog that prevents all of the tests from running. if you rebase this PR the fix for that issue should be picked up and all of these tests will turn green.

john-westcott-iv (approved):

jay-steurer (approved):

Issue #12429, opened 2022-06-24
JT Form should automatically feed resource when created from the tab inside that resource


akus062381:
Credential, inventory, and project resources all recently had a Job Template tab added inside of the individual resources. 
Currently, if a user clicks on the Job Templates tab inside of a project, and adds a new Job Template, that particular project is populated in the project field in the new Job Template form.

This is not the case for Inventory or Credential.
I think it would make sense for this to be automatically populated for the user.

To access:

Create an inventory and/or create a project.
Click on the resource you just created so that you see its detail screen.
Click on the job templates tab for that resource.
Click Add.
You should (in my opinion) see the pertinent inventory name populated in the inventory field (if you reached the JT form through the inventory) or should see the pertinent Credential name populated in the credential field (if you reached the JT form through the credential).

PR #13262, opened 2022-12-01
Prepopulates job template form with related resource

AlexSCorey:
##### SUMMARY
This addresses #12429 
When in `/credentials/id/job_templates` or in inventories/id/job_templates`.  If the user clicks Add they are navigated to  the Job Template Add form.  This change now prepopulates the resource, from which the user came, in the appropriate field in the form.  All of that info about the related resource is passed through to the job template form via a querystring in the url.

There is 1 notable change here.  Not all credentials can be a part of a job template.  Therefore, we now add the Job Templates tab to ONLY those credentials whose `kind` is able to be added to a job template.  

This functionality already exists for projects.

Users cannot add job templates from /instange_groups/id, or execution_environments/id as those resources do not have a job templates tab.
##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui/src/screens/Credential/Credential.js
+++ b/awx/ui/src/screens/Credential/Credential.js
@@ -22,6 +22,16 @@ import { CredentialsAPI } from 'api';
 import CredentialDetail from './CredentialDetail';
 import CredentialEdit from './CredentialEdit';
 
+const jobTemplateCredentialTypes = [
+  'machine',
+  'cloud',
+  'net',
+  'ssh',
+  'vault',
+  'kubernetes',
+  'cryptography',
+];
+
 function Credential({ setBreadcrumb }) {
   const { pathname } = useLocation();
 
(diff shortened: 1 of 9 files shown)


vidyanambiar (approved):
Looks good!

awx/ui/src/screens/Credential/Credential.js
```diff
@@ -22,6 +22,16 @@ import { CredentialsAPI } from 'api';
...
 
+const jobTemplateCredentialTypes = [
+  'machine',
+  'cloud',
+  'net',
+  'ssh',
+  'vault',
+  'kubernetes',
+  'cryptography',
+];
```
vidyanambiar, line 33: Curious, how is this list determined?

AlexSCorey: Job templates only support certain types of credentials and the api doesn't have a way of telling the UI, which types are JTs support so the UI has the be the source of truth for that information.

Issue #13070, opened 2022-10-17
Enable abstract Credential Types 


philipsd6:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

Enhancement to Existing Feature

### Feature Summary

If I have a playbook that needs to talk to 5 different APIs, each requiring separate credentials I must provide 5 different Credential Types, and 5 instances of those Credential Types -- even if all of these APIs are essentially identical:

    host: https:example1.com
    username: joe
    password: s3krit

This is difficult to manage, as Credential Types must be created by an Admin. If the configuration for the Credential Type could be configured to inject dynamic variables instead of hard coded ones, we could have a single "Generic API" credential type with a "prefix" field that could then be used to generate multiple credentials with differently prefixed variables. 

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [X] Other

### Steps to reproduce

For example, if we could do this with the Input for a generic API host:

```yaml
fields:
  - id: host
    label: API Host
  - id: username
    label: Username
  - id: password
    label: Password
    secret: true
  - id: prefix
    label: Prefix for the credential variables
required:
  - host
  - username
  - password
  - prefix
```
And use the prefix like this in the Injector:
```yaml
extravars:
  "{{ prefix }}_host: "{{ host }}"
  "{{ prefix }}_username: "{{ username }}"  
  "{{ prefix }}_password: "{{ password }}"
```

Then we could create credentials for different environments (dev/test/prod) or scopes (internal/external) using the same Credential Type.

### Current results

Currently it is an error to use Jinja2 templating in the key side of an Injector configuration.

### Sugested feature result

With the ability to use Jinja2 templating in the key side of injectors, we can create an abstract/generic Credential Type that can be used for multiple credentials in the same Job Template.

AlanCoding:
With https://github.com/ansible/awx/issues/1965 now addressed, it could be possible to address this issue in principal. As I've said in other discussions, I would direct attention to a rewrite of `unique_hash`.

https://github.com/ansible/awx/blob/0815f935ca359537c023efadf286671e31dc3374/awx/main/models/credential/__init__.py#L237

[...]

Issue #1965, opened 2018-06-06
Custom credentials injector does not support hash result, only flat


miha-plesko:
##### ISSUE TYPE

 - Bug Report

##### COMPONENT NAME

- API

##### SUMMARY
When defining custom credential type, I'm not able to inject fields as a dictionary, only flat string values seem to be accepted.

##### ENVIRONMENT
* AWX version: 1.0.6.15
* AWX install method: docker on linux
* Ansible version:  2.5.4
* Operating System: Ubuntu 16.04
* Web Browser: chromium

##### STEPS TO REPRODUCE

Navigate to **CREDENTIAL TYPES -> Create a new credential type** and input following data:

- name: "My Custom Credential"
- input configuration:

```yaml
---
fields:
  - id: userid
    type: string
    label: Username
  - id: password
    type: string
    label: Password
```

- injector configuration:

```yaml
---
extra_vars:
  auth:
    api_username: '{{ userid }}'
    api_password: '{{ password }}'
```

Then click **SAVE**.

##### EXPECTED RESULTS

Credential type should be saved with no problems.

##### ACTUAL RESULTS

Following error appears:

```
OrderedDict provided in relative path ['extra_vars', u'auth'], expected string
```

![capture](https://user-images.githubusercontent.com/8102426/41024247-000379ac-696f-11e8-8dda-5f9265519850.PNG)

##### ADDITIONAL INFORMATION

The result I'm trying to achieve is that when I later use such credentials type, following variable is available to my playbook:

```yaml
---
auth:
  api_username: myusername
  api_password: mypassword
```

ATM this seems not to be possible and I'm only able to pass flat values like this:

```yaml
---
auth_api_username: myusername
auth_api_password: mypassword
```

but then I need to modify all my playbooks that depend on `auth` variable only because I run them via AWX instead ansible directly...

AlanCoding:
Although it would be relatively easy to change the JSON schema of the injector validator to _allow_ nested extra_vars structures, it would require novel work to actually know how we plan to _inject_ those variables. Currently we do the obvious top-level loop:

https://github.com/ansible/awx/blob/4fe7c9ea427f5b00cd89a6a1583a7bf9c5eb9a61/awx/main/models/credential/__init__.py#L654-L656

I believe this code assumes that `tmpl` is a string. Ping @ryanpetrello for his thoughts.

ryanpetrello:
What @AlanCoding said - right now injectors only support flat strings w/ naive Jinja replacement.

[...]

PR #13267, opened 2022-12-01
Enable support for injecting complex extra vars

philipsd6:
##### SUMMARY
Until now, it's only been possible to inject flat strings with Jinja templating into extra vars from Custom Credential Types. However, it's quite desirable to be able to pass a data structure containing authentication information for certain collections/modules. For example, the [Infoblox collection](https://docs.ansible.com/ansible/latest/scenario_guides/guide_infoblox.html#credentials-and-authenticating) accepts authentication under the `nios_provider` key.

This PR enables that data structure to be created by an injector config.

related #1965 #13070

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - Other

##### AWX VERSION

```
PYTHON=python3 make VERSION
awx: 21.9.1.dev32+g4af80777dd
```

##### ADDITIONAL INFORMATION
This PR also makes it possible to template the keys of an injector configuration, as requested in #13070.

For example, given an input configuration providing values for `environment`, `host`, `apikey`, `username`, it's possible to use an extra_vars injector configuration of:
```yaml
"{{ environment }}_auth":
  host: "{{ host }}"
  apikey: "{{ apikey }}"
  username: "{{ username }}"
  verify_ssl: "{{ verify_ssl }}"
```
resulting in the following extra var available to the playbooks:
```yaml
prod_auth:
  host: http://production.server
  apikey: 5hfthisj4hisj5hgnotj6hrealgj6
  username: philipsd6
  verify_ssl: True
```
This would lay the groundwork for allowing multiple credentials of the _same_ credential type, as another credential might supply values for a different environment.

### DIFF
--- a/awx/main/fields.py
+++ b/awx/main/fields.py
@@ -856,7 +856,7 @@ class CredentialTypeInjectorField(JSONSchemaField):
                 template_name = template_name.split('.')[1]
                 setattr(valid_namespace['tower'].filename, template_name, 'EXAMPLE_FILENAME')
 
-        def validate_template_string(tmpl):
+        def validate_template_string(type_, key, tmpl):
             try:
                 sandbox.ImmutableSandboxedEnvironment(undefined=StrictUndefined).from_string(tmpl).render(valid_namespace)
             except UndefinedError as e:
@@ -874,23 +874,23 @@ class CredentialTypeInjectorField(JSONSchemaField):
                     params={'value': value},
                 )
 
-        def validate_extra_vars(node):
+        def validate_extra_vars(key, node):
             if isinstance(node, dict):
-                return {validate_extra_vars(k): validate_extra_vars(v) for k, v in node.items()}
+                return {validate_extra_vars(key, k): validate_extra_vars(k, v) for k, v in node.items()}
             elif isinstance(node, list):
-                return [validate_extra_vars(x) for x in node]
+                return [validate_extra_vars(key, x) for x in node]
             else:
-                validate_template_string(node)
+                validate_template_string("extra_vars", key, node)
 
         for type_, injector in value.items():
             if type_ == 'env':
                 for key in injector.keys():
                     self.validate_env_var_allowed(key)
             if type_ == 'extra_vars':
-                validate_extra_vars(injector)
+                validate_extra_vars(None, injector)
             else:
                 for key, tmpl in injector.items():
-                    validate_template_string(tmpl)
+                    validate_template_string(type_, key, tmpl)
 
 
 class AskForField(models.BooleanField):
--- a/awx/main/fields.py
+++ b/awx/main/fields.py
@@ -856,27 +856,41 @@ class CredentialTypeInjectorField(JSONSchemaField):
                 template_name = template_name.split('.')[1]
                 setattr(valid_namespace['tower'].filename, template_name, 'EXAMPLE_FILENAME')
 
+        def validate_template_string(tmpl):
+            try:
+                sandbox.ImmutableSandboxedEnvironment(undefined=StrictUndefined).from_string(tmpl).render(valid_namespace)
+            except UndefinedError as e:
+                raise django_exceptions.ValidationError(
+                    _('{sub_key} uses an undefined field ({error_msg})').format(sub_key=key, error_msg=e),
+                    code='invalid',
+                    params={'value': value},
+                )
+            except SecurityError as e:
+                raise django_exceptions.ValidationError(_('Encountered unsafe code execution: {}').format(e))
+            except TemplateSyntaxError as e:
+                raise django_exceptions.ValidationError(
+                    _('Syntax error rendering template for {sub_key} inside of {type} ({error_msg})').format(sub_key=key, type=type_, error_msg=e),
+                    code='invalid',
+                    params={'value': value},
+                )
+
+        def validate_extra_vars(node):
+            if isinstance(node, dict):
+                return {validate_extra_vars(k): validate_extra_vars(v) for k, v in node.items()}
+            elif isinstance(node, list):
+                return [validate_extra_vars(x) for x in node]
+            else:
+                validate_template_string(node)
+
         for type_, injector in value.items():
             if type_ == 'env':
                 for key in injector.keys():
                     self.validate_env_var_allowed(key)
-            for key, tmpl in injector.items():
-                try:
-                    sandbox.ImmutableSandboxedEnvironment(undefined=StrictUndefined).from_string(tmpl).render(valid_namespace)
-                except UndefinedError as e:
-                    raise django_exceptions.ValidationError(
-                        _('{sub_key} uses an undefined field ({error_msg})').format(sub_key=key, error_msg=e),
-                        code='invalid',
-                        params={'value': value},
-                    )
-                except SecurityError as e:
-                    raise django_exceptions.ValidationError(_('Encountered unsafe code execution: {}').format(e))
-                except TemplateSyntaxError as e:
-                    raise django_exceptions.ValidationError(
-                        _('Syntax error rendering template for {sub_key} inside of {type} ({error_msg})').format(sub_key=key, type=type_, error_msg=e),
-                        code='invalid',
-                        params={'value': value},
-                    )
+            if type_ == 'extra_vars':
+                validate_extra_vars(injector)
+            else:
+                for key, tmpl in injector.items():
+                    validate_template_string(tmpl)
 
 
 class AskForField(models.BooleanField):
--- a/awx/main/fields.py
+++ b/awx/main/fields.py
@@ -790,13 +790,14 @@ class CredentialTypeInjectorField(JSONSchemaField):
                 'extra_vars': {
                     'type': 'object',
                     'patternProperties': {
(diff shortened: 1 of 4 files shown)


philipsd6:
I'm opening this PR to generate some discussion -- I'll update with tests and documentation if this passes muster.

AlanCoding:
Let's get your example documented in `docs/credentials/custom_credential_types.md`

philipsd6:
> Let's get your example documented in `docs/credentials/custom_credential_types.md`

I'll document the major part of this PR, but not the ability to template the key (yet) as that needs some discussion.

awx/main/models/credential/__init__.py
```diff
@@ -540,6 +545,7 @@ def build_extra_vars_file(vars, private_dir):
                 os.chmod(path, stat.S_IRUSR)
                 return path
 
+            extra_vars = render_extravars(self.injectors.get('extra_vars', {}))
```
AlanCoding: Should this be a call to `build_extra_vars` instead?

awx/main/models/credential/__init__.py
```diff
@@ -528,9 +528,14 @@ class TowerNamespace:
 
         if 'INVENTORY_UPDATE_ID' not in env:
             # awx-manage inventory_update does not support extra_vars via -e
-            extra_vars = {}
-            for var_name, tmpl in self.injectors.get('extra_vars', {}).items():
-                extra_vars[var_name] = sandbox_env.from_string(tmpl).render(**namespace)
+            def build_extra_vars(node):
+                if isinstance(node, dict):
+                    return {sandbox_env.from_string(k).render(**namespace):
```
AlanCoding: my 2 cents, I'd rather call `build_extra_vars(k)` as opposed to `sandbox_env.from_string(k).render(**namespace)`. If we're recursive anyway, might as well own it.

awx/main/models/credential/__init__.py
philipsd6: YES it SHOULD! My bad, I missed that when I refactored my code...

shanemcd:
This is awesome 😄 Before we merge this we'll need to get some tests added. Check out `test_tasks.py` - there's a lot of examples of tests calling `inject_credential`.

philipsd6:
Added two tests for nested and templated extra vars keys, but still TBD: Updating the jsonschema for the `CredentialTypeInjectorField` model (the tests will fail until this is done.)

john-westcott-iv:
@philipsd6 if you are all set with your changes can you convert this from a draft PR to an opened PR?

philipsd6:
> @philipsd6 if you are all set with your changes can you convert this from a draft PR to an opened PR?

I don't think it's ready yet -- I'm struggling with the jsonschema that will allow recursively nested injector vars. Currently in my test env, I'm still getting this error when trying to create a nested extra_vars injector:

    OrderedDict([('nested_auth', OrderedDict([('username', '{{ username }}'), ('password', '{{ password }}')]))]) is not valid under any of the given schemas

philipsd6:
I updated `awx/main/fields.py:CredentialTypeInjectorField` to try to make the `extra_vars` property recursive, but while the schema is ok with an online validator: https://www.jsonschemavalidator.net/s/Yr7ExLBf it still fails in AWX when trying to update a credential with a nested extra_var injector with:

    TypeError: Can't compile non template nodes
    10.0.2.100 PATCH /api/v2/credential_types/28/ - HTTP/1.1 500

philipsd6:
> I updated `awx/main/fields.py:CredentialTypeInjectorField` to try to make the `extra_vars` property recursive, but while the schema is ok with an online validator: https://www.jsonschemavalidator.net/s/Yr7ExLBf it still fails in AWX when trying to update a credential with a nested extra_var injector with:
> 
> ```
> TypeError: Can't compile non template nodes
> 10.0.2.100 PATCH /api/v2/credential_types/28/ - HTTP/1.1 500
> ```
My bad, the jsonschema update worked fine, it was the validator that was failing on the nested part. Fixed with latest commit.

awx/main/fields.py
```diff
@@ -790,13 +790,11 @@ def schema(self, model_instance):
                 'extra_vars': {
                     'type': 'object',
                     'patternProperties': {
-                        # http://docs.ansible.com/ansible/playbooks_variables.html#what-makes-a-valid-variable-name
-                        '^[a-zA-Z_]+[a-zA-Z0-9_]*$': {'type': 'string'},
+                        r'^(?:(?:{(?:{|%)[^{}]*?(?:%|})})|(?:[a-zA-Z_]+[a-zA-Z0-9_]*)+)+$': {"anyOf": [{'type': 'string'}, {'$ref': '#/properties/extra_vars'}]}
```
philipsd6: This regex is ridiculous, but the only alternative I can see is to just make it `r'^.+$'`

awx/main/fields.py
```diff
@@ -858,27 +856,41 @@ def __str__(self):
...
+                raise django_exceptions.ValidationError(
+                    _('{sub_key} uses an undefined field ({error_msg})').format(sub_key=key, error_msg=e),
+                    code='invalid',
+                    params={'value': value},
+                )
+            except SecurityError as e:
+                raise django_exceptions.ValidationError(_('Encountered unsafe code execution: {}').format(e))
+            except TemplateSyntaxError as e:
+                raise django_exceptions.ValidationError(
+                    _('Syntax error rendering template for {sub_key} inside of {type} ({error_msg})').format(sub_key=key, type=type_, error_msg=e),
```
AlanCoding, line 875: I'm getting a server error due to the fact that `key` is undefined here.

philipsd6: yup, valid -- and not only that, so is `type_`. I'll have to revise a little deeper after closer inspection. Thanks for the review.

awx/main/fields.py
```diff
@@ -858,27 +856,41 @@ def __str__(self):
...
+                    code='invalid',
+                    params={'value': value},
+                )
+
+        def validate_extra_vars(key, node):
+            if isinstance(node, dict):
+                return {validate_extra_vars(key, k): validate_extra_vars(k, v) for k, v in node.items()}
```
AlanCoding: We really don't want to lose context in error messages. The `key` passed in should always be a string right? So what if you combined `k` and `key` for the value here, so instead of `validate_extra_vars(k, v)` do `validate_extra_vars(f'{key}:{k}', v)`, or whatever you'd prefer for delimiter. That would keep stacking layers of context on a rolling basis.

awx/main/fields.py
```diff
@@ -858,27 +856,41 @@ def __str__(self):
...
+            elif isinstance(node, list):
+                return [validate_extra_vars(key, x) for x in node]
```
AlanCoding: I'd like to change `for x in node` to `for i, x in enumerate(node)` and then stack onto `key`, like `f'{key}:{i}`. It doesn't distinguish between dicts and lists... but it shouldn't matter because the user know what injectors they submitted.

awx/main/fields.py
philipsd6: I used a '.' (dot) as the dict key delimiter, seemed appropriate for this.

awx/main/fields.py
philipsd6: I think the context is fully maintained now in the validation error messages. I used `key[i][i]` style for the validation error message, as between the dict dot style and this, it should be pseudo jq/jmespath style.

@AlanCoding: running `podman exec -it awx-tools_awx_1 make test_unit` works passes the new unit tests, But I'm not sure how to actually test those validation error paths.

awx/main/fields.py
```diff
@@ -858,27 +858,41 @@ def __str__(self):
...
-                except SecurityError as e:
-                    raise django_exceptions.ValidationError(_('Encountered unsafe code execution: {}').format(e))
-                except TemplateSyntaxError as e:
-                    raise django_exceptions.ValidationError(
-                        _('Syntax error rendering template for {sub_key} inside of {type} ({error_msg})').format(sub_key=key, type=type_, error_msg=e),
-                        code='invalid',
-                        params={'value': value},
-                    )
+            if type_ == 'extra_vars':
+                validate_extra_vars(None, injector)
```
AlanCoding, line 896: This `None` is coming through in the error message with a grammar that's not really desirable. I've actually been using the docs as an error case, because the existing example is badly formatted. That does need to be changed... but the value is:

```json
"injectors": {
    "env": {
        "THIRD_PARTY_CLOUD_API_TOKEN": "{{api_token}}"
    },
    "extra_vars": {
        "some_extra_var": "{{username}}:{{password}",
        "auth": {
            "username": "{{username}}",
            "password": "{{password}}"
        }
    }
}
```

A bit here is missing a closing `}`. The currently produced error message is:

> {'injectors': ["Syntax error rendering template for None.auth.username inside of extra_vars (unexpected '}')"]}

You know, in this line, `injector` must be a dictionary. This is fixable, but I think it will take an extra line or two of string formatting inside of `validate_extra_vars`.

I am going to write down a few validation test cases and save them for later.

jay-steurer (approved):

AlanCoding:
I poked at this some more and put up https://github.com/philipsd6/awx/pull/571, which is mainly concerned with error messages.

AlanCoding (approved):

PR #13268, opened 2022-12-01
Copy UI static files to /var/lib/awx only for ui-devel build

simaishi:
##### SUMMARY
#13125 broke productization rpm build's `make ui-release` (permission denied on `mkdir -p /var/lib/awx/public/static/css`)

##### ISSUE TYPE

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION
This change was inteneded to be made for `ui-devel` as per #13125 , so moving the mkdir/copy to ui-devel target.


simaishi:
Thanks to @shanemcd for testing out the change! 😄

shanemcd (approved):

PR #13277, opened 2022-12-02
Configure dbshell to use the 'less' pager

jbradberry:
##### SUMMARY
Configure dbshell to use the 'less' pager.  The 'more' pager used by default is obnoxious.

This ought to only apply to the dev environment.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
+++ b/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
@@ -273,7 +273,8 @@ RUN for dir in \
       /var/lib/shared/vfs-layers/layers.lock \
       /var/run/nginx.pid \
       /var/lib/awx/venv/awx/lib/python3.9/site-packages/awx.egg-link ; \
-    do touch $file ; chmod g+rw $file ; done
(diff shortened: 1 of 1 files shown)


shanemcd (approved):

tools/ansible/roles/dockerfile/templates/Dockerfile.j2
```diff
@@ -274,6 +274,8 @@ RUN for dir in \
       /var/run/nginx.pid \
       /var/lib/awx/venv/awx/lib/python3.9/site-packages/awx.egg-link ; \
     do touch $file ; chmod g+rw $file ; done
+
+RUN echo "\setenv PAGER 'less -S'" > /var/lib/awx/.psqlrc
```
relrod: Maybe combine with the `RUN` right above it, just to avoid another image layer?

jbradberry: like this?

relrod: Yep, lgtm

relrod (approved):

Issue #12266, opened 2022-05-23
Communication with receptor fails: could not load signing key file: unknown block type PRIVATE KEY


edudobay:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

I've installed AWX in a test server using the instructions from https://github.com/ansible/awx/blob/devel/tools/docker-compose/README.md, and the application fails to communicate with the receptor instances — messages like these show up in the log and in the instance status page: `Remote error: ERROR: could not load signing key file: unknown block type PRIVATE KEY`.

### AWX version

21.0.0

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs

### Installation method

docker on linux

### Modifications

no

### Ansible version

2.12.5 (ansible-core) 5.8.0 (ansible)

### Operating system

Ubuntu 22.04

### Steps to reproduce

What I did for the setup:
* Installed Docker (official installation steps for Ubuntu), Docker Compose and Ansible (`sudo pip install docker-compose ansible`)
* `git clone -b 21.0.0 https://github.com/ansible/awx.git; cd awx`
* `make docker-compose-build`
* `make docker-compose COMPOSE_UP_OPTS=-d`
* build the UI with `docker exec tools_awx_1 make clean-ui ui-devel`
* configured the user with `docker exec -ti tools_awx_1 awx-manage createsuperuser`

### Expected results

All instances should be healthy.

### Actual results

Both `receptor-1` and `receptor-2` have an error state in the Instances page: https://localhost:8043/#/instances

![image](https://user-images.githubusercontent.com/4432091/169754409-a0821764-d015-419b-add9-e80b56123d8f.png)

### Additional information

Log output from `docker logs tools_awx_1`:

```
2022-05-23 05:10:45,397 WARNING  [f6876450] awx.main.tasks.system Hop node receptor-hop, has rejoined the receptor mesh
2022-05-23 05:10:45,439 WARNING  [f6876450] awx.main.tasks.system Execution node attempting to rejoin as instance receptor-1.
2022-05-23 05:10:45,499 WARNING  [f6876450] awx.main.tasks.system Execution node attempting to rejoin as instance receptor-2.
[...]

PR #13280, opened 2022-12-02
Re-enable work signing in development environment

shanemcd:
##### SUMMARY
Effectively reverses what I did here: https://github.com/ansible/awx/pull/13200.

Using the approach described in https://github.com/ansible/awx/issues/12266#issuecomment-1171605519

##### ISSUE TYPE


shanemcd:
Holding off on this, it seems like the version of OpenSSL on macOS doesn't support `-traditional`...

shanemcd:
Yeah, never mind.

PR #13281, opened 2022-12-02
Respect work signing configuration for remote work types

shanemcd:
##### SUMMARY
Follow-up from https://github.com/ansible/awx/pull/13253

This will fix: 

- https://github.com/ansible/awx-operator/issues/1132
- https://github.com/ansible/awx/issues/13234

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
21.10.1.dev3+gc440572838
```

### DIFF
--- a/awx/main/tasks/receptor.py
+++ b/awx/main/tasks/receptor.py
@@ -61,11 +61,17 @@ def read_receptor_config():
             return yaml.safe_load(f)
 
 
-def work_signing_enabled(config_data):
-    for section in config_data:
-        if 'work-verification' in section:
-            return True
-    return False
+def work_signing_enabled(receptor_ctl, remote_node, work_type):
+    receptor_status = receptor_ctl.simple_command("status")
+
+    for node in receptor_status["Advertisements"]:
+        if node["NodeID"] == remote_node:
+            for work in node["WorkCommands"]:
+                if work["WorkType"] == work_type:
+                    return work["Secure"]
+
+    logger.warning(f"{remote_node} not found in receptor advertisments, defaulting to signed work.")
+    return True
 
 
 def get_receptor_sockfile(config_data):
@@ -301,6 +307,7 @@ class AWXReceptorJob:
         self.task = task
         self.runner_params = runner_params
         self.unit_id = None
+        self.execution_node = task.instance.execution_node
 
         if self.task and not self.task.instance.is_container_group_task:
             execution_environment_params = self.task.build_execution_environment_params(self.task.instance, runner_params['private_data_dir'])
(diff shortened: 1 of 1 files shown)


awx/main/tasks/receptor.py
```diff
@@ -61,11 +61,17 @@ def read_receptor_config():
...
+def work_signing_enabled(receptor_ctl, remote_node, work_type):
+    receptor_status = receptor_ctl.simple_command("status")
+
+    for node in receptor_status["Advertisements"]:
+        if node["NodeID"] == remote_node:
+            for work in node["WorkCommands"]:
+                if work["WorkType"] == work_type:
+                    return work["Secure"]
+
+    logger.warning(f"{remote_node} not found in receptor advertisments, defaulting to signed work.")
```
shanemcd, line 73: I'm doing this because you can technically submit work to a node that doesn't exist on the mesh yet. I don't know if this will ever happen in the AWX use case, but I just wanted a non-fatal fallback here.

relrod (approved):

awx/main/tasks/receptor.py
```diff
@@ -487,9 +494,7 @@ def receptor_params(self):
 
     @property
     def sign_work(self):
-        if self.work_type in ('ansible-runner', 'local'):
-            return work_signing_enabled(self.config_data)
-        return False
+        return work_signing_enabled(self.receptor_ctl, self.execution_node, self.work_type)
```
AlanCoding, line 497: We still don't intend to sign work for container_group jobs, right? The keys would have to be backed into the image or something and that's not workable.

Container group jobs don't use the "ansible-runner" work type, so that's how this returned False in that case. I think we still want to return False in that case.

AlanCoding:
Given the cost and consequence of running the receptorctl status command before every job, this is something that would make sense to cache as a field on the `Instance` model. We already run the status command as an integral part of execution node health checks (and they won't have capacity before then), if it's not easier to have set it from awx-manage registering command.

AlanCoding:
I put up https://github.com/ansible/awx/pull/13296 as a more direct fix of the bug I introduced in the last related merge.

shanemcd:
@AlanCoding  Totally fine merging that PR first (I've approved it) if there are concerns about the performance implications of this change (I think you are right to be cautious here).

- Do you think it's worth refactoring this PR a bit and caching the info on the Instance model?
- ...or just make it a global setting and move on?

My main concern right now is this: we unconditionally add `work-signing` to AWX's Receptor configuration. But we also give people the ability to change the value for `verifysignature` in the vars fie we generate for the install bundle. I can understand if we don't want to make this per-worktype and instead opt for an install-scoped setting inside AWX.

AlanCoding:
It took me a while to come around and see what you meant per-worktype, but considering how the install bundle organizes vars, I'm following better now.

Where I still struggle - we don't support signing work at all for some work types (like `kubernetes-runtime-auth`), and I question if it's wise to even allow turning it _off_ for execution nodes. My gut instinct is that I'd rather it not be configurable at all. I can't see through to the intended use.

Again, going off gut reactions, it feels like receptor would be a better place to handle this. We're adding in a step of communication with receptor to find a parameter to send to receptor.

AlanCoding:
I thought I knew how this PR would work with container_group jobs, and now I'm thinking I don't.

PR #13295, opened 2022-12-06
Remove un-editable Instance fields from pre-filled edit data in API browser

AlanCoding:
##### SUMMARY
You can clearly see that these are prohibited from editing in the serializer validators. For example:

```python
    def validate_hostname(self, value):
        """
        - Hostname cannot be "localhost" - but can be something like localhost.domain
        - Cannot change the hostname of an-already instantiated & initialized Instance object
        """
        if self.instance and self.instance.hostname != value:
            raise serializers.ValidationError("Cannot change hostname.")

        return value
```

It says it right there - the field is only editable in that it can be set on creation.

This is the instance _detail_ view, so there's no way you're here unless it has already been created. Including the fields is going to either do nothing or get you an error.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
for the URL

https://localhost:8043/api/v2/instances/1/

updated

Screenshot from 2022-12-06 15-19-27

before

Screenshot from 2022-12-06 15-22-30

We have a _very very similar_ type of override for InstanceGroup detail view, so it's logical to add this here.

### DIFF
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -344,6 +344,13 @@ class InstanceDetail(RetrieveUpdateAPIView):
     model = models.Instance
     serializer_class = serializers.InstanceSerializer
 
+    def update_raw_data(self, data):
+        # these fields are only valid on creation of an instance, so they unwanted on detail view
+        data.pop('listener_port', None)
+        data.pop('node_type', None)
+        data.pop('hostname', None)
+        return super(InstanceDetail, self).update_raw_data(data)
+
     def update(self, request, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


AlanCoding:
> ERROR! Unknown error when attempting to call Galaxy at 'https://galaxy.ansible.com/api/v2/collections/community/general/versions/5.8.1/': The read operation timed out

seems to be the check failing due to https://galaxy.ansible.com/community/general, but looks to be working now.

jbradberry (approved):

PR #13297, opened 2022-12-07
Set AWS_SESSION_TOKEN in addition to AWS_SECURITY_TOKEN

ilijamt:
##### SUMMARY
AWS started using `AWS_SESSION_TOKEN` instead of `AWS_SECURITY_TOKEN` since 2014. A lot of libraries check for both, but I ended up in a case where a collection `community.aws.aws_ssm` did not check both, and ended up with errors like:

```
caught exception(An error occurred (UnrecognizedClientException) when calling the StartSession operation: The security token included in the request is invalid.)
```

In this case, it was invalid because `AWS_SESSION_TOKEN` was not set.

https://aws.amazon.com/blogs/security/a-new-and-standardized-way-to-manage-credentials-in-the-aws-sdks/

##### ISSUE TYPE

##### COMPONENT NAME
 - Other (Model/Injectors)

##### AWX VERSION
21.10.0

##### ADDITIONAL INFORMATION

After setting `AWS_SESSION_TOKEN` I no longer got the error and could successfully connect to the EC2 machines.

### DIFF
--- a/awx/main/models/credential/injectors.py
+++ b/awx/main/models/credential/injectors.py
@@ -15,6 +15,7 @@ def aws(cred, env, private_data_dir):
 
     if cred.has_input('security_token'):
         env['AWS_SECURITY_TOKEN'] = cred.get_input('security_token', default='')
+        env['AWS_SESSION_TOKEN'] = env['AWS_SECURITY_TOKEN']
 
 
 def gce(cred, env, private_data_dir):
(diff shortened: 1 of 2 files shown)


john-westcott-iv:
Thanks for this contribution, we will get some QE testing going on this before merging.

jay-steurer (approved):

john-westcott-iv (approved):

PR #13298, opened 2022-12-07
Allow to use Hashicorp Vault Auth in AWX with AWS

ilijamt:
##### SUMMARY
I don't want to store AWS credentials in AWX, because I want to use short lived credentials through Vault for each run.Currently this is not possible with AWX. This is a draft implementation of a custom Vault Auth managed credential type, along with a managed credential. 

This will allow us to use Vault Auth backend with the Vault/AWS credential type.

Comments regarding implementation are welcome.  Current implementation works with AWS, and it is in use for some time.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

##### AWX VERSION
21.10.0

##### ADDITIONAL INFORMATION

* AWS Inventory
```yaml
plugin: aws_ec2
regions:
  - eu-west-1

filters:
  instance-state-name: running

hostnames:
  - tag:Name

compose:
  ansible_host: instance_id
  ansible_aws_ssm_region: placement.region

keyed_groups:
  - prefix: tag
    key: tags | lower
  - key: placement.region
    prefix: aws_region
```                                                                                                                                                                                                              
* Playbook
```yaml
- hosts: all
  gather_facts: yes
  tasks:
    - name: Display all variables/facts known for a host
      ansible.builtin.debug:
        var: hostvars
    - name: Display some secrets that need to be passed
      ansible.builtin.debug:
        msg: "{{ lookup('env', item) }}"
      with_items:
        - AWS_ACCESS_KEY_ID
        - AWS_SECRET_ACCESS_KEY
        - AWS_SESSION_TOKEN
```
* Terraform 

```hcl
terraform {
  required_providers {
    awx = {
      source = "registry.terraform.io/ilijamt/awx"
    }
  }
}

provider "awx" {
  insecure_skip_verify = true
}

data "awx_credential_type" "hashivault_auth" {
  name = "HashiCorp Vault Auth"
}

data "awx_credential_type" "vault_aws" {
  name = "Vault/AWS"
}

data "awx_organization" "default" {
  name = "Default"
}

resource "awx_credential" "hashivault_auth" {
  name            = "hashivault-auth"
  credential_type = data.awx_credential_type.hashivault_auth.id
  organization    = data.awx_organization.default.id
  description     = "We can use this to auth against Vault"
  inputs = jsonencode({
    "url"   = "https://vault.example.com"
    "token" = "token"
  })
}

resource "awx_credential" "hashivault_auth_ansible_ssm" {
  name            = "ansible-ssm"
  credential_type = data.awx_credential_type.vault_aws.id
  organization    = data.awx_organization.default.id
}

resource "awx_credential_input_source" "hashivault_auth_input_ansible_ssm" {
  input_field_name = "auth_full_path"
  metadata = jsonencode({
    "auth_full_path" : "aws/creds/ansible-ssm"
  })
  target_credential = awx_credential.hashivault_auth_ansible_ssm.id
  source_credential = awx_credential.hashivault_auth.id
}
```


john-westcott-iv:
Thanks for this PR we will work with our engineers and QE team to review internally and will review next steps for this.

ilijamt:
@john-westcott-iv any update on this?

TheRealHaoLiu:
closing and reopen to unstuck the CI...

ilijamt:
Hello,

Any update on this? Will this go forward? Are there any changes needed. Or should this be deployed differently?

AlanCoding:
I think it's fair to say that testing is a pain point for contributions like this. I have been digging around this area and saw https://github.com/ansible/awx/pull/13180, which should be fairly easy to update and get working, and we could merge that. This would allow someone to manually run against hashicorp vault, but I would like you to confirm or deny that the suggested image would work to test this.

I would like to obtain genuine regression coverage for things like this, and have https://github.com/AlanCoding/awx/pull/81 for this, which I want to expand to hashicorp vault lookup plugins specifically.

ilijamt:
@AlanCoding, yeah, adding https://github.com/ansible/awx/pull/13180 would allow us to run integration tests for this plugin.   

The only downside regarding integration testing with AWS is that we need to set up a trust between an AWS account and a role created so Vault can assume it to do proper testing.  Mock testing the plugin would be simpler. 

Regarding the credential plugin, are you suggesting that we move it to an external credential plugin? I did take a look at https://github.com/ansible/awx-custom-credential-plugin-example, but I couldn't get it to work properly with the external plugin. If you have some up-to-date documentation of how to hook this up with an external plugin, that would be great.

AlanCoding:
If you'll allow me to try to understand the intent of this change, this new credential type requires fields `['url', 'auth_full_path']`. There are two existing hashicorp credential types already.

```
In [10]: [ct.inputs['required'] for ct in CredentialType.objects.filter(namespace__startswith='hashi')]
Out[10]: 
[['url', 'secret_path', 'api_version', 'secret_key'],
 ['url', 'secret_path', 'public_key', 'role']]
```

If I look at the help text of `secret_path` this gives

> The path to the secret stored in the secret backend e.g, /some/secret/. It is recommended that you use the secret backend field to identify the storage backend and to use this field for locating a specific secret within that store. However, if you prefer to fully identify both the secret backend and one of its secrets using only this field, join their locations into a single path without any additional separators, e.g, /location/of/backend/some/secret.

I understand you giving an auth path in lieu of `secret_key` or `public_key`. If you consider the existing hashivault_kv and hashivault_ssh credential types, it seems that they _provide_ mostly the same things (via the `secret_path`), but accept different _means of authenticating_. This new proposed credential type does not follow the same pattern, and I can't figure out why.

I agree that the "external credential plugin" notion hasn't been actively maintained and I can't give any assurances about it. With multiple hashivault credentials, it would be useful to have subclasses or sub-types or something like that. For right now, I would lean in the direction of consistency, as we already have these other hasivault types, it would fit to add another type for another means of auth. Ideally... I just want to get testing for all these simultaneously.

ilijamt:
Regarding the differences, let's go over the differences of the implementations and why the current ones won't work. 

**hashivault_kv** - only reads a value from the KV store, that value won't change so often. 
**hashivault_ssh** - just signs using vault ssh signing the ssh key 

Now the problem I am facing, in my case I'm doing this for AWS (but this is applicable for Azure, GCP, Artifactory, ...). 
You set up a trust between AWS and Vault, so that Vault can generate credentials for AWS that you can use. 
Everytime we call the endpoint in Vault to get credentials for AWS, we get a new set generated with a predefined TTL.  So I cannot just point to an endpoint and retrieve the variables as they would be different every time. 

In the case of AWS, we get the three variables after we have finished authentication, that we need to pass directly to running container. 
- AWS_ACCESS_KEY_ID
- AWS_SECRET_ACCESS_KEY
- AWS_SESSION_TOKEN

So what we can do is maybe instead of defining a generic `hashivault_auth_plugin` is to refactor this to create a specific one just for AWS like `hashivault_auth_aws`. Would that work @AlanCoding ? 

In that case we can create a separate one if we want to for Azure, GCP, Artifactory and any other ones we would need. 

Me defining `hashivault_auth_plugin` was to abstract part of the logic and only have one plugin and a separate ManagedCredentialType with its injector for each of the auths we would need.

AlanCoding:
It actually flew over my head that this PR creates _two_ new credential types. So one is the lookup-plugin type, and the other is a hybrid AWS cred compatible with the other credential type. I also had confusion about the data type of `auth_full_path`, which is declared as a string but then treated as a dictionary.

I wonder if we could reduce this change to 1 credential that just handles the dictionary lookup from Hashicorp Vault, and then you might be able to create a corresponding custom credential type (strictly through the API) that applied the lookup plugin multiple types (for each sub-key of the dictionary) and then used the `injectors` to set the `AWS_*` environment variables to those values. Those are just current thoughts, still digesting.

ilijamt:
If we can reduce this to one, that would be great. But would a custom credential type work? Do you have an example of what you mean, so I can take a look?

AlanCoding:
I was talking about the basic custom credential types

https://github.com/ansible/awx/blob/devel/docs/credentials/custom_credential_types.md

The example there contains an environment variable example (probably the most common)

```json
    "injectors": {
        "env": {
            "THIRD_PARTY_CLOUD_API_TOKEN": "{{api_token}}"
        }
    }
```

Then when applying a lookup plugin credential, you apply it to an input of another credential. So whatever input key is going into the template (like `api_token` there) would need to be gotten from Hashicorp Vault, and I think that's what you're going for.

AlanCoding:
Slightly tangential to this, the AWS credential _should_ be something which can be implemented via the template system for credential types (like custom credentials). I filed https://github.com/ansible/awx/issues/14152 which is a new construct needed to make that fully airtight. For your purposes, it may still work fine if you set optional environment variables to empty string.

chrismeyersfsu:
Credential plugins have been moved out of awx and over to https://github.com/ansible/awx-plugins/tree/devel/src/awx_plugins/credentials

PR #13303, opened 2022-12-07
[constructed-inventory] Constructed inventory as alternative to smart inventory

AlanCoding:
##### SUMMARY
Connect https://github.com/ansible/awx/issues/13447

Replaces first draft of https://github.com/ansible/awx/pull/13214

##### ISSUE TYPE
 - New or Enhanced Feature

**Second Draft**

This was a full Proof-of-Concept that can replicate the steps outlined in the CLI-based demo here:

https://github.com/AlanCoding/Ansible-inventory-file-examples/tree/master/issues/AWX371

It has been edited since then, and that is no longer entirely the functionality.

Key changes with this version:
 - Introduced a new kind of inventory, "multiple" which linked to multiple source inventories
 - Introduced "constructed" inventory source which adds (at runtime in this version) constructed vars / groups

**Third Draft**

This will be continually edited as the shape of the feature changes.

Key changes with this version:
 - Instead of a "multiple" kind of inventory, this changes to a "constructed" kind of inventory
 - The inventory kind corresponds to a 1-to-1 mandatory "constructed" inventory source
 - The inventory/source updates independently

key TODOs:
 - Add `--limit` to inventory sources

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API


awx/main/tasks/jobs.py
```diff
@@ -1520,6 +1523,15 @@ def build_args(self, inventory_update, private_data_dir, passwords):
 
         args = ['ansible-inventory', '--list', '--export']
 
+        # special case for constructed inventories, we pass source inventories from database
+        # these must come in order, and in order _before_ the constructed inventory itself
+        if inventory_update.inventory.kind == 'constructed':
+            for source_inventory in inventory_update.inventory.source_inventories.all():
+                args.append('-i')
+                script_params = script_params = dict(hostvars=True, towervars=True)
```
relrod:
```suggestion
                script_params = dict(hostvars=True, towervars=True)
```

awx/main/models/inventory.py
```diff
@@ -1547,5 +1567,17 @@ class insights(PluginFileInjector):
...
 
+class constructed(PluginFileInjector):
+    plugin_name = 'constructed'
+    namespace = 'ansible'
+    collection = 'builtin'
+
+    def build_env(self, *args, **kwargs):
+        env = super().build_env(*args, **kwargs)
+        # Enable all types of inventory plugins so we pick up the script files from source inventories
+        del env['ANSIBLE_INVENTORY_ENABLED']
```
gamuniz, line 1578: Are we just removing this to enable the default enabled plugin list? Doesn't `auto` already try the list?

gamuniz (approved):

awx/api/serializers.py
```diff
@@ -1709,8 +1704,17 @@ def get_related(self, obj):
                 labels=self.reverse('api:inventory_label_list', kwargs={'pk': obj.pk}),
             )
         )
+        if obj.kind in ('', 'constructed'):
```
TheRealHaoLiu, line 1707: if these are ones that shouldn't show for "smart" inventory should we make this if statement reflect the condition?

AlanCoding: you mean change this to `obj.kind != 'smart'`?

PR #13304, opened 2022-12-07
Only allow promote and stage to run on the awx repo

john-westcott-iv:
##### SUMMARY

Per https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution we can limit what repos these workflows run on.

##### ISSUE TYPE

##### COMPONENT NAME

 - Other

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/.github/workflows/promote.yml
+++ b/.github/workflows/promote.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   promote:
-    if: github.repository == 'ansible/awx'
+    if: endsWith(github.repository, '/awx') 
     runs-on: ubuntu-latest
     steps:
       - name: Checkout awx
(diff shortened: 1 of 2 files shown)


jbradberry (approved):

.github/workflows/promote.yml
```diff
@@ -10,6 +10,7 @@ on:
 
 jobs:
   promote:
+    if: github.repository == 'ansible/awx'
```
shanemcd, line 13: I don't want to lose the ability to test releases out on our personal forks of AWX. This should probably be:

```
endsWith(github.repository, '/awx') 
```

shanemcd (changes requested):

.github/workflows/promote.yml
shanemcd:
```suggestion
    if: endsWith(github.repository, '/awx') 
```

shanemcd (approved):

Issue #13305, opened 2022-12-07
Credentials lookup in the Job Template launch prompt results in 'Not Found'


vidyanambiar:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

Launching a Job Template that has "prompt on launch" enabled for credentials, choosing multiple credentials and switching between pages changing the credential category several times results in a 'Not Found' error.

### AWX version

2.2.1

### Select the relevant components

- [X] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

N/A

### Modifications

no

### Steps to reproduce

1. Create a Job Template where 'Credentials' is set to 'Prompt on launch'.
2. Launch the Job Template and choose by multiple credentials.
3. When choosing the credentials and adding more credentials by switching between pages and changing the Credential category several times the following error is shown:
`Not Found The page you requested could not be found. Back to Dashboard. Details  GET api/v2/credentials/ 404 Invalid page.`
    - For example: If testing in the dev environment, and the template launch prompt contains 2 pages of data for "Machine" credentials, view the 2nd page of credentials and then select a different category such as "Vault". If this category does not have a 2nd page of data, then you'll see the "Not found" error.

To recreate the issue in the Add Template form:
1. Click "Add a job template" in the templates list to create a new job template 
2.  In the form for template creation, select the search/lookup button in the "Credentials" field. 
3. In the resulting dialog view page 2 of the credentials list (or any page other than page 1) and then select
a different credential category from the menu above. 

### Expected results

A 404 "Not Found" error should not be displayed when switching between credential categories.

### Actual results

When launching a Job Template (with "prompt on launch" enabled for credentials), choosing multiple credentials  and switching between pages of credentials and changing the credential category several times a Not Found error is displayed.

![not_found](https://user-images.githubusercontent.com/43621546/206274382-45089cc3-6489-4e09-b164-599c94b1919a.png)

The issue can also be seen in the Credential lookup in the "Add job template" form.

![image](https://user-images.githubusercontent.com/43621546/206275774-18a3aa17-0552-4e2c-9f5d-95d93f566d32.png)

PR #13306, opened 2022-12-07
Fixes 'Not Found' error on looking up credentials

vidyanambiar:
##### SUMMARY
This addresses #13305 

The issue appears to be that we were retaining the query params (credential.page) even when the credential category was changed and in turn making requests for pages that didn't exist on some categories.

Within the credentials lookup (either in the `Job Template launch prompt` or in the `Add Job Template` form):
1. Select a credential type that has more than one page worth of credentials data. (In the dev env I saw that the "Machine" type had more than one page of data).
2. Switch to page 2 of the credentials (or any page other than the first page)
3. Change the credential category using the dropdown to a category that doesn't have credentials or doesn't have a second page of data.
4. If there are no credentials for that type, the appropriate message should be displayed, but the UI should not show a 404 Not Found error.

##### ISSUE TYPE

##### COMPONENT NAME

 - UI

### DIFF
--- a/awx/ui/package.json
+++ b/awx/ui/package.json
@@ -74,7 +74,7 @@
     "prestart-instrumented": "lingui compile",
     "pretest": "lingui compile",
     "pretest-watch": "lingui compile",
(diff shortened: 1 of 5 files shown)


akus062381 (approved):
looks great, good catch!

marshmalien (approved):
Nice find and fix!

PR #13308, opened 2022-12-07
Ensure that the Organization.execution_environment_admin_role always gets built

jbradberry:
##### SUMMARY
Ensure that the Organization.execution_environment_admin_role always gets built.

##### ISSUE TYPE

##### COMPONENT NAME
 - API


AlanCoding:
I thought this was a good targeted solution for backports, but for the long-term future, I'd like something that's more consistent with the existing code structure. What you have seems to do the same as this.

https://github.com/ansible/awx/blob/ef29589940bb071b802671f090158f0916f9c260/awx/main/migrations/_rbac.py#L12-L38

That's called in other migrations. I could see the argument that it slows down upgrades (whereas we would only encounter the other call with a new database, which will be quick). But even so, I would be stubborn and ask that you implement this _via_ that method, maybe you pass in the resource list which will just be `["Organization"]` here.

jbradberry:
@AlanCoding all right, I've switched to using the `create_roles` function.  It doesn't sit well with me, though, and long-term _I'd_ like to see this machinery stripped out and a simpler system put in place.

AlanCoding (approved):

CFSNM (approved):

Issue #13301, opened 2022-12-07
ValueError: Duplicated timeseries in CollectorRegistry


anxstj:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

While upgrading my AWX system from 21.5.0 to 21.10.0 the following error occurred:

```
tools_awx_1 | ValueError: Duplicated timeseries in CollectorRegistry: {'awx01_example_com_messages_received'}
tools_awx_1 | 2022-12-07 16:21:09,927 INFO exited: wsbroadcast (exit status 1; not expected)
tools_awx_1 | 2022-12-07 16:21:09,927 INFO exited: wsbroadcast (exit status 1; not expected)
```

It seems that the wsbroadcast process is dying with this error and finally the tools_awx_1 container dies as well.

Changing the required *prometheus-client* back to 0.7.1 fixes (or workarounds) this error.

```
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -253,7 +253,7 @@ pexpect==4.7.0
     #   ansible-runner
 pkgconfig==1.5.5
     # via -r /awx_devel/requirements/requirements.in
-prometheus-client==0.15.0
+prometheus-client==0.7.1
     # via -r /awx_devel/requirements/requirements.in
 psutil==5.9.4
     # via -r /awx_devel/requirements/requirements.in
```

This might be related to: https://github.com/prometheus/client_python/issues/815

### AWX version

21.10.0

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

docker development environment

### Modifications

no

### Ansible version

2.12.2

### Operating system

Debian 11

### Web browser

Firefox

### Steps to reproduce

I have a setup with two controller nodes, awx01.example.com and awx02.example.com. The error message mentions always the other node. To reproduce the problem at least two controller nodes are needed, I guess.

### Expected results

The Prometheus client is registering the nodes properly.

### Actual results

```
tools_awx_1 | Adding {'awx01.example.com'} to websocket broadcast list
tools_awx_1 | Traceback (most recent call last):
[...]

PR #13311, opened 2022-12-08
Fix duped stats name and Redis for wsbroadcast

relrod:
##### SUMMARY

This fixes several things related to our wsbroadcast stats handling. This was found during the ongoing wsrelay work.

There are really three fixes here:

- Logging was not actually enabled for the analytics.broadcast_websocket module, so that has been added to our loggers config.

- analytics.broadcast_websocket was not actually able to connect to Redis due to 68614b83c00982259be18b0e6e22550f8dd0449a as part of the work in #13187. But there was no easy way to know this because the logging issue meant no exceptions showed up anywhere reasonable.

- Relatedly, and also as part of #13187, we jumped from `prometheus-client` 0.7.1 up to 0.15.0. This included a breaking change where a `Counter` ending with `_total` will clash with a `Gauge` of the same name but without `_total`. I am not 100% sure of the reasoning here, other than "OpenMetrics compatibility".

Refs #13301
Refs #13187

---------

With these changes, `awx-manage run_wsbroadcast --status` still works:

image

------

An easy reproducer for the third bullet point above is this, for anyone who wishes to play around with it some more:

```python
In [1]: from prometheus_client import (
   ...:     Gauge,
   ...:     Counter,
   ...:     CollectorRegistry,
   ...: )

In [2]: %xmode minimal
Exception reporting mode: Minimal

In [3]: registry = CollectorRegistry()

In [4]: foo = Counter('foo_bar_total', 'some description', registry=registry)

In [5]: bar = Gauge('foo_bar', 'another description', registry=registry)
ValueError: Duplicated timeseries in CollectorRegistry: {'foo_bar'}
```

----

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/analytics/broadcast_websocket.py
+++ b/awx/main/analytics/broadcast_websocket.py
@@ -2,6 +2,7 @@ import datetime
 import asyncio
 import logging
 import redis
+import redis.asyncio
 import re
 
 from prometheus_client import (
@@ -81,7 +82,7 @@ class BroadcastWebsocketStatsManager:
 
     async def run_loop(self):
         try:
-            redis_conn = await redis.asyncio.create_redis_pool(settings.BROKER_URL)
+            redis_conn = await redis.asyncio.Redis.from_url(settings.BROKER_URL)
             while True:
                 stats_data_str = ''.join(stat.serialize() for stat in self._stats.values())
                 await redis_conn.set(self._redis_key, stats_data_str)
@@ -121,8 +122,8 @@ class BroadcastWebsocketStats:
             'Number of messages received, to be forwarded, by the broadcast websocket system',
             registry=self._registry,
         )
-        self._messages_received = Gauge(
-            f'awx_{self.remote_name}_messages_received',
+        self._messages_received_current_conn = Gauge(
+            f'awx_{self.remote_name}_messages_received_currrent_conn',
             'Number forwarded messages received by the broadcast websocket system, for the duration of the current connection',
             registry=self._registry,
         )
@@ -143,13 +144,13 @@ class BroadcastWebsocketStats:
 
     def record_message_received(self):
         self._internal_messages_received_per_minute.record()
-        self._messages_received.inc()
+        self._messages_received_current_conn.inc()
         self._messages_received_total.inc()
 
     def record_connection_established(self):
         self._connection.state('connected')
         self._connection_start.set_to_current_time()
-        self._messages_received.set(0)
+        self._messages_received_current_conn.set(0)
 
     def record_connection_lost(self):
         self._connection.state('disconnected')
--- a/awx/settings/defaults.py
+++ b/awx/settings/defaults.py
@@ -853,6 +853,7 @@ LOGGING = {
         'awx.main.signals': {'level': 'INFO'},  # very verbose debug-level logs
         'awx.api.permissions': {'level': 'INFO'},  # very verbose debug-level logs
         'awx.analytics': {'handlers': ['external_logger'], 'level': 'INFO', 'propagate': False},
+        'awx.analytics.broadcast_websocket': {'handlers': ['console', 'file', 'wsbroadcast', 'external_logger'], 'level': 'INFO', 'propagate': False},
         'awx.analytics.performance': {'handlers': ['console', 'file', 'tower_warnings', 'external_logger'], 'level': 'DEBUG', 'propagate': False},
         'awx.analytics.job_lifecycle': {'handlers': ['console', 'job_lifecycle'], 'level': 'DEBUG', 'propagate': False},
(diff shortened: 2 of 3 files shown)


awx/main/analytics/broadcast_websocket.py
```diff
@@ -121,8 +122,8 @@ def __init__(self, local_hostname, remote_hostname):
             'Number of messages received, to be forwarded, by the broadcast websocket system',
             registry=self._registry,
         )
-        self._messages_received = Gauge(
-            f'awx_{self.remote_name}_messages_received',
+        self._messages_received_current_conn = Gauge(
+            f'awx_{self.remote_name}_messages_received_currrent_conn',
```
relrod, line 126: This part addresses bullet point 3

awx/main/analytics/broadcast_websocket.py
```diff
@@ -81,7 +82,7 @@ def delete_remote_host_stats(self, remote_hostname):
 
     async def run_loop(self):
         try:
-            redis_conn = await redis.asyncio.create_redis_pool(settings.BROKER_URL)
+            redis_conn = await redis.asyncio.Redis.from_url(settings.BROKER_URL)
```
relrod, line 85: This part addresses bullet point 2

awx/settings/defaults.py
```diff
@@ -853,6 +853,7 @@
         'awx.main.signals': {'level': 'INFO'},  # very verbose debug-level logs
         'awx.api.permissions': {'level': 'INFO'},  # very verbose debug-level logs
         'awx.analytics': {'handlers': ['external_logger'], 'level': 'INFO', 'propagate': False},
+        'awx.analytics.broadcast_websocket': {'handlers': ['console', 'file', 'wsbroadcast', 'external_logger'], 'level': 'INFO', 'propagate': False},
```
relrod, line 856: This part addresses bullet point 1.

One thing I'm not positive on: Are the `handlers` right here?

shanemcd: Looks right to me.

shanemcd (approved):
Amazing work my friend.

fosterseth (approved):

awx/settings/defaults.py
AlanCoding: hmm... this isn't quite jiving for me.

https://github.com/ansible/awx/blob/ef29589940bb071b802671f090158f0916f9c260/awx/main/conf.py#L572

The external logging setting allows you to select which loggers to enable. So if you logged to a new `awx.analytics.____`, then yes, it makes sense that we would miss it. Unless we had `awx.analytics` enabled by default and the new loggers propagated (which they did before, but won't now).

I don't think we can figure this out completely without also looking at the dynamic logic there is for the external logger. But I admit I don't fully understand the fix.... although if you added it to the default loggers list, that would make sense to me (with the incompletely information I have now).

AlanCoding: This clears up the rest for me.

https://github.com/ansible/awx/blob/ef29589940bb071b802671f090158f0916f9c260/awx/main/utils/filters.py#L94-L100

A few takeaways:
 - your fix still appears _incomplete_, you added "broadcast_websocket" here, but it also needs to be listed among the default external loggers
 - anything sent to `awx.analytics` is just plain wrong. That implies that nothing should propagate down to that logger

This latter point is... oof. To prevent regressions of this type, we should enforce that nothing ever logs to an un-documented sub-logger of `awx.analytics`. Maybe that can be done with a unit test?

AlanCoding: Here we go, this can't be used to fix the problem, but it encapsulates the issue.

```
In [7]: [k for k in logging.Logger.manager.__dict__['loggerDict'].keys() if k.startswith('awx.analytics')]
Out[7]: 
['awx.analytics',
 'awx.analytics.performance',
 'awx.analytics.job_lifecycle',
 'awx.analytics.job_events',
 'awx.analytics.system_tracking',
 'awx.analytics.activity_stream']

In [8]: from awx.main.analytics import broadcast_websocket

In [9]: [k for k in logging.Logger.manager.__dict__['loggerDict'].keys() if k.startswith('awx.analytics')]
Out[9]: 
['awx.analytics',
 'awx.analytics.performance',
 'awx.analytics.job_lifecycle',
 'awx.analytics.job_events',
 'awx.analytics.system_tracking',
 'awx.analytics.activity_stream',
 'awx.analytics.broadcast_websocket']
```

We can tell via inspection of the global python objects which loggers have been created (at least on import... code can always follow non-standard practice). But that doesn't work unless everything in our app has been imported, which is hard to guarantee. So if you could do a hack job to accomplish that goal, fixing it would be possible by this approach.

Issue #13245, opened 2022-11-29
Update docs for current task manager changes


AlanCoding:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

Enhancement to Existing Feature

### Feature Summary

Our docs folder needs some updating for changes that have been happening in the task manager, and changes currently in progress. Specifically talking about the folder here:

https://github.com/ansible/awx/tree/devel/docs

Giving a status roll call here:

 - :yellow_circle:  `clustering.md`
   - should remove or amend references to old versions
   - original description of HA should mention that it is about control plane
   - remove references to "setup playbook", as it's more complicated now
   - remove celery
   - Preferred Instance Group resource list is listed twice... possibly a little wrong too
 - :yellow_circle:  `capacity.md`
   - needs added content for https://github.com/ansible/awx/pull/13099
 - :red_circle:  `task_manager_system.md`
   - Needs mention of DependencyManager and WorkflowManager, or complete rewrite
   - should mentioned `dependencies_processed` along with job states
   - from refactor work, we should also more coherently present the stages of processing, separate blocking rules from the update-on-launch dependencies

### Select the relevant components

- [ ] UI
- [ ] API
- [X] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

N/A, docs issue

### Current results

:yellow_circle: docs I'm saying have distracting content, and/or notable omissions from recent features. Otherwise, there is valid documentation of existing features.

:red_circle: docs I'm saying we should consider deleting if not substantially rewritten or updated

### Sugested feature result

N/A

AlanCoding:
closing as done

PR #13315, opened 2022-12-08
update task manager docs after refactoring

fosterseth:
##### SUMMARY

issue https://github.com/ansible/awx/issues/13245

view rendered doc here https://github.com/fosterseth/awx/blob/update_task_manager_md/docs/task_manager_system.md

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 - Docs

### DIFF
--- a/docs/task_manager_system.md
+++ b/docs/task_manager_system.md
@@ -1,18 +1,108 @@
-# Task Manager Overview
+# Task Manager System Overview
 
-The task manager is responsible for deciding when jobs should be scheduled to run. When choosing a task to run, the considerations are:
+The task management system is made up of three separate components:
+1. Dependency Manager
+2. Task Manager
+3. Workflow Manager
+
+Each of these run in a separate dispatched task and can run at the same time as one another.
+
+This system is responsible for deciding when tasks should be scheduled to run. When choosing a task to run, the considerations are:
 1. Creation time
 2. Job dependencies
 3. Capacity
 
-Independent jobs are run in order of creation time, earliest first. Jobs with dependencies are also run in creation time order within the group of job dependencies. Capacity is the final consideration when deciding to release a job to be run by the task dispatcher.
+Independent tasks are run in order of creation time, earliest first. Tasks with dependencies are also run in creation time order within the group of task dependencies. Capacity is the final consideration when deciding to release a task to be run by the dispatcher.
 
-## Task Manager Architecture
 
-The task manager has a single entry point, `Scheduler().schedule()`. The method may be called in parallel, at any time, as many times as the user wants. The `schedule()` function tries to acquire a single, global lock using the Instance table first recorded in the database. If the lock cannot be acquired, the method returns. The failure to acquire the lock indicates that there is another instance currently running `schedule()`.
+## Dependency Manager
+
+Responsible for looking at each pending task and determining whether it should create a dependency for that task.
+For example, if `update_on_launch` is enabled of a task, a project update will be created as a dependency of that task. The Dependency Manager is responsible for creating that project update.
+
+Dependencies can also have their own dependencies, for example,
+
+```
++-----------+
+|           |                          created by web API call
+|   Job A   |
+|           |
++-----------+---+
+                |
+                |
+        +-------v----+
+        |  Inventory |                 dependency of Job A
+        |  Source    |                 created by Dependency Manager
+        |  Update B  |
+        +------------+-------+
+                             |
+                             |
+                      +------v------+
+                      |   Project   |  dependency of Inventory Source Update B
+                      |   Update C  |  created by Dependency Manager
+                      +-------------+
+```
+
+
+### Dependency Manager Steps
+
+1. Get pending tasks (parent tasks) that have `dependencies_processed = False`
+2. Create project update if
+    a. not already created
+    b. last project update outside of cache timeout window
+3. Create inventory source update if
+    a. not already created
+    b. last inventory source update outside of cache timeout window
+4. Check and create dependencies for these newly created dependencies
+    a. inventory source updates can have a project update dependency
+5. All dependencies are linked to the parent task via the `dependent_jobs` field
+    a. This allows us to cancel the parent task if the dependency fails or is canceled
+6. Update the parent tasks with `dependencies_processed = True`
+
+
+## Task Manager
+
+Responsible for looking at each pending task and determining whether Task Manager can start that task.
+
+### Task Manager Steps
+
+1. Get pending, waiting, and running tasks that have `dependencies_processed = True`
+2. Before processing pending tasks, the task manager first processes running tasks. This allows it to build a dependency graph and account for the currently consumed capacity in the system.
+    a. dependency graph is just an internal data structure that tracks which jobs are currently running. It also handles "soft" blocking logic
+    b. the capacity is tracked in memory on the `TaskManagerInstances` and `TaskManagerInstanceGroups` objects which are  in-memory representations of the instances and instance groups. These data structures are used to help track what consumed capacity will be as we decide that we will start new tasks, and until such time that we actually commit the state changes to the database.
+3. For each pending task:
+    a. Check if total number of tasks started on this task manager cycle is > `start_task_limit`
+    b. Check if [timed out](#Timing Out)
+    b. Check if task is blocked
+    c. Check if preferred instances have enough capacity to run the task
+4. Start the task by changing status to `waiting` and submitting task to dispatcher
+
+
+## Workflow Manager
+
+Responsible for looking at each workflow job and determining if next node can run
+
+### Worflow Manager Steps
+
+1. Get all running workflow jobs
+2. Build up a workflow DAG for each workflow job
+3. For each workflow job:
+    a. Check if [timed out](#Timing Out)
+    b. Check if next node can start based on previous node status and the associated success / failure / always logic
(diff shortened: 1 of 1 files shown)


docs/task_manager_system.md
```diff
@@ -1,33 +1,113 @@
-# Task Manager Overview
+# Task Manager System Overview
 
-The task manager is responsible for deciding when jobs should be scheduled to run. When choosing a task to run, the considerations are:
+Made up of three separate components
```
kdelee:
```suggestion
The task management system is made up of three separate components:
```

docs/task_manager_system.md
```diff
@@ -1,33 +1,113 @@
...
+
 
-## Task Manager Architecture
+### Task Manager Algorithm
+
+1. Get pending, waiting, and running tasks that have `dependencies_processed = True`
+2. For each running task, add task to dependency graph
+    a. dependency graph is just an internal data structure that tracks which
+    jobs are currently running. It also handles "soft" blocking logic
+3. Check if # tasks started on this task manager cycle is > `start_task_limit`
```
kdelee: missing also checking the task manager time out here

fosterseth: oh right, i'll add that in

docs/task_manager_system.md
```diff
@@ -1,33 +1,113 @@
...
+Responsible for looking at each pending task and determining whether it can start that task.
+
 
-## Task Manager Architecture
+### Task Manager Algorithm
+
+1. Get pending, waiting, and running tasks that have `dependencies_processed = True`
+2. For each running task, add task to dependency graph
+    a. dependency graph is just an internal data structure that tracks which
+    jobs are currently running. It also handles "soft" blocking logic
```
kdelee: in same loop that we add tasks to dependency graph, we do the initial calculation of currently consumed capacity

```
 Before processing pending tasks, the task manager first processes running tasks. This allows it to build a dependency graph and account for the currently consumed capacity in the system.
    a. dependency graph is just an internal data structure that tracks which
    jobs are currently running. It also handles "soft" blocking logic
    b. the capacity is tracked in memory on the TaskManagerInstances and
        TaskManagerInstanceGroups objects which are  in-memory representations of the instances
        and instance groups. These data structures are used to help track what consumed capacity will
        be as we decide that we will start new tasks, and until such time that we actually commit
        the state changes to the database.
```

kdelee (commented):
left a few suggestions

docs/task_manager_system.md
fosterseth: updated the algorithm and added a `Timing Out` section

docs/task_manager_system.md
fosterseth: added, thanks

docs/task_manager_system.md
```diff
@@ -1,38 +1,127 @@
...
+                             |
+                             |
+                      +------v------+
+                      |   Project   |  dependency of Inventory Source Update B
+                      |   Update C  |  created by Dependency Manager
+                      +-------------+
+```
+
+
+### Dependency Manager Algorithm
```
gundalow: Is `Algorithm` the right term here, or is this a list of steps that are run through in order?

docs/task_manager_system.md
```diff
@@ -1,38 +1,127 @@
...
 
 
 ### Job Lifecycle
 
 | Job Status |                                                       State                                                      |
-|:----------:|:------------------------------------------------------------------------------------------------------------------:|
+|:-----------|:-------------------------------------------------------------------------------------------------------------------|
 | pending    | Job has been launched.  <br>1. Hasn't yet been seen by the scheduler <br>2. Is blocked by another task <br>3. Not enough capacity |
 | waiting    | Job published to an AMQP queue.
```
AlanCoding: does pg_notify qualify as an AMQP queue? I don't think so?

docs/task_manager_system.md
```diff
@@ -1,38 +1,127 @@
...
+4. Check and create dependencies for these newly created dependencies
+    a. inventory source updates can have a project update dependency
+5. All dependencies are linked to the parent task via the `dependent_jobs` field
+    a. This allows us to cancel the parent task if the dependency fails or is canceled
+6. Update the parent tasks with `dependencies_processed = True`
+
+
+## Task Manager
+
+Responsible for looking at each pending task and determining whether it can start that task.
```
gundalow: Shouldn't use `it` as it's ambiguous, do we mean:

```suggestion
Responsible for looking at each pending task and determining whether Task Manager can start that task.
```

docs/task_manager_system.md
```diff
@@ -1,38 +1,127 @@
...
 
-## Task Manager Architecture
+### Task Manager Algorithm
```
gundalow: As above, if this is steps, then do we mean steps?

docs/task_manager_system.md
```diff
@@ -1,38 +1,127 @@
...
+
+1. Get pending, waiting, and running tasks that have `dependencies_processed = True`
+2. Before processing pending tasks, the task manager first processes running tasks. This allows it to build a dependency graph and account for the currently consumed capacity in the system.
```
gundalow, line 70: Do all running tasks have to be processed before new pending tasks are processed?

Do we need to define "processing"?

docs/task_manager_system.md
```diff
@@ -1,38 +1,127 @@
...
+    a. dependency graph is just an internal data structure that tracks which jobs are currently running. It also handles "soft" blocking logic
+    b. the capacity is tracked in memory on the `TaskManagerInstances` and `TaskManagerInstanceGroups` objects which are  in-memory representations of the instances and instance groups. These data structures are used to help track what consumed capacity will be as we decide that we will start new tasks, and until such time that we actually commit the state changes to the database.
+3. For each pending task:
+    a. Check if # tasks started on this task manager cycle is > `start_task_limit`
```
gundalow: Define `#` (if we mean number or ID, they say that)

docs/task_manager_system.md
```diff
@@ -1,38 +1,127 @@
...
+2. Build up a workflow DAG for each workflow job
+3. For each workflow job:
+    a. Check if [timed out](#Timing Out)
+    b. Check if next node can start based on previous node status and the associated success / failure / always logic
+4. Create new task and signal start
+
+
+## Task Manager System Architecture
+
+Each of the three managers has a single entry point, `schedule()`. The method may be called in parallel, at any time, as many times as the user wants. The `schedule()` function tries to acquire a single, global lock using the Instance table first recorded in the database. If the lock cannot be acquired, the method returns. The failure to acquire the lock indicates that there is another instance currently running `schedule()`.
```
gundalow: "as many times as the user wants" what's the use-case for multiple parallel runs?

docs/task_manager_system.md
```diff
@@ -46,19 +135,20 @@ Empirically, the periodic task manager has been effective in the past and will c
...
 
 
 ### Blocking Logic
 
 The blocking logic is handled by a mixture of ORM instance references and task manager local tracking data in the scheduler instance.
 
+There is a distinction between so-called "hard" vs "soft" blocking.
 
-## Acceptance Tests
+"Hard" refers to dependencies that are represented in the database via the task `dependent_jobs` field. That is, Job A will not run if any of its `dependent_jobs` are still running.
```
gundalow:
```suggestion
**Hard blocking** refers to dependencies that are represented in the database via the task `dependent_jobs` field. That is, Job A will not run if any of its `dependent_jobs` are still running.
```

murakobardenas (approved):

docs/task_manager_system.md
```diff
@@ -46,19 +135,20 @@ Empirically, the periodic task manager has been effective in the past and will c
...
 
-The new task manager should, in essence, work like the old one. Old task manager features were identified while new ones were discovered in the process of creating the new task manager. Rules for the new task manager behavior are iterated below; testing should ensure that those rules are followed.
+"Soft" refers to blocking logic that doesn't have a database representation. Imagine Job A and B are both based on the same job template, and concurrent jobs is `disabled`. Job B will be blocked from running if Job A is already running. This is determined purely by the task manager tracking running jobs via the Dependency Graph.
```
gundalow:
```suggestion
**Soft blocking** refers to blocking logic that doesn't have a database representation. Imagine Job A and B are both based on the same job template, and concurrent jobs is `disabled`. Job B will be blocked from running if Job A is already running. This is determined purely by the task manager tracking running jobs via the Dependency Graph.
```


Can we reword "database representation"

murakobardenas (approved):

murakobardenas (approved):

docs/task_manager_system.md
fosterseth: yeah this was inaccurate language. remove it and added a `bulk reschedule` section that addresses this topic

docs/task_manager_system.md
fosterseth: fixed

docs/task_manager_system.md
fosterseth: using `steps` now

docs/task_manager_system.md
fosterseth: changed to
`Job submitted to dispatcher via pg_notify`

AlanCoding (approved):

PR #13316, opened 2022-12-08
SAML enhancements

john-westcott-iv:
##### SUMMARY

Split out SAML and Social Auth pipelines.
Pull common bits between SAML and LDAP auth into common file.

##### ISSUE TYPE
 
 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.9.1.dev65+g7ac513fa18.d20221208
```

##### ADDITIONAL INFORMATION

```

```


AlanCoding:
@kdelee did you do some performance testing of this PR? Could you give like a really really brief summary? Because right now this just looks like a refactor.

awx/sso/backends.py
```diff
@@ -389,43 +385,16 @@ def on_populate_user(sender, **kwargs):
...
-                new_org = Organization.objects.create(name=org_name)
-            except IntegrityError:
-                # Another thread must have created this org before we did so now we need to get it
-                new_org = Organization.objects.get(name=org_name)
-            # Add the org name to the existing orgs since we created it and we may need it to build the teams below
-            existing_orgs[org_name] = new_org.id
-
-    # Do the same for teams
-    existing_team_names = list(Team.objects.all().values_list('name', flat=True))
+    orgs_list = set(list(org_map.keys()) + [item.get('organization', None) for item in team_map.values()])
```
jjwatt: This looks like a `TypeError`. `set() + list()` Is there a unit test for `on_populate_user`?

john-westcott-iv: Good catch, I don't think there is for backends.

kdelee (approved):
image

100 times faster in worst case
10 times faster in best case

SHIP IT!

awx/sso/common.py
```diff
@@ -0,0 +1,171 @@
...
+            try:
+                Team.objects.create(name=team_name, organization_id=existing_orgs[team_map[team_name]])
+            except IntegrityError:
+                # If another process got here before us that is ok because we don't need the ID from this team or anything
+                pass
+    # End move some day
+    # ==============================================================================================================
+
+
+def get_or_create_with_default_galaxy_cred(**kwargs):
```
AlanCoding: I might prefer to title it `get_or_create_org_...`

awx/sso/social_pipeline.py
```diff
@@ -0,0 +1,90 @@
...
+        return
+
+    org_map = backend.setting('ORGANIZATION_MAP') or {}
+    for org_name, org_opts in org_map.items():
+        organization_alias = org_opts.get('organization_alias')
+        if organization_alias:
+            organization_name = organization_alias
+        else:
+            organization_name = org_name
+        org = get_or_create_with_default_galaxy_cred(name=organization_name)
```
AlanCoding: huh, I'm surprised that we don't use `defaults` already, like, set the description of the org to "imported via SAML"

john-westcott-iv: Its just a straight building of the org. This will be a slight change to how the LDAP adapter works because it didn't used to include adding the default galaxy credential at all.

awx/sso/common.py
john-westcott-iv: Good call on making this more descriptive since this function has moved to common.

awx/sso/common.py
```diff
@@ -0,0 +1,171 @@
...
+    #
+    # org_list is a set of organization names
+    # team_map is a dict of {<team_name>: <org name>}
+    #
+    # Move this junk into save of the settings for performance later, there is no need to do that here
+    #    with maybe the exception of someone defining this in settings before the server is started?
+    # ==============================================================================================================
+
+    if not can_create:
+        logger.debug(f"Adapter {adapter} is not allowed to create orgs/teams")
```
gamuniz, line 113: in what scenario would this condition be hit? saml auto create disabled? If so, shouldn't we make the logging message indicate that incase the user is unaware of the setting or has been changed?

john-westcott-iv: The SAML adapter has a setting that says whether or not it its actually allowed to create orgs/teams. I don't think there is an equivalent for LDAP. I don't think we used to do anything before (besides skip the task) if the setting was off for SAML. So this log would just show up in the debug logs as an FYI for an admin.

gamuniz: Makes sense but maybe we should include help text to how to enable it?

john-westcott-iv: It should already be there:
image

I suppose we could do something like `Adapter {adapter} not not allowed to create orgs/team see documentation.` or somehow try and key off `{adapter}` to get more specific but that seems like overkill to me. What do you think?

gamuniz: see docs is probably enough

gamuniz (approved):

jay-steurer (approved):

john-westcott-iv:
@tvo318 this PR has a minor change to the behavior of the LDAP adapter. It now shares common code with the SAML adapter so if an LDAP login creates an organization the default galaxy credential will be added to that organization.

PR #13326, opened 2022-12-13
Handle canceled and timeout status in callback

jangel97:
This is one of the hotfixes that helped to improve performance. Backporting it to AWX.


fosterseth:
thanks @jangel97 

we had a different solution for the upstream case that involved an ansible runner change https://github.com/ansible/ansible-runner/pull/1133

so this change shouldn't be needed for upstream AWX

Issue #13191, opened 2022-11-14
Tracking: Image build changes for rsyslog decoupling


relrod:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

Enhancement to Existing Feature

### Feature Summary

We need to modify our current image build process a bit (both for the docker-compose dev env and the kube image) to account for the [rsyslog decoupling proposal](https://github.com/ansible/awx/issues/13183).

- [ ] We need to pull out the rsyslog supervisor configuration section into its own supervisor file (`supervisor_rsyslog.conf` or similar), similar to `supervisor_task.conf`
- [ ] We need a new `launch_awx_rsyslog.sh` added to the container, similar to `launch_awx_task.sh`
- [ ] Later on, once the new daemon exists, it will need to be added to the newly created `supervisor_rsyslog.conf`
- [ ] ... and to `tools/docker-compose/supervisor.conf` for the dev environment

The full completion of this task depends on #13192 being done, but there is some work that can be done before that.

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [X] Other

### Steps to reproduce

N/A

### Current results

N/A

### Sugested feature result

N/A

### Additional information

N/A

djyasin:
PR: https://github.com/ansible/awx/pull/13327

PR #13327, opened 2022-12-13
Created new rsyslog launch file.

djyasin:
##### SUMMARY

refs https://github.com/ansible/awx/issues/13191

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.8.1
```

### DIFF
--- a/tools/ansible/roles/dockerfile/files/launch_awx_rsyslog.sh
+++ b/tools/ansible/roles/dockerfile/files/launch_awx_rsyslog.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+if [ `id -u` -ge 500 ]; then
+    echo "awx:x:`id -u`:`id -g`:,,,:/var/lib/awx:/bin/bash" >> /tmp/passwd
+    cat /tmp/passwd > /etc/passwd
+    rm /tmp/passwd
+fi
+
+if [ -n "${AWX_KUBE_DEVEL}" ]; then
+    pushd /awx_devel
+    make awx-link
+    popd
+
+    export SDB_NOTIFY_HOST=$MY_POD_IP
+fi
+
+set -e
+
+wait-for-migrations
+
+# This file will be re-written when the dispatcher calls reconfigure_rsyslog(),
+# but it needs to exist when supervisor initially starts rsyslog to prevent the
+# container from crashing. This was the most minimal config I could get working.
+cat << EOF > /var/lib/awx/rsyslog/rsyslog.conf
+action(type="omfile" file="/dev/null")
+EOF
+
+exec supervisord -c /etc/supervisord.conf
--- a/tools/ansible/roles/dockerfile/files/launch_awx_rsyslog.sh
+++ b/tools/ansible/roles/dockerfile/files/launch_awx_rsyslog.sh
@@ -24,9 +24,5 @@ cat << EOF > /var/lib/awx/rsyslog/rsyslog.conf
 action(type="omfile" file="/dev/null")
 EOF
 
-#using this to start child process rsyslog and rsyslog configurer 
-#How do I tell supervisord to launch rsyslog 
-#supervisord -c  tells supervisor d what config file to use 
-#need an rsyslog config file 
 exec supervisord -c /etc/supervisor_rsyslog.conf
 
--- a/tools/ansible/roles/dockerfile/files/launch_awx_rsyslog.sh
+++ b/tools/ansible/roles/dockerfile/files/launch_awx_rsyslog.sh
@@ -24,4 +24,9 @@ cat << EOF > /var/lib/awx/rsyslog/rsyslog.conf
 action(type="omfile" file="/dev/null")
 EOF
 
-exec supervisord -c /etc/supervisord.conf
+#using this to start child process rsyslog and rsyslog configurer 
+#How do I tell supervisord to launch rsyslog 
+#supervisord -c  tells supervisor d what config file to use 
+#need an rsyslog config file 
+exec supervisord -c /etc/supervisor_rsyslog.conf
+
--- a/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
+++ b/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
@@ -217,7 +217,7 @@ ADD tools/ansible/roles/dockerfile/files/launch_awx_task.sh /usr/bin/launch_awx_
 ADD tools/ansible/roles/dockerfile/files/uwsgi.ini /etc/tower/uwsgi.ini
 ADD {{ template_dest }}/supervisor.conf /etc/supervisord.conf
 ADD {{ template_dest }}/supervisor_task.conf /etc/supervisord_task.conf
-ADD {{ template_date }}/supervisor_rsyslog.conf /etc/supervisor_rsyslog.conf
+ADD {{ template_dest }}/supervisor_rsyslog.conf /etc/supervisor_rsyslog.conf
 {% endif %}
 {% if (build_dev|bool) or (kube_dev|bool) %}
 ADD tools/docker-compose/awx.egg-link /tmp/awx.egg-link
--- a/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
+++ b/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
@@ -2,7 +2,7 @@
 ### tools/ansible/roles/dockerfile/templates/Dockerfile.j2
 ###
 ### DO NOT EDIT
-###
+###t
 
 # Build container
 FROM quay.io/centos/centos:stream9 as builder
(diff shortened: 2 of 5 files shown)


tools/ansible/roles/dockerfile/templates/Dockerfile.j2
```diff
@@ -199,6 +199,10 @@ ADD tools/ansible/roles/dockerfile/files/rsyslog.conf /var/lib/awx/rsyslog/rsysl
 ADD tools/ansible/roles/dockerfile/files/wait-for-migrations /usr/local/bin/wait-for-migrations
 ADD tools/ansible/roles/dockerfile/files/stop-supervisor /usr/local/bin/stop-supervisor
 
+# Create default awx rsyslog sidecar
```
TheRealHaoLiu: this does not create the side car 

a "sidecar" is essentially another container that's deployed in the same pod as your "main" container

https://docs.docker.com/engine/reference/builder/

this is the dockerfile documentation and ADD adds a file to your docker container image 

this statement below means "add the supervisor_rsyslog.conf.j2 template to the container image and renamed it to /var/run/supervisor/supervisor.rsyslog.sock"

lets talk about what you are trying to do with this statement

TheRealHaoLiu: for additional context

dockerfile builds docker container **image** 
when you create a container via kubernetes deployment it uses the image as basis to start the container

tools/ansible/roles/dockerfile/templates/Dockerfile.j2
```diff
@@ -216,6 +220,7 @@ ADD tools/ansible/roles/dockerfile/files/launch_awx_task.sh /usr/bin/launch_awx_
 ADD tools/ansible/roles/dockerfile/files/uwsgi.ini /etc/tower/uwsgi.ini
 ADD {{ template_dest }}/supervisor.conf /etc/supervisord.conf
 ADD {{ template_dest }}/supervisor_task.conf /etc/supervisord_task.conf
+ADD {{ template_dest }}/supervisor_rsyslog.conf.j2 /etc/supervisor_rsyslog.conf.j2
```
TheRealHaoLiu: you should not be adding the template files to the container image

tools/ansible/roles/dockerfile/templates/Dockerfile.j2
```diff
@@ -233,6 +238,7 @@ RUN for dir in \
       /var/log/nginx \
       /var/lib/postgresql \
       /var/run/supervisor \
+      /var/run/supervisor_rsyslog \
```
TheRealHaoLiu: what is the intention for this?

tools/ansible/roles/dockerfile/templates/Dockerfile.j2
TheRealHaoLiu: https://docs.docker.com/engine/reference/builder/#add

djyasin: Thank you so much for this context! I reviewed the documentation and it looks like I need to tell it where it will render: `The ADD instruction copies new files, directories or remote file URLs from <src> and adds them to the filesystem of the image at the path <dest>.`

djyasin: Actually, upon further investigation, I don't think I need this at all.

tools/ansible/roles/dockerfile/templates/Dockerfile.j2
djyasin: I thought that this would lay the foundation for the supervisor_rsyslog, but based on the provided documentation I don't think this is correct.

tools/ansible/roles/dockerfile/templates/Dockerfile.j2
TheRealHaoLiu: dockerfile does not render j2 ansible render j2 template to the config file

djyasin: Thank you so much for your reply! I think it is in order now.

tools/ansible/roles/dockerfile/templates/Dockerfile.j2
```diff
@@ -199,6 +199,7 @@ ADD tools/ansible/roles/dockerfile/files/rsyslog.conf /var/lib/awx/rsyslog/rsysl
 ADD tools/ansible/roles/dockerfile/files/wait-for-migrations /usr/local/bin/wait-for-migrations
 ADD tools/ansible/roles/dockerfile/files/stop-supervisor /usr/local/bin/stop-supervisor
 
+
```
relrod, line 200: Nix

tools/ansible/roles/dockerfile/files/launch_awx_rsyslog.sh
```diff
@@ -0,0 +1,32 @@
...
+# but it needs to exist when supervisor initially starts rsyslog to prevent the
+# container from crashing. This was the most minimal config I could get working.
+cat << EOF > /var/lib/awx/rsyslog/rsyslog.conf
+action(type="omfile" file="/dev/null")
+EOF
+
+#using this to start child process rsyslog and rsyslog configurer 
+#How do I tell supervisord to launch rsyslog 
+#supervisord -c  tells supervisor d what config file to use 
+#need an rsyslog config file
```
relrod: I would nix or clean up these comments a bit.

tools/ansible/roles/dockerfile/templates/supervisor_rsyslog.conf.j2
```diff
@@ -0,0 +1,59 @@
...
+startsecs = 30
+stopasgroup=true
+killasgroup=true
+stdout_logfile=/dev/stdout
+stdout_logfile_maxbytes=0
+stderr_logfile=/dev/stderr
+stderr_logfile_maxbytes=0
+
+[group:tower-processes]
+programs=awx-rsyslog-configurer, awx-rsyslogd
```
relrod: Maybe drop the space here for consistency with the other files.

```suggestion
programs=awx-rsyslog-configurer,awx-rsyslogd
```

tools/ansible/roles/dockerfile/files/launch_awx_rsyslog.sh
```diff
@@ -0,0 +1,32 @@
...
+set -e
+
+wait-for-migrations
+
+# This file will be re-written when the dispatcher calls reconfigure_rsyslog(),
+# but it needs to exist when supervisor initially starts rsyslog to prevent the
+# container from crashing. This was the most minimal config I could get working.
+cat << EOF > /var/lib/awx/rsyslog/rsyslog.conf
+action(type="omfile" file="/dev/null")
+EOF
```
relrod, line 25: This seems okay to me still, after checking [Jessica's PR](
https://github.com/ansible/awx/pull/13222/files#diff-017d6c505750589da91821bb9b0e8319dca2450aba649dd55c921ff20dcb6b3aR24-R25). When the configurer starts up, it'll do an initial `reconfigure_rsyslog()` before it starts blocking and waiting to be told to reconfigure it in the future, so this minimal config shouldn't last very long. :+1:

relrod (commented):
Excellent work here, it's cool to see this coming together.

A few nitpick inlines. Nothing blocking there.

But I think we also need to remove the current rsyslog stuff that this is replacing, that currently lives in the web container.

Looking at the second diagram on #13183 - at the end of this first phase, we ultimately will have created a sidecar container in the current deployment (this part is kind of independent of the web/task split deployments). That is what @thedoubl3j is working on.

That new container will run what you've done in this PR, making the current rsyslog in the web container no longer necessary.

djyasin:
@relrod thank you so much! I believe I have made the requested changes. Please let me know if anything looks wonky!

relrod:
@djyasin This looks good to me except for needing to make rsyslog not launch anymore in the web container (i.e., delete it from `tools/ansible/roles/dockerfile/templates/supervisor.conf.j2`).

djyasin:
@relrod thank you! I think this should be done now. Please let me know if I've missed anything!

tools/ansible/roles/dockerfile/templates/Dockerfile.j2
```diff
@@ -194,11 +194,10 @@ ENV _CONTAINERS_USERNS_CONFIGURED=""
...
 
-# Create default awx rsyslog config
-ADD tools/ansible/roles/dockerfile/files/rsyslog.conf /var/lib/awx/rsyslog/rsyslog.conf
 ADD tools/ansible/roles/dockerfile/files/wait-for-migrations /usr/local/bin/wait-for-migrations
 ADD tools/ansible/roles/dockerfile/files/stop-supervisor /usr/local/bin/stop-supervisor
 
+
 ## File mappings
 {% if build_dev|bool %}
 ADD tools/docker-compose/launch_awx.sh /usr/bin/launch_awx.sh
```
TheRealHaoLiu, line 203: the new launch_awx_rsyslog.sh is not added to the image

TheRealHaoLiu (changes requested):

tools/ansible/roles/dockerfile/templates/Dockerfile.j2
```diff
@@ -216,6 +215,7 @@ ADD tools/ansible/roles/dockerfile/files/launch_awx_task.sh /usr/bin/launch_awx_
 ADD tools/ansible/roles/dockerfile/files/uwsgi.ini /etc/tower/uwsgi.ini
```
TheRealHaoLiu, line 216:
```suggestion
ADD tools/ansible/roles/dockerfile/files/launch_awx_rsyslog.sh /usr/bin/launch_awx_rsyslog.sh
ADD tools/ansible/roles/dockerfile/files/uwsgi.ini /etc/tower/uwsgi.ini
```

TheRealHaoLiu: need to add the new `launch_awx_rsyslog.sh` to the `/usr/bin`

TheRealHaoLiu (approved):

PR #13332, opened 2022-12-14
Update clustering.md to be more current

fosterseth:
##### SUMMARY

these docs were wildly out of date, and also written as a proposal / ADR style

- Update to reflect k8s deployment type (as opposed to older docker style deployment pre-18.0)
- Describe communications between nodes in a clustered environment

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - Docs

### DIFF
--- a/docs/clustering.md
+++ b/docs/clustering.md
@@ -1,92 +1,169 @@
-## AWX Clustering/HA Overview
+# AWX Clustering Overview
 
-Prior to 3.1, the Ansible Tower HA solution was not a true high-availability system. This system has been entirely rewritten in 3.1 with a focus towards a proper highly-available clustered system. This has been extended further in 3.2 to allow grouping of clustered instances into different pools/queues.
-
-* Each instance should be able to act as an entry point for UI and API Access.
-  This should enable AWX administrators to use load balancers in front of as many instances as they wish and maintain good data visibility.
-* Each instance should be able to join the AWX cluster and expand its ability to execute jobs.
-* Provisioning new instance should be as simple as updating the `inventory` file and re-running the setup playbook.
-* Instances can be de-provisioned with a simple management command.
-* Instances can be grouped into one or more Instance Groups to share resources for topical purposes.
-* These instance groups should be assignable to certain resources:
-  * Organizations
-  * Inventories
-  * Job Templates
-
-  ...such that execution of jobs under those resources will favor particular queues.
-
-It's important to point out a few existing things:
-
-* PostgreSQL is still a standalone instance and is not clustered. Replica configuration will not be managed. If the user configures standby replicas, database failover will also not be managed.
-* All instances should be reachable from all other instances and they should be able to reach the database. It's also important for the hosts to have a stable address and/or hostname (depending on how you configure the AWX host).
-* Existing old-style HA deployments will be transitioned automatically to the new HA system during the upgrade process to 3.1.
-* Manual projects will need to be synced to all instances by the customer.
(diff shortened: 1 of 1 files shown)


docs/clustering.md
```diff
@@ -1,92 +1,168 @@
...
+               └───────────┘
+```
+
+There are two main deployment types, virtual machines (VM) or K8S. Ansible Automation Controller installation supports both VM or K8S deployments. The upstream AWX project only supports K8S deployments. Either deployment type supports cluster scaling.
+- Control plane nodes run a number of background services that are managed by supervisord
+  - dispatcher
+  - wsbroadcast
+  - callback receiver
+  - receptor (*managed under systemd)
+  - redis (*managed under systemd)
```
AlanCoding, line 42: Thank you for writing these out!

docs/clustering.md
```diff
@@ -1,92 +1,168 @@
...
+  - uwsgi
+  - daphne
+  - rsyslog
+- For K8S deployments, these background processes are containerized
+  - `awx-ee`: receptor
+  - `awx-web`: uwsgi, daphne, wsbroadcast, rsyslog
+  - `awx-task`: dispatcher, callback receiver
+  - `redis`: redis
+- Each control node is monolithic and contains all the necessary components for handling API requests and running jobs.
```
AlanCoding, line 51: for now... will change soon. But maybe this is a good strategy, as your updated wording makes it very clear where the new web/task split feature needs to update.

AlanCoding:
:+1: this content looks great, I just want to ask @shanemcd to help us get the proper doc scope regarding AAP, perhaps we should link to AAP installer docs. So explaining its operation isn't really a goal here, even though mentions are already fairly minimal.

docs/clustering.md
```diff
@@ -1,92 +1,168 @@
...
-  * Job Templates
+```
+       ┌───────────────────────────┐
+       │      Load-balancer        │
+       │   (configured separately) │
+       └───┬───────────────────┬───┘
+           │   round robin API │
+           ▼       requests    ▼
+
+  AWX VM/Pod 1               AWX VM/Pod 2
```
shanemcd: @fosterseth This still says "AWX VM".

docs/clustering.md
```diff
@@ -1,92 +1,168 @@
...
+## Scaling the cluster
+
+For VM (AAP) deployments, scaling up involves modifying `inventory`and re-running setup.sh
+For K8s deployments, scaling up is handled by changing the number of replicas in the AWX replica set.
+
+After scaling up, the new control plane node is registered in the database as a new `Instance`.
 
-  ...such that execution of jobs under those resources will favor particular queues.
+Instance types:
+`hybrid` - control plane node that can also run jobs
```
shanemcd: Please mention that hybrid and hop nodes are not available in default installations of AWX.

docs/clustering.md
fosterseth: I updated to be more generic `AWX Control Node`

docs/clustering.md
fosterseth: I made it clear these are part of AAP only

AlanCoding (approved):
way way better than what we had before. I think we should merge and then iterate on it from there as feature work continues.

gundalow:
> way way better than what we had before. I think we should merge and then iterate on it from there as feature work continues.

Agreed, this is a solid improvement on what we currently have.

Issue #13319, opened 2022-12-11
Tracking: Rework websockets to not rely on a shared Redis between task and web instances


relrod:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

Enhancement to Existing Feature

### Feature Summary

This is a tracking issue meant to group together all other issues and PRs related to this work.

Our goal here is to rework our websockets infrastructure to no longer depends on a shared Redis instance between task and web instances, so that task and web can be scaled independently of each other.

We envision that this will be done by fanning messages out from task nodes (where they are generated) to all web nodes who can then forward them on to clients connected to them.

Work is ongoing in the `wsrelay` branch toward this goal.

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

N/A

### Current results

Currently a shared Redis instance is required which makes it impossible to scale web and task instances separately.

### Sugested feature result

At this end of this work, we will have decoupled the websockets subsystem of AWX so that a shared Redis instance is no longer required. This will allow users to scale web and task separately.

Messages will emit from task systems like they do currently, and will be fanned out to all web servers who have clients connected to them which have expressed interested in these particular messages. We avoid sending messages to web servers who don't have clients that care about certain messages.

PR #13333, opened 2022-12-14
[wsrelay] Send subsystem metrics via wsrelay

fosterseth:
##### SUMMARY

Works by adding a dedicated producer in `wsrelay` that looks for local django channels message with group "metrics". The producer sends this to the consumer running in the web container.

The consumer running in the web container handles the message by pushing it into the local redis instance.

The django view that handles a request at the /api/v2/metrics endpoint will load this data from redis, format it, and return the response.

Refs #13319 
Refs #13334

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/analytics/subsystem_metrics.py
+++ b/awx/main/analytics/subsystem_metrics.py
@@ -9,7 +9,7 @@ from django.apps import apps
 from awx.main.consumers import emit_channel_notification
 from awx.main.utils import is_testing
 
-root_key = 'awx_metrics'
+root_key = settings.SUBSYSTEM_METRICS_REDIS_KEY_PREFIX
 logger = logging.getLogger('awx.main.analytics')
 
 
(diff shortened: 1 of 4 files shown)


awx/main/analytics/subsystem_metrics.py
```diff
@@ -309,14 +302,12 @@ def send_metrics(self):
                     'instance': self.instance_name,
                     'metrics': self.serialize_local_metrics(),
                 }
-                # store a local copy as well
-                self.store_metrics(json.dumps(payload))
 
                 # 🚨🚨🚨🚨🚨🚨🚨🚨
                 # TODO: rework how metrics are emitted and recorded. we used to exploit wsbroadcast's behavior of
```
shanemcd: I think we can remove this comment now 🙂

awx/main/wsrelay.py
```diff
@@ -107,6 +107,9 @@ def cancel(self):
         self.async_task.cancel()
 
     async def run_connection(self, websocket: aiohttp.ClientWebSocketResponse):
+        producer = self.event_loop.create_task(self.run_producer("metrics", websocket, "metrics"))
+        self.producers["metrics"] = {"task": producer, "subscriptions": {"metrics"}}
+        logger.info("Producer metrics started.")
```
shanemcd: Not sure how useful this log message is. This one here should do be enough, I think? 

https://github.com/ansible/awx/blob/43db246493ae4e1e4076341d44c6c3d989636d89/awx/main/wsrelay.py#L156

shanemcd (commented):
Left a couple comments, but overall this is very clean. Nice work 👍

relrod (approved):

PR #13338, opened 2022-12-14
tag awx-ee latest on awx release

fosterseth:
##### SUMMARY

pairs with https://github.com/ansible/awx-operator/pull/1158

each time AWX is released, tag quay.io/ansible/awx-ee:latest with the same AWX tag

this allows users to have an image to fallback to in case the lastest breaks their custom EEs

Tagging it with AWX release tags provides a regular cadence, and the tag itself allows users to roughly align awx-ee to awx images.

see https://github.com/ansible/awx-ee/issues/117
and https://github.com/ansible/awx-ee/issues/125

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

### DIFF
--- a/.github/workflows/stage.yml
+++ b/.github/workflows/stage.yml
@@ -84,6 +84,20 @@ jobs:
             -e push=yes \
             -e awx_official=yes
 
+      - name: Log in to GHCR
+        run: |
+          echo ${{ secrets.GITHUB_TOKEN }} | docker login ghcr.io -u ${{ github.actor }} --password-stdin
+
+      - name: Log in to Quay
+        run: |
+          echo ${{ secrets.QUAY_TOKEN }} | docker login quay.io -u ${{ secrets.QUAY_USER }} --password-stdin
+
+      - name: tag awx-ee:latest with version input
+        run: |
+          docker pull quay.io/ansible/awx-ee:latest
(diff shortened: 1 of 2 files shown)


.github/workflows/stage.yml
```diff
@@ -84,6 +84,16 @@ jobs:
...
 
+      - name: Log in to Quay
+        run: |
+          echo ${{ secrets.QUAY_TOKEN }} | docker login quay.io -u ${{ secrets.QUAY_USER }} --password-stdin
+
+      - name: tag awx-ee:latest with release tag
+        run: |
+          docker pull quay.io/ansible/awx-ee:latest
+          docker tag quay.io/ansible/awx-ee:latest quay.io/ansible/awx-ee:${{ github.event.release.tag_name }}
+          docker push quay.io/ansible/awx-ee:${{ github.event.release.tag_name }}
```
shanemcd: We shouldn't be pushing straight to quay in the stage workflow. There's already a pattern in place where we use ghcr.io as a staging area, run tests, then if everything succeeds we pull the staged images and _then_ push to quay.

shanemcd (changes requested):
See my comment about staging on ghcr.io.

relrod (approved):
This is awesome!

TheRealHaoLiu (approved):

rebeccahhh (approved):
thanks for doing this :)

shanemcd (approved):
Thanks for making all the requested changes! 🎉

PR #13340, opened 2022-12-15
Change workflow create/edit to null scm_branch when not provided.

gamuniz:
##### SUMMARY

The current behavior of workflow job templates is to pass in an empty string as scm_branch on all saves and edits. This becomes problematic when using job templates/workflows which allow prompt on launch for scm_branch as it may override the scm_branch set for the individual workflow nodes to an empty string. That behavior limits the usefulness of prompting scm branch as it can no longer be selected while creating workflows as they'll be overwritten.

##### ISSUE TYPE

##### COMPONENT NAME

 
 - UI
 

##### AWX VERSION

```
awx: 21.10.1.dev35+ga4b950f79b

```

##### ADDITIONAL INFORMATION

```

```


akus062381 (approved):

vidyanambiar (approved):
lgtm!

PR #13342, opened 2022-12-15
Fixing bug in LDAP reconcile loop

john-westcott-iv:
##### SUMMARY

The LDAP reconcile loop did not account for the same team name in multiple organizations. It would add or remove the user from any team by a name (which is not unique). This change now forces the reconcile loop to consider the team along with the organization. 

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.9.1.dev95+g14fa2685d5
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/sso/backends.py
+++ b/awx/sso/backends.py
@@ -477,84 +477,65 @@ def on_populate_user(sender, **kwargs):
 
 def reconcile_users_org_team_mappings(user, desired_org_states, desired_team_states, source):
     #
-    # desired_org_states: { '<org_name>': { '<role>': <boolean> } }
+    # Arguments:
+    #   user - a user object
+    #   desired_org_states: { '<org_name>': { '<role>': <boolean> or None } }
+    #   desired_team_states: { '<org_name>': { '<team name>': { '<role>': <boolean> or None } } }
+    #   source - a text label indicating the "authentication adapter" for debug messages
     #
-    # desired_team_states: { '<org_name>': { '<team name>': { '<role>': <boolean> } } }
+    # This function will load the users existing roles and then based on the deisred states modify the users roles
+    #    True indicates the user needs to be a member of the role
+    #    False indicates the user should not be a member of the role
+    #    None means this function should not change the users membership of a role
     #
     from awx.main.models import Organization, Team
 
     content_types = []
+    reconcile_items = []
     if desired_org_states:
         content_types.append(ContentType.objects.get_for_model(Organization))
+        reconcile_items.append(('organization', desired_org_states))
     if desired_team_states:
         content_types.append(ContentType.objects.get_for_model(Team))
+        reconcile_items.append(('team', desired_team_states))
 
     if not content_types:
         # If both desired states were empty we can simply return because there is nothing to reconcile
         return
 
-    # users_existing_roles is a flat set of IDs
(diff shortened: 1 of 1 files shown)


awx/sso/backends.py
```diff
@@ -473,16 +476,21 @@ def on_populate_user(sender, **kwargs):
 
 
 def reconcile_users_org_team_mappings(user, desired_org_states, desired_team_states, source):
+    #
+    # desired_org_states: { '<org_name>': { '<role>': <boolean> } }
+    #
+    # desired_team_states: { '<org_name>': { '<team name>': { '<role>': <boolean> } } }
```
AlanCoding: Really really helpful to have here, and just to capture as a TODO item, I suggest changing `<boolean>` to `<boolean> or None` for type clarity. Although it looks like the caller could pop off all None values to the same effect.

awx/sso/backends.py
```diff
@@ -491,24 +499,39 @@ def reconcile_users_org_team_mappings(user, desired_org_states, desired_team_sta
...
+            team_names = []
+            for teams_dict in desired_states.values():
+                team_names.extend(teams_dict.keys())
+                for sub_dict in teams_dict.values():
+                    for role_name in sub_dict:
+                        if sub_dict[role_name] is None:
+                            continue
+                        if role_name not in roles:
+                            roles.append(role_name)
+            model_roles = Team.objects.filter(name__in=team_names).values_list('name', 'organization__name', *roles, named=True)
```
AlanCoding, line 530: Now this query performs the same _filter_ as before, but it returns more values so that you capture the organization name... (continuing walk-through)

awx/sso/backends.py
```diff
@@ -491,24 +499,39 @@ def reconcile_users_org_team_mappings(user, desired_org_states, desired_team_sta
...
+
         for row in model_roles:
             for role_name in roles:
-                desired_state = desired_states.get(row.name, {})
-                if desired_state[role_name] is None:
+                if object_type == 'organization':
+                    desired_state = desired_states.get(row.name, {})
+                else:
+                    desired_state = desired_states.get(row.organization__name, {}).get(row.name, {})
```
AlanCoding, line 537: Here, we have the possibility that the team in question is unwanted, because it has the same name as a team listed in the configuration, but it's in another organization. In that event, `desired_states` should be lacking either the team's org name key or the team name key, and this should return `{}`

awx/sso/backends.py
```diff
@@ -491,24 +499,39 @@ def reconcile_users_org_team_mappings(user, desired_org_states, desired_team_sta
...
+
+                if desired_state.get(role_name, None) is None:
                     # The mapping was not defined for this [org/team]/role so we can just pass
-                    pass
+                    continue
```
AlanCoding, line 541: And here, where we have an unwanted "extra" team returned by the query, we hit the `continue` because it was not in the configuration (the `desired_states`). We don't remove or add the user, and that fixes the bug.

AlanCoding (approved):
I know tests are still failing, and I don't know why, but I can follow everything here and I believe it fixes the bug.

We should add test coverage, even unit tests feel workable. But for the core code changes, I can read this and see through to the fix.

john-westcott-iv:
We should def add test coverage for this function.

jay-steurer (approved):

PR #13351, opened 2022-12-19
add logging to situation in which project lock file is locked

jangel97:
Currently, if lock file of project is locked, there is no trace in logs https://github.com/ansible/awx/blob/devel/awx/main/tasks/jobs.py#L404

How to replicate the issue:
Run on a controller instance following python3 commands:

```
import os
import fcntl
lock_fd = os.open('/var/lib/awx/projects/_8__example_project.lock', os.O_RDWR | os.O_CREAT)
fcntl.lockf(lock_fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
```
Then, when running project update, the job will get stuck.

With this change I can see following log entry:
`/var/log/tower/task_system.log:2022-12-19 16:17:15,586 INFO     [dcfe5b893c2b48429d8131ff1ce4dce5] awx.main.tasks.jobs exception acquiring lock /var/lib/awx/projects/_8__example_project.lock: [Errno 11] Resource temporarily unavailable, sleeping`

When running `quit()` command project update should resume.

New or Enhanced Feature

### DIFF
--- a/awx/main/tasks/jobs.py
+++ b/awx/main/tasks/jobs.py
@@ -390,6 +390,7 @@ class BaseTask(object):
             logger.error("I/O error({0}) while trying to open lock file [{1}]: {2}".format(e.errno, lock_path, e.strerror))
             raise
 
+        emitted_lockfile_log = False
         start_time = time.time()
         while True:
             try:
@@ -401,6 +402,9 @@ class BaseTask(object):
                     logger.error("I/O error({0}) while trying to aquire lock on file [{1}]: {2}".format(e.errno, lock_path, e.strerror))
                     raise
                 else:
+                    if not emitted_lockfile_log:
+                        logger.info(f"exception acquiring lock {lock_path}: {e}")
+                        emitted_lockfile_log = True
                     time.sleep(1.0)
             self.instance.refresh_from_db(fields=['cancel_flag'])
             if self.instance.cancel_flag or signal_callback():


AlanCoding:
This could get noisy during the course of normal option, but I agree this should be fixed. It is quite common that a lot of time is spent here when a series of concurrent jobs are launched using the same project, and it polls every second. For that case, the exception looks like this:

```
tools_awx_1     | Traceback (most recent call last):
tools_awx_1     |   File "/awx_devel/awx/main/tasks/jobs.py", line 396, in acquire_lock
tools_awx_1     |     fcntl.lockf(self.lock_fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
tools_awx_1     | BlockingIOError: [Errno 11] Resource temporarily unavailable
```

If that's not distinguishable from the case you're interested in, then I don't really know what to do about that. Blocking is expected to happen as multiple jobs from the same project are trying to get a copy of the project, and they are waiting on the source tree to be updated.

jangel97:
Hi,

Do you think increasing the sleep would help maybe?

Kind regards,

fosterseth:
@jangel97 we still want a low sleep value like 1 second to keep things flowing as soon as it can

maybe we can print the log message every 20 seconds or so?

```python
curr_time = time.time()
if curr_time - time_since_log > 20:
   logger.info(f"exception acquiring lock {lock_path}: {e}, sleeping")
   time_since_log = curr_time
```

fosterseth:
@jangel97 after some more thought, I think it would be better to emit the log message just once

```python3
if not emitted_log and time.time() - start_time > 1.0:
   logger.info( ...)
   emitted_log = True
```

jangel97:
Sorry for that, I closed it accidentally. I added some code so the log line only lands once

AlanCoding:
The current form seems acceptable to me @fosterseth

awx/main/tasks/jobs.py
```diff
@@ -401,6 +402,9 @@ def acquire_lock(self, project, unified_job_id=None):
                     logger.error("I/O error({0}) while trying to aquire lock on file [{1}]: {2}".format(e.errno, lock_path, e.strerror))
                     raise
                 else:
+                    if not(emitted_lockfile_log):
```
fosterseth: super minor but can we change this to `not emitted_lockfile_log` otherwise it looks like a function call

jangel97: no problem! changed!

fosterseth (approved):

AlanCoding (approved):

fosterseth:
@jangel97 for the lint, can you run `make black` in the awx repo? you may need to pip install black first

fosterseth:
@jangel97 linter being a pain, sorry

only jobs.py should be changed. Can you rebase on devel, and rerun make black? if you still get the other file changes, maybe just ignore them an only commit the jobs.py diff

fosterseth:
@jangel97 nice contribution, thanks!

PR #13352, opened 2022-12-19
Remove `subtasks` keyword arg that can exceed pg_notify max message length

AlanCoding:
##### SUMMARY
Connect https://github.com/ansible/awx/issues/12999

This is intended to be a very conservative fix for that issue. Just don't pass a list of tasks, you can get it from the database.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
This isn't the primary objective, but I just want to note @kdelee, that this does remove a query from `start_task`, and as I look, I don't see `unifiedjob_blocked_jobs` prefetched. I believe this should follow our general direction of "move logic out of the task manager where possible".

### DIFF
--- a/awx/main/tasks/system.py
+++ b/awx/main/tasks/system.py
@@ -52,6 +52,7 @@ from awx.main.constants import ACTIVE_STATES
 from awx.main.dispatch.publish import task
 from awx.main.dispatch import get_local_queuename, reaper
 from awx.main.utils.common import (
+    get_type_for_model,
     ignore_inventory_computed_fields,
     ignore_inventory_group_removal,
     ScheduleWorkflowManager,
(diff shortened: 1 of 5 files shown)


awx/main/tasks/system.py
```diff
@@ -720,45 +721,43 @@ def handle_work_success(task_actual):
...
-                instance.websocket_emit_status("failed")
+    subtasks = instance.get_jobs_fail_chain()  # reverse of dependent_jobs mostly
+    logger.debug(f'Executing error task id {task_actual["id"]}, subtasks: {[subtask.id for subtask in subtasks]}')
+
+    deps_of_deps = {}
+
+    for subtask in subtasks:
+        if subtask.celery_task_id != instance.celery_task_id and not subtask.cancel_flag and not subtask.status in ('successful', 'failed'):
+            # If there are multiple in the dependency chain, A->B->C, and this was called for A, blame B for clarity
+            blame_job = deps_of_deps.get(subtask.id, instance)
```
AlanCoding, line 741: @fosterseth this fixes the test case that we have today, for the case where there is a project update which is needed by an inventory update, which is needed by a job.

But the whole thing doesn't actually make any sense when you look hard enough. If you have that situation, the inventory update and the job both exist and are in a holding pattern, blocked by the `TaskManager`. When the task manager tries to advance these pending jobs, it runs through their `task.dependent_jobs.all()` which is... basically... the reverse of `get_job_fail_chain()` which is used here. The task manager then attaches that message

> Previous Task Failed:

And it does so in a way indistinguishable from here.

So, which code has responsibility for marking dependent jobs as failed (the job and inventory update in this case)? It seems the answer is "both"... er... whichever gets to it first. This is not a good idea.

The test case we had seems to have been working by accident, because the `dependent_jobs` fail chain changes over time, and the `subtasks` passed to the dispatcher for starting the original update was empty. After the task manager submitted the update, the inventory update was launched, and linked the update as a dependency. Because of that, this task did nothing, which deferred to the task manager, which applied the expected message. Had the specific means of launching been different (like putting the inventory update in a workflow), it would have been unlikely to give the expected error message.

Those are exactly the bigger-picture separation-of-responsibilities problems that https://github.com/ansible/awx/pull/13061 was concerned with. I just wanted to take this change to articulate one thing.

AlanCoding:
Got results from integration tests, and got all-clear on this branch, so this is ready for review.

awx/main/tasks/system.py
```diff
@@ -720,45 +721,43 @@ def handle_work_success(task_actual):
...
+                subtask.job_explanation = 'Previous Task Failed: {"job_type": "%s", "job_name": "%s", "job_id": "%s"}' % (
+                    get_type_for_model(type(blame_job)),
+                    blame_job.name,
+                    blame_job.id,
+                )
+            subtask.save()
+            subtask.websocket_emit_status("failed")
+
+            for sub_subtask in subtask.get_jobs_fail_chain():
+                deps_of_deps[sub_subtask.id] = subtask
```
fosterseth, line 754: smart!

fosterseth (approved):

PR #13355, opened 2022-12-20
Update saml.md

tallyoh:
According to latest documentation, role and value are now "one or more" fields. So they both need to be arrays.  Entering the json data as you have in this article doesn't work. But when I added the brackets, it then worked.   Thank you

##### ISSUE TYPE

##### COMPONENT NAME

 - Docs

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/docs/auth/saml.md
+++ b/docs/auth/saml.md
@@ -107,12 +107,12 @@ Below is an example of a SAML attribute that contains admin attributes:
 These properties can be defined either by a role or an attribute with the following configuration options:
 ```
 {
-  "is_superuser_role": "awx_admins",
+  "is_superuser_role": ["awx_admins"],
   "is_superuser_attr": "is_superuser",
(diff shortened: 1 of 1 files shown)


john-westcott-iv:
@tallyoh thanks for finding and fixing this. I would like to get this merged but we need to pass the required checks first. These were not caused by your change but are due to an issue we have with rsyslogd. We have address these issues but we need to rebase your patch onto devel in order to get the fix in your branch. Please reach out to me here or in our Matrix channel if you need help with that. Once you are rebased the tests should patch and we should be able to merge this. If you no longer have the time to work on this please let me know and I will take your commit and get it rebased and into a new PR.

john-westcott-iv:
Replacing with https://github.com/ansible/awx/pull/13490.

PR #13356, opened 2022-12-20
Run collection sanity tests in CI

AlanCoding:
##### SUMMARY
Attempting to run collection sanity tests in github actions.

May require adding more dependencies.

##### ISSUE TYPE

##### COMPONENT NAME
 - API


shanemcd (approved):

Makefile
```diff
@@ -330,7 +333,12 @@ install_collection: build_collection
...
 	ansible-galaxy collection install awx_collection_build/$(COLLECTION_NAMESPACE)-$(COLLECTION_PACKAGE)-$(COLLECTION_VERSION).tar.gz
 
-test_collection_sanity: install_collection
+test_collection_sanity:
+	rm -rf awx_collection_build/
+	rm -rf $(COLLECTION_INSTALL)
+	if ! [ -x "$(shell command -v ansible-test)" ]; then pip install ansible-core; fi
+	ansible --version
+	COLLECTION_VERSION=1.0.0 make install_collection
 	cd $(COLLECTION_INSTALL) && ansible-test sanity
```
felixfontein: I would recommend to always use `--docker` when running `ansible-test sanity`, to make sure that all sanity tests are actually run, and that the correct dependencies are used:
```suggestion
	cd $(COLLECTION_INSTALL) && ansible-test sanity --docker
```

felixfontein: (See also the collection requirements for Ansible: https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst#ci-testing.)

gundalow: For $reasons, we are inside of Docker already...

felixfontein: Then `--requirements` needs to be passed, and someone has to make sure sure that the Docker container used has all Python versions installed that both the collection and ansible-core supports.

AlanCoding: Now we are running `ansible-test sanity --exclude=plugins/modules/export.py` because this particular export module requires the `awxkit` dependency. None of us can figure out how to get the `--requirements` to work. Ultimately we need a `-e` install, since awxkit is maintained in the same repo as this.

.github/workflows/ci.yml
```diff
@@ -28,6 +28,9 @@ jobs:
           - name: awx-collection
             command: /start_tests.sh test_collection_all
             label: Run Collection Tests
+          - name: awx-collection-sanity
+            command: /start_tests.sh test_collection_sanity
+            label: Run Ansible core Collection Sanity tests
```
relrod, line 33: Not directly related to your PR, but I just noticed this label never gets used.

On (new) line 69, we have a typo: `matrix.texts.label` instead of `matrix.tests.label`.

AlanCoding: I have some follow-on changes to attempt, I might combine this comment with those.

Makefile
felixfontein: Just passing `--requirements` to `ansible-test sanity` makes sure that ansible-test installs the Python requirements it needs to run the sanity tests.

About `--exclude=plugins/modules/export.py`: why not use ignore.txt entries for that one? Also what exactly is preventing the sanity tests to work for that one?

AlanCoding: Yes, only with @gundalow's advice did I get to the point of creating a file `awx_collection/tests/sanity/requirements.txt`, and passing `--requirements` in conjunction. That also didn't work. I have a long history of trying and failing to solve this problem, and at the moment I want to avoid it blocking progress other problems.

We really don't want to ignore that file, you can see I make changes in this PR that the sanity tests discovered, but I only got there by hard-coding the value that it's trying to import from awxkit. It just needs the dependency installed, which is available locally and on PyPI, if it helps for testing.

felixfontein: `awx_collection/tests/sanity/requirements.txt` has nothing to do with `--requirements`; the `--requirements` parameter tells ansible-test that it installs **its own requirements** (specified internally by ansible-test). `ansible-test sanity` never installs any requirements specified by collections (and that's fully intentional).

`--exclude=plugins/modules/export.py` should **really** be removed, totally ignoring a file is a red flag. The tests that cannot run (due to such issues as argspec needing some external library to be properly composed) should be handled with ignore.txt entries to disable exactly these problems and not anything else. By completely ignoring that module, you are also disabling all other sanity tests for it.

AlanCoding: We are already running sanity tests in the downstream checks, and have a QE team. Adding this to the ignore file would mess up expected coverage _there_, I want to be very clear that this is a best-effort hacked-together check so that we can address these issues from the community quickly and efficiently.

The `awx.awx.export` module requires the ability to import `awxkit` in order to build the python argument spec. The sanity failure is that the python argument spec conflicts with `DOCUMENTATION`. It is trivial to install `awxkit` to the global python site packages, but the problem is that `ansible-test` creates a new venv that does not use system site packages, and as you say, it doesn't install test requirements. In order for the arg spec to be correct, that awxkit python library has to be installed in the python environment. For long-term maintainability, we need a `-e` install from the `awxkit/` folder in this repo.

PR #13331, opened 2022-12-13
[proof of concept -- 2nd round] Bulk Job Launch

kdelee:
##### SUMMARY

replaces https://github.com/ansible/awx/pull/13271

The idea behind this is some way to launch a number of jobs with 1 request, the client could then subscribe to websocket channel and get updates on status of the nodes or at least just poll this one workflow until individual job nodes are done and then reach into the jobs.

This is all to provide some interface that allows clients to get information they need without barraging controller with many many requests. With a limit of 100 nodes per Bulk Job, launching 1,000 jobs would take 10 web requests -- which is totally achievable with current uwsgi setup.

First pass does not attempt to implement any workflow structure other than flat, maybe not even necessary if only goal is a bulk job launch endpoint. Even so, we support passing "identifier"s on the workflow nodes to be able to tell the jobs apart in websocket messages or the api.

A significant concern is performing all necessary RBAC checks for all nodes and promptable items like Inventory, Credentials, etc. I have implemented this a good bit of the way, but it needs review/testing for both correctness and performance depending on how many unified job templates, credentials, etc are in system and how many nodes are in workflow/use of prompt-able items for nodes.

POST /api/v2/bulk_jobs/launch
```
{
"jobs": [
{"unified_job_template": 7, "inventory": 1, "credentials": [1]},
{"unified_job_template": 7, "inventory": 1, "credentials": [1]},
{"unified_job_template": 7, "inventory": 1, "credentials": [1]},
{"unified_job_template": 7, "inventory": 1, "credentials": [1]}
]
}
```
results in:
```
HTTP 201 Created
Allow: GET, POST, OPTIONS
Content-Type: application/json
Vary: Accept
X-API-Node: awx_1
X-API-Product-Name: AWX
X-API-Product-Version: 21.6.1.dev146+g1619462fb0
X-API-Query-Count: 29
X-API-Query-Time: 0.008s
X-API-Time: 0.034s

{
    "id": 947,
    "type": "workflow_job",
    "url": "[/api/v2/workflow_jobs/947/](https://localhost:8043/api/v2/workflow_jobs/947/)",
    "related": {
        "created_by": "[/api/v2/users/3/](https://localhost:8043/api/v2/users/3/)",
        "modified_by": "[/api/v2/users/3/](https://localhost:8043/api/v2/users/3/)",
        "workflow_nodes": "[/api/v2/workflow_jobs/947/workflow_nodes/](https://localhost:8043/api/v2/workflow_jobs/947/workflow_nodes/)",
[...]

PR #13271, opened 2022-12-02
[draft] proof of concept bulk job launch

kdelee:
The idea behind this is some way to launch a number of jobs with 1 request, the client could then subscribe to websocket channel and get updates on status of the nodes or at least just poll this one workflow until individual job nodes are done and then reach into the jobs.

 This is all to provide some interface that allows clients to get information they need without barraging controller with many many requests.  **_With a limit of 100 nodes per  Bulk Job, launching 1,000 jobs would take 10 web requests -- which is totally achievable with current uwsgi setup._**

First pass does not attempt to implement any workflow structure other than flat, maybe not even necessary if only goal is a bulk job launch endpoint. Even so, we support passing "identifier"s on the workflow nodes to be able to tell the jobs apart in websocket messages or the api.

A significant concern is performing all necessary RBAC checks for all nodes and promptable items like Inventory, Credentials, etc. I have implemented this a good bit of the way, but it needs review/testing for both correctness and performance depending on how many unified job templates, credentials, etc are in system and how many nodes are in workflow/use of promptable items for nodes.

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### ADDITIONAL INFORMATION

POST to /api/v2/bulk_jobs/launch
```
{
    "name": "foobar example bulk launch",
    "workflow_job_nodes": [
        {"workflow_job_template_id": 227},
        {"job_template_id": 7},
        {"job_template_id": 7, "inventory": 1, "credentials": [18]},
        {"project_id": 6},
        {"inventory_source_id": 25}
    ]
}
```

response:
```
HTTP 201 Created
Allow: GET, POST, HEAD, OPTIONS
Content-Type: application/json
Location: /api/v2/workflow_jobs/5430/
Vary: Accept
X-API-Node: awx_1
X-API-Product-Name: AWX
X-API-Product-Version: 21.6.1.dev146+g1619462fb0
X-API-Query-Count: 54
X-API-Query-Time: 0.025s
X-API-Time: 0.207s

{
    "workflow_job": 5430,
    "id": 5430,
    "type": "workflow_job",
    "url": "[/api/v2/workflow_jobs/5430/](https://localhost:8043/api/v2/workflow_jobs/5430/)",
    "related": {
        "created_by": "[/api/v2/users/2/](https://localhost:8043/api/v2/users/2/)",
        "modified_by": "[/api/v2/users/2/](https://localhost:8043/api/v2/users/2/)",
[...]

PR #13358, opened 2022-12-20
[wip feature] Bulk job launch

kdelee:
##### SUMMARY

replaces https://github.com/ansible/awx/pull/13331, https://github.com/ansible/awx/pull/13271

Going to use this as the base feature branch and @jainnikhil30 and I will be making PRs to this branch until we are ready for final review into devel

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### ADDITIONAL INFORMATION

The idea behind this is some way to launch a number of jobs with 1 request, the client could then subscribe to websocket channel and get updates on status of the nodes or at least just poll this one workflow until individual job nodes are done and then reach into the jobs.

This is all to provide some interface that allows clients to get information they need without barraging controller with many many requests. With a limit of 100 nodes per Bulk Job, launching 1,000 jobs would take 10 web requests -- which is totally achievable with current uwsgi setup.

First pass does not attempt to implement any workflow structure other than flat, maybe not even necessary if only goal is a bulk job launch endpoint. Even so, we support passing "identifier"s on the workflow nodes to be able to tell the jobs apart in websocket messages or the api.

A significant concern is performing all necessary RBAC checks for all nodes and promptable items like Inventory, Credentials, etc. I have implemented this a good bit of the way, but it needs review/testing for both correctness and performance depending on how many unified job templates, credentials, etc are in system and how many nodes are in workflow/use of prompt-able items for nodes.

POST /api/v2/bulk_jobs/launch
```
{
"jobs": [
   {"unified_job_template": 9, "identifier":"foo", "limit": "kansas", "credentials": [3], "inventory": 2 },
   {"unified_job_template": 10, "identifier": "bar"},
   {"unified_job_template": 10}
]
}
```
results in:
```
HTTP 201 Created
Allow: GET, POST, OPTIONS
Vary: Accept
X-API-Node: awx_1
X-API-Product-Name: AWX
X-API-Product-Version: 21.6.1.dev146+g1619462fb0
X-API-Query-Count: 28
X-API-Query-Time: 0.284s
X-API-Time: 0.832s

{
    "id": 11014,
    "type": "workflow_job",
    "url": "[/api/v2/workflow_jobs/11014/](https://localhost:8043/api/v2/workflow_jobs/11014/)",
    "related": {
        "created_by": "[/api/v2/users/3/](https://localhost:8043/api/v2/users/3/)",
        "modified_by": "[/api/v2/users/3/](https://localhost:8043/api/v2/users/3/)",
        "workflow_nodes": "[/api/v2/workflow_jobs/11014/workflow_nodes/](https://localhost:8043/api/v2/workflow_jobs/11014/workflow_nodes/)",
        "labels": "[/api/v2/workflow_jobs/11014/labels/](https://localhost:8043/api/v2/workflow_jobs/11014/labels/)",
        "activity_stream": "[/api/v2/workflow_jobs/11014/activity_stream/](https://localhost:8043/api/v2/workflow_jobs/11014/activity_stream/)",
        "relaunch": "[/api/v2/workflow_jobs/11014/relaunch/](https://localhost:8043/api/v2/workflow_jobs/11014/relaunch/)",
        "cancel": "[/api/v2/workflow_jobs/11014/cancel/](https://localhost:8043/api/v2/workflow_jobs/11014/cancel/)"
    },
    "summary_fields": {
        "created_by": {
            "id": 3,
            "username": "orgadmin",
            "first_name": "",
            "last_name": ""
        },
        "modified_by": {
            "id": 3,
            "username": "orgadmin",
            "first_name": "",
            "last_name": ""
        },
        "labels": {
            "count": 0,
            "results": []
        }
    },
    "created": "2022-12-20T16:18:02.010652Z",
    "modified": "2022-12-20T16:18:02.289446Z",
    "name": "Bulk Job Launch",
    "description": "",
    "unified_job_template": null,
    "launch_type": "manual",
    "status": "pending",
    "failed": false,
    "started": null,
    "finished": null,
    "canceled_on": null,
    "elapsed": 0.0,
    "job_args": "",
    "job_cwd": "",
    "job_env": {},
    "job_explanation": "",
    "result_traceback": "",
    "launched_by": {
        "id": 3,
        "name": "orgadmin",
        "type": "user",
        "url": "[/api/v2/users/3/](https://localhost:8043/api/v2/users/3/)"
    },
    "work_unit_id": null,
    "workflow_job_template": null,
    "extra_vars": "",
    "allow_simultaneous": false,
    "job_template": null,
    "is_sliced_job": false,
    "inventory": null,
    "limit": null,
    "scm_branch": null,
    "webhook_service": "",
    "webhook_credential": null,
    "webhook_guid": "",
    "skip_tags": null,
    "job_tags": null
}
```

I've tested with up to 1,000 jobs and at this point I'm getting a constant number of queries -- e.g., to launch 10 job it takes roughtly the same number of queries as it takes to launch 1000 jobs. There is a difference on the order of 10-20 queries depending on how many different types of objects are needed and if the user is a system adminstrator or regular user. For example, if you include prompt-on-launch inventories, whether 1 or 100 different ones, we will have to do an additional couple of queries to check RBAC and get the objects, but this is per TYPE of object, not per number of objects total.

##### At WorkflowJob level:
- [ ] reach feature parity with things you can normally set on a WorkflowJobTemplate that get inherited by the WorkflowJob
    - [x] "name" (done)
    - [x] "description": "",
    - [x] "extra_vars": "",
    - [x] "organization": (see below in RBAC section)
    - [ ] "inventory": null,
    - [x] "limit": "",
    - [x] "scm_branch": ""
    - [ ] "webhook_service": null,
    - [ ] "webhook_credential": null,
    - [x] "skip_tags": "",
    - [x] "job_tags": ""

##### At workflow job node level:
- [x] implement RBAC and appropriate relating of other related objects 
  - [x] inventory
  - [x] credentials 
  - [x] instance groups
  - [x] execution environment
  - [x] labels
- [x] implement schema check and assignment of other character/integer fields 
  - [x] 'limit'
  - [x] 'char_prompts',
  - [x] 'diff_mode',
  - [x] 'extra_data',
  - [x] 'forks',
  - [x] 'job_slice_count',
  - [x] 'job_tags',
  - [x] 'job_type'
  - [x] 'scm_branch',
  - [x] 'skip_tags',
  - [x] 'survey_passwords',
  - [x] 'timeout',
  - [x] 'verbosity', 

#### RBAC
- [x] add "is_bulk_job" field to WorkflowJob model
- [x] associate the user organization to the WorkflowJob if they only have one org, or allow them to specify an organization they belong to as the org (unless they are sysadmin, in which case they can assocaite any org). If user has multiple organizations, return a validation error and tell them they need to pick one
- [x] Anyone who launched a bulk job can see the one they launched
- [x] An org admin for the Organization on a bulk job can see that bulk job
- [x] open up RBAC access for WorkflowJob to also include case when (pseudocode) `("is_bulk_job = True" & (("launched_by" == current user) | current_user.is_system_administrator | current_user.is_system_auditor))`
- [ ] Figure out how to prohibit relaunch. Probably in access.py we can do check `is_bulk_job=True` and return PermissionDenied with specific message
- [ ] record workflow job launch in activity stream. see https://github.com/ansible/awx/blob/ca07bc85cbd34a760589278eb42bb453a00039a5/awx/main/models/unified_jobs.py#L426-L428 for example

#### Maintainablity
- [ ] figure out if there is a way we can share some code between access.py and the validation method to reduce risk of drift between the two
- [ ] figure out if we can chunk up the validate method into some more sensible helper methods to make it a bit more readable/have logical sections. E.g. "_validate_permissions", "_objectify_attrs_for_create"

#### General API functionality
- [x] provide /api/v2/bulk/job_launch
- [x] list of individual nodes will be accessed via `workflow_job_nodes` related item on the workflow job returned
- [ ] return something sensible to a GET to /api/v2/bulk/job_launch. Use other launch endpoints as a reference. They appear to return a json blob with all the default values of the items you could launch
- [x] in the DjangoRestFramework API browser thing, in the "POST" form, it appears there is someway you can populate it with sample POST data. Provide something sensible for  /api/v2/bulk_jobs/launch
- [x] Provide some sensible OPTIONS response to  /api/v2/bulk/job_launch

#### Tests
- [ ] awxkit part to expose the api/v2/bulk
- [ ] awxkit part to expose api/v2/bulk/job_launch
- [ ] Add tests that do some validation that # of queries stays reasonable
- [ ] Test with LOTS of unified_job_templates in existence in the system and regular users. Concern is RBAC check queries could return a lot of pk's -- question is, where does this become a problem. I did nominal testing with ~ 10k job templates, didn't appear to be an issue but worth poking at some more.
- [ ] Add tests that do some validation of RBAC access
- [ ] Verify current behavior of launching jobs with "promptable" items that are either not allowed by the job template or you don't have access to. I think right now alot of these just get dropped on the floor e.g. silently ignored. Confirm this and compare to how it is working with /api/v2/bulk/job_launch


jainnikhil30:
@kdelee for the related fields, Please correct me if I am wrong, I think we dont need instance_groups and execution_environment. Those two are not valid fields for a workflow job right ? I dont see them in Workflow Job template and also in the serializer.

awx/api/serializers.py
```diff
@@ -4453,6 +4454,249 @@ def validate(self, attrs):
...
+            'survey_passwords',
+            'timeout',
+            # these are related objects and we need to add extra validation for them in the parent BulkJobLaunchSerializer
+            #
+        )
+
+
+class BulkJobLaunchSerializer(serializers.Serializer):
+    name = serializers.CharField(max_length=512, required=False)  # limited by max name of jobs
+    jobs = BulkJobNodeSerializer(many=True, allow_empty=False, max_length=1000)
```
kdelee: Here is where we would add the more WorkflowJob attributes like 
```
 "description": "",
 "extra_vars": "",
 "organization": (Maybe if they don't set it, set their first organization) # Here we can use PrimaryKeyRelatedField so it will automagically do rbac/turn into object
 "inventory": "", # Here we can use PrimaryKeyRelatedField so it will automagically do rbac/turn into object
 "limit": "",
 "scm_branch": ""
 "webhook_service": null,  # Here we can use PrimaryKeyRelatedField so it will automagically do rbac/turn into object, I think, I'm actually not sure how to use this
 "webhook_credential": null, # Here we can use PrimaryKeyRelatedField so it will automagically do rbac/turn into object  I think, I'm actually not sure how to use this
 "skip_tags": "",
 "job_tags": ""
```

awx/api/serializers.py
```diff
@@ -4453,6 +4454,249 @@ def validate(self, attrs):
...
+
+    class Meta:
+        fields = ('name', 'jobs')
+        read_only_fields = ()
+
+    def validate(self, attrs):
+
```
kdelee: if we add the PrimaryKeyRelated field stuff, we may want to do something like
```
jobs = attrs.pop('jobs')
attrs = self.super().validate(attrs)
```
I think that will do the magic validation/turn the PKs into objects

awx/api/serializers.py
```diff
@@ -4453,6 +4454,249 @@ def validate(self, attrs):
...
+            #
+        )
+
+
+class BulkJobLaunchSerializer(serializers.Serializer):
+    name = serializers.CharField(max_length=512, required=False)  # limited by max name of jobs
+    jobs = BulkJobNodeSerializer(many=True, allow_empty=False, max_length=1000)
+
+    class Meta:
+        fields = ('name', 'jobs')
```
kdelee: also have to add fields here when you add them above as attributes to the class

kdelee:
We've agreed that:

In Serializer, we will let anyone who has access to launch underlying jobs may launch a bulk job, and you can either provide an organization (you must be member or sysadmin) to associate with the WorkflowJob or we will take your organization if you only have one. Otherwise, return validation error if you have multiple and you did not specify.

In access.py we will modify rules so that:
  - Anyone who launched a bulk job can see the one they launched
  - An org admin for the Organization on a bulk job can see that bulk job

In Serializer/settings:
   - Make max number of jobs in a Bulk Job customization via a setting, but have a hard limit of 1000 jobs

Performance Test Cases:
Perform tests where we up the limit to much larger numbers like 20k or 50k to see how the WorkflowManager responds when spawning and how system performs with 50k+ pending jobs

mabashian:
One thing that we talked about that could make this more impactful in the UI would be to add a feature to the workflow output to filter the nodes down to only the nodes that have failed jobs.  We could peel that request out into it's own issue and could be implemented separately.

awx/api/serializers.py
```diff
@@ -4453,6 +4454,249 @@ def validate(self, attrs):
...
+
+        # all of the unified job templates and related items have now been checked, we can now grab the objects from the DB
+        # TODO: As we add more related objects like Label, InstanceGroup, etc we need to add them here
+        objectified_jobs = []
+        key_to_obj_map = {
+            "unified_job_template": {obj.id: obj for obj in UnifiedJobTemplate.objects.filter(id__in=requested_ujts)},
+            "inventory": {obj.id: obj for obj in Inventory.objects.filter(id__in=requested_use_inventories)},
+            "credentials": {obj.id: obj for obj in Credential.objects.filter(id__in=requested_use_credentials)},
+            "labels": {obj.id: obj for obj in Label.objects.filter(id__in=requested_use_labels)},
+            "instance_groups": {obj.id: obj for obj in InstanceGroup.objects.filter(id__in=requested_use_instance_groups)},
```
kdelee: need to add execution environments here

awx/api/serializers.py
```diff
@@ -4453,6 +4454,249 @@ def validate(self, attrs):
...
+        WorkflowJobNode.objects.bulk_create(nodes)
+
+        # Deal with the m2m objects we have to create once the node exists
+        for obj_type, obj_through_model in node_m2m_object_types_to_through_model.items():
+            through_models = []
+            for node_identifier in node_m2m_objects.keys():
+                if obj_type in node_m2m_objects[node_identifier] and obj_type == 'credentials':
+                    for cred in node_m2m_objects[node_identifier][obj_type]:
+                        through_models.append(obj_through_model(credential=cred, workflowjobnode=node_m2m_objects[node_identifier]['node']))
+                if obj_type in node_m2m_objects[node_identifier] and obj_type == 'labels':
```
kdelee, line 4686: need to add clause for execution environments

awx/api/serializers.py
```diff
@@ -4453,6 +4454,249 @@ def validate(self, attrs):
...
+        if 'name' not in validated_data:
+            validated_data['name'] = 'Bulk Job Launch'
+
+        wfj = WorkflowJob.objects.create(**validated_data, is_bulk_job=True)
+        nodes = []
+        node_m2m_objects = {}
+        node_m2m_object_types_to_through_model = {
+            'credentials': WorkflowJobNode.credentials.through,
+            'labels': WorkflowJobNode.labels.through,
+            'instance_groups': WorkflowJobNode.instance_groups.through,
```
kdelee, line 4633: need to add execution environments here

jainnikhil30:
TODO:

We also need to work on the awxkit part to expose the api/v2/bulk in there. Thanks @CFSNM for that pointer.

awx/api/serializers.py
```diff
@@ -4453,6 +4455,294 @@ def validate(self, attrs):
...
+        help_text=_('Inherit permissions from organization roles.'),
+    )
+    # inventory: "",  # Here we can use PrimaryKeyRelatedField so it will automagically do rbac/turn into object
+    limit: serializers.CharField(write_only=True, required=False, allow_blank=False)
+    scm_branch: serializers.CharField(write_only=True, required=False, allow_blank=False)
+    # webhook_service: null,  # Here we can use PrimaryKeyRelatedField so it will automagically do rbac/turn into object, I think, I'm actually not sure how to use this
+    # webhook_credential: null,  # Here we can use PrimaryKeyRelatedField so it will automagically do rbac/turn into object  I think, I'm actually not sure how to use this
+    skip_tags: serializers.CharField(write_only=True, required=False, allow_blank=False)
+    job_tags: serializers.CharField(write_only=True, required=False, allow_blank=False)
+    is_bulk_job: serializers.BooleanField(default=True)
```
kdelee: @jainnikhil30 we shouldn't expose this to user -- we just should set it directly when we instantiate the WorkflowJob

jainnikhil30:
> * Figure out how to prohibit relaunch. Probably in access.py we can do check `is_bulk_job=True` and return PermissionDenied with specific message

@kdelee 
This is already taken care by
https://github.com/ansible/awx/blob/devel/awx/main/access.py#L2163-L2164

We dont attach a wfjt/jt to the bulk workflow job, which is why we dont see the relaunch option. But Superuser do see a relaunch option. Do we want to not show it for Superuser too ? If yes we can add a is_bulk_job check as well on Superuser ?

awx/main/access.py
```diff
@@ -2124,7 +2124,11 @@ class WorkflowJobAccess(BaseAccess):
     )
 
     def filtered_queryset(self):
-        return WorkflowJob.objects.filter(unified_job_template__in=UnifiedJobTemplate.accessible_pk_qs(self.user, 'read_role'))
+        return WorkflowJob.objects.filter(
+            Q(unified_job_template__in=UnifiedJobTemplate.accessible_pk_qs(self.user, 'read_role'))
+            | Q(created_by__in=str(self.user.id), is_bulk_job=True)
```
jbradberry: I think this is an error: `created_by__in=str(self.user.id)`

awx/api/urls/urls.py
```diff
@@ -136,6 +140,8 @@
     re_path(r'^activity_stream/', include(activity_stream_urls)),
     re_path(r'^workflow_approval_templates/', include(workflow_approval_template_urls)),
     re_path(r'^workflow_approvals/', include(workflow_approval_urls)),
+    re_path(r'^bulk/job_launch/$', BulkJobLaunchView.as_view(), name='bulk_job_launch'),
```
jbradberry, line 143: did I say this on the other PR?  Anyway, I think you should break out the bulk urls into their own urls module.

awx/main/access.py
jainnikhil30: @jbradberry you mean we shouldn't filter on created by ? or should I not cast the id to string ? 

I tried following in shell_plus:

```
In [3]: WorkflowJob.objects.filter(created_by__in="1")
Out[3]: <PolymorphicQuerySet [<WorkflowJob: 2023-01-09 10:11:37.789655+00:00-2-successful>, <WorkflowJob: 2023-01-09 10:16:07.036121+00:00-7-successful>, <WorkflowJob: 2023-01-09 10:17:39.563689+00:00-11-successful>, <WorkflowJob: 2023-01-09 10:19:51.923923+00:00-15-successful>, <WorkflowJob: 2023-01-09 10:22:42.696832+00:00-19-successful>, <WorkflowJob: 2023-01-09 10:23:24.124887+00:00-26-successful>, <WorkflowJob: 2023-01-13 15:19:42.646826+00:00-62-successful>, <WorkflowJob: 2023-01-16 14:05:15.208858+00:00-73-successful>]>
```

If I try it with int it gives me an error:

```
In [2]: WorkflowJob.objects.filter(created_by__in=1)

TypeError: 'int' object is not iterable
```

jainnikhil30: oh wait, I could do this 
```
In [4]: WorkflowJob.objects.filter(created_by_id=1)
Out[4]: <PolymorphicQuerySet [<WorkflowJob: 2023-01-09 10:11:37.789655+00:00-2-successful>, <WorkflowJob: 2023-01-09 10:16:07.036121+00:00-7-successful>, <WorkflowJob: 2023-01-09 10:17:39.563689+00:00-11-successful>, <WorkflowJob: 2023-01-09 10:19:51.923923+00:00-15-successful>, <WorkflowJob: 2023-01-09 10:22:42.696832+00:00-19-successful>, <WorkflowJob: 2023-01-09 10:23:24.124887+00:00-26-successful>, <WorkflowJob: 2023-01-13 15:19:42.646826+00:00-62-successful>, <WorkflowJob: 2023-01-16 14:05:15.208858+00:00-73-successful>]>
```

I think thats better right ?

jbradberry: @jainnikhil30 the `__in` operator means that you are providing a sequence (usually a list).  By passing it a string, you are implying that each character in the string is a entry that you are looking for.  Not good.

I don't believe you need _any_ operator.  Try just `.filter(created_by=1)`.

jbradberry: If that doesn't work, your `created_by_id=1` is right thing.

kdelee:
closing in favor of https://github.com/ansible/awx/pull/13462

PR #13361, opened 2022-12-21
[collection] Run sanity tests outside of our container

relrod:
##### SUMMARY

Also just ignore one sanity test for the export module, instead of ignoring all of them.

Also use latest ansible-test, and make it work on GHA (by using podman instead of docker).

Maybe a follow up to https://github.com/ansible/awx/pull/13356

##### ISSUE TYPE

##### COMPONENT NAME

 - Collection


relrod:
CC @felixfontein for feedback

felixfontein (approved):

Makefile
```diff
@@ -339,7 +339,7 @@ test_collection_sanity:
 	if ! [ -x "$(shell command -v ansible-test)" ]; then pip install ansible-core; fi
 	ansible --version
 	COLLECTION_VERSION=1.0.0 make install_collection
-	cd $(COLLECTION_INSTALL) && ansible-test sanity --exclude=plugins/modules/export.py
+	cd $(COLLECTION_INSTALL) && ansible-test sanity --docker
```
gundalow: Nice `--docker` means we have the exact dependencies & Python versions installed as ansible-test requires.

awx_collection/tests/sanity/ignore-2.14.txt
```diff
@@ -0,0 +1 @@
+plugins/modules/export.py validate-modules:nonexistent-parameter-documented  # needs awxkit to construct argspec
```
gundalow, line 1: Nice
For others, this means that `ansible-test sanity` does run against plugins/modules/export.py though we ignore the errors about plugin parameters not matching documentation.

This is better than `--exclude=plugins/modules/export.py` as we still check for other errors.

.github/workflows/ci.yml
```diff
@@ -148,3 +145,21 @@ jobs:
...
+
+  collection-sanity:
+    name: awx_collection sanity
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Upgrade ansible-core
```
AlanCoding, line 158: Isn't this installing it in the first place?

Makefile
AlanCoding: One request I'd like to make - I need a local debugging cycle, and it might be hard to do that with this. If you don't mind, I'd like to add a new Makefile arg, like `SANITY_OPTIONS` that will default to `--docker`, but in the CLI we can clear them out. Because this does a lot of stuff that's good to have automated for the feedback cycle, like installing and `cd`ing into the right directory, which is very annoying manually.

While, yes, sometimes I would like to run --docker locally, that takes a long time, and a parameter can also narrow the test subset.

.github/workflows/ci.yml
shanemcd: The Ubuntu GHA runner has Ansible installed by default.

gundalow, line 158:
```suggestion
      # The containers that GitHub Actions use have Ansible installed, so upgrade to make sure we have the latest version.
      - name: Upgrade ansible-core
```

shanemcd (approved):

AlanCoding:
Operator test failing due to Galaxy flake. Will try to resolve by closing and re-opening.

PR #13364, opened 2022-12-21
Consolidate Makefile flow for building and testing collection

AlanCoding:
##### SUMMARY
This is very minor followup cleanup after https://github.com/ansible/awx/pull/13361 and others.

Changes it so that `make awx_collection_build` will _always_ re-build, and then there are some cascading changes in other targets in response to that. If you use the target to install or build the collection, then it is assumed that those targets properly clean their target directories.

##### ISSUE TYPE

##### COMPONENT NAME
 - Collection

##### ADDITIONAL INFORMATION
The CI should clear up whether we have the right version number. Out of fear of breaking passing tests I didn't remove our hacks earlier. I hope this version makes more sense... and still runs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,6 +154,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Get tags so version is correctly built
+        run: git fetch --tags origin
+
       # The containers that GitHub Actions use have Ansible installed, so upgrade to make sure we have the latest version.
       - name: Upgrade ansible-core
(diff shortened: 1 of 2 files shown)


AlanCoding:
Not working! It seems that we don't have the version correctly set for some reason.

AlanCoding:
@shanemcd do you have any ideas? Otherwise I'm going to abandon this effort.

The `VERSION` variable in the Makefile is blank. I added a step to fetch the tags, the the CI output is showing that. Still blank.

PR #13366, opened 2022-12-21
bump receptorctl version to 1.3.0

fosterseth:
##### SUMMARY

mostly need to bump for this PR https://github.com/ansible/awx/pull/12961

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

### DIFF
--- a/requirements/requirements.in
+++ b/requirements/requirements.in
@@ -42,7 +42,7 @@ python-dsv-sdk
 python-tss-sdk==1.0.0
 python-ldap
 pyyaml
-receptorctl==1.2.3
+receptorctl==1.3.0
 schedule==0.6.0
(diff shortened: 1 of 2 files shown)


AlanCoding:
For completeness, I had to ask myself where `receptor` itself comes from.

https://github.com/ansible/awx/blob/e34ce8c7950fc883d75e25467621fab4d57ab9ec/tools/ansible/roles/dockerfile/templates/Dockerfile.j2#L176

https://github.com/ansible/awx/blob/ac9871b36fd0723466065307340648aca22742c4/Makefile#L43

https://quay.io/repository/ansible/receptor?tab=tags&tag=latest

With the _floating_ tag, we automatically get receptor upgrades when the image advances. So updating receptorctl is just a matter of keeping up.

AlanCoding (approved):

Issue #13371, opened 2022-12-22
Save button not responding on the Job Settings page


vidyanambiar:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

To address Jira ticket: https://issues.redhat.com/browse/AAP-7757

On the "Edit Job settings" page and the Save button doesn't respond when AWX_ISOLATION_BASE_PATH has been provided manually via the settings file.

### AWX version

2.3

### Select the relevant components

- [X] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

Steps to Reproduce

- Add this to /etc/tower/settings.py
AWX_ISOLATION_BASE_PATH = '/var/lib/awx/job_execution'
- On the UI go to Settings -> Job Settings -> Edit

See attached doc for detailed steps.
[Reproducer for AAP-7757.pdf](https://github.com/ansible/awx/files/10288556/Reproducer.for.AAP-7757.pdf)

### Expected results

Save button should work

### Actual results

Save button doesn't respond at all

PR #13372, opened 2022-12-22
Fix for Save button not responding on Job Settings page

vidyanambiar:
Signed-off-by: Vidya Nambiar <vnambiar@redhat.com>

Issue: https://github.com/ansible/awx/issues/13371

##### ISSUE TYPE

##### COMPONENT NAME

 - UI

##### ADDITIONAL INFORMATION

The reason for this issue is that when fields like AWX_ISOLATION_BASE_PATH and SCHEDULE_MAX_JOBS are set manually via a settings file, they are excluded from the Edit Job Settings form view. However, they were still marked as "required" fields and this was silently causing the "Save" to fail.

With this PR:
AWX_ISOLATION_BASE_PATH and SCHEDULE_MAX_JOBS are only marked as required fields if their respective keys are returned by the PUT OPTIONS request (indicating that they have not been previously manually set)

[Reproducer for AAP-7757.pdf](https://github.com/ansible/awx/files/10288974/Reproducer.for.AAP-7757.pdf)

### DIFF
--- a/awx/ui/src/screens/Setting/Jobs/JobsEdit/JobsEdit.js
+++ b/awx/ui/src/screens/Setting/Jobs/JobsEdit/JobsEdit.js
@@ -141,14 +141,14 @@ function JobsEdit() {
               <FormColumnLayout>
                 <InputField
                   name="AWX_ISOLATION_BASE_PATH"
-                  config={jobs.AWX_ISOLATION_BASE_PATH}
-                  isRequired
+                  config={jobs.AWX_ISOLATION_BASE_PATH ?? null}
+                  isRequired={Boolean(options?.AWX_ISOLATION_BASE_PATH)}
                 />
                 <InputField
                   name="SCHEDULE_MAX_JOBS"
-                  config={jobs.SCHEDULE_MAX_JOBS}
-                  type="number"
-                  isRequired
+                  config={jobs.SCHEDULE_MAX_JOBS ?? null}
+                  type={options?.SCHEDULE_MAX_JOBS ? 'number' : undefined}
+                  isRequired={Boolean(options?.SCHEDULE_MAX_JOBS)}
                 />
                 <InputField
                   name="DEFAULT_JOB_TIMEOUT"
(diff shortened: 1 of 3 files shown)


awx/ui/src/screens/Setting/Jobs/JobsEdit/JobsEdit.js
```diff
@@ -141,14 +141,14 @@ function JobsEdit() {
...
+                  config={jobs.AWX_ISOLATION_BASE_PATH ?? null}
+                  isRequired={Boolean(options?.AWX_ISOLATION_BASE_PATH)}
                 />
                 <InputField
                   name="SCHEDULE_MAX_JOBS"
-                  config={jobs.SCHEDULE_MAX_JOBS}
-                  type="number"
-                  isRequired
+                  config={jobs.SCHEDULE_MAX_JOBS ?? null}
+                  type={options?.SCHEDULE_MAX_JOBS ? 'number' : undefined}
```
vidyanambiar, line 150: If `type` is passed in all cases it will trigger validation of the InputField component and in the case where the field is absent (manually set from settings file) it will cause the Save button to fail. https://github.com/jaredpalmer/formik/issues/902

awx/ui/src/screens/Setting/shared/SharedFields.js
```diff
@@ -397,7 +397,10 @@ const InputField = ({ name, config, type = 'text', isRequired = false }) => {
 };
 InputField.propTypes = {
   name: string.isRequired,
-  config: shape({}).isRequired,
+  config: shape({}),
+};
+InputField.defaultProps = {
+  config: null,
```
vidyanambiar, line 403: Since `config` is not passed to `InputField` when values have been manually set previously, it should be optional. And defaultProps need to be set for optional fields. https://github.com/jsx-eslint/eslint-plugin-react/issues/1433

kialam (approved):
LGTM! Nice job debugging the source of the bug.

akus062381 (approved):

Issue #13231, opened 2022-11-24
awx.awx.group : preserve_existing_children doesn't work, only its alias preserve_existing_groups


mooky31:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

awx.awx.group : preserve_existing_children doesn't work, only its alias preserve_existing_groups

### AWX version

21.7.0

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [X] Collection
- [ ] CLI
- [ ] Other

### Installation method

minikube

### Modifications

no

### Ansible version

5.10

### Operating system

RHEL9

### Web browser

Firefox

### Steps to reproduce

Try to add subgroupB to a parentGroup with existing subgroupA, using preserve_existing_children

### Expected results

parentGroup contains SubgroupA and subgroupB

### Actual results

parentGroup only contains subgroupB

djyasin:
PR #13374

PR #13374, opened 2022-12-22
added fix for preserve existing children issue.

djyasin:
##### SUMMARY

Fix for :https://github.com/ansible/awx/issues/13231

##### ISSUE TYPE

##### COMPONENT NAME

 - Collection

##### AWX VERSION

```
21.10.3
```

### DIFF
--- a/awx_collection/plugins/modules/group.py
+++ b/awx_collection/plugins/modules/group.py
@@ -129,7 +129,6 @@ def main():
     state = module.params.pop('state')
     preserve_existing_hosts = module.params.get('preserve_existing_hosts')
     preserve_existing_children = module.params.get('preserve_existing_children')
-    preserve_existing_children = preserve_existing_children or module.params.get('preserve_existing_children')
     variables = module.params.get('variables')
 
     # Attempt to look up the related items the user specified (these will fail the module if not found)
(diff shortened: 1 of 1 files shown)


awx_collection/plugins/modules/group.py
```diff
@@ -129,6 +129,7 @@ def main():
     state = module.params.pop('state')
     preserve_existing_hosts = module.params.get('preserve_existing_hosts')
     preserve_existing_children = module.params.get('preserve_existing_groups')
+    preserve_existing_children = preserve_existing_children or module.params.get('preserve_existing_children')
```
relrod: If a task in a playbook uses a module argument's alias, the value given to that alias will [get assigned](https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/common/parameters.py#L235) to the original argument that declared that alias in the module's argument spec.

In other words, if `preserve_existing_children` declares a `preserve_existing_groups` alias and someone passes in `preserve_existing_groups: foo`, then `"foo"` will live in `module.params['preserve_existing_children']`.

So I think an alternative/better fix here would be to just use the _actual_ parameter name instead of its alias when we assign the value to `preserve_existing_children` on line 131.

djyasin: Thank you so much, @relrod! That makes sense to me! I have revised this per your suggestion. Please let me know if I need to make any additional changes!

relrod: After that change, you should be able to remove (new) line 132 entirely :)

djyasin: Oops, thank you!

relrod (approved):
This LGTM and should fix the issue.

jay-steurer (approved):

PR #13382, opened 2022-12-29
set the real ip to logs  (login)

jordicasas1987:
##### SUMMARY
The value of (settings.REMOTE_HOST_HEADERS) is used to replace the REMOTE_ADDR.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API
 - Settings
 - Middleware
 
##### AWX VERSION
```
awx: 21.10.3.dev9+g132158586e
```

### DIFF
--- a/awx/api/generics.py
+++ b/awx/api/generics.py
@@ -145,19 +145,6 @@ class APIView(views.APIView):
         if getattr(settings, 'SQL_DEBUG', False):
             self.queries_before = len(connection.queries)
 
-        # If there are any custom headers in REMOTE_HOST_HEADERS, make sure
-        # they respect the allowed proxy list
-        if all(
-            [
-                settings.PROXY_IP_ALLOWED_LIST,
-                request.environ.get('REMOTE_ADDR') not in settings.PROXY_IP_ALLOWED_LIST,
-                request.environ.get('REMOTE_HOST') not in settings.PROXY_IP_ALLOWED_LIST,
-            ]
-        ):
-            for custom_header in settings.REMOTE_HOST_HEADERS:
-                if custom_header.startswith('HTTP_'):
-                    request.environ.pop(custom_header, None)
-
         drf_request = super(APIView, self).initialize_request(request, *args, **kwargs)
         request.drf_request = drf_request
         try:
--- a/awx/api/generics.py
+++ b/awx/api/generics.py
@@ -88,15 +88,30 @@ class LoggedLoginView(auth_views.LoginView):
 
     def post(self, request, *args, **kwargs):
         ret = super(LoggedLoginView, self).post(request, *args, **kwargs)
+
+        client_addr = request.META.get('REMOTE_ADDR', None)
+
+        # If an authorized header contains an IP, override the client_addr
+        for custom_header in settings.REMOTE_HOST_HEADERS:
+            for value in request.META.get(custom_header, '').split(','):
+                value = value.strip()
+                if value:
+                    client_addr = value
+                    break
+
+            else:
+                continue
+            break
+
         if request.user.is_authenticated:
-            logger.info(smart_str(u"User {} logged in from {}".format(self.request.user.username, request.META.get('REMOTE_ADDR', None))))
+            logger.info(smart_str(u"User {} logged in from {}".format(self.request.user.username, client_addr)))
             ret.set_cookie('userLoggedIn', 'true')
             ret.setdefault('X-API-Session-Cookie-Name', getattr(settings, 'SESSION_COOKIE_NAME', 'awx_sessionid'))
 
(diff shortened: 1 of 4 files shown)


djyasin:
Hi @jordicasas1987!
Thank you for contributing. We would like to gather a bit more information from you regarding this particular change. Could you please elaborate on the desired outcome of this change and the intended effect? 

Thank you again for your contribution!

jordicasas1987:
Hi @djyasin!

Currently, when we asking for the IP of the client making the connection, if the service uses a proxy, the IP returned is the IP of the proxy and not the IP of the client.
Ex: https://github.com/ansible/awx/blob/devel/awx/api/generics.py#L94

The idea is to replace the validation that is done in the API
https://github.com/ansible/awx/blob/devel/awx/api/generics.py#L151
and put it in the middleware so that they are executed for all the requests.


Regards,

samehif:
Could someone please review and merge this fix??

TheRealHaoLiu:
currently this PR is failing the functional test 
```
=================================== FAILURES ===================================
____________________________ test_proxy_ip_allowed _____________________________
[gw1] linux -- Python 3.9.16 /var/lib/awx/venv/awx/bin/python3.9
  File "/awx_devel/awx/main/tests/functional/api/test_generic.py", line 32, in test_proxy_ip_allowed
    assert 'HTTP_X_FROM_THE_LOAD_BALANCER' not in middleware.environ
AssertionError: assert 'HTTP_X_FROM_THE_LOAD_BALANCER' not in {'CONTENT_TYPE': 'application/octet-stream', 'HTTP_COOKIE': '', 'HTTP_X_FROM_THE_LOAD_BALANCER': 'some-actual-ip', 'PATH_INFO': '/api/v2/settings/system/', ...}
 +  where {'CONTENT_TYPE': 'application/octet-stream', 'HTTP_COOKIE': '', 'HTTP_X_FROM_THE_LOAD_BALANCER': 'some-actual-ip', 'PATH_INFO': '/api/v2/settings/system/', ...} = <awx.main.tests.functional.api.test_generic.test_proxy_ip_allowed.<locals>.HeaderTrackingMiddleware object at 0x7fd8a4333190>.environ
 ```
 
 and also failing api-lint (a simple `make black` should fix the api-lint problem`

**Comment by djyasin** [comment]
@jordicasas1987 it looks like we still have a few failing tests. It does look like they have different failures. Would you mind reviewing those failures to see if you can get those tests to a passing state? It also looks like you will need to rebase to fix some of the failing tests.

File: awx/api/generics.py [hunk 1]
```diff
@@ -88,15 +88,30 @@ def get(self, request, *args, **kwargs):
 
     def post(self, request, *args, **kwargs):
         ret = super(LoggedLoginView, self).post(request, *args, **kwargs)
+
+        client_addr = request.META.get('REMOTE_ADDR', None)
+
+        # If an authorized header contains an IP, override the client_addr
+        for custom_header in settings.REMOTE_HOST_HEADERS:
```
jjwatt, line 95: I'm not familiar with this code base, but I don't see how this loop will only apply to IP addresses. Won't this loop through every REMOTE_HOST_HEADERS? Like if it were `['REMOTE_ADDR', 'REMOTE_HOST']`, then would it not loop through those? The way the loop is constructed, it looks like it will overwrite `client_addr` with the first non-falsey value it finds in the list of `REMOTE_HOST_HEADERS`. Can you explain what is going on here and/or refactor this loop?

awx/api/generics.py
```diff
@@ -88,15 +88,30 @@ def get(self, request, *args, **kwargs):
...
+            for value in request.META.get(custom_header, '').split(','):
+                value = value.strip()
```
jjwatt, line 97: I don't like modifying the loop iterator, `value`, directly, especially if you're not going to use it. Unless `request` object has a special get that makes copies, you're not getting a value, you're getting a reference and if you change it it will change in the `request` object, too. And that might be kind of unexpected, no? It can certainly cause gotchas. EDIT: OK, I now notice that you're using `split()` and I guess that makes new objects. However, beyond that I don't understand the use of `split(',')` here or how we know that the first value will be the right `client_addr`.

jjwatt (commented):
PTAL

jjwatt (approved):

iTeV:
Sorry to bump this PR, but I was wondering: Is there an ETA regarding when this PR will be merged? I've stumbled upon the same issue where the logging shows the IP of the proxy. I've tested the changes in this PR locally, and it seemed to work just fine (I was getting/seeing the correct IP-Addresses after I've changed the remote host headers configuration)

if there is some more testing required, please let me know! I'd be happy to help :)

PR #13385, opened 2022-12-30
Bulk job launch

jainnikhil30:
##### SUMMARY
This is continuation to https://github.com/ansible/awx/pull/13358
Add more fields to the Bulk Job.

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -4463,8 +4463,7 @@ class BulkJobNodeSerializer(serializers.Serializer):
     inventory = serializers.IntegerField(required=False, min_value=1)
     credentials = serializers.ListField(child=serializers.IntegerField(min_value=1), required=False)
     identifier = serializers.CharField(required=False, write_only=True, allow_blank=False)
-
-    #  labels = serializers.ListField(child=serializers.IntegerField(min_value=1), required=False)
+    labels = serializers.ListField(child=serializers.IntegerField(min_value=1), required=False)
     #  instance_groups = serializers.ListField(child=serializers.IntegerField(min_value=1), required=False)
     #  execution_environment = serializers.IntegerField(required=False, min_value=1)
     #
@@ -4488,6 +4487,7 @@ class BulkJobNodeSerializer(serializers.Serializer):
             'inventory',
             'credentials',
             'limit',
+            'labels',
             #   'char_prompts',
             #   'diff_mode',
             #   'extra_data',
@@ -4502,7 +4502,6 @@ class BulkJobNodeSerializer(serializers.Serializer):
             #   'timeout',
             #   'verbosity',
             # these are related objects and we need to add extra validation for them in the parent BulkJobLaunchSerializer
-            #   'labels',
             #   'execution_environment',
             #   'instance_groups',
         )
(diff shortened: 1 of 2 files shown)


awx/api/serializers.py
```diff
@@ -4563,13 +4573,34 @@ def validate(self, attrs):
                     not_allowed = requested_use_credentials - accessible_use_credentials
                     raise serializers.ValidationError(_(f"Credentials {not_allowed} not found."))
 
+            if requested_use_labels:
+                accessible_use_labels = {tup[0] for tup in Label.objects.all()}
```
kdelee, line 4577: is there an equivalent `Label.accessible_pk_qs(request.user, 'use_role').all()`? I guess question is -- what is the role called and is `accessible_pk_qs` method defined for `Label`
```suggestion
                accessible_use_labels = {tup[0] for tup in Label.accessible_pk_qs(request.user, 'use_role').all()}
```

I see you using that method elsewhere -- so I'm guessing maybe it doesn't exist or maybe its like EEs and we need to look at what labels are defined for your organizations? Or maybe everyone has access to all labels -- would be good to clarify that in a comment if that is the case. If that is the case, maybe we could avoid fetching every single Label by doing a `Label.objects.filter(id__in=requested_use_labels).all()`

awx/api/serializers.py
```diff
@@ -4563,13 +4573,34 @@ def validate(self, attrs):
...
+                if requested_use_labels - accessible_use_labels:
+                    not_allowed = requested_use_labels - accessible_use_labels
+                    raise serializers.ValidationError(_(f"Labels {not_allowed} not found"))
+
+            if requested_use_instance_groups:
+                # only org admins are allowed to see instance groups
```
kdelee, line 4583: That is interesting. What about job template admins, can they not set IGs when they create a new JT?

awx/api/serializers.py
```diff
@@ -4563,13 +4573,34 @@ def validate(self, attrs):
...
         # all of the unified job templates and related items have now been checked, we can now grab the objects from the DB
         # TODO: As we add more related objects like Label, InstanceGroup, etc we need to add them here
         objectified_jobs = []
         key_to_obj_map = {
             "unified_job_template": {obj.id: obj for obj in UnifiedJobTemplate.objects.filter(id__in=requested_ujts)},
             "inventory": {obj.id: obj for obj in Inventory.objects.filter(id__in=requested_use_inventories)},
             "credentials": {obj.id: obj for obj in Credential.objects.filter(id__in=requested_use_credentials)},
+            "labels": {obj.id: obj for obj in Label.objects.filter(id__in=requested_use_labels)},
+            "instance_groups": {obj.id: obj for obj in InstanceGroup.objects.filter(id__in=requested_use_instance_groups)}
         }
```
kdelee, line 4604: TODO add ExecutionEnvironment to this map

awx/api/serializers.py
jainnikhil30: There is no `Label.accessible_pk_qs()` because accessible_pk_qs is a method from ResourceMixin class which Labels don't extend. Yeah it seems a user/member of an org can see all the labels for that org, similar to what EE does as well. So we need to may be just filter them on the request user org.

awx/api/serializers.py
jainnikhil30: Yup only org admins. Not even workflow admin or job template admin.

kdelee (approved):

kdelee:
going to merge to the base branch `bulk_job_launch` and we can continue to work from there

PR #13398, opened 2023-01-04
Workaround for events with NUL char, touch up error loop

AlanCoding:
##### SUMMARY
This fixes an error where some events would not save due to having the `0x00` character which errors in postgres. This adds a line to replace it with empty text.

Hitting that kind of event put us in an infinite error loop, so this change makes a number of changes to prevent similar loops. The showcase example is a negative counter, this is not realistic in the real world but works for unit tests.

These error loop fixes seek to establish the cases where we clear the buffer Some logic is removed from the outer loop, with the idea that ensure_connection will better distinguish flake.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
I had a laundry list of maybe 5 things I wanted to fix with this (see unit tests), and a good number of them could solve the observed bug by itself. However, I don't want to let a good crisis go to waste, and having a reproducer for this issue revealed a lot of problems with the error handling to me. With the current patch, I'm fairly happy that this checks off all the objectives I had. Including:

 - Catch duplicate key violations (seen from production logs) and remove them from the buffer (only in bulk_create, so solution doesn't follow very obviously)
 - Remove any successful saves from the buffer - I was able to show that the 0x00 error was leading to keeping _good_ events around in the buffer, which could lead to duplicates (not unique on uuid :scream: )
 - Clear the buffer if we have an unknown error in the outer loop - this may be controversial
 - Track a retry counter for individual events, which becomes our metric for "giving up" on them
 - Get rid of the "rule of 5" which created disastrous bugs and I introduced it - this relies on `ensure_connection` instead
 - Catch the specific NUL character and sanitize it

### DIFF
--- a/awx/main/dispatch/worker/callback.py
+++ b/awx/main/dispatch/worker/callback.py
@@ -3,14 +3,12 @@ import logging
 import os
 import signal
 import time
-import traceback
 import datetime
 
 from django.conf import settings
 from django.utils.functional import cached_property
 from django.utils.timezone import now as tz_now
-from django.db import DatabaseError, OperationalError, transaction, connection as django_connection
-from django.db.utils import InterfaceError, InternalError
+from django.db import transaction, connection as django_connection
 from django_guid import set_guid
 
 import psutil
@@ -64,6 +62,7 @@ class CallbackBrokerWorker(BaseWorker):
     """
 
     MAX_RETRIES = 2
+    INDIVIDUAL_EVENT_RETRIES = 3
     last_stats = time.time()
     last_flush = time.time()
     total = 0
@@ -164,38 +163,46 @@ class CallbackBrokerWorker(BaseWorker):
                     else:  # only calculate the seconds if the created time already has been set
                         metrics_total_job_event_processing_seconds += e.modified - e.created
                 metrics_duration_to_save = time.perf_counter()
+                saved_events = []
                 try:
                     cls.objects.bulk_create(events)
                     metrics_bulk_events_saved += len(events)
+                    saved_events = events
+                    self.buff[cls] = []
                 except Exception as exc:
-                    logger.warning(f'Error in events bulk_create, will try indiviually up to 5 errors, error {str(exc)}')
+                    # if the database is flaking, let ensure_connection throw a general exception
+                    # will be caught by the outer loop, which goes into a proper sleep and retry loop
+                    django_connection.ensure_connection()  # INCLUDE in final patch
+                    logger.warning(f'Error in events bulk_create, will try indiviually, error: {str(exc)}')
                     # if an exception occurs, we should re-attempt to save the
                     # events one-by-one, because something in the list is
                     # broken/stale
-                    consecutive_errors = 0
-                    events_saved = 0
                     metrics_events_batch_save_errors += 1
-                    for e in events:
+                    for e in events.copy():
+                        # special sanitization logic for postgres treatment of NUL 0x00 char
+                        if "\x00" in e.stdout:
+                            e.stdout = e.stdout.replace("\x00", "")
+
                         try:
                             e.save()
-                            events_saved += 1
-                            consecutive_errors = 0
+                            metrics_singular_events_saved += 1
+                            events.remove(e)  # Importantly, remove successfully saved events from the buffer
+                            saved_events.append(e)
                         except Exception as exc_indv:
-                            consecutive_errors += 1
-                            logger.info(f'Database Error Saving individual Job Event, error {str(exc_indv)}')
-                        if consecutive_errors >= 5:
-                            raise
-                    metrics_singular_events_saved += events_saved
-                    if events_saved == 0:
-                        raise
+                            retry_count = getattr(e, '_retry_count', 0) + 1
+                            e._retry_count = retry_count
+                            logger.info(f'Database Error Saving individual Event uuid={e.uuid} try={retry_count}, error: {str(exc_indv)}')
+                            if retry_count >= self.INDIVIDUAL_EVENT_RETRIES:
+                                logger.error(f'Hit max retries saving individual Event error: {str(exc_indv)}\ndata:\n{e.__dict__}')
+                                events.remove(e)
+
                 metrics_duration_to_save = time.perf_counter() - metrics_duration_to_save
-                for e in events:
+                for e in saved_events:
                     if not getattr(e, '_skip_websocket_message', False):
                         metrics_events_broadcast += 1
                         emit_event_detail(e)
                     if getattr(e, '_notification_trigger_event', False):
                         job_stats_wrapup(getattr(e, e.JOB_REFERENCE), event=e)
-            self.buff = {}
             self.last_flush = time.time()
             # only update metrics if we saved events
             if (metrics_bulk_events_saved + metrics_singular_events_saved) > 0:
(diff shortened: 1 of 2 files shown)


awx/main/dispatch/worker/callback.py
```diff
@@ -64,6 +62,7 @@ class CallbackBrokerWorker(BaseWorker):
     """
 
     MAX_RETRIES = 2
+    INDIVIDUAL_EVENT_RETRIES = 3
```
gamuniz, line 65: do we want to mirror the 5 attempts or we okay with lowering this?

AlanCoding: It's not that it was trying 5 times before, it would allow 5 errors (from separate events) before giving up saving the working buffer. If it hit over 5 errors, then it would retry the whole batch next time, but if it hit fewer than 5 errors, it would give up on those events right away. So call this a geometric average of 1 and infinity. 3 sounds good. This is giving 3 retries _after we check that postgres connection is working_ so that seems generous to me.

Given the reduced tolerance I'm putting in for the outer loop (adding `self.buffer = {}`), I believe a relevant argument would be to increase `MAX_RETRIES`.

gamuniz: 👍

awx/main/dispatch/worker/callback.py
```diff
@@ -164,38 +163,46 @@ def flush(self, force=False):
...
                     # if an exception occurs, we should re-attempt to save the
                     # events one-by-one, because something in the list is
                     # broken/stale
-                    consecutive_errors = 0
-                    events_saved = 0
                     metrics_events_batch_save_errors += 1
-                    for e in events:
+                    for e in events.copy():
+                        # special sanitization logic for postgres treatment of NUL 0x00 char
+                        if "\x00" in e.stdout:
```
shanemcd: Kind of nervous about the performance implications of scanning every single byte of stdout here. Would it be better to catch ValueError, check if it contains "A string literal cannot contain NUL", and then fallback to doing this?

AlanCoding: This is already in the exception block for `bulk_create`. So this only triggers when an error happens, but it runs against all events, including the perfectly good ones. Event errors (outside of database problems) _should_ be extremely rare, but things don't always work as they should.

To be more honest about things - the handling was put earlier than it should have been to make the unit test possible.

To follow the logic, this is where I see it going:

```diff
diff --git a/awx/main/dispatch/worker/callback.py b/awx/main/dispatch/worker/callback.py
index 2bb0699d35..9d6f988af7 100644
--- a/awx/main/dispatch/worker/callback.py
+++ b/awx/main/dispatch/worker/callback.py
@@ -179,10 +179,6 @@ class CallbackBrokerWorker(BaseWorker):
                     # broken/stale
                     metrics_events_batch_save_errors += 1
                     for e in events.copy():
-                        # special sanitization logic for postgres treatment of NUL 0x00 char
-                        if "\x00" in e.stdout:
-                            e.stdout = e.stdout.replace("\x00", "")
-
                         try:
                             e.save()
                             metrics_singular_events_saved += 1
@@ -191,6 +187,11 @@ class CallbackBrokerWorker(BaseWorker):
                         except Exception as exc_indv:
                             retry_count = getattr(e, '_retry_count', 0) + 1
                             e._retry_count = retry_count
+
+                            # special sanitization logic for postgres treatment of NUL 0x00 char
+                            if (retry_count == 1) and isinstance(exc_indv, ValueError) and ("\x00" in e.stdout):
+                                e.stdout = e.stdout.replace("\x00", "")
+
                             logger.info(f'Database Error Saving individual Event uuid={e.uuid} try={retry_count}, error: {str(exc_indv)}')
                             if retry_count >= self.INDIVIDUAL_EVENT_RETRIES:
                                 logger.error(f'Hit max retries saving individual Event error: {str(exc_indv)}\ndata:\n{e.__dict__}')
```

I don't have any argument against this. But doing this, I would give up on the corresponding test, because I know this would get too complicated to realistically maintain.

AlanCoding: Pushed a commit where I did this but kept the test, adding a couple of mocks instead. The test isn't quite as bad as I feared, so we'll go with it.

AlanCoding:
Integration tests look good for this. The only potential option review comment is the idea of increasing MAX_RETRIES from 2 to something higher, but I don't see that making a huge resiliency difference, since it already waits for >1 min (which is longer than the dispatcher right now).

awx/main/dispatch/worker/callback.py
shanemcd: Thanks for clarifying Alan.

awx/main/tests/functional/commands/test_callback_receiver.py
```diff
@@ -24,3 +32,108 @@ def test_wrapup_does_send_notifications(mocker):
...
+
+    @classmethod
+    def from_url(cls, *args, **kwargs):
+        return cls()
+
+    def pipeline(self):
+        return self
+
+
+class TestCallbackBrokerWorker(TransactionTestCase):
```
jbradberry, line 55: Huh, does mixing this with pytest functionality actually work?  I guess it must, since a quick search indicates that we do this elsewhere.

awx/main/dispatch/worker/callback.py
```diff
@@ -164,38 +163,48 @@ def flush(self, force=False):
...
-                    events_saved = 0
                     metrics_events_batch_save_errors += 1
-                    for e in events:
+                    for e in events.copy():
                         try:
                             e.save()
-                            events_saved += 1
-                            consecutive_errors = 0
+                            metrics_singular_events_saved += 1
+                            events.remove(e)
```
jbradberry, line 185: Since `events` is a list, `.remove` is going to be doing a linear scan each loop iteration.  I'm not thrilled with doing this.

AlanCoding: Yes, this is technically valid.

If you look at the definition of `.flush`, you can see a hard limit of `1000` events in the list. This method is called for every event received, so there's not a sneaky way for it to go over that number. The list is always 1,000 elements or less.

These are objects that we're removing, so I believe the hashing is the general universal python indexing by memory address. That being the case, this should be a completely fair calculation of the upper limit time:

```python
import random
import time

alist = [object() for i in range(1000)]

def destroy_list():
  for obj in alist.copy():
    alist.remove(obj)

s = time.time(); destroy_list(); print(time.time() - s)
```

That runs in <1 ms for me. Database interactions are typically 5 ms, so I would gladly accept this upper limit, if my evaluation is correct.

I should test this again but with Django objects, just in case they override the hash method.

AlanCoding: Testing with saved `InventoryUpdateEvent` objects, I get 1.1 ms for the same thing. Higher, but still not a performance concern.

awx/main/tests/functional/commands/test_callback_receiver.py
AlanCoding: Background I maybe should have put as comments - the callback receiver code makes copious use of catching database errors, and then taking followup actions. This does not play well with transactions, because you get the error that you are in a broken transactions so you can't make subsequent queries. That error handling is the core behavior that's being tested here, so I can't workaround other than running the test in auto-commit mode. I get that with the `TransactionTestCase`.

Not quite sure what the collateral damage is, although I suspect that it affects what `.sqlite3` files are created.

awx/main/dispatch/worker/callback.py
jbradberry: As long as this method isn't somewhere in the hot loop, I guess I'm ok with it then.

Note: `list.remove` isn't going to have anything to do with hashing.  It is going to be checking for outright equality.

jbradberry (approved):

awx/main/dispatch/worker/callback.py
AlanCoding: Haven't been able to find reference material on that yet, but what you say suggests this should be hitting the Django model `__eq__`, which could explain the timing difference.

https://github.com/django/django/blob/78f163a4fb3937aca2e71786fbdd51a0ef39629e/django/db/models/base.py#L593-L601

From some quick poking at the subject, it sounds like the performance crunch isn't really from _searching_ for an element, but that the act of removing it requires shifting positions of all the remaining pointers. I'm sure that could be optimized, but I'm still pretty convinced that the worst-case timing is already just fine.

jay-steurer (approved):

Issue #12364, opened 2022-06-13
Support of unified requirements.yml file with roles/collections do not work


dacofr:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

Even what is mentioned in this issue : https://github.com/ansible/awx/issues/6270, support of unified requirements.yml at top level do not work.

In file https://github.com/ansible/awx/blob/devel/awx/playbooks/project_update.yml#L214 
we are looking at the requirements.yml file at the root folder, but we install it with `ansible-galaxy collection install`
(https://github.com/ansible/awx/blob/devel/awx/playbooks/project_update.yml#L204)

### AWX version

20.1

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

kubernetes

### Modifications

no

### Operating system

Linux

### Steps to reproduce

- Create a repository with a requirements.yml at **the root folder** with this content :
```
---
collections:
  - name: community.general
    version: "4.8.1"
  - name: community.hashi_vault
    version: "2.5.0"

roles:
  - src: Datadog.datadog
    name: Datadog.datadog
    version: 4.15.0
```

- Create a projet
- Synchronize it

### Expected results

Collections and roles are installed 

### Actual results

Only collections are installed

AlanCoding:
It sounds like what you want is:

```diff
diff --git a/awx/playbooks/project_update.yml b/awx/playbooks/project_update.yml
index 6d90b8d586..9da65be5d2 100644
--- a/awx/playbooks/project_update.yml
+++ b/awx/playbooks/project_update.yml
@@ -193,6 +193,8 @@
       with_fileglob:
         - "{{project_path|quote}}/roles/requirements.yaml"
         - "{{project_path|quote}}/roles/requirements.yml"
+        - "{{project_path|quote}}/requirements.yaml"
+        - "{{project_path|quote}}/requirements.yml"
       changed_when: "'was installed successfully' in galaxy_result.stdout"
       environment: "{{ galaxy_task_env }}"
       when: roles_enabled|bool
```

[...]

PR #13404, opened 2023-01-05
Added additional requirements paths into project_update.yml playbook

dulin-gnet:
##### SUMMARY

There is logic issue with installing dependencies by ansible tower as single file requirements.yml can  contains roles and collections. (https://docs.ansible.com/ansible/latest/galaxy/user_guide.html)  If you are using ansible_builder you can define single file as dependency. 

This fix also resolve the problem with wrong displayed file in UI as it suggests it should be in main directory. 

image

image

##### ISSUE TYPE

##### COMPONENT NAME

 - CLI
 - UI
 - Other

### DIFF
--- a/awx/playbooks/project_update.yml
+++ b/awx/playbooks/project_update.yml
@@ -214,6 +214,8 @@
       with_fileglob:
         - "{{project_path|quote}}/roles/requirements.yaml"
         - "{{project_path|quote}}/roles/requirements.yml"
+        - "{{project_path|quote}}/requirements.yaml"
+        - "{{project_path|quote}}/requirements.yml"
       changed_when: "'was installed successfully' in galaxy_result.stdout"
       environment: "{{ galaxy_task_env }}"
       when: roles_enabled|bool


AlanCoding:
Connect https://github.com/ansible/awx/issues/12364

The comment I left there remains an accurate summary.

john-westcott-iv:
Thank you so much for your contribution. After a review with additional engineers we think this PR should be changed similar to the comments made by @dacofr in the issue @AlanCoding linked. Our idea is that we would have three tasks:
1) The role install task that would install from roles/requirements.y[a]ml
2) The collection install task that would install from collections/requirements.y[a]ml
3) A unified install task that would install from requirement.y[a]ml. This could have a when clause (like the collections task) that would require ansible >= 2.10 (which was when the unified installer came out).

Please let us know if you have any questions on this and if you are willing to continue working on this issue with us or not.

TheRealHaoLiu:
Due to inactivity @john-westcott-iv have volunteered to take up this issue and will submit an follow up PR to fix the problem

Issue #5743, opened 2020-01-22
RFE: Define field for CyberArk AIM lookup 


DonaldKallman:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Allow the user to define the field in the JSON response to be used for the username and password values.  This is a unique scenario to object types that do not map to the standard Content and UserName fields (like a cloud service).  For example, I want the value of Content for the password and the value of AWSAccessKeyID as the username from the below:

```
{
  "Content": "****************************************",
  "Name": "Cloud Service-svc-cyberark-np1",
  "CreationMethod": "PVWA",
  "SequenceID": "1",
  "LastTask": "VerifyTask",
  "CPMStatus": "success",
  "Folder": "Root",
  "Safe": "Ansible_Tower_Test",
  "LastSuccessVerification": "1579694399",
  "AWSAccessKeyID": "AKIA****************",
  "DeviceType": "Cloud Service",
  "RetriesCount": "-1",
  "LastSuccessChange": "1579558784",
  "UserName": "svc-cyberark",
  "PasswordChangeInProcess": "False"
}
```

kylecurtis-od:
I believe this has to do with the tower config not handling two different credential types from the cyberark side. 
In the scenario where the credential in the cyberark safe is, "Device Type: Operating System", making a call to the cyberkark 'password' api endpoint will work fine. However, if you make that same call to the cyberark api, and you want a cyberark stored account that is of the device type, "Cloud Service", that is different.  There cyberark stores the AWS ACCESS KEY ID into the field, "AWSAccessKeyID".  A an api call to that endpoint for a cred of that type, does return the json mentioned above, but tower isn't mapping it back properly so you can use it in an ansible tower AWS Credential type.

jakemcdermott:
You'll need to change how you're storing this data in Cyberark.

Tower will map _strings_ (returned by your secret management system) to input fields. It does not process or attempt to remove fields from strings that contain json.

Using the example above, If you want to map `Content` to a specific field tower, you'll need to store that value _on its own_ inside your Cyberark server and not as part of a json object.

jakemcdermott:
:man_facepalming: Actually, I see what you mean now. You're referring to this field: https://github.com/ansible/awx/blob/devel/awx/main/credential_plugins/aim.py#L100, which we use to get the secret string.

jakemcdermott:
[...]

PR #13411, opened 2023-01-06
Fixes #13402 allow user defined key retrieval from CYBR

### LINKED ISSUE #13402: Access to all keys out of the CyberArk object

### Feature type

Enhancement to Existing Feature

### Feature Summary

https://issues.redhat.com/browse/AAPRFE-28

We want to pull all the Keys out of the CyberArk object instead of just the content key or the content value. When we run the API call from the playbook, CyberArk presents all the keys and values associated with that object. We can set the fact and use the keys and values as variables in the playbook. But when we use the custom credential it only gives us the content value and does not give us any other keys and values. We want to be able to use custom credential to pull all the keys and values associated with an object and assigned into variables in AAP natively using the credential plugin.

This feature request is for changes to the credential lookup plugin as used in AAP2.

- [X] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

1. Configure a credential using the credential type "CyberArk Central Credential Provider Lookup".
2. Only "Content" key value from JSON response is returned.  "Content" is the password value of the account object.

### Current results

The credential plugin "CyberArk Central Credential Provider Lookup" is hardcoded to only return the "Content" key value from a JSON response that includes more keys that could be utilized.

### Sugested feature result

The suggested feature would include an additional field in the metadata of the credential plugin to allow the user to specify a different key value to utilize outside of the "Content" which should still remain as the default key.

### Additional information

Spoke with Oleksii Baranov on 01-05-2023 about how to include this feature.  Being developed by CyberArk's @infamousjoeg.  PR to come shortly.

For more information, please reference the Issue at https://issues.redhat.com/browse/AAPRFE-28.

infamousjoeg:
##### SUMMARY
An enhancement request was made for the CyberArk Central Credential Provider Lookup to support user-defined key retrieval instead of the hardcoded `Content` retrieval which equates to a password or key value.  This would allow users to specify other values to use, such as Username, Address, Port, Database, etc.

This PR adds a string field to the CyberArk Central Credential Provider Lookup credential plugin metadata for `Object Property` to be defined by the user.  If no value is provided, the default value is `Content` which falls in line with behavior prior to this PR.

related #13402
related #5743 

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - UI

##### AWX VERSION
```
awx: 0.1.dev32738+gc34509f
```

##### ADDITIONAL INFORMATION
Screenshots from testing:
image
image
image


awx/main/credential_plugins/aim.py
```diff
@@ -74,6 +80,7 @@ def aim_backend(**kwargs):
     app_id = kwargs['app_id']
     object_query = kwargs['object_query']
     object_query_format = kwargs['object_query_format']
+    object_property = kwargs['object_property']
```
obaranov: @infamousjoeg will that work if I do not provide the `object_property` key in metadata: 

```
{
  "object_query": "Safe=TestSafe;Object=testAccountName123",
  "object_query_format": "Exact",
}
```
I'm just thinking about backward compatibility with the possible existing credentials without `object_property`?

infamousjoeg: Yes, it defaults to using the "Content" key value instead which is how it works today.  So, a change will not break current integrations.

obaranov: I think this comment is still actual. I think we need to keep compatibility with old API calls. So when we call API and we do not provide object_property in the inputs we gonna fail here saying that the object_property is not in the kwargs. 

I think we should be more careful here and use the default value: 
```
object_property = kwargs.get('object_property', '')
```

obaranov:
I verified the fix with the docker development environment. Everything works correctly except this moment: https://github.com/ansible/awx/pull/13411#discussion_r1111822088

awx/main/credential_plugins/aim.py
infamousjoeg: I have tested this successfully and committed the change.

obaranov: Tested and verified as fixed.

obaranov:
Tested with the empty and missing object_property key. Eveything looks good.

awx/main/credential_plugins/aim.py
```diff
@@ -98,7 +105,16 @@ def aim_backend(**kwargs):
...
+    # CCP returns the property name capitalized, username is camel case
+    # so we need to handle that case
+    if object_property == '':
+        object_property = 'Content'
+    elif object_property.lower() == 'username':
+        object_property = 'UserName'
+    else:
+        object_property = object_property.capitalize()
+
+    return res.json()[object_property]
```
fosterseth, line 119: should we check that `object_property` is a key in the response json object, and maybe raise a 400 error if it isn't. otherwise we'll get a 500 server error if a user supplies a bogus `object_property` value

infamousjoeg: Is this necessary if on L83 the key has a defined default of `''`?

`object_property = kwargs.get('object_property', '')`

In the conditional block you highlighted, the logic goes:

1. If `object_property` is `''` then set it to `Content`.
2. If the `object_property` is `username` then set it to `UserName`.
3. Otherwise, we can assume a value is given other than nothing or `username`, so we just need to capitalize the first letter of the value.

I don't believe, since there's a default value given on L83, that there will ever be an instance where `object_property` is not a key once we evaluate the highlighted conditional you are referring to.

Please correct me if my thinking is wrong here.  Thanks for the feedback!

fosterseth: I could be wrong, so I apologize in advance if I misunderstand something

```python3
object_property = kwargs.get('object_property', '')
```

this handles just the _input_ side. that is, if the input from user _doesn't_ provide `object_property`

but what if a user puts `bogus_value` on the object property prompt in the UI?

seems this like will hit a syntax error

`res.json()[object_property]`

because the response from cyberark isn't going to have a "bogus_value" key

I don't have a cyberark server to try this on, but curious what happens if you provide a nonsense value in the UI for object_property

infamousjoeg: Ahhhhh, excellent point!  I see what you mean now.  You were referring to the return of the `object_property` in the response rather than the conditional logic.

Let me test what the outcome of this occurring would be and we can continue the discussion here.

infamousjoeg: The error that occurs when a key not contained in the JSON response is used as the value for `object_property` is:
image

@fosterseth Any recommendations for handling this if it is not an acceptable exception?

fosterseth: @infamousjoeg thanks for the screenshots that is helpful

the "test" button will call out to the /test/ endpoint, and that specific code is wrapped in a `try except` block, so the backend is still probably hitting a 500 syntax error type of problem but it doesn't show up there.

instead of using the Test, what happens if you try it out in a job run? that is, just save the form, then create a machine credential with this lookup and provide the `thisdoesnotexist` as the object property and save it. Try to run a job and see what happens (and note if any server errors pop up)

infamousjoeg: It's been a rough few weeks. I'm continuing this today...

infamousjoeg: Here is the error message returned:

```text
  File "/awx_devel/awx/main/tasks/jobs.py", line 506, in run
    passwords = self.build_passwords(self.instance, kwargs)
  File "/awx_devel/awx/main/tasks/jobs.py", line 838, in build_passwords
    value = runtime_passwords.get(field, cred.get_input('password' if field == 'ssh_password' else field, default=''))
  File "/awx_devel/awx/main/models/credential/__init__.py", line 275, in get_input
    return self._get_dynamic_input(field_name)
  File "/awx_devel/awx/main/models/credential/__init__.py", line 309, in _get_dynamic_input
    return input_source.get_input_value()
  File "/awx_devel/awx/main/models/credential/__init__.py", line 1254, in get_input_value
    return backend(**backend_kwargs)
  File "/awx_devel/awx/main/credential_plugins/aim.py", line 117, in aim_backend
    return res.json()[object_property]
KeyError: 'Alsononexistent'
```

fosterseth: I think we can either raise a 400 Bad Request if 'object_property' is not in `res.json()`, or we can look for key 'Content' (as we do in empty case)

infamousjoeg: `ValueError` is raised when an Object Property is provided that is not returned in the JSON response.
image

infamousjoeg: Maybe a KeyError would be preferred?  Could not figure out how to raise a 400 Bad Request without introducing additional dependencies.

fosterseth: yeah let's do a `KeyError` instead. Don't worry about raising a Bad Request for this

infamousjoeg: Committed the change to KeyError to PR.

fosterseth (approved):

nielskroon:
NB issue #14606 created for below

Using AWX 23.3.0 and trying to fetch the 'Address' via Central Credential Provider version 12.6 I encounter a key error.
Content and UserName works flawlessly BTW and using PowerShell or on another instance (same versions) we can reproduce this error (and Address field in the JSON response is there and has the proper value).

Please reopen this issue.

How to reproduce with a Credential Type "VMware vCenter" aiming at a self managed safe:

```
{
  "reason": "no particular",
  "object_query": "Safe=XXXXXX_Self;Object=awx-vcenter-test",
  "object_property": "Address",
  "object_query_format": "Exact"
}
```
Test button in credential edit page will show a Key Error, and playbook in debug mode using this credential in debug mode fails with this message:
```
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/jobs.py", line 528, in run
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/models/credential/__init__.py", line 455, in inject_credential
    getattr(builtin_injectors, credential.credential_type.namespace)(credential, injected_env, private_data_dir)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/models/credential/injectors.py", line 71, in vmware
    env['VMWARE_HOST'] = cred.get_input('host', default='')
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/models/credential/__init__.py", line 279, in get_input
    return self._get_dynamic_input(field_name)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/models/credential/__init__.py", line 313, in _get_dynamic_input
    return input_source.get_input_value()
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/models/credential/__init__.py", line 1258, in get_input_value
    return backend(**backend_kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/credential_plugins/aim.py", line 115, in aim_backend
    raise KeyError('Property {} not found in object'.format(object_property))
KeyError: 'Property Address not found in object'
```

PR #13418, opened 2023-01-09
8049-expose execution node var for playbook

lucas-benedito:
##### SUMMARY
**Feature Overview**
Expose the Execution Node name as a variable for playbooks(awx_execution_node).

**Background, and strategic fit**
Customers running jobs in Ansible Tower use the ansible_hostname as an identifier for external applications requests (i.e.: Splunk, Service Now, etc ). In the Ansible Automation platform, that value is the container hostname which is now known to the customer and is not useful.

**(Optional) Use Cases**
- delegate the job to the specific cluster node
- define an identifier for external services such as Splunk, Service Now, etc.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - CLI

##### AWX VERSION
```
2.3
```

### DIFF
--- a/awx/main/models/jobs.py
+++ b/awx/main/models/jobs.py
@@ -832,7 +832,8 @@ class Job(UnifiedJob, JobOptions, SurveyJobMixin, JobNotificationMixin, TaskMana
                 r['{}_job_template_id'.format(name)] = self.job_template.pk
                 r['{}_job_template_name'.format(name)] = self.job_template.name
         if self.execution_node:
-            r['{}_execution_node'.format(name)] = self.execution_node
+            for name in JOB_VARIABLE_PREFIXES:
+                r['{}_execution_node'.format(name)] = self.execution_node
         return r
 
     '''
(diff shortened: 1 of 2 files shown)


TheRealHaoLiu:
just out of curiosity what is 8049 in reference to?

TheRealHaoLiu:
is this a reference to a JIRA issue? if so are you able to see the Automation Controller Project? please edit Jira issue field in the project

lucas-benedito:
@TheRealHaoLiu yes, that is the Jira issue, i will update the field accordingly. Thanks for the feedback

ffirg:
This one might be better served once constructed inventory has landed and we can then make better use of hostvars etc

fosterseth:
we may wish to update our functional test to assert the execution node is passed in

https://github.com/ansible/awx/blob/23a34c5dc9c0c1c48e88015dff9e9440cb55a802/awx/main/tests/unit/models/test_unified_job_unit.py#L96

Can you update the PR for that?

fosterseth (approved):

fosterseth:
comfirmed that execution node is injected into the extra vars, here is the extravars file in the private data dir
```
bash-5.1$ tail -n 1 extravars 
!unsafe 'tower_execution_node': !unsafe 'awx_1'
```

PR #13423, opened 2023-01-10
Allow web and task container to be deployed in separate deployment on Kubernetes

relrod:
##### SUMMARY

WIP - this will be filled in with a way better summary before it's ready for review.

Conglomerate branch/PR - includes:
- wsrelay work from `wsrelay` branch
- rsyslog work frorm #13222 and #13327

**merge commits will be rebased out before this gets tossed at `devel`**

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME

 - API


tools/ansible/roles/dockerfile/templates/supervisor.conf.j2
```diff
@@ -22,12 +22,11 @@ stderr_logfile=/dev/stderr
 stderr_logfile_maxbytes=0
 
 [program:uwsgi]
-
 {% if kube_dev | bool %}
 command = make uwsgi
 directory = /awx_devel
 environment =
-    DEV_RELOAD_COMMAND='supervisorctl -c /etc/supervisord_task.conf restart all; supervisorctl restart tower-processes:daphne tower-processes:wsbroadcast'
+    DEV_RELOAD_COMMAND='supervisorctl -c /etc/supervisord_task.conf restart all; supervisorctl restart tower-processes:daphne'
```
TheRealHaoLiu, line 29: this file is the supervisorconf file for web container it should not try to read the supervisor conf for the task container to restart daphne?

tools/ansible/roles/dockerfile/templates/supervisor_rsyslog.conf.j2
```diff
@@ -0,0 +1,59 @@
+; needs to launch rsyslog
+; needs to launch rsyslog-configurer 
+
```
TheRealHaoLiu: @djyasin please remove these comments

tools/ansible/roles/dockerfile/templates/supervisor_task.conf.j2
```diff
@@ -40,7 +57,7 @@ stderr_logfile=/dev/stderr
 stderr_logfile_maxbytes=0
 
 [group:tower-processes]
-programs=dispatcher,callback-receiver
+programs=dispatcher,callback-receiver,wsrelay
```
TheRealHaoLiu, line 60: awx-cache-clear is not needed in task container bc dispatcher also listen to `tower_settings_change`  pg_notify and will clear the setting in redis without needing `awx-cache-clear`

tools/docker-compose/ansible/roles/sources/templates/haproxy.cfg.j2
```diff
@@ -32,15 +32,15 @@ backend nodes
     option httpchk HEAD / HTTP/1.1\r\nHost:localhost
 {% for i in range(control_plane_node_count|int) %}
     {% set container_postfix = loop.index %}
-    server tools_awx_{{ container_postfix }} tools_awx_{{ container_postfix }}:8013 check
+    server tools_awx_{{ container_postfix }} tools_awx_{{ container_postfix }}:8013 check inter 10s
```
TheRealHaoLiu, line 35: can someone help me understand why this change was made?

tools/docker-compose/supervisor.conf
```diff
@@ -8,24 +8,47 @@ command = make dispatcher
 autorestart = true
 stopasgroup=true
 killasgroup=true
-stdout_logfile=/dev/stdout
-stdout_logfile_maxbytes=0
-stderr_logfile=/dev/stderr
-stderr_logfile_maxbytes=0
+stdout_events_enabled = true
```
TheRealHaoLiu, line 11: why was this changed?

TheRealHaoLiu: there seem to be a mix of 

```
stdout_logfile=/dev/stdout
stdout_logfile_maxbytes=0
stderr_logfile=/dev/stderr
stderr_logfile_maxbytes=0
```

and 

```
stdout_events_enabled = true
stderr_events_enabled = true
```

tools/ansible/roles/dockerfile/templates/supervisor.conf.j2
TheRealHaoLiu, line 1: we should rename this to supervisor_web.conf.j2

TheRealHaoLiu: make sure u also update Dockerfile.j2 to add the file if we rename it

awx/main/management/commands/run_dispatcher.py
```diff
@@ -76,7 +76,7 @@ def handle(self, *arg, **options):
         consumer = None
 
         try:
-            queues = ['tower_broadcast_all', get_local_queuename()]
+            queues = ['tower_broadcast_all', 'tower_settings_change', 'rsyslog_configurer', get_task_queuename()]
```
TheRealHaoLiu: why is dispatcher listening to `rsyslog_configurer`

TheRealHaoLiu: to make this work the same i think we should remove rsyslog_configurer from dispatcher and in docker-compose devleopment environment's superviosr conf start the `awx-manage run_rsyslog_configurer.py`

tools/docker-compose/supervisor.conf
thedoubl3j: this was changed because we are not triggering capture mode with `stdout_capture_maxbytes` as we have it set to 0. Therefore, the change to `stdout_events_enabled` was made as we have no reason to care about the size of the logfile. I would favor using `stdout_events; stderr_events` merely for less lines but they are doing the same thing. We should standardize on one

tools/docker-compose/ansible/roles/sources/templates/haproxy.cfg.j2
thedoubl3j: per a convo @relrod we understand what it does, changing how often we check the health of the backend servers. Why this change was made, we may need to poke @shanemcd on that as he made this change

tools/ansible/roles/dockerfile/templates/supervisor.conf.j2
relrod: > it should not try to read the supervisor conf for the task container to restart daphne?

It is not. There are two separate commands in that line separated by a `;`

awx/main/management/commands/run_dispatcher.py
thedoubl3j: remove cache_clear and add awx-rsyslog-configurer to group

tools/ansible/roles/dockerfile/templates/supervisor_rsyslog.conf.j2
djyasin: Resolved by: https://github.com/ansible/awx/pull/13763

awx/main/utils/external_logging.py
```diff
@@ -124,3 +127,8 @@ def reconfigure_rsyslog():
             f.write(tmpl + '\n')
         shutil.move(path, '/var/lib/awx/rsyslog/rsyslog.conf')
     supervisor_service_command(command='restart', service='awx-rsyslogd')
+
+
+def send_pg_notify(channel: str, payload: str) -> None:
```
fosterseth: is this code needed anymore? I don't see this referenced anywhere

docs/websockets.md
```diff
@@ -10,13 +10,13 @@ To communicate between our different services we use websockets. Every AWX node
...
 
-### Broadcast Backplane
+### Relay Backplane
 
-Previously, AWX leveraged RabbitMQ to deliver Ansible events that emanated from one AWX node to all other AWX nodes so that any client listening and subscribed to the Websockets could get events from any running playbook. We are since moved off of RabbitMQ and onto a per-node local Redis instance. To maintain the requirement that any Websocket connection can receive events from any playbook running on any AWX node we still need to deliver every event to every AWX node. AWX does this via a fully connected Websocket backplane. 
+Previously, AWX leveraged RabbitMQ to deliver Ansible events that emanated from one AWX node to all other AWX nodes so that any client listening and subscribed to the Websockets could get events from any running playbook. We are since moved off of RabbitMQ and onto a per-node local Redis instance. To maintain the requirement that any Websocket connection can receive events from any playbook running on any AWX node we still need to deliver every event to every AWX node. AWX does this via a fully connected Websocket backplane.
```
fosterseth: As part of this feature, we no longer send every event to every AWX node.

We can also drop the comments regarding rabbitMQ

I would replace this paragraph with something like,

"A core feature of AWX is that any websocket connection can receive events from any running playbook, regardless of the control node that is handling those events. Each control node runs a service called `wsrelay`. This service maintains an active connection to each running web service, forming a websocket backplane. The web services are powered by Redis-based Django channels to handle websocket messages. As jobs run and events are processed, the control node sends events _only_ to web services that subscribe for those job events."

docs/websockets.md
```diff
@@ -65,14 +65,14 @@ This section will specifically discuss deployment in the context of websockets a
...
-| `wsbroadcast`   | no listening port, forwards all group messages to all cluster nodes |
+| `wsrelay`   | no listening port, forwards all group messages to all cluster nodes |
 | `supervisord` | (production-only) handles the process management of all the services except `nginx` |
 
 When a request comes in to `nginx` and has the `Upgrade` header and is for the path `/websocket`, then `nginx` knows that it should be routing that request to our `daphne` service.
 
 `daphne` handles websocket connections proxied by nginx.
 
-`wsbroadcast` fully connects all cluster nodes via the `/websocket/broadcast/` endpoint to every other cluster nodes. Sends a copy of all group websocket messages to all other cluster nodes (i.e. job event type messages).
+`wsrelay` fully connects all cluster nodes via the `/websocket/broadcast/` endpoint to every other cluster nodes. Sends a copy of all group websocket messages to all other cluster nodes (i.e. job event type messages).
```
fosterseth: wsrelay only runs on control nodes, and only connects to web services (pods). They send websocket messages to _only_ those web services that subscribe to those message types.

relrod: Yeah, these docs need to be updated, they don't reflect the new architecture at all.

ffirg: It would be great if you could add an architectural diagram for this. That would make it 100% easier to understand.

tools/ansible/roles/dockerfile/templates/supervisor.conf.j2
```diff
@@ -75,8 +74,8 @@ stdout_logfile_maxbytes=0
 stderr_logfile=/dev/stderr
 stderr_logfile_maxbytes=0
 
-[program:awx-rsyslogd]
-command = rsyslogd -n -i /var/run/awx-rsyslog/rsyslog.pid -f /var/lib/awx/rsyslog/rsyslog.conf
+[program:awx-cache-clear]
+command = awx-manage run_cache_clear
```
TheRealHaoLiu: if kube_dev conditional need to exist look at how conditional around uwsgi is done

awx/main/management/commands/run_dispatcher.py
thedoubl3j: fixed in #13768

AlanCoding: We talked it over and I agree that removing `tower_settings_change` is what we need for right now - because it is consistent with all the 5 supervisor config files out there.

AlanCoding: https://github.com/ansible/awx/pull/13768

Makefile
```diff
@@ -5,6 +5,7 @@ DOCKER_COMPOSE ?= docker-compose
 OFFICIAL ?= no
 NODE ?= node
 NPM_BIN ?= npm
+KIND_BIN ?= $(shell which kind)
```
AlanCoding: This adds overhead to running any make target, because it has to run this shell command. I also cannot understand the utility of doing this much more complicated thing instead of just typing `kind` in the `make kind-dev-load` where it is intended to be used.

EDIT: to clarify, when `kind` is not available (think of running things inside of the container) then this still has a value, which is

```
which: no kind in (/home/alancoding/repos/awx/env/bin:/home/alancoding/Downloads/google-cloud-sdk-367.0.0-linux-x86_64/google-cloud-sdk/bin:/home/alancoding/.local/bin:/home/alancoding/bin:/home/alancoding/.cargo/bin:/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/var/lib/snapd/snap/bin)
```

I really don't like having that just hanging around in totally irrelevant contexts.

awx/main/dispatch/__init__.py
```diff
@@ -14,6 +15,26 @@ def get_local_queuename():
...
+
+        return (
+            Instance.objects.filter(
+                node_type__in=[Instance.Types.CONTROL, Instance.Types.HYBRID],
+                node_state=Instance.States.READY,
+                enabled=True,
+            )
+            .only('hostname')
+            .order_by('?')
+            .first()
```
AlanCoding: I would like to formally mention this - I want to see error handling for the case that `.first()` returns `None`, meaning the database is not prepped with instances set up.

This could happen on a new install if services are started but instances are not yet registered. This may not be a "bad" thing in all cases, just want to see it handled.

awx/main/utils/external_logging.py
TheRealHaoLiu: @jessicamack can u address this comment

docs/websockets.md
TheRealHaoLiu: @relrod can u update the doc to reflect the current design

docs/websockets.md
TheRealHaoLiu: @relrod can u address this comment

tools/ansible/roles/dockerfile/templates/supervisor.conf.j2
TheRealHaoLiu: @djyasin can u address this comment

awx/main/dispatch/__init__.py
TheRealHaoLiu: ok i will address this

Makefile
TheRealHaoLiu: this change does not belong in this PR @djyasin can u revert this change

Makefile
```diff
@@ -600,6 +619,9 @@ awx-kube-build: Dockerfile
 		--build-arg HEADLESS=$(HEADLESS) \
 		-t $(DEV_DOCKER_TAG_BASE)/awx:$(COMPOSE_TAG) .
 
+kind-dev-load: awx-kube-dev-build
```
TheRealHaoLiu: this change does not belong in this PR @djyasin can u revert this change

djyasin: Resolved with PR https://github.com/ansible/awx/pull/13770

awx/main/management/commands/run_heartbeet.py
```diff
@@ -0,0 +1,67 @@
...
+    def construct_payload(self, action='online'):
+        payload = {
+            'hostname': settings.CLUSTER_HOST_ID,
+            'ip': os.environ.get('MY_POD_IP'),
+            'action': action,
+        }
+        return json.dumps(payload)
+
+    def do_hearbeat_loop(self):
+        with pg_bus_conn(new_connection=True) as conn:
```
AlanCoding, line 54: After looking at this more, I see that this is a publisher, not a listener (logical). As such, I think it would be _better_ if it didn't create a new connection.

AlanCoding (approved):

awx/main/utils/external_logging.py
jessicamack: it's not. making a PR to remove it

tools/ansible/roles/dockerfile/templates/supervisor.conf.j2
djyasin: Resolved via this pr: https://github.com/ansible/awx/pull/13771

awx/main/dispatch/__init__.py
TheRealHaoLiu: fixed in https://github.com/ansible/awx/pull/13772

CFSNM (approved):

Issue #13384, opened 2022-12-29
Thycotic Secret Server SSH key + passphrase does not work


timothydilbert:
##### ISSUE TYPE

 - Bug, Docs Fix or other nominal change

##### COMPONENT NAME
 - API
 - UI
 - CLI
 - Other

##### SUMMARY
We use Thycotic Secret Server as an external secret management platform. SSH Keys stored in Secret Server are rotated every X days. When we try to use those SSH keys with Job Templates, it causes errors in the Ansible Automation Platform and the playbook does not run.

##### ENVIRONMENT
* AWX version: Ansible Automation Platform Controller 4.2.0
* AWX install method: operator, developer environment
* AWX deployment target: N/A
* Operating System: RHEL 8
* Web Browser: N/A

##### STEPS TO REPRODUCE

1. On an Ubuntu server, create a sudo user with an SSH key (no password).
2. Create a secret in the Thycotic Secret server using the **Unix Account (SSH Key Rotation - No Password)** template.
3. Wait for the heartbeat to report as being successful, confirming that the original SSH key provided is valid.
4. Use Secret Server to rotate the SSH key and generate a new one.
5. Go to Ansible Automation Platform, add a new credential, mapping the **SSH Private Key** (private-key) and **Private Key Passphrase** (private-key-passphrase) to the Thycotic Secret and Field names.
6. Try to run a Job Template using those credentials.

##### EXPECTED RESULTS

Ansible should connect to the Host specified on the Job Template. 

##### ACTUAL RESULTS

A python error is thrown with the following output:
```
Traceback (most recent call last):
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/jobs.py", line 539, in run
    res = receptor_job.run()
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/receptor.py", line 271, in run
    res = self._run_internal(receptor_ctl)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/receptor.py", line 326, in _run_internal
    transmitter_future.result()
  File "/usr/lib64/python3.9/concurrent/futures/_base.py", line 439, in result
    return self.__get_result()
  File "/usr/lib64/python3.9/concurrent/futures/_base.py", line 391, in __get_result
    raise self._exception
  File "/usr/lib64/python3.9/concurrent/futures/thread.py", line 58, in run
[...]

PR #13425, opened 2023-01-11
Fix for Issue https://github.com/ansible/awx/issues/13384

tarunchy:
# For SSH Key types thycotic is returning a Object of Type <class 'requests.models.Response'>
    # So we need to check whether response is a string or object and based on that we need return value

##### ISSUE TYPE

 - Bug

##### COMPONENT NAME

 - Other

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```


tarunchy:
Closing this PR to follow right process of merging. Will raise another PR

PR #13428, opened 2023-01-13
[bulk job launch feature]add organization validation to the workflowjob

jainnikhil30:
Set the Organization on the Workflow job with following rule: 
- If the organization is not set in the request, set it to the request user's organization, if user is part of only 1 organization
- if the organization is not set in the request, and request user is part of multiple organization, throw an error saying "User is part of multiple organization, please provide one in the request"
- if the organization is set in the request, validate if the user has access to it or not.

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -4509,7 +4509,13 @@ class BulkJobLaunchSerializer(serializers.Serializer):
     jobs = BulkJobNodeSerializer(many=True, allow_empty=False, write_only=True, max_length=1000)
     description: serializers.CharField(write_only=True, required=False, allow_blank=False)
     extra_vars: serializers.CharField(write_only=True, required=False, allow_blank=False)
-    # organization: (Maybe if they don't set it, set their first organization) # Here we can use PrimaryKeyRelatedField so it will automagically do rbac/turn into object )
+    organization = serializers.PrimaryKeyRelatedField(
+        queryset=Organization.objects.all(),
+        required=False,
+        default=None,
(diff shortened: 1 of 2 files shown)


awx/api/serializers.py
```diff
@@ -4525,6 +4532,20 @@ class Meta:
...
+        # validate Organization
+        # - If the orgs is not set, set it to the org of the launching user
+        # - If the user is part of multiple orgs, throw a validation error saying user is part of multiple orgs, please provide one
+
+        if 'organization' not in attrs or attrs['organization'] == None or attrs['oganization'] == '':
+            if Organization.accessible_pk_qs(request.user, 'read_role').count() == 1:
+                for tup in Organization.accessible_pk_qs(request.user, 'read_role').all():
+                    attrs['organization'] = tup[0]
+            elif Organization.accessible_pk_qs(request.user, 'read_role').count() > 1:
+                raise serializers.ValidationError(_(f"User is part of multiple Organization, please set one of them in the request"))
```
```suggestion
                raise serializers.ValidationError(_(f"User has permission to multiple Organizations, please set one of them in the request"))
```

awx/api/serializers.py
```diff
@@ -4525,6 +4532,20 @@ class Meta:
...
+            else:
+                raise serializers.ValidationError(_(f"User not part of any organization, please assign an organization for the user"))
```
```suggestion
                raise serializers.ValidationError(_(f"User not part of any organization, please assign an organization to assign to the bulk job"))
```

kdelee (approved):
great job, will merge

Issue #13429, opened 2023-01-13
admin_role disassociated despite remove_admins = false


anxstj:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

After upgrading from AWX 21.5.1 to 21.10.2 all organization admin roles were removed from the users' accounts.

We're using LDAP with an organization map looking like this:
```
{
  "Org1": {
    "admins": false,
    "remove_admins": false,
    "remove_users": true,
    "users": [
      "CN=..."
    ]
  }
}
```
Users were manually granted the organization admin role. That had worked before, but after the upgrade, the role was removed from the users with the  following activity message:
```
| 13.1.2023, 14:10:40 | system | disassociated Org1 admin_role from jdoe
```
And AWX was logging the following:
```
DEBUG    [4ecb27399afb4cb7b9ad2fd5e666a8ca] awx.sso.backends LDAP adapter removing user jdoe permission of admin_role from organization Org1
```

I tried to remove '"admins": false' from the config or replace it with an empty list ('admins": [],') but that didn't change anything. "None" was not accepted by the form.
According to the Automation Platform docs that should work: https://docs.ansible.com/automation-controller/4.2.1/html/administration/social_auth.html#organization-mapping

This seems to be related to the following PR: https://github.com/ansible/awx/pull/12949

### AWX version

21.10.2

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

docker development environment

### Modifications

no

### Ansible version

2.12.2

### Operating system

Debian 11

### Web browser

Firefox

### Steps to reproduce

1) Add default LDAP settings
2) Create an organization map with "admins: false"
3) Add a user manually as organization admin to an organization
4) Login with that user and recognize that the admin_role was removed

### Expected results

The system doesn't remove the admin_role from the user.

### Actual results

The admin_role is removed from the user

PR #13430, opened 2023-01-13
Fix bug in LDAP adapter where the remove flag was not being respected for org/team membership when False

anxstj:
##### SUMMARY

_update_m2m_from_groups must return None if remove_* is false or empty, because None indicates that the user permissions will not be changed.

related #13429 

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.10.3.dev32+g6ed7b5dd21
```

##### ADDITIONAL INFORMATION

The  reconcile_users_org_team_mappings() function in backends.py has three exit options:
1) true: the user will be added to the org/team
2) false: the user will be removed from the org/team
3) None: break - do nothing

If an organization is configured with remove_* (e.g. remove_admin) = false then the users with the admin_role must not be removed from the org. That means the reconcile_users_org_team_mappings() must break and that happens if _update_m2m_from_groups() returns with None.

This fix ensures that it will do that.

```

```

### DIFF
--- a/awx/sso/backends.py
+++ b/awx/sso/backends.py
@@ -354,7 +354,9 @@ def _update_m2m_from_groups(ldap_user, opts, remove=True):
                 continue
             if ldap_user._get_groups().is_member_of(group_dn):
                 return True
-    return False
+    if remove:
+        return False
+    return None
 
 
 @receiver(populate_user, dispatch_uid='populate-ldap-user')


awx/sso/backends.py
```diff
@@ -342,7 +342,7 @@ def _update_m2m_from_groups(ldap_user, opts, remove=True):
     if opts is None:
         return None
     elif not opts:
-        pass
+        return None
```
john-westcott-iv: Thank you so much for not only pointing this out but also working on fixing it!

In this function `opts` is going to be the `admins` section of the configuration when we are looking at the administrator role. i.e.:
```json
   "admins": [ "cn=group1,ou=groups,dc=example,dc=org", "cn=group2ou=groups,dc=example,dc=org"],
```

Imagine a scenario where someone would want no admins from LDAP and to remove any LDAP users from the admin roles so their setting is:
```json
    "admins": [],
    "remove_admins": true,
```

Now lets say an LDAP user was made an org admin and was logging in, in this scenario they should not be an admin because they are not in `[]` and `remove_admins` is `true`.

The empty list `[]` would trigger this line and `return None` meaning no change to the users permissions. This behavior is incorrect in this scenario.

Instead I think we want to keep this as "pass" to short circuit the remaining if clauses, but then at the bottom of the function instead of just `return False` do something like:
```python
    if remove:
        return False
    return None
```

So if we had no `opts` configuration we would skip the checks that could `return True` and next check if we are supposed to remove users from this role and, if so, `return False` to do that. If we weren't mandated to remove the user and they were not explicitly added than we `return None` so that their permissions for the role do not change.

Please let me know if you agree with that. If so, we can update your PR and keep it moving to get it merged.

Thanks again!

john-westcott-iv (changes requested):

awx/sso/backends.py
anxstj: Yes, you're right. We need to take the remove status into account, too.

I come up with the following combinations:

| opts     | remove | result (user in opts) | result (user not in opts) |
| ---| --- | --- | ---|
| True     | True   | True                  | ---   |
| True     | False  | True                  | ---   |
| [...]    | True   | True                  | False |
| [...]    | False  | True                  | None  |
| []/False | True   | ---                   | False |
| []/False | False  | ---                   | None  |
| None     | True   | None                  | None  |
| None     | False  | None                  | None  |

I followed your suggestion and updated my code accordingly.

awx/sso/backends.py
```diff
@@ -342,7 +342,7 @@ def _update_m2m_from_groups(ldap_user, opts, remove=True):
     if opts is None:
         return None
     elif not opts:
-        pass
+        return pass
```
john-westcott-iv: Just one more minor fix, we need to just pass here instead of calling return. Other than that this is good to go!

anxstj: Of course.

john-westcott-iv (approved):

jay-steurer (approved):

Issue #13437, opened 2023-01-17
[web/task split] Rework django caching mechanism so that task and web services do not rely upon shared redis instance


jessicamack:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

New Feature

### Feature Summary

clear_setting_cache is a background task that runs whenever settings are changed via the API/UI. This runs in the context of the task container. It will delete cache keys from the redis cache if they have changed. This relies on the same instance of redis being shared between the task system and web service. This will not be the case after the split. A new service needs to be added to the web pod, almost identical to run_rsyslog_configurer in PR 13222 that listens to a pg_notify channel (tower_settings_change) and will clear the settings keys in redis if it gets one of those messages using the same method that we use now.

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

After the split the task and web containers will not share a redis instance. We store settings in the cache so the lack of a shared instance could mean the settings are temporarily different between containers. This will prevent that. 

### Current results

There is not a current issue but will be in the future.

### Sugested feature result

The API should continue to work as it does now.

PR #13438, opened 2023-01-17
[web-task-split] Rework django caching mechanism

jessicamack:
Signed-off-by: Jessica Mack <jmack@redhat.com>

##### SUMMARY
related #13437 

Added a  new service that listens to a pg_notify channel (tower_settings_change) and will clear the settings keys in redis.
Also updated the dispatcher to listen to the queues added recently.

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 
##### AWX VERSION

```
21.7.1
```


awx/main/tasks/system.py
```diff
@@ -240,15 +240,10 @@ def apply_cluster_membership_policies():
...
 def clear_setting_cache(setting_keys):
-    orig_len = len(setting_keys)
-    for i in range(orig_len):
-        for dependent_key in settings_registry.get_dependent_settings(setting_keys[i]):
-            setting_keys.append(dependent_key)
-    cache_keys = set(setting_keys)
-    logger.debug('cache delete_many(%r)', cache_keys)
-    cache.delete_many(cache_keys)
+    # notify the service to clear the cache
+    send_pg_notify('tower_settings_change', setting_keys)
```
AlanCoding: It seems like 2 types of messages are going to be going into the "tower_settings_change" channel - one with the task data for `clear_setting_cache` and one with just a list of setting_keys. I'd prefer to only deal with the task. You could have `run_clear_cache.py` process the task data and then call this method directly (similar to the dispatcher but without using a thread pool).

awx/main/management/commands/run_dispatcher.py
```diff
@@ -76,7 +76,7 @@ def handle(self, *arg, **options):
         consumer = None
 
         try:
-            queues = ['tower_broadcast_all', get_local_queuename()]
+            queues = ['tower_broadcast_all', 'tower_settings_change', 'rsyslog_configurer', get_local_queuename()]
```
relrod, line 79: I'm confused about this change (and I don't know dispatcher enough yet - I might just be way off base here).

AFAIK we're not using the dispatcher for either of these daemons (the cache buster nor the rsyslog configurererer). The daemons themselves that are listening for these messages are what is doing the listening for these messages (as opposed to dispatcher), right?

awx/main/tasks/system.py
```diff
@@ -240,15 +240,10 @@ def apply_cluster_membership_policies():
         logger.debug('Cluster policy computation finished in {} seconds'.format(time.time() - started_compute))
 
 
-@task(queue='tower_broadcast_all')
+@task(queue='tower_settings_change')
 def clear_setting_cache(setting_keys):
-    orig_len = len(setting_keys)
```
TheRealHaoLiu, line 245: i think this still need to be here since redis cache still need to be cleared in the task pods

tools/ansible/roles/dockerfile/files/launch_awx_cache_clear.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
TheRealHaoLiu: this file is not needed since awx_cache_clear is just a process within the web container and not a container 

launch scripts are used when launching a container

awx/main/tasks/system.py
TheRealHaoLiu: alternatively we just launch cache-clear in both web and task therefore this method would only be there to serve to send pg_notify

TheRealHaoLiu: additional random qustion... if i have 3 task pod would this send out 3 pg_notify or 1?

TheRealHaoLiu: can we just send the pg_notify during when we write setting to the database? this way we completely eliminate the need of even queuing this job

awx/main/management/commands/run_dispatcher.py
jessicamack: I'm not completely clear on all the functions of dispatcher as well but in the discussion on the JIRA ticket it was said that the new rsyslog channel should be added to the queues dispatcher is listening to. I assumed that meant I should add the new cache clear channel as well. Hao has clarified that this will be a service not another container though so I would assume it actually doesn't need to be added.

awx/main/tasks/system.py
jessicamack: I'm pretty sure redis will still be cleared by the handle_settings_change in the signals.py file https://github.com/ansible/awx/blob/devel/awx/conf/signals.py. 

I would assume with 3 tasks pods it would send pg_notify 3 times. Are the pods all receiving the same requests? Are we worried about duplication of work?

Do you mean removing the clear_settings_job and just calling pg_notify in on_commit?

AlanCoding: > if i have 3 task pod would this send out 3 pg_notify or 1?

@TheRealHaoLiu pg_notify does a NOTIFY to a "channel" in postgres, and every client that is LISTENing will receive the message. I am capitalizing to conform with postgres SQL syntax.

IMHO, the _best_ implementation would NOTIFY 1 channel when a setting is changed. You _could_ have a `tower_settings_change_web` channel and `tower_settings_change_task` channel, but I have never discussed that.

Assuming a single-channel implementation (because it is my preference)... if there are 3 task pods, then there will be 3 dispatcher processes, one for each pod, listening to the `tower_settings_change` channel. The NOTIFY happens once, but _postgres_ sends the message to _all listeners_, which would be 3 in this case.

> Do you mean removing the clear_settings_job and just calling pg_notify in on_commit?

To clarify what I would prefer, which is the same direction I was trying to give in my last comment - revert to put back in the prior code inside of `clear_setting_cache`, but keep the change where you changed its queue to `tower_settings_change`. Keep the channels you have defined for the dispatcher and other service to listen to.

Now, do some testing, and log / observe what messages the dispatcher service and the `run_clear_cache` service are getting. You should observe that they are _both_ getting the message to run this task. The only problem is that `run_clear_cache` isn't equipped to interpret this message. What I tried to say in prior comments is that we should look at that data that the current system (the "publisher" part of the dispatcher) sends in the message. Then `run_clear_cache` should parse the data to get the `setting_keys` and then directly (meaning, not with `.delay` or `.apply_async`) call `clear_setting_cache` with that argument.

awx/main/management/commands/run_dispatcher.py
AlanCoding: The change here to add "tower_settings_change" is coherent, because it is a `@task`. However, adding the "rsyslog_configurer" would require corresponding changes that are not in this PR. Functionally, a PATCH to `/api/v2/settings/logging/` produces:

```
tools_awx_1 |   File "/awx_devel/manage.py", line 8, in <module>
tools_awx_1 |     manage()
tools_awx_1 |   File "/awx_devel/awx/__init__.py", line 201, in manage
tools_awx_1 |     execute_from_command_line(sys.argv)
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
tools_awx_1 |     utility.execute()
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/management/__init__.py", line 413, in execute
tools_awx_1 |     self.fetch_command(subcommand).run_from_argv(self.argv)
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/management/base.py", line 354, in run_from_argv
tools_awx_1 |     self.execute(*args, **cmd_options)
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/management/base.py", line 398, in execute
tools_awx_1 |     output = self.handle(*args, **options)
tools_awx_1 |   File "/awx_devel/awx/main/management/commands/run_dispatcher.py", line 81, in handle
tools_awx_1 |     consumer.run()
tools_awx_1 |   File "/awx_devel/awx/main/dispatch/worker/base.py", line 185, in run
tools_awx_1 |     self.process_task(json.loads(e.payload))
tools_awx_1 |   File "/usr/lib64/python3.9/json/__init__.py", line 346, in loads
tools_awx_1 |     return _default_decoder.decode(s)
tools_awx_1 |   File "/usr/lib64/python3.9/json/decoder.py", line 337, in decode
tools_awx_1 |     obj, end = self.raw_decode(s, idx=_w(s, 0).end())
tools_awx_1 |   File "/usr/lib64/python3.9/json/decoder.py", line 355, in raw_decode
tools_awx_1 |     raise JSONDecodeError("Expecting value", s, err.value) from None
tools_awx_1 | json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

This is because traffic in the "rsyslog_configurer" contains non-JSON data, because it is not using a `@task` implementation.

Right now, I would probably advocate removing "rsyslog_configurer" from the queues here and dealing with it in another PR. Either way, the rsyslog re-configuration needs an additional patch because right now the messages are not processed in VM installs.

awx/main/tasks/system.py
```diff
@@ -240,15 +240,10 @@ def apply_cluster_membership_policies():
...
-    for i in range(orig_len):
-        for dependent_key in settings_registry.get_dependent_settings(setting_keys[i]):
-            setting_keys.append(dependent_key)
-    cache_keys = set(setting_keys)
-    logger.debug('cache delete_many(%r)', cache_keys)
-    cache.delete_many(cache_keys)
+    # log that cache is being cleared
+    logger.info(f"clear_setting_cache of keys {setting_keys}")
```
AlanCoding, line 247: This needs to contain the original logic with `delete_many` because that's how the settings actually get cleared. Meaning, the method _contents_ need to be what they were before.

jessicamack: Thank you. Missed that I hadn't added it back.

tools/ansible/roles/dockerfile/templates/supervisor.conf.j2
```diff
@@ -74,8 +74,19 @@ stdout_logfile_maxbytes=0
...
+stopasgroup=true
+killasgroup=true
+stdout_logfile=/dev/stdout
+stdout_logfile_maxbytes=0
+stderr_logfile=/dev/stderr
+stderr_logfile_maxbytes=0
+
 [group:tower-processes]
-programs=nginx,uwsgi,daphne
+programs=nginx,uwsgi,daphne,wsbroadcast,awx-cache-clear
```
AlanCoding: `wsbroadcast` exists in the devel branch but not in this branch or the feature branch. I'm thinking this is an artifact of a git conflict.

On the other hand... this file does contain a `heartbeet` that isn't in devel, which should be listed here.

EDIT: I had something wrong, this is the right place for the `awx-clear-cache` service to go.

tools/docker-compose/supervisor.conf
```diff
@@ -91,7 +101,7 @@ stdout_events_enabled = true
 stderr_events_enabled = true
 
 [group:tower-processes]
-programs=awx-dispatcher,awx-receiver,awx-uwsgi,awx-daphne,awx-nginx,awx-wsrelay,awx-rsyslogd,awx-heartbeet
+programs=awx-dispatcher,awx-receiver,awx-uwsgi,awx-daphne,awx-nginx,awx-wsrelay,awx-rsyslogd,awx-heartbeet,awx-cache-clear
```
AlanCoding: I don't think `awx-clear-cache` necessarily needs to be added here, because this includes the dispatcher service which will clear the cache when needed. This was needed for testing, so that might be why it's here.

AlanCoding (approved):
I have some outstanding comments about internal consistency in the supervisor.conf files, but this either isn't production-relevant, or wasn't fully fixed before.

Given that this is changing the _web_ supervisor file, I think this becomes a complete functional set of changes. Maybe it needs to be tested on openshift, but I don't have any more comments to make.

tools/ansible/roles/dockerfile/templates/supervisor.conf.j2
jessicamack: wsbroadcast and heartbeet might be better added in the feature branch since they're missing there and not relevant to this PR.

tools/docker-compose/supervisor.conf
jessicamack: Removed. Related to the comment above https://github.com/ansible/awx/pull/13438#discussion_r1093770801 if heartbeet is here does it also need to be in supervisor.conf.j2?

tools/ansible/roles/dockerfile/templates/supervisor.conf.j2
relrod: Adding `wsbroadcast` here looks incorrect to me. In the feature branch (which this branch is based off and merging into), wsbroadcast is replaced with wsrelay, and that only runs in the task container. The web container shouldn't have to run any WS processes except heartbeet.

tools/docker-compose/supervisor.conf
relrod: Yes, heartbeet needs to run in the web container.

PR #11920, opened 2022-03-18
Improves UX of workflow approval list

AlexSCorey:
##### SUMMARY
The action options on the workflow approval list lacked clarity.  This work aims to improve that.  As part of this work I was able to remove some code/files that were no longer needed.  
##### ISSUE TYPE
-enhancement

##### COMPONENT NAME
 - UI

##### ADDITIONAL INFORMATION

https://tower-mockups.testing.ansible.com/patternfly/workflow-approvals/wf-approvals-detail/
https://tower-mockups.testing.ansible.com/patternfly/workflow-approvals/wf-approvals/

PR #13450, opened 2023-01-19
Re-add workflow approval bulk actions to workflow approvals list

mabashian:
##### SUMMARY
These buttons were moved to a kabob here: https://github.com/ansible/awx/pull/11920
Then they were moved into row actions here: https://github.com/ansible/awx/pull/12548

This PR rolls the bulk actions back to a state before #11920 while keeping the row actions in place.

The expectation is that a user will be able to select any number of pending workflow approvals and do a bulk approve or deny.  Workflow cancel will still only be supported at the row level.

Screen Shot 2023-01-19 at 12 23 34 PM

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui/src/screens/WorkflowApproval/WorkflowApprovalList/WorkflowApprovalList.js
+++ b/awx/ui/src/screens/WorkflowApproval/WorkflowApprovalList/WorkflowApprovalList.js
@@ -12,11 +12,16 @@ import PaginatedTable, {
 import AlertModal from 'components/AlertModal';
 import ErrorDetail from 'components/ErrorDetail';
 import DataListToolbar from 'components/DataListToolbar';
-import useRequest, { useDeleteItems } from 'hooks/useRequest';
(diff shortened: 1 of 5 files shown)


marshmalien (approved):
👍 Tested locally and it looks good to me.

vidyanambiar (approved):
lgtm!

vidyanambiar:
@mabashian  I came back and tried this on awx and saw the same thing - https://github.com/ansible/tower/pull/6277#pullrequestreview-1267843275 Let me know if it needs to be a separate issue.

vidyanambiar:
@mabashian The issue in the comment above also occurs on the `devel` branch so it is not related to this PR. I opened a separate issue for it. https://github.com/ansible/awx/issues/13465

akus062381 (approved):

PR #13455, opened 2023-01-20
Fixes #13439 Add exception handling for `/api` on url

### LINKED ISSUE #13439: CyberArk Conjur lookup plugin does not work with open source Conjur version

### Bug Summary

The CyberArk Conjur Secrets Manager Lookup plugin no longer supports the lookup against the Conjur OSS. Most likely because the opensource conjure version API endpoint does not have the `/api/` portion recently added by this change: https://github.com/ansible/awx/pull/13121

### AWX version

2.4

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

docker development environment

### Modifications

no

### Ansible version

_No response_

### Operating system

_No response_

### Web browser

_No response_

### Steps to reproduce

1. Deploy Conjur OSS. We use the latest `cyberark/conjur` image for that. The Conjur version is
```
    Version 1.19.1-3398
    API Version 5.3.0 
```
2. Under the AWX, create the CyberArk Conjur Secrets Manager Lookup credential. Provide Conjure URL, user, and token. 
3. Click Test and provide a path. 

### Expected results

Connection to Conjur should be established. HTTP 200 response. 

### Actual results

401 error returned. 

```
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/jobs.py", line 504, in run
    args = self.build_args(self.instance, private_data_dir, passwords)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/jobs.py", line 937, in build_args
    ssh_username = creds.get_input('username', default='')
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/models/credential/__init__.py", line 275, in get_input
    return self._get_dynamic_input(field_name)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/models/credential/__init__.py", line 309, in _get_dynamic_input
    return input_source.get_input_value()
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/models/credential/__init__.py", line 1250, in get_input_value
    return backend(**backend_kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/credential_plugins/conjur.py", line 72, in conjur_backend
    raise_for_status(resp)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/credential_plugins/plugin.py", line 12, in raise_for_status
    resp.raise_for_status()
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://conjureurl.com/api/authn/test/admin/authenticate
```

### Additional information

_No response_

infamousjoeg:
##### SUMMARY

Self-Hosted Conjur Secrets Manager Enterprise and Conjur Cloud both support the usage of `/api` after the Base URL of the Conjur service.  In the case of Conjur Cloud, it is a requirement to interact with Conjur's API.  However, Conjur Open Source does not support the usage of `/api` after the Base URL for the Conjur service.

The change implemented in this PR adds a `try...exception...except` block to every `requests` command used in the Credential Plugin to first try without `/api` and if a `requests.exceptions.HTTPConnection` exception is captured, try with `/api` provided.  If that then fails, an error will be raised properly based on the response from the Conjur service API.

Related #13439 

##### ISSUE TYPE

##### COMPONENT NAME

 - UI

##### AWX VERSION

```
awx: 0.1.dev32760+g4470b80
```

##### ADDITIONAL INFORMATION

This was tested successfully using the Conjur Open Source Quick-Start Guide available at https://www.conjur.org/get-started/quick-start/oss-environment.

### DIFF
--- a/awx/main/credential_plugins/conjur.py
+++ b/awx/main/credential_plugins/conjur.py
@@ -72,8 +72,6 @@ def conjur_backend(**kwargs):
             resp = requests.post(urljoin(url, '/'.join(['authn', account, username, 'authenticate'])), **auth_kwargs)
         except requests.exceptions.ConnectionError:
             resp = requests.post(urljoin(url, '/'.join(['api', 'authn', account, username, 'authenticate'])), **auth_kwargs)
-        except:
-            resp = requests.post(urljoin(url, '/'.join(['api', 'authn', account, username, 'authenticate'])), **auth_kwargs)
     raise_for_status(resp)
     token = resp.content.decode('utf-8')
 
@@ -95,8 +93,6 @@ def conjur_backend(**kwargs):
         try:
             resp = requests.get(path, timeout=30, **lookup_kwargs)
         except requests.exceptions.ConnectionError:
-            resp = requests.get(path_conjurcloud, timeout=30, **lookup_kwargs)
-        except:
             resp = requests.get(path_conjurcloud, timeout=30, **lookup_kwargs)
     raise_for_status(resp)
     return resp.text
--- a/awx/main/credential_plugins/conjur.py
+++ b/awx/main/credential_plugins/conjur.py
@@ -68,7 +68,12 @@ def conjur_backend(**kwargs):
     with CertFiles(cacert) as cert:
         # https://www.conjur.org/api.html#authentication-authenticate-post
         auth_kwargs['verify'] = cert
-        resp = requests.post(urljoin(url, '/'.join(['api', 'authn', account, username, 'authenticate'])), **auth_kwargs)
+        try:
+            resp = requests.post(urljoin(url, '/'.join(['authn', account, username, 'authenticate'])), **auth_kwargs)
+        except requests.exceptions.ConnectionError:
+            resp = requests.post(urljoin(url, '/'.join(['api', 'authn', account, username, 'authenticate'])), **auth_kwargs)
+        except:
+            raise
     raise_for_status(resp)
     token = resp.content.decode('utf-8')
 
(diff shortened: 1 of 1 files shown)


infamousjoeg:
Fixes #13439

awx/main/credential_plugins/conjur.py
```diff
@@ -68,7 +68,12 @@ def conjur_backend(**kwargs):
     with CertFiles(cacert) as cert:
         # https://www.conjur.org/api.html#authentication-authenticate-post
         auth_kwargs['verify'] = cert
-        resp = requests.post(urljoin(url, '/'.join(['api', 'authn', account, username, 'authenticate'])), **auth_kwargs)
+        try:
+            resp = requests.post(urljoin(url, '/'.join(['authn', account, username, 'authenticate'])), **auth_kwargs)
+        except requests.exceptions.ConnectionError:
```
fosterseth, line 73: any particular reason to have the redundant exception here and below? should it just be `except:` , or did you have a plan to add to this in the future

obaranov:
I verified this fix with docker-compose deployment. It worked fine. I was able to use Conjur Opensource for lookup.

awx/main/credential_plugins/conjur.py
infamousjoeg: What this code block is doing is trying to send the API request to the targeted Conjur service first with `/api` provided in the URI.  This will work for self-hosted Conjur Secrets Manager Enterprise and Conjur Cloud.  If a `requests.exceptions.ConnectionError` is returned, this means that a Conjur service does exist, however, it does not support `/api` in the URI.  This would be if Conjur Open Source was being used.  Therefore, we capture the exception and attempt another API request to the Conjur service, this time _without_ the `/api` in the URI.  Any other errors or further errors after the attempt should be handled the same way errors are in Credential Plugins, hence no bare `except`.

fosterseth: "first with /api provided in the URI." is it backwards?

seems we first attempt to without `/api`, then with.

first try is

`'/'.join(['authn', account, username, 'authenticate']`

infamousjoeg: Sorry, you're right. I was looking on mobile and it was skewing my view.  Since API linting is failing on the bare `except`, I'm going to remove that and just keep the `except requests.exceptions.HTTPConnection` handling.

So the code block goes...
1. Try `/authn/` which is supported by Conjur Enterprise and Conjur Open Source.  If HTTPConnection error returned...
2. Try `/api/authn` which is supported by Conjur Cloud.
3. Regardless of success/failure, `raise_for_status` and toast notification in AWX will display results of test, if testing.

I'll make the change now and re-request review.

fosterseth (approved):

fosterseth:
thanks, looks good!

PR #13458, opened 2023-01-22
[bulk job launch feature]add some helpers functions in validate and some other minor fixes

jainnikhil30:
Cleaning up the validate function and some other minor fixes.

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -4490,13 +4490,17 @@ class BulkJobNodeSerializer(serializers.Serializer):
             'limit',
             'labels',
             'instance_groups',
-            'execution_environment' 'scm_branch',
-            'verbosity' 'forks' 'char_prompts',
+            'execution_environment',
+            'scm_branch',
+            'verbosity',
+            'forks',
+            'char_prompts',
             'diff_mode',
             'extra_data',
             'job_slice_count',
             'job_tags',
-            'job_type' 'skip_tags',
+            'job_type',
+            'skip_tags',
             'survey_passwords',
             'timeout',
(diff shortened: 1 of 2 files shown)


awx/api/serializers.py
```diff
@@ -4490,62 +4490,53 @@ class Meta:
...
+    limit= serializers.CharField(write_only=True, required=False, allow_blank=False)
+    scm_branch= serializers.CharField(write_only=True, required=False, allow_blank=False)
+    # not implemented yet
     # webhook_service: null,  # Here we can use PrimaryKeyRelatedField so it will automagically do rbac/turn into object, I think, I'm actually not sure how to use this
     # webhook_credential: null,  # Here we can use PrimaryKeyRelatedField so it will automagically do rbac/turn into object  I think, I'm actually not sure how to use this
-    skip_tags: serializers.CharField(write_only=True, required=False, allow_blank=False)
-    job_tags: serializers.CharField(write_only=True, required=False, allow_blank=False)
-    is_bulk_job: serializers.BooleanField(default=True)
+    skip_tags= serializers.CharField(write_only=True, required=False, allow_blank=False)
+    job_tags= serializers.CharField(write_only=True, required=False, allow_blank=False)
```
kdelee, line 4530: I think you probably need to have a venv with `black` installed and run `make black` to make this formatting compliant with the rules they hve on the PR checks

also, they use flake8 but I've never figured out the make command. you can check the errors on the ci checks or you can figure out which make command does that

kdelee (approved):
looks good, I would start thinking about getting the formatting compliant with the PR checks which involves running `black` and `flake8`

PR #13369, opened 2022-12-22
[wip] Bulk host add

kdelee:
##### SUMMARY

@ffirg :gift: 

related: https://github.com/ansible/awx/issues/6682
This RFE was previously closed, but I propose this proof of concept as a counterpoint to the arguments against this. I don't think its an unreasonable lift, and I think we can respect our contracts around:
  * RBAC Roles that dictate who can add hosts to an inventory (even be more restrictive if we want to be)
  * Host uniqueness in an inventory
  * Host limits at org level
  * Host limits at system level

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### ADDITIONAL INFORMATION

Many users administer their inventories in systems outside controller and want to be able to create inventories via the api, but adding them one by one is too slow. Instead, let's do all the uniqueness, RBAC and system/organization limit checks in one go as well as leverage `bulk_create` to minimize our interaction with database. Additionally rendering the return objects is relatively expensive. Instead, just return a link to the host list if you want to go see.

Example use:

POST to `/api/v2/bulk/host_create` (I think for #13358 I want to move that API endpoint to /api/v2/bulk/job_launch)
```
[...]

PR #13462, opened 2023-01-24
[Bulk API] Add Bulk add host and bulk job launch

jainnikhil30:
##### SUMMARY
Replaces #13358 and #13369 

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### ADDITIONAL INFORMATION

- With this change you could launch a bulk job as following:

Post to `/api/v2/bulk/job_launch`
```
{
"name": "Bulk Job Launch",
"jobs": [
   {"unified_job_template": 9, "identifier":"foo", "limit": "kansas", "credentials": [1], "inventory": 1, "execution_environment": 3},
   {"unified_job_template": 7, "identifier":"bar", "limit": "kansas", "credentials": [1], "inventory": 1, "execution_environment": 3},
   {"unified_job_template": 7, "limit": "kansas", "credentials": [1], "inventory": 1, "execution_environment": 3}
],
"inventory": 2
}
```

The above will launch a workflow job with 3 nodes in it.

- One could also bulk add host as following:

Post to `/api/v2/bulk/host_create`
```
{"inventory": 1, "hosts": [{"name": "foo-0-2022-12-22 01:29:08.995393"}, {"name": "foo-1-2022-12-22 01:29:08.995393"}, {"name": "foo-2-2022-12-22 01:29:08.995393"}, {"name": "foo-3-2022-12-22 01:29:08.995393"}, {"name": "foo-4-2022-12-22 01:29:08.995393"}, {"name": "foo-5-2022-12-22 01:29:08.995393"}, {"name": "foo-6-2022-12-22 01:29:08.995393"}, {"name": "foo-7-2022-12-22 01:29:08.995393"}, {"name": "foo-8-2022-12-22 01:29:08.995393"}, {"name": "foo-9-2022-12-22 01:29:08.995393"}, {"name": "foo-10-2022-12-22 01:29:08.995393"}, {"name": "foo-11-2022-12-22 01:29:08.995393"}, {"name": "foo-12-2022-12-22 01:29:08.995393"}, {"name": "foo-13-2022-12-22 01:29:08.995393"}, {"name": "foo-14-2022-12-22 01:29:08.995393"}, {"name": "foo-15-2022-12-22 01:29:08.995393"}, {"name": "foo-16-2022-12-22 01:29:08.995393"}, {"name": "foo-17-2022-12-22 01:29:08.995393"}, {"name": "foo-18-2022-12-22 01:29:08.995393"}, {"name": "foo-19-2022-12-22 01:29:08.995393"}, {"name": "foo-20-2022-12-22 01:29:08.995393"}, {"name": "foo-21-2022-12-22 01:29:08.995393"}, {"name": "foo-22-2022-12-22 01:29:08.995393"}, {"name": "foo-23-2022-12-22 01:29:08.995393"}, {"name": "foo-24-2022-12-22 01:29:08.995393"}, {"name": "foo-25-2022-12-22 01:29:08.995393"}, {"name": "foo-26-2022-12-22 01:29:08.995393"}, {"name": "foo-27-2022-12-22 01:29:08.995393"}, {"name": "foo-28-2022-12-22 01:29:08.995393"}, {"name": "foo-29-2022-12-22 01:29:08.995393"}, {"name": "foo-30-2022-12-22 01:29:08.995393"}, {"name": "foo-31-2022-12-22 01:29:08.995393"}, {"name": "foo-32-2022-12-22 01:29:08.995393"}, {"name": "foo-33-2022-12-22 01:29:08.995393"}, {"name": "foo-34-2022-12-22 01:29:08.995393"}, {"name": "foo-35-2022-12-22 01:29:08.995393"}, {"name": "foo-36-2022-12-22 01:29:08.995393"}, {"name": "foo-37-2022-12-22 01:29:08.995393"}, {"name": "foo-38-2022-12-22 01:29:08.995393"}, {"name": "foo-39-2022-12-22 01:29:08.995393"}, {"name": "foo-40-2022-12-22 01:29:08.995393"}, {"name": "foo-41-2022-12-22 01:29:08.995393"}, {"name": "foo-42-2022-12-22 01:29:08.995393"}, {"name": "foo-43-2022-12-22 01:29:08.995393"}, {"name": "foo-44-2022-12-22 01:29:08.995393"}, {"name": "foo-45-2022-12-22 01:29:08.995393"}, {"name": "foo-46-2022-12-22 01:29:08.995393"}, {"name": "foo-47-2022-12-22 01:29:08.995393"}, {"name": "foo-48-2022-12-22 01:29:08.995393"}, {"name": "foo-49-2022-12-22 01:29:08.995393"}, {"name": "foo-50-2022-12-22 01:29:08.995393"}, {"name": "foo-51-2022-12-22 01:29:08.995393"}, {"name": "foo-52-2022-12-22 01:29:08.995393"}, {"name": "foo-53-2022-12-22 01:29:08.995393"}, {"name": "foo-54-2022-12-22 01:29:08.995393"}, {"name": "foo-55-2022-12-22 01:29:08.995393"}, {"name": "foo-56-2022-12-22 01:29:08.995393"}, {"name": "foo-57-2022-12-22 01:29:08.995393"}, {"name": "foo-58-2022-12-22 01:29:08.995393"}, {"name": "foo-59-2022-12-22 01:29:08.995393"}, {"name": "foo-60-2022-12-22 01:29:08.995393"}, {"name": "foo-61-2022-12-22 01:29:08.995393"}, {"name": "foo-62-2022-12-22 01:29:08.995393"}, {"name": "foo-63-2022-12-22 01:29:08.995393"}, {"name": "foo-64-2022-12-22 01:29:08.995393"}, {"name": "foo-65-2022-12-22 01:29:08.995393"}, {"name": "foo-66-2022-12-22 01:29:08.995393"}, {"name": "foo-67-2022-12-22 01:29:08.995393"}, {"name": "foo-68-2022-12-22 01:29:08.995393"}, {"name": "foo-69-2022-12-22 01:29:08.995393"}, {"name": "foo-70-2022-12-22 01:29:08.995393"}, {"name": "foo-71-2022-12-22 01:29:08.995393"}, {"name": "foo-72-2022-12-22 01:29:08.995393"}, {"name": "foo-73-2022-12-22 01:29:08.995393"}, {"name": "foo-74-2022-12-22 01:29:08.995393"}, {"name": "foo-75-2022-12-22 01:29:08.995393"}, {"name": "foo-76-2022-12-22 01:29:08.995393"}, {"name": "foo-77-2022-12-22 01:29:08.995393"}, {"name": "foo-78-2022-12-22 01:29:08.995393"}, {"name": "foo-79-2022-12-22 01:29:08.995393"}, {"name": "foo-80-2022-12-22 01:29:08.995393"}, {"name": "foo-81-2022-12-22 01:29:08.995393"}, {"name": "foo-82-2022-12-22 01:29:08.995393"}, {"name": "foo-83-2022-12-22 01:29:08.995393"}, {"name": "foo-84-2022-12-22 01:29:08.995393"}, {"name": "foo-85-2022-12-22 01:29:08.995393"}, {"name": "foo-86-2022-12-22 01:29:08.995393"}, {"name": "foo-87-2022-12-22 01:29:08.995393"}, {"name": "foo-88-2022-12-22 01:29:08.995393"}, {"name": "foo-89-2022-12-22 01:29:08.995393"}, {"name": "foo-90-2022-12-22 01:29:08.995393"}, {"name": "foo-91-2022-12-22 01:29:08.995393"}, {"name": "foo-92-2022-12-22 01:29:08.995393"}, {"name": "foo-93-2022-12-22 01:29:08.995393"}, {"name": "foo-94-2022-12-22 01:29:08.995393"}, {"name": "foo-95-2022-12-22 01:29:08.995393"}, {"name": "foo-96-2022-12-22 01:29:08.995393"}, {"name": "foo-97-2022-12-22 01:29:08.995393"}, {"name": "foo-98-2022-12-22 01:29:08.995393"}, {"name": "foo-99-2022-12-22 01:29:08.995393"}]}
```
#### General API:

- [x] OPTIONS request responds with description of schema accepted
- [x] POST accepts inventory and hosts desired to create, responds with how many hosts created, and link to inventory
- [x] Have POST return id's of the created hosts
- [x]  Create ActivityStream entry/entries to create

#### RBAC
- [x] org admin for org on inventory can add hosts
- [x] inventory admin at org level for org of inventory can add hosts
- [x] superuser can add hosts, bypass rbac
- [x] inventory admin of particular inventory can add hosts

#### CLI/collection/awxkit

- [x] add collection module for bulk inventory add
- [x] add cli support for bulk inventory add
- [x] add awxkit support for bulk inventory add

#### Testing

- [x] add awx functional test asserting on number of queries it takes
- [x] add awx functional test for RBAC
- [x] add integration tests for api
- [x] add integration tests for collection
- [x] add integration tests for cli

#### At WorkflowJob level:
- [x] reach feature parity with things you can normally set on a WorkflowJobTemplate that get inherited by the WorkflowJob
- [x] "name" (done)
- [x] "description": "",
- [x] "extra_vars": "",
- [x] "organization": (see below in RBAC section)
- [x] "inventory": null,
- [x] "limit": "",
- [x] "scm_branch": ""
- [x] "skip_tags": "",
- [x] "job_tags": ""

#### At workflow job node level:
- [x] implement RBAC and appropriate relating of other related objects
- [x] inventory
- [x] credentials
- [x] instance groups
- [x] execution environment
- [x] labels
- [x] implement schema check and assignment of other character/integer fields
- [x] 'limit'
- [x] 'char_prompts',
- [x] 'diff_mode',
- [x] 'extra_data',
- [x] 'forks',
- [x] 'job_slice_count',
- [x] 'job_tags',
- [x] 'job_type'
- [x] 'scm_branch',
- [x] 'skip_tags',
- [x] 'survey_passwords',
- [x] 'timeout',
- [x] 'verbosity',

#### RBAC
- [x] add "is_bulk_job" field to WorkflowJob model
- [x] associate the user organization to the WorkflowJob if they only have one org, or allow them to specify an organization they belong to as the org (unless they are sysadmin, in which case they can assocaite any org). If user has multiple organizations, return a validation error and tell them they need to pick one
- [x] Anyone who launched a bulk job can see the one they launched
- [x] An org admin for the Organization on a bulk job can see that bulk job
- [x] open up RBAC access for WorkflowJob and WorkflowJobNodes to allow normal users view bulk jobs they launched in detail view. They won't see them in the general job list view
- [x] check that relaunch is prohibited except for super user. we get this for free since its "orphaned" (has no wfjt)
- [x] record workflow job launch in activity stream.

#### Maintainablity
- [x] figure out if there is a way we can share some code between access.py and the validation method to reduce risk of drift between the two
- [x] figure out if we can chunk up the validate method into some more sensible helper methods to make it a bit more readable/have logical sections. E.g. "_validate_permissions", "_objectify_attrs_for_create"
- [x] Add a configuration for the MAX bulk host and  MAX bulk job launch 

#### General API functionality
- [x] provide /api/v2/bulk/job_launch
- [x] list of individual nodes will be accessed via workflow_job_nodes related item on the workflow job returned
- [x] return something sensible to a GET to /api/v2/bulk/job_launch. Use other launch endpoints as a reference. They appear to return a json blob with all the default values of the items you could launch
- [x] in the DjangoRestFramework API browser thing, in the "POST" form, it appears there is someway you can populate it with sample POST data. Provide something sensible for /api/v2/bulk_jobs/launch
- [x] Provide some sensible OPTIONS response to /api/v2/bulk/job_launch

#### Tests
- [x] awxkit part to expose the api/v2/bulk
- [x] awxkit part to expose api/v2/bulk/job_launch
- [x] Add tests that do some validation that # of queries stays reasonable
- [x] Test with LOTS of unified_job_templates in existence in the system and regular users. Concern is RBAC check queries could return a lot of pk's -- question is, where does this become a problem. I did nominal testing with ~ 10k job templates, didn't appear to be an issue but worth poking at some more.
- [x] Add tests that do some validation of RBAC access
- [x] Verify current behavior of launching jobs with "promptable" items that are either not allowed by the job template or you don't have access to.


awx/api/serializers.py
```diff
@@ -4453,6 +4577,335 @@ def validate(self, attrs):
...
+        )
+
+
+class BulkJobLaunchSerializer(BaseSerializer):
+    name = serializers.CharField(default='Bulk Job Launch', max_length=512, write_only=True, required=False, allow_blank=True)  # limited by max name of jobs
+    job_node_help_text = _('List of jobs to be launched, JSON. e.g. [{"unified_job_template": 7}, {"unified_job_template": 10}]')
+    jobs = BulkJobNodeSerializer(many=True, allow_empty=False, write_only=True, max_length=1000, help_text=job_node_help_text)
+    description = serializers.CharField(write_only=True, required=False, allow_blank=False)
+    extra_vars = serializers.CharField(write_only=True, required=False, allow_blank=False)
+    organization = serializers.PrimaryKeyRelatedField(
```
AlanCoding, line 4574: I am confused what `organization` does, and to a lesser extent, name too.

Normally, when you launch or update something (a UJT, unified job template), the update/job inherits the name from the UJT, and inherits the organization from the UJT.

You could give something job a name unrelated to its template, and this may make it hard for the user to _find_ and understand the jobs, but there's no fundamental problem. But for `organization`, we depend on organization being the same between the job & the UJT for permissions to be coherent.

jainnikhil30: @AlanCoding Right, there is no ujt for a bulk job. We are not using any ujt, we are creating the workflow job directly, which is why we cant inherit the orgs and name from a ujt. We are assigning them directly.

AlanCoding: Ok that seems workable, I was operating off information from a prior demo.

AlanCoding: do you have some docs that describe the current form of the feature?

jainnikhil30: Not yet, but let me add some docs.

jainnikhil30: added some docs for the feature.

docs/bulk_api.md
```diff
@@ -0,0 +1,39 @@
...
+"name": "Bulk Job Launch",
+"jobs": [
+   {"unified_job_template": 7, "identifier":"foo", "limit": "kansas", "credentials": [1]},
+   {"unified_job_template": 8, "identifier":"bar", "inventory": 1, "execution_environment": 3},
+   {"unified_job_template": 9}
+]
+}
+```
+
+The above will launch a workflow job with 3 nodes in it.
```
AlanCoding, line 26: Could you add something to document the use of top-level parameters? Like if `inventory` was provided as a top-level key. I think the feature is that it will place those as prompts on the workflow job. But that's from somewhat indirect evidence from the tests, I am still not clear on the intent.

awx/api/serializers.py
```diff
@@ -1959,6 +1964,137 @@ def to_representation(self, obj):
...
+    variables = serializers.CharField(write_only=True, required=False)
+
+    class Meta:
+        fields = (
+            'name',
+            'enabled',
+            'instance_id',
+            'description',
+            'variables',
+        )
```
AlanCoding, line 1961: Why don't you specify `model = Host` in the `Meta`? That seems like the most obvious thing to do. I'm not sure if you should even need to declare the fields manually, although maybe subtract `inventory` from the fields.

awx/api/serializers.py
```diff
@@ -1959,6 +1964,137 @@ def to_representation(self, obj):
...
+
+    def raise_if_cannot_add_hosts(self, attrs):
+        validation_info = get_licenser().validate()
+
+        org = attrs['inventory'].organization
+
+        if org:
+            org_active_count = Host.objects.org_active_count(org.id)
+            new_hosts = [h['name'] for h in attrs['hosts']]
+            org_net_new_host_count = Host.objects.filter(inventory__organization=org.id).exclude(name__in=new_hosts).count()
```
AlanCoding: This name `org_net_new_host_count` doesn't jive with the query in my head. The new hosts, excluding name conflicts, would be `len(new_hosts)`, which is a list of host names. This query is counting hosts in the database, excluding any names in the new set of hosts. The name suggests that it is a count of the new hosts, excluding those already in the database (in the organization).

So what I read from the name would suggest

```python
org_net_new_host_count = len(new_hosts) - Host.objects.filter(inventory__organization=org.id).exclude(name__in=new_hosts).count()
```

But even this falls short, because of the corner case where a new host has a name which is present in _two_ existing hosts in the organization. This is why we do the `order_by` and `distinct` in the host count stuff. You could add this to make the above plausibly correct.

awx/api/serializers.py
```diff
@@ -1959,6 +1964,137 @@ def to_representation(self, obj):
...
+                org_admin_orgs = {tup[0] for tup in Organization.accessible_pk_qs(request.user, 'admin_role')}
+                inv_admin_orgs = {tup[0] for tup in Organization.accessible_pk_qs(request.user, 'inventory_admin_role')}
+                is_org_admin = inv.organization.id in org_admin_orgs
+                is_org_inv_admin = inv.organization.id in inv_admin_orgs
+            else:
+                is_org_admin = False
+                is_org_inv_admin = False
+            # This may not work, need to figure out what the role is called
+            is_inventory_admin = inv.admin_role.members.filter(id=request.user.id).exists()
+            if not any([is_inventory_admin, is_org_admin, is_org_inv_admin]):
```
AlanCoding: For all of this, why is it not just `if request.user not in inv.admin_role:`. The whole point of the Role model is to abstract out these nuances.

awx/api/serializers.py
```diff
@@ -1959,6 +1964,137 @@ def to_representation(self, obj):
...
+                raise serializers.ValidationError(_(f'Inventory with id {inv.id} not found or lack permissions to add hosts.'))
+        current_hostnames = {h[0] for h in Host.objects.filter(inventory=inv).values_list('name').all()}
+        new_names = [host['name'] for host in attrs['hosts']]
+        duplicate_new_names = [n for n in new_names if n in current_hostnames or new_names.count(n) > 1]
+        if duplicate_new_names:
+            raise serializers.ValidationError(_(f'Hostnames must be unique in an inventory. Duplicates found: {duplicate_new_names}'))
+
+        self.raise_if_cannot_add_hosts(attrs)
```
AlanCoding: It might be good to name this something more specific like `raise_if_host_counts_violated`

awx/api/serializers.py
```diff
@@ -1959,6 +1964,137 @@ def to_representation(self, obj):
...
+
+    def create(self, validated_data):
+        # This assumes total_hosts is up to date, and it can get out of date if the inventory computed fields have not been updated lately.
+        # If we wanted to side step this we could query Hosts.objects.filter(inventory...)
+        old_total_hosts = validated_data['inventory'].total_hosts
+        result = [Host(**attrs) for attrs in validated_data['hosts']]
+        try:
+            Host.objects.bulk_create(result)
+        except Exception as e:
+            raise serializers.ValidationError({"detail": _(f"{e}")})
```
AlanCoding: the `_` is a translation marker, and `_(f"{e}")` does not make sense as a string to send to a translator. It would make sense to mark it for translation if if it said `_(f"Cannot do that, host creation error: {e}")`, because that can be translated into other languages.

awx/api/serializers.py
```diff
@@ -4419,6 +4555,333 @@ def validate(self, attrs):
         return accepted
 
 
+class BulkJobNodeSerializer(serializers.Serializer):
+    # if we can find out the user, we can filter down the UnifiedJobTemplate objects
+    unified_job_template = serializers.IntegerField(
```
AlanCoding, line 4539: Kind of strange that you use a `PrimaryKeyRelatedField` for inventory but an `IntegerField` here

awx/api/serializers.py
```diff
@@ -4419,6 +4555,333 @@ def validate(self, attrs):
         return accepted
 
 
+class BulkJobNodeSerializer(serializers.Serializer):
+    # if we can find out the user, we can filter down the UnifiedJobTemplate objects
```
AlanCoding: not sure what this comment means.

awx/api/templates/api/bulk_host_create_view.md
```diff
@@ -0,0 +1,16 @@
...
+```
+{
+"inventory": 1,
+"hosts": [
+    {"name": "example1.com"},
+    {"name": "example2.com"}
+]
+
+}
+```
```
AlanCoding: I would really like to see info about the returned data structure, and this seems like the most appropriate place. In the returned data, for instance, you provide a "url" key, but it doesn't have the full ordinary Host serializer, almost certainly due to performance considerations. That really needs to be spelled out to the user from the get-go, because I find it very likely that people will get the response data and do something with it. This could be done with an example, that's fine. It might be good to include extra_vars, because vars format is always a pain.

awx/api/serializers.py
```diff
@@ -4419,6 +4555,333 @@ def validate(self, attrs):
...
+        default=None,
+        allow_null=True,
+        help_text=_('Inherit permissions from organization roles.'),
+    )
+    inventory = serializers.PrimaryKeyRelatedField(queryset=Inventory.objects.all(), required=False, write_only=True)
+    limit = serializers.CharField(write_only=True, required=False, allow_blank=False)
+    scm_branch = serializers.CharField(write_only=True, required=False, allow_blank=False)
+    # not implemented yet
+    # webhook_service: null,  # Here we can use PrimaryKeyRelatedField so it will automagically do rbac/turn into object, I think, I'm actually not sure how to use this
+    # webhook_credential: null,  # Here we can use PrimaryKeyRelatedField so it will automagically do rbac/turn into object  I think, I'm actually not sure how to use this
```
AlanCoding: The `BulkJobLaunchSerializer` creates a workflow job, and I think these fields are present in a WFJT, so I get why you are saying this. But it seems completely unreasonable to me to configure these in an ad-hoc launch, they should just make a WFJT in that case, and you could just put a statement in docs to that effect.

AlanCoding: Looking at this again, I don't think it _can_ make sense, because github can't call back to a template if there is no template.

Don't even document it, just delete these comments.

awx/api/serializers.py
```diff
@@ -4419,6 +4555,333 @@ def validate(self, attrs):
...
+            extra_vars_dict = parse_yaml_or_json(attrs['extra_vars'])
+            attrs['extra_vars'] = json.dumps(extra_vars_dict)
+        attrs = super().validate(attrs)
+        return attrs
+
+    def create(self, validated_data):
+        job_node_data = validated_data.pop('jobs')
+        # FIXME: Need to set organization on the WorkflowJob in order for users to be able to see it --
+        # normally their permission is sourced from the underlying WorkflowJobTemplate
+        # maybe we need to add Organization to WorkflowJob
```
AlanCoding: WorkflowJob does have an organization, and you have already added it to this serializer. So delete comment, as work has been finished?

awx/api/serializers.py
```diff
@@ -4419,6 +4555,333 @@ def validate(self, attrs):
...
+            raise serializers.ValidationError(_(f"Inventories {not_allowed} not found."))
+
+    def check_credential_permission(self, request, requested_use_credentials):
+        accessible_use_credentials = {tup[0] for tup in Credential.accessible_pk_qs(request.user, 'use_role').all()}
+        if requested_use_credentials - accessible_use_credentials:
+            not_allowed = requested_use_credentials - accessible_use_credentials
+            raise serializers.ValidationError(_(f"Credentials {not_allowed} not found."))
+
+    def check_label_permission(self, requested_use_labels):
+        accessible_use_labels = {tup.id for tup in Label.objects.all()}
```
AlanCoding: hold on, `Label.objects.all()`... why even run this method? Or you're not just checking permission, but checking permission and existence. Do labels not have any relevant permission checks? I think they do.

awx/api/serializers.py
```diff
@@ -4419,6 +4555,333 @@ def validate(self, attrs):
...
+        accessible_use_inventories = {tup[0] for tup in Inventory.accessible_pk_qs(request.user, 'use_role')}
+        if requested_use_inventories - accessible_use_inventories:
+            not_allowed = requested_use_inventories - accessible_use_inventories
+            raise serializers.ValidationError(_(f"Inventories {not_allowed} not found."))
+
+    def check_credential_permission(self, request, requested_use_credentials):
+        accessible_use_credentials = {tup[0] for tup in Credential.accessible_pk_qs(request.user, 'use_role').all()}
+        if requested_use_credentials - accessible_use_credentials:
+            not_allowed = requested_use_credentials - accessible_use_credentials
+            raise serializers.ValidationError(_(f"Credentials {not_allowed} not found."))
```
AlanCoding: they may be viable - the user doesn't have use role for them. Maybe say "not found or you don't have permission" because that's pretty normal to avoid divulging info the user shouldn't have.

awx_collection/plugins/modules/bulk_host_create.py
```diff
@@ -0,0 +1,99 @@
...
+          description:
+            - Variables to use for the host.
+          type: dict
+        instance_id:
+          description:
+            - instance_id to use for the host.
+          type: str
+    inventory:
+      description:
+        - Inventory ID the hosts should be made a member of.
```
john-westcott-iv: Most of the params we provide for relations take name or ID you'd have to make this param a string.

awx_collection/plugins/modules/bulk_host_create.py
```diff
@@ -0,0 +1,99 @@
...
+    # Create a module for ourselves
+    module = ControllerAPIModule(argument_spec=argument_spec)
+
+    # Extract our parameters
+    inventory = module.params.get('inventory')
+    hosts = module.params.get('hosts')
+
+    for h in hosts:
+      if 'variables' in h:
+        h['variables'] = json.dumps(h['variables'])
```
john-westcott-iv: I don't think you need json.dumps here since the parameter should be forced into a dict already.

awx_collection/plugins/modules/bulk_host_create.py
```diff
@@ -0,0 +1,99 @@
...
+    # Launch the jobs
+    result = module.post_endpoint("bulk/host_create", data={"inventory": inventory, "hosts": hosts})
+
+    if result['status_code'] != 201:
+        module.fail_json(msg="Failed to create hosts, see response for details", response=result)
+
+    module.json_output['changed'] = True
```
john-westcott-iv, line 96: If the bulk import fails (you don't get a 201) some of the items may have imported and changed the system right? If so, is there any way to tell from the response? Ansible usually seems to favor reporting changed of there is any posibility of something having been changed. We could make these modules like the command module where it assums changed? Along those lines, what happens if all of the hosts are already imported? Will we still get a 201? If so, we might want to expand the logic in here to see if the nodes are part of the inventory already (before the bulk import) and then only do the import if nodes are missing to give a more accurate indication of changed/ok.

awx_collection/plugins/modules/bulk_host_create.py
```diff
@@ -0,0 +1,99 @@
...
+
+    # Extract our parameters
+    inventory = module.params.get('inventory')
+    hosts = module.params.get('hosts')
+
+    for h in hosts:
+      if 'variables' in h:
+        h['variables'] = json.dumps(h['variables'])
+    # Launch the jobs
+    result = module.post_endpoint("bulk/host_create", data={"inventory": inventory, "hosts": hosts})
```
john-westcott-iv: If you convert inventory to a string for name/id lookup you should be able to do something like: `module.resolve_name_to_id('inventories', inventory)` to get back an ID.

awx_collection/plugins/modules/bulk_job_launch.py
```diff
@@ -0,0 +1,272 @@
...
+      elements: dict
+      suboptions:
+        unified_job_template:
+          description:
+            - Job template ID to use when launching.
+          type: int
+          required: True
+        inventory:
+          description:
+            - Inventory ID applied as a prompt, if job template prompts for inventory
```
john-westcott-iv, line 37: Same ID vs name/id stuff as in the other file.

awx_collection/plugins/modules/bulk_job_launch.py
```diff
@@ -0,0 +1,272 @@
...
+    jobs:
+      - unified_job_template: 7
+        skip_tags: foo
+      - unified_job_template: 10
+        limit: foo
+        extra_data:
+          food: carrot
+          color: orange
+    limit: bar
+    extra_vars: # these override / extend extra_data at the job level
```
john-westcott-iv, line 189: This comment feels conflicting for extra vars. Do the upper level override or extend?

awx_collection/plugins/modules/bulk_job_launch.py
```diff
@@ -0,0 +1,272 @@
...
+      food: grape
+      animal: owl
+    inventory: 1
+
+- name: Launch bulk jobs with lookup plugin
+  bulk_job_launch:
+    name: My Bulk Job Launch
+    jobs:
+      - unified_job_template: 7
+      - unified_job_template: "{{ lookup('awx.awx.controller_api', 'job_templates', query_params={'name': 'Demo Job Template'}, return_ids=True) }}"
```
john-westcott-iv: Is a unified_job_tempalte ID always the same as the ID of the job template itself? Also, I think this returns a list and `name` is not unique for a JT you might want to add `expect_one=True`.

awx_collection/plugins/modules/bulk_job_launch.py
```diff
@@ -0,0 +1,272 @@
...
+        organization=dict(type='int'),
+        inventory=dict(type='int'),
+        limit=dict(),
+        scm_branch=dict(),
+        extra_vars=dict(type='dict'),
+        job_tags=dict(),
+        skip_tags=dict(),
+        wait=dict(required=False, default=True, type='bool'),
+        interval=dict(required=False, default=2.0, type='float'),
+        timeout=dict(required=False, default=None, type='int'),
```
john-westcott-iv: It looks like in the doc `timeout` is a sub-option of the jobs, not at a the main level. Should it be in both or just one?

awx_collection/plugins/modules/bulk_job_launch.py
```diff
@@ -0,0 +1,272 @@
...
+    interval = module.params.get('interval')
+    name = module.params.get('name')
+
+    # Launch the jobs
+    result = module.post_endpoint("bulk/job_launch", data=post_data)
+
+    if result['status_code'] != 201:
+        module.fail_json(msg="Failed to launch bulk jobs, see response for details", response=result)
+
+    module.json_output['changed'] = True
```
john-westcott-iv, line 265: Same question about changed.

awx_collection/test/awx/test_completeness.py
```diff
@@ -73,6 +79,8 @@
     'user': ['new_username', 'organization'],
     # workflow_approval parameters that do not apply when approving an approval node.
     'workflow_approval': ['action', 'interval', 'timeout', 'workflow_job_id'],
+    # bulk
+    'bulk_job_launch': ['interval', 'wait'],
```
john-westcott-iv, line 83: You will probablly need `timeout` here if you move it to the top level.

awx_collection/test/awx/test_completeness.py
```diff
@@ -44,6 +44,12 @@
     'subscriptions',  # Subscription deals with config/subscriptions
 ]
 
+# Add modules with endpoints that are not at /api/v2
+extra_endpoints = {
```
john-westcott-iv, line 48: Nice feature! I wonder if we can leverage this for some of the other modules in `no_endpoint_for_module`.

awx_collection/tests/integration/targets/bulk_host_create/main.yml
```diff
@@ -0,0 +1,51 @@
+---
+- name: Generate a random string for test
+  set_fact:
+    test_id: "{{ lookup('password', '/dev/null chars=ascii_letters length=16') }}"
+  when: test_id is not defined
+
+- name: Generate a unique name
+  set_fact:
+    bulk_host_name: "AWX-Collection-tests-bulk_host_create-{{ test_id }}"
```
john-westcott-iv: This should probablly be `bulk_inv_name`

awx_collection/tests/integration/targets/bulk_host_create/main.yml
```diff
@@ -0,0 +1,51 @@
...
+  bulk_host_create:
+    hosts:
+      - name: "123.456.789.123"
+        description: "myhost1"
+        variables:
+          food: carrot
+          color: orange
+      - name: example.dns.gg
+        description: "myhost2"
+        enabled: false
```
john-westcott-iv, line 38: I would extend this so that you have used all of the possible options under `hosts`to make sure they all work. 
Also, it might be worth having a second identical call to see/document what happens if you run an identical test.

awx_collection/tests/integration/targets/bulk_host_create/main.yml
```diff
@@ -0,0 +1,51 @@
...
+- assert:
+    that:
+      - result is not failed
+
+# cleanup
+- name: Delete inventory
+  inventory:
+    name: "{{ bulk_host_name }}"
+    organization: Default
+    state: absent
```
john-westcott-iv, line 51: Will this also delete any created hosts as well?

awx_collection/tests/integration/targets/bulk_job_launch/main.yml
```diff
@@ -0,0 +1,69 @@
...
+
+- name: Get our collection package
+  controller_meta:
+  register: controller_meta
+
+- name: Generate the name of our plugin
+  set_fact:
+    plugin_name: "{{ controller_meta.prefix }}.controller_api"
+
+- name: Get Inventory
```
john-westcott-iv, line 19: If you provide resolution of names in your module you would't need this step.

awxkit/awxkit/api/resources.py
```diff
@@ -13,6 +13,8 @@ class Resources(object):
     _applications = 'applications/'
     _auth = 'auth/'
     _authtoken = 'authtoken/'
+    _bulk = 'bulk/'
+    _bulk_job_launch = 'bulk/job_launch/'
```
john-westcott-iv, line 17: Are we not need bulk_host_create here?

awxkit/awxkit/cli/options.py
```diff
@@ -174,15 +177,16 @@ def json_or_yaml(v):
...
-                    if isinstance(v, str) and v.startswith('@'):
-                        path = os.path.expanduser(v[1:])
-                        parsed[k] = open(path).read()
+                if expected_type is dict:
+                    for k, v in parsed.items():
+                        # add support for file reading at top-level JSON keys
+                        # (to make things like SSH key data easier to work with)
+                        if isinstance(v, str) and v.startswith('@'):
+                            path = os.path.expanduser(v[1:])
+                            parsed[k] = open(path).read()
```
john-westcott-iv, line 189: Do you not want to expand things with `@` for the list form of this function?

docs/bulk_api.md
```diff
@@ -0,0 +1,39 @@
+# Bulk API Overview
+
+Bulk API endpoints allows to perform bulk operations in single web request. There are currently following bulk api actions:
+- /api/v2/bulk/job_launch
+- /api/v2/bulk/host_create
```
john-westcott-iv, line 5: Do you want to say why I might want to do this over multile API calls? i.e. efficiency of the bulk endpoints, etc?

awx/api/serializers.py
jainnikhil30: @AlanCoding Sure, I can add the model = Host. For the fields I am following the other serializers, most of them define a model but still have all the fields too. Please let me know if there is a specific use case of defining the fields ?

awx/api/serializers.py
jainnikhil30: removed

awx/api/serializers.py
jainnikhil30: Sure, removed

awx/api/serializers.py
jainnikhil30: Oh my bad. corrected that.

awx/api/serializers.py
jainnikhil30: removed

awx/api/serializers.py
jainnikhil30: Sure. Done

awx/api/serializers.py
jainnikhil30: from what I could see there is no permission checks on labels. A user who is part of an organization can see all the labels of that org. Which is why I am checking if the labels exist and user can see it.

awx/api/templates/api/bulk_host_create_view.md
jainnikhil30: Sure, I do have an example of how to use variables in the docs section of the bulk_api feature. But I will add it here too.

awx/api/serializers.py
jainnikhil30: @AlanCoding Great catch ! I edited it to be like this now 

```
org_net_new_host_count = len(new_hosts) - Host.objects.filter(inventory__organization=1, name__in=new_hosts).values('name').distinct().count()
```

Example:

```
In [21]: Host.objects.filter(inventory__organization=1).values('name')
Out[21]: <QuerySet [{'name': 'example1.com'}, {'name': 'example2.com'}, {'name': 'host1'}, {'name': 'host1'}, {'name': 'host2'}, {'name': 'host3'}, {'name': 'host4'}, {'name': 'host5'}, {'name': 'host6'}, {'name': 'localhost'}]>

In [22]: new_hosts = ["host1"]

In [23]: Host.objects.filter(inventory__organization=1, name__in=new_hosts).values('name').distinct().count()
Out[23]: 1

In [24]: len(new_hosts) - Host.objects.filter(inventory__organization=1, name__in=new_hosts).values('name').distinct().count()
Out[24]: 0
```

awx/api/serializers.py
jainnikhil30: Sure

awx/api/serializers.py
jainnikhil30: Great ! thanks for that suggestion. I will use that

awx/api/serializers.py
kdelee: good catch yall :D

awx/api/serializers.py
AlanCoding: > from what I could see there is no permission checks on labels. A user who is part of an organization can see all the labels of that org. Which is why I am checking if the labels exist and user can see it.

Yes, the "permission" check here would be checking that the user has permission to see those labels. That seems to be missing.

awx/api/serializers.py
AlanCoding: Oh yeah, I was wrong. You need the field list. What the model can buy you is that it'll allow you to avoid _defining serializer fields_. The first level gating criteria is that OPTIONS recognizes the field types. I might check out the branch and see if anything remains that can be deleted.

awx/api/serializers.py
jainnikhil30: @AlanCoding Actually there is a reason we did that. If we do a PrimaryKeyRelatedField it increases the number of DB queries we make, which is why we take it as integer and later convert it into a object in get_objectified_jobs. That way we keep the number of queries less even with large number of jobs being launched via bulk job

awx/api/serializers.py
AlanCoding: Yeah, so first test:

```json
                "child": {
                    "type": "nested object",
                    "required": true,
                    "read_only": false,
                    "children": {
                        "name": {
                            "type": "string",
                            "required": true,
                            "read_only": false,
                            "label": "Name",
                            "max_length": 512,
                            "help_text": "Name of this host.",
                            "filterable": true
                        },
                        "enabled": {
                            "type": "boolean",
                            "required": false,
                            "read_only": false,
                            "label": "Enabled",
                            "help_text": "Is this host online and available for running jobs?",
                            "filterable": true,
                            "default": true
                        },
                        "instance_id": {
                            "type": "string",
                            "required": false,
                            "read_only": false,
                            "label": "Instance id",
                            "help_text": "The value used by the remote inventory source to uniquely identify the host",
                            "max_length": 1024,
                            "filterable": true,
                            "default": ""
                        },
                        "description": {
                            "type": "string",
                            "required": false,
                            "read_only": false,
                            "label": "Description",
                            "help_text": "Optional description of this host.",
                            "filterable": true,
                            "default": ""
                        },
                        "variables": {
                            "type": "json",
                            "required": false,
                            "read_only": false,
                            "label": "Variables",
                            "help_text": "Host variables in JSON or YAML format.",
                            "filterable": true,
                            "default": ""
                        }
                    }
                },
```

is what I get after doing

```diff
diff --git a/awx/api/serializers.py b/awx/api/serializers.py
index 14f38d004d..b2043025aa 100644
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -1965,12 +1965,6 @@ class GroupSerializer(BaseSerializerWithVariables):
 
 
 class BulkHostSerializer(HostSerializer):
-    name = serializers.CharField(required=True, allow_blank=False, max_length=512)
-    instance_id = serializers.CharField(required=False, max_length=1024)
-    description = serializers.CharField(required=False)
-    enabled = serializers.BooleanField(default=True, required=False)
-    variables = serializers.CharField(write_only=True, required=False)
-
     class Meta:
         model = Host
         fields = (
```

And this is surprisingly _better_ than with the fields in there. Currently, the enabled field gives:

```json
                        "enabled": {
                            "type": "boolean",
                            "required": false,
                            "read_only": false,
                            "label": "Enabled",
                            "filterable": true,
                            "default": true
                        },
```

Because when you specified a DRF field for `enabled` you didn't give help_text, but the model field already has it. So without defining the fields, it'll infer the properties from the Django model fields on `Host`.

AlanCoding:
I checked out your branch and the help templates were not rendering, so I made a PR to try and clear that up.

https://github.com/jainnikhil30/awx/pull/8

For changes to PRs, I personally prefer the squash & merge option to simplify git history. Otherwise you are welcome to manually apply the changes as you see fit.

awx/api/serializers.py
AlanCoding: ah, to wrap this conversation up, the inventory field I referenced was on `BulkJobLaunchSerializer`, as opposed to this class which is a child serializer. The `PrimaryKeyRelatedField` have a history of performance problems, but if there is only 1 field for the serialized / view, a query or two would be acceptable, but the child serializer may repeat the query 1000 times in the request.

awx/main/access.py
```diff
@@ -2123,7 +2127,11 @@ class WorkflowJobAccess(BaseAccess):
     )
 
     def filtered_queryset(self):
-        return WorkflowJob.objects.filter(unified_job_template__in=UnifiedJobTemplate.accessible_pk_qs(self.user, 'read_role'))
+        return WorkflowJob.objects.filter(
+            Q(unified_job_template__in=UnifiedJobTemplate.accessible_pk_qs(self.user, 'read_role'))
+            | Q(created_by_id=self.user.id, is_bulk_job=True)
+            | Q(organization__in=Organization.objects.filter(Q(admin_role__members=self.user)), is_bulk_job=True)
```
AlanCoding, line 2138: Yikes, this is going to be a tough pill to swallow. For the most part, _this_ query doesn't matter. When you go to the JOBS tab, the UI doesn't list workflow jobs. The UI lists `/api/v2/unified_jobs/` with a bunch of terribly scary filters (for scale) on it.

So right now, these jobs are probably not showing up in the UI for non-superusers. Ping @CFSNM 

How we move forward is not at all clear. The requirement that we would allow organization admins / auditors to see jobs launched by this mechanism (in their organization) is something we should have ideally pushed back on in early feature planning.

the UnifiedJob queryset is this:

https://github.com/ansible/awx/blob/17246ff1130de065557ece97b27b8defac952895/awx/main/access.py#L2501-L2509

It doesn't even contain the `created_by` filter, and importantly _I would push back on adding it_ for performance and stability reasons.

awx/main/access.py
```diff
@@ -1997,7 +1997,11 @@ class WorkflowJobNodeAccess(BaseAccess):
     )
 
     def filtered_queryset(self):
-        return self.model.objects.filter(workflow_job__unified_job_template__in=UnifiedJobTemplate.accessible_pk_qs(self.user, 'read_role'))
+        return self.model.objects.filter(
+            Q(workflow_job__unified_job_template__in=UnifiedJobTemplate.accessible_pk_qs(self.user, 'read_role'))
+            | Q(workflow_job__created_by_id=self.user.id, workflow_job__is_bulk_job=True)
+            | Q(workflow_job__organization__in=Organization.objects.filter(Q(admin_role__members=self.user)), workflow_job__is_bulk_job=True)
```
AlanCoding: As you can imagine, this is going to be a problem along with the WorkflowJob filter. But let me try to open up some options here.

Ideally, I should never care about this filter. It should be rarely used in normal operation. This is the only option to build the API list view of `/api/v2/workflow_job_nodes/`, and will remain that way. However, I see no reason the UI should ever use that. No one should ever need or want a global list of workflow job nodes. Instead, they should use the `/api/v2/workflow_job/42/workflow_nodes/` sub-list. Right now, this still uses this filter (actually worse, it uses this filter with JOIN and DISTINCT to do the parent filter).

It should be uncontroversial and easy to change `get_queryset` on `WorkflowJobWorkflowNodesList` and any similar views to completely bypass this filter. This is because permission is dictated 100% by the parent object, which is the workflow job. As you can see here, all filters flow through `workflow_job__`. So even if this grows to unworkable complexity, this particular filter may not necessarily become a feature blocker if we can make some mitigating changes.

awx/main/access.py
AlanCoding: Wait, maybe the `     | Q(organization__in=org_auditor_qs)` from the UnifiedJob query covers a lot of what you need here. And some of the update to the WorkflowJob queryset may just be bringing that to parity with the unified job queryset. However, this still leaves unsettled the question of created_by.

AlanCoding: Let me offer another way out - require `admin_role` for the `organization` field given in the bulk_launch endpoint. This would avoid the need to filter by `created_by`, but it has the downside of limiting who can use this feature.

If that's not workable, then we should question is workflow jobs are a suitable solution here. Users can still see the _jobs_ in any case, but if we let them do this without giving visibility to the workflow job, I fear this would be objectionable to the point that it's not sustainable.

jainnikhil30: > So right now, these jobs are probably not showing up in the UI for non-superusers. Ping @CFSNM

@AlanCoding The UI maybe using a different query, because as a normal user in the UI I dont see the bulk job, but I do see them in API. 
```
[REST API](https://localhost:8043/api/) [Version 2](https://localhost:8043/api/v2/) [Workflow Job List](https://localhost:8043/api/v2/workflow_jobs/) 

Workflow Job List

GET /api/v2/workflow_jobs/?is_bulk_job=True

HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Vary: Accept
X-API-Node: awx_1
X-API-Product-Name: AWX
X-API-Product-Version: 21.10.3.dev60+g28f275ee25.d20230216
X-API-Query-Count: 19
X-API-Query-Time: 0.007s
X-API-Time: 0.035s

{
    "count": 10,
    "next": null,
    "previous": null,
    "results": [
```

AlanCoding: The UI uses the `/api/v2/unified_jobs/` list. That's what matters. And the problem isn't exactly that you can get them to show, the problem is the complexity of the unified jobs query we are left with, and I am not okay adding `created_by` to that. You have to look at the plans / SQL generated. It's not the obvious, this is complicated by the polymorphic models.

jainnikhil30: @AlanCoding I see. I am not sure how do I move forward ? Currently you can't launch the bulk jobs via UI, is it fine if we say this is a API only feature and the job will be visible via API only ?

At the same time I did some testing and wanted to post some numbers here:

- Created 10,000 (10k) users, assigned them admin access on one of the job templates
```
GET /api/v2/users/

HTTP 200 OK
Allow: GET, POST, HEAD, OPTIONS
Vary: Accept
X-API-Node: awx_1
X-API-Product-Name: AWX
X-API-Product-Version: 21.10.3.dev60+g28f275ee25.d20230216
X-API-Query-Count: 124
X-API-Query-Time: 0.474s
X-API-Time: 1.376s

{
    "count": 10005,
    "next": "[/api/v2/users/?page=2](https://localhost:8043/api/v2/users/?page=2)",
    "previous": null,
    "results": [
```
- Ran multiple bulk jobs with different users (around 7k) jobs:
```
GET /api/v2/workflow_jobs/?is_bulk_job=True

HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Vary: Accept
X-API-Node: awx_1
X-API-Product-Name: AWX
X-API-Product-Version: 21.10.3.dev60+g28f275ee25.d20230216
X-API-Query-Count: 32
X-API-Query-Time: 0.039s
X-API-Time: 0.126s

{
    "count": 6839,
    "next": "[/api/v2/workflow_jobs/?is_bulk_job=True&page=2](https://localhost:8043/api/v2/workflow_jobs/?is_bulk_job=True&page=2)",
    "previous": null,
```
- logged in as one of those 10,000 (10k) user and tried fetching the workflow_jobs api endpoint:
```


GET /api/v2/workflow_jobs/

HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Vary: Accept
X-API-Node: awx_1
X-API-Product-Name: AWX
X-API-Product-Version: 21.10.3.dev60+g28f275ee25.d20230216
X-API-Query-Count: 14
X-API-Query-Time: 0.075s
X-API-Time: 0.145s

{
    "count": 11,
    "next": null,
    "previous": null,
```

API seems to very snappy, X-API-Query-Time and X-API-Time seems to be fine. I wonder if its because we have an index on `create_by_id` in the `main_unifiedjob` table ?

docs/bulk_api.md
jainnikhil30: Done. Added an  example for that

kdelee:
just to save some digging here are the collection sanity test errors
```
ERROR: Found 3 pep8 issue(s) which need to be resolved:
ERROR: plugins/modules/bulk_host_create.py:85:7: E111: indentation is not a multiple of 4
ERROR: test/awx/test_bulk.py:9:1: E302: expected 2 blank lines, found 1
ERROR: test/awx/test_bulk.py:42:39: W292: no newline at end of file
See documentation for help: https://docs.ansible.com/ansible-core/2.14/dev_guide/testing/sanity/pep8.html
ERROR: Found 1 pylint issue(s) which need to be resolved:
ERROR: test/awx/test_bulk.py:42:0: missing-final-newline: Final newline missing
See documentation for help: https://docs.ansible.com/ansible-core/2.14/dev_guide/testing/sanity/pylint.html
ERROR: Found 7 validate-modules issue(s) which need to be resolved:
ERROR: plugins/modules/bulk_host_create.py:0:0: doc-elements-mismatch: Argument 'hosts' in argument_spec does not specify elements,but elements is documented as being dict
ERROR: plugins/modules/bulk_host_create.py:0:0: invalid-documentation: DOCUMENTATION.options.hosts.suboptions.name.require: extra keys not allowed @ data['options']['hosts']['suboptions']['name']['require']. Got True
ERROR: plugins/modules/bulk_host_create.py:0:0: parameter-list-no-elements: Argument 'hosts' in argument_spec defines type as list but elements is not defined
ERROR: plugins/modules/bulk_job_launch.py:0:0: doc-elements-mismatch: Argument 'jobs' in argument_spec does not specify elements,but elements is documented as being dict
ERROR: plugins/modules/bulk_job_launch.py:0:0: parameter-list-no-elements: Argument 'jobs' in argument_spec defines type as list but elements is not defined
ERROR: plugins/modules/bulk_job_launch.py:0:0: parameter-type-not-in-doc: Argument 'timeout' in argument_spec defines type as 'int' but documentation doesn't define type
ERROR: plugins/modules/bulk_job_launch.py:0:0: undocumented-parameter: Argument 'timeout' is listed in the argument_spec, but not documented in the module documentation
```

awx_collection/plugins/modules/bulk_job_launch.py
fosterseth: it does a merge, and top level `extra_vars` takes precedence over the item level `extra_data`

extra_vars
```
{"animal": "cow",
"color": "black"}
```

extra_data
```
{"animal": "dog",
"food": "carrot"}
```

resulting extra_vars on Job
```
{"animal": "cow",
"color": "black",
"food": "carrot"
}
```

awx_collection/plugins/modules/bulk_job_launch.py
fosterseth: fixed thanks

awx_collection/tests/integration/targets/bulk_host_create/main.yml
fosterseth: good catch, fixed

awx/api/serializers.py
kdelee: https://github.com/jainnikhil30/awx/pull/13

awx_collection/tests/integration/targets/bulk_host_create/main.yml
fosterseth: yes it will remove the hosts too

awx/main/tests/functional/test_bulk.py
```diff
@@ -0,0 +1,217 @@
...
+    organization.admin_role.members.add(org_admin)
+    organization.inventory_admin_role.members.add(org_inv_admin)
+    inventory.admin_role.members.add(inventory_admin)
+
+    for u in [org_admin, inventory_admin, org_inv_admin, superuser]:
+        hosts = [{'name': uuid4()} for i in range(num_hosts)]
+        with django_assert_max_num_queries(num_queries):
+            bulk_host_create_response = post(reverse('api:bulk_host_create'), {'inventory': inventory.id, 'hosts': hosts}, u, expect=201).data
+            assert len(bulk_host_create_response['hosts']) == len(hosts), f"unexpected number of hosts created for user {u}"
+
```
fosterseth, line 40: add assertions about content of resulting hosts e.g. name

maybe add host level variables and assert on those (for all tests)

do additional Inventory.objects.all() and assert len(hosts) == num expected hosts, to make sure actual objects are created in db and linked to inventory

awx/main/tests/functional/test_bulk.py
```diff
@@ -0,0 +1,217 @@
...
+    inventory.admin_role.members.add(inventory_admin)
+    inventory.use_role.members.add(use_inv_member)
+    organization.auditor_role.members.add(auditor)
+
+    for indx, u in enumerate([org_admin, inventory_admin, org_inv_admin]):
+        bulk_host_create_response = post(
+            reverse('api:bulk_host_create'), {'inventory': inventory.id, 'hosts': [{'name': f'foobar-{indx}'}]}, u, expect=201
+        ).data
+        assert len(bulk_host_create_response['hosts']) == 1, f"unexpected number of hosts created for user {u}"
+
```
fosterseth, line 74: add same assertions as described in above

awx/main/tests/functional/test_bulk.py
```diff
@@ -0,0 +1,217 @@
...
+    for indx, u in enumerate([member, auditor, use_inv_member]):
+        bulk_host_create_response = post(
+            reverse('api:bulk_host_create'), {'inventory': inventory.id, 'hosts': [{'name': f'foobar2-{indx}'}]}, u, expect=400
+        ).data
```
fosterseth, line 78: assert no hosts were added to inventory in case of 400

awx/main/tests/functional/test_bulk.py
```diff
@@ -0,0 +1,217 @@
...
+        bulk_job_launch_response = post(reverse('api:bulk_job_launch'), {'name': 'Bulk Job Launch', 'jobs': jobs}, normal_user, expect=201).data
+
+    for u in (org_admin,):  # TODO normal_user not working because launched_by not getting set in the tests...does happen in real request
+        bulk_job = get(bulk_job_launch_response['url'], u, expect=200).data
+        assert organization.id == bulk_job['summary_fields']['organization']['id']
+        resp = get(bulk_job_launch_response['related']['workflow_nodes'], u)
+        assert resp.data['count'] == num_jobs
+        for item in resp.data['results']:
+            assert item["unified_job_template"] == jt.id
+
```
fosterseth, line 117: assert inventory, execution environment were set for each resulting Job

awx/main/tests/functional/test_bulk.py
```diff
@@ -0,0 +1,217 @@
...
+    '''
+    normal_user = user('normal_user', False)
+    org1 = Organization.objects.create(name='foo1')
+    org2 = Organization.objects.create(name='foo2')
+    org1.member_role.members.add(normal_user)
+    org2.member_role.members.add(normal_user)
+    jt = JobTemplate.objects.create(name='my-jt', inventory=inventory, project=project, playbook='helloworld.yml')
+    jt.save()
+    jt.execute_role.members.add(normal_user)
+    post(reverse('api:bulk_job_launch'), {'name': 'Bulk Job Launch', 'jobs': [{'unified_job_template': jt.id}]}, normal_user, expect=400)
```
fosterseth: assert on the specific message returned is about the multi org

awx/main/tests/functional/test_bulk.py
```diff
@@ -0,0 +1,217 @@
...
+    '''
+    normal_user = user('normal_user', False)
+    org1 = Organization.objects.create(name='foo1')
+    org2 = Organization.objects.create(name='foo2')
+    jt = JobTemplate.objects.create(name='my-jt', inventory=inventory, project=project, playbook='helloworld.yml')
+    jt.save()
+    org1.member_role.members.add(normal_user)
+    inv = Inventory.objects.create(name='inv1', organization=org2)
+    jt.execute_role.members.add(normal_user)
+    post(reverse('api:bulk_job_launch'), {'name': 'Bulk Job Launch', 'jobs': [{'unified_job_template': jt.id, 'inventory': inv.id}]}, normal_user, expect=400)
```
fosterseth: assert on message being about inventory

awx/main/tests/functional/test_bulk.py
```diff
@@ -0,0 +1,217 @@
...
+
+
```
fosterseth, line 215: add new Test

- set all of the non-related fields on individual Job, and assert resulting Job has those fields
- e.g. extra_data, diff_mode, verbosity, scm_branch, job_type, job_tags, skip_tags, limit, forks, job_slice_count, identifier, timeout

- for extra_data, assert on the merging behavior of the top-level extra_vars field (i.e. show that resulting job has acquired the extra_vars, adhering to merging behavior)

fosterseth:
Idea for future: bulk host create

- add a dry-run mode to the api call, which will return a list of objects that would have been created, but don't actually create them
- this would allow the bulk_host_create collection to be idempotent (does a dry run first, then does real post if needed, and sets changed value accordingly)

awx/main/access.py
AlanCoding: It's very difficult to create the circumstances that people will see and complain about in the real world here. The most concerning endpoint is not this, but the unified jobs, because this gets _slammed_ by the UI, and the dangerous situation is where someone actually makes heavy use of the RBAC system so there is a lot of indirect role memberships, _in addition to_ the type-specific filters.

Right now, looking at the diff, `WorkflowJobAccess.filtered_queryset` is inconsistent with `UnifiedJob.filtered_queryset`. Where this has

```python
Q(organization__in=Organization.objects.filter(Q(admin_role__members=self.user)), is_bulk_job=True)
```

the `UnifiedJobAccess` queryset has

```python
Q(organization__in=org_auditor_qs)
```

Where `org_auditor_qs = Organization.objects.filter(Q(admin_role__members=self.user) | Q(auditor_role__members=self.user))`.

These really need to be the same filters, because a workflow job _is_ a unified job. The same items should appear in `/api/v2/workflow_jobs/` as what appears in `/api/v2/unified_jobs/`, otherwise it's just inconsistent and wrong.

So I agree that you should add this org admin / auditor filter, but you should add the same on as in the `UnifiedJobAccess` filter. And then my suggestion for this feature is to abandon the `created_by` filter.

awx/main/tests/functional/test_bulk.py
jainnikhil30: Sure, done

awx/main/tests/functional/test_bulk.py
jainnikhil30: Sure, done

awx/main/tests/functional/test_bulk.py
jainnikhil30: sure, done

awx/main/tests/functional/test_bulk.py
jainnikhil30: - Added the non-related fields test
- For the merging behavior we have to do it in a integration test. In a functional test I do see the extra_data at the WorkFlowNode level and extra_vars at the WorkFlowJob level. But is playbook able to use them we need to check in a integration level test @CFSNM

awx/main/access.py
kdelee: @AlanCoding I see what you are saying about `/api/v2/unified_jobs/` returning a different list than `/api/v2/workflow_jobs/`. However, in this world, there are no "broken" links. If you view in the UI the child jobs, you can then follow the link to the parent Bulk Job. and in the API you can follow the link for the workflow you spawned.

If consistency is paramount, we could make a second query in the UnifiedJobsAccess like we do for some of the other conditions
```
diff --git a/awx/main/access.py b/awx/main/access.py
index f5eda98ad4..a999b4be63 100644
--- a/awx/main/access.py
+++ b/awx/main/access.py
@@ -2500,11 +2500,13 @@ class UnifiedJobAccess(BaseAccess):
     def filtered_queryset(self):
         inv_pk_qs = Inventory._accessible_pk_qs(Inventory, self.user, 'read_role')
         org_auditor_qs = Organization.objects.filter(Q(admin_role__members=self.user) | Q(auditor_role__members=self.user))
+        self_bulk_job_ids = WorkflowJob.objects.filter(Q(created_by_id=self.user.id, is_bulk_job=True)).values_list('id', flat=True)
         qs = self.model.objects.filter(
             Q(unified_job_template_id__in=UnifiedJobTemplate.accessible_pk_qs(self.user, 'read_role'))
             | Q(inventoryupdate__inventory_source__inventory__id__in=inv_pk_qs)
             | Q(adhoccommand__inventory__id__in=inv_pk_qs)
             | Q(organization__in=org_auditor_qs)
+            | Q(id__in=self_bulk_job_ids)
         )
         return qs
```

AlanCoding: Let me show you what that will look like when it generates the SQL. Persona is an org admin here.

https://gist.github.com/AlanCoding/9cacdb6e5770811194992ccf88b573bd

What does your proposed change add? The following, in sort of the main list of OR clauses.

```sql
        OR "main_unifiedjob"."id" IN
          (SELECT U0."unifiedjob_ptr_id"
           FROM "main_workflowjob" U0
           INNER JOIN "main_unifiedjob" U1 ON (U0."unifiedjob_ptr_id" = U1."id")
           WHERE (U1."created_by_id" = 1097
                  AND U0."is_bulk_job")))
```

Big picture - avoiding queries like this is _why_ the RBAC system was created in the first place. These quickly become unmaintainable, and people will lose the ability to use the app due to complex permissions and a large number of objects.

This isn't a linear cost, either. Every new condition is tremendously risky.

We shouldn't be doing these conditions at all. There should be _one_ filter implemented by the backend to filter with roles, plus what the UI filters by. We should not have added the workflow approval condition, and that doesn't justify adding a new condition now.

kdelee: ok talked to @AlanCoding and we came up with https://github.com/jainnikhil30/awx/pull/15 as a resolution to these problems

awx/api/serializers.py
kdelee: included in https://github.com/jainnikhil30/awx/pull/16 thanks @AlanCoding

awx/main/access.py
kdelee: we addressed changing the WorkflowJobWorkflowNodesList view  in https://github.com/ansible/awx/pull/13462/commits/03a9031f81e37f5e9c87e7a01835d345a9a059f8

kdelee (approved):
Amazing work. Great collaboration @AlanCoding @jainnikhil30 @fosterseth

awx/api/serializers.py
```diff
@@ -4405,6 +4533,271 @@ def validate(self, attrs):
...
+        wfj.signal_start()
+
+        return WorkflowJobSerializer().to_representation(wfj)
+
+    def check_organization_permission(self, attrs, request):
+        # validate Organization
+        # - If the orgs is not set, set it to the org of the launching user
+        # - If the user is part of multiple orgs, throw a validation error saying user is part of multiple orgs, please provide one
+        if not request.user.is_superuser:
+            read_org_qs = Organization.accessible_objects(request.user, 'read_role')
```
```suggestion
            read_org_qs = Organization.accessible_objects(request.user, 'member_role')
```

fosterseth (approved):

docs/bulk_api.md
```diff
@@ -0,0 +1,78 @@
...
```
kdelee, line 5:
```suggestion
- /api/v2/bulk/host_create

Making individual API calls in rapid succession or at high concurrency can overwhelm AWX's ability to serve web requests. When the application's ability to serve is exausted, clients often receive 504 timeout errors.

 Allowing the client combine actions into fewer requests allows for launching more jobs or adding more hosts with fewer requests and less time without exauhsting Controller's ability to serve requests, making excessive and repetitive database queries, or using excessive database connections (each web request opens a seperate database connection).
```

docs/bulk_api.md
kdelee: addressed in suggestion

awx/api/serializers.py
```diff
@@ -4405,6 +4533,271 @@ def validate(self, attrs):
...
+        requested_use_credentials = set()
+        requested_use_labels = set()
+        # requested_use_instance_groups = set()
+        for job in attrs['jobs']:
+            for cred in job.get('credentials', []):
+                requested_use_credentials.add(cred)
+            for label in job.get('labels', []):
+                requested_use_labels.add(label)
+            # for instance_group in job.get('instance_groups', []):
+            #     requested_use_instance_groups.add(instance_group)
```
AlanCoding, line 4619: We should probably call out _why_ we're not doing instance_groups here.

Also add a mention to the docs, because otherwise this is inconsistent behavior.

docs/bulk_api.md
```diff
@@ -0,0 +1,78 @@
...
+        ],
+        "inventory": 2
+    }
+
+In the above example, `inventory: 2` will get used for the job templates (11, 12 and 13) in which inventory is marked as prompt of launch.
+
+### RBAC For Bulk Job Launch
+
+#### Who can bulk launch?
+Anyone who is logged in can view the launch point. In order to launch a unified_job_template, you need to have either `update` or `execute` depending on the type of unified job (job template, project update, etc).
```
kdelee, line 62:
```suggestion
Anyone who is logged in can view the launch point. In order to launch a unified_job_template, you need to have either `update` or `execute` depending on the type of unified job (job template, project update, etc).

Launching using the bulk endpoint results in a workflow job being launched. For auditing purposes, in general we require to assign an organization to the resulting workflow. The logic for assigning this organization is as follows:

- Superusers may assign any organization or none. If they do not assign one, they will be the only user able to see the parent workflow.
- Users that are members of exactly 1 organization do not need to specify an organization, as their single organization will be used to assign to the resulting Workflow
- Users that are members of multiple organizations must specify the organization to assign to the resulting workflow. If they do not specify, an error will be returned indicating this requirement.

Example of specifying the organization:

    {
        "name": "Bulk Job Launch with org specified",
        "jobs": [
            {"unified_job_template": 12},
            {"unified_job_template": 13}
        ],
        "organization": 2
    }
```

awx/api/serializers.py
```diff
@@ -4405,6 +4533,271 @@ def validate(self, attrs):
...
+            nodes.append(node_obj)
+
+            # we'll need this later when we do the m2m through model bulk create
+            node_m2m_objects[node_attrs['identifier']]['node'] = node_obj
+
+        WorkflowJobNode.objects.bulk_create(nodes)
+
+        # Deal with the m2m objects we have to create once the node exists
+        for obj_type, obj_through_model in node_m2m_object_types_to_through_model.items():
+            through_models = []
```
AlanCoding: nit - this isn't a list of through _models_. It's a list of through model _objects_.

awx/api/serializers.py
```diff
@@ -4405,6 +4533,271 @@ def validate(self, attrs):
...
+            for node_identifier in node_m2m_objects.keys():
+                if obj_type in node_m2m_objects[node_identifier] and obj_type == 'credentials':
+                    for cred in node_m2m_objects[node_identifier][obj_type]:
+                        through_models.append(obj_through_model(credential=cred, workflowjobnode=node_m2m_objects[node_identifier]['node']))
+                if obj_type in node_m2m_objects[node_identifier] and obj_type == 'labels':
+                    for label in node_m2m_objects[node_identifier][obj_type]:
+                        through_models.append(obj_through_model(label=label, workflowjobnode=node_m2m_objects[node_identifier]['node']))
+                # if obj_type in node_m2m_objects[node_identifier] and obj_type == 'instance_groups':
+                #     for instance_group in node_m2m_objects[node_identifier][obj_type]:
+                #         through_models.append(obj_through_model(instancegroup=instance_group, workflowjobnode=node_m2m_objects[node_identifier]['node']))
```
AlanCoding: I think I can see through to how you _could_ implement `instance_groups`, which would be that the through model would have to have a `position` field specified, and you would have to be sure this reflected the position in the list the user provided via the API. Fine to not do that here, but I also don't like having commented code that wouldn't fully work, so I'd advocate deleting this for now.

awx/main/access.py
```diff
@@ -1997,7 +1998,15 @@ class WorkflowJobNodeAccess(BaseAccess):
...
-        return self.model.objects.filter(workflow_job__unified_job_template__in=UnifiedJobTemplate.accessible_pk_qs(self.user, 'read_role'))
+        return self.model.objects.filter(
+            Q(workflow_job__unified_job_template__in=UnifiedJobTemplate.accessible_pk_qs(self.user, 'read_role'))
+            | Q(workflow_job__organization__in=Organization.objects.filter(Q(admin_role__members=self.user)))
+        )
+
+    def can_read(self, obj):
+        if obj.workflow_job.is_bulk_job and obj.workflow_job.created_by_id == self.user.id:
+            return True
+        return super().can_read(obj)
```
AlanCoding, line 2010: non-blocking comment, but we should add a comment here, calling out that this is intentionally inconsistent with `filtered_queryset` for performance reasons, and that this is documented.

Same for the other `can_read` override.

awx/main/access.py
```diff
@@ -1997,7 +1998,15 @@ class WorkflowJobNodeAccess(BaseAccess):
     )
 
     def filtered_queryset(self):
-        return self.model.objects.filter(workflow_job__unified_job_template__in=UnifiedJobTemplate.accessible_pk_qs(self.user, 'read_role'))
+        return self.model.objects.filter(
+            Q(workflow_job__unified_job_template__in=UnifiedJobTemplate.accessible_pk_qs(self.user, 'read_role'))
+            | Q(workflow_job__organization__in=Organization.objects.filter(Q(admin_role__members=self.user)))
+        )
+
+    def can_read(self, obj):
```
kdelee, line 2006:
```suggestion
    def can_read(self, obj):
        """Overriding this opens up detail view access for bulk jobs, where the workflow job has no associated workflow job template."""
```

awx/main/access.py
```diff
@@ -2123,7 +2132,15 @@ class WorkflowJobAccess(BaseAccess):
     )
 
     def filtered_queryset(self):
-        return WorkflowJob.objects.filter(unified_job_template__in=UnifiedJobTemplate.accessible_pk_qs(self.user, 'read_role'))
+        return WorkflowJob.objects.filter(
+            Q(unified_job_template__in=UnifiedJobTemplate.accessible_pk_qs(self.user, 'read_role'))
+            | Q(organization__in=Organization.objects.filter(Q(admin_role__members=self.user)), is_bulk_job=True)
+        )
+
+    def can_read(self, obj):
```
kdelee, line 2141:
```suggestion
    def can_read(self, obj):
        """Overriding this opens up detail view access for bulk jobs, where the workflow job has no associated workflow job template."""
```

docs/bulk_api.md
```diff
@@ -0,0 +1,78 @@
...
+        "name": "Bulk Job Launch",
+        "jobs": [
+            {"unified_job_template": 11, "limit": "kansas", "credentials": [1]},
+            {"unified_job_template": 12},
+            {"unified_job_template": 13}
+        ],
+        "inventory": 2
+    }
+
+In the above example, `inventory: 2` will get used for the job templates (11, 12 and 13) in which inventory is marked as prompt of launch.
```
AlanCoding, line 55:
```suggestion
In the above example, `inventory: 2` will get used for the job templates (11, 12 and 13) in which inventory is marked as prompt of launch.

*Note:* The `instance_groups` relationship is not supported for node-level prompts, unlike `"credentials"` in the above example, and will be ignored if provided. See OPTIONS for `/api/v2/bulk/job_launch/` for what fields are accepted at the workflow and node level, as that is the ultimate source of truth to determine what fields the API will accept.
```

awxkit/awxkit/api/resources.py
fosterseth: Yes not needed -- we could add it as a stub but no custom logic was needed for host create endpoint, so it just works with whatever generic logic exists in awxkit to handle arbitrary endpoints.

awx/main/access.py
kdelee: added docstring

awx_collection/README.md
```diff
@@ -68,6 +68,7 @@ Notable releases of the `awx.awx` collection:
  - 7.0.0 is intended to be identical to the content prior to the migration, aside from changes necessary to function as a collection.
  - 11.0.0 has no non-deprecated modules that depend on the deprecated `tower-cli` [PyPI](https://pypi.org/project/ansible-tower-cli/).
  - 19.2.1 large renaming purged "tower" names (like options and module names), adding redirects for old names
+ - 21.11.0 "tower" modules deprecated and symlinks removed.
```
AlanCoding: this line looks like it hitched a ride on your PR.

awx_collection/plugins/modules/bulk_host_create.py
AlanCoding: Is there any way that a bulk action _doesn't_ change something? @kdelee I assumed that a no-op POST to the bulk endpoints wasn't possible (would get conflict error) so a 201 is always changed.

kdelee: yep either create and return 201, or reject do nothing 400

awx/api/serializers.py
```diff
@@ -4405,6 +4533,271 @@ def validate(self, attrs):
...
+    unified_job_template = serializers.IntegerField(
+        required=True, min_value=1, help_text=_('Primary key of the template for this job, can be a job template or inventory source.')
+    )
+    inventory = serializers.IntegerField(required=False, min_value=1)
+    execution_environment = serializers.IntegerField(required=False, min_value=1)
+    # many-to-many fields
+    credentials = serializers.ListField(child=serializers.IntegerField(min_value=1), required=False)
+    labels = serializers.ListField(child=serializers.IntegerField(min_value=1), required=False)
+    # TODO: Use instance group role added via PR 13584(once merged), for now everything related to instance group is commented
+    # instance_groups = serializers.ListField(child=serializers.IntegerField(min_value=1), required=False)
```
AlanCoding, line 4548: I see, very nice reference to put in, FYI @gamuniz

AlanCoding (approved):

Issue #13447, opened 2023-01-18
[constructed-inventory] Add new API views for constructed inventory edits


AlanCoding:
Proposal is to add endpoints:

 - `/api/v2/constructed_inventories/100/` - as a superset of `/api/v2/inventories/100/`
 - `/api/v2/constructed_inventories/` - as a filtering of `/api/v2/inventories/` and possibly sharing the serializer with the first point, unsure

The detail view here is the main thing needed in order to un-block UI work, so I would advocating getting that into the feature branch first.

The view will append on InventorySource fields to what is normally shown by the InventorySerializer. Any inventory source fields will be applied to the related "constructed" inventory source on the backend, to simplify things for clients.

PR #13463, opened 2023-01-24
[constructed-inventory] Add views and serializers for special constructed inventory endpoints

AlanCoding:
##### SUMMARY
This adds special endpoints for constructed inventory, as outlined in https://github.com/ansible/awx/issues/13447 so that the UI can work with it as a single form.

I come into this with some opinions, so all the special logic is generally implemented in the serializer.

We need to add validation (or some other read_only status to the field) so that `kind` cannot be given any value other than "constructed" in these views, because that would not make any sense. I do not intend to do that here, I just mention it because it becomes another item in the long list of validation cases we need to add for this feature.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -1715,6 +1715,7 @@ class InventorySerializer(LabelsListMixin, BaseSerializerWithVariables):
             res['organization'] = self.reverse('api:organization_detail', kwargs={'pk': obj.organization.pk})
         if obj.kind == 'constructed':
             res['source_inventories'] = self.reverse('api:inventory_source_inventories', kwargs={'pk': obj.pk})
+            res['url'] = self.reverse('api:constructed_inventory_detail', kwargs={'pk': obj.pk})
         return res
 
     def to_representation(self, obj):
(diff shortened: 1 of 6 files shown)


TheRealHaoLiu:
let me make sure i understand the code and the goal correctly 

for a Inventory of type `constructed`
- first inventory source is the "constructed inventory source" (essentially https://docs.ansible.com/ansible/latest/collections/ansible/builtin/constructed_inventory.html)

the constructed inventory source view/serializer will do the following
- on GET take the information out of the "constructed inventory source" and add it as fields in the view
- on CREATE and UPDATE strip the information that belongs to "constructed inventory source" and create the "constructed inventory source" as well as the Inventory that is of type "constructed"

the URL field that's added to Inventory of type "constructed" is to provide a linkage to the "constructed inventory" view

@AlanCoding am i understanding this correctly?

awx/api/serializers.py
```diff
@@ -1715,6 +1715,7 @@ def get_related(self, obj):
             res['organization'] = self.reverse('api:organization_detail', kwargs={'pk': obj.organization.pk})
         if obj.kind == 'constructed':
             res['source_inventories'] = self.reverse('api:inventory_source_inventories', kwargs={'pk': obj.pk})
+            res['url'] = self.reverse('api:constructed_inventory_detail', kwargs={'pk': obj.pk})
```
TheRealHaoLiu, line 1718: is `url` descriptive enough?

AlanCoding: it's a pre-existing field, I'm swapping the `/api/v2/inventories/42/` link with `/api/v2/constructed_inventories/42/` link. Somewhat opinionated, either would work, but I figured this is what we're going for.

AlanCoding:
> on CREATE and UPDATE strip the information that belongs to "constructed inventory source" and create the "constructed inventory source" as well as the Inventory that is of type "constructed"

Before this PR, we added logic to create the constructed inventory source automatically _in the models_. So that is not part of this PR, and this PR is only changing the views.

Otherwise, yes, on creation, we fetch the auto-created inventory source and update it with any of the added fields in the serializer.

TheRealHaoLiu (approved):

gamuniz (approved):

Issue #13469, opened 2023-01-25
Job failed with just error and without log output


Halytskyi:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

Clean deployment and I use "Demo Job Template". Job failing and no output except "Finished":
![Screenshot 2023-01-24 at 4 39 47 PM](https://user-images.githubusercontent.com/4974260/214453907-bdff6c0f-5c6a-46ee-8316-7b9e307e6f49.png)
![Screenshot 2023-01-24 at 4 40 03 PM](https://user-images.githubusercontent.com/4974260/214453935-986c6be5-26fe-4e7a-b069-1b78f90a3f00.png)

In "**awx-ee**" container logs I see:
```
DEBUG 2023/01/24 23:27:28 Client connected to control service @
DEBUG 2023/01/24 23:27:28 Control service closed
DEBUG 2023/01/24 23:27:28 Client disconnected from control service @
DEBUG 2023/01/24 23:27:28 Client connected to control service @
DEBUG 2023/01/24 23:27:29 Kubernetes version v1.23.15 is at least v1.23.14, using reconnect support
DEBUG 2023/01/24 23:27:29 [4Pmqfss9] streaming stdout with reconnect support
DEBUG 2023/01/24 23:27:29 [4Pmqfss9] Detected EOF for pod my-ns/automation-job-6-njfz8. Will retry 5 more times.
DEBUG 2023/01/24 23:27:29 [4Pmqfss9] Detected EOF for pod my-ns/automation-job-6-njfz8. Will retry 4 more times.
DEBUG 2023/01/24 23:27:29 [4Pmqfss9] Detected EOF for pod my-ns/automation-job-6-njfz8. Will retry 3 more times.
DEBUG 2023/01/24 23:27:29 [4Pmqfss9] Detected EOF for pod my-ns/automation-job-6-njfz8. Will retry 2 more times.
DEBUG 2023/01/24 23:27:29 [4Pmqfss9] Detected EOF for pod my-ns/automation-job-6-njfz8. Will retry 1 more times.
[...]

PR #13482, opened 2023-01-27
Consume `job_explanation` from runner, fix error reporting error

AlanCoding:
##### SUMMARY
An update of https://github.com/ansible/awx/pull/12089, as this failure in error handling has been reflected in recent cases, and in user reports. This seems to account for the "Job terminated due to error" reports.

This modifies some of what was introduced in https://github.com/ansible/awx/pull/12961, which is the source of the bug.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/tasks/receptor.py
+++ b/awx/main/tasks/receptor.py
@@ -431,16 +431,16 @@ class AWXReceptorJob:
                         # massive, only ask for last 1000 bytes
                         startpos = max(stdout_size - 1000, 0)
                         resultsock, resultfile = receptor_ctl.get_work_results(self.unit_id, startpos=startpos, return_socket=True, return_sockfile=True)
-                        resultsock.setblocking(False)  # this makes resultfile reads non blocking
                         lines = resultfile.readlines()
                         receptor_output = b"".join(lines).decode()
                     if receptor_output:
-                        self.task.runner_callback.delay_update(result_traceback=receptor_output)
+                        self.task.runner_callback.delay_update(result_traceback=f'Worker output:\n{receptor_output}')
                     elif detail:
-                        self.task.runner_callback.delay_update(result_traceback=detail)
+                        self.task.runner_callback.delay_update(result_traceback=f'Receptor detail:\n{detail}')
                     else:
                         logger.warning(f'No result details or output from {self.task.instance.log_format}, status:\n{state_name}')
                 except Exception:
+                    logger.exception(f'Work results error from job id={self.task.instance.id} work_unit={self.task.instance.work_unit_id}')
                     raise RuntimeError(detail)
 
         return res
--- a/awx/main/tasks/callback.py
+++ b/awx/main/tasks/callback.py
@@ -208,9 +208,10 @@ class RunnerCallback:
             # We opened a connection just for that save, close it here now
             connections.close_all()
         elif status_data['status'] == 'error':
-            result_traceback = status_data.get('result_traceback', None)
-            if result_traceback:
-                self.delay_update(result_traceback=result_traceback)
+            for field_name in ('result_traceback', 'job_explanation'):
+                field_value = status_data.get(field_name, None)
+                if field_value:
+                    self.delay_update(**{field_name: field_value})
 
     def artifacts_handler(self, artifact_dir):
         self.artifacts_processed = True


AlanCoding:
This should go in along-side of https://github.com/ansible/ansible-runner/pull/1186

fosterseth (approved):

AlanCoding:
What I originally tested this against was the runner change:

```diff
@@ -68,6 +73,7 @@ class Worker(object):
             _output = sys.stdout.buffer
         self._input = _input
         self._output = _output
+        self._output.write(b'surprise from alan!\n')
 
         self.kwargs = kwargs
         self.job_kwargs = None
```

But I've realized there is a shortcoming for a different class of errors. Consider:

```diff
diff --git a/ansible_runner/streaming.py b/ansible_runner/streaming.py
index 608b972..df14bfc 100644
--- a/ansible_runner/streaming.py
+++ b/ansible_runner/streaming.py
@@ -68,6 +68,7 @@ class Worker(object):
             _output = sys.stdout.buffer
         self._input = _input
         self._output = _output
+        raise Exception('alan!')
 
         self.kwargs = kwargs
         self.job_kwargs = None
```

This results in a bad user experience:

Screenshot from 2023-01-31 13-48-16

But! As per some speculative data related to https://github.com/ansible/awx/issues/13469, we know that receptor may not be fully functional. So we might need the `readline` data in some cases, and the full results (or truncated) results in other cases.

As I've stewed on this problem, I don't see a way out other than a special-case for the JSON parse error from the processing code. So I think I have an idea of how to move forward - which involves combining all of the data we can possibly manage, and a special string comparison for that specific message.

AlanCoding:
Error handling is still very broken without https://github.com/ansible/awx/issues/13494 fixed, so I'm not sure what I'm going to do now.

AlanCoding:
But I did adjust my ansible-runner PR to gracefully handle the Exception case I mentioned before. I think both of these are good changes, but it's still hard to get quality output from error cases.

awx/main/tasks/receptor.py
```diff
@@ -437,9 +437,9 @@ def _run_internal(self, receptor_ctl):
                         lines = resultfile.readlines()
                         receptor_output = b"".join(lines).decode()
                     if receptor_output:
-                        self.task.runner_callback.delay_update(result_traceback=receptor_output)
+                        self.task.runner_callback.delay_update(result_traceback=f'Worker output:\n{receptor_output}')
```
AlanCoding, line 437: My plan here, now, is that ansible-runner will write the line to `job_explanation`. By doing that, we _might_ echo the output twice in two places. Because of that, we need some better labeling of what data we're showing, so I added this stuff.

awx/main/tasks/receptor.py
```diff
@@ -437,9 +437,9 @@ def _run_internal(self, receptor_ctl):
                         lines = resultfile.readlines()
```
AlanCoding, line 434: recent discussion is that I might delete `resultsock.setblocking(False)`

AlanCoding:
I added tests to my related ansible-runner PR. This is one of the highest priority items I have to make sure we get in, but I don't want to merge it until the runner change lands (because don't know what the UX of that would be).

AlanCoding:
I have test cases up here:

https://github.com/AlanCoding/bad-execution-environments

Let me review those...

The "traceback" test is not tested right now due to linked receptor issue https://github.com/ansible/receptor/issues/736

#### ending_line

Screenshot from 2023-04-03 12-29-19

#### starting_line

Screenshot from 2023-04-03 12-29-09

#### artifacts

Screenshot from 2023-04-03 12-26-59

awx/main/tasks/receptor.py
```diff
@@ -431,16 +431,16 @@ def _run_internal(self, receptor_ctl):
...
                     if receptor_output:
-                        self.task.runner_callback.delay_update(result_traceback=receptor_output)
+                        self.task.runner_callback.delay_update(result_traceback=f'Worker output:\n{receptor_output}')
                     elif detail:
-                        self.task.runner_callback.delay_update(result_traceback=detail)
+                        self.task.runner_callback.delay_update(result_traceback=f'Receptor detail:\n{detail}')
                     else:
                         logger.warning(f'No result details or output from {self.task.instance.log_format}, status:\n{state_name}')
                 except Exception:
+                    logger.exception(f'Work results error from job id={self.task.instance.id} work_unit={self.task.instance.work_unit_id}')
```
gamuniz, line 443: Does receptor have the ability to expose the work_unit_id from the remote(execution_node)? It might be nice to expose that someplace for further debugging

AlanCoding: This came up before, and I agree but it will have a few difficult implementation questions. Best to raise it as a new issue, if it's not already, I can't find it.

gamuniz: Filed this issue to track that separately:

https://github.com/ansible/awx/issues/13793

gamuniz (approved):

PR #13487, opened 2023-01-30
[constructed-inventory] Add constructed inventory docs and do minor field updates

AlanCoding:
##### SUMMARY
I started trying to do a demo, and from that realized that these changes are necessary.

I consider these to be fairly bare-bones docs. We need considerably more usage guidelines, with functional examples for using Ansible facts, and the metavars as mentioned. That should be tracked in cards, but I want to get this in to have a file structure pinned down.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION
I changed `update_on_launch`, because that was the original intent, and I called it out in the docs.

I also added the `verbosity` field, because this _should_ be editable for constructed inventory. I knew we would have more fields trickle in like this.


docs/inventory/constructed_inventory.md
```diff
@@ -0,0 +1,86 @@
+### Constructed inventory in AWX
+
+Constructed inventory is a separate "kind" of inventory, along-side of
+normal (manual) inventories and "smart" inventories.
+The functionality overlaps with smart inventory, and it is intended that
+smart inventory is sunseted and will be eventually removed.
```
gamuniz: sunsetted?

gamuniz (approved):

TheRealHaoLiu (approved):

PR #13490, opened 2023-01-30
Update "one or more" fields in SAML documentation.

john-westcott-iv:
According to latest documentation, role and value are now "one or more" fields. So they both need to be arrays.  Entering the json data as you have in this article doesn't work. But when I added the brackets, it then worked.   Thank you

Replaces https://github.com/ansible/awx/pull/13355

##### ISSUE TYPE

##### COMPONENT NAME

 - Docs

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```


tallyoh:
Thank you!

fosterseth (approved):

PR #13491, opened 2023-01-31
External users should not be able to change their password

rooftopcellist:
##### SUMMARY

When Azure AD (or other external source) is configured, a user can log in with their external credentials and then set a password on their user screen. Once that is set, the user can auth with the local password into AWX. That user should NOT be able to change their password via the AWX API/UI as authorization for that user is managed by Azure AD.  

In the UI, we can see that when editing that user, currently the password field is shown and can be edited when specific combinations of external authentication are configured. For example, if SAML is configured with Azure AD than the password is not editable. However, if just Azure AD is configured the password is editable.

This is happening because `external_account` for that user is being set to `null`, which is because this is evaluating false:

https://github.com/ansible/awx/blob/ac6a82eee41feb041ff3e4d16459d4b1a774175f/awx/main/utils/common.py#L1043-L1050

With this change, it will evaluate to true, which will set `external_account: "social"`

```
>>> bool(getattr(settings, 'SOCIAL_AUTH_AZUREAD_OAUTH2_KEY', None))
True

>>> bool(user.social_auth.all())
True
```

In addition, we are changing the requirements from looking at the current configuration of AWX to just looking at the existing user attributes. Including the current configuration was causing another issue where a user could login through SAML but later be able to set a local password if SAML is disabled. While this has a byproduct of allowing users to be converted to local if the auth configuration is removed 1) it only allows this for authenticated users. 2) If SAML is reconfigured the account can bypass SAML login by using the local password.

##### ISSUE TYPE

##### COMPONENT NAME

 - API

cc @john-westcott-iv

### DIFF
--- a/awx/api/conf.py
+++ b/awx/api/conf.py
@@ -1,5 +1,4 @@
 # Django
-from django.conf import settings
 from django.utils.translation import gettext_lazy as _
 
 # Django REST Framework
@@ -9,6 +8,7 @@ from rest_framework import serializers
 from awx.conf import fields, register, register_validate
 from awx.api.fields import OAuth2ProviderField
 from oauth2_provider.settings import oauth2_settings
+from awx.sso.common import is_remote_auth_enabled
 
 
 register(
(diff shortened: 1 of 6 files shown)


gamuniz (approved):

rooftopcellist:
@john-westcott-iv This PR looks good to me.  Since more of the changes in here are yours now, please merge it when you feel it is ready.  Thank you!

john-westcott-iv:
Testing this change requires a couple of different items.

1. Setup, one at a time, each authentication adapter (SAML, OIDC, LDAP, etc) and then login as a user through the adapter and verify:
    - That you do not see the password field on the users screen in the UI. This can also be confirmed by ensuring that the `/api/v2/user/<id>/` page has a non-null value for `external_account` and that there is something in the `auth` array. 
    - Confirm that if you pass the `password` field through the same API endpoint in `a` that the password does not take affect. Note, the API will not complain, you have to log out as the user and then try and login with the username/password and confirm that you are denied access.
    - Confirm that you can not get an oauth token for the user if 1ALLOW_OAUTH2_FOR_EXTERNAL_USERS1 is `False`
    - Confirm that you can get an oauth token for the user if `ALLOW_OAUTH2_FOR_EXTERNAL_USERS` is `True`
2. Disable all external authenticate and attempt to set `DISABLE_LOCAL_AUTH` to `True` and ensure that you see the exception `There are no remote authentication systems configured.`

jay-steurer (approved):

jay-steurer (approved):

Klaas-:
I think you failed to account for a scenario where a user already has a password an is then being transformed into a social account. I pre-create a few users with a random password via awx.awx.user so I can give them superuser rights during install. Those are then transformed to social auth on the users first login. I still see in DB that the password is set (pbkdf2_....) and I can still login with that password. I can however not change it anymore via UI or API.

Shall I open a new issue for this?

john-westcott-iv:
@Klaas- Go ahead and open an issue for this and we can discuss it in the new issue.

Klaas-:
done, #13662

PR #13502, opened 2023-02-01
Update to comply with new black rules

AlanCoding:
##### SUMMARY
Failing checks on other PRs

I ran `black awx awxkit`

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/api/metadata.py
+++ b/awx/api/metadata.py
@@ -128,7 +128,7 @@ class Metadata(metadata.SimpleMetadata):
         # Special handling of notification configuration where the required properties
         # are conditional on the type selected.
         if field.field_name == 'notification_configuration':
-            for (notification_type_name, notification_tr_name, notification_type_class) in NotificationTemplate.NOTIFICATION_TYPES:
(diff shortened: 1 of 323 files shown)


awx/api/metadata.py
```diff
@@ -128,7 +128,7 @@ def get_field_info(self, field):
         # Special handling of notification configuration where the required properties
         # are conditional on the type selected.
         if field.field_name == 'notification_configuration':
-            for (notification_type_name, notification_tr_name, notification_type_class) in NotificationTemplate.NOTIFICATION_TYPES:
+            for notification_type_name, notification_tr_name, notification_type_class in NotificationTemplate.NOTIFICATION_TYPES:
```
AlanCoding, line 131: this makes me uncomfortable

gamuniz (approved):

awx/api/metadata.py
AlanCoding: Never mind, I think I've convinced myself that since this is a `for`, it really is standard practice to not use (). The behavior is the same in either case.

ryanpetrello:
SHIP IT

PR #13507, opened 2023-02-01
fix postgreSQL reconnection in case of DB failover or idle or instable connection

### LINKED ISSUE #13505: PostgreSQL reconnection in case of DB failover or idle or unstable connection

### Bug Summary

Hi team,
as discussed on matrix with @TheRealHaoLiu and @shanemcd  I notice that in case of instable connection with PostgreSQL or failover (in case of PostgreSQL HA configuration) awx.main.dispatcher lose the connection with DB and not reconnect it.

I tired multiple use case, but the actual reconnection will work only in a very short dropdown connection (less then 1 sec) or not everytime.

I notice that similar issue was already opened but closed with a wron motivation https://github.com/ansible/awx/issues/12683 . 

I opened a PR to try to fix this issue with a simple check of connection with some retry on pg_bus_conn function (as default 40 times every 4 seconds).
If connection is restored we need to kill dispatcher process on awx-task container to restart it.

I don't know if different approach it's possibile but in that way we are sure that all process restart correctly.

From UI perspective or awx-web logs I don't see any drop, everything works whell.

### AWX version

21.10.2

- [ ] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [X] Other

### Installation method

kubernetes

### Modifications

no

### Ansible version

_No response_

### Operating system

_No response_

### Web browser

_No response_

### Steps to reproduce

Use AWX normally and then try to restart or deattach the network to postgreSQL and check awx-task log.

Connection will be droppend and all jobs will be in pending status.

### Expected results

In case of PostgreSQL drop connection , reconnection should be re-established

### Actual results

Error on awx-task

### Additional information

_No response_

tanganellilore:
##### SUMMARY
I notice that in case of instable connection with PostgreSQL or failover (in case of PostgreSQL HA configuration) awx.main.dispatcher lose the connection with DB and not reconnect it.

I opened a PR to try to fix this issue with a simple check of connection with some retry on pg_bus_conn function (as default 40 times every 4 seconds, as per default).
If connection is restored we need to kill dispatcher process on awx-task container and wait to restart it.

I don't know if different approach it's possibile but in that way we are sure that all process restart correctly.

##### ISSUE TYPE

##### COMPONENT NAME
 - Other

##### AWX VERSION

```
AWX 20.10.2
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/dispatch/__init__.py
+++ b/awx/main/dispatch/__init__.py
@@ -1,16 +1,10 @@
 import psycopg2
 import select
-import logging
-import time
-import os
-import signal
 
 from contextlib import contextmanager
 
 from django.conf import settings
-from django.db import connection as pg_connection, OperationalError, Error
-
-logger = logging.getLogger('awx.dispatch.init')
+from django.db import connection as pg_connection
 
 NOT_READY = ([], [], [])
 
@@ -42,6 +36,10 @@ class PubSub(object):
         while True:
             if select.select([self.conn], [], [], select_timeout) == NOT_READY:
                 if yield_timeouts:
+                    # this is a 'ping' to db connection to make sure it's still alive
+                    # and prevent the main thread from hanging
+                    # this error will be cathced by the main thread and we enter on reconnection loop
+                    self.conn.cursor().execute('SELECT 1')
                     yield None
             else:
                 self.conn.poll()
@@ -63,11 +61,7 @@ def pg_bus_conn(new_connection=False):
     so that messages follow postgres transaction rules
     https://www.postgresql.org/docs/current/sql-notify.html
     '''
-    conn = None
-    retry_conn = False
-    MAX_RETRIES = settings.DISPATCHER_DB_DOWNTOWN_TOLLERANCE
-    POLL_SECONDS = int(settings.DISPATCHER_DB_DOWNTOWN_TOLLERANCE / 10)
-    
+
     if new_connection:
         conf = settings.DATABASES['default']
         conn = psycopg2.connect(
@@ -77,23 +71,10 @@ def pg_bus_conn(new_connection=False):
         conn.set_session(autocommit=True)
     else:
         if pg_connection.connection is None:
-            for retry_count in range(MAX_RETRIES):
-                try:
-                    pg_connection.connect()
-                    conn = pg_connection.connection
-                except (psycopg2.OperationalError, OperationalError, Error) as reconn_error:
-                    logger.warning(f'Database unavailable. Retry with new connection in next {POLL_SECONDS} seconds. Attempt {retry_count}/{MAX_RETRIES}.')
-                    time.sleep(POLL_SECONDS)
-                    retry_conn = True
-                else: 
-                    if retry_conn:
-                        logger.warning('Run dispatcher restart due to database connection loss and restore.')
-                        os.kill(os.getppid(), signal.SIGTERM)
-                    break
-            else:
-                raise RuntimeError(f'Could not connect to postgres afer {MAX_RETRIES} retries.')
-        else:
-            conn = pg_connection.connection
+            pg_connection.connect()
+        if pg_connection.connection is None:
+            raise RuntimeError('Unexpectedly could not connect to postgres for pg_notify actions')
+        conn = pg_connection.connection
 
     pubsub = PubSub(conn)
     yield pubsub
--- a/awx/main/dispatch/__init__.py
+++ b/awx/main/dispatch/__init__.py
@@ -1,11 +1,16 @@
 import psycopg2
 import select
+import logging
+import time
+import os
+import signal
 
 from contextlib import contextmanager
 
 from django.conf import settings
-from django.db import connection as pg_connection
+from django.db import connection as pg_connection, OperationalError, Error
 
+logger = logging.getLogger('awx.dispatch.init')
 
 NOT_READY = ([], [], [])
 
@@ -58,7 +63,11 @@ def pg_bus_conn(new_connection=False):
     so that messages follow postgres transaction rules
     https://www.postgresql.org/docs/current/sql-notify.html
     '''
-
+    conn = None
+    retry_conn = False
+    MAX_RETRIES = settings.DISPATCHER_DB_DOWNTOWN_TOLLERANCE
+    POLL_SECONDS = int(settings.DISPATCHER_DB_DOWNTOWN_TOLLERANCE / 10)
+    
     if new_connection:
         conf = settings.DATABASES['default']
         conn = psycopg2.connect(
@@ -68,10 +77,23 @@ def pg_bus_conn(new_connection=False):
         conn.set_session(autocommit=True)
     else:
         if pg_connection.connection is None:
-            pg_connection.connect()
-        if pg_connection.connection is None:
-            raise RuntimeError('Unexpectedly could not connect to postgres for pg_notify actions')
-        conn = pg_connection.connection
+            for retry_count in range(MAX_RETRIES):
+                try:
+                    pg_connection.connect()
+                    conn = pg_connection.connection
+                except (psycopg2.OperationalError, OperationalError, Error) as reconn_error:
+                    logger.warning(f'Database unavailable. Retry with new connection in next {POLL_SECONDS} seconds. Attempt {retry_count}/{MAX_RETRIES}.')
+                    time.sleep(POLL_SECONDS)
+                    retry_conn = True
+                else: 
+                    if retry_conn:
+                        logger.warning('Run dispatcher restart due to database connection loss and restore.')
+                        os.kill(os.getppid(), signal.SIGTERM)
+                    break
+            else:
+                raise RuntimeError(f'Could not connect to postgres afer {MAX_RETRIES} retries.')
+        else:
+            conn = pg_connection.connection
 
     pubsub = PubSub(conn)
     yield pubsub
--- a/awx/main/dispatch/worker/base.py
+++ b/awx/main/dispatch/worker/base.py
@@ -190,7 +190,7 @@ class AWXConsumerPG(AWXConsumerBase):
             except psycopg2.InterfaceError:
                 logger.warning("Stale Postgres message bus connection, reconnecting")
                 continue
-            except (db.DatabaseError, psycopg2.OperationalError):
+            except (db.DatabaseError, psycopg2.OperationalError, db.Error):
                 # If we have attained stady state operation, tolerate short-term database hickups
                 if not self.pg_is_down:
                     logger.exception(f"Error consuming new events from postgres, will retry for {self.pg_max_wait} s")
--- a/awx/main/dispatch/worker/base.py
+++ b/awx/main/dispatch/worker/base.py
@@ -190,7 +190,7 @@ class AWXConsumerPG(AWXConsumerBase):
             except psycopg2.InterfaceError:
                 logger.warning("Stale Postgres message bus connection, reconnecting")
                 continue
-            except (db.DatabaseError, psycopg2.OperationalError, db.Error):
+            except (db.DatabaseError, psycopg2.OperationalError):
                 # If we have attained stady state operation, tolerate short-term database hickups
                 if not self.pg_is_down:
                     logger.exception(f"Error consuming new events from postgres, will retry for {self.pg_max_wait} s")


tanganellilore:
Issue: #13505

awx/main/dispatch/__init__.py
```diff
@@ -68,10 +77,23 @@ def pg_bus_conn(new_connection=False):
...
+                try:
+                    pg_connection.connect()
+                    conn = pg_connection.connection
+                except (psycopg2.OperationalError, OperationalError, Error) as reconn_error:
+                    logger.warning(f'Database unavailable. Retry with new connection in next {POLL_SECONDS} seconds. Attempt {retry_count}/{MAX_RETRIES}.')
+                    time.sleep(POLL_SECONDS)
+                    retry_conn = True
+                else: 
+                    if retry_conn:
+                        logger.warning('Run dispatcher restart due to database connection loss and restore.')
```
AlanCoding: logging nit-pick, this is used more broadly than the "dispatcher" for our task system. For the most part, this will apply to pg_notify listeners.

awx/main/dispatch/__init__.py
```diff
@@ -68,10 +77,23 @@ def pg_bus_conn(new_connection=False):
...
+                        os.kill(os.getppid(), signal.SIGTERM)
```
AlanCoding: Could you explain a little bit about what you're thinking here? I'm worried that it's not going to follow the canonical crash loop behavior, which will someday make something fail in the wrong way.

AlanCoding: If you need to break out of the dispatcher main loop here

https://github.com/ansible/awx/blob/0815f935ca359537c023efadf286671e31dc3374/awx/main/dispatch/worker/base.py#L175

You could do that by throwing _any_ exception that's not one of them that gets caught. You could throw a `RuntimeError` or a pure `Exception` and this will exit the dispatcher, and then the decision-making about what to do after that (like restarting) is punted to supervisor then systemctl / kubernetes.

awx/main/dispatch/__init__.py
tanganellilore: This logging only in case of connection fail.
This seems to happen only for the Main queue when postgres go down.

awx/main/dispatch/__init__.py
tanganellilore: I notice that error arrive on pg_connection.connect() and  i need to check if with raise flow arrive to base.py. Tomorrow I will try this option. 

What i can say now is that i tried to avoid the "kill" of PID with reause pg_connection, but seems that a down of postgres broke something on what was already in place in terms of connection and queue listener. 

Any suggestion?

tanganellilore:
@AlanCoding I used a different approacha fter a lot of testing and trying different use case.

I adopted the usage of "ping" in postgres with a `SELECT 1` in the event loop only when yield_timeouts is true.
In that way if in the main thread the `pg_bus_conn` setted in the frist loop of dispatcher will go in hang (due to nw change on postgres or connection timeout or somehting similar) with this simple ping we can go directly to the exception and go to the retry loop alrready inplace.

So simply add one line of code we can catch the issue :-)

Obivously, i take a "query" every 5 seconds in case of there is no notification to be processed but i think this is acceptable.

awx/main/dispatch/__init__.py
tanganellilore: changed approach

awx/main/dispatch/__init__.py
tanganellilore: changed approach

djyasin:
@tanganellilore 
Thank you for providing this additional information! We are still trying to identify exactly what is occurring here.

Could you provide us with additional information regarding the error you experienced and what you saw? Screenshots, additional logs, or any additional information you can provide would be very helpful. 

Thank you for your time and assistance with this matter!

AlanCoding:
Specifically, as you're adding another class of error to be caught, which would make it really really relevant to be able to see what tracebacks were produced in the failover events that you had.

tanganellilore:
Good morning @AlanCoding ,
I removed the `db.Error` except, due to the fact that is not required on this case.

This PR try to catch the reconnection error also for the main loop notify. 
Actually the main loop on ` /awx/main/dispatch/worker/base.py` line 177, will not catch the connection drop beacuse it's simply create a connection first time and then listen the queue only one time. So if the connection to postgres will drop, for example for a network lag or for a failover of postgres, nobody restart the main loop and reconnect to all queue.
Going to exception, we simply restart the main loop if the retry happens before `pg_max_wait` time.


Some screenshot of error:
---
<br>

### **Simulate a failover with deattach of vip address from Master postgreql instance with actual release:**
### **Failover started on 08:16:20**

_On awx-task logs:_
image

_On AWX UI all jobs are in pending:_
image

_Inside awx-task the awx-manage command:_

image
(all worker are in idle)

image
(error on wait the listener)

---

### **Simulate a failover with deattach of vip address from Master postgreql instance with this PR:**
### **Failover failover started on 08:33:40**

_On awx-task logs:_
image
(we going to exception)

_On AWX UI all jobs are in running because we have a reconnection at 08:34:15:_
image

_Inside awx-task the awx-manage command:_

image
(worker now run some command and sync)

image
(listener work)

AlanCoding:
Thanks for the reply. I really want to understand the mechanics of what you've seen here. If I go to `awx-manage shell_plus` and then run this:

```python
In [1]: from awx.main.dispatch import pg_bus_conn

In [2]: with pg_bus_conn(new_connection=True) as conn:
   ...:     for e in conn.events(yield_timeouts=True):
   ...:         print(e)
   ...: 
None
None
None
None
```

Then I will continue getting a `None` every 5 seconds as long as the database is healthy. Now if I essentially shut postgres down (pretty normally I think, it logs about terminating connections then "LOG:  shutting down") then it will stop printing those and give:

```
None
---------------------------------------------------------------------------
OperationalError                          Traceback (most recent call last)
Cell In[2], line 2
      1 with pg_bus_conn(new_connection=True) as conn:
----> 2     for e in conn.events(yield_timeouts=True):
      3         print(e)

File /awx_devel/awx/main/dispatch/__init__.py:42, in PubSub.events(self, select_timeout, yield_timeouts)
     40         yield None
     41 else:
---> 42     self.conn.poll()
     43     while self.conn.notifies:
     44         yield self.conn.notifies.pop(0)

OperationalError: server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
```

At first glance, this seems like the same sort of traceback you had, although this is still indirect logs. The log I have, seen from the dispatcher's listener's perspective is:

```
tools_awx_1 | 2023-02-13 21:18:21,113 ERROR    [a73e7ff7] awx.main.dispatch Error consuming new events from postgres, will retry for 40 s
tools_awx_1 |   File "/awx_devel/awx/main/dispatch/worker/base.py", line 182, in run
tools_awx_1 |     for e in conn.events(yield_timeouts=True):
tools_awx_1 |   File "/awx_devel/awx/main/dispatch/__init__.py", line 42, in events
tools_awx_1 |     self.conn.poll()
tools_awx_1 | psycopg2.OperationalError: server closed the connection unexpectedly
tools_awx_1 | 	This probably means the server terminated abnormally
tools_awx_1 | 	before or while processing the request.
```

Again, consistent with my shell_plus testing. This is showing that it falls within the `(db.DatabaseError, psycopg2.OperationalError)` exceptions being caught. This already sounds like it's doing what you propose, and I am interested to close any cases of database flake that current logic does not catch.

tanganellilore:
Thanks @AlanCoding for reply. I understand What you mean, but probably I catch why you are not able to replay. 
Use case i little different than you share now.
In HA postgres solution you have 3 instances of postgres, 1 master and 2 slaves, accessible via VIP. 

One of the most standard solutions for a failover is to move the VIP ip from master to one of the slave or use an ha proxy. 
What i can suppose  is that, the main differance beetwen this case and your examples is the connection itself.
In your case you drop all connections, shutting down the postgres and all pg_notify go down.
In failover case we have different behavior and seems that the main connection it's partially broken or will be restore partially in some way (or the switch is very fast) but it will not work properly.

What I tried to understant is why, in case of failover, i receive an exception from awx.conf.settings instead from Main loop process and this was my answer. 
Put a "ping" during the "None" message to check if connection is healty, because we check the connection unhealty only in case we recive notify. But if we don'recive notify, we don't check if connection is healty... 
We suppose that it's implicit in a single instance, and it's true, but this is not the same in a ha postgre

tanganellilore:
these some logs:
Failover at 02/14/2023, 7:53:11:

From `awx-manage shell_plus`:
image

You can see that there isn't any "conenction" drop.

So I confirm that this happen only in a single instance or in HA solution where all postgressql go down in the same time.

If we have an HA, with Patroni, pg_bouncer and vip-manager or HA proxy with VIP address, old connections go down but seems is not fetched by psycopg2 library (probaly because is very short time) .

In any case I notice that nothing check if `self.conn` is valid and working when we are in `if select.select([self.conn], [], [], select_timeout) == NOT_READY:`, so in case of connection resetted by postgres we risk to have a not working connection on awx side.

tanganellilore:
good morning @AlanCoding ,
I tried different configuration for postgres HA but in some cases is not catched by exception in main loop.
To be more precise, in all cases where there is a real failover (deattach of nw from master or brute shutdown of master) this is not catched.

And this is for me normally, becasue the vip address will be moved to one of the slave and nobody told to main loop that this connection was dropped (it not receive an ack status or dropping or something like this because master was shutted down not programmatically).
The strange things is the error was catched by `awx.conf.settings` every 1 minute 
image


This is my lab (created with https://github.com/vitabaks/postgresql_cluster):
postgresql01: 10.10.1.2 - patroni, vip-manager, etcd, pgbouncer
postgresql02: 10.10.1.3 - patroni, vip-manager, etcd, pgbouncer
etcdarbiter: 10.10.1.4 - etcd
vip: 10.10.1.5

What I tried:
server side:
- with pgbouncer (using port 6432)
- without pgbouncer (using port 5432)

address side:
- with dns that resolve to vip
- without dns (directly IP address)

Failover via patroni and `patronictl failover cluster` or normal shutdown of postgres master: Success (expected, becasue manual failover or reboot drop all connection and probably send message that is catched by psycopg2)

Failover via deattach NW to vm or shutdown not gracefully: KO, psycopg2 not catch that connection go drop and re-up in a short time

AlanCoding:
Thanks for digging and the information you gave here. I am much more-so seeing why there isn't an underlying traceback associated with the problem, and that the problem is a bit lower-level on the `select.select` not realizing that a failover had happened and made the socket it was using defunct.

I brought this up at the community meeting https://github.com/ansible/awx/issues/13563

Engineering-wise, it might be obvious I don't like adding another periodic SELECT, particularly in such a low-level fast 5 second loop. But you have good points that suggest it may be unavoidable in which case increasing the timeout might be a good counter-balance. We already run tasks in the main loop on a periodic basis, and these can _already_ talk to the database (and other systems). The weird and non-obvious point to know is that this process maintains 2 database connections - one for pg_notify listening and the other for Django ORM actions. The latter connection is closed for at least scaleup events. If the connection might ever be closed, then it is no good for listening to pg_notify because it would lose messages while closed. The entire logic for this is suspicious and I don't feel like both connections should be necessary. If those could be consolidated, my preference would be to test the connection in normal dispatcher periodic tasks. This would be a much larger change that risks unexpected fallout, but I am thinking about it.

tanganellilore:
I wasn't able to join on community meeting, so i will answer here.
I have the same concern related Select every 5 sec, also if we can consider that connection to DB is already in place and a Select will not so eavy.
To me, probably the much faster and affordable solution, until a complete rework on dispatching (or probably a rework on pg_notify 😄) could be add a counter and check the connection after X Times that we recive None event.

Consider that in some of my debug session, i notice that in a awx instance used normally None event are very, very low so this contuer can be very low, like 10/12.
So in that case you check connection every 1 minute, for a connection already in place.

So in the worst case, we stay without db connection for 1 minute with one single check per minute

AlanCoding:
> after X Times that we recive None event.

Maybe!

It will not hit the timeout every loop, because it exits early if it gets a message. We don't need to pretend that we don't know anything about the event load coming in. Every single AWX node has periodic tasks running _that are submitted to the local queue_. See:

https://github.com/ansible/awx/blob/11475590e794248676e3ad77c5d9563e1ad675aa/awx/settings/defaults.py#L452

So, since there is essentially a chron job submitting a task every 20 seconds, then if you hit 4 timeouts (5x4...) then something's wrong, it's time to panic. It could be because the fork submitting periodic tasks had an error, but it could also be due to this problem, a problem with the listening connection.

CMPluto:
Hi @AlanCoding Any update on the merging the above postgres DB instable connections? 
We can still see that in 21.12 as well.

**_[-] awx.conf.settings Database settings are not available, using defaults. error: connection already closed\n","stream":"stderr"_**

CMPluto:
Hi @AlanCoding, Any update on the merging the above postgres DB instable connections?

AlanCoding:
I'm glad that you put it on the community meeting schedule. @tanganellilore has done excellent work here, and I'm on-board with the latest state of our discussion - that we need to SELECT something with the listening connection periodically to detect this case where it has gone stale.

In the meantime, other changes have been landing in the `devel` branch. Now we have expanded the use of this "listening" connection to other services to enable the web/task pod scaling split in OCP. Whatever arguments applied to this will now apply to those services as well, so we need to discuss those and come to a solution.

tanganellilore:
@AlanCoding I try to update my fork with a new branch from devel, and try to approach with different solution that we discuss (check connection after X iteration).
Is a draft but you can check here : https://github.com/tanganellilore/awx/commits/reconnect_dispatch_v2

I know that with web/task split, `conn.events` it's called in different points than before, but this is just to show us a different approach than check conn every time.
We can discuss it on today meeting.

TheRealHaoLiu:
finally got around to pulling down your change and testing it... apology for the amount of time that it took

setup: 
- multi-AZ RDS and triggered a force failover

result:
- not good...
- all pg listener connection die so no pg_notify is working
- cluster_heathcheck fails which trigger peer judgement in repeat and all instance but 1 is reaped

tanganellilore:
Hi team,
I'm happy that you are able to "reprodce" issue and I need to be clear, this implementation was tested in a release before task//web split, so I don't know/tests if this still works.

From your tests seems that not works on RDS, but i also remeber some issues with RDS in general.
Actually a i need to recreate a postgres cluster to restest new awx but now i need some time to do it.

When I'm ready i will ping you and we can see the output

TheRealHaoLiu:
@tanganellilore https://github.com/ansible/awx-operator/pull/1393 here's a PR that we are currently testing to solve the problem

seems to be working

TheRealHaoLiu:
@tanganellilore have you tested out https://github.com/ansible/awx-operator/pull/1393 to see if it works for you? if it does lets close out this PR

tanganellilore:
as per comment on https://github.com/ansible/awx/issues/13505

Thanks for solution!!

PR #13513, opened 2023-02-02
[usage-collection] Host Metrics API + migration

slemrmartin:
##### SUMMARY

- PRD: Utility to collect Ansible usage data
- ADR 0017
- https://issues.redhat.com/browse/AA-1572

##### DESCRIPTION

- Adds columns to HostMetric
  - Changes `hostname` column from primary key to unique, adds `id` autofield
- Creates HostMetric list/detail views(API)
  - including permissions

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.11.1.dev28+ga3aff3dcff.d20230202
```

##### ADDITIONAL INFORMATION

```

```


awx/main/models/inventory.py
```diff
@@ -820,9 +820,19 @@ def _get_related_jobs(self):
...
+    hostname = models.CharField(unique=True, max_length=512)
     first_automation = models.DateTimeField(auto_now_add=True, null=False, db_index=True, help_text=_('When the host was first automated against'))
     last_automation = models.DateTimeField(db_index=True, help_text=_('When the host was last automated against'))
+    last_deleted = models.DateTimeField(null=True, db_index=True, help_text=_('When the host was last deleted'))
+    automated_counter = models.BigIntegerField(default=0, help_text=_('How many times was the host automated'))
+    deleted_counter = models.BigIntegerField(default=0, help_text=_('How many times was the host deleted'))
+    deleted = models.BooleanField(
+        default=False, help_text=_('Boolean flag saying whether the host is deleted and therefore not counted into the subscription consumption')
+    )
+    used_in_inventories = models.BigIntegerField(null=True, help_text=_('How many inventories do contain this host'))
```
gamuniz: nit on grammer `How many inventories contain this host`

slemrmartin: :+1: fixed

awx/api/views/__init__.py
```diff
@@ -1531,6 +1531,21 @@ def related_search_fields(self):
...
+    serializer_class = serializers.HostMetricSerializer
+    permission_classes = (IsSystemAdminOrAuditor,)
+    search_fields = ('hostname', 'deleted')
+
+
+class HostMetricDetail(RetrieveDestroyAPIView):
+    name = _("Host Metric Detail")
+    model = models.HostMetric
+    serializer_class = serializers.HostMetricSerializer
+    permission_classes = (IsSystemAdminOrAuditor,)
```
AlanCoding, line 1546: If you're using these permission classes, then I don't think adding an _access_ class like `HostMetricAccess` makes any sense. The access classes are only helpers for the view & permission classes. If neither will call the access class, then it's probably unused.

awx/main/access.py
```diff
@@ -861,6 +862,34 @@ def can_unattach(self, obj, sub_obj, relationship, *args, **kwargs):
...
+
+    def get_queryset(self):
+        if self.user.is_superuser or self.user.is_system_auditor:
+            qs = self.model.objects.all()
+        else:
+            qs = self.filtered_queryset()
+        return qs
+
+    def can_read(self, obj):
+        return bool(self.user.is_superuser or self.user.is_system_auditor or (obj and obj.user == self.user))
```
AlanCoding, line 881: What do you think `obj.user` is supposed to be here? I don't see it as a field on the `HostMetric` model.

awx/api/views/__init__.py
slemrmartin: @AlanCoding when I omit this class, then API call produces 500 error with `KeyError` in the `register_access()` method

awx/main/access.py
slemrmartin: Yeah, I was also wondering if it's required. I just followed the pattern below => I'll remove it

slemrmartin:
closing in favor of https://github.com/ansible/awx/pull/13560

PR #13467, opened 2023-01-24
Changing to common action

john-westcott-iv:
##### ISSUE TYPE

 - Bug, Docs Fix or other nominal change

##### COMPONENT NAME

 - Other

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

PR #13516, opened 2023-02-02
Attempt to consolidate CI logic with github_ci_runner target

AlanCoding:
##### SUMMARY
I want to get more logic reuse, and also, move logic away from the github actions YAML and into the Makefile.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
Related to https://github.com/ansible/awx/pull/13467, but trying to be more conservative to clean up before we undertake larger undertakings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -301,13 +301,20 @@ test:
 	cd awxkit && $(VENV_BASE)/awx/bin/tox -re py3
 	awx-manage check_migrations --dry-run --check  -n 'missing_migration_file'
 
+## Login to Github container image registry, pull image, then build image.
 github_ci_setup:
-	echo $(CI_GITHUB_TOKEN) | docker login ghcr.io -u $(GITHUB_ACTOR) --password-stdin  # Log in to registry
+	# GITHUB_ACTOR is automatic github actions env var
+	# CI_GITHUB_TOKEN is defined in .github files
+	echo $(CI_GITHUB_TOKEN) | docker login ghcr.io -u $(GITHUB_ACTOR) --password-stdin
 	docker pull $(DEVEL_IMAGE_NAME) || :  # Pre-pull image to warm build cache
-	make docker-compose-build	 # Build image
+	make docker-compose-build
 
-github_ci_runner: github_ci_setup
-	docker run -u $(shell id -u) --rm -v $(GITHUB_WORKSPACE):/awx_devel/:Z --workdir=/awx_devel $(DEVEL_IMAGE_NAME) $(CI_GITHUB_COMMAND)
+## Runs AWX_DOCKER_CMD inside a new docker container.
+docker-runner:
+	docker run -u $(shell id -u) --rm -v $(shell pwd):/awx_devel/:Z --workdir=/awx_devel $(DEVEL_IMAGE_NAME) $(AWX_DOCKER_CMD)
+
+## Builds image and runs AWX_DOCKER_CMD in it, mainly for .github checks.
+github_ci_runner: github_ci_setup docker-runner
 
 test_collection:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -301,6 +301,14 @@ test:
 	cd awxkit && $(VENV_BASE)/awx/bin/tox -re py3
 	awx-manage check_migrations --dry-run --check  -n 'missing_migration_file'
 
+github_ci_setup:
```
john-westcott-iv, line 305: Can we add a comment above the targets so `make help` prints a message for these new targets?

Makefile
```diff
@@ -301,6 +301,14 @@ test:
...
+	echo $(CI_GITHUB_TOKEN) | docker login ghcr.io -u $(GITHUB_ACTOR) --password-stdin  # Log in to registry
+	docker pull $(DEVEL_IMAGE_NAME) || :  # Pre-pull image to warm build cache
+	make docker-compose-build	 # Build image
+
+github_ci_runner: github_ci_setup
```
john-westcott-iv: I'm wondering if this one might be a good "Generic" command to have for reuse? i.e. just call it container_runner or something, change `GI_GITHUB_COMMAND` to just `COMMAND_TO_RUN ?= awx-manage shell_plus` and also swap out `GITHUB_WORKSPACE` to something like `MOUNT_DIR ?= .`

john-westcott-iv (approved):
Minor comments added.

Makefile
AlanCoding: There's a messy mess of environment variables here, let me categorize:
 - Things that start with `CI_`  had to be added specifically for this makefile target. This is because they were in the github workflow varspace, but not as env vars, like `${{ secrets.GITHUB_TOKEN }}`, so the github actions has to add it as a (novel) env var to pass the data
 - A couple of things like `GITHUB_WORKSPACE` are documented to be provided automatically by github actions, so there's no need to pass anything, but we also don't have freedom to rename it, as long as we want to take advantage of it. That's why I wouldn't want to call it `MOUNT_DIR`
 - Many, many, things are already declared Makefile variables like `COMPOSE_TAG`, so I used those existing names.

Sure, it could be used for other uses, but I'd like to be very explicit that the first and dominating purpose is to use it for Github actions. You have a really helpful suggestion about the help text and I'll look into that, because otherwise what I wrote here about the env vars will be promptly forgotten.

AlanCoding: I tried to give this some genuine thought, so here's my answer -

I still want dedicated `github_` targets, but this is an oddly general target. Thus, I split out a new command `make docker-runner` which should be compared to `make docker-compose` and others. Then, `github_ci_runner` is just a combination of the github setup command & ansible-runner.

AlanCoding:
I pushed 1 more significant change. The `label` wasn't used correctly due to a typo. However, instead of correcting that, I'd rather get rid of the label. The naming always turns out to be redundant, see:

Screenshot from 2023-02-03 11-47-06

Additionally, I don't think this is helpful for _locating_ the test. Here's the updated version, with a bit of a better view of the context.

Screenshot from 2023-02-03 11-56-19

When we have other stages, then it's github-level stuff. The check itself is doing image stuff and running the test. This isn't getting timing for each separately, but we could add it to the Makefile. Either way, I think it's good to name one stage the same as the name of the check, because that's the actual logic with it.

john-westcott-iv (approved):

AlanCoding:
shoot, it seems that the image layer caching isn't working as it should, and checks are taking too long.

shanemcd:
The image cache is probably getting expired because you changed the Makefile, and that's one of the very first layers. Once this lands in devel and this pre-built image goes up to ghcr.io, subsequent PRs should be faster again.

PR #13521, opened 2023-02-03
Add error handling to scm_version.py script

AlanCoding:
##### SUMMARY
I hit cases where this script failed in CI, but then all it tells you is that the subprocess failed, but doesn't give the output of the subprocess.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/tools/scripts/scm_version.py
+++ b/tools/scripts/scm_version.py
@@ -14,15 +14,7 @@ except ModuleNotFoundError:
         result = subprocess.run(cmd, capture_output=True)
         if result.returncode:
             # failed, we have no version, so print output so that users can debug
-            print(f'\nCommand `{" ".join(cmd)}` failed (rc={result.returncode}).\n\nstdout:')
-            sys.stdout.buffer.write(result.stdout)
-            sys.stdout.flush()
-            sys.stderr.flush()
-            print('\nstderr:\n')
-            sys.stderr.buffer.write(result.stderr)
-            sys.stderr.flush()
-            print('\n')
-            result.check_returncode()  # raise exception from subprocess
+            raise Exception(f'\nCommand `{" ".join(cmd)}` failed (rc={result.returncode}).\n\nstdout:\n{result.stdout}\n\nstderr:\n{result.stderr}')
 
     from setuptools_scm import get_version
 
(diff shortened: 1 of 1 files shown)


AlanCoding:
The error I mentioned turns out affects all successful CI runs with this error text:

```
Unable to import setuptools-scm, attempting to install now...
  File "/home/runner/work/awx/awx/tools/scripts/scm_version.py", line 7, in <module>
    from setuptools_scm import get_version
ModuleNotFoundError: No module named 'setuptools_scm'

  File "/home/runner/work/awx/awx/tools/scripts/scm_version.py", line 12, in <module>
    subprocess.check_output([sys.executable, '-m', 'ensurepip'])
  File "/usr/lib/python3.[10](https://github.com/ansible/awx/actions/runs/4831164595/jobs/8608244693?pr=13924#step:3:11)/subprocess.py", line 4[20](https://github.com/ansible/awx/actions/runs/4831164595/jobs/8608244693?pr=13924#step:3:21), in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.10/subprocess.py", line 524, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['/usr/bin/python3', '-m', 'ensurepip']' returned non-zero exit status 1.
```

To summarize:
 - it fails to compute the version
 - error details for that error are missing
 - this error is non-fatal
 - rest of checks continue and are successful

Currently, so-far, this PR gets it to the point where it accurately reports the error, which is...

```
Exception: 
Command `/usr/bin/python3 -m ensurepip` failed (rc=1).
stdout:
b'ensurepip is disabled in Debian/Ubuntu for the system python.\n\nPython modules for the system python are usually handled by dpkg and apt-get.\n\n    apt install python3-<module name>\n\nInstall the python3-pip package to use pip itself.  Using pip together\nwith the system python might have unexpected results for any system installed\nmodule, so use it on your own risk, or make sure to only use it in virtual\nenvironments.\n\n'
```

re-formatted...

```
ensurepip is disabled in Debian/Ubuntu for the system python.

Python modules for the system python are usually handled by dpkg and apt-get.

    apt install python3-<module name>

Install the python3-pip package to use pip itself.  Using pip together
with the system python might have unexpected results for any system installed
module, so use it on your own risk, or make sure to only use it in virtual
environments.
```

Obviously, it would also be good to fix the root of the problem... or stop running this script.

shanemcd (approved):

Issue #13336, opened 2022-12-14
Testing controller hostname DNS resolution is preventing awx_collection to work with http[s]_proxy


bagasse:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

ControllerApi code check for host resolution via `getaddrinfo` (see https://github.com/ansible/awx/blob/devel/awx_collection/plugins/module_utils/controller_api.py#L146).

This prevent any module to work behind http[s]_proxy.

eg:
```
- name: Ensure organizations are configured
  ansible.controller.organization:
    controller_host: "{{ aap_controller_host }}"
    controller_username: "{{ aap_controller_username | default(omit) }}"
    controller_password: "{{ aap_controller_password | default(omit) }}"
    validate_certs: "{{ aap_controller_validate_certs | default(omit) }}"
    default_environment: "{{ organization.default_environment | default(omit) }}"
    description: "{{ organization.description | default(omit) }}"
    galaxy_credentials: "{{ organization.galaxy_credentials | default(omit) }}"
    instance_groups: "{{ organization.instance_groups | default(omit) }}"
    max_hosts: "{{ organization.max_hosts | default(omit) }}"
    name: "{{ organization.name }}"
    state: "{{ organization.state | default(omit) }}"
  environment:
    https_proxy: http://<http_proxy_host>:3128
```
Fails with `"msg": "Unable to resolve controller_host ([Errno -2] Name or service not known)`

Removing the `try/except` block for `getaddrinfo` make the module works with http proxy

As this modules talks on http(s), it should work with http proxies.

### AWX version

ansible.controller 4.3.0

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [X] Collection
- [ ] CLI
- [ ] Other

### Installation method

N/A

### Modifications

no

### Ansible version

2.13

### Operating system

RHEL 8

### Steps to reproduce

* Install awx.awx or downstream ansible.controller collection
* try to use collection modules via an http[s] proxy (see sumary for task example)

### Expected results

Collection's modules works

### Actual results

Fails with `"msg": "Unable to resolve controller_host ([Errno -2] Name or service not known)`

PR #13524, opened 2023-02-06
Related #13336 - DNS resolution is preventing awx_collection to work with http[s]_proxy

bagasse:
##### SUMMARY

ControllerApi code check for host resolution via getaddrinfo. This prevent any module to work behind http[s]_proxy when hostnames are not resolvable on the host where the code is run. See #13336 for more info.

##### ISSUE TYPE

 - Bug

##### COMPONENT NAME

 - Collection

##### AWX VERSION

```
awx: 0.1.dev32821+g0a6e0ae
```

##### ADDITIONAL INFORMATION

```

```


bagasse:
The other solution should be to remove this DNS test.

fosterseth:
@bagasse what controller url and port were you providing that fails this getaddrinfo check? if you cannot provide the exact url, just an equivalent form is fine

bagasse:
Hello @fosterseth, 
Thanks to take a look at this. The `controller_host` param that I used only contain the hostname of the controller (eg: aac.xxx.redhat.com), the module convert it in https://aac.xxx.redhat.com.

The thing is, when you use a HTTP(s) proxy, the DNS resolution is made by the proxy (not by the host that make the http request), and so in some case, only the http proxy is able to resolve the DNS name (that the case of my lab build on an internal Red Hat lab infra), so in that case the current test will throw an error for a valid host. In some company with a lot of different security zones and internal http proxy, it can be a pain.

The other solution/implementation should be to remove this dns resolution test, because if it's not resolvable the underlying http library will return the same error anyway.

fosterseth:
@shanemcd I think this is a valid case.

should we just remove the hostname DNS resolution gut check entirely and let the module fail further down below?

awx_collection/plugins/module_utils/controller_api.py
```diff
@@ -152,9 +152,11 @@ def __init__(self, argument_spec=None, direct_params=None, error_callback=None,
             self.url.hostname.replace(char, "")
         # Try to resolve the hostname
         try:
-            addrinfolist = getaddrinfo(self.url.hostname, self.url.port, proto=IPPROTO_TCP)
-            for family, kind, proto, canonical, sockaddr in addrinfolist:
-                sockaddr[0]
+            proxy_env_var_name = "{0}_proxy".format(self.url.scheme)
+            if not environ.get(proxy_env_var_name) and not environ.get(proxy_env_var_name.upper()):
```
TheRealHaoLiu, line 135:
```
TASK [Create Credential for target cluster in AWX] **************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Unable to resolve controller_host (name 'environ' is not defined): gabe.something.com"}
```

TheRealHaoLiu: `from os import access, R_OK, getcwd, environ` need to be added

TheRealHaoLiu:
tested the PR with the following scenarios 

with HTTPS_PROXY set and unresolvable hostname, error does not show
with HTTPS_PROXY not set and unresolvable hostname, error show `"Unable to resolve controller_host ([Errno 8] nodename nor servname provided, or not known): yoloswasdfasdfasfasfasdgabeag.com"`
with HTTPS_PROXy not set and resolvable hostname error not show 

after you fix the import we can merge it

TheRealHaoLiu:
please update your PR with description and version information

fosterseth:
@TheRealHaoLiu correct import statement now, want to do a final review? can probably merge soon

fosterseth (approved):

PR #13526, opened 2023-02-07
[chore] Update project_update playbook to be compliant with ansible-lint

TheRealHaoLiu:
##### SUMMARY
periodic therapeutic yak shaving...

TODO: 
- [x] fix `Commands should not change things if nothing needs doing.`
```
    - name: Delete project directory before update
      ansible.builtin.command: "find -delete"  # volume mounted, cannot delete folder itself
      args:
        chdir: "{{ project_path }}"
      tags:
        - delete
 ```
- [x] fix `File permissions unset or incorrect.`
```
        - name: Ensure the project directory is present
          ansible.builtin.file:
            dest: "{{ project_path | quote }}"
            state: directory
```
 - [x] fix `File permissions unset or incorrect.`
 ```
         - name: Ensure the project archive directory is present
          ansible.builtin.file:
            dest: "{{ project_path | quote }}/.archive"
            state: directory
```
- [x] fix `File permissions unset or incorrect.`
```
        - name: Get archive from url
          ansible.builtin.get_url:
            url: "{{ scm_url | quote }}"
            dest: "{{ project_path | quote }}/.archive/"
            url_username: "{{ scm_username | default(omit) }}"
            url_password: "{{ scm_password | default(omit) }}"
            force_basic_auth: true
```

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.11.1.dev64+gcdbf20da64
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/playbooks/project_update.yml
+++ b/awx/playbooks/project_update.yml
@@ -25,42 +25,47 @@
   connection: local
   name: Update source tree if necessary
   tasks:
-
-    - name: delete project directory before update
-      command: "find -delete"  # volume mounted, cannot delete folder itself
+    - name: Delete project directory before update
+      ansible.builtin.shell: set -o pipefail && find . -delete -print | head -2  # volume mounted, cannot delete folder itself
+      register: reg
+      changed_when: reg.stdout_lines | length > 1
       args:
         chdir: "{{ project_path }}"
       tags:
         - delete
 
-    - block:
-        - name: update project using git
-          git:
-            dest: "{{project_path|quote}}"
-            repo: "{{scm_url}}"
-            version: "{{scm_branch|quote}}"
-            refspec: "{{scm_refspec|default(omit)}}"
-            force: "{{scm_clean}}"
-            track_submodules: "{{scm_track_submodules|default(omit)}}"
-            accept_hostkey: "{{scm_accept_hostkey|default(omit)}}"
+    - name: Update project using git
+      tags:
+        - update_git
+      block:
+        - name: Update project using git
+          ansible.builtin.git:
+            dest: "{{ project_path | quote }}"
+            repo: "{{ scm_url }}"
+            version: "{{ scm_branch | quote }}"
+            refspec: "{{ scm_refspec | default(omit) }}"
+            force: "{{ scm_clean }}"
+            track_submodules: "{{ scm_track_submodules | default(omit) }}"
+            accept_hostkey: "{{ scm_accept_hostkey | default(omit) }}"
           register: git_result
 
         - name: Set the git repository version
-          set_fact:
+          ansible.builtin.set_fact:
             scm_version: "{{ git_result['after'] }}"
           when: "'after' in git_result"
-      tags:
-        - update_git
 
-    - block:
-        - name: update project using svn
-          subversion:
-            dest: "{{project_path|quote}}"
-            repo: "{{scm_url|quote}}"
-            revision: "{{scm_branch|quote}}"
-            force: "{{scm_clean}}"
-            username: "{{scm_username|default(omit)}}"
-            password: "{{scm_password|default(omit)}}"
+    - name: Update project using svn
+      tags:
+        - update_svn
+      block:
+        - name: Update project using svn
+          ansible.builtin.subversion:
+            dest: "{{ project_path | quote }}"
+            repo: "{{ scm_url | quote }}"
+            revision: "{{ scm_branch | quote }}"
+            force: "{{ scm_clean }}"
+            username: "{{ scm_username | default(omit) }}"
+            password: "{{ scm_password | default(omit) }}"
             # must be in_place because folder pre-existing, because it is mounted
             in_place: true
           environment:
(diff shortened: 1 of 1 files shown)


fosterseth (approved):

AlanCoding:
I don't see ansible-lint in any of our checks?

awx/playbooks/project_update.yml
```diff
@@ -26,127 +26,127 @@
   name: Update source tree if necessary
   tasks:
 
-    - name: delete project directory before update
-      command: "find -delete"  # volume mounted, cannot delete folder itself
+    - name: Delete project directory before update
+      ansible.builtin.command: "find -delete"  # volume mounted, cannot delete folder itself
```
AlanCoding: is this sufficiently version-compatible? A lot of people are surely using 2.9, which has collections, but not sure if this works.

TheRealHaoLiu: we run project update in controlplane ee so we should be fine 

in the latest awx-ee (stream9 based) we are running 2.14.2

awx/playbooks/project_update.yml
```diff
@@ -26,127 +26,127 @@
...
       args:
         chdir: "{{ project_path }}"
       tags:
         - delete
+      changed_when: false
```
AlanCoding: This isn't really correct. When the project folder contents are deleted, something _is_ changed. If you're giving a value to changed_when, `true` would be a better shot at being correct. What if you gave changed=true and wrote another expression to skip the task where there are no contents inside the folder? That is the only way I can see making this correct.

TheRealHaoLiu:
```
    - name: Detect if project directory is empty
      ansible.builtin.stat:
        path: "{{ project_path }}"
      register: project_dir
      tags:
        - delete

    - name: Delete project directory before update if not empty
      ansible.builtin.command: "find -delete"  # volume mounted, cannot delete folder itself
      args:
        chdir: "{{ project_path }}"
      tags:
        - delete
      changed_when: true
      when: project_dir.stat.exists and project_dir.stat.isdir and project_dir.stat.nlink > 2
 ```
 
 this?

Comment (relrod) [reply awx/playbooks/project_update.yml]: Why hardcode `changed_when` at all, as opposed to making it accurate? You can use `find -delete -print`, register the task output, and do something like `changed_when: reg.stdout != "."`

(that might not be exact, but that general idea)

Comment (TheRealHaoLiu) [reply awx/playbooks/project_update.yml]: i changed it to true is that good enough

Comment (relrod) [reply awx/playbooks/project_update.yml]: > i changed it to true is that good enough


It's not accurate:

```
[rick@fedora asdf]$ find . -delete -print
.
[rick@fedora asdf]$ find . -delete -print
.
```

Why should changed be true the second time?

Comment (AlanCoding) [reply awx/playbooks/project_update.yml]: I'm worried about performance implications of any version of `-print` because it prints every file. If this is recorded in the event data (as stdout) then that is the exact thing we got ourselves in trouble with when we used the Ansible built-in module (I think file) for this purpose. Once you get to "large" repo sizes (realistically, any major project out there) then the sheer number of files becomes a problem, particularly when the .git folder is included. This kind of situation would make the page un-navigable in the UI and cause a mess of other problems.

You know, we don't even really _need_ to make this determination in the playbook. Since the folder must be on the control node, we could determine in the task code whether there is anything to be deleted. You could even pass in a variable _just_ for the somewhat ridiculous purpose of setting the changed status. This would be preferable to risking performance regressions.

Comment (relrod) [reply awx/playbooks/project_update.yml]: > If this is recorded in the event data (as stdout) then that is the exact thing we got ourselves in trouble with when we used the Ansible built-in module (I think file) for this purpose.

Okay, what about:

```yaml
shell: find . -delete -print | head -2
register: reg
changed_when: reg.stdout_lines | length > 1
```

AlanCoding: yeah, that should be fine, because the bash commands will behave extremely efficiently in this case.

relrod: @TheRealHaoLiu if you go this route, note you'll need shell and not command

TheRealHaoLiu: updated as per @relrod's suggestion please re-review

gamuniz (approved):

awx/playbooks/project_update.yml
```diff
@@ -26,127 +26,128 @@
...
-            awx_license_type: "{{awx_license_type}}"
-            awx_version: "{{awx_version}}"
+            insights_url: "{{ insights_url }}"
+            username: "{{ scm_username }}"
+            password: "{{ scm_password }}"
+            project_path: "{{ project_path }}"
+            awx_license_type: "{{ awx_license_type }}"
+            awx_version: "{{ awx_version }}"
           register: results
-
```
relrod, line 96: Can you preserve these blank lines between tasks? They make mentally parsing the YAML easier.

relrod (commented):
Looks fine to me, but I'd like to see the blank lines between tasks preserved. I think they make enough of a difference when mentally/quickly parsing yaml.

AlanCoding:
I wrote an integration test for the "changed" status. The test basically assures the following:
 - after you create a project with `scm_delete_on_update` as true, then it auto-updates and the delete task is changed=False, because there is nothing to delete
 - upon a second update, that same task gives changed=True

This requires having exactly 1 control node in the system. If this is working, then I'd like to merge this and merge the test.

PR #13529, opened 2023-02-07
Config and monitoring

Red-HAP:
##### SUMMARY

This change adds database changes to enable monitoring of query performance via `pg_stat_statements` and is designed to work with a new database user named `db_monitor`. 

This change also adds an enhancement to `JobEvent.save()` that will truncate the `job_created` field (table partitioning key; datetime.datetime type) to the hour. The `save()` method will call `update_or_create()` instead of `super().save()`

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - PostgreSQL

##### AWX VERSION

```
awx: 21.11.1.dev6+g8574bf1594
```

###### PostgreSQL

This change will create a new config file for the engine instance. This config file will first tell PostgreSQL to load and consume the original config file, then apply all configurations from a mounted directory. These configs will be applied in lexicographic order of the file name. Also, initialization scripts are mounted in the DB container that will be executed only on creating a new database. For existing databases, a Django migration file has been added.

The result of these changes is:
* A new database role (user) named `db_monitor` is created that can only utilize public.pg_stat_statements and should not be able to perform any CRUD action against an application table.
* The `pg_stat_statements` shared library will be preloaded for docker container PostgreSQL
* The `pg_stat_statements` extension will be installed in the `public` schema.

###### Grafana

A development-only (ie: must not visible to support or other outside entities) dashboard will be created that will live-query the docker PostgreSQL database. The datasource for this database will also be created.

This dashboard includes as its top panel, a table rendered view of live-query of the pg_stat_statements data from the configured datasource. This is a tool that could be helpful to not only clients, but we could also simply select that data to have it exported to support for triage. This panel in the dashboard will show the mean execution time and max execution time (in milliseconds) of a statement, the number of times that statement has been executed and the parameterized form of that statement. This should not leak any customer data as the parameters are not supplied. Only the placeholders for those parameters are listed.

```

```

### DIFF
--- a/awx/main/models/events.py
+++ b/awx/main/models/events.py
@@ -596,7 +596,6 @@ UnpartitionedJobEvent._meta.db_table = '_unpartitioned_' + JobEvent._meta.db_tab
 
 
 class ProjectUpdateEvent(BasePlaybookEvent):
-
     VALID_KEYS = BasePlaybookEvent.VALID_KEYS + ['project_update_id', 'workflow_job_id', 'job_created']
     JOB_REFERENCE = 'project_update_id'
 
@@ -734,7 +733,6 @@ class BaseCommandEvent(CreatedModifiedModel):
 
 
 class AdHocCommandEvent(BaseCommandEvent):
-
     VALID_KEYS = BaseCommandEvent.VALID_KEYS + ['ad_hoc_command_id', 'event', 'host_name', 'host_id', 'workflow_job_id', 'job_created']
     WRAPUP_EVENT = 'playbook_on_stats'  # exception to BaseCommandEvent
     JOB_REFERENCE = 'ad_hoc_command_id'
@@ -841,7 +839,6 @@ UnpartitionedAdHocCommandEvent._meta.db_table = '_unpartitioned_' + AdHocCommand
 
 
 class InventoryUpdateEvent(BaseCommandEvent):
-
     VALID_KEYS = BaseCommandEvent.VALID_KEYS + ['inventory_update_id', 'workflow_job_id', 'job_created']
     JOB_REFERENCE = 'inventory_update_id'
 
@@ -887,7 +884,6 @@ UnpartitionedInventoryUpdateEvent._meta.db_table = '_unpartitioned_' + Inventory
 
 
 class SystemJobEvent(BaseCommandEvent):
-
     VALID_KEYS = BaseCommandEvent.VALID_KEYS + ['system_job_id', 'job_created']
     JOB_REFERENCE = 'system_job_id'
 
--- a/awx/main/models/events.py
+++ b/awx/main/models/events.py
@@ -596,6 +596,7 @@ UnpartitionedJobEvent._meta.db_table = '_unpartitioned_' + JobEvent._meta.db_tab
 
 
 class ProjectUpdateEvent(BasePlaybookEvent):
+
     VALID_KEYS = BasePlaybookEvent.VALID_KEYS + ['project_update_id', 'workflow_job_id', 'job_created']
     JOB_REFERENCE = 'project_update_id'
 
@@ -733,6 +734,7 @@ class BaseCommandEvent(CreatedModifiedModel):
 
 
 class AdHocCommandEvent(BaseCommandEvent):
+
     VALID_KEYS = BaseCommandEvent.VALID_KEYS + ['ad_hoc_command_id', 'event', 'host_name', 'host_id', 'workflow_job_id', 'job_created']
     WRAPUP_EVENT = 'playbook_on_stats'  # exception to BaseCommandEvent
     JOB_REFERENCE = 'ad_hoc_command_id'
@@ -839,6 +841,7 @@ UnpartitionedAdHocCommandEvent._meta.db_table = '_unpartitioned_' + AdHocCommand
 
 
 class InventoryUpdateEvent(BaseCommandEvent):
+
     VALID_KEYS = BaseCommandEvent.VALID_KEYS + ['inventory_update_id', 'workflow_job_id', 'job_created']
     JOB_REFERENCE = 'inventory_update_id'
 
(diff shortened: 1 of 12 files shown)


awx/main/migrations/0175_db_monitoring.py
```diff
@@ -0,0 +1,175 @@
...
+  from pg_settings
+ where name = %s
+ ;
+    """
+    cur = conn.cursor()
+    cur.execute(sql, ('shared_preload_libraries',))
+    res = cur.fetchone()
+    cur.close()
+    if 'pg_stat_statements' not in res[0]:
+        raise ConfigurationError("The 'shared_preload_libraries' setting must include 'pg_stat_statements'")
```
shanemcd: This requires a db configuration change, right? Making this a fatal error seems like it would inflict much pain and suffering upon both us and our support team. Let me know if I'm missing something here.

awx/main/models/events.py
```diff
@@ -586,6 +604,23 @@ def _update_host_summary_from_stats(self, hostnames):
     def job_verbosity(self):
         return self.job.verbosity
 
+    # If the Django primary key and the partition key have values, try update_or_create()
+    # Otherwise, fallback to inherited save()
+    def save(self, *args, **kwargs):
```
shanemcd: I think it might be worth pulling the changes in this file out into a separate PR while we think long and hard about the migration you added.

awx/main/models/events.py
```diff
@@ -586,6 +604,23 @@ def _update_host_summary_from_stats(self, hostnames):
...
+        if getattr(self, self.PartitionInfo.partition_key, None) is None:
+            setattr(self, self.PartitionInfo.partition_key, datetime.datetime.now())
+
+        if getattr(self, "created", None) is None:
+            setattr(self, "created", datetime.datetime.now())
+            setattr(self, "modified", self.created)
+        else:
+            setattr(self, "modified", datetime.datetime.now())
```
AlanCoding: we already do a version of this in another module

https://github.com/ansible/awx/blob/devel/awx/main/dispatch/worker/callback.py#L163

so I want to first understand what about that has not worked for you. Did you find events without these fields unset? Could you get the data for those?

awx/main/migrations/0175_db_monitoring.py
```diff
@@ -0,0 +1,175 @@
+# Generated by Django 3.2.16 on 2023-02-07 14:39
```
Ladas: @Red-HAP  Does this need to be a migration? It seems like lot of this is for a local docker development? (so e.g. it could go to the init script also?)

And e.g. the firehose script is also just for local dev or perf testing. (for inserting a lot of events in the DB to test performance). It will be easier to get this in if we'd scope it as local/dev tools changes. And we can think later what needs to be in the product.

awx/main/migrations/0175_db_monitoring.py
Red-HAP: This migration has been removed.

awx/main/models/events.py
Red-HAP: The JobEvent model has been restored.

awx/main/models/events.py
Red-HAP: The JobEvent model has been restored.

awx/main/migrations/0175_db_monitoring.py
Red-HAP: The migration has been removed. These changes should only impact development.

tools/scripts/firehose.py
```diff
@@ -239,7 +240,7 @@ def generate_events(events, job, time_delta):
         num_events = events
 
     for i in range(num_procs):
-        p = multiprocessing.Process(target=firehose, args=(job, num_events, created_stamp, modified_stamp))
+        p = multiprocessing.Process(target=firehose, args=(job, num_events, created_stamp, modified_stamp, job_created))
```
Ladas, line 243: @kdelee ^ firehose must have gotten pretty slow after we did partitioning? Because it wasn't leveraging partitions.

Ladas (approved):
This should now be touching only dev env code, so that looks good.

I am trying to find time to test this locally. (but you don't have to wait for me to merge this :-) )

Red-HAP:
@Ladas 
I may not have the correct permissions on the repo to merge. I still see `Merging is blocked` even with your approval.

Ladas:
@Red-HAP right, I am not AWX approver, so you'll need @AlanCoding for that. :-)

AlanCoding:
Do you need to create a new database volume for this to work? I'm getting this error

```
tools_postgres_1 | 2023-03-06 21:04:31.591 UTC [76] FATAL:  role "db_monitor" does not exist
```

I know this PR states from the get-go that it is intended to create this role, so I'm in some debugging now about why that didn't happen when I'm trying to run it.

tools/scripts/firehose.py
kdelee: I don't think its been heavliy used, hence the bitrot

AlanCoding:
I shelled into the postgres container and ran your script to create the role. Now I'm running into a problem where the server isn't responding at localhost:8043 like normal. I do see the dashboard appear (though not much data yet) in grafana, so it's promising..

tools/scripts/firehose.py
Ladas: @kdelee do you have something else now for testing event payloads at scale?

AlanCoding:
@shanemcd helped me figure it out. We had recent changes to the uWSGI config, so this branch wasn't working with current images (not surprising). I rebased the branch and the problem I mentioned in my last comment is resolved.

Would you mind both rebasing your branch and squashing your commits into 1? Then I think we should be good to merge. I am happy to see this data, as are others. Excited to see what kind of optimizations this might help us with.

AlanCoding:
Screenshot from 2023-03-07 08-50-25

PR #13530, opened 2023-02-07
Simplify OrderedManyToManyField and restrict some cases

AlanCoding:
##### SUMMARY
I'm just trying to add a little bit more sanity for this field.

One thing I wanted to do was test that the reverse relationships don't work, but it turns out that it's really not possible to access them in the first place.

##### ISSUE TYPE

##### COMPONENT NAME
 - API


AlanCoding:
pulled into feature branch.

PR #13533, opened 2023-02-07
Simplifications for DependencyManager

AlanCoding:
##### SUMMARY
I want to pull in some of the core changes from https://github.com/ansible/awx/pull/13061 which isn't viable due to other reasons.

EDITed for changes that were done since opening

Some principles here...
 - virtually all of the logic for spawning an update-on-launch project update is shared between inventory updates & jobs, as this feature applies to both of them in the same way. This replaces what seemed like less-clear code sharing before.
 - only jobs generate an inventory update, so this is left as boutique logic in `get_dep_for_job`
 - A few operations like `add_dependencies` is generic to all dependency spawning, so it is moved further up in the stack which eliminates redundancy.
 - The utility method `should_update_again` answers whether a new update is needed due to latest update and cache timeout rules. It applies to update-on-launch regardless of the type (project vs inventory update), so the method is consolidated to reflect this. The old code strongly appeared to hint at different rules for different job types, but this was never the case.
 - Projects are cached along with inventory sources. Caching is for performance, but this is done here mainly for consistency.
 - Stop tracking created dependencies vs. all dependencies, as this is covered by excluding tasks with `dependencies_processed=True` in the main loop

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -25,7 +25,6 @@ from awx.main.models import (
     InventoryUpdate,
     Job,
     Project,
-    ProjectUpdate,
     UnifiedJob,
     WorkflowApproval,
     WorkflowJob,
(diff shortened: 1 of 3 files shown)


AlanCoding:
Surprisingly, this passes all the integration tests.

awx/main/scheduler/task_manager.py
```diff
@@ -323,21 +303,13 @@ def should_update_inventory_source(self, job, latest_inventory_update):
             return False
 
         timeout_seconds = timedelta(seconds=latest_inventory_update.inventory_source.update_cache_timeout)
-        if (latest_inventory_update.finished + timeout_seconds) < now:
+        if (latest_inventory_update.finished + timeout_seconds) < job.created:
```
AlanCoding: This is not good, need to change this or otherwise think about more.

awx/main/scheduler/task_manager.py
```diff
@@ -393,7 +354,7 @@ def gen_dep_for_job(self, task):
                 continue
             if not inventory_source.update_on_launch:
                 continue
-            latest_inventory_update = self.get_latest_inventory_update(inventory_source)
+            latest_inventory_update = InventoryUpdate.objects.filter(inventory_source=inventory_source).order_by("-created").first()
```
AlanCoding: probably want `inventory_source.inventory_updates`

awx/main/scheduler/task_manager.py
```diff
@@ -410,29 +359,33 @@ def gen_dep_for_inventory_update(self, inventory_task):
...
         return created_dependencies
 
     @timeit
     def generate_dependencies(self, undeped_tasks):
         created_dependencies = []
+        processed_ids = []
         for task in undeped_tasks:
+            if task.dependencies_processed:
+                continue
+            processed_ids.append(task.id)
```
AlanCoding: @fosterseth I added this because we call `generate_dependencies` an extra time for in case of deps-of-deps, passing in deps. In most cases those deps do not need to have their own deps generated. I will trust my optimizations to create_unified_job, and if a dependent project update has `dependencies_processed` as `False` then obviously we would like to exclude it from this loop for performance and to avoid errors from re-processing.

awx/main/scheduler/task_manager.py
```diff
@@ -271,170 +270,107 @@ class DependencyManager(TaskBase):
...
+    @staticmethod
+    def should_update_again(update, cache_timeout):
         '''
-        If there's already a inventory update utilizing this job that's about to run
-        then we don't need to create one
+        If it has never updated, we need to update
+        If there is already an update in progress then we do not need to a new create one
+        If the last update failed, we always need to try and update again
+        If current time is more than cache_timeout after last update, then we need a new one
         '''
```
AlanCoding, line 313: You can summarize the diff around here by saying that we combined 2 methods into 1. That is:

`should_update_inventory_source` & `should_update_related_project` --> `should_update_again`

the method signature changed slightly from `(self, job, latest_project_update/latest_inventory_update)` to `(update, cache_timeout)`. Maybe I should call it `latest_update` instead of `update` for clarity, and the fact that it's the same as the last parameter of the old method. This can be either a project update or a inventory update. And now that I think about it, it might be better to infer `cache_timeout` inside of this method, getting us down to 1 argument, because that is truly the only thing we need.

fosterseth (approved):
+1 simplification

Issue #13540, opened 2023-02-08
[constructed-inventory] API validations for constructed_inventory


TheRealHaoLiu:
- [x] inventory_sources should not be editable for inventory of type 'constructed'
- [x] Can constructed inventories be used as inputs for other constructed inventories? NO!
- [x] Do not allow editing or deleting hosts or groups in constructed inventories
- [x] update on launch on the auto-created constructed inventory source should not be "modifiable" and should always be true
- [x] Do not allow attaching "constructed" inventory source to other non-constructed inventories
- [x] Make "kind" field a read only field for the constructed inventory views
- [x] do not allow to add host and group to constructed inventory

TheRealHaoLiu:
update on launch on the auto-created constructed inventory source should not be "modifiable"

PR #13545, opened 2023-02-09
Added readonly api endpoint for constructed hosts

gamuniz:
##### SUMMARY

Constructed inventories should not allow adding hosts directly to it, created a new inventory list endpoint specifically for constructed inventories

##### ISSUE TYPE

 - New or Enhanced Feature
 - 
##### COMPONENT NAME

 - API
 

##### AWX VERSION

```
make VERSION
awx: 21.11.1.dev87+g4e6e9bb074
```

##### ADDITIONAL INFORMATION

```

```


AlanCoding:
My first impression is that I would prefer to not introduce a new view and use `/api/v2/inventories/N/hosts/` but we need some solutions for https://github.com/ansible/awx/issues/13540

PR #13555, opened 2023-02-10
[constructed-inventory] remove ability to add constructed inventories to input inventories

gamuniz:
##### SUMMARY

related: https://github.com/ansible/awx/issues/13540

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```
make VERSION
awx: 21.11.1.dev90+gc28768cc08

```

### DIFF
--- a/awx/api/views/inventory.py
+++ b/awx/api/views/inventory.py
@@ -115,6 +115,15 @@ class InventoryInputInventoriesList(SubListAttachDetachAPIView):
     parent_model = Inventory
     relationship = 'input_inventories'
 
+    # Specifically overriding the post method on this view to disallow constructed inventories as input inventories
+    def post(self, request, *args, **kwargs):
+        obj = Inventory.objects.get(id=request.data.get('id'))
+        if obj.kind == 'constructed':
+            return Response(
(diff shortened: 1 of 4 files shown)


awx/api/views/inventory.py
```diff
@@ -115,6 +115,15 @@ class InventoryInputInventoriesList(SubListAttachDetachAPIView):
     parent_model = Inventory
     relationship = 'input_inventories'
 
+    # Specifically overriding the post method on this view to disallow constructed inventories as input inventories
+    def post(self, request, *args, **kwargs):
+        obj = Inventory.objects.get(id=request.data.get('id'))
```
AlanCoding: It would be good to implement this at some lower level, so that we do not re-invent the wheel like with this logic to get the object.

AlanCoding: this method was made for this purpose:

https://github.com/ansible/awx/blob/0815f935ca359537c023efadf286671e31dc3374/awx/api/generics.py#L633-L635

So here, I think `sub` is the input inventory, so you can check `obj.kind` without having to do `Inventory.objects.get`, which would obligate you to do error handling if you couldn't get the object.

AlanCoding (approved):

PR #13556, opened 2023-02-10
Add selective type enforcement to a test fixture

AlanCoding:
##### SUMMARY
We want to use some newer python stuff, so I wanted to use the request fixtures as a foot-in-the-door towards more type hinting. This jives well, because the method signature is a bit of a mess as it stands.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/tests/functional/api/test_job_template.py
+++ b/awx/main/tests/functional/api/test_job_template.py
@@ -1,3 +1,5 @@
+import urllib.parse
+
 import pytest
 
 # AWX
@@ -8,9 +10,11 @@ from awx.main.migrations import _save_password_keys as save_password_keys
 
 # Django
 from django.apps import apps
+from django.urls import resolve
 
 # DRF
 from rest_framework.exceptions import ValidationError
+from rest_framework.test import APIRequestFactory, force_authenticate
 
 
 @pytest.mark.django_db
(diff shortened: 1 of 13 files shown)


awx/main/tests/functional/api/test_job_template.py
```diff
@@ -217,7 +225,7 @@ def test_jt_admin_copy_edit_functional(jt_copy_edit, rando, get, post):
 
     post_data = get_response.data
     post_data['name'] = '%s @ 12:19:47 pm' % post_data['name']
-    post_response = post(reverse('api:job_template_list'), user=rando, data=post_data)
+    post_response = post(reverse('api:job_template_list'), user=rando, data=dict(post_data))
```
john-westcott-iv, line 228: `post_data` appears to already be a dict, why do you have to cast it again? Or is it not really a dict?

AlanCoding: from memory, it was a Django `ResponseDict` or something like that. This does bring up pained philosophical questions about type hints, like, does it allow subclasses of the expected type? But for the purposes of this particular fixture, I think I'm fine forcing all callers to cast to a plain dict.

requirements/requirements_dev.txt
```diff
@@ -5,6 +5,7 @@ django-rest-swagger
 ipython>=7.31.1 # https://github.com/ansible/awx/security/dependabot/30
 unittest2
 black
+type-enforced
```
john-westcott-iv, line 8: Should we move this into requirements.in/txt incase someone wants to use in the api code? Or would that cause issues with other things?

AlanCoding: I don't think you're supposed to use this in production? It's confusing, I need a python type hints 101 class myself.

AlanCoding:
closing b/c of age and non-critical

PR #13560, opened 2023-02-13
[usage-collection] HostMetric + HostMetricSummary db & api

slemrmartin:
###### API and migrations for HostMetrics
- https://issues.redhat.com/browse/AA-1572

- DB schema of HostMetric implementation 
  - [ADR chapter 1](https://docs.google.com/document/d/1hv14yISL8GlX0Kd-xO9h9JFP9snyob5wEfUDlBimwIc/edit#heading=h.zen2zh5g8n2h)
- API for HostMetric
  - GET /api/v2/host_metrics/
  - GET /api/v2/host_metrics/<pk>
  - DELETE /api/v2/host_metics/<pk>

###### HostMetric updates and soft deletes
- https://issues.redhat.com/browse/AA-1586

- HostMetric updating:
  - After the event of type "playbook_on_stats" arrives, increment automated_counter and set deleted=False (together with current first_automation/last_automation timestamps)
  -  [ADR chapter 2.1](https://docs.google.com/document/d/1hv14yISL8GlX0Kd-xO9h9JFP9snyob5wEfUDlBimwIc/edit#heading=h.y8nzfmka0vtl)
- HostMetric soft deleting:
  - Deleting is done through API call - instead classic destroy implement soft delete (deleted=True + last_deleted = now()
  - [ADR chapter 2.2](https://docs.google.com/document/d/1hv14yISL8GlX0Kd-xO9h9JFP9snyob5wEfUDlBimwIc/edit#heading=h.6at5ooqer0l)
  - [ADR chapter 5.2](https://docs.google.com/document/d/1hv14yISL8GlX0Kd-xO9h9JFP9snyob5wEfUDlBimwIc/edit#heading=h.3gpklf9qd0i6)

###### Changing compliance computation 
- https://issues.redhat.com/browse/AA-1594 

- Subscription Compliance computing filters out deleted hosts (more accurately deleted HostMetrics)
- [ADR chapter 2.3](https://docs.google.com/document/d/1hv14yISL8GlX0Kd-xO9h9JFP9snyob5wEfUDlBimwIc/edit#heading=h.xd00948k9pd4)

######  HostMetricSummary API and migrations
- https://issues.redhat.com/browse/AA-1589

- HostMetricSummary is a new model with aggegations
- DB schema: [ADR chapter 1](https://docs.google.com/document/d/1hv14yISL8GlX0Kd-xO9h9JFP9snyob5wEfUDlBimwIc/edit#heading=h.zen2zh5g8n2h)
- API: [ADR chapter 5.3](https://docs.google.com/document/d/1hv14yISL8GlX0Kd-xO9h9JFP9snyob5wEfUDlBimwIc/edit#heading=h.vafgfz50acc3)
  - GET /api/v2/host_metric_summary_monthly/
    -  past_months=_number_ # custom filter, min=1, default=12, max=36

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.11.1.dev71+gdaf842522e.d20230213
```

##### ADDITIONAL INFORMATION

```

```


docs/clustering.md
```diff
@@ -1,92 +1,169 @@
-## AWX Clustering/HA Overview
```
AlanCoding: It looks like the changes reflected in this file are due to a bad git rebase or something along those lines.

awx/main/models/inventory.py
```diff
@@ -820,9 +820,47 @@ def _get_related_jobs(self):
...
+    hostname = models.CharField(unique=True, max_length=512)
     first_automation = models.DateTimeField(auto_now_add=True, null=False, db_index=True, help_text=_('When the host was first automated against'))
     last_automation = models.DateTimeField(db_index=True, help_text=_('When the host was last automated against'))
+    last_deleted = models.DateTimeField(null=True, db_index=True, help_text=_('When the host was last deleted'))
+    automated_counter = models.BigIntegerField(default=0, help_text=_('How many times was the host automated'))
+    deleted_counter = models.BigIntegerField(default=0, help_text=_('How many times was the host deleted'))
+    deleted = models.BooleanField(
+        default=False, help_text=_('Boolean flag saying whether the host is deleted and therefore not counted into the subscription consumption')
+    )
+    used_in_inventories = models.BigIntegerField(null=True, help_text=_('How many inventories contain this host'))
```
AlanCoding: I recognize that it's hard to draw an exact line between integer vs. big integer fields. But I think a host (even localhost) will not be used in over 2 billion inventories.

awx/main/access.py
```diff
@@ -861,6 +863,61 @@ def can_unattach(self, obj, sub_obj, relationship, *args, **kwargs):
...
+        return bool(self.user.is_superuser or self.user.is_system_auditor or (obj and obj.user == self.user))
+
+    def can_add(self, data):
+        return False  # There is no API endpoint to POST new settings.
+
+    def can_change(self, obj, data):
+        return False
+
+    def can_delete(self, obj):
+        return False
```
AlanCoding: Because this is just superuser/auditor, and that is already handled in the permission classes, I don't think we should have access classes.

I left this comment in the other PR, and you mentioned getting a 500 error. I would rather work through that error than add dead code.

awx/api/serializers.py
```diff
@@ -5002,6 +5004,32 @@ class Meta:
         fields = read_only_fields
 
 
+class HostMetricSerializer(BaseSerializer):
+    show_capabilities = ['delete']
```
AlanCoding, line 5008: I'd also prefer to not add user_capabilities with this line, given the basic nature of the permission rules.

awx/main/migrations/0175_add_hostmetric_fields.py
```diff
@@ -0,0 +1,43 @@
...
+                default=False, help_text='Boolean flag saying whether the host is deleted and therefore not counted into the subscription consumption'
+            ),
+        ),
+        migrations.AddField(
+            model_name='hostmetric',
+            name='used_in_inventories',
+            field=models.BigIntegerField(null=True, help_text='How many inventories contain this host'),
+        ),
+        migrations.AddField(
+            model_name='hostmetric', name='id', field=models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')
```
AlanCoding, line 41: Let me talk this out here - it is adding an id field because you made the `hostname` no longer the primary key (but you kept it unique, which it is already now).

Why add an id field?

awx/main/models/inventory.py
```diff
@@ -820,9 +820,47 @@ def _get_related_jobs(self):
...
 class HostMetric(models.Model):
-    hostname = models.CharField(primary_key=True, max_length=512)
+    hostname = models.CharField(unique=True, max_length=512)
     first_automation = models.DateTimeField(auto_now_add=True, null=False, db_index=True, help_text=_('When the host was first automated against'))
     last_automation = models.DateTimeField(db_index=True, help_text=_('When the host was last automated against'))
+    last_deleted = models.DateTimeField(null=True, db_index=True, help_text=_('When the host was last deleted'))
+    automated_counter = models.BigIntegerField(default=0, help_text=_('How many times was the host automated'))
+    deleted_counter = models.BigIntegerField(default=0, help_text=_('How many times was the host deleted'))
+    deleted = models.BooleanField(
+        default=False, help_text=_('Boolean flag saying whether the host is deleted and therefore not counted into the subscription consumption')
```
AlanCoding, line 830: All deletion related fields, `last_deleted`, `deleted_counter`, and `deleted` are going to be very challenging to satisfy from a performance standpoint. We have a task `delete_inventory` that is the asynchronous task for deleting an inventory and its contents. It breaks frequently, and this will add additional logic to it. I can start talking out some things we need to do for that...

We should absolutely make HostMetric updates _outside of_ the database transaction where we delete the inventory objects, because I would prefer to move along with partially incorrect data as opposed to revert the inventory deletion which would be a bug that significantly interferes with the user experience.

I'm a little more concerned with the implications for inventory imports. The entire beautify of `HostMetric` as it exists now is that we only had to update it _reactively_ as jobs finished. With the addition of these fields we will have to _proactively_ update the data as a side-effect of the inventory update logic to save the content.

What if a previously-automated `Host` is _renamed_? Should the `HostMetric` model reflect that the host was deleted? If yes, then this is a costly decision.

awx/main/models/events.py
```diff
@@ -575,12 +575,26 @@ def _update_host_summary_from_stats(self, hostnames):
...
+        )
+        # bulk-update
+        batch_start, batch_size = 0, 1000
+        while batch_start <= len(updated_hosts_list):
+            batched_host_list = updated_hosts_list[batch_start : (batch_start + batch_size)]
+            HostMetric.objects.filter(hostname__in=batched_host_list).update(
+                last_automation=current_time, automated_counter=models.F('automated_counter') + 1, deleted=False
             )
-            HostMetric.objects.filter(hostname__in=updated_hosts_list).update(last_automation=current_time)
+            batch_start += batch_size
```
TheRealHaoLiu: @kdelee @AlanCoding do you have any concern about the performance implication of this change?

AlanCoding: I'd ballpark guesstimate that it would take 2x as long as it did before. Instead of bulk creating these, now this will bulk-create and then bulk-update. While that's not ideal, I believe that shortcoming was obvious from the get-go and this is a price paid to keep the operations in the domain of the ORM, which I agree with.

I would like to use the Django `batch_size` parameter as opposed to a hard-coded batch size, but I realize that's not compatible with `.update`, because the number of rows affect is unknown by Django (although known by us).

This does lead us into the obvious potential improvement, which is that we could have this python code apply additional fields like `last_automation` to the `.bulk_create` and exclude those hosts from the `.update`. Even if that includes some unlikely race conditions I think it would be my preference, but this is getting too deep into things. What we have here should probably suffice.

awx/main/migrations/0175_add_hostmetric_fields.py
slemrmartin: @AlanCoding because Detail view needs URL friendly identifier
Which means we need either id or slug (hostname is not in an url-friendly format). 
And because AWX doesn't use slug nowhere, ID is the solution - using the same convention like all other models with views.

awx/main/models/inventory.py
slemrmartin: cc @Ladas

docs/clustering.md
slemrmartin: :-o I don't get it. How could it happen? What is "bad" rebase and "good" rebase? I'm rebasing against devel branch

awx/main/models/inventory.py
slemrmartin: it sounds reasonable, maybe the counters don't need Big int as well
cc @Ladas

awx/main/access.py
slemrmartin: Ok, I'm not familiar with Django yet, so I'll ask you in Slack what is and isn't  needed regarding the permissions here

awx/main/models/events.py
slemrmartin: I tried to use upsert first, but it's not implemented in Django's ORM. And using psycopg2 with upsert is not compatible with sqlite3 in tests. 
Anyway yes, if the `bulk_create` returns created hostnames without ignored, I can exclude them from `update`

awx/api/serializers.py
Ladas: @AlanCoding can you be more specific here? We don't know the pitfalls of the permission rules. Is this needed to be able to delete via API?

awx/main/migrations/0175_add_hostmetric_fields.py
Ladas: it makes sense to use ID, unless we have a better solution here

awx/main/models/events.py
Ladas: @AlanCoding the update was there even before, so this shouldn't be adding more processing time. (or just a small fraction, due to updates in another columns that are new)

Martin is adding batching for the update, because before it wasn't batched. Afaik it should be batched because from my experience doing a query with "hostname IN <100k hostnames>" will just explode in DB. (mostly because PG will not do index scan) So this approach should be faster for jobs with a lot of hosts.

-------

Martin already has the upsert query (doing atomic insert and update as one query) which will be much more performant, but we're not able to test it with SQLite. (I was looking for a while how to make some tests run with Postgres in AWX, but it doesn't seems that trivial to do. Although e.g. Automation Analytics runs tests on Postgres)

awx/main/models/inventory.py
Ladas: @AlanCoding right now there is no cascade delete or any connection to Inventory.

Soft-Deletes from HostMetric will be done manually through UI/API and automatically via the scheduled task. (so e.g. automatically marking HostMetric rows as soft-deleted after a year with no automation, giving user the ability to tweak the timeframe via settings)

Then if the host is automated again, the deleted flag is reverted. So adding/removing from inventory will not have any effect here now.

------

Connection to Inventory is a future TBD, so e.g. we might attempt to delete the host if it will be deleted from inventory and it's a last occurence of the host. But there is no such requirement now. (the only connection here is an periodic job, that will compute how many inventories is the HostMetric row in, so customers can order by this in the UI. Later we could leverage it for scheduled delete too...)

awx/main/models/inventory.py
Ladas: @AlanCoding @slemrmartin yeah that is fair. You're right that the counter will not likely hit 2B of automations/deletes. So for the safe of storage saving, we can make them just Int.

awx/main/models/inventory.py
slemrmartin: @AlanCoding 
I tested the bulk_create:
In Django 3.2 (used in awx), the bulk_create returns all objects including ignored. And even ignored have updated values. So it's not possible to recognize what was and what wasn't updated without additional query => excluding from update is not worth. 
Regarding [django 4.2 doc](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#bulk-create) the `bulk_create` offers `update_fields` and `update_conflicts` (not clear about effectivity). So in order to keep usage of ORM and be more effective Django update is required

awx/main/models/events.py
slemrmartin: @AlanCoding 
I tested the bulk_create:
In Django 3.2 (used in awx), the bulk_create returns all objects including ignored. And even ignored have updated values. So it's not possible to recognize what was and what wasn't updated without additional query => excluding from update is not worth. 
Regarding [django 4.2 doc](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#bulk-create) the `bulk_create` offers `update_fields` and `update_conflicts` (not clear about effectivity). So in order to keep usage of ORM and be more effective Django update is required

AlanCoding: Thanks, those are great replies. I believe this is a carefully engineered set of queries, and I'm not particularly worried about performance.

One thing that is truly novel here is the 1000 manual batch size. The Host `name` field has a max of 512 characters. Are Django and postgres okay if you run a job with >1000 hosts where they all have a name 512 characters? That's the main risk I can see being introduced here, and could be handled with a fairly straightforward test case.

awx/main/migrations/0175_add_hostmetric_fields.py
AlanCoding: Thanks for the reply. Required for purpose of `/api/v2/host_metric/42/` endpoint :+1:

awx/api/serializers.py
AlanCoding: User capabilities are extra data from API dumped into "summary_fields" but the purpose is ultimately for the UI. The UI doesn't always need it, but if no UI is even planned then the need for the data won't exist.

docs/clustering.md
AlanCoding: Looks like you fixed it!

awx/main/access.py
slemrmartin: @AlanCoding I digged deeper into it. More systematic approach seems to be remove the "permission_classes" from the views. Because they are almost not used except one specific case. 
On the other side working through error means doing something like
```
if model class == 'HostMetric' or model_class == 'HostMetricSummaryMonthly':
   ...
else:
   ...
```
in User related method: [User.get_query_set()](https://github.com/ansible/awx/blob/05af2972bf3799d3059363dd721d300435ccdbb4/awx/main/access.py#L143)
that's much worse

awx/api/serializers.py
slemrmartin: @AlanCoding the UI for this feature is currently in progress

docs/clustering.md
slemrmartin: yeah, thanks to @TheRealHaoLiu :+1:

awx/main/models/events.py
Ladas: @AlanCoding right, so that is the code I was pointing you to, before we were doing no batching (look at the line being removed):
```
HostMetric.objects.filter(hostname__in=updated_hosts_list).update(last_automation=current_time)
```

I assume it must have been exploding on something like 100k hosts (postgres hates large IN queries, the performance then significantly drops), the query size limit is much bigger, like 1GB. (I normally do upserts with 10k rows as a batch). The point is to make it small enough so this still causes an index scan. 1k should be good.

Anyway, we can add it into a test to make sure, so just generating a playbook on stats with 1k hostames of max size.

awx/main/models/inventory.py
```diff
@@ -820,9 +820,47 @@ def _get_related_jobs(self):
...
+        if not self.deleted:
+            self.deleted_counter = (self.deleted_counter or 0) + 1
+            self.last_deleted = now()
+            self.deleted = True
+            self.save()
+
+    def soft_restore(self):
+        if self.deleted:
+            self.deleted = False
+            self.save()
```
AlanCoding: Where does `soft_restore` get used?

awx/main/models/inventory.py
```diff
@@ -820,9 +820,47 @@ def _get_related_jobs(self):
...
+
+    def get_absolute_url(self, request=None):
+        return reverse('api:host_metric_detail', kwargs={'pk': self.pk}, request=request)
+
+    def soft_delete(self):
+        if not self.deleted:
+            self.deleted_counter = (self.deleted_counter or 0) + 1
+            self.last_deleted = now()
+            self.deleted = True
+            self.save()
```
AlanCoding: I do really strongly prefer to use `update_fields`, since it will avoid any conflicts with other updates going on. In this case, I don't think any other process even changes these deletion-related fields, which is perfect. Requests run in a transaction. It'll be great to never have this edit a value that the task system also edits.

awx/main/models/inventory.py
slemrmartin: Actually nowhere. It's just my feeling that the model could have both methods. 
Restoring is done only through "bulk" update atm.

awx/main/models/inventory.py
slemrmartin: I have no problem to change it to `update_fields`. 
I don't understand the rest of your comment, if it's just comment, or suggestion :)

awx/main/models/events.py
slemrmartin: I changed batch_size from 100 -> 1000 in `bulk_create`, but decreased from 'unlimited' -> 1000 in `update`. 
So I can decrease back the `bulk_create`, it's the main concern of @AlanCoding if I understand correctly. 
This optimization is not directly subject of this PR. Moreover this testing can't be done in this repo's tests (because performance testing against sqlite3 doesn't make much sense)

awx/main/models/inventory.py
slemrmartin: @AlanCoding in current work, renaming is not directly reflected in HostMetrics. But customer can (soft-)delete HostMetric manually (through API/UI) or there will be cleanup scheduled task (in next PR) which will soft-delete records after certain time with no updates

AlanCoding:
These new views are not linked at `/api/v2/`. What's the ultimate intention? Is this meant to be something which users can find by a link?

awx/main/models/inventory.py
```diff
@@ -820,9 +820,47 @@ def _get_related_jobs(self):
...
+
+    def get_absolute_url(self, request=None):
+        return reverse('api:host_metric_detail', kwargs={'pk': self.pk}, request=request)
+
+    def soft_delete(self):
+        if not self.deleted:
+            self.deleted_counter = (self.deleted_counter or 0) + 1
+            self.last_deleted = now()
+            self.deleted = True
+            self.save(update_fields=['deleted', 'deleted_counter', 'last_deleted'])
```
AlanCoding, line 845: If the thing is already deleted, the API gives a somewhat confusing response, which is indistinguishable from a successful deletion. Not sure what to do about that... either increase the deletion counter anyway or give a 400 error?

AlanCoding:
I put up suggested changes at https://github.com/slemrmartin/awx/pull/1

slemrmartin:
> These new views are not linked at `/api/v2/`. What's the ultimate intention? Is this meant to be something which users can find by a link?

@AlanCoding What do you mean by that? These views are accessible only at `/api/v2/host_metric.*`

AlanCoding:
@slemrmartin if you start at `/api/` then you find a link to `/api/v2/`, and from there you get a link to a lot of views at the level of `/api/v2/**/`, like `/api/v2/metrics/` (for an example that's not a resource list view). The views you added here at in the same URL pattern of `/api/v2/**/` but they are not included in the v2 links, like:

https://github.com/ansible/awx/blob/23a34c5dc9c0c1c48e88015dff9e9440cb55a802/awx/api/views/root.py#L99

awx/main/models/inventory.py
slemrmartin: deletion counter means change from "it was deleted" -> "it is deleted", so that's not this case. 

I'm just for the approach like playbooks do - make operation idempotent. 

If it is already deleted, then there is just no action, it's not an error (moreover when awx is designed by "mass deleting illusion" on UI, but APi is called x-times per record - it's pattern used in awx, not invented here)

I don't see a reason, why user needs to distinguish it.

slemrmartin:
@AlanCoding I implemented suggested changes, except:
- I haven't removed the "show_capabilities", as you posted, it's for UI and it seems you assumed there won't be UI
- I kept the idempotency of soft-delete (replied above)

awx/main/models/inventory.py
```diff
@@ -820,9 +820,47 @@ def _get_related_jobs(self):
...
+
+    def soft_restore(self):
+        if self.deleted:
+            self.deleted = False
+            self.save(update_fields=['deleted'])
```
AlanCoding, line 850: Looks like `soft_restore` isn't being used anywhere yet. Will it?

awx/main/models/inventory.py
```diff
@@ -820,9 +820,47 @@ def _get_related_jobs(self):
...
+
+
+class HostMetricSummaryMonthly(models.Model):
+    """
+    HostMetric summaries computed by scheduled task <TODO> monthly
+    """
+
+    date = models.DateField(unique=True)
+    license_consumed = models.BigIntegerField(default=0, help_text=_("How much unique hosts are consumed from the license"))
```
AlanCoding: --> How many unique hosts

awx/main/models/inventory.py
```diff
@@ -820,9 +820,47 @@ def _get_related_jobs(self):
...
+    license_capacity = models.BigIntegerField(default=0, help_text=_("'License capacity as max. number of unique hosts"))
+    hosts_added = models.IntegerField(default=0, help_text=_("How many hosts were added in the associated month, consuming more license capacity"))
```
AlanCoding, line 861: specify whether this includes deleted hosts or not

awx/main/models/inventory.py
```diff
@@ -820,9 +820,47 @@ def _get_related_jobs(self):
...
+    hosts_deleted = models.IntegerField(default=0, help_text=_("How many hosts were deleted in the associated month, freeing the license capacity"))
+    indirectly_managed_hosts = models.IntegerField(default=0, help_text=("Manually entered number indirectly managed hosts for a certain month"))
```
AlanCoding, line 863: If this is a non-goal for the near-term, could we remove the field for now and add it in a migration later?

awx/main/models/inventory.py
```diff
@@ -820,9 +820,47 @@ def _get_related_jobs(self):
 
 
 class HostMetric(models.Model):
-    hostname = models.CharField(primary_key=True, max_length=512)
+    hostname = models.CharField(unique=True, max_length=512)
     first_automation = models.DateTimeField(auto_now_add=True, null=False, db_index=True, help_text=_('When the host was first automated against'))
     last_automation = models.DateTimeField(db_index=True, help_text=_('When the host was last automated against'))
+    last_deleted = models.DateTimeField(null=True, db_index=True, help_text=_('When the host was last deleted'))
+    automated_counter = models.IntegerField(default=0, help_text=_('How many times was the host automated'))
```
AlanCoding: @gamuniz claims he has seen job counters go over 2 billion, so this should rightfully be a BigIntegerField

awx/main/models/inventory.py
```diff
@@ -820,9 +820,47 @@ def _get_related_jobs(self):
...
+    deleted_counter = models.IntegerField(default=0, help_text=_('How many times was the host deleted'))
+    deleted = models.BooleanField(
+        default=False, help_text=_('Boolean flag saying whether the host is deleted and therefore not counted into the subscription consumption')
+    )
+    used_in_inventories = models.IntegerField(null=True, help_text=_('How many inventories contain this host'))
```
AlanCoding, line 832: I know it's a TODO to compute this in a background task, but I have some concern about that. You're planning on introducing a _monthly_ task to compute the monthly summaries but it's unclear if a monthly job to update this field would be sufficient. I'd hate to have to introduce another more frequent background task.

awx/main/models/inventory.py
Ladas: @AlanCoding what formulation would you like here? This just counts how many were added in certain month (regardless of current state). The deleted are counted separately for a month, then the total monthly addition is computed as added+deleted.

awx/main/models/inventory.py
Ladas: @AlanCoding ideally we'd keep this in. I want to do a prototype for PMs with just a textarea where we let customers fill this in. With the DB change here it can be easily backported.

awx/main/models/inventory.py
Ladas: @AlanCoding yeah. Think that all the background jobs are ok to be done monthly. I don't think that subscription consumption reporting and adjusting will be operation that is done very often. This field is just for being able to order the list of hosts.

awx/main/models/inventory.py
Ladas: Might be, yeah. If we'll introduce filtering to show also deleted hosts, having the restore operation there makes sense.

awx/main/models/inventory.py
Ladas: @AlanCoding  Interesting, I'd guess update_fields and .save() must be equvivalent? No 2 processes can touch the same in-memory structure of this (unless we have some badly done multithreading there), so in the end both will just generate a SQL query and send it. (with last update winning)

AlanCoding:
I put up a suggestion at https://github.com/slemrmartin/awx/pull/2

awx/main/models/inventory.py
AlanCoding: that's a valid point, we can't backport migrations, so..

awx/main/models/inventory.py
AlanCoding: then that answers the question - that deleted hosts are still computed. The drawback which may be desirable to document is the fact that hosts automated-against and deleted within the time frame of a month will still be reflected in this count. That might be a mouthful, but it sounds like it could be relevant for the intended use.

AlanCoding (approved):
Great work LGTM

PR #13574, opened 2023-02-15
Make docker-compose command configurable in Makefile

sinuscosinustan:
##### SUMMARY
docker-compose v1 is EOL since April 2022 and hasn't received any updates since May 2021. docker compose v2 is a complete rewrite in Go which acts as a plugin for the main docker application. The syntax is the same, but only the `compose` command differs.

This commit adds the ability to override the default `docker-compose` command using `make DOCKER_COMPOSE='docker compose'`.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - Other

##### AWX VERSION
```
21.12.1.dev1+g24a0a5dadb
```

##### ADDITIONAL INFORMATION
I know that we could alias `docker-compose` to `docker compose` in the shell however, docker compose v2 will become the default at some time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 PYTHON ?= python3.9
+DOCKER_COMPOSE ?= docker-compose
 OFFICIAL ?= no
 NODE ?= node
 NPM_BIN ?= npm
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -509,20 +509,20 @@ docker-compose-sources: .git/hooks/pre-commit
 
 
 docker-compose: awx/projects docker-compose-sources
-	docker-compose -f tools/docker-compose/_sources/docker-compose.yml $(COMPOSE_OPTS) up $(COMPOSE_UP_OPTS) --remove-orphans
+	docker compose -f tools/docker-compose/_sources/docker-compose.yml $(COMPOSE_OPTS) up $(COMPOSE_UP_OPTS) --remove-orphans
```
shanemcd: On my (relatively) new system, I get:

```
$ docker compose
docker: 'compose' is not a docker command.
```

What version of Docker did this get added? I'm not really in favor of making a change that requires everyone to go update something for not a whole lot of gain.

sinuscosinustan: I'm running Docker `20.10.21` but NixOS seems to install the compose plugin by default. According to the [Docker Compose Docs](https://docs.docker.com/compose/install/) you have to install it on your own.

Besides that, the AWX docs already says that [the Docker Compose Plugin a prerequisite](https://github.com/ansible/awx/blob/devel/tools/docker-compose/README.md?plain=1#L38) additionally to the deprecated v1 version. Maybe by accident?

sinuscosinustan: >  I'm not really in favor of making a change that requires everyone to go update something for not a whole lot of gain.

I can totally understand the reason for that. If this is a complete breaking-change, then I'd be cool to write a "if you use compose v2, please set up an alias" note or whatever instead (even though I'm not a big fan of having to build workarounds because it uses a deprecated / EOL tooling).

fosterseth: maybe have the default be `docker-compose` in a make variable at the top of the makefile, and users can override it with `docker compose` if needed

sinuscosinustan: > maybe have the default be `docker-compose` in a make variable at the top of the makefile, and users can override it with `docker compose` if needed

Great idea! Thanks Seth!

I've applied your suggestion and adjusted the commit message too.

TheRealHaoLiu (approved):

TheRealHaoLiu:
@tomsiewert thank you for your contribution, congratulation to your first contribution to the AWX project!

I approved the PR and enabled auto-merge when the CI complete the PR will be merged

Issue #4292, opened 2019-07-09
RFE: Allow assigning an organization IG to a job template without requiring org admin


Ompragash:
##### ISSUE TYPE

 - Feature Idea

##### COMPONENT NAME

 - API
 - UI

##### SUMMARY

Currently in AWX only System Admin/Org Admin can assign Instance Group but needs new role i.e. "Instance Group Admin" role and this role will help a user to assign existing instance group in a Job Template.

wenottingham:
This seems the wrong way to do this, since the roles would likely be on the IGs, and you're not administrating *them*.

ulysses-31:
It would be great if users could select instance groups assigned to their organization to their job templates (without granting them Org Admin). Perhaps via a new Org/Role or enhancement of Org/Job Template Admins?

wenottingham:
see also #2785

wenottingham:
retitled to reflect the problem that needs solved

mikeycmccarthy:
Also interested in this one! It seems a funny one in terms of what the permission is on. A job template admin can't assign the instance group, but is that because they can't 'use' the instance group effectively?

schmandforke:
+1

williamhanseneric:
+1

AlanCoding:
The challenge I see here is that the `InstanceGroup` model is not directly scoped to a single organization. When a resource has `admin_role`, the standard pattern is that `organization.admin_role` becomes a parent role for it, assuming it has an `organization` ForeignKey. Instance groups do not.

Giving permissions via a many-to-many to organizations while also using that for fallback rules introduced some debt that we're going to have to resolve here. Depending on the big-picture direction from @ffirg and others, we may _introduce_ a new ForeignKey `organization` field. I don't like that idea, because it's an addition to the data model which isn't 100% necessary. The alternative is that we have `InstanceGroup.admin_role` just have the system admin role as the parent role. This seems expedient to me.

In any case, we will require additional special data migration steps

1. Any organization admin who used to get permission to an instance group via the many-to-many relationship will probably need to be added to the instance group `admin_role` explicitly (then the hierarchy rebuilt, standard stuff)
2. We will need to create roles for the instance groups from before the migration. There are pretty standard patterns for this as well.

PR #13584, opened 2023-02-16
Add instance groups roles

gamuniz:
##### SUMMARY

Related to: https://github.com/ansible/awx/issues/4292

The changes made will do the following:

- Create role permissions for instance group objects(read,use,admin)
- remove the organization admin access to the instance groups  which was  granted by way of special logic
- migrate use_role to organization admins which had instance groups associated with their org
- adds an access_list endpoint to the instance group
- adds new parent class to InstanceGroup to allow accessible_objects
- adds some basic testing
- adds UI for user/team to allow assignment
- changes routes to expose instance_groups to all users

##### ISSUE TYPE

 
 - New or Enhanced Feature
 

##### COMPONENT NAME

 - API

##### AWX VERSION

```
make VERSION
awx: 21.11.1.dev75+g5612a548f2

```

##### ADDITIONAL INFORMATION

```

```


awx/main/access.py
```diff
@@ -588,17 +588,41 @@ def can_delete(self, obj):
...
 
     def filtered_queryset(self):
-        return InstanceGroup.objects.filter(organization__in=Organization.accessible_pk_qs(self.user, 'admin_role')).distinct()
+        if self.user.admin_of_organizations.exists():
+            return InstanceGroup.objects.filter(organization__in=Organization.accessible_pk_qs(self.user, 'admin_role')).distinct()
+        return self.model.accessible_objects(self.user, 'read_role')
+
+    @check_superuser
+    def can_use(self, obj):
+        return self.user in obj.use_role
```
AlanCoding, line 612: You have a criteria in `filtered_queryset`, which effectively sets "read" permission. This criteria allows org admins to see instance groups associated with their org (by a m2m relationship).

The point of this comment is to point out that these users (org admins) will not gain "use" permission, because the `use_role` here only has the system admin role as a parent role, not any org admin roles. So these users will be able to see those IGs, but will not be able to associate them to a JT, for example. I don't know if this is a problem or not.

awx/main/access.py
```diff
@@ -588,17 +588,41 @@ def can_delete(self, obj):
...
 
 class InstanceGroupAccess(BaseAccess):
+    """
+    I can see Instance Groups when I am:
+       - a superuser(system administrator)
+       - an organization admin
+       - at least read_role on the instance group
+    I can edit Instance Groups when I am:
+       - a superuser
+       - an organization admin
```
AlanCoding: current logic does not allow org admins to edit and Instance Group.

awx/main/access.py
```diff
@@ -588,17 +588,41 @@ def can_delete(self, obj):
...
+       - admin role on the Instance group
+    I can add/delete Instance Groups:
+       - a superuser(system administrator)
+    """
```
AlanCoding, line 602: If you're doing this doc string, you should describe "use" permission, which is permission to attach an IG to a JT and some other stuff.

awx/main/access.py
```diff
@@ -588,17 +588,41 @@ def can_delete(self, obj):
...
+
     model = InstanceGroup
     prefetch_related = ('instances',)
 
     def filtered_queryset(self):
-        return InstanceGroup.objects.filter(organization__in=Organization.accessible_pk_qs(self.user, 'admin_role')).distinct()
+        if self.user.admin_of_organizations.exists():
+            return InstanceGroup.objects.filter(organization__in=Organization.accessible_pk_qs(self.user, 'admin_role')).distinct()
```
AlanCoding: This `return` would not respect the case that a user is an org admin, but has been granted access to an instance group which is not associated with their organization. They would see IGs associated with their org, but not those they got explicit access to.

I'm not saying I _want_ them to, because this would complicate querysets significantly, and that's really something we would like to avoid.

awx/main/access.py
gamuniz: Yes that is correct, I'll fix that up

awx/main/access.py
gamuniz: Yep I clarified that language as well. I also added that org admins can `use` these instance groups by way of special logic handled at the inventory/jt access level to keep backward compatibility.

awx/main/access.py
gamuniz: I think I see what you're saying, I'm tempted to add a not statement to make it work like you'd expect

awx/main/access.py
gamuniz: There is special logic around both inventory and jt can_attach that handles this case. I've left that logic in for backward compatibility.

awx/main/access.py
```diff
@@ -588,17 +588,44 @@ def can_delete(self, obj):
...
+    """
+    I can see Instance Groups when I am:
+       - a superuser(system administrator)
+       - an organization admin to which an instance group is associated
+       - at least read_role on the instance group
+    I can edit Instance Groups when I am:
+       - a superuser
+       - admin role on the Instance group
+    I can add/delete Instance Groups:
+       - a superuser(system administrator)
```
AlanCoding, line 599: I'm not sure why we should continue the special superuser criteria for deletion after the `admin_role` has been introduced. Certainly, the special IGs like controlplane must be protected, but absent that case, we should defer to the standard logic, which would defer to `can_admin`. I would prefer to see a call to `super()` in this case.

AlanCoding:
I see that the UI gains the ability to _grant_ a user access to an instance group here. That's good progress, but how can you see who has access? The consistent answer would be to present an `/api/v2/instance_groups/42/access_list/` endpoint and hook it up the same way other resources are.

That's a loaded request, because if you do this org admins won't really show correctly. That's why my prior suggestion was to give those users this access through explicit membership via a data migration.

Both of these things are going to be more work right here, but it would result in a consistent user experience where instance groups behave the same as other role-possessing resources.

gamuniz:
> 

great idea and I had forgotten about the access_list
I am adding it in here:

```
{
            "id": 2,
            "type": "user",
            "url": "[/api/v2/users/2/](https://fedora.local:8043/api/v2/users/2/)",
            "related": {
                "teams": "[/api/v2/users/2/teams/](https://fedora.local:8043/api/v2/users/2/teams/)",
                "organizations": "[/api/v2/users/2/organizations/](https://fedora.local:8043/api/v2/users/2/organizations/)",
                "admin_of_organizations": "[/api/v2/users/2/admin_of_organizations/](https://fedora.local:8043/api/v2/users/2/admin_of_organizations/)",
                "projects": "[/api/v2/users/2/projects/](https://fedora.local:8043/api/v2/users/2/projects/)",
                "credentials": "[/api/v2/users/2/credentials/](https://fedora.local:8043/api/v2/users/2/credentials/)",
                "roles": "[/api/v2/users/2/roles/](https://fedora.local:8043/api/v2/users/2/roles/)",
                "activity_stream": "[/api/v2/users/2/activity_stream/](https://fedora.local:8043/api/v2/users/2/activity_stream/)",
                "access_list": "[/api/v2/users/2/access_list/](https://fedora.local:8043/api/v2/users/2/access_list/)",
                "tokens": "[/api/v2/users/2/tokens/](https://fedora.local:8043/api/v2/users/2/tokens/)",
                "authorized_tokens": "[/api/v2/users/2/authorized_tokens/](https://fedora.local:8043/api/v2/users/2/authorized_tokens/)",
                "personal_tokens": "[/api/v2/users/2/personal_tokens/](https://fedora.local:8043/api/v2/users/2/personal_tokens/)"
            },
            "summary_fields": {
                "direct_access": [
                    {
                        "role": {
                            "id": 38,
                            "name": "Use",
                            "description": "Can use the instance group in a job template",
                            "team_id": 1,
                            "team_name": "test",
                            "team_organization_name": "Default",
                            "resource_name": "default",
                            "resource_type": "instance_group",
                            "related": {
                                "instance_group": "[/api/v2/instance_groups/2/](https://fedora.local:8043/api/v2/instance_groups/2/)"
                            },
                            "user_capabilities": {
                                "unattach": false
                            }
                        },
                        "descendant_roles": [
                            "read_role",
                            "use_role"
                        ]
                    }
                ],
                "indirect_access": []
```

awx/main/access.py
AlanCoding: I take it back, creation should be congruent with deletion (not admin).

awx/main/access.py
AlanCoding: With the migration logic you added, I think this isn't needed anymore?

awx/main/access.py
```diff
@@ -1675,7 +1698,7 @@ def can_attach(self, obj, sub_obj, relationship, data, skip_sub_obj_read_check=F
         if relationship == "instance_groups":
             if not obj.organization:
                 return False
-            return self.user.can_access(type(sub_obj), "read", sub_obj) and self.user in obj.organization.admin_role
+            return self.user in sub_obj.use_role
```
AlanCoding: My mind is going some accounting here - "How many objects can we attach an IG to?" I think the answer is 3.

For each one of those I expect to see the `use_role` check in `can_attach`. Right now I see 2 in this change.

So now working backwards... we attach IGs to JTs, organizations, inventories.

I think you still need to update organization access.

gamuniz: I left the organization access alone because a regular user shouldn't be able to change an organization anyway. If you would like I can try adjusting it but I didn't want to make any changes that would fundamentally change the rbac.

AlanCoding: Okay, I'm not sure, I need to talk that out.

https://github.com/ansible/awx/blob/60356afeec61ad8d81ba5e7cfb6ca1545fbd2d90/awx/main/access.py#L864

https://github.com/ansible/awx/blob/60356afeec61ad8d81ba5e7cfb6ca1545fbd2d90/awx/main/access.py#L870-L873

Here, `obj` is the organization object and `sub_obj` is the instance group. The request is to add an instance group to the organization's instance group list, which is a fallback for jobs (does not delegate IG permission anymore).

I'm not seeing a sufficiently strong justification for the new pattern you're developing here, which would suggest a requirement of "admin" to `obj` (org) and "use" to `sub_obj` (IG).

awx/main/access.py
```diff
@@ -934,7 +957,7 @@ def can_run_ad_hoc_commands(self, obj):
 
     def can_attach(self, obj, sub_obj, relationship, *args, **kwargs):
         if relationship == "instance_groups":
-            if self.user.can_access(type(sub_obj), "read", sub_obj) and self.user in obj.organization.admin_role:
+            if self.user in sub_obj.use_role:
```
AlanCoding: Inventory - you're attaching an IG to the inventory. In that case you should need admin permission to the inventory, and on first reading here I'm not seeing that.

awx/main/access.py
```diff
@@ -588,17 +588,40 @@ def can_delete(self, obj):
...
+       - a superuser(system administrator)
+       - at least read_role on the instance group
+    I can edit Instance Groups when I am:
+       - a superuser
+       - admin role on the Instance group
+    I can add/delete Instance Groups:
+       - a superuser(system administrator)
+    I can use Instance Groups when I have:
+       - use_role on the instance group
+       - admin_role on the instance group
```
AlanCoding: I don't see a need to specify this, since admin implies use permission.

AlanCoding (approved):
This will still need some work in terms of docs and testing, but the main API development looks like it has pretty well wrapped up.

AlanCoding:
Wait, I forgot about permissions with providing instance groups as a job prompt on launch, and ask for @john-westcott-iv to help direct that.

obaranov:
@gamuniz @AlanCoding I want to double check the org admin behavior now. So if the user has just an Org Admin role, they won't see any instance groups, including the instance group(s) associated with the organization, right?

And to see the org instance groups user should be assigned to the "global" ig role now?.

awx/main/models/ha.py
```diff
@@ -379,6 +384,24 @@ class InstanceGroup(HasPolicyEditsMixin, BaseModel, RelatedJobsMixin):
...
+        ]
+    )
+    use_role = ImplicitRoleField(
+        parent_role=[
+            'admin_role',
+        ]
+    )
+    read_role = ImplicitRoleField(
+        parent_role=[
+            'singleton:' + ROLE_SINGLETON_SYSTEM_AUDITOR,
```
AlanCoding, line 399: You shouldn't really need this parent here, because it has admin_role as a parent, which inherits from system admin. No harm, it's just redundant.

awx/ui/src/screens/InstanceGroup/ContainerGroupEdit/ContainerGroupEdit.js
```diff
@@ -21,9 +21,9 @@ function ContainerGroupEdit({ instanceGroup }) {
     result: initialPodSpec,
   } = useRequest(
     useCallback(async () => {
-      const { data } = await InstanceGroupsAPI.readOptions();
-      return data.actions.POST.pod_spec_override.default;
-    }, []),
+      const { data } = await InstanceGroupsAPI.readInstanceGroupOptions(instanceGroup.id);
+      return data.actions.PUT.pod_spec_override.default;
```
AlanCoding, line 27: :+1:  Nice. I understood the problem this addresses - that the UI will equate the ability to add an instance group to the ability to edit an instance group (and its pod spec). This is the form of solution I was anticipating for that problem, and is ideal. Now permissions are specific to the object, so assumptions have to change.

gamuniz:
@AlexSCorey I might need to pick your brain on the failure. but everything else is 99% done. Need to separate out the migration into 2 files.

awx/main/access.py
```diff
@@ -588,17 +588,39 @@ def can_delete(self, obj):
...
+    def can_use(self, obj):
+        return self.user in obj.use_role
 
     def can_add(self, data):
         return self.user.is_superuser
 
+    @check_superuser
     def can_change(self, obj, data):
-        return self.user.is_superuser
+        return self.can_admin(obj)
```
john-westcott-iv, line 619: Is this an existing pattern? If not should we just change this to can_change? since the default can_admin just calls can_change and we aren't overriding can_admin in the model.

awx/main/migrations/0176_instance_group_admin_migration.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Migration(ActivityStreamDisabledMigration):
+    dependencies = [
+        ('main', '0175_instance_group_role_addition'),
+    ]
+    operations = [
+        migrations.RunPython(migration_utils.set_current_apps_for_migrations),
+        migrations.RunPython(rbac.create_roles),
+        migrations.RunPython(oamigrate.migrate_org_admin_to_use),
```
john-westcott-iv: Why not just add the function into this file?

john-westcott-iv: Have you tried reversing this migration? If it can't do it on its own and its something we wanted to support, there is a way to tell Django how to reverse a migration see `0170_node_and_link_state.py` for details if you are interested in that.

gamuniz: for this we don't really support backing out migrations so its probably okay.

awx/main/access.py
gamuniz: yeah I just changed it from system admin to object admin. The pattern for this was already determined.
its likely out of scope for this change as its tech debt and we should limit the impact of this PR.

kdelee:
https://github.com/ansible/awx/pull/13462 merged so we need to fixup allowing promptable instance groups in the bulk API job launch serializer. We left some comments as guiding crumbs but feel free to reach out when you want to fix that up

obaranov:
I tried to rebase this branch on top of devel. And probably migraration should be now fixed. I got this error: 
```

tools_postgres_1 | 2023-03-14 12:24:53.884 UTC [124] ERROR:  relation "conf_setting" does not exist at character 158
tools_postgres_1 | 2023-03-14 12:24:53.884 UTC [124] STATEMENT:  SELECT "conf_setting"."id", "conf_setting"."created", "conf_setting"."modified", "conf_setting"."key", "conf_setting"."value", "conf_setting"."user_id" FROM "conf_setting" WHERE ("conf_setting"."key" = 'LOG_AGGREGATOR_LEVEL' AND "conf_setting"."user_id" IS NULL) ORDER BY "conf_setting"."id" ASC LIMIT 1
tools_awx_1 | 2023-03-14 12:24:53,876 WARNING  [-] awx.main.dispatch.periodic periodic beat started
tools_postgres_1 | 2023-03-14 12:24:53.886 UTC [124] ERROR:  relation "conf_setting" does not exist at character 158
tools_postgres_1 | 2023-03-14 12:24:53.886 UTC [124] STATEMENT:  SELECT "conf_setting"."id", "conf_setting"."created", "conf_setting"."modified", "conf_setting"."key", "conf_setting"."value", "conf_setting"."user_id" FROM "conf_setting" WHERE ("conf_setting"."key" = 'LOG_AGGREGATOR_ENABLED' AND "conf_setting"."user_id" IS NULL) ORDER BY "conf_setting"."id" ASC LIMIT 1
tools_postgres_1 | 2023-03-14 12:24:53.890 UTC [125] ERROR:  relation "conf_setting" does not exist at character 158
tools_postgres_1 | 2023-03-14 12:24:53.890 UTC [125] STATEMENT:  SELECT "conf_setting"."id", "conf_setting"."created", "conf_setting"."modified", "conf_setting"."key", "conf_setting"."value", "conf_setting"."user_id" FROM "conf_setting" WHERE ("conf_setting"."key" = 'LOG_AGGREGATOR_LEVEL' AND "conf_setting"."user_id" IS NULL) ORDER BY "conf_setting"."id" ASC LIMIT 1
tools_postgres_1 | 2023-03-14 12:24:53.891 UTC [125] ERROR:  relation "conf_setting" does not exist at character 158
tools_postgres_1 | 2023-03-14 12:24:53.891 UTC [125] STATEMENT:  SELECT "conf_setting"."id", "conf_setting"."created", "conf_setting"."modified", "conf_setting"."key", "conf_setting"."value", "conf_setting"."user_id" FROM "conf_setting" WHERE ("conf_setting"."key" = 'LOG_AGGREGATOR_ENABLED' AND "conf_setting"."user_id" IS NULL) ORDER BY "conf_setting"."id" ASC LIMIT 1
tools_postgres_1 | 2023-03-14 12:24:53.913 UTC [127] ERROR:  relation "main_schedule" does not exist at character 560
tools_postgres_1 | 2023-03-14 12:24:53.913 UTC [127] STATEMENT:  SELECT "main_schedule"."id", "main_schedule"."created", "main_schedule"."modified", "main_schedule"."description", "main_schedule"."created_by_id", "main_schedule"."modified_by_id", "main_schedule"."inventory_id", "main_schedule"."char_prompts", "main_schedule"."extra_data", "main_schedule"."survey_passwords", "main_schedule"."execution_environment_id", "main_schedule"."unified_job_template_id", "main_schedule"."name", "main_schedule"."enabled", "main_schedule"."dtstart", "main_schedule"."dtend", "main_schedule"."rrule", "main_schedule"."next_run" FROM "main_schedule" ORDER BY "main_schedule"."next_run" DESC NULLS LAST, "main_schedule"."id" ASC
tools_awx_1 | 2023-03-14 12:24:53,914 ERROR    [-] awx.main.dispatch Postgres event consumer has not recovered in 40.00880026817322 s, exiting
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
tools_awx_1 |     return self.cursor.execute(sql, params)
tools_awx_1 | psycopg2.errors.UndefinedTable: relation "main_schedule" does not exist
tools_awx_1 | LINE 1: ...chedule"."rrule", "main_schedule"."next_run" FROM "main_sche...
```

tchellomello (approved):
:rocket:

PR #13585, opened 2023-02-16
[usage-collection] CLI for host usage collection

AparnaKarve:
https://issues.redhat.com/browse/AA-1576

##### SUMMARY

Extended the functionality of the `host_metrics` CLI as described in the ADR

It now supports -

```awx-manage host_metric --csv host_metric```
```awx-manage host_metric --csv host_metric_summary_monthly```

 The above options would output the contents of the CSV in the standard output

```awx-manage host_metric --tarball```

```awx-manage host_metric  --tarball --rows_per_file ``` 

The above options would produce a tarball

For e.g - 
```awx-manage host_metric  --tarball --rows_per_file 100000 ``` 

-----------------------

Sample output with the `--csv` option :
```
# awx-manage host_metric --csv host_metric
id,hostname,first_automation,last_automation,last_deleted,automated_counter,deleted_counter,deleted,used_in_inventories
1,localhost,2023-01-18 17:53:15.824452+00:00,2023-01-18 17:53:15.823961+00:00,,0,0,False,
3,awx_1,2023-01-18 18:28:40.421047+00:00,2023-01-18 19:12:34.620062+00:00,,0,0,False,
2,awx_1_copy,2023-01-18 18:32:57.004374+00:00,2023-01-18 18:32:57.003868+00:00,,0,0,False,
5,host1,2023-01-18 22:49:32.891346+00:00,2023-02-09 19:51:37.351248+00:00,,4,0,False,
4,host2,2023-01-18 22:49:32.891430+00:00,2023-02-09 19:51:37.351248+00:00,,4,0,False,
6,host2a,2023-02-02 20:14:48.108894+00:00,2023-02-09 19:51:37.351248+00:00,,3,0,False,
14,host4,2023-02-04 01:47:36.704301+00:00,2023-02-15 16:55:37.932600+00:00,,4,0,False,
15,host3,2023-02-04 01:47:36.704358+00:00,2023-02-15 16:55:37.932600+00:00,,4,0,False,
20,host9,2023-02-06 23:00:45.343491+00:00,2023-02-09 19:51:37.351248+00:00,,4,0,False,
21,host8,2023-02-06 23:00:45.343556+00:00,2023-02-09 19:51:37.351248+00:00,,4,0,False,
36,host5,2023-02-09 07:10:44.545754+00:00,2023-02-15 16:55:37.932600+00:00,,4,0,False,
37,host7,2023-02-09 07:10:44.545814+00:00,2023-02-15 16:55:37.932600+00:00,2023-02-15 16:54:57.964919+00:00,4,1,False,
38,host6,2023-02-09 07:10:44.545857+00:00,2023-02-15 16:55:37.932600+00:00,,4,0,False,
57,abc,2023-02-09 17:49:05.996232+00:00,2023-02-09 19:11:46.847399+00:00,2023-02-15 16:49:48.252720+00:00,1,1,True,
88,host10,2023-02-14 02:20:56.119105+00:00,2023-02-15 16:20:16.958107+00:00,,2,0,False,
```

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - CLI

##### AWX VERSION

```
awx: 21.12.1.dev1+g6ea81c4ec9
```

##### ADDITIONAL INFORMATION

All anonymization code that previously existed has been removed

~For the SHA256 anonymization, we could not completely leverage the existing `awx.main.utils.encryption` package because it looks like the `decrypt_value` function does not respond well if a custom salt is used during encryption.
For that reason, we had to write our own `encrypt_name` and `decrypt_name` (the way we have it in Analytics)
Note: While both functions, `encrypt_name` and `decrypt_name` are defined, for this CLI utility,  only `encrypt_name` has been used.~

### DIFF
--- a/awx/main/management/commands/host_metric.py
+++ b/awx/main/management/commands/host_metric.py
@@ -9,13 +9,14 @@ import tempfile
 import tarfile
 import csv
 
-PREFERRED_ROW_COUNT = 500000
+CSV_PREFERRED_ROW_COUNT = 500000
+BATCHED_FETCH_COUNT = 10000
 
 
 class Command(BaseCommand):
     help = 'This is for offline licensing usage'
 
-    def host_metric_queryset(self, result, offset=0, limit=PREFERRED_ROW_COUNT):
+    def host_metric_queryset(self, result, offset=0, limit=BATCHED_FETCH_COUNT):
         list_of_queryset = list(
             result.values(
                 'id',
@@ -32,7 +33,7 @@ class Command(BaseCommand):
 
         return list_of_queryset
 
-    def host_metric_summary_monthly_queryset(self, result, offset=0, limit=PREFERRED_ROW_COUNT):
+    def host_metric_summary_monthly_queryset(self, result, offset=0, limit=BATCHED_FETCH_COUNT):
         list_of_queryset = list(
             result.values(
                 'id',
@@ -67,22 +68,70 @@ class Command(BaseCommand):
 
             offset += len(list_of_queryset)
 
-    def csv_for_tar(self, temp_dir, type, filter_kwargs, single_header=False, rows_per_file=PREFERRED_ROW_COUNT):
+    def controlled_db_retrieval(self, type, filter_kwargs, offset=0, fetch_count=BATCHED_FETCH_COUNT):
+        if type == 'host_metric':
+            result = HostMetric.objects.filter(**filter_kwargs)
+            return self.host_metric_queryset(result, offset, fetch_count)
+        elif type == 'host_metric_summary_monthly':
+            result = HostMetricSummaryMonthly.objects.filter(**filter_kwargs)
+            return self.host_metric_summary_monthly_queryset(result, offset, fetch_count)
+
+    def write_to_csv(self, csv_file, list_of_queryset, always_header, first_write=False, mode='a'):
+        with open(csv_file, mode, newline='') as output_file:
+            try:
+                keys = list_of_queryset[0].keys() if list_of_queryset else []
+                dict_writer = csv.DictWriter(output_file, keys)
+                if always_header or first_write:
+                    dict_writer.writeheader()
+                dict_writer.writerows(list_of_queryset)
+
+            except Exception as e:
+                print(e)
+
+    def csv_for_tar(self, temp_dir, type, filter_kwargs, rows_per_file, always_header=True):
         for index, list_of_queryset in enumerate(self.paginated_db_retrieval(type, filter_kwargs, rows_per_file)):
             csv_file = f'{temp_dir}/{type}{index+1}.csv'
             arcname_file = f'{type}{index+1}.csv'
 
-            with open(csv_file, 'w', newline='') as output_file:
-                try:
-                    keys = list_of_queryset[0].keys() if list_of_queryset else []
-                    dict_writer = csv.DictWriter(output_file, keys)
-                    if not single_header or index == 0:
-                        dict_writer.writeheader()
-                    dict_writer.writerows(list_of_queryset)
+            first_write = True if index == 0 else False
 
-                except Exception as e:
-                    print(e)
+            self.write_to_csv(csv_file, list_of_queryset, always_header, first_write, 'w')
+            yield csv_file, arcname_file
 
+    def csv_for_tar_batched_fetch(self, temp_dir, type, filter_kwargs, rows_per_file, always_header=True):
+        csv_iteration = 1
+
+        offset = 0
+        rows_written_per_csv = 0
+        to_fetch = BATCHED_FETCH_COUNT
+
+        while True:
+            list_of_queryset = self.controlled_db_retrieval(type, filter_kwargs, offset, to_fetch)
+
+            if not list_of_queryset:
+                break
+
+            csv_file = f'{temp_dir}/{type}{csv_iteration}.csv'
+            arcname_file = f'{type}{csv_iteration}.csv'
+            self.write_to_csv(csv_file, list_of_queryset, always_header)
+
+            offset += to_fetch
+            rows_written_per_csv += to_fetch
+            always_header = False
+
+            remaining_rows_per_csv = rows_per_file - rows_written_per_csv
+
+            if not remaining_rows_per_csv:
+                yield csv_file, arcname_file
+
+                rows_written_per_csv = 0
+                always_header = True
+                to_fetch = BATCHED_FETCH_COUNT
+                csv_iteration += 1
+            elif remaining_rows_per_csv < BATCHED_FETCH_COUNT:
+                to_fetch = remaining_rows_per_csv
+
+        if rows_written_per_csv:
             yield csv_file, arcname_file
 
     def config_for_tar(self, options, temp_dir):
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: awx/main/management/commands/host_metric.py [hunk 1]
```diff
@@ -1,26 +1,196 @@
...
+
+    def config_for_tar(self, options, temp_dir):
+        config_json = json.dumps(config(options.get('since')))
+        config_file = f'{temp_dir}/config.json'
+        arcname_file = 'config.json'
+        with open(config_file, 'w') as f:
+            f.write(config_json)
+        return config_file, arcname_file
+
+    def encrypt_name(self, key, value):
```
Ladas: btw. the ADR had it needs Anonymization not Encryption (these are 2 different approaches, with anonymization being irreversible)

But it might be best to remove anonymization for now, we will need a proper design for how to deal with a drift if key is changed. (otherwise it will create chaos in the reporting)

AparnaKarve: So, remove the support for `--anonymized` for now?

Ladas: @AparnaKarve yeah that might be the best, we can try to solve it for 2.5 in general (so not just for CLI), with a plan of how to deal with the drift.

TheRealHaoLiu:
summary for the code review session:
- PR and functionality looks good 
- looks like pandas is not being used by anything critical and the csv library would be enough, we would like to avoid pulling in pandas if possible
- removing the anonymizing function due to comment from @Ladas

AparnaKarve:
Addressed above feedback.

Previously, `pandas` was used for CSV creation and for applying a column-wide anonymization function.
With the plan to remove anonymization for 2.4, it made sense to use the `csv` library (already part of awx) for CSV creation.

awx/main/management/commands/host_metric.py
```diff
@@ -33,17 +174,39 @@ def handle(self, *args, **options):
...
+        filter_kwargs_host_metrics_summary = {}
+        if since is not None:
+            filter_kwargs_host_metrics_summary['date__gte'] = since
+        if until is not None:
+            filter_kwargs_host_metrics_summary['date__lte'] = until
+
+        host_metric_row_count = HostMetric.objects.filter(**filter_kwargs).count()
+        host_metric_summary_monthly_row_count = HostMetricSummaryMonthly.objects.filter(**filter_kwargs_host_metrics_summary).count()
+
+        if (host_metric_row_count > PREFERRED_ROW_COUNT or host_metric_summary_monthly_row_count > PREFERRED_ROW_COUNT) and (
```
slemrmartin: I suggest to set the 500k as a default value (ADR is describing it too). The `--rows_per_file` can be optional customization.

awx/main/management/commands/host_metric.py
```diff
@@ -33,17 +174,39 @@ def handle(self, *args, **options):
...
         if options['json']:
-            list_of_queryset = list(result.values('hostname', 'first_automation', 'last_automation'))
-            json_result = json.dumps(list_of_queryset, cls=DjangoJSONEncoder)
-            print(json_result)
+            self.output_json(options, filter_kwargs)
+        elif options['csv']:
+            self.output_csv(options, filter_kwargs)
+        elif options['tarball']:
+            self.output_tarball(options, filter_kwargs, host_metric_row_count, host_metric_summary_monthly_row_count)
+        elif options['rows_per_file']:
```
slemrmartin: as this param could be just a customization for csv/tarball options, it shouldn't be in the if/else flow

awx/main/management/commands/host_metric.py
```diff
@@ -1,26 +1,167 @@
...
+                'date'
+            )[offset : offset + limit]
+        )
+
+        return list_of_queryset
+
+    def paginated_db_retrieval(self, type, filter_kwargs, offset=0, limit=PREFERRED_ROW_COUNT):
+        list_of_queryset = []
+        if type == 'host_metric':
+            result = HostMetric.objects.filter(**filter_kwargs)
```
slemrmartin, line 73: there could be also fine-grained db pagination in case of big rows_per_file settings (i.e. 10000)

This can produce `SELECT * from main_hostmetric WHERE ... LIMIT 500000`. That's quite a big number for db.

awx/main/management/commands/host_metric.py
```diff
@@ -33,17 +174,39 @@ def handle(self, *args, **options):
...
             filter_kwargs['last_automation__lte'] = until
 
-        result = HostMetric.objects.filter(**filter_kwargs)
+        filter_kwargs_host_metrics_summary = {}
+        if since is not None:
+            filter_kwargs_host_metrics_summary['date__gte'] = since
+        if until is not None:
+            filter_kwargs_host_metrics_summary['date__lte'] = until
+
+        host_metric_row_count = HostMetric.objects.filter(**filter_kwargs).count()
```
slemrmartin: these 2 db consuming queries can be omitted if there will be loop with increasing db offset and checked emptiness of retrieved queries.

slemrmartin (changes requested):
Overall it looks good.
The pagination parts and pagination param should be adjusted and it'll be ready

awx/main/management/commands/host_metric.py
AparnaKarve: The default value is already set to 500k.
We can set `--rows_per_file` to be an optional customization -- that's fine.

Note: This will require a change in the ADR

awx/main/management/commands/host_metric.py
AparnaKarve: So `--rows_per_file`  is the parameter for regulating how many rows to fetch at any given time.



- `For awx-manage host_metric --csv --rows_per_file 10000`, 
we will keep fetching 10000 rows until we have fetched all rows and then display the csv in standard output

- `For awx-manage host_metric --tarball --rows_per_file 10000`, 
we will fetch 10000 rows and create the tarball with several individual CSV files where each CSV would have up to 10000 rows

- `For awx-manage host_metric --tarball `, 
we will fetch 500k rows in one go since `--rows_per_file` was not specified and create the tarball

awx/main/management/commands/host_metric.py
AparnaKarve: This should be doable.

AparnaKarve:
@Ladas @slemrmartin If the [above](https://github.com/ansible/awx/pull/13585#discussion_r1123414850) resolution looks OK, please update the ADR and I will make the requested changes.
Thanks.

awx/main/management/commands/host_metric.py
Ladas: I think ADR lists that as optional (stating 500k should be default)

AparnaKarve:
Addressed the above feedback related to -
- making `--rows_per_file` optional
- Not calling the `.count` queries for the 2 tables and instead relying on the empty queryset list to determine if we have fetched all the rows from the db

awx/main/management/commands/host_metric.py
slemrmartin: yes, it's a parameter how big should be the CSV file. 
But my concern is that if we default 500k, then *default* query to the database is `SELECT .. LIMIT 500k`, that sounds dangerous. So I'm suggesting to make independent db pagination and csv pagination

Ladas: Martin is correct, unless you'll do `COPY` directly to csv (that should be batched operation), this might blow up in memory. E.g. we do this in Analytics Exporter, COPY to csv without any explicit batching (postgres takes care of that), then batch reading of the csv to convert it to different format. (so we keep the memory low) That should be the most effective way to do it.


But paginated read from DB should work ok-ish for now. (something like hardcoded page of 10k rows we load from DB at once and append to the csv)

AparnaKarve: > So I'm suggesting to make independent db pagination and csv pagination

I had thought about this, but was trying to avoid this situation because two pagination schemes are just going to add some complexity.

I would have to evaluate this to make a decision as we do not want to overcomplicate the code.

------------------

The easiest approach here  would be to set default value of `--rows_per_file` to 10000

AparnaKarve: Added functionality for fetching rows in batches of 10k and creating CSVs with whatever is specified in `--rows_per_file`
Since 500k is a really large number (even for the postgres function), its best to use 10k-chunking approach.

(Also, the postgres function might force us to use raw SQL - would prefer to use django orm when possible)

slemrmartin (approved):
Looks good now, thanks :+1:

TheRealHaoLiu (approved):

Issue #5692, opened 2020-01-17
Inventory update based on project and not job


Giusepe92:
##### ISSUE TYPE

 - Bug Report

##### COMPONENT NAME

 - API
 - UI

##### SUMMARY
We are using AWX 9.1.1
We defined a project named : DEPLOY_DEV
for this project, we allowed branch override, and we have not specified any branch.

We defined a job template : DEPLOY_DEV_AUTO
this job template ask for branch on launch, which will point to the branch SCM where we have our playbook and our inventory.

We defined an invenotry named: INVENTORY_DEV
this inventory is supposed to point to the same branch SCM as job template on run.

Expected Behaviour : 
When we launch a job based on job template DEPLOY_DEV_AUTO and specify branch "V1" , DEPLOY_DEV and INVENTORY_DEV should be synchronized with branch "V1", then the job is launched

Actual Behaviour : 
When we launch a job based on job template DEPLOY_DEV_AUTO and specify branch "V1" , DEPLOY_DEV is synchronized with branch "V1",INVENTORY_DEV is synchronized with "master" (since the field branch in project is empty) . it fails, which is normal since it's looking for inventory in master branch.
We don't want to put anything in branch field for project since we have a new branch for every job, so a new inventory and a new playbook.

##### ENVIRONMENT
* AWX version: 9.1.1

AlanCoding:
Is this the same as https://github.com/ansible/awx/issues/5698? If it is, I just want to consolidate, and set priority.

Giusepe92:
Yes

linkeal:
We are hitting the exact same problem, also AWX 9.1.1. I somehow have the feeling that it is related to the fact that there is a "prompt on launch" checkbox missing in the inventory. At least I've also tried to create a workflow with an scm branch definition on launch which updates first the project and then the inventory and then executed the template and the same problem persists, that the project is correctly checked out from a branch but the inventory is checked out from master.

Would be really helpful to get this fixed because we would also like to use AWX for our CI/CD and not create a separate project/inventory/template for every branch.
Thanks!

hardeaux:
Same issue here - looking for best of both worlds where inventory/playbooks tracked in gitlab + ability for users to easily test their inventory branch changes with AWX before requesting merge. 

glad there is already an issue opened - please fix.

-thanks

idelsink:
We have the same issue, We have an inventory with multiple sources both from AWS and sourced from a project. 

[...]

PR #13587, opened 2023-02-16
Make Job scm_branch override propagate to Inventory Source update

fosterseth:
#### SUMMARY

Aims to solve https://github.com/ansible/awx/issues/5692

The job scm_branch override does not propagate to the inventory source. For example, the job will run based on project branch `foo`, but will use inventory hosts from branch `main`.

#### Fix

Allow inventory updates to sync hosts based on an scm_branch

**Requirements**:
1. Project must have branch override enabled
2. Inventory Source must have Update on Launch enabled
3. Job and InventorySource must be based on the same project

#### :construction: Major Caveat :construction: 

Shared inventories can clash, and a Job might end up using the wrong inventory hosts.

1. JobA and JobB both share InventorySource
2. JobA launches with scm_branch set to `foo`
3. JobB launches with scm_branch set to `bar`
4. Two inventory source updates will start, one to sync to `foo` and one to `bar`
5. Both inventory source updates will occur **before** either JobA or JobB runs.
6. Therefore, by the time JobA and JobB run, they will both use inventory from the `bar` branch.

##### ISSUE TYPE

##### COMPONENT NAME

 - API


fosterseth:
@ffirg There is no toggle in this solution, it just works behind the scenes. The only thing users need to ensure are the items listed under `Requirements` above.

the major caveat may be a deal breaker, but this work represents a least invasive way to handle this issue.

awx/main/models/inventory.py
```diff
@@ -1210,6 +1210,12 @@ class Meta:
         verbose_name=_('SCM Revision'),
         help_text=_('The SCM Revision from the Project used for this inventory update.  Only applicable to inventories source from scm'),
     )
+    scm_branch = models.CharField(
+        max_length=1024,
+        default='',
+        blank=True,
+        help_text=_('Branch to use in inventory update. Project default used if blank. Only allowed if project allow_override field is set to true.'),
```
AlanCoding, line 1217: to avoid ambiguity, I would set `editable=False`.

awx/main/scheduler/task_manager.py
```diff
@@ -394,8 +396,18 @@ def gen_dep_for_job(self, task):
...
+            # if inventory source uses the same project as task, and the last inventory update has a different scm_branch, we should set branch override
+            if (
+                inventory_source.source == "scm"
+                and inventory_source.source_project_id == task.project_id
+                and latest_inventory_update.scm_branch != task.scm_branch
+            ):
+                should_set_scm_branch = True
+            else:
+                should_set_scm_branch = False
+            if should_set_scm_branch or self.should_update_inventory_source(task, latest_inventory_update):
```
AlanCoding, line 409: > 2. Inventory Source must have Update on Launch enabled

Doesn't this criteria conflict with the requirement? I see `should_set_scm_branch` has no bearing on the InventorySource update-on-launch setting, and the `or` means that it being true will spawn an update, spawning updates even when update-on-launch is False, which would be bad.

I liked what you were doing before where you tested this condition in `should_update_inventory_source`, which could more naturally come after the update-on-launch check.

awx/main/scheduler/task_manager.py
```diff
@@ -283,11 +283,13 @@ def create_project_update(self, task, project_id=None):
...
         return project_task
 
-    def create_inventory_update(self, task, inventory_source_task):
+    def create_inventory_update(self, task, inventory_source_task, should_set_scm_branch=False):
         inventory_task = InventorySource.objects.get(id=inventory_source_task.id).create_inventory_update(_eager_fields=dict(launch_type='dependency'))
 
         inventory_task.created = task.created - timedelta(seconds=2)
         inventory_task.status = 'pending'
+        if should_set_scm_branch:
+            inventory_task.scm_branch = task.scm_branch
```
AlanCoding, line 292: Even if it's more verbose, I have a gut feeling that this method should evaluate whether or not to set scm_branch from what's already available in `task` and `inventory_source_task`.

BTW, the naming of `inventory_source_task` bothers me, right? It seems to be an InventorySource, so maybe we could just call it `inventory_source`?

awx/main/scheduler/task_manager.py
fosterseth: yeah i didn't want to repeat the logic for checking whether we should set scm_branch twice:
once in `should_update_inventory_source` and another time in `create_inventory_update`

this was my way of trying to consolidate it by moving it outside of should_update method

also if `update_on_launch` is not enabled, we hit a continue above so we don't even hit this logic. So in that way it is an implicit requirement, but for code readability this is kind of messy

AlanCoding: As I'm internally churning through this in my mind, I set aside a proposal to rename `should_update_inventory_source` to `inventory_update_passed_cache_timeout`. I can see that the method checks `if latest_inventory_update.inventory_source.update_on_launch is True`, which seems to be the 3rd time that check is done. To begin with `self.all_inventory_sources` is filtered to update-on-launch sources. Then we check here, and then again in that method.

I'm still trying to catch up... so onto my next question.

Let's say that the job has an `scm_branch` which is non-blank. In what cases should `create_inventory_update` _not_ set the `scm_branch`? Implication of my question is that it should always do that. With your enhancement here, it _seems_ like every inventory update dependency will have the same `scm_branch` as the job that triggered it. If this is always true, then `should_set_scm_branch` shouldn't be a parameter, because the only time it would be `False` would be when job's `scm_branch` is blank, in which case it doesn't matter if you do or don't set it.

AlanCoding: answering my own question - it should not inherit the `scm_branch` if the source is not "scm". But if it is scm source, then it should always get the scm_branch from the job, unless there's a corner case I'm missing.

AlanCoding: ah, no, there's also the same-project criteria.

AlanCoding (approved):

AlanCoding:
It's not very obvious, but this should force us to make UI changes to show scm_branch on the inventory update details page.

AlanCoding:
I have 2 tech debt items that I want to see addressed after this merges:

 - Just general cleanup of the `DependencyManger` and internal method signature updates - https://github.com/ansible/awx/pull/13533
 - What I described here https://github.com/ansible/awx/issues/13630, when we're in this kind of situation in the `DependencyManager`, it may spawn inventory updates even when the cache timeout is not expired (added by this PR), but it _should not spawn project updates_ in those cases, because they are not needed. They don't do anything. From a big-picture app consistency standpoint, I feel like I would want this as a user. If "extra" inventory updates are spawned as a special-case due to branch-override, then I would expect that no-op project updates are _avoided_ when they are rendered pointless by those same conditions.

I don't see how we do any of these before merging your PR, so I'm fine to let this go in.

fosterseth:
we decided to go a different direction with this to avoid risk outlined in the "Major Caveat" section

replacement PR https://github.com/ansible/awx/pull/13644

PR #12721, opened 2022-08-23
Handles workflow node schema inventory

db0:
##### SUMMARY

This makes the workflow_job_template handle the inventory passed to nodes in the schema properly

This is handled fine in workflow_job_template_node, but since workflow_job_template is using its own function, this functionality is missing. I simply copied the same approach from workflow_job_template_node and tested that it is working in my installation

Related #12720
##### ISSUE TYPE

Bug

##### COMPONENT NAME

Collection

##### AWX VERSION

```
I am using the Ansible Automation Platform 2.2.0 from RedHat, not AWX
```

Bug, Docs Fix or other nominal change

PR #13588, opened 2023-02-17
fix inventory prompt on launch for workflow nodes

sean-m-sullivan:
##### SUMMARY
I was mistaken on the fix for #12721, The previous bugfix does not account for the form of the export output from the awx cli command. It also does not update the documentation, I've reverted the previous change and corrected it. I've also updated tests in regards to this.

##### ISSUE TYPE
 - Bug

##### COMPONENT NAME
 - Collection

##### AWX VERSION

```
21.12.0
```

### DIFF
--- a/awx_collection/plugins/modules/workflow_job_template.py
+++ b/awx_collection/plugins/modules/workflow_job_template.py
@@ -377,6 +377,24 @@ options:
                       description:
                         - The organization of the credentials exists in.
                       type: str
+            inventory:
+              description:
+                - Inventory to be applied to job as launch-time prompts.
+              type: dict
+              suboptions:
+                name:
+                  description:
+                    - Name Inventory to be applied to job as launch-time prompts.
+                  type: str
+                organization:
+                  description:
+                    - Name of key for use in model for organizational reference
+                  type: dict
+                  suboptions:
+                    name:
+                      description:
+                        - The organization of the credentials exists in.
+                      type: str
             labels:
               description:
                 - Labels to be applied to job as launch-time prompts.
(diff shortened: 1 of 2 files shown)


awx_collection/plugins/modules/workflow_job_template.py
```diff
@@ -736,11 +750,12 @@ def create_workflow_nodes_association(module, response, workflow_nodes, workflow
                 'failure_nodes',
                 'credentials',
                 'labels',
+                'inventory',
                 'instance_groups',
             ):
                 # Extract out information if it exists
                 # Test if it is defined, else move to next association.
-                prompt_lookup = ['credentials', 'labels', 'instance_groups']
+                prompt_lookup = ['credentials', 'labels', 'inventory', 'instance_groups']
```
AlanCoding, line 758: This seems suspicious since these are all plural, many-to-many, relationships but inventory is not.

sean-m-sullivan: It's derived from the related field
image

sean-m-sullivan:
Fixed the PR check error by removing 2nd field of Docs fix.

awx_collection/plugins/modules/workflow_job_template.py
AlanCoding: I have to admit I have a really hard time getting through this (existing) code.

```python
                if association in workflow_node['related']:
                    id_list = []
                    lookup_data = {}
                    for sub_name in workflow_node['related'][association]:
```

If the JSON data is what we expect (like the image you post) then `association` iterates the keys of "related", starting with `"inventory"`, and then `workflow_node['related'][association]` would, first, give `"/api/v2/inventories/9/"` and then the first `sub_name` would be `"/"`. That's wrong... So I have a hard time reading through the rest of it.

sean-m-sullivan: arg, closing for now, will repoen when its ready, going to do a full writeup with all the options and double check everything.

PR #13590, opened 2023-02-17
Constructed inventory hosts

AlexSCorey:
##### SUMMARY
This adds constructed inventory host lists and their related details view
TYPE
- New or Enhanced Feature

##### COMPONENT NAME
 - UI


awx/ui/src/screens/Inventory/ConstructedInventory.js
```diff
@@ -21,9 +21,9 @@ import RelatedTemplateList from 'components/RelatedTemplateList';
 import { ResourceAccessList } from 'components/ResourceAccessList';
 import RoutedTabs from 'components/RoutedTabs';
 import ConstructedInventoryDetail from './ConstructedInventoryDetail';
-import ConstructedInventoryEdit from './ConstructedInventoryEdit';
+import ConstructedInventoryEdit from './InventoryEdit';
```
marshmalien: Can we change this back to `ConstructedInventoryEdit`? I don't think we'll use the regular inventory edit form for this.

marshmalien (commented):
Discussed with @AlexSCorey. Constructed inventory hosts should be read-only, so we are going to try to reuse the smart inventory host components since they're read-only by default.

marshmalien (approved):
LGTM!

PR #13591, opened 2023-02-17
Update templates for feature removals

AlanCoding:
##### SUMMARY
Previously, the help drop-down for the config view in the API browser was not working, because the template simply had not been moved from API v1 to v2.

This fixes that and deletes some other templates.

##### ISSUE TYPE

##### COMPONENT NAME
 - API


AlanCoding:
I was going to update the tests to avoid the test-specific `timeout` kwarg, but looking further into it, the fact that this is a unit test makes that extra difficult. I'm happy with this as it is.

gamuniz (approved):

PR #13596, opened 2023-02-20
Add HOST_PROJECTS_ROOT setting

rcarrillocruz:
##### SUMMARY

When doing a project update, the execution environment mounts project_path:project_path and cache_path:cache_path. 
This prevents from running a containerized receptor, as the project_path/cache_path on host may differ to the container, specially when running containerized receptor rootless.

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - Other

##### AWX VERSION

```
awx: 0.1.dev32846+gcacb0c9
```

### DIFF
--- a/awx/main/management/commands/provision_instance.py
+++ b/awx/main/management/commands/provision_instance.py
@@ -29,7 +29,7 @@ class Command(BaseCommand):
         parser.add_argument('--uuid', type=str, help="Instance UUID")
 
     def _register_hostname(self, hostname, node_type, uuid):
-        if not hostname:
+        if settings.IS_K8S:
             if not settings.AWX_AUTO_DEPROVISION_INSTANCES:
                 raise CommandError('Registering with values from settings only intended for use in K8s installs')
 
@@ -47,8 +47,10 @@ class Command(BaseCommand):
                 max_forks=settings.DEFAULT_EXECUTION_QUEUE_MAX_FORKS,
                 max_concurrent_jobs=settings.DEFAULT_EXECUTION_QUEUE_MAX_CONCURRENT_JOBS,
             ).register()
-        else:
+        elif hostname:
             (changed, instance) = Instance.objects.register(hostname=hostname, node_type=node_type, uuid=uuid)
+        else:
+            return
         if changed:
             print("Successfully registered instance {}".format(hostname))
         else:
--- a/awx/main/models/projects.py
+++ b/awx/main/models/projects.py
@@ -196,10 +196,22 @@ class ProjectOptions(models.Model):
             if not check_if_exists or os.path.exists(smart_str(proj_path)):
                 return proj_path
 
+    def get_host_project_path(self, check_if_exists=True):
+        local_path = os.path.basename(self.local_path)
+        if local_path and not local_path.startswith('.'):
+            proj_path = os.path.join(settings.HOST_PROJECTS_ROOT or settings.PROJECTS_ROOT, local_path)
+            if not check_if_exists or os.path.exists(smart_str(proj_path)):
+                return proj_path
+
     def get_cache_path(self):
         local_path = os.path.basename(self.local_path)
         if local_path:
             return os.path.join(settings.PROJECTS_ROOT, '.__awx_cache', local_path)
+
+    def get_host_cache_path(self):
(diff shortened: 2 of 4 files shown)


rcarrillocruz:
Alternatively, I guess we could have projects_root parameter added to get_project_path/get_cache_path, defaulting to settings.PROJECTS_ROOT

AlanCoding:
Do you have some more background on the motivation for the change? This will only apply for _containerized project updates_ and will still not affect containerized jobs or normal jobs. The notion of containerized receptor suggests some changes to the deployment...

rcarrillocruz:
@AlanCoding the motivation for the change is when running containerized receptor as rootless, in which the host path differs from the container path.

What I'm trying to accomplish is to have a regular non-root user to have host paths at let's say /home/myuser/aap as the base folder to contain the AWX paths for projects and job executions in the container (e.g. /var/lib/awx/projects, /var/lib/awx/job_execution).
Then when running with process isolation with podman, the volumes exposed are something along the lines of:

```
-v /home/myuser/aap/projects:/var/lib/awx/projects
-v /home/myuser/aap/job_execution/<job id>:/var/lib/awx/job_execution/<job_id>
```

By adding the code for HOST_PROJECTS_ROOT_PATH, I can get past the issues for the creation of projects folders.
However, I'm struggling to make it work for containerized job runs as you suggested.
I added some logic to 'try' to circumvent the job execution, by adding an AWX_ISOLATION_HOST_BASE_PATH setting and , if defined, replace the dirname of private_data_dir to AWX_ISOLATION_HOST_BASE_PATH but it's failing with:

```
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/jobs.py", line 593, in run
    res = receptor_job.run()
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/receptor.py", line 317, in run
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/receptor.py", line 368, in _run_internal
    transmitter_future.result()
  File "/usr/lib64/python3.9/concurrent/futures/_base.py", line 439, in result
    return self.__get_result()
  File "/usr/lib64/python3.9/concurrent/futures/_base.py", line 391, in __get_result
    raise self._exception
  File "/usr/lib64/python3.9/concurrent/futures/thread.py", line 58, in run
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/utils/common.py", line 1212, in wrapper_cleanup_new_process
    return func(*args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/receptor.py", line 453, in transmit
    ansible_runner.interface.run(streamer='transmit', _output=_socket.makefile('wb'), **self.runner_params)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/ansible_runner/interface.py", line 210, in run
    r = init_runner(**kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/ansible_runner/interface.py", line 56, in init_runner
    dump_artifacts(kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/ansible_runner/utils/__init__.py", line 196, in dump_artifacts
    raise ValueError('private_data_dir path is either invalid or does not exist')
ValueError: private_data_dir path is either invalid or does not exist
```

Sounds to me this can't be fixed on ansible/awx side, but rather a patch to ansible-runner since there is an assumption that private_data_dir will be the same in both the host and the container.
Thus maybe we would need a patch in which there are params exposed for host_private_data_dir and container_private_data_dir ?

Thoughts?

cc @shanemcd

rcarrillocruz:
I can circument the containerized job execution issue by setting AWX_ISOLATION_BASE_PATH to /tmp/job_execution and mounting the volume /tmp/job_execution:/tmp/job_execution in both receptor and automation-task-controller.
Basicly, using the same path for both host and container.

rcarrillocruz:
Closing, from what I see this requires quite a bit of changes in runner process isolation code to decouple host/container paths.
I'm just using same paths when deploying rootless for now and it works.

PR #13599, opened 2023-02-20
Add constructed inventory add form

marshmalien:
##### SUMMARY
* Add constructed inventory add form 
* Add user friendly hints to create a constructed plugin (wip)
* Input Inventories field should filter out smart and constructed inventories
* Add variables field validation 

Screenshot 2023-02-20 at 10 02 04

Screenshot 2023-02-20 at 10 02 14

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - UI

### DIFF
--- a/awx/ui/src/screens/Inventory/shared/ConstructedInventoryHint.js
+++ b/awx/ui/src/screens/Inventory/shared/ConstructedInventoryHint.js
@@ -1,0 +1,164 @@
+import React from 'react';
+import { t } from '@lingui/macro';
+import {
+  Alert,
+  AlertActionLink,
+  CodeBlock,
+  CodeBlockAction,
+  CodeBlockCode,
+  ClipboardCopyButton,
+} from '@patternfly/react-core';
+import {
+  TableComposable,
+  Thead,
+  Tr,
+  Th,
+  Tbody,
+  Td,
+} from '@patternfly/react-table';
+import { ExternalLinkAltIcon } from '@patternfly/react-icons';
+import getDocsBaseUrl from 'util/getDocsBaseUrl';
+import { useConfig } from 'contexts/Config';
+
+function ConstructedInventoryHint() {
+  const config = useConfig();
+  const [copied, setCopied] = React.useState(false);
+
+  const clipboardCopyFunc = (event, text) => {
+    navigator.clipboard.writeText(text.toString());
+  };
+
+  const onClick = (event, text) => {
+    clipboardCopyFunc(event, text);
+    setCopied(true);
+  };
+
+  const pluginSample = `plugin: constructed
+strict: true
+use_vars_plugins: true
+groups:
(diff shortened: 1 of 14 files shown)


awx/ui/src/screens/Inventory/shared/ConstructedInventoryHint.js
```diff
@@ -0,0 +1,159 @@
...
+
+  return (
+    <Alert
+      isExpandable
+      isInline
+      variant="info"
+      title={t`How to use constructed inventory plugin`}
+      actionLinks={
+        <AlertActionLink
+          href="https://docs.ansible.com/ansible/latest/collections/ansible/builtin/constructed_inventory.html"
```
AlexSCorey: We should use this function https://github.com/ansible/awx/blob/devel/awx/ui/src/util/getDocsBaseUrl.js to help ensure that the user is taken to the correct version of the docs.

awx/ui/src/screens/Inventory/shared/Inventory.helptext.js
```diff
@@ -188,6 +188,40 @@ const getInventoryHelpTextStrings = () => ({
       </>
     );
   },
+  constructedInventorySourceVars: () => {
+    const docsUrl =
+      'https://docs.ansible.com/ansible/latest/collections/ansible/builtin/constructed_inventory.html';
```
AlexSCorey: Use the same utility here https://github.com/ansible/awx/blob/devel/awx/ui/src/util/getDocsBaseUrl.js

AlexSCorey (changes requested):
I pulled this down, tested it, and tested the translations.  Everything looks good to me, but just 2 quick changes to ensure users go to the correct docs version.

awx/ui/src/screens/Inventory/shared/ConstructedInventoryHint.js
```diff
@@ -0,0 +1,159 @@
...
+          component="a"
+          target="_blank"
+        >
+          {t`View constructed plugin documentation here`}{' '}
+          <ExternalLinkAltIcon />
+        </AlertActionLink>
+      }
+    >
+      <span>{t`WIP - More to come...`}</span>
```
AlexSCorey, line 64: Whats going to be here?

marshmalien: I left "wip" there because we haven't nailed down the content. I don't want the hint content to block this PR because the edit form depends on the form to merge. I'll create a follow-up issues just for the ConstructedInventoryHint content. 

@AlanCoding had some ideas of what examples he would like to see in the hint: 
* filtering on hostvars
* filtering on related objects, like inventory name or organization name (via special variables)
* filtering on facts
* some non-trivial combination of logic

AlexSCorey (approved):
This looks good to me.

PR #13602, opened 2023-02-20
[constructed inventory] ordered m2m

relrod:
##### SUMMARY

Ordered m2m for Inventory/Inventory relationship, including changes to our custom Ordered m2m field which previously broke if the source and target model was the same.

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API


awx/main/models/inventory.py
```diff
@@ -58,6 +58,16 @@
 logger = logging.getLogger('awx.main.models.inventory')
 
 
+class InventoryConstructedInventoryMembership(models.Model):
+    constructed_inventory = models.ForeignKey('Inventory', on_delete=models.CASCADE, related_name='constructed_inventories')
```
AlanCoding: I think it might not be a good idea to put a `related_name` on the field on a through model. Think about `inventory.constructed_inventories.all()`, you would expect this to be a list of `Inventory` objects, but if the related_name is defined here, that would be a list of `InventoryConstructedInventoryMembership`, which is :exploding_head:

AlanCoding (approved):
Looks like tests are passing.

AlanCoding:
Related to this, we looked at https://github.com/jazzband/django-sortedm2m

But we have a real deal-breaker if considering using it as a library

> django-sortedm2m runs on Python 3.6+ and Django from 2.1 to 3.0.

The lack of updates for more recent Django versions looks like it would really hurt us. Indeed, a lot of the issues look like they follow this theme.

AlanCoding (approved):

PR #13603, opened 2023-02-21
Fix a bug in clean languages

jjwatt:
The `$` was not escaped for make or shell.

##### SUMMARY

Make will try to expand '$', and even if it does not, having a bare one like this breaks some syntax highlighting. Additionally, the shell could try to expand "$" since it's in double quotes instead of single quotes.

##### ISSUE TYPE

##### COMPONENT NAME

 - Other

##### AWX VERSION

```
awx: 0.1.dev32848+gaf6549f
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ clean-schema:
 
 clean-languages:
 	rm -f $(I18N_FLAG_FILE)
-	find ./awx/locale/ -type f -regex ".*\.mo$" -delete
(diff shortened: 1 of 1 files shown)


AlanCoding (approved):
Tested and confirmed I was getting an error before this fix. I don't see any reason to not move ahead with this.

TheRealHaoLiu (approved):

PR #13604, opened 2023-02-21
Change docker-clean build rule in Makefile

jjwatt:
- Use a make foreach macro and rmi instead of grep and xargs.

##### SUMMARY

Use a make foreach macro and rmi instead of grep and xargs.

##### ISSUE TYPE

##### COMPONENT NAME

 - Other (Makefile)

##### AWX VERSION

```
awx: 0.1.dev32848+g69cf85a
```

##### ADDITIONAL INFORMATION

Please comment on the `TODO` I added if you know the answer so that we can resolve it in the PR.

```

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -547,10 +547,8 @@ docker-compose-build:
 	    --cache-from=$(DEV_DOCKER_TAG_BASE)/awx_devel:$(COMPOSE_TAG) .
 
 docker-clean:
-	$(foreach container_id,$(shell docker ps -f name=tools_awx -aq && docker ps -f name=tools_receptor -aq),docker stop $(container_id); docker rm -f $(container_id);)
-	if [ "$(shell docker images | grep awx_devel)" ]; then \
-	  docker images | grep awx_devel | awk '{print $$3}' | xargs docker rmi --force; \
-	fi
+	-$(foreach container_id,$(shell docker ps -f name=tools_awx -aq && docker ps -f name=tools_receptor -aq),docker stop $(container_id); docker rm -f $(container_id);)
+	-$(foreach image_id,$(shell docker images --filter=reference='*awx_devel*' -aq),docker rmi --force $(image_id);)
 
 docker-clean-volumes: docker-compose-clean docker-compose-container-group-clean
 	docker volume rm -f tools_awx_db tools_grafana_storage tools_prometheus_storage $(docker volume ls --filter name=tools_redis_socket_ -q)


Makefile
```diff
@@ -527,10 +527,11 @@ docker-compose-build-swagger: awx/projects docker-compose-sources
 SCHEMA_DIFF_BASE_BRANCH ?= devel
 detect-schema-change: genschema
 	curl https://s3.amazonaws.com/awx-public-ci-files/$(SCHEMA_DIFF_BASE_BRANCH)/schema.json -o reference-schema.json
-	# Ignore differences in whitespace with -b
+  # Ignore differences in whitespace with -b
 	diff -u -b reference-schema.json schema.json
 
 docker-compose-clean: awx/projects
+  # FIXME: Check for file first
```
AlanCoding: I'm not really sure what this FIXME is suggesting. Could you clarify?

Makefile
```diff
@@ -547,10 +548,9 @@ docker-compose-build:
 	    --cache-from=$(DEV_DOCKER_TAG_BASE)/awx_devel:$(COMPOSE_TAG) .
 
 docker-clean:
-	$(foreach container_id,$(shell docker ps -f name=tools_awx -aq && docker ps -f name=tools_receptor -aq),docker stop $(container_id); docker rm -f $(container_id);)
-	if [ "$(shell docker images | grep awx_devel)" ]; then \
-	  docker images | grep awx_devel | awk '{print $$3}' | xargs docker rmi --force; \
-	fi
+	-$(foreach container_id,$(shell docker ps -f name=tools_awx -aq && docker ps -f name=tools_receptor -aq),docker stop $(container_id); docker rm -f $(container_id);)
+  # TODO: Figure out if it was intentional for the old code to not remove receptor images.
```
AlanCoding: The receptor images are only used for building images and for hop nodes. This obviously only attempts to deal with the awx_devel image. It also doesn't touch the redis or postgres image, because these are external and should change infrequently. While receptor is more integral and picking up changes are more important, you get this by refreshing the awx_devel image, itself... unless you're building locally or using hop nodes.

I think I agree that in some context it would be appropriate to remove receptor images, but the distinction between `make docker-compose-clean` and `make docker-clean` is, itself, challenging to undertand.

Makefile
jjwatt: Oh yeah. If the `docker-compose.yml` file isn't there, you get kind of a nasty error. But, this FIXME could apply to a ton of these rules/commands, so it might be better to just leave it out for now.

Makefile
jjwatt: Thanks. And, I agree about those rules being challenging to understand. @TheRealHaoLiu and I have been talking about fixing up our GNU Make rules and de-complicating them.

Makefile
```diff
@@ -527,7 +527,7 @@ docker-compose-build-swagger: awx/projects docker-compose-sources
 SCHEMA_DIFF_BASE_BRANCH ?= devel
 detect-schema-change: genschema
 	curl https://s3.amazonaws.com/awx-public-ci-files/$(SCHEMA_DIFF_BASE_BRANCH)/schema.json -o reference-schema.json
-	# Ignore differences in whitespace with -b
+  # Ignore differences in whitespace with -b
```
TheRealHaoLiu: [nit] is this change intentional

jjwatt: Sort of. Can you tell if those are spaces or tabs in the other one? I have a habit of using spaces for "make" stuff and only using tabs for "command" blocks because anything that begins with tabs will be passed to the shell. Maybe it's just more spaces than my editor converted it to, or maybe it's a tab that I converted to spaces. I'll check.

jjwatt: Went ahead and just reversed that hunk

AlanCoding (approved):
updated form looks more consistent and direct

PR #13608, opened 2023-02-21
Use ansible-runner change to get periodic keep-alive messages in K8S

AlanCoding:
##### SUMMARY
This is a candidate patch for consuming https://github.com/ansible/ansible-runner/pull/1191, which adds keepalive messages.

FYI for @nitzmahone, I hope we are narrowing in on this implementation. This will require keeping and supporting the environment variable. There are no plans to use the CLI flag. I also maintain a weak preference to have the event callback pass on the keep-alive events.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API


AlanCoding:
I tested building the UI, and then I was able to edit and save the value of the setting. Without UI changes I was able to see the setting, but not able to edit it. I know some changes here are just test file changes.

mabashian:
UI changes lgtm

awx/main/conf.py
```diff
@@ -282,6 +282,16 @@
     placeholder={'HTTP_PROXY': 'myproxy.local:8080'},
 )
 
+register(
+    'AWX_RUNNER_KEEPALIVE_SECONDS',
+    field_class=fields.IntegerField,
+    label=_('K8S Ansible Runner Keep-Alive Message Interval'),
+    help_text=_('Only applies to K8S deployments and container_group jobs. If not 0, send a message every so-many seconds to keep connection open.'),
```
shanemcd:
```suggestion
    help_text=_('Only applies to jobs running in a Container Group. If not 0, send a message every so-many seconds to keep connection open.'),
```

awx/main/tasks/callback.py
```diff
@@ -85,6 +85,8 @@ def event_handler(self, event_data):
         # which generate job events from two 'streams':
         # ansible-inventory and the awx.main.commands.inventory_import
         # logger
+        if event_data.get('event') == 'keepalive':
+            return
```
AlanCoding, line 89: From discussion with @TheRealHaoLiu, we are unlikely to need this, but I would prefer to keep it in anyway. Maybe leave a comment that it should be dead code.

awx/main/conf.py
```diff
@@ -282,6 +282,16 @@
     placeholder={'HTTP_PROXY': 'myproxy.local:8080'},
 )
 
+register(
+    'AWX_RUNNER_KEEPALIVE_SECONDS',
```
AlanCoding, line 286: should we rename this to something about K8S_?

CFSNM (approved):

PR #13609, opened 2023-02-21
WIP: ui-next build

TheRealHaoLiu:
##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME

 - Other

##### AWX VERSION

```
awx: 21.12.1.dev4+g86a4a38be6
```

##### ADDITIONAL INFORMATION

```

```


awx/settings/defaults.py
```diff
@@ -83,7 +83,11 @@
 
 USE_TZ = True
 
-STATICFILES_DIRS = [os.path.join(BASE_DIR, 'ui', 'build', 'static'), os.path.join(BASE_DIR, 'static')]
+STATICFILES_DIRS = [
+    os.path.join(BASE_DIR, 'ui', 'build', 'static'),
+    os.path.join(BASE_DIR, 'ui-next', 'ansible-ui', 'build'),
+    os.path.join(BASE_DIR, 'static'),
+]
```
TheRealHaoLiu: This seems weird

Makefile
```diff
@@ -1,3 +1,8 @@
+mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
+current_dir := $(dir $(mkfile_path))
```
jjwatt: Can you use `CURDIR` here? https://www.gnu.org/software/make/manual/html_node/Quick-Reference.html

awx/ui-next/Makefile
```diff
@@ -0,0 +1,87 @@
...
+
+ANSIBLE_UI_GIT_BRANCH ?= main
+
+ANSIBLE_UI_LOCAL ?=
+
+ANSIBLE_UI_DIR ?= $(ui_next_dir)ansible-ui
+
+.PHONY: ui-next/clone-https
+## Shallow clone the ui-next repo via https skip if ANSIBLE_UI_GIT_REPO_HTTPS is undefined
+ui-next/clone-https:
```
jjwatt: I don't think these necessarily need to be separate rules. I would simplify by making it `ANSIBLE_UI_URL`. If you need to do different things when it's https or git, then you can check the `ANSIBLE_UI_URL` whenever you want. At the very least, I'd probably collapse the rules into a `ui-next/clone` or use `$(ANSIBLE_UI_SRCDIR)` as a target, and decide how to clone it in that target rule. And even decide *if* you need to clone it (see other comment about `ANSIBLE_UI_SRCDIR`)--you may get a sdist path or some random `/tmp/ui-next` joint, and then you can skip cloning entirely and try to use it.

awx/ui-next/Makefile
```diff
@@ -0,0 +1,87 @@
...
+endif
+
+ANSIBLE_UI_GIT_REPO_HTTPS ?= https://github.com/ansible/ansible-ui.git
+ANSIBLE_UI_GIT_REPO_SSH ?= git@github.com:ansible/ansible-ui.git
+
+ANSIBLE_UI_GIT_BRANCH ?= main
+
+ANSIBLE_UI_LOCAL ?=
+
+ANSIBLE_UI_DIR ?= $(ui_next_dir)ansible-ui
```
jjwatt: I don't love how this `ANSIBLE_UI_DIR` is handled. We can fix it up later, though. I would probably do something like this, though: `ANSIBLE_UI_URI` (a cloneable ui joint--e.g., a remote git repo, but could also be a local one!) and `ANSIBLE_UI_SRCDIR` could be something set irrespective of `ANSIBLE_UI_URI` (like if you already have an exploded copy, an sdist (without a `.git` dir! remember sdists don't have that!) or anything with that useable code in it). `ANSIBLE_UI_URI` could be populated after the clone when you need to clone or it could be set to the sdist dir or whatever.

jjwatt (commented):
Couple of quick comments after a quick look, so I don't forget. We can save these fixups for later, though.

awx/ui-next/Makefile
```diff
@@ -0,0 +1,87 @@
+ui_next_mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
+ui_next_dir := $(dir $(ui_next_mkfile_path))
+
+ifeq ($(current_dir), undefined)
+	ui_next_dir := $(dir $(ui_next_mkfile_path))
+else
+	ui_next_dir := $(shell realpath --relative-to $(current_dir) $(ui_next_dir))/
+endif
+
+ANSIBLE_UI_GIT_REPO_HTTPS ?= https://github.com/ansible/ansible-ui.git
```
jjwatt: These will always be set. The double-edged sword of conditional assignment :). It does mean that those `if -z` conditionals in the commands below are useless unless someone shadows these somewhere (either by {un,}setting them at the top of the rule or from the env or command line. I don't love this because it means you probably can't just run `make` without setting/unsetting them.

awx/ui-next/Makefile
```diff
@@ -0,0 +1,87 @@
...
+	elif [ -d "$(ANSIBLE_UI_DIR)" ]; then \
+		echo "SKIP: ui-next/clone-https. $(ANSIBLE_UI_DIR) already exists."; \
+	else \
+		git clone --depth 1 --branch $(ANSIBLE_UI_GIT_BRANCH) $(ANSIBLE_UI_GIT_REPO_HTTPS) $(ANSIBLE_UI_DIR) || true; \
+	fi
+
+.PHONY: ui-next/clone-ssh
+## Shallow clone the ui-next repo via ssh.
+ui-next/clone-ssh:
+	@if [ -z "$(ANSIBLE_UI_GIT_REPO_SSH)" ]; then \
```
jjwatt: Ohh. We should have style guides for awx-ish projects, and we could include a GNU Make style guide/section. But, for now I just want to say if you do decide to keep handling these somewhat mutually exclusive vars this way, then at least wrap these repetitive lines in a "user-defined function" (really, a multi-line macro with optional params):
```make
# $(call maybe-clone,repo-var)
define maybe-clone
    @if [ -z $$1 ]; then \ echo "skip: not set"; \
        elif [ -d "$ANSIBLE_UI_DIR" ]; then \
            echo 'skip: ui dir exists'; \
        else \
            git clone --depth 1 --branch $(branch) $($1) $(ANSIBLE_UI_DIR) || true; \
       fi
endef
```
Something like that. Didn't test.
But, then you could, e.g., `$(call maybe-clone,ANSIBLE_UI_GIT_REPO_SSH)` It could be simplified further by not parameterizing it, too. Just put the text you want in a `define`, and make two of them if you need to. I think even this might be simpler if you collapse the `REPO_{SSH,HTTPS}` gargoyles, too.

jjwatt: oh yeah, you might skip the subshells with like,
```
$(if $(wildcard $(dir $(ANSIBLE_UI_DIR)),$(warning "skippity doo da"),git clone --depth 1 ...)
```
My syntax might be off. Getting back into the make headspace here.

awx/ui-next/Makefile
TheRealHaoLiu: ur right!

Makefile
```diff
@@ -590,7 +590,7 @@ awx-kube-dev-build: Dockerfile.kube-dev
 	    -t $(DEV_DOCKER_TAG_BASE)/awx_kube_devel:$(COMPOSE_TAG) .
 
 ## Build awx image for deployment on Kubernetes environment.
-awx-kube-build: Dockerfile
+awx-kube-build: Dockerfile awx/ui_next/src
```
TheRealHaoLiu, line 594: note this make target is not what we actually uses for building awx 

need to make sure that the playbook works

rooftopcellist: @TheRealHaoLiu I believe we need to add `awx/ui_next/src` to the `make ui-release` target.

TheRealHaoLiu:
moving this work into a feature branch because feature branch now will trigger builds

PR #13612, opened 2023-02-22
Serve ansible UI

kialam:
##### SUMMARY

Serve ansible-ui alongside AWX

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - UI

### DIFF
--- a/awx/ui/urls.py
+++ b/awx/ui/urls.py
@@ -9,6 +9,10 @@ class IndexView(TemplateView):
 
     template_name = 'index.html'
 
+class ControllerView(TemplateView):
+
+    template_name = 'index_controller.html'
+
 
 class MigrationsNotran(TemplateView):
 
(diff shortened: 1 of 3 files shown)


awx/ui/urls.py
```diff
@@ -27,4 +31,4 @@ def get_context_data(self, **kwargs):
 
 app_name = 'ui'
 
-urlpatterns = [re_path(r'^$', IndexView.as_view(), name='index'), re_path(r'^migrations_notran/$', MigrationsNotran.as_view(), name='migrations_notran')]
+urlpatterns = [re_path(r'^$', IndexView.as_view(), name='index'), re_path(r'^controller/.*$', ControllerView.as_view(), name='index'), re_path(r'^migrations_notran/$', MigrationsNotran.as_view(), name='migrations_notran')]
```
kialam, line 34: Note the regex pattern `r'^controller/.*$'` here. We need this in order so that hard refreshes work. Otherwise django views(`/controller/projects/?sort....`) as invalid routes

PR #13619, opened 2023-02-23
Allow serving app from non-root path in dev env

shanemcd:
##### SUMMARY
Usage:

```
$ EXTRA_SOURCES_ANSIBLE_OPTS='-e ingress_path=/awx' make docker-compose 
$ curl http://localhost:8013/awx/api/v2/ping/
```

##### ISSUE TYPE

##### COMPONENT NAME

 - Other


tools/docker-compose/ansible/roles/sources/tasks/main.yml
```diff
@@ -49,19 +49,17 @@
...
+    - "local_settings.py"
     - "websocket_secret.py"
     - "haproxy.cfg"
+    - "nginx.conf"
+    - "nginx.locations.conf"
 
 - name: Delete old local_settings.py
   file:
     path: "{{ playbook_dir }}/../../../awx/settings/local_settings.py"
     state: absent
```
AlanCoding: I don't think you need the "Delete old local_settings.py" task anymore either?

tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
```diff
@@ -40,6 +41,8 @@ services:
       - "../../docker-compose/_sources/database.py:/etc/tower/conf.d/database.py"
       - "../../docker-compose/_sources/websocket_secret.py:/etc/tower/conf.d/websocket_secret.py"
       - "../../docker-compose/_sources/local_settings.py:/etc/tower/conf.d/local_settings.py"
```
AlanCoding, line 43: well this is confusing that we had 2 paths (on the host machine) where local_settings.py could be - either in `awx/settings/local_settings.py` or in `tools/docker-compose/_sources/local_settings.py`

tools/docker-compose/ansible/roles/sources/templates/nginx.conf.j2
```diff
@@ -17,106 +16,61 @@ http {
...
-            uwsgi_read_timeout 120s;
-            uwsgi_pass uwsgi;
-            include /etc/nginx/uwsgi_params;
-            {%- if extra_nginx_include is defined %}
-            include {{ extra_nginx_include }};
-            {%- endif %}
-            proxy_set_header X-Forwarded-Port 443;
-            uwsgi_param HTTP_X_FORWARDED_PORT 443;
-        }
+        include /etc/nginx/conf.d/*.conf;
```
AlanCoding, line 74: interesting approach

AlanCoding (approved):
great cleanup work

akus062381:
this PR seems to have introduced several UI failures. I created a PR reverting it and I am running tests against the reverted PR to see if the failures are fixed.

AlanCoding:
We had another reversion with https://github.com/ansible/awx/pull/13623, which reverted something known to have caused a bunch of failures. Your PR seems to have happened after that revert merged, so it might be worth getting results from the devel branch so we know it's from one or the other.

PR #13624, opened 2023-02-27
[usage-collection] Compliance computation settings

slemrmartin:
##### SUMMARY

This PR adds setting `SUBSCRIPTION_USAGE_MODEL` 
which involves deleted HostMetrics when `unique_managed_hosts`.

- https://issues.redhat.com/browse/AA-1594
- https://issues.redhat.com/browse/AA-1622

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
- CLI

##### AWX VERSION

```
awx: 21.12.1.dev41+g85fcd0d3b1.d20230227
```

##### ADDITIONAL INFORMATION

```

```


awx/settings/defaults.py
```diff
@@ -1003,3 +1003,6 @@
 
 # This is overridden downstream via /etc/tower/conf.d/cluster_host_id.py
 CLUSTER_HOST_ID = socket.gethostname()
+
+# Compliance computation involves deleted hosts (HostMetric) when True
```
AlanCoding: I'd like to make this comment a little easier to read...

> License compliance for total host count will include deleted hosts (using /api/v2/host_metrics/) when True

awx/main/conf.py
```diff
@@ -765,6 +765,17 @@
     help_text=_('Indicates whether the instance is part of a kubernetes-based deployment.'),
 )
 
+register(
+    'SUBSCRIPTION_COMPLIANCE_WITH_DELETING',
+    field_class=fields.BooleanField,
+    default=False,
+    label=_('Involve deleted hosts into Compliance computation and shows Host Metric menu.'),
+    help_text=_('If set to true, count of Hosts Automated is decreased by number of deleted host metrics.'),
+    read_only=True,
```
slemrmartin: @TheRealHaoLiu is the `read_only` flag ok?
If I set it here, the value in `awx/settings/defaults.py` and `awx/settings/local_overrides.py` is not relevant

AlanCoding: Settings defined here have very non-obvious precedence rules. If the user sets it in a settings file then it will automatically become read-only. IIRC, we very infrequently use the read_only setting here and I suspect it will have some bugs.

Ladas: @slemrmartin @TheRealHaoLiu I would maybe do this in a more future proof way, I'd say setting:

SUBSCRIPTION_USAGE_MODEL (string)

The value we'd be setting now would be "unique_managed_hosts", which affects the bacground processing and if we'll show/compute the compliance status and used/remaining hosts. (so only AoC could set this to None or to "unique_managed_hosts_montly" directly)

-----

In 2.5 we'd do  "unique_managed_hosts_montly", that will only compute monthly usage, but will  hide remaining and compliance (since it's usage based, there is no upper limit, therefore there is no compliance)

------

and lets send this in Analytics under the cluster_info.json

awx/main/utils/licensing.py
```diff
@@ -382,8 +382,15 @@ def validate(self):
 
         current_instances = Host.objects.active_count()
         license_date = int(attrs.get('license_date', 0) or 0)
-        automated_instances = HostMetric.active_objects.count()
-        first_host = HostMetric.active_objects.only('first_automation').order_by('first_automation').first()
+
+        exclude_deleted = getattr(settings, 'SUBSCRIPTION_COMPLIANCE_WITH_DELETING', False)
+        if exclude_deleted:
```
Ladas: @slemrmartin @TheRealHaoLiu so the only valid model here now is with removing the deleted.

So I'd say if SUBSCRIPTION_USAGE_MODEL != "unique_managed_nodes", we can set "automated_instances"/"free_instances" to None and Compliant status to None. And UI should hide these fields. Or we can set Compliant to True, if it would break other things, since it's kind of Compliant if there is no Compliance in the model.

slemrmartin: changed to the SUBSCRIPTION_USAGE_MODEL. We can keep it as is, the menu and values should be hidden, and we'll extend it with another model in 2.5

awx/main/utils/licensing.py
```diff
@@ -382,8 +382,15 @@ def validate(self):
...
         license_date = int(attrs.get('license_date', 0) or 0)
-        automated_instances = HostMetric.active_objects.count()
-        first_host = HostMetric.active_objects.only('first_automation').order_by('first_automation').first()
+
+        model = getattr(settings, 'SUBSCRIPTION_USAGE_MODEL', '')
+        if model == 'unique_managed_hosts':
+            automated_instances = HostMetric.active_objects.count()
+            first_host = HostMetric.active_objects.only('first_automation').order_by('first_automation').first()
+        else:
+            automated_instances = HostMetric.objects.count()
```
Ladas, line 392: Might be more understandable to put None here. But it really doesn't matter what we set here, since UI should hide the compliance/consumption related fields if model != 'unique_managed_hosts'

awx/settings/defaults.py
```diff
@@ -1003,3 +1003,9 @@
 
 # This is overridden downstream via /etc/tower/conf.d/cluster_host_id.py
 CLUSTER_HOST_ID = socket.gethostname()
+
+# License compliance for total host count. Possible values:
+# - '': No model - Subscription not counted from Host Metrics
+# - 'unique_managed_hosts': Compliant = automated - deleted hosts (using /api/v2/host_metrics/)
+# - 'unique_managed_hosts_monthly': TBD: AoC on Azure (now equal to '')
+SUBSCRIPTION_USAGE_MODEL = ''
```
bhavenst, line 1016: Will "unique_managed_hosts_monthly" be implemented with this change?  I've been informed to set SUBSCRIPTION_USAGE_MODEL = 'unique_managed_hosts_monthly' by @Ladas but looking at this PR not sure that will be the way to go for 2.4.  I tried adding it in our current 2.3 version to be ready in advance, but controller was not happy (thought maybe it would be ignored).

awx/main/conf.py
```diff
@@ -765,6 +765,19 @@
     help_text=_('Indicates whether the instance is part of a kubernetes-based deployment.'),
 )
 
+# Possible values
+# - ''
+# - unique_managed_hosts
+register(
+    'SUBSCRIPTION_USAGE_MODEL',
+    field_class=fields.CharField,
```
AlanCoding: if it has 2 possible values then `field_class=fields.ChoiceField` would take care of your validation.

awx/settings/defaults.py
Ladas: The "unique_managed_hosts_monthly" will not show anything for 2.4. (most likely, we might be able to backport something to 2.4.x, if we'll get the DB changes in 2.4)

Yeah I suppose it will fail in 2.3 since the setting is not defined there.

awx/main/conf.py
Ladas: Right, we talked we that we can even make the 'unique_managed_hosts' default, so we don't have to include installer in this.

the '' strategy (no subscription model strategy) will be covered by the productization flag (that is now hiding the whole subs section) 

Then choice field with ['unique_managed_hosts', 'unique_managed_hosts_monthly'] sounds good.

awx/main/conf.py
slemrmartin: @Ladas do you mean `instance_info.json`? 
If it's intended to this feature, it requires rebase of feature branch (because of https://issues.redhat.com/browse/AA-1514 - PRs were merged)

awx/main/utils/licensing.py
slemrmartin: as it doesn't matter, I'll not change the current behaviour for AWX and keep it

awx/settings/defaults.py
bhavenst: OK, fair enough.  We'll add that when we get 2.4.  Thanks

awx/main/conf.py
slemrmartin: added to `config.json` next to the `free_instances` value

awx/main/conf.py
```diff
@@ -775,6 +775,17 @@
     help_text=_('Indicates whether the instance is part of a kubernetes-based deployment.'),
 )
 
+register(
+    'SUBSCRIPTION_USAGE_MODEL',
+    field_class=fields.ChoiceField,
+    choices=[SUBSCRIPTION_USAGE_MODEL_UNIQUE_HOSTS, SUBSCRIPTION_USAGE_MODEL_UNIQUE_HOSTS_MONTHLY],
+    default='',
```
AlanCoding, line 788: default isn't in the choices, and that can cause problems in some corner cases. Some clients like to PATCH with defaults instead of resetting, which can give problems with stuff like this. So I think you just want to reuse one of those constants you have for the default.

slemrmartin: @AlanCoding added

AlanCoding: Here, two different values are going to have the same effect. Choice labels (via tuples) are a good way to avoid that. Like:

```python
register(
    'SUBSCRIPTION_USAGE_MODEL',
    field_class=fields.ChoiceField,
    choices=[
        ('', _('Describe what this does')),
        (SUBSCRIPTION_USAGE_MODEL_UNIQUE_HOSTS, _('Does something monthly, more info...')),
    ],
    default='',
```

slemrmartin: Ok, I'll add some description. Default and Azure choice will have the same effect now, but the value for AoC for Azure is added conciously to be able to set it in 2.4 in AoC upgrade process. At least it'll be sent to Analytics. And different behaviour and background tasks will be added later, probably 2.5 release

slemrmartin: Anyway I suppose labels shouldn't contain words like "Automation Controller", "AoC on Azure" and other downstream terminology

Ladas: (SUBSCRIPTION_USAGE_MODEL_UNIQUE_HOSTS, _('Usage based on unique managed nodes in a large historical time frame and delete functionality for no longer used managed nodes')),
(SUBSCRIPTION_USAGE_MODEL_UNIQUE_HOSTS_MONTHLY, _('Usage based on unique managed nodes in a calendar month time frame')),


Somethin like this should work, you're right we shouldn't use product names here.

bhavenst: Just as an FYI, Azure doesn't use calendar month, but uses a rolling month period based on installation date..

Ladas: @bhavenst oh is it? I was checking with Chris Hambridge that it was like this before, but then you moved to calendar month billing?

bhavenst: Azure doesn't support calendar month. On AWS it is calendar month though AFAIK.

Ladas: @slemrmartin ok, lets take the SUBSCRIPTION_USAGE_MODEL_UNIQUE_HOSTS_MONTHLY out for now. @bhavenst AoC can set '' value for now.

slemrmartin: ok, done

awx/main/conf.py
slemrmartin: https://gitlab.cee.redhat.com/automation-analytics/automation-analytics-backend/-/merge_requests/697

awx/settings/defaults.py
```diff
@@ -1008,3 +1008,9 @@
 
 # This is overridden downstream via /etc/tower/conf.d/cluster_host_id.py
 CLUSTER_HOST_ID = socket.gethostname()
+
+# License compliance for total host count. Possible values:
+# - '': No model - Subscription not counted from Host Metrics
+# - 'unique_managed_hosts': Compliant = automated - deleted hosts (using /api/v2/host_metrics/)
+# - 'unique_managed_hosts_monthly': TBD: AoC on Azure (now equal to '')
```
AlanCoding, line 1015: update comment here - since it's now a binary, not tertiary option

awx/main/conf.py
```diff
@@ -775,6 +775,23 @@
     help_text=_('Indicates whether the instance is part of a kubernetes-based deployment.'),
 )
 
+register(
+    'SUBSCRIPTION_USAGE_MODEL',
+    field_class=fields.ChoiceField,
+    choices=[
+        ('', _('Default model for AWX - no subscription')),
```
AlanCoding, line 782: I want to leave this as a non-blocking comment, but still want it to be a very strong suggestion.

To improve the coherency of docs, I would add a sentence to the end of this - "Deletion of host_metrics will not be considered for purposes of managed host counting." Otherwise, I feel like the numerical consequence of the option is lost, or left too vague to understand.

AlanCoding (approved):
still left 2 commenting / documenting suggestions, but changes looks fine.

awx/settings/defaults.py
slemrmartin: https://github.com/ansible/awx/pull/13736

awx/main/conf.py
slemrmartin: https://github.com/ansible/awx/pull/13736

PR #13626, opened 2023-02-27
Changing how associations work in awx collection

john-westcott-iv:
Some associations, like galaxy_credentials are ordered. The association logic tried to minimize the number of calls to the server when performing associations but this had the side affect of loosing order of associations for ordered fields. The new logic will ensure that order is preserved.

NOTE: There was several changes to satisfy `black` in here.

##### ISSUE TYPE

##### COMPONENT NAME

 - Collection

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx_collection/plugins/module_utils/controller_api.py
+++ b/awx_collection/plugins/module_utils/controller_api.py
@@ -680,20 +680,26 @@ class ControllerAPIModule(ControllerModule):
         response = self.get_all_endpoint(association_endpoint)
         existing_associated_ids = [association['id'] for association in response['json']['results']]
 
-        # If the current associations match the desired associations then we can just return
-        if existing_associated_ids == new_association_list:
-            return
+        # Some associations can be ordered (like galaxy credentials)
+        if association_endpoint.strip('/').split('/')[-1] in ('instance_groups', 'galaxy_credentials'):
+            if existing_associated_ids == new_association_list:
+                return  # If the current associations EXACTLY match the desired associations then we can return
+            removal_list = existing_associated_ids  # because of ordering, we have to remove everything
+            addition_list = new_association_list  # re-add everything back in-order
+        else:
+            if set(existing_associated_ids) == set(new_association_list):
+                return
+            removal_list = set(existing_associated_ids) - set(new_association_list)
+            addition_list = set(new_association_list) - set(existing_associated_ids)
 
-        # Some associations can be ordered (like galaxy credentials), because of this we have to remove everything to re-add in order
-        for an_id in existing_associated_ids:
+        for an_id in removal_list:
             response = self.post_endpoint(association_endpoint, **{'data': {'id': int(an_id), 'disassociate': True}})
             if response['status_code'] == 204:
                 self.json_output['changed'] = True
             else:
                 self.fail_json(msg="Failed to disassociate item {0}".format(response['json'].get('detail', response['json'])))
 
-        # Now we will add everything back in the received order
-        for an_id in new_association_list:
+        for an_id in addition_list:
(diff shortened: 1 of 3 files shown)


awx_collection/plugins/module_utils/controller_api.py
```diff
@@ -700,17 +677,16 @@ def modify_associations(self, association_endpoint, new_association_list):
         response = self.get_all_endpoint(association_endpoint)
         existing_associated_ids = [association['id'] for association in response['json']['results']]
 
-        # Disassociate anything that is in existing_associated_ids but not in new_association_list
-        ids_to_remove = list(set(existing_associated_ids) - set(new_association_list))
-        for an_id in ids_to_remove:
+        # Some associations can be ordered (like galaxy credentials), because of this we have to remove everything to re-add in order
+        for an_id in existing_associated_ids:
```
john-westcott-iv: These lines are the actual change along with 688 and 689.

djyasin (approved):
Reviewed with John.

awx_collection/plugins/module_utils/controller_api.py
```diff
@@ -700,17 +678,20 @@ def modify_associations(self, association_endpoint, new_association_list):
...
             response = self.post_endpoint(association_endpoint, **{'data': {'id': int(an_id), 'disassociate': True}})
             if response['status_code'] == 204:
                 self.json_output['changed'] = True
             else:
                 self.fail_json(msg="Failed to disassociate item {0}".format(response['json'].get('detail', response['json'])))
 
-        # Associate anything that is in new_association_list but not in `association`
-        for an_id in list(set(new_association_list) - set(existing_associated_ids)):
+        # Now we will add everything back in the received order
```
AlanCoding: I understand that this is what the UI does for ordered relationships (like galaxy credentials), and for those types of relationships, I agree, this should also be acceptable in the API.

But if we ran this for _unordered_ relationships, that would be significantly disruptive to users. Because the order we get from the API isn't deterministic, it will remove all objects and add them back arbitrarily. This will dramatically slow down playbooks, change task status from ok to changed incorrectly, and spam the activity stream with pointless removals and associations.

...if I'm understand this.

I don't see a choice other than distinguishing ordered from unordered relationships.

john-westcott-iv: Yea, we'd have to introduce a new field to indicate what is ordered and not right? There is nothing in the API which would tell us now?

AlanCoding: yeah that's right. I can tell you the names of ordered relationships - it's "instance_groups" and "input_inventories"

john-westcott-iv: Added https://github.com/ansible/awx/issues/13654 with a back reference here.

AlanCoding:
I put up the patch https://github.com/john-westcott-iv/awx/pull/70, because I'm not happy with the current state of the PR. We basically have to maintain two code paths for managing the association diff. There's no way around that unless we are really really disruptive.

AlanCoding (changes requested):
I requested specific changes in the patch I linked. With that, I think this is a good thing to do. Also there are conflicts to resolve.

john-westcott-iv:
> I requested specific changes in the patch I linked. With that, I think this is a good thing to do. Also there are conflicts to resolve.

@AlanCoding I pulled in your PR and made one little change to have the list of ordered associations as a class variable at the top of the class instead of buried in an `if` statement. IDK why it felt better to me that way.

AlanCoding (approved):
thanks, this was very necessary to do and adds good testing

jay-steurer (approved):

PR #13628, opened 2023-02-27
[code health] Breaking serializer.py into more manageable files

jessicamack:
##### SUMMARY

The serializer file is 5.3k lines long. Breaking it up in to separate modules will make it more manageable and maintainable.

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
21.12.1
```

##### ADDITIONAL INFORMATION

```

```


AlanCoding:
This is pretty clearly beyond saving regarding git conflicts. I don't see any indication that this can pick up changes, which is why we want to use `git mv` when possible (probably for the bulk of the content). So rebasing would have to re-apply changes that happened in devel, and this manual process can not be trusted. If restarting, we need to quickly have the work finished and merged so that we don't get into conflict hell.

PR #13631, opened 2023-02-27
add handling for job call back status 'idle_timeout'

TheRealHaoLiu:
##### SUMMARY
help distinguish between different type of timeouts

won't have any effect until 
https://github.com/ansible/ansible-runner/pull/1200
is merged

but can be merged without the PR with no harm

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

##### AWX VERSION

```
awx: 21.12.1.dev37+gb74b9ca2ea
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/tasks/jobs.py
+++ b/awx/main/tasks/jobs.py
@@ -611,9 +611,9 @@ class BaseTask(object):
             status = res.status
             rc = res.rc
 
-            if status in ('timeout', 'error'):
+            if status in ('timeout', 'idle_timeout', 'error'):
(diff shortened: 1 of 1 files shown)


AlanCoding:
shouldn't we give some indication to the user that it was an idle timeout vs. a job timeout?

dmzoneill:
Having chatted with @TheRealHaoLiu this is waiting on an upstream patch to support this.
Marking as blocked

Issue #13607, opened 2023-02-21
[constructed-inventory][UI] Filter out constructed inventory hosts in smart host filter 


marshmalien:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

New Feature

### Feature Summary

Users should not be able to view constructed inventory hosts in the smart inventory host lookup. We don't allow users to add constructed inventory hosts to a smart inventory via the API, so we should hide this in the UI. 

### Select the relevant components

- [X] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

1. Create a constructed inventory with hosts
2. Navigate to smart inventory add form
3. Open host filter lookup
4. View host list

### Current results

User can see constructed inventory hosts

### Sugested feature result

User should only be able to filter on regular inventory hosts

PR #13635, opened 2023-02-28
Filter out constructed inventory hosts from smart host filter 

marshmalien:
##### SUMMARY
Fixes https://github.com/ansible/awx/issues/13607

Remove constructed inventory hosts as an option in the smart inventory host lookup

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - UI


AlexSCorey (approved):
Pulled it down and tested it.  Looks good to me.

PR #13641, opened 2023-03-02
Fixes #6556 Expose SOCIAL_AUTH_USERNAME_IS_FULL_EMAIL

### LINKED ISSUE #6556: Azure AD username mapings
##### ISSUE TYPE
 - Change username mappings for Azure AD configuration

##### SUMMARY
At the moment when configuring Azure AD SSO the username is mapped as NameSurname. It would be nice to be able to change the mapping to a different attribute in Azure AD

Klaas-:
##### SUMMARY

Hi, this should fix #6556 - with that change you are able to set the setting that allows to force new social accounts to use email as username.

This change exposes a current default setting  https://github.com/ansible/awx/blob/a47cfc55ab8a73777576e8acf4f69b290b6350ad/awx/settings/defaults.py#L493 to be overwrite able via awx settings boolean
 

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 - UI

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

After rebasing on latest devel the local testing seems to work for me, I can see the setting in awx api and change it through ui

### DIFF
--- a/awx/sso/conf.py
+++ b/awx/sso/conf.py
@@ -148,6 +148,16 @@ register(
     placeholder=['username', 'email'],
 )
 
+register(
+    'SOCIAL_AUTH_USERNAME_IS_FULL_EMAIL',
+    field_class=fields.BooleanField,
+    default=False,
+    label=_('Use Email address for usernames'),
+    help_text=_('Enabling this setting will tell social auth to use the full Email as username instead of the full name'),
+    category=_('Authentication'),
+    category_slug='authentication',
+)
+
 ###############################################################################
 # LDAP AUTHENTICATION SETTINGS
 ###############################################################################
--- a/awx/ui/src/screens/Setting/MiscAuthentication/MiscAuthenticationDetail/MiscAuthenticationDetail.test.js
+++ b/awx/ui/src/screens/Setting/MiscAuthentication/MiscAuthenticationDetail/MiscAuthenticationDetail.test.js
@@ -40,6 +40,7 @@ describe('<MiscAuthenticationDetail />', () => {
         SOCIAL_AUTH_ORGANIZATION_MAP: {},
         SOCIAL_AUTH_TEAM_MAP: {},
         SOCIAL_AUTH_USER_FIELDS: [],
+        SOCIAL_AUTH_USERNAME_IS_FULL_EMAIL: false,
       },
     });
     await act(async () => {
@@ -79,6 +80,7 @@ describe('<MiscAuthenticationDetail />', () => {
     assertVariableDetail(wrapper, 'Social Auth Organization Map', '{}');
     assertVariableDetail(wrapper, 'Social Auth Team Map', '{}');
     assertVariableDetail(wrapper, 'Social Auth User Fields', '[]');
+    assertDetail(wrapper, 'Use Email address for usernames', 'Off');
     assertDetail(
       wrapper,
       'Allow External Users to Create OAuth2 Tokens',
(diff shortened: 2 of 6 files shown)


fosterseth:
for the django migration errors, just kill the containers and rerun `make docker-compose`. there is some kind of race condition the first time starting the containers (when migrations are needed).

Klaas-:
> for the django migration errors, just kill the containers and rerun `make docker-compose`. there is some kind of race condition the first time starting the containers (when migrations are needed).

Ah that would also explain why it's working now and it may be unrelated to my rebase on devel :)

fosterseth:
pulled down the branch and tested it out in development environment + keycloak

can confirm with `settings.SOCIAL_AUTH_USERNAME_IS_FULL_EMAIL=True` that the full email is used in place of the user name in AWX

image

Klaas-:
I think I fixed the tests

awx/sso/conf.py
```diff
@@ -148,6 +148,16 @@ def __call__(self):
     placeholder=['username', 'email'],
 )
 
+register(
+    'SOCIAL_AUTH_USERNAME_IS_FULL_EMAIL',
+    field_class=fields.BooleanField,
+    default=False,
+    label=_('Use E-Mail address for usernames'),
```
fosterseth: small nitpick, can we change
E-Mail to Email (here and elsewhere)? It will be more consistent with the usage in the rest of the codebase

fosterseth:
UI auth settings page
image

fosterseth:
image

@Klaas-  also, can you please move the setting up toward the top? maybe under "Allow anonymous users to poll metrics"

fosterseth:
image

also, can you please move the setting up toward the top? maybe under "Allow anonymous users to poll metrics"

Klaas-:
> image
> 
> @Klaas- also, can you please move the setting up toward the top? maybe under "Allow anonymous users to poll metrics"

I would if I understood how that gets there :) 

The allow anonymous users stuff is not in the js details. So I am assuming they get build automatically from settings somehow? :)
awx/api/views/metrics.py:        if settings.ALLOW_METRICS_FOR_ANONYMOUS_USERS:
awx/api/views/metrics.py:        if settings.ALLOW_METRICS_FOR_ANONYMOUS_USERS or request.user.is_superuser or request.user.is_system_auditor:
awx/api/conf.py:    'ALLOW_METRICS_FOR_ANONYMOUS_USERS',
awx/settings/defaults.py:ALLOW_METRICS_FOR_ANONYMOUS_USERS = False

I was assuming the sorting comes from here:
https://github.com/ansible/awx/blob/89c06853177880ef9bde61c203e76c9ce7e95271/awx/ui/src/screens/Setting/MiscAuthentication/MiscAuthenticationEdit/MiscAuthenticationEdit.js#L246-L249
but the allow metrics is not included there :)

Greetings
Klaas

awx/sso/conf.py
Klaas-: done

Klaas-:
Also the last failed test https://github.com/ansible/awx/actions/runs/4388506123/jobs/7772953312 I am not sure how to correct that one, or is it expected to fail if I add something to the api?

fosterseth:
@Klaas- the schema test is expected to fail since we add this new field to the API. It won't block merging

fosterseth (approved):

mabashian (approved):

fosterseth:
@Klaas- a QE engineer will sign off and merge this soon, and the changes will go into the next AWX release. thanks for your contribution!

jay-steurer:
@Klaas- could you rebase your branch? Thanks!

Klaas-:
> @Klaas- could you rebase your branch? Thanks!

@jay-steurer done

Klaas-:
I don't think those test failures are related to my change

jay-steurer (approved):

PR #13644, opened 2023-03-03
Add scm_branch to inventory source and inventory update

fosterseth:
##### SUMMARY

Allows users to set `scm_branch` on inventory sources (if type is `scm`)

Aims to solve https://github.com/ansible/awx/issues/5692

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API


awx/api/serializers.py
```diff
@@ -2051,7 +2052,7 @@ class InventorySourceSerializer(UnifiedJobTemplateSerializer, InventorySourceOpt
 
     class Meta:
         model = InventorySource
-        fields = ('*', 'name', 'inventory', 'update_on_launch', 'update_cache_timeout', 'source_project') + (
+        fields = ('*', 'name', 'inventory', 'update_on_launch', 'update_cache_timeout', 'source_project', 'scm_branch') + (
```
AlanCoding: This field was already added to `InventorySourceOptionsSerializer`, a parent class of this, so I don't think it should be added again here.

awx/main/models/inventory.py
```diff
@@ -872,6 +872,13 @@ class Meta:
         default='',
         help_text=_('Inventory source variables in YAML or JSON format.'),
     )
+    scm_branch = models.CharField(
+        max_length=1024,
+        default='',
+        blank=True,
+        null=True,
+        help_text=_('Branch to use for inventory source. Project default used if blank. Only allowed if project allow_override field is set to true.'),
```
AlanCoding: very nit-picky, but I would say "inventory update" as opposed to "inventory source", because this field applies to both models, and I think that wording would make sense in both cases. Or just say "Inventory source SCM branch"

AlanCoding:
This looks good, code is plausibly fully-functional on the API side.

 - the checks are complaining that the collection wasn't updated after a new field was added
 - UI needs this added to inventory source page, which is complicated by the source='scm' conditional in the edit field
 - UI also needs this added to inventory update page

fosterseth:
in UI, it does show source branch on the inventory update

image


we still need to add an editable field for the inventory source page itself

awx/api/serializers.py
fosterseth: fixed

awx/main/models/inventory.py
fosterseth: changed the text, thanks

fosterseth:
added scm_branch to the inventory_source collection module

fosterseth:
for testing: @CFSNM 

- create inventory source based on project
- set inventory_source.scm_branch to non-default branch of project (e.g. `foo`)
- launch inventory update, and assert that the revision matches `foo`, and that inventory contains hosts from `foo`
- ideally the hosts on `foo` are different from the hosts on the default branch

fosterseth:
@gamuniz 's UI changes are shown below

inventory source form

image

inventory source detail

image

CFSNM (approved):

AlanCoding (approved):

akus062381:
UI Tests are passing. the five failures are in the visual tests and are not actual defects in the UI itself.

akus062381 (approved):

drdamour:
how do i know what version this enhancement makes it into?

AlanCoding:
Github will list the tags a commit is a part of

https://github.com/ansible/awx/commit/d2aa1b94e348a2a1db7ba3d6640bf55ed7107ead

That shows [22.0.0](https://github.com/ansible/awx/releases/tag/22.0.0) and [21.14.0](https://github.com/ansible/awx/releases/tag/21.14.0)

PR #13652, opened 2023-03-06
[on-premise-analytics]: Analytics API

slemrmartin:
##### SUMMARY

API for forwarding requests to Analytics API on console.redhat.com

* https://issues.redhat.com/browse/AA-1466
* https://issues.redhat.com/browse/AA-1544
* https://issues.redhat.com/browse/AA-1545
* https://issues.redhat.com/browse/AA-1547
* https://issues.redhat.com/browse/AA-1548
* https://issues.redhat.com/browse/AA-1549

It supports:

URL (/api/v2/analytics/…) | Allowed HTTP methods
-- | --
/authorized/ | GET, POST(?)
/reports/ | POST
/report/`<slug>`/ | GET, POST(?)
/report_options/ | GET, POST(?)
/adoption_rate/ | POST
/adoption_rate_options/ | POST
/event_explorer/ | POST
/event_explorer_options/ | POST
/host_explorer/ | POST
/host_explorer_options/ | POST
/job_explorer/ | POST
/job_explorer_options/ | POST
/probe_templates/ | POST
/probe_template_options/ | POST
/probe_template_for_hosts/ | POST
/probe_template_for_hosts_options/ | POST
/roi_templates/ | POST
/roi_templates_options/ | POST

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.12.1.dev12+gcb811fec79.d20230306
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/api/urls/analytics.py
+++ b/awx/api/urls/analytics.py
@@ -1,0 +1,32 @@
+# Copyright (c) 2017 Ansible, Inc.
+# All Rights Reserved.
+
+from django.urls import re_path
+
+import awx.api.views.analytics as analytics
+
+
+urls = [
+    re_path(r'^$', analytics.AnalyticsRootView.as_view(), name='analytics_root_view'),
+    re_path(r'^test/$', analytics.AnalyticsTestList.as_view(), name='analytics_test_list'),
+    re_path(r'^authorized/$', analytics.AnalyticsAuthorizedView.as_view(), name='analytics_authorized'),
+    re_path(r'^reports/$', analytics.AnalyticsReportsList.as_view(), name='analytics_reports_list'),
+    re_path(r'^report/(?P<slug>[\w-]+)/$', analytics.AnalyticsReportDetail.as_view(), name='analytics_report_detail'),
+    re_path(r'^report_options/$', analytics.AnalyticsReportOptionsList.as_view(), name='analytics_report_options_list'),
+    re_path(r'^adoption_rate/$', analytics.AnalyticsAdoptionRateList.as_view(), name='analytics_adoption_rate'),
+    re_path(r'^adoption_rate_options/$', analytics.AnalyticsAdoptionRateList.as_view(), name='analytics_adoption_rate_options'),
+    re_path(r'^event_explorer/$', analytics.AnalyticsEventExplorerList.as_view(), name='analytics_event_explorer'),
+    re_path(r'^event_explorer_options/$', analytics.AnalyticsEventExplorerList.as_view(), name='analytics_event_explorer_options'),
+    re_path(r'^host_explorer/$', analytics.AnalyticsHostExplorerList.as_view(), name='analytics_host_explorer'),
+    re_path(r'^host_explorer_options/$', analytics.AnalyticsHostExplorerList.as_view(), name='analytics_host_explorer_options'),
+    re_path(r'^job_explorer/$', analytics.AnalyticsJobExplorerList.as_view(), name='analytics_job_explorer'),
+    re_path(r'^job_explorer_options/$', analytics.AnalyticsJobExplorerList.as_view(), name='analytics_job_explorer_options'),
+    re_path(r'^probe_templates/$', analytics.AnalyticsProbeTemplatesList.as_view(), name='analytics_probe_templates_explorer'),
+    re_path(r'^probe_templates_options/$', analytics.AnalyticsProbeTemplatesList.as_view(), name='analytics_probe_templates_options'),
+    re_path(r'^probe_template_for_hosts/$', analytics.AnalyticsProbeTemplateForHostsList.as_view(), name='analytics_probe_template_for_hosts_explorer'),
+    re_path(r'^probe_template_for_hosts_options/$', analytics.AnalyticsProbeTemplateForHostsList.as_view(), name='analytics_probe_template_for_hosts_options'),
+    re_path(r'^roi_templates/$', analytics.AnalyticsRoiTemplatesList.as_view(), name='analytics_roi_templates_explorer'),
+    re_path(r'^roi_templates_options/$', analytics.AnalyticsRoiTemplatesList.as_view(), name='analytics_roi_templates_options'),
+]
+
+__all__ = ['urls']
--- a/awx/api/urls/analytics.py
+++ b/awx/api/urls/analytics.py
@@ -8,7 +8,6 @@ import awx.api.views.analytics as analytics
 
 urls = [
     re_path(r'^$', analytics.AnalyticsRootView.as_view(), name='analytics_root_view'),
-    re_path(r'^test/$', analytics.AnalyticsTestList.as_view(), name='analytics_test_list'),
     re_path(r'^authorized/$', analytics.AnalyticsAuthorizedView.as_view(), name='analytics_authorized'),
     re_path(r'^reports/$', analytics.AnalyticsReportsList.as_view(), name='analytics_reports_list'),
     re_path(r'^report/(?P<slug>[\w-]+)/$', analytics.AnalyticsReportDetail.as_view(), name='analytics_report_detail'),
--- a/awx/api/views/analytics.py
+++ b/awx/api/views/analytics.py
@@ -1,0 +1,173 @@
+import requests
+import logging
+
+from django.utils.translation import gettext_lazy as _
+
+from awx.api.generics import APIView, Response
+from awx.api.permissions import IsSystemAdminOrAuditor
+from awx.api.versioning import reverse
+from rest_framework.permissions import AllowAny
+from rest_framework import status
+
+from collections import OrderedDict
+
+# docker-docker API request requires:
+# ```
+# docker network connect koku_default tools_awx_1
+# ```
+AUTOMATION_ANALYTICS_API_URL = "http://automation-analytics-backend_fastapi_1:8080/api/tower-analytics/v1"
+# AUTOMATION_ANALYTICS_API_URL = "http://localhost:8004/api/tower-analytics/v1"
+
+logger = logging.getLogger('awx.api.views')
+
+
+class GetNotAllowedMixin(object):
+    def get(self, request, format=None):
+        return Response(status=status.HTTP_405_METHOD_NOT_ALLOWED)
+
+
+class AnalyticsRootView(APIView):
+    permission_classes = (AllowAny,)
+    name = _('Automation Analytics')
+    swagger_topic = 'Automation Analytics'
+
+    def get(self, request, format=None):
+        data = OrderedDict()
+        data['reports'] = reverse('analytics_reports_list')
+        return Response(data)
+
+
+class AnalyticsTestList(APIView):
+    name = _("Testing")
+    permission_classes = (IsSystemAdminOrAuditor,)
+    swagger_topic = "AA"
+
+    def get(self, request, format=None):
+        logger.info(f"TEST: {type(request.headers)}")
+        new_headers = request.headers.copy()
+
+        data = {
+            'get': {
+                'method': request.method,
+                'content-type': request.content_type,
+                'data': request.data,
+                'query_params': request.query_params,
+                'headers': new_headers,
+                'path': request.path,
+            }
+        }
+        return Response(data)
+
+    def post(self, request, format=None):
+        return self.get(request, format)
+
+
+class AnalyticsGenericView(APIView):
+    """
+    Example:
+        headers = {
+            'Content-Type': 'application/json',
+        }
+
+        params = {
+            'limit': '20',
+            'offset': '0',
+            'sort_by': 'name:asc',
+        }
+
+        json_data = {
+            'limit': '20',
+            'offset': '0',
+            'sort_options': 'name',
+            'sort_order': 'asc',
+            'tags': [],
+            'slug': [],
+            'name': [],
+            'description': '',
+        }
+
+        response = requests.post(f'{AUTOMATION_ANALYTICS_API_URL}/reports/', params=params,
+                                 headers=headers, json=json_data)
+
+        return Response(response.json(), status=response.status_code)
+    """
+
+    permission_classes = (IsSystemAdminOrAuditor,)
+
+    def _remove_api_path_prefix(self, request_path):
+        parts = request_path.split('analytics/')
+        return parts[len(parts) - 1]
+
+    def _forward_get(self, request, format=None):
+        headers = {'Content-Type': 'application/json'}
+        analytics_path = self._remove_api_path_prefix(request.path)
+        response = requests.get(f'{AUTOMATION_ANALYTICS_API_URL}/{analytics_path}', params=request.query_params, headers=headers)
+        return Response(response.json(), status=response.status_code)
+
+    def _forward_post(self, request, format=None):
+        analytics_path = self._remove_api_path_prefix(request.path)
+        response = requests.post(f'{AUTOMATION_ANALYTICS_API_URL}/{analytics_path}', params=request.query_params, headers=request.headers, json=request.data)
+        return Response(response.json(), status=response.status_code)
+
+
+class AnalyticsGenericListView(AnalyticsGenericView):
+    def get(self, request, format=None):
+        return self._forward_get(request, format)
+
+    def post(self, request, format=None):
+        return self._forward_post(request, format)
+
+
+class AnalyticsGenericDetailView(AnalyticsGenericView):
+    def get(self, request, slug, format=None):
+        return self._forward_get(request, format)
+
+    def post(self, request, slug, format=None):
+        return self._forward_post(request, format)
+
+
+class AnalyticsAuthorizedView(AnalyticsGenericListView):
+    name = _("Authorized")
+
+
+class AnalyticsReportsList(AnalyticsGenericListView):
+    name = _("Reports")
+    swagger_topic = "Automation Analytics"
+
+
+class AnalyticsReportDetail(AnalyticsGenericDetailView):
+    name = _("Report")
+
+
+class AnalyticsReportOptionsList(AnalyticsGenericListView):
+    name = _("Report Options")
+
+
+class AnalyticsAdoptionRateList(GetNotAllowedMixin, AnalyticsGenericListView):
+    name = _("Adoption Rate")
+
+
+class AnalyticsEventExplorerList(GetNotAllowedMixin, AnalyticsGenericListView):
+    name = _("Event Explorer")
+
+
+class AnalyticsHostExplorerList(GetNotAllowedMixin, AnalyticsGenericListView):
+    name = _("Host Explorer")
+
+
+class AnalyticsJobExplorerList(GetNotAllowedMixin, AnalyticsGenericListView):
+    """TODO: Allow Http GET?"""
+
+    name = _("Job Explorer")
+
+
+class AnalyticsProbeTemplatesList(GetNotAllowedMixin, AnalyticsGenericListView):
+    name = _("Probe Templates")
+
+
+class AnalyticsProbeTemplateForHostsList(GetNotAllowedMixin, AnalyticsGenericListView):
+    name = _("Probe Template For Hosts")
+
+
+class AnalyticsRoiTemplatesList(GetNotAllowedMixin, AnalyticsGenericListView):
+    name = _("ROI Templates")
--- a/awx/api/views/analytics.py
+++ b/awx/api/views/analytics.py
@@ -1,24 +1,42 @@
 import requests
 import logging
+import urllib.parse as urlparse
 
+from django.conf import settings
 from django.utils.translation import gettext_lazy as _
+from django.utils import translation
 
 from awx.api.generics import APIView, Response
 from awx.api.permissions import IsSystemAdminOrAuditor
 from awx.api.versioning import reverse
+from awx.main.utils import get_awx_version
 from rest_framework.permissions import AllowAny
 from rest_framework import status
 
 from collections import OrderedDict
 
-# docker-docker API request requires:
-# ```
-# docker network connect koku_default tools_awx_1
-# ```
-AUTOMATION_ANALYTICS_API_URL = "http://automation-analytics-backend_fastapi_1:8080/api/tower-analytics/v1"
-# AUTOMATION_ANALYTICS_API_URL = "http://localhost:8004/api/tower-analytics/v1"
+AUTOMATION_ANALYTICS_API_URL_PATH = "/api/tower-analytics/v1"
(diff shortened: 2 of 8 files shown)


awx/api/views/analytics.py
```diff
@@ -0,0 +1,274 @@
...
+            # Unexpected text => still HTTP 404
+            message = response.text
+
+        return self._error_response(ERROR_NOT_FOUND, message, remote=True, remote_status_code=status.HTTP_404_NOT_FOUND, status_code=status.HTTP_404_NOT_FOUND)
+
+    @staticmethod
+    def _forward_response(response):
+        try:
+            content_type = response.headers['content-type'] or ''
+            if content_type.find('application/json') != -1:
```
AlanCoding, line 173: The `if` here confuses me, because if it were false, then it won't hit the `return`, but still not hit the `except`, thus it will return `None` and when I follow this into the parent methods, it's clear that's not properly accounted for.

awx/api/views/analytics.py
```diff
@@ -0,0 +1,274 @@
...
+ERROR_MISSING_URL = "missing-url"
+ERROR_MISSING_USER = "missing-user"
+ERROR_MISSING_PASSWORD = "missing-password"
+ERROR_NO_DATA_OR_ENTITLEMENT = "no-data-or-entitlement"
+ERROR_NOT_FOUND = "not-found"
+ERROR_UNAUTHORIZED = "unauthorized"
+ERROR_UNKNOWN = "unknown"
+ERROR_UNSUPPORTED_METHOD = "unsupported-method"
+
+logger = logging.getLogger('awx.api.views')
```
john-westcott-iv: Can we make this awx.api.view.analytics

awx/api/views/analytics.py
```diff
@@ -0,0 +1,274 @@
...
+            #
+            # Not found, No entitlement or No data in Analytics
+            #
+            elif response.status_code == status.HTTP_404_NOT_FOUND:
+                return self._error_response_404(response)
+            #
+            # Success or not a 401/404 errors are just forwarded
+            #
+            else:
+                return self._forward_response(response)
```
john-westcott-iv, line 220: Is there any chance for redirects here? If so, is that something you want to handle in this so the redirect does not make it back through the AWX API?

awx/api/views/analytics.py
```diff
@@ -0,0 +1,274 @@
...
+from awx.api.versioning import reverse
+from awx.main.utils import get_awx_version
+from rest_framework.permissions import AllowAny
+from rest_framework import status
+
+from collections import OrderedDict
+
+AUTOMATION_ANALYTICS_API_URL_PATH = "/api/tower-analytics/v1"
+
+CERT_PATH = "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
```
john-westcott-iv: Hard coding stuff like this makes me cringe. Do we have confirmation that this file will always be there with the proper content? If not would it be worth making this an AWX setting? It could be True/False/<file path> for different conditions and then maybe we have something like:
```
if <setting is not boolean> and <file does not exist>:
    raise "Argh...."
```

awx/api/views/analytics.py
```diff
@@ -0,0 +1,274 @@
...
+
+
+class AnalyticsRootView(APIView):
+    permission_classes = (AllowAny,)
+    name = _('Automation Analytics')
+    swagger_topic = 'Automation Analytics'
+
+    def get(self, request, format=None):
+        data = OrderedDict()
+        data['reports'] = reverse('analytics_reports_list')
```
john-westcott-iv: Do we want to show the reports at this view or should we show a list of other endpoints like the main `api/v2/` page?

awx/api/views/analytics.py
```diff
@@ -0,0 +1,274 @@
...
+        headers['X-Rh-Analytics-Source'] = 'controller'
+        headers['X-Rh-Analytics-Source-Version'] = get_awx_version()
+        headers['Accept-Language'] = translation.get_language()
+
+        return headers
+
+    @staticmethod
+    def _get_analytics_path(request_path):
+        parts = request_path.split('analytics/')
+        path_specific = parts[len(parts) - 1]
```
john-westcott-iv: `parts[-1]` should give you what you need here, no need for the `len`.
You could even just do:
```
path_specific = request_path.split('analytics/')[-1]
```

It would be really cool if we could figure out how to variablize 'analytics' here and in `urls.py` (line 137). Not that they will likely ever change.

awx/api/views/analytics.py
```diff
@@ -0,0 +1,274 @@
...
+    def _send_to_analytics(self, request, method):
+        try:
+            headers = self._request_headers(request)
+
+            self._check_upload_enabled()
+            url = self._get_analytics_url(request.path)
+            rh_user = self._get_rh_user()
+            rh_password = self._get_rh_password()
+
+            if method == "GET":
```
john-westcott-iv: This could be simplified to something like:
```
if method not in ['GET', 'POST']:
    return self._error_response(ERROR_UNSUPPORTED_METHOD, method, remote=False, status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)

response = requests.request(method, url, auth=(rh_user, rh_password), verify=CERT_PATH, params=request.query_params, headers=headers, json=request.data)
```

awx/api/views/analytics.py
```diff
@@ -0,0 +1,274 @@
...
+                response = requests.get(url, auth=(rh_user, rh_password), verify=CERT_PATH, params=request.query_params, headers=headers, json=request.data)
+            elif method == "POST":
+                response = requests.post(url, auth=(rh_user, rh_password), verify=CERT_PATH, params=request.query_params, headers=headers, json=request.data)
```
john-westcott-iv: Also, we should probably add a timeout to these requests since this is a blocking call with no timeout. Not sure if you want to make this configurable or not. If we catch a timeout we can throw a 504 or something along those lines.

awx/api/views/analytics.py
```diff
@@ -0,0 +1,274 @@
...
+            elif response.status_code == status.HTTP_404_NOT_FOUND:
+                return self._error_response_404(response)
+            #
+            # Success or not a 401/404 errors are just forwarded
+            #
+            else:
+                return self._forward_response(response)
+
+        except MissingSettings as e:
+            return self._error_response(e.args[0], remote=False)
```
john-westcott-iv, line 224: It might not be also log an info or warn here to indicate which setting is missing in the log.

awx/api/views/analytics.py
```diff
@@ -0,0 +1,274 @@
...
+class AnalyticsEventExplorerList(GetNotAllowedMixin, AnalyticsGenericListView):
+    name = _("Event Explorer")
+
+
+class AnalyticsHostExplorerList(GetNotAllowedMixin, AnalyticsGenericListView):
+    name = _("Host Explorer")
+
+
+class AnalyticsJobExplorerList(GetNotAllowedMixin, AnalyticsGenericListView):
+    """TODO: Allow Http GET?"""
```
john-westcott-iv: Can we resolve this TODO before merge?

awx/api/views/analytics.py
```diff
@@ -0,0 +1,274 @@
...
+
+class AnalyticsGenericListView(AnalyticsGenericView):
+    def get(self, request, format=None):
+        return self._send_to_analytics(request, method="GET")
+
+    def post(self, request, format=None):
+        return self._send_to_analytics(request, method="POST")
+
+
+class AnalyticsGenericDetailView(AnalyticsGenericView):
```
john-westcott-iv, line 244: This class is only used in one place. Are we thinking that we might have additional endpoints someday and that is why its separated out?

awx/api/views/analytics.py
```diff
@@ -0,0 +1,274 @@
...
+        analytics_path = self._get_analytics_path(request_path)
+        url = getattr(settings, 'AUTOMATION_ANALYTICS_URL', None)
+        if not url:
+            raise MissingSettings(ERROR_MISSING_URL)
+        url_parts = urlparse.urlsplit(url)
+        analytics_url = urlparse.urlunsplit([url_parts.scheme, url_parts.netloc, analytics_path, url_parts.query, url_parts.fragment])
+        return analytics_url
+
+    @staticmethod
+    def _check_upload_enabled():
```
john-westcott-iv: It would be good if we can add tests for the methods in here as well.

awx/api/views/analytics.py
slemrmartin: it won't return `None`, there is a return in the end of this method (line 169) covering `text/plain`, json errors etc. responses.

awx/api/views/analytics.py
slemrmartin: c.rh.c. returns 307 response if we don't add final slash to the url. But UI is aware about that and doesn't do that, I think it's not needed to handle.

awx/api/views/analytics.py
slemrmartin: This path doesn't introduce something new. it's hardcoded also here: https://github.com/ansible/awx/blob/devel/awx/main/analytics/core.py#L363

It's Red Hat IT Root CA certificate provided by installer.

awx/api/views/analytics.py
slemrmartin: this url is quite unclear for me. the `/api/v2/` is clear, but there are almost all endpoints POST only. 
List of reports can't be shown, because it doesn't go from database, but from c.th.c. Analytics API.
Is it a *must* to have it, or can I just remove the `/api/v2/analytics/` endpoint at all?

awx/api/views/analytics.py
slemrmartin: I'll just remove the comment, it won't be definitelly allowed this time

awx/api/views/analytics.py
slemrmartin: yes, I expect there could be more `/<path>/<slug>` endpoints

awx/api/views/analytics.py
slemrmartin: I agree, but we are in a time pressure. So we agreed we'll make tests later. Yolo and Slowyo etc. will be enough

awx/api/views/analytics.py
slemrmartin: I'll add warning log

awx/api/views/analytics.py
john-westcott-iv: But if you just return the response and its a 307 the location header will be passed back to the AWX user and they would get redirected from `https://<awx>/api/v2/<whatever>` to `https://c.rh.c/<whatever new>` right? Wouldn't that be a bad experience if you were typing in a URL to the browsable AWX API?

awx/api/views/analytics.py
john-westcott-iv: Are there any use cases where we would need a different cert? Like an onprem analytics server or a proxy to get outside or something like that? If not, I guess we can go with this as is.

awx/api/views/analytics.py
john-westcott-iv: Let me pull the code down and play around with it for a little and I will get back to you on this.

awx/api/views/analytics.py
slemrmartin: I'm not aware of the need of different cert. 
- Also the same path is required for local dev computers. Described here: https://redhat.service-now.com/help?id=kb_article_view&sysparm_article=KB0000183.
- Also this file contains multiple certificates (their public keys)
- it seems like it's standardized path

slemrmartin: anyway I noticed one c.rh.c. constant was moved to setting, so I'll do it with this too

awx/api/views/analytics.py
slemrmartin: 307 redirection from cloud is handled by default for GET and POST (https://requests.readthedocs.io/en/latest/user/quickstart/#redirection-and-history).  

In case of request to `'https://localhost:8043/api/v2/analytics/report/most_used_modules'` (Controller's URL without slash), it returns 301 with correct AWX location. 

**Tested.**

awx/api/views/analytics.py
slemrmartin: I removed the analytics root endpoint

awx/api/views/analytics.py
john-westcott-iv: Perfect, thanks!

awx/api/views/analytics.py
```diff
@@ -0,0 +1,274 @@
...
+class AnalyticsGenericDetailView(AnalyticsGenericView):
+    def get(self, request, slug, format=None):
+        return self._send_to_analytics(request, method="GET")
+
+    def post(self, request, slug, format=None):
+        return self._send_to_analytics(request, method="POST")
+
+
+class AnalyticsAuthorizedView(AnalyticsGenericListView):
+    name = _("Authorized")
```
john-westcott-iv, line 231: If I hit the authorized endpoint with a bad username/password I get this as an application/json response:
```
"<?xml version=\"1.0\" encoding=\"iso-8859-1\"?>\n<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Transitional//EN\"\n         \"[http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd\](http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd%5C)">\n<html xmlns=\"[http://www.w3.org/1999/xhtml\](http://www.w3.org/1999/xhtml%5C)" xml:lang=\"en\" lang=\"en\">\n\t<head>\n\t\t<title>400 - Bad Request</title>\n\t</head>\n\t<body>\n\t\t<h1>400 - Bad Request</h1>\n\t</body>\n</html>\n"
```

john-westcott-iv, line 256: If I set a username/password but "forget" to change the default `example.com` (in AWX) I get this response as application/json:
```
"<?xml version=\"1.0\" encoding=\"iso-8859-1\"?>\n<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Transitional//EN\"\n         \"[http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd\](http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd%5C)">\n<html xmlns=\"[http://www.w3.org/1999/xhtml\](http://www.w3.org/1999/xhtml%5C)" xml:lang=\"en\" lang=\"en\">\n\t<head>\n\t\t<title>400 - Bad Request</title>\n\t</head>\n\t<body>\n\t\t<h1>400 - Bad Request</h1>\n\t</body>\n</html>\n"
```

Not sure if we want to try and intercept this or look at the URL and if its `example.com` say its invalid.

slemrmartin: I'll take a look where this come from, it doesn't look like response from Analytics views. 
Anyway I don't want to hardcode specifically `example.com`  (moreover https://example.com/api/v1/tower-analytics/authorized/ returns 404, not 400)
I

awx/api/urls/analytics.py
```diff
@@ -0,0 +1,31 @@
...
+from django.urls import re_path
+
+import awx.api.views.analytics as analytics
+
+
+urls = [
+    re_path(r'^$', analytics.AnalyticsRootView.as_view(), name='analytics_root_view'),
+    re_path(r'^authorized/$', analytics.AnalyticsAuthorizedView.as_view(), name='analytics_authorized'),
+    re_path(r'^reports/$', analytics.AnalyticsReportsList.as_view(), name='analytics_reports_list'),
+    re_path(r'^report/(?P<slug>[\w-]+)/$', analytics.AnalyticsReportDetail.as_view(), name='analytics_report_detail'),
```
john-westcott-iv, line 13: Do you have an example of a slug name? One of reports had a slug of `automation_calculator` and when I tried to do `api/v2/analytics/reports/automation_calculator/` I ended up with a 404.

awx/api/urls/analytics.py
```diff
@@ -0,0 +1,31 @@
...
+
+from django.urls import re_path
+
+import awx.api.views.analytics as analytics
+
+
+urls = [
+    re_path(r'^$', analytics.AnalyticsRootView.as_view(), name='analytics_root_view'),
+    re_path(r'^authorized/$', analytics.AnalyticsAuthorizedView.as_view(), name='analytics_authorized'),
+    re_path(r'^reports/$', analytics.AnalyticsReportsList.as_view(), name='analytics_reports_list'),
```
john-westcott-iv, line 12: If I pass `limit` of 1 I still get 5 reports back. Should that only allow one to come back? The slugs on the post seem to work to this endpoint. Name does not, description does. We should probably try more.

john-westcott-iv, line 12: Also, on the reports page there is a section like:
```
"links": {
        "first": "[/api/tower-analytics/v1/reports/?offset=0](https://localhost:8043/api/tower-analytics/v1/reports/?offset=0)",
        "last": "[/api/tower-analytics/v1/reports/?offset=10](https://localhost:8043/api/tower-analytics/v1/reports/?offset=10)",
        "next": "[/api/tower-analytics/v1/reports/?offset=5](https://localhost:8043/api/tower-analytics/v1/reports/?offset=5)",
        "prev": null
    },
    ```
    Those links obviously don't work when clicked on. Do we need to transform those URLS?

File: awx/api/views/analytics.py [hunk 16]
```diff
@@ -0,0 +1,274 @@
...
+                headers[header] = request.headers.get(header)
+        headers['X-Rh-Analytics-Source'] = 'controller'
+        headers['X-Rh-Analytics-Source-Version'] = get_awx_version()
+        headers['Accept-Language'] = translation.get_language()
+
+        return headers
+
+    @staticmethod
+    def _get_analytics_path(request_path):
+        parts = request_path.split('analytics/')
```
john-westcott-iv: request_path should only ever come from request.path but if it happens to be `None` you get an exception here. Not sure if you want to make this a little more defensive incase anyone does something weird with this.

awx/api/urls/analytics.py
slemrmartin: there are `/reports/` (list only) and `/report/` (detail only)

slemrmartin: `/report/automation_calculator` should work, `report/most_used_modules` is another example

awx/api/urls/analytics.py
slemrmartin: can you post your query? It works for me as expected.

slemrmartin: hm, it means parsing of json responses, let me look into that

awx/api/views/analytics.py
slemrmartin: ok I'll handle it

awx/api/urls/analytics.py
john-westcott-iv: In talking this out with the team yesterday we think the API can handle this by fully qualifying any links from these endpoints with `settings.AUTOMATION_ANALYTICS_URL`.

john-westcott-iv: From `/api/v2/analytics/reports/` just post `{"limit": 1}`

awx/api/views/analytics.py
slemrmartin: I'm thinking about this and the Exception is appropriate. 
It's a "private" method. Together with defined URLs it's not possible to send anything else to these views in request.path. So if it happens, it's not a user fail, but programmer's future mistake.

awx/api/urls/analytics.py
slemrmartin: hm, the "limit" in POST data doesn't have effect. it has to be sent as url query param (I don't understand, why Analytics UI sends both)

awx/api/views/analytics.py
slemrmartin: When I "forgot" the default value `https://example.org` I get that response ^ wrapped to 404:

```
{
    "error": {
        "keyword": "not-found",
        "message": "<?xml version=\"1.0\" encoding=\"iso-8859-1\"?>\n<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Transitional//EN\"\n         \"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd\">\n<html xmlns=\"http://www.w3.org/1999/xhtml\" xml:lang=\"en\" lang=\"en\">\n\t<head>\n\t\t<title>404 - Not Found</title>\n\t</head>\n\t<body>\n\t\t<h1>404 - Not Found</h1>\n\t\t<script type=\"text/javascript\" src=\"//obj.ac.bcon.ecdns.net/ec_tpm_bcon.js\"></script>\n\t</body>\n</html>\n",
        "remote": true,
        "remote_status": 404
    }
}
```

as expected

awx/api/views/analytics.py
slemrmartin: added back by john's commit

awx/api/views/analytics.py
john-westcott-iv: Do we want to do any kind of validation? i.e. make sure the URL host ends in redhat.com (if that is where analytics will live)? Or is this exposed to allow people to proxy requests out?

awx/api/urls/analytics.py
john-westcott-iv: I see what happened. I had `reports` (with an s). This is our normal pattern, `/api/v2/users/` is a list view and `/api/v2/users/1/` is the detail of the user. We are ok with this being this way as you are just proxying back to analytics.

awx/api/urls/analytics.py
john-westcott-iv: Will this break anything with the UI or are we ok with this behavior?

awx/api/views/analytics.py
john-westcott-iv: Can we get rid of this variable to close out this conversation?

awx/api/views/analytics.py
john-westcott-iv: I'm ok with this. We can leave it as is.

awx/api/views/analytics.py
slemrmartin: I discussed it and no, we don't want additional validation here. Enough is the certificate validation with parameter `verify=settings.INSIGHTS_CERT_PATH` in GET/POST/OPTION

awx/api/views/analytics.py
slemrmartin: Sure, it's not used now, I only forgot to remove it

awx/api/urls/analytics.py
slemrmartin: it shouldn't break anything. pagination works in on-prem UI

PR #13658, opened 2023-03-08
Use different dockerfile for docker-compose-build

TheRealHaoLiu:
##### SUMMARY
one of my previous commit turn Dockerfile to a PHONY make target, to avoid collision between `make docker-compose-build` and `make awx-kube-build` this is a better approach

##### ISSUE TYPE

##### COMPONENT NAME
 - Other

##### AWX VERSION

```
awx: 21.13.1.dev13+ga9b60cc3e4
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -556,12 +556,21 @@ docker-compose-container-group-clean:
 	fi
 	rm -rf tools/docker-compose-minikube/_sources/
 
-## Base development image build
-docker-compose-build:
-	ansible-playbook tools/ansible/dockerfile.yml -e build_dev=True -e receptor_image=$(RECEPTOR_IMAGE)
-	DOCKER_BUILDKIT=1 docker build -t $(DEVEL_IMAGE_NAME) \
-	    --build-arg BUILDKIT_INLINE_CACHE=1 \
-	    --cache-from=$(DEV_DOCKER_TAG_BASE)/awx_devel:$(COMPOSE_TAG) .
+.PHONY: Dockerfile.dev
+## Generate Dockerfile.dev for awx_devel image
+Dockerfile.dev: tools/ansible/roles/dockerfile/templates/Dockerfile.j2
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -528,12 +528,20 @@ docker-compose-container-group-clean:
...
+		-e build_dev=True \
+		-e receptor_image=$(RECEPTOR_IMAGE)
+
 ## Base development image build
-docker-compose-build:
-	ansible-playbook tools/ansible/dockerfile.yml -e build_dev=True -e receptor_image=$(RECEPTOR_IMAGE)
-	DOCKER_BUILDKIT=1 docker build -t $(DEVEL_IMAGE_NAME) \
-	    --build-arg BUILDKIT_INLINE_CACHE=1 \
-	    --cache-from=$(DEV_DOCKER_TAG_BASE)/awx_devel:$(COMPOSE_TAG) .
+docker-compose-build: Dockerfile.dev
```
AlanCoding, line 568: Wouldn't this use a stale `Dockerfile.dev` unless you manually deleted it or touched it? That feels like it would inevitably give undesired behavior, since over time, people are almost sure to have a stale one around.

TheRealHaoLiu: that's how all the other Dockerfile target works,

since generating the Dockerfile is not very expensive what if we just turn all the Dockerfile targets PHONY? this way we avoid having the stale Dockerfile problem altogether

john-westcott-iv (approved):
Pending Alans comments.

gamuniz (approved):

AlanCoding (approved):

AlanCoding:
I think my comment was addressed by having it added to .PHONY

AlanCoding:
...hit the wrong button.

TheRealHaoLiu:
https://github.com/ansible/awx/issues/13737

PR #12627, opened 2022-08-05
Added domain entry and authorizer for TSS

vician:
##### SUMMARY
Adding domain option to Thycotic SecretServer credential plugin.

We hope this would get into Ansible Automation Platform as well.

##### ISSUE TYPE
 - Enhanced Feature
 - New or Enhanced Feature

##### COMPONENT NAME
awx/main/credential_plugins/tss.py

##### AWX VERSION

```
21.4.0
```

##### ADDITIONAL INFORMATION
It's very similar to approach of `community.general.tss` here https://github.com/ansible-collections/community.general/pull/3228

PR #13660, opened 2023-03-08
Introducing tech preview of the new AWX UI

TheRealHaoLiu:
##### SUMMARY
- Makefile file and make targets for pulling in and build new UI from https://github.com/ansible/ansible-ui
- Add banner switch between old UI and new UI
- Add new setting UI_NEXT to enable/disable new UI
- Add new config to display the setting for UI_NEXT (for non-admin)
- Add new path `/ui_next` to access the new UI

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API
 - UI
 - Other

##### AWX VERSION
```
awx: 21.13.1.dev30+ge18a755f6b
```

##### TEST SCENARIOS
Test environment: fedora 37 VM setup using https://github.com/TheRealHaoLiu/fedora-setup/tree/main
docker-compose related  tests:
🟢 able to start API server without building ui or ui-next in docker-compose development environment (verify via access to /api/v2/ping)
🟢 able to build ui and ui-next on machine directly (verified via `make ui-devel && make ui-next`)
🟢 able to build ui and ui-next in docker-compose development environment (ui-next src from git)
🟢 after building ui without building ui-next with settings.UI_NEXT enabled verified that the placehold index page show up (telling people that UI_NEXT was not built yet)
🟢 after building ui and ui-next collectstatic is able to collect the static files successfully and ui show up

kube build related tests:
🟢 able to build awx with `make awx-kube-build` and old and new UI appears successfully (with spud logo)
🟢 CI feature branch built image function as expected with both old and new UI appear successfully (with spud logo)

release build related tests:
🟢 moc stage job was able to successfully build and push the moc release image https://github.com/ansible/awx/actions/runs/4443351332/jobs/7800534464
🟢 deployment using the image generated by stage CI job produced image is successful both old and new ui function (with AWX logo)

Function related tests:
🟢 /ui_next show 404 when settings.UI_NEXT = False
🟢 banner on old UI for switching to new UI does not appear when settings.UI_NEXT = False
🟢 banner on old UI for switching to new UI appear for normal user when settings.UI_NEXT = True


awx/main/conf.py
```diff
@@ -795,6 +795,16 @@
     category_slug='bulk',
 )
 
+register(
+    'UI_NEXT',
+    field_class=fields.BooleanField,
+    default=False,
+    label=_('Enable Next Generation User Interface'),
+    help_text=_('Enable the next generation user interface.'),
```
fosterseth: optional suggestion:

"next generation" has other colloquial meanings, for example to indicate "cutting edge"

maybe we can consider "Enable experimental user interface"

fosterseth: or "tech preview" as indicated elsewhere

TheRealHaoLiu: updated

gamuniz (approved):

awx/ui/src/screens/Setting/shared/data.allSettingOptions.json
```diff
@@ -639,6 +639,15 @@
         "unit": "seconds",
         "default": 14400
       },
+      "UI_NEXT": {
+        "type": "boolean",
+        "required": false,
+        "label": "Enable Next Generation User Interface",
```
fosterseth: should be "preview of new" here and a few other spots

fosterseth (approved):
pulled it down and tested out the instructions in the readme. Was able to build new UI successfully, navigate to new UI, and tested out the banner functionality to flip between old and new UI

Makefile
```diff
@@ -449,7 +451,7 @@ HEADLESS ?= no
 ifeq ($(HEADLESS), yes)
 dist/$(SDIST_TAR_FILE):
 else
-dist/$(SDIST_TAR_FILE): $(UI_BUILD_FLAG_FILE)
+dist/$(SDIST_TAR_FILE): $(UI_BUILD_FLAG_FILE) ui-next
```
john-westcott-iv, line 455: Can we add a comment in here that ui-next is an imported target? It might save someone some time in the future.

awx/ui/public/static/media/192.png
john-westcott-iv, line 1: We are adding these files specifically for the new UI but then rebranding them as AWX on release, is that correct?

awx/ui/src/screens/Setting/MiscSystem/MiscSystemDetail/MiscSystemDetail.test.js
```diff
@@ -113,6 +114,7 @@ describe('<MiscSystemDetail />', () => {
     assertVariableDetail(wrapper, 'Remote Host Headers', '[]');
     assertVariableDetail(wrapper, 'Proxy IP Allowed List', '[]');
     assertDetail(wrapper, 'Global default execution environment', 'Foo');
+    assertDetail(wrapper, 'Enable Next Generation User Interface', 'On');
```
john-westcott-iv: Did we want to change this value to be consistent with the wording in `awx/main/conf.py` (Enable Preview of New User Interface)?

awx/ui_next/Makefile
```diff
@@ -0,0 +1,97 @@
...
+UI_NEXT_GIT_REPO ?= https://github.com/ansible/ansible-ui.git
+UI_NEXT_GIT_BRANCH ?= main
+
+.PHONY: ui-next
+## Default build target of ui-next Makefile, builds ui-next/build
+ui-next: ui-next/build
+
+.PHONY: ui-next/build
+## Build ui-next/build
+ui-next/build: $(UI_NEXT_DIR)/build
```
john-westcott-iv, line 18: Is this target necessary? Can ui-next just call `${UI_NEXT_DIR}/build` directly?

awx/ui_next/Makefile
```diff
@@ -0,0 +1,97 @@
...
+				echo "=== Cloning $(UI_NEXT_DIR)/src from $(UI_NEXT_GIT_BRANCH) of $(UI_NEXT_GIT_REPO) ==="; \
+				rm -rf $(UI_NEXT_DIR)/src; \
+				git clone --depth 1 --branch $(UI_NEXT_GIT_BRANCH) $(UI_NEXT_GIT_REPO) $(UI_NEXT_DIR)/src || true; \
+			fi; \
+		else \
+			echo "=== Cloning $(UI_NEXT_DIR)/src from $(UI_NEXT_GIT_BRANCH) of $(UI_NEXT_GIT_REPO) ==="; \
+			git clone --depth 1 --branch $(UI_NEXT_GIT_BRANCH) $(UI_NEXT_GIT_REPO) $(UI_NEXT_DIR)/src || true; \
+		fi; \
+	else \
+		echo "SKIP: ui-next/src. UI_NEXT_LOCAL and UI_NEXT_GIT_REPO are not set."; \
```
john-westcott-iv: Should we have an exit status code here since we can't actually make this target?

awx/ui_next/Makefile
```diff
@@ -0,0 +1,97 @@
...
+		fi; \
+	elif [ ! -z "$(UI_NEXT_GIT_REPO)" ]; then \
+		if [ -d $(UI_NEXT_DIR)/src ]; then \
+			GIT_REMOTE_ORIGIN=$$(cd $(UI_NEXT_DIR)/src && git remote get-url origin); \
+			GIT_REMOTE_BRANCH=$$(cd $(UI_NEXT_DIR)/src && git rev-parse --abbrev-ref HEAD); \
+			if [ "$$GIT_REMOTE_ORIGIN" = "$(UI_NEXT_GIT_REPO)" ] && [ "$$GIT_REMOTE_BRANCH" = "$(UI_NEXT_GIT_BRANCH)" ]; then \
+				echo "=== Updating $(UI_NEXT_DIR)/src from $(UI_NEXT_GIT_BRANCH) of $(UI_NEXT_GIT_REPO) ==="; \
+				git fetch && git pull; \
+			else \
+				echo "=== Cloning $(UI_NEXT_DIR)/src from $(UI_NEXT_GIT_BRANCH) of $(UI_NEXT_GIT_REPO) ==="; \
```
john-westcott-iv, line 67: This and the outer else feels like it could be a little reworked to reduce redundency. Can the if above just `exit 0` and then thise else just `rm /src` and then remove the lower else so that it just echo's and clones? (or something along those lines).

awx/ui_next/README.md
```diff
@@ -0,0 +1,39 @@
+# Instruction to build ui_next directly from this directory
+
+## Set src of the ui_next repo
+
+### via GIT
```
john-westcott-iv: Since the Makefile will use the default repo could we say something like `from a custom GIT repo`

awx/ui_next/README.md
```diff
@@ -0,0 +1,39 @@
...
+export UI_NEXT_GIT_BRANCH_REPO=git@<git repo>
+```
+
+optionally set branch (default is main)
+
+```bash
+export UI_NEXT_GIT_BRANCH_BRANCH=main
+```
+
+### via symlink to existing clone
```
john-westcott-iv: Can we say something here about which has prescidence if both UI_NEXT_GIT_BRANCH_REPO and UI_NEXT_LOCAL are both set?

awx/ui_next/README.md
john-westcott-iv: Also, it looks like a bad find and replace happened. This should be `UI_NEXT_GIT_REPO` and the branch should just be `UI_NEXT_GIT_BRANCH`

john-westcott-iv (commented):
Should we add/change anything in https://github.com/ansible/awx/blob/devel/awx/ui/CONTRIBUTING.md?

john-westcott-iv:
When you turn on/off the setting in the UI settings page and then go back to the dashboard view the message does not toggle until you do a full reload of the page. Is there a way to have it re-check if the header should be on/off on a dashboard load?

john-westcott-iv:
I put on my user hat and went through the default motions:
```
clone repo
make ui-devel
make docker-compose-build
make docker-compose
```

When I click on the link to load the new dashboard I get a 500 because the template does not exist for the view:
```
TemplateDoesNotExist at /ui_next/dashboard/
index_awx.html
```

Can we add a default template that has a message like "Did you forget to build the UI Next interface"?

john-westcott-iv:
In the blue banner in UI Next it references `Ansible Automation Platform` should that be changed to AWX for community stuff?

john-westcott-iv:
I had awx running and then did `make ui-next` and ended up with a blank screen until I redid `make docker-compose-build && make docker-compose`. Is that expected? I'm thinking yes because the index file was there to serve the HTML but collectstatic needs to be re-run to pickup all of the additional files. If thats accurate can we add it to whatever build documentation we are going to (hopefully) add?

tools/docker-compose/bootstrap_development.sh
```diff
@@ -25,6 +25,7 @@ fi
 
 # Make sure that the UI static file directory exists, Django complains otherwise.
 mkdir -p /awx_devel/awx/ui/build/static
+mkdir -p /awx_devel/awx/ui_next/build/awx
```
john-westcott-iv: Since we added a default template, I think this line should no longer be needed.

Makefile
jjwatt: Yes, and a "MaybeImported" target at that :). Please leave a comment about that or else folks might be grepping for a `ui-next` recipe in this file and feeling hurt that it's not there.

awx/ui_next/Makefile
```diff
@@ -0,0 +1,97 @@
...
+.PHONY: ui-next/src
+ui-next/src: $(UI_NEXT_DIR)/src
+
+.PHONY: $(UI_NEXT_DIR)/src
+## Real target for ui-next/src.
+$(UI_NEXT_DIR)/src:
+	@echo "=== Setting up $(UI_NEXT_DIR)/src ==="
+	@if [ ! -z "$(UI_NEXT_LOCAL)" ]; then \
+		if [ -d $(UI_NEXT_DIR)/src ]; then \
+			if [ "$$(readlink $(UI_NEXT_DIR)/src)" = "$(UI_NEXT_LOCAL)" ]; then \
```
jjwatt, line 48: Let's try not to use `readlink` so we're inclusive of BSD (and macOS, I guess).

awx/ui_next/Makefile
```diff
@@ -0,0 +1,97 @@
...
+.PHONY: ui-next
+## Default build target of ui-next Makefile, builds ui-next/build
+ui-next: ui-next/build
+
+.PHONY: ui-next/build
+## Build ui-next/build
+ui-next/build: $(UI_NEXT_DIR)/build
+
+## Real target for ui-next.
+$(UI_NEXT_DIR)/build: ui-next/src/build
```
jjwatt, line 21: Somewhat in line with John's comment: It's not always clear which of these targets are meant to be PHONY and which are meant to be real targets. And, more importantly to functionality:
1. Real targets should probably never depend on phony targets because PHONY targets will always evaluate as needing updating, and so they'll make the "real" targets always need updating, too, since the real targets *depend* on the always-out-of-date PHONY target, and if something might have changed with it, then something might have changed with our target and...
2. Real directories should almost never be real targets
Similar, but distinct reasoning: A POSIX directory is "updated" or "changed" anytime any metadata about it changes (!), so that would include any files atime (on certain filesystems), new/removed directory, etc. etc. and GNU Make looks at that as a trigger to update the target, so directories will often need updating.
It's not immediately super clear to me what the intent is, so I'm not sure what to recommend instead, but it's late on a Friday. I wouldn't block merging this if it works, but would definitely want to rewrite a lot of it in a followup commit.

awx/ui_next/Makefile
jjwatt: Again, echoing John's comment: this has warts. The first code smell is that it's just too long. In fact, that would cause `make` to fail in some environments because of argument list length restrictions! Make essentially passes everything as one long line as arguments to `sh -c`. Yes, it could use reworking, but even then it's probably doing too much. My first thought is just *lift*. Lift it to a helper script, possibly with params--you can always pass make macro vars `$()` into the helper script as-needed. Then, I don't even care that it's not great script, and it will be more likely to be improved later, will be picked up by shell linters automatically, etc. As an alternative to that, we could probably write much of the functionality in actual Make code.
That's a lie, tho. My first thought is really to let it in if it works well enough :)
But, if and when we want to fix it, we need to simplify it here and/or move it to a script.

jjwatt (approved):

awx/ui_next/Makefile
TheRealHaoLiu: this one works for macOS... i verified with BSD readlink

awx/ui_next/Makefile
TheRealHaoLiu: its not necessary...i added this for symetry...

awx/ui_next/Makefile
TheRealHaoLiu: fixed now fail `make src` if UI_NEXT_LOCAL and UI_NEXT_GIT_REPO not set

awx/ui_next/Makefile
TheRealHaoLiu: as discussed previously will address this outside of this PR...yes i hate this blob of bash too

awx/ui_next/README.md
TheRealHaoLiu: fixed

awx/ui_next/README.md
TheRealHaoLiu: fixed

tools/docker-compose/bootstrap_development.sh
TheRealHaoLiu: i ended up moving the placeholder index_awx.html out because 
image

but great idea and thanks for the commit

awx/ui/src/screens/Setting/shared/data.allSettingOptions.json
TheRealHaoLiu: updated

TheRealHaoLiu: fixed

awx/ui/public/static/media/192.png
TheRealHaoLiu: these file are for devel build for release build the icon come from https://github.com/ansible/awx-logos will override what's here

awx/ui_next/Makefile
TheRealHaoLiu: explicitly PHONY the make targets

awx/ui_next/Makefile
TheRealHaoLiu: added TODO to address this later

awx/ui/src/screens/Setting/MiscSystem/MiscSystemDetail/MiscSystemDetail.test.js
TheRealHaoLiu: fixed

Makefile
TheRealHaoLiu: added note to indicate that `ui-next` is imported

kialam:
> In the blue banner in UI Next it references `Ansible Automation Platform` should that be changed to AWX for community stuff?

Yes, those changes have already been made.

kialam:
> When you turn on/off the setting in the UI settings page and then go back to the dashboard view the message does not toggle until you do a full reload of the page. Is there a way to have it re-check if the header should be on/off on a dashboard load?

It would require another API call to get /config in order to refresh the banner. The banner doesn't auto update unless a refresh is called for now, but can revisit this.

john-westcott-iv (approved):

.gitignore
```diff
@@ -157,7 +157,11 @@ use_dev_supervisor.txt
 *.unison.tmp
 *.#
 /awx/ui/.ui-built
+/awx/ui_next/.ui-built
```
TheRealHaoLiu: remove this line that file is no longer needed

awx/ui_next/Makefile
```diff
@@ -0,0 +1,102 @@
+## UI_NEXT_DIR: Relative path to the directory containing this Makefile
+UI_NEXT_DIR := $(patsubst %/,%,$(dir $(lastword $(MAKEFILE_LIST))))
+
+# ## Path to your local clone of the UI_NEXT repo
```
TheRealHaoLiu, line 4: remove #12627

Zokormazo (approved):

rooftopcellist (approved):

akus062381 (approved):

PR #13661, opened 2023-03-08
Adds support for a pseudolocalization and lang query params

mabashian:
##### SUMMARY
This work follows https://github.com/ansible/awx/pull/9537.  The changes here allow a developer to add ?pseudolocalization=true to the query parameters in the URL to check to see if a rendered string has been marked for translation.  All strings marked for translation will be wrapped in carats like so: `>>string<<`.  Unmarked strings will render without the carats.  Note that a hard reload is required for the changes to take effect.  Also note that this only works for strings _defined in the UI_.  Strings that get rendered in the interface that come from API call responses will not be wrapped in carats.

pseudolocalization

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui/src/App.js
+++ b/awx/ui/src/App.js
@@ -142,13 +142,8 @@ function App() {
   const searchParams = Object.fromEntries(new URLSearchParams(search));
   const pseudolocalization =
     searchParams.pseudolocalization === 'true' || false;
-  let language = getLanguageWithoutRegionCode(navigator);
-
-  if (!Object.keys(locales).includes(language)) {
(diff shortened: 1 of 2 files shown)


mabashian:
Also added support for `?lang=` to allow a user to force the ui to render in a particular language regardless of the specified browser language.

lang

keithjgrant (approved):

marshmalien (approved):

jay-steurer:
Verified by QE

PR #13664, opened 2023-03-08
isolate parallel tests to file by file in awx PR

kdelee:
we've  noticed some flake, specifically around "test_lost_worker_autoscale" that could be sensitive race conditions caused by test running along side it that consumes too many resources.

Trying out this config for pytest that loads tests file-by-file so at least we have better idea what tests its running along side of.

### ISSUE TYPE

### COMPONENT NAME
Other


jainnikhil30:
Seems like the dispatcher test is still randomly failing. Looking at the failure it seems we have 2 dispatcher workers but the test is expecting it to be 1.

May be in bulk api tests we are running 10 jobs which is causing dispatcher to autoscale the workers. We can reduce that 10 jobs to maybe 3 or 5 and see how it goes ? 

https://github.com/ansible/awx/blob/devel/awx/main/tests/functional/test_bulk.py#L83

Instead of 10 lets do 3 or 5 ?

PR #13666, opened 2023-03-08
Make ui-devel run-able outside docker-compose

TheRealHaoLiu:
##### SUMMARY
remove unnecessary bits from the make ui-devel target that to allow `make ui-devel` to be run-able outside of docker-compose

##### ISSUE TYPE
 - Bug, Docs Fix or other nominal change

##### COMPONENT NAME
 - Other

##### AWX VERSION

```
awx: 21.13.1.dev37+g3b39bdf1f0
```

##### ADDITIONAL INFORMATION
tests
- `make ui-devel` outside of container is successful
- `make docker-compose` without `make ui-devel` is successful and is able to access /api/v2
- `make docker-compose` after `make ui-devel` is successful and the UI come up 
- `make docker-compose` after `make ui-clean` is successful and able to access /api/v2

PR #13667, opened 2023-03-09
update Makefile to account for being inside or outside of a container

akus062381:
when running make ui-devel outside of the container, this error will occur:
image

This change will prevent this error from showing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -411,12 +411,17 @@ ui-release: $(UI_BUILD_FLAG_FILE)
 
 ui-devel: awx/ui/node_modules
 	@$(MAKE) -B $(UI_BUILD_FLAG_FILE)
-	mkdir -p /var/lib/awx/public/static/css
-	mkdir -p /var/lib/awx/public/static/js
-	mkdir -p /var/lib/awx/public/static/media
-	cp -r awx/ui/build/static/css/* /var/lib/awx/public/static/css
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -411,12 +411,17 @@ ui-release: $(UI_BUILD_FLAG_FILE)
...
 ui-devel: awx/ui/node_modules
 	@$(MAKE) -B $(UI_BUILD_FLAG_FILE)
-	mkdir -p /var/lib/awx/public/static/css
-	mkdir -p /var/lib/awx/public/static/js
-	mkdir -p /var/lib/awx/public/static/media
-	cp -r awx/ui/build/static/css/* /var/lib/awx/public/static/css
-	cp -r awx/ui/build/static/js/* /var/lib/awx/public/static/js
-	cp -r awx/ui/build/static/media/* /var/lib/awx/public/static/media
+	@if [ -d "/var/lib/awx" ] ; then \
+		command1 ; \
```
shanemcd: This doesn't seem right... Copy pasta?

shanemcd:
Looks like @TheRealHaoLiu was also trying to solve this here: https://github.com/ansible/awx/pull/13666

Makefile
akus062381: fixed :-D

TheRealHaoLiu (approved):

PR #13670, opened 2023-03-09
docker-compose wait for pg to be ready

fosterseth:
##### SUMMARY

I've been noticing this error when running make docker-compose from fresh environment (after deleting all containers and volumes)

```
tools_awx_1 | awx-manage migrate --noinput
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/db/backends/base/base.py", line 219, in ensure_connection
tools_awx_1 |     self.connect()
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/utils/asyncio.py", line 33, in inner
tools_awx_1 |     return func(*args, **kwargs)
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/db/backends/base/base.py", line 200, in connect
tools_awx_1 |     self.connection = self.get_new_connection(conn_params)
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/utils/asyncio.py", line 33, in inner
tools_awx_1 |     return func(*args, **kwargs)
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/db/backends/postgresql/base.py", line 187, in get_new_connection
tools_awx_1 |     connection = Database.connect(**conn_params)
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/psycopg2/__init__.py", line 122, in connect
tools_awx_1 |     conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
tools_awx_1 | psycopg2.OperationalError: could not connect to server: Connection refused
tools_awx_1 | 	Is the server running on host "postgres" (172.18.0.2) and accepting
tools_awx_1 | 	TCP/IP connections on port 5432?
```

usually Control + C and restarting helps

We can fix this by waiting on the postgres service to be ready to accept connections before starting the tools_awx_* containers

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/tools/docker-compose/bootstrap_development.sh
+++ b/tools/docker-compose/bootstrap_development.sh
@@ -14,6 +14,10 @@ make awx-link
 make version_file
 
 if [[ -n "$RUN_MIGRATIONS" ]]; then
+    # wait for postgres to be ready
+    while ! nc -z postgres 5432; do
+        echo "Waiting for postgres to be ready to accept connections"; sleep 1;
+    done;
     make migrate
 else
     wait-for-migrations


fosterseth:
@shanemcd  ~~Maybe my suggested alternative above would be better, if we don't want to use the `healthcheck` directive.~~

I went ahead and went with the alternative.
So instead of a container healthcheck and `depends_on`, just add this while loop to the bootstrap_development script.

tools/docker-compose/bootstrap_development.sh
```diff
@@ -14,6 +14,10 @@ make awx-link
 make version_file
 
 if [[ -n "$RUN_MIGRATIONS" ]]; then
+    # wait for postgres to be ready
+    while ! nc -z postgres 5432; do
```
relrod, line 18: Two thoughts, both probably not blocking:

1. Would it be best to have some kind of limit here? If it fails after 10-20 (or whatever) attempts, give up?
2. I did something very similar to this in another project, and ran into issues where on some ARM containers, using `nc` for this would segfault. I ended up using a small script in another language to do the test. This was with an old CentOS image though, so maybe not relevant anymore. Just something I thought of.

I will approve because I don't think either of these are worth blocking on.

relrod (approved):

tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
```diff
@@ -37,6 +37,7 @@ services:
     working_dir: "/awx_devel"
     volumes:
       - "../../../:/awx_devel"
+      - "../../docker-compose/bootstrap_development.sh:/usr/bin/bootstrap_development.sh"
```
AlanCoding: probably want to remove this, since it seems like a temporary hack to test before building images

fosterseth: haha I was trying to sneak it in. okay I'll remove it :)

tools/docker-compose/bootstrap_development.sh
fosterseth: thanks for taking a look. I'm content with the while loop for now, and if it causes issues later we can change to a for loop as you suggest. I'm trying not to bloat this file too much

gundalow:
> [#13641 (comment)](https://github.com/ansible/awx/pull/13641#issuecomment-1460804551) "for the django migration errors, just kill the containers and rerun make docker-compose. there is some kind of race condition the first time starting the containers (when migrations are needed)."
> 
> that threw me off when I ran the tests locally :)

@Klaas- did this solve it for you?

Klaas-:
will give it another try with a clean VM

Klaas-:
I'd guess yes, clean VM, first try, works fine :)

a-haurylau:
This check successfully block containers from start when using external DB. We should use 'postgres' in nc only for internal database. For external - pg_hostname from inventory file should be used.

AlanCoding:
That's a valid point, and I wonder if ultimately we might need to do this through an `awx-manage` command. In this context, we should be able to do that. I'm just a little worried about duplicating effort, because I'm guessing that OCP has multiple similar wait steps.

My initial thoughts are that maybe the `wait-for-migrations` script could be an `awx-manage` command itself, and then we can use a flag to wait for database availability or for migrations to complete, and so on...

PR #13671, opened 2023-03-09
Fix bug with manual type alias

AlanCoding:
##### SUMMARY
While `scm_type` local variable is changed to `''` when it gets "manual", it later on gets it from `module.params` which is obviously not aliased. To avoid, we alias earlier. Also, we can't say `scm_type` is not manual when it's not provided (`None`), so act accordingly.

##### ISSUE TYPE

##### COMPONENT NAME
 - Collection

### DIFF
--- a/awx_collection/plugins/modules/project.py
+++ b/awx_collection/plugins/modules/project.py
@@ -284,13 +284,15 @@ def main():
         argument_spec=argument_spec,
     )
 
+    # Alias for manual projects
+    if module.params.get('scm_type') == "manual":
+        module.params['scm_type'] = ''
+
     # Extract our parameters
     name = module.params.get('name')
     new_name = module.params.get("new_name")
     copy_from = module.params.get('copy_from')
     scm_type = module.params.get('scm_type')
-    if scm_type == "manual":
-        scm_type = ""
     local_path = module.params.get('local_path')
     credential = module.params.get('credential')
     scm_update_on_launch = module.params.get('scm_update_on_launch')
@@ -387,7 +389,8 @@ def main():
         # this is resolved earlier, so save an API call and don't do it again in the loop above
         project_fields['organization'] = org_id
 
-    if scm_type == '' and local_path is not None:
+    # Respect local_path if scm_type is manual type or not specified
+    if scm_type in ('', None) and local_path is not None:
         project_fields['local_path'] = local_path
 
     if scm_update_cache_timeout not in (0, None) and scm_update_on_launch is not True:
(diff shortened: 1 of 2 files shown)


awx_collection/plugins/modules/project.py
```diff
@@ -284,13 +284,15 @@ def main():
...
 
+    # Alias for manual projects
+    if module.params.get('scm_type') == "manual":
+        module.params['scm_type'] = ''
+
     # Extract our parameters
     name = module.params.get('name')
     new_name = module.params.get("new_name")
     copy_from = module.params.get('copy_from')
     scm_type = module.params.get('scm_type')
```
relrod, line 295: Might as well just pull this line up too and use it above?

awx_collection/plugins/modules/project.py
```diff
@@ -387,7 +389,8 @@ def main():
         # this is resolved earlier, so save an API call and don't do it again in the loop above
         project_fields['organization'] = org_id
 
-    if scm_type == '' and local_path is not None:
+    # Respect local_path if scm_type is manual type or not specified
+    if scm_type in ('', None) and local_path is not None:
```
relrod, line 393: I am sick and not thinking clearly, but I think this change makes sense. If it's the empty string, it was given by the user as "manual" per above. If it's `None`, it wasn't given by the user, and (I think) the API will default to "" (manual) if it's a new project.

It looks like in the serializer we already correctly error the case where the user could specify `local_path` for an SCM type where it doesn't make sense. Still, there are probably some test case scenarios it would be good to add here in the collection just to be explicit. Not :+1:-blocking though.

relrod (approved):

AlanCoding:
Let me explain the testing status:

It is very challenging to add integration tests for manual projects, because this relies on some corresponding setup with the folders on the server. It is difficult or impossible to automate this in an unopinionated way.

Because of that, I wrote a unit test where I have freedom to mock whatever as I see fit. This fails without the fix, and I feel like it's pretty good regression coverage for the fix.

awx_collection/plugins/modules/project.py
AlanCoding: > If it's `None`, it wasn't given by the user, and (I think) the API will default to "" (manual) if it's a new project.

This doesn't capture my own motivation for the `None` case.

If the value is `None` and there is an _existing_ project, then the existing project could be manual, and thus, allow specifying a local_path. What we really need to avoid is silently dropping the param when we don't know if it's valid or not. If they apply a local_path to a git project, then... don't do that, why would you do that? The API will throw a 400 error message which will have all the error details they need.

PR #13674, opened 2023-03-10
expose task settings in api

kdelee:
This will surface these settings that play key role in the task management system that are especially crucial for k8s deployements where relying on system detected capacity is generally dangerous because it will lead to control instances that appear to have the capacity of the entire underlying node.

We already did some magic to set these settings if resource limits are set...but I think its good to expose them and allow them to be set via API as 1) a way to verify it worked or 2) have an easy way to set them through API instead of by setting limits

Because in k8s all the replicas are the same, setting the capacity related settings in the API is safe. For other deployments where services are deployed directly on nodes, leaving these settings unset does not change behavior and capacity will still be detected from the filesystem.

No default behavior is changed by this PR, just allows the ability to view and set these settings from the API.

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### ADDITIONAL INFORMATION

some of these settings we have kept quite well hidden and not documented. This is an attempt to help make transparent that they exist and provide some minimal documentation via the help text.

image

### DIFF
--- a/awx/main/conf.py
+++ b/awx/main/conf.py
@@ -795,6 +795,106 @@ register(
     category_slug='bulk',
 )
 
+register(
+    'SYSTEM_TASK_ABS_CPU',
+    field_class=fields.CharField,
+    default='',
+    allow_blank=True,
+    label=_('Override CPU for capacity calculations for all instances.'),
+    help_text=_(
+        'Override CPU for capacity calculations for all instances.'
+        ' When not set, the heartbeat task will dynamically detect CPU available.'
+        ' Accepts formats equivalent to kubernetes resource requests like 4 or 4000m.'
+        ' For operator based deployments, this is managed by the operator based on resource limits.'
+    ),
+    category=_('Task'),
+    category_slug='task',
+)
+
+register(
+    'SYSTEM_TASK_ABS_MEM',
+    field_class=fields.CharField,
+    default='',
+    allow_blank=True,
+    label=_('Override Memory for capacity calculations for all instances.'),
+    help_text=_(
+        'Override Memory for capacity calculations for all instances.'
+        ' When not set, the heartbeat task will dynamically detect memory available.'
+        ' Accepts formats equivalent to kubernetes resource requests like 4Gi or 4000Mi.'
+        ' For operator based deployments, this is managed by the operator based on resource limits.'
+    ),
+    category=_('Task'),
+    category_slug='task',
+)
+
+register(
+    'SYSTEM_TASK_FORKS_MEM',
+    field_class=fields.IntegerField,
+    default=100,
+    label=_('Amount of memory needed per-fork.'),
+    help_text=_('Amount of memory needed per-fork, used when calculating memory based capacity of an instance.'),
+    category=_('Task'),
+    category_slug='task',
+)
+
+register(
+    'SYSTEM_TASK_FORKS_CPU',
+    field_class=fields.IntegerField,
+    default=4,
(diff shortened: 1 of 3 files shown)


AlanCoding:
Aren't at least _some_ of these redundant with the new fields you added?

https://github.com/ansible/awx/blob/5a3900a927bc9968c4de3837d92155ea1bf4ebce/awx/main/models/ha.py#L382-L383

There are just so many knobs that all seem like they're a coefficient in the same place in the same formula. If you're in the UI, the primary place for you to view and manage capacity use should probably be the instance and instance group list. The setting allows you to muck with the math with these, but there's no feedback.

If you add these, it would be good to have some back-reference from the instance / IG list to point here. At bare minimum we can mention it in the help text.

kdelee:
@AlanCoding in k8s deployments, I don't see them being redundant because those max_forks  and max_concurrent_jobs would only really apply to container groups -- which do not have any other sense of capacity.

For regular instance groups with VM instances in them, there is overlap/interaction of the settings and those global maximums -- but again there may be valid reasons why you want to cap an instance group at some global max different than the sum of the capacity of the instances.

Generally on VMs I think no one should touch these. This is mostly useful in k8s where our default behavior if no limits are set (and set none by default in awx-operator)  is really not reasonable and ends up with awx control Instances with outsized "capacity" for what they actually have access to.

awx/main/conf.py
```diff
@@ -795,6 +795,94 @@
...
+    category=_('Task'),
+    category_slug='task',
+)
+
+register(
+    'SYSTEM_TASK_ABS_MEM',
+    field_class=fields.CharField,
+    default='',
+    allow_blank=True,
+    label=_('Override Memory for capacity calculations for **ALL** instances. Accepts formats equivalent to kubernetes resource requests like 1Gi or 500Mi'),
```
AlanCoding: The label should be shorter than the help text.

awx/main/conf.py
```diff
@@ -795,6 +795,94 @@
...
+    help_text=_(
+        'Override Memory for capacity calculations for **ALL** instances. Accepts formats equivalent to kubernetes resource requests like 1Gi or 500Mi'
```
AlanCoding: The first sentence is hard to read.

It would also help _immensely_ to spell out the behavior, if the setting is _not_ set. That behavior would be that the heartbeat task will dynamically check to find out how much memory and CPU is available.

But it's more complicated for OCP, because the expectation is that the awx-operator (IIRC) will define a value in settings, which will take precedence over the `default=''` here (again, I _think_, needs confirmation). So in that case, not setting it doesn't necessarily result in a garbage value, but an appropriate value, and users should really think twice before changing it here.

kdelee: in ocp if its not set -- then it will read from the worker node the pod is scheduled on via the heartbeat. e.g. the way the operator communicates anything that is not garbage, is by setting this setting. So you would see it. This is one motivation I have for exposing it, so it would be clear that it is set. But maybe I can spell out that in ocp, this setting is controlled by operator if resource limits are set on the task container

kdelee:
@AlanCoding made some changes in response to your comments/suggestions, let me know what you think

mabashian:
Yea so we _could_ expose these fields in the UI but we definitely have a precedent for leaving some fields out of the UI, especially if those fields allow a user to shoot themselves in the foot.  If these fields fit that mould then I'd suggest we leave them out for now.  If users really clamor for those fields we can add them later.

kdelee:
Closing, we can re-open if interest is there

PR #13676, opened 2023-03-10
Fix organization not showing all galaxy credentials for org admin

gamuniz:
##### SUMMARY

Organizations were not properly showing all the associated galaxy credentials for Organization admins. They should be visible even if the credential cannot be accessed by an Org admin.

##### ISSUE TYPE

 
##### COMPONENT NAME

 - API
 

##### AWX VERSION

```
make VERSION
awx: 21.12.1.dev95+g122b9e23be.d20230310
```

##### ADDITIONAL INFORMATION

Steps to reproduce:
1. create a galaxy credential(in a separate organization as the org admin user)
2. create an org admin 
3. associate the galaxy credential with that to which the newly created user is an admin
4. Navigate to the Organization details page

Admin user:

image

Org admin:
image

Org Admin with fix:

image
 

```

```


awx/api/generics.py
```diff
@@ -528,6 +530,8 @@ def get_queryset(self):
         self.check_parent_access(parent)
         qs = self.request.user.get_queryset(self.model).distinct()
         sublist_qs = self.get_sublist_queryset(parent)
+        if not self.filter_read_permission:
+            return sublist_qs.prefetch_related(*self.prefetched_items)
```
AlanCoding: There's an alternative method of doing this above in this file, which can avoid the contract change of adding the `prefetched_items` property to the view

https://github.com/ansible/awx/blob/4f6c853c3ef4f0b87788b55c8cd58f4b5248a4fe/awx/api/generics.py#L367-L372

It would also be good to move the `qs` definition to after this `return` because in this code path it's never used.

gamuniz: I mostly understood your comments and am going to try something like this:

```
if not self.filter_read_permission:
        access_class = access_registry[self.model]
        if access_class.prefetch_related:
            return sublist_qs.prefetch_related(*access_class.prefetch_related)
        if access_class.select_related:
            return sublist_qs.select_related(*access_class.select_related)

       I reordered the returns as it seems that prefetch_related if available returns with less DB queries even if it is a little more processing on the api

**Review by AlanCoding — APPROVED** [review]
Many other view should be able to get rid of the `get_queryset` override by using `filter_read_permission = False`, which should result in a glorious amount of code deletions.

If you want to formalize that followup, go ahead. If you don't, I never intend to leave a good chunk of unnecessary code un-deleted.

**Comment by obaranov** [comment]
What is the final behavior? Admin org user should see all the credentials and instance groups? Can the org admin user delete the instance group to which they don't have access from the organization?

**Comment by gamuniz** [comment]
> What is the final behavior? Admin org user should see all the credentials and instance groups? Can the org admin user delete the instance group to which they don't have access from the organization?

@obaranov They should be able to see any credential/instance group associated to the org, but they should not be able to remove without the correct rbac permissions

**Comment by obaranov** [comment]
@gamuniz I checked, and api works fine. But with UI I have one small concern. User cannot edit org instance groups if at least one in the list has no user role.  

image


The current user does not have access to the `ig_restricted` group, but it does have an admin role for `ig_admin`. And when I try to remove the ig_admin group, I'm still getting the forbidden error on save: 

image

The same happens when I try to add a new instance group to the list.

File: awx/api/generics.py [hunk 2]
```diff
@@ -512,6 +512,9 @@ class SubListAPIView(ParentMixin, ListAPIView):
     # And optionally (user must have given access permission on parent object
     # to view sublist):
     #   parent_access = 'read'
+    # filter_read_permission sets whether or not to override the default intersection behavior
+    # implemented here
```
john-westcott-iv, line 514: This reads a little funny. I'd say drop `implemented here` or merge it with the line above.

awx/api/generics.py
```diff
@@ -526,8 +529,14 @@ def get_description_context(self):
...
         parent = self.get_parent_object()
         self.check_parent_access(parent)
-        qs = self.request.user.get_queryset(self.model).distinct()
         sublist_qs = self.get_sublist_queryset(parent)
+        if not self.filter_read_permission:
+            access_class = access_registry[self.model]
+            if access_class.prefetch_related:
+                return sublist_qs.prefetch_related(*access_class.prefetch_related)
+            if access_class.select_related:
+                return sublist_qs.select_related(*access_class.select_related)
```
john-westcott-iv, line 536: Do you want to be safe and raise an exception here that the access_class had neither prefetch_related nor select_related? Otherwise we turned off filter_read_permission but fall through to as if we hadn't.

awx/api/views/__init__.py
```diff
@@ -2716,6 +2707,7 @@ class JobTemplateInstanceGroupsList(SubListAttachDetachAPIView):
     serializer_class = serializers.InstanceGroupSerializer
     parent_model = models.JobTemplate
     relationship = 'instance_groups'
+    filter_read_permission = False
```
john-westcott-iv, line 2774: Are there any other items we would want to apply this to? i.e. Notifications, whatever else?

awx/api/generics.py
AlanCoding: This is doing the same thing as a prior bit in this file. The argument for its robustness is that the `BaseAccess` class sets these to empty tuples.

https://github.com/ansible/awx/blob/7deddabea60166ca8808509756a080d63cffa97e/awx/main/access.py#L223

awx/api/generics.py
gamuniz: fixing

awx/api/views/__init__.py
gamuniz: I would like to add it liberally to remove a lot of the queries that do identical lookups. That was part of the discussion we had last week and came to no conclusion.

AlanCoding: Yes, there's a performance argument for using this, if possible. However, we do have a lot of views that are _reverse_ relationships. Even if a JT should list all of its credentials, that doesn't mean a credential should list all the JTs it is used in.

awx/api/generics.py
john-westcott-iv: Talked this out with @AlanCoding I'm good with how this functions as is.

awx/api/views/__init__.py
gamuniz: @john-westcott-iv @AlanCoding if you would like me to add this to other places in the api just call it out, otherwise I'm okay with just the org endpoints and the job template credentials endpoint for now

john-westcott-iv (approved):

awx/api/views/__init__.py
john-westcott-iv: I'm good with this as a start. We can add other pieces as needed.

gamuniz:
@obaranov can we merge this guy?

jay-steurer (approved):

PR #13678, opened 2023-03-11
[constructed-inventory] Allow filtering based on facts

AlanCoding:
##### SUMMARY
The main diff here is that, since we are using facts for inventory update prep, they no longer "belong" to jobs, so I moved it into its own module. It is still awkward to manage the interface at times, but this is the most sense I can make of it.

This is mostly written in iteration with an integration test for it, unit tests need work as of opening this.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/tasks/facts.py
+++ b/awx/main/tasks/facts.py
@@ -1,0 +1,126 @@
+import codecs
+import datetime
+import os
+import json
+import logging
+
+# Django
+from django.conf import settings
+from django.db.models.query import QuerySet
+from django.utils.encoding import smart_str
+from django.utils.timezone import now
+
+# AWX
+from awx.main.utils.common import log_excess_runtime
+
+
+logger = logging.getLogger('awx.main.tasks.facts')
+system_tracking_logger = logging.getLogger('awx.analytics.system_tracking')
+
+
+def _get_inventory_hosts(inventory, slice_number, slice_count, only=('name', 'ansible_facts', 'ansible_facts_modified', 'modified', 'inventory_id'), **filters):
+    """Return value is an iterable for the relevant hosts for this job"""
+    if not inventory:
+        return []
+    host_queryset = inventory.hosts.only(*only)
+    if filters:
+        host_queryset = host_queryset.filter(**filters)
+    host_queryset = inventory.get_sliced_hosts(host_queryset, slice_number, slice_count)
+    if isinstance(host_queryset, QuerySet):
+        return host_queryset.iterator()
+    return host_queryset
+
+
+@log_excess_runtime(logger, debug_cutoff=0.01, msg='Inventory {inventory_id} host facts prepared for {written_ct} hosts, took {delta:.3f} s', add_log_data=True)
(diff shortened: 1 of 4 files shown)


CFSNM (approved):

awx/main/tasks/facts.py
```diff
@@ -0,0 +1,126 @@
...
+        return host_queryset.iterator()
+    return host_queryset
+
+
+@log_excess_runtime(logger, debug_cutoff=0.01, msg='Inventory {inventory_id} host facts prepared for {written_ct} hosts, took {delta:.3f} s', add_log_data=True)
+def start_fact_cache(inventory, destination, log_data, timeout=None, slice_number=0, slice_count=1):
+    log_data['inventory_id'] = inventory.id
+    log_data['written_ct'] = 0
+    try:
+        os.makedirs(destination, mode=0o700)
```
jjwatt, line 39: nit: Can we use `exist_ok=True` and skip the try/except? Was introduced in 3.2

awx/main/tasks/facts.py
```diff
@@ -0,0 +1,126 @@
...
+            else:
+                log_data['unmodified_ct'] += 1
+        else:
+            # if the file goes missing, ansible removed it (likely via clear_facts)
+            host.ansible_facts = {}
+            host.ansible_facts_modified = now()
+            hosts_to_update.append(host)
+            system_tracking_logger.info('Facts cleared for inventory {} host {}'.format(smart_str(host.inventory.name), smart_str(host.name)))
+            log_data['cleared_ct'] += 1
+        if len(hosts_to_update) > 100:
```
jjwatt, line 122: nit: being able to inject the pivot number could make unit/integration testing easier: e.g., set it to 1 or 2 for a test and don't have to have 100 hosts.

jjwatt (approved):

awx/main/tasks/facts.py
AlanCoding: The reason the except is needed here but not before is because we loop over `input_inventories`, so the dir will exist after the first one. I've thought that we could refactor this away instead.

AlanCoding: I'll keep thinking about this but will go ahead with the merge for now.

awx/main/tasks/jobs.py
```diff
@@ -1557,6 +1579,9 @@ def build_args(self, inventory_update, private_data_dir, passwords):
 
         return args
 
+    def should_use_fact_cache(self):
+        return bool(self.instance.source == 'constructed')
```
relrod, line 1583: Why the `bool()` cast here?

awx/main/tasks/facts.py
relrod: I think I agree that using `exist_ok=True` is a bit cleaner here. It effectively does the same thing, just avoids throwing the exception at all, so the try/except/pass lines become unnecessary.

relrod (approved):

PR #13682, opened 2023-03-14
Allow modifiable _sources (output) dir for builds

jjwatt:
##### SUMMARY

This dir, `_sources` is an output/target directory used by make rules and playbooks for storing generated files and for mounting some things into containers. Before this patchset, `_sources` is sometimes hard-coded, and it is at other times in a `sources_dest` var. This patchset makes it possible to set/change this sources dir. Some simplifications along the way.

###### Use
```shell
$ make SOURCES=out docker-compose
```

###### Why?
This patch cleans things up a little, and it may fix some hidden bugs, but the main force behind it is a weird incompatibility with podman:
Docker-compose will automatically name a network after the name of the directory it's started from, but `podman`, using docker-compose yml through the socket API, *but* podman rejects the network name because it starts with an `_`. Strangely, I couldn't find where that is specified in OCI or anything, but I did find the code in podman that tests against a regexp (iirc, it uses a `hostname` or dns name regexp), so this may be a spec miss/bug in podman. But, being able to change the name makes it work as a workaround, so I set things up to change the name so that I can continue use podman for testing on RHEL 8/9.

 There are still more patches needed to make podman work. I have it working in a branch, and I'm pulling those on top of HEAD one at a time and fixing them up for PRs.

##### ISSUE TYPE

##### COMPONENT NAME

- Other

##### AWX VERSION

```
awx: 0.1.dev32930+g3088eba.d20230314
```

##### ADDITIONAL INFORMATION

- Add override for '_sources' dir in Makefile SOURCES var
- Change docker-compose rules to use $(SOURCES)
- Add $SOURCES sh var to docker-compose container's sh scripts
- Add a SOURCES make variable that defaults to `_sources` to replace hard-coded `_sources`
- Add new SOURCES env var to docker-compose template
- Change plumb* ymls to use sources_dest or _sources
- Change pre_translation.sh to enable custom SOURCES
  - Use $SOURCES if it's in the env, otherwise original behavior (use '_sources' string for the dir)
- Change migrate-from-local-docker to use var sources_dest
- Change playbook to copy over all config files used by docker-compose
  - Add copy grafana dir rule
  - Copy `supervisor.conf` and `redis.conf` to roles/sources/templates dir
  - Add `superviosr.conf` and `redis.conf` to what's copied from the playbook
- Add awx_root to sources/defaults/main.yml
  - Discover awx_root by discovering the git root
  - Default to `../../../../` like the old behavior
- Change sources/tasks/main.yml to use awx_root in playbook
  - Introduce `awx_root` var because docker-compose yml kept trying to guess directory names using a scary anchored relative path, and it would fail if `_sources` moved
- Change docker-compose template
  - Remove use of relative path with anchor (`../docker-compose/../`)
- Change plumb playbooks to use sources_dest/default
- Change `$DIR/sources_dest` references to be just `sources_dest`

```

```


tools/docker-compose/bootstrap_development.sh
```diff
@@ -53,5 +53,5 @@ fi
 
 # Create resource entries when using Minikube
 if [[ -n "$MINIKUBE_CONTAINER_GROUP" ]]; then
-    awx-manage shell < /awx_devel/tools/docker-compose-minikube/_sources/bootstrap_minikube.py
+    awx-manage shell < /awx_devel/tools/docker-compose-minikube/"${SOURCES:-_sources}"/bootstrap_minikube.py
```
jjwatt: Should be "${SOURCES:-_sources}" but might need some work to get it there.

tools/scripts/l18n/pre_translation.sh
```diff
@@ -2,7 +2,7 @@
 
 # Extract Strings from API & UI
 make docker-compose-sources
-docker-compose -f tools/docker-compose/_sources/docker-compose.yml run awx_1 make awx-link migrate po messages
+docker-compose -f tools/docker-compose/"${SOURCES:-_sources}"/docker-compose.yml run awx_1 make awx-link migrate po messages
```
jjwatt: remove the prefix, `tools/docker-compose/`

tools/docker-compose/ansible/roles/sources/templates/docker-compose.yml.j2
```diff
@@ -36,19 +37,19 @@ services:
...
+      - "{{ awx_root }}:/awx_devel"
+      - "./supervisor.conf:/etc/supervisord.conf"
+      - "./database.py:/etc/tower/conf.d/database.py"
+      - "./websocket_secret.py:/etc/tower/conf.d/websocket_secret.py"
+      - "./local_settings.py:/etc/tower/conf.d/local_settings.py"
+      - "./nginx.conf:/etc/nginx/nginx.conf"
+      - "./nginx.locations.conf:/etc/nginx/conf.d/nginx.locations.conf"
+      - "./SECRET_KEY:/etc/tower/SECRET_KEY"
+      - "./receptor/receptor-awx-{{ loop.index }}.conf:/etc/receptor/receptor.conf"
+      - "./receptor/receptor-awx-{{ loop.index }}.conf.lock:/etc/receptor/receptor.conf.lock"
```
AlanCoding, line 49: I want to get a little more confidence in this change here.

https://docs.docker.com/compose/compose-file/build/

> Context path can be absolute or relative, and if so relative path MUST be resolved from Compose file parent folder.

I think this documenting support for this means of specifying relative paths. And really, it seems we already relied on this, and the docker-compose was already in the sources directory.

jjwatt: Yeah, we already relied on it...too much, imho. The old ones had the weird `docker-compose` anchor dir in them.

tools/docker-compose/ansible/roles/sources/templates/supervisor.conf.j2
```diff
@@ -0,0 +1,127 @@
...
+serverurl=unix:///var/run/supervisor/supervisor.sock ; use a unix:// URL  for a unix socket
+
+[rpcinterface:supervisor]
+supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface
+
+[eventlistener:stdout]
+command = supervisor_stdout
+buffer_size = 100
+events = PROCESS_LOG
+result_handler = supervisor_stdout:event_handler
```
AlanCoding: I like what you're trying to do here, but this concerns me. These new files seem to account for most of the line additions here. This indicates duplication of the files. If you were to move the existing file, I'm not so worried about that. But the supervisor conf, in particular, will absolutely change in the future we do refactoring of the systems, and duplicated files has in the past, and will in the future, cause drift between those files. We already have a hard time keeping the existing supervisor config files in sync. So at the moment, this is the one thing that causes me to raise my hand and say "don't merge yet"

jjwatt: Yeah. I didn't love doing this, but felt like doing what it did with the other files was the path of least resistance. I absolutely agree with you, though. That's one reason this is a draft, for sure. I'm open to suggestions, but off the cuff:
1. Just move the files (might break something; don't know what else uses them, etc)
2. I could copy the files over from the Makefile or the playbook (from within the project) and not worry about putting them in the role templates. Not sure if I'd even put them in the role at that point, though. Oh yeah, the other thing about doing it this way is that it's less flexible for moving around (like using the role somewhere else and expecting it to work the same way).

Open to suggestions or ideas.

jjwatt:
I'm seeing two failures in CI pre that I didn't see locally. I think I see how to reproduce them. I'm going to test and try to fix them.

tools/docker-compose/ansible/roles/sources/templates/supervisor.conf.j2
AlanCoding: I get where you're coming from. We have multiple supervisor files at these locations

 - `tools/ansible/roles/dockerfile/templates/supervisor.conf.j2` - for the web container, not obvious from the name
 - `tools/ansible/roles/dockerfile/templates/supervisor_task.conf.j2`
 - `tools/docker-compose/supervisor.conf`

The last one is only used by the docker-compose development environment and contains both web & task services, so you could call it a monolithic collection of services. The docker-compose stuff is categorically never used for production, and we can be much more incautious because of that. If you were to move one of the first 2 supervisor configs, that would trigger much more scrutiny, but you're not.

There is a incongruity where we directly map some files, but template others, as shown in the difference between these 2 items in the current docker-compose.

```
      - "../../docker-compose/supervisor.conf:/etc/supervisord.conf"
      - "../../docker-compose/_sources/database.py:/etc/tower/conf.d/database.py"
```

I completely support refactoring this so that both files work in generally the same way. The structure today is an accident of history.

My only nit pick about where to move would be that `templates/supervisor.conf.j2` may not follow _Ansible_ best practices very well, because we don't need to template anything in it (yet). The more standard Ansible role structure would put it in a files subdirectory like `files/supervisor.conf`, and these are generally expected to get copied over by a role.

Before you go too far into this, I must warn that https://github.com/ansible/awx/pull/13423 is a high-priority item that should land in the next few weeks and will have conflicts with your changes. All the more reason to make sure you keep the change to a single commit and do `git mv` for the moved files so that git has the best chance of seamlessly rebasing the change.

jjwatt: Excellent explanation. I'm happy to hear that the one `supervisor.conf` is likely moveable. I thought of another way to do it, but now I like aligning the methods so that they work generally the same way, like you said.

- I will move that `supervisor.conf` and double-check for any references.
- I'll also change it so that it's not a template. That was a cheap shot that I knew I'd want to change, and I didn't see the `files` module/practice until afterwards.
- I'll fix failures that CI caught

Thanks for the heads up about #13423 ! I took a look, and the conflicts don't look too bad. I wouldn't be opposed to rebasing on it/waiting for it to land before we land this one.

jjwatt:
OK. I was getting a stuttering `tools/docker-compose/ansible/tools/docker-compose/ansible/_sources`. The problem was that I set SOURCES in the `Makefile` and then pass SOURCES to `ansible-playbook` as if `ansbile-playbook` ran from the `$(pwd)`, but apparently its `$CWD` is the same as the playbook path, at least in this case.

I'll have to either rewrite that path when I pass it to the playbook or just set a (calculated) absolute path early on, at least for the file generation. I'd be more OK with the abspath generating if we were using make a little differently--like if we actually updated/checked generated files or had real targets for them. For now, I'll probably write a GNU Make macro to prep/distinguish the path (`$(SOURCES) var) when passing it around and avoid absolute paths.

Makefile
```diff
@@ -511,6 +512,7 @@ docker-compose-sources: .git/hooks/pre-commit
 	fi;
 
 	ansible-playbook -i tools/docker-compose/inventory tools/docker-compose/ansible/sources.yml \
+			-e sources_dest=$(abspath $(SOURCES)) \
```
AlanCoding: indentation kind of weird

tools/docker-compose/bootstrap_development.sh
AlanCoding: I'm not sure what this is doing, I don't see how it can be correct. Won't the sources directory be like `tools/docker-compose/_sources/`?

Makefile
jjwatt: Had to turn on whitespace. The one at the very bottom had all spaces. Looks like the rest of the file is a tab for command blocks and then spaces. I think I am used to (and my editor is used to) just doing all tabs for the command blocks. Think I fixed it.

tools/docker-compose/bootstrap_development.sh
jjwatt: You're right. I think I missed that one. I should test this more with the minikube stuff. :/ I think I know of a way to fix it for this PR that won't effect anything else. Let me try something.

jjwatt: PTAL. I think this will actually do the trick. Fixed up the `docker-compose-sources` rule in the `Makefile` to pass in `$(SOURCES)` to the minikube playbook. The minikube role didn't need any changes as it was using its `sources_dest` var throughout. I haven't tested it yet, but I'm about to. If it's not clear, idea is instead of having a separate `tools/docker-compose-minikube/_sources` I just do the right lazy thing and use `$(SOURCES)`, which is the same place as the regular flavored `docker-compose`, sans `-minikube`. I need to test in case they write the same files, but it doesn't look like they'll conflict at all. That will be great because it fits in better with make philosophy, too (one place for outputs).

AlanCoding (approved):
There's still kind of an open question about minkube and the SOURCES env var. I'm not fluent enough to really review, so I'm going to request some other reviews on this. Either that, or just sure it works, or back that out, because I do want the rest of these changes.

jjwatt:
I took care of the minikube standout by not trying to fix it :). Basically, since it's the only thing that needs to use `_sources` from inside the container, I leave the path hard-coded for it in the `bootstrap_development.sh` script, and then I make sure that the minikube playbook copies the *generated* `bootstrap_minikube.py` *from* our newly configurable `$SOURCES/sources_dest` to the old hard-coded location so that it'll be where it expects.

So this already starts to pay off: Before, the minikube playbooks were downloading huge minikube and k8s go binaries to the `_sources` directory inside the project directory, so they were always copying all of those over when you ran the container. Now that you can specify $SOURCES outside of the project dir (e.g., `/tmp/sources`), it automatically won't re-copy those files in every time you start a container! hooray

tools/scripts/l18n/pre_translation.sh
```diff
@@ -1,10 +1,9 @@
...
 
 # Extract Strings from API & UI
-make docker-compose-sources
-docker-compose -f tools/docker-compose/_sources/docker-compose.yml run awx_1 make awx-link migrate po messages
-
-# Move extracted Strings to Translation Directory
-mv awx/locale/en-us/LC_MESSAGES/django.po translations/
-mv awx/ui/src/locales/en/messages.po translations/
+make docker-compose-sources && \
+  docker-compose -f "${SOURCES:-_sources}"/docker-compose.yml run awx_1 make awx-link migrate po messages && \
```
jjwatt, line 5: Just noticed this. Leaving a comment for myself to fix it: `_sources` should be `tools/docker-compose/_sources` or wherever it ends up when run normally.

tools/docker-compose-minikube/minikube/tasks/main.yml
```diff
@@ -107,3 +107,9 @@
     src: bootstrap_minikube.py.j2
     dest: "{{ sources_dest }}/bootstrap_minikube.py"
     mode: '0600'
+
```
TheRealHaoLiu, line 110: we might want to leave some comment on why is this being done

tools/docker-compose/launch_awx.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 set +x
```
jjwatt, line 3: Leave a comment about shell "use or set default" shortcut.

jjwatt:
Letting this sit for a bit. Hao and I are pairing on some other `Makefile` changes, and Hao has an idea to simplify the `docker-compose` file/build even more. We're going to land those other changes and see if we still need this.

AlanCoding:
@jjwatt what you mentioned in your last comment sounds like https://github.com/ansible/awx/pull/14188

And I believe this was a well-considered patch that would have been helpful. I'm not sure if it's revivable with the current conflicts, but I would support that still.

PR #13686, opened 2023-03-14
[constructed-inventory] Remove towervars from constructed inventory hosts

AlanCoding:
##### SUMMARY
This addresses an aesthetic problem with constructed inventory, that all 4 of these variables would be present in all constructed hosts. The data doesn't actually offer any value to the user, but it is necessary as an intermediary step in order to fill in `instance_id` with the id of the original host that it came from.

This addresses the test case `test_metavars_not_in_constructed_hosts` which checks that the constructed inventory host has the same variables as the original host.

Just realized this needs to be done for groups too!

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API


AlanCoding:
I tested the same case for groups, and there are no extra vars we put in those, so it's not a thing anyway.

CFSNM (approved):

Issue #13669, opened 2023-03-09
The CyberArk Credential plugin names are not getting migrated after upgrade


obaranov:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### AWX version

devel

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

docker development environment

### Modifications

no

### Steps to reproduce

After upgrading the controller to the devel, the old CyberArk credentials plugin names are still displayed. 

1) Install old version of the AWX with the old names for the CyberArk plugins (prior to this change https://github.com/ansible/awx/issues/13122) 
2) Upgrade to the most recent version. 
3) Try to create new Credential 

### Expected results

Expected names are: 

![image](https://user-images.githubusercontent.com/15234398/224034166-e45bf065-449b-462e-98a2-a7c0d44b5ffa.png)

### Actual results

Old plugin names are still displayed

![image](https://user-images.githubusercontent.com/15234398/224033745-2353d045-2295-49e5-b58e-3ee671f259b2.png)

obaranov:
@infamousjoeg can you please take a look into this one? cc @ffirg

obaranov:
@AlanCoding do you have time to take a look into this one? I am not completely sure if this issue is related to awx or maybe to the aap installer.

AlanCoding:
Current hope is that running `CredentialType.setup_tower_managed_defaults()` just fixes this.

AlanCoding:
^ to update - that method is not sufficient to solve this, because it doesn't look for existing credential types based on `namespace` (which, I think, it should). So this needs work.

obaranov:
I added POC fix here: https://github.com/ansible/awx/pull/13692

Verified with the upgrade pipeline: 

TASK [ansible.automation_platform_installer.automationcontroller : Migrate the database schema (may take awhile when upgrading).] ***
[1;30mtask path: /tmp/setup/collections/ansible_collections/ansible/automation_platform_installer/roles/automationcontroller/tasks/tasks.yml:355[0m
[...]

PR #13692, opened 2023-03-15
Add migration for new cyberark plugin names

obaranov:
##### SUMMARY

Fixed issue with the migration of the new names: https://github.com/ansible/awx/issues/13669

 
##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.13.1.dev64+g268f884e92
```

##### ADDITIONAL INFORMATION

```

```


awx/main/migrations/0177_change_cyberark_plugin_names.py
```diff
@@ -0,0 +1,31 @@
...
+from awx.main.models import CredentialType
+
+
+class Migration(migrations.Migration):
+    def update_cyberark_ccp_name(apps, schema_editor):
+        CredentialType.setup_tower_managed_defaults(apps)
+        aim_types = apps.get_model('main', 'CredentialType').objects.filter(namespace='aim').order_by('id')
+
+        if aim_types.count() == 2:
+            original, renamed = aim_types.all()
```
AlanCoding: I don't think this ordering is guaranteed.

AlanCoding:
I think you're close here, with the addition of an `.order_by('created')` _maybe_. Needs testing.

I like that a lot better than having to reference specific names. In fact, this might be something that can be written generally, and applied to all managed types (instead of 2 operations).

Should add logging, in case a change is made, it is important to alert the user to it for auditing.

awx/main/migrations/0177_change_cyberark_plugin_names.py
obaranov: Fixed. Order now by 'changed'.

AlanCoding (approved):
This is a good near-term fix, but longer-term I would like to followup and adjust the logic inside of `CredentialType.setup_tower_managed_defaults` so that we won't need to do this again. Ideally, we could do this the next time that a credential type name is changed.

obaranov:
> I think you're close here, with the addition of an `.order_by('created')` _maybe_. Needs testing.
> 
> I like that a lot better than having to reference specific names. In fact, this might be something that can be written generally, and applied to all managed types (instead of 2 operations).
> 
> Should add logging, in case a change is made, it is important to alert the user to it for auditing.

I added logging and reduced everything to one operation.  I also checked that existing credentials were properly migrated after upgrade: 
image

Also added upgrade API tests here: https://github.com/ansible/tower-qa/pull/8021

PR #13694, opened 2023-03-15
Added limit and source_vars option to inventory module

CFSNM:
To be able to create a constructed inventory via AWX collection, we need to pass the limit and source_vars variables

### DIFF
--- a/awx_collection/plugins/modules/inventory.py
+++ b/awx_collection/plugins/modules/inventory.py
@@ -8,9 +8,7 @@ from __future__ import absolute_import, division, print_function
 
 __metaclass__ = type
 
-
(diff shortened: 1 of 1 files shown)


awx_collection/plugins/modules/inventory.py
```diff
@@ -175,6 +183,12 @@ def main():
     if variables is not None:
         inventory_fields['variables'] = json.dumps(variables)
 
+    # For constructed inventory, we need to pass limit and source_vars
+    if limit is not None and kind == 'constructed':
+        inventory_fields['limit'] = prevent_instance_group_fallback
```
relrod, line 188: copy-paste error?

CFSNM: yeah, will fix. BTW, not sure if this will be merged. May cause some conflicts

PR #13700, opened 2023-03-15
[constructed-inventory] Save facts on model for original host

AlanCoding:
##### SUMMARY
This change makes the "real" host the source of truth about Ansible facts when a job runs against constructed inventory.

These two diagrams were made to articulate the goal of this:

**Saving**

Screenshot from 2023-03-15 14-09-19

**Loading**

Screenshot from 2023-03-15 14-09-01

The idea is that we keep the original host (that which exists inside the input inventory) as the source of truth about Ansible facts. This is so that we can persist facts across jobs that use normal inventories and constructed inventories, both of which may touch the same host.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -30,7 +30,7 @@ from django.utils.safestring import mark_safe
 from django.utils.timezone import now
 from django.views.decorators.csrf import csrf_exempt
 from django.template.loader import render_to_string
-from django.http import HttpResponse
+from django.http import HttpResponse, HttpResponseRedirect
 from django.contrib.contenttypes.models import ContentType
 from django.utils.translation import gettext_lazy as _
 
@@ -1584,6 +1584,14 @@ class HostDetail(RelatedJobsPreventDeleteMixin, RetrieveUpdateDestroyAPIView):
 class HostAnsibleFactsDetail(RetrieveAPIView):
     model = models.Host
     serializer_class = serializers.AnsibleFactsSerializer
+
+    def get(self, request, *args, **kwargs):
+        obj = self.get_object()
+        if obj.inventory.kind == 'constructed':
+            # If this is a constructed inventory host, it is not the source of truth about facts
+            # redirect to the original input inventory host instead
+            return HttpResponseRedirect(reverse('api:host_ansible_facts_detail', kwargs={'pk': obj.instance_id}, request=self.request))
+        return super().get(request, *args, **kwargs)
 
 
 class InventoryHostsList(HostRelatedSearchMixin, SubListCreateAttachDetachAPIView):
--- a/awx/main/models/jobs.py
+++ b/awx/main/models/jobs.py
@@ -838,7 +838,7 @@ class Job(UnifiedJob, JobOptions, SurveyJobMixin, JobNotificationMixin, TaskMana
 
     def get_hosts_for_fact_cache(self):
         """
-        Builds the querset to use for writing or finalizing the fact cache
+        Builds the queryset to use for writing or finalizing the fact cache
         these need to be the 'real' hosts associated with the job.
         For constructed inventories, that means the original (input inventory) hosts
         when slicing, that means only returning hosts in that slice
--- a/awx/main/models/jobs.py
+++ b/awx/main/models/jobs.py
@@ -11,6 +11,7 @@ from urllib.parse import urljoin
 from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.db import models
+from django.db.models.functions import Cast
 
 # from django.core.cache import cache
 from django.utils.translation import gettext_lazy as _
@@ -21,6 +22,7 @@ from rest_framework.exceptions import ParseError
 
 # AWX
 from awx.api.versioning import reverse
+from awx.main.constants import HOST_FACTS_FIELDS
 from awx.main.models.base import (
     BaseModel,
     CreatedModifiedModel,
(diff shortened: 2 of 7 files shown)


awx/main/models/jobs.py
```diff
@@ -836,6 +838,27 @@ def get_notification_templates(self):
...
+
+        if self.inventory.kind == 'constructed':
+            id_field = Host._meta.get_field('id')
+            host_qs = Host.objects.filter(id__in=Host.objects.exclude(instance_id='').values_list(Cast('instance_id', output_field=id_field)))
+        else:
+            host_qs = self.inventory.hosts
+
+        host_qs = host_qs.only(*HOST_FACTS_FIELDS)
+        host_qs = self.inventory.get_sliced_hosts(host_qs, self.job_slice_number, self.job_slice_count)
+        return host_qs
```
AlanCoding, line 858: @relrod this is ultimately derived from the method you had introduced. Took me a while but it's making more sense to me. For a given job, there's a list of potential hosts that it may touch, so this is that. There's an exception case for constructed inventory as well as for job slicing.

AlanCoding:
I need to explain some of my last commit here - we have a problem where the _UI_ will only show facts from the obvious location. I wondered if maybe it would follow the related link from the host details, but it does not. The UI will get facts from `/api/v2/hosts/465/ansible_facts/` where 465 is the id for that job details page, even if the related link points to `/api/v2/hosts/464/ansible_facts/`.

So I thought I would get sneaky for a quick fix... I put in a redirect. This is what happens when you click on Ansible facts now:

Screenshot from 2023-03-15 15-51-26
Screenshot from 2023-03-15 15-50-58

With this, the details of the 465 host will _actually_ show the facts for the 464 facts, even in the UI. This seems acceptable to me, although maybe it's considered bad form. In any case I need to loop the UI developers into this @marshmalien @AlexSCorey 

Ultimately, I do also think the UI should give more user help in some form for the host details, to clarify that there is a host in another inventory that the current host is intended to reference. The API added an "original_host" related link for this purpose.

AlanCoding:
^ what I said in that last comment is largely invalid after you've rebuilt the UI correctly. The current UI doesn't show the facts tab for constructed inventory hosts (similar to smart inventories).

awx/main/models/jobs.py
```diff
@@ -836,6 +838,27 @@ def get_notification_templates(self):
     def get_notification_friendly_name(self):
         return "Job"
 
+    def get_hosts_for_fact_cache(self):
+        """
+        Builds the querset to use for writing or finalizing the fact cache
```
relrod:
```suggestion
        Builds the queryset to use for writing or finalizing the fact cache
```

awx/api/views/__init__.py
```diff
@@ -1585,6 +1585,14 @@ class HostAnsibleFactsDetail(RetrieveAPIView):
     model = models.Host
     serializer_class = serializers.AnsibleFactsSerializer
 
+    def get(self, request, *args, **kwargs):
+        obj = self.get_object()
+        if obj.inventory.kind == 'constructed':
+            # If this is a constructed inventory host, it is not the source of truth about facts
+            # redirect to the original input inventory host instead
+            return HttpResponseRedirect(reverse('api:host_ansible_facts_detail', kwargs={'pk': obj.instance_id}, request=self.request))
```
relrod, line 1593: I tested this with awxkit too, and it seems fine.

Of course (per the serializer change), the related URL automatically points to the right thing.

And even if someone forces awxkit to point directly to the constructed host's `/ansible_facts/` endpoint, it seems to handle the redirect fine.

awx/main/tasks/facts.py
```diff
@@ -42,15 +30,14 @@ def start_fact_cache(inventory, destination, log_data, timeout=None, slice_numbe
...
-        hosts = _get_inventory_hosts(inventory, slice_number, slice_count, ansible_facts_modified__gte=timeout)
-    else:
-        hosts = _get_inventory_hosts(inventory, slice_number, slice_count)
+
+    if isinstance(hosts, QuerySet):
+        hosts = hosts.iterator()
 
     last_filepath_written = None
     for host in hosts:
+        if host.ansible_facts_modified and (host.ansible_facts_modified >= now() - datetime.timedelta(seconds=timeout)):
```
relrod: Just to logic through this...

So if I understand right, facts should be expired if now() is greater than the modified timestamp + the timeout. This would mean the current time exceeds the stored timestamp plus the delta.

now() >= modified + timeout
now() - timeout >= modified

So I think the code has a sign error?

I think in the old code, the timeout was being used to determine *inclusion* of hosts (include every host that has *unexpired* facts), but here we're determining which hosts to *skip*/continue past, so we want to be true when they are expired.

But also it's 2am, and I might just not be mathing properly.

awx/main/tests/unit/models/test_jobs.py
```diff
@@ -49,26 +44,43 @@ def test_start_job_fact_cache(hosts, job, inventory, tmpdir):
...
+
+
+def test_start_job_fact_cache_within_timeout(hosts, tmpdir):
+    fact_cache = os.path.join(tmpdir, 'facts')
+    # the hosts fixture was modified 5s ago, which is less than 7s
+    last_modified = start_fact_cache(hosts, fact_cache, timeout=7)
+    assert last_modified
+
+    for host in hosts:
+        assert os.path.exists(os.path.join(fact_cache, host.name))
```
AlanCoding, line 78: @relrod great job finding the facts expiration problem. We did have an integration test that would have picked up a failure from that bug (eventually...) but no unit tests. So I added these 2 which get regression coverage for the issue you mentioned. I confirmed that they were both failing before the updates here, because the wrong sign makes both the expired / not-expired cases be wrong.

relrod (approved):
I think this looks reasonable now. Let's see what happens :)

awx/main/tests/unit/models/test_jobs.py
relrod: Thanks for the new tests!

PR #13704, opened 2023-03-15
Adding "password": "$encrypted$" to user serializer

john-westcott-iv:
##### SUMMARY
The awx.awx.user module was not respecting the update_secrets parameter even though it looked like it. This was due to the user object not having the "password" field in the serializer.

This change adds the password field if the user is an internal user.

##### ISSUE TYPE

##### COMPONENT NAME

 - API
 
##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```


AlanCoding (approved):
Make sure to mark this for a release note!

Klaas-:
~~I have a side question -- should there also be a possibility of "no password" -- for example a social user is not created with a password (in db the password is not a valid hash but populated with some random[?] string)~~ nevermind, I wanted to post it in the issue not on the PR :)

gamuniz (approved):

jay-steurer (approved):

Jaapis (approved):

PR #13713, opened 2023-03-16
Fix race with heartbeat and reaper logic

gamuniz:
##### SUMMARY

In the current implementation of the system's heartbeat, it is possible for a job launched during the heartbeat to be inadvertently reaped as it would not be in the excluded_uuid list presented to the reaper. This PR aims to add a ref_time to avoid newly created jobs from showing up on the list to reap.

##### ISSUE TYPE

##### COMPONENT NAME

 - API
 
##### AWX VERSION

```
make VERSION 
awx: 21.13.1.dev74+g6461ecc762

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/tests/functional/test_dispatch.py
+++ b/awx/main/tests/functional/test_dispatch.py
@@ -347,8 +347,8 @@ class TestJobReaper(object):
         'status, execution_node, controller_node, modified, fail',
         [
             ('running', '', '', None, False),  # running, not assigned to the instance
-            ('running', 'awx', '', None, True),  # running, has the instance as its execution_node
-            ('running', '', 'awx', None, True),  # running, has the instance as its controller_node
+            ('running', 'awx', '', minute, True),  # running, has the instance as its execution_node
+            ('running', '', 'awx', minute, True),  # running, has the instance as its controller_node
             ('waiting', '', '', None, False),  # waiting, not assigned to the instance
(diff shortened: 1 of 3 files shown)


awx/main/tests/functional/test_dispatch.py
```diff
@@ -412,6 +419,16 @@ def test_workflow_does_not_reap(self):
         i.save()
         j = WorkflowJob(status='running', execution_node='awx')
         j.save()
-        reaper.reap(i)
+        reaper.reap(i, ref_time=now)
 
         assert WorkflowJob.objects.first().status == 'running'
+
+    def test_should_not_reap_new(self):
```
AlanCoding, line 426: Considering how nuanced and deliberate the ordering of commands in this test are, and how it's tied back to a very highly specific issue seen in production, we could use a little bit of comments about the mechanics.

awx/main/tests/functional/test_dispatch.py
```diff
@@ -412,6 +419,16 @@ def test_workflow_does_not_reap(self):
         i.save()
         j = WorkflowJob(status='running', execution_node='awx')
         j.save()
-        reaper.reap(i)
+        reaper.reap(i, ref_time=now)
```
AlanCoding: I can't see how `ref_time` is needed for _this_ test.

AlanCoding (approved):

AlanCoding (approved):

PR #13717, opened 2023-03-17
[#13711] git depth set to 1

### LINKED ISSUE #13711: [Project_update] Git depth

### Feature type

New Feature

### Feature Summary

I am suggesting adding ``depth`` attribute with value 1 in task down there. 

https://github.com/ansible/awx/blob/ededc61a71f92f399438136ce70a8665869a246d/awx/playbooks/project_update.yml#L41-L50

Using depth attirbute will create a shallow clone with a history truncated. Cloning the project will be a lot faster and it won't hurt current setup, because we are deleting the project folder before and there is no need for walking down the history line in git repository. User only wants to checkout to a specific commit/branch and that is it.  Ansible documentation also states this: ``Needs git>=1.9.1 to work correctly.`` but if git version is lower than the requested one, Ansible will print a warning that depth attribute will not be used so I also won't break older instances. 

- [ ] UI
- [ ] API
- [ ] Docs
- [X] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

Add ``depth: 1`` in the task down there. 

https://github.com/ansible/awx/blob/ededc61a71f92f399438136ce70a8665869a246d/awx/playbooks/project_update.yml#L41-L50

### Current results

Project is being cloned with it's entire history. 

### Sugested feature result

Project is being cloned with truncated history. (It depends on a project size but in my case I got 4-5 speed up, so almost I doesn't even notice this job, so my main job is spending less time in pending state).

### Additional information

_No response_

dariodsa:
##### SUMMARY

Fixes #13711.

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - Collection

##### ADDITIONAL INFORMATION

Using depth attirbute will create a shallow clone with a history truncated. Cloning the project will be a lot faster and it won't hurt current setup, because we are deleting the project folder before and there is no need for walking down the history line in git repository. User only wants to checkout to a specific commit/branch and that is it. Ansible documentation also states this: Needs git>=1.9.1 to work correctly. but if git version is lower than the requested one, Ansible will print a warning that depth attribute will not be used so it also won't break older instances.

```

```

### DIFF
--- a/awx/playbooks/project_update.yml
+++ b/awx/playbooks/project_update.yml
@@ -47,6 +47,7 @@
             force: "{{ scm_clean }}"
             track_submodules: "{{ scm_track_submodules | default(omit) }}"
             accept_hostkey: "{{ scm_accept_hostkey | default(omit) }}"
+            depth: 1
           register: git_result
 
         - name: Set the git repository version


AlanCoding:
> because we are deleting the project folder before

The "we" in this sentence would only include those who set the project delete-on-update option to true. Other users are very much not doing this. While using a depth of 1 will be more preformant for an initial clone it won't necessarily be faster (or even compatible) as it is updated on a rolling basis.

This might be acceptable, conditionally, for the delete-on-update case. That's an interesting idea that only just now occurred to me.

AlanCoding:
Closing, as per comment, cannot accept if not under some further conditionals.

Issue #13716, opened 2023-03-17
[constructed-inventory] Host recent activity and event backlinks missing for original hosts


AlanCoding:

AlanCoding:
For the "old" smart inventory, you can see that we preserve links from jobs to hosts via these 2 types of relationships:

```python
        org = factories.organization()
        inventories = [factories.inventory(organization=org) for i in range(2)]
        special_string = fauxfactory.gen_alphanumeric()
        hosts = [
            factories.host(
                inventory=inventory, name=f'{special_string}-in-inventory={inventory.id}'
            ) for inventory in inventories
        ]
        smart_inv = factories.inventory(kind='smart', host_filter=f'name__icontains={special_string}', organization=org)
        assert smart_inv.get_related('hosts').count == 2  # sanity

        jt = factories.job_template(inventory=smart_inv, project=class_project)
        job = jt.launch()
        job.wait_until_completed().assert_successful()

        events_pg = job.get_related('job_events', host__isnull=False)
        assert events_pg.count >= 2
        event_hosts = set(e.host for e in events_pg.results)
        assert set(host.id for host in hosts) == event_hosts  # event --> host link

        constructed_hosts_pg = smart_inv.get_related('hosts')
        for host in constructed_hosts_pg.results:
            assert [job['id'] for job in host.summary_fields.recent_jobs] == [job.id]  # job host summary link
```

These relationships are:
 - events link directly back to the hosts they affect
 - job host summaries link resultant host status back to the host

In the case of constructed inventory, currently, both of these relationships point back to the _constructed_ inventory host, which is essentially a placeholder host. In order to have true audibility, we need to do what we already do with the "old" smart inventory shown in the above script - link back to the original hosts. This may be _in addition to_ the constructed hosts, but the presence of a link to the original hosts is needed to track changes made to the host.

AlanCoding:
Putting this here for posterity

```python
        org = factories.organization()
        inventory = factories.inventory(organization=org)
        special_string = fauxfactory.gen_alphanumeric()
        host = factories.host(inventory=inventory, name=f'{special_string}-inv={inventory.id}')
        smart_inv = factories.inventory(kind='smart', host_filter=f'name__icontains={special_string}', organization=org)
        assert smart_inv.get_related('hosts').count == 1  # sanity

        jt = factories.job_template(inventory=smart_inv, project=class_project, playbook='sleep.yml')
        job = jt.launch()
[...]

PR #13718, opened 2023-03-17
[constructed-inventory] Backlink events to real hosts and summaries to both hosts

AlanCoding:
##### SUMMARY
I don't love everything about how this is written, but the bones of it are likely to remain in place.

Connect https://github.com/ansible/awx/issues/13716, as this is a direct solution to that issue. Where I have a script that looks at backlinks for the smart_inventory analog, they both fail when applied to constructed inventories, and this resolves it for both cases. So the spec for what this is doing is in the test.

But I'll go over it here:
 - This _duplicates_ job host summaries so that we have 1 for the original host _and_ 1 for the constructed inventory host. This may still conflict with some other downstream logic at some point, but I'm not very worried about our ability to fix that if it does.
 - This _changes_ the event -> host link from the constructed inventory host to the original host. I don't see any other alternative.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/migrations/0179_constructed_inventory.py
+++ b/awx/main/migrations/0179_constructed_inventory.py
@@ -128,7 +128,7 @@ class Migration(migrations.Migration):
             field=models.ForeignKey(
                 default=None,
                 editable=False,
-                help_text='Only for jobs ran against constructed inventories, this links to the host inside the constructed inventory.',
+                help_text='Only for jobs run against constructed inventories, this links to the host inside the constructed inventory.',
                 null=True,
                 on_delete=django.db.models.deletion.SET_NULL,
                 related_name='constructed_host_summaries',
--- a/awx/main/migrations/0179_constructed_inventory.py
+++ b/awx/main/migrations/0179_constructed_inventory.py
@@ -122,4 +122,8 @@ class Migration(migrations.Migration):
                 help_text='This field is deprecated and will be removed in a future release. Regex where only matching hosts will be imported.',
             ),
         ),
+        migrations.AlterUniqueTogether(
+            name='jobhostsummary',
+            unique_together={('job', 'host')},
+        ),
     ]
--- a/awx/main/migrations/0179_constructed_inventory.py
+++ b/awx/main/migrations/0179_constructed_inventory.py
@@ -122,8 +122,17 @@ class Migration(migrations.Migration):
                 help_text='This field is deprecated and will be removed in a future release. Regex where only matching hosts will be imported.',
             ),
         ),
-        migrations.AlterUniqueTogether(
-            name='jobhostsummary',
-            unique_together={('job', 'host')},
+        migrations.AddField(
+            model_name='jobhostsummary',
+            name='constructed_host',
+            field=models.ForeignKey(
+                default=None,
+                editable=False,
+                help_text='Only for jobs ran against constructed inventories, this links to the host inside the constructed inventory.',
+                null=True,
+                on_delete=django.db.models.deletion.SET_NULL,
+                related_name='constructed_host_summaries',
+                to='main.host',
+            ),
         ),
     ]
--- a/awx/main/models/events.py
+++ b/awx/main/models/events.py
@@ -544,9 +544,7 @@ class JobEvent(BasePlaybookEvent):
                     'id', 'name'
                 )
                 constructed_host_map = self.host_map
-                host_map = {}
-                for host in all_hosts:
-                    host_map[host.name] = host.id
+                host_map = {host.name: host.id for host in all_hosts}
             else:
                 all_hosts = Host.objects.filter(pk__in=self.host_map.values()).only('id', 'name')
                 constructed_host_map = {}
@@ -558,10 +556,10 @@ class JobEvent(BasePlaybookEvent):
             updated_hosts_list = list()
             for host in hostnames:
                 updated_hosts_list.append(host.lower())
-                host_id = host_map.get(host, None)
+                host_id = host_map.get(host)
                 if host_id not in existing_host_ids:
                     host_id = None
-                constructed_host_id = constructed_host_map.get(host, None)
+                constructed_host_id = constructed_host_map.get(host)
                 host_stats = {}
                 for stat in ('changed', 'dark', 'failures', 'ignored', 'ok', 'processed', 'rescued', 'skipped'):
                     try:
--- a/awx/main/models/events.py
+++ b/awx/main/models/events.py
@@ -549,6 +549,7 @@ class JobEvent(BasePlaybookEvent):
                     host_map[host.name] = host.id
             else:
                 all_hosts = Host.objects.filter(pk__in=self.host_map.values()).only('id', 'name')
+                constructed_host_map = {}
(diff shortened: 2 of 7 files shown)


awx/main/tasks/callback.py
```diff
@@ -98,8 +99,12 @@ def event_handler(self, event_data):
             host = event_data.get('event_data', {}).get('host', '').strip()
             if host:
                 event_data['host_name'] = host
-                if host in self.host_map:
-                    event_data['host_id'] = self.host_map[host]
+                if hasattr(self, 'constructed_backlinks'):
+                    if host in self.constructed_backlinks:
+                        event_data['host_id'] = self.constructed_backlinks[host]
```
AlanCoding: I'm going to walk-back this change. Firstly, the naming of the new data structure is impossible to understand.

This changed the prior behavior from (event) -> (constructed host id) to (event) -> (real host id). There's some tortured logic, but I don't agree with this. Best to link events back to the constructed host.

The reason we _might want_ to link the events to the real host is so that you could traverse from the host back to its events. Problem is, I don't believe this is a thing. Doing that would cross multiple partitions (as we painfully discovered in other issues), and that creates problems. I don't believe the API ever follows this relationship, although I will have to check.

With that established, linking to the constructed host actually gives more information because it's trivial to walk back to the real host, but going from the real host to the constructed host is much harder.

awx/main/models/events.py
```diff
@@ -557,6 +559,26 @@ def _update_host_summary_from_stats(self, hostnames):
                 summary = JobHostSummary(created=now(), modified=now(), job_id=job.id, host_id=host_id, host_name=host, **host_stats)
                 summary.failed = bool(summary.dark or summary.failures)
                 summaries[(host_id, host)] = summary
+                if self.job.inventory.kind == 'constructed':
+                    # make an extra copy of summary for original host
```
relrod: As we discussed on our review call, we might look at adding an FK to the constructed host in the new row instead of having a copy of the row for original and constructed.

AlanCoding: I implemented this suggestion, the change looks quite different now.

awx/main/migrations/0179_constructed_inventory.py
```diff
@@ -122,4 +122,17 @@ class Migration(migrations.Migration):
                 help_text='This field is deprecated and will be removed in a future release. Regex where only matching hosts will be imported.',
             ),
         ),
+        migrations.AddField(
+            model_name='jobhostsummary',
+            name='constructed_host',
+            field=models.ForeignKey(
+                default=None,
+                editable=False,
+                help_text='Only for jobs ran against constructed inventories, this links to the host inside the constructed inventory.',
```
relrod:
```suggestion
                help_text='Only for jobs run against constructed inventories, this links to the host inside the constructed inventory.',
```

awx/main/models/events.py
```diff
@@ -538,23 +539,38 @@ def _update_host_summary_from_stats(self, hostnames):
...
 
-            all_hosts = Host.objects.filter(pk__in=self.host_map.values()).only('id', 'name')
+            if self.job.inventory.kind == 'constructed':
+                all_hosts = Host.objects.filter(id__in=self.job.inventory.hosts.values_list(Cast('instance_id', output_field=models.IntegerField()))).only(
+                    'id', 'name'
+                )
+                constructed_host_map = self.host_map
+                host_map = {}
+                for host in all_hosts:
+                    host_map[host.name] = host.id
```
relrod: Might as well use a comprehension here.

```suggestion
                host_map = {host.name: host.id for host in all_hosts}
```

awx/main/models/events.py
```diff
@@ -538,23 +539,38 @@ def _update_host_summary_from_stats(self, hostnames):
...
+                host_map = self.host_map
+
             existing_host_ids = set(h.id for h in all_hosts)
 
             summaries = dict()
             updated_hosts_list = list()
             for host in hostnames:
                 updated_hosts_list.append(host.lower())
-                host_id = self.host_map.get(host, None)
+                host_id = host_map.get(host, None)
```
relrod: nit-pick and not new, but `None` is the default, so:

```suggestion
                host_id = host_map.get(host)
```

awx/main/models/events.py
```diff
@@ -538,23 +539,38 @@ def _update_host_summary_from_stats(self, hostnames):
...
                 if host_id not in existing_host_ids:
                     host_id = None
+                constructed_host_id = constructed_host_map.get(host, None)
```
relrod: Same here

```suggestion
                constructed_host_id = constructed_host_map.get(host)
```

awx/main/models/jobs.py
```diff
@@ -1056,6 +1051,15 @@ class Meta:
...
     )
     host = models.ForeignKey('Host', related_name='job_host_summaries', null=True, default=None, on_delete=models.SET_NULL, editable=False)
+    constructed_host = models.ForeignKey(
+        'Host',
+        related_name='constructed_host_summaries',
+        null=True,
+        default=None,
+        on_delete=models.SET_NULL,
+        editable=False,
+        help_text='Only for jobs ran against constructed inventories, this links to the host inside the constructed inventory.',
```
relrod:
```suggestion
        help_text='Only for jobs run against constructed inventories, this links to the host inside the constructed inventory.',
```

awx/main/models/inventory.py
```diff
@@ -359,6 +359,8 @@ def get_script_data(self, hostvars=False, towervars=False, show_all=False, slice
             for host in hosts:
                 data['_meta']['hostvars'][host.name] = host.variables_dict
                 if towervars:
+                    if self.kind == 'constructed':
+                        data['_meta']['hostvars'][host.name]['remote_constructed_original_id'] = int(host.instance_id)
```
relrod: This key doesn't seem to be used anywhere right now that I can see. Are there other places we need to update to make use of this?

I see these right now:

```
[rick@fedora awx]$ rg '\bremote.+id\b'
awx/settings/defaults.py
721:CONTROLLER_INSTANCE_ID_VAR = 'remote_tower_id'
759:CONSTRUCTED_INSTANCE_ID_VAR = 'remote_tower_id'

awx/main/tasks/jobs.py
324:            self.runner_callback.host_map[hostname] = hv.get('remote_tower_id', '')

awx/main/tests/functional/models/test_inventory.py
22:            'remote_tower_id': host.id,
24:            'remote_host_id': host.id,

awx/main/management/commands/inventory_import.py
674:                    for var in ('remote_{}_enabled', 'remote_{}_id'):

awx/main/models/inventory.py
365:                            f'remote_{prefix}_id': host.id,
```

relrod (approved):
A few silly nits (nothing blocking) and a question.

But this approach makes sense to me. Approving. Feel free to pull in or ignore any of the nit-picks.

awx/main/models/inventory.py
AlanCoding: no it's not used anywhere, my intent is that this is for the user.

but I don't necessarily want to obligate us to test this so maybe I'll back it out.

PR #13721, opened 2023-03-20
Add missing filtering mechanism for the Thycotic Devops Vault credential lookup

sscheib-rh:
##### SUMMARY
Thycotic's DevOps Vault returns a JSON-encoded string with all sorts of metadata, etc., in it, which is not useful as a credential lookup plugin, as one would typically require a specific field of the secret (such as username or password).
This PR introduces a new field ('secret_field') which allows filtering the returned JSON encoded string on a specific field.
The implementation is in accordance with Thycotic's other product: Thycotic Secret Server, which provides this filtering mechanism as well. It is implemented as `tss.py`

##### ISSUE TYPE

##### COMPONENT NAME
 - UI
 - Other

##### AWX VERSION
```
awx: 0.1.dev32960+g4f6e7eb
```

### DIFF
--- a/awx/main/credential_plugins/dsv.py
+++ b/awx/main/credential_plugins/dsv.py
@@ -35,8 +35,14 @@ dsv_inputs = {
             'type': 'string',
             'help_text': _('The secret path e.g. /test/secret1'),
         },
+        {
+            'id': 'secret_field',
+            'label': _('Secret Field'),
+            'help_text': _('The field to extract from the secret'),
+            'type': 'string',
+        },
     ],
-    'required': ['tenant', 'client_id', 'client_secret', 'path'],
+    'required': ['tenant', 'client_id', 'client_secret', 'path', 'secret_field'],
 }
 
 if settings.DEBUG:
@@ -52,5 +58,5 @@ if settings.DEBUG:
 dsv_plugin = CredentialPlugin(
     'Thycotic DevOps Secrets Vault',
     dsv_inputs,
-    lambda **kwargs: SecretsVault(**{k: v for (k, v) in kwargs.items() if k in [field['id'] for field in dsv_inputs['fields']]}).get_secret(kwargs['path']),
+    lambda **kwargs: SecretsVault(**{k: v for (k, v) in kwargs.items() if k in [field['id'] for field in dsv_inputs['fields']]}).get_secret(kwargs['path'])['data'][kwargs['secret_field']],
 )
--- a/awx/main/credential_plugins/dsv.py
+++ b/awx/main/credential_plugins/dsv.py
@@ -58,5 +58,5 @@ if settings.DEBUG:
 dsv_plugin = CredentialPlugin(
     'Thycotic DevOps Secrets Vault',
     dsv_inputs,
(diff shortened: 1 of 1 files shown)


fosterseth:
you may need to address the api lint error, you can run `make black` locally to see lint errors

sscheib-rh:
Thanks for coming back to me @fosterseth. I am unfamiliar with "black", so I don't know what it's supposed to do.
I ran `make black` locally, and it changed that one file (`dsv.py`) and formatted it differently:
```
[steffen@lab-development-rhel8 awx]$ make black
reformatted awx/main/credential_plugins/dsv.py

All done! ✨ 🍰 ✨
1 file reformatted, 820 files left unchanged.
```
The result looks like this:

```
[steffen@lab-development-rhel8 awx]$ git diff 
diff --git a/awx/main/credential_plugins/dsv.py b/awx/main/credential_plugins/dsv.py
index 78d5e6a4f0..d4ae8e1a4a 100644
--- a/awx/main/credential_plugins/dsv.py
+++ b/awx/main/credential_plugins/dsv.py
@@ -58,5 +58,7 @@ if settings.DEBUG:
 dsv_plugin = CredentialPlugin(
     'Thycotic DevOps Secrets Vault',
     dsv_inputs,
-    lambda **kwargs: SecretsVault(**{k: v for (k, v) in kwargs.items() if k in [field['id'] for field in dsv_inputs['fields']]}).get_secret(kwargs['path'])['data'][kwargs['secret_field']],
+    lambda **kwargs: SecretsVault(**{k: v for (k, v) in kwargs.items() if k in [field['id'] for field in dsv_inputs['fields']]}).get_secret(kwargs['path'])[
+        'data'
+    ][kwargs['secret_field']],
 )
```

If I am not mistaken, lambdas cannot be a multiline expression (but please correct me if I am wrong), so is there a way to tell 'black' to ignore that line?
Further, I don't see any 'API-linting errors', but then again, I don't know what it's supposed to look like. Maybe the error is already that black is reformatting that file.

So my two questions are:
- Can I instruct black to ignore this particular line?
- If not, is there a better way to write that lambda which I am unaware of? I am not a full-time developer, so this might be obvious for a skilled developer.

fosterseth:
@sscheib-rh 

I would try to add `# fmt: skip ` at the end of the line https://github.com/psf/black/issues/451#issuecomment-1054372101

alternatively, maybe use an in-line python function instead of lambda, which should allow you to break up the long line

sscheib-rh:
@fosterseth 
Added `# fmt: skip` - this led to `api-lint` completing successfully, however, now the `awx-operator` fails, and I am (again) failing to see why.
Sorry to bother you with that, but I am unsure how to continue; Any hint would be appreciated.
My first guess would be to simply try re-running the `awx-operator` build, but it seems I don't have the necessary permissions to do so.

fosterseth:
yeah we have a known issue with the operator ci build, once that is resolved I'll re-run the check

thedoubl3j:
@delineaKrehl can you take a look at this PR?

delineaKrehl (approved):

TheRealHaoLiu (approved):

Issue #13169, opened 2022-11-08
AWX Collections / credentials  - do not override existing value when updating just one value in the inputs field


dnguyen9:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

Enhancement to Existing Feature

### Feature Summary

the awx.awx collections is used to extensively for self-service onboarding and we create the stubs objects ahead of time, and we let our users fill in the secrets inside the AWX GUI. 

For the credential object, we would create vault secrets with a vault_id and our users would go into the AWX GUI and put in their vault password. Once we run the automation again, the user inputs would be wiped out. Would like it the collections to not delete the fields that are not specified in the  'inputs' and let it alone.

`- name: Onboarding | create stub credentials for team
   credential:  
   type: 'Vault'  
   name:  Vault_team
   inputs:
    vault_id: vault_for_database
`

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [X] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

1, Use  awx collections and specify the vault_id via input

`- name: Onboarding | create stub credentials for team
   credential:
   type: 'Vault'
   name:  Vault_team
   inputs:
    vault_id: vault_for_database
`
2.  update the password for the Vault_team cred object
3. re-run the awx collections in step #1. the password will be wiped out

### Current results

password fields is wiped out

### Sugested feature result

password fields should be left alone

john-westcott-iv:
Have you tried running this with `update_secrets` option in the module?

```
- update_secrets
        `true' will always update encrypted values.
        `false' will only updated encrypted values if a change is absolutely known to be needed.
        [Default: True]
        type: bool
```

This should alter the behavior of the module to only attempt to update the password field if something else in the credential has changed.

dnguyen9:
I tried that against collection 21 on AWX 15 but  was not successful. I'll try again and report back with more details.

dnguyen9:
@john-westcott-iv 

I did some digging and here are my finding
update_secrets param does not work in the following conditions:
[...]

Issue #13687, opened 2023-03-14
AWX Community Meeting Agenda - Apr 11 2023


john-westcott-iv:
# AWX Office Hours

## Proposed agenda based on topics
* Web Task Split @thedoubl3j   5 min
* Constructed Inventory @AlanCoding 10 min
* UI Next @TheRealHaoLiu 5 min
* Infra Config as Code @sean-m-sullivan 20min 
* https://github.com/ansible/awx/pull/13507 @tanganellilore @CMPluto @AlanCoding 15min
* Thanks to new contributors & closing 5min

## What

After a successful Contributor Summit in October, one of the bits of feedback we got was to host a regular time for the Automation Controller (AWX) Team to be available for your folks in the AWX Community, so we are happy to announce a new regular video meeting.

This kind of feedback loop is vital to the success of AWX and the AWX team wants to make it as easy as possible for you - our community - to get involved.

## Where & When
Our next meeting will be held on Tuesday, Apr 11th, 2023 at [1500 UTC](https://dateful.com/time-zone-converter?t=15:00&tz=UTC)

* [Google Meet](https://meet.google.com/vyk-dfow-cfi)
* Via Phone PIN: 842522378 [Guide](https://support.google.com/meet/answer/9518557)

This meeting is held once a month, on the second Tuesday of the month, at [1500 UTC](https://dateful.com/time-zone-converter?t=15:00&tz=UTC)

## How

Add one topic per comment in this GitHub issue

If you don't have a GitHub account, jump on [#awx:ansible.com](https://matrix.to/#/#awx:ansible.com) on Matrix and we can add the topic for you

## Talk with us

As well as the fortnightly video meeting you can join the Community (inc development team) on Matrix Chat.

* Matrix: [#awx:ansible.com](https://matrix.to/#/#awx:ansible.com) (recomended)
* libera.chat IRC: `#ansible-awx` (If you are already setup on IRC)

The Matrix & IRC channels are bridged, you'll just have a better experience on Matrix

## Links
[AWX YouTube Chanel](https://www.youtube.com/@ansible-awx)
[Previous Meeting](https://github.com/ansible/awx/issues/13563)
[Meeting recording](https://youtu.be/a75Kv8CN_Lg)
[Next Meeting](https://github.com/ansible/awx/issues/13836)

See you soon!

sean-m-sullivan:
The [Infra Configuration as Code Team](https://galaxy.ansible.com/infra/) will do a presentation of how to use their collections.

[...]

PR #13725, opened 2023-03-20
[collection] Add "exists" state for credential module

l3acon:
##### SUMMARY

Add the enumeration `exists` to `state` for the credential module. This is a relatively simple change that allows the module to be used in the following situation: 
  1.  we want to ensure a credential resource of a particular name is available, but
  2. if the resource is already configured we do not want to modify any inputs (including secrets)

Since credentials have protections in place to safeguard their privacy, overwriting them can be (at best) annoying or (at worst) dangerous. This is an attempt to address https://github.com/ansible/awx/issues/13169.

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - Collection


john-westcott-iv:
In the collection there is an `controller_api` lookup that you could leverage to see if a credential exists. So this could be implemented like so:
```
    - name: Create credential if needed
      awx.awx.credential:
        credential_type: Machine
        inputs:
          username: John
          password: 'test'
        name: Test Machine Credential
        organization: Default
      when: "lookup('awx.awx.controller_api', 'credentials', query_params={ 'name': 'Test Machine Credential' }) | length == 0"
```

That being said, I'm not opposed to an `exists` option in the state but we would probably want to look at the other modules to see if we should add exists in other areas as well for consistency.

john-westcott-iv:
Perhaps we could discuss this at the next community meeting? If that would be something you are interested in could you leave a comment on https://github.com/ansible/awx/issues/13687?

willtome:
@john-westcott-iv I was unaware of the api lookup. That's pretty cool. I'm not sure how that would integrate to work with the controller_collection though. 

@sean-m-sullivan do you think a lookup could work with the way the roles are today or would an additional parameter be better?

sean-m-sullivan:
@willtome like this?
https://github.com/redhat-cop/controller_configuration/blob/devel/roles/object_diff/tasks/credentials.yml
However this will not remove the item from being updated if it already exists, its more for cleaning up controllers.

sean-m-sullivan:
I think this state for credentials is  worth it, as you would stop changing credentials in a simple way, most of the other modules its just putting in settings, this prevents a password change unless it specifically is needed, but John is right like with all things ansible, there are 20 ways of doing it.

sean-m-sullivan (approved):

john-westcott-iv:
@l3acon Can you add some testing around this PR. Specifically in https://github.com/ansible/awx/blob/devel/awx_collection/tests/integration/targets/credential/tasks/main.yml and in https://github.com/ansible/awx/blob/devel/awx_collection/test/awx/test_credential.py.

l3acon:
@john-westcott-iv  thanks for pointing those out. I wasn't sure what the unit test should look like since the integration tests seem to cover the use-cases I have. Totally open to suggestions or ideas.

awx_collection/test/awx/test_credential.py
```diff
@@ -132,3 +132,21 @@ def test_secret_field_write_twice(run_module, organization, admin_user, cred_typ
     else:
         assert result.get('changed') is False, result
         assert Credential.objects.get(id=result['id']).get_input('token') == val1
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('state', ('present', 'absent', 'exists'))
+def test_credential_state(run_module, organization, admin_user, cred_type, state):
+    for state in ('present', 'absent', 'exists'):
```
fosterseth: this `for loop` is redundant if you already parametrize it above
I'd keep the parametrize

fosterseth (approved):

thedoubl3j (approved):

thedoubl3j:
kicking CI

john-westcott-iv:
I was thinking about this PR this morning and I am wondering if you were hitting the issue in https://github.com/ansible/awx/pull/13704. tldr; the password was always being changed even when it should have been skipped because the user api page did not contain `"password": "$encrypted$"` which threw off the module_util logic.

Issue #13702, opened 2023-03-15
[constructed-inventory] Can't use constructed inventory with collection


AlanCoding:
We have so far not introduced a "constructed_inventory" module to the AWX collection.

Presumably, you might be able to do the same thing in a 2-step action with the "inventory" and then "inventory_source" modules. However, right now our API validation logic is too strict to allow this, and will give you a 400 error for any modifications to the inventory source.

Deliverable - some playbook that will create and use a constructed inventory.

PR #13727, opened 2023-03-20
[constructed-inventory] Fix some validation for constructed inv sources

relrod:
##### SUMMARY

- When updating, we need the original object so we can make sure we aren't changing things we shouldn't be.
- We want to allow source_vars and limit, but not much else.
- We want to block everything else (at least, if it doesn't match what is in the original object...to allow the collection to work properly).
- Add two functional tests.

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -2353,14 +2353,18 @@ class InventorySourceSerializer(UnifiedJobTemplateSerializer, InventorySourceOpt
 
     # TODO: remove when old 'credential' fields are removed
     def update(self, obj, validated_data):
+        original = self.context['original']
         deprecated_fields = {}
         if 'credential' in validated_data:
             deprecated_fields['credential'] = validated_data.pop('credential')
         obj = super(InventorySourceSerializer, self).update(obj, validated_data)
         if deprecated_fields:
             self._update_deprecated_fields(deprecated_fields, obj)
-        if obj.source == 'constructed':
-            raise serializers.ValidationError({'error': _("Cannot edit source of type constructed.")})
+        if original.source == 'constructed':
+            allowed_fields = ('source_vars', 'limit')
+            for field in validated_data:
+                if validated_data[field] != getattr(original, field) and field not in allowed_fields:
+                    raise serializers.ValidationError({"error": _("Cannot change field '{}' on a constructed inventory source.").format(field)})
         return obj
 
     # TODO: remove when old 'credential' fields are removed
@@ -2388,7 +2392,7 @@ class InventorySourceSerializer(UnifiedJobTemplateSerializer, InventorySourceOpt
             if ('source' in attrs or 'source_project' in attrs) and get_field_from_model_or_attrs('source_project') is None:
                 raise serializers.ValidationError({"source_project": _("Project required for scm type sources.")})
         elif (get_field_from_model_or_attrs('source') == 'constructed') and (self.instance and self.instance.source != 'constructed'):
-            raise serializers.ValidationError({"Error": _('constructed not a valid source for inventory')})
+            raise serializers.ValidationError({"error": _('constructed not a valid source for inventory')})
         else:
             redundant_scm_fields = list(filter(lambda x: attrs.get(x, None), ['source_project', 'source_path', 'scm_branch']))
             if redundant_scm_fields:
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -45,7 +45,7 @@ from polymorphic.models import PolymorphicModel
 
 # AWX
 from awx.main.access import get_user_capabilities
-from awx.main.constants import ACTIVE_STATES, CENSOR_VALUE, CONSTRUCTED_INVENTORY_SOURCE_EDITABLE_FIELDS
+from awx.main.constants import ACTIVE_STATES, CENSOR_VALUE
 from awx.main.models import (
     ActivityStream,
     AdHocCommand,
@@ -187,6 +187,11 @@ SUMMARIZABLE_FK_FIELDS = {
     'approved_or_denied_by': ('id', 'username', 'first_name', 'last_name'),
     'credential_type': DEFAULT_SUMMARY_FIELDS,
 }
+
+
+# These fields can be edited on a constructed inventory's generated source (possibly by using the constructed
+# inventory's special API endpoint, but also by using the inventory sources endpoint).
+CONSTRUCTED_INVENTORY_SOURCE_EDITABLE_FIELDS = ('source_vars', 'update_cache_timeout', 'limit', 'verbosity')
 
 
 def reverse_gfk(content_object, request):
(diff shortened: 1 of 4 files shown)


awx/api/serializers.py
```diff
@@ -2353,14 +2353,18 @@ def create(self, validated_data):
...
         deprecated_fields = {}
         if 'credential' in validated_data:
             deprecated_fields['credential'] = validated_data.pop('credential')
         obj = super(InventorySourceSerializer, self).update(obj, validated_data)
         if deprecated_fields:
             self._update_deprecated_fields(deprecated_fields, obj)
-        if obj.source == 'constructed':
-            raise serializers.ValidationError({'error': _("Cannot edit source of type constructed.")})
+        if original.source == 'constructed':
+            allowed_fields = ('source_vars', 'limit')
```
relrod: Are there other fields we should allow here?

awx/api/views/__init__.py
```diff
@@ -2004,6 +2004,11 @@ class InventorySourceDetail(RelatedJobsPreventDeleteMixin, RetrieveUpdateDestroy
     model = models.InventorySource
     serializer_class = serializers.InventorySourceSerializer
 
+    def get_serializer_context(self):
+        context = super().get_serializer_context()
+        context.update(original=self.get_object())
```
relrod: I don't know if this is the best way to do this. I needed the original object, not the object as represented by the request data. I didn't see a better way to get the object in the serializer without stuffing it in the context. This might be horrible, I'm not sure.

awx/main/tests/functional/api/test_inventory.py
```diff
@@ -624,6 +624,35 @@ def test_patch_constructed_inventory(self, constructed_inventory, admin_user, pa
...
+                'source_path': '',
+                'source_vars': 'plugin: a.b.c',
+            },
+            expect=400,
+            user=admin_user,
+        )
+        assert r.data['error'] == "Cannot change field 'source' on a constructed inventory source."
+
+        # Make sure it didn't get updated before we got the error
+        assert inv_src == constructed_inventory.inventory_sources.first()
```
relrod: I think this won't do what I want, this probably just compares PK. I should compare the fields themselves.

relrod: If I do this after the `patch()` it fails, and I have no idea why, especially because if I test _outside_ of pytest (using the collection, getting the 400, and looking at the database), the record doesn't get updated.

```python
inv_src_after_err = constructed_inventory.inventory_sources.first()
assert inv_src.source == inv_src_after_err.source
```

awx/api/serializers.py
AlanCoding: seems like you're looking for the same list as the ConstructedInventorySerializer, which are

'source_vars', 'update_cache_timeout', 'limit', 'verbosity'

awx/api/views/__init__.py
AlanCoding: I think a better place would be in the serializer `validate` method (or `update` like now) before the `super()` call. The criteria for checking for a change goes get a little messy. For every field in `attrs`, if it's not in the allowed list for changing (given it's a constructed source type) then if `attrs` value differs from the object, then you want to throw a validation message.

relrod: This is effectively what I do in the serializer. But I needed access to the original object as it exists in the database when we reach the serializer.

The `obj` that gets passed to `update()` seems to already have the request data in it. So I can't check if it's different, because it never will be.

So here I'm sending along the original object in the serializer context so that I have something to compare the new object (in `obj`) to.

awx/api/serializers.py
relrod: I pulled this out to a constant and shared the list between the two.

awx/main/tests/functional/api/test_inventory.py
```diff
@@ -624,6 +624,40 @@ def test_patch_constructed_inventory(self, constructed_inventory, admin_user, pa
...
+            },
+            expect=400,
+            user=admin_user,
+        )
+        assert r.data['error'] == "Cannot change field 'source' on a constructed inventory source."
+
+        # Make sure it didn't get updated before we got the error
+        inv_src_after_err = constructed_inventory.inventory_sources.first()
+        assert inv_src.id == inv_src_after_err.id
+        assert inv_src.source == inv_src_after_err.source
```
relrod, line 645: This fails and I don't get why, yet.

awx/api/views/__init__.py
relrod: I moved this into `validate()` per @gamuniz's suggestion. `self.instance` is the original object, still, at this point (but not once we hit `validate()`), so this avoids having to stuff it into the context.

awx/main/tests/functional/api/test_inventory.py
relrod: This got resolved after moving the logic into `validate()`.

awx/main/constants.py
```diff
@@ -106,3 +106,7 @@
 ANSIBLE_RUNNER_NEEDS_UPDATE_MESSAGE = (
     '\u001b[31m \u001b[1m This can be caused if the version of ansible-runner in your execution environment is out of date.\u001b[0m'
 )
+
+# These fields can be edited on a constructed inventory's generated source (possibly by using the constructed
+# inventory's special API endpoint, but also by using the inventory sources endpoint).
+CONSTRUCTED_INVENTORY_SOURCE_EDITABLE_FIELDS = ('source_vars', 'update_cache_timeout', 'limit', 'verbosity')
```
AlanCoding: As a style / preference nit-pick, I feel like this should rightfully belong in the serializers module until some other module has a use for it.

awx/api/serializers.py
```diff
@@ -1787,7 +1787,7 @@ class Meta:
 
     def pop_inv_src_data(self, data):
         inv_src_data = {}
-        for field in ('source_vars', 'update_cache_timeout', 'limit', 'verbosity'):
+        for field in CONSTRUCTED_INVENTORY_SOURCE_EDITABLE_FIELDS:
```
AlanCoding, line 1795: If you're doing this, I'd request a corresponding change to this other line:

```diff
diff --git a/awx/api/serializers.py b/awx/api/serializers.py
index 53c5bbf882..6fc22072d1 100644
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -1782,7 +1785,7 @@ class ConstructedInventorySerializer(InventorySerializer):
 
     class Meta:
         model = Inventory
-        fields = ('*', '-host_filter', 'source_vars', 'update_cache_timeout', 'limit', 'verbosity')
+        fields = ('*', '-host_filter') + CONSTRUCTED_INVENTORY_SOURCE_EDITABLE_FIELDS
         read_only_fields = ('*', 'kind')
 
     def pop_inv_src_data(self, data):
```

awx/main/constants.py
relrod: Fair, I'll move it.

AlanCoding (approved):
Great work, thanks!

relrod:
Refs #13702 

But I think also maybe fixes #13732?

Issue #13714, opened 2023-03-17
[constructed inventory] edit constructed inventory variables in UI?


relrod:
Currently the UI doesn't seem to let me edit inventory variables on constructed inventories. But the API will gladly accept `variables` as a parameter.

Questions:

- Does it make sense to have constructed-inventory-global inventory variables?
- If so, does the UI need to add the field for them so they can be set?

cc @AlanCoding

AlanCoding:
The API should prevent editing of inventory-level variables, like it does for hosts and groups.

But the exact meaning of the inventory variables are the "all" group variables to Ansible. These should come via the inventory imports, so their true status would be that they should be read-only.

AlanCoding:
I realize the title of this issue is unhelpful, we need to rename it. Things that need to be done here:

 - [ ] ~Add validation in the API to prevent editing constructed inventory `variables`~ --> moving into new issue
 - [ ] ~show constructed inventory variables in the UI as read-only~ --> moving into new issue
 - [x] overwrite inventory `variables` in imports, opposed to merging them, which is what is does now, and is very wrong

Issue #11623, opened 2022-01-27
Variable stays in awx inventory even it is deleted in git repo.


kzinas-adv:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Summary

Using sourced from project inventory update group_vars/all.yaml variables are not deleted
After removing variable in git repo group_vars/all.yaml , it stays in awx default groups vars although it is gone.
Change value of variable - OK
Specific groups variables are not effected by this
Specific host vars  are not effected by this.

Inventory has these "Enabled Options":
 Overwrite local groups and hosts from remote inventory source
 Overwrite local variables from remote inventory source
 Update on project update

``` 
diff --git a/inventory/avv-prod/group_vars/all.yaml b/inventory/avv-prod/group_vars/all.yaml
index af2262cf..2a7c90de 100644
--- a/inventory/avv-prod/group_vars/all.yaml
+++ b/inventory/avv-prod/group_vars/all.yaml
@@ -52,4 +52,3 @@ git_dest: "{{ lookup('env', 'PWD') }}"
 grafana_url: "https://grafana.host.zz"
 grafana_user: "admin"
 grafana_password: "123456"
-test_var: test_zzz
```

After project refresh (+inventory update on dependency): 

https://awx-test.host.zz/#/inventories/inventory/8/details
Inventories -> avv-prod -> Variables 

```
<skiped>
grafana_url: "https://grafana.host.zz"
grafana_user: "admin"
grafana_password: "123456"
test_var: test_zzz
```

### AWX version

19.5.1

### Installation method

kubernetes

### Modifications

yes

### Ansible version

2.12.1

### Operating system

Debian 10.11

### Steps to reproduce

Install awx 19.5.1 using operator latest on Kubernetes
Create  inventory which is sourced from project using options:
```
 Overwrite local groups and hosts from remote inventory source
 Overwrite local variables from remote inventory source
 Update on project update
```
Add variable to inventory git repo group_vars/all.yaml
Commit git, push git,  update project on awx.
Check your variable in awx inventory group all. 
Delete it in git 
Commit git, push git,  update project on awx.
Check your variable in awx inventory group all.   

  

### Expected results

It should be gone.

### Actual results

It still exists.

### Additional information

[...]

PR #13731, opened 2023-03-21
[constructed-inventory] Force overwrite all vars for constructed inventory

AlanCoding:
##### SUMMARY
There's an issue for this...

https://github.com/ansible/awx/issues/13714

Item from there

> overwrite inventory variables in imports, opposed to merging them, which is what is does now, and is very wrong

That - that's what this addresses.

Corresponding test case is `test_all_group_variables` and this will flip that to passing.

tldr - If we don't have this, the constructed inventories will (eventually, over time) fail to accurately mirror the input inventories.

##### ISSUE TYPE

##### COMPONENT NAME
 - API


AlanCoding:
Link related issue https://github.com/ansible/awx/issues/11623

relrod (approved):

PR #13734, opened 2023-03-21
Allow BYTES_DISPLAY job settings to be set via UI

CWollinger:
##### SUMMARY

Able to view and set the parameters `STDOUT_MAX_BYTES_DISPLAY ` & `EVENT_STDOUT_MAX_BYTES_DISPLAY` via the UI (/settings/jobs)

Already able to set the parameters via the API

##### ISSUE TYPE

 - Enhanced Feature

##### COMPONENT NAME

 - UI

##### AWX VERSION

```
devel
```


mabashian:
👋 hello there - I just wanted to chime in on this change and say that when we transitioned the UI of AWX from AngularJS to React we talked about each of the available settings and whether or not to expose them via the UI.  That was several years ago now but I think the concern with these two fields in particular were that toggling them runs the risk of significantly impacting the system and it's performance so we decided to deliberate exclude them from the interface while still making them available via the API.  The hope was that an administrator wouldn't stumble upon them and make changes without really understanding the implications.  I'm not opposed to making them available via the UI but I figured I'd add a little context here since I think it's pertinent.

PR #13740, opened 2023-03-22
Remove --until option from the `host_metric` CLI

AparnaKarve:
##### SUMMARY

Remove the `--until` option from the CLI

The ADR was recently updated where it now states that the `--until` option should be removed from the CLI utility. 
The rationale being - if a host is managed in July and then in August, the last managed date would move to August. If we now run the CLI with `–since=2022-07-01 –until=2022-07-31`, then the CLI will not output this host, which is not desirable

##### ISSUE TYPE

##### COMPONENT NAME

 - CLI

##### AWX VERSION

```
awx: 21.13.1.dev132+g790bbeb62d
```

##### ADDITIONAL INFORMATION

```
Before:
--until option was supported 

After:
If  --until is used, the user would see the below message

usage: awx-manage host_metric [-h] [--since SINCE] [--json [JSON]] [--csv [CSV]] [--tarball] [--rows_per_file ROWS_PER_FILE] [--version] [-v {0,1,2,3}] [--settings SETTINGS]
                              [--pythonpath PYTHONPATH] [--traceback] [--no-color] [--force-color] [--skip-checks]
awx-manage host_metric: error: unrecognized arguments: --until 2023-03-22

```

### DIFF
--- a/awx/main/management/commands/host_metric.py
+++ b/awx/main/management/commands/host_metric.py
@@ -185,7 +185,6 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument('--since', type=datetime.datetime.fromisoformat, help='Start Date in ISO format YYYY-MM-DD')
-        parser.add_argument('--until', type=datetime.datetime.fromisoformat, help='End Date in ISO format YYYY-MM-DD')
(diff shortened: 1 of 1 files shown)


h-kataria:
@TheRealHaoLiu @Ladas please review. This is to address one of the additional requirements from ADR 1.1

AlanCoding (approved):

PR #13750, opened 2023-03-27
Only select task instance that are ready and enabled

TheRealHaoLiu:
When select a queue for task instance to run task only select task instance that are ready and enabled

##### SUMMARY
When select a queue for task instance to run task only select task instance that are ready and enabled

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.13.1.dev185+g58ae461e1a.d20230327
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/dispatch/__init__.py
+++ b/awx/main/dispatch/__init__.py
@@ -19,7 +19,18 @@ def get_task_queuename():
     if os.getenv('AWX_COMPONENT') == 'web':
         from awx.main.models.ha import Instance
 
-        return Instance.objects.filter(node_type__in=['control', 'hybrid']).order_by('?').first().hostname
+        return (
+            Instance.objects.filter(
+                node_type__in=[Instance.Types.CONTROL, Instance.Types.HYBRID],
+                node_state=Instance.States.READY,
+                enabled=True,
+            )
+            .only('hostname')
+            .order_by('?')
+            .first()
+            .hostname
+        )
+
     else:
         return settings.CLUSTER_HOST_ID
 


gamuniz (approved):

awx/main/dispatch/__init__.py
```diff
@@ -19,7 +19,17 @@ def get_task_queuename():
     if os.getenv('AWX_COMPONENT') == 'web':
         from awx.main.models.ha import Instance
 
-        return Instance.objects.filter(node_type__in=['control', 'hybrid']).order_by('?').first().hostname
+        return (
+            Instance.objects.filter(
+                node_type__in=['control', 'hybrid'],
+                node_state__in=['ready'],
+                enabled__in=['true'],
```
AlanCoding: This is just kind of weird, you should be able to do `enabled=True` for this kwarg, Likewise, `node_state='ready'`, although I'm skeptical that node_state should apply to these types of nodes

awx/main/dispatch/__init__.py
```diff
@@ -19,7 +19,17 @@ def get_task_queuename():
...
+            )
+            .order_by('?')
+            .first()
+            .hostname
```
AlanCoding, line 31: if you're only going to use the hostname property, you might as well add `.only('hostname')` to the queryset for performance.

awx/main/dispatch/__init__.py
TheRealHaoLiu: yah mostly... symatry and figure later on its easier to append to the list that's why i did __in

AlanCoding: Apparently, `Instance.States.READY` is a fancy way to type "ready"

TheRealHaoLiu: i will change it

awx/main/dispatch/__init__.py
TheRealHaoLiu: added

awx/main/dispatch/__init__.py
TheRealHaoLiu: updated, and also i change 'control' and 'hybrid' to Instance.Types.CONTROL and Instance.Types.HYBRID

AlanCoding (approved):

PR #13753, opened 2023-03-27
Catch SynchronousOnlyOperation and get setting in thread

TheRealHaoLiu:
##### SUMMARY
If trying to get setting from async context (in daphne) catch SynchronousOnlyOperation error and retry in a thread

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.13.1.dev185+gce7337e545.d20230327
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/conf/settings.py
+++ b/awx/conf/settings.py
@@ -11,7 +11,7 @@ from concurrent.futures import ThreadPoolExecutor
 from django.conf import LazySettings
 from django.conf import settings, UserSettingsHolder
 from django.core.cache import cache as django_cache
-from django.core.exceptions import ImproperlyConfigured
+from django.core.exceptions import ImproperlyConfigured, SynchronousOnlyOperation
 from django.db import transaction, connection
 from django.db.utils import Error as DBError, ProgrammingError
 from django.utils.functional import cached_property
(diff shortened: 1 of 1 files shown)


TheRealHaoLiu:
testing:
- remove @shanemcd's hack
- disable redis cache 
- confirm that we were able to reproduce the bug 
```
packages/django/db/models/sql/compiler.py", line 1173, in execute_sql
tools_awx_1 |     cursor = self.connection.cursor()
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/utils/asyncio.py", line 31, in inner
tools_awx_1 |     raise SynchronousOnlyOperation(message)
tools_awx_1 | django.core.exceptions.SynchronousOnlyOperation: You cannot call this from an async context - use a thread or sync_to_async.
```

note: this problem is reproducible in devel

switch to code in this PR... so far no SynchronousOnlyOperation error... im swallowing that error now tho... does it means we fix it or...did it not?

i'm skeptical of this fix TBH

TheRealHaoLiu:
if this problem still exist 
logger will use logging.WARNING if fail to get `settings.LOG_AGGREGATOR_LEVEL`
https://github.com/ansible/awx/blob/devel/awx/main/utils/filters.py/#L108-L123

AlanCoding:
I'm suspecting that this directly causes the api-test check failure we see here.

AlanCoding (approved):

PR #13764, opened 2023-03-29
HostMetrics cleanup task and more Subscription info

slemrmartin:
##### SUMMARY

- Subscription Settings page: Deleted host metrics and Active previously deleted info. Visible when `SUBSCRIPTION_USAGE_MODEL == 'unique-managed_hosts'`
  - https://issues.redhat.com/browse/AA-1642
  - https://issues.redhat.com/browse/AA-1634
- Host Metric cleanup task - soft deletes HostMetric records which were automated more than 12 months ago
  - https://issues.redhat.com/browse/AA-1593

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - UI
 - Other

##### AWX VERSION

```
21.14.1.dev34+gba108660af
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/management/commands/cleanup_host_metrics.py
+++ b/awx/main/management/commands/cleanup_host_metrics.py
@@ -1,0 +1,22 @@
+from awx.main.models import HostMetric
+from django.core.management.base import BaseCommand
+from django.conf import settings
+
+
+class Command(BaseCommand):
+    """
+    Run soft-deleting of HostMetrics
+    """
+
+    help = 'Run soft-deleting of HostMetrics'
+
+    def add_arguments(self, parser):
+        parser.add_argument('--months-ago', type=int, dest='months-ago', action='store', help='Threshold in months for soft-deleting')
+
+    def handle(self, *args, **options):
+        months_ago = options.get('months-ago') or None
+
+        if not months_ago:
+            months_ago = getattr(settings, 'CLEANUP_HOST_METRICS_THRESHOLD', 12)
+
+        HostMetric.cleanup_task(months_ago)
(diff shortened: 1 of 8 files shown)


Ladas (approved):
This looks great 👍

awx/main/management/commands/cleanup_host_metrics.py
```diff
@@ -0,0 +1,22 @@
...
+
+    help = 'Run soft-deleting of HostMetrics'
+
+    def add_arguments(self, parser):
+        parser.add_argument('--months-ago', type=int, dest='months-ago', action='store', help='Threshold in months for soft-deleting')
+
+    def handle(self, *args, **options):
+        months_ago = options.get('months-ago') or None
+
+        if not months_ago:
```
john-westcott-iv, line 19: This prevents me from doing `--months-ago=0` should that be allowed?
I can also do `-1`, do you want to do any kind of validation on it?

awx/main/models/inventory.py
```diff
@@ -849,6 +851,16 @@ def soft_restore(self):
             self.deleted = False
             self.save(update_fields=['deleted'])
 
+    @classmethod
+    def cleanup_task(cls, months_ago):
```
john-westcott-iv, line 894: `months_ago` can come from a setting. Do you want to do any kind of check to make sure its a valid, positive int? Admittedly, a user would have to manually enter something in the defaults file to change this from its default but if we ever exposed these settings later having validation on them might be good.

awx/main/management/commands/cleanup_host_metrics.py
slemrmartin: it's not supposed to be 0. Negative values will be checked in the `cleanup_task` method

awx/main/models/inventory.py
slemrmartin: nice catch, implemented

PR #13789, opened 2023-04-02
Adding basic validation for local passwords

john-westcott-iv:
##### SUMMARY

Adds 4 definable password rules for local user passwords:
  * Min length
  * Min number of upper case characters
  * Min number of digits
  * Min number of special characters

Also synced and prettied the UI files:
  * src/screens/Setting/shared/data.allSettingOptions.json
  * src/screens/Setting/shared/data.allSettings.json

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```


awx/sso/conf.py
```diff
@@ -1603,6 +1603,50 @@ def get_saml_entity_id():
     ],
 )
 
+register(
+    'LOCAL_PASSWORD_MIN_LENGTH',
+    field_class=fields.IntegerField,
+    min_value=0,
```
fosterseth, line 1609: does the django validation have a minimum? would it make sense to restrict the min to whatever that value is?

awx/api/serializers.py
```diff
@@ -995,6 +995,24 @@ def validate_password(self, value):
         django_validate_password(value)
         if not self.instance and value in (None, ''):
             raise serializers.ValidationError(_('Password required for new User.'))
+
+        # Check if a password is too long
+        password_max_length = User._meta.get_field('password').max_length
+        if len(value) > password_max_length:
+            raise serializers.ValidationError(_('Password max length is {}'.format(password_max_length)))
+        if getattr(settings, 'LOCAL_PASSWORD_MIN_LENGTH', 0) and len(value) < getattr(settings, 'LOCAL_PASSWORD_MIN_LENGTH'):
```
fosterseth, line 1003: if these settings are in the db with default values already, would
```python3
if len(value) < getattr(settings, 'LOCAL_PASSWORD_MIN_LENGTH'):
```
suffice?

awx/sso/conf.py
john-westcott-iv: Technically it can't be blank but you can set a password to `a` (so min 1) but I wanted `0` to mean "check disabled".

awx/api/serializers.py
john-westcott-iv: Probably, I was thinking that since 0 means "no check" we would start with that to prevent the second checks from even being looked at. But I suppose its only a string compare/small loops so the amount of time saved by the short circuiting the if clause is probably negligible.

fosterseth (approved):

jay-steurer (approved):

PR #13792, opened 2023-04-03
Analytics upload: HostMetrics hybrid sync

slemrmartin:
##### SUMMARY

Uploading HostMetrics to the analytics (CLI awx-manage gather_analytics)
It adds `host_metric_table.csv` to the files in the sent tarballs. 

Unlike other collectors, this one introduces hybrid full/incremental sync. It means it performs full sync of whole table once per 30 days. Incremental sync otherwise. 

PR extends the `@register` decorator with `full_sync_interval` - integer: number of days when the full sync should be performed instead of incremental. If not provided, incremental sync is performed as before. 

https://issues.redhat.com/browse/AA-1632

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - Other
 - Analytics

##### AWX VERSION

```
awx: 21.14.1.dev30+g04d5e020de.d20230404
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/analytics/collectors.py
+++ b/awx/main/analytics/collectors.py
@@ -6,7 +6,7 @@ import platform
 import distro
 
 from django.db import connection
-from django.db.models import Count
+from django.db.models import Count, Min
 from django.conf import settings
 from django.contrib.sessions.models import Session
 from django.utils.timezone import now, timedelta
@@ -35,7 +35,7 @@ data _since_ the last report date - i.e., new data in the last 24 hours)
 """
 
 
-def trivial_slicing(key, since, until, last_gather):
+def trivial_slicing(key, since, until, last_gather, **kwargs):
     if since is not None:
         return [(since, until)]
 
@@ -48,7 +48,7 @@ def trivial_slicing(key, since, until, last_gather):
     return [(last_entry, until)]
 
 
-def four_hour_slicing(key, since, until, last_gather):
+def four_hour_slicing(key, since, until, last_gather, **kwargs):
     if since is not None:
         last_entry = since
     else:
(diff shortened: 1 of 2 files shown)


awx/main/analytics/collectors.py
```diff
@@ -69,6 +69,54 @@ def four_hour_slicing(key, since, until, last_gather):
         start = end
 
 
+def host_metric_slicing(key, since, until, last_gather, **kwargs):
+    """
+    Slicing doesn't start 4 weeks ago, but sends whole table monthly or first time
+    """
+    from awx.main.models.inventory import HostMetric
+
+    if since is not None:
```
Ladas, line 78: isn't `since` always passed here?

slemrmartin: `since` is `None ` by default: 
- https://github.com/ansible/awx/blob/devel/awx/main/analytics/core.py#L164
- the only way how it isn't None is CLI `awx-manage gather_analytics --since ...`

Ladas:
this looks great, doing incremental sync with full sync in the beginning and then once a month 👍

awx/main/analytics/core.py
```diff
@@ -259,10 +260,19 @@ def gather(dest=None, module=None, subset=None, since=None, until=None, collecti
                 # These slicer functions may return a generator. The `since` parameter is
                 # allowed to be None, and will fall back to LAST_ENTRIES[key] or to
                 # LAST_GATHER (truncated appropriately to match the 4-week limit).
+                #
+                # Or it can force full table sync if interval is given
+                kwargs = dict()
+                full_sync_enabled = False
+                if func.__awx_full_sync_interval__:
+                    last_full_sync = last_entries.get(f"{key}_full")
+                    full_sync_enabled = not last_full_sync or last_full_sync < now() - timedelta(days=func.__awx_full_sync_interval__)
```
john-westcott-iv, line 269: Not a required change but do you want to add `_days` to the variable name to be clear on the units?

john-westcott-iv (approved):
Do we need to do any kind of load testing for the full files? i.e. if I had a very large number of hosts could I generate a file too big to be uploaded to c.r.c?

fosterseth (approved):

Ladas:
> Do we need to do any kind of load testing for the full files? i.e. if I had a very large number of hosts could I generate a file too big to be uploaded to c.r.c?

@john-westcott-iv there is a logic that will split the file into many, if the size of the tarball is bigger than 100MB

awx/main/analytics/core.py
slemrmartin: right, we're going to replace this code for 2.5, so I'll do it there

Issue #7804, opened 2020-08-03
Allow running with formerly-successful projects


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

If a project's last update is failed, you cannot run a JT that uses that project. Hard fail implemented here, maybe in some other ways as well:

https://github.com/ansible/awx/blob/708244881cc6cd04722d3bc36d7d6036bffc329a/awx/main/tasks.py#L1858-L1863

There are several reasons a project update could fail, and not all of them should break running of a previously-known commit from a prior project update. If a project update fails, there's no way to "revert" to the last successful commit, and jobs will refuse to run as things stand. Some of these scenarios, you _could_ keep running jobs successfully, but AWX _blocks_ the user from doing that.

Drafting some scenarios for that:

##### Project branch deletion

 - a project in AWX defines `scm_branch` to "master"
 - the repo changes its default branch to another name, like "main"
 - the maintainers delete the "master" branch
 - subsequent project update fails

Jobs will no longer run. The last-known SCM revision can still be obtained from the remote.

##### Local commit still available

In conventional single-node installs, you know for sure that the the local filesystem is up-to-date with the last project update.

 - A general error happens with the SCM (connectivity or otherwise)
 - A project update runs, it marks the project as failed

Jobs will no longer run. This behavior is debatable - in some cases people would prefer to stop jobs from running if the latest could not be pulled. In some cases, they would prefer for the server to keep using last-known revisions, which will use the local copy in some cases.

----

Some very preliminary ideas:

 - [ ] Introduce a configurable, either on project / organization / system that allows running a job with a failed project, or some other kind of implementation which will allow jobs to keep running with the previously-known revision
 - [ ] In the event that a job does run with a project that had a previously failed update, put up some kind of warning

AlanCoding:
The patch just changes the behavior, non-configurable. I think this is okay. @ffirg maybe we should discuss.

PR #13796, opened 2023-04-04
Removed automatic failure of job template launch when last project update is failed and update on launch is enabled

gamuniz:
##### SUMMARY

The current implementation of project models treats a failure of project updates as an automatic failure of a job template which uses the project. This only makes sense if the project which failed does not automatically update revision on launch. If a project has update revision on launch enabled the job template should try to update the project.  

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 22.0.1.dev1+g4f2bdc230b

```

##### ADDITIONAL INFORMATION

Steps to reproduce:
1. create project with update on launch enabled
2. create job template that depends on that project
3. in order to reproduce the condition awx-manage shell_plus --quiet ProjectUpdate.objects.filter(project_id=$PROJECTID).last().update_fields(status='failed')
4. attempt to launch job_template created in step 2 (should get a 400)

Once the issue is reproduced checkout PR branch and retry the launch. It will launch another dependent project sync and this time should launch the template(not throw a 400).

```

```


AlanCoding (approved):
Very strongly need some testing / documentation / release notes, but I agree with the change.

AlanCoding:
Link https://github.com/ansible/awx/issues/7804

jay-steurer (approved):

Issue #13781, opened 2023-03-30
AWX Collection - Inventory module doesnt allow to create constructed inventory


CFSNM:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

When using the inventory module from the AWX Collection, trying to create a constructed inventory using 2 input inventories is failing.

### AWX version

devel

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

N/A

### Modifications

no

### Ansible version

2.15

### Operating system

RHEL

### Steps to reproduce

Passing these parameters to the inventory module:
```
'module_args': {'name': 'MaximumCheek�', 'description': 'hello world', 'organization': 'Organization - LieComputerꬒ', 'kind': 'constructed', 'input_inventories': ['Inventory - DepressionCoupleᙘ', 'Inventory - SleepSurround�']
```

The list of input inventories are 2 inventories which have been created, the same for the organization

### Expected results

The constructed inventory should be created correctly.

### Actual results

Getting this error message:
The requested object could not be found at /api/v2/constructed_inventories/33/input_inventories/.

relrod:
Found this same issue last night.

Refs #13779 for a really hacky solution, but I don't think that's the right way.

@AlanCoding ideas?

relrod:
The break is https://github.com/ansible/awx/pull/13730

AlanCoding:
I don't think it would disrupt much to point the url link back to `/api/v2/inventories/id/` and then add an additional link for `/api/v2/constructed_inventories/id/`, for informational purposes. But @mabashian was hitting a similar issue with the UI for some reason, and we should get a general consensus about what direction we want to go here.

Jaapis:
Tested and it is working correctly as expected.

PR #13797, opened 2023-04-04
Only use constr. inv URL when req comes from it

relrod:
##### SUMMARY

When the API request is for /inventories/id use that as the URL in the API response. When the request is for /constructed_inventories/id use that.

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/tests/functional/api/test_inventory.py
+++ b/awx/main/tests/functional/api/test_inventory.py
@@ -680,3 +680,20 @@ class TestConstructedInventory:
         inv_src = constructed_inventory.inventory_sources.first()
         assert inv_src.update_cache_timeout == 55
         assert inv_src.limit == 'foobar'
+
+    def test_get_absolute_url_for_constructed_inventory(self, constructed_inventory, admin_user, get):
+        """
+        If we are using the normal inventory API endpoint to look at a
+        constructed inventory, then we should get a normal inventory API route
+        back. If we are accessing it via the special constructed inventory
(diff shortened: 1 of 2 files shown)


awx/main/tests/functional/api/test_inventory.py
```diff
@@ -680,3 +680,20 @@ def test_create_constructed_inventory(self, constructed_inventory, admin_user, p
...
+        """
+
+        url_const = reverse('api:constructed_inventory_detail', kwargs={'pk': constructed_inventory.pk})
+        url_inv = reverse('api:inventory_detail', kwargs={'pk': constructed_inventory.pk})
+
+        const_r = get(url=url_const, user=admin_user, expect=200)
+        inv_r = get(url=url_inv, user=admin_user, expect=200)
+        assert const_r.data['url'] == url_const
+        assert inv_r.data['url'] == url_inv
+        assert inv_r.data['url'] != const_r.data['url']
```
AlanCoding, line 699: inside the `inv_r`, could you make sure that the `constructed_url` inside of `related` still links to the expected location?

AlanCoding (approved):
This is the best solution to the problem, I think we should get it in.

awx/main/tests/functional/api/test_inventory.py
relrod: Good idea, done.

relrod:
Refs #13781

AlanCoding (approved):

PR #13813, opened 2023-04-06
[wip] Allow switching DB in dev env

fosterseth:
##### SUMMARY

Allows you to name your own db volume so that you can switch between branches more easily.

Usage:

`DBNAME=sbf make docker-compose`

this will create a docker volume called "sbf" that will persist on disk.

You can then switch back to devel or some other branch, which will use a different branch.

not setting DBNAME defaults to `tools_awx_db`

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -498,6 +498,7 @@ EXECUTION_NODE_COUNT ?= 0
 MINIKUBE_CONTAINER_GROUP ?= false
 MINIKUBE_SETUP ?= false # if false, run minikube separately
 EXTRA_SOURCES_ANSIBLE_OPTS ?=
+DBNAME ?= tools_awx_db
 
 ifneq ($(ADMIN_PASSWORD),)
 	EXTRA_SOURCES_ANSIBLE_OPTS := -e admin_password=$(ADMIN_PASSWORD) $(EXTRA_SOURCES_ANSIBLE_OPTS)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -519,9 +520,11 @@ docker-compose-sources: .git/hooks/pre-commit
 	    -e enable_ldap=$(LDAP) \
 	    -e enable_splunk=$(SPLUNK) \
 	    -e enable_prometheus=$(PROMETHEUS) \
-	    -e enable_grafana=$(GRAFANA) $(EXTRA_SOURCES_ANSIBLE_OPTS)
+	    -e enable_grafana=$(GRAFANA) $(EXTRA_SOURCES_ANSIBLE_OPTS) \
+	    -e dbname=$(DBNAME) \
 
 docker-compose: awx/projects docker-compose-sources
+	-docker rm tools_postgres_1
```
AlanCoding, line 527: this is a little weird. Can we do without it?

fosterseth: I agree this is weird and tried to get it work without this. It seems to be a limitation of docker compose -- changing the volume information in the docker-compose.yml file doesn't seem to change the volume the container uses if that container already exists.

I may try to find another way to do this.

fosterseth:
stale, closing..

PR #13831, opened 2023-04-11
Analytics API: Permissions for System Auditor

slemrmartin:
##### SUMMARY

Adds permissions for System Auditor for POST/OPTIONS requests to Analytics API.
Required because cloud's API is based on POST requests even for lists

- https://issues.redhat.com/browse/AA-1650

Also changes permissions for `/api/v2/analytics/` from everyone to the same like other analytics URLs (admins/auditors)

- https://issues.redhat.com/browse/AAP-11166

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 22.0.1.dev32+g805c55bee2
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/api/permissions.py
+++ b/awx/api/permissions.py
@@ -25,6 +25,7 @@ __all__ = [
     'UserPermission',
     'IsSystemAdminOrAuditor',
     'WorkflowApprovalPermission',
+    'AnalyticsPermission',
 ]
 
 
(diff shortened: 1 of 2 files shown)


awx/api/permissions.py
```diff
@@ -250,3 +251,16 @@ def has_permission(self, request, view):
...
+
+class AnalyticsPermission(permissions.BasePermission):
+    """
+    Allows GET/POST/OPTIONS to system admins and system auditors.
+    """
+
+    def has_permission(self, request, view):
+        if not (request.user and request.user.is_authenticated):
+            return False
+        if request.method in ["GET", "POST", "OPTIONS"]:
```
john-westcott-iv, line 264: In the view you already check for GET/POST/OPTIONS and 500 if its not one of them. Thinking about flexibility in the future, should we change this and the two likes below it so that we return False if the user is not authed and then just make the method:
```
return request.user.is_superuser or request.user.is_system_auditor
```
Or do you envision one day having a delete function which could be called by a superuser but not callable by an auditor?

slemrmartin: @john-westcott-iv yes, that's possible, also there will be PUT/PATCH in 2.5 (I suppose, the cloud API is quite non-standard) - Automation Calculator page could send data to cloud, and that's for admins only

john-westcott-iv (approved):

PR #13834, opened 2023-04-11
AAP-18284 Added a new logger configuration for activity_stream

lucas-benedito:
##### SUMMARY
- Added a new handler for activity_stream with formatter type json
- Added the new handler at the awx.analytics logger

The activity stream is a critical security log that can only be used from UI or when using an external logging aggregation.
Creating an activity_stream.log allows for easier and more effective analysis without relying only on the API endpoint.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - Other

### DIFF
--- a/awx/settings/defaults.py
+++ b/awx/settings/defaults.py
@@ -744,6 +744,9 @@ LOGGING = {
         'json': {'()': 'awx.main.utils.formatters.LogstashFormatter'},
         'timed_import': {'()': 'awx.main.utils.formatters.TimeFormatter', 'format': '%(relativeSeconds)9.3f %(levelname)-8s %(message)s'},
         'dispatcher': {'format': '%(asctime)s %(levelname)-8s [%(guid)s] %(name)s PID:%(process)d %(message)s'},
+        'activity_stream': {
+            'format': '%(asctime)s %(levelname)-8s [%(guid)s] %(name)s %(message)s, Operation: %(operation)s, Actor: %(actor)s, Changes: %(changes)s'
+        },
     },
     # Extended below based on install scenario. You probably don't want to add something directly here.
     # See 'handler_config' below.
@@ -791,7 +794,8 @@ LOGGING = {
         'awx.main.access': {'level': 'INFO'},  # very verbose debug-level logs
         'awx.main.signals': {'level': 'INFO'},  # very verbose debug-level logs
         'awx.api.permissions': {'level': 'INFO'},  # very verbose debug-level logs
-        'awx.analytics': {'handlers': ['external_logger', 'activity_stream'], 'level': 'INFO', 'propagate': False},
+        'awx.analytics': {'handlers': ['external_logger'], 'level': 'INFO', 'propagate': False},
+        'awx.analytics.activity_stream': {'handlers': ['activity_stream']},
         'awx.analytics.broadcast_websocket': {'handlers': ['console', 'file', 'wsrelay', 'external_logger'], 'level': 'INFO', 'propagate': False},
         'awx.analytics.performance': {'handlers': ['console', 'file', 'tower_warnings', 'external_logger'], 'level': 'DEBUG', 'propagate': False},
         'awx.analytics.job_lifecycle': {'handlers': ['console', 'job_lifecycle', 'external_logger'], 'level': 'DEBUG', 'propagate': False},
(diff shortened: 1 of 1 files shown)


awx/settings/defaults.py
```diff
@@ -882,7 +882,7 @@
         'awx.main.access': {'level': 'INFO'},  # very verbose debug-level logs
         'awx.main.signals': {'level': 'INFO'},  # very verbose debug-level logs
         'awx.api.permissions': {'level': 'INFO'},  # very verbose debug-level logs
-        'awx.analytics': {'handlers': ['external_logger'], 'level': 'INFO', 'propagate': False},
+        'awx.analytics': {'handlers': ['external_logger', 'activity_stream'], 'level': 'INFO', 'propagate': False},
```
AlanCoding: I don't think this is going to log what you think it will. This encompasses activity stream and other stuff like job events that you really don't want in the file. Perhaps you would do good to add a new line that does

```python
        'awx.analytics.activity_stream': {'handlers': ['activity_stream']},
```

in order to capture logs sent to here:

https://github.com/ansible/awx/blob/1a56272eaf74f34ba1ee538de1b41d9d6a8278a6/awx/main/signals.py#L73

Good chance that will capture the logs. But we shouldn't take it on faith that that is what you want either. You should do some stuff and then open up the `activity_stream.log` file and see what actually goes in. I highly suspect that the `extra=` sent to the logger will be ignored, and you will get an uninteresting message.

Assuming this is where you land with this all, and that your intent really is to get the _changes_, then your option will be to deal with the python logging formatter. Hopefully that can be done without an additional class, and it will be fairly clean to wrap up.

lucas-benedito: @AlanCoding I agree with the change to select only the awx.analytics.activity_stream events. I overlooked that when setting the logger.
Regarding the outcome on the log file, the current output is a JSON value that contains the important data about the change made, the actor, and the extras.
We may instead process the output to match other logs' format and add the extras because the default logger is not verbose enough for the purpose of troubleshooting.

lucas-benedito: @AlanCoding Based on your feedback I made additional changes.

awx/settings/defaults.py
```diff
@@ -883,6 +886,7 @@
         'awx.main.signals': {'level': 'INFO'},  # very verbose debug-level logs
         'awx.api.permissions': {'level': 'INFO'},  # very verbose debug-level logs
         'awx.analytics': {'handlers': ['external_logger'], 'level': 'INFO', 'propagate': False},
+        'awx.analytics.activity_stream': {'handlers': ['activity_stream'], 'level': 'DEBUG', 'propagate': False},
```
AlanCoding: I don't like either the level or the propagate here. That might misbehave.

Why? We need to propagate so that these logs make it to `awx.analytics` to get processed by the very important `external_logger` handler. If they don't propagate, it may stop going to the rsyslog/remote log aggregator.

the level probably wouldn't matter, but it's bad form. You don't want any level filter, because again, it might not make it up to the parent logger. In practice that would never happen because I know no logs going here will be below DEBUG level, but I still would rather allow _all_ logs through instead of putting a level filter in.

lucas-benedito: I've followed the examples from the other loggers regarding the propagate option and the level was actually planned to be INFO. 
Based on your statement, should I set the propagate to true or add the external_logger to the handlers?

AlanCoding: no, I would just delete "level" and "propagate" here (in this logger). That's the best way to keep this targeted. That way, the _only_ thing it does is send logs going to this logger to the activity stream handler.

AlanCoding (approved):
Looks good, but may need to delay merge until installer changes get unblocked.

AlanCoding:
I want to disclose some conflicts with https://github.com/ansible/awx/pull/14018 I have up.

I agree with the philosophy here, although it requires additional installer work (logrotate I believe).

In that PR, I want to go in the direction of configuring a logger-based filter using the existing `LOG_AGGREGATOR_LOGGERS` setting. This will mean turning on/off activity stream logger, not just for external logger, but for ordinary loggers, based on that setting. In OCP installs, this will go to console (if enabled). That doesn't have anything directly to do about whether it goes to the `tower.log` file or this new log file you are introducing here, but there will be conflicts to resolve in order to respect both of the intents.

As a side-effect of my objectives, I believe I have to move the `propagate` up from `awx.analytics` to the sub-logger of that. I only want to stop the tide of events / facts mostly. Otherwise I want to default to letting logs through.

dmzoneill:
Marking as blocked until the "additional installer work" has been done

Issue #3513, opened 2019-03-27
Provide means of flipping Ansible config TRANSFORM_INVALID_GROUP_CHARS


AlanCoding:
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
 - API

##### SUMMARY
In version 2.10, Ansible will implement a breaking change to group naming. Group names will be automatically renamed by default.

AWX may need some sort of strategy to deal with this.

##### ADDITIONAL INFORMATION

https://github.com/ansible/ansible/blob/d66f3de3dca5d23950c5889858f211da14bc4eb0/lib/ansible/config/base.yml#L1481-L1494

(env: ANSIBLE_TRANSFORM_INVALID_GROUP_CHARS)

Things that will break:

 - uses in `--limit`
 - references like `groups["foo"]` in playbook namespace
 - vars plugins, like in `group_vars/` folder
 - accessing groups via named URL or API
 - things I didn't think of?

At least smart inventories are not a concern, because they don't enable groups.

You can not use the configure-tower-in-tower setting to do this.

You _can_ use `ansible.cfg` in your source control, but this would need to be adjacent to inventory (in the case of SCM inventory, an increasingly common complaint) and your playbook. Is this sufficient? This could argue in favor of the "do nothing" option. Ping @wenottingham

wenottingham:
We may just need to set it by default.

AlanCoding:
> We may just need to set it by default.

You have a good point.

I made a PR to do this (it was lost in the feature changes for compatibility mode) for the built-in sources. For built-in sources you cannot apply your own `ansible.cfg`.

This issue is still not resolved, in that users do not have any sort of visible toggle. I'm more concerned with playbook runs than for inventory imports.

xarses:
Since this is the first search result for `TRANSFORM_INVALID_GROUP_CHARS`, I'm going to link this here https://github.com/ansible/ansible-documentation/issues/89.

AlanCoding:
Merged the PR that deletes this setting.

### Workaround

I have manually confirmed that I can PATCH `/api/v2/settings/jobs/` with

```json
{
    "AWX_TASK_ENV": {
        "ANSIBLE_DEVEL_WARNING": "False",
        "ANSIBLE_TRANSFORM_INVALID_GROUP_CHARS": "never"
    },
}
```

[...]

PR #13837, opened 2023-04-11
Remove Ansible config override to validate group names

AlanCoding:
##### SUMMARY
Connect https://github.com/ansible/awx/issues/3513

The idea is that there is has been enough time for the Ansible content producers to update their inventory plugins to no longer produce group names with the "invalid" characters in them. For a time, we would add these old (now invalid) names via the inventory files that we produced for inventory updates, but we also stopped doing that at some point.

This should be put as a release note.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/models/inventory.py
+++ b/awx/main/models/inventory.py
@@ -1479,8 +1479,6 @@ class PluginFileInjector(object):
     def build_env(self, inventory_update, env, private_data_dir, private_data_files):
         injector_env = self.get_plugin_env(inventory_update, private_data_dir, private_data_files)
         env.update(injector_env)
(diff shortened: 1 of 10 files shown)


gamuniz (approved):

AlanCoding:
I kicked off tests because there is concern about test failures due to this. After that, I will come back here to merge.

relrod (approved):

AlanCoding:
Integration tests are doing fine with this. I may need to call out a workaround in the docs, but otherwise I'm going to merge.

PR #13849, opened 2023-04-12
Check user permissions before fetching system settings

marshmalien:
##### SUMMARY
Only system admins and system auditors can access the settings endpoint, so this PR checks if the user has access before making a `/settings` GET request. This should resolve 403 errors in the instance list and instance details pages.

_Org admin with permission to view instances_
Screenshot 2023-04-12 at 9 44 25 AM

_Org admin without permission to view instances_
Screenshot 2023-04-12 at 9 43 17 AM

##### ISSUE TYPE

##### COMPONENT NAME

 - UI


mabashian:
Here's how I tested this:

1. Create an organization "test"
2. create a "test-admin" user with admin role inside the "test" organization
3. Browse: Administation > Instances

org_admin_instances

mabashian (approved):

akus062381 (changes requested):
The following shows when the user tries to turn on the tech preview:
image

akus062381 (approved):

Issue #13320, opened 2022-12-11
[web/task split] Implement a heartbeat daemon from web nodes to trigger wsrelay connections


relrod:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

Enhancement to Existing Feature

### Feature Summary

At the end of #13319 our websockets infrastructure will look differently than it does now.

We will no longer have a `wsbroadcast` daemon that lives in web pods. Instead, we'll have a `wsrelay`, which fans out messages to web servers, which then send them to clients/browsers.

```mermaid
graph TD;

task[Dispatcher] -->|job event message| wsrelay
wsrelay --> w1[web pod 1]
wsrelay --> w2[web pod 2]
wsrelay -..->|no user on web pod 3, we stay connected but don't send messages to it| w3[web pod 3]
wsrelay --> w4[web pod 4]
w1 --> u1[user 1]
w2 --> u2[user 2]
w2 --> u3[user 3]
w4 --> u4[user 4]
```

But how does `wsrelay` know which web nodes are out there?

The way we are planning to do this, we will run a new, small service in the web pod which periodically (every *n* seconds) beacons out and advertises the web server's existence.

It will do this by sending a message over a `pg_notify` channel. `wsrelay` will (asynchronously) be listening for messages on this `pg_notify` channel, and when it receives a message, it will connect to the web server's relay websocket endpoint, if it needs to.

Importantly, nothing happens on the `wsrelay` side, if a beacon is missed. The only time `wsrelay` takes an action on these beacons is when:

1. It hears the beacon
2. The beacon is coming from a node with a hostname different from the hostname this `wsrelay` instance is sitting on (for VM deployments, we don't need `wsrelay` to connect to its own host)
3. The beacon it hears identifies a web server that it wasn't previously connected to. Either a new hostname, or a different IP address for an existing hostname.

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

N/A

### Current results

[...]

Issue #13322, opened 2022-12-12
[web/task split] New `heartbeet` daemon should catch SIGTERM and send a "going offline" message


relrod:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

Enhancement to Existing Feature

### Feature Summary

The new [`run_heartbeet`](https://github.com/ansible/awx/blob/wsrelay/awx/main/management/commands/run_heartbeet.py) daemon will send a heartbeat message via pg_notify broadcast every *n* seconds.

When it is being shut down (i.e. gets a SIGTERM), we should have it send a similar message with `action=offline`.

This allows wsrelay to act on it a bit faster in the "normal" scale down case.

Refs #13319 

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

Use the `web-task-split` branch of awx-operator and `wsrelay` branch of awx. Scale down a web node, and `wsrelay` will usually "fail" a connection by trying to send something over the closed/closing web socket connection.

It handles this case fine and will just remove the producer and the host from it its internal state.

But we can speed up the process and handle the "normal" case a bit better if we have an actual message come across that says "this web node is going offline".

It's already prepared for this message (I think), but `run_heartbeet.py` needs to be modified to send it.

### Current results

Right now, no message is sent when `run_heartbeet.py` is terminated.

### Sugested feature result

See above :)

thedoubl3j:
@jessicamack is taking this during hardening, this is not blocking the feature

PR #13858, opened 2023-04-13
Catch SIGTERM or SIGINT and send offline message

jessicamack:
##### SUMMARY

Handle SIGTERM and SIGINT in the hearbeet daemon. Relates to #13320 

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 21.14.1
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/management/commands/run_heartbeet.py
+++ b/awx/main/management/commands/run_heartbeet.py
@@ -55,7 +55,7 @@ class Command(BaseCommand):
     def notify_listener_and_exit(self, *args):
         with pg_bus_conn(new_connection=False) as conn:
             conn.notify('web_heartbeet', self.construct_payload(action='offline'))
-        sys.exit(1)
+        sys.exit(0)
 
     def do_hearbeat_loop(self):
         with pg_bus_conn(new_connection=True) as conn:
--- a/awx/main/management/commands/run_heartbeet.py
+++ b/awx/main/management/commands/run_heartbeet.py
@@ -58,11 +58,11 @@ class Command(BaseCommand):
         sys.exit(1)
 
     def do_hearbeat_loop(self):
-        while True:
-            with pg_bus_conn(new_connection=True) as conn:
+        with pg_bus_conn(new_connection=True) as conn:
+            while True:
                 logger.debug('Sending heartbeat')
                 conn.notify('web_heartbeet', self.construct_payload())
-            time.sleep(settings.BROADCAST_WEBSOCKET_BEACON_FROM_WEB_RATE_SECONDS)
+                time.sleep(settings.BROADCAST_WEBSOCKET_BEACON_FROM_WEB_RATE_SECONDS)
 
     def handle(self, *arg, **options):
         self.print_banner()
--- a/awx/main/management/commands/run_heartbeet.py
+++ b/awx/main/management/commands/run_heartbeet.py
@@ -3,6 +3,7 @@ import logging
 import os
 import time
 import signal
+import sys
 
 from django.core.management.base import BaseCommand
 from django.conf import settings
@@ -51,14 +52,19 @@ class Command(BaseCommand):
         }
         return json.dumps(payload)
 
+    def notify_listener_and_exit(self, *args):
+        with pg_bus_conn(new_connection=False) as conn:
+            conn.notify('web_heartbeet', self.construct_payload(action='offline'))
+        sys.exit(1)
+
     def do_hearbeat_loop(self):
         with pg_bus_conn(new_connection=True) as conn:
+            signal.signal(signal.SIGTERM, self.notify_listener_and_exit)
+            signal.signal(signal.SIGINT, self.notify_listener_and_exit)
             while True:
                 logger.debug('Sending heartbeat')
                 conn.notify('web_heartbeet', self.construct_payload())
                 time.sleep(settings.BROADCAST_WEBSOCKET_BEACON_FROM_WEB_RATE_SECONDS)
-                signal.signal(signal.SIGTERM, conn.notify('web_heartbeet', self.construct_payload(action='offline')))
-                signal.signal(signal.SIGINT, conn.notify('web_heartbeet', self.construct_payload(action='offline')))
 
     # TODO: Send a message with action=offline if we notice a SIGTERM or SIGINT
     # (wsrelay can use this to remove the node quicker)
--- a/awx/main/management/commands/run_heartbeet.py
+++ b/awx/main/management/commands/run_heartbeet.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 import time
+import signal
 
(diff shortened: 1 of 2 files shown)


awx/main/management/commands/run_heartbeet.py
```diff
@@ -56,6 +57,8 @@ def do_hearbeat_loop(self):
                 logger.debug('Sending heartbeat')
                 conn.notify('web_heartbeet', self.construct_payload())
                 time.sleep(settings.BROADCAST_WEBSOCKET_BEACON_FROM_WEB_RATE_SECONDS)
+                signal.signal(signal.SIGTERM, conn.notify('web_heartbeet', self.construct_payload(action='offline')))
```
fosterseth: the signal only needs to be called once I think. if so we may move this to before the while True loop

AlanCoding: do we need something in the loop to check that the signal had been received and `break` out of the loop too?

AlanCoding:
I tried to test this locally. Steps:

 - run docker-compose
 - in 2nd tab, exec into the awx_devel container, `awx-manage run_heartbeeet`, leave it
 - in 3rd tab, exec in again, `ps aux --forest` and find process, `kill 1234`

When I do this, in the 2nd tab, I find this:

```
(awx) bash-5.1$ awx-manage run_heartbeet

   **********   **********
 ************* *************
*****************************
 ***********HEART***********
  *************************
     *******************
       *************** _._
         *********** /`._ `'.      __
           *******   \  .\| \   _'`  `)
             ***  (``_)  \| ).'` /`- /
              *   `\ `;\_ `\\//`-'` /
                    \ `'.'.| /  __/`
                     `'--v_|/`'`
                         __||-._
                      /'` `-``  `'\\
                     /         .'` )
                     \  BEET  '    )
                      \.          /
                        '.     /'`
                           `) |
                            //
                            '(.
                             `\`.
                               ``
  File "/awx_devel/awx/__init__.py", line 200, in manage
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/management/__init__.py", line 413, in execute
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/management/base.py", line 354, in run_from_argv
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/management/base.py", line 398, in execute
  File "/awx_devel/awx/main/management/commands/run_heartbeet.py", line 70, in handle
    self.do_hearbeat_loop()
  File "/awx_devel/awx/main/management/commands/run_heartbeet.py", line 60, in do_hearbeat_loop
    signal.signal(signal.SIGTERM, conn.notify('web_heartbeet', self.construct_payload(action='offline')))
  File "/usr/lib64/python3.9/signal.py", line 56, in signal
    handler = _signal.signal(_enum_to_int(signalnum), _enum_to_int(handler))
TypeError: signal handler must be signal.SIG_IGN, signal.SIG_DFL, or a callable object
```

Before your branch, it printed "Terminated" and exited, which is desirable.

awx/main/management/commands/run_heartbeet.py
fosterseth: yeah you're right we need to do a sys.exit
we can follow the pattern here https://github.com/fosterseth/awx/blob/e42461d96ffe3979c8d5c4b9e978cee0809fef35/awx/main/scheduler/task_manager.py#L121

jessicamack: > the signal only needs to be called once I think. if so we may move this to before the while True loop

my understanding is that if we want to catch both we need to call signal on both similar to how we do in base.py https://github.com/ansible/awx/blob/devel/awx/main/dispatch/worker/base.py#L129-L130

fosterseth: @jessicamack you're right we need both for SIGTERM and SIGINT

what I meant is that I don't think we need to put them into the while True block

AlanCoding:
you might have a problem with an outdated `black` version for the api-lint check failure. Just a guess.

awx/main/management/commands/run_heartbeet.py
```diff
@@ -50,8 +52,15 @@ def construct_payload(self, action='online'):
...
 
+    def notify_listener_and_exit(self, *args):
+        with pg_bus_conn(new_connection=False) as conn:
+            conn.notify('web_heartbeet', self.construct_payload(action='offline'))
+        sys.exit(1)
+
     def do_hearbeat_loop(self):
         with pg_bus_conn(new_connection=True) as conn:
+            signal.signal(signal.SIGTERM, self.notify_listener_and_exit)
+            signal.signal(signal.SIGINT, self.notify_listener_and_exit)
```
relrod: It might make more sense to register these in `handle()` instead.

relrod: And we can drop the `TODO` above `handle()` then.

awx/main/management/commands/run_heartbeet.py
```diff
@@ -50,8 +52,15 @@ def construct_payload(self, action='online'):
...
             while True:
```
relrod, line 62: Not related to this PR, but worth pondering-

It might make more sense for scaling to move this `while` up to the top of the method, out of the `with` block (and move the sleep out as well).

The difference is: Right now, we hold on to a connection to the db, in between heartbeat messages. We _could_ just connect, send the message, and disconnect each time. This frees up a persistent (mostly unused) connection to the db.

jessicamack: I don't mind rolling that change into this PR. I believe we were having an issue with the number of db connections being used so a change that helps us reduce that and is low effort is worth adding now.

awx/main/management/commands/run_heartbeet.py
jessicamack: signal catches have been moved to `handle()` and `TODO` has been removed

awx/main/management/commands/run_heartbeet.py
```diff
@@ -50,17 +52,22 @@ def construct_payload(self, action='online'):
         }
         return json.dumps(payload)
 
+    def notify_listener_and_exit(self, *args):
+        with pg_bus_conn(new_connection=False) as conn:
+            conn.notify('web_heartbeet', self.construct_payload(action='offline'))
+        sys.exit(1)
```
relrod: This is the normal way to exit the program, so we probably should exit with 0 instead.

relrod (approved):
One more inline comment. I think after that fix, this is good to go.

TheRealHaoLiu:
related to https://github.com/ansible/awx/issues/13322

PR #13869, opened 2023-04-17
Rework PersistentFilter to avoid double API call

keithjgrant:
##### SUMMARY
Fixes race condition, making persistent filter behavior more consistent.

In the previous implementation, persistent filters were restored by the list page. This led to a double-fetch from the API, first before the filters (query string) were applied, then a second time after. This led to a race condition where the second request occasionally returned first and the first (unfiltered) request overwrote the results, displaying an unfiltered list.

The new approach sets the query string directly from the "Back to {X}" button, so the list page never loads without the filters applied.

##### ISSUE TYPE

##### COMPONENT NAME
 - UI
 - Collection

### DIFF
--- a/awx/ui/src/components/PersistentFilters/PersistentFilters.js
+++ b/awx/ui/src/components/PersistentFilters/PersistentFilters.js
@@ -1,25 +1,9 @@
 import { useEffect } from 'react';
-import { useLocation, useHistory } from 'react-router';
+import { useLocation } from 'react-router';
 import { PERSISTENT_FILTER_KEY } from '../../constants';
 
 export default function PersistentFilters({ pageKey, children }) {
   const location = useLocation();
-  const history = useHistory();
-
-  useEffect(() => {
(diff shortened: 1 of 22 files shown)


AlexSCorey (approved):
I've pulled this down and tested it out.  It works as expected.  I was not able to reproduce the original bug though.

Well done!

awx/ui/src/components/PersistentFilters/PersistentFilters.js
```diff
@@ -31,3 +15,13 @@ export default function PersistentFilters({ pageKey, children }) {
 
   return children;
 }
+
+export function getPersistentFilters(key) {
```
marshmalien, line 19: 👍

marshmalien (approved):
Nice work! I pulled this down and verified that the back button applied the persisted filter to the list. There are some old references to the `?restoreFilters` in the unit tests that should be removed before merging. 
cred-filter

PR #13870, opened 2023-04-18
Bumps patternfly and react deps.  Removes codemirror

mabashian:
##### SUMMARY
Brings Patternfly and React dependencies up to latest.  Removes unused codemirror dependency.

##### ISSUE TYPE

##### COMPONENT NAME
 - UI


mabashian:
I'm now realizing that all of our enzyme tests are going to prevent this jump to react 18.  Will come up with a different plan.

PR #13874, opened 2023-04-18
Fix vault credential update error when vault_id is missing

marshmalien:
##### SUMMARY
If a vault credential has a missing `vault_id` input, don't send `vault_id` in the API request payload. 

vault

##### ISSUE TYPE

##### COMPONENT NAME

 - UI


vidyanambiar:
@marshmalien I'm unable to recreate the Vault credential update issue on `devel`. When I edit and save a credential without the vault_id, there is no error. Could some update in the API have fixed it on awx/devel?

vidyanambiar (approved):
lgtm!

Thank you for clarifying on slack that the original credential (without vault_id) needs to be created via the API. Tested the changes locally and they work well 👍

akus062381 (approved):
looks good!

PR #13879, opened 2023-04-18
Updated libraries from Thycotic to Delinea

vician:
##### SUMMARY

Migrated Thycotic integration to new library where it was moved after being bought by Delinea

##### ISSUE TYPE

 - Bug Fix 

##### COMPONENT NAME

awx/main/credential_plugins/tss.py

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

Similar changes to https://github.com/ansible-collections/community.general/pull/5150/files and related to previous PR https://github.com/ansible/awx/pull/12627


rooftopcellist:
Wouldn't this also need to change?
* https://github.com/ansible/awx/blob/33f1c35292e1d468ad479bc4b3ed5a57d455a67b/awx/main/credential_plugins/dsv.py#L5

fosterseth:
wondering if this package is needed https://github.com/fosterseth/awx/blob/886ba1ea7f1c0a40ba0d68cc6f868071b9cb67f0/requirements/requirements.in#L44

or if only dsv is required. If so, we can consider dropping it

vician:
> Wouldn't this also need to change?
> 
>     * https://github.com/ansible/awx/blob/33f1c35292e1d468ad479bc4b3ed5a57d455a67b/awx/main/credential_plugins/dsv.py#L5

That's a different library and I don't have experience with that, sorry. But possibly yes, see https://github.com/DelineaXPM/python-dsv-sdk

vician:
> wondering if this package is needed https://github.com/fosterseth/awx/blob/886ba1ea7f1c0a40ba0d68cc6f868071b9cb67f0/requirements/requirements.in#L44
> 
> or if only dsv is required. If so, we can consider dropping it

Thank you. It is required but needs updating. I added that to the PR. The currently used version is very old and should be considered insecure, I think. It's two different libraries for two different credential plugins.

fosterseth:
@vician the only import I see for `thycotic.secrets.server` was in tss.py, which you change in this PR. Where else is that module needed? or is it implicitly needed by `dsv`?

where is tss module still needed?

vician:
> @vician the only import I see for `thycotic.secrets.server` was in tss.py, which you change in this PR. Where else is that module needed? or is it implicitly needed by `dsv`?
> 
> where is tss module still needed?

This PR is aiming to fix and improve `tss` credential plugin. `dsv` is totally different plugin with different library even it has been created by the same company and it has the same logic. I don't have any experience with `dsv` as we are not using it.

This PR includes all required changes for `tss` to use newer version of the library as the one being use now is not maintained anymore.

fosterseth:
@delineaKrehl 

would you mind taking a look at this PR?

also we are wondering if we can get rid of the `python-tss-sdk` requirement

we see `from thycotic` imports elsewhere in our code, should those also be changed to `from delinea`?

thanks for your input

delineaKrehl (approved):

vician:
`python-tss-sdk` package provides `delinea.secrets.server` library. See https://pypi.org/project/python-tss-sdk/

It is the main library (and dependency) for Thycotic Secret Server credential plugin.

delineaKrehl:
> @delineaKrehl
> 
> would you mind taking a look at this PR?
> 
> also we are wondering if we can get rid of the `python-tss-sdk` requirement
> 
> we see `from thycotic` imports elsewhere in our code, should those also be changed to `from delinea`?
> 
> thanks for your input

We have migrated the majority of our open source projects to the DelineaXPM org from their previously branded locations. You should be able to find a newer release under the DelineaXPM GitHub organization. For many of the repositories this was a rebrand and not a functional change. However, any future bug and feature releases will come via DelineaXPM. If you run across one that you cannot swap to delinea let me know and we'll investigate.

We are in the process of doing updates ourselves to this AWX plug-in and as part of that the team can review any usage of thycotic.

As for removing use of python-tss-sdk removing it's use would generally mean we have to code such functionality twice as we own that SDK and I'd prefer to avoid that. For example, to support a feature request in another Ansible plug-in we are adding it first to the tss-sdk and then will pull it in as a tested and verified feature and just implement the Ansible portion.
I will discuss it with the team and update if needed.

obaranov:
Hi, do you have any updates on that ticket?

gamuniz:
looks like a maintainer did the work here [14207](https://github.com/ansible/awx/pull/14207). closing in favor of that work.

PR #13881, opened 2023-04-18
Fix copy API

TheRealHaoLiu:
##### SUMMARY
In web/task split deployment web and task container no longer share the same redis cache

In the original code we use redis cache to pass the list of sub objects that need to be copied to the new object

In this PR we extracted out the logic that computes the sub_object_list and move it into deep_copy_model_obj task

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### AWX VERSION
```
awx: 22.0.1.dev82+gd49a5345d9
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/api/generics.py
+++ b/awx/api/generics.py
@@ -5,13 +5,11 @@
 import inspect
 import logging
 import time
-import uuid
 
 # Django
 from django.conf import settings
 from django.contrib.auth import views as auth_views
 from django.contrib.contenttypes.models import ContentType
-from django.core.cache import cache
 from django.core.exceptions import FieldDoesNotExist
 from django.db import connection, transaction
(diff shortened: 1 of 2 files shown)


TheRealHaoLiu:
tested by deploying an 2 web 2 task instance on OCP, made a workflow template with 2 job template node copy it and verified the following 
- verify that the copied workflow job have 2 new and unique nodes

AlanCoding:
Did it also copy a link between the nodes?

CFSNM (approved):

AlanCoding:
Screenshot from 2023-04-19 09-33-32

Confirmed it can copy this

AlanCoding (approved):

PR #13884, opened 2023-04-19
Show related dependent_jobs list in the API

AlanCoding:
##### SUMMARY
This is populated by the `DependencyManager` but users cannot see it. I feel that the UI should surface this in some way.

For now, this is just a discussion piece.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -812,8 +812,10 @@ class UnifiedJobSerializer(BaseSerializer):
             res['stdout'] = self.reverse('api:project_update_stdout', kwargs={'pk': obj.pk})
         elif isinstance(obj, InventoryUpdate):
             res['stdout'] = self.reverse('api:inventory_update_stdout', kwargs={'pk': obj.pk})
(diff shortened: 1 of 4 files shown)


AlanCoding:
This seems pretty non-controversial.

fosterseth:
maybe we could add a new tab on job detail page called "Dependent Jobs" that would be a list view of dependent jobs

Or maybe just link to them on the detail page (e.g. 26 and 27 would be links to the jobs)

Screenshot from 2024-02-20 14-57-24

awx/api/serializers.py
```diff
@@ -812,8 +812,10 @@ def get_related(self, obj):
             res['stdout'] = self.reverse('api:project_update_stdout', kwargs={'pk': obj.pk})
         elif isinstance(obj, InventoryUpdate):
             res['stdout'] = self.reverse('api:inventory_update_stdout', kwargs={'pk': obj.pk})
+            res['dependent_jobs'] = self.reverse('api:inventory_update_dependent_jobs_list', kwargs={'pk': obj.pk})
```
fosterseth, line 815: `inventory_update_dependent_jobs_list`

does this still exist?

Issue #13737, opened 2023-03-22
Improvements to the AWX development experience


gundalow:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

Let's use this GitHub Issue to brain-dump problems we've seen.
Including:
* docker-compose
* Makefile targets

### AWX version

devel

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [X] Other

### Installation method

kubernetes

### Modifications

no

### Steps to reproduce

n/a

### Expected results

Easier to develop.

### Actual results

Not as easy as it could be

TheRealHaoLiu:
build for ARM so people on the new M1 mac dont have to keep rebuilding the images

gundalow:
Fail fast
* Check our dependencies are installed before we start running
  * [setuptools](https://groups.google.com/g/awx-project/c/YHZeab7vt-8)
  * [docker-compose](https://groups.google.com/g/awx-project/c/tIvwZYDNErQ)
  * ansible
  * git

Makefile improvements
* Document main targets, as [ansible-core](https://github.com/ansible/ansible/blob/devel/Makefile#L2-L10) does
* Ensure Contributing guide is clearer about which targets to use and when
* Remove older make targets

TheRealHaoLiu:
ability to run "e2e-test" against docker-compose development environment within vscode

TheRealHaoLiu:
ability to use vscode debugger against docker-compose development environment

vscode dev container

Klaas-:
https://github.com/ansible/awx/pull/13641#issuecomment-1460804551
"for the django migration errors, just kill the containers and rerun make docker-compose. there is some kind of race condition the first time starting the containers (when migrations are needed)."

that threw me off when I ran the tests locally :)

fosterseth:
@Klaas- thanks for mentioning that one! we fixed it recently here https://github.com/ansible/awx/pull/13670

gundalow:
`make ui-devel` is undocumented in the README
`/var/lib/awx` needs to be manually created (why) with certain permissions

shanemcd:
[...]

PR #13888, opened 2023-04-19
Make target should not call make directly

TheRealHaoLiu:
##### SUMMARY
https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html

make target should always call make with $(MAKE)

```
Recursive make commands should always use the variable MAKE, not the explicit command name ‘make’
```

##### ISSUE TYPE

##### COMPONENT NAME
 - Other

##### AWX VERSION

```
awx: 22.1.1.dev21+g329fb88bbb.d20230419
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -296,7 +296,7 @@ swagger: reports
 check: black
 
 api-lint:
-	BLACK_ARGS="--check" make black
+	BLACK_ARGS="--check" $(MAKE) black
 	flake8 awx
 	yamllint -s .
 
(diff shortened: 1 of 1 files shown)


TheRealHaoLiu:
https://github.com/ansible/awx/issues/13737

jjwatt (approved):
🤌

AlanCoding (approved):

PR #13890, opened 2023-04-19
Make `state: exists` universal in collection

john-westcott-iv:
##### SUMMARY

In #13725 we added a `state: exists` to the credential module but this could have been made universal for all modules. This PR does that making it a valid state for all API modules. In addition, the `state: exists` previously implemented did not return anything if the credential did exist whereas it would return the `id` of the credential if it had to be made. This would make it impossible to "chain" modules together like:
```
- name: Create a cred
  awx.awx.credential:
    <args>
  state: exists
  register: my_cred

- name: Use cred in JT
  awx.awx.job_template:
    <args>
    credentials:
      - "{{ my_cred.id }}"
```

This PR fixes this by modifying the `get_one` method from the controller_api  module_utl to take a `check_existing` argument which will return the module with the id if exactly one item was found.

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - Collection

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx_collection/plugins/module_utils/controller_api.py
+++ b/awx_collection/plugins/module_utils/controller_api.py
@@ -47,35 +47,14 @@ class ItemNotDefined(Exception):
 class ControllerModule(AnsibleModule):
     url = None
     AUTH_ARGSPEC = dict(
-        controller_host=dict(
-            required=False,
-            aliases=['tower_host'],
-            fallback=(env_fallback, ['CONTROLLER_HOST', 'TOWER_HOST'])),
-        controller_username=dict(
-            required=False,
-            aliases=['tower_username'],
-            fallback=(env_fallback, ['CONTROLLER_USERNAME', 'TOWER_USERNAME'])),
-        controller_password=dict(
-            no_log=True,
-            aliases=['tower_password'],
-            required=False,
-            fallback=(env_fallback, ['CONTROLLER_PASSWORD', 'TOWER_PASSWORD'])),
-        validate_certs=dict(
-            type='bool',
-            aliases=['tower_verify_ssl'],
-            required=False,
-            fallback=(env_fallback, ['CONTROLLER_VERIFY_SSL', 'TOWER_VERIFY_SSL'])),
+        controller_host=dict(required=False, aliases=['tower_host'], fallback=(env_fallback, ['CONTROLLER_HOST', 'TOWER_HOST'])),
+        controller_username=dict(required=False, aliases=['tower_username'], fallback=(env_fallback, ['CONTROLLER_USERNAME', 'TOWER_USERNAME'])),
+        controller_password=dict(no_log=True, aliases=['tower_password'], required=False, fallback=(env_fallback, ['CONTROLLER_PASSWORD', 'TOWER_PASSWORD'])),
+        validate_certs=dict(type='bool', aliases=['tower_verify_ssl'], required=False, fallback=(env_fallback, ['CONTROLLER_VERIFY_SSL', 'TOWER_VERIFY_SSL'])),
         controller_oauthtoken=dict(
-            type='raw',
-            no_log=True,
-            aliases=['tower_oauthtoken'],
-            required=False,
-            fallback=(env_fallback, ['CONTROLLER_OAUTH_TOKEN', 'TOWER_OAUTH_TOKEN'])),
-        controller_config_file=dict(
-            type='path',
-            aliases=['tower_config_file'],
-            required=False,
-            default=None),
+            type='raw', no_log=True, aliases=['tower_oauthtoken'], required=False, fallback=(env_fallback, ['CONTROLLER_OAUTH_TOKEN', 'TOWER_OAUTH_TOKEN'])
+        ),
+        controller_config_file=dict(type='path', aliases=['tower_config_file'], required=False, default=None),
     )
     short_params = {
         'host': 'controller_host',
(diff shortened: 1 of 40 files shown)


awx_collection/plugins/module_utils/controller_api.py
```diff
@@ -446,14 +423,20 @@ def get_one(self, endpoint, name_or_id=None, allow_none=True, **kwargs):
             # Or we weren't running with a or search and just got back too many to begin with.
             self.fail_wanted_one(response, endpoint, new_kwargs.get('data'))
 
+        if check_exists:
+            name_field = self.get_name_field_from_endpoint(endpoint)
+            item_id = response['json']['results'][0]['id']
+            self.json_output['id'] = item_id
+            self.exit_json(**self.json_output)
+
```
john-westcott-iv: Black tore up this file, this is the meat of the actual change.

john-westcott-iv:
@sean-m-sullivan @l3acon could you all also review and 👍 if you are good with this change.

l3acon:
👏🏾 yes this is way better thank you @john-westcott-iv.  It passes my simple test: not overwriting credentials with `exist` state.

Should we add documentation around the returned ID if the resource is present and not created?

john-westcott-iv:
> Should we add documentation around the returned ID if the resource is present and not created?

I think most of the modules don't specify their return values.

awx_collection/plugins/module_utils/controller_api.py
```diff
@@ -446,14 +423,21 @@ def get_one(self, endpoint, name_or_id=None, allow_none=True, **kwargs):
             # Or we weren't running with a or search and just got back too many to begin with.
             self.fail_wanted_one(response, endpoint, new_kwargs.get('data'))
 
+        if check_exists:
+            name_field = self.get_name_field_from_endpoint(endpoint)
+            item_id = response['json']['results'][0]['id']
+            self.json_output['id'] = item_id
```
sean-m-sullivan: In addition to ID, can we get name in the output, right now ALL it outputs is:
```
{
"result": {
    "changed": false,
    "failed": false,
    "id": 1
}
}
```
Should be something like 
```
If response['json']['results'][0]['name']
    self.json_output['name'] = response['json']['results'][0]['name']
```

Also Shouldn't this be one line?
```
self.json_output['id'] = response['json']['results'][0]['id']
````

Comment (john-westcott-iv) [reply awx_collection/plugins/module_utils/controller_api.py]: I ran a little test playbook on an org and we don't return name during cretion/no change/edit or delete:

```
$ ansible-playbook test_organization.yml 
[WARNING]: No inventory was parsed, only implicit localhost is available
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'

PLAY [Test org creation] *******************************************************************************************************************************************************************************************************************************

TASK [Create a new org] ********************************************************************************************************************************************************************************************************************************
[WARNING]: You are running collection version 0.0.1-devel but connecting to AWX version 22.1.1.dev33+gc10ada6f44
ok: [localhost]

TASK [debug] *******************************************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "msg": {
        "changed": false,
        "failed": false,
        "id": 26,
        "warnings": [
            "You are running collection version 0.0.1-devel but connecting to AWX version 22.1.1.dev33+gc10ada6f44"
        ]
    }
}

TASK [Create an existing org] **************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [debug] *******************************************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "msg": {
        "changed": false,
        "failed": false,
        "id": 26,
        "warnings": [
            "You are running collection version 0.0.1-devel but connecting to AWX version 22.1.1.dev33+gc10ada6f44"
        ]
    }
}

TASK [Exist an org] ************************************************************************************************************************************************************************************************************************************
changed: [localhost]

TASK [debug] *******************************************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "msg": {
        "changed": true,
        "failed": false,
        "id": 26,
        "warnings": [
            "You are running collection version 0.0.1-devel but connecting to AWX version 22.1.1.dev33+gc10ada6f44"
        ]
    }
}

TASK [Delete an org] ***********************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [debug] *******************************************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "msg": {
        "changed": false,
        "failed": false,
        "id": 26,
        "warnings": [
            "You are running collection version 0.0.1-devel but connecting to AWX version 22.1.1.dev33+gc10ada6f44"
        ]
    }
}

PLAY RECAP *********************************************************************************************************************************************************************************************************************************************
localhost                  : ok=8    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   

```

So I'd like to stay consistent. 

I'll make that other call one line, no reason for it not to be.

sean-m-sullivan (approved):
Can't wait on this to be merged, thinking about what we want to do with state.

thedoubl3j (approved):
looks good to me. happy with the changes and also big +1 to the idea and implementation of chaining things together

AlanCoding:
I wanted to do a teeny tiny bit of research on prior work for similar situations. Just looking in the `community.general` collection (and not exhaustively) I came up with

https://docs.ansible.com/ansible/latest/collections/community/general/rhevm_module.html#ansible-collections-community-general-rhevm-module

has choices
- "absent"
- "cd"
- "down"
- "info"
- "ping"
- "present" ← (default)
- "restarted"
- "up"

And

https://docs.ansible.com/ansible/latest/collections/community/general/proxmox_kvm_module.html#ansible-collections-community-general-proxmox-kvm-module

has choices
- "present" ← (default)
- "started"
- "absent"
- "stopped"
- "restarted"
- "current"

Compared to the total number of modules that use `state` as a parameter, I don't believe these are tremendously common, but they do exist. These are doing something that's different, by nature, than what this module is doing. It seems that the distinctions align well enough for my taste.

AlanCoding (approved):

jay-steurer (approved):

PR #13892, opened 2023-04-20
[WIP] Upgrade django to 4.2 LTS (psycopg3)

john-westcott-iv:
##### SUMMARY

Upgrade for Django LTS

Had to update the old migrations because we have both psycopg 2 and 3 and Django will give preference to 3. 
Also had to upgrade some libraries for compatibility with the new Django:
  * psycopg3
  * djangorestframework
  * django-oauth-toolkit

Restframework and psycopg3 so far look not too bad but django-oauth-toolkit caused issues.
After reverting from django-oauth-toolkit 2.x down to 1.7 its fine. 2.x had breaking changes that definitely broke us. In a future effort we should attempt a specific upgrade of django-oauth-toolkit and work through its issues.

To Do:

- [x] Get CI checks green
- [x] Run all collection integration tests
- [x] Clean up psyopg2.copy_expert
- [x] Fix deprecation errors from tests
- [ ] Test in different configs w and w/o FIPS mode

##### ISSUE TYPE

 - New or Enhanced Feature
 
##### COMPONENT NAME

 - Other

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```


john-westcott-iv:
After many trials we have decided to supersede this pr with a copy that does not include psycopg3. At a later time we will attempt an upgrade to psycopg3.
See #13961

PR #13895, opened 2023-04-20
Use separate module for pytest settings

AlanCoding:
##### SUMMARY
Just trying to take this to its logical conclusion

Update: now believe this is needed to address https://github.com/ansible/awx/issues/13901

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API


john-westcott-iv (approved):
Confirmed that this makes the tests run locally as well as the UI still using the postgres DB. Thanks for fixing this!

PR #13896, opened 2023-04-20
Fallback on PYTHON path in Makefile

jjwatt:
##### SUMMARY

If I understand correctly, this change should make '$(PYTHON)' work how we want it to everywhere. Before this change, on develpers' machines that don't have a 'python3.9' in their path, make would fail. With this change, we will prefer python3.9 if it's available, but we'll take python3 or python otherwise.

I think setting this might be an artifact from having several minor versions of Python 3 available. It may not even still be relevant. Let me know if I'm misunderstanding this or if there's somewhere where this will fail. My main goal is to avoid `make` failing because a developer does not have a bin named `python3.9` in their path (or the equiv symlink or w/e). Like, maybe I have `python3` and `python3.11` and then I have to set `PYTHON` myself for running `make`. I think we can find something that works for these cases.

I really think this does what we mean for all cases: make from dev machine, make from container, make from scripts. It will even run the venv path one if started from venv.

##### ISSUE TYPE

##### COMPONENT NAME

 - API
 - UI
 - Collection
 - CLI
 - Docs
 - Other

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 -include awx/ui_next/Makefile
 
-PYTHON ?= python3.9
+PYTHON := $(notdir $(shell for i in python3.9 python3; do command -v $$i; done|sed 1q))
 DOCKER_COMPOSE ?= docker-compose
 OFFICIAL ?= no
 NODE ?= node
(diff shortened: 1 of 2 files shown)


shanemcd (approved):
🍰

jjwatt:
Hah. This is a fun one. So, whatever shell (version of `sh`) that GitHub runs did not like my trick with `command -v python3 python3.9` It works it a lot of shells because it will return the first full path that actually exists. However, apparently the shell that runs from GH Actions -> ci.yml -> make -> ansible -> build.yml -> make print-VERSION is one that does not work that way! Instead it returns blank/empty if the first command can't be found :). So, that was causing the `Permission Denied` error I was getting in the `awx-operator` CI.

I fixed it by using a for loop in the shell macro instead of relying on my assumed `command -v` behavior. I still use `command -v` but in a loop, and it won't error out on missing commands.

I will clean this up and squash.

TheRealHaoLiu (approved):

TheRealHaoLiu:
related to https://github.com/ansible/awx/issues/13737

PR #13903, opened 2023-04-25
Enhance collection intergration tests

john-westcott-iv:
##### SUMMARY

Update all tests to respect and use test_id to allow for running all tests as a single test.
Fix timeouts in ad_hoc_command_cancel since the API now returns sub second to a cancel call (timeout really no longer applies)
Fix check in group test as it sets up group2 as a sub group of group1 so when we delete group1 group2 is deleted by default. The UI now enables two options: delete sub groups and promote sub groups so maybe we want to support that in the collection someday.
Only run instance test if we are on a k8s installation (and assume we are not)
Remove the manual_project integration test since we can no longer auto-create the project directory because we now execute jobs inside EEs.

##### ISSUE TYPE

##### COMPONENT NAME

 - Collection

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```


awx_collection/tests/integration/targets/group/tasks/main.yml
```diff
@@ -1,13 +1,18 @@
...
-    host_name1: "AWX-Collection-tests-group-host-{{ lookup('password', '/dev/null chars=ascii_letters length=16') }}"
-    host_name2: "AWX-Collection-tests-group-host-{{ lookup('password', '/dev/null chars=ascii_letters length=16') }}"
-    host_name3: "AWX-Collection-tests-group-host-{{ lookup('password', '/dev/null chars=ascii_letters length=16') }}"
+    group_name1: "AWX-Collection-tests-group-group1-{{ test_id }}"
+    group_name2: "AWX-Collection-tests-group-group2-{{ test_id }}"
+    group_name3: "AWX-Collection-tests-group-group3-{{ test_id }}"
+    inv_name: "AWX-Collection-tests-group-inv-{{ test_id }}"
+    host_name1: "AWX-Collection-tests-group-host1-{{ test_id }}"
+    host_name2: "AWX-Collection-tests-group-host2-{{ test_id }}"
+    host_name3: "AWX-Collection-tests-group-host3-{{ test_id }}"
```
AlanCoding, line 15: oh this is much better

awx_collection/tests/integration/targets/group/tasks/main.yml
```diff
@@ -117,9 +122,10 @@
     state: absent
   register: result
 
+# In this case, group 2 was last a child of group1 so deleting group1 deleted group2
 - assert:
     that:
-      - "result is changed"
+      - "result is not changed"
```
AlanCoding, line 128: Confirming - this was a preexisting failure.

john-westcott-iv: I think it used to work at one point but if you try this scenario in the UI now you get two options:
* delete the child group(s) (default)
* promote the child group(s)

The code in the group module follows the first bullet now. So when we delete group1 first group2 is now gone.

AlanCoding (approved):
Great work, these are sorely needed overdue changes.

PR #13904, opened 2023-04-25
Hop nodes for k8s

fosterseth:
##### SUMMARY

Allow users to setup hop nodes for k8s deployments

- [ ] revert using devel receptor-collection before merging

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 22.0.1
```


awx/api/serializers.py
```diff
@@ -5465,22 +5479,29 @@ def get_percent_capacity_remaining(self, obj):
...
+
+        if not self.instance and not settings.IS_K8S:
+            raise serializers.ValidationError("Can only create instances on Kubernetes or OpenShift.")
+        node_type = get_field_from_model_or_attrs("node_type")
+        peers_from_control_nodes = get_field_from_model_or_attrs("peers_from_control_nodes")
+
+        if peers_from_control_nodes and node_type not in (Instance.Types.EXECUTION, Instance.Types.HOP):
+            raise serializers.ValidationError("peers_from_control_nodes can only be enabled for execution or hop nodes.")
+
+        if node_type in (Instance.Types.CONTROL):
```
relrod: I would write this as: `if node_type == Instance.Types.CONTROL:`

But not just to be nit-picky.

The line in its current form happens to work, but I don't think it's obvious/intentional how it's working.

This is a peculiarity of Python syntax when defining a tuple with just one element. You need a trailing comma. To someone quickly skimming the code, it _looks_ like you're making a tuple here, but without a trailing comma, you aren't:

```python
>>> a = (Instance.Types.CONTROL)
>>> b = (Instance.Types.CONTROL,)
>>> type(a)
<enum 'Types'>
>>> type(b)
<class 'tuple'>
```

What you're actually checking is if `node_type` is in the tuple that `Instance.Types.CONTROL` is defined as in `models/ha.py` (which gets turned into an enum entry, where the entry itself is a tuple):

```python
    class Types(models.TextChoices):
        CONTROL = 'control', _("Control plane node")
```

So the line happens to work because `"control"` is in that tuple/enum entry. But I don't think that was the intention here when the code was written, I think you wanted to compare against a tuple that you were defining. Which is fine if you are planning to possibly have more entries here in the future (in which case, my request is that you add the comma to the end). But if you're never planning to check against more than one instance type in this conditional, `==` is more clear, and works as expected:

```python
>>> 'control' == Instance.Types.CONTROL
True
>>> 'hop' == Instance.Types.CONTROL
False
```

relrod (changes requested):

fosterseth:
reminder: use string translations on the validationerrors in the serializers

awx/api/serializers.py
fosterseth: good catch and great explanation. Will address this

awx/api/views/__init__.py
```diff
@@ -4350,3 +4351,17 @@ def post(self, request, *args, **kwargs):
...
+
+
+class PeersList(ListAPIView):
+    name = _("Peers")
+    model = models.InstanceLink
+    serializer_class = serializers.InstanceLinkSerializer
+    search_fields = ('source', 'target', 'link_state')
+
+
+class PeersDetail(RetrieveAPIView):
```
AlanCoding: The plural of "Peers" is kind of weird here - this is showing a single InstanceLink right?

awx/api/views/root.py
```diff
@@ -130,6 +130,7 @@ def get(self, request, format=None):
         data['mesh_visualizer'] = reverse('api:mesh_visualizer_view', request=request)
         data['bulk'] = reverse('api:bulk', request=request)
         data['analytics'] = reverse('api:analytics_root_view', request=request)
+        data['peers'] = reverse('api:peers_list', request=request)
```
AlanCoding: this is kind of vague to be at /api/v2/, maybe "instance_peers" or "instance_links" would be better

awx/main/migrations/0186_hop_nodes.py
```diff
@@ -0,0 +1,53 @@
+# Generated by Django 4.2 on 2023-05-17 18:31
+
+import django.core.validators
+from django.db import migrations, models
+
+
+def set_peers_from_control_nodes_true(apps, schema_editor):
+    Instance = apps.get_model('main', 'Instance')
+    Instance.objects.filter(node_type='execution').update(peers_from_control_nodes=True)
```
AlanCoding: should this still be done if it's not a K8S install?

AlanCoding: never mind my comment, that's already being worked on in https://github.com/ansible/awx/pull/14196

AlanCoding:
I think the api-test CI failure could be potentially because this branch needs to be rebased.

awx/api/views/root.py
fosterseth: this endpoint isn't used, I think we will drop it altogether

AlanCoding:
in operator check:

```
  File \"/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/jobs.py\", line 597, in run
    res = receptor_job.run()
  File \"/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/receptor.py\", line 318, in run
  File \"/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/receptor.py\", line 393, in _run_internal
    res = processor_future.result()
  File \"/usr/lib64/python3.9/concurrent/futures/_base.py\", line 446, in result
    return self.__get_result()
  File \"/usr/lib64/python3.9/concurrent/futures/_base.py\", line 391, in __get_result
    raise self._exception
  File \"/usr/lib64/python3.9/concurrent/futures/thread.py\", line 58, in run
  File \"/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/utils/common.py\", line 1189, in wrapper_cleanup_new_process
    return func(*args, **kwargs)
  File \"/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/receptor.py\", line 461, in processor
    return ansible_runner.interface.run(
  File \"/var/lib/awx/venv/awx/lib64/python3.9/site-packages/ansible_runner/interface.py\", line 210, in run
    r = init_runner(**kwargs)
  File \"/var/lib/awx/venv/awx/lib64/python3.9/site-packages/ansible_runner/interface.py\", line 107, in init_runner
    stream_processor = Processor(event_handler=event_callback_handler,
  File \"/var/lib/awx/venv/awx/lib64/python3.9/site-packages/ansible_runner/streaming.py\", line 270, in __init__
    self.artifact_dir = os.path.join(project_artifacts, ident)
  File \"/usr/lib64/python3.9/posixpath.py\", line 90, in join
    genericpath._check_arg_types('join', a, *p)
  File \"/usr/lib64/python3.9/genericpath.py\", line 152, in _check_arg_types
    raise TypeError(f'{funcname}() argument must be str, bytes, or '
TypeError: join() argument must be str, bytes, or os.PathLike object, not 'int'
```

This is fixed with https://github.com/ansible/ansible-runner/pull/1268

awx/main/models/ha.py
```diff
@@ -493,6 +536,13 @@ def on_instance_group_deleted(sender, instance, using, **kwargs):
 @receiver(post_delete, sender=Instance)
 def on_instance_deleted(sender, instance, using, **kwargs):
     schedule_policy_task()
+    if settings.IS_K8S and instance.node_type in (Instance.Types.EXECUTION, Instance.Types.HOP) and instance.peers_from_control_nodes:
+        schedule_write_receptor_config()
+
+
+@receiver(post_save, sender=InstanceLink)
+def on_instancelink_save(sender, instance, using, **kwargs):
+    logger.warning("instancelink saved")
```
fosterseth: remove this on_instancelink_save

awx/main/migrations/0186_hop_nodes.py
```diff
@@ -0,0 +1,61 @@
...
+        ),
+        migrations.AlterField(
+            model_name='instance',
+            name='listener_port',
+            field=models.PositiveIntegerField(
+                blank=True,
+                default=None,
+                help_text='Port that Receptor will listen for incoming connections on.',
+                null=True,
+                validators=[django.core.validators.MinValueValidator(1), django.core.validators.MaxValueValidator(65535)],
```
fosterseth: min value 1024

awx/main/tasks/system.py
```diff
@@ -459,6 +459,30 @@ def execution_node_health_check(node):
...
+    ctl = get_receptor_ctl()
+    mesh_status = ctl.simple_command('status')
+
+    # detect active/inactive receptor links
+    from awx.main.models import InstanceLink
+
+    all_links = InstanceLink.objects.all()
+    active_receptor_conns = mesh_status['KnownConnectionCosts']
+    update_links = []
+    for link in all_links:
```
fosterseth, line 528: ignore nodes that are in a deprovisioning state

awx/main/tasks/system.py
```diff
@@ -459,6 +459,30 @@ def execution_node_health_check(node):
...
+        if link.target.hostname in active_receptor_conns.get(link.source.hostname, {}):
+            if link.link_state is not InstanceLink.States.ESTABLISHED:
+                link.link_state = InstanceLink.States.ESTABLISHED
+                update_links.append(link)
+        else:
+            if link.link_state is not InstanceLink.States.DISCONNECTED:
```
fosterseth: ignore nodes in adding state

awx/main/models/ha.py
```diff
@@ -464,21 +478,50 @@ def on_instance_group_saved(sender, instance, created=False, raw=False, **kwargs
...
-
-        if instance.node_state == Instance.States.INSTALLED:
-            from awx.main.tasks.receptor import write_receptor_config  # prevents circular import
-
-            # broadcast to all control instances to update their receptor configs
-            connection.on_commit(lambda: write_receptor_config.apply_async(queue='tower_broadcast_all'))
+        else:
+            control_instances = set(Instance.objects.filter(node_type__in=[Instance.Types.CONTROL, Instance.Types.HYBRID]))
+            if instance.peers_from_control_nodes:
+                if (control_instances & set(instance.peers_from.all())) != set(control_instances):
```
fosterseth, line 520: @jbradberry please double check my conditional here

awx/main/models/ha.py
```diff
@@ -464,21 +478,50 @@ def on_instance_group_saved(sender, instance, created=False, raw=False, **kwargs
...
+                    instance.peers_from.add(*control_instances)
+                    schedule_write_receptor_config()  # keep method separate to make pytest mocking easier
+            else:
+                if set(control_instances) & set(instance.peers_from.all()):
```
fosterseth, line 524: remove set

awx/api/serializers.py
```diff
@@ -5368,6 +5374,11 @@ class Meta:
         fields = ('id', 'hostname', 'node_type', 'node_state', 'enabled')
 
 
+class PeersSerializer(serializers.StringRelatedField):
+    def to_internal_value(self, value):
+        return Instance.objects.get(hostname=value)
```
jbradberry: I'm curious why this was done, instead of making `InstanceSerializer.peers` something like the `source` and `target` fields on `InstanceLinkSerializer`.  It seems to me like doing it this way will result in N+1 query behavior.

awx/api/serializers.py
```diff
@@ -5376,10 +5387,11 @@ class InstanceSerializer(BaseSerializer):
     jobs_running = serializers.IntegerField(help_text=_('Count of jobs in the running or waiting state that are targeted for this instance'), read_only=True)
     jobs_total = serializers.IntegerField(help_text=_('Count of all jobs that target this instance'), read_only=True)
     health_check_pending = serializers.SerializerMethodField()
+    peers = PeersSerializer(many=True, required=False)
 
     class Meta:
         model = Instance
-        read_only_fields = ('ip_address', 'uuid', 'version')
```
jbradberry: We're allowing `ip_address` to be changed?  Doesn't that have the same problem as changing the `port` value?  A bunch of instances pointing to it will have to have their configs updated.

awx/api/serializers.py
```diff
@@ -5464,37 +5478,62 @@ def get_percent_capacity_remaining(self, obj):
...
+        listener_port = get_field_from_model_or_attrs("listener_port")
+        ip_address = get_field_from_model_or_attrs("ip_address")
+        hostname = get_field_from_model_or_attrs("hostname")
+        if not ip_address:
+            attrs["ip_address"] = hostname
+
+        if peers_from_control_nodes and node_type not in (Instance.Types.EXECUTION, Instance.Types.HOP):
+            raise serializers.ValidationError(_("peers_from_control_nodes can only be enabled for execution or hop nodes."))
+
+        if node_type == Instance.Types.CONTROL:
```
jbradberry: What about hybrid?  Those are also on the control plane, and presumably would disable this field for the same reason.

awx/api/serializers.py
```diff
@@ -5464,37 +5478,62 @@ def get_percent_capacity_remaining(self, obj):
...
 
     def validate_node_type(self, value):
-        if not self.instance:
-            if value not in (Instance.Types.EXECUTION,):
-                raise serializers.ValidationError("Can only create execution nodes.")
-        else:
-            if self.instance.node_type != value:
-                raise serializers.ValidationError("Cannot change node type.")
+        if not self.instance and value not in [Instance.Types.HOP, Instance.Types.EXECUTION]:
+            raise serializers.ValidationError("Can only create execution nodes.")
```
jbradberry: Wording.

awx/api/templates/instance_install_bundle/requirements.yml
```diff
@@ -1,4 +1,5 @@
 ---
 collections:
-  - name: ansible.receptor
-    version: 1.1.0
+  - name: https://github.com/fosterseth/receptor-collection.git
```
jbradberry: :raised_eyebrow:

awx_collection/plugins/modules/instance.py
```diff
@@ -122,6 +138,12 @@ def main():
         new_fields['node_state'] = node_state
     if listener_port is not None:
         new_fields['listener_port'] = listener_port
+    if peers is not None:
+        new_fields['peers'] = peers
+    if peers is None:
+        peers = ['']
```
jbradberry: Surely the empty list?

awx/main/tests/functional/api/test_instance_peers.py
```diff
@@ -0,0 +1,331 @@
...
+
+    def test_prevent_peering_to_self(self):
+        '''
+        cannot peer to self
+        '''
+        control_instance = Instance.objects.create(hostname='abc', node_type="control")
+        with pytest.raises(IntegrityError):
+            control_instance.peers.add(control_instance)
+
+    @pytest.mark.parametrize('node_type', ['control', 'hop', 'execution'])
```
jbradberry: hybrid

awx/main/tests/functional/api/test_instance_peers.py
```diff
@@ -0,0 +1,331 @@
...
+        listener_port can be None
+        '''
+        post(
+            url=reverse('api:instance_list'),
+            data={"hostname": "abc", "node_type": node_type, "listener_port": None},
+            user=admin_user,
+            expect=201,
+        )
+
+    @pytest.mark.parametrize('node_type, allowed', [('control', False), ('hop', True), ('execution', True)])
```
jbradberry: hybrid

awx/main/tests/functional/api/test_instance_peers.py
```diff
@@ -0,0 +1,331 @@
...
+            user=admin_user,
+            expect=400,
+        )
+
+    def test_peers_from_control_nodes_listener_port_enabled(self, admin_user, post):
+        '''
+        if peers_from_control_nodes is True, listener_port must an integer
+        Assert that all other combinations are allowed
+        '''
+        Instance.objects.create(hostname='abc', node_type="control")
```
jbradberry: I'd also toss in a (you guessed it) hybrid node, just to make sure that both types do the autolink thing.

awx/main/tests/functional/api/test_instance_peers.py
```diff
@@ -0,0 +1,331 @@
...
+                        expect=400 if disallowed else 201,
+                    )
+                    i += 1
+
+    def test_disallow_modify_peers_control_nodes(self, admin_user, patch):
+        '''
+        for control nodes, peers field should not be
+        modified directly via patch.
+        '''
+        control = Instance.objects.create(hostname='abc', node_type='control')
```
jbradberry: pretty much take as given that it would be good to cover hybrid nodes everywhere else in this module unless it explicitly doesn't make sense (but I can't think of circumstances where that would happen).

awx/api/templates/instance_install_bundle/group_vars/all.yml
```diff
@@ -14,10 +15,22 @@ custom_tls_certfile: receptor/tls/receptor.crt
...
+{% if peers %}
+receptor_peers:
+{% for peer in peers %}
+  - host: {{ peer.host }}
+    port: {{ peer.port }}
+    protocol: tcp
+{% endfor %}
+{% endif %}
+additional_python_packages:
+  - ansible-runner
```
jbradberry: As I mentioned in the review meeting, this should not be installed on hop nodes.

awx/api/urls/urls.py
```diff
@@ -153,6 +154,7 @@
     re_path(r'^bulk/$', BulkView.as_view(), name='bulk'),
     re_path(r'^bulk/host_create/$', BulkHostCreateView.as_view(), name='bulk_host_create'),
     re_path(r'^bulk/job_launch/$', BulkJobLaunchView.as_view(), name='bulk_job_launch'),
+    re_path(r'^peers/', include(peers_urls)),
```
jbradberry: I don't understand why we are doing top-level 'peers' endpoints.  Peers implies that we are looking in the context of some particular instance, yeah?

Maybe as I get further down I'll see why and this comment will just recommend changing the name to 'links' or something.

awx/api/views/__init__.py
jbradberry: I agree with @AlanCoding (with the additional question in my mind of why they exist)

awx/api/views/root.py
jbradberry: @fosterseth yeah, I think dropping it is the right thing.

awx/main/access.py
```diff
@@ -2949,6 +2950,22 @@ def filtered_queryset(self):
         return self.model.objects.filter(workflowjobtemplatenodes__workflow_job_template__in=WorkflowJobTemplate.accessible_pk_qs(self.user, 'read_role'))
 
 
+class PeersAccess(BaseAccess):
```
jbradberry: This should presumably go away along with the peers endpoint

awx/main/models/ha.py
```diff
@@ -67,19 +68,28 @@ class InstanceLink(BaseModel):
     class States(models.TextChoices):
         ADDING = 'adding', _('Adding')
         ESTABLISHED = 'established', _('Established')
+        DISCONNECTED = 'disconnected', _('Disconnected')
         REMOVING = 'removing', _('Removing')
 
     link_state = models.CharField(
-        choices=States.choices, default=States.ESTABLISHED, max_length=16, help_text=_("Indicates the current life cycle stage of this peer link.")
+        choices=States.choices, default=States.ADDING, max_length=16, help_text=_("Indicates the current life cycle stage of this peer link.")
```
jbradberry, line 74: I think this is a mistake, bound up in a misapprehension of what the modeling of the links is there to accomplish, along with several other things I see in this PR.  I'll try to write up my thoughts on this as a top-level comment.

awx/main/models/ha.py
```diff
@@ -67,19 +68,28 @@ class InstanceLink(BaseModel):
     class States(models.TextChoices):
         ADDING = 'adding', _('Adding')
         ESTABLISHED = 'established', _('Established')
+        DISCONNECTED = 'disconnected', _('Disconnected')
```
jbradberry: and this.

awx/main/models/ha.py
```diff
@@ -67,19 +68,28 @@ class InstanceLink(BaseModel):
...
 
     link_state = models.CharField(
-        choices=States.choices, default=States.ESTABLISHED, max_length=16, help_text=_("Indicates the current life cycle stage of this peer link.")
+        choices=States.choices, default=States.ADDING, max_length=16, help_text=_("Indicates the current life cycle stage of this peer link.")
     )
 
     class Meta:
         unique_together = ('source', 'target')
+        ordering = ("id",)
+        constraints = [models.CheckConstraint(check=~models.Q(source=models.F('target')), name='source_and_target_can_not_be_equal')]
```
jbradberry, line 80: :+1:

awx/api/views/instance_install_bundle.py
```diff
@@ -49,9 +53,9 @@ class InstanceInstallBundle(GenericAPIView):
     def get(self, request, *args, **kwargs):
         instance_obj = self.get_object()
 
-        if instance_obj.node_type not in ('execution',):
+        if instance_obj.node_type not in ('execution', 'hop'):
```
jbradberry, line 56: it would be good to make use of the enum objects instead of bare strings, like we do elsewhere

awx/main/access.py
AlanCoding: Right now I'm not seeing any views which are not read-only.

awx/api/views/root.py
fosterseth: removed this endpoint

awx/api/serializers.py
fosterseth: now includes hop

awx_collection/plugins/modules/instance.py
fosterseth: fixed

awx/main/tests/functional/api/test_instance_peers.py
fosterseth: the purpose of this test is not autolinking, rather it makes sure listener port is defined if peers_from_control_nodes is enabled. in fact the control node creation is not need so I'm going to remove that line.

awx/api/urls/urls.py
fosterseth: dropped this endpoint

awx/main/access.py
fosterseth: dropping this

awx/main/models/ha.py
fosterseth: Jeff and I discussed and decided to keep ADDING as the default.

awx/api/views/instance_install_bundle.py
fosterseth: - [ ] use enum where possible

kurokobo:
Created small PR to fix docs for this branch: https://github.com/ansible/awx/pull/14248

awx/main/tasks/system.py
jbradberry: This all isn't very atomic, leaving the possibility of changes getting clobbered.

awx/main/tasks/receptor.py
```diff
@@ -731,6 +769,3 @@ def remove_deprovisioned_node(hostname):
 
     # This will as a side effect also delete the InstanceLinks that are tied to it.
     Instance.objects.filter(hostname=hostname).delete()
-
-    # Update the receptor configs for all of the control-plane.
-    write_receptor_config.apply_async(queue='tower_broadcast_all')
```
jbradberry, line 736: Why is this being taken out?  Nodes that previously linked to the decommissioned node will need to be updated.

awx/main/tests/functional/api/test_instance_peers.py
```diff
@@ -0,0 +1,368 @@
...
+
+@pytest.mark.django_db
+class TestPeers:
+    @pytest.fixture(autouse=True)
+    def configure_settings(self, settings):
+        settings.IS_K8S = True
+
+    @pytest.mark.parametrize('node_type', ['control', 'hybrid'])
+    def test_prevent_peering_to_self(self, node_type):
+        '''
```
jbradberry: this is picky, I know, but we should get away from using triple-single quotes

awx/main/tests/functional/api/test_instance_peers.py
```diff
@@ -0,0 +1,368 @@
...
+                for listener_port in [None, 6789]:
+                    # only disallowed case is when peers_from is True and listener port is None
+                    disallowed = peers_from and not listener_port
+                    post(
+                        url=reverse('api:instance_list'),
+                        data={"hostname": f"abc{i}", "peers_from_control_nodes": peers_from, "node_type": node_type, "listener_port": listener_port},
+                        user=admin_user,
+                        expect=400 if disallowed else 201,
+                    )
+                    i += 1
```
jbradberry: I would have used `itertools.product` and `enumerate` to avoid manually managing the index number.

awx/main/tests/functional/api/test_instance_peers.py
```diff
@@ -0,0 +1,368 @@
...
+        peer to hop
+
+        peer to hop should be removed if hop is deleted
+        '''
+
+        hop = Instance.objects.create(hostname='hop', node_type='hop', peers_from_control_nodes=True, listener_port=6789)
+        control = Instance.objects.create(hostname='abc', node_type=node_type)
+        assert hop in control.peers.all()
+        hop.delete()
+        assert not control.peers.exists()
```
jbradberry, line 198: if this test doesn't work, it'll be because `.exists()` uses the queryset results cache populated by the `.all()` above.

awx/main/tests/functional/api/test_instance_peers.py
fosterseth: good suggestion

awx/api/templates/instance_install_bundle/install_receptor.yml
```diff
@@ -7,14 +7,11 @@
       user:
         name: "{{ receptor_user }}"
         shell: /bin/bash
-    - name: Enable Copr repo for Receptor
-      command: dnf copr enable ansible-awx/receptor -y
```
janorn, line 11: Shouldn't this be removed from previous installed receptors?

fosterseth: sorry, can you clarify the question? The next receptor-collection version will install receptor via these releases by default https://github.com/ansible/receptor/releases

thus the Copr repo is no longer needed going forward

janorn: I'm thinking of upgrades of previous installs.

awx/api/templates/instance_install_bundle/requirements.yml
```diff
@@ -1,4 +1,5 @@
 ---
 collections:
-  - name: ansible.receptor
-    version: 1.1.0
+  - name: https://github.com/ansible/receptor-collection.git
+    type: git
+    version: main
```
CFSNM: do we need to point this to 2.0.0 version?

CFSNM (approved):

djyasin (approved):

awx/api/views/instance_install_bundle.py
AlanCoding: Yeah, I think this is still an open item but only a minor code style thing.

tools/docker-compose/bootstrap_development.sh
```diff
@@ -40,7 +40,7 @@ echo "Admin password: ${DJANGO_SUPERUSER_PASSWORD}"
 awx-manage create_preload_data
 awx-manage register_default_execution_environments
 
-awx-manage provision_instance --hostname="$(hostname)" --node_type="$MAIN_NODE_TYPE"
+awx-manage provision_instance --hostname="$(hostname)" --node_type="$MAIN_NODE_TYPE" --listener_port=2222
```
AlanCoding, line 43: This concerns me a little bit about the expected practice. Does this need to be provided for VM installs in general? I don't get why we didn't need this before, but would now.

AlanCoding: So it seems that this is not required, and as such, it would make the most sense to _not_ set it here, because the VM install scenario that this corresponds to should not ordinarily be setting the listener port.

AlanCoding (approved):

AlanCoding:
Filed a followup issue https://github.com/ansible/awx/issues/14387

relrod (approved):

akus062381 (approved):

PR #13921, opened 2023-04-27
[troubleshooting][WIP] Add debug management command to stream pg_notify traffic

AlanCoding:
##### SUMMARY
Run this command while other stuff happens to see messages. Still needs more polish.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/dispatch/queues.py
+++ b/awx/main/dispatch/queues.py
@@ -1,0 +1,42 @@
+import os
+
+from django.conf import settings
+
+
+REDIS_GROUP = "tower_settings_change"
+RSYSLOG_GROUP = "rsyslog_configurer"
+TASK_GROUP = "tower_broadcast_all"
+HEARTBEET = "web_heartbeet"
+
+
+def get_local_queuename():
+    return settings.CLUSTER_HOST_ID
+
+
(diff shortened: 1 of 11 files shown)


awx/main/dispatch/queues.py
```diff
@@ -0,0 +1,42 @@
+import os
+
+from django.conf import settings
+
+
+REDIS_GROUP = "tower_settings_change"
+RSYSLOG_GROUP = "rsyslog_configurer"
+TASK_GROUP = "tower_broadcast_all"
+HEARTBEET = "web_heartbeet"
```
AlanCoding, line 9: These constants aren't faithfully used yet. Like the heartbeet should get applied:

https://github.com/ansible/awx/blob/301f3f25386d8b1c63d9fadcf54ab5fac9b04d1f/awx/main/management/commands/run_heartbeet.py#L64

and

https://github.com/ansible/awx/blob/fa05f555123b968f36cfd38e5a5f37d6ed786b7d/awx/main/wsrelay.py#L214

AlanCoding:
Going to table this for potentially revisiting later, but closing at the moment.

PR #13923, opened 2023-04-27
wsrelay: handle heartbeet shutdown and redis drop

relrod:
##### SUMMARY

This fixes two different exceptions in wsrelay.

- One resulted from heartbeet getting ability in #13858 to gracefully shut down. When we saw the message come through, we didn't fully clean up the connection to the web node.

- The second resulted when Redis disappeared. We still want to exit in that case, but it's better to log a message and exit gracefully instead of crashing out.

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/wsrelay.py
+++ b/awx/main/wsrelay.py
@@ -5,6 +5,7 @@ from typing import Dict
 
 import aiohttp
 from aiohttp import client_exceptions
+import aioredis
 
 from channels.layers import get_channel_layer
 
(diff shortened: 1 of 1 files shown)


AlanCoding (approved):
It would be good to have some more background information - like the traceback this addresses, any associated testing.

But the code changes look well thought-out and defensive.

relrod:
Superseded by #13991

PR #13924, opened 2023-04-27
Enhance the dashboard job summary endpoint

jamestalton:
##### SUMMARY
Enhance the dashboard job summary endpoint to contain `canceled` and `error` job counts.

Jobs have 4 completion states `successful`, `failed`, `canceled`, and `error`. 

The dashboard api endpoint only gives `successful` and `failed`, which makes the user think there is something wrong with the dashboard because jobs with an `error` status do not show up. This PR adds those counts to the dashboard API.

We see this on the new dashboard, where the recent jobs card shows errors but the chart shows none.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API
 
 This PR: https://github.com/ansible/ansible-ui/pull/477
Adds support to the Ansible-UI to use the updated API payload.

This is an example screenshot of what the customer sees without this PR. Notice no red in the graph. :(
Screenshot 2023-04-27 at 8 59 32 AM

### DIFF
--- a/awx/api/views/__init__.py
+++ b/awx/api/views/__init__.py
@@ -272,16 +272,24 @@ class DashboardJobsGraphView(APIView):
 
         success_query = user_unified_jobs.filter(status='successful')
         failed_query = user_unified_jobs.filter(status='failed')
+        canceled_query = user_unified_jobs.filter(status='canceled')
+        error_query = user_unified_jobs.filter(status='error')
 
         if job_type == 'inv_sync':
             success_query = success_query.filter(instance_of=models.InventoryUpdate)
             failed_query = failed_query.filter(instance_of=models.InventoryUpdate)
+            canceled_query = canceled_query.filter(instance_of=models.InventoryUpdate)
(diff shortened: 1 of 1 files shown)


jamestalton:
> **jmack slack comment**
Looks good. The query filtering and data extraction operations are the same code repeated so that could probably be consolidated since were doing it four times now

jamestalton:
> **from alancoding in slack**
Strictly from the API side, we should expand tests to check the new fields you're adding, we probably already have some dashboard tests for counts, so I think it's unlikely this would be new tests, just expanding what's there
well I'm coming up with a surprising shortage of dashboard tests...
Since this gives global counts, I think the best format for the test would be an API unit test for the view.

shanemcd:
Let's make sure those 3 merge commits don't make their way into devel. Either rebase this manually prior to, or choose the squash option when clicking the merge button.

dmzoneill (approved):

PR #13928, opened 2023-04-30
In collection, give changed status in workflow_job_template when destroying nodes

### LINKED ISSUE #8349: Explore ways to make modules project_update and inventory_source_update idempotent

### LINKED ISSUE #13387: awx.awx.workflow_job_template parameter destroy_current_nodes does not show "changed" status

### Bug Summary

When using the awx.awx.workflow_job_template module and the parameter 'destroy_current_nodes: true', Ansible returns 'ok' status even though the item was very much changed as all of the workflow nodes are removed.

This should return a 'changed' status to properly indicate that something was modified on the workflow.

### AWX version

21.7.0

- [ ] UI
- [ ] API
- [ ] Docs
- [X] Collection
- [ ] CLI
- [ ] Other

### Installation method

kubernetes

### Modifications

no

### Ansible version

2.13.7

### Operating system

_No response_

### Web browser

_No response_

### Steps to reproduce

Task using awx.awx.workflow_job_template
Set parameter 'destroy_current_nodes: true'

### Expected results

Status would be 'changed' since all existing nodes in the workflow were removed

### Actual results

Status is 'ok' and all existing nodes in the workflow were removed

### Additional information

_No response_

sean-m-sullivan:
##### SUMMARY
makes destroy current nodes schema report changed status correctly
Resolves #13387

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - Collection

##### AWX VERSION

```
22.1.0
```

### DIFF
--- a/awx_collection/plugins/modules/workflow_job_template.py
+++ b/awx_collection/plugins/modules/workflow_job_template.py
@@ -991,6 +991,7 @@ def main():
     # Destroy current nodes if selected.
     if destroy_current_nodes:
         destroy_workflow_nodes(module, response, workflow_job_template_id)
+        module.json_output['changed'] = True
 
     # Work thorugh and lookup value for schema fields
     if workflow_nodes:


AlanCoding:
This doesn't look coherent for the case where it is _not_ an SCM inventory import. Those inventory updates don't use `scm_revision` at all, and thus, it will probably be blank. That suggests to me that this would have ec2 updates, for example, always give changed=False which isn't desirable.

Even for SCM inventory updates, the revision is only valid as a metric for changed status if the content it static, which is often isn't. Cloud sources can change.

sean-m-sullivan:
@AlanCoding I did not take that into account you are correct, I've renamed the PR, and removed the inventory source update code, and just left the workflow node change.

AlanCoding (approved):

PR #13930, opened 2023-04-30
In collection, allow roles to be added to multiple teams and users

sean-m-sullivan:
##### SUMMARY
collection, allow roles to be added to multiple teams and users

Have had multiple people saw when assigning roles to teams and users they have to have the same long list of JT/WFT declared multiple times one for each team and user to receive that permission. This is a way to do that in bulk. 

This also simplifies the code a bit as well removing an entire block, I also searched and after the changes the function param_to_endpoint, was no longer needed as well.

Also should we retire the other singular parameters now, they've been depreciated for a while, or create another issue around that.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - Collection

##### AWX VERSION
```
22.1.0
```

### DIFF
--- a/awx_collection/plugins/modules/role.py
+++ b/awx_collection/plugins/modules/role.py
@@ -24,11 +24,23 @@ options:
     user:
       description:
         - User that receives the permissions specified by the role.
+        - Deprecated, use 'users'.
       type: str
+    users:
+      description:
+        - Users that receive the permissions specified by the role.
+      type: list
+      elements: str
     team:
       description:
         - Team that receives the permissions specified by the role.
+        - Deprecated, use 'teams'.
       type: str
+    teams:
+      description:
+        - Teams that receive the permissions specified by the role.
+      type: list
+      elements: str
     role:
       description:
         - The role type to grant/revoke.
@@ -161,7 +173,9 @@ def main():
 
     argument_spec = dict(
         user=dict(),
+        users=dict(type='list', elements='str'),
         team=dict(),
+        teams=dict(type='list', elements='str'),
         role=dict(
             choices=[
                 "admin",
@@ -219,9 +233,9 @@ def main():
         'projects': 'project',
         'target_teams': 'target_team',
         'workflows': 'workflow',
+        'users': 'user',
+        'teams': 'team',
     }
-    # Singular parameters
-    resource_param_keys = ('user', 'team', 'lookup_organization')
 
     resources = {}
     for resource_group, old_name in resource_list_param_keys.items():
@@ -229,9 +243,9 @@ def main():
             resources.setdefault(resource_group, []).extend(module.params.get(resource_group))
         if module.params.get(old_name) is not None:
             resources.setdefault(resource_group, []).append(module.params.get(old_name))
-    for resource_group in resource_param_keys:
-        if module.params.get(resource_group) is not None:
-            resources[resource_group] = module.params.get(resource_group)
+    if module.params.get('lookup_organization') is not None:
+        resources['lookup_organization'] = module.params.get('lookup_organization')
+
     # Change workflows and target_teams key to its endpoint name.
     if 'workflows' in resources:
         resources['workflow_job_templates'] = resources.pop('workflows')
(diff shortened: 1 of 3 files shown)


awx_collection/plugins/modules/role.py
```diff
@@ -248,43 +262,32 @@ def main():
...
-                lookup_data_populated = lookup_data
-            # Attempt to look up project based on the provided name or ID and lookup data
-            data = module.get_one('{0}s'.format(key), name_or_id=resources[key], data=lookup_data_populated)
-            if data is None:
-                module.fail_json(
-                    msg='Unable to find {0} with name: {1}'.format(key, resources[key]), changed=False
-                )
-            else:
-                actor_data[key] = module.get_one('{0}s'.format(key), name_or_id=resources[key], data=lookup_data_populated)
-                resources.pop(key)
```
AlanCoding, line 265: How is this supporting the now-deprecated singular parameters of user and team?

Shouldn't you need to get those options and then add then onto the users / teams options in order for the current use to continue working?

sean-m-sullivan: Line 236/246 in the new code Takes the individual user/team params and adds them on to the List items to process. Same as we did for the older items like credentials and such

sean-m-sullivan: and to add on to this, because we changed to lists and adding them to resource items, this logic was also moved to 279. So taken together 236/246/279 in the new code works together to eliminate this block.

AlanCoding (approved):

AlanCoding:
@john-westcott-iv if you want to sync quickly about this PR tomorrow I can fill in any gaps, and then I'd opt to go ahead with merge.

awx_collection/plugins/modules/role.py
```diff
@@ -24,11 +24,23 @@
...
+      type: list
+      elements: str
     team:
       description:
         - Team that receives the permissions specified by the role.
+        - Deprecated, use 'teams'.
       type: str
+    teams:
+      description:
+        - Teams that receives the permissions specified by the role.
```
ericzolf:
```suggestion
        - Teams that receive the permissions specified by the role.
```

awx_collection/plugins/modules/role.py
```diff
@@ -24,11 +24,23 @@
     user:
       description:
         - User that receives the permissions specified by the role.
+        - Deprecated, use 'users'.
       type: str
+    users:
+      description:
+        - Users that receives the permissions specified by the role.
```
ericzolf:
```suggestion
        - Users that receive the permissions specified by the role.
```

awx_collection/tests/integration/targets/role/tasks/main.yml
```diff
@@ -55,9 +69,12 @@
         that:
           - "result is changed"
 
-    - name: Add Joe to the update role of the default Project with lookup Organization
+    - name: Add Joe and teams to the update role of the default Project with lookup Organization
       role:
         user: "{{ username }}"
```
ericzolf, line 88: Should we still use `user` (and `team`) in tests if they're now deprecated?

awx_collection/tests/integration/targets/role/tasks/main.yml
```diff
@@ -74,6 +91,9 @@
     - name: Add Joe to the new project by ID
       role:
         user: "{{ username }}"
```
ericzolf, line 109: Should we still use `user` (and `team`) in tests if they're now deprecated?

ericzolf (commented):
Just typos and questions

awx_collection/tests/integration/targets/role/tasks/main.yml
sean-m-sullivan: I left them in as while they are depreciated they should still be tested, however I added additional tests for the users feature.

awx_collection/tests/integration/targets/role/tasks/main.yml
sean-m-sullivan: I left them in as while they are depreciated they should still be tested, however I added additional tests for the users feature.

AlanCoding:
Updated title because we are using these for release notes now, so we want a top-line message that can be shipped.

john-westcott-iv (approved):

PR #13947, opened 2023-05-03
Remove psycopg3 from wsrelay

TheRealHaoLiu:
##### SUMMARY
When upgrading to django4 we realize that django will default to psycopg3 if installed, we don't want to deal with django4 upgrade and get our code to be compatible with psycopg3 at the same time.

TODO: revert this when we move to psycopg3

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 22.1.1.dev71+g88a2f75c78
```

##### ADDITIONAL INFORMATION

```

```


awx/main/wsrelay.py
```diff
@@ -249,15 +251,15 @@ async def run(self):
...
+        async_conn = await aiopg.connect(
             dbname=database_conf['NAME'],
             host=database_conf['HOST'],
             user=database_conf['USER'],
             password=database_conf['PASSWORD'],
             port=database_conf['PORT'],
             **database_conf.get("OPTIONS", {}),
         )
-        await async_conn.set_autocommit(True)
+        # await async_conn.set_autocommit(True)
```
relrod: We can probably nix this line.

TheRealHaoLiu:
aiopg pulls in psycopg2-binary which we probably don't want to use 

@relrod is working on an alternative to use asyncpg

TheRealHaoLiu:
close in favor of https://github.com/ansible/awx/pull/13965/files

PR #13950, opened 2023-05-04
Get rid of custom `OrderedManyToManyDescriptor` and replace with library

AlanCoding:
##### SUMMARY
Our custom field `OrderedManyToManyDescriptor` did several questionable things. Instead of attempting to fix this, this deletes the field and replaces it with https://github.com/jazzband/django-sortedm2m

I've been making a few PRs to that library, touching up the status of its checks, and feel confident enough in it to use it.

The migration approach is kind of wacky. The ideal would be a simple use of [AlterSortedManyToManyField](https://github.com/jazzband/django-sortedm2m/blob/ddf321fc9736b9a9477d496c60250270111cf2ce/sortedm2m_tests/test_migrations.py#L46-L77), but no existing example shows migrating from a m2m field to a sorted-m2m field where the former had a `through` model, much less one that contains the sort position field. So that would take us well outside of charted territory. The approach here is well-understood and reliable, even if a little clunky - this adjusts the current models to match what the table/column names of the new (automatic) through model will be. Then it does a state-only operation to remove the old through model & field, and add the new field. The new model never added to the db schema, but it will instead, just pick up and start using the existing schema. This should be performant and avoid data migrations.

##### ISSUE TYPE

##### COMPONENT NAME
 - API


AlanCoding:
This looks viable, but I have one possible adjustment I want to test for the migration about whether the add-field/remove-field step can be condensed into a single AlterField step. Should that be possible, I'm unsure if it should be a state-only operation or a normal operation. In either case, I'm fairly sure that removing the through model would need to be a state-only operation.

AlanCoding:
Closing the loop on my last comment..

First attempt was to combine `RemoveField` and `AddField` operations into (locally) a `AlterField` operation, and this operation was moved out of the `state_operations` set, into the normal operations. This encounters the error from the core Django m2m field:

> ValueError: Cannot alter field main.Organization.galaxy_credentials into main.Organization.galaxy_credentials - they are not compatible types (you cannot alter to or from M2M fields, or add or remove through= on M2M fields)

The issue isn't that the custom field types are not compatible (they are) but that migrating an M2M field with a through model cannot be done to a field without a through model. Logically, because Django can't assure the table and columns will match (which is the exact manual work I did here).

Next attempt - what if we don't combine them, but just move those 2 steps out of the `state_operations`? Well, that will compile the `sqlmigrate` plan, but produces a `CREATE TABLE` step (as expected) for adding the new field. Then when you execute the migration, as you would expect, it errors because the table already exists.

For a third approach (not tried), you could maybe take the `RemoveField` operation out of the `state_operations` steps. Because it uses a `through` table anyway, it's a state-only operation (the step is not removing the through table)... even if listed in normal operations. I'm still undecided about whether this is preferable. I remain undecided, but leaning "no". The arguments for/against get pretty academic.

AlanCoding:
Integration tests are all normal with this, so marked it as ready for review.

Issue #13952, opened 2023-05-04
Better utilize database connections in the heartbeet service


jessicamack:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

Enhancement to Existing Feature

### Feature Summary

We currently open a new database connection and keep it open while we use it to send heartbeats. We don't need to use a new connection or keep the connection open for the service as it only sends occasional notifications.

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

Start the heartbeet service and check the number of active postgres connections. 

### Current results

Currently when the service starts it opens a new connection.

### Sugested feature result

The feature should work as it did before the change.

jessicamack:
The work for this has been rolled into #14034.

PR #13953, opened 2023-05-04
Free up persistent db connection in heartbeet service

jessicamack:
##### SUMMARY

related #13952

Use existing db connection instead of creating a new one and close the connection one the message has been sent.

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 22.1.1
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/management/commands/run_heartbeet.py
+++ b/awx/main/management/commands/run_heartbeet.py
@@ -58,11 +58,11 @@ class Command(BaseCommand):
         sys.exit(0)
 
     def do_hearbeat_loop(self):
-        with pg_bus_conn(new_connection=True) as conn:
-            while True:
+        while True:
+            with pg_bus_conn() as conn:
                 logger.debug('Sending heartbeat')
                 conn.notify('web_heartbeet', self.construct_payload())
-                time.sleep(settings.BROADCAST_WEBSOCKET_BEACON_FROM_WEB_RATE_SECONDS)
+            time.sleep(settings.BROADCAST_WEBSOCKET_BEACON_FROM_WEB_RATE_SECONDS)
 
     def handle(self, *arg, **options):
         self.print_banner()


AlanCoding:
If I check out this branch, I still see an open connection from the heatbeet.

```
bash-5.1$ echo "select pid,application_name, state from pg_stat_activity;" | awx-manage dbshell
 pid |             application_name              | state  
-----+-------------------------------------------+--------
  30 |                                           | 
  32 |                                           | 
  46 | awx-144-heartbeet-awx_1                   | idle
  48 | awx-146-rsyslog_configurer_listener-awx_1 | idle
  49 | awx-146-rsyslog_configurer-awx_1          | idle
  54 | awx-142-wsrelay-awx_1                     | idle
  51 | awx-150-cache_clear_listener-awx_1        | idle
  52 | awx-137-dispatcher_listener-awx_1         | idle
 588 | awx-875-dispatcher_worker-awx_1           | idle
 175 | awx-138-callback_receiver-awx_1           | idle
 597 | awx-260-dispatcher_periodic-awx_1         | idle
 598 | awx-889-dispatcher_worker-awx_1           | idle
 589 | awx-881-dispatcher_worker-awx_1           | idle
 591 | awx-887-dispatcher_worker-awx_1           | idle
 599 | awx-883-dispatcher_worker-awx_1           | idle
 600 | psql                                      | active
  28 |                                           | 
  27 |                                           | 
  29 |                                           | 
(19 rows)
```

You could get rid of this too by `from django.db import connection` and then `connection.close()` in the loop after the message is sent.

jessicamack:
@AlanCoding The service should only be using existing connections now so I'm looking into why it's showing as a dedicated connection still. Would we still want to close the connection if we're using an existing one that should be released afterward?

relrod:
> The service should only be using existing connections now

The service is running independently from everything else, so it has no existing connection to share (it's not running in wsgi or anything). It needs its own connection. The goal is just to not have it hold on to that connection in between messages for no reason.


> Would we still want to close the connection if we're using an existing one that should be released afterward?

I think if you keep `new_connection=True`, it should do what you want. When we exit the context manager, it will automatically close the connection.

AlanCoding:
> The service should only be using existing connections now so I'm looking into why it's showing as a dedicated connection still. Would we still want to close the connection if we're using an existing one that should be released afterward?

As to the question of "should", the setting `BROADCAST_WEBSOCKET_BEACON_FROM_WEB_RATE_SECONDS` makes the period 15 seconds. For a comparison point, look at uWSGI, it frequently closes connections not being used by its workers, which is why we rarely see it in the connection list, I believe.

PR #13959, opened 2023-05-05
Adds missing rel="noopener noreferrer" to each link element with target="_blank"

mabashian:
##### SUMMARY
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noopener
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noreferrer

`noopener` may be overkill at this point given:

`Note: Setting target="_blank" on <a> elements now implicitly provides the same rel behavior as setting rel="noopener" which does not set window.opener. See [browser compatibility](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#browser_compatibility) for support status.`

I didn't want to make assumptions about browsers though so I went ahead and made it consistent.

##### ISSUE TYPE

##### COMPONENT NAME
 - UI

### DIFF
--- a/awx/ui/.eslintrc.json
+++ b/awx/ui/.eslintrc.json
@@ -99,7 +99,8 @@
           "fieldName",
           "splitButtonVariant",
           "pageKey",
-          "textId"
+          "textId",
+          "rel"
         ],
         "ignore": [
           "Ansible",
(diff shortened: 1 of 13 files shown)


AlexSCorey (approved):
Looks good to me.

Issue #13927, opened 2023-04-30
Add an enforced state to the collection modules


sean-m-sullivan:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

Enhancement to Existing Feature

### Feature Summary

Right now when using the modules with the idea of config as code, If the user does not provide the field, it is skipped, and not acted on.

The proposal is to allow a user to set the modules to enforce the default values. The goal of this is to prevent any drift from changes in the GUI, or if someone wants to reset an option back to default. The end goal to allow for complete configuration as code for the Controller/AWX.

In addition some fields if provided as blank or empty lists they come back as an error as it searches for all credentials or execution environments, or gets skipped due to some logic.

An example of this is
awx.awx.schedules
When empty returns multiple objects: credentials , labels, instance_groups
Not allowed to be empty, but not required: unified_job_template
Setting "" does not enforce empty: description, job_slice_count, diff_mode, job_type, verbosity

awx.awx.organizations
Not allowed to be empty, but not required: custom_virtualenv, default_environment
Setting "" does not enforce empty: max_hosts

awx.awx.exeuction_environments
Setting "" does not enforce empty: description, credential

Sometimes this can be worked around using something like the following code to turn it on/off:
galaxy_credentials: "{{ (__controller_organizations_item.galaxy_credentials | default(( [] if controller_configuration_organizations_enforce_defaults else omit), true)) }}"

However a better solution is to enforce it at the module level.

I've made code changes and tested the organization and schedule modules of how this would work. This is based on a branch of the exists state, so its hard to do an actual fork compare.
schedule
https://editor.mergely.com/8rAVJRNB
organization
https://editor.mergely.com/z2LLlBeO

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [X] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

    - name: Create organization
      awx.awx.schedule:
        name: sched1
        state: present
        unified_job_template: Demo Job Template
        rrule: "{{ query('awx.awx.schedule_rrule', 'none', start_date='1979-09-13 03:45:07') }}"

    - name: Create organization
      awx.awx.organization:
        name: Default
        state: present

### Current results

[...]

PR #13960, opened 2023-05-06
In collection, Add an enforced state to the modules

sean-m-sullivan:
##### SUMMARY
Follow up PR from #13927, this is expected to fail tests until #13890 is merged, as it contains code from that PR

This enforces default values being submitted to the API, this PR is an overview of how that enforcement mechanism would be achieved across the collection, This PR is for feedback on how this was achieved so that I can go through and add it to the rest of the modules. The idea is to enforce a state to prevent drift when an option is not provided and option is enforced. This is compared to exists that makes no changes if the object exists, or present that only makes changes for the options given to the module.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - Collection

##### AWX VERSION

```
22.1.0
```

### DIFF
--- a/awx_collection/plugins/modules/schedule.py
+++ b/awx_collection/plugins/modules/schedule.py
@@ -145,8 +145,8 @@ options:
       type: bool
     state:
       description:
-        - Desired state of the resource.
-      choices: ["present", "absent"]
+        - Enforced state enforces default values of any option not provided.
+      choices: ["present", "absent", "exists", enforced]
       default: "present"
       type: str
 extends_documentation_fragment: awx.awx.auth
@@ -174,7 +174,7 @@ EXAMPLES = '''
     name: "{{ sched1 }}"
     state: present
     unified_job_template: "Demo Job Template"
-    rrule: "{{ query(awx.awx.schedule_rruleset, '2022-04-30 10:30:45', rules=rrules, timezone='UTC' ) }}"
+    rrule: "{{ query('awx.awx.schedule_rruleset', '2022-04-30 10:30:45', rules=rrules, timezone='UTC' ) }}"
   vars:
     rrules:
       - frequency: 'day'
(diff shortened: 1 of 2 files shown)


AlanCoding:
@john-westcott-iv brought up a good proposal about using OPTIONS to automatically get the defaults.

AlanCoding:
I have a question - what about related lists? I want to have that behave coherently given the state options of exists or enforced.

Logically, the "default" value for a related list (like credentials for a job template) would be an empty list. However, users have the ability to not specify credentials, in which case, they will not be adjusted in any way. If someone wanted it "enforced" and the JT _has_ related credentials already, should they be removed? I'm leaning towards "yes".

What about the "exists" option? If creating a new JT, should it associate related credentials? Strong lean towards "yes". Should the normal "present" state enforce the list is _exactly equal_ to the provided list, or just that the items specified are there? I lean towards the former (current behavior), but worry this might latter get us in some kind of catch 22 I haven't thought about.

sean-m-sullivan:
@AlanCoding This is why it is piggybacked on the state, 
If state == present, things behave EXACTLY as they have in the past, if State == exists, it exits out after confirming existance, ONLY if state == enforced does it start overwriting things, its why you were right in adding this in as a state option, as they cannot coexist.

AlanCoding:
My near-term question here is - if I use the `job_template` module with state=enforced, and it currently has credentials, and the module does not specify any credentials, does it remove existing credentials?

I think we can work this out quickly and efficiently here. I would say, yes, it _should_ remove credentials and any objects in any other relationships like this.

awx_collection/plugins/modules/schedule.py
```diff
@@ -265,15 +265,29 @@ def main():
...
+    # Fail early now if no unified job template provided now that enforced defaults set.
+    if unified_job_template is None and enforced_defaults:
+        module.fail_json(msg='When using enforced state unified_job_template is a required field')
+
     if credentials is not None:
         association_fields['credentials'] = []
         for item in credentials:
             association_fields['credentials'].append(module.resolve_name_to_id('credentials', item))
+    elif credentials is None and enforced_defaults:
+        association_fields['credentials'] = []
```
sean-m-sullivan, line 290: @AlanCoding This line I tested removes existing credentials when submited to the API. when none are supplied and enforced is true.

sean-m-sullivan: This would be the same as in a JT

PR #13965, opened 2023-05-08
[wsrelay] Switch from psycopg 3 to asyncpg

relrod:
##### SUMMARY

Due to dependency issues specifically around upgrading to Django 4.2, we cannot feasibly have a dependency on psycopg2 and psycopg3. The only place that was currently using psycopg3 was wsrelay.

Change wsrelay to use the asyncpg library and psycopg2 instead.

Tested locally on kind with a dev build of awx.

##### ISSUE TYPE

##### COMPONENT NAME

 - API


TheRealHaoLiu:
tested on a openshift cluster too, this works

TheRealHaoLiu (approved):

shanemcd (approved):

Issue #13910, opened 2023-04-26
[hop node] prevent manual peering of control plane nodes to hop node


fosterseth:
`peers` field in API should be read-only for control type nodes

To establish peers to control plane nodes, `peers_from_control_nodes` should be set to True on the hop / execution nodes.

PR #13966, opened 2023-05-08
Prevent manual peering of control plane nodes to hop node

djyasin:
##### SUMMARY

Prevent manual peering of control plane nodes to hop node
- related #13910

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 22.1.1
```

##### Test Scenarios:
- Ensure that if `peers` has been set, it cannot be changed to a different node
- Esnure that `peers` cannot be set manually, must be set via `peers_from_control_nodes`

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -5490,6 +5490,13 @@ class InstanceSerializer(BaseSerializer):
 
         if peers_from_control_nodes and node_type not in (Instance.Types.EXECUTION, Instance.Types.HOP):
             raise serializers.ValidationError("peers_from_control_nodes can only be enabled for execution or hop nodes.")
+
+        if node_type in (Instance.Types.CONTROL):
+            if self.instance and 'peers' in attrs and set(self.instance.peers.all()) != set(attrs['peers']):
+                raise serializers.ValidationError(
+                    "Setting peers manually for control nodes is not allowed. Enable peers_from_control_nodes on the hop and execution nodes instead."
+                )
+
         return super().validate(attrs)
 
     def validate_node_type(self, value):


awx/api/serializers.py
```diff
@@ -5474,6 +5475,32 @@ def get_percent_capacity_remaining(self, obj):
     def get_health_check_pending(self, obj):
         return obj.health_check_pending
 
+    def peers_read_only(self, attrs, value):
+        # `peers` field in API should be read-only for control type nodes
+        def get_field_from_model_or_attrs(fd):
+            return attrs.get(fd, self.instance and getattr(self.instance, fd) or None)
+
+        node_type = get_field_from_model_or_attrs("node_type")
+        if node_type == 'control' 'peers'(read_only=True):
```
bpedersen2: looks strange, the peers(readonly) should go on next line after colon

djyasin: Thank you so much! Still working on this one, but I absolutely appreciate your feedback!

awx/api/serializers.py
```diff
@@ -5,6 +5,7 @@
 import copy
 import json
 import logging
+from multiprocessing.sharedctypes import Value
```
AlanCoding: Doesn't look like this is used anymore.

djyasin: Thank you so much!! I will get this fixed as soon as I fix my rebase!

awx/api/serializers.py
```diff
@@ -5492,6 +5492,18 @@ def get_field_from_model_or_attrs(fd):
             raise serializers.ValidationError("peers_from_control_nodes can only be enabled for execution or hop nodes.")
         return super().validate(attrs)
 
+    # import sdb
+    # sdb.Sdb(notify_host='*:30080').set_trace()
+    def peers_validation(self, attrs, value):
```
fosterseth: in general validation for a specific field uses `def validate_fieldname` and the only parameter is `value`

you probably want to move the logic to `def validate`

awx/api/serializers.py
```diff
@@ -5460,14 +5460,24 @@ def get_percent_capacity_remaining(self, obj):
     def get_health_check_pending(self, obj):
         return obj.health_check_pending
 
-    def validate(self, data):
-        if self.instance:
-            if self.instance.node_type == Instance.Types.HOP:
```
gamuniz: are we losing this validation for some reason? It seems this could still apply for VM installation as HOP nodes don't have the requisite packages to do execution.

djyasin: Thank you for raising that point! Should it be renamed since we have another method called `validate`?

awx/api/serializers.py
```diff
@@ -5490,6 +5490,12 @@ def get_field_from_model_or_attrs(fd):
 
         if peers_from_control_nodes and node_type not in (Instance.Types.EXECUTION, Instance.Types.HOP):
             raise serializers.ValidationError("peers_from_control_nodes can only be enabled for execution or hop nodes.")
+
+        if self.instance.node_type == 'control' and self.instance.peers:
```
fosterseth: `self.instance.peers` _is_ valid and expected for control nodes. For example, add an execution node via the API with `peers_from_control_nodes` set to True. Now go back and look at the control instance and you'll see `peers` has something in it. If you were to then patch the control instance (say, by changing the capacity_adjustment value), then you'd hit this validation error, which is not what we want.

awx/api/serializers.py
```diff
@@ -5487,9 +5487,17 @@ def get_field_from_model_or_attrs(fd):
...
         peers_from_control_nodes = get_field_from_model_or_attrs("peers_from_control_nodes")
+        peers = get_field_from_model_or_attrs("peers")
 
         if peers_from_control_nodes and node_type not in (Instance.Types.EXECUTION, Instance.Types.HOP):
             raise serializers.ValidationError("peers_from_control_nodes can only be enabled for execution or hop nodes.")
+
+        if node_type in (Instance.Types.CONTROL) and len(peers) > 0:
+            if self.instance and set(self.instance.peers.all()) != set(attrs['peers']):
+                raise serializers.ValidationError(
+                    "Peers is only possible for execution or hop node types. Set 'peers' via 'peers_from_control_nodes' instead. This value cannot be modified once set."
```
fosterseth: Setting peers manually for control nodes is not allowed. Enable peers_from_control_nodes on the hop and execution nodes instead.

awx/api/serializers.py
```diff
@@ -5487,9 +5487,17 @@ def get_field_from_model_or_attrs(fd):
             raise serializers.ValidationError("Can only create instances on Kubernetes or OpenShift.")
         node_type = get_field_from_model_or_attrs("node_type")
         peers_from_control_nodes = get_field_from_model_or_attrs("peers_from_control_nodes")
+        peers = get_field_from_model_or_attrs("peers")
 
         if peers_from_control_nodes and node_type not in (Instance.Types.EXECUTION, Instance.Types.HOP):
             raise serializers.ValidationError("peers_from_control_nodes can only be enabled for execution or hop nodes.")
+
+        if node_type in (Instance.Types.CONTROL) and len(peers) > 0:
+            if self.instance and set(self.instance.peers.all()) != set(attrs['peers']):
```
fosterseth: you might need to check peers is a key in attrs first

fosterseth:
manual testing validation

go to api/v2/instances, click on control instance

- Hit patch with no changes to input form, assert success
- Add a peer name to the peers field, hit patch, assert validation error
- Remove a current peer name from peers field, hit patch, assert validation error
- Remove peers line entirely, hit patch, assert success

djyasin:
> manual testing validation
> 
> go to api/v2/instances, click on control instance
> 
> * Hit patch with no changes to input form, assert success
> * Add a peer name to the peers field, hit patch, assert validation error
> * Remove a current peer name from peers field, hit patch, assert validation error
> * Remove peers line entirely, hit patch, assert success

<b style="font-weight:normal;" id="docs-internal-guid-2d93f6fe-7fff-2dcf-bae0-f90a71cc9593"><p dir="ltr" style="line-height:1.38;background-color:#ffffff;margin-top:0pt;margin-bottom:12pt;"></p><div dir="ltr" style="margin-left:0pt;" align="left">

Scenario | Results
-- | --
Hit patch with no changes to input form, assert success | 👍
Add a peer name to the peers field, hit patch, assert validation error | 👍
Remove a current peer name from peers field, hit patch, assert validation error | 👍
Remove peers line entirely, hit patch, assert success | 👍

</div></b>

fosterseth (approved):

PR #13975, opened 2023-05-10
[troubleshooting] Make dispatcher --status accurate real-time

AlanCoding:
##### SUMMARY
With this change:

```
bash-5.1$ awx-manage run_dispatcher --status
2023-05-10 13:29:07,138 WARNING  [-] awx.main.dispatch checking dispatcher status for awx_1
listening on ['tower_broadcast_all', 'tower_settings_change', 'awx_1']
Recorded at: 2023-05-10 13:29:07 UTC 
awx_1[pid:78909] workers total=4 min=4 max=314 
.  worker[pid:79000] sent=1 finished=1 qsize=0 rss=125.996MB [IDLE]
.  worker[pid:79001] sent=2 finished=2 qsize=0 rss=129.141MB [IDLE]
.  worker[pid:79002] sent=1 finished=1 qsize=0 rss=127.152MB [IDLE]
.  worker[pid:79003] sent=3 finished=3 qsize=0 rss=126.535MB [IDLE]
```

You can run this _over and over_ again and it almost never shows a scheduled task running. These tasks take on the order of `0.02` seconds, which the task manager being the longest at around `0.1` seconds on a quiet system. Given that these run on periods of 20 seconds to 60 seconds, we actually _expect_ to (almost) never see such tasks. We're looking at a 1 in 200 chance.

But right now, the dispatcher regularly shows workers with active tasks. This is because those workers are not actually active. They are only shown in-memory to be active because their finished queue has not been checked.

This _does_ make it more likely that the `--status` command will hang, and while we might want some kind of balancing fix for that, I would like to look elsewhere instead of what this is removing.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/dispatch/worker/base.py
+++ b/awx/main/dispatch/worker/base.py
@@ -17,7 +17,6 @@ from django.conf import settings
 
 from awx.main.dispatch.pool import WorkerPool
 from awx.main.dispatch import pg_bus_conn
-from awx.main.utils.common import log_excess_runtime
 from awx.main.utils.db import set_connection_name
 import awx.main.analytics.subsystem_metrics as s_metrics
 
@@ -42,8 +41,6 @@ class WorkerSignalHandler:
 
 
 class AWXConsumerBase(object):
-    last_stats = time.time()
-
     def __init__(self, name, worker, queues=[], pool=None):
         self.should_stop = False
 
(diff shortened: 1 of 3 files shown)


awx/main/dispatch/worker/base.py
```diff
@@ -115,16 +112,6 @@ def process_task(self, body):
         self.pool.write(queue, body)
         self.total_messages += 1
 
-    @log_excess_runtime(logger)
-    def record_statistics(self):
```
gamuniz, line 120: is this what is being replaced by the --history in the other PR?

AlanCoding: No, I'm trying to keep these logically separated. This is replaced by _this_ PR.

Before, we periodically produced the `.debug()` template and put it into redis. Then when someone ran the `--status` command we read what was in redis. Well, this changes is so that the `--status` command sends a message to the dispatcher, dispatcher renders the template, then the dispatcher sends it back on another (temporary) channel to the `awx-manage` command. Since I'm changing how the command works, nothing else is reading this data from redis, so I remove all the logic to save it there.

PR #13980, opened 2023-05-10
Update make target for extracting strings to do so for ui_next too

rooftopcellist:
##### SUMMARY

We were only extracting strings for the `ui` directory, not the tech preview `ui_next` one as well.  

##### ISSUE TYPE

##### COMPONENT NAME

 - Localization
 - Other

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -660,10 +660,12 @@ awx-kube-dev-build: Dockerfile.kube-dev
 ## generate UI .pot file, an empty template of strings yet to be translated
 pot: $(UI_BUILD_FLAG_FILE)
 	$(NPM_BIN) --prefix awx/ui --loglevel warn run extract-template --clean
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -660,10 +660,12 @@ awx-kube-dev-build: Dockerfile.kube-dev
 ## generate UI .pot file, an empty template of strings yet to be translated
 pot: $(UI_BUILD_FLAG_FILE)
 	$(NPM_BIN) --prefix awx/ui --loglevel warn run extract-template --clean
+	$(NPM_BIN) --prefix awx/ui --loglevel warn run extract-template --clean
```
shanemcd: Is this supposed to be `ui_next` like below?

shanemcd (approved):

Issue #12776, opened 2022-08-29
[Metrics&Observability] Track dispatcher metrics


AlanCoding:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

Enhancement to Existing Feature

### Feature Summary

We have already implemented a set of metrics for the task manager (and related dependency/workflow managers) in https://github.com/ansible/awx/issues/12225

This is a request that we track similar metrics for the process that listens to pg_notify for tasks and user automation to run (the dispatcher parent process). To articulate some specifics, we could track:

 - the total number of messages processed (either on a cumulative or differential basis)
   - a subdivision of that number for user automation versus periodic system tasks versus special trigger tasks
 - the time taken in `process_task`
 - the time taken for `cleanup`, `record_metrics`, or any others that may run related to that
 - number of workers at any given time
   - maybe also the number of scale-up and scale-down events, which may not directly follow from the active number

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

N/A

We have some logs for certain threshold events, but no metrics right now.

### Current results

N/A

### Sugested feature result

These should go in `/api/v2/metrics/` and we can make graphs for them.

PR #13989, opened 2023-05-11
Add subsystem metrics for the dispatcher

AlanCoding:
##### SUMMARY
Connect https://github.com/ansible/awx/issues/12776

Demo:

```
# HELP dispatcher_pool_scale_up_events Number of times local dispatcher scaled up a worker since startup
# TYPE dispatcher_pool_scale_up_events gauge
dispatcher_pool_scale_up_events{node="awx_1"} 12
# HELP dispatcher_pool_active_task_count Number of active tasks in the worker pool when last task was submitted
# TYPE dispatcher_pool_active_task_count gauge
dispatcher_pool_active_task_count{node="awx_1"} 0
# HELP dispatcher_pool_worker_count Highest number of workers in worker pool in last collection interval, about 20s
# TYPE dispatcher_pool_worker_count gauge
dispatcher_pool_worker_count{node="awx_1"} 6
# HELP dispatcher_availability Fraction of time last interval when dispatcher was available to receive messages
# TYPE dispatcher_availability gauge
dispatcher_availability{node="awx_1"} 0.9945928728946261
```

I started with a _very_ long list of metrics that we could put in, but settled on these as providing unique value with each metric, and simple enough they can be understood. I feel like we added too much data with the task manager & callback receiver.

These have absolutely _red hot_ implications for system stability. If the availability is sub-90%, then message processing _will_ lag notably. Scale up events should also be infrequent, and they are not. This is already a reported finding, but this gives instrumentation for the perfscale team to quantify that and hold some regression coverage over it.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
I am trying to offer this as a model to use the metrics to make progress in performance. Here are patches that are intended to address these performance metrics:

 - https://github.com/ansible/awx/pull/13975 should increase (improve) `dispatcher_availability` by a static amount by avoiding running the debug template periodically
 - https://github.com/ansible/awx/pull/13990 should offer dramatic improvements by decreasing `dispatcher_pool_scale_up_events` (indeed, making it flat) and decreasing `dispatcher_pool_worker_count`
 - https://github.com/ansible/awx/pull/13993 should also increase `dispatcher_availability`

### DIFF
--- a/awx/main/analytics/subsystem_metrics.py
+++ b/awx/main/analytics/subsystem_metrics.py
@@ -209,6 +209,11 @@ class Metrics:
             SetFloatM('workflow_manager_recorded_timestamp', 'Unix timestamp when metrics were last recorded'),
             SetFloatM('workflow_manager_spawn_workflow_graph_jobs_seconds', 'Time spent spawning workflow tasks'),
             SetFloatM('workflow_manager_get_tasks_seconds', 'Time spent loading workflow tasks from db'),
+            # dispatcher subsystem metrics
+            SetIntM('dispatcher_pool_scale_up_events', 'Number of times local dispatcher scaled up a worker since startup'),
+            SetIntM('dispatcher_pool_active_task_count', 'Number of active tasks in the worker pool when last task was submitted'),
+            SetIntM('dispatcher_pool_worker_count', 'Highest number of workers in worker pool in last collection interval, about 20s'),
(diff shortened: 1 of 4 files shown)


kdelee (approved):

AlanCoding:
I added this to the demo dashboard so here is what it looks like.

Screenshot from 2023-05-17 13-46-42

john-westcott-iv (approved):

Issue #12777, opened 2022-08-29
Reduce the number of dispatcher scale up/down events


AlanCoding:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

Enhancement to Existing Feature

### Feature Summary

The task dispatcher will scale workers down at up at a rate much higher than necessary.

Expectation is that if we have a constant number of jobs running, scale ups and downs should not be necessary _at all_. Meaning, if we have 0 jobs running, we should always be fine at the minimum number of workers. But if we have 10 jobs running, we should also have no need to scale up or down, if that number has held constant for more than ~30 seconds.

Right now this is not the case. We know of cases when we may actually scale a worker _down and then up_ right before starting a new task.

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

launch a constant number of sleep tasks, observe number of workers.

this will be easier to do when https://github.com/ansible/awx/issues/12776 is completed.

### Current results

We know from the logs that workers are always scaling up and down, even when there is no need to.

### Sugested feature result

Smarter worker scaling logic.

### Additional information

I attempted this in https://github.com/ansible/awx/pull/12584 and reverted that in https://github.com/ansible/awx/pull/12766

In the end, that approach left _too many_ idle workers in reserve. We need a solution that keeps _just enough_ idle workers free. In practice, this probably means keeping as many workers as the number of scheduled tasks that might run at the same time. So, I suggest that the best solution would be to add a timeout to scheduled tasks, then compute the max number of simultaneous scheduled tasks, and then keep this number of workers available _beyond the current workers running non-scheduled tasks_. This is a much more complex heuristic than my prior attempt, but it would give a much more satisfactory result. I'm unhappy with both the increased memory use of my prior attempt, and the impact of database connection usage.

PR #13990, opened 2023-05-11
Spread out submission of scheduled tasks to avoid bursting

AlanCoding:
##### SUMMARY
Connect https://github.com/ansible/awx/issues/12777, this is my revised solution to the problem, as my prior attempt was a bust. This can be tested with these new subsystem metrics https://github.com/ansible/awx/pull/13989 to get an accurate measurement of the change.

Before:

```
dispatcher_pool_scale_up_events{node="awx_1"} 22
dispatcher_pool_active_task_count{node="awx_1"} 1
dispatcher_pool_worker_count{node="awx_1"} 6
dispatcher_availability{node="awx_1"} 0.9943700416419148
```

After:

```
dispatcher_pool_scale_up_events{node="awx_1"} 4
dispatcher_pool_active_task_count{node="awx_1"} 1
dispatcher_pool_worker_count{node="awx_1"} 4
dispatcher_availability{node="awx_1"} 0.9946726353763824
```

Specifically:
 - this aims to reduce `dispatcher_pool_scale_up_events` so that it doesn't increase with 3 jobs in-progress or fewer
 - this should also decrease `dispatcher_pool_worker_count` which is another angle of the same thing

For a system not running jobs (and hopefully a system running just a few jobs) the workers stays at a static number of 4, which should have been the original expectation. However, over time, we have accumulated a number of additional periodic tasks to this system which was not designed for it. The net result was that schedule jobs were released in a "burst" that pushes workers from their min of 4 up to 6 for on-the-minute scheduled tasks. This causes 2 workers (timing dependent) to scale up and down constantly when the system is idle (and also when busy).

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/dispatch/periodic.py
+++ b/awx/main/dispatch/periodic.py
@@ -26,6 +26,7 @@ class Scheduler(Scheduler):
             set_connection_name('periodic')  # set application_name to distinguish from other dispatcher processes
 
             while True:
+                loop_start = time.time()
                 if os.getppid() != ppid:
                     # if the parent PID changes, this process has been orphaned
                     # via e.g., segfault or sigkill, we should exit too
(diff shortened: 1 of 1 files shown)


awx/main/dispatch/periodic.py
```diff
@@ -38,10 +39,21 @@ def run():
...
+                    for job in runnable_jobs[:-1]:
+                        self._run_job(job)
+                        time.sleep(0.3)
+                    if runnable_jobs:
+                        self._run_job(runnable_jobs[-1])
+
                 except Exception:
                     logger.exception('encountered an error while scheduling periodic tasks')
-                time.sleep(idle_seconds)
+                time.sleep(idle_seconds - (time.time() - loop_start))
```
AlanCoding: The pre-existing code here seems extraordinarily lazy... as there is a documented way to do this

https://schedule.readthedocs.io/en/stable/examples.html#time-until-the-next-execution

Here, that would be

```python
time.sleep(self.idle_seconds())
```

Why on Earth we would not start out using this, I do not know, and I am a bit dumbfounded about how the library is completely circumvented with an obviously less-good alternative. This demands some remedy my changes here, because I'm adding additional sleeps and any weird skews this previously caused would worsen if this line isn't changed.

AlanCoding:
I think I've decided that I do want to test this with an inventory of sleep jobs going, will look into that tomorrow morning maybe.

AlanCoding:
Screenshot from 2023-05-18 09-01-16

This shows a burst of (I think) 7 jobs, so when they are all running we're up to 8 workers. When all is said and done there are 8 scale-up events, which is :100: perfect.

PR #13993, opened 2023-05-11
Reduce queue checks needed for task assignment

AlanCoding:
##### SUMMARY
Another patch to offer to use https://github.com/ansible/awx/pull/13989 to test with.

We have to check the finished queue to evaluate a worker's idle/busy state this is triggered
 - for _all_ workers when `should_grow` is called
 - for each worker as we are searching for an idle worker

the argument here is that we are checking the finished queue two times before dispatching a task. So we just don't check the first time.

This is presented as something which should increase the `dispatcher_availability`, but this should be most acute in the scenario that we are under very high loads, because in this case we will have the largest number of finished queues to check.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/dispatch/pool.py
+++ b/awx/main/dispatch/pool.py
@@ -4,7 +4,6 @@ import random
 import signal
 import sys
 import time
-import traceback
 from datetime import datetime
 from uuid import uuid4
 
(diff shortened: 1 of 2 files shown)


AlanCoding:
Done on code that was deleted!

PR #13809, opened 2023-04-06
Enhance usage metrics collection

john-westcott-iv:
##### SUMMARY

Subscription Settings page: Deleted host metrics and Active previously deleted info. Visible when SUBSCRIPTION_USAGE_MODEL == 'unique-managed_hosts'

Host Metric cleanup task - soft deletes HostMetric records which were automated more than 12 months ago

Uploading HostMetrics to the analytics (CLI awx-manage gather_analytics)
It adds host_metric_table.csv to the files in the sent tarballs.

Unlike other collectors, this one introduces hybrid full/incremental sync. It means it performs full sync of whole table once per 30 days. Incremental sync otherwise.

PR extends the @register decorator with full_sync_interval - integer: number of days when the full sync should be performed instead of incremental. If not provided, incremental sync is performed as before.

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 - UI

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

PR #13999, opened 2023-05-15
HostMetricSummaryMonthly command + scheduled task

slemrmartin:
##### SUMMARY

Adds command `awx-manage host_metric_summary_monthly`
Adds scheduled task `host_metric_summary_monthly` (once per week)

It computes last 36 months of HostMetricSummaryMonthly table (1 row == 1 month)
36 months is also threshold for deleting HostMetrics (will be implemented also to this feature branch)

It computes subscription consumption and unique hosts added/deleted in given month (more precisely HostMetrics)
It'll serve as a backend for UI table with graph of subscription consumption usage. 

More described in the ADR

- https://issues.redhat.com/browse/AA-1590
- ADR 0017 (available from the Jira task)

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 - CLI
- Scheduler
- 
##### AWX VERSION

```
awx: 22.2.1.dev5+g0f57e16f4f.d20230515
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/conf.py
+++ b/awx/main/conf.py
@@ -863,6 +863,15 @@ register(
 )
 
 register(
+    'HOST_METRIC_SUMMARY_TASK_LAST_TS',
+    field_class=fields.DateTimeField,
+    label=_('Last computing date of HostMetricSummaryMonthly'),
+    allow_null=True,
+    category=_('System'),
+    category_slug='system',
+)
+
+register(
     'AWX_CLEANUP_PATHS',
     field_class=fields.BooleanField,
     label=_('Enable or Disable tmp dir cleanup'),
--- a/awx/main/models/inventory.py
+++ b/awx/main/models/inventory.py
@@ -910,7 +910,7 @@ class HostMetric(models.Model):
 
 class HostMetricSummaryMonthly(models.Model):
     """
-    HostMetric summaries computed by scheduled task <TODO> monthly
+    HostMetric summaries computed by scheduled task 'awx.main.tasks.system.host_metric_summary_monthly' monthly
     """
 
     date = models.DateField(unique=True)
--- a/awx/main/models/inventory.py
+++ b/awx/main/models/inventory.py
@@ -899,13 +899,13 @@ class HostMetric(models.Model):
 
             last_automation_before = now() - dateutil.relativedelta.relativedelta(months=months_ago)
 
-            logger.info(f'Cleanup [HostMetric]: soft-deleting records last automated before {last_automation_before}')
+            logger.info(f'cleanup_host_metrics: soft-deleting records last automated before {last_automation_before}')
             HostMetric.active_objects.filter(last_automation__lt=last_automation_before).update(
                 deleted=True, deleted_counter=models.F('deleted_counter') + 1, last_deleted=now()
             )
             settings.CLEANUP_HOST_METRICS_LAST_TS = now()
         except (TypeError, ValueError):
-            logger.error(f"Cleanup [HostMetric]: months_ago({months_ago}) has to be a positive integer value")
+            logger.error(f"cleanup_host_metrics: months_ago({months_ago}) has to be a positive integer value")
 
 
 class HostMetricSummaryMonthly(models.Model):
--- a/awx/main/tasks/handlers/host_metric_summary_monthly_task.py
+++ b/awx/main/tasks/handlers/host_metric_summary_monthly_task.py
@@ -1,0 +1,169 @@
+import datetime
+from dateutil.relativedelta import relativedelta
+
+from django.conf import settings
+from django.db.models import Count
+from django.db.models.functions import TruncMonth
+from awx.main.models.inventory import HostMetric, HostMetricSummaryMonthly
+from awx.conf.license import get_license
+
+
+class HostMetricSummaryMonthlyTask:
+    """
+    This task computes last <threshold> months of HostMetricSummaryMonthly table
+    Each record in the table represents changes in HostMetric table in one month
+    <threshold> is setting CLEANUP_HOST_METRICS_HARD_THRESHOLD
+    It always overrides all the months newer than <threshold>, never updates older months
+    Algorithm:
+    - hosts_added are HostMetric records with first_automation in given month
+    - hosts_deleted are HostMetric records with deleted=True and last_deleted in given month
+    - license_consumed is license_consumed(previous month) + hosts_added - hosts_deleted
+    - - license_consumed for first updated month is computed also from all host metrics created before this month
+    - license_capacity is set only for current month, and it's never updated (value taken from current subscription)
+    """
+
+    def __init__(self):
+        self.host_metrics = {}
+        self.processed_month = self._get_first_month()
+        self.existing_summaries = None
+        self.existing_summaries_idx = 0
+        self.existing_summaries_cnt = 0
+        self.records_to_create = []
+        self.records_to_update = []
+
+    def execute(self):
+        self._load_existing_summaries()
+        self._load_hosts_added()
+        self._load_hosts_deleted()
+
+        # Get first month after last hard delete
+        month = self._get_first_month()
+        license_consumed = self._get_license_consumed_before(month)
+
+        # Fill record for each month
+        while month <= datetime.date.today().replace(day=1):
+            summary = self._find_or_create_summary(month)
+            # Update summary and update license_consumed by hosts added/removed this month
+            self._update_summary(summary, month, license_consumed)
+            license_consumed = summary.license_consumed
+
+            month = month + relativedelta(months=1)
+
+        # Create/Update stats
+        HostMetricSummaryMonthly.objects.bulk_create(self.records_to_create, batch_size=1000)
+        HostMetricSummaryMonthly.objects.bulk_update(self.records_to_update, ['license_consumed', 'hosts_added', 'hosts_deleted'], batch_size=1000)
+
+    def _get_license_consumed_before(self, month):
+        license_consumed = 0
+        for metric_month, metric in self.host_metrics.items():
+            if metric_month < month:
+                hosts_added = metric.get('hosts_added', 0)
+                hosts_deleted = metric.get('hosts_deleted', 0)
+                license_consumed = license_consumed + hosts_added - hosts_deleted
+            else:
+                break
+        return license_consumed
+
+    def _load_existing_summaries(self):
+        """Find all summaries newer than host metrics delete threshold"""
+        self.existing_summaries = HostMetricSummaryMonthly.objects.filter(date__gte=self._get_first_month()).order_by('date')
+        self.existing_summaries_idx = 0
+        self.existing_summaries_cnt = len(self.existing_summaries)
+
+    def _load_hosts_added(self):
+        """
+        SELECT 1 as id, date_trunc('month', first_automation) as month,
+               count(first_automation) AS hosts_added
+        FROM main_hostmetric
+        GROUP BY month
+        ORDER by month;
+        """
+        result = (
+            HostMetric.objects.annotate(month=TruncMonth('first_automation'))
+            .values('month')
+            .annotate(hosts_added=Count('first_automation'))
+            .values('month', 'hosts_added')
+            .order_by('month')
+        )
+
+        for host_metric in list(result):
+            month = host_metric['month']
+            if month:
+                beginning_of_month = datetime.date(month.year, month.month, 1)
+                if self.host_metrics.get(beginning_of_month) is None:
+                    self.host_metrics[beginning_of_month] = {}
+                self.host_metrics[beginning_of_month]['hosts_added'] = host_metric['hosts_added']
+
+    def _load_hosts_deleted(self):
+        """
+        SELECT 1 as id, date_trunc('month', last_deleted) as month,
+               count(last_deleted) AS hosts_deleted
+        FROM main_hostmetric
+        WHERE deleted = True
+        GROUP BY month
+        ORDER by month;
+        """
+        result = (
+            HostMetric.objects.annotate(month=TruncMonth('last_deleted'))
+            .values('month')
+            .annotate(hosts_deleted=Count('last_deleted'))
+            .values('month', 'hosts_deleted')
+            .filter(deleted=True)
+            .order_by('month')
+        )
+        for host_metric in list(result):
+            month = host_metric['month']
+            if month:
+                beginning_of_month = datetime.date(month.year, month.month, 1)
+                if self.host_metrics.get(beginning_of_month) is None:
+                    self.host_metrics[beginning_of_month] = {}
+                self.host_metrics[beginning_of_month]['hosts_deleted'] = host_metric['hosts_deleted']
+
+    def _find_or_create_summary(self, month):
+        summary = self._find_summary(month)
+
+        if not summary:
+            summary = HostMetricSummaryMonthly(date=month)
+            self.records_to_create.append(summary)
+        else:
+            self.records_to_update.append(summary)
+        return summary
+
+    def _find_summary(self, month):
+        """
+        Existing summaries are ordered by month ASC.
+        This method is called with month in ascending order too => only 1 traversing is enough
+        """
+        summary = None
+        while not summary and self.existing_summaries_idx < self.existing_summaries_cnt:
+            tmp = self.existing_summaries[self.existing_summaries_idx]
+            if tmp.date < month:
+                self.existing_summaries_idx += 1
+            elif tmp.date == month:
+                summary = tmp
+            elif tmp.date > month:
+                break
+        return summary
+
+    def _update_summary(self, summary, month, license_consumed):
+        # Get month counts from host metrics, zero if not found
+        hosts_added, hosts_deleted = 0, 0
+        if metric := self.host_metrics.get(month, None):
+            hosts_added = metric.get('hosts_added', 0)
+            hosts_deleted = metric.get('hosts_deleted', 0)
+
+        summary.license_consumed = license_consumed + hosts_added - hosts_deleted
+        summary.hosts_added = hosts_added
+        summary.hosts_deleted = hosts_deleted
+
+        # Set subscription count for current month
+        if month == datetime.date.today().replace(day=1):
+            license_info = get_license()
+            summary.license_capacity = license_info.get('instance_count', 0)
+        return summary
+
+    @staticmethod
+    def _get_first_month():
+        """Returns first month after host metrics hard delete threshold"""
+        threshold = getattr(settings, 'CLEANUP_HOST_METRICS_HARD_THRESHOLD', 36)
+        return datetime.date.today().replace(day=1) - relativedelta(months=int(threshold) - 1)
--- a/awx/main/tasks/handlers/host_metric_summary_monthly_task.py
+++ b/awx/main/tasks/handlers/host_metric_summary_monthly_task.py
@@ -4,21 +4,24 @@ from dateutil.relativedelta import relativedelta
 from django.conf import settings
 from django.db.models import Count
 from django.db.models.functions import TruncMonth
+from django.utils.timezone import now
 from awx.main.models.inventory import HostMetric, HostMetricSummaryMonthly
 from awx.conf.license import get_license
 
 
 class HostMetricSummaryMonthlyTask:
     """
-    This task computes last <threshold> months of HostMetricSummaryMonthly table
+    This task computes last [threshold] months of HostMetricSummaryMonthly table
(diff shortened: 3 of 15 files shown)


awx/settings/defaults.py
```diff
@@ -1052,4 +1053,6 @@
 # Host metrics cleanup - minimal interval between two cleanups in days
 CLEANUP_HOST_METRICS_INTERVAL = 30  # days
 # Host metrics cleanup - soft-delete HostMetric records with last_automation < [threshold] (in months)
-CLEANUP_HOST_METRICS_THRESHOLD = 12  # months
+CLEANUP_HOST_METRICS_SOFT_THRESHOLD = 12  # months
+# Host metrics cleanup - delete HostMetric record with deleted=True and last_deleted < [threshold]
```
AlanCoding: In practice, I ran `awx-manage host_metric_summary_monthly` and it created 36 records (going back to 2020), which is determined by this setting. I can't get from what this comment says to this behavior.

awx/main/tasks/handlers/host_metric_summary_monthly_task.py
```diff
@@ -0,0 +1,169 @@
...
+               count(first_automation) AS hosts_added
+        FROM main_hostmetric
+        GROUP BY month
+        ORDER by month;
+        """
+        result = (
+            HostMetric.objects.annotate(month=TruncMonth('first_automation'))
+            .values('month')
+            .annotate(hosts_added=Count('first_automation'))
+            .values('month', 'hosts_added')
```
AlanCoding: This is written kind of weirdly to my eye. Let's combine some operations:

```python
        result = (
            HostMetric.objects.annotate(month=TruncMonth('first_automation'), hosts_added=Count('first_automation'))
            .values('month', 'hosts_added')
            .order_by('month')
        )
```

awx/main/tasks/handlers/host_metric_summary_monthly_task.py
```diff
@@ -0,0 +1,169 @@
...
+        FROM main_hostmetric
+        WHERE deleted = True
+        GROUP BY month
+        ORDER by month;
+        """
+        result = (
+            HostMetric.objects.annotate(month=TruncMonth('last_deleted'))
+            .values('month')
+            .annotate(hosts_deleted=Count('last_deleted'))
+            .values('month', 'hosts_deleted')
```
AlanCoding: again, let's combine the annotate / values operations.

awx/settings/defaults.py
slemrmartin: This setting is primarily to ADR chapter 4.3 (not in this PR), but I get the point, I'll add some text for this feature too

awx/main/tasks/handlers/host_metric_summary_monthly_task.py
slemrmartin: @AlanCoding it doesn't produce the same sql, tests fail with this query

awx/main/tasks/handlers/host_metric_summary_monthly_task.py
slemrmartin: the same, it doesn't produce the same query

awx/main/tasks/handlers/host_metric_summary_monthly_task.py
AlanCoding: You can get the query by:

```python
print(result.query)
```

This produces

```sql
SELECT DATE_TRUNC(MONTH, "main_hostmetric"."first_automation" AT TIME ZONE UTC) AS "month",
       COUNT("main_hostmetric"."first_automation") AS "hosts_added"
FROM "main_hostmetric"
GROUP BY "main_hostmetric"."id",
         1
ORDER BY 1 ASC
```

The difference that my modification introduces is `GROUP BY "main_hostmetric"."id", 1` as opposed to your query which has `GROUP BY 1` only. I do not know what high-level objective this affects.

slemrmartin: my query produces:
```sql
SELECT DATE_TRUNC('month', "main_hostmetric"."first_automation" AT TIME ZONE 'UTC') AS "month", 
       COUNT("main_hostmetric"."first_automation") AS "hosts_added" 
FROM "main_hostmetric" 
GROUP BY DATE_TRUNC('month', "main_hostmetric"."first_automation" AT TIME ZONE 'UTC') 
ORDER BY "month" ASC
```
your query produces:
```sql
SELECT DATE_TRUNC('month', "main_hostmetric"."first_automation" AT TIME ZONE 'UTC') AS "month", 
       COUNT("main_hostmetric"."first_automation") AS "hosts_added" 
FROM "main_hostmetric" 
GROUP BY "main_hostmetric"."id", DATE_TRUNC('month', "main_hostmetric"."first_automation" AT TIME ZONE 'UTC') 
ORDER BY "month" ASC
```
so the `group by id` causes no grouping

AlanCoding: When I run your query it produces

```sql
SELECT DATE_TRUNC(MONTH, "main_hostmetric"."first_automation" AT TIME ZONE UTC) AS "month",
       COUNT("main_hostmetric"."first_automation") AS "hosts_added"
FROM "main_hostmetric"
GROUP BY 1
ORDER BY 1 ASC
```

That is, I run this in `awx-manage shell_plus`

```python
print((
HostMetric.objects.annotate(month=TruncMonth('first_automation'))
.values('month')
.annotate(hosts_added=Count('first_automation'))
.values('month', 'hosts_added')
.order_by('month')
).query)
```

slemrmartin: I don't know what to say - I printed it directly from the `awx-manage host_metric_summary_monthly` - if they look like your query, tests definitely won't work. The intention is to produce summaries by months

slemrmartin: also when I run it from `awx-manage shell_plus` it produces what I wrote

slemrmartin: `GROUP BY 1` is a valid SQL construction and is interchangeable with 
`GROUP BY DATE_TRUNC('MONTH', "main_hostmetric"."first_automation" AT TIME ZONE 'UTC')`.
It means "Group by 1st column". 

That's why it could produce different sqls for me and for you. And also it's an answer why it's important to keep it as it is.

AlanCoding: Yeah, the construction of the query is making sense to me now. Thanks. For the record, I still don't see why it would produce different sqls for me and you, unless you're running a different version of Django. Even if they are equivalent.

slemrmartin: good question. I'm running it in the docker container tools_awx_1, so there should be the Django from requirements.txt

awx/settings/defaults.py
```diff
@@ -468,6 +468,7 @@
     'send_subsystem_metrics': {'task': 'awx.main.analytics.analytics_tasks.send_subsystem_metrics', 'schedule': timedelta(seconds=20)},
     'cleanup_images': {'task': 'awx.main.tasks.system.cleanup_images_and_files', 'schedule': timedelta(hours=3)},
     'cleanup_host_metrics': {'task': 'awx.main.tasks.system.cleanup_host_metrics', 'schedule': timedelta(days=1)},
+    'host_metric_summary_monthly': {'task': 'awx.main.tasks.system.host_metric_summary_monthly', 'schedule': timedelta(weeks=1)},
```
AlanCoding: To offer full disclosure, I want to make it clear that the scheduler does not start a schedule on any kind of absolute or invariant time reference.

https://github.com/ansible/awx/blob/aafd4df28876f7b0d3479653a2fb77cb382e2383/awx/main/dispatch/periodic.py#L56

Rephrasing... whenever the dispatcher starts, the `cleanup_host_metrics` job will be set to run 1 day in the future. You can inspect and verify with shell_plus the item `scheduler.jobs[-1].next_run` from that startup loop I link.

My only implication is that when services are restarted, this new `host_metric_summary_monthly` task will be set to run 1 week in the future. This is true for all nodes, as all control plane nodes will run this every week, starting 1 week after their startup.

That's not a blocker for me, I just want you to know and consider if there are any negative consequences for your own goals.

slemrmartin: Does it mean first run is not immediatelly after restart but after week? 
If so, then I'll shorten the interval and check the interval also in the task (like gather-analytics)

awx/main/conf.py
```diff
@@ -831,6 +831,15 @@
     category_slug='system',
 )
 
+register(
+    'HOST_METRIC_SUMMARY_TASK_LAST_TS',
+    field_class=fields.DateTimeField,
+    label=_('Last computing date of HostMetricSummaryMonthly'),
+    allow_null=True,
+    category=_('System'),
+    category_slug='system',
```
AlanCoding, line 871: Let me recap this - `HOST_METRIC_SUMMARY_TASK_LAST_TS` is set simply whenever the weekly task is ran.

Following that technically, that should be surfaced as a read_only setting. This is problematic, as we have a lot of debt around read_only settings, as you often see hard-coded keys next to read_only checks...

https://github.com/ansible/awx/blob/aafd4df28876f7b0d3479653a2fb77cb382e2383/awx/conf/settings.py#L360

If you want to look at other examples using read_only settings, here are some:

```
In [27]: [key for key in settings.registry.get_registered_settings() if settings.registry.get_setting_field(key).read_only]
Out[27]: 
['INSTALL_UUID',
 'DEFAULT_CONTROL_PLANE_QUEUE_NAME',
 'DEFAULT_EXECUTION_QUEUE_NAME',
 'IS_K8S',
 'PENDO_TRACKING_STATE',
 'AUTHENTICATION_BACKENDS',
 'SOCIAL_AUTH_GOOGLE_OAUTH2_CALLBACK_URL',
 'SOCIAL_AUTH_GITHUB_CALLBACK_URL',
 'SOCIAL_AUTH_GITHUB_ORG_CALLBACK_URL',
 'SOCIAL_AUTH_GITHUB_TEAM_CALLBACK_URL',
 'SOCIAL_AUTH_GITHUB_ENTERPRISE_CALLBACK_URL',
 'SOCIAL_AUTH_GITHUB_ENTERPRISE_ORG_CALLBACK_URL',
 'SOCIAL_AUTH_GITHUB_ENTERPRISE_TEAM_CALLBACK_URL',
 'SOCIAL_AUTH_AZUREAD_OAUTH2_CALLBACK_URL',
 'SOCIAL_AUTH_SAML_CALLBACK_URL',
 'SOCIAL_AUTH_SAML_METADATA_URL']
```

I'm not sure what "read only" actually means in all these cases. In some cases, it means that the installer gives it a value in a file. In your case here, you mean for the task system to set it, but present it as read-only in the UI (you don't want users to change it right?)

I would find it preferable to deal with that in a separate model (we use a solo model for the user job scheduler last run) and present in the views specific to this feature. This is mainly because of our existing debt with exiting abuse of settings for purposes like these.

slemrmartin: I tried read only before and it means I can't update it even in the code. 

Can you describe more what do you mean by the solo model? Btw the same pattern is here for 2 more tasks. 
And we should keep in mind that this feature should be backportable to AAP 2.4 -> no new migrations. 
(we could move the discussion to Slack's working group channel)

awx/settings/defaults.py
slemrmartin: I changed this to run the schedule each several hours and check desired interval in the task.

awx/main/conf.py
slemrmartin: anyway, althought it's not intended to have these settings in the UI, it's perfectly safe if user changes it (i.e. with recommendation of RH support), now it can change the interval between two runs

awx/main/models/inventory.py
```diff
@@ -910,7 +910,7 @@ def cleanup_task(cls, months_ago):
 
 class HostMetricSummaryMonthly(models.Model):
     """
-    HostMetric summaries computed by scheduled task <TODO> monthly
+    HostMetric summaries computed by scheduled task 'awx.main.tasks.system.host_metric_summary_monthly' monthly
```
AlanCoding, line 913: this module path `awx.main.tasks.system.host_metric_summary_monthly` looks like it's old, now that you put it into a new `handlers/` folder inside of the task folder. I don't love the idea of introducing a folder there, because what does "handlers" mean, and why is it applicable here? However, I do strongly support keeping separate modules, so like `awx/main/tasks/host_metric_summary_monthly_task.py` would make the most sense to me, personally.

awx/main/conf.py
AlanCoding: I see that this began in https://github.com/ansible/awx/pull/13809, so it's not very much in the scope of this review.

awx/main/models/inventory.py
slemrmartin: the module path is not old, it wasn't here at all, anyway it's part of the same feature/ADR, just splitted to two releases. 

---

I made tasks.host_metrics module. I'll move, rename and extend the cleanup_host_metrics tasks in follow-up PR (according to the remaining ADR chapters, which contain 2 host metrics scheduled tasks)

awx/main/tasks/host_metrics.py
```diff
@@ -0,0 +1,202 @@
...
+            Setting.objects.filter(key='HOST_METRIC_SUMMARY_TASK_LAST_TS').first(),
+            getattr(settings, 'HOST_METRIC_SUMMARY_TASK_INTERVAL', 7) * 86400
+    ):
+        logger.info("TASK STATUS: host_metric_summary_monthly: STARTED")
+        HostMetricSummaryMonthlyTask().execute()
+        logger.info("TASK STATUS: host_metric_summary_monthly: FINISHED")
+
+
+def _is_run_threshold_reached(setting, threshold_seconds):
+    from rest_framework.fields import DateTimeField
```
AlanCoding: I don't see a particular reason why this import needs to be in-line. If putting it at the top of the file works, that would be the more standard practice.

AlanCoding: you could also obtain the field from the settings registry, demo:

```python
In [5]: settings.registry.get_setting_field('CLEANUP_HOST_METRICS_LAST_TS')
Out[5]: DateTimeField(allow_null=True, label='Last cleanup date for HostMetrics')
```

That would replace `DateTimeField()` in the current use. Either way, just trying to help this along.

awx/main/tasks/host_metrics.py
```diff
@@ -0,0 +1,202 @@
...
+from awx.main.models.inventory import HostMetric, HostMetricSummaryMonthly
+from awx.conf.license import get_license
+
+logger = logging.getLogger('awx.main.tasks.host_metric_summary_monthly')
+
+
+@task(queue=get_task_queuename)
+def host_metric_summary_monthly():
+    """Run cleanup host metrics summary monthly task each week"""
+    from awx.conf.models import Setting
```
AlanCoding: this will probably fail linters without a NOQA flag as well. If the model is needed, I expect you probably do need an in-line import to avoid circular imports. Before you do that, I would like to ask if the alternative would work to just reference `settings.HOST_METRIC_SUMMARY_TASK_LAST_TS`.

awx/main/tasks/host_metrics.py
```diff
@@ -0,0 +1,202 @@
...
+
+    if _is_run_threshold_reached(
+            Setting.objects.filter(key='HOST_METRIC_SUMMARY_TASK_LAST_TS').first(),
+            getattr(settings, 'HOST_METRIC_SUMMARY_TASK_INTERVAL', 7) * 86400
+    ):
+        logger.info("TASK STATUS: host_metric_summary_monthly: STARTED")
```
AlanCoding: Might be confusing log, since _other_ logs will say that it's starting the `host_metric_summary_monthly` task when the time criteria is not met. I'd suggest slight modification like

```python
logger.info(f"Executing host_metric_summary_monthly, last ran at {settings.HOST_METRIC_SUMMARY_TASK_LAST_TS}")
```

Otherwise the level and everything like that seems appropriate :+1:

AlanCoding:
:+1: all discussions are pretty well resolved aside from some linter failure related items, so after those get cleaned up we'll be looking at final verification and merging.

AlanCoding (approved):
wrapping up
 - review discussions resolved
 - checks are good
 - integration test results reviewed

djyasin (approved):

awx/main/tasks/host_metrics.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def _load_existing_summaries(self):
+        """Find all summaries newer than host metrics delete threshold"""
+        self.existing_summaries = HostMetricSummaryMonthly.objects.filter(date__gte=self._get_first_month()).order_by('date')
+        self.existing_summaries_idx = 0
+        self.existing_summaries_cnt = len(self.existing_summaries)
+
+    def _load_hosts_added(self):
+        """
+        SELECT date_trunc('month', first_automation) as month,
```
jjwatt: Seeing these SQL strings as the first thing in the """""", which is usually used for "docstrings" is jarring to me. Can you add descriptions to these like the other methods as the first line in the docstring (e.g., L97) and explain that the sql blocks are examples of SQL queries that describe the function? Alternatively, you could quote these a different way so that Python tools do not think they're docstrings.

awx/main/tasks/host_metrics.py
```diff
@@ -0,0 +1,198 @@
...
+            if tmp.date < month:
+                self.existing_summaries_idx += 1
+            elif tmp.date == month:
+                summary = tmp
+            elif tmp.date > month:
+                break
+        return summary
+
+    def _update_summary(self, summary, month, license_consumed):
+        # Get month counts from host metrics, zero if not found
```
jjwatt, line 185: nit: consistency. Make this a docstring like L163

awx/main/tasks/system.py
```diff
@@ -381,16 +376,25 @@ def cleanup_images_and_files():
 
 @task(queue=get_task_queuename)
 def cleanup_host_metrics():
+    """TODO: move to host_metrics in follow-up PR
```
jjwatt: nit: Do not use docstring for the TODO. Use a docstring to describe the method on the first line, adding the TODO somewhere in the body of the docstring OR use a normal comment. Consistency, too.

AlanCoding:
I have one minor bug fix I want to add to this branch:

https://github.com/slemrmartin/awx/pull/6

I'll wait for you to look over that, because I don't see a particular rush. After that I can work on next steps for shipping this.

awx/settings/defaults.py
```diff
@@ -1046,4 +1047,12 @@
 # Host metrics cleanup - minimal interval between two cleanups in days
 CLEANUP_HOST_METRICS_INTERVAL = 30  # days
 # Host metrics cleanup - soft-delete HostMetric records with last_automation < [threshold] (in months)
-CLEANUP_HOST_METRICS_THRESHOLD = 12  # months
+CLEANUP_HOST_METRICS_SOFT_THRESHOLD = 12  # months
```
jjwatt, line 1058: nit: I don't know how this, `CLEANUP_HOST_METRICS_THRESHOLD`, is used, and I don't know the typical practice in AWX, but from general principles it does not seem like this variable name MUST change, right? You could add `HARD_THRESHOLD` and leave `CLEANUP_HOST_METRICS_THRESHOLD` for backwards compatibility, and you could still add `*_SOFT_THRESHOLD` if you make sure it's always the same as the old `CLEANUP_HOST_METRICS_THRESHOLD`. You could then drop the unadorned varname, `CLEANUP_HOST_METRICS_THRESHOLD`, later on. It is probably not that important in this case, but in general I would avoid changing the settings names when possible.

jjwatt (changes requested):
Reviewed for Python readability. I have a couple of nitpicks around comments, etc.

awx/settings/defaults.py
slemrmartin: I would really not recommend that. Keeping just "threshold" and "hard_threshold" looks like the hard one will subset of the current one. But it's not true and will be confusing. 
Also this PR is part of the same ADR where the constant was introduced and will be backported. 
So to your question, yes, this variable MUST change because it'll become confusing other way

awx/main/tasks/host_metrics.py
slemrmartin: This is not description of method, but comment for next few lines. 
But I'll add docstring too, ok

jjwatt (approved):

AlanCoding:
Ready to rebase and merge.

PR #14004, opened 2023-05-15
Skip constructed_inventory in a more correct loop

AlanCoding:
##### SUMMARY
This revises what was done in https://github.com/ansible/awx/pull/13977

Yes, we do want to skip constructed inventories _in a sense_ but that was not the right place to do it...

At the point we loop over `self._dependent_resources()`, we have some type established by the class of `self`, like job templates. So we're performing `awx import --job-templates` and `self` is the job templates page. Then that loop should hit all the dependent resources of JTs, like projects, credentials, inventory.

The problem with that loop is that if you _skip_ constructed inventory, you _also_ skip inventory. This is because of an implementation detail of `_dependent_resources`, which stashes everything into a dictionary where the _keys_ are the _type_. The `/api/v2/inventories/` and `/api/v2/constructed_inventories/` are _pages_ but they both have the same inventory type. So if you allow the _inner_ loop to hit constructed inventories, then the _constructed inventory page_ gets associated with the _inventory type_.

We don't want to skip a dependent resource _type_ (because a JT would probably give "inventory" as a dependency... however that's done), we just want the _inventory type_ to be associated with the inventory page. To do that, we skip the constructed inventory page when registering pages to the type.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awxkit/awxkit/api/pages/api.py
+++ b/awxkit/awxkit/api/pages/api.py
@@ -253,7 +253,13 @@ class ApiV2(base.Base):
     # Import methods
 
     def _dependent_resources(self):
-        page_resource = {getattr(self, resource)._create().__item_class__: resource for resource in self.json}
+        page_resource = {}
+        for resource in self.json:
+            # The /api/v2/constructed_inventories endpoint is for the UI but will register as an Inventory endpoint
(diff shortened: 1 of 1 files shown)


john-westcott-iv (approved):
Great catch, thanks for fixing this!

PR #14005, opened 2023-05-15
[rsyslog] Enable disk-assisted queuing on output

relrod:
##### SUMMARY

Right now we only enable queuing on the rsyslog main_queue. This adds a
parameter to also enable it on the omhttp output action. As omhttp can
take time to process messages (e.g. blocking on the result of its HTTP
requests), this allows for queuing messages up and hopefully preventing
some messages from getting lost when the log server is slow to respond.

##### ISSUE TYPE

##### COMPONENT NAME

 - API


AlanCoding (approved):

gamuniz (approved):

awx/main/utils/external_logging.py
```diff
@@ -78,6 +79,10 @@ def escape_quotes(x):
             'action.resumeRetryCount="-1"',
             'template="awx"',
             f'action.resumeInterval="{timeout}"',
+            'queue.filename="awx-external-logger-action-queue"',
```
relrod, line 83: Needs `queue.spoolDirectory` too

PR #14007, opened 2023-05-15
[collection] Fix sanity tests on ansible-core 2.15

relrod:
##### SUMMARY

- Two new ignores (for unused imports which are used as feature flag tests)
- One legitimate fix, we documented a default that was no longer a default.
- Add 2.16 ignores in case people run tests using ansible devel

##### ISSUE TYPE

##### COMPONENT NAME

 - Collection


gundalow (approved):
Thanks for fixing this.

PR #14008, opened 2023-05-15
Add dynamically configurable debug settings

gamuniz:
##### SUMMARY

Add debug to configurable settings
Add UI elements to render in the UI

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 - UI

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```


awx/main/conf.py
```diff
@@ -831,6 +831,56 @@
     category_slug='system',
 )
 
+register(
+    'AWX_CALLBACK_PROFILE',
+    field_class=fields.BooleanField,
+    label=_('Debug callback timing'),
+    default=False,
+    help_text=_('Debug callback code timing'),
```
AlanCoding: This has a lot more explaining to do than "debug callback code timing", because this enables listening for the `SIGUSR1` signal, and if it receives it, it will then start profiling. But this means they need to shell in, find the process of interest, and send the signal.

But even _if_ they are competent enough to do that, I don't think this will work right, because the signal is registered at the start of the `work_loop` only if the setting is turned on. So then you still have to restart services for this to work.

awx/main/conf.py
```diff
@@ -831,6 +831,56 @@
...
+    category=('Debug'),
+    category_slug='debug',
+)
+
+register(
+    'DEFAULT_CONTAINER_RUN_OPTIONS',
+    field_class=fields.StringListField,
+    label=_('Container Run Options'),
+    default=['--network', 'slirp4netns:enable_ipv6=true'],
+    help_text=_('List of options to pass to podman run'),
```
AlanCoding: I think it would be fine to suggest `--debug` in the help text, if that's the idea of the typical use.

AlanCoding (approved):

mabashian (approved):

jay-steurer (approved):

jay-steurer:
Verified by QE API. I recommend getting a QE UI approval. On the API side is good to go.

PR #14018, opened 2023-05-17
[troubleshooting] Configurable filtering of high-volume logs

AlanCoding:
##### SUMMARY
The pitch here is to make DEBUG level logging useful.

Before this, if you switched log level to debug, you got swamped with periodic messages. After this, the default behavior will be relatively quiet, and then you will have to enable certain things with the new setting to get the unusually noisy logs again.

This still needs some UI work to show the setting in the UI and allow for editing it there.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/api/permissions.py
+++ b/awx/api/permissions.py
@@ -120,23 +120,17 @@ class ModelAccessPermission(permissions.BasePermission):
         return result
 
     def has_permission(self, request, view, obj=None):
+        response = self.check_permissions(request, view, obj)
         logger.debug(
-            'has_permission(user=%s method=%s data=%r, %s, %r)',
+            'has_permission(user=%s method=%s, %s, %r) returned %r',
             request.user,
             request.method,
-            request.data,
             view.__class__.__name__,
             obj,
+            response,
             extra={'volume_tag': 'rbac'},
         )
-        try:
-            response = self.check_permissions(request, view, obj)
-        except Exception as e:
-            logger.debug('has_permission raised %r', e, exc_info=True, extra={'volume_tag': 'rbac'})
-            raise
-        else:
-            logger.debug('has_permission returned %r', response, extra={'volume_tag': 'rbac'})
-            return response
+        return response
 
     def has_object_permission(self, request, view, obj):
         return self.has_permission(request, view, obj)
--- a/awx/api/permissions.py
+++ b/awx/api/permissions.py
@@ -120,14 +120,22 @@ class ModelAccessPermission(permissions.BasePermission):
         return result
 
     def has_permission(self, request, view, obj=None):
-        logger.debug('has_permission(user=%s method=%s data=%r, %s, %r)', request.user, request.method, request.data, view.__class__.__name__, obj)
+        logger.debug(
+            'has_permission(user=%s method=%s data=%r, %s, %r)',
+            request.user,
+            request.method,
+            request.data,
+            view.__class__.__name__,
+            obj,
+            extra={'volume_tag': 'rbac'},
+        )
         try:
             response = self.check_permissions(request, view, obj)
         except Exception as e:
-            logger.debug('has_permission raised %r', e, exc_info=True)
+            logger.debug('has_permission raised %r', e, exc_info=True, extra={'volume_tag': 'rbac'})
             raise
         else:
-            logger.debug('has_permission returned %r', response)
+            logger.debug('has_permission returned %r', response, extra={'volume_tag': 'rbac'})
             return response
 
     def has_object_permission(self, request, view, obj):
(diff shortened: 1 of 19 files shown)


gamuniz (approved):
looks good to me, if/when we merge the fact gathering logging we could include it here.

AlanCoding:
I have become more, and more, and more, unhappy with this implementation. My issue is that its purpose is so close to `LOG_AGGREGATOR_LOGGERS` that keeping both of would create a great deal of confusion. We faced a similar dilemma with `LOG_AGGREGATOR_LEVEL`, and in the great wisdom of our product manager at the time (and must against my own judgement at the time), we changed it to adjust the level of _both_ the external logger and the "normal" (stdout or file) logs. Time has proven that judgement call to be very good.

There's still some awkwardness figuring out how to do that... and it might not really work. I'm just still thinking about it.

In any case, system_tracking and activity_stream should be things you can turn on for normal logs, and I want some kind of unified system for that.

awx/main/scheduler/task_manager_models.py
```diff
@@ -277,7 +277,6 @@ def init_with_consumed_capacity(cls, **kwargs):
             from awx.main.models import UnifiedJob
 
             if instance_group_queryset is not None:
-                logger.debug("******************INSTANCE GROUP QUERYSET PASSED -- FILTERING TASKS ****************************")
```
AlanCoding, line 280: @TheRealHaoLiu this is a log that should have stood out to us, but I don't think it ever did because it was never configured like it should have been.

awx/main/analytics/broadcast_websocket.py
```diff
@@ -20,7 +20,7 @@
 BROADCAST_WEBSOCKET_REDIS_KEY_NAME = 'broadcast_websocket_stats'
 
 
-logger = logging.getLogger('awx.analytics.broadcast_websocket')
+logger = logging.getLogger('awx.main.analytics.broadcast_websocket')
```
AlanCoding, line 23: This appears to have been a mistake. There's only 1 log that uses this, and it's fairly incoherent to sell as one of the loggers for external logging. Clearly, it wasn't intended. This change puts it back in its place as a normal system log.

awx/api/permissions.py
```diff
@@ -120,15 +120,17 @@ def check_permissions(self, request, view, obj=None):
...
+        logger.debug(
+            'has_permission(user=%s method=%s, %s, %r) returned %r',
+            request.user,
+            request.method,
+            view.__class__.__name__,
+            obj,
+            response,
+            extra={'volume_tag': 'rbac'},
+        )
+        return response
```
AlanCoding, line 133: These are logs that are disabled by the LOGGING setting, and I'm bringing them in from the cold - letting people turn it on with the `LOG_AGGREGATOR_LOGGERS` setting here.

awx/main/signals.py
```diff
@@ -422,6 +422,7 @@ def emit_activity_stream_change(instance):
             object1=instance.object1,
             object2=instance.object2,
             summary_fields=summary_fields,
+            volume_tag='activity_stream',
```
AlanCoding, line 425: these logs are still not ideal, because they only contain the _message_ without the _data_, but I will leave the formatting problems associated with that for another day.

awx/main/tasks/system.py
```diff
@@ -508,7 +508,6 @@ def inspect_execution_nodes(instance_list):
 
 @task(queue=get_task_queuename, bind_kwargs=['dispatch_time', 'worker_tasks'])
 def cluster_node_heartbeat(dispatch_time=None, worker_tasks=None):
-    logger.debug("Cluster node heartbeat task.")
```
AlanCoding, line 511: The dispatcher task code logs when it's starting a task at DEBUG level, so this is pointless.

awx/main/utils/filters.py
```diff
@@ -83,39 +83,41 @@ def filter(self, record):
...
-            return False
-        if record.name.startswith('awx.analytics'):
-            base_path, headline_name = record.name.rsplit('.', 1)
-            return bool(headline_name in loggers)
-        else:
-            if '.' in record.name:
-                base_name, trailing_path = record.name.split('.', 1)
-            else:
-                base_name = record.name
-            return bool(base_name in loggers)
```
AlanCoding, line 105: Essentially this is moved from `ExternalLoggerEnabled` to `DynamicLevelFilter`, but we keep the "awx" logger check here as an exception specific to the external logger.

awx/settings/development.py
```diff
@@ -29,8 +29,6 @@
 # show colored logs in the dev environment
 # to disable this, set `COLOR_LOGS = False` in awx/settings/local_settings.py
 LOGGING['handlers']['console']['()'] = 'awx.main.utils.handlers.ColorHandler'  # noqa
-# task system does not propagate to AWX, so color log these too
-LOGGING['handlers']['task_system'] = LOGGING['handlers']['console'].copy()  # noqa
```
AlanCoding: this wrote task system messages twice in dev environment.

AlanCoding:
too complex to communicate, best to abandon

PR #14021, opened 2023-05-18
Upgrade psycopg2 to psycopg3

john-westcott-iv:
##### SUMMARY

Replace psycopg2 with psycopg3.

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - Other

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```


AlanCoding:
This library looks pretty similar to what we were already doing

https://github.com/rpkilby/jsonfield/blob/master/src/jsonfield/fields.py

but probably fills in all the correct gaps. Makes sense to me. This might be really close to being ready.

awx/main/dispatch/worker/callback.py
```diff
@@ -191,7 +192,7 @@ def flush(self, force=False):
                             e._retry_count = retry_count
 
                             # special sanitization logic for postgres treatment of NUL 0x00 char
-                            if (retry_count == 1) and isinstance(exc_indv, ValueError) and ("\x00" in e.stdout):
+                            if (retry_count == 1) and isinstance(exc_indv, DataError) and ("\x00" in e.stdout):
```
AlanCoding: Looks like this is the cause of

```
[gw0] linux -- Python 3.9.16 /var/lib/awx/venv/awx/bin/python3.9
  File "/awx_devel/awx/main/tests/functional/commands/test_callback_receiver.py", line 140, in test_postgres_invalid_NUL_char
    assert "\x00" not in events[0].stdout
AssertionError: assert '\x00' not in '\x00'
  '\x00' is contained here:
```

failure, which should be trivial to resolve

john-westcott-iv: Yea, it seems like two sources are seeing different raises. When I change between `ValueError` adn `DataError` two different tests fail depending on which is in here. So I simplified the if condition to remove the `isinstance` check and just wipe out the `\x00` if its there to handle both errors. Everything is green now so hopefully we can move forward.

Let me know if this is ok with you or if you think we should tack down why the two classes are seeing different things.

AlanCoding: looks fine, this sort of thing seems expected.

awx/main/fields.py
```diff
@@ -67,10 +67,61 @@ def __enum_validate__(validator, enums, instance, schema):
...
+            #     >>> string = "\x00"
+            #     >>> "\u0000" in string
+            #     True
+            dumped_value = json.dumps(value)
+            if "\x00" in dumped_value:
+                dumped_value = dumped_value.replace("\x00", '')
+            return dumped_value
+        except Exception as e:
+            logger.error(f"Failed to dump JSONField {self.name}: {e}")
+            logger.error(f"{value}")
```
AlanCoding: you should combine this into the prior log message. I'd rather have multi-line logs than logs without context.

john-westcott-iv: Fixed.

awx/main/dispatch/__init__.py
```diff
@@ -64,9 +64,9 @@ def events(self, select_timeout=5, yield_timeouts=False):
                 if yield_timeouts:
                     yield None
             else:
-                self.conn.poll()
-                while self.conn.notifies:
-                    yield self.conn.notifies.pop(0)
+                notification_generator = self.conn.notifies()
+                for notification in notification_generator:
```
AlanCoding, line 68: I was trying to read

https://www.psycopg.org/psycopg3/docs/advanced/async.html#asynchronous-notifications

But it looks like you're done it and gotten the key pattern change here. Little weird to save `notification_generator`, but this is all fine. This is a documented change from the library, and is cleaner than before.

awx/main/migrations/0185_move_JSONBlob_to_JSONField.py
```diff
@@ -0,0 +1,142 @@
...
+    dependencies = [
+        ('main', '0184_django_indexes'),
+        ('conf', '0010_change_to_JSONField'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='activitystream',
+            name='deleted_actor',
+            field=models.JSONField(null=True),
```
AlanCoding, line 17: Did this conversion not require a data migration? You could migrate locally and check the functionality of a single field and I would be satisfied. Right now I just don't see that documented anywhere (you may have told me and I forgot), so I want to see it written down.

awx/main/fields.py
```diff
@@ -67,10 +67,60 @@ def __enum_validate__(validator, enums, instance, schema):
...
+            except Exception as e:
+                logger.error(f"Failed to load JSONField {self.name}: {e}")
+
+        return value
+
+    def get_db_prep_value(self, value, connection, prepared=False):
+        if not prepared:
+            value = self.get_prep_value(value)
+        try:
+            # Null characters are not allowed in text fields and JSONBlobs are JSON data but saved as test
```
AlanCoding: test --> text

awx/main/models/unified_jobs.py
```diff
@@ -1137,11 +1137,9 @@ def result_stdout_raw_handle(self, enforce_max_bytes=True):
                     if total > max_supported:
                         raise StdoutMaxBytesExceeded(total, max_supported)
 
-                # psycopg2's copy_expert writes bytes, but callers of this
+                # psycopg3's copy writes bytes, but callers of this
                 # function assume a str-based fd will be returned; decode
                 # .write() calls on the fly to maintain this interface
```
AlanCoding: but you're not doing what this comment describes anymore now, are you?

(maybe move that comment down to your new `smart_str` use)

awx/main/wsrelay.py
```diff
@@ -209,51 +209,55 @@ def __init__(self):
...
+                        ip = hostname
+                    if ip is None:
+                        logger.warning(f"Received invalid offline ws_heartbeat, missing hostname and ip: {payload}")
+                        return
+                    self.cleanup_offline_host(ip)
+                    logger.debug(f"Web host {hostname} ({ip}) offline heartbeat received.")
+            except Exception as e:
+                # This catch-all is the same as the one above. asyncio will eat the exception
+                # but we want to know about it.
+                logger.exception(f"on_ws_heartbeat exception: {e}")
```
AlanCoding, line 252: _requests review from @relrod_

relrod: Yeah, I wrote this patch with @john-westcott-iv a few weeks ago. It should be reasonable. (If you view the github diff and tell it to ignore whitespace, it gives a clearer picture of the changes in that file.)

:+1: here (on wsrelay -- I haven't reviewed the rest)

AlanCoding (approved):

awx/main/analytics/collectors.py
```diff
@@ -399,7 +399,10 @@ def _copy_table(table, query, path):
     file_path = os.path.join(path, table + '_table.csv')
     file = FileSplitter(filespec=file_path)
     with connection.cursor() as cursor:
-        cursor.copy_expert(query, file)
+        with cursor.copy(query) as copy:
+            while data := copy.read():
+                byte_data = bytes(data)
```
jjwatt, line 404: optional/nit: Couldn't `byte_data.decode()` fail here? Is it OK that this would throw an exception from inside those two context managers? I don't see a context manager for the file, either, but maybe `FileSplitter()` tries to take care of that.

awx/main/dispatch/__init__.py
jjwatt: Yeah, agree this is fine. it will probably hold that reference that it closes over as long as the generator has more next()s to yield, but as long as it won't use all the memory, it would be fine. It wouldn't be any different if you kept it anonymous, either: i.e., `self.conn.notifies()` is the real iterator that would get closed over, and so `notification_generator` is just an alias, iiuc.

jay-steurer (approved):

awx/main/fields.py
```diff
@@ -67,10 +67,60 @@ def __enum_validate__(validator, enums, instance, schema):
...
+            #     >>> my_obj = { 'test': '\x00' }
+            #     >>> import json
+            #     >>> json.dumps(my_obj)
+            #     '{"test": "\\u0000"}'
+            # But just to be safe, lets remove them if they are there. \x00 and \u0000 are the same:
+            #     >>> string = "\x00"
+            #     >>> "\u0000" in string
+            #     True
+            dumped_value = json.dumps(value)
+            if "\x00" in dumped_value:
```
jjwatt, line 116: Note: Depending on how big dumped_value is and how often this function runs, I might consider not even checking--just always run `.replace("\x00", '')`. A `replace()` is idempotent if it isn't there, right? I know it might not be a no-op if it's not there, but it might be depending on the implementation in Python!

awx/main/wsrelay.py
```diff
@@ -209,51 +209,55 @@ def __init__(self):
...
+                        return
+                    self.known_hosts[hostname] = ip
+                    logger.debug(f"Web host {hostname} ({ip}) online heartbeat received.")
+                elif payload.get("action") == "offline":
+                    hostname = payload.get("hostname")
+                    ip = payload.get("ip")
+                    if ip is None:
+                        # If we don't get an IP, just try the hostname, maybe it resolves
+                        ip = hostname
+                    if ip is None:
```
jjwatt: nit: seeing double. I would make this clearer/less surprising. Either like, `if ip is None: if hostname: ip = hostname` or even `if hostname is None` for the last one, but that's debatable.

jjwatt (approved):

awx/main/wsrelay.py
john-westcott-iv: @relrod, do you want to change this?

awx/main/fields.py
john-westcott-iv: Yea, I debated that for a little. The last two JSONBlob fields we have are events and artifacts. Artifacts are set by users so can be random in size. Events are generated from ansible output and can also be variable in size. As an example, I was testing with a playbook that would cat files and then set them as artifacts. So the size of the events/artifact would be a file length.

Due to the number of events we process we want that code to be as performant as possible. I didn't do any timing myself but based on a [stack overflow](https://stackoverflow.com/questions/42059754/python-replace-function-should-i-check-first-if-what-i-want-to-replace-is-a-s) having the `if` is supposed to be more performant especially since I expect the decoded string to really never have `\x00` in it at all. This is a complete safety net to attempt to prevent an exception.

awx/main/analytics/collectors.py
john-westcott-iv: It could but I'm not sure what we would do if we caught an exception here. We are getting data back from the database and if we can't decode it for some reason I'd think whatever we are doing should probably stop.

@slemrmartin any thoughts on this?

awx/main/migrations/0185_move_JSONBlob_to_JSONField.py
john-westcott-iv: From sqlmigrate:
```
--
-- Alter field survey_spec on jobtemplate
--
ALTER TABLE "main_jobtemplate" ALTER COLUMN "survey_spec" TYPE jsonb USING "survey_spec"::jsonb;
```

Pre migration:
```
awx=# select unifiedjobtemplate_ptr_id, survey_spec from main_jobtemplate where unifiedjobtemplate_ptr_id = 7;
 unifiedjobtemplate_ptr_id |                                                                                                                                     survey_spec                                                                      >
---------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------->
                         7 | {"name": "", "spec": [{"max": 1024, "min": 0, "type": "multiplechoice", "choices": ["production", "development"], "default": "", "required": true, "variable": "environment", "new_question": true, "question_name": >
(1 row)
```

Post migration:
```
awx=# select unifiedjobtemplate_ptr_id, survey_spec from main_jobtemplate where unifiedjobtemplate_ptr_id = 7;
 unifiedjobtemplate_ptr_id |                                                                                                                                     survey_spec                                                                                     >
---------------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------->
                         7 | {"name": "", "spec": [{"max": 1024, "min": 0, "type": "multiplechoice", "choices": ["production", "development"], "default": "", "required": true, "variable": "environment", "new_question": true, "question_name": "Environment", >
(1 row)
```

So the data "looks" the same but it should now be a jsonb instead of a text field so we can do "nifty" JSON selects:
```
select unifiedjobtemplate_ptr_id, survey_spec->>'spec' from main_jobtemplate WHERE survey_spec->>'name' = 'Get the zrelease';
 unifiedjobtemplate_ptr_id |                                                                                                    ?column?                                                                                                    
---------------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
                        62 | [{"type": "text", "required": true, "variable": "release_version", "question_name": "What is the zrelease to branch?", "question_description": "Enter the version to zrelease in the format <major>.<minor>"}]
(1 row)
```

awx/main/wsrelay.py
jjwatt: It's a nit, though. Fine leaving it for now.

awx/main/models/unified_jobs.py
john-westcott-iv: Done.

jjwatt (approved):

PR #14025, opened 2023-05-19
Add `help_text` for all* fields and add enforcement check

AlanCoding:
##### SUMMARY
This adds a test that loops over all views, and then for every view that has a model in our app, it loops over the model fields, subtracts exceptions, and then will fail if any field has blank help text. So that explains the asterisk about the definition of "all" here.

I wrote the help text for everything here, applying knowledge from my own memory but I had to search the code to understand the behavior in a good handful of cases.

This is used in API OPTIONS, so we get stuff like this:

```json
            "job_explanation": {
                "type": "string",
                "label": "Job explanation",
                "help_text": "A status field to indicate the state of the job if it wasn't able to run and capture stdout",
                "filterable": true
            },
            "execution_node": {
                "type": "string",
                "label": "Execution node",
                "help_text": "The node the job executed on.",
                "filterable": true
            },
            "controller_node": {
                "type": "string",
                "label": "Controller node",
                "help_text": "The instance that managed the execution of this job.",
                "filterable": true
            },
            "launched_by": {
                "type": "field",
                "label": "Launched by",
                "filterable": false
            },
            "work_unit_id": {
                "type": "string",
                "label": "Work unit id",
                "help_text": "The Receptor work unit ID associated with this job.",
                "filterable": true
```

You can see here that `launched_by` still has no help_text. This is because it is a serializer field and the method here fails to capture that. Someone could delete my test and do it better if they want, but I believe just filling in the help text will have lasting value, and this will prevent new fields with no help text in the future.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - Docs


awx/main/constants.py
```diff
@@ -14,6 +14,18 @@
     'STANDARD_INVENTORY_UPDATE_ENV',
 ]
 
+NAME_HELP_TEXT = _('Name of object to be displayed in list views. Also commonly used by clients for lookups.')
+MANAGED_HELP_TEXT = _('Indicates whether this resource is created and controled automatically by server logic.')
```
gamuniz: typo on controlled.

awx/main/constants.py
```diff
@@ -14,6 +14,18 @@
...
+DESCRIPTION_HELP_TEXT = _('Extra information about this resource to show in its detail view.')
+
+# Job and ad hoc type field help_text
+JOB_TYPE_HELP_TEXT = _('If doing a check job the ansible-playbook --check option will be used.')
+DIFF_MODE_HELP_TEXT = _('If enabled, textual changes made to any templated files on the host are shown in the standard output')
+LIMIT_HELP_TEXT = _('Limit inventory to subset of hosts, passed to ansible-playbook as --limit')
+FORKS_HELP_TEXT = _('Output verbosity, passed to ansible-playbook as --limit')
```
gamuniz: copy paste fail here?

awx/main/constants.py
```diff
@@ -14,6 +14,18 @@
...
+VERBOSITY_HELP_TEXT = _('Number of parallel processes to use, passed to ansible-playbook as --forks')
```
gamuniz: same here

awx/main/fields.py
```diff
@@ -901,7 +901,10 @@ class AskForField(models.BooleanField):
     Denotes whether to prompt on launch for another field on the same template
     """
 
+    fmt_string = _('If true, prompt for the corresponding job field when a user manually ' 'launches this template. Also controls whether ')
```
AlanCoding: Black concatenation strikes again!

AlanCoding:
This idea lived on in https://github.com/ansible/django-ansible-base/pull/647

And the test should then ideally make use of the new command which we may already have, which is `awx-manage help_text_check --applications=main --ignore-file=...` so we should try that.

AlanCoding:
closing because there's a DAB utility that should be used for this

PR #14026, opened 2023-05-19
Move pid files and sockets to /tmp

### LINKED ISSUE #13951: Launch scripts don't work as non-root

### Bug Summary

In each of the launch scripts ([launch_awx_rsyslog.sh](https://github.com/ansible/awx/blob/devel/tools/ansible/roles/dockerfile/files/launch_awx_rsyslog.sh), [launch_awx_task.sh](https://github.com/ansible/awx/blob/devel/tools/ansible/roles/dockerfile/files/launch_awx_task.sh), [launch_awx_web.sh](https://github.com/ansible/awx/blob/devel/tools/ansible/roles/dockerfile/files/launch_awx_web.sh)), there's this block:

```bash
if [ `id -u` -ge 500 ]; then
    echo "awx:x:`id -u`:`id -g`:,,,:/var/lib/awx:/bin/bash" >> /tmp/passwd
    cat /tmp/passwd > /etc/passwd
    rm /tmp/passwd
fi
```

But the logic doesn't make sense to me. If our UID is greater than 500 then we don't have root privileges so we can't edit `/etc/passwd`. For us this presents as `/usr/bin/launch_awx_task.sh: line 4: /etc/passwd: Permission denied` in the container logs when running as UID = 1000. @rooftopcellist  Could you explain what this block is there to accomplish? It looks like it was added by you in #9289.

### AWX version

22.0.0

- [ ] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [X] Other

### Installation method

kubernetes

### Modifications

no

### Ansible version

_No response_

### Operating system

_No response_

### Web browser

_No response_

### Steps to reproduce

Deploy AWX with a security context that defines `runAsUser` and `runAsGroup`. Here's a script that deploys awx fresh and shows the issue.

```
#!/usr/bin/env bash

cat << EOF > kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  - github.com/ansible/awx-operator/config/default?ref=2.0.1

images:
  - name: quay.io/ansible/awx-operator
    newTag: 2.0.1

namespace: awx
EOF

kubectl apply -k .

cat << EOF > awx.yaml
---
apiVersion: awx.ansible.com/v1beta1
kind: AWX
metadata:
  name: awx
  namespace: awx
spec:
  image_pull_policy: Always
  service_type: nodeport
  security_context_settings:
    runAsGroup: 1000
    runAsUser: 1000
EOF

kubectl apply -f awx.yaml
```

After the awx-task pod comes up you can see the error in the awx-task, awx-rsyslog and awx-ee containers
```
> k logs -n awx awx-task-5c6d49b7fb-lglgx -c awx-task
/usr/bin/launch_awx_rsyslog.sh: line 4: /etc/passwd: Permission denied
[wait-for-migrations] Waiting for database migrations...
[wait-for-migrations] Attempt 1 of 30
[wait-for-migrations] Waiting 0.5 seconds before next attempt
[wait-for-migrations] Attempt 2 of 30
...
```

### Expected results

AWX runs properly as a non-root user

### Actual results

3 (awx-ee, awx-task, awx-rsyslog) of the 4 containers in the pod fail to start with similar errors `/usr/bin/launch_awx_task.sh: line 4: /etc/passwd: Permission denied`. The redis container starts up normally.

### Additional information

_No response_

ToxicGLaDOS:
Fixes #13951

Users without root access didn't have permission to `/var/run` and `/var/lib/awx`

##### SUMMARY

`/var/lib/awx/rsyslog`, `/var/run/supervisor`,  and `/var/run/awx-rsyslog` are owned by `root:root` and aren't world writable, so we can't create the sockets and temp files there unless we're the root user or in the root group. While `/tmp` isn't necessarily a conventional place to put these things (except `tempfile.TemporaryDirectory(...)` which probably should've always been there), we need a world writable place to create these files if we want to support non-root access.

An alternative could be to `chmod` `/var/lib/awx/rsyslog`, `/var/run/supervisor`, and `/var/run/awx-rsyslog` to be world writable.

##### ISSUE TYPE

##### COMPONENT NAME

 - Other

##### AWX VERSION

```
awx: 22.2.1.dev22+gfa05f55512.d20230519
```

##### ADDITIONAL INFORMATION

To reproduce you just need to set the security context to a non-root user and group. For example:

```
---
apiVersion: awx.ansible.com/v1beta1
kind: AWX
metadata:
  name: awx
  namespace: awx
spec:
  security_context_settings:
    runAsUser: 1000
    runAsGroup: 1000
```

As a side note I noticed that the awx-web pod doesn't get the security context applied, this seems to be a separate bug that I'll file an issue for separately.

### DIFF
--- a/awx/main/utils/external_logging.py
+++ b/awx/main/utils/external_logging.py
@@ -119,10 +119,10 @@ def construct_rsyslog_conf_template(settings=settings):
 def reconfigure_rsyslog():
     tmpl = construct_rsyslog_conf_template()
     # Write config to a temp file then move it to preserve atomicity
-    with tempfile.TemporaryDirectory(dir='/var/lib/awx/rsyslog/', prefix='rsyslog-conf-') as temp_dir:
+    with tempfile.TemporaryDirectory(dir='/tmp/', prefix='rsyslog-conf-') as temp_dir:
         path = temp_dir + '/rsyslog.conf.temp'
         with open(path, 'w') as f:
             os.chmod(path, 0o640)
             f.write(tmpl + '\n')
-        shutil.move(path, '/var/lib/awx/rsyslog/rsyslog.conf')
+        shutil.move(path, '/tmp/rsyslog.conf')
     supervisor_service_command(command='restart', service='awx-rsyslogd')
--- a/tools/ansible/roles/dockerfile/files/launch_awx_rsyslog.sh
+++ b/tools/ansible/roles/dockerfile/files/launch_awx_rsyslog.sh
@@ -20,7 +20,7 @@ wait-for-migrations
 # This file will be re-written when the dispatcher calls reconfigure_rsyslog(),
 # but it needs to exist when supervisor initially starts rsyslog to prevent the
 # container from crashing. This was the most minimal config I could get working.
-cat << EOF > /var/lib/awx/rsyslog/rsyslog.conf
+cat << EOF > /tmp/rsyslog.conf
 action(type="omfile" file="/dev/null")
 EOF
 
--- a/tools/ansible/roles/dockerfile/templates/supervisor_rsyslog.conf.j2
+++ b/tools/ansible/roles/dockerfile/templates/supervisor_rsyslog.conf.j2
@@ -3,10 +3,10 @@ nodaemon = True
 umask = 022
 logfile = /dev/stdout
 logfile_maxbytes = 0
-pidfile = /var/run/supervisor/supervisor.rsyslog.pid
+pidfile = /tmp/supervisor.rsyslog.pid
 
 [program:awx-rsyslogd]
-command = rsyslogd -n -i /var/run/awx-rsyslog/rsyslog.pid -f /var/lib/awx/rsyslog/rsyslog.conf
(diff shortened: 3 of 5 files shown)


shanemcd (changes requested):
This is most likely going to break our downstream product. We'll need to think about this more deeply and try to find a better way of solving this problem. Requesting changes to prevent this from being accidentally merged. cc @rooftopcellist

ToxicGLaDOS:
> This is most likely going to break our downstream product. We'll need to think about this more deeply and try to find a better way of solving this problem.

Like I had mentioned, we could also `chmod` those select directories and files to be world writable. If that's not an option I can also dig into more structural changes, but at the end of the day we need a user with an unknown UID to be able to read and write the directory where we put the PID files and sockets.

ToxicGLaDOS:
@shanemcd Would the `chmod` solution I described above be acceptable to the downstream product? Or do you think there's a bigger picture solution that we should be trying to pursue?

PR #14034, opened 2023-05-23
Reduce connection usage by conventional means

AlanCoding:
##### SUMMARY
This reduces the number of connections used by the application. We can go over each here...

 - dispatcher main process close connection before starting its pg_notify listen (because the listening uses a _separate_ connection)
 - dispatcher workers close connection before listening to IPC queue for messages from main process
 - dispatcher periodic scheduler close connection before entering its `time.sleep` line
 - callback receiver closes connection before its 60 second sleep :hugs: 
 - Do not use a separate connection for run_clear_cache listening / ORM actions :hugs: 
 - Do not use a separate connection for run_rsyslog_configurer.py listening / ORM actions :hugs: 

I have marked the "safe" changes with :hugs: , with the intent that they can be backported, if desired.

What makes something safe? Take run_clear_cache, for example. This _consolidates_ 2 connections being used by a single process (no crazy threading issues). Since it will remain as a long-lived connection, there's no real trade-off. The safe changes get a benefit, but no apparent negative.

I've tried to put all "safe" changes in the first commit.

What makes something not-safe? If it results in re-opening the connection, then I'm calling it not safe. Closing, and then re-opening a connection results in lower memory use, but more overhead of re-opening the connection later, which is churn on the postgres server. Perhaps using pgbouncer can solve this problem as well, it's not yet known. On balance I am still pretty sure this is a good idea. After all, this behavior is _already_ what uWSGI does.

This includes and replaces https://github.com/ansible/awx/pull/13953

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/dispatch/worker/base.py
+++ b/awx/main/dispatch/worker/base.py
@@ -146,6 +146,7 @@ class AWXConsumerRedis(AWXConsumerBase):
 
         while True:
             logger.debug(f'{os.getpid()} is alive')
+            db.connection.close()
             time.sleep(60)
 
 
--- a/awx/main/dispatch/worker/base.py
+++ b/awx/main/dispatch/worker/base.py
@@ -181,6 +181,13 @@ class AWXConsumerPG(AWXConsumerBase):
             self.listen_cumulative_time = 0.0
             self.last_metrics_gather = current_time
 
+        # close all connections before entering the select on pg_notify to minimize conn use
+        for django_connection in db.connections.all():
+            django_connection.close()
+
+        self.pg_is_down = False
+        self.listen_start = time.time()
+
     def run(self, *args, **kwargs):
         super(AWXConsumerPG, self).run(*args, **kwargs)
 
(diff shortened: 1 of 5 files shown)


awx/main/dispatch/worker/base.py
```diff
@@ -180,6 +181,13 @@ def run_periodic_tasks(self):
             self.listen_cumulative_time = 0.0
             self.last_metrics_gather = current_time
 
+        # close all connections before entering the select on pg_notify to minimize conn use
+        for django_connection in db.connections.all():
```
jessicamack, line 185: This excludes the connection in use automatically?

awx/main/management/commands/run_cache_clear.py
```diff
@@ -18,7 +19,7 @@ class Command(BaseCommand):
 
     def handle(self, *arg, **options):
         try:
-            with pg_bus_conn(new_connection=True) as conn:
```
jessicamack, line 21: Since these services are running separately won't they have establish new connections anyway? At least that is my understanding from my previous PR.

AlanCoding: In the body here, it does a `logger.info`, and the way we structure our logging, that requires database access (if other logic before this didn't), so that will open up a new connection. This is essentially allowing those other miscellaneous ORM actions to use the same connection as what is used for listening to pg_notify. I wish my explanation was easier to follow here.

PR #14049, opened 2023-05-25
[WIP] [feature] Specify a jobs' name from its template

john-westcott-iv:
##### SUMMARY

Adds a `job name template` field to Job Templates and Workflow Job Templates. This defaults to the templates name but can be customized via Jinja2 to include other attributes of the job. For example maybe you want a job name template of:
```
{{ job.name }} - for environment {{ extra_vars.environment }}
```
If you had a job template with a name of `Patch Servers` and survey or extra var that defined an `environment` variable with a value of `PRODUCTION` a job launched with this would be named `Patch Servers - for environment PRODUCTION`. Other fields from the job template can be added.

The use of undefined variables an empty string. In the above example, if extra_vars did not contain an environment variable the name would be `Patch Servers - for environment  ('dict object' has no attribute 'environment')`.

TODO:
- [ ] Add unit/functional tests of API
- [ ] Add integration tests for collection modules
- [X] Do we want to change the UI to not add the `{{ job.id }} - ` in the jobs screen/bread crumb trail and instead set the default value to be `{{ job.id }} - {{ job.name }}` to maintain that behavior? No, in the Next UI it is a separate column that can be added/removed. 
- [x] See if we can alter the Jinja templating to put a value of `undefined` instead of empty string for a variable that can not be found.
- [ ] Add "global" default
- [ ] Workflow template IDs
- [ ] Change field name to be more descriptive
- [ ] What happens with job slicing
- [ ] What happens if a rendered template is larger than a TextField
- [ ] Can we perform a second template on job completion to allow for additional fields/variables to be included?

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API
 - UI
 - Collection

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```


awx/main/models/unified_jobs.py
```diff
@@ -427,6 +434,26 @@ def create_unified_job(self, instance_groups=None, **kwargs):
             activity_stream_create(None, unified_job, True)
         unified_job.log_lifecycle("created")
 
+        if parent_field_name:
+            job_name_template = getattr(getattr(unified_job, parent_field_name), 'job_name_template', None)
+            if job_name_template:
+                from jinja2 import Environment
+                from jinja2.exceptions import TemplateSyntaxError
+
+                j2_environment = Environment()
```
gamuniz: https://jinja.palletsprojects.com/en/3.1.x/api/#undefined-types

fosterseth:
some discussion came up about the name "Job Name Template". the word "Template" might be a little overloaded because this field is on the job template page.

Maybe it can just be called Job Name or Job Name Pattern or something. The help text box will probably suffice for emphasizing the jinja2 nature of it

fosterseth:
another thing to consider is length. Since the length can be dynamic and only resolved at runtime, then we may wish cap the overall length that the UI displays (by truncating)

Also is there a way to preview the job name without running a job? that might be a slick thing to add if possible

pHiney:
Why have a template ? Why not just allow for a name directly off the command line.

pHiney:
Anyway. +1 if this gets us the ability to name jobs

beumont2:
+1

Issue #14047, opened 2023-05-25
Import for existing users not adding all permissions with awx cli


ArtsiomMusin:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.
- [X] I am **NOT** reporting a (potential) security vulnerability. (These should be emailed to `security@ansible.com` instead.)

### Bug Summary

When a list of users is exported, all users are exported with permissions. However, when users are imported back, no permissions are added.

### AWX version

AAP

### Select the relevant components

- [ ] UI
- [ ] UI (tech preview)
- [ ] API
- [ ] Docs
- [ ] Collection
- [X] CLI
- [ ] Other

### Installation method

openshift

### Modifications

no

### Ansible version

n/a

### Operating system

n/a

### Steps to reproduce

1. Create a user and add some permissions for this user to some objects or orgs
2. Export the user
3. Remove permissions given on step 1
4. Import the user

### Expected results

All permissions are restored

### Actual results

No permissions are restored

PR #14053, opened 2023-05-26
Use patch to update users in awx cli

ArtsiomMusin:
##### SUMMARY

Related to #14047

When exporting users, they have roles exported as well. When importing back, users are updated with `put` which fails and roles are skipped and not assigned.

##### ISSUE TYPE

##### COMPONENT NAME

 - CLI

##### AWX VERSION

```
N/A
```

##### ADDITIONAL INFORMATION

Before the fix, if a user exists, awx cli tries to update it with `put` but it fails and role assignment is skipped
```
/api/v2/users/ "user@com": Bad Request (400) received - {'password': ['This password is too short. It must contain at least 15 characters.', 'This password is too common.']}.
```

After the fix, the user is patched and after that all roles are assigned.

### DIFF
--- a/awxkit/awxkit/api/pages/api.py
+++ b/awxkit/awxkit/api/pages/api.py
@@ -317,7 +317,10 @@ class ApiV2(base.Base):
                     if asset['natural_key']['type'] == 'project' and 'local_path' in post_data and _page['scm_type'] == post_data['scm_type']:
                         del post_data['local_path']
 
-                    _page = _page.put(post_data)
+                    if asset['natural_key']['type'] == 'user':
+                        _page = _page.patch(**post_data)
+                    else:
+                        _page = _page.put(post_data)
                     changed = True
             except (exc.Common, AssertionError) as e:
                 identifier = asset.get("name", None) or asset.get("username", None) or asset.get("hostname", None)


awxkit/awxkit/api/pages/api.py
```diff
@@ -281,7 +281,10 @@ def _import_list(self, endpoint, assets):
                     if asset['natural_key']['type'] == 'project' and 'local_path' in post_data and _page['scm_type'] == post_data['scm_type']:
                         del post_data['local_path']
 
-                    _page = _page.put(post_data)
+                    if asset['natural_key']['type'] == 'user':
+                        _page = _page.patch(**post_data)
```
ArtsiomMusin, line 321: tested as well with just `patch` for all objects and didn't notice any issues. So maybe this condition is not required and we can use just `patch`

anandpatial:
*** Does not fix ***
1) As part of issue# 332 I have created the awxkit and planted the provided script from Artsiom on the given path.
2) I have exported the user and then deleted the user
```
(.venv) ansible@WIN-NLR2FVP31G6:~/CACF/local_windows_CACF/issue332/issue332_awxkit$ awx -k export --users test_awx_user0 > aws_export_users.txt
(.venv) ansible@WIN-NLR2FVP31G6:~/CACF/local_windows_CACF/issue332/issue332_awxkit$ awx -k users delete test_awx_user0
(.venv) ansible@WIN-NLR2FVP31G6:~/CACF/local_windows_CACF/issue332/issue332_awxkit$ awx -k users list --username test_awx_user0
{
     "count": 0,
     "next": null,
     "previous": null,
     "results": []
}
```
3) Imported the user and it does not work 
```
(.venv) ansible@WIN-NLR2FVP31G6:~/CACF/local_windows_CACF/issue332/issue332_awxkit$ awx -k import < awx_export_users.txt 

/api/v2/users/ "test_awx_user0": Bad Request (400) received - {'password': ['This password is too short. It must contain at least 15 characters.', 'This password is too common.']}.
```

ArtsiomMusin:
@anandpatial you tested a different scenario.

Your scenario is:
1. Create a user
2. Export it
3. Delete it
4. Import it

My scenario and fix related to:
1. Create a user
2. Give some RBAC to it
3. Export the user
4. Remove the RBAC given on step 2
5. Import the user

Expected result is RBAC are restored and should match RBAC from step 2.

Also note that the issue was related to SSO users which are logged in and their RBAC should be restored.

anandpatial:
@ArtsiomMusin , right Artsiom , now I am going to test your scenario :  (will get back shortly)
Create a user
Give some RBAC to it
Export the user
Remove the RBAC given on step 2
Import the user

anandpatial:
**test case2: successfully tested (api.py worked fine)**
1)recreate the user
2) assign permissions
3) export user
4) remove permissions
5) import user

Removing the link as the artifacts contain Kyndryl specific nodes . The Artifacts are present in issue#332
Summary
{
Import command issued:

And basically, it took 12 minutes for the script execution.
The script is retrieving/reading almost all the objects from the tower.
22,026 debug messages generated of which the message at 22,025 indicates that the roles for the required user have been posted:
```
DEBUG:urllib3.connectionpool:XXXXXMASKED"POST /api/v2/users/2384/roles/ HTTP/1.1" 204 0
```

Observing that the export data produced do not have the “user id” rather the natural key data that indicates the username.
After searching for the username “test_awx_user0” in the Debug Log I see that at Debug line# 4418 it discovers this user by natural key and by Debug line # 4694 it has the user’s ID.
At line# 4694 we have the user endpoint, and we have the import list from the exported data file.

Query: If the exported data already contained userid, could the work till debug line # 4418 be reduced?
The script goes further and looks after “all” the teams, organizations, inventories, workflow_job_templates, credential_types, credentials, projects, inventory sources, job_templates, applications, schedules etc... (Basically, all EXPORTABLE_RESOURCES) and ultimately roles, and, by this time the debug line # 22,025 is reached.
}

anandpatial:
Query: for test case1: where the user object was deleted and then the user was imported :
  is the behaviour
```
Bad Request (400) received - {'password': ['This password is too short. It must contain at least 15 characters.', 'This password is too common.']}.
```
an expected behaviour because of following:

```
if _page is None:
    if asset['natural_key']['type'] == 'user':
        ## We should only impose a default password if the resource doesn't exist.
        post_data.setdefault('password', 'abc123')
```

ArtsiomMusin:
1. Yes the performance is not good and requires some optimisation. 
2. I also saw that it should be set to a default password but looks like the password requirements became stronger since the time this default password had been set.

It would be good that the awx team comment on these findings by @anandpatial

jjwatt:
> Import command issued:
> And basically, it took 12 minutes for the script execution.

This does concern me. I'll take a closer look there. EDIT: But, was it failing before? It's not a performance regression if it just started working :)

> At line# 4694 we have the user endpoint, and we have the import list from the exported data file.
Query: If the exported data already contained userid, could the work till debug line # 4418 be reduced?
> The script goes further and looks after “all” the teams, organizations, inventories, workflow_job_templates, credential_types, credentials, projects, inventory sources, job_templates, applications, schedules etc... (Basically, all EXPORTABLE_RESOURCES) and ultimately roles, and, by this time the debug line # 22,025 is reached.

That sounds reasonable, but I don't immediately see how the patch causes that, so I'll tap another AWX team member to comment on that and the other issue about default passwords (which I don't understand, either).

jjwatt:
@fosterseth Do you understand more of the context here? Do you know who would?

jbradberry:
@anandpatial It sounds to me like you are issuing an export command that isn't specifying which resources you want.  It is possible to select a subset of types or even specific resources of particular types.  I strongly recommend that for testing you limit yourself to only what is needed, or you are going to be iterating over all of the resources on your system, as you noted.

cc @jjwatt: if you export an entire system that has real data, 12 minutes seems unsurprising.

jbradberry:
@anandpatial

"This password is too short.":  Yes, that block of code is why that's happening.  We definitely should fix that.

awxkit/awxkit/api/pages/api.py
jbradberry: I can't think of any case where `patch` would be a problem.  I say go for it.

dmzoneill:
Marking as blocked as there seems to be open questions to be answered and other actions to be completed

jbradberry (approved):
In my opinion this is correct and we should merge it.

jbradberry:
The unrelated performance issues mentioned above are almost certainly dealt with by #15128.

Issue #13152, opened 2022-11-02
events are not deleted from _unpartitioned_* tables


anxstj:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

The "_unpartitioned_*" tables (e.g. "_unpartitioned_main_jobevent") which were created during the AWX 19 update (I think) are still containing a lot of data. The issue https://github.com/ansible/awx/issues/10341 and the code in https://github.com/ansible/awx/blob/devel/awx/main/management/commands/cleanup_jobs.py#L202 indicate that this shouldn't be the case and that the events should have been removed by now.

### AWX version

21.5.0

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

docker development environment

### Modifications

no

### Ansible version

2.12.2

### Operating system

Debian 11

### Web browser

Firefox

### Steps to reproduce

Haven't tried this, but upgrading an older System (e.g. AWX 18) to a  recent version should move the existing events to the "_unpartitioned_" tables. And then starting a cleanup job shouldn't remove the data from there.

### Expected results

The "_unpartioned_*" tables are empty.

### Actual results

There is a lot of old data in those tables:
```
awx=# select count(*) from _unpartitioned_main_jobevent where created > timestamp '2022-06-01 00:00:00' limit 10;
 count 
-------
     0
awx=# select count(id) from _unpartitioned_main_jobevent where created > timestamp '2022-04-01 00:00:00' limit 10;
  count   
----------
 51360577
awx=# select count(id) from _unpartitioned_main_jobevent where created < timestamp '2022-02-15 00:00:00' limit 10;
  count   
----------
 15384905
```

The scheduler is configured to cleanup all data that is older than 120 days: 

`curl https://awx.example.com/api/v2/schedules/1/`
```
...
    "created": "2018-10-17T12:13:09.566943Z",
    "modified": "2022-09-15T16:22:41.282216Z",
    "name": "Cleanup Job Schedule",
    "description": "Automatically Generated Schedule",
    "extra_data": {
        "days": 120
    },
...
```

### Additional information

[...]

Issue #10341, opened 2021-06-03
job event cleanup - Drop unpartitioned table when empty.


chrismeyersfsu:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
With the introduction of partitioned job event tables we keep the "old" unpartitioned job events via renaming the tables `main_jobevent` --> `_unpartitioned_main_jobevent`. Jobs that were ran before the partition migration are still accessible through this table. Cleanup jobs still take action on this table. Eventually, the old unpartitioned job events table will become empty. BUT, without a postgres full vacuum the storage will not be returned to the OS ("normal" autovacuum will only make the space available within the same table). As a nicety we should drop the old unpartitioned job events table when it becomes empty. The check for "is the table empty" must be done with care because counting the table can be expensive. Also note that code and queries may still rely on this table existing. There are two design decisions to be made (1) how and when to check if the unpartitioned table(s) are empty. (2) How to "drop" the old tables.

1a. (idea from @Ladas) After a cleanup job, if no jobs are cleaned up we can presume that the number of job events in the unpartitioned table is low. Do a count to verify. The counter argument to this, maybe the customer runs a cleanup for jobs older than 31 days and they have many unpartitioned jobs in the last 30 days. In this case, the count will still be very expensive.
1b. Maybe we get an approximate count and if it's < some number we can assume it's safe to do an absolute count `SELECT reltuples::BIGINT AS approximate_row_count FROM pg_class WHERE relname = 'job_events_fk'`
1c. Some other clever query i.e. `LIMIT 1` or a query on `pg_relation_size` `pg_table_size` to answer the question "Is there at least 1 record remaining?" if so, we know we can't drop the table.

2a.  Return disk space to customer via (in a single transaction) rename empty unpartitioned table(s), create new, empty table place of the renamed table, drop renamed table. This will require no (little) change to existing code.
2b. Return disk space to customer via dropping empty unpartitioned table(s). Modify code to deal with unpartitioned table(s) existing or not existing.
2c. In a future release, we may declare that "ok, all old job events should have been cleaned up by now". In this case, we delete all code that deals with old unpartitioned job events table(s) & drop the tables.

DLV1810:
Having just hit this issue myself, why can't the cleanup mgmt jobs look for this table and delete old records based on the retention time until its empty? - The logic should already be there from AAP 1.2 (Tower 3.8.6) just add task to the cleanup job and update the table names it references? eventually the table will have 0 records

[...]

PR #14055, opened 2023-05-26
DROP unnecessary unpartioned event tables

AlanCoding:
##### SUMMARY
Connect https://github.com/ansible/awx/issues/13152

> TODO:
> adjust cleanup_jobs to remove events (particularly from the unpartitioned tables) without necessarily having to tie them to an existing job

Connect https://github.com/ansible/awx/issues/10341

> Eventually, the old unpartitioned job events table will become empty. BUT, without a postgres full vacuum the storage will not be returned to the OS ("normal" autovacuum will only make the space available within the same table). As a nicety we should drop the old unpartitioned job events table when it becomes empty.

This PR carries out the "we should drop" part of that, but I have abandoned the suggested heuristics from that issue, the "when it becomes empty". Because, refer back to #13152, you can't count on it becoming empty because orphaned events clearly exist.

So the new heuristic this uses is "drop the table when it is obviously irrelevant". For that, I just check if `created` timestamps are all before the cutoff time given to `awx-manage cleanup_jobs`.

I believe that a full and simple DROP is the safest option. This command is extremely scale-hardened, and I don't want to get into any row-by-row things. I don't even like that this keeps the old logic for cascade deleting. There is a good argument that we should ditch that too.

I tested this by artificially creating an unpartitioned job event and re-running the system job.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/management/commands/cleanup_jobs.py
+++ b/awx/main/management/commands/cleanup_jobs.py
@@ -195,14 +195,35 @@ class Command(BaseCommand):
         delete_meta.delete_jobs()
         return (delete_meta.jobs_no_delete_count, delete_meta.jobs_to_delete_count)
 
-    def _cascade_delete_job_events(self, model, pk_list):
+    def _handle_unpartitioned_events(self, model, pk_list):
+        """
+        If unpartitioned job events remain, it will cascade those from jobs in pk_list
+        if the unpartitioned table is no longe necessary, it will drop the table
+        """
+        tblname = unified_job_class_to_event_table_name(model)
+        rel_name = model().event_parent_key
+        with connection.cursor() as cursor:
+            cursor.execute(f"SELECT 1 FROM pg_tables WHERE tablename = '_unpartitioned_{tblname}';")
+            row = cursor.fetchone()
+            if row is None:
+                self.logger.debug(f'Unpartitioned table for {rel_name} does not exist, you are fully migrated')
+                return
         if pk_list:
             with connection.cursor() as cursor:
-                tblname = unified_job_class_to_event_table_name(model)
-
                 pk_list_csv = ','.join(map(str, pk_list))
-                rel_name = model().event_parent_key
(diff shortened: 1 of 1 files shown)


AlanCoding:
Next steps after this would be to delete the `UnpartitionedJobEvent` model and similar models. Then we'll need to carry this logic inside of cleanup_jobs... even if we did so indefinitely I don't think that's a big deal. Eventually it becomes a simple "check if these tables exist and drop them if they do"

gamuniz (approved):

awx/main/management/commands/cleanup_jobs.py
```diff
@@ -195,14 +195,35 @@ def cleanup_workflow_jobs_partition(self):
         delete_meta.delete_jobs()
         return (delete_meta.jobs_no_delete_count, delete_meta.jobs_to_delete_count)
 
-    def _cascade_delete_job_events(self, model, pk_list):
+    def _handle_unpartitioned_events(self, model, pk_list):
+        """
+        If unpartitioned job events remain, it will cascade those from jobs in pk_list
+        if the unpartitioned table is no longe necessary, it will drop the table
```
jessicamack: typo in the comment should be "longer necessary"

jessicamack:
> Next steps after this would be to delete the `UnpartitionedJobEvent` model and similar models. Then we'll need to carry this logic inside of cleanup_jobs... even if we did so indefinitely I don't think that's a big deal. Eventually it becomes a simple "check if these tables exist and drop them if they do"

LGTM. Noted the one typo but that's no reason not to merge. It would be helpful to open a a backlog ticket for the further work to improve the cleanup_jobs task so it's not lost in the shuffle.

jessicamack (approved):

AlanCoding:
Yes, I filed https://github.com/ansible/awx/issues/14533 as a message to the future to do this.

jay-steurer (approved):

PR #14056, opened 2023-05-26
[hop_node] Validate listener_port is defined for peers

### LINKED ISSUE #13918: [hop node] ensure peers have tcp listeners
if adding a peer, make sure that target peer has a tcp listener port defined. I think we can do this as a serializer validation.

also check listener_port is defined when peers_from_control_nodes is set to true

as part of this work, listener_port should be optional (and default to none on model)

we'll also need to make changes to the install bundle for the optional listener

thedoubl3j:
##### SUMMARY

Fixes #13918 
When adding a peer to a node, the resulting instance that is the peer needs to have a listerner_port defined. This change adds validation so that a peer cannot be added without a listener_port to either an instance or the control node.

This change also includes logic for the install bundle. If you get the install bundle for an instance that does not have a listener_port defined, then the group_vars (all.yml) receptor_listener will be turned to false (default is true) and receptor_port will not be set. If a listener_port is defined, then receptor_listener will be set to true and the port will be set.

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
devel
```

##### ADDITIONAL INFORMATION

**Test Cases**
1 - listener_port validation 
- create 2 instances via the API, one with listener_port defined (task1) and the other without it defined (task2). 
- once both are created, attempt to add task2 as a peer to task1 (this should fail with a validation error)
- after that, attempt to add task1 as a peer to task2 (this should pass)
2 - install bundle 
- using the same instances above, download the install bundle from each instance and look at the all.yml (in the group_vars dir), receptor_listener should be true and receptor_port should be set in the bundle from task1 and receptor_listener should false and receport_port should not be set in the task2 bundle. 
3 - test peer_from_control_node validation
- using the same instances above, on each instance, flip the flag to true. Task1 should have no issue but task2 should fail.


awx/api/serializers.py
```diff
@@ -5496,6 +5497,9 @@ def get_field_from_model_or_attrs(fd):
                 raise serializers.ValidationError(
                     "Setting peers manually for control nodes is not allowed. Enable peers_from_control_nodes on the hop and execution nodes instead."
                 )
+                
+        if peers_from_control_nodes and listener_port == None:
```
fosterseth: `listener_port is None` might be preferred

also what happens when the input form has

```
"listener_port": ""
```
does it still throw the validation? I suspect it does, but worth checking

awx/api/serializers.py
```diff
@@ -5496,6 +5497,9 @@ def get_field_from_model_or_attrs(fd):
...
+            raise serializers.ValidationError("listener_port cannot be empty when peers_from_control_nodes is true.")
```
fosterseth: listener_port must be a valid integer when peers_from_control_nodes is enabled.

awx/main/migrations/0185_hop_nodes.py
```diff
@@ -32,4 +32,15 @@ class Migration(migrations.Migration):
...
         ),
+        migrations.AlterField(
+            model_name='instance',
+            name='listener_port',
+            field=models.PositiveIntegerField(
+                blank=True,
+                default=None,
+                help_text='Port that Receptor will listen for incoming connections on.',
+                null=True,
+                validators=[django.core.validators.MinValueValidator(1), django.core.validators.MaxValueValidator(65535)],
```
fosterseth, line 44: oh can you append this to the top of the migration file
```python3
import django.core.validators
from django.db import migrations, models
```

we are getting import errors in the api-test CI

fosterseth (approved):

fosterseth:
api lint error is complaining about a file that is not part of this PR, so we can ignore that and fix it in feature branch

PR #14061, opened 2023-05-30
Migration file to set peers_from_control_ nodes to true for existing execution nodes

djyasin:
Added function in hop nodes migration file to set peers_from_control_ nodes to true for existing execution nodes.

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 22.3.1
```

### DIFF
--- a/awx/main/migrations/0185_hop_nodes.py
+++ b/awx/main/migrations/0185_hop_nodes.py
@@ -4,6 +4,11 @@ import django.core.validators
 from django.db import migrations, models
 
 
+def set_peers_from_control_nodes_true(apps, schema_editor):
+    Instance = apps.get_model('main', 'Instance')
+    Instance.objects.filter(node_type='execution').update(peers_from_control_nodes=True)
+
+
 class Migration(migrations.Migration):
     dependencies = [
         ('main', '0184_django_indexes'),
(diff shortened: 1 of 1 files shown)


awx/main/migrations/0186_djy.py
```diff
@@ -0,0 +1,25 @@
...
+
+import logging
+
+logger = logging.getLogger('awx.main.migrations')
+
+
+def set_peers_from_control_nodes_true(apps, schema_editor):
+    Instance = apps.get_model('main', 'Instance')
+
+    for instance in Instance.objects.all():
```
fosterseth: this block can be written a little more succinctly by using objects.filter(..) followed by a .update(..)

awx/main/migrations/0186_djy.py
```diff
@@ -0,0 +1,25 @@
...
+
+logger = logging.getLogger('awx.main.migrations')
+
+
+def set_peers_from_control_nodes_true(apps, schema_editor):
+    Instance = apps.get_model('main', 'Instance')
+
+    Instance.objects.filter(node_type='execution').update(peers_from_control_nodes=True)
+
+    logger.info("set_peers_from_control_nodes_true")
```
fosterseth: we can remove the logger message here

fosterseth (approved):

PR #14067, opened 2023-06-01
Integrate scheduler into dispatcher main loop

AlanCoding:
##### SUMMARY
This replaces https://github.com/ansible/awx/pull/13990, and does the same thing. However, the scope is expanded significantly here, with the key goals being:

1. Avoid unnecessary use of `pg_notify` for local scheduling of tasks, as we may replace it sometimes or all the time in a future systems redesign
2. Reduce the database connections we use by 1 per node (the periodic scheduler obviously)
3. (as in prior PR) Spread out tasks to avoid unnecessary scale-up events

Right now, it has a lot of very verbose DEBUG level logs (we already had some, but this has more). However, these are all very relevant near-term for feature validation. The proof of the functionality is definitively in the logs, which show that things are making their schedules as expected.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
To talk out the process and decisions made here...

If we are to eliminate the periodic scheduler's fork (and its connection along with it), then we have to integrate the "time until next run" into the dispatcher listener `select_timeout`. The schedule library has a method that gives the "time until next run" but, oddly, we never used it, and just went by the most-frequent schedule. This is correct or no depending on the divisibility of your schedules, like if you ran every 2 minutes and every 4 minutes then it's fine. But if you ran every 2 minutes and 3 minutes the current code would not be smart enough to figure out it needs to run ever 1 minute. Aaaanyway...

The problem with using the schedule library's answer to this question is that we can't _also_ accomplish the task of spreading out tasks. See my prior PR on it, I used `time.sleep`s which won't work given the other expectations of the dispatcher. This is because it refused feature requests to "evenly" spread out secondly schedules. It only allows jumbling the schedule "offset" randomly. This may be because it introduces deeper philosophical problems, like how we have 12 schedules but only like 3 of them run on the shortest interval of 20 seconds, so you're interval is 20/12, and you only have 3 slots filled for each 20 second bin. This might bother a library maintainer, but it's 100% perfect, exactly what I was looking for to solve all the above-stated problems.


awx/main/dispatch/periodic.py
```diff
@@ -1,57 +1,72 @@
...
-        apply_async = TaskWorker.resolve_callable(task['task']).apply_async
-        total_seconds = task['schedule'].total_seconds()
-        scheduler.every(total_seconds).seconds.do(apply_async)
-    scheduler.run_continuously()
+class Job:
+    def __init__(self, name: str, data: dict):
+        self.name = name
+        self.data = data
+        self.interval = int(data['schedule'].total_seconds())
+        self.offset = None
```
AlanCoding: should change the default to `0` just in case

awx/main/dispatch/periodic.py
```diff
@@ -1,57 +1,72 @@
...
+        self.jobs = [Job(name, data) for name, data in schedule.items()]
+        min_interval = min(job.interval for job in self.jobs)
+        num_jobs = len(self.jobs)
+
+        # even space out jobs over the base interval
+        for i, job in enumerate(self.jobs):
+            job.offset = (i * min_interval) // num_jobs
+
+        # internally times are all referenced relative to startup time, add grace period
+        self.global_start = int(time.time() + 2)
```
AlanCoding: I've thought about this a lot, and changed my mind that `global_start` should stay as a float. Our "internal" time model should remain all as integer math, but we shouldn't pin the target run times to on-the-second times, because if all applications did this, it would create artificial bottlenecks.

awx/main/dispatch/periodic.py
```diff
@@ -1,57 +1,72 @@
...
+    @property
+    def next_run(self):
+        """
+        Gives the time until the next run with t=0 being the global_start
+        of the scheduler class
+        """
+        return (self.index + 1) * self.interval + self.offset
+
+    def mark_run(self, relative_time):
+        new_index = (relative_time - self.offset) // self.interval
```
AlanCoding: change expectation that `relative_time` should be a float, make this normal `/`, and take an `int` of it

AlanCoding:
I added a new debug command to go along with this, demo:

```
(awx) bash-5.1$ awx-manage run_dispatcher --schedule
2023-06-02 14:28:10,385 WARNING  [-] awx.main.dispatch checking dispatcher schedule for awx_1
title: Scheduler status
started_time: 2023-06-02 14:20:52 UTC
current_time: 2023-06-02 14:28:10 UTC
current_time_relative: 437.559
total_schedules: 12
schedule_list:
  task_manager:
    last_run_seconds_ago: 12.529
    next_run_in_seconds: 7.441
    offset_in_seconds: 5
    completed_runs: 21
    missed_runs: 0
  dependency_manager:
    last_run_seconds_ago: 11.558
    next_run_in_seconds: 8.441
    offset_in_seconds: 6
    completed_runs: 21
    missed_runs: 0
  send_subsystem_metrics:
    last_run_seconds_ago: 6.558
    next_run_in_seconds: 13.441
    offset_in_seconds: 11
    completed_runs: 21
    missed_runs: 0
  metrics_gather:
    last_run_seconds_ago: 19.51
    next_run_in_seconds: 0.441
    offset_in_seconds: 18
    completed_runs: 20
    missed_runs: 0
  tower_scheduler:
    last_run_seconds_ago: 17.521
    next_run_in_seconds: 12.441
    offset_in_seconds: 0
    completed_runs: 14
    missed_runs: 0
  cluster_heartbeat:
    last_run_seconds_ago: 16.557
    next_run_in_seconds: 43.441
    offset_in_seconds: 1
    completed_runs: 7
    missed_runs: 0
  k8s_reaper:
    last_run_seconds_ago: 9.517
    next_run_in_seconds: 50.441
    offset_in_seconds: 8
    completed_runs: 7
    missed_runs: 0
  receptor_reaper:
    last_run_seconds_ago: 7.508
    next_run_in_seconds: 52.441
    offset_in_seconds: 10
    completed_runs: 7
    missed_runs: 0
  pool_cleanup:
    last_run_seconds_ago: 1.547
    next_run_in_seconds: 58.441
    offset_in_seconds: 16
    completed_runs: 7
    missed_runs: 0
  gather_analytics:
    last_run_seconds_ago: 134.534
    next_run_in_seconds: 165.441
    offset_in_seconds: 3
    completed_runs: 1
    missed_runs: 0
  cleanup_images:
    last_run_seconds_ago: null
    next_run_in_seconds: 10375.441
    offset_in_seconds: 13
    completed_runs: 0
    missed_runs: 0
  cleanup_host_metrics:
    last_run_seconds_ago: null
    next_run_in_seconds: 85977.441
    offset_in_seconds: 15
    completed_runs: 0
    missed_runs: 0
```

awx/main/dispatch/__init__.py
```diff
@@ -73,7 +74,7 @@ def close(self):
 
 
 @contextmanager
-def pg_bus_conn(new_connection=False):
+def pg_bus_conn(new_connection=False, **kwargs):
```
relrod: Can we be explicit here, just for clarity instead of using unpacking?

awx/main/dispatch/periodic.py
```diff
@@ -1,57 +1,139 @@
...
-
-
-def run_continuously():
-    scheduler = Scheduler()
-    for task in settings.CELERYBEAT_SCHEDULE.values():
-        apply_async = TaskWorker.resolve_callable(task['task']).apply_async
-        total_seconds = task['schedule'].total_seconds()
-        scheduler.every(total_seconds).seconds.do(apply_async)
-    scheduler.run_continuously()
+class Job:
```
relrod: I feel like we risk overloading this word. Maybe `ScheduledTask`? If nothing else, it makes grepping easier for people later on.

awx/main/dispatch/worker/base.py
```diff
@@ -162,24 +171,46 @@ def __init__(self, *args, **kwargs):
...
-            # so that we scale down workers and free up connections
-            self.pool.cleanup()
-            self.last_cleanup = current_time
+        for job in self.scheduler.get_and_mark_pending():
+            if 'control' in job.data:
+                job.data['control']()
+            elif 'task' in job.data:
+                body = self.worker.resolve_callable(job.data['task']).get_async_body()
+                # bypasses pg_notify for scheduled tasks
+                self.dispatch_task(body)
```
relrod, line 214: This is really cool.

awx/main/dispatch/worker/base.py
```diff
@@ -195,14 +226,12 @@ def run(self, *args, **kwargs):
...
-                    self.listen_start = time.time()
+                    conn.select_timeout = self.run_periodic_tasks()
                     for e in conn.events(yield_timeouts=True):
                         self.listen_cumulative_time += time.time() - self.listen_start
                         if e is not None:
                             self.process_task(json.loads(e.payload))
-                        self.run_periodic_tasks()
-                        self.pg_is_down = False
-                        self.listen_start = time.time()
+                        conn.select_timeout = self.run_periodic_tasks()
```
relrod, line 243: So just to wrap my head around this (and adding a comment or two explaining this might be useful, because it's not super obvious to me):

- `conn.events` will block waiting on events
- If there's no event that comes through within `select_timeout` seconds, it will give us a `None` to work with and we will move on
- If we have a task (if we didn't get a `None`), we'll queue it
- Either way (whether or not we have a task), we'll run periodic tasks
- Then we change `select_timeout`, so that (assuming no events come in), we'll time out (and get a `None`) right at the next time a task would need to run -- to re-run this loop and run the scheduler logic again. And if a message comes in before then, we'll do the same thing but earlier.

I'd really like a nearby comment or something explaining this because it's really clever and not immediately obvious (but it's a cool solution).

relrod (commented):
This is so cool. Really creative solution.

awx/main/dispatch/__init__.py
AlanCoding: I just wish kwarg default handling was better. Because if `PubSub.__init__` has a default, and `pg_bus_conn` has a default, then that renders the default in `PubSub.__init__` meaningless because it's only ever called from here. I could do `, select_timeout=None` but then I'll have to manually check for the None case in `__init__` and apply the default then. Using `**kwargs` avoids this multiple-default problem.

awx/main/dispatch/worker/base.py
AlanCoding: For some related history...

Not long ago, we were not yielding `None` for timeouts in this loop. That means we only ran logic when we got a message. Problem is that there can be some cases (when the periodic scheduler fails) when no messages come in. For a lot of good reasons, in that case, you still want to run local dispatcher periodic tasks (cleanup), as these may be necessary to walk-back from that failure state.

This patch would leave us in a very different kind of landscape, where only _reactive_ tasks normally come through this mechanism, and no _periodic_ tasks. So previously we were maybe 50/50 sometimes timing out and sometimes getting a message. With this, we'll be like 99/1 with it almost always timing out unless someone starts a job.

AlanCoding:
Some details on impact of scale-up events here: https://github.com/ansible/awx/pull/13989

AlanCoding:
This broke when rebased on top of the psycopg upgrade, needs https://github.com/ansible/awx/pull/14222 to function now.

AlanCoding:
This has some tie-ins to https://github.com/psycopg/psycopg/issues/340 so I wanted to link that here.

TheRealHaoLiu (approved):

relrod (approved):

Issue #14064, opened 2023-05-31
Add an argument to control the API call frequency with --monitor or --wait


Flashh38:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

Enhancement to Existing Feature

### Feature Summary

I would like to propose the addition of a new argument in the AWX project to provide users with the ability to control the frequency of API calls when utilizing the --monitor or --wait options. Currently, these options trigger API calls every 0.25 seconds, which can lead to security alerts.

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [X] CLI
- [ ] Other

### Steps to reproduce

1. Launch a job using the AWX command-line interface.
2. Include the --monitor or --wait options in the command to enable monitoring or waiting for job completion.
3. Observe the API calls being made every 0.25 seconds.

### Current results

When using the AWX command-line interface with the --monitor or --wait options, the API calls are made every 0.25 seconds. This frequent API call behavior can trigger security alerts in certain environments.

### Sugested feature result

The proposed feature aims to introduce a customizable argument, such as --api-interval, that would allow users to define the desired interval between API calls when utilizing the --monitor or --wait options in AWX. By incorporating this feature, administrators would have the flexibility to adjust the frequency of API calls according to their specific environment requirements, thereby reducing the occurrence of security alerts and improving the overall user experience.

AlanCoding:
@gamuniz @jay-steurer closing as resolved with https://github.com/ansible/awx/pull/14068

PR #14069, opened 2023-06-01
Add polling interval argument for --monitor/--wait

hluk:
##### SUMMARY

Adds `--poll-interval` CLI flag indicating the interval between API calls when `--monitor` or `--wait` are used.

The default value is set to 2000ms (instead of the previous 250ms) to avoid overwhelming the server in case client does not set the interval explicitly.

The logs retrieval will be less smooth (unless user overrides the interval) but it makes the job monitoring much less likely to fail by overloading the server (which can respond for example with 502 Bad Request status code).

The minimum allowed value is 250ms.

related #14064

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - CLI

##### AWX VERSION

```
awx: 22.3.1.dev19+g1bd7770c30
```

##### ADDITIONAL INFORMATION

Monitoring jobs with a default small polling interval (which was 250 ms) causes server to get overloaded (assuming there are many clients) and in turn the monitoring process fails after it gets unexpected response from the server (for example with 502 Bad Request status code).

### DIFF
--- a/awxkit/awxkit/cli/custom.py
+++ b/awxkit/awxkit/cli/custom.py
@@ -1,7 +1,7 @@
 import functools
 import json
 
-from .stdout import monitor, monitor_workflow
+from .stdout import DEFAULT_POLL_INTERVAL_SECONDS, MINIMUM_POLL_INTERVAL_SECONDS, monitor, monitor_workflow
 from .utils import CustomRegistryMeta, color_enabled
 from awxkit import api
 from awxkit.exceptions import NoContent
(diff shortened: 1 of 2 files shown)


dangoncalves (commented):
IMHO it would be worth to have poll_interval in second instead of millisecond

hluk:
> IMHO it would be worth to have poll_interval in second instead of millisecond

Yeah, that would be more consistent with the units used by the other arguments.

Is it OK to make the argument value type `float` instead of `int`? Value `0.25` would be the minimum, `2.0` the default.

dangoncalves:
I understand your POV, but it would be better if we accept both float and int.
Minimum and default are fine for me.

hluk:
> I understand your POV, but it would be better if we accept both float and int.

I've changed the unit to seconds and the type to `float`. This accepts integers on command line as well - i.e. `--poll-interval=2` would be parsed as `2.0`.

dangoncalves (approved):

hluk:
I'm closing this in favor of #14068.

PR #14073, opened 2023-06-01
Add feature flag to allow instance API on non K8s environments

rcarrillocruz:
##### SUMMARY

It should be possible to add remote execution nodes via the API, 
regardless if AWX is running on K8s or not.

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```
0.1.dev33318+g1adcbe4
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -5435,7 +5435,7 @@ class InstanceSerializer(BaseSerializer):
         res = super(InstanceSerializer, self).get_related(obj)
         res['jobs'] = self.reverse('api:instance_unified_jobs_list', kwargs={'pk': obj.pk})
         res['instance_groups'] = self.reverse('api:instance_instance_groups_list', kwargs={'pk': obj.pk})
-        if obj.node_type in [Instance.Types.EXECUTION, Instance.Types.HOP]:
+        if (settings.FEATURE_API_INSTANCE_MANAGEMENT or settings.IS_K8S) and obj.node_type in (Instance.Types.EXECUTION, Instance.Types.HOP):
             res['install_bundle'] = self.reverse('api:instance_install_bundle', kwargs={'pk': obj.pk})
         res['peers'] = self.reverse('api:instance_peers_list', kwargs={"pk": obj.pk})
         if self.context['request'].user.is_superuser or self.context['request'].user.is_system_auditor:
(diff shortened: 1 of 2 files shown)


shanemcd:
Hey @rcarrillocruz - thanks for the PR. As we discussed yesterday, we aren't ready to enable this for downstream VM installs. I think the best approach here would be to put this code under a new explicit setting instead of using `IS_K8S`. Then you can toggle that for your use case.

rcarrillocruz:
What setting name do you suggest? INSTANCE_API_VALIDATE_IS_K8S , defaulting to True? I'm terrible at naming :-)

shanemcd:
We've been talking about feature flags for a while. Might be good to try to standardize on a naming convention. How about `FEATURE_API_INSTANCE_MANAGEMENT`? Default to `False`.

rcarrillocruz:
@shanemcd let me know, seems to work. Without adding the flag, will just validate and complain about not being on K8s (on my laptop, running on docker).
Setting the flag to True, I could create a dummy '1.1.1.1' instance from the Instance API.

TheRealHaoLiu:
https://github.com/TheRealHaoLiu/awx/blob/devel/awx/api/serializers.py/#L5434 this might also need to be changed

tpouyer:
@rcarrillocruz @shanemcd @TheRealHaoLiu where do we stand on this?  This is required in AAP 2.4 release in order for AoC Self-Managed to deliver partial Automation Mesh support in the next AoC Self-Managed release.

jay-steurer:
Hi @rcarrillocruz Could you please check the branch conflicts? Thanks!

TheRealHaoLiu:
in testing we found that if we go through the API to add the instance group it will trigger code like write_receptor_conf and establish link between nodes. currently we force the link to be outbound from control node to the new execution node this is undesirable

rcarrillocruz:
Can you comment what the next course of action for this patch?
Bear in mind in my use case I don't care if the connection is outbound, we are just considering the containerized installer to be able to create an execution node via API, get the bundle and playbooks and run it from a bastion host.

TheRealHaoLiu:
currently the assumption is that all node will be peered from the controlplane in development of hop node support we added a API field to explicitly specify if we peer from controlplane or not 

there may be more changes needed 

im experimenting with that right now

PR #14076, opened 2023-06-01
Add management command to precreate partitioned tables

gamuniz:
##### SUMMARY

Useful for automating a pg_dump as it holds exclusive locks on main_jobevent etc during a pg_dump

example of automation:
```
- name: precreate partitioned tables
  shell: awx-manage precreate_partitions
  become_user: awx
  become: true
  register: partitions
```

##### ISSUE TYPE

 - New or Enhanced Feature
 

##### COMPONENT NAME

 - API
##### AWX VERSION

```
make VERSION
awx: 22.3.1.dev22+gfdce6c9b35.d20230601

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/management/commands/precreate_partitions.py
+++ b/awx/main/management/commands/precreate_partitions.py
@@ -19,7 +19,7 @@ class Command(BaseCommand):
         while count > 0:
             for table in tables:
                 create_partition(table, start)
-                print(f'Creating partitions for {table} {start}')
+                print(f'Created partitions for {table} {start}')
             start = start + timedelta(hours=1)
             count -= 1
 
--- a/awx/main/management/commands/precreate_partitions.py
+++ b/awx/main/management/commands/precreate_partitions.py
@@ -1,6 +1,6 @@
 from django.conf import settings
 from django.utils.timezone import now
-from django.core.management.base import BaseCommand
+from django.core.management.base import BaseCommand, CommandParser
 from datetime import timedelta
 from awx.main.utils.common import create_partition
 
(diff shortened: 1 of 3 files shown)


TheRealHaoLiu (approved):

awx/main/management/commands/precreate_partitions.py
```diff
@@ -0,0 +1,23 @@
...
+class Command(BaseCommand):
+    """Command used to precreate database partitions to avoid pg_dump locks"""
+
+    def add_arguments(self, parser: CommandParser) -> None:
+        parser.add_argument('--count', dest='count', action='store', help='The amount of hours of partitions to create', type=int, default=1)
+
+    def _create_partitioned_tables(self, count):
+        start = now()
+        while count > 0:
+            for table in ("main_inventoryupdateevent", "main_jobevent", "main_projectupdateevent", "main_systemjobevent"):
```
AlanCoding: I would like to make 1 request that we don't hard-code these table names. Import this method instead, you may find it necessary to move into utils:

https://github.com/ansible/awx/blob/f1196fc019313e081d40c8650b14dfe5b3f26898/awx/main/management/commands/cleanup_jobs.py#L22-L23

With that method, you don't need to provide a table name for the loop, but you do need to provide the models. Those would be Job, AdHocCommand, ProjectUpdate, InventoryUpdate, SystemJob.

I listed 5 and you listed 4, I think you're missing the ad hoc commands as well.

gamuniz: I refactored based on your feedback @AlanCoding. Let me know if you had something different in mind here.

awx/main/management/commands/precreate_partitions.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def _create_partitioned_tables(self, count):
+        tables = list()
+        for model in (Job, SystemJob, ProjectUpdate, InventoryUpdate, AdHocCommand):
+            tables.append(unified_job_class_to_event_table_name(model))
+        start = now()
+        while count > 0:
+            for table in tables:
+                create_partition(table, start)
+                print(f'Creating partitions for {table} {start}')
```
AlanCoding: I'm just slightly confused about this message coming _after_ the create_partition call. Maybe it should say "Created" instead of "creating"

AlanCoding (approved):

awx/main/management/commands/precreate_partitions.py
gamuniz: Yeah I agree, that message was mostly for me during writing/debug to make sure I had the correct tables/timestamps

PR #13118, opened 2022-10-26
Send real client remote address in TACACS+ authentication packet

pax0s:
##### SUMMARY
Addresses #1797, to add the client IP when authenticating in TACACS+
The client IP is extracted from one of the following header in the request in this order: x-forwarded-for, remote-addr.
If none of those are found it is behaving like previously (no value is passed and the default client flag is used)

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

Issue #1797, opened 2018-04-25
RFE: Send real client remote address in TACACS+ authentication packet


b0urn3:
##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
 - API

##### SUMMARY
Currently  the TACACS+ authentication backend

https://github.com/ansible/awx/blob/96370584062a15271abafab7fc557ac2879aa38c/awx/sso/backends.py#L225-L227

sends default value for client remote address 

https://github.com/ansible/tacacs_plus/blob/526b5a29c5656bbd8644accca238d1e9303dc272/tacacs_plus/flags.py#L81

The function `authenticate()` in  `TACACSClient` supports sending client remote address as parameter and should be used to correctly report client real address.

https://github.com/ansible/tacacs_plus/blob/9ba553f79efcc7b955001503d1953900ba6284a4/tacacs_plus/client.py#L158

##### ENVIRONMENT
* AWX version: 1.0.5
* AWX install method: openshift, minishift, docker on linux, docker for mac, boot2docker
* Ansible version:  N/A
* Operating System: Any
* Web Browser: Any

##### EXPECTED RESULTS

TACACS+ authentication packet will contain clients real remote address.

##### ACTUAL RESULTS

Client remote address is always set to `python_device`.

john-westcott-iv:
@b0urn3 can you confirm this is now addressed with #14077 merged in?

PR #14077, opened 2023-06-01
Send real client remote address in TACACS+ authentication packet

john-westcott-iv:
##### SUMMARY

Replaces #13118

Addresses https://github.com/ansible/awx/issues/1797, to add the client IP when authenticating in TACACS+

The client IP is extracted from one of the following header in the request in this order: x-forwarded-for, remote-addr.
If none of those are found it is behaving like previously (no value is passed and the default client flag is used)

ISSUE

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```


TheRealHaoLiu (approved):
/lgtm

PR #14081, opened 2023-06-02
Add combined roles/collection requirements on project sync

### LINKED ISSUE #12364: Support of unified requirements.yml file with roles/collections do not work

john-westcott-iv:
##### SUMMARY

Previously, during a project sync, we would run ansible-galaxy in the following ways.

For roles we would run `ansible-galaxy roles install` for roles/requirements.y[a]ml.
For collections we would run `ansible-galaxy collections install` for collections/requirements.y[a]ml or requirements.y[a]ml

With this PR we will continue to install roles from roles/requirements.y[a]ml and collections from collections/requirement.y[a]ml. But we will now now install both roles and collections from requirements.y[a]ml files **IF**:
* You are on ansible >= 2.10
* You have both collection and role syncing enabled

If someone was using a requirements.y[a]ml file for only collections this will still work because a collections file is a valid combined file.

However, if you were using a requirements.y[a]ml file for collections and have role syncing disabled, you will have to move your file into the `collections` directory or enable role syncing in order for this file to be used.

Replaces: #13404
Fixes: #12364

##### TESTING
Create a project linked to SCM with a requirements.y[a]ml file at its root, a roles/requirement.y[a]ml file and a collections/requirements.y[a]ml file.
Disable role and collection syncing and sync the project.... no tasks will be run.
Enable role syncing and sync the project.... only the roles/requirements.y[a]ml file will be run.
Disable role syncing and enable collection syncing.... only the collections/requirements.y[a]ml file will be run.
Enable both role and collection syncing... all three files will be run.

If you were to change your version of ansible to < 2.10 the requirements.y[a]ml file would no longer be processed.

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/playbooks/project_update.yml
+++ b/awx/playbooks/project_update.yml
@@ -189,11 +189,12 @@
   connection: local
   name: Install content with ansible-galaxy command if necessary
   vars:
-    galaxy_task_env:  # configure in settings
-    additional_collections_env:
-      # These environment variables are used for installing collections, in addition to galaxy_task_env
-      # setting the collections paths silences warnings
+    galaxy_task_env:  # configured in settings
+    # additional_galaxy_env contains environment variables are used for installing roles and collections and will take precedence over items in galaxy_task_env
+    additional_galaxy_env:
+      # These paths control where ansible-galaxy installs collections and roles on top the filesystem
       ANSIBLE_COLLECTIONS_PATHS: "{{ projects_root }}/.__awx_cache/{{ local_path }}/stage/requirements_collections"
+      ANSIBLE_ROLES_PATH: "{{ projects_root }}/.__awx_cache/{{ local_path }}/stage/requirements_roles"
       # Put the local tmp directory in same volume as collection destination
       # otherwise, files cannot be moved accross volumes and will cause error
       ANSIBLE_LOCAL_TEMP: "{{ projects_root }}/.__awx_cache/{{ local_path }}/stage/tmp"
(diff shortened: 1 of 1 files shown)


awx/playbooks/project_update.yml
```diff
@@ -240,12 +240,34 @@
...
+        chdir: "{{ project_path | quote }}"
+      register: galaxy_combined_result
+      with_fileglob:
+        - "{{ project_path | quote }}/requirements.yaml"
+        - "{{ project_path | quote }}/requirements.yml"
+      changed_when: "'Nothing to do.' not in galaxy_combined_result.stdout"
+      environment: "{{ additional_collections_env | combine(galaxy_task_env) | combine(paths) }}"
+      vars:
+        paths:
+          ANSIBLE_COLLECTIONS_PATHS: "{{ projects_root }}/.__awx_cache/{{ local_path }}/stage/requirements_collections"
```
gamuniz: pretty sure the plural here is going to be deprecated soon. Maybe we want to be forward thinking?

AlanCoding:
I expect this to heavily conflict with https://github.com/ansible/awx/pull/14065, I'd like to compare and contrast the objectives of the two and make some decisions on desired behavior.

awx/playbooks/project_update.yml
AlanCoding: @gamuniz that's pre-existing. Sure, rename it, but do it in the right place.

Here, it is not making any sense why ` | combine(paths)` is added. That is, why was the `paths` variable introduced at all? It appears to be the same as `additional_collections_env`. This is, again, duplicated with the use of `--roles-path` and `--collections-paths` CLI flags in specific commands (but I would not touch this right now). This also shouldn't come at higher precedence than the user's custom `galaxy_task_env` setting.

AlanCoding:
From the other PR, I agree that `req_file: "{{ req_candidates |first_found }}"` is a better pattern than the `with_fileglob` looping mechanism. So my position is now that I'd like to accept both of these. I'm looking over this to see if there's anything else that needs to be done.

AlanCoding (changes requested):
I don't see a good reason for `ANSIBLE_ROLES_PATH` to be an exception here. We should combine this with the `--roles-path` usage in the other task. I'd say _remove_ `--roles-path` and add the env var to `additional_collections_env` instead of this custom `paths` var.

I also want to note that `ANSIBLE_COLLECTIONS_PATHS` is duplicated and unnecessary. This should be removed, which results in the `paths:` var being completely removed. This will give better consistency and a smaller diff.

awx/playbooks/project_update.yml
john-westcott-iv: In, at least my version of ansible, ansible-galaxy install does not respect the --collections-paths:
```
$ ansible-galaxy install --collections-path . awx.awx
usage: ansible-galaxy [-h] [--version] [-v] TYPE ...
ansible-galaxy: error: unrecognized arguments: --collections-path
```

But it does respect the ANSIBLE_COLLECTIONS_PATHS environment variable and combine takes a dict so I just make the `paths` dict variable to do the combine to make the generic install work.

john-westcott-iv: @gamuniz do you know where the deprecation message is? I see them both in the [latest docs](https://docs.ansible.com/ansible/latest/reference_appendices/config.html) but don't see a deprecation warning on either. I could be looking in the wrong spot though.

john-westcott-iv:
@AlanCoding take a look at the latest changes and let me know if this meets your requirements.

awx/playbooks/project_update.yml
```diff
@@ -232,20 +233,37 @@
     - name: Fetch galaxy collections from collections/requirements.(yml/yaml)
       ansible.builtin.command: >
         ansible-galaxy collection install -r {{ item }}
-        --collections-path {{ projects_root }}/.__awx_cache/{{ local_path }}/stage/requirements_collections
```
AlanCoding, line 235: So I'm hoping that this was already unnecessary because `additional_collections_env` set the same thing via `ANSIBLE_COLLECTIONS_PATHS`

awx/playbooks/project_update.yml
```diff
@@ -190,6 +190,8 @@
   name: Install content with ansible-galaxy command if necessary
   vars:
     galaxy_task_env:  # configure in settings
+      # This environment variable is used for install roles
+      ANSIBLE_ROLES_PATH: "{{ projects_root }}/.__awx_cache/{{ local_path }}/stage/requirements_roles"
```
AlanCoding: This extra var is set here

https://github.com/ansible/awx/blob/86f6b54eec6758f618e5cb6d922f7ff508d75462/awx/main/tasks/jobs.py#L1281

and it doesn't merge, it just overwrites anything here. The only reason the variable was put here was to establish an expectation that the CLI ansible-playbook -e will define it.

awx/playbooks/project_update.yml
```diff
@@ -212,10 +214,9 @@
         - name: End play due to disabled content sync
           ansible.builtin.meta: end_play
 
-    - name: Fetch galaxy roles from requirements.(yml/yaml)
+    - name: Fetch galaxy roles from roles/requirements.(yml/yaml)
       ansible.builtin.command: >
         ansible-galaxy role install -r {{ item }}
-        --roles-path {{ projects_root }}/.__awx_cache/{{ local_path }}/stage/requirements_roles
```
AlanCoding, line 218: Here's how I'd do the puzzle here:
 - if you're taking out `--roles-path` here, fine, counter-balance by adding `ANSIBLE_ROLES_PATH` to the shared environment blob, which is `additional_collections_env`, and adding to to `environment:` for this task
 - when you do that, obviously it's not just collections paths you're setting (which is totally fine), so just rename it to something more general, as this defines paths for collections/roles/tmp

awx/playbooks/project_update.yml
```diff
@@ -190,13 +190,15 @@
   name: Install content with ansible-galaxy command if necessary
   vars:
     galaxy_task_env:  # configure in settings
-    additional_collections_env:
+    additional_galaxy_env:
       # These environment variables are used for installing collections, in addition to galaxy_task_env
       # setting the collections paths silences warnings
```
AlanCoding: rewrite this comment

> setting the collections paths silences warnings

That comment is written with the assumption that the _functional path_ is provided by `--collections-path` and you have changed that. This now needs to be aggressively clear that this variable provides the destination collections path to install requirements into.

awx/playbooks/project_update.yml
```diff
@@ -224,28 +225,45 @@
...
+      ansible.builtin.command: >
+        ansible-galaxy install -r {{ item }}
+        {{ ' -' + 'v' * ansible_verbosity if ansible_verbosity else '' }}
+      args:
+        chdir: "{{ project_path | quote }}"
+      register: galaxy_combined_result
+      with_fileglob:
+        - "{{ project_path | quote }}/requirements.yaml"
+        - "{{ project_path | quote }}/requirements.yml"
+      changed_when: "'Nothing to do.' not in galaxy_combined_result.stdout"
```
AlanCoding: Could your "Nothing to do" criteria also be applied to the other 2 Galaxy commands?

Also, `environment:` may be shared between them all, maybe we could set those with a `block:`

john-westcott-iv: Looks like we could do collections:
```
ansible-galaxy collection install awx.awx
Starting galaxy collection install process
Nothing to do. All requested collections are already installed. If you want to reinstall them, consider using `--force`.
```

But not the role install:
```
ansible-galaxy role install geerlingguy.jenkins
Starting galaxy role install process
[WARNING]: - geerlingguy.jenkins (4.3.0) is already installed - use --force to change version to unspecified
```

In addition to extracting environment I also made a var to shorten up the command line.

AlanCoding (approved):
Looks great :+1: 

thanks for adding more polish to this.

jay-steurer (approved):

PR #14082, opened 2023-06-04
Two silly internal cleanups

relrod:
##### SUMMARY

- Nix an unused function from run_dispatcher. This stopped being used in 558e92806b0e692939faf113029ad322960c7369 but was never removed.

- Fix a typo in run_ws_heartbeat: hearbeat -> heartbeat that has existed since the beginning of this daemon.

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/management/commands/run_dispatcher.py
+++ b/awx/main/management/commands/run_dispatcher.py
@@ -3,7 +3,6 @@
 import logging
 import yaml
 
-from django.conf import settings
 from django.core.cache import cache as django_cache
 from django.core.management.base import BaseCommand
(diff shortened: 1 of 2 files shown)


AlanCoding (approved):
I do want to ask @john-westcott-iv if pylint might catch stuff like this for us automatically. I am sure this is not the only such case.

john-westcott-iv:
For the spell checking I use a pre-commit-user hook with aspell. I'm assuming the first thing was just dead code. If so, I don't think pylint would get that but if we could get sonarcloud going that probably would.

PR #14066, opened 2023-05-31
Apply only very conservative database connection reduction changes

AlanCoding:
##### SUMMARY
Replaces https://github.com/ansible/awx/pull/14034 but avoids closing connections for clients that will periodically re-open them (with I guess the exception of the callback receiver main process, but if we have pushback we should delete the log because it is pointless).

##### ISSUE TYPE
 - Bug, Docs Fix or other nominal change

##### COMPONENT NAME
 - API

PR #14083, opened 2023-06-05
Add optional pgbouncer to dev environment

relrod:
##### SUMMARY

This is mostly just to make it easier to start playing with pgbouncer and seeing if we can make it work reasonably with AWX.

As we are starting to explore database performance options and be more proactive about which connections are long-lived (see #14066 and #14067), this provides an easy way to spin up pgbouncer in the dev environment to play around with.

A few **notes** so far:

We cannot use transaction mode, since pg_notify does not work with it. So we are limited to session mode, which negates some (not all) of the point of using pgbouncer I think.

In session mode, if the pool size is set too low, our long-lasting connections hog all the available slots so (for example) the API stops working.

For experimenting, I've set the pool size high (70) and I've been testing with:

```bash
make docker-compose PGBOUNCER=true CONTROL_PLANE_NODE_COUNT=5 EXECUTION_NODE_COUNT=5
```

This seemed to work and put us right around the 50 active client mark (one job running at the time I ran this):

```
pgbouncer=# show pools;
-[ RECORD 1 ]---------+----------
database              | awx
user                  | awx
cl_active             | 51
cl_waiting            | 0
cl_active_cancel_req  | 0
cl_waiting_cancel_req | 0
sv_active             | 51
sv_active_cancel      | 0
sv_being_canceled     | 0
sv_idle               | 1
sv_used               | 3
sv_tested             | 0
sv_login              | 0
maxwait               | 0
maxwait_us            | 0
pool_mode             | session
```

##### ISSUE TYPE

##### COMPONENT NAME

 - API


AlanCoding (approved):
Good to have, solves a pretty narrow case for the moment but I wouldn't mind keeping it around for later expansion.

PR #14093, opened 2023-06-06
Lazy init VERSION vars in Makefile

jjwatt:
##### SUMMARY
Setup VERSION and COLLECTION_VERSION vars so that they're only initialized when needed. This avoids running everything you have from shells in the top level even when you just want to `make help`. It's especially frustrating since our version calculation has an external dependency! Anyway, here's a breakdown: Start them as recursive vars and skip evaluating the rhs at the top-level (GNU Make will only eval the lhs and save the rhs for later), then on use the rhs will start expanding, and after expansion it will rewrite VERSION as a simple var holding the version that we calculate with our expensive dependent shell operations (so it won't have to run them again, either).

##### ISSUE TYPE

##### COMPONENT NAME

 - API
 - UI
 - Collection
 - CLI
 - Docs
 - Other

##### AWX VERSION

```
awx: 22.3.1.dev31+g5a7cc88ace
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ NPM_BIN ?= npm
 CHROMIUM_BIN=/tmp/chrome-linux/chrome
 GIT_BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD)
 MANAGEMENT_COMMAND ?= awx-manage
-VERSION := $(shell $(PYTHON) tools/scripts/scm_version.py)
+VERSION ?= $(shell $(PYTHON) tools/scripts/scm_version.py)
 
 # ansible-test requires semver compatable version, so we allow overrides to hack it
 COLLECTION_VERSION ?= $(shell $(PYTHON) tools/scripts/scm_version.py | cut -d . -f 1-3)


shanemcd:
Can you add some more information to either the PR description or the commit message? I'm not sure I understand what's going on here or why this is a good idea just by reading the code.

jjwatt:
Sorry! The template thing messed me up. I wrote a little summary. I also switched this to draft because I just thought of something:

I think we have an implicit dependency (maybe more than one) on version manifestation ordering. i.e., if you look at CI jobs, they get and log the wrong version all the time when they start because they either can't access git or they didn't pull tags, etc. etc. -- iow, some kind of failure in setuptools_scm that we don't catch or handle. We just go with something like `dev0.1` -- some kind of default. Anyway, some time after that, I guess even in CI the version might be right. So, what I'm worried about is that we're actually dependent on the implicit way that everything sort of shakily works right now: The same Makefile is used outside and inside the container, and it may count on "fail-passing" (passing that ugly string) in CI in the GitHub VM, but then maybe it works OK when it runs in CI: *inhale* from the Makefile from the docker container invoked by the Makefile running in GitHub CI.

So, I want to make sure we're not dependent on any of that coincidental behavior before I commit.

jjwatt:
Example. e.g., at 8c86092bf5cb92a9641bdf386da7dfb43e75c74a

Without fetching tags on Fedora 38:
```
[jwattenb@fedora awx]$ make VERSION
awx: 0.1.dev33334+g8c86092
[jwattenb@fedora awx]$ make print-VERSION
0.1.dev33334+g8c86092
```
This must be because Fedora seems to actually have setuptools_scm available to its site packages. Most of my other systems don't. I wonder if this explains why most AWX developers don't seem as annoyed at this as I do. It might be that Fedora includes the `setuptools_scm` package, but most distros don't seem to. I bet that the giant GH Actions VM probably includes it. Anyway, back to the point:

with setutpools_scm (without pull tags)        |  with git pull --tags
======================================|===================
0.1.dev33334+g8c86092                                       |  22.3.1.dev30+g8c86092bf5

So, already our VERSION is not deterministic. Nothing stops anyone from using that, and it's even in the CI logs, *and yet* we still ask that developers have that dependency. So, that's part of the motivation, too. But, here I just want to call attention to how it's different in these two places, and so now I'm unsure about *my* commit that would improve it. Like, we might now be *counting on* and *depening on* the Make invocation *re-running* the whole scm_version.py command all over again. If we do depend on that, then something might not go quite right since I essentially "cache" it and only run the "scm_version.py" thingy once.

I think it might be OK afterall, though. Doing some more testing.

jjwatt:
Maybe not. This might be a real failure on GitHub Actions awx-operator. Like, it dynamically generates a recipe for that 0.1dev version but then can't find it later :/. That's from inside the awx container, though. hmm. I think it's the only one that runs `make sdist` of all the GH Actions, so I'm going to start looking there.

Unconfirmed so far, but I have a weird feeling that with how things are currently done, we could be in situations where the "parent" Make thinks VERSION is one thing, but its submake invocations think it's something else, because:

1. We don't export the VERSION var and
2. We're strict about it sometimes and loose about it others

jjwatt:
So I don't lose it, here's the failure I'm seeing in the awx-operator CI job around the "Build AWX Image" task or whatever:
```
DONE 0.0s", "", "#18 [builder 11/11] RUN make sdist && /var/lib/awx/venv/awx/bin/pip install dist/awx.tar.gz", "#18 sha256:d7f87f9f937e45a8752ba928391d6593b3bcfabd9492f4a79d760785ec39e798", "#18 0.275 Unable to import setuptools-scm, attempting to install now...", "#18 2.897 make: *** No rule to make target 'dist/awx-0.1.dev1.tar.gz', needed by 'sdist'.  Stop.", "#18 ERROR: executor failed running [/bin/sh -c make sdist && /var/lib/awx/venv/awx/bin/pip install dist/awx.tar.gz]: exit code: 2", "------", " > [builder 11/11] RUN make sdist && /var/lib/awx/venv/awx/bin/pip install dist/awx.tar.gz:", "------", "executor failed running [/bin/sh -c make sdist && /var/lib/awx/venv/awx/bin/pip install dist/awx.tar.gz]: exit code: 2"], "stdout": "", "stdout_lines": []}
```
There are so many things that could go wrong here. Wow. But first, I think we start at the Makefile's `sdist` target and how it generates the `dist/$(SDIST_TAR_FILE)` dependency which is built from `$(VERSION)` somewhere in the soup (L63ish):
```
SDIST_TAR_NAME=$(NAME)-$(VERSION)

SDIST_COMMAND ?= sdist
SDIST_TAR_FILE ?= $(SDIST_TAR_NAME).tar.gz
```
So, yeah I think it could happen that the parent make is now saving my VERSION with the bad ("fail-pass" aka `0.1.dev`) that it gets from `setuptools_scm` whenever it *does* get the chance to run in CI without the tags fetched. Then we're crashing around the awx build step because there's no `0.1.dev.tar.gz` or whatever. So, pre-patch that VERSION var was getting the chance to change. So, we are probably unintentionally dependent on non-deterministic results there. Pushed the original Makefile back over in this PR to compare results. It needs it, but I really want to avoid fixing the VERSION machinery in this PR. I want to limit this PR to *just* preventing setuptools_scm and pip from taking over make.

jjwatt:
Yeah. That's so wild and cool that it seems to break that rule.
Snipped from logs:
Failing:
```
DONE 0.0s", "", "#18 [builder 11/11] RUN make sdist && /var/lib/awx/venv/awx/bin/pip install dist/awx.tar.gz", "#18 sha256:d7f87f9f937e45a8752ba928391d6593b3bcfabd9492f4a79d760785ec39e798", "#18 0.275 Unable to import setuptools-scm, attempting to install now...", "#18 2.897 make: *** No rule to make target 'dist/awx-0.1.dev1.tar.gz', needed by 'sdist'.  Stop.", "#18 ERROR: executor failed running [/bin/sh -c make sdist && /var/lib/awx/venv/awx/bin/pip install dist/awx.tar.gz]: exit code: 2", "------", " > [builder 11/11] RUN make sdist && /var/lib/awx/venv/awx/bin/pip install dist/awx.tar.gz:", "------", "executor failed running [/bin/sh -c make sdist && /var/lib/awx/venv/awx/bin/pip install dist/awx.tar.gz]: exit code: 2"], "stdout": "", "stdout_lines": []}
```
Passing
```
  .828 Creating tar archive\n#18 10.15 removing 'awx-0.1.dev1' (and everything under it)\n#18 10.28 Successfully built awx-0.1.dev1.tar.gz\n#18 10.32 ln -sf awx-0.1.dev1.tar.gz dist/awx.tar.gz\n#18 10.32 echo yes\n#18 10.32 yes\n#18 10.32 #############################################\n#18
```
I really gotta crack reproducing the CI environment better so I can iterate on these faster.

jjwatt:
One last little note for today. Because of this, the VERSION will never ever ever ever be right in the dev image/build anyway. So, I'm losing hair over something we don't even really use. I'll sort out how I want to approach it. Let me be clear: that's a huge problem. Like, CI builds and tests fail because this VERSION hackery doesn't work and that could limit commits and it definitely hurts velocity. Trying to sort it all out is hard work, too. If it's not important, it should not break the CI! I still think there's a simple way to get what I need now without fixing everything. It's interesting when you worry that fixing something could cause problems in other parts of the system.
https://github.com/ansible/awx/blame/8c86092bf5cb92a9641bdf386da7dfb43e75c74a/tools/ansible/roles/dockerfile/templates/Dockerfile.j2#L83
```
# Copy source into builder, build sdist, install it into awx venv
COPY . /tmp/src/
WORKDIR /tmp/src/
RUN make sdist && /var/lib/awx/venv/awx/bin/pip install dist/awx.tar.gz
```
We copy the whole damn awx repo/root dir into /tmp/src and then try to build `sdist` there :(. I'm sure there ain't no `.git` dir there and so our beloved `setuptools_scm` will just always return that ol' `0.1.dev...` string. The `Makefile` has also been hacked up to symlink whatever named version tar.gz is built to the more generic `awx.tar.gz`, too, so it could just use it like the name of a CPU reg.

Silver lining, though: I was worried about breaking it, but it's already broken.

I'm not sure about this, either:
```
RUN make sdist && /var/lib/awx/venv/awx/bin/pip install dist/awx.tar.gz
```
Every line in a Dockerfile is like shelling in and running it, and then saving and committing a layer. Above this there are lots of real simple things split up into separate layers, but this one is combined without any error handling or anything. That makes me wonder if it's intentional or if it's because Dockerfile's are confusing. This is the part that's failing for me: `make sdist` before the optimistic `&&.` Yes, it's usually good to combine stuff you want to be atomic, but because it's a call to make there's no way it's atomic. If it were separate here, any partial steps for the `make sdist` could stick around in the last layer for debugging or for continuing the build (but, we don't really take advantage of make's ability to not redo work anyway). If we did, I would definitely want to separate build and install into different layers, right? I think so.

jjwatt:
tl;dr: This should be all I need right now, and it is passing everywhere. Feel free to ignore my analysis/self-indulgent ramblings and the previous implementation. Other stuff is for later, and once I get the GitHub CI jobs running the same way locally. This one character patch should do the trick.

AlanCoding (approved):

Issue #13417, opened 2023-01-09
Incorrect deprecation warning for awx.awx.schedule_rrule


jmighion:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

When using the schedule_rrule lookup plugin in the awx collection, I get the following deprecation warning:

```
[DEPRECATION WARNING]: The lookup plugin 'awx.awx.schedule_rrule' was expected to return a list, got '<class 'str'>'
instead. The lookup plugin 'awx.awx.schedule_rrule' needs to be changed to return a list. This will be an error in Ansible
2.18. This feature will be removed in version 2.18. Deprecation warnings can be disabled by setting
deprecation_warnings=False in ansible.cfg.
```

This seems wrong because that plugin does intend to return a string and not a list. 

Ansible version:
```
$ ansible --version
ansible [core 2.14.1]
  config file = /Users/jmighion/.ansible.cfg
  configured module search path = ['/Users/jmighion/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/jmighion/.pyenv/versions/3.10.0/envs/3.10.0_dev/lib/python3.10/site-packages/ansible
  ansible collection location = /Users/jmighion/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/jmighion/.pyenv/versions/3.10.0_dev/bin/ansible
  python version = 3.10.0 (default, Dec  6 2021, 14:41:16) [Clang 12.0.0 (clang-1200.0.32.2)] (/Users/jmighion/.pyenv/versions/3.10.0/envs/3.10.0_dev/bin/python3.10)
  jinja version = 3.1.2
  libyaml = True
```

awx.awx version 21.10.0

Sample playbook:
```
- name: Test
  hosts: localhost
  gather_facts: false
  tasks:
    - name: Debug
      ansible.builtin.debug:
        msg: "{{ query('awx.awx.schedule_rrule', 'hour', start_date='2022-01-01', timezone='GMT') }}"
```

### AWX version

21.10.0

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [X] Collection
- [ ] CLI
- [ ] Other

### Installation method

N/A

[...]

PR #14095, opened 2023-06-07
Make rrule plugins return a list with 1 element

AlanCoding:
##### SUMMARY
Connect https://github.com/ansible/awx/issues/13417

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx_collection/plugins/lookup/schedule_rrule.py
+++ b/awx_collection/plugins/lookup/schedule_rrule.py
@@ -79,10 +79,13 @@ EXAMPLES = """
 """
 
 RETURN = """
-_raw:
+_list:
   description:
-    - String in the rrule format
-  type: string
+    - A list with one element which is a string in the rrule format
+  type: list
+  elements: str
+  example:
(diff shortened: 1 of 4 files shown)


AlanCoding:
Seems suspicious that all the checks passed...

john-westcott-iv:
Is there a reason we are looking to do this? Seems like it could "break" existing integrations.
@sean-m-sullivan are you using these lookup plugins anywhere?

sean-m-sullivan:
I am not using this in a lot of places, but have the same concern as John. However on looking at the underlying issue, we need to conform to sanity standards, and return the right thing

AlanCoding:
> Is there a reason we are looking to do this?

The current behavior will become an error in a future version of Ansible core.

I guess you could argue that we don't fix it until Ansible core releases a version that breaks it. In that case the question is whether you care more about compatibility with older core versions, or with newer core versions.

dmzoneill:
Marking as blocked/draft until it Ansible core breaks behaviour, per @AlanCoding feedback

AlanCoding:
Replaced by https://github.com/ansible/awx/pull/15625

Issue #14057, opened 2023-05-27
k3s max container log size leading to job error


mamercad:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.
- [X] I am **NOT** reporting a (potential) security vulnerability. (These should be emailed to `security@ansible.com` instead.)

### Bug Summary

Seeing "Job terminated due to error" in the UI on a very simple playbook (below).

In the UI it looks like this:

<img width="676" alt="image" src="https://github.com/ansible/awx/assets/6192266/e402d817-db0f-45b2-aab2-d18a9057777f">

In the logs around the end of this job I see this:

```text
awx-task 2023-05-27 14:52:07,144 INFO     [2d7014133ee34ad9b12564981b1764b2] awx.main.commands.run_callback_receiver Starting EOF event processing for Job 93
awx-task 2023-05-27 14:52:07,207 INFO     [2d7014133ee34ad9b12564981b1764b2] awx.analytics.job_lifecycle job-93 post run
awx-task 2023-05-27 14:52:07,357 INFO     [2d7014133ee34ad9b12564981b1764b2] awx.analytics.job_lifecycle job-93 finalize run
awx-task 2023-05-27 14:52:07,440 WARNING  [2d7014133ee34ad9b12564981b1764b2] awx.main.dispatch job 93 (error) encountered an error (rc=None), please see task stdout for details.
awx-task Task was destroyed but it is pending!
awx-rsyslog 2023-05-20 18:23:01,640 INFO success: awx-rsyslog-configurer entered RUNNING state, process has stayed up for > than 30 seconds (startsecs)
awx-rsyslog 2023-05-20 18:23:01,640 INFO success: awx-rsyslog-configurer entered RUNNING state, process has stayed up for > than 30 seconds (startsecs)
awx-rsyslog 2023-05-20 18:23:03,643 INFO success: awx-rsyslogd entered RUNNING state, process has stayed up for > than 30 seconds (startsecs)
awx-rsyslog 2023-05-20 18:23:03,643 INFO success: awx-rsyslogd entered RUNNING state, process has stayed up for > than 30 seconds (startsecs)
[...]

Issue #14003, opened 2023-05-15
Missing job output and log lines


mamercad:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.
- [X] I am **NOT** reporting a (potential) security vulnerability. (These should be emailed to `security@ansible.com` instead.)

### Bug Summary

I've run into this rather strange situation where I'm missing output from jobs. In an attempt to reproduce it, I created a relatively simple Ansible playbook which runs for a few hours and generates a few thousands of lines of output. For the most recent test, it ran for about 4 hours and 20 minutes and the playbook simply counted to 40,000.

The output in the UI and when downloaded looks like this (abbreviated):

![image](https://github.com/ansible/awx/assets/6192266/a3e0c85d-f2b4-4a16-a70c-d85e56c23fdb)

Notice the huge swath of missing lines.

The playbook that it's running is quite simple:

```ansible
---
- name: long_log_lines
  hosts: localhost
  connection: local
  gather_facts: false
  vars:
    how_many: 40000

  tasks:
    - name: Create a list
      ansible.builtin.command:
        cmd: seq {{ how_many }}
      register: the_list

    - name: Set a fact for the list
      ansible.builtin.set_fact:
        the_list: "{{ the_list.stdout_lines }}"

    - name: Ping list times
      ansible.builtin.ping:
      loop: "{{ the_list }}"
```

### AWX version

21.11.14

### Select the relevant components

- [X] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

kubernetes

### Modifications

no

### Ansible version

2.9

### Operating system

Linux

### Web browser

Chrome

### Steps to reproduce

This should be simple to reproduce, it's a very simple playbook.

### Expected results

That all of the output is there.

### Actual results

Missing many thousands of lines.

dylex:
We're seeing the same thing on 22.2.0 on a new k8s multi-node cluster. (We did not see this issue on an older 21.7.0, single-node k8s. Not sure what the critical difference is.) I do see the output in the automation-job pod log, but various errors in the awx-ee log (at the time the job ends):
```
[...]

PR #14097, opened 2023-06-07
[wsrelay] Give connection tasks time to clean up

relrod:
##### SUMMARY

When we close/cancel a connection to a web node, give the task time to clean up after itself and cleanly exit. Otherwise, the Python GC might clean up the task too early and this leads to ugly log messages like this: "Task was destroyed but it is pending!"

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/wsrelay.py
+++ b/awx/main/wsrelay.py
@@ -245,13 +245,13 @@ class WebSocketRelayManager(object):
                 if action == "online":
                     self.known_hosts[hostname] = ip
                 elif action == "offline":
-                    self.cleanup_offline_host(hostname)
+                    await self.cleanup_offline_host(hostname)
             except Exception as e:
                 # This catch-all is the same as the one above. asyncio will eat the exception
                 # but we want to know about it.
                 logger.exception(f"on_ws_heartbeat exception: {e}")
 
-    def cleanup_offline_host(self, hostname):
+    async def cleanup_offline_host(self, hostname):
         """
         Given a hostname, try to cancel its task/connection and remove it from
         the list of hosts we know about.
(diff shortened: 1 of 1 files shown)


fosterseth:
were you able to reproduce the scenario that logged the  "Task was destroyed.." message?

relrod:
I wasn't able to reproduce it, but I am fairly confident that this would cause that in some conditions. Even if there's something else causing that, this is still something that I think needs to be fixed.

relrod:
https://bugs.python.org/issue44665 - This is _kind of_ a reference, if you need a citation :smile:

And it specifically mentions the error we're seeing. The issue itself is about adding asyncio documentation to prevent doing exactly what the current code is doing (killing off the strong reference to the task before it is completed).

This seems to be a well-known phenomenon in python asyncio, you have to keep a strong reference around to tasks you create with `create_task()` or bad things can happen.

awx/main/wsrelay.py
```diff
@@ -323,7 +336,7 @@ async def run(self):
                 logger.info(f"Adding {new_remote_hosts} to websocket broadcast list")
 
             for h in deleted_remote_hosts:
-                self.cleanup_offline_host(h)
+                await self.cleanup_offline_host(h)
```
AlanCoding: Considering that timeouts are a possibility, you might as well run this simultaneously for all hosts in `deleted_remote_hosts`

AlanCoding (approved):

AlanCoding:
Needs a rebase.

I also _kind_ of liked the callback method in the python bug you linked `task.add_done_callback(lambda t: running_tasks.remove(t))`, since it would allow instance ejection from the tracking dictionary. But this interest is almost entirely academic, as there's no good functional reason for such a change.

You should consider my comment about running these concurrently, but either way there is probably little risk because removing a node is an infrequent task and this is unlikely to choke because the main loop gets stalled waiting for the cancellations.

AlanCoding:
I see the error mentioned in issues https://github.com/ansible/awx/issues/14057 and https://github.com/ansible/awx/issues/14003 and both mention a fairly ancient resolution of the k8s log size issue. So I think we're good.

AlanCoding (approved):
Looks great, very async

PR #14098, opened 2023-06-08
Add settings.RECEPTOR_LOG_LEVEL, update work signing key path

fosterseth:
##### SUMMARY

Offers ability to set log level of receptor (k8s only)

Intended to be paired with https://github.com/ansible/awx-operator/pull/1444

Should be backwards compatible with older operators too.

Instead of hard-coding part of the receptor config file, we can just read the current file (and pop the -peers and -listeners off)

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### AWX VERSION

```
awx: 22.3.1.dev34+g07034cfa95
```

### DIFF
--- a/awx/main/tasks/receptor.py
+++ b/awx/main/tasks/receptor.py
@@ -639,11 +639,11 @@ class AWXReceptorJob:
 #
 RECEPTOR_CONFIG_STARTER = (
     {'local-only': None},
-    {'log-level': 'info'},
+    {'log-level': settings.RECEPTOR_LOG_LEVEL},
     {'node': {'firewallrules': [{'action': 'reject', 'tonode': settings.CLUSTER_HOST_ID, 'toservice': 'control'}]}},
     {'control-service': {'service': 'control', 'filename': '/var/run/receptor/receptor.sock', 'permissions': '0660'}},
     {'work-command': {'worktype': 'local', 'command': 'ansible-runner', 'params': 'worker', 'allowruntimeparams': True}},
-    {'work-signing': {'privatekey': '/etc/receptor/signing/work-private-key.pem', 'tokenexpiration': '1m'}},
+    {'work-signing': {'privatekey': '/etc/receptor/work_private_key.pem', 'tokenexpiration': '1m'}},
     {
         'work-kubernetes': {
             'worktype': 'kubernetes-runtime-auth',
--- a/awx/settings/defaults.py
+++ b/awx/settings/defaults.py
@@ -959,6 +959,9 @@ AWX_RUNNER_KEEPALIVE_SECONDS = 0
 # Delete completed work units in receptor
 RECEPTOR_RELEASE_WORK = True
 
+# K8S only. Use receptor_log_level on AWX spec to set this properly
+RECEPTOR_LOG_LEVEL = 'info'
+
 MIDDLEWARE = [
     'django_guid.middleware.guid_middleware',
     'awx.main.middleware.SettingsCacheMiddleware',


awx/settings/defaults.py
```diff
@@ -959,6 +959,9 @@
 # Delete completed work units in receptor
 RECEPTOR_RELEASE_WORK = True
 
+# K8S only. Use receptor_log_level on AWX spec to set this properly
+RECEPTOR_LOG_LEVEL = 'info'
```
TheRealHaoLiu, line 963: does this also affect the generated instance install bundle?

TheRealHaoLiu: nvm

awx/settings/defaults.py
```diff
@@ -959,6 +959,9 @@
 # Delete completed work units in receptor
 RECEPTOR_RELEASE_WORK = True
 
+# K8S only. Use receptor_log_level on AWX spec to set this properly
```
TheRealHaoLiu, line 962:
```suggestion
# This setting should be passed in from awx-operator via AWX spec. Mismatch between receptor_log_level in AWX spec and this setting will cause controlplane receptor to not be able to start.
```

or something like that

awx/main/tasks/receptor.py
```diff
@@ -632,53 +632,21 @@ def kube_config(self):
...
-            'key': '/etc/receptor/tls/receptor.key',
-            'mintls13': False,
-        }
-    },
-)
-
-
 @task()
 def write_receptor_config():
+    current_config = read_receptor_config()  # this grabs lock
```
AlanCoding: Isn't this dangerous? By using the current config as a starter, aren't you kind of cargo-culting the config in an automated way? That could create some weird problems of config drift down the line.

Also, I'd like to move in the direction of generating `receptor_config` in a separate method so that we can test it. Before this, it looked like all we needed to test it was the `Instance` and `InstanceLink` objects in the database. We can still test this after your restructuring here, but that just means moving the "starter config" to the tests.

AlanCoding: Looking down at the line `receptor_config.append(peer)`, I can see this not working _at all_ since you're not checking in advance whether the link already exists in the config, so it could end up adding it every time it rewrites the config.

TheRealHaoLiu: **cargo culting** (uncountable) (chiefly computing) An approach that copies an existing successful approach without properly analysing and understanding it.

TheRealHaoLiu:
@fosterseth i think the safer approach is to just go ahead and use the old method of templating out the receptor config file and add the ability to config log level instead of trying to rewrite the way that we generate the receptor config file 

I understand that there's currently a PITA that when u change the file names you need to change in both awx and awx-operator we can address this separately if desired.

awx/main/tasks/receptor.py
fosterseth: The logic here strips out any "peer" entry, and we simply readd them later. So in a way we are starting "fresh" without having the hard coded template.

I think I will just revert this part back for now and we can figure out a better solution later.

awx/main/tasks/receptor.py
```diff
@@ -639,11 +639,11 @@ def kube_config(self):
 #
 RECEPTOR_CONFIG_STARTER = (
     {'local-only': None},
-    {'log-level': 'info'},
+    {'log-level': settings.RECEPTOR_LOG_LEVEL},
     {'node': {'firewallrules': [{'action': 'reject', 'tonode': settings.CLUSTER_HOST_ID, 'toservice': 'control'}]}},
     {'control-service': {'service': 'control', 'filename': '/var/run/receptor/receptor.sock', 'permissions': '0660'}},
     {'work-command': {'worktype': 'local', 'command': 'ansible-runner', 'params': 'worker', 'allowruntimeparams': True}},
-    {'work-signing': {'privatekey': '/etc/receptor/signing/work-private-key.pem', 'tokenexpiration': '1m'}},
+    {'work-signing': {'privatekey': '/etc/receptor/work_private_key.pem', 'tokenexpiration': '1m'}},
```
AlanCoding, line 646: Is this change to be _paired_ with https://github.com/ansible/awx-operator/pull/1442?

The new location looks like it's the correct one, I agree with the change, but I'd like a little bit better linking and description of the changes here, since it's not all all that this should happen from the title of this PR, and the associated changes & issue are not immediately discoverable.

AlanCoding (approved):

AlanCoding:
I'm going to merge this myself because of the stated compatibility concerns, and because Seth isn't around right now.

awx/main/tasks/receptor.py
AlanCoding: Never mind my prior comments, this was rendered defunct by https://github.com/ansible/awx/pull/14151, but did not show due to a git issue. This line is not actually changed with this merge.

PR #14101, opened 2023-06-09
Perform project updates and inventory updates on correct branch

willthames:
##### SUMMARY

Ensure that job template's branch is taken into account when performing inventory and source updates

Without this change, scm branch isn't passed to the update jobs and it ends up using HEAD, which is only ever accidentally what is wanted.

This has no effect on the actual job being run, which does run against the correct branch (but potentially with an incorrect inventory).

##### ISSUE TYPE

##### COMPONENT NAME
 - API

##### AWX VERSION

```
awx: 22.1.1.dev136+gb310f4fd56.d20230609
```

##### ADDITIONAL INFORMATION
Reproduction

* Create a project with an inventory source and job template based on an SCM repository. Project and job template should have a default branch.
* Update HEAD of the SCM repository to something different to the default branch
* Run the project - the project will run against the correct branch but the Source Control Update and Inventory Sync will run against HEAD.

### DIFF
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -328,7 +328,7 @@ class DependencyManager(TaskBase):
         '''
         if (update is None) or (update.status in ['failed', 'canceled', 'error']):
             return True
-        if update.status in ['waiting', 'pending', 'running']:
+        if update.status in ['new', 'waiting', 'pending', 'running']:
             return False
 
         return bool(((update.finished + timedelta(seconds=cache_timeout))) < tz_now())
@@ -397,7 +397,7 @@ class DependencyManager(TaskBase):
 
     @timeit
     def _schedule(self):
-        self.get_tasks(dict(status__in=["pending"], dependencies_processed=False))
+        self.get_tasks(dict(status__in=["new", "pending"], dependencies_processed=False))
 
         if len(self.all_tasks) > 0:
             deps = self.generate_dependencies(self.all_tasks)
@@ -672,7 +672,7 @@ class TaskManager(TaskBase):
         self.process_running_tasks(running_tasks)
         self.subsystem_metrics.inc(f"{self.prefix}_running_processed", len(running_tasks))
 
-        pending_tasks = [t for t in self.all_tasks if t.status == 'pending']
+        pending_tasks = [t for t in self.all_tasks if t.status in ['new', 'pending']]
 
         self.process_pending_tasks(pending_tasks)
         self.subsystem_metrics.inc(f"{self.prefix}_pending_processed", len(pending_tasks))
@@ -704,7 +704,7 @@ class TaskManager(TaskBase):
 
     @timeit
     def _schedule(self):
-        self.get_tasks(dict(status__in=["pending", "waiting", "running"], dependencies_processed=True))
+        self.get_tasks(dict(status__in=["new", "pending", "waiting", "running"], dependencies_processed=True))
 
         self.after_lock_init()
         self.reap_jobs_from_orphaned_instances()
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -333,12 +333,15 @@ class DependencyManager(TaskBase):
 
         return bool(((update.finished + timedelta(seconds=cache_timeout))) < tz_now())
 
-    def get_or_create_project_update(self, project_id):
+    def get_or_create_project_update(self, project_id, scm_branch=None):
         project = self.all_projects.get(project_id, None)
         if project is not None:
             latest_project_update = project.project_updates.filter(job_type='check').order_by("-created").first()
             if self.should_update_again(latest_project_update, project.scm_update_cache_timeout):
-                project_task = project.create_project_update(_eager_fields=dict(launch_type='dependency'))
+                fields = dict(launch_type='dependency')
+                if scm_branch and project.allow_override:
+                    fields['scm_branch'] = scm_branch
+                project_task = project.create_project_update(_eager_fields=fields)
                 project_task.signal_start()
                 return [project_task]
             else:
@@ -346,7 +349,7 @@ class DependencyManager(TaskBase):
         return []
 
     def gen_dep_for_job(self, task):
-        dependencies = self.get_or_create_project_update(task.project_id)
+        dependencies = self.get_or_create_project_update(task.project_id, task.scm_branch)
 
         try:
             start_args = json.loads(decrypt_field(task, field_name="start_args"))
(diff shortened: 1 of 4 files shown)


AlanCoding:
I thought the issue here was resolved by the merge of https://github.com/ansible/awx/pull/12073 by @fosterseth 

I want to understand a little bit about what that didn't do which is now being requested here. I'm sure there are open issues on this subject.

awx/main/scheduler/task_manager.py
```diff
@@ -279,6 +279,8 @@ def create_project_update(self, task, project_id=None):
         # Project created 1 seconds behind
         project_task.created = task.created - timedelta(seconds=1)
         project_task.status = 'pending'
+        if task.scm_branch:
+            project_task.scm_branch = task.scm_branch
```
AlanCoding: Whether or not the part about setting the inventory update branch is a good idea (and it very well could be), I think applying the branch to the project update is not a good idea. If a job's `scm_branch` matches that of the project, then this is not necessary. If the job's `scm_branch` does not match the project, then it doesn't depend on the dependent project update in the first place. This would not do anything meaningful. We could, I think, avoid creating a project update dependency entirely in the case that `update_on_launch` is true, `branch_override` is true, and the job specifies a unique `scm_branch`. So in the case you're concerned about, ideally, we should never reach this code.

willthames: I might have been overly careful here, or added this change before I tried the inventory update change. I can definitely check if removing this line of code still fixes things.

willthames:
@AlanCoding I have now properly documented my findings with a minimal reproducible test suite.

See https://github.com/willthames/awx-project-update-test-suite for the details.

Note that with the change in this PR, Test 1 still fails (my workaround for that is to set project default branch), Tests 3 and 4 both pass.

Removing lines 282 and 283 as you suggest causes Test 4 to fail - it runs using the override branch playbook but against the job-default inventory (I only retried the second part of the test suite, Test 3 does succeed)

AlanCoding:
https://github.com/ansible/awx/pull/13644 was the recent patch I was trying to think of. I didn't understand your test case fully.

Aohzan:
> #13644 was the recent patch I was trying to think of. I didn't understand your test case fully.

hello, we have the same need: we want to dynamically choose the inventory source branch from the job template task and not "statically" from the inventory source. It will allowed multiple user to run playbooks with an inventory from their branch

pierre-cegid:
That is exactly what we need, thank you very much @willthames !
We are looking to propagate the scm branch set in awx template to awx project and awx inventory sources.

willthames:
@AlanCoding can I rerequest another look at this one?

I've been using this version of task_manager.py in our AWX for months now (occasionally rebasing it during AWX upgrades) and I'm very happy with how it works - would be good to get it merged in!

willthames:
Ah, I was looking at my wrong source branch, it seems that task manager has been refactored since I raised this PR and that the remaining changes I have may no longer be required - I'll validate that and either update this PR or close it

willthames:
I've added some tests to AWX to validate the use of scm_branch with job templates and tasks and updated the code to adjust for the very significant refactor of the underlying workflows. 

I'm happy that the tests pass after passing scm_branch to the inventory update and project update tasks

There are some changes to allow for `new` status that there might be a better way to do (I couldn't get the inventory update to ever fire, even for the known good test case, without that change)

dmzoneill:
@willthames Thanks for your continued work on this.
I'm moving it to draft for the time being to make it clearer to developers what is ready for merging.
Please once the undraft once it's ready.

willthames:
Pretty sure the failing test looks like the RPM repo was temporarily down, I don't have permissions to retry a single test though

fosterseth:
@willthames thanks for the PR, reviewing but having trouble understanding the root issue

1. I setup a project with branch set to `foo`
image
2. I setup a job template based on that project, also with branch set to `foo`
image
3. I setup an inventory source based on that project, with nothing set in the branch field
image
4. importantly update on launch enabled for the project and inventory source
5. Go to github and change my inventory.ini file with a different host (127.0.0.32)

when I launch job I get the following updates spawned,

image

the job runs against the right host
image

willthames:
Hi @fosterseth , the accompanying tests in this PR should show the behaviour, as I wrote them to fail without this change.

The problem is when using a branch that isn't the default branch for the project/job template

* Create a project with default branch `main`
* Create a job template with default branch `main`
* Create a branch with an inventory update (you could use `foo` if it only had the change above) and a task change
* Run a job using `foo` branch
* It will probably use `main` for inventory and `foo` for the tasks - you can validate this even without looking at the output by going to the underlying jobs and seeing what revision the jobs ran under, but making the outputs change make it a little more obvious

Let me know if you need any further info.

mcapra:
Any updates on this? @willthames has described the problem I and others in #5692 are facing.

Underpinning all this is a desire to have a gitops style workflow for managing our inventory. As it stands, though I can pass a ref to specify playbook/role/etc changes, I can't pass a ref to specify inventory changes. I would like to have a ref representing a feature branch for the inventory, pass that ref to a Job Template running in Check Mode, and have the Inventory Sources associated with that Job Template respect the ref that was passed in the Job Template.

crypto-services:
Hey guys, is there any update on this? This issue makes it very difficult to test host configurations before merging to main. The branch specified on launch of a template should apply to roles and inventory if they both have the same source.

stefankaufmann:
This issue still persists and we quite often run into this issue. I hope the use case is understood, but I provide an example here what happens quite often to us: 

1. We adjust a playbook on `branch_a`
2. The playbook requires a new host_var `var_a` which is added to the inventory for that host
3. We launch the job template for `branch_a`
4. The job fails, because `var_a` is not defined for this host. 

The problem still is, that the inventory sync happend for the default branch configured for the inventory and not for the branch selected in the job template using the branch override functionality. 

I also tried to figure out how inventories are passed to the jobs, because I think that there might be race condition even with this fix merged. Let's assume the following

1. A template `A` job is triggered for `branch_a`
2. The fix applied causes the inventory to sync also for `branch_a`
3. The actual job `A` is pending for some reason (no executor, waiting for hosts, etc.)
4. Another job using the same inventory is triggered on `branch_b`
5. The inventory is synced for `branch_b`
6. The job `A` is now starting and building the inventory. **BUT** the last state of the inventory was synced for `branch_b`

From my understanding, the state of an inventory is global and we might run into race conditions. What might be needed is an isolation for inventories, such that a triggered inventory sync for a specific branch will be kept in a stable state until the actual job starts. 

My assumptions might be wrong and this scenario will not occur, so I will be glad for any feedback. And many thanks @willthames for your effort already put into fixing this issue.

Issue #14113, opened 2023-06-12
Facing issue while adding Roles for users and teams mapping


arnav3000:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.
- [X] I am **NOT** reporting a (potential) security vulnerability. (These should be emailed to `security@ansible.com` instead.)

### Bug Summary

While adding role mapping to users and teams using configuration as code the output says mapping is successful however when I check in Ansible controller UI, I do not see any user being mapped to a team or any role.

roles.yml :
---

controller_roles_dev:
  - user: john
    target_team: team_storage_dev_devconf
    role: member
    
  - user: sita
    target_team: team_unix_dev_devconf
    role: member

  - user: hanuman
    target_team: team_storage_admin_devconf
    role: member

  - user: ram
    target_team: team_unix_admin_devconf
    role: member

  - team: team_unix_admin_devconf
    organization: org_unix
    role: execute

  - team: team_unix_dev_devconf
    organization: org_unix
    role: execute

  - team: team_unix_dev_devconf
    organization: org_unix
    role: job_template_admin

  - team: team_storage_admin_devconf
    organization: org_storage
    role: execute

  - team: team_storage_dev_devconf
    organization: org_storage
    role: execute

  - team: team_storage_dev_devconf
    job_templates:
      - jt_configure_webserver_rhel86
      - jt_configure_services_rhel86
    role: execute

...

Please find attached screenshots and the sample file I used.
<img width="1779" alt="Screenshot 2023-06-12 at 19 25 26" src="https://github.com/ansible/awx/assets/105802773/4ba6bfd7-f693-4967-822d-3a0ebd456f85">

<img width="1779" alt="Screenshot 2023-06-12 at 22 41 12" src="https://github.com/ansible/awx/assets/105802773/fb728244-d05c-4e23-94aa-38bc29aa3c01">

### AWX version

22.3.0

### Select the relevant components

- [X] UI
- [ ] UI (tech preview)
- [ ] API
- [ ] Docs
- [ ] Collection
- [X] CLI
- [ ] Other

### Installation method

N/A

### Modifications

no

### Ansible version

ansible [core 2.14.0]

### Operating system

RHEL-86

### Web browser

Chrome

[...]

PR #13784, opened 2023-03-30
Add instance groups role to awxkit and awx collection

tanganellilore:
##### SUMMARY

This pull request will enable instance_groups role to awxkit and awx collection.
Moreover will close https://github.com/ansible/awx/issues/13783

##### ISSUE TYPE

 - Bug, Docs Fix or other nominal change

##### COMPONENT NAME

 - Collection
 - CLI

##### AWX VERSION

```
21.14.0
```

##### ADDITIONAL INFORMATION

```

```

PR #14119, opened 2023-06-13
bugfix collection role module target_teams and instance_groups options

sean-m-sullivan:
##### SUMMARY
Bugfix for #14113

Previously the target teams would be added to actor data, the two options to fix was this, or sepreating their lookup. 

##### ISSUE TYPE

##### COMPONENT NAME
 - Collection

##### AWX VERSION
```
22.3.0
```

### DIFF
--- a/awx_collection/plugins/modules/role.py
+++ b/awx_collection/plugins/modules/role.py
@@ -245,12 +245,11 @@ def main():
             resources.setdefault(resource_group, []).append(module.params.get(old_name))
     if module.params.get('lookup_organization') is not None:
         resources['lookup_organization'] = module.params.get('lookup_organization')
-
-    # Change workflows and target_teams key to its endpoint name.
+    if module.params.get('instance_groups') is not None:
+        resources['instance_groups'] = module.params.get('instance_groups')
+    # Change workflows to its endpoint name.
     if 'workflows' in resources:
         resources['workflow_job_templates'] = resources.pop('workflows')
-    if 'target_teams' in resources:
-        resources['teams'] = resources.pop('target_teams')
 
     # Set lookup data to use
     lookup_data = {}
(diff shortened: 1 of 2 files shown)


awx_collection/plugins/modules/role.py
```diff
@@ -272,12 +270,18 @@ def main():
                     lookup_data_populated = {}
                 else:
                     lookup_data_populated = lookup_data
-            data = module.get_one(key, name_or_id=resource, data=lookup_data_populated)
+                if key == 'target_teams':
+                    lookup_key = 'teams'
+                else:
+                    lookup_key = key
```
AlanCoding: I don't like the indent of this. If the condition `key in resources` is not met, then `lookup_key` stands to be undefined and then referenced on the `module.get_one` line.

awx_collection/plugins/modules/role.py
```diff
@@ -249,8 +249,6 @@ def main():
     # Change workflows and target_teams key to its endpoint name.
```
AlanCoding: Comment is wrong now, this only changes workflows here.

awx_collection/plugins/modules/role.py
```diff
@@ -268,18 +266,23 @@ def main():
         for resource in value:
             # Attempt to look up project based on the provided name or ID and lookup data
             if key in resources:
```
AlanCoding: I think you're dancing around some defunct code. Consider the main loop/condition

```python
    for key, value in resources.items():
        for resource in value:
            # Attempt to look up project based on the provided name or ID and lookup data
            if key in resources:
```

Before your change, not everything was nested under `if key in resources:`, so its issues didn't stand out as much.

First impressions reading this block, the ordering of the conditions is wrong. `key` is an iteration variable of the outer look, in which case you should never check it inside of the _inner_ loop. So that would dictate switching the 2nd `for` and the `if`.

But if you do that, then the more fundamental problem readily becomes apparent. `key` should never not be in `resources`, unless the dictionary is modified inside of the loop. I looked at the history, and I don't know if it ever was modified in the loop, and it certainly isn't anymore.

So it appears that `if key in resources:` should be deleted, because it is a condition which should always be true.

sean-m-sullivan: Went through and looked back on what this was based on.
https://github.com/ansible/awx/blob/22.0.0/awx_collection/plugins/modules/role.py#L245

it was a simplification at the time and you are right we don't need the if, or even another block, we just need the basic for as its a lookup of module parameters that were added in, and then need to be sorted to their respective set. actor_data and resource_data

I also discovered the instance group change was done incorrectly in #13784, where it was never added to a field to get the parameter, so I added it and tested it. 

I also removed an option in one of the tests so the plurality is tested on its own instead of With a singular which led to the tests not detecting this issue.

AlanCoding (approved):
This code looks much better than before, I want you to merge it. The role of the loop over `resources` makes sense now with a detailed reading.

sean-m-sullivan:
@AlanCoding I am not able to merge or run workflows, please advance this as you want to proceed.

PR #14120, opened 2023-06-13
Change parsing of docker info in dev build

jjwatt:
This is a non-functional change. The way os_info is populated with docker info and grep 'Operating System' breaks on podman and likely in other places. This makes it work on both podman and docker, and it will continue to return the exact same strings everywhere else.

##### SUMMARY

Fix call in dev build to get os_info from `docker info` so that it works on podman.

##### ISSUE TYPE

##### COMPONENT NAME

 - Other

##### AWX VERSION

```
devel
```

##### ADDITIONAL INFORMATION

```

```


jjwatt:
Important: We should really figure out what this `os_info` thing is for. As best I can tell, it gets passed to docker-compose/containers as an "OS" environment variable. But, let me tell you something: This is what the value looks like, literally:
-  Operating System: Fedora Linux 38 (Workstation Edition)
-  Operating System: NixOS 22.05 (Quokka)

Those strings are going into `os_info.stdout` (and you have to remember that it has a stdout which is annoying, too). But, why? And why does it have to be pulled from docker? There are so many better places to get the OS info and maybe even better places in that output. How would strings like the above ever be useful? Who's parsing those?

shanemcd (approved):

shanemcd:
IIRC we only use this to detect when we're running on a Mac:

https://github.com/ansible/awx/blob/7e7a9f541cae5f2da1b81e65a6bd55e2dfbb587f/tools/docker-compose/entrypoint.sh#L29-L33

jjwatt:
> IIRC we only use this to detect when we're running on a Mac:
> 
> https://github.com/ansible/awx/blob/7e7a9f541cae5f2da1b81e65a6bd55e2dfbb587f/tools/docker-compose/entrypoint.sh#L29-L33

That seems like a very legitimate use case for passing in that string to the container(s)! Since we've identified how it's used, we can probably better name/use it in the "staging" part. Like, we could even make a bool of that string that we pass through. If we really want to case over an os string instead of just looking for a static substring, then we probably would not be using this `$OS`, `os_info` for that. So we'd either end up with multiple "OS" vars or we'd end up changing this then. We should probably make a bool for the docker desktop case. I'll try to do that in a follow-up PR after this one gets merged.

PR #13867, opened 2023-04-17
OPTIONS-based CLI associations

AlanCoding:
##### SUMMARY
This is a fairly raw Proof of Concept, but I can demo the basic elements of functionality here.

What issue would this resolve? Several. I had this filter I've been using to demonstrate the problem.

https://github.com/ansible/awx/issues?q=is%3Aissue+is%3Aopen+label%3Acomponent%3Acli+associate

There are 3 issues hit by that filter that say "Unable to ____" in some sense. In response to those issues, this provides the association command as follows:

```
$ awx group associate
usage: awx groups associate [-h] (--children  | --hosts ) id

positional arguments:
  id           Primary object for the association, obtained from groups endpoint

options:
  -h, --help   show this help message and exit
  --children   The ID (or name) of the children to associate
  --hosts      The ID (or name) of the hosts to associate
```

Now you can not only associate hosts to groups, but _fully_ build inventories

```
$ awx job_template associate
usage: awx job_templates associate [-h]
                                   (--credentials  | --notification_templates_started  | --notification_templates_error  | --notification_templates_success  | --instance_groups  | --labels )
                                   id

positional arguments:
  id                    Primary object for the association, obtained from job_templates endpoint

options:
  -h, --help            show this help message and exit
  --credentials         The ID (or name) of the credentials to associate
  --notification_templates_started 
                        The ID (or name) of the notification_templates_started to associate
  --notification_templates_error 
                        The ID (or name) of the notification_templates_error to associate
  --notification_templates_success 
                        The ID (or name) of the notification_templates_success to associate
  --instance_groups     The ID (or name) of the instance_groups to associate
  --labels              The ID (or name) of the labels to associate
```

Not only can you associate _labels_ as the issue asked for, but you can associate other things you haven't thought of yet.

```
$ awx organization associate
usage: awx organizations associate [-h]
                                   (--users  | --admins  | --execution_environments  | --teams  | --notification_templates  | --notification_templates_started  | --notification_templates_error  | --notification_templates_success  | --notification_templates_approvals  | --instance_groups  | --galaxy_credentials )
                                   id

positional arguments:
  id                    Primary object for the association, obtained from organizations endpoint

options:
  -h, --help            show this help message and exit
  --users               The ID (or name) of the users to associate
  --admins              The ID (or name) of the admins to associate
  --execution_environments 
                        The ID (or name) of the execution_environments to associate
  --teams               The ID (or name) of the teams to associate
  --notification_templates 
[...]

PR #14128, opened 2023-06-15
 In collection, Add an enforced state to the modules

sean-m-sullivan:
##### SUMMARY
Follow up PR from #13927, #13960

This enforces default values being submitted to the API, this PR is an overview of how that enforcement mechanism would be achieved across the collection, This PR is for feedback on how this was achieved so that I can go through and add it to the rest of the modules. The idea is to enforce a state to prevent drift when an option is not provided and option is enforced. This is compared to exists that makes no changes if the object exists, or present that only makes changes for the options given to the module.

However some are the same as the default value set in the get parameter, others are related fields that have no defaults in the options. The idea of this PR is to do two modules to show how it would be implemented, so it can be approved, before it is extrapolated to every module.

In addition while going through things
Credential types:
      Removed managed, it technically does nothing on post, unable to change it through API.

Applications:
      Fixed bug where Skip authorization was never sent to API. Tried to track down but this has been this way in the code for 3 years.

Added blocks to following tests and brought to the standard way we do the other tests with an always to cleanup after everything.
    groups 
    job templates
    organizations
    credentials
    hosts
    inventory sources
    teams

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - Collection

##### AWX VERSION

```
22.4.0
```

### DIFF
--- a/awx_collection/plugins/module_utils/controller_api.py
+++ b/awx_collection/plugins/module_utils/controller_api.py
@@ -340,7 +340,305 @@ class ControllerAPIModule(ControllerModule):
         return self.make_request('GET', endpoint, **kwargs)
 
     def get_options_endpoint(self, endpoint, *args, **kwargs):
-        return self.make_request('OPTIONS', endpoint, **kwargs)
+        return self.make_request('OPTIONS', endpoint, **kwargs)['json']['actions']['POST']
+
+    def get_enforced_defaults(self, endpoint, *args, **kwargs):
+        endpoint_defaults = self.get_options_endpoint(endpoint)
+
+        default_fields = {
+            'applications': {
+                'new_fields': {
+                    'description': '' if endpoint != 'applications' else endpoint_defaults['description']['default'],
+                    'client_type': 'public',
+                    'redirect_uris': '' if endpoint != 'applications' else endpoint_defaults['redirect_uris']['default'],
+                    'authorization_grant_type': 'password',
+                    'skip_authorization': False if endpoint != 'applications' else endpoint_defaults['skip_authorization']['default'],
+                },
+                'association_fields': {},
+            },
+            'credentials': {
+                'new_fields': {
+                    'description': '' if endpoint != 'credentials' else endpoint_defaults['description']['default'],
+                    'inputs': {} if endpoint != 'credentials' else endpoint_defaults['inputs']['default'],
+                },
+                'association_fields': {},
+            },
+            'credential_types': {
+                'new_fields': {
+                    'description': '' if endpoint != 'credential_types' else endpoint_defaults['description']['default'],
+                    'kind': 'cloud',
+                    'inputs': {} if endpoint != 'credential_types' else endpoint_defaults['inputs']['default'],
+                    'injectors': {} if endpoint != 'credential_types' else endpoint_defaults['injectors']['default'],
+                },
+                'association_fields': {},
+            },
+            'credential_input_sources': {
+                'new_fields': {
+                    'description': '' if endpoint != 'credential_input_sources' else endpoint_defaults['description']['default'],
+                    'metadata': {} if endpoint != 'credential_input_sources' else endpoint_defaults['metadata']['default'],
+                },
+                'association_fields': {},
+            },
+            'execution_environments': {
+                'new_fields': {
+                    'description': '' if endpoint != 'execution_environments' else endpoint_defaults['description']['default'],
+                    'pull': 'missing',  # Module default, is '' in the API.
+                },
+                'association_fields': {},
+            },
+            'groups': {
+                'new_fields': {
+                    'description': '' if endpoint != 'groups' else endpoint_defaults['description']['default'],
+                    'variables': '' if endpoint != 'groups' else endpoint_defaults['variables']['default'],
+                },
+                'association_fields': {
+                    'hosts': [],
+                    'children': [],
+                },
+            },
+            'hosts': {
+                'new_fields': {
+                    'description': '' if endpoint != 'hosts' else endpoint_defaults['description']['default'],
+                    'variables': '' if endpoint != 'hosts' else endpoint_defaults['variables']['default'],
+                    'enabled': True if endpoint != 'hosts' else endpoint_defaults['enabled']['default'],
+                },
+                'association_fields': {},
+            },
+            'instance_groups': {
+                'new_fields': {
+                    'max_concurrent_jobs': 0 if endpoint != 'instance_groups' else endpoint_defaults['max_concurrent_jobs']['default'],
+                    'max_forks': 0 if endpoint != 'instance_groups' else endpoint_defaults['max_forks']['default'],
+                    'is_container_group': False,
+                    'policy_instance_percentage': 0 if endpoint != 'instance_groups' else endpoint_defaults['policy_instance_percentage']['default'],
+                    'policy_instance_minimum': 0 if endpoint != 'instance_groups' else endpoint_defaults['policy_instance_minimum']['default'],
+                    'policy_instance_list': [],
+                    'pod_spec_override': '' if endpoint != 'instance_groups' else dumps(endpoint_defaults['pod_spec_override']['default']),
+                },
+                'association_fields': {
+                    'instances': [],
+                },
+            },
+            'instances': {
+                'new_fields': {
+                    'capacity_adjustment': 1.0 if endpoint != 'instances' else endpoint_defaults['capacity_adjustment']['default'],
+                    'enabled': True if endpoint != 'instances' else endpoint_defaults['enabled']['default'],
+                    'managed_by_policy': True if endpoint != 'instances' else endpoint_defaults['managed_by_policy']['default'],
+                    'node_type': 'execution' if endpoint != 'instances' else endpoint_defaults['node_type']['default'],
+                    'listener_port': 27199 if endpoint != 'instances' else endpoint_defaults['listener_port']['default'],
+                },
+                'association_fields': {},
+            },
+            'inventory_sources': {
+                'new_fields': {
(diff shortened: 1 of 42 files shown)


awx_collection/plugins/modules/organization.py
```diff
@@ -159,49 +167,65 @@ def main():
         association_fields['instance_groups'] = []
         for item in instance_group_names:
             association_fields['instance_groups'].append(module.resolve_name_to_id('instance_groups', item))
+    elif instance_group_names is None and enforced_defaults:
+        association_fields['instance_groups'] = []
```
john-westcott-iv: @AlanCoding would we be good with always assuming associations are `[]` by default? I think its probably find and I don't think we can extract it from the API anywhere but if we every did something like "auto adding" associations on new models this would break and have to be logic we apply in the modules themselves.

awx_collection/plugins/modules/organization.py
```diff
@@ -159,49 +167,65 @@ def main():
...
 
     notifications_start = module.params.get('notification_templates_started')
     if notifications_start is not None:
         association_fields['notification_templates_started'] = []
         for item in notifications_start:
             association_fields['notification_templates_started'].append(module.resolve_name_to_id('notification_templates', item))
+    elif notifications_start is None and enforced_defaults:
+        association_fields['notification_templates_started'] = []
```
john-westcott-iv: Instead of doing a bunch of `and enforced_defaults` could we just default the `association_fields` variable if enforced is true?
```
elif state == 'enforced':
    association_fields = {
         'instance_groups': []
         'notification_templates_started': []
         ...
    }
```

We could also do the same with org_fields.... or maybe even make this more generic with having a module call like:
```
   org_fields, association_fields = module.get_enforced_state('organizations')
 ```
 
You'd have to adjust the existing initialization of these fields but this might make a better pattern for all of the endpoints.

**Review by john-westcott-iv — COMMENTED** [review]
Are you planning on adding `enforced` to the remaining modules over time?

File: awx_collection/plugins/modules/organization.py (hunk 1 above)
Comment (sean-m-sullivan) [reply awx_collection/plugins/modules/organization.py]: If you make associations [] by default, it REMOVES all associations when submitted, which is the point of the enforced, and config as code. It is also a breaking change to how the modules work, unless we want enforced state to be default.

Comment (sean-m-sullivan) [reply awx_collection/plugins/modules/organization.py]: Closing this one as it requires breaking changes.

File: awx_collection/plugins/module_utils/controller_api.py [hunk 3]
```diff
@@ -339,6 +339,47 @@ def head_endpoint(self, endpoint, *args, **kwargs):
...
+    def get_enforced_defaults(self, endpoint, *args, **kwargs):
+        endpoint_defaults = self.get_options_endpoint(endpoint)
+
+        default_fields = {
+            'organizations': {
+                'new_fields': {
+                    'description': '' if endpoint != 'organizations' else endpoint_defaults['description']['default'],
+                    'default_environment': '',
+                    'custom_virtualenv': '',
+                    'max_hosts': 0 if endpoint != 'organizations' else endpoint_defaults['max_hosts']['default'],
```
john-westcott-iv, line 534: The associations might have to be manually created for now but `new_fields` we should be able to get from `endpoint_defaults` right? i.e. something like: 
```
for option_name in response.json['POST']:
    try:
        new_fields['optoins_name'] = respose.json['actions']['POST'][option_name].get('default', 'awx_collection_undefined')
    except:
        raise Exception(f'{option_name} for model {endpoint} does not have a default!')
```
We might need to do a little bit of digging to make sure all defaults have been defined. Or if not, if there is a general default we can use like `""`.

Associations I'd like to do programatic too but I need to think about how we can accomplish that. Maybe we need the actual modules to construct that since the options page currently does not give us that information.

awx_collection/plugins/modules/schedule.py
```diff
@@ -272,8 +274,13 @@ def main():
     if state == 'absent':
         # If the state was absent we can let the module delete it if needed, the module will handle exiting from this
         module.delete_if_needed(existing_item)
-
-    association_fields = {}
+    elif state == 'enforced':
+        enforced_defaults = True
+        new_fields, association_fields = module.get_enforced_defaults('schedules')
```
john-westcott-iv, line 277: This feel much cleaner!

awx_collection/plugins/module_utils/controller_api.py
sean-m-sullivan: The if else, is because, if you use the try except on them all, and you call job templates, it doesn't return information about schedules, it will except, and exit

By using if else, it will pull from api, for only the endpoint passed.

If we had a source of truth for this on the api, generated, with all endpoints populated, we could just pull in that structure

I looked at adding additional fields, or do something like it for each endpoint, but wasn't able to suss out how we did it.

gamuniz:
@sean-m-sullivan should we open a different PR for the linting and keep the changes to what you're expecting the code to do here?

awx_collection/plugins/modules/organization.py
sean-m-sullivan: Done in latest

sean-m-sullivan:
@gamuniz I rebased, but unsure what linting you are referring to, if you can point me I can fix, but can work with what the team needs

AlanCoding:
I know you put a lot of work into this, and I apologize for taking a long time to get back to it. So let me talk out loud about potential pitfalls.

While good in theory, I'm concerned that the _API itself_ may be inconsistent, and give problems about consistency with the modules. How do we express that in a test case? I want to imagine creating a resource of every time with the minimal number of fields set. Imagine creating a job template (obviously, the messiest resource), and in this thought experiment obviously we will try to _not set_ the `extra_vars` field. A criteria I'm interested in is consistency of the "changed" status. What if the API _reports_ a certain default value, but when saving it, performs a transformation - say OPTIONS gives `""` for the default but after saving it has `"{}"` _that's_ what I'm worried about the API doing. The test steps which would correspond to this would be (1) create with state=enforced and then (2) re-run the exact same task. The re-run of the task should report changed=false.

Hypothetically, we might want to this this kind of thing for every resource type, but I want to be practical. If we don't have that sort of task for job_template... that's probably a good idea. If manual testing is already confirming that this is behaving as expected, that would give me good confidence too.

This is basically the logical track I would take myself, if I were to pull this down and try to verify.

sean-m-sullivan:
@AlanCoding Its a good question, and with my understanding it would NOT change, but that was just an assumption based on what I new of the code.

The Job template extra vars is an interesting example, but a strange one, partially because we transform it in the module, 
```
    # Special treatment of extra_vars parameter
    extra_vars = module.params.get('extra_vars')
    if extra_vars is not None:
        new_fields['extra_vars'] = json.dumps(extra_vars)
```
as the api expects a json string. which is why the default is ""

However I did make a playbook to test the issue, and it passed, and is how it should behave. If a default value is passed to the api, the api should treat it the same, if it isn't, I'd question the api. 
```
---
- name: Playbook to test enforced state
  hosts: localhost
  connection: local
  gather_facts: false

  environment:
    CONTROLLER_HOST: controller.nas
    CONTROLLER_USERNAME: admin
    CONTROLLER_PASSWORD: secret123
    CONTROLLER_VERIFY_SSL: false
    controller_state: enforced
 
  tasks:
    - name: Create job template
      awx.awx.job_template:
        name: test-template-1
        description: created by Ansible Playbook
        job_type: run
        inventory: RHVM-01
        credentials: admin@internal-RHVM-01
        project: Test Project
        playbook: helloworld.yml
        verbosity: 2
        extra_vars:
          target_hosts: infra-ansible-tower-01.example.com
        ask_execution_environment_on_launch: true
        ask_forks_on_launch: true
        ask_instance_groups_on_launch: true
        ask_job_slice_count_on_launch: true
        ask_labels_on_launch: true
        ask_timeout_on_launch: true
        prevent_instance_group_fallback: true
      register: result

    - assert:
        that:
          - "result is changed"

    - name: Create job template enforced
      awx.awx.job_template:
        name: test-template-1
        state: enforced
      register: result

    - assert:
        that:
          - "result is changed"

    - name: Create job template enforced
      awx.awx.job_template:
        name: test-template-1
        state: enforced
      register: result

    - assert:
        that:
          - "result is not changed"
...
```
With that, I added 15 tests that are a duplicate of the test above to this PR to test 15 modules.
The exceptions are:
User - because of password
Notification template - likely the token used. Not sure on this one why it reports changed, but it does.

sean-m-sullivan:
Also I want to put this up here, as its currently buried in the resolved.

The hardcoded default list is a start
We should add it into the API as an endpoint to pull from, of set new and associated fields derived from their sources.

However I am unsure of how to implement that on my own, so went with what worked now, instead of going for the perfect solution.

AlanCoding:
I agree with @john-westcott-iv 's sentiment expressed above, and I believe any stop-gap solution will become permanent.

I want to talk about the function of that dictionary, because I see it tracking a lot of _associations_. I have a PoC up for something unrelated at https://github.com/ansible/awx/pull/13867 in which I realized that our OPTIONS don't give any usable breadcrumbs for identifying what possible associations exist for an object. That adds a `"related_associations"` field to the top-level of OPTIONS which allows inferring the related link, like "credentials" for `/api/v2/job_templates/7/credentials/`, and I want to ask if maybe this might help solve the problem you're dealing with here as well.

sean-m-sullivan:
@AlanCoding I can add in the change to get associations similarly to the others, I'd have to see how its used to get a definitive answer. I think the only thing I can see is that if we go for perfect on v1, we are never going to get there as there are tons of little inconsistencies, better to build it up over time. IMO it would be great just to have an endpoint that derives Everything and we just pull in that data structure, 1 pull, gets everything, derives from there, that is essentially what the var is now. just in non API form.

PR #14132, opened 2023-06-15
[hop node] update peer validation logic 

thedoubl3j:
##### SUMMARY

@relrod pointed out a bug in the validation logic for instance type when we are checking peers. this change moves the logic from membership to comparison since the code previously was working by happen stance since "control" was in the tuple/enum. 

I found no other instance of this in the current branch that is _new_ there are other instances of this type of membership logic but it is out of scope for the feature (and is also valid use of the membership logic IMO)

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
devel
```

##### ADDITIONAL INFORMATION

Link to Rick's comment for context if folks want to see the full explanation https://github.com/ansible/awx/pull/13904#discussion_r1206104180

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -5491,7 +5491,7 @@ class InstanceSerializer(BaseSerializer):
         if peers_from_control_nodes and node_type not in (Instance.Types.EXECUTION, Instance.Types.HOP):
             raise serializers.ValidationError(_("peers_from_control_nodes can only be enabled for execution or hop nodes."))
 
-        if node_type in (Instance.Types.CONTROL):
+        if node_type == (Instance.Types.CONTROL):
             if self.instance and 'peers' in attrs and set(self.instance.peers.all()) != set(attrs['peers']):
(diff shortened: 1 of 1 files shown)


djyasin (approved):

awx/api/serializers.py
```diff
@@ -5491,7 +5491,7 @@ def get_field_from_model_or_attrs(fd):
         if peers_from_control_nodes and node_type not in (Instance.Types.EXECUTION, Instance.Types.HOP):
             raise serializers.ValidationError(_("peers_from_control_nodes can only be enabled for execution or hop nodes."))
 
-        if node_type in (Instance.Types.CONTROL):
+        if node_type == (Instance.Types.CONTROL):
```
relrod: I am guessing one of the linters would catch this, but:

```suggestion
        if node_type == Instance.Types.CONTROL:
```

fosterseth (approved):

PR #6788, opened 2020-04-21
include the AWX version as a header in all responses

ryanpetrello:
https://github.com/ansible/awx/pull/6743#discussion_r411067456

image

PR #6743, opened 2020-04-17
Adding version checking to collection

john-westcott-iv:
##### SUMMARY

Make collection check version of Tower compared to the version of the collection and warn on mismatch.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - Collection

##### AWX VERSION

```
awx: 11.0.0
```

##### ADDITIONAL INFORMATION

```

```

PR #14135, opened 2023-06-15
Only show the product version header when the requester is authenticated

rooftopcellist:
##### SUMMARY

We should only show the product version when the user is authenticated as that could be considered sensitive information. 

##### ISSUE TYPE

##### COMPONENT NAME

 - API
 

##### AWX VERSION

```
Present in latest awx version 22.3.0
```

### DIFF
--- a/awx_collection/plugins/module_utils/controller_api.py
+++ b/awx_collection/plugins/module_utils/controller_api.py
@@ -533,13 +533,7 @@ class ControllerAPIModule(ControllerModule):
                 controller_version = response.info().getheader('X-API-Product-Version', None)
 
             parsed_collection_version = Version(self._COLLECTION_VERSION).version
-            if not controller_version:
-                self.warn(
-                    "You are using the {0} version of this collection but connecting to a controller that did not return a version".format(
-                        self._COLLECTION_VERSION
-                    )
-                )
-            else:
+            if controller_version:
                 parsed_controller_version = Version(controller_version).version
                 if controller_type == 'AWX':
                     collection_compare_ver = parsed_collection_version[0]
--- a/awx/api/generics.py
+++ b/awx/api/generics.py
@@ -232,7 +232,8 @@ class APIView(views.APIView):
 
         response = super(APIView, self).finalize_response(request, response, *args, **kwargs)
         time_started = getattr(self, 'time_started', None)
-        response['X-API-Product-Version'] = get_awx_version()
+        if request.user.is_authenticated:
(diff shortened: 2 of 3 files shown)


TheRealHaoLiu (approved):

TheRealHaoLiu:
adding some context that @rooftopcellist mentioned on Slack

I'm looking a bit deeper now... It looks like Ryan P added this here:
https://github.com/ansible/awx/pull/6788
And he added it so that the awx.awx collection could throw version mismatch (between awx collection and awx) warnings
https://github.com/ansible/awx/pull/6743/files

So we might also want to add a line to not throw the warning at all if the version header was not present

rooftopcellist:
I added some logic to address @TheRealHaoLiu 's comment.  I haven't explicitly tested it with the collection use case though.

awx_collection/plugins/module_utils/controller_api.py
```diff
@@ -531,7 +531,16 @@ def make_request(self, method, endpoint, *args, **kwargs):
...
-            if not controller_version:
+            unauthenticated_endpoints = [
+                '/api/',
+                '/api',
+                '/api/v2/',
+                '/api/v2',
+                '/ping',
+                '/ping/',
+                '/authtoken/'
+            ]
```
AlanCoding: What I'm concerned about is that someone is unauthenticated (or in the process of authenticating), makes a request, logically gets `None` for `controller_version`, and thus drops into the `else:` here which first does `parsed_controller_version = Version(controller_version).version`. Importing the class used:

```python
>>> from ansible.module_utils.compat.version import LooseVersion as Version
>>> Version(None)
  File "/home/alancoding/repos/ansible/lib/ansible/module_utils/compat/version.py", line 328, in __repr__
    return "LooseVersion ('%s')" % str(self)
                                   ^^^^^^^^^
  File "/home/alancoding/repos/ansible/lib/ansible/module_utils/compat/version.py", line 325, in __str__
    return self.vstring
           ^^^^^^^^^^^^
AttributeError: 'LooseVersion' object has no attribute 'vstring'
```

This would likely error and cause trouble.

As I read this, I highly suspect the original author treated a `None` value as the case where it was running against an _old_ version which obviously did not have the header. That is less of a concern now, as the header has been in released versions for years. The safest option may be to do nothing (no warning) in the None case.

jbradberry (approved):

djyasin (approved):

AlanCoding (approved):

PR #14143, opened 2023-06-20
Remove reference to unmaintained runner image

AlanCoding:
##### SUMMARY
This default is used when doing `factories.execution_environment()` with awxkit. With builder v3, this image is sun-setted. @nitzmahone 

I don't _think_ that pinning this to the awx-ee (already the default EE) will cause a problem?

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awxkit/awxkit/api/pages/execution_environments.py
+++ b/awxkit/awxkit/api/pages/execution_environments.py
@@ -20,7 +20,7 @@ class ExecutionEnvironment(HasCreate, HasCopy, base.Base):
     NATURAL_KEY = ('name',)
 
     # fields are name, image, organization, managed, credential
-    def create(self, name='', image='quay.io/ansible/ansible-runner:devel', organization=Organization, credential=None, pull='', **kwargs):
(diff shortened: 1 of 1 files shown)


nitzmahone:
I'd argue that having a default on that arg is probably more a liability than a help, as it could mask all sorts of weird problems if someone forgets to pass the arg but things still "work". But no, off the top of my head I can't think of a reason that switching to a more regularly maintained image should make it worse...

The more I think about that default though, seems like a good opportunity to get rid of it altogether :wink:

AlanCoding:
That doesn't make sense in this case. If you intend to use the default (which is the awx-ee upstream) then you should _not create an execution environment_, which is fundamentally what `factories.execution_environment()` does. If you don't have a custom image to run jobs out of, then you have no business creating a EE via the UI or API. We are very much obligated to test the action of using a custom EE, and we probably run jobs out of it in some cases. That's why I was always grasping for a public image off-the-shelf available for testing, which isn't that challenging to maintain, although I like how builder v3 doesn't make these intermediaries into critical infrastructure.

shanemcd (approved):

Issue #14123, opened 2023-06-14
[hop node] add peers option to awx collection


fosterseth:
- add a `peers` option for instance creation
- add node type option for `hop`
- add listener_port option

PR #14153, opened 2023-06-21
Hop node AWX Collection Updates

djyasin:
##### SUMMARY

Related to issue #14123 

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - Collection

##### AWX VERSION

```
22.3.1
```

##### ADDITIONAL INFORMATION

- [x] - add a `peers` option for instance creation
- [x] - add `peers_from_control_nodes` option 
- [x] - add node type option for `hop`
- [x] - add listener_port option
- [x] - add updated information to collection documentation

```

```


awx_collection/plugins/modules/instance.py
```diff
@@ -47,6 +47,7 @@
         - Role that this node plays in the mesh.
       choices:
         - execution
+        - hop
       required: False
       type: str
     node_state:
```
gamuniz, line 53: should we add the peers to the documentation under options?

djyasin: Added that to the documentation! Thank you for the reminder 😄

awx_collection/plugins/modules/instance.py
```diff
@@ -88,9 +94,10 @@ def main():
         capacity_adjustment=dict(type='float'),
         enabled=dict(type='bool'),
         managed_by_policy=dict(type='bool'),
-        node_type=dict(type='str', choices=['execution']),
+        node_type=dict(type='str', choices=['execution', 'hop']),
         node_state=dict(type='str', choices=['deprovisioning', 'installed']),
         listener_port=dict(type='int'),
+        peers=dict(type='str'),
```
fosterseth: in the API this field can accept a list of strings
```
peers: ["hop1", "hop2"]
```
right now your code only allows a single string, which would result in `peers: "hop1"`, which isn't what the api expects.

this is what a list looks like in an ansible playbook
```yaml
peers:
  - hop1
  - hop2
```

you might look at the instances field in the instance group module to see how a list of strings is handled there

awx_collection/plugins/modules/instance.py
```diff
@@ -62,6 +63,17 @@
         - Port that Receptor will listen for incoming connections on.
       required: False
       type: int
+    peers:
+      description:
+        - Peer to a hop or execution node from a control node.
```
fosterseth:
```suggestion
        - List of peers to connect outbound to. Only configurable for hop and execution nodes.
```

awx_collection/plugins/modules/instance.py
```diff
@@ -62,6 +63,17 @@
...
+      required: False
+      type: list
+      elements: str
+    peers_from_control_nodes:
+      description:
+        - Set hop or execution node to peer from a control node.
```
fosterseth:
```suggestion
        - If enabled, control plane nodes will automatically peer to this node.
```

awx_collection/plugins/modules/instance.py
```diff
@@ -104,7 +118,8 @@ def main():
     node_type = module.params.get('node_type')
     node_state = module.params.get('node_state')
     listener_port = module.params.get('listener_port')
-
+    peers = module.params.get('peers')
+    peers_from_control_nodes =module.params.get('peers_from_control_nodes')
```
fosterseth:
```suggestion
    peers_from_control_nodes = module.params.get('peers_from_control_nodes')
```

thedoubl3j (approved):
parameters are documented and looks clean to me.

djyasin:
Added logic to remove peers. Set to empty list `[]`

djyasin:
@fosterseth Should I add how to remove peers to the docs as well?

awx_collection/plugins/modules/instance.py
```diff
@@ -122,6 +137,12 @@ def main():
         new_fields['node_state'] = node_state
     if listener_port is not None:
         new_fields['listener_port'] = listener_port
+    if peers is not None:
+        new_fields['peers'] = peers
+    if peers is None:
```
fosterseth, line 143: if a user leaves off the option `peers`, does that remove peers?

I think ideally the only way to remove peers is to do something like

```yaml
hostname: hop
node_type: execution
peers:
```

so it is user defined, but just set to empty. If peers is not set, then we should just keep current value.

fosterseth:
and yes, good idea we should include an example on how to unpeer in the Example section of the module

awx_collection/plugins/modules/instance.py
```diff
@@ -62,6 +63,18 @@
         - Port that Receptor will listen for incoming connections on.
       required: False
       type: int
+    peers:
+      description:
+        - List of peers to connect outbound to. Only configurable for hop and execution nodes.
+        - To remove the peer relationship from a node, provide that node with an empty list.
```
fosterseth:
```suggestion
        - To remove all current peers, set value to an empty list, [].
```

awx_collection/tests/integration/targets/instance/tasks/main.yml
```diff
@@ -70,3 +70,60 @@
...
+        node_state: installed
+        listener_port: 27199
+        peers: []
+      register: result
+
+    - assert:
+        that:
+          - result is changed
+
+
```
fosterseth: we'll need the `when: IS_K8S` like the above block

fosterseth:
I think we may wish to add a similar set a tests here https://github.com/fosterseth/awx/blob/05c7d3a60ed78f5389f607feb05318014e91299e/awx_collection/test/awx/test_inventory.py#L60-L61

add a new test that creates your 2 hop nodes and 1 execution node.

1. Use run_module to peer execution node to the hop nodes, then assert that execution peers has those 2 nodes (at the database level)
2. Remove the peers via run_module, and assert the changes are made at the database level

awx_collection/tests/integration/targets/instance/tasks/main.yml
```diff
@@ -70,3 +70,60 @@
...
+        node_state: installed
+        listener_port: 27199
+        peers: []
+      register: result
+
+    - assert:
+        that:
+          - result is changed
+
+  when: IS_K8S
```
fosterseth:
```suggestion
  when: IS_K8S

```

awx_collection/test/awx/test_instance.py
```diff
@@ -0,0 +1,53 @@
...
+__metaclass__ = type
+
+import pytest
+
+from awx.main.models import Instance
+from django.test.utils import override_settings
+
+
+@pytest.mark.django_db
+def test_hop_node_scenario(run_module, admin_user):
```
fosterseth:
```suggestion
def test_peers_adding_and_removing(run_module, admin_user):
```

fosterseth (approved):

PR #14157, opened 2023-06-21
Add Request time out option for collection

sean-m-sullivan:
##### SUMMARY
Found issue when updating a large number of objects, especially using async, where the client would timeout and cancel the request causing time out errors giving the message:
 "msg": "There was an unknown error when trying to connect to https://aap.controller/api/v2/project_updates/141901/: timeout The read operation timed out", "results_file": 
 
 After viewing /var/log/nginx/automationcontroller.access.log errored 499 coming up as 
 10.242.42.191 - - [20/Jun/2023:12:49:38 -0400] "GET /api/v2/projects/51/notification_templates_started/ HTTP/1.1" 499 0 "-" "Python-urllib/3.11" "-"
10.242.42.191 - - [20/Jun/2023:12:49:38 -0400] "GET /api/v2/project_updates/178/ HTTP/1.1" 499 0 "-" "Python-urllib/3.11" "-"
10.242.42.191 - - [20/Jun/2023:12:49:38 -0400] "GET /api/v2/project_updates/189/ HTTP/1.1" 499 0 "-" "Python-urllib/3.11" "-"
10.242.42.191 - - [20/Jun/2023:12:49:38 -0400] "GET /api/v2/project_updates/159/ HTTP/1.1" 499 0 "-" "Python-urllib/3.11" "-"
10.242.42.191 - - [20/Jun/2023:12:49:38 -0400] "GET /api/v2/project_updates/177/ HTTP/1.1" 499 0 "-" "Python-urllib/3.11" "-"

It was found by increasing the timeout as specified at 
https://github.com/ansible/ansible/blob/1f9836fe9a417dd4dec9fcc7c7d45a11ee831d1f/lib/ansible/module_utils/urls.py#L1309

solved this issue. As most deployments will not have this issue, a default of 10 is fine, however it should be customizable, so put it in the general auth config as a place to add this option to every module as need be.

This has been discussed here:
https://github.com/redhat-cop/controller_configuration/issues/497

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - Collection

##### AWX VERSION

```
22.3.0
```


awx_collection/tests/integration/targets/project/tasks/main.yml
```diff
@@ -53,6 +53,23 @@
...
 
+    - name: Create a git project and wait with short request timeout.
+      project:
+        name: "{{ project_name1 }}"
+        organization: Default
+        scm_type: git
+        scm_url: https://github.com/ansible/test-playbooks
+        wait: true
+        state: exists
+        request_timeout: .1
```
AlanCoding: I'd like to think that at some point this will become flaky because we can process this in under 100ms, if it can support even shorter timeouts I would prefer that.

sean-m-sullivan: Updated to .001 and tested that to make sure it kept the error message, changed from auth token to handshake timeout.

AlanCoding:
Looks good to me, @john-westcott-iv do you have any concerns? For instance, is it a problem in your vision that we have this as a common module parameter but don't (I think) process it from an env var or config file?

awx_collection/plugins/module_utils/controller_api.py
```diff
@@ -61,12 +62,14 @@ class ControllerModule(AnsibleModule):
         'username': 'controller_username',
         'password': 'controller_password',
         'verify_ssl': 'validate_certs',
+        'req_timeout': 'request_timeout',
```
AlanCoding: oh, we do process this from env vars and config file..

sean-m-sullivan: this I was not sure about, willing to change, but everything refers to req_timeout,

AlanCoding (approved):

awx_collection/plugins/doc_fragments/auth_plugin.py
```diff
@@ -68,6 +68,13 @@ class ModuleDocFragment(object):
         why: Collection name change
         alternatives: 'CONTROLLER_VERIFY_SSL'
     aliases: [ validate_certs ]
+  request_timeout:
+    description:
+    - Specify the timeout Ansible should use in requests to the controller host.
+    - Defaults to 10s, but this is handled by the shared module_utils code
```
john-westcott-iv: I don't think we need to specify that this is controlled by the module_utils code. All of these settings are.

Also.... we have a situation here (I think). This auth file was shared by both `ControllerAWXKitModule` and `ControllerAPIModule` but I don't think this new setting is respected by `ControllerAWXKitModule`. So we should consider that.... do we want to split the doc fragment into two, do we want to see if awxkit can take a timeout and respect it, something else?

AlanCoding: I don't see awxkit supporting this, maybe we can just add a line qualifying that this won't work for certain modules.

sean-m-sullivan: Updated  description lines
```
    description:
    - Specify the timeout Ansible should use in requests to the controller host.
    - Defaults to 10 seconds
    - This will not work with the export or import modules.
```

john-westcott-iv (approved):

sean-m-sullivan:
Doing further review, looks like this breaks lookup plugin, so going to look into that.

awx_collection/plugins/module_utils/controller_api.py
sean-m-sullivan: This turned out to be an issue just changed to request_timout everywhere.

sean-m-sullivan:
Found issue with the short parms with the api lookup, changed the variable to be request_timemout everywhere instead of a shortname in the api utils.

jay-steurer (approved):

PR #14160, opened 2023-06-22
Tooling for running collection tests locally ad hoc

AlanCoding:
##### SUMMARY
I have maintained a silly gist with instructions for running the collection integration tests here

https://gist.github.com/AlanCoding/2adb1590489912233117aa572b44b85a

This PR is attempting to fully finalize that to hand over to other developers to run generally.

I would like to eventually run in CI as a part of the dev environment check (because only that check actually has a server running). That requires running _all_ the integration tests, but if we do that, I believe we should cut out some slow actions.

This is stripped down a little bit from what I had before, because I really don't want people to run all the tests. This should be focused on the case of running just 1 test module, and you should wrap that in bash for loops or whatever you want to do more.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

### DIFF
--- a/awx_collection/tools/README.md
+++ b/awx_collection/tools/README.md
@@ -1,0 +1,62 @@
+## Collection tools
+
+Tools used for building, maintaining, and testing the collection.
+
+### Template Galaxy
+
+The `template_galaxy.yml` playbook ran as a pre-requisite for building the collection.
+
+```
+make awx_collection_build
+```
+
+### Generate
+
+This will template resource modules (like `group`, for groups in inventory) from a boilerplate template.
+It is intended as a tool for writing new modules or enforcing consistency.
+
+### Integration Testing
+
+These instructions assume you have ansible-core and the collection installed.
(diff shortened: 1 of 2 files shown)


awx_collection/tools/README.md
```diff
@@ -0,0 +1,61 @@
...
+### Generate
+
+This will template resource modules (like `group`, for groups in inventory) from a boilerplate template.
+It is intended as a tool for writing new modules or enforcing consistency.
+
+### Integration Testing
+
+These instructions assume you have ansible-core and the collection installed.
+To install the collection in-place (to pick up any local changes to source)
+the `make symlink_collection` will put in place symlinks.
```
gamuniz: this reads funny to me. will put in place symlinks?

AlanCoding: I'll say "will symlink the folder"

gamuniz (approved):

jessicamack (approved):

PR #14161, opened 2023-06-23
Add hashivault option as docker-compose optional container

AlanCoding:
##### SUMMARY
Replaces https://github.com/ansible/awx/pull/13180

EDIT: testing has now been completed and steps have been outlined in the README so other people can follow it as well.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API


jjwatt (approved):
Approved

awx/main/credential_plugins/hashivault.py
```diff
@@ -265,6 +265,8 @@ def kv_backend(**kwargs):
 
     if secret_key:
         try:
+            if secret_key != 'data' and 'data' in json['data']:
+                return json['data']['data'][secret_key]
```
AlanCoding, line 269: This seems to be an actual bug fix which was necessary. Example of the response data we got by adding in logging statements while debugging:

```python
{'request_id': '0577c382-e10a-9452-01e3-78d512ad67fe', 'lease_id': '', 'renewable': False, 'lease_duration': 0, 'data': {'data': {'alan': '1234asdf1234fdsa431fdsa'}, 'metadata': {'created_time': '2023-06-23T19:01:43.020126505Z', 'custom_metadata': None, 'deletion_time': '', 'destroyed': False, 'version': 1}}, 'wrap_info': None, 'warnings': None, 'auth': None}
```

You can clearly see - it is nested twice under "data". I don't see any hazard with using that entry, for the cases where the entry is present, and this debugging clearly proves that it is necessary.

tools/docker-compose/README.md
```diff
@@ -303,7 +303,7 @@ To bring up a 1 node AWX + minikube that is accessible from AWX run the followin
 Start minikube
 
 ```bash
-(host)$minikube start --cpus=4  --memory=8g --addons=ingress`
+(host)$minikube start --cpus=4  --memory=8g --addons=ingress
```
AlanCoding, line 306: this was messing up the formatting of the rest of the file.

awx/main/credential_plugins/hashivault.py
```diff
@@ -265,6 +265,8 @@ def kv_backend(**kwargs):
 
     if secret_key:
         try:
+            if secret_key != 'data' and 'data' in json['data']:
```
AlanCoding: now I'm thinking that this is not conservative enough. I think we should add to this `and secret_key not in json['data']`. Because we could be looking up a key, and "data" is another irrelevant key.

AlanCoding:
I wonder if the reason this is the only one hitting that swagger issue is that it's the only PR active that's rebuilding the image.

awx/main/credential_plugins/hashivault.py
john-westcott-iv: Do we want to split this fix out into its own PR?

awx/main/credential_plugins/hashivault.py
jjwatt: I think I ended up doing a really liberal one in javascript when I was dealing with Vault because sometimes some frameworks were unwrapping and some weren't, so I had like a recursive-ish deal that unwrapped until there was only one data with a leaf under it.

AlanCoding:
I have https://github.com/ansible/awx/pull/14165 to fix the checks, because it's not related to this.

PR #14168, opened 2023-06-26
Enhance development sidecar containers

john-westcott-iv:
##### SUMMARY
Enhanced vault integration via:
Added persistent storage for vault
Added playbooks to auto configure vault and awx

General Improvements:
Create a new pattern for custom containers where we can do initialization
Auto-install roles needed for plumbing via the Makefile

##### ISSUE TYPE

##### COMPONENT NAME

 - Other

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```


Makefile
```diff
@@ -532,6 +532,9 @@ docker-compose-sources: .git/hooks/pre-commit
             $(EXTRA_SOURCES_ANSIBLE_OPTS)
 
 docker-compose: awx/projects docker-compose-sources
+	ansible-galaxy install -r tools/docker-compose/ansible/requirements.yml;
+	ansible-playbook -i tools/docker-compose/inventory tools/docker-compose/ansible/initialize_containers.yml \
```
AlanCoding, line 536: The dev-env check can't find the `initialize_containers.yml` playbook, and I can't either.

Makefile
```diff
@@ -532,6 +532,9 @@ docker-compose-sources: .git/hooks/pre-commit
...
+	  -e enable_vault=$(VAULT);
```
AlanCoding, line 537: What if we had a new target, like `make docker-compose-plumb` which runs plumbing playbooks for all of the env vars which are set up?

Makefile
AlanCoding: oh, ha, it's `initalize_containers.yml` elsewhere (sp)

Makefile
john-westcott-iv: Yea, I've been trying to think about how to do the plumbing. A make target wouldn't be a bad idea.

Makefile
john-westcott-iv: Fixed.

tools/docker-compose/README.md
```diff
@@ -505,32 +505,45 @@ Run a HashiVault container alongside of AWX.
...
+This includes the unseal keys and a root token.
 
-You can find the generated token at `tools/docker-compose/_sources/secrets/vault_password.yml`,
-this is a root token, and it should not need a corresponding username.
-Note that the token will be different on each restart, as it is re-generated by the playbook,
-and the container does not use a persistent volume.
+You will need to unseal the HashiVault each time the container is started.
+The easiest way to do that is to run:
+```bash
+ansible-playbook tools/docker-compose/ansible/unseal_vault.yml
```
AlanCoding, line 514: It looks like unsealing is done in the initialize playbook now? I'm not sure if this is necessary then.

Makefile
jjwatt: If ya'll go this direction, please put it in a script if it's more than one line. You can export all the 'make' vars to the script if that's why you keep using make as a harness. Basically, extract any complex command blocks to a specialized script that can be run outside of make, too.

tools/docker-compose/README.md
john-westcott-iv: We have to do it in the initialize playbook so that we can create the default engine and secret.

AlanCoding (approved):
Strategically, I really want to see this merged in, and I am willing to stick my neck out for it a bit.

If you wanted a slight bit of additional feedback - you could condition the galaxy install and initialize playbook on the VAULT env var being set. This would obviously be expected to grow, as more sidecars start using this, but I don't see that as a problem. That might help soften any disruption, but I'd also be okay to move forward as-is.

Issue #12815, opened 2022-09-02
Direct links to AWX do not reroute the user after authentication


anxstj:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

If users have a link to a resource and they are not logged in, then they will be redirected to the login page and after successful login redirected to the dashboard instead of the originally requested resource.

### AWX version

21.5.0

### Select the relevant components

- [X] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

docker development environment

### Modifications

no

### Ansible version

2.12.2

### Operating system

Debian 11

### Web browser

Firefox

### Steps to reproduce

- Take a link
- Rerouted to Login Page. Login
- Taken to the dashboard instead of the original link 

### Expected results

Rerouted to the original link after login.

### Actual results

Rerouted to the dashboard.

PR #14170, opened 2023-06-27
Fix rerouting issue after user authentication (#12815)

Sasa993:
##### SUMMARY

This fixes an issue with wrong user redirection after login. 

##### ISSUE TYPE

##### COMPONENT NAME

 - UI

##### AWX VERSION

```
22.3.1.dev76+g1078ec709c
```

##### ADDITIONAL INFORMATION

Before the change, user would be redirected to the dashboard instead of the original link.
`SESSION_USER_ID` would always be `null` after authentication, causing the user to be redirected to '/home/'.

```

```

### DIFF
--- a/awx/ui/src/screens/Login/Login.js
+++ b/awx/ui/src/screens/Login/Login.js
@@ -129,6 +129,7 @@ function AWXLogin({ alt, isAuthenticated }) {
 
   const setLocalStorageAndRedirect = useCallback(() => {
     if (userId && !hasVerifiedUser.current) {
+      window.localStorage.setItem(SESSION_USER_ID, JSON.stringify(userId));
       const verifyIsNewUser = () => {
         const previousUserId = JSON.parse(
           window.localStorage.getItem(SESSION_USER_ID)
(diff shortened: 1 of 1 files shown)


djyasin:
@mabashian we would like to get some UI eyes on this PR.

marshmalien:
Hello!

Thanks very much for your submission to AWX. It means a lot to us that you have taken time to contribute.

At this time we do not want to merge this PR. Our reasons for this are: 
- When the user is logged out due to a session timeout, we return the user to the last page upon log in so that they may resume their last task.
- When the user chooses to log out (click the logout button), we return them to the home screen upon log in. Users are probably not in the middle of important work when they choose to log out.
- I noticed that this code doesn't handle the edge case of multiple users. If user-A logs out, it may be jarring for user-B to login and be redirected to user-A's last session location. 

Please know that we are always up for discussion but this project is very active. Because of this, we're unlikely to see comments made on closed PRs, and we lock them after some time. If you or anyone else has any further questions, please let us know by using any of the communication methods listed in the page below:

https://github.com/ansible/awx/#get-involved

In the future, sometimes starting a discussion on the development list prior to implementing a feature can make getting things included a little easier, but it is not always necessary.

Thank you once again for this and your interest in AWX!

PR #14179, opened 2023-06-28
[POC] Developer Documentation Revamp using mkdocs

relrod:
##### SUMMARY

I started on this a few months ago, and dusted it off and did some more work on it tonight.

Just a proof of concept for what would dev docs could look like using mkdocs.

I moved the docs files around to roughly organize them, but I am not happy with the organization yet.

mkdocs-material (the theme I am using here) has a "tags" feature which helps to group documents together, and I think this is more useful than the directory structure. But the directory structure determines the hierarchy of the navbar on the left of the site.

#### Front page :page_facing_up: 

image

#### Tags index :label: 

image

#### Search :mag: 

image

#### Built-in Mermaid graphs(!) :chart_with_upwards_trend: 

image

##### ISSUE TYPE

##### COMPONENT NAME

 - Docs


.github/workflows/dev_docs.yml
```diff
@@ -0,0 +1,39 @@
...
+          if [[ $errors -ne 0 ]]; then
+            echo "Encountered $errors errors"
+            exit 1
+          fi
+
+      - name: Build the site
+        run: |
+          mkdocs build --clean
+
+      # TODO: Deploy the site somewhere (gh pages?)
```
AlanCoding, line 39: Reading up on the limitations of this... github pages seem to want an `index.html` at the root level. Since this would be an artifact of the build process, we probably don't want to keep that in source ourselves. Then the URL would be somewhat limited, as we could have maybe `http://ansible.github.io/awx`, which would maybe be okay, maybe not.

Given what other projects are doing...
 - https://ansible.readthedocs.io/projects/builder/en/stable/
 - https://ansible.readthedocs.io/projects/runner/en/stable/

I would just assume we go with readthedocs.io

PR #14182, opened 2023-06-28
Fixed bug where a weekly rrule string without a BYDAY would result in the UI throwing a TypeError

mabashian:
##### SUMMARY
When creating a schedule with an rrule like: `DTSTART:20191219T130551Z RRULE:FREQ=WEEKLY;INTERVAL=1;COUNT=1` the UI would crash and show a TypeError.  This patch fixes that.  The problem was that `BYDAY` is missing in the rrule string.  This seems to be a valid rrule string although confusing to me.

##### ISSUE TYPE

##### COMPONENT NAME
 - UI


vidyanambiar (approved):
lgtm!

PR #14184, opened 2023-06-28
Automatically set up auth for integration tests, get rid of tower-cli reference

AlanCoding:
##### SUMMARY
This is a minor followup, as it was thought to be possible before, but I just never dived far enough in before now.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - Collection

### DIFF
--- a/awx_collection/tools/integration_testing.yml
+++ b/awx_collection/tools/integration_testing.yml
@@ -1,5 +1,28 @@
 ---
-- hosts: localhost
+# NOTE: Using a token makes tests run significantly faster
+- name: Generate an OAuth2 token for the test run
+  hosts: localhost
+  gather_facts: false
+  connection: local
+  collections:
+    - awx.awx
+  tasks:
+    - name: Create a new token using username/password
+      awx.awx.token:
+        description: 'Token used for collection tests {{ now() }}'
+        scope: write
+        state: present
+        controller_username: admin
+        controller_password: >-
+          {{ (lookup('ansible.builtin.file',
+          '{{ playbook_dir }}/../../tools/docker-compose/_sources/secrets/admin_password.yml') | from_yaml
+          )['admin_password'] }}
+        validate_certs: false
+        controller_host: https://localhost:8043/
+
+
+- name: Run the integration tests
+  hosts: localhost
   gather_facts: false
   connection: local
   collections:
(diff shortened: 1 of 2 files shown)


awx_collection/tools/integration_testing.yml
```diff
@@ -8,6 +31,11 @@
     collection_location: "{{ playbook_dir }}/.."
     loc_tests: "{{ collection_location }}/tests/integration/targets/"
     test: ad_hoc_command,host,role
+    auth_env:
+      CONTROLLER_VERIFY_SSL: false
+      CONTROLLER_HOST: https://localhost:8043/
+      CONTROLLER_OAUTH_TOKEN: "{{ controller_token['token'] }}"
```
john-westcott-iv, line 37: Have you tested this with a lookup plugin? I've had issues with some configurations where a lookup plugin wouldn't get the environment variables when defined like this. But I'm wondering it it works correctly because oh how the tasks are included?

awx_collection/tools/integration_testing.yml
```diff
@@ -1,5 +1,28 @@
...
+  tasks:
+    - name: Create a new token using username/password
+      awx.awx.token:
+        description: 'Token used for collection tests {{ now() }}'
+        scope: write
+        state: present
+        controller_username: admin
+        controller_password: >-
+          {{ (lookup('ansible.builtin.file',
+          '{{ playbook_dir }}/../../tools/docker-compose/_sources/secrets/admin_password.yml') | from_yaml
```
john-westcott-iv, line 18: I'm a little cringe on this because I change my password with awx-manage but don't update the file which would affect me if I was running this locally. You could do something like:
```
controller_password: "{{ password_override | lookup(...) }}"
```

That would allow someone to pass in password_override to give it a password instead of the file.

awx_collection/tools/integration_testing.yml
AlanCoding: huh, I guess you're right about that.

```
TASK [Try to load invalid endpoint] **********************************************************************************
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'awx.awx.controller_api'. Error was a <class 'ansible.errors.AnsibleError'>, original message: There was an unknown error when trying to connect to https://127.0.0.1/api/v2/john/: URLError <urlopen error [Errno 111] Connection refused>. There was an unknown error when trying to connect to https://127.0.0.1/api/v2/john/: URLError <urlopen error [Errno 111] Connection refused>"}
...ignoring
```

that doesn't seem great.

AlanCoding:
Guess this doesn't work for lookup plugins.

PR #14188, opened 2023-06-30
[WIP] Overhaul of how sidecar containers work in development environment

john-westcott-iv:
##### SUMMARY

Instead of using flags like `KEYCLOAK=True make docker-compose` this PR introduces a yaml variable file where you can turn options on and off and all pertinent variables for the containers are listed in the variable file. This also is going to adopt the initialize/plumb model we introduced with the vault container. 

TODO:
- [ ] Make docker_config.yml somehow git ignored (another file perhaps), maybe leverage the inventory file for the defaults?
- [ ] Lots of testing
- [ ] Make sure in SAML provider configuration the client ID is AWX and not {{ container_reference }}
- [x] Make sure in Keycloak initialization that we don't error out on creating the realm if its already created.

##### ISSUE TYPE

##### COMPONENT NAME

 - Other

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```


djyasin:
@john-westcott-iv do we still want this PR, or can we close this one?

PR #14190, opened 2023-06-30
Allow user customization of the git environment

AlanCoding:
##### SUMMARY
Feature similar to `galaxy_task_env` but applied to the git task https://github.com/ansible/awx/pull/11865

Several issues we've had probably could have been resolved with this, but I don't have something specific to point to right now.

Pragmatically, this is done so that we can fix an integration test.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/playbooks/project_update.yml
+++ b/awx/playbooks/project_update.yml
@@ -15,6 +15,7 @@
 # scm_track_submodules: true/false
 # roles_enabled: Value of the global setting to enable roles downloading
 # collections_enabled: Value of the global setting to enable collections downloading
+# git_task_env: environment variables to use with the git module
 # galaxy_task_env: environment variables to use specifically for ansible-galaxy commands
 # awx_version: Current running version of the awx or tower as a string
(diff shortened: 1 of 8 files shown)


awx/playbooks/project_update.yml
```diff
@@ -24,6 +25,8 @@
   gather_facts: false
   connection: local
   name: Update source tree if necessary
+  vars:
+    git_task_env:  # configure in settings
```
relrod, line 29: Is this necessary? Can't we just give a default below like we do for `scm_accept_hostkey` etc?

```yaml
environment: "{{ git_task_env | default('') }}"
```

relrod:
Adding @AlexSCorey for UI review

awx/playbooks/project_update.yml
AlanCoding: If we did it that way, I believe the default would be `{}`, not `''`, and we could... but I like having it at the play level to declare where it comes from.

Issue #14178, opened 2023-06-27
[hop node] migration peers_from_control_nodes only in k8s


fosterseth:
1. rename migration function `set_peers_from_control_nodes_true` to `automatically_peer_from_control_plane`
2. import settings and only run function if settings.IS_K8S is true
3. set listener_port for control nodes to None

PR #14196, opened 2023-07-03
[hop node] migration peers_from_control_nodes only in k8s

djyasin:
##### SUMMARY

Related issue #14178 

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
22.3.1
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/migrations/0186_hop_nodes.py
+++ b/awx/main/migrations/0186_hop_nodes.py
@@ -2,11 +2,11 @@
 
 import django.core.validators
 from django.db import migrations, models
-from django.test.utils import override_settings
+from django.conf import settings
 
 
 def automatically_peer_from_control_plane(apps, schema_editor):
-    with override_settings(IS_K8S=True):
+    if settings.IS_K8S:
         Instance = apps.get_model('main', 'Instance')
         Instance.objects.filter(node_type='execution').update(peers_from_control_nodes=True)
         Instance.objects.filter(node_type='control').update(listener_port=None)
--- a/awx/main/migrations/0186_hop_nodes.py
+++ b/awx/main/migrations/0186_hop_nodes.py
@@ -4,7 +4,7 @@ import django.core.validators
 from django.db import migrations, models
 
 
-def set_peers_from_control_nodes_true(apps, schema_editor):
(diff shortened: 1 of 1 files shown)


djyasin:
- [x] rename migration function set_peers_from_control_nodes_true to automatically_peer_from_control_plane
- [x] import settings and only run function if settings.IS_K8S is true
- [x] set listener_port for control nodes to None

awx/main/migrations/0186_hop_nodes.py
```diff
@@ -2,11 +2,14 @@
...
 import django.core.validators
 from django.db import migrations, models
+from django.test.utils import override_settings
 
 
-def set_peers_from_control_nodes_true(apps, schema_editor):
-    Instance = apps.get_model('main', 'Instance')
-    Instance.objects.filter(node_type='execution').update(peers_from_control_nodes=True)
+def automatically_peer_from_control_plane(apps, schema_editor):
+    with override_settings(IS_K8S=True):
```
AlanCoding: Shouldn't this just be `if settings.IS_K8S`? the override_settings is a test utility and shouldn't be used in non-test code. It's function is to override the existing value, which isn't what you want, and the models here shouldn't be running any custom logic that it could affect anyway.

djyasin: Thank you so much, Alan! I was actually working on testing this piece today. I will revise accordingly. Thank you again for your feedback!

AlanCoding (approved):

djyasin:
Testing process:

- Git reset to fb8fadc7f9 Add new ANSIBLE_COLLECTIONS_PATH
- Created new cluster
- Added two execution
- Stopped cluster 
- Checked out my branch
- Started cluster
- Execed into task pod, ran migrations for file 0816
- Confirmed execution nodes had ‘peers_from_control_nodes’ set to true
- Confirmed control node had listener_port set to None

thedoubl3j (approved):
lgtm, +1 for testing comment, help set context for me really quickly

fosterseth (approved):

PR #14205, opened 2023-07-05
Added support to collection for named urls

john-westcott-iv:
##### SUMMARY
With this change awx.awx colleciton modules can used named URLs. For example:
```
---
- name: Test a named URL
  hosts: localhost
  connection: local
  gather_facts: False
  tasks:
    - name: Create a job template
      awx.awx.job_template:
        name: named_url_test
        job_type: "run"
        inventory: "Demo Inventory++Default"
        project: "Demo Project++Default"
        playbook: "hello_world.yml"
        credentials:
          - "Demo Credential++Machine+ssh++"
        state: "present"
```

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - Collection

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx_collection/README.md
+++ b/awx_collection/README.md
@@ -68,6 +68,7 @@ Notable releases of the `awx.awx` collection:
  - 7.0.0 is intended to be identical to the content prior to the migration, aside from changes necessary to function as a collection.
  - 11.0.0 has no non-deprecated modules that depend on the deprecated `tower-cli` [PyPI](https://pypi.org/project/ansible-tower-cli/).
  - 19.2.1 large renaming purged "tower" names (like options and module names), adding redirects for old names
+ - X.X.X added support of named URLs to all modules. Anywhere that previously accepted name or id can also support named URLs
  - 0.0.1-devel is the version you should see if installing from source, which is intended for development and expected to be unstable.
 
 The following notes are changes that may require changes to playbooks:
--- a/awx_collection/plugins/module_utils/controller_api.py
+++ b/awx_collection/plugins/module_utils/controller_api.py
@@ -10,7 +10,7 @@ from ansible.module_utils.six import raise_from, string_types
 from ansible.module_utils.six.moves import StringIO
 from ansible.module_utils.six.moves.urllib.error import HTTPError
 from ansible.module_utils.six.moves.http_cookiejar import CookieJar
-from ansible.module_utils.six.moves.urllib.parse import urlparse, urlencode
+from ansible.module_utils.six.moves.urllib.parse import urlparse, urlencode, quote
 from ansible.module_utils.six.moves.configparser import ConfigParser, NoOptionError
 from socket import getaddrinfo, IPPROTO_TCP
 import time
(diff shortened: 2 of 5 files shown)


AlanCoding:
Organizations are vanilla name-unique. So you would have to do `organization: Default++`? I don't know if this is useful or intended...

sean-m-sullivan:
Vanilla names should still work, What I've been trying to do is make it so it takes the output of export out of the box, and using that as the metric, if we could an output for export to put out the named urls... well that would be /chefskiss

ericzolf:
It might make sense to make it clear that `get_one` remains the entry point by prefixing the two new functions with underscore? i.e. `_get_by_name` and `_get_by_named_url`

john-westcott-iv:
I found a bug playing around with this and will be providing an update to the code.

john-westcott-iv:
@AlanCoding items with just a single named reference (i.e. Orgs) would skip the named reference task in this current implementation. Since a named reference is a pretty simple git we could remove the check on `++` and always try the GET for a named reference first and then fall back to the lookup method. This should be much easier in the new implementation of the code. Let me know if you think its worth it, I'm currently debating if it would be "faster" since we already say to use "Name" but that would only apply to named lookups where they only have one part for the name so maybe not?

AlanCoding:
How should we document features like this? We have plugin-level docs with examples, but because this is general that's not effective. For stuff like authentication, there are parameters that correspond to features so those options can be documented on their own.

To me that leaves the README. Maybe we need a general feature overview that covers stuff like this. Otherwise, how do users find it?

john-westcott-iv:
@AlanCoding I did open an issue for updating all of the docs to say: "name, ID or named URL" #14208. Which would help but yea, we could also do something in the root readme.

john-westcott-iv:
I put a place holder in the readme which will need a last minute replacement once we know what version this will be released with.

awx_collection/README.md
```diff
@@ -68,6 +68,7 @@ Notable releases of the `awx.awx` collection:
  - 7.0.0 is intended to be identical to the content prior to the migration, aside from changes necessary to function as a collection.
  - 11.0.0 has no non-deprecated modules that depend on the deprecated `tower-cli` [PyPI](https://pypi.org/project/ansible-tower-cli/).
  - 19.2.1 large renaming purged "tower" names (like options and module names), adding redirects for old names
+ - X.X.X added support of named URLs to all modules. Anywhere that previously accepted name or id can also support named URLs
```
AlanCoding, line 71: so it's either going to be 22.5.1 or 22.6.0 or 23.0.0?

awx_collection/plugins/module_utils/controller_api.py
```diff
@@ -421,7 +443,6 @@ def get_one(self, endpoint, name_or_id=None, allow_none=True, check_exists=False
             self.fail_wanted_one(response, endpoint, new_kwargs.get('data'))
 
         if check_exists:
-            name_field = self.get_name_field_from_endpoint(endpoint)
```
AlanCoding, line 424: so surprised no linter could pick this up

AlanCoding (approved):
This seems complete, I just want to look for some feedback that people see this and use it and it doesn't mess them up.

jay-steurer (approved):

Issue #13261, opened 2022-12-01
Update python-tss-sdk dependency


john-westcott-iv:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

We pin python-tss-sdk to 1.0.0 but the latest (as of opening this issue) is 1.2.0. This library is for Thytoic secret server. In the latest 1.2.0 Thytoic was renamed to Delinea and the namespace of the package was changed but the functionality seems to be intact. We need to update references like `from thycotic.secrets.server ...` to `from delinea.secrets.server ...` and test functionality.

### AWX version

21.9.0

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

N/A

### Modifications

no

### Ansible version

N/A

### Operating system

N/A

### Steps to reproduce

Upgrade python-tss-sdk to latest version and test.

### Expected results

Tests should pass.

### Actual results

Import errors.

dimatha:
I'm currently running an PoC with Delinea and it seems that I can't get it going. The error is: "SecretServerClientError", using cloud server and no domain.  I also had to set the HTTP proxy to get it going as we don't have direct internet access. Looking through the code, it looks similar - https://github.com/DelineaXPM/python-tss-sdk. Using the API directly, I can get the secret. Not sure how to troubleshoot this. Do you guys have any progress ?

dimatha:
I did some investigation, and it looks like the password is sent empty, where the value of the password is assigned to the 'domain' attribute.

dimatha:
Ok, this issue has been fixed here: https://github.com/ansible/awx/commit/d357c1162f3d52389ac9f466cdfe4b14a777f73c

thedoubl3j:
@delineaKrehl you mind taking a look at this?

PR #14207, opened 2023-07-06
Update python-tss-sdk dependency

delinea-sagar:
**SUMMARY**

Fixed issue #13261 . Updated python-tss-sdk dependency.

**ISSUE TYPE**

**COMPONENT NAME**

awx/main/credential_plugins/tss.py


delineaKrehl (approved):

delineaKrehl (approved):

AlanCoding:
Question that came up elsewhere, if the `thycotic` import will no longer be available, don't we need to update this as well?

https://github.com/ansible/awx/blob/1ad6aec97c699626cd7df8bfd0d7c9a8bf033de4/awx/main/credential_plugins/dsv.py#L5

The bigger problem, I think, is that we need proper testing for this. That's something I've spent some time on and am still trying to get prioritized.

tylerezimmerman:
@AlanCoding i think that is outside the scope of this PR and can be addressed in a future effort.

AlanCoding:
I just wanted to be sure we're not outright breaking something. So testing:

```
python -c "from thycotic.secrets.vault import SecretsVault" # works

pip install python-tss-sdk==1.2.1

python -c "from delinea.secrets.server import DomainPasswordGrantAuthorizer, PasswordGrantAuthorizer, SecretServer, ServerSecret" # works
python -c "from thycotic.secrets.vault import SecretsVault" # works
```

Does this supply both the `delinea` and `thycotic` imports? No.

The fact that upgrading does not remove the `thycotic` import path appears to be because of imperfect cleanup from the `pip` upgrade. If I create a new environment, `python -m venv env2`, activate it, and install the 1.2.1 version, `thycotic` imports give a `ModuleNotFoundError`.

Also,  `python -c "from delinea.secrets.vault import SecretsVault"` does not work, meaning that the maintainers don't seem to be maintaining this vault section of the library anymore.

If we take the action here, we'd break any use of the `dsv.py` plugin, which raise the question of why we should continue shipping it at all if it's broken as distributed?

delinea-sagar:
@AlanCoding 
**"from delinea.secrets.vault import SecretsVault"** is coming from **python-dsv-sdk** not from **python-tss-sdk**. So, I think that is not related to this PR. For dsv.py credential plugin we need to update **python-dsv-sdk** dependency and that will be in different PR.
CC: @tylerezimmerman @delineaKrehl

AlanCoding:
^ thanks for the reply. These libraries are sometimes making major import changes in minor versions, and in testing it I have been too incautious about the exact versions I'm using. First, confirming versions

```
# pip show python-dsv-sdk python-tss-sdk
Name: python-dsv-sdk
Version: 1.0.1
Summary: The Thycotic DevOps Secret Vault Python SDK
Home-page: 
Author: Adam Migus
Author-email: adam@migus.org
License: 
Location: /var/lib/awx/venv/awx/lib/python3.9/site-packages
Requires: dataclasses, requests
Required-by: 
---
Name: python-tss-sdk
Version: 1.2.1
Summary: The Delinea Secret Server Python SDK
Home-page: 
Author: Delinea Integrations
Author-email: GitHub@delinea.com
License: 
Location: /var/lib/awx/venv/awx/lib/python3.9/site-packages
Requires: requests
Required-by: 
```

This is what we will get after this PR.

Now check imports for both plugins.

```
python -c "from thycotic.secrets.vault import SecretsVault"
python -c "from delinea.secrets.server import DomainPasswordGrantAuthorizer, PasswordGrantAuthorizer, SecretServer, ServerSecret"
```

That's working, so I believe I can give a :+1: to this.

It is true (and perhaps what you were alluding to) that upgrading `python-dsv-sdk` _also_ affect the install location. So with this PR, we get the dsv library installing to the thycotic folder and the tss folder to the delinea folders. Not at all obvious, but this incremental change appears fine.

Also, for future planning (to make sure we don't upgrade ourselves into a corner), I tried `pip install --upgrade python-dsv-sdk` and then `python -c "from delinea.secrets.vault import SecretsVault"` worked, so I think we should be able to upgrade that plugin near-term as well.

awx/main/credential_plugins/tss.py
```diff
@@ -1,7 +1,16 @@
...
+try:
+    from delinea.secrets.server import DomainPasswordGrantAuthorizer, PasswordGrantAuthorizer, SecretServer, ServerSecret
+except ImportError:
+    try:
+        from thycotic.secrets.server import DomainPasswordGrantAuthorizer, PasswordGrantAuthorizer, SecretServer, ServerSecret
+    except ImportError:
+        SecretServer = None
+        ServerSecret = None
+        PasswordGrantAuthorizer = None
+        DomainPasswordGrantAuthorizer = None
```
AlanCoding: I know this is a common pattern in Ansible-core, but it just makes things harder to debug if the imports do go wrong, because instead of troubleshooting an import error, people are troubleshooting why a class is `None` for some reason.

However, I see some wisdom in the fallback to `thycotic` to be more flexible for someone who needed to make changes to their site-packages.

So I would prefer to get rid of these `= None` lines, and also add a brief comment explaining the libary's name change.

delinea-sagar: @AlanCoding 
We can set flag **HAS_TSS_SDK = False** if tss SDK is not found while importing instead of assigning None. And if flag value is False then we will raise AnsibleError at the start of the function as below.

if not HAS_TSS_SDK:
        raise AnsibleError("python-tss-sdk must be installed to use this plugin")

what do you think about this?

CC: @tylerezimmerman

AlanCoding: I know that's the common Ansible-core pattern, but if you look at the other plugins in this folder none of them do that, I don't see what value it offers, it's hard to test, and it is easy to get it wrong.

delinea-sagar: @AlanCoding 
You want only **"from delinea.secrets.server import DomainPasswordGrantAuthorizer, PasswordGrantAuthorizer, SecretServer, ServerSecret"** there instead of whole try-catch block. right?

AlanCoding: I'm fine to keep the imports as:

```python
try:
    from delinea.secrets.server import DomainPasswordGrantAuthorizer, PasswordGrantAuthorizer, SecretServer, ServerSecret
except ImportError:
    from thycotic.secrets.server import DomainPasswordGrantAuthorizer, PasswordGrantAuthorizer, SecretServer, ServerSecret
```

I think there are valid reasons for that. I just don't want to get into this rabbit hole of potentially handling `None` values from the 2nd except block.

delinea-sagar: @AlanCoding 
I have modified the code.

AlanCoding (approved):

AlanCoding:
As an administrative thing, could you do the `git commit --signoff` thing mentioned here?

https://github.com/ansible/awx/blob/87bb70bcc0754908682de3297d23742089a6a47f/DCO_1_1.md?plain=1#L4

and we have a squash and merge policy, I can do that myself, but if you do it I'll put in the commit as-is.

delinea-sagar:
> As an administrative thing, could you do the `git commit --signoff` thing mentioned here?
> 
> https://github.com/ansible/awx/blob/87bb70bcc0754908682de3297d23742089a6a47f/DCO_1_1.md?plain=1#L4
> 
> and we have a squash and merge policy, I can do that myself, but if you do it I'll put in the commit as-is.

@AlanCoding Do I need to add signoff for every commit? I have already added for last commit.

awx/main/credential_plugins/tss.py
```diff
@@ -1,7 +1,10 @@
 from .plugin import CredentialPlugin
 from django.utils.translation import gettext_lazy as _
 
-from thycotic.secrets.server import DomainPasswordGrantAuthorizer, PasswordGrantAuthorizer, SecretServer, ServerSecret
+try:
+    from delinea.secrets.server import DomainPasswordGrantAuthorizer, PasswordGrantAuthorizer, SecretServer, ServerSecret
+except ImportError:
+    from thycotic.secrets.server import DomainPasswordGrantAuthorizer, PasswordGrantAuthorizer, SecretServer, ServerSecret
```
andrii-zakurenyi, line 7: Hi @delinea-sagar 

I know that this was already merged, but can you please explain why did you add this `ImportError` check?

There is a PR which tries to follow the same pattern, but I wonder if it is actually required.

Related PR -> #14362

delinea-sagar: @andrii-zakurenyi this is a common pattern that we follow

PR #14210, opened 2023-07-06
Migrate from django-redis to Django's built-in Redis caching support

relrod:
##### SUMMARY
This attempts to drop a dependency on django-redis and switch to the Django built in redis backend.

##### ISSUE TYPE

##### COMPONENT NAME

 - API


AlanCoding:
Trying to document some due diligence:

https://github.com/jazzband/django-redis/issues/585

> if you need a compressor, serializer, connection pool or sentinel support you can not do that with django native support

I think we're good!

> Furthermore, if you need more control over how the keys are stored, like expire_at and millisecond precision you could only do it with `django-redis`.

Don't think so, but less confident.

> `django-redis` can help you create distributed redis locks while `django` native support could not.

We do a lot with locks, but we don't do them with redis!

AlanCoding (approved):
Conditionally on passing integration tests, I say merge it :+1:

PR #14223, opened 2023-07-11
don't pass the 'organization' or other fields to the search of the instance group or execution environments

ivarmu:

##### SUMMARY

Fixes a bug detected here: https://github.com/redhat-cop/controller_configuration/issues/624

- If instance_groups are specified in the input schedule, the Schedule role fails to get the specified instance group
- If execution_environment is specified in the input schedule, the Schedule role fails to get the specified execution_environment

##### ISSUE TYPE

 - Bug
##### COMPONENT NAME

 - Collection

##### AWX VERSION

https://github.com/redhat-cop/controller_configuration/issues/624

##### ADDITIONAL INFORMATION

https://github.com/redhat-cop/controller_configuration/issues/624


ivarmu:
@gamuniz, @john-westcott-iv: any feedback related to this bugfix?

john-westcott-iv:
@ivarmu can you add/modify the tests in https://github.com/ansible/awx/blob/devel/awx_collection/tests/integration/targets/schedule/tasks/main.yml and/or https://github.com/ansible/awx/blob/devel/awx_collection/test/awx/test_schedule.py to catch this issue so we can test it before and after your change to make sure we don't get a regression in the future?

Also, is this "fully" broken or does it work now under some circumstances? I'm asking to find out if we need a release/doc note if the behavior is changing.

ivarmu:
@john-westcott-iv , I'm adding the tests and pushing the changes.

I think this is fully broken as the detailed fields are never returned through the schedules api.

ivarmu:
> @john-westcott-iv , I'm adding the tests and pushing the changes.
> 
> I think this is fully broken as the detailed fields are never returned through the schedules api.

The tests were already implemmented, the only parameter that was not tested at all was passing the `organization` field to the `schedule` module (see https://github.com/ansible/awx/blob/ab5cc2e69c68a1fa763d181ec15b986af78e531d/awx_collection/plugins/modules/schedule.py#L136, https://github.com/ansible/awx/blob/ab5cc2e69c68a1fa763d181ec15b986af78e531d/awx_collection/plugins/modules/schedule.py#L221 and https://github.com/ansible/awx/blob/ab5cc2e69c68a1fa763d181ec15b986af78e531d/awx_collection/plugins/modules/schedule.py#L251).

I've changed the direction of the patch, letting other parameters to be passed to the API calls if needed. I was thinking on removing the field `organization` like the `name` one is removed here: https://github.com/ansible/awx/blob/ab5cc2e69c68a1fa763d181ec15b986af78e531d/awx_collection/plugins/modules/schedule.py#L283C1-L285C34 ... This would be a clearer solution, but I'm not sure that the `organization` field could be used after these API calls (by the `execution_environment` API call, for example)... @john-westcott-iv  what do you think about that? Maybe a change ordering the tasks would allow that?

ivarmu:
Nah!, I've done it, it seems more elegant than before :-)

ivarmu:
@gamuniz, @john-westcott-iv: I know these may be holiday days... when you become ready again... any advance on this PR?

ivarmu:
@gamuniz, @john-westcott-iv: putting this bugfix in your radar again :-)

john-westcott-iv (approved):

Issue #14199, opened 2023-07-05
Adding host facts API


tanganellilore:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

New Feature

### Feature Summary

Hi team,

as discussed in Matrix, I will propose you to have an API endpoint dedicated to ALL hosts ansible_facts, where, with paginated api, we can access directly to list of hosts and related facts.

I know that facts can be very heavy but with pagination and "limit" of pagesize, we can have a single entrypoint where we can fetch all facts via API.

This can be very usefull if we want to have an integration beetween AWX and other source, like CMDB, without direct access to DB and query it.

I will propose a PR that address this new feature, feel free to chat me and we can discuss together.
Thanks

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

No direct access to paginated view of hosts facts

### Current results

Facts available only call `/api/v2/hosts/<id_host>/ansible_facts`

### Sugested feature result

Have an api endpoint like `/api/v2/host_facts/` and detailed view like `/api/v2/host_facts/<id_host>`

AlanCoding:
I understand the case you're making, and I see the need for something like this.

My gut impression, however, is that because `ansible_facts` is a direct field on the `Host` model, handling this by adding additional host _view_ is awkward. We seem to display the related ansible_facts endpoint based on read permission to the host, making it just like any other field, so..

I'd like to spitball a potential alternative - what if we allowed using a querystring parameter to add or remove certain fields from the serialized data? As you're getting hosts, imagine appending `/api/v2/hosts/?add_fields=ansible_facts` and then the returned hosts would include this dictionary in the returned data. This would put the task of field curation back on the client instead of on the API internal code, as your PR hand-picks a certain field list to return.

tanganellilore:
Thanks @AlanCoding for input, yes probably your is the best approach, but i need to understand how can add this on the host serializer part.
Tomorrow I will check it, if you have some example in other api, i can copy/inspirate it.

Thanks

AlanCoding:
[...]

PR #14229, opened 2023-07-12
Adding add_fields and remove_fields on api

tanganellilore:
##### SUMMARY
This PR is to allow us the possibility to add and/remove fields from API query, simpli adding new query parameters called `add_fields` and `remove_fields`.

This is related discussion with @AlanCoding on #14199 ,  the idea is to have a way to add (or remove) one or more fileds in a api view.
My approach is to allow adding and removing fields via query params, separated by comma for all API, not only for single endpoint. 

In case of `add_fields` we need to evaluate fields, retrieve him models and then create field based on model, so in that way we can extend current field defined in serializer, with all fields present in the model itself.

In all cases (add/remove), if fields requested not exists in the model, we simply skip it.

Some example:
- Add ansible_facts to host view
`http://localhost:8013/api/v2/hosts/?add_fields=ansible_facts`
- Remove related and summary_fields from api view
`http://localhost:8013/api/v2/hosts/?remove_fields=related,summary_fields`
- A combination of add/remove
`http://localhost:8013/api/v2/hosts/?remove_fields=related,summary_fields&add_fields=ansible_facts`

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API

##### AWX VERSION

```
N/A
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/api/filters.py
+++ b/awx/api/filters.py
@@ -133,7 +133,7 @@ class FieldLookupBackend(BaseFilterBackend):
     Filter using field lookups provided via query string parameters.
     """
 
-    RESERVED_NAMES = ('page', 'page_size', 'format', 'order', 'order_by', 'search', 'type', 'host_filter', 'count_disabled', 'no_truncate', 'limit')
+    RESERVED_NAMES = ('page', 'page_size', 'format', 'order', 'order_by', 'search', 'type', 'host_filter', 'count_disabled', 'no_truncate', 'limit', 'add_fields', 'remove_fields')
 
     SUPPORTED_LOOKUPS = (
         'exact',
--- a/awx/api/generics.py
+++ b/awx/api/generics.py
@@ -345,6 +345,12 @@ class GenericAPIView(generics.GenericAPIView, APIView):
     #   serializer_class = SerializerClass
 
     def get_serializer(self, *args, **kwargs):
+        add_fields = self.request.query_params.get('add_fields', None)
+        remove_fields = self.request.query_params.get('remove_fields', None)
+        if add_fields:
+            kwargs.update({'add_fields': add_fields.split(',')})
+        if remove_fields:
+            kwargs.update({'remove_fields': remove_fields.split(',')})
         serializer = super(GenericAPIView, self).get_serializer(*args, **kwargs)
         # Override when called from browsable API to generate raw data form;
         # update serializer "validated" data to be displayed by the raw data
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -39,6 +39,7 @@ from rest_framework import fields
 from rest_framework import serializers
 from rest_framework import validators
 from rest_framework.utils.serializer_helpers import ReturnList
+from rest_framework.utils import model_meta
 
 # Django-Polymorphic
 from polymorphic.models import PolymorphicModel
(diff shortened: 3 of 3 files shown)


gamuniz:
@tanganellilore are you able to add some testing and test scenarios around this?

tanganellilore:
I can try to do that, but before i refer to have @AlanCoding feedback and also fix all lint check that are under my scope

tanganellilore:
fixed one missing fields during my copy...sorry for that

jbradberry:
@tanganellilore I am philosophically opposed to this approach.  This PR does _way_ too much by being generalized.  The new `add_fields` and `remove_fields` are bypassing any serializer validation, and I could easily imagine this functionality bypassing access control logic.

Like @AlanCoding I'm uneasy with the idea of adding a parallel list view just for your use case.  My suggestion is to add a `show_facts` field to the host serializer and just write the logic to switch on or off showing the facts based on that.

tanganellilore:
Thanks @jbradberry for your feedback, probably I'm not so expert on this topic but why addind/removing fields that is present in the model used by serialization itself should bypass serialization validation?
I'm not adding something that is not already present on model used by serialization...moreover i skipped all argunent passed if it's not present in model itself. 
(I see in some points that serializer add something outside the model, like lookup, but this is not "allowed" in my implementation) 

For access control logic, I don't have any idea how test it, but i think that access control should be present on api path and method, not in the serializer fields itself. 
 

Can we evaluate to have an "allowed list" for remove/add fields? (just to restrict it). 
To me allow possibility to user to add/remove fields will be a great features, also to ui point of view (we can reduce heavy of data passed to client with removal of unused fields). 

Imagine to have an integration that non require summary_fields or related_fields, or, just my start use case,  integrate current view with something exclude from serializer but present in the model (due to performance issue like adding facts).

Anyway, I'm simply try to address what @AlanCoding asked, but probably

sdktr:
This seems to built a graphql approach into REST. Was graphql evaluated before in the AWX project?

fosterseth:
@tanganellilore after some discussion, I think an easier PR to get merged in is going with what @jbradberry proposed. A show_facts query for the Hosts endpoint only. If enabled, add ansible facts to the hosts

would you be able to make those changes? we can just close this PR (to keep it as a reference in case we want to resurrect the idea) and open a new one.

jbradberry (changes requested):
@tanganellilore 

Bypassing validation:

You are pulling the values of `add_fields` and `remove_fields` directly out of the request.  They could be any value at all and thus ought to be run through serializer fields with some validators first.  Additionally and more seriously, the only concept of "allowlist" here is that they are fields on the model.

Access control:

To the best of my recollection, we have various model fields that are either not allowed or only conditionally allowed to be present in the API.  This PR will bypass any logic we have elsewhere that determines this.  Additionally, I suspect that the machinery that you are making use of here will allow relational fields to be exposed directly.

I would very much prefer to see a more tightly focused PR that only conditionally exposes the facts, nothing else.

tanganellilore:
Clear @jbradberry,
I will try to submit a new PR to go in direct io suggested by you and @fosterseth.

You can close this PR just for future question about it.

PR #14230, opened 2023-07-12
Upgrade to postgres:15

john-westcott-iv:
##### SUMMARY

Upgrade from using postgres 12 to postgres 15.

Postgres 15 is not able to load a postgres 12 database directly. So we have modified the docker volume name to help distinguish the two version. However, please note that docker-compose will attempt to use the old volume if the `tools_postgres_1` container already exists. If this happens you will see an error in the logs like:
```
tools_postgres_1 | 
tools_postgres_1 | PostgreSQL Database directory appears to contain a database; Skipping initialization
tools_postgres_1 | 
tools_redis_1 | 1:M 12 Jul 2023 13:53:13.739 * monotonic clock: POSIX clock_gettime
tools_postgres_1 | 2023-07-12 13:53:13.919 UTC [1] FATAL:  database files are incompatible with server
tools_postgres_1 | 2023-07-12 13:53:13.919 UTC [1] DETAIL:  The data directory was initialized by PostgreSQL version 12, which is not compatible with this version 15.3 (Debian 15.3-1.pgdg120+1).
```

First, remove the container with `docker container rm tools_postgres_1`. If you then start the containers again you will get a fresh install of AWX.

If you want to migrate your existing docker-compose development installation do the following:

1. Start with latest current version running via `make docker-compose`.
2. Run `docker exec -it -u postgres tools_postgres_1 pg_dumpall --database awx -U awx > awx_dump.sql`
3. Stop your existing containers.
4. Remove the existing tools_postgres_1 container by running `docker container rm tools_postgres_1`.
5. Checkout the new version of the code.
6. Run `make docker-compose-sources` to upgrade the docker-compose.yml file.
7. Run `docker-compose -f tools/docker-compose/_sources/docker-compose.yml up postgres` to start up the new version of the container.
8. Run `docker cp awx_dump.sql tools_postgres_1:/tmp` to copy the backup file into the new postgres container.
9. Run `docker exec -it tools_postgres_1 /usr/bin/psql -U awx -f /tmp/awx_dump.sql` to restore your old data into the new postgres 15 database.
10. Stop the container being run from `docker-compose up` and start the full environment with `make docker-compose`.

I did simple testing with the keycloak integration (which uses the postgres container) and it works fine. There did not seems to be a compatibility matrix for keycloak and postgres but I did find one doc indicating that it worked with pg 15 (but that was likely intended for the latest version 20 and we are on 15).

##### ISSUE TYPE

 - Breaking Change 

##### COMPONENT NAME

 - Other

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/__init__.py
+++ b/awx/__init__.py
@@ -154,10 +154,11 @@ def manage():
     from django.conf import settings
     from django.core.management import execute_from_command_line
 
-    # enforce the postgres version is equal to 12. if not, then terminate program with exit code of 1
+    # enforce the postgres version is equal to 15. if not, then terminate program with exit code of 1
+    # The return of connection.pg_version is something like 12013
     if not os.getenv('SKIP_PG_VERSION_CHECK', False) and not MODE == 'development':
-        if (connection.pg_version // 10000) < 12:
-            sys.stderr.write("Postgres version 12 is required\n")
+        if (connection.pg_version // 10000) < 15:
+            sys.stderr.write("Postgres version 15 is required\n")
             sys.exit(1)
 
     if len(sys.argv) >= 2 and sys.argv[1] in ('version', '--version'):  # pragma: no cover
--- a/awx/__init__.py
+++ b/awx/__init__.py
@@ -154,11 +154,12 @@ def manage():
     from django.conf import settings
     from django.core.management import execute_from_command_line
 
-    # enforce the postgres version is equal to 15. if not, then terminate program with exit code of 1
+    # enforce the postgres version is a minimum of 12 (we need this for partitioning); if not, then terminate program with exit code of 1
+    # In the future if we require a feature of a version of postgres > 12 this should be updated to reflect that.
     # The return of connection.pg_version is something like 12013
     if not os.getenv('SKIP_PG_VERSION_CHECK', False) and not MODE == 'development':
-        if (connection.pg_version // 10000) < 15:
-            sys.stderr.write("Postgres version 15 is required\n")
+        if (connection.pg_version // 10000) < 12:
+            sys.stderr.write("At a minimum, postgres version 12 is required\n")
             sys.exit(1)
 
     if len(sys.argv) >= 2 and sys.argv[1] in ('version', '--version'):  # pragma: no cover
--- a/awx/main/analytics/collectors.py
+++ b/awx/main/analytics/collectors.py
@@ -419,7 +419,7 @@ def _events_table(since, full_path, until, tbl, where_column, project_job_create
                           resolved_action,
                           resolved_role,
                           -- '-' operator listed here:
-                          -- https://www.postgresql.org/docs/12/functions-json.html
+                          -- https://www.postgresql.org/docs/15/functions-json.html
                           -- note that operator is only supported by jsonb objects
                           -- https://www.postgresql.org/docs/current/datatype-json.html
                           (CASE WHEN event = 'playbook_on_stats' THEN {event_data} - 'artifact_data' END) as playbook_on_stats,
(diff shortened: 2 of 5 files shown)


john-westcott-iv:
Should this be considered an X version release since it can't be directly upgraded to? @shanemcd @AlanCoding et. al.

AlanCoding:
I wouldn't say that we support docker-compose upgrades to begin with.

rooftopcellist:
Agreed, and the operator deployments will work because there is migration logic built in.

rooftopcellist (dismissed):

AlanCoding (dismissed):

awx/main/analytics/collectors.py
```diff
@@ -419,7 +419,7 @@ def query(event_data):
                           resolved_action,
                           resolved_role,
                           -- '-' operator listed here:
-                          -- https://www.postgresql.org/docs/12/functions-json.html
+                          -- https://www.postgresql.org/docs/15/functions-json.html
                           -- note that operator is only supported by jsonb objects
                           -- https://www.postgresql.org/docs/current/datatype-json.html
```
TheRealHaoLiu, line 424: @john-westcott-iv can we remove this?

relrod (dismissed):

tools/ansible/roles/dockerfile/templates/Dockerfile.j2
```diff
@@ -296,7 +296,6 @@ RUN ln -sf /dev/stdout /var/log/nginx/access.log && \
 {% endif %}
 
 ENV HOME="/var/lib/awx"
-ENV PATH="/usr/pgsql-12/bin:${PATH}"
```
TheRealHaoLiu, line 303: lol... that dir doesnt seem to exist anyway good catch

relrod:
:shipit:

awx/main/analytics/collectors.py
john-westcott-iv: I believe they are comments so we /could/ remove them if we wanted but I'd say leave them.

rooftopcellist:
We are holding off merging this to coordinate with other related PR's. This may be a couple of weeks.  Sorry for the delay.  

If anyone needs to start using pg 15 before this merges, you can modify the following variables on your spec to override the postgres images:

* postgres_image
* postgres_image_version

awx/__init__.py
```diff
@@ -179,10 +179,12 @@ def manage():
...
     from django.core.management import execute_from_command_line
 
-    # enforce the postgres version is equal to 12. if not, then terminate program with exit code of 1
+    # enforce the postgres version is a minimum of 12 (we need this for partitioning); if not, then terminate program with exit code of 1
+    # In the future if we require a feature of a version of postgres > 12 this should be updated to reflect that.
+    # The return of connection.pg_version is something like 12013
     if not os.getenv('SKIP_PG_VERSION_CHECK', False) and not MODE == 'development':
         if (connection.pg_version // 10000) < 12:
-            sys.stderr.write("Postgres version 12 is required\n")
+            sys.stderr.write("At a minimum, postgres version 12 is required\n")
```
gundalow, line 162:
```suggestion
            sys.stderr.write("At a minimum, PostgreSQL version 12 is required\n")
```

rooftopcellist: @gundalow thank you for the suggestion, I incorporated it in the latest push.

rooftopcellist:
This PR is ready for final review, I rebased and addressed all remaining comments.

TheRealHaoLiu (approved):

PR #14232, opened 2023-07-12
Explicitly turn off autocomplete for API login form

rooftopcellist:
##### SUMMARY

Users have pointed out the the username and password should not be autocompleted for the login form.  That has been corrected for the UI in this PR, but not for the API for the username.  That is what this PR addresses.  

I also explicitly set autocomplete to off for the password field even though that appears to be the default.  

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/templates/rest_framework/login.html
+++ b/awx/templates/rest_framework/login.html
@@ -19,7 +19,7 @@
           <input type="text" name="username" maxlength="100"
               autocapitalize="off"
               autocorrect="off" class="form-control textinput textInput"
-              id="id_username" required autofocus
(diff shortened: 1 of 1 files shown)


mabashian (approved):
@rooftopcellist thanks for sending this through.  Adding autocomplete="off" to each input should work just fine.  You could also set it on the form element to hit all the inputs at once.  Both approaches should be valid.

chadmf:
lgtm! @rooftopcellist can we get this in soonish?

fosterseth (approved):

Issue #14238, opened 2023-07-13
Preemptively remove django-taggit relationships


AlanCoding:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

New Feature

### Feature Summary

Task, do the following:

 - remove taggit use in `awx/main/models`, mainly in the form of `taggit.managers.TaggableManager`
 - run `awx-manage makemigrations`

Any existing tags will be deleted in this process.

Why? Because with https://github.com/jazzband/django-taggit/pull/862 there is a lot of uncertainty about whether this library will play nicely with future Django versions, and whether it will do so in the time frame that we need it. Django has made some opinionated changes that requires a degree of synchronicity, in that the old pattern of `index_together` will stop working at some point, but the new pattern isn't supported by 3.Y versions. The library may increase their minimum version, or it may introduce long-term versions for 3 vs 4 Django versions. Either way, I'm not counting on that happening.

----

For the future, when (or if) the Django upgrade forces us to remove the taggit library, this will leave the option open to declare the current AWX version as a checkpoint version, and then in the next version:
 - Manually delete the taggit tables at the start of the migration command or some other way, or do this as SQL in another numbered migration
 - Also manually as a separate step, run `awx-manage remove_stale_contenttypes`, or maybe do that in a numbered migration

### Select the relevant components

- [ ] UI
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Steps to reproduce

N/A

### Current results

N/A

### Sugested feature result

N/A

PR #14241, opened 2023-07-14
Drop unused django-taggit dependency

relrod:
##### SUMMARY

This drops the django-taggit dependency and drops the relevant fields from old migrations.

##### ISSUE TYPE

##### COMPONENT NAME

 - API


relrod:
Refs #14238

But it goes further than the initial ask in that ticket, insofar as it actually removes the dep and references to it.

Note that this does not add a migration - because any migration will want these fields to exist, which of course they won't after this patch.

This might not be an issue given (per #14238):
> Manually delete the taggit tables at the start of the migration command or some other way, or do this as SQL in another numbered migration

But really I am just fiddling around at 3am and I am probably not thinking straight.

matburt:
👍🏻 This stuff goes back a *long* time... love to see it removed.

AlanCoding:
This is a draft, so it's not expected to be complete yet, but I wanted to round up the known items that are still TODO.

 - Introduce a data migration, and inside of that
   - Drop the tables used by the taggit app (I believe there are two)
   - Delete content types that reference the taggit app, `ContentType.objects.filter(app_label='taggit')`
   - [more speculative] delete the migration records that reference the app `MigrationRecorder.Migration.objects.filter(app='taggit').delete()`

This migration may be reversible in a superficial sense, as there will essentially be 2 forms of history. There is the upgrade scenario where the user starts with the app installed, and has it removed through the data migration. The other scenario is a new install and in the case the app is never installed. So reversing will not re-install the app, and we could allow reversing it under the 2nd form of history.

relrod:
@AlanCoding Sorry for the delay here, finally circling back to this. I added a migration, I think it is how we talked about doing this; it addresses the 3 points in your comment.

awx/main/migrations/0186_drop_django_taggit.py
```diff
@@ -0,0 +1,29 @@
...
+from django.db import migrations
+
+
+def delete_taggit_contenttypes(apps, schema_editor):
+    ContentType = apps.get_model('contenttypes', 'ContentType')
+    ContentType.objects.filter(app_label='taggit').delete()
+
+
+def delete_taggit_migration_records(apps, schema_editor):
+    from django.db.migrations.recorder import MigrationRecorder
```
AlanCoding: Since you've already imported the module at the top, you can get rid of this line and reference `migrations.recorder.MigrationRecorder`. Sometimes you can't do this, but I tested it with python shell and it works.

awx/main/migrations/0186_drop_django_taggit.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('main', '0185_move_JSONBlob_to_JSONField'),
+    ]
+
+    operations = [
+        migrations.RunSQL("DROP TABLE IF EXISTS taggit_tag CASCADE;"),
+        migrations.RunSQL("DROP TABLE IF EXISTS taggit_taggeditem CASCADE;"),
```
AlanCoding, line 24: :+1:  nicely done. This is as close to supported as we're ever going to get.

AlanCoding (approved):

AlanCoding:
I like this patch, and I want to get it in. So I took some time to front-run some of the tests and give independent verification. This is upstream testing, using the docker-compose environment. Nevertheless, I believe it obtains the necessary coverage.

 - :heavy_check_mark: Upgrade from `devel` to this branch with old data, confirm jobs and stuff still work
 - :heavy_check_mark: Create a fresh database and migrate to this PR. Confirmed accounts and jobs still work.

After the migrations, as expected the taggit app no longer appears in the showmigrations output.

Issue #14245, opened 2023-07-17
Dispatcher shutdown deadlock when redis is unavailable


AlanCoding:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.
- [X] I am **NOT** reporting a (potential) security vulnerability. (These should be emailed to `security@ansible.com` instead.)

### Bug Summary

We have identified a set of circumstances and the mechanic for how `awx-manage run_dispatcher` can hit an error but fail to shut down. Instead of exiting, it remains in a deadlocked state where it cannot process new work. Because the process is known by supervisord to be healthy (although it's not) this state is not recovered from without manual intervention.

### AWX version

devel

### Select the relevant components

- [ ] UI
- [ ] UI (tech preview)
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

N/A

### Modifications

no

### Ansible version

N/A

### Operating system

RHEL9

### Web browser

Chrome

### Steps to reproduce

There is a test scenario that will cause the deadlock every time, and that is:

 - shut down redis
 - wait until a control node reports a new heatbeat timestamp
 - start redis back up

Redis is shut down for < 20s through the process (although it may vary). The heatbeat may not actually run, but it does obtain the deadlocked state.

### Expected results

No deadlocks

### Actual results

Dispatcher main process hits an error and tries to exit, shown by the `Encountered unhandled error in dispatcher main loop` log:

```
Traceback (most recent call last):
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/dispatch/worker/base.py", line 203, in run
    self.run_periodic_tasks()
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/dispatch/worker/base.py", line 180, in run_periodic_tasks
    self.subsystem_metrics.pipe_execute()
```

The state of the dispatcher service is that some workers and the periodic scheduler thread exit, but several workers (1 to 3 seen in practice) are still running.

```
awx        73763  0.0  0.9 832528 146988 ?       S    Jul15   0:02  \_ /var/lib/awx/venv/awx/bin/python /usr/bin/awx-manage run_dispatcher
[...]

PR #14249, opened 2023-07-17
Make dispatcher timeout use SIGUSR1, not SIGTERM

relrod:
##### SUMMARY

Refs #14245

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/scheduler/task_manager.py
+++ b/awx/main/scheduler/task_manager.py
@@ -134,20 +134,19 @@ class TaskBase:
                         logger.debug(f"Not running {self.prefix} scheduler, another task holds lock")
                         return
                     logger.debug(f"Starting {self.prefix} Scheduler")
-                    # if sigterm due to timeout, still record metrics
+                    # if sigusr1 due to timeout, still record metrics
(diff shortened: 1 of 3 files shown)


awx/main/scheduler/task_manager.py
```diff
@@ -132,13 +135,19 @@ def schedule(self):
...
                     commit_start = time.time()
 
                 if self.prefix == "task_manager":
                     self.subsystem_metrics.set(f"{self.prefix}_commit_seconds", time.time() - commit_start)
                 self.record_aggregate_metrics()
+
+                # Reset the signal handler back to the default just in case anything
+                # else uses the same signal for other purposes
+                logger.debug('Resetting SIGUSR1 handler')
+                signal.signal(signal.SIGUSR1, original_sigusr1)
```
AlanCoding: I'd like to see this in a `finally:`, but I can't tell you the exact structure that should have.

AlanCoding:
I want to contribute this https://github.com/relrod/awx/pull/33

This will have _some_ conflicts with older versions, but the overall spirit (that metrics should never _exit the program_) is something that I'd like to apply uniformly.

AlanCoding (approved):

PR #14255, opened 2023-07-18
HostMetrics: Hard auto-cleanup

slemrmartin:
##### SUMMARY

Extends 
- scheduled task `cleanup_host_metrics` (https://github.com/ansible/awx/pull/13764) and
- command `awx-manage  cleanup_host_metrics`

for deleting of HostMetric rows which were soft-deleted 3+ years ago and wasn't reused for automation.

https://issues.redhat.com/browse/AA-1592

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - CLI
 - Scheduler

##### AWX VERSION

```
awx: 22.5.1.dev10+g4b53bd1352.d20230718
```

##### ADDITIONAL INFORMATION

It's part of [ADR0017](https://docs.google.com/document/d/1hv14yISL8GlX0Kd-xO9h9JFP9snyob5wEfUDlBimwIc/edit#heading=h.9b9owjzfgnyu) - will be backported

```

```

### DIFF
--- a/awx/main/management/commands/cleanup_host_metrics.py
+++ b/awx/main/management/commands/cleanup_host_metrics.py
@@ -1,24 +1,22 @@
 from django.core.management.base import BaseCommand
-from django.utils.translation import gettext_lazy as _
+from django.conf import settings
 from awx.main.tasks.host_metrics import HostMetricTask
 
 
 class Command(BaseCommand):
     """
-    Run soft/hard-deleting of HostMetrics
+    This command provides cleanup task for HostMetric model.
+    There are two modes, which run in following order:
+    - soft cleanup
+    - - Perform soft-deletion of all host metrics last automated 12 months ago or before.
+        This is the same as issuing a DELETE request to /api/v2/host_metrics/N/ for all host metrics that match the criteria.
+    - - updates columns delete, deleted_counter and last_deleted
+    - hard cleanup
+    - - Permanently erase from the database all host metrics last automated 36 months ago or before.
+        This operation happens after the soft deletion has finished.
     """
 
-    help = 'Run soft/hard-deleting of HostMetrics'
-
-    def add_arguments(self, parser):
-        parser.add_argument('--soft', type=int, nargs='?', default=None, const=0, help='Threshold in months for soft-deleting')
-        parser.add_argument('--hard', type=int, nargs='?', default=None, const=0, help='Threshold in months for hard-deleting')
+    help = 'Run soft and hard-deletion of HostMetrics'
 
     def handle(self, *args, **options):
-        soft_threshold = options.get('soft')
-        hard_threshold = options.get('hard')
-
-        if soft_threshold is None and hard_threshold is None:
-            return _("Specify either --soft or --hard argument")
-
-        HostMetricTask().cleanup(soft_threshold=soft_threshold, hard_threshold=hard_threshold)
+        HostMetricTask().cleanup(soft_threshold=settings.CLEANUP_HOST_METRICS_SOFT_THRESHOLD, hard_threshold=settings.CLEANUP_HOST_METRICS_HARD_THRESHOLD)
--- a/awx/main/management/commands/cleanup_host_metrics.py
+++ b/awx/main/management/commands/cleanup_host_metrics.py
@@ -1,22 +1,24 @@
-from awx.main.models import HostMetric
 from django.core.management.base import BaseCommand
-from django.conf import settings
+from django.utils.translation import gettext_lazy as _
+from awx.main.tasks.host_metrics import HostMetricTask
 
 
 class Command(BaseCommand):
     """
-    Run soft-deleting of HostMetrics
+    Run soft/hard-deleting of HostMetrics
     """
 
-    help = 'Run soft-deleting of HostMetrics'
+    help = 'Run soft/hard-deleting of HostMetrics'
 
     def add_arguments(self, parser):
-        parser.add_argument('--months-ago', type=int, dest='months-ago', action='store', help='Threshold in months for soft-deleting')
+        parser.add_argument('--soft', type=int, nargs='?', default=None, const=0, help='Threshold in months for soft-deleting')
+        parser.add_argument('--hard', type=int, nargs='?', default=None, const=0, help='Threshold in months for hard-deleting')
 
     def handle(self, *args, **options):
-        months_ago = options.get('months-ago') or None
+        soft_threshold = options.get('soft')
+        hard_threshold = options.get('hard')
 
-        if not months_ago:
-            months_ago = getattr(settings, 'CLEANUP_HOST_METRICS_SOFT_THRESHOLD', 12)
+        if soft_threshold is None and hard_threshold is None:
+            return _("Specify either --soft or --hard argument")
 
-        HostMetric.cleanup_task(months_ago)
+        HostMetricTask().cleanup(soft_threshold=soft_threshold, hard_threshold=hard_threshold)
--- a/awx/main/tasks/helpers.py
+++ b/awx/main/tasks/helpers.py
@@ -1,0 +1,8 @@
+from django.utils.timezone import now
+from rest_framework.fields import DateTimeField
+
+
+def is_run_threshold_reached(setting, threshold_seconds):
+    last_time = DateTimeField().to_internal_value(setting.value) if setting and setting.value else DateTimeField().to_internal_value('1970-01-01T00:00')
+
+    return (now() - last_time).total_seconds() > threshold_seconds
--- a/awx/main/tasks/helpers.py
+++ b/awx/main/tasks/helpers.py
@@ -3,6 +3,8 @@ from rest_framework.fields import DateTimeField
 
 
 def is_run_threshold_reached(setting, threshold_seconds):
-    last_time = DateTimeField().to_internal_value(setting.value) if setting and setting.value else DateTimeField().to_internal_value('1970-01-01T00:00')
-
-    return (now() - last_time).total_seconds() > threshold_seconds
+    last_time = DateTimeField().to_internal_value(setting) if setting else None
+    if not last_time:
+        return True
+    else:
+        return (now() - last_time).total_seconds() > threshold_seconds
--- a/awx/main/tasks/host_metrics.py
+++ b/awx/main/tasks/host_metrics.py
@@ -63,10 +63,10 @@ class HostMetricTask:
             threshold = getattr(settings, 'CLEANUP_HOST_METRICS_SOFT_THRESHOLD', 12)
 
         last_automation_before = now() - relativedelta(months=threshold)
-        logger.info(f'cleanup_host_metrics: soft-deleting records last automated before {last_automation_before}')
-        HostMetric.active_objects.filter(last_automation__lt=last_automation_before).update(
+        rows = HostMetric.active_objects.filter(last_automation__lt=last_automation_before).update(
             deleted=True, deleted_counter=F('deleted_counter') + 1, last_deleted=now()
         )
+        logger.info(f'cleanup_host_metrics: soft-deleted records last automated before {last_automation_before}, affected rows: {rows}')
 
     @staticmethod
     def hard_cleanup(threshold=None):
@@ -74,9 +74,9 @@ class HostMetricTask:
             threshold = getattr(settings, 'CLEANUP_HOST_METRICS_HARD_THRESHOLD', 36)
 
         last_deleted_before = now() - relativedelta(months=threshold)
-        logger.info(f'cleanup_host_metrics: hard-deleting records last deleted before {last_deleted_before}')
         queryset = HostMetric.objects.filter(deleted=True, last_deleted__lt=last_deleted_before)
-        queryset._raw_delete(queryset.db)
+        rows = queryset.delete()
+        logger.info(f'cleanup_host_metrics: hard-deleted records which were soft deleted before {last_deleted_before}, affected rows: {rows[0]}')
 
 
 class HostMetricSummaryMonthlyTask:
--- a/awx/main/tasks/host_metrics.py
+++ b/awx/main/tasks/host_metrics.py
@@ -50,17 +50,22 @@ class HostMetricTask:
         :param soft_threshold: (int)
         :param hard_threshold: (int)
         """
-        if soft_threshold is not None:
-            self.soft_cleanup(soft_threshold)
         if hard_threshold is not None:
             self.hard_cleanup(hard_threshold)
+        if soft_threshold is not None:
+            self.soft_cleanup(soft_threshold)
 
         settings.CLEANUP_HOST_METRICS_LAST_TS = now()
 
     @staticmethod
     def soft_cleanup(threshold=None):
-        if not threshold:
+        if threshold is None:
             threshold = getattr(settings, 'CLEANUP_HOST_METRICS_SOFT_THRESHOLD', 12)
+
+        try:
+            threshold = int(threshold)
+        except (ValueError, TypeError) as e:
+            raise type(e)("soft_threshold has to be convertible to number") from e
 
         last_automation_before = now() - relativedelta(months=threshold)
         rows = HostMetric.active_objects.filter(last_automation__lt=last_automation_before).update(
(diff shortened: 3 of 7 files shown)


slemrmartin:
@AlanCoding ready for review.
Needs to be backported to 2.4

awx/main/tasks/helpers.py
```diff
@@ -0,0 +1,8 @@
+from django.utils.timezone import now
+from rest_framework.fields import DateTimeField
+
+
+def is_run_threshold_reached(setting, threshold_seconds):
+    last_time = DateTimeField().to_internal_value(setting.value) if setting and setting.value else DateTimeField().to_internal_value('1970-01-01T00:00')
```
AlanCoding: The call to this is

```python
is_run_threshold_reached(getattr(settings, 'CLEANUP_HOST_METRICS_LAST_TS', None), getattr(settings, 'CLEANUP_HOST_METRICS_INTERVAL', 30) * 86400)
```

And that's incompatible with the `setting.value` syntax, which treats this as a `Setting` object, as opposed to a datetime object.

To state my preference - I would prefer that the `awx/main` app never touches a `Setting` object directly at all, and only deals with their values.

slemrmartin: good catch, fixed

awx/main/management/commands/cleanup_host_metrics.py
```diff
@@ -1,22 +1,24 @@
...
+    Run soft/hard-deleting of HostMetrics
     """
 
-    help = 'Run soft-deleting of HostMetrics'
+    help = 'Run soft/hard-deleting of HostMetrics'
 
     def add_arguments(self, parser):
-        parser.add_argument('--months-ago', type=int, dest='months-ago', action='store', help='Threshold in months for soft-deleting')
+        parser.add_argument('--soft', type=int, nargs='?', default=None, const=0, help='Threshold in months for soft-deleting')
+        parser.add_argument('--hard', type=int, nargs='?', default=None, const=0, help='Threshold in months for hard-deleting')
```
AlanCoding: It would be good for the help text to say "months ago" because I find this non-obvious.

awx/main/tasks/host_metrics.py
```diff
@@ -3,33 +3,80 @@
...
+
+    @staticmethod
+    def hard_cleanup(threshold=None):
+        if not threshold:
+            threshold = getattr(settings, 'CLEANUP_HOST_METRICS_HARD_THRESHOLD', 36)
+
+        last_deleted_before = now() - relativedelta(months=threshold)
+        logger.info(f'cleanup_host_metrics: hard-deleting records last deleted before {last_deleted_before}')
+        queryset = HostMetric.objects.filter(deleted=True, last_deleted__lt=last_deleted_before)
+        queryset._raw_delete(queryset.db)
```
AlanCoding: Given what I know about the `HostMetric` model I don't _think_ this should really behave differently from the normal `.delete()` bulk delete, but I don't care a great deal either way.

awx/main/management/commands/cleanup_host_metrics.py
```diff
@@ -1,22 +1,24 @@
...
 
 
 class Command(BaseCommand):
     """
-    Run soft-deleting of HostMetrics
+    Run soft/hard-deleting of HostMetrics
     """
 
-    help = 'Run soft-deleting of HostMetrics'
+    help = 'Run soft/hard-deleting of HostMetrics'
```
AlanCoding: I'm realizing that I'd like a little more more substance behind these docs. After reading the querysets, I will use my own words to say that.

Soft deleting will _mark_ all host metrics as deleted if they are older than the specified number of months. This is the same as when you delete those host metrics via the API, so this is just a way to automatically do that after a certain amount of time.

Hard deleting will fully delete all host metrics that fit 2 criteria - that they are older than the specified number of months AND they have already been soft-deleted. Practically, I can see why the former criteria shouldn't usually matter, because the soft deletion threshold _should_ be shorter than the hard deletion threshold, but I don't see any guarantees of this.

awx/main/tasks/system.py
```diff
@@ -374,29 +372,6 @@ def cleanup_images_and_files():
...
-    """Run cleanup host metrics ~each month"""
-    # TODO: move whole method to host_metrics in follow-up PR
-    from awx.conf.models import Setting
-
-    if is_run_threshold_reached(
-        Setting.objects.filter(key='CLEANUP_HOST_METRICS_LAST_TS').first(), getattr(settings, 'CLEANUP_HOST_METRICS_INTERVAL', 30) * 86400
-    ):
-        months_ago = getattr(settings, 'CLEANUP_HOST_METRICS_SOFT_THRESHOLD', 12)
-        logger.info("Executing cleanup_host_metrics")
-        HostMetric.cleanup_task(months_ago)
```
AlanCoding, line 441: Is `cleanup_task` method now replaced by the soft delete method in the task code? If so, then it should be deleted. EDIT: I mean deleted in the location where it was defined.

awx/main/tasks/host_metrics.py
```diff
@@ -3,33 +3,80 @@
...
+        settings.CLEANUP_HOST_METRICS_LAST_TS = now()
+
+    @staticmethod
+    def soft_cleanup(threshold=None):
+        if not threshold:
+            threshold = getattr(settings, 'CLEANUP_HOST_METRICS_SOFT_THRESHOLD', 12)
+
+        last_automation_before = now() - relativedelta(months=threshold)
+        logger.info(f'cleanup_host_metrics: soft-deleting records last automated before {last_automation_before}')
+        HostMetric.active_objects.filter(last_automation__lt=last_automation_before).update(
```
AlanCoding: Now I'm very confused about the intent of this - `HostMetric.active_objects` filters to only return the deleted host metrics. So this soft-deletes the already deleted host metrics older than so-many months ago... but why? All that would do is increment `deleted_counter` and `last_deleted`.

The more obvious initial reading I had was that it would mark all host metrics older than so-many months ago as deleted. I'm confused about how this would be useful.

EDIT: never mind, the `.active_objects` returns the host metrics which are _not_ deleted, so it seems to kind of be working.

awx/main/tasks/host_metrics.py
```diff
@@ -3,33 +3,80 @@
...
+            deleted=True, deleted_counter=F('deleted_counter') + 1, last_deleted=now()
+        )
+
+    @staticmethod
+    def hard_cleanup(threshold=None):
+        if not threshold:
```
AlanCoding: The logic of `if not threshold` is blocking me from running

```
awx-manage cleanup_host_metrics --soft 0
```

to test this PR, because `0` is a falsy value.

awx/main/tasks/host_metrics.py
```diff
@@ -3,33 +3,80 @@
...
+    @staticmethod
+    def soft_cleanup(threshold=None):
+        if not threshold:
+            threshold = getattr(settings, 'CLEANUP_HOST_METRICS_SOFT_THRESHOLD', 12)
+
+        last_automation_before = now() - relativedelta(months=threshold)
+        logger.info(f'cleanup_host_metrics: soft-deleting records last automated before {last_automation_before}')
+        HostMetric.active_objects.filter(last_automation__lt=last_automation_before).update(
+            deleted=True, deleted_counter=F('deleted_counter') + 1, last_deleted=now()
+        )
```
AlanCoding, line 73: I would like you to log the number of affected rows, you can do this by saving the result of this query, and then that is an integer for the number of affected rows. This is a somewhat higher quality information than the existing log which says it's going to run the task. It might be good to consolidate those logs, so that you write both the `last_automation_before` and the affected row count.

awx/main/management/commands/cleanup_host_metrics.py
jjwatt: Not super familiar with this, and I could be dead wrong, but it looks like if `hardlimit=softlimit`, then behavior would be order-dependent. Like if soft runs before hard vs. hard running before soft: Wouldn't you get different results?

AlanCoding: The order is that soft deletion runs before hard deletion, so assuming `softlimit<hardlimit` then anything older than the hard limit will be deleted. It appears possible that un-deleted records older than the hard limit will be kept if they have never been soft-deleted.

awx/main/tasks/host_metrics.py
slemrmartin: I'm not sure about that, if you're 100% sure that `delete()` won't generate tons of SQL queries or query like `DELETE FROM .. WHERE id IN(...)` etc. which is problematic for 10000+ rows, I can change it. But it's not obvious from the documentation.

awx/main/tasks/host_metrics.py
slemrmartin: it's not specified in the ADR that soft delete could have the meaning "delete all". 
My understanding is, that the sense of this feature is different, so this command from my point of view is invalid

slemrmartin: @Ladas ^ what's your thought?

awx/main/tasks/system.py
slemrmartin: that's true, it was my plan, I forgot

awx/main/management/commands/cleanup_host_metrics.py
slemrmartin: to solve these assumptions, we can 
- make both --soft and --hard arguments required and check if soft < hard
- rely on settings only, so no arguments possible (but then it'll be harder to test it), but it'll do the same like scheduler does

--- 

about the comments, when this review targets to the "help", I don't see any multiline "help" string among all commands. So we can keep one-liner for the help and extend class comment, ok?

awx/main/management/commands/cleanup_host_metrics.py
slemrmartin: Can you type the whole sentence? 
- "Mark host metrics last automated <x> months ago"
- "Delete host metrics soft deleted <y> months ago"

awx/main/tasks/host_metrics.py
slemrmartin: ..or as I posted above, other option is not allow to specify parameters and get values from settings only.

AlanCoding: I don't fully know what your reasons are for not allowing "delete all", but using a value of `0` months is going to be first and most obvious approach to testing. If conflicts with objectives you have, then I don't need to give pushback, we can hack around it for testing.

But if this command doesn't allow `0` months retention, silently doing nothing when given the `0` value is not the right behavior. You could throw a command error in that case. Exact implementation is up to you, I just don't want it to work in a frustrating and unexpected way.

awx/main/management/commands/cleanup_host_metrics.py
AlanCoding: I don't think we should hesitate to have multi-line command help text. Example:

https://github.com/ansible/awx/blob/devel/awx/main/management/commands/register_default_execution_environments.py

You could, for example, add a sentence to this help that says "this is also run periodically as a system task" to establish more context.

But I don't think you need the details discussed in this thread in the _command_ help text, since they can mostly be covered in the help text for the `--hard` and `--soft` options, which is discussed in another thread.

> check if soft < hard

No objections to this. It's fairly common to see `raise CommandError('you did this wrong')`, and this is often better than documenting the restriction (since users often will not read it).

awx/main/tasks/host_metrics.py
AlanCoding: If `.delete()` did generate a ton of SQL queries like that, then that would be a bit of a red flag that bypassing it could cause an integrity problem (which could be fixed by model changes), but I just have to see for myself using `awx-manage shell_plus --print-sql`

```python
In [20]: HostMetric.objects.filter(deleted=True, last_deleted__lt=hm.last_deleted + timedelta(seconds=1)).delete()
DELETE
  FROM "main_hostmetric"
 WHERE ("main_hostmetric"."deleted" AND "main_hostmetric"."last_deleted" < '2023-08-03 19:42:41.740404+00:00'::timestamptz)

Execution time: 0.001741s [Database: default]
Out[20]: (1, {'main.HostMetric': 1})
```

This gives about the _very basic_ level of validation I needed to say we _don't_ have any red flags, and that either approach should give the same behavior.

To wrap this up, my own _preference_ would be the normal .delete() syntax, but I'm fairly confident behavior will be correct either way and :+1: either way.

awx/main/management/commands/cleanup_host_metrics.py
AlanCoding: Good to ask. I'll give a shot, I can point you to our docs person if you want more feedback.

 - `--soft`: Perform soft-deletion of all host metrics last automated this-many months ago or before. This is the same as issuing a DELETE request to /api/v2/host_metrics/N/ for all host metrics that match the criteria.
 - `--hard`: Permanently erase from the database all host metrics last automated this-many months ago or before. This operation happens after the soft deletion has finished. This must be greater than the soft delete option.

I think this could cover all the docs topics we've discussed in this review.

slemrmartin: I got an answer that it's not wanted to allow configurable command. It'll be driven by settings only.

awx/main/management/commands/cleanup_host_metrics.py
slemrmartin: I got an answer that it's not wanted to allow configurable command. It'll be driven by settings only.  Accordingly I'll update the comments

awx/main/tasks/host_metrics.py
slemrmartin: I got an answer that it's not wanted to allow configurable command. It'll be driven by settings only.

jjwatt: iiuc, you can probably get what Alan and you both want with, *explicitly*, `if threshold is None` <-- This explicitly says that the value was not passed in. You should be doing this anyway from a readability perspective because you're using `None` to indicate that the value was not passed in. If you just say `if not threshold` then you'll catch a lot of situations where the value *WAS* passed in: e.g., `myfunc(threshold=0)`, `myfunc(threshold=False)`, even `myfunc(threshold=[])`, `myfunc(threshold=set())`, or the best one what if it's an empty string that you thought wasn't? `thing=""; myfunc(threshold=thing`, etc. because it matches any "falsy" value, and all of those are falsy). Add in command line parsing, like in Alan's example, and you can see how taking user input and not being explicit can be an issue here. Another one, in some shells this might do exactly what `--soft 0` does: `awx-manage cleanup_host_metrics --soft ` and this probably will do the same thing unless the type is checked for `str`: `awx-manage cleanup_host_metrics --soft ""`

AlanCoding (approved):

awx/main/tests/functional/commands/test_cleanup_host_metrics.py
```diff
@@ -0,0 +1,83 @@
...
+        if threshold == settings.CLEANUP_HOST_METRICS_HARD_THRESHOLD:
+            HostMetricTask().cleanup(hard_threshold=0)
+        else:
+            HostMetricTask().cleanup(hard_threshold=threshold)
+        assert HostMetric.objects.count() == 6
+
+        hostnames = []
+        qs = HostMetric.objects.order_by('hostname')
+        for host in list(qs):
+            hostnames.append(host.hostname)
```
AlanCoding: This is confusing, can you just do `hostnames = set(HostMetric.objects.values_list('hostname', flat=True))`? Then compare sets in the end.

awx/main/management/commands/cleanup_host_metrics.py
```diff
@@ -1,22 +1,22 @@
...
+    This command provides cleanup task for HostMetric model.
+    There are two modes, which run in following order:
+    - soft cleanup
+    - - Perform soft-deletion of all host metrics last automated 12 months ago or before.
+        This is the same as issuing a DELETE request to /api/v2/host_metrics/N/ for all host metrics that match the criteria.
+    - - updates columns delete, deleted_counter and last_deleted
+    - hard cleanup
+    - - Permanently erase from the database all host metrics last automated 36 months ago or before.
+        This operation happens after the soft deletion has finished.
     """
```
AlanCoding, line 17: What if you tried assigning this docstring to a `help` variable as in the Django docs?

https://docs.djangoproject.com/en/4.2/howto/custom-management-commands/

slemrmartin: @AlanCoding we discussed that previously. The `help` variable is here below docstring and should be one-liner. 
It's not supposed to be long description like the docstring (also according to the doc you posted ^)

awx/main/tasks/host_metrics.py
slemrmartin: now it receives 0 or "0", raises exceptions ValueError or TypeError on "", set() etc.

RedOrNot:
QA Approved!
manual testing pass

PR #14257, opened 2023-07-18
Upgrade to Python 3.11

jessicamack:
##### SUMMARY

Upgrade the default version of Python used to 3.11

##### ISSUE TYPE

##### COMPONENT NAME

 - API
 - CLI

##### AWX VERSION

```
awx: 22.5.1.dev10+g694d190e57
```

##### ADDITIONAL INFORMATION

```

```


relrod:
Jira issue: https://issues.redhat.com/browse/AAP-13369

awx/main/credential_plugins/dsv.py
```diff
@@ -2,7 +2,7 @@
 
 from django.conf import settings
 from django.utils.translation import gettext_lazy as _
-from thycotic.secrets.vault import SecretsVault
+from delinea.secrets.vault import SecretsVault
```
AlanCoding: Also see https://github.com/ansible/awx/pull/14207, it looks like we have 2 credential plugins that import this library. And multiple open PRs doing some fraction of the upgrade.

jessicamack: I think we should rebase and merge the maintainers PR and then I can rebase and include that work.

AlanCoding: That's good with me, I asked the question in that PR, because we should strive to keep all plugins functional.

AlanCoding: I had additional conversation in that PR https://github.com/ansible/awx/pull/14207 and left it with a minor request. I'm completely happy to get that in on its own. Knowing that, it would seem appropriate to remove that change from this python upgrade PR.

jessicamack: I don't mind removing it though I think it might cause errors because the lower versions of python-tss-sdk and python-dsv-sdk don't work with 3.11. Do we want to just ignore the errors for the moment since the other PR will be in shortly and address them?

AlanCoding: Errors installing them (which breaks all builds), errors when using the plugin, or more subtle hypothetical python version compatibility issues which might not affect us anyway?

The PR https://github.com/ansible/awx/pull/14207 has some clear steps forward, but they have not been responded too. Additionally, upgrading `python-dsv-sdk` is not attempted by that. Manually testing imports, I have found that upgrading `python-dsv-sdk` looks fairly straightforward and easy. In both cases, upgrading the library changes the import path.

If those changes really are needed, you could consider taking over that patch from them, including it in here, or doing a quick separate PR where you upgrade those 2 libraries. We have somewhat of a problem that we have a fragile testing situation, for which I have some potential work up at https://github.com/AlanCoding/awx/pull/81. If you want to verify functionality of these plugins after upgrade, you could try the manual instructions at `tools/docker-compose/README.md` which have been tested quite recently.

awx/main/tests/unit/api/test_views.py
```diff
@@ -110,13 +110,13 @@ def exclude(self, **kwargs):
...
+                serializer_class = mocker.patch('awx.api.serializers.InventoryUpdateDetailSerializer')
+                serializer = serializer_class.return_value
+                serializer.to_representation.return_value = {}
 
-                    view = InventoryInventorySourcesUpdate()
-                    response = view.post(mock_request)
-                    assert response.data == expected
+                view = InventoryInventorySourcesUpdate()
+                response = view.post(mock_request)
+                assert response.data == expected
```
TheRealHaoLiu, line 119: @jessicamack why was this changed

jessicamack: The tests were throwing an error and this warning about using mocks as a context manager https://pytest-mock.readthedocs.io/en/latest/remarks.html#usage-as-context-manager so I reformatted the tests.

licenses/aiohttp-retry.txt
jjwatt, line 1: I might not be synced, but we need to make sure we regen/clean-up the licenses that were changed that we went back on, right? Like we regenerated without the full updates so we should have way fewer license changes now iiuc.

jjwatt: Like maybe `git checkout devel -- $allthoselicensefiles` and then regen and then `git add $newlicensefiles` `&& git commit`, etc.

awx/main/tests/unit/api/test_views.py
jjwatt: A long shot, but could this have been related to the test that we eventually disabled? Can you try reverting these changes and then run tests a) with that naughty test disabled and b) with the naughty test enabled? If we can get past tests with this change reverted, and maybe even get past the other test with this change reverted, then we should revert it.

jjwatt (commented):
Please take a look at the code you reformatted early on and at the license files (if you haven't already).

awxkit/test/cli/test_options.py
```diff
@@ -175,9 +175,9 @@ def test_choices(self):
         assert '--verbosity {0,1,2,3,4,5}' in out.getvalue()
 
     def test_actions_with_primary_key(self):
+        page = OptionsPage.from_json({'actions': {'GET': {}, 'POST': {}}})
```
jbradberry, line 178: :+1:

awx/settings/defaults.py
```diff
@@ -476,7 +476,7 @@
 
 # Django Caching Configuration
 DJANGO_REDIS_IGNORE_EXCEPTIONS = True
-CACHES = {'default': {'BACKEND': 'awx.main.cache.AWXRedisCache', 'LOCATION': 'unix:/var/run/redis/redis.sock?db=1'}}
+CACHES = {'default': {'BACKEND': 'awx.main.cache.AWXRedisCache', 'LOCATION': 'unix://var/run/redis/redis.sock?db=1'}}
```
relrod: Just noting here so it doesn't get lost

```suggestion
CACHES = {'default': {'BACKEND': 'awx.main.cache.AWXRedisCache', 'LOCATION': 'unix:///var/run/redis/redis.sock?db=1'}}
```

Otherwise I think the path will be relative, it'll get parsed as `unix://` and `var/run/redis/redis.sock?db=1` so the path will be relative to the cwd.

jjwatt (approved):

licenses/aiohttp-retry.txt
jjwatt: bump. @jessicamack Do you have a local copy where you fixed all the license files, etc? That seems to be one of the last pieces to at least mark this as complete. That and @relrod feedback. If you want, I can try to do the cleanup.

PR #14258, opened 2023-07-18
Return back chdir to project sync to support project-local roles/coll…

eric-zadara:
…ections

##### SUMMARY
PR https://github.com/ansible/awx/pull/14081 added support for 3 requirements.yml paths, however it removed the `chdir` argument to `ansible.builtin.command`.

When a requirements.yml contains a project-relative source, this causes the following error(s) from `ansible-galaxy` during project sync:
```
ERROR! Neither the collection requirement entry key 'name', nor 'source' point to a concrete resolvable collection artifact. Also 'name' is not an FQCN. A valid collection name must be in the format <namespace>.<collection>. Please make sure that the namespace and the collection name contain characters from [a-zA-Z0-9_] only.

Could not find ./collections/ansible-netcommon-4.1.0.tar.gz.
```
This was discovered after recently updating from AWX 22.2.0 to 22.5.0, and has brought all related AWX jobs to a halt because the project sync itself is now failing.

Besides not being confident of where the EE is executing this command within the container/pod(as an admin), this also breaks support for project-local sources, and/or requirements files generated by `ansible-galaxy collection download --download-path <path> -r requirements.reference.yml`. ex:
```
collections:
  - name: ./collections/ansible-netcommon-4.1.0.tar.gz
    version: 4.1.0
  - source: ./collections/arbitrarycollection
    type: dir
```

##### ISSUE TYPE

##### COMPONENT NAME
 - API
 - Other

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```
1. Create `./requirements.galaxy.yml` file with `collections: [{name: 'ansible.netcommon', version: '4.1.0'}]` as yaml
2. Run `ansible-galaxy collection download -r requirements.galaxy.yml`
3. Revise `./collections/requirements.yml` to include `./collections/` prefix for all collections
4. Commit all changes
5. Perform project sync with role/collection installation enabled
6. Sync fails

Running `ansible-galaxy collection install -r ./collections/requirements.yml` directly by hand from the root of the project directory locally is successful, same as it was prior to AWX 22.5.0
```

### DIFF
--- a/awx/playbooks/project_update.yml
+++ b/awx/playbooks/project_update.yml
@@ -255,6 +255,9 @@
           tags:
             - install_collections
             - install_roles
+      module_defaults:
+        ansible.builtin.command:
+          chdir: "{{ project_path | quote }}"
 
       # We combine our additional_galaxy_env into galaxy_task_env so that our values are preferred over anything a user would set
       environment: "{{ galaxy_task_env | combine(additional_galaxy_env) }}"


AlanCoding:
Thanks, this is a good find and I agree with the overall change. However, you added this to 3 tasks, all of which are in a block. Can you please use `module_defaults:` to only set this in 1 place, and then the block will apply it to all tasks inside of it. To clarify, `args:` is the old way this was done (and is not compatible with `block:`), but `chdir:` can be used as a direct option in this module, so it can be used with module_default, without using args.

So take docs

https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_module_defaults.html

I am thinking something that looks like

```yaml
- block:
    - name: "..."
      ansible.builtin.command:
        cmd: "..."
    - more:
  module_defaults:
    ansible.builtin.command:
      chdir: "..."
```

john-westcott-iv:
Good catch on this functionality change. In the previous PR I hadn't considered installing a collection from within the project itself.

eric-zadara:
Sure, will make the change shortly.

I knee-jerked to whatever-it-was-before so I could get my environment back online ASAP.

AlanCoding (approved):

john-westcott-iv (approved):

relrod:
@eric-zadara thanks for the patch!

Issue #13828, opened 2023-04-10
Django-radius w/ python future.


john-westcott-iv:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

In PR https://github.com/ansible/awx/pull/13827 we unpinned the python-future library which had a hard pinning because of django-radius.

There is an opened PR against django-radius itself which would remove the future dependency: https://github.com/robgolding/django-radius/pull/27.

If that PR gets merged we should back out #13827 and update `ansible/django-radius`. Additionally, if that PR gets released, we should remove django-radius from `requirements_git.txt` all together and go back to a released version. If that PR will not be merged/released we should maybe consider merging it into `ansible/django-radius`, add some testing and offer that up as an alternative to `django-radius`?

### AWX version

N/A

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [X] Other

### Installation method

N/A

### Modifications

no

### Steps to reproduce

N/A

### Expected results

N/A

### Actual results

N/A

AlanCoding:
Just to link our fork - https://github.com/ansible/django-radius

AlanCoding:
keeping this open https://github.com/ansible/django-radius/pull/2

If nothing else happens in the next 3 months, this gives a good simplification almost for free.

PR #14265, opened 2023-07-19
Re-run the updater script after upstream removal of future

AlanCoding:
##### SUMMARY
I just merged this https://github.com/ansible/django-radius/pull/2 which picked up an upstream change from that library.

Addresses (partially) https://github.com/ansible/awx/issues/13828

##### ISSUE TYPE

##### COMPONENT NAME
 - API


jbradberry (approved):

AlanCoding:
@simaishi I just want to ask one last time here, would it be easier for you if we made separate branches for before / after this change at https://github.com/ansible/django-radius/branches? I don't expect any problems if old versions pick up this patch (it should be harmless either way), but I just don't want to introduce unexpected changes.

AlanCoding:
We have a stable branch for the fork, so we're using this here.

PR #14268, opened 2023-07-20
Split job-end host update into 2 queries to be slightly more efficient

AlanCoding:
##### SUMMARY
What I want to do here is break up the existing query into 2 that are somewhat more scale-tolerant. Previously:

```sql
UPDATE "main_host"
SET "last_job_id" = (CASE
                         WHEN ("main_host"."id" = '13') THEN '36'
                         WHEN ("main_host"."id" = '14') THEN '36'
                         WHEN ("main_host"."id" = '15') THEN '36'
                         WHEN ("main_host"."id" = '16') THEN '36'
                         WHEN ("main_host"."id" = '12') THEN '36'
                         ELSE NULL
                     END)::integer,
    "last_job_host_summary_id" = (CASE
                                      WHEN ("main_host"."id" = '13') THEN '17'
                                      WHEN ("main_host"."id" = '14') THEN '16'
                                      WHEN ("main_host"."id" = '15') THEN '14'
                                      WHEN ("main_host"."id" = '16') THEN '15'
                                      WHEN ("main_host"."id" = '12') THEN '18'
                                      ELSE NULL
                                  END)::integer
WHERE "main_host"."id" IN ('13',
                           '14',
                           '15',
                           '16',
                           '12')
```

Now, for the same setup, instead of running that, it would run these two queries.

```sql
UPDATE "main_host"
SET "last_job_id" = '36'
WHERE "main_host"."id" IN ('13',
                           '14',
                           '15',
                           '16',
                           '12')
```

```sql
UPDATE "main_host"
SET "last_job_host_summary_id" = (CASE
                                      WHEN ("main_host"."id" = '13') THEN '17'
                                      WHEN ("main_host"."id" = '14') THEN '16'
                                      WHEN ("main_host"."id" = '15') THEN '14'
                                      WHEN ("main_host"."id" = '16') THEN '15'
                                      WHEN ("main_host"."id" = '12') THEN '18'
                                      ELSE NULL
                                  END)::integer
WHERE "main_host"."id" IN ('13',
                           '14',
                           '15',
                           '16',
                           '12')
```

So this is making use of the obvious redundancy in the first CASE logic, where it always set it to `'36'`. You obviously don't need a CASE there... it's always 36.

We may want some kind of deeper analysis to see how or if we could do a similar thing for the `last_job_host_summary` update... but it does not look good.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/models/events.py
+++ b/awx/main/models/events.py
@@ -586,19 +586,15 @@ class JobEvent(BasePlaybookEvent):
             JobHostSummary.objects.bulk_create(summaries.values())
 
             # update the last_job_id and last_job_host_summary_id
-            # in single queries
             host_mapping = dict((summary['host_id'], summary['id']) for summary in JobHostSummary.objects.filter(job_id=job.id).values('id', 'host_id'))
+            Host.objects.filter(id__in=host_mapping.keys()).update(last_job_id=job.id)
+
             updated_hosts = set()
             for h in all_hosts:
-                # if the hostname *shows up* in the playbook_on_stats event
-                if h.name in hostnames:
-                    h.last_job_id = job.id
-                    updated_hosts.add(h)
                 if h.id in host_mapping:
                     h.last_job_host_summary_id = host_mapping[h.id]
                     updated_hosts.add(h)
-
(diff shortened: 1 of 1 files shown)


AlanCoding:
This has been made obsolete by @benthomasson approach to turn it into an annotation.

Issue #14051, opened 2023-05-26
Incorrect capacity for remote execution nodes


fosterseth:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.
- [X] I am **NOT** reporting a (potential) security vulnerability. (These should be emailed to `security@ansible.com` instead.)

### Bug Summary

Incorrect reporting when task container resource limits are set (in k8s). We are scaling the CPU and Memory values by `SYSTEM_TASK_ABS_CPU`, even for execution nodes

https://github.com/fosterseth/awx/blob/5afdfb11351dca986d62c7ebbfc7438c55274449/awx/main/utils/common.py#L759

we need at least these two changes,

`def save_health_data`
and
`get_cpu_effective_capacity` (which is called by refresh_capacity_fields)

### AWX version

devel

### Select the relevant components

- [ ] UI
- [ ] UI (tech preview)
- [X] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

kubernetes

### Modifications

no

### Steps to reproduce

add limits to your task container
```yaml
task_resource_requirements:
  requests:
    cpu: 250m
    memory: 1Gi
  limits:
    cpu: 1000m
    memory: 2Gi
```

then add a remote execution node. The "cpu" and "memory" fields on the api/v2/instances endpoint are not correct.

### Expected results

execution node capacity should not be affected by SYSTEM_TASK_ABS_CPU

### Actual results

it is

djyasin:
TODO: need a backport to 2.3 and 2.4

djyasin:
PR: https://github.com/ansible/awx/pull/14315

PR #14272, opened 2023-07-20
Incorrect capacity for remote execution nodes 14051

djyasin:
##### SUMMARY

Fixes issue #14051
Correct reporting for task container resource limits set (in k8s), revising the handling of execution nodes specifically. 

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
22.5.1
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/awx/main/access.py
+++ b/awx/main/access.py
@@ -366,7 +366,7 @@ class BaseAccess(object):
             report_violation = lambda message: None
         else:
             report_violation = lambda message: logger.warning(message)
-        if validation_info.get('trial', False) is True or validation_info['instance_count'] == 10:  # basic 10 license
+        if validation_info.get('trial', False) is True:
 
             def report_violation(message):  # noqa
                 raise PermissionDenied(message)
--- a/awx/main/utils/licensing.py
+++ b/awx/main/utils/licensing.py
@@ -175,7 +175,12 @@ class Licenser(object):
             license.setdefault('pool_id', sub['pool']['id'])
             license.setdefault('product_name', sub['pool']['productName'])
             license.setdefault('valid_key', True)
-            license.setdefault('license_type', 'enterprise')
+            if sub['pool']['productId'].startswith('S'):
+                license.setdefault('trial', True)
+                license.setdefault('license_type', 'trial')
+            else:
+                license.setdefault('trial', False)
+                license.setdefault('license_type', 'enterprise')
             license.setdefault('satellite', False)
             # Use the nearest end date
             endDate = parse_date(sub['endDate'])


awx/main/utils/common.py
```diff
@@ -759,14 +759,20 @@ def get_corrected_cpu(cpu_count):  # formerlly get_cpu_capacity
...
-    if env_abscpu is not None:
+    if env_abscpu is not None and obj.node_type != 'execution':
         return convert_cpu_str_to_decimal_cpu(env_abscpu)
     elif settings_abscpu is not None:
         return convert_cpu_str_to_decimal_cpu(settings_abscpu)
-
     return cpu_count  # no correction
 
 
+def get_cpu_k8s(self, obj):
```
AlanCoding: this isn't used anywhere right now

awx/main/utils/common.py
```diff
@@ -750,7 +750,7 @@ def convert_cpu_str_to_decimal_cpu(cpu_str):
     return max(0.1, round(cpu, 1))
 
 
-def get_corrected_cpu(cpu_count):  # formerlly get_cpu_capacity
+def get_corrected_cpu(cpu_count, obj):  # formerlly get_cpu_capacity
```
AlanCoding: I would prefer a change to the method signature that is backward-compatible like adding a kwarg like `, is_execution=False` which can be flipped to True when necessary.

Also, I don't see any corresponding change to set the new option, presumably in the task code for running health checks.

awx/main/utils/common.py
```diff
@@ -760,7 +760,7 @@ def get_corrected_cpu(cpu_count):  # formerlly get_cpu_capacity
     settings_abscpu = getattr(settings, 'SYSTEM_TASK_ABS_CPU', None)
     env_abscpu = os.getenv('SYSTEM_TASK_ABS_CPU', None)
 
-    if env_abscpu is not None:
+    if env_abscpu is not None and obj.node_type != 'execution':
```
AlanCoding: I don't think `obj` should be defined anymore.

awx/main/models/ha.py
```diff
@@ -354,7 +354,7 @@ def local_health_check(self):
         except redis.ConnectionError:
             errors = _('Failed to connect to Redis')
 
-        self.save_health_data(awx_application_version, get_cpu_count(), get_mem_in_bytes(), update_last_seen=True, errors=errors)
+        self.save_health_data(self, awx_application_version, get_cpu_count(), get_mem_in_bytes(), update_last_seen=True, errors=errors)
```
AlanCoding: `self` is automatically passed by python as the first argument in bound methods like these.

PR #14287, opened 2023-07-26
fixes docs link for controller versions >= 4.3

AlexSCorey:
##### SUMMARY
We have broken docs links for users that are on versions >= 4.3 because at that version we stopped using z stream releases.  
This work resolves that.  

##### ISSUE TYPE

##### COMPONENT NAME
 - UI


keithjgrant (changes requested):
Can you rewrite this without the nested ternary? Those get really confusing to read

keithjgrant (approved):

PR #14300, opened 2023-07-31
optional listener port UI

fosterseth:
##### ISSUE TYPE

##### COMPONENT NAME

 - API
 - UI

### DIFF
--- a/awx/ui/src/screens/Instances/Shared/InstanceForm.js
+++ b/awx/ui/src/screens/Instances/Shared/InstanceForm.js
@@ -63,8 +63,7 @@ function InstanceFormFields({ isEdit }) {
         label={t`Listener Port`}
         name="listener_port"
         type="number"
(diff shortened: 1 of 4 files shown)


awx/ui/src/screens/Instances/Shared/InstanceForm.js
```diff
@@ -146,17 +145,16 @@ function InstanceForm({
           description: instance.description || '',
           node_type: instance.node_type || 'execution',
           node_state: instance.node_state || 'installed',
-          listener_port: instance.listener_port || 27199,
+          listener_port: instance.listener_port,
           enabled: instance.enabled || true,
           managed_by_policy: instance.managed_by_policy || true,
-          peers_from_control_nodes: instance.peers_from_control_nodes
-            ? true
-            : !isEdit,
```
fosterseth, line 154: I can't remember why we had this conditional. I just want a default value of False

PR #14307, opened 2023-08-01
Modified URL includes to be dynamic

john-westcott-iv:
##### SUMMARY

This PR should have a net 0 affect, it just tidies up the urls.py file by allowing for dynamically including URL files from sub directories rather than having to import and include them. Each URL file can have three variables in it:
 * `extend_urls` - These are appended as is to the url list being extended
 * `urls` - These are imported with a namespace of the file by default. For example, any URLs in a file called 'object.py' would be included as being under `/object/`.
 * `entry_point` - This modifies the namespace used for the `urls` variables from the filename without the .py to this value. Again, using our example of `object.py` if we set `entry_point = 'something'` the urls defined in urls would be added under `/something/`.

##### ISSUE TYPE

 - New or Enhanced Feature
 
##### COMPONENT NAME

 - API
 
##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```


awx/api/urls/urls.py
```diff
@@ -2,177 +2,47 @@
...
-    BulkHostCreateView,
-    BulkJobLaunchView,
 )
 
-from awx.api.views.mesh_visualizer import MeshVisualizer
 
-from awx.api.views.metrics import MetricsView
-from awx.api.views.analytics import AWX_ANALYTICS_API_PREFIX
+def extend_urls(urls, dir_name):
+    urls_dir = path.join(path.dirname(inspect.getfile(ApiRootView)), '..', '..', 'api', 'urls', dir_name)
```
john-westcott-iv: I'm not in love with finding another known class and then going relative down to api/urls. Maybe we just want to import something from awx.api and go from there? Anyone else have another trick?

awx/api/urls/urls.py
```diff
@@ -2,177 +2,47 @@
...
+            if hasattr(url_module, 'entry_point'):
+                entry_point = url_module.entry_point
+            urls.append(re_path(entry_point, include(url_module.urls)))
+            imported_something = True
+        if hasattr(url_module, 'extend_urls'):
+            urls.extend(url_module.extend_urls)
+            imported_something = True
+
+        if not imported_something:
+            print(f"Not loading {fully_qualified_module} missing urls or extend_urls")
```
john-westcott-iv: I should probably change this to a logger instead of just a print.

jbradberry (changes requested):
Rather than create our own new pattern, I would like to see us make use of the modern Django pattern for this, e.g. `path("whatever/", include("foo.bar"))`.

https://docs.djangoproject.com/en/4.2/topics/http/urls/#including-other-urlconfs

jbradberry:
Closing due to staleness.

PR #14311, opened 2023-08-02
Modify main/0185 to set aside the json fields that might be a problem

jbradberry:
##### SUMMARY

Related to issue https://github.com/ansible/awx/issues/14314

Rename them, then create a new clean field of the new jsonb type. We'll use a task to do the data conversion.

AAP-14795

##### ISSUE TYPE

##### COMPONENT NAME
 - API


jbradberry:
For reference, look at

- https://github.com/ansible/awx/pull/6032
- https://github.com/ansible/awx/pull/11947

awx/main/tasks/system.py
```diff
@@ -86,6 +88,8 @@ def dispatch_startup():
     if settings.IS_K8S:
         write_receptor_config()
 
+    convert_jsonfields()
```
AlanCoding: I've debugged too many things that error in this startup method, I would like to have this in a blanket try-except. I don't ever want a failure here to prevent dispatcher startup.

awx/main/tasks/system.py
```diff
@@ -129,6 +133,49 @@ def inform_cluster_of_shutdown():
...
+                    """,
+                    (i, i + batchsize),
+                )
+                rows = cursor.rowcount
+                logger.debug(f"Batch {i} to {i + batchsize} copied on {table}, {rows} rows affected.")
+
+            column_expr = ', '.join("DROP COLUMN {column}" for column in columns)
+            cursor.execute(f"ALTER TABLE {table} {column_expr};")
+
+        logger.warning(f"Migration of {table} to jsonb is finished.")
```
AlanCoding, line 182: I think you should add `{column_expr}` to this log

AlanCoding: No, I was wrong, never mind

You could add `{columns}`, but meh.

awx/main/tasks/system.py
```diff
@@ -129,6 +133,49 @@ def inform_cluster_of_shutdown():
...
+                    update {table}
+                      set {column_expr}
+                      where id >= %s and id < %s;
+                    """,
+                    (i, i + batchsize),
+                )
+                rows = cursor.rowcount
+                logger.debug(f"Batch {i} to {i + batchsize} copied on {table}, {rows} rows affected.")
+
+            column_expr = ', '.join("DROP COLUMN {column}" for column in columns)
```
AlanCoding: I don't get it, why wouldn't this be `DROP COLUMN {column}_old`?

awx/main/tasks/system.py
jbradberry: fair enough

awx/main/tasks/system.py
jbradberry: ... because I'm wrong

jbradberry:
further to-dos:

- [x] remove any non-null constraints from the `*_old` fields, to avoid `IntegrityError` on new rows
- [x] make sure that the copy sql commands don't clobber real data in the new fields

awx/main/models/__init__.py
```diff
@@ -99,6 +100,58 @@
...
+        return
+
+    # fmt: off
+    fields = [
+        ('main_activitystream', (
+            'deleted_actor',
+            'setting',
+        )),
+        ('main_job', (
+            'survey_passwords',
```
AlanCoding, line 114: I'm asking myself "will there be any bugs due to a field being _unexpectedly_ null / empty?". To some extent, yes, this may expose encrypted values (not the plain text values).

I have argued that the `survey_passwords` field is not necessary in the first place https://github.com/ansible/awx/pull/13091 and thus, we could remove it.

If someone was concerned about the security / optics of this, you could copy the `display_extra_vars` versions from my attempted removal of the field, and hopefully allow it to become completely vestigial.

AlanCoding:
Initial impressions are :+1: this is probably good to start testing with.

awx/main/models/__init__.py
jbradberry: I'd argue vehemently that making any changes to these fields that do not recapitulate what has already landed in old main/0185 is out of scope.  This is surgery to make the change that is already out in the wild work more robustly.

awx/main/tasks/system.py
```diff
@@ -129,6 +136,51 @@ def inform_cluster_of_shutdown():
...
+        while True:
+            executor = MigrationExecutor(connection)
+            if not executor.migration_plan(executor.loader.graph.leaf_nodes()):
+                break
+            time.sleep(120)
+
+        logger.warning(f"Migrating json fields for {table}: {', '.join(columns)}")
+
+        with connection.cursor() as cursor:
+            for i in itertools.count(0, batchsize):
```
AlanCoding, line 158: Why this as opposed to using `range`? Apparently so that the loop can be infinite until the `break` is encountered. I get it, and this should work. But for this particular problem I would think that the maximum relevant pk could be known from the outset, and you could write a more robust loop with a finite max by getting the max affected pk in advance.

jbradberry: I considered this to be a more robust loop for the situation, considering that this operation could be taking non-trivial amounts of time and the table may grow.  Granted, any newly added rows will not be relevant and so the differences are mainly stylistic.

AlanCoding:
I went about manually testing this. Some notes...

To start with data-loading, I created a JT with a survey and a password. I also confirmed that I had activity stream entries with changed settings. I verified that these exist and their counts with some canary queries:

```python
list(a for a in ActivityStream.objects.values_list('setting', flat=True) if a)
[job.survey_passwords for job in Job.objects.filter(survey_passwords__isnull=False) if job.survey_passwords]
```

To migrate backwards, I had to patch the reverse taggit migrations...

```diff
diff --git a/awx/main/migrations/0186_drop_django_taggit.py b/awx/main/migrations/0186_drop_django_taggit.py
index c9c767b74c..fd56ba3fd3 100644
--- a/awx/main/migrations/0186_drop_django_taggit.py
+++ b/awx/main/migrations/0186_drop_django_taggit.py
@@ -20,8 +20,8 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RunSQL("DROP TABLE IF EXISTS taggit_tag CASCADE;"),
-        migrations.RunSQL("DROP TABLE IF EXISTS taggit_taggeditem CASCADE;"),
-        migrations.RunPython(delete_taggit_contenttypes),
-        migrations.RunPython(delete_taggit_migration_records),
+        migrations.RunSQL("DROP TABLE IF EXISTS taggit_tag CASCADE;", reverse_sql=[]),
+        migrations.RunSQL("DROP TABLE IF EXISTS taggit_taggeditem CASCADE;", reverse_sql=[]),
+        migrations.RunPython(delete_taggit_contenttypes, migrations.RunPython.noop),
+        migrations.RunPython(delete_taggit_migration_records, migrations.RunPython.noop),
     ]
```

and then in `devel`, as a hacked method, I did

```
awx-manage migrate main 0184
```

and that worked fine. Then I switched off `devel` to this branch. Once here, I did

```
awx-manage migrate main
```

Once I checked my canary queries, I found that no entries were present. This is because after the migration finished the dispatcher had not been restarted. This is an expectation on the installer, so I manually restarted.

```
supervisorctl restart tower-processes:*
```

That kicked off a number of things. Logs:

```
tools_awx_1     | awx-dispatcher stderr | 2023-08-07 18:39:05,791 WARNING  [-] awx.main.tasks.system Migrating json fields for main_activitystream: deleted_actor, setting
tools_awx_1     | awx-dispatcher stderr | 
tools_awx_1     | awx-dispatcher stderr | 2023-08-07 18:39:05,793 WARNING  [-] awx.main.tasks.system Migrating json fields for main_notification: body
tools_awx_1     | awx-dispatcher stderr | 
tools_awx_1     | awx-dispatcher stderr | 2023-08-07 18:39:05,801 WARNING  [-] awx.main.tasks.system Migrating json fields for main_joblaunchconfig: char_prompts, survey_passwords
tools_awx_1     | awx-dispatcher stderr | 
tools_awx_1     | awx-dispatcher stderr | 2023-08-07 18:39:05,824 WARNING  [-] awx.main.tasks.system Migrating json fields for main_job: survey_passwords
tools_awx_1     | awx-dispatcher stderr | 
tools_awx_1     | awx-dispatcher stderr | 2023-08-07 18:39:05,828 WARNING  [-] awx.main.tasks.system Migration of main_notification to jsonb is finished.
tools_awx_1     | awx-dispatcher stderr | 
tools_awx_1     | 2023-08-07 18:39:05,865 INFO waiting for awx-rsyslogd to stop
tools_awx_1     | 2023-08-07 18:39:05,871 INFO stopped: awx-rsyslogd (exit status 0)
tools_awx_1     | 2023-08-07 18:39:05,878 INFO spawned: 'awx-rsyslogd' with pid 38809
tools_awx_1     | awx-dispatcher stderr | 2023-08-07 18:39:05,883 WARNING  [-] awx.main.tasks.system Migration of main_activitystream to jsonb is finished.
tools_awx_1     | awx-dispatcher stderr | 2023-08-07 18:39:05,883 WARNING  [-] awx.main.tasks.system Migration of main_joblaunchconfig to jsonb is finished.
tools_awx_1     | awx-dispatcher stderr | 
tools_awx_1     | awx-dispatcher stderr | 2023-08-07 18:39:05,907 WARNING  [-] awx.main.tasks.system Migration of main_job to jsonb is finished.
tools_awx_1     | awx-dispatcher stderr | 
tools_awx_1     | awx-dispatcher stderr | 2023-08-07 18:39:05,910 WARNING  [-] awx.main.tasks.system Migrating json fields for main_workflowjob: char_prompts, survey_passwords
tools_awx_1     | awx-dispatcher stderr | 
tools_awx_1     | awx-dispatcher stderr | 2023-08-07 18:39:05,910 WARNING  [-] awx.main.tasks.system Migrating json fields for main_unifiedjob: job_env
tools_awx_1     | awx-dispatcher stderr | 2023-08-07 18:39:05,911 WARNING  [-] awx.main.tasks.system Migrating json fields for main_workflowjobnode: char_prompts, survey_passwords
tools_awx_1     | awx-dispatcher stderr | 
tools_awx_1     | awx-dispatcher stderr | 2023-08-07 18:39:05,966 WARNING  [-] awx.main.tasks.system Migration of main_workflowjob to jsonb is finished.
tools_awx_1     | awx-dispatcher stderr | 
tools_awx_1     | awx-daphne stderr | 2023-08-07 18:39:05,981 DEBUG    cleaning up Redis key b'asgi:group:jobs-status_changed'
tools_awx_1     | awx-daphne stderr | 
tools_awx_1     | awx-daphne stderr | 2023-08-07 18:39:05,984 DEBUG    cleaning up Redis key b'asgi:group:job_events-92'
tools_awx_1     | awx-daphne stderr | 
tools_awx_1     | awx-daphne stderr | 2023-08-07 18:39:05,984 DEBUG    cleaning up Redis key b'asgi:group:jobs-summary'
tools_awx_1     | awx-daphne stderr | 
tools_awx_1     | awx-dispatcher stderr | 2023-08-07 18:39:05,989 WARNING  [-] awx.main.tasks.system Migration of main_workflowjobnode to jsonb is finished.
tools_awx_1     | awx-dispatcher stderr | 2023-08-07 18:39:05,989 WARNING  [-] awx.main.tasks.system Migration of main_unifiedjob to jsonb is finished.
```

So far, looking good. Onto checking the canary queries.

I found the content to be present for both cases. There were 3 jobs with survey_passwords in both cases, expected. There were 62 activity stream settings entries before, but 38 after the process. This isn't necessary a problem with your branch, they could have been blanked in the reverse migration somehow. Tentatively, it looks like no activity stream entries were deleted, it's just that some became unintelligible. Among these, I believe `TOWER_URL_BASE` was likely one, and perhaps some esoteric read_only logic is causing it to mess up, but I really don't know.

Maybe it's a good point to watch out for, but likely due to imperfections in the hacked testing method.

jbradberry:
Confirming for the record -- it was definitely my fault.  I had a boolean logic fail for tables with multiple migrating columns.

AlanCoding (approved):
On re-testing everything seems to be working as intended.

jay-steurer (approved):

PR #14325, opened 2023-08-08
Add a retry to update host facts on deadlocks

AlanCoding:
##### SUMMARY
If updating facts gives a deadlock error, then retry saving them up to 5 times.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/tasks/facts.py
+++ b/awx/main/tasks/facts.py
@@ -58,14 +58,19 @@ def start_fact_cache(hosts, destination, log_data, timeout=None, inventory_id=No
     return None
 
 
+def raw_update_hosts(host_list):
+    Host.objects.bulk_update(host_list, ['ansible_facts', 'ansible_facts_modified'])
+
+
 def update_hosts(host_list):
     if not host_list:
         return
-    for i in range(5):
+    max_tries = 5
+    for i in range(max_tries):
         try:
-            Host.objects.bulk_update(host_list, ['ansible_facts', 'ansible_facts_modified'])
+            raw_update_hosts(host_list)
         except OperationalError as exc:
-            if 'deadlock detected' in str(exc):
+            if 'deadlock detected' in str(exc) and i + 1 < max_tries:
                 logger.info(f'Deadlock saving host facts retry {i}')
                 continue
             else:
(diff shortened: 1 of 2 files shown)


awx/main/tasks/facts.py
```diff
@@ -57,6 +58,21 @@ def start_fact_cache(hosts, destination, log_data, timeout=None, inventory_id=No
     return None
 
 
+def update_hosts(host_list):
+    if not host_list:
+        return
+    for i in range(5):
```
relrod: Future me would love a comment or two here to understand why this retry loop is necessary.

AlanCoding: Sure, this is valid, and as I looked at it, I could predict other behavior problems. So I added a few unit tests so that we at least get coverage for all the code branches.

awx/main/tasks/facts.py
```diff
@@ -57,6 +58,29 @@ def start_fact_cache(hosts, destination, log_data, timeout=None, inventory_id=No
     return None
 
 
+def raw_update_hosts(host_list):
+    Host.objects.bulk_update(host_list, ['ansible_facts', 'ansible_facts_modified'])
+
+
+def update_hosts(host_list):
```
jjwatt: is `update_hosts(host_list,max_tries=5)` not a cool thing to do here? The type in the initializer is immutable, so you won't clobber it for other callers in the fn.

awx/main/tasks/facts.py
```diff
@@ -57,6 +58,29 @@ def start_fact_cache(hosts, destination, log_data, timeout=None, inventory_id=No
...
+    max_tries = 5
+    for i in range(max_tries):
+        try:
+            raw_update_hosts(host_list)
+        except OperationalError as exc:
+            # Deadlocks can happen if this runs at the same time as another large query
+            # inventory updates and updating last_job_host_summary are candidates for conflict
+            # but these would resolve easily on a retry
+            if i + 1 < max_tries:
+                logger.info(f'OperationalError (suspected deadlock) saving host facts retry {i}, error: {exc}')
```
jjwatt: nit: Smell: Writing the string "error" in a `logger.info()` call.

jjwatt (approved):
Approved. With a couple optional nits.

jjwatt (approved):

PR #13829, opened 2023-04-11
Fix: Misconfigured LOG_AGGREGATOR_URL can cause dispatcher to go into a crash loop. #13823 

kloud-byun:
##### SUMMARY
- Integrate URL validation within the construct_rsyslog_conf_template function
- Check for proper scheme, hostname, and port in the URL
- Raise a ValueError for invalid URLs with appropriate messages

##### ISSUE TYPE
 - Bug, Docs Fix or other nominal change

##### COMPONENT NAME
 - Other

##### AWX VERSION
```
awx: 22.0.1.dev33+gf4a6301057
```

##### ADDITIONAL INFORMATION
Before the change, the construct_rsyslog_conf_template function did not validate the URL, which could lead to misconfigurations and errors. The issue occurred when the URL contained a misconfigured scheme or port, such as "http:s//test.com" or "ht:tps//test.com".

After the change, the function validates the URL by checking for the proper scheme, hostname, and port. If an invalid URL is found, it raises a ValueError with an appropriate error message. This helps prevent misconfigurations and ensures that the logging feature works as intended.

```
Before change example output:
- Misconfigured URLs were accepted without validation

After change example output:
- ValueError: Invalid URL for logging: http:s//test.com. Please check the LOG_AGGREGATOR_HOST setting.
- ValueError: Invalid port in URL for logging: ht:tps//test.com. Please check the LOG_AGGREGATOR_HOST setting.
```

PR #14333, opened 2023-08-11
Improve log aggregator host/port validation

### LINKED ISSUE #13823: Misconfigured LOG_AGGREGATOR_URL can cause dispatcher to go into a crash loop.
### ISSUE TYPE
Bug Report
SUMMARY
Misconfiguration of the logging url can cause dispatcher to go into a crash loop.

### ENVIRONMENT
AWX version: devel
AWX install method: setup.sh
Ansible version: 2.9
Operating System:
Web Browser:

### STEPS TO REPRODUCE
misconfigure logging url with something like http:s//test.com
enable logging
restart awx
launch jobs

### EXPECTED RESULTS
Misconfigured logger should not cause a crash loop.

### ACTUAL RESULTS
Traceback is in awx-dispatcher.log

  File "/var/lib/awx/venv/awx/lib64/python3.6/site-packages/awx/main/utils/external_logging.py", line 57, in construct_rsyslog_conf_template
    if parsed.port:
  File "/usr/lib64/python3.6/urllib/parse.py", line 169, in port
    port = int(port, 10)
ValueError: invalid literal for int() with base 10: 's'

  File "/usr/bin/awx-manage", line 11, in <module>
    load_entry_point('awx==3.8.2', 'console_scripts', 'awx-manage')()
  File "/var/lib/awx/venv/awx/lib64/python3.6/site-packages/awx/__init__.py", line 154, in manage
  File "/var/lib/awx/venv/awx/lib64/python3.6/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/var/lib/awx/venv/awx/lib64/python3.6/site-packages/django/core/management/__init__.py", line 375, in execute
  File "/var/lib/awx/venv/awx/lib64/python3.6/site-packages/django/core/management/base.py", line 323, in run_from_argv
  File "/var/lib/awx/venv/awx/lib64/python3.6/site-packages/django/core/management/base.py", line 364, in execute
  File "/var/lib/awx/venv/awx/lib64/python3.6/site-packages/awx/main/management/commands/run_dispatcher.py", line 66, in handle
    consumer.run()
  File "/var/lib/awx/venv/awx/lib64/python3.6/site-packages/awx/main/dispatch/worker/base.py", line 153, in run
    self.worker.on_start()

### ADDITIONAL INFORMATION
Looks like this try except block is written in a way which doesn't account for no LOGGING_AGGREGATOR_PORT being configured:

https://github.com/ansible/awx/blob/devel/awx/main/utils/external_logging.py#L61

relrod:
**Don't merge yet** - I want to add more tests.

##### SUMMARY

- Simplify validation functions for logging settings

- Abstract some host/port parsing logic out in utils.external_logging so the validation function can use it as settings are updated

- Never crash-loop if somehow http logging settings are invalid, just log errors (to the console!) and write a default /dev/null syslog config.

- Add a bunch of new test cases for various scenarios (most of them pass before the changes, but we throw some new errors now when validation fails).

Fixes #13823
Refs #13829

##### ISSUE TYPE

##### COMPONENT NAME

 - API

### DIFF
--- a/awx/main/conf.py
+++ b/awx/main/conf.py
@@ -2,6 +2,7 @@
 import logging
 
 # Django
+from django.utils.encoding import force_str
 from django.utils.translation import gettext_lazy as _
 
 # Django REST Framework
@@ -11,6 +12,7 @@ from rest_framework import serializers
 from awx.conf import fields, register, register_validate
 from awx.main.models import ExecutionEnvironment
 from awx.main.constants import SUBSCRIPTION_USAGE_MODEL_UNIQUE_HOSTS
+from awx.main.utils.external_logging import validate_http_host_port
 
 logger = logging.getLogger('awx.main.conf')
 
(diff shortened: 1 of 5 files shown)


awx/settings/defaults.py
```diff
@@ -869,6 +869,7 @@
         'awx.main.tasks': {'handlers': ['task_system', 'external_logger', 'console'], 'propagate': False},
         'awx.main.analytics': {'handlers': ['task_system', 'external_logger', 'console'], 'level': 'INFO', 'propagate': False},
         'awx.main.scheduler': {'handlers': ['task_system', 'external_logger', 'console'], 'propagate': False},
+        'awx.main.utils.external_logging': {'handlers': ['console']},  # force logging to console, if we log while trying to configure logging
```
AlanCoding: We have an existing method of handling these, by filtering them out (in the filter class).

https://github.com/ansible/awx/blob/284bd8377a2be85ab0ab163e9bd4d670be5085d0/awx/main/constants.py#L73-L81

If you want to _instead of this_ do it by the logging configuration, then go ahead. But I don't think this would work as-is. It would need `'propagate': False` if the intent was to avoid infinite logging loops. Also, in many cases we still want these in the file-based logs (although sometimes we don't), and that can get messier than what you're looking for.

relrod: Good call out, I didn't realize we already had infrastructure in place to exclude things like this. I'll just add the module to that list instead.

awx/main/conf.py
```diff
@@ -913,26 +915,56 @@
...
+            # We know it can't have a port number
+            if ':' in new_host:
+                errors.append(
+                    _('Cannot specify a port number for protocol {new_protocol}, use LOG_AGGREGATOR_PORT for that instead.').format(new_protocol=new_protocol)
+                )
+        elif new_protocol == 'https':
+            # It would be really nice to check that if a port number is set here, then it's not also set in LOG_AGGREGATOR_PORT.
+            # But that would be a breaking change vs. silently preferring the port number in LOG_AGGREGATOR_HOST.
+            try:
+                host, port, parsed = validate_http_host_port(new_host, throw=True)
```
AlanCoding, line 957: nice to reuse the validation method in both places.

awx/main/conf.py
```diff
@@ -913,26 +915,56 @@
 
 
 def logging_validate(serializer, attrs):
-    if not serializer.instance or not hasattr(serializer.instance, 'LOG_AGGREGATOR_HOST') or not hasattr(serializer.instance, 'LOG_AGGREGATOR_TYPE'):
+    needed_attrs = (
+        'LOG_AGGREGATOR_HOST',
+        'LOG_AGGREGATOR_TYPE',
+        'LOG_AGGREGATOR_PROTOCOL',
+    )
+    if not serializer.instance or not all(hasattr(serializer.instance, attr) for attr in needed_attrs) or not attrs.get('LOG_AGGREGATOR_ENABLED', False):
```
AlanCoding, line 923: This is messing me up. I'm trying to manually run your test cases, but essentially all cases are bailing early on this line and not throwing the 400 errors because of that.

Still not sure to do with this information, but that's the big thing for verifying the functionality of this PR.

PabloHiro:
Hey @relrod we are doing some cleanup on old PRs. If this is still valid, do you mind iterating with the team internally and figure out a way forward? We are closing it for now.

Thank you!

PR #14335, opened 2023-08-11
make the default JOB_EVENT_BUFFER_SECONDS 1 seconds

jainnikhil30:
This is in conjunction with experiment done for increase the JOB_EVENT_BUFFER_SECONDS from 0.1 second to 1 second. In a high load scenario with loads of event getting generated, with the default 0.1 seconds we end up with loads of small bins, which is not desirable. With 1 second we get bins with bigger sizes. It is clearly visible with following graphs from Grafana:

- With 0.1 seconds:

0 1sec

- with 1 seconds:

1sec

Thus making the default  JOB_EVENT_BUFFER_SECONDS 1 seconds makes sense.

##### ISSUE TYPE

### DIFF
--- a/awx/settings/defaults.py
+++ b/awx/settings/defaults.py
@@ -210,7 +210,7 @@ JOB_EVENT_WORKERS = 4
 
 # The number of seconds to buffer callback receiver bulk
 # writes in memory before flushing via JobEvent.objects.bulk_create()
-JOB_EVENT_BUFFER_SECONDS = 0.1
+JOB_EVENT_BUFFER_SECONDS = 1
 
 # The interval at which callback receiver statistics should be
 # recorded


AlanCoding:
Let me go ahead and make some comments public - specifically why I'm not worried about this change for _responsiveness_.

When I think of responsiveness, I have manual behavior in mind. If you are trying to do something by clicking, you presumably go to a job template and click the "launch" button. In that case, it is unlikely there are other background jobs running _which are actively making progress_ as most long-running jobs are doing work, but spend their time inside heavy tasks.

If your playbook runs fast (or just runs in bursts, which will always be true) then you'll see delay due to the _redis timeout_.

https://github.com/ansible/awx/blob/5cf93febaa4f3d6f3e520385ce11a64d4ccf5b5f/awx/main/dispatch/worker/callback.py#L89

What's probably normal and predicable (as user is watching the standard out) is that a handful of events come in, go in the buffer, and then there's nothing else to read. As such, it'll stay in the read until the timeout is hit. After that, it will flush. So that 1 second read timeout is the main thing that will delay the time until the user sees the first events. This is particularly true with a significant number of callback workers (as is factually the case, at least 4) that split the events between themselves.

I think it's worth considering how we can increase responsiveness. For example, if the _last_ read was known to be a timeout, we could flush on seeing the first new event (only if stdout is not empty). That would alert the user to a batch of events coming in. However, to back this up with measurements, we should have some first-event timing benchmarks.

AlanCoding (approved):
those are some big :bucket: s

kdelee (approved):

PR #14338, opened 2023-08-14
adding iputils

wolfsoldier47:
##### SUMMARY
- Add iputils

##### ISSUE TYPE
- Docs Fix or other nominal change

##### COMPONENT NAME
 - Other

### DIFF
--- a/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
+++ b/tools/ansible/roles/dockerfile/templates/Dockerfile.j2
@@ -22,6 +22,7 @@ RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial
 RUN dnf -y update && dnf install -y 'dnf-command(config-manager)' && \
     dnf config-manager --set-enabled crb && \
     dnf -y install \
+    iputils \
     gcc \
     gcc-c++ \
     git-core \


fosterseth:
@shanemcd what are your thoughts on including `iputils` into the awx image?

wolfsoldier47:
@fosterseth it will be a good idea to have a direct ping and tracepath command available in the development container at least so the user can check the interconnectivity within docker to docker and to the docker host too.

djyasin:
Hello,
Thank you for your contribution. At this time we would prefer not to add iputils into the container. For now, one way of achieving this is to start a sidecar container with these utilities.

Alternatively, we could consider putting iputils in a new image called `awx_debug`. Our GitHub workflows could be updated to build this image alongside the standard AWX image. If you wanted to update this PR to go more in that direction then we would certainly consider the changes there. 

Thank you again for your time!

-AWX Team

wolfsoldier47:
> Hello, Thank you for your contribution. At this time we would prefer not to add iputils into the container. For now, one way of achieving this is to start a sidecar container with these utilities.
> 
> Alternatively, we could consider putting iputils in a new image called `awx_debug`. Our GitHub workflows could be updated to build this image alongside the standard AWX image. If you wanted to update this PR to go more in that direction then we would certainly consider the changes there.
> 
> Thank you again for your time!
> 
> -AWX Team

Allright then you may take it to awx_debug maybe if you feel oright about adding it to the container then surely you may add it.

dmzoneill (approved):

PR #14351, opened 2023-08-16
Clarify that the license module requires fetching subs prior

phess:
##### SUMMARY
The `awx.license` module requires accessing a Red Hat Satellite or web-based Red Hat Subscription Management servers. However, the `awx.license` module itself won't use these parameters. Instead, the `awx.subscriptions` module should be used for querying Red Hat Satellite or Red Hat Subscription Management servers for a list of subscriptions at some point prior to `awx.license`.

This PR adds an example to the Examples section to clarify that the `awx.subscriptions` module should be used prior to `awx.license`.

##### ISSUE TYPE

##### COMPONENT NAME
 - Collection
 - Docs

##### AWX VERSION

Collection version: awx.awx==22.7.0

##### ADDITIONAL INFORMATION
none

### DIFF
--- a/awx_collection/plugins/modules/license.py
+++ b/awx_collection/plugins/modules/license.py
@@ -52,7 +52,12 @@ EXAMPLES = '''
   license:
     manifest: "/tmp/my_manifest.zip"
 
-- name: Attach to a pool
+- name: Use the subscriptions module to fetch subscriptions from Red Hat or Red Hat Satellite
+  subscriptions:
+    username: "my_satellite_username"
+    password: "my_satellite_password"
+
+- name: Attach to a pool (requires fetching subscriptions at least once before)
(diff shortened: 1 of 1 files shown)


phess:
Hi @john-westcott-iv , this is the change we talked about.
I'm only adding the example to the license module so as to make this change minimal.
Personally, I think an additional note in the "Notes" section of the online documentation would also be nice but I don't think the "Notes" section is encoded inside the module itself and I couldn't find where to add it. Please let me know if you agree with adding a note to "Notes" and whether this can be done inside this repo and this PR.

rebeccahhh (approved):

rebeccahhh:
John thinks this is great (I asked him) so I'm merging it.

PR #14347, opened 2023-08-16
Fixed typos

tuxpreacher:
##### SUMMARY

Variable used to be called DISPATCHER_DB_DOWNTOWN_TOLLERANCE but it is regarding DB downtime and tolerance is spelled incorrectly. Cleaning up the typos for clarity.

##### ISSUE TYPE
 - Bug, Docs Fix or other nominal change

##### COMPONENT NAME
 - Other

PR #14361, opened 2023-08-21
Respect old downtime setting name if user has already set it

AlanCoding:
##### SUMMARY
Followup from https://github.com/ansible/awx/pull/14347/files

The concern is that if anyone was already using this, that it will stop working with that change. We need a window for upgrades without any disruption. Eventually we can remove this.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/dispatch/worker/base.py
+++ b/awx/main/dispatch/worker/base.py
@@ -161,7 +161,7 @@ class AWXConsumerRedis(AWXConsumerBase):
 class AWXConsumerPG(AWXConsumerBase):
     def __init__(self, *args, schedule=None, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


relrod:
> Eventually we can remove this.

Do we have a way to deprecate old settings and show a [`DeprecationWarning`](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) or something in logs, if someone overrides them?

Otherwise, I don't see how we get people to stop using the old one.

AlanCoding:
There are a great deal of settings that are only documented by their existence in the defaults file. That file does have a fairly robust comment with each setting.

The setting has already been renamed in that location.

dmzoneill (approved):

PR #14362, opened 2023-08-21
Upgrade client library for dsv since tss already landed

AlanCoding:
##### SUMMARY
Followup from https://github.com/ansible/awx/pull/14207 and just to finalize and clean things up in our dependencies.

I ran the requirements/updater.sh script to get this output. Yes, it removed a dependency. Sanity check, is this expected?

https://github.com/DelineaXPM/python-tss-sdk/commit/84512b8ee437b9f86f6399956f6b82483ca0fde4

In that commit, they dropped python 3.6 support and in doing so dropped dataclasses, so... :tada: 

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API


awx/main/credential_plugins/dsv.py
```diff
@@ -2,7 +2,11 @@
 
 from django.conf import settings
 from django.utils.translation import gettext_lazy as _
-from thycotic.secrets.vault import SecretsVault
+
+try:
+    from delinea.secrets.vault import SecretsVault
+except ImportError:
+    from thycotic.secrets.vault import SecretsVault
```
andrii-zakurenyi, line 9: **_That's an off-topic question I guess. Feel free to ignore it, no worries._** 

Can you please explain why you've added this `except ImportError` while `python-dsv-sdk>=1.0.4` is set in the `requirements/requirements.in`?

I see the comment "_These module have a try-except for ImportError which will allow using the older library <...>_" in a file with tests, so there is a chance when this code will be shipped with "the older library" regardless of what is required (from requirements file) ?

AlanCoding: This shouldn't get shipped with the older library. But this is a pattern established by the tss plugin in this same folder (which just recently merged). The tests give me the confidence that the library is installing the expected module for both tss and dsv.

jessicamack (approved):

PR #13445, opened 2023-01-18
Add `disable_instance` management command

stanislav-zaprudskiy:
##### SUMMARY

Add `disable_instance` management command, which allows to disable AWX
instance, optionally waiting for all its managed jobs to finish.

related: https://github.com/ansible/awx-operator/pull/1193

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - CLI

##### AWX VERSION

```
awx: 21.10.3.dev35+g3896d58f79
```

##### ADDITIONAL INFORMATION

```console
$ awx-manage disable_instance --wait --retry=180 --retry_sleep=10 --force-color
[32;1mInstance ["awx_1"](https://towerhost/#/instances/1/details) has been disabled[0m
1/180: Waiting 10s before the next attempt to see if the following instance' managed jobs have finished: ["sleep"](https://towerhost/#/jobs/playbook/15)
2/180: Waiting 10s before the next attempt to see if the following instance' managed jobs have finished: ["sleep"](https://towerhost/#/jobs/playbook/15)
3/180: Waiting 10s before the next attempt to see if the following instance' managed jobs have finished: ["sleep"](https://towerhost/#/jobs/playbook/15)
4/180: Waiting 10s before the next attempt to see if the following instance' managed jobs have finished: ["sleep"](https://towerhost/#/jobs/playbook/15)
5/180: Waiting 10s before the next attempt to see if the following instance' managed jobs have finished: ["sleep"](https://towerhost/#/jobs/playbook/15)
6/180: Waiting 10s before the next attempt to see if the following instance' managed jobs have finished: ["sleep"](https://towerhost/#/jobs/playbook/15)
[32;1mDone waiting for instance' managed jobs to finish![0m
[0m
```

PR #14380, opened 2023-08-25
awx-manage: Add enable_instance command

sirjaren:
Taken from:
https://github.com/ansible/awx/pull/13445

Adds the ability to enable an AWX instance via `awx-manage`.

##### SUMMARY
Gives `awx-manage` another command called `enable_instance` which allows an admin to use `awx-manage` to enable a given instance in addition to disabling an instance.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME

 - CLI
 - `awx-manage`

##### AWX VERSION
```
awx: 0.1.dev33458+gee1022b
```

##### ADDITIONAL INFORMATION
This feature completes the idea of being able to disable a queue (instance) within AWX via `awx-manage`, draining jobs, and re-enabling the queue when patching/maintenance/work is completed on a given instance node.

A benefit of using `awx-manage` vs the AWX API is it gives a way to run commands via automation service accounts to perform maintenance without having to be a user or admin of AWX.

For example:

```
# Docker environment
$ docker exec -ti awx_task awx-manage enable_instance --hostname <instance_node>

# K8s environment
$ kubectl -n awx exec -ti pod/awx-task-<hash>-<hash> -c awx-task -- awx-manage enable_instance --hostname awx-task-<hash>-<hash>
```

### DIFF
--- a/awx/main/management/commands/enable_instance.py
+++ b/awx/main/management/commands/enable_instance.py
@@ -1,0 +1,78 @@
+from urllib.parse import urljoin
+
+from argparse import ArgumentTypeError
+
+from django.conf import settings
+from django.core.management.base import BaseCommand, CommandError
+from django.db.models import Q
+
+from awx.main.models import Instance, UnifiedJob
+
+
+class AWXInstance:
+    def __init__(self, **filter):
+        self.filter = filter
+        self.get_instance()
+
+    def get_instance(self):
+        filter = self.filter if self.filter is not None else dict(hostname=settings.CLUSTER_HOST_ID)
+        qs = Instance.objects.filter(**filter)
+        if not qs.exists():
+            raise ValueError(f"No AWX instance found with {filter} parameters")
+        self.instance = qs.first()
+
+    def disable(self):
+        if self.instance.enabled:
+            self.instance.enabled = False
+            self.instance.save()
+            return True
+
+    def enable(self):
+        if not self.instance.enabled:
+            self.instance.enabled = True
+            self.instance.save()
+            return True
+
+    def instance_pretty(self):
+        instance = (
+            self.instance.hostname,
+            urljoin(settings.TOWER_URL_BASE, f"/#/instances/{self.instance.pk}/details"),
+        )
+        return f"[\"{instance[0]}\"]({instance[1]})"
+
+
+class Command(BaseCommand):
+    help = "Enable instance."
+
+    @staticmethod
+    def ge_1(arg):
+        if arg == "inf":
+            return float("inf")
+
+        int_arg = int(arg)
+        if int_arg < 1:
+            raise ArgumentTypeError(f"The value must be a positive number >= 1. Provided: \"{arg}\"")
+        return int_arg
+
+    def add_arguments(self, parser):
+        filter_group = parser.add_mutually_exclusive_group()
+
+        filter_group.add_argument(
+            "--hostname",
+            type=str,
+            default=settings.CLUSTER_HOST_ID,
+            help=f"{Instance.hostname.field.help_text} Defaults to the hostname of the machine where the Python interpreter is currently executing".strip(),
+        )
+        filter_group.add_argument("--id", type=self.ge_1, help=Instance.id.field.help_text)
+
+    def handle(self, *args, **options):
+        try:
+            filter = dict(id=options["id"]) if options["id"] is not None else dict(hostname=options["hostname"])
+            instance = AWXInstance(**filter)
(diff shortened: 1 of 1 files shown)


awx/main/management/commands/enable_instance.py
```diff
@@ -0,0 +1,78 @@
...
+    def disable(self):
+        if self.instance.enabled:
+            self.instance.enabled = False
+            self.instance.save()
+            return True
+
+    def enable(self):
+        if not self.instance.enabled:
+            self.instance.enabled = True
+            self.instance.save()
```
AlanCoding, line 33: I don't want to give an opinion on the overall intent of this PR.

But if this command is added, it is fairly critical that it reduces the database contention problems by using `self.instance.save(update_fields=['enabled'])`. This is the only field it is expected to modify, and this avoids updating unrelated columns with stale data. We have a lot of problems with separate processes modifying instances around the same time.

sirjaren: The same goes for `disable_instance` as well, since it uses `self.instance.save()` instead of specifying a specific field.

jessicamack:
Thank you for opening the PR. We'll let you know if we need changes to be made.

fosterseth:
@sirjaren wondering if we could consolidate the code here with disable_instance.py

I think I'd like to see a single awx-manage command that accepts an argument to enable/disable the instance

e.g.

`awx-manage instance_set_enabled myhostname False`

thoughts?

fosterseth:
or, we can have separate enable and disable commands, but I'd like to see them call out to a shared code in a common.py or similar file

sirjaren:
Apologies on the delay.

I can still take a look at this, if there is any interest.

Initially, the motivation was to support an older AWX installation (pre-K8s), in which disabling/enabling instances allowed for automated patching/rebooting of the underlying node without impacting possibly running jobs.

Currently, I'm using a series of playbooks/roles to manage AWX on K8s, which involves safely bringing down AWX nodes without impacting jobs, patching the nodes, and bringing them down (if needed).

Now that AWX is expected to be on K8s, continued motivation for this PR is somewhat lost as it appears to be better to manage AWX instances directly via K8s vs within the AWX application.

Here is a high-level overview of what managing AWX from a patching/perspective looks like when done on a Docker install vs K8s:

### Docker environment (simplified)
```
# Disable AWX instance (node)
$ docker exec -ti awx_task awx-manage disable_instance --hostname <instance_node>

# Wait for jobs to clear (either via the UI, CLI, programmatically, etc)
# Patch/reboot node

# Enable AWX instance (node)
$ docker exec -ti awx_task awx-manage enable_instance --hostname <instance_node>
```

### K8s environment (simplified)
```
# Disable AWX instance (node)
$ kubectl cordon <instance_node>

# Wait for jobs to clear (either via the UI, CLI, programmatically, etc)

# Evict pods from node
$ kubectl drain <options> <instance_node>

# Patch/reboot node

# Enable AWX instance (node)
$ kubectl uncordon <instance_node>
```

fosterseth:
@sirjaren I think it would be a nice feature to have, especially since a lot of AAP customers still use VM based deployments.

My one suggestion would be to try to share code between enable and disable commands, then we can get it merged in

sirjaren:
@fosterseth 

Not to shirk potential work :wink:, but instead of trying to create an `awx-manage` sub-command just to enable and disable an AWX instance, what if we just documented a method that does effectively the same thing using Django's ORM via `shell_plus`?

Something like this (let's say you want to enable an instance):

```
kubectl -n awx exec pod/awx-task-6896fd6948-2g6lm -c awx-task -- awx-manage shell_plus --quiet-load --command \
'instance = Instance.objects.get(hostname="awx-task-6896fd6948-2g6lm"); instance.enabled = True; instance.save(update_fields["enabled"])'
```

This would work in both AWX on VM (`docker-compose`), K8s, and likely AAP customers not yet on K8s.
Of course, this could be expanded to include some scaffolding, return a value, etc.

Ideally, commands like this would be documented and run in some form of automation where it's not done by hand, which is where the value of `awx-manage` sub-commands are realized.

Otherwise, I do agree that some common code should be used for both these commands (maybe an argument to a sub-command like you suggested earlier) in addition to ensuring the `save()` method only updates fields that are to be changed.

Thoughts?

dmzoneill:
@sirjaren I'm marking this as draft for the time being.
I think this is a slam dunk patch.  If you can provide the requested changes, I can merge!
Thanks for your efforts

PR #14386, opened 2023-08-28
Fixes #14377 Ensure Conjur Enterprise JWT token is Base64-encoded

### LINKED ISSUE #14377: Conjur credential retrieval differs between Conjur OSS and Conjur Enterprise

- [X] I am **NOT** reporting a (potential) security vulnerability. (These should be emailed to `security@ansible.com` instead.)

### Bug Summary

When attempting to retrieve credentials using Ansible AWX 22.7.x from Conjur Enterprise 13, I consistently encounter a 401 error. After thorough investigation, I've identified that the root of the issue lies within the `conjur_backend` function, which executes two requests to Conjur in order to fetch the necessary credential information:

1. Initial authentication is performed through the endpoint `api/authn/<account>/<username>/authenticate/`. The response from this request includes a signature.
2. The actual credential retrieval takes place using the endpoint `api/secrets/<account>/variable/<path_to_secret>/`, where users provide the relevant variable or secret.

Upon closer inspection, it becomes clear that the authentication response (as described in point 1) differs between Conjur OSS and Conjur Enterprise. To illustrate these differences, I wrote a test script, and the subsequent output from the console is documented in the provided images (Image 1: Test Code, Image 2: Console Output).

Image 1:
Screenshot 2023-08-24 at 10 48 32

Image 2:
Screenshot 2023-08-24 at 10 46 50

The Conjur Enterprise API returns a response containing the keys `signature`, `payload`, and `protected`. In contrast, the Conjur OSS response presents the same keys, encoded in base64, though. The result is that the plain message from Conjur Enterprise should be encoded to base64 first, before added as a header to the second request (point 2).

~~An alternative approach involves including the header `{"Accept-Encoding": "base64"}` within the headers for the second request. This adjustment is performed within the `awx/main/credential_plugins/conjur.py` file, wherein the modification is as follows:~~

Original Code:
```python
    lookup_kwargs = {
        'headers': {'Authorization': 'Token token="{}"'.format(token)},
        'allow_redirects': False,
    }
```

Modified Code:
```python
    lookup_kwargs = {
        'headers': {'Authorization': 'Token token="{}"'.format(token), 'Accept-Encoding': 'base64'},
        'allow_redirects': False,
    }
```

~~By incorporating this change, we ensure compatibility with both Conjur OSS and Conjur Enterprise instances, enhancing the credential retrieval process from AWX.~~

---

**EDIT:** Following further investigation into this issue, it appears that my initial solution of adding headers may not be effective. I attempted to rectify this within my local AWX environment, but unfortunately, the issue persisted.

A more viable approach could involve checking whether the authentication response is already in base64 format. This can be accomplished using the following steps:

1. Begin by importing the necessary modules:
```python
import base64
import binascii
```

2. Integrate the following function into your codebase:
```python
def _is_base64(s: str) -> bool:
    try:
        return base64.b64encode(base64.b64decode(s.encode("utf-8"))) == s.encode("utf-8")
    except binascii.Error:
        return False
```

3. Adjust the declaration of the `lookup_kwargs` as shown below. This modification ensures that the authentication token is properly encoded if it's not already in base64 format:
```python
lookup_kwargs = {
    'headers': {'Authorization': 'Token token="{}"'.format(
        token if _is_base64(token) else base64.b64encode(token.encode('utf-8')).decode('utf-8')
    )},
    'allow_redirects': False,
}
```

@infamousjoeg can you help with this? And do you agree with this change?

### AWX version

22.7.1

- [ ] UI
- [ ] UI (tech preview)
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [X] Other

### Installation method

docker development environment

### Modifications

yes

### Ansible version

_No response_

### Operating system

MacOS

### Web browser

Chrome

### Steps to reproduce

- Setup a Conjur Enterprise and a Conjur OSS environment
- Create 2 credentials (both for Conjur Enterprice and OSS)
- Create identical users and variables in both environments
- Test the credential retrieval in both environments

### Expected results

The retrieval of a secret from OSS will work, but the retrieval of a secret from Enterprise will result in a 401 error.

### Actual results

The retrieval of a secret from OSS works, and the retrieval of a secret from Enterprise results in a 401 error.

### Additional information

I tested the suggested code change. That was the only change I did in AWX.

infamousjoeg:
##### SUMMARY
This change implements a base64 encoding check on the JWT returned from a Conjur Enterprise authentication.  Customers reported sporatic issues of the `Accept-Encoding: base64` header actually encoding the returned JWT.  To prevent issues, the check will handle any failed encoding attempts.

Related #14377 

##### ISSUE TYPE

##### COMPONENT NAME
 - UI

##### AWX VERSION
```
awx: 0.1.dev33459+g90e28b7
```

##### ADDITIONAL INFORMATION
n/a

### DIFF
--- a/awx/main/credential_plugins/conjur.py
+++ b/awx/main/credential_plugins/conjur.py
@@ -4,6 +4,8 @@ from urllib.parse import urljoin, quote
 
 from django.utils.translation import gettext_lazy as _
 import requests
+import base64
+import binascii
 
 
 conjur_inputs = {
@@ -50,6 +52,13 @@ conjur_inputs = {
 }
 
 
+def _is_base64(s: str) -> bool:
+    try:
+        return base64.b64encode(base64.b64decode(s.encode("utf-8"))) == s.encode("utf-8")
+    except binascii.Error:
+        return False
+
+
 def conjur_backend(**kwargs):
     url = kwargs['url']
     api_key = kwargs['api_key']
@@ -77,7 +86,7 @@ def conjur_backend(**kwargs):
     token = resp.content.decode('utf-8')
 
     lookup_kwargs = {
-        'headers': {'Authorization': 'Token token="{}"'.format(token)},
+        'headers': {'Authorization': 'Token token="{}"'.format(token if _is_base64(token) else base64.b64encode(token.encode('utf-8')).decode('utf-8'))},
         'allow_redirects': False,
     }
 


Zektono:
Thank you for acting so quickly, @infamousjoeg !

Nenodema:
Thanks! I have successfully verified this change with both Conjur Open Source and Conjur Enterprise.

Zektono (approved):
Is think this PR solves the problem of the inconsistency between the API of Conjur OSS and Conjur Ent.

Only a small consideration: maybe it is slightly more efficient to check if the auth response is JSON, instead of whether it is base64:

```
def _is_json(s: str) -> bool:
    try:
        json.loads(s)
        return True
    except json.JSONDecodeError:
        return False
```


The data volume is marginal, of course, so there is no performance benefit to change this.

infamousjoeg:
> Is think this PR solves the problem of the inconsistency between the API of Conjur OSS and Conjur Ent.
> 
> Only a small consideration: maybe it is slightly more efficient to check if the auth response is JSON, instead of whether it is base64:
> 
> ```
> def _is_json(s: str) -> bool:
>     try:
>         json.loads(s)
>         return True
>     except json.JSONDecodeError:
>         return False
> ```
> 
> The data volume is marginal, of course, so there is no performance benefit to change this.

If you think long-term the change will actually make a difference, I'm happy to implement. Please let me know.

Zektono:
I believe the current code will work for the long term as well, so maybe we can implement it now. Even when Conjur OSS and Enterprise APIs are consistent in the future, the code will still work. We have thoroughly tested this code, and it works well.

When I have tested some improved code, I will contact you.

fosterseth (approved):

fosterseth:
@infamousjoeg thanks for the PR, looks good to us

PR #13839, opened 2023-04-11
Adding PR check to ensure JIRA links are present

john-westcott-iv:
##### ISSUE TYPE

 - Bug, Docs Fix or other nominal change

##### COMPONENT NAME

 - Other

##### AWX VERSION

```

```

##### ADDITIONAL INFORMATION

```

```

PR #14389, opened 2023-08-28
Touchup of PR body checks

AlanCoding:
##### SUMMARY
This reverts https://github.com/ansible/awx/pull/13839 because that is no longer the syntax we want to go with (either the ID in the title or the branch name or the commit) and it looks non-obvious how to do it the new way.

This also restricts the existing PR body check to just the AWX repo, because the issue / PR template is only relevant or desired for the AWX repo, and I don't want it to mess with either personal AWX forks or any organizational fork.

##### ISSUE TYPE

##### COMPONENT NAME
 - API


john-westcott-iv (approved):

.github/workflows/pr_body_check.yml
```diff
@@ -7,6 +7,7 @@ on:
     types: [opened, edited, reopened, synchronize]
 jobs:
   pr-check:
+    if: github.repository_owner == 'ansible' && github.repository == 'awx'
```
AlanCoding: This was mistaken - the repository is "ansible/awx", so need to figure out how to deal with that.

Issue #14394, opened 2023-08-30
AWX Task Pod - `Worker failed to run task awx.main.tasks.system.awx_periodic_scheduler`


Daniel-Vaz:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.
- [X] I am **NOT** reporting a (potential) security vulnerability. (These should be emailed to `security@ansible.com` instead.)

### Bug Summary

On the AWX Task Pod I see the following error message being logged on a loop constantly:

```
2023-08-30 07:38:38,795 ERROR    [-] awx.main.dispatch Worker failed to run task awx.main.tasks.system.awx_periodic_scheduler(*[], **{}
Traceback (most recent call last):
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/dispatch/worker/task.py", line 103, in perform_work
    result = self.run_callable(body)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/dispatch/worker/task.py", line 78, in run_callable
    return _call(*args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/awx/main/tasks/system.py", line 745, in awx_periodic_scheduler
    state = TowerScheduleState.get_solo()
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/solo/models.py", line 55, in get_solo
    obj, created = cls.objects.get_or_create(pk=cls.singleton_instance_id)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/db/models/query.py", line 916, in get_or_create
    return self.get(**kwargs), False
  File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/db/models/query.py", line 640, in get
    raise self.model.MultipleObjectsReturned(
awx.main.models.ha.TowerScheduleState.MultipleObjectsReturned: get() returned more than one TowerScheduleState -- it returned 2!
```

Pod Startup seems normal:
```
[...]

PR #14396, opened 2023-08-30
Remove extra scheduler state save that does nothing

AlanCoding:
##### SUMMARY
While I was looking at https://github.com/ansible/awx/issues/14394 I couldn't figure out what was happening, but I did find this one gigantic red flag.

This save seems to have been introduced in 13c89ab78c081066eecfa145fb51f4e283125bd4 and it didn't make sense then either.

Between the two `state.save()` calls, there is no change made to the `state` object. So what's the point of the save? The timestamp doesn't change from one to the other. The line just re-saves the existing data on the object and seems to be completely pointless.

##### ISSUE TYPE

##### COMPONENT NAME
 - API


relrod (approved):
Maybe it was a nod to the old UNIX days where you had to `sync; sync; sync` before you umounted anything, to make sure caches were written out to disk. Except here it was `save; save; save`

PR #14397, opened 2023-08-30
Enable collection integration tests on GHA

relrod:
##### SUMMARY

There are a number of changes here:

- Abstract out a GHA composite action for running the dev environment
- Update the e2e tests to use that new abstracted action
- Introduce a new (matrixed) job for running collection integration tests. This splits the jobs up based on filename.
- Collect coverage info and generate an html report that people can download easily to see collection coverage info.

**NOTE** We should make the new checks be non-voting for a week or two, and make sure they don't have intermittent flakes.

##### ISSUE TYPE

##### COMPONENT NAME

 - Collection

### DIFF
--- a/.github/actions/run_awx_devel/action.yml
+++ b/.github/actions/run_awx_devel/action.yml
@@ -1,0 +1,89 @@
+# This currently *always* uses the "warm build cache" image
+# We should do something to allow forcing a rebuild, probably by looking for
+# some string in the commit message or something.
+
+name: Run AWX (devel environment)
+description: Runs AWX with `make docker-compose`
+inputs:
+  github-token:
+    description: GitHub Token for registry access
+    required: true
+  build-ui:
+    description: Should the UI be built?
+    required: false
+    default: false
+    type: boolean
+outputs:
+  ip:
+    description: The IP of the tools_awx_1 container
+    value: ${{ steps.data.outputs.ip }}
+  admin-token:
+    description: OAuth token for admin user
+    value: ${{ steps.data.outputs.admin_token }}
+runs:
+  using: composite
+  steps:
+    - name: Get python version from Makefile
+      shell: bash
+      run: echo py_version=`make PYTHON_VERSION` >> $GITHUB_ENV
+
+    - name: Upgrade ansible-core
+      shell: bash
+      run: python3 -m pip install --upgrade ansible-core
+
+    - name: Install system deps
+      shell: bash
+      run: sudo apt-get install -y gettext
+
+    - name: Log in to registry
+      shell: bash
+      run: |
+        echo "${{ inputs.github-token }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
+
+    - name: Pre-pull latest available devel image and build HEAD on top of it
+      shell: bash
+      run: |
+        docker pull ghcr.io/${{ github.repository_owner }}/awx_devel:${{ github.base_ref }}
+        DEV_DOCKER_TAG_BASE=ghcr.io/${{ github.repository_owner }} \
+        COMPOSE_TAG=${{ github.base_ref }} \
+        make docker-compose-build
+
+    - name: Start AWX
+      shell: bash
+      run: |
+        DEV_DOCKER_TAG_BASE=ghcr.io/${{ github.repository_owner }} \
+        COMPOSE_TAG=${{ github.base_ref }} \
+        COMPOSE_UP_OPTS="-d" \
+        make docker-compose
+
+    - name: Update default AWX password
+      shell: bash
+      run: |
+        while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' -k https://localhost:8043/api/v2/ping/)" != "200" ]]
+        do
+        echo "Waiting for AWX..."
+        sleep 5
+        done
+        echo "AWX is up, updating the password..."
+        docker exec -i tools_awx_1 sh <<-EOSH
+          awx-manage update_password --username=admin --password=password
+        EOSH
+
+    - name: Build UI
+      # This must be a string comparison in composite actions:
+      # https://github.com/actions/runner/issues/2238
+      if: ${{ inputs.build-ui == 'true' }}
+      shell: bash
+      run: |
+        docker exec -i tools_awx_1 sh <<-EOSH
+          make ui-devel
+        EOSH
+
+    - name: Get instance data
+      id: data
+      shell: bash
+      run: |
+        AWX_IP=$(docker inspect -f '{{range.NetworkSettings.Networks}}{{.IPAddress}}{{end}}' tools_awx_1)
+        ADMIN_TOKEN=$(docker exec -i tools_awx_1 awx-manage create_oauth2_token --user admin)
+        echo "ip=$AWX_IP" >> $GITHUB_OUTPUT
+        echo "admin_token=$ADMIN_TOKEN" >> $GITHUB_OUTPUT
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,3 +114,135 @@ jobs:
           # needed due to cgroupsv2. This is fixed, but a stable release
           # with the fix has not been made yet.
           ANSIBLE_TEST_PREFER_PODMAN: 1
+
+  collection-integration:
+    name: awx_collection integration
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        target-regex:
+          - name: a-h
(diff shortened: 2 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -114,3 +114,96 @@ jobs:
...
+
+      # Upload coverage report as artifact
+      - uses: actions/upload-artifact@v3
+        with:
+          name: coverage-${{ matrix.target-regex.name }}
+          path: ~/.ansible/collections/ansible_collections/awx/awx/tests/output/coverage/
+
+      - uses: ./.github/actions/upload_awx_devel_logs
+
+  collection-integration-coverage-combine:
```
AlanCoding, line 167: I see that the previous job produces coverage data... what is this combining with?

.github/actions/run_awx_devel/action.yml
```diff
@@ -0,0 +1,92 @@
+# This currently *always* uses the "warm build cache" image
+# We should do something to allow forcing a rebuild, probably by looking for
+# some string in the commit message or something.
+
+name: Run AWX (devel environment)
+description: Runs AWX with `make docker-compose`
```
AlanCoding, line 6: This whole thing is amazing, I'm stoked about this.

.github/actions/run_awx_devel/action.yml
```diff
@@ -0,0 +1,92 @@
...
+runs:
+  using: composite
+  steps:
+    - name: Get python version from Makefile
+      shell: bash
+      run: echo py_version=`make PYTHON_VERSION` >> $GITHUB_ENV
+
+    - name: Remove built-in ansible-core
+      shell: bash
+      run: /usr/bin/python3 -m pip uninstall -y ansible-core
```
AlanCoding: I'm getting a little confused here.

.github/actions/run_awx_devel/action.yml
```diff
@@ -0,0 +1,92 @@
...
+    - name: Install system deps
+      shell: bash
+      run: sudo apt-get install -y gettext
+
+    - name: Log in to registry
+      shell: bash
+      run: |
+        echo "${{ inputs.github-token }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
+
+    - name: Pre-pull image to warm build cache
```
AlanCoding: It looks like you're just running out of that image, not just warming up your cache.

relrod: That comment/name came from the original e2e_tests, IIRC. But yeah, valid, I can change it.

Actually, I think it'd be really cool to cache the *database* every once in a while. Then just let things fetch a relatively recent database and apply the latest few migrations on top of it.

But that's a rainy-day idea.

.github/actions/run_awx_devel/action.yml
relrod: I can add a comment. The idea is: the GHA runner actually ships with ansible-core installed against the default Python. This removes it so we can install a new python (with setup-python) and latest ansible-core.

.github/workflows/ci.yml
relrod: So:

- The matrix jobs will each produce a bunch of coverage files.
- They upload them as an artifact (I wish there were a better way to do that, I don't really want those to stay around as artifacts, they aren't very useful to most people... but we need them all to generate a report.)
- Then this job grabs all the coverage files from all the matrix jobs, shoves them into a place where `ansible-test` knows how to work with them, and then asks `ansible-test` to combine them into a single file and generate a report from that file.

It kind of has to be done this way because `pycoverage` has to know where to look for the actual source code to include it in the reports. And `ansible-test` does some hackery to make that all work out nicely for collections.

relrod:
@AlanCoding thanks for the review/comments.

I am planning to file a few issues for future improvements, the main one that comes to mind is nixing our `github_ci_setup` Makefile target (and related `tools/docker-compose/ansible/smoke-test.yml`).

These do basically the exact same thing as the new action. Let's standardize on one way to set up the dev env in CI.

AlanCoding:
When I introduced the `github_ci_setup`, it was because I lacked the immediate ability to do what you did with the `uses:` here. I get that you probably do want to build the image for these tests, and that will likely somehow intermingle the github actions YAML reuse. Honestly there's already existing stuff that hasn't been combined into the standard setup / running stuff.

.github/actions/run_awx_devel/action.yml
AlanCoding: makes sense, although I didn't have a particular reason to question the stock python they have

relrod:
@AlanCoding 

> I get that you probably do want to build the image for these tests, and that will likely somehow intermingle the github actions YAML reuse.

I don't think it would at all, we can just pass in another variable with `with` like I do for controlling if the UI should get built, and conditionalize based on that. If `build-head`, we `make docker-compose-build`, otherwise we pull the latest available image from ghcr.

> Honestly there's already existing stuff that hasn't been combined into the standard setup / running stuff.

I wasn't sure what you meant by this.

.github/actions/run_awx_devel/action.yml
relrod: So I've updated it so that it builds the image each time now from the current checkout.

I guess we do want that, even here for the collection tests (consider: developing a feature that also has collection changes, we want to test them in lockstep). But since it adds so much time, I might split up the job matrix a bit more to try to get the time back down.

relrod: Thinking more on this and doing some more research -- we're using buildkit's inline caching which only caches the last stage and we use a multistage build. So I think we don't really get to take advantage of the caching here.

I'd like to propose (or at least explore and discuss) making our "official" ghcr.io devel images (the ones that get built on pushes to devel and feature branches) use a [registry cache](https://docs.docker.com/build/cache/backends/registry/) instead. Basically they would push a secondary cache image alongside the branch image.

So we'd have (for example): `ghcr.io/ansible/awx_devel:devel` and `ghcr.io/ansible/awx_devel:devel_cache` or something.

I can't guarantee this will speed things up, but I think there's a chance that it could.

AlanCoding:
> I wasn't sure what you meant by this.

The `github_ci_setup` builds the image, but we have github workflow actions that run `make docker-compose-build`. To be DRY this shouldn't happen. That's what I mean. Because the github actions have expanded in scope so much, it would be good to aggressively consolidate it. It is large enough that it's a maintenance concern. What you're doing is going in the right direction.

> If build-head, we make docker-compose-build, otherwise we pull the latest available image from ghcr.

Something's not clicking for me here, so I'm going to talk in fairly basic terms.

Some PRs will modify our dependencies and make corresponding API code changes (and errors will happen if image & code is out-of-sync) - this is basically guaranteed to happen. Some PRs will modify the collection.

It doesn't make sense to run unconditional tests without first building the image, because that would almost guarantee false-failures whenever something big happens. So running the tests unconditionally (which I'm not telling you to do) forces the issue of building the image.

I would be in favor of both making this conditional to `awx_collection/` and making it non-voting in order to speed this up. In that context, it makes more sense to me to run without building the image. But in the long run, I don't see why we wouldn't build the image before running this.

.github/actions/run_awx_devel/action.yml
AlanCoding: At first glance, I don't think I disagree with your proposal for use of the registry cache, but we certainly want to keep that out of the current scope and I'd like @shanemcd to have a look. What we already have has been working _relatively_ well to keep build times fast and there are sometimes slowdown you see due to out-of-date branches or dependency changes that don't apply to small patches and aren't a real problem.

AlanCoding (approved):

.github/workflows/ci.yml
```diff
@@ -114,3 +114,138 @@ jobs:
...
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run integration tests
+        run: |
+          echo "::remove-matcher owner=python::"  # Disable annoying annotations from setup-python
+          echo '[general]' > ~/.tower_cli.cfg
+          echo 'host = https://${{ steps.awx.outputs.ip }}:8043' >> ~/.tower_cli.cfg
+          echo 'oauth_token = ${{ steps.awx.outputs.admin-token }}' >> ~/.tower_cli.cfg
+          echo 'verify_ssl = false' >> ~/.tower_cli.cfg
+          cd awxkit; pip install -r requirements.txt .; cd ..
```
AlanCoding: What kind of build system black magic caused you to have to install requirements separately from the package? Why is line not just `pip install -e awxkit/`

AlanCoding (approved):

PR #14405, opened 2023-09-01
project: add scm depth option

sinuscosinustan:
##### SUMMARY

Git projects can become big and consume a lot of space and memory as they pull all available revisions however, most times they do not require all commits / revisions of a project but the current state of a branch (e.g. depth=1).

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - API
 - UI  <-- unfinished
 - Collection
 - Docs <-- unfinished


fosterseth:
see @AlanCoding comment on this other PR that attempted to set depth=1

https://github.com/ansible/awx/pull/13717#issuecomment-1475504045

Basically, if depth is 1, we would probably need to also force `scm_delete_on_update` to be true for that project.

same argument perhaps for depth of 2, 3, etc. Maybe if depth is set, the the delete on update should be set.

sinuscosinustan:
> Maybe if depth is set, the the delete on update should be set.

Thanks for the input @fosterseth. For the UI, it should be easy to do an auto-select, but how should this be handled for the API? Same behaviour like in the UI or should it throw an explicit error that `scm_delete_on_update=true` is required when `scm_depth` is set?

AlanCoding:
Closing for now, feel free to open a new PR if you rework it to do what was discussed.

sinuscosinustan:
> Closing for now, feel free to open a new PR if you rework it to do what was discussed.

Which discussion? I asked about the handling within the API and didn’t receive any answer, plus I opened this PR as a draft…

klotzjkl:
Hello,
As an admin with a lot of projects (several hundreds) to handle, I find this request perticulerly useful, because it allows a better control over disk space used by projects; 
Indeed, some users took the bad habit of putting some binaries into some git repos; and even deleted, these binaries remain in the history of the project although there are not referenced anymore.
Also, large repositories imply some network performances drop, because in clustered installations,the project is sent to execution nodes / jobs over network.

PR #14406, opened 2023-09-01
Add check for building docsite

AlanCoding:
##### SUMMARY
To make sure people don't break the build of the docsite

ping @oraNod 

##### ISSUE TYPE

##### COMPONENT NAME
 - Docs

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -17,10 +17,3 @@ jobs:
 
       - name: Assure docs can be built
         run: tox -e docs
-
-      - name: Upload html build as artifact
-        uses: actions/upload-artifact@v3
-        with:
-          name: docsite-html
-          path: docs/docsite/build/html
-          retention-days: 5
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,7 +2,9 @@
 name: Docsite CI
 on:
   pull_request:
-    paths: 'docs/**'
+    paths:
+      - 'docs/**'
+      - '.github/workflows/docs.yml'
 jobs:
   docsite-build:
     name: docsite test build
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -23,3 +23,4 @@ jobs:
         with:
           name: docsite-html
           path: docs/docsite/build/html
+          retention-days: 5
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -21,5 +21,5 @@ jobs:
       - name: Upload html build as artifact
         uses: actions/upload-artifact@v3
         with:
-          name: coverage-report-html
+          name: docsite-html
           path: docs/docsite/build/html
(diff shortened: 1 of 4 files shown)


tvo318 (approved):
I like it!

.github/workflows/docs.yml
```diff
@@ -0,0 +1,25 @@
...
+name: Docsite CI
+on:
+  pull_request:
+    paths:
+      - 'docs/**'
+      - '.github/workflows/docs.yml'
+jobs:
+  docsite-build:
+    name: docsite test build
+    runs-on: ubuntu-latest
```
oraNod, line 11: How about using Fedora? One of my goals is to start using Fedora in all our tests. It'd be cool to start here.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: install tox
+        run: pip install tox
+
+      - name: Assure docs can be built
+        run: tox -e docs
+
+      - name: Upload html build as artifact
```
oraNod: Do we really need this step? We can enable PR previews on Read The Docs when we set up hosting. Admittedly I've put this same step in other places but I'm thinking about disabling it.

If we do want to keep it, how about reducing the retention period so we don't have loads of doc builds hanging around and taking up space. The default is 90 days but we could drop that to ~7. https://github.com/actions/upload-artifact#retention-period

.github/workflows/docs.yml
AlanCoding: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

The choices are Ubuntu, Windows, or Mac.

.github/workflows/docs.yml
AlanCoding: Oh I really wanted your feedback on this. Yes, I'm totally fine to get rid of this and even if we did keep it I agree with dropping the retention time.

I want to defer to whatever method will be used to publish them. So if the readthedocs integration is set up on their side, then it makes no sense to have this. I had in my head that maybe we would have this step, and elsewhere another step that uploaded it to the readthedocs API. That sounds unlikely so I'll push a commit to remove this.

.github/workflows/docs.yml
oraNod: nuts. I could swear I noticed Fedora in another repo somewhere and it's kind of been on my to-do list to start looking at the switch. sorry for the noise.

.github/workflows/docs.yml
oraNod: @AlanCoding Awesome. I can send a PR with the RTD config today and set up the server-side integration with you and TVo as maintainers. Then we can add it to the ansible namespace. Sound OK?

AlanCoding: I don't understand all the proper nouns you reference there, but in general it sounds good. I assume that publishing these is next in your agenda and we can sync up in some other communication channels.

oraNod: apologies, I have a habit of using "RTD" instead of "Read The Docs". I can ping you in Matrix to sync up on the publishing tomorrow if that works.

oraNod (approved):

Issue #14187, opened 2023-06-30
Double-clicking the "Launch" button creates two Jobs


flxbwr:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.
- [X] I am **NOT** reporting a (potential) security vulnerability. (These should be emailed to `security@ansible.com` instead.)

### Bug Summary

If the user accidentally double clicks the "Launch" button of a Job Template, two jobs may be started. This is not an expected behavior and is not a natural user experience.

This bug has already been solved by you two times and now is actual again.
It is well described:
https://github.com/ansible/awx/issues/11633

I looked through the changelog from 21.10.2 to the latest version (22.4.0) and didn't find fix.

### AWX version

21.10.2

### Select the relevant components

- [X] UI
- [ ] UI (tech preview)
- [ ] API
- [ ] Docs
- [ ] Collection
- [ ] CLI
- [ ] Other

### Installation method

kubernetes

### Modifications

no

### Ansible version

2.15.0rc2

### Operating system

Windows 10

### Web browser

Firefox

### Steps to reproduce

1. Open the "Templates" page.
2. Click one of the following repeatedly at high speed.
        - Launch icon on the list of the Job Templates page
        - Launch button on the Details page of the Job Template.
        - Launch button on the Propmt if some "Prompt on launch" stuff is enabled for the Job Template

### Expected results

Only one Job will be created.

### Actual results

Two or more Jobs will be created at the same time.

djyasin:
@mabashian do you know if this bug was addressed between 21.10.2 and the current version? If so, do you think this bug could be resolved via upgrading?

flxbwr:
I updated to actual 22.5.0 and the problem is still present.

mabashian:
This should be addressed with https://github.com/ansible/awx/pull/14407

flxbwr:
> This should be addressed with #14407

@mabashian, Thanks for the answer!

PR #14407, opened 2023-09-02
Prevent the user from launching multiple jobs by rapidly clicking on buttons

mabashian:
##### SUMMARY
link https://github.com/ansible/awx/issues/14187

The easiest way to address this was to go right to the source of the launch/relaunch requests.  We were already keeping track of a launching "state" so I hooked into that at the beginning of each function.  If we detect that we're already launching then we do nothing.

The launch buttons already have logic that disables them while the launch request is firing.  Disabling the launch button inside the prompt wizard (very last step) is much more difficult and would probably require some refactoring.  I decided not to go down that route and figured I'd be more likely to introduce new bugs if I did.

##### ISSUE TYPE

##### COMPONENT NAME
 - UI

##### AWX VERSION
22.5.0 and earlier

### DIFF
--- a/awx/ui/src/components/LaunchButton/LaunchButton.js
+++ b/awx/ui/src/components/LaunchButton/LaunchButton.js
@@ -47,6 +47,9 @@ function LaunchButton({ resource, children }) {
   const [error, setError] = useState(null);
 
   const handleLaunch = async () => {
+    if (isLaunching) {
+      return;
+    }
     setIsLaunching(true);
     const readLaunch =
       resource.type === 'workflow_job_template'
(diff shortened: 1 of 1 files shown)


awx/ui/src/components/LaunchButton/LaunchButton.js
```diff
@@ -47,6 +47,9 @@ function LaunchButton({ resource, children }) {
   const [error, setError] = useState(null);
 
   const handleLaunch = async () => {
+    if (isLaunching) {
+      return;
```
vidyanambiar, line 63: Could we add a toast message here to indicate that the job was launched?

mabashian:
Here's the toast in action:
template_launch_toast

vidyanambiar (approved):
lgtm

akus062381 (commented):
I didn't get to submit my approval before merge, but I did test this.

PR #14408, opened 2023-09-04
allow list for example secrets in docs

oraNod:
##### SUMMARY
InfoSec are reporting lines in the following RST as potential security leaks: docs/docsite/rst/administration/oauth2_token_auth.rst

This PR adds that RST to an allow list so it will be ignored in security scans.

##### ISSUE TYPE

##### COMPONENT NAME
 - Docs

##### AWX VERSION
N/A

##### ADDITIONAL INFORMATION
N/A

### DIFF
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -1,0 +1,5 @@
+[allowlist]
+description = "Documentation contains example secrets and passwords"
+paths = [
+  "docs/docsite/rst/administration/oauth2_token_auth.rst",
+]


tvo318 (approved):
Makes sense to me!

.gitleaks.toml
```diff
@@ -0,0 +1,4 @@
+[allowlist]
+  description = "Documentation contains example secrets and passwords"
+  paths = [
+    "docs/docsite/rst/administration/oauth2_token_auth.rst"
```
AlanCoding: Is it not a problem that this doesn't have list closure `]`?

oraNod: gosh, how silly of me. fixed. and thanks for spotting this.

AlanCoding (approved):

PR #14409, opened 2023-09-05
Added relnotes for 23.0.0

tvo318:
##### SUMMARY
Adds release notes for 23.0.0 based on https://github.com/ansible/awx/releases/tag/23.0.0

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - Docs
 
##### AWX VERSION

```
23.0.0
```

### DIFF
--- a/docs/docsite/rst/release_notes/relnotes.rst
+++ b/docs/docsite/rst/release_notes/relnotes.rst
@@ -5,24 +5,32 @@ Release Notes
 **************
 
 .. index::
-   pair: release notes; v21.13
-   pair: release notes; v21.14
-   pair: release notes; v22.0
-   pair: release notes; v22.1
-   pair: release notes; v22.2
-   pair: release notes; v22.3
-   pair: release notes; v22.4
-   pair: release notes; v22.5
-   pair: release notes; v22.6
-   pair: release notes; v22.7
-   pair: release notes; v22.8
-   pair: release notes; v22.9
-   pair: release notes; v22.10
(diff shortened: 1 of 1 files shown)


tvo318:
@oraNod - please, can you review this also? Thanks!

oraNod (approved):
LGTM. Thanks @tvo318

AlanCoding:
Certainly no issues from me, but this appears to be a new process, and I want to make sure we make things as easy as possible for the future.

Shouldn't this be the same as what's shown here?

https://github.com/ansible/awx/releases/tag/23.0.0

We annotate tags (in the AWX release process) with a lightly sanitized history of commit messages. This is why we have tried to push for better summaries in PR titles (which will be the message after squashing). So I just want to ask if there's something here that should be hooked into that release process.

tvo318:
@AlanCoding - The relnotes are the same as what you linked, but not verbatim. In addition to grammar, spelling, wordsmithing, etc., I've clarified some of them so they are more descriptive what the change was for so readers don't have to cross check in the PR itself to find out.
Having said that, we could either have one of the following processes:
- Have better PR summaries before auto-gen of the relnotes and not rewrite them (developers only)
- Not auto-generate the relnotes and have it done during this process instead (writer only)
- Cleanup PR summaries before auto-gen and not have official release notes (writer only)
- Have both - raw auto-gen list and official relnotes - the way we did it here (writer only)

AlanCoding (approved):

PR #14412, opened 2023-09-05
Add customizable batch_size for cleanup_activitystream and cleanup_jobs

dangoncalves:
##### SUMMARY
Add an option in cleanup_activitystream cli tool in order to customize batch size. This can be useful when default size (set to 500) failed.

##### ISSUE TYPE

##### COMPONENT NAME
 - CLI
 - 
##### AWX VERSION
```
awx: 23.0.1.dev11+g217c76d0b7
```


AlanCoding:
I think this is a great idea, but we unfortunately have special logic to make it usable when launched from the API. This would logically need some additions (although I'd rather refactor it to work more generally)

https://github.com/ansible/awx/blob/e072bb7668d1565ac77af38a65ee937c0b3f303a/awx/main/tasks/jobs.py#L1876

fosterseth:
@dangoncalves 

Would you be interested in adding "batch_size" as a value to pass in via the API? It would only be a couple extra lines of code, where @AlanCoding pointed out. That would made this additional parameter even more useful!

AlanCoding:
also see https://github.com/ansible/awx/pull/13717

dangoncalves:
@AlanCoding @fosterseth : I just updated my PR.

awx/main/tasks/jobs.py
```diff
@@ -1873,6 +1873,11 @@ def build_args(self, system_job, private_data_dir, passwords):
             if system_job.job_type in ('cleanup_jobs', 'cleanup_activitystream'):
                 if 'days' in json_vars:
                     args.extend(['--days', str(json_vars.get('days', 60))])
+                if 'batch_size' in json_vars:
+                    default_value = 100000
+                    if system_job.job_type == 'cleanup_activitystream':
+                        default_value = 500
+                    args.extend(['--batch-size', str(json_vars.get('batch_size', default_value))])
```
AlanCoding:
```suggestion
                    args.extend(['--batch-size', str(json_vars['batch_size'])])
```

By this point you already established that "batch_size" is in the vars, so there's no need to process default values in this location.

dangoncalves: Thanks for code review.

awx/main/tasks/jobs.py
```diff
@@ -1873,6 +1873,8 @@ def build_args(self, system_job, private_data_dir, passwords):
             if system_job.job_type in ('cleanup_jobs', 'cleanup_activitystream'):
                 if 'days' in json_vars:
                     args.extend(['--days', str(json_vars.get('days', 60))])
+                if 'batch_size' in json_vars:
```
fosterseth, line 1876: we want to move under the

```python
if system_job.job_type == 'cleanup_jobs':
```

block, because `batch-size` is not an argument to the cleanup_activitystream system job

fosterseth:
@AlanCoding do you have ideas for test coverage around this?

dangoncalves:
Hello @AlanCoding,

Just a gentle reminder: we are waiting for your answer :)

AlanCoding:
Thanks. Could you do the DCO signoff before we merge?

https://github.com/ansible/awx/blob/159dd62d84df3ac14d274571724eee0324c2ca61/DCO_1_1.md?plain=1#L4

dangoncalves:
Done

AlanCoding (approved):

PR #14413, opened 2023-09-05
Add readthedocs configuration

oraNod:
##### SUMMARY
This PR adds Read The Docs configuration to host the content generated from `docs/docsite/rst`.

##### ISSUE TYPE

##### COMPONENT NAME
 - Docs

##### AWX VERSION
N/A

##### ADDITIONAL INFORMATION
N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -166,5 +166,6 @@ use_dev_supervisor.txt
 awx/ui_next/src
 awx/ui_next/build
 
-# Docs build
+# Docs build stuff
 docs/docsite/build/
+_readthedocs/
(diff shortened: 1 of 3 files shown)


oraNod:
@AlanCoding Please take a look whenever you're ready.

A future improvement could be to add positional args to the tox command so we can pass something like `python3 -m tox -e docs -- --strict --site-dir=_readthedocs/html/` instead of using the `mkdir` and `mv` commands. However I would prefer to do that as a follow on PR.

AlanCoding:
might be good to add "_readthedocs" to the .gitignore file

Also, linter wants

```
./.readthedocs.yaml
  4:1       warning  missing document start "---"  (document-start)
```

because it's a yaml file, it's checked by yamllint. If that might cause trouble there is an ignore list somewhere.

AlanCoding:
I edited the ISSUE TYPE because our checks look for highly specific syntax with that.

AlanCoding (approved):

Issue #13233, opened 2022-11-25
awx-cli bulk delete jobs


r0b1ndot:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Feature type

Enhancement to Existing Feature

### Feature Summary

Currently the awx-cli supports only deleting  one job per ID like:
```
awx job delete ID1
```

Is there the possibility or plan to support more than one job ID or a bulk action (delete all jobs)?
The  frontend is capable of deleting all selected jobs, but for CLI usage this feature would be awesome. Maybe also combine that with filter usage.

### Select the relevant components

- [ ] UI
- [ ] API
- [ ] Docs
- [ ] Collection
- [X] CLI
- [ ] Other

### Steps to reproduce

awx job delete ID1 ID2

### Current results

only ID1 is deleted

### Sugested feature result

adding a bulk feature
* all jobs
* jobs using filter
* multiple jobs

Issue #3679, opened 2019-04-10
RFE: Allow reordering instance groups without having to remove and re-add them


msilmser:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
- Customer request

- Customer would like to be able to set a "Default" IG for job template runs rather than individually specifying specific IGs to be available for job templates.
- This would set a "preferred" IG for job template runs, and only get distributed to a different instance group if the default was at max capacity.
- This comes from a job template being inadvertently launched against an isolated instance group, rather than a standard group. This caused the job template to fail.

- An alternative solution to this would be to weight the instance groups so a job will only launch against an isolated node if either there is no remaining capacity at any other instance group, or it was intentionally specified.

wenottingham:
This is best done by setting the fallback IG as the first IG on the organization.

matburt:
Ideally what you would do here is define an Instance Group at the Org level and then only override it on job templates that need to be targeted at something different. You can also set the IG at the inventory level.

The thing about this is that IF you override it at the job template it will *not* fallback to the Org or inventory group. This is by design... picking the wrong group here from an override could be bad.

What should be done, is if you need to fallback then you should add the instance group as a fallback on the job template itself.

matburt:
You can see more details in the top section here: https://docs.ansible.com/ansible-tower/latest/html/administration/clustering.html

wenottingham:
Reopening... the fallout request is to be able to re-order the instance groups on an org to effectively re-set priorities without having to remove and recreate them all.

matburt:
This is going to require some refactoring of the relationship between instance groups and the resources that contain them rather than a simple m2m relationship.  Since right now you can't really control the ordering of those.

msilmser:
When I attempt to reorder the instance groups to change priority, the Tower group always resets to the front of the line when I save. Is this order dictated somewhere else in Tower that would override the list of IGs in the Org itself?

AlanCoding:
Link https://github.com/ansible/awx/issues/2980

Someone is asking for the ability to order hosts there. We lack an engineering mechanism for satisfying any of these such requests.

matburt:
An idea we've discussed with this:
Adjust the m2m associate to allow giving a position `{"associate":..., "id": n, "to_position": x}`

This necessitates a data migration to apply this to the existing association but would maintain backwards compatibility with existing association semantics. The api side would be responsible for recalculating position as things move around.

ryanpetrello:
[...]

Issue #19, opened 2017-09-06
Feature: Select All/Batch Operations


wenottingham:
AWX Feature Spec - Select All/Batch Operations
==============================================

User Persona
------------
AWX admin/org admin

Use cases
---------
In prior releases, complex filtering and sorting was added throughout AWX,
allowing users to easily see the 5/10/100 items they want to operate on. We
should allow them to easily select some or all of these filtered items at
once, and operate on them.

Example:
* As a user editing my inventory, I want to select all subgroups, or hosts in a group, with a single click, and then delete them/run a command on them/etc.
* As a admin/org admin/user assigning permissions to multiple job templates, or adding permissions for multiple teams, I should be able to select all items that match my filter with a single click to then add permissions.
* As an admin managing job templates, I should be able to select with a single click all items matching my filter, and delete them/launch them/etc.
* As an admin cleaning old job history, I should be able to select with a single click all items matching my filter, and delete them.

This should have proper support for pagination, including:
* selecting all items on the page
* selecting all items that match the filter regardless of pagination

API Considerations
------------------

Likely need support for batched operations, including returning status on
items that have changed.

UI/UX Considerations
--------------------

Likely needs support for refreshing a list properly after batched operations
change multiple items.

Docs
----

QA Suggestions
--------------

Notes
-----

Examples:
- Gmail multiple select and operate
- GitHub issues multiple select and operate

Updates/Changelog
-----------------
2017-09-06: initial/updated version

daemenseth:
Is there some progression on this issue?

wenottingham:
Not currently; when it is there will likely be more updates in-ticket.

spsingh1982:
Any update on this?

donnelll:
Umm...going to ask this again since this is HUGE for us since we offer it as SAAS internally to our customers; "Any Update on this ?"

ryanpetrello:
@donnelll if there's progress made on this, you'll see it here.  Like any other feature or enhancement in open source software, it's a matter of time/priority, and finding somebody to put in the work.

bac-w:
Seems to me the issue has been in place since Sep 2017. When might we reasonably expect the fix?

daemenseth:
Is there any progress on this issues, or have someone a commandline solution for me

mili988:
is there a solution for this? is it possible to delete old jobs from commandline (docker container)?

wenottingham:
The new AWX UI allows for this.

PR #14417, opened 2023-09-06
[WIP] Adding Host bulk delete

Avilir:
##### SUMMARY

This PR present new capability to the API, deleting host from inventory as bulk in one API call instead of deleting them one by one

##### ISSUE TYPE

 - New or Enhanced Feature

##### COMPONENT NAME

 - API

##### ADDITIONAL INFORMATION

- With this change you could delete a bulk of hosts as following:

Post to `/api/v2/bulk/host_delete`
```
{
    "inventory": 1,
    "hosts": [1, 2, 3, 4]
}
```
The above will delete the 4 hosts with id's : 1, 2, 3, 4 from inventory ID 1

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -2226,22 +2226,15 @@ class BulkHostDeleteSerializer(serializers.Serializer):
             if request.user not in inv.admin_role:
                 raise serializers.ValidationError(_(f'Inventory with id {inv.id} not found or lack permissions to add hosts.'))
 
-        inventory_hosts = list(Host.objects.get_queryset())
+        # Getting list of all host objects in the inventory, filtered by the list of the hosts to delete
+        inventory_hosts = list(x for x in list(Host.objects.get_queryset()) if x.id in attrs['hosts'])
+
         if len(inventory_hosts) == 0:
             raise serializers.ValidationError(_(f'f"There are no hosts in inventory : {inv}.'))
 
-        attrs['in_inv'] = list()
-        hosts_to_delete = list()
-        for host in attrs['hosts']:
-            for inv_host in inventory_hosts:
-                if host == inv_host.id:
-                    attrs['in_inv'].append(inv_host)
-                    inventory_hosts.remove(inv_host)
-                    hosts_to_delete.append(host)
-                    break
-        if len(hosts_to_delete) < len(attrs['hosts']):
+        if len(inventory_hosts) < len(attrs['hosts']):
             raise serializers.ValidationError(
-                {"inventory": "Not all hosts are in the inventory", "hosts": list(set(attrs['hosts']).difference(hosts_to_delete))}
+                {"inventory": "Not all hosts are in the inventory", "hosts": list(set(attrs['hosts']).difference(inventory_hosts))}
             )
         return attrs
 
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -2146,6 +2146,7 @@ class BulkHostCreateSerializer(serializers.Serializer):
         if request and not request.user.is_superuser:
             if request.user not in inv.admin_role:
                 raise serializers.ValidationError(_(f'Inventory with id {inv.id} not found or lack permissions to add hosts.'))
+
         current_hostnames = set(inv.hosts.values_list('name', flat=True))
         new_names = [host['name'] for host in attrs['hosts']]
         duplicate_new_names = [n for n in new_names if n in current_hostnames or new_names.count(n) > 1]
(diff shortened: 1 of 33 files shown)


kdelee (changes requested):
comments from architecture review:
prefer using DELETE method and query params to indicate the ids instead of POST and the request data
allow hosts to be from multiple inventories, but this adds new requirements:
  * if user is not system administrator, 
    * get list of pks of inventories that all the hosts belong to
    * check is inventory admin of all those
    * return error with list of inventories not allowed
 Either delete all or nothing

AlanCoding:
I wanted to link some prior issues, and I'll have some notes about the angle they were coming at the issue from. The UI version of bulk actions was formally resolved, but only by doing a UI-only solution which is pretty self-evidently not performant. I do think there is some confusion in the history, where someone started at some point assuming that the UI would do this with matching API work, and we would implement the API bulk actions as part of that. The way it turned out is that this never happened and then the host / launch bulk actions came along well after that as people felt pain and needed it for other tooling.

UI focused
Feature: Select All/Batch Operations
https://github.com/ansible/awx/issues/19

related for associations
RFE: Allow reordering instance groups without having to remove and re-add them
https://github.com/ansible/awx/issues/3679

CLI-focused
awx-cli bulk delete jobs
https://github.com/ansible/awx/issues/13233

AlanCoding:
I stumbled on d2e0b2628726ba6bafc2c647bb28ae535e5ff0f7 today, which is, in fact, a host bulk deletion functionality. I think the UI still uses this but would need to check by deleting an inventory source.

AlanCoding:
Reading more on RESTful implementations:

https://stackoverflow.com/questions/21863326/delete-multiple-records-using-rest

https://stackoverflow.com/questions/2421595/restful-way-for-deleting-a-bunch-of-items

There does seem to be a bit of discouraging a DELETE to a list view, hacking it together with ids or the querystring, and the better alternative starts to look a bit more like what you originally had here.

Avilir:
> Reading more on RESTful implementations:
> 
> https://stackoverflow.com/questions/21863326/delete-multiple-records-using-rest
> 
> https://stackoverflow.com/questions/2421595/restful-way-for-deleting-a-bunch-of-items
> 
> There does seem to be a bit of discouraging a DELETE to a list view, hacking it together with ids or the querystring, and the better alternative starts to look a bit more like what you originally had here.

I am a little bit confuse here, do you think I should revert to the original approach using POST API call ?

awx/api/serializers.py
```diff
@@ -2201,6 +2199,64 @@ def create(self, validated_data):
...
+        # Extract the hosts ids from the querystring
+        d = request.query_params.get('pk_ids', '')
+        if d == '':
+            raise serializers.ValidationError({"Error": 'No hosts ids to delete'})
+        ids = [int(pk) for pk in d.split(',')]
+        if len(ids) > settings.BULK_HOST_MAX_DELETE:
+            raise serializers.ValidationError({"Error": 'Number of hosts exceeds system setting BULK_HOST_MAX_DELETE'})
+
+        # Getting list of all host objects, filtered by the list of the hosts to delete
+        attrs['in_inv'] = Host.objects.get_queryset().filter(pk__in=ids)
```
kdelee: might use `prefetch_related('inventory')` this may save some queries
example: https://github.com/ansible/awx/blob/8feeb5f1fae951332ea4b520e5618982109da570/awx/main/scheduler/task_manager.py#L98

awx/api/serializers.py
```diff
@@ -2201,6 +2199,64 @@ def create(self, validated_data):
...
+            )
+
+        # Getting list of all inventories that hosts to delete can be in
+        hosts_inv = set([host.inventory for host in attrs['in_inv']])
+        for inv in hosts_inv:
+            if inv.kind != '':
+                raise serializers.ValidationError(_('Hosts can only be deleted from manual inventories (not smart or constructed types).'))
+            if request and not request.user.is_superuser:
+                if request.user not in inv.admin_role:
+                    raise serializers.ValidationError(_(f'Inventory with id {inv.id} not found or lack permissions to delete hosts from.'))
```
kdelee: Experiement if you can do single query that finds out if user has inventory admin on all the inventories
```
Inventory.objects.filter(id__in=hosts_inv, admin_role__contains(request.user)
```
I don't know if thats actually possible but something to that effect

Avilir:
From what i read and tested, there are inconclusive thinking about the Better approach   POST / DELETE call.

In any case, we need to send the list of IDs to the endpoint:
   POST  -  as json object
   DELETE - as query string

For organization with huge number of hosts that will delete and create hosts constantly, the host ID can become a large int, and to manage large query string with large integers can be very difficult.

This is the main reason why I am in favor of the POST approach, and i would like to get some feedback about that.

AlanCoding:
Yes, looking around, I have found that opinions on RESTful patterns pretty clearly favor a POST as opposed to a DELETE with the ids as a filter or query string, and I now agree that would be the option of least-surprise.

Avilir:
Some scaling test results in graphs :
image

image

Avilir:
Replaced by #14462

PR #14420, opened 2023-09-06
Stop using the bulky test-playbooks in tests where possible

AlanCoding:
##### SUMMARY
The test-playbooks repo has 40,000 commits by a webhook bot and these are brought over in the standard kind of clone, so this is trying to move us away from reliance on the test-playbooks and instead using the repo for the Demo Project which is relatively quiet

https://github.com/ansible/ansible-tower-samples

##### ISSUE TYPE

##### COMPONENT NAME
 - Collection


AlanCoding:
This looks marginally effective, as the `(i-p, ^[i-p])` check is 11min 17sec, which is like 15 seconds faster than a reference PR.

dmzoneill (approved):

PR #14421, opened 2023-09-06
Remove `tower_legacy` module_utils that appears unused

AlanCoding:
##### SUMMARY
This does stuff like processing the config file... but that is already done in the main controller_api.py module utils file.

:wastebasket: 

NOTE: This requires bumping the major version of AWX, which I'm :100: cool with

##### ISSUE TYPE

##### COMPONENT NAME
 - Collection


relrod:
@AlanCoding Assuming `TowerLegacyModule` still actually works, I think it be better to have a deprecation cycle here.

`module_utils` are public facing API (people could in theory write custom modules using our `module_utils` and if this suddenly goes away, we break them with no warning).

`ansible-core` has built-in handling for deprecation warnings, we just need to emit them in the output.

So in the `TowerLegacyModule` we define, we could call `self.deprecate(...)` in the overloaded constructor. In the other functions, we take the module as a parameter, so we can `module.deprecate(...)`.

If this doesn't actually work at all anymore, then probably doesn't matter as much.

AlanCoding:
> So in the TowerLegacyModule we define, we could call self.deprecate(...) in the overloaded constructor.

What do you mean "in the overloaded constructor"? As you say, warning is a `self.` call. IMO the most likely way someone will get burned by this is if they had an unused import (similar to what our test did), but the deprecation machinery gives us no way to warn about that.

relrod:
> What do you mean "in the overloaded constructor"? 

I was just saying, we already overload the constructor, we can just call deprecate() there.

For the import issue... yeah that's a concern. I'm not sure of a good way around that. :(

AlanCoding:
^ we could, I just don't think it's worth the effort. I'd rather merge this as it is.

chrismeyersfsu (approved):

PR #14422, opened 2023-09-06
Consolidate image and server setup in multiple checks

AlanCoding:
##### SUMMARY
Some time ago, it was me who introduced `make github_ci_setup` and the runner one too. I figure that it is easiest if I am the one to delete it, since we have now seen an independent solution evolve for the same problem, via shared github actions.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/.github/actions/awx_ci_setup/action.yml
+++ b/.github/actions/awx_ci_setup/action.yml
@@ -1,0 +1,36 @@
+# This currently *always* uses the "warm build cache" image
+# We should do something to allow forcing a rebuild, probably by looking for
+# some string in the commit message or something.
+
+name: Setup images for AWX
+description: Builds new awx_devel image
+inputs:
+  github-token:
+    description: GitHub Token for registry access
+    required: true
+outputs:
+  ip:
+    description: The IP of the tools_awx_1 container
+    value: ${{ steps.data.outputs.ip }}
+  admin-token:
+    description: OAuth token for admin user
+    value: ${{ steps.data.outputs.admin_token }}
+runs:
+  using: composite
+  steps:
+    - name: Get python version from Makefile
+      shell: bash
+      run: echo py_version=`make PYTHON_VERSION` >> $GITHUB_ENV
+
+    - name: Log in to registry
+      shell: bash
+      run: |
+        echo "${{ inputs.github-token }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
+
+    - name: Pre-pull latest available devel image and build HEAD on top of it
+      shell: bash
+      run: |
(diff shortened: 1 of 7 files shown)


AlanCoding:
My belief is that checks too a little longer on this because changes to the Makefile triggered the image rebuilding. I don't recall exactly why this happens but I believe I've seen it before.

.github/actions/awx_ci_setup/action.yml
```diff
@@ -0,0 +1,35 @@
...
+  github-token:
+    description: GitHub Token for registry access
+    required: true
+outputs:
+  ip:
+    description: The IP of the tools_awx_1 container
+    value: ${{ steps.data.outputs.ip }}
+  admin-token:
+    description: OAuth token for admin user
+    value: ${{ steps.data.outputs.admin_token }}
```
relrod: These steps don't exist here, nix?

.github/workflows/ci.yml
```diff
@@ -40,16 +40,31 @@ jobs:
...
 
   dev-env:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
 
+      - name: Build awx_devel image for running checks
+        uses: ./.github/actions/awx_ci_setup
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
relrod: Would it make sense to have `run_awx_devel` include `awx_ci_setup?`

While we do want to run `awx_ci_setup` independently sometimes, I don't think we'd ever want to run `run_awx_devel` without depending on `awx_ci_setup`.

.github/workflows/e2e_test.yml
```diff
@@ -26,7 +26,6 @@ jobs:
         with:
           build-ui: true
           github-token: ${{ secrets.GITHUB_TOKEN }}
-          log-filename: e2e-${{ matrix.job }}.log
```
AlanCoding, line 29: @relrod This `log-filename` is an input to the other action (upload_awx_devel_logs), not this action. It seems to have not failed anything, but this seemed to be a bug.

I could, however, see the argument for having `run_awx_devel` call `upload_awx_devel_logs` in the same way that I'm now having it call the image build action. In that case, it would be an input, and we also would be more compact by never calling `upload_awx_devel_logs`

But that would require adding it as something like a "finalizer" step, which needs research.

AlanCoding: looks like it's not a thing https://github.com/actions/runner/issues/1478

relrod: Yeah that is what I really wanted to do originally, but there's no way when using a composite action. :(

PR #14425, opened 2023-09-08
[CI] Include a lint check for the Ansible tasks/roles/playbooks in AWX

ccamacho:
##### SUMMARY

This commit includes a GitHub actions lint check for the Ansible automation that resides in this repo.

The idea of this commit is to save some time in any e2e job that might fail because an issue that can be prevented with a lint check.

##### ISSUE TYPE
 - New or Enhanced Feature

##### COMPONENT NAME
 - Other (Ansible playbooks/roles/tasks)

##### AWX VERSION
NA

### DIFF
--- a/.ansible-lint.yml
+++ b/.ansible-lint.yml
@@ -1,0 +1,34 @@
+exclude_paths:
+  - .github
+  - .tox
+
+# Failed: 1348 failure(s), 97 warning(s) on 4230 files.
+skip_list:
+  - 'command-instead-of-module'
+  - 'command-instead-of-shell'
+  - 'deprecated-local-action'
+  - 'key-order[task]'
+  - 'jinja[invalid]'
+  - 'jinja[spacing]'
+  - 'schema[tasks]'
+  - 'name[missing]'
+  - 'name[play]'
+  - 'var-naming[no-reserved]'
+  - 'var-naming[pattern]'
+  - 'yaml[document-start]'
+  - 'yaml[empty-lines]'
+  - 'name[template]'
+  - 'name[casing]'
+  - 'risky-file-permissions'
+  - 'risky-shell-pipe'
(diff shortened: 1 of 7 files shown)


awx_collection/galaxy.yml
```diff
@@ -12,7 +12,10 @@
 ---
 authors:
   - AWX Project Contributors <awx-project@googlegroups.com>
-dependencies: {}
+dependencies:
+  community.docker: '*'
```
ccamacho: Do you always use latest, or a pin to an specific version?

.ansible-lint.yml
```diff
@@ -0,0 +1,34 @@
+exclude_paths:
+  - .github
+  - .tox
+
+# Failed: 1348 failure(s), 97 warning(s) on 4230 files.
```
ccamacho, line 5: All these should be addressed in additional PRs

.ansible-lint.yml
```diff
@@ -0,0 +1,34 @@
...
+  - 'var-naming[pattern]'
+  - 'yaml[document-start]'
+  - 'yaml[empty-lines]'
+  - 'name[template]'
+  - 'name[casing]'
+  - 'risky-file-permissions'
+  - 'risky-shell-pipe'
+  - 'galaxy[no-changelog]'
+  - 'galaxy[version-incorrect]'
+  - 'ignore-errors'
```
AlanCoding, line 26: For actual playbooks this is probably a good type of rule, but I'm guessing it's hitting this in the awx.awx collection tests. You can't test effectively without a lot of error scenarios that shouldn't happen if you're following best practices.

I'd like to know how many of these rules are skipped due to the collection tests, and for maintainability, I wonder if it would be better to have a separate list for those. I'm talking about `awx_collection/tests/integration`

PR #14443, opened 2023-09-13
Added contributor guide for docs

tvo318:
##### SUMMARY
Addresses doc issue https://github.com/ansible/awx/issues/14434

##### ISSUE TYPE
 - New doc section

##### COMPONENT NAME
 - Docs

##### AWX VERSION
n/a

PR #14447, opened 2023-09-14
Contributor guide and adjusted titles

oraNod:
##### SUMMARY
This PR follows up on https://github.com/ansible/awx/pull/14443 and includes the AWX Contributor's Guide in the RST docsite. Additionally this PR reorders and renames sections in the top-level index as well as titles in the index for each guide.

##### ISSUE TYPE

##### COMPONENT NAME
 - Docs

##### AWX VERSION
N/A Latest docs

##### ADDITIONAL INFORMATION

It is probably also worth deleting the `contributors` branch if it was created by accident. https://github.com/ansible/awx/tree/contributors

### DIFF
--- a/docs/docsite/rst/index.rst
+++ b/docs/docsite/rst/index.rst
@@ -5,42 +5,32 @@ Ansible AWX helps teams manage complex multi-tier deployments by adding control,
 
 .. toctree::
    :maxdepth: 2
-   :caption: AWX Quickstart
+   :caption: Get started
 
    quickstart/index
 
 .. toctree::
    :maxdepth: 2
-   :caption: User Guide
+   :caption: Community
+
+   contributor/index
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Users
 
    userguide/index
 
 .. toctree::
    :maxdepth: 2
-   :caption: AWX Administration
-
-   administration/index
-
-.. toctree::
-   :maxdepth: 2
-   :caption: AWX REST API
+   :caption: Developers
 
    rest_api/index
 
 .. toctree::
    :maxdepth: 2
-   :caption: Upgrades and Migrations
+   :caption: Administrators
 
+   administration/index
    upgrade_migration/index
-
-.. toctree::
-   :maxdepth: 2
-   :caption: Release Notes
-
    release_notes/index
-
-.. toctree::
-   :maxdepth: 2
-   :caption: Contributor's Guide
-
-   contributor/index
--- a/docs/docsite/rst/index.rst
+++ b/docs/docsite/rst/index.rst
@@ -38,3 +38,9 @@ Ansible AWX helps teams manage complex multi-tier deployments by adding control,
    :caption: Release Notes
 
    release_notes/index
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Contributor's Guide
+
+   contributor/index
--- a/docs/docsite/rst/index.rst
+++ b/docs/docsite/rst/index.rst
@@ -33,4 +33,9 @@ Ansible AWX helps teams manage complex multi-tier deployments by adding control,
 
    administration/index
    upgrade_migration/index
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Release notes
+
    release_notes/index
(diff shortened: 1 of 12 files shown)


oraNod:
@AlanCoding @thedoubl3j Would either of you be opposed to enabling PR previews in read the docs?

AlanCoding:
Where is this information coming from? Is it taken from this file?

https://github.com/ansible/awx/blob/devel/CONTRIBUTING.md

When we're migrated content, personally, my vote would be to try to delete the source content and point to the new location. Because duplicated content always goes out-of-date, because someone updates one of the copies and then both are in some kind of partially maintained state.

oraNod:
> Where is this information coming from? Is it taken from this file?
> 
> https://github.com/ansible/awx/blob/devel/CONTRIBUTING.md
> 
> When we're migrated content, personally, my vote would be to try to delete the source content and point to the new location. Because duplicated content always goes out-of-date, because someone updates one of the copies and then both are in some kind of partially maintained state.

Hey @AlanCoding Thanks for pointing that out. I think it is migrated from that particular MD file. Should we remove the `CONTRIBUTING.md` file itself or just the content and leave a pointer to the docs?

tvo318:
@AlanCoding, @oraNod - the content in this contributor's guide is specifically for docs. The https://github.com/ansible/awx/blob/devel/CONTRIBUTING.md was used as reference only and is specific to AWX code changes and development but the two are separate.

docs/docsite/rst/index.rst
```diff
@@ -5,36 +5,32 @@ Ansible AWX helps teams manage complex multi-tier deployments by adding control,
...
    :maxdepth: 2
-   :caption: AWX Quickstart
+   :caption: Get started
 
    quickstart/index
 
 .. toctree::
    :maxdepth: 2
-   :caption: User Guide
+   :caption: Community
```
tvo318, line 14: The caption should be Users

docs/docsite/rst/index.rst
```diff
@@ -5,36 +5,32 @@ Ansible AWX helps teams manage complex multi-tier deployments by adding control,
...
 
-   userguide/index
+   contributor/index
 
 .. toctree::
    :maxdepth: 2
-   :caption: AWX Administration
+   :caption: Users
```
tvo318, line 20: This caption should be Administrators

docs/docsite/rst/index.rst
```diff
@@ -5,36 +5,32 @@ Ansible AWX helps teams manage complex multi-tier deployments by adding control,
...
    :maxdepth: 2
-   :caption: AWX REST API
+   :caption: Developers
 
    rest_api/index
 
 .. toctree::
    :maxdepth: 2
-   :caption: Upgrades and Migrations
+   :caption: Administrators
```
tvo318, line 32: I am not understanding the additional administration/index here or the caption.

tvo318 (commented):
These changes look good. I'm sorry you had to re-open a new PR! Thanks for doing that and for adding the disclaimers.

docs/docsite/rst/index.rst
oraNod: I moved things around a little in the top-level index. We can adjust but each caption matches the corresponding index. In this case line 22 is `userguide/index` and the left-hand nav renders like this:

image

oraNod:
Hmm.. I toggled the PR by closing and then opening it to get the PR preview on read the docs but I don't see it yet.

@tvo318 I can revert some of the changes that I made to the captions and the left nav but I thought it might be a good idea to adjust them so that they match the persona-led approach that we've adopted on "docs.ansible.com". Here's a screengrab of the changes:

image

oraNod:
@tvo318 Also I kind of feel like this PR is mixing things a bit too much for my taste. I didn't want to reset your commit because I cherry-picked it on to my branch to avoid that issue with the previous PR we couldn't push to (also because you're on PTO). However if you think it's OK perhaps we can break this into two separate PRs, one for the docs contributor guide and one for the titles, nav changes, and short descriptions. How would that sound?

tvo318 (approved):
Let's go ahead and merge this and see what the team thinks! Thanks for doing this, @oraNod!

PR #14451, opened 2023-09-14
Move the continue-on-error marker

AlanCoding:
##### SUMMARY
I am trying to read the tea leaves of 

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions

accurately.

My last patch where I added this originally did not work. Artifacts are still not getting saved for failed checks.

##### ISSUE TYPE

##### COMPONENT NAME
 - API


relrod (approved):
What could go wrong?

PR #14462, opened 2023-09-19
Adding hosts bulk deletion feature

Avilir:
UMMARY

This PR present new capability to the API, deleting host from inventory as bulk in one API call instead of deleting them one by one

Replacing #14417  to remove merging / rebasing  issues

ISSUE TYPE

    New or Enhanced Feature

COMPONENT NAME

    API

ADDITIONAL INFORMATION

    With this change you could delete a bulk of hosts as following:

Post to /api/v2/bulk/host_delete

{
     "hosts": [1, 2, 3, 4]
}

The above will delete the 4 hosts with id's : 1, 2, 3, 4

### DIFF
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -2262,15 +2262,18 @@ class BulkHostDeleteSerializer(serializers.Serializer):
                 errors[inv.name] = "Hosts can only be deleted from manual inventories."
         if errors != {}:
             raise serializers.ValidationError({"inventories": errors})
-
+        attrs['inventories'] = inv_list
         return attrs
 
     def delete(self, validated_data):
         result = {"hosts": dict()}
-        changes = {'deleted_hosts': list()}
+        changes = {'deleted_hosts': dict()}
+        for inventory in validated_data['inventories']:
+            changes[inventory] = list()
+
         for host in validated_data['hosts_data']:
             result["hosts"][host["id"]] = f"The host {host['name']} was deleted"
-            changes['deleted_hosts'].append({"host_id": host["id"], "host_name": host["name"]})
+            changes['deleted_hosts'][host["inventory_id"]].append({"host_id": host["id"], "host_name": host["name"]})
 
         try:
             validated_data['host_qs'].delete()
@@ -2278,12 +2281,16 @@ class BulkHostDeleteSerializer(serializers.Serializer):
             raise serializers.ValidationError({"detail": _(f"cannot delete hosts, host deletion error {e}")})
 
         request = self.context.get('request', None)
-        ActivityStream.objects.create(
-            operation='delete',
-            object1='inventory',
-            changes=json.dumps(changes),
-            actor=request.user,
-        )
+
+        for inventory in validated_data['inventories']:
+            activity_entry = ActivityStream.objects.create(
+                operation='update',
+                object1='inventory',
+                changes=json.dumps(changes['deleted_hosts'][inventory]),
+                actor=request.user,
+            )
+            activity_entry.inventory.add(inventory)
+
         return result
 
 
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -2269,7 +2269,7 @@ class BulkHostDeleteSerializer(serializers.Serializer):
         result = {"hosts": dict()}
         changes = {'deleted_hosts': dict()}
         for inventory in validated_data['inventories']:
-            changes[inventory] = list()
+            changes['deleted_hosts'][inventory] = list()
 
         for host in validated_data['hosts_data']:
             result["hosts"][host["id"]] = f"The host {host['name']} was deleted"
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -2201,6 +2201,92 @@ class BulkHostCreateSerializer(serializers.Serializer):
         return return_data
 
 
+class BulkHostDeleteSerializer(serializers.Serializer):
+    hosts = serializers.ListField(
+        allow_empty=False,
+        max_length=100000,
+        write_only=True,
+        help_text=_('List of hosts ids to be deleted, e.g. [105, 130, 131, 200]'),
+    )
+
+    class Meta:
+        model = Host
+        fields = ('hosts',)
+
+    def validate(self, attrs):
+        request = self.context.get('request', None)
+        max_hosts = settings.BULK_HOST_MAX_DELETE
+        # Validating the number of hosts to be deleted
+        if len(attrs['hosts']) > max_hosts:
+            raise serializers.ValidationError(
+                {
+                    "ERROR": 'Number of hosts exceeds system setting BULK_HOST_MAX_DELETE',
+                    "BULK_HOST_MAX_DELETE": max_hosts,
+                    "Hosts_count": len(attrs['hosts']),
+                }
+            )
+
+        # Getting list of all host objects, filtered by the list of the hosts to delete
+        attrs['host_qs'] = Host.objects.get_queryset().filter(pk__in=attrs['hosts'])
+
+        # Converting the queryset data in a dict. to reduce the number of queries when
+        # manipulating the data
+        attrs['hosts_data'] = attrs['host_qs'].values()
+
+        if len(attrs['host_qs']) == 0:
+            error_hosts = {host: "Hosts do not exist or you lack permission to delete it" for host in attrs['hosts']}
+            raise serializers.ValidationError({'hosts': error_hosts})
+
+        if len(attrs['host_qs']) < len(attrs['hosts']):
+            hosts_exists = [host['id'] for host in attrs['hosts_data']]
+            failed_hosts = list(set(attrs['hosts']).difference(hosts_exists))
+            error_hosts = {host: "Hosts do not exist or you lack permission to delete it" for host in failed_hosts}
+            raise serializers.ValidationError({'hosts': error_hosts})
+
+        # Getting all inventories that the hosts can be in
+        inv_list = list(set([host['inventory_id'] for host in attrs['hosts_data']]))
+
+        # Checking that the user have permission to all inventories
+        errors = dict()
+        for inv in Inventory.objects.get_queryset().filter(pk__in=inv_list):
+            if request and not request.user.is_superuser:
+                if request.user not in inv.admin_role:
+                    errors[inv.name] = "Lack permissions to delete hosts from this inventory."
+        if errors != {}:
+            raise PermissionDenied({"inventories": errors})
+
+        # check the inventory type only if the user have permission to it.
+        errors = dict()
+        for inv in Inventory.objects.get_queryset().filter(pk__in=inv_list):
+            if inv.kind != '':
+                errors[inv.name] = "Hosts can only be deleted from manual inventories."
+        if errors != {}:
+            raise serializers.ValidationError({"inventories": errors})
+
+        return attrs
+
+    def delete(self, validated_data):
+        result = {"hosts": dict()}
+        changes = {'deleted_hosts': list()}
+        for host in validated_data['hosts_data']:
+            result["hosts"][host["id"]] = f"The host {host['name']} was deleted"
+            changes['deleted_hosts'].append({"host_id": host["id"], "host_name": host["name"]})
+
+        try:
+            validated_data['host_qs'].delete()
+        except Exception as e:
+            raise serializers.ValidationError({"detail": _(f"cannot delete hosts, host deletion error {e}")})
+
+        request = self.context.get('request', None)
+        ActivityStream.objects.create(
+            operation='delete',
+            object1='inventory',
+            changes=json.dumps(changes),
+            actor=request.user,
+        )
+        return result
+
+
 class GroupTreeSerializer(GroupSerializer):
     children = serializers.SerializerMethodField()
 
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -2227,7 +2227,7 @@ class BulkHostDeleteSerializer(serializers.Serializer):
             )
 
         # Getting list of all host objects, filtered by the list of the hosts to delete
-        attrs['host_qs'] = Host.objects.get_queryset().filter(pk__in=attrs['hosts'])
+        attrs['host_qs'] = Host.objects.get_queryset().filter(pk__in=attrs['hosts']).only('id', 'inventory_id', 'name')
 
         # Converting the queryset data in a dict. to reduce the number of queries when
         # manipulating the data
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -2278,13 +2278,12 @@ class BulkHostDeleteSerializer(serializers.Serializer):
             raise serializers.ValidationError({"detail": _(f"cannot delete hosts, host deletion error {e}")})
 
         request = self.context.get('request', None)
-        activity_entry = ActivityStream.objects.create(
-            operation='update',
+        ActivityStream.objects.create(
+            operation='delete',
             object1='inventory',
             changes=json.dumps(changes),
             actor=request.user,
         )
-        activity_entry.inventory.add(validated_data['inventory'])
         return result
 
 
--- a/awx/api/serializers.py
+++ b/awx/api/serializers.py
@@ -2278,12 +2278,13 @@ class BulkHostDeleteSerializer(serializers.Serializer):
             raise serializers.ValidationError({"detail": _(f"cannot delete hosts, host deletion error {e}")})
 
         request = self.context.get('request', None)
-        ActivityStream.objects.create(
-            operation='delete',
+        activity_entry = ActivityStream.objects.create(
+            operation='update',
             object1='inventory',
             changes=json.dumps(changes),
             actor=request.user,
         )
+        activity_entry.inventory.add(validated_data['inventory'])
         return result
 
 
--- a/awx/main/tests/functional/test_bulk.py
+++ b/awx/main/tests/functional/test_bulk.py
@@ -309,3 +309,139 @@ def test_bulk_job_set_all_prompt(job_template, organization, inventory, project,
     assert node[0].limit == 'kansas'
     assert node[0].skip_tags == 'foobar'
     assert node[0].job_tags == 'untagged'
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('num_hosts, num_queries', [(1, 70), (10, 70), (25, 120)])
+def test_bulk_host_delete_num_queries(organization, inventory, post, get, user, num_hosts, num_queries, django_assert_max_num_queries):
+    '''
+    If I am a...
+      org admin
+      inventory admin at org level
+      admin of a particular inventory
+      superuser
+
+    Bulk Host delete should take under a certain number of queries
+    '''
+    users_list = setup_admin_users_list(organization, inventory, user)
+    for u in users_list:
+        hosts = [{'name': str(uuid4())} for i in range(num_hosts)]
+        with django_assert_max_num_queries(num_queries):
+            bulk_host_create_response = post(reverse('api:bulk_host_create'), {'inventory': inventory.id, 'hosts': hosts}, u, expect=201).data
+            assert len(bulk_host_create_response['hosts']) == len(hosts), f"unexpected number of hosts created for user {u}"
+            hosts_ids_created = get_inventory_hosts(get, inventory.id, u)
+            bulk_host_delete_response = post(reverse('api:bulk_host_delete'), {'hosts': hosts_ids_created}, u, expect=201).data
+            assert len(bulk_host_delete_response['hosts'].keys()) == len(hosts), f"unexpected number of hosts deleted for user {u}"
+
+
+@pytest.mark.django_db
+def test_bulk_host_delete_rbac(organization, inventory, post, get, user):
+    '''
+    If I am a...
+      org admin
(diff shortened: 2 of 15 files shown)


kdelee:
seeing
```
TASK [bulk_host_delete : Get hosts ids from inventory] *************************
task path: /home/runner/.ansible/collections/ansible_collections/awx/awx/tests/output/.tmp/integration/bulk_host_delete-9yauvxl1-ÅÑŚÌβŁÈ/tests/integration/targets/bulk_host_delete/tasks/main.yml:44
fatal: [testhost]: FAILED! => {
    "msg": "The task includes an option with an undefined variable. The error was: 'controller_host' is undefined. 'controller_host' is undefined\n\nThe error appears to be in '/home/runner/.ansible/collections/ansible_collections/awx/awx/tests/output/.tmp/integration/bulk_host_delete-9yauvxl1-ÅÑŚÌβŁÈ/tests/integration/targets/bulk_host_delete/tasks/main.yml': line 44, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: \"Get hosts ids from inventory\"\n  ^ here\n"
}
```
in the PR check

awx_collection/tests/integration/targets/bulk_host_delete/tasks/main.yml
```diff
@@ -0,0 +1,74 @@
...
+        enabled: "false"
+    inventory: "{{ bulk_inv_name }}"
+  register: "result"
+
+- assert:
+    that:
+      - "result is not failed"
+
+- name: "Get hosts ids from inventory"
+  uri:
```
kdelee: instead of the URI module, maybe you could do a set fact task and use the controller API lookup plugin

see https://github.com/ansible/awx/blob/devel/awx_collection/tests/integration/targets/lookup_api_plugin/tasks/main.yml#L18-L24 for how they use it in these tests and https://docs.ansible.com/ansible/latest/collections/awx/awx/controller_api_lookup.html for the docs

kdelee: that way you just inherit whatever way it is they are setting all the host/auth stuff for the target awx host in the tests

awx/main/tests/functional/test_bulk.py
```diff
@@ -309,3 +309,95 @@ def test_bulk_job_set_all_prompt(job_template, organization, inventory, project,
...
+
+
+@pytest.mark.django_db
+def test_bulk_host_delete_rbac(organization, inventory, post, get, user):
+    '''
+    If I am a...
+      org admin
+      inventory admin at org level
+      admin of a particular invenotry
+          ... I can bulk add hosts
```
kdelee:
```suggestion
          ... I can bulk delete hosts
```

awx/main/tests/functional/test_bulk.py
```diff
@@ -309,3 +309,95 @@ def test_bulk_job_set_all_prompt(job_template, organization, inventory, project,
...
+        bulk_host_create_response = post(
+            reverse('api:bulk_host_create'), {'inventory': inventory.id, 'hosts': [{'name': f'foobar2-{indx}'}]}, create_u, expect=201
+        ).data
+        print(bulk_host_create_response)
+        assert bulk_host_create_response['hosts'][0]['name'] == f'foobar2-{indx}'
+        hosts_ids_created = get_inventory_hosts(get, inventory.id, create_u)
+        print(f"Try to delete {hosts_ids_created}")
+        for delete_u in users_list:
+            bulk_host_delete_response = post(reverse('api:bulk_host_delete'), {'hosts': hosts_ids_created}, delete_u, expect=400).data
+            assert bulk_host_delete_response['ERRORS'] == ['Cannot delete host(s) from inventory']
```
kdelee: Either in this test, or in a new one, I'd like confirmation that if the list of hosts contains a mix of hosts the user DOES have permission to delete (e.g. belong to an inventory they have admin of) and hosts the user DOES NOT have permission to delete (e.g. belong to an inventory they DO NOT have admin of, they get an error and NO hosts are deleted

awx/main/tests/functional/test_bulk.py
```diff
@@ -309,3 +309,95 @@ def test_bulk_job_set_all_prompt(job_template, organization, inventory, project,
...
+        organization.member_role.members.add(u)
+    inventory.use_role.members.add(use_inv_member)
+    organization.auditor_role.members.add(auditor)
+    return [auditor, member, use_inv_member]
+
+
+def get_inventory_hosts(get, inv_id, use_user):
+    data = get(reverse('api:inventory_hosts_list', kwargs={'pk': inv_id}), use_user, expect=200).data
+    results = [host['id'] for host in data['results']]
+    return results
```
kdelee, line 447: Also would like to see a test that measures the # of queries and sets some bounds so we can get test failure if that changes dramatically in the future.

See https://github.com/ansible/awx/blob/devel/awx/main/tests/functional/test_bulk.py#L14

awx/api/serializers.py
```diff
@@ -2201,6 +2201,89 @@ def create(self, validated_data):
         return return_data
 
 
+class BulkHostDeleteSerializer(serializers.Serializer):
+    hosts = serializers.ListField(
+        allow_empty=False,
+        max_length=250,
```
```suggestion
        max_length=100000,
```
so I think in the BULK_HOST_CREATE I hard coded this large number because you can't easily access the settings at this point, but we want the setting to be the real limit.

awx/api/serializers.py
```diff
@@ -2201,6 +2201,89 @@ def create(self, validated_data):
...
+        help_text=_('List of hosts ids to be deleted, e.g. [105, 130, 131, 200]'),
+    )
+
+    class Meta:
+        model = Host
+        fields = 'hosts'
+        read_only_fields = ()
+
+    def validate(self, attrs):
+        request = self.context.get('request', None)
```
kdelee, line 2217:
```suggestion
        request = self.context.get('request', None)
        max_hosts = settings.BULK_HOST_MAX_DELETE
```
then uses `max_hosts` later on in function. Each access of `settings` is expensive/can fire queries to DB

awx/api/serializers.py
```diff
@@ -2201,6 +2201,89 @@ def create(self, validated_data):
...
+            raise serializers.ValidationError(
+                {
+                    "ERROR": 'Number of hosts exceeds system setting BULK_HOST_MAX_DELETE',
+                    "BULK_HOST_MAX_DELETE": settings.BULK_HOST_MAX_DELETE,
+                    "Hosts_count": len(attrs['hosts']),
+                }
+            )
+
+        # Getting list of all host objects, filtered by the list of the hosts to delete
+        attrs['in_inv'] = Host.objects.get_queryset().filter(pk__in=attrs['hosts'])
```
kdelee: This is a nitpick, but this key `in_inv` is not super clear to me.
Maybe better name would be `host_qs` becauses it is a host queryset

awx/api/serializers.py
```diff
@@ -2201,6 +2201,89 @@ def create(self, validated_data):
...
+
+        # Converting the queryset data in a dict. to reduce the number of queries when
+        # manipulating the data
+        hosts_data = attrs['in_inv'].values()
```
kdelee: you later call `values` again in `delete` perhaps instead you could do
`attrs['host_data'] = attrs['in_inv'].values()` and use it again later in `delete` method

awx/api/serializers.py
```diff
@@ -2201,6 +2201,89 @@ def create(self, validated_data):
...
+
+        if len(attrs['in_inv']) == 0:
+            raise serializers.ValidationError({"ERROR": "There are no hosts to delete"})
```
```suggestion
            raise serializers.ValidationError({"ERROR": f"Hosts {attrs['hosts']} do not exist or you lack permission to delete them"})
```

awx/api/serializers.py
```diff
@@ -2201,6 +2201,89 @@ def create(self, validated_data):
...
+
+        if len(attrs['in_inv']) < len(attrs['hosts']):
+            hosts_exists = [host['id'] for host in hosts_data]
+            raise serializers.ValidationError({"ERROR": "Not all hosts are exists", "hosts": list(set(attrs['hosts']).difference(hosts_exists))})
```
```suggestion
            raise serializers.ValidationError({"ERROR": "f"Hosts do not exist or you lack permission to delete them"", "hosts": list(set(attrs['hosts']).difference(hosts_exists))})
```

awx/api/serializers.py
```diff
@@ -2201,6 +2201,89 @@ def create(self, validated_data):
...
+            if inv.kind != '':
+                errors["Inventory"][inv.id] = {
+                    "Inventory_name": inv.name,
+                    "reason": "Hosts can only be deleted from manual inventories (not smart or constructed types).",
+                }
+            if request and not request.user.is_superuser:
+                if request.user not in inv.admin_role:
+                    errors["Inventory"][inv.id] = {
+                        "Inventory_name": inv.name,
+                        "reason": "Lack permissions to delete hosts from.",
```
```suggestion
                        "reason": "Lack permissions to delete hosts in this inventory",
```

kdelee (commented):
still reviewing but here are some comments

awx/api/serializers.py
```diff
@@ -2201,6 +2201,89 @@ def create(self, validated_data):
...
+        # Getting all inventories that the hosts can be in
+        inv_list = list(set([host['inventory_id'] for host in hosts_data]))
+
+        errors = {"ERRORS": "Cannot delete host(s) from inventory", "Inventory": dict()}
+        for inv in Inventory.objects.get_queryset().filter(pk__in=inv_list):
+            if inv.kind != '':
+                errors["Inventory"][inv.id] = {
+                    "Inventory_name": inv.name,
+                    "reason": "Hosts can only be deleted from manual inventories (not smart or constructed types).",
+                }
```
kdelee: we should check this after we check permissions, otherwise we are letting users check existence/type of inventories before we know they have permission to them

awx/api/serializers.py
```diff
@@ -2201,6 +2201,89 @@ def create(self, validated_data):
...
+            if request and not request.user.is_superuser:
+                if request.user not in inv.admin_role:
```
kdelee, line 2253: Take a look at https://github.com/ansible/awx/blob/87f12a569887e6ab406304941e2ae4b9884bab18/awx/api/serializers.py#L4873


This is more along lines of what we need to do than what we do in bulk host create, where we only need to check 1 inventory.

Don't necessarily need to take whole function since we are only dealing with inventories.
Example:
```              
allowed_invs =set(Inventory.accessible_objects(request.user, 'admin_role').filter(pk__in=[1,4], kind='').values_list('id', flat=True))
not_allowed_invs = set(inv_list) - allowed_invs
```

awx/api/serializers.py
```diff
@@ -2201,6 +2201,89 @@ def create(self, validated_data):
...
+                    }
+        if errors["Inventory"] != {}:
+            raise serializers.ValidationError(errors)
+
+        return attrs
+
+    def delete(self, validated_data):
+        result = {"hosts": dict()}
+        changes = {'deleted_hosts': list()}
+        for host in validated_data['in_inv'].values():
```
kdelee: if you save that `attrs['host_data']` could just that here

kdelee (commented):
couple more comments

awx_collection/tests/integration/targets/bulk_host_delete/tasks/main.yml
Avilir: done

awx/main/tests/functional/test_bulk.py
Avilir: done.

awx/main/tests/functional/test_bulk.py
Avilir: will add a new test for that

awx/main/tests/functional/test_bulk.py
Avilir: There is this kind of test, see https://github.com/Avilir/awx/blob/hbd/awx/main/tests/functional/test_bulk.py#L316

awx/api/serializers.py
Avilir: done

awx/api/serializers.py
Avilir: done

awx/api/serializers.py
Avilir: done

awx/api/serializers.py
Avilir: done

awx/api/serializers.py
Avilir: done

awx/api/serializers.py
Avilir: done

awx/api/serializers.py
Avilir: done

awx/api/serializers.py
Avilir: done

awx/api/serializers.py
Avilir: Since I need to loop over a list of inventories to check permission and kind of inventory, I think that this method is good.

awx/api/serializers.py
Avilir: done

awx/main/tests/functional/test_bulk.py
```diff
@@ -309,3 +309,139 @@ def test_bulk_job_set_all_prompt(job_template, organization, inventory, project,
     assert node[0].limit == 'kansas'
     assert node[0].skip_tags == 'foobar'
     assert node[0].job_tags == 'untagged'
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('num_hosts, num_queries', [(1, 70), (10, 70)])
```
kdelee: what if we add one more round of 100 hosts?

docs/bulk_api.md
```diff
@@ -99,3 +100,20 @@ Following is an example of a post request at the /api/v2/bulk/host_create:
...
+
+
+    {
+        "hosts": [3, 4, 5, 6, 7 ,8, 9, 10]
+    }
+
+
+The above will delete 8 hosts from the inventory.
+
+The maximum number of hosts allowed to be deleted is controlled by the setting `BULK_HOST_MAX_DELETE`. The default is 250 hosts. Additionally, nginx limits the maximum payload size, which is very likely when posting a large number of hosts in one request with variable data associated with them. The maximum payload size is 1MB unless overridden in your nginx config.
```
kdelee, line 119: :tada:

kdelee (approved):

awx/api/serializers.py
```diff
@@ -2201,6 +2201,92 @@ def create(self, validated_data):
...
+        allow_empty=False,
+        max_length=100000,
+        write_only=True,
+        help_text=_('List of hosts ids to be deleted, e.g. [105, 130, 131, 200]'),
+    )
+
+    class Meta:
+        model = Host
+        fields = 'hosts'
+        read_only_fields = ()
```
AlanCoding: the `fields` should normally be a tuple, and my concern would be that this gets parsed as `['h', 'o', 's', 't', 's']`. The observable we care about is the OPTIONS data that comes from the relevant endpoint. I also think it best to delete an empty `read_only_fields`

awx/api/serializers.py
```diff
@@ -2201,6 +2201,92 @@ def create(self, validated_data):
...
+        # manipulating the data
+        attrs['hosts_data'] = attrs['host_qs'].values()
+
+        if len(attrs['host_qs']) == 0:
+            raise serializers.ValidationError({"ERROR": f"Hosts {attrs['hosts']} do not exist or you lack permission to delete them"})
+
+        if len(attrs['host_qs']) < len(attrs['hosts']):
+            hosts_exists = [host['id'] for host in attrs['hosts_data']]
+            raise serializers.ValidationError(
+                {"ERROR": "Hosts do not exist or you lack permission to delete them", "hosts": list(set(attrs['hosts']).difference(hosts_exists))}
```
AlanCoding: It's standard to give keys in a 400 response that correspond to the form fields with errors, so it would be perfectly correct to use the "hosts" key. However, the "ERROR" field looks non-standard. I would go with a single message, and either pass that as a single string argument to `ValidationError`, or use a dict with only the "hosts" key where the value is the formatted message.

awx/api/serializers.py
```diff
@@ -2201,6 +2201,92 @@ def create(self, validated_data):
...
+        # Getting all inventories that the hosts can be in
+        inv_list = list(set([host['inventory_id'] for host in attrs['hosts_data']]))
+
+        errors = {"ERRORS": "Cannot delete host(s) from inventory", "Inventory": dict()}
+        for inv in Inventory.objects.get_queryset().filter(pk__in=inv_list):
+            if request and not request.user.is_superuser:
+                if request.user not in inv.admin_role:
+                    errors["Inventory"][inv.id] = {
+                        "Inventory_name": inv.name,
+                        "reason": "Lack permissions to delete hosts in this inventory.",
```
AlanCoding: This should be a permission error, not a `ValidationError`, a 403 code instead of a 400 code.

awx/api/serializers.py
```diff
@@ -2201,6 +2201,92 @@ def create(self, validated_data):
...
+                    }
+            # check the inventory type only if the user have permission to it.
+            if inv.kind != '' and inv.id not in errors["Inventory"]:
+                errors["Inventory"][inv.id] = {
+                    "Inventory_name": inv.name,
+                    "reason": "Hosts can only be deleted from manual inventories (not smart or constructed types).",
```
AlanCoding: If you want to be able to stack errors (I don't feel it's really necessary in this case), someone could give >100 hosts where some are from constructed or smart inventories. In that case, the problematic input field is still `hosts`. The REST framework allows giving lists for multiple errors a field, so the errors dict could be `{"hosts": ["You gave more than 100 hosts", "Hosts can only be deleted from manual inventories"]}`

Avilir: fixed.

awx/api/serializers.py
Avilir: fixed.

awx/api/serializers.py
Avilir: fixed.

awx/api/serializers.py
Avilir: fixed.

awx/main/tests/functional/test_bulk.py
Avilir: done.

kdelee (approved):

jainnikhil30 (approved):

awx/api/serializers.py
```diff
@@ -2201,6 +2201,92 @@ def create(self, validated_data):
...
+        except Exception as e:
+            raise serializers.ValidationError({"detail": _(f"cannot delete hosts, host deletion error {e}")})
+
+        request = self.context.get('request', None)
+        ActivityStream.objects.create(
+            operation='delete',
+            object1='host',
+            changes=json.dumps(changes),
+            actor=request.user,
+        )
```
AlanCoding: Let's associate the inventory for tracking purposes, like the host creation serializer

```python
        activity_entry = ActivityStream.objects.create(
            operation='update',
            object1='inventory',
            changes=json.dumps(changes),
            actor=request.user,
        )
        activity_entry.inventory.add(validated_data['inventory'])
```

Avilir: done

AlanCoding: I don't see this in the current diff, maybe the commit got lost. My request is to add something like the line `activity_entry.inventory.add(validated_data['inventory'])`. I don't think it will be any different from that example.

Avilir: yep, the commit was lost, fixed it now.

Avilir: @AlanCoding , I revert the commit since the deletion is not related to a specific inventory, there is no `inventory` key to point trhe activity_entry to.

awx/api/serializers.py
```diff
@@ -2201,6 +2201,92 @@ def create(self, validated_data):
...
+            raise serializers.ValidationError(
+                {
+                    "ERROR": 'Number of hosts exceeds system setting BULK_HOST_MAX_DELETE',
+                    "BULK_HOST_MAX_DELETE": max_hosts,
+                    "Hosts_count": len(attrs['hosts']),
+                }
+            )
+
+        # Getting list of all host objects, filtered by the list of the hosts to delete
+        attrs['host_qs'] = Host.objects.get_queryset().filter(pk__in=attrs['hosts'])
```
chrismeyersfsu: Consider using .only(id, inventory_id, name) 
I’m worried about hosts with large host vars using a lot of memory.

Avilir: done.

awx/api/serializers.py
AlanCoding: You already have some existing loops `for inv in Inventory.objects.get_queryset().filter(pk__in=inv_list):`, and the data in `validated_data['hosts_data']` already has a `inventory_id` field which exists in-memory by this point. No additional trip to the database is needed to get this. Yes it's true that we would need to list the subset of `changes` specific to that inventory which requires some additional processing. But I don't think this is a good idea to merge without the audit trail in place.

Avilir: OK, i fixed it.

chrismeyersfsu (approved):

AlanCoding (approved):

jbradberry (approved):

PR #14463, opened 2023-09-19
Fix bug, None was used instead of empty for DB outage

AlanCoding:
##### SUMMARY
It looks like something changed with the Django upgrade such that they do more referencing of settings in the startup order (probably because Django settings are canonically cached now)

This resulted in a list-valued setting with a default of `[]` returning `None` due to a surprising oversight in our error-catching wrapper.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/conf/settings.py
+++ b/awx/conf/settings.py
@@ -418,6 +418,10 @@ class SettingsWrapper(UserSettingsHolder):
         """Get value while accepting the in-memory cache if key is available"""
         with _ctit_db_wrapper(trans_safe=True):
             return self._get_local(name)
+        # If the last line did not return, that means we hit a database error
+        # in that case, we should not have a local cache value
+        # thus, return empty as a signal to use the default
+        return empty
 
     def __getattr__(self, name):
         value = empty
--- a/awx/conf/tests/unit/test_settings.py
+++ b/awx/conf/tests/unit/test_settings.py
@@ -13,6 +13,7 @@ from unittest import mock
 from django.conf import LazySettings
 from django.core.cache.backends.locmem import LocMemCache
 from django.core.exceptions import ImproperlyConfigured
+from django.db.utils import Error as DBError
 from django.utils.translation import gettext_lazy as _
 import pytest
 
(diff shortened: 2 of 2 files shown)


TheRealHaoLiu (approved):

TheRealHaoLiu:
thanks @AlanCoding for putting up this PR, this problem cause too much noises in the log when we have db connection problem and make things hard to debug

AlanCoding:
Getting the traceback:

```
tools_awx_1 | 2023-09-19 18:17:57,609 WARNING  [-] awx.conf.settings Database settings are not available, using defaults. error: connection is bad: Name or service not known
tools_awx_1 |   File "/usr/local/bin/awx-manage", line 9, in <module>
tools_awx_1 |     load_entry_point('awx', 'console_scripts', 'awx-manage')()
tools_awx_1 |   File "/awx_devel/awx/__init__.py", line 200, in manage
tools_awx_1 |     execute_from_command_line(sys.argv)
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
tools_awx_1 |     utility.execute()
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/management/__init__.py", line 436, in execute
tools_awx_1 |     self.fetch_command(subcommand).run_from_argv(self.argv)
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/management/base.py", line 412, in run_from_argv
tools_awx_1 |     self.execute(*args, **cmd_options)
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/management/base.py", line 453, in execute
tools_awx_1 |     self.check()
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/management/base.py", line 485, in check
tools_awx_1 |     all_issues = checks.run_checks(
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/checks/registry.py", line 88, in run_checks
tools_awx_1 |     new_errors = check(app_configs=app_configs, databases=databases)
tools_awx_1 |   File "/var/lib/awx/venv/awx/lib64/python3.9/site-packages/django/core/checks/compatibility/django_4_0.py", line 9, in check_csrf_trusted_origins
tools_awx_1 |     for origin in settings.CSRF_TRUSTED_ORIGINS:
tools_awx_1 | TypeError: 'NoneType' object is not iterable
tools_awx_1 | 2023-09-19 18:17:58,052 WARN exited: awx-cache-clear (exit status 1; not expected)
```

However, writing the test, I have discovered a flaw in this method. We define the default in 2 places:

https://github.com/ansible/awx/blob/7b846e1e49874c64aac979619d2773c3bc3b384e/awx/settings/defaults.py#L164

https://github.com/ansible/awx/blob/7b846e1e49874c64aac979619d2773c3bc3b384e/awx/main/conf.py#L99

Now in this case, they are both the same. We also do this somewhat systematically, where we define the default for a setting in both places. I don't know exactly why, but I'm sure that practice started due to some bug in the settings code. This doesn't matter a _huge_ deal for the current fix, except to note that it won't _follow the default value precedence_ correctly. The fix here will get us the default value from `awx.settings.defaults` module, but will not get us the default declared in the settings registry.

PR #14433, opened 2023-09-12
AAP-15927 Use ATTACH PARTITION to avoid exclusive table lock for events

AlanCoding:
##### SUMMARY
This parrots the example from postgresql docs

https://www.postgresql.org/docs/current/ddl-partitioning.html

Specifically, see the example:

```sql
CREATE TABLE measurement_y2008m02
  (LIKE measurement INCLUDING DEFAULTS INCLUDING CONSTRAINTS)
  TABLESPACE fasttablespace;

ALTER TABLE measurement_y2008m02 ADD CONSTRAINT y2008m02
   CHECK ( logdate >= DATE '2008-02-01' AND logdate < DATE '2008-03-01' );

\copy measurement_y2008m02 from 'measurement_y2008m02'
-- possibly some other data preparation work

ALTER TABLE measurement ATTACH PARTITION measurement_y2008m02
    FOR VALUES FROM ('2008-02-01') TO ('2008-03-01' );
```

This just replaces our current SQL for creating a partition with this example, editing the details to use the tables names and timestamps that we use.

Right now this is working, in that new jobs will create a partition and that goes fine.

I did add a new bit of code to this to pre-check for existence of the partition. Letting it error in postgres created a lot of log noise that we've found very distracting in the past so I wanted to go ahead and take care of this, particularly since I'm running a new version of our SQL code. I do think we should remove our existing error handling since I added this, so expect a new commit to do that.

##### ISSUE TYPE
 - Bug, Docs Fix or other nominal change

##### COMPONENT NAME
 - API

##### ADDITIONAL INFORMATION
I'm not yet sure if we need this part:

```python
                    f'ALTER TABLE {tblname}_{partition_label} ADD CONSTRAINT y{partition_label} '
                    f'CHECK ( job_created >= TIMESTAMP \'{start_timestamp}\' AND job_created < TIMESTAMP \'{end_timestamp}\' ); '
```

Because historically, the old method did not create this constraint. So I think we could just delete these lines, but including it is more compliant to the docs so I wanted to start with this.

PR #14480, opened 2023-09-25
Add null value handling in create_partition

AlanCoding:
##### SUMMARY
Fix on top of https://github.com/ansible/awx/pull/14433

I believe this may be version-dependent, but I thought I should adopt some of the just-in-case checking I saw in this code:

https://github.com/ansible/awx/blob/9ed527eb26820caa2a8a18e93cf396e345f3c679/awx/sso/migrations/0003_convert_saml_string_to_list.py#L10-L13

And in any case, we rarely directly process the result from `cursor.execute()`, so I don't want this very critical code to be the exception.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/utils/common.py
+++ b/awx/main/utils/common.py
@@ -1164,7 +1164,8 @@ def create_partition(tblname, start=None):
     try:
         with transaction.atomic():
             with connection.cursor() as cursor:
-                row = cursor.execute(f"SELECT EXISTS (SELECT FROM information_schema.tables WHERE table_name = '{tblname}_{partition_label}');").fetchone()
+                cursor.execute(f"SELECT EXISTS (SELECT FROM information_schema.tables WHERE table_name = '{tblname}_{partition_label}');")
+                row = cursor.fetchone()
                 if row is not None:
                     for val in row:  # should only have 1
                         if val is True:
--- a/awx/main/utils/common.py
+++ b/awx/main/utils/common.py
@@ -1164,8 +1164,8 @@ def create_partition(tblname, start=None):
     try:
         with transaction.atomic():
             with connection.cursor() as cursor:
-                r_tuples = cursor.execute(f"SELECT EXISTS (SELECT FROM information_schema.tables WHERE table_name = '{tblname}_{partition_label}');")
-                for row in r_tuples:  # should only have 1
+                row = cursor.execute(f"SELECT EXISTS (SELECT FROM information_schema.tables WHERE table_name = '{tblname}_{partition_label}');").fetchone()
+                if row is not None:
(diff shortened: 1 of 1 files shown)


awx/main/utils/common.py
```diff
@@ -1164,8 +1164,8 @@ def create_partition(tblname, start=None):
     try:
         with transaction.atomic():
             with connection.cursor() as cursor:
-                r_tuples = cursor.execute(f"SELECT EXISTS (SELECT FROM information_schema.tables WHERE table_name = '{tblname}_{partition_label}');")
-                for row in r_tuples:  # should only have 1
+                row = cursor.execute(f"SELECT EXISTS (SELECT FROM information_schema.tables WHERE table_name = '{tblname}_{partition_label}');").fetchone()
+                if row is not None:
```
relrod, line 1169: I think the missing `fetchone()` was the real issue here. The query should always return one row that has true or false.

So I think we can nix this `None` check, but I'm indifferent on it.

relrod (approved):

awx/main/utils/common.py
AlanCoding: If it's `None` we might already be on a collision course, so in that case my preference would be to move onto the creation of the partition and maybe that will give a nicer error.

relrod: If it's `None`, I believe someone is running a _weirdly_ modified version of postgres :sweat_smile:

AlanCoding:
It seems that I have made the mistake outlined here:

https://stackoverflow.com/questions/30421466/why-would-mysql-execute-return-none

What seems to be the root of the problem is that old Django versions returned `None` from `cursor.execute()`, but newer versions of Django _probably_ now give `cursor` as the return value. Even for the current documentation:

https://docs.djangoproject.com/en/4.2/topics/db/sql/

They don't use the returned thing. They use:

```python
def my_custom_sql(self):
        cursor.execute("UPDATE bar SET foo = 1 WHERE baz = %s", [self.baz])
        cursor.execute("SELECT foo FROM bar WHERE baz = %s", [self.baz])
        row = cursor.fetchone()

    return row
```

This conflict in coding pattern seems to be the real root issue, and I see no reason to preference my prior approach as opposed to the docs. So I'll push a commit for that.

relrod:
That pattern looks more reasonable. Thinking more about it, I guess if `execute()` returns `None`, then tagging `fetchone()` on the end wouldn't work any better than trying to iterate over `None` did :laughing:

But I maintain that for a `SELECT EXISTS` you should _always_ get a row back.

In any case my :heavy_check_mark: still applies.

PR #14488, opened 2023-09-28
Execution environment image should not be required

willthames:
##### SUMMARY
Remove requirement from image parameter as it's not always needed (e.g. when state is absent or updating an existing execution environment)

##### ISSUE TYPE

##### COMPONENT NAME
 - Collection
##### AWX VERSION

```
awx: 22.1.1.dev353+g6c4c8df3f8
`

### DIFF
--- a/awx_collection/plugins/modules/execution_environment.py
+++ b/awx_collection/plugins/modules/execution_environment.py
@@ -33,7 +33,6 @@ options:
     image:
       description:
         - The fully qualified url of the container image.
-      required: True
       type: str
     description:
       description:
@@ -79,7 +78,7 @@ def main():
     argument_spec = dict(
         name=dict(required=True),
         new_name=dict(),
-        image=dict(required=True),
+        image=dict(),
         description=dict(),
         organization=dict(),
         credential=dict(),

### CODE REVIEW

File: awx_collection/plugins/modules/execution_environment.py [hunk 1]
```diff
@@ -79,17 +78,21 @@ def main():
...
         organization=dict(),
         credential=dict(),
         state=dict(choices=['present', 'absent', 'exists'], default='present'),
         # NOTE: Default for pull differs from API (which is blank by default)
         pull=dict(choices=['always', 'missing', 'never'], default='missing'),
     )
+    required_if = [
+        ['state', 'present', ['image']],
+        ['state', 'exists', ['image']],
+    ]
```
AlanCoding: I would prefer to not even have this condition. Someone could modify a different field of an existing item. If it is missing (on creation) then the API error should suffice.

willthames:
@AlanCoding I've updated the PR based on your feedback - image is now just not required.

AlanCoding:
Thanks. This looks good to me.

Could you do the DCO signoff before we merge?

https://github.com/ansible/awx/blob/159dd62d84df3ac14d274571724eee0324c2ca61/DCO_1_1.md?plain=1#L4

AlanCoding (approved):

PR #14528, opened 2023-10-03
collection bugfix booleans default value to match api

sean-m-sullivan:
##### SUMMARY
collection bugfix booleans default value to match api
fixing issue
#14461

Also referencing
#14493

Also removes the tests folder from the final distribution tarball, I've seen a few instances where including it messes with sanity tests and others when collections like this are used as dependencies.

##### ISSUE TYPE

##### COMPONENT NAME

 - Collection

##### AWX VERSION
```
23.2.0
```

### DIFF
--- a/awx_collection/plugins/modules/host.py
+++ b/awx_collection/plugins/modules/host.py
@@ -43,6 +43,7 @@ options:
       description:
         - If the host should be enabled.
       type: bool
+      default: true
     variables:
       description:
         - Variables to use for the host.
@@ -81,7 +82,7 @@ def main():
         new_name=dict(),
         description=dict(),
         inventory=dict(required=True),
-        enabled=dict(type='bool'),
+        enabled=dict(type='bool', default=True),
         variables=dict(type='dict'),
         state=dict(choices=['present', 'absent', 'exists'], default='present'),
     )
--- a/awx_collection/galaxy.yml
+++ b/awx_collection/galaxy.yml
@@ -36,3 +36,4 @@ build_ignore:
   - galaxy.yml.j2
   - template_galaxy.yml
   - '*.tar.gz'
+  - tests
--- a/awx_collection/plugins/modules/instance.py
+++ b/awx_collection/plugins/modules/instance.py
@@ -37,11 +37,13 @@ options:
         - If true, the instance will be enabled and used.
       required: False
       type: bool
+      default: true
     managed_by_policy:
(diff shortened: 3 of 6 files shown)


sean-m-sullivan:
I was not able to get a k8's up to make good tests for the instances, if you can suggest, or I can address it here tomorrow.

awx_collection/plugins/modules/host.py
```diff
@@ -43,6 +43,7 @@
       description:
         - If the host should be enabled.
       type: bool
+      default: true
```
AlanCoding, line 46: This is going to set the `enabled` flag to true for any existing host where the option isn't specified.

sean-m-sullivan: you are right..........................................................................................................................................................................................................................................................................................................................................................................................................................................................

AlanCoding:
I think we both have felt the same frustration with the sea-saw, whack-a-mole that comes of this. Partially just from the anger I feel from multiple experiences of this, I don't want to leave the problem un-fixed, and I don't want to let structural code issues block a fix for such a simple problem.

Your point here is well-taken. Today's fix should be more targeted, and avoid wading into the insanity of types of null. I've incorporated the spirit of your approach into https://github.com/ansible/awx/pull/14493

Instead of marking the field deafults (which would mean maintaining more schema), that uses the module `argument_spec` data which already exists to know if something is a boolean field. It's still not totally correct, as I think `diff_mode` in schedules and workflow nodes may be a valid tertiary field in the API. However, I think that field serves a purpose that no one should want or care about, so I'd gladly accept this bug fix as a tradeoff.

AlanCoding (changes requested):
I'd rather not use this overall approach

sean-m-sullivan:
I agree, this is just another can of worms in the long line of whack a mole, I don't think there is a great answer. Was thinking maybe we get the argument spec from the AAP API like I was thinking for enforced, but it still has that issue of if the user doesn't pass the field, it would default it. 

Right now if anything is passed it honors it, and the API will skip that field, but I think the answer to this goes back to the enforced defaults idea. 

I am trying to think of how we can make this better, and what criteria we would want on behavior of specifically host/shcedule state on
1. creation when no bool var is passed
2. Update/change not chaning where no var is passed for that bool field.

And focus on that behaving as expected, and how best to handle it.

Issue #13241, opened 2022-11-28
"make black" target doesn't actually fix something


thedoubl3j:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.

### Bug Summary

the pre commit check for black checks to see if anything needs to be reformatted then says to make 

### AWX version

awx: 21.5.1.dev434+g3ca14c43bb

### Select the relevant components

- [ ] UI
- [ ] API
- [X] Docs
- [X] Collection
- [ ] CLI
- [ ] Other

### Installation method

N/A

### Modifications

no

### Ansible version

ansible [core 2.13.1]

### Operating system

Fedora 36

### Steps to reproduce

run `make black` at top level of the project with files that need to be reformatted. 

### Expected results

Files to be reformatted similar to targeting specific files with black. 

### Actual results

```
[thedoubl3j@localhost awx]$ git commit -m "commit msg"
would reformat awx_collection/plugins/lookup/schedule_rrule.py
would reformat awx_collection/plugins/lookup/schedule_rruleset.py

Oh no! 💥 💔 💥
2 files would be reformatted.
To fix this, run `make black` to auto-format your code prior to commit, or set AWX_IGNORE_BLACK=1
[thedoubl3j@localhost awx]$ make black
All done! ✨ 🍰 ✨
800 files left unchanged.
```

### Additional information

current versions locally installed and which is being used
```
[thedoubl3j@localhost awx]$ black --version
black, 22.10.0 (compiled: yes)
Python (CPython) 3.9.15
[thedoubl3j@localhost awx]$ which black
~/.local/bin/black
```

relrod:
I think this is because the pre-commit check runs `black` on against all changed files, without respect to files that `black` is told to ignore:

https://github.com/ansible/awx/blob/fcdff8bdfb6b041cde04c35453a57b843cfd290f/pre-commit.sh#L2-L9

And then we configure `black` to ignore `awx_collection`:

https://github.com/ansible/awx/blob/fcdff8bdfb6b041cde04c35453a57b843cfd290f/pyproject.toml#L5

thedoubl3j:
Is there any reason anyone could think of that we do not extend black to the collection?

AlanCoding:
[...]

PR #14531, opened 2023-10-03
Only block commits if black fails for certain paths

AlanCoding:
##### SUMMARY
Connect https://github.com/ansible/awx/issues/13241

This solution isn't "perfect". It doesn't reuse the ignore list in `pyproject.toml`, but this would be very hard to make work with the overall structure we have. I don't want to let the best be the enemy of the good, and this seems pretty good to me. This doesn't change how the checks work, and I don't foresee any problems with it.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 if [ -z $AWX_IGNORE_BLACK ] ; then
-	python_files_changed=$(git diff --cached --name-only --diff-filter=AM | grep -E '\.py$')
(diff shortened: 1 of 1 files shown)


john-westcott-iv (approved):

john-westcott-iv:
Do you want to change the pre-commit hook as well?

AlanCoding:
> Do you want to change the pre-commit hook as well?


What does this mean?

john-westcott-iv:
> What does this mean?

It means I was in a meeting when I reviewed this and totally spaced, thinking that you were modifying a check in the GH actions. But clearly, you are already modifying the pre-commit hook, lol. My bad 🤦‍♀️

PR #14532, opened 2023-10-03
[rsyslog] remove main_queue, add more action queue params

relrod:
##### SUMMARY

Drop `main_queue`, add a few more action queue parameters controlling things like how big the queue grows, what happens when it grows too large, etc.

NOTE: This removes the old (now unused) `LOG_AGGREGATOR_MAX_DISK_USAGE_GB`, so this should get a major version bump, since anyone setting that setting with the collection will now get an error.

##### ISSUE TYPE

##### COMPONENT NAME

 - API


awx/main/utils/external_logging.py
```diff
@@ -22,6 +22,21 @@ def construct_rsyslog_conf_template(settings=settings):
...
+    queue_options = [
+        f'queue.spoolDirectory="{spool_directory}"',
+        'queue.filename="awx-external-logger-action-queue"',
+        f'queue.maxDiskSpace="{max_disk_space_action_queue}g"',  # overall disk space for all queue files
+        'queue.maxFileSize="100m"',  # individual file size
+        'queue.type="LinkedList"',
+        'queue.saveOnShutdown="on"',
+        'queue.syncqueuefiles="on"',  # (f)sync when checkpoint occurs
+        'queue.checkpointInterval="1000"',  # Update disk queue every 1000 messages
+        'queue.size="131072"',
```
portante: Be aware that this also impacts memory consumption.

awx/main/utils/external_logging.py
```diff
@@ -22,6 +22,21 @@ def construct_rsyslog_conf_template(settings=settings):
     spool_directory = getattr(settings, 'LOG_AGGREGATOR_MAX_DISK_USAGE_PATH', '/var/lib/awx').rstrip('/')
```
portante, line 22: Consider refactoring the repeated constant, `/var/lib/awx`, into a variable and using that instead of all the embedded references.

portante (approved):
Looks good.

portante (commented):
Looks good to me, although perhaps DRY out `test_logger.py`?

awx/main/utils/external_logging.py
```diff
@@ -17,11 +17,26 @@ def construct_rsyslog_conf_template(settings=settings):
...
+        f'queue.maxDiskSpace="{max_disk_space_action_queue}g"',  # overall disk space for all queue files
+        'queue.maxFileSize="100m"',  # individual file size
+        'queue.type="LinkedList"',
+        'queue.saveOnShutdown="on"',
+        'queue.syncqueuefiles="on"',  # (f)sync when checkpoint occurs
+        'queue.checkpointInterval="1000"',  # Update disk queue every 1000 messages
+        f'queue.size="{action_queue_size}"',  # max number of messages in queue
+        f'queue.highwaterMark="{int(action_queue_size * 0.75)}"',  # 75% of queue.size
+        f'queue.discardMark="{int(action_queue_size * 0.9)}"',  # 90% of queue.size
+        'queue.discardSeverity="5"',  # Only discard notice, info, debug if we must discard anything
```
chrismeyersfsu, line 37: I'd imagine some customers think of our external logging as an audit log. Also, we don't do a great job of correctly choosing our log levels in our log level messages.

I'd rather retain the existing behavior of not dropping message on pressure. 

@chadmf Thoughts on this?

awx/main/conf.py
```diff
@@ -694,16 +694,18 @@
     category_slug='logging',
 )
 register(
-    'LOG_AGGREGATOR_MAX_DISK_USAGE_GB',
```
chrismeyersfsu, line 697: I wonder what happens with this setting in the DB? I guess there isn't much harm in letting it continue to live as an orphan in the DB. 

I'm imagining the developer that comes along down the road and picks this same setting name for a different purpose & is confused by the bug reports that come with it. Slim chances, but a funny thought.

Maybe we have an existing pattern for this case. I would think it would be some kind of special migration.

awx/main/utils/external_logging.py
relrod: > I'd rather retain the existing behavior of not dropping message on pressure.

So, I think the idea here is if we hit `queue.maxDiskSpace` GB total or `queue.size` total messages, we will start dropping everything, and with doing it this way (as written) we try to drop the less severe things first.

So if I understand right, the options are:

1) Let `queue.discardSeverity` be 8 (the default) which doesn't try to do any prioritization, it'll just fill up the queue until it can't anymore and then drop everything.

or

2)  Specify `queue.discardSeverity` with some smaller number and drop the less important messages for a while, trying to save space for more urgent messages.

@portante do you have any thoughts here? And is my understanding here correct?

portante: @chrismeyersfsu, if you don't drop messages on pressure, what are you suggesting should happen?

chrismeyersfsu: All of this context helps. Thank you. I think I better understand the problem at hand. Yeah, you have to make a decision here, and deciding to drop by log level is a good decision.

The concern I now have is I'm unsure if this resolves the customers underlying problem. Unsure if it's acceptable to have _any_ of the messages dropped. But that concern is orthogonal to the good work this PR does. So I'm :+1: on this change being an improvement.

chrismeyersfsu (approved):

awx/main/utils/external_logging.py
portante: This is a topic that unfortunately has not been well addressed by the computer science community at large.  Logging is still exactly like plumbing in your house: if the system backups up, you have a stinky mess.

Ideally (see https://www.p99conf.io/session/lets-fix-logging-once-and-for-all/) we'd have the ability of the admin / SRE to control the overall maximum rate of logs that flow in a distributed system, and have individual log producers decide (with a default) what to do what that maximum rate is hit (block or drop).

In this case, given we don't know what the overall rate should be, and we can only discover it via back-pressure (queue filling up), we start dropping to avoid problems once we have already stored as much as we are allowed to locally.

If the user wants to avoid losing any logs, then they should create an infinite disk that can hold all the logs necessary so that they can be transmitted once the connection to the sink it restored.  Or they should ensure that no client ever loses a connection to the sink and that sink can accept every log sent to it.

Back to reality, because we can't actually block the creation of logs in rsyslog, we can only queue them until a certain amount is collected and then start dropping.

chrismeyersfsu: Really cool take on solving this problem. Love the SRE perspective of  1 place to configure to **really** enforce the policy and the general take on solving the root problem here. I like the choice of implementing this mechanism in the kernel to reduce data copy. I've been reading a lot about eBPF but haven't actually played around with it. It seems like a perfect fit here. 

Now you've got me re-thinking about approaching logging. I'm trying to think how, as an app developer, I can take action to adhear to some arbitrary-set SRE limit. My first thought is "How can I code for the extreme case", "What happens if an SRE sets my app to only be able to log 1 message a second?" Let's throw that thought out. How about the normal case, SRE sets the log rate to the observed average and with a policy of block. Now imagine the log load doubling. The result of this would be my app would have random performance issues all over the place.

The above is too simplistic. A complex app would use something like journald or rsyslog to buffer the extremes. But exceeding the bandwidth would still result in backpressure.

Now let's say the makers of journald and rsyslog expose feature and configs to control the rates and policies, essentially passing the control and responsibility up to the App layer.

Now I'm back to my original thought, what can I do as an app developer to adhear to some limit(s)?
* Log less
* Operate in different modes (I'm thinking log levels are pretty close to this)
* Respecting a rate limit in my app leads me down the thought path of just building an rsyslog inside of my app.

Thinking of log provisioning as a rate-limit vs. a fixed high water mark is interesting. The enforcement policies of block or drop are limiting. The SRE should be able to take action and be given a choice of what action to take when a policy is violated.

PR #14562, opened 2023-10-11
ee_reference.rst: refert to Builder's definition docs instead of duplicating its content

Andersson007:
##### SUMMARY
https://docs.ansible.com/automation-controller/latest/html/userguide/ee_reference.html#execution-environment-definition duplicates content from [Builder's docs](https://ansible.readthedocs.io/projects/builder/en/latest/definition/), so we'll be just referring to it to avoid divergence and confusion.

##### ISSUE TYPE

##### COMPONENT NAME

 - Docs
 
##### AWX VERSION

```
latest
```


Andersson007:
ready_for_review

Zokormazo (approved):

docs/docsite/rst/userguide/ee_reference.rst
```diff
@@ -2,304 +2,8 @@
...
-      workdir: /myworkdir
-      user: bob
-
-
-version
-~~~~~~~~
-
-An integer value that sets the schema version of the execution environment definition file. Defaults to ``1``. Must be ``3`` if you are using Ansible Builder 3.x.
-
+For detailed information about the |ee| definition,
```
gundalow, line 5: Does `|ee|` need to be defined in this file, or is it global?

Andersson007: @gundalow i don't see it defined in this file, so i think it's global

Andersson007:
@jbericat @Zokormazo @gundalow thanks for reviewing and merging!

tvo318:
@Andersson007 @gundalow - So the reason why we put the builder v3 content in the automation controller docs was because I was told that we don't want to reference upstream docs for downstream readers. If that is ok, then I'll know next time where duplicate content exist to not duplicate the content.

Andersson007:
@tvo318 thanks for the context, if you recall the rationale for that, it'd be great, maybe it's reasonable, i don't know, but if we use `builder` to build the images i think we should refer to their docs as it's the most relevant and updated, etc.

PR #14566, opened 2023-10-11
Make wait-for-migrations wait forever

### LINKED ISSUE #14314: AWX DB migration gets in loop and upgrade never ends

- [X] I am **NOT** reporting a (potential) security vulnerability. (These should be emailed to `security@ansible.com` instead.)

### Bug Summary
When upgrading from 22.3 to 22.5, we noticed that the upgrade took much longer than expected, was stuck in the DB migration process, and didn't allow AWX to start. This upgrade included one of the [database migration tasks](https://github.com/ansible/awx/blob/devel/awx/main/migrations/0185_move_JSONBlob_to_JSONField.py).

This issue happened because of a bug in AWX DB migration logic. When we upgrade AWX, the AWX operator starts the DB migration. The [following script](https://github.com/ansible/awx/blob/7e4cf859f59aaa865985ade13a51a37e3c41e8f6/tools/ansible/roles/dockerfile/files/wait-for-migrations#L8) is responsible for the retry mechanism.

The script makes a total of 30 retries to check the status of AWX migration. It will wait for a maximum of 60 seconds (TIMEOUT=60) for each attempt. However, an exponential backoff strategy dynamically calculates the waiting time between attempts.

The next_sleep function calculates the waiting time before the next attempt. It starts with MIN_SLEEP (0.5 seconds) and doubles the value with each attempt until it reaches MAX_SLEEP (30 seconds). So, the waiting times between attempts will be as follows:

_______

Attempt 1: 0.5 seconds
Attempt 2: 1 second
Attempt 3: 2 seconds
Attempt 4: 4 seconds
Attempt 5: 8 seconds
Attempt 6: 16 seconds
Attempt 7: 30 seconds (maximum reached)
Attempt 8 and beyond: 30 seconds (maximum reached)

______

After the 7th attempt, the waiting time will remain constant at 30 seconds since it has reached the maximum configured value. Suppose none of the attempts succeeds within the specified TIMEOUT period (60 seconds). In that case, the script will fail with an error message "ERROR: Database migrations not applied" and return with exit code 1 and stop the migration task.

It means that if we have a large DB that requires more time to finish the particular migration task, the migration will never be finished since it will be stuck in a loop.

**The workaround:**
To resolve this issue, we found a workaround with the AWX community and implemented it. To go with a workaround, you should do the following actions:

1. In the K8s deployments in the AWX namespace:
Scale in **awx-operator-controller-manager** replicas to 0
Scale in **awx-task** replicas to 0
Scale in **awx-web** replicas to 0
2. In awx-task deployment, find the following section:

>         - name: awx-task
>           image: pegasus-docker.artifactory.cyberng.com/pegasus-awx:22.5.0
>           args:
>             - /usr/bin/launch_awx_task.sh
>           env:
>             - name: AWX_COMPONENT
>               value: task
>             - name: SUPERVISOR_CONFIG_PATH
>               value: /etc/supervisord_task.conf
>             - name: AWX_SKIP_MIGRATIONS
>               value: '1'
>             - name: MY_POD_UID
>               valueFrom:
>                 fieldRef:
>                   apiVersion: v1
>                   fieldPath: metadata.uid
And replace the args section with
>         - name: awx-task
>           image: pegasus-docker.artifactory.cyberng.com/pegasus-awx:22.5.0
>           args:
>             - awx-manage
>             - migrate
>             - --noinput
>           env:
>             - name: AWX_COMPONENT
>               value: task
>             - name: SUPERVISOR_CONFIG_PATH
>               value: /etc/supervisord_task.conf
>             - name: AWX_SKIP_MIGRATIONS
>               value: '1'
>             - name: MY_POD_UID
>               valueFrom:
>                 fieldRef:
>                   apiVersion: v1
>                   fieldPath: metadata.uid

3. Scale-out awx-task deployment to 1 replica. You should see the following message in the task container logs:
image
4. Wait till the migration is finished. The pod should finish running, and you should see the following log, which shows that the migration is finished:
image
5. Scale out the **awx-operator-controller-manager** deployment to 1 replica. It will automatically delete the task you modified and scale out the AWX deployments.

### AWX version

22.5

- [ ] UI
- [ ] UI (tech preview)
- [X] API
- [ ] Docs
- [ ] Collection
- [X] CLI
- [X] Other

### Installation method

kubernetes

### Modifications

no

### Ansible version

_No response_

### Operating system

_No response_

### Web browser

_No response_

### Steps to reproduce

Upgrade the AWX with the large database from 22.3 to 22.5

### Expected results

DB migration task is running without interruption

### Actual results

DB migration task is interrupted with the pod restart

### Additional information

_No response_

TheRealHaoLiu:
##### SUMMARY
fixes https://github.com/ansible/awx/issues/14314

Removing retry attempt for waiting for db migration to complete 

##### ISSUE TYPE

##### COMPONENT NAME

 - API
 - Other

##### AWX VERSION

```
awx: 23.3.1.dev6+g1f341c1fb4
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/tools/ansible/roles/dockerfile/files/wait-for-migrations
+++ b/tools/ansible/roles/dockerfile/files/wait-for-migrations
@@ -7,7 +7,6 @@ readonly CMDNAME=$(basename "$0")
 
 readonly MIN_SLEEP=0.5
 readonly MAX_SLEEP=30
-readonly ATTEMPTS=30
 readonly TIMEOUT=60
 
 log_message() { echo "[${CMDNAME}]" "$@" >&2; }
@@ -25,7 +24,7 @@ wait_for() {
     local check=1
 
     while true; do
-        log_message "Attempt ${attempt} of ${ATTEMPTS}"
+        log_message "Attempt ${attempt}"
 
         timeout "${TIMEOUT}" \
                 /bin/bash -c "awx-manage check" &>/dev/null
@@ -37,8 +36,7 @@ wait_for() {
                 && return || rc=$?
         fi
 
-        (( ++attempt > ATTEMPTS )) && break
-
+        attempt=$((attempt + 1))
         log_message "Waiting ${next_sleep} seconds before next attempt"
         sleep "${next_sleep}"
         next_sleep=$(next_sleep ${next_sleep})


Zokormazo (approved):

tools/ansible/roles/dockerfile/files/wait-for-migrations
```diff
@@ -25,7 +24,7 @@ wait_for() {
     local check=1
 
     while true; do
-        log_message "Attempt ${attempt} of ${ATTEMPTS}"
+        log_message "Attempt ${attempt}"
```
AlanCoding, line 27: I don't get it. These messages about attempts (not just this but another one a little later) don't make any sense to me after this change.

AlanCoding:
Does this require a supporting operator change? https://github.com/ansible/awx-operator/pull/1589 seemed the obvious one but I'm not getting it.

TheRealHaoLiu:
@AlanCoding it's independent

tools/ansible/roles/dockerfile/files/wait-for-migrations
TheRealHaoLiu: you are right im gonna still increment attempt at the end

tools/ansible/roles/dockerfile/files/launch_awx_web.sh
```diff
@@ -15,6 +15,4 @@ fi
 
 set -e
 
-wait-for-migrations
```
AlanCoding: I want to document this change, because it is deceptively unrelated to the intent of the rest of the change.

This is an unrelated kind of bug fix - because the task container's obligation in this scenario is to serve the "AWX is upgrading" page, and it obviously can't do that if the container hangs out waiting for migrations to finish before serving. So web container should always go straight to serving.

Not true for task container.

AlanCoding (approved):

TheRealHaoLiu:
i reverted the unrelated change of removing wait from launch_awx_web.sh because it seems like it cause some confusion and does not fit this PR

PR #14584, opened 2023-10-18
Set subscription type as developer for developer subscriptions.

unnecessary-username:
##### SUMMARY

The subscription type in AAP controller is enterprise by default except trial SKUs. It seems not that appropriate to show enterprise for developer subscriptions.
This PR aims to set subscription type as developer for developer subscriptions.

##### ISSUE TYPE

##### COMPONENT NAME

 - API

##### AWX VERSION

```
23.3.0
```

##### ADDITIONAL INFORMATION

Active subscriptions that match this criteria: RH00798, RH2262474, RH2262474F3.
As per current aggregation rules, these subscriptions could not be aggregated with any other applicable subscriptions, so no need to consider which subscription type to set when someone upload a manifest with both the aforementioned developer subscriptions and other Ansible subscriptions.

### DIFF
--- a/awx/main/utils/licensing.py
+++ b/awx/main/utils/licensing.py
@@ -324,7 +324,9 @@ class Licenser(object):
     def generate_license_options_from_entitlements(self, json):
         from dateutil.parser import parse
 
-        ValidSub = collections.namedtuple('ValidSub', 'sku name support_level end_date trial developer_license quantity pool_id satellite subscription_id account_number usage')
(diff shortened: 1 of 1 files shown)


CFSNM (approved):

rooftopcellist (approved):

gundalow:
@unnecessary-username Congratulations on your first merged PR

@rooftopcellist @CFSNM Thank you for the reviews & running this against the licensing tests.

PR #14596, opened 2023-10-22
Fix missing service account secret in docker-compose-minikube role

lmo5:
##### SUMMARY

Starting Kubernetes 1.24, Secrets are not automatically generated when Service Accounts are created.
In order to get a "Minikube Bearer Token", the "docker-compose-minikube" role tries to get a service account token.
I suggest that we create a secret of the type "service-account-token" with the same name as the service account and use-it to boostap minikube 

##### ISSUE TYPE
 - Bug

##### COMPONENT NAME
 - Other
 
##### ISSUE TYPE

##### AWX VERSION

```
awx: 4.0.1.dev13275+g420807589a

```

##### ADDITIONAL INFORMATION
Steps to reproduce :
 1 -  install latest minikube (minikube version: v1.31.2)
 2 -  Run "make docker-compose-container-group"
```
TASK [minikube : Register serviceAccount secret name] **********************************************************
fatal: [localhost]: FAILED! => {"msg": "The conditional check 'service_account[\"resources\"][0][\"secrets\"] | length' failed. The error was: error while evaluating conditional (service_account[\"resources\"][0][\"secrets\"] | length): 'dict object' has no attribute 'secrets'. 'dict object' has no attribute 'secrets'\n\nThe error appears to be in '/vagrant/repos/awx/tools/docker-compose-minikube/minikube/tasks/main.yml': line 85, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Register serviceAccount secret name\n  ^ here\n"} 

```


tools/docker-compose-minikube/minikube/templates/rbac.yml.j2
```diff
@@ -36,3 +36,12 @@ roleRef:
...
   name: {{ minikube_service_account_name }}
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ minikube_service_account_name }}
+  namespace: {{ minikube_service_account_namespace }}
+  annotations:
+    kubernetes.io/service-account.name: {{ minikube_service_account_name }}
+type: kubernetes.io/service-account-token
```
TheRealHaoLiu:
```suggestion
type: kubernetes.io/service-account-token

```

TheRealHaoLiu: nit: new line at the end of file

TheRealHaoLiu:
@lmo5 hey there thanks for the contribution and welcome to the AWX community!

gotta say I'm a little surprised that someone is using this make target.

fosterseth (approved):

fosterseth:
@lmo5 thanks for helping maintain our dev environment

PR #14607, opened 2023-10-26
Docs: Include REST API reference content from swagger.json

oraNod:
##### SUMMARY
Includes the swagger.json REST API reference as part of the docs build.

This PR does the following:

- Updates RST and resources to include generated API docs.
- Adds a Python script that fetches "swagger.json" from a "swagger-test" folder in the public s3 bucket.
- Updates the docs build tox session to run the Python script before doing the sphinx-build.

The https://awx-public-ci-files.s3.amazonaws.com/swagger-test/swagger.json file is added as part of a RH internal Jenkins build. As a follow up it would be ideal to replace that Jenkins build with something similar to https://github.com/ansible/awx/blob/bef0a8b23ac34db56aeaebe3eae4ff7c9673f087/.github/workflows/upload_schema.yml#L47

##### ISSUE TYPE

##### COMPONENT NAME
 - Docs

##### AWX VERSION
N/A

##### ADDITIONAL INFORMATION
N/A


oraNod:
I've used this to deploy on RTD from my fork. https://oranod-awx.readthedocs.io/en/latest/rest_api/api_ref.html

oraNod:
Any opinions on naming the folder in the S3 bucket if this approach is OK? I went with `swagger-test` to try it out but if we go ahead we should likely change that to something else. Possibly `swagger-json` or `swagger-docs` or just plain `docs`.

Edit: I changed the folder to `community-docs` so it should be super clear what it is used for. We might also need to add some bits for the generated CLI reference to that same folder so it made sense to have a more generic name.

tvo318 (approved):
Looks good to me!

oraNod:
@TheRealHaoLiu @AlanCoding @thedoubl3j Hi, could one of you - or someone else - please take a look and let me know if you have thoughts or concerns? Thanks.

tox.ini
```diff
@@ -35,4 +35,5 @@ deps =
   -r{toxinidir}/docs/docsite/requirements.in
   -c{toxinidir}/docs/docsite/requirements.txt
 commands =
+  python {toxinidir}/docs/docsite/rst/rest_api/_swagger/download-json.py
```
fosterseth, line 38: any particular reason to use a python call here instead of curl or wget? Not very familiar with tox command limitations

fosterseth:
also, would we be able to use schema.json directly? the only uploaded from this workflow https://github.com/ansible/awx/blob/bef0a8b23ac34db56aeaebe3eae4ff7c9673f087/.github/workflows/upload_schema.yml#L47

what is the difference between schema.json and swagger.json?

tox.ini
oraNod: hey @fosterseth thanks for the review and question. this really came down to the read the docs build moreso than tox. somewhat surprisingly neither wget or curl seemed to be available there. in an effort to make sure the build is consistent across environments the python call seemed like a good way to go.

oraNod:
> also, would we be able to use schema.json directly? the only uploaded from this workflow
> 
> https://github.com/ansible/awx/blob/bef0a8b23ac34db56aeaebe3eae4ff7c9673f087/.github/workflows/upload_schema.yml#L47
> 
> what is the difference between schema.json and swagger.json?

@fosterseth Thanks again for the questions. I did try to build the API reference with the schema.json but it didn't work as expected because it has additional data compared with the swagger file. I'm hopeful that we can tackle that as a follow up and simply extract the necessary bits from the schema though.

oraNod:
Thanks for merging @TheRealHaoLiu

PR #14617, opened 2023-11-01
Do not reference cache on init of db wrapper

AlanCoding:
##### SUMMARY
We have a problem getting a signal 6 when uWSGI workers reach end of their life.

Found out that this doesn't happen if we don't use our custom database wrapper. I would love nothing more than to axe this thing and revert back to `"ENGINE": "django.db.backends.postgresql",`, but feel obligated to make a perfunctory effort to salvage this debugging tool, which is this PR.

We have a string definition of both the cache and database in Django settings. I have no idea what order Django resolves and initializes them in, but one cannot assume the other is initialized in order to work, so this is making the most obvious of obvious fixes to avoid referencing the _redis_ cache in order to create the _postgres_ cursor.

##### ISSUE TYPE

##### COMPONENT NAME
 - API

### DIFF
--- a/awx/main/db/profiled_pg/base.py
+++ b/awx/main/db/profiled_pg/base.py
@@ -15,16 +15,20 @@ __loc__ = LocMemCache(str(uuid.uuid4()), {})
 __all__ = ['DatabaseWrapper']
 
 
+@memoize(ttl=1, cache=__loc__)
+def get_threshold_value():
+    try:
+        return cache.get('awx-profile-sql-threshold')
+    except Exception:
+        # if we can't reach the cache, just assume profiling's off
+        return None
+
+
 class RecordedQueryLog(object):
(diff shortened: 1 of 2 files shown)


awx/main/db/profiled_pg/base.py
```diff
@@ -15,16 +15,20 @@
 __all__ = ['DatabaseWrapper']
 
 
+@memoize(ttl=1, cache=__loc__)
+def get_threshold_value():
+    try:
+        return cache.get('awx-profile-sql-threshold')
+    except Exception:
```
webknjaz, line 22: I suppose it's
```suggestion
    except KeyError:
```
?

AlanCoding: I don't disagree, but that's also not something that is being changed in this patch.

